PR #1237, opened 2017-06-02
Do not install extlibs if SFML_USE_SYSTEM_DEPS is true.

### LINKED ISSUE #1236: Add a flag to cmake build to prevent building/including external deps.
Some people have a sane-ish setup which already includes properly compiled vorbis, openal, etc
and sfml can clobber those libraries when compiled for windows.

Add a flag to prevent this for those with this sort of setup.

ntzrmtthihu777:
If those libs are present at build time, they would be clobbered at install time.
Will close #1236 (at least for me)
Signed-off-by: Marty Plummer <ntzrmtthihu777@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -408,24 +408,26 @@ endif()
 # install 3rd-party libraries and tools
 if(SFML_OS_WINDOWS)
 
-    # install the binaries of SFML dependencies
-    if(ARCH_32BITS)
-        install(DIRECTORY extlibs/bin/x86/ DESTINATION bin)
-        if(SFML_COMPILER_MSVC AND SFML_MSVC_VERSION LESS 14)
-            install(DIRECTORY extlibs/libs-msvc/x86/ DESTINATION lib)
-        elseif(SFML_COMPILER_MSVC)
-            install(DIRECTORY extlibs/libs-msvc-universal/x86/ DESTINATION lib)
-        else()
-            install(DIRECTORY extlibs/libs-mingw/x86/ DESTINATION lib)
-        endif()
-    elseif(ARCH_64BITS)
-        install(DIRECTORY extlibs/bin/x64/ DESTINATION bin)
-        if(SFML_COMPILER_MSVC AND SFML_MSVC_VERSION LESS 14)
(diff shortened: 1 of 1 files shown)


ntzrmtthihu777:
@MarioLiebisch unless I'm misunderstanding the OSX code shortly after my change, it will only install those frameworks if they were used as part of the build (meaning, if system libraries for that are either not present or were manually prevented from use with that cmake arg).

Android would be a simple fix.

MarioLiebisch:
Yes, you're right. Doesn't apply to Mac OS X (although I'm not a real fan of those string comparisons), but the two mobile targets.

ntzrmtthihu777:
@MarioLiebisch does this work for you?

ntzrmtthihu777:
@MarioLiebisch does this work for you?

CMakeLists.txt
```diff
@@ -486,19 +488,23 @@ elseif(SFML_OS_IOS)
...
-    install(FILES extlibs/Android.mk DESTINATION extlibs)
+    if(NOT SFML_USE_SYSTEM_DEPS)
+        # install extlibs
+        install(DIRECTORY extlibs/libs-android/${ANDROID_ABI} DESTINATION extlibs/lib)
+        install(FILES extlibs/Android.mk DESTINATION extlibs)
 
-    # install Android.mk so the NDK knows how to set up SFML
-    install(FILES src/SFML/Android.mk DESTINATION .)
+        # install Android.mk so the NDK knows how to set up SFML
+        install(FILES src/SFML/Android.mk DESTINATION .)
```
MarioLiebisch: This install command has to stay even for `SFML_USE_SYSTEM_DEPS`. This `Android.mk` works similar to `pkg-config`.

ntzrmtthihu777: Achk, misread that. one second.

ntzrmtthihu777:
ah, it looks like I need to tweak  few more things to prevent building against bundled libs... give me a bit.

CMakeLists.txt
ntzrmtthihu777: Should be fixed up now.

eXpl0it3r:

Issue #1224, opened 2017-05-03
Fullscreen does not go over task bars on top and left on Ubuntu 16.04


naezith:
Fullscreen does not go over task bars on top and left on Ubuntu 16.04, SFML 2.4.1.

YouTube's and other games' fullscreen modes work normally.

Here is [a screenshot of it](http://i.imgur.com/D427jpo.png).
Here is the minimal code:
```cpp
#include <SFML/Graphics.hpp>

int main (int argc, char* argv[]) {
    sf::RenderWindow window(sf::VideoMode(1920, 1080), "Window", sf::Style::Fullscreen);

    while (window.isOpen()) {
        sf::Event event;
        while (window.pollEvent(event)) {
            switch(event.type) {
                case sf::Event::Closed:
                    window.close();
                    break;
            }
        }

        window.clear();
        window.display();
    }

    return 0;
}
```

MarioLiebisch:
Always thought this is the expected behavior on Unity, although I might be wrong. Only using Linux for occasional test builds, no full time user here.

naezith:
YouTube's and other games' fullscreen modes work normally like in Windows though.

MarioLiebisch:
Okay, then this should probably be changed. By someone actually knowing X11 * shudder *. Any volunteers?

naezith:
This is very critical for my game. Performance is very poor at windowed mode so fullscreen is a must. Can anyone help with this?

MarioLiebisch:
Tried and clearly too much X11/XRandR shenanigans going on for my likings. It seems like we'd have to use XRandR 1.2 and newer [the way described here](https://www.reddit.com/r/gamedev/comments/2w8ahe/changing_resolutions_for_fullscreen_in_x11_linux/).

naezith:
I just tested @bloodsword 's commit, it fixes the issue. Please add this.

edwrodrig:
Wow! This is a clear reason to NOT use SFML in my current project

danieljpetersen:
This fixes the issue for me.  Tested on Ubuntu 16.04.5 (Unity) in Virtualbox.
* Using latest version on master (5e10e1f0c9b0a5c095df0f8e9be96914f75297d0), fullscreen takes up the entire monitor as expected.
* I chose a build a few commits before the latest master (be3556d76ab956bab63d96b1d117546e6bf9897e) and was able to replicate the issue, fullscreen did not cover the top bar / side bar.

[...]

PR #1238, opened 2017-06-05
Dual monitor fix

### LINKED ISSUE #1226: Fullscreen runs on second monitor and closes the main one in dual monitor setup, Linux
I can't test this myself but a full-time Linux user reported it,

He has two monitors, same model, both **1920x1080**. He opens the game in full-screen, at first launch game picks the **sf::VideoMode::getDesktopMode()** -> **3840x1080** which is both monitors combined. It's not terrible but still is every user has to change it so I consider it as a bad thing. When he changes it to **1920x1080**, game runs on the second monitor only, and **main monitor shuts down**, no signal.

Can someone with similar setup try this?

bloodsword:
**SFML Tasks**

* [ ] Test and confirm fix by multiple people/distros/window managers

----

This should fix #1226.

XRRSetCrtcConfig is now used to set video mode  instead of XRRSetScreenConfig.

### DIFF
--- a/src/SFML/Window/Unix/WindowImplX11.cpp
+++ b/src/SFML/Window/Unix/WindowImplX11.cpp
@@ -173,7 +173,7 @@ namespace
 
         if (result != Success || actualType != XA_WINDOW || numItems != 1)
         {
-            if(result == Success)
+            if (result == Success)
                 XFree(data);
 
             sf::priv::CloseDisplay(display);
@@ -205,7 +205,7 @@ namespace
 
         if (result != Success || actualType != XA_WINDOW || numItems != 1)
         {
-            if(result == Success)
+            if (result == Success)
                 XFree(data);
 
             sf::priv::CloseDisplay(display);
@@ -269,7 +269,7 @@ namespace
             windowManagerName = sf::String::fromUtf8(begin, end);
         }
 
-        if(result == Success)
+        if (result == Success)
             XFree(data);
 
         sf::priv::CloseDisplay(display);
@@ -486,6 +486,7 @@ m_inputMethod    (NULL),
 m_inputContext   (NULL),
 m_isExternal     (true),
 m_oldVideoMode   (0),
+m_oldRRCrtc      (0),
 m_hiddenCursor   (0),
 m_lastCursor     (None),
 m_keyRepeat      (true),
@@ -534,6 +535,7 @@ m_inputMethod    (NULL),
 m_inputContext   (NULL),
 m_isExternal     (false),
 m_oldVideoMode   (0),
+m_oldRRCrtc      (0),
 m_hiddenCursor   (0),
 m_lastCursor     (None),
 m_keyRepeat      (true),
@@ -555,8 +557,17 @@ m_lastInputTime  (0)
     m_screen = DefaultScreen(m_display);
 
     // Compute position and size
-    int left = m_fullscreen ? 0 : (DisplayWidth(m_display, m_screen)  - mode.width) / 2;
-    int top = m_fullscreen ? 0 : (DisplayHeight(m_display, m_screen) - mode.height) / 2;
+    Vector2i windowPosition;
+    if(m_fullscreen)
+    {
+        windowPosition = getPrimaryMonitorPosition();
+    }
+    else
+    {
+        windowPosition.x = (DisplayWidth(m_display, m_screen)  - mode.width) / 2;
+        windowPosition.y = (DisplayWidth(m_display, m_screen)  - mode.height) / 2;
+    }
+
     int width  = mode.width;
     int height = mode.height;
 
@@ -571,7 +582,7 @@ m_lastInputTime  (0)
 
     m_window = XCreateWindow(m_display,
                              DefaultRootWindow(m_display),
-                             left, top,
+                             windowPosition.x, windowPosition.y,
                              width, height,
                              0,
                              visualInfo.depth,
@@ -668,9 +679,11 @@ m_lastInputTime  (0)
     {
         m_useSizeHints = true;
         XSizeHints* sizeHints = XAllocSizeHints();
-        sizeHints->flags = PMinSize | PMaxSize;
+        sizeHints->flags = PMinSize | PMaxSize | USPosition;
         sizeHints->min_width = sizeHints->max_width = width;
         sizeHints->min_height = sizeHints->max_height = height;
+        sizeHints->x = windowPosition.x;
+        sizeHints->y = windowPosition.y;
         XSetWMNormalHints(m_display, m_window, sizeHints);
         XFree(sizeHints);
     }
@@ -707,7 +720,15 @@ m_lastInputTime  (0)
     // Set fullscreen video mode and switch to fullscreen if necessary
     if (m_fullscreen)
     {
-        setPosition(Vector2i(0, 0));
+        // Disable hint for min and max size,
+        // otherwise some windows managers will not remove window decorations
+        XSizeHints *sizeHints = XAllocSizeHints();
+        long flags = 0;
+        XGetWMNormalHints(m_display, m_window, sizeHints, &flags);
+        sizeHints->flags &= ~(PMinSize | PMaxSize);
+        XSetWMNormalHints(m_display, m_window, sizeHints);
+        XFree(sizeHints);
+ 
         setVideoMode(mode);
         switchToFullscreen();
     }
@@ -721,11 +742,11 @@ WindowImplX11::~WindowImplX11()
     cleanup();
 
     // Destroy icon pixmap
-    if(m_iconPixmap)
+    if (m_iconPixmap)
         XFreePixmap(m_display, m_iconPixmap);
 
     // Destroy icon mask pixmap
-    if(m_iconMaskPixmap)
+    if (m_iconMaskPixmap)
         XFreePixmap(m_display, m_iconMaskPixmap);
 
     // Destroy the cursor
@@ -956,10 +977,10 @@ void WindowImplX11::setIcon(unsigned int width, unsigned int height, const Uint8
         return;
     }
 
-    if(m_iconPixmap)
+    if (m_iconPixmap)
         XFreePixmap(m_display, m_iconPixmap);
 
-    if(m_iconMaskPixmap)
+    if (m_iconMaskPixmap)
         XFreePixmap(m_display, m_iconMaskPixmap);
 
     m_iconPixmap = XCreatePixmap(m_display, RootWindow(m_display, m_screen), width, height, defDepth);
@@ -1233,53 +1254,99 @@ void WindowImplX11::setVideoMode(const VideoMode& mode)
         return;
 
     // Check if the XRandR extension is present
-    int version;
-    if (!XQueryExtension(m_display, "RANDR", &version, &version, &version))
+    int xRandRMajor, xRandRMinor;
+    if (!checkXRandR(xRandRMajor, xRandRMinor))
     {
         // XRandR extension is not supported: we cannot use fullscreen mode
         err() << "Fullscreen is not supported, switching to window mode" << std::endl;
         return;
     }
 
-    // Get the current configuration
-    XRRScreenConfiguration* config = XRRGetScreenInfo(m_display, RootWindow(m_display, m_screen));
+    // Get root window
+    ::Window rootWindow = RootWindow(m_display, m_screen);
 
-    if (!config)
+    // Get the screen resources
+    XRRScreenResources* res = XRRGetScreenResources(m_display, rootWindow);
+    if (!res)
     {
-        // Failed to get the screen configuration
-        err() << "Failed to get the current screen configuration for fullscreen mode, switching to window mode" << std::endl;
+        err() << "Failed to get the current screen resources for fullscreen mode, switching to window mode" << std::endl;
+        return;
+    }
+
+    RROutput output = getOutputPrimary(rootWindow, res, xRandRMajor, xRandRMinor);
+
+    // Get output info from output
+    XRROutputInfo* outputInfo = XRRGetOutputInfo(m_display, res, output);
+    if (!outputInfo || outputInfo->connection == RR_Disconnected)
+    {
+        XRRFreeScreenResources(res);
+
+        // If outputInfo->connection == RR_Disconnected, free output info
+        if (outputInfo)
+            XRRFreeOutputInfo(outputInfo);
+
+        err() << "Failed to get output info for fullscreen mode, switching to window mode" << std::endl;
+        return;
+    }
+
+    // Retreive current RRMode, screen position and rotation
+    XRRCrtcInfo* crtcInfo = XRRGetCrtcInfo(m_display, res, outputInfo->crtc);
+    if (!crtcInfo)
+    {
+        XRRFreeScreenResources(res);
+        XRRFreeOutputInfo(outputInfo);
+        err() << "Failed to get crtc info for fullscreen mode, switching to window mode" << std::endl;
+        return;
+    }
+
+    // Find RRMode to set
+    bool modeFound = false;
+    RRMode xRandMode;
+
+    for (int i = 0; (i < res->nmode) && !modeFound; i++)
+    {
+        if (crtcInfo->rotation == RR_Rotate_90 || crtcInfo->rotation == RR_Rotate_270)
+            std::swap(res->modes[i].height, res->modes[i].width);
+
+        // Check if screen size match
+        if (res->modes[i].width == static_cast<int>(mode.width) &&
+            res->modes[i].height == static_cast<int>(mode.height))
+        {
+            xRandMode = res->modes[i].id;
+            modeFound = true;
+        }
+    }
+
+    if (!modeFound)
+    {
+        XRRFreeScreenResources(res);
+        XRRFreeOutputInfo(outputInfo);
+        err() << "Failed to find a matching RRMode for fullscreen mode, switching to window mode" << std::endl;
         return;
     }
 
     // Save the current video mode before we switch to fullscreen
-    Rotation currentRotation;
-    m_oldVideoMode = XRRConfigCurrentConfiguration(config, &currentRotation);
+    m_oldVideoMode = crtcInfo->mode;
+    m_oldRRCrtc = outputInfo->crtc;
 
-    // Get the available screen sizes
-    int nbSizes;
-    XRRScreenSize* sizes = XRRConfigSizes(config, &nbSizes);
+    // Switch to fullscreen mode
+    XRRSetCrtcConfig(m_display,
+                     res,
+                     outputInfo->crtc,
+                     CurrentTime,
+                     crtcInfo->x,
+                     crtcInfo->y,
+                     xRandMode,
+                     crtcInfo->rotation,
+                     &output,
+                     1);
 
-    // Search for a matching size
-    for (int i = 0; (sizes && i < nbSizes); ++i)
-    {
-        XRRConfigRotations(config, &currentRotation);
+    // Set "this" as the current fullscreen window
+    fullscreenWindow = this;
 
-        if (currentRotation == RR_Rotate_90 || currentRotation == RR_Rotate_270)
-            std::swap(sizes[i].height, sizes[i].width);
-
-        if ((sizes[i].width  == static_cast<int>(mode.width)) && (sizes[i].height == static_cast<int>(mode.height)))
-        {
-            // Switch to fullscreen mode
-            XRRSetScreenConfig(m_display, config, RootWindow(m_display, m_screen), i, currentRotation, CurrentTime);
-
-            // Set "this" as the current fullscreen window
-            fullscreenWindow = this;
-            break;
-        }
-    }
-
-    // Free the configuration instance
-    XRRFreeScreenConfigInfo(config);
+    XRRFreeScreenResources(res);
+    XRRFreeOutputInfo(outputInfo);
+    XRRFreeCrtcInfo(crtcInfo);
 }
 
 
@@ -1288,19 +1355,55 @@ void WindowImplX11::resetVideoMode()
 {
     if (fullscreenWindow == this)
     {
-        // Get current screen info
-        XRRScreenConfiguration* config = XRRGetScreenInfo(m_display, RootWindow(m_display, m_screen));
-        if (config)
+        // Try to set old configuration
+        // Check if the XRandR extension
+        int xRandRMajor, xRandRMinor;
+        if (checkXRandR(xRandRMajor, xRandRMinor))
         {
-            // Get the current rotation
-            Rotation currentRotation;
-            XRRConfigCurrentConfiguration(config, &currentRotation);
+            XRRScreenResources* res = XRRGetScreenResources(m_display, DefaultRootWindow(m_display));
+            if (!res)
+            {
+                err() << "Failed to get the current screen resources to reset the video mode" << std::endl;
+                return;
+            }
 
-            // Reset the video mode
-            XRRSetScreenConfig(m_display, config, RootWindow(m_display, m_screen), m_oldVideoMode, currentRotation, CurrentTime);
+            // Retreive current screen position and rotation
+            XRRCrtcInfo* crtcInfo = XRRGetCrtcInfo(m_display, res, m_oldRRCrtc);
+            if (!crtcInfo)
+            {
+                XRRFreeScreenResources(res);
+                err() << "Failed to get crtc info to reset the video mode" << std::endl;
+                return;
+            }
 
-            // Free the configuration instance
-            XRRFreeScreenConfigInfo(config);
+            RROutput output;
+
+            // if version >= 1.3 get the primary screen else take the first screen
+            if ((xRandRMajor == 1 && xRandRMinor >= 3) || xRandRMajor > 1)
+            {
+                output = XRRGetOutputPrimary(m_display, DefaultRootWindow(m_display));
+
+                // Check if returned output is valid, otherwise use the first screen
+                if (output == None)
+                    output = res->outputs[0];
+            }
+            else{
+                output = res->outputs[0];
+            }
+
+            XRRSetCrtcConfig(m_display,
+                             res,
+                             m_oldRRCrtc,
+                             CurrentTime,
+                             crtcInfo->x,
+                             crtcInfo->y,
+                             m_oldVideoMode,
+                             crtcInfo->rotation,
+                             &output,
+                             1);
+
+            XRRFreeCrtcInfo(crtcInfo);
+            XRRFreeScreenResources(res);
         }
 
         // Reset the fullscreen window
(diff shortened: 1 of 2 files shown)


MarioLiebisch:
Ah, very cool. Found a similar solution but couldn't be bothered trying to figure out the correct types etc. :)

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1801,6 +1911,29 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
...
+    {
+        // XRandR extension is not supported
+        err() << "XRandR extension is not supported" << std::endl;
+        return false;
+    }
+
+    // Check XRandR version, 1.2 required
+    if(!XRRQueryVersion(m_display, &xRandRMajor, &xRandRMinor) || xRandRMajor < 1 || xRandRMinor < 2 )
+    {
+        err() << "XRandR is to old" << std::endl;
```
mantognini: s/to/too/

Check will fail once v.2 is release (although it's not for tomorrow).

MarioLiebisch:
I know it's buggy in its current form and I think this is the right way to go, but I don't have a multi monitor setup to actually test this. However, running a program in fullscreen typically means you can't interact with anything on the second screen/desktop since it has exclusive access. At least that's how it's working on Windows. You can't force the ability to continue working on a secondary display.

eXpl0it3r:
Can this be tested in a VM?

MarioLiebisch:
In theory, but never got multiple screens to work properly. I can set two displays, but "Displays" still only lists a single one.

bloodsword:
I changed XRandr check to accept a potential v2.
I fixed code style and  squashed my commits.

The message "Mouse still can't be used in the second screen." was inaccurate, actually it's possible but you have to alt+tab.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -56,7 +56,7 @@
     #include <SFML/Window/Unix/GlxContext.hpp>
     typedef sf::priv::GlxContext ContextType;
 #endif
-
+
```
mantognini: extra spaces

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1061,54 +1063,128 @@ void WindowImplX11::setVideoMode(const VideoMode& mode)
...
-    Rotation currentRotation;
-    m_oldVideoMode = XRRConfigCurrentConfiguration(config, &currentRotation);
+    // If version > 1.2 get the primary screen else take the first screen
+    RROutput output;
+    if ((xRandRMajor == 1 && xRandRMinor >= 3) || xRandRMajor > 1)
+    {
+        output = XRRGetOutputPrimary(m_display, rootWindow);
+
+        // Check if output return is valid, otherwise use the first screen
+        if(output == None)
```
mantognini: missing space after `if`

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1061,54 +1063,128 @@ void WindowImplX11::setVideoMode(const VideoMode& mode)
...
+        err() << "Failed to get the current screen resources for fullscreen mode, switching to window mode" << std::endl;
         return;
     }
 
-    // Save the current video mode before we switch to fullscreen
-    Rotation currentRotation;
-    m_oldVideoMode = XRRConfigCurrentConfiguration(config, &currentRotation);
+    // If version > 1.2 get the primary screen else take the first screen
+    RROutput output;
+    if ((xRandRMajor == 1 && xRandRMinor >= 3) || xRandRMajor > 1)
```
mantognini: If we want to follow strictly our style rules, more parentheses are needed. But I'd argue it's more readable as is. Do others agree? (It's also the case in a few other places.)

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1061,54 +1063,128 @@ void WindowImplX11::setVideoMode(const VideoMode& mode)
...
-    XRRScreenSize* sizes = XRRConfigSizes(config, &nbSizes);
+    // Get output info from output
+    XRROutputInfo* outputInfo = XRRGetOutputInfo(m_display, res, output);
+    if (!outputInfo || outputInfo->connection == RR_Disconnected)
+    {	
+        // Free screen resources
+        XRRFreeScreenResources(res);
+        
+        // If outputInfo->connection == RR_Disconnected, free output info
+        if(!outputInfo)
```
mantognini: missing space

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1061,54 +1063,128 @@ void WindowImplX11::setVideoMode(const VideoMode& mode)
...
+    // Set "this" as the current fullscreen window
+    fullscreenWindow = this;
+
+    // Free screen resources
+    XRRFreeScreenResources(res);
+
+    // Free  output info
+    XRRFreeOutputInfo(outputInfo);
+
+    // Free crtc info
```
mantognini: I think the `// Free <stuff>` comment should be removed -- they don't bring much.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1117,21 +1193,61 @@ void WindowImplX11::resetVideoMode()
...
+        int xRandRMajor, xRandRMinor;
+        if (checkXRandR(xRandRMajor, xRandRMinor))
         {
-            // Get the current rotation
-            Rotation currentRotation;
-            XRRConfigCurrentConfiguration(config, &currentRotation);
+            XRRScreenResources* res = XRRGetScreenResources(m_display, DefaultRootWindow(m_display));
+            if (!res)
+            { 
+                // Failed to get screen resources
```
mantognini: Same: these kind of comments could be removed -- the string following them explain the problem well enough I think.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1801,6 +1917,29 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
...
+    int version;
+    if (!XQueryExtension(m_display, "RANDR", &version, &version, &version))
+    {
+        // XRandR extension is not supported
+        err() << "XRandR extension is not supported" << std::endl;
+        return false;
+    }
+
+    // Check XRandR version, 1.2 required
+    if (!XRRQueryVersion(m_display, &xRandRMajor, &xRandRMinor) || xRandRMajor < 1 || (xRandRMajor == 1 && xRandRMinor < 2 ))
```
mantognini: 👍

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1801,6 +1917,29 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
...
+        return false;
+    }
+
+    return true;
+}
+
 } // namespace priv
 
-} // namespace sf
+}// namespace sf
```
mantognini: space was removed.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1061,54 +1063,128 @@ void WindowImplX11::setVideoMode(const VideoMode& mode)
...
+        return;
+    }
+
+    // Find RRMode to set 
+    bool modeFound = false;
+    RRMode xRandMode;
 
-            // Set "this" as the current fullscreen window
-            fullscreenWindow = this;
+    for (int i = 0; i < res->nmode; i++)
```
mantognini: I know it was like that before, but it would be much more elegant IMO to have `i < res->nmode && !modeFound` as a loop condition and remove the `break statement`, wouldn't it?

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1061,54 +1063,128 @@ void WindowImplX11::setVideoMode(const VideoMode& mode)
...
+        }
 
-    // Search for a matching size
-    for (int i = 0; (sizes && i < nbSizes); ++i)
+        // Failed to get output info
+        err() << "Failed to get output info for fullscreen mode, switching to window mode" << std::endl;
+        return;
+    }
+
+    // Get crtc info from crtc
```
mantognini: X11 & cie being far from trivial, I'd add in the comment what kind of concrete info we're trying to get from CRTC. (screen dimension and rotation?) Otherwise the comment seems of no additional interest than the next function call.

src/SFML/Window/Unix/WindowImplX11.hpp
```diff
@@ -256,6 +257,18 @@ class WindowImplX11 : public WindowImpl
...
+    /// \brief Check if a valid version of XRandR extension is present 
+    ///
+    /// \param xRandRMajor XRandR major version
+    ///
+    /// \param xRandRMinor XRandR minor version
+    ///
+    /// \return True if a valid XRandR version found, false otherwise
+    ///
+    ////////////////////////////////////////////////////////////
+    bool checkXRandR(int &xRandRMajor, int &xRandRMinor);
```
mantognini: Should be `int& x` and not `int &x`.

bloodsword:
I pushed a new revision. 
Sorry for code style. I hope I did not forget anything  :)

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1061,54 +1063,112 @@ void WindowImplX11::setVideoMode(const VideoMode& mode)
...
     }
 
-    // Save the current video mode before we switch to fullscreen
-    Rotation currentRotation;
-    m_oldVideoMode = XRRConfigCurrentConfiguration(config, &currentRotation);
-
-    // Get the available screen sizes
-    int nbSizes;
-    XRRScreenSize* sizes = XRRConfigSizes(config, &nbSizes);
+    // If version > 1.2 get the primary screen else take the first screen
```
mantognini: minor: should be `>=`

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1061,54 +1063,112 @@ void WindowImplX11::setVideoMode(const VideoMode& mode)
...
+    RROutput output;
+    if ((xRandRMajor == 1 && xRandRMinor >= 3) || xRandRMajor > 1)
+    {
+        output = XRRGetOutputPrimary(m_display, rootWindow);
```
mantognini: I'm actually wondering: is this function define in xrandr 1.2? Isn't it instead defined in 1.3+?

src/SFML/Window/Unix/WindowImplX11.cpp
bloodsword: It's actually version > 1.2, but maybe >= 1.3 is more readable.

bloodsword:
You are absolutely right about XRRGetOutputPrimary. 
We can do a compile time check, then no additional version requirement will be needed in CMake. 
As you said 1.3 is quite old now, so it's not really a big deal, but it does not require a lot of changes to do a compile time check, so I pushed a new revision.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1061,54 +1063,111 @@ void WindowImplX11::setVideoMode(const VideoMode& mode)
     if (mode == VideoMode::getDesktopMode())
         return;
 
-    // Check if the XRandR extension is present
-    int version;
-    if (!XQueryExtension(m_display, "RANDR", &version, &version, &version))
+    // Check if the XRandR extension
+    int xRandRMajor, xRandRMinor;
+    if (!checkXRandR(xRandRMajor, xRandRMinor))
```
mantognini: If we don't care about the actual version, but just the presence of the extension, then I'd remove the two parameters of `checkXRandR`. Having the function is fine, but the simpler it is the better.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1061,54 +1063,111 @@ void WindowImplX11::setVideoMode(const VideoMode& mode)
     if (mode == VideoMode::getDesktopMode())
         return;
 
-    // Check if the XRandR extension is present
-    int version;
-    if (!XQueryExtension(m_display, "RANDR", &version, &version, &version))
+    // Check if the XRandR extension
```
mantognini: The comment was truncated, leaving out the end of the sentence.

src/SFML/Window/Unix/WindowImplX11.cpp
mantognini: I'm not enough familiar with X11 & cie, but do we need to actually check for the extension? Or can't we assume that if we were able to link against xrandr the extension will be available at runtime? @binary1248, since you edited those line before, could you clarify that for me. Thanks. :-)

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1061,54 +1063,111 @@ void WindowImplX11::setVideoMode(const VideoMode& mode)
...
-    // Save the current video mode before we switch to fullscreen
-    Rotation currentRotation;
-    m_oldVideoMode = XRRConfigCurrentConfiguration(config, &currentRotation);
+    RROutput output;
 
-    // Get the available screen sizes
-    int nbSizes;
-    XRRScreenSize* sizes = XRRConfigSizes(config, &nbSizes);
+    // Compile time check : if version >= 1.3 get the primary screen else take the first screen
+    #if (RANDR_MAJOR == 1 && RANDR_MINOR >= 3) || RANDR_MAJOR > 1
```
mantognini: The question I'm asking myself is, do really want to support older versions? If not, it would make the code much simpler without those conditional compilation macros. What do @binary1248, @Bromeon and @MarioLiebisch (or anyone who's familiar with Linux actually) think?

src/SFML/Window/Unix/WindowImplX11.cpp
binary1248: As strange as it sounds, I've experienced implementations where not calling `XQueryExtension` would cause the extension to fail even if it should be supported. It seemed like calling `XQueryExtension` would load the extension or something. So even if it should be supported, performing the runtime "check" is still the safer variant.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1061,54 +1063,111 @@ void WindowImplX11::setVideoMode(const VideoMode& mode)
...
         return;
 
-    // Check if the XRandR extension is present
-    int version;
-    if (!XQueryExtension(m_display, "RANDR", &version, &version, &version))
+    // Check if the XRandR extension
+    int xRandRMajor, xRandRMinor;
+    if (!checkXRandR(xRandRMajor, xRandRMinor))
     {
-        // XRandR extension is not supported: we cannot use fullscreen mode
```
binary1248: There is no reason to remove this comment.

src/SFML/Window/Unix/WindowImplX11.cpp
binary1248: I would prefer runtime version checking. It is done in the above code already anyway, and X on Linux doesn't exactly make it trivial to detect library versions during compile time. X works based on Client-Server so it might even be possible that the header you compile against on the system doesn't match what the X server is actually running.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1061,54 +1063,111 @@ void WindowImplX11::setVideoMode(const VideoMode& mode)
...
-            std::swap(sizes[i].height, sizes[i].width);
+    // Get output info from output
+    XRROutputInfo* outputInfo = XRRGetOutputInfo(m_display, res, output);
+    if (!outputInfo || outputInfo->connection == RR_Disconnected)
+    {
+        XRRFreeScreenResources(res);
 
-        if ((sizes[i].width  == static_cast<int>(mode.width)) && (sizes[i].height == static_cast<int>(mode.height)))
+        // If outputInfo->connection == RR_Disconnected, free output info
+        if (!outputInfo)
```
binary1248: This should probably be `outputInfo` and single line `if` blocks omit brackets.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1061,54 +1063,111 @@ void WindowImplX11::setVideoMode(const VideoMode& mode)
...
+            modeFound = true;
         }
     }
 
-    // Free the configuration instance
-    XRRFreeScreenConfigInfo(config);
+    if (!modeFound)
+    {
+        err() << "Failed to find a matching RRMode for fullscreen mode, switching to window mode" << std::endl;
+        return;
```
binary1248: Resources not freed here.

eXpl0it3r:
@bloodsword Can you follow up on @binary1248's review?

bloodsword:
@eXpl0it3r yes. I will try to push something this week.

bloodsword:
I pushed a new revision to fit @binary1248 's review.

mantognini:
Would be nice if you could rebase this work onto master to resolve the conflict. :)

bloodsword:
@mantognini  I have fetched upstream, rebased the dual-monitor-fix branch and pushed the squashed commit. So everything should be OK now :)

mantognini:
thanks!

eXpl0it3r:

src/SFML/Window/Unix/WindowImplX11.hpp
```diff
@@ -265,6 +266,18 @@ class WindowImplX11 : public WindowImpl
     bool processEvent(XEvent& windowEvent);
 
     ////////////////////////////////////////////////////////////
+    /// \brief Check if a valid version of XRandR extension is present 
+    ///
+    /// \param xRandRMajor XRandR major version
+    ///
```
binary1248: Superfluous empty line between params.

binary1248:
So I tested this on my dual monitor system.

Manjaro Linux, Cinnamon WM
Monitors:
Left 1280x1024
Right 1920x1080 (Primary)

Using master, I can get SFML into fullscreen and as the open issues describe, the left monitor gets disabled but the fullscreen application is displayed on the primary monitor as it should without decorations.

Using this branch, I can't get SFML into fullscreen. Window decorations are still displayed, however the mouse cursor is locked within the window region. The window is also always positioned at the left-most virtual position if you stitch all monitors together. Using 1920x1080 resolution in fullscreen, a window of that size was positioned on the left monitor and it spanned a bit into the right monitor with decorations. None of the monitors were disabled. Re-arranging the monitors so the 1280x1024 was virtually to the right of the 1920x1080 caused the window to show up on my 1920x1080 still with decorations.

eXpl0it3r:
@llongi can you take a look at this and maybe test it with your super 100-WM setup? 😄

llongi:
I can take a look this week-end. Though it's a VirtualBox VM so not sure how well the multi-screen support works there anyway. I do have a three-screen setup at home (Xmonad WM), so I can at the very least test this there. Will let you know how it goes.

bloodsword:
I installed a Manjaro with Cinnamon and I got the same problem.
But the version compiled for my game works, this is probably due to the fix I made for the issue #1224. 
I haven't posted a pull request for this bug because the fix is realy ugly.
But I might have a more elegant solution to post. I'll try to post something soon.

bloodsword:
The last commit should fix #1226 and #1224.
I have tested my changes on a dual monitors setup (2 x 1280x1024), on Manjaro with Cinnamon and Ubuntu 16.10.  I haven't tested with two monitors with different resolutions like @binary1248's configuration.

naezith:
I just tested this, works perfectly.

eXpl0it3r:
I tried to run SFML in multi screen mode in a VM, but didn't notice any issue before this PR, so can't really test it.

@binary1248 Can you give this another test run?

binary1248:
I gave this another try.

Current version:

No matter what I do in the code, the fullscreen window always shows up on my left most monitor. This is an extended desktop monitor, my primary monitor is to the right of it.

If I create the window with sf::VideoMode::getDesktopMode() it takes the 1920x1080 of my primary monitor and tries to set it on my left monitor which is only 1280x1024. Like this, mode setting fails and no resolution is adjusted, both screens stay on. If I hard code 1280x1024 as window size, the fullscreen window is created on my left most monitor and my primary monitor is disabled.

Window decorations no longer show up in fullscreen as they did in the last version.

I think to fully solve this problem we need the monitor API that has been in discussion for quite a while.

bloodsword:
Monitor API would give more choice to user, but I think that selecting a monitor wouldn't be mandatory when creating a VideoMode. So the default monitor would be the primary monitor, as it should be in the curent version without the monitor API. So we still have to solve this issue. Also I think we would face the same issue when implementing the monitor API.
I have done more tests. The setup on my test machine is a bit different than the @binary1248 one (my primary screen is 1680x1050).
The behaviour on my test machine is a bit different but I got the same confusion with getDesktopMode result and the resolution of the selected monitor.
I found the issue, it comes from the setPosition(0,0).
I will try to push something soon.

bloodsword:
It took me a while to solve this window position problem :)
I have pushed my commits and updated my repo from upstream.

This code do the XRRGetCrtcInfo mechanics twice if fullscreen is on, that's
a bit "verbose". But I did not want to store X resources as member, since
there are used only when it's fullscreen and only in the constructor.

@binary1248 could you please test this version ?

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1966,6 +2069,106 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
...
+    if (!XQueryExtension(m_display, "RANDR", &version, &version, &version))
+    {
+        err() << "XRandR extension is not supported" << std::endl;
+        return false;
+    }
+
+    // Check XRandR version, 1.2 required
+    if (!XRRQueryVersion(m_display, &xRandRMajor, &xRandRMinor) || xRandRMajor < 1 || (xRandRMajor == 1 && xRandRMinor < 2 ))
+    {
+        err() << "XRandR is to old" << std::endl;
```
dolfinsbizou: too* I guess?

bloodsword: You guessed right :)
I have pushed the fix.

eXpl0it3r:
I got more reports from some people that his doesn't fix the issue. Unfortunately I don't have more details on this.

bloodsword:
To bad there isn't more information. Any information would be welcome (what happens, hardware, distribution, window manager or/and code sample used). 
With this lone statement, it might even be a different bug.

bloodsword:
I don't know what the buggy behaviour looks like since my last commit.  
Prior my last commit,  @binary1248 describes some problems, it should be fixed, but except my own test I have no feedback.
@eXpl0it3r reports that some users had problems, but I have no more details. 
If you manage to produce a bug with my pull request version (https://github.com/bloodsword/SFML/tree/dual-monitor-fix), 
please tell me what is your setup, maybe I will manage to reproduce it and to fix it.

Here is a sample code to test (press escape to quit) :

```
#include <SFML/Graphics.hpp>

int main()
{  
    sf::RenderWindow window( sf::VideoMode::getDesktopMode(), "SFML test", sf::Style::Fullscreen);

    while (window.isOpen())
    {        
        sf::Event event;
        {  
            if ((event.type == sf::Event::Closed) || 
               ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Escape)))
            {
                window.close();
                break;
            }
        } 
        window.clear();
        window.display();
        sf::sleep(sf::milliseconds(16));
    }

    return EXIT_SUCCESS;
}
```

The normal behaviour should be a black fullscreen on one of your screen, your other screen should stay active.
Mouse cursor should be locked in the fullscreen monitor except if you alt+tab.

binary1248:
I attribute the test failure during my last test in part to my "exotic" screen setup. I can't test with 2 screens of the same resolution because... well... I simply don't have 2 screens with the same resolution. If others who can test using screens with the same resolution report that this is still an improvement over the state without it, I am willing to accept it as a partial solution on the way to something more complete.

bloodsword:
I have no problem with same resolution screens (1280x1024 x 2), 
and since my last commit (01 November 2017) I have no problem with different screens either(1680x1050 + 1280x1024).
@binary1248 did you try the latest commit ? If you did, any improvement since your 16 October review ?

DanielJPetersen:
I just tried the linked commit (e18b9d2540d8d3b0551ba09c582087c183e4e1c0) out.

I'm on Linux Mint 18.3 and using two 1920x1080 monitors.   The window appears across both of my monitors.  There's a gap across the top of each screen with my desktop background showing.  The gap appears to be the exact size the window titlebar would be.

Here's a screenshot to showing it happen.  I left the monitor app open to show the monitor info in the top left of each screen.

screenshot from 2018-04-09 16-07-42

I used the example code posted above, except I added a circle to it as well (to see if it stretched the view, which has happened to me in the past with Windows though I guess that's another issue).  For the sake of completeness this was the test code I used

```
#include <SFML/Graphics.hpp>

int main()
{  
    sf::RenderWindow window( sf::VideoMode::getDesktopMode(), "SFML test", sf::Style::Fullscreen);

    sf::CircleShape shape(100.f);

    while (window.isOpen())
    {        
        sf::Event event;
        {  
            if ((event.type == sf::Event::Closed) || 
               ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Escape)))
            {
                window.close();
                break;
            }
        } 
        window.clear();
        window.draw(shape);
        window.display();
        sf::sleep(sf::milliseconds(16));
    }

    return EXIT_SUCCESS;
}
```

DanielJPetersen:
Actually, I take all of that back.  I did a double check and the damn thing was still linking to the current version of master.  e18b9d2540d8d3b0551ba09c582087c183e4e1c0 fixes the issue for me.

bloodsword:
No problem. When I booted my Mint trying to reproduce the behaviour you describe I did the same mistake :)
Thanks for your feedback :+1:

fahlmant:
Using Fedora 28, KDE, two screens at 1080p resolution, using the sample code above, this branch does not fix it for me. I get one screen completely black, and the other still shows the desktop, with the mouse locked in the black screen. The second screen also flickers.

fahlmant:
(Made sure I grabbed the right commit this time). Commit `e18b9d2` did not work for me. I do see different behaviour between this and master though. When starting the executable with a terminal, the master branch uses the display the terminal is in to start the full screen. When starting the executable with this branch from either display, the full screen starts on the primary display every time, and there is no flickering of the other screen. Running Fedora 28, KDE Plasma, two screens running at 1920x1080.

eliasdaler:
Tested on Ubuntu 18.10, NVIDIA driver 396.51. Both "master" and this branch work as expected for me...

One monitor is 1366x768. The other is 1920x1080. The fullscreen shows up on the primary monitor.

bloodsword:
Nice to read your feedbacks.
Tell me if you need me to rebase the PR on master.

eXpl0it3r:

eXpl0it3r:
Merged in 46ce05cd9c336020e911a1f6a3486132dc31b679

PR #1244, opened 2017-06-16
fix warnings C26496

fsb4000:
Hi. I'm learning SFML and reading the book: https://www.packtpub.com/game-development/beginning-c-game-programming
While learning examples from the book I am getting the warnings:
C26496	Variable 'r1MinX' is assigned only once, mark it as const. (con.4: https://go.microsoft.com/fwlink/p/?LinkID=784969)
....
(other similar)

Thanks for the cool library!

### DIFF
--- a/include/SFML/Graphics/Rect.inl
+++ b/include/SFML/Graphics/Rect.inl
@@ -78,10 +78,10 @@ bool Rect<T>::contains(T x, T y) const
     // Rectangles with negative dimensions are allowed, so we must handle them correctly
 
     // Compute the real min and max of the rectangle on both axes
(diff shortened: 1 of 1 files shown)


mantognini:
I'm not a big fan of having `const` on pretty much all variables, especially in small & trivial functions, because it makes the code more heavy to read. But one could argue that accepting this PR is a good thing as it reduces the warnings in the user projects so...

LaurentGomila:
I'm not in favor of such a change. If we do this, then we should also mark 'const' half of the local variables in SFML functions, as well as 95% of arguments passed by value. That's really verbose, and as far as I remember, this is not really a recommenced practice in C++.

Issue #1245, opened 2017-06-21
Very minor bug in Window\Win32\WglContext.cpp


GatoRat:
In function "String getErrorString(DWORD errorCode)"

`FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM, NULL, errorCode, 0, errBuff, sizeof(errBuff), NULL);`

Should be:

`FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM, NULL, errorCode, 0, errBuff, _countof(errBuff), NULL);`

Crisspl:
`sizeof` for compile-time-known-size arrays will return (or rather be replaced by) `size * sizeof(T)` where `T` is type of array's elements. What's more - `_countof` is nonstandard macro defined on msvc.

jcowgill:
@LaurentGomila Isn't this a genuine bug? If `sizeof(TCHAR)` is not 1, this might cause a stack buffer overflow.

MarioLiebisch:
No, since `errBuff` is also of type `TCHAR`, so it will always have the correct length. In this example, `sizeof(errBuff)` will be equal to `256 * sizeof(TCHAR)`.

jcowgill:
Yeah but `FormatMessage` takes the number of `TCHAR`s, not the size of the buffer (if I'm reading MSDN correctly).

MarioLiebisch:
Yep, rereading MSDN it seems  you're actually right. It's supposed to be the number of `TCHAR` rather than the memory/buffer size.

So simple solution would be using `256` in this case.

MarioLiebisch:
Just rewrote the function in question to just use system allocated memory (something Joystick implementation does as well).

LaurentGomila:
In my opinion, TCHAR and the whole set of types and macros that depend on UNICODE, should never be used. We should directly call the A or W version of the function with either char or wchar_t, and correctly deal with that.

MarioLiebisch:
Yep, I agree. We should just use the wide character versions everywhere. All supported Windows versions support it anyway.

GatoRat:
FWIW, I usually go with what LaurentGomila suggeted; be explicit with char/wchar_t and A or W.

MarioLiebisch:
Yep, but think should do that in a separate commit with other similar changes. There are many calls like that hidden throughout the Windows implementation.

PR #1246, opened 2017-06-22
Fixed potential buffer overrun in Win32 OpenGL error handling

MarioLiebisch:
This fixes issue #1245.

### DIFF
--- a/src/SFML/Window/Win32/WglContext.cpp
+++ b/src/SFML/Window/Win32/WglContext.cpp
@@ -66,12 +66,10 @@ void ensureExtensionsInit(HDC deviceContext)
 ////////////////////////////////////////////////////////////
 String getErrorString(DWORD errorCode)
 {
-    std::basic_ostringstream<TCHAR, std::char_traits<TCHAR> > ss;
-    TCHAR errBuff[256];
-    FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM, NULL, errorCode, 0, errBuff, sizeof(errBuff), NULL);
-    ss << errBuff;
-    String errMsg(ss.str());
-
+    PTCHAR buffer;
+    FormatMessage(FORMAT_MESSAGE_MAX_WIDTH_MASK | FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM, NULL, errorCode, 0, reinterpret_cast<PTCHAR>(&buffer), 0, NULL);
+    String errMsg(buffer);
+    LocalFree(buffer);
     return errMsg;
 }
 


src/SFML/Window/Win32/WglContext.cpp
```diff
@@ -66,12 +66,10 @@ void ensureExtensionsInit(HDC deviceContext)
...
 String getErrorString(DWORD errorCode)
 {
-    std::basic_ostringstream<TCHAR, std::char_traits<TCHAR> > ss;
-    TCHAR errBuff[256];
-    FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM, NULL, errorCode, 0, errBuff, sizeof(errBuff), NULL);
-    ss << errBuff;
-    String errMsg(ss.str());
-
+    PTCHAR buffer;
+    FormatMessage(FORMAT_MESSAGE_MAX_WIDTH_MASK | FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM, NULL, errorCode, 0, reinterpret_cast<PTCHAR>(&buffer), 0, NULL);
```
binary1248: Is `PTCHAR` the right type for the cast?

MarioLiebisch: Yes and no. While you're supposed to pass a `**TCHAR` to retrieve a system allocated buffer, the signature only accepts a `*TCHAR` (i.e. `PTCHAR`). Remember, we're in C land here, so no overloads. :)

binary1248: `buffer` is already a `PTCHAR`, so the `&` operator would make it a `PTCHAR*`?

MarioLiebisch: Yes, you basically cast away the second layer of reference for the signature alone. You'll pass a `PTCHAR*`, because it's a return value here.

binary1248: Why not cast it to a `LPTSTR` then?

MarioLiebisch: I actually copied the snippet from SFML's Joystick handling. That would probably be an even "more correct" solution I guess.

GatoRat: This still has two major bugs:

The FORMAT_MESSAGE_IGNORE_INSERTS flag should be added to the call.
buffer **MUST** be checked for nullptr before being used and/or freed.

eXpl0it3r:
Bump.

Any alternative solution?

binary1248:
@MarioLiebisch 

```cpp
String getErrorString(DWORD errorCode)
{
    PTCHAR buffer;
    if (FormatMessage(FORMAT_MESSAGE_MAX_WIDTH_MASK | FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS, NULL, errorCode, 0, reinterpret_cast<LPTSTR>(&buffer), 256, NULL) != 0)
    {
        String errMsg(buffer);
        LocalFree(buffer);
        return errMsg;
    }

    std::ostringstream ss;
    ss << "Error " << errorCode;
    return String(ss.str());
}
```

eXpl0it3r:
@binary1248 can you re-review this? I applied the suggested change.

PR #1248, opened 2017-07-01
Added support for (some) Hat/POV axis

mantognini:
This adds support for POV (of Hat) axes. It was solely tested with a Dualshock 4 but should gracefully fail (with a report on `sf::err`) when the assumptions behind this implementation don't hold.

### DIFF
--- a/src/SFML/Window/OSX/JoystickImpl.cpp
+++ b/src/SFML/Window/OSX/JoystickImpl.cpp
@@ -147,19 +147,16 @@ bool JoystickImpl::isConnected(unsigned int index)
                     // If there exists a device d s.t. there is no j s.t.
                     // m_locationIDs[j] == d's location then we have a new device.
 
-                    for (CFIndex didx(0); didx < size; ++didx)
+                    for (CFIndex didx(0); !state && didx < size; ++didx)
                     {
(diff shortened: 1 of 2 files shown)


src/SFML/Window/OSX/JoystickImpl.cpp
```diff
@@ -255,37 +238,79 @@ bool JoystickImpl::open(unsigned int index)
...
+                        if (IOHIDElementGetCollectionType(element) != kIOHIDElementCollectionTypeApplication)
+                        {
+                            sf::err() << std::hex << "Gamepage (vendor/product id: 0x" << m_identification.vendorId
+                                      << "/0x" << m_identification.productId << ") is not an CA but a 0x"
+                                      << IOHIDElementGetCollectionType(element) << std::dec << std::endl;
+                        }
+                        break;
+
+                    default:
+                        sf::err() << "Unexpected usage for element of Page Generic Desktop: 0x" << std::hex << IOHIDElementGetUsage(element) << std::dec << std::endl;
```
mantognini: This should be disabled in release builds.

eXpl0it3r:
Does this need adjustment for other OS?

mantognini:
As far as I know, only Mac is concerned by this.

mantognini:
I've rebased this. I believe it can be moved to "Ready".

PR #1260, opened 2017-08-03
Updated OpenAL Soft to 1.18.1 for Windows.

### LINKED ISSUE #1247: The current openal32.dll appears to have problems with new hardware and Windows 10
The current openal32.dll appears to have problems with new hardware and Windows 10, at least in 64-bit. It will throw an invalid access exception after about three minutes. We observed this on a Surface Tablet running Windows 10 Creators Update. In the forums, there is a very similar report on a Dell XPS ([https://en.sfml-dev.org/forums/index.php?topic=22000.0](https://en.sfml-dev.org/forums/index.php?topic=22000.0).)

I determined that SFML is using OpenAl 1.15.1, which is from December 2012. I took the release binary for 1.18.0, renamed it to openal32.dll, replaced the earlier version and the software has been running without issue for almost 18 hours.

I suggest replacing the opeal32.dll with the latest version.

eXpl0it3r:
This should fix #1247.


eXpl0it3r:
Just tested all the binaries [provided by our CI system](https://www.sfml-dev.org/artifacts/by-branch/feature/update_openal/) and they all work fine. If someone has Windows XP running or some other system than Windows 10, feel free to test this as well.

mantognini:
Should we take this opportunity to update the binaries for other OSes as well? E.g. OS X is at 1.16.

eXpl0it3r:
Feel free to do so. 😉

mantognini:
Okay, I'll *try* to find the time before the end of the month. Don't block this one on me though, I'll create another PR. ;-)

eXpl0it3r:
Tested on Windows XP as well.

How's it looking @mantognini? Otherwise I'll just go ahead and set this up for merging.

mantognini:
Go ahead, I won't have time to do this soon. And in any case it will be in another PR.

If someone else wants to do it, please have fun.  :)

eXpl0it3r:
If someone wants to give this a positive review, as I can't do it myself. 😉

PR #1263, opened 2017-08-08
Bugfix: ALCheck.hpp #include issue for compiling iOS

KRMisha:
**SFML Tasks**

* [ ] Create a corresponding issue & decline this PR

----

Modified the #include paths for OpenAL in ALCheck.hpp inside the #ifdef which checks for iOS. Before this change, attempting to compile SFML's static libs for iOS after building them with CMake would throw an error and fail. This changed fixes this issue, and I can confirm audio works in a compiled iOS project using libraries compiled with this fix.

[Link to forum thread](https://en.sfml-dev.org/forums/index.php?topic=22350.0)

To test, build CMake and compile it with Xcode with these [forum instructions](https://en.sfml-dev.org/forums/index.php?topic=13716.0) to validate that compiling the libraries does not fail.

### DIFF
--- a/src/SFML/Audio/ALCheck.hpp
+++ b/src/SFML/Audio/ALCheck.hpp
@@ -30,8 +30,8 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Config.hpp>
 #ifdef SFML_SYSTEM_IOS
-    #include <OpenAl/al.h>
-    #include <OpenAl/alc.h>
+    #include <AL/al.h>
+    #include <AL/alc.h>
 #else
     #include <al.h>
     #include <alc.h>


eXpl0it3r:
Looking at this again a bit more closely, I think this is not the right way to go and more things need to be adjust to let SFML use OpenAL on iOS.

The reason it might want the headers in `AL/` is because the `extlibs/headers` directory might have been added to the include search directories. However, since the SFML audio module hasn't been really made available on iOS, the [search paths for iOS](https://github.com/SFML/SFML/blob/master/src/SFML/Audio/CMakeLists.txt#L58) [are not adjusted](https://github.com/SFML/SFML/blob/master/src/SFML/Audio/CMakeLists.txt#L81). Additionally, the [OpenAL package isn't searched for](https://github.com/SFML/SFML/blob/master/src/SFML/Audio/CMakeLists.txt#L70).

However, OpenAL is still [being manually linked](https://github.com/SFML/SFML/blob/master/src/SFML/Audio/CMakeLists.txt#L90), which makes me question things again a bit. Since OpenAL is shipped natively on iOS, we have to make sure that it's found somehow, question just is how and whether we can use our own headers or whether we must use the iOS SDK headers.

mantognini:
@eXpl0it3r, your analysis seems good.

> question just is how and whether we can use our own headers or whether we must use the iOS SDK headers.

We would clearly be better off and use the header corresponding to the runtime lib, hence not ours. Hopefully the two versions should be compatible and therefore not require changes in SFML code base.

JonnyPtn:
Looking at this, I'd say the best way would be to find the openAL include directory when configuring and add to the include paths, as such:

Add this code in an `else()` condition [here](https://github.com/SFML/SFML/blob/master/src/SFML/Audio/CMakeLists.txt#L83)
```
    find_path(OPENAL_IOS_DIR al.h)
    include_directories(${OPENAL_IOS_DIR})
```

Then you can remove the conditional in the code entirely and just `#include <al.h>` and `#include <alc.h>`

If you're happy with this I can create a new PR with these changes I've made?

eXpl0it3r:
Does removing the if barrier above for find_host_package really not fill the correct include directory?

JonnyPtn:
It does, but if I understand correctly that will pick up any OpenAL in the search path, which may well not be the iOS one?

eXpl0it3r:
If that's the case, which beyond the original reported nobody has confirmed, i.e. might just be a special setup, then how is `find_path(OPENAL_IOS_DIR al.h)` going to solve this? `al.h` exists in both locations, so who guarantees that it will set the correct one?

JonnyPtn:
Oddly, I'm struggling to reproduce this now :S

@KRMisha can you still reproduce it? If so, can you add more info about your situation, e.g. target sdk, architecture, Xcode version, etc. The more the better!

JonnyPtn:
@KRMisha thanks for the info. I'm considering dropping 32 bit support anyway, as it's almost completely deprecated.

Have you tried using the new iOS toolchain from #1269 ?

JonnyPtn:
As oddly as this disappeared for me, it has now returned! 

I've investigated further, and basically I think we can remove all the special handling of openAL on iOS. I have a feeling it only had special handling because the audio module was disabled, because I can't see any other reason why it was done that way.

@KRMisha can you test the changes I've made [here](https://github.com/JonnyPtn/SFML/tree/ios_find_openal), and see if it solves the problem for you?

KRMisha:
I agree that there isn't much point to keeping 32-bit support anymore.

I tried building and compiling SFML with your branch (using the new toolchain file) and it worked perfectly!

JonnyPtn:
Wunderbar! Thanks for testing

eXpl0it3r:
Superseded by #1376

PR #1265, opened 2017-08-19
Fix Deadlock in Android Main Cleanup

JeffSwenson:
The SFML Android implementation hangs when the main function returns. We noticed this when we added an exit button to our game. Clicking the exit button would background the activity. If you tried to open the game after closing it, it opened to a blank screen. I updated the SFML example so that it closes after 10 taps. This allows you to test closing the android activity and it reproduces the issue we noticed in our game.

**Root Cause: Activity Thread Waits for Game Thread**
There are a few loops in the activity thread that wait for some action on the main thread. If the main thread has exited, the actions will never occur, and the activity thread will wait forever.

**Root Cause: Looper Gets Deleted**
By default, android will delete the game thread looper when the game thread is destroyed. The activity thread needs the looper to clean up the activity's input queue. During cleanup the activity tries to detach the input queue from a deleted looper and it deadlocks trying to lock a deleted mutex stored on the looper. Android provides an acquire/release API that is meant for reference counting a looper. The application must reference count the looper when passing it between threads.

### DIFF
--- a/src/SFML/Main/MainAndroid.cpp
+++ b/src/SFML/Main/MainAndroid.cpp
@@ -93,6 +93,13 @@ static void initializeMain(ActivityStates* states)
     ALooper* looper = ALooper_prepare(ALOOPER_PREPARE_ALLOW_NON_CALLBACKS);
     states->looper = looper;
 
+    /**
+     * Acquire increments a reference counter on the looper. This keeps android 
+     * from collecting it before the activity thread has a chance to detach its 
+     * input queue.
+     */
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
Any chance you can rebase this and remove the changes to the example, @JeffSwenson?

JeffSwenson:
Will do!

JeffSwenson:
I've removed the example app changes.

eXpl0it3r:
Merged in be3556d76ab956bab63d96b1d117546e6bf9897e

Issue #1228, opened 2017-05-07
Linux: Window getPosition returning incorrect position


JonesBlunt:
Related to #346, #372 

Testing on Ubuntu/GNOME 3 - getPosition() output differs quite drastically depending on if window does/doesn't have decorations

New to X11, though it seems `XTranslateCoordinates()` is meant to use (0, 0) for src_x/y as they are relative to the source window origin. Works on my build minus the decorations discussed in #372, though that's good for me.

https://gist.github.com/JonesBlunt/7c396e2341b69362f0d940c1b955e0e0/revisions

llongi:
I'm having a similar problem. I'm running the XMonad WM, tested on Gentoo, Fedora 25 and Fedora 26.
SetPosition() works fine, but getPosition() always returns double the expected value, as if it's adding itself. X11 utils like xwininfo report the correct position, so I don't think this is a WM or X11 bug.
So I added a few print statements to getPosition(), and it seems like localX and localY from XGetGeometry() already contain the position, which then XTranslateCoordinates() uses as offset. Just using XTranslateCoordinates() with 0/0 as source X,Y fixes this in all my tests, see https://gist.github.com/llongi/23e326ac46eed275a6be6ff5b32a5298 for a patch (that also removes XGetGeometry(), since we can get the right root window by using DefaultRootWindow(m_display) as it's done in other parts of the X11 SFML code). This is the same as GLFW does (https://github.com/glfw/glfw/blob/master/src/x11_window.c#L1982), they also had a lot of discussions and testing in https://github.com/glfw/glfw/issues/800 and in the end this was the most consistent behavior, so that getPosition() and setPosition() return the same value.

llongi:
Small test program here: https://gist.github.com/llongi/e02ea9d824572a0854a032c10b35d136 it just creates a 100x100 window, puts it at position 100,100 and on close prints the position as returned by getPosition(), which should be, in a perfect world, 100,100 again.
So with the change discussed in this bug, it is 100,100 on XMonad, and I just tested Gnome 3 too, there it is 100,137 (the +37 comes from the thickness of the toolbar).
Basically the behavior as I can see it is that with the change, X/Y as returned by getPosition() correspond to "Absolute upper-left X/Y" as displayed by xwininfo. Without the change, as the code is now, the returned X/Y are a sum of "Absolute upper-left X/Y" and "Relative upper-left X/Y", which is in all cases I've tested an (even more) incorrect number.
[...]

PR #1266, opened 2017-08-21
Window/Unix: fix SFML issue #1228, getPosition() returns incorrect results on Linux

llongi:
Fixes SFML issue #1228, getPosition() returns incorrect results on Linux depending on the used WM, as well as not returning values that are in-sync with what was given to setPosition(x, y).

See https://github.com/SFML/SFML/issues/1228#issuecomment-323557761 for a complete description of current patch, as well as testing results comparing current approach to approach in patch for over 20 different window managers.

### DIFF
--- a/src/SFML/Window/Unix/WindowImplX11.cpp
+++ b/src/SFML/Window/Unix/WindowImplX11.cpp
@@ -66,6 +66,8 @@ namespace
     sf::Mutex                             allWindowsMutex;
     sf::String                            windowManagerName;
 
+    sf::String                            wmAbsPosGood[] = { "Enlightenment", "FVWM", "i3" };
+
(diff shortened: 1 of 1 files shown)


LaurentGomila:
It's a huge work that you've done, many thanks.

I'm not a X11/WM expert, so here is my only comment: some parts or your patch don't follow our [coding guidelines](https://www.sfml-dev.org/style.php), you should fix them before we can merge your work.

eXpl0it3r:

llongi:
Squash done.

eXpl0it3r:

eXpl0it3r:
Merged in 58b7c2c48a7aa6eef1160862fb1805fedd33818f 🎉

PR #1267, opened 2017-08-24
fix gcc 5 bug

edukaj:
Unfortunately is not possible build SFML on Ubuntu Xenial (16.04) due to a GCC bug.
Further information on [this link](https://bugs.launchpad.net/ubuntu/+source/gcc-5/+bug/1568899).

### DIFF
--- a/src/SFML/Graphics/CMakeLists.txt
+++ b/src/SFML/Graphics/CMakeLists.txt
@@ -153,6 +153,12 @@ if(SFML_COMPILER_GCC)
     set_source_files_properties(${SRCROOT}/ImageLoader.cpp PROPERTIES COMPILE_FLAGS -fno-strict-aliasing)
 endif()
 
+# see https://bugs.launchpad.net/ubuntu/+source/gcc-5/+bug/1568899
+if(SFML_COMPILER_GCC AND BUILD_SHARED_LIBS)
+    message(WARNING "Applying workaround for https://bugs.launchpad.net/ubuntu/+source/gcc-5/+bug/1568899")
+    list(APPEND GRAPHICS_EXT_LIBS "-lgcc_s -lgcc")
+endif()
+
 # define the sfml-graphics target
 sfml_add_library(sfml-graphics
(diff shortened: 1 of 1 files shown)


mantognini:
If it's an upstream bug, it should be fixed upstream, not here, in my opinion.

MarioLiebisch:
If it's an issue with your local build system – which it clearly is – adjust your local `CMAKE_EXE_LINKER_FLAGS` (or equivalent matching  your target type) in your build directory with these flags.

I don't think linking both static and shared(?) version of gcc specific extensions is the way to go here, as I could imagine this might cause unintended side effects regarding dependencies.

edukaj:
@MarioLiebisch It's not my build system, unfortunately, it's a bug of GCC. Users on Ubuntu 16.04 are unable to build SFML (`cmake .. && make -j16` fail).

eXpl0it3r:
Warning about bugs in third-party tools is not something we want to do in SFML, otherwise we'd start collection tons and tons of warnings, as all software has some bugs.

Besides that, generating a warning for every GCC and shared library build is not an option. Not everyone compiles SFML on your specific Ubuntu version and even fewer run the broken compiler version. 😉

edukaj:
@eXpl0it3r I agree with you. But that's not a warning, it's a linkage error due to a bug in the compiler. GCC 5, unfortunately, is the default compiler of Ubuntu Xenial.
By the way, I signaled it.

eXpl0it3r:
> But that's not a warning

```
message(WARNING "Applying workaround for https://bugs.launchpad.net/ubuntu/+source/gcc-5/+bug/1568899")
```

Looks like a warning to me.

That you then proceed to just link against libgcc and libgcc_s is even more questionable, as I wouldn't want that on Windows when using MinGW (which also matches GCC and shared library).

texus:
What about updating the stb_image extlib, wouldn't that also solve the issue?

eXpl0it3r:
Looks like they changed things in [stb_image a while back](https://github.com/nothings/stb/pull/428), so updating it might indeed be an option now.

PR #1269, opened 2017-08-25
Use a CMake toolchain file for iOS build

### LINKED ISSUE #1268: iOS cmake build should use a toolchain file
**Problem:** Currently SFML on iOS builds itself with CMake but it does not use a toolchain file. Instead, it sets the necessary CMake variables itself instead of relying on the external toolchain file. It's worth noting that SFML **does** use a toolchain file for android. 

**Suggested Solution:** Ship an iOS toolchain file with SFML and have the user specify CMAKE_TOOLCHAIN_FILE=/path/to/toolchain.cmake instead of IOS=True

tlein:
**SFML Tasks**

* [x] Rebase PR
* [x] Review code

----

Resolves #1268 

**Description**
This changes SFML to use a toolchain file when building for iOS instead of managing all the necessary CMake values for iOS build itself. This is the standard pattern for cross compiling CMake projects and is used by SFML for Android build. 

This updated the audio package as well even though it is currently being excluded from iOS so once that is fixed this will be compatible with the audio package.

The CMake file was taken from [https://github.com/jomof/ios-cmake](https://github.com/jomof/ios-cmake). The license is BSD and I included it at the top of the toolchain file.

**Follow ups**
The tutorial on the wiki should be updated once this PR is in to reference the toolchain file instead of the IOS=True variable.

**Testing**
Open a terminal in the SFML folder and run these commands:
`mkdir build && mkdir build/ios && mkdir build/ios-install`
`cd build/ios`
`cmake ../.. -DCMAKE_INSTALL_PREFIX=../ios-install/ -DCMAKE_TOOLCHAIN_FILE=../../cmake/toolchains/iOS.toolchain.cmake -G Xcode`
`cmake --build . -- -sdk iphoneos -arch arm64`

The configuration and build should complete without a problem.


tlein:
Note, there is an unrelated build error in iOS so while the CMake configuration should succeed the build will fail. This is expected for this pull request and the build error is discussed in #1271

eXpl0it3r:
When we add a BSD licensed file, do we need to add a link to the license in our license file? @LaurentGomila

LaurentGomila:
> When we add a BSD licensed file, do we need to add a link to the license in our license file? @LaurentGomila

I'm not sure, but I don't think so.

tlein:
I don't think it's necessary, but adding a link to it or just saying "iOS toolchain file licensed under BSD" in license.md wouldn't hurt. If you want I could go ahead and add that in this PR.

eXpl0it3r:
I think it's fine the way it is right now. Can you rebase your PR in order to resolve the conflict? Thanks!

tlein:
Did the rebase. Turns out I had not yet added the BSD license like I had claimed, my mistake. Went ahead and added the license found in the toolchain repository to the top of iOS.toolchain.cmake

JonnyPtn:
I've tested this and it's great in comparison to the current method (Audio also worked, so what's outstanding for that to be included again?)

eXpl0it3r:
Merged in b6c1acab3ce4e6caf3cb0b2f7997b2f5f78b7f21

PR #1272, opened 2017-08-30
Fix iOS build error when using @import in ClipboardImpl.mm

### LINKED ISSUE #1271: iOS build fails because of import in ClipboardImpl.mm
When building master for iOS I get the following error: 

```

SFML/src/SFML/Window/iOS/ClipboardImpl.mm:30:1: error: use of '@import' when C++ modules are disabled, consider using
      fmodules and fcxx-modules
@import UIKit
^
1 error generated.
```

This can be fixed if `#import <UIKit/UIKit.h>` was used instead of the `@import UIKit` line.

tlein:
Resolves #1271 
**Description**
When compiling for iOS on master there is a build error in ClipboardImpl.mm (see accompanying issue), this fixes that issue. Seems modules get disabled when using Objective-C++ (see https://stackoverflow.com/questions/29701647/use-of-import-when-modules-are-disabled) even when the build settings enable modules.

**Testing**
Open a terminal in the SFML folder and run these commands:
`mkdir build && mkdir build/ios && mkdir build/ios-install`
`cd build/ios`
`cmake ../.. -DCMAKE_INSTALL_PREFIX=../ios-install/ -DIOS=True -G Xcode`
`cmake --build .`

SFML should compile without any errors.

### DIFF
--- a/src/SFML/Window/iOS/ClipboardImpl.mm
+++ b/src/SFML/Window/iOS/ClipboardImpl.mm
@@ -27,7 +27,7 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Window/iOS/ClipboardImpl.hpp>
 
-@import UIKit
+#import <UIKit/UIKit.h>
 
 namespace sf
 {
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
Merged in a2f448ab4cc29bffb9c91d7ee55c83f5adeaa2c8 👍

Issue #1278, opened 2017-09-08
Get rid of dependency to libjpeg


LaurentGomila:
The new version of stb_image_write (see #1270) now supports writing JPEG files. We should test it, and if it works fine then we could remove the dependency to libjpeg.

eXpl0it3r:
Superseded by #1279

PR #1279, opened 2017-09-08
Removed dependency to libjpeg

LaurentGomila:
Implements #1278.

This needs reviewing/testing regarding the stuff changed in (Android) makefiles and tools.


Foaly:
Would it also be worthwile to compare with other jpg libraries?
For example [libjpeg-turbo](https://github.com/libjpeg-turbo/libjpeg-turbo) or [mozjpeg](https://github.com/mozilla/mozjpeg) offer SIMD instructions for much faster processing as well as smaller file sizes while keeping the same image quality.
Or is this just about removing a dependency?

MarioLiebisch:
I think we should actually try to reduce the number of dependencies rather than just replacing them. Especially considering SFML is **not** an image processing library, so performance concerns should be secondary for things you do rather seldom.

If you really need that performance, e.g. to store an images every frame, people can just use an external library such as libjpeg-turbo.

Foaly:
Hmm I don't know. I agree perfomance is not a selling point as you pointed out. But I think the smaller file size on save whilst keeping the same quality is a point worth considering. At least for me.
Also isn't it kinda SFMLs philosophy that if a functionallity is offered in the library it is implemented in the best way possible and not if you want something better just do it yourself? 
Like I said I don't know if this PR is about improving functionallity or simply reducing the number of dependecies.

eXpl0it3r:
stb's image library has also various optimizations, as such one would have to first make a comparison which one does perform "better".

> a functionallity is offered in the library it is implemented in the best way possible

SFML doesn't guarantee the best possible implementation (however you measure "best") for everything. Sometimes we will sacrifice performance over maintainability or easy of use.

Having to deal with one dependency less, makes it easier for use to maintain and easier for anyone who wants to build all the dependencies themselves. libjpeg is ancient and the official source code's build files are broken (last I checked) - it's also the reason why I actually used libjpeg-turbo for the MSVC universal builds and why some package managers also use libjpeg-turbo as replacement.

File size optimization can go a long way, from different ways to compress it, to different ways to throw away data and use specific color palettes. SFML on the other hand just wants to be able to load and save image files. You can't say how the saving should be done, you can't even say what quality level you want the image to be saved. As such I don't see the need to go for a library that offers many configurations to make optimized files and then have SFML with no options to change any configurations. Plus whether "small file size" is really a criterion for "best possible way" is yet another argument to be had.

To get back to the beginning, without a clear performance test between the libraries, it's impossible to tell whether one is much better than the other and thus other things should be considered, like maintainability.

After all, if something else seems to trump stb in every way, there's nothing holding us back from switching to that lib again. :wink:

LaurentGomila:
> Like I said I don't know if this PR is about improving functionallity or simply reducing the number of dependecies.

stb_image_write can now save JPEG files, so I took this opportunity to remove the dependency to libjpeg. Nothing more. I don't think anybody cares about "improvements" to saving JPEG files in SFML -- as long as `image.saveToFile("test.jpg")` produces a valid JPEG file, everybody is happy.

JPEG has an obvious way of balancing between quality and size: the "quality" parameter. It has never been exposed by SFML (hard-coded to 90%), and so far nobody ever complained about it, so...

eXpl0it3r:
Tested this PR by loading, saving and displaying [these RGB 8bit test images](http://imagecompression.info/test_images/). Everything worked fine.

PR #1281, opened 2017-09-09
Fixed Wave file writer writing wrong header values

### LINKED ISSUE #1280: error in the SoundFileWriterWav :: close () method
Hi,
I found an error in the SoundFileWriterWav :: close () method. 
Uint32 dataChunkSize = static_cast<Uint32>(m_sampleCount * m_channelCount * 2);  <------ bug
This variable (m_sampleCount) should not be multiplied by m_channelCount.
m_sampleCount contains the number of samples saved rather than the number of samples per channel. When I write a stereo wav 44100 samplerate this wav header has an invalid length of chunks.

To fix this:
Uint32 dataChunkSize = static_cast<Uint32>(m_sampleCount * 2);  <------ bug fix

```c++
void SoundFileWriterWav::close()
{
  // If the file is open, finalize the header and close it
  if (m_file.is_open())
  {
    m_file.flush();

    //Update the main chunk size and data sub-chunk size
    //Uint32 dataChunkSize = static_cast<Uint32>(m_sampleCount * m_channelCount * 2); <------bug
    Uint32 dataChunkSize = static_cast<Uint32>(m_sampleCount * 2);  //<------ bug fix
    Uint32 mainChunkSize = dataChunkSize + 36;
    m_file.seekp(4);
    encode(m_file, mainChunkSize);
    m_file.seekp(40);
    encode(m_file, dataChunkSize);
    m_file.close();
  }
}

```

MarioLiebisch:
Fixes #1280.

Previously when updating the header fields, SFML assumed the number of samples written would be the number of samples per channel, which wasn't the case. Therefore for stereo files the written file length was actually twice the correct value. This fix uses the file size written as a basis, no longer counting the samples written altogether.

### DIFF
--- a/src/SFML/Audio/SoundFileWriterWav.cpp
+++ b/src/SFML/Audio/SoundFileWriterWav.cpp
@@ -86,9 +86,7 @@ bool SoundFileWriterWav::check(const std::string& filename)
 
 ////////////////////////////////////////////////////////////
 SoundFileWriterWav::SoundFileWriterWav() :
-m_file        (),
-m_sampleCount (0),
-m_channelCount(0)
+m_file()
 {
 }
 
@@ -118,9 +116,6 @@ bool SoundFileWriterWav::open(const std::string& filename, unsigned int sampleRa
         return false;
     }
 
-    // Save the channel count
-    m_channelCount = channelCount;
-
     return true;
 }
 
@@ -129,8 +124,6 @@ bool SoundFileWriterWav::open(const std::string& filename, unsigned int sampleRa
 void SoundFileWriterWav::write(const Int16* samples, Uint64 count)
 {
     assert(m_file.good());
-
-    m_sampleCount += count;
 
     while (count--)
         encode(m_file, *samples++);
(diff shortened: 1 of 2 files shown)


src/SFML/Audio/SoundFileWriterWav.cpp
```diff
@@ -191,8 +188,9 @@ void SoundFileWriterWav::close()
         m_file.flush();
 
         // Update the main chunk size and data sub-chunk size
-        Uint32 dataChunkSize = static_cast<Uint32>(m_sampleCount * m_channelCount * 2);
-        Uint32 mainChunkSize = dataChunkSize + 36;
+        Uint32 fileSize = m_file.tellp();
```
LaurentGomila: No compiler warning about type conversion here?

eXpl0it3r: Our VS CI builds have warnings for it.

```src\SFML\Audio\SoundFileWriterWav.cpp(191) : warning C4244: 'initializing' : conversion from 'std::streamoff' to 'sf::Uint32', possible loss of data```

tgnottingham:
Look good, but I think you can remove m_channelCount as well.

MarioLiebisch:
You're right, didn't notice!

Issue #1274, opened 2017-09-04
Graphics: Set anti-aliasing at RenderTexture


hydroper:
It should be possible to control the anti-aliasing like on `RenderWindow`. Smooth changed in nothing.

My reason: I simply need to cache a `Drawable`. On `RenderWindow` I'm able to explain the anti-aliasing amount with `ContextSettings`. [Is it possible to draw a Drawable and get a bitmap from it?](https://stackoverflow.com/q/46037260).

18kaioken:
@eXpl0it3r @LaurentGomila Yea, could you guys look this?

LaurentGomila:
I'd rather ask @binary1248 his opinion about it.

eXpl0it3r:
https://en.sfml-dev.org/forums/index.php?topic=22147.msg156438#msg156438

JonnyPtn:
I'll throw my hat in as someone who's wanted this for a while and would make use of it - I also know of others who would appreciate it.

If the answer is still "It's too much effort", then I'm happy to put effort in to work on an implementation

binary1248:
The TL;DR of my posts is: you don't have to supersample your intermediate surfaces if you make sure that aliasing can't happen when rendering to them. If you supersample both intermediate surfaces and final framebuffer you will end up losing more image quality than if you only did it in the final render pass.

This is like video editing: You don't compress your source video and edit the compressed clips into a final output. You work at the native resolution in some raw uncompressed format and only compress right before publishing it.

hydroper:
@binary1248

I did something like that with the green circle example. I drew the circle on a _render-texture_ without anti-aliasing settings, then did get the _render-texture_'s texture after drawn and drew it on a _render-window_. No anti-aliasing did happen at the end...

In order to enable anti-aliasing samples, someone suggested me to use _context-settings_ on a _context_, making sure the _render-texture_ is the active context. Though I couldn't compile my test code because the `Context` constructor is deleted. :/

naezith:
I'd also love to have this because my whole game is on a RenderTexture, I draw that to RenderWindow.

binary1248:
@naezith Are you applying some full screen post-processing or is there another reason that you need to do this?

naezith:
@binary1248 Yes, vignette effect shader.

binary1248:
Try #1285 branch [feature/rendertexture_stencil_multisample](https://github.com/SFML/SFML/tree/feature/rendertexture_stencil_multisample).

[...]

PR #1285, opened 2017-09-17
sf::RenderTexture stencil and multisampling support

binary1248:
**SFML Tasks**

* [x] Check reported finding
* [x] Review code
* [x] Test implementation

----

Added support for creation of a stencil attachment and multisampling to sf::RenderTexture. (#1274)

Test with this:
```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode(800, 600, 32), "Test", sf::Style::Titlebar | sf::Style::Close);

    sf::CircleShape circle(200, 9);
    circle.setFillColor(sf::Color::White);
    circle.setPosition(sf::Vector2f(200, 100));

    renderTexture.create(800, 600, sf::ContextSettings(0, 0, 8));

    sf::Sprite sprite;
    sprite.setTexture(renderTexture.getTexture());

    bool on = true;

    while (window.isOpen())
    {
        sf::Event event;
        {
            if ((event.type == sf::Event::Closed) || ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Escape)))
            {
                window.close();
                break;
            }

            if ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Space))
            {
                on = !on;
                renderTexture.create(800, 600, sf::ContextSettings(0, 0, on ? 8 : 0));
            }
        }

        renderTexture.clear(sf::Color::Black);
        renderTexture.draw(circle);
        renderTexture.display();

        window.clear();
        window.draw(sprite);
        window.display();
    }
}
```

### DIFF
--- a/src/SFML/Graphics/RenderTextureImplFBO.cpp
+++ b/src/SFML/Graphics/RenderTextureImplFBO.cpp
@@ -37,9 +37,13 @@ namespace priv
 {
 ////////////////////////////////////////////////////////////
 RenderTextureImplFBO::RenderTextureImplFBO() :
-m_context    (NULL),
-m_frameBuffer(0),
-m_depthBuffer(0)
+m_context               (NULL),
+m_frameBuffer           (0),
+m_multisampleFrameBuffer(0),
+m_depthStencilBuffer    (0),
+m_colorBuffer           (0),
+m_width                 (0),
+m_height                (0)
 {
 
 }
@@ -50,11 +54,25 @@ RenderTextureImplFBO::~RenderTextureImplFBO()
 {
     m_context->setActive(true);
 
-    // Destroy the depth buffer
-    if (m_depthBuffer)
+    // Destroy the color buffer
+    if (m_colorBuffer)
     {
-        GLuint depthBuffer = static_cast<GLuint>(m_depthBuffer);
-        glCheck(GLEXT_glDeleteRenderbuffers(1, &depthBuffer));
+        GLuint colorBuffer = static_cast<GLuint>(m_colorBuffer);
+        glCheck(GLEXT_glDeleteRenderbuffers(1, &colorBuffer));
+    }
+
+    // Destroy the depth/stencil buffer
+    if (m_depthStencilBuffer)
+    {
+        GLuint depthStencilBuffer = static_cast<GLuint>(m_depthStencilBuffer);
+        glCheck(GLEXT_glDeleteRenderbuffers(1, &depthStencilBuffer));
+    }
+
+    // Destroy the multisample frame buffer
+    if (m_multisampleFrameBuffer)
+    {
+        GLuint multisampleFrameBuffer = static_cast<GLuint>(m_multisampleFrameBuffer);
+        glCheck(GLEXT_glDeleteFramebuffers(1, &multisampleFrameBuffer));
     }
 
     // Destroy the frame buffer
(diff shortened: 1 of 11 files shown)


hydroper:
Thanks very much

hydroper:
@binary1248 I wish SFML was already pre-compiled with these changes.

eXpl0it3r:
See the [CI artifact builds](https://www.sfml-dev.org/artifacts/by-branch/feature/rendertexture_stencil_multisample/)

hydroper:
@eXpl0it3r Thanks. Unfortunately I see nothing. I compiled the PR example, but nothing looks to render. It's all black... I'm using MingW 32..

hydroper:
@binary1248 Thanks very much for this. It didn't work here :/. It's entirely black and when I press any key, nothing happens too.

src/SFML/Graphics/RenderTextureImplFBO.cpp
```diff
@@ -141,6 +321,18 @@ bool RenderTextureImplFBO::activate(bool active)
 ////////////////////////////////////////////////////////////
 void RenderTextureImplFBO::updateTexture(unsigned int)
 {
+
+#ifndef SFML_OPENGL_ES
+
+    if (m_multisampleFrameBuffer)
+    {
+        glCheck(GLEXT_glBindFramebuffer(GLEXT_GL_DRAW_FRAMEBUFFER, m_frameBuffer));
+        glCheck(GLEXT_glBlitFramebuffer(0, 0, m_width, m_height, 0, 0, m_width, m_height, GL_COLOR_BUFFER_BIT, GL_NEAREST));
```
Foaly: Wouldn't GL_LINEAR give smoother results? Or is the parameter irrelevant, because the textures have the same size?

binary1248: In the odd case the driver is stupid, this is faster and imposes no loss of quality.

Foaly:
Hi there!
I just tested this PR and it works beautifully for me. Thanks for the work!
One thing I noticed is that when I tried to set the AA level to 16 the example stopped working for me. I checked the return code and saw that the create function failed. Looking into the code I saw that the highest AA level on my computer is 8. I think there should be a notice in sf::err when the creation failed, because of an unsupported AA level, since AA will be the most used case IMO.
I suspect this is also why the example didn't work for @hydroper.

src/SFML/Graphics/RenderTextureImplFBO.cpp
```diff
@@ -82,52 +100,214 @@ bool RenderTextureImplFBO::isAvailable()
...
+
+#ifndef SFML_OPENGL_ES
+
+    // Check if the requested anti-aliasing level is supported
+    if (settings.antialiasingLevel)
+    {
+        GLint samples = 0;
+        glGetIntegerv(GLEXT_GL_MAX_SAMPLES, &samples);
+
+        if(settings.antialiasingLevel > samples)
```
Foaly: I would suggest something like this here: `err() << "Impossible to create render texture (unsupported anti-aliasing level, maximum supported samples: " << samples << " )" << std::endl;`

eXpl0it3r:
It's a good point. For the RenderWindow we just fallback to the minimal setting if the config doesn't match and return the difference on `sf::err()`. Would this be possible here without too much effort or at least to produce an error so it's clear why something failed.

binary1248:
I'd rather not spam the console with warnings that can be avoided with the necessary precaution. The problem with RenderWindow and this in its current state is that there is no way to check the maximum AA that is supported. It would be simple to do in both cases, but would necessitate adding static functions to both classes.

Foaly:
Well I don't know I wouldn't consider it spam, since it informs the user why the creation of the RenderTexture failed. Also pretty much every other error case in `create` prints the reason to the console. But I agree its a very good idea to also add a function to get the maximum supported AA level.

eXpl0it3r:
So you don't want to add any message @binary1248? Can you rebase it as well?

binary1248:
Updated.

PR #1287, opened 2017-09-18
Android build support for newest NDK release

AlexAUT:
Currently we ship our own android.toolchain file for cmake, which does not work anymore with the newest NDK releases (> r12). When compiling SFML with the llvm STL implementation (c++_static/shared), you will get undefined symbols errors. OpenCV has the same problems and could not fix it yet.

But the NDK comes now with his own Cmake toolchain (NDK_ROOT/build/cmake), and this PR adds support for this shipped toolchain and should not break the old one. It seems that there are some relative paths in the toolchain file, so we cannot simply copy the toolchain file into the SFML repo, but since you need the NDK anyways to build SFML for android this should not be a problem.


MarioLiebisch:
Cool. Put on my To-Do. :)

eXpl0it3r:

eXpl0it3r:

MarioLiebisch:
Argh, I thought it would actually fast forward the changes. :( Reverting back to the old plan and providing this commit in a new PR. Thanks for the fixes though. :)

PR #1290, opened 2017-09-23
Caps lock detection

pixelherodev:
Add the ability to check if caps lock is on to Windows and Linux. Mac always returns false (unable to find documentation for the check on OS X), and mobile implementations don't seem to have a isKeyPressed function.

### DIFF
--- a/include/SFML/Window/Keyboard.hpp
+++ b/include/SFML/Window/Keyboard.hpp
@@ -148,7 +148,8 @@ public:
         F13,          ///< The F13 key
         F14,          ///< The F14 key
         F15,          ///< The F15 key
-        Pause,        ///< The Pause key
+        Pause,        ///< The Pause key,
+        CAPS,          ///< Whether or not caps lock is on
 
         KeyCount      ///< Keep last -- the total number of keyboard keys
     };
--- a/src/SFML/Window/OSX/HIDInputManager.mm
+++ b/src/SFML/Window/OSX/HIDInputManager.mm
@@ -45,6 +45,7 @@ HIDInputManager& HIDInputManager::getInstance()
 ////////////////////////////////////////////////////////////
 bool HIDInputManager::isKeyPressed(Keyboard::Key key)
 {
+	if(key==Keyboard::CAPS)return false;
     return isPressed(m_keys[key]);
 }
 
--- a/src/SFML/Window/Unix/InputImpl.cpp
+++ b/src/SFML/Window/Unix/InputImpl.cpp
@@ -30,8 +30,8 @@
 #include <SFML/Window/Unix/Display.hpp>
 #include <SFML/System/Err.hpp>
 #include <X11/Xlib.h>
+#include <X11/XKBlib.h>
 #include <X11/keysym.h>
-
 
 namespace sf
 {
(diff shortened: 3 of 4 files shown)


include/SFML/Window/Keyboard.hpp
```diff
@@ -148,7 +148,8 @@ class SFML_WINDOW_API Keyboard
         F13,          ///< The F13 key
         F14,          ///< The F14 key
         F15,          ///< The F15 key
-        Pause,        ///< The Pause key
+        Pause,        ///< The Pause key,
+        CAPS,          ///< Whether or not caps lock is on
```
MarioLiebisch: I'd say the enum member should use camel case as usual: `CapsLock`

src/SFML/Window/OSX/HIDInputManager.mm
```diff
@@ -45,6 +45,7 @@
 ////////////////////////////////////////////////////////////
 bool HIDInputManager::isKeyPressed(Keyboard::Key key)
 {
+	if(key==Keyboard::CAPS)return false;
```
MarioLiebisch: This should be handled gracefully by `isPressed()`, e.g. by setting a special/undefined index for `m_keys[key]`. (Didn't check actual values/usage.)

src/SFML/Window/Unix/InputImpl.cpp
```diff
@@ -41,7 +41,16 @@ namespace priv
...
 {
     // Get the corresponding X11 keysym
-    KeySym keysym = 0;
+
+    if(key==Keyboard::CAPS){
+		Display* display = OpenDisplay();
+		unsigned int n;
+        XkbGetIndicatorState(display, XkbUseCoreKbd, &n);
+        CloseDisplay(display);
+        return (n&0x01)==1;
```
MarioLiebisch: Really weid looking magic numbers here.

src/SFML/Window/Unix/InputImpl.cpp
```diff
@@ -41,7 +41,16 @@ namespace priv
...
+	}
+
+    KeySym keysym;
```
MarioLiebisch: This results in an undefined `KeySym` for unhandled cases. Why not leave the default value of `0`?

src/SFML/Window/Win32/InputImpl.cpp
```diff
@@ -45,7 +45,7 @@ namespace priv
 ////////////////////////////////////////////////////////////
 bool InputImpl::isKeyPressed(Keyboard::Key key)
 {
-    int vkey = 0;
+    int vkey;
```
MarioLiebisch: Same as above, why remove the default case, which will cause undefined behavior?

src/SFML/Window/Win32/InputImpl.cpp
```diff
@@ -150,6 +150,7 @@ bool InputImpl::isKeyPressed(Keyboard::Key key)
         case Keyboard::F14:        vkey = VK_F14;        break;
         case Keyboard::F15:        vkey = VK_F15;        break;
         case Keyboard::Pause:      vkey = VK_PAUSE;      break;
+        case Keyboard::CAPS:       return (GetKeyState(VK_CAPITAL) & 1) !=0;
```
MarioLiebisch: Is there any reason for this async (pun intended) call to a different function?

mantognini:
What's the rational for this? It was discussed before at least on the forum -- in a nutshell, CapsLock isn't useful for the current event system: it's a modifier, not really a key. You most likely might prefer using `TextEntered` event instead. The scancode API will probably deal with this anyway. The procedure to open feature requests is explained in [the contribution guidelines](http://www.sfml-dev.org/contribute.php).

pixelherodev:
Sorry, didn't notice that.

pixelherodev:
Also, in retrospect this *was* a very localized need.

PR #1230, opened 2017-05-08
Improved method used to filter KeyRelease in X11

texus:
This should fix #1223.

The event behind KeyRelease is now only retrieved from the same window. Because peeking did not work well, the event is instead removed from the queue and put back afterwards (unless the event is KeyPress and key repeating is disabled).

PR #1291, opened 2017-09-25
Fix X11 key repeat handling not filtering out events from other windows

### LINKED ISSUE #1223: setKeyRepeatEnabled(false) doesn't work after second window.create() call on Linux
setKeyRepeatEnabled(false) doesn't work after second window.create() call on Linux.

Here is the minimal code:

`
#include <iostream>
#include <SFML/Graphics.hpp>

sf::RenderWindow window;

void createWindow() {
    window.create(sf::VideoMode(100, 100), "Window");
    window.setKeyRepeatEnabled(false);
}

int main (int argc, char* argv[]) {
    createWindow();

    while (window.isOpen()) {
        sf::Event event;
            switch(event.type) {
                case sf::Event::Closed:
                    window.close();
                    break;

                case sf::Event::KeyPressed:
                    if(event.key.code == sf::Keyboard::H) createWindow();

                    std::cout << "Key Pressed: " << event.key.code << std::endl;
                    break;

                case sf::Event::KeyReleased:
                    std::cout << "Key Released: " << event.key.code << std::endl;
                    break;
            }
        }
    }

    return 0;
}
`

binary1248:
Fixes #1223.
Closes #1230.

### DIFF
--- a/src/SFML/Window/Unix/WindowImplX11.cpp
+++ b/src/SFML/Window/Unix/WindowImplX11.cpp
@@ -77,6 +77,21 @@ namespace
 
     static const unsigned int             maxTrialsCount = 5;
 
+    // Predicate we use to find key repeat events in processEvent
+    struct KeyRepeatFinder
+    {
+        KeyRepeatFinder(unsigned int keycode, Time time) : keycode(keycode), time(time) {}
+
+        // Predicate operator that checks event type, keycode and timestamp
+        bool operator()(const XEvent& event)
+        {
+            return ((event.type == KeyPress) && (event.xkey.keycode == keycode) && (event.xkey.time - time < 2));
+        }
+
+        unsigned int keycode;
+        Time time;
+    };
+
     // Filter the events received by windows (only allow those matching a specific window)
     Bool checkEvent(::Display*, XEvent* event, XPointer userData)
     {
@@ -752,8 +767,16 @@ WindowHandle WindowImplX11::getSystemHandle() const
 void WindowImplX11::processEvents()
 {
     XEvent event;
+
+    // Pick out the events that are interesting for this window
     while (XCheckIfEvent(m_display, &event, &checkEvent, reinterpret_cast<XPointer>(m_window)))
+        m_events.push_back(event);
+
+    // Handle the events for this window that we just picked out
+    while (!m_events.empty())
     {
+        event = m_events.front();
+        m_events.pop_front();
         processEvent(event);
     }
 }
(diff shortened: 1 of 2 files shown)


eXpl0it3r:
@naezith @texus Can you give this a spin?

texus:
This solves the issues on both my linux computers. 👍

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1533,29 +1556,23 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
...
 
     // Detect repeated key events
-    // (code shamelessly taken from SDL)
     if (windowEvent.type == KeyRelease)
     {
-        // Check if there's a matching KeyPress event in the queue
-        XEvent nextEvent;
-        if (XPending(m_display))
+        // Find the next KeyPress event with matching keycode and time
+        std::deque<XEvent>::iterator iter = std::find_if(
```
LaurentGomila: Is there a reason to call `find_if` if `m_keyRepeat` is false?

eXpl0it3r:
Tested on Debian x64 Gnome. Was able to reproduce the issue and this PR fixed it.

src/SFML/Window/Unix/WindowImplX11.cpp
eXpl0it3r: I do think so, as it has to fulfill the two tasks mentioned at the function top. 

* Discard duplicated KeyRelease events when KeyRepeatEnabled is true
* Discard both duplicated KeyPress and KeyRelease events when KeyRepeatEnabled is false

When repeat is true it will now ignore the repeated KeyRelease event that X generates.

When repeat is false it will erase the repeated KeyPress event from the queue and ignore the repeated KeyRelease event.

LaurentGomila: Sorry, I didn't read the code carefully 😸

naezith:
Sorry for testing late, I just did. Absolutely nothing changed for me. #1223 is still there. @eXpl0it3r @binary1248 

Linux (Ubuntu 16.04)

eXpl0it3r:
Having tested and reviewed the code myself, I don't trust your result. Did you sure that the correct library is linked and not an old version.

If the new code is actually running, then the only reason it fails must be because the timestamps reported by X don't match, which I can't think is happening as three independent testers report it to work fine.

naezith:
@eXpl0it3r I manually went into Commits of this, did "Browse the repository at this point in the history", then downloaded .zip of whole repository from that page. Checked if the changes are there. Renamed my original SFML and build folders. Put this new one there and changed the name to SFML, generated files with cmake, then built it with "make" into the build folder. Then run this script I wrote which copies the libs to the game path:

```sh
#!/bin/bash

game_path="/home/naezith/.steam/steam/steamapps/common/Remnants of Naezith Beta"
linux_content="/media/naezith/SSD/steamworks_sdk/tools/ContentBuilder/content/linux_content"
TARGETS=("$linux_content/lib/" "$game_path/lib/") 

for TARGET_PATH in "${TARGETS[@]}"; do
	echo " "
	echo ">> NEW TARGET PATH:""$TARGET_PATH"

	# Copy SFML
	SFML_PATH=/home/naezith/sfml-build/lib/
	SFML_LIBS=(audio graphics window network system)

	for f in "${SFML_LIBS[@]}"; do
		path=$SFML_PATH"libsfml-"$f".so.2.4"
		echo $path
		cp $path "$TARGET_PATH"
	done 
done
```

Then I opened the SFML project and compiled it, and tested. Is there any suspicious part in this? Maybe I am doing this testing thing wrong.

naezith:
My bad, there was SFML installed directly to /usr/local/lib kind of folders. They were prioritized I guess. I cleaned them now. I did the same stuff as I explained and now it works! Thanks a lot.

Issue #1295, opened 2017-09-29
Support for extensible format PCM wave files


tgnottingham:
Since removing the libsndfile dependency, SFML only supports integer-encoded PCM wave files with the PCM format tag. However, there are identically encoded wave files that use the "extensible" tag instead. Since some applications use this extensible tag when saving PCM waves, it would be nice to support the format.

With an extensible format file, the example below outputs "Failed to open WAV sound file (invalid or unsupported file)" (assuming the file is named "tone_24.wav" and it's in your working directory).

```
#include <SFML/Audio.hpp>
#include <SFML/System.hpp>

int main() {
    sf::SoundBuffer sb; 
    sb.loadFromFile("tone_24.wav");

    sf::Sound s(sb);
    s.play();

    sf::sleep(sf::seconds(2.5));
}
```

I'll post a pull request shortly to help you see if you want to include support.

[Forum Thread](https://en.sfml-dev.org/forums/index.php?topic=22559.0)
[Wave File Specifications](http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/WAVE.html)
[Example of Unsupported File](https://www.dropbox.com/s/4j48bvpa35daowp/tone_24.wav?dl=1)

eXpl0it3r:
Merged in fc54dba3d7ff15d99d9e86fea7d5cb593dd86106

PR #1296, opened 2017-09-29
Added support for extensible format PCM wave files.

tgnottingham:
Implementation for issue #1295. See issue for motivation, code example, and links to relevant info.

This adds support for extensible format PCM wave files by allowing the extensible format tag, parsing additional header data if that tag is present, and rejecting extensible format files if their subformat is not PCM.

### DIFF
--- a/src/SFML/Audio/SoundFileReaderWav.cpp
+++ b/src/SFML/Audio/SoundFileReaderWav.cpp
@@ -31,6 +31,7 @@
 #include <algorithm>
 #include <cctype>
 #include <cassert>
+#include <cstring>
 
 
 namespace
@@ -88,6 +89,14 @@ namespace
     }
 
     const sf::Uint64 mainChunkSize = 12;
+
+    const sf::Uint16 waveFormatPcm = 1;
+
+    const sf::Uint16 waveFormatExtensible= 65534;
+
+    const char* waveSubformatPcm =
+        "\x01\x00\x00\x00\x00\x00\x10\x00"
+        "\x80\x00\x00\xAA\x00\x38\x9B\x71";
 }
 
 namespace sf
@@ -226,6 +235,9 @@ bool SoundFileReaderWav::parseHeader(Info& info)
         Uint32 subChunkSize = 0;
         if (!decode(*m_stream, subChunkSize))
             return false;
+        Int64 subChunkStart = m_stream->tell();
+        if (subChunkStart == -1)
+            return false;
 
         // Check which chunk it is
         if ((subChunkId[0] == 'f') && (subChunkId[1] == 'm') && (subChunkId[2] == 't') && (subChunkId[3] == ' '))
@@ -236,7 +248,7 @@ bool SoundFileReaderWav::parseHeader(Info& info)
             Uint16 format = 0;
             if (!decode(*m_stream, format))
                 return false;
-            if (format != 1) // PCM
+            if ((format != waveFormatPcm) && (format != waveFormatExtensible))
                 return false;
 
             // Channel count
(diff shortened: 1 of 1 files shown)


src/SFML/Audio/SoundFileReaderWav.cpp
```diff
@@ -88,6 +89,16 @@ namespace
     }
 
     const sf::Uint64 mainChunkSize = 12;
+
+    enum WaveFormat
+    {
+        WaveFormatPcm = 1,
+        WaveFormatExtensible = 65534,
```
mantognini: Extra comma is C++11 feature.

src/SFML/Audio/SoundFileReaderWav.cpp
```diff
@@ -272,12 +286,54 @@ bool SoundFileReaderWav::parseHeader(Info& info)
...
+
+                // Channel mask
+                Uint32 channelMask = 0;
+                if (!decode(*m_stream, channelMask))
+                    return false;
+
+                // Subformat
+                char subformat[16];
+                if (m_stream->read(subformat, sizeof(subformat)) !=
+                        sizeof(subformat))
```
mantognini: Could be on one line.

src/SFML/Audio/SoundFileReaderWav.cpp
```diff
@@ -272,12 +286,54 @@ bool SoundFileReaderWav::parseHeader(Info& info)
...
+                        " (" << bitsPerSample << " bits) differ" << std::endl;
+                    return false;
+                }
             }
+
+            // Skip potential extra information
+            Int64 subChunkEnd = subChunkStart + subChunkSize;
+
+            Int64 position = m_stream->tell();
+            if (position == -1 || position > subChunkEnd)
```
mantognini: Please add parentheses for each operand of the `||` operator here.

mantognini:
Arrg, wrong button. It's approve modulo changes, obviously...

mantognini:

src/SFML/Audio/SoundFileReaderWav.cpp
tgnottingham: Do you guys have an established max line length?

MarioLiebisch: Readability over strict max line length IMO.

src/SFML/Audio/SoundFileReaderWav.cpp
MarioLiebisch: Pretty sure that's C++98/C99. I definitely used it before C++11.

src/SFML/Audio/SoundFileReaderWav.cpp
mantognini: Mario is right. But we actually don't have such rule in our [style guide](https://www.sfml-dev.org/style.php).

src/SFML/Audio/SoundFileReaderWav.cpp
mantognini: from clang:

```
warning: commas at the end of enumerator lists are a C++11 extension [-Wc++11-extensions]
        WaveFormatExtensible = 65534,
```

;-)

tgnottingham:
Thanks, I think I've addressed the remaining style issues with an amended commit.

src/SFML/Audio/SoundFileReaderWav.cpp
```diff
@@ -88,6 +89,16 @@ namespace
     }
 
     const sf::Uint64 mainChunkSize = 12;
+
+    enum WaveFormat
```
LaurentGomila: I think you should use const integers, to remain consistent with what's already done. The enum name is not even used, all we want is to name those magical numbers.

src/SFML/Audio/SoundFileReaderWav.cpp
```diff
@@ -272,12 +286,51 @@ bool SoundFileReaderWav::parseHeader(Info& info)
...
+                Uint32 channelMask = 0;
+                if (!decode(*m_stream, channelMask))
+                    return false;
+
+                // Subformat
+                char subformat[16];
+                if (m_stream->read(subformat, sizeof(subformat)) != sizeof(subformat))
                     return false;
+
+                if (memcmp(subformat, waveSubformatPcm, sizeof(subformat)) != 0)
```
LaurentGomila: `std::memcmp`

src/SFML/Audio/SoundFileReaderWav.cpp
```diff
@@ -272,12 +286,51 @@ bool SoundFileReaderWav::parseHeader(Info& info)
...
+                            "sample container size (" << bitsPerSample << " bits) differ" << std::endl;
+                    return false;
+                }
             }
+
+            // Skip potential extra information
+            Int64 subChunkEnd = subChunkStart + subChunkSize;
+
+            Int64 position = m_stream->tell();
+            if ((position == -1) || (position > subChunkEnd))
```
LaurentGomila: Are these checks really needed? I'd say just jump to the end of the current chunk and go on (that's what was done before, anyway)

tgnottingham: It would only serve to catch malformed headers. Since the rest of the code doesn't do much of this kind of validation, I'm okay with dropping it.

eXpl0it3r:

eXpl0it3r:

PR #1298, opened 2017-10-01
Enable comparing sf::Transform and optimize resetting OpenGL back to the identity matrix

binary1248:
In cases where the RenderTarget is forced to go in and out of "cache mode", it will end up constantly uploading the matrix of Transform::Identity to OpenGL even though this is most likely a built in constant (and probably highly optimized). This patch makes use of the built-in and reduces CPU/RAM load in cases where the copying is happening excessively.

### DIFF
--- a/src/SFML/Graphics/Transform.cpp
+++ b/src/SFML/Graphics/Transform.cpp
@@ -269,4 +269,23 @@ Vector2f operator *(const Transform& left, const Vector2f& right)
     return left.transformPoint(right);
 }
 
+
+////////////////////////////////////////////////////////////
+bool operator ==(const Transform& left, const Transform& right)
+{
+    const float* a = left.getMatrix();
+    const float* b = right.getMatrix();
+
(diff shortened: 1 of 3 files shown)


binary1248:
Fixed in new version.

> And is it so obvious that using glLoadIdentity + the added float comparisons is more optimized than glLoadMatrix? (just wondering)

The values of the matrix will very likely be in the L1 cache since they were passed into draw() and applyTransform() by reference. Float comparisons are super fast, much faster than copying memory from one address space into another and then over the PCIe interface. Who knows, maybe the driver compares our data with the identity matrix and optimizes this itself, but if we do it ourselves we are on the safer side.

src/SFML/Graphics/Transform.cpp
```diff
@@ -26,6 +26,7 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Graphics/Transform.hpp>
+#include <algorithm>
```
LaurentGomila: This one can be removed now.

binary1248:
Fixed.

PR #1308, opened 2017-10-21
Added sf::VertexBuffer class

binary1248:
**SFML Tasks**

* [x] Review code (2 reviews)
* [x] Test implementation
* [x] Resolve conflict

----

Also added an example demonstrating VertexBuffer, Shader and Thread usage.

If anybody feels like tweaking the generation parameters a bit to make the example output look better feel free to do so. Being an artist is not one of my strong points...

### DIFF
--- a/include/SFML/Graphics/VertexBuffer.hpp
+++ b/include/SFML/Graphics/VertexBuffer.hpp
@@ -1,0 +1,408 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2017 Laurent Gomila (laurent@sfml-dev.org)
+//
+// This software is provided 'as-is', without any express or implied warranty.
+// In no event will the authors be held liable for any damages arising from the use of this software.
+//
+// Permission is granted to anyone to use this software for any purpose,
+// including commercial applications, and to alter it and redistribute it freely,
+// subject to the following restrictions:
+//
+// 1. The origin of this software must not be misrepresented;
+//    you must not claim that you wrote the original software.
+//    If you use this software in a product, an acknowledgment
+//    in the product documentation would be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such,
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+#ifndef SFML_VERTEXBUFFER_HPP
+#define SFML_VERTEXBUFFER_HPP
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#include <SFML/Graphics/Export.hpp>
+#include <SFML/Graphics/PrimitiveType.hpp>
+#include <SFML/Graphics/Drawable.hpp>
+#include <SFML/Window/GlResource.hpp>
+
+
+namespace sf
+{
+class RenderTarget;
+class Vertex;
+
+////////////////////////////////////////////////////////////
+/// \brief Vertex buffer storage for one or more 2D primitives
+///
+////////////////////////////////////////////////////////////
+class SFML_GRAPHICS_API VertexBuffer : public Drawable, private GlResource
+{
+public:
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Usage specifiers
+    ///
+    /// If data is going to be updated once or more every frame,
+    /// set the usage to Stream. If data is going to be set once
+    /// and used for a long time without being modified, set the
+    /// usage to Static. For everything else Dynamic should be a
+    /// good compromise.
+    ///
+    ////////////////////////////////////////////////////////////
+    enum Usage
+    {
+        Stream,  ///< Constantly changing data
+        Dynamic, ///< Occasionally changing data
+        Static   ///< Rarely changing data
+    };
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Default constructor
+    ///
+    /// Creates an empty vertex buffer.
+    ///
+    ////////////////////////////////////////////////////////////
+    VertexBuffer();
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Construct a VertexBuffer with a specific PrimitiveType
+    ///
+    /// Creates an empty vertex buffer and sets its primitive type to \p type.
+    ///
+    /// \param type Type of primitive
+    ///
+    ////////////////////////////////////////////////////////////
+    explicit VertexBuffer(PrimitiveType type);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Construct a VertexBuffer with a specific usage specifier
+    ///
+    /// Creates an empty vertex buffer and sets its usage to \p usage.
+    ///
+    /// \param usage Usage specifier
+    ///
+    ////////////////////////////////////////////////////////////
+    explicit VertexBuffer(Usage usage);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Construct a VertexBuffer with a specific PrimitiveType and usage specifier
+    ///
+    /// Creates an empty vertex buffer and sets its primitive type
+    /// to \p type and usage to \p usage.
+    ///
+    /// \param type  Type of primitive
+    /// \param usage Usage specifier
+    ///
+    ////////////////////////////////////////////////////////////
+    VertexBuffer(PrimitiveType type, Usage usage);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Copy constructor
+    ///
+    /// \param copy instance to copy
+    ///
+    ////////////////////////////////////////////////////////////
+    VertexBuffer(const VertexBuffer& copy);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Destructor
+    ///
+    ////////////////////////////////////////////////////////////
+    ~VertexBuffer();
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Create the vertex buffer
+    ///
+    /// Creates the vertex buffer and allocates enough graphics
+    /// memory to hold \p vertexCount vertices. Any previously
+    /// allocated memory is freed in the process.
+    ///
+    /// In order to deallocate previously allocated memory pass 0
+    /// as \p vertexCount. Don't forget to recreate with a non-zero
+    /// value when graphics memory should be allocated again.
+    ///
+    /// \param vertexCount Number of vertices worth of memory to allocate
+    ///
+    /// \return True if creation was successful
+    ///
+    ////////////////////////////////////////////////////////////
+    bool create(std::size_t vertexCount);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Return the vertex count
+    ///
+    /// \return Number of vertices in the vertex buffer
+    ///
+    ////////////////////////////////////////////////////////////
+    std::size_t getVertexCount() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Update the whole buffer from an array of vertices
+    ///
+    /// The \a vertex array is assumed to have the same size as
+    /// the \a created buffer.
+    ///
+    /// No additional check is performed on the size of the vertex
+    /// array, passing invalid arguments will lead to undefined
+    /// behavior.
+    ///
+    /// This function does nothing if \a vertices is null or if the
+    /// buffer was not previously created.
+    ///
+    /// \param vertices Array of vertices to copy to the buffer
+    ///
+    /// \return True if the update was successful
+    ///
+    ////////////////////////////////////////////////////////////
+    bool update(const Vertex* vertices);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Update a part of the buffer from an array of vertices
+    ///
+    /// \p offset is specified as the number of vertices to skip
+    /// from the beginning of the buffer.
+    ///
+    /// If \p offset is 0 and \p vertexCount is equal to the size of
+    /// the currently created buffer, its whole contents are replaced.
+    ///
+    /// If \p offset is 0 and \p vertexCount is greater than the
+    /// size of the currently created buffer, a new buffer is created
+    /// containing the vertex data.
+    ///
+    /// If \p offset is 0 and \p vertexCount is less than the size of
+    /// the currently created buffer, only the corresponding region
+    /// is updated.
+    ///
+    /// If \p offset is not 0 and \p offset + \p vertexCount is greater
+    /// than the size of the currently created buffer, the update fails.
+    ///
+    /// No additional check is performed on the size of the vertex
+    /// array, passing invalid arguments will lead to undefined
+    /// behavior.
+    ///
+    /// \param vertices    Array of vertices to copy to the buffer
+    /// \param vertexCount Number of vertices to copy
+    /// \param offset      Offset in the buffer to copy to
+    ///
+    /// \return True if the update was successful
+    ///
+    ////////////////////////////////////////////////////////////
+    bool update(const Vertex* vertices, std::size_t vertexCount, unsigned int offset);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Copy the contents of another buffer into this buffer
+    ///
+    /// \param vertexBuffer Vertex buffer whose contents to copy into this vertex buffer
+    ///
+    /// \return True if the copy was successful
+    ///
+    ////////////////////////////////////////////////////////////
+    bool update(const VertexBuffer& vertexBuffer);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Overload of assignment operator
+    ///
+    /// \param right Instance to assign
+    ///
+    /// \return Reference to self
+    ///
+    ////////////////////////////////////////////////////////////
+    VertexBuffer& operator =(const VertexBuffer& right);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Swap the contents of this vertex buffer with those of another
+    ///
+    /// \param right Instance to swap with
+    ///
+    ////////////////////////////////////////////////////////////
+    void swap(VertexBuffer& right);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the underlying OpenGL handle of the vertex buffer.
+    ///
+    /// You shouldn't need to use this function, unless you have
+    /// very specific stuff to implement that SFML doesn't support,
+    /// or implement a temporary workaround until a bug is fixed.
+    ///
+    /// \return OpenGL handle of the vertex buffer or 0 if not yet created
+    ///
+    ////////////////////////////////////////////////////////////
+    unsigned int getNativeHandle() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Set the type of primitives to draw
+    ///
+    /// This function defines how the vertices must be interpreted
+    /// when it's time to draw them.
+    ///
+    /// The default primitive type is sf::Points.
+    ///
+    /// \param type Type of primitive
+    ///
+    ////////////////////////////////////////////////////////////
+    void setPrimitiveType(PrimitiveType type);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the type of primitives drawn by the vertex buffer
+    ///
+    /// \return Primitive type
+    ///
+    ////////////////////////////////////////////////////////////
+    PrimitiveType getPrimitiveType() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Set the usage specifier of this vertex buffer
+    ///
+    /// This function provides a hint about how this vertex buffer is
+    /// going to be used in terms of data update frequency.
+    ///
+    /// After changing the usage specifier, the vertex buffer has
+    /// to be updated with new data for the usage specifier to
+    /// take effect.
+    ///
+    /// The default primitive type is sf::VertexBuffer::Stream.
+    ///
+    /// \param usage Usage specifier
+    ///
+    ////////////////////////////////////////////////////////////
+    void setUsage(Usage usage);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the usage specifier of this vertex buffer
+    ///
+    /// \return Usage specifier
+    ///
+    ////////////////////////////////////////////////////////////
+    Usage getUsage() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Bind a vertex buffer for rendering
+    ///
+    /// This function is not part of the graphics API, it mustn't be
+    /// used when drawing SFML entities. It must be used only if you
+    /// mix sf::VertexBuffer with OpenGL code.
+    ///
+    /// \code
+    /// sf::VertexBuffer vb1, vb2;
+    /// ...
+    /// sf::VertexBuffer::bind(&vb1);
+    /// // draw OpenGL stuff that use vb1...
+    /// sf::VertexBuffer::bind(&vb2);
+    /// // draw OpenGL stuff that use vb2...
+    /// sf::VertexBuffer::bind(NULL);
+    /// // draw OpenGL stuff that use no vertex buffer...
+    /// \endcode
+    ///
+    /// \param vertexBuffer Pointer to the vertex buffer to bind, can be null to use no vertex buffer
+    ///
+    ////////////////////////////////////////////////////////////
+    static void bind(const VertexBuffer* vertexBuffer);
+
+    ////////////////////////////////////////////////////////////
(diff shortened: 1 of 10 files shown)


include/SFML/Graphics/VertexBuffer.hpp
```diff
@@ -0,0 +1,389 @@
...
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Construct a VertexBuffer with a specific PrimitiveType
+    ///
+    /// Creates an empty vertex buffer and sets its primitive type to \p type.
+    ///
+    /// \param type Type of primitive
+    ///
+    ////////////////////////////////////////////////////////////
+    VertexBuffer(PrimitiveType type);
```
LaurentGomila: This constructor should be marked 'explicit'.

include/SFML/Graphics/VertexBuffer.hpp
```diff
@@ -0,0 +1,389 @@
...
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Construct a VertexBuffer with a specific usage specifier
+    ///
+    /// Creates an empty vertex buffer and sets its usage to \p usage.
+    ///
+    /// \param usage Usage specifier
+    ///
+    ////////////////////////////////////////////////////////////
+    VertexBuffer(Usage usage);
```
LaurentGomila: This constructor should be marked 'explicit'.

include/SFML/Graphics/VertexBuffer.hpp
```diff
@@ -0,0 +1,389 @@
...
+    ///
+    /// This function deallocates the corresponding buffer memory.
+    ///
+    /// After being cleared, the buffer has to be re-created before
+    /// performing any updates on it or drawing with it.
+    ///
+    /// \see create, update
+    ///
+    ////////////////////////////////////////////////////////////
+    void clear();
```
LaurentGomila: Is this function really necessary? What's the use case?

include/SFML/Graphics/VertexBuffer.hpp
```diff
@@ -0,0 +1,389 @@
...
+    /// \brief Get the underlying OpenGL handle of the vertex buffer.
+    ///
+    /// You shouldn't need to use this function, unless you have
+    /// very specific stuff to implement that SFML doesn't support,
+    /// or implement a temporary workaround until a bug is fixed.
+    ///
+    /// \return OpenGL handle of the vertex buffer or 0 if not yet created
+    ///
+    ////////////////////////////////////////////////////////////
+    unsigned int getNativeHandle() const;
```
LaurentGomila: What kind of useful stuff would users do on the VBO with this handle? (this is a real question, not a criticism 😄)

include/SFML/Graphics/VertexBuffer.hpp
```diff
@@ -0,0 +1,389 @@
...
+public:
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Usage specifiers
+    ///
+    ////////////////////////////////////////////////////////////
+    enum Usage
+    {
+        Stream,  ///< Constantly changing data
+        Dynamic, ///< Occasionally changing data
```
LaurentGomila: I really wonder what difference this flag (`Dynamic`) makes at driver level. I can understand that choosing between constantly changing or static data can make a difference, but "occasionally changing"? How do you optimize that?

Well, my point is, are there use cases where using `Dynamic` gives a significant performance boost compared to the 2 other options, in the context of SFML? Or is it just for the sake of completeness, given that this class is a VBO wrapper that can be used with OpenGL code?

include/SFML/Graphics/VertexBuffer.hpp
```diff
@@ -0,0 +1,389 @@
...
+    /// any attempt to use sf::VertexBuffer will fail.
+    ///
+    /// \return True if vertex buffers are supported, false otherwise
+    ///
+    ////////////////////////////////////////////////////////////
+    static bool isAvailable();
+
+private:
+
+    friend class RenderTarget;
```
LaurentGomila: Any reason for that? As far as I can tell, `RenderTarget::draw` only uses members that have a corresponding public getter.

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -317,6 +319,107 @@ void RenderTarget::draw(const Vertex* vertices, std::size_t vertexCount,
...
+        return;
+
+    // GL_QUADS is unavailable on OpenGL ES
+    #ifdef SFML_OPENGL_ES
+        if (vertexBuffer.m_primitiveType == Quads)
+        {
+            err() << "sf::Quads primitive type is not supported on OpenGL ES platforms, drawing skipped" << std::endl;
+            return;
+        }
+        #define GL_QUADS 0
```
LaurentGomila: Won't we have a preprocessor warning about GL_QUADS being defined twice (the other one being in `RenderTarget::draw(Vertex*)), when compiling with OpenGL ES?

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -317,6 +319,107 @@ void RenderTarget::draw(const Vertex* vertices, std::size_t vertexCount,
...
+        // If the texture we used to draw belonged to a RenderTexture, then forcibly unbind that texture.
+        // This prevents a bug where some drivers do not clear RenderTextures properly.
+        if (states.texture && states.texture->m_fboAttachment)
+            applyTexture(NULL);
+
+        // Update the cache
+        m_cache.useVertexCache = false;
+        m_cache.texCoordsArrayEnabled = true;
+    }
+}
```
LaurentGomila: I didn't check in detail, but maybe we could refactor some stuff, common between VA and VB drawing? At least the GL -> SFML primitive type conversion 😸

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -317,6 +319,107 @@ void RenderTarget::draw(const Vertex* vertices, std::size_t vertexCount,
...
+        // Bind vertex buffer
+        VertexBuffer::bind(&vertexBuffer);
+
+        // Always enable texture coordinates
+        if (!m_cache.texCoordsArrayEnabled)
+            glCheck(glEnableClientState(GL_TEXTURE_COORD_ARRAY));
+
+        glCheck(glVertexPointer(2, GL_FLOAT, sizeof(Vertex), reinterpret_cast<const void*>(0)));
+        glCheck(glColorPointer(4, GL_UNSIGNED_BYTE, sizeof(Vertex), reinterpret_cast<const void*>(8)));
+        glCheck(glTexCoordPointer(2, GL_FLOAT, sizeof(Vertex), reinterpret_cast<const void*>(12)));
```
LaurentGomila: Can't this be cached? This only needs to be called when a `VertexArray` was draw previously, which should happen less and less, and not anymore in SFML 3.

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -405,6 +509,9 @@ void RenderTarget::resetGLStates()
         if (shaderAvailable)
             applyShader(NULL);
 
+        if (vertexBufferAvailable)
+            glCheck(GLEXT_glBindBuffer(GLEXT_GL_ARRAY_BUFFER, 0));
```
LaurentGomila: Isn't it `VertexBuffer::bind(NULL)`?

src/SFML/Graphics/VertexBuffer.cpp
```diff
@@ -0,0 +1,373 @@
...
+    if (vertexCount >= m_size)
+    {
+        glCheck(GLEXT_glBufferData(GLEXT_GL_ARRAY_BUFFER, sizeof(Vertex) * vertexCount, 0, usageToGlEnum(m_usage)));
+
+        m_size = vertexCount;
+    }
+
+    glCheck(GLEXT_glBufferSubData(GLEXT_GL_ARRAY_BUFFER, sizeof(Vertex) * offset, sizeof(Vertex) * vertexCount, vertices));
+
+    glCheck(GLEXT_glBindBuffer(GLEXT_GL_ARRAY_BUFFER, 0));
```
LaurentGomila: I wonder if we should do like in `sf::Texture` and preserve the previous VBO binding. I don't like this strategy, but at least it avoids surprises for users that mix SFML classes and OpenGL calls.

src/SFML/Graphics/VertexBuffer.cpp
```diff
@@ -0,0 +1,373 @@
...
+void VertexBuffer::clear()
+{
+    if (m_buffer)
+    {
+        TransientContextLock contextLock;
+
+        glCheck(GLEXT_glBindBuffer(GLEXT_GL_ARRAY_BUFFER, m_buffer));
+        glCheck(GLEXT_glBufferData(GLEXT_GL_ARRAY_BUFFER, 0, 0, usageToGlEnum(m_usage)));
+        glCheck(GLEXT_glBindBuffer(GLEXT_GL_ARRAY_BUFFER, 0));
+    }
```
LaurentGomila: Can this be a simple call to `create(0)`? That would also avoid the bug of `m_size` not being set to 0 😛 

This would also add proof that this function may not be necessary in the public API.

src/SFML/Graphics/VertexBuffer.cpp
```diff
@@ -0,0 +1,373 @@
...
+}
+
+
+////////////////////////////////////////////////////////////
+void VertexBuffer::swap(VertexBuffer& right)
+{
+    std::swap(m_size,          right.m_size);
+    std::swap(m_buffer,        right.m_buffer);
+    std::swap(m_primitiveType, right.m_primitiveType);
+}
```
LaurentGomila: `m_usage` is not swapped.

src/SFML/Graphics/VertexBuffer.cpp
```diff
@@ -0,0 +1,373 @@
...
+    }
+
+    return available;
+}
+
+
+////////////////////////////////////////////////////////////
+void VertexBuffer::draw(RenderTarget&, RenderStates) const
+{
+    // This should never be called
```
LaurentGomila: Why is it a `Drawable` then?

include/SFML/Graphics/VertexBuffer.hpp
binary1248: Considering D3D also has several usage flags (see [here](https://msdn.microsoft.com/en-us/library/windows/desktop/ff476259(v=vs.85).aspx)), I take it it actually makes a difference somehow. At the end of the day, this is a usage _hint_ anyway. Depending on the driver implementation it can choose to trust us or doubt us. We aren't going to be doing it a favour by making it choose a wrong starting point. Also see chapter 28.2.2 of [this](https://www.seas.upenn.edu/~pcozzi/OpenGLInsights/OpenGLInsights-AsynchronousBufferTransfers.pdf).

include/SFML/Graphics/VertexBuffer.hpp
binary1248: If for some reason in some crazy use case, the user allocates a 512MB buffer on a GPU with only 1GB of VRAM, they might want to free the block for other usage without resorting to destroying the entire sf::VertexBuffer or resizing it twice.

include/SFML/Graphics/VertexBuffer.hpp
binary1248: Some really arcane magic (another word for optimization) where they e.g. map sections of the buffer themselves using `glMapBufferRange` in multiple threads for load balanced data streaming? 😄

src/SFML/Graphics/RenderTarget.cpp
binary1248: From the documentation:

> If a non-zero named buffer object is bound to the GL_ARRAY_BUFFER target (see glBindBuffer) while a vertex array is specified, pointer is treated as a byte offset into the buffer object's data store. Also, the buffer object binding (GL_ARRAY_BUFFER_BINDING) is saved as vertex array client-side state (GL_VERTEX_ARRAY_BUFFER_BINDING).

The buffer binding is saved as part of the array state. If you draw from another buffer, you will have to rebind the pointers since GL_ARRAY_BUFFER_BINDING changed. This is the weird kind of legacy emulation the driver has to pull off because we aren't making use of VAOs.

src/SFML/Graphics/RenderTarget.cpp
binary1248: Maybe...

src/SFML/Graphics/RenderTarget.cpp
binary1248: CI didn't seem to complain about it... doesn't mean other compilers won't though :P.

src/SFML/Graphics/VertexBuffer.cpp
binary1248: Mixing SFML and OpenGL is already complicated enough as it is. Until now we've told people that we don't mess with buffer binding state and they should make sure not to touch it or at least unbind anything that they bound. Changing it again would just make them more confused than they already were before.

I've already been thinking of a way for users to be able to draw in a separate context but to the same window framebuffer as the sf::RenderTarget. This way both SFML and user OpenGL can live in harmony without ever having to know about what the other is doing. This will deprecate the pushGLStates and popGLStates that is still necessary now.

src/SFML/Graphics/VertexBuffer.cpp
binary1248: The Drawable API until now has been built around the idea that everything is representable and drawable through a client side array. While sf::VertexBuffer is also conceptually a Drawable, it can't make use of this API because there is no client side pointer to pass to sf::RenderTarget. Maybe the Drawable API has to be extended/changed but this would be too big of a breaking change for SFML 2.

src/SFML/Graphics/RenderTarget.cpp
LaurentGomila: Ok, I didn't know (and expect) this kind of behavior.

src/SFML/Graphics/VertexBuffer.cpp
LaurentGomila: > I've already been thinking of a way for users to be able to draw in a separate context but to the same window framebuffer as the sf::RenderTarget

That would be the best option. Is it feasible?

src/SFML/Graphics/VertexBuffer.cpp
LaurentGomila: > While sf::VertexBuffer is also conceptually a Drawable

I'd say it is drawable, but it is not a `sf::Drawable`. That would translate to: it can be passed to `RenderTarget::draw`, which is exactly what you've done with the new overload.

A simple and probably common use case where making it a derived `sf::Drawable` fails, is a container of `sf::Drawable` instances and drawing them all from this container. That would call the `draw(sf::Drawable&)` overload, which would call `VertexBuffer::draw`, ending up in nothing at all being done.

At least, make `VertexBuffer::draw` call `draw(sf::VertexBuffer&)`, like `sf::VertexArray` does.

src/SFML/Graphics/VertexBuffer.cpp
binary1248: Yes... but not in this PR. 😄

src/SFML/Graphics/VertexBuffer.cpp
binary1248: Wouldn't this end up in infinite recursion?

RenderTarget::draw() -> VertexBuffer::draw() -> RenderTarget::draw() -> VertexBuffer::draw() ...

include/SFML/Graphics/VertexBuffer.hpp
LaurentGomila: What about `create(0)` then?

src/SFML/Graphics/VertexBuffer.cpp
LaurentGomila: No, because once you're in VertexBuffer::draw, you can pass a VertexBuffer& (*this) instead of a Drawable&, so it ends up in the correct overload and draws the vertex buffer.

binary1248: Yes... but the user's call into RenderTarget::draw() would already be using the VertexBuffer& overload instead of the Drawable& one...

LaurentGomila: That's why I mentioned the case where the user draws the vertex buffer instance through its base class.

```cpp
std::vector<sf::Drawable*> drawables;
drawable.push_back(new sf::VertexBuffer);
...
for (auto drawable : drawables)
    window.draw(*drawable); // calls RenderTarget::draw(sf::Drawable&)
```

binary1248:
@LaurentGomila All the issues should be addressed in the current version.

LaurentGomila:
Looks perfect.

What about:

> Have you tried to replace VA with VB in SFML implementation, to see the impact on performances (and usability)? The most interesting candidate would probably be sf::Text.

> And shouldn't we mark sf::VertexArray deprecated?

binary1248:
If I could be sure that VBOs are supported everywhere then it wouldn't be a problem. The reason why there is `sf::VertexBuffer::isAvailable()` is because SFML still claims to support everything OpenGL 1.2+. VBO support was only merged in OpenGL 1.5, so we might end up breaking some older systems unless we have code for both paths.

I also wouldn't mark `sf::VertexArray` as deprecated for the same reason. If we intend on staying backwards compatible with older systems, then they are still necessary. I would rather provide a recommendation to use `sf::VertexBuffer` whenever compatibility with 20 year old systems isn't a big concern of the user.

LaurentGomila:
An alternative would then be to add an implementation of sf::VertexBuffer that falls back to vertex array when VBOs are not supported, but that might complicate the code too much.

LaurentGomila:
Maybe the added example would deserve its own separate PR, so that it can be reviewed and discussed properly without blocking this one?

binary1248:
> An alternative would then be to add an implementation of sf::VertexBuffer that falls back to vertex array when VBOs are not supported, but that might complicate the code too much.

We can merge this first and think about optimizing the existing classes with VBO support one by one in their own PRs. It involves some work but maybe having both paths in the implementation is still the least complicated way to go forward.

> Maybe the added example would deserve its own separate PR, so that it can be reviewed and discussed properly without blocking this one?

I was already thinking about doing this anyway. SFML needs way more examples anyway, might as well start with this.

LaurentGomila:
Ok, so after you move the example to its own PR, I think we'll be ok here :+1:

binary1248:
Done. Example here: https://github.com/SFML/SFML/tree/feature/island_example

anthnich:
Had a chance to test this a little. Today, I moved my game's current implementation from sf::VertexArrays over to sf::Vertex/sf::VertexBuffer and thought I would give my casual observations.

- Everything converted over fine. Had 0 problems. Haven't noticed any oddities.
- Tested a game level that uses multiple tilemaps, with the largest being 36,736 verts.
- I observed a small drop in GPU usage (25% -> 23% from Windows Task Manager, 16% -> 13% from GPUz).
- I observed an increase in video memory usage (to be expected). ~79MB to ~93 MB.
- Didn't seem to be a major change in CPU usage. What I'm creating isn't very demanding to be honest.
- On my system, a casual profile saw the Draw time on my Tilemap class (all objects) go from 3.73s (on a 60s run) to 1.1s (!).

Note: I haven't tried to "break" this yet. Will be sure to report issues if I run into them.

Here's the sizes of the tilemaps in the level I tested.

Loaded 36736 Verts
Loaded 4092 Verts
Loaded 1144 Verts
Loaded 584 Verts
Loaded 384 Verts
Loaded 52 Verts
Loaded 120 Verts
Loaded 352 Verts
Loaded 2248 Verts
Loaded 30436 Verts
Loaded 32 Verts

src/SFML/Graphics/GLLoader.cpp
```diff
@@ -1,7 +1,7 @@
 ////////////////////////////////////////////////////////////
 //
 // SFML - Simple and Fast Multimedia Library
-// Copyright (C) 2007-2017 Laurent Gomila (laurent@sfml-dev.org)
+// Copyright (C) 2007-2015 Laurent Gomila (laurent@sfml-dev.org)
```
Iniquitatis: We're going back in time? :)

src/SFML/Graphics/GLLoader.hpp
```diff
@@ -1,7 +1,7 @@
 ////////////////////////////////////////////////////////////
 //
 // SFML - Simple and Fast Multimedia Library
-// Copyright (C) 2007-2017 Laurent Gomila (laurent@sfml-dev.org)
+// Copyright (C) 2007-2015 Laurent Gomila (laurent@sfml-dev.org)
```
Iniquitatis: Welcome to 2015 again. :)

binary1248:
Fixed.

anthnich:
In another game I developed, I moved my sf::VertexArrays code to sf::VertexBuffers and here are the results of my initial tests.  This is the vertexbuffer branch merged with the master branch as of 1/9/2018.

- It reduced the profiled (in a 60s run) draw time of NPC "vision cones" from 5.25s to .40s (!). These cones are changed nearly every tick, so there was a lot of vert data being manipulated.
- It reduced the profiled (in a 60s run) draw time of level tilemaps (walls, bgs, etc) from 1.11s to .83s. This improvement seems to line up with the results my other game, posted above.
- It dropped my GPU-z observed load by about 10%. VertexBuffers make up a significant amount of my world construction code.
- In addition to Windows, I also did some spot testing in Ubuntu 16.04 & Mac 10.13. Everything built and compiled fine. Everything looks to be working as expected and I didn't run into any oddities.

I'm still working on this game update, so if I run into any issues I'll let you know. All in all, I'm very pleased with these results. Thank you for the hard work. 👍

include/SFML/Graphics/VertexBuffer.hpp
```diff
@@ -0,0 +1,378 @@
...
+    ///
+    ////////////////////////////////////////////////////////////
+    static void bind(const VertexBuffer* vertexBuffer);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Set the type of primitives to draw
+    ///
+    /// This function defines how the vertices must be interpreted
+    /// when it's time to draw them:
+    /// \li As points
```
MarioLiebisch: I think this list is redundant here (and incomplete?). In addition, it might hide the fact that quads are deprecated and leave the impression that the strips/fans might not be supported. It should just refer to `sf::PrimitiveType`.

include/SFML/Graphics/VertexBuffer.hpp
```diff
@@ -0,0 +1,378 @@
...
+    ///
+    ////////////////////////////////////////////////////////////
+    bool update(const Vertex* vertices, std::size_t vertexCount, unsigned int offset);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Copy the contents of another buffer into this buffer
+    ///
+    /// \param vertexBuffer Vertex buffer whose contents to copy into this vertex buffer
+    ///
+    /// \return True if the copy was successful
```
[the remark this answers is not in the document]
binary1248: Hard to tell beforehand... As an example, if allocation of space for the copy fails because of some reason then it will fail. Reliably predicting GPU memory behaviour is almost impossible using OpenGL.

binary1248:
@mantognini @MarioLiebisch Documentation should be fixed now.

mantognini:
Thanks for the update! 

Quick question regarding concurrent update of the `VertexBuffer`: what is supported? I guess updating the same range of vertices at the same time from multiple threads is not legal, but do we explicitly support any concurrent update?

binary1248:
You can update the same range from multiple threads simultaneously... it wouldn't even be considered as undefined behaviour. The problem with that is there is no deterministic way to tell in which order they will actually take place in the end because we don't have any support for OpenGL fences.

Non-overlapping regions on the other hand... you don't really need to know the order if you just join the threads at some specific point in time. So yes... although not explicitly stated, concurrent updates of distinct regions is supported. I didn't know if such details were worth mentioning explicitly in the documentation, considering that people familiar with GL would know this anyway and even those who don't have already tried similar techniques in the past (and "failed" because they weren't aware of the sequential-ness of OpenGL).

RUSshy:
Nice! first step to support sprite batching in SFML ?

mantognini:
> So yes... although not explicitly stated, concurrent updates of distinct regions is supported. I didn't know if such details were worth mentioning explicitly in the documentation, considering that people familiar with GL would know this anyway

I think it should be added because the premise that SFML users (including me) know OpenGL is probably wrong. Beside, SFML is a bit more than just a OpenGL wrapper, I hope. :-)

binary1248:
Updated description with multi-threading information.

mantognini:
👍

eXpl0it3r:
Looks like there's a conflict.

binary1248:
Fixed.

mantognini:
For some reason, buildbot didn't re-run after your update...

PR #1312, opened 2017-10-29
Fixed a "used-before" init of a global  mutex

dfleury2:
Prevent this code to make an access violation due to bad ordering initialization of a global mutex 
`
#include <SFML/Audio.hpp>
sf::Sound aSound;
int main() {}
`

### DIFF
--- a/src/SFML/Audio/AlResource.cpp
+++ b/src/SFML/Audio/AlResource.cpp
@@ -35,7 +35,19 @@ namespace
 {
     // OpenAL resources counter and its mutex
     unsigned int count = 0;
-    sf::Mutex mutex;
+    sf::Mutex& mutex()
+    {
+        // To prevent uninitialized used of the mutex.
+        // Transform to bool and pointer to create a phoenix singleton if needed
+        static sf::Mutex gMutex;
+        return gMutex;
+    }
+
+    namespace {
+        sf::Mutex& force_creation = mutex();
+        // Hide a force creation on global initialization space (thread safe)
+        // To remove on C++11
+    }
 
     // The audio device is instantiated on demand rather than at global startup,
     // which solves a lot of weird crashes and errors.
@@ -50,7 +62,7 @@ namespace sf
 AlResource::AlResource()
 {
     // Protect from concurrent access
-    Lock lock(mutex);
+    Lock lock(mutex());
 
     // If this is the very first resource, trigger the global device initialization
     if (count == 0)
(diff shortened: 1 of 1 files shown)


LaurentGomila:
With this modification, why do we still need to force a creation at global scope?

dfleury2:
I did not check where AlResource constructor is called. So I have try to preserved safely the current behavior.
If it's called outside the global initialization, someone can call it concurrently from different thread.

LaurentGomila:
> If it's called outside the global initialization, someone can call it concurrently from different thread.

I still don't get it, can you explain?

dfleury2:
Prior to C++11, calling static variable constructor in functions are not thread safe.
The constructor is not guaranteed to be called once in a multi-threaded context.

Since SoundBuffer, SoundRecorder and SourceSource both inherit from AlResource, an user can call the AlResource constructor concurrently on different threads (in the main for example) and breaks the static gMutex constructor call.

In case of C++98, to avoid the multi-thread issue, I force the construction (if not already done by the mutex() call in AlResource constructor) in the global initialization phase because it's a single only thread zone.

C++11 has added a lock on static construction for these cases.

Hope I was more complete and clear this time (bare with me for my "english")

LaurentGomila:
That was very clear, thanks.

binary1248:
1. I don't see the point in going through this effort just so that people can stick with their bad habits of declaring SFML objects in namespace scope. There are always better alternatives, people need to learn about them. This isn't the only place where this "problem" exists. The reason why there haven't been PRs for the other places is because everybody who ran into problems like this just moved their declarations out of namespace scope and the problem disappeared.

2. SFML never guaranteed that declaring objects at namespace scope would always work. The same can be said about many other libraries out there as well. The mutexes scattered throughout the code are meant to prevent concurrent access to specific functions only after main() has been entered.

3. I don't understand how this PR is supposed to fix the actual problem. The `sf::Mutex` constructor is called during dynamic initialization, just like the function you declared to initialize the reference. You are replacing one function call with another and expect the global initialization order to change in a reliable way? This is not defined anywhere in the standard. If it works for you it is purely by chance and might break for others.

4. If you are really hell bent on declaring SFML objects in namespace scope, just link SFML dynamically. The dynamic library loader will guarantee that all static initialization that must be performed in dynamic libraries takes place when they are loaded, before your own statics are initialized. This is as nasty as workarounds come, but to each his own.

dfleury2:
1. Got reason you probably do not care, and not very important for this matter anyway

2. Glad I can improve SFML by providing a simple patch for this minor issue.

3. I am sorry my explanation was not clear enough for you to understand. Not sure I can do better.
Be sure that this patch works for every one due to proper use of order initialization, so that every object is fully constructed before use.

4. Got reason you probably do not care, and not very important for this matter anyway

Funny that it's blocked without even given a try. Strange state of mind.

Happy coding

Issue #1311, opened 2017-10-27
Android example does not provide the build files and directory structure for the Gradle build system


pjmlp:
The provided Android example under examples/android/ does not provide the Android Studio project file structure, still makes use of the deprecated ndk-build files and does not provide the necessary Gradle configuration files.

I already have a private fork with these issues sorted out and can gladly do a push request.

isubasinghe:
I am having trouble with the android example as well, it needs to be updated. Please update the tutorial as well

pjmlp:
@isubasinghe You can have a look at the pull request, https://github.com/SFML/SFML/pull/1315.

isubasinghe:
@pjmlp Thanks, I will try to compile it now.

isubasinghe:
I couldnt get it to work still : ( , oh well. Android support is still experimental isnt it, its not worth trying to make my game in SFML when I could just use libgdx right? , since android is supported there, and most importantly I can get it to compile.

pjmlp:
If you are fine with Java/Kotlin than it is probably better, unless you are quite comfortable with the NDK.

Have you tried to compile SFML with _-DANDROID_STL=stlport_shared_ as explained on the Wiki?

isubasinghe:
Yeah I am fine with using Java and I haven't really messed with the NDK before this. 

![image](https://user-images.githubusercontent.com/28187128/32686151-30e0e0b4-c6f4-11e7-92bb-8cc677dc4a01.png)

I got it to compile though, I had to change the gradle version to 4.1 and add 
"
 sourceSets.main {
        jniLibs.srcDir 'src/main/libs'
        jni.srcDirs = [] //disable automatic ndk-build call
   }
"
under defaultConfig

Thanks for the help and update xml file.

isubasinghe:
Ill probably stick to libgdx though for my Release version, I will rewrite it in SFML just for fun though.

Alia5:
Dunno, why this is on undiced, as NDK build, is considered deprecated for a long time, may someone enlighten me?

In any case [here](https://github.com/Alia5/SFML_AndroidStudio) is a (somewhat incomplete) example.

MarioLiebisch:
Simply a lack of time in combination with not having all potential related issues in mind when working on a PR or anything similar. Building for Android has been updated in pull request #1350 and the last open adjustments to build the example with gradle are on my HDD, waiting for a few fixes and being pushed as a new PR.

texus:
[...]

PR #1315, opened 2017-10-31
Adding support for the Android example to compile with Gradle

pjmlp:
This pull request relates to issue #1311.

As the current example still expects the old Ant based tooling, which is no longer supported, I have updated the example to rely on Gradle, while still making use of the underlying ndk-build scripts.


MarioLiebisch:
Going to include these commits/changes into #1350, so closing this PR, thanks. Still have to figure out a few issues with shared STL builds though. I also reverted the renamed "jni" folder, as `ndk-gdb` still expects it by default.

Is there any specific reason you've switched to stlport rather than using libc++?

pjmlp:
> Is there any specific reason you've switched to stlport rather than using libc++?

Not really, I think it had to do with the initial compilation I have done, and then it was easier to change it than just compile everything again.

I don't expect any major issue with libc++.

PR #1325, opened 2017-12-07
Fixed various type conversion/comparison warnings.

eXpl0it3r:
I have applied some fixed for the various type conversion/comparison warnings Visual Studio has been complaining about for a while now.

I wasn't certain for every place if the change was done properly and for some places it certainly raises the question if the chosen type is good.

There's one remaining warning that we can't easily fix. It's in the network module when calling `::send()`. [Linux](http://man7.org/linux/man-pages/man2/send.2.html) implements the socket send function with a `size_t`, while [Windows](https://msdn.microsoft.com/en-us/library/windows/desktop/ms740149(v=vs.85).aspx) uses `int`, causing a mismatch between platforms.

### DIFF
--- a/src/SFML/Network/Ftp.cpp
+++ b/src/SFML/Network/Ftp.cpp
@@ -463,7 +463,7 @@ Ftp::Response Ftp::getResponse()
                         }
 
                         // Save the remaining data for the next time getResponse() is called
-                        m_receiveBuffer.assign(buffer + static_cast<std::streamoff>(in.tellg()), length - static_cast<std::size_t>(in.tellg()));
+                        m_receiveBuffer.assign(buffer + static_cast<std::size_t>(in.tellg()), length - static_cast<std::size_t>(in.tellg()));
 
                         // Return the response code and message
                         return Response(static_cast<Response::Status>(code), message);


src/SFML/Network/Ftp.cpp
```diff
@@ -463,7 +463,7 @@ Ftp::Response Ftp::getResponse()
                         }
 
                         // Save the remaining data for the next time getResponse() is called
-                        m_receiveBuffer.assign(buffer + in.tellg(), length - in.tellg());
+                        m_receiveBuffer.assign(buffer + static_cast<std::streamoff>(in.tellg()), length - static_cast<std::size_t>(in.tellg()));
```
LaurentGomila: The first cast should not be needed.

eXpl0it3r: Visual Studio 2015 and 2017 32bits [still complain about it](https://ci.sfml-dev.org/builders/windows-vc14-32/builds/506/steps/compile/logs/warnings%20%2816%29).

According to [cppreference](http://en.cppreference.com/w/cpp/io/streamoff) `std::streamoff` is usually typedef-ed to `long long` and for the pointer manipulation you need an `int` thus there needs to be an explicit cast.

LaurentGomila: Ah, sorry I though it was the same type (std::streamoff vs std::streampos). And why not std::size_t by the way?

eXpl0it3r: I picked `std::streamoff` since that's what the compiler expected and converts to. I could try `std::size_t` as well, even though that will be `unsigned int` and not `int`, wouldn't it?

LaurentGomila: The first argument of assign is a const char*, which `buffer + <any integral type>` should produce no matter what the integral type is. So I don't get why we need a std::streamoff specifically.

eXpl0it3r: Changed it to `std::size_t`.  🙂

LaurentGomila: :+1:

PR #1326, opened 2017-12-21
Added support for interfacing with joysticks via DirectInput when it is available.

### LINKED ISSUE #1251: Unplugging gamepad crashes
A player reported today that it crashes on unplug. I asked him to give system specs, I'll update when he tells.

Also I found this on internet if it helps:
https://twitter.com/hartlabs/status/631934662378590210

Searched that **joyGetPosEx** function in SFML. 

https://github.com/SFML/SFML/blob/f053871a6c041292ecc059a8a2a435416f176390/src/SFML/Window/Win32/JoystickImpl.cpp#L198

It loops from 1 to Joystick::Count which is generally 8. I limited "Joystick::isConnected" checks to check only 1 to 4 in my code but I don't want to do the same in SFML. Maybe you guys would consider it. 

There are a lot of forum topics about this and they are not solved.

https://en.sfml-dev.org/forums/index.php?topic=18790.0
https://en.sfml-dev.org/forums/index.php?topic=13318.0

------

SFML 2.4.2
gcc (x86_64-posix-seh-rev1, Built by MinGW-W64 project) 6.3.0

binary1248:
Fixes #1251.

Test with this (might be worth making into an official example):
```cpp

// Headers
#include <SFML/Graphics.hpp>
#include <sstream>
#include <iomanip>
#include <string>
#include <map>


namespace
{
    typedef std::map<std::string, std::pair<sf::Text, sf::Text>> Texts;
    Texts texts;
    std::ostringstream sstr;

    // Helper to set text entries to a specified value
    template<typename T>
    void set(const char* label, const T& value)
    {
        sstr.str("");
        sstr << value;
        texts[label].second.setString(sstr.str());
    }
}


/// Entry point of application
///
/// \return Application exit code
///
int main()
{
    // Create the window of the application
    sf::RenderWindow window(sf::VideoMode(480, 640), "Joystick", sf::Style::Close);

    // Load the text font
    sf::Font font;
        return EXIT_FAILURE;

    // Axes labels in as C strings
    const char* axislabels[] = {"X", "Y", "Z", "R", "U", "V", "PovX", "PovY"};

    // Set up our string conversion parameters
    sstr.precision(2);
    sstr.setf(std::ios::fixed | std::ios::boolalpha);

    // Set up our label-value sf::Text objects
    texts["ID"].first.setPosition(5.f, 5.f);
    texts["ID"].second.setPosition(80.f, 5.f);

    for (unsigned int i = 0; i < sf::Joystick::AxisCount; ++i)
    {
        std::pair<sf::Text, sf::Text>& text = texts[axislabels[i]];

        text.first.setPosition(5.f, 5.f + ((i + 2) * font.getLineSpacing(14)));
        text.first.setString(std::string(axislabels[i]) + ":");

        text.second.setPosition(80.f, 5.f + ((i + 2) * font.getLineSpacing(14)));
    }

    for (unsigned int i = 0; i < sf::Joystick::ButtonCount; ++i)
    {
        sstr.str("");
        sstr << "Button " << i;
        std::pair<sf::Text, sf::Text>& text = texts[sstr.str()];

        text.first.setPosition(5.f, 5.f + ((sf::Joystick::AxisCount + i + 2) * font.getLineSpacing(14)));
        text.first.setString(sstr.str() + ":");

        text.second.setPosition(80.f, 5.f + ((sf::Joystick::AxisCount + i + 2) * font.getLineSpacing(14)));
    }

    for (Texts::iterator i = texts.begin(); i != texts.end(); ++i)
    {
        i->second.first.setFont(font);
        i->second.first.setCharacterSize(14);
        i->second.first.setFillColor(sf::Color::White);

        i->second.second.setFont(font);
        i->second.second.setCharacterSize(14);
        i->second.second.setFillColor(sf::Color::White);
    }

    while (window.isOpen())
    {
        // Handle events
        sf::Event event;
        {
            if ((event.type == sf::Event::Closed) ||
               ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Escape)))
            {
                window.close();
                break;
            }
        }

        // Clear the window
        window.clear();

        // Search for a connected joystick
        for (unsigned int i = 0; i < sf::Joystick::Count; ++i)
        {
            if (sf::Joystick::isConnected(i)) {
                // Update the label-value sf::Text objects based on the current joystick state

                // Joystick identification
                sstr.str("");
                sstr << "Joystick " << i << ":";
                texts["ID"].first.setString(sstr.str());
                texts["ID"].second.setString(sf::Joystick::getIdentification(i).name);

                // Joystick axes
                for (unsigned int j = 0; j < sf::Joystick::AxisCount; ++j)
                {
                    if (sf::Joystick::hasAxis(i, static_cast<sf::Joystick::Axis>(j)))
                        set(axislabels[j], sf::Joystick::getAxisPosition(i, static_cast<sf::Joystick::Axis>(j)));
                }

                // Joystick buttons
                for (unsigned int j = 0; j < sf::Joystick::getButtonCount(i); ++j)
                {
                    sstr.str("");
                    sstr << "Button " << j;

                    set(sstr.str().c_str(), sf::Joystick::isButtonPressed(i, j));
                }

                // Only print information for the first connected joystick
                break;
            }
        }

        // Draw the label-value sf::Text objects
        for (Texts::const_iterator i = texts.begin(); i != texts.end(); ++i)
        {
            window.draw(i->second.first);
            window.draw(i->second.second);
        }

        // Display things on screen
        window.display();
    }
}
```

### DIFF
--- a/extlibs/headers/mingw/dinput.h
+++ b/extlibs/headers/mingw/dinput.h
@@ -1,0 +1,2467 @@
+/*
+ * Copyright (C) the Wine project
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+#ifndef __DINPUT_INCLUDED__
+#define __DINPUT_INCLUDED__
+
+#define COM_NO_WINDOWS_H
+#include <objbase.h>
+#include <_mingw_dxhelper.h>
+
+#ifndef DIRECTINPUT_VERSION
+#define DIRECTINPUT_VERSION	0x0800
+#endif
+
+/* Classes */
+DEFINE_GUID(CLSID_DirectInput,		0x25E609E0,0xB259,0x11CF,0xBF,0xC7,0x44,0x45,0x53,0x54,0x00,0x00);
+DEFINE_GUID(CLSID_DirectInputDevice,	0x25E609E1,0xB259,0x11CF,0xBF,0xC7,0x44,0x45,0x53,0x54,0x00,0x00);
+
+DEFINE_GUID(CLSID_DirectInput8,		0x25E609E4,0xB259,0x11CF,0xBF,0xC7,0x44,0x45,0x53,0x54,0x00,0x00);
+DEFINE_GUID(CLSID_DirectInputDevice8,	0x25E609E5,0xB259,0x11CF,0xBF,0xC7,0x44,0x45,0x53,0x54,0x00,0x00);
+
+/* Interfaces */
+DEFINE_GUID(IID_IDirectInputA,		0x89521360,0xAA8A,0x11CF,0xBF,0xC7,0x44,0x45,0x53,0x54,0x00,0x00);
+DEFINE_GUID(IID_IDirectInputW,		0x89521361,0xAA8A,0x11CF,0xBF,0xC7,0x44,0x45,0x53,0x54,0x00,0x00);
+DEFINE_GUID(IID_IDirectInput2A,		0x5944E662,0xAA8A,0x11CF,0xBF,0xC7,0x44,0x45,0x53,0x54,0x00,0x00);
+DEFINE_GUID(IID_IDirectInput2W,		0x5944E663,0xAA8A,0x11CF,0xBF,0xC7,0x44,0x45,0x53,0x54,0x00,0x00);
+DEFINE_GUID(IID_IDirectInput7A,		0x9A4CB684,0x236D,0x11D3,0x8E,0x9D,0x00,0xC0,0x4F,0x68,0x44,0xAE);
+DEFINE_GUID(IID_IDirectInput7W,		0x9A4CB685,0x236D,0x11D3,0x8E,0x9D,0x00,0xC0,0x4F,0x68,0x44,0xAE);
+DEFINE_GUID(IID_IDirectInput8A,		0xBF798030,0x483A,0x4DA2,0xAA,0x99,0x5D,0x64,0xED,0x36,0x97,0x00);
+DEFINE_GUID(IID_IDirectInput8W,		0xBF798031,0x483A,0x4DA2,0xAA,0x99,0x5D,0x64,0xED,0x36,0x97,0x00);
+DEFINE_GUID(IID_IDirectInputDeviceA,	0x5944E680,0xC92E,0x11CF,0xBF,0xC7,0x44,0x45,0x53,0x54,0x00,0x00);
+DEFINE_GUID(IID_IDirectInputDeviceW,	0x5944E681,0xC92E,0x11CF,0xBF,0xC7,0x44,0x45,0x53,0x54,0x00,0x00);
+DEFINE_GUID(IID_IDirectInputDevice2A,	0x5944E682,0xC92E,0x11CF,0xBF,0xC7,0x44,0x45,0x53,0x54,0x00,0x00);
+DEFINE_GUID(IID_IDirectInputDevice2W,	0x5944E683,0xC92E,0x11CF,0xBF,0xC7,0x44,0x45,0x53,0x54,0x00,0x00);
+DEFINE_GUID(IID_IDirectInputDevice7A,	0x57D7C6BC,0x2356,0x11D3,0x8E,0x9D,0x00,0xC0,0x4F,0x68,0x44,0xAE);
+DEFINE_GUID(IID_IDirectInputDevice7W,	0x57D7C6BD,0x2356,0x11D3,0x8E,0x9D,0x00,0xC0,0x4F,0x68,0x44,0xAE);
+DEFINE_GUID(IID_IDirectInputDevice8A,	0x54D41080,0xDC15,0x4833,0xA4,0x1B,0x74,0x8F,0x73,0xA3,0x81,0x79);
+DEFINE_GUID(IID_IDirectInputDevice8W,	0x54D41081,0xDC15,0x4833,0xA4,0x1B,0x74,0x8F,0x73,0xA3,0x81,0x79);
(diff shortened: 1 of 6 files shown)


LaurentGomila:
Would it be possible to split the implementations (JoystickImplDInput / JoystickImplWinMM -- or at least 2 distinct sets of functions within JoystickImpl) for better readability and maintainability?

MarioLiebisch:
Oh, didn't notice those new header files from MinGW before. IMO neither of them should be included since they're part of the compiler toolset/platform SDK.

binary1248:
@LaurentGomila @MarioLiebisch Those headers are meant for compatibility with some (older?) versions of MinGW that for some reason don't come with dinput.h out of the box.

See https://ci.sfml-dev.org/builders/windows-gcc-492-tdm-32/builds/404

If you check, GLFW does the same for probably similar reasons.

MarioLiebisch:
Really confusing. [According to this bug report](https://sourceforge.net/p/tdm-gcc/feature-requests/18/) it seems like the headers should be there.

binary1248:
They are there in the 64-bit distribution: https://ci.sfml-dev.org/builders/windows-gcc-492-tdm-64/builds/401

However, they are missing in the 32-bit distribution. I just double checked on the CI server.

binary1248:
I split the DirectInput implementation into its own functions further down in JoystickImpl.cpp.

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -33,9 +33,128 @@
...
+
+        DIOBJECTDATAFORMAT data[8 + 4 + sf::Joystick::ButtonCount];
+        DIDATAFORMAT format;
+    };
+
+    const ObjectDataFormat objectDataFormat;
+
+    IDirectInput8W* directInput = NULL;
+    HMODULE dinput8dll = NULL;
+    HRESULT(WINAPI *DirectInput8CreateFunc)(HINSTANCE, DWORD, REFIID, LPVOID*, LPUNKNOWN);
```
LaurentGomila: `DirectInput8CreateFunc` doesn't follow our naming conventions (starts with capital letter). There's no need for it to be a global, it is only used in `initializeDInput`.

The DirectInput-related variables could be stuffed into a structure (like `DirectInputContext`), to limit the number of globals and make things clearer.

`objectDataFormat` is also used locally in a single place, but since its address is taken, it may have to live as long as DInput is used; do you know more about this stuff?

binary1248:
Fixed.

MarioLiebisch:
> However, they are missing in the 32-bit distribution. I just double checked on the CI server.
That sounds like a bug or issue. Why would that be the case?

binary1248:
https://sourceforge.net/projects/tdm-gcc/files/TDM-GCC%20Installer/Previous/1.1309.0/tdm-gcc-4.9.2.exe/download

You can have a look yourself.

There was an issue: https://sourceforge.net/p/mingw-w64/discussion/723797/thread/0176cb85/
The files got added in r312: https://sourceforge.net/p/mingw-w64/code/312/

Don't ask me why the distributions took so long to react. Fact is that this is only an issue for 4.9.2, meaning it will most probably never be fixed for that version since newer versions don't have this problem any more. Either we decide to drop support for 4.9.2 32-bit just because of the missing header or we bundle it ourselves.

dabonetn:
Is there any guide for getting a mxe setup with a different branch like yours?
I'd love to see if this fixes the crash error in attractmode under windows when wireless controllers disconnect.

anthnich:
Tested this. No crashing when unplugging. Controller works as expected, didn't notice any issues. Ran a performance profile for 60 secs, nothing out of the ordinary.

MarioLiebisch:
Just wanted to try this, still having MinGW-W64 (I think?) GCC 5.3.0 installed. While it got the Direct3D headers and others, it's missing the Direct Input one for whatever reason. Updated to 6.3.0 and the header still isn't there.

I'd just suggest dropping those headers, letting CMake look for them, and then print a notice about them missing. If they're missing, compile the legacy API version only? What do you think?

MarioLiebisch:
Now got GCC 4.8.3 from win-builds.org and it compiles fine…

eXpl0it3r:
Honestly, I feel it's not worth the effort. If your compiler doesn't have it, then it's a broken compiler and you should just update it.

If anything, an (advanced) feature flag in CMake would also do the trick. Either it works and if it doesn't you could pick the old implementation.

extlibs/headers/mingw/dinput.h
```diff
@@ -0,0 +1,2467 @@
+/*
+ * Copyright (C) the Wine project
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
```
eXpl0it3r: I assume LGPL headers are fine to be shipped, right?

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -152,6 +202,12 @@ namespace priv
 ////////////////////////////////////////////////////////////
 void JoystickImpl::initialize()
 {
+    // Try to initialize DirectInput
+    initializeDInput();
+
+    if (!directInput)
+        err() << "DirectInput not available, falling back to Windows joystick API" << std::endl;
```
eXpl0it3r: If someone doesn't want to use a joystick and doesn't have direct input available will this still output the error?

eXpl0it3r:
I think, shipping our own header is fine. Will make our life easier as we won't have to deal with broke compilers.

src/SFML/Window/Win32/JoystickImpl.cpp
binary1248: People have already "complained" a few times about not being able to disable joystick polling, so yes... this error will still be output as soon as window events are pumped. Realistically, these can only be Windows 98 or older systems, so if they get the rest of SFML running on those kinds of systems they deserve a medal.

extlibs/headers/mingw/dinput.h
binary1248: We are only working around a toolchain "bug". Either you use the one you already have or you use the one we provide. In both cases you are `#include`ing an LGPL header.

binary1248:
The workaround is trivial anyway, so I don't see any problems with it. AFAIK GLFW decided to go down the same route.

Foaly:
I don't think so.
What would be the use case for that? SFML 2.5 will be release soonish™ and you can already get nightly builds of the master branch from the build servers.

dabonetn:
Because I have no idea how to get the 2.5 sfml into a mxe environment so I can compile attract mode.

see here under cross compile.

http://attractmode.org/docs/Compile.html

MarioLiebisch:
I've never used mxe before (nor have I heard of it), but since it's obviously to cross-compile Win32 binaries, you'll just have to compile those yourself and replace those mxe uses. You'll probably have more luck in their community/communities though.

eXpl0it3r:
> Because I have no idea how to get the 2.5 sfml into a mxe environment so I can compile attract mode.

The same way you'd get a new "backported" build in there. If you can do that, you can also do it with a new SFML version.

Besides, backporting for minor versions is not a thing we do, especially not for features.

PR #1330, opened 2017-12-26
pushing upstream openbsd support from package.

devnexen:


MarioLiebisch:
Is it intentional to use the Android joystick implementation? If so, it might be a good idea to rename or move it.

devnexen:
Yes intentional :-) but it concerns only few files among Android implementations in another hand ... what do you think ?

src/SFML/Window/CMakeLists.txt
```diff
@@ -116,6 +116,12 @@ elseif(SFML_OS_LINUX OR SFML_OS_FREEBSD)
             ${SRCROOT}/FreeBSD/JoystickImpl.cpp
             ${SRCROOT}/FreeBSD/JoystickImpl.hpp
         )
+    elseif(SFML_OS_OPENBSD)
+        set(PLATFORM_SRC
+            ${PLATFORM_SRC}
+            ${SRCROOT}/Android/JoystickImpl.cpp
```
mantognini: Since the android implementation is a dummy one, I'd rather see those files being copied to a new directory.

mantognini:

src/SFML/Window/CMakeLists.txt
devnexen: Ok will do later in the day fair enough :-)

mantognini:

eXpl0it3r:
@devnexen We've gotten quite some changes with our CMake setup. Could you resolve the merge conflicts?

eXpl0it3r:

binary1248:
The rebase doesn't look clean. Stuff like `INSTALL_MISC_DIR` that was removed in ee08e18726788e4855218fa27319b2313b185d0f are added back in here.

binary1248:
Browsing through the SDL source code, I noticed that [SDL shares the same joystick code among all BSDs](http://hg.libsdl.org/SDL/file/fbfacc66c65c/src/joystick/bsd/SDL_sysjoystick.c). SFML's current FreeBSD Joystick implementation makes use of usbhid instead of the joy driver. usbhid should be available on OpenBSD as well no?

devnexen:
This code is not compiled for OpenBSD (e.g. SDL_JOYSTICK_USBHID is set to zero in our case)

binary1248:
I'm not an expert in OpenBSD by any means, but does this come down to whether usbhid is installed by the user? Or is it not available at all as a package on OpenBSD? [There seem to be OpenBSD man pages for usbhid with the remark that usbhid has been available since OpenBSD 3.0](https://man.openbsd.org/usbhid.3).

devnexen:
usbhid is available but no joystick support. The SDL code you mentioned earlier needs joystick.h header which we di not have.

devnexen:
To come back to SFML until now we used empty android implementation  as you can see here https://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/multimedia/sfml/patches/patch-src_SFML_Window_JoystickImpl_hpp?rev=1.2&content-type=text/x-cvsweb-markup

binary1248:
OK I guess. 👍

cmake/Config.cmake
```diff
@@ -117,3 +121,12 @@ else()
     message(FATAL_ERROR "Unsupported compiler")
     return()
 endif()
+
+# define the install directory for miscellaneous files
```
binary1248: This block is no longer used in the current master revision.

include/SFML/Config.hpp
```diff
@@ -85,6 +85,10 @@
 
         // FreeBSD
         #define SFML_SYSTEM_FREEBSD
+    #elif defined(__OpenBSD__)
```
binary1248: Empty line missing before this one.

src/SFML/Window/CMakeLists.txt
```diff
@@ -259,7 +265,7 @@ if(SFML_OS_LINUX)
     target_link_libraries(sfml-window PRIVATE UDev)
 elseif(SFML_OS_WINDOWS)
     target_link_libraries(sfml-window PRIVATE winmm gdi32)
-elseif(SFML_OS_FREEBSD)
+elseif(SFML_OS_FREEBSD OR SFML_OS_OPENBSD)
```
binary1248: If we can't interface with joysticks via usbhid do we still need to link to it on OpenBSD?

src/SFML/Window/JoystickImpl.hpp
```diff
@@ -98,7 +98,7 @@ struct JoystickState
 
     #include <SFML/Window/iOS/JoystickImpl.hpp>
 
-#elif defined(SFML_SYSTEM_ANDROID)
+#elif defined(SFML_SYSTEM_ANDROID) || defined(SFML_SYSTEM_OPENBSD)
```
binary1248: This should be split out and include the OpenBSD implementation instead.

eXpl0it3r:

src/SFML/Window/OpenBSD/JoystickImpl.hpp
```diff
@@ -0,0 +1,117 @@
...
+#ifndef SFML_JOYSTICKIMPLOPENBSD_HPP
+#define SFML_JOYSTICKIMPLOPENBSD_HPP
+
+
+namespace sf
+{
+namespace priv
+{
+////////////////////////////////////////////////////////////
+/// \brief Android implementation of joysticks
```
Foaly: Copy pasta :)

eXpl0it3r:
Merged in 9da895da8baae8d82bf2c870a50f987a03fd8f18

Issue #1329, opened 2017-12-26
Moving analog stick very slowly does not trigger sf::Event::JoystickMoved


naezith:
When I move the analog stick very slowly, sf::Event::JoystickMoved event does not trigger. However it's fine if I move it fast.

I tried this with Dualshock 4 controller and Logitech F310 which has two modes, X (XInput) and D (Dualshock?), I tried both of them and it's same on those too.

I tried this with @binary1248 's new DirectInput patch (https://github.com/SFML/SFML/pull/1326) and without it. Same problem exists.

One player also tried the same and approves.

On the other hand, Windows Control Panel -> Set up USB Game controllers -> Properties does not have the same issue, maybe it's polling the current value instead of working with events.

I already implemented XInput for my game. If I use an XInput emulator for these gamepads, Game's XInput library works fine. 

I am incredibly busy these days so I can't provide a code snippet but it's just the most basic pollEvent loop.

Windows 10
Dualshock 4
Logitech F310
g++ (x86_64-posix-seh-rev1, Built by MinGW-W64 project) 7.2.0

MarioLiebisch:
This is most likely not directly a bug, but the joystick threshold instead. Check out [`sf::Window::setJoystickThreshold()`](https://www.sfml-dev.org/documentation/2.4.2/classsf_1_1Window.php#aa45b8f54e29a6f59f1fc7ee66b2fab68).

It will basically remove tiny movement, but should most likely work in a different way (i.e. only allowing the axes to center properly).

naezith:
Yes you are correct, sorry about it. @MarioLiebisch

MarioLiebisch:
Well, you're actually right about this being a bug I think. It should prevent tiny movements, but if you've moved the axis far enough, it should be considered a movement.

MarioLiebisch:
Just checked the source and we do indeed compare the previous state with the current one rather than with the threshold itself. So we basically filter out tiny relative movement rather than tiny absolute movement.

LaurentGomila:
... and that's on purpose, to avoid spamming the event queue; this is *not* supposed to act as a dead zone.

MarioLiebisch:
But even then it shouldn't drop movement below the threshold. It should just add it up to emit an event once the total movement since the last event was significant enough, correct?

For example, moving the axis by – no idea – 0.01 between measurements shouldn't trigger the event, but after 100 steps it should emit a movement by 1, because all those tiny moves happened.

naezith:
[...]

PR #1331, opened 2017-12-28
Fixed small joystick movement getting lost due to the set axes threshold

MarioLiebisch:
This fixes issue #1329.

Simple test program:

    #include <SFML/Window.hpp>
    #include <iostream>

    int main()
    {
        sf::Window window(sf::VideoMode(100, 100), "SFML");
    
        window.setJoystickThreshold(5);

        std::vector<std::string> names = {
            "X",    ///< The X axis
            "Y",    ///< The Y axis
            "Z",    ///< The Z axis
            "R",    ///< The R axis
            "U",    ///< The U axis
            "V",    ///< The V axis
            "PovX", ///< The X axis of the point-of-view hat
            "PovY"  ///< The Y axis of the point-of-view hat
        };

        while (window.isOpen()) {
            sf::Event event;
                switch (event.type) {
                    case sf::Event::Closed:
                        window.close();
                        break;
                    case sf::Event::JoystickMoved:
                        std::cout << names[event.joystickMove.axis] << " " << event.joystickMove.position << std::endl;
                        break;
                    default:
                        break;
                }
            }

            window.display();
        }
    }

The expected result is axis movement in (at least) increments of 5. No movement must be lost, even if you move the stick very tiny steps/slow.

### DIFF
--- a/src/SFML/Window/WindowImpl.cpp
+++ b/src/SFML/Window/WindowImpl.cpp
@@ -86,7 +86,10 @@ m_joystickThreshold(0.1f)
     // Get the initial joystick states
     JoystickManager::getInstance().update();
     for (unsigned int i = 0; i < Joystick::Count; ++i)
+    {
         m_joystickStates[i] = JoystickManager::getInstance().getState(i);
+        std::fill_n(m_previousAxes[i], static_cast<std::size_t>(Joystick::AxisCount), 0.f);
+    }
 
     // Get the initial sensor states
     for (unsigned int i = 0; i < Sensor::Count; ++i)
@@ -176,6 +179,10 @@ void WindowImpl::processJoystickEvents()
             event.type = connected ? Event::JoystickConnected : Event::JoystickDisconnected;
             event.joystickButton.joystickId = i;
             pushEvent(event);
+
+            // Clear previous axes positions
+            if (connected)
+                std::fill_n(m_previousAxes[i], static_cast<std::size_t>(Joystick::AxisCount), 0.f);
         }
 
         if (connected)
@@ -186,7 +193,7 @@ void WindowImpl::processJoystickEvents()
                 if (caps.axes[j])
                 {
                     Joystick::Axis axis = static_cast<Joystick::Axis>(j);
-                    float prevPos = previousState.axes[axis];
+                    float prevPos = m_previousAxes[i][axis];
                     float currPos = m_joystickStates[i].axes[axis];
                     if (fabs(currPos - prevPos) >= m_joystickThreshold)
                     {
@@ -196,6 +203,8 @@ void WindowImpl::processJoystickEvents()
                         event.joystickMove.axis = axis;
                         event.joystickMove.position = currPos;
                         pushEvent(event);
+
+                        m_previousAxes[i][axis] = currPos;
                     }
                 }
             }
--- a/src/SFML/Window/WindowImpl.hpp
+++ b/src/SFML/Window/WindowImpl.hpp
@@ -270,10 +270,11 @@ private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    std::queue<Event> m_events;                          ///< Queue of available events
-    JoystickState     m_joystickStates[Joystick::Count]; ///< Previous state of the joysticks
-    Vector3f          m_sensorValue[Sensor::Count];      ///< Previous value of the sensors
-    float             m_joystickThreshold;               ///< Joystick threshold (minimum motion for "move" event to be generated)
+    std::queue<Event> m_events;                                              ///< Queue of available events
+    JoystickState     m_joystickStates[Joystick::Count];                     ///< Previous state of the joysticks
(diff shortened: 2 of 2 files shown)


Foaly:
I took a look at why the build fail on half the build servers. There is a warning saying `template argument uses unnamed type` and an error saying `cannot decrement expression of enum type`. I compiled it locally using clang. Locally I only get a warning. Are warning treated as errors on the build servers? If so I think we should enable it in cmake as well. 
Turns out the lines 91 and 185 use `std::fill_n` together with `Joystick::AxisCount` which is of unknown type, because its a value of an enum. Casting to an int fixes the compilation. But it makes you think. Is an enum really the best way of storing non changing counts in a class, if it leaves the data typeless? I think a const static int would do the same thing more typesafe.

src/SFML/Window/WindowImpl.cpp
```diff
@@ -176,6 +179,10 @@ void WindowImpl::processJoystickEvents()
             event.type = connected ? Event::JoystickConnected : Event::JoystickDisconnected;
             event.joystickButton.joystickId = i;
             pushEvent(event);
+
+            // Clear previous axes positions
+            if (connected)
+                std::fill_n(m_previousAxes[i], Joystick::AxisCount, 0.f);
```
Foaly: This would be:
```cpp
std::fill_n(m_previousAxes[i], static_cast<int>(Joystick::AxisCount), 0.f);
```

LaurentGomila:
It's not typeless (such a thing does not exist in C++) or less typesafe, the type here is an enum type and arithmetic operations don't exist on enum types, for obvious reasons.

Foaly:
You are of course right! Sorry for mixing that up. I still think that it is more desirable to use a type that supports arithmetic operations. Especially if you take a look at [Joystick.hpp](https://github.com/SFML/SFML/blob/master/include/SFML/Window/Joystick.hpp#L49) and see how that the enum members are defined manually anyway.

MarioLiebisch:
Changed. Wondering what the settings are as I didn't get any issues compiling with VS 2017, even on higher warning levels.

Foaly:
This is ok, but like I said I think the proper fix for this is to drop the enum and make the *Count members of type `static const std::size_t`. The API wouldn't change, but it will prevent pointless casts.

MarioLiebisch:
Just keep in mind that all the casts (except `dynamic_cast`) are compile time only, so there's no real penalty, while also avoiding a value to be maintained separately (in the unlikely case some of the stuff here changes).

Foaly:
I am aware of that. My point is like I said if you look at [Joystick.hpp](https://github.com/SFML/SFML/blob/master/include/SFML/Window/Joystick.hpp#L49) the values are already maintained manually, they don't take any advantage of being in an enum. So in order to avoid users from having similar problems (which are obviously hard to track as different compiler handle them differently) and for better readability I would change it to a static member.

MarioLiebisch:
Hm, yep, true. That's actually a bit pointless for axes, but since buttons and joysticks aren't named… Would indeed be better to have them as constants. Could just change them to `static const unsigned char` with this PR. What do the others think?

Just for reference:

    /// \brief Constants related to joysticks capabilities
    ///
    enum
    {
        Count       = 8,  ///< Maximum number of supported joysticks
        ButtonCount = 32, ///< Maximum number of supported buttons
        AxisCount   = 8   ///< Maximum number of supported axes
    };

LaurentGomila:
Don't forget that strictly speaking, changing the type of a publicly accessible variable breaks API and ABI compatibility, so this is not supposed to happen before SFML 3.

Since it's really not critical, can't we wait until SFML 3 and make them constexpr?

MarioLiebisch:
Yeah, true, so could just keep the casts for now.

Foaly:
This! I don't understand this! It literally boggles my mind! There is good progress to be made here and you (as in the team) would rather have a half-assed workaround fix and do a proper fix "sometimes later". When is this ever going to happen? I understand the desire to have a stable API, but given the current state of SFML development I would like to stongly suggest to rethink this rule. Seriously I think this attitude of "na its fine. we'll do it (properly) in SFML 3" is extremly toxic to any kind of progress.

In order to not delay the merging of this PR, please go ahead for all I care, but we need to have a discussion about this! I will open a thread in the forum.

LaurentGomila:
Aren't you exagerating this a little bit? Except avoiding a cast to int in this PR, I think nobody will care about these constants going from enum to int. Or maybe I miss the reason why it is so important, then I'll be glad to hear it :)

I feel like it is more urgent to merge something that strictly fixes the joystick issue, and leave such minor considerations for later. Unless you want the PR to remain blocked in endless discussions?

About API/ABI compatibility: it truly is important, but yes, we can be less strict when the change is tiny and really brings value.

Foaly:
Yes this is exagerated, because it is just the tip of the iceberg of a bigger underlying problem. I heard this "let's do it later" soooo may times it just makes me a bit angry. Sorry! 

You know that me out of all people does not want to endlessly block a discussion. Thats why I open a [thread](https://en.sfml-dev.org/forums/index.php?topic=23511.0) on the forum. Like I said, please continue with this PR.

PR #1119, opened 2016-07-31
GPU local texture data copying

binary1248:
In many applications, it is common for the user to stitch smaller textures into bigger ones in order to minimize the amount of OpenGL state changes necessary when drawing. This is commonly known as atlasing. Even `sf::Font` makes use of this when loading its glyphs.

The problem with atlasing in SFML is that each time a new texture has to be inserted into the atlas, users would have to copy the texture data back to RAM, resize/update it and transfer it back on to the GPU. This costs a lot of time. With this PR, the user (given the proper hardware support) will be able to copy the existing texture to a temporary texture on the GPU, resize the original texture to its new size and copy the old texture data over. This does not involve any copying of data between the GPU and CPU.

``` cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::Font font;
    if (!font.loadFromFile("resources/sansation.ttf"))
        return -1;

    for (unsigned int i = 32; i < 127; i++)
        font.getGlyph(i, 1000, false, 0);
}
```

Testing with this code yields 2 orders of magnitude performance increase. On my system, the time it takes for the font to load the codepoints decreases from 20 seconds down to under 1 second.

This is only indirectly testing the texture copying and is a contrived example, since it is meant to be a minimal demonstration. If anybody else has texture copy benchmarks, they can provide the data here as well.

In addition to this feature, I went ahead and optimized `sf::Font::cleanup()` to deallocate its pixel buffer storage. This way, if the user loads glyphs using a huge character size and later reuses the same `sf::Font` object for another font at smaller character sizes, they won't be stuck with a `std::vector` using up more memory than necessary.

PR #1034, opened 2015-12-28
Extending Font and Text to support multiple textures (no more discarded glyphs)

hobby8:
**SFML Tasks**

* [x] Update code w.r.t. outlined text (#840)
* [x] Reject and close this PR

----

Extending `Font` and `Text` to support multiple textures so that glyphs are not discarded in case one texture gets filled.
This can be useful in any of the following scenarios:
- Big text
- Large alphabets
- OpenGL implementations with a small maximum texture size (e.g. software renderers)

Discussed in this [forum thread](http://en.sfml-dev.org/forums/index.php?topic=19510.0).

PR #840, opened 2015-03-24
Added support for outlined text

zsbzsb:
Considering all the requests [over the years](https://www.google.com/search?q=sfml+text+outline+site:sfml-dev.org) for outlined text ([and the totally hackish workarounds](http://en.sfml-dev.org/forums/index.php?topic=15071.msg106581#msg106581)) and also that my current project will definitely need this at some point, I went ahead and implemented it. I made the PR for code review.

Some key points are the following.

- Outlined glyphs are stored in the same page as normal glyphs (can be changed if this is deemed that it will fill up the page texture too fast).
- ```Text::[set|get]Color(...)``` is now marked as deprecated, instead everyone should use ```Text::[set|get]FillColor(...)```
- ```Font::getGlyph(...)``` has a new parameter, outline thickness which is defaulted to 0 (to avoid API breakage). The parameter determines how thick the outline will be in pixels.
- Since outlined glyphs are stored in the same page/texture the key type has been changed to Int64 to continue being unique now that the outline thickness is part of the key.
- I refactored ```Text::ensureGeometryUpdate()``` and split the duplicated code to append vertices to separate functions ```::addLine(...)``` and ```::addGlyphQuad(...)```. These can be renamed if someone could think of more appropriate names.
- ~~Outline thickness is unsigned, I don't see any easy ways with freetype to make the outline protrude inwards with a negative thickness.~~ It is a float, but a negative value causes distorted rendering.
- Nothing should change with the current rendering of text if the user decides not to use outlines.

As stated before, please review this and give some feedback.

```
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window({ 700, 400 }, "Outline test");
    window.setVerticalSyncEnabled(true);

    sf::Font font;
    font.loadFromFile("C:/Windows/Fonts/arial.ttf");

    sf::Text text(L"\u0476ZyG\noLPfkdf", font, 140);
    text.setPosition({ 20, 20 });

    text.setFillColor(sf::Color::Red);

    text.setOutlineThickness(5);
    text.setOutlineColor(sf::Color::White);

    text.setStyle(sf::Text::Underlined | sf::Text::Italic | sf::Text::StrikeThrough);

    while (window.isOpen())
    {
        sf::Event e;
        while (window.pollEvent(e))
        {
            if (e.type == sf::Event::Closed)
                window.close();
        }

        window.clear();

        window.draw(text);

        window.display();
    }
}
```

Example Image

PR #1333, opened 2017-12-29

hobby8:
Extending `Font` and `Text` to support multiple textures so that glyphs are not discarded in case one texture gets filled.
This can be useful in any of the following scenarios:
- Big text
- Large alphabets
- OpenGL implementations with a small maximum texture size (e.g. software renderers)

Reimplementation of #1034 adjusted for #840 and #1119.

### DIFF
--- a/include/SFML/Graphics/Font.hpp
+++ b/include/SFML/Graphics/Font.hpp
@@ -37,6 +37,7 @@
 #include <map>
 #include <string>
 #include <vector>
+#include <list>
 
 
 namespace sf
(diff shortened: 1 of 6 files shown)


MarioLiebisch:
Is this really a "bug"? I'd call it "feature".

One further suggestion for the patch: When loading a `sf::Font`, it might be a good idea to allow the user to optionally set a maximum texture size. Just because you can use stuff like 4k textures doesn't necessarily mean you want to do that (e.g. for systems with lower end VRAM).

LaurentGomila:
> Is this really a "bug"? I'd call it "feature".

I'd say it's a bug because it doesn't add any new feature, it just allows some corner cases to work as expected.

> When loading a sf::Font, it might be a good idea to allow the user to optionally set a maximum texture size. Just because you can use stuff like 4k textures doesn't necessarily mean you want to do that (e.g. for systems with lower end VRAM).

Limited texture size would result in more textures, which means an equal amount of RAM but lower performances. The only way to save RAM is to generate less glyphs, and I'd rather let users decide it with less characters, less different sizes, etc. Otherwise we're just back to the current situation where SFML silently skips glyphs that don't fit.

MarioLiebisch:
I was actually thinking about potentially less texture swaps (in and out of VRAM) if you're not using all the textures. But yeah, in the end it's just another edge case in some way.

eXpl0it3r:
We had some changes, including but not limited to #1461 and as a result, there are now merge conflicts. Would you be able to resolve them and rebase onto master, @hobby8? 🙂

eXpl0it3r:
Superseded by #1645 which includes all the changes of this PR, but is brought up to date with master

PR #1337, opened 2018-01-02
Replace auto_ptr with unique_ptr to fix build errors when using the C++17 standard

ssrobins:
I was unable to build the Audio component of SFML when using VS2017 with the C++17 standard (/std:c==17 flag or use set(CMAKE_CXX_STANDARD 17) in CMake).  This solved the problem.

### DIFF
--- a/src/SFML/Audio/AudioDevice.cpp
+++ b/src/SFML/Audio/AudioDevice.cpp
@@ -107,7 +107,7 @@ bool AudioDevice::isExtensionSupported(const std::string& extension)
     // This device will not be used in this function and merely
     // makes sure there is a valid OpenAL device for extension
     // queries if none has been created yet.
-    std::auto_ptr<AudioDevice> device;
+    std::unique_ptr<AudioDevice> device;
     if (!audioDevice)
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
Thanks for the contribution! Unfortunately SFML 2.x is still using C++03 for backwards compatibility and for SFML 3 the work [has already been done](https://github.com/SFML/SFML/commits/unstable). As such we can't really merge this PR.

As for Visual Studio, you can configure the project to not handle deprecation warnings as errors or you can just disable that one specific warning (somehow). `std::auto_ptr` still compiles fine with VS 2017. 🙂

ssrobins:
It was such an easy fix, I figured there must have been a good reason it wasn't done already!

std::auto_ptr works with VS2017 when C++ Language standard is set to ISO C++14 Standard (/std:c++14), but is completely missing when set to ISO C++17 Standard (/std:c++17).

I ended up applying a patch to my build process to use unique_ptr.  I'll happily upgrade to SFML 3 when it comes out!

Anyway, thanks for your attention.

PR #1339, opened 2018-01-07
Restore fullscreen of a non-visible Unix window

dabbertorres:
Fixes the issue described here: https://en.sfml-dev.org/forums/index.php?topic=23032


eXpl0it3r:
Merged in 6d122e96e1367aa9159122712f23d2c8074184e7

PR #1235, opened 2017-05-26
Scancode

mantognini:
## Purpose

Scancodes are a very important feature that's been asked to be implemented in SFML for a long time. It gives developers a layout independent way of reading keyboard input. Suppose you're using standard "WASD" walking input scheme in your game. If you just say "if pressed sf::Keyboard::Key::A then go left", it'll work as expected only on a QWERTY and a bunch of layouts where "WASD" is in the place gamers usually expect it to be. But, it'll work wrong on AZERTY layout, which has "A" where "Q" on QWERTY is.

Scancodes prevent this problem. If we say "if pressed sf::Keyboard::ScanA then go left", we're telling something like "if pressed the key which is in the place where A on QWERTY keyboards is then go left".

## Testing

Either of the follow applications can be used to test:
- SFML-Input: https://github.com/eXpl0it3r/SFML-Input
- SFML-Test-Events (feature branch: scancode): https://github.com/mantognini/SFML-Test-Events/tree/scancode

Snapshot Builds: https://artifacts.sfml-dev.org/by-branch/feature/scancode/

## SFML Tasks

* [x] Create a complete list of key mappings
    * [x] Scancodes
    * [x] ~Description (Names)~ -> OS & Language dependant
    * [x] Windows
    * [x] Linux
    * [x] macOS
* [x] Windows: Implement scancode retrieval for events
* [x] ~Windows: Look into issue with TDM build~ -> TDM builds were decommissioned
* [x] Discussion Points
    * [x] Rename `sf::Keyboard::Key` names for Numpad to align with scancode naming?
    * [x] Find a name for Key 45
    * [x] Check on Linux & macOS whether Key 42 also maps to scancode `0x56`
* [x] Find a better way to handle descriptions on macOS or use fixed names
* [x] Check all methods are implemented for all OS
  * [x] Windows
      * [x] Key events
      * [x] Keyboard static function
      * [x] Localize
      * [x] Unlocalize
      * [x] Description
  * [x] Linux
      * [x] Key events
      * [x] Keyboard static function
      * [x] Localize
      * [x] Unlocalize
      * [x] Description
  * [x] macOS
      * [x] Key events
      * [x] Keyboard static function
      * [x] Localize
      * [x] Unlocalize
      * [x] Description
* [x] Check that the documentation is aligned
* [ ] (optional) Consider providing additional images to explain the scancode positions better
* [x] Squash commits for a better overview of the change
[...]

Issue #1045, opened 2016-01-21
More keys for sf::Keyboard


TankOs:
Before #7 is being worked on (or even decided), I'd like to simply add some missing keys to the sf::Keyboard::Key enum.

I did an implementation for Linux here; it would be nice if someone could come up with implementations for Windows, BSD and OSX as well: [feature/more_keys](https://github.com/SFML/SFML/tree/feature/more_keys)

mantognini:
FYI: [comparison against master](https://github.com/SFML/SFML/compare/feature/more_keys) (I just find it useful to see the diff in the browser. :-))

I'll keep this in mind and try to have a look before the world ends. I don't have keyboard with those keys so it's hard to test anything... But anyone is free to chip in, of course!

eXpl0it3r:
If we add more now, what's the reasoning not to do so for other keys as well, or rather why the ones selected here?

MarioLiebisch:
Guess because they're most common/standard and have identical meaning on all keyboard layouts (but are still missing)? I think we've got pretty much all others?

binary1248:
Pushed back to 2.5 since the Windows and OS X implementations are still not ready. For 2.4, #1022 can be merged instead of this in order to fix #1012.

mantognini:
Yeah, I won't have time for that for 2.4...

victorlevasseur:
It seems that Window/Unix/InputImpl.cpp hasn't been updated to work for the new keys.

SirSharpest:
I can update Inputlmpl.cpp with the key symbols if this is still being done, or is the issue abandoned for now?

MarioLiebisch:
Feel free to do so, then create a pull request for this branch once ready.

CelticMinstrel:
That linked patch looks like a good start, but I have two objections to it:

1. NumpadReturn should be called NumpadEnter. I've never seen that key labelled "return".
2. NumpadComma should be renamed, as it's a decimal point on my keyboard. I'm not sure exactly what the best name is, but Comma is not it. I see from that patch that X11 calls it "separator", and I believe SDL calls it "decimal". Either of those are better than "comma".

[...]

PR #1340, opened 2018-01-08
Added Exclamation mark/Paragraph (!/§) key handling

stephanedamo:
The !/§ key found on AZERTY keyboards wasn't handled by SFML. This should fix the issue on Windows and Unix. I didn't looked at the other platforms since i'm less familiar with them.

I chose "Exclamation" for the key name but it could be ExclamationMark too, or Paragraph.

### DIFF
--- a/include/SFML/Window/Keyboard.hpp
+++ b/include/SFML/Window/Keyboard.hpp
@@ -99,6 +99,7 @@ public:
         SemiColon,    ///< The ; key
         Comma,        ///< The , key
         Period,       ///< The . key
+        Exclamation,  ///< The Exclamation mark / Paragraph key
         Quote,        ///< The ' key
         Slash,        ///< The / key
         BackSlash,    ///< The \ key
--- a/src/SFML/Window/Unix/InputImpl.cpp
+++ b/src/SFML/Window/Unix/InputImpl.cpp
@@ -61,6 +61,7 @@ bool InputImpl::isKeyPressed(Keyboard::Key key)
         case Keyboard::LBracket:   keysym = XK_bracketleft;  break;
         case Keyboard::RBracket:   keysym = XK_bracketright; break;
         case Keyboard::Comma:      keysym = XK_comma;        break;
+        case Keyboard::Exclamation:keysym = XK_exclam;       break;
         case Keyboard::Period:     keysym = XK_period;       break;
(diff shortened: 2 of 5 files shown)


stephanedamo:
In Keyboard.hpp I think it's better to put this new key at the end of the list (before KeyCount) to avoid breaking someone's app if they used the enum values directly.

eXpl0it3r:
For reference also see #1045 which adds some control keys to the enum. Maybe this can all be connected into one PR. We however still need implementations for all platforms before this can be merged.

Also for reference see #7 and #1235.

stephanedamo:
#1045 is 2 years old and still pending. Don't we need to make SFML pull policy more flexible, allowing to add features 'as they come' when they're finished, instead of waiting for all platforms to be supported ? Unsupported keys would still return -1 on OSX if not implemented.

mantognini:
> Don't we need to make SFML merge policy more flexible, allowing to add features 'as they come' when they're finished, instead of waiting for all platforms to be supported ?

I think it would be unwise for two reasons, at least. Firstly, SFML is a multiplateforme library. If we start having features for one plateforme but not the others, then it loses its purpose. Secondly, when devising a new feature we need to have the big picture in mind, and this includes other OSes, because otherwise we risk creating some API that cannot fit nicely for another OS.

Now, in #7's discussion, the main wish seemed to be the addition of scancodes. #1235 adds an API for that. Please have a look if you want to contribute a Linux implementation. :)

The other concern emitted in #7 was the amount of Unknown keys. In #1045, @eXpl0it3r rightfully asked why those keys were added and not others. Until now we haven't actually decided what was the criteria to add a key to our `Key` enum or not. So let's try to do that now in order to move on with specific implementations such as this PR, shall we?

I'd argue that the `Key` enumeration reflects any localised key, hence we should not really refrain from adding more keys. However, I think it makes sense to limit ourselves to latin/roman based scripts.

Opinion on that?

CelticMinstrel:
Regarding the question of the key name...

1. § is "Section", not "Paragraph".
2. Generally I think it should be named after whichever character it produces if the Shift key is _not_ pressed. (I'm not sure which that is here, though.)

eXpl0it3r:
The author has vanished from GitHub. I'll close this PR and we can continue the discussion on potentially expanding the non-scancode key codes further.

PR #1342, opened 2018-01-10
Worked around render target bug on macOS

### LINKED ISSUE #1132: RenderTexture not working in OSX on 2.4.0
I get a black screen when using a render texture with SFML 2.4.0 (edit: or 2.4.1). If I switch to using the window object directly and bypass the render texture, it works correctly. If I use SFML 2.3.2 with the render texture, it also works correctly.

I am using a late-2013 Macbook Pro with the Intel Iris Pro.

Here's a small-ish program that reproduces the issue on 2.4.0:

```
#include <SFML/Graphics.hpp>
#include <iostream>

int main(int argc, char* argv[]) {
    sf::RenderWindow window(sf::VideoMode(1024, 768), "Repro");

    sf::RenderTexture render_texture;

    const sf::Color color(150, 50, 250);

    sf::CircleShape shape(16);
    shape.setFillColor(color);

    if(!render_texture.create(window.getSize().x, window.getSize().y)) {
        std::cerr << "Failed to create render texture\n";
        return -1;
    }

    while(window.isOpen()) {
        sf::Event event;

            if(event.type == sf::Event::Closed) {
                window.close();
            }

            if(event.type == sf::Event::KeyReleased) {
                if(event.key.code == sf::Keyboard::Escape || event.key.code == sf::Keyboard::Q) {
                    window.close();
                }
            }
        }

        render_texture.clear(sf::Color::Black);

        shape.setPosition(32, 32);

        render_texture.draw(shape);

        render_texture.display();

        const sf::Texture& texture = render_texture.getTexture();
        sf::Sprite sprite(texture);

        window.draw(sprite);

        window.display();
    }

    return 0;
}
```

eXpl0it3r:
Fixes #1132

I thought to open this PR here, so we can discuss if it works or not here and either merge it or reject it. 🙂 

@mantognini [mentioned](https://github.com/SFML/SFML/issues/1132#issuecomment-342885505)
> I invite others to test this as well by checking out the bugfix/osx_render_target branch.

@achpile [mentioned](https://github.com/SFML/SFML/issues/1132#issuecomment-344192612)
>this patch didn't fixed my issue.
everything renders normally if there's something drawn on RenderWindow except the sprite with RenderTexture (as in my example code).

### DIFF
--- a/src/SFML/Graphics/RenderTarget.cpp
+++ b/src/SFML/Graphics/RenderTarget.cpp
@@ -373,6 +373,12 @@ void RenderTarget::resetGLStates()
     // Check here to make sure a context change does not happen after activate(true)
     bool shaderAvailable = Shader::isAvailable();
 
+    // Workaround for states not being properly reset on
+    // macOS unless a context switch really takes place
+    #if defined(SFML_SYSTEM_MACOS)
+        setActive(false);
+    #endif
+
     if (setActive(true))
     {
         // Make sure that extensions are initialized


JonnyPtn:
For what it's worth this looks good to me. Using steps from original issue, I get a blank screen on master, and a purple circle on this branch

mantognini:
Thanks @JonnyPtn for testing this!

@achpile, could you try it again? I'm wondering if you forgot to clear the old binaries before recompiling SFML and trying the `bugfix/osx_render_target` branch. This has been a common issue with many testers.

In the meantime, since we have no better fix and it seems to fix at least some issues for some users, I'd say we go ahead and merge this. Or do you have another, better fix in mind, @binary1248?

achpile:
@mantognini sorry, i no longer have macbook :(

eXpl0it3r:
Since the change is so small and potentially fixes the issue, let's just merge it and if similar issues arise, we can always revert it again.

binary1248:
@mantognini No idea if there is another fix. During testing I just realized that resetting the states didn't result in any actual change unless you switch contexts back and forth before. This behaviour is totally unexpected and is not documented anywhere. As you noticed yourself, this is a regression since it used to work in older OS X versions. Considering that there is probably not that much software out there that makes such liberal use of multiple contexts like SFML does, I don't expect Apple to go out of their way to fix this any time soon.

mantognini:
Alright, so let's merge this!

PR #1343, opened 2018-01-11
Require CMake 3.5.1

Ceylo:
### Forum thread
https://en.sfml-dev.org/forums/index.php?topic=22954.msg159889

It is a pre-requisite for https://github.com/SFML/SFML/pull/1335
(fixes a bug/behavior with target_link_libraries() and INSTALL_INTERFACE)


mantognini:

mantognini:
Well.. seems like this actually brings some issues with compiler detection. https://ci.sfml-dev.org/builders/osx-clang-el-capitan/builds/401 Any idea why?

texus:
In cmake 3 the compiler id for the osx compiler changed from "Clang" to "AppleClang", see https://cmake.org/cmake/help/v3.0/policy/CMP0025.html

mantognini:
Thanks for the info! I guess [this](https://github.com/SFML/SFML/blob/master/cmake/Config.cmake#L84) can then be updated to use the proper values of [`CMAKE_<LANG>_COMPILER_ID`](https://cmake.org/cmake/help/v3.5/variable/CMAKE_LANG_COMPILER_ID.html#variable:CMAKE_<LANG>_COMPILER_ID), or anything more modern, actually.

LaurentGomila:
Does this mean there's absolutely no way to apply the new CMake configuration with < 3.5.1? It was released in march 2016, so although it is not bleeding-edge, it's still quite recent.

Ceylo:
For a few seconds I thought I didn't test CMake config before doing the PR. But no, it does work locally, with CMake 3.10 and Xcode 9. How… I don't know… *investigating*

LaurentGomila:
So what's the thing with CMake 3.5.1?

Ceylo:
I don’t understand your question. According to the doc the policy to have CMAKE_CXX_COMPILER_ID contain AppleClang rather than Clang is available since CMake 3.0. What do you want to do with CMake < 3.5.1 ?

There’s probably a mistake in their doc though, because they say the default is the old behavior, which should give Clang as compiler id as we don’t set the policy. But I get AppleClang.

mantognini:

LaurentGomila:
> I don’t understand your question

I think it's me who doesn't understand 😀 

I've read the linked forum thread and PR, but I still don't get why this PR is named "Now require CMake 3.5.1". What's so special with this specific version, and is there really no way to do with older versions?

Ceylo:
Oh, I just took the most recent version that is available on all the main distros according to the answer in the forum. This seems fair enough for users to easily upgrade, and it’ll allow SFML developers to have as few as possible bloated code with « if cmake version > 3.x then blabla » and automatically benefit from « latest » features.

If it was just me I would have required an even more recent version. Because the more recent the cmake version, the more find_package() modules provide imported targets, which makes linking against external stuff much easier. That’s the case of find_package(Freetype) for example since CMake 3.10.

I didn’t check everything that’s been added between CMake 3.0 and 3.5 if this is your question, but I see no reason to even consider older version if this one is easily available to everyone.

Ceylo:
Both for you and my personal knowledge I did a small summary of what's available in different CMake versions, that looks relevant for SFML, and is significant "enough".

[3.0](https://cmake.org/cmake/help/latest/release/3.0.html):
- INTERFACE libraries
- target export
- CMAKE_SYSROOT

[3.1](https://cmake.org/cmake/help/latest/release/3.1.html):
- VS2015 generator
- Windows Phone support
- target_source()
- CMAKE_CXX_STANDARD / CMAKE_CXX_EXTENSIONS
- BundleUtilities module to handle @rpath on macOS
- FindGLEW provides imported targets

[3.2](https://cmake.org/cmake/help/latest/release/3.2.html):
- wider UTF-8 support
- ANDROID_API_MIN

[3.3](https://cmake.org/cmake/help/latest/release/3.3.html):
- linking against /usr/lib/libfoo.a no longer asks the linker to search for the library (e.g. -lfoo) and now links by full path

[3.4](https://cmake.org/cmake/help/latest/release/3.4.html):
- Windows 10 SDK selection
- CMAKE_<LANG>_COMPILER_LAUNCHER for easier use of ccache/distcc
- Initial Swift support

[3.5](https://cmake.org/cmake/help/latest/release/3.5.html):
- -T option to choose toolset
- improved cmake -E copy/copy_if_different/copy_directory : multiple input support
- escaping fixes for Xcode generator

And for next releases...
[3.6](https://cmake.org/cmake/help/latest/release/3.6.html):
- Clang/C2 support for VS2015
- Various additions for simpler toolchain files
- Some support for Clang-tidy
- FindPkgConfig can provide imported targets for found packages
- Clang support in Cygwin

[3.7](https://cmake.org/cmake/help/latest/release/3.7.html):
- Cross compilation support for Android with toolchain files
- VS2017 generator
- Toolchain file improvements

[3.8](https://cmake.org/cmake/help/latest/release/3.8.html):
- C# support
- CShartpUtilities module
- static framework on macOS (wtf?)
- FindOpenGL provides imported targets

[3.9](https://cmake.org/cmake/help/latest/release/3.9.html):
- object libraries
- CMAKE_ANDROID_NDK_DEPRECATED_HEADERS
- GENERATOR_IS_MULTI_CONFIG global property 
- doxygen_add_docs() function

[3.10](https://cmake.org/cmake/help/latest/release/3.10.html):
- FindFreetype provides imported targets
- FindOpenGL also searches for EGL,  and provides both variable and imported target

LaurentGomila:
There's no point picking some random newer version, if we can perfectly go with an older one. That's a limitation that adds absolutely nothing.

So what would be the real minimum version of CMake required for SFML?

Ceylo:
That's not random. That's what's available on all common current systems.
If you go on that path, yes you can do everything with CMake 1. Manually. Actually you could also drop CMake and write your own so that there's no version issue. And write everything each time you need something. It looks to be what you want.

You want developers to waste time again for this kind of discussion next time they'd like a feature from next minor CMake version? Or write boilerplate code to handle different versions?

At some point you also need to simplify life for the developers that could help SFML.
There's no point in requiring a lower version if that one is available to everyone, and simplifies developers life, be it for features useful right now or in a few PRs.

Sorry for writing this way. I'm fed up. Because you consider it's a random choice and that it adds nothing but a limitation.

MarioLiebisch:
**TL; DR:** You're right with this request, but the approach isn't good. Please take the time and read below, because I basically agree with upping the version.

I think you didn't understand Laurent's intention. This is just a hint for the *minimum* version *required* to actually build everything successfully. It's not an optimal or target version (like other build systems have, e.g. the Android SDK).

Let's assume everything builds just fine with version 2.8 then it will also build with 3.5 or 3.9 or even 4.x.

However when you set the minimum version to 3.5 it will still compile with 3.5 or later, but it will no longer compile with 2.8. But it will be broken for everyone being stuck with any older version, even if there's *no reason to be*.

**Big however:** You're stating that this is a prerequisite for pull request #1335. That on the other hand *is* a very valid reason for setting the minimum version. As you've said, it is required to fix a bug that otherwise cripples the change/feature or at least breaks it in some way. It's not a good idea to mix different things into single pull requests, but when things belong together, there's nothing stopping you from merging them.

**So what you should do:** This branch/pull request on itself adds nothing to the project other than restriction potential users/raising the requirements without adding anything in return. Instead, you should **put the minimum version change into the other pull request with the feature requiring this**. Ideally also add a note to the pull request why this specific version (don't just link a forum thread, add a quick explanation like "bug xyz, causing abc, is fixed in 3.5.1").

P.S.: One of our work servers is still stuck with CMake 2.8 and I have no control over that version. Just because versions are readily available doesn't mean they're present everywhere. Newer versions might be available, but they could also clash with other dependencies. As such, just because a dev could use 3.0+ with SFML doesn't mean they necessarily can or want to use that with some other part of their project.

LaurentGomila:
It seems like you really didn't get my point.

All I say is: use the minimum required version of CMake needed to run your latest changes. If your PR runs with CMake 3.3 then why would we write "require CMake 3.5.1"? Makes no sense.

And... if your PR actually needs CMake 3.5.1, then I'm still waiting for the technical explanation because unless I'm missing something, it's not written anywhere.

cmake/Config.cmake
```diff
@@ -81,7 +81,7 @@ endif()
 # detect the compiler and its version
 # Note: on some platforms (OS X), CMAKE_COMPILER_IS_GNUCXX is true
 # even when CLANG is used, therefore the Clang test is done first
-if(CMAKE_CXX_COMPILER MATCHES ".*clang[+][+]" OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+if(CMAKE_CXX_COMPILER MATCHES ".*clang[+][+]" OR CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
```
MarioLiebisch: The `.*` isn't really needed here (and will in theory slow down the expression), because `MATCHES` will just match as a sub string just fine, if you're not using `^` or `$`. So looking for `Clang` will match both `AppleClang`  as well as `Clang` or `Clangtastic`.

For the same reasons you could actually use this opportunity to also simplify the first expression, only looking for `clang`.

MarioLiebisch:
@LaurentGomila From the PR text:

> (fixes a bug/behavior with target_link_libraries() and INSTALL_INTERFACE)

But as mentioned, I'd agree, there should be a few more details right in the text.

Ceylo:
> This is just a hint for the _minimum_ version required to actually build everything successfully.

Actually no. As I found out recently, a different requirement changes the policies' values that CMake will use. That is, without changing the CMake program version that actually processes the CMakeLists files. This is this original reason for making this PR.

I don't know if there is a policy that can be set for the issue I'm having, because it doesn't look to be documented.
The list of policies is there: https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html#manual:cmake-policies(7) and none seems to refer to INSTALL_INTERFACE and target_link_libraries(), which is where my first problem is.

> It's not a good idea to mix different things into single pull requests, but when things belong together, there's nothing stopping you from merging them.

This is really not flexible. I split what can be split and still makes sense for a few reasons:
- having what's ready merged earlier -> this means fewer conflicts and work available earlier to everyone
- avoiding huge PRs -> because it becomes harder to review, to test, because it delays integration and because it makes it very likely for the people working on the PR to often get conflicts because of PRs merged in the meantime

I agree that splitting just to split (ie. the PR doesn't have value on itself anymore) is wrong. I don't consider this is the case of current PR. For the reasons above and the simplified CMake code.

> P.S.: One of our work servers is still stuck with CMake 2.8 and I have no control over that version.

Do you mean at your workplace?

> It seems like you really didn't get my point.
> 
> All I say is: use the minimum required version of CMake needed to run your latest changes.  If your PR runs with CMake 3.3 then why would we write "require CMake 3.5.1"? Makes no sense.

Yes, that's what I understood and that's what I am against. For these reasons:
- doing an upgrade has a cost, so it'd better be done all at once if possible
- when working on CMake and you realize that you'd need a feature from an newer version, you either waste time doing a workaround, introduce harder to read CMake files, or pay again the time to do the upgrade

The PR about the <Project>Config most likely doesn't need exactly CMake 3.5.1. It needs a newer CMake, probably at least 3.0. Until I find another issue, waste more time, then conclude that it should actually need 3.1. And I'm only talking about myself. Consider all the people who try to improve the situation with "modern" CMake.

Also if it was only about what I needed, I would have required CMake 3.10. Because of the imported targets that are available in CMake 3.10 and 3.8 and which would simplify my other PR. With CMake 3.5 I'll need to rewrite myself a few INTERFACE libraries.

To give more examples about what I'm playing with for the SFMLConfig PR:
- CMake 3.2: The install(EXPORT) command now works with an absolute DESTINATION even if targets in the export set are installed with a destination or usage requirements specified relative to the install prefix. The value of the CMAKE_INSTALL_PREFIX variable is hard-coded into the installed export file as the base for relative references.
- CMake 3.3: The install(TARGETS) command learned to support generator expressions in the DESTINATION value.
- CMake 3.4: The install(DIRECTORY) command DESTINATION option learned to support generator expressions.
- CMake 3.4: The install(FILES) command DESTINATION option learned to support generator expressions.
- CMake 3.5: The install(DIRECTORY) command learned to support generator expressions in the list of directories.

So I'd very much appreciate to avoid useless headaches.

```
Here are the versions of CMake in Debian / Ubuntu which might be relevant:
 
Ubuntu 12.04 Precise (EOL) = 2.8.7
Debian 7 Wheezy (EOL)      = 2.8.9
Ubuntu 14.04 Trusty        = 2.8.12
Debian 8 Jessie            = 3.0.2
Ubuntu 16.04 Xenial        = 3.5.1
Debain 9 Stretch           = 3.7.2
Ubuntu 17.10 Artful        = 3.9.1
 
Ubuntu 18.04 Bionic (dev)  = 3.9.5
Debian 10 Buster (dev)     = 3.9.5
 
I think you definitely want to support Xenial and Stretch which are the
latest "big" versions of both distros. This would mean a minimum version
of 3.5. Beyond that it's up to you how much you want to support.
```
We **know** that we can ask people to have at least CMake 3.5.1. So why waste time about minimal upgrades, and waste time during development on CMake files?

cmake/Config.cmake
Ceylo: Indeed, thanks!

LaurentGomila:
I don't want you to get more headaches, modernizing the CMake config of SFML is already a tough task and you're doing an amazing job with it. You seem to have spend an incredible amount of time reading the various docs, and for sure you know what you're doing.

So, I'd say, go on and finish this, don't get stopped by my questions or comments, even if you don't like them. In case it was not clear, I totally agree that we should take this opportunity to write a new CMake config as modern and clean as possible, without being stuck on some specific version of CMake. We'll then get feedback from users, and if it appears that we need to lower our CMake requirements, then fine, we can still find workarounds at that moment.

However, when you're done with this PR and satisfied with the result, we should still write the lowest possible version in `cmake_minimum_required`, because it's not nice to reject a bunch of users for no good reasons.

> We know that we can ask people to have at least CMake 3.5.1

Is it really safe to draw conclusions based on this listing of Debian and Ubuntu versions?

Ceylo:
I'm not sure I understood. You want to merge this PR then change the requirement based on what has been used for the SFMLConfig PR ?

> Is it really safe to draw conclusions based on this listing of Debian and Ubuntu versions?

Which other distros should be checked?
Homebrew on macOS provides CMake 3.10.1.
I'm not considering Windows since installing through package manager isn't the default way on this platform, so you can always get the latest version from CMake website.

LaurentGomila:
> I'm not sure I understood. You want to merge this PR then change the requirement based on what has been used for the SFMLConfig PR ?

All I'm saying (and this has been my only concern, since the begining) is that the requirements should reflect what the CMake files do require. Nothing less, nothing more.. I'm not even trying to enforce anything, just giving my opinion.

> Which other distros should be checked?
Homebrew on macOS provides CMake 3.10.1.
I'm not considering Windows since installing through package manager isn't the default way on this platform, so you can always get the latest version from CMake website.

One can always get the latest version of CMake on any OS -- it's an open source project after all. But our user base is large, and it's hard to predict how many of them will be annoyed when we change our requirements. That's why we have to be reasonable, and only require what's required. You wouldn't make the CMake config fail on non-C++17 compilers if our code only uses C++11, would you?

feliwir:
Maybe we could go straight to CMake 3.7 since that will probably be a requirement for builtin Android support in CMake?

MarioLiebisch:
Haven't had a chance to try it. Not sure it would work with our current setup, but if it does, might indeed be an interesting jump.

feliwir:
@MarioLiebisch i did the switch to CMake 3.7 at work recently. It works like a charm. All custom android toolchain files can be pretty much dropped

eXpl0it3r:

mantognini:
I think 3.5 is a good trade off: it is widely available so it will bother only few people while providing us with more modern tools, simplifying the code (hopefully). Requiring 3.7 seems however a bit premature for now.

binary1248:
I just want to remind people of the origin of this discussion: We wanted to make use of "modern" CMake features to simplify integrating SFML into the user's own CMake configuration via target-dependency-based configuration.

From that perspective, raising the minimum version just so that we can do some internal clean up isn't worth it for me. At the end of it all, the user will not benefit from this that much, quite the opposite, if for some unfathomable reason they are running on a platform where they simply can't get hold of a newer CMake we are going in the other direction: making their lives harder than it needs to be.

Just to make the example more concrete:
I am running multiple servers (one of them being CI). They are all still running Debian Jessie, since I haven't come around to (or had the need to) upgrade them to Stretch since it came out 7 months ago. 7 months is also not that long if you are the kind of guy who plays it safe (which I hope most sysadmins are). After merging this PR, I would have to go and either perform a distribution upgrade or install a newer CMake from backports (which might cause other packages to also be upgraded). CI isn't the only thing running on the server, just one of I'd say 40 things. Would I be willing to take the risk upgrading just so I can build a bleeding edge version of SFML? Probably not. And at that point I would honestly be annoyed at the developers for making a change that does not benefit me in any way.

Yes... I myself am a developer in this context, and as a developer I must also consider that it isn't always about us. We develop this library for others to use. If we only piss them off due to decisions we make for maintenance reasons then that is a sure-fire way to convince people to start using another library.

Also, I'd not exaggerate the number of places where we would have to add version checking to incorporate newer features if they are available. Going out on a limb, I'd say that you can count them using 2 hands. This is still maintainable if you ask me. Everybody should know by now that I am a fan of modern OpenGL. If you look at SFML's graphics code, we check for versions and optional features all over the place. We support platforms providing only OpenGL 1.2.1 support and at the same time have support for geometry shaders. This involved a certain amount of effort, but if you ask me this is what makes the library what it is. Investing the effort to achieve this goal in the actual C++ code and skimping on the CMake configuration wouldn't leave a very convincing impression of the library if you ask me.

TL;DR: It's nice to make use of the newest features, but I wouldn't sacrifice compatibility for it.

Ceylo:
To balance a bit what has been said, don't forget that there are two sides to this story: SFML developers and SFML users.

Apart from balancing the wishes of each side, we may also consider having a lower requirement for SFML users than for SFML developers. This means a backward compatible FindSFML/SFMLConfig/whatever-is-used and a "modern" internal CMake usage. This sounds like a reasonable compromise. The main exception I'm aware of is users that use SFML as a sub-project of their CMake project. How does it sound to you?

@binary1248 Is the Debian Jessie server you're talking of the one used for SFML's CI ?

binary1248:
A Jessie server is used for the Buildbot master, Buildbot slave, and the XEN hypervisor for a Windows slave and the FreeBSD slave.

Ceylo:
Ok. So we could imagine, if you want to remain totally safe, that the Buildbot slave could run within the Xen hypervisor as a Debian Stretch, right?

binary1248:
Originally the slave would run in an isolated machine. The "problem" with that was that since it is a separate virtual machine you would have to dedicate hardware resources i.e. RAM to it. The memory ballooning in Xen still doesn't work very well, so if there isn't a need for dynamic RAM allocation then don't use it. In order to have a somewhat reliable environment for GCC to run in you would need to dedicate at least 512MB of RAM to the machine. Multiply this by the concurrency you want to achieve and you will be putting aside a few GB of RAM just to run builds for 1 hour per week. Not worth it for me, which is why I run the slave in the host now. Security isn't as much of an issue for me as much as reliability and stability is, this doesn't mean I don't take security seriously.

mantognini:
I guess it all boils down to this: do we think it's acceptable to have the user potentially install manually cmake on their system or not?

From my personal experience on Linux, I've always ended up installing some *non-critical* package manually such as dev tools. Why? Because no matter which distribution I've used, they are never enough up-to-date. Fortunately, it's easy to install programs in your HOME (or wherever you wish).

The story is different for critical packages (e.g. system-related components; packages you absolutely always want to have the latest version of) or for packages used for running the applications you're developing in order to make it easy to ship your programs. So, the versions of SFML's dependencies should be widely available and, generally speaking, system compatibility should be high.

But *building* SFML is another story IMO. I think it's okay to require something not 100% available in common Linux distros for that specific purpose, as long as it's easy to install manually, of course. Beside, not everybody has to build SFML themself; on such system with a standard package manager people should be able to rely on it to get SFML without even bothering with building it.

Bottomline is, asking a relatively recent version of Cmake (~2 years old) would be fine according to this criteria. But this is only my personal opinion... I just hope we decide quickly what is tolerable for SFML according to the majority and adapt code accordingly. :)

feliwir:
I agree with @mantognini here. We can totally expect people building/developing software being able to upgrade a hopelessly outdated build tool to a more recent version, that is easy accessible.

MarioLiebisch:
@feliwir @mantognini I dare to say you haven't had that situation with some tricky piece of software and its dependency hell. :D

Either way, I'd see it a bit differently and I'd aim for some "middle of the road" approach:

* Define the minimum CMake version as the one definitely required under all circumstances to build successfully. In this current case this would most likely be 3.1 or 3.5.1 or whatever. (In short: have all necessary features without excessive workarounds)
* If there are cases that require a more recent version, let's assume targeting MinGW with Codeblocks is broken before version 3.8, add a warning or error message rather than bumping minimum requirements to 3.8 for everyone.

FYI: Visual Studio 2017 deploys CMake 3.9.2 right now.

mantognini:
> I dare to say you haven't had that situation with some tricky piece of software and its dependency hell. :D

For the record, one Debian **7** wheezy, one just needs to install the `build-essential` package to get & compile successfully cmake 3.5, so we can't speak of dependency hell for our specific use case.

PR #769, opened 2015-01-08
Font Dirtyness Counter

FRex:
This solves the problem that appears when reloading sf::Font after sf::Text already built its' geometry - without this, the text would disappear or use wrong texture coords after the font reload.
This will also allow sf::Text (and other layouters written by users) to automagically know to rebuild their vertices when something forces sf::Font to flush all its' glyph pages.

The rare bug thread: http://en.sfml-dev.org/forums/index.php?topic=14019.0
The thread of first possible feature that might force flushing glyph pages: http://en.sfml-dev.org/forums/index.php?topic=17175

PR #1345, opened 2018-01-15
Fix for broken text when the font is reloaded.

Foaly:
Hi!
I took the liberty to write an update for this PR #769. There has been no activity in the last 2 years. I updated the implementaion to the decision in the forum discussion. https://en.sfml-dev.org/forums/index.php?topic=14019.30
Hopefully we can fix the bug soon!

### DIFF
--- a/include/SFML/Graphics/Text.hpp
+++ b/include/SFML/Graphics/Text.hpp
@@ -393,6 +393,7 @@ private:
     mutable VertexArray m_outlineVertices;    ///< Vertex array containing the outline geometry
     mutable FloatRect   m_bounds;             ///< Bounding rectangle of the text (in local coordinates)
     mutable bool        m_geometryNeedUpdate; ///< Does the geometry need to be recomputed?
+    mutable Uint64      m_fontTextureId;      ///< The font texture id
 };
 
 } // namespace sf
--- a/include/SFML/Graphics/Texture.hpp
+++ b/include/SFML/Graphics/Texture.hpp
@@ -35,10 +35,11 @@
 
 namespace sf
 {
-class Window;
+class InputStream;
 class RenderTarget;
 class RenderTexture;
-class InputStream;
+class Text;
+class Window;
 
 ////////////////////////////////////////////////////////////
 /// \brief Image living on the graphics card that can be used for drawing
@@ -584,6 +585,7 @@ public:
 
 private:
 
+    friend class Text;
     friend class RenderTexture;
     friend class RenderTarget;
 
--- a/src/SFML/Graphics/Text.cpp
+++ b/src/SFML/Graphics/Text.cpp
@@ -84,7 +84,8 @@ m_outlineThickness  (0),
 m_vertices          (Triangles),
 m_outlineVertices   (Triangles),
 m_bounds            (),
-m_geometryNeedUpdate(false)
+m_geometryNeedUpdate(false),
+m_fontTextureId     (0)
 {
 
 }
@@ -102,7 +103,8 @@ m_outlineThickness  (0),
 m_vertices          (Triangles),
 m_outlineVertices   (Triangles),
 m_bounds            (),
-m_geometryNeedUpdate(true)
+m_geometryNeedUpdate(true),
+m_fontTextureId     (0)
 {
 
 }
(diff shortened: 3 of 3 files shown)


include/SFML/Graphics/Text.hpp
```diff
@@ -393,6 +393,7 @@ class SFML_GRAPHICS_API Text : public Drawable, public Transformable
     mutable VertexArray m_outlineVertices;    ///< Vertex array containing the outline geometry
     mutable FloatRect   m_bounds;             ///< Bounding rectangle of the text (in local coordinates)
     mutable bool        m_geometryNeedUpdate; ///< Does the geometry need to be recomputed?
+    mutable Uint64      m_fontTextureId;      ///< The fonts texture id
```
binary1248: Comment looks weird, either add an apostrophe or leave the s away from fonts.

include/SFML/Graphics/Texture.hpp
```diff
@@ -35,10 +35,11 @@
 
 namespace sf
 {
-class Window;
+class InputStream;
 class RenderTarget;
 class RenderTexture;
-class InputStream;
+class Text;
+class Window;
```
binary1248: No need to reorder the declarations here... Add a single line instead of modifying 3.

src/SFML/Graphics/Text.cpp
```diff
@@ -346,10 +346,13 @@ void Text::draw(RenderTarget& target, RenderStates states) const
 ////////////////////////////////////////////////////////////
 void Text::ensureGeometryUpdate() const
 {
-    // Do nothing, if geometry has not changed
-    if (!m_geometryNeedUpdate)
+    // Do nothing, if geometry has not changed and the font texture has not changed
+    if (!m_geometryNeedUpdate && m_font->getTexture(m_characterSize).m_cacheId == m_fontTextureId)
```
binary1248: In certain situations, this will dereference a NULL pointer causing undefined behaviour. The check for m_font is already done further down. Either move it up or move this down.

src/SFML/Graphics/Text.cpp
```diff
@@ -346,10 +346,13 @@ void Text::draw(RenderTarget& target, RenderStates states) const
...
         return;
 
+    // Save the current fonts texture id
+    m_fontTextureId = m_font->getTexture(m_characterSize).m_cacheId;
```
binary1248: Same as above.

include/SFML/Graphics/Texture.hpp
Foaly: I though having them alphabetically made more sense then random.

Foaly:
By the way if someone wants to test I've been using exploiters old testo code

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode(800, 600), "Example");

    bool broken = false;

    const std::string filename("a font");

    sf::Font font;
    font.loadFromFile(filename);

    sf::Text text("font is broken", font, 30u);

    // run the program as long as the window is open
    while (window.isOpen())
    {
        // check all the windows events
        sf::Event event;
        {
            // "close requested" event: we close the window
                window.close();

            else if (event.type == sf::Event::KeyReleased) {
                // close if escape key was pressed
                if (event.key.code == sf::Keyboard::Escape) {
                    window.close();
                }

                if (!broken)
                {
                    broken = true;
                    font.loadFromFile(filename);
                    //text.setFont(font);
                    //text.setCharacterSize(30u);
                    //text.setString("font is broken");
                }
            }
        }

        // clear the window to black
        window.clear();

        window.draw(text);

        // display the windows content
        window.display();
    }
    return 0;
}
```

mantognini:
Works like a charm on mac. 👍 

src/SFML/Graphics/Text.cpp
binary1248: Just noticed that `m_fontTextureId` isn't initialized to a value prior to this line, meaning that in the comparison above, it would contain an undefined value.

include/SFML/Graphics/Texture.hpp
binary1248: 👍

src/SFML/Graphics/Text.cpp
Foaly: 👍 fixed

Foaly:
I think this can be merged now.

PR #1209, opened 2017-03-18
Updated the binary libs from exlibs/libs-ios

sol-prog:
Updated the binary libs from exlibs/libs-ios in order to work with Xcode 8 and iOS 10.

PR #1347, opened 2018-01-24
Don't need to find vorbisfile or vorbisenc on iOS

JonnyPtn:
As mentioned in #1209 the vorbis lib on iOS is now all inclusive, however SFML cmake configuration still looks for vorbisfile and vorbisenc, so the build fails if they aren't present


LaurentGomila:
When did that change? What about still supporting older versions of iOS? Did they also change that for other OSes?

MarioLiebisch:
My big question would possibly be whether this change is compile/link time only or run-time as well? Considering how Apple regularly deprecates older versions and forces upgrades I'm not sure how important keeping the old code would be.

JonnyPtn:
> When did that change?

When #1209 was merged

> What about still supporting older versions of iOS?

This PR has nothing to do with iOS versions, it just fixes the project to build. I don't know what effect #1209 would have on iOS version compatibility, although I still suspect not much, if any

> Did they also change that for other OSes?

Again, If you're talking about #1209, then no, as far as I'm aware other OS's still have separate vorbis files which is why this PR only affects iOS builds

> My big question would possibly be whether this change is compile/link time only or run-time as well? 

It's CMake configure time.

> Considering how Apple regularly deprecates older versions and forces upgrades I'm not sure how important keeping the old code would be.

Apple have nothing to do with this. If I understand correctly the reason #1209 changed this is because it's using a more recent version of vorbis than other OS extlibs.

LaurentGomila:
> When #1209 was merged

I mean, when did that change in the library, not "when did we apply the change in SFML".

> This PR has nothing to do with iOS versions

My bad, I thought the library was provided by the OS, but apparently we build it ourselves.

> as far as I'm aware other OS's still have separate vorbis files which is why this PR only affects iOS builds

Sounds weird that they did such a change only for iOS libraries. Would be interesting to know why they did it.

JonnyPtn:
No idea when vorbis changed, struggling to find any information on it... @sol-prog or @eXpl0it3r may be able to shed some light

I agree it's weird, and ideally the libs should be the same version/format as on other platforms.

JonnyPtn:
Any more thoughts on this? If you'd prefer me to rebuild the vorbis libraries I can do that?

This is a real blocker for any other iOS testing

eXpl0it3r:
Somehow I thought they moved on to merging them in general, but looking at the VS solution in their repo, it doesn't look to be the case.

Personally, I don't mind if the library list is different for iOS compared to other OS, as long as it works. I mean if you have the time, you could see if you build them for iOS whether they get built or if @sol-prog did some magic to include them into one library.

If you don't want to investigate this, I'll merge this PR.

JonnyPtn:
I definitely think it should match other platforms, but right now it doesn't seem clear which version is correct/best. Personally I'd say merge this for the time being as it only affects iOS and is required to get it building, then we can open a separate issue to discuss and fix up the vorbis extlibs inconsistencies?

eXpl0it3r:
Sounds fine to me, it's already this way except for the CMake-thingy being broken, so lets first fix the broken thingy and then see if/how the rest needs to be fixed. 🙂

eXpl0it3r:
Merged in d6c6345d4c4f3d9345f0a2e598cc49bdae819f06

PR #1348, opened 2018-01-25
Install extlibs on iOS if needed

JonnyPtn:
Audio extlibs specifically.

I also removed a seemingly superfluous check for `SFML_OS_IOS` as it's already in a block which checks that


JonnyPtn:
I'll try and rebase this later today, presuming there's no issues?

eXpl0it3r:
CMakeLists.txt was changed, thus it created a conflict. Would be great if you could resolve it.

eXpl0it3r:
Merged in 66369645b545d5ace07b7d58b65ca97b3f43a740

PR #1350, opened 2018-01-25
Android: Removed custom toolchain + NDK updates

MarioLiebisch:
This PR provides two changes:

* The previous custom CMake toolchain file for Android is dropped in favor of CMake's new built-in toolchain for this (CMake >3.7.2).

* Includes AlexAUT's fixes for the latest NDK releases from PR #1287.

This makes building SFML for Android a lot simpler and more straight forward, working almost as smooth as other platforms.

To configure your build directory, all you have to do is defining just a few variables the first time you invoke CMake.

**Required Variables**

* `CMAKE_SYSTEM_NAME` must be `Android`, so CMake knows we actually want to cross-compile.
* `CMAKE_ANDROID_NDK` must point to the NDK's installation directory, e.g. `/usr/android/ndk` or `c:/android/ndk`.

**Recommended Variables**
* `CMAKE_ANDROID_STL_TYPE` defines the STL implementation to be used. You should use `c++_shared`, although others might work.

**Optional Variables**
* `CMAKE_SYSTEM_VERSION` can be set to pick a specific SDK version other than the latest.
* `CMAKE_ANDROID_ARCH_ABI` defines the target architecture and ABI, for example `armeabi` or `armeabi-v7a`.

Based on your system, you might want to enforce a specific generator to prevent issues, e.g. using `MinGW Makefiles`.


MarioLiebisch:
Just to avoid confusion, the Android CI builds are supposed to fail, since variable names have changed.

eXpl0it3r:
Don't forget to provide an updated [config for the CI](https://github.com/SFML/SFML-Buildbot), so it will build fine again after this is merged.

MarioLiebisch:
Now if I'd know how to do so specifically and how to actually test it… Setup should be similar trivial though, something like this (not sure it should use extra variables for this):

    if 'android' in options:
        build_target += '-DCMAKE_SYSTEM_NAME=Android'
        build_target += '-DCMAKE_ANDROID_ARCH_ABI=armeabi-v7a'
        build_target += '-DCMAKE_ANDROID_STL_TYPE=c++_shared'
        build_target += '-DCMAKE_ANDROID_NDK=...'

eXpl0it3r:

eXpl0it3r:
Android build fails. @MarioLiebisch, can you look into what needs to be changed and tell @binary1248 if there are some CI changes needed or @mantognini if there are some build changes needed?

AlexAUT:
Works fine under linux.

But I could not get it working with the CMake-Gui, not sure if this is a important concern, because the user has to use the command line to compile SFML anyways?

MarioLiebisch:
> can you look into what needs to be changed

[See my comment above.](https://github.com/SFML/SFML/pull/1350#issuecomment-360488903) It just needs different defines and no longer uses the custom toolchain file.

@AlexAUT Can you be more specific? What exactly fails? You'll have to define the variables I've mentioned, but then it should work even with the GUI version.

eXpl0it3r:
Point is, I don't run the CI, so you don't have to tell me, but you need to tell @binary1248, at best make a pull request on the config repo. @AlexAUT has already created an issue there.

MarioLiebisch:
As mentioned above, I can't. I have no idea how the configuration works or how to even test it so it would be at best what I provided in the comment.

AlexAUT:
My issue already contains the fixes to the CI, with one small unsolved problem, when binary comments, I'm gonna do the PR (just two lines of code)

CMake-Gui: 
Even when I'm defining the variables in the GUI, it stills builds with my local x86 compiler, so the variables have no effect. (Gonna look into it again later today)

AlexAUT:
**Update:**
It does work with the CMake-Gui, but you have to define the variables before the first "configure".

binary1248:
CI is fixed.

texus:
These changes added an ANDROID_ABI option (https://github.com/SFML/SFML/pull/1350/files#diff-af3b638bc2a3e6c650974192a53c7291R133), but is it still used anywhere?

MarioLiebisch:
Oh, no, that's obviously some leftover – not 100% sure how it ended in there. The correct variable is `CMAKE_ANDROID_ARCH_ABI` and I'm not sure it can be set that late.

texus:
Running ndk-build in the examples/android folder now gives the following error for me:
```
Android NDK: jni/Android.mk: Cannot find module with tag 'sfml' in import path
Android NDK: Are you sure your NDK_MODULE_PATH variable is properly defined ?
```

It works when I change the last line to `$(call import-module, third_party/sfml)` (the "third_party/" was added) and made a similar change to src/SFML/Android.mk, but that isn't really a good option because it probably won't work when CMAKE_INSTALL_PREFIX does not have its default value.

It seems like it only looks inside the /ndk-path/sources directory for sfml and not in /ndk-path/sources/third_party folder.

MarioLiebisch:
Yeah, now I've noticed what's going on here. @eXpl0it3r merged the  PR before I could push the WIP changes including the updated example. Will have to see how I can put this straight on the weekend.

PR #1351, opened 2018-01-25
Fix Opus audio sample rate to 48kHz

SergioRAgostinho:
Following the request in https://github.com/SFML/SFML/pull/1259#issuecomment-360431188


eXpl0it3r:
Merged in 81613041c3579385971c99943d2b1989e45f424f

PR #1353, opened 2018-01-25
Correct docstring in SoundFileReader::seek

SergioRAgostinho:
Per request https://github.com/SFML/SFML/pull/1259#discussion_r163906568

### DIFF
--- a/include/SFML/Audio/InputSoundFile.hpp
+++ b/include/SFML/Audio/InputSoundFile.hpp
@@ -159,8 +159,9 @@ public:
     /// other overload.
     ///
     /// The sample offset takes the channels into account.
-    /// Offsets can be calculated like this:
-    /// `sampleNumber * sampleRate * channelCount`
+    /// If you have a time offset instead, you can easily find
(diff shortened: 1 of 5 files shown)


eXpl0it3r:
Look at the commit (c78c8106dad49dd458aa3624df06d2a989f3d0ed) when that was added, seems like there's another place that needs to be changed.

Since it's your original commit, do you agree with this change, @Foaly?

SergioRAgostinho:
I'll grep and replace all occurrences upon confirmation.

LaurentGomila:
Tiny improvement: I would make it more explicit, with something like "If you have a time offset instead, you can easily find the corresponding sample offset with the following formula:".

Foaly:
Oh yes absolutly! I appologize for spreading wrong information 😄 But looking at the commit you linked there is actually 5 places where it has to be changed.

eXpl0it3r:
Merged in 6d4e870e320be743cb7699de7e964bb308dea05a

PR #925, opened 2015-07-15
Outline vertices are only updated if there is an outline.

tylerellis21:
Improved the logic of the shape class so that the outline vertices are only updated if they exist.

PR #1356, opened 2018-01-26
Generate shape outline vertices only if necessary.

Foaly:
Improved the logic of the shape class so that the outline vertices are only updated if they exist.

Original, now abandoned #925


Foaly:
This is a really small fix. @LaurentGomila @Bromeon would one of you guys mind reviewing? Thanks :)

MarioLiebisch:
Don't forget to rebase it onto the current `master` branch.

eXpl0it3r:
> Don't forget to rebase it onto the current `master` branch.

Doesn't really matter as long as there's no conflict. It just won't auto close. :wink:

Foaly:
Thanks! I rebased to make it easier :)

PR #1357, opened 2018-01-26
Fixed Win32 modifier keys handling

hobby8:
Fixed Win32 modifier keys handling so that `alt/control/shift/system` values in `Event::KeyEvent` reflect their status at the time of `KeyPressed/KeyReleased`.

Does not affect `Keyboard::isKeyPressed()`, which can still be used to access the keyboard state directly.


LaurentGomila:
Indeed, looks more correct. Out of curiosity, what made you work on this fix? Did you experience problems?

hobby8:
Indeed, with a debugging feature that introduced occasional lags, Shift became out of sync with the actual keystrokes, that was pretty funny.

eXpl0it3r:
Merged in 1062e95e39996356dde9794ee45d392afa47d1ec

PR #1359, opened 2018-01-29
Back to the future

Foaly:
I updated the copyright year to 2018. 
Funny enough all the cursor files where still at 2016. Probably because it was in discussion for so long and got overlooked while merging 😄 
Another thought I had, was that it is no longer accurate to name one (or sometimes two or three people) as the author/copy right holder. It should be updated to something like "the SFML team" or "the SFML project".


eXpl0it3r:
Merged in 5d7843c58a850b4941d7d9794d6fede2bbc02833

Issue #1313, opened 2017-10-30
sf::Text incorrect Outline


achpile:
I just found that outline calculation is incorrect for sf::Text. Screenshots and example code are in forum thread here: https://en.sfml-dev.org/forums/index.php?topic=22690

If outlineThickness is positive - outline is drawn like it is positive and negative at the same time.
If it's negative - result is completely incorrect.

rbrich:
Hi, this looks to be a documented behaviour in SFML:

`Text::setColor`
> Setting the fill color to a transparent color with an outline
> will cause the outline to be displayed in the fill area of the text.

`Text::setOutlineThickness`
> Be aware that using a negative value for the outline
> thickness will cause distorted rendering.

Looking at the FreeType API, sf:Text uses `FT_Stroker_Set` to set the radius (=thickness), and then  `FT_Glyph_Stroke` to draw the outline. This draws both inner and outer parts of the outline ("borders"). Both using the same radius. Negative value of radius seems to be just incorrect (not documented at all).

To draw only the outer (or inner) border, there is another function: `FT_Glyph_StrokeBorder`. This would have to be used in SFML to support alpha-blended text like in your example.

Foaly:
Hi! 
Eventhough this is documented behaviour I think there should be some changes.

I don't see any reason why the outline should draw within the bounds of a character. I mean its called outline afterall. So its probably a good idea to switch to `FT_Glyph_StrokeBorder` like @rbrich suggested. 

Also in my opinion a negative outlineThickness doesn't make much sense. Personally I would clamp the `outlineThickness` to 0.f and document that. I know some devs here don't like things like that. But I think there should be at least an assertion saying that negative values will lead to distorted rendering.
Another option would be to draw `FT_Glyph_StrokeBorder` inner border for negative thicknesses.
[Docs](https://www.freetype.org/freetype2/docs/reference/ft2-glyph_stroker.html#FT_Glyph_StrokeBorder)

MarioLiebisch:
Why not leave the option to whoever is using it? If the outline is positive, use `FT_Glyph_StrokeBorder`, if it's negative, use the current behavior (with an absolute value)?

Foaly:
Actually thinking about it again I wouldn't clamp the values to positive, but instead fill negative values with some actual functionality.
When we change the behaviour to only draw an outline along the outside for positive values, we can basically pick what we want do for negative values. I think there are two options: 
[...]

Issue #1233, opened 2017-05-22
Memory leak when sf::Text::setOutlineThickness is used


Chringo:
Seems to be that [Font.cpp - line 561](https://github.com/SFML/SFML/blob/master/src/SFML/Graphics/Font.cpp#L561) creates a leak as indicated by this answer on StackOverflow:
https://stackoverflow.com/a/44074908/5044558
Followed by a discussion on the sfml-dev forums:
https://en.sfml-dev.org/forums/index.php?topic=21996.0

Working with SFML 2.4.2 on Windows 7 64-bit version, I've noticed an issue with sf::Text::setOutlineThickness(float). Once it is used in the program, except for default value 0, crtdbg dumps a memory leak of various sizes of bytes but always the same amount. I believe this is related to the size of the string, if the text gets drawn, and if the parameter of setOutlineThickness is accepted, demonstrated here:
```
/// Initial set-up
sf::Text test;
test.setString("A");
// ... Set charactersize, font, fillcolor, etc ...
test.setOutlineThickness(1);
test.setOutlineColor(sf::Color::Black);

/// Make a drawcall for test later in the program
void Game::draw(sf::RenderTarget & target, sf::RenderStates states) const
{
    target.draw(test, states);
}
```
If the line in Font.cpp is modified to,
`FT_Glyph_Stroke(&glyphDesc, stroker, true);`
Then there are no memory leaks and the program can still display the text without problems. This was tested on SFML's source code from the master branch and was built in Win32 debug and release.

agordeevw:
The leak seems to originate exactly from this line, because if argument is set to "false", then the memory allocated to glyphDesc before FT_Glyph_Stroke(...) call will be hanging, because after the call glyphDesc is redirected to a newly allocated memory, which stores modified glyph.
Slightly more words on the forums
https://en.sfml-dev.org/forums/index.php?topic=21996.msg156356#msg156356

Foaly:
Funny I was just looking into this as well. Related to #1313 I will try to fix them together.

eXpl0it3r:
Fixed with #1360 and merged in d972216c57e2fe28fd36dad55c0772078640e971

PR #1360, opened 2018-01-29
Fix memory leak in sf::Text.

Foaly:
Since #1313 will probably take a bit longer, I am sending this fix for #1233 now, so there is no unnecessary dependecy.


eXpl0it3r:
Merged in d972216c57e2fe28fd36dad55c0772078640e971

PR #1362, opened 2018-01-29
Return correct key code for delete/backspace on android.

### LINKED ISSUE #1309: Android sends Delete event instead of Backspace
SFML maps the [AKEYCODE_DEL](https://developer.android.com/ndk/reference/group___input.html#gga39fca1837c5ce7715cbf571669660c13acd013221b457d98975dc47e49817e28a) event to Keyboard::Delete ([WindowImplAndroid.cpp#L623](https://github.com/SFML/SFML/blob/013d053277c980946bc7761a2a088f1cbb788f8c/src/SFML/Window/Android/WindowImplAndroid.cpp#L623)) while AKEYCODE_DEL is actually the backspace key. The real delete key is [AKEYCODE_FORWARD_DEL](https://developer.android.com/ndk/reference/group___input.html#gga39fca1837c5ce7715cbf571669660c13a9516bc190d37fea27e07ddab0c607b51).

It was mentioned before in [this forum post](https://en.sfml-dev.org/forums/index.php?topic=13716.msg105902#msg105902), but it seems like it has been overlooked.

Foaly:
This fixes #1309.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,8 +104,8 @@ endif()
 # Android options
 if(SFML_OS_ANDROID)
     # make sure there's the android library available
-    if (CMAKE_ANDROID_API LESS 9)
-        message(FATAL_ERROR "Android API level (${CMAKE_ANDROID_API}) must be equal or greater than 9.")
+    if (CMAKE_ANDROID_API LESS 14)
+        message(FATAL_ERROR "Android API level (${CMAKE_ANDROID_API}) must be equal or greater than 14.")
     endif()
 
     # CMake doesn't support defining the STL to be used with Nsight Tegra, so warn the user
--- a/examples/android/AndroidManifest.xml
+++ b/examples/android/AndroidManifest.xml
@@ -7,11 +7,11 @@
 
     <uses-feature android:glEsVersion="0x00010001" />
 
-    <uses-sdk android:minSdkVersion="9"
+    <uses-sdk android:minSdkVersion="14"
               android:targetSdkVersion="19" />
 
     <uses-permission android:name="android.permission.VIBRATE" />
-    
+
     <application android:label="@string/app_name"
                  android:icon="@drawable/sfml_logo"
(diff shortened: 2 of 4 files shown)


texus:
This is a good example of why even trivial fixes need to be tested. The API has to be 13 or higher or it won't compile. The ANDROID_API_MIN currently still has a default value of 9, where AKEYCODE_FORWARD_DEL apparently didn't exist yet. I think the minimum version can be increased without any issues, even API 13 (android 3.2) is more than 7 years old already.
I couldn't test the delete key, but at least I can confirm that the backspace key works correctly now.

Foaly:
Funny somehow I had the gut feeling it would be a good idea to have this tested 😄
According to [stats in the andoid dashboard](https://developer.android.com/about/dashboards/index.html) the version 3.2 is not even listed anymore. Any significant user numbers don't start until 4.1.x. Unless anybody from the andoid devs (maybe @MarioLiebisch ) raises any concerns I will increase the API version.

Could you test the new version?

Foaly:
14 it is then!
I've also updated the references in the android example that I could find. I hope I didn't miss anything else.

texus:
Works without problems now. I found a keyboard with a delete key so I can confirm that both backspace and delete works.

Foaly:
Thats awesome! Thank you very much for testing! I guess if @MarioLiebisch closes his review this can be merged :)

MarioLiebisch:
Yeah, changes are fine, but not sure we should wait for #1350 with this. You'll basically need both changes to properly compile without this bug (and others).

Foaly:
I think this is ready for merging now! :)

eXpl0it3r:
Small conflict to resolve since the other Android bit has been merged. 🙂

Foaly:
Ah I see it now :D Updated!

eXpl0it3r:
Merged in 186294607f32c4ec8fcb32104af8516bed63e1da

PR #1365, opened 2018-02-05
Examples: Set startup folder for debugging in Visual Studio

MarioLiebisch:
A very trivial change that will allow direct running/debugging of the examples in Visual Studio using the proper working directories, so all resources can be found.

This requires CMake 3.8 or newer, but should be silently ignored in older versions (can't test this right now).

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -217,6 +217,9 @@ macro(sfml_add_example target)
     # set the target flags to use the appropriate C++ standard library
     sfml_set_stdlib(${target})
 
+    # set the Visual Studio startup path for debugging
+    set_target_properties(${target} PROPERTIES VS_DEBUGGER_WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+
     # link the target to its SFML dependencies
     if(THIS_DEPENDS)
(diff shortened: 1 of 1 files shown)


cmake/Macros.cmake
```diff
@@ -182,6 +182,9 @@ macro(sfml_add_example target)
     # set the target's folder (for IDEs that support it, e.g. Visual Studio)
     set_target_properties(${target} PROPERTIES FOLDER "Examples")
 
+    # set the Visual Studio startup path for debugging
+    set_target_properties(${target} PROPERTIES VS_DEBUGGER_WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
```
LaurentGomila: `${CMAKE_CURRENT_SOURCE_DIR}`, are you sure?

MarioLiebisch: Considering it's called from the examples' source dir every time, yes. Don't think there should or needs to be a new parameter to pass the path. Or which variable had you in mind?

eXpl0it3r:
@MarioLiebisch can your resolve the merge conflict?

MarioLiebisch:
Didn't notice the other comment; done.

PR #1366, opened 2018-02-06
Letter spacing bugfix

oomek:
Ties letter spacing factors to the font size.

### DIFF
--- a/include/SFML/Graphics/Text.hpp
+++ b/include/SFML/Graphics/Text.hpp
@@ -161,13 +161,15 @@ public:
     ////////////////////////////////////////////////////////////
     /// \brief Set the letter spacing factor
     ///
+    /// The default spacing between letters is defined by the font.
     /// This factor doesn't directly apply to the existing
     /// spacing between each character, it rather adds a fixed
     /// space between them which is calculated from the font
     /// metrics and the character size.
     /// Note that factors below 1 (including negative numbers) bring
-    /// characters closer to each other
-    /// 
+    /// characters closer to each other.
+    /// By default the letter spacing factor is 1.
+    ///
     /// \param spacing New letter spacing factor
     ///
     /// \see getLetterSpacing
--- a/include/SFML/Graphics/Text.hpp
+++ b/include/SFML/Graphics/Text.hpp
@@ -161,10 +161,13 @@ public:
     ////////////////////////////////////////////////////////////
     /// \brief Set the letter spacing factor
     ///
-    /// The default spacing between letters is defined by the font.
-    /// This method enables you to set a factor to the spacing
-    /// between letters. By default the letter spacing factor is 1.
-    ///
+    /// This factor doesn't directly apply to the existing
+    /// spacing between each character, it rather adds a fixed
+    /// space between them which is calculated from the font
+    /// metrics and the character size.
+    /// Note that factors below 1 (including negative numbers) bring
+    /// characters closer to each other
+    /// 
     /// \param spacing New letter spacing factor
     ///
     /// \see getLetterSpacing
--- a/include/SFML/Graphics/Text.hpp
+++ b/include/SFML/Graphics/Text.hpp
@@ -149,9 +149,9 @@ public:
     ///
     /// The default spacing between lines is defined by the font.
     /// This method enables you to set a factor for the spacing
-    /// between lines. By default the line spacing offset is 1.
+    /// between lines. By default the line spacing factor is 1.
     ///
-    /// \param spacing New line spacing factor
+    /// \param spacingFactor New line spacing factor
     ///
     /// \see getLineSpacing
     ///
@@ -170,7 +170,7 @@ public:
     /// characters closer to each other.
     /// By default the letter spacing factor is 1.
     ///
-    /// \param spacing New letter spacing factor
+    /// \param spacingFactor New letter spacing factor
     ///
     /// \see getLetterSpacing
     ///
--- a/src/SFML/Graphics/Text.cpp
+++ b/src/SFML/Graphics/Text.cpp
@@ -318,8 +318,8 @@ Vector2f Text::findCharacterPos(std::size_t index) const
     // Precompute the variables needed by the algorithm
     bool  isBold          = m_style & Bold;
     float whitespaceWidth = m_font->getGlyph(L' ', m_characterSize, isBold).advance;
-    float letterSpacing   = ( whitespaceWidth / 3.0 ) * ( m_letterSpacingFactor - 1.0 );
-    whitespaceWidth += letterSpacing;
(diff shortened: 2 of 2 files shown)


src/SFML/Graphics/Text.cpp
```diff
@@ -317,7 +317,9 @@ Vector2f Text::findCharacterPos(std::size_t index) const
 
     // Precompute the variables needed by the algorithm
     bool  isBold          = m_style & Bold;
-    float whitespaceWidth = m_font->getGlyph(L' ', m_characterSize, isBold).advance * m_letterSpacingFactor;
+    float whitespaceWidth = m_font->getGlyph(L' ', m_characterSize, isBold).advance;
+    float letterSpacing   = ( whitespaceWidth / 3.0 ) * ( m_letterSpacingFactor - 1.0 );
```
LaurentGomila: Minor comment: use floats, not doubles (3.f, 1.f, etc.).

src/SFML/Graphics/Text.cpp
```diff
@@ -425,7 +427,9 @@ void Text::ensureGeometryUpdate() const
     float strikeThroughOffset = xBounds.top + xBounds.height / 2.f;
 
     // Precompute the variables needed by the algorithm
-    float whitespaceWidth = m_font->getGlyph(L' ', m_characterSize, isBold).advance * m_letterSpacingFactor;
+    float whitespaceWidth = m_font->getGlyph(L' ', m_characterSize, isBold).advance;
+    float letterSpacing   = ( whitespaceWidth / 3.0 ) * ( m_letterSpacingFactor - 1.0 );
+    whitespaceWidth += letterSpacing;
```
Foaly: Minor code style: the = should be aligned with the ones surrounding it

Foaly:
Great! Besides the documentation that Laurent mentioned I think the commit message should be changed to something more meaningful. Maybe: Fix letter spacing factors (now tied to the font size). Or something like that.

oomek:
I’m at the hospital until evening UK time. Sorry for lagging.

Foaly:
No worries! Take your time. Health comes first :)

oomek:
I'm a Github noob, as you have probably noticed. Is it acceptable now?

oomek:
@LaurentGomila I'm not sure what documentation are you referring to. if it's the comment block in the Text.hpp it's descriptive enough I think.

LaurentGomila:
I'd replace this

```cpp
/// This method enables you to set a factor to the spacing
/// between letters.
```
(which is really really bad, but so as many other comments in SFML 😁 )

With something like this

```cpp
/// This factor doesn't directly apply to the existing
/// spacing between each character, it rather adds a fixed
/// space between them which is calculated from the font
/// metrics and the character size.
/// Note that factors below 1 (including negative numbers) bring
/// characters closer to each other
/// 
```

LaurentGomila:
I didn't mean to replace the whole comment, only the part that I posted above. Please keep the indication of default value, and we can also keep the first sentence.

oomek:
Thank you for being patient. It's my first pull request done in git. I used the website to publish my code before.

oomek:
And besides, I'm not a native English speaker, so I chose the safest way :)

Foaly:
I know this has the potential to blow up the discussion, but reading over this again, wouldn't it be a better design to set the start of the additional factor at 0 intead of 1. So it would be a factor in percent or negative percent instead of somewhat randomly starting for 1 as no additional offset.

include/SFML/Graphics/Text.hpp
```diff
@@ -162,8 +162,13 @@ class SFML_GRAPHICS_API Text : public Drawable, public Transformable
     /// \brief Set the letter spacing factor
     ///
     /// The default spacing between letters is defined by the font.
-    /// This method enables you to set a factor to the spacing
-    /// between letters. By default the letter spacing factor is 1.
+    /// This factor doesn't directly apply to the existing
+    /// spacing between each character, it rather adds a fixed
+    /// space between them which is calculated from the font
+    /// metrics and the character size.
+    /// Note that factors below 1 (including negative numbers) bring
```
Foaly: Well this is not entirely correct I would say. My proposal:
```cpp
/// This factor doesn't directly apply to the existing
/// spacing between each character. It rather uses a fixed
/// space which is calculated from the fonts whitespace 
/// and the character size. This spacing is scaled by the 
/// `spacingFactor` and added to the default spacing between 
/// two characters.
```

include/SFML/Graphics/Text.hpp
```diff
@@ -162,8 +162,13 @@ class SFML_GRAPHICS_API Text : public Drawable, public Transformable
...
+    /// characters closer to each other.
+    /// By default the letter spacing factor is 1.
     ///
     /// \param spacing New letter spacing factor
```
Foaly: Parameter name is: spacingFactor

include/SFML/Graphics/Text.hpp
LaurentGomila: > this is not entirely correct

Which part is incorrect?

> My proposal

Well, no need to describe all implementation details. Users don't need to know that.

LaurentGomila:
> wouldn't it be a better design to set the start of the additional factor at 0 intead of 1

Since it's a factor, it makes sense to start at 1. And it remains consistent with the line spacing factor. If we go too far in the direction of "this is an offset, not a factor" we'll have to revert all that stuff to use actual offsets in public interface ;)

oomek:
Starting at 1.0 makes sense as going below 0.0 results in letters overlaping each other. We don’t need factoring in (-n,n) range but rather (0,n)

Foaly:
> ... use actual offset ...
You know that is the secret plan all along 😉

The point I wanted to add was that the factor affects that fixed spacing and not the "default" spacing. But yeah I guess it doesn't matter to much.

oomek:
Soooo, could we finally close that chapter please? I’m having an impression that we are overthinking it a little bit.

Foaly:
If @LaurentGomila is fine with it, it's ok from my side.

LaurentGomila:
> The point I wanted to add was that the factor affects that fixed spacing and not the "default" spacing.

I think it is important, that's why I say it in my version of the comment. My point was: users need to know how it behaves, not how it is precisely calculated.

> Soooo, could we finally close that chapter please? I’m having an impression that we are overthinking it a little bit.

Proper documentation is not "overthinking". It's probably the most important thing, now that the technical issues have been sorted out. As long as the PR is not blocked in endless arguments, there's no hurry, let's not be afraid to discuss even the  most minor details 😉 

> If @LaurentGomila is fine with it, it's ok from my side.

If the current code has been tested and still works as expected, I'm totally fine with it.

eXpl0it3r:
Squashed and merged in deeb3a95f3ff46cc350cbabb08802cf6fb9ddac6

Issue #1284, opened 2017-09-17
Borderless (sf::Style::None) acts as Fullscreen if resolution is max


naezith:
**sf::Style::None** works at exclusive fullscreen mode as in **sf::Style::Fullscreen** when resolution is at max. It works as windowed at smaller resolutions.

Skype's screenshare is showing black screen if it's borderless & max resolution. 
Screenshot programs also shows black, or captures a very very old frame.

LaurentGomila:
Please use the forum to discuss issues. This tracker is for confirmed features and bugs.

And please provide a more complete description of your problem and environment (minimal code, OS, WM, version of SFML, etc.).

naezith:
@LaurentGomila Please don't close it, it's a known bug. Many of my players reported this, I also experience it. [It also exists in the forum](https://en.sfml-dev.org/forums/index.php?topic=21802.msg154680). I didn't see a issue here so I created one.


`#include <SFML/Graphics.hpp>

int main() {
    sf::RenderWindow window;
    window.create(sf::VideoMode::getDesktopMode(), "Borderless Window", sf::Style::Fullscreen);

    while(window.isOpen()) {
        sf::Event event;
        while(window.pollEvent(event)) if(event.type == sf::Event::Closed)  window.close();

        window.clear(sf::Color::Red);
        window.display();
    }

    return 0;
}
`

Jaex:
I'm developer of the screen capture software [ShareX](https://getsharex.com) and I was testing his game [Naezith](https://naezith.com) on Windows 10. To be able to take screen capture or record videos from his game I needed to change video settings from fullscreen to borderless. But I noticed when resolution is lower than my monitor size than it works as expected. But as soon as I change resolution to my monitor size then it switches to exclusive fullscreen mode automatically and it does not let me capture screenshots from the game. This means when game window resolution is same as my monitor then fullscreen `sf::Style::Fullscreen` and borderless `sf::Style::None` have no difference.

Veltas:
This appears to happen for me as well, notionwm window manager on Linux. Code used:

```c++
#include <SFML/Graphics.hpp>

int main()
{
  // Try with your fullscreen resolution, and a lower one that's also supported e.g. 800x600
  sf::RenderWindow window(sf::VideoMode(800, 600), "SFML Window", sf::Style::None); // Borderless fullscreen if dimensions match resolution
[...]

PR #1367, opened 2018-02-08
Prevent switching to fullscreen, if window has sf::Style::None and desktop resolution on windows.

Foaly:
This is a fix for #1284. It would be great if you could test it @naezith @Jaez @marukrap


src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -193,11 +193,7 @@ m_cursorGrabbed   (m_fullscreen)
 
     // Choose the window style according to the Style parameter
     DWORD win32Style = WS_VISIBLE;
-    if (style == Style::None)
-    {
-        win32Style |= WS_POPUP;
-    }
-    else
+    if (style != Style::None)
```
LaurentGomila: No need to test explicitly against Style::None, in this case none of the tests below will succeed and the `win32Style` variable will correctly stay at its initial value.

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -216,6 +212,12 @@ m_cursorGrabbed   (m_fullscreen)
     // Create the window
     m_handle = CreateWindowW(className, title.toWideString().c_str(), win32Style, left, top, width, height, NULL, NULL, GetModuleHandle(NULL), this);
 
+    // Remove all decorations if requested
+    if (style == Style::None)
+    {
+        SetWindowLong(m_handle, GWL_STYLE, 0);
```
LaurentGomila: How did you find this solution? Is it documented?

MarioLiebisch: This basically sets the window to [`WS_OVERLAPPED`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms632600(v=vs.85).aspx):

> The window is an overlapped window. An overlapped window has a title bar and a border. Same as the **WS_TILED** style.

This can't modify all styles after the window is created, so we're potentially just reverting one of the flags rather than everything and this works just because...

LaurentGomila: How is it different from the style argument that we pass to `CreateWindowW`? Should be the same thing, and we already pass 0 to `CreateWindowW` (we just add `WS_VISIBLE` which has nothing to do with decorations).

MarioLiebisch: At window creation all components of `WS_OVERLAPPED` can be applied, so there might be happening more than when changing the flags only.

LaurentGomila: The MSDN documentation of the `dwStyle` argument of `CreateWindow` and of the `GWL_STYLE` option of `SetWindowLong` both directly point to the window styles page, so (if the doc is correct) that should be the exact same thing.

Foaly:
Hmm no it does not seem to be documented. I took it from [here](https://stackoverflow.com/questions/7442939/opening-a-window-that-has-no-title-bar-with-win32) and assumed that 0 meant no state. I didn't double check with the windows documentation. This fixes the bug though. Tested on windows 8.1 by the way.

LaurentGomila:
If we can't find a documented source, I'm afraid we'll need more tests before we can validate it.

hobby8:
I suggest to examine with [Spy++](https://docs.microsoft.com/en-us/visualstudio/debugger/introducing-spy-increment) the window styles of a few applications that get this right on the problem OS.
If doing the same does not help, you may want to try a tool such as [API Monitor](http://www.rohitab.com/apimonitor) (enable 'Windows Application UI Development/Windows and Messages/Windows' and maybe also 'Graphics/Legacy Graphics').

oomek:
I'm actually using sf::Style::None when creating a fullscreen window on Windows 10. I set the resolution to match the screen resolution. This way I get all the benefits of fullscreen (double/triple buffering, fullscreen optimizations) but with much faster and seamless window creation. It seems Windows 10 detects it straight away I want a borderless fullscreen window. Wouldn't this commit break it? If so, wouldn't it be a good idea to add another flag sf::Style:Borderless and keep the current sf::Style::None behaviour?

Issue #1344, opened 2018-01-13
CMP0068: `RPATH` settings on macOS do not affect `install_name`.


mantognini:
Investigate & fix cmake warning about policy CMP0068 emitted with cmake 3.10+:

~~~
CMP0068: ``RPATH`` settings on macOS do not affect ``install_name``.

CMake 3.9 and newer remove any effect the following settings may have on the
``install_name`` of a target on macOS:

* ``BUILD_WITH_INSTALL_RPATH`` target property
* ``SKIP_BUILD_RPATH`` target property
* ``CMAKE_SKIP_RPATH`` variable
* ``CMAKE_SKIP_INSTALL_RPATH`` variable

Previously, setting ``BUILD_WITH_INSTALL_RPATH`` had the effect of
setting both the ``install_name`` of a target to ``INSTALL_NAME_DIR``
and the ``RPATH`` to ``INSTALL_RPATH``.  In CMake 3.9, it only affects
setting of ``RPATH``.  However, if one wants ``INSTALL_NAME_DIR`` to
apply to the target in the build tree, one may set
``BUILD_WITH_INSTALL_NAME_DIR``.

If ``SKIP_BUILD_RPATH``, ``CMAKE_SKIP_RPATH`` or
``CMAKE_SKIP_INSTALL_RPATH`` were used to strip the directory portion
of the ``install_name`` of a target, one may set ``INSTALL_NAME_DIR=""``
instead.

The ``OLD`` behavior of this policy is to use the ``RPATH`` settings for
``install_name`` on macOS.  The ``NEW`` behavior of this policy is to ignore
the ``RPATH`` settings for ``install_name`` on macOS.

This policy was introduced in CMake version 3.9.  CMake version
3.10.1 warns when the policy is not set and uses ``OLD`` behavior.
Use the ``cmake_policy()`` command to set it to ``OLD`` or ``NEW``
explicitly.

.. note::
  The ``OLD`` behavior of a policy is
  ``deprecated by definition``
  and may be removed in a future version of CMake.
~~~

JonnyPtn:
This was fixed by #1371, issue can be closed

PR #1371, opened 2018-02-10
Fix CMake warning on macOS since CMake 3.9

Ceylo:
This fixes this warning since CMake 3.9 where CMP0068 was introduced:
```
CMake Warning (dev):
  Policy CMP0068 is not set: RPATH settings on macOS do not affect
  install_name.  Run "cmake --help-policy CMP0068" for policy details.  Use
  the cmake_policy command to set the policy and suppress this warning.

  For compatibility with older versions of CMake, the install_name fields for
  the following targets are still affected by RPATH settings:

   SFML
   sfml-audio
   sfml-graphics
   sfml-network
   sfml-system
   sfml-window

This warning is for project developers.  Use -Wno-dev to suppress it.
```
Happens at CMake generation time.
Originally reported by @mantognini in https://github.com/SFML/SFML/issues/1344

I checked the output of the installed dylibs and frameworks with "otool -L", after having generated both with CMake 3.10 and 3.0, and they looked exactly the same as what's shipped for current latest SFML release.


Foaly:
I gave this a quick test and it fixes the warning as expected (cmake 2.10.3). I couldn't test the actuall install part as I am not quiet sure how.

Ceylo:
To install you just need to build the “install” target. Just make sure to select a writable install prefix in CMake config.

About testing, I don’t think installing will help much. You can but you mainly need to check that the frameworks/dylibs can be embedded in bundle apps exactly the same way as previously, or check the paths shown by “otool -L” on all generated binaries. Probably the easiest and most reliable.

Ceylo:
@mantognini Are you able to check so that this can be merged? Dunno if you're still away from home though.

mantognini:

JonnyPtn:
I've tested this and it does indeed make the warning go away. Doesn't seem to have any adverse affect on my game bundling either

mantognini:
@Ceylo, do you prefer having this one merged first, or rather after the big one #1335? I'll try this one as soon as possible.

Ceylo:
@mantognini I’d really prefer the big one to be merged first : it’ll fix CI, and some of my changes for iOS depend on it, and it’s much easier to merge and fix conflict in a small PR :)

eXpl0it3r:
Merged in 910458cfb366b4584269efb9d7f84b05f606f536

CMakeLists.txt
```diff
@@ -394,9 +397,14 @@ else()
...
     # NOTE: it's not required to link against SFML.framework
-    set_target_properties(SFML PROPERTIES
-                          BUILD_WITH_INSTALL_RPATH 1
-                          INSTALL_NAME_DIR "@rpath")
+    set_target_properties(SFML PROPERTIES INSTALL_NAME_DIR "@rpath")
+    if(NOT CMAKE_SKIP_BUILD_RPATH)
+        if (CMAKE_VERSION VERSION_LESS 3.9)
+            set_target_properties(${target} PROPERTIES BUILD_WITH_INSTALL_RPATH TRUE)
+        else()
+            set_target_properties(${target} PROPERTIES BUILD_WITH_INSTALL_NAME_DIR TRUE)
```
ChrisThrasher, line 405: Did we mean to use `SFML` here instead of `${target}`? I can't find a variable defined with the name `target` in CMakeLists.txt. I just found this out because I started enforcing uninitialized variable warnings.

https://github.com/SFML/SFML/actions/runs/6859641809/job/18652242185#step:14:43

PR #1375, opened 2018-02-24
Added support for specifying optional ThreadLocal destructor functions

binary1248:
Added support for specifying optional ThreadLocal destructor functions when they are supported by the platform.

This can be useful when managing RAII resources that can't be shared across threads without having to resort to complex and fragile cleanup machinery.

Test with:
```cpp
#include <SFML/System.hpp>
#include <iostream>

struct S
{
    S(int i)
    {
        std::cout << "S(" << i << ")" << std::endl;
    }

    ~S()
    {
        std::cout << "~S()" << std::endl;
    }
};

void destructor(void* s)
{
    delete static_cast<S*>(s);
}

sf::ThreadLocalPtr<S> s(0, destructor);

void threadFunction(int i)
{
    s = new S(i);
}

int main()
{
    sf::Thread t1(threadFunction, 1);
    sf::Thread t2(threadFunction, 2);

    t1.launch();
    t2.launch();
}
```

### DIFF
--- a/include/SFML/System/ThreadLocal.hpp
+++ b/include/SFML/System/ThreadLocal.hpp
@@ -40,6 +40,8 @@ namespace priv
     class ThreadLocalImpl;
 }
 
+typedef void(*ThreadLocalDestructorPointer)(void*);
+
 ////////////////////////////////////////////////////////////
 /// \brief Defines variables with thread-local storage
 ///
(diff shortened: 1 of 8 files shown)


binary1248:
Yes... you are right that sf::ThreadLocal's days are counted, and any serious developer working on projects now should probably have access to thread_local. However, in order to not block progress on other parts of SFML before we make the big jump to 3.x one day, we shouldn't neglect maintaining C++98 facilities for ourselves (I don't think that many people ever used sf::ThreadLocal(Ptr) in their code anyway).

Specifically, I had a sf::RenderTexture performance optimization in mind, but it would require a way to store thread local data. Each thread would also have to own the data, so we can't take the same route as we did with the internal contexts. Because the user can start and end threads from their own code at will without telling us, having TLS destructors is the only clean way to implement it.

LaurentGomila:
Can we talk about that sf::RenderTexture performance optimization before reviewing this PR?

binary1248:
I can start writing it on top of this branch if you want... Nothing convinces better than raw performance numbers anyway.

LaurentGomila:
Absolutely.

binary1248:
I found a better way of implementing the optimization. The pull request for it is in #1379. This change is no longer necessary.

LaurentGomila:
Great :+1:

PR #1376, opened 2018-02-24
Find OpenAL correctly on iOS

JonnyPtn:
Following on from #1263 

This removes the conditional checks which were stopping openAL being found properly on iOS, and updates the relevant includes

### DIFF
--- a/src/SFML/Audio/ALCheck.hpp
+++ b/src/SFML/Audio/ALCheck.hpp
@@ -29,14 +29,9 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Config.hpp>
-#ifdef SFML_SYSTEM_IOS
-    #include <OpenAl/al.h>
-    #include <OpenAl/alc.h>
-#else
-    #include <al.h>
-    #include <alc.h>
-#endif
 
+#include <al.h>
+#include <alc.h>
 
 namespace sf
 {
--- a/src/SFML/Audio/CMakeLists.txt
+++ b/src/SFML/Audio/CMakeLists.txt
@@ -67,9 +67,7 @@ endif()
 
 # find external libraries
 if(SFML_OS_IOS)
-    if(NOT SFML_OS_IOS)
-        find_host_package(OpenAL REQUIRED)
-    endif()
+    find_host_package(OpenAL REQUIRED)
     find_host_package(Vorbis REQUIRED)
     find_host_package(FLAC REQUIRED)
 else()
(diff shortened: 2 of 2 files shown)


Ceylo:
Does this find the OpenAL framework in macOS SDK, or iOS SDK, or the headers provided by SFML? In my opinion the best would be to use the framework from iOS SDK but I don’t think current code with find_host_package() does this. And actually there is most likely no search to do, just using flag «-framework OpenAL » should use the correct headers (from iOS SDK).

JonnyPtn:
For me, it finds the openAL framework in the iOS sdk: /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator11.2.sdk/System/Library/Frameworks/OpenAL.framework

Ceylo:
Ok.. then I really don’t understand the purpose of find_host_package() which is made (from what I understand) specifically to NOT find dependencies for the target system but for the host...

Can check what I get with your branch on my PC in ~2 days if you want to wait.

Ceylo:
Here is what I get with your branch
cmake cache

So for OpenAL it looks good indeed, even if I'm still puzzled by find_host_package() and its expected behavior…

JonnyPtn:
I think the intention is that find_package should only search in the relevant sdk, and find_host_package uses all the standard search paths?

If that's correct, then it should probably be find_package for openAL, instead of find_host_package?

Ceylo:
This is also my understanding… but hey, good luck getting it to work with just find_package() :)

JonnyPtn:
I just tried using find_package() for openAL, and it seems to still work as expected. It's the other 3rd party libs which don't like find_package(), because they aren't installed in the sdk, as expected. Are you able to do a quick test using find_package for openAL and confirm it still works for you too?

Ceylo:
find_package() indeed works fine for OpenAL, which makes sense indeed… so it could be fixed (at least for this dependency).

As a sidenote, it is however possible to wait for https://github.com/SFML/SFML/pull/1335 to be merged before this PR? #1335 assumes that all find_package() on iOS actually are find_host_package() so they were all factorized in sfml_find_package(). If we introduce an exception for OpenAL, it'll require significant changes in the already long-waiting PR :(

JonnyPtn:
I have no problem with waiting. I haven't changed this PR, so it still uses find_host_package, so it can either be merged in it's current state which (if I understand correctly) won't conflict with #1335, or If #1335 is merged first I will make the relevant changes to this PR

Ceylo:
See https://github.com/SFML/SFML/pull/1335/files#diff-6e3aba620b5d5027b0b256ccb9ad1063
A lot has changed actually so even if this PR was merged as-is it would conflict. I've been too greedy in changing everything, sorry 😄

JonnyPtn:
No worries, this is a super simple fix, so let's just pop it on hold, and I'll update once #1335 is merged.

PR #1377, opened 2018-02-25
Force correct iOS architecture for cmake

### LINKED ISSUE #1373: iOS build defaults to x86_64 architecture, regardless of specified platform
See discussion [here](https://en.sfml-dev.org/forums/index.php?topic=23517.msg161517#msg161517)

as the toolchain file is processed when `project(SFML)` is called, it doesn't overwrite the previous setting of CMAKE_OSX_ARCHITECTURES, configured [here](https://github.com/SFML/SFML/blob/master/CMakeLists.txt#L31)

Not 100% sure of the best way to fix this. @Ceylo fixed it by checking if a toolchain file is being used, which will work, but I think there should be a better way.

I also found that moving the call to `project()` to be above that block of Code also fixes it, and doesn't have any adverse effects on osx builds, however the 
[docs](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_ARCHITECTURES.html#variable:CMAKE_OSX_ARCHITECTURES) indicate it must be set before the first call to project(), so I'm reluctant to do that

JonnyPtn:
fixes #1373 

As per @Ceylo's suggestion, this forces the CMAKE_OSX_ARCHITECTURE variable within the toolchain file. This prevents the user from modifying this variable, but they still have full control using the IOS_PLATFORM variable

### DIFF
--- a/cmake/toolchains/iOS.toolchain.cmake
+++ b/cmake/toolchains/iOS.toolchain.cmake
@@ -180,7 +180,7 @@ elseif (${IOS_PLATFORM} STREQUAL SIMULATOR64)
     set (IOS_ARCH x86_64)
 endif (${IOS_PLATFORM} STREQUAL OS)
 
-set (CMAKE_OSX_ARCHITECTURES ${IOS_ARCH} CACHE string  "Build architecture for iOS")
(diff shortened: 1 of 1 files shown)


JonnyPtn:
1 week bump...

PR #1378, opened 2018-02-27
Add iOS demo

JonnyPtn:
I've added a basic iOS demo, mostly just copied the portable stuff from the android example.

To build, just set SFML_BUILD_EXAMPLES=TRUE when configuring (and of course use Xcode generator and iOS toolchain)


eXpl0it3r:

JonnyPtn:
Looking at the builedbot output, Not sure why the osx build is failing, looks like a permissions issue?

I think the iOS issue is because it's building for the iPhone platform as opposed to the simulator, so I should be able to fix that

eXpl0it3r:
OSX should be failing without the modern CMake changes or so. Should be fine.

JonnyPtn:
This signing stuff is proving more difficult than I expected... Getting the app running on a real device isn't hard, it just requires selecting the developer team within Xcode. The problem is, I can't find any way of setting that property from cmake, all the tips/fixes I've found just have different issues :S

examples/CMakeLists.txt
```diff
@@ -1,30 +1,36 @@
...
+    if(SFML_BUILD_AUDIO)
+        add_subdirectory(sound)
+        add_subdirectory(sound_capture)
+    endif()
+    if(SFML_BUILD_WINDOW)
+        add_subdirectory(window)
+    endif()
+    if(SFML_BUILD_GRAPHICS)
+        add_subdirectory(opengl)
+        add_subdirectory(shader)
```
Foaly: Is the following if else blocks really needed? I mean SFML_OS_IOS and SFML_OS_WINDOWS can't both be set.

eXpl0it3r:
Just tell @binary1248 what needs to be changed if you want to adapt the CI.

binary1248:
Maybe this helps?

https://stackoverflow.com/questions/40664125/cmake-and-code-signing-in-xcode-8-for-ios-projects
https://stackoverflow.com/questions/6910901/how-do-i-print-a-list-of-build-settings-in-xcode-project

JonnyPtn:
Yep those are two of the many fixes/tips I checked out, but there always seemed to be something else missing (perhaps due to using Xcode 9 instead of 8).

I'll keep investigating, but if you want to try to build on builedbot for simulator just add the cmake configuration variable "IOS_PLATFORM=SIMULATOR64"

Ceylo:
I confirm the easiest for now really is to build for simulator only. Couldn't get it to be 100% automatic when targetting real device, still had to manually select a certificate in Xcode UI. Note that whatever we try, we'll need a real (free) certificate from Apple if we want to build for device.

I saw there are two possible ways to handle code signing though:
- "partially automatic" : looks to require Xcode UI -> https://developer.apple.com/library/content/qa/qa1814/_index.html#//apple_ref/doc/uid/DTS40014030
- "manual" : https://developer.apple.com/library/content/qa/qa1911/_index.html

I guess the operations done in Xcode UI can be saved in the Xcode project file, and then reused by CI. But as in our case we generate the project with CMake, we need to find how to do the same with CMake commands.

binary1248:
Adding IOS_PLATFORM=SIMULATOR64 leads to [this](https://ci.sfml-dev.org/#/builders/4/builds/35).

JonnyPtn:
Thanks for checking @binary1248. That output is expected, as builedbot is currently set up with a manual architecture flag ("-arch arm64" on the build command, will not be necessary once #1377 is merged). If you remove that extra flag from the build command it should work with the simulator

Ceylo:
For curiosity, why was the arch flag needed at all in command line? I would expect this to be defined by CMake files, not the CI.

JonnyPtn:
It was needed because of #1373, so once that's merged it won't be required anymore

Ceylo:
Rrright ok!

eXpl0it3r:
@JonnyPtn could you resolve the merge conflict? 🙂

JonnyPtn:
No problem, should be rebased and resolved now

eXpl0it3r:
Should have run CI before merging, it's missing some stuff: https://ci.sfml-dev.org/#/builders/4/builds/95

Is this something we need to configure on CI?

texus:
The example does not seem to build on a case-sensitive filesystem, I get the following error:
```
CMake Error at cmake/Macros.cmake:205 (add_executable):
  Cannot find source file:

    /Users/texus/Desktop/SFML/examples/ios/main.cpp

  Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
  .hxx .in .txx
  examples/iOS/CMakeLists.txt:19 (sfml_add_example)
```

The example folder is called "iOS", not "ios", changing the SRCROOT value in `examples/iOS/CMakeLists.txt` seems to fix the problem. Is there a reason why the path is hardcoded instead of using CMAKE_CURRENT_SOURCE_DIR?

eXpl0it3r:
Probably no real reason, just didn't think of it. Feel free to make a PR for. 🙂

PR #1379, opened 2018-02-28
Optimized RenderTexture performance

binary1248:
Optimized RenderTexture performance when using the FBO implementation by removing unnecessary context switches and flushing.

In this change, FBOs no longer live in their own context if they don't have to. This reduces or even eliminates context switching all together in single window applications, giving "near native OpenGL performance". Furthermore, the glFlush() in display() was also removed in favour of explicit rebinding in RenderTarget in order to reduce the time waiting for the driver queue to flush to the GPU queue. This also saves a large amount of time.

In applications that make heavy use of RenderTextures, this can mean a significant performance gain if they were previously CPU-bottlenecked. This change shouldn't have an effect on the actual GPU workload in any way.

Here is a synthetic "benchmark" I used to test the difference with and without this change:
```cpp
#include <SFML/Graphics.hpp>
#include <sstream>

int main()
{
    sf::RenderWindow window(sf::VideoMode(800, 600), "Test", sf::Style::Titlebar | sf::Style::Close);
    window.setVerticalSyncEnabled(false);

    sf::Font font;
        return EXIT_FAILURE;

    sf::Text text("Test", font, 20);

    sf::RenderTexture renderTexture[10];

    for (int i = 0; i < 10; ++i)
        renderTexture[i].create(800, 600);

    sf::Sprite sprite;

    sf::Clock clock;
    std::ostringstream sstr;

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();

            if ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Escape))
                window.close();
        }

        for (int i = 0; i < 20; ++i)
        {
            for (int j = 0; j < 10; ++j)
            {
                renderTexture[j].clear(sf::Color(0, 0, 0));

                for (int j = 0; j < 10; ++j)
                    renderTexture[j].draw(text);

                renderTexture[j].display();
            }

            window.clear(sf::Color(0, 0, 0));

            for (int j = 0; j < 10; ++j)
            {
                sprite.setTexture(renderTexture[j].getTexture());

                for (int k = 0; k < 10; ++k)
                    window.draw(sprite);
            }
        }

        window.display();

        sstr.str("");
        sstr << "Test -- Frame: " << clock.restart().asSeconds() << " sec";

        window.setTitle(sstr.str());
    }

    return EXIT_SUCCESS;
}
```
Frame times on my laptop dropped from 800ms down to 200ms with this change. This is a 4x speedup.
Everybody is free to test this change for bugs and performance deltas using real-world applications. I'd like feedback regarding more realistic workloads as well.

### DIFF
--- a/include/SFML/Graphics/RenderTarget.hpp
+++ b/include/SFML/Graphics/RenderTarget.hpp
@@ -296,7 +296,7 @@ public:
     /// \return True if operation was successful, false otherwise
     ///
     ////////////////////////////////////////////////////////////
-    virtual bool setActive(bool active = true) = 0;
+    virtual bool setActive(bool active = true);
 
     ////////////////////////////////////////////////////////////
     /// \brief Save the current OpenGL render states and matrices
@@ -458,6 +458,7 @@ private:
     {
         enum {VertexCacheSize = 4};
 
+        bool      enable;         ///< Is the cache enabled?
         bool      glStatesSet;    ///< Are our internal GL states set yet?
         bool      viewChanged;    ///< Has the current view changed since last draw?
         BlendMode lastBlendMode;  ///< Cached blending mode
@@ -473,6 +474,7 @@ private:
     View        m_defaultView; ///< Default view
     View        m_view;        ///< Current view
     StatesCache m_cache;       ///< Render states cache
+    Uint64      m_id;          ///< Unique number that identifies the RenderTarget
 };
 
 } // namespace sf
--- a/include/SFML/Window/Context.hpp
+++ b/include/SFML/Window/Context.hpp
@@ -112,10 +112,25 @@ public:
     ////////////////////////////////////////////////////////////
     /// \brief Get the currently active context
     ///
+    /// This function will only return sf::Context objects.
+    /// Contexts created e.g. by RenderTargets or for internal
+    /// use will not be returned by this function.
+    ///
     /// \return The currently active context or NULL if none is active
     ///
     ////////////////////////////////////////////////////////////
     static const Context* getActiveContext();
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the currently active context's ID
+    ///
+    /// The context ID is used to identify contexts when
+    /// managing unshareable OpenGL resources.
+    ///
+    /// \return The active context's ID or 0 if no context is currently active
+    ///
+    ////////////////////////////////////////////////////////////
+    static Uint64 getActiveContextId();
 
     ////////////////////////////////////////////////////////////
     /// \brief Construct a in-memory context
--- a/include/SFML/Window/GlResource.hpp
+++ b/include/SFML/Window/GlResource.hpp
@@ -37,6 +37,8 @@ namespace sf
 
 class Context;
 
+typedef void(*ContextDestroyCallback)(void*);
+
 ////////////////////////////////////////////////////////////
 /// \brief Base class for classes that require an OpenGL context
 ///
@@ -56,6 +58,19 @@ protected:
     ///
     ////////////////////////////////////////////////////////////
     ~GlResource();
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Register a function to be called when a context is destroyed
+    ///
+    /// This is used for internal purposes in order to properly
+    /// clean up OpenGL resources that cannot be shared between
+    /// contexts.
+    ///
+    /// \param callback Function to be called when a context is destroyed
+    /// \param arg      Argument to pass when calling the function
+    ///
+    ////////////////////////////////////////////////////////////
+    static void registerContextDestroyCallback(ContextDestroyCallback callback, void* arg);
 
     ////////////////////////////////////////////////////////////
     /// \brief RAII helper class to temporarily lock an available context for use
--- a/src/SFML/Graphics/RenderTarget.cpp
+++ b/src/SFML/Graphics/RenderTarget.cpp
@@ -32,10 +32,14 @@
 #include <SFML/Graphics/VertexArray.hpp>
 #include <SFML/Graphics/VertexBuffer.hpp>
 #include <SFML/Graphics/GLCheck.hpp>
+#include <SFML/Window/Context.hpp>
+#include <SFML/System/Mutex.hpp>
+#include <SFML/System/Lock.hpp>
 #include <SFML/System/Err.hpp>
 #include <cassert>
 #include <iostream>
 #include <algorithm>
+#include <map>
 
 
 // GL_QUADS is unavailable on OpenGL ES, thus we need to define GL_QUADS ourselves
@@ -48,6 +52,36 @@
 
 namespace
 {
+    // Mutex to protect ID generation and our context-RenderTarget-map
+    sf::Mutex mutex;
+
+    // Unique identifier, used for identifying RenderTargets when
+    // tracking the currently active RenderTarget within a given context
+    sf::Uint64 getUniqueId()
+    {
+        sf::Lock lock(mutex);
+
+        static sf::Uint64 id = 1; // start at 1, zero is "no RenderTarget"
+
+        return id++;
+    }
+
+    // Map to help us detect whether a different RenderTarget
+    // has been activated within a single context
+    typedef std::map<sf::Uint64, sf::Uint64> ContextRenderTargetMap;
+    ContextRenderTargetMap contextRenderTargetMap;
+
+    // Check if a RenderTarget with the given ID is active in the current context
+    bool isActive(sf::Uint64 id)
+    {
+        ContextRenderTargetMap::iterator iter = contextRenderTargetMap.find(sf::Context::getActiveContextId());
+
+        if ((iter == contextRenderTargetMap.end()) || (iter->second != id))
+            return false;
+
+        return true;
+    }
+
     // Convert an sf::BlendMode::Factor constant to the corresponding OpenGL constant.
     sf::Uint32 factorToGlConstant(sf::BlendMode::Factor blendFactor)
     {
@@ -94,7 +128,8 @@ namespace sf
 RenderTarget::RenderTarget() :
 m_defaultView(),
 m_view       (),
-m_cache      ()
+m_cache      (),
+m_id         (getUniqueId())
 {
     m_cache.glStatesSet = false;
 }
@@ -109,7 +144,7 @@ RenderTarget::~RenderTarget()
 ////////////////////////////////////////////////////////////
 void RenderTarget::clear(const Color& color)
 {
-    if (setActive(true))
+    if (isActive(m_id) || setActive(true))
     {
         // Unbind texture to fix RenderTexture preventing clear
         applyTexture(NULL);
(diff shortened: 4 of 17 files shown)


roughbits01:
Tested it with my game which makes heavy use of RenderTextures (lightmap, glowing effect which needs multiple blur passes, heatmap, and other post effects) with a least 8 RenderTextures switches. Here is the result:

|                     |    Before      |    After    |
| ------------- |:-------------:| ----------:|
| CPU             | 12%             | 5%          |
| FPS              | 44                | 144         |

binary1248:
@LaurentGomila Hiding the fact that FBOs are unshareable in RenderTextures that you can throw around at will is nothing short of black magic. 😁

I admit, this ended up becoming a bit of a complex optimization, but still... even if one doesn't understand the code at first glance, a cursory review should still possible (to catch typos, review documentation, etc.).

myroidtc:
Tried it out. I can confirm it's a lot faster (don't have exact numbers) but there's some serious squish going on (comparison images below). Nothing changed except the SFML version. The other scenes in the game render fine, but they draw directly to the view. The simulation on the other hand is a RenderTexture's completed texture being drawn with a Sprite. The simulation's RenderTexture is 1280x720 but some 4096x4096 RenderTexture textures are drawn to it. The simulation's RenderTexture is also passed to a Shader Manager which applies bloom and other effects. It uses another couple of 1280x720s to do this.

zeran s folly 2018-03-01 21-13-00-21
zeran s folly 2018-03-01 21-19-33-88
zeran s folly 2018-03-01 21-14-26-19
zeran s folly 2018-03-01 21-21-54-49

anthnich:
Can confirm that this is much, much faster. Ran 60 sec profiles for a before and after. The seconds noted are how much time that the function took to run during the 60s of execution:

RenderTexture Function|Before|After
--------------------------|-------|------
RenderWindow:Render|1.56s|0.23s
RenderWindow:RenderToFinal|3.30s|0.31s

**RenderWindow:Render**: Handles drawing of the Rendertextures that are utilized.
**RenderWindow:RenderToFinal**: Handles the final pass with any shader effects that need to be applied.

Bugs
---------

**Missing Render Texture**

I have a separate render texture that I used to draw the HUD and menu text in game. Then, it is last in the drawing order so it is drawn on top of everything. For some reason, the render texture is not drawing.

The overlay render texture uses sf::View and the size from another render texture to set it's size and position.

**Black Render Textures**

When I use my screenshot function (where I create a texture, update it from the renderwindow, copy it to the image, then save a file), a number of my render textures are black as well. These render textures that are black in the screenshot are showing up in game, but not the screenshot.

Also, the missing overlay "HUD" render texture mentioned above shows up the screenshot.

binary1248:
@myroidtc @anthnich It would be helpful if you can provide a trace of a broken frame using something like [apitrace](http://apitrace.github.io/). Or... if you have the time, come up with a minimal code example that reproduces the problem, similar in form to the test case I provided above.

Also, if you haven't already done so, make sure to give the applications a go with debugging enabled and check to see if there are any OpenGL error messages output to the console.

naezith:
I tested in couple different levels. It boosted FPS a lot!

500 -> 750
700 -> 1000
1000 -> 1150
1000 -> 1300

There is one issue quickly appeared, my water shader which reflects stuff, does not work. Waters are invisible. No shader errors in console. 

I draw them like this

```
// This is called once
          rtFBuffer.create(rt.getSize().x, rt.getSize().y);
          reflection = std::make_unique<sf::Sprite>(rtFBuffer.getTexture());

// After this line, these run every frame

        // Copy RT to Buffer to be the reflection
        {
            rt.display();

            sprFBuffer->setScale(1.0f, 1.0f);
            sprFBuffer->setPosition(0, 0);

            rtFBuffer.draw(*Game::i().sprFBuffer);

            rtFBuffer.display();
        }

        rt.resetGLStates();
        sf::Shader::bind(&shader_water);

        // Crop reflection areas and draw waters to main RT one by one
        for(auto& idx : watersInScreen) {
              reflection->setTextureRect(boundsOfWater);
              rt.draw(reflection);
        }

        sf::Shader::bind(nullptr);
```

binary1248:
@naezith What is the performance difference with and without
```cpp
rt.resetGLStates();
sf::Shader::bind(&shader_water);
```
?

You are basically relying on undefined behaviour. SFML never guaranteed that it would not touch the shader binding within a call to `.draw()`. If it has to, it will, and you shouldn't rely on it staying this way in the future, independent from the changes in this pull request. The only guarantee that is made is that if you don't specify a shader in the `RenderStates` passed to `.draw()` that SFML with either disable a currently active shader or leave it deactivated if it is determined that none is active.

naezith:
@binary1248 I didn't try that but giving shader to draw func as an argument was a lot slower, that's what I remember. Yet this works on Windows, Mac, Linux in every distro. Still I might try again.

anthnich:
@binary1248 I sent you a PM on the SFML forums with the apitrace info.

binary1248:
@anthnich @myroidtc @naezith Give the current version a try. I fixed several bugs.

myroidtc:
Still getting the aspect ratio bug. In this scene, the background's RenderTexture (with the big circle) is drawn to the simulation's RenderTexture which is then drawn to the screen. The background's aspect ratio is even more exaggerated than the simulation's which gives me the impression that it's being applied to both of them.

mana_normal
mana_stretch

MarioLiebisch:
Out of curiosity, what happens if you use a texture/screen resolution that's a power of two? like 1024x1024? Have you tried figuring out by which factor the whole thing is deformed?

binary1248:
@myroidtc Can you provide an apitrace trace like @anthnich did or some kind of pseudo-code/snippet that demonstrates how you are drawing to the different RenderTargets? Working off of your current description, it is kind of hard to come up with a theory on why it is exhibiting this behaviour.

myroidtc:
Sent you a PM on the forums with a link to it.

naezith:
@binary1248 Why don't I see a new commit? Latest and the only one is Feb 28.

binary1248:
@naezith The commit timestamp doesn't change if I amend it.

naezith:
@binary1248 Okay I just tested it again. My shader issue is solved and pretty good FPS boost is still there but a bit less compared to old one I think, I'm not completely sure. I'll check if I have issues on Linux and Mac, if not, this will be in the next patch of my game. Thanks a lot!

binary1248:
@myroidtc @anthnich @naezith Can you test the newest version? I redid the internal RenderTarget-to-Context tracking. It should take care of the edge cases as well now.

@LaurentGomila Do you think it is worth it to move GL state tracking out of RenderTarget and into e.g. a specialized class? It becomes really messy to cache states within the RenderTarget class when you start to have multiple RenderTargets all trying to live within the same context.

myroidtc:
No improvement, sorry.

binary1248:
@myroidtc Can you send me a new trace? I narrowed it down to the projection matrix of the RenderTexture not being set to what it should be. Where the wrong value is coming from, I have yet to find out. Kind of hard debugging without access to the process memory.

LaurentGomila:
> @LaurentGomila Do you think it is worth it to move GL state tracking out of RenderTarget and into e.g. a specialized class? It becomes really messy to cache states within the RenderTarget class when you start to have multiple RenderTargets all trying to live within the same context.

If you think you can come up with something cleaner, without a lot of dependencies on `RenderTarget` internals, sure, let's do this.

binary1248:
@myroidtc I figured out what the problem was. In case the RenderTexture was created during a stage of your application before a context was set active for a longer period of time, the FBO would not be immediately created causing the RenderTexture implementation to also erroneously skip initialization of its view leading to an improper projection matrix being applied when drawing.

The current version should have fixed this problem, as well as a performance "regression" introduced in the previous version. Mind to give it try?

myroidtc:
You got it! Works perfectly now. Here's a speed comparison:

image

"Bloom" refers to a scene with a bloom effect where the scene is passed back and forth between two RenderTextures a few times.

"Water" refers to a scene with no bloom but with a simple wave shader.

anthnich:
Fantastic job! Everything seems to be working now. 

My CPU usage went from ~10% to about 6% and my GPU usage went from ~15% to 
~7% from my observance of Task Manager. My profile stands in line with the one I provided above. It's a good bit speedier.

naezith:
Would be great if you actually do a normal commit instead of "amend" thing. I try to merge this one with my own fork but it says "there is nothing to merge" because I already merged it before. It's being very confusing like I have no idea if it got updated or not. I manually did some copy pasting stuff to make sure, it's painful. Tell me if there is an easier way.

naezith:
Okay as I said it was running perfect for me on Windows 10. On Ubuntu 16.04, it was stuttering when Steam Overlay was coming up because someone types. On macOS High Sierra, it's also good. 

So I merged these with my SFML and updated my game. Yet tons of people reports visual issues.

Here is a video https://www.dropbox.com/s/6zfm8yevure5gy3/2018-03-13%2019-58-11.mp4?dl=0

naezith:
I guess this patch broke the one cheat I did and it was working before. Drawing from same RenderTexture to the same RenderTexture. I'll avoid that.

binary1248:
No matter whether you use SFML before this patch, SFML after this patch, or decide to code everything in Win32 + raw OpenGL, what you described invokes undefined behaviour. Sampling from the same texture image that you are currently updating in a draw should always be avoided. I'm actually quite surprised that you only noticed this problem after applying this patch, but then again... undefined behaviour.

Relevant links:
https://www.khronos.org/opengl/wiki/Framebuffer_Object#Feedback_Loops
https://www.khronos.org/opengl/wiki/Memory_Model#Framebuffer_objects

> It is possible to bind a texture to an FBO, bind that same texture to a shader, and then try to render with it at the same time.
> It is perfectly valid to bind one image from a texture to an FBO and then render with that texture, as long as you prevent yourself from sampling from that image. If you do try to read and write to the same image, you get undefined results. Meaning it may do what you want, the sampler may get old data, the sampler may get half old and half new data, or it may get garbage data. Any of these are possible outcomes.
> Do not do this. What you will get is undefined behavior.

Possible alternatives are listed under the second link. Performance-wise, nothing really speaks against implementing a ping-pong style approach in which you use 2 RenderTextures to achieve what you want.

naezith:
Yeah I did a quickfix, everything is fine now.

binary1248:
@LaurentGomila Considering this change set seems to work as expected now, I don't feel good about combining the required changes for improved state caching into this PR. From what I estimate, almost every GlResource in sfml-graphics would be affected by the change and it should really be tackled on its own in a separate PR environment.

binary1248:
Bump.

eXpl0it3r:
Should we consider this for SFML 2.5?

mantognini:
> Should we consider this for SFML 2.5?

If the PR is stable, yes! This way we will get feedback sooner and in the worse case we rollback this.

binary1248:
I'll still have to rebase this on the new RenderTextureImplFBO.

binary1248:
Rebased on top of current master.

Because this rebase was a bit bigger than usual, @anthnich @myroidtc @naezith Can you guys test the current version? I don't think the current code will change much even if some final adjustments are to be made.

myroidtc:
Mirroring what @anthnich said, works fine on my end.

PR #1383, opened 2018-03-13
Added raw mouse input event to sf::Window

### LINKED ISSUE #304: Raw input for mouse movement
SFML currently uses WM_MOUSEMOVE and this smooths out the movement and is dependent on the resolution. Smoothing is preferred when the game integrates with the desktop. However when the mouse controls a 3d camera smoothing is less desired.

There should therefore be a option to enable non-smoothed  "raw" input. The solution seems to be WM_INPUT http://msdn.microsoft.com/en-us/library/windows/desktop/ee418864%28v=vs.85%29.aspx

WM_INPUT also provides a resolution independent movement that might be good if the game needs to be run at low resolutions and/or sub-pixel movements are needed.

The forum thread where this issue originated from: http://en.sfml-dev.org/forums/index.php?topic=9426

sro5h:
This attempts to fix #304 by adding `MouseMoveRawEvent` to sf::Window on Linux and Windows (based on [hornta](https://github.com/hornta)s work).
A macOS implementation is still missing.

The `MouseMoveRawEvent` gets emitted every time the mouse is moved and holds the raw, unaccelerated, and unitless movement deltas in its members `deltaX` and `deltaY`.

To use the new feature you have to react to the `MouseMoveRawEvent`s that are emitted:
```c++
#include <SFML/Graphics.hpp>
#include <iostream>

int main()
{
    sf::RenderWindow window(sf::VideoMode(400, 400), "Raw Input Example");

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();

            else if (event.type == sf::Event::MouseMovedRaw)
            {
                std::cout << "raw movement, x = " << event.mouseMoveRaw.deltaX;
                std::cout << " y = " << event.mouseMoveRaw.deltaY << std::endl;
            }
        }

        window.clear();
        window.display();
    }

    return EXIT_SUCCESS;
}
```


sro5h:
Sure, will do :)

sro5h:
My fork is up to date with sfml/master and the library and my simple example both compile under Ubuntu 18.04 and Windows 8.1 :)

sro5h:
I don't know why the travis-ci build fails, could it be that it is missing the xinput2 dependency?

Keeo:
Looks like that, could you try add `libxi-dev` to `apt-packages` in `.travis.yml` file? I think i can't add commit on your branch.
https://packages.ubuntu.com/source/trusty/xinput

Anyway, thank you for the updating the PR 🥇

Unarelith:
@sro5h I rebased this PR on [my fork](https://github.com/Unarelith/SFML/tree/mouse_raw_movement)

However I've an issue... If I hold a mouse button, X11 GenericEvents are only sent when `sf::Window::setMouseCursorGrabbed` is set to `true`, otherwise it won't work.

Here is the code I used to test:
```cpp
#include <ctime>
#include <iostream>

#include <SFML/OpenGL.hpp>
#include <SFML/Window.hpp>

int main(int, char **) {
	sf::Window window;
	window.create(sf::VideoMode(800, 600), "OpenGL", sf::Style::Close);
	// window.setMouseCursorGrabbed(true);

	bool running = true;
	while (running) {
		sf::Event event;
			if (event.type == sf::Event::Closed || (event.type == sf::Event::KeyPressed && event.key.code == sf::Keyboard::Escape)) {
				running = false;
			}
			else if (event.type == sf::Event::MouseButtonPressed) {
				std::cout << "Mouse button pressed" << std::endl;
			}
			else if (event.type == sf::Event::MouseButtonReleased) {
				std::cout << "Mouse button released" << std::endl;
			}
			else if (event.type == sf::Event::MouseMovedRaw) {
				std::cout << "Mouse raw movement" << std::endl;
			}
		}

		glClear(GL_COLOR_BUFFER_BIT);
		window.display();
	}

	window.close();

	return 0;
}
```

Try holding left mouse button for example, and moving your mouse. You won't see "Mouse raw movement" displayed until you release the button.

If you uncomment the `m_window.setMouseCursorGrabbed(true);` line, it'll work.

It happens both on this PR and on my rebase. :/

sro5h:
I will look into it. I just tried your example, but i'm on my Laptop and i can't move the cursor while i press a mouse button. I will try it on my pc, as soon as possible.

include/SFML/Window/Event.hpp
```diff
@@ -87,6 +87,16 @@ class Event
         int y; ///< Y position of the mouse pointer, relative to the top of the owner window
     };
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Mouse move raw event parameter (MouseMovedRaw)
```
binary1248, line 91: Parameters not parameter.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -97,7 +98,8 @@ namespace
     Bool checkEvent(::Display*, XEvent* event, XPointer userData)
     {
         // Just check if the event matches the window
-        return event->xany.window == reinterpret_cast< ::Window >(userData);
+        return event->xany.window == reinterpret_cast< ::Window >(userData)
+                || event->type == GenericEvent;
```
binary1248, line 102: The line break here is unnecessary, we don't enforce strict line length limits.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -362,6 +364,33 @@ namespace
...
     }
 
+    void initRawMouse(::Display* disp)
+    {
+        bool enabled = false;
+        int opcode, event, error;
+
+        if (XQueryExtension(disp, "XInputExtension", &opcode, &event, &error))
+        {
+            int major = 2, minor = 0;
```
binary1248, line 374: When initializing variables where they are defined, we prefer them to be placed on separate lines.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -2071,6 +2104,45 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
 
             break;
         }
+
+        // Mouse moved raw
+        case GenericEvent:
+        {
+            if (XGetEventData(m_display, &windowEvent.xcookie)
+                    && windowEvent.xcookie.evtype == XI_RawMotion)
```
binary1248, line 2112: The line break here is unnecessary, we don't enforce strict line length limits.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -2071,6 +2104,45 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
...
+                }
+
+                Event event;
+                event.type = Event::MouseMovedRaw;
+                event.mouseMoveRaw.deltaX = relativeValues[0];
+                event.mouseMoveRaw.deltaY = relativeValues[1];
+                pushEvent(event);
+            }
+
+            XFreeEventData(m_display, &windowEvent.xcookie);
```
binary1248, line 2143: XFreeEventData would be called even if XGetEventData returns False.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -2071,6 +2104,45 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
...
         }
+
+        // Mouse moved raw
+        case GenericEvent:
+        {
+            if (XGetEventData(m_display, &windowEvent.xcookie)
+                    && windowEvent.xcookie.evtype == XI_RawMotion)
+            {
+                const XIRawEvent* rawEv;
+                rawEv = static_cast<const XIRawEvent*>(windowEvent.xcookie.data);
```
binary1248, line 2115: These lines can be combined.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -2071,6 +2104,45 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
...
+
+                const int mouseAxis = 2;
+                int relativeValues[2] = { 0, 0 };
```
binary1248, line 2118: `mouseAxis` could be used to specify the array size.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -2071,6 +2104,45 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
...
+                int limit = rawEv->valuators.mask_len;
+
+                if (limit > mouseAxis)
+                    limit = mouseAxis;
+                limit *= 8;
```
binary1248, line 2123: I don't understand why `limit` has to be multiplied with 8.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -2071,6 +2104,45 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
...
+            {
+                const XIRawEvent* rawEv;
+                rawEv = static_cast<const XIRawEvent*>(windowEvent.xcookie.data);
+
+                const int mouseAxis = 2;
+                int relativeValues[2] = { 0, 0 };
+                int limit = rawEv->valuators.mask_len;
+
+                if (limit > mouseAxis)
+                    limit = mouseAxis;
```
binary1248, line 2122: This can be replaced with `std::min`.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -2071,6 +2104,45 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
...
+                // Get relative input values
+                const double* inputValues = rawEv->raw_values;
+                for (int i = 0; i < limit; ++i)
+                {
+                    if (XIMaskIsSet(rawEv->valuators.mask, i))
+                    {
+                        relativeValues[i] = static_cast<int>(*inputValues);
+                        ++inputValues;
+                    }
+                }
```
binary1248, line 2134: If `limit` happened to be larger than 2 we could end up writing past the end of the array.

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -123,6 +123,17 @@ namespace
...
+
+    // Register a RAWINPUTDEVICE representing the mouse to receive raw
+    // mouse deltas using WM_INPUT
+    void initRawMouse()
+    {
+        RAWINPUTDEVICE rawMouse = { 0x01, 0x02, 0, NULL };
+        bool enabled = RegisterRawInputDevices(&rawMouse, 1, sizeof(RAWINPUTDEVICE)) == TRUE;
+
+        if (!enabled)
+                sf::err() << "Failed to initialize raw mouse input" << std::endl;
```
binary1248, line 135: Indent too large.

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -985,6 +1004,31 @@ void WindowImplWin32::processEvent(UINT message, WPARAM wParam, LPARAM lParam)
             pushEvent(event);
             break;
         }
+
+        // Mouse move raw event
+        case WM_INPUT:
+        {
+            HRAWINPUT hRawInput = (HRAWINPUT)lParam;
```
binary1248, line 1011: We prefer not to use C-style casts.

eXpl0it3r:
One point that was brought up on Discord is whether it wouldn't be better to integrate the event into the existing Mouse event?

Unarelith:
I think it would definitely be better to have a single `MouseMoved` event with `x`, `y`, `deltaX`, `deltaY` attributes.

binary1248:
The raw mouse movement data comes in separate events on both Linux and Windows. We can't assume that a normal mouse move event will be received at the same time (or even in which order) so we would have to buffer the data incurring an unnecessary delay.

Unarelith:
Then I'm curious about how SDL and glfw does that, because iirc they both pack the two events into a single one.

binary1248:
What SDL and GLFW do is multiplex both event sources. In simple terms, you can choose between getting "normal" mouse data or raw mouse data.

The calls to do this would be:

`glfwSetInputMode(window, GLFW_RAW_MOUSE_MOTION, GLFW_TRUE);`

and

`SDL_SetRelativeMouseMode(SDL_TRUE);`

Depending on which you decide to activate, the event structure/callback is used accordingly. In neither of those libraries do they report "normal" and raw data simultaneously because as I said, this is not supported on any windowing system.

sro5h:
Sorry for taking so long to have a look at it. It seems it had to do with the value i gave to XIEventMask::deviceid (changed it from XIAllMasterDevices to XIAllDevices and it worked as expected) when selecting for the XI_RawMotion events. It was more of an educated guess supported by some google searches though. I don't know if that's just me, but i find it really hard to find information on the X Input Extension 2. I will do some more reading and testing to see if i can find anything that i may have broken.
I will also update the pullrequest to incorporate the changes requested by @eXpl0it3r.

include/SFML/Window/Event.hpp
sro5h: I never worked with this feature. Is it ok if i just add a commit to my pull request and click "Resolve conversation" on the respective suggestion?

Unarelith: Usually yes, that's actually the purpose of that feature.

src/SFML/Window/Unix/WindowImplX11.cpp
sro5h: I tried to and it seems that [X11 defines](https://stackoverflow.com/questions/37908716/x11-xlibint-h-naming-conflict-issues-are-they-supposed-to-be-named-min-max) a macro named `min`. Should i `#undef` it at the beginning of `WindowImplX11.cpp`?

src/SFML/Window/Unix/WindowImplX11.cpp
sro5h: I didn't know C++ allowed that.

src/SFML/Window/Unix/WindowImplX11.cpp
sro5h: Honestly i don't know anymore, SDL did it in a similiar way. But you are right i should not write past the array and there is no point in iterating past `mouseAxis` if i don't use the values anyway.

binary1248:
Squashed/rebased/modernized version can be found at #3057. The discussion can be continued there.

PR #1384, opened 2018-03-14
Update Xcode templates

mantognini:
I removed a few things that are no longer relevant.

### DIFF
--- a/tools/xcode/templates/SFML/SFML Compiler.xctemplate/TemplateInfo.plist.in
+++ b/tools/xcode/templates/SFML/SFML Compiler.xctemplate/TemplateInfo.plist.in
@@ -45,74 +45,6 @@ subject to the following restrictions:
         <string>Xcode.Xcode3.ProjectTemplateUnitKind</string>
 
         <!--
-             OPTIONS
-         -->
-        <key>Options</key>
-        <array>
-            <dict>
-                <key>Identifier</key>
-                <string>compilerSettingsType</string>
-
-                <key>Name</key>
-                <string>[ADVANCED] C++ Compiler and Standard Library</string>
-
-                <key>Description</key>
-                <string>If you don't know what is it about, use the default value. Note that you'll need a version of SFML compiled with Clang and libc++ to use C++11!</string>
-
-                <key>Default</key>
-                <string>C++11 with Clang and libc++</string>
-
-                <key>NotPersisted</key>
-                <true />
-
-                <key>Type</key>
-                <string>popup</string>
-
-                <key>Units</key>
-                <dict>
-                    <!-- Use Clang and libstdc++ -->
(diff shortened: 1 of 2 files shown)


JonnyPtn:
This works fine for me.

A couple of things to mention (Not caused by this PR as they are present on master):
 - Would prefer to see high resolution capable set to "NO", so the window is the expected size
 - Is it possible to set the working directory of the console app? with the current settings it doesn't load the resources, so I have to manually specify the working directory

mantognini:
Thanks for testing!

For future readers, I've answered to the resolution remark [here](https://en.sfml-dev.org/forums/index.php?topic=23809.msg162145#msg162145).

> Is it possible to set the working directory of the console app?

Yes, you can do that quite easily with the POSIX `chdir` function, based on the parameter of the `main` function. See [here](http://en.cppreference.com/w/cpp/language/main_function):

> argv[0] is the pointer to the initial character of a null-terminated multibyte string that represents the name used to invoke the program itself (or an empty string "" if this is not supported by the execution environment).

However, I would *not* recommend doing so because it wouldn't be 100% compatible with application bundles (hence why the templates don't do it automatically).

Instead defining a `resource_path` function seems more flexible and offers a better overall design in my opinion: it encapsulates the feature that is actually needed while reducing the side-effects (no need to change the working directory). Of course, one would need to "initialise" the `resource_path` function with the information stored in `argv[0]`. But one could also use this opportunity to let the user configure a root directory for resources and therefore switch settings easily. Furthermore, one would only need to change this function at compile time to use the application bundle resource directory -- or the mechanism provided by other runtime environment for that matter.

HTH.

JonnyPtn:
I think you misunderstand - I mean the working directory of the Xcode template project, not the program itself.

mantognini:
I see. I don't know if it is possible to do this.. But you'll face the same
issue: double clicking your app in finder will result in an unexpected WD.
You'll also need to configure your profiler and other tools, probably
manually too. So implementing a robust & flexible system for loading
resources seems worth it, especially since it can be reused across
projects.

JonnyPtn:
OK, Let me try and be even clearer. I'm not talking about modifying the working directory of the app itself, just the Xcode project working directory, so it can be run in the debugger straight away. This has no impact on the product and doesn't affect the end result in the slightest. It's ONLY relevant when running the project in Xcode. It's this setting here I'm talking about:
screen shot 2018-03-18 at 12 15 04

Which I currently have to manually set before the CLT template will run correctly in Xcode. Ideally, it would be nice if this could be set in the template to "${PROJECT_DIR}/${PROJECT_NAME}" if that's possible? I'm not familiar with Xcode templates, but I'll have a look around and see if there are any suggestions for this

mantognini:
Yes, I got it in your previous message. But, as I said, I don't know if its
possible to set it through templates.

PR #1386, opened 2018-03-15
window: change case of Dbt.h inclusion

hanetzer:
Windows should have no issue with this change, as it addresses
files in a case-insensitve manner. Said header is installed in
lowercase on mingw-w64, and won't build since commit
22f1b85515242c44d7e5a1be1e3960bdf6648b52

### DIFF
--- a/src/SFML/Window/Win32/WindowImplWin32.cpp
+++ b/src/SFML/Window/Win32/WindowImplWin32.cpp
@@ -39,7 +39,10 @@
 #include <GL/gl.h>
 #include <SFML/System/Err.hpp>
 #include <SFML/System/Utf.hpp>
-#include <Dbt.h>
+// dbt.h is lowercase here, as a cross-compile on linux with mingw-w64
+// expects lowercase, and a native compile on windows, whether via msvc
+// or mingw-w64 addresses files in a case insensitive manner.
+#include <dbt.h>
 #include <vector>
 #include <cstring>
 


MarioLiebisch:
This is essentially an issue with MinGW's bundled platform SDK (or the compiler being case sensitive, if you want). The official platform SDK by Microsoft uses the uppercase version, `Dbt.h` ([example](https://msdn.microsoft.com/en-us/library/windows/desktop/aa363205%28v=vs.85%29.aspx)) and this is also the casing of the actual file names.

hanetzer:
@MarioLiebisch fair enough, but you may want to consider the examples of SDL2 and glfw3:
http://hg.libsdl.org/SDL/file/c808a74c019e/src/joystick/windows/SDL_windowsjoystick.c#l45
https://github.com/glfw/glfw/blob/master/src/win32_platform.h#L68

They both use the lowercase naming convention, and both provide mingw-w64 and msvc
compiled binary dists with no issue.

MarioLiebisch:
Not saying we can't do it – not sure what the others think about it. It's a trivial change after all – as you've mentioned. Are there any chances someone might actually use the official platform SDK for cross-compiling rather than MinGW?

hanetzer:
I suppose it could maybe be done, but I've never seen an example of it being done
in the real world.

binary1248:
I don't see any risk in this change, and if it solves a problem without introducing any side-effects, then why not? We already `#include <windows.h>` even though MSDN also lists it in upper case.

hanetzer:
@binary1248 good point on the `#include <windows.h>` bit. Glad to see this being
implemented.

LaurentGomila:
I would also add an explicit comment, so that nobody comes months or years later to change it back to upper case.

hanetzer:
@LaurentGomila should I add that to the pr prior to merging, then?

LaurentGomila:
Yes, please :)

hanetzer:
@LaurentGomila is this to your liking?

LaurentGomila:
Looks good :+1:

eXpl0it3r:
Merged in 5c20c67469e2be17a727bf6cfe3f5672bf50df83

PR #1387, opened 2018-03-15
build: generate and install pkg-config files on mingw

hanetzer:
Mingw-w64 is typically used in conjunction with a unix-like environment,
whether as a cross compiler on linux or as a 'cross' compiler with
cygwin (still win32 and win32, but kinda like cross-compiling), and the
use of pkg-config files within this sort of setup is rather common.


hanetzer:
To achieve the desired effect I added a new variable, `SFML_OS_WINDOWS_MINGW` to
specify that this is a windows compile using the mingw-w64 runtime and libraries. I could
just add a `MINGW` check to the `if(SFML_OS_LINUX OR SFML_OS_FREEBSD)` check,
which would reduce the patch to a single addition and subtraction, but I was attempting to
match your current conventions of wrapping ''global'' variables into SFML ''wrapped'' ones.

hanetzer:
Alternatively one can just generate pkg-config files for all platforms except msvc,
as just about every other platform there is can consume pkg-config files, being
as they're all POSIX to some degree or another.

eXpl0it3r:
Why close it?

hanetzer:
@eXpl0it3r because there apparently is already a commit in master which does this, more
or less. its 5fe5e5d6d7792e37685a437551ffa8ed5161fcc1, and since gentoo unconditionally
sets `SFML_INSTALL_PKGCONFIG_FILES` in its ebuild its not needed on 2.4.2 if we add this
commit as a patch.

eXpl0it3r:
I see, thanks for the clarification and the contributions! 🙂

PR #1390, opened 2018-03-18
Fixed OpenGL version string being parsed incorrectly on some platforms

### LINKED ISSUE #1249: iOS OpenGL ES version string incorrect formatting
SFML calls glGetVersionString in GLContext.cpp. It assumes the first 3 characters of the string are major.minor. example (on windows PC): "4.5.0 NVIDEA 382.53"

However, on iOS devices this format is not consistent. Here are examples below:
dev iPad Air: "OpenGL ES-CM 1.1 Apple A7 GPU - 95.58"
dev iPhone SE: "OpenGL ES-CM 1.1 Metal - 33"
sim iPhone 5: "OpenGL ES-CM 1.1 APPLE-14.0.15"
sim iPad Pro 9.7in: "OpenGL ES-CM 1.1 APPLE-14.0.15"

this causes some undefined behavior with rendering, because sfml doesn't use the correct opengl es version.

Please implement a fix to take into account iOS formatting. 
I also recommend this be tested for android as well, unfortunately I am not set with android to confirm or deny if there is even a problem for that platform.

binary1248:
Fixes #1249.

### DIFF
--- a/src/SFML/Window/GlContext.cpp
+++ b/src/SFML/Window/GlContext.cpp
@@ -38,6 +38,7 @@
 #include <set>
 #include <cstdlib>
 #include <cstring>
+#include <cctype>
 #include <cassert>
 
 #if !defined(SFML_OPENGL_ES)
@@ -199,6 +200,26 @@ namespace
 
     // Supported OpenGL extensions
     std::vector<std::string> extensions;
+
+    // Helper to parse OpenGL version strings
+    bool parseVersionString(const char* version, const char* prefix, unsigned int &major, unsigned int &minor)
+    {
+        std::size_t prefixLength = std::strlen(prefix);
+
+        if ((std::strlen(version) >= (prefixLength + 3)) &&
+            (std::strncmp(version, prefix, prefixLength) == 0) &&
+            std::isdigit(version[prefixLength]) &&
+            (version[prefixLength + 1] == '.') &&
+            std::isdigit(version[prefixLength + 2]))
+        {
+            major = version[prefixLength] - '0';
+            minor = version[prefixLength + 2] - '0';
+
+            return true;
+        }
+
+        return false;
+    }
 }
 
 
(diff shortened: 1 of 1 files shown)


binary1248:
`std::strncmp` compares two _possibly null-terminated_ character arrays. If the null terminator is encountered in either array before n characters are compared, it ends the comparison. In our case, there is no danger because both parameters are guaranteed to be null-terminated.

LaurentGomila:
I was talking about what happens inside the blocks:

```cpp
m_settings.majorVersion = version[13] - '0';
m_settings.minorVersion = version[15] - '0';
```

binary1248:
I guess you don't want to have to assume that the OpenGL implementation is standard compliant? 😁

LaurentGomila:
What standard? We are "parsing" strings we know nothing about 😀

binary1248:
Fixed.

LaurentGomila:
Looks very robust now 😆 

What about a helper function to avoid all these repetitions?

```cpp
bool parseVersionString(const char* prefix, unsigned int versionPosition, unsigned int &major, unsigned int &minor);
```

binary1248:
Naaah... an extra function just to move out 4 if statements? 😨

mantognini:
It would makes things more readable and shorter. 👍 

Maybe even one if statement:

```
if (parseVersionString(...) || parseVersionString(...) || ...) {
    m_settings.majorVersion = major;
    m_settings.minorVersion = minor;
} else { warning }
```

with some better formatting, ofc.

binary1248:
Fixed.

LaurentGomila:
Is there a reason to use strncmp with the last argument being strlen of one string? If it stops at the first encountered \0 anyway, maybe we can just use strcmp?

binary1248:
Using `std::strncmp` we can check for prefixes. `std::strncmp("A", "ABC", 1)` will return 0 while `std::strcmp("A", "ABC")` will return non-zero.

src/SFML/Window/GlContext.cpp
```diff
@@ -199,6 +200,26 @@ namespace
...
     // Supported OpenGL extensions
     std::vector<std::string> extensions;
+
+    // Helper to parse OpenGL version strings
+    bool parseVersionString(const char* version, const char* prefix, unsigned int &major, unsigned int &minor)
+    {
+        std::size_t versionPosition = std::strlen(prefix);
+
+        if ((std::strlen(version) >= (versionPosition + 3)) &&
+            !std::strncmp(version, prefix, std::strlen(prefix)) &&
```
Foaly: I guess the third parameter could be replaced with `versionPosition`, but this is a minor detail.

binary1248:
Fixed.

eXpl0it3r:
Does it now look okay, @LaurentGomila?

eXpl0it3r:
@kingnobody8 Can you try the code? Need more of an actual test than review. 🙂

binary1248:
@LaurentGomila Fixed.

JonnyPtn:
Looks good to me! tested on iPhone 8

src/SFML/Window/GlContext.cpp
```diff
@@ -582,18 +603,30 @@ void GlContext::initialize(const ContextSettings& requestedSettings)
...
-            m_settings.minorVersion = version[2] - '0';
+            // OpenGL ES Common Lite profile: The beginning of the returned string is "OpenGL ES-CL major.minor"
+            // OpenGL ES Common profile:      The beginning of the returned string is "OpenGL ES-CM major.minor"
+            // OpenGL ES Full profile:        The beginning of the returned string is "OpenGL ES major.minor"
+            // Desktop OpenGL:                The beginning of the returned string is "major.minor"
+
+            if (!parseVersionString(version, "OpenGL ES-CL ", m_settings.majorVersion, m_settings.minorVersion) &&
+                !parseVersionString(version, "OpenGL ES-CM ", m_settings.majorVersion, m_settings.minorVersion) &&
+                !parseVersionString(version, "OpenGL ES ",    m_settings.majorVersion, m_settings.minorVersion) &&
+                !parseVersionString(version, "",              m_settings.majorVersion, m_settings.minorVersion))
```
Foaly: Sorry for the minor remark, but logical ors || make more sense here, don't they?

eXpl0it3r: !A && !B <=> A || B

Since you want to check that these are not the values you expect, not-anding seems more logical to me.

kingnobody8:
Looks good. Tested on Windows, Android, Mac, and iOS, all successfully.
Just for reference, only the apple versions will do the "old way" of getting the gl version. Windows and android do the "new way".

PR #1394, opened 2018-03-20
Added a VertexBuffer implementation to all Drawables that were rendered via VertexArrays

binary1248:
As promised...

__DISCLAIMER: Save important work before testing. Your PC might not like this kind of GPU stress.__

Test with:
```cpp
#include <SFML/Graphics.hpp>
#include <sstream>
#include <cstdlib>

#define TESTTEXT

int main()
{
    sf::RenderWindow window(sf::VideoMode(800, 800), "Test");
    window.setVerticalSyncEnabled(false);

#if defined(TESTTEXT)

    sf::Font font;
        return EXIT_FAILURE;

    sf::Text drawable;
    drawable.setFont(font);

    std::string str;

    for (int i = 0; i < 1024 * 1024; ++i)
        str += (std::rand() % 0x80);

    drawable.setString(str);

#elif defined(TESTSHAPE)

    sf::CircleShape drawable(1, 4096 * 4096);

#elif defined(TESTSPRITE)

    sf::Image image;
    image.create(1, 1, sf::Color::White);

    sf::Texture texture;
    texture.loadFromImage(image);

    sf::Sprite drawable(texture);

#endif

    sf::Clock clock;
    std::ostringstream sstr;

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
        }

        window.clear();

#if defined(TESTSPRITE)

        for (int i = 0; i < 1024 * 256; ++i)
            window.draw(drawable);

#else

        window.draw(drawable);

#endif

        window.display();

        sstr.str("");
        sstr << "Test -- Frame: " << clock.restart().asSeconds() << " sec";

        window.setTitle(sstr.str());
    }
}
```
Performance improvement with `sf::Text` and any `sf::Shape` is so large that mentioning the values doesn't even make any sense. With `sf::Sprite` in order to notice any difference I had to push draw call count to something so stupid that the CPU overhead of the VBO binding/unbinding ends up dominating and making the VBO implementation slower, take with a grain of salt, no sane person would do something like this.

Test using this code by going into VertexBuffer.cpp and temporarily returning false from `isAvailable()`.

Results from real world tests are of course highly appreciated.

### DIFF
--- a/include/SFML/Graphics/Shape.hpp
+++ b/include/SFML/Graphics/Shape.hpp
@@ -32,6 +32,7 @@
 #include <SFML/Graphics/Drawable.hpp>
 #include <SFML/Graphics/Transformable.hpp>
 #include <SFML/Graphics/VertexArray.hpp>
+#include <SFML/Graphics/VertexBuffer.hpp>
 #include <SFML/System/Vector2.hpp>
 
 
@@ -305,15 +306,17 @@ private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    const Texture* m_texture;          ///< Texture of the shape
-    IntRect        m_textureRect;      ///< Rectangle defining the area of the source texture to display
-    Color          m_fillColor;        ///< Fill color
-    Color          m_outlineColor;     ///< Outline color
-    float          m_outlineThickness; ///< Thickness of the shape's outline
-    VertexArray    m_vertices;         ///< Vertex array containing the fill geometry
-    VertexArray    m_outlineVertices;  ///< Vertex array containing the outline geometry
-    FloatRect      m_insideBounds;     ///< Bounding rectangle of the inside (fill)
-    FloatRect      m_bounds;           ///< Bounding rectangle of the whole shape (outline + fill)
+    const Texture* m_texture;               ///< Texture of the shape
+    IntRect        m_textureRect;           ///< Rectangle defining the area of the source texture to display
+    Color          m_fillColor;             ///< Fill color
+    Color          m_outlineColor;          ///< Outline color
+    float          m_outlineThickness;      ///< Thickness of the shape's outline
+    VertexArray    m_vertices;              ///< Vertex array containing the fill geometry
+    VertexArray    m_outlineVertices;       ///< Vertex array containing the outline geometry
+    VertexBuffer   m_verticesBuffer;        ///< Vertex buffer containing the fill geometry
+    VertexBuffer   m_outlineVerticesBuffer; ///< Vertex buffer containing the outline geometry
+    FloatRect      m_insideBounds;          ///< Bounding rectangle of the inside (fill)
+    FloatRect      m_bounds;                ///< Bounding rectangle of the whole shape (outline + fill)
 };
 
 } // namespace sf
--- a/include/SFML/Graphics/Sprite.hpp
+++ b/include/SFML/Graphics/Sprite.hpp
@@ -33,6 +33,7 @@
 #include <SFML/Graphics/Transformable.hpp>
 #include <SFML/Graphics/Vertex.hpp>
 #include <SFML/Graphics/Rect.hpp>
+#include <SFML/Graphics/VertexBuffer.hpp>
 
 
 namespace sf
@@ -215,9 +216,10 @@ private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    Vertex         m_vertices[4]; ///< Vertices defining the sprite's geometry
-    const Texture* m_texture;     ///< Texture of the sprite
-    IntRect        m_textureRect; ///< Rectangle defining the area of the source texture to display
+    Vertex         m_vertices[4];    ///< Vertices defining the sprite's geometry
+    VertexBuffer   m_verticesBuffer; ///< Vertex buffer containing the sprite's geometry
+    const Texture* m_texture;        ///< Texture of the sprite
+    IntRect        m_textureRect;    ///< Rectangle defining the area of the source texture to display
 };
 
 } // namespace sf
--- a/include/SFML/Graphics/Text.hpp
+++ b/include/SFML/Graphics/Text.hpp
@@ -34,6 +34,7 @@
 #include <SFML/Graphics/Font.hpp>
 #include <SFML/Graphics/Rect.hpp>
 #include <SFML/Graphics/VertexArray.hpp>
+#include <SFML/Graphics/VertexBuffer.hpp>
 #include <SFML/System/String.hpp>
 #include <string>
 #include <vector>
@@ -435,20 +436,22 @@ private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    String              m_string;              ///< String to display
-    const Font*         m_font;                ///< Font used to display the string
-    unsigned int        m_characterSize;       ///< Base size of characters, in pixels
-    float               m_letterSpacingFactor; ///< Spacing factor between letters
-    float               m_lineSpacingFactor;   ///< Spacing factor between lines
-    Uint32              m_style;               ///< Text style (see Style enum)
-    Color               m_fillColor;           ///< Text fill color
-    Color               m_outlineColor;        ///< Text outline color
-    float               m_outlineThickness;    ///< Thickness of the text's outline
-    mutable VertexArray m_vertices;            ///< Vertex array containing the fill geometry
-    mutable VertexArray m_outlineVertices;     ///< Vertex array containing the outline geometry
-    mutable FloatRect   m_bounds;              ///< Bounding rectangle of the text (in local coordinates)
-    mutable bool        m_geometryNeedUpdate;  ///< Does the geometry need to be recomputed?
-    mutable Uint64      m_fontTextureId;       ///< The font texture id
+    String               m_string;                ///< String to display
+    const Font*          m_font;                  ///< Font used to display the string
+    unsigned int         m_characterSize;         ///< Base size of characters, in pixels
+    float                m_letterSpacingFactor;   ///< Spacing factor between letters
+    float                m_lineSpacingFactor;     ///< Spacing factor between lines
+    Uint32               m_style;                 ///< Text style (see Style enum)
+    Color                m_fillColor;             ///< Text fill color
+    Color                m_outlineColor;          ///< Text outline color
+    float                m_outlineThickness;      ///< Thickness of the text's outline
+    mutable VertexArray  m_vertices;              ///< Vertex array containing the fill geometry
+    mutable VertexArray  m_outlineVertices;       ///< Vertex array containing the outline geometry
+    mutable VertexBuffer m_verticesBuffer;        ///< Vertex buffer containing the fill geometry
+    mutable VertexBuffer m_outlineVerticesBuffer; ///< Vertex buffer containing the outline geometry
+    mutable FloatRect    m_bounds;                ///< Bounding rectangle of the text (in local coordinates)
+    mutable bool         m_geometryNeedUpdate;    ///< Does the geometry need to be recomputed?
+    mutable Uint64       m_fontTextureId;         ///< The font texture id
 };
 
 } // namespace sf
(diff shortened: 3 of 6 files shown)


LaurentGomila:
I don't think this kind of test is fair, as you're only drawing static geometry, which is the optimal case for VBOs (no CPU - GPU transfer). SFML drawables have visual attributes that can change dynamically (color, geometry size, texture coordinates), possibly every frame, and that require an update of the vertices. And, if I remember correctly my tests a few years ago, that's precisely where VBOs were not better than VAs.

I also wonder how good it is to duplicate all vertices, in drawable classes, ie. it's not either VA or VBO, it's always VA (in system RAM) + possibly VBO (in GPU RAM) if it's supported. So that's twice the memory usage.
I understand that it's better to keep vertex data in system RAM for updates, rather than moving it back from GPU to CPU, but can we afford the wasted memory? And what about geometry that is always reconstructed from scratch, like sf::Text? Can't we throw away the vertex buffer and put vertices directly into the VBO when we rebuild it?

binary1248:
I don't think this test is unfair... yes it is synthetic, but the frequency of `Drawable` updates correlates with the traces that were sent to me when debugging #1379. In short, I think you are overestimating the update frequency in real-world applications, which is why I am relying on real-world tests to gauge this PR.

Regarding memory usage, yes... I was also asking myself if it's a good idea to keep the VA data around after uploading it. In order to reliably de-allocate the VA buffer you would have to completely destroy the VA itself, a `clear()` is not sufficient because `std::vector` doesn't free its reserved RAM unless you do `shrink_to_fit()` (C++11) or something equivalent. This would necessitate making the VA member into a pointer which adds a dereference every time geometry has to be accessed. Not only that, but the user might even feel the performance spike every time they need to rewrite the VA because of the allocation that would be necessary if we were hellbent on getting rid of the buffer all the time. The fact that `std::vector` keeps hold of its reserved memory is a feature/optimization, a memory usage vs. CPU time trade-off. I don't think we should worry ourselves with this too much. Almost all systems will have much more system RAM that graphics RAM.

LaurentGomila:
You're right about memory if we stick to building the geometry first then uploading it. But what if we directly fill the VBO in GPU memory, using glMapBuffer? Is it considered inefficient?

I disagree about people always drawing static data with SFML, but could you first run the same test with something that changes every frame? I don't want to start arguing if it's not worth it 😬

binary1248:
So...

I started to implement code that made use of `glMapBuffer`. During my initial testing, I noticed that performance was significantly worse than just uploading via `glBufferSubData`. This is nothing new, I've already noticed this discrepancy over the last years when experimenting with buffer streaming. I always thought it was a driver anomaly, but if it still exists now, then I assume it has deeper rooted explanations.

The thing is, either way we submit geometry data, via `glBufferSubData` or implicitly within `glDrawArrays`, the driver will have to memcpy our geometry data into a buffer allocated within kernel space and instruct the GPU to pull the data over via a DMA transfer. You will never get better than 2 transfers, and they will happen regardless which way around you go. The big difference between `glMapBuffer` and the other 2 methods is that `glMapBuffer` takes the detour through virtual memory. When you map to a user-space visible address, the driver has to assume the worst. Unlike with the other 2 methods, you don't have to write to the memory in a contiguous operation. You can jump back and forth as you wish. When the time comes to transfer the new data down to the GPU, the poor driver has to figure out which bytes you touched and only transfer those. How it does this is beyond my understanding, needless to say, it has to mark which bytes you touched when you write to them. This could be the primary reason why `glMapBuffer` performance sucks so bad that is has been largely replaced by `glMapBufferRange` and the other more advanced buffer APIs.

In terms of performance, if we want to limit ourselves to ARB_vertex_buffer_object, `glBufferSubData` is the only sane way to go. Oh... and then there is the part about OpenGL ES 1.1 not even supporting `glMapBuffer`. 😁

Because I had to write 2 upload paths (with and without `glMapBuffer` support), the CPU overhead also increased a bit and the code became much more complicated. I eventually just threw it all out. 😛

Now... to the streaming performance. To be honest, I wasn't expecting any significant difference, and also didn't notice any in the end. As explained above, the driver ends up having to do the same amount of work no matter which way around you go. I don't know how you tested VBO streaming performance all those years ago, but there are a few caveats you have to be aware of to reach optimal performance. I've already implemented the necessary optimizations in the merged `VertexBuffer` implementation to make sure streaming performance is on par with the implicit `glDrawArray` performance. If you want to read more, you can search the internet or just start [here](https://www.khronos.org/opengl/wiki/Buffer_Object_Streaming).

We also mustn't forget that per-frame data streaming is the absolute worst case, and yes it can happen in certain use cases, but forfeiting realistic performance improvements when data doesn't have to be transferred per-frame doesn't make sense to me. You don't have to look any further than our own examples. In no example is an `sf::Text` modified per-frame, that would just make it obnoxious to read which kind of defeats its purpose. The OpenGL example also makes use of a permanent background `sf::Sprite` that never changes. You might think the Edge effect in the Shader example modifies geometry data every frame, but even that is not the case. Modifying anything via `sf::Transformable` only has an effect on the modelview matrix, not the geometry. This is also how objects are transformed in real-world OpenGL-only applications. The geometry data (which can be huge, e.g. 1000000 triangles) stays static, while the matrices used for e.g. kinematics are constantly updated to make characters move.

TL;DR: I still think you are trying to optimize the implementation for a problem that does not occur that frequently if at all in the wild. Any well-informed developer keen on performance would not get themselves into the situation you are afraid of in the first place. This is a problem that has been solved for quite some time now. We're just still stuck in the 1999 mindset because of our current implementation.

The final nail in the coffin is the fact that we will have to provide a VBO implementation for all Drawables anyway if we are going to stay forward compatible with future OpenGL, this includes one day providing shader support for mobile platforms. Client-side arrays and shaders are 2 features that are mutually exclusive in OpenGL ES.

Here is the code I tested streaming performance with, like I said, I don't think this is realistic in any way:
```cpp
#include <SFML/Graphics.hpp>
#include <sstream>
#include <cstdlib>

#define TESTTEXT

int main()
{
    sf::RenderWindow window(sf::VideoMode(800, 800), "Test");
    window.setVerticalSyncEnabled(false);

#if defined(TESTTEXT)

    sf::Font font;
        return EXIT_FAILURE;

    sf::Text drawable;
    drawable.setFont(font);

    std::string str;

    for (int i = 0; i < 1024 * 1024; ++i)
        str += (std::rand() % 0x80);

    drawable.setString(str);
    drawable.setOutlineColor(sf::Color::Red);
    drawable.setOutlineThickness(10.f);

#elif defined(TESTSHAPE)

    sf::CircleShape drawable(100.f, 1024 * 1024);
    drawable.setOutlineColor(sf::Color::Red);
    drawable.setOutlineThickness(10.f);

#elif defined(TESTSPRITE)

    sf::Image image;
    image.create(1, 1, sf::Color::White);

    sf::Texture texture;
    texture.loadFromImage(image);

    sf::Sprite drawable[1024];

    for (int i = 0; i < 1024; ++i)
        drawable[i].setTexture(texture);

#endif

    sf::Clock clock;
    std::ostringstream sstr;

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
        }

        window.clear();

#if defined(TESTSPRITE)

        for (int i = 0; i < 1024; ++i)
        {
            drawable[i].setColor(sf::Color::White);
            window.draw(drawable[i]);
        }

#elif defined(TESTSHAPE)

        drawable.setRadius(100.f);
        window.draw(drawable);

#elif defined(TESTTEXT)

        drawable.setStyle(sf::Text::Bold);
        drawable.setStyle(sf::Text::Regular);
        window.draw(drawable);

#endif

        window.display();

        sstr.str("");
        sstr << "Test -- Frame: " << clock.restart().asSeconds() << " sec";

        window.setTitle(sstr.str());
    }
}
```

LaurentGomila:
Thanks for the detailed explanation. I totally agree with you, at the driver level it should not be worse. I just wanted to be sure that it was really the case.

And yes, catching up with modern OpenGL is also a very good thing for the future.

The benchmark is not useless, it ensures that in the worst case we still have decent performances, and no regression compared to VA.

Now my only concern is all that boilerplate code in drawables (maintaining both VA and VBO, always testing if VBO are supported, ...). Maybe we could design a class that abstracts this? If VertexBuffer held the vertex data itself, it could manage automatically VBO updates, and switching to VA if they are not supported. What do you think?

binary1248:
I wouldn't over-generalize this early.

1. There are already people happily making use of `sf::VertexBuffer` and they don't expect/want it to internally store a copy of its data in system RAM as well. These users would only use `sf::VertexBuffer` if it was supported and they know what they are dealing with. When we use it in our Drawable implementation, it is only an optimization. In the case VBOs aren't available, we still have to fall back to `sf::VertexArray`. I wouldn't try to combine their feature sets into a single class just to save a bit of typing.

2. Exposing too much of the implementation through the public API makes it harder to optimize the implementation in the future. This drawing implementation only looks this way right now because we only have an OpenGL 1.2 implementation. When (not if) we get around to writing e.g. an OpenGL 4.6 renderer, it will look completely different. Nobody said that each Drawable even has to have its own buffer object. It is a well known fact from AZDO presentations that manually allocating fragments from a single persistently mapped buffer can lead to huge performance gains. The external API the users see will stay the same, and when they call `window.draw()` they have a certain behavioural expectation. How the data is internally managed and how draw calls are submitted are all implementation details that should remain hidden in order to allow for future optimizations. I would even go so far as to say that exposing the fact that we use `sf::VertexArray` in our Drawables is already too much. We would end up breaking ABI compatibility if we got rid of it once we deem it no longer necessary.

I've had the idea for a while, and it is currently blocked on the non-legacy shader discussion somewhere on the forum. We should turn all outward facing sfml-graphics classes into proxies. During runtime, the library can select the best OpenGL implementation for the system and use it as the concrete implementation of the proxy classes. We would just have to move the current implementation into something like GL12. We would then also have GL21, GL33 and GL46. Like this, we can work our magic much easier than if we had to worry about the public facing classes all the time.

texus:
I wanted to try using VertexBuffer in my code and I found that it actually lead to a drop in performance. So I decided to try this PR and check the performance when SFML uses the VertexBuffer internally. The numbers below aren't very accurate (e.g. if it says 0.2 it migh mean that it fluctuated between 0.18 and 0.24) but the difference between the two versions is pretty clear.

The following table shows the amount of seconds spend per frame (using the code from the first post)

|         | TESTTEXT | TESTSHAPE | TESTSPRITE |
|---------|----------|-----------|------------|
| master  |   0.08   |    0.20   |    0.21    |
| this PR |   0.01   |    0.27   |    0.62    |

As you can see I'm getting a performance boost for drawing text, but in the other cases (and even if I try to use VertexBuffer from the master branch directly in my code) the code with VertexBuffer is slower.

Tested on Manjaro Linux laptop with Intel HD Graphics 620 (i5-7200U CPU).

binary1248:
Using `sf::VertexBuffer` without adjusting the code to take advantage of it will lead to performance degradation. This will always be the case. It can be more or less dropped in place of a `sf::VertexArray`, but this won't always guarantee a performance gain without further adjustments.

`sf::VertexBuffer` only really starts to shine at larger amounts of geometry data. This is the reason why `sf::Text` benefits much more from it than the other 2 cases you tested. You didn't provide any test code, but I suspect your shape test consisted of low-vertex-count shapes. Since sprites only consist of 4 vertices it makes the least sense performance-wise to use a `sf::VertexBuffer` for it.

The primary reason why everything has to eventually internally support `sf::VertexBuffer` isn't even performance. It is forwards compatibility. We are shutting ourselves out from using any of the modern APIs by staying in this client-side-array world. There actually isn't even any naturally inherent disadvantage to using vertex buffers everywhere. What really drags performance down here is the combination of vertex buffers and the typical SFML usage pattern which doesn't fit well together.

If an SFML user gave some effort to first reduce the number of invoked draw calls by doing a bit of  batching into larger data blobs in their own code they will see significant performance gains when switching over to `sf::VertexBuffer` from `sf::VertexArray`. This doesn't happen automagically.

eXpl0it3r:
If there are no further concerns I'd like to merge this soon. 🙂

eigenbom:
As I pointed out in a [forum post](https://en.sfml-dev.org/forums/index.php?topic=25228.0), this PR leads to a non-trivial performance drop. Can we at least get a flag to disable it, or delay it until sfml3?

binary1248:
See [my reply](https://en.sfml-dev.org/forums/index.php?topic=25228.msg167409#msg167409).

PR #1395, opened 2018-03-21
Renamed a few key codes

mantognini:
Deprecated BackSlash, BackSpace, SemiColon, Dash and introduced Backslash, Backspace, Semicolon, Hyphen in a retrocompatible way.

As discussed in #1235. Note that I kept Subtract instead of merging Dash and Subtract in Minus.


Foaly:
Looks good! I know we can't mark an enum entry as deprecated in a cross-compiler way, but would it maybe be worth it to put an assert at the point where they are converted to inform the user he is using deprecated values?

eXpl0it3r:
I assume by assert you don't mean `assert()` right?

Foaly:
Well I did, but maybe thats a bit much. A console output might be better. Otherwise as a library user there is no warning that the old values have been deprecated.

mantognini:
An assertion is definitely too strong here - crashing the client app for that wouldn't be appreciated. The value is deprecated, not removed nor restricted.

Outputting something would be better, but we would end up spamming the user. Also, we wouldn't be able to catch all usages. 

I think for such small deprecation (it's just an alias in the end), letting the user read the changelog should be enough.

Foaly:
Well it would only crash in debug mode, which nobody sane would ship. So it would happen on developer machines where it notifies exactly the person we want to reach. I am actually all for deprecating APIs [like this](https://twitter.com/johnregehr/status/920691341738123264). I think we should have a console output the very least. The spamming you mention happens only as long as the user decides to keep using outdated parts of the code.

But I know we have different opinions on how to handle things like these, so in order not to block merging I will keep quiet, if the team decides against it ¯\\\_(ツ)\_/¯

mantognini:
That's an interesting example you've brought. I'm wondering however if
people will prefer not to upgrade deps with that extreme choice.

Personally, I'm open for suggestions, but as a separate PR. In any case, we
will get standard deprecation mechanism, and therefore compile time
warnings, for v3 I think. `[[deprecated]]` or something, right?

eliasdaler:
Yes, you can do this in C++14:

```enum class Test
{ 
    x [[deprecated]],
    y,
    z
};```

No good way to do it in C++03 though. I think we'll be fine with a comment for now.

PR #1398, opened 2018-03-23
GlyphQuad x position quantized to pixels

oomek:
To avoid subpixel positioning of small glyphs when character spacing is not 1.0 I propose the following changes. Subpixel positioning does only affect small characters. Quads x positions for large character sizes are quantized anyway.

### DIFF
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -727,6 +727,7 @@ IntRect Font::findGlyphRect(Page& page, unsigned int width, unsigned int height)
                 // Make the texture 2 times bigger
                 Texture newTexture;
                 newTexture.create(textureWidth * 2, textureHeight * 2);
+                newTexture.setSmooth(true);
                 newTexture.update(page.texture);
                 page.texture.swap(newTexture);
             }
--- a/src/SFML/Graphics/Text.cpp
+++ b/src/SFML/Graphics/Text.cpp
@@ -50,15 +50,17 @@ namespace
     // Add a glyph quad to the vertex array
     void addGlyphQuad(sf::VertexArray& vertices, sf::Vector2f position, const sf::Color& color, const sf::Glyph& glyph, float italicShear, float outlineThickness = 0)
     {
-        float left   = glyph.bounds.left;
-        float top    = glyph.bounds.top;
-        float right  = glyph.bounds.left + glyph.bounds.width;
-        float bottom = glyph.bounds.top  + glyph.bounds.height;
+        float padding = 1.0;
 
-        float u1 = static_cast<float>(glyph.textureRect.left);
-        float v1 = static_cast<float>(glyph.textureRect.top);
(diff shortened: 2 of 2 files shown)


LaurentGomila:
What exactly does that solve?

oomek:
sfml quantised font position

LaurentGomila:
Indeed, that looks better in this case, but here we are dealing with scene coordinates, not screen coordinates. What if the user, for some reason, wants to work with small coordinates (either with the view or the text scale), where one scene unit maps to several screen pixels?

oomek:
In the case of coordinates not mapping exactly to screen coordinates even letter spacing factor = 1.0 will look wrong.

Regarding to scrolling it’s the same thing. The font larger than 38 ( i think this value is hardcoded in the specific font when the glyphs start snapping to pixels ) will start to crawl. I overcome this by rendering the font to rendertexture and animating it.

oomek:
There is one other option I'm currently using a vertex shader for text like the one below:
```
uniform float width; // width in pixels of the rendertarger

void main()
{
    // transform the vertex position
    gl_Position = gl_ModelViewProjectionMatrix * gl_Vertex;
	gl_Position.x = floor(gl_Position.x * width / 2.0) / (width / 2.0);

    // transform the texture coordinates
    gl_TexCoord[0] = gl_TextureMatrix[0] * gl_MultiTexCoord0;

    // forward the vertex color
    gl_FrontColor = gl_Color;
}
```

oomek:
We may have an additional problem. The glyph quads may need some additional padding like 1 pixel on each side. When you scroll small font the sides of fonts are being cut off, actually it happens on all font sizes, but it’s more visible on the small ones. It’s, not related to my commit, it’s apparent on master also. Will post some video with zoom to illustrate that isssue.

oomek:
Sorry for the delay.
Here is the link to the video showing what happens to the glyph's boundaries when it's scrolled with sub-pixel speed. 

https://youtu.be/s2NBOKEeAGE

Ideally we could extend each letter's quad by 1 pixel ( 0.5 on each side as the cropping happens halfway through the transition to the next pixel), But I'm not sure yet what would be the best approach. It needs to be coordinate space agnostic and obviously cannot alter the rendering in any way.

oomek:
Here is another video showing the quads boundaries.

https://youtu.be/MljCptuOQw4

As you can see the quads are to small to accommodate the glyphs during subpixel scrolling.

LaurentGomila:
That's interesting, thanks for taking the time to make those examples.

oomek:
So what do you think? Is it something that can be controlled outside the FreeType library?
Theoreticaly it can be done in the shader. Extend the quad by 1 pixel and shrink the uv of the glyph by 1 pixel, but it's rather a workaround than a solution.

eXpl0it3r:
What's the difference between this PR and #1452, @oomek?

oomek:
I don't know what happened, but the previous commit was overriden by this one. Deleting. It's not relevant anyway as explained above. This functionality can be easily achieved with a shader.

PR #1400, opened 2018-03-29
Initial X11 scancode implementation

eliasdaler:
**SFML Tasks**

- [x] Agree on code structure (see below)
- [x] Adjust code according to the review comments
- [ ] Test ([use this code for it](https://github.com/mantognini/SFML-Test-Events/blob/scancode/Test%20Events/main.cpp) and switch your keyboard layout to non-US for harder cases)
- [ ] Merge into the feature/scancode branch (see #1235)

---

Hello everyone! This is my first PR and it was pretty hard to get this working! Keycodes, KeySyms, scancodes, SFML's scancodes - it's all different and sometimes it's hard to convert one to another.
A lot of the code was based on [GLFW implementation](https://github.com/glfw/glfw/blob/master/src/x11_init.c) because it deals with very hard cases.

I've introduced X11InputManager singleton, because someone needs to hold mappings between keycodes and SFML's scancodes. If someone comes up with a better alternative - I'll be happy to remake it. :)

---
**getDescription** was pretty hard to implement because there's no easy way to convert keycode or keysym to unicode. One way of doing it is using huge lookup tables like [this one](http://www.cl.cam.ac.uk/~mgk25/ucs/keysym2ucs.c). But thankfully, we have Xutf8LookupString now. But we need to create a fake event for it to work... which works, but isn't pretty.
And one more thing: some keys like Tab, Enter and Space produce unwanted results - they return " ", "\n" or "\t", so they have to be accounted for. It's even worse for Delete or Backspace, but it's the same thing.

---

There'll probably be a lot of things which need to be fixed with my code - I haven't done much X11 programming, so maybe I've made some mistakes - I'll be happy to fix them.


mantognini:
Fantastic! Looking forward testers feedback.

Feel free to split files in a meaningful manner. I'll probably end up doing something similar for macOS. ;-)

binary1248:
Is the singleton really necessary? We already have a reference counted `Display` in Display.cpp. You can initialize the stuff in `OpenDisplay`, deinitialize the stuff in `CloseDisplay` and convert all those methods to static methods. I would also give the XInputManager its own `XIC`. Something tells me that sharing it with some random `Window` (and even changing it whenever new `Window`s are created) won't work out so well...

eliasdaler:
Fair enough. But where should I store the mapping between the keycodes and scancodes?

binary1248:
In an anonymous namespace at the top of XInputManager.cpp like is done in all the other files.

eliasdaler:
So, I did tons of works on this:
* X11InputManager is not KeyboardImpl as it deals with keyboard only not all input, and we already have JoysticImpl, so it's similar
* We don't need a singleton anymore - everything is static
* I've made a script which generated a giant switch-case which maps KeySym to Unicode from [this file](https://www.cl.cam.ac.uk/~mgk25/ucs/keysyms.txt). Sadly, XLib doesn't provide a way to do it without such mappings. I can share the script I made to generate the function if anyone is interested. 

---

Okay, some of the namings might not be perfect, so feel free to give feedback about how to improve them.
And I'll see if I can make less conversions between KeySyms, Keycodes, sf::Keyboard::Key and sf::Keyboard::Scancode. But right now the implementation looks a lot better!

mantognini:
> The only thing I still don't like is that I still use "0" as a null KeyCode. Is it okay? For null KeySyms there's NoSymbol defined, but I've found nothing corresponding to KeyCodes.

Maybe define a local constant? This way you can refer to libX11's XKeysymToKeycode code in one place.

Let us know when you are happy with the code so we can review/test this. :)

eliasdaler:
@mantognini defined `NullKeyCode` constant.

I'm quite happy with the design now. And everything seems to work. I only tested on QWERTY layouts (but changed by layout with setxkbmap), so it'll be perfect for someone with other layout to test it.

I'm not 100% on the naming of all the stuff, so it could be improved a lot, if someone comes up with better names.

eliasdaler:
@Bromeon @texus 
Can you please take a look at this and test it if you have the time?

@binary1248, what do you think of the new code structure?

texus:
I tested all keys on my belgian azerty keyboard.
The only mistake I found is that Return and Enter are reversed, the main return key generates ScanEnter while the one on the keypad generates ScanReturn.

Although you could call the main key either Return or Enter, the one on the keypad is always an Enter key (it usually even has "Enter" written on it). In practice they do the same thing, but using ScanReturn for the one on the keypad seems wrong to me.

From [wikipedia](https://en.wikipedia.org/wiki/Enter_key#Differences_between_Enter_and_Return): `On some keyboard layouts, the return and enter key are two different keys, an artifact of the differing handling of newlines by different operating systems. As an example, on the Macintosh, the return key is the usual key, while the enter key is positioned at the lower right of the numeric key pad.`

I personally would call the return key ScanReturn and the keypad enter ScanKeypadEnter. In SDL they are called SDL_SCANCODE_RETURN and SDL_SCANCODE_KP_ENTER.

I also noticed that with one key I got the correct scancode but sometimes also ScanUnknown, but `xev` also shows "keycode 0" in this case, so this is not a problem with the code. The Fn and PrtScr keys didn't generate any scancode, but again this corresponds to the `xev` output.

eliasdaler:
Thanks a lot for testing! 

At first I made a commit to change Enter and Return, but then I realized that I've made a mistake. Right now `ScanEnter` = keyboard enter and `ScanReturn` = numpad enter. That's how @mantognini wrote it for now.

As for the naming - `getDescription` now returns "Enter" and "Enter (Numpad)". The names of scancodes should be discussed in #1235 and it seems like we may agree to have `ScanEnter` for keyboard "Enter" and `ScanNumpadEnter` for numpad "Enter". It's still not 100%, so you should probably post your point there!

eliasdaler:
That's a lot of feedback, thanks a lot! 
And I also found a bug, "Escape" wasn't shown correctly. Fixed it in the same commit.

For now, I don't know what to do with fallthrough cases. Should I just place return statements here?

binary1248:
> For now, I don't know what to do with fallthrough cases. Should I just place return statements here?

Yes.

eliasdaler:
Done!

src/SFML/Window/Unix/KeySymToUnicodeMapping.cpp
```diff
@@ -0,0 +1,1399 @@
...
+        case 0x04b1: return 0x30a2; // kana_A
+        case 0x04a8: return 0x30a3; // kana_i
+        case 0x04b2: return 0x30a4; // kana_I
+        case 0x04a9: return 0x30a5; // kana_u
+        case 0x04b3: return 0x30a6; // kana_U
+        case 0x04aa: return 0x30a7; // kana_e
+        case 0x04b4: return 0x30a8; // kana_E
+        case 0x04ab: return 0x30a9; // kana_o
+        case 0x04b5: return 0x30aa; // kana_O
+        case 0x04b6: return
```
binary1248: GitHub says there is no newline at the end of this file.

src/SFML/Window/Unix/KeyboardImpl.cpp
```diff
@@ -0,0 +1,586 @@
...
+#include <X11/keysym.h>
+#include <X11/Xlib.h>
+#include <X11/XKBlib.h>
+#include <cstring> // strcmp
+
+namespace
+{
+
+const KeyCode NullKeyCode = 0;
+KeyCode scancodeToKeycode[sf::Keyboard::ScanCodeCount];     ///< Mapping of SFML scancode to X11 KeyCode
```
binary1248: Is the space before the comment meant for alignment with the line below? I can't tell what is the intention here.

src/SFML/Window/Unix/KeyboardImpl.cpp
```diff
@@ -0,0 +1,586 @@
...
+sf::Keyboard::Scancode keycodeToScancode[256]; ///< Mapping of X11 KeyCode to SFML scancode
+bool isMappingInitialized = false;
+
+////////////////////////////////////////////////////////////
+bool isValidKeycode(KeyCode keycode)
+{
+    // Valid key code range is [8,255], according to the Xlib manual
+    return keycode >= 8 || keycode <= 255;
```
binary1248: Might not be necessary, but brackets around the comparisons makes it safer and maybe even a bit more readable.

src/SFML/Window/Unix/KeyboardImpl.cpp
```diff
@@ -0,0 +1,586 @@
...
+    XkbGetNames(display, XkbKeyNamesMask, desc);
+
+    sf::Keyboard::Scancode sc;
+
+    for (int keycode = desc->min_key_code; keycode <= desc->max_key_code; ++keycode)
+    {
+        std::memcpy(name, desc->names->keys[keycode].name, XkbKeyNameLength);
+        name[XkbKeyNameLength] = '\0';
+
+        if (strcmp(name, "TLDE") == 0)      sc = sf::Keyboard::ScanGraveAccent;
```
binary1248: I would align the `(strcmp(name, "TLDE") == 0)` with the lines below just for that extra touch.

src/SFML/Window/Unix/KeyboardImpl.cpp
```diff
@@ -0,0 +1,586 @@
...
+        else if (strcmp(name, "AE04") == 0) sc = sf::Keyboard::ScanNum4;
+        else if (strcmp(name, "AE05") == 0) sc = sf::Keyboard::ScanNum5;
+        else if (strcmp(name, "AE06") == 0) sc = sf::Keyboard::ScanNum6;
+        else if (strcmp(name, "AE07") == 0) sc = sf::Keyboard::ScanNum7;
+        else if (strcmp(name, "AE08") == 0) sc = sf::Keyboard::ScanNum8;
+        else if (strcmp(name, "AE09") == 0) sc = sf::Keyboard::ScanNum9;
+        else if (strcmp(name, "AE10") == 0) sc = sf::Keyboard::ScanNum0;
+        else if (strcmp(name, "AE11") == 0) sc = sf::Keyboard::ScanDash;
+        else if (strcmp(name, "AE12") == 0) sc = sf::Keyboard::ScanEquals;
+        else if (strcmp(name, "TAB") == 0)  sc = sf::Keyboard::ScanTab;
```
binary1248: I would align the `) == 0)` with the other lines.

eliasdaler:
Agreed with everything. Done! :+1:

src/SFML/Window/Unix/KeySymToUnicodeMapping.cpp
HiImJulien: Is there a reason, why 'magic numbers' are used instead of constants?
At first glance most of them are defined in `X11/keysym.h`.

eliasdaler: Hmm, indeed. I must have only seen X11/keysym.h, but didn't take a look at keysymdef.h, where all these codes are.
Will redo this giant switch later. :)

eliasdaler:
@HiImJulien 

Looking back at why I didn't use the mapping already present in `<X11/keysym.h>` - it doesn't really correspond to correct and actual Unicode code points! See

* [This](https://stackoverflow.com/questions/8970098/how-to-map-a-x11-keysym-to-a-unicode-character)
* [And this](https://github.com/substack/node-keysym/blob/master/data/keysyms.txt)

... So this mapping is really necessary.

eXpl0it3r:
I've rebased, fixed my bad rebase mistakes and merged your branch onto `feature/scancode`, hope you don't mind. :wink: 
Let's iron out the rest on one branch, so refactorings are a bit easier.

PR #1403, opened 2018-04-05
Added GitHub issue and PR templates

eXpl0it3r:
[GitHub will use these templates](https://blog.github.com/2016-02-17-issue-and-pull-request-templates/) to pre-fill created issues and pull requests. This should help to get people to provide better information on issues or direct them to the forum when needed and hint people to provide example/test code for pull requests.

Also moved the CONTRIBUTING.md file into the .github directory.

This is really just a draft and I'd like to hear people's opinion on what we should add to these templates.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,51 @@
+Thanks a lot for making a contribution to SFML! 🙂
+
+Before you create the pull request, we ask you to check the follow boxes. (For small changes not everything needs to ticked, but the more the better!)
+
+* [ ] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before?
+* [ ] Does the code follow the SFML [Code Style Guide](https://www.sfml-dev.org/style.php)?
+* [ ] Have you provided some example/test code for your changes?
+* [ ] If you have additional steps which need to be performed list them as tasks!
+
+----
+
+## Description
+
+Please describe your pull request.
+
+This PR is related to the issue #
+
+## Tasks
+
+* [ ] Tested on Linux
+* [ ] Tested on Windows
+* [ ] Tested on macOS
+* [ ] Tested on iOS
+* [ ] Tested on Android
+
+## How to test this PR?
+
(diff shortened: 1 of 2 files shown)


dabbertorres:
What do you think about an explicit "code goes here" area in the issue template? Would make it a little more difficult to ignore/miss.

ie:

> Your \[minimal, complete, and verifiable example](https://stackoverflow.com/help/mcve):
> \```c++
> \<code goes here>
> \```

Could even go a step farther and include a bare-bones SFML C++ program (#include's, main(), Window, (empty) event loop, etc) for people to start from. Or, it could serve as an explicit example of a MCVE in the context of SFML, for those who may wish for an example?

mantognini:
I think it's a great start! :)

Maybe the `CONTRIBUTING.md` could stay visible at the root; it can be useful for people who cloned the repo as well.

I also like @dabbertorres's suggestion. For PR, this could also be added:
 - reference to issue #,
 - instruction & code for test with clear expectation of the outputs.
 - this list to keep track of testing progress:
   - [ ] tested on Linux
   - [ ] tested on Windows
   - [ ] tested on macOS
   - [ ] tested on iOS
   - [ ] tested on Android

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,12 @@
+Thanks a lot for making a contribution to SFML! 🙂
+
+Before you create the pull request, we ask you to check the follow boxes. (For small changes not everything needs to ticked, but the more the better!)
+
+* [ ] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before?
+* [ ] Does the code follow the SFML [Code Style Guide](https://www.sfml-dev.org/style.php)?
+* [ ] Have you provided some example/test code for your changes?
+* [ ] If you additional steps which need to be performed list them as tasks!
```
mantognini: *if you* **have** *additional steps [...]*

eXpl0it3r:
Updated the templates with your suggestions, see them both here:

* [Issue template](https://github.com/SFML/SFML/blob/4c4e961f391439a97d159f00ff6aa01397ffd961/.github/ISSUE_TEMPLATE.md)
* [PR template](https://github.com/SFML/SFML/blob/4c4e961f391439a97d159f00ff6aa01397ffd961/.github/PULL_REQUEST_TEMPLATE.md)

Let me know if there are further improvements that could be made.

Issue #1273, opened 2017-09-03
Update Doxygen CSS


pixelherodev:
![image](https://user-images.githubusercontent.com/10342579/30004796-bcdec752-90a3-11e7-8746-63b9f27b2de8.png) That shows up at the top of every page of the local documentation, both with package libsfml-doc, and the copy I built myself.

![image](https://user-images.githubusercontent.com/10342579/30004817-fb598d14-90a3-11e7-9051-b117744b4390.png)
If you look at the URL bars, you can tell one is from the Debian documentation browser (aka libsfml-doc package), and on the local copy.

I cleared the JS console and refreshed the page:

```
Loading failed for the <script> with source “http://localhost/cgi-bin/dwww/usr/share/doc/libsfml-doc/html/dynsections.js”.  modules.html:9
jQuery.Deferred exception: $(...).smartmenus is not a function initMenu@http://localhost/cgi-bin/dwww/usr/share/doc/libsfml-doc/html/menu.js:25:3
@http://localhost/cgi-bin/dwww/usr/share/doc/libsfml-doc/html/modules.html:23:3
mightThrow@http://localhost/cgi-bin/dwww/usr/share/doc/libsfml-doc/html/jquery.js:3570:21
resolve/</process<@http://localhost/cgi-bin/dwww/usr/share/doc/libsfml-doc/html/jquery.js:3638:12
 undefined  jquery.js:3846
TypeError: $(...).smartmenus is not a function  menu.js:25:3
```
I checked the folder and sure enough there is no file `dynsections.js`
I copied the file from https://www.sfml-dev.org/documentation/2.4.2/dynsections.js and the error messages shrank to
```
jQuery.Deferred exception: $(...).smartmenus is not a function initMenu@http://localhost/cgi-bin/dwww/usr/share/doc/libsfml-doc/html/menu.js:25:3
@http://localhost/cgi-bin/dwww/usr/share/doc/libsfml-doc/html/modules.html:23:3
mightThrow@http://localhost/cgi-bin/dwww/usr/share/doc/libsfml-doc/html/jquery.js:3570:21
resolve/</process<@http://localhost/cgi-bin/dwww/usr/share/doc/libsfml-doc/html/jquery.js:3638:12
 undefined  jquery.js:3846
TypeError: $(...).smartmenus is not a function  menu.js:25:3
```
And the page didn't change.

np511:
I just checked the console for the official site, and there was nothing there.

[...]

PR #1405, opened 2018-04-06
Update doxygen config and fix menu title on newer doxygen

Gravechapa:
https://github.com/SFML/SFML/issues/1273
screenshot_20180407_022123


Gravechapa:
I'm afraid it will work for doxygen 1.8.14+ only, because of this bug https://bugzilla.gnome.org/show_bug.cgi?id=776964

eXpl0it3r:
Superseded by #1428 included your change there. 🎉

Issue #1192, opened 2017-02-17
NvOptimusEnablement not exported for /SUBSYSTEM:CONSOLE


eigenbom:
When /SUBSYSTEM:CONSOLE is specified the exported symbols NvOptimusEnablement and AmdPowerXpressRequestHighPerformance in MainWin32.cpp aren't brought into the executable. I think this is because WinMain isn't referenced and so the entire .obj is ignored. Here's [a related article](https://blogs.msdn.microsoft.com/oldnewthing/20140321-00/?p=1433).

eigenbom:
Current temp fix is to use declare them in your own code and then use /FORCE:MULTIPLE to allow multiple declarations.

MarioLiebisch:
It's a compiler/optimization thing. Don't think there's anything we can do about this and in the end you'd just have to ensure to include the export on your own (as you do now).

On the flip side, you don't really have to link to **sfml-main** anyway; it's just a convenience thing. Just add your own little wrapper `WinMain()` calling `main()`. In the end it's just a trivial call if you don't care about parameters.

    #ifdef WIN32
    #include <Windows.h>
    int WINAPI wWinMain(HINSTANCE, HINSTANCE, PWSTR, int) {
        return main(0, NULL);
    }
    #endif

eigenbom:
Thanks for the reply.

According to the article I referenced, you can write a dummy function to force the .lib to always get linked in. 

If it isn't fixed, then I think having the NvOptimusEnablement stuff in there is misleading as it is only exported for /subsystem:windows builds. Maybe it could be somewhere else.

MarioLiebisch:
> According to the article I referenced, you can write a dummy function to force the .lib to always get linked in.

You could actually use `NvOptimusEnablement` in your code to achieve this. But I think the cleaner option would still be to implement the export and `WinMain` yourself.

> If it isn't fixed, then I think having the NvOptimusEnablement stuff in there is misleading as it is only exported for /subsystem:windows builds. Maybe it could be somewhere else.

It is somewhat. Unfortunately there's no other part of SFML that's always linked statically. It won't work as an export in a dll file. It has to happen in the executable itself. All solutions of a hidden use result in code you'd have to manually add either way.

---

[...]

PR #1406, opened 2018-04-07
Fix discrete GPU preference symbols being exported from the wrong place

binary1248:
Moved NvOptimusEnablement and AmdPowerXpressRequestHighPerformance out of sfml-main and into a macro the user can place in their own translation unit when they need it.

The current implementation suffers from a bunch of problems:
* It doesn't work when using the console subsystem, see #1192
* It can't work when linking all of SFML dynamically (this implies not linking sfml-main obviously)
* When linking sfml-main for portability, the user can't opt out of preferring the discrete GPU when they know they won't need the extra performance
* This "feature" was never documented, only people following the commit history or who browse through the code will know this exists
* Exporting symbols via `__declspec(dllexport)` from a static library is just weird in itself anyway

With this change, the user will have full control over when they want this preference to be made, and the symbols will always be exported from the right place: the final executable.


LaurentGomila:
I'm not sure if the top of Window.hpp is the right place to define this stuff. Maybe it would deserve its own header?

> this will actually show up in the documentation as well (might still be hard to find though)

I don't know if it works with macros too, but you can use `\relates Class` to attach it to the documentation page of `Class`.

MarioLiebisch:
[Haven't tried it myself, but this seems to be the intended approach.](https://stackoverflow.com/questions/4542626/documenting-macro-functions-in-c-with-doxygen)

binary1248:
@LaurentGomila @MarioLiebisch I moved the definition into its own file and added a paragraph with a reference to it in Window.hpp. `\relates` doesn't seem to work on macros.

eXpl0it3r:
Shouldn't header be included in the top-level Window.hpp as well, so you don't specifically have to include the header and define the macro?

binary1248:
I was thinking of providing this as a "convenience header" much like what `<SFML/OpenGL.hpp>` is now. But... if it makes more sense to automatically include this with `<SFML/Window.hpp>` then I can fix that too. What do the others think?

eXpl0it3r:
I mean, it doesn't really do any harm in directly including it and potentially reduces one extra step when explaining how to use it.

binary1248:
@eXpl0it3r Done.

binary1248:
@LaurentGomila Fixed.

Ceylo:
Shouldn't it be part of current Window module? ie in SFML/Window/GpuPreference.hpp
You include it automatically from <SFML/Window.hpp> and it changes the OpenGL setup in some way, which is currently the Window module responsibility with Context settings & such. Both make me think it should belong to the Window module.

binary1248:
Theoretically you can use this macro even in a console application in which you only make use of sfml-network even though it wouldn't make much sense. I wouldn't go so far as to say it changes any behaviour we should handle differently. This also expresses a preference, not a demand, we can't and shouldn't rely on it having any effect. When choosing a spot to place it, I figured this has a similar purpose as `<SFML/OpenGL.hpp>`, being more of a convenience than a necessity so I might as well place it in the same location.

LaurentGomila:
I didn't notice that before, and yes I admit that it looks confusing. Either make it part of the window module and include it in SFML/Window.hpp, or leave it as a standalone functionality and don't include it in any module by default.

Which solution to choose, I'm not sure 😬

binary1248:
Well... until now, the user could always assume that using something from an include file would imply having to link against that module as well. This works out for `<SFML/OpenGL.hpp>` as well since it is not actually part of any module and only a convenience header. The same can be said about `<SFML/GpuPreference.hpp>`. If the user is so inclined, they could even include both of these headers for mere convenience and use something else for their window management needs.

LaurentGomila:
Ok, so let's keep it separate from the Window module.

eXpl0it3r:
I didn't realize that it wasn't "part" of a module when I suggested to add it to `Window.hpp`. Should have taken a closer look. Since it really is unrelated to SFML window or graphics, I think it's okay to just have it similar to the OpenGL header.

eXpl0it3r:
Cna you revert it again, @binary1248? 😄

binary1248:
What do you mean with revert it again?

binary1248:
@MarioLiebisch Selection is misleading, because we don't select anything. We express a preference/hint. It is still up to the driver to pick a GPU for us. If the user chooses to do so, they can still override which GPU will be used. The opposite is also true, there is no way to express a preference for the iGPU. The hints only have an effect when they are exported and their values set to 1. Setting their values to 0 would be the same as not exporting them at all meaning "no preference". In that case it is up to the driver to guess which GPU to use.

binary1248:
Not reliable at all? An iGPU in a hybrid graphics setup could be the only available GPU in another system. OpenGL also doesn't provide any reliable way of enumerating adapters like D3D or Vulkan does, so that is also sort of out of the question.

binary1248:
@eXpl0it3r Done.

PR #1411, opened 2018-04-08
Modernize iOS toolchain file

Ceylo:
- Unused BUILD_ARM64 variable is removed
- Support for Xcode < 4.3 (!!) is removed
- Renamed all variables prefixed with CMAKE_
- Made the toolchain file much less verbose (you still get the SDK being used and you can check other values from CMake cache displayed in CMake GUI)

If CI build is ok this should be ok. There is no expected side effect on SFML behavior itself.

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -1,5 +1,10 @@
 include(CMakeParseArguments)
 
+# This little macro lets you set any Xcode specific property
+macro (sfml_set_xcode_property TARGET XCODE_PROPERTY XCODE_VALUE)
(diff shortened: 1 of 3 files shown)


JonnyPtn:
Looks good to me

eXpl0it3r:
Despite this being tagged as feature, do you think we could add this to SFML 2.5.1?

JonnyPtn:
I think that would be fine

eXpl0it3r:
Merged in ff011dc51dc47373b0d49b087f8ac9d3d2bc8b06

Issue #1401, opened 2018-03-31
Missing cursor types on macos


JonnyPtn:
I noticed there are some cursor types missing on macOS:

SizeAll
SizeTopLeftBottomRight
SizeBottomLeftTopRight

Based on some quick research, it seems they just aren't exposed or documented by apple, which is helpful.... but [this](https://stackoverflow.com/a/46635398) SO answer suggests a solution which appears to work for me, implemented [here](https://github.com/JonnyPtn/SFML/commit/ecc50f318712cd60db48391e9565b7a8f39bb7b7)

I should mention I'm not really experienced with objective-C at all, so apologies if that code is bad in any way... If not, I can PR whenever you'd like

MarioLiebisch:
Would probably be worth a try. What happens if you try to pass an invalid value (let's assume something is changed later)?

JonnyPtn:
If I change `[NSCursor _________]` to use a non-existent value it compiles fine but then get SIGABRT at runtime. 

Following the SO article again, checking `if ([NSCursor respondsToSelector:@selector(______)])` would stop the SIGABRT, and could just return false from `loadFromSystem()`, which would match the current behaviour?

I'm using high Sierra 10.13.4 and Xcode 9.2 (9C40b), for the record.

MarioLiebisch:
Sounds reasonable. My company Mac isn't up to date, so can't really check at the moment.

eXpl0it3r:
What happens on minimum supported macOS versions?

mantognini:
> but this SO answer suggests a solution which appears to work for me, implemented here

Interesting! This information wasn't available to me when #827 was being worked on.

> What happens on minimum supported macOS versions?

Nothing. This would be an addition based on non-documented API from Apple so we do not enforce users to have a system supporting those cursors. Instead, the users can only be certain at runtime that the cursor exists. The documentation should therefore be very clear on that.

Implementation wise, a patch along those lines (and proper formatting) would be good I think:

```cpp
// adding a private function to anonymous namespace
bool loadFromSelector(SEL selector, NSCursorRef& cursor) {
  if ([NSCursor respondsToSelector:selector]) { cursor = [NSCursor performSelector:selector]; return true; }
  else return false;
}

...
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wundeclared-selector"
switch (type) {
[...]

PR #1413, opened 2018-04-11
Implement extra cursors on macOS

JonnyPtn:
Addresses #1401

The move cursor doesn't look quite as I'd expect (It's basically just a white arrow) but I can't see any better looking alternatives.

### DIFF
--- a/src/SFML/Window/OSX/CursorImpl.mm
+++ b/src/SFML/Window/OSX/CursorImpl.mm
@@ -98,6 +98,10 @@ bool CursorImpl::loadFromSystem(Cursor::Type type)
             return loadFromSelector(@selector(_windowResizeNorthWestSouthEastCursor), m_cursor);
         case Cursor::SizeAll:
             return loadFromSelector(@selector(_moveCursor), m_cursor);
+        case Cursor::Wait:
+            return loadFromSelector(@selector(_waitCursor), m_cursor);
+        case Cursor::Help:
+            return loadFromSelector(@selector(_helpCursor), m_cursor);
     }
 
     // Since we didn't allocate the cursor ourself, we have to retain it
--- a/src/SFML/Window/OSX/CursorImpl.mm
+++ b/src/SFML/Window/OSX/CursorImpl.mm
@@ -65,6 +65,16 @@ bool CursorImpl::loadFromPixels(const Uint8* pixels, Vector2u size, Vector2u hot
     return m_cursor != nil;
 }
 
+////////////////////////////////////////////////////////////
+bool loadFromSelector(SEL selector, NSCursorRef& cursor)
+{
+    if ([NSCursor respondsToSelector:selector])
+    {
+        cursor = [NSCursor performSelector:selector];
+        return true;
+    }
+    return false;
+}
 
 ////////////////////////////////////////////////////////////
 bool CursorImpl::loadFromSystem(Cursor::Type type)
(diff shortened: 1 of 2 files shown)


mantognini:

This looks good but haven't actually tested it. I'll try to do it this week. Thanks for the PR!

mantognini:
I guess the documentation should be updated as well, specifically these lines:

https://github.com/JonnyPtn/SFML/blob/76f3adbd978e8955d24dcc787339563edc307b90/include/SFML/Window/Cursor.hpp#L65-L67 

What about replacing *no* with *maybe (1)* and adding a note along those lines: *(1) those cursors are usually available on macOS \<insert minimal version\> and newer but without any guarantee.* (Or phrased completely differently...)

JonnyPtn:
Ah yes, Sorry I missed the docs, will update now. 

Not really sure what to put for the supported version, can't find any documentation about it, and looking at other projects doesn't yield any useful results (glfw has an open (issue)[https://github.com/glfw/glfw/issues/427], and as far as I can tell SDL just uses the closed hand cursor for these diagonal resizes (?)

src/SFML/Window/OSX/CursorImpl.mm
```diff
@@ -80,6 +90,14 @@
...
         case Cursor::Cross:           m_cursor = [NSCursor crosshairCursor];           break;
         case Cursor::NotAllowed:      m_cursor = [NSCursor operationNotAllowedCursor]; break;
+            
+        // These cursor types are undocumented, may not be available on some platforms
+        case Cursor::SizeBottomLeftTopRight:
+            return loadFromSelector(@selector(_windowResizeNorthEastSouthWestCursor), m_cursor);
+        case Cursor::SizeTopLeftBottomRight:
+            return loadFromSelector(@selector(_windowResizeNorthWestSouthEastCursor), m_cursor);
+        case Cursor::SizeAll:
+            return loadFromSelector(@selector(_moveCursor), m_cursor);
```
JonnyPtn: Will returning here cause issues as it doesn't hit `[m_cursor retain]`?

JonnyPtn:
I've added the Help and Wait cursors, and updated documentation as best I can.

You can use this test code to cycle through the new cursors with Space:
```
#include <SFML/Graphics.hpp>

int main(int, char const**)
{
    // Load cursors
    std::vector<sf::Cursor> cursors(5);
    
    cursors[0].loadFromSystem(sf::Cursor::SizeBottomLeftTopRight);
    cursors[1].loadFromSystem(sf::Cursor::SizeTopLeftBottomRight);
    cursors[2].loadFromSystem(sf::Cursor::SizeAll);
    cursors[3].loadFromSystem(sf::Cursor::Wait);
    cursors[4].loadFromSystem(sf::Cursor::Help);
    
    int currentCursor = -1;
    
    // Create the main window
    sf::RenderWindow window(sf::VideoMode(800, 600), "SFML window");

    // Start the game loop
    while (window.isOpen())
    {
        // Process events
        sf::Event event;
        {
            // Close window: exit
            {
                window.close();
            }

            // Escape pressed: exit
            if (event.type == sf::Event::KeyPressed && event.key.code == sf::Keyboard::Escape)
            {
                window.close();
            }
            if (event.type == sf::Event::KeyPressed && event.key.code == sf::Keyboard::Space)
            {
                currentCursor++;
                if (currentCursor >= cursors.size())
                    currentCursor = 0;
                window.setMouseCursor(cursors[currentCursor]);
            }
        }

        // Clear screen 
        window.clear();
    }

    return EXIT_SUCCESS;
}

```

mantognini:
Yes, most likely. Good catch!

mantognini:
OK, that's good. But we should make it clear that Apple can remove those in
the future.

I'll see to test this tomorrow or the day after. :)

mantognini:
Superseded by #1425.

PR #1414, opened 2018-04-11
cmake/Macros.cmake: ex:->examples:

hanetzer:
vim interprets the # ex: comments as a modeline, which causes editing
this file with vim to throw an error.

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -1,7 +1,7 @@
 include(CMakeParseArguments)
 
 # set the appropriate standard library on each platform for the given target
-# ex: sfml_set_stdlib(sfml-system)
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
I'd only change it because it's more explicit and not because random text editor tries to execute something based on some characters in a text file..

MarioLiebisch:
Yeah, true, but then again "ex" isn't really an abbreviation I'd use anyway. I'd rather like to see "e.g.", if you want to use something short.

eXpl0it3r:
> I'd only change it because it's more explicit

eXpl0it3r:
Merged in c828314d12ddca2744bdac11f4018aeefcf39a90 and cherry-pick to branch 2.5.1

PR #1416, opened 2018-04-12
Fix cmake config for finding dependencies on iOS

JonnyPtn:
This fixes the cmake config so it can find iOS dependencies correctly:

- As iOS currently is only set up for static linking, openGLES framework is automatically linked, so no need to find openGL
- Vorbisfile and VorbisEnc are included in the precompiled iOS vorbis lib, so no need to search for them separately.


Ceylo:
It doesn't look to be usable yet, I get these errors:
```
CMake Error at /tmp/sfml-ios/lib/cmake/SFML/SFMLConfig.cmake:139 (message):
  SFML found but some of its dependencies are missing ( FreeType Ogg Vorbis
  FLAC)
  CMakeLists.txt:8 (find_package)
```

My test example is as follow:
```
find_package(SFML 2.4 COMPONENTS audio graphics network REQUIRED)

add_executable(Client main.cpp)
target_link_libraries(Client PRIVATE sfml-graphics sfml-audio sfml-network)
```

It seems like find_library() is only looking into the iOS SDK (OpenAL_LIB is properly set).

JonnyPtn:
Ah, that's the old "find_host_package" conundrum which we haven't really found a solution for yet... but if you use find_host_package(SFML...) that should solve the error

Ceylo:
When using find_host_package(SFML ...) there is no more CMake error but it finds wrong libraries:
capture d ecran 2018-04-13 a 00 14 35

Even if the dependencies are actually in the lib dir of the iOS install:
capture d ecran 2018-04-13 a 00 16 03

Ceylo:
This is not enough however. Even if I remove the frameworks shown in above screenshot, it finds libfreetype.dylib from /usr/local/lib. Which is not the one we want either.

From what I understand find_library() works with filenames and paths only. It doesn't check the library type (static/dynamic) or architecture. This last point is what we would actually need: only find libraries that match our build architecture.

As we can't do that I would suggest restricting dependency search on iOS to the SDK and the lib dir of the installed package. It seems quite reasonable and it's very unlikely that you'll find (or even need to find) SFML's dependencies in other locations.

JonnyPtn:
Yeah I think the only way to find static libraries first is to explicitly name them first using find_library. As I understand it this is a problem on all platforms, not just iOS? I'm reluctant to stop find_library searching the system paths.

Can we not modify the config to explicitly search for "libxxx.a" if SFML_STATIC_LIBRARIES is set? That would also solve the problem on other platforms?

Ceylo:
On other platforms there's also the issue between static/dynamic libs. However the libs usually have the same architecture as the host platform so it's ok.

I don't think you can or should search by exact file name, research is designed to be by library name, so that it is platform independent.

JonnyPtn:
I'd be inclined to leave it as it is in that case - It's the users responsibility to make sure they install the correct architecture of the libs, even on desktop (32/64 bit commonly).

Even if you force the user to specify the SFML directory to search in, that still won't ensure that the correct architecture is found, as the user could easily install the wrong architecture to that folder. 

The only other option that comes to mind would be to give the iOS libs a special name (the same way it differentiates between static/dynamic SFML libs), but I'm not overly fond of that idea either.

Any input from the SFML team on this? Ultimately I'm happy to go with the majority

Ceylo:
>  It's the users responsibility to make sure they install the correct architecture of the libs

But.. who's going to install libs for iOS in /Library/Frameworks or the other default search paths? Nobody. CMake is bad at handling dependencies when target architecture differ from host architecture. But you can't expect users to install libs for another host in the default host's paths. That would be quite ugly :D

> Even if you force the user to specify the SFML directory to search in, that still won't ensure that the correct architecture is found, as the user could easily install the wrong architecture to that folder.

Even if it doesn't guarantee anything, that's the best bet in my opinion. By default it would just work. I mean, if the user is using the SFMLConfig generated from an iOS install, then the libs found in that install are the good ones! Of course he can mess up the install dir. But that's still much better than asking the user to fix all library paths in CMake cache because they were all wrongly chosen I think.

> The only other option that comes to mind would be to give the iOS libs a special name (the same way it differentiates between static/dynamic SFML libs), but I'm not overly fond of that idea either.

I'm not very fond of it either. It looks ugly. But if it can make things work I won't fight against it. Seems to require much more changes than changing the search paths though. Also it won't work if the user has its own libraries built for iOS with standard names and wants to use them. Which somewhat breaks the purpose of the find_library() flexibility.

JonnyPtn:
It should be possible to check if a lib was built for iOS, using `otool libsfml-graphics-s-d.a -lv | grep -A LC_VERSION` and parsing the output for `IPHONEOS`, which we could wrap in `execute_process()`? The bit which is confusing me is how to implement this in the configs in a nice way, as I'm not too experienced with this config stuff yet. @Ceylo If you can point out a good place to put this command in the cmake config I'll have a go at implementing

JonnyPtn:
In the absence of an elegant solution right now, I'd say this should be merged, as it at least makes sure the correct dependencies are searched for making it usable, then we can try solutions for differentiating the libraries properly in future

eXpl0it3r:
So should we merge this for now?

eXpl0it3r:
Merged in ac38b17827b8e3324a7842590687bb7c7c23968c

PR #1417, opened 2018-04-12
(iOS) Use default supported rotations when none are specified

JonnyPtn:
discussion here: https://en.sfml-dev.org/forums/index.php?topic=23507.0

- If there are no supported orientations in the info.plist, then instead of assuming no rotations are supported, use the value of 'supportedInterfaceOrientations'
- Remove the override of 'supportedInterfaceOrientations', which as it stands returns that All orientations are available, whereas the default implementation returns the correct default orientations, [reference](  https://developer.apple.com/documentation/uikit/uiviewcontroller/1621435-supportedinterfaceorientations)


Ceylo:
I confirm this fixes weird view size issues after rotations:
Before & After
capture d ecran 2018-04-12 a 22 46 21capture d ecran 2018-04-12 a 22 45 42

I guess logo should be centered but that's a distinct issue.

JonnyPtn:
The logo should move to wherever you click/touch, and starts off at position 0,0 with a centred origin, so it looks as I'd expect. Doesn't matter a huge amount anyway as that example will be removed once the others are implemented

PR #1422, opened 2018-04-15
cmake/Macros.cmake: Do not alter default naming

hanetzer:
In just about any library which is compilable for both linux and
mingw-w64, import libraries are called *.dll.a and static libraries are
called *.a; there is no benefit from changing the naming, and could
cause confusion.

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -56,10 +56,8 @@ macro(sfml_add_library target)
             set_target_properties(${target} PROPERTIES DEBUG_POSTFIX -d)
         endif()
         if (SFML_OS_WINDOWS AND SFML_COMPILER_GCC)
-            # on Windows/gcc get rid of "lib" prefix for shared libraries,
-            # and transform the ".dll.a" suffix into ".a" for import libraries
+            # on Windows/gcc get rid of "lib" prefix for shared libraries
             set_target_properties(${target} PROPERTIES PREFIX "")
-            set_target_properties(${target} PROPERTIES IMPORT_SUFFIX ".a")
         endif()
     else()
         set_target_properties(${target} PROPERTIES DEBUG_POSTFIX -s-d)


eXpl0it3r:
@LaurentGomila do you remember what the reasoning was for this? Changing it now will cause confusion with existing users (and guides need to be adjusted), yet I do see that there's no real benefit from diverging what "everyone" else is doing, plus it will be easier to explain that import libraries are .dll.a and static libraries are .a.

MarioLiebisch:
In theory ".a" might be associated and as such be a hidden extension by default, so users would only see "sfml-system.dll" (the actual ".dll.a") and "sfml-system" (the actual ".dll"). This could be confusing, but then again you'd assume at least some basic understanding and sense from people wanting to get into programming, so probably a weak point. Besides that, the ".dll.a" file for MinGW is pretty much pointless anyway, since (in contrast to Visual Studio), GCC will just accept a DLL as a direct input for the linker, so there's no need to pick the ".dll.a" (whatever extension it has).

LaurentGomila:
That was probably done for consistency. But in the end I don't really care, except that if we go this way we may also not remove the "lib" prefix.

MarioLiebisch:
Not having "lib" in front is perfectly fine, it's very uncommon on Windows anyway IMO. But one could think about dropping "-s" for static versions, since file names would no longer overlap, although I'm not 100% sure how linking logic would work in case both are installed. Are the ".dll.a" files even used?

LaurentGomila:
> Not having "lib" in front is perfectly fine, it's very uncommon on Windows anyway IMO

This is the default, we have to explicitly remove it in our CMake config.

> But one could think about dropping "-s" for static versions, since file names would no longer overlap

Only in this specific case (Windows / gcc). Visual C++, for example, only uses ".a", for both import and static libraries.

> Are the ".dll.a" files even used?

If that's what the linker finds (and it usually is, since you give it the "lib" path, not the "bin" one), yes, they are used.

eXpl0it3r:
After thinking about this a bit more, I think it makes sense to keep the current behavior.
It's aligned with how it's usually expected when working with GCC and libraries from the Linux side.
Plus a lot of guides for SFML may already use this and would have to be updated, so we're only creating more confusion from my point of view.

PR #1425, opened 2018-04-20
Implement extra cursors on macOS

mantognini:
Superseds #1413 & #1401.

@JonnyPtn when carefully reviewing your PR I noticed a few issues in my original code so I've fixed it here. Feedback welcome.

I've removed Wait and SizeAll cursors as they don't look nice (Wait is not spinning and produces a broken rendering, SizeAll is a simple white cursor as you mentioned.) I think it would be more a disappointment/confusion or a source of bug reports to keep those.

It should also be more robust w.r.t. memory handling and reduce unwanted selector warnings.

### DIFF
--- a/include/SFML/Window/Cursor.hpp
+++ b/include/SFML/Window/Cursor.hpp
@@ -53,21 +53,21 @@ public:
     /// Refer to the following table to determine which cursor
     /// is available on which platform.
     ///
-    ///  Type                               | Linux | Mac OS X | Windows
-    /// ------------------------------------|:-----:|:--------:|:--------:
-    ///  sf::Cursor::Arrow                  |  yes  |    yes   |   yes
-    ///  sf::Cursor::ArrowWait              |  no   |    no    |   yes
-    ///  sf::Cursor::Wait                   |  yes  |    no    |   yes
-    ///  sf::Cursor::Text                   |  yes  |    yes   |   yes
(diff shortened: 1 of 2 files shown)


eliasdaler:
Look at how SDL does it. Maybe we can do the same too?

https://github.com/spurious/SDL-mirror/blob/a4e95f81625b16d7454a7edb4010e7d37f72cebf/src/video/cocoa/SDL_cocoamouse.m#L136

mantognini:
> but it's better to have closed hand as SizeAll then nothing.

I'd rather let the users decide on this themselves than impose an arbitrary and imperfect solution on them. ;-)

JonnyPtn:
Isn’t it an arbitrary and imperfect solution whether we used closed hand or leave it as normal arrow cursor? 

I’d also say closed hand is better than unchanged

mantognini:
The current choice is between the closed hand (or any non optimal cursor) and returning false to explicitly let the user choose another cursor.

eliasdaler:
Agreed.

eXpl0it3r:
Since this doesn't change any public-facing API or files, I've moved this from "feature" to "bugfix" and will include it in SFML 2.5.1.

eXpl0it3r:
Merged in aeca3dce4119b8998607c6d246c6212aff6f75f7

PR #1429, opened 2018-05-03
Added missing import for memcpy in MainAndroid

AlexAUT:
With the newest android ndk an include is missing for memcpy in MainAndroid.cpp


binary1248:
We are in C++ land. We `#include <cstring>` and make use of `std::memcpy` here. 😉

AlexAUT:
Changed :+1:

eXpl0it3r:

PR #1437, opened 2018-05-15
Fixed the Unix clipboard implementation causing an abort due to internal data races in Xlib.

binary1248:
Fixes issue reported [here](https://en.sfml-dev.org/forums/index.php?topic=24012.0).

Test code:
```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode(200, 200), "Test");

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
        }

        sf::err() << sf::Clipboard::getString().toAnsiString() << std::endl;

        window.clear();
        window.display();
    }
}
```
Causes an abort after a while without the fix.

### DIFF
--- a/include/SFML/Window/Clipboard.hpp
+++ b/include/SFML/Window/Clipboard.hpp
@@ -60,7 +60,12 @@ public:
     /// This function sets the content of the clipboard as a
     /// string.
     ///
-    /// \param text sf::String containing the data to be sent 
+    /// \warning Due to limitations on some operating systems,
+    ///          setting the clipboard contents is only
+    ///          guaranteed to work if there is currently an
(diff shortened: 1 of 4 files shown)


DanielJPetersen:
I'm on Linux Mint 18.3.

Tested the code without the patch and can reproduce the issue.  

Applied the patch and it fixes it for me.

Thanks.

eXpl0it3r:
@dascandy gave it a try and it seems to work fine. Thanks again! 🙂

dascandy:
Ran test on Ubuntu 18.04 with current Git, crashes within 10 seconds. Re-ran with patch applied and was stable for 5+ minutes.

PR #1439, opened 2018-05-16
Install CMake config files into lib${LIB_SIFFIX}

jcowgill:
The CMake config files should be installed in the same location as the libraries and pkg-config files.

Dispite what the documentation for `find_package` states, CMake will look into directories like `/usr/lib32` for CMake config files (and has done for ages). The documentation has been fixed in the latest git master:
https://cmake.org/cmake/help/git-master/command/find_package.html


eXpl0it3r:
Merged in 88bb3f89a5782cd95a781277c4f665239fc914cc and cherry-pick to branch 2.5.1

PR #1443, opened 2018-05-21
Allow re-creation of the shared context as a core context

binary1248:
Allow re-creation of the shared context as a core context if the user indicates they want a core profile context. Sharing of compatibility and core profile contexts is not possible on macOS which is why we need to have a way to re-create the shared context as a core context if required in this case.

The following code breaks on macOS without this change:
```cpp
#include <SFML/Window.hpp>

int main()
{
    sf::Window window(sf::VideoMode(100, 100), "Test", sf::Style::Default, sf::ContextSettings(0, 0, 0, 3, 2, sf::ContextSettings::Core));

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
        }

        window.display();
    }
}
```

### DIFF
--- a/src/SFML/Window/GlContext.cpp
+++ b/src/SFML/Window/GlContext.cpp
@@ -212,6 +212,55 @@ namespace
     // Supported OpenGL extensions
     std::vector<std::string> extensions;
 
+    // Load our extensions vector with the supported extensions
+    void loadExtensions()
+    {
+        extensions.clear();
+
+        // Check whether a >= 3.0 context is available
+        int majorVersion = 0;
+        glGetIntegerv(GL_MAJOR_VERSION, &majorVersion);
+
+        if (glGetError() == GL_INVALID_ENUM)
+        {
+            // Try to load the < 3.0 way
+            const char* extensionString = reinterpret_cast<const char*>(glGetString(GL_EXTENSIONS));
+
+            do
+            {
+                const char* extension = extensionString;
+
+                while (*extensionString && (*extensionString != ' '))
+                    extensionString++;
+
+                extensions.push_back(std::string(extension, extensionString));
+            }
+            while (*extensionString++);
+        }
+        else
+        {
+            // Try to load the >= 3.0 way
+            glGetStringiFuncType glGetStringiFunc = NULL;
+            glGetStringiFunc = reinterpret_cast<glGetStringiFuncType>(sf::priv::GlContext::getFunction("glGetStringi"));
+
+            if (glGetStringiFunc)
+            {
+                int numExtensions = 0;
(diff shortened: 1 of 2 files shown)


src/SFML/Window/GlContext.cpp
```diff
@@ -212,6 +212,55 @@ namespace
...
+        if (glGetError() == GL_INVALID_ENUM)
+        {
+            // Try to load the < 3.0 way
+            const char* extensionString = reinterpret_cast<const char*>(glGetString(GL_EXTENSIONS));
+
+            do
+            {
+                const char* extension = extensionString;
+
+                while(*extensionString && (*extensionString != ' '))
```
mantognini: missing space: `while_(`

src/SFML/Window/GlContext.cpp
```diff
@@ -446,6 +471,25 @@ GlContext* GlContext::create(const ContextSettings& settings, unsigned int width
 
     Lock lock(mutex);
 
+    // If resourceCount is 1 we know that we are inside sf::Context or sf::Window
```
mantognini: DRY code: re-factor this code and the identical one from above into a function. Maybe a template function to accommodate for the width/height vs owner/bpp difference? If the logic is the same (except how `context` is constructed), I think it makes sense.

src/SFML/Window/Win32/WglContext.cpp
```diff
@@ -137,17 +149,23 @@ m_ownsWindow   (false)
...
-    // Make sure that extensions are initialized if this is not the shared context
-    // The shared context is the context used to initialize the extensions
-    if (shared && shared->m_deviceContext)
-        ensureExtensionsInit(shared->m_deviceContext);
-
     // Create the rendering surface (window or pbuffer if supported)
     createSurface(shared, width, height, VideoMode::getDesktopMode().bitsPerPixel);
 
     // Create the context
     if (m_deviceContext)
```
mantognini: If possible, I would also DRY the ctors.

src/SFML/Window/GlContext.cpp
binary1248: I initially tried to do this, but `GlContext::initialize()` is private.

src/SFML/Window/Win32/WglContext.cpp
binary1248: Delegating constructors are coming in C++11. 😉

mantognini: Good! Maybe add a `\todo` so we actually clean up the code at some point?

binary1248:
There... should be a bit better now. The amount of code duplication can probably be significantly reduced in C++11. I wouldn't invest too much time in trying to minimize the lines of code right now.

PR #1449, opened 2018-06-07
Fixed KeyReleased event not firing when pressing both shift keys on W…

### LINKED ISSUE #1301: KeyReleased event not produced when pressing both Shift keys.
When I press both left and right shift keys, then release both keys, I do not get a KeyReleased event for the right shift, only the left shift key.  This is on WIndows 10.

kiwon0905:
…indows.

## Description
Fixes #1301 

## How to test this PR?
Observe that the key events are correctly reported.

```cpp
#include <SFML/Window.hpp>
#include <iostream>

static std::string getKeyboardKeyName(sf::Keyboard::Key key) 
{
    switch (key) {
    default:
    case sf::Keyboard::Unknown:
        return "Unknown";
    case sf::Keyboard::KeyCount:
        return "<...>";
    case sf::Keyboard::A:
        return "A";
    case sf::Keyboard::B:
        return "B";
    case sf::Keyboard::C:
        return "C";
    case sf::Keyboard::D:
        return "D";
    case sf::Keyboard::E:
        return "E";
    case sf::Keyboard::F:
        return "F";
    case sf::Keyboard::G:
        return "G";
    case sf::Keyboard::H:
        return "H";
    case sf::Keyboard::I:
        return "I";
    case sf::Keyboard::J:
        return "J";
    case sf::Keyboard::K:
        return "K";
    case sf::Keyboard::L:
        return "L";
    case sf::Keyboard::M:
        return "M";
    case sf::Keyboard::N:
        return "N";
    case sf::Keyboard::O:
        return "O";
    case sf::Keyboard::P:
        return "P";
    case sf::Keyboard::Q:
        return "Q";
    case sf::Keyboard::R:
        return "R";
    case sf::Keyboard::S:
        return "S";
    case sf::Keyboard::T:
        return "T";
    case sf::Keyboard::U:
        return "U";
    case sf::Keyboard::V:
        return "V";
    case sf::Keyboard::W:
        return "W";
    case sf::Keyboard::X:
        return "X";
    case sf::Keyboard::Y:
        return "Y";
    case sf::Keyboard::Z:
        return "Z";
    case sf::Keyboard::Num0:
        return "0";
    case sf::Keyboard::Num1:
        return "1";
    case sf::Keyboard::Num2:
        return "2";
    case sf::Keyboard::Num3:
        return "3";
    case sf::Keyboard::Num4:
        return "4";
    case sf::Keyboard::Num5:
        return "5";
    case sf::Keyboard::Num6:
        return "6";
    case sf::Keyboard::Num7:
        return "7";
    case sf::Keyboard::Num8:
        return "8";
    case sf::Keyboard::Num9:
        return "9";
    case sf::Keyboard::Escape:
        return "Escape";
    case sf::Keyboard::LControl:
        return "LControl";
    case sf::Keyboard::LShift:
        return "LShift";
    case sf::Keyboard::LAlt:
        return "LAlt";
    case sf::Keyboard::LSystem:
        return "LSystem";
    case sf::Keyboard::RControl:
        return "RControl";
    case sf::Keyboard::RShift:
        return "RShift";
    case sf::Keyboard::RAlt:
        return "RAlt";
    case sf::Keyboard::RSystem:
        return "RSystem";
    case sf::Keyboard::Menu:
        return "Menu";
    case sf::Keyboard::LBracket:
        return "LBracket";
    case sf::Keyboard::RBracket:
        return "RBracket";
    case sf::Keyboard::SemiColon:
        return "SemiColon";
    case sf::Keyboard::Comma:
        return "Comma";
    case sf::Keyboard::Period:
        return "Period";
    case sf::Keyboard::Quote:
        return "Quote";
    case sf::Keyboard::Slash:
        return "Slash";
    case sf::Keyboard::BackSlash:
        return "BackSlash";
    case sf::Keyboard::Tilde:
        return "Tilde";
    case sf::Keyboard::Equal:
        return "Equal";
    case sf::Keyboard::Dash:
        return "Dash";
    case sf::Keyboard::Space:
        return "Space";
    case sf::Keyboard::Return:
        return "Return";
    case sf::Keyboard::BackSpace:
        return "BackSpace";
    case sf::Keyboard::Tab:
        return "Tab";
    case sf::Keyboard::PageUp:
        return "PageUp";
    case sf::Keyboard::PageDown:
        return "PageDown";
    case sf::Keyboard::End:
        return "End";
    case sf::Keyboard::Home:
        return "Home";
    case sf::Keyboard::Insert:
        return "Insert";
    case sf::Keyboard::Delete:
        return "Delete";
    case sf::Keyboard::Add:
        return "Add";
    case sf::Keyboard::Subtract:
        return "Subtract";
    case sf::Keyboard::Multiply:
        return "Multiply";
    case sf::Keyboard::Divide:
        return "Divide";
    case sf::Keyboard::Left:
        return "Left";
    case sf::Keyboard::Right:
        return "Right";
    case sf::Keyboard::Up:
        return "Up";
    case sf::Keyboard::Down:
        return "Down";
    case sf::Keyboard::Numpad0:
        return "Num0";
    case sf::Keyboard::Numpad1:
        return "Num1";
    case sf::Keyboard::Numpad2:
        return "Num2";
    case sf::Keyboard::Numpad3:
        return "Num3";
    case sf::Keyboard::Numpad4:
        return "Num4";
    case sf::Keyboard::Numpad5:
        return "Num5";
    case sf::Keyboard::Numpad6:
        return "Num6";
    case sf::Keyboard::Numpad7:
        return "Num7";
    case sf::Keyboard::Numpad8:
        return "Num8";
    case sf::Keyboard::Numpad9:
        return "Num9";
    case sf::Keyboard::F1:
        return "F1";
    case sf::Keyboard::F2:
        return "F2";
    case sf::Keyboard::F3:
        return "F3";
    case sf::Keyboard::F4:
        return "F4";
    case sf::Keyboard::F5:
        return "F5";
    case sf::Keyboard::F6:
        return "F6";
    case sf::Keyboard::F7:
        return "F7";
    case sf::Keyboard::F8:
        return "F8";
    case sf::Keyboard::F9:
        return "F9";
    case sf::Keyboard::F10:
        return "F10";
    case sf::Keyboard::F11:
        return "F11";
    case sf::Keyboard::F12:
        return "F12";
    case sf::Keyboard::F13:
        return "F13";
    case sf::Keyboard::F14:
        return "F14";
    case sf::Keyboard::F15:
        return "F15";
    case sf::Keyboard::Pause:
        return "Pause";
    }
}

int main()
{
    sf::Window window(sf::VideoMode(200, 200), "1");
   
    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();

            {
                std::cout << getKeyboardKeyName(event.key.code) << " pressed\n";
            }
            {
                std::cout << getKeyboardKeyName(event.key.code) << " released\n";

            }
        }
        window.display();
    }

    return 0;
}

```


kiwon0905:
Hello, I was wondering what it would take for this to be reviewed? Also I think raw using raw inputs might be beneficial to implementing windows scancode.

eXpl0it3r:
Sorry for getting back to only so late. I've tested the code and it seems to work, however, I have my concerns about dropping `WM_KEYUP` and just using `WM_SYSKEYUP`. It's not really clear to me, whether every keyboard device has to provide raw input and how other devices can potentially mess up the raw input.

While looking further into this I came across [GetKeyboardState](https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-getkeyboardstate), which may be useful here.

- [GLFW simply triggers two KeyReleased events](https://github.com/glfw/glfw/blob/master/src/win32_window.c#L747)
- [SDL uses GetKeyboardState](https://github.com/spurious/SDL-mirror/blob/17af4584cb28cdb3c2feba17e7d989a806007d9f/src/video/windows/SDL_windowsevents.c#L652)
- [LAF seems to only report one key-down event for modifiers](https://github.com/aseprite/laf/blob/master/os/win/window.cpp#L1109)

kiwon0905:
[MSDN doc](https://docs.microsoft.com/en-us/windows/desktop/inputdev/about-raw-input) says "The raw input API provides a stable and robust way for applications to accept raw input from any HID, including the keyboard and mouse." From reading this sentence, I think we can assume using raw input is safe with every keyboard.

JonnyPtn:
I agree that rawinput is the way to solve this, and I don't think there are many concerns - my understanding is that the previous WM_KEYUP/DOWN events are basically what windows produces after processing the raw input anyway, and that processing is also where it skips the shift key up event if another shift key is pressed.

As we now have handling for scancodes anyway I think it also makes sense for us to just do the rawinput processing ourselves and bypass any quirks from window's behaviour

~~My only concern would be that as you need to register for raw input from devices, I think we'd need to handle device dis/connection too (might also open the door for handling multiple keyboards/mice in future)~~

^ Not a concern on second reading, from docs:

> Note that an application can register a device that is not currently attached to the system. When this device is attached, the Windows Manager will automatically send the raw input to the application

This PR is also quite old now and will need fairly significant updates for SFML 3 - Would you be able/willing to do that @kiwon0905?

PR #1450, opened 2018-06-09
Fixed performance issue with reading WAV files

eXpl0it3r:
## Description

Calling tell() and thus std::ftell() for every reading iteration ate up 80-90% of the whole read call. As the check is redundant anyways (see the analogous OGG implementation), it should be safely removed. Please comment if it's not safe!

As brought up on [the forum](https://en.sfml-dev.org/forums/index.php?topic=24127.0).

I only tested this with Visual Studio 2017 (Debug & Release), it might well be that other compilers have a more performant ´std::ftell()´.

## How to test this PR?

Run the following code and time the load-time. With master it will take up to 30s to load the 10 WAV files.

```cpp
#include <SFML/Audio.hpp>
#include <vector>

int main()
{
	std::vector<sf::SoundBuffer> buffers;
	buffers.reserve(10);
	for(std::size_t i = 1; i < 11; ++i)
	{
		buffers.emplace_back();
		buffers.back().loadFromFile("assets/" + std::to_string(i) + ".wav");
	}

	sf::Sound sound;
	sound.setBuffer(buffers[0]);
	sound.play();
}
```

### DIFF
--- a/src/SFML/Audio/SoundFileReaderWav.cpp
+++ b/src/SFML/Audio/SoundFileReaderWav.cpp
@@ -155,7 +155,11 @@ Uint64 SoundFileReaderWav::read(Int16* samples, Uint64 maxCount)
     assert(m_stream);
 
     Uint64 count = 0;
-    while ((count < maxCount) && (static_cast<Uint64>(m_stream->tell()) < m_dataEnd))
+    Uint64 startPos = m_stream->tell();
+
+    // Tracking of m_dataEnd is important to prevent sf::Music from reading
+    // data until EOF, as WAV files may have metadata at the end.
+    while ((count < maxCount) && (startPos + count * m_bytesPerSample < m_dataEnd))
     {
         switch (m_bytesPerSample)
         {


eXpl0it3r:
As pointed out on the forum, it seems we can't directly remove it, because WAV files can have meta data at the end of the file, so simply reading till the end doesn't work.

This check was first introduced in commit 698bbccd6a0e2bb3971b052996a78b39272c8b3c by @FRex 

I wonder whether it's possible to track the position manually, instead of having to call tell() all the time...

FRex:
Thanks for doing at at me, I'd not have noticed this otherwise.

When it was reported on https://en.sfml-dev.org/forums/index.php?topic=19381.msg139774#msg139774 the wave bug affected only wav in sf::Music. SoundBuffer only read as many samples as reported by the parsed header so even though SoundFileReaderWav had the potential to return more data than there was it never happened with it.

I grabbed some random wav file (from Pidgin) and it has a few bytes of info at the end of file about what program exported it. That causes a small but totally audible crack with this commit applied and this code. If I append lots of random data to end of file it becomes gibberish static (but for a second only, I think this has something to do with amount of data read at once but I didn't dig into code to check). Adding looping points to Music also affected this bug slightly, with this commit and setLoop(true) there is no crack.

This check was basically introduced to never return bad data from wav reader, even if more is requested. Due to the fact that most SFML code behaves well and checks how many samples there are and reads only that amount and it only seems to manifests with wav used in Music (rare use case) it's really exotic to run into this bug since most people do ogg or something for music and wav for sound effects (but then again someone did run into it on the forum, and maybe some people don't care since they have so little music the size doesn't matter or have crisp retro audio music and could run into this, on forum Ironbell ran into it with wav music exported by FL Studio, which is kinda popular commercial program I guess..?).

Even though reader and writer classes are private they can be created via the public factory so in the unlikely event someone wants to use SoundFileReaderWav in their own code they could come across that too. I'm not too into the idea of "if you read more you can get some bad data" but then again it's so rare and minor anyway.

Then again the issue that ftell is so heavy is really silly, but yes, out input stream for FILE can be easily changed to track the current position in its own code on each read (move by amount read) and seek (with ftell) so that the virtual tell in input stream is 100% our code and thus avoids ftell.

Code to get a small pop if there is metadata at the end with this commit:
```cpp
#include <SFML/Audio.hpp>
#include <iostream>
#include <string>

int main()
{
    sf::Music mus;
    mus.openFromFile("receive.wav");
    mus.play();

    std::string str;
    std::getline(std::cin, str);
}
```

FRex:
Yes, doing own tracking in just that function to not call ftell in the while should work too. I thought you meant own tracking of current position in input stream when you said to "track it manually".

eXpl0it3r:
I've implemented @LaurentGomila's suggestion and tried the receive.wav from Pidgin and it doesn't seem to click and the load times are still in the ms range.

So others don't have to download Pidgin, I've attached the receive.wav here: [receive.zip](https://github.com/SFML/SFML/files/2087309/receive.zip) 😄

FRex:
Yes, looks good/equivalent to old code but counted in the function now.

There isn't any significance to that specific sound effect, it's just a wav I grabbed to test that also had some info at the end. :smile:

MJBrune:
Works great for me as well 👍

LaurentGomila:
We could also add a comment that explains why we need to check m_dataEnd as well, so that next time we're not tempted to just remove it 😁

eXpl0it3r:
Added a comment 🙂

src/SFML/Audio/SoundFileReaderWav.cpp
```diff
@@ -155,7 +155,11 @@ Uint64 SoundFileReaderWav::read(Int16* samples, Uint64 maxCount)
     assert(m_stream);
 
     Uint64 count = 0;
-    while ((count < maxCount) && (static_cast<Uint64>(m_stream->tell()) < m_dataEnd))
+    Uint64 startPos = m_stream->tell();
+
+    // Tracking of m_dataEnd is important to prevent sf::Music from reading
```
Debtriches, line 160: - 
```suggestion
    Tracking of m_dataEnd is important to prevent sf::Music from reading
```

eXpl0it3r: ?

PR #846, opened 2015-03-26
Added support for clipping masks (Implements #1)

zsbzsb:
_Recreated from #842_

This is ready for a code review, hoping for a SFML 2.4 inclusion. This implements #1, yes the very first issue. :smile:

I based my branch on /feature/gl_dev_new due to the OpenGL changes. Clipping is done with the stencil buffer which gets enabled when a mask is applied to the render states.

Some points from #1 are answered below about this implementation.

> And do we clip in scene coordinates or in window coordinates?

Clipping is done in scene coordinates.

> However since it can be disabled (and is currently not enabled by default) in the context settings.

It is now enabled by default in the context and render texture constructors.

> it implies that the user can disable clipping masks, without even knowing it. So this must be at least well documented, and even enforced if possible.

It is now noted in the documentation that disabling stencil buffer will disable clipping masks. As for enforcing it? Its not really needed, if the user disables it - it is their fault because it is clearly stated.

> So the main problem is to decide what we use to define the clipping area. We could start with a `const sf::Drawable*` to have maximum flexibility.

Clipping area is defined by the `sf::ClippingMask` class which is a container for multiple `const sf::Drawable*` pointers.

> Another question is: do we allow only one active mask, or do we provide the ability to add multiple masks?

Only 1 mask is active at one time, but it can be defined by multiple drawables.

As with my other PR (#840), please review and test and then give feedback. **Even textures with alpha channels are supported.**

``` cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window({ 800, 600 }, "Clipping Mask Test");
    window.setVerticalSyncEnabled(true);

    // create a circle for part of the clipping mask
    sf::CircleShape keyholetop(50);
    keyholetop.setOrigin({ 50, 50 });

    // create a square for part of the clipping mask
    sf::RectangleShape keyholebody({ 36, 150 });
    keyholebody.setOrigin({ 18, 0 });

    // create the clipping mask
    sf::ClippingMask clipmask;

    // add the shapes to the clipping mask
    // sf::ClippingMask behaves like sf::VertexArray
    // so you can append any sf::Drawable to the mask
    // the mask will not copy the sf::Drawable, it only stores a pointer
    clipmask.append(keyholetop);
    clipmask.append(keyholebody);

    sf::Texture background;
    // use any background image
    // I used the image from opengl example
    background.loadFromFile("sfml/examples/opengl/resources/background.jpg");

    sf::Sprite backgroundsprite(background);

    // set a clipping area
[...]

PR #1451, opened 2018-06-10
Added support for scissor testing.

binary1248:
This is a subset of #846 and implements a part of #1.

Test with:
```cpp
#include <SFML/Graphics.hpp>

int main()
{
    auto window = sf::RenderWindow{sf::VideoMode{{200u, 200u}}, "Test"};

    auto shape = sf::CircleShape{100.f};

    auto left = window.getView();
    left.setScissor(sf::FloatRect{{0.f, 0.f}, {0.5f, 1.f}});

    auto right = window.getView();
    right.setScissor(sf::FloatRect{{0.5f, 0.f}, {0.5f, 1.f}});

    while (window.isOpen())
    {
        {
                window.close();
        }

        window.setView(window.getDefaultView());
        window.clear();

        window.setView(left);
        window.clear(sf::Color::Red);

        window.setView(right);
        window.draw(shape);

        window.display();
    }
}
```

### DIFF
--- a/include/SFML/Graphics/View.hpp
+++ b/include/SFML/Graphics/View.hpp
@@ -120,6 +120,30 @@ public:
     void setViewport(const FloatRect& viewport);
 
     ////////////////////////////////////////////////////////////
+    /// \brief Set the target scissor rectangle
+    ///
+    /// The scissor rectangle, expressed as a factor (between 0 and 1) of
+    /// the RenderTarget, specifies the region of the RenderTarget whose
+    /// pixels are able to be modified by draw or clear operations.
+    /// Any pixels which lie outside of the scissor rectangle will
+    /// not be modified by draw or clear operations.
+    /// For example, a scissor rectangle which only allows modifications
+    /// to the right side of the target would be defined
+    /// with View.setScissor(sf::FloatRect({0.5f, 0.f}, {0.5f, 1.f})).
+    /// By default, a view has a scissor rectangle which allows
+    /// modifications to the entire target. This is equivalent to
+    /// disabling the scissor test entirely. Passing the default
+    /// scissor rectangle to this function will also disable
+    /// scissor testing.
+    ///
+    /// \param scissor New scissor rectangle
+    ///
+    /// \see getScissor
+    ///
+    ////////////////////////////////////////////////////////////
+    void setScissor(const FloatRect& scissor);
+
+    ////////////////////////////////////////////////////////////
     /// \brief Reset the view to the given rectangle
     ///
     /// Note that this function resets the rotation angle to 0.
@@ -170,6 +194,16 @@ public:
     ///
     ////////////////////////////////////////////////////////////
     const FloatRect& getViewport() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the scissor rectangle of the view
+    ///
+    /// \return Scissor rectangle, expressed as a factor of the target size
+    ///
+    /// \see setScissor
+    ///
+    ////////////////////////////////////////////////////////////
+    const FloatRect& getScissor() const;
 
     ////////////////////////////////////////////////////////////
     /// \brief Move the view relatively to its current position
@@ -240,6 +274,7 @@ private:
     Vector2f  m_size;                       //!< Size of the view, in scene coordinates
     Angle     m_rotation;                   //!< Angle of rotation of the view rectangle
     FloatRect m_viewport{{0, 0}, {1, 1}};   //!< Viewport rectangle, expressed as a factor of the render-target's size
+    FloatRect m_scissor{{0, 0}, {1, 1}};    //!< Scissor rectangle, expressed as a factor of the render-target's size
     mutable Transform m_transform;          //!< Precomputed projection transform corresponding to the view
     mutable Transform m_inverseTransform;   //!< Precomputed inverse projection transform corresponding to the view
     mutable bool      m_transformUpdated{}; //!< Internal state telling if the transform needs to be updated
@@ -269,6 +304,34 @@ private:
 /// rectangle doesn't have the same size as the viewport, its
 /// contents will be stretched to fit in.
 ///
+/// The scissor rectangle allows for specifying regions of the
+/// render target to which modifications can be made by draw
+/// and clear operations. Only pixels that are within the region
+/// will be able to be modified. Pixels outside of the region will
+/// not be modified by draw or clear operations.
+///
+/// Certain effects can be created by either using the viewport or
+/// scissor rectangle. While the results appear identical, there
+/// can be times where one method should be preferred over the other.
+/// Viewport transformations are applied during the vertex processing
+/// stage of the graphics pipeline, before the primitives are
+/// rasterized into fragments for fragment processing. Since
+/// viewport processing has to be performed and cannot be disabled,
+/// effects that are performed using the viewport transform are
+/// basically free performance-wise. Scissor testing is performed in
+/// the per-sample processing stage of the graphics pipeline, after
+/// fragment processing has been performed. Because per-sample
+/// processing is performed at the last stage of the pipeline,
+/// fragments that are discarded at this stage will cause the
+/// highest waste of GPU resources compared to any method that
+/// would have discarded vertices or fragments earlier in the
+/// pipeline. There are situations in which scissor testing has
+/// to be used to control whether fragments are discarded or not.
+/// An example of such a situation is when performing the viewport
+/// transform on vertices is necessary but a subset of the generated
+/// fragments should not have an effect on the stencil buffer or
+/// blend with the color buffer.
+//
 /// To apply a view, you have to assign it to the render target.
 /// Then, objects drawn in this render target will be
 /// affected by the view until you use another view.
--- a/src/SFML/Graphics/RenderTarget.cpp
+++ b/src/SFML/Graphics/RenderTarget.cpp
@@ -161,6 +161,10 @@ void RenderTarget::clear(const Color& color)
         // Unbind texture to fix RenderTexture preventing clear
         applyTexture(nullptr);
 
+        // Apply the view (scissor testing can affect clearing)
+        if (!m_cache.enable || m_cache.viewChanged)
+            applyCurrentView();
+
         glCheck(glClearColor(color.r / 255.f, color.g / 255.f, color.b / 255.f, color.a / 255.f));
         glCheck(glClear(GL_COLOR_BUFFER_BIT));
     }
@@ -197,6 +201,17 @@ IntRect RenderTarget::getViewport(const View& view) const
 
     return IntRect(Rect<long>({std::lround(width * viewport.left), std::lround(height * viewport.top)},
                               {std::lround(width * viewport.width), std::lround(height * viewport.height)}));
+}
+
+
+////////////////////////////////////////////////////////////
+IntRect RenderTarget::getScissor(const View& view) const
+{
+    const auto [width, height] = Vector2f(getSize());
+    const FloatRect& scissor   = view.getScissor();
+
+    return IntRect(Rect<long>({std::lround(width * scissor.left), std::lround(height * scissor.top)},
+                              {std::lround(width * scissor.width), std::lround(height * scissor.height)}));
 }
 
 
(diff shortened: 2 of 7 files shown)


binary1248:
Updated to current master.

JonnyPtn (approved):

include/SFML/Graphics/View.hpp
```diff
@@ -119,6 +119,27 @@ class SFML_GRAPHICS_API View
...
+    /// \brief Set the target scissor rectangle
+    ///
+    /// The scissor rectangle, expressed as a factor (between 0 and 1) of
+    /// the RenderTarget, specifies the region of the RenderTarget whose
+    /// pixels are able to be modified by draw or clear operations.
+    /// Any pixels which lie outside of the scissor rectangle will
+    /// not be modified by draw or clear operations.
+    /// For example, a scissor rectangle which only allows modifications
+    /// to the right side of the target would be defined
+    /// with View.setScissor(sf::FloatRect(0.5, 0, 0.5, 1)).
```
kimci86, line 132:
```suggestion
    /// with View.setScissor(sf::FloatRect({0.5, 0}, {0.5, 1})).
```

binary1248:
Updated to current master and added tests.

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -199,6 +203,18 @@ IntRect RenderTarget::getViewport(const View& view) const
 }
 
 
+////////////////////////////////////////////////////////////
+IntRect RenderTarget::getScissor(const View& view) const
+{
+    float            width   = static_cast<float>(getSize().x);
+    float            height  = static_cast<float>(getSize().y);
```
ChrisThrasher, line 210:
```suggestion
    const auto [width, height] = Vector2f(getSize());
```

test/Graphics/View.test.cpp
```diff
@@ -92,6 +92,13 @@ TEST_CASE("[Graphics] sf::View")
         CHECK(view.getInverseTransform() == Approx(sf::Transform(500, 0, 500, 0, -500, 500, 0, 0, 1)));
     }
 
+    SUBCASE("Set/get scissor")
+    {
+        sf::View view;
+        view.setScissor({{0, 0}, {0.5, 1}});
+        CHECK(view.getScissor() == sf::FloatRect({0, 0}, {0.5, 1}));
+    }
+
```
ChrisThrasher, line 101: Can you add `getScissor()` to the constructor tests to test its initial value?

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -199,6 +203,18 @@ IntRect RenderTarget::getViewport(const View& view) const
...
+    const FloatRect& scissor = view.getScissor();
+
+    return IntRect({static_cast<int>(0.5f + width * scissor.left), static_cast<int>(0.5f + height * scissor.top)},
+                   {static_cast<int>(0.5f + width * scissor.width), static_cast<int>(0.5f + height * scissor.height)});
```
ChrisThrasher, line 214: Is this "add half then cast to int" pattern a way of rounding to the nearest integer? Might I suggest `std::round` if that's our intention?

ChrisThrasher (approved):
I found a few items you should consider but besides that I'm cool with this PR 👍

binary1248:
Fixed.

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -563,9 +580,42 @@ void RenderTarget::applyCurrentView()
 {
     // Set the viewport
     const IntRect viewport = getViewport(m_view);
-    const int     top      = static_cast<int>(getSize().y) - (viewport.top + viewport.height);
+    int           top      = static_cast<int>(getSize().y) - (viewport.top + viewport.height);
     glCheck(glViewport(viewport.left, top, viewport.width, viewport.height));
 
+    const FloatRect scissor = m_view.getScissor();
+
+    if (m_cache.scissorEnabled)
```
MarioLiebisch, line 588: Is this differentiation even necessary? Feels like unnecessary code duplication just to avoid one potentially unnecessary call to `glDisable()` by doing one extra branching. And if we go that far, shouldn't it verify the rectangle actually changed since it's been enabled?

binary1248: Fixed.

binary1248:
Updated.

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -203,6 +207,17 @@ IntRect RenderTarget::getViewport(const View& view) const
...
 
 
+////////////////////////////////////////////////////////////
+IntRect RenderTarget::getScissor(const View& view) const
+{
+    const auto [width, height] = Vector2f(getSize());
+    const FloatRect& scissor   = view.getScissor();
+
+    return IntRect(Rect<long>({std::lround(width * scissor.left), std::lround(height * scissor.top)},
+                              {std::lround(width * scissor.width), std::lround(height * scissor.height)}));
```
eXpl0it3r, line 217: Any reason to prefer `std::lround` over `std::round`?

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -563,9 +580,33 @@ void RenderTarget::applyCurrentView()
...
     // Set the viewport
     const IntRect viewport = getViewport(m_view);
-    const int     top      = static_cast<int>(getSize().y) - (viewport.top + viewport.height);
+    int           top      = static_cast<int>(getSize().y) - (viewport.top + viewport.height);
     glCheck(glViewport(viewport.left, top, viewport.width, viewport.height));
 
+    // Set the scissor rectangle and enable/disable scissor testing
+    const FloatRect scissor = m_view.getScissor();
+
+    if ((scissor.left == 0) && (scissor.top == 0) && (scissor.width == 1) && (scissor.height == 1))
```
eXpl0it3r, line 589:
```suggestion
    if ((scissor.left == 0.f) && (scissor.top == 0.f) && (scissor.width == 1.f) && (scissor.height == 1.f))
```

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -563,9 +580,33 @@ void RenderTarget::applyCurrentView()
...
+    {
+        if (m_cache.scissorEnabled)
+        {
+            glCheck(glDisable(GL_SCISSOR_TEST));
+            m_cache.scissorEnabled = false;
+        }
```
eXpl0it3r, line 595: I feel like this "resetting"/disabling should be documented somehow on the API

eXpl0it3r (changes requested):
Looks good! 👍 

Some small, mostly documentation comments.  
Plus, we may want to document somewhere, what the difference between viewport and scissor testing is.

src/SFML/Graphics/RenderTarget.cpp
ChrisThrasher: `std::round` returns a floating point type and since we need need an integral type, it's easier to use `std::lround` which returns a `long`.

src/SFML/Graphics/RenderTarget.cpp
binary1248: Documented in `setScissor` documentation of latest commit.

kimci86:
It doesn't work well in interaction with `sf::Font`.

See the result of the following code:

```c++
#include <SFML/Graphics.hpp>

int main()
{
    auto target = sf::RenderTexture{};
    if (!target.create({400, 300}))
        return 1;

    auto font = sf::Font{};
    // https://github.com/SFML/SFML/raw/master/test/Graphics/tuffy.ttf
    if (!font.loadFromFile("tuffy.ttf"))
        return 1;

    auto text = sf::Text{font, "0123456789ABCDEFGHIJKLMNOPQ", 30};
    text.setPosition({0.f, 100.f});

    target.clear(sf::Color{64, 64, 64});

    auto scissorView = target.getDefaultView();
    scissorView.setScissor({{0.1f, 0.1f}, {0.8f, 0.8f}});
    target.setView(scissorView);
    target.clear(sf::Color{128, 192, 128});
    target.draw(text);

    target.display();

    if (!target.getTexture().copyToImage().saveToFile("result.png"))
        return 1;

    return 0;
}
```

result

Below is the expected result, which I get after applying this patch (quick hack, probably not a proper solution).

```diff
diff --git a/src/SFML/Graphics/Texture.cpp b/src/SFML/Graphics/Texture.cpp
index 7c6df2e5..70b5286d 100644
--- a/src/SFML/Graphics/Texture.cpp
+++ b/src/SFML/Graphics/Texture.cpp
@@ -463,6 +463,8 @@ void Texture::update(const std::uint8_t* pixels, const Vector2u& size, const Vec
         // Make sure that the current texture binding will be preserved
         const priv::TextureSaver save;
 
+        glCheck(glDisable(GL_SCISSOR_TEST));
+
         // Copy pixels from the given array to the texture
         glCheck(glBindTexture(GL_TEXTURE_2D, m_texture));
         glCheck(glTexSubImage2D(GL_TEXTURE_2D,
```

result

binary1248:
Updated. Turns out scissor testing was having an affect on our texture copying when the FBO blit implementation is used.

src/SFML/Graphics/View.cpp
```diff
@@ -87,6 +87,13 @@ void View::setViewport(const FloatRect& viewport)
 }
 
 
+////////////////////////////////////////////////////////////
+void View::setScissor(const FloatRect& scissor)
+{
+    m_scissor = scissor;
```
ChrisThrasher, line 93: Should we assert that this float rect position and size are within [0, 1]? It would be friendly to users to check this for them.

ChrisThrasher (approved):

binary1248:
Updated.

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -557,9 +574,31 @@ void RenderTarget::applyCurrentView()
 {
     // Set the viewport
     const IntRect viewport = getViewport(m_view);
-    const int     top      = static_cast<int>(getSize().y) - (viewport.top + viewport.height);
+    int           top      = static_cast<int>(getSize().y) - (viewport.top + viewport.height);
```
kimci86, line 577: I think it would be more readable to keep `top` const here and make another const int for scissor top instead of reusing this variable. Also naming them `viewportTop` and `scissorTop` respectively.

src/SFML/Graphics/View.cpp
```diff
@@ -87,6 +87,18 @@ void View::setViewport(const FloatRect& viewport)
 }
 
 
+////////////////////////////////////////////////////////////
+void View::setScissor(const FloatRect& scissor)
+{
+    assert(scissor.left >= 0.0f && scissor.left <= 1.0f && "scissor.left must lie within [0, 1]");
+    assert(scissor.top >= 0.0f && scissor.top <= 1.0f && "scissor.top must lie within [0, 1]");
+    assert(scissor.width >= 0.0f && scissor.width <= 1.0f && "scissor.width must lie within [0, 1]");
+    assert(scissor.height >= 0.0f && scissor.height <= 1.0f && "scissor.height must lie within [0, 1]");
```
kimci86, line 96: Maybe this should be checking position+size coordinates instead of size.

kimci86 (approved):
I confirm the problem with font rendering is not there anymore. Looks good to me besides minor comments below.

src/SFML/Graphics/View.cpp
ChrisThrasher:
```cpp
    assert(sf::FloatRect({0, 0}, {1, 1}).contains(scissor.getPosition()) &&
           "View::setScissor() Scissor start position must lie within the screen");
    assert(sf::FloatRect({0, 0}, {1, 1}).contains(scissor.getPosition() + scissor.getSize()) &&
           "View::setScissor() Scissor end position must lie within the screen");
```
This is what I came up with for ensuring both the start and end of this `FloatRect` are within the screen.

kimci86: That would not work because `Rect::contains` checks if point is in the product of half-open intervals [left, left+width)*[top, top+height). In particular `sf::FloatRect({0, 0}, {1, 1}).contains({1, 1})` is false.
However it could be nicely expressed if we had a contains method between two Rects.

kimci86: This might be a place where having some Rect::getAbsolute method in the future could be useful to support negative size.

binary1248:
Updated.

ChrisThrasher (approved):

PR #1452, opened 2018-06-11
Fixed two issues with glyph sub-pixel positioning

oomek:

----

## Description

A new method of swapping glyphs on the GPU, introduced a while ago, disabled smoothing of the glyph textures. The result of that is the sub-pixel positioning of bigger fonts stopped working, so we have to re-enable it.
Added 1 pixel padding for glyph uv's  and increased glyph quads boundaries by 1 pixel so the glyphs aren't cropped when text is being scrolled with sub-pixel increments

Here's a little video showing what I fixed:
https://youtu.be/4r7MQ_O7Fvg

This is the commit that have identified that caused the glyph textures to loose smoothing:
https://github.com/SFML/SFML/commit/6b71456a55c9cce3eae12ee9cdacbf96fe1c1a71

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -727,6 +727,7 @@ IntRect Font::findGlyphRect(Page& page, unsigned int width, unsigned int height)
                 // Make the texture 2 times bigger
                 Texture newTexture;
                 newTexture.create(textureWidth * 2, textureHeight * 2);
+                newTexture.setSmooth(true);
                 newTexture.update(page.texture);
                 page.texture.swap(newTexture);
             }
--- a/src/SFML/Graphics/Text.cpp
+++ b/src/SFML/Graphics/Text.cpp
@@ -50,15 +50,17 @@ namespace
     // Add a glyph quad to the vertex array
     void addGlyphQuad(sf::VertexArray& vertices, sf::Vector2f position, const sf::Color& color, const sf::Glyph& glyph, float italicShear, float outlineThickness = 0)
     {
-        float left   = glyph.bounds.left;
-        float top    = glyph.bounds.top;
-        float right  = glyph.bounds.left + glyph.bounds.width;
-        float bottom = glyph.bounds.top  + glyph.bounds.height;
+        float padding = 1.0;
 
-        float u1 = static_cast<float>(glyph.textureRect.left);
-        float v1 = static_cast<float>(glyph.textureRect.top);
-        float u2 = static_cast<float>(glyph.textureRect.left + glyph.textureRect.width);
-        float v2 = static_cast<float>(glyph.textureRect.top  + glyph.textureRect.height);
+        float left   = glyph.bounds.left - padding;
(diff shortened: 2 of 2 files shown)


oomek:
This is the commit that have identified that caused the glyphs to loose smoothing:
https://github.com/SFML/SFML/commit/6b71456a55c9cce3eae12ee9cdacbf96fe1c1a71

LaurentGomila:
Thanks a lot for the work, and the video showing the improvement :+1:

The texture.setSmooth fix is definitely right, and the padding thing, I'm not sure, but it looks ok.

oomek:
What are you not sure about Laurent?

LaurentGomila:
I mean that I didn't follow the latest changes in sf::Font and sf::Text, and that specific fix should be reviewed by someone else before validation.

My main concern is, why is it correct to put additional padding to the glyphs geometry and UVs, with no matching modification in the glyphs texture in sf::Font?

oomek:
I've added some alpha in the shader so you can see the glyph boundaries. As you can see on the animated gifs below the text stays in place also with scaled viewport.
https://image.ibb.co/izMsL8/glyph_border_zoom1_0.gif
https://image.ibb.co/n72z08/glyph_border_zoom0_5.gif

oomek:
https://github.com/SFML/SFML/blob/6b71456a55c9cce3eae12ee9cdacbf96fe1c1a71/src/SFML/Graphics/Font.cpp#L584
This was added to not to leak the neighbouring glyphs with fractional positioning. But that's only half of the solution. This border needs to be displayed, because the glyph boundaries snap to pixels, so you have cropped edges starting half way to the next integer position.

eXpl0it3r:
Please use the edit function instead of adding x comments.

oomek:
I have a question if I may. What is the proper way of rendering text on adjusted views? When I set the zoom level for example the text is rendered with interpolation so it’s blurry or jaggy. It has nothing to do with my commit, but I need to make sure I’m testing it the right way.

LaurentGomila:
Yes, scaled text (either by its own or in a zoomed view) is blurry and is therefore not recommended; the solution to get the best quality is usually to use no scale/view and increase the character size.

oomek:
Just a little update. I’ve been testing this for the last week and I haven’t observed any adverse side effects.

oomek:
Could anyone review this pr and dress it with some nice flags please?

oomek:
It appears that no checks have been completed yet for this PR. Is it done automaticaly, or needs to be manually triggered? Also is there any chance to flag this PR for 2.5.1 release?

oomek:
I'd like to advocate for my PR even more. Have you ever tried slowly rotating the text? Without my commit the straight lines of glyphs loose antialiasing.

eXpl0it3r:
Squashed and merged in f4f7ef0d705ba3eb4f1711cc5b556ae3ed7df748

PR #1453, opened 2018-06-12
Added support for stencil testing.

binary1248:
A heavily modified version of @Kipernal's stencil testing implementation. Alpha testing was removed since it relies on legacy OpenGL and can be implemented using a shader anyway. The user is also required to explicitly request a context with stencil bits if they know they are going to need stencil testing.

Test with (paste on top of OpenGL example code):
```cpp
#include <cmath>
#include <SFML/Graphics.hpp>

int main()
{
    auto window = sf::RenderWindow{sf::VideoMode{{640u, 480u}}, "Test", sf::Style::Default, sf::ContextSettings{0, 8, 8}};
    window.setFramerateLimit(60);

    auto texture = sf::Texture{};
    texture.loadFromFile("resources/texture.jpg");

    auto sprite = sf::Sprite{texture};

    auto backgroundTexture = sf::Texture{};
    backgroundTexture.loadFromFile("resources/background.jpg");

    auto backgroundSprite = sf::Sprite{backgroundTexture};

    auto shape = sf::CircleShape{40.f};
    shape.setOutlineColor(sf::Color::Black);
    shape.setOutlineThickness(3.5f);
    shape.setFillColor(sf::Color::White);
    shape.setOrigin({40.f, 40.f});

    auto clock = sf::Clock{};

    while (window.isOpen())
    {
        {
                window.close();
        }

        // Clear color and stencil buffers
        window.clear(sf::Color::Blue, 0);

        // Move the sprite and circle around a bit
        sprite.setPosition({std::sinf(clock.getElapsedTime().asSeconds()) * 100.f, 0.f});
        shape.setPosition({std::cosf(clock.getElapsedTime().asSeconds()) * 100.f + 100.f, 256.f});

        // Draw our sprite as a stencil
        window.draw(sprite, sf::StencilMode{sf::StencilMode::Comparison::Always, sf::StencilMode::UpdateOperation::Replace, 1, ~0u, true});

        // Draw our circle shape as a stencil
        window.draw(shape, sf::StencilMode{sf::StencilMode::Comparison::Always, sf::StencilMode::UpdateOperation::Replace, 1, ~0u, true});

        // Draw our background only where we've previously drawn a stencil
        window.draw(backgroundSprite, sf::StencilMode{sf::StencilMode::Comparison::Equal, sf::StencilMode::UpdateOperation::Keep, 1, ~0u, false});

        // Move the sprite and circle around a bit more
        sprite.move(static_cast<sf::Vector2f>(texture.getSize()));
        shape.move({0.f, shape.getRadius() * 2.f});

        // Draw normal sprites and shapes unrelated to the stencil
        window.draw(sprite);
        window.draw(shape);

        window.display();
    }
}
```

Test code that uses a shader for alpha testing (paste on top of OpenGL example code):
```cpp
#include <SFML/Graphics.hpp>

int main()
{
    auto window = sf::RenderWindow{sf::VideoMode{{640u, 480u}}, "Test", sf::Style::Default, sf::ContextSettings{0, 8, 8}};
    window.setFramerateLimit(60);

    auto font = sf::Font{};
    font.loadFromFile("resources/sansation.ttf");

    auto text = sf::Text{font, "Hello Stencil World", 60};
    text.setStyle(sf::Text::Bold | sf::Text::Underlined);
    text.setPosition({60.f, 200.f});

    auto backgroundTexture = sf::Texture{};
    backgroundTexture.loadFromFile("resources/background.jpg");

    auto backgroundSprite = sf::Sprite{backgroundTexture};
    backgroundSprite.setPosition({0.f, -100.f});

    auto shader = sf::Shader{};
    shader.loadFromMemory(
        "uniform sampler2D texture;"
        ""
        "void main()"
        "{"
        "    gl_FragColor = texture2D(texture, gl_TexCoord[0]) * gl_Color;"
        ""
        "    if (gl_FragColor.a == 0.0)"
        "        discard;"
        "}"
        "",
        sf::Shader::Fragment);

    shader.setUniform("texture", sf::Shader::CurrentTexture);

    auto stencilPrepare = sf::RenderStates{sf::StencilMode{sf::StencilMode::Comparison::Always, sf::StencilMode::UpdateOperation::Replace, 1, ~0u, true}};
    stencilPrepare.shader = &shader;

    auto stencilApply = sf::RenderStates{sf::StencilMode{sf::StencilMode::Comparison::Equal, sf::StencilMode::UpdateOperation::Keep, 1, ~0u, false}};

    while (window.isOpen())
    {
        {
                window.close();
        }

        // Clear color and stencil buffers
        window.clear(sf::Color::Blue, 0);

        // Mark the area where the glyphs of the text is drawn
        window.draw(text, stencilPrepare);

        // Draw the background where the glyphs of the text were drawn
        window.draw(backgroundSprite, stencilApply);

        window.display();
    }
}
```


LaurentGomila:
Do we still try to have a "vision" for SFML, as an API on its own, or is it just becoming an object-oriented direct OpenGL wrapper?

binary1248:
Because "Stencil" isn't just an OpenGL feature... It is present in every single graphics API to date, Direct3D, OpenGL and even Vulkan which means something. I don't see the advantage in giving long standing graphics concepts new names just so that SFML can look like an abstraction for something everybody associates under another name.

We aren't here to teach people that wheels can also be called something else when pretty much everybody knows what a wheel is. There is enough literature and tutorials on the internet that teach these general concepts in an API and library independent way that you will have more people coming to us and asking "Does SFML support stencil testing?" than asking "What are clipping masks and what do I use them for?". That is of course ignoring the fact that beginners probably won't even make use of them that often. Do you want to tell them that what they learned as stencil testing is named Clipping Mask just for the sake of it?

> What I mean is, in the past we talked about masks, which is a way of exposing stencil features which is far more intuitive.

If you ask me this isn't even intuitive any more, forcing people to relearn something they might already know. We need to give people the tools they need to be productive. We aren't here to reinvent graphics programming. From the discussions I have seen about clipping masks everything is present, from set theory to pushing and popping masks on a stack. We mustn't forget, users might not come from the same background as us. The fact that we perceive something as "intuitive" doesn't make it intuitive for everyone. If you ask me, I would have a harder time getting stuff done using the clipping APIs discussed in the past than just manipulating the stencil buffer directly.

> Yes it is less flexible, but if one wants all the OpenGL features then why not using OpenGL?

Because interoperation with OpenGL is still a nightmare, and SFML still somehow manages to abstract away OpenGL pretty well. Shader, Texture, VertexBuffer, Stencil, all these concepts can be ported to other APIs as well. Why tell people to give up the portability just because we don't want to give the effort to provide an API independent abstraction of these concepts? If we aren't willing to provide the features people need and keep throwing out our "just use OpenGL then" excuse, why should new users favour SFML over e.g. SDL or GLFW? For the learning experience? Great... then the first few projects are implemented using SFML and once they have a grasp of the basics they jump off to a library that they can actually implement a real project with.

You should know best... `sf::PostFX` was changed into `sf::Shader` for pretty much the same reason.

LaurentGomila:
> Because "Stencil" isn't just an OpenGL feature... It is present in every single graphics API to date, Direct3D, OpenGL and even Vulkan which means something

So as all major features, because these APIs only reflect what graphics cards can do. In terms of features they are all more or less equivalent; what makes them different is how they manage errors, states, etc.
So to me, this is not an argument.

> I don't see the advantage in giving long standing graphics concepts new names just so that SFML can look like an abstraction

Then, again, just use OpenGL. Don't forget that you are an expert in low-level graphics APIs, and other people may not be familiar with what you consider "long standing graphics concepts". If you've learnt stencil buffer, then somehow you've learnt low-level APIs (whether it is OpenGL, D3D, Vulkan or whatever). If you don't know what the stencil is, then I strongly believe that you'll have a much better experience learning the mask API of SFML.

> We aren't here to teach people that wheels can also be called something else when pretty much everybody knows what a wheel is.

If you don't want to invent anything then we're back to an OpenGL wrapper. This is clearly not what SFML is. We're supposed to provide our own concepts and abstractions, and then eventually map them to what low-level APIs provide. People who are already familiar with stencil and all these low-level graphics concepts might already know OpenGL and then have no interest in SFML.

Forgetting your background and what you know about graphics APIs, tell me honestly which code looks more intuitive / readable / maintainable from these two:

```cpp
window.draw(sprite, sf::StencilMode(sf::StencilMode::Always, sf::StencilMode::Replace, 1, ~0, true));
window.draw(shape, sf::StencilMode(sf::StencilMode::Always, sf::StencilMode::Replace, 1, ~0, true));
window.draw(backgroundSprite, sf::StencilMode(sf::StencilMode::Equal, sf::StencilMode::Keep, 1, ~0, false));
```

```cpp
mask.add(sprite);
mask.add(shape);
window.draw(backgroundSprite, mask);
```

binary1248:
> So as all major features, because these APIs only reflect what graphics cards can do. In terms of features they are all more or less equivalent; what makes them different is how they manage errors, states, etc.
So to me, this is not an argument.

Actually this isn't even 100% true. Both OpenGL and Direct3D have had bad reputations in the past for emulating features on the CPU that weren't able to be hardware accelerated. In fact this is one of the biggest problems with OpenGL compared to Direct3D, you can't even find out if a feature is hardware accelerated or not.

Why would the driver developers go through the effort of implementing emulation for something that isn't provided by the hardware? Because the graphics APIs themselves are abstractions. Shaders are nothing else but a region of memory containing instructions for the compute units on GPUs. Textures are nothing else but a region of memory containing data to be used in a specific way, and the list goes on. The big APIs have more or less agreed on what abstractions a graphics API should provide in general, and we are left with what we have now.

If we reinvented something on top of stencil functionality it would only be an abstraction on top of an already existing abstraction. Where is it going to end?

> Then, again, just use OpenGL. Don't forget that you are an expert in low-level graphics APIs, and other people may not be familiar with what you consider "long standing graphics concepts".

You don't have to be an expert to use the stencil buffer, in fact it shows up relatively early in many tutorials that are available, whether they are D3D, OpenGL, or API independent tutorials/literature. I can even imagine that a programmer who is experienced in D3D coming to use SFML will be able to use this feature fairly quickly because they already know what the stencil buffer is used for but aren't really in the mood to learn yet another graphics API like OpenGL. Saying "you know what the stencil buffer is, so just use OpenGL" is over-generalizing. There are people who know what the stencil buffer is but don't know OpenGL, and they must not be under-represented.

> If you've learnt stencil buffer, then somehow you've learnt low-level APIs (whether it is OpenGL, D3D, Vulkan or whatever). If you don't know what the stencil is, then I strongly believe that you'll have a much better experience learning the mask API of SFML.

This is also one of the problems that will come. If we are going to invent a completely new graphics programming workflow, the burden is on us to provide enough documentation and support so people can actually do what they want to do with it. Nobody else on the internet is going to explain what you can or can't do with clipping masks in a programming context.

Imagine this discussion:

User: "I want to implement SuperGameFeatureA, how would I do this in SFML?"
Team: "Read the tutorial on Clipping Masks, they do what you need."
User: "OK, I read the tutorial, and it sort of does what I want, but I also need to do SuperGameFeatureB, is that possible?"
Team: "Well, the tutorial doesn't explicitly mention it but you can do it like this: XYZ"
User: "But that isn't really what I was aiming for, (insert pictures here). Is there really no way to do this using Clipping Masks?"
Team: "No, that is an advanced feature, you are better off using OpenGL directly."
User: "Oh well, I guess I have to learn OpenGL as well now. I was hoping I could implement that last 5% with sfml-graphics but it seems like it isn't the case."

This is essentially how many discussions have ended up in the past, and I can see many discussions ending up like this in the future because we have to invent and more importantly _support_ our own concepts. If we exposed the stencil buffer directly to the user, they can refer to many other tutorials on the internet that probably explain the concept much better than we will ever do. I don't think the cost of maintaining our own concepts are worth it.

> If you don't want to invent anything then we're back to an OpenGL wrapper. This is clearly not what SFML is.

To be fair, the majority of sfml-graphics already maps more or less directly to objects in any graphics API (not just OpenGL). Sure, this was never the intention, but we ended up here because we provide a programming library for people to get stuff done with. If it turns out that the established concepts within the graphics APIs are the simplest and more importantly the best performing way of getting things done, what is the point of building yet another artificial layer on top of it?

> We're supposed to provide our own concepts and abstractions, and then eventually map them to what low-level APIs provide.

And theoretically, if it could even be measured, if graphics APIs provide an already optimal abstraction, would we still insist on building yet another one on top of it just for the sake of it? Instead of constantly saying "we need more abstractions", we should take a step back and ask ourselves whether users would even have difficulties with the abstractions already provided by the graphics APIs. If the answer is no (and I strongly believe this) then it doesn't make much sense investing effort into something that overcomplicates the concept.

> People who are already familiar with stencil and all these low-level graphics concepts might already know OpenGL and then have no interest in SFML.

There are enough people, myself included, who prefer using SFML to get moderately simple things done than rolling yet another OpenGL application. I could do it if I wanted to, but if I can get the same effects out of SFML in 10% of the time why shouldn't I? You are basically forcing me to write 4x more code just because of that last 5% of functionality that Clipping Masks shouldn't provide by design.

Programming libraries are meant to aid developers in being more productive with their time. It isn't always a black and white world, where it's either "use SFML and all the concepts we impose on you" or "just use OpenGL". People try really hard to get as much done in sfml-graphics as possible. But when they feel their artistic freedom is unnecessarily compromised, they first try begging us to support something they need, then they try to hack something together mixing OpenGL and sfml-graphics, and then in the end they finally give up and rewrite their whole application using OpenGL. At that point why are they even still using SFML? We seem to have damned ourselves to having to work exclusively with beginners for eternity, because once you actually know what you need, you won't be served by SFML any longer.

> Forgetting your background and what you know about graphics APIs, tell me honestly which code looks more intuitive / readable / maintainable from these two:

From a readability point of view, sure, where I have to type less it is also more readable. From a intuition or maintainability point of view, I would still prefer a single point of contact with the API that does everything I can possibly need. Maybe it should be mentioned, but besides `sf::StencilMode` there is nothing else you have to understand to get your work done. This along with what the stencil buffer is, is one "concept" that has to be learned as opposed to all the different operations you would be able to perform on a tentative ClippingMask object. The example you provide compares 100% of the currently proposed API to X% of the ClippingMask API, and X depends on what else has to be supported in order for people to make productive use of it.

I'm sure you are still aware of the lengthy discussions both on #846 and [here](https://en.sfml-dev.org/forums/index.php?topic=18042.0). This is the typical design by committee process where something really simple is taken and blown out of proportion for diverse reasons. Even I had a hard time figuring out what is it that would go into such a ClippingMask object. Add, Remove, Union/Combine, Push, Pop, etc. This is also the reason why I split out scissor testing as well. We don't want to have god objects.

You yourself said on the forum to @Kipernal's API:
> I haven't looked at the implementation yet, but I like the API. I would just replace all these technical terms with friendlier names: from a user point-of-view, it would be more intuitive to talk about masks, and adding to/drawing inside/drawing outside/ignoring them instead of these stencil buffer operations.

Besides renaming various things back to their graphics API counterpart and removing all the pre-defined modes (which don't really provide any value), I didn't do much with the code, which is also the reason why I left @Kipernal as the author.

If you want, I can provide even more examples that would work with this proposal but lead to a nightmare when having to juggle and manipulate ClippingMask objects. Not to mention, I still have no idea how such a mask would even be implemented in a safe and efficient way.

include/SFML/Graphics/StencilMode.hpp
```diff
@@ -0,0 +1,210 @@
...
+/// stencil value the new value will be clamped to the range from 0 to the
+/// maximum representable value given the bit width of the stencil buffer
+/// e.g. 255 if an 8-bit stencil buffer is being used.
+///
+/// The reference value is used both during the comparison with the current
+/// stencil buffer value and as the new value to be written when the operation
+/// is set to Replace.
+///
+/// The mask value is used to mask the bits of both the reference value and
+/// the value in the stencil buffer during the comparison and when updatinig.
```
eXpl0it3r: Typo in `updatinig`

binary1248:
Fixed and rebased.

eXpl0it3r:
The term "clipping mask" seems to often be used in relation to image manipulation software such as Photoshop and GIMP while "stencil buffer" is a very common term in graphics programming independent of low-level API.

From my point of view, going with a masking approach might make it easier to initially understand the concept and might even be easier to just go and add some simple masks. But to me the proposed designs quickly become rather cumbersome programmatically and in their usage: Addition state to track, creating all sorts of objects to create as mask from, apply mask order, etc.

Going with a classic stencil approach may require some more upfront reading and maybe some more effort to get the first quick-win, but there exist a lot more resources on how to use a stencil buffer programatically and giving more or less direct access to the lower-level functions, gives the user all the power they want.

I personally don't see the advantage of using masks rather than providing access to how masks are most likely implemented. If anything one could potentially consider some clipping mask implementation on top of this stencil buffer implementation.

----

Would pre-defined modes make sense here (similar to blend modes), e.g. `sf::StencileMode::Default`?

ventsin:
I'm a fan of the idea of implementing direct stencil tools and building abstractions on top of them, much like how the audio module is implemented. I think this is important to consider here because the current design doesn't feel too `Simple` to me. What I mean by this is that it seems too different from the rest of the SFML approach.

At the same time, not taking advantage of the full capabilities of the stencil technology seems rather disappointing, if the point is to make an API of it's own. Since a working prototype is already available here, not being able to make use of this would be a waste of perfectly good code.

If the matter is with naming, I'd like to bring these two images to your attention:
image
image
You see, there is a tutorial that explains how to render your `sf::Texture`s using something called `sf::Sprite`, which internally uses something known as an array of `sf::Vertex`. And in the same category, there is a tutorial detailing how to use these tools yourself to build different things.

My point is, I see no reason as to why there couldn't be some kind of `sf::StencilBuffer` wrapper around OpenGL (and in the future maybe Vulkan), with a simple to use `sf::Mask` abstraction built on top of it. And with tutorials that compliment each other much like with the above examples, this approach would probably solve both issues of being hard to grasp and being lacking in features.

binary1248:
Updated to current master.

include/SFML/Graphics/RenderStates.hpp
```diff
@@ -103,6 +113,22 @@ class SFML_GRAPHICS_API RenderStates
...
+    /// \param theTexture         Texture to use
+    /// \param theShader          Shader to use
+    ///
+    ////////////////////////////////////////////////////////////
+    RenderStates(const BlendMode&   theBlendMode,
+                 const StencilMode& theStencilMode,
+                 const Transform&   theTransform,
+                 const Texture*     theTexture,
+                 const Shader*      theShader);
+
```
ChrisThrasher, line 131: I suppose this constructor was added to be API compatible during the SFML 2 lifecycle. Because we're breaking the API, can we instead just add this extra parameter to the existing 4-parameter constructor?

include/SFML/Graphics/StencilMode.hpp
```diff
@@ -0,0 +1,214 @@
...
+////////////////////////////////////////////////////////////
+struct SFML_GRAPHICS_API StencilMode
+{
+    ////////////////////////////////////////////////////////
+    /// \brief Enumeration of the stencil test comparisons that can be performed
+    ///
+    /// The comparisons are mapped directly to their OpenGL equivalents,
+    /// specified by glStencilFunc().
+    ////////////////////////////////////////////////////////
+    enum Comparison
```
ChrisThrasher, line 49: Can we use a scope enumeration for `Comparison` and `UpdateOperation`?

include/SFML/Graphics/StencilMode.hpp
```diff
@@ -0,0 +1,214 @@
...
+
+
+namespace sf
+{
+
+////////////////////////////////////////////////////////////
+/// \brief Stencil modes for drawing
+///
+////////////////////////////////////////////////////////////
+struct SFML_GRAPHICS_API StencilMode
```
ChrisThrasher, line 41: Can you add unit tests for `StencileMode`? The [ContextSettings tests](https://github.com/SFML/SFML/blob/master/test/Window/ContextSettings.cpp) are pretty similar if you want to know what kind of boilerplate is involved.

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -171,6 +215,35 @@ void RenderTarget::clear(const Color& color)
 }
 
 
+////////////////////////////////////////////////////////////
+void RenderTarget::clear(int value)
+{
+    if (RenderTargetImpl::isActive(m_id) || setActive(true))
+    {
+        // Unbind texture to fix RenderTexture preventing clear
+        applyTexture(NULL);
```
ChrisThrasher, line 224:
```suggestion
        applyTexture(nullptr);
```

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -171,6 +215,35 @@ void RenderTarget::clear(const Color& color)
...
+}
+
+
+////////////////////////////////////////////////////////////
+void RenderTarget::clear(const Color& color, int value)
+{
+    if (RenderTargetImpl::isActive(m_id) || setActive(true))
+    {
+        // Unbind texture to fix RenderTexture preventing clear
+        applyTexture(NULL);
```
ChrisThrasher, line 238:
```suggestion
        applyTexture(nullptr);
```

src/SFML/Graphics/RenderStates.cpp
```diff
@@ -35,24 +35,35 @@ namespace sf
...
-    BlendMode::Add));
+// For similar reasons we need to provide the default stencil mode, and thus all other settings along with it
+const RenderStates RenderStates::Default(
+    BlendMode(BlendMode::SrcAlpha,
+              BlendMode::OneMinusSrcAlpha,
+              BlendMode::Add,
+              BlendMode::One,
+              BlendMode::OneMinusSrcAlpha,
+              BlendMode::Add),
+    StencilMode(StencilMode::Always, StencilMode::Keep, 0, ~0u, false),
```
ChrisThrasher, line 46: The `StencilMode` default constructor already uses these same values. Why don't we just default-initialize this member variable instead?

src/SFML/Graphics/StencilMode.cpp
```diff
@@ -0,0 +1,74 @@
...
+{
+}
+
+
+////////////////////////////////////////////////////////////
+bool operator==(const StencilMode& left, const StencilMode& right)
+{
+    return (left.stencilUpdateOperation == right.stencilUpdateOperation) &&
+           (left.stencilComparison == right.stencilComparison) && (left.stencilReference == right.stencilReference) &&
+           (left.stencilMask == right.stencilMask) && (left.stencilOnly == right.stencilOnly);
```
ChrisThrasher, line 64: I'd prefer we omit the extra parentheses, but I understand why you used them.

ChrisThrasher (changes requested):

binary1248:
Fixed issues, added tests and updated to current master.

include/SFML/Graphics/StencilMode.hpp
```diff
@@ -0,0 +1,214 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2022 Laurent Gomila (laurent@sfml-dev.org)
```
ChrisThrasher, line 4:
```suggestion
// Copyright (C) 2007-2023 Laurent Gomila (laurent@sfml-dev.org)
```
Check the other copyright headers too.

include/SFML/Graphics/StencilMode.hpp
```diff
@@ -0,0 +1,214 @@
...
+//
+// 2. Altered source versions must be plainly marked as such,
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+#ifndef SFML_STENCILMODE_HPP
+#define SFML_STENCILMODE_HPP
```
ChrisThrasher, line 26: We've switched to using `#pragma once` everywhere.

include/SFML/Graphics/StencilMode.hpp
```diff
@@ -0,0 +1,214 @@
...
+                int             theStencilReference,
+                unsigned int    theStencilMask,
+                bool            theStencilOnly);
+
+    ////////////////////////////////////////////////////////////
+    // Member Data
+    ////////////////////////////////////////////////////////////
+    Comparison stencilComparison{Comparison::Always}; //!< The comparison we're performing the stencil test with
+    UpdateOperation stencilUpdateOperation{UpdateOperation::Keep}; //!< The update operation to perform if the stencil test passes
+    int             stencilReference{0}; //!< The reference value we're performing the stencil test with
```
ChrisThrasher, line 106:
```suggestion
    int             stencilReference{}; //!< The reference value we're performing the stencil test with
```

include/SFML/Graphics/StencilMode.hpp
```diff
@@ -0,0 +1,214 @@
...
+    unsigned int stencilMask{~0u}; //!< The mask to apply to both the reference value and the value in the stencil buffer
+    bool         stencilOnly{false}; //!< Whether we should update the color buffer in addition to the stencil buffer
```
ChrisThrasher, line 108:
```suggestion
    bool         stencilOnly{}; //!< Whether we should update the color buffer in addition to the stencil buffer
```

ChrisThrasher (changes requested):

binary1248:
Fixed.

include/SFML/Graphics/StencilMode.hpp
```diff
@@ -0,0 +1,210 @@
...
+////////////////////////////////////////////////////////////
+struct SFML_GRAPHICS_API StencilMode
+{
+    ////////////////////////////////////////////////////////
+    /// \brief Enumeration of the stencil test comparisons that can be performed
+    ///
+    /// The comparisons are mapped directly to their OpenGL equivalents,
+    /// specified by glStencilFunc().
+    ////////////////////////////////////////////////////////
+    enum class Comparison
```
kimci86, line 48: Is the word "comparison" the usual word in that context? Wouldn't "condition" or "test" for example be more correct?

include/SFML/Graphics/StencilMode.hpp
```diff
@@ -0,0 +1,210 @@
...
+///
+/// The stencil comparison specifies the comparison that is performed between
+/// the reference value of the currently active sf::StencilMode and the value
+/// that is currently in the stencil buffer. This comparison determines whether
+/// the stencil test passes or fails.
+///
+/// The stencil update operation specifies how the stencil buffer is updated if
+/// the stencil test passes. If the stencil test fails, neither the color or
+/// stencil buffers will be modified. If incrementing or decrementing the
+/// stencil value the new value will be clamped to the range from 0 to the
```
kimci86, line 187:
```suggestion
/// stencil value, the new value will be clamped to the range from 0 to the
```

src/SFML/Graphics/RenderStates.cpp
```diff
@@ -35,13 +35,18 @@ namespace sf
...
 // We cannot use the default constructor here, because it accesses BlendAlpha, which is also global (and dynamically
 // initialized). Initialization order of global objects in different translation units is not defined.
-const RenderStates RenderStates::Default(BlendMode(
-    BlendMode::SrcAlpha,
-    BlendMode::OneMinusSrcAlpha,
-    BlendMode::Add,
-    BlendMode::One,
-    BlendMode::OneMinusSrcAlpha,
-    BlendMode::Add));
+// For similar reasons we need to provide the default stencil mode, and thus all other settings along with it
```
kimci86, line 38: Is that true? I don't get it.

include/SFML/Graphics/StencilMode.hpp
```diff
@@ -0,0 +1,210 @@
...
+/// The arithmetic comparison that is performed on the 2 values can also be
+/// controlled via sf::StencilMode. Depending on whether the test passes i.e.
+/// the comparison yields true, the color buffer is updated with its new RGBA
+/// value and if set in sf::StencilMode the stencil buffer is updated
+/// accordingly. The new stencil value will be used during stencil testing the
+/// next time the pixel is drawn to.
+///
+/// The class is composed of 4 components, each of which has its
+/// own public member variable:
+/// \li Stencil Comparison (@ref stencilComparison)
```
kimci86, line 173: For consistency:
```suggestion
/// \li Stencil Comparison (\ref stencilComparison)
```

kimci86 (commented):
Some nitpicking

include/SFML/Graphics/StencilMode.hpp
binary1248: The "test" refers to the operation that happens at a specific pipeline stage. It is what is done. Condition could imply that arbitrary conditions can be tested. Comparison describes what is actually performed as part of the test: a comparison of 2 integer values.

See: https://registry.khronos.org/OpenGL-Refpages/gl4/html/glStencilFunc.xhtml where the term comparison is also used to describe the technical implementation of the test.

src/SFML/Graphics/RenderStates.cpp
binary1248: No it's just a leftover from the initial implementation, will be removed.

include/SFML/Graphics/StencilMode.hpp
kimci86: Thank you for the explanation. I was asking because Never and Always do not look like comparisons to me. Anyways it is used in OpenGL documentation, so I guess we can keep it.

binary1248:
Updated.

danieljpetersen:
I've switched the build of SFML that I use to this branch.  The work you all have done on this is very much appreciated! I feel pretty strongly that this feature deserves to be in SFML

I agree that the API here is not as beginner friendly as the other SFML API's, but I do think it's a good API and not hard to use if you understand the concepts. It seems wrong to not expose the full functionality of the underlying graphics API's when we have working code which does a nice job of it. I want to echo ventsin's thought about an ideal world where SFML has two layers of abstraction, one layer designed to be easy to use built on top of a lower layer which exposes more details for those that need it. 

I had to make some small updates to the test code in the first comment to get it compiling on the latest master / gcc (which has std::sinf as just sinf apparently).  

```cpp
#include <SFML/Graphics.hpp>
#include <iostream>
#include <cmath>

int main()
{
    sf::RenderWindow window(sf::VideoMode({640, 480}), "Test", sf::Style::Default, sf::ContextSettings(0, 8));
    window.setFramerateLimit(60);

    sf::Texture texture;
    texture.loadFromFile("resources/texture.jpg");

    sf::Sprite sprite(texture);

    sf::Texture backgroundTexture;
    if (!backgroundTexture.loadFromFile("resources/background.jpg"))
    {
        std::cout << "Error loading background.jpg" << std::endl;
    }

    sf::Sprite backgroundSprite(backgroundTexture);

    sf::CircleShape shape;
    shape.setOutlineColor(sf::Color::Black);
    shape.setOutlineThickness(3.5f);
    shape.setFillColor(sf::Color::White);
    shape.setRadius(40);
    shape.setOrigin({40, 40});

    sf::Clock clock;

    while (window.isOpen())
    {
        sf::Event event;

        {
                window.close();
        }

        // Clear color and stencil buffers
        window.clear(sf::Color::Blue, 0);

        // Move the sprite and circle around a bit
        sprite.setPosition({sinf(clock.getElapsedTime().asSeconds()) * 100, 0});
        shape.setPosition({cosf(clock.getElapsedTime().asSeconds()) * 100 + 100, 256});

        // Draw our sprite as a stencil
        window.draw(sprite, sf::StencilMode(sf::StencilMode::Comparison::Always, sf::StencilMode::UpdateOperation::Replace, 1, ~0, true));

        // Draw our circle shape as a stencil
        window.draw(shape, sf::StencilMode(sf::StencilMode::Comparison::Always, sf::StencilMode::UpdateOperation::Replace, 1, ~0, true));

        // Draw our background only where we've previously drawn a stencil
        window.draw(backgroundSprite, sf::StencilMode(sf::StencilMode::Comparison::Equal, sf::StencilMode::UpdateOperation::Keep, 1, ~0, false));

        // Move the sprite and circle around a bit more
        sprite.move(static_cast<sf::Vector2f>(texture.getSize()));
        shape.move({0, shape.getRadius() * 2});

        // Draw normal sprites and shapes unrelated to the stencil
        window.draw(sprite);
        window.draw(shape);

        window.display();
    }
}
```

ChrisThrasher (approved):

include/SFML/Graphics/RenderTarget.hpp
```diff
@@ -82,6 +83,29 @@ class SFML_GRAPHICS_API RenderTarget
...
+    /// \brief Clear the entire target with a single color and stencil value
+    ///
+    /// The specified stencil value is truncated to the bit
+    /// width of the current stencil buffer.
+    ///
+    /// \param color Fill color to use to clear the render target
+    /// \param value Stencil value to clear to
+    ///
+    ////////////////////////////////////////////////////////////
+    void clear(const Color& color, int value);
```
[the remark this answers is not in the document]
Bromeon: I guess having all combinations of (color, stencil) is fine for 2 arguments, but this might not scale that well in case we ever want to provide more parameters for `clear()`. Not sure if that's realistic, we can always address such a problem only when we have it (in the worst case by deprecating existing `clear()` overloads).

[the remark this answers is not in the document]
ChrisThrasher: > or should we use std::int32_t here?

https://github.com/SFML/SFML/issues/2325 

I'd generally prefer we err more on the side of sized integers. In practice though `int` is 32 bits on everything but Win16 platforms so in this particular case it's a safe assumption. I'll do the work to move us over to more fixed width integer types if the SFML team can come to agreement on how to use them.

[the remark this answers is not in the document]
binary1248: `value` depends on the size of the stencil buffer, which itself depends on what is supported by the system and what the user requests during window creation.

`int` was chosen here because that is also what the underlying API uses. I personally also don't know how larger values would be passed if some hypothetical system were to support stencil buffers with more than 32 bits.

Usually even if a system supported more than e.g. 8 bits only 8 bits are requested so as not to waste memory. Typical rendering techniques using the stencil buffer don't need it to be that large.

Using a fixed width type here wouldn't feel right. The default type for numeric literals whether in decimal or hexadecimal notation is `int` and those are the 2 most common ways of clearing the stencil buffer.

include/SFML/Graphics/StencilMode.hpp
```diff
@@ -0,0 +1,210 @@
...
+    ////////////////////////////////////////////////////////
+    enum class UpdateOperation
+    {
+        Keep,      //!< If the stencil test passes, the value in the stencil buffer is not modified
+        Zero,      //!< If the stencil test passes, the value in the stencil buffer is set to zero
+        Replace,   //!< If the stencil test passes, the value in the stencil buffer is set to the new value
+        Increment, //!< If the stencil test passes, the value in the stencil buffer is incremented and if required clamped
+        Decrement, //!< If the stencil test passes, the value in the stencil buffer is decremented and if required clamped
+        Invert,    //!< If the stencil test passes, the value in the stencil buffer is bitwise inverted
+    };
```
eXpl0it3r, line 74: Would it make sense to move this enum class out side the struct, as the access to the enum values is rather long now.

```cpp
sf::StencilMode::Equal
// vs
sf::StencilMode::Comparison::Equal
```

Maybe something like `sf::StencileComparsion` would work.

include/SFML/Graphics/StencilMode.hpp
```diff
@@ -0,0 +1,210 @@
...
+    {
+        Never,     //!< The stencil test never passes
+        Less,      //!< The stencil test passes if the new value is less than the value in the stencil buffer
+        LessEqual, //!< The stencil test passes if the new value is less than or equal to the value in the stencil buffer
+        Greater,   //!< The stencil test passes if the new value is greater than the value in the stencil buffer
+        GreaterEqual, //!< The stencil test passes if the new value is greater than or equal to the value in the stencil buffer
+        Equal,        //!< The stencil test passes if the new value is strictly equal to the value in the stencil buffer
+        NotEqual, //!< The stencil test passes if the new value is strictly inequal to the value in the stencil buffer
+        Always    //!< The stencil test always passes
+    };
```
eXpl0it3r, line 58: Same point as for `sf::StencilMode::Comparison`

binary1248:
Updated.

eXpl0it3r (dismissed):

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -147,6 +147,50 @@ std::uint32_t equationToGlConstant(sf::BlendMode::Equation blendEquation)
...
+        case sf::StencilComparison::Greater:      return GL_GREATER;
+        case sf::StencilComparison::GreaterEqual: return GL_GEQUAL;
+        case sf::StencilComparison::Equal:        return GL_EQUAL;
+        case sf::StencilComparison::NotEqual:     return GL_NOTEQUAL;
+        case sf::StencilComparison::Always:       return GL_ALWAYS;
+    }
+    // clang-format on
+
+    sf::err() << "Invalid value for sf::StencilComparison! Fallback to sf::StencilMode::Always." << std::endl;
+    assert(false);
```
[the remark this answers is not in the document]
ChrisThrasher: Elsewhere we've kept `assert(false)` when there's an `sf::err()` log message right above it since otherwise we'd just be writing the same string twice

binary1248:
Updated.

test/Graphics/StencilMode.test.cpp
```diff
@@ -0,0 +1,299 @@
+#include <SFML/Graphics/Image.hpp>
+#include <SFML/Graphics/RectangleShape.hpp>
+#include <SFML/Graphics/RenderTexture.hpp>
+#include <SFML/Graphics/StencilMode.hpp>
```
ChrisThrasher, line 4: https://github.com/SFML/SFML/blob/00b6488c3e9350bda573e9a82a06a1d855fa494b/test/Graphics/Texture.test.cpp#L3

Can you please follow the pattern here to ensure StencilMode.hpp is the first header included? This ensures that we're testing that each header can be compiled on its own.

(Side note, CMake 3.26 adds a new feature that lets us test that each header can be compiled on its own but 3.26 is out of reach for the time unless we decide to upgrade our CMake requirement once again. This feature is a lot nicer than having to fight clang-format to ensure the correct header is included first.)

ChrisThrasher (changes requested):

ChrisThrasher (approved):

ChrisThrasher:
Rebased and fixed conflicts

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -529,11 +603,13 @@ void RenderTarget::resetGLStates()
         glCheck(glEnableClientState(GL_VERTEX_ARRAY));
         glCheck(glEnableClientState(GL_COLOR_ARRAY));
         glCheck(glEnableClientState(GL_TEXTURE_COORD_ARRAY));
-        m_cache.scissorEnabled = false;
```
kimci86, line 532: I guess deleting this line was a conflict resolution mistake.

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -171,6 +215,35 @@ void RenderTarget::clear(const Color& color)
...
+////////////////////////////////////////////////////////////
+void RenderTarget::clear(int value)
+{
+    if (RenderTargetImpl::isActive(m_id) || setActive(true))
+    {
+        // Unbind texture to fix RenderTexture preventing clear
+        applyTexture(nullptr);
+
+        glCheck(glClearStencil(value));
+        glCheck(glClear(GL_STENCIL_BUFFER_BIT));
```
kimci86, line 227: It looks to me the code related to applying the view for scissor testing we can see in `RenderTarget::clear(const Color&)` method above is missing here.
Same remark for the other `clear` method below.

test/Graphics/StencilMode.test.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+            CHECK(sf::StencilMode() !=
+                  sf::StencilMode(sf::StencilComparison::Equal, sf::StencilUpdateOperation::Replace, 1, 0u, true));
+            CHECK(sf::StencilMode(sf::StencilComparison::Greater, sf::StencilUpdateOperation::Invert, 0, ~0u, false) !=
+                  sf::StencilMode(sf::StencilComparison::Equal, sf::StencilUpdateOperation::Replace, 1, 0u, true));
+        }
+    }
+}
+
+TEST_CASE("[Graphics] sf::StencilMode Draw", runDisplayTests())
```
ChrisThrasher, line 72: I think I'd rather put all of these tests behind `runDisplayTests` than have two `TEST_CASE`s.

ChrisThrasher (changes requested):

ChrisThrasher:
Rebased and addressed some of my review comments. Not much more to say about this. I'm ready to merge this pending a few more review comments.

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -663,6 +739,41 @@ void RenderTarget::applyBlendMode(const BlendMode& mode)
...
 
 
+////////////////////////////////////////////////////////////
+void RenderTarget::applyStencilMode(const StencilMode& mode)
+{
+    using RenderTargetImpl::stencilFunctionToGlConstant;
+    using RenderTargetImpl::stencilOperationToGlConstant;
+
+    // Fast path if we have a default (disabled) stencil mode
+    if (mode == StencilMode())
```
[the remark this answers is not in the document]
binary1248: The default constructed `StencilMode` should correspond to the default state of stencil operations in a newly created OpenGL context. In this case, rendering is performed as if no stencil testing is required. Because there is always a stencil state (even if it effectively doesn't do anything) in OpenGL, using an `optional` to model this wouldn't make sense.

include/SFML/Graphics/StencilMode.hpp
```diff
@@ -0,0 +1,251 @@
...
+    /// \brief Construct a stencil mode given the function, operation and reference value.
+    ///
+    /// \param theStencilComparison      Specifies the kind of test to perform against the value in the stencil buffer
+    /// \param theStencilUpdateOperation Specifies how the stencil buffer is updated if the test passes
+    /// \param theStencilReference       Specifies the reference value to perform the stencil test with and to write into stencil buffer if the operation is Replace and the test passes
+    /// \param theStencilMask            Specifies the mask value to apply to both the reference value and the value in the stencil buffer when comparing and updating
+    /// \param theStencilOnly            True to update only the stencil buffer, false to update both stencil and color buffers
+    ///
+    ////////////////////////////////////////////////////////////
+    StencilMode(StencilComparison      theStencilComparison,
```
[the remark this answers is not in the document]
binary1248: This was done analogous to the `RenderStates` constructor. If we want to clean this up at some later point we should do it all at once instead of one at a time.

binary1248:
Rebased and updated. This should be the final version ready for review and merging.

include/SFML/Graphics/StencilMode.hpp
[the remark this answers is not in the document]
ChrisThrasher: I don't think repeating the mistakes of `RenderStates` is the right call here. We're already so close to have an aggregate type. We just have to delete this single constructor. `RenderStates` has 6 constructors of which we'd have to remove all 6 to make it an aggregate.

test/Graphics/StencilMode.test.cpp
```diff
@@ -0,0 +1,278 @@
...
+                        sf::StencilMode(sf::StencilComparison::Equal, sf::StencilUpdateOperation::Replace, 1, 0u, true));
+
+            CHECK(sf::StencilMode() !=
+                  sf::StencilMode(sf::StencilComparison::Equal, sf::StencilUpdateOperation::Replace, 1, 0u, true));
+            CHECK(sf::StencilMode(sf::StencilComparison::Greater, sf::StencilUpdateOperation::Invert, 0, ~0u, false) !=
+                  sf::StencilMode(sf::StencilComparison::Equal, sf::StencilUpdateOperation::Replace, 1, 0u, true));
+        }
+    }
+
+    sf::RenderTexture renderTexture;
```
ChrisThrasher, line 73: `StencilMode` itself is dead simple. It's just a simple struct. We just need to make sure the one constructor works and if we remove that constructor then we have even less code to test. I think that's where the tests should end.

If we want to test how it's used in rendering then I think those tests belong with the render texture tests.

ChrisThrasher (changes requested):

include/SFML/Graphics/StencilMode.hpp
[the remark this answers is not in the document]
ChrisThrasher: https://github.com/SFML/SFML/compare/master...aggregate_render_states

Turns out it's not so hard to remove all those `RenderStates` constructors. At least internally we had little use for many of those constructors.

binary1248:
Updated. Dumped all the rendering tests into the new Render.test.cpp that can be used for future full rendering pipeline tests as well.

ChrisThrasher:
> Dumped all the rendering tests into the new Render.test.cpp that can be used for future full rendering pipeline tests as well.

All of off our test files correspond to a public header of the same name. It seems odd to make this new test file when there is no Render.hpp and when these are ultimately just more `sf::RenderTexture` tests.

binary1248:
We have to use something to test functionality, by your logic we could also put these in the `Image` tests just because `Image` is used as well. This code doesn't test any `RenderTexture` specific functionality, it could have also been written to use `RenderWindow` instead, `RenderTexture` was chosen solely due to performance reasons. There will always come a point where we have to test how multiple different things interact with one another, these tests don't belong to any one of the existing types. I just picked `Render.test.cpp` because I had no other idea what to call the file. The only reason it looks strange right now is because it is the first file of its kind.

ChrisThrasher:
I now see why it makes sense to put lots of rendering tests in a single file. I’m okay with having Render.test.cpp.

ChrisThrasher (approved):

binary1248:
Rebased onto master.

include/SFML/Graphics/RenderTarget.hpp
```diff
@@ -96,6 +97,29 @@ class SFML_GRAPHICS_API RenderTarget
...
+    ////////////////////////////////////////////////////////////
+    /// \brief Clear the stencil buffer to a specific value
+    ///
+    /// The specified value is truncated to the bit width of
+    /// the current stencil buffer.
+    ///
+    /// \param value Stencil value to clear to
+    ///
+    ////////////////////////////////////////////////////////////
+    void clear(int value);
```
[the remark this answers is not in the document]
binary1248: Renamed to `clearStencil` and introduced a dedicated type that disables unintended conversions.

ChrisThrasher (approved):

eXpl0it3r (approved):

PR #1454, opened 2018-06-13
Fixed two conversion warnings (C4267)

Ironbell:

----

## Description

Two small fixes for two argument conversion warnings (C4267) appearing on Microsoft Visual C++, Version MSVC 19.13.26128.0, x64 and x86.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Compile the code with the mentioned compiler and check that these two warnings don't appear
image

### DIFF
--- a/src/SFML/Graphics/RenderTarget.cpp
+++ b/src/SFML/Graphics/RenderTarget.cpp
@@ -708,7 +708,7 @@ void RenderTarget::drawPrimitives(PrimitiveType type, std::size_t firstVertex, s
     GLenum mode = modes[type];
 
     // Draw the primitives
-    glCheck(glDrawArrays(mode, firstVertex, static_cast<GLsizei>(vertexCount)));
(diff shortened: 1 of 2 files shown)


eXpl0it3r:
Not sure if there's an easy fix for the `::send()` error as Linux seems to use `size_t` while Windows uses `int`.

https://msdn.microsoft.com/en-us/library/windows/desktop/ms740149(v=vs.85).aspx
https://linux.die.net/man/2/send

At least on the CI there doesn't seem to be any warnings.

JonnyPtn:
Strangely, I don't get any warnings whatsoever when building master for macOS, even with `-Wall`... which makes me suspicious. Anyway that's preventing me really testing this for macOS.

on iOS, the warning in sfml-graphics is fixed in this PR, but not the network one, for the same reason @eXpl0it3r mentions (`send()` returns `ssize_t`)

eXpl0it3r:
Merged in 44d3e267663d573fc7ec1079cae6ca78007a9b37

PR #1459, opened 2018-07-22
Reduced context locking & unlocking while creating textures

eXpl0it3r:
## Description

Thanks to @PKEuS for pointing out this optimization. 🙂 

## How to test this PR?

I tried to create some dummy code, but I couldn't see any performance improvement or issue, however it is better code, as such it should still be accepted, even if the gain isn't *really* noticeable (on my high-end system). 😉

### DIFF
--- a/src/SFML/Graphics/Texture.cpp
+++ b/src/SFML/Graphics/Texture.cpp
@@ -129,6 +129,11 @@ bool Texture::create(unsigned int width, unsigned int height)
         return false;
     }
 
+    TransientContextLock lock;
+
+    // Make sure that extensions are initialized
+    priv::ensureExtensionsInit();
+
     // Compute the internal texture dimensions depending on NPOT textures support
     Vector2u actualSize(getValidSize(width), getValidSize(height));
 
@@ -150,8 +155,6 @@ bool Texture::create(unsigned int width, unsigned int height)
     m_pixelsFlipped = false;
     m_fboAttachment = false;
 
-    TransientContextLock lock;
-
     // Create the OpenGL texture if it doesn't exist yet
     if (!m_texture)
     {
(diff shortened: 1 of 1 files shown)


LaurentGomila:
What if `getValidSize` is called from somewhere else? I don't like removing stuff from a function and assuming that it will always be done by the callers.

eXpl0it3r:
I've thought of the same thing, but haven't really found a better solution that doesn't require unnecessary locking or code duplication. Maybe pass the lock as reference, but that seems weird as well...

binary1248:
Just inline the implementation of `getValidSize`. It is only used in that one spot. Depending on how hard we want to keep backwards compatibility we can either leave the original function as it is or get rid of it.

LaurentGomila:
> a better solution

To what problem? You said your PR didn't really improve anything, so if we're not 100% satisfied with it, can't we just leave the code as it is?

eXpl0it3r:
> To what problem? You said your PR didn't really improve anything, so if we're not 100% satisfied with it, can't we just leave the code as it is?

None that **I** could measure.

If you started from scratch and designed that code, would you allow for three locks to be created, acquired and released to just do one thing? :wink:

Just because it's not immediately measurable on an high-end system, doesn't mean we should just allow any kind of code quality.

> Depending on how hard we want to keep backwards compatibility

I assume if we remove it, we couldn't include it in 2.5.1 as it would not be compatible as it's defined in a public header (even if it's a private function), correct?

binary1248:
Actually it's not that hard to come up with a test scenario to test this. Obviously it's about unnecessary temporary context locking, so... just try to create textures either before any other context (Window etc.) is created or do it on a secondary thread in which no context has been created yet.

> I assume if we remove it, we couldn't include it in 2.5.1 as it would not be compatible as it's defined in a public header (even if it's a private function), correct?

Yes.

PKEuS:
I can see a clear difference when running my program (which loads ~1000 textures from files) on an Intel UHD 620 on Win 10 with latest driver (so, low-end device). I attached a profiler and observed that 10,57% of entire runtime is spent with constructing and destructing the lock. After the change, this reduces to 7,4% of entire runtime. There are still some other functions called on loading which lock the context, such as setSmooth() or update() (but they do it once per texture, not three times).

eXpl0it3r:
I'm unsure now on how to proceed with this. Personally, I'd just stick with this solution at least for 2.5.1. For 2.6 we could also remove getValidSize and inline the code. Does that sound good to you guys?

PR #1463, opened 2018-08-01
Changed deprecated auto_ptr in AudioDevice to unique_ptr

JinLisek:
Yes, here: https://en.sfml-dev.org/forums/index.php?topic=24313.0
I guess so.
None, I've just changed auto_ptr to unique_ptr. I haven't found any unit tests in the repo. I ran cmake with ninja and it compiled.
None
----

## Description

I removed deprecated auto_ptr in AudioDevice, so it no longer produces warnings when compiling with -Werror. The fix is tiny, shouldn't change anything, also the pointer is not even really used, at least in my understanding.

## Tasks

* [ ] Tested on Linux
It compiles on Linux, I haven't tested it on different platforms
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?
I don't think it's possible to test, as the pointer is not even used.

### DIFF
--- a/src/SFML/Audio/AudioDevice.cpp
+++ b/src/SFML/Audio/AudioDevice.cpp
@@ -107,9 +107,9 @@ bool AudioDevice::isExtensionSupported(const std::string& extension)
     // This device will not be used in this function and merely
     // makes sure there is a valid OpenAL device for extension
     // queries if none has been created yet.
-    std::auto_ptr<AudioDevice> device;
+    std::unique_ptr<AudioDevice> device;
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
> The fix is tiny, shouldn't change anything

Except it would make SFML use C++11, which we still don't support in SFML 2.x, which is the reason that we're still using `std::auto_ptr`.

This change among others has long been implemented in the unstable branch.

Issue #1464, opened 2018-08-09
Window cursor with negative colors


blackhamm3rjack:
## Subject of the issue

Whenever I try to load a window icon over the Windows platform, using a 32-bits PNG image, the resulting colors are inverted. I determined that the color format of the image I use is RGBA meanwhile the correct format for a window cursor is BGRA. I then inverted the colors through a loop that changes every pixel in the image and the problem was fixed.

In the documentation there's written that the required format is RGBA, but I find it wrong. I don't know if it's an issue related to how the Windows API manages cursors or if I'm totally wrong.

## Your environment

* Windows 10 x64 Version 10.0.17134 Build 17134
* SFML from git master
* Microsoft (R) C/C++ Optimizing Compiler Version 19.14.26433 for x64
* No special compiler flags

## Steps to reproduce

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode(1280, 720), "Minimal, complete and verifiable example");
    window.setFramerateLimit(60);

    // Load the icon
    sf::Image image;
    sf::Cursor cursor;
    if (!image.loadFromFile("icon.png"))
        return -1;
    if (!cursor.loadFromPixels(image.getPixelsPtr(), image.getSize(), { 0, 0 }))
        return -1;
    window.setMouseCursor(cursor);

    while (window.isOpen())
    {
        sf::Event event;
        while (window.pollEvent(event))
        {
            if (event.type == sf::Event::Closed)
                window.close();
        }

        window.clear();
        window.display();
    }
}
```

## Expected behavior

The window cursor should be displayed with the correct image colors.

## Actual behavior

The cursor image appears as if the colors were negated.

## Fix

```cpp
// Swap the color format from RGBA to BGRA
for (unsigned x = 0; x < image.getSize().x; ++x)
{
	for (unsigned y = 0; y < image.getSize().y; ++y)
	{
		sf::Color pixel = image.getPixel(x, y);
		image.setPixel(x, y, { pixel.b, pixel.g, pixel.r, pixel.a });
	}
}
```

blackhamm3rjack:
Example cursor: ![arrow](https://user-images.githubusercontent.com/6911442/43892352-f92797c8-9bcb-11e8-929d-68608c8c5a6c.png)

MarioLiebisch:
Branch `bugfix/windows-cursor-colors` should fix this, give it a spin, if you can.

eXpl0it3r:
[...]

PR #1465, opened 2018-08-09
Windows: Fixed swapped colors for custom cursors

MarioLiebisch:
Previously the red and blue color channels were swapped,
since Windows expects a different channel order for DIBs.

This fixes issue #1464.

### DIFF
--- a/src/SFML/Window/Win32/CursorImpl.cpp
+++ b/src/SFML/Window/Win32/CursorImpl.cpp
@@ -89,7 +89,15 @@ bool CursorImpl::loadFromPixels(const Uint8* pixels, Vector2u size, Vector2u hot
     }
 
     // Fill our bitmap with the cursor color data
-    std::memcpy(bitmapData, pixels, size.x * size.y * 4);
+    // We'll have to swap the red and blue channels here
+    Uint8* bitmapOffset = bitmapData;
+    for (std::size_t remaining = size.x * size.y; remaining; --remaining, pixels += 4)
+    {
+        *bitmapOffset++ = pixels[2]; // Blue
+        *bitmapOffset++ = pixels[1]; // Green
+        *bitmapOffset++ = pixels[0]; // Red
+        *bitmapOffset++ = pixels[3]; // Alpha
+    }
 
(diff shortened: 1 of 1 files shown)


LaurentGomila:
Wouldn't it be much simpler to adjust the mask to our input?

```cpp
     bitmapHeader.bV5RedMask     = 0x000000ff;
     bitmapHeader.bV5GreenMask   = 0x0000ff00;
     bitmapHeader.bV5BlueMask    = 0x00ff0000;
     bitmapHeader.bV5AlphaMask   = 0xff000000;
```

MarioLiebisch:
Try it. It was also my first attempt to "fix" it by simply swapping the masks. Once done, the colors are indeed fixed, but also the alpha channel no longer worked for me. Maybe I missed something, but [documentation on this is rather thin right now](https://docs.microsoft.com/en-us/windows/desktop/api/wingdi/ns-wingdi-bitmapv5header), with many things just [described as "TBD"](https://docs.microsoft.com/en-us/windows/desktop/api/wingdi/nf-wingdi-createdibsection).

LaurentGomila:
> Try it. It was also my first attempt to "fix" it by simply swapping the masks. Once done, the colors are indeed fixed, but also the alpha channel no longer worked for me. 

Really strange. I'll try to investigate more if I have some time.

> As for endianess I'm pretty sure the WinAPI accounts for that.

The problem is that the documentation (of value `32` for `bV5BitCount`) explicitly mentions `DWORD` values. So we should really write composed colors as 32-bit integers in the pixel array, not 8-bit components separately. Which leads to a code similar to your PR anyway, so let's forget about the masks.

LaurentGomila:
The way we populate the pixel array is still not endian safe.

eXpl0it3r:
So what's the solution?

LaurentGomila:
Something like this:

```cpp
Uint32* bitmapOffset = bitmapData;
for (std::size_t remaining = size.x * size.y; remaining; --remaining, pixels += 4)
{
    *bitmapOffset++ = (pixels[3] << 24) | (pixels[0] << 16) | (pixels[1] << 8) | pixels[2];
}
```

Issue #1456, opened 2018-07-08
Pixel font size 


codecat:
## Subject of the issue

I'm trying to load a pixel font (specifically, artwiz-aqui) and setting the character size to 12, but I get the following error:

```
Failed to set bitmap font size to 12
Available sizes are: 12
```

I think this is because of this line:

https://github.com/SFML/SFML/blob/949e7aecae1101348308433ab3758f85ca5e2b9e/src/SFML/Graphics/Font.cpp#L768

According to [this page](https://www.freetype.org/freetype2/docs/tutorial/step1.html), "a value of 0 for one of the dimensions means ‘same as the other’". So, would it be possible that it's not the height that's not available, but the width?

## Your environment

* Windows 7 Ultimate x64
* SFML 2.5.0
* Visual Studio 15.7.3

## Steps to reproduce

Get `aqui.pcf` from here: https://sourceforge.net/projects/artwizaleczapka/

```cpp
#include <SFML/Graphics.hpp>

int main()
{
	sf::RenderWindow window(sf::VideoMode(1280, 720), "Minimal, complete and verifiable example");
	window.setFramerateLimit(60);

	sf::Font font;
	font.loadFromFile("aqui.pcf");

	sf::Text text;
	text.setFont(font);
	text.setCharacterSize(12);
	text.setString("Hello, world");
	text.setFillColor(sf::Color::White);

	while (window.isOpen())
	{
		sf::Event event;
		while (window.pollEvent(event))
		{
			if (event.type == sf::Event::Closed)
				window.close();
		}

		window.clear();
		window.draw(text);
		window.display();
	}
}
```

## Expected behavior

The font should be 12px high.

## Actual behavior

Strange error.

Qu3tzal:
Hello,

I looked into your problem and found out that your font is only available in 9x12 size (width x height).

LaurentGomila:
Or we could search into `face->fixed_sizes` for the matching width and use it instead of 0, in case of a bitmap font.

MarioLiebisch:
Sounds like a good idea, just pick the size available that's closest to the one requested.

Qu3tzal:
Something among the lines of 19cf28319b97a7b4941fe1ec8f63028f0a152f4f ?

LaurentGomila:
Yes, that's the idea.

eXpl0it3r:
@Qu3tzal Are you going to create a PR?

eXpl0it3r:
[...]

PR #1474, opened 2018-08-15
Fallback for fonts were the height is not equal to the width

Qu3tzal:
This PR is related to the issue #1456. In this PR we propose a way to fallback when a font does not have the same height and width. We look for the first available width size for the given height.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Get `aqui.pcf` from here: https://sourceforge.net/projects/artwizaleczapka/

```c++
#include <SFML/Graphics.hpp>

int main()
{
	window.setFramerateLimit(60);

	sf::Font font;
	font.loadFromFile("aqui.pcf");

	sf::Text text;
	text.setFont(font);
	text.setCharacterSize(12);
	text.setString("Hello, world");

	while (window.isOpen())
	{
		sf::Event event;
		{
				window.close();
		}

		window.clear();
		window.draw(text);
		window.display();
	}
}
```

### DIFF
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -777,7 +777,7 @@ bool Font::setCurrentSize(unsigned int characterSize) const
                 err() << "Failed to set bitmap font size to " << characterSize << std::endl;
                 err() << "Available sizes are: ";
                 for (int i = 0; i < face->num_fixed_sizes; ++i)
-                    err() << face->available_sizes[i].height << " ";
+                    err() << face->available_sizes[i].width << "x" << face->available_sizes[i].height << " ";
                 err() << std::endl;
             }
         }
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -782,16 +782,21 @@ bool Font::setCurrentSize(unsigned int characterSize) const
                         result = FT_Set_Pixel_Sizes(face, face->available_sizes[i].width, characterSize);
                     }
                 }
-            }
 
-            // Check if we found a way to fallback to a valid size. If we couldn't, log the error.
-            if (result == FT_Err_Invalid_Pixel_Size)
+				// Check if we found a way to fallback to a valid size. If we couldn't, log the error
+                if (result == FT_Err_Invalid_Pixel_Size)
+                {
+                    err() << "Failed to set bitmap font size to " << characterSize << std::endl;
+                    err() << "Available sizes are: ";
+                    for (int i = 0; i < face->num_fixed_sizes; ++i)
+                        err() << face->available_sizes[i].width << "x" << face->available_sizes[i].height << " ";
+                    err() << std::endl;
+                }
+			}
+            // If the font is scalable but we had an error
+			else if (result == FT_Err_Invalid_Pixel_Size) 
             {
-                err() << "Failed to set bitmap font size to " << characterSize << std::endl;
-                err() << "Available sizes are: ";
-                for (int i = 0; i < face->num_fixed_sizes; ++i)
-                    err() << face->available_sizes[i].width << "x" << face->available_sizes[i].height << " ";
-                err() << std::endl;
+                err() << "Failed to set font size to " << characterSize << std::endl;
             }
         }
 
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -775,13 +775,11 @@ bool Font::setCurrentSize(unsigned int characterSize) const
             if (!FT_IS_SCALABLE(face))
             {
                 // Look for an available size with a matching height
-                bool sizeFound = false;
-                for (int i = 0; i < face->num_fixed_sizes && !sizeFound; ++i)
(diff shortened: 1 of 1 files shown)


src/SFML/Graphics/Font.cpp
```diff
@@ -777,8 +777,24 @@ bool Font::setCurrentSize(unsigned int characterSize) const
...
+                    err() << face->available_sizes[i].width << "x" << face->available_sizes[i].height << " ";
                 err() << std::endl;
+				
+				// Look for an available size with a matching height
+				bool sizeFound = false;
+				for (int i = 0; i < face->num_fixed_sizes && !sizeFound; ++i)
+				{
+					if (face->available_sizes[i].height == characterSize)
+					{
+						result = FT_Set_Pixel_Sizes(face, , characterSize);
```
eXpl0it3r: Woops, what happened here? Two commas after each other?

src/SFML/Graphics/Font.cpp
```diff
@@ -777,8 +777,24 @@ bool Font::setCurrentSize(unsigned int characterSize) const
...
+						sizeFound = true;
+					}
+				}
+				
+				if (result == FT_Err_Invalid_Pixel_Size || !sizeFound)
+				{
+					err() << "Failed to fallback on an available size." << std::endl;
+				}
```
eXpl0it3r: You should use spaces instead of tabs for indenting. 🙂

src/SFML/Graphics/Font.cpp
```diff
@@ -790,11 +784,16 @@ bool Font::setCurrentSize(unsigned int characterSize) const
...
-                    err() << "Failed to fallback on an available size." << std::endl;
-                }
+            // Check if we found a way to fallback to a valid size. If we couldn't, log the error.
+            if (result == FT_Err_Invalid_Pixel_Size)
+            {
+                err() << "Failed to set bitmap font size to " << characterSize << std::endl;
+                err() << "Available sizes are: ";
+                for (int i = 0; i < face->num_fixed_sizes; ++i)
+                    err() << face->available_sizes[i].width << "x" << face->available_sizes[i].height << " ";
+                err() << std::endl;
```
LaurentGomila: In case of a scalable font, we can't say it is a "bitmap" font and there will probably be no fixed sizes to print. So we can't use the same error message for both cases (or if we do, make it less specific).

src/SFML/Graphics/Font.cpp
```diff
@@ -775,13 +775,11 @@ bool Font::setCurrentSize(unsigned int characterSize) const
             if (!FT_IS_SCALABLE(face))
             {
                 // Look for an available size with a matching height
-                bool sizeFound = false;
-                for (int i = 0; i < face->num_fixed_sizes && !sizeFound; ++i)
+                for (int i = 0; i < face->num_fixed_sizes && result != FT_Err_Invalid_Pixel_Size; ++i)
```
LaurentGomila: What if the function returns some error other than `FT_Err_Invalid_Pixel_Size`?

Qu3tzal:
Anything else to change ? :)

texus:
Replacing `result = FT_Set_Pixel_Sizes(face, face->available_sizes[i].width, characterSize);` by `result = FT_Select_Size(face, i);` seems to work at first sight.

texus:
I debugged the freetype code to figure out why it returned FT_Err_Invalid_Pixel_Size when using pixel size 12. The PCF_Size_Request function in the pcf driver doesn't look at `bzize->height` (which is 12, bsize is basically face->available_sizes[i]), but checks `(bsize->y_ppem + 32 ) >> 6)`, which is 11.

Calling `text.setCharacterSize(11);` in the example code works.

Using `FT_Select_Size` instead of `FT_Set_Pixel_Sizes` seems to somehow work for both pixel sizes 11 and 12, although the output on the screen is identical.

eXpl0it3r:
Yeah, I tried to debug FreeType, but failed to build debug libs for VS 2017. Thanks for getting behind this.

Weird that it calculates to 11 and that it works with `FT_Select_Size` for both 11 and 12.

@Qu3tzal Could you adjust the PR to what @texus suggested?

texus:
I figured out why it works for both font sizes 11 and 12 when using FT_Select_Size, the code isn't even being executed when using 11.

When font size 11 is used, the `FT_Set_Pixel_Sizes(face, 0, characterSize);` on top of the function already returns `FT_Err_Ok`, so the changed code is never reached.

The `FT_Select_Size` would work for font size 12 because it uses the index in face->available_sizes instead of a font height as parameter. Which index is used depends on `if (face->available_sizes[i].height == characterSize)`, which is wrongly true for the value 12.

So the code introduced in this PR (except for maybe the else case of `if (!FT_IS_SCALABLE(face))`) isn't actually needed. The original code can just be used with `err() << face->available_sizes[i].height << " ";` changed to
```c++
const unsigned int size = (face->available_sizes[i].y_ppem + 32) >> 6;
err() << size << " ";
```

eXpl0it3r:
I applied the changes in the PR #1492, as such I'll close this PR.

PR #1475, opened 2018-08-15
Add unit testing foundation to SFML

eXpl0it3r:
## Description

This PR lays the foundation with some very basic test cases. The goal here isn't to get 100% coverage, but to start somewhere and build unit tests over time. So over time, new features will generate new unit tests and unit tests will help shape features.

See the Unit tests discussion and the [SFML Test Strategy](https://en.sfml-dev.org/forums/index.php?topic=24355.0) thread on the forum.

## Tasks

* [x] Find a solution for dynamic library not being found for the runtests target
* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

## How to test this PR?

Test the tests! As in build the test step and run the tests. Would be especially interesting to hear how it runs on macOS and Linux.


CMakeLists.txt
```diff
@@ -84,6 +84,9 @@ if(SFML_BUILD_WINDOW)
     sfml_set_option(SFML_OPENGL_ES ${OPENGL_ES} BOOL "TRUE to use an OpenGL ES implementation, FALSE to use a desktop OpenGL implementation")
 endif()
 
+# add an option for building the test suite
+sfml_set_option(SFML_BUILD_TEST TRUE BOOL "TRUE to build the SFML test suite, FALSE to ignore them")
```
LaurentGomila: > SFML_BUILD_TEST

`SFML_BUILD_TESTS` or `SFML_BUILD_TEST_SUITE` would be more correct.

> FALSE to ignore them

to ignore _it_ 😉

test/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+    "${SRCROOT}/Rect.cpp"
+)
+
+include_directories("${PROJECT_SOURCE_DIR}/include")
+include_directories("${PROJECT_SOURCE_DIR}/extlibs/headers")
+
+# Using sfmltest instead of test because the latter is reserved.
+add_executable(sfmltest ${SRC})
+
+target_link_libraries(sfmltest sfml-graphics sfml-window sfml-audio sfml-network sfml-system)
```
JonnyPtn: Nothing major, but not all sfml libs need to be listed here as dependencies will automatically be linked. Could probably just list `sfml-graphics sfml-audio sfml-network`. Depends if you prefer comcise or explicit

LaurentGomila: We always prefer explicit. Would be strange to test sfml-system and not see it in the list of linked libraries.

But this raises a more important point in my opinion: why don't we have one test executable per module? SFML modules are more or less independant, and we can select which ones are built, so we should have the same modularity for tests.

JonnyPtn: This makes a lot of sense, especially as there are cmake options to enable/disable certain modules

JonnyPtn:
Strangely, on macOS I get:

```
Undefined symbols for architecture x86_64:
  "sf::String::toAnsiString(std::locale const&) const", referenced from:
      Catch::toString(sf::String const&) in UnitTests.o
ld: symbol(s) not found for architecture x86_64
```

Given that the function isn't used I can just comment it out, and everything else works fine, but I can't see why it would be undefined, and nothing else. Would be interested to see what happens on linux

test/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+
+add_custom_target(runtests ALL
```
JonnyPtn: Using `add_test` here would be the more typical CMake way. e.g.:

```
add_test(runtests sfmltest
WORKING_DIRECTORY "${PROJECT_BINARY_DIR}/test")
```

These will automatically get added to the "test" target, so you can easily add future test programs

test/CMakeLists.txt
eliasdaler: Agreed. And if you disable some modules from building, you get a linking errors

test/CMakeLists.txt
eliasdaler: Yes, and maybe it'll get us some ctest integration?
Can't we use [this](https://github.com/catchorg/Catch2/blob/3d7104c124e3af88eda573f867f79d0aa9ef6c72/contrib/Catch.cmake#L10) for the version we're currently on?

If not, it should be on the todo list as it's pretty convenient. :)

test/src/UnitTests.hpp
```diff
@@ -0,0 +1,58 @@
...
+// This ensures that string conversions are visible and can be used by Catch for debug output.
+
+#ifndef SFML_UNITTESTS_HPP
+#define SFML_UNITTESTS_HPP
+
+#include <SFML/System/Vector2.hpp>
+#include <SFML/System/Vector3.hpp>
+#include <SFML/Graphics/Rect.hpp>
+#include <sstream>
+#include <catch.hpp>
```
eliasdaler: I'd suggest to move <catch.hpp> higher (maybe so that it's the first included header?), because the list of SFML headers included here may become bigger in the future.

test/src/UnitTests.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <SFML/Window/VideoMode.hpp>
+#include <SFML/Graphics/Color.hpp>
+#include <sstream>
+
+// String conversions for Catch framework
+namespace Catch
+{
+    std::string toString(const sf::String& string)
+    {
+        return string.toAnsiString();
```
eliasdaler: Why not convert to Utf8? This is probably what will work on most platforms with non-ASCII strings
Now sure how to it handles Windows, though...

test/src/UnitTests.cpp
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    std::string toString(sf::Time time)
+    {
+        std::ostringstream stream;
+        stream << time.asMicroseconds() << "us";
```
eliasdaler: Won't we need milliseconds more frequently? Microseconds are too noisy for human readable output.

eXpl0it3r: The point isn't necessarily human readable output, but predictable and comparable output. Microseconds is the smallest unit and what's saved internally, so this makes the most sense.

test/src/UnitTests.cpp
eXpl0it3r: Because catch expects a `std::string` as return. Not sure if it can deal with `std::basic_string<Uint8>` in all cases. In theory it should work, as `sf::Uint8` is defined as `char` and `std::string` is defined as `std::basic_string<char>`.

test/src/UnitTests.hpp
eXpl0it3r: I moved it up.

test/src/UnitTests.cpp
eliasdaler: We can make a string out of sf::basic_string<Uint8> explicitly. That's what a lot of programs do now - they store UTF-8 in std::string. :)

test/CMakeLists.txt
eXpl0it3r: Source files for a specific module are now only added when the module is explicitly selected to be built and only then the lib is set. Required to split the test utilities into module-based parts.

test/src/UnitTests.cpp
eXpl0it3r: Any example code?

test/CMakeLists.txt
eXpl0it3r: The linked command doesn't seem to exist or catch 1.x.

How do you run the when using `add_test`? The current solution is quite neat, as it runs the test suite when you build said target. If the same can be done with `add_test` then let me know when you figure it out, as I can't seem to get it working.

JonnyPtn: That link's a cmake script so although it may not be in catch 1.x we could still use it in catch2 as long as it doesn't depend on any catch2 functionality.
add_test is what populates the "test" target (the one that this PR tries to avoid clashing with by naming the target "runtests". This means you can easily run the tests either by building the test target (e.g. `make test`) or just by executing `ctest` in the build directory after configuring with cmake.

there are other advantages to using the built in cmake testing stuff, for example you can add as many test programs as you want using `add_test` and they all get added to the "test" target. You can also use the built in flags for tests: `BUILD_TESTING`.

These two solutions aren't mutually exclusive either, we could easily run the proper test stuff alongside the custom command stuff if that covers all bases. 

I'll try and put together an example of this behaviour later (And also look into that catch command which supposedly improves on add_test)

test/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+    "${SRCROOT}/Vector2.cpp"
+    "${SRCROOT}/Vector3.cpp"
+    "${SRCROOT}/TestUtilities/System.hpp"
+    "${SRCROOT}/TestUtilities/System.cpp"
+)
+SET(SYSTEM_LIB "sfml-system")
+
+if(SFML_BUILD_WINDOW)
+    SET(WINDOW_SRC
+        "${SRCROOT}/Rect.cpp"
```
JonnyPtn: Shouldn't Rect be in the graphics tests?

eXpl0it3r: Woops, forgot to move it after I created the graphics section. 🙂

CMakeLists.txt
```diff
@@ -84,6 +84,9 @@ if(SFML_BUILD_WINDOW)
     sfml_set_option(SFML_OPENGL_ES ${OPENGL_ES} BOOL "TRUE to use an OpenGL ES implementation, FALSE to use a desktop OpenGL implementation")
 endif()
 
+# add an option for building the test suite
+sfml_set_option(SFML_BUILD_TEST_SUITE TRUE BOOL "TRUE to build the SFML test suite, FALSE to ignore it")
```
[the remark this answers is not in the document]
eliasdaler: It's OFF by default in most libs, I think. So we'd better do the same ;)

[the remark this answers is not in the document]
TankOs: It will also indicate errors/problems when the test suite fails for someone. Unit tests are there to be run, not to be ignored. IMO it should stay enabled by default.

If we want to make unit tests part of our workflow, then we will have to explain to contributors how to enable tests, so that their PRs get accepted. Another annoying step to do that can be avoided.

The extra build time is only required for the initial build, especially when building the test suite "main file". Consequent builds are much faster.

TankOs:
Successfully tested on Arch Linux. Btw, if tests were enabled in Buildbot, most platforms would have been tested already automatically.

```
Run tests
Test project /home/stsch/projects/sfml/build
    Start 1: System-module-tests
1/3 Test #1: System-module-tests ..............   Passed    0.00 sec
    Start 2: Window-module-tests
2/3 Test #2: Window-module-tests ..............   Passed    0.01 sec
    Start 3: Graphics-module-tests
3/3 Test #3: Graphics-module-tests ............   Passed    0.01 sec

100% tests passed, 0 tests failed out of 3

Total Test time (real) =   0.02 sec
```

eXpl0it3r:
Except the wouldn't as buildbot would fail, because it can't find the shared libraries. Check the status of previous commits. 😉

CMakeLists.txt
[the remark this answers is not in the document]
JonnyPtn: It's not just the build time, but also the run time, which could take any amount of time in the future once more (and more complex) tests are added. 

furthermore, it causes the build itself to fail meaning if a unit test fails for any reason the user won't be able to install SFML without fixing the bug or removing the test. 

I think it would be more reasonable to ask people to run tests before reporting an issue or sending a PR rather than blocking use of SFML until the issue is found and resolved

TankOs:
@eXpl0it3r I remember the talk about RPATH etc, and whether Windows and OSX have a similar feature.

JonnyPtn:
You can set the working directory for [add_test](https://cmake.org/cmake/help/v3.0/command/add_test.html), if that helps

JonnyPtn:
Works fine for me still, no local changes

JonnyPtn:
You can tick off macOS testing now. Up to you if you want to tick off iOS or not. It’s basically disabled on iOS currently until someone works out a way to run iOS unit tests (and I presume android will have the same problem?)

You can also tick off a solution for dlls, as that’s covered now, although Laurent suggested an alternative solution of building all tests and examples in the same folder as the libs. Personally I would prefer not to mix up the libs and the other guff, and don’t want to  change behaviour on all platforms when only Windows is a problem, which is why I did it the way I did, but I’m not particularly precious so feel free to change it if it’s decided that the other way is better. 

I’d say it’s best not to worry too much about iOS and Android tests right now as I’ve no idea how long that might take to set up (and also they aren’t “officially” supported yet) so let’s just make a decision on the dll situation and get this moving

LaurentGomila:
> Laurent suggested an alternative solution of building all tests and examples in the same folder as the libs. Personally I would prefer not to mix up the libs and the other guff, and don’t want to change behaviour on all platforms when only Windows is a problem

No, I said to tell CMake to put all binaries into a "bin" folder, as well as libraries in "lib". This would result in:
* .dll and .exe in "bin", .a / .lib in "lib" on Windows
* executables in "bin" , ".so" in "lib" on Linux

So it's not platform specific, CMake already differentiates correctly files types on all platforms. And I think it's quite common to do it this way (we do it for all our projects at work).

JonnyPtn:
Sorry for the misunderstanding - my preference still stands for the Same reasons, but like I said I’m not precious about it

LaurentGomila:
> my preference still stands for the Same reasons

So, what reasons? As I said, there's no mix between libraries and runtime files, and it doesn't change anything in behaviour compared to the current version.

LaurentGomila:
> Mixing the "SFML Library" files with examples/tests

Depends what you call "libraries". CMake separates "runtime" files (.dll on Windows, none on Linux) from "library" files (.a/.lib import files on Windows, .so on Linux) from "archive" files (.a/.lib static libraries on Windows, .a static libraries on Linux) -- I'm not talking about MacOS since my knownledge of this OS is too limited.

But in the end, you have all "run-time" files in "bin", and all "link-time" files (libraries) in "lib". That's pretty close to what you get in final packages anyway.

> changing the process for all platforms when only windows is the issue

We don't change anything for other OSes. We just set global rules that:
* change something on Windows, where there was an issue,
* change nothing on Linux, where things were already ok
* again, I leave MacOS aside but I guess that similar conclusions can be drawn

So, can you be a litte more specific about what exactly bothers you with this solution?

JonnyPtn:
Not really.

What’s left to get this merged then?

eXpl0it3r:
Looks like we still have some issues with DLLs with VS builds and I guess we should disable it for Android builds or figure out what syntax issue they have. 😄

LaurentGomila:
> Looks like we still have some issues with DLLs with VS builds

What issues? Now that the DLLs are copied the tests run fine.

eXpl0it3r:
You can check the Travis build log. The error code indicates that it's missing a DLL, but we don't really know which one. To be honest it reads like a Travis issue.

I tried to figure out how to run the Travis VS system on my local machine, but couldn't find any documentation on this.

I'll look into it again some time this weekend.

JonnyPtn:
Is buildbot set up to run the tests too?

LaurentGomila:
> To be honest it reads like a Travis issue.

Probably. Today I tried a fresh checkout of the this branch, with VS2015 / dynamic, and all built and ran fine.

test/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+
+# System is always built
+SET(SYSTEM_SRC
+    "${SRCROOT}/CatchMain.cpp"
+    "${SRCROOT}/System/Vector2.cpp"
+    "${SRCROOT}/System/Vector3.cpp"
+    "${SRCROOT}/TestUtilities/SystemUtil.hpp"
+    "${SRCROOT}/TestUtilities/SystemUtil.cpp"
+)
+sfml_add_test(systemtest "${SYSTEM_SRC}"  sfml-system)
```
LaurentGomila: I would name the targets differently, this looks different from the conventions used to name modules. For example, sfml-xxx-test (or test-sfml-xxx) would be consistent.

test/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
+set(SRCROOT "${PROJECT_SOURCE_DIR}/test/src")
+
+include_directories("${PROJECT_SOURCE_DIR}/include")
+include_directories("${PROJECT_SOURCE_DIR}/extlibs/headers")
+include_directories("${SRCROOT}/TestUtilities")
+
```
LaurentGomila: Why not putting each test target in its own CMakeLists.txt file, in the corresponding sub-folder?

test/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+if(SFML_BUILD_GRAPHICS)
+    SET(GRAPHICS_SRC
+        "${SRCROOT}/CatchMain.cpp"
+        "${SRCROOT}/Graphics/Rect.cpp"
+        "${SRCROOT}/TestUtilities/GraphicsUtil.hpp"
+        "${SRCROOT}/TestUtilities/GraphicsUtil.cpp"
+    )
+    sfml_add_test(graphicstest "${GRAPHICS_SRC}" sfml-graphics)
+endif()
+
```
LaurentGomila: I would also add a test target for sfml-network and sfml-audio, for completeness. And so that people who want to contribute don't have to create them.

test/src/CatchMain.cpp
```diff
@@ -0,0 +1,2 @@
+#define CATCH_CONFIG_MAIN
+#include <catch.hpp>
```
LaurentGomila: "No newline at end of file" 😉

cmake/Macros.cmake
```diff
@@ -265,6 +265,42 @@ macro(sfml_add_example target)
...
+# example: sfml_add_test(sfml-test
+#                           ftp.cpp ...
+#                           sfml-network)
+function(sfml_add_test target SOURCES DEPENDS)
+
+    # set a source group for the source files
+    source_group("" FILES ${SOURCES})
+
+    # check whether resources must be added in target
+    set(target_input ${SOURCES})
```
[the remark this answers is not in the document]
eXpl0it3r: Me neither. 🤔 
It just sets `target_input` and then uses that further down instead of the `${SOURCES}` but it doesn't get used any further. Maybe some leftover, @JonnyPtn?

test/CMakeLists.txt
eXpl0it3r: It's not hard to add and we don't want to add stuff just because it hopefully will be added one day, same as we don't add C++ functions because they might be useful someday. 😉

test/CMakeLists.txt
eXpl0it3r: Might be part of a future enhancement, for now I think we'll just stick to this.

test/CMakeLists.txt
eXpl0it3r: Renamed to test-sfml-xyz

cmake/Macros.cmake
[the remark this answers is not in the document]
eXpl0it3r: Oh, I just checked the history and noticed by chance that this is just an copy-paste artifact. The example macro adds resource files to that target if they are needed. Might be useful at one point, but not right now.

eXpl0it3r:
Looks like it's still a Travis issue. There are two steps I want to get done before merging this:

- Disable the tests on the failing Travis instance
- Activate tests on our CI system

test/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+# Automatically run the tests at the end of the build
+add_custom_target(runtests ALL
+                  DEPENDS test-sfml-system test-sfml-window test-sfml-graphics
+)
+
+add_custom_command(TARGET runtests
+                   COMMENT "Run tests"
+                   POST_BUILD COMMAND ctest ARGS --output-on-failure
+)
```
ChrisThrasher, line 44: What was the rationale for automatically running tests after they're built? This is not a conventional thing to do and not something I've seen on other C++ projects. Now that SFML has 31 test cases, the runtime impact of these tests is starting to approach the time it takes to incrementally build a few C++ files. This problem is only prone to get worse over time as we improve test coverage and the test suite takes longer and longer to run. Our CI pipeline is so rigorous that if a contributor forgets to build or run the tests, that problem is immediately caught upon submission of their PR so I'm not particularly worried about such situations.

ChrisThrasher: https://github.com/SFML/SFML/issues/2429

I filed an issue to remove this.

Issue #1472, opened 2018-08-15
[Linux] Building SFML fails with conflicting declarations when using a recent git mesa


ShadowsFriend:
When I try to build the current master version of SFML it fails with several conflicting declarations with the previous declarations all being in glxext.h. On my system, this file is provided by a current git version of mesa. I have attached a logfile containing the output of cmake and make including the errors.

## Your environment

* Gentoo Linux x86_64
* git master (83d05ea8228790d2a44ba8393393fb7a57ee498b), also tested with 2.4.2 and 2.5.0
* mesa git (70bf639328fa00d7875e82b43fb011b0687559c0)
* GCC 7.3.0

## Steps to reproduce
1. Clone the SFML git repository
2. Enter the directory where the repository has been cloned to
3. cmake .
4. make

## Expected behavior

SFML builds without an error.

## Actual behavior

SFML fails building with conflicting declarations.

[build.log](https://github.com/SFML/SFML/files/2288863/build.log)

Edit: The reason for this issue is src/SFML/Window/Unix/GlxExtensions.hpp including the system's GL/glx.h which in turn includes GL/glxext.h. And this file then provides declarations that are also in src/SFML/Window/Unix/GlxExtensions.hpp.

ShadowsFriend:
It seems the header changes introduced in mesa commit [1] break the compilation. I was able to successfully compile SFML when using mesa 18.2.0-rc2 (which does not include this commit) but when I manually apply it on top of 18.2.0-rc2, the compilation of SFML breaks.

[1] https://cgit.freedesktop.org/mesa/mesa/commit/?id=f7d42ee7d319256608ad60778f6787c140badada

oprypin:
For the record, this issue makes SFML releases before 2.5 un-buildable on any recent Linux system.
For discoverability, the error message is:

Clang:
```
In file included from src/SFML/Window/GlContext.cpp:47:
In file included from src/SFML/Window/Unix/GlxContext.hpp:32:
src/SFML/Window/Unix/GlxExtensions.hpp:123:3: error: typedef redefinition with different types ('struct GLXHyperpipeNetworkSGIX' vs 'struct GLXHyperpipeNetworkSGIX')
} GLXHyperpipeNetworkSGIX;
  ^
/usr/include/GL/glxext.h:771:3: note: previous definition is here
} GLXHyperpipeNetworkSGIX;
  ^
In file included from src/SFML/Window/GlContext.cpp:47:
[...]

PR #1476, opened 2018-08-16
Window/Unix: Fix compilation with glxext header versions >=20180525

ShadowsFriend:

----

## Description

When mesa updated their headers, they changed the include guard
from __glxext_h_ to __glx_glxext_h_, which breaks compilation
due to conflicting declarations. This commit modifies the preprocessor
directives to allow for compilation with older and newer mesa header
versions.
Fixes: #1472

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?
Just trying to compile should be enough to test the PR on Linux with mesa should be enough. I tested it with the headers from mesa 18.1.6 and with headers from a current git mesa (9a96bf0ecd071219cb975fbd64f5c68849fd5697).

### DIFF
--- a/src/SFML/Window/Unix/GlxExtensions.hpp
+++ b/src/SFML/Window/Unix/GlxExtensions.hpp
@@ -25,11 +25,12 @@
 #ifndef SF_POINTER_C_GENERATED_HEADER_GLXWIN_HPP
 #define SF_POINTER_C_GENERATED_HEADER_GLXWIN_HPP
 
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
Merged in 87aaa9e145659d6a8fc193ab8540cf847d4d0def

PR #1477, opened 2018-08-18
Fix RPATH for installed examples on Linux

eliasdaler:
## Steps to reproduce

* Configure CMake build with custom CMAKE_INSTALL_PREFIX
* `cmake --build . target --install`
* Go to any example folder and try to launch it - fails with "sfml-xxx.so is not found"
* Check executable via "ldd" - shows that .so modules are not found

```
    libsfml-graphics.so.2.5 => not found
    libsfml-window.so.2.5 => not found
```

## How this fix works

First, it computes relative path from target directory to lib folder. Then it sets it target's INSTALL_RPATH to $ORIGIN + rel_path ($ORIGIN being a location of executable).

And now, when you install the examples, everything will work fine and SFML modules will be found by the relative path. Even if you move the install folder everywhere on the system as long as its internal structure remains the same.

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -232,7 +232,7 @@ macro(sfml_add_example target)
 
     set(target_install_dir ${SFML_MISC_INSTALL_PREFIX}/examples/${target})
 
-    if(SFML_OS_LINUX OR SFML_OS_FREEBSD)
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
Merged in b81de898bf49d919354899d14ecf2b81753f9f42

PR #1478, opened 2018-08-19
Use modern cmake testing functionality including catch integration

JonnyPtn:
This uses cmake's built in testing functionality (ctest). This means the tests are added to the "test" target, and you can continue to add as many executables to that target in future as you like. 

To run tests you can now either build the "test" target, e.g. `make test`, run ctest from the build folder e.g. `ctest`, or run the executable manually yourself. I've also kept the existing behaviour to run the tests after the build

On top of that, I've added and used the [cmake integration script provided by the Catch repo](https://github.com/catchorg/Catch2/blob/master/contrib/Catch.cmake) which names the Ctest tests based on the catch test names rather than having to specify them in the cmake script, as well as other functionality which may be of use in the future.


eXpl0it3r:
Looks good. Maybe we could have a file per module. Maybe even move each module into their own sub-directory and move & rename the test utility files as well. I'm fine with either way.

JonnyPtn:
I've restructured the folders a little bit to separate out the test source for each module, and also moved the catch main code into "CatchMain.cpp" which is built with each unit test now

PR #1480, opened 2018-08-23
Add support for creating Color from HSVA

IvanFon:
Adds a static sf::Color::fromHSVA method to create a colour using HSVA values.

This PR is related to [this forum post](https://en.sfml-dev.org/forums/index.php?topic=7313.0).

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode(600, 200), "HSVA Example");

    sf::CircleShape red(100.f);
    red.setFillColor(sf::Color::fromHSVA(0, 1, 1));

    sf::CircleShape yellow(100.f);
    yellow.setFillColor(sf::Color::fromHSVA(50, 1, 1));
    yellow.setPosition(200, 0);

    sf::CircleShape green(100.f);
    green.setFillColor(sf::Color::fromHSVA(100, 1, 1));
    green.setPosition(400, 0);

    window.clear();
    window.draw(red);
    window.draw(yellow);
    window.draw(green);
    window.display();

    sf::sleep(sf::seconds(3));

}
```

### DIFF
--- a/src/SFML/Graphics/Color.cpp
+++ b/src/SFML/Graphics/Color.cpp
@@ -88,8 +88,6 @@ Uint32 Color::toInteger() const
 Color Color::fromHSVA(Uint16 hue, double sat, double val, Uint8 alpha)
 {
     hue %= 360;
-    while (hue < 0)
-        hue += 360;
 
     if (sat < 0)
         sat = 0;
--- a/src/SFML/Graphics/Color.cpp
+++ b/src/SFML/Graphics/Color.cpp
@@ -85,7 +85,7 @@ Uint32 Color::toInteger() const
 }
 
 ////////////////////////////////////////////////////////////
-Color Color::fromHSVA(Uint16 hue, double sat, double val, Uint8 alpha)
+Color Color::fromHSVA(Uint16 hue, float sat, float val, Uint8 alpha)
 {
     hue %= 360;
 
@@ -100,10 +100,10 @@ Color Color::fromHSVA(Uint16 hue, double sat, double val, Uint8 alpha)
         val = 1;
 
     int h = hue / 60;
-    double f = float(hue) / 60 - h;
-    double p = val * (1 - sat);
-    double q = val * (1 - sat * f);
-    double t = val * (1 - (1 - f) * sat);
+    float f = float(hue) / 60 - h;
+    float p = val * (1 - sat);
+    float q = val * (1 - sat * f);
+    float t = val * (1 - (1 - f) * sat);
 
     switch (h)
     {
--- a/src/SFML/Graphics/Color.cpp
+++ b/src/SFML/Graphics/Color.cpp
@@ -84,6 +84,41 @@ Uint32 Color::toInteger() const
     return (r << 24) | (g << 16) | (b << 8) | a;
 }
 
+////////////////////////////////////////////////////////////
+Color Color::fromHSVA(Uint16 hue, double sat, double val, Uint8 alpha)
+{
+    hue %= 360;
+    while (hue < 0)
+        hue += 360;
+
+    if (sat < 0)
+        sat = 0;
+    else if (sat > 1)
+        sat = 1;
+
+    if (val < 0)
+        val = 0;
+    else if (val > 1)
+        val = 1;
+
+    int h = hue / 60;
+    double f = float(hue) / 60 - h;
+    double p = val * (1 - sat);
+    double q = val * (1 - sat * f);
+    double t = val * (1 - (1 - f) * sat);
+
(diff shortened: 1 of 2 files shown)


src/SFML/Graphics/Color.cpp
```diff
@@ -84,6 +84,41 @@ Uint32 Color::toInteger() const
     return (r << 24) | (g << 16) | (b << 8) | a;
 }
 
+////////////////////////////////////////////////////////////
+Color Color::fromHSVA(Uint16 hue, double sat, double val, Uint8 alpha)
+{
+    hue %= 360;
+    while (hue < 0)
```
dmitry-t: How can uint16 become negative? :)

IvanFon:
@fahlmant I've tested on Arch Linux with i3wm and Debian with XFCE and LXDE.

dmitry-t:
A question. SFML uses `float`s [almost] [everywhere](https://github.com/SFML/SFML/search?q=double&unscoped_q=double) (single exclusion - in implementation, OSX/JoystickImpl.cpp). Here there are `double`s. This looks inconsistent. Is the double precision really necessary?

eXpl0it3r:
Would have been better to revive the linked old post on the forum first. SFML uses RGBA across the board. Given that and the fact that this function can easily exist as an independent utility function, adding it to SFML doesn't seem like such a good idea. Also brings up a lot more questions of: Why just toHSVA? Why not also fromHSVA? Why HSVA and not some other color models? etc.

My personal opinion is to not add this functionality to SFML.

IvanFon:
@eXpl0it3r That's a good point, and this could be implemented independently, but I think a user in the linked forum thread made a good point that SFML is a multimedia library, so supporting other colour spaces probably wouldn't be out of place.

I chose HSV because it's a very useful colour space for working programmatically with colours. For [example](https://stackoverflow.com/questions/340209/generate-colors-between-red-and-green-for-a-power-meter/340214#340214), a simple use is to generate colours between green and red, to be used for something like a power meter or a healthbar. While it can be done pretty easily in RGB, it doesn't look great, and doing it in HSV is simpler and looks much better.

Another example I found [here](https://gamedev.stackexchange.com/questions/46463/how-can-i-find-an-optimum-set-of-colors-for-10-players/46469#46469) is generating aesthetically pleasing colours, another thing that's done quite easily with HSV.

More colour spaces could be implemented in the future, but from what I read [here](https://www.sfml-dev.org/contribute.php#philosophy), I'd rather start small and add more later. It's my first time contributing to SFML, so I like getting feedback before making too many changes :)

Another colour space that could be added is HSL, which is similar to HSV, but has a few different uses. Perhaps Lab, CMYK as well, but I think those are less useful for software.

In terms of adding more functionality, `toHSVA` would definitely be useful for completeness.

Thanks for the feedback!

IvanFon:
@dmitry-t Oops - no particular reason to use `double`s here, I'll switch to `float`s.

eliasdaler:
I agree with @eXpl0it3r . I'm not sure this is really needed in SFML. There are tons of different color modes and they can easily be implemented as wrapper functions. You can just make some ColorUtils.h and have this:

```cpp
sf::Color fromHSV(...)
{
   ...
}
```

And later in your program do:

```cpp
sf::Color color = fromHSV(...);
```

RGBA is the lowest common denominator. And HSVA is not used that often by everyone. Yes, there are requests, but there are a lot of requests on SFML forums for stuff that can be implemented as a free function without the need to be included in SFML's public API.

eXpl0it3r:
> supporting other colour spaces probably wouldn't be out of place

It's not necessarily out of scope, but it needs a bit more planning and thinking, than just adding a random function and see from there.

If you want to contribute more in that direction, I suggest to open a forum thread outlining how SFML could provide more support for other color spaces. Keep also in mind that there is already support for sRGB to some degrees and that there have been discussions about other color spaces support for `sf::Texture`.

The reason we generally like to first have a discussion before something gets implemented is so people don't invest time into something that isn't really wanted as part of the library and so we don't have to say no as often, because it's not really fun to tell people that their work isn't getting merge. 😉

IvanFon:
Sorry for taking a while to reply - thanks for all the feedback. Will close this for now, I'm going to add more features and maybe make a separate library. I might make another pull request eventually if the features are wanted.

eXpl0it3r:
Sounds like a good idea! Looking forward to see what you can come up with. 🙂

Issue #1471, opened 2018-08-14
iOS apps just show as a blank magenta screen


JonnyPtn:
## Subject of the issue

iOS apps don't render anything, just showing as a blank magenta screen. 

## Environment

* macOS 10.13, Xcode 9, iOS 11.3
* SFML 2.5.0

## Steps to reproduce

Build SFML with the example for iOS and run it

## Expected behavior

Should render correctly

## Actual behavior

Blank magenta screen

## Details

I've tracked it down to commit 0adde249ec9fd2012c53b5f57f71989ffd88d16f, and more specifically [this line](https://github.com/SFML/SFML/blob/1dcad6087812a08d41c349511f48fbbf379763f6/src/SFML/Graphics/RenderWindow.cpp#L86) which when commented out solves the problem, but my openGL knowledge is too sparse to work out exactly what is causing it. 

edit: after some googling and trial-and-error I've found out if I bind the frame buffer in `EAGLContext::Display()` (add `glBindFramebufferOES(GL_FRAMEBUFFER_OES, m_framebuffer);` at the top of the function) then it works... I can't say whether this is a good solution or not, so I'm hoping someone else can!

PR #1481, opened 2018-08-23
Fix magenta screen on iOS (#1471)

JonnyPtn:
In the absence of any other feedback for issue #1471 , this change seems to solve this issue and make iOS apps functional again.

### DIFF
--- a/src/SFML/Window/iOS/EaglContext.mm
+++ b/src/SFML/Window/iOS/EaglContext.mm
@@ -185,6 +185,7 @@ bool EaglContext::makeCurrent(bool current)
 ////////////////////////////////////////////////////////////
 void EaglContext::display()
 {
+    glBindFramebufferOES(GL_FRAMEBUFFER_OES, m_framebuffer);
     glBindRenderbufferOES(GL_RENDERBUFFER_OES, m_colorbuffer);
     [m_context presentRenderbuffer:GL_RENDERBUFFER_OES];
 


binary1248:
I have no idea why binding the FBO is necessary here... But if you say it definitely fixes the problem then it might just be one of those Apple specific implementation details. 😁

Could you run through a few RenderTexture tests, including using RenderTextures with multiple contexts and see if it still all works as expected with this change?

JonnyPtn:
Ah yes you're right... using a rendertexture breaks it again. Does that help at all with where to look? I'm not really clued up on what the changes in that commit do, so I'm trying (and clearly failing) to understand as I go along.

binary1248:
Well the problem is that sfml-graphics in its current form is fighting against this file in terms of who gets to activate the FBO. In every other context implementation, all drawing related operations are isolated to sfml-graphics so it was never a problem.

The fact that misusing a FBO as a render target is a necessary solely on iOS is just one of those quirks that has you asking how drunk the Apple engineers had to be when they thought this was a good idea.

I have no idea what debugging for iOS looks like, but you could start by tracking/printing which FBO gets bound/unbound in what context. Remember that FBOs aren't shareable between contexts so if you see something like that then you know things are broken.

If I had to make a wild guess, I would say that the `priv::RenderTextureImplFBO::unbind()` in `RenderWindow::setActive()` is breaking it because it assumes that no implementation actually needs an FBO to be bound to draw to the window/default render target. You would need to make an exception for iOS in that case and make sure the context's FBO is bound in `EaglContext::makeCurrent()`.

JonnyPtn:
Thanks, I'll see if I can find any more out. debugging for iOS is more or less the same as macOS or any other platform. It seems that Xcode has some nifty openGLES frame capture tool though, so maybe that will help somewhat...

JonnyPtn:
@binary1248 Here's a quick sample of one frame of the iOS demo (So no render textures involved, just drawing a textured sprite to the window). Not sure if this helps at all: https://imgur.com/a/FZJ1pNO

I'll investigate further when I can, but some quick things I've noted which may or may not be of relevance:

- inspecting the call to `gldrawarrays` shows that "Framebuffer #0 does not exist"
- the call to `presentrenderbuffer` shows a warning: "Present of uninitialised colour attachment: Renderbuffer #1 did not have content rendered to all pixels in the renderbuffer"

eXpl0it3r:
According to @binary1248 the [`unbind()`](https://github.com/SFML/SFML/blob/master/src/SFML/Graphics/RenderWindow.cpp#L86) in RenderWindow should somehow be changed to bind the framebuffer which is created inside the EGAL context. So the main question is how to get that from the window to the graphics module. 🤔 

If we can't find a working solution soon, we'll have to either merge this change and get RTs fixed later or move everything, as I want to get SFML 2.5.1 released soon.

JonnyPtn:
A better solution than just merging this would be to revert the breaking commit entirely until a working implementation is found, that way at least we have a working iOS implementation again

eXpl0it3r:
That commit changed quite a few things, has brought great performance improvements, has modified public-facing interfaces and new things have built on top of it, as such it's not really possible to revert the commit on it's own, especially not for a patch version release.

Question is if we're able to come up with an overall fix in a short-ish period of time or what's worse, this PR or the current master. Just trying to find what's the best way here. 🙂

binary1248:
@JonnyPtn Give the [bugfix/renderwindow_default_framebuffer](https://github.com/SFML/SFML/tree/bugfix/renderwindow_default_framebuffer) branch a try.

JonnyPtn:
Yep, looks great to me, render textures and all

Issue #1473, opened 2018-08-15
iOS autocomplete window overlaps keyboard


JonnyPtn:
## Subject of the issue

The autocomplete window overlaps the keyboard, looking terrible and making typing difficult, see [here](https://imgur.com/a/AvmzVWG)

## Your environment

* macOS 10.13, Xcode 10.13, iOS 11.3

## Steps to reproduce

Set the virtual keyboard visible with `sf::Keyboard::setVirtualKeyboardVisible(true);`

## Expected behavior

Autocomplete window should appear on top of the keyboard, as per usual

## Actual behavior

Autocomplete overlaps keyboard

## Details

I haven't worked out why this is happening, but I have worked out how to hide the autocomplete window, which technically fixes the issue, at the expense of losing some functionality (But to be fair in my tests the autocomplete didn't fill correctly anyway, so that's an issue for another day).

If I don't get any other feedback in the next week or so about this I'll put together a PR to hide the autocomplete.

eXpl0it3r:
Fixed with #1473 and merged in 2bba983f1d90b26fac4560600a226156e6721b2a

PR #1482, opened 2018-08-23
Disable autocorrect on iOS for issue #1473

JonnyPtn:
The best I can do for #1473, I can't work out any reason why the autocorrect window would be bugged like that.


eXpl0it3r:
Merged in 2bba983f1d90b26fac4560600a226156e6721b2a

PR #1484, opened 2018-08-26
Allow creation of a window without an OpenGL context

binary1248:
This gives users the freedom to use SFML for pure window management and choose another rendering API of their choice.

Test with:
```cpp
#include <SFML/Window.hpp>

int main()
{
    sf::WindowBase window(sf::VideoMode(400, 400), "Test", sf::Style::Default);

    // Do stuff with window.getSystemHandle() here...

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
        }
    }
}
```

### DIFF
--- a/examples/X11/X11.cpp
+++ b/examples/X11/X11.cpp
@@ -210,6 +210,13 @@ int main()
         sfmlView2.display();
     }
 
+    // Close our SFML views before destroying the underlying window
+    sfmlView1.close();
+    sfmlView2.close();
+
+    // Destroy the window
+    XDestroyWindow(display, window);
+
     // Close the display
     XCloseDisplay(display);
 
--- a/examples/win32/Win32.cpp
+++ b/examples/win32/Win32.cpp
@@ -122,6 +122,10 @@ int main()
         }
     }
 
+    // Close our SFML views before destroying the underlying window
+    SFMLView1.close();
+    SFMLView2.close();
+
     // Destroy the main window (all its child controls will be destroyed)
     DestroyWindow(window);
 
--- a/include/SFML/Window/Cursor.hpp
+++ b/include/SFML/Window/Cursor.hpp
@@ -161,7 +161,7 @@ public:
 
 private:
 
-    friend class Window;
+    friend class WindowBase;
 
     ////////////////////////////////////////////////////////////
     /// \brief Get access to the underlying implementation
--- a/include/SFML/Window/Mouse.hpp
+++ b/include/SFML/Window/Mouse.hpp
@@ -34,7 +34,7 @@
 
 namespace sf
 {
-class Window;
+class WindowBase;
 
 ////////////////////////////////////////////////////////////
 /// \brief Give access to the real-time state of the mouse
@@ -101,7 +101,7 @@ public:
     /// \return Current position of the mouse
     ///
     ////////////////////////////////////////////////////////////
-    static Vector2i getPosition(const Window& relativeTo);
+    static Vector2i getPosition(const WindowBase& relativeTo);
 
     ////////////////////////////////////////////////////////////
     /// \brief Set the current position of the mouse in desktop coordinates
@@ -124,7 +124,7 @@ public:
     /// \param relativeTo Reference window
     ///
     ////////////////////////////////////////////////////////////
-    static void setPosition(const Vector2i& position, const Window& relativeTo);
+    static void setPosition(const Vector2i& position, const WindowBase& relativeTo);
 };
 
 } // namespace sf
--- a/include/SFML/Window/Touch.hpp
+++ b/include/SFML/Window/Touch.hpp
@@ -34,7 +34,7 @@
 
 namespace sf
 {
-class Window;
+class WindowBase;
 
 ////////////////////////////////////////////////////////////
 /// \brief Give access to the real-time state of the touches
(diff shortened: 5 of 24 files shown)


eXpl0it3r:
What's the advantage of using a struct for the parameter?

MarioLiebisch:
Rather than using a new type wouldn't it make more sense to potentially introduce a static `ContextSetting` that signifies "don't create a context", like `sf::ContextSetting::NoContext`?

MarioLiebisch:
Or maybe just a new `ContextSetting` field determining the render backend to use, for now a default value of `OpenGL` with an optional `None`?

binary1248:
> What's the advantage of using a struct for the parameter?

`sf::Window::NoContextType` is just a tag type, like `sf::Shader::CurrentTextureType`.

> Rather than using a new type wouldn't it make more sense to potentially introduce a static `ContextSetting` that signifies "don't create a context", like `sf::ContextSetting::NoContext`?

This was my original implementation, but I didn't want to add another field to `ContextSetting` that basically invalidates all other data contained in the structure. If you don't want a context to be created you shouldn't have to touch `ContextSetting`, this makes more sense to me.

binary1248:
I don't see what a "discussion" of this topic whether here or on the forum would bring. You yourself should know that there were instances of people who asked if using SFML with DirectX/Direct3D was possible over the last years, only to be told that SFML is oriented around OpenGL. I don't think they stayed around much longer after that considering that other comparable libraries actually provide what they want.

Nowadays Vulkan is the new hype, and guess what, people who want to get their feet wet are already either using GLFW or SDL because SFML forces an OpenGL context down your throat whenever you create a window. If you want to know their opinion you would have to ask them on some other forums/trackers. Starting a discussion on the forum is basically like trying to prevent unnecessary deaths by asking dead people how they died.

Since this feature is basically inevitable, we can get to the API discussion faster, and I prefer discussing such matters around real code and a working implementation instead of walls of text and pseudo-code on the forum. There is no point coming up with some fantastic API that turns out to be unimplementable in the end.

LaurentGomila:
Thanks.

So now let's start the API discussion. I'm not a big fan of the new overloads, I think we already have too many of them: ctor/create, size/handle, and now duplicate all that stuff again with the NoContext argument, we end up with 8 different ways to create a window. This could be confusing for the end user.

Another bad consequence of this addition, is that we now have 5 functions in `sf::Window` that become totally useless (not to say out of scope) if we choose to create a context-less window. This strongly asks for a different type rather than an option at creation time. For example, we could have a sf::Window base class and a derived sf::OpenGlWindow class (that's just an example).

The abstraction of the rendering back-end will most likely require the same kind of refactoring -- actually, this feature is just a special case of it: a null/empty back-end -- so maybe we can integrate this feature at the same time?

binary1248:
We can't change the external functionality or rename `sf::Window` without breaking existing code, so we would have to move most of the implementation into the new null window that would sit under sf::Window. This is a change that is so small I would prefer not delaying it until the next time we can make API breaks.

binary1248:
Split `sf::Window` into `sf::Window` (window with OpenGL context) and `sf::WindowBase` (basic window functionality). See example above for usage.

LaurentGomila:
That's a good starting point.

You forgot to remove `m_clock` and `m_frameTimeLimit` from `sf::WindowBase`.

In my opinion there's still one important issue to solve with this design: we have the same overloads of `create` (when you leave the `ContextSettings` argument to its default value in `sf::Window`) in both classes, but which give different results.

```cpp
sf::Window window;
window.create(mode, title, style); // OpenGL context
static_cast<sf::WindowBase&>(window).create(mode, title, style); // no OpenGL context

// we also have no way to create a sf::Window without a context, although it inherits from (is a) sf::WindowBase
```

I have no solution right now, I'll try to think about it.

binary1248:
I split the default parameters into their own overloads and made all the `::create` methods virtual. You shouldn't be able to call the wrong kind of `::create` now as in your example.

I don't mind not being able to create an `sf::Window` without a context. This is the exact reason why `sf::WindowBase` exists and the whole idea why virtual functions are a language feature. `sf::Window` is nothing more than `sf::WindowBase` + OpenGL context. Liskov substitution is still possible: You can pass an `sf::Window` to anything expecting a `sf::WindowBase`. The opposite however is understandably not possible.

LaurentGomila:
> I don't mind not being able to create an sf::Window without a context

Yes, I can live with that too. I thought more about it and the only design that came to my mind is this one:

```cpp
sf::Window window(...); // simple OS window, no context
sf::OpenGLContext context(window); // no longer owned by sf::Window

window.pollEvent(...); // window API
context.display(); // context API

void draw(sf::OpenGLContext& context); // you now draw to a context, not a window
```

I like it a lot for many reasons, but I think it would be too annoying for the end user to manage those two objects in parallel.

binary1248:
Conceptually that might have a lot of overlap with the concept of `sf::RenderTarget` that we have now...

I can't think of anything else besides `setActive()` and `display()` that would go into such an `sf::OpenGLContext` object if we wanted to avoid such overlap...

LaurentGomila:
Yes, that would be of course a replacement for `sf::RenderTarget`. But anyway, we're not really going to implement that, so let's forget about it ;)

JonnyPtn:
This works great for me with opengl/vulkan/dx/metal backend on windows and mac

MarchingCube:
Hi, is there any plan to merge this? My team is currently porting our renderer to Vulkan and forced OpenGL context from SFML does not play too well with it :) If not is there any known workaround to get rid of OpenGL context?

binary1248:
As far as I can tell, the review process left off with @LaurentGomila requesting changes?

If that is not/no longer the case I don't see any other reason to stall merging this in its current state.

PR #1485, opened 2018-08-27
Add -ObjC flag to fix static linking on macOS

JonnyPtn:
* [x] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php?topic=24169.0)

----

## Description

When static linking on macOS you get a runtime error along the lines of:

`Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[SFOpenGLView enableKeyRepeat]: unrecognized selector sent to instance 0x7fe0b762c5c0'`

This is because theres a missing linker flag which is required when using Objective-C code, documentation [here](https://developer.apple.com/library/archive/qa/qa1490/_index.html)

## Tasks
* [x] Tested on macOS
* [x] Tested on iOS

## How to test this PR?

Build SFML with examples and static linking (cmake flag `-DBUILD_SHARED_LIBS=FALSE`), then try to run any example.

### DIFF
--- a/src/SFML/Window/CMakeLists.txt
+++ b/src/SFML/Window/CMakeLists.txt
@@ -229,6 +229,11 @@ sfml_add_library(sfml-window
                  SOURCES ${SRC} ${PLATFORM_SRC})
 target_link_libraries(sfml-window PUBLIC sfml-system)
 
+# When static linking on macOS, we need to add this flag for objective C to work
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
Merged in 9fba1dff27520e9af464ebacf0c4b3ce23a2dd26

PR #1491, opened 2018-10-01
Fixed Windows cursor color conversion to be endian safe

eXpl0it3r:
The fix implemented by #1465 wasn't endian-safe so this fixes that.

## Tasks

* [X] Tested on Windows

## How to test this PR?

To test, follow the instruction from the original issue: #1464


eXpl0it3r:
Merged in 3aa156c27837ee9c1131c6f603fdc31eac5ab565

PR #1492, opened 2018-10-01
Fixed the error message when the wrong bitmap font size is selected

eXpl0it3r:
Fixes the issue that was a first wrongfully reported in #1474 and then identified as wrong error message.

## Tasks

* [ ] Tested on Linux
* [X ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Follow the instructions in #1474


eXpl0it3r:
If anyone is willing to give this a review, @Qu3tzal, @texus, @LaurentGomila

PR #1497, opened 2018-10-09
Initial travis and Appveyor configs

JonnyPtn:
Not intended as a replacement for the SFML buildbot, but to complement it.

Just needs someone from the SFML team to connect their GitHub accounts with the respective services and it will be good to go

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,56 @@
+sudo: false
+dist: trusty
+language: c++
+
+addons:
+  apt:
+    packages:
+    - cmake
+    - libxrandr-dev
+    - libudev-dev
+    - libopenal-dev
+    - libflac-dev
+    - libvorbis-dev
+    - g++
+    - clang
+
+before_script:
+- mkdir build && cd build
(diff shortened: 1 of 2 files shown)


LaurentGomila:
> Not intended as a replacement for the SFML buildbot, but to complement it.

Can you explain what it adds?

JonnyPtn:
Automatic building of PRs ( doesn’t depend on a team member to trigger the buildbot, contributors will get feedback within minutes if the build fails)

Hosted solution ( doesn’t depend on buildbot slaves all being running and healthy)

Easier collaboration (they are much more widespread than buildbot, so more people will be familiar with them. Also as configs are in the main repo people don’t need to check out the entire buildbot repo and set up their own buildbot instance just to test their changes, which would also only be limited to their platform unless they have a comprehensive build slave set up - unlikely)

Most importantly, it can just take some administrative load off some members of the SFML team so they can focus more on the important stuff.

eXpl0it3r:
Merged in 1159ac5c522a4819263ef7fd8628e0ea8d624862

PR #1499, opened 2018-10-16
Fix cast

AMDmi3:
Fixes clang error:

```
src/SFML/Window/Unix/WindowImplX11.cpp:1559:36: error: reinterpret_cast from 'nullptr_t' to 'void *' is not allowed
                                   reinterpret_cast<void*>(NULL));
                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```


LaurentGomila:
What about not casting at all? NULL should always be implicitly convertible to any type of pointer, without a cast.

AMDmi3:
Done.

eXpl0it3r:

eXpl0it3r:
Merged in 119195ac20866fcecc93e33cda457b3fe54b1b0c

Issue #1503, opened 2018-10-25
Add color blending modes to sf::BlendMode


FRex:
So the idea of 'supporting color blending' came up on the forum: https://en.sfml-dev.org/forums/index.php?topic=24653

There is apparently one 'blend color' set by `glBlendColor`. It then gets used by some blending factors. Relevant ones are `GL_CONSTANT_COLOR` and probably/maybe also `GL_ONE_MINUS_CONSTANT_COLOR`, `GL_CONSTANT_ALPHA` and `GL_ONE_MINUS_CONSTANT_ALPHA`.

It should be easy enough to implement. Add an sf::Color field to sf::BlendMode, add 1-4 relevant enums, add/change caching and applying the sf::BlendMode in sf::RenderTarget.

Arc8ne:
Is it alright if I could be assigned to this issue?

eXpl0it3r:
Feel free to create a PR for it. You don't need to be assigned to it for that. 🙂

Arc8ne:
@eXpl0it3r I see, just wanted to check first to prevent a situation where multiple people are unknowingly working on their own PRs for the same issue, thanks.

PR #1505, opened 2018-10-26
Add support for blend color

cier:
## Description

Adds support for constant color blending modes, by adding GL_EXT_blend_color OpenGL extension.
Support for OpenGL ES is not implemented yet.

This PR is related to the issue #1503.
Forum: https://en.sfml-dev.org/forums/index.php?topic=24653.0

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?
```cpp
#include <SFML/Graphics.hpp>

int main()
{
    // Create a window for the game
    sf::RenderWindow window(sf::VideoMode(800, 600), "Minimal example");

    sf::CircleShape circle(150);
    circle.setFillColor(sf::Color(255, 255, 255, 255));

    sf::BlendMode mode(sf::BlendMode::ConstColor, sf::BlendMode::Zero, sf::BlendMode::Add, sf::Color::Red);

    for (;;)
    {
        // ---------- Handle Events -----------
        sf::Event event;
        {
            // Handle events
            if (event.type == sf::Event::Closed
                || event.type == sf::Event::KeyPressed && event.key.code == sf::Keyboard::Escape)
                return 0;
        }

        // ---------- Draw the scene ----------
        window.clear();

        window.draw(circle, mode);

        // Display the frame
        window.display();
    }
}
```

### DIFF
--- a/include/SFML/Graphics/BlendMode.hpp
+++ b/include/SFML/Graphics/BlendMode.hpp
@@ -29,7 +29,7 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Graphics/Export.hpp>
-
+#include <SFML/Graphics/Color.hpp>
 
 namespace sf
 {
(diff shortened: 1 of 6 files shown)


JonnyPtn:
If I understand the example correctly this works on macOS (It's meant to show a red circle right?)

Unfortunately doesn't work on iOS though, some undeclared identifiers:

```
GL_CONSTANT_COLOR
GL_ONE_MINUS_CONSTANT_COLOR
GL_CONSTANT_ALPHA
GL_ONE_MINUS_CONSTANT_ALPHA
```

I'm no expert, but some light searching indicates to me that these weren't available until GLES 2 which we don't support, so we'll need a workaround for that

eXpl0it3r:
If we can't support this for mobile, then we need to have the API no do anything on these platforms and not prevent compilation or spew out errors from OpenGL during runtime.

Also needs a rebase.

LaurentGomila:
This PR lacks some explanations about why this feature would be a good addition (use cases, etc.).

And are the `GL_ONE_MINUS_` variants really needed? The parameter is some hard-coded constant, so it seems a little overkill to add blending factors just to make the GPU/driver compute one minus that constant. Unless I'm missing something about this mode.

JonnyPtn:
From the linked forum post:
>But why iterate over the mesh everytime the color changes to update the vertex colors, when there's already a corresponding OpenGL blend mode available?
>For certain color effects, like color flashing, you would need to update the vertex colors every frame. Depending on mesh size and the amount of objects that are affected, this seems like quite a bit of unnecessary overhead.

PR #1510, opened 2018-11-07
CMake: create SFML_LIB_SUFFIX option

markand:
----

## Description

Make the hardcoded LIB_SUFFIX a first class citizen to let the user pass
a new value if needed.

## Tasks

* [x] Tested on Linux
* [x] Tested on Android

## How to test this PR?

This will install in lib64

```
cmake .. -DSFML_LIB_SUFFIX="64" && make && make install
```

This will install in subdirectory lib/armeabi-v7a

```
cmake .. -DSFML_LIB_SUFFIX="/armeabi-v7a" && make && make install
```

Not specifying a value keep the old behaviour set with DEFAULT_LIB_SUFFIX.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,7 +115,7 @@ if(SFML_OS_ANDROID)
     set(CMAKE_INSTALL_PREFIX ${CMAKE_ANDROID_NDK}/sources/third_party/sfml)
 
     # we install libs in a subdirectory named after the ABI (lib/mips/*.so)
-    set(LIB_SUFFIX "/${CMAKE_ANDROID_ARCH_ABI}")
+    set(DEFAULT_LIB_SUFFIX "/${CMAKE_ANDROID_ARCH_ABI}")
 
     # pass shared STL configuration (if any)
     if (CMAKE_ANDROID_STL_TYPE MATCHES "_shared")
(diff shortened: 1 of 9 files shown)


eXpl0it3r:
For Android you can already achieve this by setting `${CMAKE_ANDROID_ARCH_ABI}`, I don't think we should be able to manually change it otherwise, as it's determined by the architecture used.

markand:
No, because I explicitly need to set CMAKE_ANDROID_ARCH_ABI, I can't set it to an empty string even though I would like to install in lib/ without suffix.

eXpl0it3r:
Doesn't `CMAKE_INSTALL_PREFIX` overwrite this then?

markand:
No because CMAKE_INSTALL_PREFIX controls the prefix of everything else as parent directory (include, lib, bin, share). The purpose of this option is to control only the lib/ suffix for people who prefer having everything in lib/ directly (rather than lib/<arch>). Since many build system only search lib/

OvermindDL1:
Doesn't this seem wrong on two front?

1. Any generated SFMLConfig.cmake file already gives the proper locations.

2. If you are wanting to set a debug prefix, like adding a `d` on the end of the filename of the output library as is common on Windows then why not just pass `-DCMAKE_DEBUG_POSTFIX=d` to the `cmake` call since that is the built-in way of doing that instead of inventing your own option?

sum01:
You could instead replace instances of `bin` and `lib` in `install()` calls with `${CMAKE_INSTALL_BINDIR}` and `${CMAKE_INSTALL_LIBDIR}`, which can be defaulted by doing `include(GNUInstallDirs)`. Example:
```cmake
# Sets OS-specific defaults which can be overriden
include(GNUInstallDirs)
install(TARGETS ${target} EXPORT SFMLConfigExport
  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} COMPONENT bin
  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT bin
  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT devel
)
```

Then to get the suffix desired, you'd do `cmake -DCMAKE_INSTALL_LIBDIR=lib/yoursuffixhere ..` or whatever you want the libdir to be.

This avoids project-specific options, which in my opinion too many projects are rife with.

markand:
This was planned after this PR. But I think I'll do it now since it does not gonna be merged anyway.

eXpl0it3r:
That suggestion sounds like a better approach. 🙂

PR #1511, opened 2018-11-07
CMake: don't set a predefined build type

markand:
## Description

The CMakeLists.txt should not set a CMAKE_BUILD_TYPE by default because
multiple generators do not use this variable (e.g. Visual Studio).

This makes the build process inconsistent between the CMake
configuration and the IDE.

Example with current behaviour:

1. User runs CMake for Visual Studio, CMake generates Release
2. User selects Debug in Visual Studio

The build process is inconsistent.

## Tasks

* [x] Tested on Linux

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,9 +15,6 @@ endmacro()
 
 # these options have to be set before CMake detects/configures the toolchain
 
-# determine whether to create a debug or release build
-sfml_set_option(CMAKE_BUILD_TYPE Release STRING "Choose the type of build (Debug or Release)")
-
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
I disagree. It's still inconsistent if you set nothing and then get multiple options. It's just expected behavior that some generator will provide multiple types.

ChrisThrasher:
I agree that CMake projects should not hardcode such details and that it makes more sense to default to a debug build.

Bromeon:
This has been closed 3.5 years ago 🙂
If you think we should change something, maybe open a new issue about the topic?

PR #1516, opened 2018-11-24
allow for closing sound files

vfjpl:

----

## Description

Sound files need to be closed to write metadata to files. Closing is performed automatically when the object goes out of scope. This patch allow for closing files at any moment.

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Audio/OutputSoundFile.hpp>
#include <SFML/Audio/InputSoundFile.hpp>

int main()
{
    sf::OutputSoundFile writer;
    writer.openFromFile("filename.wav", 48000, 1);
    //writer.write();
    writer.close();

    sf::InputSoundFile reader;
    reader.openFromFile("filename.wav");
    //reader.read();
    reader.close();
}
```


LaurentGomila:
Why not. But in C++, a good practice is to avoid keeping "uninitialized" variables alive ; the same way that you only construct a variable when you need it.

```cpp
#include <SFML/Audio/OutputSoundFile.hpp>
#include <SFML/Audio/InputSoundFile.hpp>

int main()
{
    {
        sf::OutputSoundFile writer;
        writer.openFromFile("filename.wav", 48000, 1);
        //writer.write();
        writer.close();
    }

    {
        sf::InputSoundFile reader;
        reader.openFromFile("filename.wav");
        //reader.read();
        reader.close();
    }
}
```

vfjpl:
@LaurentGomila If sf::OutputSoundFile is member of a class and you can't destroy object of that class then you can't write metadata. :)

Thank you for accepting the patch.

PR #1517, opened 2018-11-24
Allow changing processing interval in SoundStream

vfjpl:

----

## Description

This patch bring similar functionality that is available in SoundRecorder. It is useful to be able to change it when you have your own waiting/blocking calls in onGetData function. That way sleep in SoundRecorder don't get in your way. After this patch default behaviour is the same.

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Audio/SoundStream.hpp>
class FastSoundStreamt: public sf::SoundStream
{
    FastSoundStream()
    {
         sf::SoundRecorder::setProcessingInterval(sf::Time::Zero);
    }
    void onSeek(sf::Time timeOffset) {}
    bool onGetData(sf::SoundStream::Chunk& data)
    {
        //custom blocking calls for samples
    }
};
```

### DIFF
--- a/include/SFML/Audio/SoundStream.hpp
+++ b/include/SFML/Audio/SoundStream.hpp
@@ -251,6 +251,24 @@ protected:
     ////////////////////////////////////////////////////////////
     virtual Int64 onLoop();
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Set the processing interval
+    ///
+    /// The processing interval controls the period
+    /// between calls to the fillAndPushBuffer function. You may
+    /// want to use a small interval if you want to process the
+    /// captured data in real time, for example.
+    ///
+    /// Note: this is only a hint, the actual period may vary.
+    /// So don't rely on this parameter to implement precise timing.
+    ///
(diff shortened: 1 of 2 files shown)


include/SFML/Audio/SoundStream.hpp
```diff
@@ -251,6 +251,24 @@ class SFML_AUDIO_API SoundStream : public SoundSource
     ////////////////////////////////////////////////////////////
     virtual Int64 onLoop();
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Set the processing interval
+    ///
+    /// The processing interval controls the period
+    /// between calls to the fillAndPushBuffer function. You may
```
LaurentGomila: You should rather talk about `getData`, `fillAndPushBuffer` means nothing for the user.

eXpl0it3r:
Did you intentionally close this PR?

PR #1522, opened 2018-11-29
iOS example updates

JonnyPtn:
This adds a few things to #1520 :

- Touch support in all 3 examples
- Icons and launch screens for all 3 examples
- Handling of screen resize (i.e. device rotation) correctly on all 3 examples

The only outstanding issue is that the cube on the openGL example isn't showing correctly. not sure if it's a GLES issue or something?

I've tested on macOS and the examples still seem fine, but I'd encourage testing on other platforms to ensure my changes haven't broken anything


eXpl0it3r:

PR #1523, opened 2018-11-29
Unit test updates

JonnyPtn:
Rebased onto master and updated travis/appveyor configs to build and run the tests

Should make #1475 ready for merge


LaurentGomila:
> Should make #1475 ready for merge

What about "Find a solution for dynamic library not being found for the runtests target"?

JonnyPtn:
Good point, appears to only be an issue on Windows so I'll look into it when I get on a windows machine.

Also seems like an odd error on iOS which I'll look into

LaurentGomila:
Why do you add `${CMAKE_BINARY_DIR}/lib/Debug` to the PATH? I would rather use generator expressions to get the actual output path of one of the sfml-xxx targets.

JonnyPtn:
Fixed and ready for review.

iOS is currently disabled as it's not possible to run iOS unit tests the same way as all the other platforms. There may be some way using the xctest support [here](https://cmake.org/cmake/help/v3.3/module/FindXCTest.html) which I'll look into when I can

Anyone know what the deal with unit testing on android is? I'll add it to my list but for now I'd say let's get this going to cover the main platforms

LaurentGomila:
That was quickly merged...

Why don't we build everything into a "${CMAKE_BINARY_DIR}/bin" folder, rather than copying SFML DLLs in every sub-folder they are needed?

eXpl0it3r:
It was simply merged onto our feature branch, see #1475

LaurentGomila:
Ah, ok.

But anyway, even if it's not merged into the master branch, it would have been convenient to end this review here rather than having to open a new PR 😛

PR #1526, opened 2018-12-03
Use travis for windows builds

JonnyPtn:
Now travis has windows images, we don't need to use appveyor to cover all 3 platforms.

Also did a bit of tidying in the travis config


.travis.yml
```diff
@@ -13,44 +11,42 @@ addons:
     - libvorbis-dev
     - g++
     - clang
+    - freeglut3-dev
```
eXpl0it3r: Why did you add freeglut? Did you mean to add freetype?

JonnyPtn: As discussed, it was a stab in the dark based on googling, and we've established libgl1-mesa-dev is better, so that's changed now

PR #1529, opened 2018-12-28
made the display method a virtual part of the RenderTarget class

emersont1:
## Description

Quite simply, it makes display a virtual method of display, so you can have a renderer agnostic of whether it's using `sf::RenderWindow` or `sf::RenderTexture` like so

```cpp
class generic_renderer{
   // ...
  sf::RenderTarget* target; // or a smart pointer
  public:
  void render(){
    target->clear();
    for(auto& d: drawables){
      // draw using the target
    }
    target->display();    
  }
}
```

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    window.setFramerateLimit(60);
    sf::RenderTarget* target = &window;

    sf::CircleShape shape(50.f);

    // set the shape color to green
   shape.setFillColor(sf::Color(100, 250, 50));

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
        }

        target->clear();
        target->draw(shape);
        target->display();
    }
}
```


emersont1:
right, I'll add that in

LaurentGomila:
So, it was never really agreed to do that change 😉

eXpl0it3r:
As the current consensus is to not implement this and since the author of this PR seems to have deleted their account, I'll close this for now.

Discussion on this topic should be best continued in the [existing forum thread](https://en.sfml-dev.org/forums/index.php?topic=23803.0).

PR #1535, opened 2019-01-03
NetBSD support

niacat:


src/SFML/Window/CMakeLists.txt
```diff
@@ -79,7 +79,8 @@ if(SFML_OS_WINDOWS)
 
     # make sure that we use the Unicode version of the Win API functions
     add_definitions(-DUNICODE -D_UNICODE)
-elseif(SFML_OS_LINUX OR SFML_OS_FREEBSD OR SFML_OS_OPENBSD)
+elseif(SFML_OS_LINUX OR SFML_OS_FREEBSD OR SFML_OS_NETBSD OR SFML_OS_OPENBSD)
+	message("weh")
```
LaurentGomila: Not sure if this is really needed 😄

LaurentGomila:
Is there any significant difference between Open/Free/NetBSD, or could we just merge all that stuff into a generic BSD implemenation?

ensra:
> Is there any significant difference between Open/Free/NetBSD, or could we just merge all that stuff into a generic BSD implementation?

Since the code is filled with ifdef (linux or freebsd or netbsd) you could probably just have a unix system type and handle the special cases within that for different joystick APIs and similar.

krytarowski:
> Is there any significant difference between Open/Free/NetBSD, or could we just merge all that stuff into a generic BSD implementation?

If you want to get a generic switch, go for UNIX or POSIX. BSDs are diverged for over 20 years.

eXpl0it3r:
In the meantime we've gotten another contribution (#1723) with a working Joystick implementation, as such I'll close this PR.

PR #1538, opened 2019-01-08
Android: Set SYSTEM_UI_FLAG_HIDE_NAVIGATION when going to Fullscreen …

auzden:
…mode

----

## Description

Hides the navigation bar on API >= 14. On Android >= API 19 when immersive sticky mode is enabled the navigation bar can be swiped to be shown by the user.

https://developer.android.com/reference/android/view/View#SYSTEM_UI_FLAG_HIDE_NAVIGATION

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [x] Tested on Android

## How to test this PR?

On line 82 on main.cpp of the Android SFML example, change the window constructor to be:
```sf::RenderWindow window(screen, "", sf::Style::Fullscreen);```

### CODE REVIEW

**Comment by eXpl0it3r** [comment]
Shouldn't there be some code to remove the flag when switching back to windowed mode or is that automatically the case due to having to call `window.create()` to change the mode?

**Comment by eXpl0it3r** [comment]
So calling create() with window styling will make it appear again correctly, right?

File: src/SFML/Main/SFMLActivity.cpp [hunk 1]
```diff
@@ -183,13 +183,11 @@ void ANativeActivity_onCreate(ANativeActivity* activity, void* savedState, size_
     loadLibrary("sfml-window", lJNIEnv, ObjectActivityInfo);
     loadLibrary("sfml-graphics", lJNIEnv, ObjectActivityInfo);
     loadLibrary("sfml-audio", lJNIEnv, ObjectActivityInfo);
-    loadLibrary("sfml-network", lJNIEnv, ObjectActivityInfo);
```
eXpl0it3r: Why was that removed?

examples/android/build.gradle
```diff
@@ -1,13 +1,13 @@
...
 buildscript {
-    
+
     repositories {
         google()
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.0.0'
+        classpath 'com.android.tools.build:gradle:3.1.0'
```
eXpl0it3r: What's the difference between 3.0 and 3.1?

examples/android/app/build.gradle
```diff
@@ -4,8 +4,8 @@ android {
     compileSdkVersion 27
     defaultConfig {
         applicationId "org.sfmldev.android"
-        minSdkVersion 14
-        targetSdkVersion 19
+        minSdkVersion 19
+        targetSdkVersion 26
```
eXpl0it3r: What impact does it have updating these versions?

.gitignore
```diff
@@ -0,0 +1 @@
+/[Bb]uild/
```
eXpl0it3r: Don't think we need that.

src/SFML/Main/SFMLActivity.cpp
auzden: Bleh, my bad. Forgot I had an open PR.

examples/android/build.gradle
MarioLiebisch: It's basically the build system plugin that's downloaded and used. It's not necessarily mandatory to change, but might improve compatibility.

examples/android/app/build.gradle
MarioLiebisch: they essentially define the minimum Android version the app supports together with the API/header version it's working with. In this case it's targeting for stuff as defined in API 26, but versions down to 19 are supported with necessary adjustments/workarounds for those.

.gitignore
MarioLiebisch: the folder name is defaulted by Gradle, so I'm undecided here.

src/SFML/Main/SFMLActivity.cpp
MarioLiebisch: Just create an extra branch for the next PR. :)

eXpl0it3r:
Superseded by #1554

PR #1539, opened 2019-01-08
Fixing issue with already defined targets, silently return

meveric:
https://en.sfml-dev.org/forums/index.php?topic=25001.0
Discussed by a different user, tested and confirmed working.
Not provided, but the command used in the forum now works fine.
Small enough change that it shouldn't affect anything else.

## Description
As discussed in Forum https://en.sfml-dev.org/forums/index.php?topic=25001.0 this fixes build for GLES/EGL platforms where sfml-window tries to define targets EGL and GLES again which were already defined.

## Tasks

* [X] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?
cmake -DSFML_OPENGL_ES=1 .
should work without error about already defined targets EGL and GLES.

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -276,7 +276,7 @@ function(sfml_find_package)
     list(REMOVE_AT ARGN 0)
 
     if (TARGET ${target})
-        message(FATAL_ERROR "Target '${target}' is already defined")
+        return()
     endif()
 
     cmake_parse_arguments(THIS "" "" "INCLUDE;LINK" ${ARGN})


LaurentGomila:
Thanks :+1:

I would:
* Leave a comment that some dependencies may be searched by more than one module and thus we must skipped subsequent calls to sml_find_package
* Remove these "hacks" and put back calls to sfml_find_package, now that it's fixed:

```cmake
# Target OpenGL already defined for Window component so no sfml_find_package() here
target_link_libraries(sfml-graphics PRIVATE OpenGL)
```

eXpl0it3r:
Any chance to get an update on this, @meveric? 🙂 

Otherwise, I'll pick it up. The line that needs further change as suggested by @LaurentGomila would be: https://github.com/SFML/SFML/blob/master/src/SFML/Graphics/CMakeLists.txt#L119

LaurentGomila:
> The line that needs further change as suggested by @LaurentGomila would be: https://github.com/SFML/SFML/blob/master/src/SFML/Graphics/CMakeLists.txt#L119

There are more (L125, L129). We should review all files to make sure everything is cleaned.

meveric:
> Any chance to get an update on this, @meveric? slightly_smiling_face
> 
> Otherwise, I'll pick it up. The line that needs further change as suggested by @LaurentGomila would be: https://github.com/SFML/SFML/blob/master/src/SFML/Graphics/CMakeLists.txt#L119

I haven't done any other tests yet.
If you want you can take it over. My initial change worked fine for me.
If you want me to test something let me know and I can build and test it.

Do you guys actually have ARM boards to test this on?
Since I have a couple other issues I normally fix but I'm not sure if they are related to my board, or a general X11/EGL issue.

BTW there were also some people asking about EGL/GLES support without X11 any Idea about this?

eXpl0it3r:
Superseded by #1577

PR #1540, opened 2019-01-09
Removed old android cmake toolchain

JonnyPtn:
Since #1350 the android toolchain in the SFML repo is redundant, should be using the cmake one. Think the file removal was missed though, so this takes care of that


eXpl0it3r:

PR #1541, opened 2019-01-10
Add travis android builds

JonnyPtn:
This adds android builds to travis (x86 and armeabi-v7a)

It targets android API version 26, which is currently the minimum required target version in the [docs](https://support.google.com/googleplay/android-developer/answer/113469#targetsdk)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,5 +78,28 @@ matrix:
     env:
       - CMAKE_FLAGS="-DBUILD_SHARED_LIBS=FALSE"
 
+  - name: "Android armeabi-v7a"
+    language: android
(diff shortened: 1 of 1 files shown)


MarioLiebisch:
Not that it would make any significant difference, but targeting Android API 26 is **not** mandatory right now. This is **only** when using the new Play **Console** as far as I know, not Google Play itself.

JonnyPtn:
[This](https://developer.android.com/distribute/best-practices/develop/target-sdk) other source indicates it applies to google play too

MarioLiebisch:
Okay, sounds right then. Weird it's specifically mentioning the Console then.

PR #1495, opened 2018-10-06
Fix OpenGL extensions not always loading

hobby8:
Fix OpenGL extensions not loading in a specific Linux VM configuration ([here](https://en.sfml-dev.org/forums/index.php?topic=21384.msg165289)).

PR #1543, opened 2019-01-15
Fix OpenGL extensions not loading

eXpl0it3r:
## Description

Fix OpenGL extensions not loading in a specific Linux VM configuration ([here](https://en.sfml-dev.org/forums/index.php?topic=21384.msg165289)).

Rebased the PR #1495

### DIFF
--- a/src/SFML/Window/GlContext.cpp
+++ b/src/SFML/Window/GlContext.cpp
@@ -218,10 +218,12 @@ namespace
         extensions.clear();
 
         // Check whether a >= 3.0 context is available
+        glGetStringiFuncType glGetStringiFunc = NULL;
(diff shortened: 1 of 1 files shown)


src/SFML/Window/GlContext.cpp
```diff
@@ -234,28 +236,21 @@ namespace
                     extensionString++;
 
                 extensions.push_back(std::string(extension, extensionString));
-            }
-            while (*extensionString++);
+            } while (*extensionString++);
```
binary1248: Unnecessary change?

eXpl0it3r: Woops, oversight while resolving merge conflict.

PR #1552, opened 2019-02-07

hobby8:
Extending `Font` and `Text` to support multiple textures so that glyphs are not discarded in case one texture gets filled.

This can be useful in any of the following scenarios:
- Big text (in pixels)
- Large alphabets
- OpenGL implementations with a small maximum texture size (e.g. software renderers)

Reimplementation of #1333 (and previously, #1034) adjusted for recent SFML changes.


hobby8:
Rebased #1333 instead

PR #1554, opened 2019-02-10
Fixed navigation bar not being entirely hidden on Android

acsbendi:
## Description

In full-screen mode on Android, the navigation bar was not entirely hidden which I tested on multiple devices. After changing a flag, the navigation bar started to behave as expected.

Also refer to [the official sample about immersive mode](https://github.com/googlesamples/android-ImmersiveMode/blob/e07e2ac48fd9796240e6f9ba721d5fc3b1dd4b3a/Application/src/main/java/com/example/android/immersivemode/ImmersiveModeFragment.java#L80).

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [X] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>

using std::string;
using sf::Font;
using sf::Sprite;
using sf::Texture;

{

    sf::RenderWindow window(screen, "", sf::Style::Fullscreen);
    window.setFramerateLimit(30);

    sf::Texture texture;
    if(!texture.loadFromFile("image.png"))
        return EXIT_FAILURE;

    sf::Sprite image(texture);
    image.setPosition(screen.width / 2, screen.height / 2);
    image.setOrigin(texture.getSize().x/2, texture.getSize().y/2);

    sf::Font font;
        return EXIT_FAILURE;

    sf::Text text("Tap anywhere to move the logo.", font, 64);
    text.setFillColor(sf::Color::Blue);
    text.setPosition(10, 10);

    // Loading canary.wav fails for me for now; haven't had time to test why

    /*sf::Music music;
    if(!music.openFromFile("canary.wav"))
        return EXIT_FAILURE;
    music.play();*/

    sf::Color background = sf::Color::Black;

    bool active = true;

    while (window.isOpen())
    {
        sf::Event event;

        {
            switch (event.type)
            {
                case sf::Event::Closed:
                    window.close();
                    break;
                case sf::Event::KeyPressed:
                        window.close();
                    break;
                case sf::Event::Resized:
                    window.setView(view);
                    break;
                case sf::Event::LostFocus:
                    break;
                case sf::Event::GainedFocus:
                    break;

                case sf::Event::MouseLeft:
                    active = false;
                    break;
                case sf::Event::MouseEntered:
                    active = true;
                    break;
                case sf::Event::TouchBegan:
                    if (event.touch.finger == 0)
                    {
                        image.setPosition(event.touch.x, event.touch.y);
                    }
                    break;
                default:
                    break;
            }
        }

        if (active)
        {
            window.clear(background);
            window.draw(image);
            window.draw(text);
            window.display();
        }
        else {
        }
    }
}
```

### DIFF
--- a/src/SFML/Main/MainAndroid.cpp
+++ b/src/SFML/Main/MainAndroid.cpp
@@ -175,7 +175,7 @@ void goToFullscreenMode(ANativeActivity* activity)
     // API Level 14
     if (apiLevel >= 14)
     {
-        jfieldID FieldSYSTEM_UI_FLAG_LOW_PROFILE = lJNIEnv->GetStaticFieldID(classView, "SYSTEM_UI_FLAG_LOW_PROFILE", "I");
+        jfieldID FieldSYSTEM_UI_FLAG_LOW_PROFILE = lJNIEnv->GetStaticFieldID(classView, "SYSTEM_UI_FLAG_HIDE_NAVIGATION", "I");
         jint SYSTEM_UI_FLAG_LOW_PROFILE = lJNIEnv->GetStaticIntField(classView, FieldSYSTEM_UI_FLAG_LOW_PROFILE);
         flags |= SYSTEM_UI_FLAG_LOW_PROFILE;
     }


eXpl0it3r:
Just realized we had a similar PR open #1538 but the author didn't provide an updated one. Looking at the two changes, what's the difference between yours and the other one: https://github.com/SFML/SFML/pull/1538/files#diff-a9064007c5b52fa9f173fdd52d3f0dc3R177 ?

acsbendi:
@eXpl0it3r Apart from the other files changed in that PR, mine removes setting the other flag (SYSTEM_UI_FLAG_LOW_PROFILE) which doesn't have any effect after setting SYSTEM_UI_FLAG_HIDE_NAVIGATION and is therefore unnecessary.

eXpl0it3r:

PR #1557, opened 2019-02-15
Add support for creating Vulkan window surfaces.

binary1248:
Add support for creating Vulkan window surfaces.

Example is provided in the patch.


eliasdaler:
Can you please give me some ideas at what and how do you want this to be reviewed? 

Reviewing such a huge patch is scary, but I think if you tell me what to check here, I'll be happy to devote some time to reviewing this thoroughly. :)

binary1248:
The funny thing is that over 90% of the patch lines aren't even SFML code, it is all from the Vulkan headers.

Then the next bigger chunk of changes come from the new Vulkan example, which you could look at... but if you aren't experienced with Vulkan you would also find pretty daunting (still worth a look though).

The smallest chunk and the code that should be "most" reviewed are the actual changes to the library and its public interface. These can be found in the files in the `src` and `include` directories as usual.

include/SFML/Window/WindowBase.hpp
```diff
@@ -393,6 +394,18 @@ class SFML_WINDOW_API WindowBase : NonCopyable
...
+    /// \brief Create a Vulkan rendering surface
+    ///
+    /// \param instance  Vulkan instance
+    /// \param surface   Created surface
+    /// \param allocator Allocator to use
+    ///
+    /// \return True if surface creation was successful, false otherwise
+    ///
+    ////////////////////////////////////////////////////////////
+    bool createVulkanSurface(const VkInstance& instance, VkSurfaceKHR& surface, const VkAllocationCallbacks* allocator = 0);
```
eXpl0it3r: Since `sf::WindowsBase` acts as the base for the regular `sf::Window`, does it the `createVulkanSurface()` make sense in the context of `sf::Window` or `sf::RenderWindow`?

src/SFML/Window/Unix/VulkanImplX11.cpp
```diff
@@ -0,0 +1,228 @@
...
+#include <map>
+#include <cstring>
+
+
+namespace
+{
+    // RAII Vulkan shared library wrapper
+    struct VulkanLibraryWrapper
+    {
+        // Ctor
```
eXpl0it3r: Kind of redundant comment, but 🤷‍♂️

src/SFML/Window/Unix/VulkanImplX11.cpp
```diff
@@ -0,0 +1,228 @@
...
+        VulkanLibraryWrapper() :
+        library(NULL)
+        {
+        }
+
+        // Dtor
```
eXpl0it3r: Kind of redundant comment, but 🤷‍♀️

include/SFML/Window/Vulkan.hpp
```diff
@@ -0,0 +1,114 @@
...
+    /// If only compute is required, set \a requireGraphics
+    /// to false to skip checking for the extensions necessary
+    /// for graphics rendering.
+    ///
+    /// \param requireGraphics
+    ///
+    /// \return True if Vulkan is supported, false otherwise
+    ///
+    ////////////////////////////////////////////////////////////
+    static bool isAvailable(bool requireGraphics = true);
```
eXpl0it3r: I find `requireGraphics` a bit ambiguous in some regards. If I set it to false, will extensions still be loaded if available, because I don't *require* them, so they may still be loaded.

Maybe `requestGraphics` or `includeGraphics` or a more explicit documentation comment would work better here.

src/SFML/Window/Unix/VulkanImplX11.cpp
```diff
@@ -0,0 +1,228 @@
...
+    {
+        checked = true;
+
+        // Check if the library is available
+        computeAvailable = wrapper.loadLibrary();
+
+        // To check for instance extensions we don't need to differentiate between graphics and compute
+        graphicsAvailable = computeAvailable;
+
+        if (graphicsAvailable)
```
eXpl0it3r: If we don't require graphics, do we still need to run this block? (See also required vs request etc above)

src/SFML/Window/Vulkan.cpp
```diff
@@ -0,0 +1,95 @@
...
+#endif
+}
+
+
+////////////////////////////////////////////////////////////
+VulkanFunctionPointer Vulkan::getFunction(const char* name)
+{
+#if defined(SFML_VULKAN_IMPLEMENTATION_NOT_AVAILABLE)
+
+    return 0;
```
eXpl0it3r: Maybe use `NULL` instead?

src/SFML/Window/Win32/VulkanImplWin32.cpp
```diff
@@ -0,0 +1,222 @@
...
+#include <map>
+#include <cstring>
+
+
+namespace
+{
+    // RAII Vulkan shared library wrapper
+    struct VulkanLibraryWrapper
+    {
+        // Ctor
```
eXpl0it3r: Redundant comment?

src/SFML/Window/Win32/VulkanImplWin32.cpp
```diff
@@ -0,0 +1,222 @@
...
+        VulkanLibraryWrapper() :
+        library(NULL)
+        {
+        }
+
+        // Dtor
```
eXpl0it3r: Redundant comment?

include/SFML/Window/WindowBase.hpp
binary1248: I don't see why not... If someone finds that they are in a special situation where they need to use Vulkan and sfml-graphics simultaneously then it shouldn't be our goal to prevent something that is theoretically possible just as a matter of principle.

include/SFML/Window/Vulkan.hpp
binary1248: Our goal shouldn't be to teach people how Vulkan works. We should be able to assume that anybody who dares to touch Vulkan has some kind of idea of what they are doing.

This function doesn't actually load any extensions, because simply put, you don't _load_ extensions in Vulkan, you _require them to be enabled_ in the instance that you create. Instance creation is all or nothing, this is why you have to check beforehand if instance creation could even succeed with the required extensions enabled before going ahead and creating it.

For people who know how to work with Vulkan, the current naming/documentation should be enough for them to understand the usage.

src/SFML/Window/Unix/VulkanImplX11.cpp
binary1248: Yes, because Vulkan doesn't differentiate between graphics and non-graphics extensions. We need to make that differentiation ourselves when checking for support. We enter the block regardless what we are looking for and update the available flags accordingly.

binary1248:
Updated with requested changes.

examples/vulkan/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+
+set(SRCROOT ${PROJECT_SOURCE_DIR}/examples/vulkan)
+
+# all source files
+set(SRC ${SRCROOT}/Vulkan.cpp)
+
+# define the window target
+sfml_add_example(vulkan GUI_APP
+                 SOURCES ${SRC}
+                 DEPENDS sfml-window sfml-graphics
```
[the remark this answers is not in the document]
JonnyPtn: No, but In the past it has been specifically requested to explicitly list all dependencies

JonnyPtn:
It doesn't work on macOS, of course

examples/vulkan/Vulkan.cpp
```diff
@@ -0,0 +1,2556 @@
...
+        if (!vulkanAvailable)
+            window.setTitle("SFML window with Vulkan (Vulkan not available)");
+    }
+
+
+    // Destructor
+    ~VulkanExample()
+    {
+        // Wait until there are no pending frames
+        vkDeviceWaitIdle(device);
```
JonnyPtn: Need to check for null device

eXpl0it3r: Not sure how much error handling one wants to add, but I guess if VK isn't available this may crash, @binary1248?

binary1248:
Fixed missing check for a valid device in the destructor.

john01dav:
>     * Any real impact on requiring `dl` on Linux even when people aren't using Vulkan?
I use Linux and develop C++ here very frequently. There is no impact on requiring dl. Any system will have it as it is part of the system that loads shared libraries, even if it isn't explicitly being used.

john01dav:
> It's my opinion this is not a useful addition to SFML, as it's a platform specific renderer which isn't compatible with SFML's graphics module, and adds 16k lines of source code just to save a very niche group of users some boilerplate. Might be missing something though?

I disagree with your assertion that it is not a useful addition to SFML. Please let me explain why:
 - Vulkan is not a platform specific renderer. It has support on Windows, Linux, and [even Mac](https://en.wikipedia.org/wiki/MoltenVK). It also works on Android, although I am uncertain if that matters to SFML. In addition to this, there is [an active initiative](https://www.khronos.org/vulkan/portability-initiative) to make it available in as many places as possible. 
 - Vulkan would be an extremely useful addition to SFML to allow SFML to be used an OS-abstraction layer with a very good C++ API instead of some awful C monstrosity from the likes of SDL or native windowing APIs, which is how I use SFML. This works well with OpenGl as I can create a context with SFML and then not use SFML to render to it. It would be very useful to be able to do the same think with Vulkan, and thus would allow SFML to serve as a better OS-abstraction layer, which, to me, is its best quality.
 - Without this patch, the best alternative is to bring in some other C library with a far inferior API to that of SFML, something that is best avoided both due to the inferior API and due to the extra dependency.

eXpl0it3r:
Anyone else willing to review this?

eXpl0it3r:
Looks like there are no further reviews, so I'll merge this PR soon.

PR #1559, opened 2019-02-20
Wrong fullscreen resolution fix

### LINKED ISSUE #1349: Android - sf::RenderWindow::Create doesn't take the button drawer into account
Best described in pictures:

image_012

Over here is when the app opens, everything looks fine.

But due to the bug with the android sticky I'm manually recreating the window with sf::RenderWindow::Create on mouseEntered event, and this is the result:

image_013

It seems like sf::RenderWindow::Create doesn't take the button drawer into account.

Virtual device - Nexus5X API 26
IDE - Android Studio
NDK Version - compiled with r15c, using r16b ( When I try to compile with r16b I get a version unsupported error)
The bug persists on the pixel2 API 26

acsbendi:
## Description

A new field `fullScreenSize` has been added to the `ActivityStates`, which will hold the screen size when in full-screen mode (more precisely, immersive mode) as opposed to the default mode. This variable is set using a new function `getFullScreenSizeInPixels` which calls [`getRealMetrics`](https://developer.android.com/reference/android/view/Display#getRealMetrics(android.util.DisplayMetrics)) through JNI. The value of this new field is returned in `getFullscreenModes`. Previously, this method returned the same value as `getDesktopMode` which is incorrect, because the usable screen size in full-screen mode is always larger.

This PR will close #1349

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [X] Tested on Android

## How to test this PR?

Set the screen orientation in `AndroidManifest.xml` to landscape:
```xml
android:screenOrientation="landscape"
```

```cpp
#include <SFML/System.hpp>
#include <SFML/Window.hpp>
#include <SFML/Graphics.hpp>
#include <SFML/Audio.hpp>
#include <SFML/Network.hpp>

{
    sf::VideoMode screen(sf::VideoMode::getFullscreenModes()[0]);

    sf::RenderWindow window(screen, "", sf::Style::Fullscreen);
    window.setFramerateLimit(30);

    sf::RectangleShape rightRectangle{{50,50}}; //this should appear on the right edge of the screen
    rightRectangle.setPosition(screen.width - 50, 100);
    rightRectangle.setFillColor(sf::Color::Red);

    bool active = true;

    while (window.isOpen())
    {
        sf::Event event;

        {
            switch (event.type)
            {
                case sf::Event::Closed:
                    window.close();
                    break;
                case sf::Event::KeyPressed:
                        window.close();
                    break;
                case sf::Event::Resized:
                    window.setView(view);
                    break;
                case sf::Event::LostFocus:
                    break;
                case sf::Event::GainedFocus:
                    break;

                case sf::Event::MouseLeft:
                    active = false;
                    break;
                case sf::Event::MouseEntered:
                    active = true;
                    break;
                default:
                    break;
            }
        }

        if (active)
        {
            window.clear(background);
            window.draw(rightRectangle);
            window.display();
        }
        else {
        }
    }
}
```

### DIFF
--- a/src/SFML/Main/MainAndroid.cpp
+++ b/src/SFML/Main/MainAndroid.cpp
@@ -244,6 +244,48 @@ void getScreenSizeInPixels(ANativeActivity* activity, int* width, int* height)
 
 
 ////////////////////////////////////////////////////////////
+void getFullScreenSizeInPixels(ANativeActivity* activity, int* width, int* height)
+{
+    // Perform the following Java code:
+    //
+    // DisplayMetrics dm = new DisplayMetrics();
+    // getWindowManager().getDefaultDisplay().getRealMetrics(dm);
+
+    JNIEnv* lJNIEnv = activity->env;
+
+    jobject objectActivity = activity->clazz;
+    jclass classActivity = lJNIEnv->GetObjectClass(objectActivity);
+
+    jclass classDisplayMetrics = lJNIEnv->FindClass("android/util/DisplayMetrics");
+    jmethodID initDisplayMetrics = lJNIEnv->GetMethodID(classDisplayMetrics, "<init>", "()V");
+    jobject objectDisplayMetrics = lJNIEnv->NewObject(classDisplayMetrics, initDisplayMetrics);
+
+    jmethodID methodGetWindowManager = lJNIEnv->GetMethodID(classActivity, "getWindowManager", "()Landroid/view/WindowManager;");
+    jobject objectWindowManager = lJNIEnv->CallObjectMethod(objectActivity, methodGetWindowManager);
+
+    jclass classWindowManager = lJNIEnv->FindClass("android/view/WindowManager");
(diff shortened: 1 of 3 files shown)


eXpl0it3r:
While we do appreciate the contributions, we also expect people to not just ignore the template text, add useful information like a description and to remove unrelated text blocks.

Can you please add a description that explains the change and your reasoning behind it? 🙂

acsbendi:
@eXpl0it3r Alright, I just had the time to provide an explanation + example.

acsbendi:
@eXpl0it3r I've resolved the conflicts, fixed the build errors, and updated the description. Is there anything else I can do to get this merged?

eXpl0it3r:

eXpl0it3r (approved):
I discussed this with Zombieschannel und Discord a while back and it seems to be the way forward.

PR #1560, opened 2019-02-20
Fixed missing checks for empty vertex arrays when updating the vertex buffers in sf::Text.

### LINKED ISSUE #1558: assert in src/SFML/graphics/text.cpp
* Your OS / distro / window manager used:
Windows 7
* Your version of SFML (2.5.0, git master, etc)
git master
* Your compiler and compiler version used
VS2019

In src\SFML\Graphics\Text.cpp

                if (m_outlineVertices.getVertexCount() > 0)
Line 241 has this line:
                    m_outlineVerticesBuffer.update(&m_outlineVertices[0]);
The above line triggers an assert when m_outlineVertices is empty.
Because address of item at index 0 isn't a legal statement to make in such a situation.

I believe it needs something like this:
                if (m_outlineVertices.getVertexCount() > 0)
                    m_outlineVerticesBuffer.update(&m_outlineVertices[0]);
Or possibly a line like this:
                    m_outlineVerticesBuffer.update(
                        m_outlineVertices.getVertexCount() > 0 ? &m_outlineVertices[0] : nullptr
                    );
There are other parts of the codebase that could use similar guards.
Thanks.

binary1248:
Fixes #1558.

Test with the following in debug configuration (asserts enabled):
```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderTexture foo;
    sf::Font bar;
    sf::Text baz(" ", bar);
    foo.draw(baz); // Debug assert 1 & 2
    baz.setFillColor(sf::Color::Green);  // Debug assert 3
    baz.setOutlineColor(sf::Color::Green);  // Debug assert 4
}
```

### DIFF
--- a/src/SFML/Graphics/Text.cpp
+++ b/src/SFML/Graphics/Text.cpp
@@ -212,7 +212,8 @@ void Text::setFillColor(const Color& color)
                 if (m_verticesBuffer.getVertexCount() != m_vertices.getVertexCount())
                     m_verticesBuffer.create(m_vertices.getVertexCount());
 
-                m_verticesBuffer.update(&m_vertices[0]);
+                if (m_vertices.getVertexCount() > 0)
+                    m_verticesBuffer.update(&m_vertices[0]);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/SFML/Graphics/Text.cpp
```diff
@@ -212,7 +212,8 @@ void Text::setFillColor(const Color& color)
                 if (m_verticesBuffer.getVertexCount() != m_vertices.getVertexCount())
                     m_verticesBuffer.create(m_vertices.getVertexCount());
 
-                m_verticesBuffer.update(&m_vertices[0]);
+                if (m_vertices.getVertexCount())
```
LaurentGomila: We usually write "> 0" explicitly for this kind of tests.

binary1248: Fixed.

PR #1575, opened 2019-04-08
Fixed unable to build SFML with latest Android NDK

acsbendi:

----

## Description

SFML could not be built using the latest NDK, since CMake did not find any libraries (such as OpenAL).
This PR will fix this.
The problem was caused by the the changes added in NDK r19b. Since then, NDK adds its root path to `CMAKE_FIND_ROOT_PATH`. To tell CMake to also search for the libraries in SFML's directory, we have to add it to `CMAKE_FIND_ROOT_PATH` too.

## Tasks

* [X] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Try to build SFML using the [latest (r19c) Android NDK](https://developer.android.com/ndk/downloads). Without this PR, it will fail.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,6 +106,9 @@ endif()
 
 # Android options
 if(SFML_OS_ANDROID)
+    # avoid missing libraries when building SFML for Android with NDK r19c and later
+    set(CMAKE_FIND_ROOT_PATH "${PROJECT_SOURCE_DIR};${CMAKE_FIND_ROOT_PATH}")
+
     # make sure there's the android library available
     if (CMAKE_ANDROID_API LESS 14)
         message(FATAL_ERROR "Android API level (${CMAKE_ANDROID_API}) must be equal or greater than 14.")


auzden:
This worked for me on NDK r19c on Windows 10. Many thanks!

acsbendi:
@auzden You're welcome :)

CMakeLists.txt
```diff
@@ -44,6 +44,9 @@ endif()
 # project name
 project(SFML)
 
+# avoid missing libraries when building with latest NDK
+set(CMAKE_FIND_ROOT_PATH "${PROJECT_SOURCE_DIR};${CMAKE_FIND_ROOT_PATH}")
+
```
eXpl0it3r, line 49: As this [seems to only be](https://github.com/android/ndk/issues/912) an issue with the NDK (and potentially other cross-compilations), I'd suggest to put it inside an Android check.

acsbendi: Updated.

eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:
Thanks for this! 🙂 

Sorry it took so long 😓

acsbendi:
You're welcome!

Issue #1563, opened 2019-02-24
Use GNUInstallDirs in CMake for install flexibility


eliasdaler:
Right now, we have a lot of folders hard coded for install rules [here](https://github.com/SFML/SFML/blob/6d8143088ec8ffc31a12ee3a9a60119474407872/cmake/Macros.cmake#L176). I think that for that, we can use [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) module which sets a lot of sensible defaults which can easily be redefined as needed, e.g. if you want libraries to go to some other place than `${CMAKE_INSTALL_PREFIX}/lib`, you'll be able to set `${CMAKE_INSTALL_LIBDIR}`.

This will make lives of our *BSD folks easier, plus it will allow users of some unconventional distros to be able to configure install directories as they want.

Thoughts?

(Related to #189, #1561, #1510)

LaurentGomila:
We're happily using that at work to simplify the install configuration. It works fine for our needs on Windows/Linux/MacOS.

JonnyPtn:
Should be a pretty straightforward change, just need to clarify which variables should be used for what. So I think these are the ones we want to replace, and my suggested replacements. Once the SFML team confirm these a PR can be done pretty quickly

* `lib` replace with `CMAKE_INSTALL_LIBDIR`
* `SFML_DEPENDENCIES_INSTALL_PREFIX` : replace with `CMAKE_INSTALL_LIBDIR`
* `SFML_MISC_INSTALL_PREFIX`: replace with
    * `CMAKE_INSTALL_INFODIR` for the license/readme
    *  `CMAKE_INSTALL_DOCDIR` for the docs
* `bin`: replace with `CMAKE_INSTALL_BINDIR`

Not sure what we should do with `SFML_PKGCONFIG_INSTALL_PREFIX`?

eliasdaler:
I think for pkg-config files, it'll be good to use ```${CMAKE_INSTALL_LIBDIR}/pkgconfig``` as the default, but still make it configurable (right now, I have lots of *.pc files in /usr/local/lib/pkgconfig and in /usr/lib/...)

eXpl0it3r:
Don't forget the considerations that were taken into account for the current PKGCONFIG setup in #1433 / #1466 and others. 😉

eXpl0it3r:
Fixed with #1576

Issue #189, opened 2012-03-23
Wrong directory for example binaries 


intjelic:
Examples binaries shouldn't be placed in /usr/share (or /usr/local/share) since those directories are meant for architecture-independent data. I think /usr/lib/games is a more appropriate directory.

Source: http://www.pathname.com/fhs/pub/fhs-2.3.html#USRSHAREARCHITECTUREINDEPENDENTDATA

eXpl0it3r:
So what should be changed?

binary1248:
The option that I can think of that makes everyone happy for now is to split the install targets:
* The default install target installs the library and headers
* A separate install target installs the examples _somewhere of the user's choosing_
* Another install target installs the documentation _somewhere of the user's choosing_

I chose to separate installing the documentation as well, because let's face it, if you grab the source archive of an official release and build it, you would probably be more inclined to just go to the website than to browse your filesystem for some index.htm file. For master revisions, it's another story, but knowing what to look for (the potentially new stuff) already implies that you follow development which makes having to browse the documentation specially for that new feature irrelevant.

Maybe at some point in the future, examples can be split off into their own repository, but let's keep that discussion for another time and place.

MarioLiebisch:
Yep, sounds very reasonable to me.

eXpl0it3r:
So `CMAKE_INSTALL_PREFIX` is for libraries and headers, plus two options to activate doc/example installing as well as two paths to specify where to install them, right?
Or should building the docs/examples imply installing them?

MarioLiebisch:
Building shouldn't imply installing, e.g. you might want to keep the CHM version only, not the raw HTML files.

binary1248:
When support for separate building comes along, nothing speaks against having to separately install the products as well. We really shouldn't try to predict what the user ultimately wants to do :wink:. If they want more things to happen, let them do more for it.

mantognini:
Does someone feel like implementing it? :-)

TankOs:
AFAIK the typical situation in Linux distributions is to ship the example source code only, not binaries. That code resides in /usr/share/doc/SFML/example, for example, then.

Separating the targets makes sense in my opinion.

binary1248:
Is this still valid considering #1335 has been merged?

Bromeon:
Is this still relevant?

ChrisThrasher:
I believe I recently spoke with @eXpl0it3r who told me that examples are no longer installed meaning we can close this issue, right?

[...]

PR #1576, opened 2019-04-09
Use GNUInstallDirs module for cmake install paths

JonnyPtn:

----

## Description

This replaces our custom install paths with the ones from the cmake GnuInstallDirs module

This PR is related to the issue #1563 

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Build and install SFML (preferably to a custom CMAKE_INSTALL_PREFIX) and check the directory structure.

Seems I'm not able to request reviewers, but @eliasdaler , @fxcoudert and @Ceylo  would be great to take a look

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,11 @@ project(SFML)
 # include the configuration file
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/Config.cmake)
 
+# we use the paths from the cmake GNUInstallDirs module as defaults
+# you can override these if you like
+# https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html
+include(GNUInstallDirs)
+
 # setup version numbers
 set(VERSION_MAJOR 2)
 set(VERSION_MINOR 5)
@@ -92,14 +97,8 @@ if(SFML_OS_MACOSX)
     # add an option to build frameworks instead of dylibs (release only)
     sfml_set_option(SFML_BUILD_FRAMEWORKS FALSE BOOL "TRUE to build SFML as frameworks libraries (release only), FALSE to build according to BUILD_SHARED_LIBS")
 
-    # add an option to let the user specify a custom directory for external frameworks installation
-    sfml_set_option(SFML_DEPENDENCIES_INSTALL_PREFIX "/Library/Frameworks" PATH "External frameworks (FLAC, Freetype, Vorbis, ...) installation directory")
-
     # add an option to automatically install Xcode templates
     sfml_set_option(SFML_INSTALL_XCODE_TEMPLATES FALSE BOOL "TRUE to automatically install the Xcode templates, FALSE to do nothing about it. The templates are compatible with Xcode 4 and 5.")
-else()
-    # add an option to let the user specify a custom directory for external libraries installation
-    sfml_set_option(SFML_DEPENDENCIES_INSTALL_PREFIX "." PATH "External libraries (FLAC, Freetype, Vorbis, ...) installation directory")
 endif()
 
 # iOS specific options
@@ -125,9 +124,8 @@ if(SFML_OS_ANDROID)
     # install everything in $NDK/sources/ because this path is appended by the NDK (convenient)
     set(CMAKE_INSTALL_PREFIX ${CMAKE_ANDROID_NDK}/sources/third_party/sfml)
 
-    # we install libs in a subdirectory named after the ABI (lib/mips/*.so)
-    set(LIB_SUFFIX "/${CMAKE_ANDROID_ARCH_ABI}")
-
+    # we install libs in a subdirectory named after the ABI
+    set(CMAKE_INSTALL_LIBDIR "${CMAKE_INSTALL_LIBDIR}/${CMAKE_ANDROID_ARCH_ABI}")
     # pass shared STL configuration (if any)
     if (CMAKE_ANDROID_STL_TYPE MATCHES "_shared")
         add_definitions("-DSTL_LIBRARY=${CMAKE_ANDROID_STL_TYPE}")
@@ -158,9 +156,6 @@ elseif(SFML_OS_MACOSX)
 elseif(SFML_OS_ANDROID)
     set(DEFAULT_INSTALL_MISC_DIR ${CMAKE_ANDROID_NDK}/sources/third_party/sfml)
 endif()
-# add an option to let the user specify a custom directory for doc, examples, licence, readme and other miscellaneous files
-sfml_set_option(SFML_MISC_INSTALL_PREFIX "${DEFAULT_INSTALL_MISC_DIR}" PATH "Prefix installation path for miscellaneous files")
-
 
 # force building sfml-window, if sfml-graphics module is built
 if(SFML_BUILD_GRAPHICS AND NOT SFML_BUILD_WINDOW)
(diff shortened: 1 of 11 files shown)


eXpl0it3r:
@jcowgill can you give feedback on this, especially w.r.t. pkgconfig?

cmake/Macros.cmake
```diff
@@ -257,7 +257,7 @@ macro(sfml_add_example target)
         target_link_libraries(${target} PRIVATE ${THIS_DEPENDS})
     endif()
 
-    set(target_install_dir ${SFML_MISC_INSTALL_PREFIX}/examples/${target})
+    set(target_install_dir ${CMAKE_INSTALL_BINDIR}/examples/${target})
```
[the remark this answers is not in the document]
JonnyPtn: As discussed I've removed the examples entirely from the install target now

CMakeLists.txt
```diff
@@ -417,31 +411,31 @@ else()
             COMPONENT devel)
 endif()
 
-install(FILES license.md DESTINATION ${SFML_MISC_INSTALL_PREFIX})
-install(FILES readme.md DESTINATION ${SFML_MISC_INSTALL_PREFIX})
+install(FILES license.md DESTINATION ${CMAKE_INSTALL_INFODIR})
```
jcowgill: `${CMAKE_INSTALL_INFODIR}` (aka `/usr/share/info`) is only for [Gnu TexInfo](https://www.gnu.org/software/texinfo/) files. Use `${CMAKE_INSTALL_DOCDIR}` instead.

jcowgill:
Presumably this section of `Macros.cmake` needs updating?
```
    # add the install rule
    install(TARGETS ${target} EXPORT SFMLConfigExport
            RUNTIME DESTINATION bin COMPONENT bin
            LIBRARY DESTINATION lib${LIB_SUFFIX} COMPONENT bin
            ARCHIVE DESTINATION lib${LIB_SUFFIX} COMPONENT devel
            FRAMEWORK DESTINATION "." COMPONENT bin)
```

jcowgill:
And I expect to do this properly, all occurrences of `lib${LIB_SUFFIX}` should be replaced with `${CMAKE_INSTALL_LIBDIR}`

jcowgill:
And for pkg-config specifically, `${CMAKE_INSTALL_LIBDIR}/pkgconfig` is the correct directory to use by default (but keep the extra override for BSD systems).

CMakeLists.txt
JonnyPtn: Thanks, Done

JonnyPtn:
Thanks for the feedback @jcowgill, think I've covered them all now?

CMakeLists.txt
```diff
@@ -47,6 +47,9 @@ project(SFML)
 # include the configuration file
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/Config.cmake)
 
+# we use the gnu install dirs for our install paths
```
eliasdaler: Don't like this comment. First of all, it should be "GNU", not "gnu", plus it kinda makes it sound like these paths are set in stone and not configurable. Instead, we should somehow lead the reader to [this link](https://cmake.org/cmake/help/git-master/module/GNUInstallDirs.html) and say that all these paths are configurable.

cmake/Macros.cmake
```diff
@@ -257,44 +257,10 @@ macro(sfml_add_example target)
...
-        set_target_properties(${target} PROPERTIES
-                              INSTALL_RPATH "$ORIGIN/${rel_lib_dir}")
-    endif()
-
     if (SFML_OS_IOS)
         sfml_set_common_ios_properties(${target})
     endif()
 
-    # add the install rule
-    install(TARGETS ${target}
```
eliasdaler: Why are the examples not installed anymore?

CMakeLists.txt
JonnyPtn: Probably just adding the link is good enough? Don't think we need to go into a detailed description about cmake features?

cmake/Macros.cmake
JonnyPtn: Most people don't want them to be installed, and they commonly aren't

JonnyPtn:
Not sure how much benefit there is to us explaining these variables, and it's kind of out of SFML scope. Realistically the user should only need to worry about CMAKE_INSTALL_PREFIX, and anything else is more advanced usage which they can probably work out on their own.

I think if we add a comment linking to the cmake docs this should suffice, but I'm happy to add more if that's the decision

cmake/Macros.cmake
eliasdaler: Not sure if this be done in one go, though... @eXpl0it3r, thoughts?

CMakeLists.txt
eliasdaler: Yeah, maybe add a link and say "search for CMAKE_INSTALL_* for customizing installation directories". But maybe even link should be enough

eliasdaler:
> Realistically the user should only need to worry about CMAKE_INSTALL_PREFIX, and anything else is more advanced usage which they can probably work out on their own.

Yeah, agreed. Okay, it's good enough for now. :+1:

cmake/Macros.cmake
eXpl0it3r: I suggested to remove it, otherwise we now have to have a discussion where the examples should be installed to with the new GNU dirs, which then eventually will lead to the discussion about whether we should install them at all and then we can go round and round from there (see #189).

I'd much rather go the other way of removing the installation and let those, who feel that it's really needed, come up with a good directory suggestion. That way we won't be blocking this PR.

eliasdaler: Okay, fair enough.

CMakeLists.txt
JonnyPtn: Have added the link and mentioned that they can be set by the user (and made GNU uppercase :P)

cmake/Macros.cmake
LaurentGomila: The install target is supposed to generate a SFML package similar to the official releases. I can live with the exception of examples, but we (@eXpl0it3r) now have to remember to do the extra step of copying examples when preparing SFML releases.

eXpl0it3r:

JonnyPtn:
Have just noticed this will break the xcode templates which rely on SFML_DEPENDENCIES_INSTALL_PREFIX to copy the frameworks into the app bundle...

I can probably change it to use the generator expression for the output directory of SFML if that's suitable?

eXpl0it3r:
Sounds good to me, even though I don't fully understand the implications.

Looks like the Android build fails, do we have to adjust builbot as well?

jcowgill:
There are still a few instances of `LIB_SUFFIX` left in the pkg-config files in `tools/pkg-config/`. It's also set once in the Android code in `CMakeLists.txt`, but I'm not sure what to do about that though.

JonnyPtn:
I think that's the pkg config stuff sorted @jcowgill?

For the xcode template, even after updating the paths in the post build step it still fails to link for me, and I don't see how it would work as it doesn't set the library search paths as far as I can tell. I also haven't managed to get it working in the latest master anyway, so I'm inclined to say this is a separate issue to discuss

eXpl0it3r:

PR #1578, opened 2019-04-16
Fixed sometimes failing to read floats and doubles from Packet on Android.

### LINKED ISSUE #1565: Sometimes not able to receive floats via Packet on Android
Thanks for raising your issue here! 🙂

Before you submit the issue however, we'd like you to consider the follow points.

* We like to use the issue tracker for confirmed issues.
* If you're stuck with SFML, please use [the forum](https://en.sfml-dev.org/forums/index.php#c3) to get help.

----

## Subject of the issue

I set up a simple server-client application, the server running on desktop (Ubuntu 18.04) and the client on Android. The server after an initial message from the client sends an `sf::Packet` back to the client, which contains an `std::string` and a `float`. The client crashes when it tries to read the float after reading the string. The error message is as follows:
```2019-02-25 16:53:48.872 25082-25112/org.sfmldev.android A/libc: Fatal signal 7 (SIGBUS), code 1, fault addr 0xea02f07b in tid 25112 (sfmldev.android)```

I think the error comes from some strict requirements about the alignment of `float`s.
See [this related post](https://stackoverflow.com/questions/26158510/gcc-float-pointer-casting-in-c-causing-sigbus-error/26158641#26158641) and the [line in question](https://github.com/SFML/SFML/blob/6d8143088ec8ffc31a12ee3a9a60119474407872/src/SFML/Network/Packet.cpp#L243).
This suspicion is reinforced by the fact that the error does not always happen depending on the length of the string sent.

## Your environment

* Host enviroment: Ubuntu 18.04, client: Android 7.0 Huawei MediaPad M3
* SFML version: git master for the client (Android) and precompiled SFML 2.5.1 for Linux for the server (Ubuntu)
* I built the client (Android) SFML libraries with the following command: `cmake -DANDROID_ABI=armeabi-v7a -DANDROID_PLATFORM=android-21 -stdlib=libc++ -DCMAKE_TOOLCHAIN_FILE=${ndk_path}/build/cmake/android.toolchain.cmake -DCMAKE_ANDROID_NDK_TOOLCHAIN_VERSION=clang -DCMAKE_SYSTEM_NAME=Android -DCMAKE_ANDROID_NDK=${ndk_path} -DCMAKE_ANDROID_STL_TYPE=c++_static -DCMAKE_BUILD_TYPE=Debug ../..
`, where `ndk_path` points to the directory containing the latest NDK version (r19b).

## Steps to reproduce
Server:
```cpp
#include <iostream>
#include <cstdlib>
#include <SFML/Network/UdpSocket.hpp>
#include <SFML/Network/Socket.hpp>

int main() {
    // Choose an arbitrary port for opening sockets
    const unsigned short port = 50001;

    // Create a socket to receive a message from anyone
    sf::UdpSocket socket;

    // Listen to messages on the specified port
    if (socket.bind(port) != sf::Socket::Done)
        return EXIT_FAILURE;
    std::cout << "Server is listening to port " << port << ", waiting for a message... " << std::endl;

    // Wait for a message
    sf::Packet packet;
    sf::IpAddress sender;
    unsigned short senderPort;
    if (socket.receive(packet, sender, senderPort) != sf::Socket::Done)
        return EXIT_FAILURE;
    std::cout << "Message received from client " << sender << std::endl;

    sf::Packet packet2;
    std::string message = "message";
    packet2 << message;
    float f = 32.234f;
    packet2 << f;
    if (socket.send(packet2, sender, senderPort) != sf::Socket::Done)
        return EXIT_FAILURE;

    std::cout << "Message sent to the client: \"" << message << "\"" << std::endl;
    std::cout << "Float sent to the client: \"" << f << "\"" << std::endl;

    // Wait until the user presses 'enter' key
    std::cout << "Press enter to exit..." << std::endl;
    std::cin.ignore(10000, '\n');
    std::cin.ignore(10000, '\n');

    return EXIT_SUCCESS;
}
```
Client:
```cpp
#include <SFML/System.hpp>
#include <SFML/Window.hpp>
#include <SFML/Graphics.hpp>
#include <SFML/Audio.hpp>
#include <SFML/Network.hpp>
#include <android/log.h>

int main(int argc, char *argv[]) {

    window.setFramerateLimit(30);

    bool active = true;

    while (window.isOpen()) {
        sf::Event event;

        while (active ? window.pollEvent(event) : window.waitEvent(event)) {
            switch (event.type) {
                case sf::Event::Closed:
                    window.close();
                    break;
                case sf::Event::KeyPressed:
                        window.close();
                    break;
                case sf::Event::Resized:
                    view.setCenter(event.size.width / 2, event.size.height / 2);
                    window.setView(view);
                    break;
                case sf::Event::LostFocus:
                    break;
                case sf::Event::GainedFocus:
                    break;

                case sf::Event::MouseLeft:
                    active = false;
                    break;
                case sf::Event::MouseEntered:
                    active = true;
                    break;
                case sf::Event::TouchBegan:
                    if (event.touch.finger == 0) {
                        // Create a socket for communicating with the server
                        sf::UdpSocket socket{};
                        socket.bind(sf::Socket::AnyPort);

                        // Send a message to the server
                        sf::Packet welcomePacket;
                        std::string welcomeMessage = "Hi, I'm a client";
                        welcomePacket << welcomeMessage;
                        if (socket.send(welcomePacket, /*IP of the server*/, 50001) !=
                            sf::Socket::Done)
                        __android_log_print(ANDROID_LOG_ERROR, "Packet test", "Error while sending message! ");

                        // Receive an answer from anyone (but most likely from the server)
                        sf::Packet packet;
                        sf::IpAddress sender;
                        unsigned short senderPort;
                        if (socket.receive(packet, sender, senderPort) !=
                            sf::Socket::Done)
                            __android_log_print(ANDROID_LOG_ERROR, "Packet test", "Error while receiving message! ");

                        std::string message;
                        float f;
                        packet >> message;
                        packet >> f;
                        __android_log_print(ANDROID_LOG_DEBUG, "Packet test", "Received from server: message: %s, float: %f", message.c_str(), f);
                    }
                    break;
            }
        }

        if (active) {
            window.clear(background);
            window.display();
        } else {
        }
    }
}

```
## Expected behavior

The client application should read the `float` from the `sf::Packet`.

## Actual behavior

The client application crashes.

acsbendi:
* [X] Have you provided some example/test code for your changes? -> in the issue

----

## Description

This PR closes #1565. `reinterpret_cast` has been replaced with `memcopy` so the alignment rules for floats and doubles are not violated.

## Tasks

* [ ] Tested on Linux
* [X] Tested on Windows (server)
* [ ] Tested on macOS
* [ ] Tested on iOS
* [X] Tested on Android (client)

## How to test this PR?

The detailed test case is described in the issue that this PR will close.

### DIFF
--- a/src/SFML/Network/Packet.cpp
+++ b/src/SFML/Network/Packet.cpp
@@ -240,7 +240,7 @@ Packet& Packet::operator >>(float& data)
 {
     if (checkSize(sizeof(data)))
     {
-        data = *reinterpret_cast<const float*>(&m_data[m_readPos]);
+        std::memcpy(&data, &m_data[m_readPos], sizeof(data));
         m_readPos += sizeof(data);
     }
 
@@ -253,7 +253,7 @@ Packet& Packet::operator >>(double& data)
 {
     if (checkSize(sizeof(data)))
     {
-        data = *reinterpret_cast<const double*>(&m_data[m_readPos]);
+        std::memcpy(&data, &m_data[m_readPos], sizeof(data));
         m_readPos += sizeof(data);
     }
 
--- a/src/SFML/Network/Packet.cpp
+++ b/src/SFML/Network/Packet.cpp
@@ -116,7 +116,7 @@ Packet& Packet::operator >>(Int8& data)
 {
     if (checkSize(sizeof(data)))
     {
-        data = *reinterpret_cast<const Int8*>(&m_data[m_readPos]);
+        std::memcpy(&data, &m_data[m_readPos], sizeof(data));
         m_readPos += sizeof(data);
     }
 
@@ -129,7 +129,7 @@ Packet& Packet::operator >>(Uint8& data)
 {
     if (checkSize(sizeof(data)))
     {
-        data = *reinterpret_cast<const Uint8*>(&m_data[m_readPos]);
+        std::memcpy(&data, &m_data[m_readPos], sizeof(data));
         m_readPos += sizeof(data);
     }
 
@@ -142,7 +142,8 @@ Packet& Packet::operator >>(Int16& data)
 {
     if (checkSize(sizeof(data)))
     {
-        data = ntohs(*reinterpret_cast<const Int16*>(&m_data[m_readPos]));
+        std::memcpy(&data, &m_data[m_readPos], sizeof(data));
+        data = ntohs(data);
         m_readPos += sizeof(data);
     }
 
@@ -155,7 +156,8 @@ Packet& Packet::operator >>(Uint16& data)
 {
     if (checkSize(sizeof(data)))
     {
-        data = ntohs(*reinterpret_cast<const Uint16*>(&m_data[m_readPos]));
+        std::memcpy(&data, &m_data[m_readPos], sizeof(data));
+        data = ntohs(data);
         m_readPos += sizeof(data);
     }
 
@@ -168,7 +170,8 @@ Packet& Packet::operator >>(Int32& data)
 {
     if (checkSize(sizeof(data)))
     {
-        data = ntohl(*reinterpret_cast<const Int32*>(&m_data[m_readPos]));
+        std::memcpy(&data, &m_data[m_readPos], sizeof(data));
+        data = ntohl(data);
         m_readPos += sizeof(data);
     }
 
@@ -181,7 +184,8 @@ Packet& Packet::operator >>(Uint32& data)
 {
     if (checkSize(sizeof(data)))
     {
-        data = ntohl(*reinterpret_cast<const Uint32*>(&m_data[m_readPos]));
+        std::memcpy(&data, &m_data[m_readPos], sizeof(data));
+        data = ntohl(data);
         m_readPos += sizeof(data);
     }
 
@@ -196,7 +200,8 @@ Packet& Packet::operator >>(Int64& data)
     {
         // Since ntohll is not available everywhere, we have to convert
         // to network byte order (big endian) manually
-        const Uint8* bytes = reinterpret_cast<const Uint8*>(&m_data[m_readPos]);
+        Uint8 bytes[sizeof(data)];
+        std::memcpy(bytes, &m_data[m_readPos], sizeof(data));
         data = (static_cast<Int64>(bytes[0]) << 56) |
                (static_cast<Int64>(bytes[1]) << 48) |
                (static_cast<Int64>(bytes[2]) << 40) |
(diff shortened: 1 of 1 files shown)


binary1248:
Since you possess a platform on which you can test the alignment behaviour, did you test whether all the other supported data types work as well? Unaligned memory accesses should also cause a SIGBUS to be raised in those cases as well.

Also, this is just personal preference, but I would reword the commit message to indicate that the problem was caused by unaligned memory accesses.

acsbendi:
Yes I've tested it for some other types, such as `sf::Uint64`. It seems only the floating point data types have these strict requirements. 

I think it would be good to include some more information indeed but I think I would need to open a new PR, do you think it would be worth doing that?

eXpl0it3r:
You can also use `git commit --amend` edit the message and `git push origin <your branch> --force-with-lease` to update the PR.

binary1248:
If you ask me, since we're already at it, it would be nice if all the `reinterpret_cast`s were replaced with `memcpy`s. This way we can avoid potential future headaches on platforms with even stricter alignment requirements.

acsbendi:
@eXpl0it3r Thanks for the advice, I managed to amend the commit message in this way :)

acsbendi:
@binary1248 Yes, this is a good idea (also just for the sake of consistency), so I replaced them everywhere in a new commit.

binary1248:
I would squash the commits, the changes are similar enough that they belong together.

src/SFML/Network/Packet.cpp
```diff
@@ -219,7 +224,8 @@ Packet& Packet::operator >>(Uint64& data)
     {
         // Since ntohll is not available everywhere, we have to convert
         // to network byte order (big endian) manually
-        const Uint8* bytes = reinterpret_cast<const Uint8*>(&m_data[m_readPos]);
+        Uint8 bytes[sizeof(data)];
+        std::memcpy(bytes, &m_data[m_readPos], sizeof(data));
```
dmitry-t: There is no byte order conversion. Is Int64 really serialized not in the network byte order?

src/SFML/Network/Packet.cpp
```diff
@@ -196,7 +200,8 @@ Packet& Packet::operator >>(Int64& data)
     {
         // Since ntohll is not available everywhere, we have to convert
         // to network byte order (big endian) manually
-        const Uint8* bytes = reinterpret_cast<const Uint8*>(&m_data[m_readPos]);
```
LaurentGomila: Is this memcpy (and the one below) necessary? There should never be alignment issues with uint8.

acsbendi: They are not. I just thought it's a good idea to be consistent across all reads from `Packet`. Do you think it's better to leave `reinterpret_cast`s here?

PR #1579, opened 2019-04-24
Compiler flag is not target library

uilianries:
When passing -ObjC to apple-clang by CMake, as an option it should be passed by `target_compile_options` and not through `target_link_library`.

----

## Description

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Build SFML on OSX with `window` enabled.

    mkdir build && cd build
    cmake .. -DSFML_BUILD_WINDOW=ON -DCMAKE_VERBOSE_MAKEFILE=ON

The flag `-ObjC` must be listed when building Window objects.

### DIFF
--- a/src/SFML/Window/CMakeLists.txt
+++ b/src/SFML/Window/CMakeLists.txt
@@ -233,7 +233,7 @@ target_link_libraries(sfml-window PUBLIC sfml-system)
 
 # When static linking on macOS, we need to add this flag for objective C to work
 if ((NOT BUILD_SHARED_LIBS) AND SFML_OS_MACOSX)
-    target_link_libraries(sfml-window PRIVATE -ObjC)
+    target_compile_options(sfml-window PRIVATE -ObjC)
 endif()
 
 # find and setup usage for external libraries


LaurentGomila:
This seems to be a linker flag, so I'd say `target_link_libraries` is the right thing to use. Did you have troubles with the current version, and does your PR solves them?

uilianries:
Yes. Let me get a full log to show you.

uilianries:
Building with current master branch: [sfml-target_compiler_options.log](https://github.com/SFML/SFML/files/3114212/sfml-target_compiler_options.log)

Building with target_compile_options: [sfml-target_link_libraries.log](https://github.com/SFML/SFML/files/3114213/sfml-target_link_libraries.log)

Both builds are using the same configuration:
System Version: macOS 10.14.4
Apple LLVM version 10.0.1
Build Type: Release
Library Type: Static
Build Window: ON

When both reach 14% you will see Window files being built. However, the first log doesn't list `ObjC`, which is incorrect, because I'm building the static library and the OS is OSX.

My patch is able to fix that flag, however, a new error occur because clang is treating source input files as Objective-C.

LaurentGomila:
Didn't you mix your logs, ie. sfml-target_link_libraries.log is the current master branch? I don't see any error in that log, so please explain what exactly is wrong with the current code.

uilianries:
> Didn't you mix your logs, ie. sfml-target_link_libraries.log is the current master branch?

I'm sure the logs are correct, because I've executed the same example 2 times before to open this PR. Since master is relative, the correct SHA-1 that I'm using on master is [4043f71](https://github.com/uilianries/SFML/tree/4043f71156a7149f8d4bafe15d3ac73c440fae7b)

> so please explain what exactly is wrong with the current code.

Both logs, at line 78, are building the object Clipboard, which is related to SFML Window module:

```
# With target_link_libraries(sfml-window PRIVATE -ObjC)
cd /Users/user153599/Development/SFML/build/src/SFML/Window && \
   /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ \
   -DSFML_STATIC \
   -I/Users/user153599/Development/SFML/include \
   -I/Users/user153599/Development/SFML/src \
   -F/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/System/Library/Frameworks \
   -O3 \
   -DNDEBUG \
   -arch x86_64 \
   -isysroot \
   /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk \
   -stdlib=libc++ \
   -o CMakeFiles/sfml-window.dir/Clipboard.cpp.o \
   -c /Users/user153599/Development/SFML/src/SFML/Window/Clipboard.cpp
```

```
# With target_compile_options(sfml-window PRIVATE -ObjC)
cd /Users/user153599/Development/SFML/build/src/SFML/Window && \
   /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++  \
   -DSFML_STATIC \
   -I/Users/user153599/Development/SFML/include \
   -I/Users/user153599/Development/SFML/src \
   -F/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/System/Library/Frameworks  \
   -O3 \
   -DNDEBUG \
   -arch x86_64 \
   -isysroot \
   /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk \
   -stdlib=libc++ \
   -ObjC \
   -o CMakeFiles/sfml-window.dir/Clipboard.cpp.o \
   -c /Users/user153599/Development/SFML/src/SFML/Window/Clipboard.cpp
```

As you can see, when building sfml-window as static library on OSX, the flag `-ObjC` is not listed, however the cmake condition is:

```
if ((NOT BUILD_SHARED_LIBS) AND SFML_OS_MACOSX)
    target_link_libraries(sfml-window PRIVATE -ObjC)
endif()
```

As `-ObjC` is a compiler option and not a target, it won't be solved by CMake at that moment. When applying that flag by `target_compile_options`, it will be listed for every object in sfml-window. That's my point.

However, using `target_compile_options` we have a new error, which I don't know how to fix, or even is it expected.

LaurentGomila:
> I'm sure the logs are correct, because I've executed the same example 2 times before to open this PR

You say:
> Building with current master branch: sfml-target_compiler_options.log
Building with target_compile_options: sfml-target_link_libraries.log

But obviously the current master should be sfml-target_link_libraries.log and "with target_compile_options" should be sfml-target_compiler_options.log. So what's wrong, file contents or just file names?

> As you can see, when building sfml-window as static library on OSX, the flag -ObjC is not listed

That's not an error in itself. I'd like to see a compiler/linker/runtime error, if you have one.

You don't see the flag in case of static building, because the linker is not involved when creating static libraries. 

And just to be sure: `target_link_libraries` is also used to specify *linker* flags, not just targets to link. `target_compile_options` is used for *compiler* flags.

uilianries:
> But obviously the current master should be sfml-target_link_libraries.log and "with target_compile_options" should be sfml-target_compiler_options.log. So what's wrong, file contents or just file names?

The file names, but I just fixed now. The order is correct on the same comment.

> You don't see the flag in case of static building, because the linker is not involved when creating static libraries. It is invoked only when creating the final binary.

Understood. Maybe it could be *PUBLIC* instead? When linking to sfml-window only your application will receive that compiler option.

> Please start by the initial issue that you had

This problem came from another [issue](https://github.com/bincrafters/community/issues/736#issue-435531980). When linking an application to sfml-window as static library, the flag `-ObjC` is not listed on apple clang. As that option is PRIVATE, it won't be propagated.

> You don't see the flag in case of static building, because the linker is not involved when creating static libraries. It is invoked only when creating the final binary.

So I think `target_link_libraries(sfml-window PUBLIC -ObjC)` should be enough to solve this case. WDYT?

LaurentGomila:
> So I think `target_link_libraries(sfml-window PUBLIC -ObjC)` should be enough to solve this case. WDYT?

It looks like the right solution for this specific problem, but then everything linked to sfml-window will get that linker flag too (even when not building static libs). And I don't know exactly what it involves.

I think we need the point of view of a macOS guy (@mantognini).

eXpl0it3r:
Or @JonnyPtn

JonnyPtn:
Official documentation on this is pretty scarce, but just running some quick tests using the window example, it works if -ObjC is defined in _either_ the window module or the window example. Personally I'd say the simplest solution is to keep it in the window module, as the it's useless without it. The only downside as I understand it is a small increase in binary size.

I'd say the correct cmake method to use is actually `target_link_options`, but unfortunately that's not available until cmake version 3.13 and I guess it's safe to assume increasing to that version isn't an option? Whether to use `target_compile_options` or `target_link_libraries` is just a matter of which type of wrong you want it to be - I don't really have an opinion on that.

JonnyPtn:
Personally I would reject this for the following reasons:
- I don't see a reason to make the flag public, Although it helps you in your specific scenario, I don't think it's something every user will want so we shouldn't force it on them
- I don't think changing it to `target_compile_options` makes any practical difference, and is just semantics
- Having solved a related issue on another project, there are actually different requirements when using Xcode and Makefile generators. More recent versions of cmake have specific solutions for this which we can look at when we come to update our cmake version
- It's not exactly clear what problem this fixes, and OP mentions that it actually introduces a different issue (and this is proved in the travis checks)

PR #1584, opened 2019-05-04
Remove VBO implementations from drawables

binary1248:
Having the VBO code split over all drawables increases maintenance burden over having it in one centralized place. The latter also allows for better global optimization within a rendering backend that is able to do so.

Modular rendering backend support will come around when it is ready™, and is also a pre-requisite for supporting platforms that are not able to render using the old methodology, so it makes more sense to manage vertex data storage within the renderer than each drawable. This also helps to keep the drawable concept as platform independent as possible.

This will also fix the performance regressions reported by some users who make use of a large number of ephemeral drawable objects.

### DIFF
--- a/include/SFML/Graphics/Shape.hpp
+++ b/include/SFML/Graphics/Shape.hpp
@@ -32,7 +32,6 @@
 #include <SFML/Graphics/Drawable.hpp>
 #include <SFML/Graphics/Transformable.hpp>
 #include <SFML/Graphics/VertexArray.hpp>
-#include <SFML/Graphics/VertexBuffer.hpp>
 #include <SFML/System/Vector2.hpp>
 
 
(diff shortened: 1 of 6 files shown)


Bromeon:
The feature would then be removed indefinitely (as there's no timeline for modular rendering backend) -- should this not be done once it's ready?

Or would you say that VBOs at the current stage have overall more drawbacks than advantages?

binary1248:
This change was written long before the idea of modular rendering backends was even considered, as such this belongs to a completely different school of thought than the current "future-proofing concept" if you can even call it that.

The problem now is that it is hard to gauge how many people would miss this if it was removed because nobody really provided any positive feedback on it, that is if anybody even knew this change was made. There have been reports by people suffering from performance regressions due to this change. So if you ask me, I would say remove it now already. The performance before this change was merged seemed to be good enough for most people that they didn't notice any difference when it got merged.

Modular rendering backend is currently more or less ready to be pushed on my system. I got sidetracked often in the last few months and always had second thoughts on what the API should look like, but I guess there is only one way to get feedback from the community. I will probably push a branch in the coming days if all goes to plan.

Bromeon:
Recently I updated my SFML copy and played around with it again -- I haven't noticed any VBO impact (neither positive nor negative), but those were also no high-performance scenarios.

But great to hear you have already something in the pipeline 🙂
In that case I'm fine with removing VBOs for a while. This should not break any behavior, right?

binary1248:
This is purely an internal implementation detail. There was/is no change in external behaviour (if you disregard the performance regressions that is).

PR #1590, opened 2019-06-03
Fixed sensors stop receiving data after window is recreated on Android.

acsbendi:

----

## Description

Sensors on Android stopped receiving data after the SFML window was destroyed and recreated. This is because static variables are not recreated in SFML's library. This has been described in [this answer](https://stackoverflow.com/a/3999709/3738870) as well. 

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [X] Tested on Android

## How to test this PR?

We need another Android activity that will start SFML's activity (`NativeActivity`), for example on a button press. After we have tested that the sensors work in the `NativeActivity` (in the example, we have to touch the screen to view one of its current data), we have to press the back button, this will lead us back to the other activity. When we restart the SFML activity from here again, the sensor will not change its data anymore.

`app/src/main/AndroidManifest.xml`:
```xml
<?xml version="1.0" encoding="utf-8"?>
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
    package="org.sfmldev.android"
    android:versionCode="1"
    android:versionName="1.0">

    <uses-feature android:glEsVersion="0x00010001" />

    <uses-permission android:name="android.permission.VIBRATE" />

    <application
        android:allowBackup="true"
        android:icon="@drawable/sfml_logo"
        android:label="@string/app_name">

        <activity
            android:name="android.app.NativeActivity"
            android:configChanges="keyboardHidden|orientation|screenSize"
            android:screenOrientation="landscape"
            android:icon="@drawable/sfml_logo"
            android:label="@string/app_name">
            <meta-data
                android:name="android.app.lib_name"
                android:value="sfml-activity-d" />
            <meta-data
                android:name="sfml.app.lib_name"
                android:value="sfml-example" />
        </activity>

        <activity android:name=".MainActivity"
            android:theme="@style/Theme.AppCompat.Light.DarkActionBar">
            <intent-filter>
                <action android:name="android.intent.action.MAIN" />

                <category android:name="android.intent.category.LAUNCHER" />
            </intent-filter>
        </activity>
    </application>

</manifest>
```

`app/src/main/java/org/sfmldev/android/MainActivity.java`:
```java
package org.sfmldev.android;

import android.app.NativeActivity;
import android.content.Intent;
import android.support.v7.app.AppCompatActivity;
import android.os.Bundle;
import android.view.View;
import android.widget.Button;

public class MainActivity extends AppCompatActivity {

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_main);

        Button button = findViewById(R.id.button);
        button.setOnClickListener(new View.OnClickListener() {
            @Override
            public void onClick(View view) {
                Intent intent = new Intent(MainActivity.this, NativeActivity.class);
                startActivity(intent);
            }
        });
    }
}
```

`app/src/main/jni/main.cpp`:
```cpp
#include <SFML/System.hpp>
#include <SFML/Window.hpp>
#include <SFML/Graphics.hpp>
#include <SFML/Audio.hpp>
#include <SFML/Network.hpp>
#include <SFML/Window/Sensor.hpp>

{

    window.setFramerateLimit(30);

    sf::Font font;
        return EXIT_FAILURE;

    bool active = true;

    sf::Sensor::setEnabled(sf::Sensor::Type::Orientation, true);

    sf::Text text(std::to_string(sf::Sensor::getValue(sf::Sensor::Type::Orientation).x), font, 64);
    text.setPosition(100, 100);

    while (window.isOpen())
    {
        sf::Event event;

        {
            switch (event.type)
            {
                case sf::Event::Closed:
                    window.close();
                    break;
                case sf::Event::KeyPressed:
                        window.close();
                    break;
                case sf::Event::Resized:
                    window.setView(view);
                    break;
                case sf::Event::LostFocus:
                    break;
                case sf::Event::GainedFocus:
                    break;

                case sf::Event::MouseLeft:
                    active = false;
                    break;
                case sf::Event::MouseEntered:
                    active = true;
                    break;
                case sf::Event::TouchBegan:
                    if (event.touch.finger == 0)
                    {
                        text = sf::Text(std::to_string(sf::Sensor::getValue(sf::Sensor::Type::Orientation).x), font, 64);
                        text.setPosition(100, 100);
                    }
                    break;
            }
        }

        if (active)
        {
            window.clear(background);
            window.draw(text);
            window.display();
        }
        else {
        }
    }
}
```

### DIFF
--- a/src/SFML/Window/SensorManager.cpp
+++ b/src/SFML/Window/SensorManager.cpp
@@ -33,11 +33,22 @@ namespace sf
 {
 namespace priv
 {
+SensorManager* SensorManager::singleInstance = NULL;
+
 ////////////////////////////////////////////////////////////
 SensorManager& SensorManager::getInstance()
 {
-    static SensorManager instance;
-    return instance;
+    if(singleInstance == NULL)
+        singleInstance = new SensorManager();
+    return *singleInstance;
+}
+
+
+////////////////////////////////////////////////////////////
+void SensorManager::destroyInstance()
+{
+    delete singleInstance;
+    singleInstance = NULL;
 }
 
 
--- a/src/SFML/Window/SensorManager.hpp
+++ b/src/SFML/Window/SensorManager.hpp
@@ -48,10 +48,19 @@ public:
     ////////////////////////////////////////////////////////////
     /// \brief Get the global unique instance of the manager
     ///
+    /// If needed, this function will recreate the manager's
+    /// instance.
+    ///
     /// \return Unique instance of the sensor manager
     ///
     ////////////////////////////////////////////////////////////
     static SensorManager& getInstance();
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Destroys the global unique instance of the manager
+    ///
+    ////////////////////////////////////////////////////////////
+    static void destroyInstance();
 
     ////////////////////////////////////////////////////////////
     /// \brief Check if a sensor is available on the underlying platform
@@ -128,6 +137,7 @@ private:
     // Member data
     ////////////////////////////////////////////////////////////
     Item m_sensors[Sensor::Count]; ///< Sensors information and state
+    static SensorManager* singleInstance; ///< The single instance of the manager
 };
 
 } // namespace priv
--- a/include/SFML/Window/Sensor.hpp
+++ b/include/SFML/Window/Sensor.hpp
@@ -93,6 +93,19 @@ public:
     ///
     ////////////////////////////////////////////////////////////
     static Vector3f getValue(Type sensor);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Cleans up all resources associated with sensors
+    ///
+    /// Some platforms (e.g. Android) require that when the window
+    /// (Activity) is recreated, the sensor's resources are
+    /// recreated too.
+    /// You do not need to call this function manually, SFML
+    /// will automatically do so when the window is destroyed.
+    /// For more information, see <a href="https://stackoverflow.com/a/3999709/3738870">this answer</a>.
+    ///
+    ////////////////////////////////////////////////////////////
+    static void cleanUp();
 };
 
 } // namespace sf
--- a/src/SFML/Main/MainAndroid.cpp
+++ b/src/SFML/Main/MainAndroid.cpp
@@ -43,6 +43,7 @@
 #include <SFML/System/Thread.hpp>
 #include <SFML/System/Lock.hpp>
 #include <SFML/System/Err.hpp>
+#include <SFML/Window/Sensor.hpp>
 #include <android/window.h>
 #include <android/native_activity.h>
 #include <cstring>
@@ -114,6 +115,7 @@ static void terminateMain(ActivityStates* states)
 
     // The main thread has finished, we must explicitly ask the activity to finish
     states->mainOver = true;
+    Sensor::cleanUp();
     ANativeActivity_finish(states->activity);
 }
 
--- a/src/SFML/Window/Sensor.cpp
+++ b/src/SFML/Window/Sensor.cpp
@@ -50,4 +50,10 @@ Vector3f Sensor::getValue(Type sensor)
     return priv::SensorManager::getInstance().getValue(sensor);
 }
 
+////////////////////////////////////////////////////////////
+void Sensor::cleanUp()
+{
+    priv::SensorManager::destroyInstance();
+}
+
 } // namespace sf


eXpl0it3r:
Can this also be done without extending the public API? 
If we have a "clean up" function, then it goes from a stateless to a stateful class and as such should probably be used as instance instead of static function calls and use the destructor.

acsbendi:
@eXpl0it3r I don't think it's possible in a non-hacky way. `priv::SensorManager` cannot be accessed from MainAndroid.cpp.
Yes, it should be stateful, just like how `SensorManager` is stateful too.

acsbendi:
@eXpl0it3r I've resolved the conflicts, is there anything else I can do to get this merged?

eXpl0it3r:

ChrisThrasher:
If you're still interested in maintaining this PR, can you rebase on the `master` branch and then run clang-format to fix the formatting? Once you have clang-format-12 or newer installed, build the `format` target which will automatically run a script that formats with all the correct settings. Then you can amend your commit to include the fixed formatting and force push to update this PR.

acsbendi:
@ChrisThrasher I can do that, but is there anything else that's preventing a merge? I brought the PR up-to-date with master back in June already and it wasn't merged back then.

ChrisThrasher:
> but is there anything else that's preventing a merge?

I'm not sure. I was just pointing out two particular things that need to be addressed before it can be merged. Beyond that it's up to the maintainers.

src/SFML/Window/SensorManager.hpp
```diff
@@ -136,7 +145,8 @@ class SensorManager
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    Item m_sensors[Sensor::Count]; //!< Sensors information and state
+    Item m_sensors[Sensor::Count]; ///< Sensors information and state
+    static SensorManager* singleInstance; ///< The single instance of the manager
```
ChrisThrasher, line 149: Why not use `unique_ptr` to avoid the `new` and `delete` calls? Also, why not put this static variable in SensorManager.cpp since that's the only file which needs it?

src/SFML/Window/SensorManager.cpp
```diff
@@ -35,11 +35,22 @@ namespace sf
 {
 namespace priv
 {
+SensorManager* SensorManager::singleInstance = NULL;
```
ChrisThrasher, line 38:
```suggestion
SensorManager* SensorManager::singleInstance = nullptr;
```

acsbendi:
@ChrisThrasher Okay, I was asking because I think the code is already reviewable and I prefer not spending the time fixing the formatting/rebasing before knowing if there's anything else that's blocking it from being merged.

src/SFML/Window/SensorManager.hpp
acsbendi: I guess the original reason is that back when I wrote this, SFML didn't support C++11. So probably we can switch now.

To me, it just felt like it belongs to the class. Would it work outside of the class in the same way?

ChrisThrasher: > I guess the original reason is that back when I wrote this, SFML didn't support C++11. So probably we can switch now.

I think you're right. In 2019 this wasn't an option for you since that was before SFML 3 was being developed.

> To me, it just felt like it belongs to the class. Would it work outside of the class in the same way?

As far as I'm aware this would work just fine as a global in the .cpp file. It's worth testing though in case I'm missing something.

ChrisThrasher:
> @ChrisThrasher Okay, I was asking because I think the code is already reviewable and I prefer not spending the time fixing the formatting/rebasing before knowing if there's anything else that's blocking it from being merged.

Rebasing and fixing formatting should only take a minute or two.

acsbendi:
Rebasing yes, but I don't have clang-format-12 as far as I know

src/SFML/Window/SensorManager.hpp
acsbendi: Alright, I'll do both changes once I know if the PR is good to go otherwise the changes you mentioned.

kimci86:
> is there anything else that's preventing a merge?

As far as I understand, one blocking aspect is adding the `Sensor::cleanUp` method to SFML's public interface when it is intended to be an implementation detail.

> priv::SensorManager cannot be accessed from MainAndroid.cpp.

That is technically possible. I gave it a try to see if CI would be happy with it and it looks ok. See [this branch](https://github.com/kimci86/SFML/tree/bugfix/android_sensor_recreate).
That might be considered dirty because of namespace `sf::priv`, but `MainAndroid.cpp` functions are also in namespace `sf::priv`, so maybe that would be acceptable. What do you think? Is there a cleaner alternative?

----

Other remark:
I am not familiar with Android subtleties and not sure how to test but I am wondering: Don't users need to call `Sensor::setEnabled` again manually for the sensors they use after the window is recreated? If so, couldn't we recreate the sensor manager with the same enabled sensors?

----

Regarding formatting, if you don't want to setup clang-format, you can check the diffs in the CI log to see how clang-format changes the code. For example: https://github.com/SFML/SFML/runs/7841275748?check_suite_focus=true

acsbendi:
@kimci86 Thanks for the tips, I've updated the PR accordingly.

acsbendi:
`unique_ptr` doesn't work well here because `SensorManager` has a private destructor. While we [could make it work](https://stackoverflow.com/questions/19274058/singleton-with-private-destructor-using-stdunique-ptr), the solution is quite hacky and it's way simpler to just use a pointer instead so I chose not to use `unique_ptr` here.

PR #1592, opened 2019-06-10
Replacing switch statement by bit manipulation

kocourOggy:

----

## Description

I replaced switch statement by bit manipulation in RectangleShape::getPoint(std::size_t index) method.
Returned sf::Vector2f corresponds with function input (size_t index) in the following way:
```
input = 0 = 0000  ==> first bit=0, second bit=0, XOR(0, 0)=0  ==> (XOR(0, 0), second bit) = (0, 0) ==> output = Vectof2f(0 * m_size.x, 0 * m_size.y)
input = 1 = 0001  ==> first bit=1, second bit=0, XOR(1, 0)=1  ==> (XOR(1, 0), second bit) = (1, 0) ==> output = Vectof2f(1 * m_size.x, 0 * m_size.y)
input = 2 = 0010  ==> first bit=0, second bit=1, XOR(0, 1)=1  ==> (XOR(0, 1), second bit) = (1, 1) ==> output = Vectof2f(1 * m_size.x, 1 * m_size.y)
input = 3 = 0011  ==> first bit=1, second bit=1, XOR(1, 1)=0  ==> (XOR(1, 1), second bit) = (0, 1) ==> output = Vectof2f(0 * m_size.x, 1 * m_size.y)
```
Should we do bit manipulation instead of switch statement? I know that replacing every switch statement with unclear code is unreasonable. However if the function is called quite often (e.g. because of some computation inside game loop), there can be fairly good performance gain. And I can imagine RectangleShape::getPoint that is called inside game loop oftentimes.

Function without switch statement provides stable and fast performance that does not rely on branch prediction of processor. The function also generate less assembly code and thus it can provide better opportunity for function inlining and further compiler optimizations.

More information can be found [here](https://github.com/stepaiv3/Optimization-of-switch-statement).

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

We create safe function with switch statement that is equivalent of RectangleShape::getPoint(std::size_t index). The new implemented method RectangleShape::getPoint(std::size_t index) with bit manipulation is tested in following manner. For input 0, 1, 2, 3 we test that new implemented method produce the same Vector2f object as the older one. We do not care about inputs out of range [0, 3] because SFML documentation is already saying: "The result is undefined if index is out of the valid range."

```cpp
#include <SFML/Graphics.hpp>

#include <iostream>
#include <iomanip>
#include <random>
#include <chrono>
#include <vector>

bool cmp(const sf::Vector2f& v1, const sf::Vector2f& v2)
{
    return (v1.x == v2.x) && (v1.y == v2.y);
}

void print(const sf::Vector2f& v)
{
    std::cout << std::setprecision(10);
    std::cout << "(" << v.x << ", " << v.y << ")\n";
}

// function that is equivalent of method sf::RectangleShape.getPoint() with switch statement
sf::Vector2f getPointWithSwitchFunction(const sf::RectangleShape& shape, std::size_t index)
{
    const sf::Vector2f& m_size = shape.getSize();
    switch (index)
    {
        default:
        case 0: return sf::Vector2f(0, 0);
        case 1: return sf::Vector2f(m_size.x, 0);
        case 2: return sf::Vector2f(m_size.x, m_size.y);
        case 3: return sf::Vector2f(0, m_size.y);
    }
}

// function that is equivalent of new method sf::RectangleShape.getPoint() with bit manipulation
sf::Vector2f getPointWithoutSwitchFunction(const sf::RectangleShape& shape, std::size_t index)
{
    const sf::Vector2f& m_size = shape.getSize();
    std::size_t firstBit = index & 1;
    std::size_t secondBit = index >> 1;
    firstBit = firstBit^secondBit;

    return sf::Vector2f(m_size.x * firstBit, m_size.y * secondBit);
}

bool test(const std::size_t numberOfIterations, const float from, const float to)
{
    const std::size_t seed = std::chrono::system_clock::now().time_since_epoch().count();
    std::mt19937 randomGenerator(seed);
    std::uniform_real_distribution<float> randomFloat(from, to);


    sf::Vector2f shapeSize;
    for (std::size_t i = 0; i < numberOfIterations; ++i)
    {
        std::cout << "ITERATION: " << i << "\n";
        const float f1 = randomFloat(randomGenerator);
        const float f2 = randomFloat(randomGenerator);
        const sf::Vector2f shapeSize(f1, f2);
        const sf::RectangleShape rect(shapeSize);

        std::cout << "=================================================\n";
        std::cout << "testing rectangleShape with size: "; print(shapeSize);
        for (std::size_t j = 0; j < 4; ++j)
        {
            // function using switch that is equivalent to previous method getPoint() with switch
            const sf::Vector2f correctPoint = getPointWithSwitchFunction(rect, j);

            // getPoint() new method implemented without switch
            const sf::Vector2f testedPoint = rect.getPoint(j);

            // if new method getPoint() without switch has not been implemented yet use following function
            //const sf::Vector2f testedPoint = getPointWithoutSwitchFunction(rect, j);

            std::cout << "index: " << j << "\n";
            std::cout << "correct point: "; print(correctPoint);
            std::cout << "tested  point: ";print(testedPoint);
            if ( cmp(correctPoint, testedPoint) == false )
            {
                std::cout << "Points are NOT equal!!!\n";
                std::cout << "------------------------------\n";
                return false;
            } else {
                std::cout << "Points are equal.\n";
                std::cout << "------------------------------\n";
            }
        }
        std::cout << "=================================================\n";
    }
    return true;
}

int main()
{
    const std::size_t iterations = 2 * 10000;
    const float testingFrom = -10;
    const float testingTo = 10;

    if ( test(iterations, testingFrom, testingTo) )
    {
        std::cout << "\nTest completed succesfully!\n";
    } else {
        std::cout << "\nTest failed!!! \n";
    }


    return 0;
}
```

### DIFF
--- a/src/SFML/Graphics/RectangleShape.cpp
+++ b/src/SFML/Graphics/RectangleShape.cpp
@@ -63,14 +63,11 @@ std::size_t RectangleShape::getPointCount() const
 ////////////////////////////////////////////////////////////
 Vector2f RectangleShape::getPoint(std::size_t index) const
 {
-    switch (index)
-    {
-        default:
-        case 0: return Vector2f(0, 0);
-        case 1: return Vector2f(m_size.x, 0);
-        case 2: return Vector2f(m_size.x, m_size.y);
-        case 3: return Vector2f(0, m_size.y);
-    }
+    std::size_t firstBit = index & 1;
+    std::size_t secondBit = index >> 1;
+    firstBit = firstBit^secondBit;
+
+    return sf::Vector2f(m_size.x * firstBit, m_size.y * secondBit);
 }
 
 } // namespace sf


LaurentGomila:
Thanks for the optimization trick, and for the complete analysis and benchmark of your modification.

However I'm not a big fan of such optimizations in non-critical functions, it makes the code much less readable and maintainable. The function is not called that often, since sf::Shape caches its geometry and, in the case of sf::RectangleShape, recomputes it only when size or outline thickness changes.

Let's see what others think about it.

Bromeon:
Your benchmark is a good start -- thanks for taking the time to measure it and not optimize blindly. And it's definitely a nice trick to know.

However I agree with Laurent w.r.t. code clarity. To justify optimizations which make code considerably harder to understand, I would expect some measurements indicating that this part of the code is a bottleneck _in a typical SFML application_, and is relevant _at all_ in the bigger context. There are probably dozens of cases where similar or even more promising optimizations are possible.

After having to do with some other languages, it's also interesting to see the C++ community mindset of pushing the last processor cycle out of an executable, while other languages ecourage a code style of mindless copies of entire collections for the sake of functional code or developer productivity 🙂

kocourOggy:
My urge to optimize it is there because I wouldn't expect there is a switch statement inside RectangleShape::getPoint method. I know it is more natural to write it with branch when I look inside the library. But at first glance I wouldn't expect "heavy" switch from lightweight operation as RectangleShape::getPoint from user point of view.

For example I am OK with switches in

```
bool CursorImpl::loadFromSystem(Cursor::Type type)
{
    ...
    switch (type)
    {
        default: return false;

        case Cursor::Arrow:          shape = XC_arrow;              break;
        case Cursor::Wait:           shape = XC_watch;              break;
        case Cursor::Text:           shape = XC_xterm;              break;
        case Cursor::Hand:           shape = XC_hand1;              break;
        ...
    }
    ...
}
```

```
std::string Http::Request::prepare() const
{
	...
	switch (m_method)
	{
	        case Get:    method = "GET";    break;
	        case Post:   method = "POST";   break;
	        case Head:   method = "HEAD";   break;
	        case Put:    method = "PUT";    break;
	        case Delete: method = "DELETE"; break;
	}
	...
}
```
I understand it feels more natural to me to change getPoint method because I come up with the bit manipulation. If I was on the other side, I would probably not like the change as well.

I am sure we can agree that the best solution would be the readable switch statement working as bit manipulation under the hood. But only way I can think of is commenting switch statement to see directly what is the bit manipulation doing instead of just deleting the switch completely.

However I understand if you don't like it and you are definitely right that there are certainly better part of code for speed up improvements.

Yeah... I think it's a disease of C/C++ community to optimize at all costs :D

Hapaxia:
One thing to note is that this does not take into consideration values that are out-of-range.

For example, if "4" is passed to the original code, the result will be (0, 0). In the current bit manipulation code, an input of 4 would result in a double-height rectangle's point.

Easily fixed, though, of course: `std::size_t secondBit = (index >> 1) & 1;`
That is, if this gets chosen.

Bromeon:
My take on this is: incorrect invocation of a function should be caught and not "rectified". Such corrections almost never do what the user intends to (why should index 0 be the correct one, when 732392 is passed), but they do an outstanding job at hiding and propagating logic errors.

In my opinion, `assert(index <= 3);` would be the right choice here.

kocourOggy:
This is a nice fix! On the other hand the documentation is already saying: "Index of the point to get, in range [0 .. 3]. The result is undefined if index is out of the valid range."

Thus there's no need to fix it for values such as 4, 5, etc. The documentation of SFML doesn't guarantee a default returning value for invalid inputs. It's a programmer's mistake when he relies on undefined result. It's also very unusual when programmer is trying to get the 5th point of a rectangle although rectangle consists of only 4 points.

Another point of view could be that your fix is bringing one extra instruction that CPU has to execute. And as I said the fix is unnecessary according to the SFML's documentation.

But it is a still nice fix if we want to keep closer to the original code :)

Bromeon:
> But it is a still nice fix if we want to keep closer to the original code :)

This is exactly why undefined behavior as a concept exists: to allow breaking changes for values outside the valid input domain. We should not put in effort to keep bugs compatible 🙂

Hapaxia:
I only put forward that "fix" as a way to match the original implementation as that one returned (0, 0) if out-of-range ("default").

In later versions of SFML, I expect this will be caught and trigger an exception.

It's true, though, sure, that it doesn't matter what it returns if the input is valid if it's considered "undefined behaviour". That said, the bit manipulation version without check allows an odd hacky result that might be beneficial to someone with creativity!

eXpl0it3r:
Unfortunately, I too have to agree, that the change is a lot less readable and maintainable.

I do understand the arguments and see that switch statements might not be the best option, however for non-critical code paths, it's way more important that as many people as possible can read and understand the code with ease, than to have the best performance.

Maybe if you had a profiler report of a real application (not just benchmark) and the function was shown as performance issue, I could be convinced, but I'm confident that profiler reports won't ever show that.

I feel bad for closing the PR and have your work not being carried forward, but I hope you understand the reasoning and I'm happy to see more PRs otherwise! 🙂

PR #1594, opened 2019-06-24
Clarify image documentation

Hapaxia:
Modification of the documentation for sf::Image's getPixelsPtr() method to clarify that modifying the data within the image object does not cause the pointer to be invalidated.

No code was modified - just documentation.


LaurentGomila:
Your comment is wrong, the pointer *may* be invalidated (`create`, `loadFrom...`).

And it seems a bit overkill to use 3 commits for 2 modified lines of code, you should squash into a single clean commit before submitting your PR :wink:

Bromeon:
Since this comes probably from the recent forum thread, I'll repost [what I wrote there](https://en.sfml-dev.org/forums/index.php?topic=25472.msg168472#msg168472):

>>     Single pixels are allowed to change, and only resize/reload is a problem, right?
> With the current implementation, yes. The thing is that the pointer is not meant to be kept around as a read model for the sf::Image data, but rather requested and immediately used.

As such, I would be careful with documenting too precisely the behavior of the _current_ implementation, even when it's not likely to change.

Hapaxia:
Ah, so if it is this explicit then it should probably say "modifies the data directly" or "modifies the pixel data directly".

If this is incorrect or the behaviour is expected to change in future, it can be left as it is - no problem.

I was having some trouble with GitHub and it was showing all sorts of stuff that wasn't actually there (and vice versa) so I thought I'd managed to get this to be just that single commit. My problem was from a previous modification and PR I created but did it on the master branch instead of a separate one so when it was merged elsewhere, they didn't match and I was having trouble getting them to match.

eXpl0it3r:
I think the current documentation is sufficient. The 'may' indicates that it doesn't apply to all cases and since the idea of the pointer is, as @Bromeon said, that it's for immediate usage, it doesn't really matter how the data is modified. You're really not losing anything by calling getPixelsPtr() again in case you do any modification to the image.

Hapaxia:
I understand that this pull request is not 'clean' enough.
If you would like me to still make the changes to the documentation, please let me know and I can start a new, cleaner, 'proper' pull request (instead of this one unless this one can still be fixed somehow?) with whatever you would prefer the changes to be.
However, if you would like no changes to the documentation, please let me know that too so I can close this PR and also clean up the repository for further use.

PR #1600, opened 2019-07-12
Fixed Win32 requestFocus in multithreaded environment

hobby8:
Minor fix in Win32 implementation of requestFocus() so that it works when the process creates multiple windows from different threads, which seems to be the original intention.

### DIFF
--- a/src/SFML/Window/Win32/WindowImplWin32.cpp
+++ b/src/SFML/Window/Win32/WindowImplWin32.cpp
@@ -431,8 +431,9 @@ void WindowImplWin32::setKeyRepeatEnabled(bool enabled)
 void WindowImplWin32::requestFocus()
 {
     // Allow focus stealing only within the same process; compare PIDs of current and foreground window
-    DWORD thisPid       = GetWindowThreadProcessId(m_handle, NULL);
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
I don't really understand what this fixes exactly. Reading the [documentation](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getwindowthreadprocessid) this seems equivalent.

> Return Value
> Type: Type: DWORD
>
> The return value is the identifier of the thread that created the window.

> lpdwProcessId
> Type: LPDWORD
>
> A pointer to a variable that receives the process identifier. **If this parameter is not NULL, GetWindowThreadProcessId copies the identifier of the process to the variable;** otherwise, it does not.

So whether you capture the value from the return type or view parameter as pointer, doesn't really seem to make a difference, or am I missing something here?

hobby8:
The return value of GetWindowThreadProcessId() is the _thread_ID_, not the process ID. So if you create two windows from two different threads inside the same process, you will not be able to move focus between them.
From the comment in the code, and the naming of the variables, it appears that the process ID is what's required.

This WinAPI is a bit misleading...

eXpl0it3r:
Ah thread vs process, important detail 🙂

PR #1604, opened 2019-07-19
Updating window contents during resizes - Win32

hobby8:
This feature adds an optional interface that enables the user to update the window contents during resize (as well as other scenarios where the event loop does not run), without the need for multithreading.

Implemented for Windows, other platforms should degrade gracefully.

It seems to work nicely, but is currently intended for review of the interface and the design (e.g. it lacks proper documentation, possibly new files etc.)

Interface:
Added `Window::setWindowCallbacks()` for providing an optional `WindowCallbacks` interface, which the user will get asynchronous rendering requests through during resizes.

On each rendering request, a resize `sf::Event` will be supplied if the size has changed. Resize events received through the callback will not be received again by the normal event loop.

Implementation:
`WindowImpl` notifies `WindowBase` upon asynchronous resizes through a new internal `Resizeable` interface.


eXpl0it3r:
SFML currently doesn't use callbacks, as such the introduction of callbacks in any form, should first be discussed.

I personally see only very few edge cases where this would be really useful, if anything, I'd much more question, whether we can workaround the event queue blocking and thus have pollEvent return. Because if you do go with a callback, you now have two different entry points into the game loop if you want your game to simply continue during resize.

If we do want to provide such a feature, then I'd would want so see a cross-platform implementation (at least for Linux and macOS as well).

eliasdaler:
I don't think that's the right approach - if you want async events, then you need to poll events in another thread. Callback system seems nice, but then we'll need a callback for every window event, which I'm not sure is worth it.

eXpl0it3r:
I don't think we'll ever add this. For one, it's hard to make this useful in a cross-platform manner, and then if you really want to hook into the event loop on Windows, you can already do that without modifying SFML:

```cpp
#include <SFML/Graphics.hpp>
#include <Windows.h>

LONG_PTR SfmlEventCallback = 0x0;

LRESULT CALLBACK CustomEventCallback(HWND handle, UINT message, WPARAM wParam, LPARAM lParam)
{
    if (message == WM_MENU...)
    {
        // Your event handling
    }

    return CallWindowProcW(reinterpret_cast<WNDPROC>(SfmlEventCallback), handle, message, wParam, lParam);
}

int main()
{
    auto app = sf::RenderWindow{ { 640u, 480u }, "Custom Event Handling" };
    HWND handle = app.getSystemHandle();

    SfmlEventCallback = SetWindowLongPtrW(handle, GWLP_WNDPROC, reinterpret_cast<LONG_PTR>(CustomEventCallback));

    while (app.isOpen())
    {
        for (auto event = sf::Event{}; app.pollEvent(event);)
        {
            {
                app.close();
            }
        }

        app.clear();
        app.display();
    }
}
```

As such, I'll close this PR.

PR #1612, opened 2019-09-14
fix(compilation): clang compilation on windows

Milerius:
When using LLVM-clang on Windows which share the same ABI as `MSVC` (that means clang is able to use MSVC compiled DLL)

the CMake library path was not updated resulting on not finding dependencies.

Fixed with detecting compiler as `SFML_COMPILER_CLANG` on Windows

Command line to build with LLVM-Clang on Windows: `cmake -G "Visual Studio 16 2019" -A x64 -T "ClangCl" -DCMAKE_CXX_COMPILER="C:/Program Files/LLVM/bin/clang-cl.exe" -DCMAKE_C_COMPILER="C:/Program Files/LLVM/bin/clang-cl.exe" ../`


eXpl0it3r:

PR #1614, opened 2019-09-26
Use non-const reference to Font as self documentation

markand:

## Description

The Font object has to exist during the lifetime of the Text object. By taking a const reference it's easy for someone to pass a temporary variable and have a dangling pointer afterwards.

In many libraries (including Boost), taking a non-const reference prevents temporary variables and clearly indicates to the programmer that the object must stay alive during the lifetime of the object.

## Tasks

* [x] Tested on Linux

### DIFF
--- a/include/SFML/Graphics/Text.hpp
+++ b/include/SFML/Graphics/Text.hpp
@@ -85,7 +85,7 @@ public:
     /// \param characterSize  Base size of characters, in pixels
     ///
     ////////////////////////////////////////////////////////////
-    Text(const String& string, const Font& font, unsigned int characterSize = 30);
+    Text(const String& string, Font& font, unsigned int characterSize = 30);
 
     ////////////////////////////////////////////////////////////
     /// \brief Set the text's string
@@ -123,7 +123,7 @@ public:
     /// \see getFont
     ///
     ////////////////////////////////////////////////////////////
-    void setFont(const Font& font);
+    void setFont(Font& font);
 
     ////////////////////////////////////////////////////////////
     /// \brief Set the character size
@@ -436,7 +436,7 @@ private:
     // Member data
     ////////////////////////////////////////////////////////////
     String              m_string;              ///< String to display
-    const Font*         m_font;                ///< Font used to display the string
+    Font*               m_font;                ///< Font used to display the string
     unsigned int        m_characterSize;       ///< Base size of characters, in pixels
     float               m_letterSpacingFactor; ///< Spacing factor between letters
     float               m_lineSpacingFactor;   ///< Spacing factor between lines
--- a/src/SFML/Graphics/Text.cpp
+++ b/src/SFML/Graphics/Text.cpp
@@ -96,7 +96,7 @@ m_fontTextureId      (0)
 
 
 ////////////////////////////////////////////////////////////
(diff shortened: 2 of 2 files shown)


eliasdaler:
I'm not sure if it really helps compared to what we have now - however this makes me think that sf::Text might modify sf::Font object, which is not true. 

At least it might prevent this situation:

```cpp
sf::Text text(sf::Font{}, ...); // rvalue can't be assigned to non-const lvalue!
```

but then there's no reason to store it as a pointer to non-const. *But* - it's still only prevents this error, but you're not sure that sf::Text won't modify sf::Font based on function signature.

markand:
> At least it might prevent this situation:
> 
> ```c++
> sf::Text text(sf::Font{}, ...); // rvalue can't be assigned to non-const lvalue!
> ```

Yes, exactly why I wanted that change because I was bitten by this as well. I've passed a temporary and my application crashed. Then I saw in the documentation that it requires a existing object until sf::Text is destroyed. Most API use const references when they require immediate read or make a copy if needed.

Qt perform a kind of implicit sharing (I don't like it much but it may helps). In that case, sf::Font could have a shared_ptr internal to the real underlying object and then, copying sf::Font would just increase that reference count rather that doing a full copy. This has the advantage of not requiring a shared_ptr in user code but has the inconvenient to hide this concept internally.

LaurentGomila:
sf::Font *does* use an internal shared pointer to the FreeType font, because SFML allows copy semantics for it but FreeType provides no way to duplicate a loaded font.

However I did it only because there's no other way around it, I don't want to expose that private detail and document sf::Font as an "implicitly shared" class.

The best solution to this "problem" is to read the documentation. Everything is clearly explained in API doc and tutorials :wink:

eXpl0it3r:
I too am against this change. It may prevent one use case, but it also makes the API communicate a different meaning, breaks existing code and forces people to potentially adapt their interfaces that communicate a different meaning.

For example a function that returns a `const sf::Font&` stops working and you're forced to change the interface to `sf::Font&`, now user code suddenly allows for anyone using that function to modify the font object.

Practical example, see the [SFML Shader example](https://github.com/SFML/SFML/blob/master/examples/shader/Effect.hpp#L67) which fails to build [on Travis](https://travis-ci.org/SFML/SFML/jobs/589968186#L693).

I'll close the PR for now, feel free to open it again, if you have additional arguments.

dmitry-t:
The proposed change breaks the API backward compatibility. There is a simple rule: never never never break compatibility in a minor version.

eliasdaler:
Didn't SFML already break ABI in minor releases?

eXpl0it3r:
Yes, we allow for breaking ABI changes between minor version, as it's a pointless task to ensure this.

As for API, if it's functionally the same, maybe with some limitation, there have been changes in the past, but this seems to alter the interface too much, so our own examples fail to compile.

Bromeon:
It's not only a breaking change, but disregards const-correctness and forces the caller to either have non-const references/pointers to `sf::Font` around, or to use `const_cast`.

To prevent temporary objects from being passed, there are other idioms (e.g. const pointer), but they have their own flaws (doesn't communicate whether `nullptr` is a valid argument). SFML doesn't use this convention. Also, I doubt that passing a default-constructed `sf::Font` is an error that occurs often enough to justify an API change.

Issue #1615, opened 2019-09-28
Add the possibility to get a glyph by its index in sf::Font


JoaoBaptMG:
This is a necessity that has arisen in a game I was developing a year ago. I am using [HarfBuzz](https://github.com/harfbuzz/harfbuzz) to be able to handle text in different scripts in different languages. The goal is to be able to output text in any language, utilizing any script and complex fonts shapes.

The issue I got through (and that had me modify SFML) is that `sf::Font`'s `getGlyph` accepts _Unicode code points_, whereas HarfBuzz's `hb_shape` method returns _glyph indices_, and thus it is impossible to pass those to `getGlyph`.

I am proposing (code in a pull request) to add another function (not so conveniently called `getGlyphId`) that allows me to get information about the glyph using the glyph index, instead of the code point.

The important code (that would demonstrate its ability) is [this one](https://github.com/JoaoBaptMG/ReboundTheGame/blob/master/MainGame/drawables/TextDrawable.cpp#L197-L223):

    auto prop = wrapper.shape(curPos, nextWordPos - curPos);
    for (const auto& glyphData : prop.glyphs)
    {
        // Even though here I call the index as code point, it is really the glyph index,
        // I need to change it in my next commit
        uint32_t c = glyphData.codepoint;
        size_t cluster = glyphData.cluster;

        if (outline)
        {
            auto tempPos = position + glyphData.offset - sf::Vector2f(outlineThickness, outlineThickness);
            const auto& glyph = font.getGlyphId(c, fontSize, false, outlineThickness);
            addGlyphQuad(verticesOutline, tempPos, defaultOutlineColor, glyph);
        }

        const auto& glyph = font.getGlyphId(c, fontSize, false);
        auto p = addGlyphQuad(vertices, position + glyphData.offset, defaultColor, glyph);

        position += glyphData.advance;
        // ...
    }

Here, `wrapper.shape` is just a wrapper over HarfBuzz's API. Its implementation can be found [here](https://github.com/JoaoBaptMG/ReboundTheGame/blob/master/MainGame/unicode/HarfBuzzWrapper.cpp#L48-L84).

LaurentGomila:
Thanks for your feedback. However you don't have to open an issue if you already have a PR ; I close this one, so that discussions about your modifications are not split.

PR #1616, opened 2019-09-28
Add the possibility to get a glyph by its index in sf::Font

JoaoBaptMG:
## Description

This pull request adds the feature described in #1615

## How to test this PR?

Unfortunately, I do not have any ready code that can prove this PR does the thing it should do. However, I have used a fork with the exact modification in my [game](https://github.com/JoaoBaptMG/ReboundTheGame), using an adapted text drawable that uses HarfBuzz and needs to load the glyphs by using their index instead of the code point.

### DIFF
--- a/include/SFML/Graphics/Font.hpp
+++ b/include/SFML/Graphics/Font.hpp
@@ -160,7 +160,7 @@ public:
     const Info& getInfo() const;
 
     ////////////////////////////////////////////////////////////
-    /// \brief Retrieve a glyph of the font
+    /// \brief Retrieve a glyph of the font using the glyph's code point
     ///
     /// If the font is a bitmap font, not all character sizes
     /// might be available. If the glyph is not available at the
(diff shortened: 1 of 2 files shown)


LaurentGomila:
Thanks for your contributions.

This is however a very specific use case. This function will clearly be for working with HarfBuzz; for everyone else, a glyph index has no meaning.

And yes, the function name is not the best one, since the function doesn't return the glyph ID 😉

JoaoBaptMG:
Would you have a better idea to make this method more “useful” to be put in SFML? Maybe integrate HarfBuzz with `sf::Font` and `sf::Text` directly?

(I’ll confess it’s also a bit of inexperience of mine, I wouldn’t know how to maintain and sync my fork without tearing it down entirely each time a new update to SFML is out)

LaurentGomila:
> Maybe integrate HarfBuzz with sf::Font and sf::Text directly?

That would probably be the best option.

khaledhosny:
> This is however a very specific use case. This function will clearly be for working with HarfBuzz; for everyone else, a glyph index has no meaning.

It should be useful for any layout engine not just HarfBuzz, since these engines need to output glyph indices and positions (or do the whole rendering themselves).

LaurentGomila:
> It should be useful for any layout engine not just HarfBuzz

Any layout engine (are there many?) *based on FreeType*. And in practice, I don't see any technical reason why another layout engine wouldn't be able to work with codepoints rather than glyph indices.

khaledhosny:
There are several other text layout engines besides HarfBuzz and some new ones in the making. FreeType is font rasterization library, but not everyone needs/wants to use it.

Many font glyphs don’t have any direct association with Unicode code points and can only be accessed by glyph indices, which why text layout engines takes text as input and give glyph indices as output. Virtually all common graphics libraries have ways to draw glyphs by glyph indices, otherwise they can’t be made to draw text in many of the world’s writing systems.

JoaoBaptMG:
> 
> 
> > Maybe integrate HarfBuzz with sf::Font and sf::Text directly?
> 
> That would probably be the best option.

If you want, I can work on augmenting `sf::Text` with HarfBuzz support and all the features I have added to [my own](https://github.com/JoaoBaptMG/ReboundTheGame/blob/master/MainGame/drawables/TextDrawable.hpp) [implementation of `sf::Text`](https://github.com/JoaoBaptMG/ReboundTheGame/blob/master/MainGame/drawables/TextDrawable.cpp). This would require, however, that `sf::Font` carries a HarfBuzz struct, since I think both it and HarfBuzz copy the font data for themselves.

JoaoBaptMG:
Moreover, it doesn't support only FreeType. It can work with other font libraries as well, like GDI, GLib and Graphite. But, since building it might be a bit complicated, maybe it should be offered as an extlib alongside FreeType (or rather, maybe replacing it, since the lib will have to include FreeType as well).

oomek:
Is it ok to change the newTexture to RenderTexture? Clearing with update() from an image may add some unneccessary overhead.

eXpl0it3r:
I don't see us adding this feature. Harfbuzz support is however still something we're very much interested in implementing, but probably after we no longer ship binaries in our source code.

Issue #1622, opened 2019-11-14
Variable / Function comments not shown in Intellisense


firepro20:
## Intellisense not showing suggestions in VS2019

I am using SFML 251 x86 with Visual Studio 2019. I managed to get the green circle example running, and followed documentation. When using the IntelliSense to understand what each variable / function does in the library, I am getting this alternative text for each -

XML comment contains invalid XML: Whitespace is not allowed at this location.

This is how comment are formatted in Events.hpp as an example - 

///< Code of the key that has been pressed

## Your environment

* Windows 10 Pro x64 Version 10.0.18362 Build 18362
* SFML 2.5.1, git master

Microsoft Visual Studio Community 2019
Version 16.3.6
VisualStudio.16.Release/16.3.6+29418.71
Microsoft .NET Framework
Version 4.8.03752

Installed Version: Community

Visual C++ 2019   00435-60000-00000-AA503
Microsoft Visual C++ 2019

C# Tools   3.3.1-beta3-19461-02+2fd12c210e22f7d6245805c60340f6a34af6875b
C# components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

IntelliCode Extension   1.0
IntelliCode Visual Studio Extension Detailed Info

Microsoft JVM Debugger   1.0
Provides support for connecting the Visual Studio debugger to JDWP compatible Java Virtual Machines

Microsoft MI-Based Debugger   1.0
Provides support for connecting Visual Studio to MI compatible debuggers

Microsoft Visual C++ Wizards   1.0
Microsoft Visual C++ Wizards

Microsoft Visual Studio VC Package   1.0
Microsoft Visual Studio VC Package

NuGet Package Manager   5.3.1
NuGet Package Manager in Visual Studio. For more information about NuGet, visit https://docs.nuget.org/

ProjectServicesPackage Extension   1.0
ProjectServicesPackage Visual Studio Extension Detailed Info

Visual Basic Tools   3.3.1-beta3-19461-02+2fd12c210e22f7d6245805c60340f6a34af6875b
Visual Basic components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Visual Studio Code Debug Adapter Host Package   1.0
Interop layer for hosting Visual Studio Code debug adapters in Visual Studio

Visual Studio Tools for Unity   4.3.3.0
Visual Studio Tools for Unity

## Steps to reproduce

Follow documentation, start coding and invoke Intellisense.

## Expected behavior

[...]

PR #1623, opened 2019-11-14
Changed Doxygen single-line comment format

MarioLiebisch:
Single line comments starting with `///` are interpreted by Microsoft Visual Studio as documentation containing XML code.

Therefore single line comments starting with `///<` unfortunately cause a parsing error, as IntelliSense will assume the `<` is the start of an XML tag. This is not the case, but unfortunately, IntelliSense will still complain about the following space character rather than displaying the raw string. This commit alters all such comments to start with `//!<` instead, which prevents the issue.

Before:
Before Image

After:
After Image

This fixes issue #1622.


MarioLiebisch:
There's no difference. The it's just alternative syntax to avoid issues such as this one. The generated files are 100% identical.

eXpl0it3r:
I know, question is more, did you check or are you just trusting that it is so? 😉

MarioLiebisch:
WinMerge said everything is identical. ;)

Issue #1608, opened 2019-08-08
Documentation misleading concerning the order when combining transforms


pvigier:
Hello,

If I understand well the source code of `sf::Transform`, when we do:

```cpp
auto A = sf::Transform(...);
auto B = sf::Transform(...);
auto C = A.combine(B);
```
C is equal to A*B with * the matrix multiplication.

And when we want to transform a point:

```cpp
auto u = sf::Vector2f(...);
auto v = C.transformPoint(u);
```
We have v = A\*B\*u.

So it is B that is applied first on u, then A.

However, in the documentation of [combine](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1Transform.php#acd978f60421a0f839bb9a8263e8877ff), it is said that:

> The result is a transform that is equivalent to applying *this followed by transform. 

According, to this sentence, C would be equal to B*A (applying A followed by B). I think it is misleading.

Correct me if I am wrong.

Best regards,
Pierre

LaurentGomila:
I don't remember this kind of stuff that was written a long time ago, so the best thing would probably be to test it. A rotation combined with a translation should be a good case to figure out which one is applied first.

pvigier:
Indeed, I have done exactly that before posting to be sure.

Here is the snippet:

```cpp
auto transform = sf::Transform();
transform.translate(2.0f, 3.0f);
transform.rotate(90.0f);
```

Then, I print the entries of the matrix:

```
-4.37114e-08 -1 0 2 
1 -4.37114e-08 0 3 
0 0 1 0 
0 0 0 1 
```

Which corresponds to rotation first, translation then.

I think that the tutorial on [transforms](https://www.sfml-dev.org/tutorials/2.5/graphics-transform.php) is also misleading. It is said that :

>  You can apply several predefined transformations to the same transform as well. They will all be combined sequentially: 

I think it should be written: "They will all be combined sequentially in reversed order".

Bromeon:
Thanks for the report, I created a PR to address this.

PR #1628, opened 2019-12-07
Improve sf::Transform documentation

Bromeon:
Resolves misleading documentation mentioned in #1608, and adds clarification regarding equivalence of `sf::Transform` named functions and operators.

### DIFF
--- a/include/SFML/Graphics/Transform.hpp
+++ b/include/SFML/Graphics/Transform.hpp
@@ -100,6 +100,12 @@ public:
     ////////////////////////////////////////////////////////////
     /// \brief Transform a 2D point
     ///
+    /// These two statements are equivalent:
+    /// \code
+    /// sf::Vector2f transformedPoint = matrix.transformPoint(x, y);
+    /// sf::Vector2f transformedPoint = matrix * sf::Vector2f(x, y);
+    /// \endcode
+    ///
     /// \param x X coordinate of the point to transform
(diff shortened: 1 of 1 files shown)


include/SFML/Graphics/Transform.hpp
```diff
@@ -409,6 +427,7 @@ SFML_GRAPHICS_API Vector2f operator *(const Transform& left, const Vector2f& rig
 ///
 /// Performs an element-wise comparison of the elements of the
 /// left transform with the elements of the right transform.
+/// Matrices are expected to be exactly equal; no epsilon is involved.
```
eXpl0it3r: Kind of unrelated to the related ticket and I think it might need additional clarification. While people who would look for such a clarification will be familiar with what "no epsilon is involved" really means, those who don't know the meaning, will be more confused than helped by this comment.

Might more useful to explain that due to floating point imprecision the equality might not always be fulfilled and potentially offer a solution for said problem.

Bromeon: Thinking more about it, the main thing to express would be:

> Just because two matrices transform a vector in the same way, they need not be equal.

This can be due to floating point inaccuracy, but also going into the math details might be out of scope. I'm also not aware of a one-fits-all way to provide such a comparison. I removed the sentence for now, but we can gladly add a formulation if we find a good, intuitive one 🙂

PR #1634, opened 2020-01-04
Make the process to set a DirectInput device's axis mode more conservative

binary1248:
This patch changes the DirectInput device initialization process to only attempt setting the device's axis mode if the device reports having axes and its axis mode is not already set to absolute. This should help when dealing with broken device drivers that report being game controllers and not fully supporting game controller functionality such as was reported [here](https://en.sfml-dev.org/forums/index.php?topic=25361.0).

In order to test whether the negative case handling is improved, find someone with such a broken device driver. 😁

### DIFF
--- a/src/SFML/Window/Win32/JoystickImpl.cpp
+++ b/src/SFML/Window/Win32/JoystickImpl.cpp
@@ -80,6 +80,17 @@ namespace
 
     typedef std::vector<JoystickRecord> JoystickList;
     JoystickList joystickList;
+
+    struct JoystickBlacklistEntry
+    {
+        unsigned int vendorId;
+        unsigned int productId;
+    };
+
+    typedef std::vector<JoystickBlacklistEntry> JoystickBlacklist;
+    JoystickBlacklist joystickBlacklist;
+
+    const DWORD directInputEventBufferSize = 32;
 }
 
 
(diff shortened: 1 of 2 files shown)


binary1248:
Added support for buffered DirectInput data and a device blacklist to abort attempting to initialize the same device in the future.

bytercode:
How do i fix this ;(

PfAndrey:
@eXpl0it3r 
Y-axis of my gamepad doesn't work after this merge (for the previous commit it worked well).

> sf::Joystick::getAxisPosition(0, sf::Joystick::Axis::**PovY**) always returns 0 for me.

OS: 10.0.18363
Gamepad: defender game racer turbo rs3

eXpl0it3r:
@binary1248 can you explain this effect?

binary1248:
If I had to guess, I'd say this check fails:
https://github.com/SFML/SFML/pull/1634/files#diff-f7c9ce23dfc6bdad82ec2a0e246d1b05R764

If we can't set the axis mode to absolute although the device reports having axes the entire device is blacklisted. This verification was not done before, we just assumed that it always worked out somehow. Now that it is done we get a whole family of devices/drivers that misbehave or at least don't behave in a way that we can reason about. If we can't guarantee that the device's axis mode is set to absolute, we also can't guarantee that the values reported by our Joystick API are usable, thus this check. If the axis mode was always absolute for said device but it never reported it correctly, then... well... 🤷‍♂️

Like I said, this is only an educated guess. I don't have the hardware in question, so I can't put this hypothesis to the test. One would have to step through the Joystick code with said controller attached to see what happens and report their findings here. If it is an oversight or something trivial to fix then I have nothing against fixing the code again.

Issue #1536, opened 2019-01-04
Supporting font fallbacks for unicode


StephenLynx:
Currently, SFML supports only a single font for a sf::Text object. So what happens when your string have  multiple unicode characters is that they probably are not going to be rendered, specially if you are talking about non-textual characters, such as ☆.

Now, what can be done is to use a library such as fontconfig to use system fallbacks to render any character that isn't present on the chosen font.

As it stands, sfml text rendering is very limited because of how you can't render arbitrary unicode characters. Imagine a text client, a terminal emulator or a file browser that can't make use of these fallbacks to system fonts to render unicode.

StephenLynx:
Update: as discussed on irc, an alternative would be to have SFML to provide just a means for a programmer to provide a fallback.

Something like passing a function pointer that is used if the character isn't found on the font.

MarioLiebisch:
I actually modified `sf::Font` in the past for personal use to allow the user to add a `sf::Image` as an extra glyph for a code point of the author's choice. In my case I added icons for things such as gamepad button images and items as extra glyphs outside the code range my regular texts used. You're talking about something like that?

StephenLynx:
Yes. For example, if the programmer knows a certain glyph wasn't found on his font, he can then search for the glyph on some other font installed on his system, which is what pango by using fontconfig.

Also, I found out about FT_Get_Char_Index. This function returns 0 if the given character can't be found on the loaded font and seems to be the best way to tell if a fallback is needed to load a character.

Edgaru089:
In my opinion, SFML does not support loading system fonts in the first place, so, instead of loading system fallbacks silently, we should somehow tell the programmer that the glyph is missing in this certain font and let him/her to resolve the issue.

The quick fix to me should be something like a boolean in sf::Glyph to tell if the glyph is really provided in the font (or just a dummy). I might attempt implementing that if it sounds nice.

IMHO, something like a font manager which attempts loading a certain codepoint from a list of fonts one by one sounds nice too.

eXpl0it3r:
With #1644 merged, you can now query the font for valid glyphs. It's not exactly a fallback feature, but you can now achieve something similar, checking some special glyphs and if they don't exist try a different font.

Should we keep this issue open or do we consider this solved, @binary1248 / @LaurentGomila?

binary1248:
[...]

PR #1644, opened 2020-02-25
Added Glyph::hasGlyph field

Edgaru089:
## Description

This is a bare minimum implementation that partially fixes the issue #1536. It clearly isn't the best solution, but at least it gives users a way to handle missing glyphs themselves, with only a few lines of code.

There might be other things to be done. For instance, if multiple missing glyphs are requested in the same font, multiple copies of the same "missing character" glyph (with glyph index zero) are added to the font texture, wasting resources.

## How to test this PR?

Run the following minimum example with a font file (say, "sansation.ttf") in the working directory:
```cpp
sf::Font font;
font.loadFromFile("sansation.ttf");

sf::Glyph glyphA = font.getGlyph('A', 12, false);
std::cout << std::boolalpha << glyphA.isValid << std::endl; // true

sf::Glyph glyphB = font.getGlyph(0x03B1, 12, false); // lower case Greek letter "Alpha"
std::cout << std::boolalpha << glyphB.isValid << std::endl; // false
```

### DIFF
--- a/include/SFML/Graphics/Font.hpp
+++ b/include/SFML/Graphics/Font.hpp
@@ -166,6 +166,10 @@ public:
     /// might be available. If the glyph is not available at the
     /// requested size, an empty glyph is returned.
     ///
+    /// You may want to use \ref hasGlyph to determine if the
+    /// glyph exists before requesting it. If the glyph does not
+    /// exist, a font specific default is returned.
+    ///
     /// Be aware that using a negative value for the outline
     /// thickness will cause distorted rendering.
     ///
@@ -178,6 +182,24 @@ public:
     ///
     ////////////////////////////////////////////////////////////
     const Glyph& getGlyph(Uint32 codePoint, unsigned int characterSize, bool bold, float outlineThickness = 0) const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Determine if this font has a glyph representing the requested code point
+    ///
+    /// Most fonts only include a very limited selection of glyphs from
+    /// specific Unicode subsets, like Latin, Cyrillic, or Asian characters.
+    ///
+    /// While code points without representation will return a font specific
+    /// default character, it might be useful to verify whether specific
+    /// code points are included to determine whether a font is suited
+    /// to display text in a specific language.
+    ///
+    /// \param codePoint Unicode code point to check
+    ///
+    /// \return True if the codepoint has a glyph representation, false otherwise
+    ///
+    ////////////////////////////////////////////////////////////
+    bool hasGlyph(Uint32 codePoint) const;
 
     ////////////////////////////////////////////////////////////
     /// \brief Get the kerning offset of two glyphs
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -366,6 +366,13 @@ const Glyph& Font::getGlyph(Uint32 codePoint, unsigned int characterSize, bool b
 
 
 ////////////////////////////////////////////////////////////
+bool Font::hasGlyph(Uint32 codePoint) const
+{
+    return FT_Get_Char_Index(static_cast<FT_Face>(m_face), codePoint) != 0;
+}
+
+
+////////////////////////////////////////////////////////////
 float Font::getKerning(Uint32 first, Uint32 second, unsigned int characterSize) const
 {
     // Special case where first or second is 0 (null character)


MarioLiebisch:
Wouldn't this be more logical as `sf::Font::hasGlyph(int codePoint)`?

LaurentGomila:
> it gives users a way to handle missing glyphs themselves, with only a few lines of code.

Can you explain how? What code would a user write, to "handle missing glyphs"?

1aam2am1:
I would copy sf::Text code and in update method change way of getting glyph. As we can draw one vertexarray per font texture. We would need to create multiple arrays for different fonts. Whatchout for spacing between different fonts (kerning). And draw multiple times vertex arrays.

Edgaru089:
> Wouldn't this be more logical as `sf::Font::hasGlyph(int codePoint)`?

Wait... I do think this would be more logical. Sorry for bringing such a complete mess. Should I close this, or just force-push that to my branch?

> Can you explain how? What code would a user write, to "handle missing glyphs"?

`sf::Text` wasn't modified, so the user should write his/her own Text class.

```cpp
sf::VertexArray array[N];
sf::Font fonts[N];
float advance = 0.0f;
for (sf::Uint32 codepoint: string) // Just to keep it simple, I know we don't use C++11
{
    for (int i = 0; i < fonts.size(); i++)
    {
        sf::Glyph g = fonts[i].getGlyph(codepoint);
        if (g.isValid || i == fonts.size() - 1)
        {
            array[i].push_back(...);
            ...
            advance += g.advance;
        }
    }
}
```

We should be handling kerning (not written) only when two adjacent characters are from the same font. That may require extra bookkeeping to let us know which font every glyph is from.

LaurentGomila:
Does it make sense to add 5% of the final solution without the rest (modifications to sf::Text, multi-font system)? I don't think that many users have their own implementation of sf::Text, or will write one.

MarioLiebisch:
> Wait... I do think this would be more logical. Sorry for bringing such a complete mess. Should I close this, or just force-push that to my branch?

Just force-push, it will update the pull request as well. I think a simple query function is somewhat reasonable.

**Potential use case:** Allow the user to pick a different font face for your game or application, but at the same time test the font first to see if it's actually suitable. For example, if your program is supposed to show Chinese glyphs, you'd want to make sure the font the user selected actually supports displaying these glyphs. Otherwise they might end up with an unusable UI.

Edgaru089:
Ok. Now it works like:
```cpp
sf::Font font;
font.loadFromFile("sansation.ttf");

std::cout << std::boolalpha << font.hasGlyph('A') << std::endl;    // true
std::cout << std::boolalpha << font.hasGlyph(0x03B1) << std::endl; // false
```

src/SFML/Graphics/Font.cpp
```diff
@@ -365,6 +365,15 @@ const Glyph& Font::getGlyph(Uint32 codePoint, unsigned int characterSize, bool b
 }
 
 
+////////////////////////////////////////////////////////////
+bool Font::hasGlyph(Uint32 codePoint) const
+{
+    FT_Face face = static_cast<FT_Face>(m_face);
```
MarioLiebisch: I'd probably inline this since it's really just a cast and only used once.

include/SFML/Graphics/Font.hpp
```diff
@@ -179,6 +179,21 @@ class SFML_GRAPHICS_API Font
     ////////////////////////////////////////////////////////////
     const Glyph& getGlyph(Uint32 codePoint, unsigned int characterSize, bool bold, float outlineThickness = 0) const;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Tell if a codepoint has a corresponding glyph in the font
```
MarioLiebisch: > Determine if this font has a glyph representing the requested code point

include/SFML/Graphics/Font.hpp
```diff
@@ -179,6 +179,21 @@ class SFML_GRAPHICS_API Font
...
+    ///
+    /// A font does not have every Unicode character in it. If the glyph
+    /// of a nonexistent codepoint is requested, a font-dependent
+    /// 'missing character' (often a rectangle with a question mark inside)
+    /// is returned.
```
MarioLiebisch: How about:
> Most fonts only include a very limited selection of glyphs from specific Unicode subsets, like Latin, Cyrillic, or Asian characters. While code points without representation will return a font specific default character,  it might be useful to verify whether specific code points are included to determine whether a font is suited to display text in a specific language.

Edgaru089:
> This sentence is now effective duplicated.

Why? The previous paragraph is about bitmap fonts, and this one is about every font. But the text does seem to be a little weird.

We should somehow document the behavior of "returning a font specific dummy", however. How about

> You may want to use \ref hasGlyph to determine if the glyph exists before requesting it. If the glyph does not exist, a font specific default is returned.

eXpl0it3r:

eXpl0it3r:
Merged in 43187455e407258cc4c304064777c59ceb0d13ca 🎉

Thanks for your contribution! 👍

PR #1645, opened 2020-02-25
Font text multiple textures

JonnyPtn:
A rebase of #1333 and updated the single line documentation format to use `\\!` as per the rest of the codebase.

Not sure exactly what i need to do to trigger multiple texture usage, but I have done a cursory test on the examples which use text and they all seem fine

### DIFF
--- a/include/SFML/Graphics/Font.hpp
+++ b/include/SFML/Graphics/Font.hpp
@@ -318,7 +318,7 @@ private:
     ////////////////////////////////////////////////////////////
     // Types
     ////////////////////////////////////////////////////////////
-    typedef std::list<Page> PageList; ///< List of pages, where each page corresponds to a texture
+    typedef std::list<Page> PageList; //!< List of pages, where each page corresponds to a texture
 
     ////////////////////////////////////////////////////////////
     /// \brief Free all the internal resources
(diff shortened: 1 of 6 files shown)


MarioLiebisch:
For testing it should be enough to use a large font combined with many different glyphs, so you fill up the available texture.

hobby8:
Can be tested by overriding Texture::getMaximumSize() to return a small power of two - that's how I used to do it, at least.

eXpl0it3r:

binary1248:
My only worry with this is that we are affecting the performance of the hot path through `Text::draw` by introducing 2 loops over associative containers that weren't necessary before. I don't think we can justify this potential performance hit for the majority of the users just to be able to support a new edge case most people don't care about. The more text-heavy the application is the bigger the hit will be.

include/SFML/Graphics/Font.hpp
```diff
@@ -250,8 +251,25 @@ class SFML_GRAPHICS_API Font
...
+    ///
+    /// The texture id is used internally by sf::Text.
+    ///
+    /// \param characterSize Reference character size
+    ///
+    /// \return Texture id containing glyphs of the requested size
+    ///
     ////////////////////////////////////////////////////////////
-    const Texture& getTexture(unsigned int characterSize) const;
+    Uint64 getTextureId(unsigned int characterSize) const;
```
eXpl0it3r, line 272: Even though `getTexture` was mostly for internal usage, am I correctly understanding that with this change you can't retrieve the textures as a user anymore?

hobby8: The user can access the texture with `getGlyph().texture` just like `Text` does - it's written in the documentation of the deprecated `getTexture()`.
`getTextureId()` is used internally only for invalidating the geometry, not for building it.

JonnyPtn:
Do we want this or not?

eXpl0it3r:
I think we definitely want this. But @binary1248's comment had me pause a bit. We'd either need some close-to-real-world example to check the performance impact or alternatively merge stuff and think about potential performance improvements, once people report them (and thus provide a "real world" example).

So it would be great if you could rebase it. 👍

JonnyPtn:
Github's not updating the PR for some reason despite having rebased the original branch - will do a new PR unless someone has a novel idea to solve.

RE performance I'm not sure there's a huge amount to compare - In the typical case where this solution isn't needed it's negligible, especially in the context of everything else going on in a draw call. In the case where this fix is needed it wouldn't have worked at all previously so there's nothing to compare.

Personally, I feel the cases where the performance would matter (apps using a huge amount of sf::Text instances on limited hardware) are going to be very rare, so I would go ahead with this fix

eXpl0it3r:
Superseded by #2282

PR #1649, opened 2020-03-17
Adding a static "Zero" position member to sf::Vector[2, 3]<T>

MrZeusTheCoder:
* [X] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before? It was discussed on the SFML Discord server. [Also on the forum](https://en.sfml-dev.org/forums/index.php?topic=18065.0) (as it has now come to my attention as I write this PR)
* [X] Does the code follow the SFML [Code Style Guide](https://www.sfml-dev.org/style.php)? As far as I can tell, yes. :smile:
* [X] If you have additional steps which need to be performed list them as tasks!

----

## Description

For a long time something has been slightly lacking in the SFML vector classes: an easy to access zero position (`sf::Vector2f(0, 0);`/`sf::Vector2f();`). This PR essentially allows for easier access to the the default constructor of a vector. It is a constant static member so it will also save the time that would be taken constructing a new object (if your compiler didn't already optimize that out). It also provides nicer API (IMHO): `if(Vector1 == sf::Vector2f::Zero)` vs `if(Vector1 == sf::Vector2f())`. 

## Tasks

* [ ] Make sure I am doing this PR thing properly. (My first time, :P)
* [X] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Just use `sf::Vector[2, 3]<T>::Zero` as you would a default constructed vector!

```cpp
#include <SFML/System/Vector2.hpp>
#include <SFML/System/Vector3.hpp>

int main()
{
    bool is_zero;
    // Vector2f
    sf::Vector2f TestVector2f = sf::Vector2f(1, 0);
    is_zero = (TestVector2f == sf::Vector2f::Zero); //False
    TestVector2f = sf::Vector2f::Zero;
    is_zero = (TestVector2f == sf::Vector2f::Zero); //True
    // Vector3f
    sf::Vector3f TestVector3f = sf::Vector3f(1, 0, 0);
    is_zero = (TestVector3f == sf::Vector3f::Zero); //False
    TestVector3f = sf::Vector3f::Zero;
    is_zero = (TestVector3f == sf::Vector3f::Zero); //True
}
```

I would also like to thank the entire SFML Team for the time they take out of their busy schedules to maintain, add to, and improve SFML. It means a lot to the entire community including myself. I don't think you guys could ever hear that enough. You guys are awesome.

### DIFF
--- a/include/SFML/System/Vector2.hpp
+++ b/include/SFML/System/Vector2.hpp
@@ -74,6 +74,11 @@ public:
     ////////////////////////////////////////////////////////////
     T x; //!< X coordinate of the vector
     T y; //!< Y coordinate of the vector
+
+    ////////////////////////////////////////////////////////////
+    // Static member data
+    ////////////////////////////////////////////////////////////
+    static const Vector2<T> Zero; //!< The zero (0, 0) coordinate of the vector
 };
 
 ////////////////////////////////////////////////////////////
@@ -283,6 +288,11 @@ typedef Vector2<float>        Vector2f;
 /// can be accessed directly (there are no accessors like setX(), getX()) and it
 /// contains no mathematical function like dot product, cross product, length, etc.
 ///
+/// The sf::Vector2 class also has a nifty little static member named Zero.
+/// It is the "zero-position" of the vector (i.e. sf::Vector2(0, 0)). This is quite useful
+/// for comparisons and other situations where you would otherwise have to
+/// construct a zero-ed vector.
+///
 /// Usage example:
 /// \code
 /// sf::Vector2f v1(16.5f, 24.f);
@@ -294,6 +304,9 @@ typedef Vector2<float>        Vector2f;
 /// v3 = v1 + v2;
 ///
 /// bool different = (v2 != v3);
+///
+/// sf::Vector2f v4(0, 0);
+/// bool is_zero = (v4 == sf::Vector2f::Zero);
 /// \endcode
 ///
 /// Note: for 3-dimensional vectors, see sf::Vector3.
--- a/include/SFML/System/Vector2.inl
+++ b/include/SFML/System/Vector2.inl
@@ -22,7 +22,8 @@
 //
 ////////////////////////////////////////////////////////////
 
-
+////////////////////////////////////////////////////////////
+// Templatized Constructors
 ////////////////////////////////////////////////////////////
 template <typename T>
 inline Vector2<T>::Vector2() :
@@ -32,8 +33,6 @@ y(0)
 
 }
 
-
-////////////////////////////////////////////////////////////
 template <typename T>
 inline Vector2<T>::Vector2(T X, T Y) :
 x(X),
@@ -42,8 +41,6 @@ y(Y)
 
 }
 
-
-////////////////////////////////////////////////////////////
 template <typename T>
 template <typename U>
 inline Vector2<T>::Vector2(const Vector2<U>& vector) :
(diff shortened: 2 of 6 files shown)


Bromeon:
This change would add a redundant symbol to the API. The fact that there would be two ways to achieve exactly the same can often lead to confusion. I have seen this convention of `Zero` sometimes in libraries, where a default constructor mimicks the behavior of basic types and does not initialize the fields, so you have an actual difference.

Furthermore, static constants can be problematic due to global initialization order. Since vectors are very fundamental types, it's not unlikely that users use them for global constants, e.g.:
```cpp
const sf::Vector2f gravity = isSpace ? sf::Vector2f::Zero : sf::Vector2f(0.f, 9.8f);
```
This is an unnecessary source of errors in my opinion.

MarioLiebisch:
Correct me if I'm wrong, but this might be a better fit once using C++11 with `constexpr`. But then again with that approach one could just use the default constructor itself, evaluated at compile time.

MrZeusTheCoder:
So I fixed the XCode error and the tests are passing now. As to the PR as a whole I partially agree with you. 

I am a little confused here though:

> I have seen this convention of `Zero` sometimes in libraries, where a default constructor mimicks the behavior of basic types and does not initialize the fields, so you have an actual difference.

Are you saying that in those libraries the  "default constructor mimicks the behavior of basic types and does not initialize the fields"? If so, this is not true of SFML vectors.

Are you saying that in those libraries the default constructor actually calls `Zero`.

I honestly don't understand what you are trying to say. Just rephrase yourself and I will probably get it. Lol.

@MarioLiebisch, my code already uses that approach, if I am not mistaken.

MarioLiebisch:
> Are you saying that in those libraries the "default constructor mimics the behavior of basic types and does not initialize the fields"? If so, this is not true of SFML vectors.

Exactly. Starting a data structure uninitialized is the most effective approach. In such cases, a default `Zero` object can be useful to instantly create a pre-defined instance. But since the default constructor in SFML does this already, it's not as important to have that extra `Zero` case, since `Vector()` and `Vector::Zero` basically result in the same code.

> my code already uses that approach, if I am not mistaken.

Not exactly. While `const` and `constexpr` might behave identical, it's not guaranteed. Otherwise they wouldn't have had to add a new keyword. `const` alone has the potential issues of initialization order etc. As far as I'm aware, it's still the case when using `constexpr` (or more specific: `constexpr const`), but less likely, based on what's going on.

Bromeon:
> Exactly. Starting a data structure uninitialized is the most effective approach. 

Performance-wise, it may make a _tiny_ difference, but in most cases it will not matter. However, not initializing variables is quite a bad practice and an unnecessary source of errors, that has kept itself around for surprisingly long. I'm glad that more modern languages like Rust force you to initialize variables, or [explicitly opt in](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html) to uninitialized state -- but it doesn't just happen by default.

> As far as I'm aware, it's still the case when using `constexpr` (or more specific: `constexpr const`), but less likely, based on what's going on.

The answer [here](https://stackoverflow.com/a/50425857) says it's not possible, but I'm not 100% sure. After all, C++20 also introduces `consteval` keyword for such purposes (and `constinit`). Global initialization is a big mess in C++, and the fact there are now 4 keywords that start with "const" certainly doesn't help...

Also, SFML is still at C++03.

MrZeusTheCoder:
So does this basically mean it's a no? I am totally willing to close the PR, but I just want to know if there is anything further you guys wanted to say or do.

MarioLiebisch:
I'd vote on "no for now", since it doesn't add anything (just creating a new vector has the very same effect). Not sure about other opinions, though.

LaurentGomila:
Same reply as the one I gave on the forum 5 years ago 😉

MrZeusTheCoder:
This isn't a needed API feature. It was a good learning experience for me though. Have a good day/night, all!

Issue #1657, opened 2020-05-10
Question: Can sf::TcpListener pick a random available port?


Unarelith:
## Subject

`sf::UdpSocket` is capable of selecting a random available port as described in the [documentation](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1UdpSocket.php#ad764c3d06d90b4714dcc97a0d1647bcc), using `sf::UdpSocket::bind(sf::Socket::AnyPort)`:

> You can use the special value Socket::AnyPort to tell the system to automatically pick an available port, and then call getLocalPort to retrieve the chosen port.

But there is nothing about `sf::TcpListener::listen(sf::Socket::AnyPort)`.

I tried on my ArchLinux. It works as expected, and I can use `sf::TcpListener::getLocalPort` to know which port it decided to pick.

Is the documentation missing, or is it unsupported/not portable/something else?

## Your environment

* ArchLinux
* SFML 2.5

LaurentGomila:
I'd say the documentation is missing. Internally, `TcpListener::listen` uses the same call to` ::bind` as `sf::UdpSocket`, so it should work the same.

Unarelith:
I just opened #1658 to add documentation on that topic, closing this ticket.

PR #1658, opened 2020-05-11
Added missing documentation about TcpListener::listen and Socket::AnyPort

Unarelith:
## Description

As the title says.

This PR is related to the issue #1657 

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/include/SFML/Network/TcpListener.hpp
+++ b/include/SFML/Network/TcpListener.hpp
@@ -74,6 +74,10 @@ public:
     /// function is called, it will stop listening on the old
     /// port before starting to listen on the new port.
     ///
+    /// When providing sf::Socket::AnyPort as port, the listener
+    /// will request an available port from the system.
+    /// The chosen port can be retrieved by calling getLocalPort().
+    ///
     /// \param port    Port to listen on for incoming connection attempts
     /// \param address Address of the interface to listen on
     ///
--- a/include/SFML/Network/UdpSocket.hpp
+++ b/include/SFML/Network/UdpSocket.hpp
@@ -78,9 +78,10 @@ public:
     ///
     /// Binding the socket to a port is necessary for being
     /// able to receive data on that port.
-    /// You can use the special value Socket::AnyPort to tell the
-    /// system to automatically pick an available port, and then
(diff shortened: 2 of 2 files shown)


include/SFML/Network/TcpListener.hpp
```diff
@@ -74,6 +74,10 @@ class SFML_NETWORK_API TcpListener : public Socket
     /// function is called, it will stop listening on the old
     /// port before starting to listen on the new port.
     ///
+    /// You can use the special value Socket::AnyPort to tell the
```
eXpl0it3r, line 77: Can you use the full qualifying name `sf::Socket::AnyPort`

include/SFML/Network/TcpListener.hpp
```diff
@@ -74,6 +74,10 @@ class SFML_NETWORK_API TcpListener : public Socket
...
+    /// system to automatically pick an available port, and then
+    /// call getLocalPort to retrieve the chosen port.
```
eXpl0it3r, line 79: I'd have probably gone more with something like this:

```
/// When providing sf::Socket::AnyPort as port, the listener
/// will request an available port from the system.
/// The chosen port can be retrieved by calling getLocalPort().
```

Unarelith: I basically copy-pasted the lines from UdpSocket.hpp: https://github.com/SFML/SFML/blob/master/include/SFML/Network/UdpSocket.hpp#L81-L83

Unarelith: But I prefer your way. Do you want me do edit UdpSocket.hpp too to keep consistency?

eXpl0it3r: Oh I didn't know. Still kind of prefer my version. Either way, I think it makes sense to have the `sf::` prefix in both versions.

Unarelith: Ok done. :)

Unarelith:
Sure

PR #1666, opened 2020-05-28
Adjustable processing interval in SoundStream #1517

unl1ght:
Resurrection of previous PR with updated documentation. Added setter to
adjust processing interval in SoundStream for low-latency streams.

----

## Description

Resurrection of previous PR with updated documentation. Added setter to
adjust processing interval in SoundStream for low-latency streams.

This PR is related to the issue #1517

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Audio.hpp>
#include <iostream>
#include <limits>
#include <vector>
#include <cmath>

class SignalGenerator : public sf::SoundStream
{
public:

    SignalGenerator(const sf::Time& processingInterval, const sf::Time& chunkDuration)
    : m_buffer()
    , m_samplePeriod(0)
    , m_sampleTime(0)
    {
        unsigned int channelCount = 1;
        unsigned int sampleRate = 44100;
        m_samplePeriod = 1.0 / sampleRate;

        size_t bufferLength = chunkDuration.asSeconds() * sampleRate;
        m_buffer.resize(bufferLength);

        SoundStream::initialize(channelCount, sampleRate);
        SoundStream::setProcessingInterval(processingInterval);
    }

protected:

    bool onGetData(sf::SoundStream::Chunk& data)
    {
        for (auto& sample : m_buffer)
        {
            constexpr double amplitude = 0.5 * std::numeric_limits<sf::Int16>::max();
            sample = amplitude * std::sin(330.0 * 2.0 * M_PI * m_sampleTime);
            m_sampleTime += m_samplePeriod;
        }

        data.sampleCount = m_buffer.size();
        data.samples = m_buffer.data();
        return true;
    }
    
    void onSeek(sf::Time timeOffset) { }

private:

    std::vector<sf::Int16> m_buffer;
    double m_samplePeriod;
    double m_sampleTime;
};

int main()
{
  SignalGenerator generator(sf::milliseconds(1000), sf::milliseconds(50));
  generator.play();
  std::cin.get();
}

```

### DIFF
--- a/include/SFML/Audio/SoundStream.hpp
+++ b/include/SFML/Audio/SoundStream.hpp
@@ -251,6 +251,20 @@ protected:
     ////////////////////////////////////////////////////////////
     virtual Int64 onLoop();
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Set the processing interval
+    ///
+    /// The processing interval controls the period at which the
+    /// audio buffers are filled by calls to onGetData. A smaller
+    /// interval may be useful for low-latency streams. Note that
+    /// the given period is only a hint and the actual period may
+    /// vary. The default processing interval is 10 ms.
+    ///
+    /// \param interval Processing interval
+    ///
+    ////////////////////////////////////////////////////////////
+    void setProcessingInterval(Time interval);
+
 private:
 
     ////////////////////////////////////////////////////////////
@@ -315,8 +329,9 @@ private:
     unsigned int  m_sampleRate;               //!< Frequency (samples / second)
     Uint32        m_format;                   //!< Format of the internal sound buffers
     bool          m_loop;                     //!< Loop flag (true to loop, false to play once)
-    Uint64        m_samplesProcessed;         //!< Number of buffers processed since beginning of the stream
+    Uint64        m_samplesProcessed;         //!< Number of samples processed since beginning of the stream
     Int64         m_bufferSeeks[BufferCount]; //!< If buffer is an "end buffer", holds next seek position, else NoLoop. For play offset calculation.
+    Time          m_processingInterval;       //!< Interval for checking and filling the internal sound buffers.
 };
 
 } // namespace sf
--- a/src/SFML/Audio/SoundStream.cpp
+++ b/src/SFML/Audio/SoundStream.cpp
@@ -51,7 +51,8 @@ m_sampleRate      (0),
 m_format          (0),
 m_loop            (false),
 m_samplesProcessed(0),
-m_bufferSeeks     ()
+m_bufferSeeks     (),
+m_processingInterval(milliseconds(10))
 {
 
 }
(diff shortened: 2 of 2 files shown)


unl1ght:
Apologies, I got a bit eager, I will need to append the test code tomorrow.

eXpl0it3r:
No worries, for now I just added the test code from the other PR. If you have a more complete example, that would be great.

unl1ght:
I have updated the test which audibly demonstrates the change in processing interval. Interestingly, changing the processing interval to zero does not produce the outcome that I would have expected in terms of reducing audio latency. At the standard 10ms interval, you want about 20 to 30 ms of audio per chunk to prevent the buffers from starving, but reducing the interval to zero does not reduce the amount of audio that you need per chunk. I suppose this would also require a change to the internal audio buffer sizes.

This is not a bug, but maybe an indication that a tunable processing interval alone is not enough to acheive the desired outcome?

LaurentGomila:
> reducing the interval to zero does not reduce the amount of audio that you need per chunk. I suppose this would also require a change to the internal audio buffer sizes.

Two comments there:
- there's no "internal audio buffer size", buffers have the size you give them in `onGetData`
- be aware that `sleep(ms)` is not meant for accurate timing, OSes will usually not go lower than the standard time slice (for example, on Windows this could be 16 ms)

According to this 2nd remark, and since this feature targets low-latency usages (who's interested in `setProcessingInterval(1000)`?), I wonder if it will be able to give 100% satisfaction on every environment.

unl1ght:
> there's no "internal audio buffer size", buffers have the size you give them in onGetData

Yes, good point, bad oversight on my part there.

`be aware that sleep(ms) is not meant for accurate timing, OSes will usually not go lower than the standard time slice (for example, on Windows this could be 16 ms)`

This shouldnt hold in the case that the processing interval is negative, in which a sleep shouldnt occur at all. Would you expect that in this situation, a chunk size of 10ms should not cause the audio device to starve?

LaurentGomila:
> This shouldnt hold in the case that the processing interval is negative, in which a sleep shouldnt occur at all. Would you expect that in this situation, a chunk size of 10ms should not cause the audio device to starve?

I'm not sure where you want to go there. A negative sleep? Why? No sleep at all is not a desirable situation, as the stream thread will eat all the CPU.

I was talking about small processing intervals (between 1 and 15), which is most likely what most people will use this function for.

unl1ght:
The negative processing interval was simply to stop the thread from sleeping as the sleep function does a check for time >= 0. I was thinking about this from the perspective of using the stream class to make a synthesizer or some other realtime audio generator that might require latency to be less than 30 ms.

eXpl0it3r:

Issue #1675, opened 2020-06-11
sf::Shader is not moveable


Eren121:
## Subject of the issue

It is not possible to return `sf::Shader` from function.

## Your environment

Any environnement

## Steps to reproduce

```cpp
#include <SFML/Graphics.hpp>

sf::Shader loadShader() {
    sf::Shader shader;
    /* Probably more code here */
    return shader;
}
int main() {
    sf::Shader myShader = loadShader();
}
```

## Expected behavior

The `sf::Shader` should not be copiable but should be moveable.

## Actual behavior

Compilation error.

## Solution

The compilation error comes from the implementation of `NonCopyable`. `NonCopyable` class mean also non moveable.
Trivial change in `sf::Shader` or code design change with `NonCopyable` to accept move semantic, but may / may not break other classes.

LaurentGomila:
We know that move semantics are not implemented in SFML, what we need is someone to actually do it ;)

Bromeon:
See [this forum thread](https://en.sfml-dev.org/forums/index.php?topic=21571) for discussion on C++ standards.

eliasdaler:
Just as a workaround for now - you can use `std::unique_ptr<sf::Shader>` for non-moveable objects and do something like this:

```cpp
#include <SFML/Graphics.hpp>

std::unique_ptr<sf::Shader> loadShader() {
    auto shader = std::make_unique<sf::Shader>();
    /* Probably more code here */
    return shader;
}

int main() {
    auto myShader = loadShader();
}
```

The solution isn't perfect, but I think it's wrong to add move constructor only for one class (and before we "officially" say that we support C++11 or newer in the SFML)

ChrisThrasher:
`sf::Shader` is now moveable

PR #1676, opened 2020-06-12
Added move constructor and move assignment operator for sf::Shader

Eren121:
Yes, multiple times in multiple topics
See related issue

**Related to >=C++11 change**
It's a syntactic change that should not break any cross-platform compatibility. The compatibility is related to the compiler and not the platform, and today most if not all modern compilers support C++11. But it can break C++03 code (any project use it?... if yes, they probably also use older versions of SFML that are also compatible with C++03).

Some tests did not pass because they aren't actually... C++11 compatible.

It's can also maybe be a change for another branch forecasting C++11.

----

## Description

Make sf::Shader moveable, and keep not copiable.

This PR is related to the issue https://github.com/SFML/SFML/issues/1675
Code from the issue:
```cpp
#include <SFML/Graphics.hpp>

sf::Shader loadShader() {
    sf::Shader shader;
    /* Probably more code here */
    return shader;
}
int main() {
    sf::Shader myShader = loadShader();
}
```


unl1ght:
I noticed that the move assignment operator performs a swap of variables between the two instances. Would it be more appropriate to destroy the lhs object resources, take ownership of the rhs object resources, then reinitialise the rhs resources to default/null states?

unl1ght:
I have some concerns with the use of swap in the move assignment operator that I have voiced on the forum. Please see https://en.sfml-dev.org/forums/index.php?topic=27116.msg173691#msg173691

include/SFML/Graphics/Shader.hpp
```diff
@@ -91,12 +91,24 @@ class SFML_GRAPHICS_API Shader : GlResource, NonCopyable
     ////////////////////////////////////////////////////////////
     Shader();
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Move constructor
+    ///
+    ////////////////////////////////////////////////////////////
+    Shader(Shader&& other);
```
reconndev, line 98: Move constructors should be marked `noexcept` whenever possible.

Also, looking at the implementation this seems like it should be:
```cpp
Shader(Shader&& other) noexcept = default;
```

include/SFML/Graphics/Shader.hpp
```diff
@@ -91,12 +91,24 @@ class SFML_GRAPHICS_API Shader : GlResource, NonCopyable
...
     /// \brief Destructor
     ///
     ////////////////////////////////////////////////////////////
     ~Shader();
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Move assignement
+    ///
+    ////////////////////////////////////////////////////////////
+    Shader& operator=(Shader&& other);
```
reconndev, line 110:
```cpp
Shader& operator=(Shader&& other) noexcept = default;
```

src/SFML/Graphics/Shader.cpp
```diff
@@ -228,6 +229,14 @@ m_uniforms      ()
 }
 
 
+////////////////////////////////////////////////////////////
+Shader::Shader(Shader&& other) :
+Shader()
+{
+    *this = std::move(other);
```
reconndev, line 236: You're using operator assignment inside constructor logic which loses the benefits of in-place construction.

Solution: Definitions are not needed when the declaration states `= default`

src/SFML/Graphics/Shader.cpp
```diff
@@ -239,6 +248,22 @@ Shader::~Shader()
 }
 
 
+////////////////////////////////////////////////////////////
+Shader& Shader::operator=(Shader&& other)
```
reconndev, line 252: Same as before: should probably be removed

Bromeon:
Valid points. There are different ways to implement move semantics in C++, we should agree on one.

I brought this up in the forum a while ago, see [here](https://en.sfml-dev.org/forums/index.php?topic=27116.msg173694#msg173694) and subsequent posts.

reconndev:
> There are different ways to implement move semantics in C++, we should agree on one.

I think it is important to keep a strong exception guarantee for move semantics. The transfer should always be intuitive, and I believe that `noexcept` grants this intuitiveness. I also posted my view on the forum thread you mentioned.

JonnyPtn:
superseded by #2277 ?

eXpl0it3r:
Yep ^

PR #1681, opened 2020-07-04
Move semantics for sf::WindowBase and sf::Window classes

unl1ght:

----

## Description

Added move constructor and move assignment operator to Window and Window Base class.

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Window.hpp>
#include <iostream>

int main()
{
    sf::Window window1(sf::VideoMode(800, 600), "My window");
    sf::Window window2(std::move(window1));
    sf::Window window3;

    std::cout << "Window 1 open: " << window1.isOpen() << std::endl;
    std::cout << "Window 2 open: " << window2.isOpen() << std::endl;
    std::cout << "Window 3 open: " << window3.isOpen() << std::endl;

    window3 = std::move(window2);

    while (window3.isOpen())
    {
        sf::Event event;
        while (window3.pollEvent(event))
        {
            {
                window3.close();
            }
        }
    }
}
```


Bromeon:
Strictly speaking, the current implementation is not exception-safe. While this does not matter much for SFML 2 and current types, chances are that we introduce exceptions in SFML 3, or SFML classes which work with user-defined types that can throw exceptions.

Before implementing all the move operations, we should probably define if strong exception guarantee is something we need, and which idioms to follow for copy/move operations. I think it makes sense too continue that discussion [in the forum thread](https://en.sfml-dev.org/forums/index.php?topic=27116).

unl1ght:
Temporarily closing PR, need to refork repository and resubmit.

PR #1683, opened 2020-07-16
Implementation of copy/move constructors and copy/move operator for sf::Transform

bogdasar1985:

----

## Description
I implemented copy & move constructors, and also copy & move assignments operators for `sf::CircleShape`.

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
#include "SFML/include/SFML/Window.hpp"
#include "SFML/include/SFML/Graphics.hpp"
#include <iostream>

int main()
{
    sf::Transform tr1(10,2,3,4,5,6,7,8,9);
    sf::Transform tr2(tr1);

    sf::Transform tr3(10,2,3,4,5,6,7,8,9);
    sf::Transform tr4(std::move(tr3));
    
    sf::Transform tr5(10,2,3,4,5,6,7,8,9);
    sf::Transform tr6 = tr5;

    sf::Transform tr7(10,2,3,4,5,6,7,8,9);
    sf::Transform tr8 = std::move(tr7);
    return 0;
}
```

If we look at `sf::Transform`s variables on GDB, we'll see that everything's working out right.


LaurentGomila:
```cpp
    sf::CircleShape cs2_cp_op = cs_cp_op;
    std::cout << "Copy assignment operator: " << std::endl;
    ...

    sf::CircleShape cs_mv_op(12.5f, 50UL);
    sf::CircleShape cs2_mv_op = std::move(cs_mv_op);
```
Those will still call the constructors. You need to assign in another expression (ie. not to initialize) if you want the assignment operators to be called.

bogdasar1985:
@LaurentGomila Yeah, I didn't think about parent classes. I had to start implement move-semantic from any base class (i.e. class without parent).

In that case, for example, follow code:
```
sf::CircleShape cs_cp_constr(12.5f, 50UL);
    cs_cp_constr.setFillColor(sf::Color::Red);
    std::cout << cs_cp_constr.getRadius() << ' ' << cs_cp_constr.getPointCount() << ' ' << cs_cp_constr.getFillColor().toInteger() << std::endl;
    
    sf::CircleShape cs2_cp_constr(std::move(cs_cp_constr));
    std::cout << cs2_cp_constr.getRadius() << ' ' << cs2_cp_constr.getPointCount() << ' ' << cs2_cp_constr.getFillColor().toInteger() << std::endl;
```

Print out the following:
```
12.5 50 4278190335
12.5 50 4294967295
```
 
As you can see, `Color` didn't move, because `Color` is a member of `sf::Shape`(parent class). I should fix it. It's not a little work, because there are many classes. Instead of parent classes, we also need an move-semantics for members of classes. Well, I'm thinking of making a class tree and starting from the root or from simplest classes without inheritances (if there are any.).

unl1ght:
The sf::Transform class does not inherit from sf::NonCopyable, so I don't think that explicit copy constructor/assignment operators are required? The explicit move constructor/assignment may be useful as they leave the moved-from sf::Transform in an identity matrix state. Should these move functions be implemented in terms of a swap function for consistency as per the forum post on move semantics?

bogdasar1985:
@unl1ght So, first, you think we need only move ctor/assign oper, without copy, right? Second, we need to implement them via swap, like that:
```CPP
struct A
{
    int a, b;
    A(int _a, int _b) : a(_a), b(_b) { std::cout << "Constructor" << std::endl; };
    A(const A& _A) : a(_A.a), b(_A.b) { std::cout << "Copy constructor" << std::endl; }; //Without this copy ctor.
    
    A(A&& _A) : a(0), b(0)
    {
        swap(_A);          //With that.
        std::cout << "Move constructor" << std::endl;
    }

    void swap(A& _A)         //And with that.
    {
        std::swap(a, _A.a);
        std::swap(b, _A.b);
    }
};
```
Right?

bogdasar1985:
Class sf::Transform does not need manual implementation of move-semantic. Pull Request is closed.

PR #1684, opened 2020-07-16
Copy automatically examples resource folder to binary output location

Madour:
## Description

After compiling examples, I noticed the resource folders where not copied automatically and needed to be manually copied.
I knew it was possible to copy folders with Cmake, so I did it.

## Tasks

The commit is Cmake related, should work on all platforms

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Set SFML_BUILD_EXAMPLES to TRUE in CMakeLists.txt and do the usual build commands.

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -218,6 +218,11 @@ macro(sfml_add_example target)
         set(target_input ${target_input} ${THIS_BUNDLE_RESOURCES})
     endif()
 
+    # copy resource folder to binary output location when needed
+    if(THIS_RESOURCES_DIR)
+        file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/${THIS_RESOURCES_DIR} DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+    endif()
+
     # create the target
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
Thanks for the contribution! 🙂 

Doesn't this now copy the resources for every build of the examples?
Copying resources around and other things, should from my understanding only really been done during the "installation" (i.e. when building the INSTALL target), which I thought works correctly already. If that's not the case let me know.

Madour:
This copy the resources once when you run `cmake ..` , but not for every build.
Indeed resources are copied correctly when installing; I didn't notice, because I was building the examples targets into my build folder (in this case resources are not copied), but not installing them, oopsie !

eXpl0it3r:
I think that's the expected behavior then. Just building things doesn't prepare stuff for usage, but once you install things, the resources, headers, binaries, etc. should be placed in the correct locations

binary1248:
This is normally what you use

`add_custom_command(example POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy_if_different ...)`

for...

PR #1689, opened 2020-07-27
Use bit-packing for booleans in Packet

Dhanton:

----

## Description

This PR adds a bit-packer that reduces packet size when appending multiple booleans in succession. The implementation is very simple and most of the internals of sf::Packet are not changed. 

I believe this change could significantly reduce bandwidth usage in some cases, while also adding  virtually no overhead in the rest.

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
#include <iostream>

int main()
{
    sf::Packet packet;

    packet << true << true << false << true;

    // Size is 1 byte (would've been 4 bytes with previous implementation)
    std::cout << packet.getDataSize() << std::endl;

    // Bit-packing is not used with other types
    packet << sf::Int16(133) << true << false << false;

    // Size is 4 bytes now (would've been 9 bytes with previous implementation)
    std::cout << packet.getDataSize() << std::endl;

    return 0;
}
```

### DIFF
--- a/include/SFML/Network/Packet.hpp
+++ b/include/SFML/Network/Packet.hpp
@@ -413,10 +413,12 @@ private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    std::vector<char> m_data;    //!< Data stored in the packet
-    std::size_t       m_readPos; //!< Current reading position in the packet
-    std::size_t       m_sendPos; //!< Current send position in the packet (for handling partial sends)
-    bool              m_isValid; //!< Reading state of the packet
+    std::vector<char> m_data;        //!< Data stored in the packet
+    std::size_t       m_readPos;     //!< Current reading position in the packet
+    std::size_t       m_sendPos;     //!< Current send position in the packet (for handling partial sends)
+    std::size_t       m_boolReadPos; //!< Current reading bit position (for sending booleans)
+    std::size_t       m_boolSendPos; //!< Current send bit position (for sending booleans)
+    bool              m_isValid;     //!< Reading state of the packet
 };
 
 } // namespace sf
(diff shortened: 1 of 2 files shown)


LaurentGomila:
I don't think it is that common to stuff many consecutive booleans in a packet, without any user-side optimization / packing. Isn't it easier to let users handle this if they really want to save those few bytes?

Moreover, it does add overhead: extra code means slightly more CPU operations to run, and most importantly, a maintenance cost and potential bugs if we forget to handle these extra attributes in future modifications to the class.

Dhanton:
I think you're right that it's not a common optimization, but I also don't think it's just a few bytes. In an multiplayer prototype I'm making total server bandwidth dropped from about 110 kB/s to 56 kB/s with just this change (4 players). This could become even more significant when having multiple game servers at the same time.

I don't think there's an extensable way users can add this functionality, since it cannot be implemented by deriving from the Packet class. Having said this I absolutely agree on the maintenance cost and maybe it's not worth it.

LaurentGomila:
Users will most likely define their own sf::Packet operators for complex data structures, so there they can easily pack data as they want, if they wish to optimize size. And not only booleans, they can optimize even further, since they have more knowledge about the whole data structure than sf::Packet.

Bromeon:
I agree with Laurent here: `sf::Packet` is a simple wrapper around a byte buffer, and provides write/read operations for basic types.

Its operations do what someone would expect: in practice it's usually a `memcpy/reinterpret_cast` of the data, plus some endian handling. If a `bool` were mapped to something that doesn't occupy `sizeof(bool)` space on the wire, this might be surprising.

As mentioned, you can overload `<<` and `>>` operators for higher-level data structures in order to (de-)serialize them, or even work with a serialization format/library and just write the bytes directly to `sf::Packet`.

eXpl0it3r:
Unfortunately, I have to agree with the previous comments. While I do understand the need for optimizations in certain cases, I also think that this is better done as an `<<` / `>>` overload and more likely for a whole struct than just a primitive type.

As such I'll close this PR for now. Thanks a lot of your contribution! 🙂

PR #1690, opened 2020-08-12
Implement smoothing for fonts

JonnyPtn:
## Description

Following on from forum discussion [here](https://en.sfml-dev.org/forums/index.php?topic=21058.0) this implements smoothing for fonts, via `setSmooth();`

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Can test using the following code. Space toggles the font's smoothing

```cpp

#include <SFML/Graphics.hpp>
int main()
{

    sf::RenderWindow window( sf::VideoMode( 800, 600, 32 ), "SFML text smoothing",
                             sf::Style::Titlebar | sf::Style::Close );


    sf::Font font;
    if (!font.loadFromFile("resources/font.ttf"))
        return EXIT_FAILURE;

    sf::Text message;
    message.setFont(font);
    int textSize = 40;
    message.setScale( { 1.5f,1.5f } );
    message.setCharacterSize(textSize);
    message.setFillColor(sf::Color::White);

    message.setString( "Some text about something or other" );
    

    while (window.isOpen())
    {
        // Handle events
        sf::Event event;
        while ( window.pollEvent( event ) )
        {
            if ( (event.type == sf::Event::Closed) ||
                 ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Escape)) )
            {
                window.close();
                break;
            }

            if ( event.type == sf::Event::KeyPressed )
            {
                switch ( event.key.code )
                {
                case sf::Keyboard::Space:
                    font.setSmooth( !font.isSmooth() );
                    break;

                case sf::Keyboard::Up:
                    message.setCharacterSize( ++textSize );
                    break;

                case sf::Keyboard::Down:
                    message.setCharacterSize( --textSize );
                    break;
                }
            }
        }

        window.clear(sf::Color(50, 200, 50));
        window.draw(message);
        window.display();
    }

    return EXIT_SUCCESS;
}

```

### DIFF
--- a/include/SFML/Graphics/Font.hpp
+++ b/include/SFML/Graphics/Font.hpp
@@ -276,6 +276,32 @@ public:
     const Texture& getTexture(unsigned int characterSize) const;
 
     ////////////////////////////////////////////////////////////
+    /// \brief Enable or disable the smooth filter
+    ///
+    /// When the filter is activated, the font appears smoother
+    /// so that pixels are less noticeable. However if you want
+    /// the font to look exactly the same as its source file,
+    /// you should disable it.
+    /// The smooth filter is enabled by default.
+    ///
+    /// \param smooth True to enable smoothing, false to disable it
+    ///
+    /// \see isSmooth
+    ///
+    ////////////////////////////////////////////////////////////
+    void setSmooth(bool smooth);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Tell whether the smooth filter is enabled or not
+    ///
+    /// \return True if smoothing is enabled, false if it is disabled
+    ///
+    /// \see setSmooth
+    ///
+    ////////////////////////////////////////////////////////////
+    bool isSmooth() const;
+
+    ////////////////////////////////////////////////////////////
     /// \brief Overload of assignment operator
     ///
     /// \param right Instance to assign
@@ -373,6 +399,7 @@ private:
     void*                      m_streamRec;   //!< Pointer to the stream rec instance (it is typeless to avoid exposing implementation details)
     void*                      m_stroker;     //!< Pointer to the stroker (it is typeless to avoid exposing implementation details)
     int*                       m_refCount;    //!< Reference counter used by implicit sharing
+    bool                       m_isSmooth;    //!< Status of the smooth filter
     Info                       m_info;        //!< Information about the font
     mutable PageTable          m_pages;       //!< Table containing the glyphs pages by character size
     mutable std::vector<Uint8> m_pixelBuffer; //!< Pixel buffer holding a glyph's pixels before being written to the texture
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -88,6 +88,7 @@ m_face     (NULL),
 m_streamRec(NULL),
 m_stroker  (NULL),
 m_refCount (NULL),
+m_isSmooth (true),
 m_info     ()
 {
     #ifdef SFML_SYSTEM_ANDROID
@@ -105,7 +106,8 @@ m_stroker    (copy.m_stroker),
 m_refCount   (copy.m_refCount),
 m_info       (copy.m_info),
 m_pages      (copy.m_pages),
-m_pixelBuffer(copy.m_pixelBuffer)
+m_pixelBuffer(copy.m_pixelBuffer),
+m_isSmooth   (copy.m_isSmooth)
 {
     #ifdef SFML_SYSTEM_ANDROID
         m_stream = NULL;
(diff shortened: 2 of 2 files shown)


src/SFML/Graphics/Font.cpp
```diff
@@ -468,6 +468,26 @@ const Texture& Font::getTexture(unsigned int characterSize) const
     return m_pages[characterSize].texture;
 }
 
+////////////////////////////////////////////////////////////
+void Font::setSmooth( bool smooth )
```
eXpl0it3r, line 472: Remove spaces around argument

src/SFML/Graphics/Font.cpp
```diff
@@ -468,6 +468,26 @@ const Texture& Font::getTexture(unsigned int characterSize) const
...
+{
+    if (smooth != m_isSmooth)
+    {
+        m_isSmooth = smooth;
+
+        for (int i = 0; i < m_pages.size(); ++i)
```
eXpl0it3r, line 478: Should use the key type `unsigned int` or probably even better, iterate over the container with iterators.

src/SFML/Graphics/Font.cpp
```diff
@@ -468,6 +468,26 @@ const Texture& Font::getTexture(unsigned int characterSize) const
...
+        {
+            m_pages[i].texture.setSmooth( m_isSmooth );
```
eXpl0it3r, line 480: No spaces around the argument

include/SFML/Graphics/Font.hpp
```diff
@@ -373,6 +399,7 @@ class SFML_GRAPHICS_API Font
     void*                      m_streamRec;   //!< Pointer to the stream rec instance (it is typeless to avoid exposing implementation details)
     void*                      m_stroker;     //!< Pointer to the stroker (it is typeless to avoid exposing implementation details)
     int*                       m_refCount;    //!< Reference counter used by implicit sharing
+    bool                       m_isSmooth;    //!< Status of the smooth filter
     Info                       m_info;        //!< Information about the font
```
eXpl0it3r, line 403: - Missing default initialization in constructor
- Missing copy in copy constructor
- Missing copy in assignment operator

eXpl0it3r (changes requested):
@LaurentGomila mentioned on the linked [forum thread](https://en.sfml-dev.org/forums/index.php?topic=21058.msg150492#msg150492) that one can't just apply the smoothness like that. 
So the question is, if we simply change the texture property is this enough or do we need to re-render the already rendered glyphs?

JonnyPtn:
Yeah he also clarified afterwards that he misunderstood and to ignore that comment, unless I misunderstand? the smoothing is not an issue when rendering glyphs, but when rendering texts

src/SFML/Graphics/Font.cpp
JonnyPtn: Done, with iterators

src/SFML/Graphics/Font.cpp
JonnyPtn: Done

include/SFML/Graphics/Font.hpp
JonnyPtn: Done

src/SFML/Graphics/Font.cpp
JonnyPtn: Done

JonnyPtn:
Feedback addressed and rebased onto master

include/SFML/Graphics/Font.hpp
```diff
@@ -275,6 +275,32 @@ class SFML_GRAPHICS_API Font
...
+    /// the font to look exactly the same as its source file,
+    /// you should disable it.
+    /// The smooth filter is enabled by default.
+    ///
+    /// \param smooth True to enable smoothing, false to disable it
+    ///
+    /// \see isSmooth
+    ///
+    ////////////////////////////////////////////////////////////
+    void setSmooth( bool smooth );
```
eXpl0it3r, line 292: Wrong spacing 😉

eXpl0it3r (approved):

include/SFML/Graphics/Font.hpp
JonnyPtn: :( think It must be picking up a rogue clang-format file from somewhere as this is the style I use for other projects, will fix this though ( and out of curiosity would a clang-format PR be worthwhile? )

eXpl0it3r: I would very much appreciate such a PR. Not sure how easy it is to reproduce SFML's coding style, but I'd also be willing to change some of it in favor of an automated formatting.

JonnyPtn: Done

binary1248 (approved):

binary1248:

eXpl0it3r (approved):

PR #1708, opened 2020-10-05
OpenAL Audio effects: reverb, chorus, delay

fallahn:
## Description

Implements OpenAL sound effects for reverb, chorus and delay. Can be easily extended to more types supported by OpenAL.
This was discussed some time (10 years!) ago here: https://en.sfml-dev.org/forums/index.php?topic=2245.0 the conclusion being that available OpenAL implementations were not capable. Thankfully things have changed somewhat in the last decade. I've tested this on platforms using openal-soft (windows and ubuntu) and on macOS using Apple's OpenAL library. I was pleasantly surprised to find that the effects are supported by the Apple library, although they are audibly rendered differently - particularly the delay which sounds much softer. However it seems openal-soft is supported on macOS and building SFML against that makes the effects sound consistent across tested platforms. I'm unable to test on mobile platforms.

I'm submitting this mostly because I find the effects useful, and it would be preferable not to have to maintain my own fork of SFML. I'd also like to gauge interest to potentially develop further with features such as low/high pass filters for example.

## Tasks

* [X] Tested on Linux - xubuntu 20.04
* [X] Tested on Windows 10 - 2004
* [X] Tested on macOS - 10.14
* [ ] Tested on iOS
* [ ] Tested on Android

## API Overview

- `sf::SoundSource` - Base class of e.g. `sf::Music` or `sf::Sound`
  - `void setEffect(const SoundEffect* effect);` - Set or remove (with `NULL`) effects to anything implementing SoundSource
  - `const SoundEffect* getEffect() const;` - Get the currently set effect or `NULL` if none is set
  - `void resetEffect();` - Internal use to reset the effect when the `sf::SoundEffect` instance is destroyed

- `sf::SoundEffect` - Sound effect base class
  - `static bool isAvailable();` - Not all platforms support effects (e.g. iOS)
  - `void setVolume(float volume);` - Volume of the effect in the range or 0 to 1
  - `float getVolume() const;` - Retrieve set volume

- `sf::ChorusEffect`
  - `void setWaveform(Waveform waveform);` - Set the LFO waveform shape, either Sine or Triangle
  - `void setPhase(sf::Int32 angle);` - Set the LFO phase
  - `void setRate(float rate);` - Set the LFO modulation rate
  - `void setDepth(float depth);` - Set the amount by which the delay time is modulated
  - `void setFeedback(float amount);` - Set the amount of processed signal that is fed back to the input
  - `void setDelay(float delay);` - Set the average amount of time the sample is delayed
  - `Waveform getWaveform() const;`
  - `sf::Int32 getPhase() const;`
  - `float getRate() const;`
  - `float getDepth() const;`
  - `float getFeedback() const;`
  - `float getDelay() const;`

- `sf::DelayEffect`
  - `void setDelay(float delay);` - Set the delay between the original sound and the first 'tap', or echo instance
  - `void setLRDelay(float delay);` - Set the delay between the first 'tap' and the second 'tap'
  - `void setDamping(float damping);` - Set the amount of high frequency damping applied to each echo
  - `void setFeedback(float feedback);` - Set the amount of feedback the output signal fed back into the input
  - `void setSpread(float spread);` - Set how hard panned the individual echoes are
  - `float getDelay() const;`
  - `float getLRDelay() const;`
  - `float getDamping() const;`
  - `float getFeedback() const;`
  - `float getSpread() const;`

- `sf::ReverbEffect`
  - `void setDensity(float density);` - Set the coloration of the late reverb
  - `void setDiffusion(float diffusion);` - Set the echo density in the reverberation decay
  - `void setGain(float gain);` - Sets the max amount of reflections and reverberation added to the final sound mix
  - `void setDecayTime(float decay);` - Set the reverberation decay time
  - `void setReflectionGain(float gain);` - Set  the overall amount of initial reflections relative to the Gain property
  - `void setReflectionDelay(float delay);` - Set the amount of delay between the arrival time of the direct path from the source to the first reflection from the source
  - `void setLateReverbGain(float delay);` - Set the overall amount of later reverberation relative to the Gain property
  - `void setLateReverbDelay(float delay);` - Set the begin time of the late reverberation relative to the time of the initial reflection
  - `void setRoomRolloff(float rolloff);` - Set the attenuation of the reflected sound
  - `float getDensity() const;`
  - `float getDiffusion() const;`
  - `float getGain() const;`
  - `float getDecayTime() const;`
  - `float getReflectionGain() const;`
  - `float getReflectionDelay() const;`
  - `float getLateReverbGain() const;`
  - `float getLateReverbDelay() const;`
  - `float getRoomRolloff() const;`

## How to test this PR?

I've updated the 'sound' example in the included examples directory to demonstrate the use of effects. The syntax is similar to that of `sf::SoundBuffer` to `sf::Sound`:

    sf::ReverbEffect effect;
    effect.setDecayTime(2.f);

    sf::Music music;
    music.loadFromFile("music.wav");
    music.setEffect(&effect);
    music.play();

### DIFF
--- a/examples/sound/Sound.cpp
+++ b/examples/sound/Sound.cpp
@@ -118,7 +118,7 @@ int main()
     playMusic("ding.flac", reverbEffect);
 
     // Play music from a mp3 file
-    playMusic("ding.mp3");
+    playMusic("ding.mp3", reverbEffect);
 
     // Wait until the user presses 'enter' key
     std::cout << "Press enter to exit..." << std::endl;
--- a/examples/sound/Sound.cpp
+++ b/examples/sound/Sound.cpp
@@ -26,17 +26,28 @@ void playSound()
 
     // Create a sound instance and play it
     sf::Sound sound(buffer);
-    sound.play();
+    sf::ReverbEffect effect;
 
-    // Loop while the sound is playing
-    while (sound.getStatus() == sf::Sound::Playing)
+    sound.setEffect(&effect);
+    effect.setDecayTime(20.f);
+    effect.setDensity(0.f);
+
+    for (auto i = 0; i < 2; ++i)
     {
-        // Leave some CPU time for other processes
-        sf::sleep(sf::milliseconds(100));
+        sound.play();
 
-        // Display the playing position
-        std::cout << "\rPlaying... " << sound.getPlayingOffset().asSeconds() << " sec        ";
-        std::cout << std::flush;
+
+        // Loop while the sound is playing
+        while (sound.getStatus() == sf::Sound::Playing)
+        {
+            // Leave some CPU time for other processes
+            sf::sleep(sf::milliseconds(100));
+
+            // Display the playing position
+            std::cout << "\rPlaying... " << sound.getPlayingOffset().asSeconds() << " sec        ";
+            std::cout << std::flush;
+        }
+        sound.setEffect(nullptr);
     }
     std::cout << std::endl << std::endl;
 }
@@ -72,6 +83,7 @@ void playMusic(const std::string& filename)
         std::cout << "\rPlaying... " << music.getPlayingOffset().asSeconds() << " sec        ";
         std::cout << std::flush;
     }
+
     std::cout << std::endl << std::endl;
 }
 
--- a/examples/sound/Sound.cpp
+++ b/examples/sound/Sound.cpp
@@ -27,9 +27,9 @@ void playSound()
     sf::Sound sound(buffer);
 
     // Create a reverb effect to assign to the sound
-    sf::ReverbEffect effect;
+    sf::ChorusEffect effect;
 
-    for (auto i = 0; i < 2; ++i)
+    for (int i = 0; i < 2; ++i)
     {
         sound.play();
 
@@ -46,8 +46,12 @@ void playSound()
 
         //apply the reverb effect
         sound.setEffect(&effect);
-        effect.setDecayTime(20.f);
-        effect.setDensity(0.f);
+        effect.setPhase(80);
+        effect.setRate(5.f);
+        effect.setFeedback(-0.5f);
+        effect.setDelay(0.01f);
+       // effect.setDecayTime(20.f);
+        //effect.setDensity(0.f);
     }
     std::cout << std::endl << std::endl;
 }
--- a/examples/sound/Sound.cpp
+++ b/examples/sound/Sound.cpp
@@ -42,7 +42,7 @@ void playSound(const sf::SoundEffect& effect)
         }
 
         // apply the effect
-        sound.setEffect(&effect);
+        sound.setEffect(effect);
     }
     std::cout << std::endl << std::endl;
 }
@@ -83,7 +83,7 @@ void playMusic(const std::string& filename, const sf::SoundEffect& effect)
         music.setPlayingOffset(sf::seconds(0.f));
 
         // Apply the effect the second time
-        music.setEffect(&effect);
+        music.setEffect(effect);
     }
 
     std::cout << std::endl << std::endl;
--- a/examples/sound/Sound.cpp
+++ b/examples/sound/Sound.cpp
@@ -28,10 +28,6 @@ void playSound()
     sf::Sound sound(buffer);
     sf::ReverbEffect effect;
 
-    sound.setEffect(&effect);
-    effect.setDecayTime(20.f);
-    effect.setDensity(0.f);
-
     for (auto i = 0; i < 2; ++i)
     {
         sound.play();
@@ -47,7 +43,10 @@ void playSound()
             std::cout << "\rPlaying... " << sound.getPlayingOffset().asSeconds() << " sec        ";
             std::cout << std::flush;
         }
-        sound.setEffect(nullptr);
+
+        sound.setEffect(&effect);
+        effect.setDecayTime(20.f);
+        effect.setDensity(0.f);
     }
     std::cout << std::endl << std::endl;
 }
--- a/examples/sound/Sound.cpp
+++ b/examples/sound/Sound.cpp
@@ -6,7 +6,6 @@
 #include <iostream>
 #include <string>
 
-
 ////////////////////////////////////////////////////////////
 /// Play a sound
 ///
@@ -26,12 +25,13 @@ void playSound()
 
     // Create a sound instance and play it
     sf::Sound sound(buffer);
+
+    // Create a reverb effect to assign to the sound
     sf::ReverbEffect effect;
 
     for (auto i = 0; i < 2; ++i)
     {
         sound.play();
-
 
         // Loop while the sound is playing
         while (sound.getStatus() == sf::Sound::Playing)
@@ -44,6 +44,7 @@ void playSound()
             std::cout << std::flush;
         }
 
+        //apply the reverb effect
         sound.setEffect(&effect);
         effect.setDecayTime(20.f);
         effect.setDensity(0.f);
--- a/examples/sound/Sound.cpp
+++ b/examples/sound/Sound.cpp
@@ -27,7 +27,7 @@ void playSound()
     sf::Sound sound(buffer);
 
     // Create a reverb effect to assign to the sound
-    sf::ChorusEffect effect;
+    sf::DelayEffect effect;
 
     for (int i = 0; i < 2; ++i)
     {
@@ -46,11 +46,7 @@ void playSound()
 
         //apply the reverb effect
         sound.setEffect(&effect);
-        effect.setPhase(80);
-        effect.setRate(5.f);
-        effect.setFeedback(-0.5f);
-        effect.setDelay(0.01f);
-       // effect.setDecayTime(20.f);
+        //effect.setDecayTime(20.f);
         //effect.setDensity(0.f);
     }
     std::cout << std::endl << std::endl;
--- a/examples/sound/Sound.cpp
+++ b/examples/sound/Sound.cpp
@@ -114,7 +114,7 @@ int main()
 
     // Play music from a flac file
     reverbEffect.setDecayTime(5.f);
-    reverbEffect.setDensity(0.1f);
+    reverbEffect.setDiffusion(0.1f);
     playMusic("ding.flac", reverbEffect);
 
     // Play music from a mp3 file
--- a/examples/sound/Sound.cpp
+++ b/examples/sound/Sound.cpp
@@ -10,7 +10,7 @@
 /// Play a sound
 ///
 ////////////////////////////////////////////////////////////
-void playSound()
+void playSound(sf::ReverbEffect& effect)
 {
     // Load a sound buffer from a wav file
     sf::SoundBuffer buffer;
@@ -25,9 +25,6 @@ void playSound()
 
     // Create a sound instance and play it
     sf::Sound sound(buffer);
-
-    // Create a reverb effect to assign to the sound
-    sf::DelayEffect effect;
 
     for (int i = 0; i < 2; ++i)
     {
@@ -44,10 +41,10 @@ void playSound()
             std::cout << std::flush;
         }
 
-        //apply the reverb effect
+        // apply the reverb effect
         sound.setEffect(&effect);
-        //effect.setDecayTime(20.f);
-        //effect.setDensity(0.f);
+        effect.setDecayTime(20.f);
+        effect.setDensity(0.f);
     }
     std::cout << std::endl << std::endl;
 }
@@ -96,8 +93,11 @@ void playMusic(const std::string& filename)
 ////////////////////////////////////////////////////////////
 int main()
 {
+    // Create a reverb effect to assign to the sound
+    sf::ReverbEffect effect;
+
     // Play a sound
-    playSound();
+    playSound(effect);
 
     // Play music from an ogg file
     playMusic("doodle_pop.ogg");
--- a/examples/sound/Sound.cpp
+++ b/examples/sound/Sound.cpp
@@ -10,7 +10,7 @@
 /// Play a sound
 ///
 ////////////////////////////////////////////////////////////
-void playSound(sf::ReverbEffect& effect)
+void playSound(const sf::SoundEffect& effect)
 {
     // Load a sound buffer from a wav file
     sf::SoundBuffer buffer;
@@ -41,10 +41,8 @@ void playSound(sf::ReverbEffect& effect)
             std::cout << std::flush;
         }
 
-        // apply the reverb effect
+        // apply the effect
         sound.setEffect(&effect);
-        effect.setDecayTime(20.f);
-        effect.setDensity(0.f);
     }
     std::cout << std::endl << std::endl;
 }
@@ -54,7 +52,7 @@ void playSound(sf::ReverbEffect& effect)
 /// Play a music
 ///
 ////////////////////////////////////////////////////////////
-void playMusic(const std::string& filename)
+void playMusic(const std::string& filename, const sf::SoundEffect& effect)
 {
     // Load an ogg music file
     sf::Music music;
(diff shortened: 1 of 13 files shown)


Bromeon:
Thanks a lot for your contribution! 🙂 

A few remarks regarding the API:

* If a `sf::SoundEffect` derived instance outlives the `sf::SoundSource` that uses it, it automatically deregisters it through the pointer list. This is nice, but it can be hard to get right, when values change their address.
What would e.g. the following do?

  ```cpp
  ChorusEffect effect, otherEffect;
  // setup effects
  
  sound.setEffect(effect);
  effect = otherEffect;
  ```

* Are there custom effects that users can implement by deriving `sf::SoundEffect`?
* If `setEffect()` with a null pointer means to clear the effect, this should be documented.
* If `resetEffect()` is for internal use, it should not be public, but private (if needed with friend).
* The necessity of `const_cast` is often a design error -- you already use `mutable` for "not visible const-ness", so it should be possible to invoke const methods on the `sf::SoundEffect` through a const pointer.

There are a few smaller things in the code, but I think they can be discussed at a later stage.

fallahn:
The pointer registration mechanism is copied from `sf::SoundBuffer` - unless I've missed something then it should be at least as reliable as losing a `sf::SoundBuffer` while a `sf::Sound` is still using it. This is the same for `resetEffect()` - it mimics [sf::Sound::resetBuffer()](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1Sound.php#acb7289d45e06fb76b8292ac84beb82a7). I considered it should be hidden from the public API when writing it but decided to stick to the existing design as closely as possible. I'd suggest that it probably should be changed, and the `sf::SoundBuffer` / `sf::Sound` relationship be updated too.

No, users can't implement their own effects with `sf::SoundEffect`.  I was hoping this would be the case at the outset as it would mean I could create my own effects without having to modify SFML at all. Perhaps I could find some middle ground here and update the API to allow users to create their own custom effect.

I don't recall where I used the `const_cast` - it may have been a mid-development hack I've forgotten to remove. I'll take another look at that.

LaurentGomila:
Here is a suggestion to make the design slightly simpler and safer.

Instead of `m_effect = setType(...)` then `assert(m_effect); alEffect(m_effect, ...)`, why not let the base class completely manage m_effect and only provide a `setParameter` protected function to derived classes?

```cpp
SoundEffect::SoundEffect(Type type); // or why not directly the AL_EFFECT_XXX enum, to avoid pre-defining all derived types in base class?

void SoundEffect::setParameter(ALenum param, int value);

void SoundEffect::setParameter(ALenum param, float value);
```

Bromeon:
The problem with exposing `ALenum` as a parameter type is that you need to pull in the whole AL headers.
Since it's simply a typedef for `int`, that could work as well.

fallahn:
I completely agree, it separates concerns much more cleanly, and allows users to inherit `sf::SoundEffect` (assuming they're willing to include the OpenAL headers). I've not tested every single effect listed in the spec but I'm sure some of the more obscure ones such as 'vocal morpher' have never been implemented outside of a few mid-00s soundblaster cards. This way SFML can provide an interface for setting parameters similarly to the way it does for `sf::SoundSource` for only known good/cross platform effects such as reverb/delay, but not prevent users from creating other effects in their own specific cases.

examples/sound/Sound.cpp
```diff
@@ -26,17 +25,24 @@ void playSound()
 
     // Create a sound instance and play it
     sf::Sound sound(buffer);
-    sound.play();
 
-    // Loop while the sound is playing
-    while (sound.getStatus() == sf::Sound::Playing)
+    for (int i = 0; i < 2; ++i)
```
eliasdaler, line 29: Maybe change "2" to some constant like "NUM_TIMES_SOUND_PLAYED"?
But I'm also not sure why we need to play it 2 times...

examples/sound/Sound.cpp
```diff
@@ -60,18 +66,26 @@ void playMusic(const std::string& filename)
     std::cout << " " << music.getChannelCount()         << " channels"      << std::endl;
 
     // Play it
-    music.play();
-
-    // Loop while the music is playing
-    while (music.getStatus() == sf::Music::Playing)
+    for (int i = 0; i < 2; ++i)
```
eliasdaler, line 69: Same applies here, consider replacing "2" by a constant or removing this.

include/SFML/Audio.hpp
```diff
@@ -43,7 +43,9 @@
 #include <SFML/Audio/SoundRecorder.hpp>
 #include <SFML/Audio/SoundSource.hpp>
 #include <SFML/Audio/SoundStream.hpp>
-
+#include <SFML/Audio/ReverbEffect.hpp>
```
eliasdaler, line 46: I think effect headers should be separated via blank newline and sorted alphabetically here

include/SFML/Audio/ChorusEffect.hpp
```diff
@@ -0,0 +1,232 @@
...
+    ////////////////////////////////////////////////////////////
+    /// \brief Set the current feedback value
+    ///
+    /// This property controls the amount of processed signal that is fed back to the
+    /// input of the chorus effect. Negative values will reverse the phase of the feedback
+    /// signal. At full magnitude the identical sample will repeat endlessly. At lower
+    /// magnitudes the sample will repeat and fade out over time. Use this parameter to
+    /// create a "cascading" chorus effect.
+    /// 
+    /// \param amount new feedback amount to set, between -1.f and 1.f. Defaults to 0.25f
```
eliasdaler, line 120: This review contains a lot of magic variables like "0.25f" or "0.404f". Can you please explain where they come from? I didn't find why they're significant.
(It's especially not clear why parameter ranges are the way they are: [some_magic_number_1;some_magic_number_2])

include/SFML/Audio/SoundEffect.hpp
```diff
@@ -0,0 +1,176 @@
...
+    void detachSoundSource(SoundSource* source) const;
+
+    sf::Uint32 m_effectSlot;    //!< OpenAL Effects slot handle
+    sf::Uint32 m_effect;        //!< OpenAL Effect handle assigned to this slot
+    int m_type;                 //!< OpenAL Effect type used to track reference counting
+
+    float m_volume;             //!< Current volume of this slot
+
+    mutable std::set<SoundSource*> m_soundlist; //!< List of SoundSources using this effect
+    void ensureEffect(int);    //!< Used to ensure that there is only one instance of OpenAL effect object for each type
```
eliasdaler, line 158: Missing parameter name - is it intentional?

src/SFML/Audio/SoundSource.cpp
```diff
@@ -206,4 +223,45 @@ SoundSource::Status SoundSource::getStatus() const
     return Stopped;
 }
 
+void SoundSource::setEffect(const SoundEffect* effect)
```
eliasdaler, line 226: It's still not clear to me why effect can't be a const ref here, same as `sf::SoundBuffer` is passed by const ref to `sf::Sound::setBuffer`. To remove effect, we can add `removeEffect` function, which will be an explicit way of saying "I don't want this effect anymore".

eliasdaler (changes requested):
Very nice PR, great job. A lot of work was clearly done here.
I have a few questions/remarks, though, see them in the review.

include/SFML/Audio/ChorusEffect.hpp
fallahn: These are the default values as laid out in the OpenAL specification. I guess the comments should be at least updated to state that fact.

examples/sound/Sound.cpp
fallahn: The first loop the audio is played without the effect applied, then the second loop the audio is played with the effect, to demonstrate the difference in output.

eXpl0it3r:
Not sure if this a limitation of iOS (Apple has deprecated OpenAL anyways) or if we need to link something additionally or use a different toolchain, but there are some linker errors in regards to the effects functions.

If we can find a workaround, then that would be great to add, otherwise we might need some capability check and stub the functions for iOS.

```
  "_alAuxiliaryEffectSlotf", referenced from:
      sf::SoundEffect::SoundEffect(sf::SoundEffect const&) in libsfml-audio-s.a(SoundEffect.cpp.o)
      sf::SoundEffect::setVolume(float) in libsfml-audio-s.a(SoundEffect.cpp.o)
  "_alAuxiliaryEffectSloti", referenced from:
      sf::SoundEffect::SoundEffect(int) in libsfml-audio-s.a(SoundEffect.cpp.o)
      sf::SoundEffect::SoundEffect(sf::SoundEffect const&) in libsfml-audio-s.a(SoundEffect.cpp.o)
      sf::SoundEffect::~SoundEffect() in libsfml-audio-s.a(SoundEffect.cpp.o)
      sf::SoundEffect::setParameter(int, float) in libsfml-audio-s.a(SoundEffect.cpp.o)
      sf::SoundEffect::setParameter(int, int) in libsfml-audio-s.a(SoundEffect.cpp.o)
  "_alDeleteAuxiliaryEffectSlots", referenced from:
      sf::SoundEffect::~SoundEffect() in libsfml-audio-s.a(SoundEffect.cpp.o)
  "_alDeleteEffects", referenced from:
      sf::SoundEffect::~SoundEffect() in libsfml-audio-s.a(SoundEffect.cpp.o)
  "_alEffectf", referenced from:
      sf::SoundEffect::setParameter(int, float) in libsfml-audio-s.a(SoundEffect.cpp.o)
  "_alEffecti", referenced from:
      sf::SoundEffect::SoundEffect(int) in libsfml-audio-s.a(SoundEffect.cpp.o)
      sf::SoundEffect::setParameter(int, int) in libsfml-audio-s.a(SoundEffect.cpp.o)
  "_alGenAuxiliaryEffectSlots", referenced from:
      sf::SoundEffect::SoundEffect(int) in libsfml-audio-s.a(SoundEffect.cpp.o)
      sf::SoundEffect::SoundEffect(sf::SoundEffect const&) in libsfml-audio-s.a(SoundEffect.cpp.o)
  "_alGenEffects", referenced from:
      sf::SoundEffect::ensureEffect(int) in libsfml-audio-s.a(SoundEffect.cpp.o)
```

eXpl0it3r:

fallahn:
As far as I know OpenAL-soft supports macOS and iOS. If SFML were to use that it would also make the behaviour of the effects consistent across platforms.

examples/sound/Sound.cpp
eliasdaler: I think it would be good to have a visual indicator for this. E.g. show "No effect" for the first play and "Chorus Effect" for the second. Ideally the code would be something like this:

```cpp
playWithoutEffect(...);
playWithEffect(...);
```

include/SFML/Audio/ChorusEffect.hpp
eliasdaler: Yep, reference to OpenAL spec would be enough

eXpl0it3r:
Bringing OpenAL-soft to iOS might be a bit of a larger task, so I wouldn't necessarily have this mixed with this PR, as such, do you see a possibility to somehow stub this on iOS?

fallahn:
I've added an empty implementation for IOS, but the CI doesn't seem to recognise the SFML_SYSTEM_IOS definition from Config.hpp?

src/SFML/Audio/SoundEffect.cpp
```diff
@@ -0,0 +1,310 @@
...
+////////////////////////////////////////////////////////////
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+
+#ifndef SFML_SYSTEM_IOS
+
+#include <SFML/Audio/AudioDevice.hpp>
+#include <SFML/Audio/SoundEffect.hpp>
```
eXpl0it3r, line 32: > but the CI doesn't seem to recognise the SFML_SYSTEM_IOS definition from Config.hpp?

Took me a while to figure this out, but since you're including the SoundEffect header only inside the `#ifndef` block, `SFML_SYSTEM_IOS` won't be defined at the time you check it, as the C++ file is empty from the perspective of the pre-processor when reaching that line.
Meaning, you should move this line outside the `#ifndef` blocks.

fallahn: 🤦‍♀️ Next question is: as this breaks the style guide for alphabetical includes (trying to follow it is how I managed to introduce this problem in the first place) can an exception be made in this specific case?

eXpl0it3r: Well yeah, if there's a technical reason why stuff has to be different, then it's okay to deviate from the guidelines.

fallahn: No problem, just didn't want to make assumptions.

eXpl0it3r:
I've added the public API changes at the top to get a faster overview.
And I've hidden comments that have already been addressed to keep a bit of an overview.

----

> If resetEffect() is for internal use, it should not be public, but private (if needed with friend).

> The pointer registration mechanism is copied from sf::SoundBuffer - unless I've missed something then it should be at least as reliable as losing a sf::SoundBuffer while a sf::Sound is still using it. This is the same for resetEffect() - it mimics sf::Sound::resetBuffer(). I considered it should be hidden from the public API when writing it but decided to stick to the existing design as closely as possible. I'd suggest that it probably should be changed, and the sf::SoundBuffer / sf::Sound relationship be updated too.

Maybe @LaurentGomila has some insight on why he chose the public design for SoundBuffer over private + friend.

eXpl0it3r:
> It's still not clear to me why effect can't be a const ref here, same as sf::SoundBuffer is passed by const ref to sf::Sound::setBuffer. To remove effect, we can add removeEffect function, which will be an explicit way of saying "I don't want this effect anymore".

I think this requires an additional API discussion.
A lot of SFML functions are mostly set-and-never remove, which can be quite annoying at times (e.g. when you want to explicitly close/detach a stream).
With pointers you allow to reset this via the same API that you set it, but now you're dealing with raw pointers and an implicit state the `NULL` = no/remove effect.
Alternatives include:
- Use references, but have a separate function "remove Effect"
- NullEffect which can be set by default

Anyone else's opinion on this design?

eXpl0it3r:
`setVolume` currently deals in the range of `0.f` to `1.f`, which is different to the `sf::SoundSource::setVolume` with a range of `0.f` to `100.f`. Both work since it's just basic math, question is whether we want to have this aligned?

eXpl0it3r:
The ranges are currently silently clamped to the min or max allowed value. That way even if you pass in like `-10.f` the property will be set to for example `0.f` as minimum value.

Question is, should we silently clamp the values or should we assert them instead or assert and then still clamp?

eXpl0it3r:

fallahn:
> 
> 
> `setVolume` currently deals in the range of `0.f` to `1.f`, which is different to the `sf::SoundSource::setVolume` with a range of `0.f` to `100.f`. Both work since it's just basic math, question is whether we want to have this aligned?

This is an oversight on my part - I'd expect it to be 0 - 100 the same as the rest of SFML. If I'm honest this has always confused me slightly as OpenAL uses 0 - 1, so I'm not sure why SFML would differ, particularly when it has to be converted, but I'm happy to roll with it.

As for asserting vs clamping - I'd personally prefer assertions because it's more obvious when something is wrong rather than silently hiding an error, although I'd expect OpenAL to print some kind of range error in these cases too. Ultimately though clamping the range has proven to be adequate enough in the existing functions with no adverse side effects that I'm aware of so I'd probably stick with that for consistency.

fallahn:
On further examination `setVolume()` actually multiplies the output of the associated `SoundSource`, meaning that 0 - 1 is a more reasonable range than 0 - 100, as well as it not setting the volume to any specific value. Rather it's dependent on the output/current volume of the `SoundSource`. To that end I've renamed `setVolume()` to `setVolumeMultiplier()`.

eXpl0it3r:

OliverGlandberger:
Hi! I just wanted to say that this project looks very promising. Nicely done! :D

If anyone involved has a minute to spare, I am curious about the following things:
• What's the current status of this project?
• Is this code free/public and useable in any type of project (assuming proper credits are provided)?
• I noticed the description mentions high-/lowpass filters, any chance this will actually happen?

I've been working on a game engine on and off for the past 4 years which has SFML as its base. Would be nice if I could incorporate these features when I get around to further developing my SoundManager. (:

fallahn:
Hi! To address your questions:

- I have no idea 😄
- Of course, it's all under the SFML license - to test it out just clone [my branch](https://github.com/fallahn/SFML/tree/audio_effects)
- This will happen if the existing PR gets accepted. Biggest hold up I believe is the lack of support on iOS.

src/SFML/Audio/SoundSource.cpp
fallahn: This has been addresed in [1b41eb0](https://github.com/SFML/SFML/pull/1708/commits/1b41eb0821fb56436836929b7e25fd36781b63a6) 🙂

fallahn:
> > It's still not clear to me why effect can't be a const ref here, same as sf::SoundBuffer is passed by const ref to sf::Sound::setBuffer. To remove effect, we can add removeEffect function, which will be an explicit way of saying "I don't want this effect anymore".
> 
> I think this requires an additional API discussion. A lot of SFML functions are mostly set-and-never remove, which can be quite annoying at times (e.g. when you want to explicitly close/detach a stream). With pointers you allow to reset this via the same API that you set it, but now you're dealing with raw pointers and an implicit state the `NULL` = no/remove effect. Alternatives include:
> 
>     * Use references, but have a separate function "remove Effect"

I've updated `SoundSource::setEffect()` to take a const reference (for consistency with the existing API) and renamed `resetEffect()` to `removeEffect()` since this is what it actually did anyway. It also (albeit unintentionally) addresses the question as to whether or not `resetEffect()` should be private 😊

eXpl0it3r:

ChrisThrasher:
Closed because #2749 gets rid of OpenAL entirely

eXpl0it3r:
Would've really loved to see this in an earlier version of SFML, but with us moving to miniaudio, it doesn't quite fit in.

Thanks @fallahn for all the work, hopefully it can still be useful to someone. 🙂

PR #1709, opened 2020-10-09
Update of documentation in Vector implementation based in system module

HoopDead:
----

## Description

Hello guys!
I have been writing a simple game in SFML right now for my University Classes and I guess that I found a mistake in your documentation. When reading the example given in Vector2.hpp I tried to create a new Vector for my Player object and when I used notation 
``` C++
sf::Vector2f v1(1.2f, 3.4f)
```
It didn't work, so I tried to assign variables, like in normal vector, and finally, it compiled correctly.

``` C++
sf::Vector2f v1 = { 1.2f, 3.5f }
```

If It is my mistake, I'm really sorry for wasting your time, but still in the learning process (as everyone I guess)

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/include/SFML/System/Vector2.hpp
+++ b/include/SFML/System/Vector2.hpp
@@ -285,7 +285,7 @@ typedef Vector2<float>        Vector2f;
 ///
 /// Usage example:
 /// \code
-/// sf::Vector2f v1(16.5f, 24.f);
+/// sf::Vector2f v1 = { 16.2f, 2.0f}
 /// v1.x = 18.2f;
 /// float y = v1.y;
 ///


LaurentGomila:
Thanks for your contribution, but before fixing the documentation, you should have asked (on the forum) why it didn't compile. Fixing something without understanding the reasons is not really a fix ;)

From the API documentation generated from the code, you can clearly see that such a constructor exists. So the construction as shown in the documentation should work.
https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1Vector2.php#aed26a72164e59e8a4a0aeee2049568f1

Please do a proper report (error message, minimal code that triggers it, etc.) and *then* we'll see if there's something to fix, and what.

PR #1712, opened 2020-10-21
Update stb_image (2.16 -> 2.26) and stb_image_write (1.07 -> 1.15)

### LINKED ISSUE #1711: Image loading race condition when image failed to load
## Subject of the issue
sf::Image::loadFromFile can't be safely used from multiple threads on different image instances (each thread having its own sf::Image) if there was an error loading images (such as file not found).

## Your environment

* Arch Linux
* git master
* GCC 10.2.0

## Expected behavior
It should be thread-safe to use different sf::Image instances in different threads.

## Actual behavior
The program can crash when using sf::Image::loadFromFile from multiple threads on different image instances (each thread having its own sf::Image) if there was an error loading images (such as file not found) because stb which sf::Image uses has a global variable for the failure reason string (see: extlibs/headers/stb_image/stb_image.h:832) and sfml tries to print the global variable in src/SFML/Graphics/ImageLoader.cpp:127. Between the call to stbi_load and stbi_failure_reason another thread could be in the middle of writing to the failure reason global variable that stb uses, causing a crash in the error printing.
A crash from this happens occasionally in my program and since its a thread race condition I cant provide a reliable code example to reproduce the issue, but here is what `valgrind --tool=helgrind` complains about:
```
 542 ==860571== Possible data race during write of size 8 at 0x4916540 by thread #7
 543 ==860571== Locks held: none
 559 ==860571==    at 0x48AD188: stbi__err(char const*) (stb_image.h:841)
 560 ==860571==    by 0x48ADD97: stbi_load (stb_image.h:1135)
 561 ==860571==    by 0x48CB8D7: sf::priv::ImageLoader::loadImageFromFile(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::vector<unsigned char, std::allocator<unsigned char> >&, sf::Vector2<unsigned int>&) (ImageLoader     .cpp:104)
 562 ==860571==    by 0x48ABD2F: sf::Image::loadFromFile(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (Image.cpp:125)
```
stb_image.h probably needs to be modified to use thread-safe error string to fix this issue

Bromeon:
Should fix #1711 - race condition on `stbi_failure_reason()`

This does not only avoid the race condition on the failure string itself, but also makes failure state thread-local, meaning that each string will be correctly associated with the last image it failed to load.

I haven't had the chance to test it yet. A it's quite a big library update, we should probably wait a bit before merging.

### DIFF
--- a/src/SFML/Graphics/CMakeLists.txt
+++ b/src/SFML/Graphics/CMakeLists.txt
@@ -117,6 +117,12 @@ elseif(SFML_OS_IOS)
     target_link_libraries(sfml-graphics PRIVATE z bz2)
 endif()
 
+# starting from Visual Studio 2015, inline versions of some C functions are used; for compatibility link this library
+# see https://docs.microsoft.com/en-us/cpp/porting/overview-of-potential-upgrade-issues-visual-cpp?view=msvc-160#libraries
+if(SFML_COMPILER_MSVC AND MSVC_VERSION GREATER_EQUAL 1900)
+    target_link_libraries(sfml-graphics PRIVATE legacy_stdio_definitions.lib)
+endif()
+
 sfml_find_package(Freetype INCLUDE "FREETYPE_INCLUDE_DIRS" LINK "FREETYPE_LIBRARY")
 target_link_libraries(sfml-graphics PRIVATE Freetype)
 
--- a/extlibs/headers/stb_image/stb_image.h
+++ b/extlibs/headers/stb_image/stb_image.h
@@ -1,5 +1,5 @@
-/* stb_image - v2.16 - public domain image loader - http://nothings.org/stb_image.h
-                                     no warranty implied; use at your own risk
+/* stb_image - v2.26 - public domain image loader - http://nothings.org/stb
+                                  no warranty implied; use at your own risk
 
    Do this:
       #define STB_IMAGE_IMPLEMENTATION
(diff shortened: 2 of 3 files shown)


eXpl0it3r:
There seems to be a conflict with the `sprintf_s` function defined somewhere in FreeType2 and the use of it since [this commit](https://github.com/nothings/stb/commit/a0b521fcf2bfbc88cf188a5d0af1a1cf47fe4d99) of stb_image_write

Bromeon:
The linker error appeared [in our forum](https://en.sfml-dev.org/forums/index.php?topic=17500.0) already 5 years ago:
> freetype.lib(bdf.obj) : error LNK2019: unresolved external symbol sprintf referenced in function _bdf_parse_properties

The problem is that MSVC inlined some functions (such as `sprintf()`) in more recent versions, which means that these functions are no longer appearing as symbols in system libraries. Libraries which were compiled with older MSVC versions will thus still try to link them but fail to do so. In order to fix this, MSVC provides a compatibility library `legacy_stdio_definitions.lib`, which is what I'm linking now. A detailed explanation can be found [here](https://docs.microsoft.com/en-us/cpp/porting/overview-of-potential-upgrade-issues-visual-cpp?view=msvc-160#libraries).

This fixes the problem for me. An alternative way would be to use a more recent FreeType2 library; this would mean however updating all the binaries for all platforms. For MSVC they can be found [here](https://github.com/ubawurinna/freetype-windows-binaries). Do we have an (semi-)automated way of compiling them?

src/SFML/Graphics/CMakeLists.txt
```diff
@@ -117,6 +117,12 @@ elseif(SFML_OS_IOS)
     target_link_libraries(sfml-graphics PRIVATE z bz2)
 endif()
 
+# on newer MSVC versions, inline versions of some C functions are used; for compatibility link this library
+# see https://docs.microsoft.com/en-us/cpp/porting/overview-of-potential-upgrade-issues-visual-cpp?view=msvc-160#libraries
+if(SFML_COMPILER_MSVC)
+    target_link_libraries(sfml-graphics PRIVATE legacy_stdio_definitions.lib)
+endif()
```
eXpl0it3r, line 124: Do we need some additional library check constraint, i.e. does this also compile/link with e.g. VS 2010

Bromeon: Good point, according to [this page], they support legacy_stdio_definitions.lib from Visual Studio 2015 and later.

Which version is still officially supported by SFML? Also, there are a few pending PRs introducing move semantics, so they will anyway break very old versions.

eXpl0it3r: For SFML 3 it won't matter, but up until then, I'd add the following constraint as well.

`if(SFML_COMPILER_MSVC AND MSVC_VERSION GREATER_EQUAL 1900)`

Bromeon: Updated.

eXpl0it3r (approved):

eXpl0it3r:
I don't really mind, but if you want to, you could squash the last two commits.

Bromeon:
Combined the last 2 commits, also added some description to the commit message.

PR #1713, opened 2020-10-23
Fix warning: control may reach end of non-void function [-Wreturn-type]

Eren121:
Fix warning and/or bug in androidKeyToSF() adding a default label for all keys _Keyboard::Unknown_


eXpl0it3r (approved):

eXpl0it3r:
Thanks for providing a fix! 🙂

PR #1715, opened 2020-10-24
Fix audio stream replay for server example

davi-v:
Fix replay feature of this example. Tested on Windows.

### DIFF
--- a/examples/voip/Server.cpp
+++ b/examples/voip/Server.cpp
@@ -189,6 +189,7 @@ void doServer(unsigned short port)
     std::cin.ignore(10000, '\n');
 
     // Replay the sound (just to make sure replaying the received data is OK)
+    audioStream.setPlayingOffset(sf::Time());
     audioStream.play();
 
     // Loop until the sound playback is finished


MrZeusTheCoder:
@davi-v is this connected to a known issue? I am working on example improvements and I might add this too the list. I don't have a Windows machine to test this on.

davi-v:
> @davi-v is this connected to a known issue? I am working on example improvements and I might add this too the list. I don't have a Windows machine to test this on.
I don't think this is an issue. Maybe when this example was created, SFML internally already set the playing offset to the beginning when the sound ended, but now you need to do it manually?

MrZeusTheCoder:
But if it isn't an issue why are you fixing it, :P. It is not working for you?

MrZeusTheCoder:
Bump @davi-v .

davi-v:
> But if it isn't an issue why are you fixing it, :P. It is not working for you?

The example claims to be able to hear the recording after it finishes (replay feature), but it didn't work for me, *unless* I set the playing offset to the beginning before replaying the sound. I said I don't think it's an issue in SFML's implementation, but I think it's an issue in this example code.

LaurentGomila:
> I said I don't think it's an issue in SFML's implementation, but I think it's an issue in this example code.

That's a possibility. However this should be investigated further, since this example has existed for years and you're the first one to report such a bug.

And note that `sf::Time::Zero` is slightly more expressive than `sf::Time()` 😉

eXpl0it3r:
This has been fixed with #2037

Can you give SFML master another try @davi-v?

davi-v:
It's working as documented now 👍

Issue #150, opened 2012-01-05
SO_REUSEADDR


TankOs:
Reimplement SO_REUSEADDR (was there in 1.6). Otherwise previously sockets linger around.

PR #1717, opened 2020-11-28
Add SO_REUSEADDR to socket.

jakiki6:

----

## Description

Now, you can kill your server and instantly restart it without an error message that the address is in use.

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Build an EmptyEpsilon server with it.

### DIFF
--- a/src/SFML/Network/Socket.cpp
+++ b/src/SFML/Network/Socket.cpp
@@ -106,6 +106,9 @@ void Socket::create(SocketHandle handle)
         // Set the current blocking state
         setBlocking(m_isBlocking);
 
+		const int data = 1;
+		setsockopt(m_socket, SOL_SOCKET, SO_REUSEADDR, &data, sizeof(int));
+
         if (m_type == Tcp)
         {
             // Disable the Nagle algorithm (i.e. removes buffering of TCP packets)


LaurentGomila:
As far as I remember, this flag has other implications that made it not so suitable for every situation.

Do you have a strong experience of socket flags, or did you just happen to successfully test this one in your particular case, and thought it should be the default in SFML? 😉

JonnyPtn:
Yeah I've got experience with this flag but would not claim to be a network expert. From my understanding it's a common "quick fix" to allow you to re-bind a port almost immediately after a socket has just closed on it. Typical behaviour is the port is still bound for a short while to allow any delayed messages to come through. The inherent risk is that the new socket may receive packets which were intended for the old socket that was bound to this port

I think there's some debate over whether using this flag is a good idea or not - Some people say the correct thing to do is wait out the timeout until the port can be bound again, some say that's being over-cautious and you can just ignore any stale packets that come through.

I don't feel strongly either way - waiting for the timeout is arguably more correct, but allowing address reuse is arguably simpler.

ColinH:
General practice for `SO_REUSEADDR` on "normal" sockets is to set it when you have a TCP server socket (i.e. TCP sockets that you call `listen()` and `accept()` on) that is bound to a fixed port.

In all other cases, including TCP server sockets that are not bound to a fixed port (e.g. by passing `0` as port number to `bind()`), TCP client sockets (i.e. TCP sockets that you call `connect()` on), and all UDP sockets, there is rarely a reason to set `SO_REUSEADDR`.

(In the case of UDP sockets there can be weird effects, e.g. when the OS allows two applications to simultaneously open a UDP socket on the same address and port...)

binary1248:
Since setting this socket option depends on the concrete use case, it would be something that SFML optionally allows the developer to activate via a member function.

ColinH:
@binary1248 Given that there are separate classes `TcpSocket`, `TcpListener` and `UdpSocket` my suggestion, if fitting for SFML, would be for `TcpListener` to automatically set `SO_REUSEADDR` when being bound to a local port that is not `0`, and the other two classes to never touch that flag (or optionally have a member function for manual use). That way all three socket classes, by default, do that what is correct for (nearly) all regular use cases.

oznogon:
previously: issue #150.

eXpl0it3r:
I believe we'll need a different implementation, if we do decide to support this, as such I'm going to close this PR.

PR #1719, opened 2020-11-29
Fixed joystick POV values not being set properly for devices that are…

binary1248:
This fixes an issue where joystick/controller POV values were not set correctly for devices that are accessed via DirectInput.

See: https://github.com/SFML/SFML/pull/1634#issuecomment-667706226

DirectInput provides a combined value for both X and Y hat directions, this value has to be translated back to the separate X and Y values that SFML exposes through its API. Previously the implementation erroneously treated X and Y hat axes as separate objects which could only work if the device reported the same hat twice as distinct objects. This was the case with some devices but not all causing this bug to not always be reproducible.

This fix was tested with both a joystick and a controller (the same one mentioned in the linked comment).

### DIFF
--- a/src/SFML/Window/Win32/JoystickImpl.cpp
+++ b/src/SFML/Window/Win32/JoystickImpl.cpp
@@ -907,7 +907,7 @@ JoystickState JoystickImpl::updateDInputBuffered()
         {
             if (m_axes[j] == events[i].dwOfs)
             {
-                if (j == Joystick::PovX)
+                if ((j == Joystick::PovX) || (j == Joystick::PovY))
                 {
                     unsigned short value = LOWORD(events[i].dwData);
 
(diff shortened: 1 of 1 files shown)


Dreamykass:
Tested this PR with the joystick example with my controlled Torid Speedlink. It's got two modes, XInput and DirectInput (dunno which is which), and both appear to work fine.

(Just the second sees the back triggers as on-off buttons instead of analog input. But I think that's a controller thing.)

[https://i.imgur.com/aiD9ebg.gif](https://i.imgur.com/aiD9ebg.gif)
[https://i.imgur.com/WlakuyF.gif](https://i.imgur.com/WlakuyF.gif)

MarioLiebisch (approved):
Seems to be working fine with my Saitek X-55 (incompatible with XInput). All 8 hat directions follow the same coordinates pattern as a Xbox One Conotroller (using XInput).

eXpl0it3r (approved):

PR #1688, opened 2020-07-25
Added support for directional arrow resize cursors on Linux

texus:
This change adds the following mouse cursors:
- SizeLeft
- SizeRight
- SizeTop
- SizeBottom
- SizeTopLeft
- SizeTopRight
- SizeBottomLeft
- SizeBottomRight

Those cursors will only look like directional arrows on Linux, on Windows and macOS it will display the same double-headed arrows that are used by existing cursors (e.g. SizeTopLeftBottomRight is used when SizeTopLeft or SizeBottomRight is requested).

As stated on the [the SFML forum](https://en.sfml-dev.org/forums/index.php?topic=27397), this change is needed to have diagonal resize cursors on Linux.

I have tested all newly added cursors on Linux.
I didn't test on Windows or macOS, but the change is straightforward.

Example code to test the change (will show arrow pointing to bottom right on Linux, and double-headed arrow from top left to bottom right on Windows and macOS):
```
sf::Cursor cursor;
cursor.loadFromSystem(sf::Cursor::SizeBottomRight);
window.setMouseCursor(cursor);
```

PR #1721, opened 2020-11-29
Retain cursor if an unsupported cursor is used

eXpl0it3r:
## Description

While testing #1688, I ran into random crashes and after a while I realized that the switch-case construct didn't cover all cursor types - as documented - which then lead to random crashes when picking the unsupported cursors.

## Tasks

* [x] Tested on macOS

## How to test this PR?

You can use this example that counts through all the cursor types and will crash with `master` right now.

```cpp
#include <iostream>
#include <SFML/Graphics.hpp>

int main()
{
    auto window = sf::RenderWindow{{800, 600, 32}, "SFML Window"};
    window.setFramerateLimit(60);

    auto cursor = sf::Cursor{};

    auto start = sf::Cursor::Arrow;
    auto current = sf::Cursor::Arrow;
    auto end = sf::Cursor::NotAllowed;

    cursor.loadFromSystem(current);
    std::cout << "New cursor type: " << current << "\n";

    while (window.isOpen())
    {
        {
            {
                window.close();
                return 0;
            }

            if ((event.type == sf::Event::KeyPressed))
            {
                switch (event.key.code)
                {
                    case sf::Keyboard::Escape:
                    {
                        window.close();
                        return 0;
                    }
                    case sf::Keyboard::Up:
                        if(current < end)
                        {
                            current = static_cast<sf::Cursor::Type>(current + 1);
                            cursor.loadFromSystem(current);
                            std::cout << "New cursor type: " << current << "\n";
                        }
                        break;
                    case sf::Keyboard::Down:
                        if(current > start)
                        {
                            current = static_cast<sf::Cursor::Type>(current - 1);
                            cursor.loadFromSystem(current);
                            std::cout << "New cursor type: " << current << "\n";
                        }
                        break;
                }
            }
        }


        window.clear();
        window.display();
    }
}```

### DIFF
--- a/src/SFML/Window/OSX/CursorImpl.mm
+++ b/src/SFML/Window/OSX/CursorImpl.mm
@@ -89,7 +89,10 @@ bool CursorImpl::loadFromSystem(Cursor::Type type)
 
     switch (type)
     {
-        default: return false;
+        default:
+            if (m_cursor)
+                [m_cursor retain];
+            return false;
 
         case Cursor::Arrow:           m_cursor = [NSCursor arrowCursor];               break;
         case Cursor::Text:            m_cursor = [NSCursor IBeamCursor];               break;

### CODE REVIEW

**Review by binary1248 — APPROVED** [review]

File: src/SFML/Window/OSX/CursorImpl.mm [hunk 1]
```diff
@@ -89,7 +89,10 @@
 
     switch (type)
     {
-        default: return false;
+        default:
+            if (m_cursor)
+                [m_cursor retain];
```
eXpl0it3r, line 94: If I understand the [memory model](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/MemoryMgmt.html) of Obj-C correctly, this is not guaranteed to work, as the `release` at the top of the function, will allow the releasing of the cursor immediately, so one wouldn't be able to `retain` it here.

JonnyPtn: Yeah, I think the correct approach here is to set the member to nil after releasing it, as if it's deallocated we end up with a dangly pointer, have got a PR incoming

eXpl0it3r:
Superseded by #1736

Issue #1671, opened 2020-06-04
Mouse grabbing acts weirdly


Unarelith:
## Subject of the issue

The mouse is sometimes grabbed even though mouse grabbing is disabled.

## Your environment

* ArchLinux / i3
* Latest release and master both have this issue
* GCC 10.1.0

## Steps to reproduce

```cpp
#include <SFML/Window.hpp>

int main(int, char **) {
    sf::Window window;
    window.create(sf::VideoMode(800, 600), "SFML", sf::Style::Close);
    window.setPosition({1920 / 2 - 800 / 2, 1080 / 2 - 600 / 2});

    bool running = true;
    while (running) {
        sf::Event event;
        while (window.pollEvent(event)) {
            if (event.type == sf::Event::Closed || (event.type == sf::Event::KeyPressed && event.key.code == sf::Keyboard::Escape)) {
                running = false;
            }
            else if (event.type == sf::Event::KeyPressed) {
                if (event.key.code == sf::Keyboard::G)
                    window.setMouseCursorGrabbed(true);
                else if (event.key.code == sf::Keyboard::H)
                    window.setMouseCursorGrabbed(false);
            }
        }

        window.display();
    }

    window.close();

    return 0;
}
```

- Try to move the mouse in and out of the window, it should work
- Then press G to grab the mouse, you shouldn't be able to move the mouse out of the window
- Now press H to release the mouse, you should be able to move the mouse out of the window

## Expected behavior

The mouse should not be grabbed at all after `H` is pressed.

## Actual behavior

In my case, after pressing H, I'll be able to move out of the window only once.
**The second time it will act as if the mouse was grabbed.**

wyatthuckaby:
I was able to reproduce on PopOs 20.10 but with almost the opposite experience. The cursor was only grabbed the first time i pressed G. After pressing H, then G again, (turning off grabbing and turning it on again), the cursor was NOT grabbed. The cursor wont grab again, no matter how many times I attempt Grab or toggle grab on and off.

eXpl0it3r:
Should be fixed with #1722 

Feel free to comment/reopen in case the issue remains.

PR #1722, opened 2020-12-01
Fixed setMouseCursorGrabbed on Linux

Franckltung:
Description: Fixed the issue https://github.com/SFML/SFML/issues/1671 by setting m_cursorGrabbed to false when you disable grabbing. Without this change, every time we receive a FocusIn event (When you unfocus and refocus the window) it reconfines the mouse to the window even if setMouseCursorGrabbed was set to false. I tested with the code sample used in the issue and the problem was resolved.


binary1248 (approved):

eXpl0it3r:

eXpl0it3r:
Thanks a lot of spotting the issue and providing a fix! 🙂

PR #1723, opened 2020-12-05
NetBSD support proposal.

devnexen:


src/SFML/Window/CMakeLists.txt
```diff
@@ -124,6 +124,13 @@ elseif(SFML_OS_LINUX OR SFML_OS_FREEBSD OR SFML_OS_OPENBSD)
             ${SRCROOT}/OpenBSD/JoystickImpl.cpp
             ${SRCROOT}/OpenBSD/JoystickImpl.hpp
         )
+    elseif(SFML_OS_NETBSD)
+        set(PLATFORM_SRC
+            ${PLATFORM_SRC}
+	    ${SRCROOT}/NetBSD/JoystickImpl.cpp
+	    ${SRCROOT}/NetBSD/JoystickImpl.hpp
```
eXpl0it3r, line 131: Wrong indentation

src/SFML/Window/NetBSD/JoystickImpl.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+namespace
+{
+    std::map<unsigned int, std::string> plugged;
+    std::map<int, std::pair<int, int> > hatValueMap;
+
+    bool isJoystick(const char *name)
+    {
+        int file = ::open(name, O_RDONLY | O_NONBLOCK);
+	int id;
```
eXpl0it3r, line 54: Use spaces instead of tabs

src/SFML/Window/NetBSD/JoystickImpl.cpp
```diff
@@ -0,0 +1,349 @@
...
+        if (!desc)
+        {
+            ::close(file);
+            return false;
+        }
+
+	if (ioctl(file, USB_GET_REPORT_ID, &id) < 0) {
+		::close(file);
+		return false;
+	}
```
eXpl0it3r, line 70: Use spaces instead of tabs

src/SFML/Window/NetBSD/JoystickImpl.cpp
```diff
@@ -0,0 +1,349 @@
...
+            {
+                ::close(m_file);
+                return false;
+            }
+
+            // And the id
+	    if (ioctl(m_file, USB_GET_REPORT_ID, &m_id) < 0) {
+                ::close(m_file);
+                return false;
+	    }
```
eXpl0it3r, line 223: Use spaces instead of tabs

src/SFML/Window/NetBSD/JoystickImpl.cpp
```diff
@@ -0,0 +1,349 @@
...
+            // Get the report descriptor
+            m_desc = hid_get_report_desc(m_file);
+            if (!m_desc)
+            {
+                ::close(m_file);
+                return false;
+            }
+
+            // And the id
+	    if (ioctl(m_file, USB_GET_REPORT_ID, &m_id) < 0) {
```
eXpl0it3r, line 220: `{` should be on a new line

eXpl0it3r (changes requested):
I commented on some small stylistic changes.

As someone with no knowledge or experience with NetBSD, how would I go about testing this?

cmake/Config.cmake
```diff
@@ -35,6 +35,10 @@ elseif(CMAKE_SYSTEM_NAME MATCHES "^OpenBSD$")
     set(SFML_OS_OPENBSD 1)
     # don't use the OpenGL ES implementation on OpenBSD
     set(OPENGL_ES 0)
+elseif(CMAKE_SYSTEM_NAME MATCHES "^NetBSD$")
+	set(SFML_OS_NETBSD 1)
+	# don't use the OpenGL ES implementation on NetBSD
```
eXpl0it3r, line 40: Missed one, sorry: Use spaces instead of tab

eXpl0it3r (approved):

eXpl0it3r:

Bromeon:
Is there any difference in behavior regarding the different BSD distributions?
If not, we might think about collapsing the config and #define constants.

devnexen:
Mainly in term of joystick support in this case.

Bromeon:
Basically, your PR is very similar to the [existing FreeBSD joystick implementation](https://github.com/SFML/SFML/blob/master/src/SFML/Window/FreeBSD/JoystickImpl.cpp), in small differences e.g. how to get joystick ID. Meanwhile, the [OpenBSD implementation](https://github.com/SFML/SFML/blob/master/src/SFML/Window/OpenBSD/JoystickImpl.cpp) is an empty stub.

But in this case I'm not sure if it makes things more complicated to unify common code rather than duplicating it. The only thing I could imagine is if there is some `Window/BSD` folder with shared code -- not sure if it's worth it.

By the way, do you happen to know if the same code can be used on OpenBSD?

devnexen:
> 
> By the way, do you happen to know if the same code can be used on OpenBSD?

It might ... maybe the only difference is the ioctl lookup id, just a glance in their code tree tough.

eXpl0it3r:
I my opinion, code duplication is okay, if it's for the sake of isolating independent contexts/modules/etc.
Maybe there's a way to make share more code between the different BSD versions, but I think that should then maybe be done as a separate PR.

eXpl0it3r:
Thank you for the contribution! 🙂

PR #1724, opened 2020-12-06
Thread-safe sf::err()

Bromeon:
Very naive approach for fixing second problem in #1711.

`sf::Err` is not thread-safe, and since it's used as a global logger across many different classes, this can quickly lead to problems when the user doesn't have a single-threaded SFML application.

This implementation is not ready for merging, we need to discuss at least:
* how `sf::ThreadLocalPtr` should be used properly
* how to clean up (currently, there is a memory leak for each thread)

### DIFF
--- a/src/SFML/System/Err.cpp
+++ b/src/SFML/System/Err.cpp
@@ -26,6 +26,8 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/System/Err.hpp>
+#include <SFML/System/ThreadLocalPtr.hpp>
+#include <SFML/System/NonCopyable.hpp>
 #include <streambuf>
 #include <cstdio>
 
@@ -93,6 +95,23 @@ private:
         return 0;
     }
 };
+
+// Groups a std::ostream with its std::streambuf
+struct ThreadLocalErr : sf::NonCopyable
+{
+    DefaultErrStreamBuf buffer;
+    std::ostream stream;
+
+    ThreadLocalErr() :
+    buffer(),
+    stream(&buffer)
+    {
+    }
+};
+
+// This per-thread variable holds the current instance of DefaultErrStreamBuf
+sf::ThreadLocalPtr<ThreadLocalErr> currentErr(NULL);
+
 }
 
 namespace sf
@@ -100,11 +119,12 @@ namespace sf
 ////////////////////////////////////////////////////////////
 std::ostream& err()
 {
-    static DefaultErrStreamBuf buffer;
-    static std::ostream stream(&buffer);
+    if (currentErr == NULL)
+    {
+        currentErr = new ThreadLocalErr();
+    }
 
-    return stream;
+    return currentErr->stream;
 }
-
 
 } // namespace sf
--- a/src/SFML/System/Err.cpp
+++ b/src/SFML/System/Err.cpp
@@ -26,10 +26,13 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/System/Err.hpp>
+#include <SFML/System/Lock.hpp>
+#include <SFML/System/Mutex.hpp>
 #include <SFML/System/ThreadLocalPtr.hpp>
 #include <SFML/System/NonCopyable.hpp>
 #include <streambuf>
 #include <cstdio>
+#include <vector>
 
 
 namespace
@@ -96,6 +99,17 @@ private:
     }
 };
 
+// No-op buffer
+class NullBuffer : public std::streambuf
+{
+public:
+
+    int overflow(int character)
+    {
+        return character;
+    }
+};
+
 // Groups a std::ostream with its std::streambuf
 struct ThreadLocalErr : sf::NonCopyable
 {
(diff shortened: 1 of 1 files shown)


Bromeon:
@dec05eba if you find the time, could you check if this branch would also fix the problem you had?

dec05eba:
@Bromeon Confirmed that the crash is gone in this branch

Bromeon:
Thank you for testing!

**@SFML Team:**
Before we can merge this, we would need to address the above-mentioned points:

>   * how `sf::ThreadLocalPtr` should be used properly
>   * how to clean up (currently, there is a memory leak for each thread)

Bromeon:
Does anyone have a good idea how to address this without memory leak?

Bromeon:
Experimented a bit more with this.
Newest version:

1. Every time `sf::err()` is used from a new thread, a `ThreadLocalErr` instance is lazily initialized and assigned to the `ThreadLocalPtr`. Once the thread has its instance, it is reused.
1. To counter the memory leak, I maintain a global GC list with all thread-local instances. I register an `atexit` hook that iterates through the list, deleting each instance.
1. To make sure that `sf::err()` accesses *after* deletion don't wreak havoc, I return a no-op stream. Since I can't rely on global variables here (we're in the middle of global exit), I leak memory. This is a last resort that shouldn't happen anyway.
1. The book-keeping variables, which exist once globally, are protected by a mutex.

At this point all that is quite hacky, but so are the rules of C++ global variables and initialization order. I'm only wondering if we shouldn't just leave everything as-is and use a mutex. But let's see how it works out.

---

Now we have a small example code for testing, but I don't get it to work -- not even the version that the issue author @dec05eba said would be OK. So I'm stuck and it would be nice if someone else could also look into it.

<details><summary>Reproducing code (?)</summary>

I used the one @dec05eba provided [here](https://github.com/SFML/SFML/issues/1711#issuecomment-713945954), just changed the array to a vector and made the variables local.

For some reason I always get a stack overflow (in both my commits and the original SFML version).
Does that happen for anyone else?

```cpp
#include <SFML/Graphics/Image.hpp>
#include <thread>

int main()
{
    const int num_images = 3;
    const std::string image_paths[3] = { "working-image.png", "non-existing-image.png", "Tennis.cpp" };

    const int num_threads = 4;
    std::vector<std::thread> threads;

    for(int i = 0; i < num_threads; ++i) {
        threads.emplace_back([&]() {
            while(true) {
                for(int i = 0; i < num_images; ++i) {
                    sf::Image image;
                    image.loadFromFile(image_paths[i]);
                }
            }
        });
    }

    std::this_thread::sleep_for(std::chrono::seconds(10000000));
    return 0;
}
```
</details>

Bromeon:
Any feedback?

binary1248:
My opinion is that this change just shifts the problem from one place to another. Global initialization was/is/will always be a tricky subject in C++. While the problem has gotten less bad (yes... it still isn't good) with modern C++ (e.g. `thread_local` which is just compiler syntactic sugar/magic for doing things that are also possible by hand using platform-specific methods if you really want to), the problem hasn't gone away.

Over the years I have just trained myself to not access any non-POD, non-zero-initializable data that lives in namespace scope in my own code before `main()` starts. Whenever possible, complex initialization has to happen "downstream" from `main()` which allows me to specify deterministic construction/destuction ordering. Messing around with objects with complex initialization in global scope whose initialization order is not guaranteed is the literal definition of undefined behaviour. The same applies to destruction ordering even though it is less talked about. Even though it might not lead to noticeable effects in most cases, in my book all UB including _relying on the global initialization order_ (i.e. accessing before `main()`) of complex objects should always be avoided at any cost. If we cannot guarantee correctness of our code (even if non-/correctness has no visible side-effects) then the program is not fit for distribution, which is ultimately what most people aim for.

In the current changeset, `currentErr` and `activePtrs` (and `mutex` if you go full pedantic mode, `std::mutex`'s constructor is actually `constexpr` for this very reason) are technically non-POD with complex initialization (even relying in each case on dynamic allocation) and they are in namespace scope. In the unlucky scenario these objects are not yet initialized by the time `err()` is called you will have similar problems to what is possible now, just shifted to another location. An easy fix would be to lazily, dynamically allocate these objects as well before they are first accessed. Pointers are PODs and if they can be statically initialized to `NULL`, they will land in the .data segment and are guaranteed to have those values when the process image is loaded and before any code is executed.

I haven't evaluated the effort required, but I think instead of piling one workaround on top of another, we should just adopt air-tight habits and avoid indirect calls to `err()` from SFML library code in namespace scope whenever it is feasible.

Bromeon:
So, how do you want me to implement this?

A discussed, I can go with the assumption that no SFML code (i.e. `sf::err()`) will be run before or after `main()`, that would simplify quite a bit. Namely:
* the mutex will always be live
* `std::atexit()` is guaranteed to be run _after_ the last `sf::err()` call
* I don't need any `NullBuffer` or any logic that deals with after-main `sf::err()` calls

Is the assumption always valid? Probably not.
Is it better than the current, **broken** implementation? Yes.

Bromeon:
Could I get some feedback on this?
Do we want to port it to `2.6.x`? I would say it's worth fixing for SFML 2.
Also, if we target `master`, https://github.com/SFML/SFML/pull/1863 will lead to extra conflicts, because we waited too long with this PR.

I would say we go [with the last suggestion](https://github.com/SFML/SFML/pull/1724#issuecomment-958099762) -- assumption no `sf::err()` will be called before/after `main()`.

eXpl0it3r:
If you think it's important enough for 2.6, then lets add it there.
I personally consider any application that uses SFML objects at global scope invalid, as for most classes it really is UB and very often crashes with VS. Of course `sf::Image` and some other classes are pretty standalone, but even then using them non-globally is a better option.
Will the implementation with such an assumption lead to UB at global scope or explicitly crash?

For me personally, any solution that improves the situation, especially for multi-threading, seems okay.

Bromeon:
> I personally consider any application that uses SFML objects at global scope invalid, as for most classes it really is UB and very often crashes with VS.

As mentioned further above, that's not the issue -- we have a race condition, when using SFML objects only in `main()`.

eXpl0it3r:
> It's not possible to change the target branch of a PR, is it? Otherwise I'd need to create a new one.

Of course it is, just click "edit" at the top and there you can change it. Of course need to make sure not to include any changes from master

eXpl0it3r:
I think it's a lot easier to just make a decision to go one way, make assumptions that seem plausible and then based on that having a discussion, than to try and brain storm solutions on a GitHub issue. 😄 
The assumption makes sense to me, but as your explainers after my comments show, I seem to not fully understand the situation... 😅

Bromeon:
Not sure if this is still relevant, but it went nowhere. If someone still finds it useful, feel free to integrate the commits somewhere else :slightly_smiling_face:

PR #1725, opened 2020-12-08
[sprite] don't set default rect before the user provided one

vfjpl:
Small Optimisation. Using second constructor with user provided rextureRect we first create the default one(by setTexture method) and then overwrite it with the user provided one. With this change we simply first set user provided textureRect and then set the texture which prevents unnecesary creating the default one.

Please review.

Cheers.

### DIFF
--- a/src/SFML/Graphics/Sprite.cpp
+++ b/src/SFML/Graphics/Sprite.cpp
@@ -55,7 +55,9 @@ Sprite::Sprite(const Texture& texture, const IntRect& rectangle) :
 m_texture    (NULL),
 m_textureRect()
 {
+    // Compute the texture area
     setTextureRect(rectangle);
+    // Assign texture
     setTexture(texture, false);
 }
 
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
Can you provide an example that explains the use case here?

Also have you seen the code in `setTexture`? 

https://github.com/SFML/SFML/blob/4cbb34ddb4487a466f158712043c0ac8ef3eba05/src/SFML/Graphics/Sprite.cpp#L66-68

vfjpl:
Also have you seen the code in setTexture? yes, this is why I did this.

Exapmple is simple
`sf::Sprite newSprite(texture, userRectangle);`

Before sprite would create default internal rectangle, set texture, and then create new internal rectangle based on `userRectangle`.
Now it create new internal rectangle based on `userRectangle `and then sets the texture.

eXpl0it3r:
Ah okay, that seems to make sense to me then. Somehow had a hard time understanding the purpose. 😄 

@LaurentGomila any reason why we didn't do it this way from the beginning?

eXpl0it3r (approved):

LaurentGomila:
I don't really remember. Just a guess: the `resetRect` argument may have been added after the constructors, and we didn't really thought about this optimization.

vfjpl:
@eXpl0it3r @LaurentGomila Potential problem seems to be when the user want to set rect to `sf::IntRect(0, 0, 0, 0)`
Like:
`sf::Sprite newSprite(texture, `sf::IntRect(0, 0, 0, 0));`

before:
sprite would create internal rectangle based on texture size, set texture, and then create new internal rectangle based on userRectangle.
now:
sprite would create create new internal rectangle based on userRectangle, set texture and then create internal rectangle based on texture size (because rectangle the internal rectangle is all "0").

I'm not sure if there are any legitimate usages for setting rect to all 0 but it's something I didn't thought of yesterday.
Maybe it would be a bug in user code so we have "saved" him?

JonnyPtn (approved):

PR #1726, opened 2020-12-09
Use enum class instead of plain enum for PrimitiveType

johnBuffer:

----

## Description

Use enum class instead of plain enum for PrimitiveType as it is recommended. However this breaks retrocompatibilty since `sf::VertexArray` will have to be declared using `PrimitiveType::Points` for example.


src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -706,7 +706,7 @@ void RenderTarget::drawPrimitives(PrimitiveType type, std::size_t firstVertex, s
     // Find the OpenGL primitive type
     static const GLenum modes[] = {GL_POINTS, GL_LINES, GL_LINE_STRIP, GL_TRIANGLES,
                                    GL_TRIANGLE_STRIP, GL_TRIANGLE_FAN, GL_QUADS};
-    GLenum mode = modes[type];
+    GLenum mode = modes[int(type)];
```
Bromeon, line 709: should probably be `static_cast<std::size_t>(type)`

Bromeon (commented):
Why did you close it?

This would be a good change, but requires updating SFML to a newer standard. Since it breaks compatibility, it's only possible to have it for SFML 3.

[Upgrading SFML to C++17 is being planned.](https://github.com/SFML/SFML/wiki/SD%3A-CPP-17-Support)

johnBuffer:
I closed it because the checks didn't pass thinking that reopening later when fixed would be better.

eXpl0it3r:
Superseded by #2286

PR #1731, opened 2020-12-19
Preserve the drop shadow for system cursors on Windows

StrikerX3:

----

## Description

`CopyCursor` causes the cursor to lose the drop shadow effect. The only reason it was being used was so that it could keep an `HCURSOR` that could be destroyed later, for consistent usage in `release()`. I simply added a flag to indicate whether the cursor was loaded from a system cursor or a bitmap and a check on `release()` to only destroy the `HCURSOR` on the latter case.

## Tasks

* [x] Tested on Windows

(these changes do not affect any other OS)

## How to test this PR?

Run the code below and move the mouse over the window. It should change to a hand and honor the system-wide drop shadow setting.

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    window.setFramerateLimit(60);
        
    sf::Cursor cursor;
    cursor.loadFromSystem(sf::Cursor::Type::Hand);

    while (window.isOpen()) {
        sf::Event event;
            if (event.type == sf::Event::Closed) window.close();
        }

        window.display();
    }
}
```

### DIFF
--- a/src/SFML/Window/Win32/CursorImpl.cpp
+++ b/src/SFML/Window/Win32/CursorImpl.cpp
@@ -36,7 +36,8 @@ namespace priv
 
 ////////////////////////////////////////////////////////////
 CursorImpl::CursorImpl() :
-m_cursor(NULL)
+m_cursor(NULL),
+m_systemCursor(false)
 {
     // That's it.
 }
@@ -118,6 +119,7 @@ bool CursorImpl::loadFromPixels(const Uint8* pixels, Vector2u size, Vector2u hot
 
     // Create the cursor
     m_cursor = reinterpret_cast<HCURSOR>(CreateIconIndirect(&cursorInfo));
+    m_systemCursor = false;
 
     // The data has been copied into the cursor, so get rid of these
     DeleteObject(color);
@@ -166,8 +168,9 @@ bool CursorImpl::loadFromSystem(Cursor::Type type)
         case Cursor::NotAllowed:             shape = IDC_NO;          break;
     }
 
-    // Create a copy of the shared system cursor that we can destroy later
-    m_cursor = CopyCursor(LoadCursor(NULL, shape));
+    // Get the shared system cursor and make sure not to destroy it
+    m_cursor = LoadCursor(NULL, shape);
+    m_systemCursor = true;
 
     if (m_cursor)
     {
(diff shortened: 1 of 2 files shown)


eXpl0it3r:
Maybe you can give brief description of the fix you did here, as I may be misunderstanding something.

This seems more of a workaround than an actual fix, plus I think you might be leaking some memory or handle.

StrikerX3:
The original code created a copy of a shared cursor with `CopyCursor`, which lacks the shadow effect added by Windows if the user has enabled pointer shadows (in the Mouse properties window, see [here](https://www.thewindowsclub.com/enable-disable-mouse-pointer-shadow-windows)), whereas the original (shared) cursor loaded by `LoadCursor` retains the effect.

There's no memory/handle leak. [`LoadCursor` loads a shared system cursor unless it was already loaded](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadcursora#remarks) and [the documentation for `DestroyCursor` explicitly states that you should not destroy a shared cursor created by `LoadCursor`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-destroycursor#remarks).

As explained in the description above, my fix was to keep track of whether the cursor was loaded from a system cursor or a bitmap through the newly added `m_systemCursor` flag. Only the latter should be destroyed, which is checked in `CursorImpl::release()`. This makes the cursor behavior consistent across all Windows applications, i.e., the default Windows cursors don't lose the shadow effect when hovering an SFML application.

binary1248:
Wouldn't it be simpler to just not assign the shared cursor to `m_cursor` than to track if `m_cursor` contains a shared cursor via `m_systemCursor`? In the case a shared cursor is loaded and `m_cursor` stays `NULL`, the current `release()` implementation wouldn't have to be changed.

StrikerX3:
That will not work because `WindowImplWin32::setMouseCursor` uses the `m_cursor` field to set the cursor. Leaving it `NULL` causes the cursor to disappear.

binary1248:
Hmm... yes... now I notice I overlooked the friend class declaration.

binary1248 (approved):

binary1248:

eXpl0it3r:

eXpl0it3r:
Thank you for this fix! 🙂

PR #1732, opened 2021-01-03
Slight increase performance and code refactor

GermanAizek:

----

## Tasks

* [X] Tested on Linux
* [X] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Performed a check with minimal build.

### DIFF
--- a/src/SFML/Network/Ftp.cpp
+++ b/src/SFML/Network/Ftp.cpp
@@ -366,7 +366,7 @@ Ftp::Response Ftp::sendCommand(const std::string& command, const std::string& pa
 {
     // Build the command string
     std::string commandStr;
-    if (parameter != "")
+    if (!parameter.empty())
         commandStr = command + " " + parameter + "\r\n";
     else
         commandStr = command + "\r\n";
--- a/src/SFML/Window/WindowImpl.cpp
+++ b/src/SFML/Window/WindowImpl.cpp
@@ -181,7 +181,6 @@ void WindowImpl::processJoystickEvents()
         // Copy the previous state of the joystick and get the new one
         JoystickState previousState = m_joystickStates[i];
         m_joystickStates[i] = JoystickManager::getInstance().getState(i);
-        JoystickCaps caps = JoystickManager::getInstance().getCapabilities(i);
 
         // Connection state
         bool connected = m_joystickStates[i].connected;
@@ -199,6 +198,8 @@ void WindowImpl::processJoystickEvents()
 
         if (connected)
         {
+            JoystickCaps caps = JoystickManager::getInstance().getCapabilities(i);
+
             // Axes
             for (unsigned int j = 0; j < Joystick::AxisCount; ++j)
             {
--- a/src/SFML/Audio/SoundFileWriterFlac.cpp
+++ b/src/SFML/Audio/SoundFileWriterFlac.cpp
@@ -39,7 +39,7 @@ namespace priv
 ////////////////////////////////////////////////////////////
 bool SoundFileWriterFlac::check(const std::string& filename)
 {
-    std::string extension = filename.substr(filename.find_last_of(".") + 1);
+    std::string extension = filename.substr(filename.find_last_of('.') + 1);
     std::transform(extension.begin(), extension.end(), extension.begin(), ::tolower);
 
     return extension == "flac";
--- a/src/SFML/Audio/SoundFileWriterOgg.cpp
+++ b/src/SFML/Audio/SoundFileWriterOgg.cpp
@@ -40,7 +40,7 @@ namespace priv
 ////////////////////////////////////////////////////////////
 bool SoundFileWriterOgg::check(const std::string& filename)
 {
-    std::string extension = filename.substr(filename.find_last_of(".") + 1);
+    std::string extension = filename.substr(filename.find_last_of('.') + 1);
     std::transform(extension.begin(), extension.end(), extension.begin(), ::tolower);
 
     return extension == "ogg";
--- a/src/SFML/Audio/SoundFileWriterWav.cpp
+++ b/src/SFML/Audio/SoundFileWriterWav.cpp
@@ -77,7 +77,7 @@ namespace priv
 ////////////////////////////////////////////////////////////
 bool SoundFileWriterWav::check(const std::string& filename)
 {
-    std::string extension = filename.substr(filename.find_last_of(".") + 1);
+    std::string extension = filename.substr(filename.find_last_of('.') + 1);
     std::transform(extension.begin(), extension.end(), extension.begin(), ::tolower);
 
     return extension == "wav";
--- a/src/SFML/Network/Http.cpp
+++ b/src/SFML/Network/Http.cpp
@@ -327,7 +327,7 @@ void Http::setHost(const std::string& host, unsigned short port)
     {
         // HTTPS protocol -- unsupported (requires encryption and certificates and stuff...)
         err() << "HTTPS protocol is not supported by sf::Http" << std::endl;
-        m_hostName = "";
+        m_hostName.clear();
         m_port     = 0;
     }
     else


src/SFML/Graphics/Font.cpp
```diff
@@ -731,7 +732,7 @@ IntRect Font::findGlyphRect(Page& page, unsigned int width, unsigned int height)
             continue;
 
         // Check if there's enough horizontal space left in the row
-        if (width > page.texture.getSize().x - it->width)
+        if (width > size.x - it->width)
```
Bromeon, line 735: [`sf::Texture::getSize()` returns a precomputed variable](https://github.com/SFML/SFML/blob/a2626365908ded901fa5aef9ecf4d3dacbd03f47/src/SFML/Graphics/Texture.cpp#L310-L313), so the most you can win is an extra function call, which could be inlined anyway.

I would definitely not add this change for performance reasons, more for the sake of avoiding repetition. But then the variable should be named `pageTextureSize` or so for clarity, and then you have again almost the same verbosity as before. In addition, it may require the reader to look up the declaration of this variable further up to see what it actually does.

So: it _can_ be changed, not sure the pros outweight the cons though.

src/SFML/Graphics/Font.cpp
```diff
@@ -770,7 +771,7 @@ IntRect Font::findGlyphRect(Page& page, unsigned int width, unsigned int height)
         }
 
         // We can now create the new row
-        page.rows.push_back(Row(page.nextRow, rowHeight));
+        page.rows.emplace_back(Row(page.nextRow, rowHeight));
```
Bromeon, line 774: `emplace_back()` can't be used in C++98, and is meant for cases where arguments are passed to the constructor. In this case you would just invoke the copy/move constructor.

`push_back()` has a `T&&` overload, so this would not be faster either.

src/SFML/Graphics/Text.cpp
```diff
@@ -48,7 +48,7 @@ namespace
     }
 
     // Add a glyph quad to the vertex array
-    void addGlyphQuad(sf::VertexArray& vertices, sf::Vector2f position, const sf::Color& color, const sf::Glyph& glyph, float italicShear, float outlineThickness = 0)
+    void addGlyphQuad(sf::VertexArray& vertices, const sf::Vector2f& position, const sf::Color& color, const sf::Glyph& glyph, float italicShear, float outlineThickness = 0)
```
Bromeon, line 51: Please don't change this, it's literally a `double` being copied and makes code more wordy without benefit.

I even think that all the APIs in SFML should use `sf::Vector2f` instead of `const sf::Vector2f&` for shorter IDE suggestions and documentation.

src/SFML/Window/GlContext.cpp
```diff
@@ -261,7 +261,7 @@ namespace
                 while (*extensionString && (*extensionString != ' '))
                     extensionString++;
 
-                extensions.push_back(std::string(extension, extensionString));
+                extensions.emplace_back(std::string(extension, extensionString));
```
Bromeon, line 264: see above

src/SFML/Window/WindowImpl.cpp
```diff
@@ -199,6 +198,8 @@ void WindowImpl::processJoystickEvents()
 
         if (connected)
```
Bromeon, line 199: This is an example of a good change, as it could (depending on platform) save an expensive joystick query. It probably doesn't have huge weight compared to other joystick/input operations, but it also _increases_ code clarity as it moves declarations more local, so win/win.

The only thing I hope is that this call did not on one platform have a side effect that would be required 😬 but in that case it would be a hidden bug, so don't worry.

Bromeon (changes requested):
Thanks for your contribution!

Some of the changes are controversial due to micro-optimization vs. code clarity trade-off. In that case I tend to prefer clearer code, unless the optimization is proven to have an impact. I've commented some of these places.

src/SFML/Window/WindowImpl.cpp
GermanAizek: @Bromeon fix it.

src/SFML/Window/GlContext.cpp
GermanAizek: @Bromeon fix it.

src/SFML/Graphics/Text.cpp
GermanAizek: @Bromeon fix it.
https://github.com/SFML/SFML/pull/1732/commits/ae2d032d5a33e7f63fb2355b64426eef870eb0e5

src/SFML/Graphics/Font.cpp
GermanAizek: @Bromeon fix it.
https://github.com/SFML/SFML/pull/1732/commits/ae2d032d5a33e7f63fb2355b64426eef870eb0e5

src/SFML/Graphics/Font.cpp
GermanAizek: @Bromeon fix it.
https://github.com/SFML/SFML/pull/1732/commits/ae2d032d5a33e7f63fb2355b64426eef870eb0e5

Bromeon (approved):
Looks good to me, feedback from someone else?

src/SFML/Network/Ftp.cpp
```diff
@@ -366,7 +366,7 @@ Ftp::Response Ftp::sendCommand(const std::string& command, const std::string& pa
 {
     // Build the command string
     std::string commandStr;
-    if (parameter != "")
+    if (!parameter.empty())
```
eXpl0it3r, line 369: `std::string::empty()` is C++11 as such can't be used for SFML 2.x

eXpl0it3r (approved):

src/SFML/Graphics/Font.cpp
```diff
@@ -747,11 +748,11 @@ IntRect Font::findGlyphRect(Page& page, unsigned int width, unsigned int height)
     if (!row)
     {
         int rowHeight = height + height / 10;
-        while ((page.nextRow + rowHeight >= page.texture.getSize().y) || (width >= page.texture.getSize().x))
+        while ((page.nextRow + rowHeight >= pageTextureSize.y) || (width >= pageTextureSize.x))
```
LaurentGomila, line 751: The texture size can change inside the loop, so it's wrong to always use the size that was stored at the beginning of the function.
Then the first replacement becomes almost usless, so I would just revert changes to this file completely.

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -162,7 +162,7 @@ m_cursorGrabbed   (false)
 
 
 ////////////////////////////////////////////////////////////
-WindowImplWin32::WindowImplWin32(VideoMode mode, const String& title, Uint32 style, const ContextSettings& /*settings*/) :
+WindowImplWin32::WindowImplWin32(const VideoMode& mode, const String& title, Uint32 style, const ContextSettings& /*settings*/) :
```
LaurentGomila, line 165: I don't think that passing a bunch of integers by value is the slow thing in constructing a new window, let's keep it simple 😉

LaurentGomila (changes requested):

src/SFML/Window/Win32/WindowImplWin32.cpp
GermanAizek: Fix it.
https://github.com/SFML/SFML/pull/1732/commits/19b0c3dffc583de96cb3dfedfa82d4d2aa122725

src/SFML/Graphics/Font.cpp
GermanAizek: Fix it.
https://github.com/SFML/SFML/pull/1732/commits/19b0c3dffc583de96cb3dfedfa82d4d2aa122725

LaurentGomila (approved):

src/SFML/Graphics/Font.cpp
Bromeon: In that case, would it make sense to write a comment
```cpp
// Note: texture size can change during the loop
```
, to avoid a future pitfall?

binary1248 (approved):

PR #1735, opened 2021-01-06
Remove Travis CI Integration.

MrZeusTheCoder:
Removes CI as discussed on the SFML Discord server by @eXpl0it3r .


eXpl0it3r (approved):
Travis is basically dead, after they forced everyone to use travis-ci.com and only provide a small number of free minutes to open source projects.
With GitHub Actions and our own CI, we don't really need this, so while it's currently still working, I'm happy to remove it now already.

PR #1736, opened 2021-01-06
Fix potential crash when changing cursor - don't release the NSCursor object if it's already been released

JonnyPtn:
Follow up from #1721 which isn't suitable as it may try to retain a dangling reference. We solve this by setting our member reference to nil after we've released it.

 ## Tasks
 * [x]  Tested on macOS

### DIFF
--- a/src/SFML/Window/OSX/CursorImpl.mm
+++ b/src/SFML/Window/OSX/CursorImpl.mm
@@ -89,27 +89,23 @@ bool CursorImpl::loadFromPixels(const Uint8* pixels, Vector2u size, Vector2u hot
 ////////////////////////////////////////////////////////////
 bool CursorImpl::loadFromSystem(Cursor::Type type)
 {
-    if (m_cursor)
-    {
-        [m_cursor release];
-        m_cursor = nil;
-    }
+    NSCursor* newCursor = nil;
 
     switch (type)
     {
         default: return false;
 
-        case Cursor::Arrow:           m_cursor = [NSCursor arrowCursor];               break;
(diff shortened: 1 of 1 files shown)


src/SFML/Window/OSX/CursorImpl.mm
```diff
@@ -71,7 +71,11 @@
 ////////////////////////////////////////////////////////////
 bool CursorImpl::loadFromPixels(const Uint8* pixels, Vector2u size, Vector2u hotspot)
 {
-    [m_cursor release];
+    if (m_cursor)
+    {
+        [m_cursor release];
+        m_cursor = nil;
+    }
```
eXpl0it3r, line 78: If one had a valid cursor before and the loading would fail for some reason, the user would end up with a different state.

I guess it can be argued either way. For an UX perspective, I think it would be nicer to retain the valid cursor if the loading failed, but I'm open either way.

JonnyPtn: Yeah in many ways that would be user error - if the load function fails then you shouldn't use the cursor (same as with other resources like audio/textures) but I do agree that it makes sense for the cursor object to maintain it's state if loading fails.

Think the only way to do that is to make a temp local copy of m_cursor which is released at the end of the method - will update this later

eXpl0it3r (changes requested):
As discussed the cursor should retain its previous state if the loading fails

eXpl0it3r (approved):

eXpl0it3r:

PR #1741, opened 2021-01-15
[Feature] Adding constexpr functionality to the library

fgallegosalido:
* [X] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before? Discussed in Discord.

----

## Description

Added constexpr support for a few classes (more will be added later).

It's important to note that constexpr functionality cannot be buil separately and then linked with the rest of the library. This is because the compiler must be able to see the definition of constexpr functionality in order to perform compile-time computations.

For that reason, definitions of constexpr functionality must be done inside the header file (just like templates), so many definitions that used to be in the corresponding `.cpp` file are now inside a `.inl` file (to be consistent with the rest of the library).

Another discussion is about turning some variables that are currently const into constexpr. This should be easy in general, but static class variables which type is the same as the class they're in (e.g.: `sf::Color::Blue`, `sf::Color::Red`, `sf::Time::Zero`, ...) cannot be made constexpr (at least I'm not aware of how to do it).

For the moment I'm leaving all the variable definitions inside the `.cpp` file.

## Tasks

* [X] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

This is the template example, but using a constexpr `sf::Vector2u` to intialize `sf::VideoMode` instead of using the values directly. The constexpr functionality should work out of the box, as long as the initialization is performed at compile time.

```cpp
#include <SFML/Graphics.hpp>
#include <SFML/System.hpp>

int main()
{
    // Just showing sf::Vector2, but should work for every constexpr functionality
    constexpr sf::Vector2u windowSize(1280, 720);
    sf::RenderWindow window(sf::VideoMode(windowSize.x, windowSize.y), "Minimal, complete and verifiable example");
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
        }

        window.clear();
        window.display();
    }
}
```


Bromeon:
Moving function definitions from .cpp to header files bears similar problems as with templates:
* May require dependencies to be exposed
* Increases compile times
* Changes require recompilation of all clients (probably OK since it's a library).

Also, the `constexpr` attribute is somewhat never-ending, it can be added to loads of functions without having a real impact.
So I would be for a more selective strategy regarding if and how we add this attribute. In the cases where no one benefits from `constexpr`, we should probably not add it, for the sake of decreased compile times (would need measurements).

Eventually, `constexpr` is an _expression of intent_ -- it signals to the user "you can use this symbol in a constant expression". That implies that removing `constexpr` is a breaking API change; thus it's not suited for more complex cases which might undergo change of implementation.

fgallegosalido:
I'm okay with not doing almost everything constexpr (even though we could make 90% of SFML constexpr). I did some that I felt would benefit from constexpr (except `sf::BlendMode`, maybe).

Some classes are obviously not really great to add constexpr to them, like classes that draw stuff or classes that manage windows. We should compile a list of functions, classes and member functions that can benefit from being constexpr.

Exposing internal stuff is something we risk by adding constexpr, but we have no other option if we want this functionality.

Bromeon:
> Exposing internal stuff is something we risk by adding constexpr, but we have no other option if we want this functionality.

Yep, and this is what we should analyze.
What do we need `constexpr` for?
* Performance?
* Usage of expressions in more contexts?
* Metaprogramming?
* Keep up with latest trends?

If it boils down to point 4, we might need to reconsider it 🙂

Also, the feature is more complex than might seem at first (see e.g. Chapter 3 [here](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2043r0.pdf)), and with C++20 and `constinit`/`consteval`, a whole new can of worms is opened.

TLDR: I'm not sure SFML is a typical library that can benefit from the added complexity (more verbose signatures, moving things to headers, possibly increased compile times), so we should carefully check the benefits.

fgallegosalido:
I agree that we shouldn't go crazy, but there's really no drawback in adding it to a few classes. I can speak for myself, and I've needed it a few times. Not that it was a must, but having a const variable when a constexpr is possible is not the best.

eliasdaler:
In my opinion SFML should handle more important things related to modern C++ first. 
Removing redundant classes (sf::Thread), using std::unique_ptr or std::unordered_map in some places, etc. etc. (more was discussed on SFML forums in development section)

`constexpr` doesn't really add much here, I think (it's more useful for places where you really needed `constexpr` expressions, e.g. std::tuple manipulation, std::array initialization, etc.). Just because we can doesn't mean we should.

khatharr:
The predefined color constants can be made constexpr with the following pattern:

`struct Foo {
  int a;
  constexpr Foo(int a) a(a) {}

  static const Foo Default;
};

constexpr Foo::Default{1}; //must be in the same header as the declaration
`

This will cause the compiler will treat Foo::Default as a constexpr.

vittorioromeo:
`sf::Vector<T>` is now `constexpr` thanks to #1903.

Bromeon:
Closing as discussed in #1909 🙂

Issue #1733, opened 2021-01-04
setUniform("texture2", tex) on RenderTexture flipped.


oomek:
## Subject of the issue

When I draw a RenderTexture with a shader that has an additional texture bound to it it's drawn flipped. Checked the forum, but haven't found anything similar. Still not sure whether it is a bug or not, so please don't shout at me.

## Your environment

* Windows 10
* Your version of SFML (git master)
* Your compiler and compiler version used VS2015

## Steps to reproduce

```cpp
#include <SFML/Graphics.hpp>

int main()
{
	sf::RenderWindow window;
	window.create( sf::VideoMode::getDesktopMode(), "Flip bug", sf::Style::Fullscreen );
	window.setVerticalSyncEnabled( true );
	window.setMouseCursorVisible( false );

	sf::Vector2u size = window.getSize();

	sf::Texture texture;
	texture.loadFromFile( "a.png" );
	texture.setSmooth( true );

	sf::Sprite texture_sprite;
	texture_sprite.setTexture( texture );
	texture_sprite.setScale( (float)size.x / texture.getSize().x, (float)size.y / texture.getSize().y / 2.0 );

	sf::RenderTexture render;
	render.create( size.x, size.y / 2 );

	sf::Sprite render_sprite;
	render_sprite.setTexture( render.getTexture() );
	render_sprite.setPosition( 0, size.y / 2 );

	sf::Shader shader;
	shader.loadFromFile( "shader.frag", sf::Shader::Fragment );
	shader.setUniform( "texture2", texture );

	while (window.isOpen())
	{
		sf::Event event;
		while ( window.pollEvent( event ))
		{
			if (( event.type == sf::Event::Closed )
				|| ( event.key.code == sf::Keyboard::Escape ))
			{
				window.close();
			}
		}

		render.clear();
		render.draw( texture_sprite );
		render.display();

		window.clear();
		window.draw( texture_sprite, &shader );
		window.draw( render_sprite, &shader );
		window.display();
	}

	return 0;
}
```
shader.frag
```cpp
// This shader draws texture on the left using RB channels and
// texture2 on the right using G channel
uniform sampler2D texture;
uniform sampler2D texture2;

void main()
{
	vec4 color1 = texture2D( texture, gl_TexCoord[0].xy * vec2( 2.0, 1.0 ) + vec2( 0.0, 0.0 ));
[...]

PR #1742, opened 2021-01-15
Fix for issue #1733

AKonia:
#### Description
Solves issue #1733, but needs checking from people, who possibly could explain why textures in RenderTexture::display() previously have set `m_texture.m_pixelsFlipped = true;`


binary1248:
This does not fix the problem.

See [my comment in the original issue](https://github.com/SFML/SFML/issues/1733#issuecomment-761708945).

PR #1743, opened 2021-01-16
Fixed copy assign operator in sf::Sound so it checks for self-assignment

fgallegosalido:

----

## Description

The copy-assign operator in `sf::Sound` doesn't use the copy-and-swap idiom. It's also modifying the member variable `m_buffer`. For those two reasons, it is necessary to check for self-assignment.

If we don't check for self-assignment, and we're assigning an object to itself, look what happens in this snippet:

```cpp
if (m_buffer)
{
    stop();
    m_buffer->detachSound(this);
    m_buffer = NULL;
}
```

If `this == &right`, we are basically setting `right.m_buffer` to `NULL` too, which is undesirable in this situation.

## Tasks

* [X] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Just create a `sf::Sound` and try to assign it to itself. It shouldn't modify itself.


src/SFML/Audio/Sound.cpp
```diff
@@ -160,6 +160,13 @@ Sound& Sound::operator =(const Sound& right)
     // the list of sound instances contained in the buffers and unnecessarily
     // destroy/create OpenAL sound sources
 
+    // As we don't use the copy-and-swap idiom here, we need to check for
+    // self-assignment to prevent right.m_buffer from being modified
+    // (which will happen if this == &right)
+    if(this == &right)
```
eXpl0it3r, line 166: Missing space: `if (...`

src/SFML/Audio/Sound.cpp
```diff
@@ -160,6 +160,13 @@ Sound& Sound::operator =(const Sound& right)
     // the list of sound instances contained in the buffers and unnecessarily
     // destroy/create OpenAL sound sources
 
+    // As we don't use the copy-and-swap idiom here, we need to check for
+    // self-assignment to prevent right.m_buffer from being modified
+    // (which will happen if this == &right)
```
eXpl0it3r, line 165: I think this can be reduced a bit, maybe something more like 

```cpp
// Handle self-assignment here, as no copy-and-swap idiom is being used
```

eXpl0it3r (changes requested):
Seems to be the recommended way in the [ISO C++ FAQ](https://isocpp.org/wiki/faq/assignment-operators#self-assignment-why). Thanks for catching it!

If possible, try to squash the commits into one. Don't worry if it's not up to date with the master branch, we can easily rebase & merge.

fgallegosalido:
Okay, I'm going to undo the commits and make just one commit

fgallegosalido:
If you don't mind, I'm going to close this pr and open a new one with the correct changes

PR #1744, opened 2021-01-16
Fixed copy assign operator in sf::Sound so it checks for self-assignment

fgallegosalido:

----

## Description

The copy-assign operator in sf::Sound doesn't use the copy-and-swap idiom. It's also modifying the member variable m_buffer. For those two reasons, it is necessary to check for self-assignment.

If we don't check for self-assignment, and we're assigning an object to itself, look what happens in this snippet:

```cpp
if (m_buffer)
{
    stop();
    m_buffer->detachSound(this);
    m_buffer = NULL;
}
```

If `this == &right`, we are basically setting `right.m_buffer` to `NULL` too, which is undesirable in this situation.

## Tasks

* [X] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Just create a sf::Sound and try to assign it to itself. It shouldn't modify itself.

### DIFF
--- a/src/SFML/Audio/Sound.cpp
+++ b/src/SFML/Audio/Sound.cpp
@@ -160,6 +160,10 @@ Sound& Sound::operator =(const Sound& right)
     // the list of sound instances contained in the buffers and unnecessarily
     // destroy/create OpenAL sound sources
 
+    // Handle self-assignment here, as no copy-and-swap idiom is being used
+    if (this == &right)
+        return *this;
+
     // Delegate to base class, which copies all the sound attributes
     SoundSource::operator=(right);
 


eXpl0it3r (approved):

binary1248 (approved):

binary1248:

fgallegosalido:
Talking about this little bug, I'm finding some similar bugs using clang-tidy along with this one. Some of them can be fixed for SFML 2.6.0, but most of them would benefit from C++17 features. Maybe we should discuss if adding static analyzers to the project can benefit the project itself.

If you think this is interesting enough to discuss, let me know to open a new issue.

eXpl0it3r:
We already have some [static analyzers](https://ci.sfml-dev.org/#/builders/14) setup, but I guess we could further improve it with clang tidy.

For stuff that doesn't require C++11 and beyond and doesn't modify the public APIs, feel free to open new PRs for or if there's something to discuss, an issue first.

For C++17 features I'm on and off working on a [solution design](https://github.com/SFML/SFML/wiki/SD%3A-CPP-17-Support) for SFML 3 to figure out what API improvements can be made. If you have anything specifically, it would be great if you could open a thread in the [SFML Development forum](https://en.sfml-dev.org/forums/index.php?board=45.0).

Issue #1642, opened 2020-02-20
Text kerning is not working


oomek:
## Subject of the issue

It seems like all modern fonts that use GPOS table for storing kerning tables are not displayed properly in SFML. The only fonts that have proper kerning applied are those with KERN table in "legacy" format.

## Your environment

* Windows 10
* git master
* WSL Ubuntu x-compiled with MXE packages
* Special compiler flags used: none

## Steps to reproduce

Download a modern ttf font like for example Barlow:
https://github.com/jpt/barlow
Draw in SFML and in Photoshop
Compare kerning of pairs with high kern values like:
Vo, Wo, To, VA, AV

## Expected behavior

Kerning should be applied like in Photoshop

## Actual behavior

Kerning not applied when the kerning table is stored in GPOS table

![image](https://user-images.githubusercontent.com/2974860/74931356-578aa700-53d7-11ea-935f-5e2e1f9bf0b7.png)

LaurentGomila:
For font rendering issues you should always first check the FreeType documentation, as we highly rely on it.

> FreeType only supports kerning via the (rather simple) ‘kern’ table. For the interpretation of kerning data in the (highly sophisticated) ‘GPOS’ table you need a higher-level library like ICU or HarfBuzz since it can be context dependent (this is, the kerning may vary depending on the position within a text string, for example).

oomek:
Noted.

oomek:
Would you accept a pr adding harfbuzz for glyph advance calculations?

LaurentGomila:
If using Harfbuzz, it would then be for the whole implementation of sf::Text, no need to limit to just this specific bit.

But it's quite a big dependency, I don't know if we really want that.

oomek:
I'm not sure it's even possible. Harfbuzz does not render glyphs as far as I know. I see it rather as they're complementary. HB even takes FT font instance as a parameter. I'm willing to try to add a proper kerning to SFML, we just need to find some compromise features/complexity beforehand.

LaurentGomila:
I never talked about rendering glyphs, sf::Text doesn't do that. As I see it, it would be sf::Text == Harfbuzz and sf::Font == FreeType.

oomek:
Yeah, but getKerning() function lives in font.cpp doesn't it ;) But I get what you mean, HarfBuzz calls would reside in sf::Text

eXpl0it3r:
[...]

PR #1746, opened 2021-01-20
Added auto hinting compensation to fix too wide gaps between small glyphs

oomek:

----

## Description

Forced Autohint requires left and right bearings to be taken into acccount during kerning calculation. Without this the gaps between small glyph sizes are twice as wide on a lot of fonts.

This PR is related to the issue https://github.com/SFML/SFML/issues/1642#issuecomment-763700187

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android


src/SFML/Graphics/Font.cpp
```diff
@@ -375,30 +376,37 @@ bool Font::hasGlyph(Uint32 codePoint) const
...
 
+        // Retrieve position compensation deltas generated by FT_LOAD_FORCE_AUTOHINT flag
+        float firstRsbDelta = getGlyph(first, characterSize, bold).rsb_delta;
+        float secondLsbDelta = getGlyph(second, characterSize, bold).lsb_delta;
+
         // Get the kerning vector
         FT_Vector kerning;
-        FT_Get_Kerning(face, index1, index2, FT_KERNING_DEFAULT, &kerning);
+        kerning.x = kerning.y = 0;
+        if( FT_HAS_KERNING(face))
```
eXpl0it3r, line 400: Should be `if (`

oomek: @eXpl0it3r Thanks for spotting that. I'm usually pedantic when it comes to formatting conventions.

include/SFML/Graphics/Glyph.hpp
```diff
@@ -52,6 +52,8 @@ class SFML_GRAPHICS_API Glyph
     // Member data
     ////////////////////////////////////////////////////////////
     float     advance;     //!< Offset to move horizontally to the next character
+    int       lsb_delta;   //!< Left offset after forced autohint. Internally used by getKerning()
+    int       rsb_delta;   //!< Right offset after forced autohint. Internally used by getKerning()
```
LaurentGomila, line 56: These two members don't follow our naming convention (should be `lsbDelta` and `rsbDelta`).

src/SFML/Graphics/Font.cpp
```diff
@@ -375,30 +376,37 @@ bool Font::hasGlyph(Uint32 codePoint) const
...
+float Font::getKerning(Uint32 first, Uint32 second, unsigned int characterSize, bool bold) const
 {
     // Special case where first or second is 0 (null character)
     if (first == 0 || second == 0)
         return 0.f;
 
     FT_Face face = static_cast<FT_Face>(m_face);
 
-    if (face && FT_HAS_KERNING(face) && setCurrentSize(characterSize))
+    if (face && setCurrentSize(characterSize))
```
LaurentGomila, line 387: Since this block now handles the "no kerning" case as well, the comment in the "else" block should be adjusted.

LaurentGomila (commented):
Looks ok, except some minor adjustments

src/SFML/Graphics/Font.cpp
oomek: Shall I just leave `invalid font` or provide something more meaningful?

oomek:
Regarding that added function parameter:
https://github.com/SFML/SFML/blob/d2406fcff21947e4ad125f40f1234787b40b2cee/include/SFML/Graphics/Font.hpp#L220
Shall I update the brief as well? If so, do we need to indicate that when different bold flag is used than already cached the glyph will be loaded?

eXpl0it3r:
> Shall I update the brief as well?

If you think the brief could need an update/change, then certainly.
What you certainly need to add is `/param` for the new parameter.

> If so, do we need to indicate that when different bold flag is used than already cached the glyph will be loaded?

Does the user of the function need to care about this implementation detail?
If only in some very special cases, then no, I wouldn't add that information.

oomek:
Maybe it's just me being paranoid, but I found 2 fonts out of 50 I tested with unusually narrow whitespace.
It's not relevant to kerning compensation, but how `whitespaceWidth` is calculated. Maybe it's also a good opportunity to address it aswell by clamping the whitespaceWidth to at least 2 pixels by:
```
    float whitespaceWidth = m_font->getGlyph(L' ', m_characterSize, isBold).advance;
    float letterSpacing   = ( whitespaceWidth / 3.f ) * ( m_letterSpacingFactor - 1.f );
    whitespaceWidth      += letterSpacing;
    if (whitespaceWidth < 2.f) whitespaceWidth = 2.f;
```
Here is the screenshot of the affected font before and after:
image
Please open it in 100% browser zoom to avoid any potential scaling.

src/SFML/Graphics/Font.cpp
LaurentGomila: Whatever. Even removing the comment would be fine.

oomek:
Would you like me to commit that space width compensation as well?

LaurentGomila:
Since it's a different issue, let's address it in another PR.

oomek:
Half of the CI tasks did not start. Do they have to be triggered manually?

eXpl0it3r:

eXpl0it3r:
> Half of the CI tasks did not start. Do they have to be triggered manually?
Our own CI setup only executes on demand on third-party PRs for security reasons.
I have since triggered them.

@LaurentGomila any additional comment for the PR?

LaurentGomila (approved):

Issue #1749, opened 2021-02-04
MacOS CMake Rejects arm64 arch


jkeller51:
## Subject of the issue

SFML/CMakeLists.txt contains this fatal error in the MacOS section:
```
    # only the default architecture (i.e. 64-bit) is supported
    if(NOT CMAKE_OSX_ARCHITECTURES STREQUAL "x86_64")
        message(FATAL_ERROR "Only 64-bit architecture is supported")
    endif()
```

However, when bypass this check I am able to build and run native arm64 just fine on a MacBook Pro with an M1 processor.

## Your environment

macOS Big Sur 11.2
Xcode 12.4
CMAKE_OSX_ARCHITECTURES = "x86_64;arm64"

## Suggested Solution
Split the CMAKE_OSX_ARCHITECTURES by semicolon and make sure each item is either x86_64 or arm64. I'm not a CMake wizard but I could maybe do this within a few days.

eXpl0it3r:
Fixed with #1750

PR #1750, opened 2021-02-08
macOS: allow CMAKE_OSX_ARCHITECTURES to be arm64

jkeller51:

----

## Description

Changed CMakeLists.txt to allow `CMAKE_OSX_ARCHITECTURES ` to be either x86_64, arm64, or both. This allows native builds on apple silicon.

This PR is related to the issue #1749 

## Tasks
* [x] Tested on macOS

## How to test this PR?

Pull the code on MacOS, enter some value into `CMAKE_OSX_ARCHITECTURES `, and configure.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -253,10 +253,12 @@ if(SFML_OS_MACOSX)
         endif()
     endif()
 
-    # only the default architecture (i.e. 64-bit) is supported
-    if(NOT CMAKE_OSX_ARCHITECTURES STREQUAL "x86_64")
-        message(FATAL_ERROR "Only 64-bit architecture is supported")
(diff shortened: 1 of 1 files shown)


MarioLiebisch:
I'm not really sure, I'd say it's better to just drop the whole check, since you could also compile/build both architectures into the same file (although I haven't tried that yet with SFML).

jkeller51:
Actually, this will allow that as well. I've been distributing my MacOS app as a "universal app" which contains both architectures into a single binary (all static/dynamic libs also build into a single binary containing two architectures).

eXpl0it3r (approved):
Verified on our brand new Mac Mini M1 🎉

PR #1093, opened 2016-05-30
Added missing sRgb getter and setter to RenderTexture.

binary1248:
Fixes #1092.

PR #1757, opened 2021-03-10
Support sRGB color space in RenderTexture

### LINKED ISSUE #1092: Drawing to RenderTexture with sRGB conversion enabled results in banded colors
Window sRGB capability is turned on and the texture is loaded with sRGB conversion.
See this thread for more details and code example to reproduce: http://en.sfml-dev.org/forums/index.php?topic=20358.0

Sakarah:
* [x] Does the fix fully solve the issue

----

## Description

This PR tries to fix #1092.

When the sf::ContextSettings asks for an sRGB capable buffer, set the rendered image to sRGB mode and convert pixels when drawing.

This is the choice of simplicity compared to actually support textures with more than 8 bit color depth: see my post in forum (https://en.sfml-dev.org/forums/index.php?topic=20358.msg175496#msg175496).

It differs from #1093 because now we try to both convert to sRGB space when writing to the texture and when decoding from the texture. This repairs the rendering on multisampled textures, and on non-FBO implementations.
It also repairs rendering for sRGB RenderTexture used along a non-sRGB window (or no window at all).

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

To test the PR, you should run the following code and check that it matches the [reference output](https://user-images.githubusercontent.com/7562593/115022631-ee5cd180-9ebd-11eb-9cee-b50d0dadba93.png) for all parameter combination.
You will need to following image: [Gradient.png](https://user-images.githubusercontent.com/7562593/115022505-c3727d80-9ebd-11eb-9054-b823007e1999.png)

GitHub's image compression makes it slightly hard to see, but it is normal if configurations `lS -> sT -> lW` and `sS -> lT -> sW` show color banding, the other configurations must not.
Enabling or disabling antialiasing should not change the output at all.

The code is adapted from the issue by @Fewes in the forum thread https://en.sfml-dev.org/forums/index.php?topic=20358.msg146448#msg146448.

```cpp
#include <SFML/Graphics.hpp>

void updateWindowTitle(sf::Window& window, bool bypass_texture, bool sRgbSprite, bool sRgbTexture, bool sRgbWindow, bool antialiasing)
{
    sf::String title;
    title += sRgbSprite ? "sS" : "lS";
    title += " -> ";
    if (!bypass_texture)
    {
        title += sRgbTexture ? "sT" : "lT";
        if (antialiasing)
            title += " (MSAA x2)";
        title += " -> ";
    }
    title += sRgbWindow ? "sW" : "lW";
    window.setTitle(title);
}

int main()
{
    bool bypass_texture = false;

    sf::RenderWindow window;
    sf::VideoMode videoMode;
    videoMode.width = 800;
    videoMode.height = 600;
    sf::ContextSettings windowContextSettings;
    windowContextSettings.sRgbCapable = true;

    window.create(videoMode, "sRGB testing", sf::Style::Default, windowContextSettings);

    sf::Texture spriteTexture;
    spriteTexture.setSrgb(true);
    spriteTexture.setSmooth(true);
    spriteTexture.loadFromFile("Gradient.png");

    sf::Sprite sprite;
    sprite.setTexture(spriteTexture);

    renderTexture.setSmooth(true);
    sf::ContextSettings textureContextSettings;
    textureContextSettings.sRgbCapable = true;
    renderTexture.create(videoMode.width, videoMode.height, textureContextSettings);

    sf::Sprite renderTextureDrawable;
    renderTextureDrawable.setTexture(renderTexture.getTexture());

    updateWindowTitle(window, bypass_texture, spriteTexture.isSrgb(), textureContextSettings.sRgbCapable, windowContextSettings.sRgbCapable, textureContextSettings.antialiasingLevel);

    while (window.isOpen())
    {
        // Poll events
        sf::Event event;
        {
            // Window closed
            if (event.type == sf::Event::Closed || (event.type == sf::Event::KeyPressed && event.key.code == sf::Keyboard::Key::Escape))
            {
                window.close();
            }
            {
                if (event.key.code == sf::Keyboard::Space)
                {
                    bypass_texture = !bypass_texture;
                }
                else if (event.key.code == sf::Keyboard::A)
                {
                    textureContextSettings.antialiasingLevel = textureContextSettings.antialiasingLevel ? 0 : 2;
                    renderTexture.create(videoMode.width, videoMode.height, textureContextSettings);
                }
                else if (event.key.code == sf::Keyboard::T)
                {
                    textureContextSettings.sRgbCapable = !textureContextSettings.sRgbCapable;
                    renderTexture.create(videoMode.width, videoMode.height, textureContextSettings);
                }
                else if (event.key.code == sf::Keyboard::S)
                {
                    spriteTexture.setSrgb(!spriteTexture.isSrgb());
                    spriteTexture.loadFromFile("Gradient.png");
                }
                else if (event.key.code == sf::Keyboard::W)
                {
                    windowContextSettings.sRgbCapable = !windowContextSettings.sRgbCapable;
                    window.create(videoMode, "sRGB testing", sf::Style::Default, windowContextSettings);
                }
                updateWindowTitle(window, bypass_texture, spriteTexture.isSrgb(), textureContextSettings.sRgbCapable, windowContextSettings.sRgbCapable, textureContextSettings.antialiasingLevel);
            }
        }

        // Clear render texture
        renderTexture.clear(sf::Color(0, 0, 0, 255));
        // Draw gradient sprite
        renderTexture.draw(sprite);
        // Finished drawing to render texture
        renderTexture.display();

        // Clear window
        window.clear(sf::Color(0, 0, 0, 255));
        if (bypass_texture)
        {
            window.draw(sprite);
        }
        else
        {
            // Draw render texture drawable
            window.draw(renderTextureDrawable);
        }

        // Finished drawing to the window
        window.display();
    }

    return 0;
}
```

### DIFF
--- a/src/SFML/Graphics/RenderTextureImplFBO.cpp
+++ b/src/SFML/Graphics/RenderTextureImplFBO.cpp
@@ -296,6 +296,7 @@ bool RenderTextureImplFBO::create(unsigned int width, unsigned int height, unsig
 #ifndef SFML_OPENGL_ES
 
             // Create the multisample color buffer
+            bool srgb = settings.sRgbCapable && GLEXT_texture_sRGB;
             GLuint color = 0;
             glCheck(GLEXT_glGenRenderbuffers(1, &color));
             m_colorBuffer = static_cast<unsigned int>(color);
@@ -305,7 +306,7 @@ bool RenderTextureImplFBO::create(unsigned int width, unsigned int height, unsig
                 return false;
             }
             glCheck(GLEXT_glBindRenderbuffer(GLEXT_GL_RENDERBUFFER, m_colorBuffer));
-            glCheck(GLEXT_glRenderbufferStorageMultisample(GLEXT_GL_RENDERBUFFER, settings.antialiasingLevel, GL_RGBA, width, height));
+            glCheck(GLEXT_glRenderbufferStorageMultisample(GLEXT_GL_RENDERBUFFER, settings.antialiasingLevel, srgb ? GL_SRGB8_ALPHA8_EXT : GL_RGBA, width, height));
 
             // Create the multisample depth/stencil buffer if requested
             if (settings.stencilBits)
--- a/src/SFML/Graphics/RenderTextureImplFBO.cpp
+++ b/src/SFML/Graphics/RenderTextureImplFBO.cpp
@@ -120,7 +120,8 @@ m_height            (0),
 m_context           (NULL),
 m_textureId         (0),
 m_multisample       (false),
-m_stencil           (false)
+m_stencil           (false),
+m_sRgb              (false)
 {
     Lock lock(mutex);
 
@@ -228,6 +229,8 @@ bool RenderTextureImplFBO::create(unsigned int width, unsigned int height, unsig
         if (settings.stencilBits && !GLEXT_packed_depth_stencil)
             return false;
 
+        m_sRgb = settings.sRgbCapable && GL_EXT_texture_sRGB;
+
 #ifndef SFML_OPENGL_ES
 
         // Check if the requested anti-aliasing level is supported
@@ -296,7 +299,6 @@ bool RenderTextureImplFBO::create(unsigned int width, unsigned int height, unsig
 #ifndef SFML_OPENGL_ES
 
             // Create the multisample color buffer
-            bool srgb = settings.sRgbCapable && GLEXT_texture_sRGB;
             GLuint color = 0;
             glCheck(GLEXT_glGenRenderbuffers(1, &color));
             m_colorBuffer = static_cast<unsigned int>(color);
@@ -306,7 +308,7 @@ bool RenderTextureImplFBO::create(unsigned int width, unsigned int height, unsig
                 return false;
             }
             glCheck(GLEXT_glBindRenderbuffer(GLEXT_GL_RENDERBUFFER, m_colorBuffer));
-            glCheck(GLEXT_glRenderbufferStorageMultisample(GLEXT_GL_RENDERBUFFER, settings.antialiasingLevel, srgb ? GL_SRGB8_ALPHA8_EXT : GL_RGBA, width, height));
+            glCheck(GLEXT_glRenderbufferStorageMultisample(GLEXT_GL_RENDERBUFFER, settings.antialiasingLevel, m_sRgb ? GL_SRGB8_ALPHA8_EXT : GL_RGBA, width, height));
 
             // Create the multisample depth/stencil buffer if requested
             if (settings.stencilBits)
@@ -565,6 +567,13 @@ bool RenderTextureImplFBO::activate(bool active)
     }
 
     return createFrameBuffer();
+}
+
+
+////////////////////////////////////////////////////////////
+bool RenderTextureImplFBO::isSrgb() const
+{
+    return m_sRgb;
 }
 
 
--- a/include/SFML/Graphics/RenderTarget.hpp
+++ b/include/SFML/Graphics/RenderTarget.hpp
@@ -277,6 +277,14 @@ public:
     virtual Vector2u getSize() const = 0;
 
     ////////////////////////////////////////////////////////////
+    /// \brief Tell if the render target will use sRGB encoding when drawing on it
+    ///
+    /// \return True if the render target use sRGB encoding, false otherwise
+    ///
+    ////////////////////////////////////////////////////////////
+    virtual bool isSrgb() const;
+
+    ////////////////////////////////////////////////////////////
     /// \brief Activate or deactivate the render target for rendering
     ///
     /// This function makes the render target's context current for
--- a/include/SFML/Graphics/RenderTexture.hpp
+++ b/include/SFML/Graphics/RenderTexture.hpp
@@ -217,6 +217,18 @@ public:
     ////////////////////////////////////////////////////////////
     virtual Vector2u getSize() const;
 
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Tell if the render-texture will use sRGB encoding when drawing on it
+    ///
+    /// You can request sRGB encoding for a render-texture
+    /// by having the sRgbCapable flag set for the context parameter of create() method
+    ///
+    /// \return True if the render-texture use sRGB encoding, false otherwise
+    ///
+    ////////////////////////////////////////////////////////////
+    virtual bool isSrgb() const;
+
     ////////////////////////////////////////////////////////////
     /// \brief Get a read-only reference to the target texture
     ///
--- a/include/SFML/Graphics/RenderWindow.hpp
+++ b/include/SFML/Graphics/RenderWindow.hpp
@@ -112,6 +112,17 @@ public:
     ////////////////////////////////////////////////////////////
     virtual Vector2u getSize() const;
 
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Tell if the window will use sRGB encoding when drawing on it
+    ///
+    /// You can request sRGB encoding for a window by having the sRgbCapable flag set in the ContextSettings
+    ///
+    /// \return True if the window use sRGB encoding, false otherwise
+    ///
+    ////////////////////////////////////////////////////////////
+    virtual bool isSrgb() const;
+
     ////////////////////////////////////////////////////////////
     /// \brief Activate or deactivate the window as the current target
     ///        for OpenGL rendering
--- a/src/SFML/Graphics/RenderTarget.cpp
+++ b/src/SFML/Graphics/RenderTarget.cpp
@@ -412,6 +412,14 @@ void RenderTarget::draw(const VertexBuffer& vertexBuffer, std::size_t firstVerte
 
 
 ////////////////////////////////////////////////////////////
+bool RenderTarget::isSrgb() const
+{
+    // By default sRGB encoding is not enabled for an arbitrary RenderTarget
+    return false;
+}
+
+
+////////////////////////////////////////////////////////////
 bool RenderTarget::setActive(bool active)
 {
     // Mark this RenderTarget as active or no longer active in the tracking map
@@ -680,6 +688,16 @@ void RenderTarget::applyShader(const Shader* shader)
 ////////////////////////////////////////////////////////////
 void RenderTarget::setupDraw(bool useVertexCache, const RenderStates& states)
 {
+    // Enable or disable sRGB encoding
+    // This is needed for drivers that do not check the format of the surface drawn to before applying sRGB conversion
+    if (!m_cache.enable)
+    {
+        if (isSrgb())
+            glCheck(glEnable(GL_FRAMEBUFFER_SRGB));
+        else
+            glCheck(glDisable(GL_FRAMEBUFFER_SRGB));
+    }
+
     // First set the persistent OpenGL states if it's the very first call
     if (!m_cache.glStatesSet)
         resetGLStates();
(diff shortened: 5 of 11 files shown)


src/SFML/Graphics/RenderTextureImplFBO.cpp
```diff
@@ -513,17 +515,28 @@ bool RenderTextureImplFBO::activate(bool active)
 
     Uint64 contextId = Context::getActiveContextId();
 
+    // FIXME: This does not work because sometimes Context::getActiveContext()
+    // returns a different context than Context::getActiveContextId()
+    const Context* activeContext = Context::getActiveContext();
+    if (m_sRgb && (!activeContext || !activeContext->getSettings().sRgbCapable))
+    {
+        contextId = 0;
+    }
```
Sakarah, line 524: @binary1248 Here is some code where I need to know the effective settings of the active context in order to tell if I need to switch to a sRGB compatible one before rendering. I think I am interested in the internal active context that "manage unshareable resources" not the "outside facing" context.
However since `Context::getActiveContext()` is not the same as `Context::getActiveContextId()`, my code does not work. On my machine, even when it would be required for correct output, it makes the texture never use its own sRGB capable rendering context except during startup.
@eXpl0it3r told me that you are the maintainer aware of these tricky details, and I need some advice on how to finish this properly.

Sakarah: In my last commit, I provided a fix with a new method `Context::getActiveContextSettings()`. There might be other ways, but none seemed more satisfying.

binary1248: I don't understand why the context has to actually have an sRGB pixel format surface (the surface format is what is reported by the context settings after creation). The context used or activated by the `RenderTextureImplFBO` is merely a container to house the FBO since it is an unshareable resource. The surface of this context will essentially never be written to (we write to the texture attachment). When `GL_FRAMEBUFFER_SRGB` is enabled the conversion should still take place independent of the pixel format of a context. Isn't it possible to just enable `GL_FRAMEBUFFER_SRGB` for whatever context gets activated instead of jumping through hoops to get an sRGB context surface that we will never write to?

Sakarah: As you said we do not care about the surface of the context. However we cannot mess up with the surface of the currently activated context. Maybe it is only a Linux problem but on my machine, if I enable `GL_FRAMEBUFFER_SRGB` on a context, it will consider that its surface must be sRGB encoded, which gives unexpected results.
It seemed to me that there is no easy ways of enabling `GL_FRAMEBUFFER_SRGB` just for the `RenderTextureImplFBO` since there is no deactivation hook, and the alternative would be to call `glEnable`/`glDisable` on every draw.

Just to be clear, this situation where we must create the useless surface is niche anyway. I expect very few users to actually request a sRGB texture with a non sRGB surface. Although it is not a reason for it to be buggy, if there can be a less optimized case, it is this one.

binary1248: From https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_framebuffer_sRGB.txt

> If FRAMEBUFFER_SRGB_EXT is enabled __and the boolean
    FRAMEBUFFER_SRGB_CAPABLE_EXT state for the drawable is true__, the R,
    G, and B destination color values (after conversion from fixed-point
    to floating-point) are considered to be encoded for the sRGB color
    space and hence need to be linearized prior to their use in blending.
...
Otherwise if FRAMEBUFFER_SRGB_EXT is disabled, __or the
    drawable is not sRGB capable__, or the value corresponds to the A
    component, then cs = cl for such components.

This wording was adapted into the core specification at some point as well.

In theory one could leave `GL_FRAMEBUFFER_SRGB` enabled for all contexts in the whole application and OpenGL should intelligently recognize when it makes sense to perform the conversion (when the drawable is sRGB capable). If you enable `GL_FRAMEBUFFER_SRGB` and try to draw to a non sRGB surface/drawable it will/should have the same effect as if `GL_FRAMEBUFFER_SRGB` was not enabled.

Therefore I don't see the necessity to mess with or worry about how a context was created. What matters is the FBO attachment (has to be sRGB) and whether `GL_FRAMEBUFFER_SRGB` is enabled at the time of drawing. Nothing else should matter, assuming the implementation conforms to the specification of course.

Since `GL_FRAMEBUFFER_SRGB` should be a simple flag somewhere deep in the driver, I also wouldn't be concerned about having to enable/disable it whenever drawing to the FBO. Sure... I myself keep saying every call to the driver is one call too many, but as it stands there are many other worse offenders in the codebase.

Sakarah: Problem is, in practice, even when not requesting sRGB capable surface, I get one. Is there an easy way to disable that ? I agree it would be better. The spec allows us not to worry about disabling `GL_FRAMEBUFFER_SRGB` for non-sRGB render texture though.

About enable/disable, the problem is that I have no easy way to tell that we are finishing an FBO rendering that explicitly activated `GL_FRAMEBUFFER_SRGB`. If that is what you want, maybe you have a clue on how to do that.

binary1248: https://github.com/SFML/SFML/blob/master/src/SFML/Graphics/Texture.cpp#L204

The texture image is only created with sRGB storage if it is explicitly requested. Not requesting an sRGB `RenderTexture` should lead to not requesting an sRGB texture which should lead to no sRGB conversion taking place.

If it is true that you are getting an sRGB texture even when one really shouldn't be created then the reason behind this should be investigated.

Remember... only the target drawable is relevant. The surface format of the context itself is (or should be) completely irrelevant in this context according to the specification since we are not targeting it with our draw.

As for enable/disable, it might seem "messy" but the way SFML deals with the issue is it just makes sure the required state is set right before drawing. In this case you just leave it enabled and the next place where a draw takes place you disable it (if the target is sRGB and you don't want a conversion to happen for some reason). If it turns out that this leads to severe performance degradation, the enabled/disabled state can be tracked in `RenderTarget` as is already done for some other things. Application-wide OpenGL state tracking is something to be looked into in a future re-implementation of the graphics module.

Sakarah: Sorry I was not clear, for RenderTextures I get the requested sRGB storage and this works perfectly well.
However for window surfaces, I always get an sRGB surface even when I do not ask for it.

binary1248: Can you post a list of visuals that are provided by your system via the `glxinfo` command? Maybe we might need to tweak the visual selection to favour a non-sRGB visual if the user doesn't request one.

Sakarah: Here is the output of `glxinfo` on my machine: https://gist.github.com/Sakarah/38707bbd68d82e68286ae0424d66aca6
It seems that I have both sRGB and non-sRGB visuals available.

binary1248: You can try adding a cost to the inverse case at https://github.com/SFML/SFML/blob/master/src/SFML/Window/GlContext.cpp#L656

When the user requests no sRGB and the format is sRGB it should also be disfavoured.

Sakarah: With a cost for the inverse case SFML selects visual `0x582` with sRGB and visual `0x50a` without. However, if I enable `GL_FRAMEBUFFER_SRGB`, even with the visual `0x50a` (which should be non-sRGB) it will sRGB encode. It might be a bug in my graphics driver since unless I misunderstood OpenGL spec, it should not.

binary1248: According to many articles, sRGB support was really buggy in the past (e.g. https://www.g-truc.net/post-0720.html). I can't tell if it has gotten any better nowadays, however I wouldn't be surprised if it didn't since OpenGL has unintentionally become somewhat of a legacy API now that Khronos is spending more of their time on Vulkan.

If this is really still the case, we would have to implement this "intelligence" that should be in the driver in our own code. That can get somewhat complicated... and dirty...

Sakarah: My driver clearly seem to follow the rule stated in the article you quote:
> IF FRAMEBUFFER_BINDING is 0 then if FRAMEBUFFER_SRGB is enabled, the R, G, and B values after blending are converted into the non-linear sRGB color space. Otherwise, follow the specification.

I guess it means we need to use the enable/disable strategy then.

binary1248: Just a thought... 

Would it make sense to add another virtual member function to `RenderTarget` so that it can know if the concrete implementation has an sRGB target or not? Something like `virtual bool RenderTarget::isSrgb() const`. This would have to be overridden by `RenderWindow` and `RenderTexture`. `RenderTexture` would have to decide based on 1) the concrete Impl and 2) the target of the Impl.

The `RenderTarget` would use this information to enable/disable `GL_FRAMEBUFFER_SRGB` alongside all the other state management it already does now.

Sakarah:
Independently on how we plan to solve the issue, I updated the code to test the PR in a more reproducible way.

src/SFML/Graphics/RenderTextureImplFBO.cpp
Sakarah: Might be interesting but can we do that without introducing a breaking change ?
I don't think there is a good default for `virtual bool RenderTarget::isSrgb() const` that do not force users to adapt their code if they have a custom child class for RenderTarget.

binary1248: Can't we just assume the default is to return `false`? The function doesn't have to be pure virtual. That way it wouldn't break existing code. In the off chance users have implemented their own `RenderTarget` and it is even sRGB, they will be happy to have a mechanism to inform the draw code of this information.

Sakarah: It seems to me that both defaulting to `true` and to `false` will make a custom `RenderTarget` behave differently. If `true` then, we insert a call to `glEnable(GL_FRAMEBUFFER_SRGB)` before a draw batch, if false, we insert a call to `glDisable(GL_FRAMEBUFFER_SRGB)`. Depending on user code, both may break the drawing.
Therefore I currently think that a pure virtual is still a better choice, to make the breaking change generate an error instead of silently failing to correctly draw.
If SFML has a strict breaking change policy, we might prefer to find another workaround, such as using the `setActive` overrides to change the `GL_FRAMEBUFFER_SRGB` flag.

binary1248: Generally, it has always been the case that you can manipulate OpenGL state within any sfml-graphics objects "at your own risk". SFML only guarantees that everything will work as long as all your OpenGL stuff is not done in the same context that is used by any sfml-graphics objects. As such, as weird as it sounds, behaviour related to enabling/disabling `GL_FRAMEBUFFER_SRGB` in user code was never guaranteed and thus cannot be broken by definition.

Sakarah: I just pushed a new version using this last idea

Sakarah:
@binary1248 Any remarks on the current state of this PR ?

binary1248 (approved):

binary1248:
Looks good to me now. Test works on my system.

eXpl0it3r (approved):

Issue #1759, opened 2021-03-19
MacOS: When initializing an openGL view with an existing window, call finishInit


jkeller51:
I had some long-standing bugs in my software dealing with SFML not receiving certain window notifications. I had worked around them outside of SFML but during a recent dig through the code I found the cause. When creating an `sf::RenderWindow` with a frame size (e.g. telling SFML to make its own window), it makes its `SFOpenGLView` receive certain window notifications, including `NSWindowDidChangeScreenNotification` and `NSWindowDidChangeScreenProfileNotification`.

But when instantiating an `sf::RenderWindow` inside of an existing window, the `SFOpenGLView` is never set up to receive these notifications, so certain events are "missed" by SFML. For DPI-aware applications, this can cause unexpected behavior.

I'm not sure if there was a reason for leaving this out, but adding it back in makes more sense to me and more cleanly solves the problems I was having.

## Your environment

* MacOS 11
* SFML https://github.com/SFML/SFML/commit/252d26afb0aa6375e08ef95b754cccf60f31b6df
* Xcode 12.4

## Steps to reproduce

Create an `NSWindow` manually, then create an `sf::RenderWindow` by passing the `NSWindow` handle.

## Expected behavior

When DPI scaling changes occur (e.g. passing between windows or changing display scaling), the sf::RenderWindow should receive an event. Currently this is supposed to be passed to the `Event::Resized` event.

## Actual behavior

When the `sf::RenderWindow` is instantiated inside of an existing NSWindow, this notification is not received.

eXpl0it3r:
Fixed with #1760

PR #1760, opened 2021-03-19
MacOS: When initializing an openGL view with an existing window, call finishInit

jkeller51:
See issue https://github.com/SFML/SFML/issues/1759 for a description of the problem.

For DPI-scaling reasons, I think it is the `sf::RenderWindow`'s "business" if the scaling factor changes, so my change signs it up to receive the same notifications on the window as it would if it had created the window.

* [x] Discussed in an issue

----

## Description

This PR is related to the issue #1759

## Tasks

* [x] Tested on macOS

## How to test this PR?

See below

### DIFF
--- a/src/SFML/Window/OSX/SFWindowController.mm
+++ b/src/SFML/Window/OSX/SFWindowController.mm
@@ -123,6 +123,8 @@
 
         // Set the view to the window as its content view.
         [m_window setContentView:m_oglView];
+        
+        [m_oglView finishInit];
     }
 
     return self;


jkeller51:
Code example:

```c++
#include <SFML/Window.hpp>
#include <SFML/Graphics.hpp>
#include <iostream>

#define USE_PARENT_WINDOW 0

int main() {
    
#if USE_PARENT_WINDOW
    sf::Window parentWindow(sf::VideoMode(600,400), "SFML Test");
    sf::RenderWindow window(parentWindow.getSystemHandle());
#else
    sf::RenderWindow window(sf::VideoMode(600,400), "SFML Test");
#endif
    
    while (window.isOpen()) {
        sf::Event event;
            switch(event.type) {
                case sf::Event::Closed:
                    window.close(); // will not get here when USE_PARENT_WINDOW==1
                    // *this is expected* because the parentWindow will (and should)
                    // receive this event and decide what to do with it, which in this
                    // case is nothing
                    break;
                case sf::Event::Resized:
                    std::cout << "Window resized to " << event.size.width << " x " <<
                        event.size.height << std::endl;
                    
                    break;
                default:
                    break;
            }
            window.display();
        }
        
        
    }
    
    
    return 0;
}

```

Try flipping USE_PARENT_WINDOW and dragging the window between screens and the problem will become clear.

Using my change, the problem is solved.

eXpl0it3r (approved):
Comparing it to the [original implementation](https://github.com/SFML/SFML/commit/3f61214571a2a3d950692ef6692d8314ef5eae00), this seems mostly an oversight to me.

Thanks for catching it and creating a PR! 🙂

eXpl0it3r:
While the fix makes sense to me, I still can't close the window, even after applying this fix.
Any ideas why?

eXpl0it3r:

eXpl0it3r:
Would be great, if you could retest this. The change makes sense, but in my personal tests I was still unable to close the window even with the fix applied.

jkeller51:
You still won't be able to -- the window closing takes a different path. Clicking the close button calls the NSWindow's close methods, and the `parentWindow` in my example is the NSWindow which receives those function calls and decides how to respond. In this example, I have not handled events for the `parentWindow` so it will not close.

The resizing changes I am targeting are notifications that do not pay attention to hierarchy -- anybody can sign up to receive these notifications. Other notifications that will be fixed by this PR are related to focusing and unfocusing the window. We can probably find other solutions to address other window-specific events that will not be affected by this change like window closing, etc. 

But it makes sense to me that `parentWindow` is responsible for handling window closing before anyone else. Technically when `USE_PARENT_WINDOW=1` in my example, the `sf::RenderTexture window` is just an NSView inside of an NSWindow, so in that context it makes sense that it wouldn't handle window closing. But it might be useful for the `sf::RenderWindow` to still receive a notification just before the window closes. I would propose a separate issue for that though.

eXpl0it3r:
Ohhh, I read the comment in the code wrong. 😄

PR #1761, opened 2021-03-20
Add support for arm64 on macOS

eXpl0it3r:
## Description

The existing dependencies in the extlibs directory
have been updated to use universal frameworks
for both x86_64 and arm64, allowing SFML to built
natively on the new M1 chipset

This PR is related to multiple discussions on the forum ([1](https://en.sfml-dev.org/forums/index.php?topic=27867.0), [2](https://en.sfml-dev.org/forums/index.php?topic=27782.0))

## Tasks

* [X] Tested on macOS (M1)
* [X] Tested on macOS (Intel)

## How to test this PR?

- Feature: Build SFML on an Apple device with an M1 chip
- Regression: Build SFML on an Apple device with an Intel chip


eXpl0it3r:
Pinging @jkeller51 😉 
Would you be able to test this PR, since I know you're among the few active SFML users with an M1 device?

eXpl0it3r:

jkeller51:
To test, I built this branch using CMake in Xcode, using both shared and static linking. I also enabled SFML_BUILD_EXAMPLES.

*SFML and all example projects build successfully.*

Most of the examples would not run out-of-the-box because the resources are not being set up properly for a MacOS bundle.

By forcing the resource directory to the absolute path of the directory on my system, I was able to run pong successfully. I think this proves that freetype and OpenAL are working on arm64. I am not sure how to test the other libraries, or if I need to. I assume there would be linker errors if they were not built properly.

ChrisThrasher:
What would it take to replace these Frameworks with dylibs? Is that feasible?

The problem is that until CMake 3.28 (which isn't even released yet) creating `IMPORTED` targets out of frameworks is prone to fail. See [this CMake MR](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8586) that fixes the problem once and for all. It's only in the `master` branch, but I've tested it and confirmed that it does fix this problem finding frameworks.

SFML 3 may not be able to raise it's CMake requirement to CMake 3.28 before release so this fix is great but not something we can count on for a little while.

In the meantime, one option is that we switch back to using dylibs for macOS libraries. This would let us write code like this

```cmake
find_package(Freetype REQUIRED)
target_link_libraries(sfml-graphics PRIVATE Freetype::Freetype)
```

This is useful because it means we can remove our own FindFreetype.cmake module and use CMake's built-in FindFreetype.cmake instead. It gets us one step closer to a dependency management solution where our CMaek code only has to `find_package` on a given dependency rather than our current solution based on `sfml_find_package` which does some custom, nonstandard stuff.

cc @mantognini

PR #1766, opened 2021-03-25
Fix '-Wimplicit-fallthrough' warnings by adding a fallthrough comment

SuperV1234:

----

## Description

When compiling with a modern version of GCC or Clang, the `-Wimplicit-fallthrough` warning will be produced for switch cases that do not have an explicit "fallthrough" comment (or attribute) in the code. This PR fixes all occurrences of such warning in SFML.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Just compile. The PR doesn't change any code, it only adds comments. Modern compilers will recognize the comment and suppress warning production. See <https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html>.

### DIFF
--- a/include/SFML/System/Utf.inl
+++ b/include/SFML/System/Utf.inl
@@ -62,11 +62,11 @@ In Utf<8>::decode(In begin, In end, Uint32& output, Uint32 replacement)
         output = 0;
         switch (trailingBytes)
         {
-            case 5: output += static_cast<Uint8>(*begin++); output <<= 6;
-            case 4: output += static_cast<Uint8>(*begin++); output <<= 6;
-            case 3: output += static_cast<Uint8>(*begin++); output <<= 6;
-            case 2: output += static_cast<Uint8>(*begin++); output <<= 6;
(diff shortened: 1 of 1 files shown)


eXpl0it3r:

SuperV1234:
> Can you squash the two commits?

GitHub allows you to do that when you merge the PR, isn't that what you're looking for?

eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:
Thanks for the fix! 🙂

JonnyPtn:
Unfortunately this doesn't fix the issue with AppleClang 12

Bromeon:
@JonnyPtn could you elaborate?

JonnyPtn:
Noticed on appleclang but beleive it applies to various versions of clang - the comment doesn’t suppress the warning

eXpl0it3r:
Yes, you can see it on your open PR as well. On clang it expects `[[clang::fallthrough]]`

Issue #1765, opened 2021-03-24
sf::Clock restart() returns negative values in continuous work


Mozzg:
Hello. I'm using SFML with a program for LED displays for railroad comunication in Russia. It displays nessesary information about trains on a screen and the hardware is converting video signal to LEDs. This program must run continuously and preferrably without restarts. I found an issue with SFML and how it calculates time (see below).

* OS: Windows 10 x64
* SFML Version: 2.5.1
* Compiler: Microsoft Visual Studio Enterprise 2017

I'll go straight to the point. While testing my program, I encountered an issue, where after continuous work for about 10-11 days, the program started returning negative time and it disrupts normal work. I found the issue is in Win32 ClockImpl implementation and similar to issue #1167 . Also I found out that it depends on hardware and return value of QueryPerformanceFrequency function.
This is current version of ClockImpl.cpp:

    ...
    QueryPerformanceCounter(&time);
    ...
    // Return the current time as microseconds
    return sf::microseconds(1000000 * time.QuadPart / frequency.QuadPart);
The issue is in return statement. time.QuadPart type is signed __int64, which means if it goes over 9 223 372 036 854 775 807 it will go to negative numbers. In this calculation time.QuadPart needs to only be over 9 223 372 036 855, because this number is multiplied with 1000000 first. On my test computer, QueryPerformanceFrequency function returns 10000000, which means there is 10000000 tick per second. In this sircumstances time.QuadPart reaches it's overflow value in 9223372036855 / 10000000 seconds or 922337,2 seconds or 256,2 hours or just only 10,6 days.
On my home computer, QueryPerformanceFrequency returns 3609372. If we calculate time it takes to overflow with this frequency , it is 29,57 days, which is still not good. This program should run uninterrupted for several months or years.

This issue is easily fixable. we just need to exclude a calculation with big intermediate result, for example like this:

    // Return the current time as microseconds
    double inverse = 1000000 / frequency.QuadPart;
    return sf::microseconds(time.QuadPart * inverse);

In my program I used `sf::Time elapsed = clock.restart();`. Then I add this to my different timers to see if timer is fired, like `sf::Int64 elapsedMicro = elapsed.asMicroseconds(); foregroundElapsedTimer += elapsedMicro;` Which calculates elapsed time like:

    Time Clock::restart()
    {
        Time now = priv::ClockImpl::getCurrentTime();
        Time elapsed = now - m_startTime;
        m_startTime = now;
        return elapsed;
    }

[...]

PR #1771, opened 2021-03-29
Fix overflow when using sf::Clock for long time

Mozzg:
* [x] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before? It was an issue.

----

## Description

This PR is related to the issue #1765

When using sf::Clock class, method restart() may return negative value in continious work because of overflow when calculating current time in microseconds. This pull request is trying to fix that.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

See description in issue #1765

### DIFF
--- a/src/SFML/System/Win32/ClockImpl.cpp
+++ b/src/SFML/System/Win32/ClockImpl.cpp
@@ -54,9 +54,9 @@ namespace priv
 ////////////////////////////////////////////////////////////
 Time ClockImpl::getCurrentTime()
 {
-    // Get the frequency of the performance counter
-    // (it is constant across the program lifetime)
-    static LARGE_INTEGER frequency = getFrequency();
+    // Calculate inverse of frequency multiplied by 1000000 to prevent overflow in final calculation
+    // Frequency is constant across the program lifetime
+    static double inverse = 1000000.0 / getFrequency().QuadPart;
 
     // Detect if we are on Windows XP or older
     static bool oldWindows = isWindowsXpOrOlder();
@@ -80,7 +80,7 @@ Time ClockImpl::getCurrentTime()
     }
 
     // Return the current time as microseconds
-    return sf::microseconds(1000000 * time.QuadPart / frequency.QuadPart);
+    return sf::microseconds(static_cast<sf::Int64>(time.QuadPart * inverse));
 }
 
 } // namespace priv


Sakarah:
I am worried about the fact that this PR introduces floating point computation to resolve the overflow problem. This might be creating dangerous side effects, since precision now depends on the order of magnitude of the number and the division will be rounded to nearest even instead of floored.

I strongly suggest trying to fix the problem with integer arithmetic only instead.

Mozzg:
@Sakarah This PR does not introduce floating point computation, because there was one to begin with. In return statement there was division, so it is floating point computation and it ended up casting to integer type anyway. Also I didn't see that SFML needed something more precise then microseconds, so any potential precision problems will not affect anything. 
In the end the minimum time period is 1 microsecond and any precision problems will affect something much much less than that.

Sakarah:
Well, an integer division is clearly not a floating point operation. It is the arithmetically correct division that you can add back with the remainder to perfectly restore the original number.
The fact that you had to introduce a `.0` and a static_cast actually stems from the fact that you actually did not simply reorder the operands.

After thinking about it, I guess your fix is OK as it does not loose precision as long as you do not exceed 52 bits in the result. This is I admit very unlikely for a result in microseconds. (2^52 ms is over 100 years)

Mozzg:
The introduction of `.0` and `static_cast` is simply to avoid warnings. I thought it was a good programming practise to do those things. If you want I can remove them. I mainly program on Delphi, so I may not know some specifics.

As for floating point and precision I have to disagree, but let's move on. I would like to see your suggestions on how we can avoid those operations. I can't think of one.

Sakarah:
Yes obviously you want to explicit the conversions as you did, the warning says that the operation exist but is not written so the code might be misleading if you don't write these things.

Where do you disagree ? I just said that precision is fine until the result time is greater than 52-bits. And this is arguably enough. In case I wasn't clear I changed my mind about the necessity of pure integer arithmetic in this case. I was actually more worried than against. A comment justifying that a 52-bit precision provided by double is enough might be a good idea to convince maintainers if they are as picky as me.

Just for information, if you wanted to do this with pure integer arithmetic, you can use a code such as:
```c++
Int64 full_seconds = time.QuadPart / frequency.QuadPart;
Int64 remainder = time.QuadPart % frequency.QuadPart;
return full_seconds * 1000000 + remainder * 1000000 / frequency.QuadPart;
```

However I benched the performances out of curiosity, and your code is actually faster on x86_64, gcc -O3 than both the current SFML and my solution. Floating point multiplication is incredibly cheap compared to 64 bit integer division.

Mozzg:
@Sakarah Thank you for replies. I hope this PR will be added to master as soon as posible, Clock class is very convenient to use. 

I didn't think of your solution and maybe I think I understand your concern. In my mind every division operation is a floating point operation, because even when two operands are integers, result is a floating point type.

src/SFML/System/Win32/ClockImpl.cpp
```diff
@@ -57,6 +57,9 @@ Time ClockImpl::getCurrentTime()
     // Get the frequency of the performance counter
     // (it is constant across the program lifetime)
     static LARGE_INTEGER frequency = getFrequency();
+	
+	//calculate inverse of frequency multiplied by 1000000 to prevent overflow in final calculation
+	static double inverse = 1000000.0 / frequency.QuadPart;
```
Bromeon, line 62: Indentation seems to be off.

Bromeon (changes requested):
@Mozzg Thanks a lot for your PR. Did this code solve the problem you had in #1765?
I'll try to test it myself soon.

> In my mind every division operation is a floating point operation, because even when two operands are integers, result is a floating point type.

No, integer division is a different machine instruction than floating-point division, see e.g. [DIV](https://docs.oracle.com/cd/E18752_01/html/817-5477/ennbz.html#eoizh) and [FDIV](https://docs.oracle.com/cd/E18752_01/html/817-5477/eoizy.html#epmnq). The result of integer division is again an integer, not a floating point type.

src/SFML/System/Win32/ClockImpl.cpp
Mozzg: All done.

Mozzg:
@Bromeon @eXpl0it3r 
I tried to merge commits into one but failed and ended up creating more commits. I did not worked with GitHub on this level, so I'll leave it to you. I also adjusted the changes.

I did not used this exact code in my project, rebuilding SFML would be too much work. I just replaced sf::Clock class and time manipulation with this adjusted code. I tested the code by adding a constant, I wrote this test method in my issue #1765. The run/test without using a constant (but only with custom code without Clock class) is going right now and results should be known in 3 days.

My code looks like this:

	LARGE_INTEGER clockFreq;	
	LARGE_INTEGER clockMainNow;
	LARGE_INTEGER clockMainPrev;
	LARGE_INTEGER clockMainElapsed;
	QueryPerformanceFrequency(&clockFreq);
	QueryPerformanceCounter(&clockMainNow);		
	clockMainPrev.QuadPart = clockMainNow.QuadPart;
	clockMainElapsed.QuadPart = 0;
	double clockReverseFreq = 1000000.0 / clockFreq.QuadPart;
	sf::Uint64 elapsedMicro = 0;
	...
	while (running)
	{
		...
		QueryPerformanceCounter(&clockMainNow);
		if (clockMainNow.QuadPart < clockMainPrev.QuadPart)  //just in case
		{
			clockMainElapsed.QuadPart = 0;				
		}
		else
		{
			clockMainElapsed.QuadPart = clockMainNow.QuadPart - clockMainPrev.QuadPart;
		}
		clockMainPrev.QuadPart = clockMainNow.QuadPart;			
		elapsedMicro = static_cast<sf::Uint64>(clockMainElapsed.QuadPart * clockReverseFreq);
		...
	}

eXpl0it3r:

eXpl0it3r:
I squashed the commits and rebased onto master.

Bromeon (approved):
Looks good to me 👍🏼 

I ran the `tennis` example for a few minutes on Windows (with a displayed clock), and seems to work well from what I can tell. I didn't let it run long enough to trigger the overflow, though 😉

PR #1779, opened 2021-04-11
windows: add resource.rc files to dll builds

binary1248:
Windows uses a mechanism known as 'resource files' which provides, among
other things, metadata to a given executable/dll/driver/etc, and add a
layer of polish to a project which it would otherwise lack.

Supersedes and closes #1419.


eXpl0it3r (approved):
Looks good 🙂 

image

PR #1782, opened 2021-04-12
Fix ordering of 'sf::Font' members in 'Font.cpp'

SuperV1234:
## Description

Change the order in which members are initialized and swapped in `Font.cpp` to exactly match the order in which they were declared, fixing [`-Wreorder`](https://stackoverflow.com/questions/1828037/whats-the-point-of-g-wreorder) warnings.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Compile/run entirety of SFML and test suite.

### DIFF
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -105,10 +105,10 @@ m_face       (copy.m_face),
 m_streamRec  (copy.m_streamRec),
 m_stroker    (copy.m_stroker),
 m_refCount   (copy.m_refCount),
+m_isSmooth   (copy.m_isSmooth),
 m_info       (copy.m_info),
 m_pages      (copy.m_pages),
-m_pixelBuffer(copy.m_pixelBuffer),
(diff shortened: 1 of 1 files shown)


eXpl0it3r (approved):

eXpl0it3r:

eliasdaler:
@eXpl0it3r can we add `-Wreorder` to CI?

eXpl0it3r:
We could add some additional flags, however we also need to enforce treat warning as errors or similar, otherwise it's not really all that useful, since people usually don't check warning outputs from the CI.

This then however leads to issues, where some warnings might be emitted from non-SFML code (e.g. stb* files) and some (deprecation) warnings that are okay.

As Jonny mentioned on Discord, I think we could take Jason Turners best practices set as a starting point and step by step fix things, until we can effectively enforce treat-warning-as-error.

PR #1783, opened 2021-04-12
Remove standard '<algorithm>' depedency in common headers

SuperV1234:
## Description

The standard `<algorithm>` header is expensive to compile in comparison to other standard headers. I have benchmarked this myself using `clang++ -ftime-trace` on a [large SFML-powered codebase](https://github.com/SuperV1234/SSVOpenHexagon/) and it has also been [experienced and verified by others in the past](https://blog.magnum.graphics/backstage/reducing-cpp-compilation-time-in-magnum-code-optimizations/).

When benchmarking, I've noticed that many of my object files were taking an unexpected extra 800-900ms to compile due to `<algorithm>` being brought in by `Rect.hpp`. Upon investigation, I noticed that `Rect.hpp` only needs `std::min` and `std::max` -- including `<algorithm>` is overkill for that functionality.

Furthermore, `Rect.hpp`'s usage of `std::min` and `std::max` is only on primitive types, which can generate suboptimal codegen when taken by reference in functions. 

I removed the `<algorithm>` dependency in `Rect.hpp` by implementing two `private` static functions (`getMin` and `getMax`). I have also removed another `<algorithm>` depedency in `JoystickImpl.hpp` by using some simple loops. 

Some other SFML files relied on transitive `<algorithm>` inclusion of `std::size_t` (which is pretty bad), so I fixed those by explicitly including `<cstddef>`.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Compile/run entirety of SFML + test suite.

### DIFF
--- a/include/SFML/Graphics/Glsl.hpp
+++ b/include/SFML/Graphics/Glsl.hpp
@@ -32,7 +32,7 @@
 #include <SFML/Graphics/Color.hpp>
 #include <SFML/System/Vector2.hpp>
 #include <SFML/System/Vector3.hpp>
-
+#include <cstddef>
 
 namespace sf
 {
--- a/include/SFML/Graphics/Rect.hpp
+++ b/include/SFML/Graphics/Rect.hpp
@@ -29,7 +29,6 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/System/Vector2.hpp>
-#include <algorithm>
 
 
 namespace sf
@@ -178,6 +177,33 @@ public:
     T top;    //!< Top coordinate of the rectangle
     T width;  //!< Width of the rectangle
     T height; //!< Height of the rectangle
+
+private:
+    ////////////////////////////////////////////////////////////
+    /// \brief Return the minimum between two values
+    ///
+    /// This function is implemented here to avoid a dependency
+    /// on the <algorithm> C++ Standard header, unnecessarily
+    /// expensive to compile for the task of computing min/max.
+    ///
+    /// \param a First value
+    /// \param b Second value
+    ///
+    ////////////////////////////////////////////////////////////
+    static T getMin(T a, T b);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Return the maximum between two values
+    ///
+    /// This function is implemented here to avoid a dependency
+    /// on the <algorithm> C++ Standard header, unnecessarily
+    /// expensive to compile for the task of computing min/max.
+    ///
+    /// \param a First value
+    /// \param b Second value
+    ///
+    ////////////////////////////////////////////////////////////
+    static T getMax(T a, T b);
 };
 
 ////////////////////////////////////////////////////////////
--- a/include/SFML/Graphics/Rect.inl
+++ b/include/SFML/Graphics/Rect.inl
@@ -78,10 +78,10 @@ bool Rect<T>::contains(T x, T y) const
     // Rectangles with negative dimensions are allowed, so we must handle them correctly
 
     // Compute the real min and max of the rectangle on both axes
-    T minX = std::min(left, static_cast<T>(left + width));
-    T maxX = std::max(left, static_cast<T>(left + width));
-    T minY = std::min(top, static_cast<T>(top + height));
-    T maxY = std::max(top, static_cast<T>(top + height));
+    T minX = getMin(left, static_cast<T>(left + width));
+    T maxX = getMax(left, static_cast<T>(left + width));
+    T minY = getMin(top, static_cast<T>(top + height));
+    T maxY = getMax(top, static_cast<T>(top + height));
 
     return (x >= minX) && (x < maxX) && (y >= minY) && (y < maxY);
 }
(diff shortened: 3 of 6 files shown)


JonnyPtn:
I'm personally not a fan of this change, I don't think the benefit is worth it for the majority of users and prefer to keep the code cleaner/simpler

eliasdaler:
Not a fan of `std::min`/`std::max` change. However, replacing `std::fill` by a simple loop looks easier to understand to me.

SuperV1234:
@JonnyPtn @eliasdaler: I would keep the code as is if there weren't actual tangible benefits to this change. Unfortunately, until C++20 modules become widespread, header files come with a large baggage -- they're literally a copy-paste mechanism.

The `<algorithm>` header is one of the most expensive ones to include in terms of compilation time (see [this article](https://blog.magnum.graphics/backstage/reducing-cpp-compilation-time-in-magnum-code-optimizations/)) and I have benchmarked it myself.

I have about 20 objects files in my project that end up depending on `Rect.hpp` due to transitive dependencies, and after this PR, each one takes 700-800ms less to compile which is a huge improvement for me. If this is not merged, it is worthwhile for me to maintain my own fork of SFML.

I think this change will benefit many users with faster compilation time, and possibly better codegen as `std::min`/`std::max` always take their argument by reference, which is a pessimization for primitive types.

eliasdaler:
Okay, faster compilation argument is pretty strong.
However, I think it's better to make replacement min/max functions free functions (somewhere in `sf::priv` namespace) instead of member functions.

eXpl0it3r:
Not to invalidate this PR, but wouldn't pre-compiled headers solve this issue for your code base?
It saved quite a lot of time [in the ÖbEngine](https://github.com/Sygmei/ObEngine/pull/317), but I'm not deep enough in the whole C++ build system world to fully judge.

I vote against the JoystickImpl changes:
- Using algorithms over loops is the preferred way
- It only adds to the SFML compile-time and not SFML user code compile-time

I agree that Rect.hpp as Vector2.hpp and similar should be lightweight enough, since they should be used generously.
Changing the private interface of a class to avoid a standard header doesn't seem like the best solution.
Would templated free functions in an anonymous namespace also work?

JonnyPtn:
@SuperV1234 Can you share the results of your benchmark? I'm fully aware of how includes work and the costs involved, I'm just not convinced the cost here is significant enough to warrant the changes. 

In my benchmarks the difference is so negligible as to almost be unnoticeable - can you provide more details on your benchmarks?

SuperV1234:
Discussed on Discord, not going forward with this idea as the advantages probably do not outweigh having to reimplement and maintain our own `min`/`max` and people should use [PCH](https://github.com/Sygmei/ObEngine/pull/317/files) instead.

Bromeon:
@SuperV1234 Could you summarize the important points from the Discord discussion here?

I'm generally also a fan of reducing compile times, especially if gets significant gains (almost 1s for a user) for little changes.
PCH are nice, but still a workaround and often intrusive (one needs to change every translation unit).

I don't think SFML (or many C++ libraries, for that matter) will migrate to C++20 modules anytime soon.

SuperV1234:
@Bromeon 

> Could you summarize the important points from the Discord discussion here?

- At least three users felt that PCH is a better solution, some quotes:
   - > it's not our job to optimize the compilation process of the users
   - > pre-compiled headers are a thing and most big projects I've worked on uses them
   - > (referring to the benefit of this PR)  A few seconds for every build of every project using SFML. That would be worth it, if there's no alternative, like PCH
   - > It's not a few seconds for every project
   - > it's bad code design to do it for a few and not all. It's inconsistent and doesn't give any big real advantage. It becomes a maintenance issue
   - > Setting up PCH is as non-trivial as setting up a third party library like SFML. It's far from being as complicated as people make it to be
   - > CMake even supports it, it's a peice of cake
   - > Having never done any PCH before in my life, I was able to add it to a project I'm not really familiar with in a very short time period 

- At least one user felt that this could set a precedent for micro-optimization of compile-times
    - > It would set a precedent though - where do we draw the line on micro-optimisations for users case studies?
    - > we'd need our own utility/stdlib replacement header
    - > And this is my main argument for not changing it - current code is cleanest, and there’s no strong empirical argument to change

- On the context of possible run-time optimization when `min`/`max` do not get inlined, people felt debug performance is not important and the possible optimization is not guaranteed/worth it
    - > Debug build performance DOES NOT matter

Bromeon:
Thanks! While I don't agree with the "it's not our job" or "micro-optimization" arguments (it's almost 1s after all), I also don't see a nice solution other than rewriting `min`/`max` from scratch. I wouldn't have anything for such isolated instances, but this looks a bit like the start of "rewrite STL because we know better" trap that many game engines fell into.

But yes, dependencies in headers are a big problem in C++, especially with templates. There are countless workarounds at the build system level, but no really nice solutions in the language (not sure about modules). Just for completeness, explicit instantiation of the templates (for `float`, `int`, `unsigned int`) would also be a possibility, but I don't think it makes sense for SFML, and it would break user code that uses `Rect`, `Vector2` and `Vector3` with other types.

Issue #1786, opened 2021-04-13
OpenGL extension SGIS_texture_edge_clamp unavailable


danieljpetersen:
## Subject of the issue

From commit 4fcef893c9d67630ad6d9e9bbf2ee48a62f95eef onwards the following error appears in the terminal, presumably only on nvidia cards

OpenGL extension SGIS_texture_edge_clamp unavailable
Artifacts may occur along texture edges
Ensure that hardware acceleration is enabled if available

## Your environment

* nvidia 3080 GPU, Ubuntu 20.04 (driver 460.39) and Windows 10 (driver 461.33)
* GCC 9.3.0, Visual Studio 2019 

## Steps to reproduce

Any program making use of a texture

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode(200, 200), "SFML works!");
    sf::CircleShape shape(100.f);
    shape.setFillColor(sf::Color::Green);

    sf::RenderTexture T;
    T.create(200, 200);

   
    while (window.isOpen())
    {
        sf::Event event;
        while (window.pollEvent(event))
        {
            if (event.type == sf::Event::Closed)
                window.close();
        }

        T.clear();
        T.draw(shape);
        T.display();

        sf::Sprite spr;
        spr.setTexture(T.getTexture());

        window.clear();
        window.draw(spr);
        window.display();
    }

    return 0;
}

```

----

I managed to get the error to go away by changing line 11988 in https://github.com/SFML/SFML/blob/master/extlibs/headers/glad/include/glad/gl.h from

```cpp
    SF_GLAD_GL_SGIS_texture_edge_clamp = sf_glad_gl_has_extension(version, exts, num_exts_i, exts_i, "GL_SGIS_texture_edge_clamp");
```
to it's old value

```cpp
    SF_GLAD_GL_SGIS_texture_edge_clamp = sf_glad_gl_has_extension(version, exts, num_exts_i, exts_i, "GL_SGIS_texture_edge_clamp") | sf_glad_gl_has_extension(version, exts, num_exts_i, exts_i, "GL_EXT_texture_edge_clamp");
```

It looks like this an issue in the past which I'm assuming was manually fixed, https://github.com/SFML/SFML/issues/880. I'm assuming that when this file was regenerated the fix didn't get  carried over.

eXpl0it3r:
It's a regression from #1769.

[...]

PR #1787, opened 2021-04-13
Add support for SFML unity builds

SuperV1234:
## Description

This PR renames objects with internal linkage defined in `.cpp` files in order to avoid name collisions when performing a *unity build* of SFML. You can read the benefits of *unity builds* [in this article](https://onqtam.com/programming/2019-12-20-pch-unity-cmake-3-16/). Here's an excerpt: 

> ## Unity builds
>
> The idea is to cram the .cpp files of a CMake target into a few .cpp files which include the original .cpp files
>
> - up to 7-8 times faster builds (usually x3 or x4)
>
> - example: a project of 200 .cpp files might be divided into 16 unity .cpp (or .cxx - whatever) files each including about 13 of the original .cpp files => 16 .cpp files to build in parallel and 16 .obj files to link
>
> - the reasons for the speedup are:
>     - common headers from the different .cpp files end up being included and parsed fewer times (this is beneficial even when using PCHs!)
>     - common template instantiations with the same types in separate .cpp files (vector<int>) end up being done in fewer places
>     - the linker has to stitch much fewer .obj files in the end - there are a lot less weak symbols to deduplicate (inline/template
>     - functions from headers end up in every .obj => linkers leave just 1)
>          - surprisingly incremental builds (changing a single .cpp) will probably also be faster instead of slower (even though you compile more .cpp files together) - precisely because of the reduced number of weak duplicated symbols!
>     - less compiler invocations and less .obj files are written to disk

This PR renames `static` objects defined in different `.cpp` files to avoid name collisions. It also fixes up some includes of `wgl.h` and `gl.h` to avoid redefinition errors. The renaming is necessary -- from the linked article:

> Initial problems when trying to compile a project as unity
>
> - there will be static globals (or in anonymous namespaces) in different .cpp files with identical names which would clash
>    - either rename them or put such globals into an additional namespace - perhaps with the name of the file: for GraphVisitor.cpp I would recommend GRAPH_VISITOR_CPP (putting static symbols inside of a named namespace in a .cpp keeps their linkage to internal , and the same is true with nesting anonymous namespaces into named ones)

And here are some benchmarks of a clean build on my machine, using MSYS2 + MinGW-w64:

```bash
# with `set(CMAKE_UNITY_BUILD ON)`
rm CMakeCache.txt; cmake .. -G"MinGW Makefiles"; make clean; time make -j8

real    0m11.285s
user    0m0.000s
sys     0m0.015s
```

```bash
# with `set(CMAKE_UNITY_BUILD OFF)`
rm CMakeCache.txt; cmake .. -G"MinGW Makefiles"; make clean; time make -j8

real    0m15.871s
user    0m0.000s
sys     0m0.000s
```

On average, I get a ~4s speedup in a full rebuild of SFML. 

Note that this PR does *not* enable unity builds. It merely makes them possible for users of SFML who are interested in them and build SFML from source.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

1. Compile the entirety of SFML as is
2. Clean the build
3. Compile the entirety of SFML passing `-DCMAKE_UNITY_BUILD=ON` to CMake

### DIFF
--- a/src/SFML/Audio/AlResource.cpp
+++ b/src/SFML/Audio/AlResource.cpp
@@ -34,13 +34,13 @@
 namespace
 {
     // OpenAL resources counter and its mutex
-    unsigned int count = 0;
-    sf::Mutex mutex;
+    unsigned int sfmlAlResourceCount = 0;
+    sf::Mutex sfmlAlResourceMutex;
 
     // The audio device is instantiated on demand rather than at global startup,
     // which solves a lot of weird crashes and errors.
     // It is destroyed when it is no longer needed.
-    sf::priv::AudioDevice* globalDevice;
+    sf::priv::AudioDevice* sfmlAlGlobalDevice;
 }
 
 
@@ -50,14 +50,14 @@ namespace sf
 AlResource::AlResource()
 {
     // Protect from concurrent access
-    Lock lock(mutex);
+    Lock lock(sfmlAlResourceMutex);
 
     // If this is the very first resource, trigger the global device initialization
-    if (count == 0)
-        globalDevice = new priv::AudioDevice;
+    if (sfmlAlResourceCount == 0)
+        sfmlAlGlobalDevice = new priv::AudioDevice;
 
     // Increment the resources counter
-    count++;
+    sfmlAlResourceCount++;
 }
 
 
@@ -65,14 +65,14 @@ AlResource::AlResource()
 AlResource::~AlResource()
 {
     // Protect from concurrent access
-    Lock lock(mutex);
+    Lock lock(sfmlAlResourceMutex);
 
     // Decrement the resources counter
-    count--;
+    sfmlAlResourceCount--;
 
     // If there's no more resource alive, we can destroy the device
-    if (count == 0)
-        delete globalDevice;
+    if (sfmlAlResourceCount == 0)
+        delete sfmlAlGlobalDevice;
 }
 
 } // namespace sf
--- a/src/SFML/Audio/SoundRecorder.cpp
+++ b/src/SFML/Audio/SoundRecorder.cpp
@@ -40,7 +40,7 @@
 
 namespace
 {
-    ALCdevice* captureDevice = NULL;
+    ALCdevice* sfmlALCDeviceCaptureDevice = NULL;
 }
 
 namespace sf
@@ -81,7 +81,7 @@ bool SoundRecorder::start(unsigned int sampleRate)
     }
 
     // Check that another capture is not already running
-    if (captureDevice)
+    if (sfmlALCDeviceCaptureDevice)
     {
         err() << "Trying to start audio capture, but another capture is already running" << std::endl;
         return false;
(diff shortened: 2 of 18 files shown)


binary1248:
In my opinion it is always nice for a library to support as many development styles as is technically feasible without having to compromise other factors such as cleanliness or runtime performance. I already contemplated such a change for quite a while since (as you discovered) the required changes are (or in theory should be) purely syntactical.

When browsing through the changeset I couldn't help but have the feeling that the name changes felt a bit too "C-like" to me. In C++ we have the wonderful `namespace` feature that was designed specifically to prevent name collisions. As far as I understand, nesting a named `namespace` within an unnamed `namespace` would inherit the internal linkage as well as provide the disambiguation necessary to support unity builds. In my opinion it would also make the code easier to read since only a TU-specific prefix would have to be prepended to the respective occurrences of the names within the TU.

Regarding the issue with glad: This is a tricky one. Basically the integration method that glad chose to use (which I actually like) is that of a "single-file header-only library" although being an OpenGL loader the single-file part is basically impossible to do in a clean way. Within an entire application, there should only be a single compiled instance of glad (or to be more exact a single compiled instance of each of its files). As such, to keep supporting non-unity builds we can't define `SF_GLAD_GL_IMPLEMENTATION` in multiple places, that will just break everything when building as a static library as CI shows. I still don't fully understand the necessity to define `SF_GLAD_GL_IMPLEMENTATION` multiple times in this changeset. The concrete definition of its symbols is really only required in the graphics module (access to e.g. only the `#define`s doesn't require compilation), and whether building SFML in the traditional way or in a unity build, this shouldn't change.

I support this PR, but the current changes have to be worked on a bit more.

SuperV1234:
@binary1248: thanks for the feedback! I will wait for more maintainters' opinions, and if everyone is on-board I will clean it up and will try to figure out how to deal with `glad`.

Would modifying/splitting the `glad` header be reasonable at all? Or do we want to keep it untouched?

binary1248:
We already have a reported regression in #1786 because I updated glad and forgot to add in the tweak that I did by hand in the previous version. As such, going forward, I really prefer not having to modify glad in any way from its auto-generated form.

SuperV1234:
@binary1248: I am starting this from scratch using namespaces. This is the error I get when trying to build `sfml-graphics` with a unity build:

```
In file included from C:\OHWorkspace\SFML\build\src\SFML\Graphics\CMakeFiles\sfml-graphics.dir\Unity\unity_0_cxx.cxx:13:
C:/OHWorkspace/SFML/src/SFML/Graphics/GLExtensions.cpp: In function 'void sf::priv::ensureExtensionsInit()':
C:/OHWorkspace/SFML/src/SFML/Graphics/GLExtensions.cpp:57:9: error: 'gladLoadGL' was not declared in this scope
   57 |         gladLoadGL(reinterpret_cast<GLADloadfunc>(sf::Context::getFunction));
      |         ^~~~~~~~~~
```

How am I supposed to make that declaration visible without making everything explode?

It's kinda weird. `GLExtensions.cpp` has:

```cpp
#define SF_GLAD_GL_IMPLEMENTATION
#include <SFML/Graphics/GLExtensions.hpp>
```

And it relies on the transitive include of `GLExtensions.hpp`: 

```cpp

#ifndef SFML_GLEXTENSIONS_HPP
#define SFML_GLEXTENSIONS_HPP

// Headers
#include <SFML/Config.hpp>
#include <glad/gl.h>

// ...
```

What is the correct way of using `<glad/gl.h>` and the `SF_GLAD_GL_IMPLEMENTATION` macro in this case?

SuperV1234:
Better version in #1788

PR #1788, opened 2021-04-14
Enable support for unity builds

SuperV1234:
## Description

This PR renames objects with internal linkage defined in `.cpp` files in order to avoid name collisions when performing a *unity build* of SFML. You can read the benefits of *unity builds* [in this article](https://onqtam.com/programming/2019-12-20-pch-unity-cmake-3-16/). Here's an excerpt: 

> ## Unity builds
>
> The idea is to cram the .cpp files of a CMake target into a few .cpp files which include the original .cpp files
>
> - up to 7-8 times faster builds (usually x3 or x4)
>
> - example: a project of 200 .cpp files might be divided into 16 unity .cpp (or .cxx - whatever) files each including about 13 of the original .cpp files => 16 .cpp files to build in parallel and 16 .obj files to link
>
> - the reasons for the speedup are:
>     - common headers from the different .cpp files end up being included and parsed fewer times (this is beneficial even when using PCHs!)
>     - common template instantiations with the same types in separate .cpp files (vector<int>) end up being done in fewer places
>     - the linker has to stitch much fewer .obj files in the end - there are a lot less weak symbols to deduplicate (inline/template
>     - functions from headers end up in every .obj => linkers leave just 1)
>          - surprisingly incremental builds (changing a single .cpp) will probably also be faster instead of slower (even though you compile more .cpp files together) - precisely because of the reduced number of weak duplicated symbols!
>     - less compiler invocations and less .obj files are written to disk

This PR renames `static` objects defined in different `.cpp` files to avoid name collisions. It also fixes up some includes of `wgl.h` and `gl.h` to avoid redefinition errors. The renaming is necessary -- from the linked article:

> Initial problems when trying to compile a project as unity
>
> - there will be static globals (or in anonymous namespaces) in different .cpp files with identical names which would clash
>    - either rename them or put such globals into an additional namespace - perhaps with the name of the file: for GraphVisitor.cpp I would recommend GRAPH_VISITOR_CPP (putting static symbols inside of a named namespace in a .cpp keeps their linkage to internal , and the same is true with nesting anonymous namespaces into named ones)

And here are some benchmarks of a clean build on my machine, using MSYS2 + MinGW-w64:

```bash
# with `set(CMAKE_UNITY_BUILD ON)`
rm CMakeCache.txt; cmake .. -G"MinGW Makefiles"; make clean; time make -j8

real    0m11.285s
user    0m0.000s
sys     0m0.015s
```

```bash
# with `set(CMAKE_UNITY_BUILD OFF)`
rm CMakeCache.txt; cmake .. -G"MinGW Makefiles"; make clean; time make -j8

real    0m15.871s
user    0m0.000s
sys     0m0.000s
```

On average, I get a ~4s speedup in a full rebuild of SFML. 

Note that this PR does *not* enable unity builds. It merely makes them possible for users of SFML who are interested in them and build SFML from source.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

1. Compile the entirety of SFML as is
2. Clean the build
3. Compile the entirety of SFML passing `-DCMAKE_UNITY_BUILD=ON` to CMake

### DIFF
--- a/src/SFML/Graphics/RenderTarget.cpp
+++ b/src/SFML/Graphics/RenderTarget.cpp
@@ -52,95 +52,99 @@
 
 namespace
 {
-    // Mutex to protect ID generation and our context-RenderTarget-map
-    sf::Mutex mutex;
-
-    // Unique identifier, used for identifying RenderTargets when
-    // tracking the currently active RenderTarget within a given context
-    sf::Uint64 getUniqueId()
+    // A nested named namespace is used here to allow unity builds of SFML.
+    namespace RenderTargetImpl
     {
-        sf::Lock lock(mutex);
+        // Mutex to protect ID generation and our context-RenderTarget-map
+        sf::Mutex mutex;
 
-        static sf::Uint64 id = 1; // start at 1, zero is "no RenderTarget"
-
-        return id++;
-    }
-
-    // Map to help us detect whether a different RenderTarget
-    // has been activated within a single context
-    typedef std::map<sf::Uint64, sf::Uint64> ContextRenderTargetMap;
-    ContextRenderTargetMap contextRenderTargetMap;
-
-    // Check if a RenderTarget with the given ID is active in the current context
-    bool isActive(sf::Uint64 id)
-    {
-        ContextRenderTargetMap::iterator iter = contextRenderTargetMap.find(sf::Context::getActiveContextId());
-
-        if ((iter == contextRenderTargetMap.end()) || (iter->second != id))
-            return false;
-
-        return true;
-    }
-
-    // Convert an sf::BlendMode::Factor constant to the corresponding OpenGL constant.
-    sf::Uint32 factorToGlConstant(sf::BlendMode::Factor blendFactor)
-    {
-        switch (blendFactor)
+        // Unique identifier, used for identifying RenderTargets when
+        // tracking the currently active RenderTarget within a given context
+        sf::Uint64 getUniqueId()
         {
-            case sf::BlendMode::Zero:             return GL_ZERO;
-            case sf::BlendMode::One:              return GL_ONE;
-            case sf::BlendMode::SrcColor:         return GL_SRC_COLOR;
-            case sf::BlendMode::OneMinusSrcColor: return GL_ONE_MINUS_SRC_COLOR;
-            case sf::BlendMode::DstColor:         return GL_DST_COLOR;
-            case sf::BlendMode::OneMinusDstColor: return GL_ONE_MINUS_DST_COLOR;
-            case sf::BlendMode::SrcAlpha:         return GL_SRC_ALPHA;
-            case sf::BlendMode::OneMinusSrcAlpha: return GL_ONE_MINUS_SRC_ALPHA;
-            case sf::BlendMode::DstAlpha:         return GL_DST_ALPHA;
-            case sf::BlendMode::OneMinusDstAlpha: return GL_ONE_MINUS_DST_ALPHA;
+            sf::Lock lock(mutex);
+
+            static sf::Uint64 id = 1; // start at 1, zero is "no RenderTarget"
+
+            return id++;
         }
 
-        sf::err() << "Invalid value for sf::BlendMode::Factor! Fallback to sf::BlendMode::Zero." << std::endl;
-        assert(false);
-        return GL_ZERO;
-    }
+        // Map to help us detect whether a different RenderTarget
+        // has been activated within a single context
+        typedef std::map<sf::Uint64, sf::Uint64> ContextRenderTargetMap;
+        ContextRenderTargetMap contextRenderTargetMap;
 
-
-    // Convert an sf::BlendMode::BlendEquation constant to the corresponding OpenGL constant.
-    sf::Uint32 equationToGlConstant(sf::BlendMode::Equation blendEquation)
-    {
-        switch (blendEquation)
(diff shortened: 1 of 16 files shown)


SuperV1234:
Previous discussion here
https://github.com/SFML/SFML/pull/1787#issuecomment-819140669

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -52,95 +52,99 @@
 
 namespace
```
Bromeon, line 53: Would we still need anonymous namespaces? Their point _is_ internal linkage, and if we can't use internal linkage because of Unity builds, they serve no real purpose, or do they?

Or would it make sense to use `namespace sf` as the outer one, at the low chance that another library uses similar names?

src/SFML/Window/EglContext.cpp
```diff
@@ -44,44 +44,48 @@
 
 namespace
 {
-    EGLDisplay getInitializedDisplay()
+    // A nested named namespace is used here to allow unity builds of SFML.
+    namespace EGLContextImpl
```
Bromeon, line 48: should be `EglContextImpl`

Bromeon (commented):
Thanks for the contribution!

Is the best practice for unity builds really to manually avoid symbol conflicts? Since this seems to be an established technique (or workaround for compile times), it's a bit surprising that there is no automated tooling to rename internal linker symbols. Maybe because it could negate some of the performance gains...

Out of curiosity, did you measure the impact on compile times for SFML?

SuperV1234:
> Thanks for the contribution!
> 
> Is the best practice for unity builds really to manually avoid symbol conflicts? Since this seems to be an established technique (or workaround for compile times), it's a bit surprising that there is no automated tooling to rename internal linker symbols. Maybe because it could negate some of the performance gains...
> 
> Out of curiosity, did you measure the impact on compile times for SFML?

AFAIK, avoiding name collisions by renaming is the usual practice. I don't think it would be easy to automate that, given the complexity of a language like C++.

I have a basic full rebuild benchmark in the PR description. It shaves 4s off a full rebuild of SFML on my machine.

src/SFML/Graphics/RenderTarget.cpp
SuperV1234: We still need anonymous namespaces for non-unity builds. I think the suggestion of using `sf` is overkill, as the point of this PR is allowing SFML to be built as a unity build on its own, not mixed with other project files.

src/SFML/Window/EglContext.cpp
SuperV1234: Will change.

src/SFML/Graphics/RenderTarget.cpp
Bromeon: > We still need anonymous namespaces for non-unity builds. 

Why? Yes, theoretically a symbol is accessible outside its translation unit, but by following the de-duplication rules for unity builds, we automatically ensure there are no conflicts, no?

SuperV1234: Anonymous namespaces have other benefits. One that comes to mind is that compilers will warn about unused internal-linkage functions in a `.cpp` file, while they will not warn otherwise as they cannot prove the function is not being called from anywhere else.

Bromeon:
> AFAIK, avoiding name collisions by renaming is the usual practice. I don't think it would be easy to automate that, given the complexity of a language like C++.

Could be mangled with a path to the .cpp file being compiled (even if the symbol itself is in a header). Not sure if the linker already does something similar to deduplicate symbols with internal linkage (e.g. when LTCG is on and not every translation unit is completely isolated). Anyway, was just an off-topic question 🙂 

> I have a basic full rebuild benchmark in the PR description. It shaves 4s off a full rebuild of SFML on my machine.

Totally missed that, thanks! For 15-16s total build time, that is indeed significant.

binary1248 (approved):
Looks good to me.

I would still like to hear an opinion from @LaurentGomila regarding the naming (namespaces, `#define`s) and the frequent usage of `using` directives. Personally, I prefer fully qualified names myself even if it means the code ends up looking more verbose, but that is just my personal preference.

SuperV1234:
@binary1248 @LaurentGomila Any news on this?

I don't think the `using` directives are a big deal, as they're all in function-scope. I have not used any namespace-scope or class-scope directive.

binary1248:
I already reviewed and accepted the changeset. I was just interested to hear @LaurentGomila's opinion since the current style guide is based on his style. If there are no objections I wouldn't mind if it gets merged in the current form.

eXpl0it3r (approved):
I rebased onto master, added an unity build for VS 2019, fixed some overlap in the examples and squashed the commits.

With this setup, Unity builds currently just work on Windows.  
On Linux you get conflict with [X11's `None` macro](https://github.com/SFML/SFML/pull/1788/checks?check_run_id=3283535261) (and maybe more).  
On macOS there are [quite some more issues](https://github.com/SFML/SFML/runs/3283853149?check_suite_focus=true) and not even certain how well this is supported by Xcode.

eXpl0it3r:

eXpl0it3r:
Thanks @SuperV1234! 🙂 
If you ever need the unity build on Linux as well, feel free to open another PR

PR #1791, opened 2021-04-14
Fix several unused warnings

### LINKED ISSUE #1785: Minor warnings to fix (`-Wunused-parameter`, `-Wsign-compare`)
## Subject of the issue

There are some warnings produced when compiling SFML by source. I've looked at the code and they seem benign -- they just need to be suppressed.

## Your environment

* Windows 10, MSYS2 + MinGW
* SFML git master
* clang++ 10.x
* -Wall -Wextra -Wpedantic

## Steps to reproduce

Build.

---

```

src/SFML/Window/EglContext.cpp: In constructor ‘sf::priv::EglContext::EglContext(sf::priv::EglContext*, const sf::ContextSettings&, unsigned int, unsigned int)’:
src/SFML/Window/EglContext.cpp:163:36: warning: unused parameter ‘shared’ [-Wunused-parameter]
  163 | EglContext::EglContext(EglContext* shared, const ContextSettings& settings, unsigned int width, unsigned int height) :
      |                        ~~~~~~~~~~~~^~~~~~
src/SFML/Window/EglContext.cpp:163:67: warning: unused parameter ‘settings’ [-Wunused-parameter]
  163 | EglContext::EglContext(EglContext* shared, const ContextSettings& settings, unsigned int width, unsigned int height) :
      |                                            ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
src/SFML/Window/EglContext.cpp:163:90: warning: unused parameter ‘width’ [-Wunused-parameter]
  163 | EglContext::EglContext(EglContext* shared, const ContextSettings& settings, unsigned int width, unsigned int height) :
      |                                                                             ~~~~~~~~~~~~~^~~~~
src/SFML/Window/EglContext.cpp:163:110: warning: unused parameter ‘height’ [-Wunused-parameter]
  163 | EglContext::EglContext(EglContext* shared, const ContextSettings& settings, unsigned int width, unsigned int height) :
      |                                                                                                 ~~~~~~~~~~~~~^~~~~~



src/SFML/Window/Unix/GlxContext.cpp: In constructor ‘sf::priv::GlxContext::GlxContext(sf::priv::GlxContext*, const sf::ContextSettings&, const sf::priv::WindowImpl*, unsigned int)’:
src/SFML/Window/Unix/GlxContext.cpp:128:115: warning: unused parameter ‘bitsPerPixel’ [-Wunused-parameter]
  128 | GlxContext::GlxContext(GlxContext* shared, const ContextSettings& settings, const WindowImpl* owner, unsigned int bitsPerPixel) :



src/SFML/Window/Unix/WindowImplX11.cpp: In member function ‘void sf::priv::WindowImplX11::setVideoMode(const sf::VideoMode&)’:
src/SFML/Window/Unix/WindowImplX11.cpp:1338:33: warning: comparison of integer expressions of different signedness: ‘unsigned int’ and ‘int’ [-Wsign-compare]
 1338 |         if (res->modes[i].width == static_cast<int>(mode.width) &&
      |             ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/SFML/Window/Unix/WindowImplX11.cpp:1339:34: warning: comparison of integer expressions of different signedness: ‘unsigned int’ and ‘int’ [-Wsign-compare]
 1339 |             res->modes[i].height == static_cast<int>(mode.height))
      |             ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~



src/SFML/Graphics/RenderTextureImplFBO.cpp: In function ‘void {anonymous}::contextDestroyCallback(void*)’:
src/SFML/Graphics/RenderTextureImplFBO.cpp:80:39: warning: unused parameter ‘arg’ [-Wunused-parameter]
   80 |     void contextDestroyCallback(void* arg)
      |                                 ~~~~~~^~~

```

Bromeon:
Fixes part of #1785 (unused warnings).

## Tasks

* [ ] Tested on Linux
* [X] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## To be done

1. I don't know about the `EglContext` constructor (see issue discussion). We can remove it, but if we choose to keep it, someone else would need to help with the implementation. For those in the team, feel free to push directly to this branch.
2. I changed one argument `DefaultDepth(m_display, screen)` to what I think should be `static_cast<int>(bitsPerPixel)`. I took this from [the archaic X11 scrolls](https://tronche.com/gui/x/xlib/window/XCreateWindow.html), if someone knows X11 better and would like to review and/or test, that would be nice.
3. I'm not sure what to do with the `bitsPerPixel` parameter in the constructor `GlxContext::GlxContext(GlxContext* shared, const ContextSettings& settings, const WindowImpl* owner, unsigned int bitsPerPixel)`. I don't know if it makes sense to pass this in to an already existing window -- if it does, let me know. For now I commented the parameter out.

TLDR: review + testing on Unix needed.


binary1248:
`DefaultDepth(m_display, screen)` was purposely chosen because it minimizes (or even eliminates) the risk of the window not being able to be created because the requested bits per pixel is not provided by the X window system. GLX is the OpenGL extension to the X window system that provides surface properties via visuals. Unlike on Windows where the pixel format of the window is directly associated with the rendering surface, on Unix/Linux they are separate things. It sounds strange, but one can create a window with 8-bit depth and associate a 24-bit depth visual with it. The visual is the Unix equivalent of the pixel format on Windows and the available visuals will depend on GPU/driver support.

Bromeon:
@binary1248 Thanks for the elaboration!

The question is, should we deliberately fail if the provided `bitsPerPixel` value doesn't work (there's no real way to signal failure from a constructor without exceptions), or as you suggest silently ignore the value and do our best to create the window nevertheless? In the latter case, we should probably document that this parameter has no meaning for X11.

binary1248:
Window creation should really never fail (and thankfully it almost never does unless there is something really broken with the environment). If any Xlib call fails, the "standard" procedure is that Xlib ends up terminating the whole process with some cryptic message written to the console as can be seen often enough. Xlib error handling is... a mess. We really don't want this to happen under any circumstance. In the off chance it happens that the best surface format that we chose for the user doesn't match what they expect, we already emit a warning. As such they shouldn't be surprised either way.

Bromeon:
Reverted the use of `bitsPerPixel` on X11 and explicitly ignored parameters in `EglContext` constructor.

eXpl0it3r (approved):

eXpl0it3r (approved):

Bromeon:
I think it may need a few more rounds 😬 I'll do one more cycle now, after that feel free to push other commits on this branch.

eXpl0it3r:
Btw. you don't need to fix everything in this PR. It's fine if you just pick the things you want.
I've also started looking at some of the warnings, but I'll wait until you're done with the things you want to fix.

For the SoundWriters I implemented our own `toLower` function, since `std::tolower` takes `int` which then doesn't work in combination with `std::transform` and  `std::string`.

```cpp
namespace
{
    unsigned char toLower(unsigned char character)
    {
        return static_cast<unsigned char>(std::tolower(character));
    }
}
```

And for the `windows.h` header I used 

```cpp
#pragma warning(push, 0)
#include <windows.h>
#pragma warning(pop)
```

Bromeon:
Yes, as mentioned I won't push further for now, so feel free to add extra commits 🙂

src/SFML/System/Time.cpp
```diff
@@ -44,7 +44,7 @@ m_microseconds(0)
 ////////////////////////////////////////////////////////////
 float Time::asSeconds() const
 {
-    return m_microseconds / 1000000.f;
+    return static_cast<float>(static_cast<double>(m_microseconds) / 1000000.0);
```
Bromeon, line 47: Did this cause a warning in the first place?

Also, is double division and then cast to float more precise than float division?

src/SFML/Window/EglContext.cpp
```diff
@@ -328,21 +328,21 @@ void EglContext::updateSettings()
...
-    m_settings.stencilBits = tmp;
+    m_settings.stencilBits = static_cast<unsigned int>(tmp);
 
     eglCheck(result = eglGetConfigAttrib(m_display, m_config, EGL_SAMPLES, &tmp));
 
     if (result == EGL_FALSE)
         err() << "Failed to retrieve EGL_SAMPLES" << std::endl;
 
-    m_settings.antialiasingLevel = tmp;
+    m_settings.antialiasingLevel = static_cast<unsigned int>(tmp);
```
Bromeon, line 345: Not related to the change, but potential logic error:

If any of the `eglGetConfigAttrib()` calls fails, the result is still assigned to a `m_settings` member variable. I don't know what `eglGetConfigAttrib()` writes in such a case, but chances are that `tmp` is simply the last read value.

src/SFML/Window/Window.cpp
```diff
@@ -107,7 +107,7 @@ void Window::create(VideoMode mode, const String& title, Uint32 style, const Con
     // Check validity of style according to the underlying platform
     #if defined(SFML_SYSTEM_IOS) || defined(SFML_SYSTEM_ANDROID)
         if (style & Style::Fullscreen)
-            style &= ~Style::Titlebar;
+            style &= ~static_cast<Uint32>(Style::Titlebar);
```
Bromeon, line 110: This is strange. The negation needs `static_cast`, but the logical or just below doesn't?

src/SFML/Window/WindowImpl.cpp
```diff
@@ -208,7 +208,7 @@ void WindowImpl::processJoystickEvents()
                     Joystick::Axis axis = static_cast<Joystick::Axis>(j);
                     float prevPos = m_previousAxes[i][axis];
                     float currPos = m_joystickStates[i].axes[axis];
-                    if (fabs(currPos - prevPos) >= m_joystickThreshold)
+                    if (std::fabs(currPos - prevPos) >= m_joystickThreshold)
```
Bromeon, line 211: We should use `std::abs()` directly, as it's overloaded for `float` in C++.
This would be consistent as we don't use `std::sinf()` and other explicit C float variants, either.

src/SFML/System/Time.cpp
eXpl0it3r: Yes, because Uint64 and double have [different precision](https://stackoverflow.com/questions/30052710/why-double-can-store-bigger-numbers-than-unsigned-long-long)  if you look at the IEEE spec.

I'd assume so, if you have a very big integer and cast it to float, you will lose a lot of precision, while with a double, you'll have more precision.

But I'm open to suggestions, as I don't have too much experience with type conversions at this level. 😄

src/SFML/Window/Window.cpp
eXpl0it3r: Yeah, not sure why and I'm not even certain that this works correctly.
But you do have the different types of Uint32 for style and the enum being int, I assume by default.

Sounds to me like something we should align with C++17.

If you have a better way to do it let me know.

src/SFML/Window/WindowImpl.cpp
eXpl0it3r: Right, I somehow thought the float overload was only provided in C++11 and even checked cppreference, but apparently I can't read. 😄

Bromeon: It certainly doesn't help that [the `abs` page](https://en.cppreference.com/w/cpp/numeric/math/abs) only lists the integer overloads, and the float overloads are listed on [the `fabs` page](https://en.cppreference.com/w/cpp/numeric/math/fabs). I also had to check twice...

src/SFML/Window/Window.cpp
Bromeon: With C++17 we would probably use `enum class`, and then you'd need `static_cast` all the time for flag-like enums.
One neat idiom I've seen is overloading `operator+`:

```cpp
// instead of
static_cast<sf::Uint32>(Style::Titlebar) | static_cast<sf::Uint32>(Style::Resize)

// you would have
+Style::Titlebar | +Style::Resize
```

Or directly overload the operator for the enum values. There are also libraries like [enum-flags](https://github.com/grisumbras/enum-flags) trying to address this problem. But for now I think it's OK to just silence the warnings.

---

It's a bit a shame that `static_cast` is such a horrible syntactical choice for something occurring relatively often. It really makes code unncessarily verbose.

src/SFML/System/Time.cpp
Bromeon: Maybe we should test it for some very low and very high values. 
Increased precision would be a nice side effect of this refactoring 🙂

src/SFML/Network/TcpSocket.cpp
```diff
@@ -242,11 +242,11 @@ Socket::Status TcpSocket::send(const void* data, std::size_t size, std::size_t&
...
 
     // Loop until every byte has been sent
-    int result = 0;
-    for (sent = 0; sent < size; sent += result)
+    ssize_t result = 0;
+    for (sent = 0; sent < size; sent += static_cast<std::size_t>(result))
     {
         // Send a chunk of data
-        result = ::send(getHandle(), static_cast<const char*>(data) + sent, static_cast<int>(size - sent), flags);
+        result = ::send(getHandle(), static_cast<const char*>(data) + sent, size - sent, flags);
```
eXpl0it3r, line 249: Here any with all the other socket function we have a fun issue, that the signatures are different on Unix and on Windows.

The `size` parameter is taken as `int` [on Windows](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-send) and as `size_t` [on Linux](https://linux.die.net/man/2/send).

How do we want to solve this?

- Silence the warning somehow (anyone know how?)
- Use `#ifdef` for Unix and Windows
- Move the functions into the dedicated pImpl classes
- ???

Bromeon: There is already `priv::SocketImpl::AddrLength`, can we not use it?
```cpp
static_cast<priv::SocketImpl::AddrLength>(size - sent)
```

In case this type is not correct, let's add another typedef rather than outsourcing the function.

eXpl0it3r: `AddrLength` is differently defined for Unix, but I added a new typedef `Size`.
Let me know what you think!

src/SFML/Network/TcpSocket.cpp
```diff
@@ -242,11 +242,11 @@ Socket::Status TcpSocket::send(const void* data, std::size_t size, std::size_t&
...
 
     // Loop until every byte has been sent
-    int result = 0;
-    for (sent = 0; sent < size; sent += result)
+    ssize_t result = 0;
+    for (sent = 0; sent < size; sent += static_cast<std::size_t>(result))
     {
         // Send a chunk of data
-        result = ::send(getHandle(), static_cast<const char*>(data) + sent, static_cast<int>(size - sent), flags);
+        result = ::send(getHandle(), static_cast<const char*>(data) + sent, static_cast<priv::SocketImpl::Size>(size - sent), flags);
```
eXpl0it3r, line 249: Using a dedicated typedef, as suggested [here](https://github.com/SFML/SFML/pull/1791#discussion_r618142975)

src/SFML/Network/TcpSocket.cpp
Bromeon: Looks good, thanks! 👍🏼

examples/island/Island.cpp
```diff
@@ -273,10 +273,10 @@ float getElevation(float x, float y)
     for (int i = 0; i < perlinOctaves; i++)
     {
         elevation += stb_perlin_noise3(
-            x * perlinFrequency * std::pow(perlinFrequencyBase, i),
-            y * perlinFrequency * std::pow(perlinFrequencyBase, i),
+            x * perlinFrequency * static_cast<float>(std::pow(perlinFrequencyBase, i)),
+            y * perlinFrequency * static_cast<float>(std::pow(perlinFrequencyBase, i)),
             0, 0, 0, 0
-        ) * std::pow(perlinFrequencyBase, -i);
+        ) * static_cast<float>(std::pow(perlinFrequencyBase, -i));
```
Bromeon, line 279: Why is this necessary?

`perlinFrequencyBase` is `float`, `i` is `int`.
`pow(float, int)` returns `float` (until C++11).

I see that https://en.cppreference.com/w/cpp/numeric/math/pow mentions that the return value would be promoted to `double` since C++11. How is that not a breaking change in the C++ standard library?!

examples/island/Island.cpp
```diff
@@ -316,11 +326,11 @@ sf::Color getLowlandsTerrainColor(float moisture)
...
     sf::Color color =
         moisture < 0.27f ? sf::Color(240, 240, 180) :
-        moisture < 0.3f ? sf::Color(240 - 240 * (moisture - 0.27f) / 0.03f, 240 - 40 * (moisture - 0.27f) / 0.03f, 180 - 180 * (moisture - 0.27f) / 0.03f) :
+        moisture < 0.3f ? sf::Color(240 - static_cast<sf::Uint8>(240 * (moisture - 0.27f) / 0.03f), 240 - static_cast<sf::Uint8>(40 * (moisture - 0.27f) / 0.03f), 180 - static_cast<sf::Uint8>(180 * (moisture - 0.27f) / 0.03f)) :
         moisture < 0.4f ? sf::Color(0, 200, 0) :
-        moisture < 0.48f ? sf::Color(0, 200 - 40 * (moisture - 0.4f) / 0.08f, 0) :
+        moisture < 0.48f ? sf::Color(0, 200 - static_cast<sf::Uint8>(40 * (moisture - 0.4f) / 0.08f), 0) :
         moisture < 0.6f ? sf::Color(0, 160, 0) :
-        moisture < 0.7f ? sf::Color(34 * (moisture - 0.6f) / 0.1f, 160 - 60 * (moisture - 0.6f) / 0.1f, 34 * (moisture - 0.6f) / 0.1f) :
+        moisture < 0.7f ? sf::Color(static_cast<sf::Uint8>(34 * (moisture - 0.6f) / 0.1f), 160 - static_cast<sf::Uint8>(60 * (moisture - 0.6f) / 0.1f), static_cast<sf::Uint8>(34 * (moisture - 0.6f) / 0.1f)) :
```
Bromeon, line 333: While eliminating warnings, this makes code totally unreadable, and misses the point of the example.

It's a bit a weakness of C++ that a lot of valid code gets warnings. In theory, something like `short = short + short;` could already be problematic. And it's not like `static_cast` actually solves a real issue, it just silences the compiler.

Should we maybe add a `color()` function to the example that accepts `int`?
Or a `u8()` function as a shortcut for `static_cast<sf::Uint8>()`?
Or any other idea?

examples/island/Island.cpp
eXpl0it3r: Yeah, I was confused about this as well, but we don't really want to force people to use C++03 or disable warnings, so I guess while SFML isn't using C++11, we still should handle warnings.

I guess it sort of is a breaking change?

cmake/Macros.cmake
```diff
@@ -350,7 +350,7 @@ function(sfml_add_external)
             if (NOT include_dir)
                 message(FATAL_ERROR "No path given for include dir ${THIS_INCLUDE}")
             endif()
-            target_include_directories(${target} INTERFACE "$<BUILD_INTERFACE:${include_dir}>")
+            target_include_directories(${target} SYSTEM INTERFACE "$<BUILD_INTERFACE:${include_dir}>")
```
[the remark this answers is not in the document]
eXpl0it3r: Without telling the compiler, that these header are "system" (aka external library) headers, it would scan the thrid-party library headers as well and emit warnings from them.

examples/island/Island.cpp
eXpl0it3r: You mean, a `color()` function that accepts `float`s, right? I guess, I can do that.
On the other hand it's an advanced example focusing on how to use Vulkan, so a few `static_cast` mixed in there, shouldn't really bother anyone...

Bromeon: Yes, I meant `float`s.

True, but someone might be interested in the noise algorithm, and for that the `static_cast` is just distraction...

src/SFML/Network/TcpSocket.cpp
```diff
@@ -242,11 +242,14 @@ Socket::Status TcpSocket::send(const void* data, std::size_t size, std::size_t&
     }
 
     // Loop until every byte has been sent
-    int result = 0;
-    for (sent = 0; sent < size; sent += result)
+    ssize_t result = 0;
+    for (sent = 0; sent < size; sent += static_cast<std::size_t>(result))
     {
+        #pragma GCC diagnostic push
+        #pragma GCC diagnostic ignored "-Wuseless-cast"
```
eXpl0it3r, line 249: For Clang this is not needed, but GCC will complain about useless cast, but if you keep it, then Clang will complain about not knowing this warning, so it can't ignore it. So we can't use it, but also can't not use it.

I guess at this point using a `#if define(unix/windows)` is probably better...

examples/island/Island.cpp
Bromeon: Ah, so we're already running with C++11 settings, OK.
Yeah then the introduced change is OK (and I could live with it causing warnings on some older compilers).

Bromeon:
Should we not merge this? We can always open a new PR with further improvements.

But the longer this branch waits, the more stale it becomes and the higher the risk for extra effort due to merge conflicts.

binary1248 (approved):

Bromeon:
Is there anything left to do/review?

eXpl0it3r:
Fix the errors? A red CI pipeline isn't exactly useful to anyone.

Bromeon:
Expired logs aren't, either 😉 

Can you re-run CI?

eXpl0it3r:
Didn't see/know that build logs expire. I've rebase the branch and pushed it again, so you'll see the latest failures.
The unresolved review comments above also still need addressing.

eXpl0it3r:
Superseded by #1846

PR #1792, opened 2021-04-19
Enable compiler warnings.

binary1248:
Title says it all.

Currently not expected to build since the enabled warnings are stricter than the current requirements on the code.


eXpl0it3r:
Should we provide separate PR into master and then rebase this PR or how did you imagine to go about solving the produced errors?

binary1248:
Someone more daring than me should pick a method. 😁

cmake/CompilerWarnings.cmake
```diff
@@ -0,0 +1,88 @@
...
+    elseif(CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
+        set(FILE_WARNINGS ${CLANG_WARNINGS})
+    elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+        set(FILE_WARNINGS ${GCC_WARNINGS})
+    else()
+        message(AUTHOR_WARNING "No compiler warnings set for '${CMAKE_CXX_COMPILER_ID}' compiler.")
+    endif()
+
+    foreach(WARNING ${FILE_WARNINGS})
+        set_property(SOURCE ${ARGV} APPEND_STRING PROPERTY COMPILE_FLAGS " ${WARNING}")
```
eXpl0it3r, line 86: Shouldn't this be `target_compile_options(${TARGET} PRIVATE ${WARNING})` or maybe even a way to turn the list into space separated string.

eXpl0it3r:
Why might want to update the minimum required version to at least 3.15 or work around this bug: https://gitlab.kitware.com/cmake/cmake/-/issues/18317

Otherwise we'll get a lot of complaints about this warning. 😅

cmake/CompilerWarnings.cmake
binary1248: As the name of the function implies, we want to set warnings at a file-level. This is the least complicated way of being able to control which files we care about and which we don't.

eXpl0it3r: Hmm not sure if this works as expected, as the Windows header is still generating errors:

> `C:\Program Files (x86)\Windows Kits\8.1\include\um\combaseapi.h(229): error C2760: syntax error: unexpected token 'identifier', expected 'type specifier'`

https://ci.sfml-dev.org/#/builders/26/builds/94/steps/6/logs/stdio

binary1248: Yes... `set_property(SOURCE ... PROPERTY COMPILE_FLAGS ...)` can only set compiler flags that are used per compiler invocation, which is per TU (.cpp file after pre-processing). In order to exclude headers or disable warnings for specific headers (at least in MSVC) there is no other way than to wrap the corresponding header `#include`s in `#pragma`s. Rather painful process, but I've done it for my own projects often enough that I know you can survive it. 😁

binary1248:
> Why might want to update the minimum required version to at least 3.15 or work around this bug: https://gitlab.kitware.com/cmake/cmake/-/issues/18317
> 
> Otherwise we'll get a lot of complaints about this warning. 😅

Should be fixed with the `CMAKE_USER_MAKE_RULES_OVERRIDE` in the current changeset.

cmake/CompilerWarnings.cmake
eXpl0it3r: Do you think, we can make use of `/external` for the VS versions that support it? 

https://devblogs.microsoft.com/cppblog/broken-warnings-theory/

eXpl0it3r:
I think it makes the most sense if people would make PRs onto this branch, otherwise they don't actually see whether they've fixed all the issues.
Team members can also directly push to the branch.

cmake/CompilerWarnings.cmake
binary1248: Not a fan of relying on stuff that MS still considers as experimental. Until it is no longer experimental, it might break in random ways without near-term commitment from MS for a fix, see: https://gitlab.kitware.com/cmake/cmake/-/issues/17904

eXpl0it3r: It looks like this isn't even because of any or the warnings, but because of `/permissive`.
Can you and do we want to update the Windows SDK on the Windows build server?

See also: https://github.com/microsoft/krabsetw/issues/42

binary1248: I think we should just remove `/permissive-` for now. I don't want to think of the headaches supporting/guiding people that have issues because they are still using "broken" SDK versions.

binary1248:
Pushed changeset without `/permissive-`.

cmake/CompilerWarnings.cmake
```diff
@@ -0,0 +1,88 @@
...
+        -Wcast-align # warn for potential performance problem casts
+        -Wunused # warn on anything being unused
+        -Woverloaded-virtual # warn if you overload (not override) a virtual function
+        -Wpedantic # warn if non-standard C++ is used
+        -Wconversion # warn on type conversions that may lose data
+        -Wsign-conversion # warn on sign conversions
+        -Wnull-dereference # warn if a null dereference is detected
+        -Wdouble-promotion # warn if float is implicit promoted to double
+        -Wformat=2 # warn on security issues around functions that format output (ie printf)
+    )
```
SuperV1234, line 51: I would add `-Wimplicit-fallthrough`:

https://clang.llvm.org/docs/DiagnosticsReference.html#wimplicit-fallthrough

binary1248: Thanks for the tip, added. 😃

binary1248:
Added `-Wimplicit-fallthrough`.

cmake/CompilerWarnings.cmake
binary1248: @SuperV1234 Might want to suggest this upstream at https://github.com/lefticus/cpp_starter_project as well.

SuperV1234: @binary1248: good point, done!
https://github.com/lefticus/cpp_starter_project/pull/138

cmake/CompilerWarnings.cmake
```diff
@@ -0,0 +1,89 @@
...
+    elseif(CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
+        set(FILE_WARNINGS ${CLANG_WARNINGS})
+    elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+        set(FILE_WARNINGS ${GCC_WARNINGS})
+    else()
+        message(AUTHOR_WARNING "No compiler warnings set for '${CMAKE_CXX_COMPILER_ID}' compiler.")
+    endif()
+
+    foreach(WARNING ${FILE_WARNINGS})
+        set_property(SOURCE ${ARGV} APPEND_STRING PROPERTY COMPILE_FLAGS " ${WARNING}")
```
JonnyPtn, line 87: Could use `set_source_files_properties` here?

cmake/CompilerWarnings.cmake
```diff
@@ -0,0 +1,89 @@
+# from here:
+#
+# https://github.com/lefticus/cppbestpractices/blob/master/02-Use_the_Tools_Available.md
+
+# Helper function to enable compiler warnings for a specific set of files
+function(set_file_warnings)
```
JonnyPtn, line 6: is there a reason we set the options on a set of files rather than on the targets?

cmake/CompilerWarnings.cmake
binary1248: `set_property` lets us append to the existing property value in a single line.

cmake/CompilerWarnings.cmake
binary1248: We can select precisely for which files we want to enable our strict warnings (which are treated as errors and cause building to fail). There might be cases where we choose to compile some third party source as part of a target and don't want to enable our warnings for upstream code.

cmake/CompilerWarnings.cmake
eXpl0it3r: As mentioned elsewhere, this can't be easily resolved, because Clang only accepts `[[clang::fallthrough]]` and other compiler are okay with a `// fallthrough` comment, but would the of course error on the unknown/undefined attribute.

Options I see are:

- Create a custom macro that can be different for different compilers (do we really want to main this...)
- Remove the warning flag again until SFML 3/C++17

JonnyPtn: I’d say just wait for modern c++

binary1248: Removed `-Wimplicit-fallthrough` again for now.

binary1248:
Removed `-Wimplicit-fallthrough`.

eXpl0it3r:
Superseded, integrated and merged in #1846

PR #1795, opened 2021-05-03
Fix configuration conflicts with Clang

### LINKED ISSUE #1784: Building with Clang under a MinGW environment fails
## Subject of the issue

Building using `-DCMAKE_CXX_COMPILER=clang++` in a MinGW environment (e.g. installed via MSYS2) results in a compilation failure, because SFML's CMake files assume that when `SFML_COMPILER_CLANG` is `TRUE` then `clang-cl` is being used.

## Your environment

* Windows 10, using a MinGW x64 environment obtained via MSYS2
* `master` branch of SFML
* clang version 11
* no special compiler flags

## Steps to reproduce

1. Obtain [MSYS2](https://www.msys2.org/)
2. Open a MSYS shell and install `mingw-w64-x86_64-clang` via `pacman`
3. Open a MinGW 64-bit shell and

    ```bash
    git clone https://github.com/SFML/SFML
    cd SFML
    mkdir build
    cd build
    cmake .. -G"MinGW Makefiles" -DCMAKE_CXX_COMPILER=clang++
    ```

4. Observe the compilation failure -- SFML tries to get libraries from `libs-msvc-universal`

eXpl0it3r:
## Description

- On Windows you can use Clang with both MSVC and MinGW
- When using Clang with MSVC all the MSVC conditions should be fulfilled
- When using Clang with MinGW all the MSVC steps should not be run

This fixes #1784 

## Tasks

* [x] Tested Visual Studio build with MSVC
* [x] Tested Visual Studio build with Clang
* [x] Tested LLVM standalone with ninja (and Clang-cl)
* [x] Tested MinGW build with Clang

## How to test this PR?

Build SFML with your Clang compiler in the wanted environment

### Visual Studio with MSVC

```
mkdir build
cd build
cmake .. -G "Visual Studio 16 2019" -DCMAKE_INSTALL_PREFIX=<your custom install path>
cmake --build . --target install
```

### Visual Studio with Clang

- Install Clang as individual component in the VS installer
- Run in VS2019 command prompt

```
mkdir build
cd build
cmake .. -G "Visual Studio 16 2019" -T ClangCL -A x64 -DCMAKE_INSTALL_PREFIX=<your custom install path>
cmake --build . --target install
```


cmake/Config.cmake
```diff
@@ -115,6 +100,30 @@ elseif(MSVC)
         set(SFML_MSVC_VERSION 12)
     elseif(MSVC_VERSION EQUAL 1900)
         set(SFML_MSVC_VERSION 14)
+    elseif(MSVC_VERSION LESS_EQUAL 1919)
+        set(SFML_MSVC_VERSION 15)
+    elseif(MSVC_VERSION LESS_EQUAL 1929)
+        set(SFML_MSVC_VERSION 16)
+    endif()
```
Bromeon, line 107: According to [Wikipedia](https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B#Internal_version_numbering), MSVC++ version 14.* spans from `_MSC_VER` values 1900 (Visual Studio 2015 14.0) to 1929 (Visual Studio 2019 Version 16.9). 

Microsoft hasn't disappointed in adding even more parallel versioning schemes to confuse the hell out of the users.
For example, the latest version listed is `14.29 | 1929 (Visual Studio 2019 Version 16.9)`, which contains *four* (!) different versions:
* MSVC++ compiler version 14.29
* `_MSC_VER` 1929
* Visual Studio 2019 (product name)
* Visual Studio Version 16.9 (product version)

So the number `SFML_MSVC_VERSION` in our code seems to match the fourth of these version numbers, or at least the integral part of it. Do we actually gain anything from this, semantically -- i.e. is there a case differentiation based on this specific versioning scheme, out of the four?

I see we use `SFML_MSVC_VERSION GREATER_EQUAL 14` further below, but is this a more reliable/future-proof way rather than checking ` MSVC_VERSION GREATER_EQUAL 1900`? Or do we use it because we already maintain the version mapping anyway (which would also be OK).

Also, maybe let's link the Wikipedia article and document what this number is referring to, took me a while to figure that out...

cmake/Config.cmake
```diff
@@ -115,6 +100,30 @@ elseif(MSVC)
...
+elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER MATCHES "clang[+][+]")
```
Bromeon, line 108: Is the check for `CMAKE_CXX_COMPILER_ID` not enough? (e.g. https://stackoverflow.com/a/10055571 claims it is)

cmake/Config.cmake
eXpl0it3r: I think it's easier to deal with 14/15/16 instead of even more magic numbers, so I think the mapping still has its merits.
Historically, this mapping was essential, because every VS version required new extlibs builds, which isn't really necessary anymore since VS 2017.
In theory we could remove the old mappings and just have one for pre/post VS 2017

cmake/Config.cmake
eXpl0it3r: Could very well be. For a long time `CMAKE_CXX_COMPILER_ID` was not a thing and then after it was an undocumented thing, but now I believe it's fully documented, so the other part could be dropped.

cmake/Config.cmake
Bromeon: > I think it's easier to deal with 14/15/16 instead of even more magic numbers, so I think the mapping still has its merits.

My point was not to introduce yet another mapping, but work directly with `MSVC_VERSION` rather than `SFML_MSVC_VERSION`. But if most of the differences happen on the 14/15/16 boundaries, then the current approach is fine.


---

Taking a step back, this complexity is really insane. I mean, we're not only constantly doing manual case differentiations on compilers and versions, but we also need to deal with custom version mappings, warnings differing on each compiler, despite using the "meta build system" CMake with the promise to abstract from those very differences.

The C++ ecosystem has really become a mess.

eXpl0it3r: Yeah, I understood what you meant. It's not a huge benefit, but still makes it easier in our "own" CMake code to see 14/15/16 instead of random version numbers. Currently I only know of the two cases where it's really used, 1) for detecting if UCRT (universal CRT) should be used or the other one and 2) whether legacy_stdio_definitions.lib need to be linked.

----

Some stuff could potentially be simplified, by using more CMake specific functions. The initial modernization wasn't refactoring all parts of the scripts. Plus we try to make it as much compatible and easy to use as possible, so we allow certain configurations, that you won't necessarily find in other build setups. In the end, I still believe it's a lot easier to deal with than with like autotools or similar.

eXpl0it3r:
I've tested the three different variations that I could on my system and cross-checked with the people who reported the issue on Discord with standalone LLVM Clang and am currently just waiting on a) a review approval and b) testing for the MSYS2 setup test as mentioned in the linked issue.

cmake/Config.cmake
Bromeon: Absolutely, I think CMake is still the lesser evil, and for SFML users it's quite reasonable. 

It's just... once you look at other programming languages, you notice how extremely over-complicated C++ dependency management and build configuration are, and the whole toolchain fragmentation certainly doesn't help. Out of curiosity, I checked how much configuration code we have, hopefully with the right command:
```sh
git ls-files | grep -E "CMakeLists\\.txt|cmake/(Config|Macros)\\.cmake" | xargs wc -l
```
So 2.2 kLOC for CMake. You could write a decent little game in that amount of code 🙂

eXpl0it3r:
The latest changes broke the setup with MSYS2 + Clang. I wasn't able to setup such an environment, so I'm currently waiting for @SuperV1234 input.
But he already provided some fixes, that I will implement and maybe make it a bit clearer.

The issue is the combination "matrix" of how Clang can be used on Windows:

- In an MSVC environment with clang and clang-cl, requiring UCRT extlibs
- In a non MSVC environment with clang and clang-cl, requiring UCRT extlibs
- in a "GNU" environment with clang and ld (??), requiring MinGW extlibs

Issue #1232, opened 2017-05-17
Add support for MP3


dexter3k:
The longest-running [mp3- related patent](https://www.google.com/patents/US6009399) is **expired** as of 16th April, 2017. This makes `.mp3` patent-free in US too. Also licensing program [has been terminated](https://www.iis.fraunhofer.de/en/ff/amm/prod/audiocodec/audiocodecs/mp3.html).
So, why not to add support for it in the Audio module?

- [x] Implement decoder (reader): With [minimp3](https://github.com/lieff/minimp3), see #1851 ~~#1796~~
- [ ] Implement encoder (writer): See #2679

j4cobgarby:
Correct me if I'm wrong but as far as I can tell, the actual specification for the mp3 format costs almost 200 euros.

minirop:
not anymore. it's now in the public domain.

j4cobgarby:
Oh okay, if someone assigns me to this I'll have a go

MarioLiebisch:
Just do it, then create a pull request and refer to this issue by mentioning `#1232` in the text.

j4cobgarby:
Okay, on it

ricanteja:
How is development on this coming along @j4cobgarby? I might take it on myself if you take too long :)

j4cobgarby:
@ricanteja Feel free to have a go yourself - I've got a lot going on and probably won't be able to make much progress for a couple of weeks :)

ricanteja:
Okay so I've been spending way too much time on this and not really making any progress (while surviving a hurricane). I've looked into MAD, mpg123, LAME to try and figure this out but they all have their challenges. 
Challenge #1 is getting these libraries to build. Damn auto make. The build system is so convoluted good luck trying to figure out what broke.
Challenge #2 is licensing. MAD is GPL, while mpg123 and LAME are LGPL, but it doesn't matter because I can't build either one of them!

If anyone has any experience with GNU auto tools or has built mpg123 before please ping me in this thread. I'd love to get MP3 support in SFML but I'm starting to wonder if it is worth the headache. All this just to support 1 more file type?

eXpl0it3r:
So we only got GPL or LGPL as option? Adding another must have shared library dependency doesn't sound very exciting...

I've played around with GNU auto tools on Windows a bit for the other dependencies, but it really isn't very nice, but still doable if their automake file is useful.

ricanteja:
[...]

PR #1796, opened 2021-05-10
MP3 reader support

lieff:
This PR is related to the issue https://github.com/SFML/SFML/issues/1232 and adds mp3 reading support.
Also discussed on [the forum](https://en.sfml-dev.org/forums/index.php?topic=5959.msg39129#msg39129)
Example SFML/examples/sound/Sound.cpp is extended to play an mp3 file.

### DIFF
--- a/examples/X11/X11.cpp
+++ b/examples/X11/X11.cpp
@@ -3,7 +3,6 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Window.hpp>
-#include <SFML/System/Err.hpp>
 
 #define GLAD_GL_IMPLEMENTATION
 #include "gl.h"
--- a/include/SFML/Graphics/Font.hpp
+++ b/include/SFML/Graphics/Font.hpp
@@ -32,8 +32,6 @@
 #include <SFML/Graphics/Glyph.hpp>
 #include <SFML/Graphics/Texture.hpp>
 #include <SFML/Graphics/Rect.hpp>
-#include <SFML/System/Vector2.hpp>
-#include <SFML/System/String.hpp>
 #include <map>
 #include <string>
 #include <vector>
--- a/src/SFML/Graphics/RenderTextureImplDefault.cpp
+++ b/src/SFML/Graphics/RenderTextureImplDefault.cpp
@@ -29,7 +29,6 @@
 #include <SFML/Graphics/GLCheck.hpp>
 #include <SFML/Graphics/TextureSaver.hpp>
 #include <SFML/Window/Context.hpp>
-#include <SFML/System/Err.hpp>
 
 
 namespace sf
--- a/src/SFML/Graphics/Shape.cpp
+++ b/src/SFML/Graphics/Shape.cpp
@@ -28,7 +28,6 @@
 #include <SFML/Graphics/Shape.hpp>
 #include <SFML/Graphics/RenderTarget.hpp>
 #include <SFML/Graphics/Texture.hpp>
-#include <SFML/System/Err.hpp>
 #include <cmath>
 
 
--- a/src/SFML/Window/FreeBSD/JoystickImpl.cpp
+++ b/src/SFML/Window/FreeBSD/JoystickImpl.cpp
@@ -27,7 +27,6 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Window/JoystickImpl.hpp>
-#include <SFML/System/Err.hpp>
 #include <sys/stat.h>
 #include <dirent.h>
 #include <fcntl.h>
--- a/src/SFML/Window/JoystickImpl.hpp
+++ b/src/SFML/Window/JoystickImpl.hpp
@@ -30,7 +30,6 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Config.hpp>
 #include <SFML/Window/Joystick.hpp>
-#include <SFML/System/String.hpp>
 #include <algorithm>
 
 
--- a/src/SFML/Window/NetBSD/JoystickImpl.cpp
+++ b/src/SFML/Window/NetBSD/JoystickImpl.cpp
@@ -27,7 +27,6 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Window/JoystickImpl.hpp>
-#include <SFML/System/Err.hpp>
 #include <sys/stat.h>
 #include <dirent.h>
 #include <fcntl.h>
--- a/src/SFML/Window/Unix/InputImpl.cpp
+++ b/src/SFML/Window/Unix/InputImpl.cpp
@@ -28,7 +28,6 @@
 #include <SFML/Window/Window.hpp> // important to be included first (conflict with None)
 #include <SFML/Window/Unix/InputImpl.hpp>
 #include <SFML/Window/Unix/Display.hpp>
-#include <SFML/System/Err.hpp>
 #include <X11/Xlib.h>
 #include <X11/keysym.h>
 
--- a/src/SFML/Window/Win32/JoystickImpl.hpp
+++ b/src/SFML/Window/Win32/JoystickImpl.hpp
@@ -39,7 +39,6 @@
 #define DIRECTINPUT_VERSION 0x0800
 #include <SFML/Window/Joystick.hpp>
 #include <SFML/Window/JoystickImpl.hpp>
-#include <SFML/System/String.hpp>
 #include <windows.h>
 #include <mmsystem.h>
 #include <dinput.h>


src/SFML/Audio/SoundFileReaderMp3.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+    m_numSamples      = info.sampleCount;
+    return true;
+}
+
+
+////////////////////////////////////////////////////////////
+void SoundFileReaderMp3::seek(Uint64 sampleOffset)
+{
+    m_position = std::min(sampleOffset, m_numSamples);
```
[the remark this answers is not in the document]
lieff: There also conflict with min define from windows.h, using "(std::min)" seems resolves this problem. Updated the PR.

cmake/Modules/FindMinimp3.cmake
```diff
@@ -0,0 +1,14 @@
+#
```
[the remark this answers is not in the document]
lieff: Yeah, stb approach looks simpler, did not noticed it first time. Updated the PR.

examples/sound/Sound.cpp
```diff
@@ -93,6 +93,9 @@ int main()
     // Play music from a flac file
     playMusic("ding.flac");
 
+    // Play music from a mp3 file
+    playMusic("ding.mp3");
```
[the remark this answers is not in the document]
lieff: Added to asset_licenses.md, but not sure it it should be noticed that it's just re-encoded from ding.flac.

src/SFML/Audio/SoundFileReaderMp3.hpp
```diff
@@ -0,0 +1,131 @@
...
+//
+////////////////////////////////////////////////////////////
+
+#ifndef SFML_SOUNDFILEREADERMP3_HPP
+#define SFML_SOUNDFILEREADERMP3_HPP
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#include <minimp3/minimp3_ex.h>
```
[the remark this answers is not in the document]
lieff: Done. Updated the PR.

include/SFML/Audio/InputSoundFile.hpp
```diff
@@ -63,8 +63,9 @@ class SFML_AUDIO_API InputSoundFile : NonCopyable
     ////////////////////////////////////////////////////////////
     /// \brief Open a sound file from the disk for reading
     ///
-    /// The supported audio formats are: WAV (PCM only), OGG/Vorbis, FLAC.
+    /// The supported audio formats are: WAV (PCM only), OGG/Vorbis, FLAC, MP3.
     /// The supported sample sizes for FLAC and WAV are 8, 16, 24 and 32 bit.
+    /// For MP3 files with big (>16kb) APEv2 tag there low chance of garbage decoded at the end.
```
eXpl0it3r, line 68: The verb is missing: `... there is a low change ...`

lieff: Thanks) Updated the PR.

cmake/Modules/FindMinimp3.cmake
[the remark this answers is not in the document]
eXpl0it3r: It's simpler until you need to exclude it from the `SFML_USE_SYSTEM_DEPS` flag.

See here: https://github.com/SFML/SFML/blob/9e27096c3701004b97cf673692622349c8a9be92/CMakeLists.txt#L162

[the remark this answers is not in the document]
lieff: Hmm, CMAKE_IGNORE_PATH currently looks broken.
I've add message(${CMAKE_IGNORE_PATH}) after if(SFML_USE_SYSTEM_DEPS) block. And it prints very long line of all paths without spaces or other delimiters.

Looks like correct way is set(CMAKE_IGNORE_PATH "${CMAKE_IGNORE_PATH} ${DEP_DIR}") - then every path is separated by space.

CMakeLists.txt
```diff
@@ -174,14 +174,14 @@ if(SFML_USE_SYSTEM_DEPS)
...
-            set(CMAKE_IGNORE_PATH ${CMAKE_IGNORE_PATH} ${DEP_DIR})
+        if((NOT DEP_DIR MATCHES "/stb_image(/|$)") AND (NOT DEP_DIR MATCHES "/minimp3(/|$)"))
+            set(CMAKE_IGNORE_PATH "${CMAKE_IGNORE_PATH} ${DEP_DIR}")
         endif()
 
         get_filename_component(DEP_PARENT_DIR ${DEP_DIR} PATH)
         while(NOT DEP_PARENT_DIR STREQUAL "${CMAKE_SOURCE_DIR}/extlibs")
-            if(NOT DEP_DIR MATCHES "/stb_image(/|$)")
-                set(CMAKE_IGNORE_PATH ${CMAKE_IGNORE_PATH} ${DEP_PARENT_DIR})
+            if((NOT DEP_DIR MATCHES "/stb_image(/|$)") AND (NOT DEP_DIR MATCHES "/minimp3(/|$)"))
```
Bromeon, line 183: Could be simplified to 
```cmake
if(NOT DEP_DIR MATCHES "/(stb_image|minimp3)(/|$)")
```
Same above.

examples/sound/Sound.cpp
Bromeon, line 97: We play all the sounds/music at the same time, over each other?

include/SFML/Audio/InputSoundFile.hpp
```diff
@@ -63,8 +63,9 @@ class SFML_AUDIO_API InputSoundFile : NonCopyable
     ////////////////////////////////////////////////////////////
     /// \brief Open a sound file from the disk for reading
     ///
-    /// The supported audio formats are: WAV (PCM only), OGG/Vorbis, FLAC.
+    /// The supported audio formats are: WAV (PCM only), OGG/Vorbis, FLAC, MP3.
     /// The supported sample sizes for FLAC and WAV are 8, 16, 24 and 32 bit.
+    /// For MP3 files with big (>16kb) APEv2 tag there is a low chance of garbage decoded at the end.
```
Bromeon, line 68: This comment should be clarified. What is "garbage", why does it happen, how to avoid?

src/SFML/Audio/SoundFileReaderMp3.cpp
```diff
@@ -0,0 +1,129 @@
...
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#define MINIMP3_IMPLEMENTATION
+#include <SFML/Audio/SoundFileReaderMp3.hpp>
+#include <SFML/System/MemoryInputStream.hpp>
+#include <SFML/System/Err.hpp>
+#include <algorithm>
+
+static size_t read_cb(void* ptr, size_t size, void* data)
```
Bromeon, line 34: Function name violates SFML naming convention; use `readCallback` or so.

Instead of static methods, you could use an anonymous namespace.

Use `std::size_t` instead of `size_t`.

src/SFML/Audio/SoundFileReaderMp3.cpp
```diff
@@ -0,0 +1,129 @@
...
+{
+////////////////////////////////////////////////////////////
+bool SoundFileReaderMp3::check(InputStream& stream)
+{
+    char hdr[10];
+    if (stream.read(hdr, sizeof(hdr)) < sizeof(hdr))
+        return false;
+    if (!memcmp(hdr, "ID3", 3) && !((hdr[5] & 15) || (hdr[6] & 0x80) || (hdr[7] & 0x80) || (hdr[8] & 0x80) || (hdr[9] & 0x80)))
+        return true;
+    if (hdr_valid((const uint8_t *)hdr))
```
Bromeon, line 59: Don't use C casts. The question is, why isn't it declared as `Uint8 hdr[10];` in the first place? Then we also don't have the whole signed/unsigned issue.

src/SFML/Audio/SoundFileReaderMp3.cpp
```diff
@@ -0,0 +1,129 @@
...
+}
+
+namespace sf
+{
+namespace priv
+{
+////////////////////////////////////////////////////////////
+bool SoundFileReaderMp3::check(InputStream& stream)
+{
+    char hdr[10];
```
Bromeon, line 54: Maybe `header` for clarity, HDR (high dynamic range) is rather confusing.

src/SFML/Audio/SoundFileReaderMp3.cpp
```diff
@@ -0,0 +1,129 @@
...
+{
+    close();
+}
+
+
+////////////////////////////////////////////////////////////
+bool SoundFileReaderMp3::open(InputStream& stream, Info& info)
+{
+    // Init IO callbacks
+    m_io.read_data = m_io.seek_data = &stream;
```
Bromeon, line 86: I would split the assignment in two statements.

src/SFML/Audio/SoundFileReaderMp3.cpp
```diff
@@ -0,0 +1,129 @@
...
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#define MINIMP3_IMPLEMENTATION
```
Bromeon, line 28: Is this needed so the header file can be included in "implementation" mode? If so, maybe write a short comment, as it's unclear what this macro does and it's not used in the current file.

Bromeon (changes requested):
Thanks a lot for the contribution!

I noticed a few smaller issues, but overall it looks good 🙂

examples/sound/Sound.cpp
eXpl0it3r: They busy-loop inside the playMusic function 😉 

https://github.com/SFML/SFML/blob/master/examples/sound/Sound.cpp#L66-L75

CMakeLists.txt
lieff: Thanks) Updated the PR.

examples/sound/Sound.cpp
lieff: Blocking call, played one by one.

include/SFML/Audio/InputSoundFile.hpp
lieff: Can it be link to the issue comment? Added one with explanations.
Usually no need to avoid, chance is very low, just warning that APEv2 may not be removed and treated as mp3 data.

src/SFML/Audio/SoundFileReaderMp3.cpp
lieff: Done. Updated the PR.

src/SFML/Audio/SoundFileReaderMp3.cpp
lieff: Done. Updated the PR.

src/SFML/Audio/SoundFileReaderMp3.cpp
lieff: Done. Updated the PR.

src/SFML/Audio/SoundFileReaderMp3.cpp
lieff: Done. Updated the PR.

src/SFML/Audio/SoundFileReaderMp3.cpp
lieff: Added comment to defines from SoundFileReaderMp3.hpp as well. Updated the PR.

include/SFML/Audio/InputSoundFile.hpp
Bromeon: Ideally the documentation should be self-contained, also for people working offline. If the description is a few sentences, can you copy it into the doc comment?

examples/sound/Sound.cpp
Bromeon: I see, thanks!

src/SFML/Audio/SoundFileReaderMp3.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+namespace
+{
+std::size_t readCallback(void* ptr, std::size_t size, void* data)
+{
+    sf::InputStream* stream = static_cast<sf::InputStream*>(data);
+    return static_cast<std::size_t>(stream->read(ptr, size));
+}
+
+int seekCallback(uint64_t offset, void* data)
```
Bromeon, line 43: Sorry I missed that before, should be `std::uint64_t` (this applies to all symbols in the standard library, whether they have existed in C already or not -- compilers usually tolerate omitting it, but some compilers or linters may create warnings).

src/SFML/Audio/SoundFileReaderMp3.cpp
```diff
@@ -0,0 +1,134 @@
...
+}
+
+
+////////////////////////////////////////////////////////////
+SoundFileReaderMp3::SoundFileReaderMp3() :
+m_numSamples(0),
+m_position(0)
+{
+    m_io.read = readCallback;
+    m_io.seek = seekCallback;
```
Bromeon, line 75: The rest of `m_io` remains uninitialized, since its type is a POD (plain old data) without constructor. This might lead to UB if one of the `SoundFileReaderMp3` methods is called without a prior `open()` call, where `m_io` is fully initialized.

Also, `m_decoder` is not initialized.

include/SFML/Audio/InputSoundFile.hpp
lieff: https://github.com/SFML/SFML/issues/1232#issuecomment-641633168

This comment. OK, I try describe all in place.

src/SFML/Audio/SoundFileReaderMp3.cpp
```diff
@@ -0,0 +1,134 @@
...
+    toRead = mp3dec_ex_read(&m_decoder, samples, toRead);
+    m_position += toRead;
+    return toRead;
+}
+
+
+////////////////////////////////////////////////////////////
+void SoundFileReaderMp3::close()
+{
+    mp3dec_ex_close(&m_decoder);
```
Bromeon, line 129: It seems the only time `close()` is invoked is in the destructor, and the method is private.

Would it make sense to inline it directly (i.e. calling `mp3dec_ex_close(&m_decoder);` in the destructor)?
This would avoid accidentally calling `close()` twice if the implementation were to change in the future.

include/SFML/Audio/InputSoundFile.hpp
lieff: Updated the PR with full explanation.

Bromeon (changes requested):

src/SFML/Audio/SoundFileReaderMp3.cpp
lieff: Done. Updated the PR.

src/SFML/Audio/SoundFileReaderMp3.cpp
lieff: Added memset initialize. Done. Updated the PR.

src/SFML/Audio/SoundFileReaderMp3.cpp
lieff: Done. Updated the PR.

src/SFML/Audio/SoundFileReaderMp3.cpp
Bromeon: When the memory is zeroed, `seek()` and other operations are safe (i.e. fail properly)?
I don't know the implementation and assumptions of the underlying library.

src/SFML/Audio/SoundFileReaderMp3.cpp
Bromeon: This method can then be removed.

src/SFML/Audio/SoundFileReaderMp3.cpp
lieff: It should be fine, added tests for such case https://github.com/lieff/minimp3/commit/b18d274b998cd4406070ddc1f370f53392241af0 .
But calling open() method twice will lead to mem leak. Same for oundFileReaderOgg as I can see. So calling it twice forbidden? Or may be add protection against it?

src/SFML/Audio/SoundFileReaderMp3.cpp
lieff: Done. Updated the PR.

eXpl0it3r:
Woops, I screwed up, while trying to update the branch. Will have opened a new PR #1851 with the changes... 
Sorry about that. 😓

lieff:
Is something needed from my side? Looks like PR stuck on email delivery or something.

eXpl0it3r:
Nope, yeah seems like it somehow got lost, sorry about that.
I'll see to fix the typo as mentioned on the other PR and get the unity (not the engine) build to work.
If you want to take ownership of the PR again, then feel free to open one again, I messed up with updating the PR. 😓

Issue #1801, opened 2021-06-13
Proposal: Vertex Batch


DeltaNedas:
Everyone has tried to use our good friend `sf::Sprite` for a game, then it gets out of hand when a few hundred/thousand cripple performance.
Then you ask around on the forums and try out sf::VertexArray and your cool tilemap draws instantly!
Now you want to draw more dynamic things, and you're back to lots of draw calls or manually building VertexArrays which quickly becomes painful.
You can't add other drawables to a VertexArray.

I think that VertexArray/Buffer are fantastic, but their greatest limitation is that they can only be used to (easily) batch individual vertices.
You can't do something like this:
```c++
sf::VertexArray array;
sf::Sprite sprite;
array.append(sprite); // nope
```

This is where VertexBatch comes in.
VertexBatch is a thin wrapper over a VertexArray that when drawn to, simply appends the "drawn" vertices to be actually drawn later.
It would be both a Drawable (it has to be drawn *somewhere*) and a RenderTarget (why bother with it at all if you have to use a special set of drawSprite/drawArray/etc methods, when the RenderTarget::draw(Drawable) relationship already exists?)
Example code using it:
```c++
sf::RenderWindow window(...);
// Create a batch that wraps around a window (can be any RenderTarget, even another VertexBatch!)
sf::VertexBatch batch(window);

// Start of frame, clear old vertices
window.clear();
batch.clear();

// Queue some stuff to be drawn, won't actually draw it yet.
sf::CircleShape ellipse(10, 20);
ellipse.setPosition({200, 200});
// Transforms can be used too, but they'll be done on the CPU in batch.draw()
sf::Transform transform;
transform.scale(1, 0.5);
batch.draw(ellipse, transform);
sf::Text("i am very text", someFontHere);
batch.draw(text);

// Actually draw it! This just uses a few commands to transfer vertices to gpu, set gl states, draw them
window.draw(batch);
window.display();
```

Limitations:
- Only one `RenderStates` can be applied to the batch as a whole (1 texture, 1 shader) but we're used to this with `VertexArray`.
- No `VertexBuffer`s since the cost of copying them into CPU memory would outweigh the benefit of batched drawing, might upset someone if they don't expect batch.draw(someVertexBuffer) to draw immediately.
- Shapes with thousands of vertices will have to be transformed one-by-one on the CPU instead of on the GPU, but they should be using `VertexBuffer` anyways.

Currently writing an implementation of this.

winsvega:
[...]

PR #1802, opened 2021-06-13
Add draw batching

deltanedas:
## Description

Adds basic `sf::VertexBatch` support as described in issue #1801

## Tasks

* [X] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android
* [X] Converts not-triangles to triangles

## How to test this PR?

```cpp
#include <cstdio>

#include <SFML/Graphics.hpp>
#include <SFML/Window.hpp>

using namespace sf;

// use VertexBatch
#define BATCHED 1
// use square instead of circle, disable on integrated graphics to benchmark it
#define BAD_GPU 0

// Number of sprites to draw
constexpr int lagFactor = 2000;

int main() {
#if BATCHED
	RenderWindow window({1280, 720}, "Batched SFML");
	VertexBatch batch(window);
#else
	RenderWindow window({1280, 720}, "Unbatched SFML");
#endif

#if BAD_GPU
	// low poly thing to maximize effect of draw calls vs too many tris (hd graphics)
	sf::RectangleShape shape({50, 50});
	shape.setFillColor(sf::Color::Red);
#else
	sf::CircleShape shape(50, 24);
	shape.setFillColor(sf::Color::Blue);
#endif

	Event e;
	Clock clock;
	while (window.isOpen()) {
		while (window.pollEvent(e)) {
			if (e.type == sf::Event::Closed) {
				window.close();
			}
		}

		window.clear();
#if BATCHED
		batch.clear();
#endif

		for (int i = 0; i < lagFactor; i++) {
#if BATCHED
			batch.draw(shape);
#else
			window.draw(shape);
#endif
		}

#if BATCHED
		window.draw(batch);
		printf("Drawing %ld vertices\n", batch.getVertexCount());
#endif
		window.display();

		printf("%dms\n", clock.restart().asMilliseconds());
	}
}
```


DeltaNedas:
"Benchmark" results on my integrated graphics:
- BATCHED=0:
 - BAD_GPU=0: ~10-110ms, 45ms avg, extremely stuttery due to high volume of draw calls + vertex data
 - BAD_GPU=1: ~10-40ms, 15ms avg, pretty stuttery due to large number of draw calls
- BATCHED=1:
 - BAD_GPU=0: 144k verts @ ~20-40ms, 25ms avg
 - BAD_GPU=1: (UNCOMPENSATED!!!): 12k verts @ ~8-20ms, 11ms avg (lagFactor not multiplied by a compensation number so it's not accurate due to less gpu load)

In this test case, VertexBatch improved frametimes by about 40%, not bad considering it's literally a drop-in replacement if your drawing is centered around `RenderTarget&` and not using a global `window`.

While wouldn't use it for something regular like a tilemap (VertexBuffer is better to remove redundant copying), other things like particle effects or entities that move around would be a fantastic choice.

DeltaNedas:
second test, 20000 circle sprites unbatched and batched (compensated)
unbatched:
- range: 356-449ms
- average: 404ms
batched:
- range: 190-261ms (44% better)
- average: 213ms (47% better)

though at this point with 1.4M vertices it's more of a a ram throughput benchmark than anything
in the real world if i really needed 20000 circles that don't move i'd use a vertex buffer and most of this lag disappears

DeltaNedas:
in terms of draw calls they are actually the same (11/frame), it's just the internal batching management used by SFML is taking its time
so, if i can manage to make it *not*, then it'll be even slower with 20000 draw calls + a ton of geometry
seems that programmer-controlled batch is enough to create a significant performance boost, wow.

DeltaNedas:
works with RectangleShape and CircleShape, haven't tested anything else yet

cschreib:
Not an SFML maintainer here, but I just stumbled upon this and thought I'd comment. These are nice performance improvements, and I agree it is convenient to be able to dump vertices from sprites and other types of shapes in such a way, rather than having to rebuild vertex arrays by hand.

You mention that the implementation supports transforms, but actually `VertexBatch::draw()` does not use the provided render states at all, so the transforms cannot actually be applied. This is probably just a bug, as I see you do apply the transform in `batch()`, but you actually never pass the render state to it in `draw()`. I don't know how this can compile, so perhaps you are missing a commit or two? This should be easily fixed anyway.

Now, in my opinion, not offering support for multiple textures seems like a fundamental problem. Your API seems to allow it, but it actually doesn't support it. Even worse, it just silently ignores it. That's probably not going to go down well with anyone.

I would suggest two options:
 1. If you only support transforms from the render states, then don't pretend to do otherwise: just ask for a transform, and not a full render state. This means you can't inherit from `RenderTarget`, but I think that's for the best since you don't really implement that interface correctly.
 2. Add support for changing render states. For example, instead of inheriting from `VertexArray` (which is actually a code-smell to me, since `VertexArray` has no virtual interface to override), store a `std::vector<std::pair<VertexArray,RenderStates>>` (or something similar). In `draw()` you can detect if the new render state is compatible with the previous render state; if yes, just add to the last vertex array; if no, start a new vertex array. Then when the `VertexBatch` object itself is drawn, just draw the multiple vertex arrays in sequence. This way, you could also support all primitive types. And the cherry on the cake: `RenderTarget` could then own an `VertexBatch` instance, which it could use by default to optimize draw calls for everyone.

It seems to me that option 2 offers the most benefits, and it doesn't sound too hard to implement either.

LaurentGomila:
> Then when the VertexBatch object itself is drawn, just draw the multiple vertex arrays in sequence.
What about the initial drawing order? It seems to me that this approach would totally break it. Without any kind of depth testing (Z-buffer), this is a major limitation to take in account when implementing a batch system.

cschreib:
Drawing order would be preserved. You draw sprite A with its own texture, that goes in Vertex Array 1. You draw sprite B with another texture, that goes in Vertex Array 2. You draw sprite C with the same texture as A, that goes in Vertex Array 3 (does not reuse Vertex Array 1, even though it could, because that would indeed break rendering order).

Then when you render your batch, you render Vertex Array 1, then 2, then 3; the sprite rendering order is preserved. It won't give optimal performance, but this system would still allow automatic batching on consecutive draw calls that reuse the same states. It's up to the user to take the most advantage of it by ordering their draw calls, as with the current state caching system already in place in `RenderTarget`. It just goes a step further. The OP has already shown that there is a performance gain to be had.

I could see such a system being used with a texture atlas for example. If you have a single texture for your whole game, and no special shader/transform, you can draw your entire scene in a single draw call and still benefit from the convenience of SFML drawables.

It might give nice performance improvements by default, for users less experienced with Vertex Array and/or Vertex Buffer. In the long run I would still probably roll my own stuff with Vertex Buffer for optimal speed, though. But a mixture of VBO for big static stuff, and this for smaller dynamic objects, sounds fairly good.

DeltaNedas:
While I like the idea of supporting multiple textures/shaders per batch,
1. it offers no performance boost for atlases
2. it WILL break draw order
The latter is a massive issue which is why I can't add it.

The reason for not having VertexArray::draw(drawable, renderstates) take a transform instead is that it keeps rendertarget compatibility.
As long as you **use the same texture and shader**, you can just swap `drawEverything(window);` with `drawEverything(batch);` and it'll work.
A nice compromise is to add an explicit texture and shader to the VertexBatch constructor and add some asserts for debug builds to make sure users aren't being silly and using different ones. This isn't as clear as just taking a Transform, but it keeps RenderTarget pseudo-compatibility which is nice.

PR #1803, opened 2021-06-28
Fixed params to const reference and new operator

GermanAizek:

----

## Description

Simplest fixes related to parameter transfer and more correct syntax allocation of dynamic memory.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android


LaurentGomila:
I wouldn't call that "fixes".

Passing such small arguments by const ref instead of by value is not a fix, not even an improvement (please let's not start a fight about such micro-optimizations).

The "fix" on memory allocation is actually a new bug. `new int(1)` allocates a new integer with an initial value of 1. `new int[1]` allocates an array of one integer with undefined initial value.

Baardi:
Lmao that new[]-fix without delete[] was nonsensical.

PR #1806, opened 2021-06-30
Fixed funcs for safe use in Win32/Win64

GermanAizek:

----

## Description

Easy commit for better portability between Windown platforms.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/src/SFML/Window/Win32/WindowImplWin32.cpp
+++ b/src/SFML/Window/Win32/WindowImplWin32.cpp
@@ -343,7 +343,7 @@ void WindowImplWin32::setSize(const Vector2u& size)
     // SetWindowPos wants the total size of the window (including title bar and borders),
     // so we have to compute it
     RECT rectangle = {0, 0, static_cast<long>(size.x), static_cast<long>(size.y)};
-    AdjustWindowRect(&rectangle, GetWindowLong(m_handle, GWL_STYLE), false);
+    AdjustWindowRect(&rectangle, GetWindowLongPtr(m_handle, GWL_STYLE), false);
     int width  = rectangle.right - rectangle.left;
     int height = rectangle.bottom - rectangle.top;
 
@@ -500,8 +500,8 @@ void WindowImplWin32::switchToFullscreen(const VideoMode& mode)
     }
 
     // Make the window flags compatible with fullscreen mode
-    SetWindowLongW(m_handle, GWL_STYLE, WS_POPUP | WS_CLIPCHILDREN | WS_CLIPSIBLINGS);
-    SetWindowLongW(m_handle, GWL_EXSTYLE, WS_EX_APPWINDOW);
+    SetWindowLongPtr(m_handle, GWL_STYLE, WS_POPUP | WS_CLIPCHILDREN | WS_CLIPSIBLINGS);
+    SetWindowLongPtr(m_handle, GWL_EXSTYLE, WS_EX_APPWINDOW);
 
     // Resize the window so that it fits the entire screen
     SetWindowPos(m_handle, HWND_TOP, 0, 0, mode.width, mode.height, SWP_FRAMECHANGED);


GermanAizek (commented):
Dear @LaurentGomila and @Bromeon
Check it. Thx.

LaurentGomila:
Thanks.

You can give more details (why you changed that, and where it is documented) so that we don't have to guess.

What I found on the MSDN:

> Note  To write code that is compatible with both 32-bit and 64-bit versions of Windows, use GetWindowLongPtr. When compiling for 32-bit Windows, GetWindowLongPtr is defined as a call to the GetWindowLong function.

GermanAizek:
@LaurentGomila,
You have correctly found information on function, this pull request simplifies future migration between Windows platforms.
I also ask you to note that `GetWindowLong` will have the status of an deprecated function and possibly it will be deleted.

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -497,8 +497,9 @@ void WindowImplWin32::switchToFullscreen(const VideoMode& mode)
     }
 
     // Make the window flags compatible with fullscreen mode
-    SetWindowLongW(m_handle, GWL_STYLE, WS_POPUP | WS_CLIPCHILDREN | WS_CLIPSIBLINGS);
-    SetWindowLongW(m_handle, GWL_EXSTYLE, WS_EX_APPWINDOW);
+    SetWindowLongPtrW(m_handle, GWL_STYLE, WS_POPUP | WS_CLIPCHILDREN | WS_CLIPSIBLINGS);
+    SetWindowLongPtrW(m_handle, GWL_EXSTYLE, WS_EX_APPWINDOW);
+    ShowWindow(m_handle, SW_SHOW);
```
eXpl0it3r, line 502: Why is there an additional `ShowWindow()` call?

GermanAizek: @eXpl0it3r,
Did you check my PR in your displaying window was normal without ShowWindow()?

eXpl0it3r: I didn't, but if the mentioned function is a 1:1 replacement, then you shouldn't require additional code, right?

GermanAizek: @eXpl0it3r,
yes you u gd right

GermanAizek:
@eXpl0it3r, check here https://github.com/SFML/SFML/commit/5cc0c68fe8498de1a76c86ad7048491f4c883633

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -496,8 +496,8 @@ void WindowImplWin32::switchToFullscreen(const VideoMode& mode)
     }
 
     // Make the window flags compatible with fullscreen mode
-    SetWindowLongW(m_handle, GWL_STYLE, WS_POPUP | WS_CLIPCHILDREN | WS_CLIPSIBLINGS);
-    SetWindowLongW(m_handle, GWL_EXSTYLE, WS_EX_APPWINDOW);
+    SetWindowLongPtrW(m_handle, GWL_STYLE, WS_POPUP | WS_CLIPCHILDREN | WS_CLIPSIBLINGS);
+    SetWindowLongPtrW(m_handle, GWL_EXSTYLE, WS_EX_APPWINDOW);
```
eXpl0it3r, line 500: Only saw right now, it should be `SetWindowLongPtr` not `SetWindowLongPtrW`

eXpl0it3r:
The change shouldn't really make a difference because the window styles are of type `LONG` i.e. 32-bits values anyways. As such it doesn't matter whether you're building for 32-bits or 64-bits architecture, you will always be dealing with 32-bits values.

But it also doesn't hurt to use the recommended function. For example .NET only ever uses the `*Ptr` functions.

src/SFML/Window/Win32/WindowImplWin32.cpp
GermanAizek: @eXpl0it3r, check it https://github.com/SFML/SFML/commit/57150e3d05def23ae1d8943d32beb7a29dfc350f

eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:

eXpl0it3r:
Thanks for this fix! 🙂

Issue #1773, opened 2021-04-04
Allow using CoordinateType::Normalized in RenderTarget::draw


cschreib:
## Subject of the issue

I am maintaining a library which uses several back-ends for rending 2D graphics, and SFML is one of them. Recently, I started implementing optimized batching using vertex buffers. I am kind of stuck writing the SFML implementation for this, because `RenderTarget::draw()` always assumes that texture coordinates are given in pixels rather than normalized.

Unfortunately, in my library, when vertex buffers are created they know nothing of which texture will be used for rendering (and which size it will have). Thus the (back-end agnostic) interface expects normalized texture coordinates. There is no way to know, until the point of rendering, which texture will be used, hence which UV pixel coordinates should be sent to SFML. I could adjust this on the fly, but that would kill any performance gain from using vertex buffers in the first place.

I see that `Texture::bind()` allows specifying whether UV coordinates are in pixels or normalized, but as far as I know, to use `Texture::bind()` I would have to re-implement most of the OpenGL draw calls myself to get something rendered, as `RenderTarget` does not publicly expose functions like `applyTransform()`, `applyCurrentView()`, `applyBlendMode()`, etc., for me to reuse. That would defeat the purpose of using SFML for rendering as opposed to just using raw OpenGL.

My problem would be fixed if either of the following was implemented:
 - `RenderStates` is extended with a new field for specifying whether to bind the texture with normalized or pixel coordinates, rather than forcing `CoordinateType::Pixels`. That field would default to `Pixels` to preserve backward compatibility.
 - `RenderStates` is extended with a new field specifying the matrix transform to apply to texture coordinates (which gets combined with the matrix used internally to switch between pixel and normalised coordinates). By default this would be the identity, to preserve backward compatibility.
 - `RenderTarget::applyTexture`, delegates the choice of coordinate type to (to be added) `Drawable::getCoordinateType()` and `VertexBuffer::getCoordinateType()`. The type of coordinates to use is really a property of the drawable, not of the texture or the renderer. Unless specified, the coordinate type in a newly constructed drawable/vertex buffer would be `Pixels`, to preserve backward compatibility.

I am happy to make a PR to implement either solutions, if this is desirable.

## Your environment

* OS: Linux Mint 20.1
* SFML version: 2.5.1
* Compiler: clang++-10
* Special compiler flags used: none

## Steps to reproduce
N/A

## Expected behavior
N/A

## Actual behavior
N/A

Bromeon:
[...]

PR #1807, opened 2021-07-04
Add texture coordinate type to RenderStates

cschreib:
* [x] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before? (see https://github.com/SFML/SFML/issues/1773)

----

## Description

This pull request adds support for changing the texture coordinate type in `RenderStates`, so that `VertexArray` and `VertexBuffer` can be used with normalized texture coordinates. Other drawables still enforce texture coordinates in pixels. The rationale for this modification is provided in #1773.

This change is fully backward compatible. However, to keep the API clean and reduce header inclusion dependency, I decided to move the `CoordinateType` enum out of `sf::Texture` and into its own file. In the process, I deprecated using the `sf::Texture::CoordinateType` in `sf::Texture::bind()` and added an overload that uses `sf::CoordinateType`. I thought this would be neater in the long run, but if you think it's fine to just use `sf::Texture::CoordinateType` in `RenderStates` (which forces including the `Texture.hpp` header), and avoid the deprecation, I can revert that.

This PR is related to the issue #1773.

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode(sf::Vector2u{1280, 720}), "Minimal, complete and verifiable example");
    window.setFramerateLimit(60);

    sf::Texture texture;
    bool        success = texture.loadFromFile("../examples/assets/logo.png");

    // 'v1/s1' uses pixel coordinates
    sf::VertexArray v1(sf::PrimitiveType::Triangles, 6);
    v1[0].position = {0, 0};
    v1[1].position = {256, 0};
    v1[2].position = {256, 256};
    v1[3].position = {256, 256};
    v1[4].position = {0, 256};
    v1[5].position = {0, 0};

    v1[0].texCoords = {0, 0};
    v1[1].texCoords = {512, 0};
    v1[2].texCoords = {512, 512};
    v1[3].texCoords = {512, 512};
    v1[4].texCoords = {0, 512};
    v1[5].texCoords = {0, 0};

    sf::RenderStates s1;
    s1.texture        = &texture;
    s1.coordinateType = sf::CoordinateType::Pixels; // default

    // 'v2/s2' uses normalised coordinates
    sf::VertexArray v2(sf::PrimitiveType::Triangles, 6);
    v2[0].position = {256, 0};
    v2[1].position = {512, 0};
    v2[2].position = {512, 256};
    v2[3].position = {512, 256};
    v2[4].position = {256, 256};
    v2[5].position = {256, 0};

    v2[0].texCoords = {0, 0};
    v2[1].texCoords = {1, 0};
    v2[2].texCoords = {1, 1};
    v2[3].texCoords = {1, 1};
    v2[4].texCoords = {0, 1};
    v2[5].texCoords = {0, 0};

    sf::RenderStates s2;
    s2.texture        = &texture;
    s2.coordinateType = sf::CoordinateType::Normalized;

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
        }

        window.clear();

        window.draw(v1, s1);
        window.draw(v2, s2);

        window.display();
    }
}
```

### DIFF
--- a/include/SFML/Graphics/CoordinateType.hpp
+++ b/include/SFML/Graphics/CoordinateType.hpp
@@ -1,0 +1,43 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2023 Laurent Gomila (laurent@sfml-dev.org)
+//
+// This software is provided 'as-is', without any express or implied warranty.
+// In no event will the authors be held liable for any damages arising from the use of this software.
+//
+// Permission is granted to anyone to use this software for any purpose,
+// including commercial applications, and to alter it and redistribute it freely,
+// subject to the following restrictions:
+//
+// 1. The origin of this software must not be misrepresented;
+//    you must not claim that you wrote the original software.
+//    If you use this software in a product, an acknowledgment
+//    in the product documentation would be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such,
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+#pragma once
+
+namespace sf
+{
+
+////////////////////////////////////////////////////////////
+/// \ingroup graphics
+/// \brief Types of texture coordinates that can be used for rendering
+///
+/// \see sf::Texture::bind
+///
+////////////////////////////////////////////////////////////
+enum class CoordinateType
+{
+    Normalized, //!< Texture coordinates in range [0 .. 1]
+    Pixels      //!< Texture coordinates in range [0 .. size]
+};
+
+} // namespace sf
--- a/include/SFML/Graphics/RenderStates.hpp
+++ b/include/SFML/Graphics/RenderStates.hpp
@@ -30,6 +30,7 @@
 #include <SFML/Graphics/Export.hpp>
 
 #include <SFML/Graphics/BlendMode.hpp>
+#include <SFML/Graphics/CoordinateType.hpp>
 #include <SFML/Graphics/Transform.hpp>
 
 
@@ -93,13 +94,18 @@ struct SFML_GRAPHICS_API RenderStates
     ////////////////////////////////////////////////////////////
     /// \brief Construct a set of render states with all its attributes
     ///
-    /// \param theBlendMode Blend mode to use
-    /// \param theTransform Transform to use
-    /// \param theTexture   Texture to use
-    /// \param theShader    Shader to use
+    /// \param theBlendMode      Blend mode to use
+    /// \param theTransform      Transform to use
+    /// \param theCoordinateType Texture coordinate type to use
+    /// \param theTexture        Texture to use
+    /// \param theShader         Shader to use
     ///
     ////////////////////////////////////////////////////////////
-    RenderStates(const BlendMode& theBlendMode, const Transform& theTransform, const Texture* theTexture, const Shader* theShader);
+    RenderStates(const BlendMode& theBlendMode,
+                 const Transform& theTransform,
+                 CoordinateType   theCoordinateType,
+                 const Texture*   theTexture,
+                 const Shader*    theShader);
 
     ////////////////////////////////////////////////////////////
     // Static member data
@@ -110,10 +116,11 @@ struct SFML_GRAPHICS_API RenderStates
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    BlendMode      blendMode{BlendAlpha}; //!< Blending mode
-    Transform      transform;             //!< Transform
-    const Texture* texture{};             //!< Texture
-    const Shader*  shader{};              //!< Shader
+    BlendMode      blendMode{BlendAlpha};                  //!< Blending mode
+    Transform      transform;                              //!< Transform
+    CoordinateType coordinateType{CoordinateType::Pixels}; //!< Texture coordinate type
+    const Texture* texture{};                              //!< Texture
+    const Shader*  shader{};                               //!< Shader
 };
 
 } // namespace sf
@@ -123,10 +130,11 @@ struct SFML_GRAPHICS_API RenderStates
 /// \class sf::RenderStates
 /// \ingroup graphics
 ///
-/// There are four global states that can be applied to
+/// There are five global states that can be applied to
 /// the drawn objects:
 /// \li the blend mode: how pixels of the object are blended with the background
 /// \li the transform: how the object is positioned/rotated/scaled
+/// \li the texture coordinate type: how texture coordinates are interpreted
 /// \li the texture: what image is mapped to the object
 /// \li the shader: what custom effect is applied to the object
 ///
--- a/include/SFML/Graphics/Texture.hpp
+++ b/include/SFML/Graphics/Texture.hpp
@@ -29,6 +29,7 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Graphics/Export.hpp>
 
+#include <SFML/Graphics/CoordinateType.hpp>
 #include <SFML/Graphics/Rect.hpp>
 
 #include <SFML/Window/GlResource.hpp>
@@ -52,16 +53,6 @@ class Image;
 class SFML_GRAPHICS_API Texture : GlResource
 {
 public:
-    ////////////////////////////////////////////////////////////
-    /// \brief Types of texture coordinates that can be used for rendering
-    ///
-    ////////////////////////////////////////////////////////////
-    enum CoordinateType
-    {
-        Normalized, //!< Texture coordinates in range [0 .. 1]
-        Pixels      //!< Texture coordinates in range [0 .. size]
-    };
-
     ////////////////////////////////////////////////////////////
     /// \brief Default constructor
     ///
@@ -578,7 +569,7 @@ public:
     /// \param coordinateType Type of texture coordinates to use
     ///
     ////////////////////////////////////////////////////////////
-    static void bind(const Texture* texture, CoordinateType coordinateType = Normalized);
+    static void bind(const Texture* texture, CoordinateType coordinateType = CoordinateType::Normalized);
 
     ////////////////////////////////////////////////////////////
     /// \brief Get the maximum texture size allowed
--- a/src/SFML/Graphics/RenderStates.cpp
+++ b/src/SFML/Graphics/RenderStates.cpp
@@ -71,10 +71,12 @@ RenderStates::RenderStates(const Shader* theShader) : shader(theShader)
 ////////////////////////////////////////////////////////////
 RenderStates::RenderStates(const BlendMode& theBlendMode,
                            const Transform& theTransform,
+                           CoordinateType   theCoordinateType,
                            const Texture*   theTexture,
                            const Shader*    theShader) :
 blendMode(theBlendMode),
 transform(theTransform),
+coordinateType(theCoordinateType),
 texture(theTexture),
 shader(theShader)
 {
--- a/src/SFML/Graphics/Shape.cpp
+++ b/src/SFML/Graphics/Shape.cpp
@@ -236,6 +236,7 @@ void Shape::draw(RenderTarget& target, const RenderStates& states) const
     RenderStates statesCopy(states);
 
     statesCopy.transform *= getTransform();
+    statesCopy.coordinateType = CoordinateType::Pixels;
 
     // Render the inside
     statesCopy.texture = m_texture;
--- a/src/SFML/Graphics/Texture.cpp
+++ b/src/SFML/Graphics/Texture.cpp
@@ -862,7 +862,7 @@ void Texture::bind(const Texture* texture, CoordinateType coordinateType)
         glCheck(glBindTexture(GL_TEXTURE_2D, texture->m_texture));
 
         // Check if we need to define a special texture matrix
-        if ((coordinateType == Pixels) || texture->m_pixelsFlipped)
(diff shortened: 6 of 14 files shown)


eXpl0it3r:
Thanks a lot for this addition! 🙂 

We've already seen a lot of points from @binary1248 in the linked issue, but I'd like to hear some more thoughts from others. If you can bear with us a bit longer, we hopefully get this reviewed soon-ish. 😉

binary1248 (approved):

include/SFML/Graphics/RenderStates.hpp
```diff
@@ -76,6 +77,14 @@ class SFML_GRAPHICS_API RenderStates
     ////////////////////////////////////////////////////////////
     RenderStates(const Transform& theTransform);
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Construct a default set of render states with a custom texture coordinate type
+    ///
+    /// \param theCoordinateType Texture coordinate type to use
+    ///
+    ////////////////////////////////////////////////////////////
+    RenderStates(CoordinateType theCoordinateType);
```
Bromeon, line 86: Is this constructor actually useful or is it there just to be consistent with the other ones?

include/SFML/Graphics/CoordinateType.hpp
```diff
@@ -0,0 +1,46 @@
...
+/// \brief Types of texture coordinates that can be used for rendering
+///
+/// \see sf::Texture::bind
+///
+////////////////////////////////////////////////////////////
+enum CoordinateType
+{
+    Normalized, //!< Texture coordinates in range [0 .. 1]
+    Pixels      //!< Texture coordinates in range [0 .. size]
+};
```
Bromeon, line 42: This spills the constants to the `sf` namespace, i.e. `sf::Normalized` and `sf::Pixels`. Is that what we want (like with [`PrimitiveType`](https://www.sfml-dev.org/documentation/2.5.1/group__graphics.php#ga5ee56ac1339984909610713096283b1b))?

include/SFML/Graphics/Texture.hpp
```diff
@@ -52,6 +53,9 @@ class SFML_GRAPHICS_API Texture : GlResource
     ////////////////////////////////////////////////////////////
     /// \brief Types of texture coordinates that can be used for rendering
     ///
+    /// This enum is present for backward compatibility only. Prefer using
+    /// sf::CoordinateType instead.
+    ///
```
Bromeon, line 58: What about changing this `enum` to a `typedef` referring to `sf::CoordinateType`?
Then we don't have to deprecate `bind()`, just this typedef (not sure if possible with `SFML_DEPRECATED`).

include/SFML/Graphics/Texture.hpp
```diff
@@ -569,7 +573,16 @@ class SFML_GRAPHICS_API Texture : GlResource
     /// \param coordinateType Type of texture coordinates to use
     ///
     ////////////////////////////////////////////////////////////
-    static void bind(const Texture* texture, CoordinateType coordinateType = Normalized);
+    static void bind(const Texture* texture, sf::CoordinateType coordinateType = sf::CoordinateType::Normalized);
```
Bromeon, line 576: `sf::CoordinateType::Normalized` is C++11, we should use `sf::Normalized`.

src/SFML/Graphics/RenderStates.cpp
```diff
@@ -41,61 +41,91 @@ const RenderStates RenderStates::Default(BlendMode(
...
 
 ////////////////////////////////////////////////////////////
 RenderStates::RenderStates(const BlendMode& theBlendMode) :
-blendMode(theBlendMode),
-transform(),
-texture  (NULL),
-shader   (NULL)
+blendMode     (theBlendMode),
+transform     (),
+coordinateType(CoordinateType::Pixels),
```
Bromeon, line 79: C++11, see above

src/SFML/Graphics/Texture.cpp
```diff
@@ -742,7 +742,7 @@ void Texture::bind(const Texture* texture, CoordinateType coordinateType)
         glCheck(glBindTexture(GL_TEXTURE_2D, texture->m_texture));
 
         // Check if we need to define a special texture matrix
-        if ((coordinateType == Pixels) || texture->m_pixelsFlipped)
+        if ((coordinateType == sf::Pixels) || texture->m_pixelsFlipped)
```
Bromeon, line 745: In case we go for the typedef, this change could be reverted.

Bromeon (commented):
Thanks for the PR!
Added a few comments and questions 🙂

include/SFML/Graphics/RenderStates.hpp
cschreib: I only added it for consistency; I doubt it would be of much use on its own without providing also a texture. Would you rather remove it?

include/SFML/Graphics/CoordinateType.hpp
cschreib: The right thing would be to use C++11 `enum class`, but I think that's not an option for SFML. To prevent pollution, we could do something like
```c++
namespace Coordinate
{
    enum Type
    {
        Normalized,
        Pixels
    };
}
```

That will still allow implicit conversions, which `enum class` wouldn't allow, but it's less of an issue. I have no objection going either way.

include/SFML/Graphics/Texture.hpp
cschreib: That is what I did at first, and discovered that indeed the deprecation propagates to the typedef too :(

include/SFML/Graphics/CoordinateType.hpp
cschreib: Or, with more boilerplate ([kind of like this](https://musingstudio.com/2014/05/20/how-to-emulate-c11s-enum-class-in-c03/)):
```c++
class CoordinateType
{
public:

    static const CoordinateType Normalized;
    static const CoordinateType Pixels;

    CoordinateType() {}
    explicit CoordinateType(int theValue) : value(theValue) {}
    operator int() const { return value; }

private:
 
    int value;
};
```

But this may start to become a bit obscure.

include/SFML/Graphics/Texture.hpp
cschreib: My C++03 is so rusty! Fixed.

src/SFML/Graphics/RenderStates.cpp
cschreib: My C++03 is so rusty! Fixed.

cschreib:
Thank you for the comments @Bromeon, I have replied to all of them.

include/SFML/Graphics/Texture.hpp
Bromeon: I would still do it, we can also just mention "deprecated" in the docs.
There was no standard deprecation mechanism in C++ until recently, so we have to live with the options available...

include/SFML/Graphics/CoordinateType.hpp
Bromeon: I wouldn't overdo it, both the simple `enum` and the `Coordinate` namespace are good for me. Since this will likely become a C++11 `enum class` in SFML 3, we can also keep things as is for now. I was more asking to make sure everyone is OK with the symbols being in `sf` namespace.

include/SFML/Graphics/RenderStates.hpp
Bromeon: I tend to have the mindset "if it's not useful, don't add it" as we can always add things later, but not remove them. How do others see this?

include/SFML/Graphics/Texture.hpp
Bromeon: 🦀

cschreib:
The last round of comments from @Bromeon was asking for opinion from other maintainers, but there has been no reply since. Should we still wait for these comments, or should I just go ahead an implement the changes suggested by @Bromeon?

include/SFML/Graphics/Texture.hpp
cschreib: I've re-targeted this PR against `master` now, so I just removed the old function since we don't need to preserve backward compatibility in 3.0.

src/SFML/Graphics/Texture.cpp
cschreib: Done

cschreib:
I have re-targeted this PR against master, hence:
 - Using C++17 to improve the code.
 - Remove functions without care for breaking backward compatibility.

This addresses all the remaining comments from @Bromeon.

include/SFML/Graphics/CoordinateType.hpp
```diff
@@ -0,0 +1,46 @@
...
+//    in the product documentation would be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such,
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+#ifndef SFML_COORDINATETYPE_HPP
```
Bambo-Borris, line 25: Pretty sure SFML master now uses `#pragma once` instead of the C era header guards, worth changing to be consistent with the rest of master.

cschreib: Indeed, done. I also noticed the copyright year was out of date (this PR is old!), so I fixed that too.

eXpl0it3r:

cschreib:
If there are no further comments on this PR and the CI passes, are there any remaining blockers for it be merged? I have been maintaining this branch for close to two years now :)

src/SFML/Graphics/Shape.cpp
```diff
@@ -192,6 +192,7 @@ void Shape::draw(RenderTarget& target, const RenderStates& states) const
     RenderStates statesCopy(states);
 
     statesCopy.transform *= getTransform();
+    statesCopy.coordinateType = CoordinateType::Pixels;
```
eXpl0it3r, line 195: Could the state not be overwritten? when calling draw?

cschreib: Hmm there are two ways I can understand your question, but in both cases the answer is no :)
 - If you mean "the state could be overwritten in the `target.draw()` call", then no because `RenderTarget::draw()` takes the state as a const reference, and does not modify it.
 - If you mean "we should not overwrite the state in Shape::draw()", then I'd say we have to, because the interface of `Shape` imposes integer texture coordinates (see `Shape::setTextureRect()`, `Shape::updateTexCoords()`). So the class would break if a user asked for different coordinate type. We could check and assert instead of silently overwrite though, but I'm not sure it's best.

eXpl0it3r: What happens if I call `window.draw(shape, sf::CoordinateType::Normalized)`?

cschreib: As currently implemented, this is equivalent to `window.draw(shape)`. It calls `Shape::draw()` which overrides the coordinate type to `Pixels`.

It's probably an error to do this, because `Shape` in its current form cannot work with normalized coordinates. There's nothing fundamental about it, we could update all SFML classes to also support normalized coordinates. I left this out of this PR because there wasn't demand for it (in the issue https://github.com/SFML/SFML/issues/1773 it was made clear that pixel coordinates must remain the default), and also to keep the PR focused on the core capability.

eXpl0it3r: Hmm, I guess that makes sense. I wonder though whether we should have an assert if someone does choose to pass normalized as type 🤔

cschreib: Thinking about this some more; I'm not sure the assert or `sf::err` is necessary, because there is already a precedent for silently overwriting the render state without warning. Indeed this is already done today for the texture:
https://github.com/SFML/SFML/blob/1e9161b57e475764e91e1efd9bde80102f16618c/src/SFML/Graphics/Shape.cpp#L197
https://github.com/SFML/SFML/blob/1e9161b57e475764e91e1efd9bde80102f16618c/src/SFML/Graphics/Shape.cpp#L203
https://github.com/SFML/SFML/blob/1e9161b57e475764e91e1efd9bde80102f16618c/src/SFML/Graphics/Sprite.cpp#L138
https://github.com/SFML/SFML/blob/1e9161b57e475764e91e1efd9bde80102f16618c/src/SFML/Graphics/Text.cpp#L374

If someone were to call `window.draw(sprite, someOtherTexture)`, the call would ignore `someOtherTexture` and just use the texture set up inside `sprite`, without warning. So it seems there is already an implicit contract that render states may be overwritten inside `draw()`, if it is necessary to display the `Drawable` as intended.

eXpl0it3r:

eXpl0it3r (approved):
Sorry for all the delays! 😔

Looks good to me!

include/SFML/Graphics/CoordinateType.hpp
```diff
@@ -0,0 +1,43 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2023 Laurent Gomila (laurent@sfml-dev.org)
```
ChrisThrasher, line 4:
```suggestion
// Copyright (C) 2007-2024 Laurent Gomila (laurent@sfml-dev.org)
```

ChrisThrasher (approved):

eXpl0it3r:
Thank you for creating this PR and keep it alive over time! 🙂

Issue #353, opened 2013-02-15
Support Retina Displays and other High DPI Displays


mjbshaw:
I have a MacBook Pro with Retina Display. Creating an 800x600 window (for example, building and running the default sample code in a new SFML project in Xcode) results in a 1600x1200 window with a scaling factor 2 applied (much like any window created that doesn't support Apple's Retina Display, typically windows not using Cocoa or not setting NSHighResolutionCapable to true in the Info.plist file). This results in a very pixelated look.

If SFML could support high DPI displays, graphics would look sharper.

One potential issue with supporting high DPI displays is maintaining support for low DPI displays. Some apps may do their calculations in pixels, and thus on a high DPI display their end results are much smaller than they are on low DPI displays. A potential solution to this problem is providing a method in SFML to obtain the window's DPI and requiring the developer to make their app DPI independent however they may choose to do so. If a user would rather not handle DPI independence and would prefer the graphics to just be scaled up (like they currently are) on high DPI displays, that could also be provided as an option in SFML (I think).

This issue will become even more important if SFML supports mobile platforms like iOS or Android in the future.

PR #1817, opened 2021-08-15
Idea to improve high DPI handling on OS X

jqdg:
* [x] Forum posts: [1](https://en.sfml-dev.org/forums/index.php?topic=17092.0), [2](https://en.sfml-dev.org/forums/index.php?topic=18840.0), [3](https://en.sfml-dev.org/forums/index.php?topic=17354.0). Issues and PRs: #353, #388

----

### Problem Description

SFML doesn't really have complete support for DPI awareness (it is my understanding that this is planned for SFML version 3).

This can be an issue on MacBooks with retina displays (among other situations). There are two possible cases:

1. `NSHighResolutionCapable` is set to true, and all windows appear to be half the size in each dimension as expected.
2. `NSHighResolutionCapable` is set to false in the Info.plist file, and the whole app renders in low resolution.

Both of these cases are highly problematic.

In case 1, by default, all windows and graphics will be drawn far too small to be usable. In order to render at the correct size, the application developer must scale all of the window and graphics sizes by the correct scale. However, SFML does not expose an API to identify if the display is high DPI, and to return the correct scaling. This is a fairly invasive change to the application code, and will require platform-specific code to identify the display scaling.

In case 2, everything draws at the correct size, but the window decorations and fonts are blurry. Furthermore, if one uses a raw executable (rather than application bundle) then it is not possible to set this option using Info.plist.

| `NSHighResolutionCapable` true | `NSHighResolutionCapable` false |
| --- | --- |
| high_dpi | low_dpi |

---

### Proposed Solution

I believe that the **default** behavior should be as follows:

* The apparent window sizes should be the same on high DPI displays and low DPI displays
* Window decorations and fonts should be rendered in high resolution
* Graphics should be rendered at the same apparent size on high DPI and low DPI displays (without making SFML DPI-aware, which is a much bigger project, this means that graphics will be slightly pixelated on high DPI displays)

**If** the user wants to avoid the scaled rendering, and instead put in the extra effort to support high DPI rendering, then they can **opt-in** by setting a high DPI window style.

This PR accomplishes this by adding a new window style `HighDPI` (which has an effect only on Mac OS, but this could potentially be extended to other platforms). This style is not set by default. This will cause the window and graphics to render at the same apparent size on high DPI and low DPI displays, while retaining high-resolution window decorations and fonts. This is because if the style is not set, then `setWantsBestResolutionOpenGLSurface` is set to `NO` in the `SFOpenGLView`.

On the other hand, if the application developer wants to avoid possible pixelation from scaling the window, they can set the `HighDPI` window style, which will call `setWantsBestResolutionOpenGLSurface` to `YES`, and also enable scaling of window dimensions by the display factor. The developer will then need to manually scale all of the windows and graphics sizes in their rendering code to render everything at 2x.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::Uint32 style = sf::Style::Default;
    // Uncomment the next line to enable high DPI mode:
    // style |= sf::Style::HighDPI;
    const char *title = (style & sf::Style::HighDPI) ? "High DPI" : "Default";
    sf::RenderWindow window(sf::VideoMode(300, 300), title, style);
    window.setFramerateLimit(60);
    sf::CircleShape circle(150, 100);
    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
        }
        window.clear();
        window.draw(circle);
        window.display();
    }
}
```

| Default mode | High DPI mode | Low DPI mode (previous) |
| --- | --- | --- |
| Screen Shot 2021-08-15 at 1 50 52 PM | Screen Shot 2021-08-15 at 1 51 02 PM | Screen Shot 2021-08-15 at 1 51 19 PM |

From the above screenshots, we can see that the "default mode" preserves the advantages of setting `NSHighResolutionCapable` to be false (everything renders at the correct size), while avoiding the blurry window decorations and fonts. Additionally, the application developer still retains the option to render everything in high resolution if desired.

### DIFF
--- a/src/SFML/Window/OSX/SFOpenGLView.h
+++ b/src/SFML/Window/OSX/SFOpenGLView.h
@@ -78,6 +78,7 @@ namespace sf {
     BOOL                          m_cursorGrabbed;  ///< Is the mouse cursor trapped?
     CGFloat                       m_deltaXBuffer;   ///< See note about cursor grabbing above
     CGFloat                       m_deltaYBuffer;   ///< See note about cursor grabbing above
+    BOOL                          m_highDpi;        ///< Is high-DPI enabled?
 
     // Hidden text view used to convert key event to actual chars.
     // We use a silent responder to prevent sound alerts.
@@ -99,7 +100,7 @@ namespace sf {
 /// \return an initialized view
 ///
 ////////////////////////////////////////////////////////////
--(id)initWithFrame:(NSRect)frameRect fullscreen:(BOOL)isFullscreen;
+-(id)initWithFrame:(NSRect)frameRect fullscreen:(BOOL)isFullscreen highDpi:(BOOL)isHighDpi;
 
 ////////////////////////////////////////////////////////////
 /// \brief Finish the creation of the SFML OpenGL view
--- a/src/SFML/Window/OSX/SFOpenGLView.mm
+++ b/src/SFML/Window/OSX/SFOpenGLView.mm
@@ -86,11 +86,17 @@
 ////////////////////////////////////////////////////////
 -(id)initWithFrame:(NSRect)frameRect
 {
-    return [self initWithFrame:frameRect fullscreen:NO];
+    return [self initWithFrame:frameRect fullscreen:NO highDpi:NO];
 }
 
 ////////////////////////////////////////////////////////
 -(id)initWithFrame:(NSRect)frameRect fullscreen:(BOOL)isFullscreen
+{
+    return [self initWithFrame:frameRect fullscreen:isFullscreen highDpi:NO];
+}
+
+////////////////////////////////////////////////////////
+-(id)initWithFrame:(NSRect)frameRect fullscreen:(BOOL)isFullscreen highDpi:(BOOL)isHighDpi
 {
     if ((self = [super initWithFrame:frameRect]))
     {
@@ -118,8 +124,12 @@
         m_hiddenTextView = [[NSTextView alloc] initWithFrame:NSZeroRect];
         [m_hiddenTextView setNextResponder:m_silentResponder];
 
-        // Request high resolution on high DPI displays
-        [self setWantsBestResolutionOpenGLSurface:YES];
+        // If high DPI is requested, then use high resolution for the OpenGL view.
+        // Currently, isHighDpi is always expected to be NO, and so the OpenGL view will render scaled.
+        // Note: setWantsBestResolutionOpenGLSurface requires YES to work properly, rather than simply non-zero value.
+        // isHighDpi is an Objective-C BOOL, which can have non-zero values other than YES.
+        m_highDpi = isHighDpi ? YES : NO;
+        [self setWantsBestResolutionOpenGLSurface:m_highDpi];
 
         // At that point, the view isn't attached to a window. We defer the rest of
         // the initialization process to later.
@@ -224,7 +234,7 @@
     NSWindow* window = [self window];
     NSScreen* screen = window ? [window screen] : [NSScreen mainScreen];
     CGFloat oldScaleFactor = m_scaleFactor;
-    m_scaleFactor = [screen backingScaleFactor];
+    m_scaleFactor = m_highDpi ? [screen backingScaleFactor] : 1.0;
 
     // Send a resize event if the scaling factor changed
     if ((m_scaleFactor != oldScaleFactor) && (m_requester != 0)) {
--- a/src/SFML/Window/OSX/SFWindowController.h
+++ b/src/SFML/Window/OSX/SFWindowController.h
@@ -62,6 +62,7 @@ namespace sf {
     sf::priv::WindowImplCocoa*  m_requester;        ///< Requester
     BOOL                        m_fullscreen;       ///< Indicate whether the window is fullscreen or not
     BOOL                        m_restoreResize;    ///< See note above
+    BOOL                        m_highDpi;          ///< Support high-DPI rendering or not
 }
 
 ////////////////////////////////////////////////////////////
--- a/src/SFML/Window/OSX/SFWindowController.mm
+++ b/src/SFML/Window/OSX/SFWindowController.mm
@@ -99,6 +99,7 @@
         m_requester = 0;
         m_fullscreen = NO; // assuming this is the case... too hard to handle anyway.
         m_restoreResize = NO;
+        m_highDpi = NO;
 
         // Retain the window for our own use.
         m_window = [window retain];
@@ -111,7 +112,8 @@
 
         // Create the view.
         m_oglView = [[SFOpenGLView alloc] initWithFrame:[[m_window contentView] frame]
-                                             fullscreen:NO];
+                                             fullscreen:NO
+                                                highDpi:NO];
 
         if (m_oglView == nil)
         {
@@ -153,6 +155,7 @@
         m_requester = 0;
         m_fullscreen = (style & sf::Style::Fullscreen);
         m_restoreResize = NO;
+        m_highDpi = NO;
 
         if (m_fullscreen)
             [self setupFullscreenViewWithMode:mode];
@@ -218,7 +221,8 @@
     NSRect oglRect = NSMakeRect(x, y, width, height);
 
     m_oglView = [[SFOpenGLView alloc] initWithFrame:oglRect
-                                         fullscreen:YES];
+                                         fullscreen:YES
+                                            highDpi:m_highDpi];
 
     if (m_oglView == nil)
     {
@@ -277,7 +281,8 @@
 
     // Create the view.
     m_oglView = [[SFOpenGLView alloc] initWithFrame:[[m_window contentView] frame]
-                                         fullscreen:NO];
+                                         fullscreen:NO
+                                            highDpi:m_highDpi];
 
     if (m_oglView == nil)
     {
--- a/src/SFML/Window/OSX/WindowImplCocoa.mm
+++ b/src/SFML/Window/OSX/WindowImplCocoa.mm
@@ -130,9 +130,6 @@ m_showCursor(true)
     // Transform the app process.
     setUpProcess();
 
-    // Use backing size
-    scaleInWidthHeight(mode, nil);
-
     m_delegate = [[SFWindowController alloc] initWithMode:mode andStyle:style];
     [m_delegate changeTitle:sfStringToNSString(title)];
     [m_delegate setRequesterTo:this];
@@ -544,4 +541,3 @@ bool WindowImplCocoa::hasFocus() const
 } // namespace priv
 
 } // namespace sf
-


eXpl0it3r:
Maybe taking one step back again, I think there are essentially two issues here:

1) The drawing canvas/buffer needs to be able to differentiate/scale between the full DPI (high DPI) and the configured DPI scaling
2) The window styling needs to either handle high DPI and scaled DPI, but not just by rescaling a low-res version

I'm not as deep into macOS nor do I have a retina display to fully understand the defaults and possible options.
Do you think my assumption is correct, when I believe that if you have enabled DPI scaling, you'd expect *every* application to uphold that scaling? Or at least for the window styling?
With that in mind, I'd say that we shouldn't provide a window style for high DPI.

For the buffer/canvas itself, the idea was to provide an API which gives not just access to the window size, but also the non-scaled canvas size. That way you should be able to calculate the wanted difference yourself and if wanted render at "high DPI" while the window style itself is at scaled DPI.

From my limited understanding, I'd suggest the following changes:
- Always respect DPI scaling for the window styling
- As long as we don't have a way to provide the non-scaled DPI resolution, I'd force the canvas size to match the DPI scaling

Do you think this makes sense?

jqdg:
Thanks for the feedback!

> 1. The drawing canvas/buffer needs to be able to differentiate/scale between the full DPI (high DPI) and the configured DPI scaling
> 2. The window styling needs to either handle high DPI and scaled DPI, but not just by rescaling a low-res version
> 
> I'm not as deep into macOS nor do I have a retina display to fully understand the defaults and possible options.
> Do you think my assumption is correct, when I believe that if you have enabled DPI scaling, you'd expect _every_ application to uphold that scaling? Or at least for the window styling?

Agreed. In particular, every Mac application should display high-resolution window decorations and fonts on a retina display. GUI applications with native controls should display high-resolution UI elements. Some legacy application (i.e. those built before high DPI support was widespread) do not do this, and instead have pixelated/scaled low-res UI elements (like in some of the screenshots I posted). These applications generally stick out like a sore thumb (Mac users tend to notice details like this), and this situation should be avoided if possible. That is the reason that the current solution/workaround of setting `NSHighResolutionCapable` to false is not really satisfactory.

The problem with the *current* SFML functionality is that if you use high-res window decorations, you get a tiny window size. If you want a full-sized window (without manually scaling everything in your rendering code, keeping in mind that the display scale is not accessible with an SFML API), then the window decorations and fonts will be low-res.

What this PR does is make the default mode for windows have high-resolution decorations and fonts, but keep the scaled window size, and simply scale the contents of the OpenGL view.

> * Always respect DPI scaling for the window styling
> * As long as we don't have a way to provide the non-scaled DPI resolution, I'd force the canvas size to match the DPI scaling
> 
> Do you think this makes sense?

If I understand you correctly, that is what the new "default mode" of this PR does.

Now, one potential issue with this approach is that graphics within the OpenGL view will not be rendered at high-resolution. That is the reason I introduced the `HighDPI` window style: if the application developer wants to implement themselves high-resolution rendering (for example, getting the display scale factor from platform-specific APIs), then they can choose to do so by enabling this window style. In other words, the extra work of supporting high DPI rendering becomes opt-in.

If you believe that this window style adds extra complexity for little benefit, then I am completely fine getting rid of it. The consequence of that would mean that rendering high DPI graphics within the OpenGL view would not be possible.

yuupsup:
I currently don't know anything about high DPI, but I'm willing to do research into whatever is needed. This issue has been the reason I have not been able to dive into SFML on my Macbook Pro. 😭

jqdg:
> I currently don't know anything about high DPI, but I'm willing to do research into whatever is needed. This issue has been the reason I have not been able to dive into SFML on my Macbook Pro. 😭

Hi @yuupsup, would you mind giving this PR a try and let me know if it works according to your needs? Thanks!

yuupsup:
@jqdg Sorry, I never got a notification that you responded. PR means "pull request", correct? Do you mind working with me on how I can try the request? How do I pull the changes you've made?
Also, I assume I'll need to build the code again, correct?

Bromeon:
@yuupsup Yes, "PR" means "pull request".

You can checkout this request locally by doing this:
```bash
# Clone SFML (you can skip if you have done this already)
git clone https://github.com/SFML/SFML.git
cd SFML

# Check out branch of this PR locally
git fetch origin pull/1817/head:high-dpi
git checkout high-dpi
```

In the future, you can **overwrite** your local branch with newest PR changes like this:
```bash
git fetch origin pull/1817/head:high-dpi -uf
```

Alternatively, directly track the remote branch in the original repo: https://github.com/jqdg/SFML/tree/osx-high-dpi.

yuupsup:
@Bromeon Thanks I was able to get it. :)

@jqdg I'm amazed haha, it worked. Using the default (the only one I have currently tested) style is displaying the correct window and image size.

jqdg:
> @Bromeon Thanks I was able to get it. :)
> 
> @jqdg I'm amazed haha, it worked. Using the default (the only one I have currently tested) style is displaying the correct window and image size.

Thanks for checking! Glad it looks like it's working for you. I suspect that the default style should work well in almost all cases (I only introduced the `HighDPI` style so that people who want to manually implement high-DPI rendering themselves will still have the option to opt-in should they want to).

> **Edit**
> I hope this message will be seen, but I'm looking at the list of expected checks in order for the pull request to get accepted.. these are just all the other machines (and compilers) that need to be tested, correct?

There are some more CI checks that are needed. I think the maintainers would also need to decide that they want to adopt this design. I would definitely like to make whatever modifications are needed to get this PR accepted. It would be great to make working with high DPI displays easier on Mac OS.

eXpl0it3r:
> Now, one potential issue with this approach is that graphics within the OpenGL view will not be rendered at high-resolution. That is the reason I introduced the HighDPI window style: if the application developer wants to implement themselves high-resolution rendering (for example, getting the display scale factor from platform-specific APIs), then they can choose to do so by enabling this window style. In other words, the extra work of supporting high DPI rendering becomes opt-in.

But isn't this also possible by adjusting the view after the fact? Say you have a 150% DPI scaling, if you as a library user *somehow* retrieve this scaling information, you could just zoom out the view by a factor of `1/1.5`, no?

I still believe the HighDPI as a style is not the way to go, but stick to defaults and later down the line provide a function that returns the "unscaled" canvas size, so the user can get that factor on their own.

jqdg:
> But isn't this also possible by adjusting the view after the fact? Say you have a 150% DPI scaling, if you as a library user _somehow_ retrieve this scaling information, you could just zoom out the view by a factor of `1/1.5`, no?

No, I don't think so. If the high-DPI mode is not enabled, the entire OpenGL view will be rendered at 2x. If you use a view to zoom out, you can make sure that everything is rendered at the same apparent size as it would with high-DPI enabled, but it will still be pixelated. Compare the following:

| `HighDPI` style |
| --- |
| high_dpi|

| Default style with scaling |
| --- |
| scaled |

> I still believe the HighDPI as a style is not the way to go, but stick to DPS scaling as default and later down the line provide a function that returns the "unscaled" canvas size, so the user can get that factor on their own.

I would also be happy to just remove the `HighDPI` style from this PR, in which case there would be no (easy) option for users to render with high DPI.

I think the default behavior covers 90+% of the use-cases, I imagine most users of SFML are not handling high-DPI rendering themselves. In which case, the new default behavior introduced in this PR is a big improvement compared with the status quo, which is a big stumbling block for those wanting to use SFML on Macs with retina displays.

jqdg:
> I still believe the HighDPI as a style is not the way to go, but stick to DPS scaling as default and later down the line provide a function that returns the "unscaled" canvas size, so the user can get that factor on their own.

@eXpl0it3r, what do you think about the option to just remove the `HighDPI` style and always render scaled? Or is there some other API that would be better to expose this option in your opinion?

Perhaps as a first step, we could do the first (render everything scaled), and then separately (in a different PR) consider different options for public APIs to enable high DPI rendering?

eXpl0it3r:
I think it's the way to go, unfortunately, I'm somehow having a hard time to fully understand what you answered to my question (it's me not you 😅).

I mean the "Default style with scaling" looks kind of bad, but it's also not really representative of how it actually looks, since nobody would scale that much, correct?

eXpl0it3r:
Okay, then my assumptions were correct so far.

But then doesn't my point still hold true, that you could expand/zoom out in the scaled view and render things at a higher resolution?
For example if I have an texture that is draw scaled as 100x100px and now I zoom out the view by a factor of `0.5f` and then render a higher resolution texture of 200x200px. Will the end result not be of higher quality? Or does the scaled version prevent any usage of "more pixels"?

jqdg:
> But then doesn't my point still hold true, that you could expand/zoom out in the scaled view and render things at a higher resolution? For example if I have an texture that is draw scaled as 100x100px and now I zoom out the view by a factor of `0.5f` and then render a higher resolution texture of 200x200px, which visually will be shown as the previous 100x100px. Will the end result not be of higher quality? Or does the scaled version prevent any usage of "more pixels"?

Right, the scaled version prevents the use of more pixels. In the scaled version, every pixel rendered to the OpenGL view will be blown up to become a 2x2 block of pixels.

That is exactly what is happening in the two screenshots that I posted previously. The SFML logo (included in the repo at `examples/tennis/resources/sfml_logo.png`) is 373×113 pixels. On a 2x retina display, displaying this image without any scaling, it will appear to be the same size as a 186x56 pixel image on a non-high DPI display (just with higher pixel density). In high DPI mode, if I draw this image without scaling, I indeed obtain an image that is 186x56 *points* (i.e. half size when compared with non-high DPI), where each point consists of 2x2 pixels. This is expected.  The result is the "high DPI" screenshot from the previous post.

Now, I try to replicate this behavior using the scaling, without high DPI mode. I use the new default window style from this PR, and create a half-sized window, and render the texture and 0.5 scale. What happens is that the texture is down-sampled to 186x56 pixels, and then drawn to the OpenGL view, which (as I mentioned above) will draw each pixel of the down-sampled texture as a block of 2x2 pixels. So I get an image that's the same size as the high DPI case, but pixelated.

There is no way (that I am aware of) to obtain the same result as the "high DPI" screenshot using only the scaled mode.

eXpl0it3r:
Okay, thank you very much to have all this patience with my stupid questions, just wanted to make sure I understand it fully.

I am still for removing the HighDPI style and use the default DPI scaling. As a user, you enabled DPI scaling, so you should expect to have your applications scaled.

Have support of High DPI will look like API wise, should be discussed at a later stage.

Any objections to this, @SFML/sfml team?

jqdg:
I got rid of the proposed `HighDPI` window style, so now the window will always render scaled. Let me know what you think.

eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:
FYI: I've squashed the commits, so it looks a bit nicer in the git history.

Thanks a lot for this suggestion! I hope, we'll get some more positive feedback as it makes it to various macOS users! 🙂

eXpl0it3r:
First success story: https://github.com/TankOs/SFGUI/issues/88 🙂

Btw. do you want to continue the conversation on how to allow the user to decide whether they want to use high-dpi on the [forum thread](https://en.sfml-dev.org/forums/index.php?topic=17092.0)?

williamhCode:
I have a proposal, is it possible to make the HighDPI flag make everything render in high resolution, but keep the window size and coordinates the same? This is the default behavior in most other libraries such as SDL, and I'm confused as to why nobody pointed this out.

yuupsup:
> forum thread

I've been looking for this issue haha. I'm glad to see everything turned out well.
Sorry to revive this thread, but is the request @williamhCode mentioned possible?

Thanks again @jqdg for initially creating this pull request.

jkeller51:
Hi, I updated my fork from SFML 2.5.1 to 2.6.2 and the dpi scaling on Mac is a regression for me. I already had crisp retina graphics on Mac in my application, and now it's forced to render at a lower resolution in SFML 2.6.x.

By changing SFWindowController `m_highDpi = NO` to `m_highDpi = YES` (2 instances) and SFOpenGLView `highDpi:NO` to `highDpi:YES` (2 instances), I was able to restore the previous behavior on my fork.

MacOS is already adjusting window/view sizes for the screen dpi scale (known as backing factor in Mac world). SFML ends up multiplying and dividing these numbers so when you ask window->getSize(), it gives you the number of raw pixels in the window (i.e. SFOpenGLView).

Probably the expected behavior in the future will be for window sizes to be independent of dpi scaling as @williamhCode asked. But for now, you can get high DPI scaling with the above tweak, and the below snippet to run whenever the window size changes:

```c++
    // sf::RenderWindow m_window
    // float m_width, m_height is the expected window size in "SFML coordinates"

    sf::Vector2u windowSize = m_window->getSize();

    dpiScale = windowSize.x / m_width;

    m_view.setSize(m_width, m_height);  /* size of the view in SFML coordinates */
    m_view.setCenter(m_width / 2.0f, m_height / 2.0f);
    m_Window->setView(m_view);
```

You can then use `dpiScale` to increase the internal resolution of all your RenderTextures etc, to keep them 1:1 with the screen DPI.

PR #1821, opened 2021-08-30
Silence deprecation warnings on Mac OS X

jqdg:
Building SFML on Mac OS X results in ~170 compiler warnings (all of which are deprecation warnings), mostly because OpenAL and OpenGL are deprecated on Mac.

Given that SFML's use of OpenAL and OpenGL is pretty hard-coded (at least for the foreseeable future) it probably makes sense to silence these warnings.

### DIFF
--- a/examples/cocoa/CMakeLists.txt
+++ b/examples/cocoa/CMakeLists.txt
@@ -55,6 +55,7 @@ sfml_add_example(cocoa
                  SOURCES ${SRC}
                  BUNDLE_RESOURCES ${RESOURCES}
                  DEPENDS sfml-system sfml-window sfml-graphics)
+target_compile_options(cocoa PRIVATE -Wno-deprecated-declarations)
 set_target_properties(cocoa PROPERTIES
                       MACOSX_BUNDLE TRUE
                       MACOSX_BUNDLE_INFO_PLIST ${SRCROOT}/resources/Cocoa-Info.plist)
--- a/src/SFML/Audio/CMakeLists.txt
+++ b/src/SFML/Audio/CMakeLists.txt
@@ -85,6 +85,10 @@ if(SFML_OS_ANDROID)
     target_link_libraries(sfml-audio PRIVATE android OpenSLES)
 endif()
 
+if(SFML_OS_MACOSX)
+    target_compile_options(sfml-audio PRIVATE -Wno-deprecated-declarations)
+endif()
+
 target_link_libraries(sfml-audio
                       PUBLIC sfml-system
                       PRIVATE VORBIS FLAC)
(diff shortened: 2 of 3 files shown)


eXpl0it3r:
Is there a way to be more specific about which warnings are disabled? Because using `Wno-deprecated-declarations` would also silence deprecation warnings about `NSMouseMoved` / `NSLeftMouseDragged`/ `NSResizableWindowMask` / etc.

jqdg:
I don't believe there is any way to silence only specific deprecation warnings (e.g. only those about OpenGL or OpenAL).

It is possible to disable the warnings on a per-file or per-section basis using pragmas.

On the other hand, it might make sense to ignore even those other deprecation warnings for compatibility with older versions of the Mac OS SDK. For example, `NSEventTypeLeftMouseDragged` is only available on 10.12+, so if we want to compile with older SDKs, it makes sense to use `NSLeftMouseDragged` even though it is deprecated.

jqdg:
Any further thoughts on this? I think that it makes sense to choose one of two possible options:

1. Fix the deprecation warnings that we can (may break compatibility with older OS versions, impossible to do for OpenGL and OpenAL).
2. Suppress the deprecation warnings.

I don't think it makes sense to keep around deprecation warnings which are being deliberately ignored. We _know_ that OpenGL and OpenAL are deprecated on Mac, being reminded of this many times every recompile isn't useful.

In my mind, 2 is the logical option, since 1 would break backwards compatibility, with no clear benefit.

HeavyRain266:
I think, only solution would be full new macOS specific backend which uses Metal and their SDKs for sound

eXpl0it3r:
Having enabled all sorts of warnings and treat warnings as errors on the `2.6.x` as well as `master` branch, is this still needed?

jqdg:
> [Having enabled](https://github.com/SFML/SFML/pull/1846) all sorts of warnings and treat warnings as errors on the `2.6.x` as well as `master` branch, is this still needed?

Compiling the master branch on Mac OS, I still see > 140 compiler warnings.

reconndev:
That would silence all deprecated declarations for `sfml-audio` and `sfml-window`.

A better approach would be to build macOS system dependencies (*.mm and *.hpp) as a separate static library and include headers with `-isystem`.

This is CMake equivalent of `-isystem`:
```cmake
target_include_directories(cocoa SYSTEM PRIVATE ${APPLE_DEPENDENCIES})
```

That requires more work but wouldn't suppress all deprecation declaration warnings.

ChrisThrasher:
I haven't seen these deprecation warnings in a while. I think this got fixed in c824d3d57a5ef905a4453c9c2fde892c2aa8737b.

eXpl0it3r:
I believe we have resolved this issue already with #1908  
Our macOS builds (e.g. [here](https://github.com/SFML/SFML/actions/runs/3153410499/jobs/5129853058)) don't seem to report such warnings and @ChrisThrasher also confirms to not see them (anymore).

As such I'll close this PR and mark it as superseded, if you still see something that needs to be fixed, feel free to update the PR and reopen it. 🙂

Issue #1826, opened 2021-10-07
Incorrect glyph position with text outline


cschreib:
## Subject of the issue

I am using SFML as one of multiple back-ends for a GUI library. I have experienced some issues with the outline glyph positions returned by `sf::Font`, see the `:` and `+` glyphs on the screenshot below (from current master):
![Screenshot from 2021-10-07 07-53-15](https://user-images.githubusercontent.com/2236577/136334927-328e2af5-fa3e-4571-b56d-c366caad79ae.png)

## Your environment

* Your OS / distro / window manager used: Linux Mint 20.2 Cinnamon
* Your version of SFML (2.5.0, git master, etc): master
* Your compiler and compiler version used: GCC 9.3 and Clang 10 or 11
* Special compiler flags used: none

## Steps to reproduce

Tell us how to reproduce this issue. Please provide a [minimal, complete and verifiable example](https://stackoverflow.com/help/mcve), you can use the follow template as a start:

 - Font I use for testing: [noto.ttf.zip](https://github.com/SFML/SFML/files/7299511/noto.ttf.zip)
 - Code:
```cpp
#include <SFML/Graphics.hpp>
#include <cmath>

int main() {
    sf::RenderWindow window(sf::VideoMode(768, 512), "Minimal, complete and verifiable example");
    window.setFramerateLimit(60);

    sf::Font fnt;
    fnt.loadFromFile("noto.ttf");

    while (window.isOpen())
    {
        sf::Event event;
        while (window.pollEvent(event))
        {
            if (event.type == sf::Event::Closed)
                window.close();
        }

        window.clear(sf::Color(128, 128, 128));

        float y = 0.0;
        for (std::size_t i = 5; i < 20; ++i) {
            sf::Text txt("Language: C++", fnt, i);
            txt.setFillColor(sf::Color::White);
            txt.setOutlineColor(sf::Color::Black);
            txt.setOutlineThickness(2.0f);
            txt.setPosition(0.0, std::round(y));
            window.draw(txt);

            y += 1.5f*i;
        }

        y = 0.0;
        for (std::size_t i = 5; i < 20; ++i) {
            sf::Text txt("Language: C++", fnt, i);
            txt.setFillColor(sf::Color::White);
            txt.setStyle(sf::Text::Bold);
            txt.setOutlineColor(sf::Color::Black);
            txt.setOutlineThickness(2.0f);
            txt.setPosition(256.0, std::round(y));
            window.draw(txt);

            y += 1.5f*i;
[...]

PR #1827, opened 2021-10-07
Fix incorrect glyph rect for text outline

cschreib:

----

## Description

This PR fixes the glyph rect calculation in `sf::Font` to use the most accurate information from the FreeType rendered bitmap data, rather than the glyph metrics. The returned rect is guaranteed to contain the full rendered glyph with no stretching. It also simplifies the positioning in `sf::Text`, since the offsetting from the outline is now correctly taken into account in the returned rect from `sf::Font`.

As a side effect, it seems to have "fixed" an unintended stretching for bold fonts:
fix

This PR is related to the issue https://github.com/SFML/SFML/issues/1826

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

 - Font I use for testing: [noto.ttf.zip](https://github.com/SFML/SFML/files/7299511/noto.ttf.zip)
 - Code:
```cpp
#include <SFML/Graphics.hpp>
#include <cmath>

int main() {
    sf::RenderWindow window(sf::VideoMode(768, 512), "Minimal, complete and verifiable example");
    window.setFramerateLimit(60);

    sf::Font fnt;
    fnt.loadFromFile("noto.ttf");

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
        }

        window.clear(sf::Color(128, 128, 128));

        float y = 0.0;
        for (std::size_t i = 5; i < 20; ++i) {
            sf::Text txt("Language: C++", fnt, i);
            txt.setFillColor(sf::Color::White);
            txt.setOutlineColor(sf::Color::Black);
            txt.setOutlineThickness(2.0f);
            txt.setPosition(0.0, std::round(y));
            window.draw(txt);

            y += 1.5f*i;
        }

        y = 0.0;
        for (std::size_t i = 5; i < 20; ++i) {
            sf::Text txt("Language: C++", fnt, i);
            txt.setFillColor(sf::Color::White);
            txt.setStyle(sf::Text::Bold);
            txt.setOutlineColor(sf::Color::Black);
            txt.setOutlineThickness(2.0f);
            txt.setPosition(256.0, std::round(y));
            window.draw(txt);

            y += 1.5f*i;
        }

        y = 0.0;
        for (std::size_t i = 5; i < 20; ++i) {
            sf::Text txt("Language: C++", fnt, i);
            txt.setFillColor(sf::Color::White);
            txt.setStyle(sf::Text::Italic);
            txt.setOutlineColor(sf::Color::Black);
            txt.setOutlineThickness(2.0f);
            txt.setPosition(512.0, std::round(y));
            window.draw(txt);

            y += 1.5f*i;
        }

        window.display();
    }
}
```

### DIFF
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -631,7 +631,7 @@ Glyph Font::loadGlyph(Uint32 codePoint, unsigned int characterSize, bool bold, f
     }
 
     // Compute the glyph's advance offset
-    glyph.advance = std::round(static_cast<float>(bitmapGlyph->root.advance.x) / static_cast<float>(1 << 16));
+    glyph.advance = static_cast<float>(bitmapGlyph->root.advance.x >> 16);
     if (bold)
         glyph.advance += static_cast<float>(weight) / static_cast<float>(1 << 6);
 


eXpl0it3r:
Summoning our fonts expert, @oomek 😄 
Can you take a peek at this?

oomek:
Sure, on it.

oomek:
There was indeed funny business going on with outline and style flags applied. @cschreib commit seems to fix the rendering with every font, style flag and outline combination I tested so far. Forced autohint advance compensation is still working as before. Thanks for spotting it and fixing.
Just one thing would change is the line:
https://github.com/SFML/SFML/blob/40a0584636dab89f3fa1747e63f26fa895036ce1/src/SFML/Graphics/Font.cpp#L634
From what I see the `root.advance.x` is now `long` and multilier of 65536 so you could replace it with just:
```cpp
glyph.advance = static_cast<float>(bitmapGlyph->root.advance.x >> 16);
```

cschreib:
> ust one thing I would change is the line:
> 
> https://github.com/SFML/SFML/blob/40a0584636dab89f3fa1747e63f26fa895036ce1/src/SFML/Graphics/Font.cpp#L634
> 
> From what I see the `root.advance.x` is now of `long` type and a multiplier of 65536 so you could replace it with just:
> 
> ```c++
> glyph.advance = static_cast<float>(bitmapGlyph->root.advance.x >> 16);
> ```
> 
> Update: a little explanation why we can just `>> 16` Sine we use forced hinting flag the advance will always be integer and since it's in 16.16 fixed point we can just skip the fractional part. The actual fractional advance is handled by bearings.

Thanks for the explanation, that makes sense. I work in parallel on my own FreeType+OpenGL renderer, which does not use hinting, and it needed `round()` to get the right advance. Now I know why :) I thought it was a general thing, that the advance had to be rounded, so I ported it to SFML just in case. But you are right, there is absolutely no difference in output from SFML whether I use  `round()`  or `floor()`.

I have pushed a commit to apply this suggestion.

> I've just noticed some new odd artifact when outline is large enough to make internal circles collapse and turn inside-out

It sounds like it may be a fundamental problem with the Freetype stroker:
https://lists.nongnu.org/archive/html/freetype/2017-05/msg00017.html

When the outline thickness becomes large enough compared to the font size, it is probably best to simply convolve the non-outlined glyph with a circular mask (of radius = outline thickness). This could quickly become a can of worms, so I'm not sure this should be done in this PR.

eXpl0it3r:
If the two issues aren't directly related, then I too suggest to separate these. If it's really something that SFML can fix/improve then feel free to create another issue, so it can be tracked.

oomek:
> If the two issues aren't directly related, then I too suggest to separate these. If it's really something that SFML can fix/improve then feel free to create another issue, so it can be tracked.

I agree, the guy from freetype repo just confirmed that limitation of the stroker algorithm, I'll keep researching the subject.

eXpl0it3r:
Thanks a lot for finding this, reporting and fixing it, highly appreciated! 🙂

cschreib:
Yep my bad, I did not look closely enough at the code. I thought `addLine` meant "add a line of text" rather than "add a straight line for underline or strike-through", and blindly removed anything related to outline. I'll open a PR now with the fix for the fix...

Issue #1679, opened 2020-07-01
Link error when adding SFML as a subdirectory and with a static runtime


baugid:
## Subject of the issue

When adding SFML as a subdirectory to a cmake project and compiling it with MSVC using the static runtime a link time error occurs (`LNK2038 mismatch detected for 'RuntimeLibrary': value 'MD_DynamicRelease' doesn't match value 'MT_StaticRelease' in main.obj`)

## Environment

* Windows 10 64bit version 1909
* current SFML git master
* current version of MSVC/Visual studio 2019
* Tried to use /MT compiler flag

## Steps to reproduce
Just try to build the following code with MSVC and clone SFML to a directory called SFML

main.cpp:
```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode(1280, 720), "Example code");
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        sf::Event event;
        while (window.pollEvent(event))
        {
            if (event.type == sf::Event::Closed)
                window.close();
        }

        window.clear();
        window.display();
    }
}
```

CMakeLists.txt:
```CMake
cmake_minimum_required(VERSION 3.15)

project(Project VERSION 1.0 LANGUAGES CXX)

set(BUILD_SHARED_LIBS OFF)

add_executable(out main.cpp)

set_property(TARGET out PROPERTY
                MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")

set(SFML_USE_STATIC_STD_LIBS ON CACHE BOOL "Use a static c++ runtime" FORCE)
add_subdirectory(SFML)
target_link_libraries(out PRIVATE sfml-graphics)
```

## Expected behavior

The project should build and the configuration of the sfml targets should show the /MT flag in Visual Studio ( in Release mode).

## Actual behavior

The project doesn't link (with the above error message) and visual studio shows that the /MD flag is selected (in Release mode)

## Possible Fix

I fixed the problem by setting the `MSVC_RUNTIME_LIBRARY` property for all SFML targets to `"MultiThreaded$<$<CONFIG:Debug>:Debug>"` in my CMakeLists.txt.  This change probably needs to be copied to the appropriate files inside the SFML Project.

JonnyPtn:
[...]

PR #1834, opened 2021-10-18
Use modern CMake property to handle static runtime linking with MSVC

eXpl0it3r:
## Description

In some cases /MD* flags aren't part of the CXX CMake flags and thus
won't be replaced with /MT, for example when adding SFML as
sub-directory to another CMake project.

This PR is related to the issue #1679

## Tasks

* [x] Tested on Windows

## How to test this PR?

Source Example

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    auto window = sf::RenderWindow{ sf::VideoMode{ { 1280, 720} }, "Example code" };
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        {
            {
                window.close();
            }
        }

        window.clear();
        window.display();
    }
}
```

CMake Example

```cmake
cmake_minimum_required(VERSION 3.15)

project(Project VERSION 1.0 LANGUAGES CXX)

set(BUILD_SHARED_LIBS OFF)

add_executable(out main.cpp)

set_property(TARGET out PROPERTY
                MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")

set(SFML_USE_STATIC_STD_LIBS ON CACHE BOOL "Use a static c++ runtime" FORCE)
add_subdirectory(SFML)
target_link_libraries(out PRIVATE sfml-graphics)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,17 +207,6 @@ if(SFML_OS_WINDOWS)
     if(BUILD_SHARED_LIBS AND SFML_USE_STATIC_STD_LIBS)
         message(FATAL_ERROR "BUILD_SHARED_LIBS and SFML_USE_STATIC_STD_LIBS cannot be used together")
     endif()
-
-    # for VC++, we can apply it globally by modifying the compiler flags
-    if((SFML_COMPILER_MSVC OR (SFML_COMPILER_CLANG AND NOT MINGW)) AND SFML_USE_STATIC_STD_LIBS)
-        foreach(flag
-                CMAKE_CXX_FLAGS CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE
-                CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO)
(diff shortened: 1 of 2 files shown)


Bromeon:
> Yeah, would be great if we could use that, but not sure we can really upgrade CMake's minimal version to 3.15.

I think it's OK -- the alternative is that we have another point of technical debt in the configuration, or a PR which remains in undecided state.

I'm also OK with leaving the current implementation, but let's then add a `TODO: ` comment.

ChrisThrasher:
> Yeah, would be great if we could use that, but not sure we can really upgrade CMake's minimal version to 3.15.
> 
> https://repology.org/project/cmake/versions

We're using 3.16 now. Does that impact this PR?

eXpl0it3r:
Yes, if you wait long enough you can suddenly start using new shiny things. 😄 

I've reimplemented the original PR and use `MSVC_RUNTIME_LIBRARY` now.

Bambo-Borris:
Tested on Windows 11 64-bit with Visual Studio 2022 and works perfectly for me too!

JonnyPtn (approved):

Bambo-Borris (approved):

PR #1845, opened 2021-11-19
Simplify management of `ActivityStates`

SuperV1234:

* [x] Does the code follow the SFML [Code Style Guide](https://www.sfml-dev.org/style.php)?[^1]
* [x] Have you provided some example/test code for your changes?[^2]
* [x] If you have additional steps which need to be performed list them as tasks![^3]

[^1]: `ActivityStates` is used a dynamically-allocated object via `new`, but the changed code doesn't care about or manage its lifetime, so it is appropriate to use it as a reference there. The code that manages the lifetime of `ActivityStates` objects still uses raw pointers as specified by the coding guidelines.

[^2]: I have tested the existing test suite and compiled for Android.

[^3]: No additional tests required.

----

## Description

This is a refactoring PR that simplifies the way `ActivityStates` is managed, making it clear that the `NULL` state is almost always *not* a valid domain value. It was prompted by some improper use of SFML for an Android project, which resulted in some puzzling crashes due to `*states` being accessed without ever asserting `states != NULL`. 

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [x] Tested on Android

## How to test this PR?

This is a refactoring PR. The behavior of the code should not have changed. Any unlikely regression will be likely found by CI or testing suites.

### DIFF
--- a/src/SFML/System/Android/Activity.cpp
+++ b/src/SFML/System/Android/Activity.cpp
@@ -29,6 +29,7 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/System/Android/Activity.hpp>
 #include <android/log.h>
+#include <cassert>
 
 #define LOGE(...) ((void)__android_log_print(ANDROID_LOG_INFO, "sfml-error", __VA_ARGS__))
 
@@ -56,14 +57,24 @@ namespace sf
 {
 namespace priv
 {
-ActivityStates* getActivity(ActivityStates* initializedStates, bool reset)
+
+ActivityStates*& getActivityStatesPtr()
 {
     static ActivityStates* states = NULL;
-
-    if (!states || reset)
-        states = initializedStates;
-
     return states;
 }
+
+void resetActivity(ActivityStates* initializedStates)
+{
+    getActivityStatesPtr() = initializedStates;
+}
+
+ActivityStates& getActivity()
+{
+    ActivityStates* const states = getActivityStatesPtr();
+    assert(states != NULL);
+    return *states;
+}
+
 }
 }
--- a/src/SFML/System/Android/Activity.hpp
+++ b/src/SFML/System/Android/Activity.hpp
@@ -91,7 +91,9 @@ struct ActivityStates
     LogcatStream logcat;
 };
 
-SFML_SYSTEM_API ActivityStates* getActivity(ActivityStates* initializedStates=NULL, bool reset=false);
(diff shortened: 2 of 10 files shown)


src/SFML/System/Android/Activity.hpp
```diff
@@ -91,7 +91,9 @@ struct ActivityStates
     LogcatStream logcat;
 };
 
-SFML_SYSTEM_API ActivityStates* getActivity(ActivityStates* initializedStates=NULL, bool reset=false);
+SFML_SYSTEM_API ActivityStates*& getActivityStatesPtr();
+SFML_SYSTEM_API void resetActivity(ActivityStates* initializedStates);
+SFML_SYSTEM_API ActivityStates& getActivity();
```
SuperV1234, line 96: This is the most significant change, I split a single monolithic function into two smaller ones with a more well-defined and specific responsibility. `getActivityStatesPtr` could be hidden in the `.cpp` but I wasn't sure whether providing `static` linkage for the function itself would screw up things.

eXpl0it3r:

eXpl0it3r:

src/SFML/System/Android/Activity.cpp
```diff
@@ -56,14 +57,24 @@ namespace sf
...
-        states = initializedStates;
+void resetActivity(ActivityStates* initializedStates)
+{
+    getActivityStatesPtr() = initializedStates;
+}
 
-    return states;
+ActivityStates& getActivity()
+{
+    ActivityStates*& states = getActivityStatesPtr();
```
eXpl0it3r, line 74: Do you gain anything with a reference to a pointer?

SuperV1234: Not on the caller side, I will tweak this one.

SuperV1234:

SuperV1234:
It doesn't listen to me :(

eXpl0it3r:

eXpl0it3r:
> It doesn't listen to me :( 

That's kind of the point of not running it automatically, to prevent people executing random code on our machines 😄

eXpl0it3r (approved):

eXpl0it3r:

eXpl0it3r:
Thanks for this refactoring! 🙂

PR #1846, opened 2021-11-19
Fix several unused warnings (continued)

SuperV1234:
Resume work on #1791.

PR #1847, opened 2021-11-19
Proof of concept conversion to C++17 

vittorioromeo:
This PR, [motivated by this thread](https://en.sfml-dev.org/forums/index.php?topic=28301.0), is a proof of concept conversion of SFML to C++17. It removes the following classes:

- `sf::Thread` (replaced with `std::thread`).
- `sf::Mutex` (replaced with `std::recursive_mutex`).
- `sf::Lock` (replaced with `std::scoped_lock`).
- `sf::ThreadLocal` (replaced with `thread_local`).
- `sf::NonCopyable` (replaced with `delete`).

Other changes:

- Replace `NULL` with `nullptr`.
- Add `CXX_STANDARD 17` requirement in `CMakeLists.txt`.
- Replace `SFML_DEPRECATED` with `[[deprecated]]`.
- Replace `// fallthrough` with `[[fallthrough]]`.
- Turn `sf::Color`, `sf::Vector2`, `sf::Vector3`, `sf::Rect`, and `sf::Vertex` into `constexpr`-friendly and fully-`inline` classes.
- Remove all interfaces taking individual components (e.g. `setPosition(float x, float y)`), in favor of interfaces taking vectors (e.g. `setPosition(const sf::Vector2f&)`).
- Remove all deprecated functions from `sf::Shader` and `sf::Text`.
- Changed some occurrences of `std::map` to `std::unordered_map` (e.g. in `sf::Font` and `sf::Shader`).
- Use zero-initialization via `{}` instead of `std::fill` calls for arrays  in some places.
- Use `using` instead to `typedef` in some places.
- ~~`sf::Shader::getUniformLocation` is now `public`.~~
- ~~`sf::Shader::setUniform` now takes an integer location instead of a `std::string`.~~

Other included PRs:

- Includes #1845.
- Includes #1846.
- Includes #1848.

Testing status:

- Tested on both Windows x64 and Linux x64 by running the test suite and recompiling [Open Hexagon](https://en.sfml-dev.org/forums/index.php?topic=28291.0) with this branch. Seems to work.

- Definitely not thoroughly tested.


include/SFML/Graphics/Color.hpp
```diff
@@ -59,36 +60,36 @@ class SFML_GRAPHICS_API Color
...
-    static const Color Transparent; //!< Transparent (black) predefined color
+    static Color Black;          //!< Black predefined color
+    static Color White;    //!< White predefined color
+    static Color Red;          //!< Red predefined color
+    static Color Green;        //!< Green predefined color
+    static Color Blue;         //!< Blue predefined color
+    static Color Yellow;     //!< Yellow predefined color
+    static Color Magenta;    //!< Magenta predefined color
+    static Color Cyan;       //!< Cyan predefined color
+    static Color Transparent; //!< Transparent (black) predefined color
```
kimci86, line 92: Shouldn't this be constexpr or at least const? (also the comments are not aligned)

SuperV1234: https://stackoverflow.com/questions/29432283/c-static-constexpr-field-with-incomplete-type/41570881

sam20908:
So, is this suppose to supersede other PRs?

SuperV1234:
> So, is this suppose to supersede other PRs?

I am not sure, I just got fed up with the lack of activity towards a Modern C++ update and kinda started doing my own thing. If the SFML maintainers would like to merge something like this I would be more than happy to split this PR into smaller ones, or just create a new branch starting from this one.

Otherwise I might maintain my own fork for a while, who knows.

eXpl0it3r:

Bromeon:
Thank you very much for this massive contribution! 

I still need to have a detailed view of the code.
Most of the high-level changes look reasonable so far, a few comments:

> Turn `sf::Color`, `sf::Vector2`, `sf::Vector3`, `sf::Rect`, and `sf::Vertex` into `constexpr`-friendly and fully-`inline` classes.

For the templates this should be OK, but not sure about `Color` and `Vertex`, see [my comments here](https://github.com/SFML/SFML/pull/1741#issuecomment-762838738). For colors I understand the motivation to declare compile-time constants. But probably no one needs `constexpr` in `sf::Vertex`, so it would be a trade-off of run vs. compile-time with yet unknown benefits, and similar thoughts would be applicable to many other SFML classes.

> Remove all interfaces taking individual components (e.g. `setPosition(float x, float y)`), in favor of interfaces taking vectors (e.g. `setPosition(const sf::Vector2f&)`).

We don't yet do it, but do you think it would make sense to pass classes like `sf::Vector2f` by value?

>     * Use zero-initialization via `{}` instead of `std::fill` calls for arrays  in some places.

Maybe we could even go for `std::array`. Doesn't have to be part of this PR, of course.

---

Since this contains many breaking changes, it needs to become part of SFML 3. So we can revive [the discussion](https://en.sfml-dev.org/forums/index.php?topic=21571.50) if SFML 3 should continue on `master` while we maintain a `sfml-2` branch, or vice versa.


Also relevant in this context: [C++17 Solution Design](https://github.com/SFML/SFML/wiki/SD:-CPP-17-Support)

include/SFML/Graphics/Color.hpp
kimci86: To make it clear for others, `constexpr` would require constants to be defined here, but the `Color` type is not complete at this point so it is not possible.
Using `const` is possible (as it was before) and constants would be defined later, when the `Color` type is complete.
A workaround to use `constexpr` is given in the next commit https://github.com/SFML/SFML/pull/1847/commits/77edd6718a6b90478b8d9340a1c5a04aa29e7a03. It uses an auxiliary struct `ColorInit` implicitly convertible to `Color`.

SuperV1234:
> Thank you very much for this massive contribution! It's great to see we're moving forward on this front 🙂

You're welcome! Please look at my thread on the forum, I really wanna get the ball rolling.
https://en.sfml-dev.org/forums/index.php?topic=28301.0

---

> For the templates this should be OK, but not sure about `Color` and `Vertex`, see [my comments here](https://github.com/SFML/SFML/pull/1741#issuecomment-762838738). For colors I understand the motivation to declare compile-time constants. But probably no one needs `constexpr` in `sf::Vertex`, so it would be a trade-off of run vs. compile-time with yet unknown benefits, and similar thoughts would be applicable to many other SFML classes.

For me, the argument is the complete opposite. `sf::Color` and `sf::Vertex` should be aggregates, which are implicitly `constexpr`.  If it were me, I'd remove all constructors whatsoever, e.g. https://gcc.godbolt.org/z/GroYYbazP

Same applies for `sf::Vector` as well. These should all be simple aggregates with no constructors, implicitly `constexpr`-friendly. 

I also would be extremely surprised if any measurable compilation time increase would occur as a result of changing `sf::Color` or `sf::Vertex` to be inline. They're basic classes, with almost no dependencies whatsoever. It also seems beneficial for run-time performance to make these types easier for the compiler to inline. I'd want proof that compilation time significantly increases in order to revert this change.

---

> We don't yet do it, but do you think it would make sense to pass classes like `sf::Vector2f` by value?

I am not sure, we should measure this. With inlining, it wouldn't matter. On ABI boundaries, performance might differ a lot between Windows and other architectures. Then again, if you have a bottleneck that relies on `sf::Vector2f` being on an ABI boundary that cannot be inlined... is it really SFML's problem?

---

> Maybe we could even go for `std::array`. Doesn't have to be part of this PR, of course.

+1. Also, consider this PR as a "throw everything in". My recommendation would be to start a SFML 3 branch, and gradually merge smaller PRs into it after review. We can use this one as testing grounds.

---

> I fully agree that we shouldn't get lost in endless discussions and let this PR go stale. Especially if it will become part of a "SFML 3 WIP version", it's not critical that everything is perfect right from the start; and we can do further (breaking) changes where necessary.

This is also a reasonable way forward. Make some progress, then refine it. I am honestly up for anything!

SuperV1234:
Here's a compilation time analysis created with `-ftime-trace` and ClangBuildAnalyzer for this PR, after the removal of `<algorithm>` from headers:
https://gist.github.com/SuperV1234/8c906f53fde8baca315f169c324ae17d

dgcor:
Maybe the next version should also be compatible with C++20 modules. The major issue I find is that you can't `import` SFML headers because of some things. Static variables can't be used with modules. that means static colors can't be used without redefining them. the network include is the worst because it uses sf::Time::Zero in one function signature.

some workarounds I use:

https://github.com/dgcor/DGEngine-Modules/blob/master/DGEngine.core.modules/src/Modules/sfml.ixx

Bromeon:
> For me, the argument is the complete opposite. `sf::Color` and `sf::Vertex` should be aggregates, which are implicitly `constexpr`. If it were me, I'd remove all constructors whatsoever, e.g. https://gcc.godbolt.org/z/GroYYbazP

I don't think it makes much sense constructing half the object (like R and G values for Color) and let the rest default-initialize. Aggregates may be nice from a code-amount point of view, but their C++ implementation is not great, e.g. compared to Kotlin `data class`, Rust `struct` or even Java `record`.

What is the benefit of compile-time `Vertex`? Where do we draw the boundary for other similar classes, like the event-related ones?

In other words: for me it's not enough justification that something is _possible_ in C++17, it needs to have a concrete advantage. Second, we should decide on some conventions _when_ we use certain features and when not (at least for non-obvious cases). Otherwise we're making the code harder to understand with `constexpr` in some, but not all places. Unfortunately, modern idioms in C++ are often more verbose, which puts the gains a bit into perspective.

---

> I am not sure, we should measure this. With inlining, it wouldn't matter. On ABI boundaries, performance might differ a lot between Windows and other architectures. Then again, if you have a bottleneck that relies on `sf::Vector2f` being on an ABI boundary that cannot be inlined... is it really SFML's problem?

The advantage would be that `Vector2f` is more concise than `const Vector2f&`, as well as more convenient (one can directly modify the parameter). You would typically also pass `double` by value, which has the same size as `Vector2f` or `Vector2i`.

---


> Maybe the next version should also be compatible with C++20 modules. The major issue I find is that you can't `import` SFML headers because of some things. Static variables can't be used with modules. that means static colors can't be used without redefining them. the network include is the worst because it uses sf::Time::Zero in one function signature.

If I understand you correctly, you mean "make SFML's API module-friendly" and not "rewrite SFML in modules", right? 🙂 

Would your issues be solved by making `Color` and `Time` use `constexpr`?

dgcor:
> If I understand you correctly, you mean "make SFML's API module-friendly" and not "rewrite SFML in modules", right? slightly_smiling_face
> 
> Would your issues be solved by making `Color` and `Time` use `constexpr`?

Yes, that's what I mean. What I did for colors and sf::Time::Zero is to replace occurrences with a constructor (sf::miliseconds(0) or sf::Color(0,0,0,0))

You can see this blog post to see the suggested changes: https://devblogs.microsoft.com/cppblog/moving-a-project-to-cpp-named-modules/

I guess making them constexpr won't hurt. Colors aren't a big problem, you can always redefine them in your own project. The big problem is adding default parameters to static variables like in the network header where one function has val = sf::Time::Zero in the signature. also macros have to be redefined, but those aren't a problem. of course if they were constexpr values it would work for both pre/post modules code.

To add colors I think the easiest would be to define them in the header as constexpr. that should work for both pre/post modules code.

eXpl0it3r:
Do you want to keep this open for the GitHub Action checks and/or to show what could be possible, or can we close this one?

sam20908:
Whatever action, I think this can be kept as a branch, for people who want C++17 (like me).

sam20908:
Also, I'm not sure what the maintainers would think of this, but I think this could be a good fit for SFML 3.0.

ChrisThrasher:
I feel like this list of C++17 changes is mostly complete. How much longer do we need to keep this PR open?

Bromeon:
> I feel like this list of C++17 changes is mostly complete. How much longer do we need to keep this PR open?

I agree with this. We can keep working on a branch without having this PR open. I'm closing it, so the list of open PRs/issues is more focused on actual tasks. Of course, if I'm mistaken, feel free to reopen 🙂

PR #1849, opened 2021-11-21
Remove some include directives for unused headers

SuperV1234:

----

## Description

Self-explanatory. 

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI will be good enough.

### DIFF
--- a/examples/X11/X11.cpp
+++ b/examples/X11/X11.cpp
@@ -3,7 +3,6 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Window.hpp>
-#include <SFML/System/Err.hpp>
 
 #define GLAD_GL_IMPLEMENTATION
 #include "gl.h"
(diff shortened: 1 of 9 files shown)


eXpl0it3r:

Bromeon (approved):
Nice, thanks! 🙂 

Out of curiosity, did you use tooling to detect unused headers?

SuperV1234:
> Nice, thanks! 🙂
> 
> Out of curiosity, did you use tooling to detect unused headers?

Yes, but not in the way you think. I was actually benchmarking SFML's compilation time via [ClangBuildAnalyzer](https://github.com/SuperV1234/ClangBuildAnalyzer). The biggest bottleneck, by far, is `sf::Utf`, followed by `sf::String`. I was surprised to see that `sf::String` was being included many times, then I did some manual searches and found unnecessary includes.

We could try out this next: https://github.com/include-what-you-use/include-what-you-use

Issue #1840, opened 2021-11-01
Input method support on Linux


Edgaru089:
## Subject of the issue

Input methods (I'm using Fcitx) does not work with SFML on Linux.

I looked into the source code and found that SFML is simply generating TextEntered events on key presses.

A input method is a piece of software used to input characters not present on the keyboard, mostly for ideographic scripts which is not even composed of a alphabet, and must be entered by imaginative ways, like most significantly the famous CJK Ideographs.

## Your environment

Arch Linux, KDE Plasma, SFML 2.5.1

Fcitx4 with aur/fcitx-sogoupinyin

## Steps to reproduce

Sadly, for those who don't speak Chinese or Japanese, using a input method (and the fact that we even need it) is not very meaningful and thus difficult to reproduce. Even if you do know a little about one of those it's quite complicated to set up a working input method on Linux for those who don't use it on a daily basis.

If you have a working IM however, simply create a window and dump the TextEntered and KeyPressed events. 

## Behaviour

You will find they closely resemble each other and the window does not respond to the installed input method at all.

Edgaru089:
The sad thing about input methods on Linux is that there are so many frameworks, all of which are incompatible which each other. Take a look at the chart on the [Arch Wiki page](https://wiki.archlinux.org/title/Input_method) if you're interested. Each column is a input method framework patching the already cacophonous X input stack, and each row is a actual input method doing the real work, translating your keystrokes to characters.

X has a protocol called XIM (X Input Method), but it predates Unicode and uses 16-bit code pages, and is only there for legacy compatibility. I can hardly find any documentation on it, other than the offical X Specification, which is notoriously hard to understand.

Edgaru089:
Just went digging the SDL source, supporting Fcitx and ibus.

Turns out they both require (or at least SDL is using) a dbus connection to the local daemon! Simple grepping shows dbus is only used for RealtimeKit1 (thread priority), UPower (battery status), and Fcitx/ibus.

Using dbus might be too much a overhead for fixing a bug (or maybe, implementing a feature) not meaningful to many. Or at least, copy-pasting works with Unicode, and input method support on Linux is so scarce that people are getting used to it, with even utilities for entering text on an overlay, and copy-pasting them via clipboard with a hotkey.

Edgaru089:
Should we use DBus, or just leave it be? Or should we investigate using XIM?

eXpl0it3r:
[...]

PR #1850, opened 2021-11-21
X11: fix XIM input method support

Edgaru089:
## Description

This PR is a (good enough) fix to #1840.

It fixes the existing XIM code, supporting input methods (at least with Fcitx5).

### How to test

Dump all the KeyPressed/TextInput event from any window. If you don't have a input method, nothing should change. If you do have one, it should be working now.

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode(800, 600), "Window");
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        sf::Event event;
        {
            switch (e.type) {
                case sf::Event::Closed:
                    win.close();
                    break;
                case sf::Event::KeyPressed:
                    printf("Key pressed: %d\n", (int)e.key.code);
                    break;
                case sf::Event::KeyReleased:
                    printf("Key released: %d\n", (int)e.key.code);
                    break;
                case sf::Event::TextEntered:
                    printf("Text entered: %d\n", (int)e.text.unicode);
                    break;
            }
        }

        window.clear();
        window.display();
    }
}
```

### DIFF
--- a/src/SFML/Window/Unix/Display.cpp
+++ b/src/SFML/Window/Unix/Display.cpp
@@ -40,6 +40,8 @@ namespace
     // The shared display and its reference counter
     Display* sharedDisplay = NULL;
     unsigned int referenceCount = 0;
+    XIM sharedXIM = NULL;
+    unsigned int referenceCountXIM = 0;
     sf::Mutex mutex;
 
     typedef std::map<std::string, Atom> AtomMap;
@@ -85,6 +87,56 @@ void CloseDisplay(Display* display)
         XCloseDisplay(display);
 }
 
+////////////////////////////////////////////////////////////
+XIM OpenXIM()
+{
+    Lock lock(mutex);
+
+    assert(sharedDisplay != NULL);
+
+    if (referenceCountXIM == 0)
+    {
+        // Create a new XIM instance
+
+        // We need the default (environment) locale and X locale for opening
+        // the IM and properly receiving text
+        // First save the previous ones (this might be able to be written more elegantly?)
+        const char* p;
+        std::string prevLoc((p = setlocale(LC_ALL, NULL)) ? p : "");
+        std::string prevXLoc((p = XSetLocaleModifiers(NULL)) ? p : "");
+
+        // Set the locales from environment
+        setlocale(LC_ALL, "");
+        XSetLocaleModifiers("");
+
+        // Create the input context
+        sharedXIM = XOpenIM(sharedDisplay, NULL, NULL, NULL);
+
+        // Restore the previous locale
+        if (prevLoc.length() != 0)
+            setlocale(LC_ALL, prevLoc.c_str());
+
+        if (prevXLoc.length() != 0)
+            XSetLocaleModifiers(prevXLoc.c_str());
+    }
+
+    referenceCountXIM++;
+
+    return sharedXIM;
+}
+
+////////////////////////////////////////////////////////////
+void CloseXIM(XIM xim)
+{
+    Lock lock(mutex);
+
+    assert(xim == sharedXIM);
+
+    referenceCountXIM--;
+
+    if ((referenceCountXIM == 0) && (xim != NULL))
+        XCloseIM(xim);
+}
 
 ////////////////////////////////////////////////////////////
 Atom getAtom(const std::string& name, bool onlyIfExists)
--- a/src/SFML/Window/Unix/Display.hpp
+++ b/src/SFML/Window/Unix/Display.hpp
@@ -56,6 +56,27 @@ Display* OpenDisplay();
 void CloseDisplay(Display* display);
 
 ////////////////////////////////////////////////////////////
+/// \brief Get the shared XIM context for the Display
+///
+/// This function increments the reference count of the XIM context,
+/// it must be matched with a call to CloseXIM.
+///
+/// It must be called with a display already opened.
+///
+/// \return XIM handle (a pointer) of the context
+///
+////////////////////////////////////////////////////////////
+XIM OpenXIM();
+
+////////////////////////////////////////////////////////////
+/// \brief Release a reference to the shared XIM context
+///
+/// \param xim XIM context to release
+///
+////////////////////////////////////////////////////////////
+void CloseXIM(XIM xim);
+
+////////////////////////////////////////////////////////////
 /// \brief Get the atom with the specified name
 ///
 /// \param name         Name of the atom
--- a/src/SFML/Window/Unix/WindowImplX11.cpp
+++ b/src/SFML/Window/Unix/WindowImplX11.cpp
@@ -101,7 +101,9 @@ namespace
         Bool checkEvent(::Display*, XEvent* event, XPointer userData)
         {
             // Just check if the event matches the window
-            return event->xany.window == reinterpret_cast< ::Window >(userData);
+            // The input method sometimes sends ClientMessages with a different window ID,
+            // our event loop has to process them for the IM to work
+            return (event->xany.window == reinterpret_cast< ::Window >(userData)) || (event->type == ClientMessage);
         }
 
         // Find the name of the current executable
@@ -803,7 +805,7 @@ WindowImplX11::~WindowImplX11()
 
     // Close the input method
     if (m_inputMethod)
-        XCloseIM(m_inputMethod);
+        CloseXIM(m_inputMethod);
 
     // Close the connection with the X server
     CloseDisplay(m_display);
@@ -1607,7 +1609,7 @@ void WindowImplX11::initialize()
     using namespace WindowsImplX11Impl;
 
     // Create the input context
-    m_inputMethod = XOpenIM(m_display, NULL, NULL, NULL);
+    m_inputMethod = OpenXIM();
 
     if (m_inputMethod)
     {
@@ -1845,27 +1847,31 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
         // Close event
         case ClientMessage:
         {
-            static Atom wmProtocols = getAtom("WM_PROTOCOLS");
-
-            // Handle window manager protocol messages we support
-            if (windowEvent.xclient.message_type == wmProtocols)
+            // Input methods might want random ClientMessage events
+            if (!XFilterEvent(&windowEvent, None))
             {
-                static Atom wmDeleteWindow = getAtom("WM_DELETE_WINDOW");
-                static Atom netWmPing = ewmhSupported() ? getAtom("_NET_WM_PING", true) : None;
+                static Atom wmProtocols = getAtom("WM_PROTOCOLS");
 
-                if ((windowEvent.xclient.format == 32) && (windowEvent.xclient.data.l[0]) == static_cast<long>(wmDeleteWindow))
+                // Handle window manager protocol messages we support
+                if (windowEvent.xclient.message_type == wmProtocols)
                 {
-                    // Handle the WM_DELETE_WINDOW message
-                    Event event;
-                    event.type = Event::Closed;
-                    pushEvent(event);
-                }
-                else if (netWmPing && (windowEvent.xclient.format == 32) && (windowEvent.xclient.data.l[0]) == static_cast<long>(netWmPing))
-                {
-                    // Handle the _NET_WM_PING message, send pong back to WM to show that we are responsive
-                    windowEvent.xclient.window = DefaultRootWindow(m_display);
+                    static Atom wmDeleteWindow = getAtom("WM_DELETE_WINDOW");
+                    static Atom netWmPing = ewmhSupported() ? getAtom("_NET_WM_PING", true) : None;
 
-                    XSendEvent(m_display, DefaultRootWindow(m_display), False, SubstructureNotifyMask | SubstructureRedirectMask, &windowEvent);
+                    if ((windowEvent.xclient.format == 32) && (windowEvent.xclient.data.l[0]) == static_cast<long>(wmDeleteWindow))
+                    {
+                        // Handle the WM_DELETE_WINDOW message
+                        Event event;
+                        event.type = Event::Closed;
+                        pushEvent(event);
+                    }
+                    else if (netWmPing && (windowEvent.xclient.format == 32) && (windowEvent.xclient.data.l[0]) == static_cast<long>(netWmPing))
+                    {
+                        // Handle the _NET_WM_PING message, send pong back to WM to show that we are responsive
+                        windowEvent.xclient.window = DefaultRootWindow(m_display);
+
+                        XSendEvent(m_display, DefaultRootWindow(m_display), False, SubstructureNotifyMask | SubstructureRedirectMask, &windowEvent);
+                    }
                 }
             }
             break;
@@ -1904,7 +1910,7 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
                 if (m_inputContext)
                 {
                     Status status;
-                    Uint8  keyBuffer[16];
+                    Uint8  keyBuffer[64];
 
                     int length = Xutf8LookupString(
                         m_inputContext,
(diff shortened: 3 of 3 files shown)


src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -814,19 +799,14 @@ WindowHandle WindowImplX11::getSystemHandle() const
...
-    // Handle the events for this window that we just picked out
-    while (!m_events.empty())
+    // Handle pending events
+    // Input methods might drop/change the events in the queue while they are filtered
+    // so we must process them one by one
+    while (XPending(m_display))
     {
-        event = m_events.front();
-        m_events.pop_front();
+        XNextEvent(m_display, &event);
```
Edgaru089, line 809: Wait a minute... XNextEvent pulls the next event from any window associated with this process... We should use XCheckWindowEvent?

Edgaru089: No... The IME requires ClientMessage events to be fed through XFilterEvent(), which cannot be obtained by XCheckWindowEvent.

We should fall back to using XCheckIfEvent?

Edgaru089:
I added a `printf()` in `checkEvent()`:
```cpp
// Filter the events received by windows (only allow those matching a specific window)
Bool checkEvent(::Display*, XEvent* event, XPointer userData)
{
	printf("CheckEvent: event type=%d, window=%ld, user=%ld\n",event->type,event->xany.window, userData);
	// Just check if the event matches the window
	return event->xany.window == reinterpret_cast< ::Window >(userData);
}
```
Looks like upon keypresses, the IME is firing ClientMessage events into the queue with a different Window ID:
```
...
CheckEvent: event type=33, window=94371847, user=94371845
CheckEvent: event type=33, window=94371847, user=94371845
CheckEvent: event type=33, window=94371847, user=94371845
... (33 = ClientMessage)
```
I added a `|| event->type == ClientMessage` to the condition and the IME worked again, at least for a single window.

This is what happened when I press a key:
```
                   This is the output: | CheckEvent: event type=2, window=94371845, user=94371845
The first KeyPress event is filtered.  | event type=2, filtered
A ClientMessage is sent by the IME.    | CheckEvent: event type=33, window=94371847, user=94371845
The ClientMessage is also filtered.    | event type=33, filtered
After that, the real KeyPress is sent. | CheckEvent: event type=2, window=94371845, user=94371845
This event is (at last!) not filtered. | event type=2

Main: Event::TextEntered, keycode=99(c)
```
Gonna test it with multiple windows.

Edgaru089:
With 2 windows, the IME sometimes glitches out if we switch to the other window while typing something.

I looked up some documentations, and it seems that the XIM object is per-display, and the XIC context is bound to a specific window. So I saved the XIM object to the namespace `WindowsImplX11Impl` and have all the window share the same XIM object. It works now! It still need to be mutex-protected though.

This is the snippet I used to test:
```cpp
#include <SFML/Window.hpp>
#include <SFML/Graphics.hpp>
#include <iostream>

sf::RenderWindow win1,win2;

bool processEvent(sf::Event& event,int win){
	switch(event.type){
	case sf::Event::Closed:
		win1.close();
		win2.close();
		return false;
	case sf::Event::KeyPressed:
		printf("(Win%d) Key Pressed: %d\n",win,event.key.code);
		break;
	case sf::Event::TextEntered:
		printf("(Win%d) Text Entered: %d\n",win,event.text.unicode);
		break;
	}

	return true;
}

{
	win1.create(sf::VideoMode(400, 200), "Window1");
	win1.setVerticalSyncEnabled(true);
	win2.create(sf::VideoMode(400, 200),"Window2");
	win2.setVerticalSyncEnabled(true);

	while (win1.isOpen() && win2.isOpen()) {
		sf::Event event;
		while (win1.pollEvent(event))
			if(!processEvent(event,1))
				return 0;
		while(win2.pollEvent(event))
			if(!processEvent(event,2))
				return 0;

		win1.clear(sf::Color::Magenta);
		win2.clear(sf::Color::Cyan);
		win1.display();
		win2.display();
	}
	return 0;
}
```

Edgaru089:
The updated code uses `XCheckIfEvent` to pick out events only for a single window.

It also keeps a global XIM object instead of creating a new one with every window. Maybe the commit name should include that?

Edgaru089:
Rebased with latest master. @eXpl0it3r Any updates?

eXpl0it3r:
I don't know much about the whole X11 setup and hope @binary1248 can have a closer look at this 🙂

eXpl0it3r:

binary1248:
I still don't quite understand why it was necessary to make such significant changes to the logic around `m_events`. `m_events` was introduced in #1291 to fix #1223. Removing this code risks re-introducing the bug the code was written to prevent.

I would prefer this PR is modified to affect the existing `m_events` logic as little as possible.

Alternatively, this would have to be tested on multiple different systems/configurations to verify that the original bug is not re-introduced. Maybe @naezith or @texus could help with this since they worked on the issue.

Edgaru089:
> I would prefer this PR is modified to affect the existing m_events logic as little as possible.

This is possible. The only things we need to do are:

- `setlocale(LC_ALL, "")` on XIM creation
- Call `XFilterEvent(&event, None)` on ClientMessage events
- Global XIM context across SFML windows (XICs are per-window)
- Send multiple TextInput events from one X event (there may be multiple codepoints in a single UTF-8 string from the raw event)

Working on it. Should we change the commit/pr name?

Edgaru089:
Rebased with latest master.

Sorry for the mess... It was indeed not necessary to change this much. To my defense, there are very little and even conflicting information on how to handle XIM correctly, not to mention with multiple windows.

I removed `m_events` because I observed some strange KeyPress events being sent both to the IM and the application early on while hacking the code. It was not necessarily related to `m_events` itself, but more likely because of not calling `XFilterEvent()` properly, or even the crappy test snippet I was using. The drastic code changes in the first commit were more of a guess and less of reasoning, and you have my more sincere apologies for that...

I rewrote the code, pushed the second commit and was never able to repeat the bug.

------

Just being curious, is it possible for this pull request to make it into SFML 2.6.0? I sure can do a back port.

binary1248:
There is no need to apologize for standard development practice. 😉

After all is said and done this PR should/would be squashed and rebased anyway, so any intermediate steps wouldn't have become part of the commit history.

Regarding the backport to 2.6.0: @eXpl0it3r will have to comment on that. I personally don't have anything against it.

src/SFML/Window/Unix/Display.cpp
```diff
@@ -85,6 +87,54 @@ void CloseDisplay(Display* display)
...
+        XSetLocaleModifiers("");
+
+        // Create the input context
+        sharedXIM = XOpenIM(sharedDisplay, nullptr, nullptr, nullptr);
+
+        // Restore the previous locale
+        if(prevLoc.length() != 0)
+            setlocale(LC_ALL, prevLoc.c_str());
+        if(prevXLoc.length() != 0)
+            XSetLocaleModifiers(prevXLoc.c_str());
```
binary1248, line 119: Formatting, and I would leave a space between the 2 blocks.

src/SFML/Window/Unix/Display.cpp
```diff
@@ -85,6 +87,54 @@ void CloseDisplay(Display* display)
...
+    }
+
+    referenceCountXIM++;
+    return sharedXIM;
```
binary1248, line 123: Would leave a space between these lines.

src/SFML/Window/Unix/Display.cpp
```diff
@@ -85,6 +87,54 @@ void CloseDisplay(Display* display)
...
+}
+
+////////////////////////////////////////////////////////////
+void CloseXIM(XIM xim)
+{
+    Lock lock(mutex);
+
+    assert(xim == sharedXIM);
+
+    referenceCountXIM--;
```
binary1248, line 133: Would leave a space under this line.

src/SFML/Window/Unix/Display.cpp
```diff
@@ -85,6 +87,54 @@ void CloseDisplay(Display* display)
...
+    if (referenceCountXIM == 0)
+        if (xim != nullptr)
+            XCloseIM(xim);
```
binary1248, line 136: These if statements can be combined.

src/SFML/Window/Unix/Display.hpp
```diff
@@ -55,6 +55,27 @@ Display* OpenDisplay();
...
+///
+/// \return XIM handle (a pointer) of the context
+///
+////////////////////////////////////////////////////////////
+XIM OpenXIM();
+
+////////////////////////////////////////////////////////////
+/// \brief Release a reference to the shared XIM context
+///
+/// \param display XIM context to release
```
binary1248, line 74: Wrong parameter name.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -101,7 +101,9 @@ namespace
         Bool checkEvent(::Display*, XEvent* event, XPointer userData)
         {
             // Just check if the event matches the window
-            return event->xany.window == reinterpret_cast< ::Window >(userData);
+            // The input method sometimes send ClentMessages with a different window ID,
```
binary1248, line 104: The input method sometimes sends ClientMessages with a different window ID,

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -101,7 +101,9 @@ namespace
...
+            // our event loop have to process that for the IM to work
```
binary1248, line 105: our event loop has to process them for the IM to work

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -101,7 +101,9 @@ namespace
...
+            return (event->xany.window == reinterpret_cast< ::Window >(userData)) || event->type == ClientMessage;
```
binary1248, line 106: Parentheses around RHS of the condition.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1915,16 +1921,24 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
                         &status
                     );
 
-                    if (length > 0)
+                    if(status == XBufferOverflow)
```
binary1248, line 1924: Missing space.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1915,16 +1921,24 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
...
+                        err() << "A TextEntered event has more than 64 bytes of UTF-8 input, and is discarded" << std::endl << "This means either you have typed a very long string (more than 20 chars), or your input method is broken in obscure ways." << std::endl;
```
binary1248, line 1925:
```cpp
err() << "A TextEntered event contains more than 64 bytes of UTF-8 data, and "
         "has been discarded\nThis means either you have typed a very long string "
         "(more than 20 chars), or your input method is broken in obscure ways." << std::endl;
```

binary1248 (changes requested):

Edgaru089:
All fixed. You really have those eagle eyes👍

eXpl0it3r:
Yeah, I think it could be a great addition and is isolated enough for SFML 2.6.
Could you rebase the commit onto the `2.6.x` branch and change the PR target to the 2.6.x branch?

Edgaru089:
@eXpl0it3r I still don't quite get it. Is it the case that the feature changes on 2.6.x will be ported to 3.0 once all the major API changes are sorted out? That sounds like a major undertaking.

eXpl0it3r:
No, we'll continuously merge changes from 2.6.x to master.
It's a bit work, but at the same time, we don't really plan to add much more to SFML 2.6

Edgaru089:
Rebased with `2.6.x`.

src/SFML/Window/Unix/Display.cpp
```diff
@@ -85,6 +87,56 @@ void CloseDisplay(Display* display)
...
+void CloseXIM(XIM xim)
+{
+    Lock lock(mutex);
+
+    assert(xim == sharedXIM);
+
+    referenceCountXIM--;
+
+    if ((referenceCountXIM == 0) && (xim != NULL))
+            XCloseIM(xim);
```
binary1248, line 138: Alignment.

src/SFML/Window/Unix/Display.hpp
```diff
@@ -55,6 +55,27 @@ Display* OpenDisplay();
...
+///
+/// \return XIM handle (a pointer) of the context
+///
+////////////////////////////////////////////////////////////
+XIM OpenXIM();
+
+////////////////////////////////////////////////////////////
+/// \brief Release a reference to the shared XIM context
+///
+/// \param XIM context to release
```
binary1248, line 74: The first token after \param has to be the parameter name (case sensitive).

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -101,7 +101,9 @@ namespace
         Bool checkEvent(::Display*, XEvent* event, XPointer userData)
         {
             // Just check if the event matches the window
-            return event->xany.window == reinterpret_cast< ::Window >(userData);
+            // The input method sometimes sends ClentMessages with a different window ID,
```
binary1248, line 104: ClientMessages

binary1248 (changes requested):

eXpl0it3r:

binary1248 (approved):

binary1248:
Looks good to me, although I must admit I don't really have the technical expertise or experience to test these changes because I have never had to type using an input method before. Perhaps there are others willing to give this a try.

eXpl0it3r:
Thank you for diving into this! 🙂

kimci86:
I noticed since this commit that more events are generated, which looks like a bug to me.

Testing with [SFML-Input](https://github.com/eXpl0it3r/SFML-Input) and SFML feature/scancode branch on Ubuntu 20.04: when pressing a key, two KeyPressed events are generated with a small delay between them, instead of one event. If the key is held down, many KeyPressed events are regularly generated when the key is down (which is normal) but also after the key is released (not normal) for some time.

I did not test the new support input method though. I never used such input method.

Edgaru089:
@kimci86 Can you try again with `export XMODIFIERS=@im=none`? This effectively disables the entire mess with XIM.

Maybe you have the input method framework IBus installed. It comes with GNOME and is notoriously hard to replace, many of us end up giving up GNOME just to not use IBus.

kimci86:
Running again with `XMODIFIERS=@im=none` restores the expected behavior indeed :+1:
`GTK_IM_MODULE` is empty and `QT_IM_MODULE` is `ibus` in my environment.

Is there a way to have this behavior by default, without having to set an environment variable?

Edgaru089:
@kimci86 IBus is, how to say... Neglected. It kind of works, but most people who care more about input methods often use others like Fcitx or Uim. Also it's part of good-old GNOME. The fault is most likely IBus's IMHO.

> Is there a way to have this behavior by default, without having to set an environment variable?

AFAIK I'm afraid no, but I'm not an expert. I've never used IBus before. Gonna do some testing.

Edgaru089:
@kimci86 Oh no, the fault is mine in fact...

If you look at [WindowImplX11.cpp:L1880](https://github.com/SFML/SFML/blob/master/src/SFML/Window/Unix/WindowImplX11.cpp#L1880)
```cpp
// Key down event
case KeyPress:
{
	// Fill the event parameters
	Event event;
	event.type        = Event::KeyPressed;
	event.key.code    = (......)
	pushEvent(event);

	// Generate a TextEntered event
	if (!XFilterEvent(&windowEvent, None)) // Feed the event to the IME, discard it if requested
	{
		event.type = Event::TextEntered;
		......
	}
}
```

Clearly there's something wrong here, and the `if (!XFilterEvent(&windowEvent, None))` should go before sending the KeyPress event. I'm sorry...

I wasn't able to catch that mainly because when the IME takes some text, IBus sends a KeyPress without text input attached, while Fcitx and XWayland (what I use) sends no KeyPress at all. Thus if the previous event is not filtered, it gets duplicated by IBus.

PR #1851, opened 2021-11-23
MP3 reader support using minimp3

eXpl0it3r:
## Description

Accidently pushed the wrong changes, closing the other PR #1796 
Sorry about that, @lieff
Here's the rebased changes

This PR is related to the issue #1232 

## How to test this PR?

The sound example was extended and can be executed as a test


eXpl0it3r:
@Bromeon can you give this another look, as you reviewed the original PR #1796

src/SFML/Audio/SoundFileReaderMp3.hpp
```diff
@@ -0,0 +1,129 @@
...
+////////////////////////////////////////////////////////////
+
+#ifndef SFML_SOUNDFILEREADERMP3_HPP
+#define SFML_SOUNDFILEREADERMP3_HPP
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#define NOMINMAX          // To avoid windows.h and std::min issue.
+#define MINIMP3_NO_STDIO  // Minimp3 control define, eliminate file manipulation code which useless  here.
```
kimci86, line 32: There is a typo in the comment. 'is' is missing.

include/SFML/Audio/InputSoundFile.hpp
```diff
@@ -63,8 +63,11 @@ class SFML_AUDIO_API InputSoundFile : NonCopyable
     ////////////////////////////////////////////////////////////
     /// \brief Open a sound file from the disk for reading
     ///
-    /// The supported audio formats are: WAV (PCM only), OGG/Vorbis, FLAC.
+    /// The supported audio formats are: WAV (PCM only), OGG/Vorbis, FLAC, MP3.
     /// The supported sample sizes for FLAC and WAV are 8, 16, 24 and 32 bit.
+    /// Because of minimp3_ex limitation, for MP3 files with big (>16kb) APEv2 tag,
+    /// it may not be properly removed, tag data will be treated as MP3 data
+    /// and there is a low chance of garbage decoded at the end of file.
```
Bromeon, line 70: Thanks for this update! I've read again your [GitHub comment](https://github.com/SFML/SFML/issues/1232#issuecomment-641633168) explaining things. Maybe as you suggested, we can still provide a link to the ReadMe:

```cpp
/// See also: https://github.com/lieff/minimp3
```

The ReadMe in https://github.com/lieff/minimp3 states:

> At end of stream just pass rest of the buffer, sync procedure should work even with just 1 frame in stream (except for free format and **garbage at the end can mess things up, so id3v1 and ape tags must be removed first**).

Possibly we should also create an issue for this problem, and address it later so _all_ MP3 files can be read, and we can get rid of this leaky abstraction. It's of course a bit tricky, since the API relies on chunk-by-chunk streaming without lookahead.

include/SFML/Audio/InputSoundFile.hpp
```diff
@@ -63,8 +63,11 @@ class SFML_AUDIO_API InputSoundFile : NonCopyable
     ////////////////////////////////////////////////////////////
     /// \brief Open a sound file from the disk for reading
     ///
-    /// The supported audio formats are: WAV (PCM only), OGG/Vorbis, FLAC.
+    /// The supported audio formats are: WAV (PCM only), OGG/Vorbis, FLAC, MP3.
     /// The supported sample sizes for FLAC and WAV are 8, 16, 24 and 32 bit.
+    /// Because of minimp3_ex limitation, for MP3 files with big (>16kb) APEv2 tag,
```
Bromeon, line 68: Empty line (only `///`) before, to create separate paragraph for the implementation details.

src/SFML/Audio/SoundFileReaderMp3.cpp
```diff
@@ -0,0 +1,130 @@
...
+{
+namespace priv
+{
+////////////////////////////////////////////////////////////
+bool SoundFileReaderMp3::check(InputStream& stream)
+{
+    Uint8 header[10];
+    if (stream.read(header, sizeof(header)) < sizeof(header))
+        return false;
+    if (!memcmp(header, "ID3", 3) && !((header[5] & 15) || (header[6] & 0x80) || (header[7] & 0x80) || (header[8] & 0x80) || (header[9] & 0x80)))
```
Bromeon, line 62: Maybe short explanation regarding the header format would help understanding here.
The second part, the `&& !(...)` checks if _none_ of the bytes 5 to 9 match a certain value/bit pattern?

Bromeon (commented):
Thanks for the update, @lieff and @eXpl0it3r!
Most looks good now, added a few comments.

Has someone tested this with common MP3 files yet?

---

> Accidently pushed the wrong changes, closing the other PR #1796

This happened to me as well in the past; you should be able to reopen it by pushing to the a with same name (and possibly clicking the "Reopen" button). Just for next time 🙂

eXpl0it3r:
> This happened to me as well in the past; you should be able to reopen it by pushing to the a with same name (and possibly clicking the "Reopen" button). Just for next time 🙂

That is true if you own the repository you're pushing to. But I accidentally reset lieff's master branch to the official master branch state and by doing so, the PR got closed and also removed the access to push to their repository.
As I didn't want to "lose" the changes, the only option was to push it on our side. 😅

src/SFML/Audio/SoundFileReaderMp3.cpp
eXpl0it3r: @lieff do you have some insights to what this checks exactly?

include/SFML/Audio/InputSoundFile.hpp
eXpl0it3r: Since this essentially is a limitation of minimp3, do we need to create an issue on our side?

src/SFML/Audio/SoundFileReaderMp3.cpp
lieff: Hi) Mp3 detection usually more complex and ideally mp3dec_detect_cb should be used here. But big portion of file can be read here to do it properly.
So I write here light version that only checks for ID3 tag and first valid frame header. The `&& !(...)` part checks that ID3 tag header is valid.
This light method will fail on l3-sin1k0db.bit test vector for example, and can give false positives on non-mp3 files (because only 1 mp3 frame header is checked). But should work on most normal mp3 files.

eXpl0it3r: Ah okay, I'll change the check to `std::memcmp(...) == 0` as it's IMHO more understandable as the negated version and add a short comment. 
I think for now, this is good enough. If people run into issues, then we'll have to double check again. 👍

src/SFML/Audio/SoundFileReaderMp3.cpp
```diff
@@ -0,0 +1,133 @@
...
+namespace priv
+{
+////////////////////////////////////////////////////////////
+bool SoundFileReaderMp3::check(InputStream& stream)
+{
+    Uint8 header[10];
+    if (stream.read(header, sizeof(header)) < sizeof(header))
+        return false;
+
+    // Basic MP3 format detection by checking for the ID3 tag and a valid frame header
```
kimci86, line 63: "or" would more correct than "and" in this comment. The logic is to look first for an ID3 header, and if it is not there (which is acceptable because metadata is optional) then look for a valid MP3 frame header.
Maybe putting the condition to check for an ID3 header in a function with a good name would help readability and make this comment unnecessary.

include/SFML/Audio/InputSoundFile.hpp
Bromeon: I'm not sure, is it really a limitation of minimp3 in the sense that a better _implementation_ (not API) could address it? I understood that as a consequence of minimp3's streaming/batch processing, it cannot detect the id3v1/ape tag when it's larger than one batch window (16kB).

Also, for the SFML user, minimp3 is an implementation detail. If some MP3 files cannot be played, that still negatively affects user experience. But I'm also OK if we only create an issue when it's becoming a real problem; this pull request should contain enough information.

eXpl0it3r: I don't really like to open issues, unless we have a clear plan on how to tackle it. When users open issues, then we at least know that it has a "real world" impact, while this here is just theoretical guessing (unless we notice that it's a more common issue, during testing).

Bromeon: That's a good take 👍

src/SFML/Audio/SoundFileReaderMp3.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#define MINIMP3_IMPLEMENTATION // Minimp3 control define, places implementation in this file.
+#define NOMINMAX          // To avoid windows.h and std::min issue
+#define MINIMP3_NO_STDIO  // Minimp3 control define, eliminate file manipulation code which is useless here
+#include <minimp3_ex.h>
+#undef NOMINMAX
+#undef MINIMP3_NO_STDIO
```
kimci86, line 33: The problem here is that `MINIMP3_IMPLEMENTATION` is still defined when including `minimp3_ex.h` a second time in  `SoundFileReaderMp3.hpp` just below. There is no include guard for the implementation part, so the implementation is repeated.
I can see a few solutions:
- `#undef MINIMP3_IMPLEMENTATION`
- Move this include for the implementation after having included `SoundFileReaderMp3.hpp`.
- Add an include guard in `minimp3_ex.h` as it was already done in `minimp3.h` (see: https://github.com/lieff/minimp3/commit/644e0fb7fed34f803b6634f72e5ad8cc20a520f7).

eXpl0it3r: @lieff could we add an include guard in `minimp3_ex.h` as well?

lieff: Sure, added upstream https://github.com/lieff/minimp3/commit/50d2aaf360a53653b718fead8e258d654c3a7e41

eXpl0it3r: I've updated the header, but seems like that's not enough, the unity build still keeps failing... 🤔

kimci86: Here is what happens in a unity build batch:
```c++
#include <minimp3_ex.h>
```
This pulls `minimp3_ex.h` declarations.
This declarations part includes `minimp3.h` which pulls `minimp3.h` declarations.

Then, later:
```c++
#define MINIMP3_IMPLEMENTATION
#include <minimp3_ex.h>
```

As they were already included, `minimp3_ex.h` declarations are skipped.
It also skips `#includes <minimp3.h>`, so `minimp3.h` implementation is not included.
Then comes `minimp3_ex.h` implementation which cannot find `minimp3.h` implementation, as it was not included.

A workaround is to include both `minimp3.h` and `minimp3_ex.h`, but it can be fixed in minimp3 directly with the following patch. (I also found that size_t type was missing.)  @lieff 
```patch
diff --git a/extlibs/headers/minimp3/minimp3_ex.h b/extlibs/headers/minimp3/minimp3_ex.h
index 4ccc5963..64d2f39f 100644
--- a/extlibs/headers/minimp3/minimp3_ex.h
+++ b/extlibs/headers/minimp3/minimp3_ex.h
@@ -7,6 +7,7 @@
     See <http://creativecommons.org/publicdomain/zero/1.0/>.
 */
 #include "minimp3.h"
+#include <stddef.h> // size_t
 
 /* flags for mp3dec_ex_open_* functions */
 #define MP3D_SEEK_TO_BYTE   0      /* mp3dec_ex_seek seeks to byte in stream */
@@ -130,6 +131,8 @@ int mp3dec_ex_open_w(mp3dec_ex_t *dec, const wchar_t *file_name, int flags);
 
 #if defined(MINIMP3_IMPLEMENTATION) && !defined(_MINIMP3_EX_IMPLEMENTATION_GUARD)
 #define _MINIMP3_EX_IMPLEMENTATION_GUARD
+
+#include "minimp3.h"
 #include <limits.h>
 
 static void mp3dec_skip_id3v1(const uint8_t *buf, size_t *pbuf_size)
```

eXpl0it3r: vittorioromeo managed to solve the warnings

eXpl0it3r:
Even though it's currently not a "symmetric" implementation of decoder/encoder it makes sense to also include it in SFML 2.6
We might get more "can't save as MP3" forum posts, but that's small pain to deal with, when on the other hand users can load MP3 files.

But before merging this, we should spend some time, testing it with various mp3. 🙂

lieff:
For encoder part, https://github.com/cpuimage/tinymp3 can be used, if quality is not the primary concern.

vittorioromeo:
> But before merging this, we should spend some time, testing it with various mp3. 🙂

Anything in particular you'd like to test out

eXpl0it3r:
Grab some mp3's from your collection (if that's still a thing) and play them with SFML.
Best also when they're from different source, times, encoders, with and without tags, etc.
One could also try common/well-known audio applications and see whether their output mp3 play fine with SFML.

vittorioromeo:
Some testing on Windows 10 x64, using MSYS2/MinGW:

- I was able to play the provided `ding.mp3` just fine.

- I wasn't able to play `minimal.mp3` (attached), with the following error: 
    > Failed to open sound file "resources/minimal.mp3" (format not supported)
    - The error is pretty unhelpful. Why is the format not supported?
    - I can play the file in any music player just fine.
   
- I was able to play the `T-Range.mp3` file (attached) just fine.

- I was able to play the `swap.mp3` file (attached) just fine.

- I was able to play the `click.mp3` file (attached) just fine.

This is what VLC reports for `minimal.mp3`:
vlc_minimal

This is what VLC reports for `ding.mp3`:
vlc_ding

Test files:
[test_mp3_resources.zip](https://github.com/SFML/SFML/files/7669605/test_mp3_resources.zip)

lieff:
`minimal.mp3` is not regular mp3 file, but mp4 file with mp3 inside:

image

So yeah, this intended.
I've tested some files too and they all play fine.

vittorioromeo:
> `minimal.mp3` is not regular mp3 file, but mp4 file with mp3 inside:
> 
> image
> 
> So yeah, this intended. I've tested some files too and they all play fine.

Thanks for the clarification. I am quite ignorant regarding music formats, is it uncommon to have an mp3 inside a mp4 container? Would it be hard to support it in minimp3, if it's reasonably common to encounter?

lieff:
Not sure how it's common. I've never seen just mp3 audio in mp4 before, aac audio is more native and common to mp4.

Minimp3 is not the place where demuxing should be done. https://github.com/lieff/minimp4 can be used for demuxing.

Note that mp4 container can contain different audio and video codecs and multiple tracks, so such cases should be handled somehow. For example limit to only audio in mp3 and first track, return error otherwise.

Also note that there more popular containers: avi, wmv(wma), mkv, webm, so full proper support for containers is big. Better use ffmpeg or gstreamer in this case.

vittorioromeo (approved):
From my perspective, I am happy with this PR. I would prefer merging it as it is, get some community testing/feedback, and iterate on any possible issue/improvement that might arise from that.

vittorioromeo:
@eXpl0it3r: happy to merge?

eXpl0it3r:
Thanks a lot @lieff and everyone involved, looking forward to adding MP3 as supported audio format to the changelog 🙂

ChrisThrasher:
Can we close #1232?

eXpl0it3r:
Since nobody seems to read the issue description, I guess we can close it and open a new issue for implementing a MP3 decoder 😄

PR #1852, opened 2021-11-25
Update to make compilation possible of 2.4.x on Arch.

poqudrof:
Hello, 

I could not compile SFML 2.4 on my linux (Arch Linux) due to OpenGL updates I guess. 

## Disclaimer: 

I did not check the forum, **I did not write code except copy more recent code and comment a line**. 

----

## Description

My answer to the problem was brute force, I just copied the version of the incriminated files from 2.5.x which compiled OK. 

Now it compiles and I made a PKGBUILD to create a package from it : 
https://gist.github.com/poqudrof/96eff1a0f143e0e2dd13d981944bc78f 

I may release it as an AUR package if it gets included in SFML sfml-2.4 just like Nakano did : https://aur.archlinux.org/packages/sfml2.3/ 

This PR is not related to an issue.

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

**This affects only Linux.** 

## How to test this PR?

Check out the code and compile on Arch linux, or try the PKGBUILD linked above. 

Please let me know if I need to update anything, it would be great for Arch Linux users that need to maintain compatibility with previous versions. 

Regards, 

Jeremy.

### DIFF
--- a/src/SFML/Window/Unix/GlxContext.cpp
+++ b/src/SFML/Window/Unix/GlxContext.cpp
@@ -25,8 +25,8 @@
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
+#include <SFML/Window/Unix/WindowImplX11.hpp> // important to be included first (conflict with None)
 #include <SFML/Window/Unix/GlxContext.hpp>
-#include <SFML/Window/Unix/WindowImplX11.hpp>
 #include <SFML/Window/Unix/Display.hpp>
 #include <SFML/System/Mutex.hpp>
 #include <SFML/System/Lock.hpp>
--- a/src/SFML/Window/Unix/GlxExtensions.hpp
+++ b/src/SFML/Window/Unix/GlxExtensions.hpp
@@ -25,11 +25,12 @@
 #ifndef SF_POINTER_C_GENERATED_HEADER_GLXWIN_HPP
 #define SF_POINTER_C_GENERATED_HEADER_GLXWIN_HPP
 
-#ifdef __glxext_h_
+#if defined(__glxext_h_) || defined(__glx_glxext_h_)
 #error Attempt to include glx_exts after including glxext.h
 #endif
 
 #define __glxext_h_
+#define __glx_glxext_h_
 
 #include <X11/Xlib.h>
 #include <X11/Xutil.h>


src/SFML/Window/Unix/GlxContext.cpp
```diff
@@ -172,6 +172,9 @@ m_ownsWindow(false)
 ////////////////////////////////////////////////////////////
 GlxContext::~GlxContext()
 {
+    // Notify unshared OpenGL resources of context destruction
+    // cleanupUnsharedResources();
+
```
Bromeon, line 177: This code is inactive, is this intended?

src/SFML/Window/Unix/GlxContext.cpp
```diff
@@ -1,7 +1,7 @@
 ////////////////////////////////////////////////////////////
 //
 // SFML - Simple and Fast Multimedia Library
-// Copyright (C) 2007-2017 Laurent Gomila (laurent@sfml-dev.org)
+// Copyright (C) 2007-2018 Laurent Gomila (laurent@sfml-dev.org)
```
Bromeon, line 4: The `2.4.x` branch uses 2017, let's keep it that way.
Could you revert all the year changes? If possible with `git commit --amend` to keep 1 commit 🙂

Bromeon:
Thanks! I don't have Arch Linux myself, maybe someone else can additionally test it.

Bromeon (commented):
Thanks! I don't have Arch Linux myself, maybe someone else can additionally test it.

src/SFML/Window/Unix/GlxContext.cpp
poqudrof: Yes it is intended, I just checked the method does not exist in 2.4, or did have another name.

poqudrof:
I have to use it to run the Orbbec Astra samples : https://orbbec3d.com/download-sdk/ 

While Ubuntu moves slowly, Arch is always up to date. I find it quite common to have older 
versions of software to be able to run Ubuntu packages. 

Unfortunately Orbbec does not have a public repository with the sources that I know of. Their githup repo does not keep up with their releases. So it is easier do maintain older libraries than upgrade their examples.

src/SFML/Window/Unix/GlxContext.cpp
eXpl0it3r: Then better to not add it. If it wasn't there and isn't used now, it shouldn't be added as commented out code 😉

poqudrof: It's done !

Bromeon (approved):
Looks good now.

So are you going to use this branch directly? There won't be any released package on Arch, right?

eXpl0it3r:
CI failures are expected, since they're not setup for SFML 2.4

I'm also not quite sure how you're going to use this now.
If you build from source anyways, what's stopping you from using SFML 2.5? They are API compatible.

poqudrof:
I want to use this in an AUR package for Arch linux.
I updated an existing build file to create new package, just like it was done previously by [Nakano](https://aur.archlinux.org/packages/?K=nakano&SeB=m).

I will ask him to update the [current package](https://aur.archlinux.org/packages/astra-sdk/) to depend on this newly created package that will be built from source. The update requires SFML 2.4.

eXpl0it3r:
Okay, for me it's fine to have this fix on the 2.4.x branch as well, but seems there's nothing really stopping you from just using SFML 2.5 and maybe having to add some sym link.

PR #1856, opened 2021-12-03
Use list-initialization to zero-initialize 'JoystickImpl.hpp' structures

vittorioromeo:
## Description

This PR introduces the first changes that actually require C++11 support to the codebase. It's a simple refactoring changes that removes a public dependency on `<algorithm>` and simplifies the initialization of some data members in `JoystickImpl.hpp`.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI and testing existing joystick examples would be enough.

### DIFF
--- a/src/SFML/Window/JoystickImpl.hpp
+++ b/src/SFML/Window/JoystickImpl.hpp
@@ -30,7 +30,6 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Config.hpp>
 #include <SFML/Window/Joystick.hpp>
-#include <algorithm>
 
 
 namespace sf
@@ -43,14 +42,8 @@ namespace priv
 ////////////////////////////////////////////////////////////
 struct JoystickCaps
 {
-    JoystickCaps()
-    {
-        buttonCount = 0;
-        std::fill(axes, axes + Joystick::AxisCount, false);
-    }
-
-    unsigned int buttonCount;               //!< Number of buttons supported by the joystick
-    bool         axes[Joystick::AxisCount]; //!< Support for each axis
+    unsigned int buttonCount{0};              //!< Number of buttons supported by the joystick
+    bool         axes[Joystick::AxisCount]{}; //!< Support for each axis
 };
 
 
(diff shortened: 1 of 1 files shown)


eXpl0it3r (approved):
I'm okay with this way have handling structs 👍

Bromeon:
The problem with aggregates is that they allow partial initialization, and that order of fields matters.
It's typically not helpful to initialize only the first N fields (and leave the others to defaults), such as
```cpp
JoystickState state{false};
```
and I think we should definitely not allow it for types in the public API.

That said, it might be OK for internal ones, but I also don't see much benefit to be honest. It's typically less error-prone if a constructor ensures that all fields are provided with a value, especially in light of future refactorings (adding fields).

Inline initialization for members looks good (instead of `std::fill()` etc), and it's a bit weird that we hadn't just used the constructor initializer list before.

vittorioromeo:
> The problem with aggregates is that they allow partial initialization, and that order of fields matters. It's typically not helpful to initialize only the first N fields (and leave the others to defaults), such as
> 
> ```c++
> JoystickState state{false};
> ```

I agree with that, but we should also be pragmatic about it. A small internal type such as `JoystickState` is very unlikely to end up being used incorrectly in a way that's not caught by testing or during developmet.

> and I think we should definitely not allow it for types in the public API.

I'm not too sure about that. For example, `sf::Vector2<T>` and `sf::Color` are great candidates for public-facing aggregates. We would simplify the implementation, get `constexpr` support for free, and it would allow users to use [designated initializers](https://en.cppreference.com/w/cpp/language/aggregate_initialization#Designated_initializers):

```cpp
sf::Vector2f v{.x{10}, .y{15}};
sf::Color c{.r{255}, .g{255}, .b{255}, .a{100}};
```
 
> That said, it might be OK for internal ones, but I also don't see much benefit to be honest. It's typically less error-prone if a constructor ensures that all fields are provided with a value, especially in light of future refactorings (adding fields).
> 
> Inline initialization for members looks good (instead of `std::fill()` etc), and it's a bit weird that we hadn't just used the constructor initializer list before.

True, we could have just used C++98 value-initialization in the constructor. I am happy to re-introduce a constructor if you strongly feel that it should be done, but I honestly don't think it's a big deal in this particular situation.

Bromeon:
As mentioned, for internal types I don't have a very strong opinion. If we see that one way (either aggregates or classical constructors) cause issues, we can always refactor. That said, I would not go on a hunt now to change all the signatures -- after all, being pragmatic also means that things which are already written and well-working don't have priority in being updated 😉 

> I'm not too sure about that. For example, `sf::Vector2<T>` and `sf::Color` are great candidates for public-facing aggregates. We would simplify the implementation, get `constexpr` support for free, and it would allow users to use [designated initializers](https://en.cppreference.com/w/cpp/language/aggregate_initialization#Designated_initializers):
> 
> ```c++
> sf::Vector2f v{.x{10}, .y{15}};
> sf::Color c{.r{255}, .g{255}, .b{255}, .a{100}};
> ```

I disagree here. First, this is a very convoluted syntax, hardly anyone would use it over 
```cpp
sf::Vector2f v(10, 15);
sf::Color c(255, 255, 255, 100);
```
especially since these 2 classes are so central to SFML (or even gamedev in general), that no one will accidentally confuse the order and try to construct an YX-vector. 

Second, this is a trade-off and would allow constructions like
```cpp
sf::Vector2f v{10};
sf::Color c{255, 255};
```
which are highly misleading, and our job as API designers to inhibit.

In general, we should consider practicability and user-friendliness when switching over to C++17 features, and not only do it because it's possible or more modern.

PR #1857, opened 2021-12-03
Use 'nullptr' instead of 'NULL'

vittorioromeo:
## Description

This PR changes all the uses of `NULL` to `nullptr`, and also updates documentation to say "a null pointer" instead of "a NULL pointer" so that grepping for "NULL" does not result in any occurrence in the source code.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI is enough.


vittorioromeo:

eXpl0it3r (approved):

PR #1861, opened 2021-12-03
Add .clang-format file

metaquarx:
## Description

This is a pull request that attemps to match a .clang-format file to the existing style guidelines (at https://www.sfml-dev.org/style.php), using clang-format version 10 (as thats the latest available on `ubuntu-latest`, which may be required depending on how this may be integrated into the CI system).

Incompatibilities:
 * There is no way to have a space after the `operator` keyword
 * There is no way to leave an empty line after a member access modifier, until clang-format 14
 * There is no way to force breaking of constructor initializer lists. Not until the column limit is reached.
 * Include categories are dictated using regex. For the "third party dependencies" category, for now I have only added the ones I was able to pick out, but I have probably missed some. It's also ambiguous whether OS-specific includes fall under this category.

## Tasks

* [X] Tested on Linux 
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI should be fine.


examples/X11/X11.cpp
```diff
@@ -80,51 +80,24 @@ void draw(sf::Window& window, float elapsedTime)
...
-         50,  50, -50,  0, 1, 1,
-
-        -50, -50,  50,  0, 1, 1,
-         50, -50,  50,  0, 1, 1,
-        -50,  50,  50,  0, 1, 1,
-        -50,  50,  50,  0, 1, 1,
-         50, -50,  50,  0, 1, 1,
-         50,  50,  50,  0, 1, 1
-    };
+    static const GLfloat cube[] = {// positions    // colors
```
vittorioromeo, line 83: This seems like a nice place for `//clang-format off`

examples/opengl/OpenGL.cpp
```diff
@@ -8,11 +8,11 @@
...
 
 #ifdef SFML_SYSTEM_IOS
-#include <SFML/Main.hpp>
+    #include <SFML/Main.hpp>
 #endif
 
 #ifndef GL_SRGB8_ALPHA8
-#define GL_SRGB8_ALPHA8 0x8C43
+    #define GL_SRGB8_ALPHA8 0x8C43
 #endif
```
vittorioromeo, line 16: I would definitely revert indentation of preprocessor directives

examples/opengl/OpenGL.cpp
```diff
@@ -117,51 +118,24 @@ int main()
...
-             20,  20, -20,  1, 1,
-
-            -20, -20,  20,  0, 0,
-             20, -20,  20,  1, 0,
-            -20,  20,  20,  0, 1,
-            -20,  20,  20,  0, 1,
-             20, -20,  20,  1, 0,
-             20,  20,  20,  1, 1
-        };
+        static const GLfloat cube[] = {// positions    // texture coordinates
```
vittorioromeo, line 121: Ditto, nice place to manually turn off the formatting

include/SFML/Audio/SoundBuffer.hpp
```diff
@@ -23,17 +23,19 @@
...
+    // Headers
+    ////////////////////////////////////////////////////////////
+    #include <SFML/Audio/Export.hpp>
+
+    #include <SFML/Audio/AlResource.hpp>
+    #include <SFML/System/Time.hpp>
+
+    #include <set>
+    #include <string>
+    #include <vector>
```
vittorioromeo, line 38: Yeah this is very jarring, I have not seen include directives indented ever before. Would strongly suggest to revert

include/SFML/Network/Http.hpp
```diff
@@ -338,11 +335,11 @@ class SFML_NETWORK_API Http : NonCopyable
...
         // Member data
         ////////////////////////////////////////////////////////////
-        FieldTable   m_fields;       //!< Fields of the header
-        Status       m_status;       //!< Status code
+        FieldTable m_fields;         //!< Fields of the header
+        Status m_status;             //!< Status code
         unsigned int m_majorVersion; //!< Major HTTP version
         unsigned int m_minorVersion; //!< Minor HTTP version
-        std::string  m_body;         //!< Body of the response
+        std::string m_body;          //!< Body of the response
```
vittorioromeo, line 342: Can we align data members like before automatically?

metaquarx:
* Unindented PPDs
 * Unformatted the mentioned array definitions, + another similar scenario
 * Found the `catch.hpp` header in extlibs had been accidentally formatted. Entirely my fault there.

Bromeon:
Thank you! This is a great contribution and could potentially avoid lots of bikeshedding regarding formatting 👍 

Some thoughts:
1. It _must_ be enforced via CI, to guarantee that all code in `master` always follows the formatting guidelines.
2. Since it changes almost every file in the entire code base, merging this will create a ton of conflicts. We should probably merge some other open PRs first. It's easy to auto-format later, but very hard to manually fix merge conflicts.

So, maybe this PR should initially only include:
* Configuration files
* Some description on using it offline
* The exceptions in code where formatting needs to be turned off
* Possibly some CI (but can obviously not be enabled before whole code is formatted)

Then we can coordinate the actual formatting with other open PRs, and you could submit a 2nd PR (or changes to this in separate commit). I would also say we limit ourselves to SFML 3.

metaquarx:
Thank you for the feedback!

I do very much agree that it must be enforced by the CI system.

However, that may cause problems when trying to format offline. Unfortunately, clang-format is very much non-backwards compatible. Using the same style file and same input codebase, different versions of clang-format will format files differently. So even if a contributor took the time to clang-format their changes, the CI job might still complain. I'm not entirely sure what the solution to this would be, as we can't exactly expect every contributor to make sure they have the correct clang-format version install.

That said, clang-format can be made to output diffs/patch files, which could be shown within the CI logs.

I'd be happy to revert the actual formatting and only keep the config file and the clang-off comments, to simplify this.

ChrisThrasher:
> Unfortunately, clang-format is very much non-backwards compatible. Using the same style file and same input codebase, different versions of clang-format will format files differently.

I recently dealt with something similar. clang-format-11 and clang-format-7 (what we used CI) were producing different results. I solved this by making the .clang-format file as verbose as possible. The problem was that the preset Webkit style had slight differences between versions but once we explicitly listed every option, there was no more room for clang-format to style them differently between versions. Because your .clang-format file does not use a preset style and is very verbose, I don't suspect we'll have many issues with different clang-format versions producing different results.

metaquarx:
Alright, the diff is now just the clang-format file itself, and the specific instances where it needs to be turned off (for linux, atleast)

I haven't yet added it to CI, as I'm not exactly sure on the workflow for github actions (I come from gitlab). I'll continue to look into that.

Since we aren't formatting the whole repository, if we push to master, contributors' code editors might pick up the `.clang-format` file and automatically format the files that they open and change. This might create lots of diff noise when they eventually PR their changes.

Bromeon:
> Since we aren't formatting the whole repository, if we push to master, contributors' code editors might pick up the .clang-format file and automatically format the files that they open and change. This might create lots of diff noise when they eventually PR their changes.

Good point -- we should have a plan _when_ exactly we want to merge this PR.

Regarding CI, the expectation is not that you implement everything by yourself 🙂 we can assist where needed.
There seem to be already some GitHub actions:
* https://github.com/marketplace/actions/clang-format-check
* https://github.com/marketplace/actions/clang-format-lint
* https://github.com/marketplace/actions/clang-format-github-action
* https://github.com/HorstBaerbel/action-clang-format
* probably more

.clang-format
```diff
@@ -0,0 +1,156 @@
+---
+# Project
+DisableFormat: false
+Language: Cpp
+Standard: c++03
```
reconndev, line 5:
```suggestion
Standard: c++17
```

ChrisThrasher:
>Since it changes almost every file in the entire code base, merging this will create a ton of conflicts. We should probably merge some other open PRs first. It's easy to auto-format later, but very hard to manually fix merge conflicts.

I'm happy to reformat all of my MRs once this lands so don't worry about any outstanding work I have waiting to get merged. 

---

I'd like to propose a mechanism for running clang-format that I've had a lot of success with. 
https://github.com/ChrisThrasher/boids/blob/master/CMakeLists.txt#L18-L20
```cmake
add_custom_target(format
    COMMAND clang-format -i `git ls-files *.cpp *.h`
    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
```
If you create a custom target named `format`, you can run `make format` (or however any other build system is invoked) to run clang-format with a specific set of options defined in source control. This means developers don't have to think hard about how to write the clang-format command. It also provides an abstraction that lets us change the formatting settings without breaking developer workflows. 

`git ls-files` is used to make sure files outside the repo aren't accidentally committed. For SFML, the command would look like `git ls-files *.hpp *.inl *.cpp`. I'd prefer to explicitly ignore the extlibs/ directory but for now no files in there get returned since none match the specified patterns.

This can be enforced in a CI with a few shells commands
```
cmake -B build
cmake --build build --target format
git diff
exit `git status --porcelain | wc -l`
```
although there maybe better ways to do that.

ChrisThrasher:
https://github.com/ChrisThrasher/SFML/commit/9fc65363a29aac3fd0f5e3e9ce41e6112023d0ae

https://github.com/ChrisThrasher/SFML/commit/32a83a40775886bde1c1647b61911357d4fbb101

I added that custom target and tried running clang-format to see what the diff would be. Spoiler: it's pretty big. Hope this helps.

.github/workflows/ci.yml
```diff
@@ -56,3 +56,11 @@ jobs:
...
       shell: bash
       run: cmake --build $GITHUB_WORKSPACE/build --config Release --target install
+
+    - name: Check formatting
+      if: matrix.platform.name == 'Linux Clang'
+      run: |
+        sudo apt-get install clang-format
+        cmake --build $GITHUB_WORKSPACE/build --target format
+        git diff
+        exit `git status --porcelain | wc -l`
```
kimci86, line 66: This could be a separate job, with the benefit of showing success or failure of formatting only, instead or combined build and formatting. Something like (untested):
```
  format:
    name: Check formatting
    runs-on: ubuntu-latest

    steps:
    - name: Checkout Code
      uses: actions/checkout@v2

    - name: Install Dependencies
      run: sudo apt-get update && sudo apt-get install clang-format

    - name: Configure CMake
      run: cmake -S $GITHUB_WORKSPACE -B $GITHUB_WORKSPACE/build -DCMAKE_VERBOSE_MAKEFILE=ON
      
    - name: Run clang-format
      run: |
        cmake --build $GITHUB_WORKSPACE/build --target format
        git diff
        exit `git status --porcelain | wc -l`
```

ChrisThrasher:
Something we still need to solve is crafting a shell command that will find all files ending in .hpp, .inl, and .cpp but not if they're in the extlibs directory. As it stands now if any .hpp, .inl, or .cpp files are added to extlibs, they will get formatted as well which is no good.

CMakeLists.txt
```diff
@@ -540,3 +540,11 @@ string(REGEX REPLACE "/" "\\\\\\\\" NSIS_IMAGE_PATH ${NSIS_IMAGE_PATH})
...
 
 include(CPack)
+
+if(NOT CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
+    return()
+endif()
+
+add_custom_target(format
+    COMMAND clang-format -i `git ls-files *.hpp *.inl *.cpp`
+    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
```
ChrisThrasher, line 550: Something we still need to solve is crafting a shell command that will find all files ending in .hpp, .inl, and .cpp but not if they're in the extlibs directory. As it stands now if any .hpp, .inl, or .cpp files are added to extlibs, they will get formatted as well which is no good.

.github/workflows/ci.yml
metaquarx: Yep was working on that, just wanted to test it out quick. still getting my head around github actions 😅

CMakeLists.txt
ChrisThrasher: There's another problem where `git ls-files` finds files ending in .hpp or .cpp that also contain spaces in the filenames which them breaks when clang-format tries to interpret that filename as two files which don't actually exist. I like using `git ls-files` since it has a better chance of working in a Windows shell but we may have to use Unix shell commands that reduce portability but are more correct.

CMakeLists.txt
```diff
@@ -546,5 +546,7 @@ if(NOT CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
 endif()
 
 add_custom_target(format
-    COMMAND clang-format -i `git ls-files *.hpp *.inl *.cpp`
+    COMMAND "git ls-files -- '*.hpp' '*.inl' '*.cpp' ':!:extlibs/*' -z | xargs -0 clang-format -i"
+    # -z makes git separate lines with null, instead of \n
+    # -0 makes xargs separate by null, not \n, and makes other characters be taken literally
```
ChrisThrasher, line 551:
```
thrasher ~/Projects/sfml$ git ls-files -- '*.hpp' '*.inl' '*.cpp' ':!:extlibs/*' -z | xargs -0 clang-format -i
File name too long
```
Something is going wrong for me when I run this.

CMakeLists.txt
metaquarx: Turns out git has its own wildcard processor, and we can exclude directories within ls-files using `':!:extlibs/*`.

To solve the issue of spaces, pipe to xargs instead of using substitution. Unfortunately that does limit us to *nix, but I don't think that should be a problem, since we only need to run clang-format on one platform.

CMakeLists.txt
metaquarx: oh... lemme check that

ChrisThrasher: Do we need quotes around this command or not? I don't think we do.

kimci86: `-z` needs to be before `--` to be interpreted as an option.

ChrisThrasher: > -z needs to be before -- to be interpreted as an option.

This fixed my problem and now clang-format runs.

metaquarx: > `-z` needs to be before `--` to be interpreted as an option.

ahh yes, nice spot



> Do we need quotes around this command or not? I don't think we do.

im not exactly sure how cmake might handle the different `'`s inside it, I added them just to be safe

ChrisThrasher: It actually won't for me until I remove the quotes so safe to say we don't need them.

metaquarx: Ahh my bad, removed.

ChrisThrasher:
It works for me! Great work, @metaquarx. Feel free to squash my one commit away since it's probably best that we don't add any back-and-forth debugging to the commit history. This PR should probably be 1 commit adding .clang-format, the `format` target, and the CI jobs then a second commit where clang-format is actually run.

metaquarx:
It seems we need to install the full dependencies even if we're only configuring the project (makes sense).

Actually the whole point of the cmake target is to be a cross platform way, so contributors don't need to run clang-format themselves. I'll find a way to bypass xargs first.

CMakeLists.txt
```diff
@@ -540,3 +540,13 @@ string(REGEX REPLACE "/" "\\\\\\\\" NSIS_IMAGE_PATH ${NSIS_IMAGE_PATH})
...
+
+if(NOT CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
+    return()
+endif()
+
+add_custom_target(format
+    COMMAND git ls-files -z -- '*.hpp' '*.inl' '*.cpp' ':!:extlibs/*' | xargs -0 clang-format -i
+    # -z makes git separate lines with null, instead of \n
+    # -0 makes xargs separate by null, not \n, and makes other characters be taken literally
+    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
```
ChrisThrasher, line 552: We might need to quote the WORKING_DIRECTORY since someone could have a source directory that includes spaces.

ChrisThrasher:
```cmake
if(NOT CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
     return()
endif()
```
for anyone not familiar with this pattern I used, the idea is that if SFML is a submodule in someone's project, don't configure this `format` target. Anything that shouldn't get passed onto users including SFML as a submodule can go after this line and they're guaranteed to not get it. `add_subdirectory(test)` should maybe go after this line so that tests and examples and other unessential things aren't configured or built as part of a larger project containing SFML. This speeds up their build times and reduces the possibility of our targets conflicting with user-defined targets.

eXpl0it3r:

kimci86:
Installing dependencies to be able to configure the project but not build it (only format code) seems overkill.
It is simpler to just call clang-format in the CI script without configuring the project, or use an existing action from the GitHub marketplace, but then the "format" target and CI could get out of sync. So, I thought a CMake script for both CI and the "format" target could be nice. I came up with this: https://github.com/kimci86/SFML/commit/215c6898cb294c7ba9a8447db6b41a406ccb6f2d It does not rely on having git installed.

metaquarx:
I don't think having to configure would be a problem, as whoever is contributing code needs to configure to test their changes anyway.

Nevertheless, I do agree it is overkill, and using a cmake script for this is a great idea. I merged your commit into this PR.

ChrisThrasher:
> So, I thought a CMake script for both CI and the "format" target could be nice. I came up with this: [kimci86@215c689](https://github.com/kimci86/SFML/commit/215c6898cb294c7ba9a8447db6b41a406ccb6f2d) It does not rely on having git and xargs installed.

This is really cool! This is a great way to avoid needing `xargs` and avoiding needing to configure the problem just to run formatting which removes the single biggest downside of the approach I proposed while keeping the developer ergonomics of running `make format`. One small simplification you can make is using a relative path to the script to avoid some extra variable dereferencing.
```cmake
add_custom_target(format
    COMMAND ${CMAKE_COMMAND} -P cmake/Scripts/format.cmake
    WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}")
```

eXpl0it3r:

eXpl0it3r:

vittorioromeo:
I've done some more work on this and added many more exceptions for large aligned switches and data arrays. How do we want to proceed? Merge this PR, then format the whole codebase in another PR? @eXpl0it3r @Bromeon @binary1248

vittorioromeo:

Bromeon:
Maybe squash the commits (either one per author, or one in total with `Co-authored-by`).

It might be good to apply the formatting here, so we can iterate if we see some mistakes. Just keep the auto-formatting as a separate commit, and attribute the changes to the original author @metaquarx 🙂

Also, this might cause mass merge conflicts, so let's see if there are some open but near-finished PRs we can merge first. I would appreciate #1979, for example.

ChrisThrasher:
> Maybe squash the commits (either one per author, or one in total with `Co-authored-by`).
> 
> It might be good to apply the formatting here, so we can iterate if we see some mistakes. Just keep the auto-formatting as a separate commit, and attribute the changes to the original author @metaquarx

Agreed. One commit for adding the infrastructure to run clang-format then a second commit which does the massive change. I think I'm co-authored on one commit but I'm fine giving all the credit to @metaquarx. I have some open PRs that I'm more than happy to fix up once this get merged. Luckily all my big PRs recently got merged.

vittorioromeo:
@Bromeon @eXpl0it3r @binary1248 -- I commented out the CI step so we can merge this as is. I will create a new PR afterwards that formats the entire codebase, or maybe multiple PRs for each SFML module.

Bromeon:
> It might be good to apply the formatting here, so we can iterate if we see some mistakes.

What I meant here, is that having the config without seeing it in action is not ideal. Chances are that we overlook something, we deviate too much from the existing style, get too opinionated, etc. It might need some feedback cycles, where we see the application and can adjust the config accordingly.

In other words: applying is the easy part. This PR is about finding a good config and automating it.

metaquarx:
> What I meant here, is that having the config without seeing it in action is not ideal. Chances are that we overlook something, we deviate too much from the existing style, get too opinionated, etc. It might need some feedback cycles, where we see the application and can adjust the config accordingly.
> 
> In other words: applying is the easy part. This PR is about finding a good config and automating it.

@ChrisThrasher ran the format [here](https://github.com/ChrisThrasher/SFML/commit/32a83a40775886bde1c1647b61911357d4fbb101) the only changes since then were related to CI & cmake

ChrisThrasher:
I agree with @Bromeon that we need to run clang-format and see the diff so we can discuss whether or not the current .clang-format config is ideal. I'd rather not merge this PR then immediately start tweaking the .clang-format file we just added. If we keep this PR as two commits, the first can contain changes to the config plus all the other infrastructure we need then the second can be where clang-format gets applied.

metaquarx:
> If we keep this PR as two commits, the first can contain changes to the config plus all the other infrastructure we need then the second can be where clang-format gets applied.
> 
> We also need to run the CI pipeline on the newly formatted code to make sure that the changes generated by clang-format locally match up with what the version of clang-format used in CI expects. We may need to be more explicit about certain settings to make sure various different versions of clang-format are all in agreement. This is tedious but not impossible to do.

Alright, I'll get to that then

Bromeon:
The thing is: as soon as we merge this PR, it will instantly break _all open pull requests_ due to CI enforcing the formatting. There's no point of separate "formatting PR" unless it's merged immediately after, because the current `master` is in a broken state.

So I'd suggest we only enable the `fail_ci_if_error` once we are sure about the changes, *and* we've merged outstanding PRs. Otherwise we can go and fix every single PR, and I could imagine better use of our time.

We can gladly merge this with a purely informational CI, if we're still accepting of config changes in subsequent PRs.

vittorioromeo:
@metaquarx: are you going to update this PR, or should we move towards merging this as-is?

ChrisThrasher:
> @metaquarx: are you going to update this PR, or should we move towards merging this as-is?

I think we still need more discussion about whether or not the current clang-format options are exactly what we want but as long as the formatting CI job remains commented out, this can be merged without any harm. It certainly represents a lot of progress towards the goal of automated formatted even if more later PRs amend the exactly formatting rules before we eventually do the giant reformatting PR and start enforcing formatting.

metaquarx:
Will update, just gotta test each version of clang-format on it, and try and match the newer versions to the older ones. Sorry been ab it busy, will try and get it done soon.

metaquarx:
<sup><sup>Sorry for the delay</sup></sup>

Following some talks in discord, I've upped the minimum version of clang-format to 12. Getting earlier versions to cooperate was pretty much impossible.

I've added a flag to override the executable that the format cmake script uses, incase its named something different (like the one provided by the ubuntu 20.04 `clang-format-12` package, used in CI):
```bash
cmake -DCLANG_FORMAT_EXECUTABLE=clang-format-12 -P cmake/Scripts/format.cmake
```
The same option can be used when configuring the cmake build, so it only needs to be run once when developing. Behind the scenes, it checks the `--version` output to ensure its version 12 or 13 (which both produce the same consistent output).

You can see the format applied as it currently sits [here](https://github.com/metaquarx/SFML/commit/42e48983fce9f5bd8b3dc523c28f961709280905), to be applied whenever ready.

cmake/Scripts/format.cmake
```diff
@@ -0,0 +1,34 @@
...
+
+# Check executable version
+execute_process(COMMAND ${CLANG_FORMAT_EXECUTABLE} --version OUTPUT_VARIABLE CLANG_FORMAT_VERSION)
+string(REGEX MATCH "clang-format version ([0-9]+)" CLANG_FORMAT_VERSION ${CLANG_FORMAT_VERSION})
+unset(CLANG_FORMAT_VERSION)
+if(CMAKE_MATCH_1 EQUAL 12 OR CMAKE_MATCH_1 EQUAL 13)
+    message(STATUS "Using clang-format version ${CMAKE_MATCH_1}")
+else()
+    message(FATAL_ERROR "clang-format version ${CMAKE_MATCH_1} is not supported")
+endif()
```
ChrisThrasher, line 25: Do we want a closed set of version numbers or are we better off future proofing ourselves by requiring a minimum but accepting newer versions? clang-format-15 is already in development so it's safe to say that we're not far off from somebody attempting to use a newer version. I'm of the opinion that we should try to support to those newer versions until we run into issues.

metaquarx: That would be fine by me, I was more concerned about some non-trivial change that could break it without overriding some option.

Bromeon: Are these releases generally backwards-compatible? If yes, then I see no reason to prevent newer versions.

metaquarx: > Are these releases generally backwards-compatible?

While older versions weren't due to non-toggleable bug fixes, 12 and 13 were. I don't suspect this will change in 14/15, but if it does then I'm sure we can catch it and change it. I'll go ahead and change it to `>12`.

metaquarx:
Changed version checking from `12 || 13` to `>=12`

eXpl0it3r:

ChrisThrasher:
`brew` just started shipping clang-format-14. If start with this PR and run `make format`, I get mostly identical results to @metaquarx's `feature/apply-clang-format` branch but with changes to some string literals in src/SFML/Window/EGLCheck.cpp and this one line diff in test/TestUtilities/SystemUtil.hpp.

```
diff --git a/test/TestUtilities/SystemUtil.hpp b/test/TestUtilities/SystemUtil.hpp
index 5d098930..900c5e88 100644
--- a/test/TestUtilities/SystemUtil.hpp
+++ b/test/TestUtilities/SystemUtil.hpp
@@ -84,7 +84,7 @@ public:
     ~TemporaryFile();
 
     // Prevent copies.
-    TemporaryFile(const TemporaryFile&) = delete;
+    TemporaryFile(const TemporaryFile&)            = delete;
     TemporaryFile& operator=(const TemporaryFile&) = delete;
 
     // Return the randomly generated path.
```

metaquarx:
Will attempt to fix as soon as arch linux's clang package upgrades to 14, shouldn't take too long.

eXpl0it3r:
Superseded by #2159

Issue #1754, opened 2021-02-26
Performance: Replace std::map with std::unordered_map


dbechrd:
## SFML is using trees instead of hash tables in many places

I was profiling my SFML app and noticed this:

![image](https://user-images.githubusercontent.com/707367/109253229-2f7d1300-77a4-11eb-84d8-87e54e3fc60a.png)

https://github.com/SFML/SFML/blob/b762276a27efac365968c8330bdb0b5cca44bf52/include/SFML/Graphics/Font.hpp#L332 https://github.com/SFML/SFML/blob/b762276a27efac365968c8330bdb0b5cca44bf52/include/SFML/Graphics/Font.hpp#L392

Both `GlyphTable` and `PageTable` should be `unordered_map`. Ordered iteration is not required by the Font class, and this causes `sf::Font::GetGlyph` to be much slower than is necessary.

There are numerous other instances of `std::map` in the SFML codebase at a quick glance. Most or all of them could likely be replaced with `std::unordered_map` for noticeable performance gains.

LaurentGomila:
Can you do the modification, and show us the results using your profiler?

Bromeon:
Probably needless to say, but make sure you profile in Release mode **and without Debugger running** (these are two independent settings in MSVC).

Also, this should probably become part of the [C++17 solution design](https://github.com/SFML/SFML/wiki/SD%3A-CPP-17-Support). I would even go as far as saying that after C++11, `std::unordered_map` should be the default choice for key-value stores, not `std::map`. That's also what virtually every other programming language on this planet does.

dbechrd:
I spent a few minutes trying to get SFML build, but was getting a bunch of errors. I don't entirely know how to configure it correctly. I generally just use the pre-built binaries for VS 2019 from the "Releases" section on Github so that I don't have to fiddle around with CMake.

You can set up a simple test case yourself by just rendering an sf::Text string to the window with a large enough number of characters for `GetGlyph` to show up in your profiler (100-200 should be plenty, more if you want to stress test it). The profiler in the screenshot above is Intel VTune, which is freely available here (https://software.intel.com/content/www/us/en/develop/tools/oneapi/components/vtune-profiler.html).

[...]

PR #1862, opened 2021-12-03
Change associative containers to their respective 'unordered' version

vittorioromeo:

* [x] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before? (issue #1754)

----

## Description

This PR changes all instances of `std::map` to `std::unordered_map`, and also changes all instances of `std::set` to either `std::unordered_set` or to `std::unordered_map` (when a `std::set<std::pair<T, U>>` was being used). I have done my best to understand the code, and I don't believe that there was any order-specific requirement anywhere, so the behavior should be equivalent to before. Regardless, a thorough code review should be performed.

This PR is related to the issue #1754.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI, thorough review, and test it in an existing project. I will test it out in Open Hexagon.


Bromeon (commented):
Thanks a lot for taking care of this! 🙂 

What we need to consider is that hash and tree maps have different ordering semantics. Blindly replacing one with the other works in most cases, but there may be situations where we implicitly rely on ordering (e.g. for iteration), and the change introduces bugs.

Unless we carefully inspect that this is nowhere the case, the potential performance gains (to be confirmed) may be outweighted by the risk...

src/SFML/Graphics/RenderTextureImplFBO.cpp
```diff
@@ -84,9 +85,9 @@ namespace
         sf::Uint64 contextId = sf::Context::getActiveContextId();
 
         // Destroy active frame buffer objects
-        for (std::set<std::map<sf::Uint64, unsigned int>*>::iterator frameBuffersIter = frameBuffers.begin(); frameBuffersIter != frameBuffers.end(); ++frameBuffersIter)
+        for (std::unordered_set<std::unordered_map<sf::Uint64, unsigned int>*>::iterator frameBuffersIter = frameBuffers.begin(); frameBuffersIter != frameBuffers.end(); ++frameBuffersIter)
```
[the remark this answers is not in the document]
vittorioromeo: Yes, I'd like the change to be separate.

vittorioromeo:
> What we need to consider is that hash and tree maps have different ordering semantics. Blindly replacing one with the other works in most cases, but there may be situations where we implicitly rely on ordering (e.g. for iteration), and the change introduces bugs.

I am aware -- I have checked all the usages and I didn't find anything that seemed order-dependent. I did *not* blindly replace all types. Regardless, I might have missed some cases where order matters or misunderstood some code, so this is why I am asking for a thorough review.

> Unless we carefully inspect that this is nowhere the case, the potential performance gains (to be confirmed) may be outweighted by the risk...

It's not just about performance, it's also about semantics. If we don't need ordering, we shouldn't use ordered containers in the first place.

src/SFML/Graphics/RenderTextureImplFBO.cpp
```diff
@@ -32,23 +32,24 @@
...
+    std::unordered_set<std::unordered_map<sf::Uint64, unsigned int>*> frameBuffers;
 
     // Set to track all stale FBOs
     // This is used to free stale FBOs after their owning
     // RenderTextureImplFBO has already been destroyed
     // An FBO cannot be destroyed until it's containing context
     // becomes active, so the destruction of the RenderTextureImplFBO
     // has to be decoupled from the destruction of the FBOs themselves
-    std::set<std::pair<sf::Uint64, unsigned int> > staleFrameBuffers;
+    std::unordered_map<sf::Uint64, unsigned int> staleFrameBuffers;
```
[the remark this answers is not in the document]
vittorioromeo: I will likely revert this one, it is a bit iffy. When I checked, the `Uint64` is the ID of the context. I assumed that each context can have one framebuffer but I now I see that might not be the right assumption. 

What I could do instead is also change this to:

```cpp
std::unordered_map<sf::Uint64, std::unordered_set<unsigned int>>
```

What do you think?

Bromeon:
> I am aware -- I have checked all the usages and I didn't find anything that seemed order-dependent. I did _not_ blindly replace all types

Ah, that's great to hear 👍 
I'll try to have another look, but with our great test coverage it's really easy to miss some subtle implications 🙈 

> It's not just about performance, it's also about semantics. If we don't need ordering, we shouldn't use ordered containers in the first place.

We should not only argue from a point of view, what would be the correct thing to do when starting a new project nowadays. Most people would agree that `unordered_map/set` is the default choice since C++11 🙂 

Instead, we have an existing system and need to argue _for the change_ and how it improves that system, also taking into account potential regressions. I think we can move forward with this, but from an economic perspective this may mean more bugfixing/maintenance in the future at little added benefit. Probably, consistency/code clarity is worth it.

eXpl0it3r:
> Instead, we have an existing system and need to argue *for the change* and how it improves that system, also taking into account potential regressions. I think we can move forward with this, but from an economic perspective this may mean more bugfixing/maintenance in the future at little added benefit. Probably, consistency/code clarity is worth it.

If your point is focused on considering reasons why something was implemented in the current situation and what potential regressions can be introduced or what impact the change has, then I agree.
But I disagree in having to justify every modernization change of status quo. When there is a commonly agreed on "new way" of doing something, then that should be reason enough to adapt it, rather than having to argue why the current state should be replaced.
I much rather unintentionally break something, than being afraid of unknown potential issues or having to write change proposals for already established practices in the wider C++ community. 😉

Bromeon:
> If your point is focused on considering reasons why something was implemented in the current situation and what potential regressions can be introduced or what impact the change has, then I agree.

Yep, here it's mostly the pragmatic view, aka "don't fix what's not broken". The more we can limit the bug surface in the SFML 2 -> 3 transition, the better. That said, at some point we need to make progress and I don't think this change is very critical, just something to keep in mind.



> But I disagree in having to justify every modernization change of status quo. When there is a commonly agreed on "new way" of doing something, then that should be reason enough to adapt it, rather than having to argue why the current state should be replaced.

Agree, especially for this example here. There are many things which are almost always better in modern C++, such as lambdas vs. one-time function objects, `std::unique_ptr` vs owning pointers, `std::optional` vs. bool-and-output-parameter idiom, standardized `std::thread` etc.

---

_Off-topic:_

Unfortunately though, quite a few newer C++ features are not thought through, or are not "clearly better" but simply a trade-off to existing solutions. Being part of a newer standard doesn't necessarily make older features obsolete. There's also great controversy about some changes in the C++ community (which is natural, given that everything can be achieved in 10 different ways meanwhile).

Some examples -- I don't mean to discuss them individually, just to show that there are points where we should be skeptical:
* "Uniform" initialization, imho one of the bigger fails in modern C++. There's not much uniform in creating 4 near-equivalent but subtly different syntaxes. Plus, it's badly integrated with `std::initializer_list`. It even made into the [Wikipedia list](https://en.wikipedia.org/wiki/Criticism_of_C%2B%2B#Uniform_initialization_syntax) 😬
* Type inference. For iterators etc. sure; but in some cases the type information makes code easier to understand. This is especially true for function signatures. Emplacing also goes a bit in that direction.
* Error handling. Exceptions vs. `std::expected` (not yet standardized) vs. custom solutions around `std::error_code` vs bool/pointer return typ
* [`std::visit()`](https://bitbashing.io/std-visit.html)
* ...

See also [my post in the forum](https://en.sfml-dev.org/forums/index.php?topic=24372.msg173561#msg173561). I don't think those features are _bad_. But, like anything else, we should know when to use them and when not 🙂

vittorioromeo:
@Bromeon: I reverted the change that changed semantics (`staleFrameBuffers`). I did not use an `std::unordered_set` as we would have to introduce a `std::hash` specialization for `std::pair` and I didn't think it was worth it. We can revisit later.

include/SFML/Graphics/Shader.hpp
```diff
@@ -739,8 +739,8 @@ class SFML_GRAPHICS_API Shader : GlResource, NonCopyable
     ////////////////////////////////////////////////////////////
     // Types
     ////////////////////////////////////////////////////////////
-    typedef std::map<int, const Texture*> TextureTable;
-    typedef std::map<std::string, int> UniformTable;
+    typedef std::unordered_map<int, const Texture*> TextureTable;
+    typedef std::unordered_map<std::string, int> UniformTable;
```
Bromeon, line 743: I had another look at this. One thing that looks a bit suspicious is the following code:
https://github.com/SFML/SFML/blob/eef309e870f523363339e02b95d4f6eb2a18b429/src/SFML/Graphics/Shader.cpp#L971-L985

We iterate over the textures with both index and iterator, and assign a uniform based on the index.
I'm not sure if the actual index value matters, but this is an example where `unordered_map` might change semantics.

include/SFML/Network/Http.hpp
```diff
@@ -173,7 +173,7 @@ class SFML_NETWORK_API Http : NonCopyable
         ////////////////////////////////////////////////////////////
         // Types
         ////////////////////////////////////////////////////////////
-        typedef std::map<std::string, std::string> FieldTable;
+        typedef std::unordered_map<std::string, std::string> FieldTable;
```
Bromeon, line 176: This also changes semantics.
https://github.com/SFML/SFML/blob/eef309e870f523363339e02b95d4f6eb2a18b429/src/SFML/Network/Http.cpp#L121-L125

The HTTP request now has a different payload.

What is even worse however is that `unordered_map` has no ordering guarantees whatsoever, so the same SFML code run on different stdlib implementations will produce different HTTP payloads. This might even happen on the same machine in different run-throughs, not sure if `unordered_map` has deterministic ordering or hash DoS protection.

I would revert this change and write a comment why we need `std::map`'s ordering, to make sure it's not refactored in the future.

Bromeon (changes requested):
Thanks!

Had a deeper look, found some potential regressions; see comments.

include/SFML/Network/Http.hpp
vittorioromeo: I realized that this would lead to different ordering, but is that actually a problem? The HTTP payload seems to rely on a key-value based approach, why does the ordering matter? 

I assumed that the associative nature of the payload would mean that it would still work -- am I wrong?

src/SFML/Graphics/RenderTextureImplFBO.hpp
```diff
@@ -136,17 +136,17 @@ class RenderTextureImplFBO : public RenderTextureImpl, GlResource
...
-    unsigned int                   m_colorBuffer;             //!< Optional multisample color buffer attached to the frame buffer
-    unsigned int                   m_width;                   //!< Width of the attachments
-    unsigned int                   m_height;                  //!< Height of the attachments
-    Context*                       m_context;                 //!< Backup OpenGL context, used when none already exist
-    unsigned int                   m_textureId;               //!< The ID of the texture to attach to the FBO
-    bool                           m_multisample;             //!< Whether we have to create a multisample frame buffer as well
-    bool                           m_stencil;                 //!< Whether we have stencil attachment
-    bool                           m_sRgb;                    //!< Whether we need to encode drawn pixels into sRGB color space
+    std::unordered_map<Uint64, unsigned int> m_frameBuffers;            //!< OpenGL frame buffer objects per context
+    std::unordered_map<Uint64, unsigned int> m_multisampleFrameBuffers; //!< Optional per-context OpenGL frame buffer objects with multisample attachments
```
Bromeon, line 140: Not 100% sure if this change is safe, either. Consider:
https://github.com/SFML/SFML/blob/eef309e870f523363339e02b95d4f6eb2a18b429/src/SFML/Graphics/RenderTextureImplFBO.cpp#L87-L102

We have an `erase()` here, but never use the returned iterator. Instead, we keep incrementing the (now erased) iterator in the `for` header. 

This works for associative containers, but does it too for hash ones?

include/SFML/Graphics/Shader.hpp
vittorioromeo: I don't think this would actually change anything in practice, would it? As far as I can tell, the indexes are not really used anywhere else, they're just being used as unique identifiers for the texture units. Am I missing something?

include/SFML/Network/Http.hpp
Bromeon: > I realized that this would lead to different ordering, but is that actually a problem? The HTTP payload seems to rely on a key-value based approach, why does the ordering matter?

The main problem is that it's not deterministic. The same SFML code can lead to different HTTP payloads.

While I agree that _in theory_ neither HTTP nor the server code _should_ care about the order, in practice we don't know what exactly the other end does. Even something simple as logging HTTP requests and finding diffs between them for debugging purposes becomes an issue.

Since there's no advantage in using `unordered_map` here, I'd go with a more user-friendly behavior and potentially save some unnecessary debugging for poor developers 🙂

src/SFML/Graphics/RenderTextureImplFBO.cpp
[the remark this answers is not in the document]
Bromeon: Would this nested data structure give us an actual benefit? If we only need it for lookup of (contextId, frameBufferId) pairs, the existing one is easier.

Also, I'd be careful about having too many not-directly-related changes in this PR; it makes it harder to hunt down bugs (`git bisect`) in case we break something.

include/SFML/Graphics/Shader.hpp
Bromeon: You're probably right, but I haven't touched this code for a while, so I thought I'd highlight it for review. Maybe @binary1248 knows this?

binary1248: @vittorioromeo is right. We just have to assign each texture to the next free texture unit. The ordering shouldn't make a difference. Unless the user changes `m_textures` often there should be no performance penalties either.

vittorioromeo: OK, I am going to keep this change. Thank you for confirming.

include/SFML/Network/Http.hpp
vittorioromeo: I am somewhat convinced. I will revert this particular change and add a comment.

src/SFML/Graphics/RenderTextureImplFBO.hpp
vittorioromeo: Good catch. This is actually undefined behavior -- I will fix it.

vittorioromeo:
@Bromeon: I've addressed your latest comments, please take a look when you have a chance :)

src/SFML/Graphics/RenderTextureImplFBO.cpp
```diff
@@ -84,17 +86,17 @@ namespace
...
+            for (std::unordered_map<sf::Uint64, unsigned int>::iterator iter = (*frameBuffersIter)->begin(); iter != (*frameBuffersIter)->end(); ++iter)
             {
                 if (iter->first == contextId)
                 {
                     GLuint frameBuffer = iter->second;
                     glCheck(GLEXT_glDeleteFramebuffers(1, &frameBuffer));
 
                     // Erase the entry from the RenderTextureImplFBO's map
-                    (*frameBuffersIter)->erase(iter);
+                    iter = (*frameBuffersIter)->erase(iter);
```
Bromeon, line 99: I don't think this is correct; we're skipping an element with the increment in the `for` header.
Shouldn't it be:
```cpp
if (...)
{
    ...
    iter = (*frameBuffersIter)->erase(iter);
}
else
{
    ++iter;
}
```

vittorioromeo: You are right, I didn't think this through. Will fix

vittorioromeo: Actually, I think this has always been fine... there's a `break;` right after the `erase` call. I might just revert to the prior implementation

Bromeon: Ah, true. Too bad [`erase_if()`](https://en.cppreference.com/w/cpp/container/unordered_map/erase_if) is only part of C++20.

Could you maybe add a comment above the inner for loop, something along the lines "remove first frame buffer which matches context ID"?

vittorioromeo: I'd rather revisit this code and do that in a separate PR. I think there's room for improvement here beyond adding a simple comment.

Bromeon: Makes sense, maybe we can also revisit it after the range-based-for / auto iterator PR.

vittorioromeo: Yup, that would be my preference. Let me know if you are happy with the current state of the PR :)

Bromeon (approved):

PR #1863, opened 2021-12-03
Replace SFML multithreading primitives with standard C++ ones

vittorioromeo:

----

## Description

This PR removes the following classes in favour of standard C++ library components and core language features:

- `sf::Thread` (replaced with `std::thread`).
- `sf::Mutex` (replaced with `std::recursive_mutex`).
- `sf::Lock` (replaced with `std::scoped_lock`).
- `sf::ThreadLocal` (replaced with `thread_local`).

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI should be fine. I have also recompiled Open Hexagon with these changes and everything seems fine.


vittorioromeo:

Bromeon (commented):
It's very satisfying to delete so much code, no? 😀 

Most looks good! There might be a subtle difference in behavior from standard and SFML threading primitives. It's hard to test everything now, but we should probably keep this change in mind if we find weird behavior in the next months.

examples/island/CMakeLists.txt
```diff
@@ -11,4 +11,9 @@ sfml_add_example(island GUI_APP
...
 
 # external dependency headers
-target_include_directories(island SYSTEM PRIVATE ${PROJECT_SOURCE_DIR}/examples/island)
\ No newline at end of file
+target_include_directories(island SYSTEM PRIVATE ${PROJECT_SOURCE_DIR}/examples/island)
+
+# link against pthread
+if(SFML_OS_LINUX OR SFML_OS_FREEBSD OR SFML_OS_MACOSX)
+    target_link_libraries(island PRIVATE pthread)
+endif()
```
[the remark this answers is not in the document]
vittorioromeo: Yes, this was required to get the code to link on my platform.

src/SFML/Audio/SoundRecorder.cpp
```diff
@@ -128,7 +128,11 @@ void SoundRecorder::stop()
     if (m_isCapturing)
     {
         m_isCapturing = false;
-        m_thread.wait();
+
+        if (m_thread.has_value() && m_thread->joinable())
+        {
+            m_thread->join();
+        }
```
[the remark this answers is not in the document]
vittorioromeo: > This is used twice, could probably be a separate function awaitCapturingThread().

Will do.

> Also, would it make sense to empty m_thread after join(), since it has now finished running?

There's no real reason to do so, `.emplace` will replace the current one anyway.

src/SFML/Audio/SoundRecorder.cpp
```diff
@@ -203,7 +211,7 @@ bool SoundRecorder::setDevice(const std::string& name)
 
         // Start the capture in a new thread, to avoid blocking the main thread
         m_isCapturing = true;
-        m_thread.launch();
+        m_thread.emplace(&SoundRecorder::record, this);
```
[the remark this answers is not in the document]
vittorioromeo: I don't think it is worth introducing a helper function here.

src/SFML/Window/Android/InputImpl.cpp
```diff
@@ -138,8 +138,8 @@ bool InputImpl::isMouseButtonPressed(Mouse::Button button)
 {
     ALooper_pollAll(0, nullptr, nullptr, nullptr);
 
-    priv::ActivityStates& states = priv::getActivity();
-    Lock lock(states.mutex);
+    ActivityStates& states = getActivity();
```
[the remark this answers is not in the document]
vittorioromeo: It wasn't moved out, we are already in the `priv` namespace so there's no need to qualify it again.

vittorioromeo:
@Bromeon: I addressed your review comments, my replies might have gotten lost due to the force-push

examples/island/CMakeLists.txt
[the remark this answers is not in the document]
ChrisThrasher: Might we be better served with the CMake Threads target? It's a cross platform way to ensure that the correct threading library is linked that lets us avoid writing more platform-specific conditionals. I've used this in a number of projects that are compiled in Windows, various Linux distros, macOS, and more obscure OSes like TI-RTOS and VxWorks and it's always been reliable and correct in my experience.

https://cmake.org/cmake/help/latest/module/FindThreads.html

Its use usually looks like this:
```cmake
find_package(Threads REQUIRED)
...
target_link_libraries(my-target PRIVATE Threads::Threads)
```

src/SFML/Audio/SoundRecorder.cpp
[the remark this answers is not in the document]
Bromeon: It's mostly about the name, "launch capturing thread" is more expressive than "emplace record function", then we don't need extra comments. It would also be symmetrical to `awaitCapturingThread()` (so the `m_isCapturing = true` should be part of it).

vittorioromeo:
@eXpl0it3r, @binary1248: The MinGW CI builds are failing because that version of MinGW does not provide an implementation of the C++ standard synchronization primitives. I suggest using MSYS2/MinGW in CI instead, which works fine out of the box. Alternatively we could use something like https://github.com/meganz/mingw-std-threads but that seems more work and just an outdated process. The `android-armeabi-v7a-api13` also fails as that version of the NDK does not have `<optional>`. It is a fairly old version, I think we should update that as well.

src/SFML/Audio/SoundRecorder.cpp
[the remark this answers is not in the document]
vittorioromeo: I see, I think it is worthwhile doing if we include `m_isCapturing = true`. I will make that change.

include/SFML/Audio/SoundRecorder.hpp
```diff
@@ -285,13 +286,13 @@ class SFML_AUDIO_API SoundRecorder : AlResource
...
     // Member data
     ////////////////////////////////////////////////////////////
-    Thread             m_thread;             //!< Thread running the background recording task
-    std::vector<Int16> m_samples;            //!< Buffer to store captured samples
-    unsigned int       m_sampleRate;         //!< Sample rate
-    Time               m_processingInterval; //!< Time period between calls to onProcessSamples
-    bool               m_isCapturing;        //!< Capturing state
-    std::string        m_deviceName;         //!< Name of the audio capture device
-    unsigned int       m_channelCount;       //!< Number of recording channels
+    std::optional<std::thread> m_thread;             //!< Thread running the background recording task
```
reconndev, line 289: `std::thread` would suffice here.

I assume this `std::optional` was introduced to allow for resetting `m_thread` to a default value but `std::thread::joinable() == false` is equivalent to `std::nullopt` here.

The change for resetting would be:
```cpp
std::thread thread_1;
std::optional<std::thread> thread_2;
// Reset
thread_1 = {};
thread_2 = std::nullopt;
// Check
if (thread_1.joinable()) { /* do work */ }
if (thread_2 && thread_2->joinable()) { /* do work */ }
```

eXpl0it3r:
> MinGW does not provide an implementation of the C++ standard synchronization primitives

The MinGW-w64 MinGW Builds come in two variants. One is with a win32 threading model, that still doesn't properly support `std::thread` and other concurrent types and one with a posix threading model, which does in fact work and also compiles this branch.

@binary1248 can you swap out the versions on your server with the following two:
- [x86_64-8.1.0-release-posix-seh-rt_v6-rev0.7z](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/8.1.0/threads-posix/seh/x86_64-8.1.0-release-posix-seh-rt_v6-rev0.7z/download)
- [i686-8.1.0-release-posix-dwarf-rt_v6-rev0.7z](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/8.1.0/threads-posix/dwarf/i686-8.1.0-release-posix-dwarf-rt_v6-rev0.7z/download)

> I suggest using MSYS2/MinGW

Maybe this has changed, but don't MSYS2 binaries require an additional MSYS2 lib, to workaround the illusion that one is on a Unix-like system?

One version of GCC/Clang that has been getting more popular in recent months is [WinLibs](https://winlibs.com/), which might be worth adding to the pipeline and even try and support both GCC and Clang.

> The android-armeabi-v7a-api13 also fails as that version of the NDK does not have <optional>. It is a fairly old version, I think we should update that as well.

Oh I'm sure @binary1248 is very thrilled to update it (again). 😄

src/SFML/Main/MainAndroid.cpp
```diff
@@ -568,8 +568,10 @@ JNIEXPORT void ANativeActivity_onCreate(ANativeActivity* activity, void* savedSt
     sf::err().rdbuf(&states->logcat);
 
     // Launch the main thread
-    sf::Thread* thread = new sf::Thread(sf::priv::main, states);
-    thread->launch();
+    auto* thread = new std::thread(sf::priv::main, states);
+
+    // Intentionally leak the main thread
+    (void) thread;
```
reconndev, line 574: Didn't you mean this?
```cpp
auto thread = std::thread{sf::priv::main, states};
thread.detach();
```

include/SFML/Audio/SoundRecorder.hpp
Bromeon: Nice. That would also get rid of having two separate null states for `m_thread` (empty optional, non-joinable thread).

src/SFML/Audio/SoundRecorder.cpp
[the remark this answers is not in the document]
Bromeon: > There's no real reason to do so, .emplace will replace the current one anyway.

We wouldn't even need this with the suggestion here: https://github.com/SFML/SFML/pull/1863#discussion_r762568259

include/SFML/Audio/SoundRecorder.hpp
vittorioromeo: Thanks, that should work. It's a nice improvement.

src/SFML/Main/MainAndroid.cpp
vittorioromeo: Ditto, will change it.

examples/island/CMakeLists.txt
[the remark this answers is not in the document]
vittorioromeo: @ChrisThrasher: didn't know about that module, very cool. Will try it out.

binary1248:
> > The android-armeabi-v7a-api13 also fails as that version of the NDK does not have . It is a fairly old version, I think we should update that as well.
> 
> Oh I'm sure @binary1248 is very thrilled to update it (again). 😄

As soon as someone cleans up the Android build process. Our Android build configuration is either (possibly) horribly out of date, or Google decided that the "one true way" to develop for Android is using holy Android Studio, which I am definitely not going to install on a server without a desktop environment.

A nice tutorial/guide documenting the necessary steps from fresh Debian install to finished build-capable command-line environment would be helpful. I haven't found such a thing yet myself.

vittorioromeo:
> A nice tutorial/guide documenting the necessary steps from fresh Debian install to finished build-capable command-line environment would be helpful. I haven't found such a thing yet myself.

Isn't it sufficient to download the latest NDK and build with CMake like this?

```bash
cmake .. -DCMAKE_BUILD_TYPE=Debug -GNinja -DSFML_BUILD_EXAMPLES=1 -DSFML_BUILD_TEST_SUITE=1 -DCMAKE_SYSTEM_NAME=Android -DANDROID=1 -DCMAKE_ANDROID_NDK=/c/android-ndk-r21e -DCMAKE_ANDROID_ARCH_ABI=armeabi-v7a
```

That works for me on MSYS2/MinGW.

examples/island/CMakeLists.txt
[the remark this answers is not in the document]
vittorioromeo: @ChrisThrasher: seems like iOS doesn't like it:

https://ci.sfml-dev.org/#/builders/31/builds/237

reconndev:
> A nice tutorial/guide documenting the necessary steps from fresh Debian install to finished build-capable command-line environment would be helpful. I haven't found such a thing yet myself.

@binary1248  I can help you with that but I don't have Debian around. I managed to build it extremely easily on macOS, though. Full doc step-by-step with examples is [here](https://developer.android.com/ndk/guides/cmake).

The command I used that worked "right off the bat" for macOS looked like this:
```
cd SFML
export NDK="/my/path/to/Android/sdk/ndk/24.0.7856742"
cmake -B build -DCMAKE_TOOLCHAIN_FILE=$NDK/build/cmake/android.toolchain.cmake -DANDROID_ABI=armeabi-v7a -DANDROID_PLATFORM=android-19
```

For me, android-19 was the lowest I could go with.  #1872 fails on android for a strange reason so I could gladly help you with that.

binary1248:
@vittorioromeo CI seems to pass for the current changeset, so... 🤷‍♂️

I will try to get the Android CI build environment into a less hacky state when I have the time to tinker (maybe this weekend).

binary1248:
As long as CI passes nothing really _has to_ be changed with it in order to merge. 😛 Doesn't mean I can't improve it when I have the time.

vittorioromeo:
> As long as CI passes nothing really _has to_ be changed with it in order to merge. 😛 Doesn't mean I can't improve it when I have the time.

OK, great. Will fix the conflicts... eager to get this in! :)

binary1248:
Android CI environment should be updated now.

src/SFML/Audio/SoundStream.cpp
```diff
@@ -498,4 +483,31 @@ void SoundStream::clearQueue()
...
+    // Request the thread to join
+    {
+        std::scoped_lock lock(m_threadMutex);
+        m_isStreaming = false;
+    }
+
+    if (m_thread.joinable())
+    {
+        m_thread.join();
+    }
```
binary1248, line 510: No curly braces for single statement body.

binary1248 (changes requested):

src/SFML/Audio/SoundStream.cpp
vittorioromeo: I don't like this style decision, and I don't think it should be enforced, but I will apply the change as I don't feel very strongly about it and I'd like to see this PR merged :)

ChrisThrasher: This style can have a purpose. Removing these braces signals to the reader that this conditional shall only have a single statement inside of it which discourages others from putting more code in here which in this case is probably not a valid thing to do. It expresses more intent. It's a nice side effect that it reduces line count which lets more code fit on a screen which can aid in reading the code.

I do think this style depends on the existence of an auto formatter so that code like this:

```cpp
if (condition)
    do_thing();
    do_another_thing();
```

gets turn into this

```cpp
if (condition)
    do_thing();
do_another_thing();
```

so that no one is confused about the scope of that conditional. In the absence of an auto formatter (or strict code review to catch these formatting issues) it's harder to justify removing these braces.

vittorioromeo: Counter-points:
- Consistently using braces for every language construct... 
    - prevents mistakes caused by misleading indentation
    - builds muscle memory for "good practice"
    - makes maintenance of code easier (requires less keystrokes to move some code in or out of a `if` body, for example)
    - does not require external tools like autoformatters or linters to verify correctness

But people will never agree about style. :)

eXpl0it3r: Consistency is king, thus as this is the current style, we should adhere to it.
If we want to change the style, let's have a discussion on it separately. 🙂

vittorioromeo: I agree that adhering to the rules is good, I did perform the change as soon as requested. But I am still going to complain about rules I don't like :D 

Of course, discussing on the forum or a separate issue is much better.

src/SFML/Main/MainAndroid.cpp
```diff
@@ -568,8 +568,7 @@ JNIEXPORT void ANativeActivity_onCreate(ANativeActivity* activity, void* savedSt
     sf::err().rdbuf(&states->logcat);
 
     // Launch the main thread
-    sf::Thread* thread = new sf::Thread(sf::priv::main, states);
-    thread->launch();
+    std::thread(sf::priv::main, states).detach();
```
eXpl0it3r, line 571: Did we before really just leak the thread?
Should this really be just detached? 🤔

src/SFML/Window/Context.cpp
```diff
@@ -27,8 +27,8 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Window/Context.hpp>
 #include <SFML/Window/GlContext.hpp>
-#include <SFML/System/ThreadLocalPtr.hpp>
 #include <SFML/System/Err.hpp>
+#include <cstddef>
```
eXpl0it3r, line 31: What's this header included for?

src/SFML/Audio/SoundRecorder.cpp
```diff
@@ -327,4 +323,24 @@ void SoundRecorder::cleanup()
...
+    m_thread = std::thread(&SoundRecorder::record, this);
+}
+
+
+////////////////////////////////////////////////////////////
+void SoundRecorder::awaitCapturingThread()
+{
+    m_isCapturing = false;
+
+    if (m_thread.joinable())
```
eXpl0it3r, line 342: I'm not quite following why we "allow" non-joinable threads...
Are the certain platforms that don't allow for it?

eXpl0it3r (approved):
Just some basic understanding questions...

src/SFML/Audio/SoundRecorder.cpp
vittorioromeo: `std::thread::joinable()` returns whether the thread is currently active or not (suitable to be joined), not whether the platform is capable or joining or not .Calling `std::thread::join()` requires `std::thread::joinable()` to return `true` as a precondition, otherwise an exception will be thrown. 

`sf::Thread::wait()` does not have such a precondition, and looking at `ThreadImpl.cpp` checking if the thread is active performed internally inside `ThreadImpl::wait()`. 

I am explicitly checking for `std::thread::joinable()` here as I am not sure what the state of the thread is, and I want to maintain the exact same semantics as we had before. If I were to remove the joinable check, the semantics would differ from what `sf::Thread::wait()` used to do. Maybe the check is not necessary at all, but that is outside of the scope of this PR, as I just want to replace SFML-specific threading primitives with standard ones, avoiding or minimizing any semantical change to the behavior of SFML.

src/SFML/Main/MainAndroid.cpp
vittorioromeo: > Did we before really just leak the thread?
Yes.

> Should this really be just detached?
Probably not. Outside of the scope of this PR, but definitely something to look into.

src/SFML/Window/Context.cpp
vittorioromeo: Seems like this is a remnant of something else, it is not necessary. I will remove it.

src/SFML/Audio/SoundStream.cpp
reconndev: I support @vittorioromeo claim. I find this to be fragile and inconsistent:

```cpp
if (condition)
    do_thing();
else
    do_another_thing();
```
but suddenly:

```cpp
if (condition)
{
    do_thing();
}
else
{
    // a comment
    do_another_thing();
}
```

It would be better to not allow it at all or enforce this to be a one-liner single expression, non-else, if-statement.

vittorioromeo:
@binary1248: waiting for your approval! :D

binary1248 (approved):

ChrisThrasher:
Great work, @vittorioromeo!

examples/island/CMakeLists.txt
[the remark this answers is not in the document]
ChrisThrasher: https://github.com/SFML/SFML/pull/2704

Seems like whatever caused this to fail before isn't a problem anymore.

PR #1865, opened 2021-12-04
Use 'auto', range-'for', and structured bindings to improve some loops

vittorioromeo:
## Description

Refactoring change. Should not affect behavior at all. The `auto`, range-`for`, and structued bindings features have been used to improve the readability of some loops and usages of iterators.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.

### DIFF
--- a/examples/joystick/Joystick.cpp
+++ b/examples/joystick/Joystick.cpp
@@ -142,15 +142,15 @@ int main()
         object.value.setString("N/A");
     }
 
-    for (Texts::iterator it = texts.begin(); it != texts.end(); ++it)
+    for (auto& [label, joystickObject] : texts)
     {
-        it->second.label.setFont(font);
-        it->second.label.setCharacterSize(14);
-        it->second.label.setFillColor(sf::Color::White);
+        joystickObject.label.setFont(font);
+        joystickObject.label.setCharacterSize(14);
+        joystickObject.label.setFillColor(sf::Color::White);
 
-        it->second.value.setFont(font);
-        it->second.value.setCharacterSize(14);
-        it->second.value.setFillColor(sf::Color::White);
+        joystickObject.value.setFont(font);
+        joystickObject.value.setCharacterSize(14);
+        joystickObject.value.setFillColor(sf::Color::White);
     }
 
     // Update initially displayed joystick values if a joystick is already connected on startup
@@ -187,8 +187,8 @@ int main()
             else if (event.type == sf::Event::JoystickDisconnected)
             {
                 // Reset displayed joystick values to empty
-                for (Texts::iterator it = texts.begin(); it != texts.end(); ++it)
-                    it->second.value.setString("N/A");
+                for (auto& [label, joystickObject] : texts)
+                    joystickObject.value.setString("N/A");
 
                 texts["ID"].label.setString("<Not Connected>");
                 texts["ID"].value.setString("");
@@ -226,10 +226,10 @@ int main()
         window.clear();
 
         // Draw the label-value sf::Text objects
-        for (Texts::const_iterator it = texts.begin(); it != texts.end(); ++it)
+        for (const auto& [label, joystickObject] : texts)
         {
-            window.draw(it->second.label);
-            window.draw(it->second.value);
+            window.draw(joystickObject.label);
+            window.draw(joystickObject.value);
         }
 
         // Display things on screen
--- a/src/SFML/Graphics/Image.cpp
+++ b/src/SFML/Graphics/Image.cpp
@@ -308,8 +308,8 @@ void Image::flipHorizontally()
 
         for (std::size_t y = 0; y < m_size.y; ++y)
         {
-            std::vector<Uint8>::iterator left = m_pixels.begin() + static_cast<std::vector<Uint8>::iterator::difference_type>(y * rowSize);
-            std::vector<Uint8>::iterator right = m_pixels.begin() + static_cast<std::vector<Uint8>::iterator::difference_type>((y + 1) * rowSize - 4);
+            auto left = m_pixels.begin() + static_cast<std::vector<Uint8>::iterator::difference_type>(y * rowSize);
+            auto right = m_pixels.begin() + static_cast<std::vector<Uint8>::iterator::difference_type>((y + 1) * rowSize - 4);
 
             for (std::size_t x = 0; x < m_size.x / 2; ++x)
             {
@@ -328,10 +328,10 @@ void Image::flipVertically()
 {
     if (!m_pixels.empty())
     {
-        std::vector<Uint8>::iterator::difference_type rowSize = static_cast<std::vector<Uint8>::iterator::difference_type>(m_size.x * 4);
+        auto rowSize = static_cast<std::vector<Uint8>::iterator::difference_type>(m_size.x * 4);
 
-        std::vector<Uint8>::iterator top = m_pixels.begin();
-        std::vector<Uint8>::iterator bottom = m_pixels.end() - rowSize;
+        auto top = m_pixels.begin();
+        auto bottom = m_pixels.end() - rowSize;
 
         for (std::size_t y = 0; y < m_size.y / 2; ++y)
         {
--- a/src/SFML/Graphics/RenderTextureImplFBO.cpp
+++ b/src/SFML/Graphics/RenderTextureImplFBO.cpp
@@ -62,18 +62,18 @@ namespace
     {
         sf::Uint64 contextId = sf::Context::getActiveContextId();
 
-        for (std::set<std::pair<sf::Uint64, unsigned int> >::iterator iter = staleFrameBuffers.begin(); iter != staleFrameBuffers.end();)
+        for (auto it = staleFrameBuffers.begin(); it != staleFrameBuffers.end();)
         {
-            if (iter->first == contextId)
+            if (it->first == contextId)
             {
-                GLuint frameBuffer = static_cast<GLuint>(iter->second);
+                GLuint frameBuffer = static_cast<GLuint>(it->second);
                 glCheck(GLEXT_glDeleteFramebuffers(1, &frameBuffer));
 
-                staleFrameBuffers.erase(iter++);
+                staleFrameBuffers.erase(it++);
             }
             else
             {
-                ++iter;
+                ++it;
             }
         }
     }
@@ -86,17 +86,17 @@ namespace
         sf::Uint64 contextId = sf::Context::getActiveContextId();
 
         // Destroy active frame buffer objects
-        for (std::unordered_set<std::unordered_map<sf::Uint64, unsigned int>*>::iterator frameBuffersIter = frameBuffers.begin(); frameBuffersIter != frameBuffers.end(); ++frameBuffersIter)
+        for (auto* frameBuffer : frameBuffers)
         {
-            for (std::unordered_map<sf::Uint64, unsigned int>::iterator iter = (*frameBuffersIter)->begin(); iter != (*frameBuffersIter)->end(); ++iter)
+            for (auto it = frameBuffer->begin(); it != frameBuffer->end(); ++it)
             {
-                if (iter->first == contextId)
+                if (it->first == contextId)
                 {
-                    GLuint frameBuffer = iter->second;
-                    glCheck(GLEXT_glDeleteFramebuffers(1, &frameBuffer));
+                    GLuint frameBufferId = it->second;
+                    glCheck(GLEXT_glDeleteFramebuffers(1, &frameBufferId));
 
                     // Erase the entry from the RenderTextureImplFBO's map
-                    (*frameBuffersIter)->erase(iter);
+                    frameBuffer->erase(it);
 
                     break;
                 }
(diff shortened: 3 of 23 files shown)


vittorioromeo:
SFMLBuildMaster: build this please.

src/SFML/Graphics/RenderTextureImplFBO.cpp
```diff
@@ -160,11 +160,11 @@ RenderTextureImplFBO::~RenderTextureImplFBO()
...
     // Move all frame buffer objects to stale set
-    for (std::map<Uint64, unsigned int>::iterator iter = m_frameBuffers.begin(); iter != m_frameBuffers.end(); ++iter)
-        staleFrameBuffers.emplace(iter->first, iter->second);
+    for (auto& [k, v] : m_frameBuffers)
+        staleFrameBuffers.emplace(k, v);
 
-    for (std::map<Uint64, unsigned int>::iterator iter = m_multisampleFrameBuffers.begin(); iter != m_multisampleFrameBuffers.end(); ++iter)
-        staleFrameBuffers.emplace(iter->first, iter->second);
+    for (auto& [k, v] : m_multisampleFrameBuffers)
+        staleFrameBuffers.emplace(k, v);
```
Bromeon, line 167: Maybe use more expressive variable names, such as `contextId` and `frameBufferId` or so?
That's how they are named [here](https://github.com/SFML/SFML/blob/master/src/SFML/Graphics/RenderTextureImplFBO.cpp#L91-L93), for example.

examples/joystick/Joystick.cpp
```diff
@@ -142,15 +142,15 @@ int main()
         object.value.setString("N/A");
     }
 
-    for (Texts::iterator it = texts.begin(); it != texts.end(); ++it)
+    for (auto& [k, v] : texts)
```
Bromeon, line 145: Maybe change the names to `[label, joystick]` ("k/v" is not really more expressive than "first/second").

Also below.

Bromeon (commented):
Thank you! I like how you kept the type name in the range-based for loop instead of `auto`-ing everything, probably helps the understanding.

src/SFML/Graphics/RenderTextureImplFBO.cpp
```diff
@@ -594,8 +594,8 @@ void RenderTextureImplFBO::updateTexture(unsigned int)
 
         Lock lock(mutex);
 
-        std::map<Uint64, unsigned int>::iterator iter = m_frameBuffers.find(contextId);
-        std::map<Uint64, unsigned int>::iterator multisampleIter = m_multisampleFrameBuffers.find(contextId);
+        auto iter = m_frameBuffers.find(contextId);
+        auto multisampleIter = m_multisampleFrameBuffers.find(contextId);
```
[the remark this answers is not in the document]
vittorioromeo: Yeah, will change. Consistency is good :)

src/SFML/Graphics/RenderTextureImplFBO.cpp
vittorioromeo: Good point, will change.

examples/joystick/Joystick.cpp
vittorioromeo: Ditto, will change.

src/SFML/Network/Http.cpp
```diff
@@ -154,8 +155,7 @@ m_minorVersion(0)
 ////////////////////////////////////////////////////////////
 const std::string& Http::Response::getField(const std::string& field) const
 {
-    FieldTable::const_iterator it = m_fields.find(toLower(field));
-    if (it != m_fields.end())
+    if (auto it = m_fields.find(toLower(field)); it != m_fields.end())
```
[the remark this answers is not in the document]
vittorioromeo: I like the if-with-initializer, I would like to encourage its use. It might not be that useful here, but it's nice to start to be consistent with the feature. I will remove the `else` branch (redundant) as well, which does limit the scope of `it` in a useful way.

vittorioromeo:
@Bromeon: I addressed your comments, please take a look :)

ChrisThrasher:
I found a few more places to use ranged-for:

src/SFML/Window/Unix/WindowImplX11.cpp line 369
src/SFML/Window/Win32/JoystickImpl.cpp lines 468 and 1052

vittorioromeo:
> I found a few more places to use ranged-for:
> 
> src/SFML/Window/Unix/WindowImplX11.cpp line 369 src/SFML/Window/Win32/JoystickImpl.cpp lines 468 and 1052

Thanks! I will handle those in a separate PR (or feel free to do that yourself), I'd like this one to be merged since CI is happy with it :)

ChrisThrasher:
Sorry, I was looking at the master branch, not your branch when I referenced those files! I did find one loop you missed though at src/SFML/Network/Packet.cpp:578. I might submit a PR after yours get merged. Looks like a good first contribution for me.

src/SFML/Graphics/Image.cpp
```diff
@@ -328,10 +328,10 @@ void Image::flipVertically()
 {
     if (!m_pixels.empty())
     {
-        std::vector<Uint8>::iterator::difference_type rowSize = static_cast<std::vector<Uint8>::iterator::difference_type>(m_size.x * 4);
+        auto rowSize = static_cast<std::vector<Uint8>::iterator::difference_type>(m_size.x * 4);
```
Bromeon, line 331: I know it's technically correct, but this will always be `std::ptrdiff_t`, no?
Should we use that to help readability?

src/SFML/Graphics/RenderTextureImplFBO.cpp
```diff
@@ -60,18 +60,18 @@ namespace
     {
         sf::Uint64 contextId = sf::Context::getActiveContextId();
 
-        for (std::set<std::pair<sf::Uint64, unsigned int> >::iterator iter = staleFrameBuffers.begin(); iter != staleFrameBuffers.end();)
+        for(auto it = staleFrameBuffers.begin(); it != staleFrameBuffers.end();)
```
Bromeon, line 63: missing space before (

src/SFML/Graphics/RenderTextureImplFBO.cpp
```diff
@@ -594,13 +594,13 @@ void RenderTextureImplFBO::updateTexture(unsigned int)
 
         Lock lock(mutex);
 
-        std::map<Uint64, unsigned int>::iterator iter = m_frameBuffers.find(contextId);
-        std::map<Uint64, unsigned int>::iterator multisampleIter = m_multisampleFrameBuffers.find(contextId);
+        auto it = m_frameBuffers.find(contextId);
+        auto multisampleIter = m_multisampleFrameBuffers.find(contextId);
```
Bromeon, line 598: Maybe name the two `frameBufferIt` and `multisampleIt`?

src/SFML/Network/Http.cpp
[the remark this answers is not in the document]
Bromeon: I'm not really convinced that:
```cpp
if (auto it = m_fields.find(toLower(field)); it != m_fields.end())
{
     return it->second;
}

static const std::string empty = "";
return empty;
```
is more readable than:
```cpp
auto it = m_fields.find(toLower(field));
if (it != m_fields.end())
{
      return it->second;
}
else
{
     static const std::string empty = "";
     return empty;
}
```
or better, without negation and useless static:
```cpp
auto it = m_fields.find(toLower(field));
if (it == m_fields.end())
     return std::string();
else
     return it->second;
```

In the latter example, it's immediately clear that:
* the condition is to check an iterator against `end()`
* there are two outcomes (if/else) on which it is dispatched

Any other view on this, e.g. @eXpl0it3r? Could of course be more compact with `{}` and `?:`, but this is not necessarily what we're optimizing for 😉 

--- 

Apart from that, we should just return a default-constructed `std::string` instance, a reasonable implementation should not allocate memory.

Bromeon (commented):
Thank you! Some minor comments:

src/SFML/Network/Http.cpp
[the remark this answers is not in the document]
eXpl0it3r: I guess readability for a new language/syntax feature is a bit tricky, as the good-old style of course will be "more readable" as we've been used to it for the past decade or longer.
I like the functionality that if-with-initializer provides, so I would at least try it out for a while before judging its readability.

I'm very much used to early-returns and dropping useless conditional branching (i.e. no `else` block after `return` in an `if` block). So that part would be fine for me.

Is the `static const std::string` really "useless"? As it would ensure that it's only created once, or am I wrong?

One alternative not yet mentioned would be:
```
auto  it = m_fields.find(toLower(field));
return it == m_fields.end() ? std::string() : it->second;
```

I have no strong opinions on any of the styles and don't know how much time we should invest in finding the most pleasing solution. 😄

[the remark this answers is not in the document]
vittorioromeo: I don't think we should focus on the style of this code much. All mentioned alternatives are quite readable. I prefer my own because it limits the scope of the iterator to the `if`, and reducing scope of variables is IMHO a good thing.

Also, we are returning `const std::string&', so we cannot return a temporary.

src/SFML/Graphics/Image.cpp
vittorioromeo: Hmm, I wouldn't risk it to be honest. We support many platforms and I would have to ensure that all of them do indeed use `std::ptrdiff_t`. I'd rather be "technically correct" in this case.

src/SFML/Graphics/RenderTextureImplFBO.cpp
vittorioromeo: Will fix

src/SFML/Graphics/RenderTextureImplFBO.cpp
vittorioromeo: Will fix

vittorioromeo:
@Bromeon: I think this can be merged if you are happy with the latest changes.

src/SFML/Network/Http.cpp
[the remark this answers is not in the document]
Bromeon: Regarding if-with-initializer: it can make sense if we have a bigger scope and we want to keep lifetime of variables as local as possible (this may also be an indicator for a separate method). However, none of the advantages play out here:
* it's not a big scope
* there is no code coming after, which could accidentally use the variable
* it's not an early return, it's the last statement in the function

I agree it's not worth discussing for hours, but maybe precisely for that reason we should address the larger issues before getting into bikeshedding territory 🙂  we can leave this particular change in, but please let's not hunt down `if` statements in SFML now 😆 

---

> Also, we are returning const std::string&, so we cannot return a temporary. 
> We are also not allocating any memory by creating an empty static std::string.

True, I missed that we return a const-reference.

We're not dynamically allocating memory, but we reserve static memory until the end of the program, even if this method is only used once. Plus, we add potential thread unsafety. Both are very minor issues here, but keep in mind that they're trade-offs for another minor issue (copying a `std::string` in a code that likely suffers from much larger performance impact, network latency). 

Anyway, this just as an aside 🙂

src/SFML/Window/OSX/JoystickImpl.cpp
```diff
@@ -356,8 +362,8 @@ JoystickCaps JoystickImpl::getCapabilities() const
     caps.buttonCount = m_buttons.size();
 
     // Axis:
-    for (AxisMap::const_iterator it(m_axis.begin()); it != m_axis.end(); ++it)
-        caps.axes[it->first] = true;
+    for (const auto& [axis, iohidElementRef] : m_axis)
```
Bromeon, line 365: There's no way to ignore a destructuring binding (here `iohidElementRef`), but doesn't this cause unused-variable warnings?

Unfortunately, `std::ignore` only works with `std::tie()`, which cannot be used for initialization. ([some SO discussion](https://stackoverflow.com/q/40673080))

Bromeon (commented):
Thanks again!

Looks good, just 1 comment about potential warning. If that is not an issue, we can go ahead 🙂

src/SFML/Window/OSX/JoystickImpl.cpp
vittorioromeo: There's really no way to ignore a single binding in a destructed binding declaration. This is a (known) annoying limitation.

Bromeon: There was a [SO comment](https://stackoverflow.com/questions/40673080/stdignore-with-structured-bindings#comment76711644_40680014) saying it would generate a warning, but maybe that's not accurate or no longer the case.

I'm happy if this passes CI, no need to create problems where there are none 🙂

vittorioromeo: CI seems happy, let's merge!

vittorioromeo: @Bromeon ping :)

Bromeon (approved):

PR #1867, opened 2021-12-05
Fix improper cast inside packet operator

szymonj99:

----

## Description

Very tiny PR to fix a cast issue for one of the Packet operators.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Before the following code would return output as being 0.
Now it returns the correct value.

```cpp
sf::Packet newPacket;
sf::Int32 input = 1234;
sf::Int32 output;
newPacket << input;
newPacket >> output;
std::cout <<  "input: " << input << " output: " << output << "\n";
```

### DIFF
--- a/src/SFML/Network/Packet.cpp
+++ b/src/SFML/Network/Packet.cpp
@@ -430,7 +430,7 @@ Packet& Packet::operator <<(Uint16 data)
 ////////////////////////////////////////////////////////////
 Packet& Packet::operator <<(Int32 data)
 {
-    Int32 toWrite = static_cast<Int16>(htonl(static_cast<uint32_t>(data)));
+    Int32 toWrite = static_cast<Int32>(htonl(static_cast<uint32_t>(data)));
     append(&toWrite, sizeof(toWrite));
     return *this;
 }


eXpl0it3r:
Regression from: 0468612ac0bbd3095b160b8a23ba5346d8705095

eXpl0it3r (changes requested):
Could you rebase on `2.6.x` and change the base branch to `2.6.x` for the PR?

vittorioromeo:
Could we add a unit test for this? Would be nice to catch these issues automatically.

szymonj99:
> Could we add a unit test for this? Would be nice to catch these issues automatically.

Sure, would be nice.
I'm not able to do it however. I have limited time available meaning others would have to write and maintain the tests.

I'll rebase and push the change in a bit.

reconndev:
> Could we add a unit test for this? Would be nice to catch these issues automatically.

Good suggestion. In fact, I did write a unit test just for that and the code breaks because this change is insufficient. There a duplicate mistake at `src/SFML/Network/Packet.cpp:181` which requires fixing as well.

I created a separate PR with fix and unit tests here: #1871 rebased onto 2.6.x as well.

vittorioromeo:
I am closing this in favour of #1871, which includes the fix provided here plus another one and a unit test. Thanks everyone!

PR #1871, opened 2021-12-05
Fix wrong cast in sf::Packet

reconndev:
A follow-up to #1867

Added unit tests as well which proved #1867 to be insufficient and a breaking change.

----

## Description

`sf::Packet` does not work correctly for `sf::Int32` integer type due to a bug in casting.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/src/SFML/Network/Packet.cpp
+++ b/src/SFML/Network/Packet.cpp
@@ -178,7 +178,7 @@ Packet& Packet::operator >>(Int32& data)
     if (checkSize(sizeof(data)))
     {
         std::memcpy(&data, &m_data[m_readPos], sizeof(data));
-        data = static_cast<Int16>(ntohl(static_cast<uint32_t>(data)));
+        data = static_cast<Int32>(ntohl(static_cast<uint32_t>(data)));
         m_readPos += sizeof(data);
     }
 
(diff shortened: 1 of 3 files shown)


eXpl0it3r (approved):

eXpl0it3r:

binary1248 (approved):

eXpl0it3r:
Thanks for this fix and unit tests! 🙂 
Also thanks to @szymonj99 for reporting the issue initially! 🥳

Bromeon:
Thank you for this PR and adding unit tests! 

Maybe a tip for the future: if the PR bases on another (here @szymonj99), it can be nice to either include their original commit separately, or mention them via `Co-Authored-By:` in the combined commit message. Even if it's just one line of code, the effort lies not in writing that code, but understanding + reproducing the problem, and filing the PR. Thanks! 🙂

reconndev:
You're right. Initially, I was using GitHub.dev for that which is vscode in the browser. It's still kind of limited so I wasn't able to (figure out how to) modify commits there. But I definitely should have done that when working locally.

PR #1007, opened 2015-11-19
Windows: Removed thread affinity changes in sf::Clock

MarioLiebisch:
This should prevent timing issues on Windows XP and earlier with multi-core CPUs and broken BIOS while avoiding unnecessary threading changes for all other versions of Windows.

The whole thing has been [discussed in this thread](http://en.sfml-dev.org/forums/index.php?topic=19375) and in a few other occasions as well I think.

**TL;DR:** This change prevents consecutive clock/timer queries to return time stamps going back in time, while removing unnecessary changes to the current thread, which possibly weren't even applied properly (see discussion) and also could be non-effective, e.g. in case the whole process isn't assigned to the first core.

Only open question for me: What do you think? Would this need a mutex/lock for thread safety? Or would that be overkill?

PR #1872, opened 2021-12-06
Update sf::Clock to C++17

reconndev:
* [x] Code does not modify the logic

----

## Description

Following with similar intentions as #1863.

* Discards platform-dependent implementation files for time-related classes
  * Favors compiler implementations mandated by the C++ Standard
* Updates `sf::Clock` with C++11 `<chrono>` types

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [x] Tested on Android

## How to test this PR?

This refactoring/update tries its best to not modify program logic.


include/SFML/Config.hpp
```diff
@@ -178,25 +180,20 @@ namespace sf
     // we can use them without doing any kind of check
 
     // 8 bits integer types
-    typedef signed   char Int8;
-    typedef unsigned char Uint8;
+    using Int8 = std::int_least8_t;
+    using Uint8 = std::uint_least8_t;
```
Bromeon, line 184: These should not be least, but exact types.
It matters for things like serialization and networking.

include/SFML/System/Clock.hpp
```diff
@@ -74,11 +75,23 @@ class SFML_SYSTEM_API Clock
...
+    ////////////////////////////////////////////////////////////
+    /// \brief Convert clock duration to microseconds
+    ///
+    /// This function acts as a utility for converting clock
+    /// duration type instance into sf::Time
+    ///
+    /// \return Time in microseconds
+    ///
+    ////////////////////////////////////////////////////////////
+    static Time toMicroseconds(ClockImpl::duration duration);
```
Bromeon, line 89: `sf::Time` is not microseconds, but an opaque type which is convertible to microsecond duration. Maybe name the method `durationToTime()` or so?

include/SFML/Config.hpp
```diff
@@ -178,25 +180,20 @@ namespace sf
...
     // 64 bits integer types
-    #if defined(_MSC_VER)
-        typedef signed   __int64 Int64;
-        typedef unsigned __int64 Uint64;
-    #else
-        typedef signed   long long Int64;
-        typedef unsigned long long Uint64;
-    #endif
+    using Int64 = std::int_least64_t;
+    using Uint64 = std::uint_least64_t;
```
eXpl0it3r, line 196: I feel like this "core" change, would need to be discussed in a separate PR.
One might even go as far as argue, that the custom `sf::*` types are no longer needed.

eXpl0it3r:
See also the [forum discussion](https://en.sfml-dev.org/forums/index.php?topic=21639.0) on this topic

include/SFML/System/Clock.hpp
```diff
@@ -74,11 +75,23 @@ class SFML_SYSTEM_API Clock
     Time restart();
 
 private:
+    using ClockImpl = std::chrono::system_clock;
```
Bromeon, line 78: `sf::Clock` is not a date/time API, but a monotonic high-precision clock. As such, `system_clock` is the wrong choice.

From https://en.cppreference.com/w/cpp/chrono/system_clock:

> Class std::chrono::system_clock represents the system-wide real time wall clock.
> 
> It may not be monotonic: on most systems, the system time can be adjusted at any moment. It is the only C++ clock that has the ability to map its time points to C-style time. 

We should probably use [`high_resolution_clock`](https://en.cppreference.com/w/cpp/chrono/high_resolution_clock) instead. It could be that they end up using the same implementation, but this is the platform's choice.

Bromeon (changes requested):
Thanks a lot for your contribution! 🙂 

Since clocks are such a central thing and there were issues in the past with multiple stdlib implementations not using high-precision clocks ([g++](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=63400), [MSVC](https://stackoverflow.com/q/16299029)), we should be _really_ sure we're not introducing bugs with this change.

What do you think about having some unit-tests? I know it's a bit hard to achieve with clocks in a somewhat deterministic way. The alternative is that we conduct some manual tests, e.g. by running example projects which make heavy use of clocks.

include/SFML/Config.hpp
reconndev: In the description, I provided an explanation of why it's impossible to achieve major cross-compatibility and exact bit width. I could do that but I didn't know if that's acceptable for SFML to drop the rare cases. Let me know

Bromeon: I'd rather have a compile error because of unsupported `std::int16_t`, than subtle bugs at runtime because a wider type is used. The point of these typedefs is exactly to have a precise size -- when SFML doesn't care about the size, it can use `int`, `short` etc.

Do you know platforms that we support, but which don't have `std::[u]intN_t` types?

include/SFML/System/Clock.hpp
reconndev: A good thing to note about `high_resolution_clock` which leans me more towards `steady_clock`:

>The high_resolution_clock is not implemented consistently across different standard library implementations, and its use should be avoided. It is often just an alias for std::chrono::steady_clock or std::chrono::system_clock, but which one it is depends on the library or configuration. When it is a system_clock, it is not monotonic (e.g., the time can go backwards). For example, for gcc's libstdc++ it is system_clock, for MSVC it is steady_clock, and for clang's libc++ it depends on configuration.

>Generally one should just use std::chrono::steady_clock or std::chrono::system_clock directly instead of std::chrono::high_resolution_clock: use steady_clock for duration measurements, and system_clock for wall-clock time.

Specifically:
> use steady_clock for duration measurements

That seems to be what sf::Clock needs to represent. So how about `std::chrono::steady_clock`?

include/SFML/Config.hpp
reconndev: Good point. Supported platforms will be fine. I removed the definitions per @eXpl0it3r's request.

include/SFML/System/Clock.hpp
Bromeon: I'll answer in the main thread, since it concerns the PR as a whole.

Bromeon:
Found [this](https://gamedev.stackexchange.com/a/131095) as a basic explanation of `steady_clock` and `high_performance_clock`.
However, `steady_clock` seems to also have issues:
* [stops while in background](https://github.com/android/ndk/issues/1101) (Android)
* [bad resolution](https://stackoverflow.com/questions/29551126/does-steady-clock-have-only-10ms-resolution-on-windows-cygwin) (Windows/Cygwin)
* [general uncertainity, suggestion of platform-specific workarounds](https://www.reddit.com/r/cpp/comments/fmosnb/how_can_anyone_use_stdchronosteady_clock_portably)

Why does C++ always make portability so hard? It's now a full decade after Chrono was standardized, yet we're in the situation where we need to manually evaluate different options for something like measuring durations. Migrating to Chrono at this point feels more like a downgrade to me.

I'm not generally opposed, but we should double-check the implementation on each of our systems _very carefully_. Keep in mind that we spent quite some effort over the years to iron out the implementation for exotic cases (https://github.com/SFML/SFML/issues/1765, https://github.com/SFML/SFML/pull/1258, https://github.com/SFML/SFML/pull/1007, ...). I don't want to throw that over board just so we can say we use Chrono. In the end we need to provide the best possible implementation for the user, and I wish we could just have that confidence in the standard library 😬

include/SFML/System/Clock.hpp
```diff
@@ -74,11 +76,26 @@ class SFML_SYSTEM_API Clock
...
+    ////////////////////////////////////////////////////////////
+    /// \brief Convert clock duration to microseconds
+    ///
+    /// This function acts as a utility for converting clock
+    /// duration type instance into sf::Time
+    ///
+    /// \return Time in microseconds
+    ///
+    ////////////////////////////////////////////////////////////
+    static Time durationToTime(ClockImpl::duration duration);
```
vittorioromeo, line 93:
```suggestion
    [[nodiscard]] static Time durationToTime(ClockImpl::duration duration);
```

reconndev:
I understand your view. I did some digging as well and it's been a rollercoaster of emotions but I think I feel better with Chrono. I added static checks to assert precision and a monotonic clock.

I believe the problem is not with C++ here in terms of portability. C++20 included even more clocks but the issue is because of the variety of system architectures. It's almost impossible to standardize and maintain everything with one simple solution. It's a step in a more general direction nevertheless.

However, let me try to defend the changes this PR introduces using tasks.
Linux:
  [x] Previously: `CLOCK_MONOTONIC`. Chrono clocks with `clock::is_steady == true` use monotonic too
  [x] Can use `std::chrono::steady_clock`
Android:
  [x] Android-specific clock implementation does not exist in SFML so the problem you found still persist
  [ ] I can try to fix it in this PR but I'm not sure I will be able to fully test
macOS:
  [ ] Need to verify the code is better or the same - will edit
iOS:
  [x] Uses `CLOCK_MONOTONIC` which updates even when the system is suspended (opposite of what Android, unfortunately, has - hence the bug)
  [x] Can use `std::chrono::steady_clock`
Windows:
  [ ] Need to verify the code is better or the same - will edit

reconndev:
In conclusion to my research, I believe to have carefully tested and checked (almost) all supported platforms and major compiler implementations of std::chrono clocks. The results favor `steady_clock` but a more general solution was provided (taking `high_resolution_clock` into account if the clock is monotonic).

I did not check iOS and Windows XP. Here's the summary:
* I did not implement unit tests (this is close to impossible) but I did include compile-time checks which will ensure at least microsecond precision and clock steadiness
* **iOS** seems to not have the system suspension issue as Apple decided to implement it that way
* **macOS** uses `CLOCK_MONOTONIC` for `steady_clock`
* **Android/Linux** reverted the change to make `CLOCK_MONOTONIC` suspend-aware due to backward compatibility with `systemd`. Thus, users would have to use `CLOCK_BOOTTIME` to extend capabilities
* **Windows** 10 was checked with success, however, Windows XP is out of my range
* `high_resolution_clock` is an alias on almost every compiler. Reason: it's the more precise of `system_clock` or `steady_clock` until architectures provide clocks with higher precision than nanoseconds

Please, consider reviewing this PR again with that in mind. Thank you.

CMakeLists.txt
```diff
@@ -110,6 +110,11 @@ endif()
 
 # Android options
 if(SFML_OS_ANDROID)
+    sfml_set_option(SFML_ANDROID_USE_SUSPEND_AWARE_CLOCK FALSE BOOL "TRUE to use the suspend-aware sf::Clock implementation, FALSE to default to another available monotonic clock")
```
Bromeon, line 113: Suspend-aware means it continues running or stops during activity suspensions? Maybe mention this instead, the term "suspend-aware" is already in the variable name 🙂

I don't know what a good default here is, depends a bit on how the developers use the clock. On Desktop environments, the clock keeps running even if the user loses window focus; but then again mobile has so many different paradigms.

include/SFML/System/Clock.hpp
```diff
@@ -30,10 +30,34 @@
...
+
+struct MostSuitableClockDetector
+{
+#if defined(SFML_SYSTEM_ANDROID) && defined(SFML_ANDROID_USE_SUSPEND_AWARE_CLOCK)
+    using type = SuspendAwareClock;
+#else
+    using type = std::conditional_t<
+        std::chrono::high_resolution_clock::is_steady,
+        std::chrono::high_resolution_clock,
+        std::chrono::steady_clock>;
```
Bromeon, line 55: Could you maybe write a comment (and if it helps, a link) as to why this is done? Would be nice to preserve your valuable research.

src/SFML/System/Android/SuspendAwareClock.hpp
```diff
@@ -22,40 +22,51 @@
...
 ///
 ////////////////////////////////////////////////////////////
-class ClockImpl
+class SuspendAwareClock
 {
 public:
+    using duration = std::chrono::nanoseconds;
+    using rep = duration::rep;
+    using period = duration::period;
+    using time_point = std::chrono::time_point<SuspendAwareClock, duration>;
```
Bromeon, line 61: These types define the [`TrivialClock`](https://en.cppreference.com/w/cpp/named_req/TrivialClock) concept and is expected for standard-like clocks, correct?

Could you maybe add a comment to express this? Otherwise people might wonder about the naming convention.

include/SFML/System/Clock.hpp
```diff
@@ -30,10 +30,34 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/System/Export.hpp>
 #include <SFML/System/Time.hpp>
+#include <chrono>
+#include <ratio>
```
Bromeon, line 34: The `<ratio>` header (and the `static_assert`s) could probably be moved to the .cpp file, to reduce header dependencies.

src/SFML/System/Android/SuspendAwareClock.cpp
```diff
@@ -22,31 +22,24 @@
...
-
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
-#include <SFML/Config.hpp>
-#include <SFML/System/Time.hpp>
+#include <SFML/System/Android/SuspendAwareClock.hpp>
+#include <ctime>
 
+#ifdef CLOCK_BOOTTIME
```
Bromeon, line 31: I stumbled upon [this mailing list entry](https://lwn.net/Articles/428176/), did you find proper documentation about it?

Also, isn't it an error if `CLOCK_BOOTTIME` is not defined? Maybe we should directly use `#error`, because now the user would get some compile errors about missing symbols instead (unless I'm missing something).

src/SFML/System/Android/SuspendAwareClock.cpp
```diff
@@ -22,31 +22,24 @@
...
-////////////////////////////////////////////////////////////
-void sleepImpl(Time time);
-
-} // namespace priv
 
-} // namespace sf
+SuspendAwareClock::time_point SuspendAwareClock::now() noexcept
+{
+    timespec ts;
+    clock_gettime(CLOCK_BOOTTIME, &ts);
```
Bromeon, line 39: Qualify those two with `std::` (they're part of `<ctime>`, which _officially_ is part of `std` in contrast to `<time.h>`).

Bromeon (commented):
@reconndev This is absolutely awesome. Thank you very much for your extensive research, testing, and dedication to stay true to the existing behavior -- all that in such a short time ❤️ 

I think Windows XP alone shouldn't be an obstacle to progress; we can adjust the implementation if we find users need it.

Even with some platform-dependent selection, this PR cuts down on a lot of code and OS-specific APIs. I probably won't have the chance to test the changes myself in the near future; maybe someone else can, for dedicated platforms? Was there a test project/code you used?

Otherwise, it's likely that some of the corner cases will only be detected once people start using SFML 3, and for this we need to merge changes eventually. In that sense, you have my green light (after the few comments), but I'd appreciate the input of other SFML contributors/members here.

CMakeLists.txt
reconndev: It takes suspension time into account once the system goes into sleep. I'll think of a more descriptive doc here 👍 

Personally, a good default is the one consistent across platforms. Any monotonic clock with a microsecond precision is good enough. Suspend-awareness feels more like a gimmick. Android should provide a separate interface for observers working in the background. I am not an Android developer so I don't know but I would expect such a thing.

src/SFML/System/Android/SuspendAwareClock.hpp
reconndev: Correct, `TrivialClock` to be specific but more importantly it will pass the check for a regular `Clock` concept [here](https://en.cppreference.com/w/cpp/chrono/is_clock).

Yes, the naming inconsistency might startle some. I'll add comments.

src/SFML/System/Android/SuspendAwareClock.cpp
reconndev: Linux man pages describe this clock well. https://linux.die.net/man/2/clock_gettime

I have seen mailing lists about trying to merge `CLOCK_MONOTONIC` and `CLOCK_BOOTTIME` (which was accepted even by Linus Torvalds) but later on, it proved to be a mistake as various components had already been unprepared for suspend-awareness which caused bugs (e.g. `systemd`) and so [they reverted it](https://www.spinics.net/lists/linux-tip-commits/msg43709.html).

> Also, isn't it an error if CLOCK_BOOTTIME is not defined? 

You're probably right. I will update.

include/SFML/System/Clock.hpp
reconndev: `<ratio>` is already part of `<chrono>`. I added that for clarity.

I couldn't do it because `ClockImpl` is a private member of the class so `sf::Clock::ClockImpl` is hidden outside of the class. I could make it work by putting it into the constructor but I didn't want to do that because then every other constructor in the future would have to include that too.

There is one other reason though. I don't know if that's the rule but I think `static_assert`'s should reside in header files when affecting the class as a whole. I was just able to code a mini example that would suppress static assertion when put in the source file.

src/SFML/System/Android/SuspendAwareClock.cpp
reconndev: Not possible. `timespec` and `clock_gettime` are not part of the Standard. Those are Linux-specific.

I chose `<ctime>` because the header existence is guaranteed by the Standard (e.g. older OSX would include `mach_time.h`).

include/SFML/System/Clock.hpp
reconndev: Added documentation to the class

reconndev:
> Was there a test project/code you used?

I was only able to write simple applications, for clock testing purposes, for Android and macOS during that time.

I would appreciate and encourage any feedback whether it's testing it on a larger project or other members' review.

eXpl0it3r:

src/SFML/System/Android/SuspendAwareClock.cpp
Bromeon: C++17 added [`timespec`](https://en.cppreference.com/w/cpp/chrono/c/timespec), but it looks like `clock_gettime()` is really POSIX.

In that case, it's fine if you leave them out of the namespace 👍

include/SFML/System/Clock.hpp
Bromeon: Makes sense 👍

CMakeLists.txt
```diff
@@ -110,6 +110,11 @@ endif()
 
 # Android options
 if(SFML_OS_ANDROID)
+    sfml_set_option(SFML_ANDROID_USE_SUSPEND_AWARE_CLOCK FALSE BOOL "TRUE to use an sf::Clock implementation which takes system sleep time into account, FALSE to default to another available monotonic clock")
```
Bromeon, line 113: "To take into account" is almost a synonym for "to be aware". What is still not unambiguously clear, is whether that means that the clock _stops_ during system sleep time (I would use the term "suspend" here), or keeps running. I would really be as explicit as possible. Maybe something like:

> TRUE to use an sf::Clock implementation which pauses while the activity is suspended, FALSE for a monotonic clock that keeps running.

or the other way around, if I got this wrong.

include/SFML/System/Clock.hpp
```diff
@@ -74,11 +123,28 @@ class SFML_SYSTEM_API Clock
...
+    ////////////////////////////////////////////////////////////
+    /// \brief Convert clock duration to microseconds
+    ///
+    /// This function acts as a utility for converting clock
+    /// duration type instance into sf::Time
+    ///
+    /// \return Time in microseconds
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] static Time durationToTime(ClockImpl::duration duration);
```
Bromeon, line 142: (Bikeshed) I don't think this is a good `[[nodiscard]]` candidate (I think I missed/misjudged this before).

Yes, the return value must be used, but so do 95% of all functions which return something. We don't want that attribute on functions like `Sprite::getPosition()` or `Sound::isRelativeToListener()`, even if they'd _technically_ qualify. This would litter our APIs without real-world benefit, and drown actual cases where `[[nodiscard]]` has a deeper meaning.

`[[nodiscard]]` is primarily useful to prevent mistakes in non-obvious cases where a return value could be accidentally ignored. For example:
```cpp
sf::Image image;
image.loadFromFile(...);
// WARN: didn't check 'bool' return type!
```
Same for methods which could be interpreted in a different way due to English language ambiguity:
```cpp
std::vector<T> container = ...;

container.empty(); // empty the container
// WARN: this is a noun, not a verb!
```

Here, we have a conversion function with a name `durationToTime()` -- it's 100% obvious that we get a `duration` and return a `Time`.

src/SFML/System/Android/SuspendAwareClock.cpp
reconndev: Oh, my bad then. I see an error squiggle at `std::timespec` but it compiles. Very weird.

Bromeon: It _probably_ works because the `std` symbol is simply pulled in with `using` from the global one, but maybe it's safer to keep both in the global namespace. You can qualify it with `::timespec` to be explicit...

include/SFML/System/Clock.hpp
reconndev: Exactly, that was my thinking but I changed it anyway.

I still would like to push this further and completely propose the replacement of `sf::Time` with Chrono as discussed [here](https://en.sfml-dev.org/forums/index.php?topic=21639.15).  Then `durationToTime` would be removed.

src/SFML/System/Android/SuspendAwareClock.cpp
reconndev: It was because I installed clang-tidy which pulled Clang-13 into PATH. This must be a clang-13 bug with M1. Now, it's alright

reconndev: Android fails too. Compilers supporting C++17 but lacking `std::timespec`. I'll revert to global namespace.

reconndev:
I updated the branch to the latest master via rebasing. 

@Bromeon I applied your review suggestions to the code. Unfortunately, the Android CI compiler does not fully support C++17 (`std::timespec` not in <ctime>). Updated docs, though.

@eXpl0it3r can we start the rest of the CI checks? Thanks

Bromeon (approved):

eXpl0it3r:

eXpl0it3r:
Can you squash the commits?

eXpl0it3r:

Bromeon:
One failed CI job:
```
make: *** [/usr/local/share/android-sdk-linux/ndk/21.4.7075529/build/core/build-binary.mk:478: /usr/local/share/buildslave/android-armeabi-v7a-api13/build/examples/android/app/build/intermediates/ndkBuild/release/obj/local/armeabi-v7a/objs-debug/sfml-example/main.o] Error 1
:app:externalNativeBuildRelease FAILED
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':app:externalNativeBuildRelease'.
> Build command failed.
  Error while executing process /usr/local/share/android-sdk-linux/ndk/21.4.7075529/ndk-build with arguments {NDK_PROJECT_PATH=null APP_BUILD_SCRIPT=/usr/local/share/buildslave/android-armeabi-v7a-api13/build/examples/android/app/src/main/jni/Android.mk NDK_APPLICATION_MK=/usr/local/share/buildslave/android-armeabi-v7a-api13/build/examples/android/app/src/main/jni/Application.mk APP_ABI=armeabi-v7a NDK_ALL_ABIS=armeabi-v7a NDK_DEBUG=0 APP_PLATFORM=android-16 NDK_OUT=/usr/local/share/buildslave/android-armeabi-v7a-api13/build/examples/android/app/build/intermediates/ndkBuild/release/obj NDK_LIBS_OUT=/usr/local/share/buildslave/android-armeabi-v7a-api13/build/examples/android/app/build/intermediates/ndkBuild/release/lib /usr/local/share/buildslave/android-armeabi-v7a-api13/build/examples/android/app/build/intermediates/ndkBuild/release/obj/local/armeabi-v7a/libsfml-example.so}
  [armeabi-v7a] Compile++ thumb: sfml-example <= main.cpp
  In file included from /usr/local/share/buildslave/android-armeabi-v7a-api13/build/examples/android/app/src/main/jni/main.cpp:1:
  In file included from /usr/local/share/buildslave/android-armeabi-v7a-api13/install/sfml/include/SFML/System.hpp:33:
  /usr/local/share/buildslave/android-armeabi-v7a-api13/install/sfml/include/SFML/System/Clock.hpp:38:10: fatal error: 'SFML/System/Android/SuspendAwareClock.hpp' file not found
  #include <SFML/System/Android/SuspendAwareClock.hpp>
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  1 error generated.
  make: *** [/usr/local/share/android-sdk-linux/ndk/21.4.7075529/build/core/build-binary.mk:478: /usr/local/share/buildslave/android-armeabi-v7a-api13/build/examples/android/app/build/intermediates/ndkBuild/release/obj/local/armeabi-v7a/objs-debug/sfml-example/main.o] Error 1
```

Other than that, good from my side ✔️

reconndev:
I have no idea why this android build fails, honestly. The files are correctly included in CMakeLists.txt. It cross-compiles for Android for me. Must be some kind of a build hiccup.

eXpl0it3r:

include/SFML/System/Clock.hpp
```diff
@@ -30,10 +30,59 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/System/Export.hpp>
 #include <SFML/System/Time.hpp>
+#include <chrono>
+#include <ratio>
+#include <type_traits>
+
+#ifdef SFML_SYSTEM_ANDROID
+#include <SFML/System/Android/SuspendAwareClock.hpp>
+#endif
```
eXpl0it3r, line 39: The reason the Android build is failing, is because you're including a "private" header in public header. The SuspendAwareClock is located in the `src` directory, which is not installed with the SFML distribution, yet the System header in the public `include` directory, tries to include it.
Thus, when building the Android example and using the public SFML headers, it fails, as the private header can't be found.

The preferred solution would be to use the pImpl pattern to hide the Android implementation.

include/SFML/System/Clock.hpp
```diff
@@ -30,10 +30,59 @@
...
+struct MostSuitableClockDetector
+{
+#if defined(SFML_SYSTEM_ANDROID) && defined(SFML_ANDROID_USE_SUSPEND_AWARE_CLOCK)
+    using type = SuspendAwareClock;
+#else
+    using type = std::conditional_t<
+        std::chrono::high_resolution_clock::is_steady,
+        std::chrono::high_resolution_clock,
+        std::chrono::steady_clock>;
+#endif
```
vittorioromeo, line 81: A bit simpler:

```suggestion
#if defined(SFML_SYSTEM_ANDROID) && defined(SFML_ANDROID_USE_SUSPEND_AWARE_CLOCK)
    using MostSuitableClock = SuspendAwareClock;
#else
    using MostSuitableClock = std::conditional_t<
        std::chrono::high_resolution_clock::is_steady,
        std::chrono::high_resolution_clock,
        std::chrono::steady_clock>;
#endif
```

include/SFML/System/Clock.hpp
```diff
@@ -74,11 +123,28 @@ class SFML_SYSTEM_API Clock
     Time restart();
 
 private:
+    using ClockImpl = typename priv::MostSuitableClockDetector::type;
```
vittorioromeo, line 126:
```suggestion
    using ClockImpl = priv::MostSuitableClock;
```

vittorioromeo (changes requested):

include/SFML/System/Clock.hpp
reconndev: Pimpl would be good at hiding implementation details at run-time. However, the choice of the clock is done at compile-time.

It would be very inelegant to move it to run-time unless you have a better solution to that.

eXpl0it3r:

reconndev:
FYI, the compiler for android-armeabi-v7a seems not to be fully c++17-capable. Might wanna upgrade to C++17 full library support. I'll change the code, meanwhile.
```cpp
error: no template named 'ratio_less_equal_v' in namespace 'std'; did you mean 'ratio_less_equal'?
      static_assert(std::ratio_less_equal_v<ClockImpl::period, std::micro>,
                    ~~~~~^~~~~~~~~~~~~~~~~~
                         ratio_less_equal
  /opt/android-sdk/ndk/21.4.7075529/sources/cxx-stl/llvm-libc++/include/ratio:484:29: note: 'ratio_less_equal' declared here
```

binary1248:
@reconndev The NDK version currently used by CI is 21.4 which does support C++17. If you look closely at the CI logs, the failure occurs when building the Android example as opposed to the library itself. Indeed, when the switch was made over to C++17, adjusting the flags with which the Android example is built was forgotten.

The relevant line would be: https://github.com/SFML/SFML/blob/master/examples/android/app/build.gradle#L14

I'm not an Android expert, but adding an uncommented directive to add `-std=c++17` to the compiler flags seems to have did the trick for me.

eXpl0it3r:

include/SFML/System/Clock.hpp
vittorioromeo: @reconndev: I don't understand, you could do something like this:

```cpp
#ifdef SFML_SYSTEM_ANDROID
class SuspendAwareClock;
#endif

// ...

#if defined(SFML_SYSTEM_ANDROID) && defined(SFML_ANDROID_USE_SUSPEND_AWARE_CLOCK)
    using MostSuitableClock = SuspendAwareClock;
#else
    using MostSuitableClock = std::conditional_t<
        std::chrono::high_resolution_clock::is_steady,
        std::chrono::high_resolution_clock,
        std::chrono::steady_clock>;
#endif

// ...

using ClockImpl = std::unique_ptr<priv::MostSuitableClock>;
```

You would have to expose inner type aliases such as `time_point` though, so I'm not sure how worth it this change would be.

include/SFML/System/Clock.hpp
vittorioromeo: @reconndev: this was resolved, but I don't see the change applied.

include/SFML/System/Clock.hpp
reconndev: You have to write more code to see the problem:
https://godbolt.org/z/fGEEsojfP

vittorioromeo: That's what I meant with *"You would have to expose inner type aliases such as time_point though, so I'm not sure how worth it this change would be."*, so I don't think it's worth performing this change.

include/SFML/System/Clock.hpp
vittorioromeo: ^

vittorioromeo (changes requested):
LGTM except one small unresolved comment

include/SFML/System/Clock.hpp
reconndev: Must have been reverted accidentally. Done

eXpl0it3r:

eXpl0it3r:

eXpl0it3r (approved):

vittorioromeo (approved):

eXpl0it3r:
Thanks a lot for this update and all the research you've done! 🎉

PR #1873, opened 2021-12-07
Fix Vulkan example exceptions

Si1ver:
## Description

While testing another change I faced two exceptions in Vulkan example application.

First one was on application startup in `void setupInstance()` when trying to get `&validationLayers[0]`. On my PC `validationLayers.size()` is equals to zero.
To fix the issue I've changed the way the address of buffer is obtained to `validationLayers.data()`.

The second one was on application shutdown in call `void recreateSwapchain()` while trying to draw to closed window.
To fix that I've added a check that window is still open before trying drawing to it.

I was using Windows 10 21H1 build 19043.1348 and Visual Studio 2019 (16.11.7) with Windows SDK 10.0.22000.0 to debug and test this change.

## Tasks

* [NO] Tested on Linux
* [YES] Tested on Windows
* [NO] Tested on macOS
* [NO] Tested on iOS
* [NO] Tested on Android

## How to test this PR?

Compile SFML with examples and run Vulkan example.

If `validationLayers` is not empty on your environment, you may comment lines 451 and 455 to test my first change.
```cpp
//                validationLayers.push_back("VK_LAYER_LUNARG_standard_validation");
...
//                validationLayers.push_back("VK_LAYER_LUNARG_monitor");
```

To test the second change, run Vulkan example from Visual Studio with debugger attached. You should see no exceptions after window was closed.

### DIFF
--- a/examples/vulkan/Vulkan.cpp
+++ b/examples/vulkan/Vulkan.cpp
@@ -2479,7 +2479,8 @@ public:
                     swapchainOutOfDate = true;
             }
 
-            if (vulkanAvailable)
+            // Check that window was not closed before drawing to it
+            if (vulkanAvailable && window.isOpen())
             {
                 // Update the uniform buffer (matrices)
                 updateUniformBuffer(clock.getElapsedTime().asSeconds());
--- a/examples/vulkan/Vulkan.cpp
+++ b/examples/vulkan/Vulkan.cpp
@@ -473,9 +473,9 @@ public:
         instanceCreateInfo.sType = VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO;
         instanceCreateInfo.pApplicationInfo = &applicationInfo;
         instanceCreateInfo.enabledLayerCount = static_cast<sf::Uint32>(validationLayers.size());
-        instanceCreateInfo.ppEnabledLayerNames = &validationLayers[0];
+        instanceCreateInfo.ppEnabledLayerNames = validationLayers.data();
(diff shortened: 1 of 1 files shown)


binary1248:
I think the change from `&vector[0]` to `vector.data()` belongs to the family of code modernizations that @vittorioromeo is currently undertaking. In C++11, `.data()` was provided that allows us to do what was previously done with `&vector[0]` without causing undefined behaviour when working with an empty container. Because we now force Visual Studio to build with C++17 it correctly activates additional debug assertions that were not active without C++17 thus leading to this situation.

I suggest that this modernization is performed throughout the codebase instead of only where it currently triggers the assertion. Other containers might be empty under other circumstances and trigger the assertion as well.

@vittorioromeo Were you planning on making this change yourself, or is this something @Si1ver can help with?

vittorioromeo:
@binary1248: I am in favour of this change, and I would appreciate any contribution or help from @Si1ver. I do not mind merging this PR on its own, as long as we capture the intention of performing this change on the entire codebase as an issue. 

@Si1ver, let me know if you'd like to make this change yourself in the entire codebase so that we do not duplicate work :)

Si1ver:
@vittorioromeo There is not much places where we need to change `&vector[0]` to `vector.data()` so I will be happy to submit PR implementing that change 🙂

vittorioromeo:

binary1248:

binary1248 (approved):

PR #1874, opened 2021-12-07
[WIP] Implement wrapper for loading libraries on Win32

Si1ver:
## Description

This change adds private `LibraryLoader` utility class that automates loading and unloading Win32 dynamic libraries within it's scope.

The library is loaded when object is created and released upon it's destruction.
The implementation of `loadEntryPoint` routine in current Vulkan library loading wrapper implementation was very nice. After trying a few different approaches I've decided that current function definition is most elegant for almost all usage scenarios that we have at the moment.

I've updated Window implementation, Joystick implementation and Vulkan implementation to use new LibraryLoader class.

The documentation for `LibraryLoader ` is not completed yet. I would like to get feedback about this change before writing the documentation.

I've implemented no automated tests for implemented class yet. Please let me know if I should add tests. 

I was using Windows 10 21H1 build 19043.1348 and Visual Studio 2019 (16.11.7) with Windows SDK 10.0.22000.0 to develop and test this change.

## Tasks

* [NO] Tested on Linux
* [YES] Tested on Windows
* [NO] Tested on macOS
* [NO] Tested on iOS
* [NO] Tested on Android

## How to test this PR?

I've manually tested example applications `joystick`, `opengl`, `vulkan`, `win32` and `window` to make sure that the change isn't breaking anything.
There are no visible changes or added features.

```

### CODE REVIEW

**Comment by eXpl0it3r** [comment]
As it's WIP and currently only supports Windows, I've change the PR into a draft PR.

[Previous discussions](https://en.sfml-dev.org/forums/index.php?topic=17205.0) on the topic of providing library loading facilities.

**Comment by Si1ver** [comment]
Thanks for the link to the discussion! I've missed that.

My implementation matches what Laurent said in discussion. It's simple API with 3 functions, it hides all the casts and there is no demangling stuff.

As of now it's private (does not expose to SFML user). That fact also match Laurent's decision in the beginning of disussion.

@eXpl0it3r Do you think that I should implement this wrapper for another platforms in this PR? It's located inside Win32 at the moment and I'm only using it in Win32Impls. In does not change or brake anything outside of Win32 and may be considered complete without suppotring another platforms.

**Comment by eXpl0it3r** [comment]
Ah, I didn't realize that this was just an internal implementation.

Re-reading the thread again, it seems like the overall conclusion was that a simple, public library loading feature in sfml-system would be quite useful in multiple scenarios and I guess, we could even use it internally (which was often used as justification to implement features).

If you have the time and interest, could you come up with implementations for Linux/macOS or at least check, that the current API would also work on those platforms? And then move things into a public API?

File: src/SFML/System/Win32/LibraryLoader.hpp [hunk 1]
```diff
@@ -0,0 +1,120 @@
...
+    ////////////////////////////////////////////////////////////
+    virtual ~LibraryLoader();
+
+    ////////////////////////////////////////////////////////////
+    /// \brief todo
+    ///
+    /// \return todo
+    ///
+    ////////////////////////////////////////////////////////////
+    bool getIsLoaded();
```
eXpl0it3r, line 81: `getIsX` is quite odd, it's usually either `getX` or `isX` 😉

src/SFML/System/Win32/LibraryLoader.hpp
```diff
@@ -0,0 +1,120 @@
...
+    ///
+    /// todo
+    ///
+    /// \param procName todo
+    ///
+    /// \return True if the TODO
+    ///
+    ////////////////////////////////////////////////////////////
+    template<typename TFuncType>
+    bool getProcAddress(TFuncType& procAddress, const sf::String& procName);
```
eXpl0it3r, line 94: Write out the full name of terms, i.e. `getProcedureAddress`

Si1ver:
I have a macOS laptop and I think I can easilly setup a linux environment in VirtualBox.

I would like to make more valuable contribution so I'll check how much time I need to see how dynamic library loading is used in SFML on Linux/macOS and to add implementation for this platforms.

I'll let you know shortly! 😊

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -398,35 +402,31 @@ JoystickState JoystickImpl::update()
...
-        DirectInput8CreateFunc directInput8Create = reinterpret_cast<DirectInput8CreateFunc>(reinterpret_cast<void*>(GetProcAddress(dinput8dll, "DirectInput8Create")));
+    if (!localDinput8LibraryPtr->getIsLoaded())
+        return;
 
-        if (directInput8Create)
-        {
-            // Try to acquire a DirectInput 8.x interface
-            HRESULT result = directInput8Create(GetModuleHandleW(nullptr), 0x0800, guids::IID_IDirectInput8W, reinterpret_cast<void**>(&directInput), nullptr);
+    // Try to get the address of the DirectInput8Create entry point
+    typedef HRESULT(WINAPI *DirectInput8CreateFunc)(HINSTANCE, DWORD, REFIID, LPVOID*, LPUNKNOWN);
```
reconndev, line 411: Since this is C++11, how about `using` declarations instead of `typedef`'s?

```suggestion
    using DirectInput8CreateFunc = HRESULT (WINAPI*)(HINSTANCE, DWORD, REFIID, LPVOID*, LPUNKNOWN);
```
and so on

src/SFML/Window/Win32/VulkanImplWin32.cpp
```diff
@@ -35,72 +36,51 @@
...
-        library(nullptr)
-        {
-        }
-
         ~VulkanLibraryWrapper()
         {
-            if (library)
-                FreeLibrary(library);
+            if (libraryPtr)
+                libraryPtr.release();
```
reconndev, line 50: This is a memory leak. "Release" does not mean "free memory". It releases ownership so that another code has to free it. Unique pointers don't have to be freed manually. That's the whole point

src/SFML/System/Win32/LibraryLoader.hpp
```diff
@@ -0,0 +1,120 @@
...
+    ////////////////////////////////////////////////////////////
+    LibraryLoader(const sf::String& libraryName);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Default destructor
+    ///
+    /// todo
+    ///
+    ////////////////////////////////////////////////////////////
+    virtual ~LibraryLoader();
```
reconndev, line 73: Is the intention for `LibraryLoader` to be a base class? Judging by the public interface it seems to  favor composition. Consider changing the class to final and removing `virtual`:
```cpp
class SFML_SYSTEM_API LibraryLoader final : public NonCopyable
```

src/SFML/Window/Win32/JoystickImpl.cpp
Si1ver: Indeed. Nice idea.
I think I need a permission of @vittorioromeo to move from `typedef` to `using` in on;y one place of codebase.

src/SFML/System/Win32/LibraryLoader.hpp
Si1ver: Nope, no inheritance intended. That's totally because it's inherited from `NonCopyable`. I believe that we should make `~NonCopyable()` also a `virtual`.

src/SFML/Window/Win32/VulkanImplWin32.cpp
Si1ver: Yep, that's mistake. It should be `reset()`.

src/SFML/System/Win32/LibraryLoader.hpp
reconndev: `NonCopyable` should be deprecated in SFML 3.0.
```cpp
LibraryLoader(const LibraryLoader&) = delete;
```

Si1ver: Oh, that's perfect solution to the problem 🙂 I'll switch to `=delete` for copy constructor and assignment operator.

src/SFML/Window/Win32/VulkanImplWin32.cpp
reconndev: For the record, the reset will be called automatically on destruction too.

Si1ver: Indeed. It's totally pointless to call anything here. I'll just remove destructor.
(well except if we want to leak some memory 🙂 )

src/SFML/System/Win32/LibraryLoader.cpp
```diff
@@ -0,0 +1,75 @@
...
+{
+namespace priv
+{
+
+////////////////////////////////////////////////////////////
+LibraryLoader::LibraryLoader(const sf::String& libraryName) :
+m_libraryName(libraryName)
+{
+    m_moduleHandle = LoadLibraryW(m_libraryName.toWideString().c_str());
+}
```
vittorioromeo, line 44:
```suggestion
m_libraryName(libraryName), m_moduleHandle(LoadLibraryW(m_libraryName.toWideString().c_str()))
{
}
```

src/SFML/System/Win32/LibraryLoader.hpp
vittorioromeo:
```suggestion
    [[nodiscard]] bool getIsLoaded();
```

src/SFML/System/Win32/LibraryLoader.hpp
vittorioromeo:
```suggestion
    [[nodiscard]] bool getProcAddress(TFuncType& procAddress, const sf::String& procName);
```

src/SFML/System/Win32/LibraryLoader.hpp
```diff
@@ -0,0 +1,120 @@
...
+    /// todo
+    ///
+    ////////////////////////////////////////////////////////////
+    void freeLibrary();
+
+    ////////////////////////////////////////////////////////////
+    // Member data
+    ////////////////////////////////////////////////////////////
+    sf::String m_libraryName;           //!< todo
+    HMODULE    m_moduleHandle{nullptr}; //!< todo
```
vittorioromeo, line 110: Is the `{nullptr}` initialization needed at all? The constructor always seems to initialize `m_moduleHandle`.

src/SFML/System/Win32/LibraryLoader.inl
```diff
@@ -0,0 +1,42 @@
...
+    if (m_moduleHandle == nullptr)
+    {
+        procAddress = nullptr;
+        return false;
+    }
+
+    FARPROC farprocProcAddress = GetProcAddress(m_moduleHandle, procName.toAnsiString().c_str());
+
+    // Avoid using reinterpret_cast<>
+    memcpy(&procAddress, &farprocProcAddress, sizeof(farprocProcAddress));
```
vittorioromeo, line 39:
```suggestion
    std::memcpy(&procAddress, &farprocProcAddress, sizeof(farprocProcAddress));
```

src/SFML/System/Win32/LibraryLoader.hpp
```diff
@@ -0,0 +1,120 @@
...
+#define SFML_LIBRARY_LOADERWIN32_HPP
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#include <SFML/System/Export.hpp>
+#include <SFML/System/NonCopyable.hpp>
+#include <SFML/System/String.hpp>
+#include <windows.h>
+
```
vittorioromeo, line 35:
```suggestion
#include <cstring>

```

For `std::memcpy`

src/SFML/Window/Win32/JoystickImpl.cpp
vittorioromeo: @Si1ver: Using `using` here is fine, I don't have a problem with in-flight PRs doing so. I will likely change all existing `typedef` to `using` before this PR is merged, anyway.

src/SFML/System/Win32/LibraryLoader.hpp
```diff
@@ -0,0 +1,120 @@
...
+#ifndef SFML_LIBRARY_LOADERWIN32_HPP
+#define SFML_LIBRARY_LOADERWIN32_HPP
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#include <SFML/System/Export.hpp>
+#include <SFML/System/NonCopyable.hpp>
+#include <SFML/System/String.hpp>
+#include <windows.h>
```
vittorioromeo, line 34: I don't like exposing `windows.h` in public headers. I would suggest changing this class to use the PImpl idiom, unless the intention for `LibraryLoader.hpp` is to be an implementation header, and only expose it through PImpl to the users. Please confirm.

vittorioromeo: Also, do we want
```cpp
#ifndef WIN32_LEAN_AND_MEAN
#define WIN32_LEAN_AND_MEAN
#endif
```
here?

vittorioromeo (commented):
Some coding comments and questions. I like the idea and appreciate the effort!

src/SFML/Window/Win32/JoystickImpl.cpp
Si1ver: Good. I'll follow your style then.

src/SFML/System/Win32/LibraryLoader.hpp
Si1ver: I feel the same here for both sentenses.

Since we decided to move LibraryLoader from internal implementation to public, I'll implement PImpl the same way other classes in SFML are implemented. This should deal with exposing windows.h.

The thing with WIN32_LEAN_AND_MEAN - it's better to add but we are using it in some files and not using in other.
This is confusing. What do you think if we will add a small header file for including windows.h? I've done an implementation for another project, you can see it here https://github.com/teonet-co/teobase/blob/master/include/teobase/windows.h

src/SFML/System/Win32/LibraryLoader.hpp
Bromeon: To extend on eXpl0it3r's comment, should be `isLoaded()`, not `getLoaded()`.

Also agree with Vittorio, return types indicating errors should be `[[nodiscard]]`.

Si1ver: Done.

src/SFML/System/Win32/LibraryLoader.hpp
Si1ver: Done. Added [[nodiscard]]. Fixed shortened names for functions and some variables.

Si1ver:
I believe that I've done/fixed all suggested small changes.

What's next on my list:
* Make LibraryLoader public and implement PImpl idiom
* Correctly include windows.h (see my suggestion above in comment #discussion_r764025447 )
* Add Unix implementation using `dlopen` and `dlclose`
* Finish documentation

There is no example where to place and how to arrange deleted (`= delete`) class members in header file.

The change is getting bigger than I initially assumed. Should I create discussion on the forum?

src/SFML/System/Win32/LibraryLoader.hpp
Bromeon: > Also agree with Vittorio, return types indicating errors should be `[[nodiscard]]`.

I was a bit tired when I wrote this 😬

My take on `[[nodiscard]]` is actually [described here](https://github.com/SFML/SFML/pull/1872#discussion_r764657674). I just realized that `isLoaded()` is a simple getter, not an error reporting method, so we don't gain much with `[[nodiscard]]`.

Si1ver: I see. Use to force error handling or when misuse of routine is possible.

eXpl0it3r:

Si1ver:
It won't build yet. Unix implementation is included but not yet implemented.

Si1ver:
I've had some progress on that PR. PImpl is done and Unix implementation (Linux / macOS) is done. Also rebased everything on top of master branch.
At the moment I still haven't tested it well yet. Documentation and testing are still need to be done.

The most important question I have right now: do we want to change the name to better match SFML naming style? To something like `DynamicLibrary`?

The other question is there is an ugly place where C++ templates meet PImpl. Not sure if I've done it right. Especially with function pointers.
The only improvement I can think about is defining a type for a `void *` like we do with OpenGL functions
`using GlFunctionPointer = void (*)();`

I will be happy to get some feedback on this iteration if someone will have some free time 🙂

PR #1875, opened 2021-12-07
Use std::vector.data() where appropriate

Si1ver:
## Description

There were some places where an address of the first element of the `std::vector` was used as a pointer to array of vector's elements. These places were replaced with `std::vector.data()` that semantically better for that usage scenario.

This change was discussed in PR #1873

Two occurrences in file `examples/vulkan/Vulkan.cpp` was left untouched to avoid merge conflict with PR #1873.

Also in file `examples/vulkan/Vulkan.cpp` in `void setupRenderpass()` routine additional change to an array was made.
Array `VkAttachmentReference attachmentReferences[2];` was split into two separate variables for better readability and clearer variable usage.

I was using Windows 10 21H1 build 19043.1348 and Visual Studio 2019 (16.11.7) with Windows SDK 10.0.22000.0 to develop and test this change. 

Also I was using Ubuntu 20.04.3 with latest updates running under VirtualBox to build and test this change. Clang 10.0.0 was used for building.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

I've manually tested example applications to make sure that the change isn't breaking anything:
* on Windows: island, joystick, opengl, shader, sockets, sound, tennis, vulkan, win32 and window
* on Linux: island, joystick, opengl, shader, sockets, sound, tennis, window and x11

Sadly I was unable to test vulkan example on Linux. Seems that running Vulkan under VirtualBox involves some magic tricks that I'm not familiar with yet.

There are no visible changes or added features.

### DIFF
--- a/examples/vulkan/Vulkan.cpp
+++ b/examples/vulkan/Vulkan.cpp
@@ -341,7 +341,7 @@ public:
             vkWaitForFences(device, 1, &fences[i], VK_TRUE, std::numeric_limits<uint64_t>::max());
 
         if (commandBuffers.size())
-            vkFreeCommandBuffers(device, commandPool, static_cast<sf::Uint32>(commandBuffers.size()), &commandBuffers[0]);
+            vkFreeCommandBuffers(device, commandPool, static_cast<sf::Uint32>(commandBuffers.size()), commandBuffers.data());
 
         commandBuffers.clear();
 
@@ -424,7 +424,7 @@ public:
 
         layers.resize(objectCount);
 
-        if (vkEnumerateInstanceLayerProperties(&objectCount, &layers[0]) != VK_SUCCESS)
+        if (vkEnumerateInstanceLayerProperties(&objectCount, layers.data()) != VK_SUCCESS)
         {
             vulkanAvailable = false;
             return;
@@ -486,7 +486,7 @@ public:
             requiredExtentions.pop_back();
 
             instanceCreateInfo.enabledExtensionCount = static_cast<sf::Uint32>(requiredExtentions.size());
-            instanceCreateInfo.ppEnabledExtensionNames = &requiredExtentions[0];
+            instanceCreateInfo.ppEnabledExtensionNames = requiredExtentions.data();
 
             result = vkCreateInstance(&instanceCreateInfo, 0, &instance);
         }
@@ -553,7 +553,7 @@ public:
 
         devices.resize(objectCount);
 
-        if (vkEnumeratePhysicalDevices(instance, &objectCount, &devices[0]) != VK_SUCCESS)
+        if (vkEnumeratePhysicalDevices(instance, &objectCount, devices.data()) != VK_SUCCESS)
         {
             vulkanAvailable = false;
             return;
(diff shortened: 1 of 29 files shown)


examples/vulkan/Vulkan.cpp
```diff
@@ -981,22 +981,20 @@ class VulkanExample
         attachmentDescriptions[1].initialLayout = VK_IMAGE_LAYOUT_UNDEFINED;
         attachmentDescriptions[1].finalLayout = VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL;
 
-        VkAttachmentReference attachmentReferences[2];
+        VkAttachmentReference colorAttachmentReference = VkAttachmentReference();
```
vittorioromeo, line 984:
```suggestion
        VkAttachmentReference colorAttachmentReference;
```

examples/vulkan/Vulkan.cpp
```diff
@@ -981,22 +981,20 @@ class VulkanExample
...
+        colorAttachmentReference.layout = VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL;
 
-        attachmentReferences[0] = VkAttachmentReference();
-        attachmentReferences[0].attachment = 0;
-        attachmentReferences[0].layout = VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL;
-
-        attachmentReferences[1] = VkAttachmentReference();
-        attachmentReferences[1].attachment = 1;
-        attachmentReferences[1].layout = VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL;
+        VkAttachmentReference depthStencilAttachmentReference = VkAttachmentReference();
```
vittorioromeo, line 988:
```suggestion
        VkAttachmentReference depthStencilAttachmentReference;
```

vittorioromeo (changes requested):
Looks good, just a minor thing. Thanks!

eXpl0it3r:

examples/vulkan/Vulkan.cpp
Bromeon: This changes semantics, from zero-initialized to uninitialized.

Maybe 
```cpp
VkAttachmentReference depthStencilAttachmentReference{};
// or
VkAttachmentReference depthStencilAttachmentReference = {};
```
(the 2nd notation makes it more clear it's a C struct and not a constructor call)

examples/vulkan/Vulkan.cpp
Bromeon: see below

examples/vulkan/Vulkan.cpp
Si1ver: I think that this should be zero initialized in case some implementation have (or may have in the future) additional fields. I've changed both occurrences to `= {};`
Also rebased branch on top of master.

vittorioromeo (approved):
Looks good! Thank you!

@Si1ver: could you please rebase on master?

eXpl0it3r:

PR #1877, opened 2021-12-08
Use alias declarations instead of 'typedef'

vittorioromeo:
## Description

Pure refactoring change. Mostly automated with `clang-tidy`, but also some manual work to fix-up some comments and some places where `clang-tidy` failed.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI. I've also ran some of the examples and everything was fine.


eXpl0it3r (approved):
Looks good to me, I assume a type alias can do anything a typedef can (for the given changes), correct?

vittorioromeo:
> Looks good to me, I assume a type alias can do anything a typedef can (for the given changes), correct?

Yes, it has the same exact semantics. Just alternative syntax.

PR #1879, opened 2021-12-08
Minor modernization changes: 'nullptr', range-based 'for' loops, etc...

vittorioromeo:
## Description

This PR introduces various mixed modernization changes, mostly automated with `clang-tidy`. I used the following checks: 
```bash
run-clang-tidy -header-filter='.*' -checks="-*,modernize-use-bool-literals,modernize-shrink-to-fit, \
modernize-avoid-bind,modernize-deprecated-headers,modernize-loop-convert,\
modernize-raw-string-literal,modernize-redundant-void-arg,modernize-use-bool-literals,\
modernize-use-default,modernize-use-emplace,modernize-use-nullptr" -fix
```

I then manually went through every change and:
- Ensured nothing was broken. For example, `clang-tidy` uses `nullptr` as much as possible, but APIs like Vulkan do not guarantee that handles are always pointers (they can also be int64), so I reverted those to `0`.
- Rename things to make more sense. `clang-tidy` gives bogus names when converting a normal loop to a range-based one. 

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.

### DIFF
--- a/src/SFML/Network/Ftp.cpp
+++ b/src/SFML/Network/Ftp.cpp
@@ -544,12 +544,12 @@ Ftp::Response Ftp::DataChannel::open(Ftp::TransferMode mode)
             Uint8 data[6] = {0, 0, 0, 0, 0, 0};
             std::string str = response.getMessage().substr(begin);
             std::size_t index = 0;
-            for (int i = 0; i < 6; ++i)
+            for (unsigned char& datum : data)
             {
                 // Extract the current number
                 while (isdigit(str[index]))
                 {
-                    data[i] = static_cast<Uint8>(static_cast<Uint8>(data[i] * 10) + static_cast<Uint8>(str[index] - '0'));
+                    datum = static_cast<Uint8>(static_cast<Uint8>(datum * 10) + static_cast<Uint8>(str[index] - '0'));
                     index++;
                 }
 
--- a/src/SFML/Window/WindowBase.cpp
+++ b/src/SFML/Window/WindowBase.cpp
@@ -308,7 +308,7 @@ bool WindowBase::hasFocus() const
 ////////////////////////////////////////////////////////////
 WindowHandle WindowBase::getSystemHandle() const
 {
-    return m_impl ? m_impl->getSystemHandle() : 0;
+    return m_impl ? m_impl->getSystemHandle() : WindowHandle{};
 }
 
 
--- a/examples/island/Island.cpp
+++ b/examples/island/Island.cpp
@@ -529,12 +529,12 @@ void threadFunction()
 
     std::vector<sf::Vertex> vertices(resolutionX * rowBlockSize * 6);
 
-    WorkItem workItem = {0, 0};
+    WorkItem workItem = {nullptr, 0};
 
     // Loop until the application exits
     for (;;)
     {
-        workItem.targetBuffer = 0;
+        workItem.targetBuffer = nullptr;
 
         // Check if there are new work items in the queue
         {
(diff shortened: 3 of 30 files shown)


src/SFML/Network/Ftp.cpp
```diff
@@ -544,12 +544,12 @@ Ftp::Response Ftp::DataChannel::open(Ftp::TransferMode mode)
             Uint8 data[6] = {0, 0, 0, 0, 0, 0};
             std::string str = response.getMessage().substr(begin);
             std::size_t index = 0;
-            for (int i = 0; i < 6; ++i)
+            for (unsigned char& datum : data)
```
Bromeon, line 547: Maybe name it `byte`?

src/SFML/Window/WindowBase.cpp
```diff
@@ -308,7 +308,7 @@ bool WindowBase::hasFocus() const
 ////////////////////////////////////////////////////////////
 WindowHandle WindowBase::getSystemHandle() const
 {
-    return m_impl ? m_impl->getSystemHandle() : 0;
+    return m_impl ? m_impl->getSystemHandle() : WindowHandle{};
```
Bromeon, line 311: According to the [SFML doc](https://www.sfml-dev.org/documentation/2.5.1/group__window.php#gaed947028b0698a812cad2f97bfe9caa3), most handle types are pointers (`HWND` is one, not sure about X11 `Window`), so I wonder if we shouldn't use `nullptr` here?

`WindowHandle{}` is more generic, but could possibly also represent a type which cannot be explicitly checked for a "null state". So if we lose that ability (and the user cannot check the returned handle for validity), I'd probably prefer a compile error.

Bromeon (commented):
Thank you!

src/SFML/Window/WindowBase.cpp
vittorioromeo: Force-pushing is the culprit here, but originally I had changed this to 

```cpp
return m_impl ? m_impl->getSystemHandle() : nullptr;
```

which caused a compilation failure on Unix, because we have

```cpp
#elif defined(SFML_SYSTEM_LINUX) || defined(SFML_SYSTEM_FREEBSD) || defined(SFML_SYSTEM_OPENBSD) || defined(SFML_SYSTEM_NETBSD)

    // Window handle is Window (unsigned long) on Unix - X11
    using WindowHandle = unsigned long;
```

So using `WindowHandle{}` is the best choice, IMHO.

src/SFML/Network/Ftp.cpp
vittorioromeo: Hmm... not entirely against the suggestion, but:
1. C++17 introduces `std::byte`, and it might cause some confusion
2. The container is named `data`. Maybe the container should be named `bytes` instead? If so, I'd rather perform that refactoring in a separate PR.

Let me know if you feel strongly about this.

Bromeon: Not very strongly, just thought "datum" is quite unusual. Naming the container `bytes` is also OK. Not sure if a PR just for that isn't too much effort; but maybe you can combine it with other things.

src/SFML/Window/WindowBase.cpp
Bromeon: I see, thanks for explanation!

Bromeon:
Are there notable changes since my last review? Otherwise, good to go ✔️

vittorioromeo:
> Are there notable changes since my last review? Otherwise, good to go ✔️

Nope, just rebased on master. Feel free to approve if you are fine with the resolution of your two comments above.

Bromeon (approved):

PR #1880, opened 2021-12-08
Use 'auto' where no information is lost

vittorioromeo:
## Description

Run `clang-tidy` with `modernize-use-auto`. It did a really good job here. I've checked all usage sites, and no information was lost -- the type is always visible. `auto` was only used to reduce redundancy.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.


Bromeon (approved):
Nice! I like how you limited the changes to places which suffered from repetition; I also think for such cases `auto` is great 🙂

PR #1883, opened 2021-12-09
Mark some non-'void' 'const'-qualified member functions as '[[nodiscard]]'

vittorioromeo:
## Description

Automatic run of `clang-tidy`'s `modernize-use-nodiscard` check with minor manual adjustments. This check is very conservative and marks only "obvious" `const`-qualified member functions that return something. There are more places where we could use `[[nodiscard]]`, but we can get there with multiple PRs. In other words, this is a very low-risk one to merge.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.

### DIFF
--- a/examples/shader/Effect.hpp
+++ b/examples/shader/Effect.hpp
@@ -25,7 +25,7 @@ public:
         s_font = &font;
     }
 
-    const std::string& getName() const
+    [[nodiscard]] const std::string& getName() const
     {
         return m_name;
     }
--- a/include/SFML/Audio/InputSoundFile.hpp
+++ b/include/SFML/Audio/InputSoundFile.hpp
@@ -106,7 +106,7 @@ public:
     /// \return Number of samples
     ///
     ////////////////////////////////////////////////////////////
-    Uint64 getSampleCount() const;
+    [[nodiscard]] Uint64 getSampleCount() const;
 
     ////////////////////////////////////////////////////////////
     /// \brief Get the number of channels used by the sound
@@ -114,7 +114,7 @@ public:
     /// \return Number of channels (1 = mono, 2 = stereo)
     ///
     ////////////////////////////////////////////////////////////
-    unsigned int getChannelCount() const;
+    [[nodiscard]] unsigned int getChannelCount() const;
 
     ////////////////////////////////////////////////////////////
     /// \brief Get the sample rate of the sound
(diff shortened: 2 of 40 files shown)


Bromeon:
In that regard, maybe see my forum post yesterday: https://en.sfml-dev.org/forums/index.php?topic=28340

vittorioromeo:
@Bromeon: gotcha, I can definitely understand your point of view. I'm more of the opinion that verbosity is a small price to pay for consistency, and I'm the kind of person that likes to use `const` and `[[nodiscard]]` whenever possible, even if it is highly unlikely that a function would be misused.

I feel like `const` and `[[nodiscard]]` should have been the defaults, so I see it as a language deficiency, and I can't help but try to do my best to promote their uses. While in some situations it might not be necessary, use of `[[nodiscard]]` throughout an entire codebase promotes consistency and acts as documentation that a function doesn't have any meaningful side effect. Therefore, any function *without* `[[nodiscard]]` becomes subject of a deeper analysis when invoked by a user.

I understand that this is subjective so I won't push for it. I think it would be more impactful to mark the member functions you mentioned as `[[nodiscard]]` instead.

Bromeon:
Thanks for elaborating! I understand your point of view, as well. Over the decades and with its C compatibility, it became clear that C++ has quite a few "wrong defaults". Some are more or less universally agreed upon, others a bit more controversial:
* `const` vs mutable
* `explicit` vs accepting implicit conversions
* `noexcept` vs throwing
* `final` vs `open` (see Kotlin)
* variable-sized primitives vs `std::int{N}_t`
* variables uninitialized by default
* unchecked container access for common syntax
* copy vs move (?)
* ...

They can typically be mitigated in Modern C++ (with varying level of success). However, almost always at a cost, often extra verbosity. So it shifts from an obvious default choice to being a trade-off. Some are widely accepted because they're not very intrusive, others depend heavily on the code style.

Well possible that this discussion would be different, were `[[nodiscard]]` opt-out instead of opt-in. However, even that seems not entirely uncontroversial (unlike some points in the above list) -- for example, there's an interesting discussion about Rust's `#[must_use]` attribute [on Reddit](https://www.reddit.com/r/rust/comments/hwlemy/why_is_must_use_optin_instead_of_optout).

vittorioromeo:
No consensus.

PR #1885, opened 2021-12-09
Replace sf::Time with chrono (draft proposal)

reconndev:
  * [Forum topic](https://en.sfml-dev.org/forums/index.php?topic=21639.15)

----

## Description

This is a draft proposal to introduce `std::chrono` types as a replacement for `sf::Time`.

_Before reviewing for alternate solutions please see the "**Alternatives to consider**" section below_

`sf::Time` has been a generic time-representing class that stored its value in microsecond precision. It was able to return it in any other time unit format with member functions like: `sf::Time::asMilliseconds()`.

* `std::this_thread::sleep_for` was moved into SFML namespace and replaced `sf::sleep` (naming convention inconsistency)
* `std::chrono::duration_cast` was moved into SFML namespace (naming convention inconsistency)
* `std::chrono::*seconds` were alias-templated to allow for switching between different representation types

This PR is dependent on:
* #1872

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## Usage example

```cpp
constexpr auto t1 = sf::Seconds<float>(0.1f);
static_assert(t1.count() == 0.1f);

// Overflow possible - need to use duration_cast
constexpr auto milli = sf::duration_cast<sf::Milliseconds<>>(t1).count();
static_assert(milli == 100);

constexpr auto t2 = sf::Milliseconds<>(30);
static_assert(t2.count() == 30);

// No overflow possible - constructor available
constexpr auto micro = sf::Microseconds<>(t2);
static_assert(micro.count() == 30'000);

constexpr auto t3 = sf::Microseconds<>(-800'000);
static_assert(t3.count() == -800'000);

constexpr float sec = sf::duration_cast<sf::Seconds<float>>(t3).count();
static_assert(sec == -0.8f);

struct Game
{
    constexpr void update(sf::Seconds<float> elapsed)
    {
        position += speed * elapsed.count();
    };

    float position = 0.f;
    const float speed = 2.f;
};
static_assert([] {
    Game game;
    game.update(sf::Milliseconds<>(100));
    // game.position == 0.2f
    game.update(sf::Nanoseconds<>(100'000'000));
    // game.position == 0.4f
    return game.position;
}() == 0.4f);
```

# Advantages

* Time unit is known from function signatures at compile-time
* Implicit casts are only allowed when safe
* Explicit casts are done using `sf::duration_cast`
* `using namespace std::chrono_literals` for easy literal suffix notation, e.g.: `1s`, `500ms`, `0.3s`.
* Alias templates allow for specifying time representation type as floating-point or integral numbers and converting between them
* Compatibility with `<chrono>`
* `constexpr` types allow for compile-time evaluation (see static_asserts in Usage example above)

# Alternatives to consider

Since this is a draft I don't want to pressure onto using this exact implementation. Here are the other suggestions:

### Drop the alias templates - use chrono types directly from the `std::chrono` namespace.
  - Bad: There are places where code uses floating-point seconds. Chrono does not alias these types and forces users to write `duration<float, std::micro>` - microseconds, `duration<double, std::nano>` - nanoseconds, `duration<double, std::ratio<60>>` - minutes.
  - Bad: Users need to write explicitly type all namespaces or append `using` declarations: `std::chrono::nanoseconds`.

### Specify non-templated aliases in SFML namespace.
```cpp
using Seconds = std::chrono::seconds;
using SecondsF = std::chrono::duration<float, std::ratio<1>>;
```
  - Bad: Less flexible. Can't specify the exact needed type, cast required, e.g.: `sf::Seconds<decltype(time.sec)>(elapsed)` or `sf::Milliseconds<sf::Int16>(t)`
  -  Good: No need to specify empty template arguments with `<>` like in `sf::Nanoseconds<>`

### Fetch `std::chrono` types and utilities into SFML namespace
  - Good: Shortens the syntax to `sf::nanoseconds`
  - Bad: Inconsistent naming convention as opposed to expected `sf::Nanoseconds`

### Other ideas are welcome
...


Bromeon:
While I agree that we should use Chrono internally, I'm personally not convinced that replacing `sf::Time` is the way to go. If I look at the change list, almost every single occurrence of code is now either more verbose or less expressive than before.

Just a random sample:
```diff
- effects[current]->update(clock.getElapsedTime().asSeconds(), x, y);
+ effects[current]->update(sf::Seconds<float>(clock.getElapsedTime()).count(), x, y);
      
- osstr << "Frame:  " << clock.restart().asMilliseconds() << "ms\n"
+ osstr << "Frame:  " << sf::duration_cast<sf::Milliseconds<>>(clock.restart()).count() << "ms\n"

- Response connect(..., Time timeout = Time::Zero);
+ Response connect(..., Microseconds<> timeout = Microseconds<>::zero());
```

You linked the forum thread, but there was no consensus that we would remove `sf::Time`. On the contrary, Laurent himself [was against it](https://en.sfml-dev.org/forums/index.php?topic=21639.msg153739#msg153739). I'd prefer if we keep SFML simple -- Chrono is a rather low-level library that is designed to fit all possible purposes, but in *S*FML we aim to make things as simple as possible for the user.

I think what _does_ makes sense is:
* to expose conversion functions from/to standard durations
* consider integration of user-defined literals, but they should be optional and pulled in via namespace

reconndev:
Thank you for the prompt feedback. I agree it is more verbose. This is a draft only to show how it would look/behave, and to get things moving as @vittorioromeo suggested rather than keeping things stuck on theoretical discourse.

It certainly **does not** look miles better.

> I feel a bit sorry for the huge amount of work that went into this.

No worries, I did not expect it to be merged. While forum discussions are good in their way, coding things gives a bigger perspective. As Linus Torvalds says: "Talk is cheap, show me the code". Regardless, a couple of regexes + vscode and things are coded fast. Also, M1 compiles SFML _really_ fast so I don't wait too long for compiler feedback.

I will probably create a separate draft for implementing your suggestions so that this doesn't get lost in case somebody thought: "How about fully integrating chrono into SFML?". As a precaution :)

Bromeon:
Ah OK, good to hear you saw it as a proof-of-concept. I fully agree that it's often better to try something out, rather than theoretizing endlessly 🙂

That said, are you still interested in conversion methods between `sf::Time` and Chrono, maybe after #1872 is merged? If yes, maybe we could quickly discuss the API on the forum first.

reconndev:
> That said, are you still interested in conversion methods between `sf::Time` and Chrono, maybe after #1872 is merged? If yes, maybe we could quickly discuss the API on the forum first.

Yes, definitely. Any hints API-wise on the forum would be of help and much appreciated.

I am still waiting for #1872 to be merged and waiting for the android build fix I provided for @binary1248 [here](https://github.com/SFML/SFML/pull/1863#issuecomment-988804070) because my new files seem to affect #1872 so somethings not right with the build job.

eXpl0it3r:
Btw if you need something to tinker with while waiting for the Android build fix, you can see the current Buildbot build setup in its dedicated repository. Maybe you can add your suggested build steps in there somehow. 🙂 
https://github.com/SFML/SFML-Buildbot

PR #1886, opened 2021-12-09
Remove deprecated 'PrimitiveType::Quads'

vittorioromeo:
## Description

This PR removes the deprecated `sf::PrimitiveType::Quads` enumerator and updates some usages and documentation. It is a breaking change IMHO very appropriate for SFML 3.x.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.

### DIFF
--- a/include/SFML/Graphics/Drawable.hpp
+++ b/include/SFML/Graphics/Drawable.hpp
@@ -110,7 +110,7 @@ protected:
 ///         target.draw(m_vertices, states);
 ///
 ///         // ... or draw with OpenGL directly
-///         glBegin(GL_QUADS);
+///         glBegin(GL_TRIANGLES);
 ///         ...
 ///         glEnd();
 ///     }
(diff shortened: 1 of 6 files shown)


binary1248 (approved):

nathanmyersc:
Hmm i downloaded your source like 10 hours ago. and had to change my code alot after you depreciated the quad. Made me learn some shit. its still not right. lots of buggy elements even within the sf::rectangleshape. 
But since we are eliminating the Quad can you guys implement something like an Element index array in sfml? So that we can save on vertexes and still only use 4.

vittorioromeo:
> Hmm i downloaded your source like 10 hours ago. and had to change my code alot after you depreciated the quad. Made me learn some shit. its still not right. lots of buggy elements even within the sf::rectangleshape. 

Are you saying that you found some issues with `sf:: RectangleShape`? Or in how your code uses it?

> But since we are eliminating the Quad can you guys implement something like an Element index array in sfml? So that we can save on vertexes and still only use 4.

I actually was talking about this to @eXpl0it3r a few days ago. I'd like something similar as well, we might consider implementing it.

nathanmyersc:
Well actually i had compiled 3 forms of sfml one debug one release one release with execptions.
The version i compiled sfml with execptions on has buggy buttons for some reason
only difference between release with execptions mode and release mode is i compiled sfml
without the -fnoexceptions flag and then did not have the -fnoexceptions flag in my exception accepting project.
Not sure why it causes a difference really. code is the same otherwise.

PR #1888, opened 2021-12-09
Strategic use of '[[nodiscard]]' in 'Graphics' module

vittorioromeo:
## Description

A more targeted approach to `[[nodiscard]]` compared to #1883. This has been done completely manually, applying the attribute to functions where is either not obvious that the return value is significant or where it would be a bug to forget to handle it.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.


examples/island/Island.cpp
```diff
@@ -2,6 +2,7 @@
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
+#include <cstdlib>
```
binary1248, line 5: I would move this include under `<cmath>`.

examples/island/Island.cpp
```diff
@@ -11,6 +12,7 @@
 #include <algorithm>
 #include <cstring>
 #include <cmath>
+#include <iostream>
```
binary1248, line 15: I would move this include above `<cstring>`.

examples/opengl/OpenGL.cpp
```diff
@@ -3,6 +3,8 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Graphics.hpp>
+#include <cstdlib>
+#include <iostream>
```
binary1248, line 7: I would swap these includes so that C++ facilities are included before C facilities.

examples/opengl/OpenGL.cpp
```diff
@@ -75,10 +78,14 @@ int main()
         // Attempt to generate a mipmap for our cube texture
         // We don't check the return value here since
         // mipmapping is purely optional in this example
-        texture.generateMipmap();
+        (void) texture.generateMipmap();
```
binary1248, line 81:
```cpp
static_cast<void>(texture.generateMipmap());
```

binary1248 (changes requested):

examples/opengl/OpenGL.cpp
eXpl0it3r: Do we have no warning enabled for C-casts? 🤔

vittorioromeo: Eh, `(void)` is ubiquitous. It's not really a bad use of a C-style cast IMHO.

examples/X11/X11.cpp
```diff
@@ -10,18 +10,24 @@
...
+///
 ////////////////////////////////////////////////////////////
-void initialize(sf::Window& window)
+[[nodiscard]] bool initialize(sf::Window& window)
 {
     // Activate the window
-    window.setActive();
+    if (!window.setActive())
+    {
+        std::cerr << "Failed to set window to active" << std::endl;
```
Bromeon, line 28: Should this be `sf::err()`?

(applies to others too)

examples/island/Island.cpp
vittorioromeo: Will fix

examples/opengl/OpenGL.cpp
Bromeon: With @eXpl0it3r's remark in mind, it's probably still good to use `static_cast`. Do we want to enable CI lints for C-style and function-style casts? Although we would need to limit it to SFML code itself (not extlibs)...

I agree that C++ chose the worst possible cast syntax of all languages, but well...

The alternative would be something like `SFML_UNUSED(...)`, but not sure if that's much nicer. I'd say casting to void is commonly understood.

src/SFML/Graphics/RenderTexture.cpp
```diff
@@ -143,13 +143,11 @@ bool RenderTexture::generateMipmap()
...
-    bool result = m_impl && m_impl->activate(active);
-
     // Update RenderTarget tracking
-    if (result)
-        RenderTarget::setActive(active);
+    if (m_impl && m_impl->activate(active))
+        return RenderTarget::setActive(active);
 
-    return result;
+    return false;
```
Bromeon, line 150: Nice refactoring!

Could it happen though that `m_impl->activate()` succeeds and `RenderTarget::setActive()` fails? In that case, would we end up in an inconsistent state? (It wouldn't be worse than previous behavior, just asking).

examples/X11/X11.cpp
vittorioromeo: I have always assumed that `sf::err` is SFML's *internal* error stream. Isn't that the case?

examples/opengl/OpenGL.cpp
vittorioromeo: I disagree, as I mentioned everyone knows that `(void) foo;` means "discard `foo`", and there is absolutely no risk in using that syntax in modern C++. I hate C as much as everyone else, but using `static_cast` here is needless. I feel a bit in the same way you feel about me trying to shoehorn `[[nodiscard]]` into everything :)

By the way, C-style cast warnings **intentionally exclude** `(void)` casts. See https://stackoverflow.com/a/53485874/598696

Bromeon: > I feel a bit in the same way you feel about me trying to shoehorn [[nodiscard]] into everything :)

So it's revenge! 😁

Jokes aside, if `(void) expr` is really a common idiom and is explicitly whitelisted by compilers, that changes things of course. I just checked with VS 2019, seems to be the same. Warning C4189 is not emitted with `(void) expr`, on level /W4.

src/SFML/Graphics/RenderTexture.cpp
vittorioromeo: Good point -- I guess it is technically possible, I am not sure exactly how the behavior here could be improved. Maybe @binary1248 is more familiar with this part of the code?

examples/X11/X11.cpp
Bromeon: Oh, I missed that this is an example, you're right.
Only the Vulkan.cpp example uses `sf::err()`, and that's probably by accident.

src/SFML/Graphics/RenderTexture.cpp
binary1248: Well on first glance, if `RenderTarget::setActive(active);` fails something was going to go wrong anyway. The question is: Have users gotten used to the current behaviour enough that this change will break their code? I must admit, when I wrote that code I must have had something in mind, otherwise I wouldn't have went through the extra effort of saving `result` to a temporary. I just can't remember what the reason was. In any case, I think if we want to err on the side of caution I don't think we should change the logic unless it is proven it is currently definitely broken (which I haven't heard of yet).

vittorioromeo: @binary1248: my assumption is that you meant to write 

```cpp
if (result)
    result = RenderTarget::setActive(active);
```

in order to propagate the result of the inner `setActive` call to the caller -- it's the only thing that makes sense to me, because the inner `setActive` can fail as well. My refactoring does exactly that.

Not propagating the value of the inner `setActive` is weird, because to the caller it looks like `RenderTexture::setActive` succeeded while in reality it didn't.

binary1248: That is also my thought when looking at the code now. 😁

But like I said, the current implementation that ignores the result of `RenderTexture::setActive` hasn't led to any bug reports (which doesn't have to mean anything).

I you want, we can make the change and see whether the visibly more correct logic will lead to reports.

vittorioromeo: @binary1248: I think that would be best. I will take responsibility for any bug report 😜

examples/opengl/OpenGL.cpp
binary1248: Just FYI: we already have `-Wold-style-cast` activated, see [here](https://github.com/SFML/SFML/blob/master/cmake/CompilerWarnings.cmake#L46). Like already mentioned, because it is so ubiquitous, `(void)` doesn't trigger the warning.

binary1248 (approved):

PR #1894, opened 2021-12-09
X11: fix KeyPressed duplicated with IBus

Edgaru089:
## Description

This PR is a small patch to not send `KeyPressed` events when the underlying X11 event is taken by the input method. It essentially only involves moving a `if` statement to include the part sending `KeyPressed` events, and indenting.

This fixes a bug in #1850 not discovered until @kimci86 tested it on GNOME with the bundled IM framework IBus. ([here](https://github.com/SFML/SFML/pull/1850#issuecomment-989994414))

## How to test this PR?

Dump `KeyPressed` events from a window.

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode(100, 100), "Keys");
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
                printf("KeyPressed, Key=%d\n", (int)event.key.code);
            else if (event.type == sf::Event::TextEntered)
                printf("TextEntered, Text=%d\n", (int)event.text.unicode);
        }

        window.clear();
        window.display();
    }
}
```

### DIFF
--- a/src/SFML/Window/Unix/WindowImplX11.cpp
+++ b/src/SFML/Window/Unix/WindowImplX11.cpp
@@ -34,6 +34,7 @@
 #include <SFML/System/Mutex.hpp>
 #include <SFML/System/Lock.hpp>
 #include <SFML/System/Sleep.hpp>
+#include <bitset> // <X11/Xlibint.h> defines min/max as macros, so <bitset> has to come before that
 #include <X11/Xlibint.h>
 #include <X11/Xutil.h>
 #include <X11/Xatom.h>
@@ -70,6 +71,7 @@ namespace
         std::vector<sf::priv::WindowImplX11*> allWindows;
         sf::Mutex                             allWindowsMutex;
         sf::String                            windowManagerName;
+        std::bitset<sf::Keyboard::KeyCount>   isKeyFiltered; // Maybe we should use scancodes instead of keys?
 
         sf::String                            wmAbsPosGood[] = { "Enlightenment", "FVWM", "i3" };
 
@@ -486,6 +488,45 @@ namespace
             }
 
             return sf::Keyboard::Unknown;
+        }
+
+        // Convert a XKeyEvent struct to a sf::Keyboard::Key
+        sf::Keyboard::Key xkeyToSFKey(XKeyEvent* xkey)
+        {
+            sf::Keyboard::Key key = sf::Keyboard::Unknown;
+
+            // Try each KeySym index (modifier group) until we get a match
+            for (int i = 0; i < 4; ++i)
+            {
+                // Get the SFML keyboard code from the keysym of the key that has been pressed
+                key = keysymToSF(XLookupKeysym(xkey, i));
+
+                if (key != sf::Keyboard::Unknown)
+                    break;
+            }
+
+            return key;
+        }
+
+        // Generate a KeyPressed event from a XKeyEvent struct
+        //
+        // If sfKey is not given, it calls xkeyToSFKey to try get
+        // the real sf::Keyboard::Key
+        sf::Event xkeyToKeyPressedEvent(XKeyEvent* xkey, sf::Keyboard::Key sfKey = sf::Keyboard::Unknown)
+        {
+            if(sfKey == sf::Keyboard::Unknown)
+                sfKey = xkeyToSFKey(xkey);
+
+            // Fill the event parameters
+            // TODO: if modifiers are wrong, use XGetModifierMapping to retrieve the actual modifiers mapping
+            sf::Event event;
+            event.type        = sf::Event::KeyPressed;
+            event.key.code    = sfKey;
+            event.key.alt     = xkey->state & Mod1Mask;
+            event.key.control = xkey->state & ControlMask;
+            event.key.shift   = xkey->state & ShiftMask;
+            event.key.system  = xkey->state & Mod4Mask;
+            return event;
         }
     }
 }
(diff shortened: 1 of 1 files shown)


kimci86:
I tested it with the following code to see KeyReleased events as well.
Apparently, framerate has an impact on the behavior so I added a command line argument to choose framerate.

```c++
#include <SFML/Graphics.hpp>

{
    sf::RenderWindow window(sf::VideoMode(100, 100), "Keys");
    window.setFramerateLimit(1 < argc ? std::atoi(argv[1]) : 60);

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
                printf("KeyPressed, Key=%d\n", (int)event.key.code);
            else if (event.type == sf::Event::KeyReleased)
                printf("KeyReleased, Key=%d\n", (int)event.key.code);
            else if (event.type == sf::Event::TextEntered)
                printf("TextEntered, Text=%d\n", (int)event.text.unicode);
        }

        window.clear();
        window.display();
    }
}
```

```cmake
project(test)
find_package(SFML 2.5 COMPONENTS graphics REQUIRED)
add_executable(test main.cpp)
target_link_libraries(test sfml-graphics)
install(TARGETS test DESTINATION .)
```

- [e458f4651e3cbfd168b9be0cfa328a401dccf895](https://github.com/SFML/SFML/commit/e458f4651e3cbfd168b9be0cfa328a401dccf895) (One commit before 2.6.x)
  + All good!
- [256755146990aee69c77103797dca140744f6c3a](https://github.com/SFML/SFML/commit/256755146990aee69c77103797dca140744f6c3a) (2.6.x)
  + The first KeyPressed event is duplicated
  + Several KeyPressed events after release when the key was pressed for some time and especially when FPS is low (e.g. 15).
- [d073a9b06217d411d9534d78ba4c6359c39ced05](https://github.com/SFML/SFML/pull/1894/commits/d073a9b06217d411d9534d78ba4c6359c39ced05) (Edgaru089:2.6.x)
  + I suspect the first KeyPressed event is filtered, and the second is made available, but it comes after a small delay. It can come after the KeyReleased event.
  + Several KeyPressed events after release when the key was pressed for some time and especially when FPS is low (e.g. 15).

So it does not seems to fix the issue. :thinking:

Edgaru089:
Oh no. Key and text events should be two different kinds of things...

Maybe we can keep track of which kinds of `Keyboard::Key` have been filtered away, and, for them:

- If this event is filtered by the IME: Send `KeyPressed` only
- If this event is not filtered: Send `TextEntered` only

This should not affect code discarding repeating key events.

Edgaru089:
Went digging through SDL1.2 source code. What they're doing is like:

- If a event is filtered by the IME: Send `KeyPressed` only
- If a event is not filtered:
- - If `X(utf8)LookupString()` returned both a key press and some text: send `TextEntered` only
- - If `X(utf8)LookupString()` returned only a key press: send `KeyPressed` only (there's no text anyway)

It basically assumes the IME filters every text event before re-sending it, which makes some sense.

eXpl0it3r:
Any updates on this?

Edgaru089:
> Any updates on this?

@eXpl0it3r Um, maybe I can use a `std::bitset<Keyboard::KeyCount>` to keep track on key event filtering? Or maybe scancodes are a better index so this should come after the scancode commits...

kimci86:
Thank you for you commit, but it does not seem to be a good fix.

With the test program at 25 fps (which is not extremely low in my opinion):
- A short press on A: good
  ```
  KeyPressed, Key=0
  TextEntered, Text=97
  KeyReleased, Key=0
  ```
- A short press on ² (a key on my AZERTY keyboard not in the sf::Keyboard::Key enum): repeated KeyPressed event.
  ```
  KeyPressed, Key=-1
  KeyPressed, Key=-1
  TextEntered, Text=178
  KeyReleased, Key=-1
  ```
- A long press on A: events come after a small delay. Some KeyPressed and TextEntered event arrive after KeyReleased.
  ```
  KeyPressed, Key=0
  TextEntered, Text=97
  KeyPressed, Key=0
  TextEntered, Text=97
  KeyPressed, Key=0
  KeyPressed, Key=0
  TextEntered, Text=97
  KeyPressed, Key=0
  TextEntered, Text=97
  KeyPressed, Key=0
  TextEntered, Text=97
  KeyPressed, Key=0
  KeyPressed, Key=0
  TextEntered, Text=97
  KeyPressed, Key=0
  TextEntered, Text=97
  KeyReleased, Key=0
  KeyPressed, Key=0
  TextEntered, Text=97
  TextEntered, Text=97
  TextEntered, Text=97
  ```

I don't know how it can be fixed. All I can think of for now is to check if IBus is used somehow (maybe environment variables) and fallback to previous implementation if so, but that is not really a fix and it makes the codes harder to test with more variability. But for now, key events are not reliable :face_with_head_bandage:

Edgaru089:
> A short press on ² (a key on my AZERTY keyboard not in the sf::Keyboard::Key enum): repeated KeyPressed event.

This should be resolved if we use scancodes instead of key codes, see the review.

> A long press on A: events come after a small delay. Some KeyPressed and TextEntered events arrive after KeyReleased.

I can't repeat that behaviour on my standard QWERTY, even with my keyboard layout set to AZERTY (pressing Q types A). Maybe you can add `printf()`s to the 2 places where `KeyPressed`s are emitted ([L1929](https://github.com/SFML/SFML/pull/1894/commits/8de34151331868ace49383169c12dda5d4e137ce#diff-3373daee8e5b8c4a748f9035abd44a47c467702a388e2887b095946ca6eaf6b2R1929) and [L1962](https://github.com/SFML/SFML/pull/1894/commits/8de34151331868ace49383169c12dda5d4e137ce#diff-3373daee8e5b8c4a748f9035abd44a47c467702a388e2887b095946ca6eaf6b2R1962)) and see what happens. With a text key, only the first branch should be triggered.

kimci86:
I am trying to understand and debug this. IBus seems to need some special handling. See https://github.com/libsdl-org/SDL/issues/1659

Edgaru089:
Just went through the hassle of installing a brand new Ubuntu 20.04 in a virtual machine, with GNOME and IBus bundled.

> A long press on A: events come after a small delay. Some KeyPressed and TextEntered events arrive after KeyReleased.

Well I set my keyboard layout to AZERTY, pressed and held the key on the right of Tab, and did not see any `KeyPressed` arriving before `KeyReleased` (and I guess that's the whole point). Plus, the number of `KeyPressed` I received matches the number of `TextEntered`.

`TextEntered`, yes, sometimes comes after `KeyReleased` and we might need to document that somehow, but they are text input events and they, by their very nature, should not be related to key presses in any way, nor should the user ever assume the `TextEntered` events are generated by `KeyPressed` (or at least by the ones the application actually receives).

eXpl0it3r:
IMHO doing some best-effort known key filtering, isn't really an acceptable solution. As soon as the best-effort isn't covered anymore, SFML shouldn't just send whatever combinations of pressed and released events.

Why do we get with this change get weird released/pressed event behavior?

> `TextEntered`, yes, sometimes comes after `KeyReleased` and we might need to document that somehow, but they are text input events and they, by their very nature, should not be related to key presses in any way, nor should the user ever assume the `TextEntered` events are generated by `KeyPressed` (or at least by the ones the application actually receives).

We don't guarantee the order of these events, as such this is fine and doesn't really need to be documented.

Edgaru089:
> IMHO doing some best-effort known key filtering, isn't really an acceptable solution. As soon as the best-effort isn't covered anymore, SFML shouldn't just send whatever combinations of pressed and released events.

The commit now uses a `std::bitset` to keep track of which `sf::Keyboard::Key`s are filtered at runtime, it's not known key filtering.

This does not work under keyboard layouts other than QWERTY, but once the scancode PR gets merged we can instead keep track of which scancodes are filtered, and things should work for any key pressed by a user as long as the scancodes are correct.

---

> Why do we get with this change get weird released/pressed event behavior?

Because X is bad. The client have to manually feed the events to the IM, and the IM gives text by inserting (`KeyPress`!) events into the event queue.

Without XIM, text inputs look like:
> XNextEvent() -> event, If event.type == KeyPress?
> > XLookupString() -> text input

Simple. But with XIM, things become messy:
> XNextEvent() -> event, event.type == KeyPress
> XFilterEvent(event) -> bool (manually present the event to the IM)
> - True: Key pressed, but the IM filtered it away.
> - False: The IM did not filter it away. Either the IM is not interested in this type of keypress (modifiers etc., should send `sf::KeyPressed`), or this event is sent by the IM itself as text input (should not send `sf::KeyPressed`). And no, `XLookupString()` returns a status of `XLookupBoth` if the text input has a keysym (A...Z, numbers etc.) so we cannot simply tell which one is which.

We need to send `sf::KeyPressed`s only on the original events not sent by the IM (to keep them before `sf::KeyReleased`s), so the best thing we can do is to send `sf::KeyPressed` when
- This event is filtered away, or
- This event is not filtered away, and has never been filtered before.

@binary1248 I really need your help on this.

eXpl0it3r:
Superseded by #2242

PR #1895, opened 2021-12-09
Add PCH builds via 'SFML_ENABLE_PCH' CMake option

vittorioromeo:
## Description

Enable PCH via CMake. Achieves roughly a 2x fresh compilation speedup on my machine, using MSYS2/MinGW. Looking for feedback on the direction and approach.

---

<details>
<summary>Machine Specs</summary>

- GIGABYTE Z390 AORUS MASTER-CF
- i9-9900k (stock settings)
- Corsair CMK16GX4M2B3000C15, 4x8GB sticks (XMP enabled)

</details>

---

<details>
<summary>Benchmark</summary>

Methodology:

```
cd build
rm -Rf src test examples

cmake .. -DCMAKE_BUILD_TYPE=Debug -GNinja -DSFML_BUILD_EXAMPLES=1 \ 
    -DCMAKE_EXPORT_COMPILE_COMMANDS=1 -DCMAKE_CXX_COMPILER=clang++ \
    -DCMAKE_CXX_FLAGS="-ftime-trace"

time ninja
```

Results without this PR:
- `ninja  0.00s user 0.01s system 0% cpu 5.599 total`

Results with this PR:
- `ninja  0.00s user 0.01s system 0% cpu 9.586 total`

</details>

---

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Not sure yet.

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -3,6 +3,14 @@ include(CMakeParseArguments)
 # include the compiler warnings helpers
 include(${CMAKE_CURRENT_LIST_DIR}/CompilerWarnings.cmake)
 
+# helper function to tweak visibility of public symbols
+function(set_public_symbols_hidden target)
+    # ensure public symbols are hidden by default (exported ones are explicitly marked)
+    set_target_properties(${target} PROPERTIES
+                          CXX_VISIBILITY_PRESET hidden
+                          VISIBILITY_INLINES_HIDDEN YES)
+endfunction()
+
 # This little macro lets you set any Xcode specific property
 macro(sfml_set_xcode_property TARGET XCODE_PROPERTY XCODE_VALUE)
     set_property (TARGET ${TARGET} PROPERTY XCODE_ATTRIBUTE_${XCODE_PROPERTY} ${XCODE_VALUE})
@@ -80,6 +88,13 @@ macro(sfml_add_library module)
     endif()
 
     set_target_warnings(${target})
+    set_public_symbols_hidden(${target})
+
+    # enable precompiled headers
+    if (SFML_ENABLE_PCH AND (NOT ${target} STREQUAL "sfml-system"))
+        message(VERBOSE "enabling PCH for SFML library '${target}'")
+        target_precompile_headers(${target} REUSE_FROM sfml-system)
+    endif()
 
     # define the export symbol of the module
     string(REPLACE "-" "_" NAME_UPPER "${target}")
@@ -164,17 +179,21 @@ macro(sfml_add_library module)
                                   PDB_NAME "${target}${SFML_PDB_POSTFIX}"
                                   PDB_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/lib")
         else()
-            # Static libraries have no linker PDBs, thus the compiler PDBs are relevant
-            set_target_properties(${target} PROPERTIES
-                                  COMPILE_PDB_NAME "${target}-s${SFML_PDB_POSTFIX}"
-                                  COMPILE_PDB_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/lib")
+            if(SFML_ENABLE_PCH)
+                message(VERBOSE "overriding PDB name for '${target}' with \"sfml-s${SFML_PDB_POSTFIX}\" due to PCH being enabled")
+
+                # For PCH builds with PCH reuse, the PDB name must be the same as the target that's being reused
+                set_target_properties(${target} PROPERTIES
+                                      COMPILE_PDB_NAME "sfml-system"
+                                      COMPILE_PDB_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/lib")
+            else()
+                # Static libraries have no linker PDBs, thus the compiler PDBs are relevant
+                set_target_properties(${target} PROPERTIES
+                                      COMPILE_PDB_NAME "${target}-s${SFML_PDB_POSTFIX}"
+                                      COMPILE_PDB_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/lib")
+            endif()
         endif()
     endif()
-
-    # ensure public symbols are hidden by default (exported ones are explicitly marked)
-    set_target_properties(${target} PROPERTIES
-                          CXX_VISIBILITY_PRESET hidden
-                          VISIBILITY_INLINES_HIDDEN YES)
 
     # build frameworks or dylibs
     if(SFML_OS_MACOSX AND BUILD_SHARED_LIBS AND NOT THIS_STATIC)
@@ -285,7 +304,14 @@ macro(sfml_add_example target)
         set_property(TARGET ${target} PROPERTY MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
     endif()
 
+    # enable precompiled headers
+    if (SFML_ENABLE_PCH)
+        message(VERBOSE "enabling PCH for SFML example '${target}'")
+        target_precompile_headers(${target} REUSE_FROM sfml-system)
+    endif()
+
     set_target_warnings(${target})
+    set_public_symbols_hidden(${target})
 
     # set the debug suffix
     set_target_properties(${target} PROPERTIES DEBUG_POSTFIX -d)
@@ -322,6 +348,12 @@ function(sfml_add_test target SOURCES DEPENDS)
     # create the target
     add_executable(${target} ${SOURCES})
 
+    # enable precompiled headers
+    if (SFML_ENABLE_PCH)
+        message(VERBOSE "enabling PCH for SFML test '${target}'")
+        target_precompile_headers(${target} REUSE_FROM sfml-system)
+    endif()
+
     # set the target's folder (for IDEs that support it, e.g. Visual Studio)
     set_target_properties(${target} PROPERTIES FOLDER "Tests")
 
@@ -329,6 +361,7 @@ function(sfml_add_test target SOURCES DEPENDS)
     target_link_libraries(${target} PRIVATE ${DEPENDS} sfml-test-main)
 
     set_target_warnings(${target})
+    set_public_symbols_hidden(${target})
 
     # If coverage is enabled for MSVC and we are linking statically, use /WHOLEARCHIVE
     # to make sure the linker doesn't discard unused code sections before coverage can be measured
--- a/src/SFML/PCH.hpp
+++ b/src/SFML/PCH.hpp
@@ -1,0 +1,62 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2022 Laurent Gomila (laurent@sfml-dev.org)
+//
+// This software is provided 'as-is', without any express or implied warranty.
+// In no event will the authors be held liable for any damages arising from the use of this software.
+//
+// Permission is granted to anyone to use this software for any purpose,
+// including commercial applications, and to alter it and redistribute it freely,
+// subject to the following restrictions:
+//
+// 1. The origin of this software must not be misrepresented;
+//    you must not claim that you wrote the original software.
+//    If you use this software in a product, an acknowledgment
+//    in the product documentation would be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such,
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+#ifndef SFML_SFML_PCH_HPP
+#define SFML_SFML_PCH_HPP
+
+////////////////////////////////////////////////////////////
+// Precompiled Headers
+////////////////////////////////////////////////////////////
+
+#include <SFML/Config.hpp>
+
+#ifdef SFML_SYSTEM_WINDOWS
+
+#define UNICODE  1
+#define _UNICODE 1
+#include <SFML/System/Win32/WindowsHeader.hpp>
+
+#endif // SFML_SYSTEM_WINDOWS
+
+#include <SFML/System/Err.hpp>
+#include <SFML/System/String.hpp>
+#include <SFML/System/Time.hpp>
+#include <SFML/System/Vector2.hpp>
+
+#include <algorithm>
+#include <cassert>
+#include <chrono>
+#include <cmath>
+#include <cstdint>
+#include <cstring>
+#include <filesystem>
+#include <iostream>
+#include <memory>
+#include <mutex>
+#include <ostream>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#endif // SFML_SFML_PCH_HPP
--- a/src/SFML/System/CMakeLists.txt
+++ b/src/SFML/System/CMakeLists.txt
@@ -67,6 +67,12 @@ endif()
 sfml_add_library(System
                  SOURCES ${SRC} ${PLATFORM_SRC})
 
+# enable precompiled headers
+if (SFML_ENABLE_PCH)
+    message(VERBOSE "enabling PCH for SFML library 'sfml-system' (reused as the PCH for other SFML libraries)")
+    target_precompile_headers(sfml-system PRIVATE ${PROJECT_SOURCE_DIR}/src/SFML/PCH.hpp)
+endif()
+
 if(SFML_OS_ANDROID)
     # glad sources
     target_include_directories(sfml-system PRIVATE "${PROJECT_SOURCE_DIR}/extlibs/headers/glad/include")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -288,6 +288,17 @@ if(SFML_INSTALL_PKGCONFIG_FILES)
     endforeach()
 endif()
 
+# option to enable precompiled headers
+sfml_set_option(SFML_ENABLE_PCH FALSE BOOL "TRUE to enable precompiled headers for SFML builds -- only supported on Windows/Linux and for static library builds")
+
+if(SFML_ENABLE_PCH AND BUILD_SHARED_LIBS)
+    message(FATAL_ERROR "Precompiled headers are currently not supported for shared library builds")
+endif()
+
+if(SFML_ENABLE_PCH AND SFML_OS_MACOSX)
+    message(FATAL_ERROR "Precompiled headers are currently not supported in MacOS builds")
+endif()
+
 # setup the install rules
 if(NOT SFML_BUILD_FRAMEWORKS)
     install(DIRECTORY include


vittorioromeo:
@binary1248: any idea why CI is unhappy with the changes? The GCC error is particularly weird...

binary1248:
I don't have any experience using PCH, definitely not with CMake.

sam20908:
Did you ensure that the targets reusing the PCH have the same compiler definitions, options, and flags?

Quoted from CMake documentation:
> The REUSE_FROM form requires the same set of compiler options, compiler flags and compiler definitions for both <target> and <other_target>. Some compilers (e.g. GCC) may issue a warning if the precompiled header file cannot be used (-Winvalid-pch).

vittorioromeo:
> Did you ensure that the targets reusing the PCH have the same compiler definitions, options, and flags?
> 
> Quoted from CMake documentation:
> 
> > The REUSE_FROM form requires the same set of compiler options, compiler flags and compiler definitions for both  and <other_target>. Some compilers (e.g. GCC) may issue a warning if the precompiled header file cannot be used (-Winvalid-pch).

There are no targets reusing the PCH, I've commented all of that out... The CI still fails with cryptic errors even when enabling PCH just for `sfml-system`.

vittorioromeo:
Hah, yes! That was the issue. Removing `-Wpedantic` from `CompilerWarnings.cmake` makes PCH work! I don't understand why though, as we do call `set_file_warnings` on the PCH header. Does anybody want to investigate this further? @ChrisThrasher

vittorioromeo:
I solved with 

```cmake
target_precompile_headers(sfml-system PRIVATE "${CMAKE_SOURCE_DIR}/src/SFML/PCH.hpp")
set_file_warnings("${CMAKE_BINARY_DIR}/src/SFML/System/CMakeFiles/sfml-system.dir/cmake_pch.hxx.cxx")
```

but it is really ugly...

ChrisThrasher:
Precompiled headers require CMake 3.16. I see this PR doesn't change the minimum CMake version requirement. Is it possible that some jobs are failing due to old CMake versions?

vittorioromeo:
At this point in time, PCH files seem to work only for Linux and MinGW, and only when SFML is built as a static library. For Visual Studio, the issue is:

```
CMake Error at cmake/Macros.cmake:[69](https://github.com/SFML/SFML/actions/runs/3165984825/jobs/5155401526#step:8:70) (add_library):
  PRECOMPILE_HEADERS_REUSE_FROM property is set on target ("sfml-window").
  Reusable precompile headers requires the COMPILE_PDB_NAME property to have
  the value "sfml-system"

  src/SFML/Window/CMakeLists.txt:255 (sfml_add_library)
```

So to make it work, `COMPILE_PDB_NAME` has to be the same as the PCH target we are reusing the PCH from.

Regarding shared library builds: I cannot make PCH reuse work because of differing macro defines for `SFML_XXX_EXPORTS`. All targets must have the same macros. I can try having a per-library PCH, but I'm not sure if it's worth it.

vittorioromeo:
I have no idea what the MacOS issue is about:

```
CMake Error in src/SFML/Window/CMakeLists.txt:
  Unable to resolve full path of PCH-header
  '/Users/runner/work/SFML/SFML/build/src/SFML/System/CMakeFiles/sfml-system.dir/cmake_pch.objc.h'
  assigned to target sfml-window, although its path is supposed to be known!
```

vittorioromeo:
I have no idea what the MacOS issue is about:

```
CMake Error in src/SFML/Window/CMakeLists.txt:
  Unable to resolve full path of PCH-header
  '/Users/runner/work/SFML/SFML/build/src/SFML/System/CMakeFiles/sfml-system.dir/cmake_pch.objc.h'
  assigned to target sfml-window, although its path is supposed to be known!
```

vittorioromeo:
@ChrisThrasher @eXpl0it3r @binary1248 @Bromeon please opine. The speedup on MinGW/Linux is quite large, a fresh compile is at minimum 2x faster on my machine. A few points:

- PCH builds are disabled by default. Can be enabled with the `SFML_ENABLE_PCH` option.

- PCH builds are only supported on Windows/Linux. MacOS is not working for some weird reason and I cannot debug it easily.

- PCH builds are only supported for static builds of SFML. Shared library support is not worth it because the PCH cannot be reused due to mismatching macro definitions of `SFML_XXX_EXPORTS` between targets.

- It would be nice to have "static + PCH" CI jobs for Linux, MinGW, and MSVC.

cmake/Macros.cmake
```diff
@@ -281,11 +300,21 @@ macro(sfml_add_example target)
         add_executable(${target} ${target_input})
     endif()
 
+    # enable C++17 support
+    target_compile_features(${target} PUBLIC cxx_std_17)
```
ChrisThrasher, line 304: Why add this? It's redundant since this requirement is already picked up when these targets depend on `SFML::` targets. I removed this a PR back in like February.

cmake/Macros.cmake
```diff
@@ -281,11 +300,21 @@ macro(sfml_add_example target)
...
     endif()
 
+    # enable precompiled headers
+    if (SFML_ENABLE_PCH)
+        message(VERBOSE "enabling PCH for SFML example '${target}'")
+        target_precompile_headers(${target} REUSE_FROM sfml-system)
+    endif()
+
     set_target_warnings(${target})
+    set_public_symbols_hidden(${target})
```
ChrisThrasher, line 317: Why do we need to hide symbols in executable targets? I thought that was only meaningful for library.

cmake/Macros.cmake
```diff
@@ -322,13 +351,23 @@ function(sfml_add_test target SOURCES DEPENDS)
     # create the target
     add_executable(${target} ${SOURCES})
 
+    # enable C++17 support
+    target_compile_features(${target} PUBLIC cxx_std_17)
```
ChrisThrasher, line 355: See above comment about this being redundant.

ChrisThrasher (changes requested):
In the ideal case, is it possible to support pre-compiled headers on all platforms or does CMake simply not allow that?

What do you think about automatically enabling PCH where possible so that we're not expanding the configuration space of the library? That would let us stop worrying about adding CI jobs and the exponential (polynomial?) expansion of jobs that entails.

cmake/Macros.cmake
vittorioromeo: PCH reuse requires the flags to be exactly the same for all projects using the same PCH.

cmake/Macros.cmake
vittorioromeo: Will try removing, might be unnecessary.

src/SFML/PCH.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+#ifdef SFML_SYSTEM_WINDOWS
+
+#define UNICODE  1
+#define _UNICODE 1
+#include <SFML/System/Win32/WindowsHeader.hpp>
+
+#endif // SFML_SYSTEM_WINDOWS
+
+#include <SFML/Graphics/Rect.hpp>
```
ChrisThrasher, line 42: What happens if somebody doesn't build the Graphics module? Will this still work since all headers are accessible no matter whether or not all modules are built?

vittorioromeo: I will circumvent the issue right now by removing this line.

ChrisThrasher:
Here are my results testing this PR. I'm using an Ubuntu 22 VM running in macOS with 6 CPUs and 16 GB of memory.

---

`master` (939ec267dfa98e97438a5b76f79317c7e0738a92)
```
cmake -B build -DSFML_BUILD_EXAMPLES=ON -DSFML_BUILD_TEST_SUITE=ON -DBUILD_SHARED_LIBS=OFF
time cmake --build build -j6
```

```
Total Test time (real) =   0.35 sec
[100%] Built target runtests

real	0m13.779s
user	1m1.117s
sys	0m6.092s
```

---

`feature/use-pch` (ec0a03610d22d55fe0ed4916bf0c1dbb9076352e)
```
cmake -B build -DSFML_BUILD_EXAMPLES=ON -DSFML_BUILD_TEST_SUITE=ON -DBUILD_SHARED_LIBS=OFF -DSFML_ENABLE_PCH=ON
time cmake --build build -j6
```

```
Total Test time (real) =   0.35 sec
[100%] Built target runtests

real    0m12.228s
user    0m46.263s
sys    0m7.278s
```

vittorioromeo:
My results with MSYS2/MinGW on my Windows 10 x64 laptop:

```
cmake .. -GNinja -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CXX_FLAGS="-ftime-trace" -DSFML_BUILD_TEST_SUITE=1 -DSFML_BUILD_EXAMPLES=1 -DSFML_BUILD_DOC=0 -DBUILD_SHARED_LIBS=0 -DSFML_ENABLE_PCH=0
```

```
time ninja
ninja  0.01s user 0.00s system 0% cpu 31.183 total
```

---

```
cmake .. -GNinja -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CXX_FLAGS="-ftime-trace" -DSFML_BUILD_TEST_SUITE=1 -DSFML_BUILD_EXAMPLES=1 -DSFML_BUILD_DOC=0 -DBUILD_SHARED_LIBS=0 -DSFML_ENABLE_PCH=1
```

```
time ninja
ninja  0.00s user 0.01s system 0% cpu 22.085 total
```

src/SFML/PCH.hpp
```diff
@@ -0,0 +1,62 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2021 Laurent Gomila (laurent@sfml-dev.org)
```
Bromeon, line 4: Should be 2022.
Likely also affects the source from which the license header was copy-pasted.

src/SFML/System/CMakeLists.txt
```diff
@@ -67,6 +67,12 @@ endif()
 sfml_add_library(System
                  SOURCES ${SRC} ${PLATFORM_SRC})
 
+# enable precompiled headers
+if (SFML_ENABLE_PCH)
+    message(VERBOSE "enabling PCH for SFML library 'sfml-system' (reusable)")
```
Bromeon, line 72: What is meant with `(reusable)` here? Is there an easy way to clarify the message?

cmake/Macros.cmake
```diff
@@ -80,6 +88,13 @@ macro(sfml_add_library module)
     endif()
 
     set_target_warnings(${target})
+    set_public_symbols_hidden(${target})
+
+    # enable precompiled headers
+    if (SFML_ENABLE_PCH AND (NOT ${target} STREQUAL "sfml-system"))
+        message(VERBOSE "enabling PCH for SFML library '${target}'")
+        target_precompile_headers(${target} REUSE_FROM sfml-system)
+    endif()
```
Bromeon, line 97: Small nitpick: `if(` doesn't have space in other places.

I like how you log the use of PCHs thoroughly, so that users can see what exactly happens!

Bromeon (commented):
This is a very nice addition to SFML, thanks a lot for this PR! 👍

I'm not sure if we should switch CI over to use PCH=on; probably a lot of SFML users would still use the classic setup, and I think it's important we have the confidence that the basic configuration is not broken.

The combinatoric growth is a problem; there are a few options though:
* simply add more jobs
   * as long as we can still run them in parallel, it shouldn't take much longer (not sure about initial PCH compile times)
* run CI with PCH=on only on rare periodic intervals
   * not a fan of this, as it makes it hard to link it to changes
* only run PCH when CMake is changed
   * C++ code ideally shouldn't break PCH. would however need quite some extra GitHub Actions logic

vittorioromeo:
> This is a very nice addition to SFML, thanks a lot for this PR! 👍

Thanks!

> I'm not sure if we should switch CI over to use PCH=on; probably a lot of SFML users would still use the classic setup, and I think it's important we have the confidence that the basic configuration is not broken.

Honestly, this feature is mostly to speed up builds for people working on SFML's codebase itself. Even if it's never consumed by our users, it's still beneficial for us to have as part of our development cycle, and potentially to speed up CI.

My personal preference would be to enable PCH=on on all our CI except one job that tests a non-PCH build just to make sure it still works properly -- one of the reasons to have PCH is to speed up the whole CI process.

Happy to hear what others think.

Bromeon:
> My personal preference would be to enable PCH=on on all our CI except one job that tests a non-PCH build just to make sure it still works properly -- one of the reasons to have PCH is to speed up the whole CI process.

That sounds reasonable to me. 👍

vittorioromeo:
> I assume miscompilations due to PCH are no longer the dreadful thing they were 1-2 decades ago? 😄

My personal experience with PCH has been very positive in this aspect, I never had any miscompilation issue. Also, as you can see by the pain I went through, any sort of inconsistency between files causes a total compilation failure -- PCH needs every flag to be exactly the same in order to avoid weird issues. I think we should be pretty safe!

vittorioromeo:
I would suggest merging this PR as-is, then having a separate PR to enable PCH in CI.

src/SFML/PCH.hpp
```diff
@@ -0,0 +1,62 @@
...
+////////////////////////////////////////////////////////////
+// Precompiled Headers
+////////////////////////////////////////////////////////////
+
+#include <SFML/Config.hpp>
+
+#ifdef SFML_SYSTEM_WINDOWS
+
+#define UNICODE  1
+#define _UNICODE 1
```
Bromeon, line 37: This cannot lead to issues somewhere down the line? Or should they be `#undef`ed after including the Windows headers?

Bromeon (commented):
One small question, other than that good to go from my side!

src/SFML/PCH.hpp
vittorioromeo: This matches the same definitions in the CMakeLists, which are never undefined. It should not cause any issue:

```cmake
if(SFML_OS_WINDOWS)
    set(PLATFORM_SRC
        ${SRCROOT}/Win32/CursorImpl.hpp
        ...
    )
    source_group("windows" FILES ${PLATFORM_SRC})

    # make sure that we use the Unicode version of the Win API functions
    add_definitions(-DUNICODE -D_UNICODE)
```

vittorioromeo: @Bromeon let me know if you're happy with that before merging :)

Bromeon: Ok, fine with me in that case! 🚀

cmake/Macros.cmake
```diff
@@ -164,18 +179,22 @@ macro(sfml_add_library module)
...
-            set_target_properties(${target} PROPERTIES
-                                  COMPILE_PDB_NAME "${target}-s${SFML_PDB_POSTFIX}"
-                                  COMPILE_PDB_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/lib")
+            if(SFML_ENABLE_PCH)
+                message(VERBOSE "overriding PDB name for '${target}' with \"sfml-s${SFML_PDB_POSTFIX}\" due to PCH being enabled")
+
+                # For PCH builds with PCH reuse, the PDB name must be the same as the target that's being reused
+                set_target_properties(${target} PROPERTIES
+                                    COMPILE_PDB_NAME "sfml-system"
+                                    COMPILE_PDB_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/lib")
```
ChrisThrasher, line 188:
```suggestion
                                      COMPILE_PDB_NAME "sfml-system"
                                      COMPILE_PDB_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/lib")
```

cmake/Macros.cmake
```diff
@@ -164,18 +179,22 @@ macro(sfml_add_library module)
...
+            else()
+                # Static libraries have no linker PDBs, thus the compiler PDBs are relevant
+                set_target_properties(${target} PROPERTIES
+                                    COMPILE_PDB_NAME "${target}-s${SFML_PDB_POSTFIX}"
+                                    COMPILE_PDB_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/lib")
```
ChrisThrasher, line 193:
```suggestion
                                      COMPILE_PDB_NAME "${target}-s${SFML_PDB_POSTFIX}"
                                      COMPILE_PDB_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/lib")
```

src/SFML/System/CMakeLists.txt
```diff
@@ -67,6 +67,12 @@ endif()
 sfml_add_library(System
                  SOURCES ${SRC} ${PLATFORM_SRC})
 
+# enable precompiled headers
+if (SFML_ENABLE_PCH)
+    message(VERBOSE "enabling PCH for SFML library 'sfml-system' (reused as the PCH for other SFML libraries)")
+    target_precompile_headers(sfml-system PRIVATE "${CMAKE_SOURCE_DIR}/src/SFML/PCH.hpp")
```
ChrisThrasher, line 73:
```suggestion
    target_precompile_headers(sfml-system PRIVATE ${PROJECT_SOURCE_DIR}/src/SFML/PCH.hpp)
```
`PROJECT_SOURCE_DIR` ensures that it's always starting from the root of SFML and not the root of whatever project may be including SFML via `add_subdirectory`.

ChrisThrasher (changes requested):
Can you share some stats for how this changes total CI runtime? Can we expect the pipeline to complete any quicker than before?

vittorioromeo:
@ChrisThrasher 

> Can you share some stats for how this changes total CI runtime? Can we expect the pipeline to complete any quicker than before?

I do not have any CI stats at the moment -- the goal of this PR is to add the PCH mechanism to SFML so that we can see how much of an improvement it is. I do expect a significant speedup for all Windows and Linux static library CI builds.

I have addressed your suggestions (thanks!) -- my suggestion would be to merge this PR, create a PR to enable PCH builds where appropriate in CI, and get some rough measurements. If there's no benefit, I'd be happy to revert this PR.

vittorioromeo:

ChrisThrasher (approved):

eXpl0it3r:
Thanks for this! 🙂  
As for the next steps, I'd suggest to select some builds. If we just do all or nothing, we're blind sighting ourselves in either direction. One could also consider enabling it on say GitHub Actions and leave it as is for Buildbot.

JonnyPtn:
What's the difference in performance if you don't reuse the PCH? I'm fairly sure that's the cause of the issues. So if we avoid doing that we can cover all platforms and build configurations?

Also solves the problem of adding more configuration - the typical way I've done this is to enable PCH by default then users can disable using `CMAKE_DISABLE_PRECOMPILE_HEADERS` without us having to do anything extra

vittorioromeo:
@JonnyPtn 

All the following results are on Windows 10 x64, clean static library build, build both tests and examples as well, timed via `time ninja`.

```bash
# no PCH at all
ninja  0.01s user 0.00s system 0% cpu 34.032 total # baseline

# PCH with reuse (this PR)
ninja  0.01s user 0.03s system 0% cpu 19.381 total # 15s improvement!

# PCH without reuse  
ninja  0.01s user 0.00s system 0% cpu 28.787 total # 6s improvement :/
```

So as you can see, reusing is quite significant, but without reusing we do still get a minor improvement. Maybe we could automatically pick reuse or no-reuse depending on whether we are building static or shared?

Also, I had to hack in `include_directories(${PROJECT_SOURCE_DIR}/src)` to make the `PCH.hpp` file find `WindowsHeader.hpp`. Not sure how to make that work cleanly, because that header is under `System`'s "private" headers.

PR #1896, opened 2021-12-10
Reduce compile-time impact of 'windows.h'

vittorioromeo:
## Description

This PR reduces the compile-time impact of `windows.h` by isolating it a bit more (e.g. moving it from a header to a source file) and by extensively using the `WIN32_LEAN_AND_MEAN` preprocessor definition. 

The results are very good: a clean rebuild on MSYS2/MinGW used to spend around ~39s (in total) dealing with `windows.h`. After this PR, a clean rebuild only spends ~25s on the same task. This was measured via [ClangBuildAnalyzer](https://github.com/aras-p/ClangBuildAnalyzer).

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.

### DIFF
--- a/examples/win32/Win32.cpp
+++ b/examples/win32/Win32.cpp
@@ -3,6 +3,9 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Graphics.hpp>
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
 #include <windows.h>
 #include <cmath>
 
--- a/src/SFML/Network/Win32/SocketImpl.hpp
+++ b/src/SFML/Network/Win32/SocketImpl.hpp
@@ -28,15 +28,8 @@
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
-#ifdef _WIN32_WINDOWS
-    #undef _WIN32_WINDOWS
-#endif
-#ifdef _WIN32_WINNT
-    #undef _WIN32_WINNT
-#endif
-#define _WIN32_WINDOWS 0x0501
-#define _WIN32_WINNT   0x0501
 #include <SFML/Network/Socket.hpp>
+#include <SFML/System/Win32/WindowsHeader.hpp>
 #include <winsock2.h>
 #include <ws2tcpip.h>
 
--- a/src/SFML/Window/Win32/CursorImpl.hpp
+++ b/src/SFML/Window/Win32/CursorImpl.hpp
@@ -32,8 +32,6 @@
 #include <SFML/System/NonCopyable.hpp>
 #include <SFML/System/Vector2.hpp>
 
-#include <windows.h>
-
 namespace sf
 {
 
@@ -92,8 +90,8 @@ private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    HCURSOR m_cursor;
-    bool m_systemCursor;
+    void* m_cursor; // Type erasure via `void*` is used here to avoid depending on `windows.h`
+    bool  m_systemCursor;
 };
 
 } // namespace priv
--- a/src/SFML/Window/WindowImpl.hpp
+++ b/src/SFML/Window/WindowImpl.hpp
@@ -35,7 +35,6 @@
 #include <SFML/Window/CursorImpl.hpp>
 #include <SFML/Window/Event.hpp>
 #include <SFML/Window/Joystick.hpp>
-#include <SFML/Window/JoystickImpl.hpp>
 #include <SFML/Window/Sensor.hpp>
 #include <SFML/Window/SensorImpl.hpp>
 #include <SFML/Window/VideoMode.hpp>
@@ -266,6 +265,7 @@ protected:
     virtual void processEvents() = 0;
 
 private:
+    struct JoystickStatesImpl;
 
     ////////////////////////////////////////////////////////////
     /// \brief Read the joysticks state and generate the appropriate events
@@ -282,11 +282,11 @@ private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    std::queue<Event> m_events;                                              //!< Queue of available events
-    JoystickState     m_joystickStates[Joystick::Count];                     //!< Previous state of the joysticks
-    Vector3f          m_sensorValue[Sensor::Count];                          //!< Previous value of the sensors
-    float             m_joystickThreshold;                                   //!< Joystick threshold (minimum motion for "move" event to be generated)
-    float             m_previousAxes[Joystick::Count][Joystick::AxisCount];  //!< Position of each axis last time a move event triggered, in range [-100, 100]
+    std::queue<Event>   m_events;                                              //!< Queue of available events
+    JoystickStatesImpl* m_joystickStatesImpl;                                  //!< Previous state of the joysticks (PImpl)
+    Vector3f            m_sensorValue[Sensor::Count];                          //!< Previous value of the sensors
+    float               m_joystickThreshold;                                   //!< Joystick threshold (minimum motion for "move" event to be generated)
+    float               m_previousAxes[Joystick::Count][Joystick::AxisCount];  //!< Position of each axis last time a move event triggered, in range [-100, 100]
 };
 
 } // namespace priv
--- a/include/SFML/OpenGL.hpp
+++ b/include/SFML/OpenGL.hpp
@@ -40,6 +40,9 @@
 
     // The Visual C++ version of gl.h uses WINGDIAPI and APIENTRY but doesn't define them
     #ifdef _MSC_VER
+        #ifndef WIN32_LEAN_AND_MEAN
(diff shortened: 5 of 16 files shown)


src/SFML/Window/WindowImpl.hpp
```diff
@@ -282,11 +282,11 @@ class WindowImpl : NonCopyable
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    std::queue<Event> m_events;                                              //!< Queue of available events
-    JoystickState     m_joystickStates[Joystick::Count];                     //!< Previous state of the joysticks
-    Vector3f          m_sensorValue[Sensor::Count];                          //!< Previous value of the sensors
-    float             m_joystickThreshold;                                   //!< Joystick threshold (minimum motion for "move" event to be generated)
-    float             m_previousAxes[Joystick::Count][Joystick::AxisCount];  //!< Position of each axis last time a move event triggered, in range [-100, 100]
+    std::queue<Event>   m_events;                                              //!< Queue of available events
+    JoystickStatesImpl* m_joystickStatesImpl;                                  //!< Previous state of the joysticks (PImpl)
```
binary1248, line 286: Is there a reason this can't be a `std::unique_ptr`? As long as we make sure the destructor implementation lives in the `.cpp` the compiler shouldn't complain about incomplete types.

examples/win32/Win32.cpp
```diff
@@ -3,7 +3,12 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Graphics.hpp>
+
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
```
binary1248, line 9: Is there any advantage of adding this block to all these files instead of just adding it as a compile definition in the CMake configuration? As long as it is declared `PRIVATE` in CMake it should only have an effect when building the library and should not have any effects on user code.

binary1248 (changes requested):

src/SFML/Window/WindowImpl.hpp
vittorioromeo: I wanted to create a PR that turns every `new`/`delete` into `std::unique_ptr` at once, in order to have a consistent codebase.

examples/win32/Win32.cpp
vittorioromeo: There is one place where we cannot use this macro, `SleepImpl.hpp`. I think also the Network module might have a place like that as well. My personal preference would be creating something like `SFML/Priv/WindowsLeanAndMean.hpp` and stick all the repetition there, then just include that header instead. But we don't have a precedent for that. Let me know what you think.

Bromeon: SleepImpl.hpp is being removed in https://github.com/SFML/SFML/pull/1872. o in case there's nothing additional in the Network module, having it in CMake should be OK..

cmake/Macros.cmake
```diff
@@ -47,6 +47,20 @@ function(sfml_set_common_ios_properties target)
     endif()
 endfunction()
 
+# set 'windows.h'-related definitions
+function(sfml_set_windowsh_definitions target)
```
eXpl0it3r, line 51: windowsh looks a bit odd as a term, maybe windows_header?

src/SFML/Window/Win32/CursorImpl.hpp
```diff
@@ -92,8 +90,8 @@ class CursorImpl : NonCopyable
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    HCURSOR m_cursor;
-    bool m_systemCursor;
+    void* m_cursor;
```
eXpl0it3r, line 93: Might want to add a comment to justify the usage of a void* here, so it's not removed in the next refactoring

vittorioromeo: Will do

eXpl0it3r (approved):

vittorioromeo:
@binary1248: I have addressed your comments, could you please approve if you are happy?

include/SFML/Config.hpp
```diff
@@ -43,8 +43,22 @@
...
+    #endif
+
+    #if !defined(_WIN32_WINDOWS) || !defined(_WIN32_WINNT) || !defined(WINVER) || \
+        (_WIN32_WINDOWS != 0x0501) || (_WIN32_WINNT != 0x0501) || (WINVER != 0x0501)
+        #error Both '_WIN32_WINDOWS' and '_WIN32_WINNT' must be defined '0x0501'
+    #endif
+
+    #if !defined(DIRECTINPUT_VERSION) || (DIRECTINPUT_VERSION != 0x0800)
+        #error 'DIRECTINPUT_VERSION' must be defined as '0x0800'
     #endif
```
binary1248, line 62: Is this really required? This file is part of the public interface and is directly/indirectly included by any file as soon as the user touches an SFML header in their code. All these definitions just affect which parts of the Windows headers are active. The nice thing about the Windows API is that it is a C API which means that we don't have to be too careful about mixing different API versions between SFML and user code. Microsoft has made sure that any additions are backwards compatible.

If you put yourself into the shoes of the user, I would find these compiler errors confusing/annoying, especially for a beginner who had no intention of doing Windows programming in their own code (that's what they chose SFML for). Just including an SFML header will almost always generate these 4 errors for them if they don't explicitly add these definitions to their project.

If the goal was to ensure that we build SFML with the same definitions everywhere then this has to be done in an internal header that is not exposed via the public interface.

binary1248 (changes requested):

include/SFML/Config.hpp
vittorioromeo: I understand your concern, and I believe you're right.

> If the goal was to ensure that we build SFML with the same definitions everywhere then this has to be done in an internal header that is not exposed via the public interface.

That was indeed the goal. How I should I call that internal header and where should I put it? Can I create a `Priv` subfolder under `SFML`?

binary1248: Maybe just dump it into src/SFML? 😁

When it comes to such things, I am not the right guy to ask.

@SFML/sfml

vittorioromeo: I would actually like to have a `Priv` subfolder for implementation-only headers. It would be extremely useful when trying to reuse small utilities between different SFML components. I am actually surprised there is nothing like that, and I've seen some repetition of internal utilities due to the lack of such structure.

How do you feel about `SFML/Priv/WindowsHeader.h`? And then we could include that instead of `<windows.h>` whenever required.

binary1248: `<windows.h>` is a special unicorn. Unlike the Posix headers where you can include what you use (and nothing more) the general expectation when developing for Windows is that you just `#include <windows.h>` and that is it. This is also the reason why they have tried (e.g. WIN32_LEAN_AND_MEAN) to retroactively fix the problem they created, a problem that does not exist on other platforms. I am just a bit concerned that the contents of this directory would be very Windows heavy and there would not be analogous files for other platforms.

vittorioromeo: @binary1248: let's discuss on Slack please.

src/SFML/Network/Win32/SocketImpl.hpp
```diff
@@ -28,15 +28,8 @@
...
-#ifdef _WIN32_WINDOWS
-    #undef _WIN32_WINDOWS
-#endif
-#ifdef _WIN32_WINNT
-    #undef _WIN32_WINNT
-#endif
-#define _WIN32_WINDOWS 0x0501
-#define _WIN32_WINNT   0x0501
 #include <SFML/Network/Socket.hpp>
+#include <SFML/Window/Win32/WindowsHeader.hpp>
```
kimci86, line 32: Wouldn't it make more sense to have the WindowsHeader.hpp file in the System module? Here the Network module depends on the Window module. It is just a header so it woks fine, but is seems odd to me.

eXpl0it3r: You're right, it's an invalid dependency (the Network module should not depend on the Window module), I thought it was the Window module only.

Bromeon (commented):
Looks good to me; most important is that `<windows.h>` is not unnecessarily included in user code, but this seems to be the case here.

src/SFML/Network/Win32/SocketImpl.hpp
vittorioromeo: OK, I will move it to `System/`.

vittorioromeo: I have moved the header to `System/Win32`.

vittorioromeo: @binary1248: your approval is still required for this PR

kimci86: You removed WindowsHeader.hpp instead of moving it in your last commit.

eXpl0it3r (approved):

binary1248 (approved):

PR #1901, opened 2021-12-14
Replace 'sf::NonCopyable' with '= delete'

vittorioromeo:
## Description

Removes `sf::NonCopyable` and replaces existing usages with `= delete`. Note that some of these classes might be made *movable*, but that's for a separate PR.

[Forum thread](https://en.sfml-dev.org/forums/index.php?topic=21781.0

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.


include/SFML/Audio/InputSoundFile.hpp
```diff
@@ -60,6 +59,13 @@ class SFML_AUDIO_API InputSoundFile : NonCopyable
     ////////////////////////////////////////////////////////////
     ~InputSoundFile();
 
+    ////////////////////////////////////////////////////////////
+    /// Prevent copies.
```
kimci86, line 63: Doxygen will interpret that as documentation for `InputSoundFile(const InputSoundFile&)` and `InputSoundFile& operator=(const InputSoundFile&)` will not be documented.

vittorioromeo: How do you suggest to fix this? I am not an expert on Doxygen. Maybe just change this to a normal comment (i.e. `//`)?

kimci86: Either change to normal comment `//` or document each method separately e.g. (in SFML style):
```cpp
/// \brief Disabled copy constructor
///
InputSoundFile(const InputSoundFile&) = delete;

/// \brief Disabled assignment operator
///
InputSoundFile& operator=(const InputSoundFile&) = delete;
```

src/SFML/Graphics/ImageLoader.hpp
```diff
@@ -28,7 +28,7 @@
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
-#include <SFML/System/NonCopyable.hpp>
+#include <SFML/Config.hpp>
```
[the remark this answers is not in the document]
vittorioromeo: For types like `Uint8`, used in line 69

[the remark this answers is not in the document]
Bromeon: Ah, this was transitively included through `NonCopyable.hpp` before. Ok 👍

include/SFML/Audio/InputSoundFile.hpp
eXpl0it3r: Cutting corners on documentation isn't really advisable for a library that depends on auto-generated API documentation. 😉 

With the current state (84a2a854893207cd0ff451e125193bea40cca8fb) it looks like this, which isn't that great IMHO:

image

I'd much rather have a more verbose header so that there is some documentation, than expecting everyone to see and understand the `= delete` on those functions.

The suggested change would look like this:

image
image
image

Maybe there's room later down the line to make the "delete" label more prominent, but that's then more a styling topic for doxygen

vittorioromeo: I don't think this is "cutting corners"... I don't feel like documenting everything for the sake of it is a good idea in the first place. Documentation should add value, not noise. And both users and SFML developers may go through the header files themselves, and additional noise just makes them less readable. 

For example, look at this in `Sound.hpp`:

```cpp
    /// \brief Default constructor
    ///
    Sound();
```

How is this documentation useful? What value does it provide? To me, it feels the same as:
```cpp
int x; // initialize a variable named `x` of type `int`
x += 1; // add `1` to `x`
```

Documenting everything is a common antipattern. It's like spreading `[[nodiscard]]` everywhere... :) we should only do it when it adds value.

In this particular scenario, I don't think that we add any value by individually documenting these member functions. Anyone familiar with C++11 will know that `=delete` means "disabled". And to cite @binary1248, we are not trying to teach people C++...

My personal preference would be to have no comments at all, maybe just a `// Prevent copies.` just to be stylistically consistent with every member function starting with a comment line. I strongly believe that individually providing Doxygen documentation for each `=delete` function is a waste of time, as it doesn't provide any value to the users of SFML, and it instead detracts from the readability of the header files. 

I would also remove the useless documentation of constructors like `Sound::Sound()`. Everybody knows that's a default constructor, and unless we have to talk about some specific behavior that happens when invoking it, why point out the obvious?

Bromeon: My dilemma is:
* Generally I agree with @vittorioromeo: we should only add documentation where it adds value. Unless there's a reason we want to highlight as to _why_ the copy ctor/op= is not implemented, "it's disabled" is already implied by the signature `= delete`.
* On the other hand, SFML has traditionally followed a slight "over-document" approach. Not only things like "Default constructor", but also the `\brief` keyword and other stuff that can be inferred by Doxygen. Whatever we end up choosing, would be good if it were consistent (at least once we release SFML 3, I'm not suggesting to refactor all that now).

vittorioromeo: @Bromeon: your second point reads to me as: *"I know we're doing something which doesn't bring value, but since we were doing it before, we should keep doing it for consistency."*

Which brings me to [**this**](https://cdn.shopify.com/s/files/1/0535/6917/products/consistencydemotivator.jpeg?v=1414004030) 😄

Bromeon: It's not as uncontroversial, otherwise Laurent (originally) and eXpl0it3r wouldn't support it. "Not adding value" is my own opinion 🙂 

And regarding consistency: there should be agreement whether  special member functions without noteworthy semantics are documented or not, and then this convention should be applied to the whole codebase. Can of course be done gradually, not everything has to be 100% perfect at all times, especially during migrations ⚙️ 

TLDR: the presence/absence of docs should probably not block this PR, at some point a separate discussion might be needed.

eXpl0it3r: - IMHO every public interface should be documented
- That it adds "just noise" has little to do with the documentation and more with the format requirements set out by doxygen
- Additionally, it's only really "noise" if you read the header file directly, on the generated API documentation it's more confusing when there's no information for certain selected APIs
- Sticking to "document all public interface" makes thing simpler, as you never have to check, which things need to be documented and which don't need to be documented and is overall more consistent (e.g. if we don't document default constructors, but parametered-constructors, you have a inconsistency between constructor definitions)
- IMHO the introduced "noise" is really not as problematic as portraited

> the presence/absence of docs should probably not block this PR, at some point a separate discussion might be needed.

If we introduce a inconsistency now, then yes it should be blocking, because there's no guarantee that it will be aligned later. Better to align things now, than to delay it to some other point in time, which might also be never... 😉 

In my opinion, things should be documented, but if the general consensus is not to do it, then I'll still survive. 😉

vittorioromeo: I just want to get the PR merged, as I think that's more important than discussing style, so I'll make the changes even if I disagree with them. Consistency for the sake of it is something I don't really agree with -- we should be consistently good, not just consistent.

@eXpl0it3r:  I believe there are some flaws in your reasoning:

> IMHO every public interface should be documented

That includes implicitly generated compiler functions. Should we explicitly define those so that we can just document them? If we follow that logic, we should define classes like

```cpp
struct Whatever
{
    /// obvious docs here
    Whatever() = default;

    /// obvious docs here
    ~Whatever() = default;
        
    /// obvious docs here
    Whatever(const Whatever&) = default;

    /// obvious docs here
    Whatever(Whatever&&) = default;
    
    /// obvious docs here
    Whatever& operator=(const Whatever&) = default;
    
    /// obvious docs here
    Whatever& operator=(Whatever&&) = default;

    int x, y;
};
```

instead of 

```cpp
struct Whatever
{
    int x, y;
};
```

Is that really what we want? 

> That it adds "just noise" has little to do with the documentation and more with the format requirements set out by doxygen

We shouldn't do something that we know is bad just because our tool requires it. Tooling is supposed to work for us and add value, not the other way around.

> Additionally, it's only really "noise" if you read the header file directly

People that work on SFML and want to contribute to it will read the header file. People using SFML will use "jump to declaration" in IDEs and read the header file. People interested in knowing how SFML works will read the header file. If readability of our source code is not a priority, we might as well stop formatting it...

> Sticking to "document all public interface" makes thing simpler, as you never have to check, which things need to be documented and which don't need to be documented and is overall more consistent

We could make the same argument for `[[nodiscard]]`, `const`, and other similarly verbose C++ entities. And we all know it's a bad argument. We should add verbosity and extra information only when it adds value. We should not blindly add it everywhere just because it makes us think less -- I'd argue it's a good thing to carefully select what we document. If we document everything, then our documentation becomes less useful, thus people will rely on it less. Exact same issue as sprinkling `[[nodiscard]]` everywhere.

> IMHO the introduced "noise" is really not as problematic as portraited

See answer above.

> We might not set out to teach people C++, but we can't overlook that most of the SFML users are C++ beginners, not sure if that's really as relevant here.

I don't think so. Where do you draw the line? I don't think we should be teaching that `~Foo()` means "destructor of Foo" in our documentation.

vittorioromeo: @eXpl0it3r: I have added the requested documentation. I am happy to merge this PR as is, then we can decide what to do with the docs in a separate one. Here's a real example of what I mean by noise:

```cpp
/// \brief RAII helper class to temporarily lock an available context for use
///
class SFML_WINDOW_API TransientContextLock
{
public:
    /// \brief Default constructor
    ///
    TransientContextLock();

    /// \brief Destructor
    ///
    ~TransientContextLock();

    /// \brief Deleted copy constructor
    ///
    TransientContextLock(const TransientContextLock&) = delete;

    /// \brief Deleted copy assignment
    ///
    TransientContextLock& operator=(const TransientContextLock&) = delete;
};
```

The above documentation is, frankly, BS. All it does it take up more than 30 lines of vertical estate for ZERO added value. If I'm trying to read the code and understand what `TransientContextLock` does I am forced to scroll up an down more than normal just to skip the documentation. Now, here's the alternative I propose:

```cpp
/// \brief RAII helper class to temporarily lock an available context for use
///
class SFML_WINDOW_API TransientContextLock
{
public:
    TransientContextLock();
    ~TransientContextLock();

    // Prevent copies.
    TransientContextLock(const TransientContextLock&) = delete;
    TransientContextLock& operator=(const TransientContextLock&) = delete;
};
```

No information is lost. The vertical estate taken by `TransientContextLock` is significantly reduced. I can actually focus on what matters when I read the code, rather than having to skip redundant information. Don't you agree it is better?

src/SFML/Window/GlContext.cpp
```diff
@@ -217,6 +217,18 @@ namespace
...
+            ////////////////////////////////////////////////////////////
+            /// Prevent copies.
+            ///
+            ////////////////////////////////////////////////////////////
+            TransientContext(const TransientContext&) = delete;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Deleted copy assignment
+    ///
+    ////////////////////////////////////////////////////////////
```
eXpl0it3r, line 229: Indentation

eXpl0it3r (changes requested):
Just some white space issue

include/SFML/Audio/InputSoundFile.hpp
eXpl0it3r: We two obviously read code differently. Maybe I'm too much reliant on the generated API documentation, while you might too easily dismiss it altogether. I don't find it a lot harder to read, but I also don't write/read C++ as much, compared to C#.

vittorioromeo:
@eXpl0it3r: should be fixed now.

eXpl0it3r (approved):

include/SFML/Audio/InputSoundFile.hpp
```diff
@@ -60,6 +59,18 @@ class SFML_AUDIO_API InputSoundFile : NonCopyable
...
     ~InputSoundFile();
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Deleted copy constructor
+    ///
+    ////////////////////////////////////////////////////////////
+    InputSoundFile(const InputSoundFile&) = delete;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Deleted copy assignment
```
kimci86, line 69: I think it is called assignment operator and not copy assignment. (Same remark for other files.)

vittorioromeo: https://en.cppreference.com/w/cpp/language/copy_assignment

Close enough. I specifically say "*copy* assignment" because we have "*move* assignment" operators in C++11 and above.

include/SFML/Audio/InputSoundFile.hpp
Bromeon: The only small benefit I _could_ see, is that there are people who are faster at reading text than parsing syntax, so for them `Copy assignment operator` is helpful. This is definitely arguable, as those people still need to understand the syntax-only variant if they're dealing with C++ code, but I'm purely talking about documentation here.

Other than that, agree with Vittorio's post, he summarized the issue nicely in his reply.

---

What would solve _both_ problems -- and I'm a bit surprised that this still isn't a thing in 2021 (unless I missed it) -- is if Doxygen took care of this. These kinds of comments should really be generated by Doxygen. Alone the fact, that all can be inferred from source code, shows that their documentation contains no inherent information.

In an ideal world, Doxygen would document **all** the special member functions (if requested), not just the publicly declared ones. Why? Because they are all part of the public interface, whether compiler-generated or not. Who generates them is an implementation detail.

In the same vein, `= default` should be hidden from documentation. It's purely an implementation detail and of no interest to the user. Unlike `= delete`, which tells about absence of semantics.

A really powerful doc-tool could even go a step further and auto-generate classifiers (traits) for a class, depending on the existing special member functions (or even other functions).

> **class TransientLock**
> This transiently locks stuff, bla bla...
> 
> _Supports: copy, default construction_

or: 

> _Supports: move, equality comparison, hash_

Ok, dreamt enough.

PR #1902, opened 2021-12-14
Remove redundant APIs taking '(x, y)' in favour of ones taking 'sf::Vector'

vittorioromeo:
## Description

This PR removes various interfaces accepting `(float x, float y)` in favour of ones accepting `(const sf::Vector2f&)`, as they are redundant and do not promote the consistent use of one SFML's main vocabulary types.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.

### DIFF
--- a/examples/android/app/src/main/jni/main.cpp
+++ b/examples/android/app/src/main/jni/main.cpp
@@ -87,8 +87,8 @@ int main(int argc, char *argv[])
         return EXIT_FAILURE;
 
     sf::Sprite image(texture);
-    image.setPosition(screen.width / 2, screen.height / 2);
-    image.setOrigin(texture.getSize().x/2, texture.getSize().y/2);
+    image.setPosition({screen.width / 2.f, screen.height / 2.f});
+    image.setOrigin({texture.getSize().x / 2.f, texture.getSize().y / 2.f});
 
     sf::Font font;
     if (!font.loadFromFile("tuffy.ttf"))
@@ -96,7 +96,7 @@ int main(int argc, char *argv[])
 
     sf::Text text("Tap anywhere to move the logo.", font, 64);
     text.setFillColor(sf::Color::Black);
-    text.setPosition(10, 10);
+    text.setPosition({10, 10});
 
     sf::View view = window.getDefaultView();
 
@@ -124,7 +124,7 @@ int main(int argc, char *argv[])
                     break;
                 case sf::Event::Resized:
                     view.setSize(event.size.width, event.size.height);
-                    view.setCenter(event.size.width/2, event.size.height/2);
+                    view.setCenter({event.size.width / 2.f, event.size.height / 2.f});
                     window.setView(view);
                     break;
                 case sf::Event::LostFocus:
@@ -145,7 +145,7 @@ int main(int argc, char *argv[])
                 case sf::Event::TouchBegan:
                     if (event.touch.finger == 0)
                     {
-                        image.setPosition(event.touch.x, event.touch.y);
+                        image.setPosition({static_cast<float>(event.touch.x), static_cast<float>(event.touch.y)});
 #if defined(USE_JNI)
                         vibrate(sf::milliseconds(10));
 #endif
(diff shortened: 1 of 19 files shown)


examples/android/app/src/main/jni/main.cpp
```diff
@@ -87,16 +87,16 @@ int main(int argc, char *argv[])
         return EXIT_FAILURE;
 
     sf::Sprite image(texture);
-    image.setPosition(screen.width / 2, screen.height / 2);
-    image.setOrigin(texture.getSize().x/2, texture.getSize().y/2);
+    image.setPosition({screen.width / 2.f, screen.height / 2.f});
+    image.setOrigin({texture.getSize().x/2.f, texture.getSize().y/2.f});
```
eXpl0it3r, line 91: Add spaces before and after `/ `

eXpl0it3r (changes requested):
I personally prefer adding a space after `{` and before `}`, as it makes it a bit more readable, especially when it's mixed among other parenthesis (e.g. `setPosition({ 1.f, 2.f })`).

- `sf::Rect::contains(T x, T y)` should also be removed
- Consider whether `sf::Image::create` and `sf::Image::copy` should use `sf::Vector2u` as well
- Consider whether `sf::Image::getPixel` and `sf::Image::setPixel` should use `sf::Vector2u` as well
- Consider whether `sf::RenderTexture::create` should use `sf::Vector2u` as well
- Consider whether `sf::Texture::create` and `sf::Texture::update` should use `sf::Vector2u` as well
- Consider whether the `sf::VideoMode` constructor should use `sf::Vector2u` as well

Bromeon:
> I personally prefer adding a space after `{` and before `}`, as it makes it a bit more readable, especially when it's mixed among other parenthesis (e.g. `setPosition({ 1.f, 2.f })`).

There's some truth in that. We should then probably also use that style (in the future) for aggregates and `std::initializer_list`?
```cpp
std::vector<int> v = { 1, 2, 3 };
```

vittorioromeo:
> I personally prefer adding a space after `{` and before `}`, as it makes it a bit more readable, especially when it's mixed among other parenthesis (e.g. `setPosition({ 1.f, 2.f })`)

I'm not a fan of that style, I find the extra spacing unnecessary, especially coming from the point of view that list-initialization with curly braces is just another form of initialization:

```cpp 
Foo foo(1, 2, 3); // OK
Foo foo{1, 2, 3}; // OK, consistent
Foo foo{ 1, 2, 3 }; // why?
```

I think we should have that discussion, but not as part of this PR.

---

> I agree that other methods taking X/Y pairs of numbers can be "vectorized". Can also be a separate PR, as those are technically not redundant ones; but I don't mind too much either way.

I'd prefer that as well, I like to keep PRs small. I will apply the requested review changes and start investigating the other functions for a future PR.

examples/android/app/src/main/jni/main.cpp
vittorioromeo: Done

eXpl0it3r (approved):

eXpl0it3r:
> `Foo foo{ 1, 2, 3 }; // why?`

Why not? It's a different syntax, as such the consistency argument doesn't necessarily take priority here.
I'd argue it's even a good thing to distinguish between a normal constructor call and a init list/uniform init call.

But yeah, we can have that discussion somewhere else as well. 🙂

vittorioromeo:
> > `Foo foo{ 1, 2, 3 }; // why?`
> 
> Why not? It's a different syntax, as such the consistency argument doesn't necessarily take priority here. I'd argue it's even a good thing to distinguish between a normal constructor call and a init list/uniform init call.
> 
> But yeah, we can have that discussion somewhere else as well. 🙂

Regardless of what we choose, we should do it in a way that's compatible with our automatic formatting plans. Using #1861 should be a priority IMHO

PR #1903, opened 2021-12-14
Make 'Vector2' and 'Vector3' 'constexpr' classes

vittorioromeo:

----

## Description

This PR turns both `sf::Vector2<T>` and `sf::Vector3<T>` into `constexpr`-friendly classes.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.

### DIFF
--- a/include/SFML/System/Vector2.inl
+++ b/include/SFML/System/Vector2.inl
@@ -25,7 +25,7 @@
 
 ////////////////////////////////////////////////////////////
 template <typename T>
-inline Vector2<T>::Vector2() :
+constexpr Vector2<T>::Vector2() :
 x(0),
 y(0)
 {
@@ -35,7 +35,7 @@ y(0)
 
 ////////////////////////////////////////////////////////////
 template <typename T>
-inline Vector2<T>::Vector2(T X, T Y) :
+constexpr Vector2<T>::Vector2(T X, T Y) :
 x(X),
 y(Y)
 {
@@ -46,7 +46,7 @@ y(Y)
 ////////////////////////////////////////////////////////////
 template <typename T>
 template <typename U>
-inline Vector2<T>::Vector2(const Vector2<U>& vector) :
+constexpr Vector2<T>::Vector2(const Vector2<U>& vector) :
 x(static_cast<T>(vector.x)),
 y(static_cast<T>(vector.y))
 {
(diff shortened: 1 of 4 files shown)


include/SFML/System/Vector2.inl
```diff
@@ -25,7 +25,7 @@
 
 ////////////////////////////////////////////////////////////
 template <typename T>
-inline Vector2<T>::Vector2() :
+constexpr inline Vector2<T>::Vector2() :
```
Bromeon, line 28: `constexpr` alone is enough, see https://stackoverflow.com/q/14391272
Also other occurrences.

Bromeon (commented):
Nice!

Could you elaborate the reasoning behind selecting precisely those operators for `[[nodiscard]]`?

include/SFML/System/Vector2.inl
vittorioromeo: The same is true of `template`: https://stackoverflow.com/a/10535711/598696

Since `inline` was there already, I assumed that we wanted to be more explicit (yet redundant) about it. Do you still want me to remove all `inline` specifiers?

vittorioromeo:
> Could you elaborate the reasoning behind selecting precisely those operators for `[[nodiscard]]`?

Those are the only operators for which it never makes any sense to discard the result. The main realistic use case is preventing typos of this kind:

```cpp
void integrate(sf::Vector2f& position, sf::Vector2f velocity)
{
    position + velocity; // whoops, meant to write `+=` 
}
```

include/SFML/System/Vector2.inl
Bromeon: True, good point. Not sure why it was originally there.

In my opinion, `inline` can be removed, yes. Let's see what others think.

Bromeon:
> Those are the only operators for which it never makes any sense to discard the result. The main realistic use case is preventing typos of this kind:

Thanks! Might be a bit of a border case, but I agree this has higher potential than ordinary functions to be accidentally misused, so let's do it.

include/SFML/System/Vector2.inl
vittorioromeo: @eXpl0it3r, @binary1248: any opinion on the matter?

binary1248: I think anybody slightly more experienced with C++ would know that templates are implicitly inline, thus the `inline` keyword would just be noise for them (for the compiler it is basically useless since it does not change anything). `constexpr` on the other hand does modify compilation and thus is not only an indicator to the user but the compiler as well.

I don't think we should base API design around the idea of educating users on the principles of C++. If someone wants to know whether a templated function is inline they should be reading a good C++ book instead of coming to us. 😉

eXpl0it3r (approved):

PR #1904, opened 2021-12-14
Make 'Color' a 'constexpr' class

vittorioromeo:

----

## Description

This PR turns `sf::Color` into a `constexpr`-friendly class.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.

### DIFF
--- a/include/SFML/Graphics/Color.hpp
+++ b/include/SFML/Graphics/Color.hpp
@@ -37,7 +37,7 @@ namespace sf
 /// \brief Utility class for manipulating RGBA colors
 ///
 ////////////////////////////////////////////////////////////
-class SFML_GRAPHICS_API Color
+class Color
 {
 public:
 
@@ -48,7 +48,7 @@ public:
     /// sf::Color(0, 0, 0, 255).
     ///
     ////////////////////////////////////////////////////////////
-    Color();
+    constexpr Color();
 
     ////////////////////////////////////////////////////////////
     /// \brief Construct the color from its 4 RGBA components
@@ -59,7 +59,7 @@ public:
     /// \param alpha Alpha (opacity) component (in the range [0, 255])
     ///
     ////////////////////////////////////////////////////////////
-    Color(Uint8 red, Uint8 green, Uint8 blue, Uint8 alpha = 255);
+    constexpr Color(Uint8 red, Uint8 green, Uint8 blue, Uint8 alpha = 255);
 
     ////////////////////////////////////////////////////////////
     /// \brief Construct the color from 32-bit unsigned integer
@@ -67,7 +67,7 @@ public:
     /// \param color Number containing the RGBA components (in that order)
     ///
     ////////////////////////////////////////////////////////////
-    explicit Color(Uint32 color);
+    constexpr explicit Color(Uint32 color);
 
     ////////////////////////////////////////////////////////////
     /// \brief Retrieve the color as a 32-bit unsigned integer
@@ -75,20 +75,20 @@ public:
     /// \return Color represented as a 32-bit unsigned integer
     ///
     ////////////////////////////////////////////////////////////
-    Uint32 toInteger() const;
+    constexpr Uint32 toInteger() const;
 
     ////////////////////////////////////////////////////////////
     // Static member data
     ////////////////////////////////////////////////////////////
-    static const Color Black;       //!< Black predefined color
-    static const Color White;       //!< White predefined color
-    static const Color Red;         //!< Red predefined color
-    static const Color Green;       //!< Green predefined color
-    static const Color Blue;        //!< Blue predefined color
-    static const Color Yellow;      //!< Yellow predefined color
-    static const Color Magenta;     //!< Magenta predefined color
-    static const Color Cyan;        //!< Cyan predefined color
-    static const Color Transparent; //!< Transparent (black) predefined color
+    SFML_GRAPHICS_API static const Color Black;       //!< Black predefined color
+    SFML_GRAPHICS_API static const Color White;       //!< White predefined color
+    SFML_GRAPHICS_API static const Color Red;         //!< Red predefined color
+    SFML_GRAPHICS_API static const Color Green;       //!< Green predefined color
+    SFML_GRAPHICS_API static const Color Blue;        //!< Blue predefined color
+    SFML_GRAPHICS_API static const Color Yellow;      //!< Yellow predefined color
+    SFML_GRAPHICS_API static const Color Magenta;     //!< Magenta predefined color
+    SFML_GRAPHICS_API static const Color Cyan;        //!< Cyan predefined color
+    SFML_GRAPHICS_API static const Color Transparent; //!< Transparent (black) predefined color
 
     ////////////////////////////////////////////////////////////
     // Member data
@@ -111,7 +111,7 @@ public:
 /// \return True if colors are equal, false if they are different
 ///
 ////////////////////////////////////////////////////////////
-SFML_GRAPHICS_API bool operator ==(const Color& left, const Color& right);
+[[nodiscard]] constexpr bool operator ==(const Color& left, const Color& right);
 
 ////////////////////////////////////////////////////////////
 /// \relates Color
@@ -125,7 +125,7 @@ SFML_GRAPHICS_API bool operator ==(const Color& left, const Color& right);
 /// \return True if colors are different, false if they are equal
 ///
 ////////////////////////////////////////////////////////////
-SFML_GRAPHICS_API bool operator !=(const Color& left, const Color& right);
+[[nodiscard]] constexpr bool operator !=(const Color& left, const Color& right);
 
 ////////////////////////////////////////////////////////////
 /// \relates Color
@@ -140,7 +140,7 @@ SFML_GRAPHICS_API bool operator !=(const Color& left, const Color& right);
 /// \return Result of \a left + \a right
 ///
 ////////////////////////////////////////////////////////////
-SFML_GRAPHICS_API Color operator +(const Color& left, const Color& right);
+[[nodiscard]] constexpr Color operator +(const Color& left, const Color& right);
 
 ////////////////////////////////////////////////////////////
 /// \relates Color
@@ -155,7 +155,7 @@ SFML_GRAPHICS_API Color operator +(const Color& left, const Color& right);
 /// \return Result of \a left - \a right
 ///
 ////////////////////////////////////////////////////////////
-SFML_GRAPHICS_API Color operator -(const Color& left, const Color& right);
+[[nodiscard]] constexpr Color operator -(const Color& left, const Color& right);
 
 ////////////////////////////////////////////////////////////
 /// \relates Color
@@ -172,7 +172,7 @@ SFML_GRAPHICS_API Color operator -(const Color& left, const Color& right);
 /// \return Result of \a left * \a right
 ///
 ////////////////////////////////////////////////////////////
-SFML_GRAPHICS_API Color operator *(const Color& left, const Color& right);
+[[nodiscard]] constexpr Color operator *(const Color& left, const Color& right);
 
 ////////////////////////////////////////////////////////////
 /// \relates Color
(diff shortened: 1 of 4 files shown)


include/SFML/Graphics/Color.hpp
```diff
@@ -29,6 +29,7 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Graphics/Export.hpp>
+#include <algorithm>
```
Bromeon: Exposing this mass-header is a bit unfortunate for compile times. But as you suggested [in the other issue](https://github.com/SFML/SFML/issues/1893#issuecomment-990180997), for `min`/`max` alone we could have our own lightweight implementation, in a later PR.

include/SFML/Graphics/Color.inl
```diff
@@ -22,31 +22,23 @@
...
-const Color Color::Transparent(0, 0, 0, 0);
+inline const Color Color::Black{0, 0, 0};          //!< Black predefined color
+inline const Color Color::White{255, 255, 255};    //!< White predefined color
+inline const Color Color::Red{255, 0, 0};          //!< Red predefined color
+inline const Color Color::Green{0, 255, 0};        //!< Green predefined color
+inline const Color Color::Blue{0, 0, 255};         //!< Blue predefined color
+inline const Color Color::Yellow{255, 255, 0};     //!< Yellow predefined color
+inline const Color Color::Magenta{255, 0, 255};    //!< Magenta predefined color
+inline const Color Color::Cyan{0, 255, 255};       //!< Cyan predefined color
+inline const Color Color::Transparent{0, 0, 0, 0}; //!< Transparent (black) predefined color
```
Bromeon: Why not directly `constexpr`?

Also, let's stay with the current SFML code style for now (constructor parentheses), otherwise we get an inconsistent code base. See also section "Uniform initialization" in [this forum thread](https://en.sfml-dev.org/forums/index.php?topic=28340.0).

include/SFML/Graphics/Color.inl
```diff
@@ -22,31 +22,23 @@
...
 
 
 ////////////////////////////////////////////////////////////
-Color::Color() :
+constexpr inline Color::Color() :
```
Bromeon, line 27: Depending on outcome [of this discussion](https://github.com/SFML/SFML/pull/1903#discussion_r769076008), only `constexpr`

vittorioromeo:
@binary1248: the Android building is failing with 

```
/home/buildbot-worker/android-armeabi-v7a/install/sfml/include/SFML/Graphics/Color.inl:86:29: note: lambda closure types are non-literal types before C++17
      const auto clampedAdd = [](Uint8 lhs, Uint8 rhs) -> Uint8
```

Any action I need to take to tell Android that we are indeed using C++17?

include/SFML/Graphics/Color.inl
vittorioromeo: It doesn't work, see https://stackoverflow.com/questions/11928089/static-constexpr-member-of-same-type-as-class-being-defined -- super annoying. I reverted this change anyway because VS2017 doesn't work properly with `inline` variables AFAICS.

vittorioromeo: What we should do here is change `sf::Color::Red` to `sf::Colors::Red` with `Colors` being a namespace or class, then we can define all colors as `constexpr`. But that's for a subsequent PR.

include/SFML/Graphics/Color.hpp
vittorioromeo: You are a fast reviewer :) I recognized this issue and changed my approach, implementing some local lambdas to deal with the casting and clamping of the various operations. I think the end result is cleaner than before as we avoid quite a lot of C-style casting, and we completely avoid the dependency on `<algorithm>`, which is indeed a compile-time hog.

include/SFML/Graphics/Color.inl
vittorioromeo: I created a StackOverflow question here: https://stackoverflow.com/questions/70356275/working-around-limitation-of-constexpr-static-data-member-with-same-type-as-en

binary1248:
@vittorioromeo [This line](https://github.com/SFML/SFML/blob/master/examples/android/app/build.gradle#L14) needs to be uncommented and adjusted to C++17. Don't know about the other flags.

include/SFML/Graphics/Color.inl
```diff
@@ -0,0 +1,149 @@
...
+
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr inline Color::Color(Uint32 color) :
+r(static_cast<Uint8>((color & 0xff000000) >> 24)),
+g(static_cast<Uint8>((color & 0x00ff0000) >> 16)),
+b((color & 0x0000ff00) >> 8 ),
+a((color & 0x000000ff) >> 0 )
```
Bromeon, line 53: No space before `)`

include/SFML/Graphics/Color.inl
```diff
@@ -0,0 +1,149 @@
...
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr inline Color operator *(const Color& left, const Color& right)
+{
+    const auto scaledMul = [](Uint8 lhs, Uint8 rhs) -> Uint8
+    {
+        const int intResult = static_cast<int>(lhs) * static_cast<int>(rhs);
+        return static_cast<Uint8>(intResult / 255);
```
Bromeon, line 121: On very weird architectures where `int` has 16 bits, this could cause an overflow: `255 * 255 > 2^(16-1)`

I know this was already a problem before, so not sure if it would change anything in practice. In any case, `Uint16` should do the trick.

include/SFML/Graphics/Color.hpp
```diff
@@ -37,7 +37,7 @@ namespace sf
 /// \brief Utility class for manipulating RGBA colors
 ///
 ////////////////////////////////////////////////////////////
-class SFML_GRAPHICS_API Color
+class Color
```
Bromeon, line 40: With `constexpr`, the entire class is inline and header-only, so will not be linked, right?
(Except maybe the `static` members, depending on how they're declared).

include/SFML/Graphics/Color.hpp
```diff
@@ -59,36 +59,36 @@ class SFML_GRAPHICS_API Color
...
+    constexpr explicit Color(Uint32 color);
 
     ////////////////////////////////////////////////////////////
     /// \brief Retrieve the color as a 32-bit unsigned integer
     ///
     /// \return Color represented as a 32-bit unsigned integer
     ///
     ////////////////////////////////////////////////////////////
-    Uint32 toInteger() const;
+    [[nodiscard]] constexpr Uint32 toInteger() const;
```
Bromeon, line 78: The "to" prefix implies a conversion; I don't see what the expectation of an expression like `color.toInteger();` would be. I would say this falls under the same as other getters, and thus doesn't need `[[nodiscard]]`.

include/SFML/Graphics/Color.inl
vittorioromeo: Will fix

include/SFML/Graphics/Color.inl
vittorioromeo: Honestly curious -- is there any such platform on which SFML could be used in this day and age? I will make the change anyway, it seems like a good idea to be explicit with our signedness and boundaries here.

include/SFML/Graphics/Color.hpp
vittorioromeo: Yes, that is my understanding. CI seems to agree...

include/SFML/Graphics/Color.hpp
vittorioromeo: Fair enough, I will remove this one.

include/SFML/Graphics/Color.inl
vittorioromeo: I've applied the change, but I'm not sure if it's an improvement compared to what we had before. We should consider reverting unless there actually is a platform where `int` has 16 bits which is widely used by SFML

Bromeon: I'm not aware of such platforms. Likely, this would also have led to problems elsewhere already...

In C++, there's on one side the standard crusaders who made correcting each other on StackOverflow their hobby, and on the other the "it works" people relying on various UB and implementation-defined semantics (gamedev is a great example). And then a range in between 😬

But since so many things are loosely specified, it's sometimes really hard to know whether something is a real issue in practice or not.

> We should consider reverting unless there actually is a platform where `int` has 16 bits which is targeted by SFML.

Do we need to be as overly verbose at all? Otherwise, let's use integral promotion (`Uint8 * Uint8 == int`).
```cpp
int result = lhs * rhs;
return static_cast<Uint8>(result / 255);
```
Or does that issue a warning?

vittorioromeo: We will get warnings without casts, but using `int` instead of `Uint16` does require less casts, assuming that we do not care about platforms using 16-bit `int`. How would you like me to proceed? IMHO, we should not support such platforms unless we get explicitly requested to do so.

Bromeon: I tend to agree, however that would raise the question why we have `sf::Uint8` etc. in the first place (or `uint8_t`, that's not the point).

There are two main approaches:
1. We embrace the fact that standard types are variable-sized and don't rely on "99%" assumptions. We use fixed-size types where we need them. This might be the safest and most portable option, plus it could be more expressive about the underlying assumptions in code.
2. We deliberately restrict SFML to certain platforms (possibly enforced via preprocessor, e.g. `#error` on wrong `INT_MAX` value or so). We should only do this after deeper research however, as it's really not nice to restrict certain valid use cases for "cosmetic" reasons. If we go for this, we should also be pragmatic w.r.t. other "common" implementations (`ptrdiff_t` etc).

I think it's a bigger topic, so maybe let's discuss it elsewhere.

About here, I don't know. Reducing/beautifying casts wasn't exactly a concern during https://github.com/SFML/SFML/pull/1846, so maybe we should still go for the more verbose way with `Uint16` here?

vittorioromeo: > maybe we should still go for the more verbose way with Uint16 here?

I am personally fine with that.

vittorioromeo:
@Bromeon: I think I've addressed all your requested changes, please take a look when you have some free time.

include/SFML/Graphics/Color.inl
Bromeon: Looks like there might be a way with https://stackoverflow.com/a/70356507, do you think that's applicable here?

vittorioromeo: > Looks like there might be a way with https://stackoverflow.com/a/70356507, do you think that's applicable here?

I am considering a few options, I'd like to merge this PR as is and deal with the constants in a separate one.

vittorioromeo:
@Bromeon: ping :)

include/SFML/Graphics/Color.inl
```diff
@@ -0,0 +1,149 @@
...
+
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Color::Color(Uint32 color) :
+r(static_cast<Uint8>((color & 0xff000000) >> 24)),
+g(static_cast<Uint8>((color & 0x00ff0000) >> 16)),
+b((color & 0x0000ff00) >> 8),
+a((color & 0x000000ff) >> 0)
```
Bromeon, line 53: There should be integral promotion for all 4 of them, why cast some but not all?
The code would probably be more readable 😉 

Also, not sure if `>> 0` is that helpful, we also don't do the opposite in `toInteger()`.

Bromeon (commented):
Just a small thing, after that you can merge!

include/SFML/Graphics/Color.inl
vittorioromeo: @Bromeon Done! :)

Bromeon (approved):

PR #1907, opened 2021-12-15
Unify toLower implementations

ChrisThrasher:
## Description

I found 3 instances where different files needed to transform a string into all lower case letters. Two implementations were identical and a 3rd used a standard algorithm. 3 reimplementations of the same thing seemed like justification to create a new private abstraction to handle this. Following patterns I saw elsewhere, I added a `priv` namespace to String.hpp within which `sf::priv::toLower` is implemented. 

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/src/SFML/Audio/SoundFileWriterWav.cpp
+++ b/src/SFML/Audio/SoundFileWriterWav.cpp
@@ -27,8 +27,7 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Audio/SoundFileWriterWav.hpp>
 #include <SFML/System/Err.hpp>
-#include <algorithm>
-#include <cctype>
+#include <SFML/System/Utils.hpp>
 #include <cassert>
 
 
@@ -68,11 +67,6 @@ namespace
         };
         stream.write(reinterpret_cast<const char*>(bytes), sizeof(bytes));
     }
-
-    unsigned char toLower(unsigned char character)
-    {
-        return static_cast<unsigned char>(std::tolower(character));
-    }
 }
 
 namespace sf
@@ -82,8 +76,7 @@ namespace priv
 ////////////////////////////////////////////////////////////
 bool SoundFileWriterWav::check(const std::string& filename)
 {
-    std::string extension = filename.substr(filename.find_last_of('.') + 1);
-    std::transform(extension.begin(), extension.end(), extension.begin(), toLower);
+    const std::string extension = toLower(filename.substr(filename.find_last_of('.') + 1));
 
     return extension == "wav";
 }
--- a/src/SFML/Network/Http.cpp
+++ b/src/SFML/Network/Http.cpp
@@ -27,23 +27,10 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Network/Http.hpp>
 #include <SFML/System/Err.hpp>
-#include <cctype>
+#include <SFML/System/Utils.hpp>
 #include <iterator>
 #include <sstream>
 #include <limits>
-
-
-namespace
-{
-    // Convert a string to lower case
-    std::string toLower(std::string str)
-    {
-        for (char& c : str)
-            c = static_cast<char>(std::tolower(c));
-
-        return str;
-    }
-}
 
 
 namespace sf
--- a/src/SFML/System/Utils.hpp
+++ b/src/SFML/System/Utils.hpp
@@ -1,0 +1,45 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2021 Laurent Gomila (laurent@sfml-dev.org)
+//
+// This software is provided 'as-is', without any express or implied warranty.
+// In no event will the authors be held liable for any damages arising from the use of this software.
+//
+// Permission is granted to anyone to use this software for any purpose,
+// including commercial applications, and to alter it and redistribute it freely,
+// subject to the following restrictions:
+//
+// 1. The origin of this software must not be misrepresented;
+//    you must not claim that you wrote the original software.
+//    If you use this software in a product, an acknowledgment
+//    in the product documentation would be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such,
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+#ifndef SFML_UTILS_HPP
+#define SFML_UTILS_HPP
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#include <cctype>
+#include <string>
+
+
+namespace sf
+{
+    [[nodiscard]] inline std::string toLower(std::string str)
+    {
+        for (char& c : str)
+            c = static_cast<char>(std::tolower(static_cast<unsigned char>(c)));
+        return str;
+    }
+} // namespace sf
+
+#endif // SFML_UTILS_HPP
--- a/src/SFML/Audio/SoundFileWriterFlac.cpp
+++ b/src/SFML/Audio/SoundFileWriterFlac.cpp
@@ -27,18 +27,11 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Audio/SoundFileWriterFlac.hpp>
 #include <SFML/System/Err.hpp>
+#include <SFML/System/Utils.hpp>
 #include <algorithm>
 #include <cctype>
 #include <cassert>
 
-
-namespace
-{
-    unsigned char toLower(unsigned char character)
-    {
-        return static_cast<unsigned char>(std::tolower(character));
-    }
-}
 
 namespace sf
 {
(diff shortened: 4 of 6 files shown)


include/SFML/System/String.hpp
```diff
@@ -30,13 +30,24 @@
...
 
 
 namespace sf
 {
+namespace priv
+{
+    inline std::string toLower(std::string str)
+    {
+        for (char& c : str)
+            c = static_cast<char>(std::tolower(c));
```
kimci86, line 46:
```suggestion
            c = static_cast<char>(std::tolower(static_cast<unsigned char>(c)));
```
See https://en.cppreference.com/w/cpp/string/byte/tolower#Notes

kimci86 (commented):
`sf::priv::toLower` can be simplified into `priv::toLower` or `toLower` when used in the corresponding namespace.

include/SFML/System/String.hpp
ChrisThrasher: Thanks. Fixed.

src/SFML/Network/Http.cpp
```diff
@@ -27,26 +27,13 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Network/Http.hpp>
 #include <SFML/System/Err.hpp>
-#include <cctype>
+#include <SFML/System/String.hpp>
```
vittorioromeo, line 30: `String.hpp` is one of the most expensive SFML headers to compile. We are adding it here with the only purpose of exposing `toLower` -- I am not sure that is a good idea in the context of compilation times.

Bromeon: I agree. Chances are there are more such utilities like `sf::toLower()`, and those could maybe combined in an own header `System/Utils.hpp` (if dependency-heavy, with separate `.cpp` file)?

ChrisThrasher: @vittorioromeo I totally agree. I originally wanted to put this in a private header somewhere in src/ since this is only needed in .cpp files. However, I saw no internal generic utilities file/library and I saw that many files included a `priv::` namespace in public headers for things like this so I copied that pattern.

@Bromeon I think a Utils.[hpp|cpp] is a great idea. There ought to be a central location for common implementation details like this so we can encourage better code reuse.

Bromeon: The only problem I see is that such "utils" tend to aggregate all kinds of unrelated functionality, which often comes with various different `#include`s, ultimately increasing compilation time for each of the client files.

For example, I was thinking this might be a place for the min/max functionality used in `Rect` and `Vector2/3`, but that would mean those need to include `<string>`, which is a no-no.

We can still start with `Utils`, and rename it as we see fit.

vittorioromeo: I honestly would also be fine with something like `Utils/ToLower.hpp`, and have `Utils` as a folder.

Bromeon: Also a possibility. In case of small dependencies (here `<cctype>` which might already be included by `<string>`), we can also abandon the .cpp file entirely.

ChrisThrasher: What if we put private headers in src/? That way we can avoid `priv::` and `Priv/` and have a much stronger guarantee that no user will depend on these utilities.

ChrisThrasher:
This latest revision moves this utility out of String.hpp and into System/Priv/Utils.hpp. This is good because it means we don't need to bring in the entire String.hpp header when these 3 files just need this one utility. I think we can do better though. Because we ideally want these private implementation details to be unobservable to users, I think the ideal solution is to move this file into src/. I'm still considering how exactly to implement that in a way that is more granular than simply adding 

```cmake
include_directories(.)
```

to src/CMakeLists.txt. This works, but it means that the System module could, for example, depend on files from Network without having to express a dependency on that module which we probably don't want to allow. We can't add any new public include paths inside src/ since those will be visible to users so we might be forced to do something like that `include_directories(.)` line to allow sharing files between .cpp files without them being visible to users.

ChrisThrasher:
https://github.com/ChrisThrasher/SFML/commit/c8e963243c2e3feb9f12dd63c300c671428530d6

Here's one way we could potentially add private headers to src/ to prevent having to add more implementation details to public headers.

kimci86:
@ChrisThrasher It is already possible to put implementation headers in `src/` without new CMake commands.
See https://github.com/SFML/SFML/blob/master/cmake/Macros.cmake#L212-L214

ChrisThrasher:
> @ChrisThrasher It is already possible to put implementation headers in `src/` without new CMake commands.
> See https://github.com/SFML/SFML/blob/master/cmake/Macros.cmake#L212-L214

Oh, lovely. In that case I feel ever better about moving this function to src/SFML/System/Utils.hpp (or maybe ToLower.hpp). I merged that commit into this PR. I'll squash commits as requested once we all agree on a solution. I'll keep both commits for now in case we choose to remove the latest commit I added.

eXpl0it3r:

eXpl0it3r:

src/SFML/Audio/SoundFileWriterWav.cpp
```diff
@@ -27,8 +27,8 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Audio/SoundFileWriterWav.hpp>
 #include <SFML/System/Err.hpp>
+#include <SFML/System/Utils.hpp>
 #include <algorithm>
```
vittorioromeo, line 31: Can you check if this include can be removed now? `<algorithm>` is quite expensive as a header, would be nice to get rid of it as part of this PR if it's unnecessary.

vittorioromeo (changes requested):
One minor thing, LGTM otherwise

src/SFML/Audio/SoundFileWriterWav.cpp
ChrisThrasher: Looks like it can. I'll remove it.

src/SFML/System/Utils.hpp
```diff
@@ -0,0 +1,45 @@
...
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#include <cctype>
+#include <string>
+
+
+namespace sf
+{
+    inline std::string toLower(std::string str)
```
roughbits01, line 37: Is there a reason `str` is passed by copy? (I know it was like this before your refactor)

ChrisThrasher: A copy is required if we want to modify the argument then return it.

roughbits01: Oh that makes sense! 👍

ChrisThrasher:
@vittorioromeo I found two more files with custom `toLower` implementations and was able to remove `<algorithm>` in two more places!

ChrisThrasher:
```
  D:\a\SFML\SFML\src\SFML\Audio\SoundFileWriterFlac.cpp(101): error C2039: 'min': is not a member of 'std' [D:\a\SFML\SFML\build\src\SFML\Audio\sfml-audio.vcxproj]
793
  D:\a\SFML\SFML\src\SFML\Audio\SoundFileWriterFlac.cpp(101): error C3861: 'min': identifier not found [D:\a\SFML\SFML\build\src\SFML\Audio\sfml-audio.vcxproj]
794
  D:\a\SFML\SFML\src\SFML\Audio\SoundFileWriterOgg.cpp(145): error C2039: 'min': is not a member of 'std' [D:\a\SFML\SFML\build\src\SFML\Audio\sfml-audio.vcxproj]
795
  D:\a\SFML\SFML\src\SFML\Audio\SoundFileWriterOgg.cpp(145): error C3861: 'min': identifier not found [D:\a\SFML\SFML\build\src\SFML\Audio\sfml-audio.vcxproj]
796
  D:\a\SFML\SFML\src\SFML\Audio\SoundFileWriterOgg.cpp(150): error C2039: 'min': is not a member of 'std' [D:\a\SFML\SFML\build\src\SFML\Audio\sfml-audio.vcxproj]
797
  D:\a\SFML\SFML\src\SFML\Audio\SoundFileWriterOgg.cpp(150): error C3861: 'min': identifier not found [D:\a\SFML\SFML\build\src\SFML\Audio\sfml-audio.vcxproj]
```

I may have been too quick to remove `<algorithm>`.

@vittorioromeo Don't we have a non-standard min/max implementation that we're using to avoid more inclusions of `<algorithm>`? I think we need to add that to src\SFML\Audio\SoundFileWriterOgg.cpp and src\SFML\Audio\SoundFileWriterFlac.cpp.

eXpl0it3r:

ChrisThrasher:
I followed through on @vittorioromeo's recommendation but the PR still shows as being blocked by an unaddressed request. Not sure what to make of that but it appears that everything else is good to go.

eXpl0it3r:
> the PR still shows as being blocked by an unaddressed request

It's shown as such, until the reviewer approves it 😉

vittorioromeo (approved):
LGTM, thanks!!

PR #1908, opened 2021-12-16
Fix warnings for clang

reconndev:
## Description

~That's a quick fix as `-DWARNINGS_AS_ERRORS=1` does not work for clang compilers because of how the logic was ordered.~

This PR can't be merged without all clang warnings removed.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```bash
cmake -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=1 -DWARNINGS_AS_ERRORS=1
# Assert the flag was enabled
grep 'WARNINGS_AS_ERRORS' ./build/CMakeCache.txt
# Output should be: WARNINGS_AS_ERRORS:BOOL=1
grep -om 1 '\-Werror' ./build/compile_commands.json 
# Output should be: -Werror
```

### DIFF
--- a/cmake/CompilerWarnings.cmake
+++ b/cmake/CompilerWarnings.cmake
@@ -4,12 +4,7 @@
 
 # Helper function to enable compiler warnings for a specific set of files
 function(set_file_warnings)
-    if(APPLE)
-        # Temporarily disable Apple default until warnings are fixed
-        option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" FALSE)
-    else()
-        option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" TRUE)
-    endif()
+    option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" TRUE)
 
     set(MSVC_WARNINGS
         /W4 # Baseline reasonable warnings
--- a/cmake/CompilerWarnings.cmake
+++ b/cmake/CompilerWarnings.cmake
@@ -4,7 +4,12 @@
 
 # Helper function to enable compiler warnings for a specific set of files
 function(set_file_warnings)
-    option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" TRUE)
+    if(APPLE)
+        # Temporarily disable Apple default until warnings are fixed
+        option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" FALSE)
+    else()
+        option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" TRUE)
+    endif()
 
     set(MSVC_WARNINGS
         /W4 # Baseline reasonable warnings
@@ -69,16 +74,15 @@ function(set_file_warnings)
         ${NON_ANDROID_CLANG_AND_GCC_WARNINGS}
     )
 
+    if(WARNINGS_AS_ERRORS)
+        set(CLANG_AND_GCC_WARNINGS ${CLANG_AND_GCC_WARNINGS} -Werror)
+        set(MSVC_WARNINGS ${MSVC_WARNINGS} /WX)
+    endif()
 
     set(CLANG_WARNINGS
         ${CLANG_AND_GCC_WARNINGS}
         -Wno-unknown-warning-option # do not warn on GCC-specific warning diagnostic pragmas
     )
-
-    if(WARNINGS_AS_ERRORS)
-        set(CLANG_AND_GCC_WARNINGS ${CLANG_AND_GCC_WARNINGS} -Werror)
-        set(MSVC_WARNINGS ${MSVC_WARNINGS} /WX)
-    endif()
 
     set(GCC_WARNINGS
         ${CLANG_AND_GCC_WARNINGS}
--- a/examples/cocoa/CocoaAppDelegate.mm
+++ b/examples/cocoa/CocoaAppDelegate.mm
@@ -31,6 +31,14 @@
 #define GREEN   @"Green"
 #define RED     @"Red"
 
+#if defined(__APPLE__)
+    #if defined(__clang__)
+        #pragma clang diagnostic ignored "-Wdeprecated-declarations"
+    #elif defined(__GNUC__)
+        #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+    #endif
+#endif
+
 // Our PIMPL
 struct SFMLmainWindow
 {
@@ -48,7 +56,7 @@ struct SFMLmainWindow
         sf::FloatRect rect = sprite.getLocalBounds();
         sf::Vector2f size(rect.width, rect.height);
         sprite.setOrigin(size / 2.f);
-        sprite.scale(0.3, 0.3);
+        sprite.scale(0.3f, 0.3f);
 
         unsigned int ww = renderWindow.getSize().x;
         unsigned int wh = renderWindow.getSize().y;
@@ -117,7 +125,7 @@ struct SFMLmainWindow
         self.visible = YES;
 
         // Launch the timer to periodically display our stuff into the Cocoa view.
-        self.renderTimer = [NSTimer timerWithTimeInterval:1.f/60.f
+        self.renderTimer = [NSTimer timerWithTimeInterval:1.0/60.0
                                                    target:self
                                                  selector:@selector(renderMainWindow:)
                                                  userInfo:nil
@@ -148,7 +156,7 @@ struct SFMLmainWindow
     self.sfmlView           = nil;
     self.textField          = nil;
 
-    delete (SFMLmainWindow*) self.mainWindow;
+    delete static_cast<SFMLmainWindow*>(self.mainWindow);
     self.mainWindow         = 0;
     self.renderTimer        = nil;
 
--- a/src/SFML/Graphics/Shader.cpp
+++ b/src/SFML/Graphics/Shader.cpp
@@ -1035,256 +1035,256 @@ Shader::~Shader()
 
 
 ////////////////////////////////////////////////////////////
-bool Shader::loadFromFile(const std::string& filename, Type type)
+bool Shader::loadFromFile(const std::string& /* filename */, Type /* type */)
 {
     return false;
 }
 
 
 ////////////////////////////////////////////////////////////
-bool Shader::loadFromFile(const std::string& vertexShaderFilename, const std::string& fragmentShaderFilename)
+bool Shader::loadFromFile(const std::string& /* vertexShaderFilename */, const std::string& /* fragmentShaderFilename */)
 {
     return false;
 }
 
 
 ////////////////////////////////////////////////////////////
-bool Shader::loadFromFile(const std::string& vertexShaderFilename, const std::string& geometryShaderFilename, const std::string& fragmentShaderFilename)
+bool Shader::loadFromFile(const std::string& /* vertexShaderFilename */, const std::string& /* geometryShaderFilename */, const std::string& /* fragmentShaderFilename */)
 {
     return false;
 }
 
 
 ////////////////////////////////////////////////////////////
-bool Shader::loadFromMemory(const std::string& shader, Type type)
+bool Shader::loadFromMemory(const std::string& /* shader */, Type /* type */)
 {
     return false;
 }
 
 
 ////////////////////////////////////////////////////////////
-bool Shader::loadFromMemory(const std::string& vertexShader, const std::string& fragmentShader)
+bool Shader::loadFromMemory(const std::string& /* vertexShader */, const std::string& /* fragmentShader */)
 {
     return false;
 }
 
 
 ////////////////////////////////////////////////////////////
-bool Shader::loadFromMemory(const std::string& vertexShader, const std::string& geometryShader, const std::string& fragmentShader)
+bool Shader::loadFromMemory(const std::string& /* vertexShader */, const std::string& /* geometryShader */, const std::string& /* fragmentShader */)
 {
     return false;
 }
 
 
 ////////////////////////////////////////////////////////////
-bool Shader::loadFromStream(InputStream& stream, Type type)
+bool Shader::loadFromStream(InputStream& /* stream */, Type /* type */)
 {
     return false;
 }
 
 
 ////////////////////////////////////////////////////////////
-bool Shader::loadFromStream(InputStream& vertexShaderStream, InputStream& fragmentShaderStream)
+bool Shader::loadFromStream(InputStream& /* vertexShaderStream */, InputStream& /* fragmentShaderStream */)
 {
     return false;
 }
 
 
 ////////////////////////////////////////////////////////////
-bool Shader::loadFromStream(InputStream& vertexShaderStream, InputStream& geometryShaderStream, InputStream& fragmentShaderStream)
+bool Shader::loadFromStream(InputStream& /* vertexShaderStream */, InputStream& /* geometryShaderStream */, InputStream& /* fragmentShaderStream */)
 {
     return false;
 }
 
 
 ////////////////////////////////////////////////////////////
-void Shader::setUniform(const std::string& name, float x)
+void Shader::setUniform(const std::string& /* name */, float)
 {
 }
 
 
 ////////////////////////////////////////////////////////////
-void Shader::setUniform(const std::string& name, const Glsl::Vec2& v)
+void Shader::setUniform(const std::string& /* name */, const Glsl::Vec2&)
 {
 }
 
 
 ////////////////////////////////////////////////////////////
-void Shader::setUniform(const std::string& name, const Glsl::Vec3& v)
+void Shader::setUniform(const std::string& /* name */, const Glsl::Vec3&)
 {
 }
 
 
 ////////////////////////////////////////////////////////////
-void Shader::setUniform(const std::string& name, const Glsl::Vec4& v)
+void Shader::setUniform(const std::string& /* name */, const Glsl::Vec4&)
 {
 }
 
 
 ////////////////////////////////////////////////////////////
-void Shader::setUniform(const std::string& name, int x)
+void Shader::setUniform(const std::string& /* name */, int)
 {
 }
 
 
 ////////////////////////////////////////////////////////////
-void Shader::setUniform(const std::string& name, const Glsl::Ivec2& v)
+void Shader::setUniform(const std::string& /* name */, const Glsl::Ivec2&)
 {
 }
 
 
 ////////////////////////////////////////////////////////////
-void Shader::setUniform(const std::string& name, const Glsl::Ivec3& v)
+void Shader::setUniform(const std::string& /* name */, const Glsl::Ivec3&)
 {
 }
 
 
 ////////////////////////////////////////////////////////////
-void Shader::setUniform(const std::string& name, const Glsl::Ivec4& v)
+void Shader::setUniform(const std::string& /* name */, const Glsl::Ivec4&)
 {
 }
 
 
 ////////////////////////////////////////////////////////////
-void Shader::setUniform(const std::string& name, bool x)
+void Shader::setUniform(const std::string& /* name */, bool)
 {
 }
 
 
 ////////////////////////////////////////////////////////////
-void Shader::setUniform(const std::string& name, const Glsl::Bvec2& v)
+void Shader::setUniform(const std::string& /* name */, const Glsl::Bvec2&)
 {
 }
 
 
 ////////////////////////////////////////////////////////////
-void Shader::setUniform(const std::string& name, const Glsl::Bvec3& v)
+void Shader::setUniform(const std::string& /* name */, const Glsl::Bvec3&)
 {
 }
 
 
 ////////////////////////////////////////////////////////////
-void Shader::setUniform(const std::string& name, const Glsl::Bvec4& v)
+void Shader::setUniform(const std::string& /* name */, const Glsl::Bvec4&)
 {
 }
 
 
 ////////////////////////////////////////////////////////////
-void Shader::setUniform(const std::string& name, const Glsl::Mat3& matrix)
+void Shader::setUniform(const std::string& /* name */, const Glsl::Mat3& /* matrix */)
 {
 }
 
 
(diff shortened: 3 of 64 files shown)


eXpl0it3r:
I think we may want to fix this for the `2.6.x` branch as well, could you rebase/cherry-pick that commit and change the target branch?

eXpl0it3r:

eXpl0it3r:
Are you also interested in providing fixes for the failures?

cmake/CompilerWarnings.cmake
```diff
@@ -4,7 +4,7 @@
 
 # Helper function to enable compiler warnings for a specific set of files
 function(set_file_warnings)
-    option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" TRUE)
```
binary1248, line 7: Is these a reason the default has to be changed?

reconndev: Users compiling SFML do not necessarily want warnings to be treated as errors if it's a third-party library.

If that's intended, they're free to use a special case `-DWARNINGS_AS_ERRORS=1`.

This was not a problem before because gcc did not report warnings, and clang did but they were accidentally ignored. When I left this option as TRUE and allowed clang users to throw errors on warnings, all clang-based CI failed. Mostly because of macOS deprecation warnings and those similar.

I think WARNINGS_AS_ERRORS=1 could be an option for CI (after clang warnings are removed). As far as a user is concerned, the library should default to not include `-Werror` as it is most useful to SFML developers.

vittorioromeo: My understanding was that `WARNINGS_AS_ERRORS` only applies to SFML itself, and that seems true from my testing. It shouldn't propagate to user projects including SFML. But it should be enabled by default for SFML itself.

binary1248: @reconndev When we write code, we can never foresee any platform/architecture specific problems that can arise when building for certain platforms that we didn't have access to when writing the code. The compiler that is building the code knows best if some code might be problematic on a specific architecture hence the warnings that it emits. This can be seen when we didn't take alignment on ARM architectures into account when first building with warnings.

Ideally when building SFML no warnings should ever be emitted regardless what architecture it is built on. Because compiler warnings are very conservative and any false positives are treated as serious bugs by the compiler maintainers one can be fairly confident that they are basically always correct.

The real issue here is that Apple has (as usual) chosen to misuse something that was originally meant as an aid for C++ developers. If a feature is deprecated from the C++ standard (like `std::auto_ptr`) to be removed in the future, that is a case where a warning from a _C++ compiler_ would be OK. Because this happens over the course of many years, and the necessary adjustments to the code are fairly obvious, they are warnings that everybody happily accepts. Apple has a tradition of deprecating APIs fairly frequently and expects developers to change perfectly working code to deal with the warnings. This is a business decision and has nothing to do with ensuring the code does what it is supposed to do (as with the alignment warning on ARM).

Imagine if a user were to really build SFML as a dependency on some fairly new architecture. Warnings would be emitted informing the user that something doesn't feel right. It turns out that some code does invoke undefined behaviour on said platform. The user files a bug report with us. Conventionally we don't ask them if they noticed if any warnings were generated when they were building SFML as a dependency (I'm sure less strict libraries generate a heap of other warnings as well). This would be a really annoying bug to fix, since most of us will not have access to this platform to do our own testing.

In order to reduce the maintenance burden on us (having to ask if the library was "correctly" built) whenever an issue is opened, we have the expectation that the library is built without errors/warnings. If the user chooses to ignore warnings, they do so at their own risk and we can't really support them under those circumstances.

If building on macOS is causing warnings to be generated, then that has to be fixed as opposed to changing the default for all other platforms as well. Whether it involves disabling the deprecation warnings or if all else fails then just changing the default when building for macOS. We can't let business policies impact what developers on all other platforms have to deal with as well.

reconndev: > My understanding was that WARNINGS_AS_ERRORS only applies to SFML itself, and that seems true from my testing.

Did you test it after my fix on a clang compiler? Otherwise, you wouldn't have seen the difference because this simply did not work.

In short, if I (as a user) wanted to enable `WARNINGS_AS_ERRORS` it wouldn't have worked for clang. Now, I fixed that but since Clang has lots of warnings it fails on CI.

reconndev: @binary1248 I don't think you know what this PR is about.

I fixed `WARNINGS_AS_ERRORS` for clang. SFML build ignored warnings on clang before anyway because the flag didn't work properly. Do you expect me to fix all ~200 warnings on macOS in a PR that only wanted to fix `WARNINGS_AS_ERRORS`?

I changed the default value to FALSE so that the CI doesn't fail on all clang-based platforms because *only now* it errors out on warnings and there are many unfixed warnings.

eXpl0it3r: Seems like there has been a bit of a misunderstanding 😄 

We can't / don't really want to disable the default behavior, otherwise all new changes, won't be properly checked, thus can easily introduce new warnings, that we need to provide additional fixes again.

There isn't really an easy way to get around fixing all the remaining errors for clang, if you want to utilize the CI to indicate whether the PR actually fixed the issues.

We of course don't expect you to fix all the errors, thus my question whether you're willing to help with the fixes, just so we know, whether we should pick it up and nobody is doing redundant work. 😅 

As for macOS, it might be that we're not able to fix all the warnings, which I think is what binary1248 was pointing to, because Apple marked a lot of APIs as deprecated, which we still want to make use of. So we'll have to find ways to silence those or maybe some can actually be updated to newer APIs.

binary1248: Bugs are instances where actual behaviour doesn't correspond intended behaviour. Yes, it is correct that the actual "bug" is that the ordering of the flags was wrong, and that would be what needs to be fixed. Changing a default for all platforms isn't part of the fix for the actual "bug".

If fixing the issue causes CI to fail for macOS then maybe those warnings would have to be dealt with in a separate PR before this one is merged. If the warnings would have lead to build failure in the initial PR by @vittorioromeo where the sweeping changes were made to resolve warnings on all other platforms the PR wouldn't have been merged. It was an oversight, but not something that makes it necessary to revert the default back now.

I'm not a fan of workarounds for problems that can be fixed with a proportionate amount of effort. Going from memory, many of the deprecation warnings were specific to OpenGL/OpenAL. Those warnings can be disabled on a file by file basis, and there aren't as many distinct files as the warnings might have one believe.

binary1248: This file list is from the CI log of the last failure due to deprecation warnings:
```
examples/cocoa/CocoaAppDelegate.mm
src/SFML/Audio/ALCheck.cpp
src/SFML/Audio/AudioDevice.cpp
src/SFML/Audio/Sound.cpp
src/SFML/Audio/SoundBuffer.cpp
src/SFML/Audio/SoundRecorder.cpp
src/SFML/Audio/SoundSource.cpp
src/SFML/Audio/SoundStream.cpp
src/SFML/Window/OSX/SFApplication.m
src/SFML/Window/OSX/SFApplicationDelegate.m
src/SFML/Window/OSX/SFContext.mm
src/SFML/Window/OSX/SFKeyboardModifiersHelper.mm
src/SFML/Window/OSX/SFOpenGLView+mouse.mm
src/SFML/Window/OSX/SFOpenGLView.h
src/SFML/Window/OSX/SFOpenGLView.mm
src/SFML/Window/OSX/SFViewController.mm
src/SFML/Window/OSX/SFWindowController.mm
src/SFML/Window/OSX/WindowImplCocoa.hpp
src/SFML/Window/OSX/WindowImplDelegateProtocol.h
src/SFML/Window/OSX/cg_sf_conversion.mm
```

It seems the easiest way to disable warnings in these files by adding the following lines somewhere near the top of each:
```cpp
#pragma clang diagnostic ignored "-Wdeprecated-declarations"
#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```

reconndev: Since I did not know about the "no warnings" policy you have at SFML (especially considering I was getting so many warnings before), I wasn't sure whether the default was correct, too. That's why I changed it.

My understanding was that most projects allow some warnings to happen and so CI instances could make use of `-DWARNINGS_AS_ERRORS=1` on e.g. GCC builds, which were clean to not introduce any new ones.

I didn't want to fix all the warnings in one PR. I thought it would be easier gradually. Because of the "no warnings" policy, I will have to do it in this PR or not at all to pass CI checks so thank you for the clarification now, though lengthily delivered.

binary1248: This might also tie in to #1896, the current idea is that there will be a header that is used only internally when building SFML. These `#pragma`s could be placed in there and apply only when building on macOS.

Bromeon: > Because of the "no warnings" policy, I will have to do it in this PR or not at all to pass CI checks

I agree that "extending PRs" like this is not really in our interest, it just increases the time-to-merge of the first half of the feature, leading to subsequent merge conflicts and needless extra work.

What if we disabled the treat-warnings-as-errors for now, but immediately open a new issue/PR where we revert it and fix the MacOS errors? That way, we can take our time, and we don't block the existing parts of this PR from becoming part of `master`.

binary1248:
@reconndev Just a heads up, I went ahead and did the legwork on fixing the warnings that would pop up when clang does it's job. You can rebase/cherry-pick onto the [bugfix/more_warnings](https://github.com/SFML/SFML/tree/bugfix/more_warnings) and submit both commits in this PR if you want.

reconndev:
Thank you for that, however, our work overlaps now. I did very similar steps on two separate local branches. I will use your commits to cherry-pick but after I merge mine so that I can resolve potential conflicts clearly.

cmake/CompilerWarnings.cmake
reconndev: Changed the default to FALSE only when CMake declares `APPLE`.

Now, I am working on resolving clang-based compiler errors. The last part will be with deprecations from Apple but I see @binary1248 already did some work on that so I only expect to merge that branch or cherry-pick commits.

src/SFML/Graphics/Shader.cpp
```diff
@@ -1317,7 +1317,7 @@ bool Shader::isGeometryAvailable()
 
 
 ////////////////////////////////////////////////////////////
-bool Shader::compile(const char* vertexShaderCode, const char* geometryShaderCode, const char* fragmentShaderCode)
+bool Shader::compile(const char* /* vertexShaderCode */, const char* /* geometryShaderCode */, const char* /* fragmentShaderCode */)
```
vittorioromeo, line 1320: We could use `[[maybe_unused]]` for these instead of comments, but I don't particularly care.

src/SFML/Graphics/VertexBuffer.cpp
```diff
@@ -206,6 +206,7 @@ bool VertexBuffer::update(const Vertex* vertices, std::size_t vertexCount, unsig
 ////////////////////////////////////////////////////////////
 bool VertexBuffer::update(const VertexBuffer& vertexBuffer)
 {
+    (void) vertexBuffer;
```
vittorioromeo, line 209: But we should at least be consistent. Either always use `(void)`, or `[[maybe_unused]]`, or comment. Might be worthwhile to see what the rest of the library already does.

src/SFML/Main/SFMLActivity.cpp
```diff
@@ -68,7 +68,7 @@ const char *getLibraryName(JNIEnv* lJNIEnv, jobject& objectActivityInfo)
     }
 
     // Convert the application name to a C++ string and return it
-    const jsize applicationNameLength = lJNIEnv->GetStringUTFLength(valueString);
+    const size_t applicationNameLength = static_cast<size_t>(lJNIEnv->GetStringUTFLength(valueString));
```
vittorioromeo, line 71:
```suggestion
    const std::size_t applicationNameLength = static_cast<std::size_t>(lJNIEnv->GetStringUTFLength(valueString));
```

We always qualify standard types with `std::`, please adjust other uses as well.

src/SFML/Window/Android/SensorImpl.cpp
```diff
@@ -93,7 +97,7 @@ bool SensorImpl::open(Sensor::Type sensor)
     Time minimumDelay = microseconds(ASensor_getMinDelay(m_sensor));
 
     // Set the event rate (not to consume too much battery)
-    ASensorEventQueue_setEventRate(sensorEventQueue, m_sensor, minimumDelay.asMicroseconds());
+    ASensorEventQueue_setEventRate(sensorEventQueue, m_sensor, static_cast<sf::Int32>(minimumDelay.asMicroseconds()));
```
vittorioromeo, line 100:
```suggestion
    ASensorEventQueue_setEventRate(sensorEventQueue, m_sensor, static_cast<Int32>(minimumDelay.asMicroseconds()));
```

src/SFML/Window/Android/WindowImplAndroid.cpp
```diff
@@ -402,7 +409,7 @@ int WindowImplAndroid::processKeyEvent(AInputEvent* _event, ActivityStates& stat
         event.type = Event::KeyReleased;
         forwardEvent(event);
 
-        if (int unicode = getUnicode(_event))
+        if (sf::Uint32 unicode = static_cast<sf::Uint32>(getUnicode(_event)))
```
vittorioromeo, line 412:
```suggestion
        if (Uint32 unicode = static_cast<Uint32>(getUnicode(_event)))
```

src/SFML/Window/Android/WindowImplAndroid.cpp
```diff
@@ -425,7 +432,7 @@ int WindowImplAndroid::processKeyEvent(AInputEvent* _event, ActivityStates& stat
             // https://code.google.com/p/android/issues/detail?id=33998
             return 0;
         }
-        else if (int unicode = getUnicode(_event)) // This is a repeated sequence
+        else if (sf::Uint32 unicode = static_cast<sf::Uint32>(getUnicode(_event))) // This is a repeated sequence
```
vittorioromeo, line 435:
```suggestion
        else if (Uint32 unicode = static_cast<Uint32>(getUnicode(_event))) // This is a repeated sequence
```

src/SFML/Window/Android/WindowImplAndroid.cpp
```diff
@@ -445,23 +452,23 @@ int WindowImplAndroid::processKeyEvent(AInputEvent* _event, ActivityStates& stat
...
     int32_t device = AInputEvent_getSource(_event);
-    int32_t action = AMotionEvent_getAction(_event);
+    // int32_t action = AMotionEvent_getAction(_event);
 
     Event event;
 
     if (device == AINPUT_SOURCE_MOUSE)
         event.type = Event::MouseMoved;
-    else if (device & AINPUT_SOURCE_TOUCHSCREEN)
+    else if (static_cast<sf::Uint32>(device) & AINPUT_SOURCE_TOUCHSCREEN)
```
vittorioromeo, line 461:
```suggestion
    else if (static_cast<Uint32>(device) & AINPUT_SOURCE_TOUCHSCREEN)
```

vittorioromeo (changes requested):
Just had a quick look so far, thanks for the work!

src/SFML/Graphics/Shader.cpp
reconndev: I did that but this has to be C++98 compliant (because of 2.6.x) and I had to revert because while gcc and clang ignore unknown attributes sometimes, MSVC throws errors.

src/SFML/Graphics/VertexBuffer.cpp
reconndev: I chose to comment out those which are not used at all and cast to (void) those which are partially used.

`[[maybe_unused]]` as stated earlier is out of the capabilities.

src/SFML/Main/SFMLActivity.cpp
reconndev: Still, this is not targeted for C++11 unfortunately, yet.

src/SFML/Graphics/Shader.cpp
vittorioromeo: Oh, sorry -- I didn't realize this was being merged into 2.x.

src/SFML/Main/SFMLActivity.cpp
vittorioromeo: `std::size_t` is not a C++11 feature.

reconndev: Hm, you may be right.

My problem was that I wasn't sure if std::size_t was uniformly standardized in <cstddef> and whether this header existed on all versions of major compilers (as opposed to stddef.h). That's why I decided to go with the safer choice.

I'll change to std::size and see if all CI passes later.

eXpl0it3r:

examples/cocoa/CocoaAppDelegate.mm
```diff
@@ -31,6 +31,14 @@
 #define GREEN   @"Green"
 #define RED     @"Red"
 
+#if defined(__APPLE__)
+    #if defined(__clang__)
+        #pragma clang diagnostic ignored "-Wdeprecated-declarations"
+    #elif defined(__GNUC__)
+        #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+    #endif
+#endif
```
vittorioromeo, line 40: I think that the code as written doesn't ignore that warning only for this specific file, but also for any other subsequent one. See [gcc docs](https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html).

I think it should be like this:

```cpp
// Around the beginning of file...
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wdeprecated-declarations"

// ...stuff...

// Around the end of the file...
#pragma GCC diagnostic pop
```

vittorioromeo (changes requested):
Thanks, everything looks good to me except for the way warnings get disabled via pragmas -- I think we need to take another look at that and use push/pop.

examples/cocoa/CocoaAppDelegate.mm
binary1248: The GCC documentation assumes the user is aware that anything that has an influence on the compilation process only applies to the current invocation. Since a new GCC process is created for every TU we compile it is impossible for any diagnostic state to carry over to future TUs.

We deliberately ignore after the `#include`s and don't pop the state since we want the deprecation warnings to be ignored for the whole file. You can think of there being an implicit pop at the end of the file (where the stack ceases to exist).

vittorioromeo: @binary1248: sure, it's fine in a source file... but in the PR that pattern is used in a few`.hpp` files, that's why I mentioned it.

binary1248: The deprecation warnings were generated from contents in a header file as well. They would be generated in every TU that includes that header, thus the pragma has to be there too as well. It doesn't look "clean" but this is a case where I think a "dirty" solution is necessary to fix a "dirty" problem.

reconndev: Yes, it's going to be propagated through the whole translation unit but not others. Still, the user warnings can be affected so I think @vittorioromeo has a point.

We probably don't want an unguarded (push/pop) pragma in a **header file** because it could suppress the user warnings. https://godbolt.org/z/hnf963n4W

I'll change and see if any new warning emerges.

reconndev: I wonder, though, is checking for `__clang__` and `__GNUC__` necessary at all @binary1248? I mean unknown pragmas will get discarded, won't they?

binary1248: These are internal headers we are talking about, the only code it affects is ours, and since all warnings will cause building to fail I am OK with disabling Apple deprecation warnings wherever necessary in SFML's internals. The day Apple decides that these APIs aren't just deprecated but also removes them CI will fail to build completely and that will warrant a complete rewrite of the Apple implementation in SFML. Of course, if anybody is willing to take the time to rewrite now, they are free to do so as well.

Regarding unknown pragmas: I have seen instances where compilers warn about them, since we are building with `-Werror` those warnings will cause building to fail.

src/SFML/Window/OSX/SFOpenGLView.h
```diff
@@ -28,6 +28,14 @@
 ////////////////////////////////////////////////////////////
 #import <AppKit/AppKit.h>
 
+#if defined(__APPLE__)
+    #if defined(__clang__)
+        #pragma clang diagnostic ignored "-Wdeprecated-declarations"
+    #elif defined(__GNUC__)
+        #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+    #endif
+#endif
```
vittorioromeo, line 37: Missed one

src/SFML/Window/OSX/WindowImplDelegateProtocol.h
```diff
@@ -31,6 +31,14 @@
 
 #import <AppKit/AppKit.h>
 
+#if defined(__APPLE__)
+    #if defined(__clang__)
+        #pragma clang diagnostic ignored "-Wdeprecated-declarations"
+    #elif defined(__GNUC__)
+        #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+    #endif
+#endif
```
vittorioromeo, line 40: This one too

vittorioromeo (changes requested):
Only two minor changes and I'm happy to merge!

eXpl0it3r:

vittorioromeo (approved):
Good stuff, thank you so much! LGTM

vittorioromeo:
@eXpl0it3r: are you going to backmerge this into master?

eXpl0it3r:
Yes 🙂

PR #1855, opened 2021-11-30
Enable and require C++17 support

vittorioromeo:
Thanks a lot for making a contribution to SFML! 🙂

Before you create the pull request, we ask you to check the follow boxes. (For small changes not everything needs to ticked, but the more the better!)

* [x] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before?
* [x] Does the code follow the SFML [Code Style Guide](https://www.sfml-dev.org/style.php)?
* [ ] Have you provided some example/test code for your changes?
* [ ] If you have additional steps which need to be performed list them as tasks!

----

## Description

This change enables (and requires) C++17 support in all library and executable targets generated by SFML's CMake build.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Compiling via CI and executing tests should be enough. There are no expected behavioral changes that might affect the correctness of the existing code.

I've also checked that `-std=c++17` is actually passed by using `CMAKE_EXPORT_COMPILE_COMMANDS`.

PR #1910, opened 2021-12-16
Make 'Vertex' a 'constexpr' class

vittorioromeo:
## Description

This PR turns `sf::Vertex` into a `constexpr`-friendly class. It depends on #1904, I am creating the PR before that one is merged just for early code review. CI is expected to fail at this stage.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.

### DIFF
--- a/include/SFML/Graphics/Vertex.hpp
+++ b/include/SFML/Graphics/Vertex.hpp
@@ -28,7 +28,6 @@
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
-#include <SFML/Graphics/Export.hpp>
 #include <SFML/Graphics/Color.hpp>
 #include <SFML/System/Vector2.hpp>
 
@@ -39,7 +38,7 @@ namespace sf
 /// \brief Define a point with color and texture coordinates
 ///
 ////////////////////////////////////////////////////////////
-class SFML_GRAPHICS_API Vertex
+class Vertex
 {
 public:
 
@@ -47,7 +46,7 @@ public:
     /// \brief Default constructor
     ///
     ////////////////////////////////////////////////////////////
-    Vertex();
+    constexpr Vertex();
 
     ////////////////////////////////////////////////////////////
     /// \brief Construct the vertex from its position
@@ -57,7 +56,7 @@ public:
     /// \param thePosition Vertex position
     ///
     ////////////////////////////////////////////////////////////
-    Vertex(const Vector2f& thePosition);
+    constexpr Vertex(const Vector2f& thePosition);
 
     ////////////////////////////////////////////////////////////
     /// \brief Construct the vertex from its position and color
@@ -68,7 +67,7 @@ public:
     /// \param theColor    Vertex color
     ///
     ////////////////////////////////////////////////////////////
-    Vertex(const Vector2f& thePosition, const Color& theColor);
+    constexpr Vertex(const Vector2f& thePosition, const Color& theColor);
 
     ////////////////////////////////////////////////////////////
     /// \brief Construct the vertex from its position and texture coordinates
@@ -79,7 +78,7 @@ public:
     /// \param theTexCoords Vertex texture coordinates
     ///
     ////////////////////////////////////////////////////////////
-    Vertex(const Vector2f& thePosition, const Vector2f& theTexCoords);
+    constexpr Vertex(const Vector2f& thePosition, const Vector2f& theTexCoords);
 
     ////////////////////////////////////////////////////////////
     /// \brief Construct the vertex from its position, color and texture coordinates
@@ -89,7 +88,7 @@ public:
     /// \param theTexCoords Vertex texture coordinates
     ///
     ////////////////////////////////////////////////////////////
-    Vertex(const Vector2f& thePosition, const Color& theColor, const Vector2f& theTexCoords);
+    constexpr Vertex(const Vector2f& thePosition, const Color& theColor, const Vector2f& theTexCoords);
 
     ////////////////////////////////////////////////////////////
     // Member data
@@ -98,6 +97,8 @@ public:
     Color     color;     //!< Color of the vertex
     Vector2f  texCoords; //!< Coordinates of the texture's pixel to map to the vertex
 };
+
+#include <SFML/Graphics/Vertex.inl>
 
 } // namespace sf
 
--- a/include/SFML/Graphics/Vertex.inl
+++ b/include/SFML/Graphics/Vertex.inl
@@ -22,16 +22,9 @@
 //
 ////////////////////////////////////////////////////////////
 
-////////////////////////////////////////////////////////////
-// Headers
-////////////////////////////////////////////////////////////
-#include <SFML/Graphics/Vertex.hpp>
 
-
-namespace sf
-{
 ////////////////////////////////////////////////////////////
-Vertex::Vertex() :
+constexpr Vertex::Vertex() :
 position (0, 0),
 color    (255, 255, 255),
 texCoords(0, 0)
@@ -40,7 +33,7 @@ texCoords(0, 0)
 
 
 ////////////////////////////////////////////////////////////
-Vertex::Vertex(const Vector2f& thePosition) :
+constexpr Vertex::Vertex(const Vector2f& thePosition) :
 position (thePosition),
 color    (255, 255, 255),
 texCoords(0, 0)
@@ -49,7 +42,7 @@ texCoords(0, 0)
 
 
 ////////////////////////////////////////////////////////////
-Vertex::Vertex(const Vector2f& thePosition, const Color& theColor) :
+constexpr Vertex::Vertex(const Vector2f& thePosition, const Color& theColor) :
 position (thePosition),
 color    (theColor),
 texCoords(0, 0)
@@ -58,7 +51,7 @@ texCoords(0, 0)
 
 
 ////////////////////////////////////////////////////////////
-Vertex::Vertex(const Vector2f& thePosition, const Vector2f& theTexCoords) :
+constexpr Vertex::Vertex(const Vector2f& thePosition, const Vector2f& theTexCoords) :
 position (thePosition),
 color    (255, 255, 255),
 texCoords(theTexCoords)
@@ -67,11 +60,9 @@ texCoords(theTexCoords)
 
 
 ////////////////////////////////////////////////////////////
-Vertex::Vertex(const Vector2f& thePosition, const Color& theColor, const Vector2f& theTexCoords) :
+constexpr Vertex::Vertex(const Vector2f& thePosition, const Color& theColor, const Vector2f& theTexCoords) :
 position (thePosition),
 color    (theColor),
 texCoords(theTexCoords)
 {
 }
-
-} // namespace sf
--- a/src/SFML/Graphics/CMakeLists.txt
+++ b/src/SFML/Graphics/CMakeLists.txt
@@ -47,8 +47,8 @@ set(SRC
     ${INCROOT}/Transformable.hpp
     ${SRCROOT}/View.cpp
     ${INCROOT}/View.hpp
-    ${SRCROOT}/Vertex.cpp
     ${INCROOT}/Vertex.hpp
+    ${INCROOT}/Vertex.inl
 )
 source_group("" FILES ${SRC})
 


Bromeon:
Basically still [this point](https://github.com/SFML/SFML/pull/1847#issuecomment-974899307) (copy-pasted):

---

> Turn `sf::Color`, `sf::Vector2`, `sf::Vector3`, `sf::Rect`, and `sf::Vertex` into `constexpr`-friendly and fully-`inline` classes.

For the templates this should be OK, but not sure about `Color` and `Vertex`, see [my comments here](https://github.com/SFML/SFML/pull/1741#issuecomment-762838738). For colors I understand the motivation to declare compile-time constants. But probably no one needs `constexpr` in `sf::Vertex`, so it would be a trade-off of run vs. compile-time with yet unknown benefits, and similar thoughts would be applicable to many other SFML classes.

---

Are there people concretely benefitting from this change?
Where do we draw the border?

Core utility classes like vectors, rects and colors are relatively easy to argue for, since they're so ubiquitious across the APIs, and there's a high chance that people want to declare compile-time constants for them (to tweak gameplay parameters, for example). This is not the case for `sf::Vertex`.

vittorioromeo:
Hey, thanks for sharing your thoughts. 

> But probably no one needs `constexpr` in `sf::Vertex`, so it would be a trade-off of run vs. compile-time with yet unknown benefits, and similar thoughts would be applicable to many other SFML classes.

I can think of some use cases for `constexpr` vertices. As an example, I might want to define a shape represented by some vertices with a `constexpr` array:

```cpp
constexpr std::array myShape{
    sf::Vertex{-100,  100},
    sf::Vertex{100,   100},
    sf::Vertex{100,  -100},
    sf::Vertex{-100, -100},
};
```

Reducing the run-time overhead of creating such array might be significant for low-power devices or to save on battery. On a similar note, we can now pre-compute shapes at compile-time and embed them into a binary, instead of having to compute them at run-time:

```cpp
template <std::size_t N>
constexpr std::array rotateBy90Degrees(int times, const std::array<sf::Vertex, N>& data) { /* ... */ };

constexpr std::array tetrisShape_LPiece{ /* ... */ };
constexpr std::array tetrisShape_LPiece_Rotations
{
    tetrisShape_LPiece,
    rotateBy90Degrees(1, tetrisShape_LPiece), // 90
    rotateBy90Degrees(2, tetrisShape_LPiece), // 180
    rotateBy90Degrees(3, tetrisShape_LPiece) // 270
};
```

The above might be a good way to precompute some shapes or masks in a situation where trading off memory for run-time overhead is highly favorable (again, e.g. low-power machines or saving battery).

Which brings me to my main point: there's no reason for `sf::Vertex` to *not* be `constexpr`: 
- It is just "plain old data". 
- It has no invariants. 
- The header dependency chain is very inexpensive. 

> Are there people concretely benefitting from this change? Where do we draw the border?

I've shown you two possible use cases before, and I don't really think we should limit people's creativity for the sake of it. There are individuals who have made [compile-time raytracers](https://github.com/tcbrindle/raytracer.hpp). It would be cool to see someone make a compile-time renderer that uses SFML and `sf::Vertex`. 

I strongly believe that making `sf::Vertex` a `constexpr` class does not pose any real risk, and does not impact compile-time negatively. It's pretty much an aggregate of some basic types, there are no expensive includes in any of its dependencies, and it might be useful in some cases and/or spark some creativity in clever people. Furthermore, having it inlined could also allow the compiler to more aggressively optimize code using `sf::Vertex`, which often needs to be as high-performance as possible to improve the framerate of an application.

I hope you understand my point of view :)

Bromeon:
Thanks for your elaborate answer!

> I can think of some use cases for `constexpr` vertices. As an example, I might want to define a shape represented by some vertices with a `constexpr` array [...] There are individuals who have made compile-time raytracers.

You could construct similar use cases for many classes in SFML: `ContextSettings`, `IpAddress`, `Transform`, etc.
I'm not sure if that is really something that SFML users want to do. SFML doesn't otherwise offer capabilities for compile-time computations, so a whole framework would need to be constructed around it anyway.


> Which brings me to my main point: there's no reason for `sf::Vertex` to _not_ be `constexpr`:

I see a few disadvantages:
1. If we follow the "constexpr everything we can" approach rather than "constexpr only what is undoubtedly useful for compile-time constants", we're turning part of SFML into a header-only library. A single .inl file may not add much, but when it needs to be parsed repeatedly, this adds up, and all headers used in the implementation need to be exposed.
1. `constexpr` is a commitment, removing the keyword is a breaking change. That means that APIs cannot be changed in a way that would no longer allow compile-time computation (e.g. caching, invoking non-constexpr functions, etc).
1. `constexpr` implies `inline`, so all the reasons why you would not want to make something inline apply here, too. This is particularly important because `constexpr` _does not imply compile-time-evaluated_. If the function parameters are only known at runtime, the function will be inlined, duplicating the body for each call-site. We don't add `inline` everywhere in SFML, just because we could.
1. Can we ensure that [compile-time vs. run-time discrepancies](https://youtu.be/fZjYCQ8dzTc?t=2013) are never happening? They are admittedly unlikely corner cases, but we also distribute precompiled binaries, and if this caused bugs, it would be an absolute nightmare.

I agree that for the current `sf::Vertex` implementation, the above-mentioned points have limited weight, as we currently only have constructors that assign fields. But we have to consider that we might extend this class, and if there's no .cpp file, new additions will naturally go into the .inl file (or we have the weird .hpp + .inl + .cpp triple).

Plus, as mentioned there are many similar candidates in SFML, so I think we need to find a pragmatic balance somewhere. There are good reasons to `constexpr` the most ubiquitious gamedev classes like `Color`, `Vector2`, `Vector3`, `Rect`, as they're very close to scalar types (and 3 of those 4 are already templates), but I don't think those reasons necessarily extend to all POD-like classes, especially once they have more methods.

vittorioromeo:
@Bromeon: thank you for sharing your concerns. 

---

> You could construct similar use cases for many classes in SFML: `ContextSettings`, `IpAddress`, `Transform`, etc. I'm not sure if that is really something that SFML users want to do. 

Or they never realized they could do it. We could potentially be limiting a user's creativity here, which I don't feel it's the right choice especially for a creative library.

> SFML doesn't otherwise offer capabilities for compile-time computations, so a whole framework would need to be constructed around it anyway.

I'm not sure what you mean here. Could you elaborate on *"a whole framework would need to be constructed around it"*?

> 1. If we follow the "constexpr everything we can" approach rather than "constexpr only what is undoubtedly useful for compile-time constants", we're turning part of SFML into a header-only library. 

Turning *"part of a library into a header-only library"* is not a meaningful statement. A library is either header-only, or it isn't -- there's no in-between. 

> A single .inl file may not add much, but when it needs to be parsed repeatedly, this adds up, and all headers used in the implementation need to be exposed.

Parsing is from expensive. Refer to the [rule of Chiel](https://old.reddit.com/r/cpp/comments/6gur2x/the_rule_of_chiel_aka_compiletime_cost_of/). The things that actually slow down compilation are mostly related to excessive instantiation of templates. And I did measure compilation times via ClangBuildAnalyzer (see #1893), even on the #1855 branch which makes things like `sf::Vertex` fully `constexpr` and `inline`. They don't even show up between the offenders of compilation time, we have bigger fish to fry if we care about that. 

My measurements have shown that this change does not negatively affect compilation times in any substantial way. If you can prove otherwise, I will reconsider it.

> 2. `constexpr` is a commitment, removing the keyword is a breaking change. That means that APIs cannot be changed in a way that would no longer allow compile-time computation (e.g. caching, invoking non-constexpr functions, etc).

Of course, and this is a good reason to not blindly use `constexpr` everywhere. Data types like `sf::Vertex` are IMHO the perfect candidate for `constexpr` use, as they have no invariants in their API. Being simple "plain old data" types, they will realistically never require any sort of action that cannot be performed at compile time. Do you expect to allocate dynamic memory in `sf::Vertex`? Do you expect to throw an exception? 

`constexpr` in C++17 can do pretty much anything except actions that would affect global state and I/O -- it's a good thing that we ensure those nasty side effects do not happen in a class like `sf::Vertex`. I welcome `constexpr` as a sanity check to avoid those!

> 3. `constexpr` implies `inline`, so all the reasons why you would not want to make something inline apply here, too. This is particularly important because `constexpr` _does not imply compile-time-evaluated_. If the function parameters are only known at runtime, the function will be inlined, duplicating the body for each call-site. We don't add `inline` everywhere in SFML, just because we could.

That is not how `inline` works. It does not instruct the compiler to inline -- it merely allows it to. The compiler will then only inline if it is deemed beneficial by looking at different heuristics. Please [read this answer](https://stackoverflow.com/a/1759575/598696).

`sf::Vertex` actually is a great place to use `inline`, as it is one of the classes that are often used to build up large datasets to send to the GPU for rendering. Any CPU transformation or preparation on these datasets might be better optimized by the compilers if they are given the possibility to `inline`.

> 4. Can we ensure that [compile-time vs. run-time discrepancies](https://youtu.be/fZjYCQ8dzTc?t=2013) are never happening? They are admittedly unlikely corner cases, but we also distribute precompiled binaries, and if this caused bugs, it would be an absolute nightmare.

Relying the correctness of a program on a specific floating-point representation of a `sf::Vertex` object is an unrealistic concern. Besides, that issue would already happen without `constexpr` when compiling that same program on a different machine/platform, or even when changing compilation flags (e.g. using `-ffast-math`). 

> I agree that for the current `sf::Vertex` implementation, the above-mentioned points have limited weight, as we currently only have constructors that assign fields. 

I agree with that assessment as well -- while your points are valid concerns in some specific situations, they are not concerns for common usages of `constexpr` IMHO, including `sf::Vertex` and `sf::Color`. 

> But we have to consider that we might extend this class, and if there's no .cpp file, new additions will naturally go into the .inl file (or we have the weird .hpp + .inl + .cpp triple).

I don't find the use of .hpp + .inl + .cpp "weird". Can you elaborate why you think it's weird? Regardless, `sf::Vertex` is clearly designed to be a plain-old-data type with no logic inside. 

> Plus, as mentioned there are many similar candidates in SFML, so I think we need to find a pragmatic balance somewhere. There are good reasons to `constexpr` the most ubiquitious gamedev classes like `Color`, `Vector2`, `Vector3`, `Rect`, as they're very close to scalar types (and 3 of those 4 are already templates), but I don't think those reasons necessarily extend to all POD-like classes, especially once they have more methods.

I agree that we should decide where to apply `constexpr` on a per-case basis, but I strongly think that we should lean towards "make it `constexpr`" rather than "don't make it `constexpr`". 

Creating a `constexpr` variable instead of a `const` one can be beneficial even just for readability, clearly stating to the reader that such object is statically-initialized and that its value is known at compile-time.

---

I hope I have convinced you that it would be worthwhile and very-low-risk to merge this PR. Please let me know if you have any further concerns.

Bromeon:
> Or they never realized they could do it. We could potentially be limiting a user's creativity here, which I don't feel it's the right choice especially for a creative library.

Sure, but this could be said about a lot of features we rejected in the past. It's always a balance, whether something is a very specialized case or useful for almost everyone. This more as a general statement, for `sf::Vertex` in particular we don't make the user experience considerably worse with `constexpr`, so I agree that the potential gains should be considered.

> I'm not sure what you mean here. Could you elaborate on _"a whole framework would need to be constructed around it"_?

I meant that having some functions marked `constexpr` is not enough to build heavy compile-time machinery, such as a raytracer. Most rendering- or state-related functionality in SFML is runtime only, it's unlikely that someone builds the whole rendering pipeline but wants to use precisely the SFML vertex struct 😉 

> Parsing is from expensive. Refer to the [rule of Chiel](https://old.reddit.com/r/cpp/comments/6gur2x/the_rule_of_chiel_aka_compiletime_cost_of/). The things that actually slow down compilation are mostly related to excessive instantiation of templates. And I did measure compilation times via ClangBuildAnalyzer (see #1893), even on the #1855 branch which makes things like `sf::Vertex` fully `constexpr` and `inline`. They don't even show up between the offenders of compilation time, we have bigger fish to fry if we care about that.
> 
> My measurements have shown that this change does not negatively affect compilation times in any substantial way. If you can prove otherwise, I will reconsider it.

Thanks for the explanation, and especially for measuring! I think you're right that the impact is negligible as long as the code is simple (not TMP for example) and no heavy headers such as `<algorithm>` are included.



> `constexpr` in C++17 can do pretty much anything except actions that would affect global state and I/O -- it's a good thing that we ensure those nasty side effects do not happen in a class like `sf::Vertex`. I welcome `constexpr` as a sanity check to avoid those!

That's an interesting take, I never viewed it from this angle 👍 




> That is not how `inline` works. It does not instruct the compiler to inline -- it merely allows it to. The compiler will then only inline if it is deemed beneficial by looking at different heuristics. Please [read this answer](https://stackoverflow.com/a/1759575/598696).

I know it's not an obligation, that's why there's `__forceinline` etc. But the comments on that answer seem to indicate that `inline` does make a difference, however it's only one of multiple factors for the compiler's decision. Linker settings (LTO/LTCG etc.) might also have a big influence.

> I don't find the use of .hpp + .inl + .cpp "weird". Can you elaborate why you think it's weird? Regardless, `sf::Vertex` is clearly designed to be a plain-old-data type with no logic inside.

I should maybe have said "unusual", at least I didn't encounter it often. But I'm not against embracing it as a pattern, if we find it makes sense that a class is half-header or half-cpp.

--- 



> I agree with that assessment as well -- while your points are valid concerns in some specific situations, they are not concerns for common usages of `constexpr` IMHO, including `sf::Vertex` and `sf::Color`.
> 
> I hope I have convinced you that it would be worthwhile and very-low-risk to merge this PR. Please let me know if you have any further concerns.

Thanks a lot for taking the time to elaborate. After reading all, I think it's OK to have `constexpr` for `sf::Vertex` -- while the average user probably doesn't gain much, there's no real damage done, so chances are that we enable certain new use cases.

The rule "is something POD-like?" (i.e. public member variables + constructors) may indeed be a good classifier to know whether `constexpr` makes sense. As you mentioned, the risk of more complex init logic is very low, also because the fields are already public.

Incidental question: what do you think about having POD implementations (constructors doing 1:1 field assignment) directly in the .hpp file? I used to be very strict about header/impl separation, but working with other languages (_none_ of which does anything close to C++) made me more open here. But I'm also fine with keeping .inl -- for templates or more complex cases, we should probably keep it anyway.

vittorioromeo:
@Bromeon: are you happy to accept this PR?

Bromeon (approved):
Yes, sorry if that wasn't clear! 🙂 

For future candidates, we might want to keep this in mind:

> The rule "is something POD-like?" (i.e. public member variables + constructors) may indeed be a good classifier to know whether `constexpr` is appropriate. As you mentioned, the risk of more complex init logic is very low, also because the fields are already public. On the other hand, it also means we might want to reconsider this when a POD grows and gets a lot of extra methods.

PR #1911, opened 2021-12-16
Use smart pointers to manage memory

vittorioromeo:

----

## Description

This PR aims to remove all raw `new`/`delete` calls and manage the memory automatically via smart pointers.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.

### DIFF
--- a/include/SFML/Audio/InputSoundFile.hpp
+++ b/include/SFML/Audio/InputSoundFile.hpp
@@ -30,6 +30,7 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Audio/Export.hpp>
 #include <SFML/System/Time.hpp>
+#include <memory>
 #include <string>
 #include <cstddef>
 
@@ -218,17 +219,32 @@ public:
     void close();
 
 private:
+    ////////////////////////////////////////////////////////////
+    /// \brief Deleter for input streams that only conditionally deletes
+    ///
+    ////////////////////////////////////////////////////////////
+    struct StreamDeleter
+    {
+        StreamDeleter(bool theOwned);
+
+        // To accept ownership transfer from usual std::unique_ptr<T>
+        template <typename T>
+        StreamDeleter(const std::default_delete<T>&);
+
+        void operator()(InputStream* ptr) const;
+
+        bool owned;
+    };
 
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    SoundFileReader* m_reader;       //!< Reader that handles I/O on the file's format
-    InputStream*     m_stream;       //!< Input stream used to access the file's data
-    bool             m_streamOwned;  //!< Is the stream internal or external?
-    Uint64           m_sampleOffset; //!< Sample Read Position
-    Uint64           m_sampleCount;  //!< Total number of samples in the file
-    unsigned int     m_channelCount; //!< Number of channels of the sound
-    unsigned int     m_sampleRate;   //!< Number of samples per second
+    std::unique_ptr<SoundFileReader>            m_reader;       //!< Reader that handles I/O on the file's format
+    std::unique_ptr<InputStream, StreamDeleter> m_stream;       //!< Input stream used to access the file's data
+    Uint64                                      m_sampleOffset; //!< Sample Read Position
+    Uint64                                      m_sampleCount;  //!< Total number of samples in the file
+    unsigned int                                m_channelCount; //!< Number of channels of the sound
+    unsigned int                                m_sampleRate;   //!< Number of samples per second
 };
 
 } // namespace sf
--- a/include/SFML/Audio/SoundFileFactory.hpp
+++ b/include/SFML/Audio/SoundFileFactory.hpp
@@ -29,6 +29,7 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Audio/Export.hpp>
+#include <memory>
 #include <string>
 #include <vector>
 
@@ -86,8 +87,6 @@ public:
     ////////////////////////////////////////////////////////////
     /// \brief Instantiate the right reader for the given file on disk
     ///
-    /// It's up to the caller to release the returned reader
-    ///
     /// \param filename Path of the sound file
     ///
     /// \return A new sound file reader that can read the given file, or null if no reader can handle it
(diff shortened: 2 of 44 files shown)


Bromeon (approved):
Awesome, thanks a lot for this refactoring! 👍 

One issue I sometimes encountered when migrating to `std::unique_ptr`, was that it required the full type to be known in the header (to implement the default deleter). Was this not an issue here? Or do the containing classes have a destructor defined in the .cpp file?

kimci86:
> One issue I sometimes encountered when migrating to std::unique_ptr, was that it required the full type to be known in the header (to implement the default deleter). Was this not an issue here? Or do the containing classes have a destructor defined in the .cpp file?

Right, CI shows this issue happens with MSVC Shared builds for the sf::Cursor class.

include/SFML/Audio/InputSoundFile.hpp
```diff
@@ -211,13 +212,13 @@ class SFML_AUDIO_API InputSoundFile : NonCopyable
...
-    SoundFileReader* m_reader;       //!< Reader that handles I/O on the file's format
-    InputStream*     m_stream;       //!< Input stream used to access the file's data
-    bool             m_streamOwned;  //!< Is the stream internal or external?
-    Uint64           m_sampleOffset; //!< Sample Read Position
-    Uint64           m_sampleCount;  //!< Total number of samples in the file
-    unsigned int     m_channelCount; //!< Number of channels of the sound
-    unsigned int     m_sampleRate;   //!< Number of samples per second
+    std::unique_ptr<SoundFileReader> m_reader;       //!< Reader that handles I/O on the file's format
+    InputStream*                     m_stream;       //!< Input stream used to access the file's data
+    bool                             m_streamOwned;  //!< Is the stream internal or external?
```
kimci86, line 217: `m_stream` could a unique_ptr with a custom deleter. See https://stackoverflow.com/a/24060783

vittorioromeo:
> One issue I sometimes encountered when migrating to `std::unique_ptr`, was that it required the full type to be known in the header (to implement the default deleter). Was this not an issue here? Or do the containing classes have a destructor defined in the .cpp file?

`std::unique_ptr<T>` requires `T` to be complete only during its destructor call. In practice this means that you can use a foward declaration in the header, and define the destructor in the source file.

Bromeon:
Yep, that's what I meant with
> Or do the containing classes have a destructor defined in the .cpp file?

So if this is the case here, good 👍

src/SFML/Graphics/RenderTexture.cpp
```diff
@@ -29,23 +29,17 @@
...
 
 namespace sf
 {
 ////////////////////////////////////////////////////////////
-RenderTexture::RenderTexture() :
-m_impl(nullptr)
-{
-
-}
+RenderTexture::RenderTexture() = default;
```
reconndev, line 38: How about moving `= default` to the class itself? You're losing the benefits of e.g. trivial construction doing it that way.

https://godbolt.org/z/EY6bbMTMM

vittorioromeo: I don't think it is a worthwhile change, the class is quite big anyway and trivial construction wouldn't bring any practical benefit unless I am missing something.

include/SFML/Audio/InputSoundFile.hpp
vittorioromeo: I've implemented that, thanks for the suggestion. Please take a look and let me know what you think :)

kimci86: Nice! The asymmetry between allocation with `new` and freeing with `std::default_delete` got me curious and I did some more tinkering. I got rid of the pointer casts in the process. See https://github.com/kimci86/SFML/commit/51574b4fce039625c9d87bb8b6b6851cbd30dfce.
Also I reordered some parts so that pointers are immediately deleted upon failure instead of waiting for the next call to `close()`. Feel to cherry-pick or take inspiration from what I did.

vittorioromeo: @kimci86: that's a great improvement -- it is so much cleaner. Thanks! I want to cherry-pick, but I am not totally certain that removing those `close()` calls is correct -- `close()` also resets the sound file attributes to zero in case of failure, and it doesn't seem like we do that anymore if we fail to e.g. read a file. Can you confirm this will not cause any issue?

kimci86: `close()` is called at the beginning of each `openFromXXX` method and sound file attributes are not modified unless opening succeeds so it looks to me that it is not needed to reset them on failure.

vittorioromeo: I am convinced, thanks! Will cherry pick ;)

src/SFML/Graphics/RenderTexture.cpp
Bromeon: Declaring `= delete` in the header is however what we did during https://github.com/SFML/SFML/pull/1901.
It could make sense to do the same here with `= default`. A compiler-generated constructor would behave the same.

kimci86: Having `RenderTexture() = default` in the header fails to compile (at least for me with GCC 9.3).

vittorioromeo:
The examples seem broken with this PR. I will convert to draft until I figure the issue out.

vittorioromeo:
> The examples seem broken with this PR. I will convert to draft until I figure the issue out.

Nevermind, I was missing the resource files in my examples build folder. But I am not getting any error messages unless I compile in debug mode. For instance, when running the shader example, I get the 

> Failed to load font "resources/tuffy.ttf" (failed to create the font face)

output if compiled in debug mode, but in release mode it just exits without any output. Why?

I can't find anything that makes `err()` logging conditional on debug mode in the code.

Anyway, this PR is not the culprit :)

include/SFML/Audio/SoundFileFactory.hpp
```diff
@@ -95,7 +96,7 @@ class SFML_AUDIO_API SoundFileFactory
     /// \see createReaderFromMemory, createReaderFromStream
     ///
     ////////////////////////////////////////////////////////////
-    static SoundFileReader* createReaderFromFilename(const std::string& filename);
+    static std::unique_ptr<SoundFileReader> createReaderFromFilename(const std::string& filename);
```
kimci86, line 99: In the documentation for this method and the following ones, this sentence should be removed:
> It's up to the caller to release the returned reader

vittorioromeo: Good catch, fixing...

vittorioromeo:
I've tested all examples and the 2.1.4 branch of Open Hexagon. Everything seems fine, let's merge!

PR #1912, opened 2021-12-16
Add tests for 'Vector' destructuring and 'constexpr' support

vittorioromeo:
## Description

Just making sure this stuff works as intended on every platform.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.

### DIFF
--- a/test/src/System/Vector2.cpp
+++ b/test/src/System/Vector2.cpp
@@ -1,5 +1,6 @@
 #include <SFML/System/Vector2.hpp>
 #include "SystemUtil.hpp"
+#include <type_traits>
 
 // Use sf::Vector2i for tests. Test coverage is given, as there are no template specializations.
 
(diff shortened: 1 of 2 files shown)


test/src/System/Vector2.cpp
```diff
@@ -145,4 +146,47 @@ TEST_CASE("sf::Vector2 class template", "[system]")
...
         }
     }
+
+    SECTION("Structured bindings")
+    {
+        sf::Vector2i vector(1, 2);
+
+        SECTION("destructure by value")
+        {
+            auto [x, y] = vector;
```
eXpl0it3r, line 156: Is this something, that can be broken by our code, meaning are we testing here our code or C++'s structured binding feature?

vittorioromeo: We are ensuring that the way `sf::Vector` is implemented allows usage of structured bindings. Turning `sf::Vector` from an aggregate into something else (e.g. by adding a `private` data member) would break this code. It's a sanity check for something that should always work, but could potentially be broken in a very unlikely scenario.

test/src/System/Vector2.cpp
```diff
@@ -145,4 +146,47 @@ TEST_CASE("sf::Vector2 class template", "[system]")
...
+            CHECK(vector.x == 3);
+        }
+    }
+
+    SECTION("Constexpr support")
+    {
+        constexpr sf::Vector2i vector(1, 2);
+        static_assert(vector.x == 1);
+        static_assert(vector.y == 2);
+        static_assert(vector + sf::Vector2i(2, 1) == sf::Vector2i(3, 3));
```
eXpl0it3r, line 190: Once we've updated Catch2 in #1898, we could also use `STATIC_REQUIRE`

vittorioromeo: Is there any practical benefit over `static_assert`?

test/src/System/Vector3.cpp
```diff
@@ -158,4 +159,50 @@ TEST_CASE("sf::Vector3 class template", "[system]")
...
+            static_assert(std::is_same_v<decltype(x), decltype(vector.x)>);
+
+            x = 3;
+
+            CHECK(x == 3);
+            CHECK(vector.x == 3);
+        }
+    }
+
+    SECTION("Constexpr support")
```
Bromeon, line 200: Would it make sense to add a tiny test for the `constexpr` conversion constructor (e.g. `float` -> `int`)?

Bromeon (approved):

test/src/System/Vector3.cpp
vittorioromeo: We already have a `"Conversion constructor"` section, the fact that it is now `constexpr` doesn't affect its behavior at all.

test/src/System/Vector2.cpp
eXpl0it3r: It's properly tracked as Catch2 assertion. Currently the test will just fail because of C++, but Catch2 won't really "know" about it.

See https://github.com/catchorg/Catch2/issues/1356

vittorioromeo: I see, I am happy to update in a future PR.

PR #1913, opened 2021-12-16
Use more smart pointers to manage FreeType resources

kimci86:
## Description

My attempt to use more smart pointers to manage FreeType resources in sf::Font class.
Potential addition to or inspiration for @vittorioromeo's PR #1911

### DIFF
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -41,6 +41,7 @@
 #include <cstdlib>
 #include <cstring>
 #include <cmath>
+#include <type_traits>
 
 
 namespace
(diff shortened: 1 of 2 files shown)


src/SFML/Graphics/Font.cpp
```diff
@@ -83,13 +84,32 @@ namespace
...
+        void operator()(FT_Library library) { FT_Done_FreeType(library); }
+        void operator()(FT_Face face)       { FT_Done_Face(face); }
+        void operator()(FT_Stroker stroker) { FT_Stroker_Done(stroker); }
+    };
+
+public:
+    std::unique_ptr<std::remove_pointer_t<FT_Library>, Deleter> library   = nullptr; // Pointer to the internal library interface
+    std::unique_ptr<FT_StreamRec>                               streamRec = nullptr; // Pointer to the stream rec instance
+    std::unique_ptr<std::remove_pointer_t<FT_Face>,    Deleter> face      = nullptr; // Pointer to the internal font face
+    std::unique_ptr<std::remove_pointer_t<FT_Stroker>, Deleter> stroker   = nullptr; // Pointer to the stroker
```
vittorioromeo, line 105:
```suggestion
    std::unique_ptr<std::remove_pointer_t<FT_Library>, Deleter> library;   //< Pointer to the internal library interface
    std::unique_ptr<FT_StreamRec>                               streamRec; //< Pointer to the stream rec instance
    std::unique_ptr<std::remove_pointer_t<FT_Face>,    Deleter> face;      //< Pointer to the internal font face
    std::unique_ptr<std::remove_pointer_t<FT_Stroker>, Deleter> stroker;   //< Pointer to the stroker
```

vittorioromeo: Is it guaranteed that `FT_XXX` types are always pointers on every platform? Could they be something else (e.g. handles)?

src/SFML/Graphics/Font.cpp
```diff
@@ -83,13 +84,32 @@ namespace
...
+} // namespace priv
+
 ////////////////////////////////////////////////////////////
 Font::Font() :
-m_library  (nullptr),
-m_face     (nullptr),
-m_streamRec(nullptr),
-m_stroker  (nullptr),
-m_refCount (),
+m_font     (nullptr),
```
vittorioromeo, line 112:
```suggestion
m_font     (),
```

vittorioromeo (changes requested):
I like this approach, it's much better than what I was going for. I'll merge it in as soon as some minor things are addressed :)

src/SFML/Graphics/Font.cpp
kimci86: It might be a hack to assume so because those types are documented as opaque handles, but looking at FreeType headers, they are indeed typedefs to pointer types regardless of the platform.

Bromeon (approved):
Nice, a very elegant approach to avoid all the type erasure!

src/SFML/Graphics/Font.cpp
kimci86: I added some static assertions to make sure those types are pointer types.

kimci86:
I squashed and rebased my changes.

vittorioromeo (approved):
Great job, I will merge into my draft PR. Thanks!

src/SFML/Graphics/Font.cpp
vittorioromeo: > I added some static assertions to make sure those types are pointer types.

That is not what I meant, I meant checking out the FreeType documentation. The code wouldn't compile with or without the static assertion. But if, for example `FT_Library` is a pointer on Win32 and a `int` on Linux, your approach wouldn't work.

PR #1916, opened 2021-12-19
Only compile Catch main once

ChrisThrasher:

----

## Description

Because Catch is so expensive to compile, it's really helpful for initial compile times if this target doesn't have to be recompiled so often. This won't make a huge impact on developers who are mostly doing incremental builds but can have a meaningful impact on CI runners that are always doing clean builds.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,44 +1,45 @@
 set(SRCROOT "${PROJECT_SOURCE_DIR}/test/src")
 
-include_directories("${PROJECT_SOURCE_DIR}/include")
-include_directories("${PROJECT_SOURCE_DIR}/extlibs/headers")
-include_directories("${SRCROOT}/TestUtilities")
+add_library(sfml-test-config "${SRCROOT}/CatchMain.cpp")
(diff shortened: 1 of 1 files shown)


ChrisThrasher:
@vittorioromeo Here's an atomic PR for you while I work on rearranging some commits for #1898

ChrisThrasher:
On second thought, I will include this in #1898 instead since.

ChrisThrasher:
Reopening to prevent having more than one commit in a given pull request.

ChrisThrasher:
PR still shows as closed. Maybe that's because I deleted the remote branch then recreated it. I'm not sure. The "Reopen and comment" button is greyed out.

vittorioromeo:
> PR still shows as closed. Maybe that's because I deleted the remote branch then recreated it. I'm not sure. The "Reopen and comment" button is greyed out.

Feel free to create another. But I would rename `sfml-test-config` to `sfml-test-main`, it seems more appropriate.

ChrisThrasher:
See #1918

ChrisThrasher:
@eXpl0it3r This can be removed from the 3.0 project.

PR #1919, opened 2021-12-19
Don't use 'sf::err' in public example code

vittorioromeo:
`sf::err` is an *internal* error message stream, so it should not be used in example code. To be consistent with other examples, I've changed it to `std::cerr`. Includes have just been alphabetically ordered.

### DIFF
--- a/examples/vulkan/Vulkan.cpp
+++ b/examples/vulkan/Vulkan.cpp
@@ -9,10 +9,11 @@
 #include <SFML/Graphics.hpp>
 
 #include <SFML/Window.hpp>
-#include <vector>
+#include <iostream>
 #include <limits>
-#include <cstring>
+#include <vector>
 #include <cmath>
+#include <cstring>
 
 
 ////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


Bromeon:
Maybe related to this, we currently display `sf::err()` in the public API documentation: 
https://www.sfml-dev.org/documentation/2.5.1/group__system.php#ga7fe7f475639e26334606b5142c29551f

If we intend it to be internal-only, should we hide it from the doc?

Bromeon (approved):

vittorioromeo:
> Maybe related to this, we currently display `sf::err()` in the public API documentation: https://www.sfml-dev.org/documentation/2.5.1/group__system.php#ga7fe7f475639e26334606b5142c29551f
> 
> If we intend it to be internal-only, should we hide it from the doc? Or at least mention its intended purpose?

Hmm... we could also move it to `src`.

eXpl0it3r:
Please wait for #1920 before merging this 🙂

eXpl0it3r (approved):

PR #1920, opened 2021-12-20
Feature/backmerge

eXpl0it3r:
## Description

Merging changes from the `2.6.x` branch back to `master`

Please don't merge things to master in the meantime, so we can cleanly backmerge things.

## How to test this PR?

Successful CI run


eXpl0it3r:
@vittorioromeo can you provide conversion fixes for your custom `priv::copy` implementation?

The way to handle this PR:
- `--amend` changes to the merge commit
- Force push the change merge commit
- Once everything is green from CI it will still complain that it can't be merged
- I'll need to temporarily allow merge-commit merging
- Try to avoid merging stuff to master in the meantime

vittorioromeo (approved):

include/SFML/System/Utf.inl
```diff
@@ -38,9 +38,18 @@
 template <typename InputIt, typename OutputIt>
 OutputIt priv::copy(InputIt first, InputIt last, OutputIt d_first)
 {
+#if defined(__clang__)
+    #pragma clang diagnostic push
+    #pragma clang diagnostic ignored "-Wsign-conversion"
+#endif
```
reconndev, line 44: Is it a good idea to silence this warning for all usages of this algorithm?

I feel like it's not the algorithm that's the problem here.

vittorioromeo: All the usages of this algorithm are limited to this file. I would "fix" the warning by saying:

```cpp
    while (first != last)
        *d_first++ = static_cast<decltype(*d_first)>(*first++);
```

but I cannot use C++11 here and I don't want to get `iterator_traits` involved.

eXpl0it3r: Oh, you can use C++11, since this is a backmerge into the master branch. You can't use it on the 2.6.x branch, but that change doesn't exist there

vittorioromeo: Oh, I had it in my head that I was limited to C++03 for this PR. Sorry. I will try removing the pragmas.

vittorioromeo:
@eXpl0it3r @reconndev should be good to go

PR #1924, opened 2021-12-21
Manipulate library names on Windows + Clang

### LINKED ISSUE #1923: Using Clang + MSYS2 on Windows generates shared library names starting with "lib"
This seems inconsistent with what GCC does. For example, compiling with GCC results in `sfml-system-3.dll`, while compiling with Clang results in `libsfml-system-3.dll`. I suspect some mistake in the CMakeLists.

vittorioromeo:
Fixes #1923.

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -102,8 +102,8 @@ macro(sfml_add_library target)
         else()
             set_target_properties(${target} PROPERTIES DEBUG_POSTFIX -d)
         endif()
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
Can you rebase/cherry-pick it onto `2.6.x` and change the target branch?

vittorioromeo:
@eXpl0it3r: should be good now.

eXpl0it3r (approved):

PR #1927, opened 2021-12-21
Use a smart pointer to manage std::FILE pointer in FileInputStream

kimci86:
## Description

This is a small addition to the now merged PR #1911 
I found one more opportunity to use a smart pointer.

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI and proof-reading should be enough.

### DIFF
--- a/include/SFML/System/FileInputStream.hpp
+++ b/include/SFML/System/FileInputStream.hpp
@@ -137,7 +137,16 @@ private:
 #ifdef SFML_SYSTEM_ANDROID
     std::unique_ptr<priv::ResourceStream> m_file;
 #else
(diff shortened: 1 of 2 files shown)


Bromeon:
Good idea!

I was first thinking to use `std::fclose` directly as a deleter like [in this Gist](https://gist.github.com/t-mat/5849549), however that bears the risk of forgetting to initialize the deleter in the constructor; a functor might thus be safer.

eXpl0it3r:

vittorioromeo (approved):
LGTM.

PR #1928, opened 2021-12-22
Stop configuration when GCC is detected on macOS

ChrisThrasher:
## Description

The GCC build on macOS fails because `-stdlib=libc++` as specified on [cmake/Macros.cmake:27](https://github.com/SFML/SFML/blob/master/cmake/Macros.cmake#L27) is an unrecognized command line option. If you hack around this, the compilation fails like so:
```
[  8%] Building CXX object src/SFML/Window/CMakeFiles/sfml-window.dir/Joystick.cpp.o
In file included from /Users/thrasher/Projects/sfml/src/SFML/Window/OSX/JoystickImpl.hpp:34,
                 from /Users/thrasher/Projects/sfml/src/SFML/Window/JoystickImpl.hpp:88,
                 from /Users/thrasher/Projects/sfml/src/SFML/Window/JoystickManager.hpp:32,
                 from /Users/thrasher/Projects/sfml/src/SFML/Window/Joystick.cpp:29:
/Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk/System/Library/Frameworks/IOKit.framework/Headers/hid/IOHIDDevice.h:295:34: error: 'dispatch_block_t' has not been declared
  295 |                                  dispatch_block_t               handler)
      |                                  ^~~~~~~~~~~~~~~~
```

I can't find a fix that doesn't [require modifying system headers](https://hamelot.io/programming/osx-gcc-dispatch_block_t-has-not-been-declared-invalid-typedef/) so the next best thing is to explicitly fail the build when this scenario is detected so no one is confused about whether this is a valid configuration.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android


eXpl0it3r:
I'm a bit conflicted about a hard block. It would certainly help the user to point out that it's not a supported process, but if they want to manually patch stuff, then should we actively stand in their way?

What's the workaround for libc++ exactly?

Since Maverick [one apparently](https://stackoverflow.com/questions/19774778/when-is-it-necessary-to-use-the-flag-stdlib-libstdc) doesn't/shouldn't specify `-stdlib=libc++` anyways. Would it make sense to have some SDK version check instead of the GCC check?

If GCC is detected, I could see us simply emitting a warning like `message(WARNING "Woops")` and then let CMake/GCC fail the usual way. That way if the user wants to patch something on their own, they can, but they've been warned. 😄

ChrisThrasher:
It is weird that we manually specify a standard library. I've never seen a CMake project need to do that.

A warning would suffice too. I just want some kind of intentional action taken when this known bad config is detected.

eXpl0it3r:

ChrisThrasher:
A simpler approach would be something like this:
```cmake
    if (SFML_OS_MACOSX)
        if (${CMAKE_GENERATOR} MATCHES "Xcode")
            sfml_set_xcode_property(${target} CLANG_CXX_LIBRARY "libc++")
        elseif(NOT SFML_COMPILER_CLANG)
            message(FATAL_ERROR "Clang is the only supported compiler on macOS")
        endif()
    endif()
```
which lets also skip that redundant `-stdlib=` flag which shouldn't be necessary anyway. `FATAL_ERROR` can certainly be `WARN` if we want to be a bit more patient with people who may want to hack around with fixing the GCC macOS build.

ChrisThrasher:
The current build failure looks like this:
```
[  1%] Building CXX object src/SFML/System/CMakeFiles/sfml-system.dir/Clock.cpp.o
g++-11: error: unrecognized command-line option '-stdlib=libc++'
make[2]: *** [src/SFML/System/CMakeFiles/sfml-system.dir/Clock.cpp.o] Error 1
make[1]: *** [src/SFML/System/CMakeFiles/sfml-system.dir/all] Error 2
make: *** [all] Error 2
```
so an explicit CMake error is a lot more clear. If someone wants to hack around with using GCC, either way they'll have to patch the build to allow it to continue onto whatever the first error is.

vittorioromeo:

eXpl0it3r (approved):

PR #1925, opened 2021-12-21
Add 'NetworkUtil' for network module tests

vittorioromeo:
Depends on #1921, I plan to add more tests here when we add stuff like move semantics so I wanted to make this module consistent with the rest.

PR #1932, opened 2021-12-23
Add move semantics to 'System' module

vittorioromeo:
## Description

This PR adds move semantics to the 'System' module. The only eligible class is `FileInputStream`. It was lacking tests, so I've also added some.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.


vittorioromeo:
OSX build fails because `<filesystem>` is not supported, with this error:
>  error: 'path' is unavailable: introduced in macOS 10.15

raspbian-gcc-armhf seems to fail to link against `<filesystem>` sources:
> /home/pi/buildbot/raspbian-gcc-armhf/build/test/src/TestUtilities/SystemUtil.cpp:37: undefined reference to `std::filesystem::temp_directory_path[abi:cxx11]()'

`<filesystem>` is a standard C++17 facility -- could we upgrade our CI to a compliant compiler?

test/src/TestUtilities/SystemUtil.hpp
```diff
@@ -42,4 +43,31 @@ namespace Catch
...
+        // Close and delete the generated file.
+        ~TemporaryFile();
+
+        // Prevent copies.
+        TemporaryFile(const TemporaryFile&) = delete;
+        TemporaryFile& operator=(const TemporaryFile&) = delete;
+
+        // Allow moves.
+        TemporaryFile(TemporaryFile&&) = default;
+        TemporaryFile& operator=(TemporaryFile&&) = default;
```
kimci86, line 66: Shouldn't those move operations swap each `m_path` explicitly?

eXpl0it3r:
> OSX build fails because `<filesystem>` is not supported, with this error

https://github.com/SFML/SFML-Buildbot/pull/31

> raspbian-gcc-armhf seems to fail to link against `<filesystem>` sources

Updating to Raspbian Bullseye right now (takes quite a while...)

vittorioromeo:
> > OSX build fails because `<filesystem>` is not supported, with this error
> 
> [SFML/SFML-Buildbot#31](https://github.com/SFML/SFML-Buildbot/pull/31)
> 
> > raspbian-gcc-armhf seems to fail to link against `<filesystem>` sources
> 
> Updating to Raspbian Bullseye right now (takes quite a while...)

Awesome. Thanks for the quick update.

test/src/TestUtilities/SystemUtil.hpp
vittorioromeo: That's not the real issue, the issue is that the destructor would be invoked for the moved-from object and misbehave. I have removed the move operations as they are not currently used in this PR.

vittorioromeo:
@Bromeon @eXpl0it3r @binary1248: anyone that can approve and merge if happy with the changes?

eXpl0it3r (approved):

eXpl0it3r:
@vittorioromeo Only noticed just now, that your commit log message doesn't actually match the content here, I guess the message was related to #1925 😉

PR #1934, opened 2021-12-23
Make 'Time' a 'constexpr' class

vittorioromeo:
## Description

This PR turns `sf::Time` into a class with `constexpr` support. Same principles and consideration as in #1903, #1904, and #1910 apply for this PR.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.


include/SFML/System/Time.hpp
```diff
@@ -361,9 +361,9 @@ SFML_SYSTEM_API Time& operator *=(Time& left, Int64 right);
 /// \return \a left divided by \a right
 ///
 ////////////////////////////////////////////////////////////
-SFML_SYSTEM_API Time operator /(Time left, float right);
+[[nodiscard]] constexpr Time operator /(Time left, float right);
 
-////////////////////////////////////////////////////////////
+/////////////////////////[[nodiscard]] ///////////////////////////////////
```
Bromeon, line 366: probably unintended 🙂

Bromeon (changes requested):
Looks good, just a minor formality 👍 

A question, it seems like the .cpp -> .inl rename was not detected by Git as such (probably because of the many `constexpr` lines changing). Are we OK with this, or do we want to enable continuous Git history by having a separate rename-only commit?

It's not something we considered in the past, so I'm absolutely fine if we keep things as they were; just thought I'd quickly bring it up.

include/SFML/System/Time.hpp
vittorioromeo: Whoops :)

include/SFML/System/Time.hpp
```diff
@@ -37,7 +37,7 @@ namespace sf
 /// \brief Represents a time value
 ///
 ////////////////////////////////////////////////////////////
-class SFML_SYSTEM_API Time
+class Time
```
eXpl0it3r, line 40: Doesn't the class still need to be exported?

src/SFML/System/Time.cpp
```diff
@@ -31,230 +31,8 @@
...
-    return microseconds(left.asMicroseconds() % right.asMicroseconds());
-}
-
-
-////////////////////////////////////////////////////////////
-Time& operator %=(Time& left, Time right)
-{
-    return left = left % right;
-}
+const Time Time::Zero;
```
Bromeon, line 36: If we added this in the header, we could rename the whole file .cpp -> .inl, and wouldn't need to export any more symbols.

```cpp
constexpr Time Time::Zero(0);
```

Example:
https://onlinegdb.com/5uK6_OF8a

include/SFML/System/Time.hpp
vittorioromeo: No, it's an inline class now. It's the same as in #1904.

src/SFML/System/Time.cpp
vittorioromeo: This is the same problem as with `sf::Color`, see #1904 for discussion. I remember trying the workaround you suggested, but it failed CI at one point. I will retry in a separate PR for both `sf::Color` and `sf::Time`.

Bromeon: Yes, I thought might be easier for you than having to do the administrative stuff twice. Wasn't aware of CI problems.

But your suggestion sounds good to me👍

vittorioromeo: @Bromeon: please don't forget to approve otherwise I cannot merge :)

vittorioromeo: @Bromeon: ping

Bromeon (approved):

PR #1935, opened 2021-12-23
Add tests for sf::VideoMode

ChrisThrasher:
## Description

I added tests for `sf::VideoMode`. I was unsure how to test `sf::VideoMode::getDesktopMode`, `sf::VideoMode::getFullscreenMode`, and `sf::VideoMode::isValid` in a way that would work on all machines including the presumably headless CI runners. The `test-sfml-window` executable has always been compiled but until now didn't actually include any test assertions so now this test executable will provide some meaningful feedback.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/test/Window/VideoMode.cpp
+++ b/test/Window/VideoMode.cpp
@@ -1,0 +1,100 @@
+#include <SFML/Window/VideoMode.hpp>
+
+#include <doctest.h>
+#include <ostream>
+
+namespace sf
+{
+std::ostream& operator <<(std::ostream& os, const sf::VideoMode& videoMode)
+{
+    os << videoMode.width << " x " << videoMode.height << " x " << videoMode.bitsPerPixel;
+    return os;
+}
+}
+
+TEST_CASE("sf::VideoMode class - [window]")
+{
+    SUBCASE("Construction")
+    {
+        SUBCASE("Default constructor")
+        {
+            sf::VideoMode videoMode;
+            CHECK(videoMode.width == 0);
+            CHECK(videoMode.height == 0);
+            CHECK(videoMode.bitsPerPixel == 0);
+        }
+
+        SUBCASE("Width, height constructor")
+        {
+            sf::VideoMode videoMode(800, 600);
+            CHECK(videoMode.width == 800);
+            CHECK(videoMode.height == 600);
+            CHECK(videoMode.bitsPerPixel == 32);
+        }
+
+        SUBCASE("Width, height, bit depth constructor")
+        {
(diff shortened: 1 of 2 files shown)


eXpl0it3r:

vittorioromeo (approved):
LGTM, thanks!

eXpl0it3r:

test/Window/VideoMode.cpp
```diff
@@ -0,0 +1,100 @@
...
+            CHECK(sf::VideoMode(720, 720) != sf::VideoMode(720, 720, 24));
+            CHECK(sf::VideoMode(1080, 1920, 16) != sf::VideoMode(400, 600));
+        }
+
+        SUBCASE("operator<")
+        {
+            CHECK(sf::VideoMode() < sf::VideoMode(0, 0, 1));
+            CHECK(sf::VideoMode(1080, 1920, 24) < sf::VideoMode(800, 800, 48));
+            CHECK(sf::VideoMode(400, 400, 48) < sf::VideoMode(400, 600, 48));
+            CHECK(sf::VideoMode(400, 400, 48) < sf::VideoMode(600, 400, 48));
```
Bromeon, line 67: [Our `operator<`](https://github.com/SFML/SFML/blob/master/src/SFML/Window/VideoMode.cpp#L106-L123) considers, in this order:
* bits per pixel
* width
* height

So we should have the following test cases, which all resolve to `true` for `operator<`:
* bits less, width greater, height greater
* bits equal, width less, height greater
* bits equal, width equal, height less

And maybe we could additionally check that two equal ones are never less-than each other. Or there might even be a predefined way to test ordering relations for their consistency (e.g. what `std::map` does when using them as keys).

Same applies for other operators.

ChrisThrasher: This is great feedback. I tweaked the `operator<` and `operator>` tests to include these operations. I included these improved assertions in the `operator<=` and `operator>=` tests as well.

ChrisThrasher:
By the way, the comparison logic can be vastly simplified with `std::tie` like this:

https://github.com/ChrisThrasher/SFML/commit/253ba16d2de0e84994518ca107adb62ece548070

Either way the logic is still correct but this pattern is a lot easier to get right if you're writing `operator<` from scratch.

eXpl0it3r:

ChrisThrasher:
Any other comments on this? @vittorioromeo @Bromeon @eXpl0it3r

vittorioromeo (approved):

PR #1936, opened 2021-12-23
Add tests for sf::Clock

ChrisThrasher:
## Description

I added tests for `sf::Clock`. This was a little tricky. `sf::Clock` is implemented in terms of `std::chrono` and `std::chrono` is being used to test it. This shouldn't be a problem though since we must assume that our standard library implementation is correct. There's no getting around that assumption.

The idea for these tests is that if you periodically check the state of the timer at 100 millisecond intervals you can check that the clock is reporting roughly the elapsed time one would expect. I used a rather generous 10 millisecond margin since I'm not sure what level of accuracy and precision we'd expect. On my machine I see <5 millisecond deviations so I doubled that to be safe since flaky tests are a huge problem.

Another test philosophy would be to measure the elapsed time, sleep a few milliseconds, measure again, then test that the elapsed time increased. This is less prone to flakiness but is also a less rigorous test that what is currently written.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/test/System/Clock.cpp
+++ b/test/System/Clock.cpp
@@ -1,0 +1,26 @@
+#include <SFML/System/Clock.hpp>
+#include "SystemUtil.hpp"
+#include <thread>
+
+#include <doctest.h>
+
+TEST_CASE("sf::Clock class - [system]")
+{
+    SUBCASE("getElapsedTime()")
+    {
+        const sf::Clock clock;
+        CHECK(clock.getElapsedTime() >= sf::microseconds(0));
+        const auto elapsed = clock.getElapsedTime();
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+        CHECK(clock.getElapsedTime() > elapsed);
+    }
+
+    SUBCASE("restart()")
+    {
+        sf::Clock clock;
+        CHECK(clock.restart() >= sf::microseconds(0));
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+        const auto elapsed = clock.restart();
+        CHECK(clock.restart() < elapsed);
+    }
+}
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -8,6 +8,7 @@ target_compile_features(sfml-test-main PRIVATE cxx_std_17)
 
 # System is always built
 SET(SYSTEM_SRC
+    "${SRCROOT}/System/Clock.cpp"
     "${SRCROOT}/System/FileInputStream.cpp"
     "${SRCROOT}/System/Time.cpp"
     "${SRCROOT}/System/Vector2.cpp"


ChrisThrasher:
Already getting some flakiness showing up. I don't think my approach is ideal. Eager to hear how y'all think we can test this.

vittorioromeo:
I'm not comfortable with sleeping in tests, as you can see it is very flaky and it does increase the execution time of CI. I would definitely have some "sanity checks" for the basic API of `sf::Clock` that do not require any sleeping, but I am not sure what the best way of testing this would be otherwise.

Regardless, if we keep using sleep, I'd lower the sleeping time significantly (e.g. 1ms).

Bromeon:
We might consider running the tests in a separate thread or even process. This would solve both "delays CI" and "too small precision" problems. You can then easily sleep for 1s, so you won't be affected by the precision issue that much.

The problem is, [Doctest doesn't seem to support parallel test execution out-of-the-box](https://github.com/doctest/doctest/blob/master/doc/markdown/faq.md#is-doctest-thread-aware):

> Most macros/functionality is safe to use in a multithreaded context: assertion and logging macros can be safely used from multiple threads spawned from a single test case. **This however does not mean that multiple test cases can be ran in parallel - test cases are still ran serially.**

So it might need some more work. Like running some tests as a separate bash command, as suggested [here](https://github.com/doctest/doctest/issues/168) or [here](https://github.com/doctest/doctest/issues/184). Or we would use _ordered_ test execution, spin up the thread in an early test, and await its result in a late test (where it's already available).

It may all sound too much work for the benefit, but it would give us some confidence on the different clock implementations, as introduced in https://github.com/SFML/SFML/pull/1872. Otherwise I don't see us keep an overview over the various Chrono corner cases and limitations.

eXpl0it3r:

ChrisThrasher:
I retooled the tests to be less rigorous in their checks but also (hopefully) much less flaky. I still need some sleeps to guarantee that timer can measure a non-zero duration but now we're only sleeping for a few milliseconds which is probably too short to notice and only adds a tenth of a second to the test suite.

eXpl0it3r:

vittorioromeo:

test/System/Clock.cpp
```diff
@@ -0,0 +1,32 @@
+#include <SFML/System/Clock.hpp>
+
+#include <doctest.h>
+#include <chrono>
+#include <ostream>
+#include <thread>
+
+namespace sf
+{
+std::ostream& operator <<(std::ostream& os, const sf::Time& time);
```
vittorioromeo, line 10: Is this defined anywhere?

ChrisThrasher: Yes in test/System/Time.cpp

vittorioromeo: This is where a utilities header would make more sense...

ChrisThrasher: Is there a reason this operator isn't part of the library itself?

vittorioromeo: > Is there a reason this operator isn't part of the library itself?

You can't really assume that everybody wants to print `sf::Time` in the same way, and also you would introduce a dependency on `<ostream>`. I wouldn't add this operator to the library.

ChrisThrasher: Fair point. Fixed.

vittorioromeo:

ChrisThrasher:
The MinGW builds are failing to measure a 100 microsecond sleep. Does that indicate a bug on those systems? Should we go back to sleeping for some number of milliseconds even though`sf::Clock` advertises microsecond precision? I'll increase the sleep to 1 millisecond in hopes that those MinGW tests stop failing.

Bromeon:
> The MinGW builds are failing to measure a 100 microsecond sleep. Does that indicate a bug on those systems?

Well possible. We know that Chrono is implemented very inconsistently across platforms, wouldn't surprise me if `this_thread::sleep_for()` is unreliable, too.

I'm not sure if this is the implementation that's actually used (or even the correct repo), but [this would explain the millisecond precision](https://github.com/mirror/mingw-w64/blob/518d3a79d208e194c215df29a111e9d3727ac629/mingw-w64-crt/misc/mingw_usleep.c#L10-L17):
```cpp
int __cdecl
usleep (useconds_t us)
{
  if (us != 0)
    Sleep (us / 1000);  // <-----

  return 0;
}
```

---

> Should we go back to sleeping for some number of milliseconds even though`sf::Clock` advertises microsecond precision? I'll increase the sleep to 1 millisecond in hopes that those MinGW tests stop failing.

I don't think that's a good idea -- the whole point of these tests is to ensure portability. By increasing sleep duration, we also worsen test coverage on other platforms, and might miss future regressions.

ChrisThrasher:
If sleeping for milliseconds isn't acceptable and having failing CI jobs isn't acceptable, what else can we do? These tests aren't ideal but they're a lot better than nothing.

Bromeon:
We can fix the implementation (as suggested, using `sf::sleep()` with different impls depending on system).

But this shouldn't block this PR, and I agree some tests are better than none. Maybe we can leave it at 1000us now and create an issue for MinGW sleep?

ChrisThrasher:
> Maybe we can leave it at 1000us now and create an issue for MinGW sleep?
 
I can't think of anything better. An issue on repo will record our findings and the tests as-is will serve as a starting point for whoever wants to improve them later.

Bromeon:
Created issue, feel free to add info/code snippet if you want 🙂

ChrisThrasher:
@vittorioromeo @Bromeon Could we run the full CI pipeline on this to make sure the MinGW tests still pass?

eXpl0it3r:

vittorioromeo:

eXpl0it3r:

eXpl0it3r:

vittorioromeo (approved):

PR #1937, opened 2021-12-24
Make 'Color' constants 'constexpr' and add tests

vittorioromeo:
## Description

Follow-up to #1904, using @Bromeon's suggestion to make constants `constexpr`, and also added some tests.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.


vittorioromeo:
@Bromeon: this is what I meant... I remembered trying this, but VS2017 is broken :/

include/SFML/Graphics/Color.inl
```diff
@@ -147,3 +147,17 @@ constexpr Color& operator *=(Color& left, const Color& right)
...
+////////////////////////////////////////////////////////////
+inline constexpr Color Color::Black(0, 0, 0);
+inline constexpr Color Color::White(255, 255, 255);
+inline constexpr Color Color::Red(255, 0, 0);
+inline constexpr Color Color::Green(0, 255, 0);
+inline constexpr Color Color::Blue(0, 0, 255);
+inline constexpr Color Color::Yellow(255, 255, 0);
+inline constexpr Color Color::Magenta(255, 0, 255);
+inline constexpr Color Color::Cyan(0, 255, 255);
+inline constexpr Color Color::Transparent(0, 0, 0, 0);
```
Bromeon, line 163: `constexpr` implies `inline`, so it should work without that keyword.

eXpl0it3r: I feel like we're going in circles 😄 

https://github.com/SFML/SFML/pull/1904#discussion_r769082066
https://github.com/SFML/SFML/pull/1910#issuecomment-997181070

vittorioromeo: > `constexpr` implies `inline`, so it should work without that keyword.

Without inline, VS2019 fails to compile as well, complaining about multiple definitions.

Bromeon: > Without inline, VS2019 fails to compile as well, complaining about multiple definitions. My first commit did not use inline :)

I see, thanks. Maybe we should add a comment, otherwise someone in the future might wonder or even attempt to remove `inline`:
```cpp
// Note: inline keyword needed because ...
```

vittorioromeo:
@Bromeon @binary1248 @eXpl0it3r: how do you feel about dropping VS2017 support? It is clearly broken...

vittorioromeo:
> @Bromeon @binary1248 @eXpl0it3r: how do you feel about dropping VS2017 support? It is clearly broken...

Ping :)

Bromeon:
Given that 2017 lies now half a decade in the past, I think it's OK to drop support for VS 2017, especially if it struggles to be C++17 compatible.

It's of course a question if this can be worked around somehow (with reasonable effort).

eXpl0it3r:
Well we knew going in, that we'll have to abandon some compilers. So that's fine for me. Wondering if we can easily do some branch filtering in Build Bot or drop VS 2017 for 2.6.x as well 🤔

eXpl0it3r:
The windows-2016 image is deprecated and will be removed in March 2022 anyways, so GitHub Actions couldn't easily cover VS 2017, see https://github.blog/changelog/2021-10-19-github-actions-the-windows-2016-runner-image-will-be-removed-from-github-hosted-runners-on-march-15-2022/

vittorioromeo:
@eXpl0it3r: it seems CI still tries to compile against VS2017. Is there a PR for the removal of this compiler?

eXpl0it3r:
No, I didn't do anything. If we all agree, I can create one.

ChrisThrasher:
Is this issue unblocked now that VS 2017 has been removed?

vittorioromeo:
CI passes now. @Bromeon @eXpl0it3r happy to approve?

eXpl0it3r (approved):

PR #1939, opened 2021-12-24
Keep test utilities alongside the files that use them

ChrisThrasher:
## Description

A given .hpp/.cpp pair of test utilities is only needed by one module's tests. It's more readable and less error prone to keep those files alongside the .cpp files that need them than to keep everything in one directory where unexpected dependencies can be created that make tests harder to reason about and manually verify their correctness.

Ultimately we should just delete these files since I know Catch2 didn't use these `toString` overloads and doctest may not either. I recommend the pattern I used in test/System/Time.cpp where I added an `operator<<` overload at the top of the file. This means a given public header can be tested with a single test .cpp file and nothing more. Let's keep things simple.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android


eXpl0it3r (approved):
I'm okay with this, I guess the only "confusing" part that could remain, is that not *all* files in the specific module directory contains actual tests.

What's the opinions of others'

eXpl0it3r:

Bromeon:
There are two ways to see this -- none is significantly more readable or less error-prone per se:
* All files in the `Graphics`, `System` folders etc. are actual test cases, so everything that's not a test case is in a separate directory
* Everything related to a module is in its corresponding folder

The 2nd approach (this PR) breaks down however as soon as you have utilities that are needed by multiple modules (unless you want to include `System/SystemUtil.hpp` from a `Graphics` test). I'm not even sure if this categorization of utils works in the first place.

Other than that, I don't mind much either way.

ChrisThrasher:
We can have both benefits though. We just don't need these utilities files. The relevant bits from each file can live in a single .cpp file. If you're testing Widget.hpp, there will be a test file called Widget.cpp that contains the tests as well as an `operator<<` overload for `Widget` that gives you pretty printing. No accessory files needed. Every .cpp file in test/ will contain tests.

ChrisThrasher:
I went ahead and removed the utilities files so you can see what it would like to have no test utilities files. As far as I'm aware, removing these files in this manner does not remove functionality whatsoever. Any desired printing functionality is implemented within the test file itself.

Turns out the printing function for `sf::Rect` was in the WindowUtils.hpp file which doesn't make sense. Those are the kinds of unexpected things that I'm trying to prevent by not having these utilities files.

vittorioromeo:
SFMLBuildMaster: Build this pleaes.

vittorioromeo:

ChrisThrasher:
Closing this PR because as I write more tests I'm realizing how the current layout is better than what I'm recommending. In particular the tests for other modules will need the printers from SystemUtils because so many APIs use stuff like `sf::Vector2`. I still think there's some cleanup to be done to minimize the complexity of these utilities files but that's for a different PR.

PR #1941, opened 2021-12-24
Prevent crash in Unix `JoystickImpl` with too many file descriptors

### LINKED ISSUE #1900: Linux: Crash when creating window if more than 1024 file descriptors are open
## Subject of the issue

My (linux) application using SFML is opening many files (not using SFML at all) then create a SFML Window. Everything works fine if I open less than FD_SETSIZE (1024)  files.
If I open more than 1024 files (I'm using ulimit -n 4096) the application crashes in SFML when creating the window.

## Your environment

* Linux (Ubuntu 20.04)
* SFML 2.5.1 (ubuntu package or compiled from 2.5.1 tag)
* gcc 9.3
* ulimit -n 4096

## Steps to reproduce

Using SFML Release build, any app opening files then creating a window:

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    /* do something to open more than 1024 files so file descriptor values are > 1024 */
....

    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
        }

        window.clear();
        window.display();
    }
}
```

## Expected behavior

No crash and SFML window created

## Actual behavior

Crash with "*** buffer overflow detected ***: terminated" message.

Crash comes from hasMonitorEvent() in src/SFML/Window/Unix/JoystickImpl.cpp 

```cpp
     int monitorFd = udev_monitor_get_fd(udevMonitor);
     fd_set descriptorSet;
     FD_ZERO(&descriptorSet);
     FD_SET(monitorFd, &descriptorSet);
     timeval timeout = {0, 0};
      return (select(monitorFd + 1, &descriptorSet, NULL, NULL, &timeout) > 0) &&
               FD_ISSET(monitorFd, &descriptorSet);
```
In my case monitorFd is > FD_SETSIZE and FD_SET crashes (FD_SET / select crash when the file descriptor is > FD_SET_SIZE  is a well know linux issue)

For some reason I can't explain I don't reproduce the crash with a debug SFML build (i.e. CMAKE_BUILD_TYPE=Debug) even if the monitorFd has the exact same value.

vittorioromeo:
## Description

Fixes #1900 by using the check @fvallee-bnx suggested. Note that `hasMonitorEvent` is used in `isConnected` here:

```cpp
bool JoystickImpl::isConnected(unsigned int index)
{
    // See if we can skip scanning if udev monitor is available
    if (!udevMonitor)
    {
        // udev monitor is not available, perform a scan every query
        updatePluggedList();
    }
    else if (hasMonitorEvent())
    {
        // Check if new joysticks were added/removed since last update
        udev_device* udevDevice = udev_monitor_receive_device(udevMonitor);

        // If we can get the specific device, we check that,
        // otherwise just do a full scan if udevDevice == nullptr
        updatePluggedList(udevDevice);

        if (udevDevice)
            udev_device_unref(udevDevice);
    }

    if (index >= joystickList.size())
        return false;

    // Then check if the joystick is connected
    return joystickList[index].plugged;
}
```

`joystickList[index].plugged` might return the wrong value in that case, but it seems fine as it is an edge case (over 1024 file descriptors). Not sure if there's a nice way to avoid a hard limit on file descriptors, but this should at least solve the crash.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

It needs to be tested on Linux, I have not had a chance to do that yet.

### DIFF
--- a/src/SFML/Window/Unix/JoystickImpl.cpp
+++ b/src/SFML/Window/Unix/JoystickImpl.cpp
@@ -29,6 +29,7 @@
 #include <SFML/System/Err.hpp>
 #include <linux/joystick.h>
 #include <libudev.h>
+#include <poll.h>
 #include <unistd.h>
 #include <fcntl.h>
 #include <errno.h>
@@ -253,13 +254,9 @@ namespace
         // This will not fail since we make sure udevMonitor is valid
         int monitorFd = udev_monitor_get_fd(udevMonitor);
 
-        fd_set descriptorSet;
-        FD_ZERO(&descriptorSet);
-        FD_SET(monitorFd, &descriptorSet);
-        timeval timeout = {0, 0};
+        pollfd fds{ monitorFd, POLLIN, 0 };
 
-        return (select(monitorFd + 1, &descriptorSet, NULL, NULL, &timeout) > 0) &&
-               FD_ISSET(monitorFd, &descriptorSet);
+        return (poll(&fds, 1, 0) > 0) && ((fds.revents & POLLIN) != 0);
     }
 
     // Get a property value from a udev device


binary1248:
> Not sure if there's a nice way to avoid a hard limit on file descriptors, but this should at least solve the crash.

I was thinking of just replacing `select` by `poll`. The limits of `select` have been known for a long time, which is why `poll` was designed to achieve the same functionality without the limitations.

I don't think we should jump to conclusions a bit too fast. We don't know what the user might be doing with their system. I can imagine for scientific/prosumer purposes that need some kind of visualisation SFML could have been used if it wasn't for this limitation.

https://man7.org/linux/man-pages/man2/select.2.html

> WARNING: select() can monitor only file descriptors numbers that are less than FD_SETSIZE (1024)—an unreasonably low limit for many modern applications—and this limitation will not change. All modern applications should instead use poll(2) or epoll(7), which do not suffer this limitation.

binary1248:
Something along the lines of:

```cpp
#include <poll.h>

bool hasMonitorEvent()
{
    // This will not fail since we make sure udevMonitor is valid
    int monitorFd = udev_monitor_get_fd(udevMonitor);

    pollfd fds{ monitorFd, POLLIN, 0 };

    return (poll(&fds, 1, 0) > 0) && ((fds.revents & POLLIN) != 0);
}
```

binary1248:
@vittorioromeo Would it be possible to fix this problem with the `poll` snippet I provided?

binary1248:
@vittorioromeo I updated the commit to the `poll` snippet I posted above.

eXpl0it3r:
@fvallee-bnx can you give this branch a try? Note that you may need to adjust more, since the master branch is undergoing major changes towards SFML 3

binary1248:
This patch can also be backported to 2.6.x without issues.

fvallee-bnx:
@eXpl0it3r looks good to me (patch backported to 2.5.3 - no crash). Thanks a lot!

eXpl0it3r:
I've backported the change to SFML 2.6 then we can forward merge it onto master later on

eXpl0it3r (approved):

PR #1943, opened 2021-12-27
Let tests and examples inherit language requirement from upstream targets

ChrisThrasher:

----

## Description

It's not necessary to re-specify cxx_std_17 since any example or test which depends on a core target (which should be all of them) will pick up this language require that should be a public property of those targets. If that changes, these examples and tests will potentially fail to compile and correctly catch the bug that was introduced in the core library targets.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -268,9 +268,6 @@ macro(sfml_add_example target)
         add_executable(${target} ${target_input})
     endif()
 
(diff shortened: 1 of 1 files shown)


vittorioromeo:

PR #1944, opened 2021-12-29
Remove redundant doctest formatters

ChrisThrasher:
## Description

I added `operator<<` to types that already had Doctest formatters so I removed them. 

Something to keep in mind is how GraphicsUtils.hpp depends on WindowUtils.hpp which depends on SystemUtils.hpp. Is this desired? If these util headers are so tightly coupled, why not just have one big header where we put all formatters? That would be easier to maintain.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/test/System/Time.cpp
+++ b/test/System/Time.cpp
@@ -1,18 +1,7 @@
 #include <SFML/System/Time.hpp>
-
-#include <ostream>
-#include <doctest.h>
+#include "SystemUtil.hpp"
 
 using doctest::Approx;
-
-namespace sf
-{
-std::ostream& operator <<(std::ostream& os, const sf::Time& time)
-{
-    os << time.asMicroseconds() << "us";
-    return os;
-}
-}
 
 TEST_CASE("sf::Time class - [system]")
 {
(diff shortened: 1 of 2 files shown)


ChrisThrasher:
@vittorioromeo Do you have a preference for using `doctest::string toString` over `operator<<`? The latter is more canonical in my opinion and doesn't require we depend on `doctest.h` just to define a printer function for a given type. It also means zero extra effort if we change testing frameworks going forward. This is even the [first thing](https://github.com/doctest/doctest/blob/master/doc/markdown/stringification.md) the doctest docs recommend doing.

Here's what that could look like:

https://github.com/ChrisThrasher/SFML/commit/9a06adf8fd80e3a7f94fc994a27b6c7c00d61fe3

vittorioromeo:

vittorioromeo (approved):
LGTM. I'm fine with using `operator<<`, I agree it is more idiomatic.

ChrisThrasher:
> I'm fine with using `operator<<`, I agree it is more idiomatic.

I'll submit a PR for this later.

Issue #1562, opened 2019-02-24
Add aliases for CMake targets


eliasdaler:
This was previously discussed at Discord/IRC, but I want to discuss it here too.
Right now, SFML only offers targets like `sfml-graphics`, `sfml-system` and so on. What I want is to have something like `SFML::Graphics`, `SFML::System` and so on. 
The advantage of these aliases is that if the target `sfml-graphics` is not found at configuration time, CMake simply will add `-lsml-graphics` and be done with it, If I misspell `sfml-graphics`, for example, by writing `sfml-garphics`, CMake will add `-lsfml-garphics` and I'll have an error at **link** time.
On the other hand, if I link to `SFML::Graphics`, I'll get error at **configuration** time if this target is not found.

The thing I want to discuss here is how these aliases should be named.

1) SFML::sfml-graphics
2) sfml::graphics
3) SFML::Graphics
4) sfml::sfml-graphics

In my opinion the second one looks the best, because it looks like "sfml-graphics". Plus, making "Graphics" from "graphics" string in CMake is not a trivial task... So, what do you think?

Bromeon:
The common convention seems to be `PascalCase`, without repetition.

Examples:
* `JPEG::JPEG` ([libjpeg](https://github.com/Kitware/CMake/blob/master/Modules/FindJPEG.cmake))
* `PNG::PNG` ([libpng](https://cmake.org/cmake/help/v3.6/module/FindPNG.html))
* `OpenSSL::Crypto` ([OpenSSL](https://cmake.org/cmake/help/v3.6/module/FindOpenSSL.html))
* `Qt5::Core` ([Qt](https://doc.qt.io/qt-5/cmake-manual.html))
* `JsonCpp::JsonCppStatic` ([jsoncpp](https://github.com/rpavlik/cmake-modules/blob/master/FindJsonCpp.cmake))


Some deviate also in the component part:
* `Boost::filesystem` ([Boost](https://cmake.org/cmake/help/v3.8/module/FindBoost.html))

**I would thus suggest `SFML::Graphics`**.

See also: https://stackoverflow.com/q/48524359

LaurentGomila:
If we want to be consistent with the call to find_package, it would be

```
find_package(SFML COMPONENTS graphics)
target_link_libraries(target PRIVATE SFML::graphics)
```

Bromeon:
[...]

PR #1947, opened 2021-12-30
Add SFML:: namespace to targets

ChrisThrasher:
## Description

This removes the sfml- prefixed targets from the export set. The sfml- prefixed targets are still available within the build tree but not to downstream users thus making this an API breaking change when compared to the 2.x releases. To keep things consistent, usage of the sfml- targets were replaced with their namespaced counterparts.

This has a number of benefits:

  1. It's more idiomatic. Modern CMake libraries are expected to
     have namespaced targets.

  2. Namespaces targets are less likely to collide with user-defined
     targets. No one will accidentally define a SFML:: target.

This PR is related to the issue #1562

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Build and install SFML. Find and use the package in the following way:

```cmake
find_package(SFML 3.0.0 REQUIRED COMPONENTS graphics)
...
target_link_libraries(some-target PRIVATE SFML::graphics)
```

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -50,10 +50,10 @@ function(sfml_set_common_ios_properties target)
 endfunction()
 
 # add a new target which is a SFML library
-# example: sfml_add_library(sfml-graphics
+# example: sfml_add_library(Graphics
 #                           SOURCES sprite.cpp image.cpp ...
 #                           [STATIC]) # Always create a static library and ignore BUILD_SHARED_LIBS
-macro(sfml_add_library target)
+macro(sfml_add_library module)
 
     # parse the arguments
     cmake_parse_arguments(THIS "STATIC" "" "SOURCES" ${ARGN})
@@ -62,11 +62,13 @@ macro(sfml_add_library target)
     endif()
 
     # create the target
+    string(TOLOWER sfml-${module} target)
     if (THIS_STATIC)
         add_library(${target} STATIC ${THIS_SOURCES})
     else()
         add_library(${target} ${THIS_SOURCES})
     endif()
+    add_library(SFML::${module} ALIAS ${target})
 
     # enable C++17 support
     target_compile_features(${target} PUBLIC cxx_std_17)
@@ -77,6 +79,9 @@ macro(sfml_add_library target)
     string(REPLACE "-" "_" NAME_UPPER "${target}")
     string(TOUPPER "${NAME_UPPER}" NAME_UPPER)
     set_target_properties(${target} PROPERTIES DEFINE_SYMBOL ${NAME_UPPER}_EXPORTS)
+
+    # define the export name of the module
+    set_target_properties(${target} PROPERTIES EXPORT_NAME ${module})
 
     # adjust the output file prefix/suffix to match our conventions
     if(BUILD_SHARED_LIBS AND NOT THIS_STATIC)
@@ -232,7 +237,7 @@ endmacro()
 # example: sfml_add_example(ftp
 #                           SOURCES ftp.cpp ...
 #                           BUNDLE_RESOURCES MainMenu.nib ...    # Files to be added in target but not installed next to the executable
-#                           DEPENDS sfml-network
+#                           DEPENDS SFML::Network
 #                           RESOURCES_DIR resources)             # A directory to install next to the executable and sources
 macro(sfml_add_example target)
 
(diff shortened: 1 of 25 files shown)


Bromeon:
Should we not use this chance and make module names consistent everywhere, i.e. `PascalCase`?

We have `#include <SFML/Graphics.hpp>`, so `SFML::Graphics` would be standing to reason in my opinion.

ChrisThrasher:
That's fine by me. I'll make the change as soon as the team comes to a consensus.

ChrisThrasher:
If we move to a `SFML::Graphics` style do we still want to keep the internal target names like `sfml-graphics`? That can be done but will require some find-replace string operation in `sfml_add_library` to automatically capitalize the first letter. `sfml-Graphics` is a bit unsightly so figuring out that string manipulation is probably worth it.

Bromeon:
If the user-facing names are consistent (i.e. `Graphics`), that's already a good start.

What came to my mind now, are the output libraries/binaries, which are probably still `sfml-graphics.{dll|so}`. I guess that could stay; other namings like `SFML-Graphics` or `SFML_Graphics` are rather weird for those.

eXpl0it3r:

ChrisThrasher:
I think I figured out a good way to keep the existing `sfml-graphics` style targets while adding `SFML::Graphics` style targets for export and use by downstream targets. I feel good about this implementation.

ChrisThrasher:
I updated the component names so instead of writing
```cmake
find_package(SFML 3.0.0 REQUIRED COMPONENTS graphics audio)
```
you write
```cmake
find_package(SFML 3.0.0 REQUIRED COMPONENTS Graphics Audio)
```

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:
As a German-speaking person I've noticed a major bias towards capitalizing things, so things like that make it hard for me to judge what's better. 😄 



> I updated the component names so instead of writing
> 
> ```cmake
> find_package(SFML 3.0.0 REQUIRED COMPONENTS graphics audio)
> ```
> 
> you write
> 
> ```cmake
> find_package(SFML 3.0.0 REQUIRED COMPONENTS Graphics Audio)
> ```

Does that mean the lower-case variant will no longer be working?

ChrisThrasher:
> Does that mean the lower-case variant will no longer be working?

That’s right. I believe it’s theoretically possible to support both upper and lower case component names but that would be difficult to implement, especially if you wanted `graphics` and `Graphics` to be mutually exclusive. Currently it looks like this when users want to depend on SFML
```cmake
find_package(SFML 3.0.0 REQUIRED COMPONENTS Graphics)
…
target_link_libraries(app PRIVATE SFML::Graphics)
```
```cpp
#include <SFML/Graphics.hpp>
```
While I don’t feel strong about what is or isn’t capitalized, it does look nice that everything follows this “SFML” + “Graphics” pattern in all places.

cmake/Macros.cmake
```diff
@@ -50,10 +50,10 @@ function(sfml_set_common_ios_properties target)
 endfunction()
 
 # add a new target which is a SFML library
-# example: sfml_add_library(sfml-graphics
+# example: sfml_add_library(graphics
```
eXpl0it3r, line 53: I wonder, if we're using capitalized module names everywhere now, shouldn't this also work as such and reverse the functionality internally to lowercase the module name?

ChrisThrasher: Do you mean change the convention to pass in an upper case module name to `sfml_add_library` like `sfml_add_library(Graphics)` then convert that to lower case before creating the internally used `sfml-graphics` style CMake target names? That would certainly make for a simpler implementation of `sfml_add_library` since converting strings to lowercase is a lot more simple than uppercasing the first letter.

eXpl0it3r: Yep, that's what I meant

ChrisThrasher: I like that idea. I'll try implementing it to make sure it's as simple as I'm expecting it to be.

ChrisThrasher: All done!

eXpl0it3r:

eXpl0it3r:

eXpl0it3r (approved):

PR #1942, opened 2021-12-26
Remove redundant API taking x,y pair instead of sf::Vector

ChrisThrasher:
Thanks a lot for making a contribution to SFML! 🙂

Before you create the pull request, we ask you to check the follow boxes. (For small changes not everything needs to ticked, but the more the better!)

* [ ] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before?
* [x] Does the code follow the SFML [Code Style Guide](https://www.sfml-dev.org/style.php)?
* [ ] Have you provided some example/test code for your changes?
* [ ] If you have additional steps which need to be performed list them as tasks!

----

## Description

Expanding on @vittorioromeo's work in https://github.com/SFML/SFML/pull/1902. I removed a few more instances of (x, y) pairs in APIs alongside `sf::Vector` variants. I grep'd around the repo for as many instances of this pattern as possible and couldn't find any more other than some `sf::Image` APIs that have no `sf::Vector` counterpart (but maybe should).

The biggest changes were in `sf::Transform` which have no tests but I have a branch where I'm starting to test it. Happy to submit a PR for the tests I've added so far but there's plenty more work to go there. 

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

PR #1948, opened 2021-12-31
Remove redundant 4-param sf::Rect<T> constructor

ChrisThrasher:

----

## Description

While writing tests for `sf::Transform` I discovered that `sf::Rect` has a constructor redundant with another overload that uses existing vocabulary types.

This PR is related to #1902 and #1942.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/src/SFML/Graphics/RenderTarget.cpp
+++ b/src/SFML/Graphics/RenderTarget.cpp
@@ -202,10 +202,10 @@ IntRect RenderTarget::getViewport(const View& view) const
     float height = static_cast<float>(getSize().y);
     const FloatRect& viewport = view.getViewport();
 
-    return IntRect(static_cast<int>(0.5f + width  * viewport.left),
-                   static_cast<int>(0.5f + height * viewport.top),
-                   static_cast<int>(0.5f + width  * viewport.width),
-                   static_cast<int>(0.5f + height * viewport.height));
+    return IntRect({static_cast<int>(0.5f + width  * viewport.left),
+                    static_cast<int>(0.5f + height * viewport.top)},
+                   {static_cast<int>(0.5f + width  * viewport.width),
+                    static_cast<int>(0.5f + height * viewport.height)});
 }
 
 
@@ -555,7 +555,7 @@ void RenderTarget::resetGLStates()
 void RenderTarget::initialize()
 {
     // Setup the default and current views
-    m_defaultView.reset(FloatRect(0, 0, static_cast<float>(getSize().x), static_cast<float>(getSize().y)));
+    m_defaultView.reset(FloatRect({0, 0}, Vector2f(getSize())));
     m_view = m_defaultView;
 
     // Set GL states only on first draw, so that we don't pollute user's states
--- a/src/SFML/Graphics/Shape.cpp
+++ b/src/SFML/Graphics/Shape.cpp
@@ -66,7 +66,7 @@ void Shape::setTexture(const Texture* texture, bool resetRect)
     {
         // Recompute the texture area if requested, or if there was no texture & rect before
         if (resetRect || (!m_texture && (m_textureRect == IntRect())))
-            setTextureRect(IntRect(0, 0, static_cast<int>(texture->getSize().x), static_cast<int>(texture->getSize().y)));
+            setTextureRect(IntRect({0, 0}, Vector2i(texture->getSize())));
     }
 
     // Assign the new texture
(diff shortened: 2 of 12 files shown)


src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -555,7 +555,7 @@ void RenderTarget::resetGLStates()
 void RenderTarget::initialize()
 {
     // Setup the default and current views
-    m_defaultView.reset(FloatRect(0, 0, static_cast<float>(getSize().x), static_cast<float>(getSize().y)));
+    m_defaultView.reset(FloatRect({0, 0}, {static_cast<float>(getSize().x), static_cast<float>(getSize().y)}));
```
kimci86, line 558:
```suggestion
    m_defaultView.reset(FloatRect({0, 0}, Vector2f(getSize())));
```

src/SFML/Graphics/Shape.cpp
```diff
@@ -66,7 +66,7 @@ void Shape::setTexture(const Texture* texture, bool resetRect)
     {
         // Recompute the texture area if requested, or if there was no texture & rect before
         if (resetRect || (!m_texture && (m_textureRect == IntRect())))
-            setTextureRect(IntRect(0, 0, static_cast<int>(texture->getSize().x), static_cast<int>(texture->getSize().y)));
+            setTextureRect(IntRect({0, 0}, {static_cast<int>(texture->getSize().x), static_cast<int>(texture->getSize().y)}));
```
kimci86, line 69:
```suggestion
            setTextureRect(IntRect({0, 0}, Vector2i(texture->getSize())));
```

src/SFML/Graphics/Sprite.cpp
```diff
@@ -69,7 +69,7 @@ void Sprite::setTexture(const Texture& texture, bool resetRect)
     if (resetRect || (!m_texture && (m_textureRect == sf::IntRect())))
     {
         Vector2i size = Vector2i(texture.getSize());
-        setTextureRect(IntRect(0, 0, size.x, size.y));
+        setTextureRect(IntRect({0, 0}, {size.x, size.y}));
```
kimci86, line 72:
```suggestion
        setTextureRect(IntRect({0, 0}, Vector2i(texture.getSize())));
```
and removing the line before, now useless.

kimci86 (commented):
There are some opportunities to use `Vector2<T>` conversion constructor instead of `static_cast`. See the suggested changes.

ChrisThrasher:
Great suggestions, @kimci86!

Bromeon:
While I think this is a good change, just like the `(float, float) -> Vector2f` one, we should be aware that we're breaking a _massive_ amount of existing code for quite little benefit. In fact, the "benefit" is a rather arbitrary syntax preference by the SFML developers, and the justification _seems_ mostly "this is how it's done now". (I do see the consistency and smaller API surface, which are good things -- but it's still an opinionated change to some extent).

These are changes that will cause a lot of work when migrating SFML 2 -> 3 codebase, and in extreme cases may even be reason to stay with SFML 2. I'm not saying we shouldn't commit the changes, but I have the feeling we're currently breaking a lot of stuff without deeply thinking about the implications. If we go forward with them, do we plan to assist the users in one way or another? By that I mean not just a changelog -- rather a migration guide, maybe some regexes for search&replace, or even some automated best-effort tooling.

ChrisThrasher:
You're right and I don't have a great response. I think these changes are ultimately worthwhile but I hadn't fully realized how much harder these changes will make transitioning large codebases.

ChrisThrasher:
@vittorioromeo Do you have any good ideas for how we can ease the SFML 2->3 transition for heavy users of the APIs we have removed?

vittorioromeo:

vittorioromeo:
> @vittorioromeo Do you have any good ideas for how we can ease the SFML 2->3 transition for heavy users of the APIs we have removed?

The nicest thing would be providing a custom `clang-tidy` fixit that can be automatically ran over an entire codebase and execute the fixes. See https://www.youtube.com/watch?v=38tYYrnfNrs, https://www.youtube.com/watch?v=_T-5pWQVxeE, and search online for similar things. It's not super-complicated to create a new `clang-tidy` check, but it requires a bit of time and effort, and it also requires the users to do some work to set it up. But once it works, it can automagically update an entire codebase from SFML 2 to SFML 3.

Alternatively, I think that a "migration guide" or "checklist" with some helpful steps or regexes would be the bare minimum we should provide. Some sort of step-by-step guide that anyone with an SFML codebase can go through to ensure that most if not all breaking changes are addressed.

vittorioromeo (approved):
LGTM

PR #1955, opened 2022-01-02
Add tests for sf::Transform

ChrisThrasher:

----

## Description

Added tests for the entire API of sf::Transform. The goal is to slowly work up in the Graphics dependency graph. Because `sf::Vector` and `sf::Rect` already had tests, this was the next thing to test before eventually moving to the higher level Graphics units like the various pre-made shapes.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/test/Graphics/Transform.cpp
+++ b/test/Graphics/Transform.cpp
@@ -1,0 +1,264 @@
+#include <SFML/Graphics/Transform.hpp>
+#include "GraphicsUtil.hpp"
+#include "SystemUtil.hpp"
+#include <vector>
+
+#include <doctest.h>
+
+using doctest::Approx;
+
+namespace std
+{
+std::ostream& operator<<(std::ostream& out, const std::vector<float>& vector)
+{
+    out << "{ ";
+    for (size_t i = 0; i + 1 < vector.size(); ++i)
+        out << vector[i] << ", ";
+    out << vector.back() << " }";
+    return out;
+}
+}
+
+TEST_CASE("sf::Transform class - [graphics]")
+{
+    SUBCASE("Construction")
+    {
+        SUBCASE("Default constructor")
+        {
+            CHECK(sf::Transform() == sf::Transform::Identity);
+        }
+
+        SUBCASE("3x3 matrix constructor")
+        {
+            const sf::Transform transform(10.0f, 11.0f, 12.0f,
+                                          13.0f, 14.0f, 15.0f,
+                                          16.0f, 17.0f, 18.0f);
+            const std::vector<float> matrix(transform.getMatrix(), transform.getMatrix() + 16);
+            CHECK(matrix == std::vector<float>{10.0f, 13.0f, 0.0f, 16.0f,
+                                               11.0f, 14.0f, 0.0f, 17.0f,
+                                                0.0f,  0.0f, 1.0f,  0.0f,
+                                               12.0f, 15.0f, 0.0f, 18.0f});
+        }
+    }
+
+    SUBCASE("Identity matrix")
+    {
+        const std::vector<float> matrix(sf::Transform::Identity.getMatrix(), sf::Transform::Identity.getMatrix() + 16);
+        CHECK(matrix == std::vector<float>{1.0f, 0.0f, 0.0f, 0.0f,
+                                           0.0f, 1.0f, 0.0f, 0.0f,
+                                           0.0f, 0.0f, 1.0f, 0.0f,
+                                           0.0f, 0.0f, 0.0f, 1.0f});
+    }
+
+    SUBCASE("getInverse()")
+    {
+        CHECK(sf::Transform::Identity.getInverse() == sf::Transform::Identity);
+        CHECK(sf::Transform(1.0f, 2.0f, 3.0f, 4.0f, 5.0f, 6.0f, 7.0f, 8.0f, 9.0f).getInverse() == sf::Transform::Identity);
+        CHECK(sf::Transform(1.0f, 2.0f, 3.0f, 4.0f, 5.0f, 4.0f, 3.0f, 2.0f, 1.0f).getInverse() == sf::Transform(0.375f, -0.5f, 0.875f, -1.0f, 1.0f, -1.0f, 0.875f, -0.5f, 0.375f));
+    }
+
+    SUBCASE("transformPoint()")
+    {
+        CHECK(sf::Transform::Identity.transformPoint({-10.0f, -10.0f}) == sf::Vector2f(-10.0f, -10.0f));
+        CHECK(sf::Transform::Identity.transformPoint({-1.0f, -1.0f}) == sf::Vector2f(-1.0f, -1.0f));
+        CHECK(sf::Transform::Identity.transformPoint({-1.0f, 0.0f}) == sf::Vector2f(-1.0f, 0.0f));
+        CHECK(sf::Transform::Identity.transformPoint({0.0f, 0.0f}) == sf::Vector2f(0.0f, 0.0f));
+        CHECK(sf::Transform::Identity.transformPoint({0.0f, 1.0f}) == sf::Vector2f(0.0f, 1.0f));
+        CHECK(sf::Transform::Identity.transformPoint({1.0f, 1.0f}) == sf::Vector2f(1.0f, 1.0f));
+        CHECK(sf::Transform::Identity.transformPoint({10.0f, 10.0f}) == sf::Vector2f(10.0f, 10.0f));
+
+        const sf::Transform transform(1.0f, 2.0f, 3.0f,
+                                      4.0f, 5.0f, 4.0f,
+                                      3.0f, 2.0f, 1.0f);
+        CHECK(transform.transformPoint({-1.0f, -1.0f}) == sf::Vector2f(0.0f, -5.0f));
+        CHECK(transform.transformPoint({0.0f, 0.0f}) == sf::Vector2f(3.0f, 4.0f));
+        CHECK(transform.transformPoint({1.0f, 1.0f}) == sf::Vector2f(6.0f, 13.0f));
+    }
+
+    SUBCASE("transformRect()")
+    {
+        CHECK(sf::Transform::Identity.transformRect({{-200.0f, -200.0f}, {-100.0f, -100.0f}}) == sf::FloatRect({-300.0f, -300.0f}, {100.0f, 100.0f}));
+        CHECK(sf::Transform::Identity.transformRect({{0.0f, 0.0f}, {0.0f, 0.0f}}) == sf::FloatRect({0.0f, 0.0f}, {0.0f, 0.0f}));
+        CHECK(sf::Transform::Identity.transformRect({{100.0f, 100.0f}, {200.0f, 200.0f}}) == sf::FloatRect({100.0f, 100.0f}, {200.0f, 200.0f}));
+
+        const sf::Transform transform(1.0f, 2.0f, 3.0f,
+                                      4.0f, 5.0f, 4.0f,
+                                      3.0f, 2.0f, 1.0f);
+        CHECK(transform.transformRect({{-100.0f, -100.0f}, {200.0f, 200.0f}}) == sf::FloatRect({-297.0f, -896.0f}, {600.0f, 1800.0f}));
+        CHECK(transform.transformRect({{0.0f, 0.0f}, {0.0f, 0.0f}}) == sf::FloatRect({3.0f, 4.0f}, {0.0f, 0.0f}));
+        CHECK(transform.transformRect({{100.0f, 100.0f}, {200.0f, 200.0f}}) == sf::FloatRect({303.0f, 904.0f}, {600.0f, 1800.0f}));
+    }
+
+    SUBCASE("combine()")
+    {
+        auto identity = sf::Transform::Identity;
+        CHECK(identity.combine(sf::Transform::Identity) == sf::Transform::Identity);
+        CHECK(identity.combine(sf::Transform::Identity).combine(sf::Transform::Identity) == sf::Transform::Identity);
+
+        sf::Transform transform(1.0f, 2.0f, 3.0f,
+                                4.0f, 5.0f, 4.0f,
+                                3.0f, 2.0f, 1.0f);
+        CHECK(identity.combine(transform) == transform);
+        CHECK(transform.combine(sf::Transform::Identity) == transform);
+        CHECK(transform.combine(transform) == sf::Transform(18.0f, 18.0f, 14.0f,
+                                                            36.0f, 41.0f, 36.0f,
+                                                            14.0f, 18.0f, 18.0f));
+        CHECK(transform.combine(sf::Transform(10.0f,  2.0f,  3.0f,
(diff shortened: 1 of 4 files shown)


test/TestUtilities/GraphicsUtil.cpp
```diff
@@ -16,4 +17,14 @@ namespace sf
 
         return stream.str().c_str();
     }
+
+    doctest::String toString(const sf::Transform& transform)
+    {
+        std::ostringstream stream;
+        stream << transform.getMatrix()[0] << ", " << transform.getMatrix()[4] << ", " << transform.getMatrix()[12] << ", ";
+        stream << transform.getMatrix()[1] << ", " << transform.getMatrix()[5] << ", " << transform.getMatrix()[13] << ", ";
+        stream << transform.getMatrix()[3] << ", " << transform.getMatrix()[7] << ", " << transform.getMatrix()[15];
```
Bromeon, line 26: Maybe store the matrix in a variable for readability?

test/Graphics/Transform.cpp
```diff
@@ -0,0 +1,353 @@
...
+            CHECK(transform.getMatrix()[6] == 0.0f);
+            CHECK(transform.getMatrix()[7] == 0.0f);
+            CHECK(transform.getMatrix()[8] == 0.0f);
+            CHECK(transform.getMatrix()[9] == 0.0f);
+            CHECK(transform.getMatrix()[10] == 1.0f);
+            CHECK(transform.getMatrix()[11] == 0.0f);
+            CHECK(transform.getMatrix()[12] == 0.0f);
+            CHECK(transform.getMatrix()[13] == 0.0f);
+            CHECK(transform.getMatrix()[14] == 0.0f);
+            CHECK(transform.getMatrix()[15] == 1.0f);
```
Bromeon, line 32: I wonder if it's not better to compare the whole matrix at once. Sure, one would need to find out which elements differ, but with a good `operator<<` implementation, this should be quickly visible.

On the other hand, comparing each element means that a test will fail at the first possible element, not giving any context about the matrix's state and requiring debugging/manual print statements.

Maybe we could have something like
```cpp
std::array<float, 4 * 4> expected = {
   a, b, c, d,
   e, f, g, h,
   i, j, k, l
};
```
(with substituted values for the letters), which would also make the row/columns more clearer than a 1D view?

test/TestUtilities/GraphicsUtil.cpp
```diff
@@ -16,4 +17,14 @@ namespace sf
...
+
+        return stream.str().c_str();
```
Bromeon, line 28: I found it a bit weird that their `String` class didn't accept conversion from `std::string`, but [it's indeed the case](https://github.com/doctest/doctest/blob/a384864346f48c7ff5608051c176e7bd505712ae/doctest/parts/doctest_fwd.h#L487-L543).

Possibly it's to reduce header dependencies.

test/Graphics/Transform.cpp
```diff
@@ -0,0 +1,353 @@
...
+            CHECK(sf::Transform::Identity == sf::Transform::Identity);
+            CHECK(sf::Transform() == sf::Transform());
+            CHECK(sf::Transform(0, 0, 0, 0, 0, 0, 0, 0, 0) == sf::Transform(0, 0, 0, 0, 0, 0, 0, 0, 0));
+            CHECK(sf::Transform(0.0001f, 0.0001f, 0.0001f, 0.0001f, 0.0001f, 0.0001f, 0.0001f, 0.0001f, 0.0001f)
+               == sf::Transform(0.0001f, 0.0001f, 0.0001f, 0.0001f, 0.0001f, 0.0001f, 0.0001f, 0.0001f, 0.0001f));
+            CHECK(sf::Transform(1000.0f, 1000.0f, 1000.0f, 1000.0f, 1000.0f, 1000.0f, 1000.0f, 1000.0f, 1000.0f)
+               == sf::Transform(1000.0f, 1000.0f, 1000.0f, 1000.0f, 1000.0f, 1000.0f, 1000.0f, 1000.0f, 1000.0f));
+        }
+
+        SUBCASE("operator*=")
```
Bromeon, line 332: You probably meant `!=`.

test/Graphics/Transform.cpp
```diff
@@ -0,0 +1,353 @@
...
+                                                         14.0f, 18.0f, 18.0f));
+            CHECK(transform * sf::Transform(10.0f,  2.0f,  3.0f,
+                                             4.0f, 50.0f, 40.0f,
+                                            30.0f, 20.0f, 10.0f)
+                == sf::Transform(108.0f, 162.0f, 113.0f,
+                                 180.0f, 338.0f, 252.0f,
+                                  68.0f, 126.0f,  99.0f));
+        }
+
+        SUBCASE("operator*+")
```
Bromeon, line 284: Typo: `*=`

test/Graphics/Transform.cpp
```diff
@@ -0,0 +1,353 @@
...
+    SUBCASE("translate()")
+    {
+        sf::Transform transform(9, 8, 7, 6, 5, 4, 3, 2, 1);
+        CHECK(transform.translate({10.0f, 20.0f}) == sf::Transform(9, 8, 257, 6, 5, 164, 3, 2, 71));
+        CHECK(transform.translate({10.0f, 20.0f}) == sf::Transform(9, 8, 507, 6, 5, 324, 3, 2, 141));
+    }
+
+    SUBCASE("rotate()")
+    {
+        SUBCASE("About origin")
```
Bromeon, line 212: Probably clearer: "_around_ origin"
(same below)

test/Graphics/Transform.cpp
```diff
@@ -0,0 +1,353 @@
...
+            CHECK(transform.getMatrix()[13] == Approx(-1));
+            CHECK(transform.getMatrix()[3] == Approx(0));
+            CHECK(transform.getMatrix()[7] == Approx(0));
+            CHECK(transform.getMatrix()[15] == Approx(1));
+        }
+    }
+
+    SUBCASE("scale()")
+    {
+        SUBCASE("About origin")
```
Bromeon, line 245: I'm not sure what a good preposition is here, but I don't think it's "about".
Maybe "from" or "based in".

test/Graphics/Transform.cpp
```diff
@@ -0,0 +1,353 @@
...
+            CHECK(transform.getMatrix()[13] == 0.0f);
+            CHECK(transform.getMatrix()[14] == 0.0f);
+            CHECK(transform.getMatrix()[15] == 1.0f);
+        }
+
+        SUBCASE("3x3 matrix constructor")
+        {
+            const sf::Transform transform(3.14f, 3.14f, 3.14f,
+                                          3.14f, 3.14f, 3.14f,
+                                          3.14f, 3.14f, 3.14f);
```
Bromeon, line 39: Wouldn't it be better to have distinct components?
This test can pass by having those values in the wrong places.

test/Graphics/Transform.cpp
```diff
@@ -0,0 +1,353 @@
...
+            CHECK(inverse.getMatrix()[6] == 0.0f);
+            CHECK(inverse.getMatrix()[7] == 0.0f);
+            CHECK(inverse.getMatrix()[8] == 0.0f);
+            CHECK(inverse.getMatrix()[9] == 0.0f);
+            CHECK(inverse.getMatrix()[10] == 1.0f);
+            CHECK(inverse.getMatrix()[11] == 0.0f);
+            CHECK(inverse.getMatrix()[12] == 0.0f);
+            CHECK(inverse.getMatrix()[13] == 0.0f);
+            CHECK(inverse.getMatrix()[14] == 0.0f);
+            CHECK(inverse.getMatrix()[15] == 1.0f);
```
Bromeon, line 99: You can just do:
```cpp
CHECK(inverse == sf::Transform::Identity);
```

After all, we already test `operator==` elsewhere, and this separates concerns.

test/Graphics/Transform.cpp
```diff
@@ -0,0 +1,353 @@
...
```
Bromeon, line 122: See above

Bromeon (commented):
Thanks a lot for your great effort! Added some comments.

Something I noticed, out of scope for this pull request -- we should probably change the semantics of operations that can fail. For example, `getInverse()` for a 0-determinant matrix shouldn't return the identity transform, that just propagates logic errors.

`std::optional<sf::Transform>` is a possible return type. And if that's too un-ergonomic for the typical case, let's at least add an assertion before the `Identity` is returned.

vittorioromeo:

test/TestUtilities/GraphicsUtil.cpp
ChrisThrasher: I also found that odd. Once #1944 is merged I'll submit another PR that replaces all these doctest formatters with `operator <<` which Vittorio agrees is a better choice.

test/TestUtilities/GraphicsUtil.cpp
ChrisThrasher: Good idea. Fixed.

test/Graphics/Transform.cpp
ChrisThrasher: >  comparing each element means that a test will fail at the first possible element

I don't believe this is true. At least with Catch2, `CHECK` will report the error and keep running unlike `REQUIRE` which aborts the test after the first failure. You're right though that comparing the whole matrix makes more sense. I was overusing `getMatrix()` when I started writing tests. I don't need to use that all over the place.

test/Graphics/Transform.cpp
ChrisThrasher: Thanks. Fixed.

test/Graphics/Transform.cpp
ChrisThrasher: You're right. I changed this so each value is unique.

test/Graphics/Transform.cpp
ChrisThrasher: Fixed thanks.

test/Graphics/Transform.cpp
ChrisThrasher: "About" was something I often said in my engineering undergraduate courses in this context. Not sure if that's just an American English thing or something even more specific to the region where I went to school but as far as I'm aware it's a common way to refer to operation or transformation that take place with respect to a given point in space. For now I changed it to "Around".

test/Graphics/Transform.cpp
ChrisThrasher: Fixed!

ChrisThrasher:
> Something I noticed, out of scope for this pull request -- we should probably change the semantics of operations that can fail. For example, getInverse() for a 0-determinant matrix shouldn't return the identity transform, that just propagates logic errors.
 
@Bromeon I think I agree. I think throwing an exception would be reasonable but your idea to use `optional` in the return channel is also reasonable. I wonder how other libraries like Eigen handle this problem.

eXpl0it3r:

Bromeon:
I think `optional` is better since it makes the potential absence of a value clearer. Also, we don't use exceptions in SFML yet -- would need a discussion about error handling.

ChrisThrasher:
Updated to fix some compilation errors caused by #1948

test/Graphics/Transform.cpp
Bromeon: You're right with `CHECK` vs. `REQUIRE`, remembered that wrong.

Printing both full matrices in case of a mismatch would still help, no?
Maybe this context makes it clearer _how_ the error happened.

eXpl0it3r:

test/Graphics/Transform.cpp
Bromeon: I see, thanks for the explanation! I never heard this term, but I'm also not a native English speaker. I'm fine with both "around" or "about" in that case.

test/Graphics/Transform.cpp
ChrisThrasher: How else would we test the constructor? 
```cpp
CHECK(sf::Transform(10.0f, 11.0f, 12.0f, 13.0f, 14.0f, 15.0f, 16.0f, 17.0f, 18.0f)
   == sf::Transform(10.0f, 11.0f, 12.0f, 13.0f, 14.0f, 15.0f, 16.0f, 17.0f, 18.0f));
```
is a tautological comparison. It doesn't really hurt to include something like that but we need something else to compare the constructor against. Using `getMatrix()` to test the constructor is the equivalent of tests like this:
```cpp
SUBCASE("Default constructor")
{
    const sf::Vertex vertex;
    CHECK(vertex.position == sf::Vector2f(0.0f, 0.0f));
    CHECK(vertex.color == sf::Color(255, 255, 255));
    CHECK(vertex.texCoords == sf::Vector2f(0.0f, 0.0f));
}
```
Also notice how there is no explicit `SUBCASE("getMatrix()")` because that method is implicitly tested by this 3x3 constructor test as well as the `sf::Transform::Identity` test.

eXpl0it3r:

test/Graphics/Transform.cpp
ChrisThrasher: > Printing both full matrices in case of a mismatch would still help, no?
 
We could try copy the `float*` return value from `getMatrix()` into a `std::array` then compare those two arrays which Doctest should automatically print to the console if a single element doesn't match. Does that sound good?

ChrisThrasher: Here's another option:
```cpp
            const sf::Transform transform(10.0f, 11.0f, 12.0f,
                                          13.0f, 14.0f, 15.0f,
                                          16.0f, 17.0f, 18.0f);
            const std::vector<float> matrix(transform.getMatrix(), transform.getMatrix() + 16);
            CHECK(matrix == std::vector<float>{10.0f, 13.0f, 0.0f, 16.0f,
                                               11.0f, 14.0f, 0.0f, 17.0f,
                                                0.0f,  0.0f, 1.0f,  0.0f,
                                               12.0f, 15.0f, 0.0f, 18.0f});
```
but Doctest won't automatically format this for us so I'd have to define another `operator<<` for this.

ChrisThrasher:
@Bromeon I took a shot at comparing entire matrices instead of doing piecewise comparisons with `.getMatrix()[I]`. Check out https://github.com/SFML/SFML/pull/1955/commits/7e2e0b45e0a9f2e8ebfdfa50662df8c40f3d68a6 to see what that entails.

Bromeon (approved):
Looks good to me!

Some of the equality checks are a bit hard to read, but that's due to the expressions being so long. I don't think we should spend time formatting nicely before https://github.com/SFML/SFML/pull/1861 lands.

However, this test might be eligible for exceptions (turning off the formatter) as auto-formatting might easily ruin the layout.

eXpl0it3r:

vittorioromeo:
@ChrisThrasher: please fix the conflicts caused by merging #1959 and I will merge :)

ChrisThrasher:
@vittorioromeo Good to go now!

eXpl0it3r:

vittorioromeo:

PR #1957, opened 2022-01-03
Define fixed-width integer type aliases in terms of standard fixed-width types

ChrisThrasher:
## Description

C++11 added standard fixed-width integral types so we can more easily define the `sf::` fixed with integer type aliases in terms of these standard types. However I think we should go one step further and remove these type aliases altogether for the same reason the `sf::` threading primitives were removed in favor of standard ones.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/include/SFML/Config.hpp
+++ b/include/SFML/Config.hpp
@@ -27,6 +27,12 @@
 
 
 ////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#include <cstdint>
+
+
+////////////////////////////////////////////////////////////
 // Define the SFML version
 ////////////////////////////////////////////////////////////
 #define SFML_VERSION_MAJOR 3
@@ -173,37 +179,21 @@
 ////////////////////////////////////////////////////////////
 namespace sf
 {
-    // All "common" platforms use the same size for char, short and int
-    // (basically there are 3 types for 3 sizes, so no other match is possible),
-    // we can use them without doing any kind of check
-
     // 8 bits integer types
-    using Int8 = signed char;
-    using Uint8 = unsigned char;
+    using Int8 = std::int8_t;
+    using Uint8 = std::uint8_t;
 
     // 16 bits integer types
-    using Int16 = short;
-    using Uint16 = unsigned short;
+    using Int16 = std::int16_t;
+    using Uint16 = std::uint16_t;
 
     // 32 bits integer types
-    using Int32 = int;
-    using Uint32 = unsigned int;
+    using Int32 = std::int32_t;
+    using Uint32 = std::uint32_t;
 
     // 64 bits integer types
-    #if defined(_MSC_VER)
-        using Int64 = signed   __int64;
-        using Uint64 = unsigned __int64;
-    #else
-        #if defined(__clang__)
-            #pragma clang diagnostic push
-            #pragma clang diagnostic ignored "-Wc++11-long-long"
-        #endif
-        using Int64 = long long;
-        using Uint64 = unsigned long long;
-        #if defined(__clang__)
-            #pragma clang diagnostic pop
-        #endif
-    #endif
+    using Int64 = std::int64_t;
+    using Uint64 = std::uint64_t;
 
 } // namespace sf
 


eXpl0it3r:

eXpl0it3r:
I still kind of wonder, whether we shouldn't get rid of our own definitions altogether?
Any opinions on this from anyone?

Bromeon:
> I still kind of wonder, whether we shouldn't get rid of our own definitions altogether? 
> Any opinions on this from anyone?

The only advantage I can think of is that they're less verbose in the codebase (`Uint64` vs `std::uint64_t`), which arguably _does_ make a small difference. For example, I noticed how much cleaner Rust code looks, where you have `u16`, `i64` etc. There's literally no reason to have long identifiers for fundamental, ubiquitous types.

We have a lot of already ugly lines like these in SFML: https://github.com/SFML/SFML/blob/3197dc80293a3eaf659849e2459134fa4c57577b/src/SFML/Network/Ftp.cpp#L564

Now, with standard types:
```cpp
datum = static_cast<std::uint8_t>(static_cast<std::uint8_t>(datum * 10) + static_cast<std::uint8_t>(str[index] - '0')); 
```

It's so bad that GitHub needs to add a horizontal scroll bar 😄 

But honestly, maybe we could provide some utility functions, considering the exorbitant number of casts in our codebase.
If we had something like:
```cpp
template <typename Src>
std::uint8_t u8(Src value)
{
    return static_cast<std::uint8_t>(value);
}
```
We could simplify the above line to:
```cpp
datum = u8(u8(datum * 10) + u8(str[index] - '0')); 
```
and we're not really losing anything. This has the potential to create a **massive** improvement in readability.

eXpl0it3r:
> maybe we could provide some utility functions, considering the exorbitant number of casts in our codebase

I think, it would probably be better to first evaluate how many of these casts are *really* needed and how many can be removed if we work with the "correct" types from the start or using appropriate literals in certain singed/unsigned cases.

Other than that, I guess it can be quite useful, but one also needs to be careful not to alienate the whole code base with custom functions and conventions, otherwise you run risk of creating your own esoteric C++ alternative, making it harder for others to read/contribute and in the long term, we might "teach" bad C++ practices, by being open source and source of inspiration for others. But that's just big picture thinking. 🙂

Bromeon:
> I think, it would probably be better to first evaluate how many of these casts are really needed and how many can be removed if we work with the "correct" types from the start or using appropriate literals in certain singed/unsigned cases.

A lot of those casts were introduced [during The Great Warning Purge of '21](https://github.com/SFML/SFML/pull/1846). From a first glance, they fall mostly in these categories:
* Conversion to standard index and offset types, like `size_t` or `vector::difference_type`
* Conversion to third-party types, like `ALsizei`, `unsigned int` -> `int` (`stbi_load_from_memory()`), ...
* Integral promotion: `Int16 + Int16 == int`, result needs to be cast back
* `float` <-> `int` conversions for calculations: font size, padding, color RGBA, ...

Maybe a handful can be avoided, but with all the extlibs interaction and computations, it seems like a majority is justified.


> Other than that, I guess it can be quite useful, but one also needs to be careful not to alienate the whole code base with custom functions and conventions, otherwise you run risk of creating your own esoteric C++ alternative, making it harder for others to read/contribute and in the long term, we might "teach" bad C++ practices, by being open source and source of inspiration for others. But that's just big picture thinking. 🙂

I agree, although I'm not sure that littering the codebase with verbose `static_cast` expressions is necessarily a better teaching example 😉 I think what is important is that all those utility functions follow the same convention, so that once you know one, you know all.

And I wouldn't just blindly search&replace everything. Something like `static_cast<ALsizei>` can stay, but maybe `static_cast<sf::Uint32>` can be simplified.

Maybe related question: we have a lot of places where we use `(unsigned) int` but actually mean `Uint32/Int32`. Even very central ones like: https://github.com/SFML/SFML/blob/3197dc80293a3eaf659849e2459134fa4c57577b/include/SFML/System/Vector2.hpp#L252-L255 Should we make those explicit?

ChrisThrasher:
https://github.com/ChrisThrasher/SFML/tree/integer-aliases

I've started replacing these aliases with their standard counterparts to get an idea for how big of a change it would be.

ChrisThrasher:
I finished removing all these aliases in https://github.com/ChrisThrasher/SFML/tree/integer-aliases.

Bromeon:
> I finished removing all these aliases.
> [ChrisThrasher@6cdb9c5](https://github.com/ChrisThrasher/SFML/commit/6cdb9c5567d17c6c989c127ee3369ae35694f648)

Hm... as expected, this _does_ make code less readable, and that for no real gain... 😕

In general, I think it's OK to have `typedef`/`using` for type names which are verbose. If it's true for `Vector2<float>` -> `Vector2f`, why not also for `std::uint8_t` -> `Uint8`?

ChrisThrasher:
> Hm... as expected, this _does_ make code less readable

Fine by me. In that case this PR is good to go now. If we're keeping these aliases we might as well define them in the simplest terms possible.

 > If it's true for Vector2<float> -> Vector2f, why not also for std::uint8_t -> Uint8?
 
Good point. I love those vector aliases.

eXpl0it3r:
> Maybe related question: we have a lot of places where we use `(unsigned) int` but actually mean `Uint32/Int32`. Even very central ones like:  https://github.com/SFML/SFML/blob/3197dc80293a3eaf659849e2459134fa4c57577b/include/SFML/System/Vector2.hpp#L252-L255 Should we make those explicit?

Do we really *mean* `Uint32` here? I'm sure there are some places that might assume `Uint32`, but in general, I don't see why the underlying type for `Vector2u` or `Vector2i` needs to fixed?

Bromeon:
You're right. In practice it probably doesn't matter, unless we keep explicitly converting between `int` and `sf::Uint32`.

I find the coexistence of fixed-size and variable-size types a bit... unfortunate. But maybe it's just after working with many languages that provide fixed-size types (Java, Kotlin, Rust...), that I realized you typically don't need variable-sized ones. But I don't know what's best from an API design perspective. Both have their advantages:
* Fixed-size ones whenever possible (exceptions are `size_t` + `ptrdiff_t`, which _can_ vary in practice).
  Pro: more uniform, fewer conversions.
* Fixed-size only when exact number of bytes are required (this is the current approach).
  Pro: APIs which truly require fixed-size types are recognizable as such (`sf::Color` and `sf::Packet`).

So it's probably fine to keep as-is.

vittorioromeo:

vittorioromeo:
I am personally in favour of this particular PR, and would be happy to discuss removing the aliases. I think it would be a good idea in a vacuum, but we also need to take into account the fact that we'd be breaking a lot of code, probably more than what we already broke by removing the `(x, y)` constructors in favour of vector ones. Not sure if that's worth it.

eXpl0it3r (approved):

Bromeon (approved):

PR #1958, opened 2022-01-05
Allow examples to run from any directory

ChrisThrasher:

----

## Description

I added a compiler definition to each example that allows for passing absolute paths to SFML functions that expect file paths. This means any example can be run from any directory instead of having to be in the example's source directory.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Build examples and attempt to run them from any directory on your system.

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -271,6 +271,7 @@ macro(sfml_add_example target)
     endif()
 
     set_file_warnings(${target_input})
+    target_compile_definitions(${target} PRIVATE RESOURCE_PATH=std::string\(\"${CMAKE_CURRENT_SOURCE_DIR}/resources/\"\))
 
     # set the debug suffix
     set_target_properties(${target} PROPERTIES DEBUG_POSTFIX -d)
--- a/examples/island/Island.cpp
+++ b/examples/island/Island.cpp
@@ -92,7 +92,7 @@ int main()
     window.setVerticalSyncEnabled(true);
 
     sf::Font font;
-    if (!font.loadFromFile("resources/tuffy.ttf"))
+    if (!font.loadFromFile(RESOURCE_PATH + "tuffy.ttf"))
         return EXIT_FAILURE;
 
     // Create all of our graphics resources
@@ -127,7 +127,7 @@ int main()
     {
         statusText.setString("Shaders and/or Vertex Buffers Unsupported");
     }
-    else if (!terrainShader.loadFromFile("resources/terrain.vert", "resources/terrain.frag"))
+    else if (!terrainShader.loadFromFile(RESOURCE_PATH + "terrain.vert", RESOURCE_PATH + "terrain.frag"))
     {
         prerequisitesSupported = false;
 
--- a/examples/joystick/Joystick.cpp
+++ b/examples/joystick/Joystick.cpp
@@ -92,7 +92,7 @@ int main()
 
     // Load the text font
     sf::Font font;
-    if (!font.loadFromFile("resources/tuffy.ttf"))
+    if (!font.loadFromFile(RESOURCE_PATH + "tuffy.ttf"))
         return EXIT_FAILURE;
 
     // Set up our string conversion parameters
--- a/examples/opengl/OpenGL.cpp
+++ b/examples/opengl/OpenGL.cpp
@@ -22,7 +22,7 @@ std::string resourcesDir()
 #ifdef SFML_SYSTEM_IOS
     return "";
 #else
-    return "resources/";
+    return RESOURCE_PATH;
 #endif
 }
 
--- a/examples/shader/Shader.cpp
+++ b/examples/shader/Shader.cpp
@@ -24,12 +24,12 @@ public:
     bool onLoad() override
     {
         // Load the texture and initialize the sprite
-        if (!m_texture.loadFromFile("resources/background.jpg"))
+        if (!m_texture.loadFromFile(RESOURCE_PATH + "background.jpg"))
             return false;
         m_sprite.setTexture(m_texture);
 
         // Load the shader
-        if (!m_shader.loadFromFile("resources/pixelate.frag", sf::Shader::Fragment))
+        if (!m_shader.loadFromFile(RESOURCE_PATH + "pixelate.frag", sf::Shader::Fragment))
             return false;
         m_shader.setUniform("texture", sf::Shader::CurrentTexture);
 
@@ -93,7 +93,7 @@ public:
         m_text.setPosition({30.f, 20.f});
 
         // Load the shader
-        if (!m_shader.loadFromFile("resources/wave.vert", "resources/blur.frag"))
+        if (!m_shader.loadFromFile(RESOURCE_PATH + "wave.vert", RESOURCE_PATH + "blur.frag"))
             return false;
 
         return true;
@@ -146,7 +146,7 @@ public:
         }
 
         // Load the shader
-        if (!m_shader.loadFromFile("resources/storm.vert", "resources/blink.frag"))
+        if (!m_shader.loadFromFile(RESOURCE_PATH + "storm.vert", RESOURCE_PATH + "blink.frag"))
             return false;
 
         return true;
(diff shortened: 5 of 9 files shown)


eXpl0it3r:
That's why I personally always fought for installing the examples as well, so everything is neatly put next to each other.

This feels rather hacky in a way and users can't use the code as-is anymore, but they need to know, that you have to provide a define via build command...
Personally, I'd much rather vote for an CMake option to optionally install the examples again.

ChrisThrasher:
For me this is about convenience. I think users are most likely to run example programs from the binary directory while they're playing around with the source code. I doubt many users want to install example programs to their system. They also probably don't want to keep changing directories to figure out where they have to run these programs to get them to work. I want the examples to be easy to build and run. Making them easy to copy-paste is a lesser concern if they can't even get it working in the first place.

ChrisThrasher:
> This feels rather hacky in a way

What's the recommended way of providing paths to assets such that an executable can be run from any directory? This is the pattern I've settled on but I'm sure there are more solutions I've never seen.

Bromeon:
> I doubt many users want to install example programs to their system.

What do you mean? If I'm not mistaken, installing is the standard way of "getting clean CMake output". It's also what we recommend when building SFML from source.

It just copies stuff in an arranged way to a user-selected directory, it's not installing in the sense of "placing deep regedit entries" or similar...

ChrisThrasher:
> installing is the standard way of "getting clean CMake output"

I'm not sure what you mean by this. I just want future developers to have an easier time running example programs. I'm happy if we settle on a different pattern as long as we no longer require everybody to `cd` around the source tree trying to figure out where they have to launch the examples. That's what I'm trying to prevent. If you have a better idea for how to achieve that goal, let's make it happen.

eXpl0it3r:
Examples essentially serve two purposes:

1. Give library users an idea of how to use the library
2. Help library developers have a quick test whether the change broke something

This change is rather confusing for the first group, since there's now some magic RESOURCEPATH. Granted, it's not *very* complex, but still, the example stops being self-contained.

For the second group, I think it's not unreasonable to just install the example along side the library while you're developing on stuff. I personally always execute the install target anyways, thus having an option for those who want, to also install the examples would "fix" this, while keeping the examples self-contained.

Bromeon:
@ChrisThrasher I see, you mean developers of SFML itself, I was mostly thinking about users who install SFML after building.

But isn't the resource path only part of the problem? For shared libraries, the working dir or path to .dll/.so needs to be correct as well (unless you start from an IDE where things are automagically set up). That said, if you think this change helps speed up the workflow, I'm not against it.

Should we not define the macro directly as a `std::string` object? This would make example code more readable, by allowing `RESOURCE_PATH` to be used instead of `std::string(RESOURCE_PATH)`.

ChrisThrasher:
@eXpl0it3r 

 > This change is rather confusing for the first group, since there's now some magic RESOURCEPATH. Granted, it's not very complex, but still, the example stops being self-contained.

This is a fair point. I think the complexity is worthwhile because it adds useful functionality but I understand why you might not like that.

 > For the second group, I think it's not unreasonable to just install the example along side the library while you're developing on stuff.

How exactly would we install the resources with the examples? Once we do that, won't that still mean that people have to cd into the installed location of the example program before they can run them since those relative file paths require doing that? I'm not sure how installing the examples lets you more easily run the examples. It just changes what directory you have to run them from.

@Bromeon 

 >  I see, you mean developers of SFML itself
 
Right, I'm focused on SFML members and contributors and anyone curious with the SFML source code having an easier time running the example programs since they're really cool and I want all new SFML developers to have fun with them. 

 > For shared libraries, the working dir or path to .dll/.so needs to be correct as well
 
I'm not sure about this. I don't know enough about .dll and .so files to say how these PR might affect that.

 > Should we not define the macro directly as a std::string object? This would make example code more readable, by allowing RESOURCE_PATH to be used instead of std::string(RESOURCE_PATH).
 
This is preferable but may not be possible. I tried changing
```cmake
target_compile_definitions(${target} PRIVATE RESOURCE_PATH="${CMAKE_CURRENT_SOURCE_DIR}/resources/")
```
to
```cmake
target_compile_definitions(${target} PRIVATE RESOURCE_PATH="std::string(${CMAKE_CURRENT_SOURCE_DIR}/resources/)")
```
but apparently that's not a portable thing you can do.
```
WARNING: Function-style preprocessor definitions may not be passed on the compiler command line because many compilers do not support it.
CMake is dropping a preprocessor definition: std::string(/Users/thrasher/Projects/sfml/examples/ftp/resources/)
Consider defining the macro in a (configured) header file.
```

Bromeon:
Does
```cmake
target_compile_definitions(${target} PRIVATE RESOURCE_PATH="std::string(${CMAKE_CURRENT_SOURCE_DIR}/resources/)")
```
not lack the quotes? We might need to escape them, e.g.
```cmake
target_compile_definitions(${target} PRIVATE RESOURCE_PATH="std::string(\"${CMAKE_CURRENT_SOURCE_DIR}/resources/\")")
```
But not sure about the exact syntax.

Escaping might even be necessary for parentheses, see e.g. [here](https://github.com/pavelliavonau/cmakeconverter/issues/58).

ChrisThrasher:
@Bromeon  I got it to work! Thanks for leading me in the right direction. We need to escape the parentheses otherwise CMake will think the `target_compile_definitions` call will fail. Then we need to escape the quotes because need those to appear in the C++ code. We can't quote the whole thing because we need this to expand into C++ code, not a string literal that looks like C++ code.
```cmake
target_compile_definitions(${target} PRIVATE RESOURCE_PATH=std::string\(\"${CMAKE_CURRENT_SOURCE_DIR}/resources/\"\))
```

vittorioromeo:
We should limit the proliferation of macros. I would suggest two improvements:

1. Rename the macro to something a bit more verbose and less collision-prone, e.g. `SFML_EXAMPLES_RESOURCE_PATH`.

2. Set the macro to a C-style string via CMake, and instead create a `constexpr std::string_view` or `const std::string` in every source file. E.g.

```cpp
constexpr std::string_view resourcePath = SFML_EXAMPLES_RESOURCE_PATH;
// ... 
if (!font.loadFromFile(resourcePath + "tuffy.ttf"))
    return EXIT_FAILURE;
```

ChrisThrasher:
> more verbose and less collision-prone
 
I figure because these examples are not library code we could get away without verbose prefixes on identifiers. Are you worried that `RESOURCE_PATH` may conflict with 3rd party headers or something like that?

 > Set the macro to a C-style string via CMake, and instead create a constexpr std::string_view or const std::string in every source file. E.g.
 
I can do that but it doesn't change the fact that we still have to use a macro to inject user-specific paths into the program. Now you just have two different identifiers that mean the same thing.

vittorioromeo:
> > more verbose and less collision-prone
> 
> I figure because these examples are not library code we could get away without verbose prefixes on identifiers. Are you worried that `RESOURCE_PATH` may conflict with 3rd party headers or something like that?

Possibly. Also, we should promote good practices in example code -- people are going to use it as a basis for learning.

> > Set the macro to a C-style string via CMake, and instead create a constexpr std::string_view or const std::string in every source file. E.g.
> 
> I can do that but it doesn't change the fact that we still have to use a macro to inject user-specific paths into the program. Now you just have two different identifiers that mean the same thing.

The difference is that what I suggested is much simpler to grok, especially if the macro is used in multiple places in the same source file. When you see `RESOURCE_PATH + something`, you have to understand what `RESOURCE_PATH` expands to, if there is any possibility of the `+` operator getting somehow used incorrectly due to the macro expansion, and so on. If you see `constexpr std::string_view resourcePath = SFML_EXAMPLE_RESOURCE_PATH;`, it is quite obvious that you're just initializing a string. And when you see `resourcePath + something`, you can rely on the fact that `operator+` of `std::string_view` is being invoked without having to know what the macro expands to.

Also, you could do

```cpp
constexpr std::string_view resourcePath = SFML_EXAMPLE_RESOURCE_PATH;
#undef SFML_EXAMPLE_RESOURCE_PATH
```

which limits the scope of the macro to a single use, avoiding any possible issue/comprehension difficulty, and -- again -- promoting good practice for our users.

eXpl0it3r:
> It just changes what directory you have to run them from.

I actually didn't consider this, but I'm still not happy with this change.

- If you build shared libraries and don't install them to a global library directory (default on Windows), all of this doesn't *really* matter, since you still have to copy the shared libraries next to the executable
- Our focus should IMHO be much more on the end-user experience and way less on our own development experience

Examples not only show you have to use SFML, but they also guide new users on how to do things.
This isn't just theoretical, you can see lots of people use example code provides in examples, tutorials or documentation.
I don't want to explain to people where the magic `RESOURCE_PATH` is coming from, nor would I want to see them adapt this "practice" in their code, when only purpose for this is, that you save yourself some directory navigations...
Additionally, I would even call it a bad practice to embed absolute paths in your application.

In my experience people are more likely to copy & paste the code, than to (re)compile SFML and build the example that way.

PR #1959, opened 2022-01-06
Use operator<< for printing during test failure

ChrisThrasher:
## Description

Doctest's first recommendation for test formatters is `operator<<`. `doctest::String toString()` is recommended if `operator<<` already exists and you want a test-specific formatter but SFML does not define `operator<<(std::ostream&, const T&)` for any types so we can use `operator<<` for our test formatters. This has the added benefit of removing `doctest.h` from our utility files and makes the implementation of these formatters slightly simpler.

See #1944 for more discussions on this.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/test/Graphics/Color.cpp
+++ b/test/Graphics/Color.cpp
@@ -1,6 +1,8 @@
 #include <SFML/Graphics/Color.hpp>
 #include "GraphicsUtil.hpp"
 
+#include <doctest.h>
+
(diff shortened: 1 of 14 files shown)


eXpl0it3r:

vittorioromeo (approved):

PR #1949, opened 2021-12-31
Only build Cocoa example when using Xcode

ChrisThrasher:
## Description

The Cocoa example program requires Xcode so I disabled this when using any other CMake generator on macOS.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Build SFML with SFML_BUILD_EXAMPLES=ON on macOS using Makefiles. Observe the following error:
```
xcode-select: error: tool 'ibtool' requires Xcode, but active developer directory '/Library/Developer/CommandLineTools' is a command line tools instance
```

PR #1918, opened 2021-12-19
Only compile Catch main once

ChrisThrasher:
## Description

Because Catch is so expensive to compile, it's really helpful for initial compile times if this target doesn't have to be recompiled so often. This won't make a huge impact on developers who are mostly doing incremental builds but can have a meaningful impact on CI runners that are always doing clean builds.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

PR #1960, opened 2022-01-08
Fix test and example incompatibility with ARM Macs

ChrisThrasher:
## Description

This includes a few commits from the master branch which fix issues with compiling the examples and tests on Macs, particle Apple Silicon Macs. For the sake of having simple cherry-picks without conflicts I included a commit that reduces the number of times CatchMain.cpp is compiled which is also a nice benefit since that speeds up compile times significantly. 

See #1918, #1898, and #1949 for more discussion about these commits.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android


eXpl0it3r:

eXpl0it3r (changes requested):
Can't really update Catch on the 2.6.x branch as it requires C++17 and also bumps the CMake version.

ChrisThrasher:
Catch2 v2 uses C++11, not 17 but now I remember that SFML 2.x uses C++03 so it doesn't matter either way. I expected that to have broken the CI pipeline but everything is still passing. Not sure what to make of that.

The latest commit in the Catch2 1.x release series doesn't build for Arm Macs so I suppose the tests will always be broken for Arm Macs. That's too bad. I'll keep the example fix in here since that's still a small compatibility improvement.

ChrisThrasher:
> it requires C++17

Are you referring to this line in test/CMakeLists.txt?

```cmake
target_compile_features(sfml-test-main PRIVATE cxx_std_17)
```

I forgot this was in here but I use Catch2 v2.13.7 in many other C++11 projects without issue. I tried removing that and my local build still works. Either way it's not a huge deal so I'll just keep those commits out of this PR.

eXpl0it3r:
Don't exactly see an easy solution for this either. I want to keep SFML 2.x C++03 compatible, if it means that tests can't be executed on a newer platform, then I find this a better trade off, than requiring all the other platforms to support C++11.

SFML 2.6 will be the last SFML 2.x version (excluding potential bugfix releases), so there shouldn't really be a big issue not having tests executed on macOS ARM

eXpl0it3r:

eXpl0it3r (approved):

PR #1964, opened 2022-01-10
Use std::filesystem::path

ChrisThrasher:
## Description

This PR replaces uses of `std::string` with `std::filesystem::path` when representing filesystem paths. This better expresses intent and opens the door for better supporting non-ASCII file paths. Because `std::filesystem::path` can freely be converted to and from strings, this change doesn't introduce much more type safety but also minimally breaks existing code that is using strings to represent paths.

This PR includes a few instances where I had to refactor some small functions that used `std::string` member functions and were easily replaced by `std::filesystem::path` member functions, in particular `std::filesystem::path::extension()`.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android


ChrisThrasher:
```
In file included from /home/runner/work/SFML/SFML/src/SFML/System/FileInputStream.cpp:28:
/home/runner/work/SFML/SFML/include/SFML/System/FileInputStream.hpp:37:10: fatal error: 'filesystem' file not found
#include <filesystem>
         ^~~~~~~~~~~~
```
The Android build isn't happy with using `<filesystem>`. Anyone know what this is about? @vittorioromeo Is the only one I know of to have used this header in the master branch.

ChrisThrasher:
I'm also running into plenty of issues with `std::filesystem::path::c_str()` returning a type that isn't convertible to `const char*` which is breaking external APIs that expect `const char*` paths. Not yet sure how to handle this. I guess we'll have to just brute force convert the path to an ASCII representation and hope that we're not discarding non-ASCII characters that may be in the path.

Eremiell:
Update of the Android toolchain may help with the issue, the latest NDK r23b has clang12 toolchain. The failed build above seems to suggest the currently used one sports gcc4.9. https://github.com/android/ndk/wiki/Changelog-r23

eXpl0it3r:

src/SFML/Network/Ftp.cpp
```diff
@@ -269,43 +269,43 @@ Ftp::Response Ftp::deleteDirectory(const std::string& name)
...
     if (response.isOk())
     {
         // Tell the server to start the transfer
-        response = sendCommand("RETR", remoteFile);
+        response = sendCommand("RETR", remoteFile.string());
         if (response.isOk())
         {
             // Extract the filename from the file path
-            std::string filename = remoteFile;
+            std::string filename = remoteFile.string();
```
reconndev, line 302: Since you're already introducing `<filesystem>` it would be better to use proper functions instead of only converting to string every time. `std::filesystem::path` is already implicitly convertible to `std::string` so only that change would not introduce any benefit at all:
```cpp
void usePath(const std::string& path);

usePath(std::filesystem::current_path());  // works
```

Here the comment says: `// Extract the filename from the file path` so maybe the intention was `remoteFile.filename()` or `remoteFile.stem()`

src/SFML/System/FileInputStream.cpp
```diff
@@ -60,13 +60,17 @@ FileInputStream& FileInputStream::operator=(FileInputStream&&) = default;
...
 ////////////////////////////////////////////////////////////
-bool FileInputStream::open(const std::string& filename)
+bool FileInputStream::open(const std::filesystem::path& filename)
 {
 #ifdef SFML_SYSTEM_ANDROID
     m_file = std::make_unique<priv::ResourceStream>(filename);
     return m_file->tell() != -1;
+#else
+#ifdef SFML_SYSTEM_WINDOWS
+    m_file.reset(_wfopen(filename.c_str(), L"rb"));
```
reconndev, line 70: `std::fopen` does not work on Windows? The cppreference documentation seems to state it is possible.

eXpl0it3r: I don't know what the "right" solution is, but handling non ASCII characters is still a major issue and I think `std::fopen` can't handle it properly. Not sure if `c_str()` however defeats this. I could expect a `filename.wstring()` here if we go with `_wfopen`.

src/SFML/Network/Ftp.cpp
ChrisThrasher: `std::filesystem::path` is convertible to `std::string` on Linux and macOS but not on Windows. I had to add `.string()` all over the place to appease the Windows CI jobs that kept failing without that. Maybe we should using `auto` in more contexts so Windows builds can automatically use wide character strings but other OSes can use `std::string`.

src/SFML/System/FileInputStream.cpp
ChrisThrasher: `std::filesystem::path::c_str()` returns a `const wchar_t*` on Windows which is not convertible to the `const char*` that `std::fopen` needs. This is one instance of a larger problem that Windows paths use wide characters and many external APIs we're using (including `std::fopen`) use `const char*` for paths. You can do a lossy conversion like `path.string().c_str()` which is guaranteed to convert to a `std::string` then a `const char*` but that may break code that actually needs to use non-ASCII characters in paths. 

Because there's apparently a Windows-specific analog to `std::fopen` I added preprocess logic to use that when appropriate but in most other circumstances, the lossy conversion is the best option I could come up with.

ChrisThrasher:
The iOS build is complaining about some `std::filesystem` utilities not being available until iOS 13.
```
/Users/SFML/Desktop/buildbot/ios-clang-64/build/src/SFML/System/FileInputStream.cpp:72:38: error: 'c_str' is unavailable: introduced in iOS 13.0
    m_file.reset(std::fopen(filename.c_str(), "rb"));
```
Does this warrant upgrading our iOS requirement to 13? That's a few years old now.

src/SFML/Network/Ftp.cpp
ChrisThrasher: > it would be better to use proper functions instead of only converting to string every time
You're right. I was just too lazy to replace this code without whatever the `<filesystem>` counterpart is. I just updated the PR doing exactly that. It's a lot shorter and easier to read now.

src/SFML/System/FileInputStream.cpp
reconndev: Right, I checked and unless `std::fstream` is used, one has to work around using macros like this, sadly.

ChrisThrasher: I feel like once `std::filesystem` support lands in the API, there will be a whole second effort testing and validating that non-ASCII paths can be used if that's even possible. I know that Freetype and OpenAL are two of the external APIs we're using that expects `const char*` paths so there may be limitation to how widespread we can make non-ASCII support if 3rd party APIs don't support it.

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

ChrisThrasher:
Looks like the last problem is the SFML-Buildbot iOS build job failing because it's targeting something older than iOS 13 where `<filesystem>` was added. We think the problem is that the machine running that CI job is too old to have a version of Xcode that can target iOS 13 or newer so we suspect upgrading to a newer macOS version will fix that problem.

Does anyone know anything about the machine running that job or how we might fix the issue of iOS 12 (or older) being used still?

eXpl0it3r:

ChrisThrasher:
I think we're good now! Thanks to help from @Eremiell we were able to work through issues caused by the Android and iOS toolchains being too out of date to work with C++17's `<filesystem>`. Now the CI pipeline is happy with these changes. If the maintainers would prefer, the earlier commits in this PR can be broken out into some number of smaller PRs but I think it's fitting to keep this all together since the updates to the toolchains are only necessitated by the addition of `<filesystem>`.

eXpl0it3r:

src/SFML/Audio/ALCheck.cpp
```diff
@@ -42,14 +42,13 @@ namespace sf
 namespace priv
 {
 ////////////////////////////////////////////////////////////
-void alCheckError(const char* file, unsigned int line, const char* expression)
+void alCheckError(const std::filesystem::path& file, unsigned int line, const char* expression)
```
eXpl0it3r, line 45: Does this work as expected? Why didn't we use `std::string` on the function directly in the past? 🤔

src/SFML/Audio/AudioDevice.cpp
```diff
@@ -108,7 +108,7 @@ AudioDevice::~AudioDevice()
 
 
 ////////////////////////////////////////////////////////////
-bool AudioDevice::isExtensionSupported(const std::string& extension)
+bool AudioDevice::isExtensionSupported(const std::filesystem::path& extension)
```
eXpl0it3r, line 111: Is this really a path and not just the extension (e.g. .flac)?

include/SFML/Network/Ftp.hpp
```diff
@@ -505,7 +506,7 @@ class SFML_NETWORK_API Ftp
     /// \see download
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] Response upload(const std::string& localFile, const std::string& remotePath, TransferMode mode = Binary, bool append = false);
+    [[nodiscard]] Response upload(const std::filesystem::path& localFile, const std::filesystem::path& remotePath, TransferMode mode = Binary, bool append = false);
```
eXpl0it3r, line 509: Are these *always* valid paths?

src/SFML/Graphics/GLCheck.cpp
```diff
@@ -35,14 +35,13 @@ namespace sf
 namespace priv
 {
 ////////////////////////////////////////////////////////////
-void glCheckError(const char* file, unsigned int line, const char* expression)
+void glCheckError(const std::filesystem::path& file, unsigned int line, const char* expression)
```
eXpl0it3r, line 38: Same question as for alCheckError

src/SFML/Graphics/ImageLoader.cpp
```diff
@@ -238,40 +239,39 @@ bool ImageLoader::loadImageFromStream(InputStream& stream, std::vector<Uint8>& p
...
 {
     // Make sure the image is not empty
     if (!pixels.empty() && (size.x > 0) && (size.y > 0))
     {
         // Deduce the image type from its extension
 
         // Extract the extension
-        const std::size_t dot = filename.find_last_of('.');
-        const std::string extension = dot != std::string::npos ? toLower(filename.substr(dot + 1)) : "";
+        const std::string extension = filename.extension().string();
```
eXpl0it3r, line 250: Does it need to be a  `std::string`?

src/SFML/Audio/AudioDevice.cpp
ChrisThrasher: I used `std::filesystem::path` because `std::filesystem::path::extension` returns a `std::filesystem::path`. Apparently that's how the standard library views extensions.

src/SFML/Audio/ALCheck.cpp
ChrisThrasher: As far as I can tell it's fine. No idea why we didn't use `std::string` before.

include/SFML/Network/Ftp.hpp
ChrisThrasher: What do you mean?

src/SFML/Graphics/GLCheck.cpp
ChrisThrasher: Seems fine ¯\_(ツ)_/¯

src/SFML/Graphics/ImageLoader.cpp
ChrisThrasher: A quick local test shows that `extension` could be a `std::filesystem::path` and the comparisons to character arrays on later lines would still compile and presumably behave the same.

include/SFML/Network/Ftp.hpp
eXpl0it3r: I don't know exactly, just wondering if the FTP protocol enforces normal paths or if there are also cases where you might not have "normal" paths but just some random strings 🤔

ChrisThrasher: I see. `std::filesystem::path` does normalize the path to remove unnecessary bits like `../` or multiple consecutive backslashes since those things add no extra info to a path. Maybe that removal of characters could break something here. I'm not sure.

eXpl0it3r:

ChrisThrasher:
If this gets merged we should also be able to remove the GCC 8 workarounds in the unit tests that remove `<filesystem>` for that compiler.

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

include/SFML/Network/Ftp.hpp
vittorioromeo: Has any research/testing been done here? Maybe worthwhile revert the changes here to merge the PR

ChrisThrasher: No I haven't tested this particular snippet. I'll revert this for now.

eXpl0it3r:

eXpl0it3r:

vittorioromeo (approved):
LGTM, thank you so much!

vittorioromeo:

eXpl0it3r (approved):
Looks good to me. Thank you a lot for all this effort! 🙂 

Btw. did you ever give the FTP client a test run, after these changes?

vittorioromeo:
> Looks good to me. Thank you a lot for all this effort! 🙂
> 
> Btw. did you ever give the FTP client a test run, after these changes?

I had @ChrisThrasher remove the code that you were concerned about, but I do agree it would be a good idea to test it anyway.

ChrisThrasher:
> Btw. did you ever give the FTP client a test run, after these changes?

I did not. I did however revert one instance of `filesystem` in the FTP class which looked suspicious.

vittorioromeo:
I'll merge and try out the FTP example on `master`. Will let you know if there is anything fishy...

src/SFML/Audio/AudioDevice.cpp
ChrisThrasher: I think you're totally right. Now that I look more deeply, I don't see any references to filesystem extensions. See #2024. Thanks!

include/SFML/Network/Ftp.hpp
ChrisThrasher: On second inspection, `std::filesystem::path::path` does not appear to modify the string you give it upon construction. It merely stores that string. Any manipulation of that path happens later upon request when calling functions like [`std::filesystem::canonical`](https://en.cppreference.com/w/cpp/filesystem/canonical). I believe it's safe to use `std::filesystem` here.

PR #1965, opened 2022-01-11
Reintroduce 'SleepImpl' sleep implementation

### LINKED ISSUE #1962: Recent `sf::Time` changes result in choppy gameplay with frametime limit enabled
PR #1872 resulted in dramatic performance degradation (multiple reports of choppiness, unstable FPS, input lag) in Open Hexagon when `Window::setFramerateLimit` is used due to commit db83f857feeacaa30a482aa6059b9358e9ccb2ef. Reverting that commit fixes the issue. 

I am not entirely sure what the problem is, but `sf::sleep` seems to be very inaccurate with that commit, possibly related to #1946. For example, setting an FPS limit to 60 results in very varied FPS between 50-64, while with that commit reverted it is quite accurate. I think we might have to revert this one once we figure out exactly why it is inaccurate and if that can be fixed.

vittorioromeo:
Fixes #1962.

### DIFF
--- a/src/SFML/System/Win32/SleepImpl.cpp
+++ b/src/SFML/System/Win32/SleepImpl.cpp
@@ -1,7 +1,7 @@
 ////////////////////////////////////////////////////////////
 //
 // SFML - Simple and Fast Multimedia Library
-// Copyright (C) 2007-2021 Laurent Gomila (laurent@sfml-dev.org)
+// Copyright (C) 2007-2022 Laurent Gomila (laurent@sfml-dev.org)
 //
 // This software is provided 'as-is', without any express or implied warranty.
 // In no event will the authors be held liable for any damages arising from the use of this software.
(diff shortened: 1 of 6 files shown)


Bromeon:
Is `this_thread::sleep_for()` only broken on MinGW, or Windows in general?

One option would be to use standard implementation on those platforms where it's reliable, but I also see that this may be an unnecessary risk for bugs/regressions in the future. Maybe we should just stick to something we know to work.

vittorioromeo:
> Is `this_thread::sleep_for()` only broken on MinGW, or Windows in general?

This problem is specific to the `winpthreads` library that is bundled with any MinGW-w64 distribution, including MSYS2, Nuwen, and the CodeBlocks ones, which are the most popular. I don't think it's "broken", it's just not implemented optimally. For example,  
a [one-line change](https://stackoverflow.com/a/54495870/598696) might fix the issue, but I have not tried it yet. It also seems that [Visual Studio/MSVC suffers from similar issues](https://stackoverflow.com/a/32963363/598696).

> One option would be to use standard implementation on those platforms where it's reliable, but I also see that this may be an unnecessary risk for bugs/regressions in the future. Maybe we should just stick to something we know to work.

I don't really know how to test reliability. I think that the latter is a better option, unless someone else wants to try other possibilities.

vittorioromeo:
I did some more research and I'm not sure this PR is a good idea. Check https://github.com/SFML/SFML/issues/1962.

vittorioromeo:
I did some more testing/research and this PR is probably the best way to go if we don't want to change how `setFramerateLimit` works.

src/SFML/System/Win32/SleepImpl.cpp
```diff
@@ -0,0 +1,55 @@
...
+////////////////////////////////////////////////////////////
+void sleepImpl(Time time)
+{
+    // Get the minimum supported timer resolution on this system
+    static const UINT periodMin = []
+    {
+        TIMECAPS tc;
+        timeGetDevCaps(&tc, sizeof(TIMECAPS));
+        return tc.wPeriodMin;
+    }();
```
eXpl0it3r, line 43: I assume this is to reduce the calls to `timeGetDevCaps` right?

eXpl0it3r (approved):

src/SFML/System/Win32/SleepImpl.cpp
vittorioromeo: Yes, this was originally suggested by @reconndev. It's not really about performance, but more about communicating the fact that `wPeriodMin` will never change -- it's hardware-specific.

PR #1966, opened 2022-01-11
Add tests for sf::BlendMode

ChrisThrasher:
## Description

Next up in my crusade to test the whole Graphics module is `sf::BlendMode`. This is effectively another leaf in the dependency graph since it only depends on `SFML/Graphics/Export.hpp`. These tests are pretty straightforward so hopefully they'll be self explanatory.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -32,6 +32,7 @@ endif()
 
 if(SFML_BUILD_GRAPHICS)
     SET(GRAPHICS_SRC
+        "${SRCROOT}/Graphics/BlendMode.cpp"
         "${SRCROOT}/Graphics/Color.cpp"
         "${SRCROOT}/Graphics/Rect.cpp"
         "${SRCROOT}/Graphics/Transform.cpp"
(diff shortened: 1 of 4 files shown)


ChrisThrasher:
`SFML/Graphics/RenderStates.hpp` is next since it depends on `BlendMode` and `Transform` which both have tests.

vittorioromeo:

Bromeon:
I wonder if it makes sense if we already qualify the enums, as we're likely going to turn them into enum classes soon.
E.g.:
```cpp
// instead of:
sf::BlendMode::Zero

// use:
using Factor = sf::BlendMode::Factor;
Factor::Zero
```

But we can also do everything together in a later PR. [These constants](https://github.com/SFML/SFML/blob/master/include/SFML/Graphics/BlendMode.hpp#L152-L157) are also candidates for nesting, for example.

ChrisThrasher:
@Bromeon There are certainly many unscoped enumeration that should be scoped. We could feasible do one big PR that changes lots of classic enumerations into scoped enumerations. I did also find it odd that those constants were not in the scope of the BlendMode class and would like to change that.

eXpl0it3r (approved):

eXpl0it3r:

PR #717, opened 2014-10-12
Added a wrapper class for angles

MarioLiebisch:
`sf::Angle` works in a similar way to `sf::Time` providing conversion between radians and degrees in addition to basic mathematical operands.

Angles are always normalized to provide values within _[0, 360[ deg_ resp. _[0, 2 PI[ rad_.

In addition, there's a constructor taking a floating point value to provide backward compatibility with existing source code.

Right now conversion offers radians as well as degrees, although I could imagine adding [some more units](http://en.wikipedia.org/wiki/Angle#Units), such as _grads_ or _seconds of arc_.

Issue #1747, opened 2021-01-21
Suggestion : remove unnecessary use of runtime polymorphism 


groundswellaudio:
I've found a lot of places in the API where virtual inheritance is used without an apparent reason. 
I propose to remove them in order to : 
* improve performance
* simplify the design 
* encourage better coding practices for beginners
All uses that i have seen can be rewritten without virtual inheritance without breaking existing code (at least without breaking the code that doesn't store objects with base classes pointers, but all of these base classes weren't really designed for that anyway) 

Here is a non-exhaustive list, followed by proposed solutions : 

1/ RenderTarget : use inheritance as a way to reuse the common code in the RenderTarget base class. 
2/ Drawable : use a virtual function draw() for no particular reason. Used by RenderTarget interface
3/ Shape : use a virtual function for accessing the points
4/ Socket : has a virtual destructor but no virtual functions
5/ SoundSource : has play/pause/stop virtual functions
6/ Window : has onCreate() and onResize() virtual functions

Propositions : 
1/ remove the base class, stick the reusable code in a CRTP or mixin class, make all the functions taking a RenderTarget a template, and forward the needed data of Target type to the implementation in the .cpp file (e.g instead of querying getSize() through the vtable, take a template type and pass the result of getSize() to the implementation) 
2/ remove the base class, and transform the draw(Drawable) methods for "RenderTargetable" types into a template, since there is zero logic inside of it already
3/ same as 1/, remove the base class, make template functions, assume that any ShapeType has a points() methods returning a range of points
4/ remove virtual destructor and the enum, have SocketSelector store a std::variant 
5/ it's not clear to me why does this class needs virtual functions? code reuse? if so, replace by a mixin or CRTP class
6/ i don't understand why these functions exists : if i want some code to run after i called create() in my program, i can just shadow the create() function in my class, and resize has a dedicated event already. remove virtual inheritance, stick common code somewhere else

There might be a few more. If these changes are approved, i can work on it.

LaurentGomila:
Just to be clear: this is about *public* inheritance, not *virtual* inheritance; these are two different (but non-exclusive) concepts in C++.

I strongly disagree with your proposed changes 😉 

### Templates
Converting half of SFML and user code to templated code, just to avoid indirections (v-table calls), is just insane.
- Templates increase compilation times, make you expose unecessary code, and create more dependencies between files.
- Templates don't enforce any design -- how do I know the contract to implement to make my class compatible with the template functions? With a base class, everything is clear.
[...]

PR #1969, opened 2022-01-14
Add sf::Angle

ChrisThrasher:

----

## Description

Adds `sf::Angle`, a type safe representation of angles. Similar to `sf::Time`, `sf::Angle` provides an API for working with angles and provides named functions for converting to and from degrees and radians.

This PR is related to the issue #717

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/examples/X11/X11.cpp
+++ b/examples/X11/X11.cpp
@@ -47,8 +47,7 @@
     // Setup a perspective projection
     glMatrixMode(GL_PROJECTION);
     glLoadIdentity();
-    static const float pi = 3.141592654f;
-    float extent = std::tan(90.0f * pi / 360.0f);
+    float extent = std::tan(sf::degrees(45).asRadians());
 
 #ifdef SFML_OPENGL_ES
     glFrustumf(-extent, extent, -extent, extent, 1.0f, 500.0f);
--- a/include/SFML/Graphics/Transform.hpp
+++ b/include/SFML/Graphics/Transform.hpp
@@ -31,6 +31,7 @@
 #include <SFML/Graphics/Export.hpp>
 #include <SFML/Graphics/Rect.hpp>
 #include <SFML/System/Vector2.hpp>
+#include <SFML/System/Angle.hpp>
 
 
 namespace sf
@@ -156,7 +157,7 @@ public:
     /// can be chained.
     /// \code
     /// sf::Transform transform;
-    /// transform.translate(sf::Vector2f(100, 200)).rotate(45);
+    /// transform.translate(sf::Vector2f(100, 200)).rotate(sf::degrees(45));
     /// \endcode
     ///
     /// \param offset Translation offset to apply
@@ -175,17 +176,17 @@ public:
     /// can be chained.
     /// \code
     /// sf::Transform transform;
-    /// transform.rotate(90).translate(50, 20);
+    /// transform.rotate(sf::degrees(90)).translate(50, 20);
     /// \endcode
     ///
-    /// \param angle Rotation angle, in degrees
+    /// \param angle Rotation angle
     ///
     /// \return Reference to *this
     ///
     /// \see translate, scale
     ///
     ////////////////////////////////////////////////////////////
-    Transform& rotate(float angle);
+    Transform& rotate(Angle angle);
 
     ////////////////////////////////////////////////////////////
     /// \brief Combine the current transform with a rotation
@@ -199,10 +200,10 @@ public:
     /// can be chained.
     /// \code
     /// sf::Transform transform;
-    /// transform.rotate(90, sf::Vector2f(8, 3)).translate(sf::Vector2f(50, 20));
+    /// transform.rotate(sf::degrees(90), sf::Vector2f(8, 3)).translate(sf::Vector2f(50, 20));
     /// \endcode
     ///
-    /// \param angle Rotation angle, in degrees
+    /// \param angle Rotation angle
     /// \param center Center of rotation
     ///
     /// \return Reference to *this
@@ -210,7 +211,7 @@ public:
     /// \see translate, scale
     ///
     ////////////////////////////////////////////////////////////
-    Transform& rotate(float angle, const Vector2f& center);
+    Transform& rotate(Angle angle, const Vector2f& center);
 
     ////////////////////////////////////////////////////////////
     /// \brief Combine the current transform with a scaling
@@ -219,7 +220,7 @@ public:
     /// can be chained.
     /// \code
     /// sf::Transform transform;
-    /// transform.scale(sf::Vector2f(2, 1)).rotate(45);
+    /// transform.scale(sf::Vector2f(2, 1)).rotate(sf::degrees(45));
     /// \endcode
     ///
     /// \param factors Scaling factors
--- a/include/SFML/System/Angle.hpp
+++ b/include/SFML/System/Angle.hpp
@@ -1,0 +1,519 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2022 Laurent Gomila (laurent@sfml-dev.org)
+//
+// This software is provided 'as-is', without any express or implied warranty.
+// In no event will the authors be held liable for any damages arising from the use of this software.
+//
+// Permission is granted to anyone to use this software for any purpose,
+// including commercial applications, and to alter it and redistribute it freely,
+// subject to the following restrictions:
+//
+// 1. The origin of this software must not be misrepresented;
+//    you must not claim that you wrote the original software.
+//    If you use this software in a product, an acknowledgment
+//    in the product documentation would be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such,
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+#ifndef SFML_ANGLE_HPP
+#define SFML_ANGLE_HPP
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#include <SFML/System/Export.hpp>
+#include <cassert>
+
+
+namespace sf
+{
+////////////////////////////////////////////////////////////
+/// \brief Represents an angle value.
+///
+////////////////////////////////////////////////////////////
+class Angle
+{
+public:
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Default constructor
+    ///
+    /// Sets the angle value to zero.
+    ///
+    ////////////////////////////////////////////////////////////
+    constexpr Angle();
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Return the angle's value in degrees
+    ///
+    /// \return Angle in degrees
+    ///
+    /// \see asRadians
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] constexpr float asDegrees() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Return the angle's value in radians
+    ///
+    /// \return Angle in radians
+    ///
+    /// \see asDegrees
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] constexpr float asRadians() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Wrap to a range such that -180° <= angle < 180°
+    /// 
+    /// Similar to a modulo operation, this returns a copy of the angle
+    /// constrained to the range [-180°, 180°) == [-Pi, Pi).
+    /// The resulting angle represents a rotation which is equivalent to *this.
+    ///
+    /// The name "signed" originates from the similarity to signed integers:
+    /// <table>
+    /// <tr>
+    ///   <th></th>
+    ///   <th>signed</th>
+    ///   <th>unsigned</th>
+    /// </tr>
+    /// <tr>
+    ///   <td>char</td>
+    ///   <td>[-128, 128)</td>
+    ///   <td>[0, 256)</td>
+    /// </tr>
+    /// <tr>
+    ///   <td>Angle</td>
+    ///   <td>[-180°, 180°)</td>
+    ///   <td>[0°, 360°)</td>
+    /// </tr>
+    /// </table>
+    ///
+    /// \return Signed angle, wrapped to [-180°, 180°)
+    ///
+    /// \see wrapUnsigned
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] constexpr Angle wrapSigned() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Wrap to a range such that 0° <= angle < 360°
+    /// 
+    /// Similar to a modulo operation, this returns a copy of the angle
+    /// constrained to the range [0°, 360°) == [0, Tau) == [0, 2*Pi).
+    /// The resulting angle represents a rotation which is equivalent to *this.
+    ///
+    /// The name "unsigned" originates from the similarity to unsigned integers:
+    /// <table>
+    /// <tr>
+    ///   <th></th>
+    ///   <th>signed</th>
+    ///   <th>unsigned</th>
+    /// </tr>
+    /// <tr>
+    ///   <td>char</td>
+    ///   <td>[-128, 128)</td>
+    ///   <td>[0, 256)</td>
+    /// </tr>
+    /// <tr>
+    ///   <td>Angle</td>
+    ///   <td>[-180°, 180°)</td>
+    ///   <td>[0°, 360°)</td>
+    /// </tr>
+    /// </table>
+    ///
+    /// \return Unsigned angle, wrapped to [0°, 360°)
+    ///
+    /// \see wrapSigned
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] constexpr Angle wrapUnsigned() const;
+
+    ////////////////////////////////////////////////////////////
+    // Static member data
+    ////////////////////////////////////////////////////////////
+    SFML_SYSTEM_API static const Angle Zero; //!< Predefined 0 degree angle value
+
+private:
+
+    friend constexpr Angle degrees(float angle);
+    friend constexpr Angle radians(float angle);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Construct from a number of degrees
+    ///
+    /// This function is internal. To construct angle values,
+    /// use sf::radians or sf::degrees instead.
+    ///
+    /// \param degrees Angle in degrees
+    ///
+    ////////////////////////////////////////////////////////////
+    constexpr explicit Angle(float degrees);
+
+private:
+
+    ////////////////////////////////////////////////////////////
+    // Member data
+    ////////////////////////////////////////////////////////////
+    float m_degrees; //!< Angle value stored as degrees
+};
+
+////////////////////////////////////////////////////////////
+/// \brief Construct an angle value from a number of degrees
+///
+/// \param angle Number of degrees
+///
+/// \return Angle value constructed from the number of degrees
+///
+/// \see radians
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr Angle degrees(float angle);
+
+////////////////////////////////////////////////////////////
+/// \brief Construct an angle value from a number of radians
+///
+/// \param angle Number of radians
+///
+/// \return Angle value constructed from the number of radians
+///
+/// \see degrees
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr Angle radians(float angle);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of == operator to compare two angle values
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (an angle)
+///
+/// \return True if both angle values are equal
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr bool operator ==(Angle left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of != operator to compare two angle values
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (an angle)
+///
+/// \return True if both angle values are different
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr bool operator !=(Angle left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of < operator to compare two angle values
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (an angle)
+///
+/// \return True if \a left is less than \a right
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr bool operator <(Angle left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of > operator to compare two angle values
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (an angle)
+///
+/// \return True if \a left is greater than \a right
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr bool operator >(Angle left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of <= operator to compare two angle values
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (an angle)
+///
+/// \return True if \a left is less than or equal to \a right
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr bool operator <=(Angle left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of >= operator to compare two angle values
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (an angle)
+///
+/// \return True if \a left is greater than or equal to \a right
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr bool operator >=(Angle left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of unary - operator to negate an angle value.
+///
+/// Represents a rotation in the opposite direction.
+///
+/// \param right Right operand (an angle)
+///
+/// \return Negative of the angle value
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr Angle operator -(Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of binary + operator to add two angle values
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (an angle)
+///
+/// \return Sum of the two angle values
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr Angle operator +(Angle left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of binary += operator to add/assign two angle values
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (an angle)
+///
+/// \return Sum of the two angle values
+///
+////////////////////////////////////////////////////////////
+constexpr Angle& operator +=(Angle& left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of binary - operator to subtract two angle values
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (an angle)
+///
+/// \return Difference of the two angle values
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr Angle operator -(Angle left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of binary -= operator to subtract/assign two angle values
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (an angle)
+///
+/// \return Difference of the two angle values
+///
+////////////////////////////////////////////////////////////
+constexpr Angle& operator -=(Angle& left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of binary * operator to scale an angle value
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (a number)
+///
+/// \return \a left multiplied by \a right
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr Angle operator *(Angle left, float right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of binary * operator to scale an angle value
+///
+/// \param left  Left operand (a number)
+/// \param right Right operand (an angle)
+///
+/// \return \a left multiplied by \a right
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr Angle operator *(float left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of binary *= operator to scale/assign an angle value
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (a number)
+///
+/// \return \a left multiplied by \a right
+///
+////////////////////////////////////////////////////////////
+constexpr Angle& operator *=(Angle& left, float right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of binary / operator to scale an angle value
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (a number)
+///
+/// \return \a left divided by \a right
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr Angle operator /(Angle left, float right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of binary /= operator to scale/assign an angle value
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (a number)
+///
+/// \return \a left divided by \a right
+///
+////////////////////////////////////////////////////////////
+constexpr Angle& operator /=(Angle& left, float right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of binary / operator to compute the ratio of two angle values
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (an angle)
+///
+/// \return \a left divided by \a right
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr float operator /(Angle left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of binary % operator to compute modulo of an angle value.
+///
+/// Right hand angle must be greater than zero.
+///
+/// Examples:
+/// \code
+/// sf::degrees(90) % sf::degrees(40)  // 10 degrees
+/// sf::degrees(-90) % sf::degrees(40) // 30 degrees (not -10)
+/// \endcode
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (an angle)
+///
+/// \return \a left modulo \a right
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr Angle operator %(Angle left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of binary %= operator to compute/assign remainder of an angle value
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (an angle)
+///
+/// \return \a left modulo \a right
+///
+////////////////////////////////////////////////////////////
+constexpr Angle& operator %=(Angle& left, Angle right);
+
+namespace Literals
+{
+
+////////////////////////////////////////////////////////////
+/// \relates sf::Angle
+/// \brief User defined literal for angles in degrees, e.g. 10.5_deg
+///
+/// \param angle Angle in degrees
+///
+/// \return \a Angle
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr Angle operator "" _deg(long double angle);
+
+////////////////////////////////////////////////////////////
+/// \relates sf::Angle
+/// \brief User defined literal for angles in degrees, e.g. 90_deg
+///
+/// \param angle Angle in degrees
+///
+/// \return \a Angle
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr Angle operator "" _deg(unsigned long long int angle);
+
+////////////////////////////////////////////////////////////
+/// \relates sf::Angle
+/// \brief User defined literal for angles in radians, e.g. 0.1_rad
+///
+/// \param angle Angle in radians
+///
+/// \return \a Angle
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr Angle operator "" _rad(long double angle);
+
+////////////////////////////////////////////////////////////
+/// \relates sf::Angle
+/// \brief User defined literal for angles in radians, e.g. 2_rad
+///
+/// \param angle Angle in radians
+///
+/// \return \a Angle
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr Angle operator "" _rad(unsigned long long int angle);
+
+} // namespace Literals
+
+#include <SFML/System/Angle.inl>
+
+} // namespace sf
+
+
+#endif // SFML_ANGLE_HPP
+
+
+////////////////////////////////////////////////////////////
+/// \class sf::Angle
+/// \ingroup system
+///
+/// sf::Angle encapsulates an angle value in a flexible way.
+/// It allows for defining an angle value either as a number
+/// of degrees or radians. It also works the other way
+/// around. You can read an angle value as either a number
+/// of degrees or radians.
+///
+/// By using such a flexible interface, the API doesn't
+/// impose any fixed type or unit for angle values and lets
+/// the user choose their own preferred representation.
+///
+/// Angle values support the usual mathematical operations.
+/// You can add or subtract two angles, multiply or divide
+/// an angle by a number, compare two angles, etc.
+///
+/// Usage example:
+/// \code
+/// sf::Angle a1  = sf::degrees(90);
+/// float radians = a1.asRadians(); // 1.5708f
+///
+/// sf::Angle a2 = sf::radians(3.141592654f);
+/// float degrees = a2.asDegrees(); // 180.0f
+///
+/// using namespace sf::Literals;
+/// sf::Angle a3 = 10_deg;   // 10 degrees
+/// sf::Angle a4 = 1.5_deg;  // 1.5 degrees
+/// sf::Angle a5 = 1_rad;    // 1 radians
+/// sf::Angle a6 = 3.14_rad; // 3.14 radians
+/// \endcode
+///
+////////////////////////////////////////////////////////////
--- a/include/SFML/System/Angle.inl
+++ b/include/SFML/System/Angle.inl
@@ -1,0 +1,259 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2022 Laurent Gomila (laurent@sfml-dev.org)
+//
+// This software is provided 'as-is', without any express or implied warranty.
+// In no event will the authors be held liable for any damages arising from the use of this software.
+//
+// Permission is granted to anyone to use this software for any purpose,
+// including commercial applications, and to alter it and redistribute it freely,
+// subject to the following restrictions:
+//
+// 1. The origin of this software must not be misrepresented;
+//    you must not claim that you wrote the original software.
+//    If you use this software in a product, an acknowledgment
+//    in the product documentation would be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such,
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+namespace priv
+{
+    constexpr float pi = 3.141592654f;
+
+    constexpr float positiveRemainder(float a, float b)
+    {
+        assert(b > 0.0f);
+        const float val = a - static_cast<float>(static_cast<int>(a / b)) * b;
+        if (val >= 0.f)
+            return val;
+        else
+            return val + b;
+    }
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle::Angle() :
+m_degrees(0.0f)
+{
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr float Angle::asDegrees() const
+{
+    return m_degrees;
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr float Angle::asRadians() const
+{
+    return m_degrees * (priv::pi / 180);
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle Angle::wrapSigned() const
+{
+    return degrees(priv::positiveRemainder(m_degrees + 180, 360) - 180);
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle Angle::wrapUnsigned() const
+{
+    return degrees(priv::positiveRemainder(m_degrees, 360));
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle::Angle(float degrees) :
+m_degrees(degrees)
+{
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle degrees(float angle)
+{
+    return Angle(angle);
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle radians(float angle)
+{
+    return Angle(angle * (180 / priv::pi));
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr bool operator ==(Angle left, Angle right)
+{
+    return left.asDegrees() == right.asDegrees();
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr bool operator !=(Angle left, Angle right)
+{
+    return left.asDegrees() != right.asDegrees();
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr bool operator <(Angle left, Angle right)
+{
+    return left.asDegrees() < right.asDegrees();
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr bool operator >(Angle left, Angle right)
+{
+    return left.asDegrees() > right.asDegrees();
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr bool operator <=(Angle left, Angle right)
+{
+    return left.asDegrees() <= right.asDegrees();
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr bool operator >=(Angle left, Angle right)
+{
+    return left.asDegrees() >= right.asDegrees();
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle operator -(Angle right)
+{
+    return degrees(-right.asDegrees());
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle operator +(Angle left, Angle right)
+{
+    return degrees(left.asDegrees() + right.asDegrees());
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle& operator +=(Angle& left, Angle right)
+{
+    return left = left + right;
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle operator -(Angle left, Angle right)
+{
+    return degrees(left.asDegrees() - right.asDegrees());
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle& operator -=(Angle& left, Angle right)
+{
+    return left = left - right;
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle operator *(Angle left, float right)
+{
+    return degrees(left.asDegrees() * right);
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle operator *(float left, Angle right)
+{
+    return right * left;
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle& operator *=(Angle& left, float right)
+{
+    return left = left * right;
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle operator /(Angle left, float right)
+{
+    return degrees(left.asDegrees() / right);
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle& operator /=(Angle& left, float right)
+{
+    return left = left / right;
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr float operator /(Angle left, Angle right)
+{
+    return left.asDegrees() / right.asDegrees();
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle operator %(Angle left, Angle right)
+{
+    return degrees(priv::positiveRemainder(left.asDegrees(), right.asDegrees()));
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle& operator %=(Angle& left, Angle right)
+{
+    return left = left % right;
+}
+
+namespace Literals
+{
+
+////////////////////////////////////////////////////////////
+constexpr Angle operator "" _deg(long double angle)
+{
+    return degrees(static_cast<float>(angle));
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle operator "" _deg(unsigned long long angle)
+{
+    return degrees(static_cast<float>(angle));
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle operator "" _rad(long double angle)
+{
+    return radians(static_cast<float>(angle));
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle operator "" _rad(unsigned long long angle)
+{
+    return radians(static_cast<float>(angle));
+}
+
+} // namespace Literals
--- a/src/SFML/Graphics/Transformable.cpp
+++ b/src/SFML/Graphics/Transformable.cpp
@@ -35,7 +35,7 @@ namespace sf
 Transformable::Transformable() :
 m_origin                    (0, 0),
 m_position                  (0, 0),
-m_rotation                  (0),
+m_rotation                  (),
 m_scale                     (1, 1),
 m_transform                 (),
 m_transformNeedUpdate       (true),
@@ -61,11 +61,9 @@ void Transformable::setPosition(const Vector2f& position)
 
 
 ////////////////////////////////////////////////////////////
-void Transformable::setRotation(float angle)
+void Transformable::setRotation(Angle angle)
 {
-    m_rotation = std::fmod(angle, 360.f);
-    if (m_rotation < 0)
-        m_rotation += 360.f;
+    m_rotation = angle.wrapUnsigned();
 
     m_transformNeedUpdate = true;
     m_inverseTransformNeedUpdate = true;
@@ -98,7 +96,7 @@ const Vector2f& Transformable::getPosition() const
 
 
 ////////////////////////////////////////////////////////////
-float Transformable::getRotation() const
+Angle Transformable::getRotation() const
 {
     return m_rotation;
 }
@@ -126,7 +124,7 @@ void Transformable::move(const Vector2f& offset)
 
 
 ////////////////////////////////////////////////////////////
-void Transformable::rotate(float angle)
+void Transformable::rotate(Angle angle)
 {
     setRotation(m_rotation + angle);
 }
@@ -145,7 +143,7 @@ const Transform& Transformable::getTransform() const
     // Recompute the combined transform if needed
     if (m_transformNeedUpdate)
     {
-        float angle  = -m_rotation * 3.141592654f / 180.f;
+        float angle  = -m_rotation.asRadians();
         float cosine = std::cos(angle);
         float sine   = std::sin(angle);
         float sxc    = m_scale.x * cosine;
--- a/src/SFML/Graphics/View.cpp
+++ b/src/SFML/Graphics/View.cpp
@@ -35,7 +35,7 @@ namespace sf
 View::View() :
 m_center             (),
 m_size               (),
-m_rotation           (0),
+m_rotation           (),
 m_viewport           ({0, 0}, {1, 1}),
 m_transformUpdated   (false),
 m_invTransformUpdated(false)
@@ -48,7 +48,7 @@ m_invTransformUpdated(false)
 View::View(const FloatRect& rectangle) :
 m_center             (),
 m_size               (),
-m_rotation           (0),
+m_rotation           (),
 m_viewport           ({0, 0}, {1, 1}),
 m_transformUpdated   (false),
 m_invTransformUpdated(false)
@@ -61,7 +61,7 @@ m_invTransformUpdated(false)
 View::View(const Vector2f& center, const Vector2f& size) :
 m_center             (center),
 m_size               (size),
-m_rotation           (0),
+m_rotation           (),
 m_viewport           ({0, 0}, {1, 1}),
 m_transformUpdated   (false),
 m_invTransformUpdated(false)
@@ -98,11 +98,9 @@ void View::setSize(const Vector2f& size)
 
 
 ////////////////////////////////////////////////////////////
-void View::setRotation(float angle)
+void View::setRotation(Angle angle)
 {
-    m_rotation = std::fmod(angle, 360.f);
-    if (m_rotation < 0)
-        m_rotation += 360.f;
+    m_rotation = angle.wrapUnsigned();
 
     m_transformUpdated    = false;
     m_invTransformUpdated = false;
@@ -123,7 +121,7 @@ void View::reset(const FloatRect& rectangle)
     m_center.y = rectangle.top + rectangle.height / 2.f;
     m_size.x   = rectangle.width;
     m_size.y   = rectangle.height;
-    m_rotation = 0;
+    m_rotation = Angle::Zero;
 
     m_transformUpdated    = false;
     m_invTransformUpdated = false;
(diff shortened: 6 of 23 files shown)


src/SFML/Graphics/Transformable.cpp
```diff
@@ -145,9 +143,8 @@ const Transform& Transformable::getTransform() const
     // Recompute the combined transform if needed
     if (m_transformNeedUpdate)
     {
-        float angle  = -m_rotation * 3.141592654f / 180.f;
-        float cosine = std::cos(angle);
-        float sine   = std::sin(angle);
+        float cosine = static_cast<float>(std::cos(-m_rotation.asRadians()));
+        float sine   = static_cast<float>(std::sin(-m_rotation.asRadians()));
```
kimci86, line 147:
```suggestion
        float cosine = std::cos(-m_rotation.asRadians());
        float sine   = std::sin(-m_rotation.asRadians());
```

src/SFML/Graphics/View.cpp
```diff
@@ -186,9 +184,8 @@ const Transform& View::getTransform() const
     if (!m_transformUpdated)
     {
         // Rotation components
-        float angle  = m_rotation * 3.141592654f / 180.f;
-        float cosine = std::cos(angle);
-        float sine   = std::sin(angle);
+        float cosine = static_cast<float>(std::cos(m_rotation.asRadians()));
+        float sine   = static_cast<float>(std::sin(m_rotation.asRadians()));
```
kimci86, line 188:
```suggestion
        float cosine = std::cos(m_rotation.asRadians());
        float sine   = std::sin(m_rotation.asRadians());
```

test/System/Angle.cpp
```diff
@@ -0,0 +1,204 @@
...
+        {
+            CHECK(sf::radians(0.0f) * 10.0f == sf::Angle::Zero);
+            CHECK((sf::degrees(10.0f) * 2.5f).asDegrees() == Approx(25.0f));
+            CHECK((sf::degrees(100.0f) * 10.0f).asDegrees() == Approx(280.0f));
+
+            CHECK(10.0f * sf::radians(0.0f) == sf::Angle::Zero);
+            CHECK((2.5f * sf::degrees(10.0f)).asDegrees() == Approx(25.0f));
+            CHECK((10.0f * sf::degrees(100.0f)).asDegrees() == Approx(280.0f));
+
+            CHECK(sf::radians(0.0f) * static_cast<sf::Int64>(10) == sf::Angle::Zero);
```
kimci86, line 134: This does not compile without the static cast. GCC complains:
```
error: ambiguous overload for ‘operator*’ (operand types are ‘sf::Angle’ and ‘int’)
note: candidate: ‘constexpr sf::Angle sf::operator*(sf::Angle, float)’
note: candidate: ‘constexpr sf::Angle sf::operator*(sf::Angle, sf::Int64)’
```

It looks like a problem.
Could operators just take `int` instead of `Int64`? Or maybe only accept `float` as for `Vector2f`.

kimci86 (commented):
It looks interesting. I commented some remarks.

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,426 @@
...
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief User defined literal for angles in radians
+///
+/// \param angle Angle in radians
+///
+/// \return \a Angle
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr Angle operator "" _rad(long double angle);
```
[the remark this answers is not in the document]
ChrisThrasher: Curious what @Bromeon and @eXpl0it3r think about adding something like a `sf::literals` namespace for this purpose. In the meantime the following is possible
```cpp
using sf::operator""_deg;
```

include/SFML/Graphics/Transform.hpp
```diff
@@ -175,17 +176,17 @@ class SFML_GRAPHICS_API Transform
     /// can be chained.
     /// \code
     /// sf::Transform transform;
-    /// transform.rotate(90).translate(50, 20);
+    /// transform.rotate(sf::degree(90)).translate(50, 20);
```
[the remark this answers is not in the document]
ChrisThrasher: Good catch. I'll fix this.

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,426 @@
...
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief User defined literal for angles in degrees
+///
+/// \param angle Angle in degrees
+///
+/// \return \a Angle
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr Angle operator "" _deg(long double angle);
```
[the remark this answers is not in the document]
ChrisThrasher: The compiler didn't like me using `float` as the parameter. It might be a language limitation. I'm not totally sure. @vittorioromeo may know more about how this works.

test/System/Angle.cpp
ChrisThrasher: I think we should only have `float` overloads here.

include/SFML/System/Angle.inl
```diff
@@ -0,0 +1,241 @@
...
+////////////////////////////////////////////////////////////
+constexpr Angle degrees(float degrees)
+{
+    // Normalize to [0, 360)
+    while (degrees < 0.0f)
+        degrees += 360.0f;
+    degrees = priv::fmodf(degrees, 360.0f);
+
+    return Angle(degrees * priv::pi / 180.0f);
+}
```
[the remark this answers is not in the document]
ChrisThrasher: Yeah you're probably right. I think that makes more sense.

ChrisThrasher:
@kimci86 semi-related to this PR, I think SFML needs a header with constants. Right now I can't think of a constant we'd need besides pi but it at least means removing a bunch of instances of PI all over the place. Something like `include/SFML/System/Numbers.hpp` that will contain `sf::priv::pi` among potentially other constants.

test/System/Angle.cpp
ChrisThrasher: I removed those integer overloads since all they did was cast to float before doing the same calculations. Users can easily achieve the same thing on their own without us adding these ambiguous overloads.

include/SFML/System/Angle.inl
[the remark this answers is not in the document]
ChrisThrasher: I change the implementation accordingly. Looks better now thanks.

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,426 @@
...
+/// Angle values support the usual mathematical operations:
+/// you can add or subtract two angles, multiply or divide
+/// an angle by a number, etc.
+///
+/// Usage example:
+/// \code
+/// sf::Angle a1  = sf::degrees(90.f);
+/// float radians = a1.asRadians(); // 1.5708
+///
+/// sf::Angle a2 = sf::radians(PI);
```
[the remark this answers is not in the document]
ChrisThrasher: Replaced that with a float literal so that comment will become valid code.

include/SFML/System/Angle.hpp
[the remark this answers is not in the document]
ChrisThrasher: I went ahead and added a `literals` namespace and really like how it looks to write `using namespace sf::literals` instead of the ugly `using sf::operator""_deg` that many users may not realize is possible. This also makes it easy add more literals like `sf::Time` literals or anything else and users have the same conveniently entry point.

include/SFML/System/Angle.hpp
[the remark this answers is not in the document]
ChrisThrasher: https://docs.microsoft.com/en-us/cpp/cpp/user-defined-literals-cpp?view=msvc-170#user-defined-literal-operator-signatures

Looks like using `long double` is required here since a `float` parameter doesn't match one of the standard-approved user defined literal signatures.

ChrisThrasher:
https://github.com/ChrisThrasher/SFML/commit/6e0eef46a19d6754887546bd9f690156b60e7cde

I tried to find as many redefinitions of pi and replace them with a single definition. Because we can't use C++20's `<numbers>`, this is the next best thing.

eXpl0it3r:

MarioLiebisch:
> Right now I can't think of a constant we'd need besides pi

**Tau:** τ = 2π

Not as common as pi, but still exists, in use, and related.

kimci86:
In my opinion, it would make sense to make the pi constant available wrapped in an `sf::Angle` as we are trying to make angle computations more explicit and safe. Maybe `sf::Angle::Straight`.

However, a potential problem I see is that `2.f * sf::Angle::Straight` would evaluate to a zero angle. A full angle is not representable. It makes sense but I can imagine it might surprise users.
For example, consider `something * 2.f * sf::Angle::Straight` and `2.f * sf::Angle::Straight * something`.
Anyways, it is already easy to write potentially surprising expressions such as `0.1f * 360_deg`.

If really needed, pi would be available with `sf::Angle::Straight.asRadians()`.
It is very verbose though, so `sf::pi` could stay.

In addition, I think sfml examples should be updated to use `sf::Angle` and not compute angles using primitive types. Maybe this can be a separate PR for later? I could do it if you are not eager to do it.

include/SFML/System/Angle.inl
```diff
@@ -0,0 +1,218 @@
...
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+namespace priv
+{
+    constexpr float pi = 3.141592654f;
+
+    constexpr float fmodf(float x, float y)
```
MarioLiebisch, line 29: Even if controversial, I'd move the normalization of negative values in here. `fmodf(-500)` returning `-500` is as unexpected/undefined.

kimci86:
I am starting to think that angles should not be normalized in order to have commutative multiplication (at least as much commutative as float multiplication).

MarioLiebisch:
Not 100% sure I can follow – it's associative either way?

> -5° * 18 = -90° = 270°
-5° * 2 * 9 = -10° * 9 = -90° = 270°
-5° * 9 * 2 = -45° * 2 = -90° = 270°
355° * 18 = 6390° = 270°
355° * 2 * 9 = 710° * 9 = 6390° = 270°
355° * 9 * 2 = 3195° * 2 = 6390° = 270°

kimci86:
Sorry, I should have given examples. Here:

> (2 * pi) / 3 = 0
> 2 * (pi / 3) = 2.09439510239 rad = 120 °

Surprising things might happen when multiplying with non-integer values (1/3 in this example).

MarioLiebisch:
Ah, yeah, true!

Should probably follow your typical vector example and just provide a convenient way for normalization.

This just begs the question how (in)equality should be handled. Since mathematically, 0° and 360° aren't necessarily the same, but by common definition they would be.

Bromeon (changes requested):
Thanks for this proof-of-concept! Seems like it doesn't need so many changes in the API.

We should also add comparison operators.

The whole "normalize to [0, 360[" looks good at first glance, but brings its own share of problems.

First, normalization needs to be done after every operation. Unless you do lazy normalization, which needs `mutable`, an extra bool flag (i.e. double the size) and comes with its own bunch of issues.

More importantly, it can be unintuitive, because several algeraic fundamentals are violated in the "cyclic" domain [0, 360[:
   ```cpp
   sf::Angle a = sf::degrees(300.f);
   sf::Angle b = sf::degrees(100.f);
   sf::Angle mean = (a + b) / 2.f;
   // guess what, it's not 200
   ```
   or:
   ```cpp
   sf::Angle a = sf::degrees(110.f);
   sf::Angle b = sf::degrees(130.f);
         a < b       // true
   (3 * a) < (3 * b) // false
   ```
   etc.

I'm not sure if there's a better approach, but this will certainly be a pitfall for some users. This is a problem that doesn't exist with floats (sure, the user has to do wrap-around themselves, but it typically means they're aware of how it works).

include/SFML/System/Angle.inl
```diff
@@ -0,0 +1,218 @@
...
+    {
+        while (x >= y)
+            x -= y;
```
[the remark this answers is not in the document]
ChrisThrasher: I can't say I totally understand how this implementation works but when I replace the current implementation with this, the tests still pass so it seems good!

include/SFML/System/Angle.inl
```diff
@@ -0,0 +1,218 @@
...
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle::Angle(float radians)
+    : m_radians(0.0f)
+{
+    // Normalize to [0, 2*pi)
+    while (radians < 0.0f)
+        radians += 2 * priv::pi;
```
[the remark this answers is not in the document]
ChrisThrasher: You're not wrong haha. Surely there's a better way to do this same calculation but I'm not sure what that would be.

src/SFML/Graphics/Transformable.cpp
```diff
@@ -145,9 +143,8 @@ const Transform& Transformable::getTransform() const
     // Recompute the combined transform if needed
     if (m_transformNeedUpdate)
     {
-        float angle  = -m_rotation * 3.141592654f / 180.f;
-        float cosine = std::cos(angle);
-        float sine   = std::sin(angle);
+        float cosine = std::cos(-m_rotation.asRadians());
+        float sine   = std::sin(-m_rotation.asRadians());
```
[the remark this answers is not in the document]
ChrisThrasher: Okay I'll fix that.

include/SFML/System/Angle.inl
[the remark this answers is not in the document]
Bromeon: I mean, I just showed an implementation of `fmodf()` that works for negative numbers.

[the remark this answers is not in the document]
ChrisThrasher: Oh it does? How exactly would this loop get replaced with `fmodf()`? I can't picture how that would work. Something like this?
```cpp
constexpr Angle::Angle(float radians)
    : m_radians(0.0f)
{
    // Normalize to [0, 2*pi)
    m_radians = priv::fmodf(radians, 2 * priv::pi);
    if (radians < 0.0f)
        m_radians += 2 * priv::pi;
}
```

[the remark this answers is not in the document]
Bromeon: Just this:
```cpp
    m_radians = priv::fmodf(radians, 2 * priv::pi);
```

By the way, I think this method should not be called `fmodf()` since it has different semantics and never returns negative numbers. `signedRemainer()` would  be more appropriate.

I used this snippet to run a few tests, it's probably a good idea if we add this to unit tests as well. Might even become part of our "tools" header.
```cpp
#include<iostream>
#include<cassert>

using namespace std;

float unsignedRemainder(float a, float b)
{
    assert(b > 0);
    if (a >= 0)
        return a - static_cast<int>(a / b) * b;
    else
        return a + b - static_cast<int>(a / b) * b;
}

int main()
{
    cout<< unsignedRemainder(90, 40) << endl;
    cout<< unsignedRemainder(-90, 40) << endl;
    cout<< unsignedRemainder(50, 40) << endl;
    cout<< unsignedRemainder(-50, 40) << endl;
    cout<< unsignedRemainder(30, 40) << endl;
    cout<< unsignedRemainder(-30, 40) << endl;
    cout<< unsignedRemainder(0, 40) << endl;
    
    // should assert:
    cout<< unsignedRemainder(90, -40) << endl;
    cout<< unsignedRemainder(-90, -40) << endl;
    
    return 0;
}
```

ChrisThrasher:
@Bromeon 

> We should also add comparison operators.

Your later example about unintuitive comparison operators is why I didn't include those operators. It doesn't always make sense to say what angle is bigger than another since these angles are analogous to points on the perimeter of a circle. Comparison operators might make more sense with a [-pi, pi) domain but I'm not sure if I want to use that domain.

> This is a problem that doesn't exist with floats (sure, the user has to do wrap-around themselves, but it typically means they're aware of how it works).

This is the argument for using C 😄. If users want supreme flexibility at the expensive of expressiveness then C is always available.

include/SFML/System/Angle.inl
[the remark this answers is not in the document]
ChrisThrasher: I'm reluctant to test `priv::fmodf` (or whatever it's name should be since it's no longer analogous to `std::fmodf`) because it's an implementation detail and not part of the public API. That changes if it get elevated out of the `priv::` namespace but currently that's not the case.

include/SFML/System/Angle.inl
ChrisThrasher: Replaced this with an implementation Bromeon wrote and changed the name since this is no longer the same thing as `fmodf`.

include/SFML/System/Angle.inl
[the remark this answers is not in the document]
ChrisThrasher: It works! Thanks for helping me get rid of all the loops.

include/SFML/System/Angle.inl
```diff
@@ -0,0 +1,218 @@
...
+// 2. Altered source versions must be plainly marked as such,
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+namespace priv
+{
+    constexpr float pi = 3.141592654f;
```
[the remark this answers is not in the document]
ChrisThrasher: I don't know and am fine either way. You just tell me what you want.

MarioLiebisch:
Just as another note against forced normalization: You actually might want to define an angle bigger than 360°. Imagine wanting some animation to do two full rotations, so you might want to tween from 0° to 720°. With normalization you wouldn't move at all.

Bromeon:
@Bromeon 

> Your later example about unintuitive comparison operators is why I didn't include those operators. It doesn't always make sense to say what angle is bigger than another since these angles are analogous to points on the perimeter of a circle. Comparison operators might make more sense with a [-pi, pi) domain but I'm not sure if I want to use that domain.

True, but we already have that problem anyway -- not having comparison operators just makes people write
 `a.asRadians() < b.asRadians()`, which is subject to the same issue, only less readable.

You bring up another good point: some people want a [0°, 360°[ range, others a [-180°, 180°[ one. Some don't care at all.
And Mario brought up _yet another_ reason why this might be a bad idea.

Maybe we should instead add two methods to explicitly normalize and let the user decide?
```rs
Angle Angle::clamped180() const; // [-180°, 180°[
Angle Angle::clamped360() const; // [   0°, 360°[
```

ChrisThrasher:
@MarioLiebisch 

> Just as another note against forced normalization: You actually might want to define an angle bigger than 360°. Imagine wanting some animation to do two full rotations, so you might want to tween from 0° to 720°. With normalization you wouldn't move at all.

```cpp
for (int angle = 0; angle <= 720; ++angle)
    object.setRotation(sf::degrees(angle));
```

Maybe I'm misunderstanding you but I'm not sure how this would be a problem. You can make any number of angles between 0 and 720 degrees. As far as the user is concerned, they don't even have to know that normalization is happening. The API doesn't stop you from making bigger angles. The tests do that many times to make sure the API is still correct when users do that.

@Bromeon 

> True, but we already have that problem anyway -- not having comparison operators just makes people write
a.asRadians() < b.asRadians(), which is subject to the same issue, only less readable.

If a user deliberately breaks out of the type safety of the API then I'm not sure what to tell them. It's the same as writing `std::chrono::milliseconds(10).count() > std::chrono::seconds(1).count()`. `.count()` is useful but dangerous. Same argument goes for `asRadians` and `asDegrees`. 

> You bring up another good point: some people want a [0°, 360°[ range, others a [-180°, 180°[ one. Some don't care at all.
And Mario brought up yet another reason why this might be a bad idea.

https://github.com/SFML/SFML/issues/1747#issuecomment-764708732

Earlier this year you how SFML tries to be simple and isn't trying to be Boost. Giving users access to control angle ranges (even if only via const member functions) is working against the idea of SFML being simple. I think the first thing we can agree upon is that we shouldn't let users pick the range of angles we use. Once we offer that control we can never take it back and it goes against the ethos of keeping things simple.

As for whether or not normalization should occur, it's already happening. See the definition of `sf::Transformable::rotate`. That normalization logic got removed in this PR because it's now encapsulated in `sf::Angle`. Imagine a user repeatedly calling `sf::Transformable::rotate(sf::degrees(10));` to show an object spinning forever. Internally that `float` would grow unbounded. Because SFML chooses to not provide our own `operator<<` overloads, a user would then have to call `asDegrees` to print the angle to the console just to see something confusing like `444255.0011386719` which most would find surprising and not particularly helpful. Run that code long enough and they'll just see `inf` or `NaN`.

I hope we can come to agreement what range of normalization to use, but I think we should certainly have _some_ degree of normalization to prevent users (including elsewhere in SFML) from having think hard about how to avoid these pitfalls.

kimci86:
As I understand, the first purpose of `sf::Angle` is to not be confused by units. This is achieved.

Then I see two concerns:
1. Taking care of precision by normalizing values, avoiding the angle going towards infinity.
2. Being usable intuitively in any arithmetic epxression.

Even though 1. is nice, 2. seems more important. Among what have already been pointed out, it allows uses like:
```c++
for(auto angle = 0_deg; angle < 720_deg; angle += 10_deg)
    object.setRotation(angle);
```

If we agree going that way, users can manually take care of precision with the modulo operator.

Bromeon:
Thanks for the elaborate answer!

---

> Maybe I'm misunderstanding you but I'm not sure how this would be a problem. You can make any number of angles between 0 and 720 degrees. As far as the user is concerned, they don't even have to know that normalization is happening. The API doesn't stop you from making bigger angles. 

No, the user needs to know it.
```cpp
sf::Angle a = sf::degrees(390);
a.asDegrees() // NOT 390
```
Clamping is not an implementation detail. Without users knowing the [0, 360[ semantics, they will write bugs.

---

> If a user deliberately breaks out of the type safety of the API then I'm not sure what to tell them. It's the same as writing `std::chrono::milliseconds(10).count() > std::chrono::seconds(1).count()`. `.count()` is useful but dangerous. Same argument goes for `asRadians` and `asDegrees`.

By not providing an ordering relation, we _force_ the user to break out of the type safety. Of course we can say "that's your problem", but in reality there are valid use cases for comparing angles and the currently proposed API does simply not enable them.

---

> [#1747 (comment)](https://github.com/SFML/SFML/issues/1747#issuecomment-764708732)
> 
> Earlier this year you said how SFML tries to be simple and isn't trying to be Boost. Giving users access to control angle ranges (even if only via const member functions) is working against the idea of SFML being simple. I think the first thing we can agree upon is that we shouldn't let users pick the range of angles we use. Once we offer that control we can never take it back and it goes against the ethos of keeping things simple.

There are two choices we can make in the API:
1. Everything is implicitly clamped to [0, 360]. This means less work for the user, but can lead to surprises.
2. `Angle` is just wrapping a float. Conversions to a certain range are explicit, but thus also less error-prone.

I could agree that 1. were simpler if the clamping logic would "just work". But in fact this logic is a leaky abstraction, so the user _must_ understand it in order to use `Angle` correctly. We make it _appear_ simple, but the complexity is still there, just less obvious.

Several valid use cases, like bisecting an angle, become suddenly quite complex with this approach.

---

> As for whether or not normalization should occur, it's already happening. See the definition of `sf::Transformable::rotate`. That normalization logic got removed in this PR because it's now encapsulated in `sf::Angle`. Imagine a user repeatedly calling `sf::Transformable::rotate(sf::degrees(10));` to show an object spinning forever. Internally that `float` would grow unbounded. Because SFML chooses to not provide our own `operator<<` overloads, a user would then have to call `asDegrees` to print the angle to the console just to see something confusing like `444255.0011386719` which most would find surprising and not particularly helpful. Run that code long enough and they'll just see `inf` or `NaN`.

I don't see the problem here. If the code wants the rotation to be clamped, this is exactly where `clampedTo360()` comes into play.

Where I am not sure, is whether we _always_ want that.

---

> I hope we can come to agreement what range of normalization to use, but I think we should certainly have _some_ degree of normalization to prevent users (including elsewhere in SFML) from having think hard about how to avoid these pitfalls.

Unfortunately I don't think it's possible to avoid all kinds of pitfalls. Ultimately, we need to ask ourselves what the task of `sf::Angle` is. I understood it as adding type safety against `float`, as well as enabling equivalent use of degrees and radians. You additionally understand it as limiting the domain (range) of allowed values. None is more correct, they just have different trade-offs.

When looking at existing gamedev libraries and linear algebra APIs, one thing I noticed is that almost none have a distinct angle type. Those who do (e.g. Rust [euclid](https://docs.rs/euclid/0.22.6/euclid/struct.Angle.html) or [Ogre3D](https://www.ogre3d.org/docs/api/1.9/class_ogre_1_1_degree.html)) seem to do nothing more than wrap a float, sometimes even publicly. Maybe those issues we're encountering now were also encountered by them?

ChrisThrasher:
@Bromeon Safe to say that normalization won't be happening then. With that removed, can you elaborate on what extra member functions you'd want so I can implement and test those real quick?

Bromeon:
@ChrisThrasher For now I'd start without implicit clamping. We still have the option to revise the semantics until the release of SFML 3.0.0; overall we need some practical experience with `sf::Angle`, as it's an entirely new thing for both SFML developers and users.

I would add these two methods, plus the relational operators `<` `<=` `>` `>=`.
```cpp
Angle Angle::clampedTo180() const; // [-180°, 180°[
Angle Angle::clampedTo360() const; // [   0°, 360°[
```
That plus the available arithmetic operators probably cover most of the use cases.

ChrisThrasher:
@Bromeon I made those changes including tests for those two extra member functions.

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,458 @@
...
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr bool operator >=(Angle left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of unary - operator to negate an angle value
+///
+/// \param right Right operand (an angle)
+///
+/// \return Opposite of the angle value
```
Bromeon, line 225: Maybe to not only mention the obvious, we could state that `-angle` represents a rotation in the opposite direction of `angle`? (I know most people are aware of this, but at least it's not implied by the operator itself).

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,458 @@
...
+/// \param right Right operand (an angle)
+///
+/// \return \a left divided by \a right
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr float operator /(Angle left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of binary % operator to compute remainder of an angle value
```
Bromeon, line 352: "remainder" is technically not precise, due to how it handles negative values; I think "modulo" would fit better here.
See: https://stackoverflow.com/q/13683563

We should probably mention the semantics -- people reading this likely know what % does, but they might be interested in the handling of negative numbers. Maybe something like:
```cpp
/// Examples:
/// sf::degrees(90.f) % sf::degrees(40.f)  // approximately 10
/// sf::degrees(-90.f) % sf::degrees(40.f) // approximately 30 (not -10)
```

In addition, having right-hand side negative currently fails with an assertion. We should either document this, or define the behavior for negative `right` operands.

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,458 @@
...
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief User defined literal for angles in radians
+///
+/// \param angle Angle in radians
+///
+/// \return \a Angle
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr Angle operator "" _rad(unsigned long long angle);
```
Bromeon, line 419: Maybe we should for each mention a short example usage, like
```cpp
sf::Angle angle = 22.5_deg;
```

This would also alleviate potential confusion with the parameter types `long double` and `unsigned long long`.

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,458 @@
...
+///
+/// sf::Angle encapsulates an angle value in a flexible way.
+/// It allows to define an angle value either as a number of
+/// degrees or radians. It also works the other way round:
+/// you can read an angle value as either a number of
+/// degrees or radians.
+///
+/// By using such a flexible interface, the API doesn't
+/// impose any fixed type or resolution for angle values,
+/// and let the user choose their own favorite representation.
```
Bromeon, line 443: "resolution" is not really true, we do limit the implementation to 32-bit floats.

Maybe we should mention "...the API doesn't impose any unit for angle values..."?

include/SFML/System/Angle.inl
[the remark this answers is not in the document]
Bromeon: We also use it here 🙂 although locally

https://github.com/SFML/SFML/blob/871c021a75e3d728e96086f04f7ffdf61ebd83cb/src/SFML/Graphics/CircleShape.cpp#L73-L82

I'm fine to leave this as-is for now. Should we ever make it public, it would become `PascalCase`.

include/SFML/System/Angle.inl
```diff
@@ -0,0 +1,258 @@
...
+    constexpr float two_pi = 2 * pi;
```
Bromeon, line 28: There's a name for this: `tau`

Tau has become the de-facto name in the programming world. From [Wikipedia](https://en.wikipedia.org/wiki/Turn_(angle)#Tau_proposals):

> In 2012, the educational website Khan Academy began accepting answers expressed in terms of τ.
In June 2017, for release 3.6, the Python programming language adopted the name tau to represent the number of radians in a turn.
The τ-functionality is made available in the Google calculator and in several programming languages like Python, Raku, Processing, Nim, and Rust.
It has also been used in at least one mathematical research article,[41] authored by the τ-promoter Peter Harremoës.
In 2020, for release 5.0, Tau was added to .NET Core (which is being rebranded as ".NET" for the 5.0 release).

include/SFML/System/Angle.inl
```diff
@@ -0,0 +1,258 @@
...
+constexpr bool operator <=(Angle left, Angle right)
+{
+    return left.asRadians() <= right.asRadians();
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr bool operator >=(Angle left, Angle right)
+{
+    return left.asRadians() <= right.asRadians();
```
Bromeon, line 134: Wrong operator.

Bromeon (commented):
Thanks a lot for the great effort so far! 
I now had a more detailed look at the code 🔬

kimci86:
I find the name "clamped" confusing because it reminds me `std::clamp` which has a different meaning. Sadly, I do not known a better alternative.

Bromeon:
> I find the name "clamped" confusing because it reminds me `std::clamp` which has a different meaning. Sadly, I do not known a better alternative.

Very good point, "clamp" actually means "limit to either edge".

Matlab uses [`wrapTo180`](https://ch.mathworks.com/help/map/ref/wrapto180.html) and [`wrapTo360`](https://ch.mathworks.com/help/map/ref/wrapto360.html), maybe we can adopt this?

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,458 @@
...
+/// \relates Angle
+/// \brief Overload of < operator to compare two angle values
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (an angle)
+///
+/// \return True if \a left is lesser than \a right
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr bool operator <(Angle left, Angle right);
```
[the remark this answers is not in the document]
ChrisThrasher: Because these operators don't modify any inputs and modify no global state, the only observable behavior is in the form of its return value so it must be used. This is the case with `sf::Time`s comparison operators as well.

include/SFML/System/Angle.hpp
ChrisThrasher: Oops, I didn't write tests for negative values. I added that test case for your negative example. 

I actually didn't add the assertions you originally wrote in `positiveRemainder`. I assume SFML is fine having assertions in code that could potentially be triggered by user code? I wasn't sure what the policy was on that.

include/SFML/System/Angle.hpp
ChrisThrasher: That's a better way to put it. For it's worth, `sf::Time` also uses similar language which also doesn't correctly represent what the code is doing. I doubt anybody is excited to comb through documentation comments to find these kinds of inaccuracies but it's probably another worthwhile task to do in the 3.x release series even if it doesn't land for 3.0.

include/SFML/System/Angle.hpp
Bromeon: Error handling is also one of the points where we don't have a clear convention yet.

In any case, `assert` is always better than buggy/untested results. But we can also define the behavior for negative divisors -- but we would need to think a bit about use cases and how one would typically expect things to work.

include/SFML/System/Angle.inl
ChrisThrasher: Ah yes I forgot about tau. I'll fix that.

include/SFML/System/Angle.inl
ChrisThrasher: Haha, I copy pasted the wrong implementation and wrong tests but they both lined up so I never noticed. Thanks for catching that silly mistake.

include/SFML/System/Angle.inl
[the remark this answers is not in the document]
ChrisThrasher: I have another local commit that promotes this constant into the public interface of `sf::Angle` and then replaces all redundant definitions with that one. Let me know if you're interested in this:

https://github.com/ChrisThrasher/SFML/commit/4110f22de49438bd2c83ed37d9b3e21ffd5d842b

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,458 @@
...
+    // Member data
+    ////////////////////////////////////////////////////////////
+    float m_radians; //!< Angle value stored as radians
+};
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Construct an angle value from a number of degrees
+///
+/// \param amount Number of degrees
```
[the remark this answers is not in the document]
ChrisThrasher: Fixed, thanks.

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,458 @@
...
+[[nodiscard]] constexpr bool operator >(Angle left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of <= operator to compare two angle values
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (an angle)
+///
+/// \return True if \a left is lesser or equal than \a right
```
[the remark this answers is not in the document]
ChrisThrasher: Fixed, thanks.

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,458 @@
...
+[[nodiscard]] constexpr bool operator !=(Angle left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of < operator to compare two angle values
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (an angle)
+///
+/// \return True if \a left is lesser than \a right
```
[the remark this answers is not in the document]
ChrisThrasher: Fixed thanks

include/SFML/System/Angle.hpp
ChrisThrasher: I agree. I replaced the `assert` and removed the term "remainder" from the docs.

ChrisThrasher:
I appreciate the feedback! I think I've addressed everything but let me know if I missed anything. Can we run the full pipeline now? I bet the 32 bit Mingw build will have some test failures due to its slightly different float point number handling. I probably just need to add `Approx` in a few more places in the tests.

include/SFML/System/Angle.inl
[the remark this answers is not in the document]
Bromeon: I'd say we discuss that separately (maybe in an issue before a PR). An option is also that we don't need numeric π, but could add something like `Angle::Pi` or `Angle::HalfCircle` or so.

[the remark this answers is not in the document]
Bromeon: Btw, if you mark conversations as resolved but expect answers, there's a high chance I will miss them and won't respond 😉

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,465 @@
...
+    ///
+    /// \return Angle in radians
+    ///
+    /// \see asDegree
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] constexpr float asRadians() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Return the angle clamped to the range [-180°, 180°)
```
kimci86, line 74: "clamped" on this line a few times below should be replaced with "wrapped".

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,465 @@
...
+/// \return \a left modulo \a right
+///
+////////////////////////////////////////////////////////////
+constexpr Angle& operator %=(Angle& left, Angle right);
+
+inline namespace literals
+{
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
```
kimci86, line 385: Doxygen needs `\relates sf::Angle` for the user defined literals as they are in a different namespace.

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,465 @@
...
+/// an angle by a number, compare two angles, etc.
+///
+/// Usage example:
+/// \code
+/// sf::Angle a1  = sf::degrees(90.0f);
+/// float radians = a1.asRadians(); // 1.5708f
+///
+/// sf::Angle a2 = sf::radians(3.141592654f);
+/// float degrees = a2.asDegree(); // 180.0f
+/// \endcode
```
kimci86, line 463: IMO, it would be nice to have an example of using user defined literals here, showing `using namespace sf::literals`.

include/SFML/System/Angle.hpp
ChrisThrasher: Good catch. I fixed that.

include/SFML/System/Angle.hpp
ChrisThrasher: Thanks. Fixed.

include/SFML/System/Angle.hpp
ChrisThrasher: Good idea. Do you think this is sufficient?
```cpp
/// using namespace sf::literals;
/// sf::Angle a3 = 10_deg; // 10 degrees
/// sf::Angle a4 = 1_rad;  // 1 radian
```

include/SFML/System/Angle.inl
```diff
@@ -0,0 +1,259 @@
...
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr Angle& operator %=(Angle& left, Angle right)
+{
+    return left = left % right;
+}
+
+inline namespace literals
```
Bromeon, line 229: Two questions:
1. Why `inline`? I think typical usage will be `using namespace sf::literals;`, no?
2. This naming convention is inconsistent with `sf::Glsl`. Maybe let's keep `PascalCase` for now, and discuss change of naming convention in a separate issue (if that's desired).

include/SFML/System/Angle.hpp
kimci86: Nice. Maybe have `1.42_rad` or something to illustrate both integer and floating literals.

include/SFML/System/Angle.inl
ChrisThrasher: `inline` was @kimci86's idea. Apparently it allows us to use these literals within the `sf::` namespace without having to add that `using namespace sf::literals` declaration but still requires that users outside the `sf::` namespace do use that.

Are you saying that `literals::` should be `Literals::`? That's no problem to change.

include/SFML/System/Angle.hpp
ChrisThrasher: Done. Now it's this:
```cpp
/// using namespace sf::Literals;
/// sf::Angle a3 = 10_deg;   // 10 degrees
/// sf::Angle a4 = 1.5_deg;  // 1.5 degrees
/// sf::Angle a5 = 1_rad;    // 1 radian
/// sf::Angle a6 = 3.14_rad; // 3.14 radians
```

include/SFML/System/Angle.inl
ChrisThrasher: I went ahead and changed `literals::` to `Literals::`.

kimci86: I was inspired by standard library when suggesting `inline`.

Bromeon: Are the symbols only available for SFML itself? I don't know if we want `using namespace sf;` to also spill them out.

Yes, `Literals::` for now. I'm also torn if it's better or not, but we should find a rule that holds for all namespaces 🙂
The thing is that there could be namespaces used like types, e.g. [`sf::Listener`](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1Listener.php) might as well be a namespace.

ChrisThrasher: Turns out you're right. All it takes to get these UDLs is to add `using namespace sf`. I wasn't aware it worked that way but it makes more sense now. That's why we don't need that using declaration when inside the `sf::` namespace.

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,471 @@
...
+    ///
+    /// \see wrappedTo180
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] constexpr Angle wrappedTo360() const;
+
+    ////////////////////////////////////////////////////////////
+    // Static member data
+    ////////////////////////////////////////////////////////////
+    SFML_SYSTEM_API static const Angle Zero; //!< Predefined 0 degree angle value
```
vittorioromeo, line 96:
```suggestion
    ////////////////////////////////////////////////////////////
    // Static member data
    ////////////////////////////////////////////////////////////
    SFML_SYSTEM_API static const Angle Zero; //!< Predefined 0 degree angle value
```

I don't really see the point of this. It has the usual issue with `constexpr` support, and is it really any better than `sf::degrees(0)` or `sf::Angle{}`?

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,471 @@
...
+#include <SFML/System/Export.hpp>
+#include <cassert>
+
+
+namespace sf
+{
+////////////////////////////////////////////////////////////
+/// \brief Represents an angle value
+///
+////////////////////////////////////////////////////////////
```
vittorioromeo, line 40: We should mention that the internal representation is radians, and that it should be preferred to degrees whenever possible. Yes, it's a "leaky" detail, but it helps people avoid unnecessary conversions.

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,471 @@
...
+///
+/// \param left  Left operand (an angle)
+/// \param right Right operand (an angle)
+///
+/// \return \a left modulo \a right
+///
+////////////////////////////////////////////////////////////
+constexpr Angle& operator %=(Angle& left, Angle right);
+
+inline namespace Literals
```
vittorioromeo, line 381: Nice, I like this!

include/SFML/System/Angle.hpp
ChrisThrasher: That's fair. This follows `sf::Time`'s established pattern of a named constant for zero. I think it's nice to have but it's not that important.

include/SFML/System/Angle.hpp
ChrisThrasher: I'll add a note about that. It's pretty unlikely we ever change the underlying representation so I think this is a reasonable thing to "leak" such a detail. It does call into question whether degrees should be the underlying representation since doing so would likely prevent more lossy floating point operations since most people would rather reason about angles in terms of degrees than radians. We could even do something a little weird and represent angles as millidegrees (or microdegrees?) to avoid the losses of doing many floating point calculations. We don't exactly need the range of `float`. Who really needs angles greater than `std::numeric_limits<sf::Int32>::max()` anyways?

include/SFML/System/Angle.hpp
ChrisThrasher: Great idea to add UDLs! This was my first time implementing them and it was great to learn about.

include/SFML/System/Angle.hpp
Bromeon: I agree with @ChrisThrasher here, and there might actually be more useful constants in the future, such as `Pi` and `Tau` (see comment [here](https://github.com/SFML/SFML/pull/1969#discussion_r785478790)).

The `constexpr` problem is something we need to find a solution for, or at least workaround -- but I don't think it should influence whether we want to have constants in our API or not. Otherwise we'd need to remove `sf::Time::Zero`, `sf::Color::*` and many others.

include/SFML/System/Angle.hpp
Bromeon: >  It does call into question whether degrees should be the underlying representation since doing so would likely prevent more lossy floating point operations since most people would rather reason about angles in terms of degrees than radians. 

That's a good point actually -- and maybe it's a reason why we should _not_ document how it's implemented.

If we document this, then people might specifically optimize for it and make their code more ugly than necessary by using `sf::radians(1.57075f)`. Also, once documented, people have a right to rely on this to remain true throughout SFML 3.

People who are really interested in the internals will just look at the code -- and they can do that, but we give no guarantee. By documenting it, we make it part of the API and encourage a certain way of using `sf::Angle`, which goes against its primary goal.

include/SFML/System/Angle.inl
Bromeon: Then I would say we make it a regular `namespace` instead of `inline namespace`. 

Users can still achieve the same, but usage of user-defined literals is explicit and not accidental. If we as SFML developers want to use them, we can still do so with selective `using` statements inside .cpp files or function bodies.

I would avoid that symbols appear in two different namespaces.

include/SFML/System/Angle.inl
```diff
@@ -0,0 +1,259 @@
...
+constexpr Angle::Angle() :
+m_radians(0.0f)
+{
+}
+
+
+////////////////////////////////////////////////////////////
+constexpr float Angle::asDegrees() const
+{
+    return m_radians * 180 / priv::pi;
```
Bromeon, line 51: We might want to write `m_radians * (180.f / priv::pi)`, ensuring that we use a constant factor.
Avoiding the divison at runtime also mitigates concerns about conversions and loss of precision.

Same is of course true for `sf::degrees()`.

include/SFML/System/Angle.inl
kimci86: I interpreted the standard library pattern as a good practice, but we are free to be a little more strict about namespaces. As you prefer.

include/SFML/System/Angle.inl
ChrisThrasher: Good catch.

include/SFML/System/Angle.inl
ChrisThrasher: I removed `inline` from the `Literals` namespace.

include/SFML/System/Angle.hpp
ChrisThrasher: I removed the comment about radians being preferred so we can change our mind in the future. I still want to experiment with something like millidegrees or micro degrees to see if that makes a noticeable improvement in some way.

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

include/SFML/System/Angle.inl
```diff
@@ -0,0 +1,259 @@
...
+    constexpr float tau = 2 * pi;
+
+    constexpr float positiveRemainder(float a, float b)
+    {
+        assert(b > 0.0f);
+        if (a >= 0.f)
+            return a - static_cast<float>(static_cast<int>(a / b)) * b;
+        else
+            return a + b - static_cast<float>(static_cast<int>(a / b)) * b;
+    }
```
Marioalexsan, line 37: This method doesn't seem to return a remainder in the interval 0 <= r < |b| for negative numbers that divide exactly by b.
This is evident from one of the remainder tests: `CHECK((sf::degrees(-90.0f) % sf::degrees(30.0f)).asDegrees() == Approx(30.0f))`.

```suggestion
    constexpr float positiveRemainder(float a, float b)
    {
        assert(b > 0.0f);
        const float val = a - static_cast<float>(static_cast<int>(a / b)) * b;
        
        if (val >= 0.f)
            return val;
        else
            return val + b;
    }
```

ChrisThrasher: Thanks this works! Once I swapped implementation the test case you mentioned started failing. I have this fixed locally and will update the remote branch shortly.

ChrisThrasher:
I reimplemented `sf::Angle` in terms of degrees instead of radians. Because SFML uses degrees and because degrees are generally easier for humans to reason about than degrees, it's a safe assumption to make that most users will construct an `sf::Angle` via `sf::degrees` and will print its value to the console via `sf::Angle::asDegrees`. By encoding angles as degrees, we minimize the number of floating point operators that occur which maximizes accuracy and minimizes the work that the computer has to do.

eXpl0it3r:

Bromeon (approved):
For me, this looks good! Thanks a lot for all the effort 🙂

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,471 @@
...
+private:
+
+    ////////////////////////////////////////////////////////////
+    // Member data
+    ////////////////////////////////////////////////////////////
+    float m_degrees; //!< Angle value stored as degrees
+};
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
```
kimci86, line 123: A small documentation remark: `\relates` is not needed for friend functions.
Removing `\relates` and adding the parameter name `angle` in the friend declaration makes doxygen happy.
Same remark for `degrees` function.

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,471 @@
...
+[[nodiscard]] constexpr float operator /(Angle left, Angle right);
+
+////////////////////////////////////////////////////////////
+/// \relates Angle
+/// \brief Overload of binary % operator to compute modulo of an angle value
+///        Right hand angle must be greater than zero
+///
+/// Examples:
+/// sf::degrees(90) % sf::degrees(40)  // 10 degrees
+/// sf::degrees(-90) % sf::degrees(40) // 30 degrees (not -10)
```
kimci86, line 359:
```suggestion
/// \code
/// sf::degrees(90) % sf::degrees(40)  // 10 degrees
/// sf::degrees(-90) % sf::degrees(40) // 30 degrees (not -10)
/// \endcode
```

include/SFML/System/Angle.hpp
ChrisThrasher: Good catch. Removing that `\relates` line actually made the generated docs more expressive. I also added the parameter name to the declaration so it would show up more readily in the docs.

include/SFML/System/Angle.hpp
ChrisThrasher: Thanks! I just pushed a fix for that.

eXpl0it3r:

Bromeon:
This should be ready now; or anything left to do? 🚀

ChrisThrasher:
No, I don't have any more work to do on this PR.

eXpl0it3r:

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,473 @@
...
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Return the angle wrapped to the range [-180°, 180°)
+    ///
+    /// \return Angle wrapped to [-180°, 180°)
+    ///
+    /// \see wrappedTo360
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] constexpr Angle wrappedTo180() const;
```
eXpl0it3r, line 81: What are the use cases for this functionality?

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,473 @@
...
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Return the angle wrapped to the range [0°, 360°)
+    ///
+    /// \return Angle wrapped to [0°, 360°)
```
eXpl0it3r, line 86: I think, we should explain what the `[x, y)` range means

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,473 @@
...
+////////////////////////////////////////////////////////////
+/// \relates sf::Angle
+/// \brief User defined literal for angles in degrees, e.g. 90_deg
+///
+/// \param angle Angle in degrees
+///
+/// \return \a Angle
+///
+////////////////////////////////////////////////////////////
+[[nodiscard]] constexpr Angle operator "" _deg(unsigned long long angle);
```
eXpl0it3r, line 406: Wouldn't this technically need to be `unsigned long long int`? 😄

src/SFML/System/Angle.cpp
```diff
@@ -0,0 +1,38 @@
...
+////////////////////////////////////////////////////////////
+#include <SFML/System/Angle.hpp>
+
+
+namespace sf
+{
+////////////////////////////////////////////////////////////
+// Static member data
+////////////////////////////////////////////////////////////
+const Angle Angle::Zero;
```
eXpl0it3r, line 36: Shouldn't this be part of the class itself?
See the current implementation of `sf::Time::Zero`: https://github.com/SFML/SFML/blob/master/include/SFML/System/Time.hpp#L85
And the potential constexpr change for color: https://github.com/SFML/SFML/pull/1937#issuecomment-1022991504

eXpl0it3r (changes requested):
I think, there are some smaller changes needed

include/SFML/System/Angle.hpp
ChrisThrasher: No because the standard requires that integral user defined literals use `unsigned long long`. Negative numbers use `operator-` to negate the value as necessary. I believe I wrote test cases for example that situation.

include/SFML/System/Angle.hpp
ChrisThrasher: Should we add a line below this that reads something like this?
```
Inclusive of 0, exclusive of 360
```

include/SFML/System/Angle.hpp
ChrisThrasher: Bromeon recommended we add this. https://github.com/SFML/SFML/pull/1969#issuecomment-1013732277

src/SFML/System/Angle.cpp
ChrisThrasher: It is a member of the class, hence the `Angle::` namespace. See https://github.com/ChrisThrasher/SFML/blob/sf-angle/include/SFML/System/Angle.hpp#L96

include/SFML/System/Angle.hpp
Bromeon: Both are the same.

We can even use [`long int unsigned long`](https://www.online-cpp.com/WFLmZv13Tx) 🤦‍♂️

include/SFML/System/Angle.hpp
Bromeon: `-180° .. 180°` is a range that represents a coordinate system where 0 is the middle, and positive/negative angles cancel each other out. It's often more intuitive, because algebra behaves close to the domain of real numbers.

Examples where it helps:
* `angle > 0` and `angle < 0` to see if an object (e.g. game entity) turns left/right
* `abs(angle)` gives you the deviation from "forward" (0)
* `-angle` is the same deviation in the opposite direction

include/SFML/System/Angle.hpp
Bromeon: I think only one of the two should be adjusted, the `\brief` could maybe stay.

Maybe we can directly use the relational operators?
```cpp
    /// \brief Return the angle wrapped to the range [0°, 360°)
    ///
    /// \return Angle >= 0° and < 360°
```

eXpl0it3r: I'd have gone the other way.
The return description can be rather crisp, while the function description should try and clarify in more details what stuff means and how they can/should be used etc.

include/SFML/System/Angle.hpp
ChrisThrasher: Oops sorry I misread what you said, @eXpl0it3r. I can add `int` at the end if that better matches SFML style. I don't believe it technically needs to be `unsigned long long int` though.

include/SFML/System/Angle.hpp
ChrisThrasher: I think it makes sense to have `0° <= angle < 360°` somewhere.

include/SFML/System/Angle.hpp
eXpl0it3r: I just mentioned this, because I've seen that type listed on both these pages:

- https://en.cppreference.com/w/cpp/language/user_literal
- https://docs.microsoft.com/en-us/cpp/cpp/user-defined-literals-cpp?view=msvc-170#user-defined-literal-operator-signatures

In the end it shouldn't really matter, was merely wonder 😅

include/SFML/System/Angle.hpp
Bromeon: > The return description can be rather crisp, while the function description should try and clarify in more details what stuff means and how they can/should be used etc.

Doesn't `\brief` literally mean "short/concise"? I think it's important that this one is relatively fast to read, as it will be listed among other function declarations.

But I'm totally fine with something like
```cpp
    /// \brief Wrap to a range, so that 0° <= angle < 360°
    ///
    /// \return Angle in [0°, 360°), which represents a rotation equivalent to *this.
```

eXpl0it3r: One could also add it in between the two, which then is a normal function description. The `\brief` is indeed listed in the function list overview.

image

I think both brief and return text can use the same notation and then if someone doesn't understand the notation, they should be able to find more details in the description.

include/SFML/System/Angle.hpp
kimci86: When cross product and maybe vector reflection will be added, I think using this `wrappedTo180` will not be so useful.

include/SFML/System/Angle.hpp
```diff
@@ -0,0 +1,473 @@
...
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Return the angle wrapped to the range [0°, 360°)
+    ///
+    /// \return Angle wrapped to [0°, 360°)
+    ///
+    /// \see wrappedTo180
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] constexpr Angle wrappedTo360() const;
```
eXpl0it3r, line 91: Actually just realized that the naming is also not very good, when you consider that Angle is now a unit agnostic class, so assuming 360 to mean degree and encode it on a function call, kind of works against this (why no `wrappedToTau`? 😉).
As such I'd recommend picking someone other naming like `wrappedToFullRotation` or similar.

include/SFML/System/Angle.hpp
eXpl0it3r: But is there any evidence that people actually want this? What about the ones that want some other section of the circle wrapped (e.g. -90 / +90)?

Do you have any concrete examples where you implemented this yourself or wished for it to exist?

eXpl0it3r: I guess one thing I failed to really consider that those two are more or less the only ways to "wrap" a circle (symmetrically).
You either go around once, or you go "half" and "half", as such I think it starts to make more sense and there's also enough math that deals with π and -π vs 2π I can start to see different cases 🤔 

As mentioned on the other function, I'd however rethink the naming of it.

include/SFML/System/Angle.hpp
Bromeon: Yes, but we should also avoid duplicating information, it just wastes the reader's time. (I know we're doing it everywhere else, and still think it's bad practice).

For me,
```cpp
\return Angle in [0°, 360°), which represents a rotation equivalent to *this.
```
seems to be clear enough -- but if we want to elaborate, we could also do it in `\return` directly. The middle section is not strictly necessary.

include/SFML/System/Angle.hpp
Bromeon: Yeah, that was also one of my concerns. The thing is, both are full rotations, just with different start/end points. I was also thinking about `wrappedToTau()` and `wrappedToPi()` (where tau = 2*pi), but this is even less known and suffers from the same issue, just in radians.

Another idea I had was to use "signed/unsigned". E.g. `wrapSigned()` and `wrapUnsigned()`.

From a readability point of view, "un" is quite a small difference (in more complex expressions), so having uppercase "S" and "U" helps quickly differentiate the two.

eXpl0it3r: But I'd have no clue what `wrapSigned()` means from an API perspective and the -180, 180 range would contain both signed and unsigned values.

Another idea I had was focusing more on the unit circle, so wrap FullCircle or HalfCircle, but that might be rather confusing as well.

I guess 180 vs 360 is the most intuitive way, but we'd then have to state that this is of course not limited to just degrees (which ofc doesn't even make sense since `sf::Angle` is unit agnostic...).

Bromeon: > But I'd have no clue what wrapSigned() means from an API perspective 

I think in general this concept is not well-known enough that we could make it 100% obvious in 1-2 words. There is not even a name for either of the two conventions. So maybe we should focus our efforts more on recognizability (once you know the function) than self-explanatory-ness. Reading the documentation once and understanding it from then on is acceptable in my opinion.

> and the -180, 180 range would contain both signed and unsigned values.

That's exactly what "signed" means, it has a sign 🙂 you probably meant "positive and negative".

The reason why "signed/unsigned" are quite elegant, is that their ranges behave _precisely_ like those of signed/unsigned integers (just with float):

|| signed | unsigned |
|---|---|---|
| char | [-128, 128( | [0, 256( |
| Angle| [-180°, 180°( | [0°, 360°( |

Bromeon: Maybe that discussion we're having about naming rationale and analogy to signed/unsigned could indeed fit the detailed section of the doc (between `\brief` and `\return`).

```cpp
    /// \brief Wrap to a range, so that -180° <= angle < 180°
    /// 
    /// Similar to a modulo operation, this returns a copy of the angle
    /// constrained to the range [-180°, 180°) == [-Pi, Pi).
    /// The resulting angle represents a rotation which is equivalent to *this.
    ///
    /// The name "signed" originates from the similarity to signed integers:
    /// signed char for example has range [-128, 128), just with discrete steps.
    /// In contrast, unsigned char has range [0, 256).
    ///
    /// \return Signed angle, wrapped to [-180°, 180°)
    /// \see wrapUnsigned
    ///
    [[nodiscard]] constexpr Angle wrapSigned() const;


    /// \brief Wrap to a range, so that 0° <= angle < 360°
    /// 
    /// Similar to a modulo operation, this returns a copy of the angle
    /// constrained to the range [0°, 360°) == [0, Tau) == [0, 2*Pi).
    /// The resulting angle represents a rotation which is equivalent to *this.
    ///
    /// The name "unsigned" originates from the similarity to unsigned integers:
    /// unsigned char for example has range [0, 256), just with discrete steps.
    /// In contrast, signed char has range [-128, 128).
    ///
    /// \return Unsigned angle, wrapped to [0°, 360°)
    /// \see wrapSigned
    ///
    [[nodiscard]] constexpr Angle wrapUnsigned() const;
```

include/SFML/System/Angle.hpp
ChrisThrasher: I’ll change that to `unsigned long long int` to matcher better that documentation and match how SFML uses `unsigned int` everywhere instead of `unsigned.`

include/SFML/System/Angle.hpp
ChrisThrasher: > The reason why "signed/unsigned" are quite elegant, is that their ranges behave precisely like those of signed/unsigned integers, just with floating point types instead of integers:

I think this is the best argument for a specific name I've heard so far. I like how it borrows terms already familiar to C++ programmers and accurately replicated the behavior of signed and unsigned integers. Quite elegant :)

ChrisThrasher:
I just pushed a new version of this commit which changes the naming of `wrappedTo180` and `wrappedTo360` to `wrapSigned` and `wrapUnsigned`. I mostly just copy-pasted what @Bromeon wrote but made a few small fixes to the Doxygen code to fix table formatting. This change should address all 3 outstanding unresolved issues which are all related to the naming and documentation of these functions.

eXpl0it3r:

ChrisThrasher:
I found some more uses of `sf::Angle` mostly in the example programs.

eXpl0it3r:

eXpl0it3r (approved):

include/SFML/System/Angle.hpp
ChrisThrasher: @Bromeon @eXpl0it3r Now that the two of you have approved this PR, can I resolve this last remaining issue?

examples/X11/X11.cpp
```diff
@@ -47,8 +47,7 @@
     // Setup a perspective projection
     glMatrixMode(GL_PROJECTION);
     glLoadIdentity();
-    static const float pi = 3.141592654f;
-    float extent = std::tan(90.0f * pi / 360.0f);
+    float extent = std::tan(sf::degrees(90).asRadians());
```
Marioalexsan, line 50: I think this should be `float extent = std::tan(sf::degrees(45).asRadians());` instead.

src/SFML/Graphics/CircleShape.cpp
```diff
@@ -72,11 +72,10 @@ std::size_t CircleShape::getPointCount() const
 ////////////////////////////////////////////////////////////
 Vector2f CircleShape::getPoint(std::size_t index) const
 {
-    static const float pi = 3.141592654f;
-
-    float angle = static_cast<float>(index) * 2.f * pi / static_cast<float>(m_pointCount) - pi / 2.f;
-    float x = std::cos(angle) * m_radius;
-    float y = std::sin(angle) * m_radius;
+    Angle angle = static_cast<float>(index) / static_cast<float>(m_pointCount) * sf::degrees(360) - sf::degrees(180);
```
Marioalexsan, line 75: Here, `pi / 2.f` should be `sf::degrees(90)` instead of `sf::degrees(180)`.

examples/X11/X11.cpp
Bromeon: Indeed, 90/360 * pi = pi/4 rad = 180°/4 deg

src/SFML/Graphics/CircleShape.cpp
ChrisThrasher: Thank you! It should be fixed now.

examples/X11/X11.cpp
ChrisThrasher: Thank you! It should be fixed now.

Marioalexsan (approved):

eXpl0it3r:

vittorioromeo (approved):

PR #1971, opened 2022-01-18
Add support for VS 2022 to the GitHub Actions workflow

eXpl0it3r:
## Description

Add Windows Server 2022 image + VS 2022 support to the GitHub Actions CI pipeline.

Even though Windows Server 2016 + VS 2017 [should be removed](https://github.blog/changelog/2021-10-19-github-actions-the-windows-2016-runner-image-will-be-removed-from-github-hosted-runners-on-march-15-2022/) in March 2022, I left it in for now...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,8 +11,9 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-        - { name: Windows VS2017, os: windows-2016   }
-        - { name: Windows VS2019, os: windows-latest }
+        - { name: Windows VS2017, os: windows-2016 }
+        - { name: Windows VS2019, os: windows-2019 }
+        - { name: Windows VS2022, os: windows-2022 }
         - { name: Linux GCC,      os: ubuntu-latest  }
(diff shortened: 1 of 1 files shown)


Bromeon (approved):
Looks good to me!

I was first not sure why exactly these images:
```yaml
- { name: Windows VS2017, os: windows-2016 }
- { name: Windows VS2019, os: windows-2019 }
- { name: Windows VS2022, os: windows-2022 }
```

But then I found the list of installed toolchains for each runner image. It's amazing how much is installed!
* [`windows-2016`](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2016-Readme.md#visual-studio-enterprise-2017)
* [`windows-2019`](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#visual-studio-enterprise-2019)
* [`windows-2022`](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md#visual-studio-enterprise-2022)

---

And regarding VS 2017, one option would be to find a workaround for the `constexpr` issue (depending on `#ifdef`, we could have a runtime constant, like in SFML 2). Then we wouldn't need to drop support for the entire toolchain because of one unsupported C++17 feature.

But if there are more issues or the general consensus is that such workarounds are not worth the effort, we can also drop support.

PR #1972, opened 2022-01-20
Use built-in CMake project version functionality

ChrisThrasher:

----

## Description

CMake already has a notion of project version and various variables that encode it so let's use that. It's more idiomatic and requires less code.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android


eXpl0it3r (approved):
Nice find! 🙂

eXpl0it3r:

ChrisThrasher:
I think I fixed the failing MinGW builds.

eXpl0it3r:

CMakeLists.txt
```diff
@@ -46,9 +46,6 @@ include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/Config.cmake)
 include(GNUInstallDirs)
 
 # setup version numbers
-set(VERSION_MAJOR 3)
-set(VERSION_MINOR 0)
-set(VERSION_PATCH 0)
 set(VERSION_IS_RELEASE OFF)
```
kimci86, line 49: This could be right after the `project` command to keep related things together.

eXpl0it3r:

CMakeLists.txt
ChrisThrasher: Good idea. I moved that up.

PR #1973, opened 2022-01-21
Add tests for sf::Transformable

ChrisThrasher:

----

## Description

Next up in the series of Graphics module tests is `sf::Transformable`. This is a great next candidate because it's only dependency, `sf::Transform` is already tested. These tests are particularly valuable because they get modified in the `sf::Angle` pull request so the tests can be used to help verify the correctness of the `sf::Angle` implementation.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/test/Graphics/Transformable.cpp
+++ b/test/Graphics/Transformable.cpp
@@ -1,0 +1,120 @@
+#include <SFML/Graphics/Transformable.hpp>
+#include "GraphicsUtil.hpp"
+#include "SystemUtil.hpp"
+
+#include <doctest.h>
+
+using doctest::Approx;
+
+TEST_CASE("sf::Transformable class - [graphics]")
+{
+    SUBCASE("Construction")
+    {
+        const sf::Transformable transformable;
+        CHECK(transformable.getPosition() == sf::Vector2f(0, 0));
+        CHECK(transformable.getRotation() == sf::Angle::Zero);
+        CHECK(transformable.getScale() == sf::Vector2f(1, 1));
+        CHECK(transformable.getOrigin() == sf::Vector2f(0, 0));
+        CHECK(transformable.getTransform() == sf::Transform());
+        CHECK(transformable.getInverseTransform() == sf::Transform());
+    }
+
+    SUBCASE("Setters and getters")
+    {
+        sf::Transformable transformable;
+
+        transformable.setPosition({3, 4});
+        CHECK(transformable.getPosition() == sf::Vector2f(3, 4));
+
+        transformable.setRotation(sf::degrees(3.14f));
+        CHECK(transformable.getRotation() == sf::degrees(3.14f));
+        transformable.setRotation(sf::degrees(540));
+        CHECK(transformable.getRotation() == sf::degrees(180));
+        transformable.setRotation(sf::degrees(-72));
+        CHECK(transformable.getRotation() == sf::degrees(288));
+
+        transformable.setScale({5, 6});
+        CHECK(transformable.getScale() == sf::Vector2f(5, 6));
+
+        transformable.setOrigin({7, 8});
+        CHECK(transformable.getOrigin() == sf::Vector2f(7, 8));
+
+        sf::Transform transform;
+        transform.translate(transformable.getPosition() - transformable.getOrigin());
+        transform.rotate(transformable.getRotation(), transformable.getOrigin());
+        transform.scale(transformable.getScale(), transformable.getOrigin());
+
+        CHECK(transformable.getTransform().getMatrix()[0] == Approx(transform.getMatrix()[0]));
+        CHECK(transformable.getTransform().getMatrix()[1] == Approx(transform.getMatrix()[1]));
+        CHECK(transformable.getTransform().getMatrix()[2] == Approx(transform.getMatrix()[2]));
+        CHECK(transformable.getTransform().getMatrix()[3] == Approx(transform.getMatrix()[3]));
+        CHECK(transformable.getTransform().getMatrix()[4] == Approx(transform.getMatrix()[4]));
+        CHECK(transformable.getTransform().getMatrix()[5] == Approx(transform.getMatrix()[5]));
+        CHECK(transformable.getTransform().getMatrix()[6] == Approx(transform.getMatrix()[6]));
+        CHECK(transformable.getTransform().getMatrix()[7] == Approx(transform.getMatrix()[7]));
+        CHECK(transformable.getTransform().getMatrix()[8] == Approx(transform.getMatrix()[8]));
+        CHECK(transformable.getTransform().getMatrix()[9] == Approx(transform.getMatrix()[9]));
+        CHECK(transformable.getTransform().getMatrix()[10] == Approx(transform.getMatrix()[10]));
+        CHECK(transformable.getTransform().getMatrix()[11] == Approx(transform.getMatrix()[11]));
+        CHECK(transformable.getTransform().getMatrix()[12] == Approx(transform.getMatrix()[12]));
+        CHECK(transformable.getTransform().getMatrix()[13] == Approx(transform.getMatrix()[13]));
+        CHECK(transformable.getTransform().getMatrix()[14] == Approx(transform.getMatrix()[14]));
+        CHECK(transformable.getTransform().getMatrix()[15] == Approx(transform.getMatrix()[15]));
+
+        const sf::Transform inverse_transform = transform.getInverse();
+        CHECK(transformable.getInverseTransform().getMatrix()[0] == Approx(inverse_transform.getMatrix()[0]));
+        CHECK(transformable.getInverseTransform().getMatrix()[1] == Approx(inverse_transform.getMatrix()[1]));
+        CHECK(transformable.getInverseTransform().getMatrix()[2] == Approx(inverse_transform.getMatrix()[2]));
+        CHECK(transformable.getInverseTransform().getMatrix()[3] == Approx(inverse_transform.getMatrix()[3]));
+        CHECK(transformable.getInverseTransform().getMatrix()[4] == Approx(inverse_transform.getMatrix()[4]));
+        CHECK(transformable.getInverseTransform().getMatrix()[5] == Approx(inverse_transform.getMatrix()[5]));
+        CHECK(transformable.getInverseTransform().getMatrix()[6] == Approx(inverse_transform.getMatrix()[6]));
+        CHECK(transformable.getInverseTransform().getMatrix()[7] == Approx(inverse_transform.getMatrix()[7]));
+        CHECK(transformable.getInverseTransform().getMatrix()[8] == Approx(inverse_transform.getMatrix()[8]));
+        CHECK(transformable.getInverseTransform().getMatrix()[9] == Approx(inverse_transform.getMatrix()[9]));
+        CHECK(transformable.getInverseTransform().getMatrix()[10] == Approx(inverse_transform.getMatrix()[10]));
+        CHECK(transformable.getInverseTransform().getMatrix()[11] == Approx(inverse_transform.getMatrix()[11]));
+        CHECK(transformable.getInverseTransform().getMatrix()[12] == Approx(inverse_transform.getMatrix()[12]));
+        CHECK(transformable.getInverseTransform().getMatrix()[13] == Approx(inverse_transform.getMatrix()[13]));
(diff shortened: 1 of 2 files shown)


test/Graphics/Transformable.cpp
```diff
@@ -0,0 +1,123 @@
...
+        CHECK(transformable.getTransform().getMatrix()[6] == Approx(transform.getMatrix()[6]));
+        CHECK(transformable.getTransform().getMatrix()[7] == Approx(transform.getMatrix()[7]));
+        CHECK(transformable.getTransform().getMatrix()[8] == Approx(transform.getMatrix()[8]));
+        CHECK(transformable.getTransform().getMatrix()[9] == Approx(transform.getMatrix()[9]));
+        CHECK(transformable.getTransform().getMatrix()[10] == Approx(transform.getMatrix()[10]));
+        CHECK(transformable.getTransform().getMatrix()[11] == Approx(transform.getMatrix()[11]));
+        CHECK(transformable.getTransform().getMatrix()[12] == Approx(transform.getMatrix()[12]));
+        CHECK(transformable.getTransform().getMatrix()[13] == Approx(transform.getMatrix()[13]));
+        CHECK(transformable.getTransform().getMatrix()[14] == Approx(transform.getMatrix()[14]));
+        CHECK(transformable.getTransform().getMatrix()[15] == Approx(transform.getMatrix()[15]));
```
ChrisThrasher, line 65: I'm particularly proud of this. What I'm doing is using `sf::Transform` to recreate what `sf::Transformable::getTransform` is doing with lots of dense math expressions.

eXpl0it3r:

test/Graphics/Transformable.cpp
Bromeon: Maybe we should compare the entire matrix (not sure if this works with `Approx`). 

Generally, we should not re-test implementation of another unit in this unit test. `getMatrix()` is tested separately in the `sf::Transform` test suite, here we only care if two transforms are (approximately) equal. Feel free to reuse code from there.

Same for the inverse one below.

test/Graphics/Transformable.cpp
```diff
@@ -0,0 +1,123 @@
...
+        CHECK(transformable.getOrigin() == sf::Vector2f(0, 0));
+        CHECK(transformable.getTransform() == sf::Transform());
+        CHECK(transformable.getInverseTransform() == sf::Transform());
+    }
+
+    SUBCASE("Setters and getters")
+    {
+        sf::Transformable transformable;
+
+        CHECK(transformable.getPosition() == sf::Vector2f(0, 0));
```
Bromeon, line 25: We tested this already above. Same for rotation and scale.

test/Graphics/Transformable.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+        CHECK(transformable.getScale() == sf::Vector2f(1, 1));
+        transformable.setScale({5, 6});
+        CHECK(transformable.getScale() == sf::Vector2f(5, 6));
+
+        CHECK(transformable.getOrigin() == sf::Vector2f(0, 0));
+        transformable.setOrigin({7, 8});
+        CHECK(transformable.getOrigin() == sf::Vector2f(7, 8));
+
+        sf::Transform transform;
```
Bromeon, line 45: Maybe from here onward should be a separate test case; not about setters/getters.

test/Graphics/Transformable.cpp
ChrisThrasher: Comparing entire matrices does not work with `Approx`. 

This not testing `getMatrix()`. It's making sure that the transform returned by `getTransform` is what we would expect for that specific rotation, scale, position, and origin. Doing element-wise comparison is required because lots of floating point calculations are occurring. If not this, how else would we test `sf::Transformable::getTransform`?

test/Graphics/Transformable.cpp
ChrisThrasher: I lumped them all together since `getTransform` and `getInverseTransform` are still `get` functions. If the transform getters became their own test case then I'd just copy all the set calls from this case and we'd end up with two significantly similar test cases.

test/Graphics/Transformable.cpp
ChrisThrasher: I was on the fence about those precondition asserts but I'll remove them.

test/Graphics/Transformable.cpp
ChrisThrasher: > Maybe we should compare the entire matrix 

This is certainly what I first tried but then had a few matrix elements differ by approximately 1e-6 so I was forced to dig down deeper to write approximate comparisons. 

> here we only care if two transforms are (approximately) equal. Feel free to reuse code from there.

This is how you test if two transforms are approximately equal. You have to get the individual values of each matrix cell. I reused this pattern from the `sf::Transform` tests I wrote.

test/Graphics/Transformable.cpp
Bromeon: I would avoid testing the same thing twice. It makes different test cases dependent on each other.

test/Graphics/Transformable.cpp
Bromeon: The thing is, we have now 4 styles for "compare matrix elements against expected result":

* the one here

* https://github.com/SFML/SFML/blob/60deef9997e1fe3e1c1fb2a5ed949f02553c4d0c/test/Graphics/Transform.cpp#L36-L40

* https://github.com/SFML/SFML/blob/60deef9997e1fe3e1c1fb2a5ed949f02553c4d0c/test/Graphics/Transform.cpp#L103-L105

* https://github.com/SFML/SFML/blob/60deef9997e1fe3e1c1fb2a5ed949f02553c4d0c/test/Graphics/Transform.cpp#L127-L135

Already then I suggested to write a function that compares an entire matrix; it could go into the utils directory.

It's probably easier (and more readable) to have one or two functions which accept a matrix in its literal notation. I don't think `Approx` really hurt; some operations just happen to have no rounding errors right now.

The `sf::Transform` constructor can still be used when a 3x3 matrix is needed.

ChrisThrasher: There are 3 comparisons going on. One is an approximate comparison of transforms. This is what styles 1 and 4 do (same thing). The 2nd is a perfect comparison of the 4x4 matrices returned by `sf::Transform::getMatrix`. This is just how we have to test `getMatrix`. You have to make sure each element of that array is exactly what you expect. The 3rd is a perfect comparison of `sf::Transform`s. 

I don't see any unifying solution for these 3 distinct tasks. For lack of a good unifying abstraction, I'd prefer to write the tests in a more verbose fashion that doesn't require extra infrastructure like adding new `operator<<` overloads for non-`sf` types (see test/Graphics/Transform.cpp:12). All those extra bits and pieces required for testing (but which aren't tests) add surface for bugs and reduce the readability of tests which must be correct by manual inspection.

Bromeon: As mentioned, we can have 2 functions (one for 3x3, one for 4x4). If we really need to differentiate exact/approx, we can also have a third one. But we currently don't differentiate them very consequently -- some transform operations use exact comparisons because they happen to work on integers, others don't. Setting and getting is the main operation where exact would be strictly required.

"Strictly" means that we could change the implementation so that `2.0 * 2.0 == 4.00001`. It's not nice, but definitely not unreasonable. However, we cannot change it so that `setX(2.0); getX() == 2.00001`.

---

Regarding abstraction, why not something like this?
```cpp
checkExact(transform, 18.0f, 18.0f, 14.0f, 
                      36.0f, 41.0f, 36.0f, 
                      14.0f, 18.0f, 18.0f);

checkExact(transform, 10.0f, 13.0f, 0.0f, 16.0f, 
                      11.0f, 14.0f, 0.0f, 17.0f, 
                       0.0f,  0.0f, 1.0f,  0.0f, 
                      12.0f, 15.0f, 0.0f, 18.0f);

checkApprox(transform, 10.0f, 13.0f, 0.0f, 16.0f, 
                       11.0f, 14.0f, 0.0f, 17.0f, 
                        0.0f,  0.0f, 1.0f,  0.0f, 
                       12.0f, 15.0f, 0.0f, 18.0f);
```
We can use macros if we want to preserve line numbers.

It's not just about code reuse, but also readability and uniform style for similar operations. All three of these lines make immediately clear what happens, and what is the expected result. In contrast, the following code (especially with the shuffled indexes) makes the expected value very hard to decipher, and failing tests harder to debug:
```cpp
 CHECK(transform.getMatrix()[0] == Approx(0)); 
 CHECK(transform.getMatrix()[4] == Approx(-1)); 
 CHECK(transform.getMatrix()[12] == Approx(0)); 
 CHECK(transform.getMatrix()[1] == Approx(1)); 
 CHECK(transform.getMatrix()[5] == Approx(0)); 
 CHECK(transform.getMatrix()[13] == Approx(0)); 
 CHECK(transform.getMatrix()[3] == Approx(0)); 
 CHECK(transform.getMatrix()[7] == Approx(0)); 
 CHECK(transform.getMatrix()[15] == Approx(1)); 
 ```

As usual, I'm not saying you have to go and fix the other test now 🙂 but this came to my mind as I saw more code with this pattern. We can gladly delegate any potential `sf::Transform` test changes to a future PR.

Comment (ChrisThrasher) [reply test/Graphics/Transformable.cpp:61]:
```cpp
checkExact(transform, 18.0f, 18.0f, 14.0f, 
                      36.0f, 41.0f, 36.0f, 
                      14.0f, 18.0f, 18.0f);
```
This already exists in the form of `operator==` for `sf::Transform` so that one is easy.

---

This is how I picture the various abstraction we need for approximate and exact comparisons of `sf::Transform` and `float[16]`. 

|                | sf::Transform                                                     | float[16] |
| ---        | ---                                                                        | --- |
| Exact   | operator==                                                          | `checkExact(float[16], float[16])` | 
| Approx | `checkApprox(sf::Transform, sf::Transform)` | `checkApprox(float[16], float[16])` |

One thing we can do to make comparisons easier is to change the `sf::Transform::getMatrix` return type from `float*` to `std::array<float, 16>` which gives us `operator==` for free and removes the `checkExact(float[16], float[16])` overload on top of simply being an easier type to use and harder type to misuse. That leaves us with two macros for approximate comparisons (to preserve line numbers) plus an `operator<<` overload for `std::array<float, 16>`. `std::array::data` is still present for anyone who wants a `float*` pointing to the first element so this change will easily work with existing uses of `getMatrix()`.

I can make these changes (including the `sf::Transform::getMatrix` return type change) in this PR or we can suffer through some inconsistent comparisons for a bit longer and I'll make all these changes in a later PR that changes the API while unifying the style of these tests.

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

tamhuynh3861:
Hi, I'd like to contribute by writing test cases. Are there any issues/code I can look at to get a general idea of what we're working on?

ChrisThrasher:
> Hi, I'd like to contribute by writing test cases. Are there any issues/code I can look at to get a general idea of what we're working on?

Look at the existing tests to get an idea for how we're testing things. Be sure to copy all the patterns used in other test files. Then find something which hasn't yet been tested and start writing tests. Join the [SFML Discord](https://www.sfml-dev.org/community.php) to talk more about contribution.

tamhuynh3861:
What do you recommend for running the test cases?

Bromeon:
@tamhuynh3861 If you need more guidance regarding testing, please open a thread on the forum.

This pull request is for the proposed tests about `sf::Transformable` in particular, let's stay on-topic please.

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

vittorioromeo:

vittorioromeo (approved):
I'm happy with this, we can improve/refactor in a future PR.

binary1248:
Is it just me or was the new Transformable.cpp not added to the CMake test configuration? 😅

Seems kinda suspicious that Codecov doesn't report higher coverage for this changeset. 😛

ChrisThrasher:
> Is it just me or was the new Transformable.cpp not added to the CMake test configuration? 😅
> 
> Seems kinda suspicious that Codecov doesn't report higher coverage for this changeset. 😛

Oh my god...

Haha I actually know why this happened and it's 100% my fault. I'll submit a fixup PR real quick. Sorry, everybody!

ChrisThrasher:
See #2023

PR #1974, opened 2022-01-21
Remove GCC 4 workaround

ChrisThrasher:

----

## Description

No more need to work around GCC 4 shortcomings. 

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/include/SFML/Config.hpp
+++ b/include/SFML/Config.hpp
@@ -148,20 +148,8 @@
 
     #else // Linux, FreeBSD, Mac OS X
 
-        #if __GNUC__ >= 4
-
-            // GCC 4 has special keywords for showing/hidding symbols,
-            // the same keyword is used for both importing and exporting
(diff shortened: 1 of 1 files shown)


eXpl0it3r (approved):

eXpl0it3r:

Bromeon:
Just to understand the context, why is this no longer needed? Because we don't support that version anymore?

If so, do we want to make it explicit? Would be nicer than weird errors down the line.
```cpp
#if __GNUC__ < 4
   #error This version of g++ is not supported by SFML.
#endif
```

eXpl0it3r:
GCC 4 doesn't support C++17, so it should fail when trying to enable C++17

https://gcc.gnu.org/projects/cxx-status.html#cxx17

ChrisThrasher:
@eXpl0it3r is right. It's because GCC 4 only supports some subset of C++14.

PR #1976, opened 2022-01-21
Simplify sf::Transform tests

kimci86:
## Description

Add some utilities to make writing `Transform` tests simpler.

Below is the previous description. It turns out using `std::array` in this context is not so great.

> This PR replaces C-style arrays in `sf::Tranform` and `sf::priv::Matrix` with `std::array`.
>
> The main motivation is to make testing the `Transform` class easier. I created this to contribute to the discussion in #1973.
By having the underlying matrix as an `std::array`, it makes it possible to compare matrices in a single statement using comparison operators on `std::array` which plays nice with doctest macros.
> 
> As floating-point values need a tolerant comparison for tests and `doctest::Approx` only handles single values, the utility class `ApproxArray` is proposed to compare floating-point arrays.
> 
> I looked for other opportunities to replace C-style arrays with `std::array`. There are other places where C-style arrays are used, but I feel replacing them with an `std::array` brings no benefit, so I did not do it.
Many functions are taking a pointer and a size, but this is not where `std::array` would be useful. This is where `std::span` would be nice but it is not available with C++17, only from C++20.

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

- CI
- Checking debug strings are sensible when running `./test-sfml-graphics --success --test-case=sf::Transform*`

### DIFF
--- a/test/Graphics/Transform.cpp
+++ b/test/Graphics/Transform.cpp
@@ -1,23 +1,12 @@
 #include <SFML/Graphics/Transform.hpp>
 #include "GraphicsUtil.hpp"
-#include "SystemUtil.hpp"
-#include <vector>
+#include "Array.hpp"
+#include "Approx.hpp"
 
 #include <doctest.h>
 
-using doctest::Approx;
-
-namespace std
-{
-std::ostream& operator<<(std::ostream& out, const std::vector<float>& vector)
-{
-    out << "{ ";
-    for (size_t i = 0; i + 1 < vector.size(); ++i)
-        out << vector[i] << ", ";
-    out << vector.back() << " }";
-    return out;
-}
-}
+using sf::Testing::toArray;
+using sf::Testing::Approx;
 
 TEST_CASE("sf::Transform class - [graphics]")
 {
@@ -33,21 +22,20 @@ TEST_CASE("sf::Transform class - [graphics]")
             const sf::Transform transform(10.0f, 11.0f, 12.0f,
                                           13.0f, 14.0f, 15.0f,
                                           16.0f, 17.0f, 18.0f);
-            const std::vector<float> matrix(transform.getMatrix(), transform.getMatrix() + 16);
-            CHECK(matrix == std::vector<float>{10.0f, 13.0f, 0.0f, 16.0f,
-                                               11.0f, 14.0f, 0.0f, 17.0f,
-                                                0.0f,  0.0f, 1.0f,  0.0f,
-                                               12.0f, 15.0f, 0.0f, 18.0f});
+
+            CHECK(toArray<16>(transform.getMatrix()) == std::array{10.0f, 13.0f, 0.0f, 16.0f,
+                                                                   11.0f, 14.0f, 0.0f, 17.0f,
+                                                                    0.0f,  0.0f, 1.0f,  0.0f,
+                                                                   12.0f, 15.0f, 0.0f, 18.0f});
         }
     }
 
     SUBCASE("Identity matrix")
     {
-        const std::vector<float> matrix(sf::Transform::Identity.getMatrix(), sf::Transform::Identity.getMatrix() + 16);
-        CHECK(matrix == std::vector<float>{1.0f, 0.0f, 0.0f, 0.0f,
-                                           0.0f, 1.0f, 0.0f, 0.0f,
-                                           0.0f, 0.0f, 1.0f, 0.0f,
-                                           0.0f, 0.0f, 0.0f, 1.0f});
+        CHECK(toArray<16>(sf::Transform::Identity.getMatrix()) == std::array{1.0f, 0.0f, 0.0f, 0.0f,
+                                                                             0.0f, 1.0f, 0.0f, 0.0f,
+                                                                             0.0f, 0.0f, 1.0f, 0.0f,
+                                                                             0.0f, 0.0f, 0.0f, 1.0f});
     }
 
     SUBCASE("getInverse()")
@@ -124,30 +112,18 @@ TEST_CASE("sf::Transform class - [graphics]")
         {
             sf::Transform transform;
             transform.rotate(90);
-            CHECK(transform.getMatrix()[0] == Approx(0));
-            CHECK(transform.getMatrix()[4] == Approx(-1));
-            CHECK(transform.getMatrix()[12] == Approx(0));
-            CHECK(transform.getMatrix()[1] == Approx(1));
-            CHECK(transform.getMatrix()[5] == Approx(0));
-            CHECK(transform.getMatrix()[13] == Approx(0));
-            CHECK(transform.getMatrix()[3] == Approx(0));
-            CHECK(transform.getMatrix()[7] == Approx(0));
-            CHECK(transform.getMatrix()[15] == Approx(1));
+            CHECK(transform == Approx(sf::Transform(0.0f, -1.0f, 0.0f,
+                                                    1.0f,  0.0f, 0.0f,
+                                                    0.0f,  0.0f, 1.0f)));
         }
 
         SUBCASE("Around custom point")
         {
             sf::Transform transform;
             transform.rotate(90, {1.0f, 0.0f});
-            CHECK(transform.getMatrix()[0] == Approx(0));
-            CHECK(transform.getMatrix()[4] == Approx(-1));
-            CHECK(transform.getMatrix()[12] == Approx(1));
-            CHECK(transform.getMatrix()[1] == Approx(1));
-            CHECK(transform.getMatrix()[5] == Approx(0));
-            CHECK(transform.getMatrix()[13] == Approx(-1));
-            CHECK(transform.getMatrix()[3] == Approx(0));
-            CHECK(transform.getMatrix()[7] == Approx(0));
-            CHECK(transform.getMatrix()[15] == Approx(1));
+            CHECK(transform == Approx(sf::Transform(0.0f, -1.0f,  1.0f,
+                                                    1.0f,  0.0f, -1.0f,
+                                                    0.0f,  0.0f,  1.0f)));
         }
     }
 
--- a/test/TestUtilities/Approx.hpp
+++ b/test/TestUtilities/Approx.hpp
@@ -1,0 +1,31 @@
+#ifndef SFML_TESTUTILITIES_APPROX_HPP
+#define SFML_TESTUTILITIES_APPROX_HPP
+
+#include <ostream>
+
+namespace sf::Testing
+{
+    template <typename T>
+    struct Approx
+    {
+        Approx(const T& value)
+        : value{value}
+        {}
+
+        const T& value;
+    };
+
+    template <typename T>
+    bool operator ==(const T& value, const Approx<T>& approx)
+    {
+        return almostEqual(value, approx.value);
+    }
+
+    template <typename T>
+    std::ostream& operator <<(std::ostream& out, const sf::Testing::Approx<T>& approx)
+    {
+        return out << approx.value;
+    }
+}
+
+#endif // SFML_TESTUTILITIES_APPROX_HPP
--- a/test/TestUtilities/Array.hpp
+++ b/test/TestUtilities/Array.hpp
@@ -1,0 +1,41 @@
+#ifndef SFML_TESTUTILITIES_ARRAY_HPP
+#define SFML_TESTUTILITIES_ARRAY_HPP
+
+#include <array>
+#include <limits>
+#include <ostream>
+
+namespace sf::Testing
+{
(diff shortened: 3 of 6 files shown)


src/SFML/Graphics/Shader.cpp
```diff
@@ -646,9 +646,12 @@ void Shader::setUniformArray(const std::string& name, const Glsl::Mat3* matrixAr
 {
     const std::size_t matrixSize = 3 * 3;
 
-    std::vector<float> contiguous(matrixSize * length);
+    std::vector<float> contiguous;
+    contiguous.reserve(matrixSize * length);
```
ChrisThrasher, line 650: Why use `.reserve` over the `vector` constructor which does the same thing?

test/TestUtilities/ApproxArray.hpp
```diff
@@ -0,0 +1,72 @@
...
+    bool operator!=(const ApproxArray<T, size>& approx, const std::array<T, size>& array)
+    {
+        return !(approx == array);
+    }
+
+    template <typename T, std::size_t size>
+    bool operator!=(const std::array<T, size>& array, const ApproxArray<T, size>& approx)
+    {
+        return !(approx == array);
+    }
```
ChrisThrasher, line 63: What's the use case for != in this context?

test/TestUtilities/ApproxArray.hpp
```diff
@@ -0,0 +1,72 @@
...
+{
+    template <typename T, std::size_t size>
+    std::ostream& operator<<(std::ostream& out, const std::array<T, size>& array)
+    {
+        if (size == 0)
+            return out << "{}";
+
+        out << "{ ";
+        for (size_t i = 0; i + 1 < array.size(); ++i)
+            out << array[i] << ", ";
```
ChrisThrasher, line 17: There's a tiny inconsistency here with the use of `size_t` instead of `std::size_t`.

ChrisThrasher:
I like this! It's great how this lets us keep using the `CHECK` macro which gives us line number info without us having to write any macros.

test/TestUtilities/ApproxArray.hpp
```diff
@@ -0,0 +1,72 @@
...
+        out << array.back() << " }";
```
ChrisThrasher, line 18: We may need to increase the precision of the `ostream` to make sure all significant information is printed otherwise you may issues where something like `1.00000 == 1.00000` gets printed to console when the real numbers are something like `1.000000000 == 1.000000211`. I had this issue when writing the `sf::Angle` tests and later realized I needed the following to make sure the full precision of a `float` got printed.
```
out << std::fixed << std::setprecision(std::numeric_limits<float>::max_digits10);
```

test/TestUtilities/ApproxArray.hpp
```diff
@@ -0,0 +1,72 @@
...
+    template <typename T, std::size_t size>
+    class ApproxArray
+    {
+    public:
+        ApproxArray(const std::array<T, size>& values)
+        : values{values}
+        {}
+
+        const std::array<T, size>& values;
+    };
```
ChrisThrasher, line 35: Should this just be a struct so we can depend on the default `public` access of everything?

src/SFML/Graphics/Glsl.cpp
```diff
@@ -27,7 +27,6 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Graphics/Glsl.hpp>
-#include <algorithm>
```
ChrisThrasher, line 30: @vittorioromeo will love this :)

include/SFML/Graphics/Glsl.inl
```diff
@@ -57,15 +48,13 @@ struct Matrix
...
+    ///             The elements are copied to the instance.
     ///
     ////////////////////////////////////////////////////////////
-    explicit Matrix(const float* pointer)
-    {
-        copyMatrix(pointer, Columns * Rows, array);
-    }
+    explicit Matrix(const std::array<float, Columns * Rows>& data)
+    : array{data}
+    {}
```
Bromeon, line 57: This is less powerful than the previous API.

`const float*` allows for various APIs which don't use `std::array`. Not just C ones, even `std::vector`, or an `std::array` with more (unused) elements.

We could add a constructor overload, for symmetry reasons with `getMatrix()`.

src/SFML/Graphics/Shader.cpp
Bromeon: You probably confuse it with `resize()`. The `vector` constructor accepting a number default-constructs those elements, doesn't just allocate space.

But what was the idea behind this change? Simpler code?

test/TestUtilities/ApproxArray.hpp
Bromeon: I agree, we could settle for the convention `exact == approx` and `exact != approx` everywhere, this is not exposed to user code.

src/SFML/Graphics/Shader.cpp
kimci86: I first though `copyMatrix(const float* source, std::size_t elements, float* dest)` was not needed anymore when editing Glsl.inl/cpp and removed it. Then this code was broken. I replaced it with a loop without thinking too much about it.
Using `reserve` instead of `resize` or the constructor with a size allows me to use `push_back` to take care of the destination index instead of computing it myself in the loop below. The idea was me being lazy.
Investigating some more with [a small benchmark](https://www.quick-bench.com/q/K7iEBSn4GxLy_uPqDJKbthwUY2o), it seems that using `push_back` is much slower than `std::copy` or computing the destination index, so this code will change.

test/TestUtilities/ApproxArray.hpp
kimci86: Right, we only use `exact == approx`, so I will remove other comparisons.

eXpl0it3r:

include/SFML/Graphics/Glsl.inl
kimci86: I put back the constructor for raw data.
I wonder if the overload with `std::array` is so useful. The good abstraction would be `std::span` but it comes with C++20.

src/SFML/Graphics/Shader.cpp
kimci86: I went with `copyMatrix` as before.

src/SFML/Graphics/Glsl.cpp
kimci86: Sorry, I went back to using `std::copy`!
I noticed there is `priv::copy` is `System/Utf.hpp`. Maybe this can be in a dedicated header to reuse it and avoid `<algorithm>`?

ChrisThrasher: If we're just copying contiguous memory from one place to another then `std::memcpy` will do the job quickly and without an expensive header inclusion. Right?

ChrisThrasher:
```cpp
    for (std::size_t i = 0; i < length; ++i)
        std::memcpy(&contiguous[matrixSize * i], matrixArray[i].array.data(), matrixArray[i].array.size());
```
We might also be able to remove the `matrixSize` local variable since that's encoded `.array.size()`.

kimci86: Well, `std::memcpy` this is very much C, more error-prone, taking void pointers. Not sure we want that.

kimci86: > more error-prone

See, in your example above, `matrixArray[i].array.size()` should be `matrixArray[i].array.size() * sizeof(float)`

ChrisThrasher: Haha oops. Point taken. So `memcpy` isn't the way to go.

ChrisThrasher: Ultimately it depends how desperate we are to avoid `<algorithm>`. It is possible to get `std::memcpy` right even if I messed it up on my first attempt. We can make it look a little better like this:
```cpp
std::memcpy(&contiguous[matrixSize * i], matrixArray[i].array.data(), sizeof(matrixArray[i].array));
```
which should be correct since the sizeof a `std::array` is actually what we mean, not the number of elements.

test/TestUtilities/ApproxArray.hpp
kimci86: I would not have thought of it. Thank you. I don't think `std::fixed` is needed though.
The precision is now appropriately set for arrays of floating-point values.

include/SFML/Graphics/Glsl.inl
kimci86: The constructor taking `std::array` seems useless or too specialized after all. I removed it.

src/SFML/Graphics/Glsl.cpp
kimci86: I think the C-style ugliness is not worth it, but let's see what others think.

kimci86:
Thank you for your quick feedback. I think it is ready now so I converted the draft PR into a regular PR. I will squash my commits when we agree.

src/SFML/Graphics/Glsl.cpp
Bromeon: I think it's OK to have `<algorithm>` in .cpp files, the user compiles them only once (if at all) and doesn't pay for the extra overhead on each compilation.

The old `copyMatrix()` implementation has done its job well, I don't think we need to go out of our way to find a new implementation.

eXpl0it3r:

include/SFML/Graphics/Glsl.inl
Bromeon: Following this line of thought, don't we also want to revert to returning `const float*` in `getMatrix()`?

By returning `const std::array&`, we allow precisely one implementation and that is having a field which is `std::array`. So it's almost as if we had the field public for read access.

Again, a `std::span` (C++20) would fit here, but `float*` is more generic. We just need to document how many elements are behind the pointer.

kimci86: The motivation for this PR was to return `const std::array&` to make writing unit tests simpler.
I am now not very sure it is the way to go. I agree it does not feel so elegant in `sf::Transform`'s interface. What we would need to have a nice interface is indeed `std::span`.

Now, I think we only need a testing infrastructure that can handle `const float*` arrays and switching to `std::array` bring little to no benefit.

Bromeon: What if we just have a thin wrapper type around `const float*` that is used in tests?
Could even be `std::array<float, 16>` with an easy constructor function `toArray(floatPtr)` or so...

eXpl0it3r:

kimci86:
As having `std::array` in `sf::Transform`'s interface does not seem so elegant after all, I tried another approach with just a few additional testing utilities. I looks better now in my opinion.

include/SFML/Graphics/Glsl.inl
kimci86: I agree, I implemented this approach.

ChrisThrasher:
I find it odd that we're keeping a C-style API detail in spite of how much work it imposes on us trying to test it. The fact that we have to jump through hoops to test this function should be an indication that a raw pointer maybe isn't the best return type we can pick. I think `std::array` is our best bet currently but it's worth looking into if there should be a `sf::Span` that handles cases like this.

eXpl0it3r:

kimci86:
I am sorry for the back and forth (and that the branch name is not relevant now), but I feel `std::array` without `std::span` does not bring much.
In the meantime, this PR makes tests simpler, which was my original goal, with little changes.

ChrisThrasher:
> In the meantime, this PR makes tests simpler, which was my original goal, with little changes.

Select parts of the tests are simpler at the cost of more hidden complexity elsewhere. We're slowly building an ad hoc testing library. If our intention is to have a private library for implementing tests (including stuff like all the `operator<<` overloads) then let's lean into that idea and make it an intentional choice instead of an incidental one. 

I'd argue that if you need an extra library to test your library then that highlights issues with the library you're testing. We should be thinking harder about why we need everything inside `TestUtilities` instead of simply adding more and more to that directory.

kimci86:
> I'd argue that if you need an extra library to test your library then that highlights issues with the library you're testing. We should be thinking harder about why we need everything inside TestUtilities instead of simply adding more and more to that directory. In an ideal world, TestUtilties wouldn't need to exist.

I think it is more an issue with the programming language than the library. Ideally, a testing framework like doctest would not be needed.

Anyways, I will start over after #1979 gets merged.

PR #1977, opened 2022-01-22
Remove unnecessary CMake version requirement

ChrisThrasher:

----

## Description

This is left over from 0f83e3d but we forgot to remove it. Nothing about this file requires an elevated minimum CMake version now that the FetchContent usage is gone.

The fact that its inclusion didn't break any CI jobs means we can probably safely upgrade to CMake 3.14 (or newer) in the future with minimal impact. I'm not proposing raising that requirement without cause, but we should consider what improvements can be made if we do raise that requirement. CMake really hit its stride with 3.14 and it'd be a great minimum version to target.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,5 +1,3 @@
-cmake_minimum_required(VERSION 3.14)
-
 set(SRCROOT "${PROJECT_SOURCE_DIR}/test")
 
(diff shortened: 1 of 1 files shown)


eXpl0it3r:

eXpl0it3r (approved):

PR #1979, opened 2022-01-25
Extension methods for Vector2

### LINKED ISSUE #1810: More functionality for Vectors (2D)
## Subject of the issue

More functions and features for Vectors including but not limited to:

#### fromAngle
Calculates and returns a new 2D unit vector from the specified angle value
#### mag
Calculates the magnitude (length) of the vector and returns the result as a float
#### dist
Calculates the Euclidean distance between two points (considering a point as a vector object)
#### dot
Calculates the dot product of two vectors
#### cross
Calculates and returns a vector composed of the cross product between two vectors
#### normalize
Normalize the vector to length 1 (make it a unit vector)
#### limit
Limit the magnitude of the vector to the value used for the parameter
#### setMag
Set the magnitude of the vector to the value used for the parameter
#### heading
Calculates the angle of rotation for a vector
#### rotate
Rotates a vector by the specified angle while maintaining the same magnitude
#### angleBetween
Calculates and returns the angle between two vectors

I used the Processing PVector class as the reference, here's the link: [PVector](https://processing.org/reference/PVector.html)

Bromeon:
## Description

Adds extension methods to `sf::Vector2f`, heavily inspired by my library [**Thor/Vectors**](https://github.com/Bromeon/Thor/tree/master/include/Thor/Vectors) (API doc for [2D](https://bromeon.ch/libraries/thor/documentation/v2.1/_vector_algebra2_d_8hpp.html) and [3D](https://bromeon.ch/libraries/thor/documentation/v2.1/_vector_algebra3_d_8hpp.html)).

First and foremost, I apologize for creating this PR while others are working on something similar. But [I did mention this](https://github.com/SFML/SFML/issues/1810#issuecomment-990214480) more than a month ago, which was overlooked. I wanted to avoid that people spend even more time thinking of an API and implementation, when a lot of prior work has already been available and widely used over the past decade. Unfortunately, as [ChrisTrasher noted correctly](https://github.com/SFML/SFML/issues/1810#issuecomment-1021503712), writing a useful vector library is not a trivial task at all.

By no means do I think the Thor implementation is ready-to-use for SFML. On the contrary, I think design goals for SFML (especially SFML 3) likely differ. However, I wanted to show a relatively feature-complete implementation as a starting point, so we don't need to start API discussions from zero.

I added the implementation on top of the pending `sf::Angle` proposal, but you can select a sub-range of commits to see only new changes.

Closes #1810.

## Tasks

This PR is in draft state; a lot of API questions need to be resolved.

### 1\. Do we want to support `sf::Vector<T>` or only `sf::Vector2f`?
   
   In Thor, I created `TrigonometricTraits` allowing generic extensions to support any type `T` (such as custom decimal). In the scope of SFML, I think that would be over-engineering. The entire SFML API accepts only `sf::Vector2f` for floating point vectors and there are no typedefs for `double` or `long double`, so I think it's reasonable to limit the extensions to `float`.

Almost all extension methods are meaningless for integer types; even those theoretically computable (dot product) will in practice be applied to float vectors.

### 2\. Should the methods be free or member functions?
   
   There are arguments for both, so even a mixture might be possible:
   * Symmetry: `sf::dot(vec0, vec1)`
   * Clarity: `vec.rotatedBy(angle)`

### 3\. File structure
   
   Should we add the functionality to `Vector2.hpp` or a separate header? It probably makes sense if the `Vector2` related functionality is self-contained. I also wanted to keep `<cmath>` out of the header, since `Vector2.hpp` is such a heavily depended-upon file.

### 4\. Which methods do we need and how do we name them?

I adopted the current set from Thor and found them useful for my own games, but opinions can differ.

Naming is also a topic, but it depends on point 2, so we should first decide if we go for free functions, methods or both. In general, one thing that would be super-nice are relatively short names. Doesn't have to be extreme like `len()`, but maybe `cross()` instead of `crossProduct()` etc. Keep in mind that those operations are often part of larger algebraic expressions.

### 5\. Component-wise operators

Having `vec0 * vec1` and `vec0 / vec1` for component-wise multiplication/division is useful for scaling. Most vector libraries I have encountered over the years implement this, and I think the risk for misunderstanding (as dot or cross product) is quite small. It's also possible to provide a named function for it, at slightly increased verbosity cost.

### 6\. 3D vectors

   Should very limited support for 3D vectors be added (dot/cross product, angle, etc)?
   If yes, I'll gladly provide my Thor implementation, but this could also be done at a later stage.

### DIFF
--- a/include/SFML/System/Vector2.hpp
+++ b/include/SFML/System/Vector2.hpp
@@ -25,11 +25,16 @@
 #ifndef SFML_VECTOR2_HPP
 #define SFML_VECTOR2_HPP
 
+#include <SFML/System/Angle.hpp>
+#include <cassert>
+#include <cmath>
+#include <type_traits>
+
 
 namespace sf
 {
 ////////////////////////////////////////////////////////////
-/// \brief Utility template class for manipulating
+/// \brief Class template for manipulating
 ///        2-dimensional vectors
 ///
 ////////////////////////////////////////////////////////////
@@ -68,13 +73,142 @@ public:
     ////////////////////////////////////////////////////////////
     template <typename U>
     constexpr explicit Vector2(const Vector2<U>& vector);
+    
+    ////////////////////////////////////////////////////////////
+    /// \brief Length of the vector <i><b>(floating-point)</b></i>.
+    ///
+    /// If you are not interested in the actual length, but only in comparisons, consider using lengthSq().
+    ///
+    ////////////////////////////////////////////////////////////
+    T length() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Square of vector's length.
+    /// 
+    /// Suitable for comparisons, more efficient than length().
+    ///
+    ////////////////////////////////////////////////////////////
+    constexpr T lengthSq() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Vector with same direction but length 1 <i><b>(floating-point)</b></i>.
+    /// 
+    /// \pre \c *this is no zero vector.
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] Vector2 normalized() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Signed angle from \c *this to \c rhs <i><b>(floating-point)</b></i>.
+    /// 
+    /// \return The smallest angle which rotates \c *this in positive
+    /// or negative direction, until it has the same direction as \c rhs.
+    /// The result has a sign and lies in the range [-180, 180°).
+    /// \pre Neither \c *this nor \c rhs is a zero vector.
+    ///
+    ////////////////////////////////////////////////////////////
+    Angle angleTo(const Vector2& rhs) const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Signed angle from +X or (1,0) vector <i><b>(floating-point)</b></i>.
+    /// 
+    /// For example, the vector (1,0) corresponds to 0 degrees, (0,1) corresponds to 90 degrees.
+    /// 
+    /// \return Angle in the range [-180°, 180°).
+    /// \pre This vector is no zero vector.
+    ///
+    ////////////////////////////////////////////////////////////
+    Angle angle() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Rotate by angle \c phi <i><b>(floating-point)</b></i>.
+    /// 
+    /// Returns a vector with same length but different direction.
+    ///
+    /// In SFML's default coordinate system with +X right and +Y down,
+    /// this amounts to a clockwise rotation by \c phi.
+    /// 
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] Vector2 rotatedBy(Angle phi) const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Projection of this vector onto \c axis <i><b>(floating-point)</b></i>.
+    /// 
+    /// \param axis Vector being projected onto. Need not be normalized.
+    /// \pre \c axis must not have length zero.
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] constexpr Vector2 projectedOnto(const Vector2& axis) const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Returns a perpendicular vector.
+    /// 
+    /// Returns \c *this rotated by +90 degrees; (x,y) becomes (-y,x).
+    /// For example, the vector (1,0) is transformed to (0,1).
+    ///
+    /// In SFML's default coordinate system with +X right and +Y down,
+    /// this amounts to a clockwise rotation.
+    ///
+    ////////////////////////////////////////////////////////////
+    constexpr Vector2 perpendicular() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Dot product of two 2D vectors.
+    ///
+    ////////////////////////////////////////////////////////////
+    constexpr T dot(const Vector2& rhs) const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Z component of the cross product of two 2D vectors.
+    /// 
+    /// Treats the operands as 3D vectors, computes their cross product
+    /// and returns the result's Z component (X and Y components are always zero).
+    ///
+    ////////////////////////////////////////////////////////////
+    constexpr T cross(const Vector2& rhs) const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Component-wise multiplication of \c *this and \c rhs.
+    ///
+    /// Computes <tt>(lhs.x*rhs.x, lhs.y*rhs.y)</tt>.
+    /// 
+    /// Scaling is the most common use case for component-wise multiplication/division.
+    /// This operation is also known as the Hadamard or Schur product.
+    ///
+    ////////////////////////////////////////////////////////////
+    constexpr Vector2 cwiseMul(const Vector2& rhs) const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Component-wise division of \c *this and \c rhs.
+    /// 
+    /// Computes <tt>(lhs.x/rhs.x, lhs.y/rhs.y)</tt>.
+    /// 
+    /// Scaling is the most common use case for component-wise multiplication/division.
+    /// 
+    /// \pre Neither component of \c rhs is zero.
+    ///
+    ////////////////////////////////////////////////////////////
+    constexpr Vector2 cwiseDiv(const Vector2& rhs) const;
+
 
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
     T x; //!< X coordinate of the vector
     T y; //!< Y coordinate of the vector
+
+
+    ////////////////////////////////////////////////////////////
+    // Static member data
+    ////////////////////////////////////////////////////////////
+    static const Vector2 UnitX; //!< The X unit vector (1, 0), usually facing right
+    static const Vector2 UnitY; //!< The Y unit vector (0, 1), usually facing down
 };
+
+// Define the most common types
+using Vector2i = Vector2<int>;
+using Vector2u = Vector2<unsigned int>;
+using Vector2f = Vector2<float>;
 
 ////////////////////////////////////////////////////////////
 /// \relates Vector2
@@ -93,12 +227,12 @@ template <typename T>
 /// \brief Overload of binary operator +=
 ///
 /// This operator performs a memberwise addition of both vectors,
-/// and assigns the result to \a left.
+/// and assigns the result to \c left.
 ///
 /// \param left  Left operand (a vector)
 /// \param right Right operand (a vector)
 ///
-/// \return Reference to \a left
+/// \return Reference to \c left
 ///
 ////////////////////////////////////////////////////////////
 template <typename T>
@@ -109,12 +243,12 @@ constexpr Vector2<T>& operator +=(Vector2<T>& left, const Vector2<T>& right);
 /// \brief Overload of binary operator -=
 ///
 /// This operator performs a memberwise subtraction of both vectors,
-/// and assigns the result to \a left.
+/// and assigns the result to \c left.
 ///
 /// \param left  Left operand (a vector)
 /// \param right Right operand (a vector)
 ///
-/// \return Reference to \a left
+/// \return Reference to \c left
 ///
 ////////////////////////////////////////////////////////////
 template <typename T>
@@ -153,7 +287,7 @@ template <typename T>
 /// \param left  Left operand (a vector)
 /// \param right Right operand (a scalar value)
 ///
-/// \return Memberwise multiplication by \a right
+/// \return Memberwise multiplication by \c right
 ///
 ////////////////////////////////////////////////////////////
 template <typename T>
@@ -166,7 +300,7 @@ template <typename T>
 /// \param left  Left operand (a scalar value)
 /// \param right Right operand (a vector)
 ///
-/// \return Memberwise multiplication by \a left
+/// \return Memberwise multiplication by \c left
 ///
 ////////////////////////////////////////////////////////////
 template <typename T>
@@ -176,13 +310,13 @@ template <typename T>
 /// \relates Vector2
 /// \brief Overload of binary operator *=
 ///
-/// This operator performs a memberwise multiplication by \a right,
-/// and assigns the result to \a left.
+/// This operator performs a memberwise multiplication by \c right,
+/// and assigns the result to \c left.
 ///
 /// \param left  Left operand (a vector)
 /// \param right Right operand (a scalar value)
 ///
-/// \return Reference to \a left
+/// \return Reference to \c left
 ///
 ////////////////////////////////////////////////////////////
 template <typename T>
@@ -195,7 +329,7 @@ constexpr Vector2<T>& operator *=(Vector2<T>& left, T right);
 /// \param left  Left operand (a vector)
 /// \param right Right operand (a scalar value)
 ///
-/// \return Memberwise division by \a right
+/// \return Memberwise division by \c right
 ///
 ////////////////////////////////////////////////////////////
 template <typename T>
@@ -205,13 +339,13 @@ template <typename T>
 /// \relates Vector2
 /// \brief Overload of binary operator /=
 ///
-/// This operator performs a memberwise division by \a right,
-/// and assigns the result to \a left.
+/// This operator performs a memberwise division by \c right,
+/// and assigns the result to \c left.
 ///
 /// \param left  Left operand (a vector)
 /// \param right Right operand (a scalar value)
 ///
-/// \return Reference to \a left
+/// \return Reference to \c left
 ///
 ////////////////////////////////////////////////////////////
 template <typename T>
@@ -226,7 +360,7 @@ constexpr Vector2<T>& operator /=(Vector2<T>& left, T right);
 /// \param left  Left operand (a vector)
 /// \param right Right operand (a vector)
 ///
-/// \return True if \a left is equal to \a right
+/// \return True if \c left is equal to \c right
 ///
 ////////////////////////////////////////////////////////////
 template <typename T>
@@ -241,18 +375,13 @@ template <typename T>
 /// \param left  Left operand (a vector)
 /// \param right Right operand (a vector)
 ///
-/// \return True if \a left is not equal to \a right
+/// \return True if \c left is not equal to \c right
 ///
 ////////////////////////////////////////////////////////////
 template <typename T>
 [[nodiscard]] constexpr bool operator !=(const Vector2<T>& left, const Vector2<T>& right);
 
 #include <SFML/System/Vector2.inl>
-
-// Define the most common types
-using Vector2i = Vector2<int>;
-using Vector2u = Vector2<unsigned int>;
-using Vector2f = Vector2<float>;
 
 } // namespace sf
 
@@ -267,31 +396,40 @@ using Vector2f = Vector2<float>;
 /// sf::Vector2 is a simple class that defines a mathematical
 /// vector with two coordinates (x and y). It can be used to
 /// represent anything that has two dimensions: a size, a point,
-/// a velocity, etc.
+/// a velocity, a scale, etc.
 ///
+/// The API provides basic arithmetic (addition, subtraction, scale), as
+/// well as more advanced geometric operations, such as dot/cross products,
+/// length and angle computations, projections, rotations, etc.
+/// 
 /// The template parameter T is the type of the coordinates. It
 /// can be any type that supports arithmetic operations (+, -, /, *)
 /// and comparisons (==, !=), for example int or float.
-///
+/// Note that some operations are only meaningful for vectors where T is
+/// a floating point type (e.g. float or double), often because
+/// results cannot be represented accurately with integers.
+/// The method documentation mentions "(floating-point)" in those cases.
+/// 
 /// You generally don't have to care about the templated form (sf::Vector2<T>),
 /// the most common specializations have special type aliases:
 /// \li sf::Vector2<float> is sf::Vector2f
 /// \li sf::Vector2<int> is sf::Vector2i
 /// \li sf::Vector2<unsigned int> is sf::Vector2u
 ///
-/// The sf::Vector2 class has a small and simple interface, its x and y members
-/// can be accessed directly (there are no accessors like setX(), getX()) and it
-/// contains no mathematical function like dot product, cross product, length, etc.
+/// The sf::Vector2 class has a simple interface, its x and y members
+/// can be accessed directly (there are no accessors like setX(), getX()).
 ///
 /// Usage example:
 /// \code
-/// sf::Vector2f v1(16.5f, 24.f);
-/// v1.x = 18.2f;
-/// float y = v1.y;
+/// sf::Vector2f v(16.5f, 24.f);
+/// v.x = 18.2f;
+/// float y = v.y;
 ///
-/// sf::Vector2f v2 = v1 * 5.f;
-/// sf::Vector2f v3;
-/// v3 = v1 + v2;
+/// sf::Vector2f w = v * 5.f;
+/// sf::Vector2f u;
+/// u = v + w;
+///
+/// float s = v.dot(w);
 ///
 /// bool different = (v2 != v3);
 /// \endcode
--- a/include/SFML/System/Vector2.inl
+++ b/include/SFML/System/Vector2.inl
@@ -55,6 +55,128 @@ y(static_cast<T>(vector.y))
 
 ////////////////////////////////////////////////////////////
 template <typename T>
+T Vector2<T>::length() const
+{
+    static_assert(std::is_floating_point_v<T>, "Vector2::length() is only supported for floating point types");
+
+    return std::hypot(x, y);
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+constexpr T Vector2<T>::lengthSq() const
+{
+    return this->dot(*this);
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+Vector2<T> Vector2<T>::normalized() const
+{
+    static_assert(std::is_floating_point_v<T>, "Vector2::normalized() is only supported for floating point types");
+
+    assert(*this != Vector2<T>());
+    return (*this) / length();
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+Angle Vector2<T>::angleTo(const Vector2<T>& rhs) const
+{
+    static_assert(std::is_floating_point_v<T>, "Vector2::angleTo() is only supported for floating point types");
+
+    assert(*this != Vector2<T>());
+    assert(rhs != Vector2<T>());
+    return radians(std::atan2(this->cross(rhs), this->dot(rhs)));
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+Angle Vector2<T>::angle() const
+{
+    static_assert(std::is_floating_point_v<T>, "Vector2::angle() is only supported for floating point types");
+
+    assert(*this != Vector2<T>());
+    return radians(std::atan2(y, x));
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+Vector2<T> Vector2<T>::rotatedBy(Angle angle) const
+{
+    static_assert(std::is_floating_point_v<T>, "Vector2::rotatedBy() is only supported for floating point types");
+   
+    // No zero vector assert, because rotating a zero vector is well-defined (yields always itself)
+    T cos = std::cos(angle.asRadians());
+    T sin = std::sin(angle.asRadians());
+
+    // Don't manipulate x and y separately, otherwise they're overwritten too early
+    return Vector2<T>(
+        cos * x - sin * y,
+        sin * x + cos * y);
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+constexpr Vector2<T> Vector2<T>::projectedOnto(const Vector2<T>& axis) const
+{
+    static_assert(std::is_floating_point_v<T>, "Vector2::projectedOnto() is only supported for floating point types");
+
+    assert(axis != Vector2<T>());
+    return this->dot(axis) / axis.lengthSq() * axis;
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+constexpr Vector2<T> Vector2<T>::perpendicular() const
+{
+    return Vector2<T>(-y, x);
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+constexpr T Vector2<T>::dot(const Vector2<T>& rhs) const
+{
+    return x * rhs.x + y * rhs.y;
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+constexpr T Vector2<T>::cross(const Vector2<T>& rhs) const
+{
+    return x * rhs.y - y * rhs.x;
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+constexpr Vector2<T> Vector2<T>::cwiseMul(const Vector2<T>& rhs) const
+{
+    return Vector2<T>(x * rhs.x, y * rhs.y);
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+constexpr Vector2<T> Vector2<T>::cwiseDiv(const Vector2<T>& rhs) const
+{
+    assert(rhs.x != 0);
+    assert(rhs.y != 0);
+    return Vector2<T>(x / rhs.x, y / rhs.y);
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
 constexpr Vector2<T> operator -(const Vector2<T>& right)
 {
     return Vector2<T>(-right.x, -right.y);
@@ -159,3 +281,14 @@ constexpr bool operator !=(const Vector2<T>& left, const Vector2<T>& right)
 {
     return (left.x != right.x) || (left.y != right.y);
 }
+
+
+////////////////////////////////////////////////////////////
+// Static member data
+////////////////////////////////////////////////////////////
+
+template <typename T>
+constexpr Vector2<T> Vector2<T>::UnitX(static_cast<T>(1), static_cast<T>(0));
+
+template <typename T>
+constexpr Vector2<T> Vector2<T>::UnitY(static_cast<T>(0), static_cast<T>(1));
--- a/test/System/Vector2.cpp
+++ b/test/System/Vector2.cpp
@@ -4,7 +4,12 @@
 
 #include <doctest.h>
 
-// Use sf::Vector2i for tests. Test coverage is given, as there are no template specializations.
+using namespace sf::Literals;
+using doctest::Approx;
+
+// Use sf::Vector2i for tests (except for float vector algebra).
+// Test coverage is given, as there are no template specializations.
+
 
 TEST_CASE("sf::Vector2 class template - [system]")
 {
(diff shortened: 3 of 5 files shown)


ChrisThrasher:
> 1. Do we want to support sf::Vector<T> or only sf::Vector2f?

I'm inclined to support all template parameters. `auto` return types make this relatively easy. It's what I've been using in [my own implementation](https://github.com/ChrisThrasher/SFML/commit/c25448c1a687bf1b831b877eaa27fcaf5f160382#diff-448ddb42036c20195e875440d37de426afb34ebe5b21c8c8402edb25ce8a1857R57-R61). I'm fine with only supporting `sf::Vector2f` through. Presumably that would require some SFINAE magic to make sure that certain member functions aren't available when using a different type. I'm not totally sure if that's possible though. 

> 2. Should the methods be free or member functions?

The non-modifying functions can be either. I'm leaning towards making them member functions just because users will have an easier time finding them that way. We can have 

> 3. File structure

I want to avoid a situation where a user can't use `sf::dot` because they only included `SFML/Graphics.hpp` and not also `SFML/System/VectorUtils.hpp`. That will be confusing. Is `<cmath>` particularly expensive to parse? I thought those C headers had really low compiletime overhead. 

> 4. Which methods do we need and how do we name them?

I'll have think harder about this but would prefer shorter names like `cross` over `crossProduct` since `vector1.cross(vector2)` reads like English. 

> 5. Component-wise operators

Let's keep those operators as element-wise operations.

> 6. 3D vectors

Eventually I'd like to extend this to 3D vectors but that can wait for us to figure out 2D vectors first. That way we're not having to double all the work and rework we go through.

ChrisThrasher:
Do we want to make a final decision on `sf::Angle` before going too deep in this discussion? We might avoid some extra rework if we know the state of `sf::Angle` before spending too much time debating vector utilities that involve angles.

Bromeon:
Thanks for the feedback! I agree with most statements.

> I'm inclined to support all template parameters. `auto` return types make this relatively easy. It's what I've been using in [my own implementation](https://github.com/ChrisThrasher/SFML/commit/c25448c1a687bf1b831b877eaa27fcaf5f160382#diff-448ddb42036c20195e875440d37de426afb34ebe5b21c8c8402edb25ce8a1857R57-R61). I'm fine with only supporting `sf::Vector2f` through. Presumably that would require some SFINAE magic to make sure that certain member functions aren't available when using a different type. I'm not totally sure if that's possible though.

Good that you mention this, I wasn't aware of the `double sqrt(int)` overload. My question is more: do we need this? `double` is a type that's incredibly absent from SFML, and we have conversion constructors (e.g. `Vector2i` -> `Vector2f`) if we really need the length of an `int` vector.

A `float`-centered API might be a bit easier to understand. It also doesn't lead to template errors if `T` doesn't support any of the operations used by the implementation. Is anyone even using any `Vector2<T>` instantiation besides the 3 typedefs?

Replacing a parameter type with a template parameter should be a non-breaking change, so maybe this can still be done later if enough people request it? _(I'm sure it's possible to construct a weird case in C++'s name lookup where this doesn't apply, but in general...)_

> I want to avoid a situation where a user can't use `sf::dot` because they only included `SFML/Graphics.hpp` and not also `SFML/System/VectorUtils.hpp`. That will be confusing. Is `<cmath>` particularly expensive to parse? I thought those C headers had really low compiletime overhead. 

Not sure about `<cmath>`, would need measuring. Yes, I'm also for self-contained vector utilities. Both `<SFML/Graphics.hpp>` and `<SFML/System/Vector2.hpp>` should include the functionality discussed here.

> Let's keep those operators as element-wise operations.

You mean named methods or operators?

ChrisThrasher:
> My question is more: do we need this? double is a type that's incredibly absent from SFML,

`sf::Vector2i::mag()` returning a `double` is less confusing than `sf::Vector2i::mag()` not existing. It's annoying that we can't have that template parameter return a `float` but such are the realities of template meta programming.

> It also doesn't lead to template errors if T doesn't support any of the operations used by the implementation.

This is a feature though. We want our templates to ideally support any type that meets the requirements of the implementation. If you pass in `std::string` and it doesn't compile, that's a good thing.

> You mean named methods or operators?

I meant that we can keep the existing `operator*` for `Vector2<T>`. No need to add named versions of them either.

Bromeon:
> `sf::Vector2i::mag()` returning a `double` is less confusing than `sf::Vector2i::mag()` not existing. It's annoying that we can't have that template instantiation return a `float` but such are the realities of template meta programming.

I don't think `auto` is very helpful, the reader of the doc still doesn't know the return type. We could let it return `float` by going through a type metafunction.
```cpp
template <typename T>
struct VectorTypes
{
    typedef T Length;
};

template <>
struct VectorTypes<int>
{
    typedef float Length;
};

// and then:
template <typename T>
struct Vector2
{
   typename VectorTypes<T>::Length length();
};
```
And this is exactly where the over-engineering begins 🙂
But more on this below.

---

> This is a feature though. We want our templates to ideally support any type that meets the requirements of the implementation. 

Is it though? The primary reason why `sf::Vector2` is a template, is so that we don't have to duplicate code for `float`, `int` and `unsigned int`. SFML could easily have added a `Vector2d` or `Vector2ld` typedef, but it didnt. The CSFML binding has three separate structs, which is enough to cover the APIs.

I have not encountered code which uses different instantiations -- and instead of adding genericity we might not need, I would evolve APIs more naturally: start small and add when there is real demand. Adding can be done in a (mostly) non-breaking way later, removing is impossible.

And the above situation, where some operations make sense for floating-point `T` but not for integral `T`, **is a clear indication that the `Vector2<T>` abstraction is breaking down**.

We need to either:
* Provide functions operating on `Vector2f` as free functions, so they're not constrained by the type itself.
* Specialize it for different types, which can be confusing if the provided API is quite different (as you noted in your reply).
* Forcibly provide the same methods on integral vectors, even if they create an inconsistent mixture, where half the operations yields `T` and the other half `double`.
* Split `Vector2<T>` into `FloatVector2<T>` and `IntVector2<T>`.
  Or even just `Vector2f` (class) and `IntVector2<T>` (class template).

The last option means some code duplication (depending on how we write it), but might make most sense conceptually. Specialization would be a compromise, but I'm not sure if a) differing APIs could be conveyed in the docs in a user-friendly way, and b) there's actually a good reason for vectors to be a single class template.

---

> I meant that we can keep the existing `operator*` for `Vector2<T>`. No need to add named versions of them either.

Still not sure if I understand correctly -- there is no existing `operator*` taking two vector operands, only vector and scalar. Multiplying and dividing vectors component-wise is a new addition. So I think you'd prefer operators here?

ChrisThrasher:
You make good points that make me question some of my original points. I feel like the easiest place to start is with free functions like `sf::dot(const sf::Vector2f&, const sf::Vector2f&)`, `sf::mag(const sf::Vector2f&)`, and `sf::heading(const sf::Vector2f&)` which only apply to `Vector2f`. These are easy to implement and test yet still add value to anyone working with vectors. This approach lets us get some of the basic work behind us and settle questions like how to name various operations so that future PRs which tackle the harder problems require a little less bike shedding.

ChrisThrasher:
https://github.com/ChrisThrasher/SFML/commit/35120e640acd46028b3c1f5def9f6875f93469f6

This is what I had in mind. Just some simple free functions using `Vector2f`. I think I also want to add a squared magnitude function since that's so useful for comparing the length of angles without incurring the cost of a square root.

vittorioromeo:
> ### 1. Do we want to support `sf::Vector<T>` or only `sf::Vector2f`?
> In Thor, I created `TrigonometricTraits` allowing generic extensions to support any type `T` (such as custom decimal). In the scope of SFML, I think that would be over-engineering. The entire SFML API accepts only `sf::Vector2f` for floating point vectors and there are no typedefs for `double` or `long double`, so I think it's reasonable to limit the extensions to `float`.
> 
> Almost all extension methods are meaningless for integer types; even those theoretically computable (dot product) will in practice be applied to float vectors.

We definitely want to support `sf::Vector<T>` not just `sf::Vector2f`. There are valid cases for vectors of doubles or for applying transformations to vectors of integral types. Only supporting `sf::Vector2f` doesn't make much sense to me.



> ### 2. Should the methods be free or member functions?
> There are arguments for both, so even a mixture might be possible:
> 
> * Symmetry: `sf::dot(vec0, vec1)`
> * Clarity: `vec.rotatedBy(angle)`

Free functions are good for encapsulation, but the truth is that they're painful to use. If go down the free function route, I would strongly recommend having all vector utilities in a namespace/class, to avoid name collisions and for organization purposes. User code would then become too verbose.

I would recommend using member functions here, even if free functions can do the job. Member functions are more readable and easier to chain/nest. I could be convinced otherwise, though... We could also do something like:

```cpp
sf::Vector2f myVec{10.f, 20.f};
myVec = sf::VectorOperations{myVec}.rotateBy(sf::degrees(5)).scaleBy(5.f);
```

That would allow us to have member syntax (reads left-to-right), chaining, and also completely separate these functions from `sf::Vector2` itself. (See point below).



> ### 3. File structure
> Should we add the functionality to `Vector2.hpp` or a separate header? It probably makes sense if the `Vector2` related functionality is self-contained. I also wanted to keep `<cmath>` out of the header, since `Vector2.hpp` is such a heavily depended-upon file.

I would suspect most uses of `sf::Vector2` to end up dealing with some math sooner or later, so I don't think that the `<cmath>` dependency would be too much of a problem, but we'd have to measure it. The `sf::VectorOperations` approach seen above would avoid the dependency.



> ### 4. Which methods do we need and how do we name them?
> I adopted the current set from Thor and found them useful for my own games, but opinions can differ.
> 
> Naming is also a topic, but it depends on point 2, so we should first decide if we go for free functions, methods or both. In general, one thing that would be super-nice are relatively short names. Doesn't have to be extreme like `len()`, but maybe `cross()` instead of `crossProduct()` etc. Keep in mind that those operations are often part of larger algebraic expressions.

Don't have anything concrete here, but I'd like to remind everyone that we can start with a very small subset, and add new operations over time.


> ### 5. Component-wise operators
> Having `vec0 * vec1` and `vec0 / vec1` for component-wise multiplication/division is useful for scaling. Most vector libraries I have encountered over the years implement this, and I think the risk for misunderstanding (as dot or cross product) is quite small. It's also possible to provide a named function for it, at slightly increased verbosity cost.

I would prefer named functions here, I generally dislike ambiguity when reading code.


> ### 6. 3D vectors
> Should very limited support for 3D vectors be added (dot/cross product, angle, etc)? If yes, I'll gladly provide my Thor implementation, but this could also be done at a later stage.

Later stage sounds good to me.

Bromeon:
Thanks also to you for the elaborate answer! 🙂 

---

> We definitely want to support `sf::Vector<T>` not just `sf::Vector2f`. There are valid cases for vectors of doubles or for applying transformations to vectors of integral types. Only supporting `sf::Vector2f` doesn't make much sense to me.

The problem is, integer vectors are really not designed to be used with a large part of the operations we're talking about.
Consider:
```cpp
v.setLength(len);
sf::Vector2i w = v.normalized();
sf::Vector2i w = v.rotatedBy(angle);
sf::Vector2i u = v.projectedOnto(w);
```
None of the above can be implemented correctly for `Vector2i`.
These are floating-point operations at their very core.

Limiting float vectors to only the "get" methods is a compromise that significantly limits the usefulness of this feature.
And it does so only to provide functionality to integer vectors, which is a) rarely used (if at all) and b) can _always_ be trivially achieved by `sf::Vector2f(intVector).operation()`. Since its so easy to convert vectors, I really don't see why int vectors should have these operations, when half of them cannot be properly implemented, and results are surprising at best.

I'm also open to make this use case _even_ nicer to use, e.g. `intVector.toFloat().operation()` or similar.

---

As mentioned, this shows a deeper issue: integer and float vectors are **different concepts**. The right thing to do in my opinion would be two separate classes. The one-fits-all approach is done for code reuse, we don't use vectors generically in SFML (which is why I wouldn't use specialization). And we see the limitation it brings.

Maybe consider it from a first-principle point of view -- if we didn't have SFML 2 (which conveniently bypassed this issue by not providing extension functions 😉), but needed to design a vector library from scratch.

The Rust vector library [glam](https://github.com/bitshifter/glam-rs) is a modern example of a quite well-designed vector API for gamedev, in my opinion. It has types [`Vec2`](https://docs.rs/glam/0.20.2/glam/f32/struct.Vec2.html) and [`IVec2`](https://docs.rs/glam/0.20.2/glam/i32/struct.IVec2.html) for example. By design, `IVec2` lacks all the operations which make no sense, such as `distance()`, `length()`, `normalize()`, 

---

> Free functions are good for encapsulation, but the truth is that they're painful to use. [...] I would recommend using member functions here, even if free functions can do the job. Member functions are more readable and easier to chain/nest. 

I tend to agree here. For things like `dot(a, b)` ergonomics may be similar, but other expressions like `v.projectedOnto(w)`, `v.rotatedBy(angle)`, `v.length()` are generally nicer to use.

With a separate `Vector2f` class, this would be no problem to implement.

---

> I would suspect most uses of `sf::Vector2` to end up dealing with some math sooner or later, so I don't think that the `<cmath>` dependency would be too much of a problem, but we'd have to measure it. The `sf::VectorOperations` approach seen above would avoid the dependency.

Yes, probably it's OK to have that dependency.

If we split integer and float vectors, we could keep the former basic and without `<cmath>`, while the latter can include it.

vittorioromeo:
> The problem is, integer vectors are really not designed to be used with a large part of the operations we're talking about. [...]

I see. I would be comfortable with having a `sf::Vector2<T>` and `sf::IVector2<T>` with different interfaces/requirements, and implementing extensions only for the floaty one. I am still against specifically implementing extensions only for `sf::Vector2f`.

Do we all agree that we want to try out that route? @eXpl0it3r @binary1248 

Also, it seems that we have ~7 PRs open for vector extensions, all of them from different authors. What should we do?

binary1248:
Sometimes I have the feeling that we haven't gotten used to the fact that we have finally landed in C++17+ land. 😄

The capabilities of the compiler to do our work for us and protect the developer from their own mistakes _at compile time_ has grown by an order of magnitude over the last 20 years.

If we want to prevent operations that make no sense e.g. `sf::Vector2<int>::normalized()`, we just have to communicate to the compiler that such operations shouldn't be allowed to be instantiated for non-floating-point types via `std::is_floating_point`, `std::enable_if` and friends. This is one of those areas where C++ concepts would make perfect sense, but I wouldn't make the jump to C++20 just yet. 😄

ChrisThrasher:
> Also, it seems that we have ~7 PRs open for vector extensions, all of them from different authors. What should we do?

@vittorioromeo We've talked about this a lot in Discord. @eXpl0it3r has even done some deeper research into what was going on. Those PR authors were dumping their college coursework onto this project. None of them have put any effort into refining their PRs or acknowledging our comments. It's safe to say we may delete these PRs.

Bromeon:
> Sometimes I have the feeling that we haven't gotten used to the fact that we have finally landed in C++17+ land. 😄
> 
> The capabilities of the compiler to do our work for us and protect the developer from their own mistakes _at compile time_ has grown by an order of magnitude over the last 20 years.
> 
> If we want to prevent operations that make no sense e.g. `sf::Vector2<int>::normalized()`, we just have to communicate to the compiler that such operations shouldn't be allowed to be instantiated for non-floating-point types via `std::is_floating_point`, `std::enable_if` and friends. This is one of those areas where C++ concepts would make perfect sense, but I wouldn't make the jump to C++20 just yet. 😄

SFINAE is _very_ ugly though, and leads to Boost-like APIs and error messages. It's mostly helpful to remove candidates from overload resolution without causing errors (the "NAE" part). Since we don't really have overloads, if we want to go down that road, we should probably just do
```cpp
static_assert(std::is_floating_point_v<T>, "Operation is only supported for floating point types");
```

But I still wonder if it's the right approach. These methods will all appear in the documentation and in IDE auto-completion, so it involves more trial&error from the user. There's no easy way to list only the int methods, for example.

I've been thinking about alternatives -- on one hand classic template specialization, but also conditional inheritance. Specialization still needs code duplication, and conditional inheritance is very annoying to implement, and likely confusing in the doc too.

---

So, does `Vector2<T>` need to be a template at all? It provides two main benefits:
* Code reuse (mostly for basic arithmetic operations)
* Genericity (APIs that work with `Vector2<T>`)
   * however, we don't really use this in SFML
   * if we really need it, we can still do `scale(V vec, T scalar) { return vec * scalar; }` with slightly less type safety

Avoiding code duplication is a valid goal, but it seems to me it's one of the "library author inconvenience vs. user inconvenience" trade-offs, where I generally tend to favor the former.

glam-rs solved the code duplication issues [with code-generation macros](https://github.com/bitshifter/glam-rs/blob/e5a183a53a8037f4ce6c608ee3dd8c130d862507/src/vec2.rs#L190) (i.e. metaprogramming). That wouldn't be very nice in C++ though, and likely also confuse Doxygen...

---

Another topic: if we really want to keep things generic, we should probably write code like:
```cpp
T length() const
{
    using std::sqrt;
    return sqrt(x*x + y*y); // ADL
}
```

vittorioromeo:
> SFINAE is _very_ ugly though, and leads to Boost-like APIs and error messages. It's mostly helpful to remove candidates from overload resolution without causing errors (the "NAE" part). Since we don't really have overloads, if we want to go down that road, we should probably just do

I agree that SFINAE is not great, and it is one of the slowest things for compilers to work with (see Rule of Chiel). I think that `static_assert` is a pragmatic compromise between making the library maintainable and improving the user experience. A line of documentation and a proper error string will make it really obvious to everyone that such functions can only be used with floating point vectors.

ChrisThrasher:
Heads up, a lot of newly added code is indented with tabs instead of spaces. GitHub renders those tabs as 8 characters so they're easy to spot if you view the diff online.

Bromeon:
Went ahead with the `static_assert` approach for now. I added a few commits for easy selecting, will squash in the end.
Once the API is agreed-upon, I can write some unit tests.

Design:
* everything is a member function
* some methods are limited to floating-point via `static_assert`, those are also non-`constexpr`able
* tried to keep names short (`dotProduct -> dot` etc.), since those operations often happen in more verbose expressions
* every method is `const` and returns a new object
* new constants `UnitX` and `UnitY`. I was considering to call them `Right` and `Down` for readability, but that may not always be true.
* `cwiseMul()` + `cwiseDiv()` as named methods; not sure if I'm super-happy with this. It likely makes computations less readable.

I initially had methods that could set length/angle and return a new vector, but realized that with the constants, those can be emulated relatively easily:
* `v.withLength(len)` --> `len * v.normalized()`
* `v.withPolarAngle(phi)` --> `Vector2::UnitX.rotatedBy(phi)`

API looks like this:
grafik

include/SFML/System/Vector2.inl
```diff
@@ -53,6 +53,138 @@ y(static_cast<T>(vector.y))
...
+Vector2<T> Vector2<T>::rotatedBy(Angle angle) const
+{
+    static_assert(std::is_floating_point_v<T>, "Vector2::rotatedBy() is only supported for floating point types");
+        
+    using std::sin; // allow ADL
+    using std::cos;
+
+    // No zero vector assert, because rotating a zero vector is well-defined (yields always itself)
+    T c = cos(angle.asRadians());
+    T s = sin(angle.asRadians());
```
ChrisThrasher, line 127: I've considered a few times whether defining `sf::sin` and `sf::cos` for `sf::Angle` makes sense. It would allow easy ADL usage of `Angle`s and remove many more `asRadians()` calls.

include/SFML/System/Vector2.inl
```diff
@@ -53,6 +53,138 @@ y(static_cast<T>(vector.y))
...
+template <typename T>
+Angle Vector2<T>::polarAngle() const
+{
+    static_assert(std::is_floating_point_v<T>, "Vector2::polarAngle() is only supported for floating point types");
+
+    using std::atan2; // allow ADL
+
+    assert(*this != Vector2<T>());
+    return radians(atan2(y, x));
+}
```
ChrisThrasher, line 113: Would a better name for this be `toAngle` or `asAngle`? That might read easier and better matches other functions that start with verbs.

include/SFML/System/Vector2.inl
```diff
@@ -53,6 +53,138 @@ y(static_cast<T>(vector.y))
...
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+Vector2<T> Vector2<T>::normalized() const
+{
+    static_assert(std::is_floating_point_v<T>, "Vector2::normalized() is only supported for floating point types");
+
+    assert(*this != Vector2<T>());
+    return (*this) / length();
```
ChrisThrasher, line 85: `operator*` has higher precedence than `operator/` so the extra parentheses could be removed if you wanted to.

include/SFML/System/Vector2.inl
```diff
@@ -53,6 +53,138 @@ y(static_cast<T>(vector.y))
...
 
+////////////////////////////////////////////////////////////
+template <typename T>
+T Vector2<T>::length() const
+{
+    static_assert(std::is_floating_point_v<T>, "Vector2::length() is only supported for floating point types");
+
+    using std::sqrt; // allow ADL
+
+    return sqrt(lengthSq());
```
ChrisThrasher, line 64: No issues here, just commenting on how satisfying it is to see these helpers used to implement other helpers.

include/SFML/System/Vector2.inl
Bromeon: I would say that's a good problem to solve once we actually have it 😉 

I'm not even sure if many people use `Vector<double>` for example. And I've never seen anyone use custom types. I completely overengineered that in Thor already, probably for nothing. The ADL-enabled trigonometric functions are likely good enough, if there's need for it at all.

include/SFML/System/Vector2.inl
Bromeon: This suggests however that the vector is an angle or representable as angle, like with `Time::asSeconds()` or `Angle::asDegrees()`. But this is not the case.

(btw, "as"/"to" are not verbs 🙂)

include/SFML/System/Vector2.inl
ChrisThrasher: Right, not a problem for this PR, just a thought I wanted to share. I might make a PR for that later if I can find good motivating examples to add those extra free functions.

include/SFML/System/Vector2.inl
Bromeon: I had it without parentheses first, thought might add it for readability. I don't really mind, any way is good.

include/SFML/System/Vector2.inl
Bromeon: Ah, sorry I totally misread. You were talking about `sf::Angle`, I was understanding that we extend the trigonometry, so that user can specialize it for their own types.

A first great addition would actually be `sf::Angle::Pi` and `sf::Angle::Tau`.

include/SFML/System/Vector2.inl
ChrisThrasher: > (btw, "as"/"to" are not verbs 🙂)

Go easy on me. I've only been speaking and writing English for 27 years.

What about `Vector2<T>::heading`? Heading is defined as the angle relative to X axis. That's already what this implementation is doing.

include/SFML/System/Vector2.hpp
```diff
@@ -68,14 +73,132 @@ class Vector2
...
+    ////////////////////////////////////////////////////////////
+    T lengthSq() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Vector with same direction but length 1 <i><b>(floating-point)</b></i>.
+    /// 
+    /// \pre \c vector is no zero vector.
+    ///
+    ////////////////////////////////////////////////////////////
+    Vector2 normalized() const;
```
vittorioromeo, line 99:
```suggestion
    [[nodiscard]] Vector2 normalized() const;
```

Probably reasonable to add it here to avoid people thinking that it normalizes the vector in place. The name is quite obvious, but still prone to misinterpretation or typos (imagine a non-native English speaker).

include/SFML/System/Vector2.hpp
```diff
@@ -68,14 +73,132 @@ class Vector2
...
+    ////////////////////////////////////////////////////////////
+    Angle polarAngle() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Rotate by given angle <i><b>(floating-point)</b></i>.
+    /// 
+    /// The vector (1,0) corresponds 0 degrees, (0,1) corresponds 90 degrees.
+    ///
+    ////////////////////////////////////////////////////////////
+    Vector2 rotatedBy(Angle angle) const;
```
vittorioromeo, line 129:
```suggestion
    [[nodiscard]] Vector2 rotatedBy(Angle angle) const;
```

Ditto here. I don't feel strongly about it, but wouldn't hurt...

include/SFML/System/Vector2.hpp
```diff
@@ -68,14 +73,132 @@ class Vector2
...
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Projection of \c vector onto \c axis <i><b>(floating-point)</b></i>.
+    /// 
+    /// \param vector Vector to project.
+    /// \param axis Vector being projected onto. Need not be normalized, but must not have length zero.
+    ///
+    ////////////////////////////////////////////////////////////
+    Vector2 projectedOnto(const Vector2& axis) const;
```
vittorioromeo, line 138:
```suggestion
    [[nodiscard]] Vector2 projectedOnto(const Vector2& axis) const;
```

Ditto.

include/SFML/System/Vector2.hpp
```diff
@@ -68,14 +73,132 @@ class Vector2
...
+    ////////////////////////////////////////////////////////////
+    T length() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Square of vector's length <i><b>(floating-point)</b></i>.
+    /// 
+    /// Suitable for comparisons, more efficient than length().
+    ///
+    ////////////////////////////////////////////////////////////
+    T lengthSq() const;
```
vittorioromeo, line 91: Can't this be `constexpr`?

include/SFML/System/Vector2.hpp
vittorioromeo: Can't this be `constexpr` if we make `lengthSq` so?

include/SFML/System/Vector2.hpp
```diff
@@ -68,14 +73,132 @@ class Vector2
...
+    ////////////////////////////////////////////////////////////
+    constexpr T cross(const Vector2& rhs) const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Component-wise multiplication of \c *this and \c rhs.
+    /// 
+    /// Computes <tt>(lhs.x*rhs.x, lhs.y*rhs.y)</tt>. Main use case are scales.
+    ///
+    ////////////////////////////////////////////////////////////
+    constexpr Vector2 cwiseMul(const Vector2& rhs) const;
```
vittorioromeo, line 170: I don't mind this name, but maybe consider `compWiseMul`?

include/SFML/System/Vector2.inl
```diff
@@ -53,6 +53,138 @@ y(static_cast<T>(vector.y))
 }
 
 
+////////////////////////////////////////////////////////////
+template <typename T>
+T Vector2<T>::length() const
+{
+    static_assert(std::is_floating_point_v<T>, "Vector2::length() is only supported for floating point types");
+
+    using std::sqrt; // allow ADL
```
vittorioromeo, line 62: Are we doing this in case of a custom `T` used for `Vector2<T>` that has its own sqrt implementation?

include/SFML/System/Vector2.inl
```diff
@@ -53,6 +53,138 @@ y(static_cast<T>(vector.y))
 }
 
 
+////////////////////////////////////////////////////////////
+template <typename T>
+T Vector2<T>::length() const
+{
+    static_assert(std::is_floating_point_v<T>, "Vector2::length() is only supported for floating point types");
```
vittorioromeo, line 60: The message here is not very helpful, it is redundant with the type trait predicate itself, which does appear as part of compiler diagnostics. I suggest explaining *why* instead:

```suggestion
    static_assert(std::is_floating_point_v<T>, "Calculating the length of an integer vector is a lossy operation");
```

Or something like that.

include/SFML/System/Vector2.inl
vittorioromeo: If so, `std::is_floating_point<T>` would fail. I don't really see the point...

include/SFML/System/Vector2.inl
vittorioromeo: I like the parentheses. I don't want to have to think about operator precedence when I read the code :)

include/SFML/System/Vector2.inl
```diff
@@ -53,6 +53,138 @@ y(static_cast<T>(vector.y))
...
+{
+    return Vector2<T>(x * rhs.x, y * rhs.y);
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+constexpr Vector2<T> Vector2<T>::cwiseDiv(const Vector2<T>& rhs) const
+{
+    assert(rhs.x != 0 && rhs.y != 0);
```
vittorioromeo, line 183: Splitting this in two separate assertions will produce better diagnostics.

include/SFML/System/Vector2.inl
```diff
@@ -159,3 +291,18 @@ constexpr bool operator !=(const Vector2<T>& left, const Vector2<T>& right)
...
+// Static member data
+////////////////////////////////////////////////////////////
+
+// Note: the 'inline' keyword here is technically not required, but VS2019 fails
+// to compile with a bogus "multiple definition" error if not explicitly used.
+template <typename T>
+inline constexpr Vector2<T> Vector2<T>::UnitX(static_cast<T>(1), static_cast<T>(0));
+
+// Note: the 'inline' keyword here is technically not required, but VS2019 fails
+// to compile with a bogus "multiple definition" error if not explicitly used.
```
vittorioromeo, line 306: For consistency with other files, we only repeat this comment once at the beginning, not in every `inline constexpr` variable.

Furthermore, `inline` might not be required here as it is implied by `template`, maybe VS2017 will be happy with it without `inline`... give it a try.

vittorioromeo (changes requested):

include/SFML/System/Vector2.inl
Bromeon: I was under the impression that `is_floating_point` can be specialized, but [cppref](https://en.cppreference.com/w/cpp/types/is_floating_point) says:
> The behavior of a program that adds specializations for is_floating_point or is_floating_point_v (since C++17) is undefined. 

So you're right, I'll revert this.

include/SFML/System/Vector2.hpp
Bromeon: This is even more verbose, while not really being much more expressive. I'd rather consider overloading `*` and `/` directly, like it's done in every other vector library 🙂

include/SFML/System/Vector2.hpp
Bromeon: Good point!

include/SFML/System/Vector2.inl
Bromeon: It's not redundant, the default error message is one of those that make C++ (in)famous, at least on VS 2019. I only have a German one installed on this machine, but it basically says "when compiling the class template member function", i.e context only.
```cpp
1>{path}/SFML/System/Vector2.inl(60,24): error C2338: Vector2::length() is only supported for floating point types
1>{path}/SFML/System/Vector2.inl(59): message : beim Kompilieren der Klasse Vorlage-Memberfunktion "T sf::Vector2<T>::length(void) const"
1>        with
1>        [
1>            T=int
1>        ]
1>{path}/examples\tennis\Tennis.cpp(33): message : Siehe Verweis auf die gerade kompilierte Instanziierung "T sf::Vector2<T>::length(void) const" der Funktions-Vorlage.
1>        with
1>        [
1>            T=int
1>        ]
1>{path}/SFML/Graphics/Image.hpp(217): message : Siehe Verweis auf die gerade kompilierte Klasse Vorlage-Instanziierung "sf::Vector2<int>".
1>Die Erstellung des Projekts "tennis.vcxproj" ist abgeschlossen -- FEHLER.
========== Erstellen: 0 erfolgreich, 1 fehlerhaft, 5 aktuell, 0 übersprungen ==========
```

I think the "why" part makes sense to mention, but maybe rather in the documentation? The reason is the same for all the methods: they cannot be meaningfully/correctly implemented for integers.

include/SFML/System/Vector2.inl
Bromeon: Didn't know about this method! It might mean however that the `x*x + y*y` part can no longer be computed at compile time, but that's going toward micro-optimization territory.

I'll adjust this 🙂

include/SFML/System/Vector2.hpp
```diff
@@ -68,14 +73,132 @@ class Vector2
...
+    /// \param vector Vector to project.
+    /// \param axis Vector being projected onto. Need not be normalized, but must not have length zero.
+    ///
+    ////////////////////////////////////////////////////////////
+    Vector2 projectedOnto(const Vector2& axis) const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Returns a perpendicular vector.
+    /// 
+    /// Returns \c vector rotated by 90 degrees counter clockwise; (x,y) becomes (-y,x).
```
kimci86, line 143: > counter clockwise

It does not really make sense to me. That depends on the coordinate system.
Maybe using the word "positive" would be better?

Bromeon: You're right, I'll change it.

The majority of SFML code uses the convention where +X is right and +Y is down; should we mention this somewhere? Or remain completely agnostic to the coordinate system, but also make things a bit harder to the user?

Maybe `sf::Vector2` doc is not the right place for the coordinate system conventions; rather the tutorials or so.

Bromeon: I added a line "in SFML's default coordinate system, ..." to make things explicit.

include/SFML/System/Vector2.inl
Bromeon: Seems to work without `inline` in VS 2019.

include/SFML/System/Vector2.hpp
kimci86: Good, but it is clockwise then. :wink:

Bromeon: You got me there 😬 thanks!

include/SFML/System/Vector2.inl
Bromeon: One alternative would also be to just call the method `phi()`.

It's an often-used variable name of the angle in math contexts. And I think in any case, we will need doc to be 100% clear.

ChrisThrasher: Yes it's a common term in my professional field of robotics which uses linear algebra and coordinate systems quite often. Heading is defined relative to the X axis. North is sometimes the X axis (see NED coordinate frames) but sometimes East is the X axis (see ENU coordinate frames). Many coordinate frames have no notion of north or east though since some frames are attached to moving vehicles. The only assumption is that we're using a [right handed frame](https://en.wikipedia.org/wiki/Right-hand_rule), which is the conventional way to reason about coordinate frames.

vittorioromeo:
I am quite happy with the state of this PR. What other changes are you planning, if any?

include/SFML/System/Vector2.hpp
```diff
@@ -68,14 +73,138 @@ class Vector2
...
+    ////////////////////////////////////////////////////////////
+    /// \brief Angle from +X or (1,0) vector <i><b>(floating-point)</b></i>.
+    /// 
+    /// The vector (1,0) corresponds to 0 degrees, (0,1) corresponds to 90 degrees.
+    /// 
+    /// \return Angle in degrees in the interval [-180,180].
+    /// \pre This vector is no zero vector.
+    ///
+    ////////////////////////////////////////////////////////////
+    Angle polarAngle() const;
```
ChrisThrasher, line 121: I think it's more accurate to say this returns merely an angle or if you want to be explicit
```
/// \return Angle in the range [-180°, 180°)
```
This also applies to the comments for `singleAngleTo`

include/SFML/System/Vector2.hpp
```diff
@@ -68,14 +73,138 @@ class Vector2
...
+    ////////////////////////////////////////////////////////////
+    /// \brief Signed angle from \c *this to \c rhs <i><b>(floating-point)</b></i>.
+    /// 
+    /// \return Angle in degrees in the interval [-180,180].
+    /// The angle determines how much you have to rotate \c *this
+    ///  in either direction, until it is parallel to \c rhs.
+    /// \pre Neither \c *this nor \c rhs is a zero vector.
+    ///
+    ////////////////////////////////////////////////////////////
+    Angle signedAngleTo(const Vector2& rhs) const;
```
ChrisThrasher, line 110: Calling this `signedAngleTo` insinuates that there might be a different overload returning the unsigned angle when that is not the case. We can refer to angles as simply angles and don't need extra qualifiers on them. The `Angle` abstraction handles messy detail like signed/unsigned and degrees/radians. For good reason we don't say "time in microseconds" because a time is merely a time. We don't need to also speak to a specific detail of the `sf::Time` abstraction. Let the abstractions be abstract.

include/SFML/System/Vector2.hpp
Bromeon: see other comment(s)

include/SFML/System/Vector2.inl
Bromeon: Thanks for the input! 

Given my argument about `signedAngleTo()`, we might want to clarify the signed-ness here too.

Bromeon:
Thanks for all the feedback so far!

Changes:
* Refined doc
* API changes
  * Rename `polarAngle()` -> `angle()`
  * Rename `signedAngleTo()` -> `angleTo()`
  * Make `lengthSq()` usable for integer vectors
* Unit tests
  * Test all methods added in this PR
  * Discovered idiom `ApproxVec` for better readability; we could consider using this elsewhere
* Squashed to 2 commits (tests + approx kept separate)

From my side it should be ready.

kimci86:
Regarding unit tests, I came up with a slightly different idiom when trying to simplify `sf::Transform` tests in #1976 : a template `Approx` wrapper class which defines the output operator using the underlying type output operator and comparison using a `almostEqual` function.
When a new type needs approximate comparison, we would need to define only `almostEqual` for that type and the output operator for the wrapper `Approx` type would come for free. See [Approx.hpp](https://github.com/SFML/SFML/blob/2da4f8c8de1882a0d45e01a402486486b137336b/test/TestUtilities/Approx.hpp). What do you think?

Bromeon:
@kimci86 That looks nice, I like that it reuses the underyling `operator<<` and just needs an `almostEqual()` function.

Direct comparison:
```cpp
CHECK(v.normalized() == ApproxVec(0.624695, 0.780869));
// vs.
CHECK(v.normalized() == Approx(sf::Vector2f(0.624695f, 0.780869f));
```
```cpp
CHECK(w.angleTo(v) == ApproxDeg(158.9902));
// vs.
CHECK(w.angleTo(v) == Approx(sf::degrees(158.9902)); // or
CHECK(w.angleTo(v) == Approx(158.9902_deg);
```

My observations: 
* Angles with user-defined literals are very concise.
* The vector case is _slightly_ more wordy with your approach -- something that could be addressed with an emplace-like parameter list + typedef (if we want to do that).
* Another small difference are the `float` vs. `double` literals. The `doctest::Approx` takes `double`, so I tried to stay close to that; it also makes inspection of hardcoded values slightly faster without the `f` suffix.
* A custom `Approx*` type could have different constructor than the underlying type, one specifically optimized for easy construction of testing values. This could also be done with generic `Approx<T>` + free factory functions, though.

All in all I think the differences are minor. To also decrease the PR inter-dependence, should I go ahead with the current approach for now and we can revisit these two idioms? Maybe as we write more tests, we find out which pattern works better.

include/SFML/System/Vector2.hpp
```diff
@@ -68,14 +73,143 @@ class Vector2
...
+    /// \pre \c *this is no zero vector.
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] Vector2 normalized() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Signed angle from \c *this to \c rhs <i><b>(floating-point)</b></i>.
+    /// 
+    /// \return The smallest angle which rotates \c *this in positive
+    /// or negative direction, until it is parallel to \c rhs.
```
kimci86, line 105: > parallel

Parallel can also be in the opposite direction.

Bromeon: Your eagle eyes again 🦅 thanks!

kimci86:
> All in all I think the differences are minor. To also decrease the PR inter-dependence, should I go ahead with the current approach for now and we can revisit these two idioms? Maybe as we write more tests, we find out which pattern works better.

Sure, feel free to go with the current approach. This can be revisited later.

include/SFML/System/Vector2.hpp
Bromeon: Fixed.

vittorioromeo (approved):
LGTM

test/System/Vector2.cpp
```diff
@@ -184,11 +189,97 @@ TEST_CASE("sf::Vector2 class template - [system]")
         }
     }
 
+    SUBCASE("Length and normalization")
+    {
+        sf::Vector2f v(2.4f, 3.0f);
+
+        CHECK(v.length() == Approx(3.84187));
+        CHECK(v.lengthSq() == Approx(14.7599650969));
+        CHECK(v.normalized() == ApproxVec(0.624695, 0.780869));
```
ChrisThrasher, line 198: I've made a point to mark such local variables as `const` in tests because it allows us to test that all the functions we're calling are marked as `const` as well. This fact is obvious from reading the code but I think it's worthwhile to maintain const-correctness in tests since certain member functions do make distinctions between being `const` or non-`const`. This can be fixed later since there are other tests where I'd like to add `const` to local variables.

ChrisThrasher (approved):

vittorioromeo:
@Bromeon: can we merge this?

Bromeon:
Absolutely 🚀

PR #1983, opened 2022-01-27
Don't add SFML:: namespace to external targets

### LINKED ISSUE #1981: [Regression] Wrong Targets for Dependencies while Static Linking
## Subject of the issue

With the #1947 SFML:: namespace change, it would appear that all the targets defined within SFML receive the `SFML::` namespace, including the dependencies (e.g. Freetype, opengl, ...), which is not an issue when building shared libraries, but fails when trying to link SFML statically with CMake.

## Your environment

* Windows
* SFML 3 - master
* VS 2022
* `BUILD_SHARED_LIBS=OFF`

## Steps to reproduce

- Build SFML statically (`BUILD_SHARED_LIBS=OFF`)
- Have your CMake project look for SFML static libs

```cmake
cmake_minimum_required(VERSION 3.15)

project(SFML-Testbed)

find_package(SFML 3 COMPONENTS Graphics REQUIRED)
add_executable(${PROJECT_NAME} main.cpp)

target_link_libraries(${PROJECT_NAME} SFML::Graphics)
```

## Actual behavior

The SFML Config scripts fail with for example:

```
CMake Error at cmake/SFML/SFMLConfigDependencies.cmake:33 (set_property):
set_property could not find TARGET Freetype.  Perhaps it has not yet been created.
```

or

```
CMake Error at cmake/SFML/SFMLConfigDependencies.cmake:53 (set_property):
set_property could not find TARGET OpenGL.  Perhaps it has not yet been created.
```

## Expected behavior

It should be fully configured

## Potential Fix 1

https://github.com/SFML/SFML/blob/7c80f302e489ec1754c4cc03cb1cd41f68785850/cmake/SFMLConfigDependencies.cmake.in#L32-L36

Add the `SFML::` prefix in front of the `${THIS_TARGET}`.

This is in my view not really desired, as non-SFML targets (e.g. freetype, opengl, etc.) shouldn't be part of the `SFML::` namespace.

## Potential Fix 2

https://github.com/SFML/SFML/blob/7c80f302e489ec1754c4cc03cb1cd41f68785850/cmake/Macros.cmake#L453-L456

Make sure that the `SFML::` prefix here, isn't applied to SFML's dependencies.

ChrisThrasher:
## Description

Closes #1981 and fixes an issue I introduced in #1947 where the `SFML::` namespace was prepended to external targets as well as our internal targets. This apparently is fine when building shared libraries but broke when @eXpl0it3r attempted to build and use static libraries. Luckily CMake lets you add namespaces in the `EXPORT_NAME` property of a target so we can simply add `SFML::` there instead of relying on CMake to add `SFML::` to all targets in the export set.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -81,7 +81,7 @@ macro(sfml_add_library module)
     set_target_properties(${target} PROPERTIES DEFINE_SYMBOL ${NAME_UPPER}_EXPORTS)
 
     # define the export name of the module
-    set_target_properties(${target} PROPERTIES EXPORT_NAME ${module})
+    set_target_properties(${target} PROPERTIES EXPORT_NAME SFML::${module})
 
     # adjust the output file prefix/suffix to match our conventions
     if(BUILD_SHARED_LIBS AND NOT THIS_STATIC)
@@ -452,7 +452,6 @@ function(sfml_export_targets)
 
     install(EXPORT SFMLConfigExport
             FILE ${targets_config_filename}
-            NAMESPACE SFML::
             DESTINATION ${config_package_location})
 
     install(FILES "${CMAKE_CURRENT_BINARY_DIR}/SFMLConfig.cmake"


eXpl0it3r:

eXpl0it3r (approved):
Tested on Windows and makes sense 🙂

eXpl0it3r:
Thanks for getting this fixed! 🎉

PR #1984, opened 2022-01-28
Fix SFML App Xcode template

GP2P:
## Description

Renamed `TemplateInfo.plist.in` to `TemplateInfo.plist`
File named incorrectly so the SFML App Xcode template can't be loaded
All other .plist files in other templates are correct and don't have the `.in`

## Tasks

* [x] Tested on macOS

## How to test this PR?

Load the old and new files in Xcode, only the new file will be recognized and opened.


ChrisThrasher:
`.in` is a CMake convention referring to files which CMake will transform into some file without that `.in` extension.  This filename is intentional. The CMake build uses that file to create TemplateInfo.plist. 

https://github.com/SFML/SFML/blob/master/CMakeLists.txt#L474-L477

Bromeon:
@ChrisThrasher is right, the `.in` convention is on purpose. Often, there needs to be some preprocessing by CMake (like variable substitution); in that case, the file couldn't be used directly.

GP2P:
Thanks to you two for clearing that for me! And sorry for the false PR. I'm a normal user and when I wanted to find the most up to date Xcode template for my school project, I came to this GitHub repo to find it. (The SFML installed via homebrew and snapshots don't contain Xcode templates)

Can you please tell me where to actually find the most up to date version of the Xcode template? I'm on an Apple Silicon machine and the 2.5.1 zip on the website isn't compatible because it was from 2018. @Bromeon

eXpl0it3r:
No worries! 🙂 

You can either build SFML yourself (I recommend the 2.6.x branch, as there's a lot of changes on the master branch).
Or you can download a snapshot build here: https://artifacts.sfml-dev.org/by-branch/2.6.x/

GP2P:
Thanks for the link, though I don't think the macOS snapshot versions contain any Xcode template. That was the reason why I tried to come here to find a xcode template that works (the old template from the official website downloads does not work for some reason, probably outdated)

I tried to build SFML myself but I got confused by the multi-component structure. Although I made each component (audio, network, window etc) build, I can't find the code that builds the xcode templates. It would be great if there's an guide for building SFML in the README

But again, I basically just need a Xcode template that works, so I can work on my starter school project on my Apple Silicon machine on Xcode

PR #1995, opened 2022-02-05
Remove VS 2017 as GitHub Actions compiler

eXpl0it3r:
## Description

As discussed in #1937, due to poor C++17 support in VS 2017, we'll remove VS 2017 from the build pipeline

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,6 @@ jobs:
       fail-fast: false
       matrix:
         platform:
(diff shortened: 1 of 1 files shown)


Bromeon (approved):
Sounds good!

I assume we will then also stop supporting VS 2017 officially for SFML 3, e.g. no longer distribute binaries.

eXpl0it3r:
Well yeah, kind of hard to create binaries if the compiler can't build them 😄

PR #1997, opened 2022-02-07
Added support for coverage reporting.

binary1248:
Figured with all the unit tests added recently it would make sense to start measuring test coverage as well.

[Codecov](https://about.codecov.io/) seemed nice and integrates well with GitHub so I gave it a try. I am by no means an expert of any kind and the current changes should serve merely as a starting point.

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -1,0 +1,31 @@
+codecov:
+  require_ci_to_pass: yes
+
+coverage:
+  precision: 2
+  round: down
+  range: "70...100"
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 100%
+        base: auto
+
+parsers:
+  gcov:
+    branch_detection:
+      conditional: yes
+      loop: yes
+      method: no
+      macro: no
+
+comment:
+  layout: "reach,diff,flags,files,footer"
+  behavior: default
+  require_changes: no
+  require_head: no
+  require_base: no
+
+github_checks:
+  annotations: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,10 +35,23 @@ jobs:
     - name: Checkout Code
       uses: actions/checkout@v2
 
-    - name: Install Linux Dependencies
+    - name: Detect Linux Clang Version
       if: runner.os == 'Linux'
-      run: sudo apt-get update && sudo apt-get install libxrandr-dev libxcursor-dev libudev-dev libopenal-dev libflac-dev libvorbis-dev libgl1-mesa-dev libegl1-mesa-dev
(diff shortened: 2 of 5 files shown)


Bromeon:
Thanks for taking the initiative! 

I have to admit, my experience with code coverage is very limited. Have you made good experiences with it, in the sense that it helped detect untested code parts, or at least gave you an overall confidence level?

That said, as long as it's opt-in in CMake, it's a nice addition. With current resources it's probably not realistic (or worthwhile) to bring coverage near 100%, but it's even satisfying to see that a PR increases coverage by a small amount 🙂

eXpl0it3r:
We use code coverage at work and it can help notice whether your tests cover all the cases of your newly written code and it can overall point to modules or sections of code that are still untested and thus have a higher risk of silently breaking during some refactoring. So if I have a bit of down time, I'd pick a section and add one or the other test.

How does coverage work exactly for SFML, given that we have a lot of platform-specific code. Does it accumulate the reports from all builds on the different platforms?

binary1248:
@Bromeon Like @eXpl0it3r said, the main benefit of measuring coverage is improving developer time utilization. Anybody can open up a PR and add a bunch more tests, but if it doesn't actually cover more code then the change doesn't really bring that much value. In its ultimate use case, coverage is combined with fuzzing and [mutation testing](https://en.wikipedia.org/wiki/Mutation_testing) to eliminate the human factor completely.

Anybody who has an idea what code coverage is will know that unless you are writing a purely algorithmic library it is practically impossible to get 100% coverage in the real world. SFML interacts a lot with the underlying operating system and other external systems. While it isn't impossible to test these things (e.g. mocking), considering the required effort, I personally don't see much value in doing so since we know that OS/hardware behaviour is not guaranteed to stay the same, thus passing tests would not prevent defects in specific user scenarios.

The current configuration reports coverage for all GitHub Actions jobs that allow coverage to be measured (tests need to be run, tooling needs to be available, building in debug must be possible). All jobs besides macOS frameworks and the mobile platforms generate coverage data. The data is uploaded to Codecov and is aggregated based on *magic* to create the coverage report that includes data from the different operating systems.

Such a report can be seen here:
https://codecov.io/gh/SFML/SFML/commit/8560c6e3fafb5b4b3141a7df3caf7dbda74dea0d

Since this will be the first commit for which data is collected, the report isn't as meaningful since there was no previous baseline available. All future commits based on this one will report more useful information (including the comment bot). Because I can't see what the delta would look like until after the initial configuration has been merged, it isn't unlikely that the settings will have to be tweaked at some point in the future. To get to that point this will have to be merged first.

eXpl0it3r (approved):

eXpl0it3r:
Thanks a lot for getting this added. Let's see how it operates when it's actually "live" 🙂

binary1248:
Let the race to [70%](https://github.com/SFML/SFML/blob/66339bc09c9470b5a809743459ce5b942288765b/.github/.codecov.yml#L7) begin. 😁

PR #1999, opened 2022-02-10
Turn more compile-time constants into 'constexpr' variables

vittorioromeo:
Turns many `static const` variables and static data members whose values are known at compile-time `static constexpr` variables instead.


src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -77,14 +77,14 @@ namespace
 
         sf::String                            wmAbsPosGood[] = { "Enlightenment", "FVWM", "i3" };
 
-        static const unsigned long            eventMask = FocusChangeMask      | ButtonPressMask     |
+        static constexpr unsigned long        eventMask = FocusChangeMask      | ButtonPressMask     |
                                                         ButtonReleaseMask    | ButtonMotionMask    |
                                                         PointerMotionMask    | KeyPressMask        |
                                                         KeyReleaseMask       | StructureNotifyMask |
                                                         EnterWindowMask      | LeaveWindowMask     |
                                                         VisibilityChangeMask | PropertyChangeMask;
```
ChrisThrasher, line 85: Should the following lines be indented as well?

include/SFML/System/Angle.inl
```diff
@@ -257,3 +257,13 @@ constexpr Angle operator "" _rad(unsigned long long angle)
...
 } // namespace Literals
+
+
+////////////////////////////////////////////////////////////
+// Static member data
+////////////////////////////////////////////////////////////
+
+// Note: the 'inline' keyword here is technically not required, but VS2019 fails
+// to compile with a bogus "multiple definition" error if not explicitly used.
+
```
Bromeon, line 268: Should we remove this empty line, so it's clear the comments refer to the following definition?

src/SFML/Window/Unix/WindowImplX11.cpp
Bromeon: That would probably be undone by https://github.com/SFML/SFML/pull/1861, so not sure if it's worth to invest any effort in horizontal formatting.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -655,23 +655,23 @@ m_lastInputTime  (0)
...
+            static constexpr unsigned long MWM_DECOR_MENU        = 1 << 4;
+            static constexpr unsigned long MWM_DECOR_MINIMIZE    = 1 << 5;
+            static constexpr unsigned long MWM_DECOR_MAXIMIZE    = 1 << 6;
+
+            //static constexpr unsigned long MWM_FUNC_ALL          = 1 << 0;
+            static constexpr unsigned long MWM_FUNC_RESIZE       = 1 << 1;
+            static constexpr unsigned long MWM_FUNC_MOVE         = 1 << 2;
+            static constexpr unsigned long MWM_FUNC_MINIMIZE     = 1 << 3;
+            static constexpr unsigned long MWM_FUNC_MAXIMIZE     = 1 << 4;
+            static constexpr unsigned long MWM_FUNC_CLOSE        = 1 << 5;
```
Bromeon, line 674: I guess we can do this here, because the Unix impl won't be compiled by VS 2019?

Still a bit strange to use different "styles" across the code base, this may not be obvious to someone who wasn't involved in the discussions we have now 🧐

src/SFML/Window/Window.cpp
```diff
@@ -164,7 +164,7 @@ void Window::close()
 ////////////////////////////////////////////////////////////
 const ContextSettings& Window::getSettings() const
 {
-    static const ContextSettings empty(0, 0, 0);
+    static constexpr ContextSettings empty(0, 0, 0);
```
Bromeon, line 167: Can we not remove the `static` entirely?
Also below.

src/SFML/Window/Unix/WindowImplX11.cpp
vittorioromeo: Nah, the problem with having to add `inline` was specific to out-of-class `constexpr` definitions of non-`constexpr` static data members, these are in function-scope, so it's all good.

include/SFML/System/Angle.inl
vittorioromeo: Sure.

src/SFML/Window/Window.cpp
Bromeon: I've read that SO answer multiple times and still struggle to understand it. It explains things very incoherently, in my opinion. In particular, I fail to see how this is a best practice.

> First, note that static and constexpr are completely independent of each other. static defines the object's lifetime during execution; constexpr specifies that the object should be available during compilation. Compilation and execution are disjoint and discontiguous, both in time and space. So once the program is compiled, constexpr is no longer relevant.

I don't see how two keywords about object lifetime and storage can be fully independent.
* An object, which is fully substituted at compile time, is simply a series of values in the code segment of the executable. Any occurrences to it in the code can be hardcoded, no memory load instructions are needed at runtime. (This assuming that the compiler indeed substitutes the `constexpr`).
* A function-local `static` object is initialized on first invocation and destroyed at the end of the application. It typically has an address.

Now the compiler has a lot of optimization potential and would likely treat the variable like `constexpr` in our case, since there is no constructor/destructor with side-effects. Which again means that writing `constexpr` directly would convey the intent more clearly.

In general, I think the "store my small object as `static` so I don't have to re-initialize it" is a hard anti-pattern, which likely pessimizes performance (and even created extra issues with thread safety in older C++ standards). If we want a compile-time constant, let's use `constexpr`. If we _need_ something to be initialized only once and/or live until the end of times, then `static` is the keyword.

I guess this counts as "feel strongly"? 😀 Most important to me is that we convey the intent of the code. Things like `static constexpr` are definitely unusual sights, and I'd guess most people don't know what it does, let alone why we chose it over `constexpr` or `static`. Simpler is better.

vittorioromeo: > An object, which is fully substituted at compile time

That's not a guarantee for `constexpr` variables. You can ODR-use them, which means they need to have a memory address (and storage). The simplest example is that you are able to take the address of a `constexpr` variable.

> Any occurrences to it in the code can be hardcoded, no memory load instructions are needed at runtime

This is not always true due to what I mentioned above.

> Now the compiler has a lot of optimization potential and would likely treat the variable like constexpr in our case

Ah, I see. You're saying that those points apply if the `constexpr` variable is actually substituted. I agree with that.

> Which again means that writing constexpr directly would convey the intent more clearly.

I am not sure it would. In function scope, I see both `static constexpr` and `constexpr` as "compile-time constant". I can agree that the latter is simpler, but it would still make sense to use `static constexpr` in situations where the object is ODR-used and it can be large.

> In general, I think the "store my small object as static so I don't have to re-initialize it" is a hard anti-pattern [...]

I agree for small objects.

> I guess this counts as "feel strongly"?

I don't feel strongly about either option, so you win this one :)

vittorioromeo: Oh, by the way, this is the only place where the `static` is necessary, because we are returning a `const&`!!!

vittorioromeo: @Bromeon: I've removed `static` for small function-scope variables. I've kept it for larger ones (e.g. arrays). And I've kept it here because it is necessary to avoid UB.

Bromeon: Your points are true, which is why I tried to carefully limit my statements to the applicable bounds 😉 ("if `constexpr` is indeed substituted"; "for small objects"). But thanks for shedding another light on this, I also had to think a bit about the exact semantics.

> Oh, by the way, this is the only place where the `static` is necessary, because we are returning a `const&`!!!

True, but that's also bad design -- a good return type would probably be `std::optional<ContextSettings>` to signal when there are no settings. But even using `ContextSettings` would be better; I can't imagine a situation where we cannot afford this copy and have to resort to `static` objects only to satisfy the reference...

Anyway, that's a bit out-of-scope of this PR 🙂

vittorioromeo: @Bromeon: don't forget to approve if you are happy with the current changeset :)

Bromeon (approved):

PR #2000, opened 2022-02-10
Turn 'sf::Transform' into a 'constexpr' class

vittorioromeo:

----

## Description

This PR turns `sf::Transform` into a `constexpr`-friendly class.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.


include/SFML/Graphics/Transform.hpp
```diff
@@ -31,16 +31,17 @@
 #include <SFML/Graphics/Export.hpp>
 #include <SFML/Graphics/Rect.hpp>
 #include <SFML/System/Vector2.hpp>
-#include <SFML/System/Angle.hpp>
 
 
 namespace sf
 {
+class Angle;
+
```
ChrisThrasher, line 39: Is this trying to minimize compile time? Is that compile time improvement measurable or meaningful?

ChrisThrasher (commented):
Can we add a `static_assert` to the unit tests that ensure this can be used at compile time?

eXpl0it3r:

Bromeon:
Before we turn SFML into a header-only library, do we know that we'll never want to go back? 🙂 

Removing `constexpr` will be a breaking change. We give up the freedom to change implementation (let's say by using a vector library); at least not without sacrificing a lot of compile time. Maybe that's an OK trade-off, but we should keep it in mind.

Also, while a single `constexpr` class likely makes no meaningful difference in compile time, do we know that this is still the case when a significant part of SFML has its implementation in the header? I assume this is negligible compared to monomorphization (template instantiation) of standard headers?

Bromeon (approved):
I haven't checked every single line moved into the header, so I assume semantics are unchanged 😀

If we have the buy-in from everyone regarding above-mentioned caveats, this is good to go!

PR #2001, opened 2022-02-10
Fix license detection and add helpful links to external libraries and their licenses

binary1248:
GitHub uses [licensee](https://github.com/licensee/licensee) to detect a repository's license for display and indexing purposes. In its current state, the license text in [license.md](https://github.com/SFML/SFML/blob/e458f4651e3cbfd168b9be0cfa328a401dccf895/license.md) isn't recognized as Zlib because there are too many discrepancies for the matcher to output a high enough confidence value. As such SFML is currently not indexed with a license. This has an impact on the discoverability/findability of the library.

If a developer is bound by legal reasons to use libraries that are clearly under certain licenses they would save themselves the time and filter during their initial search.

Compare [without license filter](https://github.com/search?q=multimedia&type=Repositories) and [with license filter](https://github.com/search?q=multimedia+license%3Azlib&type=Repositories).

This change makes licensee output a 100% match to Zlib and adds helpful links to the repositories and license texts of all external libraries.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -38,3 +38,19 @@ There are several places to learn SFML:
 ## Contribute
 
 SFML is an open-source project, and it needs your help to go on growing and improving. If you want to get involved and suggest some additional features, file a bug report or submit a patch, please have a look at the [contribution guidelines](https://www.sfml-dev.org/contribute.php).
+
+## License
+
+The SFML libraries and source code are distributed under the [zlib/libpng license](https://opensource.org/licenses/Zlib). See [license.md](license.md). External libraries used by SFML are distributed under their own licenses.
+
+In short, SFML is free for any use (commercial or personal, proprietary or open-source). You can use SFML in your project without any restriction. You can even omit to mention that you use SFML -- although it would be appreciated.
+
+## External libraries used by SFML
+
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -38,3 +38,19 @@ There are several places to learn SFML:
...
+## License
+
+The SFML libraries and source code are distributed under the [zlib/libpng license](https://opensource.org/licenses/Zlib). See [license.md](license.md). External libraries used by SFML are distributed under their own licenses.
+
+In short, SFML is free for any use (commercial or personal, proprietary or open-source). You can use SFML in your project without any restriction. You can even omit to mention that you use SFML -- although it would be appreciated.
+
+## External libraries used by SFML
+
+  * [_OpenAL-Soft_](https://github.com/kcat/openal-soft) is under the [LGPL license](https://github.com/kcat/openal-soft/blob/master/COPYING)
+  * [_stb\_image_ and _stb_image_write_](https://github.com/nothings/stb) are [public domain](https://github.com/nothings/stb/blob/master/LICENSE)
```
eXpl0it3r, line 51: Should we mention the dual license here MIT / unlicense?

eXpl0it3r (approved):

readme.md
Bromeon: We can, but I think it's mostly relevant which of the license we use when we have it as a SFML dependency. Either should be fine, public domain is the least restrictive.

PR #2004, opened 2022-02-10
Make `sf::Clock` pausable

MetGang:

## Description

Ability to pause `sf::Clock` was discussed and proposed on SFML forum many times - yet it was not implemented (and there is no plans to do so) in SFML 2.x. With this PR this goal is achieved (and some more) with SFML 3 release in mind.
This is very light PR that may cause more changes to come in the future.

## Tasks

* [ ] Tested on Linux
* [X] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Internal repository tests have been updated to cover changes from this PR.
*Passed on my machine.*

### DIFF
--- a/include/SFML/System/Clock.hpp
+++ b/include/SFML/System/Clock.hpp
@@ -104,21 +104,61 @@ public:
     Time getElapsedTime() const;
 
     ////////////////////////////////////////////////////////////
+    /// \brief Check whether the clock is running
+    ///
+    /// \return True if the clock is running, false otherwise
+    ///
+    ////////////////////////////////////////////////////////////
+    bool isRunning() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Start the clock
+    ///
+    /// \see stop
+    ///
+    ////////////////////////////////////////////////////////////
+    void start();
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Stop the clock
+    ///
+    /// \see start
+    ///
+    ////////////////////////////////////////////////////////////
+    void stop();
+
+    ////////////////////////////////////////////////////////////
     /// \brief Restart the clock
     ///
-    /// This function puts the time counter back to zero.
-    /// It also returns the time elapsed since the clock was started.
+    /// This function puts the time counter back to zero, returns
+    /// the elapsed time, and leaves the clock in a running state.
     ///
     /// \return Time elapsed
     ///
+    /// \see reset
+    ///
     ////////////////////////////////////////////////////////////
     Time restart();
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Reset the clock
+    ///
+    /// This function puts the time counter back to zero, returns
+    /// the elapsed time, and leaves the clock in a paused state.
+    ///
+    /// \return Time elapsed
+    ///
+    /// \see restart
+    ///
+    ////////////////////////////////////////////////////////////
+    Time reset();
 
 private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    priv::ClockImpl::time_point m_startTime{priv::ClockImpl::now()}; //!< Time of last reset
+    priv::ClockImpl::time_point m_refPoint{priv::ClockImpl::now()}; //!< Time of last reset
+    priv::ClockImpl::time_point m_stopPoint;                        //!< Time of last stop
 };
 
 } // namespace sf
@@ -130,7 +170,7 @@ private:
 ///
 /// sf::Clock is a lightweight class for measuring time.
 ///
-/// Its provides the most precise time that the underlying
+/// It provides the most precise time that the underlying
 /// OS can achieve (generally microseconds or nanoseconds).
 /// It also ensures monotonicity, which means that the returned
 /// time can never go backward, even if the system time is
@@ -143,6 +183,8 @@ private:
 /// Time time1 = clock.getElapsedTime();
 /// ...
 /// Time time2 = clock.restart();
+/// ...
+/// Time time3 = clock.reset();
 /// \endcode
 ///
 /// The sf::Time value returned by the clock can then be
--- a/src/SFML/System/Clock.cpp
+++ b/src/SFML/System/Clock.cpp
@@ -34,17 +34,54 @@ namespace sf
 ////////////////////////////////////////////////////////////
 Time Clock::getElapsedTime() const
 {
-    return std::chrono::duration_cast<std::chrono::microseconds>(priv::ClockImpl::now() - m_startTime);
+    if (isRunning())
+        return std::chrono::duration_cast<std::chrono::microseconds>(priv::ClockImpl::now() - m_refPoint);
+    return std::chrono::duration_cast<std::chrono::microseconds>(m_stopPoint - m_refPoint);
+}
+
+
+////////////////////////////////////////////////////////////
+bool Clock::isRunning() const
+{
+    return m_stopPoint == priv::ClockImpl::time_point();
+}
+
+
+////////////////////////////////////////////////////////////
+void Clock::start()
+{
+    if (!isRunning())
+    {
+        m_refPoint += priv::ClockImpl::now() - m_stopPoint;
+        m_stopPoint = {};
+    }
+}
+
+
+////////////////////////////////////////////////////////////
+void Clock::stop()
+{
+    if (isRunning())
+        m_stopPoint = priv::ClockImpl::now();
 }
 
 
 ////////////////////////////////////////////////////////////
 Time Clock::restart()
 {
-    const auto now     = priv::ClockImpl::now();
-    const auto elapsed = std::chrono::duration_cast<std::chrono::microseconds>(now - m_startTime);
-    m_startTime        = now;
+    const Time elapsed = getElapsedTime();
+    m_refPoint         = priv::ClockImpl::now();
+    m_stopPoint        = {};
+    return elapsed;
+}
 
+
+////////////////////////////////////////////////////////////
+Time Clock::reset()
+{
+    const Time elapsed = getElapsedTime();
+    m_refPoint         = priv::ClockImpl::now();
+    m_stopPoint        = m_refPoint;
     return elapsed;
 }
 
--- a/test/System/Clock.test.cpp
+++ b/test/System/Clock.test.cpp
@@ -9,6 +9,8 @@
 
 TEST_CASE("[System] sf::Clock")
 {
+    using namespace std::chrono_literals;
+
     SECTION("Type traits")
     {
         STATIC_CHECK(std::is_copy_constructible_v<sf::Clock>);
@@ -17,21 +19,50 @@ TEST_CASE("[System] sf::Clock")
         STATIC_CHECK(std::is_nothrow_move_assignable_v<sf::Clock>);
     }
 
+    SECTION("Construction")
+    {
+        const sf::Clock clock;
+        CHECK(clock.isRunning());
+        CHECK(clock.getElapsedTime() >= sf::microseconds(0));
+    }
+
     SECTION("getElapsedTime()")
     {
         const sf::Clock clock;
         CHECK(clock.getElapsedTime() >= sf::microseconds(0));
         const auto elapsed = clock.getElapsedTime();
-        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+        std::this_thread::sleep_for(1ms);
         CHECK(clock.getElapsedTime() > elapsed);
+    }
+
+    SECTION("start/stop")
+    {
+        sf::Clock clock;
+        clock.stop();
+        CHECK(!clock.isRunning());
+        const auto elapsed = clock.getElapsedTime();
+        std::this_thread::sleep_for(1ms);
+        CHECK(elapsed == clock.getElapsedTime());
+
+        clock.start();
+        CHECK(clock.isRunning());
+        CHECK(clock.getElapsedTime() >= elapsed);
     }
 
     SECTION("restart()")
     {
         sf::Clock clock;
         CHECK(clock.restart() >= sf::microseconds(0));
-        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+        CHECK(clock.isRunning());
+        std::this_thread::sleep_for(1ms);
         const auto elapsed = clock.restart();
         CHECK(clock.restart() < elapsed);
+    }
+
+    SECTION("reset()")
+    {
+        sf::Clock clock;
+        CHECK(clock.reset() >= sf::microseconds(0));
+        CHECK(!clock.isRunning());
     }
 }


include/SFML/System/Clock.hpp
```diff
@@ -107,7 +108,47 @@ class SFML_SYSTEM_API Clock
...
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Check whether certain time has passed
+    ///
+    /// This function returns boolean value based on whether certain time has passed.
+    ///
+    /// \return True whether certain time has passed, false otherwise
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] bool hasTimePassed(const Time& time) const noexcept;
```
ChrisThrasher, line 121: Is this equivalent to `clock.getElapsedTime() > time`?

include/SFML/System/Time.hpp
```diff
@@ -86,9 +86,9 @@ class Time
 
 private:
 
-    friend constexpr Time seconds(float);
-    friend constexpr Time milliseconds(Int32);
-    friend constexpr Time microseconds(Int64);
+    friend constexpr Time seconds(float) noexcept;
+    friend constexpr Time milliseconds(Int32) noexcept;
+    friend constexpr Time microseconds(Int64) noexcept;
```
ChrisThrasher, line 91: I believe you're the first person to start (rightfully) decorating functions as `noexcept` where appropriate. This PR would be the first to introduce such changes. @eXpl0it3r @Bromeon @vittorioromeo is this a pattern we should start applying to everything which cannot throw?

test/System/Clock.cpp
```diff
@@ -16,12 +16,53 @@ TEST_CASE("sf::Clock class - [system]")
...
+        CHECK(clock.hasTimePassed(sf::microseconds(0)));
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+        CHECK(clock.hasTimePassed(sf::milliseconds(1)));
+    }
+
+    SUBCASE("start()")
+    {
+        sf::Clock clock;
+        clock.start();
+        CHECK(clock.isRunning());
```
ChrisThrasher, line 31: Should we check that `isRunning` is false after construction?

test/System/Clock.cpp
```diff
@@ -16,12 +16,53 @@ TEST_CASE("sf::Clock class - [system]")
...
+    }
+
+    SUBCASE("pause()")
+    {
+        sf::Clock clock;
+        clock.pause();
+        CHECK(clock.isPaused());
```
ChrisThrasher, line 38: Should we check that `isPaused` is false after construction?

Bromeon:
Regarding `[[nodiscard]]` and `noexcept`:
[Forum thread](https://en.sfml-dev.org/forums/index.php?topic=28340.0)
[GitHub comment about `[[nodiscard]]`](https://github.com/SFML/SFML/pull/1883#issuecomment-989583140)

Would be nice if we didn't have to reiterate the discussion for every pull request 🙂

include/SFML/System/Time.hpp
Bromeon: see main thread

Bromeon (commented):
Thanks for the contribution!

This changes a bit many things at once, e.g. the `[[nodiscard]]`/`noexcept` is highly controversial and shouldn't be mixed in a functionality-PR, if it changes a lot of code beyond the additions. See my message above for rationale.

Also, it's not really clear whether stop-watches are something we need/want in SFML, and what the design would be. For example, the current approach makes `sf::Clock` bigger, even if the pause functionality is not used.

test/System/Clock.cpp
MetGang: That's a good point since clock by default is running.

test/System/Clock.cpp
MetGang: According to my previous response `!isPaused` is more appropriate.

include/SFML/System/Clock.hpp
MetGang: It is equivalent. Motivation was just to have dedicated method to perform these checks. I understand some may find this addition redundant. Also I can see your point with this method being less versetile than comparision operators which sf::Time has plenty.

include/SFML/System/Clock.hpp
```diff
@@ -107,7 +108,47 @@ class SFML_SYSTEM_API Clock
...
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Toggle clock state
+    ///
+    /// This function toggles clock state between running and paused.
+    ///
+    /// \see start, pause, isRunning, isPaused
+    ///
+    ////////////////////////////////////////////////////////////
+    void toggle() noexcept;
```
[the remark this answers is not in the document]
MetGang: I read posts about `[[nodiscard]]`/`noexcept` and I understand that this PR may push this concept too much when its main goal is ability to pause a clock. Nonetheless, sooner or later these changes **may** be necessary but now for I'll reiterate changes for more consistency.

As for the point with size increase there was idea to seperate this feature and make `sf::PausableClock` along with unchanged `sf::Clock`. Then, user wouldn't have to pay for pause feature.

`toggle()` was motivated with simplifying code like below but since `setX(bool)` is more used in library and is more versatile I understand your point. It can substituted with `clock.setRunning(!clock.isRunning())` though.
```cpp
if (clock.isRunning())
{
    clock.pause();
}
else
{
    clock.start();
}
```

With addition of `setRunning(bool)` I'd drop `isPaused()` and keep `isRunning()` for consistency but keep `start()` and `pause()`.

I'll wait for responses and more feedback then commit accordingly ;)

vittorioromeo:
> For example, the current approach makes `sf::Clock` bigger, even if the pause functionality is not used.

@Bromeon: This is technically true, but it really is insignificant. :)

---

@MetGang: I would simplify the entire PR by a lot. Drop all the unnecessary changes (e.g. `noexcept` and `[[nodiscard]]`), replace the multiple functions with `setRunning`, remove `hasTimePassed`. It will be easier for us to review it and approve it. :)

src/SFML/System/Clock.cpp
```diff
@@ -31,37 +31,102 @@
...
 
 
 ////////////////////////////////////////////////////////////
-Time Clock::getElapsedTime() const
+Time Clock::getElapsedTime() const noexcept
 {
-    return durationToTime(ClockImpl::now() - m_startTime);
+    if (isRunning())
+    {
+        return sf::microseconds((ClockImpl::now() - m_refPoint).count());
```
kimci86, line 56: This implicitly requires `ClockImpl` to give time in microseconds, which might not be the case.

src/SFML/System/Clock.cpp
```diff
@@ -31,37 +31,102 @@
...
+    m_pausePoint = m_refPoint = ClockImpl::now();
+
+    return current;
+}
+
+
+////////////////////////////////////////////////////////////
+bool Clock::isRunning() const noexcept
+{
+    return m_pausePoint.time_since_epoch() == ClockImpl::duration::zero();
```
kimci86, line 122: Why not `m_pausePoint == ClockImpl::time_point{}`?

src/SFML/System/Clock.cpp
```diff
@@ -31,37 +31,102 @@
...
+
+}
+
+
 ////////////////////////////////////////////////////////////
-Clock::Clock() :
-m_startTime(ClockImpl::now())
+bool Clock::hasTimePassed(const Time& time) const noexcept
 {
+    return time.asMicroseconds() < getElapsedTime().asMicroseconds();
```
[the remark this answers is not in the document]
Bromeon: As mentioned by ChrisThrasher and vittorio already, this method should go entirely.

src/SFML/System/Clock.cpp
```diff
@@ -31,37 +31,102 @@
...
+}
 
-    return elapsed;
+
+////////////////////////////////////////////////////////////
+Time Clock::restart() noexcept
+{
+    Time current = getElapsedTime();
+
+    m_pausePoint = ClockImpl::time_point{};
```
Bromeon, line 100: Please use `type()` constructor syntax for consistency

src/SFML/System/Clock.cpp
MetGang: Addressed in new commits.

src/SFML/System/Clock.cpp
```diff
@@ -31,37 +31,102 @@
...
 
 
 ////////////////////////////////////////////////////////////
-Time Clock::restart()
+void Clock::start() noexcept
 {
-    const ClockImpl::time_point now = ClockImpl::now();
-    Time elapsed = durationToTime(now - m_startTime);
-    m_startTime = now;
+    m_refPoint += ClockImpl::now() - m_pausePoint;
```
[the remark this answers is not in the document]
MetGang: Addressed in new commits.

src/SFML/System/Clock.cpp
```diff
@@ -31,37 +31,102 @@
...
+
+    m_pausePoint = ClockImpl::time_point{};
+}
+
+
+////////////////////////////////////////////////////////////
+void Clock::pause() noexcept
+{
+    m_pausePoint = ClockImpl::now();
```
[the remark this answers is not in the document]
MetGang: Addressed in new commits.

src/SFML/System/Clock.cpp
MetGang: `m_pausePoint == ClockImpl::time_point()` is also a valid (and maybe simpler) way but I'd personally stick to existing one since it reflects intentions (`m_pausePoint` must be zero to identify that clock is running) better. I'm open to hear others' opinions though.

src/SFML/System/Clock.cpp
MetGang: Addressed in new commits.

include/SFML/System/Clock.hpp
```diff
@@ -134,15 +192,18 @@ class SFML_SYSTEM_API Clock
     /// This function acts as a utility for converting clock
     /// duration type instance into sf::Time
     ///
+    /// \param duration std::chorno's duration to be converted
```
ananace, line 195: Typo, should be `std::chrono`

MetGang: I misstyped `std::chrono` as `std::chorno`

DelinWorks: What they look soo the same 😂

include/SFML/System/Time.hpp
```diff
@@ -1,7 +1,7 @@
 ////////////////////////////////////////////////////////////
 //
 // SFML - Simple and Fast Multimedia Library
-// Copyright (C) 2007-2021 Laurent Gomila (laurent@sfml-dev.org)
+// Copyright (C) 2007-2022 Laurent Gomila (laurent@sfml-dev.org)
```
ChrisThrasher, line 4: Did you mean to change the copyright year on a file you didn't modify?

include/SFML/System/Time.inl
```diff
@@ -1,7 +1,7 @@
 ////////////////////////////////////////////////////////////
 //
 // SFML - Simple and Fast Multimedia Library
-// Copyright (C) 2007-2021 Laurent Gomila (laurent@sfml-dev.org)
+// Copyright (C) 2007-2022 Laurent Gomila (laurent@sfml-dev.org)
```
ChrisThrasher, line 4: Did you mean to change this copyright year?

src/SFML/System/Time.cpp
```diff
@@ -30,6 +30,7 @@
 
 namespace sf
 {
+
```
ChrisThrasher, line 33: Did you mean to modify whitespace in a file you otherwise didn't change?

src/SFML/System/Clock.cpp
```diff
@@ -31,36 +31,103 @@
...
 Time Clock::restart()
 {
-    const ClockImpl::time_point now = ClockImpl::now();
-    Time elapsed = durationToTime(now - m_startTime);
-    m_startTime = now;
+    const Time current = getElapsedTime();
+
+    m_pausePoint = ClockImpl::time_point();
+
+    m_refPoint = ClockImpl::now();
```
ChrisThrasher, line 101: Could you remove a few of these extra empty newlines to better match the style of existing code? Most SFML code puts expressions on consecutive lines instead of double-spacing everything.

src/SFML/System/Clock.cpp
```diff
@@ -31,36 +31,103 @@
...
+
+
+////////////////////////////////////////////////////////////
+Time Clock::reset()
+{
+    const Time current = getElapsedTime();
+
+    m_pausePoint = m_refPoint = ClockImpl::now();
+
+    return current;
```
ChrisThrasher, line 114: See previous commenting about double-spacing all code.

Bromeon (commented):
Having 6 functions with overlapping semantics, yet completely different terminology is confusing:
* (constructor)
* setRunning
* start
* pause
* reset
* restart

I also believe we don't have to provide every possible combination of operations "for convenience", and we might want to split functionalty. Maybe `start`/`stop` are still justified, as most of the cases, the logic around will know whether the stopwatch should be running or not, and thus `setRunning(bool)` might not be necessary.

What if we keep the API simple:
```cpp
class Clock {
public:
    // Running state
    void start();
    void stop(); // formerly pause()
    bool isRunning() const;

    // Elapsed time state
    void reset();
    Time getElapsedTime() const;
};
```

One option would be to just keep the two states completely orthogonal, so `reset()` would not affect running/stopped state. But in practice, one is likely to either stop the clock or restart it, so maybe `reset()` can have a fixed meaning.

I'm not sure if we need the methods who act as both setters and getters simultaneously (current `restart()`) -- while convenient, it introduces some redundancy and makes the API surface broader. So far there was only one state (elapsed time), so it was manageable, but now we additionally introduce running/stopped, so we might focus on making sure the user always knows which state he is manipulating.

include/SFML/System/Clock.hpp
```diff
@@ -101,14 +100,47 @@ class SFML_SYSTEM_API Clock
...
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Set the clock state
+    ///
+    /// This function starts the clock if running is true
+    /// and pauses it if running is false.
+    ///
+    /// \param running True to start, false to pause
+    ///
+    /// \return True if operation was successful, false otherwise
```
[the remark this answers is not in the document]
MetGang: Thanks for pointing out docs error.

src/SFML/System/Clock.cpp
MetGang: Seems reasonable, gonna address that.

include/SFML/System/Time.hpp
MetGang: Actually Time files were changed too, then changed again and there is some leftover after those changes. I think I should revert all changes then.

include/SFML/System/Clock.hpp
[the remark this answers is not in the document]
MetGang: Regarding API changes. Changing `pause()` into `stop()` is okay. Taking return values from `restart/reset()` as form of simplifying API since we manage two states now - okay. Though I would not consider deleting `restart()` method.

ChrisThrasher:
This PR has become significantly diverged from the `master` branch. Can you rebase and squash on `master`?

ChrisThrasher:
Can you squash this into some smaller number of commits and remove the merge commits you added? That'll have to happen before it can get merged.

eXpl0it3r:

ChrisThrasher:
@MetGang Are you still interested in maintaining this PR?

MetGang:
@ChrisThrasher Sure, it seems that clang-format changes from master result in conflicts with my changes. Should be easy to fix when I find some time.

MetGang:
Yep... I metged a wrong thing here again ^^"

ChrisThrasher:
I updated this to fix all the merge conflicts. This idea is growing on me.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher (approved):

ChrisThrasher:
I renamed `pause()` to `stop()` and removed `setRunning()` since I'd rather users call `start()` and `stop()` explicitly. I still do like `reset()` and `restart()` returning a `Time` so I hope that can stay. I find that really convenient. I hope we can make a decision on this since soon.

ChrisThrasher:

estebanc08:
I am currently using a clock object to synchronize the notes being played on piano to a midi file that reads in music as timed ticks. Im wanting to add a function to pause the notes and the music and therefore pause the clock so the notes playing can stay synchronized with the clock and a pause function is really the only way I see the issue being resolved.

ChrisThrasher:

ChrisThrasher:

Bambo-Borris:
On a project of my own I've encountered a need for this, and I've also helped users in the SFML Discord recently who could have benefitted from this as a feature. I think it's a good win in terms of improving a very commonly used utility class from SFML.

Bambo-Borris (approved):

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

binary1248 (approved):

kimci86:
One small inconsistency is that a default-constructed clock is running but a clock after `reset` is not running.
One could expect `reset` to put the clock in the state of a default-constructed clock.

Looking online for other similar interfaces, I see a clock (or rather stopwatch) is not running after construction there:
- .NET: https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch
- Apache Commons: https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/time/StopWatch.html
- Dart: https://api.dart.dev/stable/3.0.2/dart-core/Stopwatch-class.html

Maybe a default-constructed clock should be stopped?
Because creating a running clock can be convenient, we could add a parameter to the clock constructor or maybe a factory function? This is what Apache commons and Guava do.
- Guava: https://guava.dev/releases/18.0/api/docs/com/google/common/base/Stopwatch.html

ChrisThrasher:
Good point on a clock maybe being stopped by default. Ideally we'd rename it to sf::Stopwatch and thus better justify some radical changes. However if we keep the same name and keep the default constructor, I'm aprehensive about a change that will break so many existing programs at runtime.

Bromeon:
We could also consider named constructor idiom:
```cpp
auto s = sf::Clock::stopped();
auto r = sf::Clock::running();
```

Do we need a default constructor? It may be convenient, but it's not obvious what it does, as the discussion here shows.

A problem with the above proposal is that static methods can also be called on an instance (another huge C++ mistake), i.e.
```cpp
sf::Clock c = ...;

sf::Clock d = c.stopped(); // does not touch c
```

ChrisThrasher:

test/System/Clock.test.cpp
```diff
@@ -17,21 +19,53 @@ TEST_CASE("[System] sf::Clock")
...
+        std::this_thread::sleep_for(1ms);
         CHECK(clock.getElapsedTime() > elapsed);
     }
 
+    SECTION("start()")
+    {
+        sf::Clock clock;
+        clock.start();
+        CHECK(clock.isRunning());
+    }
```
eXpl0it3r, line 43: There's no test case for the behavior of:

- Stop a running clock
- Take measure of the elapsed time
- Start a stopped clock
- Elapsed time isn't reset

eXpl0it3r (approved):
I'm not entirely convinced of the API, as I feel `start()`, `restart()` and `reset()` aren't exactly obvious in their operations. Why does start not reset the elapsed time, yet restart does? Why does reset leave the clock in a stopped state, when the default construction is a started state?

And I think the whole change really makes a strong argument for a different name, for which we have [a discussion thread on the forum](https://en.sfml-dev.org/forums/index.php?topic=28969.0) already.

Why an approval then? I'd like to see progress made and maybe it's not the final state of the whole feature, but we're also done with SFML 3, so lets get it in and maybe think about the fine tuning of the API along the way.  
It's a bit "radical" to how we've done stuff otherwise in SFML, but happy to try something a bit different.

ChrisThrasher:

ChrisThrasher:
Thanks everyone for their feedback on this new feature! Let's continue the API discussion in #https://github.com/SFML/SFML/issues/2570.

PR #1954, opened 2022-01-01
Update year to 2022 (2.6.x branch)

Bromeon:
## Description

Updates the copyright year to 2022.

I noticed we have some old years with a single author, e.g. in BSD/Android (`Window/OpenBSD/JoystickImpl.hpp`):
```cpp
////////////////////////////////////////////////////////////
//
// SFML - Simple and Fast Multimedia Library
// Copyright (C) 2013 Jonathan De Wachter (dewachter.jonathan@gmail.com)
//
// This software is provided 'as-is', without any express or implied warranty.
```
In others, we have this but still with Laurent and an up-to-date date (`Window/FreeBSD/JoystickImpl.cpp`):
```cpp
////////////////////////////////////////////////////////////
//
// SFML - Simple and Fast Multimedia Library
// Copyright (C) 2007-2022 Laurent Gomila (laurent@sfml-dev.org)
//               2013-2013 David Demelier (demelier.david@gmail.com)
//
// This software is provided 'as-is', without any express or implied warranty.
```

Should we maybe adopt the 2nd form for those too?

## How to test this PR?

CI + manual review.

I can repeat the same process for `master` once we decided to do it for `2.6.x`. It's definitely easier than fixing merge conflicts in the forward-merge (but maybe they will then occur anyway).

PR #2005, opened 2022-02-10
Update year to 2022 (master branch)

Bromeon:
Follow-up to https://github.com/SFML/SFML/pull/1954.

Excludes license file, which is handled by https://github.com/SFML/SFML/pull/2001.

### DIFF
--- a/examples/cocoa/CocoaAppDelegate.h
+++ b/examples/cocoa/CocoaAppDelegate.h
@@ -1,7 +1,7 @@
 ////////////////////////////////////////////////////////////
 //
 // SFML - Simple and Fast Multimedia Library
-// Copyright (C) 2007-2021 Marco Antognini (antognini.marco@gmail.com),
+// Copyright (C) 2007-2022 Marco Antognini (antognini.marco@gmail.com),
 //                         Laurent Gomila (laurent@sfml-dev.org)
 //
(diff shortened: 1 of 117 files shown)


eXpl0it3r (approved):
Hmm odd, your [other commit](https://github.com/SFML/SFML/commit/3b069726022929ec00cc91417db2c5114ff9c86e) is on master already, guess we just overlooked some files 😄

Bromeon:
Oh, good catch! Would it make more sense if I fix that (and we then back-merge `2.6.x` to `master`)?
Or do you want me to create a separate PR for `2.6.x`?
Maybe in terms of potential future merge conflicts, one "source of truth" (2.6.x) is better. There are a few new files though...

Let me know what you think.

eXpl0it3r:
I guess it would make sense to update it on the `2.6.x` branch as well and then one for the new files on master

Bromeon:
Created a separate PR for `2.6.x`.

I wondered why I missed the initial changes. I use a `grep` version on Git-for-Windows; maybe I used it wrong? 🤔

vittorioromeo:
LGTM, merging it in

eXpl0it3r:
Will be interesting to see how git handles the backmerge from `2.6.x` 😅

PR #2008, opened 2022-02-15
Fixed warnings reported by LGTM and Coverity Scan.

binary1248:
Reports:
* https://lgtm.com/projects/g/SFML/SFML/
* https://scan.coverity.com/projects/sfml-sfml (to view the issues, you can request to join the project using your GitHub account)


vittorioromeo:
I would strongly recommend to split this in separate PRs. The move operation implementations is something we should tackle separately, and I think what we have in this PR is not optimal. If you remove those changes from the PR, I would be happy to approve.

src/SFML/Audio/SoundStream.cpp
```diff
@@ -82,7 +82,11 @@ void SoundStream::initialize(unsigned int channelCount, unsigned int sampleRate)
     m_channelCount = channelCount;
     m_sampleRate = sampleRate;
     m_samplesProcessed = 0;
-    m_isStreaming = false;
+
+    {
+        std::scoped_lock lock(m_threadMutex);
+        m_isStreaming = false;
+    }
```
vittorioromeo, line 89: Why do we need a lock here? What are we synchronizing against?

src/SFML/Audio/SoundStream.cpp
```diff
@@ -494,8 +498,11 @@ void SoundStream::clearQueue()
 ////////////////////////////////////////////////////////////
 void SoundStream::launchStreamingThread(Status threadStartState)
 {
-    m_isStreaming = true;
-    m_threadStartState = threadStartState;
+    {
+        std::scoped_lock lock(m_threadMutex);
+        m_isStreaming = true;
+        m_threadStartState = threadStartState;
+    }
```
vittorioromeo, line 505: Ditto.

binary1248: In a hypothetical case, the write to the variables might not be visible to the thread that is started below. There are no formal guarantees for this.

src/SFML/Audio/SoundStream.cpp
binary1248: The write to `m_isStreaming` is not guaranteed to be visible to reads from another thread started after this call returns.

vittorioromeo (approved):
LGTM

JonnyPtn:
Should the RenderStates parameter of the draw function have been made const here? I presume it was reported as it's never modified in SFML code, but users with custom drawables may want to modify the texture, transform, shader, etc.

Issue #1996, opened 2022-02-05
X11 miss Delete Key


VitalyVint:
## Subject of the issue

SFML does not react on Delete key (numpad)

## Your environment

* Your OS / distro / window manager used - Linux/Gentoo/X11
* Your version of SFML (2.5.0, git master, etc) - 2.5.1
* Your compiler and compiler version used  - gcc g++ (Gentoo 10.3.0 p1) 10.3.0
* Special compiler flags used

## Steps to reproduce

```cpp
#include <SFML/Graphics.hpp>
#include <iostream>
int main()
{
    sf::RenderWindow window(sf::VideoMode(1280, 720), "Minimal, complete and verifiable example");
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        sf::Event event;
        while (window.pollEvent(event))
        {
            if (event.type == sf::Event::KeyPressed && event.key == sf::Keyboard::Key::Delete)
                std::cout << "delete pressed" << std::endl;
        }

        window.clear();
        window.display();
    }
}
```

## Expected behavior

delete key on num pad pressed and "delete pressed" printed to stdout

## Actual behavior

nothing printed

I found that /usr/include/X11/keysymdef.h defined two delete key constants
XK_Delete and XK_KP_Delete.
SFML-2.5.1/src/SFML/Window/Unix/WindowImplX11.cpp and SFML-2.5.1/src/SFML/Window/Unix/InputImpl.cpp used XK_Delete.

kimci86:
Pressing this key yields events with `sf::Keyboard::Period` keycode.
There is some ongoing work to improve keyboard handling with scancodes in #1235.

VitalyVint:
on my configuration i got 0xffffffff key.code on Delete key (i think it is a 104 key from #1235 ) with or without NumLock.
xev -event keyboard output
```
KeyPress event, serial 28, synthetic NO, window 0x8400001,
    root 0x6bd, subw 0x8400002, time 3786550347, (49,36), root:(1349,714),
    state 0x0, keycode 91 (keysym 0xff9f, KP_Delete), same_screen YES,
    XLookupString gives 0 bytes: 
    XmbLookupString gives 0 bytes: 
    XFilterEvent returns: False
```
and with numlock
```
KeyPress event, serial 28, synthetic NO, window 0x8400001,
    root 0x6bd, subw 0x8400002, time 3786553355, (49,36), root:(1349,714),
[...]

PR #2009, opened 2022-02-15
Fixes for X11 scancode implementation

kimci86:
Here are some fixes for the scancode feature implementation for X11.

## Related to #1235

- Fixed issues with right Alt key.
- Decimal (Numpad) is now detected with `isKeyPressed`.
- Fixed `getDescription` for dead keys (in particular '^' on AZERTY keyboard).

Tested with [SFML-Input](https://github.com/eXpl0it3r/SFML-Input) on Ubuntu 20.04 with QWERTY (UK) and AZERTY (FR) keyboards.
I run it with `XMODIFIERS=@im=none ./SFML-Input` to avoid behavior described in #1894. 

## Related to #1996

Events triggered by Decimal (Numpad) key should now be associated to `sf::Keyboard::Period` more consistently.
@VitalyVint Could you please test this on your system?


src/SFML/Window/Unix/KeySymToUnicodeMapping.cpp
```diff
@@ -405,28 +405,31 @@ Uint32 keysymToUnicode(KeySym keysym)
...
+        case 0xfe5a: return 0x02c7; // 0x030c // dead_caron
         case 0xfe62: return 0x031b; // dead_horn
         case 0x1eff: return 0x0323; // combining_belowdot
         case 0xfe60: return 0x0323; // dead_belowdot
-        case 0xfe5b: return 0x0327; // dead_cedilla
-        case 0xfe5c: return 0x0328; // dead_ogonek
-        case 0xfe5d: return 0x0345; // dead_iota
+        case 0xfe5b: return 0x00b8; // 0x0327 // dead_cedilla
+        case 0xfe5c: return 0x02db; // 0x0328 // dead_ogonek
+        case 0xfe5d: return 0x037a; // 0x0345 // dead_iota
```
eXpl0it3r, line 431: Does the combine character code provide any additional information? 
Otherwise it might make more sense to remove it.
If it's useful, then we might want to change the secondary `//` as it's a bit confusing to have "comments" in comments 😄

kimci86: I think it is nice to show which values from the original table were changed. I will try to make it look better.

eXpl0it3r:

VitalyVint:
Hello, i got sf::Keyboard::Key::Period with and without NumLock pressed.
I think it should be sf::Keyboard::Key::Delete without NumLock?

kimci86:
> I think it should be sf::Keyboard::Key::Delete without NumLock?

I don't know. In the current implementation, keys generally are not sensitive to caps lock or num lock, except the numpad on Windows. (I just tested it in Virtual Box.)
@eXpl0it3r What do you think? Should we unify the behavior across systems?

kimci86:
After some more thinking, I think `Keyboard::Key` values were designed at first to be used to identify the pressed key regardless of the active modifiers.
Now that scancodes are being added, it would make sense to allow modifier keys to have impact `Keyboard::Key` values.

eXpl0it3r (approved):

PR #2012, opened 2022-02-15
Simplify how tests are built

ChrisThrasher:
## Description

Removes unnecessary CMake code which obfuscates the true intent of the build. There is no requirement that source files use absolute paths. CMake will just as happily accept these relative paths. 

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android


vittorioromeo:

vittorioromeo (approved):
LGTM, seems like a nice simplification. Thanks!

PR #2014, opened 2022-02-16
Add move semantics to 'sf::Packet'

vittorioromeo:
## Description

This PR makes `sf::Packet`'s move operations actually move the internal buffer instead of silently copying it. 

> Why doesn't `sf::Packet` actually move the buffer?

https://www.foonathan.net/images/special-member-functions.png

> Can you prove it?

https://gcc.godbolt.org/z/57hnW4fEh

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.


ChrisThrasher:
Couldn't we just remove the empty destructor instead?

vittorioromeo:
> Couldn't we just remove the empty destructor instead?

It's `virtual`, there might be someone out there using `sf::Packet` as a base class...

binary1248:
Based on existing copy assignment operators, I think the style is `Packet& operator =(const Packet&);` (note the space between `operator` and `=`). Disclaimer: I didn't invent this style. 😄

vittorioromeo:
> Based on existing copy assignment operators, I think the style is `Packet& operator =(const Packet&);` (note the space between `operator` and `=`). Disclaimer: I didn't invent this style. 😄

There are 31 occurrences of `operator=` and 44 occurrences of `operator =` in the entire codebase. One more reason to switch to `clang-format` :)

ChrisThrasher:
> There are 31 occurrences of operator= and 44 occurrences of operator = in the entire codebase. One more reason to switch to clang-format :)

#1861 is in a great state!

binary1248:
I think a search in the 2.1 codebase would yield more conclusive results, there was only a single ~de-facto maintainer~ style back then. 😄

binary1248 (approved):

Bromeon:
Should we declare the virtual destructor as `= default`, too?

It doesn't seem to do any TU trickery (like a `T` in `unique_ptr<T>` as an incomplete type in the header).

vittorioromeo:
> Should we declare the virtual destructor as `= default`, too?
> 
> It doesn't seem to do any TU trickery (like a `T` in `unique_ptr<T>` as an incomplete type in the header).

Possibly, but it doesn't really make a difference here. I'd rather remove the `virtual` altogether, to be fair. Who would derive from `sf::Packet` and why...?

kimci86:
> Who would derive from sf::Packet and why...?

There is an example `ZipPacket` in the tutorial: https://www.sfml-dev.org/tutorials/2.5/network-packet.php#custom-packets

vittorioromeo:
> > Who would derive from sf::Packet and why...?
> 
> There is an example `ZipPacket` in the tutorial: https://www.sfml-dev.org/tutorials/2.5/network-packet.php#custom-packets

Hm, interesting. I'd argue that it would make more sense to compress/decompress an entire packet, though... not convinced that's a good use case.

PR #2018, opened 2022-02-16
Remove Cygwin workaround

ChrisThrasher:
## Description

This line was intended to silence a warning, however in the time since that code was added in 95828a85a, the minimum required CMake version was increased from 2.8.3 to 3.8. As of CMake 2.8.4 this warning is no longer emitted which means we can remove this line.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android


eXpl0it3r (approved):

eXpl0it3r:

PR #2020, opened 2022-02-17
Modernize CMake

ChrisThrasher:
## Description

Manually manipulating CMake variable is rather error prone and has fallen out of fashion in favor of newer CMake commands that more easily achieve the same outcome. This PR uses `include_directories` and `link_directories` to simplify how header and linker paths are appended. Eventually we should consider using `target_include_directories` and `target_link_directories` but those aren't simple drop-in replacements and thus out of scope of this PR.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [x] Tested on iOS
* [ ] Tested on Android


binary1248 (approved):

binary1248 (changes requested):
`CMAKE_INCLUDE_PATH` and `CMAKE_LIBRARY_PATH` are used to find packages, not as input to the compiler/linker.

ChrisThrasher:
> CMAKE_INCLUDE_PATH and CMAKE_LIBRARY_PATH are used to find packages, not as input to the compiler/linker.

It appears you're right. My bad. I still wonder if there's a better way to specify those paths without doing so much variable manipulation.

vittorioromeo:
@ChrisThrasher: maybe you use <https://cmake.org/cmake/help/latest/command/list.html#append>?

binary1248:
This is more of a problem on Windows than UNIX-based operating systems since there is no standard filesystem hierarchy for development files on Windows.

If one was hellbent on not having to manually manipulate those variables, additional search paths can be specified in the `find_*` commands, but that would look equally verbose.

ChrisThrasher:
Nevermind, everyone! Turns out I misunderstood what `CMAKE_INCLUDE_PATH` and `CMAKE_LIBRARY_PATH` do.

> @ChrisThrasher: maybe you use https://cmake.org/cmake/help/latest/command/list.html#append?

Yeah I think this is the best we can do. It'll clean things up a bit. I'll try that out.

vittorioromeo (approved):
If CI is happy, this is definitely a (small) improvement.

binary1248 (approved):

PR #2024, opened 2022-02-18
Revert accidental filesystem conversion

ChrisThrasher:
## Description

This fixes a mistake I introduced in #1964. "extension" in this context is not referring to a file extension.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android


eXpl0it3r (approved):
I think, I've looked it and confused it with the MP3 extension checking, so I didn't drill down further 😄 
Glad you caught it

PR #2025, opened 2022-02-18
Simplify CI workflow

kimci86:
## Description

I was bothered by the CI failing in my fork because it cannot upload code coverage results so I added a condition to upload coverage results only from the original repository.

While at it, I simplified the workflow file.
It changes a few things that I thought would be ok:
- Code coverage generation is done in separate jobs.
- ~~Code coverage is generated for shared libraries but no more for static libraries.~~
- ~~Code coverage is no more generated for VS Unity build.~~

@binary1248 Does it look ok?

## How to test this PR?

CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   build:
-    name: ${{ matrix.platform.name }} ${{ matrix.config.name }}
+    name: ${{ matrix.platform.name }} ${{ matrix.config.name }} ${{ matrix.type.name }}
     runs-on: ${{ matrix.platform.os }}
 
     strategy:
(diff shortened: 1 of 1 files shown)


kimci86:
Oops, `Could NOT find OpenCppCoverage (missing: OpenCppCoverage_BINARY)`. I am looking into that.

kimci86:
I got tangled up between bash and powershell, but it is fixed now.

binary1248:
Is there a reason why coverage can't also be measured for static/unity builds? Depending on what the compiler has to generate, the output might lead to a different set of executable paths through the code.

Codecov reports that this changeset reduces the total number of potentially executed lines by quite a bit.

I think it is good to be conservative and run tests/coverage for as many different configurations as possible. We have had bug reports in the past that only occurred for some configurations. Tests are there to make sure we catch these whenever we can and coverage is there to tell us how good our tests are.

kimci86:
> Is there a reason why coverage can't also be measured for static/unity builds? Depending on what the compiler has to generate, the output might lead to a different set of executable paths through the code.

I was hoping shared builds would give more than or the same information as other builds.

> Codecov reports that this changeset reduces the total number of potentially executed lines by quite a bit.

That is bad indeed. Having a closer look, it seems it is caused by lines made only of an opening or closing bracket in the audio module not being reported as tracked lines. I wonder why.

> I think it is good to be conservative and run tests/coverage for as many different configurations as possible.

Ok, I will change the workflow to generate coverage in more configurations.

binary1248 (approved):

kimci86:
Squashed.

PR #2026, opened 2022-02-19
Abort infinite looping in SoundStream::streamData

### LINKED ISSUE #1831: High CPU usage from SoundStream after removal of Audio Device
## Subject of the issue

When an audio device is removed while it is being used by a SoundStream, the playback loop will produce error messages without yielding execution resulting in 100% usage on one CPU core.

## Your environment

* Windows 10 x64
* SFML 2.5.1
* VS2019 x86
* Static runtime

## Steps to reproduce

* Play an sf::Music object
* Open the sound device settings in Windows and disable the device that is currently being used as a default device.

## Expected behavior

The music should stop and CPU usage should not change.

## Actual behavior

The CPU usage on one core rises to 100%. The code producing the problem: https://github.com/SFML/SFML/blob/master/src/SFML/Audio/SoundStream.cpp#L335

Sound playback should be stopped or execution should be yielded while errors are occurring due to the device having been removed.

I don't know if any other momentary errors are expected from alGetSourcei. If alGetSourcei is supposed to never fail, then it would make sense to stop playback entirely instead of continuing and yielding.

Also: SFML doesn't seem to handle audio device changes at all. It would be great if SFML would be able to switch the audio device internally when the user switches the default device in Windows.

binary1248:
Fixes #1831.

It is not uncommon that audio adapters are used that have a USB interface. When all audio adapters are no longer available on Windows, OpenAL will "go crazy". Most OpenAL function calls will cause an OpenAL error to be reported, and any queries will return unpredictable data.

When running the sound example using an OpenAL DLL from the latest release (version 1.21.1) and disabling all audio adapters while music is playing, certain conditions in OpenAL could lead to the main loop in `SoundStream::streamData` never terminating and also never sleeping causing 100% CPU load on that thread.

This PR fixes this problem by forcibly terminating the loop if an OpenAL error is detected.

Steps to reproduce:
1. Build latest SFML with examples
2. Replace the bundled openal32.dll with a DLL from https://openal-soft.org/openal-binaries/openal-soft-1.21.1-bin.zip (rename the DLL to openal32.dll)
3. Run the sound example
4. When the music is playing, disable all audio output devices in Windows
5. If you are (un-)lucky you will end up in an infinite loop, if testing in debug, the console will get flooded so fast with error messages that the program might end up crashing
6. If you can't reproduce the problem, keep repeating steps 3 and 4 until it occurs

### DIFF
--- a/src/SFML/Audio/ALCheck.cpp
+++ b/src/SFML/Audio/ALCheck.cpp
@@ -38,6 +38,15 @@
     #endif
 #endif
 
+namespace
+{
+    // A nested named namespace is used here to allow unity builds of SFML.
+    namespace AlCheckImpl
+    {
+        thread_local ALenum lastError(AL_NO_ERROR);
+    }
+}
+
 namespace sf
 {
 namespace priv
@@ -50,6 +59,8 @@ void alCheckError(const std::filesystem::path& file, unsigned int line, const ch
 
     if (errorCode != AL_NO_ERROR)
     {
+        AlCheckImpl::lastError = errorCode;
+
         std::string error = "Unknown error";
         std::string description = "No description";
 
@@ -99,6 +110,13 @@ void alCheckError(const std::filesystem::path& file, unsigned int line, const ch
               << "\nError description:\n   " << error << "\n   " << description << '\n'
               << std::endl;
     }
+}
+
+
+////////////////////////////////////////////////////////////
+ALenum alGetLastErrorImpl()
+{
+    return std::exchange(AlCheckImpl::lastError, AL_NO_ERROR);
 }
 
 } // namespace priv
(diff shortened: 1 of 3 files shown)


src/SFML/Audio/ALCheck.cpp
```diff
@@ -101,6 +112,17 @@ void alCheckError(const std::filesystem::path& file, unsigned int line, const ch
...
 
+
+////////////////////////////////////////////////////////////
+ALenum alGetLastErrorImpl()
+{
+    ALenum errorCode = AlCheckImpl::lastError;
+
+    AlCheckImpl::lastError = AL_NO_ERROR;
+
+    return errorCode;
```
vittorioromeo, line 123:
```suggestion
    return std::exchange(AlCheckImpl::lastError, AL_NO_ERROR);
```

binary1248: Fixed.

src/SFML/Audio/ALCheck.cpp
```diff
@@ -38,6 +38,15 @@
     #endif
 #endif
 
+namespace
+{
+    // A nested named namespace is used here to allow unity builds of SFML.
+    namespace AlCheckImpl
+    {
+        thread_local ALenum lastError(AL_NO_ERROR);
```
eXpl0it3r, line 46: Does `thread_local` require a `<thread>` include?

eXpl0it3r (approved):

PR #2027, opened 2022-02-21
Add `sf::Span`

kimci86:
## Description

We noticed in #1976 that `std::span` could be helpful to modernize the API.
As it comes with C++20, I tried to use an implementation compatible with C++17 [tcbrindle/span](https://github.com/tcbrindle/span). There are other implementation. I did not review alternatives but this one seems to work great.

## Tasks

* [x] Update Doxygen comments
* [ ] Decide of a suitable license text for `Span.inl` which is based on [tcbrindle/span](https://github.com/tcbrindle/span).
* [ ] Review
* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

- CI
- Run SFML example programs
- How else?


examples/voip/Client.cpp
```diff
@@ -67,12 +67,12 @@ class NetworkRecorder : public sf::SoundRecorder
     /// \see SoundRecorder::onProcessSamples
     ///
     ////////////////////////////////////////////////////////////
-    bool onProcessSamples(const sf::Int16* samples, std::size_t sampleCount) override
+    bool onProcessSamples(std::span<const sf::Int16> samples) override
```
ChrisThrasher, line 70: Is it convention to pass spans by value when you just need to read from the span? I haven't worked with them before.

include/SFML/Audio/Music.hpp
```diff
@@ -140,7 +140,7 @@ class SFML_AUDIO_API Music : public SoundStream
     /// \see openFromFile, openFromStream
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool openFromMemory(const void* data, std::size_t sizeInBytes);
+    [[nodiscard]] bool openFromMemory(std::span<const std::byte> data);
```
ChrisThrasher, line 143: What's the significance of a span of `const T` versus a span of `T`? As in, how does it different from a `const` span of `T`?

include/SFML/Graphics/Transform.inl
```diff
@@ -47,7 +47,7 @@ constexpr Transform::Transform(float a00, float a01, float a02,
 
 
 ////////////////////////////////////////////////////////////
-constexpr const float* Transform::getMatrix() const
+constexpr std::span<const float, 16> Transform::getMatrix() const
```
ChrisThrasher, line 50: 👍 👏 🥳

test/Graphics/Transform.cpp
```diff
@@ -34,7 +34,7 @@ TEST_CASE("sf::Transform class - [graphics]")
             const sf::Transform transform(10.0f, 11.0f, 12.0f,
                                           13.0f, 14.0f, 15.0f,
                                           16.0f, 17.0f, 18.0f);
-            const std::vector<float> matrix(transform.getMatrix(), transform.getMatrix() + 16);
+            const std::vector<float> matrix(transform.getMatrix().begin(), transform.getMatrix().end());
```
ChrisThrasher, line 37: Can we easily initialize a `const std::array<float, 16>` from this span? I'm not sure if there's a one-line way to do that.

include/SFML/Audio/Music.hpp
1aam2am1: Span of const T is like pointer to non chanable object. And const span is non changeable span.

examples/voip/Client.cpp
kimci86: Passing a span by value is similar to passing a pointer and a size (both by value).
This what [C++ Core Guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f24-use-a-spant-or-a-span_pt-to-designate-a-half-open-sequence) recommends.

include/SFML/System/Span.hpp
```diff
@@ -0,0 +1,43 @@
...
+// Include <span> if C++20 or greater is used.
+// Otherwise, use an alternative implementation for older compilers.
+////////////////////////////////////////////////////////////
+#if __cplusplus >= 202002L || (defined(_MSVC_LANG) && _MSVC_LANG >= 202002L)
+
+    #include <span>
+
+#else
+
+    #define TCB_SPAN_NAMESPACE_NAME std
```
Bromeon, line 38: Isn't it [UB to define custom symbols in `std`](https://en.cppreference.com/w/cpp/language/extending_std)?

Bromeon: It would also cause problems if the user in fact uses C++20, with `std::span` available, but defined differently.

kimci86: Right. I guess this could be ok in an application but that is a problem for a library.
So we could have `sf::span` instead for example or `sf::Span` for consistency with SFML's naming conventions.
I wonder if `sf::Span` could be made compatible with `std::span` (the "real" one). That is not absolutely necessary but could be nice.

test/Graphics/Transform.cpp
kimci86: As far as I know, there is no trivial way of doing so. That would require a small helper function based on `std::copy` for example.
Alternatively, what we really want here is to compare spans' pointed-to content as if using `std::equal`, but with `operator==` to have a pretty binary check for doctest. I think this is out of scope for this PR but I will be happy to try it later.

include/SFML/System/Span.hpp
Bromeon: Can we not conditionally compile?
```cpp
#ifdef CPP_AT_LEAST_20
   #include <3rd-party-span.h> // namespace tcb
   namespace sf { 
       using Span = tcb::span;
   }
#else
   #include <span>
   namespace sf { 
       using Span = std::span;
   }
#endif
```

Someone said `__cplusplus` is not reliable on some Visual Studio versions?

ChrisThrasher: Maybe use C++20's feature testing macros?

https://en.cppreference.com/w/cpp/feature_test

The macro for `std::span` is `__cpp_lib_span`.

Bromeon: That's even better -- assuming the feature macros _are_ properly supported on all major compilers.

kimci86: > Can we not conditionally compile?

I am not sure. Could it create issues if the library is compiled for C++17 and used in a C++20 project?
Maybe the C++ version detection could happen only when the library is compiled?

> Someone said __cplusplus is not reliable on some Visual Studio versions?

It needs (or it needed at some point) a command line option to be reliable. See [this Microsoft blog post](https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/). I did not test it though.

kimci86: > However, I think we're best off doing what SFML 2 does by simply reimplementing library features that aren't yet available. In the same way SFML 2 has the equivalent of C++11's thread, SFML 3 will have the equivalent of C++20's span.

I agree, that seems better. In addition, spans are constructible from anything with a `.data()` and `.size()` methods so that should be compatible with `std::span` without doing anything.

include/SFML/Graphics/Glsl.inl
```diff
@@ -81,7 +81,7 @@ struct Matrix
         copyMatrix(transform, *this);
     }
 
-    float array[Columns * Rows]; //!< Array holding matrix data
+    std::array<float, Columns * Rows> array; //!< Array holding matrix data
```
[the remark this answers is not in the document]
Bromeon: I'm not sure if a 2D array is needed or brings us any advantage here. It's quite common in gamedev to represent matrices as 1D arrays.

src/SFML/Audio/SoundFileReaderWav.cpp
```diff
@@ -42,13 +42,13 @@ namespace
 
     bool decode(sf::InputStream& stream, sf::Uint8& value)
     {
-         return static_cast<std::size_t>(stream.read(&value, sizeof(value))) == sizeof(value);
+        return stream.read(sf::asWritableBytes(sf::Span(&value, 1))) == static_cast<sf::Int64>(sizeof(value));
```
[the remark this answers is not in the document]
Bromeon: Agree, and also don't hesitate to declare variables outside the expression 🙂

include/SFML/Graphics/Glsl.inl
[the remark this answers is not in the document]
ChrisThrasher: It's not necessary but it better expresses our intent. While we're currently using a 1D array, what we really mean to express is how there are two independent dimensions here. Matrices are inherently 2 dimensional. Using a 2D array would let us much more clearly write code that needs to access certain elements by their row and column index. It's not immediately obvious what index 13 means since you have to know the size of the matrix and do some mental math, but if you express that as row 3 column 2 then you can much more easily understand what's going on plus that code is much more resilient to potential changes in the size of the array.

include/SFML/Window/WindowBase.hpp
```diff
@@ -282,7 +283,7 @@ class SFML_WINDOW_API WindowBase
     /// \see setTitle
     ///
     ////////////////////////////////////////////////////////////
-    void setIcon(unsigned int width, unsigned int height, const Uint8* pixels);
+    void setIcon(Span<const Uint8> pixels, Vector2u size);
```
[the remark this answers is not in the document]
kimci86: Well spotted, I wanted to discuss this.
I wanted to change to order to be more consistent with other methods:
```c++
bool Cursor::loadFromPixels(Span<const Uint8> pixels, Vector2u size, Vector2u hotspot);
bool Texture::loadFromMemory(Span<const std::byte> data, const IntRect& area);
```
And this one that I missed could be changed:
```c++
void Image::create(unsigned int width, unsigned int height, Span<const Uint8> pixels);
```

src/SFML/Audio/SoundFileReaderWav.cpp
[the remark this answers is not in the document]
kimci86: The value 1 means that we make a span one element. Then `asWritableBytes` makes the necessary multiplication by the type's size to make a span over a memory region.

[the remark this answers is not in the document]
ChrisThrasher: My bad. Ignore my other comments about that then.

include/SFML/Window/WindowBase.hpp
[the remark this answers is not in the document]
ChrisThrasher: Sounds good to me. Let's collapse those `unsigned int`s into a `Vector2u` in `Image::create` while we're at it. I guess the other 3 commits that aimed to replace these number pair interfaces with vectors missed this one and a few others you've fixed.

vittorioromeo:
I think it would be worthwhile to measure the compilation time impact on this PR, either with `time` or with `ClangBuildAnalyzer`.

ChrisThrasher:
@vittorioromeo 

Compile time in seconds compared against the master branch commit from which this branch was started.

| | make | make -j10 |
| -- | -- | -- |
| fab2c93d7a56e85f60e26f065c051630ffff420e | 41.865 | 7.330 |
| 0cb5cf9b7758501acec4c074e218924dbf52f1e5 | 43.001 | 7.452 |

Fairly negligible impact in my not especially scientific measurements that I only ran once. Absolutely worth the additional expressiveness and power this adds to the interface.

include/SFML/Window/WindowBase.hpp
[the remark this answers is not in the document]
kimci86: There are in fact more methods that take two unsigned ints and could take a `Vector2u` instead.
- `Image::create`
- `Image::copy`
- `Image::setPixel`
- `Image::getPixel`
- `RenderTexture::create`
- `Texture::create`
- `Texture::update`

I prefer not to change those in this already huge PR and keep it for later.
Changing `WindowBase::setIcon` was an easy case as is has little impact and improves consistency within the window module (consistent with `Cursor` class).

include/SFML/Graphics/Glsl.inl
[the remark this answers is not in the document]
kimci86: I am also not sure it is worth doing. I agree it is semantically more correct to implement a matrix as a 2D array. However, a span over a 2D array would still be 1D (unless we add a multi-dimensional span class) so we would still have to use 1D indices at some places.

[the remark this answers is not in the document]
kimci86: There is no obvious consensus and this can be kept for later so I resolve this discussion for now.

include/SFML/Graphics/Transform.hpp
```diff
@@ -268,7 +269,7 @@ class Transform
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    float m_matrix[16]; //!< 4x4 matrix defining the transformation
+    std::array<float, 16> m_matrix; //!< 4x4 matrix defining the transformation
```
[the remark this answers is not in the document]
kimci86: Let's keep it for later.

kimci86:
As there was no negative feedback, I assume this has chances to get merged, so I updated the doxygen comments. This is now ready to be reviewed :)

eXpl0it3r:

include/SFML/System/Span.inl
```diff
@@ -0,0 +1,351 @@
...
+
+/*
+This is an implementation of C++20's std::span
+http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/n4820.pdf
+*/
+
+//          Copyright Tristan Brindle 2018.
+// Distributed under the Boost Software License, Version 1.0.
+//    (See accompanying file ../../LICENSE_1_0.txt or copy at
+//          https://www.boost.org/LICENSE_1_0.txt)
```
eXpl0it3r, line 12: Is the Boost license really compatible with SFML's zlib license?

include/SFML/System/Span.inl
```diff
@@ -0,0 +1,351 @@
+
+// TODO license text for a modified work
```
eXpl0it3r, line 2: TODO 😉

kimci86: I am not sure what to do here.

include/SFML/System/Span.inl
Bromeon: These are the relevant parts:

> **Permission is hereby granted**, free of charge, to any person or organization
obtaining a copy of the software and accompanying documentation covered by
this license (the "Software") to use, reproduce, display, distribute,
execute, and transmit the Software, and **to prepare derivative works of the
Software**, and to permit third-parties to whom the Software is furnished to
do so, all **subject to the following:**
> 
> **The copyright notices in the Software and this entire statement**, including
the above license grant, this restriction and the following disclaimer,
**must be included in all copies of the Software**, in whole or in part, and
all derivative works of the Software, unless such copies or derivative
works are solely in the form of machine-executable object code generated by
a source language processor.
>
> [no warranty etc]

So, looks like you can do anything, but need to ship the license statement somewhere. I don't know if linking to it counts as "shipping" in legal terms.

eXpl0it3r:
Personally I would want to have a broader discussion on whether this is the right/wanted way to go.

Should we really implement and adapt a future library/language feature? Is there any data that this feature is actually being adapted by others? Is it the "best" alternative or just the only standardized one? Are the new APIs actually more intuitive/easier/less error prone to use? etc

If @SFML/sfml team members could chip in as well 🙂

binary1248:
SFML has always been rather conservative with language feature usage. It only took *searches through discussion history* ... many ... years for us to make the switch to "modern C++" without actually having an explicit language standard target, just C++11 or newer.

I myself for my projects use whatever is available to me in the development environment I am currently bound to. If certain future language features that I find very useful aren't available yet out of the box I use the polyfills (yes... this is the proper term) that are available to bridge the gap until built-in support is available.

All that being said, this is a decision I made for my own project. I know what environment I am working on, I have a plan on how long I want to use said polyfills and I know the cost I pay for using them. These are all based on assumptions that I can make for my own project. Said assumptions are harder to make when we are talking about a library.

In contrast to application code, library code has to run on a multitude of platforms, build in environments that have varying standard support and on top of it all, boast a certain level of stability so that the users of said library don't have to be surprised by changes all too often.

I don't find it very effective to write our own polyfill just to bridge the gap for a feature that has been standardized but does not yet enjoy ubiquitous support. Because the feature has already been standardized it is not a question of if but when the sweeping majority of developers will have access to it. Like I said above, if I had an application I have to work on and I know that development will be "done" in a foreseeable time frame i.e. before said language feature makes it into my development environment then it would make sense to use a polyfill in the application. Because SFML development shouldn't end in the next few years *knock on wood* I don't see why we should be impatient and try to push usage of `std::span` forward when it will come naturally anyway.

We've already converted many legacy SFML 2.x utilities (thread, etc.) to their C++11 versions. These utilities were written in a time long before said features were even standardized let alone available with major compiler implementations. What we are trying to do with `std::span` is the opposite, implementing our own version after it has already been standardized, which makes little sense to me.

If we were to take a conservative stance as to which compiler versions we still fully want to support, I would say the last 2 Visual Studio versions and whatever GCC is available on Debian stable. Apple Clang has always been a mixed bag (they actually tie language standard support to the version of the operating system if you can believe it) so I don't have an opinion on it.

kimci86:
Thank you for your feedback.
If adding a polyfill is not acceptable, then there is nothing I can do except wait for C++20 to be used.
Should I close this PR?

eXpl0it3r:

Bromeon:
I wouldn't say it's off the table yet. But as https://github.com/SFML/SFML/pull/2055#issuecomment-1092052955 has shown, there needs to be some general discussion on the topic:
> Should SFML emulate idioms that are helpful in certain situations, even if it means providing the library infrastructure for them?

And with it, how much ease-of-use are we willing to accept, at which cost? (Unknown-ness of the idiom, build and run time, pitfalls, ...). binary1248 gave already a great answer on this matter.

ChrisThrasher:
I appreciate your PR but the team has decided that we will not make this change in v3. By the time v4 comes around we will be using `std::span`. The idea of this PR is great, we choose to wait until `std::span` is available to act on it.

PR #2028, opened 2022-02-24
Add sin/cos/tan helpers for sf::Angle

ChrisThrasher:
## Description

I noticed there were many instances of using `asRadians` just to use standard trigonometry functions so I added some additional functions that make it more easy to use `sf::Angle` with trig functions. This is a nice bit of syntactic sugar but also avoids the pitfalls of users calling `asRadians` which eschews the type safety of `sf::Angle`. We'd prefer any use of `sf::Angle` be in a type safe manner.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/examples/X11/X11.cpp
+++ b/examples/X11/X11.cpp
@@ -47,7 +47,7 @@
     // Setup a perspective projection
     glMatrixMode(GL_PROJECTION);
     glLoadIdentity();
-    float extent = std::tan(sf::degrees(45).asRadians());
+    float extent = sf::tan(sf::degrees(45));
 
 #ifdef SFML_OPENGL_ES
     glFrustumf(-extent, extent, -extent, extent, 1.0f, 500.0f);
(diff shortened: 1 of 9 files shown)


eXpl0it3r:
I'm not convinced of this change. From my point of view this more API bloat than anything else. It puts the whole usage of `sf::Angle` in question, when we internally write short hands, because we consider the API too long/verbose. Additionally, while sin, cos and tan are the most common functions, what about asin, acos, atan, atan2, sinh, cosh, tanh, asinh, acosh, atanh? Or what about some other functions that takes angle?
On top of that, I really don't want SFML to pretend to be a math library, which by implementing our own version of sin, cos and tan you start to blur the line.

Bromeon:
Once https://github.com/SFML/SFML/pull/1979 is merged, the need to explicitly use `sin`/`cos` will also be reduced. Maybe not so much for SFML itself, but definitely for the user. 

In this regard, I see vector APIs as an abstraction layer above trigonometric functions.

PR #2030, opened 2022-02-28
Simplify CMake build by removing source_group and headers as source files

ChrisThrasher:
## Description

Strictly speaking CMake puts no requirement on using `source_group` or listing headers as source files. As I've heard, both of these practices are conveniences to Visual Studio users. Having never used Visual Studio myself I struggle to understand how much of a convenience this is. I'm submitting this PR to get more feedback on what VS users want and to demonstrate exactly how much code can be simplified if we're willing to stop doing these things.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Build this PR in Visual Studio and try to observe any difference in your workflow, particularly if this PR makes SFML development more difficult or annoying. Share any specific observations you make. Mention what version of Visual Studio you are using.

### DIFF
--- a/cmake/CompilerWarnings.cmake
+++ b/cmake/CompilerWarnings.cmake
@@ -3,7 +3,7 @@
 # https://github.com/lefticus/cppbestpractices/blob/master/02-Use_the_Tools_Available.md
 
 # Helper function to enable compiler warnings for a specific set of files
-function(set_file_warnings)
+function(set_file_warnings target)
     option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" TRUE)
 
     set(MSVC_WARNINGS
@@ -106,7 +106,5 @@ function(set_file_warnings)
         message(AUTHOR_WARNING "No compiler warnings set for '${CMAKE_CXX_COMPILER_ID}' compiler.")
     endif()
 
-    foreach(WARNING ${FILE_WARNINGS})
-        set_property(SOURCE ${ARGV} APPEND_STRING PROPERTY COMPILE_FLAGS " ${WARNING}")
-    endforeach()
+    target_compile_options(${target} PRIVATE ${FILE_WARNINGS})
 endfunction()
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -51,12 +51,11 @@ endfunction()
 
 # add a new target which is a SFML library
 # example: sfml_add_library(Graphics
-#                           SOURCES sprite.cpp image.cpp ...
 #                           [STATIC]) # Always create a static library and ignore BUILD_SHARED_LIBS
 macro(sfml_add_library module)
 
     # parse the arguments
-    cmake_parse_arguments(THIS "STATIC" "" "SOURCES" ${ARGN})
+    cmake_parse_arguments(THIS "STATIC" "" "" ${ARGN})
     if (NOT "${THIS_UNPARSED_ARGUMENTS}" STREQUAL "")
         message(FATAL_ERROR "Extra unparsed arguments when calling sfml_add_library: ${THIS_UNPARSED_ARGUMENTS}")
     endif()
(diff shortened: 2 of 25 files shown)


eXpl0it3r:

eXpl0it3r:

vittorioromeo:
My two cents: I don't use Visual Studio to work on SFML, but even if I did I would prefer to see a large simplification of the build scripts rather than a few grouped files in the solution explorer.

ChrisThrasher:
The Windows Unity builds are failing now because of a local variable shadowing a global. This makes me think that the previous way we were applying compiler warnings did not actually apply warnings to all source files, otherwise this warning would have been caught earlier.

eXpl0it3r:

ChrisThrasher:
https://github.com/ChrisThrasher/SFML/commit/1e99ee4ea8e1a093d9c3d4be45e63eee7650f9c8

I have another commit tweaking some CMake stuff that I can add to this PR if you'd rather it not become a PR of its own later.

binary1248:
So... what would be the straight-forward way of opening header files in IDEs with this change? And no... i'm not counting "Open File ..." or right-click open header, etc.

ChrisThrasher:
> So... what would be the straight-forward way of opening header files in IDEs with this change? And no... i'm not counting "Open File ..." or right-click open header, etc.

Use the file system navigator in your IDE to open any given file in the include/ directory.

binary1248:
If you mean e.g. the "Solution Explorer" pane in VS then I only see the source files as can be seen in @eXpl0it3r's screenshot. I would personally find it cumbersome having to manually open header files through the File -> Open dialog. It makes it even more annoying knowing it used to be easier at some point in the past.

I don't see why we should all of a sudden treat .hpp files differently than .cpp files, especially now that we are working more actively on modernizing the codebase.

ChrisThrasher:
I don't mean the Solution Explorer. I mean just a simple view of all files in the repo. I don't use Visual Studio so I don't know exactly what that might be called. In VS Code it looks like this. 

image

Does Visual Studio not provide something similar?

eXpl0it3r:
You can switch to a file system view, but that defeats part of the point of having a project generated.

For VS I really think it's not a big issue anymore, because opening the folder directly and having CMake automatically triggered is in many ways superior than to generate a project file and working with that.

One question that wasn't asked is, whether there are tools that also depend on having headers also listed in the source list (e.g. some static analyzer)

ChrisThrasher:
> One question that wasn't asked is, whether there are tools that also depend on having headers also listed in the source list (e.g. some static analyzer)

In my experience, static analyzers don't directly analyze header files but rather analyze them in the context of the .cpp files that include them. This the case with clang-tidy at least. The clang-format PR (#1861) uses CMake to collect a list of source files (which includes headers) to pass to clang-format. We can reuse that code to pass a list of all .hpp and .cpp to whatever other tools we want to use.

binary1248:
Back before CMake even existed, the way of adding files to your project was by explicitly adding them to the vcproj so that they would be displayed as part of the project/solution, hence why the pane that lists all tracked files is called the "Solution Explorer". Files that aren't explicitly added to the project will not be visible.

When VS added CMake support I tried it out and simply put, it wasn't a very good experience. It has gotten better, but an undeniable fact is that VS will always trail standalone CMake when it comes to supporting newer features (and even more so if you have to use an experimental version of CMake for whatever reason). I've never had a problem running the initial solution generation using CMake and never having to open it up until I really need to change some options or wipe my build directory. I don't see why we should put our energy into invalidating a method of working that has worked since the beginning of CMake support for VS.

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

friendlyanon:
Sort of relevant: https://github.com/friendlyanon/cmake-init/issues/39

In particular the command there could be of interest to you. It combines `target_sources` and `source_group` into a single command, so that eliminates variables. Maybe this would introduce better DX for both sides. Headers will have to be listed either way though.

ChrisThrasher:
I'm not getting much support in simplifying the build so I'll close this PR. Happy to reevaluate this later if the team reconsiders.

PR #2031, opened 2022-02-28
Use Pitchfork project layout

ChrisThrasher:
## Description

https://api.csswg.org/bikeshed/?force=1&url=https://raw.githubusercontent.com/vector-of-bool/pitchfork/develop/data/spec.bs

EDIT: Try this link if the above link isn't loading: https://github.com/vector-of-bool/pitchfork/blob/develop/data/spec.bs

It helps to follow community conventions for the sake of providing an easy onboarding process for new contributors and team members and minimizing cognitive load when switching between projects. The Pitchfork Project Layout is the closest thing C++ has to a standard directory layout. SFML already adheres to most major tenets of this layout, hence this relatively small diff. Because we're so close to already following this, I think it's reasonable to take the last few steps to totally adhere to it.

doc -> docs
extlibs -> external
test -> tests

This draft PR is mostly about having a conversation about this. These kinds of directory layout issues are in fact bikeshedding so I won't make us talk about this for too long but I want to submit this in case the changes are well-received.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android


eXpl0it3r:

Bromeon:
Is this convention used significantly in the C++ ecosystem? Otherwise it's just one of many "standards". Are there competing alternatives?

That said, regarding directory structure, it's probably better to follow an existing convention than to have our own.

ChrisThrasher:
> Is this convention used significantly in the C++ ecosystem? Otherwise it's just one of many "standards". Are there competing alternatives?

Many C++ projects I come across use this layout or something closely resembling it. I've never heard of any competing standards or conventions. Seemingly nobody besides vector-of-bool has taken the time to document and rationalize their idea for a standard project layout.

kimci86:
> I've never heard of any competing standards or conventions.

There is also at least this one: [Canonical Project Structure](http://open-std.org/JTC1/SC22/WG21/docs/papers/2018/p1204r0.html).

eXpl0it3r:
From a value point of view, this doesn't change anything. Even if there was a layout that's more commonly used, I highly doubt anyone would fail to identify `extlibs` as `external` etc.

The docs & tests I don't really mind changing, but again value-wise, we don't gain anything here.

Personally, I'd first much rather see how we want to handle dependencies going forward. This would then also put the whole `extlibs` directory into question. In case we'd still ship header-only files, it might be more fitting to use `external` instead of `extlibs` as the `libs` part could be more misleading.

Bromeon:
The link in the original post is blank now. It's still possible to directly open the part behind `url=`, but rather hard to read.

ChrisThrasher:
@eXpl0it3r 

> From a value point of view, this doesn't change anything. Even if there was a layout that's more commonly used, I highly doubt anyone would fail to identify extlibs as external etc.

No one thinks that `extlibs` is illegible. SFML's layout is already quite good. The goal of this PR is simply to make a modest improvements over the status quo.

> Personally, I'd first much rather see how we want to handle dependencies going forward. This would then also put the whole extlibs directory into question. In case we'd still ship header-only files, it might be more fitting to use external instead of extlibs as the libs part could be more misleading.

The directory name `external/` has the nice property of being correct and descriptive whether or not the directory contains compiled libraries or merely some header files.

@Bromeon 

> The link in the original post is blank now. It's still possible to directly open the part behind url=, but rather hard to read.

That's so weird. I edited the PR description to share the link you described even if it's hard to read.

ChrisThrasher:
It seems that I'm alone in wanting to adhere to this project layout so I'll close this PR.

Bromeon:
Further up I mentioned:

> Is this convention used significantly in the C++ ecosystem? Otherwise it's just one of many "standards". 

That was not an argument against the PR, more to see if this is the best-possible choice, or if there's something even more widespread.

I think this PR is a valid proposal. Not everything has to be game-changing.

I also don't know what kind of thought went into the original naming, could be that these were just the most straightforward names around at the time. The only drawback I see is that third-party script/code that relies on a certain SFML structure might break -- but often that's either questionable practice or breaks with SFML 3 already.

PR #2032, opened 2022-02-28
Use 'ccache' if available

vittorioromeo:
## Description

This PR enables the use of `ccache` by default if available on the current machines. `ccache` *massively* speeds up builds when most of the files have not changed. It is a beneficial default, and we should also make sure `ccache` is installed on our CI machines.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,13 @@ project(SFML VERSION 3.0.0)
 
 set(VERSION_IS_RELEASE OFF)
 
+# use ccache if available
+find_program(CCACHE_PROGRAM ccache)
+if(CCACHE_PROGRAM)
+    message(STATUS "Found ccache in ${CCACHE_PROGRAM}")
+    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_PROGRAM}")
+endif()
+
 # include the configuration file
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/Config.cmake)
 


CMakeLists.txt
```diff
@@ -1,5 +1,12 @@
 cmake_minimum_required(VERSION 3.8)
 
+# use ccache if available
+find_program(CCACHE_PROGRAM ccache)
+if(CCACHE_PROGRAM)
+    message(STATUS "Found ccache in ${CCACHE_PROGRAM}")
+    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_PROGRAM}")
+endif()
+
```
ChrisThrasher, line 9: Does this need to happen before the call to `project()`?

vittorioromeo: I don't think so. Should it be moved below?

ChrisThrasher: Preferably so. By putting it before `project()` we're insinuating that it _must_ be there when that is likely not the case. I'd prefer to put as much code after `project()` as possible since most code need not exist before it.

eXpl0it3r:

Bromeon (commented):
Good idea! There's also [sccache](https://github.com/mozilla/sccache) as an alternative, but I don't have experience with either of the two.

Regarding CI, using `ccache` alone would probably not help, since the runner is freshly initialized. It's probably necessary to combine them with a caching GitHub action, which persists already compiled things over runs.

But the caching in CI should probably be limited to external dependencies, while SFML itself is rebuilt from scratch.

Bromeon:
Also, should we maybe use [`CMAKE_CXX_COMPILER_LAUNCHER`](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_LAUNCHER.html)?

CMakeLists.txt
vittorioromeo: Done

vittorioromeo:
> Regarding CI, using `ccache` alone would probably not help, since the runner is freshly initialized. It's probably necessary to combine them with a caching GitHub action, which persists already compiled things over runs.

I think figuring out a way to use `ccache` in our CI system would lead to great speedups. If we cannot do it on GitHub actions, can we install `ccache` on our custom machines as a starter?
 
But there seem to be solutions for `ccache` on GitHub actions:
- https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/
- https://github.com/hendrikmuhs/ccache-action

@binary1248, @eXpl0it3r: want to take a look?

> But the caching in CI should probably be limited to external dependencies, while SFML itself is rebuilt from scratch.

Why? We would lose on the performance benefits of `ccache` for our own code. As far as I understand, it is smart enough to only cache things that are completely unchanged.

> Also, should we maybe use [CMAKE_CXX_COMPILER_LAUNCHER](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_LAUNCHER.html)?

As far as I can tell from my non-extensive research, the technique I used is the recommended one to enable `ccache` on a project-wide basis with CMake.

Bromeon:
> I think figuring out a way to use `ccache` in our CI system would lead to great speedups.

Yes, just meant that we would need somewhere to store the cache, the runners are stateless by default. I'm not familiar with the custom machine setup, maybe they're stateful and could do the job.

> Why? We would lose on the performance benefits of `ccache` for our own code. As far as I understand, it is smart enough to only cache things that are completely unchanged.

There's a few things we need to keep in mind when caching a distributed system:
* How does this work if 3 branches are open in parallel, and each has their own set of changes? Will the push of one branch invalidate the other branches' state? Can there be something like race conditions? Or can we set the key in a way that each branch gets their own local state?
* If we cache the SFML code, it means we have to store back the cache after each commit (as opposed to only on dependency change). This is a trade-off and depends on how costly storing is. It also takes more disk space (thus increasing it time spent for network transfer). See also [here](https://github.com/Swatinem/rust-cache/issues/37#issuecomment-944697938).
* And last, is `ccache` 100% bulletproof without corner cases? Does it recognize not just code changes, but also configuration/compiler flags/anything that could influence the code? If not, we risk that our CI becomes non-deterministic.

We can gladly give it a try and experiment with different setups 🙂 one option would also be to cache precompiled headers (once we have them).

vittorioromeo:
> How does this work if 3 branches are open in parallel, and each has their own set of changes? Will the push of one branch invalidate the other branches' state? Can there be something like race conditions? Or can we set the key in a way that each branch gets their own local state?

`ccache` works by hashing the contents of source files: https://ccache.dev/manual/4.3.html#_how_ccache_works
So having multiple branches living at the same time is not an issue, different modifications will result in different hashes.

> If we cache the SFML code, it means we have to store back the cache after each CI run (as opposed to only on dependency change). It also takes more disk space (thus increasing it time spent for network transfer). 

I am not sure how GitHub actions work, where the cache would be stored, etc. But `ccache` allows you to set a limit on the size of the cache quite easily, as seen from the docs.

> And last, is ccache 100% bulletproof without corner cases? Does it recognize not just code changes, but also configuration/compiler flags/anything that could influence the code? 

`ccache` has been around and actively maintained for almost 20 years -- I would expect it to be robust. It never failed me for my own projects and we do use it at work on large projects as well. I cannot guarantee it will be 100% bulletproof, but I'm confident that in the extremely rare eventuality that we find an edge case, the time spent working around it will be minimal compared to the time saved avoiding unnecessary recompilation.

vittorioromeo:
Fair enough, I would advise reconsidering the use of this in CI, but I'm fine if we merge this just for local development as is.

CMakeLists.txt
ChrisThrasher: Did you mean to move this after `project()`?

vittorioromeo: @ChrisThrasher, sorry -- should be fine now.

eXpl0it3r (approved):

ChrisThrasher:
I'm still unsure why this is preferable when [`CMAKE_CXX_COMPILER_LAUNCHER`](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_LAUNCHER.html#variable:CMAKE_%3CLANG%3E_COMPILER_LAUNCHER) exists.

I'd rather keep our build scripts minimum and not impose particular compiler launchers automatically. That's up to users to decide and not for us to impose on them.

PR #2034, opened 2022-03-01
Use new sf::Vector2<T> utilities

ChrisThrasher:
## Description

I found a few more places where we can make use of newly added vector utilities to remove now-duplicate code.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/src/SFML/Graphics/Shape.cpp
+++ b/src/SFML/Graphics/Shape.cpp
@@ -36,17 +36,11 @@ namespace
     // Compute the normal of a segment
     sf::Vector2f computeNormal(const sf::Vector2f& p1, const sf::Vector2f& p2)
     {
-        sf::Vector2f normal(p1.y - p2.y, p2.x - p1.x);
-        float length = std::sqrt(normal.x * normal.x + normal.y * normal.y);
+        sf::Vector2f normal = (p2 - p1).perpendicular();
+        float length = normal.length();
         if (length != 0.f)
             normal /= length;
         return normal;
-    }
-
-    // Compute the dot product of two vectors
-    float dotProduct(const sf::Vector2f& p1, const sf::Vector2f& p2)
-    {
-        return p1.x * p2.x + p1.y * p2.y;
     }
 }
 
@@ -280,9 +274,9 @@ void Shape::updateOutline()
 
         // Make sure that the normals point towards the outside of the shape
         // (this depends on the order in which the points were defined)
-        if (dotProduct(n1, m_vertices[0].position - p1) > 0)
+        if (n1.dot(m_vertices[0].position - p1) > 0)
             n1 = -n1;
-        if (dotProduct(n2, m_vertices[0].position - p1) > 0)
+        if (n2.dot(m_vertices[0].position - p1) > 0)
             n2 = -n2;
 
         // Combine them to get the extrusion direction


eXpl0it3r:

src/SFML/Graphics/Shape.cpp
```diff
@@ -275,14 +255,14 @@ void Shape::updateOutline()
         Vector2f p2 = m_vertices[index + 1].position;
 
         // Compute their normal
-        Vector2f n1 = computeNormal(p0, p1);
-        Vector2f n2 = computeNormal(p1, p2);
+        Vector2f n1 = (p0 - p1).perpendicular();
+        Vector2f n2 = (p1 - p2).perpendicular();
```
vittorioromeo, line 259: Are you sure we want `perpendicular` here and not `normalized`?

vittorioromeo (changes requested):

src/SFML/Graphics/Shape.cpp
ChrisThrasher: I believe so. I double checked the implementation of `computeNormal` on paper and saw that it was returning a vector perpendicular to the vector defined by the two points passed as arguments. "Normal" in this context is referring to this:

https://en.wikipedia.org/wiki/Normal_(geometry)

vittorioromeo: But the original code was also computing the length and dividing the vector by such length. Was that unnecessary?

ChrisThrasher: Good point. Idk how necessary the normalization is but I'll add that back since I can't confirm whether or not we need it.

ChrisThrasher: Fixed now. If you'd rather I not add a local lambda and instead keep that as a free function then I can do that to minimize the diff.

src/SFML/Graphics/Shape.cpp
```diff
@@ -275,14 +255,21 @@ void Shape::updateOutline()
         Vector2f p2 = m_vertices[index + 1].position;
 
         // Compute their normal
-        Vector2f n1 = computeNormal(p0, p1);
-        Vector2f n2 = computeNormal(p1, p2);
+        const auto findNormal = [](const sf::Vector2f& vector)
+        {
+            if (vector.lengthSq() == 0.f)
+                return vector;
```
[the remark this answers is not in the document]
ChrisThrasher: This case is unfortunately possible when `i` equals 0.
```cpp
        Vector2f p0 = (i == 0) ? m_vertices[count].position : m_vertices[index - 1].position;
        Vector2f p1 = m_vertices[index].position;
``` 
In that case `p0` and `p1` are both `m_vertices[index].position`. I feel like the loop index should start at 1 instead of 0 to prevent that. Unfortunately we haven't yet gotten around to testing `sf::Shape` so we can't lean on our unit tests to confirm that we didn't break anything.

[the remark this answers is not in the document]
Bromeon: Yes, the `(i == 0)` condition is confusing, a more standard approach here would be
`m_vertices[(index - 1) % count].position`.

I guess it _is_ possible to have duplicate vertices, e.g. by using [`sf::ConvexShape`](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1ConvexShape.php) and manually inserting the same point twice. Which could already be seen as another invalidated variant, but we don't check this.

What does everyone think -- is this something worth hardening? If yes, we could open an issue so we don't forget it. If no, no action is needed.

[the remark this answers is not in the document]
vittorioromeo: I don't think it's worth addressing, the checks would introduce some overhead and I don't think anyone has ever complained about this issue.

src/SFML/Graphics/Shape.cpp
```diff
@@ -275,14 +255,21 @@ void Shape::updateOutline()
...
+            return vector.perpendicular().normalized();
```
vittorioromeo, line 262: I believe that this PR pessimizes the performance of this code, as now we're computing the length twice. Maybe it is worthwhile only using the `.dot` member function and keeping the older code for the perpendicular+normalized?

ChrisThrasher: That's fair. I just pushed a simpler version of this PR.

vittorioromeo (approved):

vittorioromeo:

PR #2035, opened 2022-03-06
Add tests for sf::VertexArray

ChrisThrasher:
## Description

The journey to test the Graphics module continues with `sf::VertexArray`. Since this is mostly just a wrapper for `std::vector` it was fairly simple. I was able to write all the tests without reading the implementation which is always a good sign. 

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/test/Graphics/VertexArray.cpp
+++ b/test/Graphics/VertexArray.cpp
@@ -1,0 +1,104 @@
+#include <SFML/Graphics/VertexArray.hpp>
+#include "GraphicsUtil.hpp"
+
+#include <doctest.h>
+
+TEST_CASE("sf::VertexArray class - [graphics]")
+{
+    SUBCASE("Construction")
+    {
+        SUBCASE("Default constructor")
+        {
+            const sf::VertexArray vertexArray;
+            CHECK(vertexArray.getVertexCount() == 0);
+            CHECK(vertexArray.getPrimitiveType() == sf::PrimitiveType::Points);
+            CHECK(vertexArray.getBounds() == sf::FloatRect({0, 0}, {0, 0}));
+        }
+
+        SUBCASE("Explicit constructor with default argument")
+        {
+            const sf::VertexArray vertexArray(sf::PrimitiveType::Lines);
+            CHECK(vertexArray.getVertexCount() == 0);
+            CHECK(vertexArray.getPrimitiveType() == sf::PrimitiveType::Lines);
+            CHECK(vertexArray.getBounds() == sf::FloatRect({0, 0}, {0, 0}));
+        }
+
+        SUBCASE("Explicit constructor")
+        {
+            const sf::VertexArray vertexArray(sf::PrimitiveType::Lines, 10);
+            CHECK(vertexArray.getVertexCount() == 10);
+            CHECK(vertexArray.getPrimitiveType() == sf::PrimitiveType::Lines);
+            CHECK(vertexArray.getBounds() == sf::FloatRect({0, 0}, {0, 0}));
+            for (std::size_t i = 0; i < vertexArray.getVertexCount(); ++i)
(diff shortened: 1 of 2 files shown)


test/Graphics/VertexArray.cpp
```diff
@@ -0,0 +1,88 @@
...
+                CHECK(vertexArray[i].texCoords == sf::Vertex().texCoords);
+            }
+        }
+    }
+
+    SUBCASE("Resize array")
+    {
+        sf::VertexArray vertexArray;
+        vertexArray.resize(42);
+        CHECK(vertexArray.getVertexCount() == 42);
```
Bromeon, line 45: Should we check the value of an inserted element (e.g. at index 41)?

ChrisThrasher: The explicit constructor tests already test that, I believe. I loop through all the vertices and make sure their values match that of a default constructed `sf::Vertex`.

Bromeon: Yes, but the constructor and `resize()` are two separate APIs.

Maybe we can extract a function for "compare equality of `position`, `color` and `texCoords`".

ChrisThrasher: Fair. I just realized that the explicit constructor isn't actually calling `resize()` in the implementation. I added those checks to this test.

Bromeon: Even if it were calling `resize()` internally -- strictly speaking, this is an implementation detail that can change at any time and is exactly the reason why we have tests 🙂 

(We have quite someconfidence that the _current_ implementation works, it's more if anything would ever change. Rather unlikely in this case though.)

eXpl0it3r:

test/Graphics/VertexArray.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+        const sf::Vertex otherVertex({2, 3}, {4, 5, 6, 7}, {8, 9});
+        vertexArray[0] = otherVertex;
+        CHECK(vertexArray.getVertexCount() == 1);
+        CHECK(vertexArray[0].position == otherVertex.position);
+        CHECK(vertexArray[0].color == otherVertex.color);
+        CHECK(vertexArray[0].texCoords == otherVertex.texCoords);
+    }
+
+    SUBCASE("Set primative type")
```
eXpl0it3r, line 74: primitive*

test/Graphics/VertexArray.cpp
```diff
@@ -0,0 +1,94 @@
...
+        vertexArray.resize(42);
+        CHECK(vertexArray.getVertexCount() == 42);
+        for (std::size_t i = 0; i < vertexArray.getVertexCount(); ++i)
+        {
+            CHECK(vertexArray[i].position == sf::Vertex().position);
+            CHECK(vertexArray[i].color == sf::Vertex().color);
+            CHECK(vertexArray[i].texCoords == sf::Vertex().texCoords);
+        }
+        vertexArray.clear();
+        CHECK(vertexArray.getVertexCount() == 0);
```
eXpl0it3r, line 53: Since clearing is a separate operation, I'd move it into a separate test, otherwise one doesn't really know that it's tested when looking at the cases

eXpl0it3r (changes requested):
Two small changes in the way the cases are structured and a typo, otherwise looking good 👍

eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:

eXpl0it3r:
Thanks for this new set of tests 🥳

PR #2036, opened 2022-03-06
Mark external headers as SYSTEM headers

ChrisThrasher:
## Description

Past PRs have made a point to mark external include directories as `SYSTEM` include directories. This PR fixes the last few instances where `SYSTEM` was missing. If you're not familiar with system includes, they're include directories which the compiler knows you do not own or control so it will not emit diagnostics about headers in those paths. This is super useful because it means any compiler warnings we enable for SFML code won't also apply to 3rd party code which we cannot fix like `doctest.h` or OpenGL headers. 

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/src/SFML/Window/CMakeLists.txt
+++ b/src/SFML/Window/CMakeLists.txt
@@ -247,7 +247,7 @@ if ((NOT BUILD_SHARED_LIBS) AND SFML_OS_MACOSX)
 endif()
 
 # Vulkan headers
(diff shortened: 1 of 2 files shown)


eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:
Thanks for this update! 🥳

PR #2040, opened 2022-03-14
Fixes compilation error on MacOS

roughbits01:

----

## Description

This change fixes the following compilation error on MacOS with Clang 13 (mainline clang installed through HomeBrew `brew install llvm`).
```
error: implicit conversion from 'size_t' (aka 'unsigned long') to 'CGFloat' (aka 'double') may lose precision [-Werror,-Wimplicit-int-float-conversion]
```

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Compile SFML with `-Werror,-Wimplicit-int-float-conversion`

### DIFF
--- a/src/SFML/Window/OSX/SFWindowController.mm
+++ b/src/SFML/Window/OSX/SFWindowController.mm
@@ -631,8 +631,7 @@
     NSDictionary* deviceDescription = [[m_window screen] deviceDescription];
     NSNumber* screenNumber = [deviceDescription valueForKey:@"NSScreenNumber"];
     CGDirectDisplayID screenID = static_cast<CGDirectDisplayID>([screenNumber intValue]);
-    CGFloat height = CGDisplayPixelsHigh(screenID);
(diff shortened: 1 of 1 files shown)


eXpl0it3r:

ChrisThrasher:
Did you also compile the tests and examples with that extra warning you added? If we're going to fix this warning, I'd prefer that we add it to CompilerWarnings.cmake and see what else gets caught when the CI pipeline runs.

ChrisThrasher (approved):
I switched to this same version of Clang and was able to recreate this error without enabling additional compiler warnings so we must already be using this warning. The tests and examples build fine too. Thanks for the contribution!

eXpl0it3r (approved):

eXpl0it3r:
Thanks for this fix! 🥳

PR #2043, opened 2022-03-15
Print absolute paths when file not found

ChrisThrasher:
## Description

Printing absolute paths is helpful when debugging why files won't load. By printing the whole path we're making it more clear to the user exactly what file is failing to load instead of simply repeating the relative path they already wrote in their code.

Here's some more info on [`std::filesystem::absolute`](https://en.cppreference.com/w/cpp/filesystem/absolute). If a path isn't already absolute, it creates an absolute path. This means our APIs still work whether users use relative or absolute paths. I don't believe a change like this would have been feasible before switching over to `std::filesystem`.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

For an easy example of this, build the master branch then try to run `./examples/tennis/tennis` from the CMake binary directory. This will fail with an error that looks like this:

```
AL lib: (EE) FillCPUCaps: Failed to open /proc/cpuinfo, cannot check for NEON support
Failed to open sound file "resources/ball.wav" (couldn't open stream)
```

After this PR, the error instead looks like this:

```
AL lib: (EE) FillCPUCaps: Failed to open /proc/cpuinfo, cannot check for NEON support
Failed to open sound file "/Users/thrasher/Projects/sfml/build/resources/ball.wav" (couldn't open stream)
```

The first output is a bit unclear because there is certainly a file called `ball.wav` in the `resources` directory which lives next to the source file where that code is written. However the second output shows the real truth that I'm actually attempting to load a file from the CMake binary direction which doesn't exist.

### DIFF
--- a/src/SFML/System/Utils.hpp
+++ b/src/SFML/System/Utils.hpp
@@ -29,17 +29,28 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <cctype>
+#include <filesystem>
+#include <sstream>
 #include <string>
 
 
 namespace sf
 {
-    [[nodiscard]] inline std::string toLower(std::string str)
-    {
-        for (char& c : str)
-            c = static_cast<char>(std::tolower(static_cast<unsigned char>(c)));
-        return str;
-    }
+[[nodiscard]] inline std::string toLower(std::string str)
+{
+    for (char& c : str)
+        c = static_cast<char>(std::tolower(static_cast<unsigned char>(c)));
+    return str;
+}
+
+[[nodiscard]] inline std::string formatDebugPathInfo(const std::filesystem::path& path)
+{
+    std::ostringstream stream;
+    stream << "    Provided path: " << path << '\n';
+    stream << "    Absolute path: " << std::filesystem::absolute(path);
+    return stream.str();
+}
+
 } // namespace sf
 
 #endif // SFML_UTILS_HPP
--- a/src/SFML/Audio/SoundFileFactory.cpp
+++ b/src/SFML/Audio/SoundFileFactory.cpp
@@ -36,6 +36,7 @@
 #include <SFML/System/FileInputStream.hpp>
 #include <SFML/System/MemoryInputStream.hpp>
 #include <SFML/System/Err.hpp>
+#include <SFML/System/Utils.hpp>
 #include <ostream>
 
 
@@ -74,7 +75,7 @@ std::unique_ptr<SoundFileReader> SoundFileFactory::createReaderFromFilename(cons
     // Wrap the input file into a file stream
     FileInputStream stream;
     if (!stream.open(filename)) {
-        err() << "Failed to open sound file " << filename << " (couldn't open stream)" << std::endl;
+        err() << "Failed to open sound file (couldn't open stream)\n" << formatDebugPathInfo(filename) << std::endl;
         return nullptr;
     }
 
@@ -92,7 +93,7 @@ std::unique_ptr<SoundFileReader> SoundFileFactory::createReaderFromFilename(cons
     }
 
     // No suitable reader found
-    err() << "Failed to open sound file " << filename << " (format not supported)" << std::endl;
+    err() << "Failed to open sound file (format not supported)\n" << formatDebugPathInfo(filename) << std::endl;
     return nullptr;
 }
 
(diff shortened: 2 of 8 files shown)


eXpl0it3r:
On the other side, printing the absolute path when you specified a relative path, can also be confusing and potentially misleading. Like why is it printing the absolute path when I explicitly specified the relative path? Or a confused user may think that they need to use an absolute path, because that's what the error message returned.

I'm okay with the change, just wanted to open a discussion from another point of view, to ensure we aren't making things worse on a broader spectrum.

ChrisThrasher:
> On the other side, printing the absolute path when you specified a relative path, can also be confusing and potentially misleading. Like why is it printing the absolute path when I explicitly specified the relative path? Or a confused user may think that they need to use an absolute path, because that's what the error message returned.

We could raise the abstraction level even further and create a helper function which given a `std::filesystem::path` will simply print the path if it's already absolute or if it's relative print both the relative and absolute paths. I avoided this mostly in the name of maintaining a simple implementation but something like this is certainly possible.

Bromeon:
Just brainstorming -- a third option might be the path _relative to the executable_. (Is this always the one the user specifies? What about different working directories?)

Path transformations can quickly get fiddly though, as soon as things like symlinks are involved. Paths are non-trivial...

If clarity is a concern, it's also possible to provide both user-provided and absolute path in the error message:
```
Error: Failed to open sound file (couldn't open stream).
   path:       /to/file.ogg
   absolute:   /user/absolute-path/to/file.ogg
```

ChrisThrasher:
> Just brainstorming -- a third option might be the path _relative to the executable_. (Is this always the one the user specifies? What about different working directories?)

I don't think there is a 3rd path to reason about. There is just the user-provided relative path (which is implicitly relative to where the executable is run, not where it is compiled) and the absolute path equivalent of that. If the user provides an absolute path then there is only one path to reason about.

> If clarity is a concern, it's also possible to provide both user-provided and absolute path in the error message:
> ```
> Error: Failed to open sound file (couldn't open stream).
>    path:       /to/file.ogg
>    absolute:   /user/absolute-path/to/file.ogg
> ```

This is easy to implement so I'd support doing it in this PR. I tried it out locally and it looks good.

> Edit: no matter what we choose, I'd actually appreciate extracting the "context data" (such as path, filename, error details, etc) from the message itself. With something like
> ```
> Error: Failed to open sound file C:/Users/Me Who Has a Long Name/Documents/SFML/Projects/XY/Audio/File.ogg (couldn't open stream).
> it's very easy to miss the "(couldn't open stream)" part.
> ```

Good point. Irrespective of this PR, it's worth taking a closer look at all uses of `sf::err()` to make sure the wording of the errors is as clear as possible.

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:
I quite like @Bromeon's suggestion with showing both. It also helps people understand where their file is expected to be and what relative path they've entered. So it might actually lead to fewer "can't load image" support requests 🙂

ChrisThrasher:
> I quite like @Bromeon's suggestion with showing both. It also helps people understand where their file is expected to be and what relative path they've entered. So it might actually lead to fewer "can't load image" support requests 🙂

I took a crack at providing more info. Now when an operation involving a path fails, the output looks like this:

```
Failed to open sound file (couldn't open stream)
    Provided path: "resources/ball.wav"
    Absolute path: "/Users/thrasher/Projects/sfml/build/resources/ball.wav"
```

I like how "Provided" and "Absolute" have the same number of characters. I wanted to make it clear that we're showing the user the path they themselves provided to the API whether not that path was already absolute.

src/SFML/System/Utils.hpp
```diff
@@ -29,17 +29,28 @@
...
-        return str;
-    }
+[[nodiscard]] inline std::string toLower(std::string str)
+{
+    for (char& c : str)
+        c = static_cast<char>(std::tolower(static_cast<unsigned char>(c)));
+    return str;
+}
+
+[[nodiscard]] inline std::string printDebugPathInfo(const std::filesystem::path& path)
```
kimci86, line 46: Having "print" in the name makes me expect this function to write to an output stream. Maybe it could be renamed to remove the "print" part.
An alternative would be to have a functor with a reference to a path as member data which actually prints to an `std::ostream`.

ChrisThrasher: Good point. I renamed it to `formatDebugPathInfo`.

eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:

eXpl0it3r:
Thanks for opening this idea and I hope we'll soon see the benefit from it 🙂

PR #2042, opened 2022-03-15
Remove redundant quotes when formatting filesystem paths

ChrisThrasher:
## Description

`operator<<` for `std::filesystem::path` already adds quotes around the path so we no longer have to keep manually adding them ourselves. 

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

PR #2044, opened 2022-03-15
Use `std::quoted`

ChrisThrasher:
## Description

While working on #2042 and #2043 I realized that there are a good number of instances of us manually adding quotes around something so I went about finding all those instances and replacing them with `std::quoted` which I think better expresses the intent of the code. This is my first time using `std::quoted` but I believe this is a valid use case for it. 

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/examples/sockets/TCP.cpp
+++ b/examples/sockets/TCP.cpp
@@ -3,6 +3,7 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Network.hpp>
+#include <iomanip>
 #include <iostream>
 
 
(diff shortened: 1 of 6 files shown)


eXpl0it3r:

eXpl0it3r (approved):

Bromeon (commented):
Technically, we would need
```cpp
#include <iomanip>
```
in each file, no?

ChrisThrasher:
> Technically, we would need
> 
> ```c++
> #include <iomanip>
> ```
> 
> in each file, no?

I don't think we technically need it as evident by CI passing but it's not a bad idea to add it anyways. Want me to do that?

ChrisThrasher:
> Technically, we would need
> 
> ```c++
> #include <iomanip>
> ```
> 
> in each file, no?

I don't think we technically need it as evident by CI passing but it's not a bad idea to add it anyways. Want me to do that?

eXpl0it3r:

eXpl0it3r:
I find relying on transitive includes isn't the best idea, unless the transitivity is actually guaranteed.

eXpl0it3r:

ChrisThrasher:
I added `#include <iomanip>` to all the files using `std::quoted`.

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:
Thanks! 🙂

Issue #2046, opened 2022-03-16
Problems with building SFML 3 (and 2.6) on Windows with Clang


nedoxff:
Thanks for raising your issue here! 🙂

Before you submit the issue however, we'd like you to consider the follow points.

* We like to use the issue tracker for confirmed issues.
* If you're stuck with SFML, please use [the forum](https://en.sfml-dev.org/forums/index.php#c3) to get help.

----

## Subject of the issue

While I have been constructing a minimal example for another issue (with `sf::String`s and wide strings), I noticed there were problems with building SFML 2.6 & 3.0 when using non-MSVC compilers (so ~~MinGW~~ and both Clang-like command-lines (`clang-cl.exe` & `clang++.exe`))

## Your environment

* Your OS / distro / window manager used: Windows 10, 64bit, build 21H2
![изображение](https://user-images.githubusercontent.com/65343244/158676400-e4ba12a4-c13d-49b5-b87d-013d10c15e4a.png)

* Your version of SFML (2.5.0, git master, etc): 2.6 (branch `2.6.x`, tag `3315456dc32e2da000145bb7f97083154806b59b`) and 3.0 (branch `master`, tag `5c9b571c70eeb0171affbce08b005a1d65765115`)
* Your compiler and compiler version used: Clang and MinGW (GNU 11.2.0, bundled with CLion)
Clang versions:
![изображение](https://user-images.githubusercontent.com/65343244/158677019-9760c6a4-f172-4a75-86a1-88b88463a22c.png)

* Special compiler flags used: none

## Steps to reproduce

Here is the ZIP archive I used for both text and build tests (the standard is set to C++17):  [SFMLTest.zip](https://github.com/SFML/SFML/files/8270902/SFMLTest.zip)
**MD5 Checksum:** `4c99ca98215ccc919181fd2dd8d24c79`

Here is just the `main.cpp` without any resources:
```cpp
#include <iostream>
#include <SFML/Graphics.hpp>
int main() {
	sf::RenderWindow window(sf::VideoMode(500, 100), "SFML Wide String Text Example");

	sf::Font font;
	if(!font.loadFromFile("C:\\Windows\\Fonts\\Arial.ttf"))
	{
		std::cerr << "uh oh! (failed loading font)" << std::endl;
		return 1;
	}

	sf::Text text;
	text.setCharacterSize(48);
	text.setString(L"test test тест");
	text.setFont(font);

	while(window.isOpen())
	{
		sf::Event event{};
[...]

PR #1914, opened 2021-12-19
Fix `-Wmissing-braces` warning under clang

vittorioromeo:
Fixes this warning under clang:

```
src/SFML/Window/Win32/WindowImplWin32.cpp:212:147: warning: suggest braces around initialization of subobject [-Wmissing-braces]
    DEV_BROADCAST_DEVICEINTERFACE deviceInterface = {sizeof(DEV_BROADCAST_DEVICEINTERFACE), DBT_DEVTYP_DEVICEINTERFACE, 0, GUID_DEVINTERFACE_HID, 0};
                                                                                                                                                  ^
                                                                                                                                                  {}
```

PR #2048, opened 2022-03-16
Fix `-Wmissing-braces` warning under clang (for SFML 2.6)

NedoProgrammer:

----

## Description

Cherry-picked [this commit](https://github.com/SFML/SFML/commit/c78ddc1f377667c70fd1d440dbfd174a3a420817) because SFML 2.6 had the same problem with building with Clang..

This PR is related to the issue #2046 and PR #1914.

## Tasks

This PR only effects Windows.

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Please check the original commit and the #2046 issue.

### DIFF
--- a/src/SFML/Window/Win32/WindowImplWin32.cpp
+++ b/src/SFML/Window/Win32/WindowImplWin32.cpp
@@ -220,7 +220,7 @@ m_cursorGrabbed   (m_fullscreen)
     m_handle = CreateWindowW(className, title.toWideString().c_str(), win32Style, left, top, width, height, NULL, NULL, GetModuleHandle(NULL), this);
 
     // Register to receive device interface change notifications (used for joystick connection handling)
-    DEV_BROADCAST_DEVICEINTERFACE deviceInterface = {sizeof(DEV_BROADCAST_DEVICEINTERFACE), DBT_DEVTYP_DEVICEINTERFACE, 0, GUID_DEVINTERFACE_HID, 0};
+    DEV_BROADCAST_DEVICEINTERFACE deviceInterface = {sizeof(DEV_BROADCAST_DEVICEINTERFACE), DBT_DEVTYP_DEVICEINTERFACE, 0, GUID_DEVINTERFACE_HID, {0}};
     RegisterDeviceNotification(m_handle, &deviceInterface, DEVICE_NOTIFY_WINDOW_HANDLE);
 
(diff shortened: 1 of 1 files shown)


ChrisThrasher (approved):

eXpl0it3r:

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -220,7 +220,7 @@ m_cursorGrabbed   (m_fullscreen)
     m_handle = CreateWindowW(className, title.toWideString().c_str(), win32Style, left, top, width, height, NULL, NULL, GetModuleHandle(NULL), this);
 
     // Register to receive device interface change notifications (used for joystick connection handling)
-    DEV_BROADCAST_DEVICEINTERFACE deviceInterface = {sizeof(DEV_BROADCAST_DEVICEINTERFACE), DBT_DEVTYP_DEVICEINTERFACE, 0, GUID_DEVINTERFACE_HID, 0};
+    DEV_BROADCAST_DEVICEINTERFACE deviceInterface = {sizeof(DEV_BROADCAST_DEVICEINTERFACE), DBT_DEVTYP_DEVICEINTERFACE, 0, GUID_DEVINTERFACE_HID, {0}};
```
eXpl0it3r, line 223: Is this C++03 compliant?

For some reason I'm having a hard time to understand a) why the braces are needed and b) what it's actually constructing here... 🤔

NedoProgrammer: I think braces are required because the struct that is constructed requires not a char, but an array with a single char in it
```cpp
char  dbcc_name[1];
```
So you can't really initialize an array with a number

And the C++03 standard says that any aggregate type can be brace-initialized, so I would guess it works out

I will test it tomorrow with the 03 standard and share the results 😄

ChrisThrasher: https://godbolt.org/z/rrsca7GMY

Even ancient versions of GCC are fine with the extra braces around that `0` so I think we're good and not accidentally using modern C++ in SFML 2.

eXpl0it3r (approved):

eXpl0it3r:
Thanks for this fix! 🙂

PR #2050, opened 2022-03-24
Require all modules are built before building tests

ChrisThrasher:
## Description

Currently the tests must recreate a lot of the configuration complexity that the core library has. If you disable a specific module, it is expected that the tests for that module don't get built or run. This approach would be fine if a given module's tests only depended on that one module (in addition to the test framework). This could in fact be true if we were willing to add `operator<<` to each class we test. It would make our test code significantly simpler and avoid the problems this PR seeks to solve. However because we refuse to do that, we're forced to create a 2nd class library only used for testing called `sfml-test-main` which contains definitions for those `operator<<` functions that let us see the contents of a type when a test assertion fails. These operators have the same inter-module dependencies as the core library. If you get the `operator<<` for `sf::Transform` then you also need the `operator<<` for `sf::Vector2<T>` which is used in `sf::Transform`'s interface. That means `sfml-test-main` also needs to recreate the optional module logic such that when you disable the Graphics module for instance, the corresponding Graphics test utilities don't get built.

Maintaining all this optional module build logic in 2-3 places is difficult. Because these code paths get tested so seldom, they're prone to bugs. Our best solution is to add `operator<<` to the core library thus allowing us to totally sidestep this problem. I think many users would appreciate this as well. Our second best solution is to simply require that all modules be built if anyone wants to build the tests. It's reasonable to assume that a potential contributor who wants to run the tests will also be able and willing to enable all modules.  

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Disable any module and observe how the tests are not configured, compiled, or ran.


eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

CMakeLists.txt
```diff
@@ -282,7 +282,7 @@ endif()
 if(SFML_BUILD_TEST_SUITE)
     if (SFML_OS_IOS)
         message( WARNING "Unit testing not supported on iOS")
-    else()
+    elseif(SFML_BUILD_WINDOW AND SFML_BUILD_GRAPHICS AND SFML_BUILD_NETWORK AND SFML_BUILD_AUDIO)
```
eXpl0it3r, line 285: I think we should output at least a warning message, that tests aren't being executed.
Otherwise someone might enable tests and just build some modules and not realize tests aren't even being executed.

eXpl0it3r (changes requested):
Small usability improvement

CMakeLists.txt
ChrisThrasher: Good idea. I just added a warning for that.

eXpl0it3r (approved):

eXpl0it3r:
Thanks 🎉

PR #2055, opened 2022-03-30
Use `sf::Vector2<T>` for numeric parameter pairs

ChrisThrasher:
## Description

This PR further expands on the work from #1902, #1942, and #1948. It originated from a conversation in #2027 which found more instances of pairs of numeric parameters. I thought it'd be good to go ahead and fix this but quickly found just how many more numeric parameter pairs we had left in SFML. I tried to find as many instances of this pattern as possible but if any still remain, I'd love to fix them I this PR.

I know this PR is massive. While finding and changing all these interfaces, I didn't try to break this apart into multiple commits although it's not unreasonable to do so. If this is simply an unmanageably large PR to review then let me know and I'll try to break it apart.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

### DIFF
--- a/include/SFML/Graphics/RenderTexture.hpp
+++ b/include/SFML/Graphics/RenderTexture.hpp
@@ -78,14 +78,13 @@ public:
     /// requires a depth or stencil buffer. Otherwise it is unnecessary, and
     /// you should leave this parameter at its default value.
     ///
-    /// \param width    Width of the render-texture
-    /// \param height   Height of the render-texture
+    /// \param size     Width and height of the render-texture
     /// \param settings Additional settings for the underlying OpenGL texture and context
     ///
     /// \return True if creation has been successful
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool create(unsigned int width, unsigned int height, const ContextSettings& settings = ContextSettings());
+    [[nodiscard]] bool create(const Vector2u& size, const ContextSettings& settings = ContextSettings());
 
     ////////////////////////////////////////////////////////////
     /// \brief Get the maximum anti-aliasing level supported by the system
--- a/include/SFML/Graphics/Texture.hpp
+++ b/include/SFML/Graphics/Texture.hpp
@@ -91,13 +91,12 @@ public:
     ///
     /// If this function fails, the texture is left unchanged.
     ///
-    /// \param width  Width of the texture
-    /// \param height Height of the texture
+    /// \param size Width and height of the texture
     ///
     /// \return True if creation was successful
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool create(unsigned int width, unsigned int height);
+    [[nodiscard]] bool create(const Vector2u& size);
 
     ////////////////////////////////////////////////////////////
     /// \brief Load the texture from a file on disk
@@ -272,13 +271,11 @@ public:
     /// texture was not previously created.
     ///
     /// \param pixels Array of pixels to copy to the texture
-    /// \param width  Width of the pixel region contained in \a pixels
-    /// \param height Height of the pixel region contained in \a pixels
-    /// \param x      X offset in the texture where to copy the source pixels
-    /// \param y      Y offset in the texture where to copy the source pixels
+    /// \param size   Width and height of the pixel region contained in \a pixels
+    /// \param dest   Coordinates of the destination position
     ///
     ////////////////////////////////////////////////////////////
-    void update(const Uint8* pixels, unsigned int width, unsigned int height, unsigned int x, unsigned int y);
+    void update(const Uint8* pixels, const Vector2u& size, const Vector2u& dest);
 
     ////////////////////////////////////////////////////////////
     /// \brief Update a part of this texture from another texture
@@ -304,18 +301,17 @@ public:
     /// \brief Update a part of this texture from another texture
     ///
     /// No additional check is performed on the size of the texture,
-    /// passing an invalid combination of texture size and offset
+    /// passing an invalid combination of texture size and destination
     /// will lead to an undefined behavior.
     ///
     /// This function does nothing if either texture was not
     /// previously created.
     ///
     /// \param texture Source texture to copy to this texture
-    /// \param x       X offset in this texture where to copy the source texture
-    /// \param y       Y offset in this texture where to copy the source texture
+    /// \param dest    Coordinates of the destination position
     ///
     ////////////////////////////////////////////////////////////
-    void update(const Texture& texture, unsigned int x, unsigned int y);
+    void update(const Texture& texture, const Vector2u& dest);
 
     ////////////////////////////////////////////////////////////
     /// \brief Update the texture from an image
@@ -341,18 +337,17 @@ public:
     /// \brief Update a part of the texture from an image
     ///
     /// No additional check is performed on the size of the image,
-    /// passing an invalid combination of image size and offset
+    /// passing an invalid combination of image size and destination
     /// will lead to an undefined behavior.
     ///
     /// This function does nothing if the texture was not
     /// previously created.
     ///
     /// \param image Image to copy to the texture
-    /// \param x     X offset in the texture where to copy the source image
-    /// \param y     Y offset in the texture where to copy the source image
+    /// \param dest  Coordinates of the destination position
     ///
     ////////////////////////////////////////////////////////////
-    void update(const Image& image, unsigned int x, unsigned int y);
+    void update(const Image& image, const Vector2u& dest);
 
     ////////////////////////////////////////////////////////////
     /// \brief Update the texture from the contents of a window
@@ -378,18 +373,17 @@ public:
     /// \brief Update a part of the texture from the contents of a window
     ///
     /// No additional check is performed on the size of the window,
-    /// passing an invalid combination of window size and offset
+    /// passing an invalid combination of window size and destination
     /// will lead to an undefined behavior.
     ///
     /// This function does nothing if either the texture or the window
     /// was not previously created.
     ///
     /// \param window Window to copy to the texture
-    /// \param x      X offset in the texture where to copy the source window
-    /// \param y      Y offset in the texture where to copy the source window
+    /// \param dest   Coordinates of the destination position
     ///
     ////////////////////////////////////////////////////////////
-    void update(const Window& window, unsigned int x, unsigned int y);
+    void update(const Window& window, const Vector2u& dest);
 
     ////////////////////////////////////////////////////////////
     /// \brief Enable or disable the smooth filter
--- a/include/SFML/Window/Context.hpp
+++ b/include/SFML/Window/Context.hpp
@@ -30,6 +30,7 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Window/Export.hpp>
 #include <SFML/Window/GlResource.hpp>
+#include <SFML/System/Vector2.hpp>
 #include <memory>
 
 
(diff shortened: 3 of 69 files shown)


eXpl0it3r:

include/SFML/Graphics/RenderTexture.hpp
```diff
@@ -78,14 +78,13 @@ class SFML_GRAPHICS_API RenderTexture : public RenderTarget
...
-    /// \param width    Width of the render-texture
-    /// \param height   Height of the render-texture
+    /// \param size     Width and height of the render-texture
     /// \param settings Additional settings for the underlying OpenGL texture and context
     ///
     /// \return True if creation has been successful
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool create(unsigned int width, unsigned int height, const ContextSettings& settings = ContextSettings());
+    [[nodiscard]] bool create(const sf::Vector2u& size, const ContextSettings& settings = ContextSettings());
```
JonnyPtn, line 87: I think sf namespace is unnecessary here?

include/SFML/Window/Context.hpp
```diff
@@ -30,6 +30,7 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Window/Export.hpp>
 #include <SFML/Window/GlResource.hpp>
+#include <SFML/System/Vector2.hpp>
```
JonnyPtn, line 33: Can we forward declare instead?

JonnyPtn (changes requested):

include/SFML/Graphics/Texture.hpp
```diff
@@ -272,13 +271,11 @@ class SFML_GRAPHICS_API Texture : GlResource
     /// texture was not previously created.
     ///
     /// \param pixels Array of pixels to copy to the texture
-    /// \param width  Width of the pixel region contained in \a pixels
-    /// \param height Height of the pixel region contained in \a pixels
-    /// \param x      X offset in the texture where to copy the source pixels
-    /// \param y      Y offset in the texture where to copy the source pixels
+    /// \param size   Width and height of the pixel region contained in \a pixels
+    /// \param offset Offset in the texture where to copy the source pixels
```
Bromeon, line 275: > Offset in the texture where to copy the source pixels

This sounds confusing (I know it was before). Is this source or destination?
Maybe we should also rename parameters for clarity.

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -226,7 +226,7 @@ m_cursorGrabbed   (m_fullscreen)
 
     // By default, the OS limits the size of the window the the desktop size,
     // we have to resize it after creation to apply the real size
-    setSize(Vector2u(mode.width, mode.height));
+    setSize(Vector2u(mode.size));
```
kimci86, line 229:
```suggestion
    setSize(mode.size);
```

include/SFML/Window/VideoMode.hpp
```diff
@@ -53,12 +54,11 @@ class SFML_WINDOW_API VideoMode
...
     /// \brief Construct the video mode with its attributes
     ///
-    /// \param modeWidth        Width in pixels
-    /// \param modeHeight       Height in pixels
+    /// \param modeSize         Width and height in pixels
     /// \param modeBitsPerPixel Pixel depths in bits per pixel
     ///
     ////////////////////////////////////////////////////////////
-    VideoMode(unsigned int modeWidth, unsigned int modeHeight, unsigned int modeBitsPerPixel = 32);
+    VideoMode(const Vector2u& modeSize, unsigned int modeBitsPerPixel = 32);
```
kimci86, line 61: Note that this is now allowed:
```c++
sf::VideoMode mode = sf::Vector2u(800, 600);
```
Should this constructor be `explicit`?

kimci86 (commented):
I have some comments and suggestions below.
In addition, `View::move(float, float)` method could be removed.

include/SFML/Window/VideoMode.hpp
Bromeon: Good catch -- this also means that we will have silent code breaking from SFML 2 -> 3 (code still compiles, but causes nasty bugs at runtime).

include/SFML/Window/Context.hpp
ChrisThrasher: Is that possible given how `Vector2` is a class template and `Vector2u` a type alias? I just tried this locally and couldn't get it to work.

src/SFML/Window/Win32/WindowImplWin32.cpp
ChrisThrasher: Thanks!

ChrisThrasher:
> In addition, `View::move(float, float)` method could be removed.

@kimci86 Good catch. The definition for that was already deleted but the declaration stuck around for some reason.

eXpl0it3r:

include/SFML/Window/VideoMode.hpp
ChrisThrasher: Okay! I just fixed it.

eXpl0it3r:

ChrisThrasher:
```cpp
    struct SizeEvent
    {
        unsigned int width;  //!< New width, in pixels
        unsigned int height; //!< New height, in pixels
    };
```
Should `sf::Event::SizeEvent` contain a `Vector2u` as well? I was unsure if this was a good idea so I didn't change it.

eXpl0it3r:

Bromeon:
> Should `sf::Event::SizeEvent` contain a `Vector2u` as well? I was unsure if this was a good idea so I didn't change it.

I remember this was originally avoided because in C++98, unions were not allowed to store types with constructors (see e.g. [here](https://en.sfml-dev.org/forums/index.php?topic=18399). I'm not fully up-to-date; this might have been relaxed in recent standards (as indicated [here](https://stackoverflow.com/q/10693913), for example). Someone would need to confirm this.

If yes, this change can also be applied to the other events which hold a position (mouse, touch).

ChrisThrasher:
> I remember this was originally avoided because in C++98, unions were not allowed to store types with constructors (see e.g. [here](https://en.sfml-dev.org/forums/index.php?topic=18399)). I'm not fully up-to-date; this might have been relaxed in recent standards (as indicated [here](https://stackoverflow.com/q/10693913), for example). Someone would need to confirm this.

AppleClang didn't like it when I tried to do this. Maybe I'm doing something wrong or maybe it really isn't possible, sadly.

```
/Users/thrasher/Projects/sfml/src/SFML/Window/WindowImpl.cpp:196:19: error: call to implicitly-deleted default constructor of 'sf::Event'
            Event event;
```

Bromeon:
> One downside is the the API starts to look weird when using these events. `event.size.width` becomes `event.size.size.x` which frankly looks bad. I'm on the fence about whether this is a worthwhile change to make.

I think the real problem is the naming of `SizeEvent` and `size` (the event). Should probably be `Resize`; then `event.resize.size.x` or `event.resized.size.x` would also look less problematic.

> It can certainly wait for a later PR since it's a bit outside the scope of this PR which aims to only change APIs using x/y or width/height pairs.

Yes, I'd say let's discuss the `sf::Event` API separately.

include/SFML/Graphics/Texture.hpp
ChrisThrasher: Reading through the source, `offset` appears to representation a destination, not a source.

eXpl0it3r:

eXpl0it3r:
> One possibility would also be to look into std::variant -- but the visiting API is quite unfortunate in C++, even [the tricks](https://stackoverflow.com/a/62344801) attempting to make it more ergonomic aren't ideal.

Just as a side-note - and I agree with the separate discussion for the Event API - @lethal-guitar gave a talk on using `std::variant` with SFML's event system at Meeting C++ 2018: https://www.youtube.com/watch?v=CELWr9roNno

eXpl0it3r:

include/SFML/Window/Context.hpp
eXpl0it3r: Yeah, you can't since it's a template

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

Bromeon:
It's very hard to keep track of PRs which last for a month or more and change 60+ files, especially when small changes are continuously added by force-pushing/overwriting 1 commit. As a reviewer, it would require me to repeatedly re-review everything and memorize the changes from a week ago.

I would suggest we keep PRs smaller and merge them quicker, aka "don't let perfect be the enemy of good" -- we can iteratively improve the codebase without getting everything done at once. In addition to easier reviewing, this would be more aligned with the continous integration philosophy and would also avoid merge conflicts as other changes get merged in the meantime.

Keeping separate commits (until final squash) is an option too, but it only addresses half of the issues.

eXpl0it3r:

include/SFML/Graphics/Texture.hpp
eXpl0it3r: Should it be renamed to destination everywhere?

ChrisThrasher: I like that idea

eXpl0it3r (approved):
Looks good to me!

Thanks for the massive effort 🙂 

Before we merge this, we should however make sure, that it really is an intentional sematic change from (width, height) to (size.x, size.y).

I've been using the vector class like this for a long time and thus am totally fine with the change. What do others think about it?

Bonkt:
I was recently very confused as i rebuilt my SFML and had used pairs instead of sf::VectorT for some of my initialization, and couldn't find the overloaded methods in SFML GH repo, nor anything in the change log.
What made it worse was that the documentation (2.5.1) is still stating support for numeric pairs. 

Although i will admit i might have misunderstood what version the "master" Branch here on GH actually is. Is it 2.5.1 or some newer? Maybe i should have cloned an older  version? Sorry - i'm new to open source GH.

Also - sorry if this is an inopportune comment in this forum.

Bromeon:
> Although i will admit i might have misunderstood what version the "master" Branch here on GH actually is. Is it 2.5.1 or some newer? Maybe i should have cloned an older version? Sorry - i'm new to open source GH.

A bit off-topic, but we actively maintain 2 branches on GitHub:
* `master` -- follows SFML 3 development
* `2.6.x` -- continuation of SFML 2 for some time

So yes, `master` branch will not be in line with the documentation you find on the website. You'll need to generate the docs for SFML 3 yourself with Doxygen, at the moment.

ChrisThrasher:
Rebasing on master revealed some compilation errors with DRM code, so I made the same `width, height` -> `size` changes and everything appears to be working now.

eXpl0it3r:

src/SFML/Window/DRM/DRMContext.cpp
```diff
@@ -204,8 +204,7 @@ m_config     (NULL),
 m_currentBO  (NULL),
 m_nextBO     (NULL),
 m_gbmSurface (NULL),
-m_width      (0),
-m_height     (0),
+m_size       (0, 0),
```
ChrisThrasher, line 207: Do we care about explicitly initializing vectors to zero when the default constructor already does this?

ChrisThrasher:
I improved some parameter names and rebased on master since there were minor conflicts to fix.

eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:
Thank you for this massive PR! ❤️🙂 

Lots of breaking changes to document 😅

PR #2063, opened 2022-04-08
Add `sf::LineShape`

ChrisThrasher:
## Description

`sf::LineShape` models a line segment defined by two points. The first point specifies the position of the line segment and the segment is drawn to the second point. The goal is to give users the simplest possible API for something they're already doing, drawing lines.

[Here](https://github.com/ChrisThrasher/fourier-draw/commit/14498cce2755ca25e8359f2d313da62fa6be70c4) is an example of me using `sf::LineShape` to replace `sf::RectangleShape` in a project of mine. 

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

### DIFF
--- a/src/SFML/Graphics/LineShape.cpp
+++ b/src/SFML/Graphics/LineShape.cpp
@@ -1,0 +1,108 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2022 Laurent Gomila (laurent@sfml-dev.org)
+//
+// This software is provided 'as-is', without any express or implied warranty.
+// In no event will the authors be held liable for any damages arising from the use of this software.
+//
+// Permission is granted to anyone to use this software for any purpose,
+// including commercial applications, and to alter it and redistribute it freely,
+// subject to the following restrictions:
+//
+// 1. The origin of this software must not be misrepresented;
+//    you must not claim that you wrote the original software.
+//    If you use this software in a product, an acknowledgment
+//    in the product documentation would be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such,
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#include <SFML/Graphics/LineShape.hpp>
+
+
+namespace sf
+{
+////////////////////////////////////////////////////////////
+LineShape::LineShape(const Vector2f& beginning, const Vector2f& end, float thickness) :
+m_direction(end - beginning),
+m_thickness(thickness)
+{
+    setPosition(beginning);
+    update();
+}
+
+
+////////////////////////////////////////////////////////////
+void LineShape::setThickness(float thickness)
+{
+    m_thickness = thickness;
+    update();
+}
+
+
+////////////////////////////////////////////////////////////
+void LineShape::setEndPoint(const Vector2f& point)
+{
+    m_direction = point - getPosition();
+    update();
+}
+
+
+////////////////////////////////////////////////////////////
+float LineShape::getThickness() const
+{
+    return m_thickness;
+}
+
+
+////////////////////////////////////////////////////////////
+Vector2f LineShape::getEndPoint() const
+{
+    return getPosition() + m_direction;
+}
+
+
+////////////////////////////////////////////////////////////
+float LineShape::getLength() const
+{
+    return m_direction.length();
+}
+
+
+////////////////////////////////////////////////////////////
+std::size_t LineShape::getPointCount() const
+{
+    return 4;
+}
+
+
+////////////////////////////////////////////////////////////
+Vector2f LineShape::getPoint(const std::size_t index) const
+{
+    auto offset = Vector2f();
+    if (m_direction != Vector2f(0, 0))
+        offset = (m_thickness / 2.f) * m_direction.normalized().perpendicular();
+
+    switch (index)
+    {
+        default:
+        case 0:
+            return offset;
+        case 1:
+            return m_direction + offset;
+        case 2:
+            return m_direction - offset;
+        case 3:
+            return -offset;
+    }
+}
+
+} // namespace sf
--- a/include/SFML/Graphics.hpp
+++ b/include/SFML/Graphics.hpp
@@ -37,6 +37,7 @@
 #include <SFML/Graphics/Font.hpp>
 #include <SFML/Graphics/Glyph.hpp>
 #include <SFML/Graphics/Image.hpp>
+#include <SFML/Graphics/LineShape.hpp>
 #include <SFML/Graphics/PrimitiveType.hpp>
 #include <SFML/Graphics/Rect.hpp>
 #include <SFML/Graphics/RectangleShape.hpp>
--- a/include/SFML/Graphics/LineShape.hpp
+++ b/include/SFML/Graphics/LineShape.hpp
@@ -1,0 +1,160 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2022 Laurent Gomila (laurent@sfml-dev.org)
+//
+// This software is provided 'as-is', without any express or implied warranty.
+// In no event will the authors be held liable for any damages arising from the use of this software.
+//
+// Permission is granted to anyone to use this software for any purpose,
+// including commercial applications, and to alter it and redistribute it freely,
+// subject to the following restrictions:
+//
+// 1. The origin of this software must not be misrepresented;
+//    you must not claim that you wrote the original software.
+//    If you use this software in a product, an acknowledgment
+//    in the product documentation would be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such,
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+#ifndef SFML_LINESHAPE_HPP
+#define SFML_LINESHAPE_HPP
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#include <SFML/Graphics/Export.hpp>
+
+#include <SFML/Graphics/Shape.hpp>
+
+
+namespace sf
+{
+////////////////////////////////////////////////////////////
+/// \brief Specialized shape representing a line
+///
+////////////////////////////////////////////////////////////
+class SFML_GRAPHICS_API LineShape : public Shape
+{
+public:
+    ////////////////////////////////////////////////////////////
+    /// \brief Default constructor
+    ///
+    /// \param beginning Beginning of line
+    /// \param end       End of line
+    /// \param thickness Thickness of line
+    ///
+    ////////////////////////////////////////////////////////////
+    LineShape(const Vector2f& beginning, const Vector2f& end, float thickness);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Set the thickness of the line
+    ///
+    /// \param size New thickness of the line
+    ///
+    /// \see getThickness
+    ///
+    ////////////////////////////////////////////////////////////
+    void setThickness(float thickness);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Set the end point of the line
+    ///
+    /// \param size New end point of the line
+    ///
+    /// \see setEndPoint
+    ///
+    ////////////////////////////////////////////////////////////
+    void setEndPoint(const Vector2f& point);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the thickness of the line
+    ///
+    /// \return Thickness of the line
+    ///
+    /// \see setThickness
+    ///
+    ////////////////////////////////////////////////////////////
+    float getThickness() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the end point of the line
+    ///
+    /// \return End point of the line
+    ///
+    /// \see setEndPoint
+    ///
+    ////////////////////////////////////////////////////////////
+    Vector2f getEndPoint() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the length of the line
+    ///
+    /// \return Length of the line
+    ///
+    ////////////////////////////////////////////////////////////
+    float getLength() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the number of points defining the shape
+    ///
+    /// \return Number of points of the shape. For line
+    ///         shapes, this number is always 4.
+    ///
+    ////////////////////////////////////////////////////////////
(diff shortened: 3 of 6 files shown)


src/SFML/Graphics/LineShape.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+
+namespace sf
+{
+////////////////////////////////////////////////////////////
+LineShape::LineShape(const Vector2f& point1, const Vector2f& point2)
+    : m_direction(point2 - point1)
+    , m_thickness(2.f)
+{
+    assert(point1 != point2);
```
ChrisThrasher, line 39: I want to remove this assertion but it will require an additional condition in `getPoint` to ensure we're not calling `m_direction.normalize()` when `m_direction` has a length of zero which is an undefined operation. That may or may not have an acceptable performance penalty. I believe `getPoint` is called particularly often when drawing the shape so I wanted to keep it as efficient as possible.

src/SFML/Graphics/LineShape.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <SFML/Graphics/LineShape.hpp>
+#include <cassert>
+
+
+namespace sf
+{
+////////////////////////////////////////////////////////////
+LineShape::LineShape(const Vector2f& point1, const Vector2f& point2)
+    : m_direction(point2 - point1)
+    , m_thickness(2.f)
```
ChrisThrasher, line 37: Instead of a default thickness maybe we require a user-specified thickness in the constructor? Or default to zero thickness?

JonnyPtn: I think a thickness should be required as that’s the fundamental reason for this class

src/SFML/Graphics/LineShape.cpp
JonnyPtn: Is the call to normalise necessary? Can we not cache the points? Even with the assert the crash would still need to be prevented

JonnyPtn:
I personally don’t see much value in adding this, there are already great libraries built on top of SFML which offer this (or it’s relatively simple to roll your own) and although it seems simple on the surface, a line shape like this can actually get quite complex use cases.

src/SFML/Graphics/LineShape.cpp
ChrisThrasher: Right, no good reason to keep that default value. Easy fix.

src/SFML/Graphics/LineShape.cpp
ChrisThrasher: > Is the call to normalise necessary?

I believe so.

> Can we not cache the points? 

This should be possible.

> Even with the assert the crash would still need to be prevented

I added a condition that checks if the direction vector has zero length before trying to normalize it so that we can remove that assertion and allow a `LineShape` to have zero length.

Bromeon: I honestly think the impact of assertions is overblown -- even more so because they can be disabled. But even if they couldn't, good diagnostics are often worth more than saving a tiny bit of performance in a (often non-critical) branch.

So in my opinion, please add assertions, if they help make code clearer and catch potential bugs :)

src/SFML/Graphics/LineShape.cpp
```diff
@@ -0,0 +1,95 @@
...
+////////////////////////////////////////////////////////////
+size_t LineShape::getPointCount() const { return 4; }
+
+
+////////////////////////////////////////////////////////////
+Vector2f LineShape::getPoint(size_t index) const
+{
+    auto offset = Vector2f();
+    if (m_direction != Vector2f(0, 0))
+        offset = (m_thickness / 2.f) * m_direction.normalized().perpendicular();
```
Bromeon, line 80: Instead of caching each point (as discussed in the other comment), it's also possible to just cache `offset`. 
Then, we:
* don't have to check for zero vector
* can do a simple vector addition in `getPoint()`

ChrisThrasher: I noticed that other shapes don't do any caching. Is this something we want to start doing with all shapes?

If we do cache offset, the only time it would get invalidated is when `setThickness` or `setEndPoint` are called, right?

Bromeon:
> although it seems simple on the surface, a line shape like this can actually get quite complex use cases.

Isn't that rather a reason _for_ adding a line shape to SFML?

In the past, it was often recommended that `sf::RectangleShape` can be used for lines. And while technically true (at least for non-rounded ends), it adds a lot of overhead for someone who simply wants to draw a few debug graphics.

---

Regarding implementation, I like the current one, as it is fairly simple. One possible change would be to cache the `offset` vector perpendicular to the direction (see [comment](https://github.com/SFML/SFML/pull/2063#pullrequestreview-937027255)).

src/SFML/Graphics/LineShape.cpp
ChrisThrasher: I'm working through adding caching of the offset but I don't see how we get around checking `m_direction` for zero-length. At some point the offset has to be calculated and it'll change depending on whether `m_direction` has length.

Bromeon: I'm also fine with not caching things, and only restart this discussion once we have evidence that performance is indeed an issue. Since these shapes tend to be used for debug graphics, it's well possible that an uncached implementation is completely fine. And it's less complex (regarding invalidation, etc), meaning less error-prone.

> I'm working through adding caching of the offset but I don't see how we get around checking m_direction for zero-length.

True, what I meant is that we don't have to check it each frame, only when direction changes.

ChrisThrasher: > I'm also fine with not caching things, and only restart this discussion once we have evidence that performance is indeed an issue. 

This was my thought as well. If other shapes can get away without caching, then I think it's safe to avoid caching in `LineShape`.

JonnyPtn:
> Isn't that rather a reason for adding a line shape to SFML?

Sorry I don't quite understand how? It would mean we have to implement or at least consider the more complex use-cases as well as dealing with further feature requests (as you say). It would also like make the implementation/API much less simple than it currently is

>  someone who simply wants to draw a few debug graphics.

Yeah this use-case has been mentioned a few times in this discussion but I struggle to get on board with it. Debug lines are easily done with existing primitives, with this new class only offering the benefit of thickness, which is not something I've ever desired with my debug tools.

Based on my experience using lines with thickness (for end products, not debugging) there are a couple of things I've needed in my own implementations that I would question for this one:
- How should textures be applied? With the current implementation I believe the texture would not rotate with the line, which is often what I would want/expect
- Will we offer any control over the orientation of the ends of the line? A common use-case given for this is line graphs, but with any thickness larger than a couple of pixels the joins between the lines won't look good. In my own implementations i've made the ends of the line vertical for cases like this, but this is quite case-specific

ChrisThrasher:
> with this new class only offering the benefit of thickness

We can't dismiss how this class provides much improved semantics as well. It lets you reason about lines as lines as opposed to stretched and rotated rectangles. Being able to set the end point is a much more friendly API.

> How should textures be applied? With the current implementation I believe the texture would not rotate with the line, which is often what I would want/expect

Currently line segments can use any two start and end points while the rotation remains zero. An alternative solution is to dynamically update the rotation to match the direction to the end point. Then we just store a length and thickness internally. This approach might be friendlier to applying textures. 

> Will we offer any control over the orientation of the ends of the line? A common use-case given for this is line graphs, but with any thickness larger than a couple of pixels the joins between the lines won't look good. 

This is why we also ought to consider `sf::CurveShape` for drawing curves comprised of many line segments.

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:
While I can understand most arguments against a `sf::LineShape`, I still think that it should be implemented.

The fact that just about every SFML extension library has added such a line shape in one way or another and we have a multiple implementations in the community wiki is a strong indication that this is something SFML should provide out of the box.

- [Thor's Line Function](https://bromeon.ch/libraries/thor/documentation/v2.1/namespacethor_1_1_shapes.html#aaf7a2812214d46e49746d9b07e79762a)
- [Thor's Arrow Class](https://bromeon.ch/libraries/thor/documentation/v2.1/classthor_1_1_arrow.html)
- [SelbaWard's Line](https://github.com/Hapaxia/SelbaWard/wiki/Line)
- [SelbaWard's Spline](https://github.com/Hapaxia/SelbaWard/wiki/Spline)
- [Wiki: Drawable Line Shape](https://github.com/SFML/SFML/wiki/Source:-Drawable-Line-Shape)
- [Wiki: Line with thickness](https://github.com/SFML/SFML/wiki/Source%3A-line-with-thickness)
- [Wiki: Line Segment with thickness](https://github.com/SFML/SFML/wiki/Source%3A-Line-segment-with-thickness)
- [Wiki: Line Shape](https://en.sfml-dev.org/forums/index.php?topic=12825.0)

Additionally, it's an often requested feature, which is also a good measure to show what people want - interestingly enough, each of these requests, is immediately followed by a link to one of the implementations above.

- [sf::Line](https://en.sfml-dev.org/forums/index.php?topic=20176.0)
- [sf::LineShape](https://en.sfml-dev.org/forums/index.php?topic=18893.0)
- [A simple sf::Line](https://en.sfml-dev.org/forums/index.php?topic=12761.0)
- [sf::LineShape and sf::EllipseShape](https://en.sfml-dev.org/forums/index.php?topic=26259.0)
- [Drawing bold lines](https://en.sfml-dev.org/forums/index.php?topic=9902.0)
- *Many more on IRC/Discord*

With that, I believe it's quite undeniable that there is demand for such a functionality.

----

Now for some of the arguments...

> It's well possible that there may be further feature requests though, e.g. end style (rectangular, pointy, rounded, arrow), line style (constant, dotted, dashed), etc.
> ~ *Bromeon*

> It would mean we have to implement or at least consider the more complex use-cases as well as dealing with further feature requests (as you say). It would also like make the implementation/API much less simple than it currently is
> ~ *Jonny*

There are many potential more complex use cases for other shapes as well (dotted outlines, rounded rectangles, different colored for each side outline, etc.). Just because there's a potential for additional feature requests, doesn't mean that we have to follow through with the initial requests, nor that they will actually happen. It's good to discuss some of those points for the initial implementation and define what problem the sf::LineShape solves, but it's not a good argument against the feature overall.

On the other hand, why are things like an arrow class or a rounded line or a spline not something SFML should have from the get go?

> What worries me about lines with thickness, is that they are just rotated rectangles. Thus, the overlap between this class and sf::RectangleShape is close to 100%. The only thing that differs is how to define the shape, but the shape itself is still a rectangle. So maybe a better option would be to provide alternative ways to build a rectangle?
> ~ *Laurent*

I believe this highly depends on the API we build for `sf::LineShape`. Yes, it might be very similar to a rectangle shape underneath, but if the access APIs work with start/end points, then the usage becomes quite different and does justify a dedicated class for me.

> > A dedicated LineShape can add thickness to the sides only.
>
> That's not obvious. It would feel strange / inconsistent to have outline only on two sides. That's in my opinion a good example of what kind of problems we'd get with a line-with-thickness class
> ~ *Laurent*

This is certainly something we need to make a decision on. Should a line only have an outline on the "long" side or all around it? Or should it be configurable?

*Are there any additional arguments?*

----

My personal design goals for a `sf::LineShape` class are:

- Needs to fully implement the `sf::Shape` interface
- "Simple" API to handling the two end points / constructing a line
- Needs to clearly define the scope of the class:
	- No arrows
	- No rounded ends
	- No joints
	- Just a line between two points with an optional thickness

ChrisThrasher:
Rebased after #2055 was merged just to be sure that there we no new conflicts.

eXpl0it3r:

ChrisThrasher:
Is the team still interested in this PR? If not then you may go ahead and close it.

Bromeon:
In my opinion, `LineShape` would be a nice addition! There are still some open questions though: https://github.com/SFML/SFML/pull/2063#issuecomment-1113970104

It might be some work to do a bigger refactoring (and possibly out of this PR's scope), but on the other hand I think SFML 3 is the time to think about such things. So I'd appreciate other opinions on the matter of the `sf::Shape` hierarchy.

kimci86:
I think this PR is not already merged because, in principle, a line does not have a surface, as opposed to other implemented shapes.
I am in favor of Bromeon's suggestion number 1:
> Add a new intermediate interface sf::AreaShape or sf::FilledShape which provides methods regarding texture, texture rect, fill color. These methods would be moved down from sf::Shape. All existing shapes would inherit the new interface, while sf::Line would inherit the (now reduced) sf::Shape.

ChrisThrasher:
What I'm gathering is that `sf::LineShape`'s inclusion hinges on us crafting a better, more complicated abstraction for prebuilt shapes. I'm happy to get into that but I think a better question to ask is, what shapes _do_ we want? I've previously mentioned `sf::EllipseShape` and `sf::CurveShape` (a piecewise-linear curve). What other shapes would hypothetically inherit from a stripped-down `sf::Shape`?

Bromeon:
Things that come to my mind: Bezier curve, 2D capsule (a circle which is stretched _in the middle_ 💊).

Not saying SFML is the right abstraction layer for these shapes, but such an abstract base class would also be open to be implemented by users and extension libraries.

ChrisThrasher:
Rebasing to keep up with `master`. No changes.

eXpl0it3r:

ChrisThrasher:
Fixed trivial merge conflict.

eXpl0it3r:

ChrisThrasher:

ChrisThrasher:
Rebased and fixed compiler error.

ChrisThrasher:
Rebased because it's been a while and because I needed to convert some `size_t`s to `std::size_t`s.

vittorioromeo:

vittorioromeo:
Consider calling this `sf::LineSegmentShape` so that we can use `sf::LineShape` for a more full-fledged line class.

vittorioromeo:
@eXpl0it3r @Bromeon: would you be in favour of this PR if we rename it as `sf::LineSegmentShape` and make it clear that this is just a basic primitive over a line with some thickness, and then we can build on top of it (e.g. `sf::LineShape` could be a bunch of `sf::LineSegmentShape`, or we could have a `sf::FancyLineSegmentShape` where we can add rounded edges/arrows and stuff like that).

What do you think?

Bromeon (approved):
I'm happy with this PR, and I don't think it's necessary to rename `LineShape` to `LineSegmentShape`.

In 2D graphics, the term "line" is often used to refer to the concept as embodied by this PR. I encountered "line segment" more in mathematical contexts where lines are infinite, or when the graphics library needs to explicitly distinguish the two.

ChrisThrasher:

ChrisThrasher:
Rebased to get new CI jobs.

eXpl0it3r:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:
Fixed small merge conflict with `master`.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:
I'm not interested in maintaining or discussing this anymore so I'm closing it (again). If motivation returns we can resurrect this but it's not worth keeping this open in the meantime.

PR #2068, opened 2022-04-11
Remove iOS 7 workarounds

ChrisThrasher:
## Description

Removing some iOS 7 workaround introduced by cac4d58b3 in 2014 when iOS 7 was only 1 year old. Now we require iOS 13 or new so this code should never get executed.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android


eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:
Thanks for the clean up!

PR #2070, opened 2022-04-11
Fix more compiler warnings

ChrisThrasher:
## Description

Changing how we apply compiler warning revealed an issue where Windows unity builds were missing a shadowed variable warning. Not sure why exactly this was happening but simplifying how we apply warnings meant this shadowed variable is now being detected. While fixing this I also discovered that warnings weren't be applied to the unit tests so I fixed that as well and fixed the one warning in the tests.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android


ChrisThrasher:
I forgot that the Windows jobs are failing with more compiler warnings I have to fix. I'll get around to fixing those soon.

src/SFML/Audio/SoundFileWriterWav.cpp
```diff
@@ -137,8 +137,7 @@ bool SoundFileWriterWav::writeHeader(unsigned int sampleRate, unsigned int chann
     m_file.write(mainChunkId, sizeof(mainChunkId));
 
     // Write the main chunk header
-    Uint32 mainChunkSize = 0; // placeholder, will be written later
-    encode(m_file, mainChunkSize);
+    encode(m_file, static_cast<Uint32>(0));
```
kimci86, line 140: I would keep the comment here about 0 being a temporary placeholder.

cmake/CompilerWarnings.cmake
```diff
@@ -63,7 +63,6 @@ function(set_file_warnings)
         -Woverloaded-virtual # warn if you overload (not override) a virtual function
         -Wconversion # warn on type conversions that may lose data
         -Wsign-conversion # warn on sign conversions
-        -Wdouble-promotion # warn if float is implicit promoted to double
```
Bromeon, line 66: Why remove this? Sure, there's technically no loss of precision, but this is not so uncontroversial. Java allows implicit widening conversions, while Kotlin and Rust do not.

test/TestUtilities/SystemUtil.hpp
```diff
@@ -54,8 +54,8 @@ struct ApproxVec
 // Utilities for approximate equality
 struct ApproxDeg
 {
-    ApproxDeg(double degrees)
-        : degrees(static_cast<float>(degrees)) {}
+    ApproxDeg(double deg)
+        : degrees(static_cast<float>(deg)) {}
```
Bromeon, line 58: This is an annoying warning, which also forced us to do ugly stuff like `thePosition` in other places. It's even worse when the parameter name appears in the public API and thus the doc.

I guess it's not possible to selectively disable it only for constructors?

cmake/CompilerWarnings.cmake
ChrisThrasher: This is from the extended commit message from the commit that made this change:

> Promoting from `float` to `double` is a lossless operation. We do
not need to introduce a `static_cast` whenever this happens. Lossy
conversions back to `float` are easily caught by `-Wconversion`. I
don't think this warning adds enough value to justify the increased
verbosity and decreased readabilty of all the additional casts it
requires we write.

One particular problem is that we keep this warning, we must add many dozen `static_cast<double>(someFLoat)` calls to the tests which drastically impacts readability while providing no extra correctness.

test/TestUtilities/SystemUtil.hpp
ChrisThrasher: `-Wshadow` does get very annoying when writing constructors for structs. It pushes me to use aggregate initialization for structs because of this problem. I don't believe it can be selectively disabled for constructors.

src/SFML/Audio/SoundFileWriterWav.cpp
ChrisThrasher: Fixed

cmake/CompilerWarnings.cmake
kimci86: Having an `Approx` helper that works on float would indeed solve this cleanly.

ChrisThrasher: Right, maybe we have to write our own wrappers around `doctest::Approx(double)` to work around this. I wasn't excited to remove `-Wdouble-promotion` and would prefer to keep it if it didn't mean destroying test legibility. However I'm also not excited to write wrappers around the test framework since I already think our testing infrastructure is growing too complicated.

test/TestUtilities/SystemUtil.hpp
Bromeon: What I found: 
* clang: https://reviews.llvm.org/D18271?id=55044
* gcc: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78147

> Clang has an extra warning flag, -Wshadow-field-in-constructor, which controls this (from what I can tell).  By default in Clang, -Wshadow doesn't enable this you need to separately enable it.

Doesn't look like g++ has adopted anything similar.

ChrisThrasher:
I fixed the last few MSVC warnings so this PR is ready for review now.

cmake/CompilerWarnings.cmake
ChrisThrasher: I reintroduced this warning. Luckily it wasn't as hard as I thought to keep it.

cmake/CompilerWarnings.cmake
```diff
@@ -3,7 +3,7 @@
 # https://github.com/lefticus/cppbestpractices/blob/master/02-Use_the_Tools_Available.md
 
 # Helper function to enable compiler warnings for a specific set of files
-function(set_file_warnings)
+function(set_file_warnings target)
```
kimci86, line 6: The description in comment and function name were not updated.

test/System/Time.cpp
```diff
@@ -25,8 +25,6 @@ TEST_CASE("sf::Time class - [system]")
             CHECK(time.asMicroseconds() == 123'000'000);
 
             CHECK(sf::seconds(1'000.0f).asMicroseconds() == 1'000'000'000);
-            CHECK(sf::seconds(0.000002f).asMicroseconds() == Approx(2).epsilon(1.f));
```
kimci86, line 28: Why are some checks removed?

cmake/CompilerWarnings.cmake
ChrisThrasher: Good point. The function name is out of date too. I fixed the comment and function name.

eXpl0it3r:

ChrisThrasher:
More MinGW floating point precision issues plaguing the tests. It'll take some back and forth to fix those precision issues and get all the tests passing in that environment.

eXpl0it3r:

eXpl0it3r:

eXpl0it3r:

ChrisThrasher:
Sweet, it looks like the MinGW test failures have been addressed. I don't have any more changes planned for this PR.

eXpl0it3r:

eXpl0it3r:

.gitattributes
```diff
@@ -1,13 +0,0 @@
...
-cmake/toolchains/android.toolchain.cmake -text -eol
-
-*.png -text -eol
-*.jpg -text -eol
-*.wav -text -eol
-*.ogg -text -eol
-*.flac -text -eol
-*.tff -text -eol
-*.icns -text -eol
-*.rtf -text -eol
```
eXpl0it3r, line 13: Please don't remove this file.

Git can be configured in different ways to deal with line endings. The `.gitattributes` files ensures that line endings are handled correctly for the repository (e.g. not enforcing different line endings in the extlibs directory).

See also https://git-scm.com/docs/gitattributes

src/SFML/Audio/SoundFileWriterWav.cpp
```diff
@@ -137,8 +137,7 @@ bool SoundFileWriterWav::writeHeader(unsigned int sampleRate, unsigned int chann
     m_file.write(mainChunkId, sizeof(mainChunkId));
 
     // Write the main chunk header
-    Uint32 mainChunkSize = 0; // placeholder, will be written later
-    encode(m_file, mainChunkSize);
+    encode(m_file, static_cast<Uint32>(0)); // 0 is a placeholder, will be written later
```
eXpl0it3r, line 140: Does `0` really need casting? What about using `0u`?

.gitattributes
ChrisThrasher: That was an accident sorry! I deleted that file to see what would happen then probably got distracted and forgot to undo that change so it got accidentally committed.

src/SFML/Audio/SoundFileWriterWav.cpp
ChrisThrasher: I added that because there are many overloads of `encode` that expect different types for that 2nd parameter and I wanted to be absolutely sure the same overload was being called. `0u` would be correct on many platforms but I don't believe it's guaranteed that `0u` will be 32 bits wide so I went with the static cast to be safe.

eXpl0it3r: Ah right, didn't remember that here are different overloads, in that case it makes sense.
Thanks for clarifying. 🙂

eXpl0it3r:

test/TestUtilities/SystemUtil.hpp
eXpl0it3r: As we're not enabling any new warnings here, lets continue the shadow discussion separately.

eXpl0it3r (approved):

eXpl0it3r:
Thanks for these fixes 🥳

PR #2064, opened 2022-04-08
Fix setSmooth(false) not having an effect before first text drawcall

oomek:
* [X] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before?
On Discord
* [X] Does the code follow the SFML [Code Style Guide](https://www.sfml-dev.org/style.php)?
* [X] Have you provided some example/test code for your changes?

----

## Description

When font.setSmooth(false) is called before the first text drawcall it appears that glyph pages are not yet generated.

## Tasks

* [X] Tested on Linux
* [X] Tested on Windows

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode(1280, 720), "Minimal, complete and verifiable example");
    window.setFramerateLimit(60);

    sf::Font font;
    font.loadFromFile("8bit.ttf");

    sf::Text text("Font Smoothing Test", font);
    text.setCharacterSize(16);
    text.setScale(4,4);
    font.setSmooth(false);

    while (window.isOpen())
    {
        sf::Event event;
        while (window.pollEvent(event))
        {
            if (event.type == sf::Event::Closed)
                window.close();
        }

        window.clear();
        window.draw(text);
        window.display();
    }
}
```

PR #2072, opened 2022-04-12
Fix font pages not being created with the desired smoothness

kimci86:
## Description

This is an alternative solution for the issue described in #2064
As `try_emplace` is not available with C++03, extra `Page` copies occur, but at most once per character size and per font.

Besides, `Page` not having a default constructor anymore implies it is not possible to use `operator[]` on `m_pages`, which is a good thing to force that the desired smoothness is always used to initialize new glyph pages.

For reference, smoothing for fonts was first added in #1690.

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

An example is provided in the original PR #2064

### DIFF
--- a/include/SFML/Graphics/Font.hpp
+++ b/include/SFML/Graphics/Font.hpp
@@ -335,7 +335,7 @@ private:
     ////////////////////////////////////////////////////////////
     struct Page
     {
-        Page();
+        explicit Page(bool smooth);
 
         GlyphTable       glyphs;  //!< Table mapping code points to their corresponding glyph
         Texture          texture; //!< Texture containing the pixels of the glyphs
@@ -348,6 +348,16 @@ private:
     ///
     ////////////////////////////////////////////////////////////
     void cleanup();
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Find or create the glyphs page corresponding to the given character size
+    ///
+    /// \param characterSize Reference character size
+    ///
+    /// \return The glyphs page corresponding to \a characterSize
+    ///
+    ////////////////////////////////////////////////////////////
+    Page& loadPage(unsigned int characterSize) const;
 
     ////////////////////////////////////////////////////////////
     /// \brief Load a new glyph and store it in the cache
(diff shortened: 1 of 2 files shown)


eXpl0it3r:

eXpl0it3r (approved):

include/SFML/Graphics/Font.hpp
```diff
@@ -335,7 +335,7 @@ class SFML_GRAPHICS_API Font
     ////////////////////////////////////////////////////////////
     struct Page
     {
-        Page();
+        Page(bool smooth);
```
Bromeon, line 338: should be `explicit`

src/SFML/Graphics/Font.cpp
```diff
@@ -348,7 +348,11 @@ const Font::Info& Font::getInfo() const
 const Glyph& Font::getGlyph(Uint32 codePoint, unsigned int characterSize, bool bold, float outlineThickness) const
 {
     // Get the page corresponding to the character size
-    GlyphTable& glyphs = m_pages[characterSize].glyphs;
+    PageTable::iterator pageIterator = m_pages.find(characterSize);
+    if (pageIterator == m_pages.end())
+        pageIterator = m_pages.insert(std::make_pair(characterSize, Page(m_isSmooth))).first;
+
+    GlyphTable& glyphs = pageIterator->second.glyphs;
```
Bromeon, line 355: Since this code appears 3 times, would it make sense to have a dedicated function?
```cpp
Page& Font::loadPage(unsigned int characterSize);
```

include/SFML/Graphics/Font.hpp
kimci86: Fixed.

src/SFML/Graphics/Font.cpp
kimci86: It looks much better with a dedicated function. Thank you.

eXpl0it3r:

eXpl0it3r:
Thank you for backporting this to 2.6! 🎉

PR #2080, opened 2022-04-21
2.6.x -> master backmerge

eXpl0it3r:
Merging back the fixes from the 2.6.x branch to master


kimci86:
Is it the moment to remove `Font::loadPage` and use `try_emplace` instead, or would it be done later on master?

eXpl0it3r:
That should be done afterwards, we're "copying" (i.e. merging) the original commits into master first, so you have the same commit on both branches.

PR #2081, opened 2022-04-21
Add tests for `sf::err`

ChrisThrasher:
## Description

Because `sf::err` is just a `std::ostream&` we can redirect it to capture whatever is printed to it. These tests are largely just testing `std::ostream` since the custom logic inside `sf::err` is difficult to observe. Regardless, these tests do a good job demonstrating how to test error text output for classes that use `sf::err` in their implementation like those which were changed in https://github.com/SFML/SFML/pull/2043.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

### DIFF
--- a/test/System/Err.cpp
+++ b/test/System/Err.cpp
@@ -1,0 +1,40 @@
+#include <SFML/System/Err.hpp>
+#include <sstream>
+
+#include <doctest.h>
+
+TEST_CASE("sf::err - [system]")
+{
+    SUBCASE("Overflow default buffer")
+    {
+        // No assertion macros in this subcase since nothing about this can be directly observed.
+        // Intention is to ensure DefaultErrStreamBuf::overflow gets called.
+        sf::err() << "SFML is a simple, fast, cross-platform and object-oriented multimedia API."
+            "It provides access to windowing, graphics, audio and network."
+            "It is written in C++, and has bindings for various languages such as C, .Net, Ruby, Python.";
+    }
+
+    SUBCASE("Redirect buffer to observe contents")
+    {
+        sf::err() << "We'll never be able to observe this" << std::endl; // Ensure buffer is flushed
+        const auto defaultStreamBuffer = sf::err().rdbuf();
+        CHECK(defaultStreamBuffer != nullptr);
+
+        std::stringstream stream;
+        sf::err().rdbuf(stream.rdbuf());
+        sf::err() << "Something went wrong!\n";
+        CHECK(stream.str() == "Something went wrong!\n");
+
+        sf::err().rdbuf(nullptr);
+        sf::err() << "Sent to the abyss";
+        CHECK(stream.str() == "Something went wrong!\n");
+
+        sf::err().rdbuf(stream.rdbuf());
+        sf::err() << "Back to the stringstream :)\n";
+        CHECK(stream.str() == "Something went wrong!\nBack to the stringstream :)\n");
+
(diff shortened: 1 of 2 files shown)


eXpl0it3r:

eXpl0it3r:

test/System/Err.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include <doctest.h>
+
+TEST_CASE("sf::err - [system]")
+{
+    SUBCASE("Overflow default buffer")
+    {
+        sf::err() << "SFML is a simple, fast, cross-platform and object-oriented multimedia API."
+            "It provides access to windowing, graphics, audio and network."
+            "It is written in C++, and has bindings for various languages such as C, .Net, Ruby, Python.";
+    }
```
Bromeon, line 13: What does this test? There's no assertion 🤔

test/System/Err.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+    SUBCASE("Redirect buffer to observe contents")
+    {
+        sf::err() << "We'll never be able to observe this" << std::endl;
```
Bromeon, line 17: Maybe comment that the `endl` is truly required for the flush, not that this gets replaced by `\n` in the future.

eXpl0it3r:

test/System/Err.cpp
ChrisThrasher: Yeah that's a weird subcase. It's there to make sure [`DefaultErrStreamBuffer::overflow`](https://github.com/SFML/SFML/blob/master/src/SFML/System/Err.cpp#L61-L79) gets run. In theory that function could cause a crash which would be observable as a test failure but there is seemingly nothing else we can observe about `overflow`.

test/System/Err.cpp
ChrisThrasher: Done

test/System/Err.cpp
eXpl0it3r: Can you redirect `std::cerr` (not `sf::err()`) to at least check that the content is properly outputted?

ChrisThrasher: I tried that and it didn't work. `sf::err` and `std::cerr` are separate output streams which appears to be in the intention behind `DefaultErrStreamBuffer`'s existence. If we just wanted to use `std::cerr` then there would no longer be a reason for `sf::err` to exist.

eXpl0it3r: Yeah, it doesn't exactly [look simple](https://stackoverflow.com/a/5419409/1034248), so I'd rather not hack anything together for that. The "issue" is, that SFML directly writes to `stderr` and doesn't use `std::cerr` which could be redirected, which kind of begs the question, why we aren't using `std::cerr` to begin with. Like you can just as easily do `std::cerr.rdbuf(buffer.rdbuf())`.

Anyways off topic in this review comment.

I'd probably add a comment explaining why there's no `CHECK` where the check would usually be.

ChrisThrasher: > I'd probably add a comment explaining why there's no CHECK where the check would usually be.

Done

eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:
Test coverage 📈 🥳

PR #2082, opened 2022-04-26
Backmerge 2.6.x to master

eXpl0it3r:
Bringing the DRM / KMS feature from 2.6.x to master


eXpl0it3r:
@oomek / @mickelson / @substring could you give this PR a try, to see that the DRM/KMS implementation also works properly with SFML 3 and/or that I didn't screw something up with the "porting"?

oomek:
After a few fixes works great on Arch Linux, but fails compiling on Raspbian `#include <filesystem>` seems to be the culprit.

oomek:
My test app is running ok on Raspbian and Arch Linux. Will check Attract Mode later after I strip it off of sf::Lock, Thread and Mutex.

eXpl0it3r:
I decided to merge it for now, if you find anything broken, then don't hesitate to open a new issue 🙂

oomek:
I've tried my best to port Attract Mode to SFLM 3.0.0. The amount of breakage just overwhelmed me halfway through besause of simulteanous c++17 upgrade. We'll have to stick to synthetic tests for now I'm afraid.

PR #2085, opened 2022-04-29
Add polar coordinate `sf::Vector2<T>` constructor

ChrisThrasher:
## Description

This is the corollary to [`sf::Vector2<T>::angle()`](https://github.com/SFML/SFML/blob/master/include/SFML/System/Vector2.hpp#L121) which produces an angle from a vector. To avoid circular dependencies and to ensure we can construct vectors of all types from angles, I made this a new `sf::Vector2<T>` constructor overload instead of adding a new member function to `sf::Angle`. 

The motivation for this was simplifying how we currently create unit vectors from angles which looks like [`sf::Vector2f::UnitX.rotatedBy(rotation)`](https://github.com/ChrisThrasher/boids/blob/sfml-v3/src/boid.cpp#L23). I wouldn't expect many users to realize this is possible since at first glance it's not obvious what I'm trying to do here.

As an aside, this PR also removes the degree sign from Vector2.hpp which is an extended ASCII character. SFML appears to avoid extended ASCII characters in source files so I wanted to bring this file in line with the rest of the library.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

### DIFF
--- a/examples/tennis/Tennis.cpp
+++ b/examples/tennis/Tennis.cpp
@@ -200,8 +200,7 @@ int main()
             }
 
             // Move the ball
-            float factor = ballSpeed * deltaTime;
-            ball.move({std::cos(ballAngle.asRadians()) * factor, std::sin(ballAngle.asRadians()) * factor});
+            ball.move({ballSpeed * deltaTime, ballAngle});
 
             #ifdef SFML_SYSTEM_IOS
             const std::string inputString = "Touch the screen to restart.";
--- a/include/SFML/System/Vector2.hpp
+++ b/include/SFML/System/Vector2.hpp
@@ -52,7 +52,7 @@ public:
     constexpr Vector2();
 
     ////////////////////////////////////////////////////////////
-    /// \brief Construct the vector from its coordinates
+    /// \brief Construct the vector from cartesian coordinates
     ///
     /// \param X X coordinate
     /// \param Y Y coordinate
@@ -73,6 +73,22 @@ public:
     ////////////////////////////////////////////////////////////
     template <typename U>
     constexpr explicit Vector2(const Vector2<U>& vector);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Construct the vector from polar coordinates
+    ///
+    /// \param r   Length of vector (can be negative)
+    /// \param phi Angle from X axis
+    ///
+    /// Note that this constructor is lossy: calling length() and angle()
+    /// may return values different to those provided in this constructor.
+    ///
+    /// In particular, these transforms can be applied:
+    /// * Vector2(r, phi) == Vector2(-r, phi + 180_deg)
+    /// * Vector2(r, phi) == Vector2(r, phi + n * 360_deg)
+    ///
+    ////////////////////////////////////////////////////////////
+    Vector2(T r, Angle phi);
     
     ////////////////////////////////////////////////////////////
     /// \brief Length of the vector <i><b>(floating-point)</b></i>.
@@ -103,7 +119,7 @@ public:
     /// 
     /// \return The smallest angle which rotates \c *this in positive
     /// or negative direction, until it has the same direction as \c rhs.
-    /// The result has a sign and lies in the range [-180, 180°).
+    /// The result has a sign and lies in the range [-180, 180) degrees.
     /// \pre Neither \c *this nor \c rhs is a zero vector.
     ///
     ////////////////////////////////////////////////////////////
@@ -114,7 +130,7 @@ public:
     /// 
     /// For example, the vector (1,0) corresponds to 0 degrees, (0,1) corresponds to 90 degrees.
     /// 
-    /// \return Angle in the range [-180°, 180°).
+    /// \return Angle in the range [-180, 180) degrees.
     /// \pre This vector is no zero vector.
     ///
     ////////////////////////////////////////////////////////////
--- a/test/System/Vector2.cpp
+++ b/test/System/Vector2.cpp
@@ -37,6 +37,57 @@ TEST_CASE("sf::Vector2 class template - [system]")
             CHECK(vector.x == static_cast<int>(sourceVector.x));
             CHECK(vector.y == static_cast<int>(sourceVector.y));
         }
+
+        SUBCASE("Length and angle constructor")
+        {
+            CHECK(sf::Vector2f(0, sf::degrees(0)) == sf::Vector2f(0, 0));
+            CHECK(sf::Vector2f(0, sf::degrees(45)) == sf::Vector2f(0, 0));
+            CHECK(sf::Vector2f(0, sf::degrees(90)) == sf::Vector2f(0, 0));
+            CHECK(sf::Vector2f(0, sf::degrees(135)) == sf::Vector2f(0, 0));
+            CHECK(sf::Vector2f(0, sf::degrees(180)) == sf::Vector2f(0, 0));
+            CHECK(sf::Vector2f(0, sf::degrees(270)) == sf::Vector2f(0, 0));
+            CHECK(sf::Vector2f(0, sf::degrees(360)) == sf::Vector2f(0, 0));
+            CHECK(sf::Vector2f(0, sf::degrees(-90)) == sf::Vector2f(0, 0));
+            CHECK(sf::Vector2f(0, sf::degrees(-180)) == sf::Vector2f(0, 0));
+            CHECK(sf::Vector2f(0, sf::degrees(-270)) == sf::Vector2f(0, 0));
+            CHECK(sf::Vector2f(0, sf::degrees(-360)) == sf::Vector2f(0, 0));
+
+            CHECK(sf::Vector2f(1, sf::degrees(0)) == sf::Vector2f(1, 0));
+            CHECK(sf::Vector2f(1, sf::degrees(45)) == ApproxVec2(std::sqrt(2.f) / 2.f, std::sqrt(2.f) / 2.f));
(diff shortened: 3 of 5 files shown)


Bromeon:
Nice idea! I think the constructor should take angle **and length** though.
Basically the polar-coordinate equivalent `(r, phi)` to cartesian `(x, y)`.

Also, we should disallow construction from angles for non-float vector types.

ChrisThrasher:
> Nice idea! I think the constructor should take angle **and length** though. Basically the polar-coordinate equivalent `(r, phi)` to cartesian `(x, y)`.

Would you rather the length be an optional parameter or keep two constructors, one for unit vectors and one for vectors with a specified length?

> Also, we should disallow construction from angles for non-float vector types.

Good point. I'll fix that right now.

Bromeon:
> Would you rather the length be an optional parameter or keep two constructors, one for unit vectors and one for vectors with a specified length?

Personally I would not treat the `(1, phi)` construction as a special case, as we don't do that for any value of `x`, `y` or `phi` either. 

A general `(r, phi)` constructor should be a good trade-off regarding flexibility and API surface.

include/SFML/System/Vector2.hpp
```diff
@@ -73,6 +73,14 @@ class Vector2
...
     template <typename U>
     constexpr explicit Vector2(const Vector2<U>& vector);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Construct a unit vector from an angle
+    ///
+    /// \param angle Angle
+    ///
+    ////////////////////////////////////////////////////////////
+    explicit Vector2(const Angle& angle);
```
Bromeon, line 83: I'd probably pass `Angle` by value and name it `phi` like here:
https://github.com/SFML/SFML/blob/c7327353b3eaa1ac62c13195a1424033121fb5de/include/SFML/System/Vector2.hpp#L140

(also applies once the constructor signature changes)

ChrisThrasher: Weird, the implementation uses the name `angle`. I'll use `phi` if you want though.

https://github.com/SFML/SFML/blob/3ae85854e4674debe3b30ae2f6cf9d9926738789/include/SFML/System/Vector2.inl#L110

Bromeon: Oh, that's an oversight, feel free to fix it. Ultimately it's not _that_ important 🙂

ChrisThrasher: Might be worth some library-wide rename alongside changing many `const Angle&` params to `Angle`. For a later PR but I'll make a note to check for consistent `sf::Angle` use.

ChrisThrasher:
@Bromeon I changed that constructor to use a length/angle pair and added some new tests cases to make sure everything working as intended. I agree this is a better plan than two constructors and/or something involving optional parameters.

include/SFML/System/Vector2.hpp
Bromeon: I was mostly thinking about calling an angle `phi` in the context of polar coordinates (i.e. a vector) -- i.e. in cases where there is another coordinate `r` (radius). Not sure if that necessarily applies to the rest of the code base; so `angle` might still be fine elsewhere.

examples/tennis/Tennis.cpp
```diff
@@ -200,8 +200,7 @@ int main()
             }
 
             // Move the ball
-            float factor = ballSpeed * deltaTime;
-            ball.move({std::cos(ballAngle.asRadians()) * factor, std::sin(ballAngle.asRadians()) * factor});
+            ball.move({ballSpeed * deltaTime, ballAngle});
```
Bromeon, line 203: Nice one! This also shows the strength of a type-safe `sf::Angle` type, otherwise we would have needed something like `sf::Vector2f::fromPolar(r, phi)`.

Are people generally OK with this being a constructor, i.e. allowing short syntax?

test/System/Vector2.cpp
```diff
@@ -37,6 +37,48 @@ TEST_CASE("sf::Vector2 class template - [system]")
             CHECK(vector.x == static_cast<int>(sourceVector.x));
             CHECK(vector.y == static_cast<int>(sourceVector.y));
         }
+
+        SUBCASE("Length and angle constructor")
+        {
+            CHECK(sf::Vector2f(0, sf::degrees(0)) == sf::Vector2f(0, 0));
+            CHECK(sf::Vector2f(0, sf::degrees(180)) == sf::Vector2f(0, 0));
```
Bromeon, line 44: Could we maybe add some arbitrary values for `phi` here (not just half-circles), to see if it's still the zero vector?
Possibly we need `ApproxVec` for rounding errors.

include/SFML/System/Vector2.hpp
```diff
@@ -73,6 +73,15 @@ class Vector2
...
     constexpr explicit Vector2(const Vector2<U>& vector);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Construct the vector from a length and angle
+    ///
+    /// \param length Length of vector
+    /// \param phi    Angle off X axis
+    ///
+    ////////////////////////////////////////////////////////////
+    Vector2(T length, Angle phi);
```
Bromeon, line 84: Do we want to allow negative values of `length`? (as it's currently handled in tests)

* If yes, we should probably rename the parameter, since it's no longer a length and it won't match the getter `length()`. Maybe `r`.
* Also, in that case we should document that it is allowed, and maybe even that `Vector2(r, phi)` is equivalent to `Vector2(-r, phi + 180°)`.
* Should we mention something about `Vector2(r, phi)` being equivalent to `Vector2(r, phi + n * 360°)`? 

Those things may sound obvious, but I think the important message is that the polar angle constructor is lossy, and it may not be able to retrieve the values later as they were provided in the constructor.

Bromeon: Furthermore, maybe we should change the `\brief` part to:
```cpp
/// \brief Construct the vector from polar coordinates
```
while for the `(x, y)` constructor, we could say:
```cpp
/// \brief Construct the vector from cartesian coordinates
```

test/System/Vector2.cpp
```diff
@@ -37,6 +37,48 @@ TEST_CASE("sf::Vector2 class template - [system]")
...
+            CHECK(sf::Vector2f(3.14f, sf::degrees(45)) == ApproxVec(3.14f * std::sqrt(2.f) / 2.f, 3.14f * std::sqrt(2.f) / 2.f));
+            CHECK(sf::Vector2f(3.14f, sf::degrees(90)) == ApproxVec(0, 3.14f));
+            CHECK(sf::Vector2f(3.14f, sf::degrees(135)) == ApproxVec(-3.14f * std::sqrt(2.f) / 2.f, 3.14f * std::sqrt(2.f) / 2.f));
+            CHECK(sf::Vector2f(3.14f, sf::degrees(180)) == ApproxVec(-3.14f, 0));
+            CHECK(sf::Vector2f(3.14f, sf::degrees(270)) == ApproxVec(0, -3.14f));
+            CHECK(sf::Vector2f(3.14f, sf::degrees(360)) == ApproxVec(3.14f, 0));
+            CHECK(sf::Vector2f(3.14f, sf::degrees(-90)) == ApproxVec(0, -3.14f));
+            CHECK(sf::Vector2f(3.14f, sf::degrees(-180)) == ApproxVec(-3.14f, 0));
+            CHECK(sf::Vector2f(3.14f, sf::degrees(-270)) == ApproxVec(0, 3.14f));
+            CHECK(sf::Vector2f(3.14f, sf::degrees(-360)) == ApproxVec(3.14f, 0));
```
Bromeon, line 80: Is there a particular reason why you use "nearly Pi"? It doesn't seem like you make use of any of its properties, so I wonder if another arbitrary magic number might be less suggestive.

test/System/Vector2.cpp
ChrisThrasher: Fixed

test/System/Vector2.cpp
ChrisThrasher: Haha yeah that was a bad idea. I'll change that.

include/SFML/System/Vector2.hpp
ChrisThrasher: > Do we want to allow negative values of length? (as it's currently handled in tests)

In the Tennis example I modified, imagine if you wanted the ball to go backwards along a certain trajectory. In that case it's really helpful to use a negative radius/length instead of having extra logic to flip the angle. Besides, the math is perfectly well defined for all values of `r` and `phi`.

> If yes, we should probably rename the parameter, since it's no longer a length and it won't match the getter length(). Maybe r.

No problem. I'll adopt that r/phi terminology. I think it looks good and I like how it helps avoid confusion with `length()` and `angle()` which as you pointed out won't necessarily return the same values that were provided to this constructor. 

> Should we mention something about Vector2(r, phi) being equivalent to Vector2(r, phi + n * 360°)?

No harm in a little extra documentation :)

ChrisThrasher:
Rebasing on master and fixing the new compiler warnings added by #2070 would require making changes that #2089 renders unnecessary so I'd prefer to block this PR until a decision can be reached on #2089.

ChrisThrasher:
@Bromeon Anything more you want to discuss about this?

Bromeon (approved):
Looks good to me!

eXpl0it3r:

eXpl0it3r:

ChrisThrasher:
Rebased after #2055 was merged just to be sure that there we no new conflicts.

eXpl0it3r:
Thanks for this addition. Lets see how it will be received/utilized. 🙂

PR #2086, opened 2022-04-29
Add extension methods to `Vector3<T>`

Bambo-Borris:
Thanks a lot for making a contribution to SFML! 🙂

Before you create the pull request, we ask you to check the follow boxes. (For small changes not everything needs to ticked, but the more the better!)

* [ ] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before?
* [x] Does the code follow the SFML [Code Style Guide](https://www.sfml-dev.org/style.php)?
* [ ] Have you provided some example/test code for your changes?
* [ ] If you have additional steps which need to be performed list them as tasks!

----

## Description
Added extension methods to `Vector3<T>` based on existing functionality for `Vector2<T>` and similar functionality present within the Thor library. Tests also added for these extension methods. There are additional methods on `Vector2<T>` which weren't implemented as I didn't deem them necessary, but if the community desires them then I'm open to adding more.

Methods added: 
- length
- legthSq
- normalized
- cross
- dot
- cwiseMul
- cwiseDiv

This PR is related to the issue #2084

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

PR #2089, opened 2022-05-02
Remove unnecessary lossy conversions in test utilities

ChrisThrasher:
## Description

Both ApproxVec and ApproxDeg accept a double then convert that to a float just to then convert it back to a double. We can save an extra lossy conversion by just accepting a float instead of a double in the first place. I never realized this until #2070 was recently merged which started applying compiler warnings to the unit tests.

On top of that I simplified how we construct the approximate comparison structures for vectors and angles such that we're using strong types to their fullest potential. I kept those as separate commits in case we change our mind about those changes and only use the first.

If accepted, I'd prefer this get merged before #2085 or #2086 as it will cause significant (but easily fixable) conflicts with both. Even just rebasing #2085 on master will require I add a lot of `static_casts` which would then get obviated by this PR. 

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

### DIFF
--- a/test/System/Vector2.cpp
+++ b/test/System/Vector2.cpp
@@ -195,31 +195,31 @@ TEST_CASE("sf::Vector2 class template - [system]")
 
         CHECK(v.length() == Approx(3.84187));
         CHECK(v.lengthSq() == Approx(14.7599650969));
-        CHECK(v.normalized() == ApproxVec(0.624695, 0.780869));
+        CHECK(v.normalized() == ApproxVec(0.624695f, 0.780869f));
 
         const sf::Vector2f w(-0.7f, -2.2f);
         
         CHECK(w.length() == Approx(2.30868));
         CHECK(w.lengthSq() == Approx(5.3300033));
-        CHECK(w.normalized() == ApproxVec(-0.303204, -0.952926));
+        CHECK(w.normalized() == ApproxVec(-0.303204f, -0.952926f));
     }
 
     SUBCASE("Rotations and angles")
     {
         const sf::Vector2f v(2.4f, 3.0f);
         
-        CHECK(v.angle() == ApproxDeg(51.3402));
-        CHECK(sf::Vector2f::UnitX.angleTo(v) == ApproxDeg(51.3402));
-        CHECK(sf::Vector2f::UnitY.angleTo(v) == ApproxDeg(-38.6598));
+        CHECK(v.angle() == ApproxDeg(51.3402f));
+        CHECK(sf::Vector2f::UnitX.angleTo(v) == ApproxDeg(51.3402f));
+        CHECK(sf::Vector2f::UnitY.angleTo(v) == ApproxDeg(-38.6598f));
 
         const sf::Vector2f w(-0.7f, -2.2f);
 
-        CHECK(w.angle() == ApproxDeg(-107.65));
-        CHECK(sf::Vector2f::UnitX.angleTo(w) == ApproxDeg(-107.65));
-        CHECK(sf::Vector2f::UnitY.angleTo(w) == ApproxDeg(162.35));
+        CHECK(w.angle() == ApproxDeg(-107.65f));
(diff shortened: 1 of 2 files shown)


Bromeon:
I think this comes [from the `Approx` constructor taking `double` or immediately casting to it](https://github.com/doctest/doctest/blob/7b9885133108ae301ddd16e2651320f54cafeba7/doctest/doctest.h#L1040-L1054):
```cpp
class DOCTEST_INTERFACE Approx
{
public:
    explicit Approx(double value);

    Approx operator()(double value) const;

#ifdef DOCTEST_CONFIG_INCLUDE_TYPE_TRAITS
    template <typename T>
    explicit Approx(const T& value,
                    typename detail::enable_if<std::is_constructible<double, T>::value>::type* =
                            static_cast<T*>(nullptr)) {
        *this = Approx(static_cast<double>(value));
    }
#endif // DOCTEST_CONFIG_INCLUDE_TYPE_TRAITS
```

But you're right, this causes unnecessary conversions back to `float`:
```cpp
bool operator==(const sf::Angle& lhs, const ApproxDeg& rhs)
{
    return lhs.asDegrees() == doctest::Approx(rhs.degrees);
}
```

So it makes sense to retain a `float` instead of a `double`, and let `Approx` deal with conversions.

---


What I would _not_ do however, is make the constructor calls more syntax-verbose (by adding `{}` and `sf::degrees()`). The entire point of those structs is readability in tests.

ChrisThrasher:
> What I would not do however, is make the constructor calls more syntax-verbose (by adding {} and sf::degrees()). The entire point of those structs is readability in tests.

My argument is that the extra verbosity does make them more readable, but I won't die on that hill so say the word and I'll remove the latter two commits.

I think a reasonable end goal is to change our utilities such the tests never include `ApproxVec` or `ApproxDeg` or anything else like that and instead write overloads that let us simply type `Approx` around any type we want approximately compared. That will make approximate comparison and exact comparison look very similar in tests.

```cpp
CHECK(getVec1() == sf::Vector(1, 2));
CHECK(getVec2() == Approx(sf::Vector(1, 2));
```

Bromeon:
> My argument is that the extra verbosity does make them more readable, but I won't die on that hill so say the word and I'll remove the latter two commits.

For the angle one, we could probably still use `ApproxAngle(123.54_deg)` instead of `ApproxDeg(123.54)` or `ApproxAngle(sf::degrees(123.54))`.



> I think a reasonable end goal is to change our utilities such the tests never include `ApproxVec` or `ApproxDeg` or anything else like that and instead write overloads that let us simply type `Approx` around any type we want approximately compared. That will make approximate comparison and exact comparison look very similar in tests.

That sounds good! Back when I wrote some tests, I was already wondering how we could achieve genericity with the existing restricted `Approx()` -- it would likely need us to make our own adapter.

ChrisThrasher:
> That sounds good! Back when I wrote some tests, I was already wondering how we could achieve genericity with the existing restricted Approx() -- it would likely need us to make our own adapter.

How about this: We scrap the 2nd and 3rd commit and make this PR as simple as possible to unblock the other PRs ASAP. Then I come back when I have a good abstraction that lets us use `Approx` with any type. Does that sound good?

Bromeon:
Great idea! 👍

ChrisThrasher:
https://github.com/ChrisThrasher/SFML/commit/0cc4ee4bfc211f162e3647b82b13261fcffeb336

Here's my shot at writing a better abstraction for approximate comparisons. Thanks to @kimci86 for their idea for how to implement this.

eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:
Wait, so is that other branch an additional change, or how am I to read this? 😅

ChrisThrasher:
> Wait, so is that other branch an additional change, or how am I to read this? 

It's a potential follow-on to this PR.

eXpl0it3r:
Thanks for the fix 🙂

PR #2090, opened 2022-05-03
try_emplace in Font::loadPage

CosminPerRam:
## Description

Just a simple `try_emplace` change in `Font::loadPage`

This PR is related to the issue #2072

### DIFF
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -534,12 +534,7 @@ void Font::cleanup()
 ////////////////////////////////////////////////////////////
 Font::Page& Font::loadPage(unsigned int characterSize) const
 {
-    // TODO: Remove this method and use try_emplace instead when updating to C++17
-    PageTable::iterator pageIterator = m_pages.find(characterSize);
-    if (pageIterator == m_pages.end())
-        pageIterator = m_pages.insert(std::make_pair(characterSize, Page(m_isSmooth))).first;
-
-    return pageIterator->second;
+    return m_pages.try_emplace(characterSize, Page(m_isSmooth)).first->second;
 }
 
 
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -534,7 +534,7 @@ void Font::cleanup()
 ////////////////////////////////////////////////////////////
 Font::Page& Font::loadPage(unsigned int characterSize) const
 {
-    return m_pages.try_emplace(characterSize, Page(m_isSmooth)).first->second;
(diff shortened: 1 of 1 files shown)


eXpl0it3r:

eXpl0it3r:
@kimci86 if you want to check with your initial suggestions 🙂

kimci86:
I suggest removing the method `Font::loadPage` and call `try_emplace` instead wherever `Font::loadPage` was used.

src/SFML/Graphics/Font.cpp
```diff
@@ -534,12 +534,7 @@ void Font::cleanup()
 ////////////////////////////////////////////////////////////
 Font::Page& Font::loadPage(unsigned int characterSize) const
 {
-    // TODO: Remove this method and use try_emplace instead when updating to C++17
-    PageTable::iterator pageIterator = m_pages.find(characterSize);
-    if (pageIterator == m_pages.end())
-        pageIterator = m_pages.insert(std::make_pair(characterSize, Page(m_isSmooth))).first;
-
-    return pageIterator->second;
+    return m_pages.try_emplace(characterSize, Page(m_isSmooth)).first->second;
```
kimci86, line 537: Passing the boolean argument for Page constructor instead of a Page instance avoids unnecessary instantiation and copy.
```suggestion
    return m_pages.try_emplace(characterSize, m_isSmooth).first->second;
```

kimci86 (changes requested):
I suggest removing the method `Font::loadPage` and call `try_emplace` instead wherever `Font::loadPage` was used (at three places).
I think it makes sense as it becomes a function of one line. On the other hand, code might be less readable without the method.
Anyways, whether `Font::loadPage` gets removed or not, `try_emplace` should take the constructor's arguments and not and instance. See the suggested change below.

eXpl0it3r:
I wouldn't duplicate the code, even if it's just one line. Having the function name "loadPage" is much more explicit, than thinking through the logic of what `try_emplace` now does exactly.

eXpl0it3r:

kimci86:
I was thinking using `try_emplace` directly would be clear enough as it is from the standard library, but being more explicit doesn't hurt.
It looks good to me.

eXpl0it3r:
What I meant, that while it's clear what `try_emplace` itself does, the meaning behind the operation (i.e. loading a page), requires extra brain power, so having a dedicated function for it makes sense and being able to re-use it (DRY principle) justifies it even more.

eXpl0it3r (approved):

eXpl0it3r:
Thanks a lot for updating this to C++17 🥳

PR #2101, opened 2022-05-14
Fix compiler errors for clang++ & clang-cl on Windows

Bambo-Borris:

----

## Description

As highlighted in #2046 the 2.6.x & master branches both produce compile errors when using either clang++ or clang-cl. This PR attempts to fix some of these issues by doing the following: 

1. Supressing errors about deprecated Windows API functions (these require Windows SDK & Win8.1+ to resolve)
2. Supressing `#pragma GCC` directives which the clang-cl interface seems to dislike. 
3. Fixing the deprecated function suggested alternative.

Regarding 1: 
I chose the supression in places where the alternative required Windows SDK to use the alternative functions

Regarding 2: 
For some reason the directive below seemed to be throwing up a compilation error in the clang-cl interface. 
```cpp
#pragma GCC diagnostic ignored "-Wuseless-cast"
```
However in the CMake configuration for compiler warnings we explicitly instruct MSVC to ignore these GCC pragma directives. This compiler flag is passed to the clang-cl `/wd4068` but the clang-cl doesn't seem to pay attention to it. The error generated if this isn't supressed is:
```
error : unknown warning group '-Wuseless-cast', ignored [-Werror,-Wunknown-warning-option]
#pragma GCC diagnostic ignored "-Wuseless-cast"
```

Regarding 3: 
Where possible to use the suggested alternative to a deprecated WinAPI function I have.

This PR is related to the issue #2046 

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Compile this branch with `clang-cl` or `clang++` on Windows and if it compiles with no issues the errors are fixed...


ChrisThrasher:
Should this PR target https://github.com/SFML/SFML/tree/feature/vs2022-clang or at least include its commits so that we can get continuous CI confirmation that this is working?

eXpl0it3r:
Yeah, can you include the CI changes, so we see the fixes to the errors working?

Bambo-Borris:
Mistakenly rebased to wrong branch whilst messing with other stuff, remade this PR with #2114 and also a follow up PR to come for the clang++ interface on Windows. I think they're better off split anyway as I still have somethings to figure out about the second one.

PR #2103, opened 2022-05-17
Simplify approximate comparisons of SFML types

ChrisThrasher:
## Description

With the help of @kimci86 I refactored how we handle approximate comparisons with the goal of making our test utilities more generic which reduces code size and more clean by removing the need for so many `static_cast`s. Along the way I added some extra utilities for approximate comparisons of `sf::Transform` since those tests were pretty verbose. 

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

### DIFF
--- a/test/Graphics/Transform.cpp
+++ b/test/Graphics/Transform.cpp
@@ -6,8 +6,6 @@
 
 #include <doctest.h>
 
-using doctest::Approx;
-
 namespace std
 {
 std::ostream& operator<<(std::ostream& out, const std::vector<float>& vector)
@@ -125,30 +123,18 @@ TEST_CASE("sf::Transform class - [graphics]")
         {
             sf::Transform transform;
             transform.rotate(sf::degrees(90));
-            CHECK(transform.getMatrix()[0] == Approx(0));
-            CHECK(transform.getMatrix()[4] == Approx(-1));
-            CHECK(transform.getMatrix()[12] == Approx(0));
-            CHECK(transform.getMatrix()[1] == Approx(1));
-            CHECK(transform.getMatrix()[5] == Approx(0));
-            CHECK(transform.getMatrix()[13] == Approx(0));
-            CHECK(transform.getMatrix()[3] == Approx(0));
-            CHECK(transform.getMatrix()[7] == Approx(0));
-            CHECK(transform.getMatrix()[15] == Approx(1));
+            CHECK(transform == Approx(sf::Transform(0, -1,  0,
+                                                    1,  0,  0,
+                                                    0,  0,  1)));
         }
 
         SUBCASE("Around custom point")
         {
             sf::Transform transform;
             transform.rotate(sf::degrees(90), {1.0f, 0.0f});
-            CHECK(transform.getMatrix()[0] == Approx(0));
-            CHECK(transform.getMatrix()[4] == Approx(-1));
-            CHECK(transform.getMatrix()[12] == Approx(1));
-            CHECK(transform.getMatrix()[1] == Approx(1));
-            CHECK(transform.getMatrix()[5] == Approx(0));
-            CHECK(transform.getMatrix()[13] == Approx(-1));
-            CHECK(transform.getMatrix()[3] == Approx(0));
-            CHECK(transform.getMatrix()[7] == Approx(0));
-            CHECK(transform.getMatrix()[15] == Approx(1));
+            CHECK(transform == Approx(sf::Transform(0, -1,  1,
+                                                    1,  0, -1,
+                                                    0,  0,  1)));
         }
     }
 
--- a/test/System/Time.cpp
+++ b/test/System/Time.cpp
@@ -3,8 +3,6 @@
 
 #include <doctest.h>
 
-using doctest::Approx;
-
 TEST_CASE("sf::Time class - [system]")
 {
     SUBCASE("Construction")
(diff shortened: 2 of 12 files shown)


test/TestUtilities/SystemUtil.hpp
```diff
@@ -41,45 +41,23 @@ namespace sf
...
+struct Approx
 {
-    ApproxVec2(float x, float y)
-        : vector(x, y) {}
-
-    explicit ApproxVec2(const sf::Vector2f& v)
-        : vector(v) {}
-
-    sf::Vector2f vector;
+    Approx(const T& t) : value(t) {}
```
Bromeon, line 47: Should this be explicit, or do we have places where we don't use the `Approx(expectedValue)` syntax?

test/System/Time.cpp
```diff
@@ -163,7 +161,7 @@ TEST_CASE("sf::Time class - [system]")
             CHECK(sf::seconds(1) / static_cast<sf::Int64>(2) == sf::seconds(0.5f));
             CHECK(sf::seconds(42) / static_cast<sf::Int64>(2) == sf::seconds(21));
             CHECK(sf::seconds(1) / sf::seconds(1) == 1.0f);
-            CHECK(sf::milliseconds(10) / sf::microseconds(1) == Approx(10'000.0).epsilon(1e-6));
+            CHECK(sf::milliseconds(10) / sf::microseconds(1) == doctest::Approx(10'000.0).epsilon(1e-6));
```
Bromeon, line 164: Do you think it would make sense to add `Approx<float>::epsilon()`? E.g. as specialization, or even as method in the general template, which is simply not invoked for non-floats...

We are likely going to use this here and there...

test/System/Vector2.cpp
```diff
@@ -244,76 +243,76 @@ TEST_CASE("sf::Vector2 class template - [system]")
...
     SUBCASE("Rotations and angles")
     {
         const sf::Vector2f v(2.4f, 3.0f);
         
-        CHECK(v.angle() == ApproxDeg(51.3402f));
-        CHECK(sf::Vector2f::UnitX.angleTo(v) == ApproxDeg(51.3402f));
-        CHECK(sf::Vector2f::UnitY.angleTo(v) == ApproxDeg(-38.6598f));
+        CHECK(v.angle() == Approx(sf::degrees(51.3402f)));
+        CHECK(sf::Vector2f::UnitX.angleTo(v) == Approx(sf::degrees(51.3402f)));
+        CHECK(sf::Vector2f::UnitY.angleTo(v) == Approx(sf::degrees(-38.6598f)));
```
Bromeon, line 263: What do you think of using the `_deg` notation here (and in similar places)?

Bromeon (commented):
Very very nice! This is wonderfully generic :sli

test/TestUtilities/SystemUtil.hpp
ChrisThrasher: Yeah you're probably right. I can't think of a good reason you'd want to except an argument which wasn't precise the same same type as `T`.

test/System/Time.cpp
ChrisThrasher: Eventually we might. With enough time I could eliminate the need for `.epsilon` here. I'm using it probably because of the 32bit MinGW build using especially low precision floats. If I change the values we're testing then I can eventually produce a result that is close enough to just use `Approx(10'000.f)`.

test/System/Vector2.cpp
ChrisThrasher: Sounds good to me. I'll use `_deg` pervasively throughout that file since it already has `using namespace sf::Literals` at the top.

test/Graphics/Transform.cpp
```diff
@@ -125,30 +123,14 @@ TEST_CASE("sf::Transform class - [graphics]")
...
-            CHECK(transform.getMatrix()[0] == Approx(0));
-            CHECK(transform.getMatrix()[4] == Approx(-1));
-            CHECK(transform.getMatrix()[12] == Approx(0));
-            CHECK(transform.getMatrix()[1] == Approx(1));
-            CHECK(transform.getMatrix()[5] == Approx(0));
-            CHECK(transform.getMatrix()[13] == Approx(0));
-            CHECK(transform.getMatrix()[3] == Approx(0));
-            CHECK(transform.getMatrix()[7] == Approx(0));
-            CHECK(transform.getMatrix()[15] == Approx(1));
+            CHECK(transform == Approx(sf::Transform(0, -1, 0, 1, 0, 0, 0, 0, 1)));
```
kimci86, line 126: What do you think about formatting this to make the 3x3 matrix visible?
```suggestion
            CHECK(transform == Approx(sf::Transform(0, -1, 0,
                                                    1,  0, 0,
                                                    0,  0, 1)));
```

ChrisThrasher: Sure I can do that.

kimci86:
Thank you for carrying out this refactor. I would say all that is missing is a small bit of documentation to explain the design and how it can be extended to other types.

ChrisThrasher:
> I would say all that is missing is a small bit of documentation to explain the design and how it can be extended to other types.

Would a comment block in SystemUtils.hpp suffice? Something like

```cpp
////////////////////////////////////////////////////////
/// Class template for creating custom approximate comparisons.
/// To register a new type, simply implement a custom operator==
/// overload for that type. 
////////////////////////////////////////////////////////
template <typename T>
struct Approx
{
    explicit Approx(const T& t) : value(t) {}
    const T& value;
};
```

ChrisThrasher:
Added some simple docs for `Approx<T>` to clarify its purpose and how to use it.

eXpl0it3r:

ChrisThrasher:
https://github.com/ChrisThrasher/SFML/tree/test_graphics

I have some more Graphics module unit tests lined up that help further reinforce the importance of simpler approximate comparisons of `sf::Transform`.

ChrisThrasher:
@Bromeon Anything else you'd like to see change about this?

Bromeon (approved):
I think it's good! The `doctest::Approx(10'000.0).epsilon(1e-6)` can be made "native" later, nothing urgent 🙂

PR #2105, opened 2022-05-18
Assorted Small Build Simplifications

ChrisThrasher:
## Description

The commits are mostly self explanatory but in short, I remove unnecessarily setting two CMake policies and modernize how we hide symbols to be more cross platform.

The CMake docs linked in the extended message explain how these two policies are no longer necessary as long as your minimum CMake version is greater than 3.0 and 3.9 respectively. Because we now require 3.15 we can rely on the default value of those policies instead of having to explicitly specify new behavior. The new way to enable visibility flags can be validated by configuring with `-DCMAKE_EXPORT_COMPILE_COMMANDS` and see how all library targets continue to get built with `-fvisibility=hidden` and `-fvisibility-inlines-hidden`.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,10 +13,6 @@ endmacro()
 # determine whether to create a debug or release build
 sfml_set_option(CMAKE_BUILD_TYPE Release STRING "Choose the type of build (Debug or Release)")
 
-if (NOT CMAKE_VERSION VERSION_LESS 3.9)
-    cmake_policy(SET CMP0068 NEW)
-endif()
-
 if(NOT CMAKE_OSX_SYSROOT)
     # query the path to the default SDK, will fail on non-macOS, but it's okay.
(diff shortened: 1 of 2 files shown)


ChrisThrasher:
https://github.com/SFML/SFML/blob/1e6dc823043b25cdec77f5e1e52cad0b3e35e0a6/cmake/CompilerWarnings.cmake#L90-L97

I also wanted to remove this GCC 8 workaround but I can't get my hands on an old enough version of GCC to test this. CI passes but it's using GCC 9.4 so of course it does. We'd need @binary1248 to help test if this is still necessary.

eXpl0it3r (approved):

eXpl0it3r:

Issue #2100, opened 2022-05-13
2.6.x Build fails with GCC 12.1.0 (Arch Linux)


rewrking:
## 2.6.x Build fails with GCC 12.1.0 (Arch Linux)

Building SFML from source (2.6.x branch) fails with GCC 12.1.0.

## Environment

* Arch Linux, upgraded with `sudo pacman -Syu`
* 2.6.x branch
* GCC 12.1.0

## Steps to reproduce

1. Upgrade latest packages with `sudo pacman -Syu`
2. Install `ninja` and `cmake` if necessary
3. Build with:

```bash
git clone -b 2.6.x https://github.com/SFML/SFML.git
cd SFML
mkdir build
cmake -G Ninja -DCMAKE_C_COMPILER=/usr/bin/gcc -DCMAKE_CXX_COMPILER=/usr/bin/g++ -DCMAKE_BUILD_TYPE=Release -S . -B build

# Will fail, but keep going until any possible files are built
cmake --build build -j 8 -- -v -k 0 

# Run a second time to see the error in question
cmake --build build -j 8 -- -v -k 0 
```

## Expected behavior

It should build.

## Actual behavior

It did not build. :(

It fails on src/Network/SFML/Http.cpp:264 with:

```bash
/usr/bin/g++ -DSFML_NETWORK_EXPORTS -I/home/username/dev/tmp/SFML/include -I/home/username/dev/tmp/SFML/src -Wno-error -O3 -DNDEBUG -fPIC -fvisibility=hidden  -Wall -Wextra -Wshadow -Wnon-virtual-dtor -Wcast-align -Wunused -Woverloaded-virtual -Wconversion -Wsign-conversion -Wdouble-promotion -Wformat=2 -Wnull-dereference -Wold-style-cast -Wpedantic -Werror -Wmisleading-indentation -Wduplicated-cond -Wlogical-op -Wduplicated-branches -MD -MT src/SFML/Network/CMakeFiles/sfml-network.dir/Http.cpp.o -MF src/SFML/Network/CMakeFiles/sfml-network.dir/Http.cpp.o.d -o src/SFML/Network/CMakeFiles/sfml-network.dir/Http.cpp.o -c /home/username/dev/tmp/SFML/src/SFML/Network/Http.cpp
In file included from /usr/include/c++/12.1.0/ios:43,
                 from /usr/include/c++/12.1.0/istream:38,
[...]

PR #2106, opened 2022-05-22
bugfix for #2100: null-dereference in GCC 12.1.0

rewrking:
## bugfix for #2100: null-dereference in GCC 12.1.0

This PR is related to the issue #2100

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

See issue #2100

### DIFF
--- a/src/SFML/Network/Http.cpp
+++ b/src/SFML/Network/Http.cpp
@@ -261,7 +261,10 @@ void Http::Response::parse(const std::string& data)
             std::istreambuf_iterator<char> it(in);
             std::istreambuf_iterator<char> itEnd;
             for (std::size_t i = 0; ((i < length) && (it != itEnd)); i++)
-                m_body.push_back(*it++);
+            {
+                m_body.push_back(*it);
+                ++it;
+            }
         }
 
         // Drop the rest of the line (chunk-extension)


ChrisThrasher:
Did you ever figure out if GCC is truly detecting a bug or is this just a way to get that warning to stop emitting false positives?

rewrking:
It definitely looks to be a false positive fixed in GCC trunk:
https://godbolt.org/z/74TPf7ez3

This commit alleviates the warning & error in 12.1.

src/SFML/Network/Http.cpp
```diff
@@ -261,7 +261,10 @@ void Http::Response::parse(const std::string& data)
             std::istreambuf_iterator<char> it(in);
             std::istreambuf_iterator<char> itEnd;
             for (std::size_t i = 0; ((i < length) && (it != itEnd)); i++)
-                m_body.push_back(*it++);
+            {
+                m_body.push_back(*it);
+                ++it;
```
ChrisThrasher, line 266: Thanks for the awesome Godbolt link! Because subsequent readers may not pick up on why we're doing this, can you add a quick comment mentioning how this is a GCC 12.1 workaround? Maybe something like this?

```suggestion
                ++it; // Iterate in separate expression to work around false positive -Wnull-dereference warning in GCC 12.1.0
```

ChrisThrasher (approved):

src/SFML/Network/Http.cpp
rewrking: Not a problem! Added.

ChrisThrasher:
If you wouldn't mind, could you confirm that this bug exists in the `master` branch too?

rewrking:
> 

Yes. This bug exists in `master` as well.

ChrisThrasher:
Thank you for checking! We periodically merge `2.6.x` into `master` so both branches will be fixed in time.

@eXpl0it3r What do you think?

eXpl0it3r:

eXpl0it3r:
Thanks for this fix! 🙂

PR #2107, opened 2022-05-22
Update Text.cpp

DanielRabl:
inconsistent comparison to '\r', usually comparing against 16bit wide characters in this file, but here the L is missing.


src/SFML/Graphics/Text.cpp
```diff
@@ -453,7 +453,7 @@ void Text::ensureGeometryUpdate() const
         Uint32 curChar = m_string[i];
 
         // Skip the \r char to avoid weird graphical issues
-        if (curChar == '\r')
```
Baardi, line 456: Probably stupid question, but wouldn't uppercase U be more correct here? Considering it's a 32bit character type

DanielRabl: @Baardi you could be right, I don't know much about utf32. I would guess it's not needed, as the only characters in question are special characters ('\n', '\t', '\r' etc) and their bit structure (probably) won't change from 16 to 32 bits.

ChrisThrasher:
Is this causing a bug that you can reproduce? Do you think we need to change the other non-wide characters literals in that same file?

eXpl0it3r:
Could you rebase your commit onto the `2.6.x` branch and change the PR target to 2.6.x?

src/SFML/Graphics/Text.cpp
eXpl0it3r: We can't in SFML 2.6.x

eXpl0it3r (approved):

Bromeon:
I don't really see the point of this change -- as @Baardi mentioned, we would use `U` prefix (`std::char32_t`, see [here](https://docs.microsoft.com/en-us/cpp/cpp/string-and-character-literals-cpp)). Now we can't do this because of C++98, OK. But how is `L` a compromise here?

The `L` prefix (`wchar_t`) is not more correct than no prefix (`char`). In both cases, we are converting to `sf::Uint32`.

However, the `L` prefix explicitly suggests that we're dealing with `wchar_t` (wide character set), which is mostly used for legacy APIs (such as Windows API). And I think this creates more confusion than it solves.

eXpl0it3r:
Leaving things in an unaligned state seems to me less desirable.
It's currently already using `L"\..."` for all the other characters, why should we leave this one unchanged?

Bromeon:
>  It's currently already using L"\..." for all the other characters

Yes, but that doesn't make it more correct.

The main reason I see for keeping `L` everywhere is that it will be easier to search&replace `L'` with `U'` in SFML 3 🙂

DanielRabl:
I will gladly do the pull request! You can find it here: https://github.com/SFML/SFML/pull/2121

PR #2114, opened 2022-06-04
Disable warnings as errors for clang-cl builds

Bambo-Borris:

----

## Description

As highlighted in https://github.com/SFML/SFML/issues/2046 the 2.6.x & master branches both produce compile errors when using either clang++ or clang-cl. This PR supresses some warnings which are not supported by clang-cl, and also disables `/WX` due to an issue where dependencies from `extlib/` were being compiled with `/WX` when they shouldn't have been. This would cause the CI to fail due to `/WX`, so this option has been temporarily disabled for Windows + clang-cl builds. 

For CI logs showing the issue with the clang-cl + `/WX` see [here](https://github.com/SFML/SFML/actions/runs/2327878406)

This does **not** resolve the compiler errors that originate from using clang++ on Windows (the GNU-like interface). Though I intend to add a follow up PR which will fix the clang++ build and add the necessary CI to confirm it's working. 

## Tasks

* [ ] Tested on Linux
* [X] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Compile with clang-cl on Windows.

### DIFF
--- a/cmake/CompilerWarnings.cmake
+++ b/cmake/CompilerWarnings.cmake
@@ -68,7 +68,13 @@ function(set_file_warnings)
         # -Wimplicit-fallthrough # warn when a missing break causes control flow to continue at the next case in a switch statement (disabled until better compiler support for explicit fallthrough is available)
         ${NON_ANDROID_CLANG_AND_GCC_WARNINGS}
     )
-
+    
+    # For now if we're using MSVC-like clang interface on Windows
+    # we'll disable warnings as errors 
+    if(SFML_OS_WINDOWS AND SFML_COMPILER_CLANG_CL)
+        set(WARNINGS_AS_ERRORS FALSE)
+    endif()
+    
     if(WARNINGS_AS_ERRORS)
         set(CLANG_AND_GCC_WARNINGS ${CLANG_AND_GCC_WARNINGS} -Werror)
         set(MSVC_WARNINGS ${MSVC_WARNINGS} /WX)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,7 @@ jobs:
         platform:
         - { name: Windows VS2019, os: windows-2019 }
         - { name: Windows VS2022, os: windows-2022 }
+        - { name: Windows VS2022 Clang, os: windows-2022, flags: -T ClangCL }
         - { name: Linux GCC,      os: ubuntu-latest  }
         - { name: Linux Clang,    os: ubuntu-latest, flags: -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ }
         - { name: MacOS XCode,    os: macos-latest   }
--- a/cmake/Config.cmake
+++ b/cmake/Config.cmake
@@ -107,6 +107,11 @@ if(MSVC)
     elseif(MSVC_VERSION LESS_EQUAL 1939)
         set(SFML_MSVC_VERSION 17)
     endif()
+	
+	if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+		set(SFML_COMPILER_CLANG_CL 1)
+	endif()
+	
 elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     set(SFML_COMPILER_CLANG 1)
 


cmake/CompilerWarnings.cmake
```diff
@@ -68,7 +68,13 @@ function(set_file_warnings)
         # -Wimplicit-fallthrough # warn when a missing break causes control flow to continue at the next case in a switch statement (disabled until better compiler support for explicit fallthrough is available)
         ${NON_ANDROID_CLANG_AND_GCC_WARNINGS}
     )
-
+    
+    # For now if we're using MSVC-like clang interface on Windows
+    # we'll disable warnings as errors 
+    if((SFML_OS_WINDOWS AND SFML_COMPILER_CLANG_CL))
```
ChrisThrasher, line 74:
```suggestion
    if(SFML_OS_WINDOWS AND SFML_COMPILER_CLANG_CL)
```
You've got extraneous parentheses here.

ChrisThrasher (commented):
After disabling warnings-as-errors for ClangCL, do we still need all the preprocessor changes you made in other files or can that wait for later once we finally fix the compiler warnings we're currently working around?

src/SFML/Network/TcpSocket.cpp
```diff
@@ -245,12 +245,15 @@ Socket::Status TcpSocket::send(const void* data, std::size_t size, std::size_t&
     int result = 0;
     for (sent = 0; sent < size; sent += static_cast<std::size_t>(result))
     {
+#if !(defined(_MSC_VER) && defined(__clang__))
         #pragma GCC diagnostic push
         #pragma GCC diagnostic ignored "-Wuseless-cast"
+#endif
```
Bromeon, line 251: We should decide if we keep preprocessor directives indented at the beginning of the line, or make it dependent on scope -- but not mix the two styles.

ChrisThrasher: I'm inclined to put all preprocessor code on the left margin since the preprocessor has no concept of scope nor do scopes have any concept of the preprocessor. I'd be willing to make a PR unifying preprocessor style since the library is already using a mix of styles.

Bambo-Borris:
> After disabling warnings-as-errors for ClangCL, do we still need all the preprocessor changes you made in other files or can that wait for later once we finally fix the compiler warnings we're currently working around?

If we remove the preprocessor changes the build will still succeed, but it will leave the known errors in place. It possible for these to be pushed in a separate commit if that's what is favoured overall. As mentioned in the description there is to be a follow up PR for this which is for the GNU-like interface to build. So maybe after the GNU-like interface these macros can go in on a third commit.

src/SFML/Network/TcpSocket.cpp
Bambo-Borris: Given what @ChrisThrasher said with these likely to go in on a future commit then I can make sure they all follow a consistent indentation level wherever I re-add them. I'll revert them from this commit for now.

ChrisThrasher (approved):
Great work!

Bambo-Borris:
Important to note, I did go through the CI logs to verify `/WX`/`-Werror` were present on the relevant builds that should retain them. It was unlikely that they would be effected, but good to be certain.

Bambo-Borris:
@Bromeon I removed the preprocessor stuff so that this commit is just the changes required to enable the ClangCL to build. It's ready for a review again whenever you're able to

eXpl0it3r:

vittorioromeo:
This only affects 2.6.x, right? Will this backmerged in 3.x? If so, will it prevent any useful warning-as-error from being generated?

Bambo-Borris:
> This only affects 2.6.x, right? Will this backmerged in 3.x? If so, will it prevent any useful warning-as-error from being generated?

It's targeted at 2.6.x but is applicable to 3.x but I don't know if it will be backmerged. It may prevent us spotting some useful warnings-as-errors, but it's a temporary solution to allow ClangCl to actually build 2.6.x. Currently 2.6.x won't build via ClangCl due to the issue highlighted in the description. There was discussion of this in Discord which provides additional information for those curious. I haven't tested building 3.x with ClangCl for a whil3 so it's something I'm open to verifying if there's a problem there too.

Bambo-Borris:
Is there any additional concerns about this PR or is it good to merge?

eXpl0it3r (approved):

eXpl0it3r:
Another 🎉 for the Big Blue Bambooooo 🙂

eXpl0it3r:
> This only affects 2.6.x, right? Will this backmerged in 3.x? If so, will it prevent any useful warning-as-error from being generated?

It will be back merged to master.

The issue right now is that SFML doesn't compile, neither on 2.6.x nor on master with MSVC + Clang-Cl.
So in order to not block people for now, the treat warnings as errors is being disable and it gives us time to fix them either on 2.6.x or master.

PR #2117, opened 2022-06-06
Add tests for sf::Image

Bambo-Borris:

----

## Description

Currently there are no tests for the `sf::Image` class, this PR will add tests for the following functions 
- Default constructor
- `create()`
- `getPixel()`/`setPixel()`
- `copy()`
- `flipHorizontally()`
- `flipVertically()`

I have plans to come back and add tests for the `copy(Image, Vector2u, IntRect, bool)`. And also for the `createMaskFromColor()` but I want to understand them both properly before I do that. I'd prefer to at least get these tests in now as there's presently no existing tests so this is gives us a base to work from. 
## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Run the tests! 😄


test/Graphics/Image.cpp
```diff
@@ -0,0 +1,151 @@
...
+            CHECK(image2.getSize() == sf::Vector2u(10, 10));
+            CHECK(image2.getPixelsPtr() != nullptr);
+
+            for (sf::Uint32 i = 0; i < 10; ++i)
+            {
+                for (sf::Uint32 j = 0; j < 10; ++j)
+                {
+                    CHECK(image2.getPixel(sf::Vector2u(i, j)) == sf::Color(0, 0, 0));
+                }
+            }
```
ChrisThrasher, line 48: Can these be separate subcases since they appear to be independent and testing different things?

test/Graphics/Image.cpp
```diff
@@ -0,0 +1,151 @@
...
+{
+    SUBCASE("Construction")
+    {
+        SUBCASE("Default constructor")
+        {
+            const sf::Image image;
+            CHECK(image.getSize() == sf::Vector2u());
+            CHECK(image.getPixelsPtr() == nullptr);
+        }
+    }
```
ChrisThrasher, line 17: You don't need this redundant scope since there is only one constructor to test.

test/Graphics/Image.cpp
```diff
@@ -0,0 +1,151 @@
...
+                {
+                    CHECK(image2.getPixel(sf::Vector2u(i, j)) == sf::Color(0, 0, 0));
+                }
+            }
+        }
+
+        SUBCASE("create(Vector2, Uint8*)")
+        {
+            // 10 x 10, with 4 colour channels array
+            std::array<sf::Uint8, 400> pixels;
```
ChrisThrasher, line 54: Would it make more sense if this was a `std::array<sf::Color, 100>`? I believe that will still yield the exactly same data layout but with clearer semantics. Maybe even use an array of arrays to further improve semantics by making it abundantly clear that we're talking about a 10 by 10 array of colors.

test/Graphics/Image.cpp
```diff
@@ -0,0 +1,151 @@
...
+            for (std::size_t i = 0; i < pixels.size(); i += 4)
+            {
+                pixels[i] = 255;     // r
+                pixels[i + 1] = 0;   // g
+                pixels[i + 2] = 0;   // b
+                pixels[i + 3] = 255; // a
+            }
+
+            sf::Image image;
+            image.create(sf::Vector2u{10, 10}, pixels.data());
```
ChrisThrasher, line 64: Use parentheses to initialize this vector to better follow SFML convention.

test/Graphics/Image.cpp
```diff
@@ -0,0 +1,151 @@
...
+        {
+            // 10 x 10, with 4 colour channels array
+            std::array<sf::Uint8, 400> pixels;
+            for (std::size_t i = 0; i < pixels.size(); i += 4)
+            {
+                pixels[i] = 255;     // r
+                pixels[i + 1] = 0;   // g
+                pixels[i + 2] = 0;   // b
+                pixels[i + 3] = 255; // a
+            }
```
ChrisThrasher, line 61: If we use an array of colors, we can easy initialize all elements with a standard algorithm instead of this loop.

test/Graphics/Image.cpp
Bambo-Borris: We need the `sf::Uint8` array here because that's one of the arguments for the create function it's testing

ChrisThrasher:
```cpp
// 10 x 10, with 4 colour channels array
std::array<sf::Color, 100> pixels;
std::fill(pixels.begin(), pixels.end(), sf::Color(255, 0, 0, 255));

sf::Image image;
image.create(sf::Vector2u{10, 10}, &pixels.front().r);
```

eXpl0it3r:

eXpl0it3r:
Looking at the coverage report it seems there some cases that could easily be covered as well:

- Negative case for create() function: https://app.codecov.io/gh/SFML/SFML/compare/2117/changes#D1L110
- Copy & Move constructor/assignment: https://app.codecov.io/gh/SFML/SFML/compare/2117/changes#D1L57
- Pass an image with size (0, 0) to the copy function (either source or destination): https://app.codecov.io/gh/SFML/SFML/compare/2117/changes#D1L208
- Pass a bad sourceRect to the copy function: https://app.codecov.io/gh/SFML/SFML/compare/2117/changes#D1L235
- Test case for `applyAlpha = true` with the copy function: https://app.codecov.io/gh/SFML/SFML/compare/2117/changes#D1L250
- Test case for createMaskFromColor: https://app.codecov.io/gh/SFML/SFML/compare/2117/changes#D1L186

ChrisThrasher:
For later tests that end up touching classes with non-trivial destructors, I do think we should start explicitly testing copy and move semantics to be sure that they all work since supporting move semantics is a significant SFML 3 feature. For `sf::Image` with it's soon-to-be-removed destructor, I don't think we should bother. Once SFML perfectly follows the rule of 0/5 then it'll be much more apparent what needs testing and what does not.

test/Graphics/Image.cpp
ChrisThrasher: 😢 Dang that's too bad. I thought it was a safe operation. Well that explains why `vector<Color>` isn't used in the implementation either.

ChrisThrasher: https://godbolt.org/z/1rG7xE7rE

Here's a Godbolt example of this problem if anyone wants to play around with it. It seems to work and I can't get any tools to emit diagnostics on this code, but that alone isn't sufficient evidence to ship code like this.

ChrisThrasher: I built an example program with AddressSanitizer and UndefinedSanitizer (macOS Monterey), neither of which reported any errors. I added unit tests that do this indexing and [all CI jobs passed](https://github.com/ChrisThrasher/SFML/actions/runs/2456576883). I'm running out of ways to demonstrate a failure.

Bambo-Borris: Are we okay with this change going into the tests, or should it be something that is updated after some `static_assert` is put into the `sf::Color` code?

Bromeon: > I'm running out of ways to demonstrate a failure.

I think it was never about demonstrating, I see this more as a theoretical exercise. As mentioned, gamedev libraries have done stuff like this for decades. It works in practice (mostly 😉).

ChrisThrasher: No don't bother. I want to think on this construct more before we start using it.

vittorioromeo: > _Disclaimer: my answer assumes that we want to do things_ correctly _and not apply the Yolo approach which the gamedev industry has been following for decades 😬_
> 
> What I found:
> 
> * https://stackoverflow.com/a/39432189
> * https://stackoverflow.com/a/44338825
> 
> From the second answer, I read that it is _explicitly defined behavior_ to reinterpret a type as `char`, `unsigned char` or `std::byte` (interestingly, they don't mention `signed char`). `sf::Uint8` is `unsigned char`.
> 
> So this might actually work. We should however do a `static_assert(sizeof(sf::Color) == 4)` (and maybe `alignof`) to prevent an exotic compiler from doing anything weird. It's extremely unlikely that `sf::Color` is padded, but IIRC not forbidden. What _is_ forbidden in C++ is reordering struct fields.

Good observation. This happens to be valid only because `sf::Uint8` is a `char`, which is a bit subtle. It also helps that this has been tested on all our CI machines, so I guess I don't have a big problem with this approach if we indeed add a static assertion to verify the alignment/size guarantees.

vittorioromeo (approved):
Honestly happy with this as-is, we can always provide improvements as subsequent PRs.

Bambo-Borris:
> Honestly happy with this as-is, we can always provide improvements as subsequent PRs.

Haven't had the chance to push my subsequent changes yet but I did identify an out of bounds read occurring within `sf::Image`. I'll aim to push the other changes today.

Bambo-Borris:
I added the `createMaskFromColor` and missing `copy` tests, the test that I think reveals a hidden issue is the test for **Copy (Out of bounds sourceRect)**. It will fail on the initial push but I'm adding another commit with a 'temporary' fix, I don't think the fix is ideal but it will prevent the test from failing.

In that test I have an image of dimension 5x5, and I set the `sourceRect` parameter of the `copy()` to be `{5, 5, 9, 9}` which should be totally out of bounds for the `sourceRect`. What I assumed would happen is no copy would occur. Instead it appears that a copy still happens since the pixels of `image2` are changed despite the `sourceRect` being invalid from what I can tell. You can check the logs of the CI and you should see during the failure completely unexpected colour values in the test logs.

src/SFML/Graphics/Image.cpp
```diff
@@ -208,6 +208,10 @@ void Image::copy(const Image& source, const Vector2u& dest, const IntRect& sourc
     if ((source.m_size.x == 0) || (source.m_size.y == 0) || (m_size.x == 0) || (m_size.y == 0))
         return;
 
+    // Make sure the sourceRect is valid
+    if (sourceRect.left < 0 || static_cast<unsigned>(sourceRect.left) >= source.m_size.x || sourceRect.top < 0 || static_cast<unsigned>(sourceRect.top) >= source.m_size.y)
+        return;
+
```
vittorioromeo, line 214: Maybe it's worthwhile splitting this in a separate PR, and having `Image::copy` return a `[[nodiscard]] bool` so that users can actually check the copy happened. Thoughts? @eXpl0it3r

test/Graphics/Image.cpp
```diff
@@ -0,0 +1,248 @@
...
+        SUBCASE("Copy (Image, Vector2u, IntRect, bool)")
+        {
+            const sf::Color dest(255, 0, 0, 180);
+            const sf::Color source(0, 255, 0, 255);
+
+            // Create the composited colour for via the alpha composite over operation
+            const sf::Uint8 a = static_cast<sf::Uint8>(source.a + (dest.a * (255 - source.a)));
+            const sf::Uint8 r = static_cast<sf::Uint8>(((source.r * source.a) + ((dest.r * dest.a) * (255 - source.a))) / a);
+            const sf::Uint8 g = static_cast<sf::Uint8>(((source.g * source.a) + ((dest.g * dest.a) * (255 - source.a))) / a);
+            const sf::Uint8 b = static_cast<sf::Uint8>(((source.b * source.a) + ((dest.b * dest.a) * (255 - source.a))) / a);
```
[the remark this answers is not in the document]
Bambo-Borris: Yeah I totally agree, I'm more than happy to add this change in

ChrisThrasher:
This PR added tests, found a bug, reproduced the bug, and fixed the bug. That's a lovely, succinct 2-commit PR. Adding a `[[nodiscard]] bool` return value is a good idea now that we're more aware of the ways copying can fail, but that shouldn't cause us to change the scope of this PR. Let's get these tests through first.

src/SFML/Graphics/Image.cpp
Bambo-Borris: Maybe worth just having this PR with a commit of tests, and a commit of the fix go in as one. Then a future PR to revisit the `sf::Image::copy()` to mark it as `[[nodiscard]] bool`. At least that way we get the fix in sooner and the tests are also in to verify future changes to `sf::Image`

eXpl0it3r:

src/SFML/Graphics/Image.cpp
vittorioromeo: Let's extend the comment by briefly mentioning why we only have to check the top-left corner

vittorioromeo (approved):
LGTM

eXpl0it3r:
Another success story. Implementing tests, finding a bug and fixing it 🥳

vittorioromeo:
Yes, great job @Bambo-Borris!

src/SFML/Graphics/Image.cpp
```diff
@@ -177,6 +177,11 @@ void Image::copy(const Image& source, const Vector2u& dest, const IntRect& sourc
     if ((source.m_size.x == 0) || (source.m_size.y == 0) || (m_size.x == 0) || (m_size.y == 0))
         return;
 
+    // Make sure the sourceRect left & top  and the {left, top} + {width, height} within bounds
+    if (static_cast<unsigned int>(sourceRect.left) >= source.m_size.x || static_cast<unsigned int>(sourceRect.left + sourceRect.width) > source.m_size.x ||
+        static_cast<unsigned int>(sourceRect.top) >= source.m_size.y || static_cast<unsigned int>(sourceRect.top + sourceRect.height) > source.m_size.y)
+        return;
```
kimci86, line 183: I am late, but casting to unsigned int wraps negative values. I don't think that is wanted.
Are negative left and top values for the source rectangle allowed? It is documented not to be allowed since @vittorioromeo's PR #2137, but this is not what this condition checks.

vittorioromeo: I think I was referring to the check at L209

kimci86: If you are referring to this check, `width` and `height` are `unsigned int` at this line, so it is actually checking that both values are not zero.
https://github.com/SFML/SFML/blob/6cd07a043fc36ae75f8f094414d110a5f5333dbf/src/SFML/Graphics/Image.cpp#L208-L210

I was referring to this piece of documentation:

> this function can fail [...] if \a sourceRect is not within the boundaries of the \a source parameter

From this piece of documentation, I understand that `sourceRect` must be inside the rectangle (0, 0, width, height) where width and height are the size of the source image. Do we agree?
If so, I think this implementation must be improved. I will give it a try in the next few days if nobody does it before.

vittorioromeo: That sounds good, thanks.

PR #2119, opened 2022-06-07
Limit the scope of event object

ChrisThrasher:
## Description

Thanks to @eXpl0it3r for this neat little trick to limit the scope of the `sf::Event` object. I figured it was worth making this sweeping change to all event loops in the library so that anyone copy-pasting code gets slightly better code than before. I even changed the issue and PR templates.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -32,8 +32,7 @@ int main()
 
     while (window.isOpen())
     {
-        sf::Event event;
-        while (window.pollEvent(event))
+        for (sf::Event event; window.pollEvent(event);)
         {
             if (event.type == sf::Event::Closed)
                 window.close();
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -37,8 +37,7 @@ int main()
 
     while (window.isOpen())
     {
-        sf::Event event;
-        while (window.pollEvent(event))
+        for (sf::Event event; window.pollEvent(event);)
         {
             if (event.type == sf::Event::Closed)
                 window.close();
--- a/doc/mainpage.hpp
+++ b/doc/mainpage.hpp
@@ -44,8 +44,7 @@
 ///     while (window.isOpen())
 ///     {
 ///         // Process events
-///         sf::Event event;
-///         while (window.pollEvent(event))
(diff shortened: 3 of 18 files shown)


eXpl0it3r (approved):
I like the change, but I also want to hear the opinion from someone else 🙂

eXpl0it3r:

Bromeon (commented):
To be perfectly honest, I think this makes things a bit less readable, because:
* `while (window.pollEvent(event))` is pretty expressive -- "while the window polls events". The same cannot be said about the `for` variant.
* We now have only one line which does everything instead of two, and the `for` loop without an update part is quite unusual. People will likely wonder what exactly happens.

Compared to that, the limited scope might be nice, but it doesn't look to me like it solves a real problem that we had in the past. Generally, I would try to check trade-offs of common C++ best practices depending on their context.

eXpl0it3r:
> it doesn't look to me like it solves a real problem that we had in the past

What about all those issues people ran into when using the event object outside of the event loop?
That's not a one of thing and keeps happening quite regularly.
The source of the issue is usually also not very clear, as people report stuff like "moving my mouse, closes the window" or so, because the event object is in being checked in an invalid state.

Additionally it removes the discussions that happen less frequent of whether the event object should be a member variable, because "performance" (having to recreate the event object each iteration) etc.

> I think this makes things a bit less readable

I agree it's not as nicely matching the English phrasing, but in my opinion the issues it solves outweigh the slightly reduced readability.

vittorioromeo:
I like the change as well, and it is similar in spirit to C++17's introduction of `if(/* init */; /* predicate */)`, so people will likely understand it if they know about that C++17 addition.

Bromeon:
OK, seems like 3 people are in favor 🙂 

> I like the change as well, and it is similar in spirit to C++17's introduction of if(/* init */; /* predicate */), so people will likely understand it if they know about that C++17 addition.

Yep, I had to think of that as well. Whether it's more or less readable depends a bit of the complexity of both parts (and matter of taste/style of course). That made me wonder why they didn't also introduce `while (init; cond)`, that would fit perfectly here. Only found fairly non-convincing answers [here](https://stackoverflow.com/q/59985550).

PR #2120, opened 2022-06-07
Add tests for `sf::ConvexShape`

ChrisThrasher:
## Description

More tests!

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android


Bromeon (commented):
Awesome, thanks for adding so many tests! :) 

The tests look good, they made me think about some definitions we use in the documentation though:

* The doc of [`setPointCount()`](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1ConvexShape.php#a56e6e79ade6dd651cc1a0e39cb68deae) says:
   > _count_ must be greater than 2 to define a valid shape.

   Now, "valid shape" is subject to bikeshedding. Maybe we can clarify the docs to say "for a well-defined graphical representation". 

* Similarly, [`setPoint()`](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1ConvexShape.php#a5929e0ab0ba5ca1f102b40c234a8e92d) says:
   > Don't forget that the polygon must remain convex, and the points need to stay ordered! setPointCount must be called first in order to set the total number of points. The result is undefined if index is out of the valid range.

   We create a temporarily non-convex shape (I think so at least; having 41 point at the origin and 1 outside is a degenerate case). Maybe the docs can explicitly allow temporary violation of the invariants, in order to gradually initialize the shape.

ChrisThrasher:
> Now, "valid shape" is subject to bikeshedding. Maybe we can clarify the docs to say "for a well-defined graphical representation".

How about this?
```cpp
    /// \a count must be greater than 2 to see anything on screen.
```

> We [can] create a temporarily non-convex shape (I think so at least; having 41 point at the origin and 1 outside is a degenerate case). Maybe the docs can explicitly allow temporary violation of the invariants, in order to gradually initialize the shape.

The bit about keeping points ordered is odd because the underlying container is ordered so I think we can remove that verbiage. How about this?
```cpp
    /// Don't forget that the shape must be convex when drawn.
    /// Point count must be specified beforehand. The result is
    /// undefined if \a index is greater than or equal to getPointCount
```

Bromeon:
> How about this?
> 
> ```c++
>     /// \a count must be greater than 2 to see anything on screen.
> ```

I'm not sure if that's true -- isn't a line drawn for 2 points? And possibly a point for 1? (especially with outline)

Maybe rather
```c++
     /// \a For the shape to be rendered as expected, count must be greater or equal to 3.
```

---

> The bit about keeping points ordered is odd because the underlying container is ordered so I think we can remove that verbiage. How about this?
> 
> ```c++
>     /// Don't forget that the shape must be convex when drawn.
>     /// Point count must be specified beforehand. The result is
>     /// undefined if \a index is greater than or equal to getPointCount
> ```

I understood "ordered" as "in an order around the center". And I would explicitly mention that temporary violation is OK. Also, "undefined behavior" instead of "undefined result".

```c++
/// Don't forget that the shape must be convex and the 
/// order of points matters. Points should not overlap.
/// This applies to rendering; it is explicitly allowed
/// to temporarily have non-convex or degenerate shapes
/// when not drawn (e.g. during shape initialization).
///
/// Point count must be specified beforehand. The behavior is
/// undefined if \a index is greater than or equal to getPointCount.
```

ChrisThrasher:
Rebased on `master` and updated the docs.

eXpl0it3r:

PR #2115, opened 2022-06-05
SFML/Graphics/Text.cpp: Compare utf32-characters to utf32 character literals consistently

Baardi:
Thanks a lot for making a contribution to SFML! 🙂

Before you create the pull request, we ask you to check the follow boxes. (For small changes not everything needs to ticked, but the more the better!)

* [X] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before?
* [x] Does the code follow the SFML [Code Style Guide](https://www.sfml-dev.org/style.php)?
* [ ] Have you provided some example/test code for your changes?
* [ ] If you have additional steps which need to be performed list them as tasks!

----

## Description

Please describe your pull request.

As discussed during [https://github.com/SFML/SFML/pull/2107](https://github.com/SFML/SFML/pull/2107), here's a pull-request replacing L-literals with U-literals, when comparing towards sf::Uint32-character literals (from sf::String). I tried searching for other cases requiring a similar fix, but I wasn't able to find any. I'm not too familiar withthis codebase, so I might've missed some, though.

This PR is related to the issue #
[https://github.com/SFML/SFML/pull/2107](https://github.com/SFML/SFML/pull/2107)

## Tasks

* [ ] Tested on Linux
* [X] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Describe how to best test these changes. Please provide a [minimal, complete and verifiable example](https://stackoverflow.com/help/mcve) if possible, you can use the follow template as a start:

This can be tested, simply by creating and drawing sf::text-objects in a simple project.
Test specifically '\n', '\t ', '\r' and ' ' in this project, as well as a "Hello World". It shouldn't appear any different than before. Test that it compiles for all the platforms this is supposed to be compiled for in sfml 3.0.

PR #2121, opened 2022-06-07
Update Text.cpp

DanielRabl:
Making all character comparisons to u32 instead of wchar_t since sf characters are of type sf::Uint32.


ChrisThrasher:
Redundant with #2115

ChrisThrasher:
Yes, please. If you found anything that was missed in #2115 then mention that in that PR thread.

PR #2125, opened 2022-06-08
Unify float literal usage in tests

ChrisThrasher:
## Description

The inconsistent use of numeric literals is 100% the fault of whoever has been writing all these unit tests so run `git blame` and yell at that idiot. Unifying how we use literals also has the knock-on effect of removing tons of code (as measured by character count) which has readability gains. For anyone curious about the `int` -> `float` conversion, this is 100% safe because these `int` literals are known by the compiler to convert losslessly to `float`s so there's no correctness lost in this PR.

The best way to look at this PR is me going back and cleaning up inconsistencies in my own code from the last few months.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

### DIFF
--- a/test/Graphics/Transform.cpp
+++ b/test/Graphics/Transform.cpp
@@ -29,63 +29,63 @@ TEST_CASE("sf::Transform class - [graphics]")
 
         SUBCASE("3x3 matrix constructor")
         {
-            const sf::Transform transform(10.0f, 11.0f, 12.0f,
-                                          13.0f, 14.0f, 15.0f,
-                                          16.0f, 17.0f, 18.0f);
+            const sf::Transform transform(10.f, 11.f, 12.f,
+                                          13.f, 14.f, 15.f,
+                                          16.f, 17.f, 18.f);
             const std::vector<float> matrix(transform.getMatrix(), transform.getMatrix() + 16);
-            CHECK(matrix == std::vector<float>{10.0f, 13.0f, 0.0f, 16.0f,
-                                               11.0f, 14.0f, 0.0f, 17.0f,
-                                                0.0f,  0.0f, 1.0f,  0.0f,
-                                               12.0f, 15.0f, 0.0f, 18.0f});
+            CHECK(matrix == std::vector<float>{10.f, 13.f, 0.f, 16.f,
+                                               11.f, 14.f, 0.f, 17.f,
+                                                0.f,  0.f, 1.f,  0.f,
+                                               12.f, 15.f, 0.f, 18.f});
         }
     }
 
     SUBCASE("Identity matrix")
     {
         const std::vector<float> matrix(sf::Transform::Identity.getMatrix(), sf::Transform::Identity.getMatrix() + 16);
-        CHECK(matrix == std::vector<float>{1.0f, 0.0f, 0.0f, 0.0f,
-                                           0.0f, 1.0f, 0.0f, 0.0f,
-                                           0.0f, 0.0f, 1.0f, 0.0f,
-                                           0.0f, 0.0f, 0.0f, 1.0f});
+        CHECK(matrix == std::vector<float>{1.f, 0.f, 0.f, 0.f,
+                                           0.f, 1.f, 0.f, 0.f,
+                                           0.f, 0.f, 1.f, 0.f,
+                                           0.f, 0.f, 0.f, 1.f});
     }
 
     SUBCASE("getInverse()")
     {
         CHECK(sf::Transform::Identity.getInverse() == sf::Transform::Identity);
-        CHECK(sf::Transform(1.0f, 2.0f, 3.0f, 4.0f, 5.0f, 6.0f, 7.0f, 8.0f, 9.0f).getInverse() == sf::Transform::Identity);
-        CHECK(sf::Transform(1.0f, 2.0f, 3.0f, 4.0f, 5.0f, 4.0f, 3.0f, 2.0f, 1.0f).getInverse() == sf::Transform(0.375f, -0.5f, 0.875f, -1.0f, 1.0f, -1.0f, 0.875f, -0.5f, 0.375f));
+        CHECK(sf::Transform(1.f, 2.f, 3.f, 4.f, 5.f, 6.f, 7.f, 8.f, 9.f).getInverse() == sf::Transform::Identity);
+        CHECK(sf::Transform(1.f, 2.f, 3.f, 4.f, 5.f, 4.f, 3.f, 2.f, 1.f).getInverse() == sf::Transform(0.375f, -0.5f, 0.875f, -1.f, 1.f, -1.f, 0.875f, -0.5f, 0.375f));
     }
 
     SUBCASE("transformPoint()")
     {
-        CHECK(sf::Transform::Identity.transformPoint({-10.0f, -10.0f}) == sf::Vector2f(-10.0f, -10.0f));
-        CHECK(sf::Transform::Identity.transformPoint({-1.0f, -1.0f}) == sf::Vector2f(-1.0f, -1.0f));
-        CHECK(sf::Transform::Identity.transformPoint({-1.0f, 0.0f}) == sf::Vector2f(-1.0f, 0.0f));
-        CHECK(sf::Transform::Identity.transformPoint({0.0f, 0.0f}) == sf::Vector2f(0.0f, 0.0f));
-        CHECK(sf::Transform::Identity.transformPoint({0.0f, 1.0f}) == sf::Vector2f(0.0f, 1.0f));
-        CHECK(sf::Transform::Identity.transformPoint({1.0f, 1.0f}) == sf::Vector2f(1.0f, 1.0f));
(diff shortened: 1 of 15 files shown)


Bromeon:
Sorry for always nitpicking on stylistic issues, I feel bad given the huge effort you have put into SFML over the last months 😬

One issue with using integer literals as floats is that we might not be able to do that 100% of the time (thus not fully "unifying" things). Examples I can imagine are arguments to generic functions, overloaded functions, `{ }` initialization syntax (like `{2, 1}` for vectors), etc.

Other than that it's probably fine, as long as people agree it's more readable that way. I usually write the `.f` explicitly, but that might also be a habit stemming from old compilers, which issued warnings for such cases.

ChrisThrasher:
> One issue with using integer literals as floats is that we might not be able to do that 100% of the time (thus not fully "unifying" things).

Maybe "unify" is the wrong word here. I mostly just wanted to make code less verbose wherever possible. I said "unify" because in earlier unit tests I was always adding `.0f` then realized I could just do `.f` _then_  I realized could actually just use integer literals in most places so if you look closely, the style of the unit tests I wrote shifted over time to get marginally simpler. 

> One issue with using integer literals as floats is that we might not be able to do that 100% of the time (thus not fully "unifying" things). Examples I can imagine are arguments to generic functions, overloaded functions, { } initialization syntax (like {2, 1} for vectors), etc.

In a few places I had to be sure I wasn't accidentally calling a different overload but for the most part SFML doesn't overload on `float`s and `int`s. 

> Other than that it's probably fine, as long as people agree it's more readable that way. I usually write the .f explicitly, but that might also be a habit stemming from old compilers, which issued warnings for such cases.

I'm cool with anything so long as we're not requiring the trailing `0` after the decimal point. I know I'm getting _really_ into the weeds with this but I figured it was worth a PR cleaning up the mess I made in `test/` even if we don't make a point to maintain a specific style going forward.

Bromeon:
Sounds good. I'm personally all for readable tests! It's just that this may be highly subjective, some people prefer things to be "explicit but more verbose". I'm fine either way here, concise tests means also they are slightly more motivating to write 🙂

ChrisThrasher:
> It's just that this may be highly subjective

This is probably a one-time change since maintaining such rigorous consistency with `float` literals will be painfully tedious. 

For what it's worth, `src/` and `include/` are already very consistent with `float` literals. `examples/` has a significant number of inconsistencies but there aren't meaningful readability gains to be made there. Fixing that up is mostly an academic exercise. I omitted those changes from this PR for the sake keep scope from exploding.

Bromeon:
Hm, there's probably no tooling available to automate that? Like a linter with a built-in "fix" mode? 🔧

ChrisThrasher:
Not that I'm aware of. clang-tidy has a built-in fix mode but clang-tidy doesn't address these kinds of discrepancies. As an aside, I've played around with running clang-tidy over SFML and eventually want to add that to our CI pipeline.

eXpl0it3r:

eXpl0it3r:
I personally prefer `1.f` etc as well, as it's more explicit of what I'm wanting to achieve and removes the need to differentiate if there are int and float overloads.

vittorioromeo:
I'll bite -- if we need a float expression, isn't it better to explicitly use a float literal rather than using an integer literal and relying on an implicit conversion...? The former seems more sensible to me.

ChrisThrasher:
2 votes to 1 from the maintainers so it looks like we have a decision. I still want to remove all the instances of `.0f` I added but I'll replace those with `.f` instead.

eXpl0it3r:

test/Graphics/Transform.cpp
```diff
@@ -94,27 +94,27 @@ TEST_CASE("sf::Transform class - [graphics]")
...
+                              752.f, 1288.f,  942.f));
     }
 
     SUBCASE("translate()")
     {
         sf::Transform transform(9, 8, 7, 6, 5, 4, 3, 2, 1);
-        CHECK(transform.translate({10.0f, 20.0f}) == sf::Transform(9, 8, 257, 6, 5, 164, 3, 2, 71));
-        CHECK(transform.translate({10.0f, 20.0f}) == sf::Transform(9, 8, 507, 6, 5, 324, 3, 2, 141));
+        CHECK(transform.translate({10.f, 20.f}) == sf::Transform(9, 8, 257, 6, 5, 164, 3, 2, 71));
+        CHECK(transform.translate({10.f, 20.f}) == sf::Transform(9, 8, 507, 6, 5, 324, 3, 2, 141));
```
eXpl0it3r, line 117: Shouldn't these be floats as well in the Transform constructor?

test/Graphics/Transform.cpp
```diff
@@ -143,16 +143,16 @@ TEST_CASE("sf::Transform class - [graphics]")
         SUBCASE("About origin")
         {
             sf::Transform transform(1, 2, 3, 4, 5, 4, 3, 2, 1);
-            CHECK(transform.scale({2.0f, 4.0f}) == sf::Transform(2, 8, 3, 8, 20, 4, 6, 8, 1));
-            CHECK(transform.scale({0.0f, 0.0f}) == sf::Transform(0, 0, 3, 0, 0, 4, 0, 0, 1));
-            CHECK(transform.scale({10.0f, 10.0f}) == sf::Transform(0, 0, 3, 0, 0, 4, 0, 0, 1));
+            CHECK(transform.scale({2.f, 4.f}) == sf::Transform(2, 8, 3, 8, 20, 4, 6, 8, 1));
+            CHECK(transform.scale({0.f, 0.f}) == sf::Transform(0, 0, 3, 0, 0, 4, 0, 0, 1));
+            CHECK(transform.scale({10.f, 10.f}) == sf::Transform(0, 0, 3, 0, 0, 4, 0, 0, 1));
```
eXpl0it3r, line 148: Shouldn't these be floats as well in the Transform constructor?

test/Graphics/Transform.cpp
```diff
@@ -143,16 +143,16 @@ TEST_CASE("sf::Transform class - [graphics]")
...
         }
 
         SUBCASE("About custom point")
         {
             sf::Transform transform(1, 2, 3, 4, 5, 4, 3, 2, 1);
-            CHECK(transform.scale({1.0f, 2.0f}, {1.0f, 0.0f}) == sf::Transform(1, 4, 3, 4, 10, 4, 3, 4, 1));
-            CHECK(transform.scale({0.0f, 0.0f}, {1.0f, 0.0f}) == sf::Transform(0, 0, 4, 0, 0, 8, 0, 0, 4));
+            CHECK(transform.scale({1.f, 2.f}, {1.f, 0.f}) == sf::Transform(1, 4, 3, 4, 10, 4, 3, 4, 1));
+            CHECK(transform.scale({0.f, 0.f}, {1.f, 0.f}) == sf::Transform(0, 0, 4, 0, 0, 8, 0, 0, 4));
```
eXpl0it3r, line 155: Shouldn't these be floats as well in the Transform constructor?

eXpl0it3r (changes requested):
Looks like not all `sf::Transform` constructor call used float literals to begin with 🤔

ChrisThrasher:
> Looks like not all sf::Transform constructor call used float literals to begin with

Haha I was wondering if anyone would notice that. In my latest tests I started using loads of integer literals, hence this PR. Sadly I can't easy grep for integer literals that are being used to initialize `float`s but I can go through and find and fix those.

Bromeon:
Looks like you opened Pandora's Box 😁

ChrisThrasher:
I guess I deserve this for opening the PR ¯\_(ツ)_/¯

ChrisThrasher:
I added `.f` in about a hundred more places. Let me know if there's anything I missed.

ChrisThrasher:
I found a few hundred more places where float literals could be used. The current state of this PR illustrates why I've been using integer literals. I don't think hundreds of unnecessary `.f` suffixes make the code any easier to read. Plus it's just a huge pain to audit the code for missing `.f` suffixes whereas finding and removing unnecessary `.f` suffixes is pretty easy to enforce.

Bromeon:
Guys, I know it's "correct" but are we solving a real problem here? Readability _is_ subjective and while there _may_ be obscure corner-case bugs related to incorrect float/integer literal usage, I have the feeling this is a bit of an academic exercise. Especially since the other literals haven't been more correct in C++98, yet no one has complained -- at least to my knowledge.

Talking of readability, a bigger pain imho is the proliferation of `static_cast<Type>(expression)` all over the place. I know it's often unavoidable, but even things like `toUint8(expression)` would help a lot; but I know this is also controversial.

Yet even that is a relatively small issue. I'm not sure if we shouldn't focus our resources on more pressing issues which add real value to users -- like coming to a consensus on the long overdue scancode design.

ChrisThrasher:
I'm definitely sucking up much of the oxygen in the room with this PR and maybe regret opening it. This is far from my top priority so if we can't come to a consensus on this I'm fine closing this. I'd also much rather we spend our time discussing scancodes or SFML 3 design decisions. I've tried proposing some bigger changes I want to make but haven't gotten very much feedback, hence me filling my time with these lesser important PRs.

PR #2126, opened 2022-06-10
Simplify how vectors are deallocated

ChrisThrasher:
## Description

After discussion in Discord, I came to learn that this "swap with empty temporary" pattern was a way to clear _and deallocate_ a vector. This PR maintains that behavior but uses the more expressive C++11 syntax for default initializing a variable. 

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

### DIFF
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -527,7 +527,7 @@ void Font::cleanup()
 
     // Reset members
     m_pages.clear();
-    std::vector<Uint8>().swap(m_pixelBuffer);
+    m_pixelBuffer = {};
 }
 
 
--- a/src/SFML/Graphics/Image.cpp
+++ b/src/SFML/Graphics/Image.cpp
@@ -66,7 +66,7 @@ void Image::create(const Vector2u& size, const Color& color)
     else
     {
         // Dump the pixel buffer
-        std::vector<Uint8>().swap(m_pixels);
+        m_pixels = {};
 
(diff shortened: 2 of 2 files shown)


eXpl0it3r:

vittorioromeo (changes requested):
...did you test this?
https://gcc.godbolt.org/z/eGsrYKoTe

ChrisThrasher:
> ...did you test this?
https://gcc.godbolt.org/z/eGsrYKoTe

Woah, that's surprising. What am I missing about `= {}`? I could have sworn using `= {}` would be identical to ` = std::vector<Uint8>()` which does do what I expected.

vittorioromeo:
> > ...did you test this?
> > https://gcc.godbolt.org/z/eGsrYKoTe
> 
> Woah, that's surprising. What am I missing about `= {}`? I could have sworn using `= {}` would be identical to ` = std::vector<Uint8>()` which does do what I expected.

I think that `= {}` might be hitting the `std::initializer_list` assignment operator? I also don't think that ` = std::vector<Uint8>()` is *required* to clear capacity, it's just assignment. I think a conforming implementation could check if it's empty and do nothing at all.

The most idiomatic way to achieve what you want is

```cpp
v.clear(); 
v.shrink_to_fit();
```

but `shrink_to_fit` is also not guaranteed to do anything at all :D

PR #2127, opened 2022-06-10
Remove empty destructors

ChrisThrasher:
## Description

The first commit removes empty destructors. These serve no purpose and implicitly disable move semantics which is no good. The 2nd commit finds all the empty virtual destructors that we must keep around and implements them with `= default` which gives the compiler a little more information to work with and is generally more expressive.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android


ChrisThrasher:
Maybe it's worth doing a full "Rule of 0/5" audit to make sure we're doing the best we can to support all the semantics we want for SFML types. Along similar lines, I'm curious why we use `= default` in the .cpp file instead of in the header. Putting `= default` in the header seems like an easy win for code readability and requires less code in general but maybe I'm missing a good reason for the status quo.

eXpl0it3r:

eXpl0it3r:
There are two instances ([1](https://github.com/SFML/SFML/blob/master/src/SFML/Graphics/RenderTextureImpl.hpp#L53), [2](https://github.com/SFML/SFML/blob/master/include/SFML/Network/Ftp.hpp#L254)) where it's used in the header file, but I agree, it seems to make more sense to have it set in the header file, as we do with `= delete`.
I don't necessarily see it brought up in #1911, so maybe @vittorioromeo can share his reason for going with source file `= default`

vittorioromeo:
> There are two instances ([1](https://github.com/SFML/SFML/blob/master/src/SFML/Graphics/RenderTextureImpl.hpp#L53), [2](https://github.com/SFML/SFML/blob/master/include/SFML/Network/Ftp.hpp#L254)) where it's used in the header file, but I agree, it seems to make more sense to have it set in the header file, as we do with `= delete`. I don't necessarily see it brought up in #1911, so maybe @vittorioromeo can share his reason for going with source file `= default`

That's necessary because we are only using a forward declaration of the type managed by the smart pointer in the header. The source file has the full definition, so that's the place were we want to "instantiate" the destructor, otherwise compilation would fail.

src/SFML/Graphics/Font.cpp
```diff
@@ -129,13 +129,6 @@ m_pixelBuffer(copy.m_pixelBuffer)
 }
 
 
-////////////////////////////////////////////////////////////
-Font::~Font()
-{
-    cleanup();
```
vittorioromeo, line 135: We don't need to call this anymore...?

vittorioromeo: Also I think you need to keep this in the source file to avoid breaking the Android build.

vittorioromeo (changes requested):

src/SFML/Graphics/Font.cpp
ChrisThrasher: https://github.com/SFML/SFML/blob/709530d062ac9c452f50b06d37cdd800c2f9ba68/src/SFML/Graphics/Font.cpp#L522-L531

`Font::cleanup` doesn't do anything that the destructors of class members aren't already doing.

> Also I think you need to keep this in the source file to avoid breaking the Android build.

Can you help me understand why this is? Why isn't the implicitly defined destructor good enough?

vittorioromeo: > Can you help me understand why this is? Why isn't the implicitly defined destructor good enough? Would it have something to do with the to move-related special member functions that are defaulted but can now be removed?

If you have

```cpp
// someclass.h

struct SomeDependency;

struct SomeClass
{
    std::unique_ptr<SomeDependency> m_dep;
};
```

that will fail to compile because `std::unique_ptr<SomeDependency>::~unique_ptr()` is going to attempt to call `SomeDependency::~SomeDependency()` which is not available in scope.

If you instead do:

```cpp
// someclass.h

struct SomeDependency;

struct SomeClass
{
    std::unique_ptr<SomeDependency> m_dep;
    ~SomeClass();
};


// someclass.cpp
#include <someclass.h>
#include <somedependency.h>

SomeClass::~SomeClass() = default;
```

then the destructor of `std::unique_ptr` will have visibility of `SomeDependency::~SomeDependency()`.

ChrisThrasher: So the destructor is necessary because we choose to use forward declarations instead of including the header?

eXpl0it3r:

src/SFML/Graphics/Font.cpp
vittorioromeo: > So the destructor is necessary because we choose to use forward declarations instead of including the header?

Yes, and we chose the latter to improve compilation speed and reduce physical dependencies between SFML's components.

vittorioromeo (approved):
LGTM now :)

PR #2130, opened 2022-06-12
Modernize examples

ChrisThrasher:
## Description

Chipping away at some of the C++17 items mentioned [here](https://github.com/SFML/SFML/wiki/SD%3A-CPP-17-Support#filesystem-api). I also found an easy way to simplify the CMake code for all the examples that should help improve readability and maintainability.

I ran the Tennis, Island, and Shader examples on macOS and could not observe any different behavior.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

### DIFF
--- a/examples/X11/X11.cpp
+++ b/examples/X11/X11.cpp
@@ -8,6 +8,7 @@
 #include <gl.h>
 
 #include <X11/Xlib.h>
+#include <array>
 #include <iostream>
 #include <cmath>
 #include <cstdlib>
@@ -93,7 +94,7 @@
     glRotatef(elapsedTime * 18.f, 0.f, 0.f, 1.f);
 
     // Define a 3D cube (6 faces made of 2 triangles composed by 3 vertices)
-    static const GLfloat cube[] =
+    constexpr std::array<GLfloat, 216> cube =
     {
         // positions    // colors
         -50, -50, -50,  1, 1, 0,
@@ -140,8 +141,8 @@
     };
 
     // Draw the cube
-    glVertexPointer(3, GL_FLOAT, 6 * sizeof(GLfloat), cube);
-    glColorPointer(3, GL_FLOAT, 6 * sizeof(GLfloat), cube + 3);
+    glVertexPointer(3, GL_FLOAT, 6 * sizeof(GLfloat), cube.data());
+    glColorPointer(3, GL_FLOAT, 6 * sizeof(GLfloat), cube.data() + 3);
     glDrawArrays(GL_TRIANGLES, 0, 36);
 
     return true;
--- a/examples/island/Island.cpp
+++ b/examples/island/Island.cpp
@@ -8,6 +8,7 @@
 #include <SFML/Graphics.hpp>
 
 #include <algorithm>
+#include <array>
 #include <deque>
 #include <iostream>
 #include <mutex>
@@ -161,8 +162,8 @@ int main()
     statusText.setPosition({(windowWidth - statusText.getLocalBounds().width) / 2.f, (windowHeight - statusText.getLocalBounds().height) / 2.f});
 
     // Set up an array of pointers to our settings for arrow navigation
-    Setting settings[] =
-    {
+    constexpr std::array<Setting, 9> settings =
+    {{
         {"perlinFrequency",     &perlinFrequency},
         {"perlinFrequencyBase", &perlinFrequencyBase},
         {"heightBase",          &heightBase},
@@ -172,10 +173,9 @@ int main()
         {"heightFactor",        &heightFactor},
         {"heightFlatten",       &heightFlatten},
         {"lightFactor",         &lightFactor}
-    };
+    }};
 
-    const int settingCount = 9;
-    int currentSetting = 0;
+    std::size_t currentSetting = 0;
 
     std::ostringstream osstr;
     sf::Clock clock;
@@ -199,8 +199,8 @@ int main()
                 switch (event.key.code)
                 {
                     case sf::Keyboard::Enter: generateTerrain(terrainStagingBuffer.data()); break;
-                    case sf::Keyboard::Down:  currentSetting = (currentSetting + 1) % settingCount; break;
-                    case sf::Keyboard::Up:    currentSetting = (currentSetting + settingCount - 1) % settingCount; break;
+                    case sf::Keyboard::Down:  currentSetting = (currentSetting + 1) % settings.size(); break;
+                    case sf::Keyboard::Up:    currentSetting = (currentSetting + settings.size() - 1) % settings.size(); break;
                     case sf::Keyboard::Left:  *(settings[currentSetting].value) -= 0.1f; break;
                     case sf::Keyboard::Right: *(settings[currentSetting].value) += 0.1f; break;
                     default: break;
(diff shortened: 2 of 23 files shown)


examples/X11/X11.cpp
```diff
@@ -93,7 +93,7 @@
     glRotatef(elapsedTime * 18.f, 0.f, 0.f, 1.f);
 
     // Define a 3D cube (6 faces made of 2 triangles composed by 3 vertices)
-    static const GLfloat cube[] =
+    constexpr std::array<GLfloat, 216> cube =
```
vittorioromeo, line 96: I guess we are not using CTAD here because we need to make sure that the values are `GLfloat`, right?

examples/shader/Shader.cpp
```diff
@@ -135,15 +135,20 @@ class StormBlink : public Effect
 
     bool onLoad() override
     {
+        std::mt19937 rng = std::mt19937(std::random_device()());
```
vittorioromeo, line 138: A few things here:

- The random device should be initialized once at the beginning of the program. I guess a global/singleton would be OK in this case...

The code can also be simplified a bit:

```suggestion
        std::mt19937 rng(global_random_device());
```

examples/vulkan/Vulkan.cpp
```diff
@@ -1464,7 +1465,7 @@ class VulkanExample
     // Create our index buffer and upload its data
     void setupIndexBuffer()
     {
-        uint16_t indexData[] = {
+        constexpr std::array<uint16_t, 36> indexData = {
```
vittorioromeo, line 1468:
```suggestion
        constexpr std::array<std::uint16_t, 36> indexData = {
```

vittorioromeo (changes requested):
Good improvements, just a few things to tweak

examples/X11/X11.cpp
ChrisThrasher: That's correct.

examples/shader/Shader.cpp
ChrisThrasher: I was running into the most vexing parse which forced me to use ` = `. I would normally just use `auto` but I wanted to match SFML style. Otherwise you're right I need to stop reinitializing that generator over and over again.

ChrisThrasher: I made that `mt19937` a global so that it's state persists across calls to that function. Not sure why I didn't just do that to begin with. For what it's worth, I'm aware that `random_device()` doesn't fully seed a `mt19937`. Doing that looks like this:
```cpp
static auto rng = []() {
    auto seed_data = std::array<int, std::mt19937::state_size>();
    auto rd = std::random_device();
    std::generate_n(seed_data.data(), seed_data.size(), std::ref(rd));
    auto seed_seq = std::seed_seq(seed_data.begin(), seed_data.end());
    return std::mt19937(seed_seq);
}();
```
I figured that all this complexity wasn't worth it for an example though.

examples/tennis/Tennis.cpp
```diff
@@ -29,7 +27,7 @@ std::filesystem::path resourcesDir()
 ////////////////////////////////////////////////////////////
 int main()
 {
-    std::srand(static_cast<unsigned int>(std::time(nullptr)));
+    std::mt19937 rng = std::mt19937(std::random_device()());
```
vittorioromeo, line 30: Ditto

vittorioromeo (changes requested):

examples/shader/Shader.cpp
```diff
@@ -4,9 +4,14 @@
 ////////////////////////////////////////////////////////////
 #include "Effect.hpp"
 #include <array>
-#include <cmath>
+#include <random>
 
 
+namespace
+{
+    std::mt19937 rng = std::mt19937(std::random_device()());
```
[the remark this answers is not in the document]
ChrisThrasher: I don't really want to keep the `=`. I would prefer to use `std::mt19937 rng(std::random_device()());` but this is what AppleClang++ reports
```
/Users/thrasher/Projects/sfml/examples/shader/Shader.cpp:12:40: error: function cannot return function type 'std::random_device ()'
    std::mt19937 rng(std::random_device()());
```
I'll use `auto` instead. It's what I most prefer anyways.

examples/tennis/Tennis.cpp
ChrisThrasher: Another vexing parse. Either we break with SFML style and use `auto` or we accept this awkward syntax.

ChrisThrasher:
I found a new solution to the most vexing parse that was forcing me to initial variables with `=`. Instead I just make a new local variable for the random device. This means we can continue to match SFML's style of not using `auto`.

vittorioromeo:
> This means we can continue to match SFML's style of not using `auto`.

Is that our official stance? Certainly not my personal preference... :)

@eXpl0it3r

ChrisThrasher:
For the record I'd also prefer we start using `auto` where there is technical merit and potentially go so far as dictating that all future patches use `auto` for all variable declarations. I haven't already used it because I wanted to make sure my patches blended into SFML as well as possible.

Bromeon:
> For the record I'd also prefer we start using `auto` more often and potentially go so far as dictating that all future patches use `auto` for all variable declarations.

Direct question: why? 🙂

ChrisThrasher:
> Direct question: why? 🙂

For starters, it's not a big deal and I'm fine sticking with the current style since there's so much inertia behind it. There are many other things I'd rather the team focus on than my preference to use `auto`.

That being said, here's why I like `auto` for variable declarations:
- Requires that you initialize the variable
- Easier to visually scan for variable declarations (somewhat subjective, but true for me personally)
- I just think it's prettier (this is absolutely subjective lol)
- Impossible to encounter the most vexing parse
- Reduces the likelihood of accidentally making implicit conversions
- Lets the compiler automatically propagate type changes throughout the codebase with less effort on our part

examples/island/Island.cpp
```diff
@@ -161,8 +162,8 @@ int main()
     statusText.setPosition({(windowWidth - statusText.getLocalBounds().width) / 2.f, (windowHeight - statusText.getLocalBounds().height) / 2.f});
 
     // Set up an array of pointers to our settings for arrow navigation
-    Setting settings[] =
-    {
+    std::array<Setting, 9> settings =
```
kimci86, line 165: This array can also be constexpr.

eXpl0it3r:

examples/island/Island.cpp
ChrisThrasher: Oh yeah you're right! I didn't mark that constexpr because I saw the following lines farther down:
```cpp
case sf::Keyboard::Left:  *(settings[currentSetting].value) -= 0.1f; break;
case sf::Keyboard::Right: *(settings[currentSetting].value) += 0.1f; break;
```
It almost appears as if the array is being mutated but only some non-const pointed-to values are being changed. Thank you!

ChrisThrasher:
Addressed a comment and added a missing header that was breaking one of the extended CI jobs. This PR is good to go now.

eXpl0it3r:

ChrisThrasher:
Looks like I fixed the last failing CI job 👍

eXpl0it3r:
🥳

eXpl0it3r:
> Is that our official stance? Certainly not my personal preference... :)

I'm an [AAA (Almost Always Auto)](https://dev.my-gate.net/2021/11/19/almost-always-auto/) kind of person, but I'm aware that it's controversial.

PR #2131, opened 2022-06-12
Use scoped enumerations in Network module

ChrisThrasher:
## Description

Chipping away at more C++17 modernizations from [here](https://github.com/SFML/SFML/wiki/SD%3A-CPP-17-Support#filesystem-api). To keep scope manageable, this PR only touches the Network module. 

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android


eXpl0it3r:

eXpl0it3r:
I think there's an initial discussion to be had about scoped-enums.

@SFML/sfml is the approach shown here, how we want to do it? Or did you imagine something else (e.g. having the enums outside the given class with a different name)?

Bromeon:
There's a general question about the utility of SFML's HTTP and FTP sockets in the room. In the late 2000s when SFML was designed, this made sense, as a lot of the web was using HTTP and the alternative was often using C socket libraries.

In 2022, providing unencrypted HTTP and FTP is close to useless beyond any toy examples, and the one place where unencrypted HTTP is still widely used are VPCs (inside hosted clouds), not a typical domain of SFML applications. I'd even go as far as saying we shouldn't promote bad security practices and rather point to other libraries from the start. Even if people start integrating `sf::Http` or `sf::Ftp` successfully in their games, they will need to rewrite their code using a different library as soon as things get more serious.

At the same time, [there is a plethora of networking libraries for C++](https://github.com/fffaraz/awesome-cpp#networking), most of which provide better support for today's web. As such, I would vote for the **removal of `sf::Http` and `sf::Ftp` in SFML 3.** This would allow us to focus resources on other parts of SFML which bring more value to the user.

Bromeon:
Regarding the enum discussion, using scoped enums is probably the more idiomatic approach, but I dislike the deterioration in ergonomics in this case. And as you know, in that regard I'm more the pragmatic than the 100% idiomatic kind of guy 😉 but of course, we should have some guidelines for consistency.

`sf::SoundStream::Status::Stopped` is quite a bit more verbose than `sf::SoundStream::Stopped`, while not adding meaningful information. On the other hand, it does make clearer that the type itself is called `sf::SoundStream::Status`.

One option would be what eXpl0it3r suggested:
```cpp
// in namespace sf
enum class SocketStatus {
    Done,
    NotReady,
    Partial,
    Disconnected,
    Error
};
```
The slight advantage here would be that there is no confusion with inheritance, and there's one way to express the type. Otherwise, some people write `TcpSocket::Status::Done`, some write `UdpSocket::Status::Done` and some `Socket::Status::Done` -- which looks like 3 different types but is 1.

ChrisThrasher:
I did some experimentation and C++20's `using enum` feature solves this problem perfectly. We continue to maintain the same user experience without requiring an extra `EnumName::` at all use sites but gain the type safety of scoped enumerations. Luckily this feature is already implemented across the latest version of the big OSes. Check this out:

https://github.com/ChrisThrasher/SFML/commit/2ec8ee049395610d143e18791fb2d7f62462574a

One notable change is how the `ubuntu-latest` and `mac-latest` GitHub Actions runners [aren't actually the latest versions of those OSes](https://github.com/actions/virtual-environments#available-environments). I had to manually specify a newer version but once I did that, you can see that today's operating systems with their default compilers already support a non-zero subset of C++20 features. That's very encouraging because it means SFML 3 can start to consider using a subset of C++20 features. There are still issues with Android but hopefully that's a solvable problem.

SMFL 3 should seriously consider whether targeting C++20 is a wise move. SFML 3 is on track for maybe a 2023 or 2024 release. By then, we'll have a few more updates to various popular toolchains that will continue to improve on C++20 compatibility. It will be reasonable to ask why a library's new major release uses a 6 or 7 year old language standard and currently we have no good answer for that.

eXpl0it3r:

ChrisThrasher:
Rebased because this touches `IpAddress.cpp` which was recently modified and tested.

eXpl0it3r:

ChrisThrasher:
Fixed conflicts with `master`

eXpl0it3r:

vittorioromeo:
> @vittorioromeo Could you weigh in on how, if at all, SFML ought to make use of scoped enumerations?

I take the same stance I do in my book -- if an enumeration is supposed to represent a fixed set of choices, then `enum class` is the way. If an enumeration is supposed to work more-or-less like an integer or bitflags (i.e. we want implicit conversions), traditional `enum` wrapped in a `class` is the way.

Bromeon:
A bit off-topic, but a nice idiom I found to allow both strong type safety and non-verbose integer conversions is the unary plus operator:
```cpp
#define INTEGRIFY(Enum)                                  \                                  
inline std::underlying_type_t<Enum> operator+(Enum e)    \
{                                                        \
    return static_cast<std::underlying_type_t<Enum>>(e); \
}
```
Usage:
```cpp
enum class Thing { A, B, C };
INTEGRIFY(Thing)

int main()
{
    int a = +Thing::A;
}

**Comment by ChrisThrasher** [comment]

**Comment by ChrisThrasher** [comment]
Does the team have any more thoughts on this?

File: examples/sound/Sound.cpp [hunk 1]
```diff
@@ -30,7 +30,7 @@ void playSound()
     sound.play();
 
     // Loop while the sound is playing
-    while (sound.getStatus() == sf::Sound::Playing)
+    while (sound.getStatus() == sf::Sound::Status::Playing)
```
kimci86, line 33: Is this intended? This line and a few other lines are related to Audio module.

ChrisThrasher: I think that's left around from an early version of this PR where I was also using scoped enums in the audio module. Good catch I'll remove that change.

ChrisThrasher:
Rebased on master and removed some old changes to Audio code left over from an older incarnation of this branch.

ChrisThrasher:
Monthly rebase just for the sake of it.

vittorioromeo:

vittorioromeo (approved):
@eXpl0it3r, @Bromeon: I am generally in favour of this change. Basically, any `enum` that should not be used as an integer but only as a choice between a closed set of options is much better (and more safely) modeled by `enum class` rather than `enum`. The ones chosen by Chris in this PR are good candidates IMHO. 

Are you against merging this? If we merge this, we should perform the same changes to the rest of the `enum` types in SFML that fit the description above.

ChrisThrasher:

ChrisThrasher:
Rebased to get new CI jobs.

eXpl0it3r:

eXpl0it3r (changes requested):
The example code in the doxygen doc comments need to be updated as well

ChrisThrasher:
> The example code in the doxygen doc comments need to be updated as well

Alright I'll try to track down where they're out of date.

ChrisThrasher:
> The example code in the doxygen doc comments need to be updated as well

I found more places where the docs were out of date. Unsure if I caught everything though, especially docs that may live in separate files.

ChrisThrasher:
I searched deeper and found more out-of-date docs.

eXpl0it3r:

eXpl0it3r (approved):

Issue #2129, opened 2022-06-12
sf::Text::getLocalBounds() inconsistency


oomek:
I've recently realized that `sf::Text::getLocalBounds()` is a bit inconsistent.
When the text object has an outline applied then `top` and `left` return correctly compensated negative values, but `width` and `height` still contain the outline. This is a bit problematic to align the text to the right without it being affected by the outline.

danieljpetersen:
Is this on all versions (2.5.1, 2.6, master)?

oomek:
2.6.X

eXpl0it3r:
Am I understanding this correctly that we subtract the outline width top-left, but don't add anything to the overall width-height of the local bounds?

The fact that the top-left is adjusted certainly makes sense to maintain a stable baseline. So the question essentially becomes whether the outline should affect the bounds size.

I currently can't think of a reason why it shouldn't. 🤔

Would this affect the line height, i.e. distance between lines?

oomek:
I've prepared an example which will help you to understand what I mean.
When I try to align text to the right edge with fractional outline values the text is crawling.
Does not matter whether I align to outline or to text body.
With localBoundingBox() returning size without outline it would be cleaner. We could always add the outline value back instead of counting advance of outlined glyphs.

```cpp
#include <cmath>
#include <SFML/Window.hpp>
#include <SFML/Graphics.hpp>

int main(int argc, char *argv[])
{
    sf::RenderWindow window;
    window.create(sf::VideoMode(800, 600, 32), "SFML test");
    window.setVerticalSyncEnabled(true);
    window.setMouseCursorVisible(true);

    sf::Font font;
    std::vector<sf::Text> text;

    if(!font.loadFromFile("segoeui.ttf"))
        return EXIT_FAILURE;

    text.push_back(sf::Text("Lorem ipsum dolor sit amet,", font));
    text.push_back(sf::Text("consectetur adipiscing elit. Nam id", font));
    text.push_back(sf::Text("nisi id libero hendrerit eleifend. Sed", font));
    text.push_back(sf::Text("tincidunt vitae neque maximus", font));
    text.push_back(sf::Text("imperdiet.", font));

    int outline = 0;
    float lineHeight = 40;
    int alignToTextBody = 0;

    int pos = 0;
    for (auto& t : text)
    {
        t.setCharacterSize(27);
        t.setFillColor(sf::Color::White);
        t.setPosition(sf::Vector2f(100.0, 100.0 + pos));
[...]

PR #2132, opened 2022-06-13
Fix corelation between text setOutlineThickness() and getLocalBounds()

oomek:
This fixes text crawling explained in issue [#2129](https://github.com/SFML/SFML/issues/2129)

----

## Description

This PR is related to the issue [#2129](https://github.com/SFML/SFML/issues/2129)

## Tasks

* [ ] Tested on Linux
* [X] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Example code provided in the issue linked above

### DIFF
--- a/src/SFML/Graphics/Text.cpp
+++ b/src/SFML/Graphics/Text.cpp
@@ -500,19 +500,8 @@ void Text::ensureGeometryUpdate() const
         {
             const Glyph& glyph = m_font->getGlyph(curChar, m_characterSize, isBold, m_outlineThickness);
 
-            float left   = glyph.bounds.left;
-            float top    = glyph.bounds.top;
-            float right  = glyph.bounds.left + glyph.bounds.width;
-            float bottom = glyph.bounds.top  + glyph.bounds.height;
-
             // Add the outline glyph to the vertices
             addGlyphQuad(m_outlineVertices, Vector2f(x, y), m_outlineColor, glyph, italicShear);
-
-            // Update the current bounds with the outlined glyph bounds
-            minX = std::min(minX, x + left   - italicShear * bottom);
-            maxX = std::max(maxX, x + right  - italicShear * top);
-            minY = std::min(minY, y + top);
-            maxY = std::max(maxY, y + bottom);
         }
 
         // Extract the current glyph's description
@@ -521,22 +510,29 @@ void Text::ensureGeometryUpdate() const
         // Add the glyph to the vertices
         addGlyphQuad(m_vertices, Vector2f(x, y), m_fillColor, glyph, italicShear);
 
-        // Update the current bounds with the non outlined glyph bounds
-        if (m_outlineThickness == 0)
-        {
-            float left   = glyph.bounds.left;
-            float top    = glyph.bounds.top;
-            float right  = glyph.bounds.left + glyph.bounds.width;
-            float bottom = glyph.bounds.top  + glyph.bounds.height;
+        // Update the current bounds
+        float left   = glyph.bounds.left;
+        float top    = glyph.bounds.top;
+        float right  = glyph.bounds.left + glyph.bounds.width;
+        float bottom = glyph.bounds.top + glyph.bounds.height;
 
-            minX = std::min(minX, x + left  - italicShear * bottom);
-            maxX = std::max(maxX, x + right - italicShear * top);
-            minY = std::min(minY, y + top);
-            maxY = std::max(maxY, y + bottom);
-        }
+        minX = std::min(minX, x + left - italicShear * bottom);
+        maxX = std::max(maxX, x + right - italicShear * top);
+        minY = std::min(minY, y + top);
+        maxY = std::max(maxY, y + bottom);
 
         // Advance to the next character
         x += glyph.advance + letterSpacing;
+    }
+
+    // If we're using outline, update the current bounds
+    if (m_outlineThickness != 0)
+    {
+        float outline = std::abs(std::ceil(m_outlineThickness));
+        minX -= outline;
+        maxX += outline;
+        minY -= outline;
+        maxY += outline;
     }
 
     // If we're using the underlined style, add the last line


src/SFML/Graphics/Text.cpp
```diff
@@ -506,19 +506,19 @@ void Text::ensureGeometryUpdate() const
...
             const Glyph& glyph = m_font->getGlyph(curChar, m_characterSize, isBold, m_outlineThickness);
 
-            float left   = glyph.bounds.left;
-            float top    = glyph.bounds.top;
-            float right  = glyph.bounds.left + glyph.bounds.width;
-            float bottom = glyph.bounds.top  + glyph.bounds.height;
+            // float left   = glyph.bounds.left;
+            // float top    = glyph.bounds.top;
+            // float right  = glyph.bounds.left + glyph.bounds.width;
+            // float bottom = glyph.bounds.top  + glyph.bounds.height;
```
ChrisThrasher, line 512: Just remove code instead of commenting it out.

src/SFML/Graphics/Text.cpp
```diff
@@ -527,24 +516,31 @@ void Text::ensureGeometryUpdate() const
...
+        maxX = std::max(maxX, x + right - italicShear * top);
+        minY = std::min(minY, y + top);
+        maxY = std::max(maxY, y + bottom);
 
         // Advance to the next character
         x += glyph.advance + letterSpacing;
     }
 
+    // If we're using outline, update the current bounds
+    if ((m_outlineThickness != 0) && (x > 0))
```
kimci86, line 535: I don't think the condition `x > 0` is needed.

oomek: It actually is so we don't apply outline offset for empty strings.

kimci86: The function exists early (line 421) for empty strings.
I understand the condition you want would be to check if the string is made or whitespace characters only. Should outline be ignored in this case?

oomek: Sorry I mean strings containing only special symbols tab next line etc. I just followed the other 2 conditions below.

oomek: Correction, x stays 0 only for `\n` and `\r`

src/SFML/Graphics/Text.cpp
```diff
@@ -527,24 +516,31 @@ void Text::ensureGeometryUpdate() const
...
+    {
+        float outline = std::abs(std::ceil(m_outlineThickness));
```
vittorioromeo, line 537: Is `std::abs` necessary here? Do we even support negative outline thicknesses? @eXpl0it3r 
Also, the name `outline` is not very descriptive here. Maybe `absOutlineThickness`?

vittorioromeo (commented):
Seems reasonable, would like to see a before/after picture in a noticeable case with a large outline. Also, would we need to do the same for shapes?

src/SFML/Graphics/Text.cpp
oomek: We don't clamp negative thickness, and it's looking weird ( diamond shape, not round ) hence the need for `abs`

oomek:
> Seems reasonable, would like to see a before/after picture in a noticeable case with a large outline. Also, would we need to do the same for shapes?

I'm not sure what you mean? You have to run the example code provided to understand what it's about. The screenshots won't tell you the story I'm afraid, as they will look the same.

src/SFML/Graphics/Text.cpp
kimci86: What about `std::ceil`? Impacted variables are of type `float`, so why not use the absolute thickness without rounding?

eXpl0it3r: The documentation states:

> Note that negative values are allowed (so that the outline expands towards the center of the shape)

So yes, we support to an extend negative outline thicknesses

eXpl0it3r: The documentation states:

> Be aware that using a negative value for the outline thickness will cause distorted rendering.

So "support" is a strong word, but it's at least a "valid" value, the output is just kind of undefined/not guaranteed

oomek: `ceil` is  required here because the bounding box of each glyph increases in integer steps, when you increment the outline with fractional values. These are texture sizes we're talking about.

eXpl0it3r:
I think the example is okay to demonstrate the issue + fix.

For me the main question is, whether there's something that speaks against including it, just so we don't "fix" something that was intentionally designed that way. But if nobody knows about it, then it should be fine for me as well.

src/SFML/Graphics/Text.cpp
kimci86: I tested and noticed that, before your patch, the bounding box grows when there is some visible outline, i.e. not around whitespace characters. It can happen on one side and not the other, for example if the text starts with whitespace or a blank line.
With your current path, the outline impacts the bounding box around any character unless the text ends with a newline character. This special case looks like a logic error to me. Without the condition `(x > 0)`, the outline impacts the bounding box around any character. That is still changing the behavior compared to before your patch but it looks consistent.

oomek: Ok I know what you mean. In that case shouldn't we check both string ends for whitespace characters and add outline for each side separately?

kimci86: Guessing precisely where outline appears might be more complex than checking both ends. Note that an sf::Text instance can represent several lines of text as the string can include newline characters.
Maybe the most consistent (and also simple to implement) is to consider that whitespace characters have an invisible outline that impacts the bounding box. It is consistent with the fact that whitespace characters already impact the bounding box even though those characters are invisible.

kimci86: To summarize, I think once we remove the condition `(x > 0)`, this PR will be good to merge.

eXpl0it3r:

eXpl0it3r:
Do you just want to fix this on master or also for 2.6.x?

oomek:
Yes, also on 2.6.x please.

eXpl0it3r:
Do you want to retarget it or shall I? 😄

oomek:
Please do. I'm not going to be in front of my PC today.

eXpl0it3r:

eXpl0it3r (approved):

ChrisThrasher (approved):
Running the example myself, I can observe this PR making things significantly smooth when scaling the outline up and down.

kimci86 (approved):

eXpl0it3r:
Thanks for the fix! 🥳

PR #2133, opened 2022-06-17
Improve `sf::Time` interoperability with `<chrono>`

ChrisThrasher:
## Description

This PR implements the ideas outlined [here](https://github.com/SFML/SFML/wiki/SD%3A-CPP-17-Support#chrono-api). 100% code coverage is maintained for `sf::Time`.

One important design note is how the constructor from `std::chrono::duration` does not perform a `duration_cast`. This is intentional because I don't want to make it too easy for users to provide a duration type that requires a lossy conversion. The onus is on the user to do that lossy conversion.


include/SFML/System/Time.hpp
```diff
@@ -79,6 +87,14 @@ class Time
     ////////////////////////////////////////////////////////////
     constexpr Int64 asMicroseconds() const;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Return the time value as a std::chorono::duration
+    ///
+    /// \return Time in microseconds
+    ///
+    ////////////////////////////////////////////////////////////
+    constexpr std::chrono::microseconds toDuration() const;
```
kimci86, line 96: Wouldn't it be more consistent to name this method `asDuration`? to be more like `asSeconds` etc.

include/SFML/System/Time.inl
```diff
@@ -30,22 +30,37 @@ m_microseconds(0)
...
+m_microseconds(duration)
+{
+}
+
+
 ////////////////////////////////////////////////////////////
 constexpr float Time::asSeconds() const
 {
-    return static_cast<float>(static_cast<double>(m_microseconds) / 1000000.0);
+    return std::chrono::duration_cast<std::chrono::duration<float>>(m_microseconds).count();
```
kimci86, line 44: `duration_cast` is not needed here.
```suggestion
    return std::chrono::duration<float>(m_microseconds).count();
```

kimci86 (commented):
The private constructor `Time(Int64 microseconds)` could be removed as it is used only in `microseconds` function which could use the new constructor instead.
Also, the class documentation could mention that conversion to/from duration is supported.

eXpl0it3r:

include/SFML/System/Time.hpp
```diff
@@ -49,6 +50,13 @@ class Time
     ////////////////////////////////////////////////////////////
     constexpr Time();
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Construct from std::chrono::duration
+    ///
+    ////////////////////////////////////////////////////////////
+    template<class Rep, class Period>
```
Bromeon, line 57:
```suggestion
    template <typename Rep, typename Period>
```
for consistency. Same further below

include/SFML/System/Time.hpp
```diff
@@ -49,6 +50,13 @@ class Time
...
+    constexpr Time(const std::chrono::duration<Rep, Period>& duration);
```
Bromeon, line 58: Is that constructor not explicit on purpose?

Bromeon (commented):
Nice to have the conversions! I'm not sure if we should change the member variable though. What do we effectively gain? Code becomes quite a bit more verbose (mainly because `duration_cast` is such an ugly API). Does `sizeof(Time)` also increase?

include/SFML/System/Time.inl
```diff
@@ -30,22 +30,37 @@ m_microseconds(0)
 }
 
 
+////////////////////////////////////////////////////////////
+template<class Rep, class Period>
```
vittorioromeo, line 34:
```suggestion
template <typename Rep, typename Period>
```

include/SFML/System/Time.hpp
ChrisThrasher: That's right. We want to allow these implicit conversions.

include/SFML/System/Time.hpp
ChrisThrasher: I thought about that. I'm not entirely sure. We're using `to` in `sf::Color::toInteger` for what it's worth.

include/SFML/System/Time.hpp
ChrisThrasher: Fixed!

ChrisThrasher:
@kimci86

> The private constructor Time(Int64 microseconds) could be removed as it is used only in microseconds function which could use the new constructor instead.

Good point. I'll remove that.

> Also, the class documentation could mention that conversion to/from duration is supported.

Right I forgot that. I'll add some extra information to that section describing the new functionality. 

@Bromeon

> I'm not sure if we should change the member variable though. What do we effectively gain? Code becomes quite a bit more verbose (mainly because duration_cast is such an ugly API).

It makes it easier to remove magic numbers from the implementation and gives us the added assurance that `<chrono>`'s type safety brings. 

> Does sizeof(Time) also increase?

No I don't believe so. `std::chrono::microseconds` will use a signed integer with at least 55 bits which means any sensible implementation is using a `std::int64_t` which is definitionally the same as `sf::Int64`.

include/SFML/System/Time.hpp
kimci86: On second thought, maybe it is ok not to be consistent as conversion between units and types is not exactly the same idea.

ChrisThrasher:
I squashed some commits too since I thought 4 or 5 commits was excessive for a change of this scale.

include/SFML/System/Time.hpp
```diff
@@ -474,6 +482,9 @@ constexpr Time& operator %=(Time& left, Time right);
 ///
 /// sf::Time t3 = sf::microseconds(-800000);
 /// float sec = t3.asSeconds(); // -0.8
+///
+/// sf::Time t4 = std::chrono::milliseconds(250);
+/// std::chrono::microseconds milli2 = t4.asDuration(); // 250000us
```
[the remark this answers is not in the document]
ChrisThrasher: 🤦🏻‍♂️

Bromeon:
Implicit conversions between `sf::Time` and `std::chrono::duration` are probably OK, since both are already "duration" types and there's less that can go wrong than when implicitly converting to seconds/milliseconds. But I would say we make either both conversion constructor and operator implicit, or none, for symmetry.

ChrisThrasher:
I went ahead and added the implicit conversion to `std::chrono::duration`.

eXpl0it3r:

ChrisThrasher:
No changes. Just keeping up with `master`.

eXpl0it3r:

ChrisThrasher:
Rebased on master and fixed new compilation error introduced by a57640c2c89042d7ecc2c5f87712b553c8e14612

eXpl0it3r:

include/SFML/System/Time.inl
```diff
@@ -30,52 +30,68 @@ m_microseconds(0)
...
 {
+    return m_microseconds;
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr Time seconds(float amount)
 {
-    return Time(static_cast<Int64>(amount * 1000000));
+    return Time(std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::duration<float>(amount)));
```
Bromeon, line 80: Is this the most concise that constructing from seconds gets? I looked for some [other answers](https://stackoverflow.com/questions/17899684/convert-seconds-as-double-to-stdchronoduration), but they assume `duration<double>`.

The Chrono API is really awful... Who thought it was a good idea to keep `duration` generic? 😕

Bromeon (commented):
Looks mostly good to me, left small remarks.

Another thing that came to my mind: how good is debugger support now across compilers? Before, it was possible to see something like `sf::Time { m_microseconds: 323811 }` in debug watches and always know what's going on. Is the new internal `chrono::duration` field displayed as a deeply nested, barely decipherable type, or is there decent support for debug printing?

test/System/Time.cpp
```diff
@@ -3,6 +3,17 @@
...
 
+using namespace std::chrono_literals;
+
+namespace std
+{
+template <typename Rep, typename Period>
+std::ostream& operator<<(std::ostream& out, const std::chrono::duration<Rep, Period>& duration)
+{
+    return out << std::chrono::microseconds(duration).count() << "us";
+}
```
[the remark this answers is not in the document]
kimci86: I am not sure why, but by trying it I find that if we include <doctest.h> after declaring (or defining) the output operator, putting it in namespace std is not required.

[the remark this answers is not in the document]
kimci86: Apparently this is some C++ weirdness related to argument-dependent lookup.
This paper gives details. https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0934r0.pdf
In particular section 3.4 describes a similar situation, and section 4.1 explains why a global `operator<<` is generally not a good idea.
To my understanding, it is better as it is now in this PR.

vittorioromeo (approved):
Generally LGTM. Nice work! I'll leave final approval to @Bromeon

include/SFML/System/Time.inl
ChrisThrasher: Converting from floating point seconds to integer microseconds is lossy so the <chrono> API makes that look ugly. The syntax is much cleaner when lossy conversions aren't required. I suppose we could still do something like
```
return Time(std::chrono::microseconds(amount * 1'000'000));
```
but my goal was to remove all those magic numbers.

test/System/Time.cpp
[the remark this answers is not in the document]
kimci86: The third stringification alternative doctest supports seems to have none of the drawbacks we noticed.
```c++
namespace doctest
{
    template <typename Rep, typename Period>
    struct StringMaker<std::chrono::duration<Rep, Period>>
    {
        static String convert(const std::chrono::duration<Rep, Period>& duration)
        {
            return toString(std::chrono::microseconds(duration).count()) + "us";
        }
    };
}
```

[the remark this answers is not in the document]
ChrisThrasher: Oh yeah idk why I looked past that. I'll let @Bromeon decide if we care about avoiding `namespace std`.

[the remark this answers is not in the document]
Bromeon: > The third stringification alternative doctest supports seems to have none of the drawbacks we noticed.

If doctest provides a way that works without drawbacks, should we go for this one, just to be on the safe side? Maybe along with a comment why we don't use `operator<<` like everywhere else.

But I can definitely also live with the `std` solution, if you find that nicer. I would add a comment in that case as well.

Bromeon:
This is still unaddressed: 

> Another thing that came to my mind: how good is debugger support now across compilers? Before, it was possible to see something like sf::Time { m_microseconds: 323811 } in debug watches and always know what's going on. Is the new internal chrono::duration field displayed as a deeply nested, barely decipherable type, or is there decent support for debug printing?

Because as mentioned in the early review, I don't see a practical advantage of storing `std::duration` internally -- yes, we remove magic numbers (which are quite obvious in that context), but introduce very hard-to-read chrono shenanigans. So let's at least make sure that usability of `sf::Time` doesn't deteriorate.

include/SFML/System/Time.inl
Bromeon: I see, thanks for the explanation. It's weird to me that one of the universally most common operations of any Time API -- converting between seconds and durations -- is made so extremely hard. But it's maybe a case of APIs designed by people who don't need them for everyday productive use 😬

Anyway I'm glad we still keep the `sf::Time` class, it's very nice to work with.

ChrisThrasher:
> but introduce very hard-to-read chrono shenanigans

To be fair, this is idiomatic `<chrono>` code. However verbose the API is, anyone familiar with C++11 time utilities will know what is going on.

> So let's at least make sure that usability of sf::Time doesn't deteriorate.

image

In the VS Code debugger I can still easily see the value of the duration.

include/SFML/System/Time.inl
ChrisThrasher: `sf::Time` is ergonomic because it's less correct. If we removed lossy conversions from `sf::Time` and made accept a generic type as the underlying representation, we'd quickly approach an interface as verbose as `<chrono>`. If you really want shorter code, we can liberally apply `using name std::chrono` within select scopes to minimize the character count.

Bromeon (approved):
The debug view has one extra level of indirection than before, but that's OK. In VS 2019, it's even better:

grafik



> To be fair, this is idiomatic `<chrono>` code. However verbose the API is, anyone familiar with C++11 time utilities will know what is going on.

True, since it's such a central API, and relatively simple (functionality-wise, not ergonomics). But I wouldn't assume the same about any C++ feature 😉

test/System/Time.cpp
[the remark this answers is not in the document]
ChrisThrasher: I went ahead and used the option that doesn't requiring opening `namespace std`. I went ahead and did the same thing for the `std::vector` printer in test/Graphics/Transform.cpp since I'm not positive that that's legal either.

ChrisThrasher:
I changed how we're printing standard types such that we never have to open up `namespace std` since I'm not 100% sure that's legal.

vittorioromeo:

ChrisThrasher:
> Now it only needs a small comment to explain why we use StringMaker for stringification.

Done

vittorioromeo:

vittorioromeo:

ChrisThrasher:

ChrisThrasher:
With Bromeon's and Vittorio's approval, is there anything else we're waiting on before this is good to merge?

ChrisThrasher:
Rebased on master, fixed a compiler warning caused by Doctest changes, and found somewhere to use this new implicit `chrono::duration`->`sf::Time` conversion to simplify code.

include/SFML/System/Time.hpp
```diff
@@ -78,6 +87,23 @@ class Time
...
+    constexpr std::chrono::microseconds toDuration() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Implicit conversion to std::chrono::duration
+    ///
+    /// \return Duration in microseconds
+    ///
+    ////////////////////////////////////////////////////////////
+    template <typename Rep, typename Period>
+    [[nodiscard]] constexpr operator std::chrono::duration<Rep, Period>() const;
```
Bromeon, line 105: Out of curiosity regarding `[[nodiscard]]`: is it even possible to invoke this operator _without_ using the result? Since its whole point is to be called when `sf::Time` is used in a context that would allow the conversion.

That is, apart from the academic `time.operator std::chrono::duration<Rep, Period>()`.

include/SFML/System/Time.inl
Bromeon: Forgot to answer here (and it's not direct feedback to this PR anyway, more interested in the discussion 🙂)

> If we removed lossy conversions from `sf::Time` and made accept a generic type as the underlying representation, we'd quickly approach an interface as verbose as `<chrono>`.

This is exactly my point: making `chrono::duration` generic was a design mistake in my opinion. No other language has generic durations and especially "business software" languages like Java need to deal _a lot_ with time (it also took them several attempts to get things finally right, but the current `Instant`/`Duration` API is pretty decent).

I don't know what the premise was, maybe someone thought the API must be capable of handling simulations of the Large Hadron Collider. I find it sometimes a bit sad that so many C++ APIs are designed to fit the 1% use cases while making the 99% extra hard 😕

test/Graphics/Transform.cpp
```diff
@@ -6,17 +6,22 @@
...
+struct StringMaker<std::vector<float>>
 {
-    out << "{ ";
-    for (size_t i = 0; i + 1 < vector.size(); ++i)
-        out << vector[i] << ", ";
-    out << vector.back() << " }";
-    return out;
-}
-} // namespace std
+    static String convert(const std::vector<float> vector)
```
Bromeon, line 15: You probably meant `const std::vector<float>& vector` here (const-reference).

include/SFML/System/Time.hpp
ChrisThrasher: I can't think of any such scenario.

test/Graphics/Transform.cpp
ChrisThrasher: Thanks!

ChrisThrasher:
Added a missing `&` on a parameter and rebased on master.

include/SFML/System/Time.hpp
Bromeon: Then maybe remove it, as it's more confusing to the code reader if it doesn't have a real purpose?

ChrisThrasher: I don't think it would ever result in a net negative to the user. I much prefer liberally applying `[[nodiscard]]` but I know you prefer omitting it in many instances so I'll get rid of it here.

eXpl0it3r:

include/SFML/System/Time.hpp
Bromeon: It's more a case of reducing clutter where it doesn't add useful information. If there are fewer instances of `[[nodiscard]]`, people will pay more attention to them and understand that those are potential pitfalls.

ChrisThrasher:
@Bromeon would you mind checking this out?

test/Graphics/Transform.cpp
```diff
@@ -6,17 +6,22 @@
...
-    for (size_t i = 0; i + 1 < vector.size(); ++i)
-        out << vector[i] << ", ";
-    out << vector.back() << " }";
-    return out;
-}
-} // namespace std
+    static String convert(const std::vector<float>& vector)
+    {
+        doctest::String out = "{ ";
+        for (size_t i = 0; i + 1 < vector.size(); ++i)
```
Bromeon, line 18: Great attention to detail, avoiding `i < vector.size() - 1`! 👍

Bromeon (approved):
Looks good to me!

There was no real conclusion about `asDuration`/`toDuration`, but it's something that could easily be changed later without blocking this PR. Maybe as we design some other APIs, certain patterns will come up.

test/Graphics/Transform.cpp
ChrisThrasher: I actually did that for style reasons but now I'm realizing that if `size()` returns `0` then _ouch_. I've never thought of that failure mode before!

kimci86: As a side note, an empty vector would fail two lines later, with `vector.back()`. Fortunately we do not print empty vectors.

Bromeon: That's ironic, it was _precisely_ the style that made me think "why the hell does he not compare the loop variable directly" 😅 

Do you think it's worth a comment regarding underflow, or is it clear enough?

Bromeon: @kimci86 Good catch!

Maybe `if (vector.empty())` special case + early return? Then we could also use _normal_™ style 😛 

If the assumption is "we never print empty vectors", we could also do `assert(!vector.empty())` -- better than UB and potentially flaky tests for sure.

ChrisThrasher:
```cpp
        if (vector.empty())
            return "{ }";
        doctest::String out = "{ ";
        for (size_t i = 0; i + 1 < vector.size(); ++i)
            out += toString(vector[i]) + ", ";
        out += toString(vector.back()) + " }";
        return out;
```
A guard clause seems like the easiest option if we want to catch empty vectors

ChrisThrasher: Seems like we're currently printing no empty vectors so we can get away with an assert and just wait for this to maybe become a problem later. I think that's what I'd prefer for now just to be safe and without us spending too much time on this.

ChrisThrasher:
Rebased on master and added an `assert` to handle a potential edge case that could cause UB in the tests.

kimci86 (approved):

eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:
Woho! Thanks for this addition 🥳

PierceLBrooks:
This change may be related to a a master branch compilation issue I am now seeing on Windows:
```
C:\UserData\SFML\include\SFML/System/Time.inl(223,42): error C2039: 'int64_t': is not a
member of 'std' [C:\UserData\SFML\build\src\SFML\System\sfml-system.vcxproj]
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.28.29333\include\chrono(24): message : s
ee declaration of 'std' [C:\UserData\SFML\build\src\SFML\System\sfml-system.vcxproj]
```

ChrisThrasher:
Very possible there's simply a missing `#include <cstdint>` somewhere. Can you share how you can reproduce this error?

PierceLBrooks:
@ChrisThrasher Of course! My steps were as follows:
```
git clone https://github.com/SFML/SFML.git
cd SFML
git checkout master
mkdir build
cd build
cmake ..
cmake --build .
```
However, this assumes that Visual Studio 2019 will be the default pick of your system, so you may need to specify the appropriate `-G` flag for CMake.

ChrisThrasher:
That's it? Weird, because we already test VS 2019 in CI. Anyways, should still be an issue fix. Mind filing an issue about this so we don't lose track of this bug? Thanks!

PierceLBrooks:
Will do, I can try to address it with a pull request later on as well if I can track down where the `cstdint` include belongs! FYI, checking out the 2.5.1 release version branch allows the build to succeed just fine.

PierceLBrooks:
https://github.com/SFML/SFML/issues/2211#issue-1381582926

PR #2134, opened 2022-06-18
Add tests for `sf::IpAddress`

ChrisThrasher:
## Description

Finally trying out some Network module tests. I updated Doctest because without doing that I was getting linker errors in macOS Release builds. I got the new Doctest header from [here](https://github.com/doctest/doctest/releases/tag/v2.4.9).

While staring at this code I also find some other enhancements I'd like to make in a later PR.

https://github.com/ChrisThrasher/SFML/tree/ipaddress_enhancements

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

### DIFF
--- a/test/Network/IpAddress.cpp
+++ b/test/Network/IpAddress.cpp
@@ -1,0 +1,226 @@
+#include <SFML/Network/IpAddress.hpp>
+
+#include <doctest.h>
+
+#include <sstream>
+
+using namespace std::string_literals;
+
+TEST_CASE("sf::IpAddress class - [network]")
+{
+    SUBCASE("Construction")
+    {
+        SUBCASE("Default constructor")
+        {
+            const sf::IpAddress ipAddress;
+            CHECK(ipAddress.toString() == "0.0.0.0"s);
+            CHECK(ipAddress.toInteger() == 0);
+            CHECK(ipAddress == sf::IpAddress::None);
+        }
+
+        SUBCASE("std::string constructor")
+        {
+            const sf::IpAddress ipAddress = "192.168.0.1"s;
+            CHECK(ipAddress.toString() == "192.168.0.1"s);
+            CHECK(ipAddress.toInteger() == 0xC0A80001);
+            CHECK(ipAddress != sf::IpAddress::None);
+            CHECK(ipAddress != sf::IpAddress::Any);
+            CHECK(ipAddress != sf::IpAddress::Broadcast);
+            CHECK(ipAddress != sf::IpAddress::LocalHost);
+
+            const sf::IpAddress broadcast("255.255.255.255"s);
+            CHECK(broadcast.toString() == "255.255.255.255"s);
+            CHECK(broadcast.toInteger() == 0xFFFFFFFF);
+            CHECK(broadcast == sf::IpAddress::Broadcast);
+
+            const sf::IpAddress any("0.0.0.0"s);
+            CHECK(any.toString() == "0.0.0.0"s);
+            CHECK(any.toInteger() == 0x00000000);
+            CHECK(any == sf::IpAddress::Any);
+
+            const sf::IpAddress localHost("localhost"s);
+            CHECK(localHost.toString() == "127.0.0.1"s);
+            CHECK(localHost.toInteger() == 0x7F000001);
+            CHECK(localHost == sf::IpAddress::LocalHost);
+
+            const sf::IpAddress invalidIpAddress("255.255.255.256"s);
+            CHECK(invalidIpAddress.toString() == "0.0.0.0"s);
+            CHECK(invalidIpAddress.toInteger() == 0);
+            CHECK(invalidIpAddress == sf::IpAddress::None);
+        }
+
+        SUBCASE("const char* constructor")
+        {
+            const sf::IpAddress ipAddress = "192.168.0.1";
+            CHECK(ipAddress.toString() == "192.168.0.1"s);
+            CHECK(ipAddress.toInteger() == 0xC0A80001);
+            CHECK(ipAddress != sf::IpAddress::None);
+
+            const sf::IpAddress broadcast("255.255.255.255");
+            CHECK(broadcast.toString() == "255.255.255.255"s);
+            CHECK(broadcast.toInteger() == 0xFFFFFFFF);
+            CHECK(broadcast == sf::IpAddress::Broadcast);
+
+            const sf::IpAddress any("0.0.0.0");
+            CHECK(any.toString() == "0.0.0.0"s);
+            CHECK(any.toInteger() == 0x00000000);
+            CHECK(any == sf::IpAddress::Any);
+
+            const sf::IpAddress localHost("localhost");
+            CHECK(localHost.toString() == "127.0.0.1"s);
+            CHECK(localHost.toInteger() == 0x7F000001);
+            CHECK(localHost == sf::IpAddress::LocalHost);
+
+            const sf::IpAddress invalidIpAddress("255.255.255.256");
+            CHECK(invalidIpAddress.toString() == "0.0.0.0"s);
+            CHECK(invalidIpAddress.toInteger() == 0);
+            CHECK(invalidIpAddress == sf::IpAddress::None);
+        }
+
+        SUBCASE("Byte constructor")
+        {
+            const sf::IpAddress ipAddress(142, 250, 69, 238);
+            CHECK(ipAddress.toString() == "142.250.69.238"s);
+            CHECK(ipAddress.toInteger() == 0x8EFA45EE);
+            CHECK(ipAddress != sf::IpAddress::None);
+        }
+
+        SUBCASE("Uint32 constructor")
+        {
+            const sf::IpAddress ipAddress(0xDEADBEEF);
+            CHECK(ipAddress.toString() == "222.173.190.239"s);
+            CHECK(ipAddress.toInteger() == 0xDEADBEEF);
+            CHECK(ipAddress != sf::IpAddress::None);
+        }
+    }
+
+    SUBCASE("Static functions")
+    {
+        SUBCASE("getLocalAddress")
+        {
+            const sf::IpAddress ipAddress = sf::IpAddress::getLocalAddress();
+            CHECK(ipAddress.toString() != "0.0.0.0");
+            CHECK(ipAddress.toInteger() != 0);
+            CHECK(ipAddress != sf::IpAddress::None);
+        }
+
+        SUBCASE("getPublicAddress")
+        {
+            const sf::IpAddress ipAddress = sf::IpAddress::getPublicAddress();
+            CHECK(ipAddress.toString() != "0.0.0.0");
+            CHECK(ipAddress.toInteger() != 0);
+            CHECK(ipAddress != sf::IpAddress::None);
+        }
+    }
+
+    SUBCASE("Static constants")
+    {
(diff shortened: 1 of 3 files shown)


ChrisThrasher:
I tried to best test the static functions but I'm not sure how to more rigorously test those. Please let me know if we can improve the coverage on those functions.

eXpl0it3r:

test/Network/IpAddress.cpp
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        SUBCASE("operator<")
+        {
+            CHECK(sf::IpAddress(1) < sf::IpAddress(2));
+            CHECK(sf::IpAddress() < sf::IpAddress(0, 0, 0, 0));
+            CHECK(sf::IpAddress(0, 0, 0, 0) < sf::IpAddress(1, 0, 0, 0));
+            CHECK(sf::IpAddress(1, 0, 0, 0) < sf::IpAddress(0, 1, 0, 0));
+            CHECK(sf::IpAddress(0, 1, 0, 0) < sf::IpAddress(0, 0, 1, 0));
+            CHECK(sf::IpAddress(0, 0, 1, 0) < sf::IpAddress(0, 0, 0, 1));
```
kimci86, line 153: Isn't it surprising? I would have expected the opposite, the lexicographical order.
This current behavior is caused by the fact that the address is stored in network byte order (big-endian) in `m_address` and used like that in `operator<`.
A quick fix to get lexicographical ordering, if we agree this is wanted, would be to use the address converted in host byte order calling `ntohl` in `operator<`, but I think it would make more sense to store the address in host byte order instead.

ChrisThrasher: It definitely surprised me. I started writing the tests the other way around and was worried I found a bug when it didn't behave like I expected.

Bromeon: I'm with @kimci86 here and don't see a reason to sort anti-lexicographically 🙂 
Also, SFML 3 would be the place to fix this, as it's quite a bit of a breaking change, even if rarely encountered.

test/Network/IpAddress.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+            const sf::IpAddress localHost("localhost"s);
+            CHECK(localHost.toString() == "127.0.0.1"s);
+            CHECK(localHost.toInteger() == 0x7F000001);
+            CHECK(localHost == sf::IpAddress::LocalHost);
+
+            const sf::IpAddress invalidIpAddress("255.255.255.256"s);
+            CHECK(invalidIpAddress.toString() == "0.0.0.0"s);
+            CHECK(invalidIpAddress.toInteger() == 0);
+            CHECK(invalidIpAddress == sf::IpAddress::None);
```
Bromeon, line 46: Side note: this might also a bit unexpected, but I don't really see how we can fix it, as long as error handling is not sorted out. The SFML 2 resource approach of `sf::Image im; if (!im.loadFromY(...))` is too clumsy for such a simple type. There are a few options (`optional`, exceptions) but we should probably postpone that until we find a general way to address error handling.

test/Network/IpAddress.cpp
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        SUBCASE("std::string constructor")
+        {
+            const sf::IpAddress ipAddress = "192.168.0.1"s;
+            CHECK(ipAddress.toString() == "192.168.0.1"s);
+            CHECK(ipAddress.toInteger() == 0xC0A80001);
+            CHECK(ipAddress != sf::IpAddress::None);
+
+            const sf::IpAddress broadcast("255.255.255.255"s);
```
Bromeon, line 28: Any reason why the previous constructor call is "implicit" (with `var = value` syntax) and this one "explicit" (with `var(value)` syntax)?

test/Network/IpAddress.cpp
ChrisThrasher: Want to fix that in this same PR or save it for later?

ChrisThrasher: What is `operator<` used for in this context besides sorting in `std::map`? If that's the main use case, the implementation of `operator<` doesn't really matter and most Network module users will never notice.

test/Network/IpAddress.cpp
```diff
@@ -0,0 +1,223 @@
...
+            CHECK(sf::IpAddress(0, 0, 1, 0) >= sf::IpAddress(0, 1, 0, 0));
+            CHECK(sf::IpAddress(0, 0, 0, 1) >= sf::IpAddress(0, 0, 1, 0));
+            CHECK(sf::IpAddress(1, 0, 0, 1) >= sf::IpAddress(0, 0, 0, 1));
+
+            CHECK(sf::IpAddress() >= sf::IpAddress());
+            CHECK(sf::IpAddress("") >= sf::IpAddress(""s));
+            CHECK(sf::IpAddress("8.8.8.8") >= sf::IpAddress(8, 8, 8, 8));
+            CHECK(sf::IpAddress(0x42, 0x69, 0x96, 0x24) >= sf::IpAddress(0x42699624));
+            CHECK(sf::IpAddress(0xABCDEF01) >= sf::IpAddress("171.205.239.1"));
+        }
```
Bromeon, line 200: I have the feeling that we might want to test all relational operators in a generic way in the future. This would reduce the amount of test code to write and enforce internal consistency between the operators.

Something like this:
```cpp
#define CHECK_LESS(lower, upper) \
    CHECK(lower < upper);        \
    CHECK(upper > lower);        \
    CHECK(!(lower >= upper));    \
    CHECK(!(upper <= lower));
```

test/Network/IpAddress.cpp
ChrisThrasher: This is a textbook use case for exceptions since signaling constructor failure is very hard through other means. I'm of the opinion that exceptions are a good fit for SFML but I know the team made the very specific reason not to use them so I see no better way to signal failure. Maybe in these cases we also write to `sf::err()` to signal an error occurred? That would be in line with other "might fail" APIs.

test/Network/IpAddress.cpp
ChrisThrasher: `var(value)` is what all SFML tests tend to use by default but I added a few exceptional `var = value` uses to make sure we test the implicit-ness of these constructors.

test/Network/IpAddress.cpp
Bromeon: What @vittorioromeo mentioned is btw a very common idiom in Rust. There are no constructors in the language (and no exceptions), just "static factory" methods, which return either the object itself or a wrapper object for fallible construction:
* `Option<T>` which is pretty much 1:1 `std::optional`
* `Result<T, E>` which contains either a value or an error (there's a [proposal for `std::expected`](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4015.pdf)).

Of course, the situation is different because the language was designed with those in mind and it's thus "idiomatic" by definition, while in C++, fallible factory methods are just one of many approaches. I don't mean this as a way to say Rust is right here, more that there's prior art for this idiom, which has worked well on a large scale.

I think this topic needs more discussion outside this PR; we can also leave the error handling as-is for now, until we discussed the topic more thoroughly.

test/Network/IpAddress.cpp
ChrisThrasher: It ought to be a macro to preserve line number information.

test/Network/IpAddress.cpp
ChrisThrasher: > We cannot easily signal why the conversion failed

A good short term solution for a later PR is to add some `sf::err` logging. We can add unit tests for this as well. I implemented this locally and it works just fine.

test/Network/IpAddress.cpp
Bromeon: Sounds great, thanks @ChrisThrasher!

I think macros have too bad of a reputation. Yes, they shouldn't be used for `MAX(a, b)` and other cases where existing abstraction mechanisms work just fine, but they have their niche where nothing comes close.

test/Network/IpAddress.cpp
```diff
@@ -0,0 +1,223 @@
...
+            CHECK(ipAddress.toInteger() == 0);
+            CHECK(ipAddress != sf::IpAddress::Any);
+        }
+
+        SUBCASE("std::string constructor")
+        {
+            const sf::IpAddress ipAddress = "192.168.0.1"s;
+            CHECK(ipAddress.toString() == "192.168.0.1"s);
+            CHECK(ipAddress.toInteger() == 0xC0A80001);
+            CHECK(ipAddress != sf::IpAddress::None);
```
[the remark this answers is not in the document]
ChrisThrasher: As in confirm inequality with `Any` and `Broadcast`?

test/Network/IpAddress.cpp
```diff
@@ -0,0 +1,223 @@
...
+TEST_CASE("sf::IpAddress class - [network]")
+{
+    SUBCASE("Construction")
+    {
+        SUBCASE("Default constructor")
+        {
+            const sf::IpAddress ipAddress;
+            CHECK(ipAddress.toString() == "0.0.0.0"s);
+            CHECK(ipAddress.toInteger() == 0);
+            CHECK(ipAddress != sf::IpAddress::Any);
```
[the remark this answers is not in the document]
ChrisThrasher: Because default-constructed IP addresses are invalid yet still report an address of 0.0.0.0. Maybe I should check equality with `sf::IpAddress::None` instead.

test/Network/IpAddress.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+            const sf::IpAddress broadcast("255.255.255.255"s);
+            CHECK(broadcast.toString() == "255.255.255.255"s);
+            CHECK(broadcast.toInteger() == 0xFFFFFFFF);
+            CHECK(broadcast == sf::IpAddress::Broadcast);
+
+            const sf::IpAddress any("0.0.0.0"s);
+            CHECK(any.toString() == "0.0.0.0"s);
+            CHECK(any.toInteger() == 0x00000000);
+            CHECK(any == sf::IpAddress::Any);
```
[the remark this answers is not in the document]
ChrisThrasher: Hahah yes these tests were cumbersome to write because the implementation often behaved in ways I was not expecting

test/Network/IpAddress.cpp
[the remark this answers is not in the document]
vittorioromeo: > As in confirm inequality with `Any` and `Broadcast`?

Yes, seems weird that you're only testing specifically against `None` but not the others.

test/Network/IpAddress.cpp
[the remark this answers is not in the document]
vittorioromeo: > Maybe I should check equality with `sf::IpAddress::None` instead.

I think that would be good

ChrisThrasher:
Addressed a few review comments and rebased on master.

test/Network/IpAddress.cpp
Bromeon: It can be used for anything, like `std::sort()`, `std::binary_search()`, etc.

It probably does not matter in many cases, but if it's easy to fix, we might as well make it more intuitive for the few people where it does matter. I would say fixes to `sf::IpAddress` belong to a separate PR.

Bromeon (approved):

vittorioromeo (approved):

eXpl0it3r:

PR #2137, opened 2022-06-21
'sf::Image::copy' returns a 'bool' to signal success/failure

vittorioromeo:
Follow-up to #2117, I made `sf::Image::copy` return a boolean as the function can fail. CC @Bambo-Borris


vittorioromeo:

PR #2141, opened 2022-06-23
Replace 'TcpSocket::send' local buffer with data member

vittorioromeo:
As the title says, this PR replaces a local `blockToSend` variable that resulted in a dynamic memory allocation on every call to `sf::TcpSocket::send` with a data member that can be reused between multiple calls. It aims to be a run-time performance optimization and to reduce the number of dynamic allocations.

**This PR is untested.** I would like to hear some thoughts on the change, and want to know if I am missing something *(is there a reason why it was not a data member before?)* and the best way to test it.

The only possible case where this change might break something is if `sf::TcpSocket::send` is meant to be concurrently called by multiple threads at the same time. Is that our expectation for this API? Seems unlikely...

### DIFF
--- a/include/SFML/Network/TcpSocket.hpp
+++ b/include/SFML/Network/TcpSocket.hpp
@@ -228,7 +228,8 @@ private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    PendingPacket m_pendingPacket; //!< Temporary data of the packet currently being received
+    PendingPacket     m_pendingPacket;     //!< Temporary data of the packet currently being received
+    std::vector<char> m_blockToSendBuffer; //!< Buffer used to prepare data being sent from the socket
 };
 
 } // namespace sf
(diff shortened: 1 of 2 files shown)


Bromeon (approved):
Is `TcpSocket` thread-safe at all? It's not advertised in the docs, so I'd expect it not to be.

The one potential disadvantage this change brings is that when someone has loads of `TcpSocket` instances open and uses them very rarely, but sends a lot of data. In that case, the new implementation keeps a big amount of unused memory allocated.

binary1248:
If TcpSocket were to be used simultaneously from multiple threads it would have also broke before. When using the Packet API it is the user's responsibility to check the return value to know if they need to retry sending the packet if it couldn't be completely sent in the first call.

> The one potential disadvantage this change brings is that when someone has loads of TcpSocket instances open and uses them very rarely, but sends a lot of data. In that case, the new implementation keeps a big amount of unused memory allocated.

This change only affects the Packet API. If the user really wanted to send huge blocks of data, I don't think they would be using the Packet API for that. Anything less than 64KiB is insignificant by today's standards, especially since it isn't recommended to open more connections than necessary anyway.

eXpl0it3r (approved):

Issue #2138, opened 2022-06-21
Overhaul `sf::IpAddress`


vittorioromeo:
So apparently these tests pass:

```cpp
const sf::IpAddress ipAddress;
CHECK(ipAddress.toString() == "0.0.0.0"s);
CHECK(ipAddress.toInteger() == 0);
CHECK(ipAddress != sf::IpAddress::Any);

const sf::IpAddress any("0.0.0.0"s);
CHECK(any.toString() == "0.0.0.0"s);
CHECK(any.toInteger() == 0x00000000);
CHECK(any == sf::IpAddress::Any);
```

This is extremely worrisome to me -- because there's a hidden "invalid" state that cannot be distinguished from `toString` and `toInteger`.

So here's my proposal:

1. Rename `sf::IpAddress` to `sf::IpV4Address`. This can be done as a pure renaming refactoring PR.
2. `sf::IpV4Address` is an always-valid value semantic class. It's just a wrapper around a single `sf::Uint32`.
3. To create `sf::IpV4Address` instances, a factory function returning `std::optional<sf::IpV4Address>` must be used. The constructor of `sf::IpV4Address` is private.

Thoughts? @ChrisThrasher @eXpl0it3r @Bromeon

ChrisThrasher:
I'd most prefer we just throw during construction. This is impossible for users to ignore, trivial to test, and maintains the clean constructor syntax that C++ developers expect. By returning an optional we're just imposing the error handling on the user who still has to correctly handle that optional or risk throwing a `std::bad_optional_access`. At least if we throw our own exception we can pick the error message such that it's as clear as possible why construction failed. This also means we get to maintain the other constructors like `sf::IpAddress(sf::Uint32)` which cannot fail without creating a weird split between fallible construction methods and infallible construction methods. 

I know this breaks with SFML convention but maybe we reconsider this design decision because as illustrated here, it's causing us to write messier interfaces. 

Exceptions aside, my 2nd favorite option would be `std::optional<sf::IpAddress> sf::IpAddress::create(/* all constructor overloads here */)` and remove all existing constructors, even those that cannot fail, for the sake of consistency. 

---

Unsure why we want to specify that these are IPv4 addresses unless we plan to add IPv6 addresses in the future. I suppose it's future proofing in case `sf::IpV6Address` gets added in a later 3.x minor release.

---

> To prevent a physical dependency between sf::IpV4Address and <string>/<string_view>/<optional>,...

[...]

PR #2145, opened 2022-06-23
`sf::IpAddress` is always valid, resolution is performed via static member function

vittorioromeo:

----

## Description

This PR is related to the issue #2138.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI and examples.


test/Network/IpAddress.cpp
```diff
@@ -3,121 +3,86 @@
...
+            REQUIRE(localHost.has_value());
+            CHECK(localHost->toString() == "127.0.0.1"s);
+            CHECK(localHost->toInteger() == 0x7F000001);
+            CHECK(*localHost == sf::IpAddress::LocalHost);
+
+            const auto invalidIpAddress = sf::IpAddress::create("255.255.255.256"s);
+            CHECK(!invalidIpAddress.has_value());
+
+            const auto invalidIpAddress2 = sf::IpAddress::create("");
+            CHECK(!invalidIpAddress2.has_value());
```
ChrisThrasher, line 47:
```suggestion
            CHECK(!sf::IpAddress::create("255.255.255.256"s).has_value());
            CHECK(!sf::IpAddress::create("").has_value());
```

src/SFML/Network/Http.cpp
```diff
@@ -369,7 +369,7 @@ Http::Response Http::sendRequest(const Http::Request& request, Time timeout)
     Response received;
 
     // Connect the socket to the host
-    if (m_connection.connect(m_host, m_port, timeout) == Socket::Done)
+    if (m_connection.connect(m_host.value(), m_port, timeout) == Socket::Done)
```
ChrisThrasher, line 372: Do we want to check this contains a value to avoid risking a `std::bad_optional_access`? Maybe just `assert` that it has a value?

include/SFML/Network/IpAddress.hpp
```diff
@@ -32,6 +32,7 @@
 #include <SFML/System/Time.hpp>
 #include <iosfwd>
 #include <optional>
+#include <string_view>
 #include <string>
```
ChrisThrasher, line 36: I think we can remove `<string>`.

ChrisThrasher (changes requested):

include/SFML/Network/IpAddress.hpp
```diff
@@ -43,47 +44,40 @@ namespace sf
...
     /// (ex: "192.168.1.56") or a network name (ex: "localhost").
-    /// This is equivalent to the constructor taking a std::string
-    /// parameter, it is defined for convenience so that the
-    /// implicit conversions from literal strings to IpAddress work.
     ///
     /// \param address IP address or network name
     ///
     ////////////////////////////////////////////////////////////
-    IpAddress(const char* address);
+    static std::optional<IpAddress> create(std::string_view address);
```
Bromeon, line 73: If we use the [Named Constructor Idiom](https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Named_Constructor), should we not give it a different name, like `fromString()`?

ChrisThrasher:
https://github.com/SFML/SFML/blob/d7f09c1167edb7e16c9aa325eec2d925c8f59444/include/SFML/Network/UdpSocket.hpp#L260

Here's another reference to the old `getLocalAddress()` API.

include/SFML/Network/IpAddress.hpp
vittorioromeo: I am up for suggestions. `resolve` was also another idea, but `fromString` sounds good to me. @eXpl0it3r, @ChrisThrasher, thoughts?

include/SFML/Network/IpAddress.hpp
```diff
@@ -301,8 +294,6 @@ SFML_NETWORK_API std::ostream& operator <<(std::ostream& stream, const IpAddress
 ///
 /// Usage example:
 /// \code
-/// sf::IpAddress a0;                                     // an invalid address
-/// sf::IpAddress a1 = sf::IpAddress::None;               // an invalid address (same as a0)
 /// sf::IpAddress a2("127.0.0.1");                        // the local host address
```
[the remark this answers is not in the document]
vittorioromeo: Will update.

include/SFML/Network/IpAddress.hpp
ChrisThrasher: `resolve` is my favorite name so far.

Bromeon: "resolve" sounds a bit like an actual connection to the IP address is established.

binary1248: I don't know... if one isn't familiar with how names are _resolved_ to addresses during name resolution, maybe that is something to read into before starting with network programming. 😉

binary1248: Also, calling it `fromString` is misleading. The almost exclusively used dotted decimal notation is the de facto standard way of representing an IPv4 address. The user might think that `fromString` can only be used to convert an IPv4 address in its textual representation to an `sf::IpAddress`. The function in question performs a complete name resolution.

src/SFML/Network/IpAddress.cpp
```diff
@@ -37,68 +37,69 @@
...
-IpAddress::IpAddress(const std::string& address)
-{
-    if (address == "255.255.255.255")
+    if (address == "255.255.255.255"sv)
     {
         // The broadcast address needs to be handled explicitly,
         // because it is also the value returned by inet_addr on error
-        m_address = INADDR_BROADCAST;
-        return;
+        return IpAddress(NoByteOrderConversionTag{}, static_cast<Uint32>(INADDR_BROADCAST));
```
binary1248, line 59:
```suggestion
        return Broadcast;
```

src/SFML/Network/IpAddress.cpp
```diff
@@ -37,68 +37,69 @@
...
     }
 
-    if (address == "0.0.0.0")
+    if (address == "0.0.0.0"sv)
     {
-        m_address = INADDR_ANY;
-        return;
+        return IpAddress(NoByteOrderConversionTag{}, static_cast<Uint32>(INADDR_ANY));
```
binary1248, line 64:
```suggestion
        return Any;
```

src/SFML/Network/IpAddress.cpp
```diff
@@ -37,68 +37,69 @@
...
         std::memcpy(&sin, result->ai_addr, sizeof(*result->ai_addr));
-        ip = sin.sin_addr.s_addr;
+
+        const Uint32 ip = sin.sin_addr.s_addr;
         freeaddrinfo(result);
-        m_address = ip;
-        return;
+
+        return IpAddress(NoByteOrderConversionTag{}, ip);
     }
```
binary1248, line 87: Is the overload with `NoByteOrderConversionTag` necessary? It is only used internally yet requires being part of the externally visible API. Since this functionality is required from within the class itself, couldn't one just create a value that is initialized to e.g. `sf::IpAddress::Any` and manually set m_value in it? Any compiler that is reasonable at optimizing should be able to combine the statements into a single operation.

binary1248 (changes requested):

eXpl0it3r:

include/SFML/Network/IpAddress.hpp
vittorioromeo: Will change the name to `resolve`.

include/SFML/Network/IpAddress.hpp
vittorioromeo: Will double-check and remove if possible.

vittorioromeo: It is necessary for `toString`.

vittorioromeo:

vittorioromeo:
@binary1248: please take one more look, I removed the private constructor.

src/SFML/Network/IpAddress.cpp
```diff
@@ -37,68 +37,62 @@
...
 ////////////////////////////////////////////////////////////
-IpAddress::IpAddress() = default;
+std::optional<IpAddress> IpAddress::resolve(std::string_view address)
+{
+    using namespace std::string_view_literals;
 
+    if (address.empty())
+    {
+        return std::nullopt;
+    }
```
binary1248, line 53:
```suggestion
    if (address.empty())
        return std::nullopt;
```

src/SFML/Network/IpAddress.cpp
```diff
@@ -37,68 +37,62 @@
...
+        return Broadcast;
     }
 
-    if (address == "0.0.0.0")
+    if (address == "0.0.0.0"sv)
     {
-        m_address = INADDR_ANY;
-        return;
+        return Any;
     }
```
binary1248, line 65: Single line `if` doesn't require braces.

src/SFML/Network/IpAddress.cpp
```diff
@@ -37,68 +37,62 @@
...
 
     // Try to convert the address as a byte representation ("xxx.xxx.xxx.xxx")
-    sf::Uint32 ip = inet_addr(address.c_str());
-    if (ip != INADDR_NONE)
+    if (const Uint32 ip = inet_addr(address.data()); ip != INADDR_NONE)
     {
-        m_address = ip;
-        return;
+        return IpAddress(ntohl(ip));
     }
```
binary1248, line 71: Same as above.

binary1248 (changes requested):

vittorioromeo:
@binary1248: applied the requested formatting changes.

vittorioromeo:

src/SFML/Network/Http.cpp
ChrisThrasher: Was there ever a conclusion made about this or is it not possible for `.value()` to be called on a null optional here?

PR #2147, opened 2022-06-23
Explicitly instantiate floating-point--only `sf::Vector2` and `sf::Vector3` members in a source file

vittorioromeo:

----

## Description

This PR removes the `<cmath>` and `<type_traits>` includes from `Vector2.hpp` and `Vector3.hpp`, moving them to source files alongside the definition of floating-point--only member functions for SFML vectors. This is done to improve compilation speed and avoid transitive includes.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI.


vittorioromeo:

include/SFML/System/Vector2.hpp
```diff
@@ -88,77 +87,77 @@ class Vector2
     /// * Vector2(r, phi) == Vector2(r, phi + n * 360_deg)
     ///
     ////////////////////////////////////////////////////////////
-    Vector2(T r, Angle phi);
-    
+    SFML_SYSTEM_API Vector2(T r, Angle phi);
```
kimci86, line 90: I now realize `<i><b>(floating-point)</b></i>` is missing from this method documentation.

kimci86 (commented):
Isn't it also required to declare explicit template instantiations in header files?

Something like this:
```c++
extern template class sf::Vector2<float>;
```

include/SFML/System/Vector2.hpp
vittorioromeo: > Isn't it also required to declare explicit template instantiations in header files?
> 
> Something like this:
> 
> ```c++
> extern template class sf::Vector2<float>;
> ```

I don't think it would really help in this case, because only a partial subset of the template member functions are explicitly instantiated. Where would you put the `extern template` declaration?

vittorioromeo: > I now realize `<i><b>(floating-point)</b></i>` is missing from this method documentation.

Will fix

ChrisThrasher (approved):

vittorioromeo:

eXpl0it3r (approved):

PR #2151, opened 2022-06-24
Improve test coverage of existing tests

ChrisThrasher:
## Description

I noticed that a few types had near-100% coverage so this PRs gets those types up to 100% coverage. In some cases it required modest changes to the underlying implementation. 

In the case of
```cpp
if (condition)
{
    return x;
} // Reported as untested code
else
{
    return y;
} 
```
the `else` is necessary and results in code coverage reporting an untested line so it's a win/win to transform it into
```cpp
if (condition)
{
    return x;
}

return y; 
```
which reports no untested code and is easier to read.
## Tasks

* [x] Tested on Linux
* [x]  Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

### DIFF
--- a/test/Graphics/Color.cpp
+++ b/test/Graphics/Color.cpp
@@ -9,95 +9,145 @@ TEST_CASE("sf::Color class - [graphics]")
     {
         SUBCASE("Default constructor")
         {
-            sf::Color c;
-            CHECK(c.r == 0);
-            CHECK(c.g == 0);
-            CHECK(c.b == 0);
-            CHECK(c.a == 255);
+            const sf::Color color;
+            CHECK(color.r == 0);
+            CHECK(color.g == 0);
+            CHECK(color.b == 0);
+            CHECK(color.a == 255);
         }
 
         SUBCASE("(r, g, b) constructor")
         {
-            sf::Color c(1, 2, 3);
-            CHECK(c.r == 1);
-            CHECK(c.g == 2);
-            CHECK(c.b == 3);
-            CHECK(c.a == 255);
+            const sf::Color color(1, 2, 3);
+            CHECK(color.r == 1);
+            CHECK(color.g == 2);
+            CHECK(color.b == 3);
+            CHECK(color.a == 255);
         }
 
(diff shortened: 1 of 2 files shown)


ChrisThrasher:
For whatever reason I can't get `sf::Rect` up to 100% coverage. I'm not sure what is missing.

Bromeon:
> In the case of
> 
> ```c++
> if (condition)
> {
>     return x;
> } // Reported as untested code
> else
> {
>     return y;
> } 
> ```
> 
> the `else` is necessary and results in code coverage reporting an untested line so it's a win/win to transform it into
> 
> ```c++
> if (condition)
> {
>     return x;
> }
> 
> return y; 
> ```
> 
> which reports no untested code and is easier to read.

This seems silly to me. Both codes are exactly equivalent, one having more coverage seems like a limitation of the tool.

I disagree that the 2nd snippet is more readable. `if/else` is more symmetrical and makes immediately clear that the two branches are mutually exclusive. It's also very common in expression-oriented languages to have:
```kotlin
return if (condition) {
    x
} else {
    y
}
```

I fear that we start changing perfectly valid code based on (false-positive) metrics of imperfect tools, and in the process make code harder to understand.

Bromeon:
Is there maybe an option to configure our Codecov integration in that regard?

ChrisThrasher:
> What bothers me is that we don't approach this change from a readability/code style perspective.

This is a readability improvement disguised as a code coverage improvement. I'd like to make such a chance even if the code coverage tool did not struggle with it. This change is largely motivated by my experience with clang-tidy which flags `else` after `return` since simpler (and IMO more readable) alternatives exist. 

https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/readability-else-after-return.html

It's not a big deal to remove this though. We'll still have like ~98% coverage without it.

ChrisThrasher:
I removed the changes I made to the implementation.

eXpl0it3r:

vittorioromeo (approved):

PR #2152, opened 2022-06-25
Add tests for `sf::MemoryInputStream`

Bambo-Borris:
## Description

`sf::FileInputStream` tests were already covered but noticed `sf::MemoryInputStream` tests were missing. Hoping these tests provide adequate coverage. 

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?
CI runs the tests!

### DIFF
--- a/test/System/MemoryInputStream.cpp
+++ b/test/System/MemoryInputStream.cpp
@@ -1,0 +1,33 @@
+#include <SFML/System/MemoryInputStream.hpp>
+#include <string_view>
+#include <ostream>
+
+#include <doctest.h>
+
+TEST_CASE("sf::MemoryInputStream class - [system]")
+{
+    SUBCASE("Empty stream")
+    {
+        sf::MemoryInputStream mis;
+
+        CHECK(mis.read(nullptr, 0) == -1);
+        CHECK(mis.seek(0) == -1);
+        CHECK(mis.tell() == -1);
+        CHECK(mis.getSize() == -1);
+    }
+
+    SUBCASE("Open memory stream")
+    {
+        using namespace std::literals::string_view_literals;
+        constexpr auto memoryContents = "hello world"sv;
+        sf::MemoryInputStream mis;
+        mis.open(memoryContents.data(), sizeof(char) * memoryContents.size());
+
+        char buffer[32];
+        CHECK(mis.read(buffer, 5) == 5);
+        CHECK(std::string_view(buffer, 5) == std::string_view(memoryContents.data(), 5));
+        CHECK(mis.seek(10) == 10);
+        CHECK(mis.tell() == 10);
+        CHECK(mis.getSize() == 11);
+    }
+}
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -17,6 +17,7 @@ SET(SYSTEM_SRC
     System/Clock.cpp
     System/Err.cpp
     System/FileInputStream.cpp
+    System/MemoryInputStream.cpp
     System/Time.cpp
     System/Vector2.cpp
     System/Vector3.cpp


test/System/MemoryInputStream.cpp
```diff
@@ -0,0 +1,30 @@
+#include <SFML/System/MemoryInputStream.hpp>
+#include "SystemUtil.hpp"
+
+#include <doctest.h>
+
+TEST_CASE("sf::MemoryInputStream class - [system]")
+{
+	SUBCASE("Empty stream")
```
kimci86, line 8: `getSize()` should also be tested in this subcase.
Also, please indent with spaces instead of tabs.

Bambo-Borris: Yeah didn't notice it was set to tabs, I'll add a `getSize()` check and format it :)

test/System/MemoryInputStream.cpp
```diff
@@ -0,0 +1,31 @@
+#include <SFML/System/MemoryInputStream.hpp>
+#include "SystemUtil.hpp"
+
```
kimci86, line 3: Maybe also include `<string_view>`

test/System/MemoryInputStream.cpp
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    SUBCASE("Temporary memory stream")
+    {
+        const std::string memoryContents = "hello world";
+        sf::MemoryInputStream mis;
+        mis.open(memoryContents.data(), sizeof(char) * memoryContents.size());
+
+        char buffer[32];
+        CHECK(mis.read(buffer, 5));
```
kimci86, line 25:
```suggestion
        CHECK(mis.read(buffer, 5) == 5);
```

ChrisThrasher (approved):
Test coverage goes brrrrrr

test/System/MemoryInputStream.cpp
```diff
@@ -0,0 +1,31 @@
...
+    {
+        sf::MemoryInputStream mis;
+
+        CHECK(mis.read(nullptr, 0) == -1);
+        CHECK(mis.seek(0) == -1);
+        CHECK(mis.tell() == -1);
+        CHECK(mis.getSize() == -1);
+    }
+
+    SUBCASE("Temporary memory stream")
```
kimci86, line 18: I am not sure "Temporary" is meaningful here.

Bambo-Borris: I took the case name from the `FileInputStream` one, could just change it to something like "Memory stream open" or something

test/System/MemoryInputStream.cpp
Bambo-Borris: Should we not verify that the `mis.read()` actually returned the expected number of bytes read in this case?

test/System/MemoryInputStream.cpp
kimci86: In the context of `FileInputStream`, a temporary file is meaningful, but a temporary memory doesn't make so much sense. I think "Open memory stream" would be more meaningful here.

test/System/MemoryInputStream.cpp
kimci86: Yes, this is exactly what I suggested if I understand correctly.

Bambo-Borris: Oh wait yeah I misread, seems I'm reading my diffs wrong today. Yeah I'm perfectly happy with changing this.

test/System/MemoryInputStream.cpp
```diff
@@ -0,0 +1,32 @@
+#include <SFML/System/MemoryInputStream.hpp>
+#include <string_view>
+#include "SystemUtil.hpp"
```
kimci86, line 3: One last thing, standard library headers come last by convention. See https://www.sfml-dev.org/style.php#includes
```suggestion
#include <SFML/System/MemoryInputStream.hpp>
#include "SystemUtil.hpp"
#include <string_view>
```

ChrisThrasher: I don't even think `SystemUtil.hpp` is getting used.

Bambo-Borris: > I don't even think `SystemUtil.hpp` is getting used.

No seems not, I think it's safe to remove it from the includes

kimci86: `<string>` is missing now :smile:
A raw array could also be used instead of the string.

Bambo-Borris: It appears it was how this file was getting `std::string` for the Linux GCC build so I'll also add the `<string>` include on

kimci86 (approved):

eXpl0it3r:

test/System/MemoryInputStream.cpp
```diff
@@ -0,0 +1,33 @@
+#include <SFML/System/MemoryInputStream.hpp>
+#include <string_view>
+#include <string>
+#include <sstream>
```
vittorioromeo, line 4: I think we can avoid these two.

test/System/MemoryInputStream.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+        CHECK(mis.read(nullptr, 0) == -1);
+        CHECK(mis.seek(0) == -1);
+        CHECK(mis.tell() == -1);
+        CHECK(mis.getSize() == -1);
+    }
+
+    SUBCASE("Open memory stream")
+    {
+        const std::string memoryContents = "hello world";
```
vittorioromeo, line 22:
```suggestion
        using std::literals::string_view_literals;
        constexpr auto memoryContents = "hello world"sv;
```

vittorioromeo (changes requested):
Looks good, but I think we can avoid depending on `string` and `sstream`.

test/System/MemoryInputStream.cpp
Bambo-Borris: It's possible to make this change but you still require `<sstream>` to build for me locally so we'd only be dropping 1 include `<string>`

vittorioromeo (approved):

vittorioromeo:

vittorioromeo:
Thanks!

PR #2153, opened 2022-06-25
Reduce the scope of helper class

ChrisThrasher:
## Description

`TemporaryFile` is only used in 1 .cpp file so it doesn't need to be part of SystemUtil. This also let me remove a few headers from SystemUtil.hpp which should result in a modest compile time improvement. 

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android


vittorioromeo (approved):
LGTM :)

vittorioromeo:

PR #2154, opened 2022-06-26
Allow implicit construction of `sf::VideoMode`

ChrisThrasher:
## Description

In #2055 I marked `sf::VideoMode::VideoMode` as explicit because a 2 parameter constructor became a 1 parameter constructor and I blindly followed the "mark converting constructors as `explicit`" advice. We've since received feedback about how annoying this is. In hindsight this change isn't helpful and simply requires users write excessively verbose code so this PR rolls that back.

This PR also fixes a number of incorrect docs which still reference the old `float, float` constructor.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -27,7 +27,7 @@ Tell us how to reproduce this issue. Please provide a [minimal, complete and ver
 
 int main()
 {
-    sf::RenderWindow window(sf::VideoMode(1280, 720), "Minimal, complete and verifiable example");
+    sf::RenderWindow window({{1280, 720}}, "Minimal, complete and verifiable example");
     window.setFramerateLimit(60);
 
     while (window.isOpen())
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -32,7 +32,7 @@ Describe how to best test these changes. Please provide a [minimal, complete and
 
 int main()
 {
-    sf::RenderWindow window(sf::VideoMode(1280, 720), "Minimal, complete and verifiable example");
+    sf::RenderWindow window({{1280, 720}}, "Minimal, complete and verifiable example");
     window.setFramerateLimit(60);
 
     while (window.isOpen())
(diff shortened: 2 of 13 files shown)


kimci86:
I have been thinking about how to require less braces too. It can be done by adding more constructors, but maybe a nice solution would be to make `VideoMode` and `Vector2` aggregate types, which allows brace elision.
See https://en.cppreference.com/w/cpp/language/aggregate_initialization
Proof of concept: https://godbolt.org/z/bbfG6v65a
This principle could also be applied to `Rect`, but would require quite a lot of rewriting.

eXpl0it3r:

vittorioromeo:
Not convinced about this PR. Yes, writing out `sf::VideoMode{w, h}` explicitly is more verbose, but it clearly tells readers what the type is. In contrast `{w, h}` could be anything, including a `sf::Vector2u`, `sf::Vector2f`, `std::initializer_list`, etc.

ChrisThrasher:
> Yes, writing out sf::VideoMode{w, h} explicitly is more verbose, but it clearly tells readers what the type is.

Users still have the choice to be explicit if they so choose.

>  In contrast {w, h} could be anything, including a sf::Vector2u, sf::Vector2f, std::initializer_list, etc.

Fortunately the type system handles this for us. If someone chooses the less explicit syntax, the type systems will continue to ensure they're writing code that makes some sense.

Bromeon:
> Users still have the choice to be explicit if they so choose.

I think that misses the point. A lot of time in programming is spent reading code, not writing it. Especially when it's code written by others or by oneself after a long time, slightly more expressive (even if verbose) code helps quickly understand what's happening.

Also, using the new convention in examples effectively makes it "canon", suggesting SFML users to adopt this for their own code.

JonnyPtn:
I'd reopen this - We've made the API less simple for no tangible gain.

Being explicit about the type has no benefit here - it's clear these parameters are the size of the window and whether that's through a videomode type or not is irrelevant

JonnyPtn:
We've had yet more people complain about this - any danger of reconsidering?

vittorioromeo:
> We've had yet more people complain about this - any danger of reconsidering?

What are the complaints?

JonnyPtn:
Pretty much what’s been discussed in this thread - the API is more difficult to use for no improvement


Also worth mentioning it currently requires 2 parentheses/braces not just one as you used in your example

PR #2155, opened 2022-06-26
Assorted unit testing improvements

ChrisThrasher:
## Description

This PR contains a few minor improvements to the tests that I had been wanting to make but didn't belong in any other PR so I combined them into one.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -9,6 +9,7 @@ add_library(sfml-test-main STATIC
 )
 target_include_directories(sfml-test-main SYSTEM PUBLIC "${PROJECT_SOURCE_DIR}/extlibs/headers")
 target_include_directories(sfml-test-main PUBLIC TestUtilities)
+target_compile_definitions(sfml-test-main PUBLIC DOCTEST_CONFIG_REQUIRE_STRINGIFICATION_FOR_ALL_USED_TYPES)
 target_link_libraries(sfml-test-main PUBLIC SFML::System)
 set_target_warnings(sfml-test-main)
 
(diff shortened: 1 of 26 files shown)


eXpl0it3r:

vittorioromeo:
What are the consequences for `DOCTEST_CONFIG_REQUIRE_STRINGIFICATION_FOR_ALL_USED_TYPES`? What's the benefit of enabling this?

ChrisThrasher:
> What are the consequences for DOCTEST_CONFIG_REQUIRE_STRINGIFICATION_FOR_ALL_USED_TYPES? What's the benefit of enabling this?

The extended commit message has more information.

> This ensures that if a printing function is not provided, compilation
fails. This prevents problems where a header is accidentally removed
that was previously providing an operator<< overload or prevents
new tests from being added without print support for all directly
tested types.

Bromeon (commented):
Enforcing concious `<<` overloads is a good idea!

The new #include style makes me wonder, would it make sense to add the directory above to the include path or add one level of nesting, in order to enable this?
```cpp
#include <Tests/GraphicUtils>
```
(This is very deep into cosmetics territory; I would also be OK with the current way 🙂)

ChrisThrasher:
> The new #include style makes me wonder, would it make sense to add the directory above to the include path or add one level of nesting, in order to enable this?

I don't suppose I feel strongly either way. I could be convinced one way or another.

vittorioromeo (approved):

Issue #2136, opened 2022-06-21
Add API to get monitor/screen information


vittorioromeo:
Inspired by https://github.com/eliasdaler/imgui-sfml/pull/213. It would be nice if SFML provided an API to get information about all the available monitors (e.g. x/y/width/height, refresh rate, rotation, etc.).

There was some work done [here](https://github.com/SFML/SFML/compare/master...Foaly:multi-monitor-windows), see [forum thread](https://en.sfml-dev.org/forums/index.php?topic=18950.0) that gives an idea on how to do this on Win32.

Is this something we'd like to target for 3.x? 

Other multi-monitor support ideas:
- API to switch from one monitor to the other when in fullscreen mode

eXpl0it3r:
I guess we should make sure not to mix multi-monitor support and retrieving of monitor information too much. There's certainly an overlay, but the former requires some more effort than the latter.

I've seen some requests regarding retrieving this information and also some for multi-monitor support (surprisingly few though, unless they've all found the past discussions).
So I think this is definitely something to tackle, but I'd suggest to target more something like SFML 3.1 and if necessary maybe prepare some windowing API changes for SFML 3.
Adding features in minor versions is always possible, thus I rather invest the resources/energy into the API changes for SFML 3.

danieljpetersen:
We could look to GLFW for inspiration. They have a strong cross platform monitor API (Linux, Windows, MacOS, though note that there's no mobile version of glfw).

https://www.glfw.org/docs/3.3/monitor_guide.html

https://www.glfw.org/docs/3.3/group__monitor.html

They define a monitor struct from which you can 
* Get a videomode
* Get the physical size of the monitor in millimetres 
* Get the content scale
* Get the virtual desktop position of the monitor in screen coordinates
* Get the work area of a monitor (the area not occupied by global task bars or menu bars)
* Get a human-readable name
* Get / set the gamma ramp

They provide a function to get an array of all connected monitors. The primary monitor is always the first object in this array, though you can also ask for the primary monitor explicitly.

SDL2 by contrast seems to have less support than GLFW. They do still provide a way to query for the number of monitors via

int numMonitors = SDL_GetNumVideoDisplays();

You can also obtain the resolutions of a given display via SDL_GetDisplayMode(...), where the first argument is an index into the relevant display. They also provide a way to obtain the desktop bounds of a window

https://wiki.libsdl.org/SDL2/SDL_GetDisplayMode

[...]

PR #2156, opened 2022-06-26
Use in-class member initializers

ChrisThrasher:
## Description

This PR uses in-class member initializers in the System module. The plan is that if this PR is accepted 1 or more subsequent PRs will employ this pattern in the other 5 modules as well as the tests and examples. 

See C++ Core Guidelines item C.45 for more information.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c45-dont-define-a-default-constructor-that-only-initializes-data-members-use-in-class-member-initializers-instead

_"Why use braced initializers?"_

There are some technical advantages to braced initializers including how lossy conversions are not possible. That advantage is nice but I also appreciate how initializing with `{}` makes it more obvious to the reader how they're not reading stack variables which by SFML convention are typically initialized with `()` or `=`. If a reader sees `{}` initializers it is more apparent that they're reading a class declaration and not procedural code.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android


vittorioromeo:
I'm generally OK with this but there are a few caveats:

- During development, making a change to a data member implies a change to a header, so more files might have to be re-compiled compared to making the change in a `.cpp` file. Usually not a big deal.

- For classes with multiple constructors, it can be annoying for readability to have to jump back and forth between the `.hpp` and `.cpp` to figure out what gets initialized by the constructor and what gets its value from the in-class member initializers.

- In some cases, it is possible to avoid a header dependency in the `.hpp` file by using it to initialize a data member in `.cpp` file. We definitely don't want to apply this change in those cases.

As far as I can see, none of these points apply to this PR. So, I'm OK with this, I think there are more worthwhile things to work on if you are willing to contribute to SFML in your spare time, but I am generally happy to merge this. 

@eXpl0it3r @Bromeon thoughts?

ChrisThrasher:
>  I think there are more worthwhile things to work on if you are willing to contribute to SFML in your spare time, but I am generally happy to merge this.

For the last 6 months I've had more time to contribute to SFML than the maintainers have had time to review my PRs so that's the last thing you need to worry about :) I'm routinely holding back PRs to avoid having 10 active PRs at a time.

vittorioromeo:
> > I think there are more worthwhile things to work on if you are willing to contribute to SFML in your spare time, but I am generally happy to merge this.
> 
> For the last 6 months I've had more time to contribute to SFML than the maintainers have had time to review my PRs so that's the last thing you need to worry about :) I'm routinely holding back PRs to avoid having 10 active PRs at a time.

And your work is very appreciated. However, while small refactoring are almost always a positive change, I would suggest trying to tackle more impactful and bigger issues. For example #2136 would a very cool contribution IMHO. Another thing I'd really like to see done is #1861, or some sort of work on batching. Maybe picking stuff out of the backlog here would also be a good idea (https://github.com/SFML/SFML/projects/7).

I will admit that I'm guilty of this myself, and maybe our time will be better spent having a meeting on Discord/Zoom and agreeing on what SFML 3.x should be, setting some concrete goals/deadlines, and working towards those. At the moment pretty much everyone (you included) is doing useful work, but I feel like there's a lack of direction.

ChrisThrasher:
> Another thing I'd really like to see done is https://github.com/SFML/SFML/pull/1861

I recently asked Metaquarx if I could take over #1861 to which they said I could. I made substantial progress on it yesterday so expect a PR for that soon. Check out my progress [here](https://github.com/ChrisThrasher/SFML/tree/clang-format). I just rebased it on master minutes ago.

I'm always working on bigger changes, it's just these little refactoring PRs only take minutes to implement so they're really easy to knock out while taking my time on the bigger stuff. It's hard to not do these when they're such easy, low hanging fruit to pick. 

> maybe our time will be better spent having a meeting on Discord/Zoom and agreeing on what SFML 3.x should be, setting some concrete goals/deadlines, and working towards those

I'm routinely asking in the Discord server what the SFML team thinks about various proposed changes and I seldom get feedback. I'd love to talk through these PRs before I write them but I'm mostly talking to myself in the #contributing channel so for lack of prompt feedback, I just submit the PRs and hope for the best.

I'd love do something like Discord voice calls every 2 weeks to talk about our direction and to talk through the big PRs which often require intense discussion before a decision can get made. Currently a few maintainers don't use Discord and I'm not sure how popular of a proposition that is to them but hopefully the maintainers can find some way to routinely convene.

ChrisThrasher:

eXpl0it3r:

Bromeon:
> * For classes with multiple constructors, it can be annoying for readability to have to jump back and forth between the `.hpp` and `.cpp` to figure out what gets initialized by the constructor and what gets its value from the in-class member initializers.

This is an issue indeed. By design, non-default constructors cannot use in-class member initializers exclusively, so they need additionally the constructor initializer list or body. For a reader, the previous "to see initial values of all members, check constructor" then becomes "read initial values from the class definition, unless you also see the fields in the constructor" 🤔 

Regarding conversions, does it make a difference? Then how is `T x{0}` in `Vector2<float>` working?

ChrisThrasher:
> This is an issue indeed. By design, non-default constructors cannot use in-class member initializers exclusively, so they need additionally the constructor initializer list or body. For a reader, the previous "to see initial values of all members, check constructor" then becomes "read initial values from the class definition, unless you also see the fields in the constructor" 🤔

Unfortunately, solving this problem requires repeating code in multiple places which I feel like is a worse solution still.

> Regarding conversions, does it make a difference? Then how is T x{0} in Vector2<float> working?

Probably because the compiler knows that the integer literal `0` can be losslessly converted to a `float`

ChrisThrasher:
https://github.com/ChrisThrasher/SFML/commit/0a012ca2fccf10d742fe992afa3b1535fd5e316d#diff-67e5ebd29f25fb6580b7215a91bcf1234c94d5fba8c80253148ef16a4144e795

If we continue with this pattern, here are some good examples of what we can expect. Just look at how much we can avoid repeating ourselves.

image

vittorioromeo:
> [ChrisThrasher@0a012ca#diff-67e5ebd29f25fb6580b7215a91bcf1234c94d5fba8c80253148ef16a4144e795](https://github.com/ChrisThrasher/SFML/commit/0a012ca2fccf10d742fe992afa3b1535fd5e316d#diff-67e5ebd29f25fb6580b7215a91bcf1234c94d5fba8c80253148ef16a4144e795)
> 
> If we continue with this pattern, here are some good examples of what we can expect. Just look at how much we can avoid repeating ourselves.
> 
> image

This is somewhat misleading, as delegating constructors could be used to both avoid the repetition and keep the initialization in the `.cpp` files.

ChrisThrasher:
> This is somewhat misleading, as delegating constructors could be used to both avoid the repetition and keep the initialization in the .cpp files.

Fair, but I stand by in-class initializers being the superior choice. For what it's worth I'd support a PR that made use of delegating constructors if anyone was willing to write that. That would also be an improvement over the status quo.

ChrisThrasher:
Rebased on master and made a minor simplification.

ChrisThrasher:
Rebased on master and fixed conflicts.

ChrisThrasher:
Rebased to fix modest merge conflicts with `master`.

ChrisThrasher:
Rebased to get new CI jobs.

eXpl0it3r:

include/SFML/System/String.hpp
```diff
@@ -56,7 +56,8 @@ class SFML_SYSTEM_API String
     ////////////////////////////////////////////////////////////
     // Static member data
     ////////////////////////////////////////////////////////////
-    static const std::size_t InvalidPos; //!< Represents an invalid position in the string
+    static inline const std::size_t InvalidPos{
+        std::basic_string<std::uint32_t>::npos}; //!< Represents an invalid position in the string
```
eXpl0it3r, line 60: Is this driven by clang formatting? Looks rather unreadable to me.

eXpl0it3r:
I have this readability issue currently anyways already in C++. It's very annoying if you just look at code on GitHub, want to know the type of a member variable and have to hunt down the header file. IDEs usually solve this by providing smarter navigations, especially for C++ there's often a fast switch to the header, so I'm not sure if that argument should have such a weight.

I've been using in-class member initialization for sometime, but I use C++ very seldomly, so my personally liking might not be a good indicator of anything.

eXpl0it3r (changes requested):
Double check if formatting can be improved

include/SFML/System/String.hpp
ChrisThrasher: Yeah that's clang-format failing to fit the expression and its trailing comment within our width limit. For the record, removing `std::` from `size_t` and `uint32_t` allow clang-format to fit this whole expression on one line. 

I moved that comment up one line and tweaked it such that it still produces the correct documentation. This is what the docs look like after this change.

Screen Shot 2022-09-29 at 2 59 44 PM

ChrisThrasher:

include/SFML/System/String.hpp
kimci86: You dropped one `std::` in the process.
You can have `static inline const std::size_t InvalidPos{std::basic_string<std::uint32_t>::npos};` on one line.

ChrisThrasher: Sorry! It's fixed now.

kimci86: Also, starting documentation comment with `///` would me more consistent with the rest of the code base.

ChrisThrasher:

include/SFML/System/String.hpp
ChrisThrasher: Are there any other member variables we document on the preceding line? I wasn't aware of any so I just went with what worked first.

kimci86: As far as I know, the convention is to use `///` except when putting documentation after member because some IDE would parse `///<` as XML, so we use `//!<` there.

ChrisThrasher: Alright I can do that. I tried it locally and Doxygen still produces the same output.

ChrisThrasher:

Bromeon:
I think the implementation looks good, just a general comment:

> _"Why use braced initializers?"_
> 
> There are some technical advantages to braced initializers including how lossy conversions are not possible. That advantage is nice but I also appreciate how initializing with `{}` makes it more obvious to the reader how they're not reading stack variables which by SFML convention are typically initialized with `()` or `=`. If a reader sees `{}` initializers it is more apparent that they're reading a class declaration and not procedural code.

I haven't seen that convention used elsewhere (and generally find the `{}` initialization rather unreadable, no other language does that); so in my opinion it could be `=` as well. But if everyone agrees that this is a convention worth using, I'm fine with it. Most important is that we have some consistency. 🙂

ChrisThrasher:
> (and generally find the `{}` initialization rather unreadable, no other language does that)

Agreed. I push back on using list initialization in most other contexts.

> I haven't seen that convention used elsewhere

I've observed no consensus on syntax for in-class initializers so I had to pick one.

I did think of another advantage to using list initialization here. For types like `sf::Vector2f` which require two arguments, you can easily initialize those like this:

```cpp
sf::Vector2f m_size{1080, 1920};
```

Doing the same with `=` simply requires more code

```cpp
sf::Vector2f m_size = {1080, 1920};
```
or
```cpp
sf::Vector2f m_size = sf::Vector2f(1080, 1920);
```

The problem ultimately comes down to the fact that you cannot use `auto` to declare class members which makes list initialization rather nice.

ChrisThrasher:

Bromeon (approved):

eXpl0it3r (approved):

PR #2159, opened 2022-06-27
Automate formatting with clang-format

ChrisThrasher:
## Description

Supersedes #1861 with the permission of its author @metaquarx. 

This PR is _big_ so I'll try to break down my rationale before anyone spends quite literally hours reviewing every line of code that got changed.

#1861 established that there is broad consensus behind adopting automated formatting and clang-format is the obvious tool to solve that. While #1861 didn't reach a consensus on how many PRs this might require, I advocate for 1 PR that both adds the .clang-format file and all the infrastructure to reformat the codebase _and_ performs said refactoring. By reformatting the code in this PR, it's easy for everyone to see the impact of the style rules chosen.

Because SFML does not already have perfectly consistent formatting, some amount of formatting changes are going to happen no matter how perfectly we configure the formatting rules. Fixing inconsistencies necessary generates a non-zero diff. Certainly though the goal is for this initial formatting pass to generate the most minimal diff possible. I tweaked as many settings as I could to ensure our .clang-format file best represents existing SFML style. 

#1861 left off with discussions about what minimum clang-format version to require. Because newer clang-format versions add new settings incompatible with older versions, we must specific a minimum version. At the time of authoring that PR, at least clang-format-12 was widely available. On the latest LTS Ubuntu and macOS, clang-format-14 is shipping. Unsure what the situation is on Windows. We were previously experiencing issues with clang-format-12 and clang-format-14 not producing identical results so were considering upgrading the minimum version to clang-format-14 but luckily only [minimal changes](https://github.com/ChrisThrasher/SFML/commit/0458cdf4629c7b308cdc1b8db6813812690a2f64) were required to get the two programs to produce identical results.

Now that I was able to get multiple versions of clang-format to produce identical results, I wanted CI to verify that. There was already a CI job for running clang-format on ubuntu-latest (Ubuntu 22.04 at the time of this PR) from #1861 so I added a second macOS formatting job which is using clang-format-14. These two formatting jobs make it even less likely that we end up in a situation where a PR submits code that provides different formatting across the two versions. I expect we may add more formatting jobs as more versions of clang-format get released and become usable in CI. Over time I also expect we periodically increase our clang-format version since newer version do provide features and bug fixes we may find useful.

So that leads us to the last commit of this PR which runs clang-format, touching 373 files with 8,214 additions and 7,743 deletions. It's a doozy to review this all but the good news is that we'll never have to review such a gigantic reformatting PR ever again!

There are certain aspects of SFML style which cannot be captured by clang-format including the empty newline after access specifiers and the extra space after `operator`. Otherwise please let me know if there are ways we can tweak the .clang-format to better capture SFML style. There are also certain code blocks where formatting was explicitly disabled via `// clang-format off` and `// clang-format on`. I suspect we may add or remove some of those depending on where we think manual formatting is worthwhile.

https://releases.llvm.org/12.0.0/tools/clang/docs/ClangFormatStyleOptions.html

Here are the docs for clang-format-12's formatting options.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android


eXpl0it3r:

ChrisThrasher:
I couldn't get certain constructor initializer lists to format as we wanted with clang-format-12. clang-format-14 has new options that do a better job formatting those initializers but we can't use those settings unless we decide v14 is the new requirement. See [PackConstructorInitializers](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) for more info.

cmake/Scripts/format.cmake
```diff
@@ -0,0 +1,29 @@
+# Check executable exists
```
eXpl0it3r, line 1: Is `Scripts` a common CMake term/folder? To me it seems a bit out of place, I'd probably just put it in the `cmake/` folder directly

examples/ftp/Ftp.cpp
```diff
@@ -190,12 +190,11 @@ int main()
                 break;
             }
         }
-
-    } while (choice != 0);
+    }
+    while (choice != 0);
```
eXpl0it3r, line 194: I guess there's no option for do-while loops? I can live with that change, since it's not used that much, but find the other formatting better.

cmake/Scripts/format.cmake
ChrisThrasher: Agreed. I hadn't thought twice about that placement. I'd rather move it to cmake/Format.cmake.

examples/ftp/Ftp.cpp
ChrisThrasher: That's totally possible. I thought we wanted the brace before the while but I'll remove that instead.

examples/X11/X11.cpp
```diff
@@ -169,31 +171,49 @@ int main()
     XSetWindowAttributes attributes;
     attributes.background_pixel = BlackPixel(display, screen);
     attributes.event_mask       = KeyPressMask;
-    Window window = XCreateWindow(display, RootWindow(display, screen),
-                                  0, 0, 650, 330, 0,
+    Window window               = XCreateWindow(display,
+                                  RootWindow(display, screen),
```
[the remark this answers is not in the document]
ChrisThrasher: It's aligning the assignments with the previous assignments. I agree it looks a bit odd. clang-format-s `align` settings can be very nice but I tend to avoid them because of situations like this where clang-format gets a bit too eager to vertically align everything for no meaningful gain.

ChrisThrasher:
I'm going to avoid any more rebasing of this PR for a while since I don't want to invalidate a review that someone has already sunk a meaningful amount of time into. I'll only push extra commits with the intent to fix up the commit history only if we decide this is something worth merging. So don't mind the potentially messy commit history. That's only temporary.

eXpl0it3r:

vittorioromeo:
Skimmed through the code. Most of it looks fine to me. My personal stance is that I'd take imperfect automated formatting over perfect manual formatting every day. For me, it's a "yes".

.github/workflows/ci.yml
```diff
@@ -94,3 +94,39 @@ jobs:
...
+      run: cmake -DCLANG_FORMAT_EXECUTABLE=clang-format-12 -P $GITHUB_WORKSPACE/cmake/Format.cmake
+
+    - name: Format Code on macOS
+      if: runner.os == 'macOS'
+      run: cmake -DCLANG_FORMAT_EXECUTABLE=/usr/local/Cellar/clang-format/14.0.5/bin/clang-format -P $GITHUB_WORKSPACE/cmake/Format.cmake
+
+    - name: Check Formatting
+      run: |
+        git diff
+        exit `git status --porcelain | wc -l`
```
eXpl0it3r, line 132: Does this lead to a failure or just output somewhere the line count?

.github/workflows/ci.yml
```diff
@@ -94,3 +94,39 @@ jobs:
...
+      if: runner.os == 'macOS'
+      run: brew install clang-format
+
+    - name: Format Code on Linux
+      if: runner.os == 'Linux'
+      run: cmake -DCLANG_FORMAT_EXECUTABLE=clang-format-12 -P $GITHUB_WORKSPACE/cmake/Format.cmake
+
+    - name: Format Code on macOS
+      if: runner.os == 'macOS'
+      run: cmake -DCLANG_FORMAT_EXECUTABLE=/usr/local/Cellar/clang-format/14.0.5/bin/clang-format -P $GITHUB_WORKSPACE/cmake/Format.cmake
```
eXpl0it3r, line 127: Should this be as hard coded? Wouldn't this break if a new patch or minor version is released?
Or asked differently, if we hard code it like that, shouldn't we also ensure that the specific version is installed via brew?

.github/workflows/ci.yml
ChrisThrasher: It's saying "if changes are observed to the repo, return a non-zero exit code". It's some shell trickery I came up with a year or so ago. We need some way of returning non-zero if we detect that running the auto formatter produces changes to the codebase.

examples/shader/Effect.hpp
```diff
@@ -71,16 +68,14 @@ class Effect : public sf::Drawable
     }
 
 private:
-
     // Virtual functions to be implemented in derived effects
-    virtual bool onLoad() = 0;
-    virtual void onUpdate(float time, float x, float y) = 0;
+    virtual bool onLoad()                                                               = 0;
+    virtual void onUpdate(float time, float x, float y)                                 = 0;
     virtual void onDraw(sf::RenderTarget& target, const sf::RenderStates& states) const = 0;
```
eXpl0it3r, line 74: This seems like a rather unnecessary and easily overlooked alignment, but I can also live with it

.github/workflows/ci.yml
ChrisThrasher: > Should this be as hard coded? 

Probably not.

> Wouldn't this break if a new patch or minor version is released?

Probably yes.

I never have to do this on my own macOS machine so surely there's a better way to get that clang-format binary into our `PATH`. Because Brew installs to different paths on x86 and ARM, I don't have a great way of debugging this but hopefully an x86 Mac user can help me out. I'd expect that binary to also exist in `/user/local/bin/clang-format` but I haven't tried that path yet.

examples/vulkan/Vulkan.cpp
```diff
@@ -253,6 +269,7 @@ class VulkanExample
         if (vulkanAvailable) setupDraw();
         if (vulkanAvailable) setupSemaphores();
         if (vulkanAvailable) setupFences();
+        // clang-format on
```
eXpl0it3r, line 272: Would be reformatted IMHO as well. Either that or then the block below shouldn't be reformatted either 😄

ChrisThrasher: Good point. I'll remove that `// clang-format off` block. I mostly want to reserve our use of `// clang-format off` for tabular data.

include/SFML/Config.hpp
```diff
@@ -123,7 +123,7 @@
 ////////////////////////////////////////////////////////////
 #if !defined(NDEBUG)
 
-    #define SFML_DEBUG
+#define SFML_DEBUG
 
 #endif
```
eXpl0it3r, line 128: Since this is just a list of marcos, it might make sense to preserve the formatting here, otherwise it becomes harder to read

include/SFML/Graphics/PrimitiveType.hpp
```diff
@@ -43,7 +43,7 @@ enum PrimitiveType
     LineStrip,     //!< List of connected lines, a point uses the previous point to form a line
     Triangles,     //!< List of individual triangles
     TriangleStrip, //!< List of connected triangles, a point uses the two previous points to form a triangle
-    TriangleFan    //!< List of connected triangles, a point uses the common center and the previous point to form a triangle
+    TriangleFan //!< List of connected triangles, a point uses the common center and the previous point to form a triangle
```
eXpl0it3r, line 46: That's kind of odd, but I can live without perfection 😄

ChrisThrasher: I looked into that. I believe the problem is that the comment on line 46 is so long that it exceeds our line limit so clang-format removes those alignment spaces to make sure the line is as short as possible. If we want better vertical alignment, we need shorter identifiers and/or shorter comments _or_ increase the maximum column width.

include/SFML/Config.hpp
Bromeon: Maybe finally time to move to `#pragma once` 😛

ChrisThrasher: I ran a quick test and can confirm that with `#pragma once` the preprocessor indentation behaves as expected and desired. With header guards, some files result in code like 
```cpp
#ifndef SFML_TIME_HPP
    #define SFML_TIME_HPP

    // Headers
    #include <SFML/System/Export.hpp>
```
but other files get their preprocessor directives indented as desired. So to get the preprocessor indentation without other undesirable side effects, we need to use `#pragma once`. Personally I don't mind left-indenting all preprocessor directives but I don't feel strongly either way.

eXpl0it3r: I think it's generally okay to not indent preprocessor directives, just thought maybe here disabling formatting might be useful. Maybe also in the Export.hpp

test/Graphics/Transform.cpp
```diff
@@ -28,31 +29,27 @@ TEST_CASE("sf::Transform class - [graphics]")
...
-            const sf::Transform transform(10.0f, 11.0f, 12.0f,
-                                          13.0f, 14.0f, 15.0f,
-                                          16.0f, 17.0f, 18.0f);
+            const sf::Transform      transform(10.0f, 11.0f, 12.0f, 13.0f, 14.0f, 15.0f, 16.0f, 17.0f, 18.0f);
             const std::vector<float> matrix(transform.getMatrix(), transform.getMatrix() + 16);
-            CHECK(matrix == std::vector<float>{10.0f, 13.0f, 0.0f, 16.0f,
-                                               11.0f, 14.0f, 0.0f, 17.0f,
-                                                0.0f,  0.0f, 1.0f,  0.0f,
-                                               12.0f, 15.0f, 0.0f, 18.0f});
+            CHECK(
```
Bambo-Borris, line 34: In situations like this I think the previous formatting was helpful because it allows you to see the data in its matrix layout, but after formatting it's harder to notice things like this

src/SFML/Window/Win32/WglContext.cpp
```diff
@@ -88,7 +89,14 @@ namespace priv
 String getErrorString(DWORD errorCode)
 {
     PTCHAR buffer;
-    if (FormatMessage(FORMAT_MESSAGE_MAX_WIDTH_MASK | FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS, nullptr, errorCode, 0, reinterpret_cast<LPTSTR>(&buffer), 256, nullptr) != 0)
+    if (FormatMessage(FORMAT_MESSAGE_MAX_WIDTH_MASK | FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM |
+                          FORMAT_MESSAGE_IGNORE_INSERTS,
```
Bambo-Borris, line 93: _Might just be me being pedantic_ The formatting here looks out of place with the section below it, I can understand why but it did catch my eye

test/Graphics/Transform.cpp
ChrisThrasher: Formatting matrices is maybe the canonical use case for `// clang-format off` but we'd basically have to wrap this entire file in a `// clang-format off` block if we wanted to format all matrices as blocks.

.github/workflows/ci.yml
Bromeon: There's also the [`--exit-code` argument](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---exit-code), which returns 0 if no differences, and 1 if differences are found.
You could use it as follows:
```sh
git diff --exit-code || exit 1
```
(or possibly add some error output). The problem with ```exit `cmd` ```  is that it exits unconditionally, so when people add another step in the future, they might be confused as to why it's not executed. `cmd || exit 1` avoids this.

**Note:** when using those commands in the past, I encountered quite a few subtleties regarding git-ignored and indexed/staged files; e.g. there's also the `--cached` argument.

Your approach with using the line count as exit code is pretty cool, I have to say 🙂

ChrisThrasher: > There's also the [--exit-code argument](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---exit-code), which returns 0 if no differences, and 1 if differences are found.

That's perfect! I didn't know that existed. I'll have to try that out later. Probably a good fit for this PR.

src/SFML/Window/Win32/WglContext.cpp
ChrisThrasher: > There's also the [--exit-code argument](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---exit-code), which returns 0 if no differences, and 1 if differences are found.

That is correct. See the [BinPackParameters](https://releases.llvm.org/12.0.0/tools/clang/docs/ClangFormatStyleOptions.html) option. I prefer we not use this, but it does exist.

.github/workflows/ci.yml
ChrisThrasher: Wow this is incredible. 
```
git diff
exit `git status --porcelain | wc -l`
```
becomes simply
```
git diff --exit-code
```
I love it! I used a separate branch to check that this fails when expected then pushed that change up to this branch.

.github/workflows/ci.yml
ChrisThrasher: I figured out what I was doing wrong. I changed `-DCLANG_FORMAT_EXECUTABLE=/usr/local/Cellar/clang-format/14.0.5/bin/clang-format` to `-DCLANG_FORMAT_EXECUTABLE=clang-format` which is possible because that executable is symlinked into /usr/loca/bin which is already part of the `PATH`.

ChrisThrasher:
Good news! We have a Windows formatting job! Ubuntu 20 uses clang-format-12, Windows uses clang-format-13, and macOS uses clang-format-14. We're really covering all our bases with this setup so it's hard for us to get surprised by different machines producing different formatting. If 1 formatting job fails, all of them should fail.

src/SFML/Graphics/Texture.cpp
```diff
@@ -364,7 +381,8 @@ Image Texture::copyToImage() const
         // Texture is either padded or flipped, we have to use a slower algorithm
 
         // All the pixels will first be copied to a temporary array
-        std::vector<Uint8> allPixels(static_cast<std::size_t>(m_actualSize.x) * static_cast<std::size_t>(m_actualSize.y) * 4);
+        std::vector<Uint8> allPixels(
+            static_cast<std::size_t>(m_actualSize.x) * static_cast<std::size_t>(m_actualSize.y) * 4);
```
eXpl0it3r, line 385: Personally, I could live with longer max line sizes. Quite a few of these changes make it IMHO less readable

eXpl0it3r:
Looks like the Objective-C *.mm files didn't get formatted. Does clang-format support them?

src/SFML/Graphics/Texture.cpp
ChrisThrasher: I could live with that too. 
```diff
-    HWND     m_handle;        //!< Win32 handle of the window
-    LONG_PTR m_callback;      //!< Stores the original event callback function of the control
-    bool     m_cursorVisible; //!< Is the cursor visible or hidden?
-    HCURSOR m_lastCursor; //!< Last cursor used -- this data is not owned by the window and is required to be always valid
-    HICON   m_icon;       //!< Custom icon assigned to the window
-    bool    m_keyRepeatEnabled; //!< Automatic key-repeat state for keydown events
-    Vector2u m_lastSize;        //!< The last handled size of the window
-    bool     m_resizing;        //!< Is the window being resized?
-    Uint16 m_surrogate; //!< First half of the surrogate pair, in case we're receiving a Unicode character in two events
-    bool   m_mouseInside;   //!< Mouse is inside the window?
-    bool   m_fullscreen;    //!< Is the window fullscreen?
-    bool   m_cursorGrabbed; //!< Is the mouse cursor trapped?
+    HWND     m_handle;           //!< Win32 handle of the window
+    LONG_PTR m_callback;         //!< Stores the original event callback function of the control
+    bool     m_cursorVisible;    //!< Is the cursor visible or hidden?
+    HCURSOR  m_lastCursor;       //!< Last cursor used -- this data is not owned by the window and is required to be always valid
+    HICON    m_icon;             //!< Custom icon assigned to the window
+    bool     m_keyRepeatEnabled; //!< Automatic key-repeat state for keydown events
+    Vector2u m_lastSize;         //!< The last handled size of the window
+    bool     m_resizing;         //!< Is the window being resized?
+    Uint16   m_surrogate;        //!< First half of the surrogate pair, in case we're receiving a Unicode character in two events
+    bool     m_mouseInside;      //!< Mouse is inside the window?
+    bool     m_fullscreen;       //!< Is the window fullscreen?
+    bool     m_cursorGrabbed;    //!< Is the mouse cursor trapped?
```
These are the kinds of changes we get when increasing the column width.

ChrisThrasher:
> Looks like the Objective-C *.mm files didn't get formatted. Does clang-format support them?

Good point. That was an oversight. I'll fix that real quick.

examples/cocoa/CocoaAppDelegate.mm
```diff
@@ -27,218 +28,202 @@
...
-
-    sf::RenderWindow    renderWindow;
-    sf::Font            font;
-    sf::Text            text;
-    sf::Texture         logo;
-    sf::Sprite          sprite;
-    sf::Color           background;
+struct SFMLmainWindow {
+  SFMLmainWindow(sf::WindowHandle win)
+      : renderWindow(win), background(sf::Color::Blue) {
```
eXpl0it3r, line 46: This seems to be wrongly formatted. Do Obj-C files take different configs than C++ files?

ChrisThrasher: They use the same .clang-format file but I'm guessing our ObjC settings need improvement. Adding `IndentWidth: 4` seems to fix it.

eXpl0it3r: Braces placement is still wrong though

ChrisThrasher: Right, seems like many options from the C++ section need to get copied into the ObjC section.

ChrisThrasher: I think I figured out how to make sure all C++ settings apply to ObjC so that we don't have to repeat ourselves between the two languages.

ChrisThrasher:
It was recently pointed out that we're not formatting .mm ObjC++ files which made me realized we're not formatting .h files which exist in both src/ and in examples/. However. when adding .h files to the list of files to format, the runtime for clang-format on my machine increases from <1s to 50s. This is needless to say a shocking difference and makes formatting so slow so I question whether auto formatting those .h files is worth it. I think the biggest culprits are the 3 gl.h headers in examples. I'm not sure what we should do about that.

src/SFML/Graphics/Texture.cpp
Bromeon: Absolutely agree. We live in a time where screens tend to be super wide 🙂

eXpl0it3r:
The `gl.h` are generated anyways and shouldn't be reformatted, if those are the culprit, then we can easily just exclude them. Not sure what other *.h file exist and whether they're really "part" of the SFML code base

ChrisThrasher:
> Not sure what other *.h file exist and whether they're really "part" of the SFML code base

@eXpl0it3r 

See the src/SFML/Window/OSX directory. Looks like 1st party code. It has the SFML copyright header.

src/SFML/Graphics/Texture.cpp
ChrisThrasher: Just pick whatever bigger width you want and I'll make the change.

ChrisThrasher:
> if those are the culprit, then we can easily just exclude them.

I'm not actually sure how to pull this off in Format.cmake. Why can't all those examples just reference a single gl.h file? That would help mitigate this problem.

eXpl0it3r:
`list(FILTER ${folder_files} EXCLUDE REGEX "gl.h")` or similar might do it

> Why can't all those examples just reference a single gl.h file?

The examples wouldn't be self-contained anymore.

Bromeon:
Regarding function calls that are broken into multiple lines, have you considered using
```cpp
Function(
   firstArg,
   secondArg,
   etc
);
```
instead of 
```cpp
Function(firstArg,
         secondArg,
         etc);
```

It looks like this can be configured with `AlignAfterOpenBracket` [here](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) -- I'll copy-paste since they don't have section links:

grafik

And setting these options to `false` would put each argument/parameter on their own line:

grafik


See also: https://stackoverflow.com/q/21966530
The config could then be:
```yaml
AlignAfterOpenBracket: BlockIndent
BinPackArguments: false
BinPackParameters: false
```
I'm also OK with `AlwaysBreak` instead of `BlockIndent`, if you prefer -- I've just seen the former more often:
```cpp
Function(
   firstArg,
   secondArg,
   etc);
```

eXpl0it3r:
> Regarding function calls that are broken into multiple lines, have you considered using
> 
> ```c++
> Function(
>    firstArg,
>    secondArg,
>    etc
> );
> ```
> 
> instead of
> 
> ```c++
> Function(firstArg,
>          secondArg,
>          etc);
> ```

I prefer the second one, so haven't really considered the first one 😄

Bromeon:
Apart from style, I see two disadvantages with the "align on opening `(`" that you don't have with the "fixed indent":
* If the function call starts late in the line (e.g. part of another expression or long initialization), some arguments may not fit the remaining space and need to be _additionally_ broken.
* If a function is renamed, this re-aligns arguments on call sites and parameters on declaration sites, creating larger diffs.

examples/X11/X11.cpp
```diff
@@ -169,31 +171,49 @@ int main()
...
+                                  0,
+                                  650,
+                                  330,
+                                  0,
                                   DefaultDepth(display, screen),
                                   InputOutput,
                                   DefaultVisual(display, screen),
-                                  CWBackPixel | CWEventMask, &attributes);
+                                  CWBackPixel | CWEventMask,
+                                  &attributes);
```
Bromeon, line 185: What's happening here -- why are the arguments aligned with the function name and not the first argument?

ChrisThrasher: ¯\_(ツ)_/¯

examples/island/stb_perlin.h
```diff
@@ -64,12 +64,41 @@
 
 
 #ifdef __cplusplus
-extern "C" {
+extern "C"
+{
```
eXpl0it3r, line 68: File should be excluded as well

examples/vulkan/vulkan.h
```diff
@@ -42,80 +42,81 @@
 #define GLAD_OPTION_VULKAN_HEADER_ONLY
 
 #ifdef __cplusplus
-extern "C" {
+extern "C"
+{
```
eXpl0it3r, line 46: File should be excluded as well

src/SFML/Window/OSX/HIDInputManager.mm
```diff
@@ -397,477 +382,805 @@
...
-        case kHIDUsage_KeyboardCancel:              return 0xff;
-        case kHIDUsage_KeyboardClear:               return 0xff;
-        case kHIDUsage_KeyboardPrior:               return 0xff;
-        case kHIDUsage_KeyboardReturn:              return 0xff;
-        case kHIDUsage_KeyboardSeparator:           return 0xff;
-        case kHIDUsage_KeyboardOut:                 return 0xff;
-        case kHIDUsage_KeyboardOper:                return 0xff;
-        case kHIDUsage_KeyboardClearOrAgain:        return 0xff;
-        case kHIDUsage_KeyboardCrSelOrProps:        return 0xff;
-        case kHIDUsage_KeyboardExSel:               return 0xff;
```
eXpl0it3r, line 587: Might want to keep, but either way works for me

src/SFML/Window/OSX/SFOpenGLView+mouse_priv.h
```diff
@@ -27,6 +27,7 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Window/Mouse.hpp>
+#import <SFML/Window/OSX/SFOpenGLView.h>
```
eXpl0it3r, line 30: Isn't this doing more than just formatting?

src/SFML/Window/OSX/SFOpenGLView+keyboard_priv.h
```diff
@@ -27,6 +27,7 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Window/Mouse.hpp>
+#import <SFML/Window/OSX/SFOpenGLView.h>
```
eXpl0it3r, line 30: Isn't this doing more than just formatting?

src/SFML/Window/OSX/SFOpenGLView.mm
```diff
@@ -26,21 +27,18 @@
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
-#include <SFML/Window/OSX/WindowImplCocoa.hpp>
 #include <SFML/System/Err.hpp>
-
-#import <SFML/Window/OSX/SFOpenGLView.h>
```
eXpl0it3r, line 32: I guess it's move out from here

src/SFML/Window/OSX/SFOpenGLView+mouse_priv.h
ChrisThrasher: Fixing compiler errors caused by clang-format reordering headers.

src/SFML/Window/OSX/HIDInputManager.mm
ChrisThrasher: To what extent do we care about that? There are seemingly many instances of big switch cases with custom vertical alignment. There is a clang-format option that allows for short case statements on a single line which we might consider using.

examples/vulkan/vulkan.h
ChrisThrasher: Good catch. I found that stb_perlin.h was getting formatted as well so I excluded that file and reverted the formatting.

examples/island/stb_perlin.h
ChrisThrasher: Thanks! I fixed that.

ChrisThrasher:
There were some nontrivial conflicts with `master` and I needed to squash a lot of commits so I did both. While I was at it, I realized that I previously pushed erroneous formatting changes to the Objective-C files so I reverted all that. The problem was, the .clang-format file originally applied different rules to the non-C++ code so those files got formatted with lots of default settings that resulted in unwanted changes.

src/SFML/Window/NetBSD/JoystickImpl.cpp
```diff
@@ -44,117 +45,124 @@
...
-    {
-        /*
+void updatePluggedList()
+{
+    /*
          * Devices /dev/uhid<x> are shared between joystick and any other
          * human interface device. We need to iterate over all found devices
          * and check if they are joysticks. The index of JoystickImpl::open
          * does not match the /dev/uhid<index> device!
          */
```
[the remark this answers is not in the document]
ChrisThrasher: Weird anomaly but if I manually fix that, clang-format doesn't undo those changes so this is easily fixable.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -25,367 +25,359 @@
...
+        if (name == windowManagerName)
+            return true;
+    }
+
+    return false;
+}
+
+sf::Keyboard::Key keysymToSF(KeySym symbol)
+{
+    // clang-format off
```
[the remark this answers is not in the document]
ChrisThrasher: Good catch. I'll fix that.

src/SFML/Window/Unix/GlxContext.cpp
```diff
@@ -25,13 +25,18 @@
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
+
+// clang-format off
```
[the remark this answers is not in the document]
ChrisThrasher: I'm testing this in CI.

https://github.com/ChrisThrasher/SFML/actions/runs/2618634518

Seems to be okay?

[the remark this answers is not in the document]
kimci86: Yes, it confirms what I expected. :+1:

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -25,367 +25,359 @@
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
+
+// clang-format off
```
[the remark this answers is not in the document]
ChrisThrasher: Fixed

src/SFML/Window/Win32/WglContext.cpp
```diff
@@ -25,9 +25,10 @@
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
-#include <SFML/Window/WindowImpl.hpp> // included first to avoid a warning about macro redefinition
-#include <SFML/Window/Win32/WglContext.hpp>
 #include <SFML/System/Err.hpp>
+#include <SFML/Window/Win32/WglContext.hpp>
+#include <SFML/Window/WindowImpl.hpp> // included first to avoid a warning about macro redefinition
```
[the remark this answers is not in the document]
ChrisThrasher: Fixed

src/SFML/Window/Unix/GlxContext.cpp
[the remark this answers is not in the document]
ChrisThrasher: Fixed

include/SFML/Graphics/Glsl.hpp
```diff
@@ -58,136 +59,136 @@ namespace priv
...
+/// 4D float vectors can be implicitly converted from sf::Color
+/// instances. Each color channel is normalized from integers
+/// in [0, 255] to floating point values in [0, 1].
+/// \code
+/// sf::Glsl::Vec4 zeroVector;
+/// sf::Glsl::Vec4 vector(1.f, 2.f, 3.f, 4.f);
+/// sf::Glsl::Vec4 color = sf::Color::Cyan;
+/// \endcode
+////////////////////////////////////////////////////////////
+using Vec4 = implementation - defined;
```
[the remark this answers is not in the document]
ChrisThrasher: Agreed. It used to be `implementation-defined` until clang-format treated that as a subtraction operation.

src/SFML/Graphics/Shape.cpp
```diff
@@ -25,24 +25,25 @@
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
-#include <SFML/Graphics/Shape.hpp>
 #include <SFML/Graphics/RenderTarget.hpp>
+#include <SFML/Graphics/Shape.hpp>
```
ChrisThrasher, line 29: clang-format by default will try to include the main header file first but that seemingly only works for headers within the same directory that are included via quotes without any directory prefixes. I've tried messing around with the header sorting options to fix this but cannot figure it out.

ChrisThrasher:
This PR accumulated another 15 or so commit so I did another round of rebasing to clean up the commit history. I'm partly through reviewing every file and found a number of more fixes to make which I've since pushed and squashed. It's tedious work but I think we're converging on something good enough to merge very soon. The things I'm fixing are getting smaller and smaller.

ChrisThrasher:
I just finished another complete review of all 420 files. I did one quick little rebase to fix up the last commit I added. I'm happy with the state of the code and the state of the commit history so as far as I'm aware, this is good to go.

We might consider increasing the column width from 120 to something like 132 but other than that, I don't see any significant changes left to be made.

src/SFML/Graphics/Shape.cpp
ChrisThrasher: We can fix this by used quoted include paths but at the moment we don't want to do that for these reasons:

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#sf12-prefer-the-quoted-form-of-include-for-files-relative-to-the-including-file-and-the-angle-bracket-form-everywhere-else

Maybe we reevaluate this later.

src/SFML/Graphics/View.cpp
```diff
@@ -46,11 +47,11 @@ m_invTransformUpdated(false)
...
-m_size               (),
-m_rotation           (),
-m_viewport           ({0, 0}, {1, 1}),
-m_transformUpdated   (false),
+m_center(),
+m_size(),
+m_rotation(),
+m_viewport({0, 0}, {1, 1}),
+m_transformUpdated(false),
 m_invTransformUpdated(false)
```
Bromeon, line 55: Unfortunately it doesn't look like there's an option to align constructor initializations, at least I haven't found one. But I think we can live with that 🙂

ChrisThrasher: Sadly no, as far as I'm aware. Luckily tons of these initializers can be removed (anything default initialized) or moved to the class declaration like I've started doing in #2156.

Bromeon:
> I just finished another complete review of all 420 files.

Wow 🤯 thanks a lot for the huge continuous effort and all the repeated improvement iterations!

I looked at some samples, and most looked good. What was not really answered (except by eXpl0it3r) is [my comment regarding function style](https://github.com/SFML/SFML/pull/2159#issuecomment-1169616429). But if the current style pleases more people, that's fine for me too, I don't want to bikeshed any longer 🙂

> I'm sure we could spend a few more weeks really refining the formatting options but such a big PR can't get postponed forever without causing significant rework and wasted effort.

Absolutely, after this round of comments we should definitely merge it. As mentioned [further up](https://github.com/SFML/SFML/pull/2159#issuecomment-1168670377), we can always improve things in the future.

ChrisThrasher:
> Wow 🤯 thanks a lot for the huge continuous effort and all the repeated improvement iterations!

Thanks! :)

> What was not really answered (except by eXpl0it3r) is https://github.com/SFML/SFML/pull/2159#issuecomment-1169616429.

I'm open to the alternate style you proposed but would prefer to stick with the current style for now. It does occasionally produce odd results but generally I think it's the better option.

src/SFML/System/Clock.cpp
```diff
@@ -32,8 +32,7 @@
 namespace sf
 {
 ////////////////////////////////////////////////////////////
-Clock::Clock() :
-m_startTime(ClockImpl::now())
+Clock::Clock() : m_startTime(ClockImpl::now())
```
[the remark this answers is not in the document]
Bromeon: Ah, I see. Should we add a comment in the config file "for future clang-format" so that we won't forget to upgrade?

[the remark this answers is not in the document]
ChrisThrasher: I'm fine either way but would rather not. I suspect we'll drop Ubuntu 20 somewhere in the SFML 3 release cycle. I'm not going anywhere so I think I'll remember to make that change as soon as possible. Or maybe this comment will look silly in a few years 😅

vittorioromeo (approved):
Skimmed through the changes again. Still looks good to me, haven't found any change that I think should be manually overridden. I would suggest to merge and check for regressions or incrementally improve the situation where required.

@ChrisThrasher, if you haven't already, could you run all the example programs? Just to make sure the PR doesn't break anything.

vittorioromeo:

ChrisThrasher:
> @ChrisThrasher, if you haven't already, could you run all the example programs? Just to make sure the PR doesn't break anything.

Sure thing. When I get back to my computer tomorrow I'll run all the examples that compile on macOS and/or Ubuntu.

ChrisThrasher:
> @ChrisThrasher, if you haven't already, could you run all the example programs? Just to make sure the PR doesn't break anything.

@vittorioromeo Looks like all the examples that I can run are working just fine. I wasn't able to run the Android, Cocoa, Vulkan, or Win32 examples, but I have no reason to believe there's any issues with them.

eXpl0it3r (approved):

src/SFML/Graphics/Texture.cpp
eXpl0it3r: Let's just go with what we have right now, maybe sometime later, someone is bothered enough to define a wider width

Bambo-Borris (approved):

ChrisThrasher:
https://github.com/SFML/SFML/pull/2159#discussion_r917086694

Now that we're using CMake 3.22, we're not longer providing seamless support to Ubuntu 20 users. I think it's also reasonable now to drop the clang-format-12 requirement and move up to requiring clang-format-14 which ships with Ubuntu 22 and gives us a few more useful options for enforcing more facets of style.

PR #2157, opened 2022-06-26
Dynamically load EGL with glad on Linux (DRM)

kimci86:
## Description

When running a program using SFML built in debug mode on Linux with DRM (e.g. on Raspberry Pi), the program crashes.
That is because the function `sf::priv::eglCheckError` which checks for EGL errors in debug mode expects the function `eglGetError` to be dynamically loaded by glad, but the `DRMContext` class did not trigger this loading (as opposed to `EglContext`).
So I made `DRMContext` load EGL with glad, inspired by how `EglContext` does it.

Then, `DRMContext` and `EglContext` both contain gald/egl implementation with the same symbols, so they cannot be linked in a library together. I edited the CMake scripts so at most one of those files is used at once in `sfml-window`.

Because the `glad/egl.h` header was including X11 headers, the None macro was in conflict with `sf::Style::None`.
Updating `glad/egl.h` with a freshly generated header solved it.
I edited the generated file adding sf_ and SF_ prefix and making loading functions static like it was done on the previous version.

This PR is a continuation of the PR #1603

I am discovering the topic while I work on this PR, so I might be doing stupid things. Feedback is welcome. @binary1248
Thank you @oomek for helping me on this.

## How to test this PR?

On Raspberry PI, with no desktop environment running.

1. Build SFML with DRM in Debug mode.
```
cmake -B build -DCMAKE_BUILD_TYPE=Debug -DSFML_USE_DRM=True
cmake --build build
```

2. Build and run the following example program.

```cpp
#include <SFML/Graphics.hpp>

int main(int argc, char *argv[])
{
    sf::RenderWindow window;

    window.create(sf::VideoMode::getDesktopMode(), "SFML Test", sf::Style::Fullscreen);
    window.setVerticalSyncEnabled(true);

    int w = window.getSize().x;
    int h = window.getSize().y;

    sf::RectangleShape rect(sf::Vector2f(w / 64.0f, h));

    sf::VertexArray quad(sf::TriangleStrip, 4);
    quad[0].position = sf::Vector2f(0, 0);
    quad[1].position = sf::Vector2f(w, 0);
    quad[2].position = sf::Vector2f(0, h);
    quad[3].position = sf::Vector2f(w, h);

    quad[0].color = sf::Color::Red;
    quad[1].color = sf::Color::Yellow;
    quad[2].color = sf::Color::Blue;
    quad[3].color = sf::Color::Cyan;

    while (window.isOpen())
    {
        sf::Event event;
        while (window.pollEvent(event))
        {
[...]

PR #2160, opened 2022-06-27
Fix compilation on Linux with SFML_OPENGL_ES option

kimci86:
## Description

Last night in #2157, I introduced without realizing some nonsense in the Window/CMakeLists.txt file.
I hoped CI would catch this kind of mistakes but there is no check with SFML_OPENGL_ES option enabled.

This PR makes it possible to build SFML on Linux with any combination of SFML_USE_DRM and SFML_OPENGL_ES as it was before #2157.

## How to test this PR?

```
cmake -B build -DCMAKE_BUILD_TYPE=Debug -DWARNINGS_AS_ERRORS=False
cmake -B build -DSFML_USE_DRM=False -DSFML_OPENGL_ES=False && cmake --build build
cmake -B build -DSFML_USE_DRM=True  -DSFML_OPENGL_ES=False && cmake --build build
cmake -B build -DSFML_USE_DRM=False -DSFML_OPENGL_ES=True  && cmake --build build
cmake -B build -DSFML_USE_DRM=True  -DSFML_OPENGL_ES=True  && cmake --build build
```

I had to disable WARNINGS_AS_ERRORS because some OpenGL ES specific parts trigger warnings, but that is a different topic.

Tested on Ubuntu and Raspbian.

### DIFF
--- a/src/SFML/Window/CMakeLists.txt
+++ b/src/SFML/Window/CMakeLists.txt
@@ -91,18 +91,9 @@ if(SFML_OS_WINDOWS)
     # make sure that we use the Unicode version of the Win API functions
     add_definitions(-DUNICODE -D_UNICODE)
 elseif(SFML_OS_LINUX OR SFML_OS_FREEBSD OR SFML_OS_OPENBSD OR SFML_OS_NETBSD)
-    if(SFML_OPENGL_ES)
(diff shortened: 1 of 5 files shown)


eXpl0it3r:
Thanks for the fix! 🙂

PR #2161, opened 2022-07-12
Add tests for SFML/Config.hpp

ChrisThrasher:
## Description

I just realized that there are some things in Config.hpp that can easily be tested. These are rather trivial tests but the more the merrier, right? It's easy to say "why bother testing this?" but it's hard to draw fine line between "too trivial to justify testing" and "complex enough to justify testing" so I'd rather we err on the side of testing all code which can be tested this easily. 

I tried to test more code in this header but most of it lives behind platform-specific preprocessor switches and I'm not sure how we'd test that without basically recreating the entire contents of the header in these tests.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

### DIFF
--- a/test/System/Config.cpp
+++ b/test/System/Config.cpp
@@ -1,0 +1,29 @@
+#include <SFML/Config.hpp>
+
+#include <doctest/doctest.h>
+
+TEST_CASE("SFML/Config.hpp")
+{
+    SUBCASE("Version macros")
+    {
+        CHECK(SFML_VERSION_MAJOR == EXPECTED_SFML_VERSION_MAJOR);
+        CHECK(SFML_VERSION_MINOR == EXPECTED_SFML_VERSION_MINOR);
+        CHECK(SFML_VERSION_PATCH == EXPECTED_SFML_VERSION_PATCH);
+        CHECK(SFML_VERSION_IS_RELEASE == EXPECTED_SFML_VERSION_IS_RELEASE);
+    }
+
+    SUBCASE("Fixed width types")
+    {
+        CHECK(sizeof(sf::Int8) == 1);
+        CHECK(sizeof(sf::Uint8) == 1);
+
+        CHECK(sizeof(sf::Int16) == 2);
+        CHECK(sizeof(sf::Uint16) == 2);
+
+        CHECK(sizeof(sf::Int32) == 4);
+        CHECK(sizeof(sf::Uint32) == 4);
+
+        CHECK(sizeof(sf::Int64) == 8);
+        CHECK(sizeof(sf::Uint64) == 8);
+    }
+}
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -24,6 +24,7 @@ set_target_warnings(sfml-test-main)
 SET(SYSTEM_SRC
     System/Angle.cpp
     System/Clock.cpp
+    System/Config.cpp
     System/Err.cpp
     System/FileInputStream.cpp
     System/MemoryInputStream.cpp
@@ -32,6 +33,12 @@ SET(SYSTEM_SRC
     System/Vector3.cpp
 )
 sfml_add_test(test-sfml-system "${SYSTEM_SRC}" SFML::System)
+target_compile_definitions(test-sfml-system PRIVATE
+    EXPECTED_SFML_VERSION_MAJOR=${SFML_VERSION_MAJOR}
+    EXPECTED_SFML_VERSION_MINOR=${SFML_VERSION_MINOR}
+    EXPECTED_SFML_VERSION_PATCH=${SFML_VERSION_PATCH}
+    EXPECTED_SFML_VERSION_IS_RELEASE=$<IF:$<BOOL:${VERSION_IS_RELEASE}>,true,false>
+)
 
 SET(WINDOW_SRC
     Window/ContextSettings.cpp


test/System/Config.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include <doctest.h>
+
+TEST_CASE("SFML/Config.hpp")
+{
+    SUBCASE("Version macros")
+    {
+        CHECK(SFML_VERSION_MAJOR == 3);
+        CHECK(SFML_VERSION_MINOR == 0);
+        CHECK(SFML_VERSION_PATCH == 0);
+        CHECK_FALSE(SFML_VERSION_IS_RELEASE);
```
Bromeon, line 12: Are tests of this form really useful? It's basically the exact same code as in `Config.hpp` and just means that when updating the version macros, it has to be done in two places.

In other words, what is this test trying to catch?

ChrisThrasher: It's making sure that more of the SFML API is correct and will remain correct. Do I expect this to ever be a source of bugs? No. But, it's easy to test and we ought to be testing as much of the API as possible. Just because certain aspects of the API are trivial doesn't mean we should intentionally never test them. 

These particular macros are actually never once used anywhere within SFML itself. If they got moved, removed, or their value changed, we have no way of knowing that through any other means than manual inspection which we know is prone to fail.

eXpl0it3r: One can argue that it's part of the public API and thus would make sense to test.  
On the other hand, the tests are rather questionable and need to be updated with every version change. 😄 

It doesn't hurt us that much, so I'm fine with keeping them.

eXpl0it3r:

test/System/Config.cpp
ChrisThrasher: I'm open to the possibility that I may be wrong about this being worth testing but luckily it's trivial to remove at any point in the future if we discover more reasons why it's not a good idea.

eXpl0it3r (approved):

test/System/Config.cpp
Bromeon: > Do I expect this to ever be a source of bugs? No. 

Then I would say there is no point in testing it. 

> But, it's easy to test and we ought to be testing as much of the API as possible.

It sounds philosophical, but why do we test? 😉

I personally would argue it's to prevent current and future bugs. So if there's very low probability of that happening, then tests lack a purpose in my opinion. There are of course other valid views, e.g. tests acting as a sort of API specification, or fulfilling a coverage metric, etc.

---

While I agree that tests are easy to write and thus tempting, you need to keep in mind each test comes with a tiny maintenance cost. Sure, adjusting two places with constants instead of one may only be a few seconds more, but the bigger issue is when a developer forgets it. Someone updates the `Config.hpp` version, pushes the code. Someone else notices that the CI fails and messages the first developer, who is already working on something else and has to context-switch again. Quickly half an hour wasted for no gain.

Of course I'm speculating, but things like these happen all the time. A lot of people forget that usual software engineering principles (e.g. don't repeat yourself, think about coupling) also apply to tests. It's just that in many cases, violations of them are tolerated because the benefits of having tests often outweigh the downsides.

The other question is, whether copy-pasting the implementation code in the test code is a good strategy. By design, such tests cannot find mistakes in the implementations because they're identical copies. The only thing they are good for is to detect _changes_ in implementation -- which then means the test will be updated to match the implementation 1:1 again, and we're back to square one.

Another issue I have with this test in particular is that it doesn't prevent the most likely scenario: a new SFML version is released and someone forgets to update the config header. Of course, the code alone cannot verify this. Maybe instead we can think of something on the "meta" level, e.g. a CI job that is run when a new Git tag is created and verifies that `Config.hpp` matches the tag's version?

ChrisThrasher: > The other question is, whether copy-pasting the implementation code in the test code is a good strategy.

I'm not sure where you got this idea of me pasting the implementation into the tests. That's never been the case.

> Another issue I have with this test in particular is that it doesn't prevent the most likely scenario: a new SFML version is released and someone forgets to update the config header.

Would you be more inclined to accept this if the tests used the CMake `PROJECT_VERSION` variables to ensure that the macros in Config.hpp always line up with what is specified in the call to `project`? I just pushed up this change. It means that if you change the version number whatsoever or change VERSION_IS_RELEASE without updating Config.hpp, the tests fail. I hadn't previously considering doing this but it does make these tests much much more valuable.

Bromeon: > I'm not sure where you got this idea of me pasting the implementation into the tests. That's never been the case.

That was unfortunate wording of mine, sorry for that -- I didn't mean this as literal copy-pasting, let alone laziness (for this I've followed your contributions for too long 😉).

I meant that implementation code is semantically identical with test code (impl declares constant X with value Y, test checks if constant X has value Y).


> Would you be more inclined to accept this if the tests used the CMake `PROJECT_VERSION` variables to ensure that the macros in Config.hpp always line up with what is specified in the call to `project`?

Absolutely! This suddenly adds a lot of value because it makes sure two independent components stay in sync! 👍👍

kimci86:
A few comments wouldn't hurt IMHO. A future reader might be wondering at first sight what those `EXPECTED` definitions are about.

ChrisThrasher:
Replaced an awkward CMake conditional with a more elegant, terse generator expression.

friendlyanon:
Why do it this way? You can just configure the header file with the variables and you only have maintain things in a single place. https://github.com/friendlyanon/SFML/blob/feature/cmake-rework/include/SFML/Config.hpp.in#L38-L41

ChrisThrasher:
@friendlyanon 

> Why do it this way?

No particular reason Config.hpp is entirely hand-written instead of autogenerated.

> You can just configure the header file with the variables and you only have maintain things in a single place.

Care to submit a PR that autogenerates Config.hpp as you've shown? I know you have a larger CMake PR in the works but I think the team would prefer multiple smaller PRs that make focused changes that are easier for reviewers to understand.

friendlyanon:
> Care to submit a PR that autogenerates Config.hpp as you've shown?

I have returned to working on it. You are more than welcome to pick things like this from it in the meantime. What's holding me back is indeed its size and how best take it apart into more digestable servings.

eXpl0it3r:

eXpl0it3r (approved):

vittorioromeo:
> Care to submit a PR that autogenerates Config.hpp as you've shown? 

Would that cause any issue with IDEs? E.g. autocompletion issues

friendlyanon:
`configure_file` does its thing during configure phase. You'd get squigglies only if the generated files are outputs of custom commands, because those happen at build time. Even then it's trivial to [make the squigglies go away](https://github.com/friendlyanon/cmake-init-gif-engine/blob/master/cmake/dev-mode.cmake#L28-L34).

eXpl0it3r:

eXpl0it3r:
Thanks for adding this 👍 
Feel free to create an issue if you want to continue to other part of the discussion.

PR #2163, opened 2022-07-12
Simplify how warnings are ignored on Apple platforms

ChrisThrasher:
## Description

GCC can't build SFML on Apple platforms so it's a given that within Apple-specific code, Clang is being used. Furthermore within files that are only compiled on Apple platforms, there is no need to use the preprocessor to check whether Apple is the current platform. That much is implied within those files.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

### DIFF
--- a/examples/cocoa/CocoaAppDelegate.mm
+++ b/examples/cocoa/CocoaAppDelegate.mm
@@ -32,13 +32,7 @@
 #define GREEN @"Green"
 #define RED   @"Red"
 
-#if defined(__APPLE__)
-#if defined(__clang__)
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-#elif defined(__GNUC__)
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-#endif
-#endif
 
 // Our PIMPL
 struct SFMLmainWindow
--- a/src/SFML/Audio/ALCheck.cpp
+++ b/src/SFML/Audio/ALCheck.cpp
@@ -33,11 +33,7 @@
 #include <utility>
 
 #if defined(__APPLE__)
-#if defined(__clang__)
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-#elif defined(__GNUC__)
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-#endif
 #endif
 
 namespace
--- a/src/SFML/Audio/ALCheck.hpp
+++ b/src/SFML/Audio/ALCheck.hpp
@@ -33,13 +33,8 @@
 #include <filesystem>
 
 #if defined(__APPLE__)
-#if defined(__clang__)
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-#elif defined(__GNUC__)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-#endif
 #endif
 
 #include <al.h>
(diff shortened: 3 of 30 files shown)


JonnyPtn:
gcc _could_ work on Apple, even if it doesn't currently. This will just make it harder for anybody who does want to use/fix the gcc implementation with no other benefit

eXpl0it3r:
Yeah, it currently doesn't hurt us and if someone manages to get SFML to build again with GCC, then we'd have to add those back in.

ChrisThrasher:
> Yeah, it currently doesn't hurt us and if someone manages to get SFML to build again with GCC, then we'd have to add those back in.

Happy to volunteer to do this if we ever get to that point. Hard to say how much farther in the future that may be though. In the meantime I don't think it's worth keeping around code which we know never gets executed.

eXpl0it3r:
I just changed:

https://github.com/SFML/SFML/blob/dd0f0ff4c6e0f487e2007192516caa5d6877e3cb/cmake/Macros.cmake#L29-L31

to

```cmake
        else()
            target_compile_options(${target} PRIVATE "-stdlib=libstdc++")
            target_link_libraries(${target} PRIVATE "-stdlib=libstdc++")
        endif()
```

and successfully built and linked SFML and its examples with GCC.

```
cmake -DCMAKE_C_COMPILER=/usr/bin/gcc -DCMAKE_CXX_COMPILER=/usr/bin/g++ -DSFML_BUILD_EXAMPLES=ON  -S . -B build
cmake --build build
```

So what exactly was the issue again with #1928? Why can't we utilize `libstdc++` for GCC on macOS?

ChrisThrasher:
> and successfully built and linked SFML and its examples with GCC.

That's awesome!

> So what exactly was the issue again with https://github.com/SFML/SFML/pull/1928? Why can't we utilize libstdc++ for GCC on macOS?

Besides what was explained in that PR, I simply never tried to explicitly specify libstdc++ with GCC. I perhaps foolishly assumed that it would use libstdc++ if we never specified any particular implementation. I wouldn't have guessed that explicitly specifying `libstdc++` fixes this but that's awesome. I'd much rather simply fix the GCC issues than remove all these pragmas.

Once there's a PR that fixes these issues and adds a GCC macOS CI job, I'll close this issue. There is still preprocessor code that can be removed though. In particular, we don't need `#if __APPLE__` in ObjectiveC++ files but I'll wait to remove that until these compiler errors are sorted.

ChrisThrasher:
I'll submit a PR soon that fixes that GCC issues on macOS. If that gets merged, I'll amend this PR because no matter what there is still unnecessary preprocessor code that can and should be removed.

ChrisThrasher:
> and successfully built and linked SFML and its examples with GCC.
> 
> ```
> cmake -DCMAKE_C_COMPILER=/usr/bin/gcc -DCMAKE_CXX_COMPILER=/usr/bin/g++ -DSFML_BUILD_EXAMPLES=ON  -S . -B build
> cmake --build build
> ```

On macOS, `/usr/bin/gcc` and `/usr/bin/g++` are just aliases for AppleClang. When I try use GCC as provided by brew, I get errors about `-stdlib=libstdc++` being an unrecognized command-line option.

ChrisThrasher:
Because Clang can recognize GCC pragmas, I removed all `#pragma clang` lines and exclusively used `#pragma GCC`. This lets the Clang builds still work but ensures zero rework should anybody ever figure out how to fix the issues using GCC on macOS. I think these latest changes address all the concerns reviewers had.

ChrisThrasher:
@eXpl0it3r Do you have any thoughts on the revised version of this PR that maintains potential GCC support?

eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:
Thanks for sticking to the whole back and forth 👍

PR #2166, opened 2022-07-17
Add SpriteBatch

Marioalexsan:
(previously appeared in forum posts as requests, and also in an issue / PR: https://github.com/SFML/SFML/issues/1802)

----

## Description

This is a draft PR for a sprite batching implementation. I have more work to do on the implementation, but I've reached a point where I have something I can show, and I'd like to hear thoughts from other people about this whole batching business.

The PR adds two classes: SpriteBatch, which implements the batching logic, and Batchable, an interface implemented by drawables that support batching. Batchable is similar in functionality with Drawable, and the classes that implement it send data to the batcher using a predefined method that takes in vertices and other state information.

The PR also adds the Batchable interface to Sprite, Text and Shape. In theory, anything that uses triangle primitives should work with the batcher, but I have only tested sprites so far.

The batcher has different modes that can be used for batching:
* `Deferred` groups sprites that use the same texture and follow each other in the same batch, which makes the end result identical with RenderTarget. However, when rendering many sprites that use the same texture in sequence, the batcher should be more efficient with the draw calls.
* `TextureSort` sorts the sprites by texture before grouping and rendering. This causes draw order to be lost; it's as if a "layer" is created for each group of sprites that have the same texture. This can be useful for drawing things that don't overlap, as in that case draw order is not relevant, and the batcher obtains maximum performance.
* `DepthSort` sorts the sprites by texture and an optional "depth" parameter before grouping and rendering. This is a compromise between `Deferred` and `TextureSort`, as you can create multiple layers that have proper ordering between each other, while drawables within a single layer follow `TextureSort` logic. This can be useful for drawing scenes that use distinct layers, such as Background, Foreground, GUI.
* There is also a fourth mode called `Immediate`, which just calls RenderTarget directly on every draw call.

Code that uses SpriteBatch requires calling `begin` and `end` for setting shaders and blend modes:

```cpp
// Begin a batch using the given settings
spriteBatch.begin(sf::SpriteBatch::BatchMode::DepthSort, window, nullptr, sf::BlendAlpha);

// Draw some stuff
spriteBatch.draw(sprite, depth);
spriteBatch.draw(text, depth);
spriteBatch.draw(shape, depth);

// Every call to begin() needs to have a matching end()
// This is also where the data is sent for rendering
spriteBatch.end();
```

## Tasks

* [x] Tested on Linux (checked if it compiles and runs)
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

You can test it by running the `examples/sprite_batch` demo I added. It shows the differences between regular rendering and batching, and also works as a basic benchmark for performance. It has controls for raising and lowering the sprite count (W and S), and for switching batching modes (F and G).


Bambo-Borris:
This is a feature I have seen discussed endlessly, and seen a couple of PRs for in the past. I think of the ones I've seen I currently like this approach the most, but I need to spend a bit more time looking into this and run the examples. But I like the approach, it feels consistent with other areas of SFML (namely how `sf::Drawable` stuff works internally). So I like consistency, and I like the feature idea. The area I probably will drill down into later is the 3 different modes for the `SpriteBatch`.

include/SFML/Graphics/SpriteBatch.hpp
```diff
@@ -0,0 +1,251 @@
...
+    ////////////////////////////////////////////////////////////
+    struct TriangleInfo
+    {
+        const Texture* texture = nullptr;
+        float          depth   = 0.0f;
+
+        TriangleInfo() = default;
+        TriangleInfo(const Texture* theTexture, float theDepth) : texture(theTexture), depth(theDepth)
+        {
+        }
```
ChrisThrasher, line 192: I believe we can omit these constructors if we want to rely on aggregate initialization instead.

include/SFML/Graphics/SpriteBatch.hpp
```diff
@@ -0,0 +1,251 @@
...
+    };
+
+    ////////////////////////////////////////////////////////////
+    // Member data
+    ////////////////////////////////////////////////////////////
+    // Batched Triangles
+    std::vector<TriangleInfo> m_triangleInfos  = {}; //!< Info about batched triangles
+    std::vector<Vertex>       m_vertices       = {}; //!< Vertices currently batched
+    std::vector<Vertex>       m_sortedVertices = {}; //!< Storage used for sorting vertices
```
ChrisThrasher, line 201: ` = {}` is unnecessary here. The default constructor will still get called.

src/SFML/Graphics/SpriteBatch.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+////////////////////////////////////////////////////////////
+void SpriteBatch::draw(const Vertex*    vertices,
+                       std::size_t      count,
+                       PrimitiveType    type,
+                       const Texture*   texture,
+                       const Transform& transform,
+                       float            depth)
+{
+    bool unsupportedType = type != PrimitiveType::TriangleFan && type != PrimitiveType::TriangleStrip &&
```
ChrisThrasher, line 71: Can this be marked `const`?

src/SFML/Graphics/SpriteBatch.cpp
```diff
@@ -0,0 +1,266 @@
...
+    {
+        // Immediate is the same as using a RenderTarget directly,
+        // so we don't use VertexBuffer here
+        m_renderTarget->draw(vertices, count, type, {m_blendMode, m_transform * transform, texture, m_shader});
+        return;
+    }
+
+    // Split into triangles if needed
+
+    if (type == PrimitiveType::TriangleStrip)
```
ChrisThrasher, line 93: Would this make more sense as a switch case than `if`/`else if` blocks?

src/SFML/Graphics/SpriteBatch.cpp
```diff
@@ -0,0 +1,266 @@
...
+    {
+        std::vector<std::size_t> indices(m_triangleInfos.size());
+        std::size_t              triangleCount = m_triangleInfos.size();
+
+        for (std::size_t i = 0; i < triangleCount; i++)
+            indices[i] = i;
+
+        if (m_batchMode == BatchMode::TextureSort)
+        {
+            const auto& comp = [&](const std::size_t a, const std::size_t b)
```
ChrisThrasher, line 168: What are the implications of binding a lambda to a `const&`? I've only ever see it bound to a const value.

src/SFML/Graphics/SpriteBatch.cpp
```diff
@@ -0,0 +1,266 @@
...
+            { return m_triangleInfos[a].texture < m_triangleInfos[b].texture; };
+
+            std::stable_sort(indices.begin(), indices.end(), comp);
```
ChrisThrasher, line 171: Any particular reason to stable sort instead of a non-stable sort?

src/SFML/Graphics/SpriteBatch.cpp
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+        // Create the array of sorted vertices
+        // We need them sorted so that we feed chunks to RenderTarget
+        m_sortedVertices.resize(m_vertices.size());
+
+        for (std::size_t i = 0; i < indices.size(); i++)
+        {
+            std::size_t newPos           = i * 3;
+            std::size_t oldPos           = indices[i] * 3;
```
ChrisThrasher, line 194: Can these be marked `const`?

src/SFML/Graphics/SpriteBatch.cpp
```diff
@@ -0,0 +1,266 @@
...
+        {
+            sendVertices(&m_sortedVertices[startIndex * 3],
+                         (nextIndex - startIndex) * 3,
+                         m_triangleInfos[indices[startIndex]].texture);
+        }
+    }
+
+    m_vertices.resize(0);
+    m_triangleInfos.resize(0);
+    m_sortedVertices.resize(0);
```
ChrisThrasher, line 231: What are you intentions here? Do you need to just empty the containers? Do you also need to deallocate all the elements? Resizing to zero seems like a more complicated way to simply call `.clear()`.

ChrisThrasher (changes requested):

include/SFML/Graphics/SpriteBatch.hpp
Bromeon: > I believe we can omit these constructors if we want to rely on aggregate initialization instead.

Do we want that, though? The rest of SFML has "proper" constructors. Having dedicated functions also allows to do custom logic in the future (e.g. validations).

examples/sprite_batch/Batching.cpp
```diff
@@ -0,0 +1,265 @@
...
+    window.setVerticalSyncEnabled(true);
+
+    std::vector<sf::Texture> textures;
+    sf::Texture              tempTexture;
+
+    int currentTex = 0;
+    while (tempTexture.loadFromFile(resourcesDir() / ("Tex" + std::to_string(currentTex) + ".png")))
+    {
+        textures.push_back(std::move(tempTexture));
+        tempTexture = sf::Texture();
```
Bromeon, line 59: This is probably not going to be necessary once the move constructor for [`sf::Texture`](https://github.com/SFML/SFML/blob/master/include/SFML/Graphics/Texture.hpp) is implemented, but I'm also fine with leaving it for explicitness.

Could also be that some compilers issue a "access to moved variable" warning otherwise.

examples/sprite_batch/Batching.cpp
```diff
@@ -0,0 +1,265 @@
...
+int main()
+{
+    std::srand(static_cast<unsigned int>(std::time(nullptr)));
+
+    // Define some constants
+    const int         gameWidth         = 800;
+    const int         gameHeight        = 600;
+    const int         maxTimesSaved     = 60;
+    const int         sameTextureChance = 85;
+    const std::size_t minSprites        = 0;
```
Bromeon, line 43: Counts like `maxTimesSaved` could be `size_t`.

examples/sprite_batch/Batching.cpp
```diff
@@ -0,0 +1,265 @@
...
+                            batchMode = modes[i == 0 ? modeCount - 1 : i - 1];
+                        }
+                        else
+                        {
+                            batchMode = modes[i == modeCount - 1 ? 0 : i + 1];
+                        }
+
+                        break;
+                    }
+                }
```
Bromeon, line 138: I wonder if it wouldn't be easier to understand if the selection of the batch mode would be done more explicitly and directly. Using the array as a middleman might not even be necessary.

Basically a `switch` with explicit state transitions. and avoid the complicated index operations.

Bromeon (commented):
Thanks a lot for this huge effort!
First quick feedback -- I'll need to reserve some time to look if/how this generally fits into SFML's scope.

What is missing in your description is a motivation/rationale for this feature. It may be obvious to you ("performance"), but could you elaborate more precisely where you encountered the biggest limits with current SFML, and how this solution would address those?

I like that you added quite some description to each of the 4 modes. What would the decision process for a user look like (e.g. how to choose between Deferred/Immediate/no batching)?

Could the `begin()` + `end()` method calls be replaced with an RAII guard?
```cpp
{
    sf::BatchGuard guard = batch.begin();
    ...
    ...
} // automatic end()
```

src/SFML/Graphics/SpriteBatch.cpp
Marioalexsan: `unsupportedType`? Is there any benefit from marking it as `const`?

include/SFML/Graphics/SpriteBatch.hpp
Marioalexsan: I'm not sure if this simple data struct will ever need special logic, though.

src/SFML/Graphics/SpriteBatch.cpp
Marioalexsan: The `switch` version looks a bit more cluttered to me.

src/SFML/Graphics/SpriteBatch.cpp
ChrisThrasher: Just the general benefits of liberally using `const` on stack variables. Makes the code easier to read and more maintainable as it communicates to future developers that this variable is not and should not be mutated.

src/SFML/Graphics/SpriteBatch.cpp
ChrisThrasher: Using `switch` is more expressive and means that if a new enum is added, the compiler will be sure to emit a warning about not handling that new enum. It makes the code easier to understand and easier for others to modify months or years from now. If the language already offers a tailor-made solution, we ought to use it.

src/SFML/Graphics/SpriteBatch.cpp
Marioalexsan: I'm not sure, but I don't think there's much difference. I'll change it to regular `const` though.

src/SFML/Graphics/SpriteBatch.cpp
Marioalexsan: Just emptying it. I'll change it to `clear`.

src/SFML/Graphics/SpriteBatch.cpp
ChrisThrasher: Don't rely on `std::sort` happening to be stable if stable sorting is really what you need. That property is liable to change across implementations and over time since `std::sort` makes no stability guarantees.

Marioalexsan:
@Bromeon I have to admit that I've yet to do my homework on the reasoning part. What I planned to do was to try and use the SpriteBatch inside open source SFML games to see if it would improve performance, particularly in cases where it's shy of being a consistent 60 FPS. It would also help with reducing or eliminating the manual batching inside VertexArrays that people do. 

For instance, [At odds](https://peoplepancakes.itch.io/at-odds) seems to lag a bit on my crappy laptop. If the batcher would nudge the performance above the 60 FPS mark while being easy to use (or swap to), that would be great news. However, I still need to analyze the source code and whatnot.

Marioalexsan:
@Bromeon Regarding the batch modes, first I need to mention that Immediate is the same as not using batching, since it just calls the RenderTarget directly. In hindsight, I'm going to remove it since it's not really useful.

Deferred is similar to regular drawing, since draw order is preserved (i.e. same as draw call order). This makes it easy to change RenderTarget logic to SpriteBatch logic using Deferred, and should provide performance enhancements when drawing many things in a row that use the same texture. So, an use case would be drawing all entities that use a wooden box texture, but which also have physics and can be rotated and moved around (and thus can't be inserted into a VertexArray or something).

TextureSort is useful if your application is structured into layers, such as top-down or side-scroller games. Within a layer, it's likely that objects don't overlap each other (or the order isn't very important), so if they use different textures, you can reorder them by texture to create bigger batches to send for rendering (thus you get some nice performance improvements).

DepthSort is a combination of the previous two modes, that is you can specify an explicit draw order if needed via the depth parameter, but you can also arrange your objects in multiple layers by using preset depths for each layer.

So, to summarize:
- If you care for strict draw order or render many things using one texture, use Deferred
- If you don't care for order and render many things using different textures, use TextureSort
- If you need a partial order and render many things using different textures, use DepthSort and specify the order you need

src/SFML/Graphics/SpriteBatch.cpp
Bromeon: [Off-topic] Arguably, this might be a case of "wrong defaults" in C++. Other languages have stable sorting as the only algorithm (e.g. [Java](https://docs.oracle.com/javase/9/docs/api/java/util/Collections.html#sort-java.util.List-) or [Python](https://docs.python.org/3/library/functions.html#sorted)), or they explicitly name the unstable sort differently ([Rust](https://doc.rust-lang.org/std/primitive.slice.html#method.sort_unstable)).

I agree that stable sorting is probably more intuitive in this case.

Bromeon:
> For instance, [At odds](https://peoplepancakes.itch.io/at-odds) seems to lag a bit on my crappy laptop. If the batcher would nudge the performance above the 60 FPS mark while being easy to use (or swap to), that would be great news. However, I still need to analyze the source code and whatnot.

Any such insights would be very interesting! I think there's a lot of potential, but since the batching API adds quite some complexity, it would be good to be sure it's beneficial in typical cases 🙂 

---

> Regarding the batch modes, first I need to mention that Immediate is the same as not using batching, since it just calls the RenderTarget directly. In hindsight, I'm going to remove it since it's not really useful.

I think it is useful, just needs to be documented that it's equivalent to using no batching. Why useful? Because it allows to quickly switch implementations and compare performances without rewriting the whole rendering code.

Also, your description following that comment is great! Do you think you could add this to the documentation? 

---

> Regarding the batch guard, I'm not sure if `begin()` and `end()` is something that is necessarily restricted to a scope. Cases like this could be possible:

I have the feeling these scenarios (having no clear owner of the batch) are _precisely_ the situations which are hard to reason about and cause potential errors. Stuff like early returns or exceptions will immediately cause bugs that can be difficult to debug.

If batching should be "stackable", then maybe we should provide a dedicated API, which also helps restoring previous state. But one probably loses most of the advantages when switching batch anyway, no?

Marioalexsan:
I pushed a modified implementation along with some changes to `examples/sprite_batch`. It's a bit random, but I feel like the current approach works better. I also tried to apply the suggestions given, although some are still in my backlog.

I've removed the `begin()` / `end()` approach in favor of making SpriteBatch work similar to a Drawable. Now, you set the batch mode using `setBatchMode(batchMode)`, batch objects using `batch(batchable, depth)`, and use the Drawable interface for rendering to RenderTarget. There is also a `clear()` method that removes all drawables from the batch. The current logic is like this:

```cpp
// Set batch strategy
spriteBatch.setBatchMode(batchMode);

spriteBatch.batch(sprite, depth);
spriteBatch.batch(text, depth);
spriteBatch.batch(shape, depth);

// Draw the batch
// RenderStates texture is ignored here, since the batch sets the textures itself
spriteBatch.draw(window, sf::RenderStates(sf::BlendAlpha, sf::Transform::Identity, nullptr, shader));

// Optionally clear the batch for rendering on the next frame
spriteBatch.clear();
```

The advantage of the new implementation is that you can now keep the batch and reuse it for static scenes, which was not possible in the previous version. It also prepares the batches only once, and reuses the result if multiple calls to `draw()` are made.

Regarding the batch example, I've added texts, shapes and shapes with no texture for checking more drawable types. As expected, the batcher struggles with dynamic texts and shapes, since they're high vertex count drawables and kind of mini-batches by themselves. However, for static stuff, inserting into a batch and reusing said batch works almost in the same way as manually batching using VertexArray, and works pretty well if sorting is applied.

examples/sprite_batch/Batching.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+            if (count % 250 == 0)
+                depth -= 0.01f;
+        }
+
+        batch.end();
+
+        auto time = clock.restart();
+
+        timeHistory.push_back(static_cast<float>(time.asMicroseconds()) / 1000.f);
```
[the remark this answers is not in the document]
Marioalexsan: I've switched to `asSeconds()`, but I'm still doing `timeReference.asSeconds() * 1000.0f` since I want floating point milliseconds. Not sure if there's a better way to do it.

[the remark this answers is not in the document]
ChrisThrasher: Ideally you'd store `sf::Time` objects instead of storing floats. The whole point of `sf::Time` is that you don't have to weak, primitive types to store durations.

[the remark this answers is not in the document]
Marioalexsan: Yeah, that is what I'm doing right now. However, I'm doing the conversion to display the values to window:
```
debugInfoText.setString(
            std::string(batchEnabled ? "true" : "false") + "\n" +
            getDisplayString(benchmarkedDrawable) + "\n" +
            std::to_string(timeReference.asSeconds() * 1000.0f) + "\n" +
            std::to_string(average.asSeconds() * 1000.0f) + "\n" +
            std::to_string(drawableCount) + "\n" +
            getDisplayString(batchMode) + "\n" +
            (useStaticBatch ? "Static batch is active" : ""));
```

examples/sprite_batch/Batching.cpp
```diff
@@ -0,0 +1,446 @@
...
+{
+    Sprite,
+    Text,
+    Shape,
+    ShapeNoTexture
+};
+
+std::string getRandomString(std::mt19937& mt)
+{
+    static std::array<std::string, 8> texts =
```
ChrisThrasher, line 40: Can this be marked `const`?

examples/sprite_batch/Batching.cpp
```diff
@@ -0,0 +1,446 @@
...
+         "'Tis but a flesh wound.",
+         "SFML Test Demo Thingy Stuff (Epic)",
+         "?!?!?!111(one)"};
+
+    return texts[std::uniform_int_distribution<std::size_t>(0, texts.size() - 1)(mt)];
+}
+
+std::string getDisplayString(sf::SpriteBatch::BatchMode batchMode)
+{
+    static std::map<sf::SpriteBatch::BatchMode, std::string> modeDisplay =
```
ChrisThrasher, line 58: Can this be marked `const`?

examples/sprite_batch/Batching.cpp
```diff
@@ -0,0 +1,446 @@
...
+        {{sf::SpriteBatch::BatchMode::Deferred, "Deferred"},
+         {sf::SpriteBatch::BatchMode::TextureSort, "TextureSort"},
+         {sf::SpriteBatch::BatchMode::DepthSort, "DepthSort"}};
+
+    if (modeDisplay.find(batchMode) != modeDisplay.end())
+    {
+        return modeDisplay[batchMode];
+    }
+
+    return std::string("Unknown");
```
ChrisThrasher, line 68: The explicit cast to `std::string` is not necessary.

examples/sprite_batch/Batching.cpp
```diff
@@ -0,0 +1,446 @@
...
+}
+
+std::string getDisplayString(BatchDrawables drawableMode)
+{
+    static std::map<BatchDrawables, std::string> modeDisplay = {{BatchDrawables::Sprite, "Sprite"},
```
ChrisThrasher, line 73: Can this be marked `const`?

examples/sprite_batch/Batching.cpp
```diff
@@ -0,0 +1,446 @@
...
+            sprites[i].setRotation(sprites[i].getRotation() + amount);
+            texts[i].setRotation(texts[i].getRotation() + amount / 8);
+            shapes[i].setRotation(shapes[i].getRotation() + amount);
+            shapesNoTexture[i].setRotation(shapesNoTexture[i].getRotation() + amount);
+        }
+
+        // Batch drawables
+        clock.restart();
+
+        float       depth = 1.00f;
```
ChrisThrasher, line 320: The extra two trailing zeros aren't necessary. `1.f` means the same thing.

examples/sprite_batch/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+
```
ChrisThrasher, line 1: Can you remove this leading empty newline? I recently removed all these empty leading newlines from all the example build scripts.

examples/sprite_batch/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+set(SRCROOT ${PROJECT_SOURCE_DIR}/examples/sprite_batch)
```
ChrisThrasher, line 2: `SRCROOT` is not necessary. Just use relative file paths and we can avoid the complexity of creating and using this variable. I recently removed all use of `SRCROOT` in the examples.

examples/sprite_batch/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+        ${CMAKE_CURRENT_SOURCE_DIR}/resources/Tex3.png
+        ${CMAKE_CURRENT_SOURCE_DIR}/resources/Tex4.png
+        ${CMAKE_CURRENT_SOURCE_DIR}/resources/Tex5.png
+        ${CMAKE_CURRENT_SOURCE_DIR}/resources/Tex6.png
+        ${CMAKE_CURRENT_SOURCE_DIR}/resources/Tex7.png
+        ${CMAKE_CURRENT_SOURCE_DIR}/resources/Tex8.png)
+    set_source_files_properties(${RESOURCES} PROPERTIES MACOSX_PACKAGE_LOCATION Resources)
+endif()
+
+# define the pong target
```
ChrisThrasher, line 20: Pong target? I think you forgot to update this comment.

include/SFML/Graphics/Batchable.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+#ifndef SFML_BATCHABLE_HPP
+#define SFML_BATCHABLE_HPP
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#include <SFML/Graphics/Export.hpp>
+
+#include <functional>
```
ChrisThrasher, line 33: What is `<functional>` being used for?

src/SFML/Graphics/Shape.cpp
```diff
@@ -148,6 +149,28 @@ FloatRect Shape::getGlobalBounds() const
 }
 
 
+////////////////////////////////////////////////////////////
+void Shape::batch(SpriteBatch& spriteBatch, float depth) const
+{
+    // Draw the shape itself
+    {
+        const std::size_t shapeCount = m_vertices.getVertexCount();
+        const Vertex*     shapePtr   = shapeCount > 0 ? &m_vertices[0] : nullptr;
```
ChrisThrasher, line 158: `m_vertices.data()` may be a more expressive way to get a pointer to the first element than taking the address of index zero.

src/SFML/Graphics/SpriteBatch.cpp
```diff
@@ -0,0 +1,267 @@
...
+#include <SFML/System/Err.hpp>
+
+#include <algorithm>
+
+namespace sf
+{
+////////////////////////////////////////////////////////////
+void SpriteBatch::setBatchMode(BatchMode batchMode)
+{
+    if (m_triangles.size() > 0)
```
ChrisThrasher, line 41:
```suggestion
    if (!m_triangles.empty())
```

src/SFML/Graphics/SpriteBatch.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+        case PrimitiveType::Triangles:
+            for (std::size_t i = 2; i < count; i += 3)
+            {
+                pushTriangle(vertices[i - 2], vertices[i - 1], vertices[i], transform, texture, depth);
+            }
+            break;
+
+        default:
+            // Shouldn't ever be hit
```
ChrisThrasher, line 103: Instead of this comment, you can simply omit this default case. If you add a new enum, the compiler will warn that there is an unhandled enum value.

src/SFML/Graphics/SpriteBatch.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+    m_updateRequired = true;
+}
+
+
+////////////////////////////////////////////////////////////
+void SpriteBatch::updateBatch() const
+{
+    // Sanity check
+    if (m_triangles.size() == 0)
```
ChrisThrasher, line 157:
```suggestion
    if (m_triangles.empty())
```

ChrisThrasher: Better yet, just use `assert` if this is something that you think will always be false. Debug builds will catch the error but release builds won't incur the performance penalty of this check.

src/SFML/Graphics/SpriteBatch.cpp
```diff
@@ -0,0 +1,267 @@
...
+            m_batches.emplace_back(lastTexture, (m_triangles.size() - startIndex) * 3);
+        }
+    }
+
+    else if (m_batchMode == BatchMode::TextureSort || m_batchMode == BatchMode::DepthSort)
+    {
+        std::vector<std::size_t> indices(m_triangles.size());
+
+        for (std::size_t i = 0; i < m_triangles.size(); i++)
+            indices[i] = i;
```
ChrisThrasher, line 192: [`std::iota`](https://en.cppreference.com/w/cpp/algorithm/iota) may come in handy here if you just want a bunch of sequentially increasing values.

src/SFML/Graphics/SpriteBatch.cpp
```diff
@@ -0,0 +1,267 @@
...
+        {
+            const auto comp = [&](const std::size_t a, const std::size_t b)
+            { return m_triangles[a].texture < m_triangles[b].texture; };
+
+            std::stable_sort(indices.begin(), indices.end(), comp);
+        }
+
+        else
+        {
+            const auto comp = [&](const std::size_t a, const std::size_t b)
```
ChrisThrasher, line 204: Instead of using this default capture mode (`[&]`) is it feasible to list the things you want to capture?

src/SFML/Graphics/SpriteBatch.cpp
```diff
@@ -0,0 +1,267 @@
...
+    else if (m_batchMode == BatchMode::TextureSort || m_batchMode == BatchMode::DepthSort)
+    {
+        std::vector<std::size_t> indices(m_triangles.size());
+
+        for (std::size_t i = 0; i < m_triangles.size(); i++)
+            indices[i] = i;
+
+        if (m_batchMode == BatchMode::TextureSort)
+        {
+            const auto comp = [&](const std::size_t a, const std::size_t b)
```
ChrisThrasher, line 196: Instead of using this default capture mode (`[&]`) is it feasible to list the things you want to capture?

src/SFML/Graphics/SpriteBatch.cpp
```diff
@@ -0,0 +1,267 @@
...
+        }
+
+        else
+        {
+            const auto comp = [&](const std::size_t a, const std::size_t b)
+            {
+                if (m_triangles[a].depth != m_triangles[b].depth)
+                    return m_triangles[a].depth > m_triangles[b].depth;
+                else
+                    return m_triangles[a].texture < m_triangles[b].texture;
```
ChrisThrasher, line 209: Omitting the braces in this `if/else` helps readability. I'd recommend removing the braces for other simple loops and conditions you've added which only contain a single expression. It halves the total line count and keeps relevant code near each other. Because we have clang-format to check formatting, there is no risk of accidentally writing something like 
```cpp
if (condition)
    run1();
    run2(); // Uh oh, this always get run no matter the value of condition
```

ChrisThrasher (changes requested):

examples/sprite_batch/CMakeLists.txt
Marioalexsan: Yeah, I just plagiarized the cmake script from that example project 😓

src/SFML/Graphics/SpriteBatch.cpp
Marioalexsan: Omitting it would mean I'm not handling the line and point primitives. Maybe I should swap the default case with three cases for those enums, to keep the warns for future new enums?

ChrisThrasher: My bad. I assumed all cases were being handled. If you're truly confident that the default case will never get hit, add an `assert` in there to catch any instances where it might accidentally get called.

include/SFML/Graphics/Batchable.hpp
Marioalexsan: Seems like I left it over from a different, more cursed variant of the batchable interface.

src/SFML/Graphics/Shape.cpp
Marioalexsan: That's a `VertexArray` though. I don't think it has such a method.

src/SFML/Graphics/SpriteBatch.cpp
Marioalexsan: Do I just put `assert(false)` in the default case, then?

src/SFML/Graphics/SpriteBatch.cpp
Marioalexsan: This one is a runtime check, although I'm not sure if it's needed given the method's logic.

src/SFML/Graphics/SpriteBatch.cpp
ChrisThrasher: Yeah that should be all you need. You ought to add a message too like this:
```cpp
assert(("This is what happened", false));
```
It (ab)uses the comma operator to make sure you get some useful output when that assert fires. 
```
example.cpp:5: int main(): Assertion `("This is what happened", false)' failed.
```

src/SFML/Graphics/SpriteBatch.cpp
ChrisThrasher: If you're not sure it's needed, then it's a great candidate for an assert. `assert`s are runtime checks.

src/SFML/Graphics/SpriteBatch.cpp
Marioalexsan: Would capturing `this` work?

ChrisThrasher: No because that's the other default capture mode. Try to explicitly list the objects you're capturing. Sometimes that is not feasible but at least try to make it work as to minimize how much stuff gets put into the scope of the lambda.

src/SFML/Graphics/SpriteBatch.cpp
Marioalexsan: I agree in the case of simple loops / `if`s, although in some cases I feel like the extra space given by adding braces helps readability.

src/SFML/Graphics/SpriteBatch.cpp
Marioalexsan: It's not needed in the sense that it may be redundant. `m_triangles` is supposed to have a varying size (including 0 triangles) 😅

ChrisThrasher: If it's perfectly valid to have zero size, then it's not a sanity check so you can remove that comment. To me, "sanity check" connotes code that is likely unnecessary. If you always need to handle that case, it's not a sanity check. It's just code that is required to make the implementation correct.

src/SFML/Graphics/SpriteBatch.cpp
Marioalexsan: I think capturing `m_triangles` directly won't work, but I can do something like this:
```cpp
auto&      triangles  = this->m_triangles;
const auto comp      = [triangles](const std::size_t a, const std::size_t b)
{
    if (triangles[a].depth != triangles[b].depth)
        return triangles[a].depth > triangles[b].depth;
    else
        return triangles[a].texture < triangles[b].texture;
};
```

ChrisThrasher: Hmm, maybe you're right to use `[this]`. You can copy `m_triangles` into the lambda but that's wasteful. I don't hate the idea of making a local reference and using that instead. In either case, it's more explicit than using `[&]` which will also capture everything in the current scope.

Marioalexsan: It seems like it's possible to capture `m_triangles` with an initializer such as `[&triangles = m_triangles]`. Maybe I'll use that instead.

ChrisThrasher: I think that's just making a non-const copy.

src/SFML/Graphics/SpriteBatch.cpp
Marioalexsan: CI is complaining about the message, it seems.
```
/home/runner/work/SFML/SFML/src/SFML/Graphics/SpriteBatch.cpp:106:21: error: left operand of comma operator has no effect [-Werror=unused-value]
  106 |             assert(("A non-triangle primitive type was encountered when decomposing into triangles.", false));
```

ChrisThrasher: Oh, that's too bad. I guess we'll just have to make do with no message.

Marioalexsan: I fixed it by casting the string to `void`. Although now that I think of it, I used a C style cast to do that.

ChrisThrasher:
I rebased on `master` and fixed some issues related to style changes that have happened since this PR was last updated. That includes some clang-tidy fixes.

ChrisThrasher:

viencent2:
Ah, I have implemented similar system for shapes and sprites. With barches and layers.
Seems like a lot of work that is already done in raylib

PR #2167, opened 2022-07-18
Remove unnecessary CMake version checks

ChrisThrasher:
## Description

Now that the project requires CMake version 3.15, all of these checks are guaranteed to eveluate one way or the other so they can be removed.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android


eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:
Thanks for the clean up 🧹

PR #2168, opened 2022-07-18
Added config for Android arm64-v8a and x86_64

Waldegger:
Added CMake flags for Android ABIs arm64-v8a and x86_64


eXpl0it3r:
Should be part of #2165 / "superseded" by #2165

Issue #1991, opened 2022-01-31
Vendoring dependencies presents maintenance issues


ChrisThrasher:
SFML currently vendors a number of binaries and headers for external dependencies. See the [`extlibs` ](https://github.com/SFML/SFML/tree/master/extlibs)directory. Vendoring provides a more convenient developer experience by removing the need to install dependencies at the cost of a few problems.

1. Increased maintenance burden. Rebuilding binaries as needed for new architectures or OSes needs to be reproducible and documented and manually performed by someone. 
2. Potential issues with dependency graphs. If someone depends on SFML and has another direct dependency on a different version of one of SFML's dependencies then it can be ambiguous which version of the library to use.
3. Potential ABI issues with different compilers being used to build SFML versus the vendored binaries.
4. Binaries cannot be diff'd efficiently by Git causing the repo size to bloat as binaries get updated.
5. External headers add tens of thousands of lines of code to the repo.
6. Separating 3rd party code from 1st party code is hard when it lives in the same repo. We must be sure tools like `clang-format` are never applied to vendored code.

SFML uses vendored dependencies for only a select few operating systems. For example, Linux users end up using `find_package`, CMake's canonical dependency retrieval mechanism. Users on Windows instead get the vendored dependencies. The key discriminator between whether a given OS uses `find_package` or vendored dependencies is the presence or absence of an official package manager. We want to keep this behavior.

---

What I propose as a first step is that we remove all binaries from `extlibs` and replace them with the source code for a given dependency as retrieved by CMake's `FetchContent` module but only on OSes that aren't already using `find_package`. `FetchContent` will automatically download and build the project and provide the SFML build tree with access to all of the CMake targets from a given dependency.

This also entails a fully target-based approach for expressing dependencies. No more modifying `CMAKE_INCLUDE_PATH` or `CMAKE_LIBRARY_PATH`. No matter how a dependency is retrieved, we must depend on a target from that project. This means that users of a given dependency don't have to know anything about how that dependency is retrieved. Separating the retrieval of dependencies from its usage opens the door for easier future changes to how dependencies are handled and brings with it the other benefits of modern target-based CMake code.

I still have more research to do into the existing dependency management scheme and the specifics of SFML's particular dependencies but I hope this issue helps start the conversation so we can start making steps towards improving this situation.

Bromeon:
Those are some very good points.
See also: https://en.sfml-dev.org/forums/index.php?topic=25053

[...]

PR #2169, opened 2022-07-18
Remove vendored doctest header

ChrisThrasher:
## Description

This PR uses the CMake FetchContent module to automatically download a specific release of Doctest and build it alongside the rest of SFML. This has the benefit of letting us removing another ~7000 SLOC from the repo. When updating Doctest in the future, the diff will be a single line as opposed to the status quo that requires many thousands of lines of changes.

Because we now have access to the entire repo, we also have access to the additional CMake modules Doctest provides including the module which includes the `doctest_discover_tests` function. I expect this will come in handy later. It isn't available to those who only have a copy of the doctest.h. See [this commit](https://github.com/ChrisThrasher/SFML/commit/e969658c6fb2af59a6cbc129131b78d8181eafe4) to see what that would look like.

Because the doctest header is now no longer a `SYSTEM` include, there were some warnings generated that I had to fix, hence the minor changes to the source code of the tests. 

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android


eXpl0it3r:

Bromeon:
I think `FetchContent` is nice and probably the way to go in the future, possibly also for other libraries. We should however be aware of the downsides mentioned in https://github.com/SFML/SFML/issues/1991:

> [...] `FetchContent` at build time would make us rely on several third-party resources to be available, or the build is broken. C++ doesn't have a centralized, stable, always-up dependency repository like crates.io, so we would have to trust each provider individually. [...] Downtime is not the only problem. GitHub repos change, are renamed, deleted, transferred, ...
> 
> I don't like the thought of people's build to break because one of the dependency maintainer decides to changes something or makes a mistake. You probably heard of the [leftpad](https://www.theregister.com/2016/03/23/npm_left_pad_chaos/) disaster, or just a few weeks ago, [colors and faker](https://www.bleepingcomputer.com/news/security/dev-corrupts-npm-libs-colors-and-faker-breaking-thousands-of-apps/). The first case made every single dependent code break, the second still broke everything with "use latest version" policy. And those are published in npm, likely the most widespread package manager, which at least has some basic stability. Unlike a GitHub repo which can change entirely through git push or two clicks.
> 
> And those are just some of the issues, security is another one. If any contributor with push access in any of these repos decides to go rogue (or is hacked), and even if this only remains undetected for a few hours, a SFML user can end up downloading malware.
>
> If we don't want to version external code in SFML's own repo, we can always create a separate SFML/extlibs repo and let the user pull dependencies from there. But this remains in our organization, under our control, and dependency updates are a conscious decision.

So what do you guys think about a `SFML/extlibs` repo? It can have a structure like:
```
Repos.txt
doctest/
freetype2/
ogg/
OpenAL/
stb_image/
...
```
and the `Repos.txt` would contain each link to a dependency on a separate line, which then could be cloned and checked into Git by some CI job.

Bromeon:
To clarify my last comment; this doesn't have to be a blocker for this PR -- the upstream repo for `FetchContent` could be moved to SFML in a later step. I'm fine with directly depending on `doctest` as an *intermediate* solution (to unblock development quicker).

But I'd like us to have a somewhat aligned vision of SFML's supply chain, and the general direction we want to take to migitate such risks.

ChrisThrasher:
> But I'd like us to have a somewhat aligned vision of SFML's supply chain, and the general direction we want to take to migitate such risks.

I can't make up my mind on what solution I most like so I don't have much to share at the moment, sorry. I'm following along but don't have much to contribute on that topic.

eXpl0it3r (approved):

eXpl0it3r:
Personally, I don't think supply chain issues are as big of an issue, but let's maybe continue the discussion over at #1991

eXpl0it3r:
Thanks for this initial step of using FetchContent 🙂

PR #2170, opened 2022-07-19
Separate test utilities between modules

kimci86:
## Description

For now, test utilities for all SFML modules are compiled in a single library, which means for example that tests for the System module are linked with Graphics testing utilities. That was ok because apparently the linker ignores what is not used in a static library, but that looks like a hack to me. It starts to be a problem with linker errors in the work in progress [feature/test_graphics](https://github.com/SFML/SFML/tree/feature/test_graphics) branch.

To fix the issue, one solution would be to have a single test executable for all modules instead of one per module.
Another solution is to have specialized test utility libraries for each module.
This is the approach I took with this PR.

As a bonus, it then becomes simple to enable testing when not all modules are built, which was required since #2050.
#2050 was motivated by simplicity, so I hope my PR adds only a reasonable amount of complexity.

## How to test this PR?

CI


ChrisThrasher:
My biggest issue with this PR is that it's solving two somewhat unrelated problems. The first is the problem that all tests are getting symbols from potentially unrelatedly libraries and that is prone to cause issues. At a minimum, it's just a bit ugly to link Graphics symbols into the System tests. The second problem is the desire to minimize compile time for those who aren't building all of SFML but still want to build the tests. I have previously explained why I don't like supporting this use case in #2050 so I won't repeat myself.

I can get behind solving the first problem is the implementation is good. My most significant request is that if we do make the test infrastructure more complicated, we do it with a clear plan that will lead to complexity leveling out soon. I hope that as tests get more sophisticated, our testing infrastructure won't need continuous hacks to keep up.

cmake/Macros.cmake
```diff
@@ -301,11 +301,39 @@ macro(sfml_add_example target)
...
+
+    # set the target's folder (for IDEs that support it, e.g. Visual Studio)
+    set_target_properties(${target} PROPERTIES FOLDER "Tests/Runners")
+
+    # link the target to its SFML dependencies and its test runner library
+    target_link_libraries(${target} PUBLIC ${THIS_DEPENDS})
+
+    set_target_warnings(${target})
+endfunction()
+
```
ChrisThrasher, line 330: Can we move this to the test/ directory since nothing outside of the unit tests needs to use this? For starters we can put this in test/CMakeLists.txt.

cmake/Macros.cmake
```diff
@@ -301,11 +301,39 @@ macro(sfml_add_example target)
...
 # add a new target which is a SFML test
 # example: sfml_add_test(sfml-test
-#                           ftp.cpp ...
-#                           SFML::Network)
-function(sfml_add_test target SOURCES DEPENDS)
+#                        ftp.cpp ...
+#                        SFML::Network
+#                        sfml-test-runner-system)     # Library providing the test runner and related utilities
+function(sfml_add_test target SOURCES DEPENDS TEST_RUNNER)
```
ChrisThrasher, line 336: Where does the "runner" terminology come from? To me that connotes CI runners, the machines that execute our CI jobs. Could we change that to some term that reflects that each "runner" is in fact a helper library?

cmake/Macros.cmake
```diff
@@ -316,8 +344,8 @@ function(sfml_add_test target SOURCES DEPENDS)
     # set the target's folder (for IDEs that support it, e.g. Visual Studio)
     set_target_properties(${target} PROPERTIES FOLDER "Tests")
 
-    # link the target to its SFML dependencies
-    target_link_libraries(${target} PRIVATE ${DEPENDS} sfml-test-main)
+    # link the target to its SFML dependencies and its test runner library
+    target_link_libraries(${target} PRIVATE ${DEPENDS} ${TEST_RUNNER})
```
ChrisThrasher, line 348: Each utils library already needs to link to the library under test so I don't think this function needs to specify any more than a single link library dependency.

test/CMakeLists.txt
```diff
@@ -23,48 +23,63 @@ SET(SYSTEM_SRC
     System/Vector2.cpp
     System/Vector3.cpp
 )
-sfml_add_test(test-sfml-system "${SYSTEM_SRC}" SFML::System)
+sfml_add_test(test-sfml-system "${SYSTEM_SRC}" SFML::System sfml-test-runner-system)
```
ChrisThrasher, line 26: Can't we remove `SFML::System` since it's already an public dependency of `sfml-test-runner-system`?

ChrisThrasher (changes requested):

kimci86:
Thank you for your feedback @ChrisThrasher.

> My biggest issue with this PR is that it's solving two somewhat unrelated problems. The first is the problem that all tests are getting symbols from potentially unrelatedly libraries and that is prone to cause issues. At a minimum, it's just a bit ugly to link Graphics symbols into the System tests. The second problem is the desire to minimize compile time for those who aren't building all of SFML but still want to build the tests. I have previously explained why I don't like supporting this use case in #2050 so I won't repeat myself.

You are right. Regarding the first described problem, I see two solution:
- merging test executables into one.
- splitting the test utility library into several.

I don't have strong arguments for one or the other. My preference went to the second option. That seems more consistent with the fact that SFML is a collection of libraries.

If we choose the second solution like in this PR, then solving the second described problem is as simple as four `if`/`endif` pairs, so at this point, why not.
Indeed, contributors interested in running tests might usually build all modules, but in some rare cases such as working on the raspberry pi, disabling modules can save minutes of compilation time.

> My most significant request is that if we do make the test infrastructure more complicated, we do it with a clear plan that will lead to complexity leveling out soon. I hope that as tests get more sophisticated, our testing infrastructure won't need continuous hacks to keep up.

As I see it, the infrastructure suggested in this PR seems generic enough so that we do not have to constantly hack the build system to write more expressive tests. At least, that is my guess.

cmake/Macros.cmake
kimci86: Yes, that seems reasonable.

test/CMakeLists.txt
kimci86: `SFML::System` is the `DEPENDS` argument or `sfml_add_test` and `sfml-test-runner-system` is the `TEST_RUNNER` argument. This is not a list or dependencies.

cmake/Macros.cmake
Bromeon: It's "test runner", which I think is quite an established term.

Not sure what else you could use... "test engine", "test framework" etc. seem also a bit out of place, as we're really running them and not providing the underlying infrastructure.

Maybe just "test" or "tests"?

kimci86: That comes from vague memories of documentation I skimmed through. I am not sure if it is the best word.
After some research, [boost.Test documentation](https://www.boost.org/doc/libs/1_79_0/libs/test/doc/html/boost_test/adv_scenarios/test_module_runner_overview.html) seems to agree. It also gives alternatives "orchestrator" and "driver".
I am open to suggestion.

ChrisThrasher: > Not sure what else you could use... "test engine", "test framework" etc. seem also a bit out of place, as we're really running them and not providing the underlying infrastructure.

I'd name each utils library `sfml-${module}-test-utils` because it's just some test utilities. It's not running anything. It's just some compiled C++ code with some headers for using the code. Each one of those utils library provides a header called `{Module}Utils.hpp` so it's all very consistent this way.

kimci86: > It's not running anything.

That is not true. It comes with a main function from DoctestMain.cpp which does all the magic of command line parsing, test running and console output.
Still `sfml-${module}-test-utils` sounds more like a library than `sfml-test-runner-${module}` so that could be make more sense.

ChrisThrasher: > It comes with a main function from DoctestMain.cpp

I'm aware and I still disagree that it's "running" the code so to speak. Sometimes libraries implement `main()`. Non-header-only test frameworks like Catch v3 and GTest do this. Can we at least name it `sfml-${module}-test-runner` since that reads better than `sfml-test-runner-${module}`?

kimci86: I changed names to `sfml-${module}-test-utils`. I am still not sure what is best so do not hesitate to ask for something else.

cmake/Macros.cmake
kimci86: I tried something else: passing the module name as an argument, as it is enough to derive the library under test which is needed for the coverage-related command.
Regarding adding test libraries for Network and Audio modules, I would prefer not to do it now because libraries without sources must be INTERFACE and not STATIC so its requires some conditionals that I would prefer not to add.

cmake/Macros.cmake
kimci86: Done

test/CMakeLists.txt
```diff
@@ -7,72 +7,163 @@ FetchContent_Declare(doctest
...
-    System/MemoryInputStream.cpp
-    System/Time.cpp
-    System/Vector2.cpp
-    System/Vector3.cpp
-)
-sfml_add_test(test-sfml-system "${SYSTEM_SRC}" SFML::System)
+# add a new static library which provides a main function and testing utilities for SFML types
+# example: sfml_add_test_utils(sfml-window-test-utils
+#                              SOURCES TestUtilities/WindowUtil.hpp TestUtilities/WindowUtil.cpp
+#                              DEPENDS SFML::Window sfml-system-test-utils)
```
ChrisThrasher, line 13: I'm not sure we need these comments anymore since all the users of this function reside in the same file. Same goes for `sfml_add_test`'s description comment.

test/CMakeLists.txt
```diff
@@ -7,72 +7,163 @@ FetchContent_Declare(doctest
...
+    cmake_parse_arguments(THIS "" "" "SOURCES;DEPENDS" ${ARGN})
+    if (NOT "${THIS_UNPARSED_ARGUMENTS}" STREQUAL "")
+        message(FATAL_ERROR "Extra unparsed arguments when calling sfml_add_test: ${THIS_UNPARSED_ARGUMENTS}")
+    endif()
+
+    # set a source group for the source files
+    source_group("" FILES ${THIS_SOURCES})
+
+    # create the target
+    string(TOLOWER test-sfml-${module} target)
```
ChrisThrasher, line 53: What do you think about `sfml_add_test_utils` also using this same naming logic where instead of providing a target name, you just provide the module name?

ChrisThrasher (approved):
I'm still not in love with going back to optionally building tests but at least with this new scheme of having 1 util library per module, we can actually implement that in a reasonable fashion. I made a couple small remarks but not enough to stand in the way of my approval so long as most of the commits get squashed. Thanks, Kimci!

test/CMakeLists.txt
kimci86: I was tempted to do it, but I think it might make the code a little less clear because it would be less obvious where the test utility library target names come from. But I don't feel strongly about it, I could be convinced otherwise.

Consider:
```cmake
sfml_add_test_utils(sfml-system-test-utils ...)
target_include_directories(sfml-system-test-utils PUBLIC TestUtilities)
```
versus
```cmake
sfml_add_test_utils(System ...)
target_include_directories(sfml-system-test-utils PUBLIC TestUtilities)
```

test/CMakeLists.txt
kimci86: I don't see a big difference if the code using those functions is in the same file or not.
A small description doesn't hurt IMHO.

kimci86:
Rebased and cleaned up commit history.

PR #2171, opened 2022-07-27
Test copy and move semantics

ChrisThrasher:
## Description

One of SFML 3's specific goals is to support move semantics. One such obstacle to implementing move semantics is not necessary knowing when the compiler is or is not automatically generating the move special member functions. By using `<type_traits>` we can test to make sure that at a minimum, the special member functions we expect do exist.

These tests will let us catch regressions. If a destructor is added to a class or a member variable is marked `const`, move semantics are implicitly deleted. These assertions will immediately catch that bug and force the developer to reckon with the bug they have introduced.

These tests also serve to document what types are still lacking move semantics. Search the codebase for `static_assert(!std::is_nothrow_move` and you will find the tests for all types that fail to implement a move constructor and/or move assignment operator. Once you add move semantics, the test will fail at which point you remove the `!` and you're good to go.

By testing copy and move semantics on all remaining SFML types, this PR also knocks out a lot of the boilerplate code required to write additional tests. Future developers will have an easier time improving the test suite when the test file already exists for the file they're testing. They can easily start writing code in that file and see it run which will be quite handy and hopefully encourage writing more tests.

Beyond testing for the existence of move semantics, these tests also assert the *lack* of copy semantics for types which are non copyable. Because one cannot write a test using a deleted function, this method is the only way to assert that a type is non copyable.

These tests are easily expanded to test for any number of relevant type traits. For example if a type promises to not be constructible or to be convertible from another type or to be abstract, we can write a test for that ensuring that those promises are never broken.

---

> *Why don't you just write tests that validate that the move assignment operator and move constructor are correct and skip using type traits?*

I already mentioned non copyable types as a justification for testing with type traits in some capacity. As for testing the _implementation_ of move operators, yes that is a good idea and something I'd like to eventually do. However it can be quite difficult (perhaps infeasible for some types) and we're [yak shaving](https://en.wiktionary.org/wiki/yak_shaving) so let's take things one step at a time. Besides, most types have implicitly defined move semantics which we know are going to be correct so I don't think it's worth anyone's time to write those tests.

> *What about putting these assertions in the classes themselves?*

I don't hate this idea but let me explain why keeping the assertions in the tests is a better idea. First, these assertions are tests. Just because we _can_ put them in the core library doesn't mean that makes sense. These assertions are only relevant to us developers. If you're merely using the library you do not care about compiling these static asserts. On top of that, putting these assertions in class templates means adding a new transitive include to all users of SFML. This is not a huge cost but it's certain a non-zero cost that we and all SFML users would be incurring.


ChrisThrasher:
In [this branch](https://github.com/ChrisThrasher/SFML/tree/move_semantics) I've started enabling move semantics for more types. As expected, the tests detect that move semantics now exist.

Waldegger (approved):
I happily help out adding the move semantics

test/Graphics/Shape.cpp
```diff
@@ -99,4 +100,12 @@ TEST_CASE("sf::Shape class - [graphics]")
...
         CHECK(triangleShape.getGlobalBounds().height == Approx(2.f));
     }
+
+    SUBCASE("Copy and move semantics")
+    {
+        static_assert(std::is_copy_constructible<TriangleShape>::value);
+        static_assert(std::is_copy_assignable<TriangleShape>::value);
+        static_assert(std::is_move_constructible<TriangleShape>::value);
+        static_assert(std::is_move_assignable<TriangleShape>::value);
+    }
```
Marioalexsan, line 110: Why is `TriangleShape` being tested and not `Shape`?

test/Network/Socket.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+#include <doctest/doctest.h>
+
+#include <type_traits>
+
+TEST_CASE("sf::Socket class - [network]")
+{
+    SUBCASE("Copy and move semantics")
+    {
+        static_assert(!std::is_constructible<sf::Socket>::value);
```
Marioalexsan, line 11: I'm guessing that only protected / private constructors are tested, and not stuff like abstract classes not being constructible?

test/Graphics/Shape.cpp
ChrisThrasher: Good catch. To be consistent with other non-constructible types, I need to directly test `sf::Shape`.

test/Network/Socket.cpp
ChrisThrasher: It would make sense for all non-constructible classes to be tested but I suppose I just forgot to add that assertions to some types.

test/Audio/AlResource.cpp
```diff
@@ -0,0 +1,17 @@
...
+{
+    SUBCASE("Type traits")
+    {
+        static_assert(!std::is_constructible<sf::AlResource>::value);
+        static_assert(!std::is_copy_constructible<sf::AlResource>::value);
+        static_assert(std::is_copy_assignable<sf::AlResource>::value);
+        static_assert(!std::is_nothrow_move_constructible<sf::AlResource>::value);
+        static_assert(std::is_nothrow_move_assignable<sf::AlResource>::value);
+    }
+}
```
vittorioromeo, line 17:
```suggestion
#include <SFML/Audio/AlResource.hpp>

#include <type_traits>

// Compile-time tests.
static_assert(!std::is_constructible<sf::AlResource>::value);        static_assert(!std::is_copy_constructible<sf::AlResource>::value);        static_assert(std::is_copy_assignable<sf::AlResource>::value);        static_assert(!std::is_nothrow_move_constructible<sf::AlResource>::value);        static_assert(std::is_nothrow_move_assignable<sf::AlResource>::value);
```

There's really not any point in wrapping the `static_assert` calls into a `TEST_CASE` or `SUBCASE`, as they will be tested during compilation. The `SUBCASE` you have here is basically a compile-time no-op.

ChrisThrasher: Fair. That still accomplishes my goal of having a .cpp file where runtime tests may eventually be written but it means we no longer build the doctest header just to not even use it. I'll make that PR-wide refactor soon.

ChrisThrasher:
Refactored the PR to avoid creating a `SUBCASE` that contains no runtime code. This means some entire `TEST_CASE`s go away which has the added benefit of minimizing how often we have to compile doctest.h. We still benefit from the existence of files like test/System/String.cpp so that the boilerplate is already in place for someone to eventually write runtime tests for that type and many others lacking tests.

eXpl0it3r:

test/Audio/OutputSoundFile.cpp
```diff
@@ -0,0 +1,8 @@
+#include <SFML/Audio/OutputSoundFile.hpp>
+
+#include <type_traits>
+
+static_assert(!std::is_copy_constructible<sf::OutputSoundFile>::value);
+static_assert(!std::is_copy_assignable<sf::OutputSoundFile>::value);
+static_assert(!std::is_nothrow_move_constructible<sf::OutputSoundFile>::value);
+static_assert(!std::is_nothrow_move_assignable<sf::OutputSoundFile>::value);
```
vittorioromeo, line 8: Missing newline

ChrisThrasher: Fixed

ChrisThrasher:
Rebased and fixed merge conflicts

eXpl0it3r:

ChrisThrasher:
Rebased on master and fixed merge conflicts.

eXpl0it3r:

test/Graphics/Shape.cpp
```diff
@@ -3,6 +3,13 @@
 #include <doctest/doctest.h>
 
 #include <GraphicsUtil.hpp>
+#include <type_traits>
+
+static_assert(!std::is_constructible<sf::Shape>::value);
+static_assert(!std::is_copy_constructible<sf::Shape>::value);
+static_assert(std::is_copy_assignable<sf::Shape>::value);
+static_assert(!std::is_move_constructible<sf::Shape>::value);
+static_assert(std::is_move_assignable<sf::Shape>::value);
```
vittorioromeo, line 12: Why are these throwing operations? Ditto for all the other derived shape classes.

vittorioromeo (changes requested):
One last concern before I can proceed: I see that in some tests you check for `!is_copy_constructible` and `!is_nothrow_move_constructible`, but you don't check for `!is_move_constructible`.

Doesn't that imply that adding a throwing move constructor to one of these types will not cause the test to fail? 

Is that intended? Or should the test also explicitly check for `!is_move_constructible` (and assignable)?

test/Graphics/Shape.cpp
ChrisThrasher: I'm not certain. At one point I did look into this but forgot what my conclusion was.

ChrisThrasher:
> One last concern before I can proceed: I see that in some tests you check for !is_copy_constructible and !is_nothrow_move_constructible, but you don't check for !is_move_constructible.

I've work been working under the assumption that `is_nothrow_move_constructible` implies `is_move_constructible`. If that is not true then I'll need to reevaluate how I'm testing move semantics.

vittorioromeo (changes requested):
Discussed in person, basically `!is_nothrow_move_XXX` does not imply `!is_move_XXX`, so we'd have to test for both. See [this example on Compiler Explorer](https://gcc.godbolt.org/z/hj5snrM8j).

Furthermore, we should change all the `std::whatever<T>::value` to `std::whatever_v<T>` as it's equivalent but shorter.

ChrisThrasher:
Rebased on master, switched to the `_v` version of these type traits, and added addition type traits for types which are moveable but not nothrow movable. This is a weird distinction but some types with copy semantics are "movable" in the sense that calling `std::move` just makes a copy but they're not nothrow movable which means we still have some work to do if we want the performance gains that moving can give us.

vittorioromeo:

ChrisThrasher:

eXpl0it3r:

eXpl0it3r:
@vittorioromeo are the changes what you expected?

ChrisThrasher:
Fixed conflicts with `master`.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:
Fixed conflicts with `master`.

eXpl0it3r (approved):

PR #1940, opened 2021-12-24
Set Doxygen 'SEARCHENGINE' to 'YES'

vittorioromeo:
Fixes #1841. We'd need to look at a preview first and see how it looks. I've used Doxygen in other projects before, and while the client-side bar is not perfect, it is quite useful at times.

PR #2173, opened 2022-07-31
Update doxygen configuration and enable searchbar

### LINKED ISSUE #1841: Feature request: adding search button to doxygen API site
it will be very helpful if I can search certain APIs, doxygen provides it for free on the client side.

kimci86:
## Description

This is another attempt to fix #1841 inspired by @vittorioromeo 's #1940 but with some more adjustments.

With those minimal changes, it looks not too bad when I generate the HTML documentation locally with doxygen 1.9.4. (See pictures below.)
The result box is oddly placed but there is some javascript involved that I am not familiar with.
If we want a better look, we could benefit from using some modern CSS template in the future instead of maintaining our own.

I am not sure if some fixes would be needed to work on the website too.
@eXpl0it3r What do you think?

------

Capture d’écran de 2022-07-31 22-31-56
Capture d’écran de 2022-07-31 22-33-50


eXpl0it3r:

eXpl0it3r (approved):
Just tested it locally and it seems to work well 👍 

The SFML Website bits, we can look at separately, it uses its own generation file anyways.

eXpl0it3r:
Thanks a lot for fixing this up!

Issue #2079, opened 2022-04-15
Build tooling and package interface improvements for v3


friendlyanon:
Since v3 is around the corner and it's a major version increase, I think it would be time to revamp the build tooling a little bit and follow proper library interface etiquette.

I'm opening this issue is just to spitball some ideas and enumerate wishes regarding what could be better.

One immediate thing I noticed is the presence of binary blobs in source control. This is some icky stuff that can be replaced with package managers easily. Conan and vcpkg are pretty much the only options here. Conan would be nice, because it has a `deploy` generator where the dependency graph can be installed to a staging prefix, but it requires more upfront work with setup, because Conan v2 no longer tries to guess details of the system. vcpkg requires less work upfront, since it just plugs into the CMake configure process using its toolchain script, but I'm not sure what's the proper way is to deploy with it (just copy the `vcpkg_installed` folder from build dir?).   
Either way, neither requires hardcoding details into the build scripts nor coupling, which means that users, developers and package managers (the people) can freely plug whatever they want to into the build. Both allow transparent integration with little to no work required.

There is tons of duplicate code of things that CMake already provides as a built-in, such as the setting of the CRT type on Windows. In fact, what's in the CML file right now if insufficient for providing the debug runtimes. CMake already has an alternative for this, so this code requires no new code, only some documentation in the install/build docs pointing people towards it.

SFML's install interface could also use improvements. The CMake package currently provided is essentially broken. Some details about the broken-ness can be found [here](https://github.com/friendlyanon/cmake-init-sfml-example/blob/master/CMakeLists.txt#L47-L62). As an aside, this is also what prompted #2052, which is just a partial fix, but the commit in that PR also contains the ideal UX, as in that would be all the code ever needed to use SFML in CMake.

The `cmake/Macros.cmake` file could (should) also be sacked, as it does a lot of things to break (sub)system level conventions and it again duplicates a lot of things that CMake provides using built-in variables. I'm very much not a fan of projects overriding user requirements and providing custom magic suffixes like `-static` for static libraries, when the users could make that consciuous decision themselves with no costs to bear for every downstream user. Most of these things should just be documented in the install/build document for users and recommending them to use CMake, since that takes care of all these little differences. Looking at the [Getting Started](https://www.sfml-dev.org/tutorials/2.5/#getting-started) section for example shows no usage with CMake right now, which stinks imo.  
[...]

PR #2174, opened 2022-08-01
Require CMake 3.16 for ObjC and ObjC++ sources

friendlyanon:
Both languages gained native CMake support in 3.16:
https://cmake.org/cmake/help/latest/release/3.16.html#languages

Issue: #2079

## Tasks

Build only test

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android


friendlyanon:
Oops, my master was outdated. Sorry!

src/SFML/Window/CMakeLists.txt
```diff
@@ -272,11 +274,6 @@ target_link_libraries(sfml-window PUBLIC SFML::System)
 # glad sources
 target_include_directories(sfml-window SYSTEM PRIVATE "${PROJECT_SOURCE_DIR}/extlibs/headers/glad/include")
 
-# When static linking on macOS, we need to add this flag for objective C to work
-if ((NOT BUILD_SHARED_LIBS) AND SFML_OS_MACOSX)
-    target_link_libraries(sfml-window PRIVATE -ObjC)
-endif()
-
```
ChrisThrasher, line 279: If I'm understanding correctly, without us explicitly enabling OBJC and OBJCXX as languages (which was not possible when requiring 3.15), this snippet was required but once we express to CMake that we're using OBJC and OBJCXX it can automatically add this flag as required?

friendlyanon: The only reference I found for this flag is just to treat sources as ObjC:

https://clang.llvm.org/docs/CommandGuide/clang.html#cmdoption-objc

> Treat source input files as Objective-C and Object-C++ inputs respectively.

ChrisThrasher: The build on `master` works with this snippet removed but perhaps that's only because I'm using a version of CMake much newer than 3.15. In either case, I support its removal if it's unnecessary. I was just hoping to gain some more insight into how to best handle building Objective-C and Objective-C++ code.

friendlyanon:
OK, quick edits done. I was just missing `enable_language` from the Main module when building for iOS, no need to enable things at the top. I didn't test iOS, because I don't have the iOS SDK on the Mac machine I have access to.

ChrisThrasher (approved):
LGTM. Thanks for finding this! @eXpl0it3r what do you think?

eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:

src/SFML/Window/CMakeLists.txt
```diff
@@ -203,6 +204,7 @@ elseif(SFML_OS_MACOSX)
     )
     source_group("mac" FILES ${PLATFORM_SRC})
 elseif(SFML_OS_IOS)
+    enable_language(OBJCXX)
```
eXpl0it3r, line 207: What's the advantage of splitting things up, compared to https://github.com/SFML/SFML/commit/90af02f318655dd51c3bf640d7c37582f751e2e0 ?

ChrisThrasher: Less maintenance how?

friendlyanon: Wrt the conditionals. It's easier to keep things in sync when the language is enabled at the usage site.

eXpl0it3r:
Thank you for this CMake enhancement! 🙂

src/SFML/Window/CMakeLists.txt
ChrisThrasher: This was originally added [here](https://github.com/SFML/SFML/pull/1485) to fix static linking on macOS. With it gone, I get runtime errors when trying to use static libs. Putting it back seems to fix that problem.

PR #2178, opened 2022-08-06
Assorted CMake fixes

ChrisThrasher:
## Description

Over time I accumulate lots of little fixes that I find while solving other problems. I've collected enough build system improvements to justify a PR now. Each commit explains my rationale and many include extended commit messages further explaining the changes so look at the commit log for more info.


cmake/Config.cmake
```diff
@@ -50,14 +50,6 @@ elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
...
         # don't use the OpenGL ES implementation on Mac OS X
         set(OPENGL_ES 0)
-
-        # detect OS X version. (use '/usr/bin/sw_vers -productVersion' to extract V from '10.V.x'.)
-        EXEC_PROGRAM(/usr/bin/sw_vers ARGS -productVersion OUTPUT_VARIABLE MACOSX_VERSION_RAW)
-        STRING(REGEX REPLACE "10\\.([0-9]+).*" "\\1" MACOSX_VERSION "${MACOSX_VERSION_RAW}")
-        if(${MACOSX_VERSION} LESS 7)
-            message(FATAL_ERROR "Unsupported version of OS X: ${MACOSX_VERSION_RAW}")
-            return()
-        endif()
```
eXpl0it3r, line 60: Is the requirement now given through other incompatibilities?

ChrisThrasher: That's right. Here's the extended commit message where I go into more detail:

> [Remove check for old macOS versions](https://github.com/SFML/SFML/pull/2178/commits/cae9733067814f901efb1f7596eb5fb198e73a2d)  …
There is no need to check for macOS 10.6 or lower. Snow Leopard came
out in 2009. Because it's nearly 13 years old, it's safe to say
nobody will be trying to build SFML 3 on Snow Leopard.
This was added in https://github.com/ChrisThrasher/SFML/commit/de70f691e10b502e09d1683b2d241d559f199948 way back in 2011 when Snow Leopard was
a mere cub.

src/SFML/CMakeLists.txt
```diff
@@ -31,9 +31,6 @@ elseif(SFML_OS_ANDROID)
     list(APPEND CMAKE_LIBRARY_PATH "${PROJECT_SOURCE_DIR}/extlibs/libs-android/${CMAKE_ANDROID_ARCH_ABI}")
 endif()
 
-# define the path of our additional CMake modules
-set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/Modules/")
```
eXpl0it3r, line 35: How does CMake find the scripts in: https://github.com/SFML/SFML/tree/master/cmake/Modules ?

Or can we get rid of those?

ChrisThrasher: That same code also exists [here](https://github.com/SFML/SFML/blob/master/cmake/Macros.cmake#L406).

I suppose it would make more sense to keep this line and remove the one deep in Macros.cmake instead so I'll amend this PR to do that instead. That will keep the code more readable.

vittorioromeo:

cmake/Modules/FindUDev.cmake
```diff
@@ -9,45 +9,45 @@
 # Redistribution and use of this file is allowed according to the terms of the BSD license.
```
eXpl0it3r, line 9: Not sure if we can/should modify this file

ChrisThrasher: My impression was that BSD licenses allowed modification but I'm no software license expert.

eXpl0it3r (approved):

eXpl0it3r:

eXpl0it3r:
Thanks for cleaning that up! 🥳

PR #2181, opened 2022-08-14
Set warnings with `target_compile_options`

ChrisThrasher:
## Description
The use of target_compile_options makes it easy to append warnings to a target. The use of generator expressions also more succinctly handles a few edge cases for compiler bugs and platform-specific oddities. The way to read this file is in four chunks: MSVC warnings, GCC/Clang warnings, GCC-specific warnings, and Clang-specific warnings. I think this reads much easier than the status quo which its use of variables that, in my opinion, are a bit tough to trace out manually. I also realized that this is a good place to be using the `SFML_COMPILER_` variables set in Config.cmake.

### DIFF
--- a/cmake/CompilerWarnings.cmake
+++ b/cmake/CompilerWarnings.cmake
@@ -6,95 +6,77 @@
 function(set_target_warnings target)
     option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" TRUE)
 
-    set(MSVC_WARNINGS
-        /W4 # Baseline reasonable warnings
-        /w14242 # 'identifier': conversion from 'type1' to 'type1', possible loss of data
-        /w14254 # 'operator': conversion from 'type1:field_bits' to 'type2:field_bits', possible loss of data
-        /w14263 # 'function': member function does not override any base class virtual member function
-        /w14265 # 'classname': class has virtual functions, but destructor is not virtual instances of this class may not be destructed correctly
-        /w14287 # 'operator': unsigned/negative constant mismatch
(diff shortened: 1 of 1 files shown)


cmake/CompilerWarnings.cmake
```diff
@@ -6,95 +6,77 @@
...
+            -Wconversion # warn on type conversions that may lose data
+            -Wsign-conversion # warn on sign conversions
+            -Wdouble-promotion # warn if float is implicit promoted to double
+            -Wformat=2 # warn on security issues around functions that format output (ie printf)
+            -Wimplicit-fallthrough # warn when a missing break causes control flow to continue at the next case in a switch statement
+            -Wsuggest-override # warn when 'override' could be used on a member function overriding a virtual function
+            -Wnull-dereference # warn if a null dereference is detected
+            -Wold-style-cast # warn for c-style casts
+            -Wpedantic # warn if non-standard C++ is used
+            $<$<BOOL:${ANDROID}>:-Wno-main> # allow main() to be called
```
eXpl0it3r, line 61: Not your error, but I think this should actually be `SFML_OS_ANDROID`

eXpl0it3r (approved):

eXpl0it3r:

ChrisThrasher:
I checked the CI logs. As far as I can tell all the warnings are still being applied exactly as expected. Clang and GCC warnings are still separated. Only Android is getting that `-Wno-main` warning.

eXpl0it3r:
Thanks for this improvement 🥳

PR #2184, opened 2022-08-21
Use `std::size_t`

ChrisThrasher:
## Description

Found a few places using `size_t` instead of `std::size_t` so I went through a did an exhaustive search for everywhere else using the shorter form so that the whole codebase is now consistent about using the more verbose option.

### DIFF
--- a/src/SFML/Audio/SoundFileReaderOgg.cpp
+++ b/src/SFML/Audio/SoundFileReaderOgg.cpp
@@ -36,7 +36,7 @@
 
 namespace
 {
-size_t read(void* ptr, size_t size, size_t nmemb, void* data)
+std::size_t read(void* ptr, std::size_t size, std::size_t nmemb, void* data)
 {
     auto* stream = static_cast<sf::InputStream*>(data);
     return static_cast<std::size_t>(stream->read(ptr, static_cast<sf::Int64>(size * nmemb)));
--- a/src/SFML/Window/FreeBSD/JoystickImpl.cpp
+++ b/src/SFML/Window/FreeBSD/JoystickImpl.cpp
@@ -228,7 +228,7 @@ bool JoystickImpl::open(unsigned int index)
 
             // Then allocate a buffer for data retrieval
             m_length = hid_report_size(m_desc, hid_input, m_id);
-            m_buffer.resize(static_cast<size_t>(m_length));
+            m_buffer.resize(static_cast<std::size_t>(m_length));
 
             m_state.connected = true;
 
@@ -300,7 +300,7 @@ Joystick::Identification JoystickImpl::getIdentification() const
 ////////////////////////////////////////////////////////////
 JoystickState JoystickImpl::JoystickImpl::update()
 {
-    while (read(m_file, m_buffer.data(), static_cast<size_t>(m_length)) == m_length)
+    while (read(m_file, m_buffer.data(), static_cast<std::size_t>(m_length)) == m_length)
     {
         hid_data_t data = hid_start_parse(m_desc, 1 << hid_input, m_id);
 
--- a/examples/vulkan/Vulkan.cpp
+++ b/examples/vulkan/Vulkan.cpp
@@ -1611,7 +1611,7 @@ public:
     void setupUniformBuffers()
     {
         // Create a uniform buffer for every frame that might be in flight to prevent clobbering
-        for (size_t i = 0; i < swapchainImages.size(); ++i)
+        for (std::size_t i = 0; i < swapchainImages.size(); ++i)
         {
             uniformBuffers.push_back(0);
             uniformBuffersMemory.push_back(0);
--- a/src/SFML/Main/MainAndroid.cpp
+++ b/src/SFML/Main/MainAndroid.cpp
@@ -474,7 +474,7 @@ static void onConfigurationChanged(ANativeActivity* /* activity */)
 
 
 ////////////////////////////////////////////////////////////
-static void* onSaveInstanceState(ANativeActivity* /* activity */, size_t* outLen)
+static void* onSaveInstanceState(ANativeActivity* /* activity */, std::size_t* outLen)
 {
     *outLen = 0;
 
@@ -489,7 +489,7 @@ static void onLowMemory(ANativeActivity* /* activity */)
 
 
 ////////////////////////////////////////////////////////////
-JNIEXPORT void ANativeActivity_onCreate(ANativeActivity* activity, void* savedState, size_t savedStateSize)
+JNIEXPORT void ANativeActivity_onCreate(ANativeActivity* activity, void* savedState, std::size_t savedStateSize)
 {
     // Create an activity states (will keep us in the know, about events we care)
     auto* states = new sf::priv::ActivityStates;
--- a/src/SFML/Main/SFMLActivity.cpp
+++ b/src/SFML/Main/SFMLActivity.cpp
@@ -38,7 +38,7 @@
 
 namespace
 {
-using activityOnCreatePointer = void (*)(ANativeActivity*, void*, size_t);
+using activityOnCreatePointer = void (*)(ANativeActivity*, void*, std::size_t);
 }
 
 const char* getLibraryName(JNIEnv* lJNIEnv, jobject& objectActivityInfo)
@@ -71,8 +71,8 @@ const char* getLibraryName(JNIEnv* lJNIEnv, jobject& objectActivityInfo)
     }
 
     // Convert the application name to a C++ string and return it
-    const size_t applicationNameLength = static_cast<size_t>(lJNIEnv->GetStringUTFLength(valueString));
-    const char*  applicationName       = lJNIEnv->GetStringUTFChars(valueString, nullptr);
+    const std::size_t applicationNameLength = static_cast<std::size_t>(lJNIEnv->GetStringUTFLength(valueString));
+    const char*       applicationName       = lJNIEnv->GetStringUTFChars(valueString, nullptr);
 
     if (applicationNameLength >= 256)
     {
@@ -80,7 +80,7 @@ const char* getLibraryName(JNIEnv* lJNIEnv, jobject& objectActivityInfo)
         exit(1);
     }
 
-    strncpy(name, applicationName, static_cast<size_t>(applicationNameLength));
+    strncpy(name, applicationName, static_cast<std::size_t>(applicationNameLength));
     name[applicationNameLength] = '\0';
     lJNIEnv->ReleaseStringUTFChars(valueString, applicationName);
 
@@ -132,7 +132,7 @@ void* loadLibrary(const char* libraryName, JNIEnv* lJNIEnv, jobject& ObjectActiv
     return handle;
 }
 
-void ANativeActivity_onCreate(ANativeActivity* activity, void* savedState, size_t savedStateSize)
+void ANativeActivity_onCreate(ANativeActivity* activity, void* savedState, std::size_t savedStateSize)
 {
     // Before we can load a library, we need to find out its location. As
     // we're powerless here in C/C++, we need the JNI interface to communicate
--- a/src/SFML/Network/Unix/SocketImpl.hpp
+++ b/src/SFML/Network/Unix/SocketImpl.hpp
@@ -55,7 +55,7 @@ public:
     // Types
     ////////////////////////////////////////////////////////////
     using AddrLength = socklen_t;
-    using Size       = size_t;
+    using Size       = std::size_t;
 
     ////////////////////////////////////////////////////////////
     /// \brief Create an internal sockaddr_in address
--- a/src/SFML/System/Android/Activity.hpp
+++ b/src/SFML/System/Android/Activity.hpp
@@ -67,8 +67,8 @@ struct ActivityStates
     EGLDisplay  display;
     EglContext* context;
 
-    void*  savedState;
-    size_t savedStateSize;
+    void*       savedState;
+    std::size_t savedStateSize;
 
     std::recursive_mutex mutex;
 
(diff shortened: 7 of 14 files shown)


src/SFML/Audio/SoundFileReaderOgg.cpp
```diff
@@ -36,7 +36,7 @@
 
 namespace
 {
-size_t read(void* ptr, size_t size, size_t nmemb, void* data)
+std::size_t read(void* ptr, std::size_t size, std::size_t nmemb, void* data)
```
kimci86, line 39: This function is used to fill an `ov_callbacks` structure whose read member is of type `size_t (*)(void*, size_t, size_t, void*)`.
See: https://www.xiph.org/vorbis/doc/vorbisfile/ov_callbacks.html
Wouldn't it be cleaner to keep using the exact same type?
When using C interfaces, I am not sure it makes sense to use `std::size_t`. What do you think?

eXpl0it3r:

src/SFML/Audio/SoundFileReaderOgg.cpp
ChrisThrasher: I don't really know what I prefer but I guess I'll remove the `std::` prefixes here.

src/SFML/Window/FreeBSD/JoystickImpl.cpp
```diff
@@ -300,7 +300,7 @@ Joystick::Identification JoystickImpl::getIdentification() const
 ////////////////////////////////////////////////////////////
 JoystickState JoystickImpl::JoystickImpl::update()
 {
-    while (read(m_file, m_buffer.data(), static_cast<size_t>(m_length)) == m_length)
+    while (read(m_file, m_buffer.data(), static_cast<std::size_t>(m_length)) == m_length)
```
kimci86, line 303: and this change on this line is to be reverted too.

kimci86 (commented):
Having a closer look, many instances of `size_t` (without `std::`) are justified because we are using C interfaces.
If we agree keeping `size_t` (without adding `std::`) when using C interfaces is what we want, files to revert in this PR are:
- `src/SFML/Main/MainAndroid.cpp`
- `src/SFML/Main/SFMLActivity.cpp`
- `src/SFML/Network/Unix/SocketImpl.hpp`
- `src/SFML/System/Android/Activity.hpp`
- `src/SFML/System/Android/ResourceStream.cpp`
- `src/SFML/Window/Android/WindowImplAndroid.cpp`

src/SFML/Window/FreeBSD/JoystickImpl.cpp
ChrisThrasher: How come? This is C++ code. It calls a C function but it's weird to use exceptional rules here. The cognitive overhead to maintain these rules is not trivial.

ChrisThrasher:
> Having a closer look, many instances of size_t (without std::) are justified because we are using C interfaces.

Calling `.data()` to produce an argument to a C function is using C++ code to produce something C can understand (a pointer). Casting to `std::size_t` is using C++ code to produce something that C can understand (an `unsigned long long`). I don't see why we need to make these carveouts where C interop is happening since these are not `extern "C"` blocks. It's all still C++ code.

kimci86:
I find it cleaner to pass arguments of whatever type the used interface expects.
In the case of `size_t` and `std::size_t`, both types are equivalent so technically there is no problem, but thinking about this equivalence adds a little step (an obvious one I admit) when reading the code to figure out if it makes sense.
But if having `std::size_t` all the time helps maintainability, I can live with it.

eXpl0it3r:

Bromeon:
It depends on whether you see the namespace as "an abstraction of the type", just like `int32_t` can be an abstraction for `int`. In that case, one could argue that the abstract versions should be used for maximum compatibility and indication of semantics.

I don't see it like this however -- `std::size_t` and `size_t` are identical and always will be. The only benefit of differentiating them is maybe documentation purposes, although I don't really see the need to say "this is the qualified version and this is not, but both are the same". It's a bit as if you would sometimes use `std::vector<T>` and sometimes `std::vector<T, std::allocator<T>>` -- they are identical by definition, so using one version consistently leads to easier rules.

ChrisThrasher:
> It's a bit as if you would sometimes use `std::vector<T>` and sometimes `std::vector<T, std::allocator<T>>`

For this same reason I'd prefer we drop the superfluous `std::` prefix. `std::size_t` no more clearly expresses intent than `size_t` yet adds 5 characters in over 300 places. Removing them shortens line length which contributes to clearer, simpler formatting. 

> so using one version consistently leads to easier rules.

But of course no matter what we choose, I'd prefer we be absolute with what we pick and not make exemptions. I most prefer we exclusively use `size_t` but my second favorite option is to exclusively use `std::size_t` with a mix of the two in a distant third place.

These arguments apply to all C++ types which are defined as nothing more than aliases for C types. 

---

Curious what it would look like if we removed all those `std::` prefixes from `size_t`? Check it out [here](https://github.com/ChrisThrasher/SFML/commit/c4afce68272891c0a443e04112c3a37a44341448).

Bromeon:
> For this same reason I'd prefer we drop the superfluous `std::` prefix. `std::size_t` no more clearly expresses intent than `size_t` yet adds 5 characters in over 300 places. Removing them shortens line length which contributes to clearer, simpler formatting.

I was an almost religious "we must write `std::` everywhere" advocate for a long time, until I realized that C++ is the only language that thoroughly requires that. Every other modern language has a module and import system, and you don't risk littering some "global namespace" from a file included elsewhere. That is, you _can_ qualify (or import-alias) to address ambiguities, but don't _have to_.

Also, Rust taught me that shorter keywords for often-used symbols makes everything _more_ readable, not less. Consider:
```cpp
// C++
std::vector<std::uint64_t>
std::unordered_map<std::size_t, std::shared_ptr<std::string>>
std::numeric_limits<std::uint8_t>::max()
static_cast<unsigned short>(expr)
```
```rs
// Rust
Vec<u64>
HashMap<usize, Option<String>>
u8::MAX
expr as u16
// btw, HashMap is the only symbol here that needs an import at all
```
So I'm a big fan of shortening ubiquitous identifiers and would support this change in cases, where:
* such identifiers are already present and don't need `using namespace` (that's obviously a no-no in headers)
* and don't lead to 2 separate code styles (qualified in header and unqualified in impl).

Looks like `size_t` would qualify here. Minor points where I'm not sure though:
* In theory, the C version would require `#include <header.h>` instead of `#include <cheader>`, no? In practice probably everything is included transitively anyway, so it's more an academic issue.
* Do linters complain when using the C types?

ChrisThrasher:
> In theory, the C version would require #include <header.h> instead of #include <cheader>, no? In practice probably everything is included transitively anyway, so it's more an academic issue.

https://eel.is/c++draft/cstddef.syn#header:%3ccstddef%3e

I can't find language in the standard which *guarantees* this but I think it's a safe assumption in either case. For this PR I was digging into libc++ and libstdc++'s `<cstddef>` and both transitively included definitions of `size_t` from the C standard library.

> Do linters complain when using the C types?

Not that I've ever come across. `size_t` is as much of a C type as `int` is so I can't think of any reason it would ever give us issues.

kimci86:
To clarify, I was only pointing out what looked like small inconsistencies, but I don't feel strongly about it. Feel free to move on with whatever convention is preferred.

ChrisThrasher:
Rebased on master and reverted back to using `std::size_t` exclusively.

Bromeon:
> Rebased on master and reverted back to using `std::size_t` exclusively.

Why that now? To my understanding the discussion was more in favor of `size_t`, am I missing something? 🙂

ChrisThrasher:
> To my understanding the discussion was more in favor of size_t, am I missing something?

May I? At a minimum I wanted to make this PR internally consistent but didn't think I was given the go-ahead to use `size_t`. I'd still prefer to remove that `std::` namespace if the SFML team are cool with that.

Bromeon:
I just realized that with all my devil-advocating, my point of view may have come across as quite ambiguous 😬 
This expresses what I meant:

> So I'm a big fan of shortening ubiquitous identifiers and would support this change in cases, where:
>
>  * such identifiers are already present and don't need `using namespace` (that's obviously a no-no in headers)
>  * and don't lead to 2 separate code styles (qualified in header and unqualified in impl).
> 
> Looks like `size_t` would qualify here.

TLDR: I would support `size_t`.
But let's see what other team members think.

ChrisThrasher:
I pushed another commit that uses `size_t` so you can easily see both in this PR. Once the team makes a decision I'll edit the history to remove whatever changes are not needed.

vittorioromeo:
I am strongly in favour of explicitly using `std::` for this PR and #2192 (if there's consensus on that one) for a few reasons:

1. `size_t` is C, `std::size_t` is C++. SFML is a C++ library, so we should be consistent with the language we picked.

2. Using `size_t` requires including a C header such as `<stddef.h>`. There's no guarantee that `<cstddef>` will define a global `size_t`. From Jonathan Wakely (emphasis mine):

    > If you include `<cstddef>` you are guaranteed to get `std::size_t` and you **might** also get `::size_t` **(but it's not guaranteed)**. If you include `<stddef.h>` you're guaranteed to get `::size_t ` and you **might** also get std::size_t **(but it's not guaranteed)**. It was different in C++03 but that was effectively unimplementable and fixed as a defect
    
3. When reading code, I value namespaces as a readability aid. Seeing `std::` immediately puts my brain into the context of *"something that comes from the Standard"*. Seeing a symbol without a namespace requires me to do a bit of extra name lookup in my head :)

Regardless of what we choose, we need to make sure that the proper headers are included everywhere we use `std::size_t` or `std::intXX_t`.

kimci86:
@vittorioromeo What is your opinion about what to do when we use C interfaces? When we use a C function taking `size_t`, should we pass a `std::size_t` to be consistent with our C++ code base or `size_t` to be consistent with the C interface?

Bromeon:
@vittorioromeo These are all good points you're making. I just wonder if we shouldn't favor ergonomics a bit more in SFML, because a lot of code in SFML is much more ugly than it could be.

I see it like this: `int`, `short` etc. were a mistake in C, which is the reason why `int16_t` & Co. exist in the first place. Those have been in the standard for a long time and every C or C++ programmer knows these types. The _only_ reason why they are not keywords is the C++ commitee's paranoid stance on compatibility, aka. "what if some 1993 code contains those exact names" or "what if some archaic system doesn't have this type". Which I guess is fair, but on the other hand it's also 2022 now, and we're specifically talking about SFML.

As such, I don't see what the `std::` prefix really adds -- it doesn't help readability, and it doesn't make it more clear that those are from the standard library.

Regarding standard compliance, I see two possible approaches:
* We include `<stddef.h>` in `Config.hpp`, to _guarantee_ these identifiers are available.
* We keep working with `<cstddef>` and in practice it will still work. If it doesn't, we have CI for the platforms we support.

ChrisThrasher:
> Using size_t requires including a C header such as <stddef.h>. There's no guarantee that <cstddef> will define a global size_t.

@vittorioromeo What implementations these days are C++17 compliant yet don't include `size_t` in `<cstddef>`? If any exist then it shouldn't be too hard to find them.

vittorioromeo:
> @vittorioromeo What is your opinion about what to do when we use C interfaces? When we use a C function taking `size_t`, should we pass a `std::size_t` to be consistent with our C++ code base or `size_t` to be consistent with the C interface?

Can you provide a concrete example? `std::size_t` is guaranteed to be the same as a `size_t`, so it is fine to keep using `std::size_t`.



> a lot of code in SFML is much more ugly than it could be

This is subjective. To me, `size_t` is honestly uglier than `std::size_t`. It reeks of C :D



> We include <stddef.h> in Config.hpp, to guarantee these identifiers are available.

See https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2340r1.html, recently approved in plenary:

cheaders

I stand by my original opinion and by what is mentioned above in P2340.

ChrisThrasher:
I hope somebody can find an implementation that doesn't provide `size_t` in `<cstddef>`. I don't believe any such implementation exists but if it does that would be good to know about since that's ultimately what we're worried about.

With 1 maintainer wanting `size_t` and the other `std::size_t`, where do I go from here?

eXpl0it3r:

eXpl0it3r:
While originally on the fence about this, I had some discussions on the topic and come to the same conclusion as @vittorioromeo 

- We can all agree that `std::size_t` and `size_t` are equivalent as defined by the standard, as such there's not *technical* argumentation to be had
- Drive a hardline on this, takes away and brain gymnastics to keep track of when to use what, as such we should either choose with `std::` or without `std::`
- As not all types can be represented without `std::` and as it's the "idiomatic C++ style", I vote for using `std::` everywhere, even when calling C functions.

ChrisThrasher:
With 2 votes to 1 I think we have a decision. I'll remove the `size_t` stuff from the commit history so the PR is ready to merge.

eXpl0it3r:

ChrisThrasher:
Rebased and fixed merge conflicts.

eXpl0it3r:

JonnyPtn:
I think it's better without the namespace for what it's worth. No reasons other than those that have already been mentioned

Bromeon:
So, from all participants here (team members _italics_), from comments and 👍 reactions:

* `std::size_t` -- *@eXpl0it3r*, *@vittorioromeo*, @Ruckamongus, @kimci86 
* `size_t`: *@Bromeon*, @ChrisThrasher, @JonnyPtn, @therocode

It looks like number-wise it's a tie, but more team members are in favor of `std::size_t`.

What is our consensus-finding process?
Not just for this particular issue, but also for the future.

ChrisThrasher:
> tl;dr: let's use std::size_t and merge :P

Sounds good to me!

eXpl0it3r (approved):

eXpl0it3r:
Phew 😃  

Thanks for keeping the discussion active and making us all think a lot 😄 
At least I asked my questions I never asked before ^^

PR #2192, opened 2022-09-03
Start removing custom fixed width integer aliases

ChrisThrasher:
## Description

Because the fixed width `sf::` integers are merely aliases for C++11 types, there is no need to have them. They're not even particularly shorter than the standard equivalents and are actually longer if you include the `sf::` namespace as is the case with all 3rd party code. This PR only removes the 8-bit integer aliases for the purposes of keeping the scope of the PR manageable and because I want more feedback before I take the effort to remove the other 6 aliases. If SFML were written today, surely these aliases never would have been added.


ChrisThrasher:
Glad to see this PR got accepted. Is this the right amount of changes to include in this PR or would the team rather see this change in its entirety in a single PR? Currently we're on track for 4 total PRs to complete this.

eXpl0it3r:
I at least think it's worth doing. We only defined these SFML type, because C++ didn't provide us fixed length ones in C++03.

I'm fine with multiple PRs, makes it easier to review.

Bromeon (approved):
Looks good to me! I'm OK with the PR size, too.

Only small subtlety we should have a look a is when either `uint8_t` or `int8_t` is the same type as `char` (not `unsigned/signed char`). In very rare edge cases we might rely on a certain type, although no such case comes to my mind in SFML (maybe pointers or so).

ChrisThrasher:
Btw as per the majority decision made in #2184, I will add `std::` to all these fixed width integer types. I'd appreciate it if someone could help me write a regex that finds all un-namespaced instances of `T` and adds `std::`. I struggled with this in the `size_t` PR since merely searching for `size_t` yields many other types which you do not want to add `std::` to.

Bromeon:
Probably easier than finding a regex (negatives are always a bit of a pain) would be two replacements:
```
size_t -> std::size_t
std::std::size_t -> std::size_t
```
For genericity over `size_t`, `uint8_t`, `ptrdiff_t` etc, you could use `[a-z]+[0-9]*_t`.

Bromeon:
> I love that trick! I've used it before successfully. Sadly I think the issue was that in that first transition, you end up picking up irrelevant types like `ssize_t` or `khronos_ssize_t`.

This can typically be countered by the "full words only" filter in IDEs. You can also explicitly prefix the regex by `(^|[^a-zA-Z0-9])`, meaning "anything that either begins the line or is preceded by a non-alphanumeric character" (untested).

Btw, we might want to await final resolution of https://github.com/SFML/SFML/pull/2184#issuecomment-1238261692. Some team members (medium active) like binary1248 have not voiced their opinion yet. Although, in the worst case reverting back wouldn't be difficult anymore.

ChrisThrasher:
Amended the latest commit just to get CI to re-run. It was full of code coverage failures a few hours ago.

ChrisThrasher:
https://github.com/ChrisThrasher/SFML/commits/more_fixed_width_ints

I've got the removal of the 16 bit integer aliases queued up for after this PR.

eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:
!next 😄 

Thanks a lot for this!

ChrisThrasher:
See #2196 for more!

PR #2193, opened 2022-09-05
Edit gitignore to ignore clion default output directory

LDprg:

----

## Description

Just a small fix to prevent git to add Clion standard ouput paths.

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [x] Tested on macOS
* [x] Tested on iOS
* [x] Tested on Android

## How to test this PR?

Obviously this change will work on every platform.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@
 
 # Build folders
 build/
+cmake-build-*/
 
 # Cache/IDE folders
 .cache/


ChrisThrasher:
Just use a binary directory named `build` to avoid needing to do this. Otherwise, set up a global .gitignore so that we don't have to keep adding stuff to this file.

https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

Bromeon:
> Just use a binary directory named `build` to avoid needing to do this. Otherwise, set up a global .gitignore so that we don't have to keep adding stuff to this file. Or you can also make a binary directory outside of the repo. You've got a few options.

So we explicitly gitignore the project folder `.idea`, but then don't go the full mile and instead say "you need to take care of the build folder yourself"? That doesn't make sense to me.

Of course people can set up things themselves, but the point of this `.gitignore` file is precisely that they don't have to.

CLion is a popular IDE, and @LDprg already put in the effort to open a pull request. I don't see a reason to _not_ merge this.

ChrisThrasher:
> So we explicitly gitignore the project folder .idea, but then don't go the full mile and instead say "you need to take care of the build folder yourself"? That doesn't make sense to me.

You're confusing my ideas (keep the .gitignore minimal) with the SFML team's idea (make the .gitignore file big). In past discussions I've disagreed with adding stuff like .idea and other IDE-specific directories to the .gitignore when that file need not include anything more than the `build` directory.

Bromeon:
@ChrisThrasher Ah, I see.

My opinion on such things is: it's a very small change for us (here even more since a PR is readily mergable), with near-zero downsides, but it saves _many_* users from spending extra time on configuration or having ignored files show up in Git.

I'm not sure if it's our job to educate people on using global .gitignores. Which btw also comes with drawbacks, as certain projects want their IDE config at least partially versioned (notably single developers or a team on the same IDE).

\_________
\* hard to say how many. Purely utilitarian, even 1-2 might pay off already time-wise.

eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:
Thanks for this enhancement! 🙂

Issue #2190, opened 2022-09-03
Replace SFML Vectors with GLM or Eigen


LDprg:
SFML Vectors have been around for plenty of years and I think with the desicion to move c++17 in SFML 3 we need to get away from them.  I would suggest to use Vectors as abstraction layer to some real math Libary like GLM or Eigen. This step would speed up SFML a whole lot, especially for collision detection and other math heavy calculations, but it wouldn't need that much changes in the Libary itself.

ChrisThrasher:
> This step would speed up SFML a whole lot, especially for collision detection and other math heavy calculations

What makes you say this?

LDprg:
I did some very basic Testing.

ChrisThrasher:
Care to share those results?

LDprg:
Here is my Testing code:
`#define TEST_SFML
//#define TEST_EIGEN

#ifdef TEST_SFML
#include <SFML/Graphics.hpp>
#endif

#ifdef TEST_EIGEN
#include <Eigen/Core>
#endif

#include <iostream>
#include <chrono>
#include <cmath>

#define v11 1.1
#define v12 2.5

#define v21 3.3
#define v22 4.2

#define iter 1000000000

int main() {
    auto start = std::chrono::high_resolution_clock::now();

    for(int i = 0; i < iter; i++) {
#ifdef TEST_SFML
        sf::Vector2f sv1(v11 * i, v12 * i);
        sf::Vector2f sv2(v21 * i, v22 * i);

        sf::Vector2f sDiff = sv1 - sv2;

        [[maybe_unused]] float sDiffLen = std::sqrt(sDiff.x * sDiff.x + sDiff.y * sDiff.y);
#endif
#ifdef TEST_EIGEN
        Eigen::Vector2f ev1(v11 * i, v12 * i);
        Eigen::Vector2f ev2(v21 * i, v22 * i);

        Eigen::Vector2f eDiff = ev1 - ev2;

        [[maybe_unused]] float eDiffLen = eDiff.norm();
#endif
    }

    auto finish = std::chrono::high_resolution_clock::now();

    std::cout
#ifdef TEST_SFML
    << "SFML\tsf::Vector2f\t"
#endif
#ifdef TEST_EIGEN
    << "Eigen\tEigen::Vector2f\t"
#endif
    << std::chrono::duration_cast<std::chrono::nanoseconds>(finish-start).count() << " ns" <<std::endl;
    return 0;
}`

[...]

PR #2195, opened 2022-09-06
Change Vector2 and Vector3 length function

LDprg:
* [x] Has this change been discussed on [Github](https://github.com/SFML/SFML/issues/2190) before?

----

## Description

Replacing hypot because of bad performance.

Perfomance with hypot:
grafik

Performance with a normal sqrt instead
grafik

GLM and EIgen only for comparison reasons

As you can see using a normal sqrt speeds up the length function by a factor of ~10.

This PR is related to the issue #2190

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Testet with the benchmark code. 

Test and Benchmark code:
[main.zip](https://github.com/SFML/SFML/files/9494334/main.zip)!

Obviously the change is platform independent.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@
 
 # Build folders
 build/
+cmake-build-*/
 
 # Cache/IDE folders
 .cache/
--- a/src/SFML/System/Vector2.cpp
+++ b/src/SFML/System/Vector2.cpp
@@ -106,8 +106,9 @@ template <typename T>
 T Vector2<T>::length() const
 {
     static_assert(std::is_floating_point_v<T>, "Vector2::length() is only supported for floating point types");
-    
-    return std::sqrt(x*x + y*y);
+
+    // don't use std::hypot because of slow performance
+    return std::sqrt(x * x + y * y);
 }
 
 } // namespace sf
--- a/src/SFML/System/Vector2.cpp
+++ b/src/SFML/System/Vector2.cpp
@@ -106,8 +106,8 @@ template <typename T>
 T Vector2<T>::length() const
 {
     static_assert(std::is_floating_point_v<T>, "Vector2::length() is only supported for floating point types");
-
-    return std::hypot(x, y);
+    
+    return std::sqrt(x*x + y*y);
 }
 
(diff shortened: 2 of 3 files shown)


Bromeon:
This is a crazy result!

An explanation can be found on StackOverflow: https://stackoverflow.com/questions/3764978/why-hypot-function-is-so-slow

Feedback for the code:
* formatting currently doesn't pass (there should be space between operands, and no space on empty lines)
* maybe add a comment that `std::hypot` is too slow, just to avoid regressions in the future

eXpl0it3r:

LDhtl:
From my side this is finsihed couldn't find any other uses of hypot, which should be replaced.

Bromeon (approved):

vittorioromeo (approved):
LGTM, thank you for the contributions and the benchmarks. @LDprg: the next step to speed this up even further is replace `std::sqrt` with the `fastInverseSqrt` that other libraries use -- I would be happy to see a PR for that.

Bromeon:
@LDhtl and @LDprg the same person? 🙂 

Thanks a lot for spotting and improving this performance bottleneck!

Marioalexsan:
> LGTM, thank you for the contributions and the benchmarks. @LDprg: the next step to speed this up even further is replace `std::sqrt` with the `fastInverseSqrt` that other libraries use -- I would be happy to see a PR for that.

Fast inverse square root does not have perfect accuracy unlike `std::sqrt`, so I'm not sure if this is a straightforward replacement.

LDprg:
Well there are different ways of doing a fast inverse sqrt, with different accuracy and performance. But it seems like they are all working only with float type (so no double precision possible). Maybe it's the best to add a separate function like fastlength().

vittorioromeo:
Some more interesting reading here:
https://stackoverflow.com/questions/71608126/is-it-still-worth-using-the-quake-fast-inverse-square-root-algorithm-nowadays-on

PR #2196, opened 2022-09-09
Continue removing custom fixed width integer aliases

ChrisThrasher:
## Description

Building on what I started in #2192.


eXpl0it3r:

Bromeon (approved):
Some stuff like
```cpp
const auto uint16Result = static_cast<std::uint16_t>(
    static_cast<std::uint16_t>(lhs) * static_cast<std::uint16_t>(rhs));
```
is now excessively verbose. 
The equivalent C code (not saying we should use this, but interesting comparison):
```c
const uint16_t uint16Result = (uint16_t)((uint16_t)lhs * (uint16_t)rhs);
```

I'm personally still not a fan that we cling to "the true C++ way" with visible readability deterioration, but if the majority thinks this is the way to go, let's do it 🙂

But so that I'm not only complaining, maybe I can try out a few things to keep things C++-y _and_ more readable, once all these `std::[u]int*_t` refactorings are done.

---

The PR itself looks good to me!

ChrisThrasher:
Functional casts (which are only a C++ feature) would drastically cut down on verbosity which still having identical behavior. No need to resort to C-style casts.

vittorioromeo (changes requested):
Two things:

1. I am generally happy with this change and removing `sf::*Int*` typedefs, but is there any drawback in doing so? I cannot think of any off the top of my head, but just wondering if there's something we're missing.

2. [`<cstdint>`](https://en.cppreference.com/w/cpp/header/cstdint) needs to be explicitly included everywhere these fixed aliases are used, otherwise we're relying on transitive inclusions.

ChrisThrasher:
> Two things:
> 
> 
> 
> 1. I am generally happy with this change and removing `sf::*Int*` typedefs, but is there any drawback in doing so? I cannot think of any off the top of my head, but just wondering if there's something we're missing.
> 
> 
> 
> 2. [`<cstdint>`](https://en.cppreference.com/w/cpp/header/cstdint) needs to be explicitly included everywhere these fixed aliases are used, otherwise we're relying on transitive inclusions.

The main drawback is verbosity since we're insisting on using std:: everywhere.

How about I wait until the last PR in this series to remove <cstdint> from Config.hpp and add that to all other files which need it? We already heavily rely on transitive inclusions and I've never once seen it be a problem.

Bromeon:
> We already heavily rely on transitive inclusions and I've never once seen it be a problem.

Also, don't we have CIs for various platforms to precisely test things like this? We already support more configurations than many C++ projects, and if someone has yet another setup, they can help us configure it with a PR.

If it's easy to write code to follow best-practices, why not. But lately we lose a lot of time with theoretical issues that have absolutely zero impact on any user (and don't get me wrong, I'm fully part of that)... 😉

ChrisThrasher:
> Also, don't we have CIs for various platforms to precisely test things like this?

Exactly. As long as there are no CI failures and as long as the contributors aren't annoyed by transitive inclusions, I see no problem that needs to be fixed. We're just creating work for ourselves which will present an ongoing maintenance burden. In my time working on SFML, I've never once seen a complaint about our use of transitive inclusions nor have I personally thought twice about it. Everything works and I'm happy that I don't have to add `#include` statements all over the place.

> If it's easy to write code to follow best-practices, why not.

I do not think this constitutes best practice.

> But lately we lose a lot of time with theoretical issues that have absolutely zero impact on any user (and don't get me wrong, I'm fully part of that)... 😉

Establishing precedent is good and this is yet another precedent which (as far as I'm aware) has not be established.

---

I've used include-what-you-use in the past but came away annoyed at its recommendation to include non-standard headers. It provided some useful feedback (which I actually used to _remove_ unnecessary `#include` statements) but I had to careful cherry-pick advice to make use of it.

If someone can successfully integrate IWYU up to the standards of our clang-format integration, I'm happy to revisit the idea of potentially following the include-what-you-use philosophy. For today given our lack of tooling support, I'd much prefer we continue to rely on transitive inclusions and I'm confident they won't present a problem in the future.

ChrisThrasher:
> Some stuff like
> 
> ```c++
> const auto uint16Result = static_cast<std::uint16_t>(
>     static_cast<std::uint16_t>(lhs) * static_cast<std::uint16_t>(rhs));
> ```
> 
> is now excessively verbose.

@Bromeon Drastically less verbose options exist without needing any C features. I hope we give this more consideration since it stands to let us remove a lot of code while perfectly maintaining existing behavior. It's also an easy find-replace command so the writing of a PR won't take much time either. Topic for another PR but think about it.

```c++
const auto uint16Result = std::uint16_t(std::uint16_t(lhs) * std::uint16_t(rhs));
```

Bromeon:
> @Bromeon Drastically less verbose options exist without needing any C features. I hope we give this more consideration since it stands to let us remove a lot of code while perfectly maintaining existing behavior. It's also an easy find-replace command so the writing of a PR won't take much time either. Topic for another PR but think about it.
> 
> ```c++
> const auto uint16Result = std::uint16_t(std::uint16_t(lhs) * std::uint16_t(rhs));
> ```

Function-style casts in particular suffer from the same problem as C casts -- namely not that they are C (even if that alone were ground for religious aversion to some C++ developers 😛), but rather that they are less type-safe than `static_cast`. They can be accidentally used to do pointer-casts or const-casts, for example.

A better solution in my opinion would be named functions:
```cpp
const auto uint16Result = toUint16(toUint16(lhs) * toUint16(rhs));
```
People will still argue that this confuses everyone because it takes 10 seconds to look at the definition once... but we can't have everything.

---

For me it's somewhat interesting that we always argue about issues that are solved problems in any modern language.
Examples:
* no language needs `std::` before every identifier
* no language has such an excessive cast syntax
* no language has variable-sized basic types and requires the standard library just to express something as basic as "16-bit int"
* no language has 4 different initialization syntaxes


Reality is that C++ allows many different styles, and even something as simple as declaring a variable needs trade-offs. I do think however we can learn from other languages how to do things _more pragmatically_. For me the best way is not necessarily "the standard way" (if that exists) but one that gets the job done equally well but is also readable. I think people underestimate how important readability is -- not just to understand what the code is doing, but it can also pose an entry barrier for new contributors, for example. Readability is of course subjective, and we need to think who our target audience is -- is it people with lots of years of C++ experience? Or people who have used other languages and are relatively new?

Anyway, this just some brainstorming 🙂 I think in the end we should use one style that most of us are comforable with and go with it.

ChrisThrasher:
> Function-style casts in particular suffer from the same problem as C casts

I'm fully aware and recommend functional casts with all of that in mind. The nature of their syntax makes them meaningfully harder to abuse than C-style casts. You can't write code like `float*(&i)` whereas you can easily write `(float*)&i`. When we're annoyed by how verbose `static_cast` is, let's consider an alternative already baked into the language. It has big potential to improve readability.

---

> we need to think who our target audience is -- is it people with lots of years of C++ experience?

For PRs like this yes it is. We have no obligation to make the implementation of SFML itself be understandable to novices and we should not be afraid to use sophisticated solutions where appropriate. For the API itself, the audience includes those with exceedingly little C++ experience.

vittorioromeo (approved):
LGTM, discussed the lambda examples in person and they might be handled in a separate PR. The transitive includes seem to not be a problem right now, we can deal with this in a separate PR.

PR #2198, opened 2022-09-12
Time out test for public IP address

ChrisThrasher:
## Description

On poor network connections, the call to getPublicAddress may hang indefinitely. If it is to succeed, let's require that it succeed without a reasonable time frame to put an upper limit on how long tests take to run.

While I'm at it, I'm also letting this test pass even if nullopt is returned because that behavior is part of the contract of this API. Otherwise this test is requiring that a stable network connection is present meaning your tests would fail on an airplane for no good reason.

### DIFF
--- a/test/Network/IpAddress.cpp
+++ b/test/Network/IpAddress.cpp
@@ -72,9 +72,11 @@ TEST_CASE("sf::IpAddress class - [network]")
         SUBCASE("getPublicAddress")
         {
             const std::optional<sf::IpAddress> ipAddress = sf::IpAddress::getPublicAddress(sf::seconds(1));
-            REQUIRE(ipAddress.has_value());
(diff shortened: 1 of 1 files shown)


vittorioromeo:

vittorioromeo (approved):
LGTM

Bromeon:
This _could_ also fail for unrelated reasons and we would simply keep ignoring the error. But I don't know a way to reliably test this without blowing up the complexity (like mimicking a server).

Effectively this means it becomes a best-effort test, but I think it's fine for this feature.

vittorioromeo:

eXpl0it3r:
And that's why one usually mocks IO related code paths 😄 
But I guess that's a topic of its own.

eXpl0it3r (approved):

PR #2203, opened 2022-09-15
Test reinterpreting a `sf::Color` as a `std::uint8_t*`

ChrisThrasher:
## Description

The purpose of this PR is to let us run the extended CI pipeline in search of potential issues with reinterpreting a `Color` as an array of `uint8_t`.

### DIFF
--- a/test/Graphics/Color.cpp
+++ b/test/Graphics/Color.cpp
@@ -3,6 +3,7 @@
 #include <doctest/doctest.h>
 
 #include <GraphicsUtil.hpp>
+#include <vector>
 
 TEST_CASE("sf::Color class - [graphics]")
 {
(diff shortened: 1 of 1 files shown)


eXpl0it3r:

vittorioromeo:

vittorioromeo:
BTW, this doesn't really prove anything. If we do this, I'd like to see strong guarantees that it is well-defined and allowed by the Standard.

Bromeon (approved):
Great idea!

ChrisThrasher:
> BTW, this doesn't really prove anything.

Agreed, but it's much-needed evidence to inform our opinion. We need to do this science since for all I know these tests will fail on a given platform which basically ends the conversation right now.

> I'd like to see strong guarantees that it is well-defined and allowed by the Standard.

Agreed, but I'm struggling to find anyone who can give an authoritative answer nor can I find any tool that expresses discontent with this. I'll keep looking.

Bromeon:
> BTW, this doesn't really prove anything. If we do this, I'd like to see strong guarantees that it is well-defined and allowed by the Standard.

It proves that the layout and conversion behavior of `sf::Color` is as expected _on the tested platforms_ (and to be very pedantic, in the given code context).

That doesn't mean it's a universal rule, but much better than nothing, and very likely enough for any practical purposes.

Bromeon:
Btw, CI failed due to Codecov upload error, so proof is not violated ;)

ChrisThrasher:
Rebased to get new CI jobs.

vittorioromeo:

ChrisThrasher:
I'm not yet convinced that I want to reimplement SFML such that it takes advantage of this `reinterpret_cast` but I still am in favor of adding these tests since I believe this is something users are doing and I want to know if any changes to the library or changes in our test environments break these tests. Because these tests are far from critical, we can easily remove these should they ever become a problem so I see the risk as minimal to zero.

PR #2206, opened 2022-09-19
Test install interface

ChrisThrasher:
## Description

This PR adds tests for the package install interface. I chose to skip the platforms that are either too hard to test or have a broken install interface. The project that is consuming the install interface does little more than find the package and build its headers. This seems to be sufficient.

These tests will better enable some fixes to install code and CMake changes that result in more generated files which we must ensure get installed correctly. I'm happy with how little code ended up being necessary to implement these.

### DIFF
--- a/test/install/CMakeLists.txt
+++ b/test/install/CMakeLists.txt
@@ -1,0 +1,15 @@
+cmake_minimum_required(VERSION 3.16)
+project(test-sfml-install CXX)
+
+# This skips the find_package call when building via add_subdirectory since that will fail under those circumstances
+# It's a CMake 3.16-compatible replacement to PROJECT_IS_TOP_LEVEL
+# https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html
+if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
+    if(NOT BUILD_SHARED_LIBS)
+        set(SFML_STATIC_LIBRARIES TRUE)
+    endif()
+    find_package(SFML 3.0.0 EXACT CONFIG REQUIRED COMPONENTS Graphics Network Audio)
+endif()
+
+add_executable(test-sfml-install Install.cpp)
+target_link_libraries(test-sfml-install PRIVATE SFML::Graphics SFML::Network SFML::Audio)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,6 +101,13 @@ jobs:
         files: ./build/coverage.out
         fail_ci_if_error: true
 
+    - name: Test Install Interface
+      if: matrix.platform.name != 'Android' && matrix.config.name != 'Frameworks' && matrix.config.name != 'iOS' && matrix.config.name != 'Static DRM'
+      shell: bash
+      run: |
+        cmake -S $GITHUB_WORKSPACE/test/install -B $GITHUB_WORKSPACE/test/install/build -DSFML_ROOT=$GITHUB_WORKSPACE/install -DCMAKE_VERBOSE_MAKEFILE=ON ${{matrix.platform.flags}} ${{matrix.config.flags}} ${{matrix.type.flags}}
+        cmake --build $GITHUB_WORKSPACE/test/install/build --config ${{ matrix.type.name == 'Debug' && 'Debug' || 'Release' }}
+
   format:
     name: Formatting on ${{ matrix.platform.name }}
     runs-on: ${{ matrix.platform.os }}
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -9,6 +9,8 @@ FetchContent_MakeAvailable(doctest)
 list(APPEND CMAKE_MODULE_PATH ${doctest_SOURCE_DIR}/scripts/cmake)
 include(doctest)
 
+add_subdirectory(install)
+
 add_library(sfml-test-main STATIC
     DoctestMain.cpp
     TestUtilities/SystemUtil.hpp
--- a/test/install/Install.cpp
+++ b/test/install/Install.cpp
@@ -1,0 +1,37 @@
+#include <SFML/Audio.hpp>
+#include <SFML/Graphics.hpp>
+#include <SFML/Network.hpp>
+
+// Instantiate some types from each module to test for linker issues. This program is not meant be ran.
+int main()
+{
+    // Audio
+    sf::InputSoundFile input_sound_file;
+    sf::Listener       listener;
+    sf::Music          music;
+    sf::Sound          sound;
+
+    // Graphics
+    sf::Color        color;
+    sf::Font         font;
+    sf::RenderWindow render_window;
+    sf::Sprite       sprite;
+    sf::Vertex       vertex;
+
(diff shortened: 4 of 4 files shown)


eXpl0it3r:

ChrisThrasher:
Rebased to get new CI jobs.

eXpl0it3r:

examples/find_package/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+cmake_minimum_required(VERSION 3.16)
+project(find_package CXX)
+
+if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
+    if(NOT BUILD_SHARED_LIBS)
+        set(SFML_STATIC_LIBRARIES TRUE)
+    endif()
```
vittorioromeo, line 7: Is this necessary? I would assume that `SFML_STATIC_LIBRARIES` is implied if `NOT BUILD_SHARED_LIBS`.

vittorioromeo (commented):
Generally looks good to me, but I'll defer to @eXpl0it3r or @binary1248 as I'm not confident in the build system / CMake stuff.

examples/find_package/CMakeLists.txt
ChrisThrasher: I believe so. `SFML_STATIC_LIBRARIES` is not initialized anywhere. If you omit this in static libraries builds, CI fails because you're implicitly asking for shared libraries yet only static libraries were built.

kimci86: This is confirmed by comments here:

https://github.com/SFML/SFML/blob/0e475f3f502d6f84744b9800d6c4fa058dd64043/cmake/SFMLConfig.cmake.in#L19-L25

ChrisThrasher: https://github.com/ChrisThrasher/SFML/actions/runs/3191808515

I'm testing this in CI as we speak (type?)

ChrisThrasher: I'm fine with SFMLConfig.cmake not automatically reacting to the value of BUILD_SHARED_LIBS even if that would be convenient for certain use cases.

examples/find_package/FindPackage.cpp
```diff
@@ -0,0 +1,7 @@
+#include <SFML/Audio.hpp>
+#include <SFML/Graphics.hpp>
+#include <SFML/Network.hpp>
```
eXpl0it3r, line 3: Just including headers isn't really enough to verify that the linking actually worked. You add the include directory and include headers, but then not link anything and it will build fine, because you didn't use any of the symbols.

Suggest to instantiate some class from each module. It can be something very basic.

ChrisThrasher: That's fair. Can this be a nonsense program that just creates lots of objects and bares no resemblance to what real SFML applications may look like?

examples/find_package/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+cmake_minimum_required(VERSION 3.16)
+project(find_package CXX)
+
+if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
```
eXpl0it3r, line 4: What does this check do? 🤔

examples/find_package/FindPackage.cpp
eXpl0it3r: At first I thought it was in the test directory, but if it's in the example directory, we should have something more or less useful, as they are examples after all 😅

examples/find_package/CMakeLists.txt
ChrisThrasher: Ah yes it's super unreadable haha. It's a CMake 3.16-compatible equivalent to [PROJECT_IS_TOP_LEVEL](https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html). It detects whether or not this project is being included via `add_subdirectory`. If it's included via `add_subdirectory` then I don't want the `find_package` code to be executed since that will fail.

examples/find_package/FindPackage.cpp
ChrisThrasher: Alternatively, I could convert an existing example into an installation test. Would that be better?

ChrisThrasher: This does make me wonder why I didn't put these in the test/ directory. That makes more sense because I don't want this code used as an example. In that case I would be able to write nonsense code that maximizes testing potential without worrying about others trying to copy it.

ChrisThrasher:
Because these are tests and not examples, I moved them from examples/ to tests/ and renamed some things accordingly. Now this program is called `test-sfml-install` which matches the names of all the other test executables. This is not meant to ever get run though so it will not get registered with CTest.

I also instantiated some objects in here to better test for linker issues. I picked a few different types from each module.

ChrisThrasher:

vittorioromeo:

ChrisThrasher:

examples/find_package/CMakeLists.txt
kimci86: I think having to answer this question during review means the code is not obvious so it can be a good idea to add a comment in the file.

eXpl0it3r: I agree, since it's not self explanatory, there should be a comment.

test/install/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+cmake_minimum_required(VERSION 3.16)
+project(test-sfml-install CXX)
+
+if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
```
eXpl0it3r, line 4: Add a comment to understand what this does

eXpl0it3r (changes requested):

test/install/CMakeLists.txt
ChrisThrasher: Fixed

ChrisThrasher:

test/install/CMakeLists.txt
eXpl0it3r: I was expecting more of comment on what this actually does and less of an explanation on what to do. I now know just want to do, but no idea why that's needed.

ChrisThrasher: > I was expecting more of comment on what this actually does

That was my intent. find_package is called if this is built as a standalone project. I can rephrase it though.

ChrisThrasher:

eXpl0it3r (approved):

test/install/CMakeLists.txt
eXpl0it3r: My bad, I read it as instruction "Call x to do y". I guess even now it reads more like an instruction than a comment, I'd probably expect "(This) Skips the find_package call below...", but maybe I'm just tired, so I guess it's okay

ChrisThrasher: I found a better way to write it such that it no longer sounds imperative. I'm fine taking our time to get these comments right.

ChrisThrasher:

eXpl0it3r:
🥳

PR #2210, opened 2022-09-20
Copy OpenAL DLL into runtime directory alongside other DLLs and EXEs

ChrisThrasher:
## Description

Backporting changes from master that ensure all DLLs get put in the same place which make them easy to copy out and makes it easier to run unit tests and example programs on DLL platforms. I'd tried to exactly match what's already in `master` so that a future back merge goes as smoothly as possible.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -265,6 +265,9 @@ if(SFML_OS_MACOSX)
     set(XCODE_TEMPLATES_ARCH "\$(NATIVE_ARCH_ACTUAL)")
 endif()
 
+# set the output directory for SFML DLLs and executables
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
+
 # enable project folders
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 set_property(GLOBAL PROPERTY PREDEFINED_TARGETS_FOLDER "CMake")
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -311,17 +311,7 @@ function(sfml_add_test target SOURCES DEPENDS)
     endif()
     
     # Add the test
-    add_test(${target} ${target})
-
-    # If building shared libs on windows we must copy the dependencies into the folder
-    if (WIN32 AND BUILD_SHARED_LIBS)
-        foreach (DEPENDENCY ${DEPENDS})
-            add_custom_command(TARGET ${target} PRE_BUILD
-                                COMMAND ${CMAKE_COMMAND} -E copy
-                                $<TARGET_FILE:${DEPENDENCY}>
-                                $<TARGET_FILE_DIR:${target}>)
-        endforeach()
-    endif()
+    add_test(NAME ${target} COMMAND ${target})
 endfunction()
 
 # Create an interface library for an external dependency. This virtual target can provide
--- a/src/SFML/Audio/CMakeLists.txt
+++ b/src/SFML/Audio/CMakeLists.txt
@@ -93,3 +93,11 @@ endif()
 target_link_libraries(sfml-audio
                       PUBLIC sfml-system
                       PRIVATE VORBIS FLAC)
+
+if(SFML_OS_WINDOWS AND NOT SFML_USE_SYSTEM_DEPS)
+    add_custom_command(
+        TARGET sfml-audio
+        COMMENT "Copy OpenAL DLL"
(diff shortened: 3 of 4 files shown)


eXpl0it3r:

ChrisThrasher:
Waiting for full CI results then I have a tiny whitespace fix to push then this PR should be good to go.

ChrisThrasher:
Screen Shot 2022-09-20 at 4 32 59 PM

All green. Whitespace fix incoming. No more changes planned.

eXpl0it3r:

CMakeLists.txt
```diff
@@ -265,6 +265,9 @@ if(SFML_OS_MACOSX)
     set(XCODE_TEMPLATES_ARCH "\$(NATIVE_ARCH_ACTUAL)")
 endif()
 
+# set the output directory for SFML DLLs and executables
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
```
eXpl0it3r, line 269: Should this be done no a per-target basis with `set_target_properties`?

ChrisThrasher: I can't think of any reason we'd want to do that.

eXpl0it3r: Any reasons to set it globally instead of target specific then?

ChrisThrasher: Must less code is required plus it keeps the door open for easily converting this into an option in the future which I think may be necessary to make sure the CMake template can easily put all these DLLs alongside the executables that users write.

eXpl0it3r: Not sure that really convinced me (beyond less code not really being an argument). I'm no expert, but I've been told on multiple occasions not to use global modifiers, as they can also easily pollute consuming projects and instead configure things on the target themselves.

Maybe @friendlyanon has some insights on this?

friendlyanon: `CMAKE_RUNTIME_OUTPUT_DIRECTORY` is a **user-controllable** cache variable. It shouldn't be hardcoded in project code.

Ideally, SFML would just use a package manager where this is taken care for you or you modify `PATH` manually. Windows lacks an RPATH mechanism, so what can you do.

ChrisThrasher: > ...not to use global modifiers

This is generally sound advice. The problem comes from people who use commands like `include_directories` when they need to specify `INTERFACE` requirements since those global modifiers only specify private details of a target. This leads to people writing targets that fail to specify all of their usage requirements. That's not a problem in our case since we don't want this to be a usage requirement of our targets.

> as they can also easily pollute consuming projects and instead configure things on the target themselves.

I'm fairly certain that as-written, setting this variable will not change where downstream projects place their own binaries so there is no pollution to be had. I do need to make this variable a CACHE variable so it can more easily be changed by users.

JonnyPtn: Yeah currently this will affect all targets so we should just set it on the SFML targets that need it

ChrisThrasher: I want this to affect all targets.

JonnyPtn: even non-SFML ones? this means if I use fetchcontent for SFML my exe is going to go into SFML's build directory (assuming it's a cached value as you say)

ChrisThrasher: Ah you're right. When setting this as a cache variable (which is something I think we need to do) then that means projects using SFML via `add_subdirectory` will have their own executables placed into build/_deps/sfml-build/bin which is definitely not desirable.

I'll experiment with setting them on a per-target basis to see if that lets us achieve both the convenient behavior we want while also letting downstream users pick the directory where all the SFML DLLs go. I think there's room to improve on the status quo but I'm still unsure if the ideal solution I'd like is possible.

JonnyPtn: I'd imagine a better solution would be to make sure openAL (and any other required) dlls are part of the RUNTIME_DLL list. 

Not something I've done before but expect it should be doable

ChrisThrasher:
Some valid issues were brought up so I'm marking this as a draft until I can get this PR to a state I'm happy with.

CMakeLists.txt
JonnyPtn: Have since proved we can get openAL on the runtime DLL list so can probably just bail on this

JonnyPtn (changes requested):

ChrisThrasher:
https://github.com/ChrisThrasher/SFML/commit/3d9e464efb907859beee1e5a0e963b26280828e4

I have a solution working that specifies the runtime output directory at configuration. I think a long-term solution will involve this in some capacity. This does mean though that anyone building on Windows will have to specify this option to run the tests. This would not be a problem however if we changed the default library type from Shared to Static. Is it worth making SFML less ergonomic to build to make it more correct? Can we switch to Static libs by default so this is only a problem for those opting into Shared library builds? I feel like we keep running into issues with this Shared library default value and I hope we reconsider.

Issue #2211, opened 2022-09-21
Windows compilation with Visual Studio 2019 on 3.x master branch fails on int64_t usages in sf::Time


PierceLBrooks:
- Subject of the issue

When compiling a fresh clone of the SFML repo on Windows with Visual Studio 2019, the "Time" class errors with complaints about the "int64_t" type not being found in the "std" namespace:

```
C:\UserData\SFML\include\SFML/System/Time.inl(223,42): error C2039: 'int64_t': is not a
member of 'std' [C:\UserData\SFML\build\src\SFML\System\sfml-system.vcxproj]
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.28.29333\include\chrono(24): message : s
ee declaration of 'std' [C:\UserData\SFML\build\src\SFML\System\sfml-system.vcxproj]
```

- My environment

* My OS used: Windows 10
* My version of SFML: git master
* My compiler and compiler version used: Visual Studio 2019

- Steps to reproduce

```
git clone https://github.com/SFML/SFML.git
cd SFML
git checkout master
mkdir build
cd build
cmake ..
cmake --build .
```

- Expected behavior

SFML should compile.

- Actual behavior

SFML does not compile.

- P.S.

I believe I have found the source of this issue and intend to submit a pull request to resolve it soon.

ChrisThrasher:
https://github.com/SFML/SFML/actions/runs/3047035798

Do you know why our CI pipeline may not have caught this? Simply adding `<cstdint>` is likely the easy fix but I want to know how we missed a build break for a common OS and toolchain.

PierceLBrooks:
@ChrisThrasher The pipeline's invokations of CMake seems fine to me, I am currently updating my copy of Visual Studio 2019 in order to see whether that may resolve it. In any case, I still intend to open a pull request for inserting that missing include line where it belongs at https://github.com/SFML/SFML/blob/master/include/SFML/System/Time.hpp#L33 unless that is felt to be unnecessary by your team.

ChrisThrasher:
Go ahead with the PR. It's exactly what I would have done anyways. I just hope in the meantime we can figure out how this slipped through the cracks.

ChrisThrasher:
[...]

PR #2212, opened 2022-09-21
Fix Visual Studio 2019 complaints regarding the int64_t type not being in the std namespace

PierceLBrooks:
This resolves the issue that I logged here:
https://github.com/SFML/SFML/issues/2211

### DIFF
--- a/include/SFML/System/Time.hpp
+++ b/include/SFML/System/Time.hpp
@@ -31,6 +31,7 @@
 #include <SFML/System/Export.hpp>
 
 #include <chrono>
+#include <cstdint>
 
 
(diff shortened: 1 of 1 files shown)


ChrisThrasher (approved):
Thanks!

ChrisThrasher:
TL;DR This problem only exists in older version of VS 2019. Because we can't rule out people still using such versions I'm fine merging this just in case.

eXpl0it3r (approved):

eXpl0it3r:

eXpl0it3r:
Thank you for the fix! 🥳

PR #2214, opened 2022-09-27
Return of the Testing PRs (`sf::Packet`)

ChrisThrasher:
## Description

I thought the `sf::Packet` tests were lacking so I shored them up to bring them up to the standard of all the other unit tests.

### DIFF
--- a/test/Network/Packet.cpp
+++ b/test/Network/Packet.cpp
@@ -2,20 +2,62 @@
 
 #include <doctest/doctest.h>
 
+#include <array>
 #include <limits>
 
-#define CHECK_PACKET_STREAM_OPERATORS(expected) \
-    do                                          \
-    {                                           \
-        sf::Packet packet;                      \
-        packet << expected;                     \
-        decltype(expected) received;            \
-        packet >> received;                     \
-        CHECK(expected == received);            \
+#define CHECK_PACKET_STREAM_OPERATORS(expected)              \
+    do                                                       \
+    {                                                        \
+        sf::Packet packet;                                   \
+        packet << expected;                                  \
+        CHECK(packet.getReadPosition() == 0);                \
+        CHECK(packet.getData() != nullptr);                  \
+        CHECK(packet.getDataSize() == sizeof(expected));     \
+        CHECK(!packet.endOfPacket());                        \
+        CHECK(static_cast<bool>(packet));                    \
+                                                             \
+        decltype(expected) received;                         \
+        packet >> received;                                  \
+        CHECK(packet.getReadPosition() == sizeof(expected)); \
+        CHECK(packet.getData() != nullptr);                  \
+        CHECK(packet.getDataSize() == sizeof(expected));     \
+        CHECK(packet.endOfPacket());                         \
+        CHECK(static_cast<bool>(packet));                    \
+        CHECK(expected == received);                         \
     } while (false)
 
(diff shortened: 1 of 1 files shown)


eXpl0it3r:

test/Network/Packet.cpp
```diff
@@ -1,53 +1,95 @@
...
+                                                             \
+        decltype(expected) received;                         \
+        packet >> received;                                  \
+        CHECK(packet.getReadPosition() == sizeof(expected)); \
+        CHECK(packet.getData() != nullptr);                  \
+        CHECK(packet.getDataSize() == sizeof(expected));     \
+        CHECK(packet.endOfPacket());                         \
+        CHECK(static_cast<bool>(packet));                    \
+        CHECK(expected == received);                         \
+    } while (false)
```
eXpl0it3r, line 27: I'm not so convinced about the arguments provided on Discord regarding use of a macro here.  
The TEST_CASE + SUBCASE + CHECK should IMHO be enough information to identify and locate a failing test, so we don't need preserve line number information, not even sure if that will be "retained" given that you expand a single line to 19 lines or so.

In which cases are TEST_CASE + SUBCASE + CHECK not enough to investigate a test failure?

ChrisThrasher: Here's what failures look like when using macros to create test utilities.
```
TEST CASE:  sf::Packet class - [network]
  Stream operators
  std::int8_t

/Users/thrasher/Projects/sfml/test/Network/Packet.cpp:65: ERROR: CHECK( packet.getReadPosition() == 1 ) is NOT correct!
  values: CHECK( 0 == 1 )
  ```
Note how it mentions line 65 with the failure. You can CTRL+Click on this to go straight to the offending test. These tests are more precise because they tell you precisely where the failure occurred. Now let's go back to using a function.
```
TEST CASE:  sf::Packet class - [network]
  Stream operators
  std::int8_t

/Users/thrasher/Projects/sfml/test/Network/Packet.cpp:13: ERROR: CHECK( packet.getReadPosition() == 1 ) is NOT correct!
  values: CHECK( 0 == 1 )
  ```
Now we're looking at the line in the utility function that failed. CTRL+Clicking on this takes you to the utility function which isn't particularly useful because you don't know what data was passed to that function. You can still see that it failed in the `std::int8_t` subcase but you cannot know which of the four expressions in that subcase failed. Information is lost and you have to spend more time figuring out where the test failure actually occurred.

---

I know, macros are ugly and we do the right thing by largely avoiding them in SFML itself but test code has very different requirements like needing to know line numbers which make macros the superior, albeit unintuitive, choice for abstractions like this. It's not much different from how all pre-C++20 test frameworks use macros. Preserving line number and being able to print out the failing expression is very useful and just not something you can do without macros in C++17.

eXpl0it3r: I guess for now this is okay, as there isn't really a great alternative with the current setup. Personally, I'd prefer something like [Parameterized Test](https://github.com/doctest/doctest/blob/master/doc/markdown/parameterized-tests.md) at one point, instead of packing multiple different test into one subcase.

eXpl0it3r (approved):

test/Network/Packet.cpp
ChrisThrasher: I want to look deeper into parameterizing tests because it may have the potential to drastically improve test coverage and write less code while we're at it.

Issue #2217, opened 2022-09-28
Compiler warning in Http class


b0guslaw:
## Wrestrict triggered in Http.cpp

The use of `std::string::insert()` in Http.cpp:32 causes the compiler warning -Wrestrict to trigger. This was noticed in a C++20 environment.

## Your environment

* EndeavourOS with gcc 12.2.0
* 2.6.x but occurs on all branches including master
* Set language standard to C++20
* -Wall -Werror -Wpedantic

## Steps to reproduce

Simply build a minimal SFML example from source

## Expected behavior

No warnings should be triggered while building SFML

## Actual behavior

```
inlined from ‘void sf::Http::Request::setUri(const std::string&)’ at <snip>/build/_deps/sfml-src/src/SFML/Network/Http.cpp:82:21:
/usr/include/c++/12.2.0/bits/char_traits.h:431:56: error: ‘void* __builtin_memcpy(void*, const void*, long unsigned int)’ accessing 9223372036854775810 or more bytes at offsets [2, 9223372036854775807] and 1 may overlap up to 9223372036854775813 bytes at offset -3 [-Werror=restrict]
  431 |         return static_cast<char_type*>(__builtin_memcpy(__s1, __s2, __n));
      |                                        ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
make[2]: *** [_deps/sfml-build/src/SFML/Network/CMakeFiles/sfml-network.dir/build.make:90: _deps/sfml-build/src/SFML/Network/CMakeFiles/sfml-network.dir/Http.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:389: _deps/sfml-build/src/SFML/Network/CMakeFiles/sfml-network.dir/all] Error 2
make: *** [Makefile:156: all] Error 2
```

kimci86:
This is a known bug in GCC when using C++20 and optimizations. See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105651

ChrisThrasher:
I think the patch in #2218 is a reasonable workaround. I want people using cutting edge compilers with warnings enabled. `std::string` provides 10 redundant APIs for every operation so it's an easy to switch to an iterator-based API instead. Plus I think it makes the code marginally more expressive which is nice.

eXpl0it3r:
Fixed with #2218

PR #2218, opened 2022-09-28
Use iterator and correct string insert

b0guslaw:
* [x] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before? Mentioned it on Discord

----

## Description

This PR fixes the compiler warning I have encountered when building SFML in C++20 on gcc 12 by using a different, more appropriate function signature.

This PR is related to the issue #2217 

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

The fix does not change any functionality and as such no special tests are required

### DIFF
--- a/src/SFML/Network/Http.cpp
+++ b/src/SFML/Network/Http.cpp
@@ -79,7 +79,7 @@ void Http::Request::setUri(const std::string& uri)
 
     // Make sure it starts with a '/'
     if (m_uri.empty() || (m_uri[0] != '/'))
-        m_uri.insert(0, "/");
+        m_uri.insert(m_uri.begin(), '/');
 }
 
 


ChrisThrasher:

vittorioromeo:

vittorioromeo (approved):
I am happy with this change as it works around a compiler bug and makes the code a bit nicer (IMHO) at the same time. I see no risk in merging it, assuming CI is happy.

ChrisThrasher (approved):

eXpl0it3r:
Can you please retarget this to `2.6.x`?

eXpl0it3r (approved):
Generally don't want to work around compiler bugs that might be fixed soon-ish, but it's not like this hurts us.

vittorioromeo:

eXpl0it3r:
Thank you for this fix 🥳

PR #2220, opened 2022-10-01
Fix SoundRecorder example code

kimci86:
About a month ago, a user on discord was confused by `SoundRecorder`'s documentation because the example code is missing access specifiers. At least, the destructor needs to be public so the class can be instantiated. Here is a fix for that, plus missing namespace and semicolon.


ChrisThrasher:

ChrisThrasher (approved):

vittorioromeo (approved):

PR #2222, opened 2022-10-01
Fix minor inconsistency in type cast in voip example

kimci86:
because m_samples is a vector of Int16 values.

That's it for today. Thank you.

### DIFF
--- a/examples/voip/Server.cpp
+++ b/examples/voip/Server.cpp
@@ -85,7 +85,7 @@ private:
         // (don't forget that we run in two separate threads)
         {
             sf::Lock lock(m_mutex);
-            m_tempBuffer.assign(m_samples.begin() + static_cast<std::vector<sf::Int64>::difference_type>(m_offset), m_samples.end());
+            m_tempBuffer.assign(m_samples.begin() + static_cast<std::vector<sf::Int16>::difference_type>(m_offset), m_samples.end());
         }
 
         // Fill audio data to pass to the stream


eXpl0it3r:

examples/voip/Server.cpp
```diff
@@ -85,7 +85,7 @@ class NetworkAudioStream : public sf::SoundStream
         // (don't forget that we run in two separate threads)
         {
             sf::Lock lock(m_mutex);
-            m_tempBuffer.assign(m_samples.begin() + static_cast<std::vector<sf::Int64>::difference_type>(m_offset), m_samples.end());
+            m_tempBuffer.assign(m_samples.begin() + static_cast<std::vector<sf::Int16>::difference_type>(m_offset), m_samples.end());
```
vittorioromeo, line 88:
```suggestion
            m_tempBuffer.assign(m_samples.begin() + static_cast<decltype(m_samples)::difference_type>(m_offset), m_samples.end());
```

Maybe this expresses the intent better?

kimci86: You are right. However that is C++11 and I target 2.6.x branch which uses C++03.

vittorioromeo (approved):

vittorioromeo:
@eXpl0it3r: we can merge into 2.6.x, right?

examples/voip/Server.cpp
ChrisThrasher: Back in SFML 3 land I would like to eventually find more of these places where the names of types are being repeated and replace them with `decltype`.

ChrisThrasher (approved):

examples/voip/Server.cpp
eXpl0it3r: Except this is targeting 2.6.x thus no decltype for you 😉

eXpl0it3r:
> @eXpl0it3r: we can merge into 2.6.x, right?

Fixes are fine, features should be discussed first.

PR #2231, opened 2022-10-06
WIP - safer 'std::optional'-based 'sf::Font' API 

vittorioromeo:
Same as #2228, but for `sf::Font`.

### DIFF
--- a/examples/android/app/src/main/jni/main.cpp
+++ b/examples/android/app/src/main/jni/main.cpp
@@ -90,11 +90,8 @@ int main(int argc, char* argv[])
     image.setPosition(sf::Vector2f(screen.size) / 2.f);
     image.setOrigin(sf::Vector2f(texture.getSize()) / 2.f);
 
-    std::optional<sf::Font> font = sf::Font::loadFromFile("tuffy.ttf");
-    if (!font.has_value())
-        return EXIT_FAILURE;
-
-    sf::Text text("Tap anywhere to move the logo.", *font, 64);
+    sf::Font font = sf::Font::loadFromFile("tuffy.ttf").value();
+    sf::Text text("Tap anywhere to move the logo.", font, 64);
     text.setFillColor(sf::Color::Black);
     text.setPosition({10, 10});
 
--- a/examples/android/app/src/main/jni/main.cpp
+++ b/examples/android/app/src/main/jni/main.cpp
@@ -90,11 +90,11 @@ int main(int argc, char* argv[])
     image.setPosition(sf::Vector2f(screen.size) / 2.f);
     image.setOrigin(sf::Vector2f(texture.getSize()) / 2.f);
 
-    sf::Font font;
-    if (!font.loadFromFile("tuffy.ttf"))
+    std::optional<sf::Font> font = sf::Font::loadFromFile("tuffy.ttf");
+    if (!font.has_value())
         return EXIT_FAILURE;
 
-    sf::Text text("Tap anywhere to move the logo.", font, 64);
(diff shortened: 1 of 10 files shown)


examples/android/app/src/main/jni/main.cpp
```diff
@@ -90,11 +90,11 @@ int main(int argc, char* argv[])
     image.setPosition(sf::Vector2f(screen.size) / 2.f);
     image.setOrigin(sf::Vector2f(texture.getSize()) / 2.f);
 
-    sf::Font font;
-    if (!font.loadFromFile("tuffy.ttf"))
+    std::optional<sf::Font> font = sf::Font::loadFromFile("tuffy.ttf");
+    if (!font.has_value())
```
Bromeon, line 94: This could simply be `if (!font)`, no?

Might be more consistent if we later use expressions like `if (std::optional<T> var = ...)`.

include/SFML/Graphics/Font.hpp
```diff
@@ -124,7 +117,7 @@ class SFML_GRAPHICS_API Font
     /// \see loadFromMemory, loadFromStream
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool loadFromFile(const std::filesystem::path& filename);
+    static std::optional<Font> loadFromFile(const std::filesystem::path& filename);
```
Bromeon, line 120: `std::optional` itself is not `[[nodiscard]]` to my knowledge -- so should these methods maybe add such an attribute?

After all, C++ even allows static methods to be called on instances (possibly with warnings, depends on QoI).
We could thus avoid mistakes of the form:
```cpp
// reload the font
myExistingFont.loadFromFile("bla");
```

vittorioromeo: +1, good point for the `.loadFromFile` issue.

src/SFML/Graphics/Font.cpp
```diff
@@ -192,32 +189,31 @@ bool Font::loadFromFile(const std::filesystem::path& filename)
...
-    return true;
+    return {result};
 
 #else
 
-    m_stream = std::make_unique<priv::ResourceStream>(filename);
-    return loadFromStream(*m_stream);
+    Font result;
+    result.m_stream = std::make_unique<priv::ResourceStream>(filename);
+    return loadFromStream(*result.m_stream);
```
kimci86, line 207: I think this will not work. There is an ownership problem here. `result` is a temporary owning the stream that will be deleted when the function is exited.
Maybe something like this would be better (untested):
```c++
    auto stream = std::make_unique<priv::ResourceStream>(filename);
    if(auto result = loadFromStream(*stream))
    {
        result->m_stream = std::move(stream);
        return result;
    }
    return std::nullopt;
```

include/SFML/Graphics/Font.hpp
```diff
@@ -330,6 +323,14 @@ class SFML_GRAPHICS_API Font
     Font& operator=(const Font& right);
 
 private:
+    ////////////////////////////////////////////////////////////
+    /// \brief Default constructor
+    ///
+    /// This constructor defines an empty font
+    ///
+    ////////////////////////////////////////////////////////////
+    Font();
```
ChrisThrasher, line 332: Do we even care about documenting private constructors?

ChrisThrasher (commented):
I won't repeat everything from my review of #2228 but those comments apply to this as well. This is great and I'm eager to get it merged!

ChrisThrasher:
Rebased on `master`

ChrisThrasher:

eXpl0it3r:
Superseded by #3020

PR #2233, opened 2022-10-07
Remove unnecessary static casts

ChrisThrasher:
## Description

Found some static casts that were not necessary and could be easily avoided.

### DIFF
--- a/include/SFML/Graphics/Color.inl
+++ b/include/SFML/Graphics/Color.inl
@@ -73,7 +73,7 @@ constexpr Color operator+(const Color& left, const Color& right)
 {
     const auto clampedAdd = [](std::uint8_t lhs, std::uint8_t rhs) -> std::uint8_t
     {
-        const int intResult = static_cast<int>(lhs) + static_cast<int>(rhs);
+        const int intResult = int{lhs} + int{rhs};
         return static_cast<std::uint8_t>(intResult < 255 ? intResult : 255);
     };
 
@@ -89,7 +89,7 @@ constexpr Color operator-(const Color& left, const Color& right)
 {
     const auto clampedSub = [](std::uint8_t lhs, std::uint8_t rhs) -> std::uint8_t
     {
-        const int intResult = static_cast<int>(lhs) - static_cast<int>(rhs);
+        const int intResult = int{lhs} - int{rhs};
         return static_cast<std::uint8_t>(intResult > 0 ? intResult : 0);
     };
 
@@ -105,8 +105,7 @@ constexpr Color operator*(const Color& left, const Color& right)
 {
     const auto scaledMul = [](std::uint8_t lhs, std::uint8_t rhs) -> std::uint8_t
     {
-        const auto uint16Result = static_cast<std::uint16_t>(
-            static_cast<std::uint16_t>(lhs) * static_cast<std::uint16_t>(rhs));
+        const auto uint16Result = static_cast<std::uint16_t>(std::uint16_t{lhs} * std::uint16_t{rhs});
         return static_cast<std::uint8_t>(uint16Result / 255u);
     };
 
--- a/test/TestUtilities/GraphicsUtil.cpp
+++ b/test/TestUtilities/GraphicsUtil.cpp
@@ -16,9 +16,8 @@ std::ostream& operator<<(std::ostream& os, const BlendMode& blendMode)
 
 std::ostream& operator<<(std::ostream& os, const Color& color)
 {
-    return os << "0x" << std::hex << color.toInteger() << std::dec << " (r=" << static_cast<int>(color.r)
-              << ", g=" << static_cast<int>(color.g) << ", b=" << static_cast<int>(color.b)
-              << ", a=" << static_cast<int>(color.a) << ")";
+    return os << "0x" << std::hex << color.toInteger() << std::dec << " (r=" << int{color.r} << ", g=" << int{color.g}
+              << ", b=" << int{color.b} << ", a=" << int{color.a} << ")";
 }
 
 std::ostream& operator<<(std::ostream& os, const Transform& transform)
--- a/src/SFML/Audio/Music.cpp
+++ b/src/SFML/Audio/Music.cpp
@@ -238,7 +238,7 @@ void Music::initialize()
     m_loopSpan.length = m_file.getSampleCount();
 
     // Resize the internal buffer so that it can contain 1 second of audio samples
-    m_samples.resize(static_cast<std::size_t>(m_file.getSampleRate()) * static_cast<std::size_t>(m_file.getChannelCount()));
+    m_samples.resize(m_file.getSampleRate() * m_file.getChannelCount());
 
     // Initialize the stream
     SoundStream::initialize(m_file.getChannelCount(), m_file.getSampleRate());
--- a/src/SFML/Graphics/Texture.cpp
+++ b/src/SFML/Graphics/Texture.cpp
@@ -331,7 +331,7 @@ Image Texture::copyToImage() const
     priv::TextureSaver save;
 
     // Create an array of pixels
-    std::vector<std::uint8_t> pixels(static_cast<std::size_t>(m_size.x) * static_cast<std::size_t>(m_size.y) * 4);
+    std::vector<std::uint8_t> pixels(m_size.x * m_size.y * 4);
 
 #ifdef SFML_OPENGL_ES
 
@@ -365,8 +365,7 @@ Image Texture::copyToImage() const
         // Texture is either padded or flipped, we have to use a slower algorithm
 
         // All the pixels will first be copied to a temporary array
-        std::vector<std::uint8_t> allPixels(
-            static_cast<std::size_t>(m_actualSize.x) * static_cast<std::size_t>(m_actualSize.y) * 4);
+        std::vector<std::uint8_t> allPixels(m_actualSize.x * m_actualSize.y * 4);
         glCheck(glBindTexture(GL_TEXTURE_2D, m_texture));
         glCheck(glGetTexImage(GL_TEXTURE_2D, 0, GL_RGBA, GL_UNSIGNED_BYTE, allPixels.data()));
 


ChrisThrasher:

test/TestUtilities/GraphicsUtil.cpp
```diff
@@ -16,9 +16,8 @@ std::ostream& operator<<(std::ostream& os, const BlendMode& blendMode)
 
 std::ostream& operator<<(std::ostream& os, const Color& color)
 {
-    return os << "0x" << std::hex << color.toInteger() << std::dec << " (r=" << static_cast<int>(color.r)
-              << ", g=" << static_cast<int>(color.g) << ", b=" << static_cast<int>(color.b)
-              << ", a=" << static_cast<int>(color.a) << ")";
+    return os << "0x" << std::hex << color.toInteger() << std::dec << " (r=" << +color.r << ", g=" << +color.g
+              << ", b=" << +color.b << ", a=" << +color.a << ")";
```
vittorioromeo, line 20: I don't think this is an improvement. The unary `operator+` is hard to spot, and the rules here are much less obvious than a `static_cast`. It seems that we are optimizing for conciseness compared to readability, and I'm not confident that is a worthwhile goal/metric.

vittorioromeo (dismissed):

test/TestUtilities/GraphicsUtil.cpp
eXpl0it3r: Yeah, I'd consider this a hack, as it works around a technical "issue" and destroys the readability of the code. A large percentage of C++ developers won't be able to tell you *what* `+color.r` does and even fewer *why" it's not just `color.r`.

Bromeon: I agree with vittorio and eXpl0it3r here. As much as I hate verbose `static_cast`, I'd probably use something like `toUint8` or so, which is at least somewhat intuitive.

One thing where I did find unary `+` useful though is for `enum class` and converting to their underlying type. So you could have similar conciseness as in C++98, but wouldn't need type-unsafe enums.

vittorioromeo: @Bromeon: Note that C++23 provides `std::to_underlying` (https://en.cppreference.com/w/cpp/utility/to_underlying) which clearly expresses that intent. I'd rather have a SFML polyfill of that compared to using `operator+`.

include/SFML/Graphics/Color.inl
```diff
@@ -73,9 +73,9 @@ constexpr bool operator!=(const Color& left, const Color& right)
 ////////////////////////////////////////////////////////////
 constexpr Color operator+(const Color& left, const Color& right)
 {
-    const auto clampedAdd = [](std::uint8_t lhs, std::uint8_t rhs) -> std::uint8_t
+    const auto clampedAdd = [](int lhs, int rhs) -> std::uint8_t
     {
-        const int intResult = static_cast<int>(lhs) + static_cast<int>(rhs);
+        const int intResult = lhs + rhs;
```
vittorioromeo, line 78: The name of this variable doesn't make much sense as part of this refactoring.

include/SFML/Graphics/Color.inl
```diff
@@ -73,9 +73,9 @@ constexpr bool operator!=(const Color& left, const Color& right)
...
         return static_cast<std::uint8_t>(intResult < 255 ? intResult : 255);
     };
```
vittorioromeo, line 81: Now we have implicit conversions in `clampedAdd`, but they are not visible anymore. Is this really an improvement?

vittorioromeo:
I am really not certain that this PR is worthwhile merging -- what problem is it solving? It is removing some `static_cast`s, but what is the benefit of that? A drawback is that now some casts are implicit (e.g. `std::uint8_t` -> `int`, or `std::uint8_t` -> `std::size_t`).

Bromeon:
The problem is really that in C++, integral promotions are so pervasive that you can almost not avoid them.
```cpp
std::uint8_t a = 3;
auto b = a + a;
// b is int
```

Should we use `static_cast` in such situations?
What if we assign the result to `uint8_t` again? 

These are not rhetorical question, I think it's hard to find a good convention, as "mark every conversion/promotion explicitly" will turn the code into a big mess. Ideally there could be some CI tweaks that follow certain rules of "explicitness", but I have not much hope that it would provide the degree of customizability we'd need.

test/TestUtilities/GraphicsUtil.cpp
ChrisThrasher: Fair. This is perhaps too clever.

ChrisThrasher:
> what problem is it solving?

The code is more complicated than it needs to be.

> what is the benefit of that?

We more clearly express when lossy conversions are needed and when they are not.

> A drawback is that now some casts are implicit

This is an upside to me. I don't want syntactic overhead when safe conversions are occurring.

---

I don't want this PR stealing attention from the bigger changes we have planned so if we can't reach a quick consensus I'm closing this. It's not consequential enough to warrant more than a tiny bit of attention.

ChrisThrasher:
I found a simpler way to remove these `static_cast`s through the use of list initialization. I see no good reason to use `static_cast`s for lossless conversions so this is a much better way to express that we don't actually want to discard information.

ChrisThrasher:

include/SFML/Graphics/Color.inl
ChrisThrasher: Removed implicit conversions in my latest version of this PR.

ChrisThrasher:

ChrisThrasher:
Found another `static_cast` to remove and rebased on `master`.

ChrisThrasher:

ChrisThrasher:
Can I bump this PR? Since the initial round of reviews I found a more satisfying way to express the explicit promotions we need. I think the PR is much more agreeable than it initially was. While I was at it I rebased on `master` too.

ChrisThrasher:

ChrisThrasher:

Bromeon:
```cpp
const int intResult = int{lhs} + int{rhs};
```
While very elegant, this looks like sorcery😁

I just wonder if that makes code truly more readable, or just shorter but leaves everyone wondering what's happening here. But maybe it's just me who is unfamiliar with this "idiom".

ChrisThrasher:
> I just wonder if that makes code truly more readable, or just shorter but leaves everyone wondering what's happening here.

It's certainly not a common feature to use since it's rare to need to explicitly promote an arithmetic type, but I think the compiletime safety benefits and added terseness make up for it's unfamiliarity.

include/SFML/Graphics/Color.inl
```diff
@@ -107,8 +107,7 @@ constexpr Color operator*(const Color& left, const Color& right)
 {
     const auto scaledMul = [](std::uint8_t lhs, std::uint8_t rhs) -> std::uint8_t
     {
-        const auto uint16Result = static_cast<std::uint16_t>(
-            static_cast<std::uint16_t>(lhs) * static_cast<std::uint16_t>(rhs));
+        const auto uint16Result = static_cast<std::uint16_t>(std::uint16_t{lhs} * std::uint16_t{rhs});
```
JonnyPtn, line 110: What's the reason for leaving one static cast here? could it not be braces init like the others on this line?

ChrisThrasher: I use braces when we need to do a promotion (`uint8_t` -> `uint16_t` loses no data so braces can be used) and `static_cast` when we need to discard information. Multiplying two `uint16_t`s produces an `int` which must be `static_cast`'ed down back to a `uint16_t`.

Bromeon:
Looks like the only thing left to decide is the idiom
```cpp
const int intResult = int{lhs} + int{rhs};
```
If other developers are fine with this, this PR is good-to-go from my side.

ChrisThrasher:

ChrisThrasher:
Rebased to catch up with `master`

include/SFML/Graphics/Color.inl
```diff
@@ -73,7 +73,7 @@ constexpr Color operator+(const Color& left, const Color& right)
 {
     const auto clampedAdd = [](std::uint8_t lhs, std::uint8_t rhs) -> std::uint8_t
     {
-        const int intResult = static_cast<int>(lhs) + static_cast<int>(rhs);
+        const int intResult = int{lhs} + int{rhs};
```
eXpl0it3r, line 76: I'd vote to keep the static_cast. It's clearer what's going on and I don't see the gain in saving a few characters.

include/SFML/Graphics/Color.inl
```diff
@@ -89,7 +89,7 @@ constexpr Color operator-(const Color& left, const Color& right)
 {
     const auto clampedSub = [](std::uint8_t lhs, std::uint8_t rhs) -> std::uint8_t
     {
-        const int intResult = static_cast<int>(lhs) - static_cast<int>(rhs);
+        const int intResult = int{lhs} - int{rhs};
```
eXpl0it3r, line 92: I'd vote to keep the static_cast. It's clearer what's going on and I don't see the gain in saving a few characters.

eXpl0it3r (changes requested):

include/SFML/Graphics/Color.inl
ChrisThrasher: The problem is, a static cast is more than we need. A static cast signals that we may be doing a narrowing conversion which in this case we don't need so not only are we making the code shorter, but we're also correctly expressing how this conversion ought not discard any information.

I looked through the C++ Core Guidelines for guidance and it seems to also recommend using `{}` in these cases where no narrowing is occurring. I believe using `{}` in this case is technically not even a cast but rather treated as construction.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-casts
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-casts-named
Screenshot 2022-12-30 at 11 06 23 AM

---

While I acknowledge that arguing over this particular example is not productive, this is a precedent I think is worth setting for the whole library since I believe we all agree that we ought to do as little casting as possible.

include/SFML/Graphics/Color.inl
ChrisThrasher: We gain the extra safety that a narrowing conversion isn't happening.

eXpl0it3r: Is it not? Looking at cppinsights, it just does a static_cast, so the only different is the difference in syntax, but maybe I misunderstand how cppinsights works 😅 

image

ChrisThrasher: I'm not entirely sure how cppinsights works but I can guarantee you that the two syntaxes are not identical and the braces are more restrictive in what they allow.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:
I found another cast that can be replaced with the construction of a wider type.

ChrisThrasher:

ChrisThrasher:

eXpl0it3r (approved):
After reading some more, it seems like the way forward, even if the syntax takes a bit getting used to.

ChrisThrasher:

PR #2234, opened 2022-10-07
More Assorted CMake Fixes

ChrisThrasher:
## Description

Successor to #2178. These are even more assorted build system improvements I've accumulated over the last few months bundled together into a single PR. Read the individual commit messages for more information about each change. This PR makes the most sense if digested on a commit-by-commit basis instead of looking at all changes at once.


ChrisThrasher:

test/CMakeLists.txt
```diff
@@ -121,12 +121,12 @@ if(SFML_ENABLE_COVERAGE AND OpenCppCoverage_FOUND)
 
     add_custom_command(TARGET runtests
                        COMMENT "Run tests"
-                       POST_BUILD COMMAND "${OpenCppCoverage_BINARY}" ARGS --quiet --export_type cobertura:"${CMAKE_BINARY_DIR}/coverage.out" --cover_children --excluded_modules "${COVERAGE_EXCLUDE}" --sources "${COVERAGE_SRC}" --sources "${COVERAGE_INCLUDE}" -- "${CMAKE_CTEST_COMMAND}" --output-on-failure -C $<CONFIG>
-    )
+                       POST_BUILD COMMAND "${OpenCppCoverage_BINARY}" ARGS --quiet --export_type cobertura:${PROJECT_BINARY_DIR}/coverage.out --cover_children --excluded_modules "${COVERAGE_EXCLUDE}" --sources "${COVERAGE_SRC}" --sources "${COVERAGE_INCLUDE}" -- "${CMAKE_CTEST_COMMAND}" --output-on-failure -C $<CONFIG>
```
kimci86, line 124: Aren't quotes needed around `${PROJECT_BINARY_DIR}/coverage.out`?

kimci86: Nervermind, after checking [CMake language manual](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#unquoted-argument), I see that quotes are not needed as long as the referenced variables do not contain an unescaped semicolon.

ChrisThrasher: That's right. And it turns out if the path to your CMakeLists.txt contains a semicolon in it, CMake bugs out anyways so there's no way that these lack of quotes can cause any issues. SFML already regularly omits these quotes and it's never been an issue as far as I'm aware. It's just one more subtle way to reduce the amount of code we write.

CMakeLists.txt
```diff
@@ -39,7 +39,7 @@ include(cmake/Config.cmake)
 
 # we use the paths from the cmake GNUInstallDirs module as defaults
 # you can override these if you like
-# https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html
+# https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html
```
JonnyPtn, line 42: would it not be better to link the minimum supported version?

CMakeLists.txt
```diff
@@ -229,13 +229,6 @@ if(SFML_OS_MACOSX)
         endif()
     endif()
 
-    # only the default x86_64 or arm64 are supported
-    foreach(arch IN LISTS CMAKE_OSX_ARCHITECTURES)
```
JonnyPtn, line 233: 32 bit iOS devices are a bit more recent than Macs, so want to make sure we don't plan to support them

eXpl0it3r (approved):

CMakeLists.txt
ChrisThrasher: That seems too hard to maintain since it's so easy to forget to update random links in comments when the minimum CMake version changes.

CMakeLists.txt
ChrisThrasher: I think 32 bit Macs are around a decade old at this point so I'm not worried about what may happen if someone tries to build SFML on such a machine.

CMakeLists.txt
eXpl0it3r: I don't think it's worth tracking some exact or minimum version here. Not like someone couldn't figure out what version is available to them if they need to know it.

PR #2235, opened 2022-10-09
Consolidate redundant gl.h headers

ChrisThrasher:
## Description

I consolidated 3 identical headers into 1. This saves thousands of lines of code, simplifies maintenance, and makes search results clearer by not repeating results.


ChrisThrasher:

vittorioromeo (approved):
LGTM, thanks!

eXpl0it3r:
I rather have self-contained examples than "save" some lines of code...

Bromeon:
Maybe we should get into the habit of letting at least 1-2 days pass before merging (unless _several_ people approve it before) 🙂 

Probably this PR seemed uncontroversial, but it that's often subjective.

vittorioromeo:
> Probably this PR seemed uncontroversial

Yeah, that's why I went ahead. It only touches examples and it's an internal refactoring change that is not visible to users. Otherwise I wouldn't have be so quick.

Happy to revert if you feel like this was the wrong move!

PR #2236, opened 2022-10-09
Revert "Consolidate redundant gl.h headers"

eXpl0it3r:
This reverts commit 7a5b862fd0876f8820f6aec74db63abf8d4a5d31 from #2235 

In my option each example should be as self-contained as possible, thus having a header duplicated makes sense from that perspective.  
Additionally, having an `include/` directory in the examples directory can be a bit confusing.


vittorioromeo:
I disagree with this rationale, I feel like it's entirely reasonable for multiple SFML examples to share some code between them, thus a `examples/include` folder is not confusing or weird IMHO.

Not going to block this because the status quo was changed, but I really think that @ChrisThrasher's refactoring was a step in the right direction -- copy pasting the same 3rd party header multiple times is a sign that we're doing something poorly.

Bambo-Borris:
_(Just repeating something discussed in the Discord server for the eyes of those who aren't in there)._
I don't necessarily agree/disagree with the original PR but I would like to raise the point that maybe PRs of a certain size need to live slightly longer so at least more maintainer & contributor eyes see them before they are merged. Maybe there is a rationale to why x is done, and maybe the rationale is worth changing but at least seeing that rationale explained in the PR prior to merging means a PR for reverting might never come about.

ChrisThrasher:
As someone with months-old PRs waiting for more reviews I'm very reluctant to propose SFML slow down the rate at which we merge changes but I acknowledge that it's reasonable to wait 12-24 hours on substantially sized PRs. Time zone realities and professional obligations mean it's very easy for someone to not see a new PR until near a day or more after it's submitted.

My best idea for systematically preventing more issues like this is to require that PRs get 2 non-author maintainer approvals before merging. Perhaps we raise that to 3 approvals for API changes. That means in the case of #2235 I would not count as an approver and a 2nd maintainer would have had to approve it as well. Practically speaking the time it will take to reach 2 non-author maintainer approvals will exceed this 12-24 hour time frame. This still means that simple PRs can get merged quickly so I don't think it will compromise our ability to get stuff done.

ChrisThrasher:
> In my option each example should be as self-contained as possible

Self-containment isn't bad but we already acknowledge the practical limits of this. For example, the examples use an external CMake macro to declare and build the targets. If we actually wanted examples to be as self contained as possible we would rewrite much or all of these CMake code in each example's build script but we don't do that because we acknowledge that self-containment isn't more important than a huge increase in code-reuse.

> Additionally, having an include/ directory in the examples directory can be a bit confusing.

How about we move this gl.h header somewhere inside extlibs instead? We've already had to hack around having 3rd party code in examples/ so it would be nice to move more of that outside of our 1st party examples/ directory.

eXpl0it3r:
> How about we move this gl.h header somewhere inside extlibs instead? We've already had to hack around having 3rd party code in examples/ so it would be nice to move more of that outside of our 1st party examples/ directory.

That's an even worse idea as you're then mixing non-library concerns in the extlibs folder

JonnyPtn (approved):
I'd agree with keeping the examples self-contained

Bromeon:
> Self-containment isn't bad but we already acknowledge the practical limits of this. For example, the examples use an external CMake macro to declare and build the targets. If we actually wanted examples to be as self contained as possible we would rewrite much or all of this CMake code in each example's build script but we don't do that because we acknowledge that self-containment isn't more important than a huge increase in code-reuse.

I think this is a good argument. At this point it can become a bit arbitrary what "self-contained" means. And there's of course the practical aspect -- `gl.h` is probably not the header that changes often, but if we add more shared headers over time, it's easier and less error-prone if they need to be changed in one place than in multiple. E.g. there was this discussion about utility functions for resource loading a while ago -- had it been accepted, it would have been a good contender for shared headers.

But the aspect I find the most important: a single header is a very straightforward way to signal to the user "this code is exactly the same across examples".

That said, if we don't find consensus, we will need to revert, as the original PR was merged too quickly.

Issue #2139, opened 2022-06-21
Error handling in SFML 3


Bromeon:
Back in 2015, I created several forum threads around the topic of error handling/reporting:
* [to promote the use of assertions](https://en.sfml-dev.org/forums/index.php?topic=19427)
* [to discuss logging with `sf::err()`](https://en.sfml-dev.org/forums/index.php?topic=19282)

Some discussions around the topic date back to 2011 and 2013:
* [why SFML didn't use exceptions](https://en.sfml-dev.org/forums/index.php?topic=5228)
* [Exceptions in SFML 3?](https://en.sfml-dev.org/forums/index.php?topic=13240)

These are quite interesting, already to understand historic context and how SFML 1 and 2 evolved to be without exceptions.

---

# Error handling in SFML 3

Now the year is 2022, a lot has changed, and at last we have the option to introduce breaking changes and redesign the API for better userfriendlyness. Handling errors is an important topic, and there are different options to tackle them in SFML 3. I'll jump right into them:

1. **Boolean return types**
   E.g. SFML 2 `sf::Image::loadFromFile()`
   * ✔️ Simple.
   * ✔️ Easy to port to bindings.
   * :x: Can be forgotten (`[[nodiscard]]` is needed on every single method).
   * ❌ Needs separate output parameters.
   * ❌ No way to transport error messages, needs `sf::err()` side channel.

2. **Optionals**
   Basically all the disadvantages of `bool` except no need for output parameters.
   * ✔️ Simple to understand and document.
   * ✔️ Easy to port to bindings.
   * :x: Can be forgotten (`[[nodiscard]]` is needed on every single method).
   * ❌ No way to transport error messages.

3. **Assertions**
   Listed for completeness, but they are not really a solution for runtime errors. But we might want to use them more often to complement other error mechanisms in the case of logic errors.
   * ✔️ Immediate breakpoint, must be fixed.
   * ✔️ No runtime overhead in release.
   * :x: Only useful for errors that are bugs (not loading files etc).
   * ❌ Do not allow the user to be lenient about certain errors.

5. **Exceptions**
   * ✔️ User cannot ignore them.
   * ✔️ Automatically propagate upward.
   * ✔️ Can be polymorphic, allowing automatic fallbacks and higher-level handling.
   * ✔️ Can theoretically transport extra data beyond message, although rarely done in practice.
   * ✔️ Can be used in constructors.
[...]

PR #2237, opened 2022-10-09
WIP - Replace font load error handling with exceptions

eXpl0it3r:
## Description

This is a counter proposal to error handling as demonstrated with #2231 by leveraging exceptions instead of optionals.

For general error handling discussions, please use #2139  
For specific error handling discussions with exceptions, comments are highly appreciated! 🙂 

Longer comment will soon follow on #2139, but some of the advantages I see over optionals:
- Allows for carrying an error message into the user code
- Tried and true concept, you'll find a lot of developers instantly getting exceptions, where optionals may need some more googling
- Keeps the API simple, no need for `->` accessors or other casts
- Handle errors at the layer where you want
- (IDE support for stack traces and pointing to where the error occurred)

Additionally, this also addresses the discussion point about non-default constructability.

## How to test this PR?

**Note:** The simple case presented here doesn't demonstrate the full power of exception handle

```cpp
#include <SFML/Graphics.hpp>
#include <iostream>

int main()
{
    auto window = sf::RenderWindow(sf::VideoMode{{1280u, 720u}}, "Minimal, complete and verifiable example");

    try
    {
        const auto font = sf::Font::loadFromFile("MyFont.ttf");
        auto text = sf::Text{"Hello World", font, 100};

        while (window.isOpen())
        {
            {
                {
                    window.close();
                }
            }

            window.clear();
            window.draw(text);
            window.display();
        }
    }
    catch (sf::LoadException& e)
    {
        std::cout << e.what();
        return -1;
    }
}
```

image

### DIFF
--- a/doc/mainpage.hpp
+++ b/doc/mainpage.hpp
@@ -27,9 +27,7 @@
 ///     sf::Sprite sprite(texture);
 ///
 ///     // Create a graphical text to display
-///     sf::Font font;
-///     if (!font.loadFromFile("arial.ttf"))
-///         return EXIT_FAILURE;
+///     sf::Font font = sf::Font::loadFromFile("arial.ttf");
 ///     sf::Text text("Hello SFML", font, 50);
 ///
 ///     // Load a music to play
--- a/examples/android/app/src/main/jni/main.cpp
+++ b/examples/android/app/src/main/jni/main.cpp
@@ -90,8 +90,7 @@ int main(int argc, char* argv[])
     image.setPosition(sf::Vector2f(screen.size) / 2.f);
     image.setOrigin(sf::Vector2f(texture.getSize()) / 2.f);
 
-    sf::Font font;
-    if (!font.loadFromFile("tuffy.ttf"))
+    sf::Font font = sf::Font::loadFromFile("tuffy.ttf"))
         return EXIT_FAILURE;
 
     sf::Text text("Tap anywhere to move the logo.", font, 64);
--- a/examples/android/app/src/main/jni/main.cpp
+++ b/examples/android/app/src/main/jni/main.cpp
@@ -90,7 +90,7 @@ int main(int argc, char* argv[])
     image.setPosition(sf::Vector2f(screen.size) / 2.f);
     image.setOrigin(sf::Vector2f(texture.getSize()) / 2.f);
 
-    sf::Font font = sf::Font::loadFromFile("tuffy.ttf"))
+    sf::Font font = sf::Font::loadFromFile("tuffy.ttf"));
         return EXIT_FAILURE;
 
     sf::Text text("Tap anywhere to move the logo.", font, 64);
--- a/examples/cocoa/CocoaAppDelegate.mm
+++ b/examples/cocoa/CocoaAppDelegate.mm
@@ -55,19 +55,18 @@ struct SFMLmainWindow
         unsigned int wh = renderWindow.getSize().y;
         sprite.setPosition(sf::Vector2f(ww, wh) / 2.f);
 
-        if (!font.loadFromFile(resPath + "/tuffy.ttf"))
-            NSLog(@"Couldn't load the font");
+        font = sf::Font::loadFromFile(resPath + "/tuffy.ttf")
 
         text.setFillColor(sf::Color::White);
         text.setFont(font);
     }
 
-    sf::RenderWindow renderWindow;
-    sf::Font         font;
-    sf::Text         text;
-    sf::Texture      logo;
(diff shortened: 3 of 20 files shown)


vittorioromeo:
Not sure if I have the energy for a full-on "ADTs versus exceptions" discussion, but I will quote myself [from a while ago, on Reddit](https://old.reddit.com/r/cpp/comments/vr771l/exceptions_yes_or_no/ietkbwc/):

> If a function can fail, and (as the caller) I am expected to react to such failure, I want to know *when* and *how* the function can fail. Therefore, I want the outcome of the function to be visible as part of the type system, rather than be completely hidden from its declaration. For such functions, I would use algebraic data types such as `std::optional`, `std::variant`, `std::expected`, etc.
> 
> I strongly believe that most functions fit the situation mentioned above. Exceptions might have a role for errors that cannot be reasonably immediately handled such as `std::bad_alloc`, but honestly I've never caught and resolved such an exception in any real-world code.
> 
> My conclusion is: exceptions most of the time obfuscate the control flow of your code and make failure cases less obvious to the reader. Prefer *not* using exceptions, unless it's a rare case that can only be reasonably handled a few levels upstream.

Just by looking at your example, we can think about the consequences of introducing exception-based error handling in SFML:

- Prior to this PR, SFML promised to never throw any exception. Now, users of SFML have to know that literally every line of SFML code they write might throw. However, most things in SFML will likely never throw an exception, yet we cannot make that promise on a library-wide basis anymore.

- Due to the implicit nature of exceptions, it is not clear what code inside a `try` block can throw. In your example, `sf::Font::loadFromFile("MyFont.ttf")` is the only expression that can throw a `sf::LoadException`, yet you wrapped the whole program in a `try { ... } catch`, reducing clarity on where the error can actually happen.

Other examples where `std::optional` or something like `std::expected` shine below. Explicit vs implicit error handling:

```cpp
// WITH EXCEPTIONS
const auto font = sf::Font::loadFromFile("MyFont.ttf"); // implicit error handling

// WITH ADTs
const auto font = sf::Font::loadFromFile("MyFont.ttf").value(); // explicit error handling
```

Falling back to a default font:

```
// WITH EXCEPTIONS
const auto font = [&]
{
    try 
    {
        return sf::Font::loadFromFile("MyFont.ttf");
    }
    catch (const sf::LoadException& e)
    {
        return defaultFont;
    }
}();

// WITH ADTs
const auto font = sf::Font::loadFromFile("MyFont.ttf").value_or(defaultFont); // ;)
```

Bromeon:
One problem I see here is that the `try` block has a far too wide scope. Only one line can fail, yet almost the entire function is covered in `try`, for technical reasons only -- the variable must remain in scope.

Unfortunately, this is a general problem with exception handling.
With optionals you could do this:
```cpp
const std::optional<sf::Font> maybeFont = sf::Font::loadFromFile("MyFont.ttf");
if (!maybeFont)
    return EXIT_FAILURE;
const sf::Font font = *maybeFont;
// Now use `font` normally
```
And possibly even better once we figure out our "unwrap" idiom.
But how would you do early exit with exceptions?
```cpp
// declare something here...
try
{
    const auto font = sf::Font::loadFromFile("MyFont.ttf");
}
catch (sf::LoadException& e)
{
    return EXIT_FAILURE;
}

// ...so that font can be used normally here.
```

You can't declare an `sf::Font` outside try-catch if it doesn't have a default constructor. So you'd still need to resort to something like optionals. 

Of course you could structure your code differently, but I see both
* not being able to early-exit
* enveloping a huge block of non-fallible code in `try`

as disadvantages that you don't have with `std::optional`.

Bromeon:
Btw, why the try-catch pattern works well in languages like C#, Java or Kotlin is because
a) references can be initialized to null (aka optional by accident), or 
b) static analysis allows to declare variables and initialize them later in all possible code paths, or
c) because they're expression-oriented (e.g. Kotlin), and try-catch blocks have a value that can be assigned to a variable

In C++, you have _none_ of the three. Instead, you need to work with old-school procedural control structures and extremely limited initialization rules.

eXpl0it3r:
See some of my counter arguments [here](https://github.com/SFML/SFML/issues/2139#issuecomment-1272597347) (where your general error handling comments should've been put as well 😉).

I basically expected to have people cling to the sub-optimal example (thus the note) and will provide a better example sometime later.

ChrisThrasher:
Rebased on `master`. Only made one minor change to fix a test.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

PR #2238, opened 2022-10-10
Hide examples, tests, and documentation from projects consuming SFML via `add_subdirectory`

ChrisThrasher:
## Description

Currently projects like the official CMake template inherit some SFML cache variables like `SFML_BUILD_TEST_SUITE`. Here's a screenshot from a project of mine that uses `FetchContent` to build the `master` branch.

Screen Shot 2022-10-09 at 10 41 07 PM

Luckily these options default off but downstream projects don't need to inherit these to build and use SFML. This PR makes it such that these developer-only options are hidden to those consuming SFML via `add_subdirectory`.

Note how we've already been using this pattern with the `format` target. It's perfectly valid for a downstream project to want to make their own target called `format` so the fact that we hide this target from them helps ensure we're not squatting on target names.

A few reasons why we benefit from doing this:
1. Less clutter in users' CMakeCache.txt files so users can more easily see the options they're most likely to change like disabling certain modules or enabling frameworks.
1. It lets us change the default values of options without worrying about users suddenly building more things by default. If we want to build tests by default for example, all `add_subdirectory` users would now be building the SFML unit tests and would then have to explicitly opt out of that to avoid doing that unnecessary work in their projects.
1. It's very useful to make the `add_subdirectory` interface match the find_package interface. [Recent `FetchContent` features](https://cmake.org/cmake/help/latest/module/FetchContent.html#integrating-with-find-package) added in CMake 3.24 mean it's only getting more common that users swap between find_package and add_subdirectory. Ensuring both have the most similar interfaces possible mean users never have issues when switching from one to the other.
1. We let ourselves use simpler option names. This is useful for the `SFML_BUILD_TEST_SUITE` option which is better replaced by the canonical [`BUILD_TESTING`](https://cmake.org/cmake/help/latest/module/CTest.html) option provided by the CTest module. If we hide tests from downstream users, we can safely use this preexisting option to enable or disable testing and maintain one less project-specific abstraction that all new contributors must learn about. We can't do that now because users with `BUILD_TESTING=ON` in their own projects would end up also enabling SFML's unit tests so we're forced to use a unique, nonstandard name.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,21 +53,11 @@ else()
     endif()
 endif()
 
-# add an option for building the examples
-if(NOT SFML_OS_ANDROID)
-    sfml_set_option(SFML_BUILD_EXAMPLES FALSE BOOL "TRUE to build the SFML examples, FALSE to ignore them")
-else()
-    set(SFML_BUILD_EXAMPLES FALSE)
-endif()
-
 # add options to select which modules to build
 sfml_set_option(SFML_BUILD_WINDOW TRUE BOOL "TRUE to build SFML's Window module. This setting is ignored, if the graphics module is built.")
 sfml_set_option(SFML_BUILD_GRAPHICS TRUE BOOL "TRUE to build SFML's Graphics module.")
 sfml_set_option(SFML_BUILD_AUDIO TRUE BOOL "TRUE to build SFML's Audio module.")
 sfml_set_option(SFML_BUILD_NETWORK TRUE BOOL "TRUE to build SFML's Network module.")
-
-# add an option for building the API documentation
-sfml_set_option(SFML_BUILD_DOC FALSE BOOL "TRUE to generate the API documentation, FALSE to ignore it")
 
 if(SFML_BUILD_WINDOW)
     # add an option for choosing the OpenGL implementation
@@ -78,12 +68,6 @@ if(SFML_BUILD_WINDOW)
         sfml_set_option(SFML_USE_DRM FALSE BOOL "TRUE to use DRM windowing backend")
     endif()
 endif()
-
-# add an option for building the test suite
-sfml_set_option(SFML_BUILD_TEST_SUITE FALSE BOOL "TRUE to build the SFML test suite, FALSE to ignore it")
-
-# add an option for enabling coverage reporting
-sfml_set_option(SFML_ENABLE_COVERAGE FALSE BOOL "TRUE to enable coverage reporting, FALSE to ignore it")
 
 # macOS specific options
 if(SFML_OS_MACOSX)
@@ -247,22 +231,6 @@ set_property(GLOBAL PROPERTY PREDEFINED_TARGETS_FOLDER "CMake")
 
 # add the subdirectories
 add_subdirectory(src/SFML)
-if(SFML_BUILD_EXAMPLES)
-    add_subdirectory(examples)
-endif()
-if(SFML_BUILD_DOC)
-    add_subdirectory(doc)
-endif()
-if(SFML_BUILD_TEST_SUITE)
-    if(SFML_OS_IOS)
-        message( WARNING "Unit testing not supported on iOS")
-    elseif(SFML_BUILD_WINDOW AND SFML_BUILD_GRAPHICS AND SFML_BUILD_NETWORK AND SFML_BUILD_AUDIO)
-        enable_testing()
-        add_subdirectory(test)
-    else()
-        message(WARNING "Cannot build unit testing unless all modules are enabled")
-    endif()
-endif()
 
 # on Linux and BSD-like OS, install pkg-config files by default
 set(SFML_INSTALL_PKGCONFIG_DEFAULT FALSE)
@@ -533,8 +501,45 @@ set(CPACK_NSIS_INSTALLER_MUI_ICON_CODE "!define MUI_WELCOMEFINISHPAGE_BITMAP \\\
 
 include(CPack)
 
-if(NOT CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
+# configure extras by default when building SFML directly, otherwise hide them
+if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
+    set(SFML_CONFIGURE_EXTRAS_DEFAULT TRUE)
+else()
+    set(SFML_CONFIGURE_EXTRAS_DEFAULT FALSE)
+endif()
+sfml_set_option(SFML_CONFIGURE_EXTRAS ${SFML_CONFIGURE_EXTRAS_DEFAULT} BOOL "TRUE to configure extras, FALSE to ignore them")
+
+if(NOT SFML_CONFIGURE_EXTRAS)
     return()
+endif()
+
+# add an option for building the API documentation
+sfml_set_option(SFML_BUILD_DOC FALSE BOOL "TRUE to generate the API documentation, FALSE to ignore it")
+if(SFML_BUILD_DOC)
+    add_subdirectory(doc)
+endif()
+
+# add an option for building the examples
+sfml_set_option(SFML_BUILD_EXAMPLES FALSE BOOL "TRUE to build the SFML examples, FALSE to ignore them")
+if(SFML_BUILD_EXAMPLES AND NOT SFML_OS_ANDROID)
+    add_subdirectory(examples)
+endif()
+
+# add an option for building the test suite
+sfml_set_option(SFML_BUILD_TEST_SUITE FALSE BOOL "TRUE to build the SFML test suite, FALSE to ignore it")
+
+# add an option for enabling coverage reporting
+sfml_set_option(SFML_ENABLE_COVERAGE FALSE BOOL "TRUE to enable coverage reporting, FALSE to ignore it")
+
+if(SFML_BUILD_TEST_SUITE)
+    if(SFML_OS_IOS)
+        message(WARNING "Unit testing not supported on iOS")
+    elseif(SFML_BUILD_WINDOW AND SFML_BUILD_GRAPHICS AND SFML_BUILD_NETWORK AND SFML_BUILD_AUDIO)
+        enable_testing()
+        add_subdirectory(test)
+    else()
+        message(WARNING "Cannot build unit testing unless all modules are enabled")
+    endif()
 endif()
 
 sfml_set_option(CLANG_FORMAT_EXECUTABLE clang-format STRING "Override clang-format executable, requires version 12, 13, or 14")


ChrisThrasher:

eXpl0it3r:
But that's not just hiding it, it's also preventing from building these things in the first place?

I get that in most cases one just wants to use the library when adding it as sub-dir, but I don't think preventing other uses cases just so we don't expose additional flags is a worthy trade-off.

ChrisThrasher:
> But that's not just hiding it, it's also preventing from building these things in the first place?
> 
> I get that in most cases one just wants to use the library when adding it as sub-dir, but I don't think preventing other uses cases just so we don't expose additional flags is a worthy trade-off.

Can you describe a use case for including SFML as a subdirectory but still wanting to build any of these accessory subdirectories?

JonnyPtn:
Yeah this is quite common for downstream projects to set these variables as they see fit so we shouldn't hide them, especially as there's no benefit to doing so

JonnyPtn:
Regardless of how rare you think it is (it's not that rare, I tend to set options on the majority of dependencies i pull in this way) completely disallowing it removes options for the end user

We are accountable for changes to these options regardless of whether users use this method or no

ChrisThrasher:
I modified the PR description to better articulate my technical arguments in favor of this change.

eXpl0it3r:
As said already privately the SFML CMake scripts aren't part of the SFML API. Sure people can build systems around it, but we won't be versioning our CMake script and ensuring that everything works all the time. Heck, we even broken the whole system when moving away from Find* modules to Config modules. Yes, it broke a lot of existing code, but you really don't want to restrict yourself on changes for the build system. It's not our duty to fix someone else's build system they build around the SFML build system, especially when they do more complex things.

As such, actively preventing people from using certain build options, just because it's done "programmatically" is bad reason. Building libraries with `add_subdirectory` on its own is already a bit of an edge case and SFML didn't (or doesn't?) properly support this for some time. Maybe with FetchContent this becomes a bit more "standard", but still if someone wants to build the SFML examples or the tests when including SFML, they should have the _option_ to do so. If they don't want these targets in their build system, then they have the _option_ to disabled them.

Your comparison with private headers only works as long as you view examples, tests, docs, formatting, etc. as "private" matters, which I personally don't.  
My counter example would be that we'd prevent using of sfml-system as a standalone library, because we deem it only as "private" library for SFML. You can only use sfml-system functions if you also use one of the other modules, which highlights the artificial restriction that doesn't really provide any gain to anyone.

Bromeon:
> [...] the SFML CMake scripts aren't part of the SFML API. Sure people can build systems around it, but we won't be versioning our CMake script and ensuring that everything works all the time. Heck, we even broken the whole system when moving away from Find* modules to Config modules. Yes, it broke a lot of existing code, but you really don't want to restrict yourself on changes for the build system. It's not our duty to fix someone else's build system that they built around the SFML build system, especially when they do more complex things.

I don't understand the reasoning "we give stability guarantees on the C++ code, but not on any infrastructure code around it". People are more and more shifting towards automated builds, GitHub Actions, etc.; so any breaking change in the build setup has the same effect as breaking C++ code. 

I know it's convenient for us to break things at will, and may be unavoidable at certain points, as CMake advances much faster than C++. However I don't agree with the general take "build config is not part of the public API". C++ dependency management is painful enough as it is; imagine someone spending all the manual effort to integrate SFML into their software only to rewrite everything a few months later. We should at least _try_ to keep it as stable as we can, and possibly document which parts exactly are stable and which aren't.

_(this solely referring to that point, not about the discussion in that PR)_

eXpl0it3r:
> We should at least _try_ to keep it as stable as we can

That's for me a given, but you can't say it's part of the API *and* try to keep it stable. It's either part of it and remains stable or you do best effort and then it's not part of the API.

And I argue that you can't make it part of the API, otherwise you back yourself in a corner where you can't update the build system, even if it's borken, because it would break the API.

As for the PR, if you remove options you just incentive people to hack their way into the SFML CMake file, apply patches and what not, is that what we want?  
I rather occasionally break something slightly somewhere, but at least provide all the options to those who want, than having people patch in their own fixes, just to use a "standard" build feature that was deemed too "private" to share with anyone else.

We've had the current options available for years and I've yet to come across someone that was really hurt by any change we did beyond the fundamental Find* -> *Config change. I'm certain people who build such a system are also smart enough to add a fix when something breaks.

Honestly you can argue all you want, my position on this won't change.  
But I could accept a compromise of hiding the options behind some "advance"/"might break" option, if you really deem it as such a problematic thing.

ChrisThrasher:
>  if you remove options you just incentive people to hack their way into the SFML CMake file, apply patches and what not, is that what we want?

Yes, I want to make it difficult for users to depend on such things. By making class implementation details `private:`, we're forcing users to use `#define private public` or apply patches to access them. It's good that it's not easy to access such things from downstream projects.

> We've had the current options available for years and I've yet to come across someone that was really hurt by any change we did beyond the fundamental Find* -> *Config change.

I wager this is because so few people are already depending on the SFML examples, tests, or docs, but I'm willing to admit I'm wrong if we merge this and get pushback from users who needed this functionality. Perhaps many of those users do exist at which point I'd support adding a `DEVELOPER_MODE` option that exposes these things to `add_subdirectory` users.

vittorioromeo:
@eXpl0it3r: I genuinely do not understand where you are coming from. 

Could you please help me understand why you think being able to build docs and examples as a subdirectory is valuable at all? 

I'm confident no one ever decided to manually run `cmake -DSFML_BUILD_EXAMPLES=1` on a project that was using SFML as a subdirectory (i.e. a *dependency*), even if they wanted to make a contribution directly to SFML.

It's a contrived workflow that we should not support. 

If you figure out something is wrong with SFML and want to contribute, you don't do it within your own project -- you clone the SFML repo in its own separate directory, create a branch, apply the changes there, build with tests and examples (which should be the default in a non-subdirectory build, IMHO), submit a PR, rinse and repeat.

At the very least, if you don't want to clone SFML again, you should just `cd` into the SFML subfolder in your project, and do your work there. Anything else is unreasonable, IMHO, and we should not support nor promote it.

JonnyPtn:
> I'm confident no one ever decided to manually run cmake -DSFML_BUILD_EXAMPLES=1 on a project that was using SFML as a subdirectory (i.e. a dependency),

Most of my recent contributions have come from a subdirectory

vittorioromeo:
> > I'm confident no one ever decided to manually run cmake -DSFML_BUILD_EXAMPLES=1 on a project that was using SFML as a subdirectory (i.e. a dependency),
> 
> Most of my recent contributions have come from a subdirectory

Just to understand your use case -- so you have a structure like

```
myproject/
myproject/CMakeLists.txt
myproject/SFML
myproject/SFML/CMakeLists.txt
```

From the `myproject` folder itself, you invoke `cmake` to build both your own project and your dependencies (including SFML). Then you make some changes to SFML, and again from the `myproject` folder you want to run `cmake -DSFML_BUILD_EXAMPLES=1' -- is that correct?

If so, wouldn't it make more sense to `cd` into SFML first, and do a local build of SFML there before preparing a PR? Wouldn't that allow you to keep SFML checked out as a subdirectory, but still invoke `cmake` on SFML as its own project and not as a dependency?

vittorioromeo:
> Typically I always configure SFML to build with examples anyway, so no that's incorrect

This is what I do not understand, why do you care about building SFML examples if you're working on a separate project? If SFML is a dependency of `myproject` and you are working on `myproject`, then why not treat SFML as a "black box" dependency?

> I wouldn't say it makes more sense - I've already got everything building and running at this point, why would it make more sense to configure, generate and build it all again?

Let me offer an alternative. Personally, I have multiple projects that depend on SFML, and in all of them I consume SFML as a dependency using a `FetchContent` based mechanism. 

Let's say projects A, B, and C all depend on SFML. And that from time to time, I contribute to SFML. It is much more easier and simpler for me to have an individual and separate clone of the SFML repo for contributing to SFML itself, separate from my projects. So my structure is:

```
workspace/A/extlibs/SFML
workspace/B/extlibs/SFML
workspace/C/extlibs/SFML
workspace/SFML
```

For A, B, and C, I mainly care about building those projects, not their dependencies. So I don't want to generate examples, documentation, or tests for the SFML dependency -- I just want to consume it.

For the top-level SFML folder, I care about contributions to SFML itself -- that's where I want to build examples, tests, and documentation.

I don't think it's worthwhile supporting the IMHO contrived workflow of working on SFML directly when it is being consumed as a dependency:

- If you want to work on SFML itself, have your own clone of it, and there you will see examples/tests/documentation options exposed (and they should be on by default, IMHO).

- If you want to use SFML as a dependency, you should never need to build examples/tests/documentation, so it should not be something we support.

vittorioromeo:
> How can you say it's not worthwhile to support other workflows when there is zero cost to supporting them, as we already do currently?

Well, there are costs. This PR and discussion itself is part of the cost. But more importantly, if we want to support this workflow explicitly, it should be documented and tested. And I don't think it's worthwhile doing so because it's a niche/unusual one.

JonnyPtn:
Think you might have it the wrong way round - nobody wants to explicitly support any particular workflow.

Currently the SFML configuration is (effectively) always the same, regardless of how users are using it. With this change it will explicitly change how SFML builds in certain circumstance, based on factors outside the SFML project. So this is not about explicitly supporting certain workflows, it's about explicitly limiting some. 

In actual fact this change is what would cause the extra work - we should add tests for this workflow and document the differences in behaviours from the normal workflow.

vittorioromeo:
My axioms/premises are:

A1. Most users that want to consume SFML as dependency do *not* want to contribute to SFML.

A2. Most users that build SFML as a subdirectory do *not* want to build SFML examples, tests and documentation.

A3. Most users that contribute to SFML will do that by building it from its *own* directory.

A4. All users that contribute to SFML will benefit by having tests and examples built by default.

A5. A very small number of users (possibly just you) contribute to SFML by building it as a subdirectory, from another top-level project, and manually enabling tests, examples, and documentation as they do so.

From those, it follows that:

B1. This PR will not affect the majority of SFML users or SFML contributors due to A1, A2, and A3.

B2. This PR will lay out the groundwork for A4, which will benefit SFML contributors.

B3. The tiny fraction of users mentioning in A5 will have to change their workflow to add a simple `cd` into the SFML folder. 

B4. If we don't want to merge this PR, we are implying that we want to support the A5 use case. If so, we should make it part of our documentation and embrace it as an official workflow supported by SFML.

@JonnyPtn: do you see my perspective, and/or can you point out any flaws in my reasoning?

My preferred course of action would be:

1. Merge this PR.

2. Create a new PR to build tests and examples by default when SFML is not a subdirectory (i.e. someone is working directly on SFML itself).

JonnyPtn:
No point continuing this, let's just see what the other opinions are

Bromeon:
The discussion around whether the workflow "have developer options, when using SFML as a subdirectory" is circling.
I'll try to summarize this thread.

**Reasons to disallow this workflow (motivation for the PR):**

1. You only need those configuration options as a contributor, and in that case you _should_ have SFML as a standalone directory.
2. We don't want to provide stability on non-public configuration options.
3. There is less clutter in user's CMake configuration without rarely needed options.

**Reasons to allow this workflow (counter-arguments to the PR):**

1. There are people actively using it at the moment, whose workflow would be broken.
2. It is a typical path to first use a library, and turning into a contributor when hitting limitations. This would be easier under that workflow.
3. It costs nothing to provide those options as long as there is no stability guarantee.

---

Since there is no consensus, what about a counter-proposal:

1. We **keep** those variables accessible for the time being.
1. We mark and document them explicitly as "unstable", making clear that anyone relying on them does so at their own risk.
   * Could e.g. be done with `SFML_UNSTABLE_*` naming or also just documentation.
1. To declutter, they could be marked as advanced in the CMake config.
1. When using in "developer mode" (SFML is not a subdirectory), examples and tests are enabled by default.

JonnyPtn:
Couple more technical notes on this:
- It will only apply if the user adds the SFML subdirectory at certain points in their script, after they've declared their project
- If the user defines these options in their own script they can still access these features

ChrisThrasher:

JonnyPtn:
What's the current state of this? Seems like using an option(s) is most sensible

ChrisThrasher:
Rebased to fix merge conflict.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:
Rebased to fix merge conflict.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:
In an attempt to reconcile the desires of this PR with #2239, I added a new option called `SFML_CONFIGURE_EXTRAS` which can be used by `add_subdirectory` users to enable all non-essential build targets and options. This means we still get the added peace of mind knowing that most `add_subdirectory` users will not be impacted by changes to these options but those who really want access to them still can do so.

JonnyPtn (changes requested):

CMakeLists.txt
```diff
@@ -533,10 +501,47 @@ set(CPACK_NSIS_INSTALLER_MUI_ICON_CODE "!define MUI_WELCOMEFINISHPAGE_BITMAP \\\
...
 include(CPack)
 
-if(NOT CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
+# configure extras by default when building SFML directly, otherwise hide them
+if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
+    set(SFML_CONFIGURE_EXTRAS_DEFAULT TRUE)
+else()
+    set(SFML_CONFIGURE_EXTRAS_DEFAULT FALSE)
+endif()
+sfml_set_option(SFML_CONFIGURE_EXTRAS ${SFML_CONFIGURE_EXTRAS_DEFAULT} BOOL "TRUE to configure extras, FALSE to ignore them")
```
JonnyPtn, line 510: Just default this to true - The expectation is that the configuration is the same whether it's a subdirectory or not, then users can decide which targets they want to enable

JonnyPtn (changes requested):

CMakeLists.txt
eXpl0it3r: The main idea behind this PR is to get `add_subdirectory` (and thus `FetchContent`) to behave as similarly as `find_package`. I can see merits and supporting these use cases more.  
As such `SFML_CONFIGURE_EXTRAS` is a good compromise to hide additional options in those scenarios, while still providing the possibility to access the extra options.

CMakeLists.txt
```diff
@@ -533,10 +501,47 @@ set(CPACK_NSIS_INSTALLER_MUI_ICON_CODE "!define MUI_WELCOMEFINISHPAGE_BITMAP \\\
...
+
+if(NOT SFML_CONFIGURE_EXTRAS)
```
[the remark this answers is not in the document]
eXpl0it3r: I also wouldn't mind to have some extra options available in either case, but I can also live with them being hidden.  
As I understood Thrasher and also Vee, they'd rather not expose extra options, as to reduce the default surface area when using it as `add_subdirectory` and `FetchContent`.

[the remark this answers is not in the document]
ChrisThrasher: I treat access to the extras as all or nothing. Moving this check down belies the point of this PR.

CMakeLists.txt
ChrisThrasher: I won't be defaulting it true because the point of this PR is to not expose such things to downstream projects by default. It's expected for users to not want such things and exceptional if they do.

CMakeLists.txt
```diff
@@ -533,10 +501,47 @@ set(CPACK_NSIS_INSTALLER_MUI_ICON_CODE "!define MUI_WELCOMEFINISHPAGE_BITMAP \\\
...
+        enable_testing()
+        add_subdirectory(test)
+    else()
+        message(WARNING "Cannot build unit testing unless all modules are enabled")
+    endif()
+endif()
+
 sfml_set_option(CLANG_FORMAT_EXECUTABLE clang-format STRING "Override clang-format executable, requires version 12, 13, or 14")
 add_custom_target(format
     COMMAND ${CMAKE_COMMAND} -DCLANG_FORMAT_EXECUTABLE=${CLANG_FORMAT_EXECUTABLE} -P ./cmake/Format.cmake
```
eXpl0it3r, line 547: Does this break by the way, if I don't have `clang-format` available?

eXpl0it3r (approved):
Assuming CPack doesn't somehow need the docs or example targets available before including it, this looks good to me.

Will have to be ready to explain the change to those who might have a broken setup now. 😄

CMakeLists.txt
ChrisThrasher: It’ll fail to run but it won’t fail to configure or otherwise break anything.

ChrisThrasher:
> Assuming CPack doesn't somehow need the docs or example targets available before including it, this looks good to me.

That's a good point. Running `make package` works for me locally. I read through the CPack code and saw no reference to any of these extras so I think we're good. I opened issue https://github.com/SFML/SFML/issues/2357 to make a note about how we ought to consider testing the packaging code.

CMakeLists.txt
JonnyPtn: Target shouldn't be created in cases where it fails to run

CMakeLists.txt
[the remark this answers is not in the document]
JonnyPtn: But it's not all or nothing because I some of them I can enable individually but some of them I can't?

PR #2239, opened 2022-10-10
Don't remove format target unnecessarily

JonnyPtn:
Noticed we're removing the format target, I think the intention is to prevent it being available when added as a subdirectory.

I don't see any advantage to doing this - Whether I've got SFML as a subdirectory or not doesn't really have any bearing on whether I want to format it, and if we're encouraging contributors to use it then it should be available at all times

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -528,13 +528,12 @@ set(CPACK_NSIS_INSTALLER_MUI_ICON_CODE "!define MUI_WELCOMEFINISHPAGE_BITMAP \\\
 
 include(CPack)
 
-if(NOT CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
-    return()
+sfml_set_option(SFML_FORMAT TRUE BOOL "Set to enable SFML's format target")
+if(SFML_FORMAT)
+    sfml_set_option(CLANG_FORMAT_EXECUTABLE clang-format STRING "Override clang-format executable, requires minimum version 12")
+    add_custom_target(format
+        COMMAND ${CMAKE_COMMAND} "-DCLANG_FORMAT_EXECUTABLE=${CLANG_FORMAT_EXECUTABLE}" -P ./cmake/Format.cmake
+        WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}" VERBATIM)
+
+    add_custom_target(tidy COMMAND run-clang-tidy -p ${PROJECT_BINARY_DIR} VERBATIM)
 endif()
-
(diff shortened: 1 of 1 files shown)


vittorioromeo:
What's the use case for having SFML as a subdirectory without it being a dependency of another project?

JonnyPtn:
Having it as a dependency of another project is enough of a use-case on it's own - if I notice a bug or something in my dependencies code that I want to change I'll often just change it in situ and push directly from the subdirectory, in which case I'd probably like to format my changes before doing that.

With the current implementation I'm forced to clone SFML separately, then either manually copy that code into my project to test my changes, or push it to my remote and modify my project to fetch that branch and test, neither of which are good workflows.

I could ask what is a use-case or benefit for preventing people from being able to format the SFML code?

eXpl0it3r:

ChrisThrasher:
This will immediately break projects of mine which define their own `format` target. By making these kinds of private details public, we're effectively expanding the API of SFML which means more private details that that we can freely change now must be API promises whose stability we promise over time. I'd much rather keep implementation details as implementation details.

JonnyPtn:
The solution to all your problems is to add a configuration option for it, not to remove it entirely

eXpl0it3r:
I guess we could add some `SFML_BUILD_FORMAT` or whatever is better fitting, so the consumers can decide whether they want it or not.

JonnyPtn:
Personally don't think a configuration option is worth it for this - users can easily pick different names and/or we can make ours more specific (e.g `sfml-format`) to avoid clashes

Not too fussed either way though so happy to go with consensus

eXpl0it3r:
Just thought it would allow for the flexibility of enabling/disabling and consequently including/hiding the target and be more inline with how other target visibility are handled (see my comment [here](https://github.com/SFML/SFML/pull/2238#issuecomment-1274686555)).

JonnyPtn:
Some projects I've seen have a more generic option like `SFML_BUILD_EXTRAS` to cover various things like formatting, analysis, etc so that's another one to consider

JonnyPtn:
What more is needed here? Do you want an option added?

JonnyPtn:
Have added an option for it - anything else required?

JonnyPtn:
Bump - also noticed we don't remove these targets in cases where we _know_ they don't work (e.g if the tools aren't available) so there's a precedent for not removing them already, and I know you all love consistency

ChrisThrasher:
As of the merging of https://github.com/SFML/SFML/pull/2238 the `format` target and all other build system extras are available to `add_subdirectory` users.

PR #2240, opened 2022-10-10
Rename test files to be unique

JonnyPtn:
Currently most of the test files have identical names to other files in the project which is generally considered bad practice. It makes the project harder to navigate and talk about, as well as running the risk of compilation errors with name clashes

Various options for naming the test files, I've gone for a common one that seems to make sense: if `X.cpp` contains the code for X then `XTests.cpp` contains the code for X's tests


eXpl0it3r:

eXpl0it3r (approved):
I'm used to this pattern as well and searches should still bring up the header, source and test file

ChrisThrasher:
> which is generally considered bad practice

I've never heard this.

> It makes the project harder to navigate and talk about

I think it makes the project easier to navigate since it takes zero brainpower to know what file contains the tests for a given header.

> as well as running the risk of compilation errors with name clashes

There is no way this is true. Even if we were globbing source files this would still not be true,

vittorioromeo:
I'm indifferent to this, but how about using the `.t.cpp` suffix instead of repeating `Tests` for each filename?

JonnyPtn:
@thr

ChrisThrasher:
How could the current name scheme cause compilation issues? We explicitly tell CMake what files to build. There is no ambiguity in the current naming convention.

kimci86:
For what it's worth, those two C++ project structure conventions suggest `.test.cpp` extension. I am not aware of other documented C++ project structure conventions.
- [The Pitchfork Layout](https://api.csswg.org/bikeshed/?force=1&url=https://raw.githubusercontent.com/vector-of-bool/pitchfork/develop/data/spec.bs#src.tests)
- [Canonical Project Structure](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html#tests-unit)

In addition, they suggest putting unit test files next to the files under test (referenced to as "merged test placement").

JonnyPtn:
> How could the current name scheme cause compilation issues? We explicitly tell CMake what files to build. There is no ambiguity in the current naming convention.

Because the output of compilation may go to the same directory

Bromeon (approved):
In my opinion, this is a nice change that helps with recognizability/findability, even if the overall impact may be low.

I would prefer @kimci86's suggestion of `X.test.cpp`, it gives visual space and preserves the class name as its own identifier. I would keep the test folder separate though.

There are some merge conflicts to fix.

ChrisThrasher:
> I would prefer @kimci86's suggestion of `X.test.cpp`

I would also rather add this extension than add a `Tests` suffix to the filename.

JonnyPtn:
Renamed and rebase

Bromeon:
All Debug CI jobs failed due to network error in Codecov, I restarted them.

eXpl0it3r:

JonnyPtn:
Bump

Bromeon (approved):
The `File.test.cpp` convention looks good to me.

eXpl0it3r:
Thanks for the small enhancement!

Issue #2053, opened 2022-03-28
Problems when using CJK IME on Linux


zenera:
## Subject of the issue

I could not input CJK text when using imgui-SFML on Ubuntu that uses ibus as a default IME. Showing the text was O.K. I've found the reason is behind the SFML source. X_HAVE_UTF8_STRING macro  is used in the SFML source(SFML/Window/Unix/WindowImplX11.cpp) but never defined. I've defined the macro and recompiled SFML libraries and now I can input CJK text. I want more official support for CJK characters. Is there any problem? 

## Your environment

* Linux / Ubuntu 21.10 / Gnome Desktop
* SFML 2.5.0 and git master + imgui git master + imgui-SFML git master
* gcc 11.2

## Steps to reproduce

I've tested with modified imgui-SFML minimal example.
https://github.com/eliasdaler/imgui-sfml/blob/master/examples/minimal/main.cpp

```cpp

#include <SFML/Graphics/CircleShape.hpp>
#include <SFML/Graphics/RenderWindow.hpp>
#include <SFML/System/Clock.hpp>
#include <SFML/Window/Event.hpp>

int main() {
    sf::RenderWindow window(sf::VideoMode(800, 600), "ImGui + SFML = <3");
    window.setFramerateLimit(60);
    if(!ImGui::SFML::Init(window)) return -1;
    auto io = ImGui::GetIO();
    io.Fonts->Clear();
    io.Fonts->AddFontFromFileTTF("/usr/share/fonts/truetype/nanum/NanumGothic.ttf", 16.0f,
                                  nullptr, io.Fonts->GetGlyphRangesKorean());
    if(!ImGui::SFML::UpdateFontTexture()) return -1;

    sf::CircleShape shape(100.f);
    shape.setFillColor(sf::Color::Green);

    sf::Clock deltaClock;
    while (window.isOpen()) {
        sf::Event event;
        while (window.pollEvent(event)) {
            ImGui::SFML::ProcessEvent(window, event);

            if (event.type == sf::Event::Closed) {
                window.close();
            }
        }

        ImGui::SFML::Update(window, deltaClock.restart());

        ImGui::ShowDemoWindow();

        ImGui::Begin("Hello, world!");
        ImGui::Button("Look at this pretty button 가나다");
        char buf[256];
        ImGui::InputText("한글 입력 테스트", buf, sizeof(buf));
        ImGui::End();

        window.clear();
        window.draw(shape);
        ImGui::SFML::Render(window);
[...]

PR #2242, opened 2022-10-11
Fix order and remove duplicates of KeyPressed events

### LINKED ISSUE #2191: Key Released event is either not properly ordered or missing when fps is limited
## Subject of the issue

Key Released event is either not properly ordered or missing when fps is limited, creating a "key stuck" effect.

## Your environment

* Your OS / distro / window manager used: 
  * Manajro KDE 5.18.5
* Your version of SFML (2.5.0, git master, etc): 
  * Tested on latest snapshot, but it also existed on snapshots from April. Very likely that it has existed for quite a while.
* Your compiler and compiler version used: 
  * GCC 12
* Special compiler flags used: 
```cmake
# relevant cmake commands used:
set (CMAKE_CXX_STANDARD 11)
#Find shared libraries in the same location as the executable
SET(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-rpath -Wl,$ORIGIN")
set(CMAKE_CXX_FLAGS "-fpermissive -std=c++0x")  #For surpressing goto warning
```
## Steps to reproduce
1. Limit the framerate, either using built-in set frame limit method or manually set the thread to sleep.
2. Hold any key for a few seconds and release, in this case I am using the "a" key.  

```cpp
#include <iostream> 
#include "SFML/Graphics.hpp"
#include "SFML/Window.hpp"
#include <thread>

int main()
{
    sf::RenderWindow window(sf::VideoMode(sf::Vector2u(500, 500)), "SFML works!");

    window.setFramerateLimit(30);

    bool pressed = false;
    while (window.isOpen())
    {
        sf::Event event;
        {
            if(event.type == sf::Event::KeyPressed)
            {
                std::cout<<("key pressed\n");

                if(event.key.code == sf::Keyboard::Key::A)
                    pressed = true;

                continue;
            }
            else if(event.type == sf::Event::KeyReleased)
            {
                std::cout<<("key released\n");

                if(event.key.code == sf::Keyboard::Key::A)
                    pressed = false;
                continue;
            }

                window.close();
        }

        if(pressed)
            window.clear(sf::Color(255, 0, 0, 255));
        else
            window.clear(sf::Color(255, 255, 255, 255));

        window.display();
        // std::this_thread::sleep_for(std::chrono::milliseconds(20));
    }

    return 0;
}
```

## Expected behavior

The release key event should always come after the key pressed event. 
With the example I have given, the background should always be white unless the "a" key is being held.

## Actual behavior

The release key event comes before the last key pressed event. 
With the example I have given, the background gets "stuck" with the red background occasionally.

Peek 2022-09-03 18-30

kimci86:
## Description

This is an attempt to fix broken keyboard events on X11 with IBus input method.

What this PR fixes (hopefully):
1. Avoid duplicated `KeyPressed` events.
   This is an alternative to #1894 from which I took inspiration, but more robust by indexing the bitset with X11 keycodes instead of `sf::Keyboard::Key` enumeration.

2. Fix events being in the wrong order, i.e. fixes #2191
   I noticed that getting as many events as possible from X11 queue and only then processing them (behavior added in #1291) makes the input method filtering go wrong, especially when framerate is low.
   Popping and processing events in order, with some contortions to handle key repeat events, seems to work just great.
   There is a risk of reintroducing bug #1223, but I do not reproduce that bug when testing this present PR.

What this PR does not fix, but improves a little bit:
- `TextEntered` events coming slowly when framerate is low.
  As far as I understand, when IBus input method is used and a key is pressed, SFML pops a `KeyPress` event from X11 events queue and notifies the input method about it through `XFilterEvent`. The input methods gives an answer with text after a tiny delay through another X11 `KeyPress` event, so it is available for SFML on the next frame. This delay is noticeable when framerate is low.
   Before this PR, when holding a key down with a low framerate, I had one `TextEntered` event every two frames. Somehow with this PR, I get one `TextEntered` each frame. So the problem becomes less noticeable.
   A workaround to avoid this delay could be to process events in a separate thread but that would add complexity and I am not sure X11 would not explode.

## Tasks

* [x] Tested on Linux
* [ ] Get more people to test @Edgaru089 @Neko-Box-Coder

### Test program

```c++
#include <iostream>
#include <SFML/Graphics.hpp>

{
    // usage: test <framerate> <keyRepeat>
    const int framerate = 1 < argc ? std::atoi(argv[1]) : 60;
    const bool keyRepeat = 2 < argc ? std::atoi(argv[2]) : true;

    sf::Window window(sf::VideoMode(100, 100), "Window");

    auto createWindow = [&] {
        window.create(sf::VideoMode(100, 100), "Window");
        window.setFramerateLimit(framerate);
        window.setKeyRepeatEnabled(keyRepeat);
    };
    createWindow();

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
            {
                std::cout << "Key Pressed: " << event.key.code << std::endl;
                if (event.key.code == sf::Keyboard::Key::Space)
                    createWindow();
            }
            else if (event.type == sf::Event::KeyReleased)
                std::cout << "Key Released: " << event.key.code << std::endl;
            else if (event.type == sf::Event::TextEntered)
            {
                const auto utf8 = sf::String(event.text.unicode).toUtf8();
                std::cout << "Text Entered: " << std::string(utf8.begin(), utf8.end()) << std::endl;
            }
        }

        window.display();
    }
}
```

### Building the test program

Save the `main.cpp` file above in a directory along with the following `CMakeLists.txt` file:
```cmake
cmake_minimum_required(VERSION 3.16)
project(test)

include(FetchContent)
FetchContent_Declare(SFML
    GIT_TAG bugfix/x11_key_events)
if(NOT sfml_POPULATED)
  FetchContent_Populate(SFML)
  add_subdirectory(${sfml_SOURCE_DIR} ${sfml_BINARY_DIR} EXCLUDE_FROM_ALL)
endif()

add_executable(test main.cpp)
target_link_libraries(test PRIVATE sfml-window)
target_compile_features(test PRIVATE cxx_std_17)
```

Then run those commands from the directory the files are saved in:
```sh
cmake -S . -B build -DCMAKE_BUILD_TYPE=Debug
cmake --build build
```

### Running the test program

From the same directory:
```sh
# by default, running at 60 fps with key repeat enabled
build/test
# running at 20 fps with key repeat enabled
build/test 20
# running at 20 fps with key repeat disabled
build/test 20 0
```

### Testing steps

Using the test program, with different framerates (first command line argument):
- Check that a single `KeyPressed` event is generated when pressing a key for a short time.
- Check that a single `KeyReleased` event is generated and comes after all the `KeyPressed` events when pressing a key for a long time.
- With key repetition disabled (second command line argument set to 0), check that a long press generates the same events as a short press.
  + Check that it behaves the same after recreating the window by pressing the space bar.


eXpl0it3r:

kimci86:
@zenera You might also be interested in testing this PR as it is probably related to #2053.

Neko-Box-Coder:
I will give it a go this weekend if free if I can manage to figure how to build sfml again :sweat_smile:

kimci86:
@Neko-Box-Coder I updated the first message with more instructions to make it easier.

Neko-Box-Coder:
Thanks, that will help quite a lot. :+1:

Neko-Box-Coder:
Hi :wave: @kimci86,

Thanks a lot for the CMake script and test program. It worked flawlessly.
I am using fcitx5,
I ran your test and also my demo just to make sure and the key pressed and released order is working good now.

The only thing is just TextEntered is coming after KeyReleased.
Peek 2022-10-16 17-18

But as you said this is probably the Xorg problem so I guess can't really do much bout it without adding a lot of complexity.
Also for most of the use case Text and key input is processed separately so it probably won't be too much of a problem.

Thanks a lot for the fix

eXpl0it3r:

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -832,14 +819,63 @@ void WindowImplX11::processEvents()
...
+        // Detect repeated key events
+        while (event.type == KeyRelease)
+        {
+            XEvent nextEvent;
+            if (XCheckIfEvent(m_display, &nextEvent, checkEvent, reinterpret_cast<XPointer>(m_window)))
+            {
+                if (nextEvent.type == KeyPress &&
+                    nextEvent.xkey.keycode == event.xkey.keycode &&
+                    event.xkey.time <= nextEvent.xkey.time &&
+                    nextEvent.xkey.time <= event.xkey.time + 1)
```
eXpl0it3r, line 842: Can you not assume the same time?

kimci86: Maybe. I just used the same condition as before which was `(event.xkey.time - time < 2)`, but rewrote it to be more explicit, i.e. not rely on unsigned integer wrapping. Involved variables are of type `Time` which is an Xlib typedef to `unsigned int`.

kimci86:
Did some formatting and rebased.

eXpl0it3r (approved):

eXpl0it3r:
Any other brave testers before we merge this? 🙂 

Maybe @Edgaru089 or @zenera?

eXpl0it3r:

eXpl0it3r:
Kind of unsurprising, but I couldn't reproduce the issue via WSL2 on Windows. I guess Windows handling events and passing them to the XServer somehow prevents the issue from occurring, but the new implementation works as well and the implementation makes sense.

PR #2245, opened 2022-10-13
[DRM] Fix forecefully switching to the preferred mode, use the current mode rather

substring:

----

## Description

In order to prevent an unnecessary modeswitch and use the current TTY resolution, don't use the preferred mode reported by DRM. This is useful for people who set a custom resolution at boot and are forced into the monitor's preferred resolution when a SFML DRM app starts

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Make sure your TTY is not set at your monitor's preferred resolution and start any app

### DIFF
--- a/extlibs/headers/drm/drm-common.c
+++ b/extlibs/headers/drm/drm-common.c
@@ -285,34 +285,6 @@ int init_drm(struct drm *drm, const char *device, const char *mode_str,
 			printf("requested mode not found, using default mode!\n");
 	}
 
-	/* find preferred mode or the highest resolution mode: */
-	if (!drm->mode)
-	{
-		for (i = 0, area = 0; i < connector->count_modes; i++)
-		{
-			drmModeModeInfo *current_mode = &connector->modes[i];
-
-			if (current_mode->type & DRM_MODE_TYPE_PREFERRED)
-			{
-				drm->mode = current_mode;
-				break;
-			}
-
-			int current_area = current_mode->hdisplay * current_mode->vdisplay;
-			if (current_area > area)
-			{
-				drm->mode = current_mode;
-				area = current_area;
-			}
-		}
-	}
-
-	if (!drm->mode)
-	{
-		printf("could not find mode!\n");
-		return -1;
-	}
-
 	/* find encoder: */
 	for (i = 0; i < resources->count_encoders; i++)
 	{
@@ -349,7 +321,17 @@ int init_drm(struct drm *drm, const char *device, const char *mode_str,
     // get original display mode so we can restore display mode after program exits
     drm->original_crtc = drmModeGetCrtc(drm->fd, drm->crtc_id);
 
-	if(getenv("SFML_DRM_DEBUG"))
+	/* Let's use the current mode rather than the preferred one if the user didn't
+	 * specify a mode with env vars
+	 */
+	if (!drm->mode)
+	{
+		if(getenv("SFML_DRM_DEBUG"))
+			printf("DRM using the current mode\n");
+		drm->mode = &(drm->original_crtc->mode);
+	}
+
+	if (getenv("SFML_DRM_DEBUG"))
 	{
 		printf("DRM Mode used: %s@%d\n", drm->mode->name, drm->mode->vrefresh);
 	}
--- a/src/SFML/Window/DRM/DRMContext.cpp
+++ b/src/SFML/Window/DRM/DRMContext.cpp
@@ -94,14 +94,18 @@ namespace
         if (!initialized)
             return;
 
-        drmModeSetCrtc(drmNode.fd,
-                       drmNode.original_crtc->crtc_id,
-                       drmNode.original_crtc->buffer_id,
-                       drmNode.original_crtc->x,
-                       drmNode.original_crtc->y,
-                       &drmNode.connector_id,
-                       1,
-                       &drmNode.original_crtc->mode);
+        /* Avoid a modeswitch if possible */
+        if (drmNode.mode != &drmNode.original_crtc->mode)
+            drmModeSetCrtc(drmNode.fd,
+                           drmNode.original_crtc->crtc_id,
+                           drmNode.original_crtc->buffer_id,
+                           drmNode.original_crtc->x,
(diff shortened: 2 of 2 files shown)


oomek (approved):

eXpl0it3r:

extlibs/headers/drm/drm-common.c
```diff
@@ -285,34 +285,6 @@ int init_drm(struct drm *drm, const char *device, const char *mode_str,
...
-			}
-		}
-	}
-
-	if (!drm->mode)
-	{
-		printf("could not find mode!\n");
-		return -1;
-	}
-
```
eXpl0it3r, line 315: We was under the impression that this is essentially a library. If we have to adjust this on our own, I feel like the functionality should not be part of a file in the extlib folder, but instead be part of SFML itself. 🤔

oomek: It's technically just a set of common functions written by Rob Clark, taken from kmscube example. I doubt it's still actively maintained. Btw, I thought we settled on it sitting inside the extlibs folder. Does the fact of it sitting there prevent us from improving it?

ChrisThrasher: I remember bringing this up. If we are to add 3rd party sources to the src/ directory then we need to (1) make sure the licensing allows that, (2) wholly change the code to match SFML conventions and standards, and (3) be willing to maintain that code for perpetuity. I don't think this code checked off all 3 requirements so it got relegated to extlibs/. The assumption then becomes that because we put the code in extlibs/, we're never going to modify it since it's not our code and only lives in our repo because we couldn't be bothered to consume it from some external source. 

We're now in this ugly middle ground where we have 3rd party that we don't actually want to maintain but do need to modify and I'm not sure how to square those two facts.

oomek: Me neither, any volunteers to SFML it ;) ?

ChrisThrasher: Whoever is most invested in the DRM backend ought to volunteer and SFML-itize this code. Happy to help but I can't lead that effort. At a minimum, let's make sure there are no licensing concerns. @oomek can you look into the original license of this code lets us totally take and rewrite it?

oomek: https://github.com/freedesktop/mesa-kmscube/blob/master/common.c

As long as the header is intact we can modify it.

ChrisThrasher: The copyright header? Or the C++ header?

oomek: Copyright header

oomek: Here is the copyright note from the original repo, the github one is just a mirror
https://gitlab.freedesktop.org/mesa/kmscube/-/blob/master/COPYING

oomek: We can store both files inside src, up to you

ChrisThrasher: As in, does the 3rd party DRM header _need_ to be part of SFML's public interface or is merely an implementation detail?

oomek: Not sure if it's required to add it to public interface it's referenced just by
`src/SFML/Window/DRM/VideoModeImpl.cpp`
`src/SFML/Window/DRM/WindowImplDRM.cpp`
`src/SFML/Window/DRM/DRMContext.hpp`

What if we take it over, host on our repo, and maintain it as an extlib? From what I read the licence allows it.

ChrisThrasher: It sounds like SFML-itizing these 2 files and keeping them as implementation details in src/ is viable. I'm curious what the rest of the team thinks about this though.

eXpl0it3r: I was under the impression that it's a file/library that "everyone is using", and thus thought it was unmodified or only slightly wrapped. If we can't use it as quasi library, then we should move the functionality into our own classes and structures, meaning the goal shouldn't necessarily be to have a `src/SFML/Window/DRM/Common.cpp` but instead integrate the functionality in the existing Impl and Context classes.

oomek: Whatever makes you happy. Shouldn't we delegate this task to a separate PR then?

ChrisThrasher: > I was under the impression that it's a file/library that "everyone is using", and thus thought it was unmodified or only slightly wrapped.

This was my impression too.

> Shouldn't we delegate this task to a separate PR then?

I think this reorganization should be its own PR before we start modifying the code from how we originally got it, but I still want some more of the team to weigh in on this.

ChrisThrasher: https://github.com/ChrisThrasher/SFML/actions/runs/3267610846/jobs/5373007062

The existence of 3rd party source files in our build is getting in the way of using clang-tidy. I could probably figure out a workaround but would prefer not to. Add that to the list of reasons we ought to integrate this file into SFML itself.

@oomek Can you try integrating these files into SFML itself in a separate branch not including any of this PR's changes? At a minimum, move the files to an appropriate place in src/ and modify the build system to treat them as 1st party files. Presumably that won't be too hard and then we can go from there determining whether we want to start heavily modifying the files to look like 1st party SFML code.

oomek: @ChrisThrasher This commit targets 2.6.X Can we open a separate issue for this please? It's getting really messy here.

ChrisThrasher:
https://github.com/SFML/SFML/issues/2252

I made an issue for discussing moving these DRM files into SFML. Please continue any relevant conversation in there.

eXpl0it3r:

eXpl0it3r:
If pushed two small style fixes

eXpl0it3r (approved):

eXpl0it3r:
Thanks for this fix! 🙂

PR #2246, opened 2022-10-13
Fix DRM using the first possible video card without even checking if a monitor is connected

substring:

----

## Description

So far, SFML was just checking for DRM capable GFX cards without considering plugged monitors. On systems with multiple GFX cards where all monitors are connected to 1 card, SFML could pick the wrong card. To fix that, check if card connectors are connected to select/skip that card

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?
You need several GFX cards, monitor(s) plugged to only one GFX card.

### DIFF
--- a/extlibs/headers/drm/drm-common.c
+++ b/extlibs/headers/drm/drm-common.c
@@ -170,6 +170,26 @@ static int get_resources(int fd, drmModeRes **resources)
 	return 0;
 }
 
+static int has_monitor_connected(int fd, drmModeRes* resources)
+{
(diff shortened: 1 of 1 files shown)


eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:
Thank you for adding this additional check! 🙂

PR #2249, opened 2022-10-16
Use doctest-provided `main` implementation

ChrisThrasher:
## Description

We must have overlooked this `doctest::doctest_with_main` target when replacing the vendored doctest header with the current FetchContent scheme. I'm happy to use this rather than maintain our own equivalent.


eXpl0it3r (approved):

eXpl0it3r:

PR #2251, opened 2022-10-17
Add clang-tidy integration

ChrisThrasher:
## Description

With most of the easy unit testing out of the way, this is the next step in SFML's journey towards better software engineering practices. For the uninitiated, read about clang-tidy [here](https://clang.llvm.org/extra/clang-tidy/).

This PR does 2 things, 1 in each commit.

1. The first commit prepares the way for clang-tidy by ensuring that Doctest's source files do not appear in the compile commands database and that it's public headers are correctly listed as `SYSTEM` headers. I'm merely mimicking how Doctest would appear if we consumed it via `find_package`.
1. The second commit adds a new custom target named `tidy` which invokes [`run-clang-tidy`](https://github.com/llvm-mirror/clang-tools-extra/blob/master/clang-tidy/tool/run-clang-tidy.py), a Python script from the LLVM project which automatically runs clang-tidy in parallel across all source files in the compile commands database. This behavior is critical because different configurations (often split on OS boundaries) compile different source files. If you try to analyze an X11 source file on macOS, you won't get meaningful results since clang-tidy won't know what settings are required to build such a file. Vice versa when trying to analyze a Objective-C++ in Ubuntu.

I chose to only run clang-tidy twice since I have no reason to believe that testing it with debug/release/shared/static will ever yield different result. Given that choice, it made the most sense to make separate jobs for running it, mimicking the pattern already established for building the `format` target.

You may notice that I actually fixed no clang-tidy findings in that PR and that's intentional. I chose to ignore all the warnings that were cropping up so that this PR could focus on only the infrastructure to run clang-tidy. It will require its own separate discussion to address the things clang-tidy is actually finding. There are many `clang-analyze-*` checks to enable and many more families of checks I want to start enabling.

I do specifically want to mention the [`readability-identifier-naming`](https://releases.llvm.org/11.1.0/tools/clang/tools/extra/docs/clang-tidy/checks/readability-identifier-naming.html) which can let us enforce things like UpperCase class names and lowerCamelCase variable and function names. These are probably the checks I'm most eager to enable.

_"But what about Windows?"_ One of the requirements to using clang-tidy is having a compile commands database. This is a file you may have seen in your build directory called compile_commands.json. CMake will generate it (when using Makefiles or Ninja) with the `-DCMAKE_EXPORT_COMPILE_COMMANDS` variable. Unfortunately MS Build will not generate such a file for us so I'm not aware of any way to run clang-tidy on the Windows-only source files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,3 +134,34 @@ jobs:
 
     - name: Check Formatting
       run: git diff --exit-code
+
+  tidy:
+    name: Analyzing on ${{ matrix.platform.name }}
+    runs-on: ${{ matrix.platform.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+        - { name: Linux, os: ubuntu-latest }
+        - { name: macOS, os: macos-12 }
+
+    steps:
+    - name: Checkout Code
+      uses: actions/checkout@v3
+
+    - name: Install Linux Dependencies
+      if: runner.os == 'Linux'
+      run: sudo apt-get update && sudo apt-get install libxrandr-dev libxcursor-dev libudev-dev libopenal-dev libflac-dev libvorbis-dev libgl1-mesa-dev libegl1-mesa-dev
+
+    - name: Install macOS Dependencies
+      if: runner.os == 'macOS'
+      run: |
+        brew install llvm
+        echo /usr/local/opt/llvm/bin >> $GITHUB_PATH
+
+    - name: Configure
+      run: cmake -S $GITHUB_WORKSPACE -B $GITHUB_WORKSPACE/build -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DSFML_BUILD_EXAMPLES=TRUE -DSFML_BUILD_TEST_SUITE=TRUE
+
+    - name: Analyze Code
+      run: cmake --build $GITHUB_WORKSPACE/build --target tidy
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -2,12 +2,16 @@ include(FetchContent)
 
 set(DOCTEST_NO_INSTALL ON)
 FetchContent_Declare(doctest
-    GIT_REPOSITORY "https://github.com/doctest/doctest.git"
+    GIT_REPOSITORY https://github.com/doctest/doctest.git
     GIT_TAG v2.4.9
 )
 FetchContent_MakeAvailable(doctest)
-list(APPEND CMAKE_MODULE_PATH ${doctest_SOURCE_DIR}/scripts/cmake)
-include(doctest)
+include(${doctest_SOURCE_DIR}/scripts/cmake/doctest.cmake)
+
+# Ensure that doctest sources and headers are not analyzed by any tools
+set_target_properties(doctest_with_main PROPERTIES COMPILE_OPTIONS "" EXPORT_COMPILE_COMMANDS OFF)
+get_target_property(DOCTEST_INCLUDE_DIRS doctest INTERFACE_INCLUDE_DIRECTORIES)
+target_include_directories(doctest SYSTEM INTERFACE ${DOCTEST_INCLUDE_DIRS})
 
 add_subdirectory(install)
 
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,14 @@
+Checks: >
+  -*,
+  clang-analyzer-*,
+  -clang-analyzer-core.NonNullParamChecker,
+  -clang-analyzer-core.NullDereference,
+  -clang-analyzer-nullability.NullablePassedToNonnull,
+  -clang-analyzer-optin.cplusplus.VirtualCall,
+  -clang-analyzer-optin.osx.*,
+  -clang-analyzer-optin.portability.UnixAPI,
+  -clang-analyzer-osx.*,
+  -clang-analyzer-security.insecureAPI.rand,
+  -clang-analyzer-unix.Malloc,
+HeaderFilterRegex: '.*'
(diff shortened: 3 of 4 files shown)


ChrisThrasher:

ChrisThrasher:

JonnyPtn:
Looks good, except I wouldn't put the doctest fetchcontent inside a find module - it will be in the global scope so will affect any downstream projects, and I don't believe that's worth it just on the off-chance it saves us changing a few lines in future

ChrisThrasher:

ChrisThrasher:

JonnyPtn (approved):

.github/workflows/ci.yml
```diff
@@ -134,3 +134,34 @@ jobs:
...
+        - { name: Linux, os: ubuntu-latest }
+        - { name: macOS, os: macos-12 }
+
+    steps:
+    - name: Checkout Code
+      uses: actions/checkout@v3
+
+    - name: Install Linux Dependencies
+      if: runner.os == 'Linux'
+      run: sudo apt-get update && sudo apt-get install libxrandr-dev libxcursor-dev libudev-dev libopenal-dev libflac-dev libvorbis-dev libgl1-mesa-dev libegl1-mesa-dev libdrm-dev libgbm-dev
```
kimci86, line 155: Installing `libdrm-dev` and `libgbm-dev` is not needed as long as option `SFML_USE_DRM` is false.

test/CMakeLists.txt
```diff
@@ -2,12 +2,14 @@ include(FetchContent)
...
-    GIT_REPOSITORY "https://github.com/doctest/doctest.git"
+    GIT_REPOSITORY https://github.com/doctest/doctest.git
     GIT_TAG v2.4.9
 )
 FetchContent_MakeAvailable(doctest)
-list(APPEND CMAKE_MODULE_PATH ${doctest_SOURCE_DIR}/scripts/cmake)
-include(doctest)
+set_target_properties(doctest_with_main PROPERTIES COMPILE_OPTIONS "" EXPORT_COMPILE_COMMANDS OFF)
+get_target_property(DOCTEST_INCLUDE_DIRS doctest INTERFACE_INCLUDE_DIRECTORIES)
+target_include_directories(doctest SYSTEM INTERFACE ${DOCTEST_INCLUDE_DIRS})
```
kimci86, line 11: A small comment explaining why those extra step are needed would be great.

.github/workflows/ci.yml
ChrisThrasher: That makes me wonder if I should go ahead and add clang-tidy + DRM jobs since currently those source files are not getting analyzed.

ChrisThrasher:
https://github.com/ChrisThrasher/SFML/actions/runs/3267610846/jobs/5373007062

It's not hard to start analyze the DRM build as well but it's currently reporting errors in the 3rd party drm-common.c file which we compile into sfml-window. [Discussions](https://github.com/SFML/SFML/pull/2245#discussion_r994829633) are ongoing about whether to convert that into a 1st party source file which would give us the ability to rewrite it in C++ and address these clang-tidy errors. For now I won't add a DRM clang-tidy job to this PR but I hope to add it in the future.

test/CMakeLists.txt
ChrisThrasher: I added a few comments for this since it is rather unintuitive.

ChrisThrasher:

kimci86 (approved):

Bambo-Borris (approved):

.github/workflows/ci.yml
```diff
@@ -134,3 +134,34 @@ jobs:
...
+  tidy:
+    name: Analyzing on ${{ matrix.platform.name }}
+    runs-on: ${{ matrix.platform.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+        - { name: Linux, os: ubuntu-latest }
+        - { name: macOS, os: macos-12 }
```
Bromeon, line 147: Is there reason to assume that clang-tidy (a linter) will yield different results on different platforms?

Just trying to keep the number of overall CI jobs in reasonable bounds (this also applies to `format` etc).

Bromeon (commented):
Thanks for adding this! You said you don't want to enable clang-tidy yet -- is that the case? Job setup doesn't seem to have anything special (`fail-fast` is unrelated), is it in the command itself?

The tough part with linters/inspections is usually to find a good balance between automation and annoyance/bikeshedding for non-issues, so the configuration will be interesting 🙂

.github/workflows/ci.yml
eXpl0it3r: You have different code paths available on Linux vs macOS (vs Windows), so there will be differences

ChrisThrasher:
> You said you don't want to enable clang-tidy yet -- is that the case?

clang-tidy is already being run, it's just that I disabled all the checks that were causing warnings to appear. We will enable and fix those checks in later PRs.

> Job setup doesn't seem to have anything special (`fail-fast` is unrelated), is it in the command itself?

`cmake --build build --target tidy` is where the magic is happening. That builds the `tidy` target which runs the run-clang-tidy script which invokes clang-tidy in parallel across all the source files built for the given configuration. As will all good software engineering there's an abundance of indirections but the `tidy` target is the entry point to all this.

> The tough part with linters/inspections is usually to find a good balance between automation and annoyance/bikeshedding for non-issues, so the configuration will be interesting 🙂

This is a non-trivial discussion, hence me postponing this to later PRs once the infrastructure is in place.

.github/workflows/ci.yml
ChrisThrasher: That's right, different OSes compile different source files which means different source files get analyzed.

Bromeon: Makes sense, thanks!

Bromeon (approved):

eXpl0it3r (approved):

PR #2254, opened 2022-10-18
Add 32-bits Visual Studio builds to the CI

eXpl0it3r:
## Description

I came to realize that we aren't building 32-bits targets (anymore?), which could potentially hide some breaking changes, especially also w.r.t. type conversions.  
It may not be the most recommended architecture these days, but it can still have relevance and doesn't really hurt us to support at the moment.

Adding it to the configuration of course another eight builds, so I'm also open for alternative suggestions.

## How to test this PR?

Run the GitHub Actions CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,8 +11,10 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-        - { name: Windows VS2019,       os: windows-2019 }
-        - { name: Windows VS2022,       os: windows-2022 }
+        - { name: Windows VS2019 x86,   os: windows-2019, flags: -A Win32 }
+        - { name: Windows VS2019 x64,   os: windows-2019, flags: -A x64 }
+        - { name: Windows VS2022 x86,   os: windows-2022, flags: -A Win32 }
+        - { name: Windows VS2022 x64,   os: windows-2022, flags: -A x64 }
         - { name: Windows VS2022 Clang, os: windows-2022, flags: -T ClangCL }
         - { name: Linux GCC,            os: ubuntu-latest }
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -11,8 +11,10 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-        - { name: Windows VS2019,       os: windows-2019 }
-        - { name: Windows VS2022,       os: windows-2022 }
+        - { name: Windows VS2019 x86,   os: windows-2019, flags: -A Win32 }
+        - { name: Windows VS2019 x64,   os: windows-2019, flags: -A x64 }
+        - { name: Windows VS2022 x86,   os: windows-2022, flags: -A Win32 }
+        - { name: Windows VS2022 x64,   os: windows-2022, flags: -A x64 }
         - { name: Windows VS2022 Clang, os: windows-2022, flags: -T ClangCL }
```
Bromeon, line 18: Tiny nitpick, should the `flags` be aligned with the one below for _Linux Clang_?

Btw, YAML is really special in what it treats as strings without quotes, but if it works... 😉

eXpl0it3r: No because that one isn't aligned with the MacOS Xcode either and don't think, aligning things with "empty" lines in between makes that much sense. With auto formatters I start to not care as much about these alignments 😄

ChrisThrasher:
Does this PR imply that SFML 3 may ship with x86 Windows binaries? If so, are there any potential issues we may run into supporting this config for the lifetime of SFML 3?

ChrisThrasher (approved):
I suppose as long as we have these 32 bit Windows binaries and all the associated code paths, it makes sense to add these CI jobs. I hope we consider removing these 32 bit binaries however. Not only would that get us one step closer to our ideal binary-free repository, it would also let us eliminate some code paths that I wager are very seldom exercised.

.github/workflows/ci.yml
Bromeon: It seems weird that we pay attention to align the 2nd column, but not the third 😅 but it's not important to bikeshed further, so let's go with the changes!

Bromeon (approved):

PR #2258, opened 2022-10-23
Enforce parameter case

ChrisThrasher:
## Description

The clang-tidy PRs keep coming. This time I added a check for the style of parameters. The style in a few places were wrong. One fix was straightforward but the changes to the vector types were a bit more complicated.

Identifiers like `X` and `Y` don't adhere to our `camelBack` style. I explored a few options. First I tried using a leading underscore but that didn't match the `camelBack` style. Trailing underscores didn't fit it either. I briefly considered using a `the` prefix like in [`sf::Vertex::Vertex`](https://github.com/SFML/SFML/blob/master/include/SFML/Graphics/Vertex.hpp#L58) but that just looked bad. Next I tried adding `NOLINT(readability-identifier-naming)` comments on the offending code but that also felt wrong. I didn't like how I had to add this comment in Vector2.hpp _and_ Vector2.inl. Plus that fix is overkill. That would disable all identifier naming checks for that line, not just those about the parameter names.

I eventually settled on using `x` and `y` style identifiers. I liked how this looked the best and didn't require adding extra comments to all the headers. I only had to temporarily disable `-Wshadow` in the .inl files. Curiously Clang didn't emit `-Wshadow` warnings. It was only GCC which took issue.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -14,5 +14,6 @@ Checks: >
 CheckOptions:
   - { key: readability-identifier-naming.ClassCase,           value: CamelCase }
   - { key: readability-identifier-naming.FunctionCase,        value: camelBack }
+  - { key: readability-identifier-naming.ParameterCase,       value: camelBack }
 HeaderFilterRegex: '.*'
 WarningsAsErrors: '*'
--- a/include/SFML/Graphics/Glsl.inl
+++ b/include/SFML/Graphics/Glsl.inl
@@ -102,15 +102,18 @@ struct Vector4
     ////////////////////////////////////////////////////////////
     /// \brief Construct from 4 vector components
     ///
-    /// \param X Component of the 4D vector
-    /// \param Y Component of the 4D vector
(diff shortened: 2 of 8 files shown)


Bromeon (approved):
It's weird to me that g++ complains with a shadow warning, as this is a very widespread pattern for constructor initialization, but it is what it is 🙂

ChrisThrasher:
> It's weird to me that g++ complains with a shadow warning, as this is a very widespread pattern for constructor initialization, but it is what it is 🙂

Yeah it's a bit annoying. I want to remove `the` from other constructors but am not eager to add 3 more lines of pragmas for each.

Bromeon:
See also: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78147
6 years, the reporter even submitted a patch, 0 progress...

ChrisThrasher:

Bambo-Borris (approved):

Issue #2252, opened 2022-10-17
Incorporate 3rd party drm-common.[h|c] files into SFML


ChrisThrasher:
## Subject of the issue

While discussing #2245 it came to our attention how it was odd that 3rd party DRM code needed our attention. Vendoring 3rd party code which we never modify is acceptable as is maintaining 1st party code over which we have full control, but 3rd party code which we must modify is an awkward middle ground.

I propose we integrate [drm-common.h](https://github.com/SFML/SFML/blob/master/extlibs/headers/drm/drm-common.h) and [drm-common.c](https://github.com/SFML/SFML/blob/master/extlibs/headers/drm/drm-common.c) into SFML itself. This means a few things which can be broken apart into multiple commits/PRs as we see fit.

1. Verify the license lets us do this. I believe we're legally cleared to take this code and heavily modify it for our purposes. I'm unsure however if that means maintaining the existing copyright notice or if extensive enough modification lets us use the normal SFML copyright header.
2. Move drm-common.h and drm-common.c into SFML's src/ directory. I'm guessing the precise location will be src/SFML/Window/DRM. This entails patching the build system to point at these new files and will probably result in simpler build code too.

The above steps are sufficient for a PR by itself but I think we need to go further with this.

3. Rename the files to Common.hpp and Common.cpp. This will start building these files as C++ code.

4. Remove any utilities we're not using. The following steps are easier if we initially minimize how much code we're having to modify.

5. Replace C idioms with C++ idioms. Easy examples of this are replacing `struct T` parameters and local variables with `T`. Other easy changes are replacing C standard header with their C++ equivalent.

6. Use SFML C++ code style. This includes conventions for naming variables, functions, and types and adding namespaces.

7. Potentially start manually inlining this code into the callsities where it's used. For example, if a given function is only needed in one file, it's perhaps better to define that function in that one file instead of maintaining it in a separate .hpp/.cpp pair. If we keep doing this, it's possible that both files disappear but I'm unsure until this process actually begins.

ChrisThrasher:
https://github.com/ChrisThrasher/SFML/commit/2ecb24c363c90eae5fc9bb09a7905f0ddbb34660

I started working this out. I can already tell there are a lot of changes we want to make to this C++ to get it up to standards of the rest of SFML.

ChrisThrasher:
@substring @oomek Can you weigh in on this? Would you be willing to upgrade to SFML 3 if it meant getting to make as many changes to DRM code as you want?

[...]

PR #2259, opened 2022-10-24
Add DRM sources to sfml-window

ChrisThrasher:
## Description

This moves the 3rd party DRM sources into SFML itself since we need to regularly modify these files. Note that we can't immediately start building this as C++ code since it uses C features or extensions that don't work in C++. That transition will require some extra work.

Addresses #2252 

While I was working on this I realized there was no cache variable for `SFML_USE_DRM` so I added that. This means you can edit the CMakeCache to enable or disable this setting without having to reconfigure the project.

How far do we want to take this? Do we want to refactor this as C++ code and do some modest style fixes in the same PR that moves the code or do we tackle this in smaller pieces?

EDT: I expanded the scope of this PR to also include all of the refactoring required to make this code match the rest of SFML.


ChrisThrasher:
Here's a sneak peak of the errors we get when trying to build this C code as C++.

```
/home/thrasher/Projects/sfml/src/SFML/Window/DRM/Common.cpp: In function ‘void drm_fb_destroy_callback(gbm_bo*, void*)’:
/home/thrasher/Projects/sfml/src/SFML/Window/DRM/Common.cpp:46:29: error: invalid conversion from ‘void*’ to ‘drm_fb*’ [-fpermissive]
   46 |         struct drm_fb *fb = data;
      |                             ^~~~
      |                             |
      |                             void*
/home/thrasher/Projects/sfml/src/SFML/Window/DRM/Common.cpp: In function ‘drm_fb* drm_fb_get_from_bo(gbm_bo*)’:
/home/thrasher/Projects/sfml/src/SFML/Window/DRM/Common.cpp:57:49: error: invalid conversion from ‘void*’ to ‘drm_fb*’ [-fpermissive]
   57 |         struct drm_fb *fb = gbm_bo_get_user_data(bo);
      |                             ~~~~~~~~~~~~~~~~~~~~^~~~
      |                                                 |
      |                                                 void*
/home/thrasher/Projects/sfml/src/SFML/Window/DRM/Common.cpp:66:20: error: invalid conversion from ‘void*’ to ‘drm_fb*’ [-fpermissive]
   66 |         fb = calloc(1, sizeof *fb);
      |              ~~~~~~^~~~~~~~~~~~~~~
      |                    |
      |                    void*
/home/thrasher/Projects/sfml/src/SFML/Window/DRM/Common.cpp:103:79: error: ISO C++ forbids compound-literals [-Werror=pedantic]
  103 |                 memcpy(handles, (uint32_t [4]){gbm_bo_get_handle(bo).u32,0,0,0}, 16);
      |                                                                               ^
/home/thrasher/Projects/sfml/src/SFML/Window/DRM/Common.cpp:103:47: error: taking address of temporary array
  103 |                 memcpy(handles, (uint32_t [4]){gbm_bo_get_handle(bo).u32,0,0,0}, 16);
      |                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/thrasher/Projects/sfml/src/SFML/Window/DRM/Common.cpp:104:75: error: ISO C++ forbids compound-literals [-Werror=pedantic]
  104 |                 memcpy(strides, (uint32_t [4]){gbm_bo_get_stride(bo),0,0,0}, 16);
      |                                                                           ^
/home/thrasher/Projects/sfml/src/SFML/Window/DRM/Common.cpp:104:47: error: taking address of temporary array
  104 |                 memcpy(strides, (uint32_t [4]){gbm_bo_get_stride(bo),0,0,0}, 16);
      |                                               ^~~~~~~~~~~~~~~
```

Fixing the compound literals is easy if we could use C++11's `std::array` but will be a bit harder when restricted to C++03. We also need to do some style fixes like replacing all the tabs with spaces. Much of this won't be hard, we just need to agree on what exactly we want to change.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:
I expanded the scope of this PR to also include all of the refactoring required to make this code match the rest of SFML.

ChrisThrasher:
@substring @oomek Can you try this out and make sure I didn't break anything? Thanks!

oomek:
@ChrisThrasher Sorry, I'm a bit preoccupied by the bathroom renovation, I'll check as soon as I can. Thank you very much for getting your hands dirty.

ChrisThrasher:

ChrisThrasher:

substring:
We've stumbled upon some very rare cases where that part would fail
 https://github.com/SFML/SFML/blob/4134e358310cf200ec52fbda3fb7b83027e6c1e1/src/SFML/Window/DRM/DRMContext.cpp#L107-L118

Can you just keep L109->116 from that block ? Let's always trigger the `drmModeSetCrtc`, whichever the situation.

Beside that case, tests were good so far. I'll let @oomek comment further if he wishes

oomek:
> Is this a bug I introduced in this PR?

Not sure if it's strictly a bug but an optimization added by @substring that does not play nicely when the drm app is launched from linux session in VirtualBox. It resulted in a black screen, no console visible upon exiting the app, so it's better to remove it until further investigation. Thanks for the commit reverting it, and your overall involvement in SFML-ifying the common drm functions. I had a chance to test it this evening and all seems to be working as expected.

oomek (approved):

ChrisThrasher:

src/SFML/Window/DRM/DRMContext.cpp
```diff
@@ -130,6 +128,308 @@ namespace
...
+        fb->bo = &bo;
+
+        const sf::Uint32 width = gbm_bo_get_width(&bo);
+        const sf::Uint32 height = gbm_bo_get_height(&bo);
+        const sf::Uint32 format = gbm_bo_get_format(&bo);
+
+        sf::Uint32 strides[4] = {0};
+        sf::Uint32 handles[4] = {0};
+        sf::Uint32 offsets[4] = {0};
+        uint64_t modifiers[4] = {0};
```
ChrisThrasher, line 159: I tried to use `sf::Uint64` here instead but it failed to compile because `Uint64` didn't match the definition of `uint64_t`. I believe one of them was an `unsigned long` and the other an `unsigned long long`.

src/SFML/Window/DRM/DRMContext.cpp
```diff
@@ -130,6 +128,308 @@ namespace
...
+    }
+
+    sf::priv::DrmFb* drmFbGetFromBo(gbm_bo& bo)
+    {
+        int drmFd = gbm_device_get_fd(gbm_bo_get_device(&bo));
+        sf::priv::DrmFb* fb = reinterpret_cast<sf::priv::DrmFb*>(gbm_bo_get_user_data(&bo));
+        if (fb)
+            return fb;
+
+        fb = reinterpret_cast<sf::priv::DrmFb*>(std::calloc(1, sizeof(*fb)));
```
ChrisThrasher, line 149: I'd prefer to replace this with `operator new` but I'm not sure if `new + memset` is better than `calloc + reinterpret_cast`. I seldom write code so low level so hopefully someone can help find a better way to express with within the limits of C++03.

src/SFML/Window/DRM/DRMContext.cpp
```diff
@@ -130,6 +128,308 @@ namespace
...
 
+    void drmFbDestroyCallback(gbm_bo* bo, void* data)
+    {
+        int drmFd = gbm_device_get_fd(gbm_bo_get_device(bo));
+        sf::priv::DrmFb* fb = reinterpret_cast<sf::priv::DrmFb*>(data);
+
+        if (fb->fbId)
+            drmModeRmFB(drmFd, fb->fbId);
+
+        std::free(fb);
```
ChrisThrasher, line 139: Can we get around manually calling `free` when we don't have access to smart pointers? I couldn't think of a way to get rid of this easily.

src/SFML/Window/DRM/DRMContext.cpp
```diff
@@ -130,6 +128,308 @@ namespace
...
+    {
+        drmModeResPtr resources;
+        int ret;
+
+        if (device)
+        {
+            drm.fd = open(device, O_RDWR);
+            ret = getResources(drm.fd, resources);
+            if (ret < 0 && errno == EOPNOTSUPP)
+                std::printf("%s does not look like a modeset device\n", device);
```
ChrisThrasher, line 320: Should we replace the use `printf` with `sf::err`?

src/SFML/Window/DRM/DRMContext.cpp
```diff
@@ -130,6 +128,307 @@ namespace
...
+        drmDevicePtr devices[maxDrmDevices] = { NULL };
+
+        const int numDevices = drmGetDevices2(0, devices, maxDrmDevices);
+        if (numDevices < 0)
+        {
+            sf::err() << "drmGetDevices2 failed: " << std::strerror(-numDevices) << std::endl;
+            return -1;
+        }
+
+        int fd = -1;
```
ChrisThrasher, line 281: Is `fd` a well understood abbreviation? I'm assuming it means file descriptor but I'm not sure. Maybe we ought to use more expressive variable names.

src/SFML/Window/DRM/DRMContext.cpp
```diff
@@ -130,6 +128,307 @@ namespace
...
+        for (int i = 0; i < numDevices; ++i)
+        {
+            drmDevicePtr device = devices[i];
+            int ret;
```
ChrisThrasher, line 285: I hope we can find a better name than `ret`. `ret` tells us nothing about what the variable actually means.

src/SFML/Window/DRM/DRMContext.cpp
Bromeon: This should not be `reinterpret_cast`, anyway.
[`std::calloc()`](https://en.cppreference.com/w/cpp/memory/c/calloc) returns `void*`, which is convertible to other pointer types via `static_cast`.

But couldn't we just do this?
```cpp
fb = new DrmFb();
```
That type is defined as a POD:
https://github.com/SFML/SFML/blob/22033670f8800f077c2f4cd6419b4405feec1240/src/SFML/Window/DRM/DRMContext.hpp#L59-L63

All members (also PODs) will be zero-initialized.

src/SFML/Window/DRM/DRMContext.cpp
Bromeon: It's probably OK as long as the semantics are easy to follow. Workarounds are likely more complex, and this would also allow relatively easy migration to `std::unique_ptr` in SFML 3 (especially when using `new/delete`).

src/SFML/Window/DRM/DRMContext.cpp
ChrisThrasher: Are you sure they're zero-initialized? My impression was that they're left uninitialized.

Bromeon: The subtle difference between:
```cpp
fb = new DrmFb; // uninit
fb = new DrmFb(); // zero
```
😉 

It also works for arrays:
```cpp
ptr = new int[5]; // uninit
ptr = new int[5](); // zero
```

ChrisThrasher: https://godbolt.org/z/jYn3Kq398

That's cool, I didn't know that! In that case this is an easy fixup then.

src/SFML/Window/DRM/DRMContext.cpp
```diff
@@ -130,6 +128,307 @@ namespace
...
+
+        if (fd < 0)
+            sf::err() << "No drm device found!" << std::endl;
+        return fd;
+    }
+
+    int initDrm(sf::priv::Drm& drm, const char* device, const char* modeStr, unsigned int vrefresh)
+    {
+        drmModeResPtr resources;
+        int ret;
```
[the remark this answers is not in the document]
ChrisThrasher: Good find! I'll declare this variable where it's being initialized which also lets me mark it `const`.

src/SFML/Window/DRM/DRMContext.cpp
```diff
@@ -130,6 +128,306 @@ namespace
...
+            fd = -1;
+        }
+        drmFreeDevices(devices, numDevices);
+
+        if (fd < 0)
+            sf::err() << "No drm device found!" << std::endl;
+        return fd;
+    }
+
+    int initDrm(sf::priv::Drm& drm, const char* device, const char* modeStr, unsigned int vrefresh)
```
ChrisThrasher, line 309: I want to change `device` and `modeStr` to `std::string` but I'm not sure how to handle line 359.
```cpp
if (modeStr && *modeStr)
```
I suppose this is saying "if the string is not null and the first character is not null"? Is that the right way to read this? Would this translate to `if (!modeStr.empty()`?

ChrisThrasher:

src/SFML/Window/DRM/DRMContext.cpp
ChrisThrasher: This `char*` return value is seemingly behaving like a `std::optional<std::string>`. There may be no string, or there may be an empty string. Those are distinct shapes. Because of that, I don't think we can replace this with merely a `std::string` since that would fail to model the distinction between a null string and an empty string.

src/SFML/Window/DRM/DRMContext.cpp
```diff
@@ -130,6 +128,306 @@ namespace
...
+            int ret;
+
+            if (!(device->available_nodes & (1 << DRM_NODE_PRIMARY)))
+                continue;
+            // OK, it's a primary device. If we can get the drmModeResources, it means it's also a KMS-capable device.
+            fd = open(device->nodes[DRM_NODE_PRIMARY], O_RDWR);
+            if (fd < 0)
+                continue;
+            ret = getResources(fd, resources);
+            if(std::getenv("SFML_DRM_DEBUG"))
```
ChrisThrasher, line 294: I'm not a fan of using an environment variable to dynamic enable or disable debug logs. I want to change this to instead use the `SFML_DEBUG` preprocessor symbol. That way we stop doing so many environment variable lookups.

eXpl0it3r:

ChrisThrasher:

src/SFML/Window/DRM/DRMContext.hpp
```diff
@@ -176,7 +195,7 @@ class DRMContext : public GlContext
     /// \brief Get Direct Rendering Manager pointer
     ///
     ////////////////////////////////////////////////////////////
-    static drm* getDRM();
+    static Drm* getDRM();
```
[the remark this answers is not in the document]
ChrisThrasher: Good idea! The diff to do this was tiny so I figured it was worth adding to this PR even if it's a little bit out of scope.

ChrisThrasher:

Bambo-Borris (approved):
Feels like we're at diminishing returns style improvement wise. I can't see any obvious issues so I reckon it's ready

ChrisThrasher:
@substring @oomek I made substantial changes since I last pinged you about this PR but have since finished everything I wanted to do. If possible I'd really appreciate you give this a review and test it out on your end so we can get this merged. I think this will help speed up all future work on DRM code. Thanks!

oomek:
Just tested. All looking good. Sorry for the late response.

oomek (approved):

substring:
LGTM

src/SFML/Window/DRM/DRMContext.cpp
```diff
@@ -130,6 +134,306 @@ namespace
...
+
+        int ret = drmModeAddFB2WithModifiers(drmFd, width, height, format, handles, strides, offsets, modifiers, &fb->fbId, flags);
+
+        if (ret)
+        {
+            std::memset(handles, 0, 16);
+            handles[0] = gbm_bo_get_handle(&bo).u32;
+            std::memset(strides, 0, 16);
+            strides[0] = gbm_bo_get_stride(&bo);
+            std::memset(offsets, 0, 16);
```
eXpl0it3r, line 190: Some memset you'll get to rewrite once it's merged to master 😄

src/SFML/Window/DRM/DRMContext.cpp
```diff
@@ -46,8 +48,14 @@
 
 namespace
 {
+    struct DrmFb
```
eXpl0it3r, line 51: What's an Fb?

src/SFML/Window/DRM/DRMContext.cpp
```diff
@@ -130,6 +134,306 @@ namespace
...
+            delete fb;
+            return NULL;
+        }
+
+        gbm_bo_set_user_data(&bo, fb, drmFbDestroyCallback);
+
+        return fb;
+    }
+
+    sf::Uint32 findCrtcForEncoder(const drmModeRes& resources, const drmModeEncoder& encoder)
```
eXpl0it3r, line 206: What's a Crtc?

ChrisThrasher: https://events.static.linuxfound.org/sites/events/files/slides/brezillon-drm-kms.pdf

Screenshot 2022-11-27 at 3 15 30 PM

Looks like `CrtController` is a more apt name.

substring: CRTC is DRM slang, so to say. I suggest you read the kernel documentation, https://docs.kernel.org/gpu/drm-kms.html#crtc-abstraction is a good source explaining what a CRTC is. Scroll up to the top for a short presentation of DRM. There is a number of other acronyms that are related to DRM, and should always be kept to report the DRM naming

src/SFML/Window/DRM/DRMContext.cpp
ChrisThrasher: I'd honestly rather keep it this way than learn the details of how `long long` behaves across C++ standards. Plus it's nice to avoid adding casts.

ChrisThrasher:

eXpl0it3r (approved):

PR #2260, opened 2022-10-24
Enforce variable case

ChrisThrasher:
## Description

Next up we're enforcing variable case. This check caught the most issues but it also required hacking around some issues with how static variables often include prefixes or use an `UpperCase` style. I wasn't going to start changing the API so I had to work around this. The use of clang-tidy-14's `NOLINTBEGIN` and `NOLINTEND` let me create blocks where identifier style was not checked. This is why the clang-tidy Linux image was updated from 20.04 LTS (`ubuntu-latest` still targets this) to 22.04 LTS. I doubled check that these comments do not appeared in the generated documentation.


ChrisThrasher:

ChrisThrasher:
I updated this PR to include checks for having a new enough version of clang-tidy since it's otherwise pretty confusing when it fails.

ChrisThrasher:

ChrisThrasher:

include/SFML/Network/IpAddress.hpp
```diff
@@ -161,9 +161,11 @@ class SFML_NETWORK_API IpAddress
     ////////////////////////////////////////////////////////////
     // Static member data
     ////////////////////////////////////////////////////////////
+    // NOLINTBEGIN(readability-identifier-naming)
     static const IpAddress Any;       //!< Value representing any address (0.0.0.0)
     static const IpAddress LocalHost; //!< The "localhost" address (for connecting a computer to itself locally)
     static const IpAddress Broadcast; //!< The "broadcast" address (for sending UDP messages to everyone on a local network)
+    // NOLINTEND(readability-identifier-naming)
```
eXpl0it3r, line 168: Can the rule not be adjusted to not care about public (static) const?

ChrisThrasher: Not that I'm aware of. I tried but couldn't carve a rule for these `UpperCase` constants we use. The options are [vast](https://clang.llvm.org/extra/clang-tidy/checks/readability/identifier-naming.html) so hopefully I just missed it.

ChrisThrasher:

Bambo-Borris (approved):

eXpl0it3r (approved):

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -632,26 +632,26 @@ m_lastInputTime(0)
...
-            constexpr unsigned long MWM_FUNC_RESIZE   = 1 << 1;
-            constexpr unsigned long MWM_FUNC_MOVE     = 1 << 2;
-            constexpr unsigned long MWM_FUNC_MINIMIZE = 1 << 3;
-            constexpr unsigned long MWM_FUNC_MAXIMIZE = 1 << 4;
-            constexpr unsigned long MWM_FUNC_CLOSE    = 1 << 5;
+            constexpr unsigned long mwmFuncResize   = 1 << 1;
+            constexpr unsigned long mwmFuncMove     = 1 << 2;
+            constexpr unsigned long mwmFuncMinimize = 1 << 3;
+            constexpr unsigned long mwmFuncMaximize = 1 << 4;
+            constexpr unsigned long mwmFuncClose    = 1 << 5;
```
kimci86, line 654: I think those constant should keep the previous formatting to match the constants of [Motif library](https://sourceforge.net/p/motif/code/ci/master/tree/lib/Xm/MwmUtil.h) from which they come from if I understand correctly.

ChrisThrasher: Because these are all defined in succession it's not too hard to wrap them in a `// NOLINTBEGIN` block. I could be convinced to make that change if others also don't want to change these.

Bromeon: Agree with kimci86 here; maybe a comment could clarify why there is this exception.

Maybe a bit off-topic, but why don't we directly include that header? Dependency decoupling?

ChrisThrasher: No problem. I reverted those changes.

ChrisThrasher:

kimci86 (approved):

ChrisThrasher:
Fixed merge conflicts

ChrisThrasher:

eXpl0it3r (approved):

PR #2261, opened 2022-10-25
Turn off Codecov annotations

Bromeon:
## Description

Those annotations create enormous clutter in diff views on GitHub, making PR reviews complicated. Since we are currently not headed toward 100% coverage, they provide little benefit.

Codecov stats aggregation still runs, this is specifically about the inline dialogs.

grafik

See also: https://github.com/codecov/codecov-action/issues/135

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

To truly test this, we might need to have some actual code changes 🤔

Since this changes only the config, we can't see the effect inside this PR.

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -37,4 +37,4 @@ comment:
   require_base: no
 
 github_checks:
-  annotations: true
+  annotations: false


eXpl0it3r:
You can hide annotations on a file basis if it's too distracting.

image

I personally would keep them enabled. It gives a good hint for files we do have tests and also a constant reminder to enhance our tests.

Bromeon:
> You can hide annotations on a file basis if it's too distracting.

Thanks, that's a nice trick -- but it's a pity that it needs to be done for each file.



> I personally would keep them enabled. It gives a good hint for files we do have tests and also a constant reminder to enhance our tests.

But do we plan to concretely act on those annotations?

Or should we maybe check out aggregated coverage statistics and see which files have low coverage? That might be more expedient than on a per-line level.

eXpl0it3r:
I don't get why they didn't add an UI option, but you can just press `a` on a PR and all annotations vanish 😄 

https://twitter.com/natfriedman/status/1366417484698578953

> But do we plan to concretely act on those annotations?

If the code isn't an implementation detail with low-level dependency (i.e. socket/file/etc) then, I think we should act on them (at one point), because it would mean you added a new code path that's not covered by a test.

The aggregated view you can get and it's indeed quite useful.

Bromeon:
> I don't get why they didn't add an UI option, but you can just press `a` on a PR and all annotations vanish 😄
> (Bonus tip: You can press `i` and all comments get hidden)

Amazing, thanks a lot! True GitHub wizard 🧙

This is probably a workaround I could live with, if others deem the Codeconv annotations truly useful. For me they were mostly in the way, but maybe other contributors see this differently.

ChrisThrasher (approved):
When I writing a bunch of unit tests earlier this year I didn't get any value out of the PR annotations but I have routinely been annoyed at their existence. It's not helpful to be told that some audio code is untested since we're so far from being able to test code paths like that.

Thanks @eXpl0it3r for sharing those workarounds though! That's really useful to know about.

PR #1222, opened 2017-04-30
implemented Event::FilesDropped for Windows

Fytch:
**Disclaimer** : this feature is not ready to be merged yet

Related forum thread: https://en.sfml-dev.org/forums/index.php?topic=19096

I have implemented this feature for personal use. I write simple map editors and other small tools with the aid of SFML because I don't want to bring out the big guns like Qt. For something like a map editor, the drag and drop functionality is very handy, especially for textures and materials.

As can be seen in the linked forum thread, I am not the only one wanting this feature. Therefore, I went ahead and created a pull request. This feature is not complete yet but it already works on Windows. I figured that once I take the first step maybe others are more inclined to chime in and collaborate.

Progress:
- [x] Modify `sf::Event` to store the relevant information
- [x] Implement interface in `sf::Window` (see `setFileDroppingEnabled`)
- [x] Implement backend for Windows
- [ ] Implement backend for *nix
- [ ] Implement backend for OSX
- [x] Add an example or another kind of test

Issues:
- `sf::Event::FileDropEvent` is inevitably part of a `union`, so I couldn't use a vector but had to resort to a raw pointer. Personally, I think this is not a big deal because the usage is still very straight forward:
```for (int i=0; i<event.droppedFiles.count; ++i) event.droppedFiles.files[i];```
But maybe there's a better solution?
- As `sf::Event::FileDropEvent` only stores pointers but not the filenames themselves, their lifetime is bound to the window, not to the event. Is there any case in which this window-specific buffer can cause trouble?
- Am I right in assuming that `WindowImplWin32::processEvent` doesn't need to be thread-safe? Otherwise the aforementioned buffer would need to be protected by a mutex, ensuring that no concurrent accesses are possible.

Unfortunately, I'm not versed well enough in X11 to implement this feature for Linux, but I have found [this](https://github.com/mpv-player/mpv/blob/master/input/event.c) which might be handy for anyone wanting to implement this feature.

PR #2265, opened 2022-10-30
Add drag and drop on X11

justanobdy:
## Description

This adds item dropping to SFML in X11, using the XDnD protocol. (Currently this only supports dragging and dropping when the item being dropped is plain text.)

This is called "Item dropping" instead of "File dropping" because on X11 you can drag and drop anything. (a file from a file manager, text from a browser, etc.)

## Example

```cpp
#include <SFML/Graphics.hpp>
#include <iostream>

int main()
{
    sf::RenderWindow window(sf::VideoMode(sf::Vector2u(1280, 720)), "SFML XDnD Example");
    window.setFramerateLimit(60);

    window.setItemDroppingEnabled(true);

    while (window.isOpen())
    {
        {
                window.close();
            if(event.type == sf::Event::ItemDropped) 
                std::cout << event.itemDropped.item->toAnsiString() << std::endl;
        }

        window.clear();
        window.display();
    }
}
```


JonnyPtn:
I think the main thing needed here (along with #1222) is a decision on whether to support multiple types or not. Otherwise it looks reasonable.

For what it's worth I would suggesting starting with support for files and text, then it can be expanded later if desired

JonnyPtn:
I think the main thing needed here (along with #1222) is a decision on whether to support multiple types or not. Otherwise it looks reasonable.

For what it's worth I would suggesting starting with support for files and text, then it can be expanded later if desired

justanobdy:
I agree, only supporting files and text is enough for most people right now. The only thing we still need to do is agree on an API which works on all platforms, and make sure file dropping works on all the platforms that support it.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -2292,6 +2364,60 @@ Vector2i WindowImplX11::getPrimaryMonitorPosition()
...
+
+    // Read the data in one go.
+    XGetWindowProperty(dpy, w, p, 0, static_cast<long>(size), false, AnyPropertyType, &da, &di, &dul, &dul, &prop_ret);
+
+    String value = reinterpret_cast<char*>(prop_ret);
+
+    XFree(prop_ret);
+
+    // Signal the selection owner that we have successfully read the data.
+    XDeleteProperty(dpy, w, p);
```
Bromeon, line 2416: We have two `XGetWindowProperty` calls, but only one `XDeleteProperty`.

Is this correct? Do we not leak memory?

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -2292,6 +2364,60 @@ Vector2i WindowImplX11::getPrimaryMonitorPosition()
...
+}
+
+sf::String WindowImplX11::getWindowStringProperty(Display* dpy, Window w, Atom p)
+{
+    // Taken from https://www.uninformativ.de/blog/postings/2017-04-02/0/POSTING-en.html
+
+    Atom           da, incr, type;
+    int            di;
+    unsigned long  size, dul;
+    unsigned char* prop_ret = NULL;
```
Bromeon, line 2395: Please declare variables as late as possible -- only once you need them. If possible, initialize them immediately during initialization.

Declaring everything at the beginning of the function is relatively outdated C practice, either mandated by old compilers or a way to measure stack memory usage.

Bromeon (commented):
Thanks for the PR!

This seems like yet another PR which might suffer from having no modern `sf::Event` API (see comment). I would advocate we do not block it on that issue though, since that may take forever. Also, SFML 3 APIs are in flux and can be modified at any time.

But we should probably write a `TODO` and possibly a GitHub issue, so we don't forget this.

src/SFML/Window/Unix/WindowImplX11.cpp
justanobdy: I took this code from https://www.uninformativ.de/blog/postings/2017-04-02/0/POSTING-en.html, and i didn't test it very thoroughly, I just wanted to make a quick draft PR so we could figure out how to get it to work on multiple platforms, and then I would go and polish it up before merging it into master.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -2292,6 +2364,60 @@ Vector2i WindowImplX11::getPrimaryMonitorPosition()
...
+        err() << "Data too large and INCR mechanism not implemented\n";
+        return "";
+    }
+
+    // Read the data in one go.
+    XGetWindowProperty(dpy, w, p, 0, static_cast<long>(size), false, AnyPropertyType, &da, &di, &dul, &dul, &prop_ret);
+
+    String value = reinterpret_cast<char*>(prop_ret);
+
+    XFree(prop_ret);
```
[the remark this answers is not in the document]
justanobdy: see https://github.com/SFML/SFML/pull/2265/files/f49594a978e8c1307ace93a505c9a49cbf37433f#r1026700494

src/SFML/Window/Unix/WindowImplX11.cpp
justanobdy: see https://github.com/SFML/SFML/pull/2265/files/f49594a978e8c1307ace93a505c9a49cbf37433f#r1026700494

src/SFML/Window/Unix/WindowImplX11.cpp
ChrisThrasher: I'd also like to see Bromeon's comments be addressed even if this is just a draft PR.

justanobdy: Ok, I've completely changed how the selection is converted. I modded the conversion code from https://github.com/SFML/SFML/blob/master/src/SFML/Window/Unix/ClipboardImpl.cpp to work with mine.

ChrisThrasher:
I see you're using `clang-format-15`. Sadly that version breaks lots of vertical alignment rules so it produces very different results from versions 12, 13, and 14 which we're using in CI. I don't have a great solution other than try to use clang-tidy-14 if you can your hands on an installation of that.

justanobdy:
Ah, that's why it changed some random stuff. I upgraded from Leap 15.4 to Tumbleweed, so that's why it changed. I'll try to see if I can get 14 on my machine.

justanobdy:
Ok, I've used `clang-format-13` and pushed the changes.

src/SFML/Window/OSX/SFKeyboardModifiersHelper.h
```diff
@@ -36,7 +36,7 @@ namespace priv
 {
 class WindowImplCocoa;
 }
-}
+} // namespace sf
```
ChrisThrasher, line 39: I'm guessing these got added because you used clang-format-15 to format. You can remove these since we now disallow using clang-format-15.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -2182,6 +2234,67 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
...
+                                                0x7fffffff,
+                                                False,
+                                                AnyPropertyType,
+                                                &type,
+                                                &format,
+                                                &items,
+                                                &remainingBytes,
+                                                &data);
+
+                sf::String value;
```
ChrisThrasher, line 2267: Where does this get used?

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -1011,6 +1019,54 @@ void WindowImplWin32::processEvent(UINT message, WPARAM wParam, LPARAM lParam)
...
             break;
         }
+
+        // File(s) dropped event
+        case WM_DROPFILES:
+        {
+            const HDROP hDrop = (HDROP)wParam;
+
+            // Get the count of files dropped
+            int count = DragQueryFileW(hDrop, 0xFFFFFFFF, NULL, 0);
```
ChrisThrasher, line 1029:
```suggestion
            const int count = DragQueryFileW(hDrop, 0xFFFFFFFF, NULL, 0);
```

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -1011,6 +1019,54 @@ void WindowImplWin32::processEvent(UINT message, WPARAM wParam, LPARAM lParam)
...
+            // Add the temporary item to a vector of strings
+            m_droppedItems.push_back(item);
+
+            // Create the event
+            Event event;
+
+            // Get the position of the dropped item
+            POINT point;
+            DragQueryPoint(hDrop, &point);
+
```
ChrisThrasher, line 1057:
```suggestion
            // Get the position of the dropped item
            POINT point;
            DragQueryPoint(hDrop, &point);

            // Create the event
            Event event;
```
Let's try to keep variable declarations as close as possible to where they're used.

ChrisThrasher (changes requested):

src/SFML/Window/OSX/SFKeyboardModifiersHelper.h
justanobdy: Visual Studio formatted everything for me, and I guess it used clang-format-15.

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -1011,6 +1018,53 @@ void WindowImplWin32::processEvent(UINT message, WPARAM wParam, LPARAM lParam)
 
             break;
         }
+
+        // File(s) dropped event
+        case WM_DROPFILES:
+        {
+            const HDROP hDrop = (HDROP)wParam;
```
ChrisThrasher, line 1025:
```suggestion
            const HDROP hDrop = static_cast<HDROP>(wParam);
```
I'm not entirely sure `static_cast` is appropriate but in any case, please stop using this C-style cast.

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -1011,6 +1018,53 @@ void WindowImplWin32::processEvent(UINT message, WPARAM wParam, LPARAM lParam)
...
+            // Get all the files' names
+            for (int i = 0; i < count; i++)
+            {
+                std::vector<wchar_t> buffer;
+
+                buffer.resize(DragQueryFileW(hDrop, i, NULL, 0) + 1);
+                DragQueryFileW(hDrop, i, &buffer.front(), static_cast<UINT>(buffer.size()));
+
+                // Insert the file name at the end
+                item.insert(item.getSize(), sf::String(&buffer.front()) + "\n");
```
ChrisThrasher, line 1044:
```suggestion
                item.insert(item.getSize(), sf::String(buffer.data()) + "\n");
```

src/SFML/Window/iOS/EaglContext.mm
```diff
@@ -85,7 +85,7 @@ void ensureInit()
             sf::priv::EaglContext::getFunction("glRenderbufferStorageOES"));
     }
 }
-}
+} // namespace
```
ChrisThrasher, line 88:
```suggestion
}
```

src/SFML/Window/iOS/SFAppDelegate.mm
```diff
@@ -38,7 +38,7 @@
 
 // Current touches positions
 std::vector<sf::Vector2i> touchPositions;
-}
+} // namespace
```
ChrisThrasher, line 41:
```suggestion
}
```

src/SFML/Window/iOS/SensorImpl.mm
```diff
@@ -38,7 +38,7 @@ float toDegrees(float radians)
 {
     return sf::radians(radians).asDegrees();
 }
-}
+} // namespace
```
ChrisThrasher, line 41:
```suggestion
}
```

justanobdy:
is there a way to get clang-format to reformat the whole codebase? because clang-format-13 isn't fixing everything, and I don't want to mess with visual studio's clang-format version

justanobdy:
should I merge all the commits together and start a new pr on a new branch?

ChrisThrasher:
> is there a way to get clang-format to reformat the whole codebase?

Yes, that's what happens when you build the `format` target.

> should I merge all the commits together and start a new pr on a new branch?

Please don't make a new branch or PR. If you want to clean up the commit history then go for it, but you can wait to do that later if you want.

justanobdy:
> Please don't make a new branch or PR. If you want to clean up the commit history then go for it, but you can wait to do that later if you want. If you do that, you can just `git push --force` to update this PR with your new commits.

Ok, I wasn't just wasn't sure if it was ok that a branch named "X11DnD" also had windows support.

ChrisThrasher:
You're good. Branches names aren't that important since they don't appear in the commit history after a merge.

src/SFML/Window/macOS/SFOpenGLView.mm
```diff
@@ -186,6 +186,9 @@ - (void)finishInit
     // Now that we have a window, set up correctly the scale factor and cursor grabbing
     [self updateScaleFactor];
     [self updateCursorGrabbed]; // update for fullscreen
+
+    // Register the drag types that this view can accept
+    //[self registerForDraggedTypes:[NSArray arrayWithObjects:NSFilenamesPboardType, nil]];
```
ChrisThrasher, line 191: Do you recall why this is commented out?

PR #2267, opened 2022-11-05
Disable certain GCC pragmas when using other compilers

### LINKED ISSUE #2266: #pragma GCC issue warning with MSVC
## #pragma GCC issue warning with MSVC

Hello ! with the latest change in https://github.com/SFML/SFML/blob/master/include/SFML/System/Vector2.inl from commit https://github.com/SFML/SFML/commit/cd985e37d2239437b33449b95466579c4239d82b
I get this warning at compilation :
SFML/System/Vector2.inl(40,9): warning C4068: unknown pragma 'GCC'

## Your environment

* windows 10
* latest SFML master
* MSVC

I checked the code and I think it could be easily fixed by either adding
`#if defined(__GNUC__)`
around the pragma or removing it totally and using other names for the parameter. Would it be okay to do this ? Would you like me to do the change and do a Pull Request ?

thanks for your awesome library !

ChrisThrasher:
## Description

Fixes MSVC warning C4068 in public headers while keeping that warning disabled for uses of GCC pragmas in SFML source files. Closes #2266.

I'd like reenable the C4068 warning so that we can systematically catch these issues going forward but the issue is that there are tons of GCC pragmas in TcpSocket.cpp and it would be pretty ugly to disable them all. For now I'm making the smallest change possible to fix the immediate issue.

### DIFF
--- a/include/SFML/Graphics/Glsl.inl
+++ b/include/SFML/Graphics/Glsl.inl
@@ -108,12 +108,16 @@ struct Vector4
     /// \param w Component of the 4D vector
     ///
     ////////////////////////////////////////////////////////////
+#if defined(__GNUC__)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wshadow"
+#endif
     Vector4(T x, T y, T z, T w) : x(x), y(y), z(z), w(w)
     {
     }
+#if defined(__GNUC__)
 #pragma GCC diagnostic pop
+#endif
 
     ////////////////////////////////////////////////////////////
     /// \brief Conversion constructor
--- a/include/SFML/System/Vector2.inl
+++ b/include/SFML/System/Vector2.inl
@@ -31,13 +31,17 @@ constexpr Vector2<T>::Vector2() : x(0), y(0)
 
 
 ////////////////////////////////////////////////////////////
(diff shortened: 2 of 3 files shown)


ChrisThrasher:

eXpl0it3r (approved):

Bromeon (approved):
Looks good.

Should we consider treating compiler warnings as errors in CI? In my opinion, warnings appearing to the end user are bugs in SFML that should be fixed.

Or are there too many warnings at the moment?

ChrisThrasher:
> Should we consider treating compiler warnings as errors in CI? In my opinion, warnings appearing to the end user are bugs in SFML that should be fixed.
> 
> Or are there too many warnings at the moment?

I covered this in the PR description. I may enable that particular warning in the future but I wasn't satisfied with the brute force solution I would have had to use. I would had to copy the pattern used in this PR another ~20 times in some networking code. This was too ugly of a solution for me to propose merging it.

I know test frameworks define convenient macros for disabling warnings in a cross-platform manner. I want to figure out how to do that and then use that pattern on SFML so that it doesn't require 7 lines of code to disable a single warning around a block of code. Once that's in place we can bring back /C4068 and ensure that this problem never happens again.

PR #2269, opened 2022-11-09
Remove the use of `std::memset`

ChrisThrasher:
## Description

`std::memset` was being used to zero-initialize some C structs so I replaced that with their modern C++ equivalent. In some cases, no extra initialization code was required since functions like `make_unique` already ensure the members get initialized.

https://en.cppreference.com/w/cpp/language/zero_initialization

### DIFF
--- a/src/SFML/Audio/SoundFileReaderMp3.cpp
+++ b/src/SFML/Audio/SoundFileReaderMp3.cpp
@@ -97,10 +97,8 @@ bool SoundFileReaderMp3::check(InputStream& stream)
 
 
 ////////////////////////////////////////////////////////////
-SoundFileReaderMp3::SoundFileReaderMp3() : m_numSamples(0), m_position(0)
+SoundFileReaderMp3::SoundFileReaderMp3() : m_io(), m_decoder(), m_numSamples(0), m_position(0)
 {
-    std::memset(&m_io, 0, sizeof(m_io));
-    std::memset(&m_decoder, 0, sizeof(m_decoder));
     m_io.read = readCallback;
     m_io.seek = seekCallback;
 }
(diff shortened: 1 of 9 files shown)


ChrisThrasher:

src/SFML/Audio/SoundFileReaderMp3.cpp
```diff
@@ -99,8 +99,8 @@ bool SoundFileReaderMp3::check(InputStream& stream)
 ////////////////////////////////////////////////////////////
 SoundFileReaderMp3::SoundFileReaderMp3() : m_numSamples(0), m_position(0)
 {
-    std::memset(&m_io, 0, sizeof(m_io));
-    std::memset(&m_decoder, 0, sizeof(m_decoder));
+    m_io      = {};
+    m_decoder = {};
```
Bromeon, line 103: Could we not use the constructor initializer list here?

ChrisThrasher: Good catch! I fixed that.

ChrisThrasher:

src/SFML/Window/Unix/JoystickImpl.cpp
```diff
@@ -398,8 +398,7 @@ std::string getJoystickName(unsigned int index)
     if (fd >= 0)
     {
         // Get the name
-        char name[128];
-        std::memset(name, 0, sizeof(name));
+        char name[128] = {0};
```
kimci86, line 401: This could be simply `char name[128] = {};` or `char name[128] = "";`.

ChrisThrasher: Fixed, thanks

ChrisThrasher:

JonnyPtn (approved):

Bambo-Borris (approved):

eXpl0it3r (approved):

src/SFML/Audio/SoundFileReaderMp3.cpp
eXpl0it3r: @Bromeon does the fix match your expectations? 😄

eXpl0it3r:
Thanks for the 🧹🪣 🙂

PR #2273, opened 2022-11-17
sf::Shader Move Semantics

Jim-Marsden:

----

## Description

Adding Move construction and assignment to sf::Shader

https://github.com/SFML/SFML/wiki/SD:-SFML-3

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

I've changed Shader.test.cpp to ensure sf::Shader is both is_nothrow_move_constructible, and is_nothrow_move_assignable. Talking on the discord I'm not sure how to best test moving. I appreciate any insights on how to better test this. (Aside from type_traits).

```cpp
#include <SFML/Graphics.hpp>
#include <type_traits>

static_assert(!std::is_copy_constructible_v<sf::Shader>);
static_assert(!std::is_copy_assignable_v<sf::Shader>);
static_assert(std::is_nothrow_move_constructible_v<sf::Shader>);
static_assert(std::is_nothrow_move_assignable_v<sf::Shader>);
```


ChrisThrasher:
Can you squash your 2nd commit?

include/SFML/Graphics/Shader.hpp
```diff
@@ -99,6 +99,12 @@ class SFML_GRAPHICS_API Shader : GlResource
     ////////////////////////////////////////////////////////////
     ~Shader();
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Defaulted move constructor
+    ///
+    ////////////////////////////////////////////////////////////
+    Shader(Shader &&) noexcept = default;
```
Bromeon, line 106: We use the `Shader&&` convention (ref/ptr symbol next to type).

Maybe we should agree on a parameter name for move constructors and operators? In some cases we will need to implement it. What about `source`?

include/SFML/Graphics/Shader.hpp
```diff
@@ -111,6 +117,13 @@ class SFML_GRAPHICS_API Shader : GlResource
     ////////////////////////////////////////////////////////////
     Shader& operator=(const Shader&) = delete;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Defaulted move assignment
+    ///
+    ////////////////////////////////////////////////////////////
+    Shader& operator=(Shader&&) noexcept = default;
```
Bromeon, line 124: For assignment operators, we typically use the parameter name `right`.

Bromeon (changes requested):
Thanks for the PR!

Just minor nitpicks -- I know that existing code is not 100% consistent regarding parameter names, but I think it would be good to a) specify them in declarations, and b) keep their names the same for special member functions. Feel free to also add parameter names to other methods in `sf::Shader`.

Bambo-Borris:
Hopefully the first of many PRs to come! Anyway, I second @Bromeon review stuff.

eXpl0it3r:

JonnyPtn:
I don't see any advantage adding parameter names for unused parameters so I would leave them out still

Jim-Marsden:
I'm not opposed to adding parameter names but I think that they add clutter when they won't or can't be used.

include/SFML/Graphics/Shader.hpp
```diff
@@ -111,6 +117,13 @@ class SFML_GRAPHICS_API Shader : GlResource
     ////////////////////////////////////////////////////////////
     Shader& operator=(const Shader&) = delete;
```
Bambo-Borris, line 118: @Bromeon Does this one not also require a `right` param name too then?

Bambo-Borris (approved):

ChrisThrasher:

ChrisThrasher (dismissed):
Thank you!

Jim-Marsden:
As I plan on doing a lot move semantics, I'll keep note of that

JonnyPtn:
So are we committing to declaring redundant/unused parameters then?

ChrisThrasher:
I’d prefer we not name parameters for defaulted operations so we can avoid having to figure out a name and keep consistent.

Jim-Marsden:
> I’d prefer we not name parameters for defaulted operations so we can avoid having to figure out a name and keep consistent.

Should I do anything?

Bambo-Borris:
@Bromeon seems the consistent style for `= delete`/`= default` methods is to omit a parameter name so consistency would be to not add those names in. I guess @Jim-Marsden to be consistent removing those names is the preferred for style consistency.

include/SFML/Graphics/Shader.hpp
ChrisThrasher: Our existing precedent is to not name parameters in defaulted or deleted special member functions. Can we just keep doing that instead of figuring out what to name unused params?

include/SFML/Graphics/Shader.hpp
```diff
@@ -98,7 +98,7 @@ class SFML_GRAPHICS_API Shader : GlResource
     ///
     ////////////////////////////////////////////////////////////
     ~Shader();
-
+
```
ChrisThrasher, line 101: This trailing whitespace will cause the formatting CI jobs to fail.

ChrisThrasher:
Please rebase on `master` and squash this into a single commit.

include/SFML/Graphics/Shader.hpp
```diff
@@ -809,4 +821,4 @@ class SFML_GRAPHICS_API Shader : GlResource
 ///
 /// \see sf::Glsl
 ///
-////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////
```
ChrisThrasher, line 824: Please add this trailing newline back.

Jim-Marsden: I'm so sorry I misunderstood before.

eXpl0it3r:

Bromeon:
@Bambo-Borris 
> seems the consistent style (after looking through the code) for `= delete`/`= default` methods is to omit a parameter name so consistency would be to not add those names in. I guess @Jim-Marsden to be consistent removing those names is the preferred for style consistency.

Yeah, but the inconsistent part here is that we don't do that for any other method.

**Your rule is:** "add parameter names to method declarations, except when they are deleted/defaulted special member functions".
**My rule is:** "add parameter names to method declarations".

A mental model is simpler if there are fewer rules to keep in mind at all times, _especially_ since there is no good reason for this differentiation and it's documented nowhere.

---

Anyway, this can be cleaned up in a separate PR, we might also discuss the parameter names there. I for one don't find `right` a very good name for `op=`'s parameter, but from the perspective of it being an operator, it's probably OK. This is already getting into bikeshedding territory... 😉

Bambo-Borris:
> @Bambo-Borris
> 
> > seems the consistent style (after looking through the code) for `= delete`/`= default` methods is to omit a parameter name so consistency would be to not add those names in. I guess @Jim-Marsden to be consistent removing those names is the preferred for style consistency.
> 
> Yeah, but the inconsistent part here is that we don't do that for any other method.
> 
> **Your rule is:** "add parameter names to method declarations, except when they are deleted/defaulted special member functions". **My rule is:** "add parameter names to method declarations".
> 
> A mental model is simpler if there are fewer rules to keep in mind at all times, _especially_ since there is no good reason for this differentiation and it's documented nowhere.
> 
> Anyway, this can be cleaned up in a separate PR, also since it reaches far beyond just `sf::Shader`. We might also discuss the parameter names there. I for one don't find `right` a very good name for `op=`'s parameter, but from the perspective of it being an operator, it's probably OK. This is already getting into bikeshedding territory... 😉

Yeah I see what you mean now, but as you say beyond the scope of this PR at this stage and definitely better left for one that is specific to ironing out inconsistency in code style surrounding this. I think we would all like to see Jim get a first PR merged 😄

eXpl0it3r:

JonnyPtn:
Not quite - my rule is don't declare parameters that aren't used, and applies to every method so is completely consistent

ChrisThrasher:

ChrisThrasher (approved):

ChrisThrasher:
🎉 Congrats on your first contribution! We hope there's more to come!

kimci86:
Did I miss something or this lacks proper resource management about `m_shaderProgram` member variable?

ChrisThrasher:
> Did I miss something or this lacks proper resource management about `m_shaderProgram` member variable?

I think you're right. A default move operation will leave the moved-from shader in a state where it still attempt to destroy the shader program. I think that value needs to be set to zero when being moved-from.

PR #2280, opened 2022-11-21
Add `sf::UniqueResource` to simplify the implementation of move semantics

ChrisThrasher:
## Description

`sf::UniqueResource` is a class template that owns a handle to a resource in such a way that you can safely move that handle while ensuring that you never free a given resource more than once. Inspired by [`std::experimental::unique_resource`](https://en.cppreference.com/w/cpp/experimental/unique_resource).

This has potential to be used in a number of place. Whether this is worth merging largely hinges on whether it can be used in 3 or more places.

Places to consider using `sf::UniqueResource`:
* `sf::Texture` (Won't let us remove custom copy operations but move operations are simplified)
* `sf::Shader`
* `sf::VertexBuffer`
* `sf::SoundBuffer` (Doesn't let us entirely remove destructor)
* `sf::AudioDevice` (Doesn't let us entirely remove destructor)
* `sf::SoundSource` (This potentially breaks the assumption that `0`-value handles represent a "null" state but maybe I'm misunderstanding OpenAL)
* `XFreePixmap` and other X11 resource cleanup functions (This is possible but painfully verbose for little gain other than purity and satisfaction)
* ~~`sf::priv::WindowImplWin32`~~ (`std::unique_ptr` is better fit for `m_icon` since it's a pointer type and the other members need destruction logic too complicated to capture in a single deleter function)
* ~~`sf::Font::FontHandles`~~ (`std::unique_ptr` is a better fit since these are pointer types but that's also not possible because other Freetype APIs need to take the address of the underlying pointer which is not possible)

### DIFF
--- a/include/SFML/System/UniqueResource.hpp
+++ b/include/SFML/System/UniqueResource.hpp
@@ -27,17 +27,14 @@
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
-#include <optional>
 #include <type_traits>
 #include <utility>
-
-#include <cassert>
 
 
 namespace sf
 {
 ////////////////////////////////////////////////////////////
-/// \brief Class template for storing non-copyable resource handles
+/// \brief Class template for storing noncopyable resource handles
 ///
 ////////////////////////////////////////////////////////////
 template <typename Handle, typename Deleter>
@@ -51,18 +48,12 @@ class UniqueResource
 
 public:
     ////////////////////////////////////////////////////////////
-    /// \brief Default constructor
-    ///
-    ////////////////////////////////////////////////////////////
-    UniqueResource();
-
-    ////////////////////////////////////////////////////////////
     /// \brief Construct from handle
     ///
-    /// \param handle Handle to the resource to store
+    /// \param handle Handle to the resource
     ///
     ////////////////////////////////////////////////////////////
-    UniqueResource(Handle handle);
+    explicit UniqueResource(Handle handle = {});
 
     ////////////////////////////////////////////////////////////
     /// \brief Delete the resource
@@ -101,17 +92,10 @@ public:
     void release();
 
     ////////////////////////////////////////////////////////////
-    /// \brief Dispose of the resource by calling the deleter
+    /// \brief Delete the resource and assign new handle
     ///
     ////////////////////////////////////////////////////////////
-    void reset();
-
-    ////////////////////////////////////////////////////////////
-    /// \brief Dispose of the resource by calling the deleter
-    /// then assign to the resource
-    ///
-    ////////////////////////////////////////////////////////////
-    void reset(Handle handle);
+    void reset(Handle handle = {});
 
     ////////////////////////////////////////////////////////////
     /// \brief Access the underlying handle
@@ -119,9 +103,17 @@ public:
     ////////////////////////////////////////////////////////////
     [[nodiscard]] Handle get() const;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Determine if resource has non-null handle
+    ///
+    /// \return True if handle is non-null
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] operator bool() const;
+
 private:
-    std::optional<Handle> m_handle;    //!< Handle to the managed resource
-    Deleter               m_deleter{}; //!< Functor to clean up resource
+    Handle  m_handle{};  //!< Handle to the managed resource
+    Deleter m_deleter{}; //!< Functor to clean up resource
 };
 
 #include <SFML/System/UniqueResource.inl>
@@ -136,6 +128,13 @@ private:
 /// This class provides a wrapper around a handle to a resource
 /// with a custom deleter that cleans up that resource at the
 /// end of its lifetime.
+///
+/// Like std::unique_ptr, sf::UniqueResource uses the default
+/// constructed state of the handle to represent the "null" state
+/// of the resource. Default construction of a unique resource
+/// or moving out of a unique resource entails the handle being
+/// reset to its default constructed state which typically means
+/// a value of 0.
 ///
 /// Usage example:
 /// \code
--- a/include/SFML/System/UniqueResource.hpp
+++ b/include/SFML/System/UniqueResource.hpp
@@ -1,0 +1,161 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2023 Laurent Gomila (laurent@sfml-dev.org)
+//
+// This software is provided 'as-is', without any express or implied warranty.
+// In no event will the authors be held liable for any damages arising from the use of this software.
+//
+// Permission is granted to anyone to use this software for any purpose,
+// including commercial applications, and to alter it and redistribute it freely,
+// subject to the following restrictions:
+//
+// 1. The origin of this software must not be misrepresented;
+//    you must not claim that you wrote the original software.
+//    If you use this software in a product, an acknowledgment
+//    in the product documentation would be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such,
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+#pragma once
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#include <optional>
+#include <type_traits>
+#include <utility>
+
+#include <cassert>
+
+
+namespace sf
+{
+////////////////////////////////////////////////////////////
+/// \brief Class template for storing non-copyable resource handles
+///
+////////////////////////////////////////////////////////////
+template <typename Handle, typename Deleter>
+class UniqueResource
+{
+    static_assert(!std::is_pointer_v<Handle>, "Handle cannot be a pointer");
+    static_assert(!std::is_const_v<Handle>, "Handle cannot be const");
+    static_assert(std::is_trivially_constructible_v<Handle>, "Handle must be trivially constructible");
+    static_assert(std::is_trivially_destructible_v<Handle>, "Handle must be trivially destructible");
+    static_assert(std::is_invocable_v<Deleter, Handle>, "Deleter must define operator()(Handle)");
+
+public:
+    ////////////////////////////////////////////////////////////
+    /// \brief Default constructor
+    ///
+    ////////////////////////////////////////////////////////////
+    UniqueResource();
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Construct from handle
+    ///
+    /// \param handle Handle to the resource to store
+    ///
+    ////////////////////////////////////////////////////////////
+    UniqueResource(Handle handle);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Delete the resource
+    ///
+    ////////////////////////////////////////////////////////////
+    ~UniqueResource();
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Deleted copy constructor
+    ///
+    ////////////////////////////////////////////////////////////
+    UniqueResource(const UniqueResource&) = delete;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Deleted copy assignment
+    ///
+    ////////////////////////////////////////////////////////////
+    UniqueResource& operator=(const UniqueResource&) = delete;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Move constructor
+    ///
+    ////////////////////////////////////////////////////////////
+    UniqueResource(UniqueResource&& other) noexcept;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Move assignment
+    ///
+    ////////////////////////////////////////////////////////////
+    UniqueResource& operator=(UniqueResource&& other) noexcept;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Release ownership of the managed resource
+    ///
+    ////////////////////////////////////////////////////////////
+    void release();
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Dispose of the resource by calling the deleter
+    ///
+    ////////////////////////////////////////////////////////////
+    void reset();
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Dispose of the resource by calling the deleter
+    /// then assign to the resource
+    ///
+    ////////////////////////////////////////////////////////////
+    void reset(Handle handle);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Access the underlying handle
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] Handle get() const;
+
+private:
+    std::optional<Handle> m_handle;    //!< Handle to the managed resource
+    Deleter               m_deleter{}; //!< Functor to clean up resource
+};
+
+#include <SFML/System/UniqueResource.inl>
+
+} // namespace sf
+
+
+////////////////////////////////////////////////////////////
+/// \class sf::UniqueResource
+/// \ingroup system
+///
+/// This class provides a wrapper around a handle to a resource
+/// with a custom deleter that cleans up that resource at the
+/// end of its lifetime.
+///
+/// Usage example:
+/// \code
+/// struct Deleter
+/// {
+///     void operator()(int) const
+///     {
+///         std::cout << "Cleanup\n";
+///     }
+/// };
+///
+/// sf::UniqueResource<int, Deleter> uniqueResource(42);
+/// useResource(uniqueResource.get());
+/// ...
+/// // When destructing, prints "Cleanup"
+/// \endcode
+///
+/// This type is not copyeable but is moveable. If you move
+/// from this type, the deleter will not be called on the
+/// moved-from instance. This ensures that the resource it
+/// owns will never be deleted twice.
+///
+////////////////////////////////////////////////////////////
--- a/include/SFML/System/UniqueResource.inl
+++ b/include/SFML/System/UniqueResource.inl
@@ -1,0 +1,102 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2023 Laurent Gomila (laurent@sfml-dev.org)
+//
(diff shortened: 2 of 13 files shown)


kimci86:
> Whether this is worth merging largely hinges on whether it can be used in 2 or more places.

Having a quick look at public headers, I think it could be used to manage OpenAL identifiers in SoundBuffer, SoundSource and SoundStream classes and OpenGL identifiers in Shader, Texture, and VertexBuffer classes.
It could also be used as a cleaner alternative to `std::unique_ptr<std::remove_pointer_t<T>, D>` to manage FreeType handles in Font.cpp.

test/System/UniqueResource.test.cpp
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    SUBCASE("Operators")
+    {
+        SUBCASE("Move assignment")
+        {
+            const auto currentDeleteCount = deleteCount;
+            {
+                sf::UniqueResource<int, Deleter>       movedFromUniqueResource(42, {});
+                const sf::UniqueResource<int, Deleter> uniqueResource = std::move(movedFromUniqueResource);
```
kimci86, line 102: This is actually testing the move constructor.

ChrisThrasher: Is that because I'm using ` = ` to begin the lifetime of the object?

ChrisThrasher: Fixed 👍

JonnyPtn:
Personally not sure the benefits of this are worth it, it's still an experimental proposal and adds unnecessary complexity.

Don't feel too strongly though - would rather we just get a decision made quickly so this doesn't continue to hold up other work

src/SFML/Graphics/Shader.cpp
```diff
@@ -235,19 +235,19 @@ struct Shader::UniformBinder
 
 
 ////////////////////////////////////////////////////////////
-Shader::Shader() : m_shaderProgram(0), m_currentTexture(-1), m_textures(), m_uniforms()
+void Shader::ShaderProgramDeleter::operator()(const unsigned int shaderProgram)
```
Bromeon, line 238: SFML doesn't really use top-level CV qualifiers on function parameters; there's also no benefit here.

include/SFML/System/UniqueResource.hpp
```diff
@@ -0,0 +1,160 @@
...
+/// sf::UniqueResource<int, Deleter> uniqueResource;
+/// useResource(uniqueResource.get());
+/// ...
+/// // When destructing, prints "Cleanup"
+/// \endcode
+///
+/// This type is not copyeable but is moveable. If you move
+/// from this type, the deleter will never be called. This
+/// ensures that the resource it owns will never be deleted
+/// twice.
```
Bromeon, line 158: > If you move from this type, the deleter will never be called.

This is confusing -- the deleter _will_ be called, just on the new and not on the moved-from instance.

include/SFML/System/UniqueResource.hpp
```diff
@@ -0,0 +1,160 @@
...
+    ////////////////////////////////////////////////////////////
+    /// \brief Access the underlying resource
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] const T& get() const;
+
+private:
+    T       m_resource; //!< Handle to the managed resource
+    Deleter m_deleter;  //!< Functor to clean up resource
+    bool    m_delete;   //!< Whether or not a resource is currently owned
```
Bromeon, line 120: `m_delete` is not really a clear name, especially since we already have `m_deleter` -- better would be something like `m_isOccupied`.

include/SFML/System/UniqueResource.inl
```diff
@@ -0,0 +1,106 @@
...
+    m_delete   = true;
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T, typename Deleter>
+const T& UniqueResource<T, Deleter>::get() const
+{
+    return m_resource;
+}
```
Bromeon, line 106: This could have an assertion to check that the resource is populated.

Bromeon (commented):
This is an interesting idea, and https://github.com/SFML/SFML/pull/2273#issuecomment-1321155282 shows that this could prevent bugs. A small disadvantage of this approach is the asymmetry (alloc functions are inline in the method, but free functions are non-local in the `operator()`).

Regarding default constructor + `reset()`: this means that every resource has now an additional "nullable" state. This may be good in cases where `-1` or something else was used to mark "invalid", but is it always applicable?

Also, should `reset()` and `operator=` maybe reuse the move constructor? We could possibly add a `swap()` method.

The current design requires `T` to be default-constructible; which may be valid for low-level C types, but is it something we can assume? We also keep a dummy value around after destruction. `std::optional<T>` might be more general here.

If `T` itself has a destructor, this approach will cause surprises. Either we should define semantics for this case, or statically assert that `T` is a POD type (or however the latest C++ revision fancies to call those).

src/SFML/Graphics/Shader.cpp
```diff
@@ -235,19 +235,19 @@ struct Shader::UniformBinder
...
+
+    // Destroy effect program
+    if (shaderProgram)
+        glCheck(GLEXT_glDeleteObject(castToGlHandle(shaderProgram)));
 }
 
 
 ////////////////////////////////////////////////////////////
-Shader::~Shader()
+Shader::Shader() : m_shaderProgram(0, {}), m_currentTexture(-1), m_textures(), m_uniforms()
```
kimci86, line 249: This is taking ownership of a shader program identified with 0.
Default initialization would make more sense, meaning the shader does not own a shader program yet.

src/SFML/Graphics/Shader.cpp
```diff
@@ -786,11 +786,7 @@ bool Shader::compile(const char* vertexShaderCode, const char* geometryShaderCod
     }
 
     // Destroy the shader if it was already created
-    if (m_shaderProgram)
-    {
-        glCheck(GLEXT_glDeleteObject(castToGlHandle(m_shaderProgram)));
-        m_shaderProgram = 0;
-    }
+    m_shaderProgram.reset(0);
```
kimci86, line 789: This is taking ownership of a shader program identified with 0, which seems wrong.
Because `m_shaderProgram.reset();` cleans up resource but does not change the stored identifier (which would be bad), I think `m_shaderProgram = {};` is what we need here.

src/SFML/Graphics/Shader.cpp
ChrisThrasher: We don't add `const` to value parameters _in the declarations_ because doing so has no effect. This is good. However, adding `const` to the parameter _in the definition_ does matter and is something SFML ought to do. I'll remove this for the sake of consistency but we gain nothing by allowing ourselves to mutate these parameters. Hopefully we can change tact and start adding `const` in more places where it can potentially catch bugs and better expresses our intent.

include/SFML/System/UniqueResource.hpp
ChrisThrasher: Right, that's confusing. I'll fix that.

include/SFML/System/UniqueResource.hpp
```diff
@@ -0,0 +1,160 @@
...
+/// \code
+/// struct Deleter
+/// {
+///     void operator()(int)
+///     {
+///         std::cout << "Cleanup\n";
+///     }
+/// };
+///
+/// sf::UniqueResource<int, Deleter> uniqueResource;
```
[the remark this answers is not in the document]
ChrisThrasher: Ah you're right. Thanks.

src/SFML/Graphics/Shader.cpp
Bromeon: > we gain nothing by allowing ourselves to mutate these parameters

I would agree if const were the default and mutable would need explicit annotating, but at the moment we gain consistency between declaration and definition signature by keeping them in sync.

Furthermore, I think `const` parameters is a bit of a theoretical issue that makes code less readable while hardly preventing bugs in real-life code.

It also opens a whole can of worms like
```cpp
void doStuff(const Texture* const texture) { ... }
```

include/SFML/System/UniqueResource.hpp
ChrisThrasher: I swapped this out for an `optional<Deleter>` and I think the implementation looks better now so I just sidestepped that naming question.

include/SFML/System/UniqueResource.inl
ChrisThrasher: Are there any circumstances under which you'd want to access this handle without accessing the resource it points to? If so then we can't `assert` here but I'm struggling to come up with such an example.

include/SFML/System/UniqueResource.hpp
```diff
@@ -0,0 +1,156 @@
...
+    UniqueResource();
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Construct from resource and deleter
+    ///
+    /// \param resource Resource to store
+    /// \param deleter Deleter for the resource
+    ///
+    ////////////////////////////////////////////////////////////
+    UniqueResource(T&& resource, Deleter&& deleter);
```
Bromeon, line 60: Would it make sense to provide an explicit constructor that only accepts a `T&& resource` and default-initializes the deleter? Even the example in the doc would then not need a `{}` argument.

Or do you think it's a compatibility hazard toward `std::experimental::unique_resource`?

test/System/UniqueResource.test.cpp
```diff
@@ -0,0 +1,108 @@
...
+static_assert(std::is_nothrow_move_constructible_v<sf::UniqueResource<int, Deleter>>);
+static_assert(std::is_nothrow_move_assignable_v<sf::UniqueResource<int, Deleter>>);
+
+TEST_CASE("[System] sf::UniqueResource")
+{
+    SUBCASE("Construction")
+    {
+        SUBCASE("Default constructor")
+        {
+            const auto currentDeleteCount = deleteCount;
```
Bromeon, line 32: Maybe call it `prevDeleteCount` to highlight that it preceeded the operation?

test/System/UniqueResource.test.cpp
```diff
@@ -0,0 +1,108 @@
+#include <SFML/System/UniqueResource.hpp>
+
+#include <doctest/doctest.h>
+
+#include <SystemUtil.hpp>
+#include <type_traits>
+
+namespace
+{
+size_t deleteCount = 0;
```
Bromeon, line 10: Unfortunately we decided for `std::size_t` in https://github.com/SFML/SFML/pull/2184 😛

Bromeon (commented):
Looks nice to me!

What is currently untested are stateful deleters, could e.g. be done with a pointer to a local variable. Or maybe conditional:
```cpp
struct Deleter
{
    // not sure if valid default value, otherwise 2 ctors
    explicit Deleter(std::size_t* counter = globalCount) : 
    deleteCount(counter)
    {
    }

    void operator()(int)
    {
        ++(*deleteCount);
    }

    std::size_t* deleteCount;
};
```
And then in some tests:
```cpp
        SUBCASE("Resource and deleter constructor")
        {
            std::size_t localCount = 0;
            {
                const sf::UniqueResource<int, Deleter> uniqueResource(42, Deleter(&localCount));
                CHECK(uniqueResource.get() == 42);
            }
            CHECK(localCount == 1);
        }
```

include/SFML/System/UniqueResource.hpp
ChrisThrasher: I want there to be a clear path to `std::unique_resource` but I don't mind if it requires some minor changes like this.

ChrisThrasher:
I rebased on `master` and a few more CI issues. There's still the bigger problem of why `sf::Shader` is moveable on all platforms except Windows. I'm not sure why a type would moveable on one platform but not the other. 

I'm doing enough to maintain this PR so it doesn't go stale but I still haven't gotten back to putting any serious thought into it.

ChrisThrasher:
Periodic maintenance rebase. No changes made.

ChrisThrasher:
https://github.com/SFML/SFML/blob/839ad6cf70276c6ee78a2f7adc5555e8fc07a6bb/src/SFML/Window/Unix/WindowImplX11.cpp#L625

X11 includes a few non-pointer types that need special cleanup functions. This could benefit from `sf::UniqueResource` as well.

ChrisThrasher:
I'm thinking about this more and I think I want to focus on making this more similar to `std::unique_ptr` for a few reasons.

1. `std::experimental::unique_resource` is not even standardized so it seems excessive to try to match its interface. It was a good starting point but I'm not going to let that stop me from making different decisions.

2. `std::unique_ptr` has proven itself to be a great interface and one that most C++ programmers understand quite well. It's probably the most common move-only type we work with. If we match its semantics, we're replicating a known, good solution for resource management. While these resource handles aren't exactly equivalent to pointers, they're used in a similar manner so I think it's a reasonable parallel to draw.

So how do we get closer to `std::unique_ptr`? First we need some robust way to represent a null state for the resource handle. One option is to treat the default-constructed state of `T` as representing `null`. This is what `std::unique_ptr` does because we've all collectively agreed that pointer types use their default-constructed (or "zero") state to mean "there is no value". In other words `std::unique_ptr` depends on a sentinel value to represent a default initialized or moved-from state.

`sf::UniqueResource` aims to work with arbitrary resource handle types. Those resource handle types often do use the default-constructed state (aka the value `0`) to represent a null state but this is not a safe assumption to make in the general case. Because of this we maybe cannot use a sentinel value to represent an empty or moved-from state. Luckily we can use `std::optional` within the class to bolt on an "empty" state to any arbitrary `T`.

At this point we can now mostly copy the `std::unique_ptr` interface. We can use `operator*` and `operator->` to easily access (but not modify) the contained handle and modify it as needed. We can have `.get()` which returns an `std::optional<T>` just like `std::unique_ptr::get()` returns the raw pointer it's storing inside. We have `operator bool` to easily query whether the unique resource has a value. We match the semantics of `std::unique_ptr` wherein a non-null pointer/resource indicates that the pointer owns something. This eliminates issues where someone tries to use the underlying resource handle after it has been moved. `.release()` and `.reset()` still work as intended when you need to free the current resource or replace the current handle with a handle to a new resource.

There's a lot of value in a type that already obeys existing semantics we all know and understand.

ChrisThrasher:
I've been working on this off and on the last few weeks and pushed up an updated version of `sf::UniqueResource` alongside using it in `sf::Shader` and `sf::Texture`. I'm still on the fence about whether the underlying handle should be nullable via `std::optional` or nullable by using its default-constructed state as the "null" state (like `std::unique_ptr` does) but so far I'm sticking with the `std::optional<Handle>` route. If we use `sf::UniqueResource` in a bunch of places and it still ends up that all handles use `0` as their "null" state then we can remove the `std::optional` and simplify `sf::UniqueResource` in a few ways. However if a single kind of resource handle uses `0` as a valid state then we can't do that.

ChrisThrasher:
I was able to use `UniqueResource` to remove the custom destructors for `sf::Shader`, `sf::Texture`, and `sf::VertexBuffer` which means we've met my minimum requirement for this being a worthwhile addition to the library! After doing that I saw that every single resource we were managing used `0` as its null state so I added one more commit to this PR that takes advantage of that fact and removes `std::optional` from `UniqueResource` in favor of treating the `Handle`'s default-constructed state as being its "null" state. This lets us simplify things significantly.

Are there resources with handles where the default-constructed state is valid? Yeah maybe. However those handles are either not used in SFML or if they are used it's not a big deal to maintain the status quo and handle them in a more manual fashion. This most recent change means that the majority of resource handles are treated in a simple, elegant fashion that is very similar to `std::unique_ptr`.

I know I've changed my mind a few times on this PR but this is the most certain about it I've been so far. I'm marking this as ready for review so we can resume the discussion in hopes of getting this merged soon.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

Bromeon:
> So how do we get closer to `std::unique_ptr`? First we need some robust way to represent a null state for the resource handle.

Do we though? I found `unique_ptr` being and owned pointer _and_ representing a null state a bad separation of responsibilities. It's basically the Billion Dollar Mistake all over again.

I wonder if `sf::UniqueResource` should not have an always-valid state, and use `std::optional<sf::UniqueResource<T>>` if we need an absent value.

Maybe the use cases we handle become too verbose as a result, though; in that case, might be indeed better to have built-in nullability.

ChrisThrasher:
> Do we though?

I reverted that decision between writing that comment and the latest commit. My new preferred solution involves no additional `optional`s.

> Maybe the use cases we handle become too verbose as a result, though; in that case, might be indeed better to have built-in nullability.

The built-in nullability comes from the handles themselves being nullable as per the conventional of the handles we're wrapping. The libraries giving us these handles treat `0` as "null" so we can take advantage of that.

ChrisThrasher:

ChrisThrasher:

Rebased on master and updated tests to Catch2

kimci86:
If we require that a default-constructed handle corresponds to a null state, maybe UniqueResource needs to be in sf::priv namespace as it cannot be used in the general case?
Anyways this convention could be added to the class documentation.

ChrisThrasher:
> If we require that a default-constructed handle corresponds to a null state, maybe UniqueResource needs to be in sf::priv namespace as it cannot be used in the general case?

I'm fine with that. I don't plan on users using this. My only intention is to clean up some of our implementations. We can always change our mind and make it public later if we so choose.

ChrisThrasher:

Bromeon:
> If we require that a default-constructed handle corresponds to a null state, maybe UniqueResource needs to be in sf::priv namespace as it cannot be used in the general case?

If this is at some point no longer the case, we could always have compile-time metaprogramming that would construct a null state for a handle type.

Either something like
```cpp
template <typename T>
struct HandleDefault {
    static T value() { return T(); }
};

// now specialize for concrete T
```
Or if we have lots of handles that are `int` but fundamentally different types:
```cpp
class UniqueResource<T, HandleOp = Default<T>> { ... };
```

---

If the type is only used internally, it should probably be in `sf::priv`. Then we can also iterate on the API, including breaking changes, until we converge on something that works well for a lot of cases. It could also be promoted to the public API as soon as we either want to expose `UniqueResource` in interfaces, or want the user to benefit from similar functionality.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ZXShady:
I also have a fork of this that internally uses unique_ptr as implementation (borderline abuse)

https://github.com/ZXShady/SFML/tree/unique_res

PR #2282, opened 2022-11-22

JonnyPtn:
Rebase of #1645 as it seems GitHub has got it's knickers in a twist


JonnyPtn:
CI failure looks unrelated, think it just needs retrying

include/SFML/Graphics/Font.hpp
```diff
@@ -364,6 +363,11 @@ class SFML_GRAPHICS_API Font
         std::vector<Row> rows;    //!< List containing the position of all the existing rows
     };
 
+    ////////////////////////////////////////////////////////////
+    // Types
+    ////////////////////////////////////////////////////////////
```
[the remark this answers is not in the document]
JonnyPtn: Good spot, suspect this happened during a rebase - there is already the "Types" separator above so will tidy up this area

include/SFML/Graphics/Font.hpp
```diff
@@ -418,15 +422,15 @@ class SFML_GRAPHICS_API Font
     // Types
     ////////////////////////////////////////////////////////////
     class FontHandles;
-    using PageTable = std::unordered_map<unsigned int, Page>; //!< Table mapping a character size to its page (texture)
+    using PageListTable = std::unordered_map<unsigned int, PageList>; ///!< Table mapping a character size to its list of pages (textures)
```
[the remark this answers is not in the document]
JonnyPtn: Yep good spot - thought I caught them all when rebasing but must have missed this

include/SFML/Graphics/Font.hpp
```diff
@@ -280,18 +281,16 @@ class SFML_GRAPHICS_API Font
...
+    /// The texture id is used internally by sf::Text.
     ///
     /// \param characterSize Reference character size
     ///
-    /// \return Texture containing the glyphs of the requested size
+    /// \return Texture id containing glyphs of the requested size
     ///
     ////////////////////////////////////////////////////////////
-    const Texture& getTexture(unsigned int characterSize) const;
+    std::uint64_t getTextureId(unsigned int characterSize) const;
```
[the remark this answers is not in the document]
JonnyPtn: I thought the same, just not sure what the best alternative is - return a container of texture ID's? not pretty but I'm not too fussed as it's just an implementation detail

src/SFML/Graphics/Font.cpp
```diff
@@ -535,15 +553,16 @@ void Font::cleanup()
...
     std::vector<std::uint8_t>().swap(m_pixelBuffer);
 }
 
 
 ////////////////////////////////////////////////////////////
 Font::Page& Font::loadPage(unsigned int characterSize) const
 {
-    return m_pages.try_emplace(characterSize, m_isSmooth).first->second;
+    auto pageList = m_pageLists.try_emplace(characterSize, PageList{}).first->second;
+    return *pageList.insert(pageList.begin(), Page{m_isSmooth});
```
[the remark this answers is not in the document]
JonnyPtn: Yeah seems it's redundant, perhaps was needed when this PR was originally created, will remove

JonnyPtn:
Great feedback @kimci86, thanks.

Think I've covered all the comments now

src/SFML/Graphics/Font.cpp
```diff
@@ -639,11 +656,38 @@ Glyph Font::loadGlyph(std::uint32_t codePoint, unsigned int characterSize, bool
...
 
-        // Get the glyphs page corresponding to the character size
-        Page& page = loadPage(characterSize);
+        // Get the pages list corresponding to the character size
+        PageList& pages = m_pageLists[characterSize];
 
-        // Find a good position for the new glyph into the texture
-        glyph.textureRect = findGlyphRect(page, {width, height});
+        // Find a page that can fit well the glyph
+        Page* page = NULL;
```
Bromeon, line 663: `nullptr`

src/SFML/Graphics/Font.cpp
```diff
@@ -639,11 +656,38 @@ Glyph Font::loadGlyph(std::uint32_t codePoint, unsigned int characterSize, bool
...
+            glyph.textureRect = findGlyphRect(*it, {width, height});
+
+            if (glyph.textureRect.width > 0)
+                page = &*it;
+        }
+
+        // If we didn't find a matching page, create a new one
+        if (!page)
+        {
+            page = &*pages.insert(pages.end(), Page{m_isSmooth});
```
Bromeon, line 676: Maybe `if (page != nullptr)` for explicitness.

I have the feeling
```cpp
pages.push_back(Page(m_isSmooth));
page = &pages.back();
```
might be clearer than this address-deref-insert-at-end expression.

Last, use constructor syntax `Page(m_isSmooth)`.

src/SFML/Graphics/Font.cpp
```diff
@@ -481,11 +495,19 @@ float Font::getUnderlineThickness(unsigned int characterSize) const
...
 
-
 ////////////////////////////////////////////////////////////
-const Texture& Font::getTexture(unsigned int characterSize) const
+std::unordered_set<std::uint64_t> Font::getTextureIds(unsigned int characterSize) const
 {
-    return loadPage(characterSize).texture;
+    PageList& pages = m_pageLists[characterSize];
+    if (pages.empty())
+        pages.push_back(Page{m_isSmooth});
```
Bromeon, line 503: Use constructor syntax `Page(m_isSmooth)`.

src/SFML/Graphics/Font.cpp
```diff
@@ -639,11 +656,38 @@ Glyph Font::loadGlyph(std::uint32_t codePoint, unsigned int characterSize, bool
...
-        // Get the glyphs page corresponding to the character size
-        Page& page = loadPage(characterSize);
+        // Get the pages list corresponding to the character size
+        PageList& pages = m_pageLists[characterSize];
 
-        // Find a good position for the new glyph into the texture
-        glyph.textureRect = findGlyphRect(page, {width, height});
+        // Find a page that can fit well the glyph
+        Page* page = NULL;
+        for (PageList::iterator it = pages.begin(); it != pages.end() && !page; ++it)
```
Bromeon, line 664: `&& page != nullptr`

src/SFML/Graphics/Font.cpp
```diff
@@ -639,11 +656,38 @@ Glyph Font::loadGlyph(std::uint32_t codePoint, unsigned int characterSize, bool
...
+        {
+            // Try to find a good position for the new glyph into the texture
+            glyph.textureRect = findGlyphRect(*it, {width, height});
+
+            if (glyph.textureRect.width > 0)
+                page = &*it;
```
Bromeon, line 670: This assigns to `glyph.textureRect` even if no good rect is found, and leaves this assigned beyond the `for` loop. I see that you have another `glyph.textureRect = findGlyphRect(*page, {width, height});` further down, so it's correct.

However, this flow is very prone to errors on refactorings. The function is also getting rather big, and variable mutations are mixed with "found" state. A separate function could make things easier:
```cpp
struct MatchingPage {
    Page* page = nullptr;
    IntRect textureRect;
};

std::optional<MatchingPage> findMatchingPage(...);
```
```cpp
// usage:
if (auto matching = findMatchingPage(...)) {
    page = matching->page;
    glyph.textureRect = matching->textureRect;
    ...
} else {
    ...
}
```

src/SFML/Graphics/Text.cpp
```diff
@@ -484,19 +519,19 @@ void Text::ensureGeometryUpdate() const
         // If we're using the underlined style and there's a new line, draw a line
         if (isUnderlined && (curChar == U'\n' && prevChar != U'\n'))
         {
-            addLine(m_vertices, x, y, m_fillColor, underlineOffset, underlineThickness);
+            addLine(m_fillVerticesMap[NULL], x, y, m_fillColor, underlineOffset, underlineThickness);
```
Bromeon, line 522: `nullptr`

Bromeon (commented):
Thanks a lot for this! 👍

src/SFML/Graphics/Font.cpp
ChrisThrasher: Why deref then take the address? Can't we just use the return value of `std::unordered_map::insert`?

JonnyPtn: > Why deref then take the address? Can't we just use the return value of `std::list::emplace_back`?
> 
> ```c++
> page = pages.emplace_back(m_isSmooth);
> ```

I can't say why as I didn't originally write this, but using emplace_back seems like a good shout, will do


> Maybe `if (page == nullptr)` for explicitness.

  I don't think comparing to nullptr makes it more explicit, if anything it makes it less intuitive (we're checking if there's not a page, which is exactly how this code reads)


> Last, use constructor syntax `Page(m_isSmooth)`.

why?

include/SFML/Graphics/Font.hpp
```diff
@@ -364,22 +364,15 @@ class SFML_GRAPHICS_API Font
         std::vector<Row> rows;    //!< List containing the position of all the existing rows
     };
 
+    using PageList = std::list<Page>; //!< List of pages, where each page corresponds to a texture
+    using PageListTable = std::unordered_map<unsigned int, PageList>; //!< Table mapping a character size to its list of pages (textures)
```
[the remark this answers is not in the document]
JonnyPtn: I can't say the original reasoning why, but not sure it hugely matters - I presume it was to avoid reallocations. If you think there's a better container to use I'm all ears

src/SFML/Graphics/Font.cpp
```diff
@@ -495,9 +517,12 @@ void Font::setSmooth(bool smooth)
     {
         m_isSmooth = smooth;
```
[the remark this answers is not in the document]
JonnyPtn: I'm not sure what you mean by "involved". If it's to avoid the extra level of scope I'm not sure it's worth it. The function is only a dozen lines and 3 levels of scope

src/SFML/Graphics/Font.cpp
JonnyPtn: As above, I don't consider this an improvement

src/SFML/Graphics/Font.cpp
JonnyPtn: Personally I think this would over-engineer it for little gain - the behaviour is only required in one place so I would prefer to use a lambda limited to this scope, but then I don't think that would really address any of your concerns.

Would like to hear other opinions on this though

src/SFML/Graphics/Font.cpp
JonnyPtn: Why?

JonnyPtn: Actually can just replace with emplace_back then it doesn't matter

Bromeon: > Why?

Because that's the convention SFML uses and there is no reason to deviate in this instance.

Bromeon:
AFAIK there's no convention in SFML regarding pointer nullability checks.
To get a quick impression what people prefer, maybe react to this message...

with 🎉 for this style:
```cpp
if (ptr)
if (!ptr)
```

and 👀 for this style:
```cpp
if (ptr != nullptr)
if (ptr == nullptr)
```

I'm fine with either, but would be good to settle. I'll add 1 reaction each.

Similar topic: `if (integer == 0)`, here I would advocate always explicit checks.

src/SFML/Graphics/Font.cpp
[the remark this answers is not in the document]
Bromeon: I meant as an optimization, to not run the rest of the function if nothing changes.

[It's what we already do in `sf::Texture::setSmooth()`.](https://github.com/SFML/SFML/blob/master/src/SFML/Graphics/Texture.cpp#L643) Having an extra indentation is not great though, hence suggestion of early return.

include/SFML/Graphics/Font.hpp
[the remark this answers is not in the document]
Bromeon: I think it was rather to ensure that pointers to elements remain valid even if new elements are inserted or existing ones removed.

If this happens only at the end or beginning, `std::deque` would also work -- in any case, let's not change this within this PR, too much risk of breaking 🙂

ChrisThrasher:

src/SFML/Graphics/Font.cpp
```diff
@@ -352,26 +353,40 @@ const Font::Info& Font::getInfo() const
...
-        return it->second;
-    }
-    else
-    {
-        // Not found: we have to load it
-        Glyph glyph = loadGlyph(codePoint, characterSize, bold, outlineThickness);
-        return glyphs.emplace(key, glyph).first->second;
+        // Search the glyph into the cache
+        const GlyphTable&          glyphs = page.glyphs;
+        GlyphTable::const_iterator it2    = glyphs.find(key);
```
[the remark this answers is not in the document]
JonnyPtn: Nope, and a few other small bits here could be tidied up

src/SFML/Graphics/Font.cpp
```diff
@@ -639,11 +657,38 @@ Glyph Font::loadGlyph(std::uint32_t codePoint, unsigned int characterSize, bool
...
-        // Get the glyphs page corresponding to the character size
-        Page& page = loadPage(characterSize);
+        // Get the pages list corresponding to the character size
+        PageList& pages = m_pageLists[characterSize];
 
-        // Find a good position for the new glyph into the texture
-        glyph.textureRect = findGlyphRect(page, {width, height});
+        // Find a page that can fit well the glyph
+        Page* page{};
+        for (PageList::iterator it = pages.begin(); it != pages.end() && !page; ++it)
```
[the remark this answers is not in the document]
JonnyPtn: Sure why not

JonnyPtn:
What else are we waiting for on this?

Bromeon:
@SFML/sfml Any more comments on this one? We should decide if we want to move ahead with this PR or not, as merge conflicts start accumulating.

ChrisThrasher:
> @SFML/sfml Any more comments on this one? We should decide if we want to move ahead with this PR or not, as merge conflicts start accumulating.

I defer to the judgment of @binary1248 or anyone else who understands OpenGL since this is outside of my expertise. It seems like a good idea from what I can tell though.

binary1248 (approved):

binary1248:
My only concern in #1645 was that performance might take a hit, but that can be optimized after this gets merged. It would still be interesting to hear from text heavy users if/how this commit affects their performance.

JonnyPtn:
Going to close this for the time being as I don't want to keep updating it and I've never heard anybody with a use-case for it since it was originally requested many years ago, and I don't fancy manufacturing an artificial test case for it

Happy to re-open if there's more motivation in future

PR #2283, opened 2022-11-22
Retry choco install step up to 10 times

JonnyPtn:
Frustrating having CI fail because this step sporadically fails

This change should make it retry up to 10 times when needed. Obviously hard to test without being able to cause a failure, but if it still goes wrong we can tweak as needed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,10 +78,13 @@ jobs:
       run: brew install gcovr
 
     - name: Install OpenCppCoverage and add to PATH
+      uses: nick-fields/retry@v2
       if: matrix.type.name == 'Debug' && runner.os == 'Windows'
-      run: |
-        choco install OpenCppCoverage -y
-        echo "C:\Program Files\OpenCppCoverage" >> $env:GITHUB_PATH
+      with:
+        max_attempts: 10
+        command: |
+          choco install OpenCppCoverage -y
+          echo "C:\Program Files\OpenCppCoverage" >> $env:GITHUB_PATH
 
     - name: Configure CMake
       shell: bash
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -78,10 +78,14 @@ jobs:
...
 
     - name: Install OpenCppCoverage and add to PATH
+      uses: nick-fields/retry@v2
       if: matrix.type.name == 'Debug' && runner.os == 'Windows'
-      run: |
-        choco install OpenCppCoverage -y
-        echo "C:\Program Files\OpenCppCoverage" >> $env:GITHUB_PATH
+      with:
+        max_attempts: 10
+        timeout_minutes: 3
```
ChrisThrasher, line 85: Does this mean it will attempt 10 times in a span of 3 minutes or that each attempt may take up to 3 minutes? If it's 10 attempts in 3 minutes then I think that's great.

JonnyPtn: This is relatively new to me, but my understanding is this:
- If the step fails it will retry up to 10 times before continuing as a failure
- If the step takes longer than 3 minutes it will count that as a "failure" and will retry (I don't expect this ever to happen, as our success should take around a minute and the failure only takes a couple of seconds)
- There is a default (but configurable if desired) pause of 10 seconds between each retry.

So in terms of what delay this has on the CI, there would be none if everything works, and if nothing works then it would be delayed by around 2 minutes while retrying

If something happens in future which means this step gets stuck and hits the timeout then it would cause a much more significant delay, but I think that's very speculative, and even if it does happen it would be better to have this 3 minute timeout rather than no timeout which could cause the job to hang indefinitely (or whatever the overall GitHub actions job timeout is)

ChrisThrasher:

ChrisThrasher (approved):
This looks good. I'm happy to not re-run CI quite so often. The nature of this PR is such that it's hard to verify in isolation so I think we'll have to roll with this for a few days/weeks before we understand how to best tweak the various parameters.

PR #2284, opened 2022-11-23
Only run formatting job on linux

JonnyPtn:
It's currently failing on mac as brew has updated to clang-format 15 which has different rules.

Formatting shouldn't be platform specific anyway, so we really only need to run it on one platform, and linux is fixed on version 12

May also want to consider adding a max version to the check in Format.cmake too, as it seems >=15 won't work

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,28 +115,19 @@ jobs:
         cmake --build $GITHUB_WORKSPACE/test/install/build --config ${{ matrix.type.name == 'Debug' && 'Debug' || 'Release' }}
 
   format:
-    name: Formatting on ${{ matrix.platform.name }}
-    runs-on: ${{ matrix.platform.os }}
+    name: Formatting
+    runs-on: ubuntu-latest
 
     strategy:
       fail-fast: false
-      matrix:
-        platform:
-        - { name: Windows, os: windows-2022,  executable: clang-format }
-        - { name: Linux,   os: ubuntu-latest, executable: clang-format-12 }
-        - { name: macOS,   os: macos-12,      executable: clang-format }
 
     steps:
     - name: Checkout Code
       uses: actions/checkout@v3
 
-    - name: Install macOS Dependencies
-      if: runner.os == 'macOS'
-      run: brew install clang-format
-
     - name: Format Code
       shell: bash
-      run: cmake -DCLANG_FORMAT_EXECUTABLE=${{ matrix.platform.executable }} -P $GITHUB_WORKSPACE/cmake/Format.cmake
+      run: cmake -DCLANG_FORMAT_EXECUTABLE=clang-format-12 -P $GITHUB_WORKSPACE/cmake/Format.cmake
 
     - name: Check Formatting
       run: git diff --exit-code


ChrisThrasher:
> Formatting shouldn't be platform specific anyway, so we really only need to run it on one platform, and linux is fixed on version 12

It's not platform specific but those 3 formatting jobs test with clang-format versions 12, 13, and 14 so they cover a wide range of potential versions users may have. I knew if we forced users to use a specific installation of clang-format, it would drastically reduce the number of contributors who could run the autoformatter locally.

> and linux is fixed on version 12

It's worth mentioning that this will get bumped up to version 14 soon when GitHub changes the `ubuntu-latest` image to point at Ubuntu 22 instead. This should be fine though.

> May also want to consider adding a max version to the check in Format.cmake too, as it seems >=15 won't work

I think this is our best option. Because we already have code for checking the minimum version, I think we can easily do the same to stop people from using v15. 

---

I also wanted to mention how some of the formatting discrepancies can be solved. In particular the trailing `//namespace foo` comments that clang-format-15 produces are not rejected by the prior versions, they were just missed by those versions. We can merge [this commit](https://github.com/ChrisThrasher/SFML/commit/0ecb6e9323639d46aa658a11a831411a7b2e9935) to reduce the diff between what clang-format-14 and 15 produce.

JonnyPtn:
I'd say if we wanted to test multiple versions of clang-format it would still be better using a linux image to test them all.

I'm still not convinced this is a good idea though - the fact that it currently works with multiple versions is more by chance than by design. There may well be some rule we want to enable in future that doesn't work or works differently across those versions so aiming for any kind of backwards compatibility is just adding extra hassle.

Also don't buy into the fact that it will severely limit contributors ability to format - Installing clang-format and installing a specific version of clang format are mostly the same amount of hassle

This is all by the by though - main goal here is just to fix the CI checks

eXpl0it3r:
I don't really mind having it on just one OS, beyond ensuring in general that it can be run on all OS.

But if we're currently running all versions, I'd keep it for now and rather remove it, once we cross that bridge.

JonnyPtn:
I'm not entirely sure to be honest - I don't see where it gets installed in our config so perhaps is on the machine by default?

we're not really currently running all versions currently, at least not intentionally - Linux was the only one with an explicit version (12), mac just used whatever default version brew provided before 15 (I would guess 14 but not entirely sure) and I'm not entirely sure where the windows one comes from either so don't know what version that is.

Regardless, It doesn't seem particularly easy from my perspective to cover all 3 versions we claim to support (and as I said before, I don't think supporting multiple versions is a good idea in the first place) so would prefer to just get this merged to fix the checks, unless somebody else can provide an easy way to check all the versions

ChrisThrasher:
> Regardless, It doesn't seem particularly easy from my perspective to cover all 3 versions we claim to support

https://github.com/SFML/SFML/pull/2287

It's not too hard to run all 3 clang-format jobs on Linux. When I originally wrote this I didn't think about how package managers like `brew` may upgrade clang-format without us knowing so it's valid to lock down the precise versions we're using.

eXpl0it3r:
Superseded by #2287

PR #2285, opened 2022-11-23
Set option warnings as errors to false #2046

MarcDirven:
Temp fix for #2046

### DIFF
--- a/cmake/CompilerWarnings.cmake
+++ b/cmake/CompilerWarnings.cmake
@@ -4,7 +4,7 @@
 
 # Helper function to enable compiler warnings for a specific set of files
 function(set_file_warnings)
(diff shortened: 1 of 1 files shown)


ChrisThrasher:
If you need a temporary fix then you can add `-DWARNINGS_AS_ERRORS=OFF` to your CMake configuration command. That's why we make that an option.

MarcDirven:
Then I think I misunderstood. You want me to fix the deprecations itself instead?

ChrisThrasher:
> Then I think I misunderstood. You want me to fix the deprecations itself instead?

No, you don't need to bother with that. First we need a CI job which uses the compiler in question. We can disable warnings as errors (via the above mentioned CMake option) for that specific CI job so that CI still passes.

PR #2286, opened 2022-11-24
Use enum class instead of plain enum for PrimitiveType

JonnyPtn:
Rebase of #1726


include/SFML/Graphics/PrimitiveType.hpp
```diff
@@ -36,7 +36,7 @@ namespace sf
 /// and view.
 ///
 ////////////////////////////////////////////////////////////
-enum PrimitiveType
+enum class PrimitiveType
```
Jim-Marsden, line 39: I thought there was talk about using enum struct over enum class?

JonnyPtn: As far as I know there's no difference, but `enum class` is the de facto standard

ChrisThrasher: `enum class` is what SFML is already using so let's please keep using that.

ChrisThrasher:

ChrisThrasher (approved):
Glad to see us chipping away at converting more of these old style enumerations into scoped enumerations.

Bromeon (approved):

Bambo-Borris (approved):

PR #2293, opened 2022-12-11
Enable warnings-as-errors for clang-cl

ChrisThrasher:
## Description

When #2114 was merged, there were some clang-cl compiler warnings we couldn't entirely make sense of so we disabled warnings as errors for that compiler. This PR fixes those compiler errors.

clang-cl recognizes `/WX /W4` which translate to `-Werror -Wdeprecated-declarations` internally to Clang. Because we use some deprecated APIs in the Windows code, this caused a build break. This isn't a problem on MSVC because we also added `/wd4996` which disables deprecation warnings in the compiler. However, that flag isn't recognized by clang-cl which continues to use `-Werror -Wdeprecated-declarations`.

Luckily there are some preprocessor definitions we can set which disable deprecation warnings in the code so that no matter what compiler we use, we don't get deprecation warnings for those specific APIs.

### DIFF
--- a/cmake/CompilerWarnings.cmake
+++ b/cmake/CompilerWarnings.cmake
@@ -38,7 +38,6 @@ function(set_target_warnings target)
             /permissive- # standards conformance mode
 
             # Disables, remove when appropriate
-            /wd4996 # disable warnings about deprecated functions
             /wd4068 # disable warnings about unknown pragmas (e.g. #pragma GCC)
             /wd4505 # disable warnings about unused functions that might be platform-specific
             /wd4800 # disable warnings regarding implicit conversions to bool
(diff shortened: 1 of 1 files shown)


ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

Bambo-Borris (approved):

Bambo-Borris:
Glad to see us getting a simpler approach for supporting `clang-cl`. We still need to setup a CI for the other form of clang available on Windows (we've seen issues with that form mentioned recently). That's the just `clang` form.

ChrisThrasher:
I tweaked how I disable deprecation warnings so that it will also fix issues with building with Clang.exe which I plan on adding a CI job for in the near future.

ChrisThrasher:

cmake/CompilerWarnings.cmake
```diff
@@ -80,9 +73,15 @@ function(set_target_warnings target)
...
 
-    if(SFML_COMPILER_CLANG)
+    if(SFML_COMPILER_CLANG OR SFML_COMPILER_CLANG_CL)
         target_compile_options(${target} PRIVATE
             -Wno-unknown-warning-option # do not warn on GCC-specific warning diagnostic pragmas
         )
     endif()
+
+    # Disable certain deprecation warnings
+    if(SFML_OS_WINDOWS)
```
eXpl0it3r, line 83: Should those really be enabled for all Windows compilers?

ChrisThrasher: Yes, because we need to disable those deprecation warnings for MSVC, clang-cl, and Clang.

eXpl0it3r (approved):

PR #2294, opened 2022-12-11
Fix Clang build on Windows

ChrisThrasher:
## Description

Now that #2293 is merged, I believe this will fix the last remaining issues presented in #2046.


ChrisThrasher:

eXpl0it3r (approved):

cmake/CompilerWarnings.cmake
```diff
@@ -71,7 +71,7 @@ function(set_file_warnings)
     
     # For now if we're using MSVC-like clang interface on Windows
     # we'll disable warnings as errors
```
kimci86, line 73: I think this comment needs to be updated because we disable warnings are errors with clang too (not only MSVC-like clang).

ChrisThrasher:

Bambo-Borris (approved):
Looks good to me. Will be glad to see Clang + Windows combo be smoother to work with for 2.6.x! 😄

Issue #2297, opened 2022-12-14
Fix Clang warnings in DRM code


ChrisThrasher:
## Subject of the issue

```
/home/thrasher/Projects/sfml/src/SFML/Window/DRM/InputImplUDev.cpp:415:36: error: implicit conversion changes signedness: 'sf::Keyboard::Key' to 'std::vector<bool, type-parameter-0-0>::size_type' (aka 'unsigned long') [-Werror,-Wsign-conversion]
                            keyMap[kb] = inputEvent.value;
                            ~~~~~~ ^~
/home/thrasher/Projects/sfml/src/SFML/Window/DRM/InputImplUDev.cpp:496:39: error: use of old-style cast [-Werror,-Wold-style-cast]
        newTerminalConfig.c_lflag &= ~(tcflag_t)ICANON;
                                      ^         ~~~~~~
/home/thrasher/Projects/sfml/src/SFML/Window/DRM/InputImplUDev.cpp:567:19: error: implicit conversion changes signedness: 'Keyboard::Key' to 'std::vector<bool, type-parameter-0-0>::size_type' (aka 'unsigned long') [-Werror,-Wsign-conversion]
    return keyMap[key];
           ~~~~~~ ^~~
/home/thrasher/Projects/sfml/src/SFML/Window/DRM/InputImplUDev.cpp:695:35: error: use of old-style cast [-Werror,-Wold-style-cast]
    newTerminalConfig.c_lflag &= ~(tcflag_t)ECHO;         // disable console feedback
                                  ^         ~~~~
/home/thrasher/Projects/sfml/src/SFML/Window/DRM/InputImplUDev.cpp:696:35: error: use of old-style cast [-Werror,-Wold-style-cast]
    newTerminalConfig.c_lflag &= ~(tcflag_t)ISIG;         // disable signals
                                  ^         ~~~~
5 errors generated.
```

Clang generates some warnings in the DRM code that GCC does not. It's just a matter of fixing some casts. The warnings are present in `2.6.x` and `master` so let's fix this in `2.6.x` so both branches can benefit from it.

ChrisThrasher:
Fixed in #2297

PR #2298, opened 2022-12-14
Replace c-style casts & fix implicit conversion warnings in DRM code

zakinadhif:

----

## Description

Replaces c-style casts with c++ cast operators and fixes "implicit conversion changes singedness" warnings.

This PR is related to the issue #2297

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/src/SFML/Window/DRM/InputImplUDev.cpp
+++ b/src/SFML/Window/DRM/InputImplUDev.cpp
@@ -412,7 +412,7 @@ namespace
                             event.key.shift = shiftDown();
                             event.key.system = systemDown();
 
-                            keyMap[kb] = inputEvent.value;
+                            keyMap[static_cast<std::size_t>(kb)] = inputEvent.value;
 
                             if (special && inputEvent.value)
                                 doDeferredText = special;
@@ -493,7 +493,7 @@ namespace
         //
         // We only clear the ICANON flag for the time of reading
 
-        newTerminalConfig.c_lflag &= ~(tcflag_t)ICANON;
+        newTerminalConfig.c_lflag &= ~static_cast<tcflag_t>(ICANON);
         tcsetattr(STDIN_FILENO, TCSANOW, &newTerminalConfig);
 
         timeval timeout;
@@ -564,7 +564,7 @@ bool InputImpl::isKeyPressed(Keyboard::Key key)
         return false;
 
     update();
-    return keyMap[key];
+    return keyMap[static_cast<std::size_t>(key)];
 }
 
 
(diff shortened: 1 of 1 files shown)


src/SFML/Window/DRM/InputImplUDev.cpp
```diff
@@ -692,8 +692,8 @@ void InputImpl::setTerminalConfig()
 
     tcgetattr(STDIN_FILENO, &newTerminalConfig);          // get current terminal config
     oldTerminalConfig = newTerminalConfig;                // create a backup
-    newTerminalConfig.c_lflag &= ~(tcflag_t)ECHO;         // disable console feedback
-    newTerminalConfig.c_lflag &= ~(tcflag_t)ISIG;         // disable signals
+    newTerminalConfig.c_lflag &= ~static_cast<tcflag_t>(ECHO);         // disable console feedback
```
zakinadhif, line 695: Should this be rewritten as such?
```cpp
newTerminalConfig.c_lflag &= static_cast<tcflag_t>(~ECHO);
```

ChrisThrasher: I do't believe so because the cast should be occurring before the `~` is applied.

src/SFML/Window/DRM/InputImplUDev.cpp
```diff
@@ -412,7 +412,7 @@ namespace
                             event.key.shift = shiftDown();
                             event.key.system = systemDown();
 
-                            keyMap[kb] = inputEvent.value;
+                            keyMap[static_cast<size_t>(kb)] = inputEvent.value;
```
ChrisThrasher, line 415:
```suggestion
                            keyMap[static_cast<std::size_t>(kb)] = inputEvent.value;
```
We use the `std::` prefix for integer types ubiquitously now.

src/SFML/Window/DRM/InputImplUDev.cpp
```diff
@@ -564,7 +564,7 @@ bool InputImpl::isKeyPressed(Keyboard::Key key)
         return false;
 
     update();
-    return keyMap[key];
+    return keyMap[static_cast<size_t>(key)];
```
ChrisThrasher, line 567:
```suggestion
    return keyMap[static_cast<std::size_t>(key)];
```

ChrisThrasher (changes requested):
Pending some minor changes I think this is good!

ChrisThrasher:

src/SFML/Window/DRM/InputImplUDev.cpp
zakinadhif: Alrighty

src/SFML/Window/DRM/InputImplUDev.cpp
zakinadhif: Done, should I also replace the occurrences in line 304 - 305?

ChrisThrasher (approved):
Thanks for taking care of this so quickly!

ChrisThrasher:

eXpl0it3r (approved):

PR #2299, opened 2022-12-15
Add -ObjC flag to fix static linking on macOS

ChrisThrasher:
## Description

Fixes regression introduced in #2174 of a bug fixed in #1485.

I think this made it through CI since it only appears when running certain Window module code and our tests don't run that since such tests would fail on Linux.

### DIFF
--- a/src/SFML/Window/CMakeLists.txt
+++ b/src/SFML/Window/CMakeLists.txt
@@ -288,6 +288,12 @@ target_link_libraries(sfml-window PUBLIC SFML::System)
 # glad sources
 target_include_directories(sfml-window SYSTEM PRIVATE "${PROJECT_SOURCE_DIR}/extlibs/headers/glad/include")
 
+# When static linking on macOS, we need to add this flag for objective C to work
+# https://developer.apple.com/library/archive/qa/qa1490/_index.html
+if ((NOT BUILD_SHARED_LIBS) AND SFML_OS_MACOSX)
+    target_link_libraries(sfml-window PRIVATE -ObjC)
+endif()
+
 # Vulkan headers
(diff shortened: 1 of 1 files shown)


src/SFML/Window/CMakeLists.txt
```diff
@@ -288,6 +288,11 @@ target_link_libraries(sfml-window PUBLIC SFML::System)
 # glad sources
 target_include_directories(sfml-window SYSTEM PRIVATE "${PROJECT_SOURCE_DIR}/extlibs/headers/glad/include")
 
+# When static linking on macOS, we need to add this flag for objective C to work
+if ((NOT BUILD_SHARED_LIBS) AND SFML_OS_MACOSX)
+    target_link_libraries(sfml-window PRIVATE -ObjC)
```
friendlyanon, line 293: While we're here, this should probably be using [`target_link_options`](https://cmake.org/cmake/help/latest/command/target_link_options.html) instead. Based on the [Apple docs](https://developer.apple.com/library/archive/qa/qa1490/_index.html), this just needs to be a linker flag.

friendlyanon (changes requested):

src/SFML/Window/CMakeLists.txt
ChrisThrasher: I tried out `target_link_options` and got the same runtime failure as before. Good idea to link to the docs though. I added that.

ChrisThrasher:

JonnyPtn:
Seems a good opportunity to add a test to catch this regression? Even if it's disabled on linux (what is it that fails?)

ChrisThrasher:
> Seems a good opportunity to add a test to catch this regression?

Once #2302 is done, that may be possible. I hope by adding some basic tests for RenderWindow it will implicitly catch this bug.

eXpl0it3r (approved):

PR #2302, opened 2022-12-16
Enable testing types which require the presence of a display

ChrisThrasher:
## Description

Certain SFML types like `sf::Texture` require a display to be present to even construct an instance. This means we cannot write unit tests for them on headless machines as it the case with the Linux CI runners. This PR uses `xvfb-run` to run a dummy X11 server that lets us construct such types and run the code in CI.

Note that we don't yet have a solution for the DRM CI jobs so I'm disabling tests on those builds. This solution is not ideal but it is pragmatic. It's hard to imagine tests failing on those jobs but not elsewhere so I think it's a reasonable tradeoff to make to let us increase the Graphics module code coverage significantly.

I went ahead and added some basic `sf::Texture` tests to prove that this works. Basically all Graphics module types without runtime tests are candidates for unit testing now!

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -157,14 +157,10 @@ if(SFML_ENABLE_COVERAGE AND OpenCppCoverage_FOUND)
     string(REPLACE "/" "\\" COVERAGE_SRC "${PROJECT_SOURCE_DIR}/src")
     string(REPLACE "/" "\\" COVERAGE_INCLUDE "${PROJECT_SOURCE_DIR}/include")
 
-    add_custom_command(TARGET runtests
-                       COMMENT "Run tests"
-                       POST_BUILD COMMAND "${OpenCppCoverage_BINARY}" ARGS --quiet --export_type cobertura:${PROJECT_BINARY_DIR}/coverage.out --cover_children --excluded_modules "${COVERAGE_EXCLUDE}" --sources "${COVERAGE_SRC}" --sources "${COVERAGE_INCLUDE}" -- "${CMAKE_CTEST_COMMAND}" --output-on-failure -C $<CONFIG>
-                       VERBATIM)
-else()
-    # Run tests without a coverage runner
-    add_custom_command(TARGET runtests
-                       COMMENT "Run tests"
-                       POST_BUILD COMMAND "${CMAKE_CTEST_COMMAND}" --output-on-failure -C $<CONFIG>
-                       VERBATIM)
+    set(COVERAGE_PREFIX ${OpenCppCoverage_BINARY} ARGS --quiet --export_type cobertura:${PROJECT_BINARY_DIR}/coverage.out --cover_children --excluded_modules "${COVERAGE_EXCLUDE}" --sources "${COVERAGE_SRC}" --sources "${COVERAGE_INCLUDE}" --)
 endif()
+
+add_custom_command(TARGET runtests
+                   COMMENT "Run tests"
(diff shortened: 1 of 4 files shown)


ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

eXpl0it3r:
Graphical tests! 🥳 

But I don't think that's the correct approach chosen here. Yes, it allows us and our CI to operate correctly, but it doesn't really consider the rest of the user base:

- Everyone who runs the test suite on their dev machine and not as part of a CI, doesn't actually need `xvfb`, because their graphics driver will happily render stuff
- We're making `xvfb` and `xvfb-run` a new dependency for running SFML tests, which might not always be available
- While `xvfb` is only required for a certain type of tests, we require it for all tests

I believe we should have two options:

- One to enable/disable "headless mode", which then utilizes xvfb
- One to enable/disable any graphical tests, which then also allows to selectively only disable those tests e.g. for DRM or FreeBSD

---

Unfortunately FreeBSD doesn't seem to have a `xvfb-run`, but as far as I know `xvfb-run` is mostly just a wrapper. In theory we could wrap it, but I think it might not necessarily be worth the effort, so I'd just leave it out. (According to some source, one would need to install the package `pgk_add -r xorg-vfbserver`).

ChrisThrasher:

ChrisThrasher:

test/CMakeLists.txt
```diff
@@ -79,15 +81,19 @@ set(GRAPHICS_SRC
...
     Graphics/Transformable.test.cpp
     Graphics/Vertex.test.cpp
     Graphics/VertexArray.test.cpp
     Graphics/VertexBuffer.test.cpp
     Graphics/View.test.cpp
 )
-sfml_add_test(test-sfml-graphics "${GRAPHICS_SRC}" SFML::Graphics)
+if(SFML_BUILD_OPENGL_TESTS)
+    set(EXTRA_GRAPHICS_SRC
+        Graphics/Texture.test.cpp
```
eXpl0it3r, line 93: I guess there's a discussion point of whether we'd need dedicated classes, as the asserts could for example be executed on non-graphical instances as well.

ChrisThrasher: Instead of selectively building different .cpp files we could have
```cmake
if(SFML_BUILD_DISPLAY_TESTS)
    target_compile_definitions(sfml-test-graphics PRIVATE SFML_BUILD_DISPLAY_TESTS)
endif()
```
and
```cpp
#if SFML_BUILD_DISPLAY_TESTS
// All the stuff that requires a display
#endif
```
That would ensure that all headers and their respective `static_asserts` always get built. I'm on the fence about whether this is worth it.

ChrisThrasher: This was not hard to implement so I switched to a different method which ensures these .cpp files always get built.

ChrisThrasher:
I figured out how we can configure Doctest to build all tests all the time but selectively not run certain tests. Now we have the peace of mind that all test code is always being compiled.

ChrisThrasher:

ChrisThrasher:
A pragmatic short-term option would be to only enable these tests on macOS and Windows. This will still give us substantial test coverage but let us temporarily avoid having to modify the Buildbot. We can save that work for later while immediately unblocking anyone who wants to start writing more tests.

ChrisThrasher:

eXpl0it3r (approved):

ChrisThrasher:

ChrisThrasher:
Did a small commit history rewrite and a rebase. Pending CI passing we can merge this.

ChrisThrasher:

PR #2303, opened 2022-12-18
Ensure diagnostic pop occur before end of header guard

ChrisThrasher:
## Description

This was breaking unity builds in a separate project. See how the `#pragma` appeared after the end of the header guard.

Maybe we just switch to `#pragma once`...


ChrisThrasher:

JonnyPtn (approved):

eXpl0it3r (approved):

PR #2301, opened 2022-12-15
Use in-class member initializers

ChrisThrasher:
## Description

Continuing the work from #2156 and #2275.

PR #2275, opened 2022-11-19
Use in-class member initializers

ChrisThrasher:
## Description

Follows #2156 using in-class member initializers in the Window module.

PR #2304, opened 2022-12-18
Use in-class member initializers

ChrisThrasher:
## Description

Follows #2156, #2275, and #2301.


ChrisThrasher:

include/SFML/Network/UdpSocket.hpp
```diff
@@ -199,7 +199,7 @@ class SFML_NETWORK_API UdpSocket : public Socket
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    std::vector<char> m_buffer; //!< Temporary buffer holding the received data in Receive(Packet)
+    std::vector<char> m_buffer{std::vector<char>(MaxDatagramSize)}; //!< Temporary buffer holding the received data in Receive(Packet)
```
ChrisThrasher, line 202: I can't just use `{MaxDatagramSize}` since that tries to constructor a vector of size 1 with a single element of value `MaxDatagramSize`. If we switch from `char` to `std::byte` then this problem is fixed. Hopefully we can start using `std::byte` in more places soon.

Jim-Marsden (approved):

ChrisThrasher:

Jim-Marsden (approved):

kimci86 (approved):

eXpl0it3r (approved):

PR #2306, opened 2022-12-19
Changed sf::String::InvalidPos from 'inline const' to constexpr

Jim-Marsden:

----

## Description

While looking through all of the header files for an unrelated issue I noticed this and wanted to update it for both modernization reasons as well as clarity. (It's a very trivial change)

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [x] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/System.hpp>
#include <iostream>
int main() try
{
    sf::String str{};
    if(str.find("empty") == sf::String::InvalidPos) throw std::runtime_error("this should not throw");
    return 0;
    
}
catch(std::runtime_error const & e){
   std::cerr << e.what();
}
```

### DIFF
--- a/include/SFML/System/String.hpp
+++ b/include/SFML/System/String.hpp
@@ -58,7 +58,7 @@ public:
     ////////////////////////////////////////////////////////////
     // NOLINTBEGIN(readability-identifier-naming)
     /// Represents an invalid position in the string
-    static inline const std::size_t InvalidPos{std::basic_string<std::uint32_t>::npos};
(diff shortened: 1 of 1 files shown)


ChrisThrasher:

ChrisThrasher (approved):
👍🏻

eXpl0it3r (approved):

PR #2308, opened 2022-12-20
Rename `Key::Tilde` to `Key::Grave`

kimci86:
## Description

`sf::Keyboard::Key::Tilde` is actually used to represent a key which writes <code>\`</code>, and not `~`, so it makes sense not to name it `Key::Tilde`.
I suggest `Key::Grave` instead, which is consistent with the naming used for scancode `Scan::Grave`.

This PR is related to #1235.

## Tasks

* [x] Tested on Linux (X11)
* [x] Tested on Windows
* [x] Tested on macOS

~~I can test this patch on other platforms later.~~ Done.

## How to test this PR?

Running [SFML-Input](https://github.com/eXpl0it3r/SFML-Input) and pressing the key above tab or a key that writes <code>\`</code> depending on keyboard layout, checking that it shows consistent feedback.

The behavior looks strange on windows when not using US layout, but that is already the case before this patch.

### DIFF
--- a/include/SFML/Window/Keyboard.hpp
+++ b/include/SFML/Window/Keyboard.hpp
@@ -108,7 +108,7 @@ public:
         Apostrophe,   //!< The ' key
         Slash,        //!< The / key
         Backslash,    //!< The \ key
-        Tilde,        //!< The ~ key
+        Grave,        //!< The ` key
         Equal,        //!< The = key
         Hyphen,       //!< The - key (hyphen)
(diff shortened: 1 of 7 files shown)


kimci86:
I found and fixed a small bug while testing on macOS :)

ChrisThrasher (approved):

eXpl0it3r (approved):

eXpl0it3r:
Thank youuu ❤️

PR #2309, opened 2022-12-21
Update Catch latest v1.x

kimci86:
## Description

I updated my Ubuntu installation, so I now have GCC 11.3 which reports compilation errors in catch.hpp.
```
In file included from /usr/include/signal.h:328,
                 from <path>/SFML/extlibs/headers/catch.hpp:6520,
                 from <path>/SFML/test/src/CatchMain.cpp:2:
<path>/SFML/extlibs/headers/catch.hpp:6543:33: error: size of array ‘altStackMem’ is not an integral constant-expression
 6543 |         static char altStackMem[SIGSTKSZ];
      |                                 ^~~~~~~~
<path>/SFML/extlibs/headers/catch.hpp:6594:45: error: size of array ‘altStackMem’ is not an integral constant-expression
 6594 |     char FatalConditionHandler::altStackMem[SIGSTKSZ] = {};
      |                                             ^~~~~~~~

```
Updating catch solves it.

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

CI

### DIFF
--- a/extlibs/headers/catch.hpp
+++ b/extlibs/headers/catch.hpp
@@ -1,6 +1,6 @@
 /*
  *  Catch v1.12.2
- *  Generated: 2018-05-14 15:10:01.112442
+ *  Generated: 2022-12-21 23:22:34.433382
  *  ----------------------------------------------------------
  *  This file has been merged from multiple headers. Please don't edit it directly
  *  Copyright (c) 2012 Two Blue Cubes Ltd. All rights reserved.
@@ -214,7 +214,7 @@
 ////////////////////////////////////////////////////////////////////////////////
 
 // Use variadic macros if the compiler supports them
-#if ( defined _MSC_VER && _MSC_VER > 1400 && !defined __EDGE__) || \
+#if ( defined _MSC_VER && _MSC_VER >= 1400 && !defined __EDGE__) || \
     ( defined __WAVE__ && __WAVE_HAS_VARIADICS ) || \
     ( defined __GNUC__ && __GNUC__ >= 3 ) || \
     ( !defined __cplusplus && __STDC_VERSION__ >= 199901L || __cplusplus >= 201103L )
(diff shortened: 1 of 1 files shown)


ChrisThrasher:
This is an unfortunate situation because Catch2 v2 requires C++11 and *technically* that's not allowed on 2.6.x. I think this is a pragmatic solution because (1) who is still using a C++03-only compiler these days, (2) who cares if the tests need a newer language standard, and (3) this will fix building the SFML 2 tests on ARM Macs which is currently a sore spot for me whenever I switch over to that branch.

If the team is cool with SFML 2 having a soft requirement on C++11 then I'm also cool with this.

kimci86:
> Catch2 v2 requires C++11

Good to know! We currently have v1.12.2. I thought that was a small update because I read v2.12.2 :facepalm:
CI seems to agree as Xcode doesn't like `noexcept`.

kimci86:
I will try with latest version from Catch1.x branch, which includes a bugfix for what I was encountering.

extlibs/headers/catch.hpp
```diff
@@ -1,6 +1,6 @@
 /*
  *  Catch v1.12.2
```
ChrisThrasher, line 2: Have these fixes not been released yet?

kimci86: Nope. See: https://github.com/catchorg/Catch2/compare/v1.12.2...Catch1.x

kimci86: For reference, I ran the python2 script `generateSingleHeader.py` to generate the single header.

ChrisThrasher:

ChrisThrasher (approved):
This fixes both the GCC error you noted and building tests on ARM Macs. Thanks!

eXpl0it3r (approved):

Issue #2124, opened 2022-06-08
Calling Window.setSize() while resizing makes the window grow on Linux


Chnossos:
## Calling Window.setSize() while resizing makes the window grow

[[Original forum thread](https://en.sfml-dev.org/forums/index.php?topic=28276)]

I want to restrict the size of the window so it doesn't shrink more than a certain size.
Using the following code, when I start to resize horizontally (mouse button pressed then mouse move) the window will grow in height until it reaches the top of the screen or until I release the resize handle (mouse button released).

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode(1280, 720, sf::VideoMode::getDesktopMode().bitsPerPixel), "W");

    while (window.isOpen())
    {
        sf::Event event;
        while (window.pollEvent(event))
        {
            /**/ if (event.type == sf::Event::Closed)
                window.close();
            else if (event.type == sf::Event::Resized)
            {
                if (event.size.width < 1280 || event.size.height < 720)
                    window.setSize({ 1280, 720 });
            }
        }
        window.clear(sf::Color::White);
        window.display();
    }
}
```

## Your environment

* Fedora 32 (Wayland)
* SFML 2.5.1 (conan)
* `g++ (GCC) 10.3.1 20210422 (Red Hat 10.3.1-1)`
* No special compiler flags used

## Steps to reproduce

conanfile.txt
```ini
[requires]
sfml/2.5.1

[generators]
cmake
```

CMakeLists.txt
```cmake
cmake_minimum_required(VERSION 3.0)
project(sfml-resize-bug)

include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
conan_basic_setup(TARGETS)

add_executable(${PROJECT_NAME} main.cpp)
conan_target_link_libraries(${PROJECT_NAME})
```

Compile, build & launch
```bash
$ conan install -if build .
$ cmake -B build
$ cmake --build build
$ ./build/bin/sfml-resize-bug
```

## Expected behavior

Window should not shrink more than size specified in `window.setSize({ ... })` call. It should not grow on its own either.

## Actual behavior

![Screencast from 08-06-22 13 22 22](https://user-images.githubusercontent.com/7984363/172608310-f2d1078a-a979-4ddd-9d3a-c1477da9968e.gif)

EtoileScintillante:
I tried out your code and  first it would not compile:

`
[...]

PR #2316, opened 2022-12-25
[2124] fix. Window.setSize() while resizing causes extraneous behavior

dogunbound:
## Description

This fix forces XResizeWindow to occur after the user is done resizing the window. If you use the test code I have provided, you will see that setSize is called after a resize event is done. Before this commit, setSize would run XResizeWindow, but since the user's mouse was still holding onto the window, the window would resize to whatever size the user's mouse is at. With my fix, XResizeWindow is GUARENTEED to occur once the user is done dragging the window.  

Here is a video demonstrating my fix working: https://youtu.be/mLMVFbeuRIs 

This PR is related to the issue #2124 

## Tasks

X11 LINUX ISSUE ONLY
* [ Works] Tested on Linux
----------INVALID TESTS----------
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

After you install SFML, go ahead and compile, link, and run the code below. You will see that the resize event is seemingly not called.

Install my fix, compile, link, and run the code below. You will see the resize event being called and resizing the window.

```cpp
#include <SFML/Graphics.hpp>
#include <SFML/Window.hpp>
#include <SFML/System.hpp>

int main() {

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
            if (event.type == sf::Event::Resized) {
                const sf::Vector2u window_size(event.size.width, event.size.height);

                if (window_size.x < 200) {
                    window.setSize(sf::Vector2u(200, window_size.y));
                }
                if (window_size.y < 200) {
                    window.setSize(sf::Vector2u(window_size.x, 200));
                }
            }
        }

        window.clear();
        window.display();
    }

    return 0;
}

```

### DIFF
--- a/src/SFML/Window/Unix/WindowImplX11.cpp
+++ b/src/SFML/Window/Unix/WindowImplX11.cpp
@@ -495,6 +495,7 @@ m_hiddenCursor   (0),
 m_lastCursor     (None),
 m_keyRepeat      (true),
 m_previousSize   (-1, -1),
+m_resizeOccuredDuringSetSize(false),
 m_useSizeHints   (false),
 m_fullscreen     (false),
 m_cursorGrabbed  (false),
@@ -546,6 +547,7 @@ m_hiddenCursor   (0),
 m_lastCursor     (None),
 m_keyRepeat      (true),
 m_previousSize   (-1, -1),
+m_resizeOccuredDuringSetSize(false),
 m_useSizeHints   (false),
 m_fullscreen     ((style & Style::Fullscreen) != 0),
 m_cursorGrabbed  (m_fullscreen),
@@ -981,7 +983,12 @@ void WindowImplX11::setSize(const Vector2u& size)
         XFree(sizeHints);
     }
 
-    XResizeWindow(m_display, m_window, size.x, size.y);
+    m_resizeOccuredDuringSetSize = false;
+    while(getSize() != size && !m_resizeOccuredDuringSetSize)
+    {
+        processEvents();
+        XResizeWindow(m_display, m_window, size.x, size.y);
+    }
     XFlush(m_display);
 }
 
@@ -1837,6 +1844,7 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
             // ConfigureNotify can be triggered for other reasons, check if the size has actually changed
             if ((windowEvent.xconfigure.width != m_previousSize.x) || (windowEvent.xconfigure.height != m_previousSize.y))
             {
+                m_resizeOccuredDuringSetSize = true;
                 Event event;
                 event.type        = Event::Resized;
                 event.size.width  = static_cast<unsigned int>(windowEvent.xconfigure.width);
--- a/src/SFML/Window/Unix/WindowImplX11.hpp
+++ b/src/SFML/Window/Unix/WindowImplX11.hpp
@@ -300,25 +300,26 @@ private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    ::Window   m_window;         ///< X identifier defining our window
-    ::Display* m_display;        ///< Pointer to the display
-    int        m_screen;         ///< Screen identifier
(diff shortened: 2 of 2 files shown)


ChrisThrasher:

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -981,7 +983,12 @@ void WindowImplX11::setSize(const Vector2u& size)
         XFree(sizeHints);
     }
 
-    XResizeWindow(m_display, m_window, size.x, size.y);
+    m_resizeOccuredDuringSetSize = false;
+    while(getSize() != size && !m_resizeOccuredDuringSetSize)
+    {
+        processEvents();
+        XResizeWindow(m_display, m_window, size.x, size.y);
+    }
```
eXpl0it3r, line 991: Spinning loop with X-calls is unlikely the way to go 🤔

dogunbound: Hmmm. I guess I'll have to tinker with other ideas later. I tried setting the mouse position in the window.setSize function, but that unfortunately doesn't really solve the issue.

dogunbound: I added a timeout to force quit this loop. Do another code review on the new code.

dogunbound:
SDL2 does the same thing. They loop X calls for the window set size function. https://github.com/libsdl-org/SDL/blob/main/src/video/x11/SDL_x11window.c 

Scroll down to X11_SetWindowSize and at the end of the function you will see them looping X calls.

The end result is similar to what I am doing. You will notice that if you drag and drop an sdl window quickly enough, a forced resize can occur after a resize event has been called.

This is the min reproducible with sdl:
```c++
// SDL2 Hello, World!
// This should display a white screen for 2 seconds
// compile with: clang++ main.cpp -o hello_sdl2 -lSDL2
// run with: ./hello_sdl2
#include <SDL2/SDL.h>
#include <stdio.h>

#define SCREEN_WIDTH 640
#define SCREEN_HEIGHT 480

int main(int argc, char* args[]) {
    SDL_Window* window = NULL;
    SDL_Surface* screenSurface = NULL;
    if (SDL_Init(SDL_INIT_VIDEO) < 0) {
        fprintf(stderr, "could not initialize sdl2: %s\n", SDL_GetError());
        return 1;
    }
    window = SDL_CreateWindow(
            "hello_sdl2",
            SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED,
            SCREEN_WIDTH, SCREEN_HEIGHT,
            SDL_WINDOW_RESIZABLE
            );
    if (window == NULL) {
        fprintf(stderr, "could not create window: %s\n", SDL_GetError());
        return 1;
    }
    bool game_is_still_running = true;
    while (game_is_still_running) {
        SDL_Event event;
        while (SDL_PollEvent(&event)) {  // poll until all events are handled!
                                         // decide what to do with this event.
            switch (event.type) {
                case SDL_QUIT:
                    game_is_still_running = false;
                    break;
                case SDL_WINDOWEVENT: 
                    switch (event.window.event) {
                        case SDL_WINDOWEVENT_RESIZED:
                            {
                                const int new_w = event.window.data1, new_h = event.window.data2;
                                if (new_w < 200) {
                                    SDL_SetWindowSize(window, 200, new_h);
                                }
                                if (new_h < 200) {
                                    SDL_SetWindowSize(window, new_w, 200);
                                }
                                break;
                            }
                        default:
                            break;
                    }
                    break;
                default:
                    break;
            }
        }

        // update game state, draw the current frame
        screenSurface = SDL_GetWindowSurface(window);
        SDL_FillRect(screenSurface, NULL, SDL_MapRGB(screenSurface->format, 0xFF, 0xFF, 0xFF));
        SDL_UpdateWindowSurface(window);
    }
    SDL_DestroyWindow(window);
    SDL_Quit();
    return 0;
}
```

eXpl0it3r:

PR #2318, opened 2022-12-26
Update x11 sf::Cursor::Hand to be consistent across OS's

danieljpetersen:
## Description

It's my understanding that sf::Cursor::Hand is intended to be the mouse cursor indicating a clickable item. I am not able to test what it looks like on macOS, but on Windows sf::Cursor::Hand is indeed the clickable hand icon. On x11, however, there are two hand icons -- xc_hand1, and xc_hand2. 

SFML is currently using xc_hand1, which is an open hand that is usually used in the context of dragging windows. This pr changes sf::Cursor::Hand to use xc_hand2 as that's the clickable indicator. 

Please see the image below

hand

It's possible that there are some themes out there which use xc_hand1 as the clickable indicator, but the default themes for every major distro I looked at (adwaita, ubuntu, fedora, breeze kde, linux mint, xubuntu) all use hand1(open hand) hand2(clickable). 

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>

int main()
{

    sf::RenderWindow window(sf::VideoMode({600, 800}, 32), "Cursor Test");

    sf::Cursor cursor;
    if (cursor.loadFromSystem(sf::Cursor::Hand))

    while (window.isOpen())
    {
        {
            {
                window.close();
            }
        }

        window.clear();
        window.display();
    }
}
```

### DIFF
--- a/src/SFML/Window/Unix/CursorImpl.cpp
+++ b/src/SFML/Window/Unix/CursorImpl.cpp
@@ -172,7 +172,7 @@ bool CursorImpl::loadFromSystem(Cursor::Type type)
         case Cursor::Arrow:           shape = XC_arrow;               break;
         case Cursor::Wait:            shape = XC_watch;               break;
         case Cursor::Text:            shape = XC_xterm;               break;
-        case Cursor::Hand:            shape = XC_hand1;               break;
+        case Cursor::Hand:            shape = XC_hand2;               break;
         case Cursor::SizeHorizontal:  shape = XC_sb_h_double_arrow;   break;
         case Cursor::SizeVertical:    shape = XC_sb_v_double_arrow;   break;
         case Cursor::SizeLeft:        shape = XC_left_side;           break;


ChrisThrasher:

ChrisThrasher:
Can confirm macOS uses the 2nd style cursor

danieljpetersen:
I've also checked SDL2 and GLFW, and they both use hand2. 

If accepted can we also merge this into 2.6 branch as well?

ChrisThrasher:
> If accepted can we also merge this into 2.6 branch as well?

I don't see why not. In that case it'd be easier for this PR to target 2.6.x then it will automatically land in `master` after we do our next backmerge.

danieljpetersen:
> > If accepted can we also merge this into 2.6 branch as well?
> 
> I don't see why not. In that case it'd be easier for this PR to target 2.6.x then it will automatically land in `master` after we do our next backmerge.

Is there a way I can change the target branch, or do I have to resubmit?

ChrisThrasher (approved):

ChrisThrasher:

eXpl0it3r (approved):

eXpl0it3r:
Thanks for spotting and reporting this!

PR #2321, opened 2022-12-27
Remove defaulted or empty, non-virtual destructors #2312

### LINKED ISSUE #2312: Remove defaulted or empty, non-virtual destructors
## Subject of the issue

I'm not quite sure how they ended up like but a few types have non-virtual destructors that do nothing. The existence of these destructors implicitly disables move semantics even if they're `= default`'d so it's best to remove them entirely. Here are the types where I found this happening. There may be more.

- `sf::Font`
- `sf::SocketSelector`
- `sf::Cursor`
- `sf::Font`

benjamintli:

----

## Description

resolves #2312 

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android


ChrisThrasher:

include/SFML/Graphics/Font.hpp
```diff
@@ -95,14 +95,6 @@ class SFML_GRAPHICS_API Font
...
     Font& operator=(Font&&) noexcept;
 
-    ////////////////////////////////////////////////////////////
-    /// \brief Destructor
-    ///
-    /// Cleans up all the internal resources used by the font
-    ///
-    ////////////////////////////////////////////////////////////
-    ~Font();
-
```
ChrisThrasher, line 105: A CI job is failing because we removed this declaration. It's hard to explain and I don't totally understand it myself but because we use forward declarations to speed up compilation in other parts of the library, it forces us to declare destructors even if we want them to be default.

benjamintli: interesting....let me add this back in then

ChrisThrasher:

include/SFML/Graphics/Font.hpp
kimci86: That is because, on Android, `Font` contains a `std::unique_ptr` to `priv::ResourceStream` which is an incomplete type at the moment of `Font` class definition.

ChrisThrasher (approved):

src/SFML/Graphics/Shader.cpp
```diff
@@ -989,12 +989,6 @@ Shader::CurrentTextureType Shader::CurrentTexture;
 Shader::Shader() = default;
 
 
-////////////////////////////////////////////////////////////
-Shader::~Shader()
-{
-}
```
eXpl0it3r, line 995: What about the declaration in the header file?

ChrisThrasher: I'm not actually sure how CI passes with this removed. I suppose it's something related to how you don't need to define symbols that don't get used and the configurations that use this half of the `Shader` implementation never actually try to use a `Shader`. I think this needs to be restored.

ChrisThrasher:
https://github.com/SFML/SFML/issues/2312#issuecomment-1366776905

@kimci86 made a great point in the issue thread how this type needs an explicitly declared destructor given how it's being implemented. I was able to corroborate this by simply attempting to construct an instead of `sf::Cursor` and see the build fail. I think coincidentally this passed CI because an `sf::Cursor` is never constructed (and thus destroyed) anywhere in the repo itself so we were able to omit a destructor without that issue being detected.

ChrisThrasher:

ChrisThrasher:
I'm closing this PR since we can't easily remove these destructors after all.

PR #2326, opened 2023-01-02
Simplify in-class initializer expressions

ChrisThrasher:
## Description

While developing #2310 we realized that our member initializer expressions could be simpler and shorter for many types. This PR applies that knowledge to the rest of SFML.


ChrisThrasher:

eXpl0it3r (approved):

PR #2329, opened 2023-01-04
Enable move semantics for `sf::InputSoundFile` and `sf::OutputSoundFile`

ChrisThrasher:
## Description

The three commits in this PR tell the story for how I discovered this. First I realized that explicitly deleting the copy operations was unnecessarily since these types are implemented in terms of unique pointers whose presence implicitly disable copy semantics.

Then I realized that the destructors were unnecessary so I made those default.

After that it was apparent that these types should be default moveable. The only thing holding that back were some forward declarations which necessitated declaring destructors. These were presumably in place for build time optimization reasons. Getting rid of those forward declarations let me remove the destructors which in turn implicitly enabled move semantics. The type trait tests also reflect that this is now true.

I'll probably want this PR squashed but for now I want the separate commits to aid in code review.


ChrisThrasher:

ChrisThrasher:

include/SFML/Audio/InputSoundFile.hpp
```diff
@@ -206,7 +206,7 @@ class SFML_AUDIO_API InputSoundFile
     /// \brief Deleter for input streams that only conditionally deletes
     ///
     ////////////////////////////////////////////////////////////
-    struct StreamDeleter
+    struct SFML_AUDIO_API StreamDeleter
```
kimci86, line 209: Exporting private helper does seem right :thinking:

ChrisThrasher: Seems like it fixed the issue. Can't understand why though.

kimci86: Not sure but I guess that a compiler-generated destructor can be inlined. The compiler-generated implementation inlined in client code could not find StreamDeleter.
A cleaner solution IMHO is to declare the destructor and provide a default implementation in cpp file as before. To enable move semantics, declaring default move constructor and move assignment operator should work.

ChrisThrasher: > To enable move semantics, declaring default move constructor and move assignment operator should work.

The whole point of this PR was to avoid needing to do that though.

ChrisThrasher:

Jim-Marsden (approved):

include/SFML/Audio/InputSoundFile.hpp
kimci86: On second thought, exported classes have their private methods visible to the linker. With that in mind, it make sense to have this helper class exported too. It just felt weird to see it explicitly at first.

kimci86 (approved):

src/SFML/Audio/OutputSoundFile.cpp
```diff
@@ -36,14 +36,6 @@ namespace sf
 OutputSoundFile::OutputSoundFile() = default;
 
 
-////////////////////////////////////////////////////////////
-OutputSoundFile::~OutputSoundFile()
-{
-    // Close the file in case it was open
-    close();
-}
```
eXpl0it3r, line 44: Why is it okay to not call `close()` at destruction?

kimci86: `close()` only deletes an `std::unique_ptr`'s managed object, which is what a default constructor would do anyways.

ChrisThrasher: Kimci is right. The [corresponding commit](https://github.com/SFML/SFML/pull/2329/commits/cacba32491024bb9bbc433e4c4db56303f642b29) described my rationale.

eXpl0it3r (approved):

Issue #2314, opened 2022-12-24
Remove unnecessary `return EXIT_SUCCESS;`


ChrisThrasher:
## Subject of the issue

The C++ specification requires that `0` be returned if you reach the end of `main` so you never have to add `return 0;` or `return EXIT_SUCCESS;` as the last expression in your program. This pattern can be removed which also entails potentially removing `#include <cstdlib>` if this is the last place where that header is being used.

maddy020:
I want to work over this issue 
Please assign it to me

ChrisThrasher:
You don’t need to ask for permission

eXpl0it3r:
Implemented with #2330

PR #2330, opened 2023-01-04
Removed return EXIT_SUCCESS and cstdlib

HelloShiv:
* [X] Has this change been discussed on [the forum](https://github.com/SFML/SFML/issues/2314) or in an issue before?

----

## Description

removed exit_suceess and cstdlib.h

This PR is related to the issue #2314


examples/island/Island.cpp
```diff
@@ -10,7 +10,6 @@
 #include <algorithm>
 #include <array>
 #include <cmath>
-#include <cstdlib>
```
ChrisThrasher, line 13: This ought to stay since this file still uses `EXIT_FAILURE`

examples/opengl/OpenGL.cpp
```diff
@@ -5,7 +5,6 @@
 #include <SFML/Graphics.hpp>
 
 #include <array>
-#include <cstdlib>
```
ChrisThrasher, line 8: This ought to stay since this file still uses `EXIT_FAILURE`

examples/window/Window.cpp
```diff
@@ -13,7 +13,6 @@
 #endif
 
 #include <array>
-#include <cstdlib>
```
ChrisThrasher, line 16: This ought to stay since this file still uses `EXIT_FAILURE`

ChrisThrasher (changes requested):

kimci86 (commented):
You can remove cstdlib include directive from `examples/sockets/Sockets.cpp` if I understand correctly.

examples/sockets/Sockets.cpp
```diff
@@ -54,6 +54,4 @@ int main()
     std::cout << "Press enter to exit..." << std::endl;
     std::cin.ignore(10000, '\n');
     std::cin.ignore(10000, '\n');
-
-    return EXIT_SUCCESS;
```
ChrisThrasher, line 58: You can remove <cstdlib> from this file

examples/window/Window.cpp
```diff
@@ -182,6 +182,4 @@ int main()
         // Finally, display the rendered frame on screen
         window.display();
     }
-
-    return EXIT_SUCCESS;
```
ChrisThrasher, line 186: I believe this has to stay because it causes the iOS build to fail. For those builds the main function is converted into `sfmlMain` which because it's just a normal function at that point loses this special property where we can omit the return value. I think the same goes for two other examples but CI will eventually catch those errors.

examples/sockets/Sockets.cpp
HelloShiv: I actually removed it  in c184656

ChrisThrasher:

ChrisThrasher:
Screenshot 2023-01-04 at 11 25 53 AM

Tennis.cpp is another file built by iOS where we can't remove EXIT_SUCCESS

ChrisThrasher:

ChrisThrasher:
Screenshot 2023-01-04 at 12 36 46 PM

I believe this is the last file where we have to re-add EXIT_SUCCESS and then CI will pass and I can approve this!

ChrisThrasher:

ChrisThrasher (approved):
LGTM

eXpl0it3r (approved):

eXpl0it3r:
Thanks a lot for this change! 🙂

PR #2332, opened 2023-01-05
Prevent constructing `sf::Text` with a temporary `sf::Font`

ChrisThrasher:
## Description

By deleting this constructor overload, it fails to compile if you pass a temporary font to this parameter slot. That includes code like

```cpp
  sf::Text text("", sf::Font());
```

but more importantly it prohibits code like this

```cpp
  sf::Font getFont()
  {
    sf::Font font;
    // load a font...
    return font;
  }

  sf::Text text("", getFont());
```

Credit to @JonnyPtn for the idea.

---

We ought to copy this idea to other APIs that do similar things taking the address of a const reference.


ChrisThrasher:

kimci86:
Nice, the same idea could be used for Sprite and Sound constructors too.

About "Co-authored-by", I think it needs to be at the end of the commit message if this blog post is still accurate https://github.blog/2018-01-29-commit-together-with-co-authors/#how-it-works.

kimci86:
I just realized `= delete` can be used on any method, not only constructors or special methods.
So this approach could be applied to `setFont` too to make it safer.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

eXpl0it3r (approved):

ChrisThrasher:

eXpl0it3r (approved):

ChrisThrasher:
> Nice, the same idea could be used for Sprite and Sound constructors too.

https://github.com/ChrisThrasher/SFML/tree/lifetime_safety

I also found somewhere in `sf::Shader` where this trick would come in handy.

JonnyPtn (approved):

PR #2337, opened 2023-01-06
Fix bug where clang-tidy CI job always passes

ChrisThrasher:
## Description

The CMake script itself isn't returning non-zero when run-clang-tidy fails meaning clang-tidy errors are going undetected in CI.


ChrisThrasher:

eXpl0it3r (approved):

PR #2343, opened 2023-01-07
Added modernize-use-nullptr check

gnawme:

----

## Description

Adds `modernize-use-nullptr` clang-tidy check.

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::Vector2u modeSize{200, 200};
    sf::RenderWindow window(sf::VideoMode(modeSize), "SFML works!");
    sf::CircleShape shape(100.f);

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
        }

        window.clear();
        window.draw(shape);
        window.display();
    }

    return 0;
}

g++ main.cpp -o sfml-app -I ./include/ -L./build/lib -lsfml-audio -lsfml-graphics -lsfml-system -lsfml-window

./sfml-app
```

### DIFF
--- a/examples/vulkan/Vulkan.cpp
+++ b/examples/vulkan/Vulkan.cpp
@@ -209,36 +209,36 @@ public:
     maxFramesInFlight(2),
     currentFrame(0),
     swapchainOutOfDate(false),
-    instance(0),
-    debugReportCallback(0),
-    surface(0),
-    gpu(0),
+    instance(),
+    debugReportCallback(),
+    surface(),
+    gpu(),
     queueFamilyIndex(-1),
-    device(0),
-    queue(0),
+    device(),
+    queue(),
     swapchainFormat(),
     swapchainExtent(),
-    swapchain(0),
+    swapchain(),
     depthFormat(VK_FORMAT_UNDEFINED),
-    depthImage(0),
-    depthImageMemory(0),
-    depthImageView(0),
-    vertexShaderModule(0),
-    fragmentShaderModule(0),
-    descriptorSetLayout(0),
-    pipelineLayout(0),
-    renderPass(0),
-    graphicsPipeline(0),
-    commandPool(0),
-    vertexBuffer(0),
-    vertexBufferMemory(0),
-    indexBuffer(0),
-    indexBufferMemory(0),
-    textureImage(0),
-    textureImageMemory(0),
-    textureImageView(0),
-    textureSampler(0),
-    descriptorPool(0)
+    depthImage(),
+    depthImageMemory(),
+    depthImageView(),
+    vertexShaderModule(),
+    fragmentShaderModule(),
+    descriptorSetLayout(),
+    pipelineLayout(),
+    renderPass(),
+    graphicsPipeline(),
+    commandPool(),
+    vertexBuffer(),
+    vertexBufferMemory(),
+    indexBuffer(),
+    indexBufferMemory(),
+    textureImage(),
+    textureImageMemory(),
+    textureImageView(),
+    textureSampler(),
+    descriptorPool()
     {
         // Vulkan setup procedure
         if (vulkanAvailable)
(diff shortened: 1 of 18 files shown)


examples/vulkan/Vulkan.cpp
```diff
@@ -17,7 +17,8 @@
 #include <limits>
 #include <vector>
 
-
+// Disable clang-tidy for this file
+// NOLINTBEGIN
```
ChrisThrasher, line 21:
```suggestion
// NOLINTBEGIN(modernize-use-nullptr)
```
As written, _all_ clang-tidy checks get silence for this block.

examples/vulkan/Vulkan.cpp
```diff
@@ -2658,3 +2669,4 @@ int main()
 
     return EXIT_SUCCESS;
 }
+// NOLINTEND
```
ChrisThrasher, line 2672:
```suggestion
// NOLINTEND(modernize-use-nullptr)
```

src/SFML/Graphics/Font.cpp
```diff
@@ -129,7 +129,7 @@ m_fontHandles(copy.m_fontHandles),
 m_isSmooth(copy.m_isSmooth),
 m_info(copy.m_info),
 m_pages(copy.m_pages),
-m_pixelBuffer(copy.m_pixelBuffer)
+m_pixelBuffer(copy.m_pixelBuffer.begin(), copy.m_pixelBuffer.end())
```
ChrisThrasher, line 132: How does this relate to the nullptr checks you added?

src/SFML/Graphics/Font.cpp
```diff
@@ -521,7 +523,7 @@ Font& Font::operator=(const Font& right)
 #ifdef SFML_SYSTEM_ANDROID
     std::swap(m_stream, temp.m_stream);
 #endif
-
+    }
```
ChrisThrasher, line 526: More non-nullptr changes

src/SFML/Window/OSX/SFWindowController.mm
```diff
@@ -99,7 +99,7 @@ - (id)initWithWindow:(NSWindow*)window
     {
         m_window        = nil;
         m_oglView       = nil;
-        m_requester     = 0;
+        m_requester     = nullptr;
```
ChrisThrasher, line 102: I don't know Objective-C++ well but do we want to use `nil` here?

examples/vulkan/Vulkan.cpp
gnawme: I don't know why, but clang-tidy was changing non-pointer 0-values to `nullptr` in this file, so I disabled it after then initial conversion

src/SFML/Graphics/Font.cpp
gnawme: Dang cherry-pick...

examples/vulkan/Vulkan.cpp
ChrisThrasher: I played around with this yesterday and it appears that on Visual Studio 2019, the Vulkan API itself is expecting an int instead of a pointer. Perhaps we use `{}` for these arguments so that we can avoid disabling this check but still make it work on both builds.

src/SFML/Window/OSX/SFWindowController.mm
gnawme: Yes, not sure why clang-tidy flagged that

examples/vulkan/Vulkan.cpp
gnawme: I did that for a couple, I'll try using `{}` in the constructor initialization list, which is where most of the problematic ones were

ChrisThrasher: You can replace `(0)` in the initializer list with `()` to default initialize the members. That will give you `0` when the members are integers and `nullptr` when the members are pointers.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:
I think I can approve this PR but I first want someone to review the Objective-C++ changes first since I don't understand `nil` versus `nullptr` all that well.

gnawme:
> I think I can approve this PR but I first want someone to review the Objective-C++ changes first since I don't understand `nil` versus `nullptr` all that well.

According to [this](https://stackoverflow.com/a/557613/124755), `nil` should only be used for objects of type `id`; I'll take a look when I get a chance.

ChrisThrasher:
> should only be used for objects of type `id`;

I read that too but "should only be used for objects of type id" isn't something I'm familiar with. I'm not sure what types we're using that meet that definition.

examples/vulkan/Vulkan.cpp
```diff
@@ -462,7 +461,7 @@ class VulkanExample
     void setupInstance()
     {
         // Load bootstrap entry points
-        gladLoadVulkan(0, getVulkanFunction);
+        gladLoadVulkan({}, getVulkanFunction);
```
eXpl0it3r, line 464: Would `nullptr` make sense here?

examples/vulkan/Vulkan.cpp
```diff
@@ -558,7 +557,7 @@ class VulkanExample
         }
 
         // Load instance entry points
-        gladLoadVulkan(0, getVulkanFunction);
+        gladLoadVulkan({}, getVulkanFunction);
```
eXpl0it3r, line 560: Would `nullptr` make sense here?

src/SFML/Graphics/Shader.cpp
```diff
@@ -737,7 +737,7 @@ void Shader::bind(const Shader* shader)
     else
     {
         // Bind no shader
-        glCheck(GLEXT_glUseProgramObject(0));
+        glCheck(GLEXT_glUseProgramObject({}));
```
eXpl0it3r, line 740: Would `nullptr` make sense here?

eXpl0it3r (changes requested):
I feel like we should still be careful with the usage of `{}` and `nullptr`.

Saying something is "default initialized" (`{}`) carries a certain intention of "I want this thing default initialized", which doesn't always communicate the same thing as, "I want to pass a nullptr to this function".

For example in Shader.cpp `Bind no shader` is better communicated with `GLEXT_glUseProgramObject(nullptr)` than `GLEXT_glUseProgramObject({})`. One really says to set it to "nothing" (nullptr), while the other just says to set it to the default value, which only implicitly then means nullptr, but it could also be some default object.

examples/vulkan/Vulkan.cpp
ChrisThrasher: This is tricky because for whatever reason, the VS 2019 jobs fail if we try to use `nullptr` because in that configuration, these functions expect an integer instead of a pointer. I'm not sure why this is the case. `0` works since that is a valid pointer and integer value as is `{}` but using `{}` means we silence this clang-tidy check. If we go back to using zero then we just have to disable this check for this file.

eXpl0it3r: ☹️

eXpl0it3r (approved):

src/SFML/Graphics/Shader.cpp
gnawme: Not sure why `clang-tidy` is looking for `nullptr` here when the argument is a `GLuint`

ChrisThrasher:
```
/Users/thrasher/Projects/sfml/src/SFML/Graphics/Shader.cpp:740:42: error: cannot initialize a parameter of type 'GLhandleARB' (aka 'void *') with an rvalue of type 'int'
        glCheck(GLEXT_glUseProgramObject(1));
                                         ^
```
On my machine it's expecting a `void*`.

ChrisThrasher:

examples/vulkan/Vulkan.cpp
```diff
@@ -17,7 +17,6 @@
 #include <limits>
 #include <vector>
 
-
```
ChrisThrasher, line 20: Can you add this back?

ChrisThrasher (changes requested):
One tiny formatting fix but I'm otherwise ready to approve and merge this

ChrisThrasher (approved):

examples/vulkan/Vulkan.cpp
gnawme: Done, squashed the commit with the others

ChrisThrasher: I thought it was easier to just fix this myself and save you the time.

ChrisThrasher:

ChrisThrasher:
Thank you! I'd love to see `modernize-use-default-member-init` added next

Issue #2342, opened 2023-01-07
`final` specifier in derived classes


rtkid-nt:
`x86-64 gcc 12.2` with `-O3` 

Without `final` specifier:
```cpp
#include <iostream>

class Base 
{ 
public:
    virtual ~Base() = default;
    virtual void f() const = 0; 
}; 
 
class Derived : public Base 
{ 
public:
    void f() const override { std::cout << "Derived\n"; } 
};

void call_f(const Derived& d) { 
    d.f(); 
}
```
```asm
.LC0:
        .string "Derived\n"
Derived::f() const:
        mov     edx, 8
        mov     esi, OFFSET FLAT:.LC0
        mov     edi, OFFSET FLAT:_ZSt4cout
        jmp     std::basic_ostream<char, std::char_traits<char> >& std::__ostream_insert<char, std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*, long)
call_f(Derived const&):
        mov     rax, QWORD PTR [rdi]
        mov     rax, QWORD PTR [rax+16]
        cmp     rax, OFFSET FLAT:Derived::f() const
        jne     .L4
        mov     edx, 8
        mov     esi, OFFSET FLAT:.LC0
        mov     edi, OFFSET FLAT:_ZSt4cout
        jmp     std::basic_ostream<char, std::char_traits<char> >& std::__ostream_insert<char, std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*, long)
.L4:
        jmp     rax
_GLOBAL__sub_I_call_f(Derived const&):
        sub     rsp, 8
        mov     edi, OFFSET FLAT:_ZStL8__ioinit
        call    std::ios_base::Init::Init() [complete object constructor]
        mov     edx, OFFSET FLAT:__dso_handle
        mov     esi, OFFSET FLAT:_ZStL8__ioinit
        mov     edi, OFFSET FLAT:_ZNSt8ios_base4InitD1Ev
        add     rsp, 8
        jmp     __cxa_atexit
```

With `final` specifier:
```cpp
#include <iostream>

class Base 
{ 
public:
    virtual ~Base() = default;
    virtual void f() const = 0; 
}; 
 
class Derived final : public Base 
{ 
public:
    void f() const override { std::cout << "Derived\n"; } 
};

void call_f(const Derived& d) { 
    d.f(); 
}
```

```asm
.LC0:
        .string "Derived\n"
call_f(Derived const&):
[...]

PR #2344, opened 2023-01-08
Mark derived classes as `final`

rtkid-nt:
#2342

### DIFF
--- a/include/SFML/Audio/Music.hpp
+++ b/include/SFML/Audio/Music.hpp
@@ -46,7 +46,7 @@ class InputStream;
 /// \brief Streamed music played from an audio file
 ///
 ////////////////////////////////////////////////////////////
-class SFML_AUDIO_API Music : public SoundStream
+class SFML_AUDIO_API Music final : public SoundStream
 {
 public:
     ////////////////////////////////////////////////////////////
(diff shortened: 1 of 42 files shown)


ChrisThrasher:
Adding `final` to every single class removes lots of useful functionality from the library. If we are to use `final` it must be in very specific places where the benefits are provable and justify the reduced functionality of the library.

Bromeon:
> Adding `final` to every single class removes lots of useful functionality from the library. If we are to use `final` it must be in very specific places where the benefits are provable and justify the reduced functionality of the library.

It's also possible to look at it from another perspective: classes are final by default, and can be opened up if designed to be inherited. While it restricts use cases, it can prevent mistakes and guide the user to use composition instead.

I'm not saying one or the other is the correct way, but this is e.g. how we handle `explicit` constructors. We consider it a design flaw of C++ itself to have chosen the wrong default. Kotlin for example restricts inheritance and requires the `open` keyword to enable it -- and this despite being Java-compatible, a language which is _extremely_ driven by inheritance based designs.

JWStaiert:
Just $.02 from a long time c++ developer, in the general case, final should actually be the default. Unless classes are explicitly designed and (more importantly) tested for use by inheritance they frequently won’t work as expected. Moreover, code designed for composition typically performs better and can be better optimized by a compiler. Proper use of a library designed around inheritance is a much steeper learning curve.

Bromeon:
So is this decision... final? 😁

ChrisThrasher:
I'm open to adding `final` in select places where derivation is prone to cause bugs. We can also add type trait tests to enforce the presence of final and conversely to ensure that certain tests remain open to subclassing.

Issue #2347, opened 2023-01-10
Replace VulkanAPI `C` style to `C++` style


rtkid-nt:
Why you don't use [Vulkan-Hpp](https://github.com/KhronosGroup/Vulkan-Hpp)?

ChrisThrasher:
Someone else brought this up recently. I can't think of any particular reason we don't do that. I'm open to that possibility. It depends on what the implementation would look like.

rtkid-nt:
> Someone else brought this up recently. I can't think of any particular reason we don't do that. I'm open to that possibility. It depends on what the implementation would look like.

Can you assign it to me?

Bambo-Borris:
> > Someone else brought this up recently. I can't think of any particular reason we don't do that. I'm open to that possibility. It depends on what the implementation would look like.
> 
> Can you assign it to me?

It's probably simpler to just go do the work and tag this issue in your PR than expect assigning per-issue you raise for things. I don't think the assigning matters, just dive in and let's see this prospective PR!

PR #2348, opened 2023-01-10
WIP - add vulkan hpp headers

rtkid-nt:
This WIP PR  add support for using not only `C` vulkan headers, but and `C++` headers #2347 

What you think about:
1.  Replace in `Vulkan.hpp`, `VulkanImpl*.hpp` headers and code to use `vulkan.hpp` header?
Basically we can leave it as is.   
2. Add vulkancpp example?


ChrisThrasher:
Adding a quarter million lines of code to the repo feels like a bit much just to be able to use the Vulkan C++ API. I don't think I can approve such a huge influx of 3rd party code into the repo when we're hoping to get rid of 3rd party code from the repo.

Bambo-Borris:
@rtkid-nt Have you got Discord, I've noticed you're eager to contribute and it may be easier to join the discord server and have a discussion in there with some of the maintainers. There's lots to do, and you clearly have some ideas of further things that could be done so feel free to boogie on down to Discord!

PR #2358, opened 2023-01-14
Run clang-tidy on Windows in CI

ChrisThrasher:
## Description

Luckily there weren't many Windows clang-tidy errors to fix. Note that I had to use static libraries because when building shared libraries it produced a number of .rc files in the build/src directory which got analyzed by clang-tidy but failed because Windows .rc files are a bit weird and not standard C++. Add this to the list of reasons why SFML should not default to shared library. While I was at it I realized that we can enable color output which is great because color is turned off by default on macOS. That'll make it easier for developers to find errors in the console output.

These are all the configs we should need! I don't expect adding any more so from here on out we're in a great spot to add more warnings without worrying about large swaths of the library going unanalyzed.

Thanks to @Bambo-Borris who helped answer a lot of Windows-specific questions very quickly.


ChrisThrasher:

cmake/Tidy.cmake
```diff
@@ -19,8 +19,16 @@ else()
     message(FATAL_ERROR "clang-tidy version ${CMAKE_MATCH_1} is too low")
 endif()
 
+# Find Python and run-clang-tidy script
+find_package(Python 3 REQUIRED)
```
eXpl0it3r, line 23: Why do we need a python script on Windows?
Do we need this also on other OS?
Just wondering if the requirement for Python as dependency applies to all OS

ChrisThrasher: run-clang-tidy is a python script, it’s just that on macos and Linux they can be run as executable scripts whereas on windows you have to invoke python. I wanted all OSes to use the same code so now we invoke python on all OSes.

.clang-tidy
```diff
@@ -20,3 +20,4 @@ CheckOptions:
   - { key: readability-identifier-naming.ParameterCase,       value: camelBack }
 HeaderFilterRegex: '.*'
 WarningsAsErrors: '*'
+UseColor: true
```
eXpl0it3r, line 23: If it doesn't break on the CI, I think it's okay to go with this for now, but it can cause issues on certain terminals.  
But that's really more an edge case. If someone reports troubles, we can still disable it again.

eXpl0it3r (approved):

.clang-tidy
ChrisThrasher: I have no problem reverting this later if someone reports problems

PR #2359, opened 2023-01-14
Use built-in CMake support for iOS

JonnyPtn:
Now we're using more modern cmake, iOS support is first-class so we don't need to ship our own toolchain.

Practically there should be no difference except for generally being easier to work with iOS (e.g don't need to reconfigure to switch between simulator/device)


ChrisThrasher:

ChrisThrasher:

cmake/Config.cmake
```diff
@@ -39,21 +39,22 @@ elseif(CMAKE_SYSTEM_NAME MATCHES "^NetBSD$")
...
     # don't use the OpenGL ES implementation on NetBSD
     set(OPENGL_ES 0)
+elseif(${CMAKE_SYSTEM_NAME} STREQUAL "iOS")
+    set(SFML_OS_IOS 1)
+    # As we want to find packages in our extlibs folder too
+    # we need to tell CMake to search there as well
+    set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY NEVER)
+    set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE NEVER)
+    set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE NEVER)
+    set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
```
eXpl0it3r, line 49: The comment says "our extlibs folder too", but then the option is set to `NEVER`, doesn't that mean, we're exclusively searching the `extlibs` folder?

.github/workflows/ci.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         - platform: { name: MacOS, os: macos-11 }
           config: { name: Frameworks, flags: -DSFML_BUILD_FRAMEWORKS=TRUE }
         - platform: { name: MacOS, os: macos-11 }
-          config: { name: iOS, flags: -DCMAKE_TOOLCHAIN_FILE=$GITHUB_WORKSPACE/cmake/toolchains/iOS.toolchain.cmake -DIOS_PLATFORM=SIMULATOR }
+          config: { name: iOS, flags:  -DCMAKE_SYSTEM_NAME=iOS }
```
ChrisThrasher, line 41:
```suggestion
          config: { name: iOS, flags: -DCMAKE_SYSTEM_NAME=iOS }
```

CMakeLists.txt
```diff
@@ -12,15 +12,7 @@ endmacro()
...
-
-if(NOT CMAKE_OSX_SYSROOT)
-    # query the path to the default SDK, will fail on non-macOS, but it's okay.
-    execute_process(COMMAND xcodebuild -sdk macosx -version Path
-                    COMMAND head -n 1
-                    COMMAND tr -d '\n'
-                    OUTPUT_VARIABLE CMAKE_OSX_SYSROOT
-                    ERROR_QUIET)
-endif()
+sfml_set_option(CMAKE_OSX_DEPLOYMENT_TARGET "13.0" STRING "The minimal iOS version that will be able to run the built binaries. Cannot be lower than 13.0")
```
ChrisThrasher, line 15: Can we move this somewhere after the `project` call? I'd like to minimum how much code we add before `project()`.

ChrisThrasher (changes requested):
I'm not familiar with CMake for iOS but if this works then it looks like a great simplification.

CMakeLists.txt
JonnyPtn: Nope it has to be before the first call to project https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html

ChrisThrasher:

cmake/Config.cmake
JonnyPtn: Yeah good spot - originally set to BOTH but then changed to NEVER, will update comment

ChrisThrasher:
https://github.com/SFML/SFML-Buildbot/blob/0ab146c329327a85a191f716e8de41c8070f7325/buildfactories.py#L67-L69

Looks like we need to update the Buildbot too

JonnyPtn:
Think I've done all I can/should on this for now - Ended up going down a bit of a rabbit hole with our Xcode/frameworks stuff which is a bit FUBAR right now, so this just adds the tests for iOS static libraries right now and we can address the framework issues separately as there is some overlap with macOS

ChrisThrasher:

ChrisThrasher (changes requested):
Being able to get rid of that iOS toolchain and test the install interface on iOS is promising. I've got a few more clarifying questions since idk much about iOS development.

.github/workflows/ci.yml
```diff
@@ -112,7 +118,7 @@ jobs:
       shell: bash
       run: |
         cmake -S $GITHUB_WORKSPACE/test/install -B $GITHUB_WORKSPACE/test/install/build -DSFML_ROOT=$GITHUB_WORKSPACE/install -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=${{ matrix.type.name == 'Debug' && 'Debug' || 'Release' }} ${{matrix.platform.flags}} ${{matrix.config.flags}} ${{matrix.type.flags}}
-        cmake --build $GITHUB_WORKSPACE/test/install/build --config ${{ matrix.type.name == 'Debug' && 'Debug' || 'Release' }}
+        cmake --build $GITHUB_WORKSPACE/test/install/build --config ${{ matrix.type.name == 'Debug' && 'Debug' || 'Release' }} ${{matrix.platform.suffix}}
```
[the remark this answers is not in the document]
JonnyPtn: Ah, they should be running - probably got lost in the crossfire when I was trying to get the framework tests running, will amend

examples/assets/info.plist
```diff
@@ -23,11 +23,11 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string></string>
+	<string>2.5.1</string>
```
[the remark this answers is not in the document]
JonnyPtn: I don't recall, probably just left over from when I first looked at this, will amend

src/SFML/Window/CMakeLists.txt
```diff
@@ -342,7 +340,7 @@ elseif(SFML_OS_FREEBSD)
 elseif(SFML_OS_MACOSX)
     target_link_libraries(sfml-window PRIVATE "-framework Foundation" "-framework AppKit" "-framework IOKit" "-framework Carbon")
 elseif(SFML_OS_IOS)
-    target_link_libraries(sfml-window PRIVATE "-framework Foundation" "-framework UIKit" "-framework CoreGraphics" "-framework QuartzCore" "-framework CoreMotion")
+    target_link_libraries(sfml-window PUBLIC "-framework Foundation" "-framework UIKit" "-framework CoreGraphics" "-framework QuartzCore" "-framework CoreMotion")
```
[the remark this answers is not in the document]
JonnyPtn: The frameworks need to be linked by the dependent build. I believe this is propagated automatically if we use find_package along with proper targets but as we just use dumb compiler flags we have to do this (or require that anybody using it links the frameworks themselves)

test/install/CMakeLists.txt
```diff
@@ -8,8 +8,16 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
     if(NOT BUILD_SHARED_LIBS)
         set(SFML_STATIC_LIBRARIES TRUE)
     endif()
+
+    # cmake by default will only search inside the iOS SDK for packages/libraries, so we need to tell it to look elsewhere
+    if (IOS)
+        set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE BOTH)
+        set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY BOTH)
+    endif()
```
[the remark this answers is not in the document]
JonnyPtn: Yeah as above - I've fixed the issues with the install tests (except for frameworks which is a separate can of worms) so need to enable the tests

test/install/CMakeLists.txt
```diff
@@ -8,8 +8,16 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
...
+
     find_package(SFML 3.0.0 EXACT CONFIG REQUIRED COMPONENTS Graphics Network Audio)
 endif()
 
 add_executable(test-sfml-install Install.cpp)
 target_link_libraries(test-sfml-install PRIVATE SFML::Graphics SFML::Network SFML::Audio)
+message(STATUS ${CMAKE_OSX_SYSROOT})
```
ChrisThrasher, line 23: Why this message?

test/install/CMakeLists.txt
```diff
@@ -8,8 +8,16 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
     if(NOT BUILD_SHARED_LIBS)
         set(SFML_STATIC_LIBRARIES TRUE)
     endif()
+
+    # cmake by default will only search inside the iOS SDK for packages/libraries, so we need to tell it to look elsewhere
+    if (IOS)
```
ChrisThrasher, line 13:
```suggestion
    if(IOS)
```

ChrisThrasher (approved):
Other than two minor comments, I like this! I can tell this is going to unblock some things I've previously tried to modernize how we handle dependencies. I'm eager to try those again once this is merged.

test/install/CMakeLists.txt
JonnyPtn: It was mostly useful for diagnosing SDK compatibility issues so you can see which one is being used - happy to take any direction on whether it should be removed or formatted differently

eXpl0it3r:
Looks like, we'll need to adapt BuildBot afterwards

eXpl0it3r:
Created a change for Build Bot, but I'm unsure if we can support this both for master and 2.6.x based branches: https://github.com/SFML/SFML-Buildbot/pull/39

binary1248:
> Created a change for Build Bot, but I'm unsure if we can support this both for master and 2.6.x based branches: [SFML/SFML-Buildbot#39](https://github.com/SFML/SFML-Buildbot/pull/39)

This is possible... and has been implemented.

eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:
Thanks for cleaning this up! 🧹🪣

ChrisThrasher:
https://github.com/SFML/SFML/blob/3f6403e279d3f0965938250f263cdd9fdfef5f4a/cmake/Macros.cmake#L35-L36

Should these lines be removed? They expand into nothing and due to how `set_property` works that results in technically valid (albeit not useful) CMake code.

JonnyPtn:
From memory, they existed so that people could specify the properties themselves on the command line when building SFML - Can still be done with the global CMAKE_XCODE_ATTRIBUTE_... variables if these were removed, but then it wouldn't be possible to have different settings for different targets.

Benefit and cost are both very small so not sure it really matters much

PR #2360, opened 2023-01-15
Simplify options for extra targets

JonnyPtn:
Currently the options for enabling/disabling our extra cmake targets are inconsistent, have unnecessary dependencies between each other and don't give users full control over which targets they can use.

This PR:
- Makes each target only relate to one configuration option, so is simpler
- Gives the user full control over which targets they include in the SFML project
- Makes the tidy/format options false by default to be consistent with the other extra targets (so also avoids a logic branch in our cmake script)
- Means less maintenance/support overall

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,7 +132,7 @@ jobs:
       run: sudo apt-get install clang-format-${{ matrix.version }}
 
     - name: Format Code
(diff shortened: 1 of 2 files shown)


eXpl0it3r:

ChrisThrasher:
Closed as it's basically just a revert of #2238

JonnyPtn:
Not a revert - offers the same functionality but with the user having control over it

PR #2361, opened 2023-01-15
Add more modernize- clang-tidy checks

ChrisThrasher:
## Description

Each commit adds 1 check and makes all required fixes so read this PR on a per-commit basis to get the best picture for why a given change was made.

Here are all the checks being added:
```Enabled checks:
    modernize-avoid-bind
    modernize-deprecated-headers
    modernize-deprecated-ios-base-aliases
    modernize-loop-convert
    modernize-make-shared
    modernize-make-unique
    modernize-raw-string-literal
    modernize-redundant-void-arg
    modernize-replace-auto-ptr
    modernize-replace-disallow-copy-and-assign-macro
    modernize-replace-random-shuffle
    modernize-shrink-to-fit
    modernize-unary-static-assert
    modernize-use-bool-literals
    modernize-use-default-member-init
    modernize-use-emplace
    modernize-use-noexcept
    modernize-use-override
    modernize-use-transparent-functors
    modernize-use-uncaught-exceptions
    modernize-use-using
```


ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

src/SFML/Window/OSX/SFContext.mm
```diff
@@ -31,9 +31,9 @@
 #include <SFML/Window/OSX/SFContext.hpp>
 #include <SFML/Window/OSX/WindowImplCocoa.hpp>
 
+#include <cstdint>
```
eXpl0it3r, line 34: Where's cstdint used here?

src/SFML/Window/VideoMode.cpp
```diff
@@ -60,7 +60,7 @@ const std::vector<VideoMode>& VideoMode::getFullscreenModes()
     static const auto modes = []
     {
         std::vector<VideoMode> result = priv::VideoModeImpl::getFullscreenModes();
-        std::sort(result.begin(), result.end(), std::greater<VideoMode>());
+        std::sort(result.begin(), result.end(), std::greater<>());
```
eXpl0it3r, line 63: I don't know std::sort or std::greater well enough, why can we leave out the template type?

eXpl0it3r (changes requested):

src/SFML/Window/OSX/SFContext.mm
ChrisThrasher: It's replacing `stdint.h` as part of the `modernize-deprecated-headers check` check. I think it's just catching any uses of C headers when the C++ equivalent could be used instead.

src/SFML/Window/VideoMode.cpp
ChrisThrasher: Honesty not sure how transparent functors work but apparently it's a thing we can do.

src/SFML/Window/DRM/InputImplUDev.cpp
```diff
@@ -87,8 +87,8 @@ bool systemDown()
 
 void uninitFileDescriptors()
 {
-    for (std::vector<int>::iterator itr = fileDescriptors.begin(); itr != fileDescriptors.end(); ++itr)
-        close(*itr);
+    for (const auto fileDescriptor : fileDescriptors)
```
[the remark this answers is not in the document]
ChrisThrasher: I used a value since this is an `int` and using a reference wouldn't be helpful. I'm fine using a reference though if you want.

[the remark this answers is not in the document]
eXpl0it3r: You're still referencing a specific value in a vector. I think you'll save yourself a lot of brain power, if we default to auto& instead of always thinking about whether we're now just reading or also writing data, and it keeps it consistent with everything else. (Note how all the other changes use auto&).

src/SFML/Window/OSX/SFContext.mm
eXpl0it3r: Yes, but that doesn't really answer the question 😄  
If we change a header and there's no change in code, is the header truly needed?

ChrisThrasher: Oh I see. I don't see anything in particular in this file that needs either header. I do know this header is already transitively included so we can definitely remove it from this file without breaking anything.

ChrisThrasher:

ChrisThrasher:

eXpl0it3r (approved):

PR #2363, opened 2023-01-16
Added per-character coloring as well as custom underline color.

Zombieschannel:

----

## Description

This PR is related to the https://en.sfml-dev.org/forums/index.php?topic=24995.0 as well as some other posts like https://en.sfml-dev.org/forums/index.php?topic=22890.0

The goal was to create per-character coloring and to also keep it as simple as possible.
For this, I added a setFillColor() function with a special second parameter which is the index of the character in the string.
The underline/strikethrough style has a special line that is drawn for the whole text and not per character.
Therefore, I added 2 new functions that allow for setting the color of this line.
Also, all of the colors are stored in a std::vector that resizes based on your usage:
-if you plan to set the whole text to a single color, the std::vector size will always be 1
-if you, however, plan to change color, you can enter the "multicolor mode" by just using setFillColor() which
has 2 parameters, and then the std::vector will resize so that you can change the color for each character

Example image:
image

## Tasks

* [ ] Tested on Linux
* [X] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>
using namespace sf;
struct App
{
    RenderWindow window;
    Font font;
    Text text;
    void Start()
    {
        window.create(VideoMode({ 500, 300 }), "");
        window.setFramerateLimit(60);
        font.loadFromFile("Uni_Sans_Heavy.otf");//you have to load a custom font
        text.setFont(font);
        text.setCharacterSize(50);
        text.setString("Imagine a dragon");//any text should work
        text.setOutlineThickness(3);
        text.setStyle(sf::Text::Underlined);

        //new function to set underline/strikethrough color of the text
        text.setUnderlineColor(Color(255, 255, 255));
        //new function to set underline/strikethrough outline color of the text
        text.setUnderlineOutlineColor(Color(255, 0, 0, 64));

        for (int i = 0; i < text.getString().getSize(); i++)
        {
            //new function to set a color for each character
            text.setFillColor(Color((float)i / text.getString().getSize() * 255, 0, 128), i);

            int val = (float)i / text.getString().getSize() * 255;
            //new function to set an outline color for each character
            text.setOutlineColor(Color(100, val, val), i);
        }
    }
    void app()
    {
        while (window.isOpen())
        {
            Event event;
            {
                switch (event.type)
                {
                case Event::Closed:
                    window.close();
                    break;
                }
            }
            window.clear();
            window.draw(text);
            window.display();
        }
    }
};
int main()
{
    App app;
    app.Start();
    app.app();
}
```


src/SFML/Graphics/Text.cpp
```diff
@@ -191,16 +191,38 @@ void Text::setStyle(std::uint32_t style)
 ////////////////////////////////////////////////////////////
 void Text::setFillColor(const Color& color)
 {
-    if (color != m_fillColor)
+    //In case we use the same color on all characters, no need to store
+    //the same color multiple times
+    if (m_fillColor.size() != 1)
+        m_fillColor.resize(1, Color(255, 255, 255));
```
ChrisThrasher, line 197:
```suggestion
        m_fillColor.resize(1, Color::White);
```
Prefer named colors to using the raw RGB codes

include/SFML/Graphics/Text.hpp
```diff
@@ -445,8 +541,10 @@ class SFML_GRAPHICS_API Text : public Drawable, public Transformable
     float                 m_letterSpacingFactor{1.f};                  //!< Spacing factor between letters
     float                 m_lineSpacingFactor{1.f};                    //!< Spacing factor between lines
     std::uint32_t         m_style{Regular};                            //!< Text style (see Style enum)
-    Color                 m_fillColor{Color::White};                   //!< Text fill color
-    Color                 m_outlineColor{Color::Black};                //!< Text outline color
+    std::vector<Color>    m_fillColor{{255, 255, 255}};                //!< Text fill color
+    Color                 m_underlineFillColor{255, 255, 255};         //!< Text underline/strikethrough fill color
+    std::vector<Color>    m_outlineColor{{0, 0, 0}};                   //!< Text outline color
+    Color                 m_underlineOutlineColor{0, 0, 0};            //!< Text underline/strikethrough outline color
```
ChrisThrasher, line 547: Please use named colors

src/SFML/Graphics/Text.cpp
```diff
@@ -191,16 +191,38 @@ void Text::setStyle(std::uint32_t style)
...
+    }
+}
+
+
+////////////////////////////////////////////////////////////
+void Text::setFillColor(const Color& color, unsigned int index)
+{
+    //Resize the vector so that each character can have a unique color
+    if (m_fillColor.size() != m_string.getSize())
+        m_fillColor.resize(m_string.getSize(), Color(255, 255, 255));
```
ChrisThrasher, line 215: Please use named colors

src/SFML/Graphics/Text.cpp
```diff
@@ -209,16 +231,38 @@ void Text::setFillColor(const Color& color)
 ////////////////////////////////////////////////////////////
 void Text::setOutlineColor(const Color& color)
 {
-    if (color != m_outlineColor)
+    //In case we use the same color on all characters, no need to store
+    //the same color multiple times
+    if (m_outlineColor.size() != 1)
+        m_outlineColor.resize(1, Color(0, 0, 0));
```
ChrisThrasher, line 237: Please use named colors

src/SFML/Graphics/Text.cpp
```diff
@@ -209,16 +231,38 @@ void Text::setFillColor(const Color& color)
...
+    }
+}
+
+
+////////////////////////////////////////////////////////////
+void Text::setOutlineColor(const Color& color, unsigned int index)
+{
+    //Resize the vector so that each character can have a unique color
+    if (m_outlineColor.size() != m_string.getSize())
+        m_outlineColor.resize(m_string.getSize(), Color(0, 0, 0));
```
ChrisThrasher, line 255: Please use named colors

ChrisThrasher (changes requested):

ChrisThrasher:

include/SFML/Graphics/Text.hpp
```diff
@@ -445,8 +541,10 @@ class SFML_GRAPHICS_API Text : public Drawable, public Transformable
     float                 m_letterSpacingFactor{1.f};                  //!< Spacing factor between letters
     float                 m_lineSpacingFactor{1.f};                    //!< Spacing factor between lines
     std::uint32_t         m_style{Regular};                            //!< Text style (see Style enum)
-    Color                 m_fillColor{Color::White};                   //!< Text fill color
-    Color                 m_outlineColor{Color::Black};                //!< Text outline color
+    std::vector<Color>    m_fillColor{Color::White};                   //!< Text fill color
+    Color                 m_underlineFillColor{Color::White};          //!< Text underline/strikethrough fill color
+    std::vector<Color>    m_outlineColor{Color::Black};                //!< Text outline color
+    Color                 m_underlineOutlineColor{Color::Black};       //!< Text underline/strikethrough outline color
```
ChrisThrasher, line 547: Do we expect `m_fillColor` and `m_outlineColor` to have the same size at all times? If so then perhaps we combine them into a single vector to better represent this invariant.

src/SFML/Graphics/Text.cpp
```diff
@@ -235,6 +279,44 @@ void Text::setOutlineThickness(float thickness)
...
+    // (if geometry is updated anyway, we can skip this step)
+    if (!m_geometryNeedUpdate)
+    {
+        bool isUnderlined    = m_style & Underlined;
+        bool isStrikeThrough = m_style & StrikeThrough;
+        for (std::size_t i = m_outlineVertices.getVertexCount() - 6 * isUnderlined - 6 * isStrikeThrough;
+             i < m_outlineVertices.getVertexCount();
+             ++i)
+            m_outlineVertices[i].color = color;
+    }
```
Bromeon, line 316: Code duplicated from above (could pass `m_outlineVertices` as a parameter).

src/SFML/Graphics/Text.cpp
```diff
@@ -209,16 +231,38 @@ void Text::setFillColor(const Color& color)
...
+    m_outlineColor[0] = color;
+
+    // Change vertex colors directly, no need to update whole geometry
+    // (if geometry is updated anyway, we can skip this step)
+    if (!m_geometryNeedUpdate)
     {
-        m_outlineColor = color;
+        for (std::size_t i = 0; i < m_outlineVertices.getVertexCount(); ++i)
+            m_outlineVertices[i].color = color;
+    }
```
Bromeon, line 246: Code duplicated from above (could use `m_outlineVertices` as a parameter).

src/SFML/Graphics/Text.cpp
```diff
@@ -235,6 +279,44 @@ void Text::setOutlineThickness(float thickness)
...
+    if (!m_geometryNeedUpdate)
+    {
+        bool isUnderlined    = m_style & Underlined;
+        bool isStrikeThrough = m_style & StrikeThrough;
+        for (std::size_t i = m_vertices.getVertexCount() - 6 * isUnderlined - 6 * isStrikeThrough;
+             i < m_vertices.getVertexCount();
+             ++i)
+            m_vertices[i].color = color;
+    }
+}
```
Bromeon, line 298: As ugly as this API gets, this should be called `setUnderlineFillColor` for consistency with all the other fill + outline pairs.

Bromeon (commented):
Thank you for this PR! 🙂 

The logic -- which character has which color -- seems rather brittle in the current implementation.
So maybe before you go into the source-level comments, some general questions:

1. How do we know which element in this `std::vector` refers to which character?
   * You use `vector::resize`, initializing all _preceding_ characters to white, but leaving all _following_ characters without color.
   * You use the `size() == 1` condition as a check whether the whole text has 1 single color. How do you differ this from having 1 character with the given color (and possibly more characters)?
2. What happens to the colors if the underlying string is changed? How are they kept in sync?
3. There are now quite a few interactions which are non-obvious to the user. Can you elaborate the desired behavior?
   * fill entire text with color A, then fill a single character with B
   * fill a single character with A, then the entire text with B
   * set outline of a single character with A, then **_fill_** the entire text with B

TDLR: every state must be unambiguously representable in exactly one way. How full-text and per-character colors interact should be specified, as well as whether fill/outline colors are completely independentor not. Behavior in case of text modifications should also be specified.

src/SFML/Graphics/Text.cpp
Zombieschannel: Thanks for the feedback! I'll do my part.

Zombieschannel: Ok, fixed it.

include/SFML/Graphics/Text.hpp
```diff
@@ -436,19 +550,31 @@ class SFML_GRAPHICS_API Text : public Drawable, public Transformable
...
+    const Font*                m_font{};                              //!< Font used to display the string
+    unsigned int               m_characterSize{30};                   //!< Base size of characters, in pixels
+    float                      m_letterSpacingFactor{1.f};            //!< Spacing factor between letters
+    float                      m_lineSpacingFactor{1.f};              //!< Spacing factor between lines
+    std::uint32_t              m_style{Regular};                      //!< Text style (see Style enum)
+    Color                      m_underlineFillColor{Color::White};    //!< Text underline/strikethrough fill color
+    Color                      m_underlineOutlineColor{Color::Black}; //!< Text underline/strikethrough outline color
+    bool                       m_multiFillColor{0};          //!< Should the text have more than a single fill color
+    bool                       m_multiOutlineColor{0};       //!< Should the text have more than a single outline color
+    float                      m_outlineThickness{0.f};      //!< Thickness of the text's outline
```
ChrisThrasher, line 574:
```suggestion
    bool                       m_multiFillColor{};           //!< Should the text have more than a single fill color
    bool                       m_multiOutlineColor{};        //!< Should the text have more than a single outline color
    float                      m_outlineThickness{};         //!< Thickness of the text's outline
```
We can use default initialization to simplify this. Plus it's better to not initialize a boolean with an integer since that makes the code a bit harder to read.

include/SFML/Graphics/Text.hpp
```diff
@@ -436,19 +550,31 @@ class SFML_GRAPHICS_API Text : public Drawable, public Transformable
...
-    mutable VertexArray   m_vertices{PrimitiveType::Triangles};        //!< Vertex array containing the fill geometry
+    String                     m_string;                              //!< String to display
+    const Font*                m_font{};                              //!< Font used to display the string
+    unsigned int               m_characterSize{30};                   //!< Base size of characters, in pixels
+    float                      m_letterSpacingFactor{1.f};            //!< Spacing factor between letters
+    float                      m_lineSpacingFactor{1.f};              //!< Spacing factor between lines
+    std::uint32_t              m_style{Regular};                      //!< Text style (see Style enum)
+    Color                      m_underlineFillColor{Color::White};    //!< Text underline/strikethrough fill color
+    Color                      m_underlineOutlineColor{Color::Black}; //!< Text underline/strikethrough outline color
+    bool                       m_multiFillColor{0};          //!< Should the text have more than a single fill color
```
ChrisThrasher, line 572: Is there a better name for this member? Seeing this variable used in context it's not really obvious what it means without reading the doc comment next to its declaration.

include/SFML/Graphics/Text.hpp
```diff
@@ -436,19 +550,31 @@ class SFML_GRAPHICS_API Text : public Drawable, public Transformable
...
+    bool                       m_multiOutlineColor{0};       //!< Should the text have more than a single outline color
+    float                      m_outlineThickness{0.f};      //!< Thickness of the text's outline
+    mutable std::vector<Color> m_fillColor{Color::White};    //!< Text fill color(s)
+    mutable std::vector<Color> m_outlineColor{Color::Black}; //!< Text outline color(s)
```
ChrisThrasher, line 576:
```suggestion
    mutable std::vector<Color> m_fillColors{Color::White};    //!< Text fill color(s)
    mutable std::vector<Color> m_outlineColors{Color::Black}; //!< Text outline color(s)
```
Can we pluralize these member variable names so it's more clear that they're now containers?

src/SFML/Graphics/Text.cpp
```diff
@@ -191,16 +191,50 @@ void Text::setStyle(std::uint32_t style)
 ////////////////////////////////////////////////////////////
 void Text::setFillColor(const Color& color)
 {
-    if (color != m_fillColor)
+    //In case we use the same color on all characters, no need to store
+    //the same color multiple times
```
ChrisThrasher, line 195:
```suggestion
    // In case we use the same color on all characters, no need to store
    // the same color multiple times
```

ChrisThrasher: Check for other places where there is no space between `//` and the first word of the comment but I won't bother making comments for every single occurrence.

src/SFML/Graphics/Text.cpp
```diff
@@ -568,4 +702,38 @@ void Text::ensureGeometryUpdate() const
     m_bounds.height = maxY - minY;
 }
 
+
+////////////////////////////////////////////////////////////
+void Text::updateUnderline(VertexArray& arr, const Color& color) const
+{
+    bool isUnderlined    = m_style & Underlined;
+    bool isStrikeThrough = m_style & StrikeThrough;
```
ChrisThrasher, line 710:
```suggestion
    const bool isUnderlined    = m_style & Underlined;
    const bool isStrikeThrough = m_style & StrikeThrough;
```

src/SFML/Graphics/Text.cpp
```diff
@@ -568,4 +702,38 @@ void Text::ensureGeometryUpdate() const
...
+    //In case of newlines, find them and change lines
+    unsigned int  lineCount = 0;
+    std::uint32_t prevChar  = 0;
+    for (unsigned int i = 0; i < m_string.getSize(); ++i)
+    {
+        std::uint32_t curChar = m_string[i];
+
+        if (isUnderlined && (curChar == U'\n' && prevChar != U'\n'))
+        {
+            for (unsigned int j = 0; j < 6; j++)
```
ChrisThrasher, line 721: What's the significance of this `6`?

ChrisThrasher (changes requested):

src/SFML/Graphics/Text.cpp
Zombieschannel: SFML uses triangles to render quads (quads aren't available on OpenGLES), therefore the number of vertices per character is 6 instead of 4.
The append() function gets called 6 times for each vertex.
image

include/SFML/Graphics/Text.hpp
Zombieschannel: Tbh, not sure. Maybe m_styleFillColor? Then again, bold and italics don't have any special lines.

Bromeon (commented):
Thanks for the updates!

I'm still not convinced by the usage of `std::vector` for single characters. What I don't like is that setting color at index N implicitly assigns colors from 0 to N-1. In other words, it fills the left-hand-side range, but not the right.

This has surprising implications.
```cpp
sf::Text text = ...; // assume length 10
text.setFillColor(green);
text.setCharacterFillColor(7, blue);

// now, we have:
// 0..6 green
//  7   blue
// 8..9 green (but not in the vector)

sf::Color at6 = text.getCharacterFillColor(6); // green
sf::Color at8 = text.getCharacterFillColor(8); // undefined behavior
```

There are ways to address that, e.g. hashmap from index to color, but it's not the full story.

---

There's another problem -- character-based colors are not in sync with the length of the text.
```cpp
sf::Text text = /* length 10 */;
text.setFillColor(green);
text.setCharacterFillColor(7, blue);

text.setText(/* length 5 */);
// now all characters are green

text.setText(/* length 8 */);
// now the last character is suddenly blue, 
// because of stale state from a previous assignment!
```

A major issue of the current design is that the character's **color** state is separate from the characters themselves, and there are lots of issues when you change one but not the other. This goes already a bit toward Rich-Text territory. Many Rich-Text APIs support use cases like this one by inline markup (think BBCode), exactly because it keeps the text and representation together.

This is likely out of scope for a basic `sf::Text` class, or SFML in general. How far do we want to drive this? This PR adds support for per-character colors, but we don't support per-character styles, sizes or outline thicknesses. I'm not saying we should, but we need to have a vision what `sf::Text` should be capable of and what should be delegated to extensions.

---

All in all, this seemingly simple feature adds a lot of complexity in practice. If we don't find a way to make this behavior intuitive and robust, I'm not sure if it's a feature worth having, in the spirit of SFML's "simple". A part of such robustness would be:
* an intuitive and hard-to-misuse API
* a vision about the scope of per-character features
* unit-testing of the functionality, particularly edge cases mentioned above

What do you (everyone) think?

Bromeon:
**TLDR:** Before continuing implementation work, we should clarify design and scope topics mentioned above.

src/SFML/Graphics/Text.cpp
```diff
@@ -280,14 +376,28 @@ std::uint32_t Text::getStyle() const
...
 {
-    return m_fillColor;
+    return getCharacterFillColor(0);
+}
+
+
+////////////////////////////////////////////////////////////
+const Color& Text::getCharacterFillColor(unsigned int index) const
+{
+    return m_fillColors[index];
```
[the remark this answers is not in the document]
Zombieschannel: > What I don't like is that setting color at index N implicitly assigns colors from 0 to N-1. In other words, it fills the left-hand-side range, but not the right.

I don't think that is how it works. When you switch to the multi-color mode, there is this check which makes sure that the vector is at least the string size (in your case, 10).

```cpp
    // Set the new per character colors to the color set by setFillColor()
    if (!m_multiFillColor)
    {
        m_fillColors.resize(m_string.getSize());
        for (std::size_t i = 0; i < m_fillColors.size(); i++)
            m_fillColors[i] = m_fillColors[0];
    }```

If you however provide an index that is out of bounds (ex. 12), the vector will resize to size 13, so that it can keep your index in the array. This is meant for use such as this:
```cpp
Text text;
//add position, rotation, scale, font, style...
//at this point text.setString() has not been called
text.setCharacterFillColor(3, Color::Red);
//this index is 3 which is now out of bounds
//the vector will resize to size 4, so that in case that you set text.setString() later (and the string will have 4 or more characters), 
//it will be seen when you draw it on the screen)
text.setString("Test");
//now the string has been set to "Test" and the vector is still at size 4.
//however the last character will now be red
```



> // now the last character is suddenly blue, 
> // because of stale state from a previous assignment!

As I said earlier, this is completely intended. Since if you create a sf::Text, first set the character to some color before setting the string, the idea was for it to work the same if you first set the string and then set the character color.

> m_perCharacterFillColor? more verbose but clearer.
Idk, to me `m_multiFillColor` is also ok (maybe instead of multi -> multiple)

> Could you add this in a comment?
Sure

> e.g. in loops, you also break when size is reached, not when size-1 is exceeded
Well it seems like some of us think of this differently :)

> This causes UB if the index is out-of-bounds.
Yup, as it says, if you pass an invalid index, you'll get undefined behaviour (though, I should probably mention "out of bounds" somewhere). 
```cpp
    /// The validity of the index is not checked, passing an invalid
    /// index results in an undefined behaviour.```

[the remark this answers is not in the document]
Comment (Bromeon) [reply src/SFML/Graphics/Text.cpp:386]: > Yup, as it says, if you pass an invalid index, you'll get undefined behaviour

There's two problems with that:
* First, it's not clear what constitutes "invalid". Anything _less_ than the maximum set index will return a color. Anything _greater_ will cause UB. The user faces a left/right asymmetry for no apparent reason.
* Undefined behavior is an unnecessary source of bugs in applications. It's really just C and C++ where UB is even a socially accepted thing that you just have to live with, any other language tries to prevent it as hard as possible. We should at least have an `assert` here. Some stdlib implementations likely do that inside `operator[]`, but it's not something we can rely on.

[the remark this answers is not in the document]
Comment (Zombieschannel) [reply src/SFML/Graphics/Text.cpp:386]: Hm, one idea of how to fix this is instead of using the vector for both single and multi color modes, keep a sf::Color for single color and a std::vector for multi color. Then if you request an index that is out of bounds, it will return the color stored in single color mode. For example:
```cpp
        text.setString("Test"); //set a string of size 4
        text.setFillColor(Color::Green); //set all characters to green
        text.setCharacterFillColor(3, Color::Blue); //set the fourth character to blue
        Color x = text.getCharacterFillColor(5); //5 is out of bounds for the m_fillColors, therefore, return the green color
```

PR #2365, opened 2023-01-16
readability-convert-member-functions-to-static check

gnawme:

----

## Description

Added `clang-tidy` check `readability-convert-member-functions-to-static`. Will be followed up by `[readability-static-accessed-through-instance](https://clang.llvm.org/extra/clang-tidy/checks/readability/static-accessed-through-instance.html)` once CI builds.

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        {
                window.close();
        }

        window.clear();
        window.display();
    }
}
```

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -9,6 +9,7 @@ Checks: >
   readability-convert-member-functions-to-static,
   readability-identifier-naming,
   readability-isolate-declaration,
+  readability-static-accessed-through-instance,
   -clang-analyzer-core.NonNullParamChecker,
   -clang-analyzer-core.NullDereference,
   -clang-analyzer-nullability.NullablePassedToNonnull,
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -6,6 +6,7 @@ Checks: >
   modernize-use-equals-delete,  
   modernize-use-nullptr,
   readability-container-size-empty,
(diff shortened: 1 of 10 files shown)


ChrisThrasher:
I'm undecided on this for a few reasons. Certain platform implementations of features have trivial implementations that merely return true or false. While these functions can technically become static functions, that's not how they're meant to be used per se since other implementations do require non-trivial implementations that cannot be static. It's weird to have some implementations be static and others not. The other reason is that once we make an interface `static`, we cannot remove that later without breaking the API. It's a promise we must keep for the lifetime of a release cycle.

I think it's certainly possible that some of these functions ought to be made `static` but I'm afraid this particularly check may be too aggressive for SFML.

gnawme:
> I think it's certainly possible that some of these functions ought to be made `static` but I'm afraid this particularly check may be too aggressive for SFML.

You're probably right. I was mainly curious how extensive the changes would be. I'll go after some of the smaller `readability-` checks

PR #2367, opened 2023-01-17
Added readability-qualified-auto, readability-redundant-access-specifiers

gnawme:

----

## Description

Instead of adding to the list of `readability-*` checks, this PR enables them all, and excludes those not wanted.

[`readability-else-after-return`](https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html) dominates this PR.

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Enabled unit tests.


examples/shader/Shader.cpp
```diff
@@ -97,10 +97,7 @@ class WaveBlur : public Effect
         m_text.setPosition({30.f, 20.f});
 
         // Load the shader
-        if (!m_shader.loadFromFile("resources/wave.vert", "resources/blur.frag"))
-            return false;
-
-        return true;
+        return m_shader.loadFromFile("resources/wave.vert", "resources/blur.frag");
```
ChrisThrasher, line 100: What check caught this? This is a no-brainer check to add and would make for a great PR by itself

gnawme: Likely `readability-simplify-boolean-expr`. Yeah, this PR is starting to get bloated.

include/SFML/Graphics/Texture.hpp
```diff
@@ -61,7 +61,7 @@ class SFML_GRAPHICS_API Texture : GlResource
         Pixels      //!< Texture coordinates in range [0 .. size]
     };
 
-public:
+
```
ChrisThrasher, line 64:
```suggestion
```
I think I'd rather delete these extra `private:` tags than replace them with empty space. We don't need these extra empty newlines.

ChrisThrasher (commented):
I like this! Just some formatting comments about removing `private:` but I think this is close to mergeable

examples/shader/Effect.hpp
```diff
@@ -70,7 +70,7 @@ class Effect : public sf::Drawable
     virtual void onUpdate(float time, float x, float y)                                 = 0;
     virtual void onDraw(sf::RenderTarget& target, const sf::RenderStates& states) const = 0;
 
-private:
+
```
Bromeon, line 73: The `private:` and `public:` qualifiers acted as a separator between multiple groups in the class definition (e.g. methods, variables).

This is an opinionated change by a tool, not a universally accepted best practice. It means we lose that visual separation.

ChrisThrasher: I don't disagree that it's opinionated. I personally like this opinion but won't protest if others disagree.

include/SFML/Graphics/Texture.hpp
gnawme: Yeah, I'll go hunt for them. I thought that the `format` build might catch them, but apparently not.

examples/shader/Effect.hpp
Bromeon: I'm fine with either as well, it's just a question if this change is necessary and truly improves something.
Just because a tool says so...

include/SFML/Graphics/Texture.hpp
ChrisThrasher: We allow for up to 2 consecutive empty newlines which is why this wasn't automatically removed.

ChrisThrasher (approved):

ChrisThrasher:

ChrisThrasher:

eXpl0it3r (approved):

PR #2371, opened 2023-01-20
Streamlined readability-* clang-tidy checks

gnawme:

----

## Description

This PR streamlines the `readability-*` checks for `clang-tidy`, simplifying the `readability` selector to `readability-*` and completing the list of `readability-` checks to exclude. It effectively adds these checks:

- readability-avoid-const-params-in-decls
- readability-const-return-type
- readability-container-data-pointer
- readability-delete-null-pointer
- readability-duplicate-include
- readability-inconsistent-declaration-parameter-name
- readability-make-member-function-const
- readability-misleading-indentation
- readability-misplaced-array-index
- readability-non-const-parameter
- readability-static-accessed-through-instance
- readability-static-definition-in-anonymous-namespace
- readability-string-compare
- readability-suspicious-call-argument
- readability-uniqueptr-delete-release
- readability-use-anyofallof

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Enabled and built tests, all pass.

### DIFF
--- a/include/SFML/Audio/SoundSource.hpp
+++ b/include/SFML/Audio/SoundSource.hpp
@@ -35,6 +35,7 @@
 
 namespace sf
 {
+// NOLINTBEGIN(readability-make-member-function-const)
 ////////////////////////////////////////////////////////////
 /// \brief Base class defining a sound's properties
 ///
@@ -291,6 +292,7 @@ protected:
     unsigned int m_source; //!< OpenAL source identifier
 };
 
+// NOLINTEND(readability-make-member-function-const)
 } // namespace sf
 
 
--- a/src/SFML/Audio/SoundSource.cpp
+++ b/src/SFML/Audio/SoundSource.cpp
@@ -34,6 +34,7 @@
 
 namespace sf
 {
+// NOLINTBEGIN(readability-make-member-function-const)
 ////////////////////////////////////////////////////////////
 SoundSource::SoundSource()
 {
@@ -204,5 +205,6 @@ SoundSource::Status SoundSource::getStatus() const
 
     return Stopped;
 }
+// NOLINTEND(readability-make-member-function-const)
 
 } // namespace sf
--- a/src/SFML/Window/DRM/InputImplUDev.cpp
+++ b/src/SFML/Window/DRM/InputImplUDev.cpp
@@ -28,6 +28,7 @@
 #include <SFML/System/Err.hpp>
 #include <SFML/Window/DRM/InputImplUDev.hpp>
 
+#include <algorithm>
 #include <cerrno>
 #include <cstdio>
 #include <cstdlib>
@@ -664,13 +665,9 @@ void InputImpl::setMousePosition(const Vector2i& position, const WindowBase& /*r
 ////////////////////////////////////////////////////////////
 bool InputImpl::isTouchDown(unsigned int finger)
 {
-    for (const auto& slot : touchSlots)
-    {
-        if (slot.id == static_cast<int>(finger))
-            return true;
-    }
-
-    return false;
+    return std::any_of(touchSlots.cbegin(),
+                       touchSlots.cend(),
+                       [finger](const TouchSlot& slot) { return slot.id == static_cast<int>(finger); });
 }
 
 
--- a/src/SFML/Window/OSX/HIDInputManager.mm
+++ b/src/SFML/Window/OSX/HIDInputManager.mm
@@ -34,8 +34,8 @@
 
 namespace
 {
-static const std::uint8_t unknownVirtualCode = 0xff;
-static const bool         isIsoKeyboard      = (KBGetLayoutType(LMGetKbdType()) == kKeyboardISO);
+const std::uint8_t unknownVirtualCode = 0xff;
+const bool         isIsoKeyboard      = (KBGetLayoutType(LMGetKbdType()) == kKeyboardISO);
 }
 
 namespace sf::priv
@@ -962,7 +962,7 @@ CFSetRef HIDInputManager::copyDevices(std::uint32_t page, std::uint32_t usage)
 
 
 ////////////////////////////////////////////////////////////
-bool HIDInputManager::isPressed(IOHIDElements& elements)
+bool HIDInputManager::isPressed(IOHIDElements& elements) const
 {
     bool pressed = false;
     for (auto it = elements.begin(); it != elements.end() && !pressed; /* noop */)
--- a/src/SFML/Window/Unix/WindowImplX11.cpp
+++ b/src/SFML/Window/Unix/WindowImplX11.cpp
@@ -92,6 +92,7 @@ constexpr unsigned long eventMask = FocusChangeMask | ButtonPressMask | ButtonRe
 constexpr unsigned int maxTrialsCount = 5;
 
 // Filter the events received by windows (only allow those matching a specific window)
+// NOLINTNEXTLINE(readability-non-const-parameter)
 Bool checkEvent(::Display*, XEvent* event, XPointer userData)
 {
     // Just check if the event matches the window
@@ -350,13 +351,9 @@ bool isWMAbsolutePositionGood()
     if (!ewmhSupported())
         return false;
 
-    for (const sf::String& name : wmAbsPosGood)
-    {
-        if (name == windowManagerName)
-            return true;
-    }
-
-    return false;
+    return std::any_of(std::begin(wmAbsPosGood),
+                       std::end(wmAbsPosGood),
+                       [&](const sf::String& name) { return name == windowManagerName; });
 }
 } // namespace WindowsImplX11Impl
 } // namespace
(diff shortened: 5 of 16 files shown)


ChrisThrasher:

include/SFML/Audio/SoundSource.hpp
```diff
@@ -81,7 +81,7 @@ class SFML_AUDIO_API SoundSource : AlResource
     /// \see getPitch
     ///
     ////////////////////////////////////////////////////////////
-    void setPitch(float pitch);
+    void setPitch(float pitch) const;
```
ChrisThrasher, line 84: It's odd that these setter functions don't mutate the class. I guess I'm fine making these const but it just seems weird.

src/SFML/Window/DRM/InputImplUDev.cpp
```diff
@@ -628,13 +629,9 @@ void InputImpl::setMousePosition(const Vector2i& position, const WindowBase& /*r
...
-    for (const auto& slot : touchSlots)
-    {
-        if (slot.id == static_cast<int>(finger))
-            return true;
-    }
-
-    return false;
+    return std::any_of(touchSlots.cbegin(),
+                       touchSlots.cend(),
+                       [finger](const TouchSlot& slot) { return slot.id == static_cast<int>(finger); });
```
ChrisThrasher, line 634: This is neat. Which check recommended this?

gnawme: `readability-use-anyofallof`

include/SFML/Audio/SoundSource.hpp
ChrisThrasher: @Bromeon any insight into this and whether you think this is a good idea?

eXpl0it3r: They do mutate stuff, but not its own members, but instead OpenAL states.  
I don't think we should change the interface, as it would limit our implementation flexibility.

ChrisThrasher: I'm fine disabling that check for this file then

src/SFML/Audio/SoundSource.cpp
```diff
@@ -204,5 +205,6 @@ SoundSource::Status SoundSource::getStatus() const
 
     return Stopped;
 }
+// NOLINTEND(readability-make-member-function-const)
```
ChrisThrasher, line 208: Is this needed in both the header and the implementation? Can we just put it in one or the other?

gnawme: It seems to need it in both places.

ChrisThrasher:

include/SFML/Audio/SoundSource.hpp
```diff
@@ -21,8 +21,8 @@
 // 3. This notice may not be removed or altered from any source distribution.
 //
 ////////////////////////////////////////////////////////////
-
```
ChrisThrasher, line 24: Can you put this newline back?

include/SFML/Audio/SoundSource.hpp
```diff
@@ -21,8 +21,8 @@
...
 #pragma once
+// NOLINTBEGIN(readability-make-member-function-const)
```
ChrisThrasher, line 25: Can this be moved down such that it only wraps the class and not also all those headers?

ChrisThrasher (approved):

ChrisThrasher:

include/SFML/Audio/SoundSource.hpp
```diff
@@ -33,6 +33,7 @@
 #include <SFML/System/Vector3.hpp>
 
 
+// NOLINTBEGIN(readability-make-member-function-const)
```
kimci86, line 36: This could be moved inside the namespace like it is done in the corresponding source file.

ChrisThrasher:

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -473,13 +474,9 @@ void JoystickImpl::cleanupDInput()
...
-    for (const JoystickRecord& record : joystickList)
-    {
-        if (record.index == index)
-            return true;
-    }
-
-    return false;
+    return std::any_of(joystickList.cbegin(),
+                       joystickList.cend(),
+                       [index](const JoystickRecord& record) { return record.index == index; });
```
Bromeon, line 479: `std::find()`?

Possibly also above...

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -91,7 +91,8 @@ constexpr unsigned long eventMask = FocusChangeMask | ButtonPressMask | ButtonRe
 constexpr unsigned int maxTrialsCount = 5;
 
 // Filter the events received by windows (only allow those matching a specific window)
-Bool checkEvent(::Display*, XEvent* event, XPointer userData)
+// NOLINTNEXTLINE(readability-non-const-parameter)
+Bool checkEvent(::Display*, XEvent* event, const XPointer userData)
```
Bromeon, line 95: Why do we add an exception here? I don't see how this is different from any other parameter.

Same above

Bromeon (dismissed):

src/SFML/Window/Win32/JoystickImpl.cpp
gnawme: `std::any_of` has the virtue of returning `bool` directly

src/SFML/Window/Unix/WindowImplX11.cpp
gnawme: `clang-tidy` changed `userData` to `const` in the first pass, then kept coming back and complaining that `userData` could be `const`.

So I kept it `const` and added the exception to shut it up.

ChrisThrasher: This also confused me so I played with it locally and found no way to satisfy clang-tidy without replacing `XPointer` with something else but then we lose some expressiveness so I didn't want to do that. Turning it off here seemed reasonable.

Bromeon: > `clang-tidy` changed `userData` to `const` in the first pass

Would it then not make more sense to disable the first lint, and not annotate the parameter `const`?

src/SFML/Window/Win32/JoystickImpl.cpp
Bromeon: I see, OK.


<details><summary>Off-topic:</summary>

To be fair, either way is not great, STL algorithms still lack expressive power in 2023. 
The Kotlin equivalent of these 3 lines:
```kotlin
return joystickList.any { it.index == index }
```

</details>

src/SFML/Window/Unix/WindowImplX11.cpp
gnawme: > > `clang-tidy` changed `userData` to `const` in the first pass
> 
> Would it then not make more sense to disable the first lint, and not annotate the parameter `const`?

No, I think it's useful to declare the function's intent to not modify `userData`

Bromeon: > No, I think it's useful to declare the function's intent to not modify `userData`. Achieving some measure of [const-correctness](https://isocpp.org/wiki/faq/const-correctness) is always worthwhile.

Following this argumentation, the function should look like this:
```cpp
Bool checkEvent(::Display* const, XEvent* const event, const XPointer userData);
```

To clarify, we're talking about _top-level_ `const` here, it doesn't express if you have read-only access through the pointer. All it does is protect you from doing `userData = somethingElse` in the function, i.e. reassigning the parameter.

My point is that we don't use this convention in SFML, so now having a single parameter out of 3 marked `const` raises more questions than it answers.

ChrisThrasher: I'm fine with both options so I'll defer to what Bromeon wants.

gnawme: > To clarify, we're talking about _top-level_ `const` here

Point taken; the rest of the `const` in this PR were applied to functions, not function arguments

ChrisThrasher:
Can you rebase on master to fix the merge conflicts?

src/SFML/Window/OSX/HIDInputManager.mm
```diff
@@ -754,6 +754,7 @@
     }
 
     NSSet* keyboards = static_cast<NSSet*>(underlying); // Toll-Free Bridge
+    CFSetGetValues(keyboards, devicesArray.data());
```
kimci86, line 757: It looks like a rebase mistake. You can remove this line.

ChrisThrasher:

ChrisThrasher:

eXpl0it3r (approved):

Issue #2366, opened 2023-01-16
Replace `std::strcmp` with `std::string_view::operator==`


ChrisThrasher:
## Subject of the issue

<img width="287" alt="Screenshot 2023-01-16 at 4 41 43 PM" src="https://user-images.githubusercontent.com/39244355/212780820-a8c793e6-9c17-4979-8088-b71670c4ef1b.png">

Looks like we're only using `std::strcmp` to detect equality of strings which we can do more expressively with `std::string_view::operator==`.

Bromeon:
I guess you mean `operator==` 😉

But are all these cases using at least one `std::string` argument? Doesn't make sense to allocate if we only have C strings...

ChrisThrasher:
> I guess you mean operator== 😉

Oops yes!

> But are all these cases using at least one std::string argument? Doesn't make sense to allocate if we only have C strings...

I think the readability gains are worth perhaps incurring a single allocation. Besides some of these strings will fall within the small string optimizations so they may not even allocate.

kimci86:
We can also use `std::string_view` to avoid dynamic allocation.

PR #2373, opened 2023-01-21
Replace std::strcmp by std::string_view::operator==

litelawliet:

----

## Description

This PR is related to the issue https://github.com/SFML/SFML/issues/2366

I couldn't find all the references found in the issue but it should be exhaustive for master branch. Feel free to point out any missing references or any remarks.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/examples/vulkan/Vulkan.cpp
+++ b/examples/vulkan/Vulkan.cpp
@@ -470,11 +470,11 @@ public:
             // -- VK_LAYER_GOOGLE_unique_objects
             // These layers perform error checking and warn about bad or sub-optimal Vulkan API usage
             // VK_LAYER_LUNARG_monitor appends an FPS counter to the window title
-            if (!std::strcmp(layer.layerName, "VK_LAYER_LUNARG_standard_validation"))
+            if (std::string_view(layer.layerName) == "VK_LAYER_LUNARG_standard_validation")
             {
                 validationLayers.push_back("VK_LAYER_LUNARG_standard_validation");
             }
-            else if (!std::strcmp(layer.layerName, "VK_LAYER_LUNARG_monitor"))
+            else if (std::string_view(layer.layerName) == "VK_LAYER_LUNARG_monitor")
             {
                 validationLayers.push_back("VK_LAYER_LUNARG_monitor");
             }
@@ -611,7 +611,7 @@ public:
 
             for (VkExtensionProperties& extension : extensions)
             {
-                if (!std::strcmp(extension.extensionName, VK_KHR_SWAPCHAIN_EXTENSION_NAME))
+                if (std::string_view(extension.extensionName) == VK_KHR_SWAPCHAIN_EXTENSION_NAME)
                 {
                     supportsSwapchain = true;
                     break;
--- a/src/SFML/Window/DRM/DRMContext.cpp
+++ b/src/SFML/Window/DRM/DRMContext.cpp
@@ -365,7 +365,7 @@ int initDrm(sf::priv::Drm& drm, const char* device, const char* modeStr, unsigne
         {
             drmModeModeInfoPtr currentMode = &connector->modes[i];
 
-            if (std::strcmp(currentMode->name, modeStr) == 0)
+            if (std::string_view(currentMode->name) == modeStr)
             {
                 if (vrefresh == 0 || currentMode->vrefresh == vrefresh)
                 {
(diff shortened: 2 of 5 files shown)


ChrisThrasher:
Why couldn’t you find those other strcmp uses?

ChrisThrasher:

ChrisThrasher:
I double checked and `master` does include more uses of `std::strcmp` to remove so please fix those in this PR too.

litelawliet:
Thank you for the list, somehow VS couldn't find more references in the solution, I reconfigured the project via CMake and that should be all now

ChrisThrasher:

examples/vulkan/Vulkan.cpp
```diff
@@ -470,11 +470,11 @@ class VulkanExample
             // -- VK_LAYER_GOOGLE_unique_objects
             // These layers perform error checking and warn about bad or sub-optimal Vulkan API usage
             // VK_LAYER_LUNARG_monitor appends an FPS counter to the window title
-            if (!std::strcmp(layer.layerName, "VK_LAYER_LUNARG_standard_validation"))
+            if (std::string_view(layer.layerName) == std::string_view("VK_LAYER_LUNARG_standard_validation"))
```
ChrisThrasher, line 473:
```suggestion
            if (std::string_view(layer.layerName) == "VK_LAYER_LUNARG_standard_validation")
```
We can compare string views to `const char*`s whether they're Vulkan constant or string literals so we don't need to explicitly construct a second string view. This applies to many other places in this PR so I'll only mention it once.

ChrisThrasher (changes requested):

examples/vulkan/Vulkan.cpp
litelawliet: You're right, I completely overlook string literals in the process. Commit 2767be2e9e7961f7a4bc6032be9d957f6ed6c20c should fix those instances.

ChrisThrasher:

ChrisThrasher:
Please build the `format` target which will run clang-format and then those formatting CI jobs will start passing.

litelawliet:
Would it be possible to support LLVM 15 and above ? I just saw the project only support up to LLVM 14

ChrisThrasher:
No because clang-format-15 produces different formatted code than versions 12 through 14.

ChrisThrasher:

ChrisThrasher (approved):

src/SFML/Window/DRM/DRMContext.cpp
```diff
@@ -365,7 +365,7 @@ int initDrm(sf::priv::Drm& drm, const char* device, const char* modeStr, unsigne
         {
             drmModeModeInfoPtr currentMode = &connector->modes[i];
 
-            if (std::strcmp(currentMode->name, modeStr) == 0)
+            if (std::string_view(currentMode->name) == std::string_view(modeStr))
```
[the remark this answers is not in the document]
litelawliet: Right, fixed in d124e7164f9185f2550f8c03b4ec8b02dd2ca30e

kimci86 (approved):

ChrisThrasher:

ChrisThrasher:
Can we remove `#include <cstring>` from any of these files now?

ChrisThrasher:

litelawliet:
So after a few builds (sorry for that) that I couldn't test for Linux GCC and Clang, it appears the only
`#include <cstring>` we can remove are in `SFML\Window\Unix\VulkanImplX11.cpp` and `SFML\Window\Win32\VulkanImplWin32.cpp` files.

ChrisThrasher:

ChrisThrasher:
For whatever reason the CI pipeline isn't running? At least it appears so to me. Can you run `git commit --amend` then `git push --force` to change the commit hash and push that so CI gets triggered again? Unsure why it's acting up.

litelawliet:
Just did the two commands you suggested, doesn't look like CI is running ?

I generally interactively rebase my branch, choose to squash, fixup or edit commits I did (usually to have only one remaining) then force push the entire thing, could it introduce any problem for CI ?

ChrisThrasher:

ChrisThrasher:
> I generally interactively rebase my branch, choose to squash, fixup or edit commits I did (usually to have only one remaining) then force push the entire thing, could it introduce any problem for CI ?

This is perfectly fine. I think we just hit a weird one-time issue.

eXpl0it3r (approved):

Issue #2370, opened 2023-01-20
Write tests for types that require the presence of a display


ChrisThrasher:
## Subject of the issue

Thanks to #2302 we can now test types that previously were untestable without headless CI machines failing at runtime. Here's a (potentially incomplete) list of types for which we now can and should write tests.

Window
- [x] `sf::Context`
- [x] `sf::Cursor` (#2402)
- [ ] `sf::GlResource`
- [x] `sf::Window` (https://github.com/SFML/SFML/compare/master...ChrisThrasher:SFML:test_window)
- [x] `sf::WindowBase` (#2390)

Graphics
- [x] `sf::Drawable` (#2375)
- [x] `sf::Font` (#2605)
- [ ] `sf::RenderTarget`
- [ ] `sf::RenderTexture`
- [x] `sf::RenderWindow`
- [x] `sf::Shader`
- [ ] `sf::Sprite`
- [x] `sf::Text` (#2638)
- [x] `sf::Texture` (#2479)
- [x] `sf::VertexBuffer` (#2405)

As we did before, I'd prefer that tests for a given type not be written until tests for all dependent types are written so let's focus on testing types for which all of its dependencies are already tested. For example I want to see tests for `sf::Drawable` before tests for `sf::Text`.

Currently we just ignore testing interfaces that load files so you may continue to ignore that in the short term. I hope to figure out a clean way of organizing tests that need to load files since that incurs some additional requirements and I don't want to start doing that until we have a clear idea for how we're going to pull it off.

ChrisThrasher:
We now have tests for all these types 👍🏻

PR #2375, opened 2023-01-23
Test `sf::Drawable`

Jim-Marsden:
## Description

Adding tests for sf::Drawable

This PR is related to the issue #2370

### DIFF
--- a/test/Graphics/Drawable.test.cpp
+++ b/test/Graphics/Drawable.test.cpp
@@ -1,5 +1,9 @@
 #include <SFML/Graphics/Drawable.hpp>
+#include <SFML/Graphics/RenderTexture.hpp>
 
+#include <doctest/doctest.h>
+
+#include <GraphicsUtil.hpp>
 #include <type_traits>
 
 static_assert(!std::is_constructible_v<sf::Drawable>);
@@ -7,3 +11,41 @@ static_assert(!std::is_copy_constructible_v<sf::Drawable>);
 static_assert(std::is_copy_assignable_v<sf::Drawable>);
 static_assert(!std::is_nothrow_move_constructible_v<sf::Drawable>);
 static_assert(std::is_nothrow_move_assignable_v<sf::Drawable>);
+static_assert(std::is_abstract_v<sf::Drawable>);
+static_assert(std::has_virtual_destructor_v<sf::Drawable>);
+
+class DrawableTest : public sf::Drawable
+{
+public:
+    int callCount() const
+    {
+        return m_callCount;
+    }
+
+private:
+    void draw(sf::RenderTarget&, const sf::RenderStates&) const override
+    {
+        ++m_callCount;
+    }
+
+    mutable int m_callCount{};
+};
+
+TEST_CASE("[Graphics] sf::Drawable" * doctest::skip(skipDisplayTests))
+{
+    SUBCASE("Construction")
+    {
+        const DrawableTest drawableTest;
+        CHECK(drawableTest.callCount() == 0);
+    }
+
+    SUBCASE("draw()")
+    {
+        const DrawableTest drawableTest;
+        sf::RenderTexture  renderTexture;
+        CHECK(renderTexture.create({32, 32}));
+        CHECK(drawableTest.callCount() == 0);
+        renderTexture.draw(drawableTest);
+        CHECK(drawableTest.callCount() == 1);
+    }
+}


test/Graphics/Drawable.test.cpp
```diff
@@ -1,4 +1,7 @@
 #include <SFML/Graphics/Drawable.hpp>
+#include <SFML/Graphics/RenderTexture.hpp>
```
ChrisThrasher, line 2: Luckily `Drawable` comes before `RenderTarget` alphabetically but if the opposite was true then the header under test would not be compiled first meaning we wouldn't be able to detect missing `#include`s within Drawable.hpp. I hope we can find some way to enforce that the header under test always gets sorted first by clang-format.

test/Graphics/Drawable.test.cpp
```diff
@@ -7,3 +10,40 @@ static_assert(!std::is_copy_constructible_v<sf::Drawable>);
...
 static_assert(std::is_nothrow_move_assignable_v<sf::Drawable>);
+
+class DrawableTest : public sf::Drawable
+{
+public:
+    void draw(sf::RenderTarget&, const sf::RenderStates&) const override
+    {
+        m_called += 1;
+    }
+    int callCount() const
```
ChrisThrasher, line 21:
```suggestion

    int callCount() const
```

test/Graphics/Drawable.test.cpp
```diff
@@ -7,3 +10,40 @@ static_assert(!std::is_copy_constructible_v<sf::Drawable>);
...
+    {
+        return m_called;
+    }
+
+protected:
```
ChrisThrasher, line 26:
```suggestion
private:
```

test/Graphics/Drawable.test.cpp
```diff
@@ -7,3 +10,40 @@ static_assert(!std::is_copy_constructible_v<sf::Drawable>);
...
+    mutable int m_called{};
+};
+
+TEST_CASE("[Graphics] sf::Drawable")
+{
+    SUBCASE("Construction")
+    {
+        const DrawableTest drawableTest;
+        REQUIRE(drawableTest.callCount() == 0);
```
ChrisThrasher, line 35:
```suggestion
        CHECK(drawableTest.callCount() == 0);
```
Prefer using `CHECK` over `REQUIRE` since a failed `CHECK` will let the rest of the tests keep running. Only use `REQUIRE` if you're trying to guard against UB in the tests should a given assertion fail like ensuring the size of a container is greater than 0 before you try to access an element.

test/Graphics/Drawable.test.cpp
```diff
@@ -7,3 +10,40 @@ static_assert(!std::is_copy_constructible_v<sf::Drawable>);
...
+    SUBCASE("draw()")
+    {
+        const DrawableTest drawableTest;
+        sf::RenderTexture  renderTexture;
+        if (renderTexture.create({32, 32}))
+        {
+            REQUIRE(drawableTest.callCount() == 0);
+            renderTexture.draw(drawableTest);
+            REQUIRE(drawableTest.callCount() == 1);
+        }
```
ChrisThrasher, line 47: I think that `if` should become another test assertion so that if it fails, we don't just quietly stop running the tests contained within that block.

ChrisThrasher (changes requested):

test/Graphics/Drawable.test.cpp
```diff
@@ -7,3 +10,40 @@ static_assert(!std::is_copy_constructible_v<sf::Drawable>);
...
+    int callCount() const
+    {
+        return m_called;
+    }
+
+protected:
+    mutable int m_called{};
+};
+
+TEST_CASE("[Graphics] sf::Drawable")
```
ChrisThrasher, line 30: Go look at the `sf::Texture` tests to see how we skip display-dependent tests. There's some special code you need to add to this `TEST_CASE` call.

test/Graphics/Drawable.test.cpp
```diff
@@ -1,9 +1,49 @@
...
+    {
+        const DrawableTest drawableTest;
+        CHECK(drawableTest.callCount() == 0);
+    }
+
+    SUBCASE("draw()")
+    {
+        const DrawableTest drawableTest;
+        sf::RenderTexture  renderTexture;
+        CHECK(renderTexture.create({32, 32}));
```
ChrisThrasher, line 44: If this `.create` call fails, are the following tests still valid or do they become UB?

Jim-Marsden: Oh, you're right.

Jim-Marsden: It should be well defined because we aren't actually manipulating anything.

We might not even need to call create? Nothings being drawn, and nothing is being changed. It just needed a sf::RenderTarget that exists in memory (to be passed by reference).

https://github.com/SFML/SFML/blob/ca5ca650040580a15890136b953d82d51c4077a0/include/SFML/Graphics/RenderTarget.hpp#L249

https://github.com/SFML/SFML/blob/ca5ca650040580a15890136b953d82d51c4077a0/src/SFML/Graphics/RenderTarget.cpp#L247

test/Graphics/Drawable.test.cpp
Bromeon: If you add a comment in the lines in between the `#include`s, will clang-format still mess with the order?

ChrisThrasher: believe adding a comment will suffice but I haven't tested that yet

ChrisThrasher:

test/Graphics/Drawable.test.cpp
```diff
@@ -1,9 +1,49 @@
...
+        CHECK(drawableTest.callCount() == 0);
+        renderTexture.draw(drawableTest);
```
kimci86, line 46: Isn't this testing RenderTexture rather than Drawable?

ChrisThrasher: Somewhat, but how else do we test `Drawable`? We need something that call `.draw` on the drawable since users can't call that directly.

test/Graphics/Drawable.test.cpp
```diff
@@ -1,9 +1,49 @@
...
 static_assert(!std::is_constructible_v<sf::Drawable>);
 static_assert(!std::is_copy_constructible_v<sf::Drawable>);
 static_assert(std::is_copy_assignable_v<sf::Drawable>);
 static_assert(!std::is_nothrow_move_constructible_v<sf::Drawable>);
 static_assert(std::is_nothrow_move_assignable_v<sf::Drawable>);
+
+class DrawableTest : public sf::Drawable
+{
+public:
+    void draw(sf::RenderTarget&, const sf::RenderStates&) const override
```
ChrisThrasher, line 18: This function can be made private

test/Graphics/Drawable.test.cpp
kimci86: Drawable has no implementation, it is only an interface, so there is no behavior to assert.
However we can assert properties of the interface itself via type_traits such being abstract or having a virtual destructor, or implement some template magic to check that draw is virtual if we want to go that far. (But isn't it only repeating the type definition then?)

ChrisThrasher: Can we assert the presence of a protected pure virtual function?

test/Graphics/Drawable.test.cpp
```diff
@@ -1,9 +1,51 @@
...
+public:
+    int callCount() const
+    {
+        return m_called;
+    }
+
+private:
+    void draw(sf::RenderTarget&, const sf::RenderStates&) const override
+    {
+        m_called += 1;
```
ChrisThrasher, line 28:
```suggestion
        ++m_callCount;
```

test/Graphics/Drawable.test.cpp
```diff
@@ -1,9 +1,51 @@
...
+    }
+
+    mutable int m_called{};
```
ChrisThrasher, line 31:
```suggestion
    mutable int m_callCount{};
```

eXpl0it3r:

ChrisThrasher (approved):
Thanks for getting started with these display-dependent tests!

test/Graphics/Drawable.test.cpp
kimci86: > Can we assert the presence of a protected pure virtual function?

I don't know if it can be done. At least deriving the class lets us ensure `draw` is a virtual method by using `override` in the derived class.
Apart from that and static_asserts, we are actually testing the behavior of a test class. Is it useful? What error in Drawable can this detect? :thinking:

ChrisThrasher: > Is it useful?

Useful enough to warrant not just closing this PR.

> What error in Drawable can this detect? 🤔

Nothing that won't already be caught my the loads of other classes in the library that also derive from `sf::Drawable`. I know you don't think these tests are worthwhile but do you think this PR is worth closing or are you fine having some relatively low value tests in the library?

kimci86: As we discussed privately, those tests are not so much about catching bugs but rather being a live documentation. Then I understand the value :+1:

eXpl0it3r (approved):

PR #2378, opened 2023-01-25
[2124] FIX Window.setSize() while resizing not working

dogunbound:
I deleted my repository causing
https://github.com/SFML/SFML/pull/2316 to disappear. This is a new PR to resolve that.

## Description

This fix forces XResizeWindow to occur after the user is done resizing the window. If you use the test code I have provided, you will see that setSize is called after a resize event is done. Before this commit, setSize would run XResizeWindow, but since the user's mouse was still holding onto the window, the window would resize to whatever size the user's mouse is at. With my fix, XResizeWindow is GUARENTEED to occur once the user is done dragging the window.

Here is a video demonstrating my fix working: https://youtu.be/mLMVFbeuRIs

This PR is related to the issue #2124 

## Tasks

X11 LINUX ISSUE ONLY
* [x] Tested on Linux

## How to test this PR?

After you install SFML, go ahead and compile, link, and run the code below. You will see that the resize event is seemingly not called.

Install my fix, compile, link, and run the code below. You will see the resize event being called and resizing the window if you drag and drop the window quickly enough.

```cpp
#include <SFML/Graphics.hpp>
#include <SFML/Window.hpp>
#include <SFML/System.hpp>

int main() {

    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
            if (event.type == sf::Event::Resized) {
                const sf::Vector2u window_size(event.size.width, event.size.height);

                if (window_size.x < 200) {
                    window.setSize(sf::Vector2u(200, window_size.y));
                }
                if (window_size.y < 200) {
                    window.setSize(sf::Vector2u(window_size.x, 200));
                }
            }
        }

        window.clear();
        window.display();
    }

    return 0;
}
```

## SDL2 comparison:

SDL2 does the same thing. They loop X calls for the window set size function. https://github.com/libsdl-org/SDL/blob/main/src/video/x11/SDL_x11window.c

Scroll down to X11_SetWindowSize and at the end of the function you will see them looping X calls.

The end result is similar to what I am doing. You will notice that if you drag and drop an sdl window quickly enough, a forced resize can occur after a resize event has been called.

This is the min reproducible with sdl:
```c++
// SDL2 Hello, World!
// This should display a white screen for 2 seconds
// compile with: clang++ main.cpp -o hello_sdl2 -lSDL2
// run with: ./hello_sdl2
#include <SDL2/SDL.h>
#include <stdio.h>

#define SCREEN_WIDTH 640
#define SCREEN_HEIGHT 480

int main(int argc, char* args[]) {
    SDL_Window* window = NULL;
    SDL_Surface* screenSurface = NULL;
    if (SDL_Init(SDL_INIT_VIDEO) < 0) {
        fprintf(stderr, "could not initialize sdl2: %s\n", SDL_GetError());
        return 1;
    }
    window = SDL_CreateWindow(
            "hello_sdl2",
            SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED,
            SCREEN_WIDTH, SCREEN_HEIGHT,
            SDL_WINDOW_RESIZABLE
            );
    if (window == NULL) {
        fprintf(stderr, "could not create window: %s\n", SDL_GetError());
        return 1;
    }
    bool game_is_still_running = true;
    while (game_is_still_running) {
        SDL_Event event;
        while (SDL_PollEvent(&event)) {  // poll until all events are handled!
                                         // decide what to do with this event.
            switch (event.type) {
                case SDL_QUIT:
                    game_is_still_running = false;
                    break;
                case SDL_WINDOWEVENT: 
                    switch (event.window.event) {
                        case SDL_WINDOWEVENT_RESIZED:
                            {
                                const int new_w = event.window.data1, new_h = event.window.data2;
                                if (new_w < 200) {
                                    SDL_SetWindowSize(window, 200, new_h);
                                }
                                if (new_h < 200) {
                                    SDL_SetWindowSize(window, new_w, 200);
                                }
                                break;
                            }
                        default:
                            break;
                    }
                    break;
                default:
                    break;
            }
        }

        // update game state, draw the current frame
        screenSurface = SDL_GetWindowSurface(window);
        SDL_FillRect(screenSurface, NULL, SDL_MapRGB(screenSurface->format, 0xFF, 0xFF, 0xFF));
        SDL_UpdateWindowSurface(window);
    }
    SDL_DestroyWindow(window);
    SDL_Quit();
    return 0;
}
```

## ADDITIONAL NOTES:
I made a slight change from the original pull request. I added a 2-millisecond sleep during my X calls. That way my loop isn't slamming X calls. 

Edit:
I noticed the window felt funny to use with the sleep. Decided to get rid of that.

### DIFF
--- a/src/SFML/Window/Unix/WindowImplX11.cpp
+++ b/src/SFML/Window/Unix/WindowImplX11.cpp
@@ -882,7 +882,6 @@ void WindowImplX11::setSize(const Vector2u& size)
         (loopStartTimer.getElapsedTime() < sf::milliseconds(50))) {
         processEvents();
         XResizeWindow(m_display, m_window, size.x, size.y);
-        sf::sleep(sf::milliseconds(2));
     }
     XFlush(m_display);
 }
--- a/src/SFML/Window/Unix/WindowImplX11.cpp
+++ b/src/SFML/Window/Unix/WindowImplX11.cpp
@@ -889,7 +889,11 @@ void WindowImplX11::setSize(const Vector2u& size)
         XSync(m_display, False);
         sf::sleep(sf::milliseconds(10));
     }
+    // Flush all of the queued resizes to prevent the window from spazzing.
     XSync(m_display, True);
+    // Do one last resize to ensure the correct resize is queued
+    // This just makes our setSize a bit less lossy than SDL
+    XResizeWindow(m_display, m_window, size.x, size.y);
     XFlush(m_display);
 }
 
--- a/src/SFML/Window/Unix/WindowImplX11.cpp
+++ b/src/SFML/Window/Unix/WindowImplX11.cpp
@@ -873,18 +873,23 @@ void WindowImplX11::setSize(const Vector2u& size)
         XFree(sizeHints);
     }
 
-    // Sometimes, Xorg doesn't want to let the XResizeWindow function call to work.
-    // Looping is an attempt to convice Xorg to allow this resize to occur.
-    // If xorg doesn't allow this resize after 100 milliseconds, or another resize occurs, quit trying.
     Vector2u original_size = getSize();
+    XSync(m_display, False);
+    XResizeWindow(m_display, m_window, size.x, size.y);
+
+    // Sometimes, Xorg doesn't want to let the XResizeWindow function call to
+    // work. Looping is an attempt to convice Xorg to allow this resize to
+    // occur. If xorg doesn't allow this resize after 100 milliseconds, or
+    // another resize occurs, quit trying.
     sf::Clock cancellationTimer;
     while((getSize() != size) &&
         getSize() == original_size &&
         (cancellationTimer.getElapsedTime() < sf::milliseconds(100))) 
     {
-        XResizeWindow(m_display, m_window, size.x, size.y);
+        XSync(m_display, False);
         sf::sleep(sf::milliseconds(10));
     }
+    XSync(m_display, True);
     XFlush(m_display);
 }
 
--- a/src/SFML/Window/Unix/WindowImplX11.cpp
+++ b/src/SFML/Window/Unix/WindowImplX11.cpp
@@ -35,6 +35,8 @@
 #include <SFML/System/Mutex.hpp>
 #include <SFML/System/Lock.hpp>
 #include <SFML/System/Sleep.hpp>
+#include <SFML/System/Time.hpp>
+#include <SFML/System/Clock.hpp>
 #include <bitset> // <X11/Xlibint.h> defines min/max as macros, so <bitset> has to come before that
 #include <X11/Xlibint.h>
 #include <X11/Xutil.h>
@@ -385,6 +387,7 @@ m_hiddenCursor   (0),
 m_lastCursor     (None),
 m_keyRepeat      (true),
 m_previousSize   (-1, -1),
+m_resizeOccuredDuringSetSize(false),
 m_useSizeHints   (false),
 m_fullscreen     (false),
 m_cursorGrabbed  (false),
@@ -436,6 +439,7 @@ m_hiddenCursor   (0),
 m_lastCursor     (None),
 m_keyRepeat      (true),
 m_previousSize   (-1, -1),
+m_resizeOccuredDuringSetSize(false),
 m_useSizeHints   (false),
 m_fullscreen     ((style & Style::Fullscreen) != 0),
 m_cursorGrabbed  (m_fullscreen),
@@ -871,7 +875,15 @@ void WindowImplX11::setSize(const Vector2u& size)
         XFree(sizeHints);
     }
 
-    XResizeWindow(m_display, m_window, size.x, size.y);
+    m_resizeOccuredDuringSetSize = false;
+    sf::Clock loopStartTimer;
+    while((getSize() != size) &&
+        !m_resizeOccuredDuringSetSize &&
+        (loopStartTimer.getElapsedTime() < sf::milliseconds(50))) {
+        processEvents();
+        XResizeWindow(m_display, m_window, size.x, size.y);
+        sf::sleep(sf::milliseconds(2));
+    }
     XFlush(m_display);
 }
 
@@ -1735,6 +1747,8 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
 
                 m_previousSize.x = windowEvent.xconfigure.width;
                 m_previousSize.y = windowEvent.xconfigure.height;
+
+                m_resizeOccuredDuringSetSize = true;
             }
             break;
         }
--- a/src/SFML/Window/Unix/WindowImplX11.cpp
+++ b/src/SFML/Window/Unix/WindowImplX11.cpp
@@ -873,7 +873,7 @@ void WindowImplX11::setSize(const Vector2u& size)
         XFree(sizeHints);
     }
 
-    Vector2u original_size = getSize();
+    Vector2u originalSize = getSize();
     XSync(m_display, False);
     XResizeWindow(m_display, m_window, size.x, size.y);
 
@@ -883,7 +883,7 @@ void WindowImplX11::setSize(const Vector2u& size)
     // another resize occurs, quit trying.
     sf::Clock cancellationTimer;
     while((getSize() != size) &&
-        getSize() == original_size &&
+        getSize() == originalSize &&
         (cancellationTimer.getElapsedTime() < sf::milliseconds(100))) 
     {
         XSync(m_display, False);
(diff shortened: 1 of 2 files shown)


eXpl0it3r:

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -871,7 +875,14 @@ void WindowImplX11::setSize(const Vector2u& size)
         XFree(sizeHints);
     }
 
-    XResizeWindow(m_display, m_window, size.x, size.y);
+    m_resizeOccuredDuringSetSize = false;
+    sf::Clock loopStartTimer;
+    while((getSize() != size) &&
+        !m_resizeOccuredDuringSetSize &&
+        (loopStartTimer.getElapsedTime() < sf::milliseconds(50))) {
```
eXpl0it3r, line 882: `{` should be on a new line

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -871,7 +875,14 @@ void WindowImplX11::setSize(const Vector2u& size)
         XFree(sizeHints);
     }
 
-    XResizeWindow(m_display, m_window, size.x, size.y);
+    m_resizeOccuredDuringSetSize = false;
+    sf::Clock loopStartTimer;
```
eXpl0it3r, line 879: I'd call it `cancelationTimer`

src/SFML/Window/Unix/WindowImplX11.hpp
```diff
@@ -300,25 +300,26 @@ class WindowImplX11 : public WindowImpl
...
+    XIM        m_inputMethod;                 ///< Input method linked to the X display
+    XIC        m_inputContext;                ///< Input context used to get unicode input in our window
+    bool       m_isExternal;                  ///< Tell whether the window has been created externally or by SFML
+    RRMode     m_oldVideoMode;                ///< Video mode in use before we switch to fullscreen
+    RRCrtc     m_oldRRCrtc;                   ///< RRCrtc in use before we switch to fullscreen
+    ::Cursor   m_hiddenCursor;                ///< As X11 doesn't provide cursor hiding, we must create a transparent one
+    ::Cursor   m_lastCursor;                  ///< Last cursor used -- this data is not owned by the window and is required to be always valid
+    bool       m_keyRepeat;                   ///< Is the KeyRepeat feature enabled?
+    Vector2i   m_previousSize;                ///< Previous size of the window, to find if a ConfigureNotify event is a resize event (could be a move event only)
+    bool       m_resizeOccuredDuringSetSize;  ///< if resize occurs during setSize, setSize knows to stop looping XResizeWindow
```
eXpl0it3r, line 315: I'd simplify `m_resizeOccured` and then as documentation `Set during resize, used to signal setSize that resize occurred`

src/SFML/Window/Unix/WindowImplX11.cpp
eXpl0it3r, line 882: Looks like SDL is doing 100ms (not 10ms as you said originally), I'd also go with 100ms then 😄

eXpl0it3r (changes requested):
Some small changes, after that, this should be ready 🙂

kimci86:
I just tested it on Ubuntu 22.04 and it does not work for me. I will try to give more details soon.

dogunbound:
https://youtu.be/xGN3NRxceR8 

@eXpl0it3r it would be best if you watched that. I personally think that we should ignore this very far stretch of an edge case I managed to stumble upon, but this isn't up to me. The great SFML maintainers are the last ones to make a decision. 

On another note, I did manage to find a way to isolate everything into the setSize function. I got rid of the `m_resizeOccured` variable! This is an improvement to the readability of the code as it isolates logic into one function.

kimci86:
I tested again. At least it doesn't freeze my desktop this time, but it does not solve the issue for me. It makes the window spasm or grow. If using 2.6.x instead, the window can grow but not spasm, which is not worse.

You mention SDL does the same, but if I understand correctly, it does not. The looping is only to get and cache the actual window size, not to set it.
https://github.com/libsdl-org/SDL/blob/02fbf2cf991bf8310d1a6400abaf05ff5ffff52c/src/video/x11/SDL_x11window.c#L967-L968

I also had a quick look at other implementation.
- Godot does a similar thing than SDL: https://github.com/godotengine/godot/blob/0b1d516f67a3b2b7dd158b923559f192ec103a85/platform/linuxbsd/x11/display_server_x11.cpp#L1826-L1838
- GLFW doesn't have such loop: https://github.com/glfw/glfw/blob/00967cbb8a4a51ea8ff6f8b8a7efcf1f761cc664/src/x11_window.c#L2216

Overall, I think we are trying to solve a problem with the wrong tool. The original problem is to force a window to have a minimal size. A proper solution is to ask the window manager to take care of it.
This is implemented in other libraries: [SDL](https://github.com/libsdl-org/SDL/blob/02fbf2cf991bf8310d1a6400abaf05ff5ffff52c/src/video/x11/SDL_x11window.c#L847-L873), [GLFW](https://github.com/glfw/glfw/blob/00967cbb8a4a51ea8ff6f8b8a7efcf1f761cc664/src/x11_window.c#L2222-L2230), [Godot](https://github.com/godotengine/godot/blob/0b1d516f67a3b2b7dd158b923559f192ec103a85/platform/linuxbsd/x11/display_server_x11.cpp#L1635-L1672).

dogunbound:
> I tested again. At least it doesn't freeze my desktop this time, but it does not solve the issue for me. It makes the window spasm or grow. If using 2.6.x instead, the window can grow but not spasm, which is not worse.
> 
> You mention SDL does the same, but if I understand correctly, it does not. The looping is only to get and cache the actual window size, not to set it. https://github.com/libsdl-org/SDL/blob/02fbf2cf991bf8310d1a6400abaf05ff5ffff52c/src/video/x11/SDL_x11window.c#L967-L968
> 
> I also had a quick look at other implementations.
> 
>     * Godot does a similar thing than SDL: https://github.com/godotengine/godot/blob/0b1d516f67a3b2b7dd158b923559f192ec103a85/platform/linuxbsd/x11/display_server_x11.cpp#L1826-L1838
> 
>     * GLFW doesn't have such a loop: https://github.com/glfw/glfw/blob/00967cbb8a4a51ea8ff6f8b8a7efcf1f761cc664/src/x11_window.c#L2216
> 
> 
> Overall, I think we are trying to solve a problem with the wrong tool. The original problem is to force a window to have a minimal size. A proper solution is to ask the window manager to take care of it. This is implemented in other libraries: [SDL](https://github.com/libsdl-org/SDL/blob/02fbf2cf991bf8310d1a6400abaf05ff5ffff52c/src/video/x11/SDL_x11window.c#L847-L873), [GLFW](https://github.com/glfw/glfw/blob/00967cbb8a4a51ea8ff6f8b8a7efcf1f761cc664/src/x11_window.c#L271-L318), [Godot](https://github.com/godotengine/godot/blob/0b1d516f67a3b2b7dd158b923559f192ec103a85/platform/linuxbsd/x11/display_server_x11.cpp#L1635-L1672).

I got rid of the spazzing by forcing xorg server to clear all the queued resizes.

ChrisThrasher:

dogunbound:
After more experimenting, there is no way to set the minimum size of your window in Linux reliably. 

If you use this method in Wayland, your window will start teleporting (this is before and after my fix. SDL does this too). If you use this in x11, it has a chance of straight up not working.

All my fix does is give x11 users a chance for the setSize to stick. The current version of my fix will have slight performance degradation to the setSize function as it will "freeze" the app until the resize is done, or fails after 100ms. That means tiling window managers who disallow resize to occur will probably notice a 100ms freeze if they try to reduce the window size below the minimum and readjust the window to the correct parameters in such a way. 

The real solution is to give x11 hints as to what the desired minimum size of your window is. If you want to see this, just let me know and I'll start hacking a set minimum size for windows in x11. GLFW, SDL, godot, and other engines have this feature probably because of situations like this.

eXpl0it3r:
Since this seems to be limited to a specific window manager setup and the given fix doesn't fully resolve the issue, while adding more complexity and also potentially some odd performance issue, I'd rather not merge this for SFML 2.6

We may experiment some further for SFML 2.6.1 (assuming that we will have a patch release), but I think, as you said yourself, the better option would be to have a min/max size limit. I've opened [a discussion on the forum](https://en.sfml-dev.org/forums/index.php?topic=28956.0), so feel free to chip in on that as well.

dogunbound:
This bug will stay prevalent forever, but we will implement https://github.com/SFML/SFML/pull/2519 as a way to remedy this issue.

Issue #2290, opened 2022-12-01
sf::Packet: Floats are not in network byte order


Nekotekina:
They are serialized in native endianness. It looks problematic for interoperability between machines with different endianness.

Bambo-Borris:
Do you have some evidence for this, or a minimal reproducibility case so that it's possible for someone to check this? More information would be helpful otherwise this issue looks a little barebones for someone who would like to resolve it.

Nekotekina:
Unfortunately, I was just reading the source code. It would be hard to reproduce.

Bambo-Borris:
> Unfortunately, I was just reading the source code. It would be hard to reproduce.

Any additional information (even a link to the line in a SFML source file) would be more useful than the just a couple of sentences.

Nekotekina:
Here are the functions which read float/double:
https://github.com/SFML/SFML/blob/7faa5507a4ccc60dd669dbc0156db2f03238d1c6/src/SFML/Network/Packet.cpp#L257-L280

For comparison, reading uint64:
https://github.com/SFML/SFML/blob/7faa5507a4ccc60dd669dbc0156db2f03238d1c6/src/SFML/Network/Packet.cpp#L235-L254

Integral reads convert from network byte order to native. Floating points don't do that.

gnawme:
Addressed in my branch: `feature/use-generic-endian-swap`; PR to come

JonnyPtn:
For clarity - the int functions don't do anything to help interoperability between different endian platforms. They assume little endian on both sides, and just convert to big endian while in transit (which is wasted work in most cases)

Interop between big and little endian machines is not something SFML currently supports and I don't think it should (but it also doesn't stop users from doing it if needed, which would be an edge case)

PR #2380, opened 2023-01-27
Switched to generic endian swap

gnawme:

----

## Description
Switched to use [generic (templated) endian swap method](https://mklimenko.github.io/english/2018/08/22/robust-endian-swap/) that I've used successfully in the past.

This PR is related to the issue #2290

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?
Extended the `Packet` stream tests to include `float` and `double`


include/SFML/Network/Socket.hpp
```diff
@@ -31,13 +31,37 @@
...
 {
 class SocketSelector;
 
+////////////////////////////////////////////////////////////
+/// \brief Robust endianness conversion
+/// \see https://mklimenko.github.io/english/2018/08/22/robust-endian-swap/
+///
+////////////////////////////////////////////////////////////
+template <typename T>
+T swapEndian(T& val, typename std::enable_if<std::is_arithmetic<T>::value, std::nullptr_t>::type = nullptr)
```
ChrisThrasher, line 50: Why a mutable l-value reference? Why not just copy it? We're only talking about types up to 8 bytes so it's not like that copy is any worse than taking a reference.

include/SFML/Network/Socket.hpp
```diff
@@ -31,13 +31,37 @@
...
+{
+    union U
```
ChrisThrasher, line 52: I believe we can avoid giving this union a name

include/SFML/Network/Socket.hpp
```diff
@@ -31,13 +31,37 @@
...
+    {
+        T                                   val;
+        std::array<std::uint8_t, sizeof(T)> raw;
+    } src, dst;
+
+    src.val = val;
+    std::reverse_copy(src.raw.begin(), src.raw.end(), dst.raw.begin());
```
ChrisThrasher, line 59: Doesn't this constitute type punning through a union and thus is UB?

include/SFML/Network/Socket.hpp
ChrisThrasher, line 50: This ought to be moved to the .cpp file so that it doesn't become a new part of the API.

src/SFML/Network/Packet.cpp
```diff
@@ -200,7 +200,7 @@ Packet& Packet::operator>>(std::uint32_t& data)
     if (checkSize(sizeof(data)))
     {
         std::memcpy(&data, &m_data[m_readPos], sizeof(data));
-        data = ntohl(data);
+        data = sf::swapEndian(data);
```
ChrisThrasher, line 203:
```suggestion
        data = swapEndian(data);
```
You can drop the `sf::` namespace whenever you're calling this.

test/Network/Packet.test.cpp
```diff
@@ -97,5 +97,21 @@ TEST_CASE("[Network] sf::Packet")
             CHECK_PACKET_STREAM_OPERATORS(std::numeric_limits<std::int64_t>::min());
             CHECK_PACKET_STREAM_OPERATORS(std::numeric_limits<std::int64_t>::max());
         }
+
+        SUBCASE("float")
+        {
+            CHECK_PACKET_STREAM_OPERATORS(float(0.f));
+            CHECK_PACKET_STREAM_OPERATORS(float(1.f));
```
ChrisThrasher, line 104:
```suggestion
            CHECK_PACKET_STREAM_OPERATORS(0.f);
            CHECK_PACKET_STREAM_OPERATORS(1.f);
```

test/Network/Packet.test.cpp
```diff
@@ -97,5 +97,21 @@ TEST_CASE("[Network] sf::Packet")
...
+            CHECK_PACKET_STREAM_OPERATORS(std::numeric_limits<float>::min());
+            CHECK_PACKET_STREAM_OPERATORS(std::numeric_limits<float>::max());
+        }
+
+        SUBCASE("double")
+        {
+            CHECK_PACKET_STREAM_OPERATORS(double(0.0));
+            CHECK_PACKET_STREAM_OPERATORS(double(1.0));
```
ChrisThrasher, line 112:
```suggestion
            CHECK_PACKET_STREAM_OPERATORS(0.0);
            CHECK_PACKET_STREAM_OPERATORS(1.0);
```

ChrisThrasher (changes requested):

include/SFML/Network/Socket.hpp
ChrisThrasher, line 50: I don't think we need SFINAE here and can instead just use a `static_assert`. SFINAE is great when dealing with overload sets but we don't have an overload set here. We just need some way to break compilation if an unacceptable type is used.

ChrisThrasher:
If I understand correct, this PR would make sure bytes always get swapped around but what about the case where the machine's endianness matches network endianness? In that case we don't need to swap bytes around but this would seemingly still swap them.

ChrisThrasher:
https://github.com/ChrisThrasher/SFML/commit/26ae62e694ef7866a1fe69d1a0ef91f902b973ae

This inspired me to write tests that make sure we're using the correct network ordering. Let me know if you can run your code with these tests and see if anything fails.

JonnyPtn:
I'm not convinced SFML should be doing any reordering - in the vast majority of cases it's going to be little endian machines on both sides, using SFML sockets. In which cases the conversion is just redundant extra work for both sides

I think the only use case this benefits would be people communicating with other services they have no control over, in which case they can still do the conversion themselves?

At best this should be optional, and off by default

gnawme:
> [ChrisThrasher@26ae62e](https://github.com/ChrisThrasher/SFML/commit/26ae62e694ef7866a1fe69d1a0ef91f902b973ae)
> 
> This inspired me to write tests that make sure we're using the correct network ordering. Let me know if you can run your code with these tests and see if anything fails.

Where can I access these tests?

include/SFML/Network/Socket.hpp
gnawme: Switched to not use a union.

ChrisThrasher:
https://github.com/SFML/SFML/compare/master...ChrisThrasher:SFML:test_network_order

include/SFML/Network/Socket.hpp
```diff
@@ -31,12 +31,55 @@
...
 namespace sf
 {
 class SocketSelector;
+////////////////////////////////////////////////////////////
+/// \brief Endian test
+///
+////////////////////////////////////////////////////////////
+inline bool isBigEndian()
+{
+    int one{1};
```
ChrisThrasher, line 50:
```suggestion
    int one = 1;
```
We don't use braces for initialization in this context.

include/SFML/Network/Socket.hpp
```diff
@@ -31,12 +31,55 @@
...
+T swapEndian(T val)
+{
+    if (isBigEndian())
+    {
+        return val;
+    }
+
+    std::array<std::uint8_t, sizeof(T)> raw;
+
+    auto* cast{reinterpret_cast<std::uint8_t*>(&val)};
```
ChrisThrasher, line 71:
```suggestion
    auto* cast = reinterpret_cast<std::uint8_t*>(&val);
```

include/SFML/Network/Socket.hpp
```diff
@@ -31,12 +31,55 @@
...
+    std::memcpy(raw.data(), cast, sizeof(T));
+
+    std::array<std::uint8_t, sizeof(T)> dst;
+    std::reverse_copy(raw.begin(), raw.end(), dst.begin());
+    std::memcpy(cast, dst.data(), sizeof(T));
+
+    auto* valp{reinterpret_cast<T*>(cast)};
```
ChrisThrasher, line 78:
```suggestion
    auto* valp = reinterpret_cast<T*>(cast);
```

src/SFML/Network/IpAddress.cpp
```diff
@@ -62,8 +62,8 @@ std::optional<IpAddress> IpAddress::resolve(std::string_view address)
         return Any;
 
     // Try to convert the address as a byte representation ("xxx.xxx.xxx.xxx")
-    if (const std::uint32_t ip = inet_addr(address.data()); ip != INADDR_NONE)
-        return IpAddress(ntohl(ip));
+    if (auto ip = inet_addr(address.data()); ip != INADDR_NONE)
```
ChrisThrasher, line 65:
```suggestion
    if (const auto ip = inet_addr(address.data()); ip != INADDR_NONE)
```

src/SFML/Network/IpAddress.cpp
```diff
@@ -75,10 +75,10 @@ std::optional<IpAddress> IpAddress::resolve(std::string_view address)
         sockaddr_in sin;
         std::memcpy(&sin, result->ai_addr, sizeof(*result->ai_addr));
 
-        const std::uint32_t ip = sin.sin_addr.s_addr;
+        auto ip = sin.sin_addr.s_addr;
```
ChrisThrasher, line 78:
```suggestion
        const auto ip = sin.sin_addr.s_addr;
```

src/SFML/Network/IpAddress.cpp
```diff
@@ -111,7 +111,8 @@ std::string IpAddress::toString() const
 ////////////////////////////////////////////////////////////
 std::uint32_t IpAddress::toInteger() const
 {
-    return ntohl(m_address);
+    auto address{m_address};
+    return swapEndian(address);
```
ChrisThrasher, line 115:
```suggestion
    return swapEndian(m_address);
```

src/SFML/Network/IpAddress.cpp
```diff
@@ -128,7 +129,8 @@ std::optional<IpAddress> IpAddress::getLocalAddress()
         return std::nullopt;
 
     // Connect the socket to localhost on any port
-    sockaddr_in address = priv::SocketImpl::createAddress(ntohl(INADDR_LOOPBACK), 9);
+    std::uint32_t kcabpool{INADDR_LOOPBACK};
```
ChrisThrasher, line 132:
```suggestion
    std::uint32_t kcabpool = INADDR_LOOPBACK;
```

include/SFML/Network/Socket.hpp
```diff
@@ -31,12 +31,55 @@
...
+
+
+////////////////////////////////////////////////////////////
+/// \brief Generic endianness conversion; if host is big-endian,
+/// incoming network byte order can be handled without swap, and
+/// host byte order is already network byte order
+///
+////////////////////////////////////////////////////////////
+template <typename T>
+T swapEndian(T val)
```
ChrisThrasher, line 62:
```suggestion
[[nodiscard]] T swapEndian(T val)
```

ChrisThrasher (changes requested):
A lot of these functions are collapsing down to nearly identical code. I wonder if this new pattern means we can replace a lot of these function overloads with a template and cut out hundreds of lines of code.

src/SFML/Network/IpAddress.cpp
ChrisThrasher: Now that `swapEndian` no longer takes a mutable reference, we can put `const` back where it was removed.

src/SFML/Network/TcpSocket.cpp
```diff
@@ -68,7 +68,7 @@ unsigned short TcpSocket::getLocalPort() const
         priv::SocketImpl::AddrLength size = sizeof(address);
         if (getsockname(getHandle(), reinterpret_cast<sockaddr*>(&address), &size) != -1)
         {
-            return ntohs(address.sin_port);
+            return swapEndian(address.sin_port);
```
ChrisThrasher, line 71: I don't think we necessarily need to stop using `ntohs` and related functions. The motivation for this PR is to fix issues with how `sf::Packet` handles floats so I'd rather not touch other classes.

gnawme:
> A lot of these functions are collapsing down to nearly identical code. I wonder if this new pattern means we can replace a lot of these function overloads with a template and cut out hundreds of lines of code.

I was thinking about that. Is there a consistent pattern that can be applied to character strings? Those are also not converted to network byte order.

src/SFML/Network/TcpSocket.cpp
gnawme: Reverted all except `Packet.cpp`

src/SFML/Network/Packet.cpp
```diff
@@ -29,12 +29,57 @@
...
+    auto* cast = reinterpret_cast<std::uint8_t*>(&val);
+    std::memcpy(raw.data(), cast, sizeof(T));
+
+    std::array<std::uint8_t, sizeof(T)> dst;
+    std::reverse_copy(raw.begin(), raw.end(), dst.begin());
+    std::memcpy(cast, dst.data(), sizeof(T));
+
+    auto* valp = reinterpret_cast<T*>(cast);
+
+    return *valp;
```
ChrisThrasher, line 78: If we're copying into the address of `val`, can't we then just return `val` whose bytes have just been rewritten? As written this appears to be starting the lifetime of a type with `reinterpret_cast` which is UB.

src/SFML/Network/Packet.cpp
```diff
@@ -29,12 +29,57 @@
...
+
+
+////////////////////////////////////////////////////////////
+/// \brief Generic endianness conversion; if host is big-endian,
+/// incoming network byte order can be handled without swap, and
+/// host byte order is already network byte order
+///
+////////////////////////////////////////////////////////////
+template <typename T>
+T swapEndian(T val)
```
ChrisThrasher, line 60:
```suggestion
T swapEndian(T value)
```
There's no need to abbreviate words that are already 5 characters. We allow for 120 character columns so vertical space is not a precious resource. Same goes for other abbreviated variable names. Please be descriptive.

ChrisThrasher:
> Is there a consistent pattern that can be applied to character strings? Those are also not converted to network byte order.

Each character is one byte which means the endianness doesn't really matter. If each character is >1 byte then I'd expect each character to get its bytes rearranged but not to reverse the string itself.

ChrisThrasher:
I'm going to pause on the PR feedback until some other maintainers can weigh in on whether they like reimplementing `sf::Packet`. I believe that we can use a template with a closed set of specializations to simplify how many of these `<<` and `>>` overloads are implemented much like how we explicitly specialize the `sf::Vector2<T>` template for floating point types to avoid having to put quite so much of that code in the header.

ChrisThrasher:

gnawme:
> believe that we can use a template with a closed set of specializations

I imagine that would be the subject of a separate PR

ChrisThrasher:
> I imagine that would be the subject of a separate PR

I see it as part of justifying this PR. It's easier to get this approved and merged this if it also means getting to delete huge swaths of code.

gnawme:
> I see it as part of justifying this PR.

Any sense of when and how the maintainers want to move forward?

ChrisThrasher:
> Any sense of when and how the maintainers want to move forward?

Sorry I'm not sure.

ZXShady:
this is UB you cannot just access 2 union members at a time

```cpp
template <typename T>
void SwapEndian(T &val) {
    union U {
        T val;
        std::array<std::uint8_t, sizeof(T)> raw; // UB!
    } src, dst;

    src.val = val; // UB
    std::reverse_copy(src.raw.begin(), src.raw.end(), dst.raw.begin());
    val = dst.val; // UB
}
```

it is UB to access 2 union members in C++ by the standard it is allowed in C (since C99 and explicitly allowed in C11) but not C++ at all but all compilers make this behavoir defined but I would not like SFML to rely on possible UB even if it is supported

PR #2381, opened 2023-01-29
Use Ninja in CI

ChrisThrasher:
## Description

Ninja automatically builds in parallel and is generally faster than Makefiles. Given lots of conflating variables that affect CI runtime, it's hard to get precise measurements but a good way to observe the difference is to look at the Linux and macOS build job times between [master](https://github.com/ChrisThrasher/SFML/actions/runs/4021739056) and [this PR](https://github.com/ChrisThrasher/SFML/actions/runs/4034738927).

`master`:
Screenshot 2023-01-28 at 8 08 09 PM

With Ninja:
Screenshot 2023-01-28 at 8 09 24 PM

Note that some jobs like the iOS builds and of course the Xcode build won't use Ninja. While we can use Ninja on Windows (and for the Clang jobs) we still want to explicitly test VS 2019 and VS 2022 so those won't be changed or removed. Let me know if I missed any jobs that could use Ninja but don't.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,9 +17,9 @@ jobs:
         - { name: Windows VS2022 x64,     os: windows-2022, flags: -A x64 }
         - { name: Windows VS2022 ClangCL, os: windows-2022, flags: -T ClangCL }
         - { name: Windows VS2022 Clang,   os: windows-2022, flags: -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -GNinja }
-        - { name: Linux GCC,            os: ubuntu-latest, prefix: xvfb-run -a }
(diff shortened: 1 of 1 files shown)


ChrisThrasher:

.github/workflows/ci.yml
```diff
@@ -75,8 +75,8 @@ jobs:
         sudo apt-get install gcovr ${{ matrix.platform.name == 'Linux Clang' && 'llvm-$CLANG_VERSION' || '' }}
 
     - name: Install macOS Tools
-      if: matrix.type.name == 'Debug' && runner.os == 'macOS'
```
eXpl0it3r, line 78: I guess there's no noticeable time penalty in installing gcovr even when we don't use it in release mode?

eXpl0it3r (approved):
That great! 🙂

.github/workflows/ci.yml
ChrisThrasher: That's what I was banking on. I tried installing it locally and it only took 8 seconds so I thought we could get away with always doing that.

PR #2386, opened 2023-01-31
Use `std::string_view` for parameter type

ChrisThrasher:
## Description

We can use a slightly more expressive type for this internal interface. I'm trying to find more places where we can replace `const char*` with more expressive and safer types.


ChrisThrasher:

eXpl0it3r (approved):
Looks good, did you try and make one of the calls fail, to see if it works?

ChrisThrasher:
Screenshot 2023-01-31 at 4 38 56 PM

I hacked on `glCheckError` to always print errors then ran the Island example and got all these errors. If I add that same hack to `master` I get the same console output so it seems like the console output has not been changed.

PR #2390, opened 2023-02-05
Add tests for `sf::WindowBase`

ChrisThrasher:
## Description

Unit tests are back on the menu!

The first commit fixes an oversight from #2302 where I forgot to enable these display tests for the Linux jobs that are already using `xvfb-run`. The second commit fixes a DRM bug where the code will attempt to terminate a null display causing a segfault. I'm not sure why this only appeared in these tests but it did. 

The main commit adds tests for `sf::WindowBase`. Certain APIs had inconsistent behavior across platforms so those tests were omitted. For example, the value of `sf::WindowBase::hasFocus` is not consistent. Other APIs like `sf::WindowBase::setMouseCursor` result in no observable changes we could test for so I didn't add tests for those.

It's important to realize that going forward we'll still need to manually build and run tests with the DRM backend since we can't run these display tests with the DRM backend in CI.


ChrisThrasher:

eXpl0it3r (approved):

PR #2391, opened 2023-02-05
Update versions in Config.hpp to 2.6.0

bluszcz:


ChrisThrasher:
https://github.com/SFML/SFML/commit/0dec051a759637c6b888109c3159709a85406188

This has already been taken care of in the `feature/prep_next_release` branch.

eXpl0it3r:
Thanks for the PR. We're all looking forward to up that version number. As mentioned, I've already prepared that a while ago and we're waiting for the final bugfix(es).

The version handling is a bit different between SFML 2 and will be for SFML 3, so we don't want to raise the version before the actual release. 🙂

PR #2402, opened 2023-02-12
Add tests for `sf::Cursor`

ChrisThrasher:
## Description

Fun fact, that trivial constructor test would have caught the fundamental issue with #2312. Until now we never constructed `sf::Cursor` in the whole codebase so accidentally making it non-destructible wouldn't have been detected. Goes to show how even the simplest of tests can still add value.


ChrisThrasher:

ChrisThrasher:

test/Window/Cursor.test.cpp
```diff
@@ -1,8 +1,36 @@
...
+{
+    SUBCASE("Construction")
+    {
+        const sf::Cursor cursor;
+    }
+
+    SUBCASE("loadFromPixels()")
+    {
+        sf::Cursor   cursor;
+        std::uint8_t pixels[4];
```
eXpl0it3r, line 23: Doesn't this array need some initialization?
What about using `std::array`? 🤔

ChrisThrasher: I figured it was fine letting these array elements be junk since we’re not testing anything about the particular pixel values. I can switch to std::array and initialize them if you want.

ChrisThrasher:

test/Window/Cursor.test.cpp
eXpl0it3r: Tests can be considered as sort of documentation, so I think it makes sense to initialize stuff and use modern constructs 🙂

ChrisThrasher:

ChrisThrasher:

test/Window/Cursor.test.cpp
```diff
@@ -1,8 +1,37 @@
...
+        CHECK(!cursor.loadFromPixels(nullptr, {}, {}));
+        CHECK(!cursor.loadFromPixels(pixels.data(), {0, 1}, {}));
+        CHECK(!cursor.loadFromPixels(pixels.data(), {1, 0}, {}));
+        CHECK(cursor.loadFromPixels(pixels.data(), {1, 1}, {}));
+    }
+
+    SUBCASE("loadFromSystem()")
+    {
+        sf::Cursor cursor;
+        CHECK(cursor.loadFromSystem(sf::Cursor::Hand));
```
eXpl0it3r, line 35: What about testing all the cursors that are defined to be available on all platforms? 🤔

ChrisThrasher: I suppose I can try. I can at least test more than this one.

ChrisThrasher:

eXpl0it3r (approved):

Issue #2403, opened 2023-02-12
OpenGL extension string is sometimes null causing crash


ChrisThrasher:
## Subject of the issue

The `glGetStringFunc` call [here](https://github.com/SFML/SFML/blob/master/src/SFML/Window/GlContext.cpp#L278) sometimes returns null. According to [the OpenGL docs](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glGetString.xhtml) this indicates an error is occurring. This has caused issues with in-development tests for `sf::Texture` and `sf::VertexBuffer`. See a failing CI run [here](https://github.com/ChrisThrasher/SFML/actions/runs/4157535248/jobs/7192096471).

Adding an assertion catches this error but doesn't get us any closer to fixing the underlying issue. It may still be worth merging this for the sake of avoiding UB.

https://github.com/ChrisThrasher/SFML/commit/bdf2f9b1e38b3efcfde004aa2b92ea3658f208a1

jeremyelalouf:
Hello @ChrisThrasher,

This issue still need some work to be done ?
I would love to contribute to SFML as my first open source contribution.

ChrisThrasher:
Not sure if there is anything else to do about this issue so I'm closing it.

PR #2404, opened 2023-02-12
Assert extension string is not null

ChrisThrasher:
## Description

This PR is related to the issue #2403 

Ideally we'd be able to recover from this error in some way but an assert is a step in the right direction to avoid UB in Debug builds.


ChrisThrasher:

Bromeon (approved):

binary1248 (approved):

eXpl0it3r (approved):

PR #2405, opened 2023-02-13
Add tests for `sf::VertexBuffer`

ChrisThrasher:
## Description

I had to skip running the tests on certain platforms because all the Windows CI runners reported that vertex buffers were unavailable. I'm not sure why this is the case.

### DIFF
--- a/test/Graphics/VertexBuffer.test.cpp
+++ b/test/Graphics/VertexBuffer.test.cpp
@@ -1,5 +1,12 @@
 #include <SFML/Graphics/VertexBuffer.hpp>
 
+// Other first party headers
+#include <SFML/Graphics/Vertex.hpp>
+
+#include <doctest/doctest.h>
+
+#include <GraphicsUtil.hpp>
+#include <array>
 #include <type_traits>
 
 static_assert(std::is_copy_constructible_v<sf::VertexBuffer>);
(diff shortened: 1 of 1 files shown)


ChrisThrasher:

ChrisThrasher:

test/Graphics/VertexBuffer.test.cpp
```diff
@@ -8,3 +14,162 @@ static_assert(std::is_move_constructible_v<sf::VertexBuffer>);
...
+    {
+        sf::VertexBuffer vertexBuffer;
+        CHECK(vertexBuffer.create(100));
+        CHECK(vertexBuffer.getVertexCount() == 100);
+    }
+
+    SUBCASE("update()")
+    {
+        sf::VertexBuffer vertexBuffer;
+        sf::Vertex       vertices[128]{};
```
eXpl0it3r, line 97: Might want to use std::array here as well

ChrisThrasher:

test/Graphics/VertexBuffer.test.cpp
```diff
@@ -8,3 +15,162 @@ static_assert(std::is_move_constructible_v<sf::VertexBuffer>);
...
+        SUBCASE("Vertices, count, and offset")
+        {
+            CHECK(vertexBuffer.create(128));
+
+            SUBCASE("Count + offset too large")
+            {
+                CHECK(!vertexBuffer.update(vertices.data(), 100, 100));
+            }
+
+            CHECK(vertexBuffer.update(vertices.data(), 128, 0));
```
eXpl0it3r, line 128: Would it make sense to test something other than 0 offset, to see that the param actually has an effect?

I guess this might need some test data that isn't just default initialized vertices.

ChrisThrasher: How would we observe the effect of an offset without drawing the buffer which is out of scope of the tests at the moment?

eXpl0it3r (approved):

PR #2409, opened 2023-02-13
Assume dinput.h is present

ChrisThrasher:
## Description

Draft PR so I can let the MinGW jobs run and see if this is still necessary.


ChrisThrasher:

ChrisThrasher:
Apparently some MinGW installations are new enough to support C++17 but old enough to not include this file

PR #2412, opened 2023-02-15
Disable unreliable tests

ChrisThrasher:
## Description

The `sf::Cursor` tests fail with the DRM backend which means keeping these would mess with any DRM users trying to develop the library. The `sf::VertexBuffer` tests fail on Linux about 10% of the time which is enough for at least one to fail every CI run which means it almost never passes on the first pass.

This commit is basically a soft revert of those two tests. However It's still worth keeping the code around even if we never run these tests.


ChrisThrasher:

eXpl0it3r (approved):

Bromeon (approved):
> However It's still worth keeping the code around even if we never run these tests.

Is the plan to run those tests locally on-demand?

PR #2410, opened 2023-02-14
Add getAspectRatio() function to VideoMode

dustdfg:
## Description

Add getAspectRatio() function to VideoMode

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android
* [ ] Add detailed description
* [ ] Add some tests

## How to test this PR?

Describe how to best test these changes. Please provide a [minimal, complete and verifiable example](https://stackoverflow.com/help/mcve) if possible, you can use the follow template as a start:

```cpp
#include <SFML/Graphics.hpp>
#include <iostream>

int main()
{
    sf::VideoMode vm = sf::VideoMode({1280, 720});
    sf::Vector2u aspectRatio = vm.getAspectRatio();

    std::cout << aspectRatio.x << " " << aspectRatio.y << std::endl;
}
```

Issue #2413, opened 2023-02-15
SFML 3 tests compile error


dustdfg:
## Subject of the issue

Compile error on tests.

## Your environment

* Debian 10
* gcc 8.3.0
* sfml git master

## Steps to reproduce

Compile sfml3 with tests

## Expected behavior

Compilation and linking

## Actual behavior

All the tests are compiling but linker has another opinion.

I tried to disable the tests one by one, the problems targets all sfml modules.  Looks like it needs c++17 because they are all counting on std::filesystem

Error message 1:

```
[ 62%] Linking CXX executable ../bin/test-sfml-system
/usr/bin/ld: CMakeFiles/test-sfml-system.dir/System/FileInputStream.test.cpp.o: in function `DOCTEST_ANON_FUNC_2()':
FileInputStream.test.cpp:(.text+0x555): undefined reference to `std::filesystem::temp_directory_path[abi:cxx11]()'
/usr/bin/ld: FileInputStream.test.cpp:(.text+0x55d): undefined reference to `std::filesystem::__cxx11::path::has_root_directory() const'
/usr/bin/ld: FileInputStream.test.cpp:(.text+0x6ed): undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
/usr/bin/ld: FileInputStream.test.cpp:(.text+0x6fc): undefined reference to `std::filesystem::__cxx11::path::has_root_directory() const'
/usr/bin/ld: FileInputStream.test.cpp:(.text+0x860): undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
/usr/bin/ld: FileInputStream.test.cpp:(.text+0xc8e): undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
/usr/bin/ld: FileInputStream.test.cpp:(.text+0x121e): undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
/usr/bin/ld: FileInputStream.test.cpp:(.text+0x1226): undefined reference to `std::filesystem::remove(std::filesystem::__cxx11::path const&)'
/usr/bin/ld: FileInputStream.test.cpp:(.text+0x12fc): undefined reference to `std::filesystem::__cxx11::path::has_filename() const'
/usr/bin/ld: FileInputStream.test.cpp:(.text+0x1386): undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
[...]

PR #2414, opened 2023-02-15
Add getAspectRatio() function to VideoMode

dustdfg:
## Description

Add getAspectRatio() function to VideoMode

Reopening of #2410
## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android
* [ ] Add detailed description
* [ ] Add some tests

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>
#include <iostream>

int main()
{
    sf::VideoMode vm = sf::VideoMode({1280, 720});
    sf::Vector2u aspectRatio = vm.getAspectRatio();

    std::cout << aspectRatio.x << " " << aspectRatio.y << std::endl;
}
```


Bromeon:
[As mentioned](https://github.com/SFML/SFML/pull/2410), could you add a test case with 5-10 resolutions in this file?

https://github.com/SFML/SFML/blob/master/test/Window/VideoMode.test.cpp

dustdfg:
> [As mentioned](https://github.com/SFML/SFML/pull/2410#pullrequestreview-1298457057), could you add a test case with 5-10 resolutions in this file?
> 
> https://github.com/SFML/SFML/blob/master/test/Window/VideoMode.test.cpp

Yes, I am working on it. I just had some troubles with compiler #2413

src/SFML/Window/VideoMode.cpp
```diff
@@ -76,6 +76,13 @@ bool VideoMode::isValid() const
     return std::find(modes.begin(), modes.end(), *this) != modes.end();
 }
 
+////////////////////////////////////////////////////////////
+Vector2u VideoMode::getAspectRatio() const
+{
+    const unsigned int divisor = std::gcd(this->size.x, this->size.y);
```
ChrisThrasher, line 82:
```suggestion
    const unsigned int divisor = std::gcd(size.x, size.y);
```
We don't use `this->`.

src/SFML/Window/VideoMode.cpp
```diff
@@ -76,6 +76,13 @@ bool VideoMode::isValid() const
...
+
+    return Vector2u(this->size.x / divisor, this->size.y / divisor);
```
ChrisThrasher, line 84:
```suggestion
    return size / divisor;
```
`sf::Vector2u` defines `Vector2u operator/(const Vector2u& left, unsigned int right);` so I believe this should be possible. Maybe we even collapse this whole function down to this:
```cpp
return size / std::gcd(size.x, size.y);
```

ChrisThrasher (changes requested):

ChrisThrasher:
> > [As mentioned](https://github.com/SFML/SFML/pull/2410#pullrequestreview-1298457057), could you add a test case with 5-10 resolutions in this file?
> > https://github.com/SFML/SFML/blob/master/test/Window/VideoMode.test.cpp
> 
> Everything isn't so rainbow as I expected:
> 
> ```diff
> +
> +        SUBCASE("Aspect ratios")
> +        {
> +            SUBCASE("4:3")
> +            {
> +                sf::Vector2u aspectRatio(4, 3);
> +                CHECK(sf::VideoMode({640, 480}).getAspectRatio() == aspectRatio);
> +                CHECK(sf::VideoMode({768, 576}).getAspectRatio() == aspectRatio);
> +                CHECK(sf::VideoMode({800, 600}).getAspectRatio() == aspectRatio);
> +                CHECK(sf::VideoMode({1024, 768}).getAspectRatio() == aspectRatio);
> +                CHECK(sf::VideoMode({1152, 864}).getAspectRatio() == aspectRatio);
> +                CHECK(sf::VideoMode({1400, 1050}).getAspectRatio() == aspectRatio);
> +                CHECK(sf::VideoMode({1600, 1200}).getAspectRatio() == aspectRatio);
> +                CHECK(sf::VideoMode({2048, 1536}).getAspectRatio() == aspectRatio);
> +            }
> +
> +            SUBCASE("16:9")
> +            {
> +                sf::Vector2u aspectRatio(16, 9);
> +                CHECK(sf::VideoMode({1024, 576}).getAspectRatio() == aspectRatio);
> +                CHECK(sf::VideoMode({1280, 720}).getAspectRatio() == aspectRatio);
> +                CHECK(sf::VideoMode({1366, 768}).getAspectRatio() == aspectRatio); //Fail
> +                CHECK(sf::VideoMode({1600, 900}).getAspectRatio() == aspectRatio);
> +                CHECK(sf::VideoMode({1920, 1280}).getAspectRatio() == aspectRatio); //Fail
> +                CHECK(sf::VideoMode({2560, 1440}).getAspectRatio() == aspectRatio);
> +            }
> +        }
> ```
> 
> You can see fails in places where the real math conflicts with people agreements... It looks like the feature doesn't have much sense 😔
> 
> @Bromeon What can you say about tests in general? What can you say about math?

You should actually push these tests so we can see precisely why the tests are failing.

ChrisThrasher:
The test for `1366, 768` is failing because that aspect ratio is only _approximately_ 16:9. It's actually very slightly off. 1366 / 758 = 1.778 whereas 16 / 9 = 1.77. 1920 / 1280 is even farther off with a ratio of 1.5.

dustdfg:
> The test for `1366, 768` is failing because that aspect ratio is only _approximately_ 16:9. It's actually very slightly off. 1366 / 758 = 1.778 whereas 16 / 9 = 1.77. 1920 / 1280 is even farther off with a ratio of 1.5.

I've just found a typo... I needed to write 1920, 1080

ChrisThrasher:
> but people agreements treat it as 16:9...

Then how do you want to reconcile the fact that what humans perceive isn't what math dictates? Does that warrant an API change? An implementation change?

dustdfg:
> > but people agreements treat it as 16:9...
> 
> Then how do you want to reconcile the fact that what humans perceive isn't what math dictates? Does that warrant an API change? An implementation change?

No. I am about fate of PR. About it is closing as unreliable. 
`proportionally changing of window size (it's weak argument)` - the only use case that I see now... It isn't so good if it has only one use case...

Bromeon:
>About 1366, 768 you are right that it's _approximately_ but people agreements treat it as 16:9...

Do they though? Just a quick search has shown both sides, many people asking why this non-16:9-resolution exists at all.

There are two behaviors possible:
* Stay true to math and return the actual fraction. This is typically not that useful in practice (and is something the user can do themselves).
* Special-case known resolutions and document this. I.e., return `(16, 9)` for `(1366, 768)`. This might make the function _actually_ useful, since it's no longer something that users can quickly do themselves.

We could also accept a parameter `bool approximate` to allow both.

---

On a separate note, this is a great example why tests are so helpful. We might not have seen this fundamental design issue until much later, possibly at a time where breaking changes are not possible anymore.

dustdfg:
> There are two behaviors possible:
> 
>     * Stay true to math and return the actual fraction. This is typically not that useful in practice (and is something the user can do themselves).
> 
>     * Special-case known resolutions and document this. I.e., return `(16, 9)` for `(1366, 768)`. This might make the function _actually_ useful, since it's no longer something that users can quickly do themselves.
> 
> 
> We could also accept a parameter `bool approximate` to allow both.
> 

In my opinion, it is look a bit messy to use bool. It looks like a look up table that uses an almost useless function :/

If it is so necessary, shouldn't it just be a `constexpr` look up table for common resolutions in a separate namespace? Without any `getAspectRatio` functions...

As internet says the world mobile devices have a more complicated zoo of display resolutions (for which I even don't know statistic and common examples) and aspect ratios *(which aren't good represent with integers) `18:9`, `18.5:9`, `19:9`, `19.5:9`, `20:9`, `21:9` 

So both ideas of special cases or look up table isn't so good

*(`19.5:9` is `39:18`)

Bromeon:
Hm... maybe you're right and it's not SFML's job to do that. Or if it is, we should at least have a clearer idea about the use case and thus work out a design accordingly 🤔

src/SFML/Window/VideoMode.cpp
```diff
@@ -76,6 +76,11 @@ bool VideoMode::isValid() const
     return std::find(modes.begin(), modes.end(), *this) != modes.end();
 }
 
+////////////////////////////////////////////////////////////
+Vector2u VideoMode::getAspectRatio() const
+{
+    return size / std::gcd(size.x, size.y);
```
kimci86, line 82: This should handle the case where gcd is 0. (That happens with a default-constructed VideoMode instance.)

dustdfg: You are right

MarioLiebisch:
Don't think a lookup table would be needed here. Keep in mind this is for `VideoMode` after all. The user can define any dimensions, it's not just what the system reports for an active screen. So if the result is 22:9 or whatever, that should be reported.

However, there could be a second variation (or changed by parameter?) that tries to report the *closest* common aspect ratio, e.g. a 21.5:9 would result in returning 21:9.

eXpl0it3r:
I personally wouldn't add this to SFML, given that there can be different options on how to handle it. It's not a fundamental feature that users couldn't do without SFML providing additional APIs, it's not *that* complicated to calculate some ratio on your own, and it leaves the decision on how to deal with odd ratios to each developer, rather than us trying to implement an overcomplicated API in an attempt to satisfy everyone's preference.

eXpl0it3r:
Thank you for the suggestion! With the reasons given in my previous comment, I'll close this PR.

I think the code could be extracted and wrapped up in a small extension that could find a good place on the wiki: https://github.com/SFML/SFML/wiki/Sources

PR #2415, opened 2023-02-15
Add `misc-const-correctness` clang-tidy check

ChrisThrasher:
## Description

I wanted to submit this to show how much more we could be using `const`. While the extra `const` is not useful in some places, I think this check is still worthwhile since it will ensure that all future PRs at least attempt to use `const` where appropriate. I know how hard it is to manually audit code for the right use of `const`, so I'd rather err on the side of too much const than too little. 

One quirk about this is that in Objective-C++ files, clang-tidy recommended we add `const` to pointers (as opposed to the pointed-to type) which is not something it recommended elsewhere so I just added that `const` to satisfy it even though we don't do that in C++ files. There's an option to enable this on all pointers but that may be excessive


ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

Bromeon:
"Everything mutable" is one of the "wrong defaults" in C++. I'm not sure how much we gain by trying hard to fix this everywhere. It seems to me that a very small number of bugs have actually resulted from that -- and maybe we'd be better off using `const` selectively in critical cases.

On the other hand, there's a very high price for this:
* Readability.
* Signaling effect: cluttering everything with `const` waters down the keyword and makes us less aware of _important_ cases like `const Ptr*`, where the `const` has deeper semantic meaning.
* Inconsistency: we don't do it for pointer types and for parameters, only for local non-pointer variables. (I'm not suggesting we should do it for parameters too, that would make things even less readable).

While well-intended, I'm not sure about the cost-benefit-ratio of this change. How do others see this?

ChrisThrasher:
You made the same argument against the safety gains of `[[nodiscard]]`. What I find so confusing is why you say "the defaults are wrong" then "we should keep using these bad default". We have automated tools to fix these endemic language problems yet you actively prefer the opposite. I can't make sense of that.

ChrisThrasher:
> The "minuscule cost" is where I disagree. I mentioned three concrete disadvantages, one of them being "we can't even fix it properly".

Your "readability" point is not obvious to me. I think code is much easier to read when `const` is used more often. When I see a block of non-const local variables, I have to slow down reading to take careful note of where those variables get modified instead of simply checking for a valid initial value and moving along.

I don't think the signaling effect of `const` or `[[nodiscard]]` is as impactful as you say. Using these opt-in safety features means the code is correct by compilation so the "signal" we care about is when these things are absent. In the case of an absent `const`, that tells me nothing. Its absence gives me nearly no information. Maybe the variable gets modified. Maybe it doesn't. I can't know that without reading all code that touches it. However, if we start programmatically enforcing `const` and `[[nodiscard]]` then the absence of such things starts to have a stronger signal. If we use `const` as much as possible then a missing `const` almost certainly means "this gets modified".

The consistency argument can be address if you really care about it. clang-tidy has options for adding const to pointers as well but it's disabled by default since it does create rather verbose code. I think it's fine to not const the pointer itself as long as the pointer is const where possible. As for parameters, clang-tidy won't enforce that for us but I've said before that I wish we used `const` on parameters in their definitions so if you let me have my way it would be more consistent than this PR.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:
Does the team have any more thoughts on this? It's a huge, automatic, cheap win for safety and correctness.

Bromeon:
For me still kind of the same as [in my last comment](https://github.com/SFML/SFML/pull/2415#issuecomment-1438371033). 
I doubt the "huge win" in terms of "actual bugs prevented".

Would be good to hear other team members' view on this! E.g. @vittorioromeo @eXpl0it3r @binary1248?

eXpl0it3r:
I'm not sure how much this can/should be automated. Personally, I do prefer to use `const` for local variables, however it shouldn't just indicated "the value is never changed", but it should indicate "the value *should* never be changed". The `const` should communicate the intent of the value being constant and not just the fact, that we so far haven't written code that changes it.

On the other hand, it doesn't necessarily hurt to mark stuff which is just set once as `const`, and on the plus side might highlight potential mistakes, when you intended to change something later on, but forgot to do so.

Having said that, I find kind of hard to accept marking lock guards or `sf::Clock` as `const`. I understand the technical reasons, I just find it really confusing, but might just be my mental model having accepted those classes as un-reassignable 😄

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:
Here's a middle ground solution:

1. Add `const` on most local variables where it applies
2. Don't add `const` to locks and other select places it's not wanted
3. Keep `misc-const-correctness` disabled

This gives us the readability and safety benefits of `const` when objects aren't being mutated, but we're aren't forced to add it in select places where we don't want it. So long as we keep `misc-const-correctness` disabled then we can be selective. clang-tidy is a useful tool for finding where `const` _could_ be added but we're not forced to enable that check if we just want to use clang-tidy as a one-time refactoring tool.

This would set a precedent that we prefer `const` on local variables but aren't dogmatic about it.

If we can't come to an agreement on such a middle ground then we're setting a precedent that we don't put much value on `const` local vars and ought to close to PR.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

eXpl0it3r (approved):

Bromeon:
Honestly, this will likely lead to _more_ bikeshedding because now the question is always "should this be const or not".

I'm not too happy with this change being merged without consensus (and not sure about @vittorioromeo who 👍'ed my original comment), but I also don't think it's worth spending months and months of discussions on.

PR #2417, opened 2023-02-16
Allow setting window icon with `sf::Image`

metaquarx:
## Description

Currently, setting window icon is only allowed using a size & pixel pointer.

This overload wraps that, allowing users to set the window icon using an `sf::Image` directly.
`sf::Image` has the same 32bit RGBA pixel format as required by `setIcon`.
This is only implemented for `sf::RenderWindow`, as `sf::Image` is part of the graphics module.

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/include/SFML/Graphics/RenderWindow.hpp
+++ b/include/SFML/Graphics/RenderWindow.hpp
@@ -35,6 +35,8 @@
 
 namespace sf
 {
+class Image;
+
 ////////////////////////////////////////////////////////////
 /// \brief Window that can serve as a target for 2D drawing
 ///
@@ -112,6 +114,18 @@ public:
     ////////////////////////////////////////////////////////////
     Vector2u getSize() const override;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Change the window's icon
+    ///
+    /// The OS default icon is used by default.
+    ///
+    /// \param icon Image to use as the icon. The image is copied,
+    ///             so you need not keep the source alive after
+    ///             calling this function.
+    ///
+    ////////////////////////////////////////////////////////////
+    void setIcon(const Image& icon);
+    using Window::setIcon;
 
     ////////////////////////////////////////////////////////////
     /// \brief Tell if the window will use sRGB encoding when drawing on it
--- a/src/SFML/Graphics/RenderWindow.cpp
+++ b/src/SFML/Graphics/RenderWindow.cpp
@@ -26,6 +26,7 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Graphics/GLCheck.hpp>
+#include <SFML/Graphics/Image.hpp>
 #include <SFML/Graphics/RenderTextureImplFBO.hpp>
 #include <SFML/Graphics/RenderWindow.hpp>
 #include <SFML/Graphics/Texture.hpp>
@@ -62,6 +63,13 @@ RenderWindow::~RenderWindow() = default;
 Vector2u RenderWindow::getSize() const
 {
     return Window::getSize();
+}
+
+
+////////////////////////////////////////////////////////////
+void RenderWindow::setIcon(const Image& icon)
+{
+    setIcon(icon.getSize(), icon.getPixelsPtr());
 }
 
 


include/SFML/Graphics/RenderWindow.hpp
```diff
@@ -112,6 +114,19 @@ class SFML_GRAPHICS_API RenderWindow : public Window, public RenderTarget
...
+    ///
+    /// The OS default icon is used by default.
+    ///
+    /// \param icon Image to use as the icon. The image is copied,
+    ///             so you need not keep the source alive after
+    ///             calling this function.
+    ///
+    ////////////////////////////////////////////////////////////
+    void setIcon(const sf::Image& icon);
+    using Window::setIcon;
```
ChrisThrasher, line 128: What does this `using` alias do?

include/SFML/Graphics/RenderWindow.hpp
```diff
@@ -112,6 +114,19 @@ class SFML_GRAPHICS_API RenderWindow : public Window, public RenderTarget
...
+    /// \brief Change the window's icon
+    ///
+    /// The OS default icon is used by default.
+    ///
+    /// \param icon Image to use as the icon. The image is copied,
+    ///             so you need not keep the source alive after
+    ///             calling this function.
+    ///
+    ////////////////////////////////////////////////////////////
+    void setIcon(const sf::Image& icon);
```
ChrisThrasher, line 127:
```suggestion
    void setIcon(const Image& icon);
```

src/SFML/Graphics/RenderWindow.cpp
```diff
@@ -65,6 +66,13 @@ Vector2u RenderWindow::getSize() const
 }
 
 
+////////////////////////////////////////////////////////////
+void RenderWindow::setIcon(const sf::Image& icon)
```
ChrisThrasher, line 70:
```suggestion
void RenderWindow::setIcon(const Image& icon)
```

ChrisThrasher (changes requested):
You can drop the `sf::` namespace in a few places

include/SFML/Graphics/RenderWindow.hpp
metaquarx: By default, an overload in a derived class overshadows a function of the same name in a base class (name hiding). This just imports it back into the current scope, so that both versions can be used (keeping the api a bit more consistent between an `sf::Window` and an `sf::RenderWindow`)

ChrisThrasher: https://godbolt.org/z/7zYeMeGjx

Interesting. I did know that.

include/SFML/Graphics/RenderWindow.hpp
```diff
@@ -112,6 +114,19 @@ class SFML_GRAPHICS_API RenderWindow : public Window, public RenderTarget
...
+    /// The OS default icon is used by default.
+    ///
+    /// \param icon Image to use as the icon. The image is copied,
+    ///             so you need not keep the source alive after
+    ///             calling this function.
+    ///
+    ////////////////////////////////////////////////////////////
+    void setIcon(const Image& icon);
+    using Window::setIcon;
+
```
ChrisThrasher, line 129:
```suggestion
```

ChrisThrasher:
Is there a particular use case that motivated you to add this?

metaquarx:
It's a somewhat common point of confusion, both on Discord & the forums. The advice is always to load it as an image, and pass in the size and pointer (as the PR does). There seems to be no particular reason for it not to be included in the first place.

I would assume the majority of users would be setting the icon using a loaded `sf::Image`, rather than using something like an embedded binary resource, due to its simplicity.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:
I'm open to this idea. I'm generally a fan of these little helpers that add convenience and make it easier to do functional programming where function return values can be directly passed to other functions. This change is in the same vein as the PRs that replaced `(float, float)` APIs with `Vector2f` which have aged well.

src/SFML/Graphics/RenderWindow.cpp
```diff
@@ -65,6 +66,13 @@ Vector2u RenderWindow::getSize() const
 }
 
 
+////////////////////////////////////////////////////////////
+void RenderWindow::setIcon(const Image& icon)
+{
+    Window::setIcon(icon.getSize(), icon.getPixelsPtr());
```
ChrisThrasher, line 72:
```suggestion
    setIcon(icon.getSize(), icon.getPixelsPtr());
```

ChrisThrasher:

eXpl0it3r:

ChrisThrasher (approved):

Bambo-Borris:
Love this as a convenience feature for users, I think basically all of my projects with SFML would have used this. It may be a small change, but it's a QoL one for sure.

Bromeon (approved):
Great addition!

eXpl0it3r (approved):

PR #2418, opened 2023-02-16
Test install interface with frameworks

ChrisThrasher:
## Description

Fixes another bit of tech debt added in #2206. Because SFML defaults to shared libs but the install test uses CMake's default of static libs, it was trying to find the static build of the frameworks which doesn't exist causing it to fail. When writing #2206 I was too lazy to fix this so I just disabled the install tests when using frameworks.


ChrisThrasher:

eXpl0it3r (approved):

Issue #2395, opened 2023-02-06
Android OpenGL context 0.0 and antialiasing 


Zombieschannel:
## Subject of the issue

When creating a window on Android, you get the created version 0.0 error. The app does not crash and works fully, there is just this annoying error + "Failed to activate the window's context" spam. If you also add antialiasing (which isn't supported on OpenGL ES afaik), the app also does not crash, you just get a black screen (common OpenGL stuff). Also for some reason when you set antialiasing to 1, it will say that it created version 0.0 and antialiasing x2.

![image](https://user-images.githubusercontent.com/54750550/217073589-682a770c-a6ac-4d63-ada0-f0df40dad6fc.png)

## Your environment

Android Studio 2022.1.1 on Windows 10
SFML 2.6.x branch
Used CMake, Ninja, Android NDK 25

## Steps to reproduce

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode::getDesktopMode(), "App", sf::Style::Fullscreen, sf::ContextSettings(0, 0, 2));
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        for (sf::Event event; window.pollEvent(event);)
        {
            if (event.type == sf::Event::Closed)
                window.close();
        }

        window.clear();
        window.display();
    }
}
```

## Expected behavior

It should create a valid OpenGL context.

## Actual behavior

Logs show an error +"Failed to activate the window's context" spam. With antialiasing: black screen.

PR #2422, opened 2023-02-19
Fixed Android OpenGL context 0.0 and antialiasing

Zombieschannel:

----

## Description

Fixes the OpenGL context 0.0 bug and fixes the black screen and egl surface crash when antialiasing is > 4.

This PR is related to the issue https://github.com/SFML/SFML/issues/2395

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [X] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>
int main()
{
    sf::RenderWindow window;
    //change antialiasing and major minor version of sf::ContextSettings to test
    window.create(sf::VideoMode::getDesktopMode(), "SFML works!", sf::Style::Fullscreen, sf::ContextSettings(0, 0, 8));
    window.setFramerateLimit(3);
    sf::CircleShape shape(400.f);
    while (window.isOpen())
    {
        sf::Event event;
        {
                window.close();
            if (event.type == sf::Event::LostFocus)
                window.setActive(0);
            if (event.type == sf::Event::GainedFocus)
                window.setActive(1);
                    window.close();
        }
        if (!window.hasFocus())
            continue;
        window.clear();
        window.draw(shape);
        window.display();

    }
}
```

### DIFF
--- a/src/SFML/Window/GlContext.cpp
+++ b/src/SFML/Window/GlContext.cpp
@@ -293,6 +293,8 @@ namespace
             }
         }
 
+#ifndef SFML_SYSTEM_ANDROID
+
         // Helper to parse OpenGL version strings
         bool parseVersionString(const char* version, const char* prefix, unsigned int &major, unsigned int &minor)
         {
@@ -312,6 +314,9 @@ namespace
 
             return false;
         }
+
+#endif
+
     }
 }
 
@@ -756,10 +761,6 @@ void GlContext::initialize(const ContextSettings& requestedSettings)
     // Activate the context
     setActive(true);
 
-    // Retrieve the context version number
-    int majorVersion = 0;
-    int minorVersion = 0;
-
     // Try the new way first
     glGetIntegervFuncType glGetIntegervFunc = reinterpret_cast<glGetIntegervFuncType>(getFunction("glGetIntegerv"));
     glGetErrorFuncType glGetErrorFunc = reinterpret_cast<glGetErrorFuncType>(getFunction("glGetError"));
@@ -772,6 +773,11 @@ void GlContext::initialize(const ContextSettings& requestedSettings)
         err() << "Could not load necessary function to initialize OpenGL context" << std::endl;
         return;
     }
+#ifndef SFML_SYSTEM_ANDROID
+
+    // Retrieve the context version number
+    int majorVersion = 0;
+    int minorVersion = 0;
 
     glGetIntegervFunc(GL_MAJOR_VERSION, &majorVersion);
     glGetIntegervFunc(GL_MINOR_VERSION, &minorVersion);
@@ -812,6 +818,14 @@ void GlContext::initialize(const ContextSettings& requestedSettings)
             err() << "Unable to retrieve OpenGL version string, defaulting to 1.1" << std::endl;
         }
     }
+
+#else
+
+    // Android contexts are always defaulting to 1.1 since SFML is still using GLES 1
+    m_settings.majorVersion = 1;
+    m_settings.minorVersion = 1;
+
+#endif
 
     // 3.0 contexts only deprecate features, but do not remove them yet
     // 3.1 contexts remove features if ARB_compatibility is not present
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,3 @@ build/
 .cache/
 .vscode/
 .idea/
-/.vs
-/out/install
-/CMakeSettings.json
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,6 @@ build/
 .cache/
 .vscode/
 .idea/
+/.vs
+/out/install
+/CMakeSettings.json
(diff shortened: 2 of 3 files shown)


ChrisThrasher:

src/SFML/Window/GlContext.cpp
```diff
@@ -772,6 +773,11 @@ void GlContext::initialize(const ContextSettings& requestedSettings)
         err() << "Could not load necessary function to initialize OpenGL context" << std::endl;
         return;
     }
+#ifndef SFML_SYSTEM_ANDROID
```
ChrisThrasher, line 776:
```suggestion

#ifndef SFML_SYSTEM_ANDROID
```

ChrisThrasher (changes requested):

binary1248:
What is the actual problem that this change is supposed to fix? Based on #2395 it seems like the `GL_MAJOR_VERSION` and `GL_MINOR_VERSION` queries don't even fail, so 0.0 is being reported by OpenGL ES itself and isn't a result of wrong handling within SFML. If this is the case, does this happen on all platforms? I would be wary building in a workaround for a platform bug that might even be well documented somewhere.

The same goes for the anti-aliasing level. This change clamps its value to 4. Why did the selection fail before? `getBestConfig` only returned configurations that are reported by the operating system, so why is it that it could still fail?

We should try to understand the underlying problems before replacing large blocks of code with magic numbers.

Zombieschannel:
> Why did the selection fail before?

In EglContext.cpp in getBestConfig there are these lines:
```cpp
const EGLint attributes[] = {
        EGL_BUFFER_SIZE, static_cast<EGLint>(bitsPerPixel),
        EGL_DEPTH_SIZE, static_cast<EGLint>(settings.depthBits),
        EGL_STENCIL_SIZE, static_cast<EGLint>(settings.stencilBits),
        EGL_SAMPLE_BUFFERS, settings.antialiasingLevel ? 1 : 0,
        EGL_SAMPLES, static_cast<EGLint>(settings.antialiasingLevel), //<---- this line specifically
        EGL_SURFACE_TYPE, EGL_WINDOW_BIT | EGL_PBUFFER_BIT,
        EGL_RENDERABLE_TYPE, EGL_OPENGL_ES_BIT,
        EGL_NONE
    };
```
There are no checks whether the user set the antialiasing level correctly. EGL_SAMPLES can be 4 at max. Otherwise you get a black screen and many EGL_BAD_CONFIG errors:

image

> it seems like the GL_MAJOR_VERSION and GL_MINOR_VERSION queries don't even fail

not sure why does it not fail since `GL_MAJOR_VERSION` and `GL_MINOR_VERSION` don't seem to appear in OpenGL ES until version 3

binary1248:
Then the problem is that the calls to `glGetIntegervFunc` don't fail properly as the code assumes. As you can see in the following code block, we already have code that is specifically written to parse the version out of the `GL_VERSION` string that can be queried in OpenGL ES 1.x. We just have to make sure we get there. I would prefer we add an additional check that `majorVersion != 0` (which we can assume should never be a valid version). If it is 0 then we try to parse `GL_VERSION`. This should cover even such broken implementations that silently fail when querying for unsupported enums.

I also realize now that the EGL context creation still relies on the platform provided pixel format selection. When creating other platform context types, we consciously made the decision not to rely on the built-in functionality because it was both too strict and also not flexible enough for our needs. The format selection process for all other platform context types consists of first collecting all formats supported by the platform and evaluating their fit against what the user requested via the API (using `GlContext::evaluateFormat`). This avoids the catastrophic failure that you experienced since surface creation is basically guaranteed, and warns the user if they requested something that isn't exactly supported on the platform they are running on. It would be nice if EGL surface format selection could follow the other context types. We would have a uniform process (that would be easier to document), selection behaviour that doesn't differ based on the platform and a solution for the issue at hand.

Zombieschannel:
By adding `majorVersion != 0`, it also seems like something is broken with glGetStringFunc(GL_VERSION), the function does not fail but it does not return anything, therefore we get the message `Unable to retrieve OpenGL version string, defaulting to 1.1`. It fails like this on any Android device I try, even though it is mentioned [here](https://registry.khronos.org/OpenGL-Refpages/es1.1/xhtml/glGetString.xml) that GL_VERSION is a valid parameter.

binary1248:
Would it be possible to run a barebones Java example on those platforms to see what the aforementioned functions return via the Java API? It could be that this is a NDK-specific bug.

https://developer.android.com/develop/ui/views/graphics/opengl/about-opengl#java

Zombieschannel:
Not sure, unfortunately, I'm not very familiar with Java.

MarioLiebisch:
> Would it be possible to run a barebones Java example on those platforms to see what the aforementioned functions return via the Java API?

Wouldn't an [app like this](https://play.google.com/store/apps/details?id=de.saschawillems.glescapsviewer) be enough? [Getting this on my Pixel 7 Pro.](http://opengles.gpuinfo.org/displayreport.php?id=6245)

binary1248:
That app doesn't explicitly request a legacy context. Could be that the platform developers have just started to neglect supporting legacy OpenGL because they assume everybody moved on.

MarioLiebisch:
Ah, true.

PR #2425, opened 2023-02-20
Remove unused headers and reduce physical dependencies

vittorioromeo:
More header cleanup, similar to #2424, in order to improve compilation times.


src/SFML/Window/DRM/ClipboardImpl.hpp
```diff
@@ -24,11 +24,10 @@
...
-////////////////////////////////////////////////////////////
-#include <SFML/System/String.hpp>
-
+namespace sf
+{
+class String;
+}
 
 namespace sf::priv
 {
```
ChrisThrasher, line 33: Do we prefer this over only opening up `namespace sf` once? There are plenty of files where the only reason we're not using a nested namespace for `sf::priv::` is because we have some forward declarations in the `sf::` namespace.

ChrisThrasher (commented):
Almost ready to approve. Glad to remove more unnecessary `#include`s.

include/SFML/Graphics/Image.hpp
```diff
@@ -156,7 +156,7 @@ class SFML_GRAPHICS_API Image
     /// \see create, loadFromFile, loadFromMemory, saveToFile
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool saveToMemory(std::vector<std::uint8_t>& output, const std::string& format) const;
```
kimci86, line 159: Should we pass `string_view` by value instead of reference?

Bromeon (commented):
Great work!

I agree with the comments regarding `std::string_view` copy and unclarity of `sf { priv {` vs `sf::priv {` -- other than that, good from my side 🙂

include/SFML/Graphics/Image.hpp
vittorioromeo: Yes, will change.

src/SFML/Window/DRM/ClipboardImpl.hpp
vittorioromeo: Will make it consistent with existing practice.

ChrisThrasher:

ChrisThrasher (approved):

PR #2430, opened 2023-02-28
Remove code coverage from build scripts

### LINKED ISSUE #2429: Remove `runtests` target
## Subject of the issue

The `runtests` target was added in #1475 without much fanfare. I tried to find any discussion about its rationale but didn't see any so I'm not sure what its proponents would say about it.

I propose we remove it for the following reasons:
1. It increases build times. No matter what, each build must run the tests. That means a no-op incremental build actually takes a few seconds since it has to run the tests even though no code has changed. As we add more tests, this build time penalty gets bigger and bigger. 
2. It's non-standard. I'm not aware any CMake project that does this. It's a bit surprising to have the tests get automatically run when all you did was ask CMake to build the project.
3. It does not produce color output. Running `ctest` will give you nicely colored output but the `runtests` target due to it using `add_custom_command` only produces a monochromatic output making it a bit harder to parse.
4. It's additional code to maintain. It's not particularly difficult to maintain, but it's a non-zero cost nonetheless.

Because our use of OpenCppCoverage on Windows ties into `runtests`, some work will be required to extract that code from the build scripts into the CI pipeline but I believe this is a solvable problem.

ChrisThrasher:
## Description

Closes #2429 

The first commit is the part that closes #2429. The second commit removes the `SFML_ENABLE_COVERAGE` option entirely because after the first commit, there was little remaining need to keep that option. Coverage can be enabled from outside of the build so there's no particular reason that needs to have its own designated option in the build. It's not as if any developers are routinely checking coverage statistics locally so no convenience is being lost.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
         - { name: Windows VS2022 ClangCL, os: windows-2022, flags: -T ClangCL }
         - { name: Windows VS2022 Clang,   os: windows-2022, flags: -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -GNinja }
         - { name: Linux GCC,            os: ubuntu-latest, prefix: xvfb-run -a, flags: -DSFML_RUN_DISPLAY_TESTS=ON -GNinja }
-        - { name: Linux Clang,          os: ubuntu-latest, prefix: xvfb-run -a, flags: -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DSFML_RUN_DISPLAY_TESTS=ON -GNinja , gcovr_options: '--gcov-executable="llvm-cov-$CLANG_VERSION gcov"' }
+        - { name: Linux Clang,          os: ubuntu-latest, prefix: xvfb-run -a, flags: -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DSFML_RUN_DISPLAY_TESTS=ON -GNinja }
         - { name: MacOS,                os: macos-11, flags: -GNinja }
         - { name: MacOS Xcode,          os: macos-11, flags: -GXcode }
         config:
@@ -26,7 +26,7 @@ jobs:
         - { name: Static, flags: -DBUILD_SHARED_LIBS=FALSE }
         type:
         - { name: Release }
-        - { name: Debug, flags: -DCMAKE_BUILD_TYPE=Debug -DSFML_ENABLE_COVERAGE=TRUE }
+        - { name: Debug, flags: -DCMAKE_BUILD_TYPE=Debug }
 
         include:
         - platform: { name: Windows VS2022, os: windows-2022 }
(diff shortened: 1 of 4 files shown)


ChrisThrasher:

ChrisThrasher:
The new code coverage figure is down from before and I might know why. Previously we didn't do debug builds of the DRM backend which meant we never ran coverage on all the DRM-specific code. The new coverage job does include the DRM code. Because most of that code is untested, I think it's bringing down the percentage although that's good because it means the new coverage figure is more accurate.

kimci86:
I don't think DRM-specific code alone gives the explanation.
Looking at the details, we see less files and less lines are included in the coverage report.
```
@@            Coverage Diff             @@
##           master    #2430      +/-   ##
==========================================
- Coverage   23.18%   18.15%   -5.03%     
==========================================
  Files         212      145      -67     
  Lines       18063    12167    -5896  
```

ChrisThrasher:

ChrisThrasher:
https://app.codecov.io/gh/SFML/SFML/commit/b288af2cd4e303cdc674cf13015d21dbfc7ea9ee

The Windows coverage report contains no coverage information which is probably the real reason coverage stats are so messed up. I need to figure out what I'm doing wrong with OpenCppCoverage.

ChrisThrasher:
I'm tempted to remove OpenCppCoverage if I can't get it to work and just use Clang + llvm-cov to do coverage on all platforms. If we use the same coverage tool everywhere then we can stop maintaining as much platform-specific code. I'd love to get rid of another platform-specific tool.

ChrisThrasher:
Codecov is no longer reporting any errors in the coverage reports yet we're still seeing that -5% coverage figure.

ChrisThrasher:

ChrisThrasher:
The first commit (https://github.com/SFML/SFML/pull/2430/commits/af4e95316f5ce6d9372d6ddd23c5a4f669763130) which does the bulk of the work refactoring code coverage only [reduces coverage by 1.93%](https://app.codecov.io/gh/SFML/SFML/commit/af4e95316f5ce6d9372d6ddd23c5a4f669763130). It seems like I've fixed some of the issues by fixing that empty Windows coverage report.

What's also weird is how the second commit (https://github.com/SFML/SFML/pull/2430/commits/7fba87ae8714b1f359187c700e41bc8f3c740c34) which just removes a CMake option [reduces coverage by another 1.42%](https://app.codecov.io/github/SFML/SFML/commit/7fba87ae8714b1f359187c700e41bc8f3c740c34).

I'm not sure how to improve this any more. Either Codecov is buggy or I'm forgetting to run coverage on a codepath that was previously being covered.

ChrisThrasher:

ChrisThrasher:
Screenshot 2023-03-02 at 4 08 43 PM

I don't know why but adding a lot more coverage jobs in a89bca5 seemingly fixed the coverage gap.

ChrisThrasher:

ChrisThrasher:
I now realize how the intent of this PR could largely be addressed by merely removing the `ALL` form the `runtests` target and I'm open to a PR that does just that (and adds a new testing CI job to ensure tests still get run). Removing coverage from the build scripts is a separate matter that can come later if we want.

ChrisThrasher:
I still want to do this but I'd rather wait until I can come up with an LLVM-based solution that will work the same on all platforms like clang-tidy does.

PR #2433, opened 2023-03-04
Replace `std::rand` with `<random>`

ChrisThrasher:
## Description

https://sonarcloud.io/project/issues?resolved=false&severities=BLOCKER&id=SFML_SFML&open=AYaqzWMqgNPC4qRKFLQx

Addresses some static analysis warnings about our use of `std::rand`.

### DIFF
--- a/src/SFML/Audio/SoundFileWriterOgg.cpp
+++ b/src/SFML/Audio/SoundFileWriterOgg.cpp
@@ -32,8 +32,8 @@
 #include <algorithm>
 #include <cassert>
 #include <cctype>
-#include <cstdlib>
 #include <ostream>
+#include <random>
 
 
 namespace sf::priv
@@ -63,7 +63,8 @@ bool SoundFileWriterOgg::open(const std::filesystem::path& filename, unsigned in
     m_channelCount = channelCount;
 
     // Initialize the ogg/vorbis stream
-    ogg_stream_init(&m_ogg, std::rand());
+    static std::mt19937 rng(std::random_device{}());
+    ogg_stream_init(&m_ogg, std::uniform_int_distribution(0, std::numeric_limits<int>::max())(rng));
     vorbis_info_init(&m_vorbis);
 
     // Setup the encoder: VBR, automatic bitrate management
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -9,7 +9,6 @@ Checks: >
   -clang-analyzer-optin.cplusplus.VirtualCall,
   -clang-analyzer-optin.osx.*,
   -clang-analyzer-osx.*,
-  -clang-analyzer-security.insecureAPI.rand,
   -clang-analyzer-unix.Malloc,
   -misc-const-correctness,
   -misc-misplaced-const,
--- a/examples/shader/Shader.cpp
+++ b/examples/shader/Shader.cpp
@@ -279,8 +279,9 @@ public:
         {
             // Spread the coordinates from -480 to +480
             // So they'll always fill the viewport at 800x600
-            m_pointCloud[i].position.x = static_cast<float>(rand() % 960) - 480.f;
-            m_pointCloud[i].position.y = static_cast<float>(rand() % 960) - 480.f;
+            std::uniform_real_distribution<float> positionDistribution(-480, 480);
+            m_pointCloud[i].position.x = positionDistribution(rng);
+            m_pointCloud[i].position.y = positionDistribution(rng);
         }
 
         // Load the texture


ChrisThrasher:

ChrisThrasher:

src/SFML/Audio/SoundFileWriterOgg.cpp
```diff
@@ -32,8 +32,15 @@
...
-#include <cstdlib>
 #include <ostream>
+#include <random>
+
+
+namespace
+{
+std::random_device rd;
+std::mt19937       rng(rd());
+} // namespace
```
Bromeon, line 43: Is this save to initialize at global startup, or could these types depend on global objects initialized somewhere else, potentially at a later stage?

Also a more general question, do we want to advocate having RNGs as global variables? Examples are also used to highlight best practices from our point of view.

Bromeon (commented):
Thanks for addressing this! My know-how regarding global RNGs in C++ is not up-to-date, so a few comments regarding that 🙂

src/SFML/Audio/SoundFileWriterOgg.cpp
ChrisThrasher: > Is this save to initialize at global startup, or could these types depend on global objects initialized somewhere else, potentially at a later stage?

We use this pattern elsewhere in SFML and I've used it in many other projects with no issue so I believe it's perfectly safe.

> Also a more general question, do we want to advocate having RNGs as global variables? Examples are also used to highlight best practices from our point of view.

The alternative is to use a `static` local variable which I'm fine doing in this case if you prefer.

ChrisThrasher:
If we really want to set a good example for randomness we can add a new `sf::rng()` function that returns a `thread_local` random number generator that is properly seeded. Here's what it takes to seed the entirety of the state of a `std::mt19937`:

https://github.com/ChrisThrasher/boids/blob/master/src/Boid.cpp#L8-L14

It's not trivial nor easy to learn. It's safe to expect that most random number generators are not seeded as well as they could be. This would obviously replace all existing `mt19937`s in SFML.

Why `thread_local`? Because there's no good reason to want to use the same random number generator across threads. Doing so is prone to produce race conditions if you don't synchronize and prone to slow things down if you do synchronize access to it. Might as well make a new one per thread to avoid that problem.

Bromeon:
Thanks a lot! Maybe in that case we can leave the global variable here in the example.

Thread-local RNGs inside SFML is something we can think about for code reuse. I'm not sure if it should be part of the public API though -- if SFML and user code shares the same RNG, it will be very hard to achieve deterministic results, and the behavior may change between SFML versions. That is, unless SFML strictly separates publicly-exposed and internally-used ones.

I also agree that it's a pity that C++ doesn't offer higher-level APIs for random number generation. Sure, the low-level one is very flexible, and it's not that difficult to get _something_ up and running, but as your example shows, having a good sequence is hard to achieve. Then again, C and C++ games have used `rand()` for decades, and it was good enough. Scientific domains might be a different story.

ChrisThrasher:
https://github.com/SFML/SFML/issues/2434

I made a separate issue to continue to `sf::rng` discussion.

src/SFML/Audio/SoundFileWriterOgg.cpp
vittorioromeo:
```suggestion
std::mt19937& rng() 
{
    static std::mt19937 result{std::random_device{}()};    
    return result;
}
```

Or maybe `thread_local`? Do we care about thread-safety here?

src/SFML/Audio/SoundFileWriterOgg.cpp
```diff
@@ -63,7 +70,8 @@ bool SoundFileWriterOgg::open(const std::filesystem::path& filename, unsigned in
     m_channelCount = channelCount;
 
     // Initialize the ogg/vorbis stream
-    ogg_stream_init(&m_ogg, std::rand());
+    std::uniform_int_distribution serialNumberDistribution(0, RAND_MAX);
+    ogg_stream_init(&m_ogg, serialNumberDistribution(rng));
```
vittorioromeo, line 74:
```suggestion
    ogg_stream_init(&m_ogg, std::uniform_int_distribution{0, RAND_MAX}(rng()));
```

vittorioromeo (changes requested):

src/SFML/Audio/SoundFileWriterOgg.cpp
ChrisThrasher: What's the advantage of hiding the mt19937 within a function?

ChrisThrasher:
I inlined a few unnecessary local variables and moved the `mt19937` into the one function that needed it since there was no reason it needed to be global.

ChrisThrasher:

src/SFML/Audio/SoundFileWriterOgg.cpp
```diff
@@ -63,7 +63,8 @@ bool SoundFileWriterOgg::open(const std::filesystem::path& filename, unsigned in
     m_channelCount = channelCount;
 
     // Initialize the ogg/vorbis stream
-    ogg_stream_init(&m_ogg, std::rand());
+    static std::mt19937 rng(std::random_device{}());
+    ogg_stream_init(&m_ogg, std::uniform_int_distribution(0, RAND_MAX)(rng));
```
vittorioromeo, line 67: I would use `{}` here as it makes it more obvious that the first set of parentheses is initialization and the second is a function call.

vittorioromeo (approved):

src/SFML/Audio/SoundFileWriterOgg.cpp
ChrisThrasher: I didn't do that because it's not precedent in SFML

eXpl0it3r, line 67: Feels a bit odd to use `RAND_MAX` here, which is essentially more connected to `std::rand()`. It also means, that technically, you wouldn't be allowed to remove the `cstdlib` header.

Can't we just use int limit or some other large number?

eXpl0it3r: I mean one could argue whether we should actually use some unique identifier generator, but I guess a random number is okay for a short lived writing process. To prevent potential collisions, you'd just want a big pool for the distribution, so I'd just take `std::numeric_limits<int>::max()`.

ChrisThrasher:

eXpl0it3r (approved):

Issue #2113, opened 2022-06-04
Utf<8>::decode asserts in Visual Studio with bad UTF-8 string


texus:
Calling sf::String::fromUtf8 with an invalid UTF-8 string leads to a debug assertion in Visual Studio when compiling in debug mode.

Example code:
```c++
#include <SFML/System.hpp>
int main()
{
    std::string s(1, (char)251);
    sf::String::fromUtf8(s.begin(), s.end());
}
```

The `if (begin + trailingBytes < end)` line at [include/SFML/System/Utf.inl#L69](https://github.com/SFML/SFML/blob/master/include/SFML/System/Utf.inl#L69) checks whether there are sufficient trailing bytes, but if there aren't then then `begin + trailingBytes` will be an invalid iterator. Visual Studio asserts about this inside the `std::string::operator+` function (with "Expression: cannot seek string iterator after end" message).

Changing the if statement to something like the following would fix the assertion:
```c++
if (trailingBytes < std::distance(begin, end))
```

Bromeon:
What should the intended outcome be if the UTF-8 string is indeed invalid?

Assertion is better than just returning a valid string and hiding the error. But it should probably validate UTF-8 in the first place and not show up as a side effect in some STL implementations.

texus:
Personally I feel like providing an invalid string is a programming error and an assertion would be a valid response.
But whether the code silently hides it, asserts for it in debug mode or even throws an exception, the most important thing is that it is consistent on all platforms.

The behavior doesn't matter for me though as I don't actually use the affected code, I'm only reporting this because I discovered this error after someone reported a "crash" in my library (which uses code that was based on fromUtf8's implementation).

PR #2435, opened 2023-03-05
Fix condition for trailing bytes count in UTF-8 decoder.

123navyblue:
This PR fixes bug reported here: https://github.com/SFML/SFML/issues/2113

Originally, the `if` condition in question would be incorrect (and triggered an assertion in Visual Studio) when there are less bytes in the input range than inferred from the last decoded byte.
Using std::distance lets me safely calculate the remaining bytes available in the input sequence and then the result can be compared to the required number of trailing bytes.

To reproduce the bug:

```cpp
#include <SFML/System.hpp>
int main()
{
    std::string s(1, (char)251);
    sf::String::fromUtf8(s.begin(), s.end());
}
```

### DIFF
--- a/include/SFML/System/Utf.inl
+++ b/include/SFML/System/Utf.inl
@@ -69,7 +69,7 @@ In Utf<8>::decode(In begin, In end, std::uint32_t& output, std::uint32_t replace
 
     // decode the character
     int trailingBytes = trailing[static_cast<std::uint8_t>(*begin)];
-    if (begin + trailingBytes < end)
+    if (trailingBytes < std::distance(begin, end))
     {
         output = 0;
 
--- a/test/System/String.test.cpp
+++ b/test/System/String.test.cpp
@@ -230,18 +230,30 @@ TEST_CASE("[System] sf::String")
 
     SUBCASE("fromUtf8()")
     {
-        constexpr std::array<std::uint8_t, 4> characters{'w', 'x', 'y', 'z'};
-        const sf::String                      string = sf::String::fromUtf8(characters.begin(), characters.end());
-        CHECK(std::string(string) == "wxyz"s);
-        CHECK(std::wstring(string) == L"wxyz"s);
-        CHECK(string.toAnsiString() == "wxyz"s);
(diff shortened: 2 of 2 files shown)


kimci86 (approved):
Looks good. It would be nice to add a unit test about this after #2466.

ChrisThrasher:

ChrisThrasher:
Those `sf::String` tests just got merged if you'd like to take a shot and writing a test that reproduces this error.

123navyblue:
I've added a test which reproduces the situation and checks if a replacement character is added in such a case.

ChrisThrasher:
Fixed formatting

ChrisThrasher:
I think the test you added may be insufficient because when I add the tests but don't add the fix, the tests still passes. I'd expect this test to fail if the corresponding fix isn't included.

kimci86:
> I think the test you added may be insufficient because when I add the tests but don't add the fix, the tests still passes.

Did you try with MSVC? It was reportedly triggering an assert.

ChrisThrasher:

ChrisThrasher:
> > I think the test you added may be insufficient because when I add the tests but don't add the fix, the tests still passes.
> 
> Did you try with MSVC? It was reportedly triggering an assert.

Oh okay. I can recreate this crash and its subsequent fix with MSVC.

ChrisThrasher (approved):

Bromeon (approved):
Bugfix looks good, thanks!

I wonder if `fromUtf8` should not be made fallible (return `optional`) instead of returning 0. This would likely require bigger changes though, as such could be a separate discussion outside this PR.

ChrisThrasher:
> I wonder if `fromUtf8` should not be made fallible (return `optional`) instead of returning 0. This would likely require bigger changes though, as such could be a separate discussion outside this PR.

https://github.com/SFML/SFML/issues/2139#issuecomment-1515137328 Noted 👍

eXpl0it3r:
Thank you for the fix! 🙂

PR #2436, opened 2023-03-06
Make moving Shapes, swapping Textures and swapping VertexBuffers noexcept

binary1248:
As reported at https://sonarcloud.io/project/issues?resolved=false&severities=BLOCKER&id=SFML_SFML certain move operations and swaps weren't noexcept. The C++ core guidelines recommend these types of operations should never fail because they really shouldn't in properly designed code.

This changeset fixes these issues.

Additional points:
- In order to make `sf::Texture::swap()` noexcept, `getUniqueId()` also has to be noexcept, the new implementation uses an atomic counter
- In order to use certain atomic features, GCC requires explicitly linking libatomic when building on certain platforms e.g. armhf
- A `std::swap` test for `sf::Texture` was not added because it would cause too many tests to fail on Windows due to OpenGL issues (hopefully will be fixed at some point)

### DIFF
--- a/include/SFML/Graphics/Shape.hpp
+++ b/include/SFML/Graphics/Shape.hpp
@@ -47,12 +47,6 @@ class SFML_GRAPHICS_API Shape : public Drawable, public Transformable
 {
 public:
     ////////////////////////////////////////////////////////////
-    /// \brief Virtual destructor
-    ///
-    ////////////////////////////////////////////////////////////
-    ~Shape() override;
-
-    ////////////////////////////////////////////////////////////
     /// \brief Change the source texture of the shape
     ///
     /// The \a texture argument refers to a texture that must
--- a/include/SFML/Graphics/VertexBuffer.hpp
+++ b/include/SFML/Graphics/VertexBuffer.hpp
@@ -224,7 +224,7 @@ public:
     /// \param right Instance to swap with
     ///
     ////////////////////////////////////////////////////////////
-    void swap(VertexBuffer& right);
+    void swap(VertexBuffer& right) noexcept;
 
     ////////////////////////////////////////////////////////////
     /// \brief Get the underlying OpenGL handle of the vertex buffer.
@@ -337,6 +337,15 @@ private:
     PrimitiveType m_primitiveType{PrimitiveType::Points}; //!< Type of primitives to draw
     Usage         m_usage{Stream};                        //!< How this vertex buffer is to be used
 };
+
+////////////////////////////////////////////////////////////
+/// \brief Swap the contents of one vertex buffer with those of another
+///
+/// \param left First instance to swap
+/// \param right Second instance to swap
+///
+////////////////////////////////////////////////////////////
+void swap(VertexBuffer& left, VertexBuffer& right) noexcept;
 
 } // namespace sf
 
--- a/src/SFML/Graphics/VertexBuffer.cpp
+++ b/src/SFML/Graphics/VertexBuffer.cpp
@@ -33,7 +33,6 @@
 
 #include <cstddef>
 #include <cstring>
-#include <mutex>
 #include <ostream>
 #include <utility>
 
@@ -42,8 +41,6 @@ namespace
 // A nested named namespace is used here to allow unity builds of SFML.
 namespace VertexBufferImpl
 {
-std::recursive_mutex isAvailableMutex;
-
 GLenum usageToGlEnum(sf::VertexBuffer::Usage usage)
 {
     switch (usage)
@@ -274,7 +271,7 @@ VertexBuffer& VertexBuffer::operator=(const VertexBuffer& right)
 
 
 ////////////////////////////////////////////////////////////
-void VertexBuffer::swap(VertexBuffer& right)
+void VertexBuffer::swap(VertexBuffer& right) noexcept
 {
     std::swap(m_size, right.m_size);
     std::swap(m_buffer, right.m_buffer);
@@ -333,22 +330,15 @@ VertexBuffer::Usage VertexBuffer::getUsage() const
 ////////////////////////////////////////////////////////////
 bool VertexBuffer::isAvailable()
 {
-    std::lock_guard lock(VertexBufferImpl::isAvailableMutex);
-
-    static bool checked   = false;
-    static bool available = false;
-
-    if (!checked)
+    static const bool available = []() -> bool
     {
-        checked = true;
-
         TransientContextLock contextLock;
 
         // Make sure that extensions are initialized
         sf::priv::ensureExtensionsInit();
 
-        available = GLEXT_vertex_buffer_object;
-    }
+        return GLEXT_vertex_buffer_object;
+    }();
 
     return available;
 }
@@ -359,6 +349,13 @@ void VertexBuffer::draw(RenderTarget& target, const RenderStates& states) const
 {
     if (m_buffer && m_size)
         target.draw(*this, 0, m_size, states);
+}
+
+
+////////////////////////////////////////////////////////////
+void swap(VertexBuffer& left, VertexBuffer& right) noexcept
+{
+    left.swap(right);
 }
 
 } // namespace sf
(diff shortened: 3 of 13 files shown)


cmake/Macros.cmake
```diff
@@ -35,6 +35,17 @@ function(sfml_set_stdlib target)
...
+    # on some platforms (e.g. ARM), GCC requires linking libatomic to use <atomic> features
+    if(SFML_COMPILER_GCC)
+        include(CheckCXXSourceCompiles)
+        check_cxx_source_compiles("#include <atomic>
+            #include <cstdint>
+            int main(){std::atomic<std::uint64_t> x(1); return x.fetch_add(1);}" ATOMIC_TEST)
+        if(NOT ATOMIC_TEST)
+            target_link_libraries(${target} PUBLIC "-latomic")
+        endif()
+    endif()
```
ChrisThrasher, line 48: I frequently use GCC on ARM and have never seen this before. What versions of GCC on what platforms are giving you issue?

binary1248: [Our own armhf worker](https://ci.sfml-dev.org/#/builders/29/builds/1864/steps/5/logs/stdio).

cmake/Macros.cmake
```diff
@@ -35,6 +35,17 @@ function(sfml_set_stdlib target)
             message(FATAL_ERROR "Clang is the only supported compiler on macOS")
         endif()
     endif()
+
+    # on some platforms (e.g. ARM), GCC requires linking libatomic to use <atomic> features
```
ChrisThrasher, line 39: Can we clarify that it's specific armhf/RasPi that is the main motivator here?

cmake/Macros.cmake
```diff
@@ -35,6 +35,17 @@ function(sfml_set_stdlib target)
...
+    if(SFML_COMPILER_GCC)
```
ChrisThrasher, line 40: Does this same problem apply to those using Clang on Raspberry Pis?

cmake/Macros.cmake
```diff
@@ -35,6 +35,17 @@ function(sfml_set_stdlib target)
...
+        include(CheckCXXSourceCompiles)
+        check_cxx_source_compiles("#include <atomic>
+            #include <cstdint>
+            int main(){std::atomic<std::uint64_t> x(1); return x.fetch_add(1);}" ATOMIC_TEST)
```
ChrisThrasher, line 44: Minor knit but if you replaced `std::uint64_t` with `int` you could drop `<cstdint>`. Presumably it's `fetch_add` that causes issues, not that we used a 64 bit unsigned int.

include/SFML/Graphics/VertexBuffer.hpp
```diff
@@ -338,6 +338,15 @@ class SFML_GRAPHICS_API VertexBuffer : public Drawable, private GlResource
...
 };
 
+////////////////////////////////////////////////////////////
+/// \brief Swap the contents of one vertex buffer with those of another
+///
+/// \param left First instance to swap
+/// \param right Second instance to swap
+///
+////////////////////////////////////////////////////////////
+void swap(VertexBuffer& left, VertexBuffer& right) noexcept;
```
ChrisThrasher, line 348: Why did you add this free function? Does it mean we should make `sf::VertexBuffer::swap` a private method that is merely an implementation detail of this overload?

test/Graphics/RectangleShape.test.cpp
```diff
@@ -7,8 +7,8 @@
 
 static_assert(std::is_copy_constructible_v<sf::RectangleShape>);
 static_assert(std::is_copy_assignable_v<sf::RectangleShape>);
-static_assert(std::is_move_constructible_v<sf::RectangleShape>);
-static_assert(std::is_move_assignable_v<sf::RectangleShape>);
+static_assert(std::is_nothrow_move_constructible_v<sf::RectangleShape>);
+static_assert(std::is_nothrow_move_assignable_v<sf::RectangleShape>);
```
ChrisThrasher, line 11: 🥳

cmake/Macros.cmake
```diff
@@ -35,6 +35,17 @@ function(sfml_set_stdlib target)
...
+        if(NOT ATOMIC_TEST)
+            target_link_libraries(${target} PUBLIC "-latomic")
```
ChrisThrasher, line 46: Because we're only using this in Texture.cpp, can we link this as a `PRIVATE` dependency? I don't see why this needs to become a transitive dependency. Also, can't this merely be a link library for the graphics module only instead of a link library for every single module?

ChrisThrasher (commented):
I like this. Just a few comments about implementation tweaks then I'm good to approve.

cmake/Macros.cmake
binary1248: I've only seen people reporting about this using GCC since it was an active decision of theirs to split the atomic implementation out of the standard runtime library that is automatically linked in. The fact that this doesn't happen when building with GCC for aarch64 (RPi 64-bit) makes me think they regret this decision and have just followed what all other sane compilers do.

cmake/Macros.cmake
binary1248: Depending on whether the operation is translated into a compiler intrinsic or a function implementation, linking to the external library might not be required. Since this is what we use in the code, I figured it is safer to test for this exact code even though your version would be shorter. I can build in the shorter version and see what happens.

include/SFML/Graphics/VertexBuffer.hpp
binary1248: `std::is_nothrow_swappable_v` requires a non-member swap being available. There are people on both sides of the fence on whether to have only 1 or both implementations, but I personally don't see why we can't just keep both especially since removing the member function at this point would just be unnecessarily causing API breakage without much benefit.

cmake/Macros.cmake
binary1248: We have to be careful here. Currently `<atomic>` is only used in Texture.cpp. Because it is likely that we might use `<atomic>` in more places as we modernize the code and replace stuff with `atomic<T>`s, I wouldn't want to have to copy and paste the same block of code into multiple places. This is a very unusual library because it is needed to support standard library features, as such I wouldn't put them in the same basket as the other libraries. Depending on whether we end up using `<atomic>` in a header somewhere, the library might also have to be made public. I can change it to private for now if you want, but there is a possibility we will have to make it public again some time in the future.

ChrisThrasher: > I wouldn't want to have to copy and paste the same block of code into multiple places.

You'll never need to do that. As soon as we need this in two module we'll put it in some higher common location.

> there is a possibility we will have to make it public again some time in the future.

This is perfectly fine. We'll make that change when the time comes.

ChrisThrasher (approved):

include/SFML/Graphics/Shape.hpp
```diff
@@ -52,6 +52,30 @@ class SFML_GRAPHICS_API Shape : public Drawable, public Transformable
     ////////////////////////////////////////////////////////////
     ~Shape() override;
```
ChrisThrasher, line 53: https://godbolt.org/z/cYs4exPs3

@kimci86 Brought up how we can apparently remove `~Shape` which will implicit define all these other special member functions. Because `sf::Shape` inherits from a class that has a virtual distractor, it too inherits a virtual destructor. The above Godbolt link shows that this won't introduce memory leaks. We can use `std::has_virtual_destructor_v<T>` to assert this fact and ensure regressions don't slip in.

binary1248:
Fixed.

ChrisThrasher (approved):
😊

src/SFML/Graphics/VertexBuffer.cpp
```diff
@@ -333,22 +330,15 @@ VertexBuffer::Usage VertexBuffer::getUsage() const
...
-
         TransientContextLock contextLock;
 
         // Make sure that extensions are initialized
         sf::priv::ensureExtensionsInit();
 
-        available = GLEXT_vertex_buffer_object;
-    }
+        return GLEXT_vertex_buffer_object;
+    }();
```
eXpl0it3r, line 341: My C++-fu isn't the best here, does the static guarantee some atomicy, or why is the mutex no longer needed?

binary1248: https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables

eXpl0it3r (approved):

ChrisThrasher:
Thanks! Now let's see if SonarCloud is happier with these changes.

PR #2437, opened 2023-03-06
Use unity builds to speed up clang-tidy runtime

ChrisThrasher:
## Description

I recently learned that clang-tidy runs a lot faster when using unity builds. I can't exactly explain why it's true but the numbers seems to speak for themself:

Before:
image

After:
image

I had to disable `readability-redundant-declaration` in WindowUtil.hpp because we do compile time (over)optimization by not including WindowUtils.hpp in WindowUtils.cpp so clang-tidy can't know that this isn't really a redundant declaration.


ChrisThrasher:

ChrisThrasher:
Here are the results from the last time CI ran on `master`:
Screenshot 2023-03-06 at 4 35 31 PM

And this PR:
Screenshot 2023-03-06 at 4 34 58 PM

eXpl0it3r:
Could this potentially create blind spots in non unity builds?

eXpl0it3r (approved):

PR #2438, opened 2023-03-07
EGL pixel format selection and OpenGL version parsing fixes

### LINKED ISSUE #2395: Android OpenGL context 0.0 and antialiasing

binary1248:
We've seen that in certain situations e.g. running on certain Android platforms (#2395) or running tests on the GitHub actions VMs, the OpenGL implementation can become pretty broken and start returning values that don't follow the specification. Because failing to parse an OpenGL context version doesn't really hinder us from continuing operation if everything else works, this change just assumes a conservative version and continues instead of causing the application to terminate.

The EGL pixel format selection procedure before this change relied on `eglChooseConfig()`. Using this function to select a pixel format bears the same problems as its `wgl` and `glX` counterparts (which we don't use): The way the SFML API is designed, the user has to resort to guessing a conservative configuration to request or else it will fail and graphics becomes unusable or even causes the application to terminate (see #2395). `eglChooseConfig()` takes a set of attributes that it tries to match exactly to a pixel format available on the system. If an exact match isn't found it will just fail. The other context implementations are a bit more "user friendly" in this respect, if an exact match isn't found they will still return something that is a close enough match and emit a warning to the console if a certain attribute could not be met. This gives the user helpful feedback and allows them to adjust their requested settings based on e.g. tester feedback. This changeset makes EGL pixel format selection follow the same procedure as the other context types.

Fixes #2395
Closes #2422

### DIFF
--- a/src/SFML/Window/EglContext.cpp
+++ b/src/SFML/Window/EglContext.cpp
@@ -31,6 +31,7 @@
 #include <SFML/Window/EglContext.hpp>
 #include <SFML/Window/WindowImpl.hpp>
 
+#include <memory>
 #include <mutex>
 #include <ostream>
 #ifdef SFML_SYSTEM_ANDROID
@@ -80,18 +81,27 @@ EGLDisplay getInitializedDisplay()
 ////////////////////////////////////////////////////////////
 void ensureInit()
 {
-    static bool initialized = false;
-    if (!initialized)
-    {
-        initialized = true;
+    static std::once_flag flag;
 
-        // We don't check the return value since the extension
-        // flags are cleared even if loading fails
-        gladLoaderLoadEGL(EGL_NO_DISPLAY);
+    std::call_once(flag,
+                   []()
+                   {
+                       if (!gladLoaderLoadEGL(EGL_NO_DISPLAY))
+                       {
+                           // At this point, the failure is unrecoverable
+                           // Dump a message to the console and let the application terminate
+                           sf::err() << "Failed to load EGL entry points" << std::endl;
 
-        // Continue loading with a display
-        gladLoaderLoadEGL(getInitializedDisplay());
-    }
+                           assert(false);
+
+                           return false;
+                       }
+
+                       // Continue loading with a display
+                       gladLoaderLoadEGL(getInitializedDisplay());
+
+                       return true;
+                   });
 }
 } // namespace EglContextImpl
 } // namespace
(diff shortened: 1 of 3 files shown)


Zombieschannel (commented):
So, now that sfml 3 actually works on android, I tried this and the results are... questionable.
In normal sfml 3 I get these warnings/errors when running the app:
image

With these changes, I get the following:
image

But yeah generally, it does remove the OpenGL 0.0 error, but there still are errors.

My code for testing
```
#include <SFML/Graphics.hpp>
using namespace sf;

int main()
{
    sf::RenderWindow window;
    window.create(sf::VideoMode::getDesktopMode(), "SFML works!", sf::Style::Fullscreen);
    window.setFramerateLimit(10);
    sf::CircleShape shape(400.f);
    while (window.isOpen())
    {
        sf::Event event;
                window.close();
            if (event.type == sf::Event::LostFocus)
                window.setActive(0);
            if (event.type == sf::Event::GainedFocus)
                window.setActive(1);
                    window.close();
        }
        if (!window.hasFocus())
            continue;
        window.clear(Color::Black);
        window.draw(shape);
        window.display();
    }
}
```

binary1248:
That is the expected output with these changes. Because we are assuming version 1.1 if the version query fails, there is nothing for the settings check to warn about.

The context is still not being able to be set active, thus the error messages. Are you running the code in debug configuration? If you run the code in debug configuration, `eglCheck` should print additional information to the console explaining what is going wrong when trying to set the contexts active.

src/SFML/Window/EglContext.cpp
```diff
@@ -287,67 +297,100 @@ EGLConfig EglContext::getBestConfig(EGLDisplay display, unsigned int bitsPerPixe
...
+    // Retrieve the list of available configs
+    auto configs = std::make_unique<EGLConfig[]>(static_cast<std::size_t>(configCount));
+
+    eglCheck(eglGetConfigs(display, configs.get(), configCount, &configCount));
+
+    // Evaluate all the returned configs, and pick the best one
+    int       bestScore = 0x7FFFFFFF;
+    EGLConfig bestConfig{};
+
+    for (std::size_t i = 0; i < static_cast<std::size_t>(configCount); ++i)
```
ChrisThrasher, line 313: Do we know how big this number may get? Could it ever get big enough to cause problems?

src/SFML/Window/EglContext.cpp
```diff
@@ -287,67 +297,100 @@ EGLConfig EglContext::getBestConfig(EGLDisplay display, unsigned int bitsPerPixe
...
+        int caveat;
+        eglCheck(eglGetConfigAttrib(display, configs[i], EGL_RED_SIZE, &red));
+        eglCheck(eglGetConfigAttrib(display, configs[i], EGL_GREEN_SIZE, &green));
+        eglCheck(eglGetConfigAttrib(display, configs[i], EGL_BLUE_SIZE, &blue));
+        eglCheck(eglGetConfigAttrib(display, configs[i], EGL_ALPHA_SIZE, &alpha));
+        eglCheck(eglGetConfigAttrib(display, configs[i], EGL_DEPTH_SIZE, &depth));
+        eglCheck(eglGetConfigAttrib(display, configs[i], EGL_STENCIL_SIZE, &stencil));
+        eglCheck(eglGetConfigAttrib(display, configs[i], EGL_SAMPLE_BUFFERS, &multiSampling));
+        eglCheck(eglGetConfigAttrib(display, configs[i], EGL_SAMPLES, &samples));
+        eglCheck(eglGetConfigAttrib(display, configs[i], EGL_CONFIG_CAVEAT, &caveat));
```
ChrisThrasher, line 341: Is it worth abstraction this `eglGetConfigAttrib` call into a lambda that we can call with the `EGL_` macro as an argument thus allowing all these local vars to be `const` and reduce code duplication or is that overkill?

```cpp
const auto getConfigAttrib = [display, configs, i](auto idkWhatVarNameToUseHere) {
    int attribute;
    eglCheck(eglGetConfigAttrib(display, configs[i], idkWhatVarNameToUseHere, &attribute));
    return attribute;
}

const int red = getConfigAttrib(EGL_RED_SIZE);
```

src/SFML/Window/EglContext.cpp
binary1248: It's not that big... typically in the hundreds on a desktop system with a powerful GPU, in a mobile context probably less.

src/SFML/Window/EglContext.cpp
binary1248: This code was intentionally written to be identical in structure to the code in Glx and Wgl contexts for consistency reasons. We can consider refactoring them all at the same time as you propose in a separate issue later.

ChrisThrasher (approved):

binary1248:
Rebased to master.

eXpl0it3r (approved):

PR #2439, opened 2023-03-07
Make use of CODECOV_TOKEN in ci.yml in order to work around Github Actions API errors that occur

binary1248:
Title.

It seems like the failures have been getting pretty bad lately. If we provide CODECOV_TOKEN ourselves then codecov can skip the Github Actions API query that fails and we should see much less failures.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,7 @@ jobs:
       if: matrix.type.name == 'Debug' && github.repository == 'SFML/SFML' # Disable upload in forks
       uses: codecov/codecov-action@v3
       with:
+        token: ${{ secrets.CODECOV_TOKEN }}
         directory: ./build
         files: ./build/coverage.out
         fail_ci_if_error: true


ChrisThrasher (approved):
Thank you! I'm so tired of code coverage uploads constantly failing.

eXpl0it3r (approved):

ChrisThrasher:
https://github.com/SFML/SFML/actions/runs/4420970832/jobs/7751288583

Code coverage uploads are still failing en masse :( I'm not sure to what extent we can do anything more about this problem.

binary1248:
Add a retry action to automatically retry with a retry_wait_seconds that is something larger than the default. It should be very unlikely that the upload fails multiple times in a row when spaced further apart.

ChrisThrasher:
> Add a retry action to automatically retry with a retry_wait_seconds that is something larger than the default.

I'd love to do that but I didn't think I could use two actions in the same CI step. We already use the `codecov/codecov-action@v3` action.

binary1248:
See #2455.

PR #2442, opened 2023-03-10
Add support for using software rendering and improve CI

binary1248:
This is a moderately large change set.

First thing's first, the biggest feature is the addition of support to CMake to automatically download and install the Mesa 3D library to the build output directory so that it will be used instead of the system provided library on Windows.

This opens up multiple opportunities:
- We can use software rendering when running CI since it will provide a much better OpenGL implementation than the lousy one Windows provides by default
- We can use OpenGL->DX12 translation if for whatever reason on developer workstations the provided OpenGL implementation is lousy or no longer up-to-date
- We can build and test OpenGL ES code even on non-mobile platforms (currently only Windows) where normally only "desktop" OpenGL is provided by the operating system

Just this change alone should lower the barrier to entry of working on OpenGL/OpenGL ES code independent of the concrete system the developer is working on. It isn't unreasonable to assume that DX will always be well supported and widely available on Windows systems.

To make use of Mesa 3D, all the user has to do is enable the SFML_USE_MESA3D CMake option and CMake will take care of the rest when it runs the configuration. The necessary *.dlls will be automatically copied to the right places when examples/tests are built. For a developer working in their IDE this is not noticeable at all and they can keep working as they always have.

Further changes:
- Added Visual Studio OpenGL ES jobs to CI so that we can provide testing and coverage for that code as well
- Force all CI Mesa 3D implementations to use llvmpipe (software renderer), even when running on Unix-based platforms, this should provide a more consistent environment to work in
- Added graphics/display related tests that were impractical when using the default OpenGL implementation on Windows
- Fixed bugs that were discovered by the newly added tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,9 @@ name: CI
 
 on: [push, pull_request]
 
+env:
+  GALLIUM_DRIVER: llvmpipe # Use Mesa 3D software OpenGL renderer
+
 jobs:
   build:
     name: ${{ matrix.platform.name }} ${{ matrix.config.name }} ${{ matrix.type.name }}
@@ -11,12 +14,12 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-        - { name: Windows VS2019 x86,     os: windows-2019, flags: -A Win32 }
-        - { name: Windows VS2019 x64,     os: windows-2019, flags: -A x64 }
-        - { name: Windows VS2022 x86,     os: windows-2022, flags: -A Win32 }
-        - { name: Windows VS2022 x64,     os: windows-2022, flags: -A x64 }
-        - { name: Windows VS2022 ClangCL, os: windows-2022, flags: -T ClangCL }
-        - { name: Windows VS2022 Clang,   os: windows-2022, flags: -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -GNinja }
+        - { name: Windows VS2019 x86,     os: windows-2019, flags: -DSFML_USE_MESA3D=TRUE -A Win32 }
+        - { name: Windows VS2019 x64,     os: windows-2019, flags: -DSFML_USE_MESA3D=TRUE -A x64 }
+        - { name: Windows VS2022 x86,     os: windows-2022, flags: -DSFML_USE_MESA3D=TRUE -A Win32 }
+        - { name: Windows VS2022 x64,     os: windows-2022, flags: -DSFML_USE_MESA3D=TRUE -A x64 }
+        - { name: Windows VS2022 ClangCL, os: windows-2022, flags: -DSFML_USE_MESA3D=TRUE -T ClangCL }
+        - { name: Windows VS2022 Clang,   os: windows-2022, flags: -DSFML_USE_MESA3D=TRUE -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -GNinja }
         - { name: Linux GCC,            os: ubuntu-latest, flags: -DSFML_RUN_DISPLAY_TESTS=ON -GNinja }
         - { name: Linux Clang,          os: ubuntu-latest, flags: -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DSFML_RUN_DISPLAY_TESTS=ON -GNinja , gcovr_options: '--gcov-executable="llvm-cov-$CLANG_VERSION gcov"' }
         - { name: MacOS,                os: macos-11, flags: -GNinja }
@@ -38,10 +41,10 @@ jobs:
 
         include:
         - platform: { name: Windows VS2022, os: windows-2022 }
-          config: { name: Unity, flags: -DBUILD_SHARED_LIBS=TRUE -DCMAKE_UNITY_BUILD=ON }
+          config: { name: Unity, flags: -DSFML_USE_MESA3D=TRUE -DBUILD_SHARED_LIBS=TRUE -DCMAKE_UNITY_BUILD=ON }
           type: { name: Release }
         - platform: { name: Windows VS2022, os: windows-2022 }
-          config: { name: Unity, flags: -DBUILD_SHARED_LIBS=TRUE -DCMAKE_UNITY_BUILD=ON }
+          config: { name: Unity, flags: -DSFML_USE_MESA3D=TRUE -DBUILD_SHARED_LIBS=TRUE -DCMAKE_UNITY_BUILD=ON }
           type: { name: Debug, flags: -DCMAKE_BUILD_TYPE=Debug -DSFML_ENABLE_COVERAGE=TRUE }
         - platform: { name: MacOS, os: macos-11 }
           config: { name: Frameworks, flags: -GNinja -DSFML_BUILD_FRAMEWORKS=TRUE -DBUILD_SHARED_LIBS=TRUE }
@@ -112,6 +115,8 @@ jobs:
         shell: bash
         command: |
           if [ "${{ runner.os }}" == "Windows" ]; then
+            # Make use of a test to print OpenGL vendor/renderer/version info to the console
+            find $GITHUB_WORKSPACE/build/bin -name test-sfml-window.exe -exec {} --test-case="[Window] sf::Context" --subcase="Version String" \; | grep OpenGL
             # Run the tests
             cmake --build $GITHUB_WORKSPACE/build --target runtests --config ${{ matrix.type.name == 'Debug' && 'Debug' || 'Release' }}
             # Coverage is already generated on Windows when running tests.
@@ -124,6 +129,8 @@ jobs:
               fluxbox > /dev/null 2>&1 &
               sleep 5
             fi
+            # Make use of a test to print OpenGL vendor/renderer/version info to the console
+            find $GITHUB_WORKSPACE/build/bin -name test-sfml-window -exec {} --test-case="[Window] sf::Context" --subcase="Version String" \; | grep OpenGL
             # Run the tests
             ctest --test-dir $GITHUB_WORKSPACE/build --output-on-failure --config ${{ matrix.type.name == 'Debug' && 'Debug' || 'Release' }}
             # Run gcovr to extract coverage information from the test run
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -180,6 +180,10 @@ if(SFML_OS_WINDOWS)
     if(BUILD_SHARED_LIBS AND SFML_USE_STATIC_STD_LIBS)
         message(FATAL_ERROR "BUILD_SHARED_LIBS and SFML_USE_STATIC_STD_LIBS cannot be used together")
     endif()
+
+    sfml_set_option(SFML_USE_MESA3D FALSE BOOL "TRUE to use the Mesa 3D graphics library for rendering, FALSE to use the system provided library for rendering")
+
+    include(cmake/Mesa3D.cmake)
 endif()
 
 # setup Mac OS X stuff
--- a/cmake/Mesa3D.cmake
+++ b/cmake/Mesa3D.cmake
@@ -1,0 +1,86 @@
+set(MESA3D_URL "https://github.com/pal1000/mesa-dist-win/releases/download/23.0.0/mesa3d-23.0.0-release-msvc.7z")
+set(MESA3D_SHA256 "FEF8A643689414A70347AE8027D24674DEFD85E8D6428C8A9D4145BB3F44A3B0")
+
+get_filename_component(MESA3D_ARCHIVE "${MESA3D_URL}" NAME)
+get_filename_component(MESA3D_ARCHIVE_DIRECTORY "${MESA3D_URL}" NAME_WLE)
+
+if(${ARCH_64BITS})
+    set(MESA3D_ARCH "x64")
+else()
+    set(MESA3D_ARCH "x86")
+endif()
+
+set(MESA3D_ARCHIVE_PATH "${PROJECT_BINARY_DIR}/${MESA3D_ARCHIVE_DIRECTORY}/${MESA3D_ARCHIVE}")
+set(MESA3D_ARCH_PATH "${PROJECT_BINARY_DIR}/${MESA3D_ARCHIVE_DIRECTORY}/${MESA3D_ARCH}")
+
+# we support automatically installing and uninstalling the necessary files
+
+# if SFML_USE_MESA3D is set and true during configuration, we add custom commands to
+# automatically copy over the necessary files whenever an executable/test target is built
+
+# if SFML_USE_MESA3D is not set or false but the Mesa 3D directory is present, use its file list to
+# remove any files that were previously copied into ${PROJECT_BINARY_DIR}/bin and subdirectories
+
+if(SFML_OS_WINDOWS AND SFML_USE_MESA3D)
+    # we are installing the files
+
+    # if the Mesa 3D directory is not yet present, download and extract the
+    # files relevant for the architecture we are configured to build for
+    if(NOT EXISTS "${MESA3D_ARCH_PATH}")
+        message(STATUS "Downloading ${MESA3D_ARCHIVE}")
+
+        file(DOWNLOAD "${MESA3D_URL}" "${MESA3D_ARCHIVE_PATH}" SHOW_PROGRESS EXPECTED_HASH SHA256=${MESA3D_SHA256})
+
+        if(NOT EXISTS "${MESA3D_ARCHIVE_PATH}")
+            message(FATAL_ERROR "Failed to download ${MESA3D_URL}")
+        endif()
+
+        message(STATUS "Extracting ${MESA3D_ARCH} files from ${MESA3D_ARCHIVE}")
+
+        execute_process(COMMAND "${CMAKE_COMMAND}" -E tar x "${MESA3D_ARCHIVE_PATH}" -- ${MESA3D_ARCH} WORKING_DIRECTORY "${PROJECT_BINARY_DIR}/${MESA3D_ARCHIVE_DIRECTORY}")
+
+        file(REMOVE "${MESA3D_ARCHIVE_PATH}")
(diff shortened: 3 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -94,6 +103,15 @@ jobs:
...
 
+    - name: Run Tests
+      if: matrix.platform.name != 'Android' && matrix.config.name != 'iOS'
+      uses: nick-fields/retry@v2
+      with:
+        max_attempts: 10
+        timeout_minutes: 3
+        shell: bash
+        command: ${{matrix.platform.prefix}} cmake --build $GITHUB_WORKSPACE/build --config ${{ matrix.type.name == 'Debug' && 'Debug' || 'Release' }} --target runtests
+
```
ChrisThrasher, line 114: Why add this when the tests already get run as part of the `install` target?

CMakeLists.txt
```diff
@@ -196,6 +196,10 @@ if(SFML_OS_WINDOWS)
     if(BUILD_SHARED_LIBS AND SFML_USE_STATIC_STD_LIBS)
         message(FATAL_ERROR "BUILD_SHARED_LIBS and SFML_USE_STATIC_STD_LIBS cannot be used together")
     endif()
+
+    sfml_set_option(SFML_USE_MESA3D FALSE BOOL "TRUE to use the Mesa 3D graphics library for rendering, FALSE to use the system provided library for rendering")
+
+    include(cmake/Mesa3D.cmake)
```
ChrisThrasher, line 202: Should this `include` be hidden behind an `if(SFML_USE_MESA3D)`?

cmake/Mesa3D.cmake
```diff
@@ -0,0 +1,91 @@
...
+    add_custom_command(OUTPUT ${MESA3D_INSTALLED_FILES} COMMAND "${CMAKE_COMMAND}" ARGS "-E" "copy_if_different" ${MESA3D_FILE_LIST} "${PROJECT_BINARY_DIR}/bin$<IF:$<BOOL:${IS_MULTI_CONFIG}>,/$<CONFIG>,>")
+
+    add_custom_target("install-mesa3d" DEPENDS ${MESA3D_INSTALLED_FILES})
+
+    set_target_properties("install-mesa3d" PROPERTIES EXCLUDE_FROM_ALL TRUE)
+
+    # provide a helper to link to the install-mesa3d target
+    macro(sfml_use_mesa3d target)
+        add_dependencies(${target} "install-mesa3d")
+    endmacro()
```
ChrisThrasher, line 66: Does something this small really need to be an abstraction?

cmake/Mesa3D.cmake
```diff
@@ -0,0 +1,91 @@
...
+    foreach(MESA3D_FILE ${MESA3D_FILE_LIST})
+        get_filename_component(MESA3D_FILE_NAME "${MESA3D_FILE}" NAME)
+
+        list(APPEND MESA3D_INSTALLED_FILES "${PROJECT_BINARY_DIR}/bin/$<IF:$<BOOL:${IS_MULTI_CONFIG}>,$<CONFIG>/,>${MESA3D_FILE_NAME}")
+    endforeach()
+
+    # if files are missing from the target directory of the configuration being built, copy them over
+    add_custom_command(OUTPUT ${MESA3D_INSTALLED_FILES} COMMAND "${CMAKE_COMMAND}" ARGS "-E" "copy_if_different" ${MESA3D_FILE_LIST} "${PROJECT_BINARY_DIR}/bin$<IF:$<BOOL:${IS_MULTI_CONFIG}>,/$<CONFIG>,>")
+
+    add_custom_target("install-mesa3d" DEPENDS ${MESA3D_INSTALLED_FILES})
```
ChrisThrasher, line 59: We can omit quotes in tons of places but an easy to start is not quoting target names.

test/Graphics/RenderWindow.test.cpp
```diff
@@ -1,8 +1,60 @@
...
+
+        texture.update(window);
+
+        CHECK(texture.copyToImage().getPixel(sf::Vector2u({128, 128})) == sf::Color::Green);
+
+        window.clear(sf::Color::Blue);
+
+        texture.update(window);
+
+        CHECK(texture.copyToImage().getPixel(sf::Vector2u({196, 196})) == sf::Color::Blue);
```
ChrisThrasher, line 58: We can not double-space our code? It doubles the line count with aiding in readability in my opinion. There are some other places in this PR where I could say the same.

test/Graphics/RenderWindow.test.cpp
```diff
@@ -1,8 +1,60 @@
...
 static_assert(!std::is_copy_constructible_v<sf::RenderWindow>);
 static_assert(!std::is_copy_assignable_v<sf::RenderWindow>);
 static_assert(!std::is_nothrow_move_constructible_v<sf::RenderWindow>);
 static_assert(!std::is_nothrow_move_assignable_v<sf::RenderWindow>);
+
+TEST_CASE("[Graphics] sf::RenderWindow" * doctest::skip(skipDisplayTests))
+{
+    SUBCASE("Construction")
+    {
+        sf::RenderWindow window(sf::VideoMode(sf::Vector2u({256, 256}), 24),
```
ChrisThrasher, line 21:
```suggestion
        const sf::RenderWindow window(sf::VideoMode(sf::Vector2u(256, 256), 24),
```
I won't comment this everywhere, but please put `const` as many local variables used in the tests as possible. It's important to ensure that member functions that should be marked `const` actually are.

test/Graphics/Shader.test.cpp
```diff
@@ -1,8 +1,160 @@
...
+        float push_distance = storm_inner_radius + len / storm_total_radius * (storm_total_radius - storm_inner_radius);
+        vertex.xy = storm_position + normalize(offset) * push_distance;
+    }
+
+    gl_Position = gl_ProjectionMatrix * vertex;
+    gl_TexCoord[0] = gl_TextureMatrix[0] * gl_MultiTexCoord0;
+    gl_FrontColor = gl_Color;
+}
+)";
+constexpr auto geometrySource = R"(
```
ChrisThrasher, line 35:
```suggestion

constexpr auto geometrySource = R"(
```

test/Window/Context.test.cpp
```diff
@@ -1,8 +1,59 @@
...
+
+#if defined(SFML_SYSTEM_WINDOWS)
+        using glGetStringFuncType = const char*(__stdcall*)(unsigned int);
+#else
+        using glGetStringFuncType = const char* (*)(unsigned int);
+#endif
+
+        auto glGetStringFunc = reinterpret_cast<glGetStringFuncType>(sf::Context::getFunction("glGetString"));
+
+        REQUIRE_UNARY(!!glGetStringFunc);
```
ChrisThrasher, line 38: What does `REQUIRE_UNARY` do?

test/Window/Context.test.cpp
```diff
@@ -1,8 +1,59 @@
...
+    {
+        sf::Context context;
+
+        CHECK(context.setActive(true));
+
+#if defined(SFML_SYSTEM_WINDOWS)
+        using glGetStringFuncType = const char*(__stdcall*)(unsigned int);
+#else
+        using glGetStringFuncType = const char* (*)(unsigned int);
+#endif
```
ChrisThrasher, line 34: Seeing platform-related preprocessor conditionals in the tests makes me wonder if we're straying from the purpose of testing SFML's cross-platform API. If we have to do platform-specific things in the tests, that implies that users may have to do this too. Do we really need tests that are mostly just checking OpenGL implementation details? Do the existence of these tests hint at the need for an addition API for validation OpenGL details?

ChrisThrasher (changes requested):
I'm still trying to understand the big picture of this PR but here are some preliminary comments.

.github/workflows/ci.yml
binary1248: They don't any more since `runtests` was removed from all. Tests are run by explicitly building `runtests` and don't run when building any other target. This gives us more control over the timing and environment we run tests in, such as in this case retrying just running the tests if they spuriously fail. This is also kind of a solution to the issues brought up in #2429 and #2430.

CMakeLists.txt
binary1248: Mesa3D.cmake also includes functionality to remove the associated files if `SFML_USE_MESA3D` evaluates to `FALSE`.

cmake/Mesa3D.cmake
binary1248: I find it cleaner to encapsulate the implementation of this functionality since the consumer does not have to know "what happens behind the scenes" and it allows us make changes to the implementation without changing the call sites.

.github/workflows/ci.yml
ChrisThrasher: > They don't any more since runtests was removed from all

No, not yet. https://github.com/SFML/SFML/pull/2430

test/Window/Context.test.cpp
binary1248: `REQUIRE` aborts the subcase if the check fails, and `_UNARY` omits doctest trying to convert the expression into something it can pretty print. We only care if the pointer is non-null so `!!` converts it into a `bool`.

test/Window/Context.test.cpp
binary1248: A normal user would realistically never have to care about these specific details. The normal assumption is that they distribute their programs to people to run on hardware accelerated systems. In our case because we care about running in virtual CI environments as well, this is the simplest way for us to verify the OpenGL implementation is what we expect. The strings are dumped to the console before the tests are run.

.github/workflows/ci.yml
binary1248: I removed it from `ALL` in this PR already.

test/Window/Context.test.cpp
ChrisThrasher: I'm not sure how this improves on `REQUIRE(glGetStringFunc)`. We already do this testing the non-nullness of pointers. The printed output looks like this:

```
 /Users/thrasher/Projects/sfml/test/Network/Packet.test.cpp:98: SUCCESS: CHECK( packet.getData() != nullptr ) is correct!
  values: CHECK( 0x1410043b0 != nullptr )
```

The `_UNARY` suffix gives us this instead

```
/Users/thrasher/Projects/sfml/test/Network/Packet.test.cpp:98: SUCCESS: CHECK_UNARY( packet.getData() != nullptr ) is correct!
  values: CHECK_UNARY( true )
```

which I don't think is any better since you no longer see the code that failed inside the macro. When it fails you get much less information:

```
/Users/thrasher/Projects/sfml/test/Network/Packet.test.cpp:98: ERROR: CHECK_UNARY( packet.getData() == nullptr ) is NOT correct!
  values: CHECK_UNARY( false )
```

.github/workflows/ci.yml
ChrisThrasher: Ah I see now. That should let us remove the `${{matrix.platform.prefix}}` from the preceding build command.

cmake/Mesa3D.cmake
ChrisThrasher: It would be easier for me to understand the code if I could just read that we're adding a dependency between two targets rather than having to track down what this macro means. The lack of CTRL+Clicking on CMake macros to find their definitions means CMake abstractions have a higher cost to use than C++ abstractions.

test/CMakeLists.txt
```diff
@@ -176,5 +176,7 @@ endif()
 
 add_custom_command(TARGET runtests
                    COMMENT "Run tests"
-                   POST_BUILD COMMAND ${COVERAGE_PREFIX} ${CMAKE_CTEST_COMMAND} --output-on-failure -C $<CONFIG>
+                   POST_BUILD
+                   COMMAND "$<TARGET_FILE:test-sfml-window>" "--test-case=[Window] sf::Context" "--subcase=Version String" -m -ne -ni -nv -nc
+                   COMMAND ${COVERAGE_PREFIX} ${CMAKE_CTEST_COMMAND} --output-on-failure -C $<CONFIG>
                    VERBATIM)
```
ChrisThrasher, line 182: How is this going to complicate #2430? Can I still run all the tests by running `ctest` or does this mean I need to invoke special alternative command now? I hope running our tests locally is never more complicated than simply invoking `ctest`.

test/Graphics/RenderWindow.test.cpp
```diff
@@ -1,8 +1,60 @@
+#include <SFML/Graphics/Image.hpp>
 #include <SFML/Graphics/RenderWindow.hpp>
```
ChrisThrasher, line 2: We need to ensure that `SFML/Graphics/RenderWindow.hpp` remains the first include. Including it first means we're testing that it includes all the headers required to build itself. See how I fixed this in [WindowBase.test.cpp](https://github.com/SFML/SFML/blob/master/test/Window/WindowBase.test.cpp).

test/Graphics/Texture.test.cpp
```diff
@@ -24,4 +25,219 @@ TEST_CASE("[Graphics] sf::Texture" * doctest::skip(skipDisplayTests))
...
+
+            CHECK(texture.copyToImage().getPixel(sf::Vector2u({0, 0})) == sf::Color::Yellow);
+        }
+
+        SUBCASE("Pixels, size and destination")
+        {
+            REQUIRE(texture.create(sf::Vector2u({2, 1})));
+
+            texture.update(yellow, sf::Vector2u({1, 1}), sf::Vector2u({0, 0}));
+            texture.update(cyan, sf::Vector2u({1, 1}), sf::Vector2u({1, 0}));
```
ChrisThrasher, line 87:
```suggestion
            texture.update(cyan, sf::Vector2u(1, 1), sf::Vector2u(1, 0));
```
You can omit also these extra braces you put inside the `sf::Vector2` constructor calls.

test/Graphics/Texture.test.cpp
```diff
@@ -24,4 +25,219 @@ TEST_CASE("[Graphics] sf::Texture" * doctest::skip(skipDisplayTests))
...
     }
+
+    SUBCASE("create()")
+    {
+        sf::Texture texture;
+
+        CHECK(texture.create(sf::Vector2u({123, 456})));
+
+        CHECK(texture.getSize() == sf::Vector2u({123, 456}));
+    }
```
ChrisThrasher, line 36:
```cpp
    SUBCASE("create()")
    {
        sf::Texture texture;

        SUBCASE("At least one zero dimension")
        {
            CHECK(!texture.create({}));
            CHECK(!texture.create({0, 1}));
            CHECK(!texture.create({1, 0}));
        }

        SUBCASE("Valid size")
        {
            CHECK(texture.create({100, 100}));
            CHECK(texture.getSize() == sf::Vector2u(100, 100));
            CHECK(texture.getNativeHandle() != 0);
        }

        SUBCASE("Too large")
        {
            CHECK(!texture.create({100'000, 100'000}));
            CHECK(!texture.create({1'000'000, 1'000'000}));
        }
    }
```
    
Here are some more thorough `sf::Texture::create` tests I wrote [here](https://github.com/ChrisThrasher/SFML/commit/167f1d5bc5280d11d08d21bcbd002ced01f9afaf).

test/TestUtilities/GraphicsUtil.hpp
```diff
@@ -11,6 +11,12 @@
 #include <iomanip>
 #include <limits>
 
+#ifndef SFML_OPENGL_ES
+static constexpr bool skipOpenGLESTests = false;
+#else
+static constexpr bool skipOpenGLESTests = true;
+#endif
```
ChrisThrasher, line 18: Might be clearer to instead write this like "if macro is defined, set true, else set false" rather than "if macro is not defined set false, otherwise set true". It removes an extra negative from the expression.

CMakeLists.txt
ChrisThrasher: What do you mean remove the files? Is it bad if those files are present even if that cache variable is set to `OFF`?

binary1248: You wouldn't be able to use the system OpenGL implementation as long as they are there.

test/Window/Context.test.cpp
binary1248: Those functions return pointers to objects, we have a function pointer here. doctest tries to convert it to `void*` in order to print it as part of the expression decomposition and that will always fail because converting between function pointers and object pointers is something the compiler really tries to prevent you from doing.

test/CMakeLists.txt
binary1248: Nothing changes for people running ctest themselves. The first `COMMAND` line was just added for OpenGL implementation verification purposes when we build in CI.

src/SFML/Graphics/Texture.cpp
```diff
@@ -100,6 +100,34 @@ Texture::~Texture()
...
+
+////////////////////////////////////////////////////////////
+Texture& Texture::operator=(Texture&& right) noexcept
+{
+    Texture temp(std::move(right));
+
+    swap(temp);
+
+    return *this;
+}
```
vittorioromeo, line 128: This is a pretty inefficient way to move-assign. All we need to do is set `right.m_texture` to zero, but swapping all the other fields is unnecessary.

binary1248:
Updated with suggested changes and fixed context management in a couple places.

binary1248:
Rebased to master.

binary1248:
Rebased to master.

test/Window/Context.test.cpp
```diff
@@ -1,8 +1,61 @@
...
```
vittorioromeo, line 36:
```suggestion
#if defined(SFML_SYSTEM_WINDOWS)
#define STDCALL __stdcall
#else
#define STDCALL 
#endif

using glGetStringFuncType = const char*(STDCALL*)(unsigned int);
```

Makes it more clear that both branches are identical except for `__stdcall`.

binary1248: Fixed.

ChrisThrasher (approved):

binary1248:
Rebased to master.

vittorioromeo (approved):

PR #2446, opened 2023-03-14
Don't automatically run tests after build

### LINKED ISSUE #2429: Remove `runtests` target

ChrisThrasher:
## Description

Closes #2429
Supersedes #2430 

Removing `ALL` from the `runtests` target means it's no longer automatically built as part of a default build. This means the tests no longer get run after every build. Keeping this target around also means we're not forced to refactor all the code coverage so that can stay in place. I still hope to eventually remove code coverage from the build scripts and to eventually swap it all out with a uniform LLVM-based solution that works on all platforms but that was never my goal so it can wait.

See the extended commit message for there are two "Test" CI jobs. TL;DR It's because of OpenCppCoverage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,17 @@ jobs:
 
     - name: Build
       shell: bash
-      run: ${{matrix.platform.prefix}} cmake --build $GITHUB_WORKSPACE/build --config ${{ matrix.type.name == 'Debug' && 'Debug' || 'Release' }} --target install
+      run: cmake --build $GITHUB_WORKSPACE/build --config ${{ matrix.type.name == 'Debug' && 'Debug' || 'Release' }} --target install
+
+    - name: Test
+      if: runner.os == 'Windows'
+      shell: bash
+      run: cmake --build $GITHUB_WORKSPACE/build --target runtests --config ${{ matrix.type.name == 'Debug' && 'Debug' || 'Release' }}
+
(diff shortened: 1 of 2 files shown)


ChrisThrasher:

eXpl0it3r (approved):

ChrisThrasher:

eXpl0it3r (approved):

binary1248 (approved):

PR #2452, opened 2023-03-15
Switch to Catch2 v3

### LINKED ISSUE #2353: Replace Doctest with Catch2 v3
## Subject of the issue

SFML originally used Catch2 v2 but we switched to Doctest because Doctest had much better compile times. This was fair because Catch2 v2 was still a header-only library and its compile times were long enough to disrupt the flow of development. Since then Catch2 has released a new major version which fixes those compile time concerns by moving much of the implementation of the library into .cpp files that allow for better parallelization and shorter incremental build times.

Since the compile time concerns are no more, I think both libraries are relatively equal footing and we can reevaluate whether we want to keep using Doctest in the long run. Here's why I think we should switch back to Catch2:

1. The two libraries still have very similar interfaces so the work required to do this is minimum. I've done test implementations of this in the past and it was not particularly difficult. It's maybe 30 minutes of work.
2. Catch2 [generators](https://github.com/catchorg/Catch2/blob/devel/docs/generators.md) will let us more succinctly create large blocks of data that let us cover a wide range of inputs without a huge number of lines of code. [Here's](https://github.com/SFML/SFML/blob/cdce9f7147e01b0ef675cc5f115666bf09feeaec/test/Graphics/Transform.test.cpp#L70-L76) an example of where we might use them.
3. Catch2 [`STATIC_REQUIRE`](https://github.com/catchorg/Catch2/blob/devel/docs/other-macros.md) lets us better write tests for `constexpr` code. I find this more expressive than `static_assert`s and it should provide better feedback when failures occur too.
4. Catch2 [matcher expressions](https://github.com/catchorg/Catch2/blob/devel/docs/matchers.md#top) give us a powerful way to express test failure that includes things like regex matches and unit of least precision matching of floating point numbers.
5. Catch2 has built-in support for printing containers and `chrono` types which reduces how many custom printers we need to write and maintain.
6. Catch2's `SKIP` macro gives us another useful option for skipping tests that are not possible if certain hardware is not present. It's particularly useful for checking `sf::VertexBuffer::isAvailable` before attempting to run any tests.
7. Catch2 is generally a more feature-filled library than Doctest. I'm optimistic that we'll find more features that we can take advantage of in helpful ways.

It's also worth mentioning that I do personally prefer Catch2 and use it in many personal and professional projects so it's something Im more familiar with and that familiarity will help me get the most out of it. Conversely I don't use Doctest outside of SFML so I have no extra expertise to lend if we stick to it.

ChrisThrasher:
## Description

Closes #2353 

The 2nd commit takes advantage of Catch2's constexpr testing functionality to ensure that our `constexpr` interfaces are in fact able to be used in a `constexpr` context. What's great about this is that if these tests fail, you can set a preprocessor symbol that will instead run them normally so you can use a debugger and see a more verbose output to diagnose the failure.


ChrisThrasher:

vittorioromeo:
One of the reasons of choosing `doctest` was the fast compilation speed of the test suite. Could you please measure the impact of switching to Catch3?

ChrisThrasher:
> One of the reasons of choosing `doctest` was the fast compilation speed of the test suite. Could you please measure the impact of switching to Catch3?

https://github.com/SFML/SFML/issues/2353#issuecomment-1386527192

ChrisThrasher:
Here are some new build time measurements which should be more accurate now that tests won't get automatically run. I'm testing the 1st commit of this PR so that the impact of those compiletime tests is not measured.

`master`:
```
Benchmark 1: ninja clean && ninja
  Time (mean ± σ):      7.388 s ±  0.037 s    [User: 49.558 s, System: 6.837 s]
  Range (min … max):    7.325 s …  7.442 s    10 runs
```

https://github.com/SFML/SFML/commit/3869c1b67a7d61709bd385f6222af0db8f938947
```
Benchmark 1: ninja clean && ninja
  Time (mean ± σ):     10.605 s ±  0.104 s    [User: 74.648 s, System: 9.973 s]
  Range (min … max):   10.426 s … 10.780 s    10 runs
  ```

**Comment by ChrisThrasher** [comment]
It's hard to get a good read on how much this increases CI times. Looking at sample size of only 4 it appears to add maybe 15-45 seconds to the build job. Unsure how impacts the overall CI runtime since the overall CI runtime is prone to vary wildly from run to run.

**Comment by ChrisThrasher** [comment]
I enabled [`CATCH_CONFIG_FAST_COMPILE`](https://github.com/catchorg/Catch2/blob/devel/docs/configuration.md#catch_config_fast_compile) which disables some features we're not using in the name of faster compile times. Here are new build time measurements:

https://github.com/SFML/SFML/pull/2452/commits/0c613f47f3332bdcc91d96d1d74f15f18822890c
```
Benchmark 1: ninja clean && ninja
  Time (mean ± σ):     10.436 s ±  0.054 s    [User: 74.418 s, System: 9.786 s]
  Range (min … max):   10.343 s … 10.525 s    10 runs
```

Total PR (https://github.com/SFML/SFML/pull/2452/commits/11da66a22a2b92585e9f0433bb41fae26a093504)
```
Benchmark 1: ninja clean && ninja
  Time (mean ± σ):     11.408 s ±  0.574 s    [User: 83.280 s, System: 10.554 s]
  Range (min … max):   10.798 s … 12.472 s    10 runs
```

I can't measure a speedup but this seems worthwhile nonetheless.

**Comment by ChrisThrasher** [comment]

**Comment by ChrisThrasher** [comment]

**Comment by ChrisThrasher** [comment]

File: test/CMakeLists.txt [hunk 1]
```diff
@@ -67,9 +73,6 @@ set(WINDOW_SRC
     Window/WindowBase.test.cpp
 )
 sfml_add_test(test-sfml-window "${WINDOW_SRC}" SFML::Window)
-if(SFML_RUN_DISPLAY_TESTS)
-    target_compile_definitions(test-sfml-window PRIVATE SFML_RUN_DISPLAY_TESTS)
-endif()
```
eXpl0it3r, line 72: Is this no longer needed as it's being set on `sfml-test-main`

test/Graphics/VertexBuffer.test.cpp
```diff
@@ -3,30 +3,33 @@
...
-static_assert(std::is_copy_assignable_v<sf::VertexBuffer>);
-static_assert(std::is_move_constructible_v<sf::VertexBuffer>);
-static_assert(!std::is_nothrow_move_constructible_v<sf::VertexBuffer>);
-static_assert(std::is_move_assignable_v<sf::VertexBuffer>);
-static_assert(!std::is_nothrow_move_assignable_v<sf::VertexBuffer>);
-static_assert(std::is_nothrow_swappable_v<sf::VertexBuffer>);
-
 // Skip these tests because they produce flakey failures in CI when using xvfb-run
-TEST_CASE("[Graphics] sf::VertexBuffer" * doctest::skip(true))
+TEST_CASE("[Graphics] sf::VertexBuffer", "[.]")
```
eXpl0it3r, line 13: Maybe enhance the comment the explain what this magic string does `[.]`

test/Window/Cursor.test.cpp
```diff
@@ -1,25 +1,28 @@
...
 #include <type_traits>
 
-static_assert(!std::is_copy_constructible_v<sf::Cursor>);
-static_assert(!std::is_copy_assignable_v<sf::Cursor>);
-static_assert(!std::is_nothrow_move_constructible_v<sf::Cursor>);
-static_assert(!std::is_nothrow_move_assignable_v<sf::Cursor>);
-
 // Skip these tests because they fail when using DRM which hasn't implemented sf::Cursor
-TEST_CASE("[Window] sf::Cursor" * doctest::skip(true))
+TEST_CASE("[Window] sf::Cursor", "[.]")
```
eXpl0it3r, line 10: Also here, extend the comment

test/TestUtilities/WindowUtil.cpp
```diff
@@ -3,6 +3,17 @@
...
 #include <ostream>
 
+std::string runDisplayTests()
+{
+#ifdef SFML_RUN_DISPLAY_TESTS
+    return "";
+#else
+    // https://github.com/catchorg/Catch2/blob/devel/docs/test-cases-and-sections.md#special-tags
+    // The "[.]" tag tells Catch2 to not run a given TEST_CASE
+    return "[.]";
```
eXpl0it3r, line 13: Wouldn't it be possible to make this more speaking, like `[.displaytest]` or similar?

eXpl0it3r (changes requested):
Just some small documentation improvement around the skip tag

test/CMakeLists.txt
ChrisThrasher: That's right

test/TestUtilities/WindowUtil.cpp
ChrisThrasher: Good idea. I didn't know that was an option until I double-checked the Catch docs.

ChrisThrasher:

vittorioromeo:
> In general I'm not a fan of merging this PR in, the small extra conveniences that we obtain by using Catch2 instead of Doctest do not seem to justify the higher compilation times in CI IMHO.

Added rationale here: https://github.com/SFML/SFML/issues/2353#issuecomment-1485446000

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

Fixed some conflicts.

ChrisThrasher:
Rebased and fixed conflicts

ChrisThrasher:

kimci86:
ci.yml needs to be updated to use Catch's command line interface to run the test printing OpenGL vendor/renderer/version info to the console.

ChrisThrasher:
> ci.yml needs to be updated to use Catch's command line interface to run the test printing OpenGL vendor/renderer/version info to the console.

Thank you for catching this! It should be fixed now.

ChrisThrasher:

ChrisThrasher:

vittorioromeo:
@eXpl0it3r: have you taken a decision about this PR? I'm personally against merging due to the reasons mentioned [here](https://github.com/SFML/SFML/issues/2353#issuecomment-1485446000). I think this is where we need our "dictator" to make a choice :)

ChrisThrasher:
This is a textbook example of why we need a BDFL :) @eXpl0it3r

eXpl0it3r:
@vittorioromeo did you run this branch in your setup to see what the difference is for you?

I'm personally in favor or this change. Catch2 is from my experience more popular, thus you can get more people who are familiar with it. And it seems more feature rich, even if the usefulness of the features can be argued about.

ChrisThrasher:

Rebased on `master` and fixed conflicts.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

Rebased and fixed merge conflict.

eXpl0it3r (approved):
Haven't realized how many tests we've added over time, pretty cool 🙂

PR #2455, opened 2023-03-16
Manually upload coverage report instead of using codecov-action

binary1248:
Title.

According to Codecov maintainers, the reason why uploading sporadically fails is because they try to query _some_ data from the GitHub API and that can fail if they manage to hit their rate limit, e.g. if by coincidence many projects simultaneously try to upload coverage data at the same time. This seems to be independent of whether we provide the `CODECOV_TOKEN` ourselves or not. Obviously the most elegant solution would be for their backend to handle the limit and just retry a few seconds/minutes later but it seems like they are taking the route of trying to convince GitHub to make a rate limit exception for them.

Whatever they end up doing, because I don't like holding my breath over such things, this change adds the ability to keep on trying ~till we run out of cake~ until the upload finally succeeds. To be able to do this, we have to roll our own implementation of what the codecov-action would have did itself and call the binary using the retry action.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,8 @@ jobs:
   build:
     name: ${{ matrix.platform.name }} ${{ matrix.config.name }} ${{ matrix.type.name }}
     runs-on: ${{ matrix.platform.os }}
+    env:
+      HAVE_CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN != '' }}
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -108,14 +110,32 @@ jobs:
       if: matrix.type.name == 'Debug' && runner.os != 'Windows' # Coverage is already generated on Windows when running tests.
       run: gcovr -r $GITHUB_WORKSPACE -x $GITHUB_WORKSPACE/build/coverage.out -s -f 'src/SFML/.*' -f 'include/SFML/.*' ${{ matrix.platform.gcovr_options }} $GITHUB_WORKSPACE
 
+    - name: Download Codecov Uploader
+      if: matrix.type.name == 'Debug' && github.repository == 'SFML/SFML' && env.HAVE_CODECOV_TOKEN == 'true' # Disable upload in forks and when Codecov token isn't available
```
ChrisThrasher, line 114: > and when Codecov token isn't available

Why would the token not be available inside SFML/SFML itself?

.github/workflows/ci.yml
```diff
@@ -108,14 +110,32 @@ jobs:
...
+      run: |
+        platform=$(echo "${{ runner.os }}" | tr '[:upper:]' '[:lower:]')
+        if [ "$platform" == "windows" ]; then CODECOV_BINARY="codecov.exe"; else CODECOV_BINARY="codecov"; fi
+        curl https://keybase.io/codecovsecurity/pgp_keys.asc | gpg --no-default-keyring --keyring trustedkeys.gpg --import
+        curl -s --remote-name-all "https://uploader.codecov.io/latest/$platform/{$CODECOV_BINARY,$CODECOV_BINARY.SHA256SUM,$CODECOV_BINARY.SHA256SUM.sig}"
+        gpgv "$CODECOV_BINARY.SHA256SUM.sig" "$CODECOV_BINARY.SHA256SUM"
+        if [ "$platform" == "windows" ]; then powershell 'If ($(Compare-Object -ReferenceObject $(($(certUtil -hashfile codecov.exe SHA256)[1], "codecov.exe") -join "  ") -DifferenceObject $(Get-Content codecov.exe.SHA256SUM)).length -eq 0) {echo "codecov: OK"} Else {exit 1}';
+        else shasum -a 256 -c codecov.SHA256SUM; fi
+        chmod +x "$CODECOV_BINARY"
+        echo "CODECOV_BINARY=$CODECOV_BINARY" >> $GITHUB_ENV
```
ChrisThrasher, line 125: Is this just copied out of the codecov action itself?

ChrisThrasher (commented):
I agree this is a reasonable approach. Pending some minor questions I should be okay to approve this. It's a shame Codecov can't provide a more elegant solution.

.github/workflows/ci.yml
binary1248: There can be obscure scenarios where secrets are not available even when working within the main repository. I've seen cases of jobs in the past week or 2 in which the Codecov uploader complained that a token wasn't provided and reverted to auto-detect. It really shouldn't happen, but as many things in life things that shouldn't happen still do. It would also make it easier for us to detect since 2 whole steps would be skipped in the log and this is easier to scan for than having to open up the upload step and check whether it took the token or not.

.github/workflows/ci.yml
binary1248: More or less from [here](https://docs.codecov.com/docs/codecov-uploader)... with the exception of us using bash for most of the heavy lifting on Windows as well.

ChrisThrasher (approved):

PR #2456, opened 2023-03-17
Reduce TransientContextLock overhead if there is already a context active on the current thread

### LINKED ISSUE #2450: Calling `setUniform` on a `sf::Shader` results in a lock + several allocations
As the thread says, it's currently quite expensive to call `sf::Shader::setUniform`, so much that it dominates the performance overhead during some profiling of Open Hexagon.

```cpp
void Shader::setUniform(const std::string& name, float x)
{
    UniformBinder binder(*this, name);
    if (binder.location != -1)
        glCheck(GLEXT_glUniform1f(binder.location, x));
}
```

The `UniformBinder` contains a `TransientContextLock`:

```
GlResource::TransientContextLock::TransientContextLock()
{
    priv::GlContext::acquireTransientContext();
}

GlResource::TransientContextLock::~TransientContextLock()
{
    priv::GlContext::releaseTransientContext();
}
```

`acquireTransientContext`:

```cpp
void GlContext::acquireTransientContext()
{
    // Protect from concurrent access
    std::lock_guard lock(mutex); // <== GRABS LOCK (!)

    // If this is the first TransientContextLock on this thread
    // construct the state object
    if (!transientContext.has_value())
        transientContext.emplace(); // <== CREATES TRANSIENT CONTEXT (!)

    // Increase the reference count
    ++transientContext->referenceCount;
}
```

`TransientContext`:

```cpp
TransientContext()
{
    if (resourceCount == 0)
    {
        context.emplace(); // <== CREATES SF CONTEXT, LOCKS, ALLOCATES, ETC (!)
    }
    else if (!currentContext)
    {
        sharedContextLock.emplace(mutex); // <== TAKES LOCK (!)
        useSharedContext = true;
        sharedContext->setActive(true);
    }
}
```

---

All of this stuff is done every time a uniform is set. This is extremely expensive especially when changing multiple uniforms multiple times per frame. 

In a single-threaded application, all of this seems unnecessary. Even in a multithreaded application, it should be possible to create the context once, and then set uniforms multiple times.

**Q:** Is the transient context necessary in a single-threaded application? If so, why? *(i.e. why can't I just call `GLEXT_glUniform1f' and be done with it?)*

**Q:** Are all the allocations and context creation necessary? Can we somehow cache the "last accessed transient context" and avoid re-creating it until a different one is needed?

binary1248:
Title.

@vittorioromeo measured relatively high overhead when using TransientContextLock in #2450. This was mainly due to the locking that was necessary to check the value of `resourceCount`. This change adds a fast path to `acquireTransientContext()` and `releaseTransientContext()` that skips the locking and construction of `transientContext` if it is determined that there is already a context active on the current thread. Because `acquireTransientContext()` and `releaseTransientContext()` should be used symmetrically by `TransientContextLock` on the same thread, we shouldn't get any mismatches as long as the `TransientContextLock` isn't moved between threads.

I've tested this code in several different scenarios and the exotic use cases that have been reported to us over the years. This works for all of them. More testing would still be welcome, especially performance benchmarks of real-world applications.

Closes #2450

### DIFF
--- a/src/SFML/Window/GlContext.cpp
+++ b/src/SFML/Window/GlContext.cpp
@@ -177,6 +177,7 @@ unsigned int resourceCount = 0;
 
 // This per-thread variable holds the current context for each thread
 thread_local sf::priv::GlContext* currentContext(nullptr);
+thread_local unsigned int         currentContextTransientCount(0);
 
 // The hidden, inactive context that will be shared with all other contexts
 std::unique_ptr<ContextType> sharedContext;
@@ -201,14 +202,27 @@ struct TransientContext
     ////////////////////////////////////////////////////////////
     TransientContext()
     {
+        // TransientContext should never be created if there is
+        // already a context active on the current thread
+        assert(!currentContext);
+
+        std::unique_lock lock(mutex);
+
         if (resourceCount == 0)
         {
+            // No GlResources, no shared context yet
+            assert(!sharedContext);
+
+            // Create a Context object for temporary use
             context.emplace();
         }
-        else if (!currentContext)
+        else
         {
-            sharedContextLock.emplace(mutex);
-            useSharedContext = true;
+            // GlResources exist, currentContext not yet set
+            assert(sharedContext);
+
+            // Lock the shared context for temporary use
+            sharedContextLock = std::move(lock);
             sharedContext->setActive(true);
         }
     }
@@ -219,7 +233,7 @@ struct TransientContext
     ////////////////////////////////////////////////////////////
     ~TransientContext()
     {
-        if (useSharedContext)
+        if (sharedContextLock)
             sharedContext->setActive(false);
     }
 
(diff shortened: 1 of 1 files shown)


src/SFML/Window/GlContext.cpp
```diff
@@ -399,10 +400,22 @@ void GlContext::registerContextDestroyCallback(ContextDestroyCallback callback,
...
+    using GlContextImpl::currentContext;
+    using GlContextImpl::currentContextTransientCount;
     using GlContextImpl::mutex;
     using GlContextImpl::TransientContext;
     using GlContextImpl::transientContext;
 
+    // Fast path if we already have a context active on this thread
+    if (currentContext)
+    {
+        ++currentContextTransientCount;
```
ChrisThrasher, line 412: It feels like we're approximating a shared pointer. Does `currentContextTransientCount` not constitute a ref count?

binary1248: Both `currentContext` and `currentContextTransientCount` are `thread_local` so it lets us skip the expensive locking that would otherwise be required. Also `currentContext` is non-owning. `currentContextTransientCount` just helps us keep track of how many times we are "borrowing" the `currentContext`, there are no ownership semantics in play.

ChrisThrasher (approved):

ChrisThrasher:
@vittorioromeo have you had a chance to test this out yet?

vittorioromeo:
> @vittorioromeo have you had a chance to test this out yet?

Not yet, hope to give it a try later today or tomorrow.

binary1248:
Rebased onto master and removed more unnecessary code.

ChrisThrasher (approved):
I love all the debug asserts

eXpl0it3r (approved):

PR #2458, opened 2023-03-21
Prevent undefined behaviour in context destructors

cyanskies:
GL contexts access contextDestroyCallbacks in their destructors to iterate over remaining callbacks. During program shutdown contextDestroyCallbacks is destroyed before sharedContext, which causes the context to access a destroyed object. 

Global objects are destroyed in reverse order of their textual declaration, so moving contextDestroyCallbacks above the declaration of sharedContext ensures that it is destroyed after sharedContext's destructor has been called.

### DIFF
--- a/src/SFML/Window/GlContext.cpp
+++ b/src/SFML/Window/GlContext.cpp
@@ -178,18 +178,18 @@ unsigned int resourceCount = 0;
 // This per-thread variable holds the current context for each thread
 thread_local sf::priv::GlContext* currentContext(nullptr);
 
-// The hidden, inactive context that will be shared with all other contexts
-std::unique_ptr<ContextType> sharedContext;
-
-// Unique identifier, used for identifying contexts when managing unshareable OpenGL resources
-std::uint64_t id = 1; // start at 1, zero is "no context"
-
 // Set containing callback functions to be called whenever a
 // context is going to be destroyed
 // Unshareable OpenGL resources rely on this to clean up properly
 // whenever a context containing them is destroyed
 using ContextDestroyCallbacks = std::unordered_map<sf::ContextDestroyCallback, void*>;
 ContextDestroyCallbacks contextDestroyCallbacks;
+
+// The hidden, inactive context that will be shared with all other contexts
+std::unique_ptr<ContextType> sharedContext;
+
+// Unique identifier, used for identifying contexts when managing unshareable OpenGL resources
+std::uint64_t id = 1; // start at 1, zero is "no context"
 
 // This structure contains all the state necessary to
 // track TransientContext usage


ChrisThrasher:
Is this something caught by something like UBSan or by careful reading of the code?

ChrisThrasher:

cyanskies:
I encountered it when I had a project using std::exit. Under normal shutdown the callbacks list is empty before it's destroyed and the context destructor iterates over the empty list without issue even though it is a already destructed object.

With std::exit the list still has elements when it's destroyed so the size is reported as non-zero and the iteration accesses freed memory.

ChrisThrasher:
Is it possible to make a small program that demonstrates this error so we can reproduce it?

cyanskies:
Here's an example app, triggering Event::KeyPressed will try to close with std::exit and exhibit a memory access violation. Closing with the close button allows the app to close normally without issue.

```
#include "SFML/Window.hpp"
int main()
{
    // Create the main window
    sf::Window window(sf::VideoMode({ 800, 600 }), {}, sf::Style::Close);
    // Start the game loop
    while (window.isOpen())
    {
        // Process events
        {
            // Close window: exit
                window.close();
                std::exit(EXIT_SUCCESS);
        }
    }
    return EXIT_SUCCESS;
}
```

cyanskies:
All globals have static storage duration(except those with thread_local). See: [Storage class specifiers](https://en.cppreference.com/w/cpp/language/storage_duration) -> Storage Duration -> Static.

[std::exit](https://en.cppreference.com/w/cpp/utility/program/exit) specifies that objects with static storage duration will be destroyed in the reverse order of their initialisation.

This fix should work on any standards conforming implementation.

eXpl0it3r:
Ah right, been a while since I looked at this in detail 😄 

But isn't the issue then just that the initialization order is undefined (static initialization order fiasco), so it doesn't really help us to know that the destruction order is in reverse to the initialization order, if the initialization order is undefined?

cyanskies:
I believe this isn't an issue for this bug, see [Initialization](https://en.cppreference.com/w/cpp/language/initialization).
Under **Dynamic initialization**:
> 3) Ordered dynamic initialization, which applies to all other non-local variables: within a single translation unit, initialization of these variables is always [sequenced](https://en.cppreference.com/w/cpp/language/eval_order) in exact order their definitions appear in the source code. Initialization of static variables in different translation units is indeterminately sequenced. Initialization of thread-local variables in different translation units is unsequenced.

Initialization is well defined for a single translation unit(our case). Static initialization order fiasco applies when trying to sequence the initialization of objects from different translation units.

ChrisThrasher:
This fix definitely makes sense in theory but I can't reproduce it on macOS or Linux. Is it possible this is only affecting other platforms?

Bambo-Borris:
Managed to reproduce this crash with MSVC 17.2 on Windows with SFML master, and also Clang 15.0 on Windows with SFML master.

ChrisThrasher (approved):

Bambo-Borris:
Can confirm the fix works locally for me on both Clang & MSVC versions I mentioned, on Windows 11.

binary1248:
This change doesn't fix the actual problem. It just suppresses the segmentation fault/access violation that results from objects being accessed in the wrong order. `sharedContext` only became a `std::unique_ptr` in the recent mechanical conversion to smart pointers. This is one of the cases where we never intended for `sharedContext` to make use of smart pointer semantics. Before the conversion, calling `std::exit` would have just leaked the object and not cause any perceptible effects. `sharedContext` is and never was supposed to be destroyed in global context, it is just an unintended side-effect of the conversion.

When we write code in SFML, we make use of language guarantees that C++ offers us, such as RAII. The code is designed to have a deterministic (and correct) order or destruction if all language guarantees are maintained. By calling `std::exit` you are breaking these guarantees and although the behaviour of `std::exit` itself is defined in the standard you cannot expect C++ libraries to be written with `std::exit` usage in mind. `std::exit` is just baggage that C++ has to inherit from C for compatibility. Why the C++ committee decided to sprinkle a bit of their own fairy dust onto `exit` is beyond me. Just like you should never use `malloc` or `free` in a C++ program you should also never use `exit` or its bigger brother `std::exit`. There are enough other ways to terminate the process depending on what you want.
- `return 0`
- `std::terminate`
- `std::abort`
- `throw`
- `std::quick_exit`

`std::exit` is the only variant that basically rolls the dice on which destructors get to be called and which don't. It is so inconsistent, that the committee decided they needed `std::quick_exit` in C++11 to have something that was at least consistent in that it doesn't call any destructors whatsoever. If you want a way to end the process quickly while still returning a normal return status you should use it instead.

`std::exit` usage was never supported by SFML. The fact that we only noticed it here doesn't mean that things don't break in many other places as well.

Currently SFML still makes use of many global objects, way more than anybody would like to see. The biggest reason for this is because there was no better way to implement things in the pre-modern-C++ era. Now that we have finally started work on modernizing the code, we intend on replacing many of these constructs with better alternatives. It is a given that this specific spot will definitely get rewritten to reduce the number of objects that have to live in global namespace since we already have an object ownership hierarchy available to us.

Until the refactoring is done, I would still recommend staying away from `std::exit` and using `std::quick_exit` if absolutely necessary.

vittorioromeo:
@binary1248: given your last reply, should this PR be closed?

binary1248:
I already have a draft of a commit that would make this PR obsolete, so yes.

PR #2460, opened 2023-03-21
Export symbol required to use sfml-activity

ChrisThrasher:
## Description

I believe what was going on is that those on Windows building for Android didn't execute this code path meaning the Android shared libraries continued to have all symbols exported by default as is the case with GCC and Clang. https://github.com/SFML/SFML/commit/5fde1ca613b7c308b6155594e5d067b1a02f5179 meant that now even on Android those symbols are hidden which revealed that some symbols in that library aren't getting exported like they need to be.

### DIFF
--- a/src/SFML/Main/SFMLActivity.cpp
+++ b/src/SFML/Main/SFMLActivity.cpp
@@ -132,7 +132,7 @@ void* loadLibrary(const char* libraryName, JNIEnv* lJNIEnv, jobject& ObjectActiv
     return handle;
 }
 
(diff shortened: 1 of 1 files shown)


Bauumm:
It fixes the issue for me.

Zombieschannel:
> @Zombieschannel would you please check this out and let me know if it fixes the issues you were having the last time you tried to use SFML 3 on Android?

Yup, it works flawlessly now.

Zombieschannel (approved):
Can confirm that it fixed the issue.

binary1248 (approved):

PR #2461, opened 2023-03-21
SFML doesn't work on Android with extractNativeLibs=false (default for minSdkVersion >= 23 with Android Gradle Plugin >= 3.6)

texus:
When minSdkVersion is set to 23 or higher (Android 6.0 or newer) in the `app/build.gradle` file of the android example, the Android app will fails to start with an error like the following:
```
sfml-activity: dlopen("/data/app/.../lib/arm64/libopenal.so"): dlopen failed: library "/data/app/.../lib/arm64/libopenal.so" not found
```

The libopenal.so file itself has nothing to do with this, it is merely the first library that sfml-activity attempts to load. The reason lies in some changes that were made in Android 6 related to uncompressed vs compressed native libraries. The libraries aren't extracted to the location where dlopen tries to load them anymore.

I found references to an `uncompressNativeLibraries` option that could be changed to fix the issue, but I didn't test it because comments mentioned that changing it had negative side effects. So I ended up with the same strategy that is being used in SDL (https://github.com/libsdl-org/SDL/commit/18bcafffcedbf5c7ad6d74a7e74fd18aa02e0aca).

Instead of passing an absolute path to dlopen, we only need to provide the filename, which we already got by calling `System.mapLibraryName(libname)` (where `libname` would be e.g. `openal` while we need `libopenal.so`). We don't need to manually add the `nativeLibraryDir` path anymore. Since I have no idea on which configurations you still need a path, the existing code is kept and the new code is only executed if the old code failed.

### DIFF
--- a/src/SFML/Main/SFMLActivity.cpp
+++ b/src/SFML/Main/SFMLActivity.cpp
@@ -122,8 +122,21 @@ void* loadLibrary(const char* libraryName, JNIEnv* lJNIEnv, jobject& ObjectActiv
     void* handle = dlopen(libraryPath, RTLD_NOW | RTLD_GLOBAL);
     if (!handle)
     {
-        LOGE("dlopen(\"%s\"): %s", libraryPath, dlerror());
-        exit(1);
+        // When android:extractNativeLibs is set to false (which is the default when using AGP >= 3.6 and minSdkVersion >= 23),
+        // the above dlopen call may fail to find the library. This is because native libraries would be stored uncompressed
+        // in the APK and they would no longer be extracted to the system directory that was passed to dlopen.
+        // Applications are supposed to load the native library directory from the APK in this case. We can simply pass
+        // the library filename, without path, to dlopen to achieve this.
+        jstring     javaLibraryFilename = static_cast<jstring>(ObjectName);
+        const char* libraryFilename     = lJNIEnv->GetStringUTFChars(javaLibraryFilename, NULL);
+        handle                          = dlopen(libraryFilename, RTLD_NOW | RTLD_GLOBAL);
+        if (!handle)
+        {
+            LOGE("dlopen(\"%s\"): %s", libraryPath, dlerror());
+            exit(1);
+        }
+
+        lJNIEnv->ReleaseStringUTFChars(javaLibraryFilename, libraryFilename);
     }
 
     // Release the Java string


ChrisThrasher:

eXpl0it3r:
@zombieschannel does that work for you?

Zombieschannel:
Oof, well unfortunately, I do not have any issues in my test project when setting the minSdk to 23 or above. But I do usually keep it at 22 since that seemed to be the default. @texus should it not work on sfml 2 or sfml 3? And what device are you testing on?

texus:
I actually can’t build the SFML example on my PC without any changes, I always use a project with more modern settings. I expected the issue to happen for everyone, but I guess some other condition(s) need to be met then on top of just setting minSdkVersion to 23.

Information about my setup:
Phone: Nokia 2.2 (running Android 11)
Java version: 11.0.18 (JAVA_HOME=/usr/lib/jvm/java-11-openjdk/)
NDK version: 25.2.9519653
SFML version: master (ccda8a9)
OS: Manjaro Linux

I just started from a fresh clone of the SFML master branch, and made the following changes to the SFML/example/android folder. (I tried keeping all changed versions as low as possible)

In build.gradle, I’ve change the `com.android.tools.build:gradle` dependency from `3.0.0` to `4.0.0`

I've set both `compileSdkVersion` and `targetSdkVersion` to 23 as well in `app/build.gradle`, to demonstrate that this isn't caused by targetting new versions.

I use gradlew to build with a specific gradle version, I've added those files to the example directory and I’ve set gradle-6.1.1-all.zip in gradle-wrapper.properties (which is the oldest version that can be used with Android Gradle plugin 4.0.0). I run `./gradlew installDebug` in the example/android directory to build and install the example to my phone.

With just the above changes, if I set `minSdkVersion` to 22 then the app runs, if I set it to 23 then I get the following error:
```
sfml-activity: dlopen("/data/app/~~Oth7x9lYQ_fvZz33USB3Ag==/org.sfmldev.android-Cd4gzOLPtKyvBJFUFno3VA==/lib/arm/libopenal.so"): dlopen failed: library "/data/app/~~Oth7x9lYQ_fvZz33USB3Ag==/org.sfmldev.android-Cd4gzOLPtKyvBJFUFno3VA==/lib/arm/libopenal.so" not found
```

The APK contains `libopenal.so` in `/lib/armeabi-v7a/`, both when minSdkVersion is 22 and 23.

For completeness, SFML itself was configured with the following command:
```
cmake -DCMAKE_SYSTEM_NAME=Android -DCMAKE_ANDROID_NDK=/home/texus/Documents/android/ndk/25.2.9519653 -DCMAKE_ANDROID_ARCH_ABI=armeabi-v7a -DCMAKE_ANDROID_STL_TYPE=c++_static -DCMAKE_BUILD_TYPE=Debug
```

texus:
Since Android Gradle Plugin 3.6, the default value for "extractNativeLibs" is changed from "true" to "false" (https://developer.android.com/studio/past-releases/past-agp-releases/agp-3-6-0-release-notes#extractNativeLibs)

If I manually add `android:extractNativeLibs="true"` to the manifest (as attribute of the `application` tag), then I can run the app even with minSdkVersion set to 23.

So I the app only fails to launch for me (without this PR) when using both extractNativeLibs=false and minSdkVersion >= 23.

PR #2464, opened 2023-03-23
Upgrade to stb_image 2.28

ChrisThrasher:
## Description

https://src.fedoraproject.org/rpms/SFML/pull-request/2

Fixes some CVEs in stb_image 2.27

### DIFF
--- a/extlibs/headers/stb_image/stb_image.h
+++ b/extlibs/headers/stb_image/stb_image.h
@@ -1,4 +1,4 @@
-/* stb_image - v2.27 - public domain image loader - http://nothings.org/stb
+/* stb_image - v2.28 - public domain image loader - http://nothings.org/stb
                                   no warranty implied; use at your own risk
 
    Do this:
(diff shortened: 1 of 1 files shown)


binary1248 (approved):

eXpl0it3r:
Can you do this for `2.6.x`?

ChrisThrasher:
@binary1248 Extended CI is failing because it's trying to run clang-tidy on 2.6.x.

ChrisThrasher:
For the record, the addition of this PR introduced a new compiler warning that will get trigged if you're using a version of CMake older than 3.24. In 3.24 CMake fixed an issue that caused MSVC to emit warnings on 3rd party code. Without that fix, MSVC will detect and emit and warning about some lossy conversions in stb_image.h.

https://cmake.org/cmake/help/latest/release/3.24.html#generators
> The [Visual Studio Generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#visual-studio-generators) now support SYSTEM headers when using VS 2019 Update 11 or later.

Is this something worth fixing in SFML itself to avoid downstream users having to deal with it? I hope we can use `pragmas` as a workaround since I think modifying stb_image.h itself it out of the question.

PR #2465, opened 2023-03-23
Fix two issues present in Debian patches

ChrisThrasher:
## Description

Fixes https://salsa.debian.org/games-team/libsfml/-/blob/master/debian/patches/01_remove-googleapi-css.patch and https://salsa.debian.org/games-team/libsfml/-/blob/master/debian/patches/02_build-doc-once.patch.

The first one about removing a comment is mostly a curtesy to the Debian folks but we don't like commented code either so it makes sense.

The second one about the `doc` target is something I agree with. I never build the docs because I have to modify a cache variable then deal with the massive console output that Doxygen produces every single time I build the library. It's much better if I can leave that option enabled but only build docs on demand. I think it would be even better if we made Doxygen an optional dependency and only configured the docs if Doxygen is found. That way we don't even need the `SFML_BUILD_DOC` option.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -45,7 +45,7 @@ configure_file(${DOXYGEN_INPUT_DIR}/doc/header.html.in ${DOXYGEN_OUTPUT_DIR}/hea
 configure_file(${DOXYGEN_INPUT_DIR}/doc/doxygen.css ${DOXYGEN_OUTPUT_DIR}/html/doxygen.css COPYONLY)
 
 # target setup
-add_custom_target(doc ALL
+add_custom_target(doc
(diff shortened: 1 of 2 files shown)


binary1248 (approved):

eXpl0it3r (approved):

eXpl0it3r:
After this change the setup is kind of broken.

- If I set `SFML_BUILD_DOC` I expect the docs to actually be built and not just when I explicitly build the `doc` target.
- If you set `SFML_BUILD_DOC` and then build the `install` target, the build fails, because it tries to still install the docs, more specifically the `SFML.tag` file.
- If you want to install the docs this has now become a two step process, where first have to explicitly build the `doc` target and then build the `install` target.

Personally, I'd want to challenge this change again.

PR #2467, opened 2023-03-23
Fix clang-tidy-16 errors

ChrisThrasher:
## Description

clang-tidy-16 adds a new [`misc-use-anonymous-namespace`](https://clang.llvm.org/extra/clang-tidy/checks/misc/use-anonymous-namespace.html) check. We already prefer anonymous namespaces over using `static` so this check is a great addition.

LLVM was released last week and has just now landed in Homebrew so we should expect this error to start appearing in CI soon. I ran clang-tidy-16 locally to go ahead and catch all the new checks that we're violating.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,9 @@ jobs:
 
     - name: Install macOS Tools
       if: runner.os == 'macOS'
-      run: brew install gcovr ninja
+      run: |
+        brew update
+        brew install gcovr ninja
 
     - name: Install OpenCppCoverage and add to PATH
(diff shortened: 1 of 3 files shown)


ChrisThrasher:
I added `brew update` calls in both places where we're using brew so that it would start downloading LLVM 16 now and will ensure that it stays up to date going forward.

ChrisThrasher:

ChrisThrasher:
Screenshot 2023-03-25 at 7 15 35 PM

Can confirm LLVM 16 is now being used

binary1248 (approved):

PR #2469, opened 2023-03-24
Add WindowBase::setImePreEditPosition

Edgaru089:
## Description

This PR adds a method to `WindowBase`, `setImePreEditPosition(const Vector2i& pos)`, that hints to the OS/IME framework where to put the preedit window.

## Tasks

* [x] Tested on Linux (Fcitx5)
* [x] Tested on Linux (IBus)
* [x] Tested on Windows
* [ ] Tested on macOS (Not implemented)

I'm completely illiterate when it comes to macOS/Objective-C (or any other Apple stuff), and I need help on that part.

## How to test this PR?

Install an IME if you don't have one, and run this example.

This is relatively straightforward on Windows 10: install a language of Chinese Simplified/Traditional or Japanese in the Settings app, and press Win+Space to switch to that input.

```cpp
#include <SFML/Graphics.hpp>

int main() {
    sf::RenderWindow window(sf::VideoMode({ 400, 400 }), "IME position");
    window.setFramerateLimit(60);

    while (window.isOpen()) {
        for (sf::Event event; window.pollEvent(event);) {
                window.close();
            else if (event.type == sf::Event::MouseButtonPressed) {
                auto pos = sf::Mouse::getPosition(window);
                window.setImePreEditPosition({pos.x, pos.y});  // Try negating these values and see what happens    
            }
        }

        window.clear();
        window.display();
    }
}
```
Click around in the opened window, and try typing with the IME engaged.
图片


Edgaru089:
This patch also works with IBus on a fresh Ubuntu VM.
图片

src/SFML/Window/OSX/WindowImplCocoa.mm
```diff
@@ -523,4 +523,11 @@ void showMouseCursor()
 }
 
 
+////////////////////////////////////////////////////////////
+void WindowImplCocoa::setIMEPreeditPosition(const Vector2i& /* position */)
+{
+    // To implement
```
vittorioromeo, line 529:
```suggestion
    // TODO: To implement
```

Would be easier to grep.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1213,6 +1213,19 @@ void WindowImplX11::grabFocus()
...
+////////////////////////////////////////////////////////////
+void WindowImplX11::setIMEPreeditPosition(const Vector2i& position)
+{
+    XPoint xpos;
+    xpos.x = static_cast<short>(position.x);
+    xpos.y = static_cast<short>(position.y);
+    XVaNestedList preeditAttr;
+    preeditAttr = XVaCreateNestedList(0, XNSpotLocation, &xpos, NULL);
+    XSetICValues(m_inputContext, XNPreeditAttributes, preeditAttr, NULL);
+    XFree(preeditAttr);
```
vittorioromeo, line 1225:
```suggestion
    const XPoint xpos{static_cast<short>(position.x), static_cast<short>(position.y)};
    const XVaNestedList preeditAttr = XVaCreateNestedList(0, XNSpotLocation, &xpos, nullptr);
    XSetICValues(m_inputContext, XNPreeditAttributes, preeditAttr, nullptr);
    XFree(preeditAttr);
```

`const` can be omitted if the other functions do not accept it.

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -460,6 +460,22 @@ bool WindowImplWin32::hasFocus() const
...
+    HIMC himc;
+    if ((himc = ImmGetContext(m_handle)) != nullptr)
+    {
+        COMPOSITIONFORM cf;
+        cf.ptCurrentPos.x = static_cast<LONG>(position.x);
+        cf.ptCurrentPos.y = static_cast<LONG>(position.y);
+        cf.dwStyle        = CFS_FORCE_POSITION; // Don't let the IME adjust the position
+        ImmSetCompositionWindow(himc, &cf);
+        ImmReleaseContext(m_handle, himc);
+    }
```
vittorioromeo, line 475:
```suggestion
    if (HIMC himc = ImmGetContext(m_handle); himc != nullptr)
    {
        COMPOSITIONFORM cf;
        cf.ptCurrentPos.x = static_cast<LONG>(position.x);
        cf.ptCurrentPos.y = static_cast<LONG>(position.y);
        cf.dwStyle        = CFS_FORCE_POSITION; // Don't let the IME adjust the position
        ImmSetCompositionWindow(himc, &cf);
        ImmReleaseContext(m_handle, himc);
    }
```

Bromeon (commented):
Small note: we generally use PascalCase for abbreviations (`Utf`, `UdpSocket`, `Http`, `AlResource`).

This would mean `IMEPreeditPosition` should be `ImePreeditPosition`.

(Strictly speaking, `WindowImplDRM` is also not consistent, but that's out of scope for this PR).

src/SFML/Window/Win32/WindowImplWin32.cpp
Edgaru089: Ah yes, my coding habits are not quite C++17 yet :laughing:

src/SFML/Window/Unix/WindowImplX11.cpp
Edgaru089: `XSetICValues` takes a handle and a variable length argument list, and `XFree` takes a `void*` which is also the type for most X11 handles. Both of them make copies of data, so `const` should be fine, they're good old C functions anyway.

Edgaru089:
Issues fixed and rebased to latest master.

eXpl0it3r:
Is preedit really the most commonly referred to name for that window?  
I like the "Composition" term that the Windows API uses "nicer", mostly wonder though what term IME users generally use.

If we stick with preedit, I suggest to use `PreEdit` instead: `setImePreeditPosition` vs `setImePreEditPosition`

Edgaru089:
Rebased to latest master.

I won't prefer the word 'Composition'. Please correct me if I'm wrong, but in my understanding the word more describes the use of the Compose key to combine accent marks with Latin alphabet in some European languages. IMEs are something quite different.

eXpl0it3r:

Edgaru089:
Got pinged, tested and this still works on my own Linux box at least.

I also looked at SDL3, their version `SDL_SetTextInputArea` takes a rectangular area and an X offset, and in the backends:

 - X11/XIM only accepts a position so ( rect.left + offset, rect.centerY ) is used
 - Fcitx and IBus (called via DBus) both only accept a rect, so ( left=rect.left+offset, top=rect.top, width=rect.height, height=rect.height ) is used. Note for some reason the width is set to rect.height, instead of rect.width-offset as you would expect
 - Wayland (`zwp_text_input_v3`) only accepts a rect, so ( left=rect.left+offset, top=rect.top, width=1, height=rect.height ) is used. The width=1 part, well, it works i guess (associated pull request https://github.com/libsdl-org/SDL/pull/13087)
 - Windows accepts a rect and a position, so the rect and position (offset, rect.centerY) is used
 - Mac (Cocoa) seems to only accept a rect, the offset is not mentioned in the ObjC file and I'm not sure if it's used or not. I can't read ObjC.

So I'd suggest we change the API to take only a rect, and leave out the offset part, since the most we can do with the offset is just add it to rect.left?

PR #2473, opened 2023-03-26
Fixed GlContext not being activated if it shares the same address as a previously destroyed GlContext.

binary1248:
Depending on how merciful the memory allocator is, by chance we might end up creating a `sf::Context` (directly or indirectly through another object) that lives at the same memory address as a previously destroyed `sf::Context`. Because the check for whether a context is active or not is performed based on its address, this can mean a new context is not activated if it has the same address as an older one.

This change performs the comparison based on a unique ID and makes sure to reset the current context ID when it is destroyed if necessary. Even if a newer context is created at the same address, it should get a higher ID than the previous one and force the code to activate it if it wasn't already active.

Because this behaviour is highly dependent on the memory allocator, you might need a bit of luck to trigger it. The best chance of seeing it is by running the following code for a while in a 32-bit build.
```cpp
#include <SFML/Window/Context.hpp>

int main()
{
    while (true)
        sf::Context context;
}
```
It will eventually crash or trigger an assert somewhere.

### DIFF
--- a/src/SFML/Window/CMakeLists.txt
+++ b/src/SFML/Window/CMakeLists.txt
@@ -346,3 +346,14 @@ elseif(SFML_OS_IOS)
 elseif(SFML_OS_ANDROID)
     target_link_libraries(sfml-window PRIVATE android)
 endif()
+
+# on some platforms (e.g. Raspberry Pi 3 armhf), GCC requires linking libatomic to use <atomic> features
+# that aren't supported by native CPU instructions (64-bit atomic operations on 32-bit architecture)
+if(SFML_COMPILER_GCC)
+    include(CheckCXXSourceCompiles)
+    check_cxx_source_compiles("#include <atomic>
+        int main(){std::atomic_ullong x(1); return x.fetch_add(1);}" ATOMIC_TEST)
+    if(NOT ATOMIC_TEST)
+        target_link_libraries(sfml-window PRIVATE atomic)
+    endif()
+endif()
--- a/src/SFML/Window/GlContext.cpp
+++ b/src/SFML/Window/GlContext.cpp
@@ -32,6 +32,7 @@
 #include <glad/gl.h>
 
 #include <algorithm>
+#include <atomic>
 #include <cassert>
 #include <cctype>
 #include <cstdlib>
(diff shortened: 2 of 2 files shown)


ChrisThrasher (approved):

src/SFML/Window/GlContext.cpp
```diff
@@ -176,14 +177,15 @@ std::recursive_mutex mutex;
...
+thread_local std::uint64_t        currentContextId(0);
 thread_local sf::priv::GlContext* currentContext(nullptr);
 thread_local unsigned int         currentContextTransientCount(0);
 
 // The hidden, inactive context that will be shared with all other contexts
 std::unique_ptr<ContextType> sharedContext;
 
 // Unique identifier, used for identifying contexts when managing unshareable OpenGL resources
-std::uint64_t id = 1; // start at 1, zero is "no context"
+std::atomic<std::uint64_t> id(1); // start at 1, zero is "no context"
```
MarioLiebisch, line 188: I'd probably call it `nextContextId`, just to have it a bit more speaking.
```cpp
std::atomic<std::uint64_t> nextContextId(1); // internal ID for the next created context
```

src/SFML/Window/CMakeLists.txt
```diff
@@ -346,3 +346,14 @@ elseif(SFML_OS_IOS)
 elseif(SFML_OS_ANDROID)
     target_link_libraries(sfml-window PRIVATE android)
 endif()
+
+# on some platforms (e.g. Raspberry Pi 3 armhf), GCC requires linking libatomic to use <atomic> features
+# that aren't supported by native CPU instructions (64-bit atomic operations on 32-bit architecture)
+if(SFML_COMPILER_GCC)
+    include(CheckCXXSourceCompiles)
+    check_cxx_source_compiles("#include <atomic>
+        int main(){std::atomic_ullong x(1); return x.fetch_add(1);}" ATOMIC_TEST)
```
MarioLiebisch, line 355: Probably makes no difference, but since we're using `std::atomic<…>` below, wouldn't it be better to use that in this test, too? Over the years I've seen some rather interesting halfway-complete headers on some MinGW versions for example.

MarioLiebisch (approved):

src/SFML/Window/CMakeLists.txt
binary1248: This is what was already added to `CMakeLists.txt` for sfml-graphics, we didn't want to include `<cstdint>`. Since `std::atomic_ullong` is just a type alias to `std::atomic<unsigned long long>` I really don't see why it could have been left out when they added atomic support. There are some atomic types that are optional according to the standard, but this isn't one of them.

binary1248:
Updated.

MarioLiebisch (approved):

ChrisThrasher (approved):

PR #2474, opened 2023-03-26
Replaced xvfb-run with manually starting Xvfb and running fluxbox

binary1248:
Tests were failing sporadically due to the fact that the Xvfb server that is started by xvfb-run did not have enough time to start up before the tests were run.

This change replaces xvfb-run with manually starting Xvfb and running fluxbox to emulate a window manager. Many aspects of SFML deal with window manager interaction and thus having one to talk to will make the environment in which we test more realistic. Between Xvfb start, fluxbox start and test start we add delays of 5 seconds which should be enough to avoid races.

Since tests might still fail for various other unforeseen reasons, both test steps and coverage step are combined into a single script invocation that can be passed to the retry action.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,8 +17,8 @@ jobs:
         - { name: Windows VS2022 x64,     os: windows-2022, flags: -A x64 }
         - { name: Windows VS2022 ClangCL, os: windows-2022, flags: -T ClangCL }
         - { name: Windows VS2022 Clang,   os: windows-2022, flags: -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -GNinja }
-        - { name: Linux GCC,            os: ubuntu-latest, prefix: xvfb-run -a, flags: -DSFML_RUN_DISPLAY_TESTS=ON -GNinja }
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -95,18 +95,34 @@ jobs:
...
-      run: gcovr -r $GITHUB_WORKSPACE -x $GITHUB_WORKSPACE/build/coverage.out -s -f 'src/SFML/.*' -f 'include/SFML/.*' ${{ matrix.platform.gcovr_options }} $GITHUB_WORKSPACE
+      uses: nick-fields/retry@v2
+      with:
+        max_attempts: 10
+        timeout_minutes: 3
+        retry_wait_seconds: 5
+        warning_on_retry: false
+        shell: bash
+        command: |
+          if [ "${{ runner.os }}" == "Windows" ]; then
```
MarioLiebisch, line 106: Any specific reason to move the OS check down into the executed script instead of keeping it on the workflow level? I don't see too much duplicated settings otherwise.

If you keep it that way, should probably use single quotes (`'…'`), since it's static strings.

binary1248: The OS check is performed only once to decide if the step is run or not. In this case we have a second branch in the script to check if we are macOS or Linux. If we wanted to move it out to workflow level we would have 3 Test steps with identical action parameters (which actually occupy more lines than the command itself).

MarioLiebisch (approved):

eXpl0it3r (approved):

ChrisThrasher (approved):

PR #2478, opened 2023-03-29
Fixed OpenGL entry point loading on Windows

binary1248:
The way OpenGL entry points are loaded on Windows is unique compared to other platforms. Whereas on other platforms, the entry points returned by e.g. `glXGetProcAddressARB` can be used independently of the context they were queried in, the entry points returned by `wglGetProcAddress` should only be used in the context they were queried in. This means in theory, any OpenGL function could have multiple different addresses on Windows depending on the context that was active when the query was made.

In practice, we just have to differentiate between 2 cases on Windows, either an ICD (hardware accelerated vendor driver) is available or an ICD is not available. In the former case, `wglGetProcAddress` forwards the query to the ICD providing the context that is currently active and it returns the address of the function from the concrete ICD .dll. In the latter case, when an ICD is not available, Windows provides a generic GDI software rendering implementation. For whatever reason, when the generic implementation is in use, `wglGetProcAddress` cannot be used to query entry points available for the active context, it will always return `NULL`. Instead they have to be loaded from the generic implementation library (OpenGL32.dll) itself using `GetProcAddress`.

So in summary:
- ICD available -> Create ICD context, activate and load using `wglGetProcAddress`
- ICD not available -> Create generic context, activate and load using `GetProcAddress`

The current implementation in SFML isn't 100% correct since it will try to fall back to `GetProcAddress` even when an ICD context is active but `wglGetProcAddress` does not return an address for whatever reason.

This is bad for 2 reasons:
1. We get entry points that are incompatible with the context they are supposed to be used in
2. We get a false impression that something is supported when it really isn't

This change fixes OpenGL entry points being loaded from OpenGL32.dll instead of failing to load if they are not provided by the ICD on Windows.

This is technically still not a 100% correct implementation, since in theory there could be multiple ICDs installed on a given system, each providing their specific entry points. The entry points would have to be stored per context that is created instead of globally for the entire application. Because this scenario occurs so rarely if at all, any OpenGL entry point loader (like glad which SFML uses) assumes the same entry points can be used across any context that is created by the application and thus stores them globally. This is correct on any platform except Windows, but I guess they weren't willing to make an exception just for Windows.

The only way to test this is by running a test application on both a system that has an ICD installed and one that only provides the generic implementation and inspecting where the function pointers originate from. In the former case, they should be addresses in the ICD (e.g. atioglxx.dll for AMD, nvoglv32.dll for Nvidia) and in the latter case they should be addresses inside OpenGL32.dll.

### DIFF
--- a/src/SFML/Window/Win32/WglContext.cpp
+++ b/src/SFML/Window/Win32/WglContext.cpp
@@ -51,6 +51,18 @@ namespace WglContextImpl
 thread_local sf::priv::WglContext* currentContext(nullptr);
 
 
+// We use a different loader for wgl functions since we load them directly from OpenGL32.dll
+sf::GlFunctionPointer getOpenGl32Function(const char* name)
+{
+    static const HMODULE module = GetModuleHandleA("OpenGL32.dll");
+
+    if (module)
+        return reinterpret_cast<sf::GlFunctionPointer>(GetProcAddress(module, reinterpret_cast<LPCSTR>(name)));
+
+    return nullptr;
+}
+
+
 ////////////////////////////////////////////////////////////
 void ensureInit()
 {
@@ -59,7 +71,7 @@ void ensureInit()
     {
         initialized = true;
 
-        gladLoadWGL(nullptr, sf::priv::WglContext::getFunction);
+        gladLoadWGL(nullptr, getOpenGl32Function);
     }
 }
 
@@ -74,7 +86,7 @@ void ensureExtensionsInit(HDC deviceContext)
 
         // We don't check the return value since the extension
         // flags are cleared even if loading fails
-        gladLoadWGL(deviceContext, sf::priv::WglContext::getFunction);
+        gladLoadWGL(deviceContext, getOpenGl32Function);
     }
 }
 } // namespace WglContextImpl
@@ -186,6 +198,8 @@ WglContext::~WglContext()
 ////////////////////////////////////////////////////////////
 GlFunctionPointer WglContext::getFunction(const char* name)
 {
+    assert(WglContextImpl::currentContext != nullptr);
+
     auto address = reinterpret_cast<GlFunctionPointer>(wglGetProcAddress(reinterpret_cast<LPCSTR>(name)));
 
     if (address)
@@ -197,13 +211,9 @@ GlFunctionPointer WglContext::getFunction(const char* name)
             return address;
     }
 
-    static HMODULE module = nullptr;
-
-    if (!module)
-        module = GetModuleHandleA("OpenGL32.dll");
-
-    if (module)
-        return reinterpret_cast<GlFunctionPointer>(GetProcAddress(module, reinterpret_cast<LPCSTR>(name)));
+    // If we are using the generic GDI implementation, try loading directly from OpenGL32.dll as well
+    if (WglContextImpl::currentContext->m_isGeneric)
+        return WglContextImpl::getOpenGl32Function(name);
 
     return nullptr;
 }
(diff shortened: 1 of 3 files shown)


src/SFML/Window/Win32/WglContext.cpp
```diff
@@ -51,6 +51,18 @@ namespace WglContextImpl
 thread_local sf::priv::WglContext* currentContext(nullptr);
 
 
+// We use a different loader for wgl functions since we load them directly from OpenGL32.dll
+sf::GlFunctionPointer getOpenGl32Function(const char* name)
+{
+    static const HMODULE module = []() { return GetModuleHandleA("OpenGL32.dll"); }();
```
ChrisThrasher, line 57: How is this different from below?

```cpp
static const HMODULE module = GetModuleHandleA("OpenGL32.dll");
```

src/SFML/Window/Win32/WglContext.hpp
```diff
@@ -181,6 +181,7 @@ class WglContext : public GlContext
     HDC         m_deviceContext{}; //!< Device context associated to the context
     HGLRC       m_context{};       //!< OpenGL context
     bool        m_ownsWindow{};    //!< Do we own the target window?
+    bool        m_generic{};       //!< Is this context provided by the generic GDI implementation?
```
ChrisThrasher, line 184:
```suggestion
    bool        m_isGeneric{};       //!< Is this context provided by the generic GDI implementation?
```

ChrisThrasher (changes requested):

binary1248:
Updated.

ChrisThrasher (approved):

binary1248:
Rebased to master.

src/SFML/Window/Win32/WglContext.cpp
```diff
@@ -460,6 +470,14 @@ void WglContext::updateSettingsFromPixelFormat()
         return;
     }
 
+    // Detect if we are running using the generic GDI implementation and warn
+    if (actualFormat.dwFlags & PFD_GENERIC_FORMAT)
+    {
+        m_isGeneric = true;
+
+        err() << "Warning: Detected \"Microsoft Corporation GDI Generic\" OpenGL implementation" << std::endl;
```
eXpl0it3r, line 478: Was "The current OpenGL implementation is not hardware-accelerated" misleading, or why did you not add it here?

binary1248: At least according to the Microsoft documentation, there is a theoretical possibility that the GDI implementation can also be hardware accelerated. I'm pretty sure this never happened in reality, but the new wording corresponds more to their documentation. Besides, whether accelerated or not, if you are using the GDI implementation, you are going to have a bad time anyway (OpenGL 1.1, no extensions, etc.) so this one warning line should be enough as an indicator.

src/SFML/Window/Win32/WglContext.cpp
```diff
@@ -186,6 +198,8 @@ WglContext::~WglContext()
 ////////////////////////////////////////////////////////////
 GlFunctionPointer WglContext::getFunction(const char* name)
 {
+    assert(WglContextImpl::currentContext);
```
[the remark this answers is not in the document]
vittorioromeo: This comment is non-blocking, BTW.

src/SFML/Window/Win32/WglContext.cpp
binary1248: I added in a check for acceleration anyway. 😁

src/SFML/Window/Win32/WglContext.cpp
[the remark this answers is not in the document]
binary1248: Fixed.

vittorioromeo (approved):
Not confident on the OpenGL-related logic here but I am happy with the code changes.

ChrisThrasher:
https://github.com/SFML/SFML/issues/2495

This is causing crashes.

PR #2276, opened 2022-11-20
sf::Texture move constructor and operator 

Jim-Marsden:
Changed copy assignment to default copy assignment. Changed move tests to match nothrow move operations.

Thanks a lot for making a contribution to SFML! 🙂

Before you create the pull request, we ask you to check the follow boxes. (For small changes not everything needs to ticked, but the more the better!)

* [ ] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before?
* [x] Does the code follow the SFML [Code Style Guide](https://www.sfml-dev.org/style.php)?
* [x] Have you provided some example/test code for your changes?
* [ ] If you have additional steps which need to be performed list them as tasks!

----

## Description

I added a default move constructor and move assignment operator. 

I also removed the implementation for the copy assignment operator, making it a default operation.

I also rearranged things to better match the style as I understand it.

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Describe how to best test these changes. Please provide a [minimal, complete and verifiable example](https://stackoverflow.com/help/mcve) if possible, you can use the follow template as a start:

```cpp
#include <fmt/format.h>
#include <SFML/Graphics.hpp>

constexpr auto texture_loc{"data/textures/test.png"};

auto make_sprite(){
    struct {sf::Texture texture; sf::Sprite sprite;} result;

    if(not result.texture.loadFromFile(texture_loc)){
        throw std::runtime_error(fmt::format("Unable to open file: {}\n", texture_loc));
    }

    result.sprite.setTexture(result.texture);
    return result;
}


int main() try {
    sf::RenderWindow window{sf::VideoMode{{128, 128}}, "shader move test"};

    auto [texture, sprite] = make_sprite();

    while(window.isOpen()){
        for(sf::Event event{}; window.pollEvent(event);){
            if(sf::Event::Closed == event.type){
                window.close();
            }
        }

        sf::Texture move_test = std::move(texture);
        if(texture.getNativeHandle() == move_test.getNativeHandle()){
            throw std::runtime_error("handles match :C");
        }
        sprite.setTexture(move_test);

        window.clear();
        window.draw(sprite);
        window.display();

        texture = std::move(move_test);
    }
}
catch (std::runtime_error const & e){
    fmt::print(stderr, "Exception caught: {}\n", e.what());
}

```

PR #2479, opened 2023-03-29
Added sf::Texture tests

binary1248:
Title.

### DIFF
--- a/test/Graphics/Texture.test.cpp
+++ b/test/Graphics/Texture.test.cpp
@@ -1,5 +1,8 @@
 #include <SFML/Graphics/Texture.hpp>
 
+// Other 1st party headers
+#include <SFML/Graphics/Image.hpp>
+
 #include <doctest/doctest.h>
 
 #include <GraphicsUtil.hpp>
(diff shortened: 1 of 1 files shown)


ChrisThrasher (approved):

src/SFML/Graphics/Texture.cpp
```diff
@@ -100,6 +100,43 @@ Texture::~Texture()
...
+m_size(source.m_size),
+m_actualSize(source.m_actualSize),
+m_texture(source.m_texture),
+m_isSmooth(source.m_isSmooth),
+m_sRgb(source.m_sRgb),
+m_isRepeated(source.m_isRepeated),
+m_pixelsFlipped(source.m_pixelsFlipped),
+m_fboAttachment(source.m_fboAttachment),
+m_hasMipmap(source.m_hasMipmap),
+m_cacheId(TextureImpl::getUniqueId())
```
vittorioromeo, line 114: Is this correct, or do we want to reuse the same cache id as the moved-from texture? Not familiar with how the cache really works.

binary1248: We had problems before with trying to be smart about reducing the number of texture bind calls we did. Since this didn't really work out, we always assign new ids to force a rebind any time the object changes, even if the underlying texture object was moved.

eXpl0it3r: That's the main question of why #2276 wasn't merged yet 😅

Bromeon: > We had problems before with trying to be smart about reducing the number of texture bind calls we did. Since this didn't really work out, we always assign new ids to force a rebind any time the object changes, even if the underlying texture object was moved.

Should we maybe add a code comment to clarify this for future readers? 🙂

binary1248:
Remove texture implementation code since we can merge #2276 instead which is essentially identical.

eXpl0it3r (approved):

PR #2480, opened 2023-03-29
Implement `sf::String` in terms of `std::u32string`

### LINKED ISSUE #2470: Use `std::char32_t` for UTF-32 character type in `sf::String`
## Subject of the issue

https://en.cppreference.com/w/cpp/language/types#char32_t

I admittedly have no experience with non-ASCII strings but `std::char32_t` appears to be the modern C++ type we ought to use to represent a single character of a UTF-32 string. If you use a UTF-32 string literal, you get a pointer to this type which means users using UTF-32 string literals have to do a reinterpret_cast to use it with `sf::String`. It also means that strongly type string literals like `U"\U00104321"s` cannot be used with the `String(const std::basic_string<std::uint32_t>& u32String)` constructor since they're of type `std::u32string`, another C++11 type not accessible prior to SFML 3.

Perhaps we should implement `sf::String` in terms of `std::u32string` but that's a separate matter.

EDIT: This change would all but require we also implement `sf::String` in terms of `std::u32string`.

ChrisThrasher:
## Description

Closes #2470 

This will make it easier to do things like use UTF-32 string literals since `sf::String` will have a constructor that perfectly matches that type.


ChrisThrasher:

src/SFML/Network/Packet.cpp
```diff
@@ -381,7 +381,7 @@ Packet& Packet::operator>>(String& data)
         {
             std::uint32_t character = 0;
             *this >> character;
-            data += character;
+            data += static_cast<char32_t>(character);
```
Bromeon, line 384: Is there no `operator>>`  overload for `char32_t`?

Bromeon (commented):
Thanks! Mostly good, 2 minor comments 🙂

src/SFML/Network/Packet.cpp
ChrisThrasher: I think the `std::uint32_t` stream operations on `sf::Packet` need to be replaced with `char32_t` overloads.

src/SFML/System/String.cpp
```diff
@@ -106,15 +106,15 @@ String::String(const std::wstring& wideString)
...
     if (utf32String)
         m_string = utf32String;
 }
 
 
 ////////////////////////////////////////////////////////////
-String::String(const std::basic_string<std::uint32_t>& utf32String) : m_string(utf32String)
+String::String(const std::u32string& utf32String) : m_string(utf32String)
 {
 }
```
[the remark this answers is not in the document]
ChrisThrasher: I think there's a clang-tidy check for that very thing if you're interested in adding it. I can do that in a later PR.

https://clang.llvm.org/extra/clang-tidy/checks/performance/move-constructor-init.html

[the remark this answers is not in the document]
Bromeon: Ah, interesting. From the description:

> The check flags user-defined move constructors that have a ctor-initializer initializing a member or base class through a copy constructor instead of a move constructor.

That wouldn't fire here though, as this is not a move constructor, or would it? Might need testing...

src/SFML/Network/Packet.cpp
Bromeon: Makes sense. They could probably reuse the `uint32_t` implementation internally.

ChrisThrasher: On second thought, all those integer stream operations on `sf::Packet` exist for all integer types so we shouldn't remove the `std::uint32_t` overloads. We could add a `char32_t` but that seems a bit weird since no other character type gets its own overloads. _Pointers_ to character types are included but not character types by value. I'm not really sure what is the most proper solution is here.

ChrisThrasher: @eXpl0it3r Any API design preferences here? Do we add a convenience for `char32_t` serialization in spite of no other character types getting the same treatment? Do we add overloads for `char`, `char16_t`, and `char32_t`?

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

test/System/String.test.cpp
```diff
@@ -74,8 +84,8 @@ TEST_CASE("[System] sf::String")
             CHECK(string.toAnsiString() == ""s);
             CHECK(string.toWideString() == L""s);
             CHECK(string.toUtf8() == std::basic_string<std::uint8_t>());
-            CHECK(string.toUtf16() == std::basic_string<std::uint16_t>());
-            CHECK(string.toUtf32() == std::basic_string<std::uint32_t>());
+            CHECK(string.toUtf16() == u""s);
+            CHECK(string.toUtf32() == U""s);
```
Bromeon, line 88: Minor thing: there is a `operator==` between `basic_string<T>` and `const T*`, no?
So it's not needed to use the `s` postfix and allocate for each right-hand-side argument.

(same for other occurrences)

test/System/String.test.cpp
```diff
@@ -179,38 +189,38 @@ TEST_CASE("[System] sf::String")
...
+            CHECK(string.toUtf32() == U"\U0010AFAF"s);
             CHECK(string.getSize() == 1);
             CHECK(!string.isEmpty());
             CHECK(string.getData() != nullptr);
         }
 
         SUBCASE("UTF-32 C string constructor")
         {
-            const sf::String string(reinterpret_cast<const std::uint32_t*>(U"\U0010ABCDrs"));
+            const sf::String string(U"\U0010ABCDrs");
```
Bromeon, line 201: Since those constructors are implicit, should we use the `const sf::String string = U"...";` syntax?

test/System/String.test.cpp
ChrisThrasher: That's true. The problem then becomes that we need to add even more `doctest::StringMaker` specializations for all those different types of null-terminal strings. We can do that, it's just tedious.

ChrisThrasher:

test/System/String.test.cpp
Bromeon: Ah, I see, in that case let's keep the tests simpler.

> In either case, I don't think any allocations are occurring since all these strings are so small.

I wasn't aware that small string optimization was common practice, [but it seems to be](https://stackoverflow.com/q/21694302). Thanks!

Bromeon (approved):

PR #2484, opened 2023-04-01
Disable stringop-overflow warning when including minimp3 since it produces false positives.

binary1248:
When building with 32-bit GCC 12.2.0 with -O3 the stringop-overflow warning produces false positives when including minimp3.

https://gcc.godbolt.org/z/aq7zf1z1q

CI currently fails to because of this, see [here](https://ci.sfml-dev.org/#/builders/42/builds/3/steps/13/logs/stdio).

This change disables stringop-overflow when including minimp3.

Until this change is merged, no other PRs will pass checks since MinGW 12.2.0 is a requirement now.

### DIFF
--- a/src/SFML/Audio/SoundFileReaderMp3.cpp
+++ b/src/SFML/Audio/SoundFileReaderMp3.cpp
@@ -34,12 +34,17 @@
 #ifdef _MSC_VER
 #pragma warning(push)
 #pragma warning(disable : 4242 4244 4267 4456 4706)
+#else
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-overflow"
 #endif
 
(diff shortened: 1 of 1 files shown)


ChrisThrasher (approved):

Bromeon:
Would it make sense to add a short comment explaining what this warning suppression does?

binary1248:
I don't really see the point in commenting this addition since 1. -isystem should have prevented the need for this 2. this warning is known to create false positives and 3. we already disable a bunch of warnings for MSVC without a comment either. It should be obvious when looking at this include why we had to disable the warning so a comment here wouldn't add much value.

eXpl0it3r:
I rebased to `2.6.x` since we'd need it there as well.

PR #2486, opened 2023-04-02
Remove bugprone `sf::Text` constructor

### LINKED ISSUE #2194: Replace the default constructor of sf::Text with one that takes an sf::Font
_Inspired by a case of a user asking the Discord server for help in why their text was not displaying despite them using `setString` `setStyle` `setFillColor` `setCharacterSize`. They lacked `setFont`._

This suggests a breaking change, so I propose this be added to SFML 3. The change is trivial so doesn't take much work, but provides safety and a clearer API for the user.

### Proposal

Change the default constructor of `sf::Text` from:
`Text()`
Into:
`Text(const Font& font, unsigned int charachterSize = 30)` 

The reasoning is that allowing the user to create an `sf::Text` with no font is not useful, and opens up for confusion and bugs since rendering an `sf::Text` with no font will just display nothing. Especially with SFML's history of providing a default font, and also with many beginners not realising that a font needs to be set, this opens up a pitfall. 

By removing the ability to even create an `sf::Text` in this state removes that class of bugs and goes along good API practices of making APIs harder to use without compromising their usefulness.

## Drawback

The only obvious rebuttal I see to this are things like:
- But I want to put uninitialised `sf::Text` in an `std::vector`
- But I create `sf::Text`s in my engine that are going to reference not-yet-loaded `sf::Font`s and I'll use `setFont` when the time comes
- ...similar...

These all boil down to "Sometimes it is useful to be able to construct uninitialised values for later use". This is true, and I'd argue the proper modern C++ way of doing this would be through `std::optional<sf::Text>` or `std::unique_ptr<sf::Text>` or similar, which to some might seem more verbose but this is a good type of verbose as it:
- Makes clear *which* usages of `sf::Text` might be uninitialised (currently ALL usages of `sf::Text` could be, which is far from the default case)
- Uses well tried and tested standard provided methods of creating uninitialised values that communicates intent and provides helper methods for handling it.
- Makes it hard to accidentally come across uninitalised text objects, in  contract to currently when ANY `sf::Text` opaquely might be not ready for use.

With this in mind, I personally don't see any further drawbacks with this change.

## Compile errors

@ChrisThrasher did a quick test to build SFML with the default constructor removed, and there was an error in the joystick code where it seems like it uses an uninitialised instane of `sf::Text`. This is hopefully trivially fixed by changing it to `std::optional<sf::Text>`. 

## Summary

All in all, I think this would be a great change since it adds seatbelts against an issue that might otherwise trip up unexpecting users. It brings the API closer to being safe and modern, utilising the type system for bulletproof eradication of the described bugs. The change is seemingly easy to apply and I don't see any unaddressed drawbacks.

Interested in hearing the thoughts from the regular SFML maintainers in case I've missed something.

ChrisThrasher:
## Description

Closes #2194 

The default constructor for `sf::Text` made it possible to accidentally forget to assign a font. Removing that default constructor means users are forced to supply a font making `sf::Text` harder to misuse. While I was at it, I also tweaked the existing constructor to allow users to specify a font without a default string to avoid the needless `""` argument.

This required some changes to the Joystick example which made heavy use of default constructed text objects. The modifications to that example prove that a default constructor is not required to effectively use `sf::Text`.

### DIFF
--- a/examples/joystick/Joystick.cpp
+++ b/examples/joystick/Joystick.cpp
@@ -41,8 +41,9 @@ void updateIdentification(unsigned int index)
 {
     sstr.str("");
     sstr << "Joystick " << index << ":";
-    texts.at("ID").label.setString(sstr.str());
-    texts.at("ID").value.setString(sf::Joystick::getIdentification(index).name);
+    auto& [label, value] = texts.at("ID");
+    label.setString(sstr.str());
+    value.setString(sf::Joystick::getIdentification(index).name);
 }
 
 // Update joystick axes
@@ -103,37 +104,40 @@ int main()
     sstr.setf(std::ios::fixed | std::ios::boolalpha);
 
     // Set up our joystick identification sf::Text objects
-    texts.emplace("ID", JoystickObject{{"<Not Connected>", font}, {"", font}});
-    texts.at("ID").label.setPosition({5.f, 5.f});
-    texts.at("ID").value.setPosition({80.f, 5.f});
+    {
+        auto [it, success]   = texts.emplace("ID", JoystickObject{{font, "<Not Connected>"}, {font}});
+        auto& [label, value] = it->second;
+        label.setPosition({5.f, 5.f});
+        value.setPosition({80.f, 5.f});
+    }
 
     // Set up our threshold sf::Text objects
     sstr.str("");
     sstr << threshold << "  (Change with up/down arrow keys)";
-
-    texts.emplace("Threshold", JoystickObject{{"Threshold:", font}, {sstr.str(), font}});
-    texts.at("Threshold").label.setPosition({5.f, 5.f + 2 * font.getLineSpacing(14)});
-    texts.at("Threshold").value.setPosition({80.f, 5.f + 2 * font.getLineSpacing(14)});
+    {
+        auto [it, success]   = texts.emplace("Threshold", JoystickObject{{font, "Threshold:"}, {font, sstr.str()}});
+        auto& [label, value] = it->second;
+        label.setPosition({5.f, 5.f + 2 * font.getLineSpacing(14)});
+        value.setPosition({80.f, 5.f + 2 * font.getLineSpacing(14)});
+    }
 
     // Set up our label-value sf::Text objects
     for (unsigned int i = 0; i < sf::Joystick::AxisCount; ++i)
     {
-        auto& object = texts.insert({axislabels[i], {{axislabels[i] + ":", font}, {"N/A", font}}}).first->second;
-
-        object.label.setPosition({5.f, 5.f + (static_cast<float>(i + 4) * font.getLineSpacing(14))});
-        object.value.setPosition({80.f, 5.f + (static_cast<float>(i + 4) * font.getLineSpacing(14))});
+        auto [it, success]   = texts.emplace(axislabels[i], JoystickObject{{font, axislabels[i] + ":"}, {font, "N/A"}});
+        auto& [label, value] = it->second;
+        label.setPosition({5.f, 5.f + (static_cast<float>(i + 4) * font.getLineSpacing(14))});
+        value.setPosition({80.f, 5.f + (static_cast<float>(i + 4) * font.getLineSpacing(14))});
     }
 
     for (unsigned int i = 0; i < sf::Joystick::ButtonCount; ++i)
     {
         sstr.str("");
         sstr << "Button " << i;
-        auto& object = texts.insert({sstr.str(), {{sstr.str() + ":", font}, {"N/A", font}}}).first->second;
-
-        object.label.setPosition(
-            {5.f, 5.f + (static_cast<float>(sf::Joystick::AxisCount + i + 4) * font.getLineSpacing(14))});
-        object.value.setPosition(
-            {80.f, 5.f + (static_cast<float>(sf::Joystick::AxisCount + i + 4) * font.getLineSpacing(14))});
+        auto [it, success]   = texts.emplace(sstr.str(), JoystickObject{{font, sstr.str() + ":"}, {font, "N/A"}});
+        auto& [label, value] = it->second;
+        label.setPosition({5.f, 5.f + (static_cast<float>(sf::Joystick::AxisCount + i + 4) * font.getLineSpacing(14))});
+        value.setPosition({80.f, 5.f + (static_cast<float>(sf::Joystick::AxisCount + i + 4) * font.getLineSpacing(14))});
     }
 
     for (auto& [label, joystickObject] : texts)
@@ -176,8 +180,9 @@ int main()
                 for (auto& [label, joystickObject] : texts)
                     joystickObject.value.setString("N/A");
 
-                texts.at("ID").label.setString("<Not Connected>");
-                texts.at("ID").value.setString("");
+                auto& [label, value] = texts.at("ID");
+                label.setString("<Not Connected>");
+                value.setString("");
 
                 sstr.str("");
                 sstr << threshold << "  (Change with up/down arrow keys)";
--- a/examples/joystick/Joystick.cpp
+++ b/examples/joystick/Joystick.cpp
@@ -24,16 +24,16 @@ Texts              texts;
 std::ostringstream sstr;
 float              threshold = 0.1f;
 
-// Axes labels in as C strings
-constexpr std::array axislabels = {"X", "Y", "Z", "R", "U", "V", "PovX", "PovY"};
+// Axes labels in as strings
+const std::array<std::string, 8> axislabels = {"X", "Y", "Z", "R", "U", "V", "PovX", "PovY"};
 
 // Helper to set text entries to a specified value
 template <typename T>
-void set(const char* label, const T& value)
+void set(const std::string& label, const T& value)
 {
     sstr.str("");
     sstr << value;
-    texts[label].value.setString(sstr.str());
+    texts.at(label).value.setString(sstr.str());
 }
 
 // Update joystick identification
@@ -41,8 +41,8 @@ void updateIdentification(unsigned int index)
 {
     sstr.str("");
     sstr << "Joystick " << index << ":";
-    texts["ID"].label.setString(sstr.str());
-    texts["ID"].value.setString(sf::Joystick::getIdentification(index).name);
+    texts.at("ID").label.setString(sstr.str());
+    texts.at("ID").value.setString(sf::Joystick::getIdentification(index).name);
 }
 
 // Update joystick axes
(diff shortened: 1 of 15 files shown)


ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

Bromeon (commented):
Repeating my comment from [here](https://github.com/SFML/SFML/issues/2194#issuecomment-1237724949):

> A case we need to think about:
> 
> ```c++
> sf::Text text(font);
> addText(std::move(text)); // give ownership away
> 
> // now `text` is like a "default-constructed" instance -- what are its semantics?
> ```
> 
> Also, `sf::Font` relates to `sf::Text` the same way that `sf::Texture` relates to `sf::Sprite`, or `sf::SoundBuffer` to `sf::Sound`. What do you think about those resources?

This will force users to use `std::optional<sf::Text>` in classes that use text that cannot be immediately initialized -- which is probably a good thing to highlight nullability. But we might need to check user feedback here.

---

On a tangential note, in retrospective I think removing the default font was a user-unfriendly decision. Sure, it keeps the binary smaller, but it's quite a bit of extra effort to get some simple text or debug info displayed. It's probably in line though with seeing SFML as a more mid-level media abstraction, rather than a "batteries included" sort of game engine.

examples/joystick/Joystick.cpp
```diff
@@ -103,58 +103,43 @@ int main()
...
 
     // Set up our joystick identification sf::Text objects
-    texts["ID"].label.setPosition({5.f, 5.f});
-    texts["ID"].value.setPosition({80.f, 5.f});
-
-    texts["ID"].label.setString("<Not Connected>");
-    texts["ID"].value.setString("");
+    texts.emplace("ID", JoystickObject{{font, "<Not Connected>"}, {font}});
+    texts.at("ID").label.setPosition({5.f, 5.f});
+    texts.at("ID").value.setPosition({80.f, 5.f});
```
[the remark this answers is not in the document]
ChrisThrasher: The return type is pretty ugly. 

```cpp
    auto& id = texts.emplace("ID", JoystickObject{{font, "<Not Connected>"}, {font}}).first->second;
    id.label.setPosition({5.f, 5.f});
    id.value.setPosition({80.f, 5.f});
```
    
I don't expect anyone to understand what `.first->second` means, but I'll use this pattern instead if you prefer it.

ChrisThrasher:
> On a tangential note, in retrospective I think removing the default font was a user-unfriendly decision.

I'm also open to the idea of bringing back a default font should someone choose to lead that effort. It's very user-friendly to not require everyone load their own font. It's in line with how we vendor all dependencies instead of making users supply them.

ChrisThrasher:
Regarding moved-from types, that's a good point. We define `sf::Sound` to nullify the underlying `SoundBuffer` when copied which means it's valid for that class to hold a null pointer. It's not unreasonable to want a moved-from `Text` to have no font associated with it. If it's possible for a given `Text` to have its font removed from it, that implies that maybe it's also valid to allow for default construction. I think this ultimately comes down to what semantics we want for moving/copying `Text`. As it stands now, it's reasonable to have no default constructor but if we change its copy or move semantics in the future then we may want to put it back.

examples/joystick/Joystick.cpp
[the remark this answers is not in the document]
Bromeon: I wanted to suggest structured bindings, but then it came to my mind that there's no way to ignore a part of it.
Why is every feature always half-done :disappointed_relieved:

It's probably still better to not throw that information away instead of doing the lookup twice, not sure how others see this...

Bromeon:
> If it's possible for a given `Text` to have its font removed from it, that implies that maybe it's also valid to allow for default construction.

Exactly. We can basically define the "moved-from" state in two ways:

1. Safe to destroy, but don't touch anymore. That instance is poisoned.
2. Defined semantics, can be re-initialized. 

I'm also totally OK when going with (1), in that case we wouldn't need a default constructor as it's not part of the public API to reuse "empty" `sf::Text` instances. Are there ways to generate compiler/linter warnings upon use of moved-from custom classes?

examples/joystick/Joystick.cpp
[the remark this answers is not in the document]
ChrisThrasher:
```cpp
    auto& [id_key, id] = *texts.emplace("ID", JoystickObject{{font, "<Not Connected>"}, {font}}).first;
    id.label.setPosition({5.f, 5.f});
    id.value.setPosition({80.f, 5.f});
```
Here I'm destructuring the pair that was just emplaced. It's a bit awkward to pick variable names that won't conflict with the `.at("Threshold")` snippet a few lines down which will need the same treatment. I can't use a generic variable name like `key` without shadowing warnings breaking the build.

Because this is example code I don't feel strongly and will change it to one of these alternative options if you prefer. I now see we're already using the `.first->second` pattern elsewhere in this file so I'm less reluctant to add more of it.

ChrisThrasher:
> Are there ways to generate compiler/linter warnings upon use of moved-from custom classes?

I don't believe we can make any changes to the library to make such bugs easier to detect. It's squarely on users to not use moved-from objects.

> Exactly. We can basically define the "moved-from" state in two ways:
> 
> 1. Safe to destroy, but don't touch anymore. That instance is poisoned.
> 1. Defined semantics, can be re-initialized.

I'm inclined to stick to (1). As far as I'm aware, our intent with adding more move semantics is to elide copies for the sake of performance. It's worth bearing in mind that if we start with option (1), we can always make moved-from objects _more_ defined in the future, but if we promise that a given moved-from object is safe to use without reinitializing, we can't rescind that promise without breaking the API. (1) is the lazy option that gives us more leeway going forward.

vittorioromeo:
I'm generally in favour of this change. 

> 1. Safe to destroy, but don't touch anymore. That instance is poisoned.

I'm also in favour of this approach.

> Are there ways to generate compiler/linter warnings upon use of moved-from custom classes?

No, but we could track these things at run-time in debug mode. Would probably not be worth the extra code complexity added.

examples/joystick/Joystick.cpp
```diff
@@ -24,25 +24,25 @@ Texts              texts;
...
 
 // Update joystick identification
 void updateIdentification(unsigned int index)
 {
     sstr.str("");
     sstr << "Joystick " << index << ":";
-    texts["ID"].label.setString(sstr.str());
-    texts["ID"].value.setString(sf::Joystick::getIdentification(index).name);
+    texts.at("ID").label.setString(sstr.str());
+    texts.at("ID").value.setString(sf::Joystick::getIdentification(index).name);
```
vittorioromeo, line 45: Avoid a double lookup:

```suggestion
    auto& [label, value] = texts.at("ID");
    label.setString(sstr.str());
    value.setString(sf::Joystick::getIdentification(index).name);
```

examples/joystick/Joystick.cpp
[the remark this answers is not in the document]
vittorioromeo: +1 for destructuring, but I would do it like this for readability:

```cpp
auto [it, success] = texts.emplace("ID", JoystickObject{{font, "<Not Connected>"}, {font}});
auto& [label, value] = *it;
label.setPosition({5.f, 5.f}); 
value.setPosition({80.f, 5.f});
```

Or just this:

```cpp
auto [it, success] = texts.emplace("ID", JoystickObject{{font, "<Not Connected>"}, {font}});
it->label.setPosition({5.f, 5.f}); 
it->value.setPosition({80.f, 5.f});
```

examples/joystick/Joystick.cpp
```diff
@@ -103,58 +103,43 @@ int main()
...
     sstr << threshold << "  (Change with up/down arrow keys)";
 
-    texts["Threshold"].label.setPosition({5.f, 5.f + 2 * font.getLineSpacing(14)});
-    texts["Threshold"].value.setPosition({80.f, 5.f + 2 * font.getLineSpacing(14)});
-
-    texts["Threshold"].label.setString("Threshold:");
-    texts["Threshold"].value.setString(sstr.str());
+    texts.emplace("Threshold", JoystickObject{{font, "Threshold:"}, {font, sstr.str()}});
+    texts.at("Threshold").label.setPosition({5.f, 5.f + 2 * font.getLineSpacing(14)});
+    texts.at("Threshold").value.setPosition({80.f, 5.f + 2 * font.getLineSpacing(14)});
```
vittorioromeo, line 116: Ditto, apply same change here.

examples/joystick/Joystick.cpp
```diff
@@ -103,58 +103,43 @@ int main()
...
 
     // Set up our label-value sf::Text objects
     for (unsigned int i = 0; i < sf::Joystick::AxisCount; ++i)
     {
-        JoystickObject& object = texts[axislabels[i]];
+        auto& object = texts.insert({axislabels[i], {{font, axislabels[i] + ":"}, {font, "N/A"}}}).first->second;
```
vittorioromeo, line 121:
```suggestion
        auto [it, success] = texts.insert({axislabels[i], {{font, axislabels[i] + ":"}, {font, "N/A"}}});
        auto& object = *it;
```

vittorioromeo: Or use `it->` below.

examples/joystick/Joystick.cpp
```diff
@@ -103,58 +103,43 @@ int main()
...
         object.value.setPosition({80.f, 5.f + (static_cast<float>(i + 4) * font.getLineSpacing(14))});
-        object.value.setString("N/A");
     }
 
     for (unsigned int i = 0; i < sf::Joystick::ButtonCount; ++i)
     {
         sstr.str("");
         sstr << "Button " << i;
-        JoystickObject& object = texts[sstr.str()];
+        auto& object = texts.insert({sstr.str(), {{font, sstr.str() + ":"}, {font, "N/A"}}}).first->second;
```
vittorioromeo, line 131: Ditto.

examples/joystick/Joystick.cpp
```diff
@@ -191,13 +176,13 @@ int main()
                 for (auto& [label, joystickObject] : texts)
                     joystickObject.value.setString("N/A");
 
-                texts["ID"].label.setString("<Not Connected>");
-                texts["ID"].value.setString("");
+                texts.at("ID").label.setString("<Not Connected>");
+                texts.at("ID").value.setString("");
```
vittorioromeo, line 180: Ditto, avoid double lookup.

vittorioromeo (changes requested):
Some minor readability changes.

ChrisThrasher:

Thanks for the feedback. I used structured bindings to get rid of all those double lookups.

By the way, [here](https://github.com/ChrisThrasher/SFML/commit/9ee5d5a7643daade3d00694c46cf59a7d924af74) are some more places we can use structured bindings. Might submit a PR if others are eager to use structured bindings in more places.

ChrisThrasher:

vittorioromeo (approved):

Bromeon (approved):

PR #2489, opened 2023-04-04
Prefer fine-grained headers to module-level ones in examples

vittorioromeo:
A few reasons for this PR:

1. Promote header hygiene in users of SFML and newcomers to C++. Using fine-grained headers is good practice.

2. Avoid relying on transitive includes. Using fine-grained headers will help us catch regressions of that sort.

3. Tiny compilation time improvement for examples.


vittorioromeo:

ChrisThrasher:
What build time difference were you able to measure?

ChrisThrasher (commented):
> Avoid relying on transitive includes. Using fine-grained headers will help us catch regressions of that sort.

It's worth mentioning how the examples serve as real-world tests of our module-level headers. If we stop using those headers in here then regressions in Graphics.hpp and the like are prone to creep in.

I'm mostly indifferent to this PR but leaning towards a "no". I won't veto it though if others like it. I hope we don't start a trend of adding way more `#include` statements all over the library.

Bromeon:
I have a mixed feelings here -- on one hand I like that good practice is demonstrated, on the other it makes examples a bit more complex, and a user taking examples as a base for experimentation will need to do more micro-managing.

Do you know how good different IDEs are meanwhile at auto-completing includes? How much effort is involved, if one takes an example without a single `#include` statement, and wants to have all the headers like in this PR?

vittorioromeo:
> Examples really should focus on brevity and teaching the user something in as little space as possible. If they want to optimize their own build times at some point they will have to look at all their headers, not only ours. We shouldn't try to teach users too many things at the same time.

@binary1248: The main point of this PR is teaching users good practices -- i.e. only include what you need. Lack of proper header hygiene causes various problems in C++, not only regarding build performance. The small benefit in compilation is a (beneficial) side effect of this PR and not one of the main goals.

---

> I have a mixed feelings here -- on one hand I like that good practice is demonstrated, on the other it makes examples a bit more complex, and a user taking examples as a base for experimentation will need to do more micro-managing.

@Bromeon: It's a tradeoff between convenience and teaching good practices to newcomers to C++. I think the latter has a lot of value, especially because poor header hygiene is a very common problem.

---

> Do you know how good different IDEs are meanwhile at auto-completing includes?

@Bromeon: Not really, unfortunately. I usually do not rely on IDEs.

---

> How much effort is involved, if one takes an example without a single #include statement, and wants to have all the headers like in this PR?

@Bromeon: The way I did it was pretty manual but quick. Grep for `sf::`, add the headers myself, try to compile, if it fails, add what's missing, and repeat. I'm confident that IDEs can do a better job.

binary1248:
I am not opposed to this idea in general, just bear in mind if we go down this route, we should really fix up the broken `#include`s at the top of each page in our documentation. 😁

vittorioromeo:
Created https://github.com/SFML/SFML/issues/2490

binary1248:
I was referring more to the broken header line in a class page like: https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1RenderWindow.php

The page makes the reader think `#include <RenderWindow.hpp>` is the right path to include. It really won't be fun for them to figure out which path to include for a specific class if we were to force them to think this way.

vittorioromeo:

eXpl0it3r:
I have mixed feelings about this. I understand the reasoning of teach good practices, but I can't really justify huge include lists in a simple example. If we wanted to teach *really* good practices, we'd also had wrap everything in nice classes or at least have some modularity with different files, etc. which I think we can all agree, would go a bit far, since we're talking about examples here.

image

Given the more general lean towards no, I'd say let's keep the simple header includes.

PR #2492, opened 2023-04-05
Inline `sfml_add_external`

ChrisThrasher:
## Description

After https://github.com/SFML/SFML/pull/2359 there were no more uses of `sfml_add_external` except within `sfml_find_package`. That lets us combine the two functions and get rid of a fair bit of code. Fewer layers of indirection should aid readability too. This is one small step towards cleaning up SFML's dependency management story.


ChrisThrasher:

eXpl0it3r (approved):

PR #2494, opened 2023-04-05
Remove default `sf::Sprite` constructor

ChrisThrasher:
## Description

Continuation of the new precedent established in #2486.

### DIFF
--- a/src/SFML/Graphics/Sprite.cpp
+++ b/src/SFML/Graphics/Sprite.cpp
@@ -29,15 +29,12 @@
 #include <SFML/Graphics/Sprite.hpp>
 #include <SFML/Graphics/Texture.hpp>
 
+#include <cassert>
 #include <cstdlib>
 
 
 namespace sf
 {
-////////////////////////////////////////////////////////////
-Sprite::Sprite() = default;
-
-
 ////////////////////////////////////////////////////////////
 Sprite::Sprite(const Texture& texture)
 {
@@ -133,14 +130,13 @@ FloatRect Sprite::getGlobalBounds() const
 ////////////////////////////////////////////////////////////
 void Sprite::draw(RenderTarget& target, const RenderStates& states) const
 {
-    if (m_texture)
-    {
-        RenderStates statesCopy(states);
+    assert(m_texture);
 
-        statesCopy.transform *= getTransform();
-        statesCopy.texture = m_texture;
-        target.draw(m_vertices, 4, PrimitiveType::TriangleStrip, statesCopy);
-    }
+    RenderStates statesCopy(states);
+
+    statesCopy.transform *= getTransform();
+    statesCopy.texture = m_texture;
+    target.draw(m_vertices, 4, PrimitiveType::TriangleStrip, statesCopy);
 }
 
 
--- a/examples/cocoa/CocoaAppDelegate.mm
+++ b/examples/cocoa/CocoaAppDelegate.mm
@@ -37,7 +37,7 @@
 // Our PIMPL
 struct SFMLmainWindow
 {
-    SFMLmainWindow(sf::WindowHandle win) : renderWindow(win), text(font), background(sf::Color::Blue)
+    SFMLmainWindow(sf::WindowHandle win) : renderWindow(win), text(font), sprite(logo), background(sf::Color::Blue)
     {
         std::string resPath = [[[NSBundle mainBundle] resourcePath] tostdstring];
         if (!logo.loadFromFile(resPath + "/logo.png"))
@@ -45,7 +45,6 @@ struct SFMLmainWindow
 
         logo.setSmooth(true);
 
-        sprite.setTexture(logo, true);
         sf::FloatRect rect = sprite.getLocalBounds();
         sf::Vector2f  size(rect.width, rect.height);
         sprite.setOrigin(size / 2.f);
--- a/examples/shader/Shader.cpp
+++ b/examples/shader/Shader.cpp
@@ -31,7 +31,7 @@ public:
         // Load the texture and initialize the sprite
         if (!m_texture.loadFromFile("resources/background.jpg"))
             return false;
-        m_sprite.setTexture(m_texture);
+        m_sprite.emplace(m_texture);
 
         // Load the shader
         if (!m_shader.loadFromFile("resources/pixelate.frag", sf::Shader::Fragment))
@@ -50,13 +50,13 @@ public:
     {
         sf::RenderStates statesCopy(states);
         statesCopy.shader = &m_shader;
-        target.draw(m_sprite, statesCopy);
+        target.draw(*m_sprite, statesCopy);
     }
 
 private:
-    sf::Texture m_texture;
-    sf::Sprite  m_sprite;
-    sf::Shader  m_shader;
+    sf::Texture               m_texture;
+    std::optional<sf::Sprite> m_sprite;
+    sf::Shader                m_shader;
 };
 
 
@@ -199,8 +199,8 @@ public:
         m_entityTexture.setSmooth(true);
 
         // Initialize the background sprite
-        m_backgroundSprite.setTexture(m_backgroundTexture);
-        m_backgroundSprite.setPosition({135.f, 100.f});
+        m_backgroundSprite.emplace(m_backgroundTexture);
+        m_backgroundSprite->setPosition({135.f, 100.f});
 
         // Load the moving entities
         for (int i = 0; i < 6; ++i)
@@ -234,7 +234,7 @@ public:
 
         // Render the updated scene to the off-screen surface
         m_surface.clear(sf::Color::White);
-        m_surface.draw(m_backgroundSprite);
+        m_surface.draw(*m_backgroundSprite);
         for (const sf::Sprite& entity : m_entities)
             m_surface.draw(entity);
         m_surface.display();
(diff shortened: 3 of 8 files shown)


ChrisThrasher:
What other types may need to get this treatment? The pattern is that types that hold a raw pointer to a resource ought to not have a default constructor.

`sf::Sound` is a potential candidate for this change. The problem with `sf::Sound` is that when copied that pointer to the sound buffer goes null so to do this we'd have to change the copy semantics of `sf::Sound`. Maybe we remove copy semantics in favor of making it a move-only type to address this.

Bromeon:
> What other types may need to get this treatment? The pattern is that types that hold a raw pointer to a resource ought to not have a default constructor.

Possibly some of shapes?

_[CircleShape](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1CircleShape.php#aaebe705e7180cd55588eb19488af3af1) (float radius=0, std::size_t pointCount=30)
[RectangleShape](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1RectangleShape.php#a83a2be157ebee85c95ed491c3e78dd7c) (const [Vector2f](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1Vector2.php) &size=[Vector2f](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1Vector2.php)(0, 0))_

Are these meaningful defaults? Not necessarily.

But it's important that we're all fully aware that we _do_ restrict quite some use cases by not having things default-constructible, like:

* `std::array` / raw arrays
* `std::vector::resize`
* `std::map::operator[]`
* conditional initialization with `if/else`
* ...

---


> `sf::Sound` is a potential candidate for this change. The problem with `sf::Sound` is that when copied that pointer to the sound buffer goes null so to do this we'd have to change the copy semantics of `sf::Sound`.

It doesn't seem to reset to null?

https://github.com/SFML/SFML/blob/21af6fecf142882b08e1f2aa006bb406a3631e96/src/SFML/Audio/Sound.cpp#L50-L55
https://github.com/SFML/SFML/blob/21af6fecf142882b08e1f2aa006bb406a3631e96/src/SFML/Audio/Sound.cpp#L89-L102

This auto-attach/detaqch seems to me more like something where we could unify the behavior of resources. It can be surprising that sound auto-unlikes itself upon destruction, while others just run into UB. I understand it's probably because sounds run continuously on a separate thread and sounds need to be centrally managed, but still...

---

> Maybe we remove copy semantics in favor of making it a move-only type to address this.

That seems too restrictive to me. These "frontend" classes (sprite, sound, text) are relatively lightweight and I don't see why we should forbid copying.

ChrisThrasher:
> It doesn't seem to reset to null?

https://github.com/SFML/SFML/blob/21af6fecf142882b08e1f2aa006bb406a3631e96/src/SFML/Audio/Sound.cpp#L172

ChrisThrasher:
> > What other types may need to get this treatment? The pattern is that types that hold a raw pointer to a resource ought to not have a default constructor.
> Possibly some of shapes?
> 
> [CircleShape](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1CircleShape.php#aaebe705e7180cd55588eb19488af3af1) (float radius=0, std::size_t pointCount=30)
> [RectangleShape](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1RectangleShape.php#a83a2be157ebee85c95ed491c3e78dd7c) (const [Vector2f](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1Vector2.php) &size=[Vector2f](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1Vector2.php)(0, 0))
> 
> Are these meaningful defaults? Not necessarily.

If a type can only reach a given state through its default constructor, that default constructor maybe shouldn't exist. For example, a `std::vector` is default constructed with zero elements, but it's still valid to call `.clear()` to get back to that state later. Seems reasonable to me. 

However, an `sf::Sprite` can only have a null texture upon default construction and not through any calls to `setTexture`. The fact that `setTexture` takes a const reference implies that we don't want users to nullify that texture later. It implies to me that having a null texture was never an intentionally valid state but rather a quirk of development that was never second-guessed. This argument probably applies to `sf::Sound` but apparently details of OpenAL complicate things so more research will be required.

By that logic I think it's fine to let an `sf::CircleShape` or `sf::RectangleShape` default construct with a zero size because that's a state you can already achieve in the API. We wrote an API that says "zero size is valid" so it's also valid to have a default constructor that takes advantage of that possibility. 

TL;DR
* Forcing users to specify some initial elements would be annoying when an empty `std::vector` is already a permissible state to put a vector into.
* Forcing users to specify a texture when constructing an `sf::Sprite` makes sense because there is no way to nullify that texture later.
* Forcing users to specify a radius/size on a `sf::RectangleShape` or `sf::CircleShape` would be annoying when it's already valid to set those values to all zeros.

Bromeon:
> If a type can only reach a given state through its default constructor, that default constructor maybe shouldn't exist. For example, a `std::vector` is default constructed with zero elements, but it's still valid to call `.clear()` to get back to that state later. Seems reasonable to me.
> 
> However, an `sf::Sprite` can only have a null texture upon default construction and not through any calls to `setTexture`. The fact that `setTexture` takes a const reference implies that we don't want users to nullify that texture later.

This is a great observation, makes total sense 👍 

I'm not sure if it alone is a sufficient criterion though. `std::vector` being empty by default is _not only_ reasonable because one can clear it, but also because it's a very useful default. In a ton of scenarios, one adds only elements over time (like bullet entities in a game), and starting out empty is desired.

In contrast, a rectangle or circle with 0 extents is a degenerate case and almost never what someone wants -- even in the few hard-to-construct scenarios where someone intentionally needs this, it's always possible to explicitly provide such arguments.

To be clear, I'm not saying `sf::Shape` derivates cannot have default constructors. But these default constructors help mainly to get around technical limitations (e.g. using `std::array`), but are semantically useless.

src/SFML/Graphics/Sprite.cpp
```diff
@@ -133,14 +129,11 @@ FloatRect Sprite::getGlobalBounds() const
...
-        RenderStates statesCopy(states);
+    RenderStates statesCopy(states);
 
-        statesCopy.transform *= getTransform();
-        statesCopy.texture = m_texture;
-        target.draw(m_vertices, 4, PrimitiveType::TriangleStrip, statesCopy);
-    }
+    statesCopy.transform *= getTransform();
+    statesCopy.texture = m_texture;
+    target.draw(m_vertices, 4, PrimitiveType::TriangleStrip, statesCopy);
```
vittorioromeo, line 136: Should we `assert(m_texture)`?

ChrisThrasher: Under what circumstances could that pointer become null? I used to have that check in there then removed it because as far as I can tell it's impossible for it to get to that state.

vittorioromeo: > as far as I can tell it's impossible for it to get to that state

That seems a use case for `assert` :P

ChrisThrasher: `assert`s are so cheap that I won't protest its addition but I'm still really curious how under any circumstances other than `memcpy` shenanigans that pointer could become null.

ChrisThrasher:

src/SFML/Graphics/Sprite.cpp
```diff
@@ -133,14 +129,13 @@ FloatRect Sprite::getGlobalBounds() const
 ////////////////////////////////////////////////////////////
 void Sprite::draw(RenderTarget& target, const RenderStates& states) const
 {
-    if (m_texture)
-    {
-        RenderStates statesCopy(states);
+    assert(m_texture);
```
vittorioromeo, line 132: We need to include `<cassert>` if we want to keep this assert.

vittorioromeo (changes requested):

ChrisThrasher:

MarioLiebisch:
> In contrast, a rectangle or circle with 0 extents is a degenerate case and almost never what someone wants -- even in the few hard-to-construct scenarios where someone intentionally needs this, it's always possible to explicitly provide such arguments.

I think it would be reasonable to force an explicit radius in many cases (but not in all of them; as a member variable it would be annoying), but I'd also consider that even with the defaulted radius of 0 the object in itself is still valid. The other calls are not failing after all.

Also, using this approach, one could even question if it should be possible to default a (geometrical) vector to all zeroes, because a zero length vector would be a special edge case just like a zero radius circle.

Bromeon:
> I think it would be reasonable to force an explicit radius in many cases (but not in all of them; as a member variable it would be annoying)

Is "annoying" a good enough reason though? It's also annoying to be forced to initialized `sf::Text` in the constructor; there may not be a `sf::Font` available at the time. On the other hand, it gives you the guarantee that whenever you see a `sf::RectangleShape` object, it is already initialized.

> Also, using this approach, one could even question if it should be possible to default a (geometrical) vector to all zeroes, because a zero length vector would be a special edge case just like a zero radius circle.

Fully disagree, zero vectors are geometrically meaningful for a huge number of cases: coordinate origin, zero offset, zero velocity, etc. They are a _very_ reasonable default value, the same way that 0 is for `int`.

MarioLiebisch:
But just like a zero velocity make sense (even if there's no direction), wouldn't a zero radius circle make sense? A vector is defined by length and orientation, yet a zero vector lacks a clear orientation. A circle is defined by a radius (and number of segments, technically), so would a zero radius circle still be a circle? Maybe getting a bit too philosophical…😉

I think `sf::Sprite` is a completely different case. It's a visual representation of a part of a texture, so without a texture it doesn't make any sense, unless you want to do drawing in shaders only. So maybe even introduce a `sf::Texture::None` static for cases where the texture isn't existing?

ChrisThrasher:

Hapaxia:
> > It doesn't seem to reset to null?
> 
> https://github.com/SFML/SFML/blob/21af6fecf142882b08e1f2aa006bb406a3631e96/src/SFML/Audio/Sound.cpp#L172

This is removing the current buffer from the sound to which the new sound has been applied so that the new sound's buffer can replace it (since we're copying from that sound):
https://github.com/SFML/SFML/blob/21af6fecf142882b08e1f2aa006bb406a3631e96/src/SFML/Audio/Sound.cpp#L176-L177

vittorioromeo (approved):
LGTM

ChrisThrasher:

eXpl0it3r (approved):

PR #2498, opened 2023-04-06
Fixed wglGetProcAddress not providing OpenGL 1.1 functions when the context is provided by an Nvidia ICD

### LINKED ISSUE #2495: Access Violation within GLExtensions.cpp
## Access Violation within GLExtensions.cpp

Describe your issue here.
Whilst using SFML master branch with the CMake template project I encountered a crash on 100% of runs. Details below:

`Exception thrown at 0x0000000000000000 in CMakeSFMLProject.exe: 0xC0000005: Access violation executing location 0x0000000000000000.`

With a debugger attached the exception is thrown at: 
`GLExtensions.cpp` in `sf::ensureExtensionsInit`, when attempting to retreive the context version number.
Tracing 

Output to the console Window prior to this was:
```
Could not load necessary function to initialize OpenGL context
Could not load necessary function to initialize OpenGL context
Setting vertical sync not supported
```
Having debugged this a bit further I followed back to where the `GlContext` was being created and noticed that all calls to `GlContext::getFunction()` returned `nullptr`. Looking towards how the `WGlContext::getFunction` call (which is called by `GlContext::getFunction()`, it seems that `wglGetProcAddress` is always returning null. Further down in that function is a check for if the current context is generic. This flag is `false` when the crash happens, if forced to `true` (through a debugger) the crash doesn't occur.

## Development Environment
* Windows 11 64-bit
* Version 10.0.22621 Build 22621
* Master
* MSVC 17

## Hardware Info
* CPU : i7-12700KF
* GPU : Nvidia GeForce RTX 3070Ti
* RAM : 32GB 
* GPU Driver : Game Ready 531.41

## Steps to reproduce

Clone CMake template project (provided by SFML) then set it to use master branch, and change the constructor arguments for the `sf::RenderWindow` to be compatible.

```cpp
#include <SFML/Graphics.hpp>

int main()
{
	auto window = sf::RenderWindow{ sf::VideoMode{ {1920u, 1080u} }, "CMake SFML Project" };

	while (window.isOpen())
	{
		{
			{
				window.close();
			}
		}

		window.clear();
		window.display();
	}
}
```

### LINKED ISSUE #2497: "Could not load necessary function to initialize OpenGL context" after 21af6fec
## "Could not load necessary function to initialize OpenGL context" error on start, after 21af6fec

SFML-3 apps that used to work fine now fail to start after the change in the Windows OpenGL init code.

## Your environment

* Windows 10
* SFML 485d367, static libs
* VS 2022
* 
  ~~~~
  cl -MD -O2 -DNDEBUG -nologo -c -W4 -Iinclude -DSFML_STATIC -EHsc -std:c++latest demo.cpp
  link -out:demo.exe sfml-graphics-s.lib sfml-window-s.lib sfml-system-s.lib opengl32.lib freetype.lib user32.lib kernel32.lib gdi32.lib winmm.lib advapi32.lib demo.obj
  ~~~~

E.g.:

    #include <SFML/Graphics.hpp>
    int main()
    {
        sf::RenderWindow window(sf::VideoMode({800, 600}), "");
        return 0;
    }

Console output:

    Could not load necessary function to initialize OpenGL context
    Could not load necessary function to initialize OpenGL context
    Setting vertical sync not supported

_(Sorry if I'm missing something important, I'm late with other duties already, but wanted to still give you guys a quick heads-up, erring on the side of assuming this might be helpful as-is, nonetheless.)_

binary1248:
So it seems like Nvidia driver developers thought it would make sense not to provide OpenGL 1.1 functions via `wglGetProcAddress` and force us to use `GetProcAddress` instead for whatever reason.

Fixes #2495.
Fixes #2497.

### DIFF
--- a/src/SFML/Window/Win32/WglContext.cpp
+++ b/src/SFML/Window/Win32/WglContext.cpp
@@ -86,7 +86,7 @@ void ensureExtensionsInit(HDC deviceContext)
 
         // We don't check the return value since the extension
         // flags are cleared even if loading fails
-        gladLoadWGL(deviceContext, getOpenGl32Function);
+        gladLoadWGL(deviceContext, sf::priv::WglContext::getFunction);
     }
 }
 } // namespace WglContextImpl
@@ -200,22 +200,22 @@ GlFunctionPointer WglContext::getFunction(const char* name)
 {
     assert(WglContextImpl::currentContext != nullptr);
 
-    auto address = reinterpret_cast<GlFunctionPointer>(wglGetProcAddress(reinterpret_cast<LPCSTR>(name)));
-
-    if (address)
+    // If we are using the generic GDI implementation, skip to loading directly from OpenGL32.dll since it doesn't support extensions
+    if (!WglContextImpl::currentContext->m_isGeneric)
     {
-        // Test whether the returned value is a valid error code
-        auto errorCode = reinterpret_cast<ptrdiff_t>(address);
+        auto address = reinterpret_cast<GlFunctionPointer>(wglGetProcAddress(reinterpret_cast<LPCSTR>(name)));
 
-        if ((errorCode != -1) && (errorCode != 1) && (errorCode != 2) && (errorCode != 3))
-            return address;
+        if (address)
+        {
+            // Test whether the returned value is a valid error code
+            auto errorCode = reinterpret_cast<ptrdiff_t>(address);
+
+            if ((errorCode != -1) && (errorCode != 1) && (errorCode != 2) && (errorCode != 3))
+                return address;
+        }
     }
 
-    // If we are using the generic GDI implementation, try loading directly from OpenGL32.dll as well
-    if (WglContextImpl::currentContext->m_isGeneric)
-        return WglContextImpl::getOpenGl32Function(name);
-
-    return nullptr;
+    return WglContextImpl::getOpenGl32Function(name);
 }
 
 


binary1248:
@Bambo-Borris @xparq Care to test?

Bambo-Borris:
Will test this now & comment on how it goes

ChrisThrasher (approved):

Bambo-Borris:
@binary1248 This does indeed fix the crash, however I still see the following being output to the log:
```
Setting vertical sync not supported
```

However, I tested to see if vsync does enable and it appears to function correctly. I also noticed that if I create a default `sf::ContextSettings` then set any antialiasing level I will get the following output (where prior to #2478 I saw no such output).

```
Warning: The created OpenGL context does not fully meet the settings that were requested
Requested: version = 1.1 ; depth bits = 0 ; stencil bits = 0 ; AA level = 2 ; core = false ; debug = false ; sRGB = false
Created: version = 4.6 ; depth bits = 24 ; stencil bits = 0 ; AA level = 0 ; core = false ; debug = false ; sRGB = false
Setting vertical sync not supported
```

binary1248:
Updated.

@Bambo-Borris Can you test the current fix?

eXpl0it3r:
Could reproduce the issue and this seems to fix it. No error on my end with the latest version.

eXpl0it3r (approved):

xparq:
> @Bambo-Borris @xparq Care to test?

Sorry, had to leave & will be away from my work env for a while. Thx for the quick fix nonetheless!

Bambo-Borris:
Didn't get the chance to check this earlier but just like @eXpl0it3r I can confirm it fixes the crash & erroneous console output for me also.

xparq:
Late for the festivities, but for completeness also confirming the fix works on my end, too.

PR #2499, opened 2023-04-06
Allow for manually triggering CI

ChrisThrasher:
## Description

This is a useful thing to have. It lets us manually trigger CI when we want to observe something about it like build times or log messages since the logs for a given CI run expire after a certain amount of time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
 
 jobs:
   build:


ChrisThrasher:

Bromeon:
One thing to keep in mind is that [GitHub Action variables](https://docs.github.com/en/actions/learn-github-actions/variables) may have subtly different meanings depending on the event that triggered the workflow.

Not sure if it matters in our case though.

eXpl0it3r (approved):

binary1248:
This is actually something I was also going to do myself. Can we add a filter so that only pushes to master and 2.x.y and 3.x.y branches trigger a workflow? It is kind of excessive to build every PR twice just because we push a development branch to the main repository.

ChrisThrasher:
> One thing to keep in mind is that [GitHub Action variables](https://docs.github.com/en/actions/learn-github-actions/variables) may have subtly different meanings depending on the event that triggered the workflow.
> 
> Not sure if it matters in our case though.

I can't tell if it would matter? I know I've used this pattern on other GItHub projects for the better part of a year and so far I haven't noticed any issues but maybe that's just a coincidence.

> This is actually something I was also going to do myself. Can we add a filter so that only pushes to master and 2.x.y and 3.x.y branches trigger a workflow? It is kind of excessive to build every PR twice just because we push a development branch to the main repository.

Would this disable automatic CI runs on branch push in forks? That's a key part of my workflow. If it was not automatic then a lot of my un-PR'd branches would never get verified in CI. It would be a major chore to keep manually triggered CI until it passed just so that I could open a PR.

binary1248:
I myself wouldn't really mind clicking a button after every push. I do mind having to wait 20 mins longer because actions take double as long as they should if the branch is pushed to the main repository. If anyone has a creative solution to fix the double building problem I'm all for it.

Bromeon:
I haven't found a great solution for this either -- my usual workaround was to modify the `.yml` file while in development, and undo that when opening a PR.

There seem to be other workarounds here: https://github.com/orgs/community/discussions/26276

ChrisThrasher:
https://github.com/SFML/SFML/actions/runs/4646631652/jobs/8223281083#step:6:150

It didn't take long to find another reason I wish I could manually trigger CI on `master`. Homebrew seems to be having difficulties and I want to know whether the same error exists on `master` or if my PR is to blame. I can't figure that out since I cannot manually re-run CI on `master`.

binary1248 (approved):

PR #2500, opened 2023-04-06
Clean up namespace usage

ChrisThrasher:
## Description

I found a few places where we were not obeying our own rules for namespaces. For those unaware, we always omit `sf::` where possible, always use `std::` where possible, and always omit a class's namespace within the class.

### DIFF
--- a/src/SFML/Main/SFMLActivity.cpp
+++ b/src/SFML/Main/SFMLActivity.cpp
@@ -67,7 +67,7 @@ const char* getLibraryName(JNIEnv* lJNIEnv, jobject& objectActivityInfo)
     if (valueString == nullptr)
     {
         LOGE("No meta-data 'sfml.app.lib_name' found in AndroidManifest.xml file");
-        exit(1);
+        std::exit(1);
     }
 
     // Convert the application name to a C++ string and return it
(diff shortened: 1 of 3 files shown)


ChrisThrasher:

src/SFML/Main/SFMLActivity.cpp
```diff
@@ -217,7 +217,7 @@ JNIEXPORT void ANativeActivity_onCreate(ANativeActivity* activity, void* savedSt
     if (!ANativeActivity_onCreate)
     {
         LOGE("sfml-activity: Undefined symbol ANativeActivity_onCreate");
-        exit(1);
+        std::exit(1);
```
eXpl0it3r, line 220: Makes me wonder, after [recent discussion](https://github.com/SFML/SFML/pull/2458#issuecomment-1478830510) whether we should really be using `std::exit` 🤔

eXpl0it3r (approved):

src/SFML/Main/SFMLActivity.cpp
vittorioromeo: Definitely not, but at least this PR will make it easier to spot where we use it.

vittorioromeo (approved):

PR #2501, opened 2023-04-06
Remove unnecessary `std::string::c_str()` calls

ChrisThrasher:
## Description

Another random fixup I found while reading code a while ago. We're calling `std::string::c_str()` just to construct another `std::string` in a few places. In one spot I was able to replace it with a `std::filesystem` API call instead which is a lot cleaner.


ChrisThrasher:

eXpl0it3r (approved):
Ah nice! The string overload only exists since C++11.

PR #2503, opened 2023-04-08
Moved most variables in sfml-graphics, GlResource and GlContext out of namespace scope

binary1248:
Everybody knows the age-old explanation of why things might go wrong with some code: "don't create global SFML objects".

Due to historical reasons, it was very difficult (although not impossible), effort-wise to implement working context management without the use of objects living in namespace scope. With the additions from modern C++ this has become much easier.

Having objects living in namespace scope introduces construction ordering requirements between translation units which has never been guaranteed by C++, this essentially makes this code broken. Saying users shouldn't put their stuff in namespace scope is just a workaround for something we can and really should fix. While we still shouldn't recommend this practice, we have to be realistic, it is still going to happen, and we have better things to do than to play code police. Standard library objects support this usage without causing undefined behaviour and SFML classes should strive to be that robust as well.

The changes in this changeset also set up a better facility to manage other types of OpenGL resources that can't be shared between contexts. This might be necessary in the future.

This code would have broken in so many ways without these changes:
```cpp
#include <SFML/Graphics.hpp>

sf::RenderWindow  window(sf::VideoMode({1280, 720}), "Minimal, complete and verifiable example");
sf::Texture       texture;
sf::VertexBuffer  vbo;
sf::Shader        shader;

int main()
{
    window.setFramerateLimit(60);

    if (!texture.create(sf::Vector2u(1, 1)))
        return -1;

    if (!vbo.create(1))
        return -1;

    if (!shader.loadFromMemory("void main() {}", sf::Shader::Vertex))
        return -1;

    if (!renderTexture.create(sf::Vector2u(1, 1)))
        return -1;

    while (window.isOpen())
    {
        {
                window.close();
        }

        window.clear();
        window.display();
    }
}
```

### DIFF
--- a/include/SFML/Window/GlResource.hpp
+++ b/include/SFML/Window/GlResource.hpp
@@ -29,6 +29,8 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Window/Export.hpp>
 
+#include <memory>
+
 
 namespace sf
 {
@@ -48,23 +50,24 @@ protected:
     GlResource();
 
     ////////////////////////////////////////////////////////////
-    /// \brief Destructor
-    ///
-    ////////////////////////////////////////////////////////////
-    ~GlResource();
-
-    ////////////////////////////////////////////////////////////
-    /// \brief Register a function to be called when a context is destroyed
+    /// \brief Register an OpenGL object to be destroyed when its containing context is destroyed
     ///
     /// This is used for internal purposes in order to properly
     /// clean up OpenGL resources that cannot be shared between
     /// contexts.
     ///
-    /// \param callback Function to be called when a context is destroyed
-    /// \param arg      Argument to pass when calling the function
+    /// \param object Object to be destroyed when its containing context is destroyed
     ///
     ////////////////////////////////////////////////////////////
-    static void registerContextDestroyCallback(ContextDestroyCallback callback, void* arg);
+    static void registerUnsharedGlObject(std::shared_ptr<void> object);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Unregister an OpenGL object from its containing context
+    ///
+    /// \param object Object to be unregistered
+    ///
+    ////////////////////////////////////////////////////////////
+    static void unregisterUnsharedGlObject(std::shared_ptr<void> object);
 
     ////////////////////////////////////////////////////////////
     /// \brief RAII helper class to temporarily lock an available context for use
@@ -97,6 +100,12 @@ protected:
         ////////////////////////////////////////////////////////////
         TransientContextLock& operator=(const TransientContextLock&) = delete;
     };
+
+private:
+    ////////////////////////////////////////////////////////////
+    // Member data
+    ////////////////////////////////////////////////////////////
+    std::shared_ptr<void> m_sharedContext; //!< Shared context used to link all contexts together for resource sharing
 };
 
 } // namespace sf
--- a/src/SFML/Graphics/RenderTarget.cpp
+++ b/src/SFML/Graphics/RenderTarget.cpp
@@ -54,30 +54,35 @@ namespace
 namespace RenderTargetImpl
 {
 // Mutex to protect ID generation and our context-RenderTarget-map
-std::recursive_mutex mutex;
+std::recursive_mutex& getMutex()
+{
+    static std::recursive_mutex mutex;
+    return mutex;
+}
 
 // Unique identifier, used for identifying RenderTargets when
 // tracking the currently active RenderTarget within a given context
 std::uint64_t getUniqueId()
 {
-    const std::lock_guard lock(mutex);
-
-    static std::uint64_t id = 1; // start at 1, zero is "no RenderTarget"
-
+    const std::lock_guard lock(getMutex());
+    static std::uint64_t  id = 1; // start at 1, zero is "no RenderTarget"
     return id++;
 }
 
 // Map to help us detect whether a different RenderTarget
 // has been activated within a single context
 using ContextRenderTargetMap = std::unordered_map<std::uint64_t, std::uint64_t>;
-ContextRenderTargetMap contextRenderTargetMap;
+ContextRenderTargetMap& getContextRenderTargetMap()
+{
+    static ContextRenderTargetMap contextRenderTargetMap;
+    return contextRenderTargetMap;
+}
 
 // Check if a RenderTarget with the given ID is active in the current context
 bool isActive(std::uint64_t id)
 {
-    const auto it = contextRenderTargetMap.find(sf::Context::getActiveContextId());
-
-    return (it != contextRenderTargetMap.end()) && (it->second == id);
+    const auto it = getContextRenderTargetMap().find(sf::Context::getActiveContextId());
+    return (it != getContextRenderTargetMap().end()) && (it->second == id);
 }
 
 // Convert an sf::BlendMode::Factor constant to the corresponding OpenGL constant.
@@ -393,12 +398,13 @@ bool RenderTarget::setActive(bool active)
 {
     // Mark this RenderTarget as active or no longer active in the tracking map
     {
-        const std::lock_guard lock(RenderTargetImpl::mutex);
+        const std::lock_guard lock(RenderTargetImpl::getMutex());
 
         const std::uint64_t contextId = Context::getActiveContextId();
 
-        using RenderTargetImpl::contextRenderTargetMap;
-        const auto it = contextRenderTargetMap.find(contextId);
+        using RenderTargetImpl::getContextRenderTargetMap;
+        auto& contextRenderTargetMap = getContextRenderTargetMap();
+        auto  it                     = contextRenderTargetMap.find(contextId);
 
         if (active)
         {
--- a/src/SFML/Graphics/RenderTextureImplFBO.cpp
+++ b/src/SFML/Graphics/RenderTextureImplFBO.cpp
@@ -36,110 +36,37 @@
 
 #include <mutex>
 #include <ostream>
-#include <set>
-#include <unordered_map>
-#include <unordered_set>
-#include <utility>
-
-
-namespace
-{
-// Set to track all active FBO mappings
-// This is used to free active FBOs while their owning
-// RenderTextureImplFBO is still alive
-std::unordered_set<std::unordered_map<std::uint64_t, unsigned int>*> frameBuffers;
-
-// Set to track all stale FBOs
-// This is used to free stale FBOs after their owning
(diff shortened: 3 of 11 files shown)


ChrisThrasher (commented):
I tried creating a global `RenderWindow` in a separate project and observed a crash which goes away when using this PR so it seems to be working.

src/SFML/Graphics/RenderTextureImplFBO.cpp
```diff
@@ -153,15 +80,22 @@ RenderTextureImplFBO::~RenderTextureImplFBO()
...
-        staleFrameBuffers.emplace(contextId, frameBufferId);
+    // Unregister FBOs with the contexts if they haven't already been destroyed
+    for (auto& entry : m_frameBuffers)
+    {
+        auto frameBuffer = entry.second.lock();
 
-    for (auto& [contextId, multisampleFrameBufferId] : m_multisampleFrameBuffers)
-        staleFrameBuffers.emplace(contextId, multisampleFrameBufferId);
+        if (frameBuffer)
+            unregisterUnsharedGlObject(frameBuffer);
```
vittorioromeo, line 89:
```suggestion
            unregisterUnsharedGlObject(std::move(frameBuffer));
```

src/SFML/Graphics/RenderTextureImplFBO.cpp
```diff
@@ -153,15 +80,22 @@ RenderTextureImplFBO::~RenderTextureImplFBO()
...
+    }
+
+    for (auto& entry : m_multisampleFrameBuffers)
+    {
+        auto frameBuffer = entry.second.lock();
 
-    // Clean up FBOs
-    destroyStaleFBOs();
+        if (frameBuffer)
+            unregisterUnsharedGlObject(frameBuffer);
```
vittorioromeo, line 97:
```suggestion
            unregisterUnsharedGlObject(std::move(frameBuffer));
```

src/SFML/Graphics/RenderTextureImplFBO.cpp
```diff
@@ -460,33 +393,30 @@ bool RenderTextureImplFBO::createFrameBuffer()
...
-    {
-        std::lock_guard lock(mutex);
+    // Insert the FBO into our map
+    m_frameBuffers.emplace(Context::getActiveContextId(), frameBuffer);
 
-        // Insert the FBO into our map
-        m_frameBuffers.emplace(Context::getActiveContextId(), frameBuffer);
-    }
+    // Register the object with the currently context so it is automatically destroyed
+    registerUnsharedGlObject(frameBuffer);
```
vittorioromeo, line 404:
```suggestion
    registerUnsharedGlObject(std::move(frameBuffer));
```

src/SFML/Graphics/RenderTextureImplFBO.cpp
```diff
@@ -515,19 +445,17 @@ bool RenderTextureImplFBO::createFrameBuffer()
...
-        {
-            std::lock_guard lock(mutex);
+        // Insert the FBO into our map
+        m_multisampleFrameBuffers.emplace(Context::getActiveContextId(), multisampleFrameBuffer);
 
-            // Insert the FBO into our map
-            m_multisampleFrameBuffers.emplace(Context::getActiveContextId(), multisampleFrameBuffer);
-        }
+        // Register the object with the currently context so it is automatically destroyed
+        registerUnsharedGlObject(multisampleFrameBuffer);
```
vittorioromeo, line 458:
```suggestion
        registerUnsharedGlObject(std::move(multisampleFrameBuffer));
```

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -49,13 +49,18 @@ namespace
 namespace RenderTargetImpl
 {
 // Mutex to protect ID generation and our context-RenderTarget-map
-std::recursive_mutex mutex;
+std::recursive_mutex& getMutex()
+{
+    static std::recursive_mutex mutex;
+
+    return mutex;
```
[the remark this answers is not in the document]
ChrisThrasher: I wanted to say the same thing

src/SFML/Graphics/RenderTextureImplFBO.cpp
ChrisThrasher: Some recently clang-tidy changes I merged will automatically enforce this once Binary can rebase the PR.

src/SFML/Graphics/RenderTarget.cpp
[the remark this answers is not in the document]
binary1248: Newlines are always a coin toss for me... I have no personal preference really, but it would be nice to have a guideline somewhere since this isn't taken care of by tidy and not having to push newline corrections would save a lot of time.

include/SFML/Window/GlResource.hpp
```diff
@@ -48,23 +50,24 @@ class SFML_WINDOW_API GlResource
...
+    ///
+    /// This is used for internal purposes in order to properly
+    /// clean up OpenGL resources that cannot be shared bwteen
+    /// contexts.
+    ///
+    /// \param object Object to be destroyed when its containing context is destroyed
     ///
     ////////////////////////////////////////////////////////////
-    ~GlResource();
+    static void registerUnsharedGlObject(std::shared_ptr<void> object);
```
MarioLiebisch, line 62: `registerPrivateGlObject()`?

include/SFML/Window/GlResource.hpp
```diff
@@ -48,23 +50,24 @@ class SFML_WINDOW_API GlResource
...
-    /// clean up OpenGL resources that cannot be shared between
-    /// contexts.
-    ///
-    /// \param callback Function to be called when a context is destroyed
-    /// \param arg      Argument to pass when calling the function
+    /// \param object Object to be unregister
     ///
     ////////////////////////////////////////////////////////////
-    static void registerContextDestroyCallback(ContextDestroyCallback callback, void* arg);
+    static void unregisterUnsharedGlObject(std::shared_ptr<void> object);
```
MarioLiebisch, line 70: `unregisterPrivateGlObject`?

include/SFML/Window/GlResource.hpp
binary1248: Private is different from not being shared. Some OpenGL objects like FBOs can't be shared among different contexts. They can however still be accessed by anyone who knows which context to activate to use them, so if you want to argue from an access control perspective they are still public.

src/SFML/Window/GlContext.cpp
```diff
@@ -167,45 +167,162 @@ namespace
...
+    ////////////////////////////////////////////////////////////
+    static std::shared_ptr<SharedContext> get()
+    {
+        auto& weakSharedContext = getWeakPtr();
+        auto  sharedContext     = weakSharedContext.lock();
+
+        if (!sharedContext)
+        {
+            sharedContext = std::make_shared<GlContext::SharedContext>();
+
```
ChrisThrasher, line 241:
```suggestion
```

src/SFML/Window/GlContext.cpp
```diff
@@ -167,45 +167,162 @@ namespace
...
+    ////////////////////////////////////////////////////////////
+    /// \brief Get weak_ptr
+    ///
+    /// \return weak_ptr to the SharedContext
+    ///
+    ////////////////////////////////////////////////////////////
+    static std::weak_ptr<SharedContext>& getWeakPtr()
+    {
+        static std::weak_ptr<SharedContext> weakSharedContext;
+
```
ChrisThrasher, line 221:
```suggestion
```

src/SFML/Window/GlContext.cpp
```diff
@@ -167,45 +167,162 @@ namespace
...
 {
     std::uint64_t        id{};
     sf::priv::GlContext* ptr{};
     unsigned int         transientCount{};
+
+    // This per-thread variable holds the current context information for each thread
+    static CurrentContext& get()
+    {
+        thread_local CurrentContext currentContext;
+
```
ChrisThrasher, line 182:
```suggestion
```

src/SFML/Window/GlContext.cpp
```diff
@@ -260,176 +375,143 @@ struct TransientContext
...
+    /// This per-thread variable tracks if and how a transient
+    /// context is currently being used on the current thread
+    ///
+    /// \return The thread local TransientContext
+    ///
+    ////////////////////////////////////////////////////////////
+    static std::optional<TransientContext>& get()
+    {
+        thread_local std::optional<TransientContext> transientContext;
+
```
ChrisThrasher, line 390:
```suggestion
```

src/SFML/Window/GlContext.cpp
```diff
@@ -260,176 +375,143 @@ struct TransientContext
...
+    /// \brief Get mutex protecting unshared objects
+    ///
+    /// \return Mutex protecting unshared objects
+    ///
+    ////////////////////////////////////////////////////////////
+    static std::mutex& getUnsharedGlObjectsMutex()
     {
-        major = static_cast<unsigned int>(version[prefixLength] - '0');
-        minor = static_cast<unsigned int>(version[prefixLength + 2] - '0');
+        static std::mutex mutex;
```
ChrisThrasher, line 461:
```suggestion
```

ChrisThrasher (approved):
Beyond some nitpicks, this seems fine to me. A test from earlier proved to me that this does solve a real problem about not being able to use global instances of `sf::` types. It'll be good to remove that limitation. I want this merged sooner rather than later so our SFML 3 testers can use it since that's going to find any potential remaining issues much faster than code review will. A PR of this scale requires extensive testing that cannot happen until it's merged.

src/SFML/Window/GlContext.cpp
```diff
@@ -165,45 +165,162 @@ namespace
...
+    // We need to make sure that no operating system context
+    // or pixel format operations are performed simultaneously
+    // This mutex is also used to protect the shared context
+    // from being locked on multiple threads
+    std::recursive_mutex mutex;
 
-// The hidden, inactive context that will be shared with all other contexts
-std::unique_ptr<ContextType> sharedContext;
+    // Supported OpenGL extensions
+    std::vector<std::string> extensions;
```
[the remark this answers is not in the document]
binary1248: The specification is vague on the lifetime of the returned strings. I would not put it past some driver implementations to free up the memory at an unexpected time because they assume the strings are copied by the application.

binary1248:
Updated with suggestions.

eXpl0it3r (approved):

PR #2508, opened 2023-04-10
WglContext: Caching pixel formats for a massive speed boost in multiple context programs.

lucasz93:

----

## Description
Hello! My program creates lots of worker OpenGL contexts and it was taking 20+ seconds to start. The bottleneck was trying to find the best WGL pixel format. Since all my contexts use the same pixel format, it made sense to cache the pixel format. Creating 64 contexts could then happen in under 1 second.

## Tasks

* [ ] Tested on Linux
* [X] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/include/SFML/Window/ContextSettings.hpp
+++ b/include/SFML/Window/ContextSettings.hpp
@@ -26,6 +26,7 @@
 #define SFML_CONTEXTSETTINGS_HPP
 
 #include <SFML/Config.hpp>
+#include <tuple>
 
 namespace sf
 {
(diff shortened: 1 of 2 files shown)


ChrisThrasher:
If you rebase on `master` then most of the CI errors will go away

src/SFML/Window/Win32/WglContext.cpp
```diff
@@ -376,6 +386,11 @@ int WglContext::selectBestPixelFormat(HDC deviceContext, unsigned int bitsPerPix
                 }
             }
         }
+
+        formatCache.try_emplace(
+            std::make_tuple(bitsPerPixel, settings),
+            bestFormat
+        );
```
vittorioromeo, line 393: Why did you pick `try_emplace` compared to `emplace`? At this point, we should be sure that there is no such format in the cache. 

I would rewrite as:

```suggestion
        [[maybe_unused]] const auto [it, emplaced] = formatCache.emplace(
            std::make_tuple(bitsPerPixel, settings),
            bestFormat
        );
        
        assert(emplaced);
```

include/SFML/Window/ContextSettings.hpp
```diff
@@ -26,6 +26,7 @@
 #define SFML_CONTEXTSETTINGS_HPP
 
 #include <SFML/Config.hpp>
+#include <tuple>
```
vittorioromeo, line 29: `<tuple>` is a bit of an expensive header, which can increase compilation times a lot. I would prefer to not use it at all, or to (more simply) hide it in `ContextSettings.cpp`, so that it doesn't get virally included by users of `ContextSettings.hpp`.

vittorioromeo (changes requested):
Mostly code-quality--related comments, @binary1248 should take a look at the PR for a higher-level overview of the change.

binary1248:
Superseded by #2616.

PR #2513, opened 2023-04-12
Add clang-tidy `bugprone-*` checks

### LINKED ISSUE #2505: Enable clang-tidy `bugprone-*` checks
## Subject of the issue

https://clang.llvm.org/extra/clang-tidy/checks/list.html

These checks end up being a bit noisy so many of them have to be disabled. However it does include a lot of useful stuff so we're still better off enabling them.

ChrisThrasher:
## Description

Closes #2505. The `bugprone-*` family is very large and contains some opinionated checks we don't care about like `bugprone-assignment-in-if-condition` or `bugprone-easily-swappable-parameters` but it also contains a lot of very useful checks that we'll be happy to have included in our CI pipeline.

Here are the additional checks being added:

```
    bugprone-argument-comment
    bugprone-assert-side-effect
    bugprone-bad-signal-to-kill-thread
    bugprone-bool-pointer-implicit-conversion
    bugprone-copy-constructor-init
    bugprone-dangling-handle
    bugprone-dynamic-static-initializers
    bugprone-fold-init-type
    bugprone-forward-declaration-namespace
    bugprone-forwarding-reference-overload
    bugprone-inaccurate-erase
    bugprone-infinite-loop
    bugprone-lambda-function-name
    bugprone-macro-parentheses
    bugprone-macro-repeated-side-effects
    bugprone-misplaced-operator-in-strlen-in-alloc
    bugprone-misplaced-pointer-arithmetic-in-alloc
    bugprone-move-forwarding-reference
    bugprone-multiple-statement-macro
    bugprone-no-escape
    bugprone-not-null-terminated-result
    bugprone-parent-virtual-call
    bugprone-posix-return
    bugprone-redundant-branch-condition
    bugprone-reserved-identifier
    bugprone-shared-ptr-array-mismatch
    bugprone-signal-handler
    bugprone-sizeof-container
    bugprone-sizeof-expression
    bugprone-spuriously-wake-up-functions
    bugprone-standalone-empty
    bugprone-string-constructor
    bugprone-string-literal-with-embedded-nul
    bugprone-stringview-nullptr
    bugprone-suspicious-enum-usage
    bugprone-suspicious-memory-comparison
    bugprone-suspicious-memset-usage
    bugprone-suspicious-missing-comma
    bugprone-suspicious-realloc-usage
    bugprone-suspicious-semicolon
    bugprone-suspicious-string-compare
    bugprone-swapped-arguments
    bugprone-terminating-continue
    bugprone-throw-keyword-missing
    bugprone-too-small-loop-variable
    bugprone-undefined-memory-manipulation
    bugprone-undelegated-constructor
    bugprone-unhandled-exception-at-new
    bugprone-unhandled-self-assignment
    bugprone-unused-raii
    bugprone-unused-return-value
    bugprone-use-after-move
    bugprone-virtual-near-miss
```


eXpl0it3r (approved):
No expert in macro magic, but I assume the added parenthesis don't break anything, correct?

ChrisThrasher:
> No expert in macro magic, but I assume the added parenthesis don't break anything, correct?

https://clang.llvm.org/extra/clang-tidy/checks/bugprone/macro-parentheses.html

Here's the check that recommended we add braces. It just makes us doubly sure that our macro arguments aren't getting interpreted in the wrong way after expansion. Because we use macros in rather simple ways I don't expect the code to behave any differently now.

binary1248 (approved):

PR #2516, opened 2023-04-13
Use `std::byte`

ChrisThrasher:
## Description

I've been rebasing this commit for ages so I figured I'd submit a PR finally. I tried to find all places we could easily drop in `std::byte`.  I'm keeping it as a draft for now until I'm more confident we've found all the places we should be using `std::byte`

### DIFF
--- a/include/SFML/Network/Packet.hpp
+++ b/include/SFML/Network/Packet.hpp
@@ -432,10 +432,10 @@ private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    std::vector<char> m_data;          //!< Data stored in the packet
-    std::size_t       m_readPos{};     //!< Current reading position in the packet
-    std::size_t       m_sendPos{};     //!< Current send position in the packet (for handling partial sends)
-    bool              m_isValid{true}; //!< Reading state of the packet
+    std::vector<std::byte> m_data;          //!< Data stored in the packet
+    std::size_t            m_readPos{};     //!< Current reading position in the packet
+    std::size_t            m_sendPos{};     //!< Current send position in the packet (for handling partial sends)
+    bool                   m_isValid{true}; //!< Reading state of the packet
 };
 
 } // namespace sf
--- a/include/SFML/Network/TcpSocket.hpp
+++ b/include/SFML/Network/TcpSocket.hpp
@@ -35,6 +35,8 @@
 
 #include <optional>
 
+#include <cstddef>
+
 
 namespace sf
 {
@@ -219,16 +221,16 @@ private:
     ////////////////////////////////////////////////////////////
     struct PendingPacket
     {
-        std::uint32_t     Size{};         //!< Data of packet size
-        std::size_t       SizeReceived{}; //!< Number of size bytes received so far
-        std::vector<char> Data;           //!< Data of the packet
+        std::uint32_t          Size{};         //!< Data of packet size
+        std::size_t            SizeReceived{}; //!< Number of size bytes received so far
+        std::vector<std::byte> Data;           //!< Data of the packet
     };
 
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    PendingPacket     m_pendingPacket;     //!< Temporary data of the packet currently being received
-    std::vector<char> m_blockToSendBuffer; //!< Buffer used to prepare data being sent from the socket
+    PendingPacket          m_pendingPacket;     //!< Temporary data of the packet currently being received
+    std::vector<std::byte> m_blockToSendBuffer; //!< Buffer used to prepare data being sent from the socket
 };
 
 } // namespace sf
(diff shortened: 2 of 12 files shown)


ChrisThrasher:
@kimci86 I think you found an issue with this a while ago but I forgot what it was.

Bromeon (dismissed):
While I agree with using `std::byte`, I'm not sure if this PR should simultaneously change the code style from `static_cast` to function-style casts.

That change is controversial because function-style casts provide less type safety, and we already use another strange idiom `int{value}` as of https://github.com/SFML/SFML/pull/2233. I don't think adding a 3rd convention helps overall consistency.

`static_cast` is admittedly the ugliest of all ways, and to me the cleanest solution to that problem would still be dedicated functions `toByte()`, `toUint32()`, etc. It's even shorter than function-style casts (no `std::` needed), while otherwise providing the same syntax without the "crowbar" behavior of the latter.

vittorioromeo (changes requested):

ChrisThrasher:
> Would `std::byte` give us any concrete advantages beyond "being the correct type" and avoiding interpretation as `char`?

The fact that `std::byte` correctly models the behavior we want is the whole argument. Because it's more restrictive it makes it harder to misuse. I'd also it's easier to read but "readability" arguments have been a major source of disagreement between us two so I won't harp too much on that. If that's not a convincing enough argument to the team then @eXpl0it3r ought to close this before we put any more time into it.

Bromeon:
I think it's worth modernizing.

But I could be wrong and real-world usage of SFML shows that `std::byte` makes things more complicated than it helps. In such cases, we'd need to act based on user feedback to also potentially revert things, but I'm not too worried about that 🙂

kimci86:
> @kimci86 I think you found an issue with this a while ago but I forgot what it was.

The trap I noticed when experimenting with span in #2027 was with this kind of statement:
```c++
value = static_cast<std::int16_t>(bytes[0] | (bytes[1] << 8));
```
because there is no integer promotion with `std::byte`. For all `b` of type `std::byte`, `b << 8` is `std::byte{0}`. I used `std::to_integer` function to solve it.
There is this problem in `SoundFileReaderWav.cpp`.

ChrisThrasher:
> because there is no integer promotion with `std::byte`. For all `b` of type `std::byte`, `b << 8` is `std::byte{0}`. I used `std::to_integer` function to solve it.

Ah yes that's what I wanted to be reminded of. I think we need a helper function to abstract over this problem of converting a few bytes into a larger integer since it's pretty ugly to manually write out all the code to do that. I can mull on that and try to come up with something palatable and more expressive.

Alternatively this can be seen as another downside of `std::byte` if it requires this extra machinery to do basic things. Depends whether you prefer terse code with invisible semantics or verbose code with explicit semantics.

In either case this highlights the need to test these audio encoding and decoding utilities.

ChrisThrasher:
```cpp
// Convert byte sequence into integer
// { 0x12, 0x34, 0x56 } == 0x563412
template <typename IntegerType, typename... Bytes>
[[nodiscard]] IntegerType toInteger(const Bytes&... bytes)
{
    static constexpr std::size_t numBytes = sizeof...(Bytes);
    static_assert(sizeof(IntegerType) >= numBytes);

    const std::common_type_t<Bytes...> byteSequence[numBytes] = {bytes...};

    IntegerType integer = 0;
    for (std::size_t offset = 0; offset < numBytes; ++offset)
        integer |= static_cast<IntegerType>(byteSequence[offset]) << 8 * offset;
    return integer;
}
```

I added a new function that fixes our issues with converting sequences of `std::byte` into integer types. I was able to use parameter packs to make this function ergonomic to use. Just list out as many bytes as you want to shove into an integer. Along with fixing our `std::byte` issues, I wrote this to be useful in other places like [here](https://github.com/SFML/SFML/blob/92dbac623bfa1880520d46256172813233c87e65/src/SFML/Window/Win32/CursorImpl.cpp#L87) where we're constructing integers from bytes. There are a few other places in SFML we could use this pattern so we no longer have to write out so much bit-shifting code.

ChrisThrasher:

ChrisThrasher:
@kimci86 Found some great ways to simplify the implementation of `toInteger` through better use of parameter packs. I feel good enough about this PR to open it for review.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

Bromeon (commented):
`toInteger` is indeed a cool idea!

I'm not sure about the name (it's more composing using bitwise OR rather than conversion), but I also don't have a better one, so that's OK🙂 

Should it maybe be in a `sf::priv` namespace? Or is that header not part of public headers?

eXpl0it3r (approved):

ChrisThrasher:
Made a tiny tweak to the function docs and rebased on master.

I also wanted to mention that we may use `std::byte` more in the future, particular where we use `void*` parameters but that will happen later once I have the time to take a deeper look into what that would entail.

ChrisThrasher:
Removed unused header

PR #2451, opened 2023-03-15
Adding support for std::string_view to sf::String

khatharr:
## Description

Revision of https://github.com/SFML/SFML/pull/2447

I went ahead and cleaned up formatting and added comments, just to get it done.

This PR is related to the issue #2445

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?
Test driver (same as before):

```cpp
#include <SFML/Main.hpp>
#include <SFML/Graphics.hpp>
#include <SFML/Window.hpp>

#include <string_view>

int main() {
  sf::RenderWindow window(sf::VideoMode(sf::Vector2u(800, 600)), "SFML Testbed");
  window.setFramerateLimit(60);

  sf::Font font;
  bool derp = font.loadFromFile("c:/Windows/Fonts/arial.ttf");
  
  const char data[] = "derpdurr";
  const wchar_t wdata[] = L"herpboing";

  std::string str(data + 4, 4);
  std::string_view sv(data, 4);
  std::wstring wstr(wdata, 4);
  std::wstring_view wsv(wdata + 4);

  std::vector<sf::Text> texts;
  texts.emplace_back(wstr, font);
  texts.back().setPosition(sf::Vector2f(10, 0));
  texts.emplace_back(sv, font);
  texts.back().setPosition(sf::Vector2f(10, 40));
  texts.emplace_back(str, font);
  texts.back().setPosition(sf::Vector2f(10, 80));
  texts.emplace_back(wsv, font);
  texts.back().setPosition(sf::Vector2f(10, 120));
  
  for(bool run = true; run;) {
    window.clear();
    for(auto& text : texts) {
      window.draw(text);
    }
    window.display();

    sf::Event event;
    while(window.pollEvent(event)) {
      if(event.type == sf::Event::Closed) {
        run = false; break;
      }
    }
  }
}
```

PR #2517, opened 2023-04-13
Support for `std::string_view` in `sf::String`

### LINKED ISSUE #2445: Support for std::string_view in sf::String?
Noticed today when trying to load an sf::Text that when a function requires an sf::String argument, std::string_view will not work, since there's no conversion. I'm not sure what the status is in terms of what C++ version most users are on, but it would be handy to have an overload for sf::String to accept std::string_view as an initialization argument. I reckon it could be segregated into a preprocessor block that checks for support before including it.

This would imply overloads for
- String::String(const std::string& ansiString, const std::locale& locale)
- String::String(const wchar_t* wideString)
- String::String(const std::basic_string<Uint32>& utf32String)

...and the inclusion of <string_view> in the header.

Thinking about it for a minute, the only potential issue that jumps out at me is that it would require some code repetition. This could potentially be avoided by having the acting overload accept std::string_view (or variant) as the relevant argument, then having the overloads simply forward their arguments into that function, since std::string_view will convert from the relevant types. If compatibility is a concern, then the preprocessor checks could just be in the header file, blocking the inclusion of <string_view> and the declaration of the overloads that use it in that scope, effectively making it an internal implementation detail, although that could potentially be an issue for users who compile the library themselves.

I'm not sure what the posture of the project is towards this kind of thing, but I figured I'd make the suggestion.

Cheers.

khatharr:
## Description

Revision of https://github.com/SFML/SFML/pull/2451

This is in a new branch and up to date with the origin's master as of this post.

Closes #2445

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?
Test driver (modified to match changes in sf::Text):

```cpp
#include <SFML/Main.hpp>
#include <SFML/Graphics.hpp>
#include <SFML/Window.hpp>

#include <string_view>

int main() {
  sf::RenderWindow window(sf::VideoMode(sf::Vector2u(800, 600)), "SFML Testbed");
  window.setFramerateLimit(60);

  sf::Font font;
  bool derp = font.loadFromFile("c:/Windows/Fonts/arial.ttf");
  
  const char data[] = "derpdurr";
  const wchar_t wdata[] = L"herpboing";

  std::string str(data + 4, 4);
  std::string_view sv(data, 4);
  std::wstring wstr(wdata, 4);
  std::wstring_view wsv(wdata + 4);

  std::vector<sf::Text> texts;
  texts.emplace_back(font, wstr);
  texts.back().setPosition(sf::Vector2f(10, 0));
  texts.emplace_back(font, sv);
  texts.back().setPosition(sf::Vector2f(10, 40));
  texts.emplace_back(font, str);
  texts.back().setPosition(sf::Vector2f(10, 80));
  texts.emplace_back(font, wsv);
  texts.back().setPosition(sf::Vector2f(10, 120));
  
  for(bool run = true; run;) {
    window.clear();
    for(auto& text : texts) {
      window.draw(text);
    }
    window.display();

    sf::Event event;
      if(event.type == sf::Event::Closed) { run = false; break; }
    }
  }
}
```

### DIFF
--- a/src/SFML/System/String.cpp
+++ b/src/SFML/System/String.cpp
@@ -173,7 +173,13 @@ String::String(const char* ansiString, const std::locale& locale)
 
 
 ////////////////////////////////////////////////////////////
-String::String(const std::string& ansiString, const std::locale& locale)
+String::String(const std::string& ansiString, const std::locale& locale) : String(std::string_view(ansiString), locale)
+{
+}
+
+
+////////////////////////////////////////////////////////////
+String::String(std::string_view ansiString, const std::locale& locale)
 {
     m_string.reserve(ansiString.length() + 1);
     Utf32::fromAnsi(ansiString.begin(), ansiString.end(), std::back_inserter(m_string), locale);
@@ -196,7 +202,13 @@ String::String(const wchar_t* wideString)
 
 
 ////////////////////////////////////////////////////////////
-String::String(const std::wstring& wideString)
+String::String(const std::wstring& wideString) : String(std::wstring_view(wideString))
+{
+}
+
+
+////////////////////////////////////////////////////////////
+String::String(std::wstring_view wideString)
 {
     m_string.reserve(wideString.length() + 1);
     Utf32::fromWide(wideString.begin(), wideString.end(), std::back_inserter(m_string));
@@ -211,6 +223,12 @@ String::String(const char32_t* utf32String) : m_string(utf32String ? utf32String
 
 ////////////////////////////////////////////////////////////
 String::String(std::u32string utf32String) : m_string(std::move(utf32String))
+{
+}
+
+
+////////////////////////////////////////////////////////////
+String::String(std::u32string_view utf32String) : m_string(utf32String)
 {
 }
 
--- a/include/SFML/System/String.hpp
+++ b/include/SFML/System/String.hpp
@@ -33,6 +33,7 @@
 
 #include <locale>
 #include <string>
+#include <string_view>
 
 #include <cstddef>
 #include <cstdint>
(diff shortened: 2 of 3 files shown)


khatharr:
I'll be more careful in the future.
Thank you for your patience.

src/SFML/System/String.cpp
```diff
@@ -118,7 +129,13 @@ String::String(const char32_t* utf32String)
 
 
 ////////////////////////////////////////////////////////////
-String::String(const std::u32string& utf32String) : m_string(utf32String)
+String::String(std::u32string_view utf32String) : m_string(utf32String)
```
ChrisThrasher, line 132: What about the `u16string` overloads?

Bromeon (commented):
Thanks for the PR!

I know it's for consistency, but the API surface for `wchar_t`, etc. is getting quite big, for it being mainly a Windows oddity that surfaces when dealing with legacy APIs. `u16string` is similarly rare.

Are types like `std::wstring_view` actually being used?

khatharr:
The current implementation in origin:master has constructors for
string/wstring/u32string. I added string-view/wstring_view/u32string_view
to match those. I didn't see anything relevant for u16.

Bromeon:
> I always thought it was used for large character sets like Hangul or Mandarin.

You probably mean Unicode in general, but most modern code uses either UTF-8 or UTF-32 to represent them, which can be encoded in various types in C++ (e.g. `char`, `std::uint8_t`, `std::char8_t` for UTF-8).

`wchar_t` on the other hand is a legacy type, mostly to deal with UCS-2/UTF-16 encoding at the time. See also [StackOverflow](https://stackoverflow.com/q/13509733) or [Wikipedia](https://en.wikipedia.org/wiki/Wide_character). You generally don't see modern APIs use it except for interop; newer proramming languages don't even have such a type.

I'm not saying `sf::String` should not provide interop for those, but I think their usefulness is overstated. And in my experience, these legacy APIs tend to be low-level (i.e. pointer/size pairs rather than C++17 `wstring_view`). But maybe it's worth providing just for consistency.

WarlockD:
std::string_view is used a lot in C++20 projects now.  Its just very lightweight and gets over the stigma of copying a std::string.  However the specs state the size() is the size of the pointer array. That is its just a `{const <T> *, int size}` wrapper so if you want to know the length of the string for the purposes of formatting you have to reiterate over the size rather than it being stored.  I guess it doesn't really mater when using char32_t but I have had quite a few projects where people use std::string or std::string_view in with unicode and just expect it to format without any pre processing.  

Don't get me started on how idiots are using std::span:P

ChrisThrasher:

PR #2518, opened 2023-04-13
Add minimum and maximum window size implementation

dogunbound:
## Description

My PR is to allow users to set the minimum and maximum size. Right now it only implements it for X11, but the plan is to do this for all other OS.

Related actions:
https://github.com/SFML/SFML/pull/2378
https://github.com/SFML/SFML/issues/2124

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Just run and compile this code bellow and try and the window to be bigger or smaller than the defined sizes.
```cpp
#include <SFML/Graphics.hpp>
#include <SFML/Window.hpp>

int main() {
  window.setMinimumSize(sf::Vector2u(200, 200));
  window.setMaximumSize(sf::Vector2u(800, 800));

  while (window.isOpen()) {
    sf::Event event;
        window.close();
    }

    window.clear();
    window.display();
  }

  return 0;
}
```

### DIFF
--- a/include/SFML/Window/WindowBase.hpp
+++ b/include/SFML/Window/WindowBase.hpp
@@ -244,6 +244,22 @@ public:
     void setSize(const Vector2u& size);
 
     ////////////////////////////////////////////////////////////
+    /// \brief Sets a minimum window rendering region size
+    ///
+    /// \param minimumSize New minimum size, in pixels
+    ///
+    ////////////////////////////////////////////////////////////
+    void setMinimumSize(const Vector2u& minimumSize);
+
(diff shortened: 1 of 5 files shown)


dogunbound:
I believe iOS, android, and linux DRM can use stub operations. Please correct me if I'm wrong. If I'm right, just let me know and I'll fix this PR to have iOS, android, and linux DRM implemented as stub operations.

dogunbound:
Was told that this would be done on master, not 2.6.x

PR #2519, opened 2023-04-13
Add `sf::WindowBase::set{Min|Max}imumSize()`

dogunbound:
## Description

My PR is to allow users to set the minimum and maximum sizes. ~~Right now it only implements it for X11, but the plan is to do this for all other OS.~~

MAINTAINER EDIT: This PR is implemented for all relevant OSes now.

Related to https://github.com/SFML/SFML/pull/2378
Closes https://github.com/SFML/SFML/issues/2124

## How to test this PR?

https://github.com/ChrisThrasher/sfml-test-2519

Use this test project.


dogunbound:
I believe iOS, android, and linux DRM can use stub operations. Please correct me if I'm wrong. If I'm right, just let me know and I'll fix this PR to have iOS, android, and linux DRM implemented as stub operations.

As for Windows, and Mac, it would be GREATLY appreciated someone could look into implementing it. That same would be true if iOS, android, and linux DRM aren't dummy operations as I have not experience developing on any other platform aside from linux.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -825,6 +825,12 @@ Vector2u WindowImplX11::getSize() const
 ////////////////////////////////////////////////////////////
 void WindowImplX11::setSize(const Vector2u& size)
 {
+    // We could set the min and max to the new size in its respective dimension if setSize violates 
+    // previously set min and max sizes
+    assert(size.x > m_minimumSize.x);
+    assert(size.y > m_minimumSize.y);
+    assert(size.x < m_maximumSize.x);
+    assert(size.y < m_maximumSize.y);
```
vittorioromeo, line 833:
```suggestion
    assert(size.x >= m_minimumSize.x);
    assert(size.y >= m_minimumSize.y);
    assert(size.x <= m_maximumSize.x);
    assert(size.y <= m_maximumSize.y);
```

Also, do we really want to `assert` this? The user could call `setSize` from the public interface of `sf::Window`.

If we want to `assert` and not return a `bool`, we should add something like: *"The behavior is undefined if this function is called with a `size` that does not fit in the previously specified minimum and maximum size boundaries. If such boundaries have never been previously specified, any value of `size` is valid."* to the documentation of `Window::setSize`.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -840,6 +846,57 @@ void WindowImplX11::setSize(const Vector2u& size)
     XFlush(m_display);
 }
 
+////////////////////////////////////////////////////////////
+void WindowImplX11::setMinimumSize(const Vector2u &minimumSize)
+{
+    assert(minimumSize.x < m_maximumSize.x);
+    assert(minimumSize.y < m_maximumSize.y);
```
vittorioromeo, line 853: Should this be `<=`?

```suggestion
    assert(minimumSize.x <= m_maximumSize.x);
    assert(minimumSize.y <= m_maximumSize.y);
```

Ditto for the use of `assert`. It means "this should never happen". Do we want these conditions to be documented undefined behavior, or do we want to return a `bool` instead to signal failure?

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -840,6 +846,57 @@ void WindowImplX11::setSize(const Vector2u& size)
...
+        XSetWMNormalHints(m_display, m_window, sizeHints);
+        XFree(sizeHints);
+    }
+}
+
+////////////////////////////////////////////////////////////
+void WindowImplX11::setMaximumSize(const Vector2u &maximumSize)
+{
+    assert(maximumSize.x > m_minimumSize.x);
+    assert(maximumSize.y > m_minimumSize.y);
```
vittorioromeo, line 879: As above.

vittorioromeo (dismissed):

src/SFML/Window/Unix/WindowImplX11.cpp
dogunbound: no. If the minimum size and maximum size are the same, that is still a valid state. It's basically forcing the window not to resize. in the respective dimension.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -840,6 +846,57 @@ void WindowImplX11::setSize(const Vector2u& size)
...
+        m_maximumSize = maximumSize;
+        XSizeHints* sizeHints = XAllocSizeHints();
+        sizeHints->flags = PMaxSize | PMinSize;
+        sizeHints->max_width = static_cast<int>(maximumSize.x);
+        sizeHints->max_height = static_cast<int>(maximumSize.y);
+        sizeHints->min_width = static_cast<int>(m_minimumSize.x);
+        sizeHints->min_height = static_cast<int>(m_minimumSize.y);
+        XSetWMNormalHints(m_display, m_window, sizeHints);
+        XFree(sizeHints);
+    }
```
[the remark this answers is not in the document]
dogunbound: maybe, but not yet. Every assert is me signifying undefined behavior that we need to figure out the best way of handling. What should we do if the user decides to set a min window size despite not using a resizable window style.

src/SFML/Window/Unix/WindowImplX11.cpp
dogunbound: This is lacking implementation. I do not think we should assert either, but we need to define what we want here. There are a few options:

1. Ignore the setSize call
2. Ignore the setSize call for the respective dimension effectively allowing the set size to work if either the new width or height is valid but only for the respective dimension.
3. of course we could still assert and crash the program
4. set a new minimum and maximum size in each respective dimension if the new setSize function violates the current minimum / maximum size

I think 4 is the right call, but do let me know! And it shouldn't be `<=`. We should allow users to perfectly butt up against the limits if they so choose to.

dogunbound: @vittorioromeo Anyways, what about the other options i detailed here https://github.com/SFML/SFML/pull/2519#discussion_r1165808282

src/SFML/Window/Unix/WindowImplX11.cpp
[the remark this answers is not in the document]
dogunbound: What we could do here:

1. Do nothing if the window style is set to no resize
2. Allow a resize to occur to the min size if the setMinimumSize is called and do the same for the set maximum method.
3. Assert

I think the 2nd option is what I'd go with.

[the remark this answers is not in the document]
dogunbound: Also, they aren't identical

> Consider extracting this mostly-duplicated code to a private function, or -- even better -- a function in an anonymous namespace.

They aren't identical. I don't see how you could abstract this.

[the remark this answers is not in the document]
kimci86: >They aren't identical.

They can be identical if you use member variables `m_minimumSize` and `m_maximumSize` to setup the `XSizeHints` instance instead of arguments.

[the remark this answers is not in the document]
dogunbound: > > They aren't identical.
> 
> They can be identical if you use member variables `m_minimumSize` and `m_maximumSize` to setup the `XSizeHints` instance instead of arguments.

Oh huh. yeah. ill do that then. thanks.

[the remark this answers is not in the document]
vittorioromeo: > > Consider extracting this mostly-duplicated code to a private function, or -- even better -- a function in an anonymous namespace.
> 
> They aren't identical. I don't see how you could abstract this with a function.

The parts that differ can be passed as parameters.

```cpp
namespace {

void updateSizeHints(const Vector2u& mins, const Vector2u& maxs, ::Display* display, ::Window window)
{
    XSizeHints* sizeHints = XAllocSizeHints();
    sizeHints->flags = PMinSize | PMaxSize;
    sizeHints->min_width = static_cast<int>(mins.x);
    sizeHints->min_height = static_cast<int>(mins.y);
    sizeHints->max_width = static_cast<int>(maxs.x);
    sizeHints->max_height = static_cast<int>(maxs.y);
    XSetWMNormalHints(display, window, sizeHints);
    XFree(sizeHints);
}

}

void WindowImplX11::setMinimumSize(const Vector2u &minimumSize)
{
    assert(minimumSize.x <= m_maximumSize.x);
    assert(minimumSize.y <= m_maximumSize.y);

    // Not sure what to do when resizing is disabled.
    // Ignore setMinimumSize is what I would recommend
    if (m_useSizeHints)
    {
        assert(false);
    }
    else
    {
        m_minimumSize = minimumSize;
        updateSizeHints(minimumSize, m_maximumSize, m_display, m_window);
    }
}

void WindowImplX11::setMaximumSize(const Vector2u &maximumSize)
{
    assert(maximumSize.x >= m_minimumSize.x);
    assert(maximumSize.y >= m_minimumSize.y);

    // Not sure what to do when resizing is disabled.
    // Ignore setMaximumSize is what I would recommend
    if (m_useSizeHints)
    {
        assert(false);
    }
    else
    {
        m_maximumSize = maximumSize;
        updateSizeHints(maximumSize, m_minimumSize, m_display, m_window);
    }
}
```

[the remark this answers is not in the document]
dogunbound: Done. I just didn't think about that for some reason.

ChrisThrasher:
https://developer.apple.com/documentation/appkit/nswindow/1419206-minsize
https://developer.apple.com/documentation/appkit/nswindow/1419595-maxsize

Looks like Cocoa has the functions we need to set the min and max size on an `NSWindow`. I can play with this later to see if I can get it working. It claims to include the size of the title bar. Does X11 include the title bar in the size of the window?

ShadowMarker789:
That's a good question. 

Is the Minimum Size the client-region or does it include the total non-client region of the window?

ChrisThrasher:
Didn't mean to push to the PR. Ignore that for now. I'm not done.

ChrisThrasher:
I've got it working on macOS! I've never done any Objective-C before so this may have flaws but it seems to work. Hopefully another macOS user can test this as well.

ChrisThrasher:
I fixed all CI jobs except the Windows job. Now it's up to someone else to finish the Windows side of this and we'll be able to do final review and testing to clean up all the details!

ChrisThrasher:

ShadowMarker789:
The Windows implementation has been completed, and all the checks have passed now, which is good. 👍

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -747,6 +761,90 @@ void WindowImplWin32::processEvent(UINT message, WPARAM wParam, LPARAM lParam)
...
+                // we also don't like this
+                if (isTop)
+                {
+                    ncsRect->top = ncsRect->bottom - static_cast<LONG>(m_maximumSize.y);
+                }
+                else
+                {
+                    ncsRect->bottom = ncsRect->top + static_cast<LONG>(m_maximumSize.y);
+                }
+            }
```
Bromeon, line 843: This is very verbose. Please reuse some code; the branches of `if` and `else if` are almost identical.

Also, could you make use of [`std::clamp()`](https://en.cppreference.com/w/cpp/algorithm/clamp), or at least `std::min/max()`?

Last, what is the "we don't like this" comment intended to mean? 🙂

src/SFML/Window/Win32/WindowImplWin32.hpp
```diff
@@ -291,6 +307,8 @@ class WindowImplWin32 : public WindowImpl
     bool          m_mouseInside{};   //!< Mouse is inside the window?
     bool          m_fullscreen{};    //!< Is the window fullscreen?
     bool          m_cursorGrabbed{}; //!< Is the mouse cursor trapped?
+    Vector2u m_minimumSize{};        //!< The minimum size of a window, as set by setMinimumSize(Vector2u&)
+    Vector2u m_maximumSize{};        //!< The maximum size of a window, as set by setMaximumSize(Vector2u&)
```
Bromeon, line 311: Alignment

Bromeon (commented):
Thanks! Looks like a good feature to have.

The API leaves it unclear how the min/max size can be reset; i.e. restore the state before it's called. I would _not_ use the zero vector for that -- that's C style and easy to accidentally interpret as a valid size.

What about a `std::optional<Vector2u>` parameter?

src/SFML/Window/Win32/WindowImplWin32.hpp
ChrisThrasher: clang-format CI jobs are passing. This is what we're stuck with ¯\_(ツ)_/¯

src/SFML/Window/Win32/WindowImplWin32.cpp
ShadowMarker789: "We don't like this" => The resize size proposed to us from the operating system violates the minimum or maximum constraints that the consuming developer has specified, so we must take action to adjust the rectangle to comply with the minimum or maximum values.

src/SFML/Window/Win32/WindowImplWin32.hpp
Bromeon: I see, I shall stand corrected. 🏰

src/SFML/Window/OSX/WindowImplCocoa.mm
```diff
@@ -436,6 +436,22 @@ void showMouseCursor()
...
+    AutoreleasePool pool;
+    [m_delegate setMinimumSize:NSMakeSize(minimumSize.x, minimumSize.y)];
+}
+
+
+////////////////////////////////////////////////////////////
+void WindowImplCocoa::setMaximumSize(const Vector2u& maximumSize)
+{
+    AutoreleasePool pool;
+    [m_delegate setMaximumSize:NSMakeSize(maximumSize.x, maximumSize.y)];
```
shane-tran-whitmire, line 451: Does this consider the window style of no resize? Does MAC not allow windows to disable resizing?

src/SFML/Window/OSX/WindowImplCocoa.mm
```diff
@@ -436,6 +436,22 @@ void showMouseCursor()
 }
 
 
+////////////////////////////////////////////////////////////
+void WindowImplCocoa::setMinimumSize(const Vector2u& minimumSize)
+{
+    AutoreleasePool pool;
+    [m_delegate setMinimumSize:NSMakeSize(minimumSize.x, minimumSize.y)];
```
shane-tran-whitmire, line 443: Does this consider the window style of no resize? Does MAC not allow windows to disable resizing?

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -825,6 +825,12 @@ Vector2u WindowImplX11::getSize() const
 ////////////////////////////////////////////////////////////
 void WindowImplX11::setSize(const Vector2u& size)
 {
+    // We could set the min and max to the new size in its respective dimension if setSize violates
+    // previously set min and max sizes
+    assert(size.x >= m_minimumSize.x);
+    assert(size.y >= m_minimumSize.y);
+    assert(size.x <= m_maximumSize.x);
+    assert(size.y <= m_maximumSize.y);
```
shane-tran-whitmire, line 833: ALL of the other setSizes need to consider this edge case too.

ChrisThrasher:
I think it makes sense to store the min and max sizes as optional vectors in `sf::WindowImpl`

ChrisThrasher:
I changed the API to take an `optional<Vector2u>`. If you pass it a null option that corresponds to un-setting the minimum or maximum size. This works on macOS but testing on Linux still shows some issues where un-setting the maximum size makes the window never appear.

MarioLiebisch:
I've added the two example lines (200x200 and 800x800) to the **window** example and it works, but my actual minimum/maximum window size don't add up on Windows 11.

* The window is initialized with `sf::VideoMode({640, 480})`, which for me results in a window that is actually 642x512 (extra width/height for title bar, border, and drop-shadow). The drawing area/canvas is 640x480.
* If I resize the window using `window.setSize(sf::Vector2u(400, 400));`, I get an actual window that is 402x432. This is once again adding the extra window decoration margins etc.
* However, setting the minimum/maximum size to 200x200/800x800 the actual minimum/maximum window sizes are 186x193 and 786x793.

MarioLiebisch:
One more thing, but I think this screws with other programs, too: If I drag the window more wide or high than is set, the window stays smaller, but once I hit the border of the screen, Windows will offer to maximize the window in that direction only (not actually maximizing it) and that will work. To get back to the example above, I can set a maximum size of 800x800, but if I drag the cursor to the top of the screen, in my case it would still resize to 800x1440.

ShadowMarker789:
> I've added the two example lines (200x200 and 800x800) to the **window** example and it works, but my actual minimum/maximum window size don't add up on Windows 11.
> 
> * The window is initialized with `sf::VideoMode({640, 480})`, which for me results in a window that is actually 642x512 (extra width/height for title bar, border, and drop-shadow). The drawing area/canvas is 640x480.
> * If I resize the window using `window.setSize(sf::Vector2u(400, 400));`, I get an actual window that is 402x432. This is once again adding the extra window decoration margins etc.
> * However, setting the minimum/maximum size to 200x200/800x800 the actual minimum/maximum window sizes are 186x193 and 786x793.

Ah, I think the WM_SIZING message gives us the TOTAL NCSRect whereas the sf::VideoMode is recard client-rectangle. 



> One more thing, but I think this screws with other programs, too: If I drag the window more wide or high than is set, the window stays smaller, but once I hit the border of the screen, Windows will offer to maximize the window in that direction only (not actually maximizing it) and that will work. To get back to the example above, I can set a maximum size of 800x800, but if I drag the cursor to the top of the screen, in my case it would still resize to 800x1440.

What would you expect the behaviour TO be for a window that has a set maximum size but is maximizable?

MarioLiebisch:
> What would you expect the behaviour TO be for a window that has a set maximum size but is maximizable?

I guess it should just get maximized anyway. But that's not how *this* Windows feature works. It will just resize the window, it won't be maximized.

Excuse the pun, but this is a literal edge case, so I wouldn't consider this a blocking issue or anything. 😉

https://user-images.githubusercontent.com/1854245/232190298-b6d51c64-3184-493d-988d-13f0ef2edf62.mp4

Also do you think setting a minimum/maximum size should have an impact on fullscreen windows? Should they ignore it? Should they automatically resize the viewport to limit the drawing area to the center of the screen?

src/SFML/Window/Win32/WindowImplWin32.cpp
MarioLiebisch: `std::min(a, b)` and `std::max(a, b)` just return the smaller or larger value. They just got added in C++11/14.
`std::clamp(value, low, high)` was added in C++17 and returns `value`, but ensuring it's neither lower nor higher than `low` and `high`.

Macro equivalents:
```cpp
#define min(a, b) ((a) < (b) ? (a) : (b))
#define max(a, b) ((a) > (b) ? (a) : (b))
#define clamp(v, l, h) ((v) < (l) ? (l) : ((v) > (h) ? (h) : (v)))
```
For example, to set the new bottom edge:
```cpp
ncsRect->bottom = ncsRect->top + std::clamp(height, m_minimumSize.y, m_maximumSize.y);
```
Or to set the new top edge:
```cpp
ncsRect->top = ncsRect->bottom - std::clamp(height, m_minimumSize.y, m_maximumSize.y);
```

dogunbound:
You can use this code to test enabling/disabling min/max window sizing

```c++
int main() {
  auto window =
      sf::RenderWindow{sf::VideoMode{{400u, 400u}}, "CMake SFML Project"};
  while (window.isOpen()) {
    for (auto event = sf::Event{}; window.pollEvent(event);) {
      switch (event.type) {
      case sf::Event::Closed:
        window.close();
        break;
      case sf::Event::KeyPressed:
        switch (event.key.code) {
        case sf::Keyboard::Q:
          window.setMinimumSize(sf::Vector2u(200, 200));
          break;
        case sf::Keyboard::A:
          window.setMinimumSize({});
          break;
        case sf::Keyboard::W:
          window.setMaximumSize(sf::Vector2u(800, 800));
          break;
        case sf::Keyboard::S:
          window.setMaximumSize({});
          break;
        default:
          break;
        }
        break;
      default:
        break;
      }
    }

    window.clear();
    window.display();
  }
}
```

ChrisThrasher:
> You can use this code to test enabling/disabling min/max window sizing

Thanks for this test program! I put it into a repo and added some debug info to the window so we can see what's going on. This should help us test since we need to be testing the same things across many platforms and it's annoying to manually set up this test project. 

Screenshot 2023-04-16 at 5 33 20 PM

I can reproduce the weird issue where setting a minimum window size of 400x400 actually lets you make the window smaller than that. I think we need to do some extra math to account for the size of the title bar or something like that.

ChrisThrasher:
https://developer.apple.com/documentation/appkit/nswindow/1419670-contentminsize

There's a convenient Cocoa API for setting the size of the _content_ of the window instead of the entire window itself. Now we get exactly the size we expect on macOS 🙂

ChrisThrasher:
I'm getting some weird behavior on Linux. If I scale the window beyond the max, then set the maximum size to a smaller size, it automatically snaps down to that size which differs from macOS but is still acceptable behavior to me. However if then unset the maximum size, it snaps back to the bigger size it used to be. That much definitely seems like a bug to me.

src/SFML/Window/OSX/SFViewController.mm
```diff
@@ -172,6 +172,18 @@ - (void)resizeTo:(unsigned int)width by:(unsigned int)height
...
+{
+}
+
+
+////////////////////////////////////////////////////////
+- (void)setMaximumSize:(NSSize)size
+{
+}
+
+
```
ChrisThrasher, line 186: I wasn't entirely sure what this file did. Does it make sense to set the min or max size on a view?

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -747,6 +759,90 @@ void WindowImplWin32::processEvent(UINT message, WPARAM wParam, LPARAM lParam)
             break;
         }
 
+        // tell OS whether we're happy with the proposed changes to the window size
```
ChrisThrasher, line 762:
```suggestion
        // Tell OS whether we're happy with the proposed changes to the window size
```

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -747,6 +759,90 @@ void WindowImplWin32::processEvent(UINT message, WPARAM wParam, LPARAM lParam)
...
+        case WM_SIZING:
+        {
+            RECT* ncsRect = reinterpret_cast<RECT*>(lParam);
+            assert(ncsRect->left < ncsRect->right);
+            assert(ncsRect->top < ncsRect->bottom);
+            auto width  = static_cast<unsigned int>(ncsRect->right - ncsRect->left);
+            auto height = static_cast<unsigned int>(ncsRect->bottom - ncsRect->top);
+            bool isTop  = false; // are we being resized on the top? If not, it's the bottom
```
ChrisThrasher, line 770:
```suggestion
            bool resizingFromTop  = false; // Are we being resized on the top? If not, it's the bottom
```
If this big comment is required to explain what the variable does, perhaps its name isn't expressive enough.

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -747,6 +759,90 @@ void WindowImplWin32::processEvent(UINT message, WPARAM wParam, LPARAM lParam)
...
+            bool isLeft = false; // are we being resized on the left? If not, it's the right
```
ChrisThrasher, line 771:
```suggestion
            bool resizingFromLeft = false; // Are we being resized on the left? If not, it's the right
```

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -747,6 +759,90 @@ void WindowImplWin32::processEvent(UINT message, WPARAM wParam, LPARAM lParam)
...
+                    break;
+                case WMSZ_TOP:
+                    isTop = true;
+                    break;
+            }
+
+
+            if (m_minimumSize && width < m_minimumSize->x)
+            {
+                // we don't like this
```
ChrisThrasher, line 795: These comments aren't obvious to me. Explain more about why we "don't like this".

src/SFML/Window/Win32/WindowImplWin32.hpp
```diff
@@ -110,6 +110,22 @@ class WindowImplWin32 : public WindowImpl
     ////////////////////////////////////////////////////////////
     void setSize(const Vector2u& size) override;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Set the minimum size of the window, preventing it from being user-resized to smaller than this. Defaults to 0,0
```
ChrisThrasher, line 114:
```suggestion
    /// \brief Sets a minimum window rendering region size
```
Let's use the same `\brief` as the other implementations.

src/SFML/Window/Win32/WindowImplWin32.hpp
```diff
@@ -110,6 +110,22 @@ class WindowImplWin32 : public WindowImpl
...
+    ///
+    /// \param size New minimum size, in pixels
+    ///
+    ////////////////////////////////////////////////////////////
+    void setMinimumSize(const std::optional<Vector2u>& minimumSize) override;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Set the maximum size of the window, preventing it from being user-resized to bigger than this. 0,0 (default) means no maximum.
```
ChrisThrasher, line 122:
```suggestion
    /// \brief Sets a maximum window rendering region size
```

src/SFML/Window/WindowBase.cpp
```diff
@@ -217,6 +217,23 @@ void WindowBase::setSize(const Vector2u& size)
...
+    }
+}
+
+////////////////////////////////////////////////////////////
+void WindowBase::setMaximumSize(const std::optional<Vector2u>& maximumSize)
+{
+    if (m_impl)
+    {
+        m_impl->setMaximumSize(maximumSize);
+    }
```
ChrisThrasher, line 235:
```suggestion
    if (m_impl)
        m_impl->setMaximumSize(maximumSize);
```
Braces aren't used for similar things elsewhere in this file.

src/SFML/Window/WindowBase.cpp
```diff
@@ -217,6 +217,23 @@ void WindowBase::setSize(const Vector2u& size)
     }
 }
 
+////////////////////////////////////////////////////////////
+void WindowBase::setMinimumSize(const std::optional<Vector2u>& minimumSize)
+{
+    if (m_impl)
+    {
+        m_impl->setMinimumSize(minimumSize);
+    }
```
ChrisThrasher, line 226:
```suggestion
    if (m_impl)
        m_impl->setMinimumSize(minimumSize);
```

ChrisThrasher (changes requested):

src/SFML/Window/OSX/SFViewController.mm
dogunbound: I don't think so.

dogunbound:
> I'm getting some weird behavior on Linux. If I scale the window beyond the max, then set the maximum size to a smaller size, it automatically snaps down to that size which differs from macOS but is still acceptable behavior to me. However if I then unset the maximum size, it snaps back to the bigger size it used to be. That much definitely seems like a bug to me.
> 
> EDIT: I'm apparently using Wayland under the hood and this problem doesn't exist for those using X11 directly.

This is a wayland bug. We could fix this by calling setsize and clamp it down, but this bug might get fixed by wayland anyways. Plus this has a very low likelihood of cropping up, and doesn't mean much in the grand scheme of things.

ChrisThrasher:
Rebased on `master` and addressed review comment

src/SFML/Window/Win32/WindowImplWin32.cpp
Bromeon: This is still unadressed. I would not solve it with macros though.

ChrisThrasher: I cleaned up this implementation a bit but I still can't figure out `std::min`/`std::max`/`std::clamp` would help. The problem is that we have to do different things based on the direction from which resizing is occurring. We're not simply taking a vector and clamping it a min and max. We have to take into account the side of the window in which resizing is occurring.

Regardless, shouldn't we be using something like this: https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-getminmaxinfo?redirectedfrom=MSDN

Surely Windows provides ways of telling it what the minimum and maximum sizes are instead of implementing all of this logic ourselves, right?

ChrisThrasher:
I just realized we probably need someone to test the macOS implementation on a non-retina display. I may need to call into some functions that take into account stuff like DPI. Can anyone do that?

ChrisThrasher:
I learned that you can call `sf::Window::setSize` to something outside of the min/max bounds on macOS. Not yet sure how best to handle this. I suppose we query the min/max size and clamp accordingly before setting the size? I was hoping Cocoa would automatically handle this for us.

dogunbound:
@ChrisThrasher I think we need to store the min and max window size if we want to be synonymous with all of the OS. The reason is that some OS don't have a good way to get the minimum window size.

SDL does this:

```c++
/* Define the SDL window structure, corresponding to toplevel windows */
struct SDL_Window
{
    const void *magic;
    SDL_WindowID id;
    char *title;
    SDL_Surface *icon;
    int x, y;
    int w, h;
    int min_w, min_h;
    int max_w, max_h;
    int last_pixel_w, last_pixel_h;
    Uint32 flags;
    SDL_bool fullscreen_exclusive;  /* The window is currently fullscreen exclusive */
    SDL_DisplayID last_fullscreen_exclusive_display;  /* The last fullscreen_exclusive display */
    SDL_DisplayID last_displayID;

    /* Stored position and size for windowed mode */
    SDL_Rect windowed;

    /* Whether or not the intial position was defined */
    SDL_bool undefined_x;
    SDL_bool undefined_y;

    SDL_DisplayMode requested_fullscreen_mode;
    SDL_DisplayMode current_fullscreen_mode;

    float opacity;

    SDL_Surface *surface;
    SDL_bool surface_valid;

    SDL_bool is_hiding;
    SDL_bool restore_on_show; /* Child was hidden recursively by the parent, restore when shown. */
    SDL_bool is_destroying;
    SDL_bool is_dropping; /* drag/drop in progress, expecting SDL_SendDropComplete(). */

    SDL_Rect mouse_rect;

    SDL_WindowShaper *shaper;

    SDL_HitTest hit_test;
    void *hit_test_data;

    SDL_WindowUserData *data;

    SDL_WindowData *driverdata;

    SDL_Window *prev;
    SDL_Window *next;

    SDL_Window *parent;
    SDL_Window *first_child;
    SDL_Window *prev_sibling;
    SDL_Window *next_sibling;
};
```

ShadowMarker789:
> @ShadowMarker789 Can you look into whether this API will work for our needs? https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-getminmaxinfo?redirectedfrom=MSDN

This is for adjusting how the window should be when it's maximized! 

Good find!

maximized-sfml-window

We can restrict the size of the window by modifying the parameters in this window message. In the above example I set the maximized height and width to 1280x720. 

```cs
case WM.WM_GETMINMAXINFO:
{
	MINMAXINFO* mmi = (MINMAXINFO*)lParam;
	mmi->ptMaxSize.x = 1280;
	mmi->ptMaxSize.y = 720;
}
```

Though in modern C++ you'd be doing a reinterpret_cast<MINMAXINFO*> right?

ShadowMarker789:
> @ShadowMarker789 Can you look into whether this API will work for our needs? https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-getminmaxinfo?redirectedfrom=MSDN

https://user-images.githubusercontent.com/37165910/234185922-58f8ea26-1732-43a4-bd07-b2e6c9cf43ef.mp4

Works great for what we want. We can clamp the window size without needing to do the bunch of work that we were otherwise doing in the WM_SIZING message. 

My C# code to do the above looks like this: 

```cs
case WM.WM_GETMINMAXINFO:
{
	MINMAXINFO* mmi = (MINMAXINFO*)lParam;
	mmi->ptMaxSize.x = 1280;
	mmi->ptMaxSize.y = 720;
	mmi->ptMinTrackSize.x = 640;
	mmi->ptMinTrackSize.y = 360;
	mmi->ptMaxTrackSize.x = 1280;
	mmi->ptMaxTrackSize.y = 720;
	return 0;
}
```
Are you happy to get this translated into C++ code to have the same behaviour as shown in the video above?

ChrisThrasher:
> Though in modern C++ you'd be doing a reinterpret_cast<MINMAXINFO*> right?

Yeah that's right.

> Works great for what we want. We can clamp the window size without needing to do the bunch of work that we were otherwise doing in the WM_SIZING message.

That's awesome! So glad to see a radically simplified implementation.

> Are you happy to get this translated into C++ code to have the same behavior as shown in the video above?

Yes please do.

ChrisThrasher:
I think we need to use `sf::WindowBase::setSize` to do some checks where we don't let the window get resized to somethign outside of the min/max bounds. We also probably need to modify `sf::WindowBase::setMinimumSize` and `sf::WindowBase::setMaximumSize` to disallow setting either of those if the window is using the `Fullscreen` style. In general I think we can use `sf::WindowBase` to enforce a lot of this shared logic.

ChrisThrasher:
I made a big change that now enforces the minimum and maximum sizes much more strongly. Here are a few things it does:

1. Asserts if minimum exceeds maximum or vice versa
2. Ensures calls to `sf::WindowBase::setSize` adhere to size constraints
3. Resizes window as needed if current size is outside of size constraints

I was even able to add some unit tests for this to be extra sure it's working.

I expect the implementation still needs some cleanup but for now I'm focused on functionally. I think this is the last bit of functionality we needed unless I'm missing something so please let me know if you find any oddities. Reminder there's my test project [here](https://github.com/ChrisThrasher/sfml-test-2519) you can use to more easily test this.

ChrisThrasher:

ChrisThrasher:
To solve this I had to think hard about data organization. The minimum and maximum sizes are owned by `WindowImpl`. This is the base class for all OS-specific implementations. However, there is some common behaviors we need to implement across all platforms. That behavior is implemented in `WindowBase` so that we get consistent behavior across platforms without code duplication. This `WindowBase` behavior required that `WindowImpl` provide both setters and getters for min/max size.

I could have directly used `m_{min|max}imumSize` within the various `WindowImpl` implementations but that would have required using `protected` data members which I did not want to do. Since I already needed to add `sf::priv::WindowImpl::get{min|max}imumSize()` for `WindowBase` I figured I'd just use those in the derived classes too.

eXpl0it3r:
Another small but cool detail is that at least on macOS, the OS will change the dragging mouse cursor, so if you're at the limit, it would indicate the only available direction 🙂 
image

dogunbound (commented):
dfa

src/SFML/Window/WindowBase.cpp
```diff
@@ -207,18 +211,67 @@ void WindowBase::setSize(const Vector2u& size)
...
+{
+    if (m_impl)
+    {
+        [[maybe_unused]] const auto minimumSizeIsLessThanMaximumSize = [this, minimumSize]()
+        {
+            if (!minimumSize.has_value() || !m_impl->getMaximumSize().has_value())
+                return true;
+            return minimumSize->x < m_impl->getMaximumSize()->x && minimumSize->y < m_impl->getMaximumSize()->y;
+        };
+        assert(minimumSizeIsLessThanMaximumSize() && "Minimum size must be less than maximum size");
```
dogunbound, line 248: I'm not so sure we should assert here. I think clamping and outputting a descript error would be better.

Output examples:
"Maximum size (200x400) is smaller than minimum size (300x200)! Ignoring"
"Minimum size (900x700) is larger than maximum size (800x600)! Ignoring"

Make sure this goes to stderr.

It would probably look something like this:
```c++
if (!minimumSizeIsLessThanMaximumSize())
{
    std::cerr << "Minimum size (" << minimumSize.x << "x" << minimumSize.y << ") is larger than maximum size (" << m_impl->getMaximumSize()->x << "x" << m_impl->getMaximumSize()->y << ")! Ignoring\n";
    return;
}
```

src/SFML/Window/WindowBase.cpp
```diff
@@ -207,18 +211,67 @@ void WindowBase::setSize(const Vector2u& size)
...
+{
+    if (m_impl)
+    {
+        [[maybe_unused]] const auto maximumSizeIsGreaterThanMinimumSize = [this, maximumSize]()
+        {
+            if (!maximumSize.has_value() || !m_impl->getMinimumSize().has_value())
+                return true;
+            return maximumSize->x > m_impl->getMinimumSize()->x && maximumSize->y > m_impl->getMinimumSize()->y;
+        };
+        assert(maximumSizeIsGreaterThanMinimumSize() && "Maximum size must be greater than minimum size");
```
dogunbound, line 267: I'm not so sure we should assert here. I think ignoring and outputting a descript error would be better.

Output examples:
"Maximum size (200x400) is smaller than minimum size (300x200)! Ignoring"
"Minimum size (900x700) is larger than maximum size (800x600)! Ignoring"

Make sure this goes to stderr.

It would probably look something like this:
```c++
if (!maximumSizeIsGreaterThanMinimumSize())
{
    std::cerr << "Maximum size (" << maximumSize.x << "x" << maximumSize.y << 
        ") is smaller than minimum size (" << m_impl->getMinimumSize()->x << 
        "x" << m_impl->getMinimumSize()->y << ")! Ignoring\n";
    return;
}
```

src/SFML/Window/WindowBase.cpp
dogunbound: Also, inverting the output and renaming the `minimumSizeIsLessThanMaximumSize` function to `minimumSizeIsGreaterThanMaximumSize` may be more optimal in this situation.

src/SFML/Window/WindowBase.cpp
dogunbound: Also, inverting the output and renaming the `maximumSizeIsGreaterThanMinimumSize` function to `maximumSizeIsSmallerthanMaximumSize` may be more optimal in this situation.

dogunbound: https://github.com/SFML/SFML/pull/2519#discussion_r1181098760

Comment on this one. I'm resolving this conversation to avoid people splitting off conversations between the two reviews.

src/SFML/Window/WindowBase.cpp
dogunbound: Another suggestion would be to assert and do an early return that way debug builds crash and release builds have well defined behavior.

ChrisThrasher: When writing much of this I knew there were a lot of options for error detection. I thought using an `assert` was an easy place to start but I'm open to a solution like this. There are a number ways we can define what happens when users provide conflict minimum and maximum sizes. I'm curious to hear the rest of team weigh in on how we should handle this case.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -2139,4 +2154,28 @@ Vector2i WindowImplX11::getPrimaryMonitorPosition()
     return monitorPosition;
 }
 
+
+////////////////////////////////////////////////////////////
+void WindowImplX11::setWindowSizeConstraints() const
```
kimci86, line 2159: I think this function should not do anything for a window that is not resizable, i.e. created without the flag sf::Style::Resize. Otherwise it can turn the window into a resizable one.

This can be done like so:
```c++
    if(m_useSizeHints)
        return;
```

src/SFML/Window/WindowBase.cpp
```diff
@@ -207,18 +211,67 @@ void WindowBase::setSize(const Vector2u& size)
...
+////////////////////////////////////////////////////////////
+void WindowBase::setMinimumSize(const std::optional<Vector2u>& minimumSize)
+{
+    if (m_impl)
+    {
+        [[maybe_unused]] const auto minimumSizeIsLessThanMaximumSize = [this, minimumSize]()
+        {
+            if (!minimumSize.has_value() || !m_impl->getMaximumSize().has_value())
+                return true;
+            return minimumSize->x < m_impl->getMaximumSize()->x && minimumSize->y < m_impl->getMaximumSize()->y;
```
kimci86, line 246: I guess it should be valid for minimum and maximum sizes to be the same.
```suggestion
            return minimumSize->x <= m_impl->getMaximumSize()->x && minimumSize->y <= m_impl->getMaximumSize()->y;
```
Maybe adjust lambda name and assert message accordingly.

src/SFML/Window/WindowBase.cpp
```diff
@@ -207,18 +211,67 @@ void WindowBase::setSize(const Vector2u& size)
...
+////////////////////////////////////////////////////////////
+void WindowBase::setMaximumSize(const std::optional<Vector2u>& maximumSize)
+{
+    if (m_impl)
+    {
+        [[maybe_unused]] const auto maximumSizeIsGreaterThanMinimumSize = [this, maximumSize]()
+        {
+            if (!maximumSize.has_value() || !m_impl->getMinimumSize().has_value())
+                return true;
+            return maximumSize->x > m_impl->getMinimumSize()->x && maximumSize->y > m_impl->getMinimumSize()->y;
```
kimci86, line 265:
```suggestion
            return maximumSize->x >= m_impl->getMinimumSize()->x && maximumSize->y >= m_impl->getMinimumSize()->y;
```

src/SFML/Window/WindowBase.cpp
ChrisThrasher: Fixed 👍🏻

src/SFML/Window/WindowBase.cpp
ChrisThrasher: Fixed 👍🏻

ChrisThrasher:

src/SFML/Window/WindowBase.cpp
```diff
@@ -207,18 +211,67 @@ void WindowBase::setSize(const Vector2u& size)
...
+            if (!minimumSize.has_value() || !m_impl->getMaximumSize().has_value())
+                return true;
+            return minimumSize->x <= m_impl->getMaximumSize()->x && minimumSize->y <= m_impl->getMaximumSize()->y;
+        };
+        assert(validateMinimumSize() && "Minimum size cannot be bigger than the miximum size along either axis");
+
+        m_impl->setMinimumSize(minimumSize);
+        setSize(getSize());
+    }
+}
```
ChrisThrasher, line 253: If the current window style was stored within the class, we could return early if this function or `setMaximumSize` is called when the style include `sf::Style::Fullscreen` or excludes `sf::Style::Resize`. That would let me remove some fixes I pushed to the X11 implementation and help ensure all implementations always behave the same.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher (approved):
While I think it's possible we tweak the behavior in the future, particular related to error handling, I still think it's reasonable to merge this and get it into the hands of SFML 3 users. Thank you to everyone who helped test this!

src/SFML/Window/OSX/WindowImplCocoa.mm
ChrisThrasher: That's all automatically handled. Just like the other platforms, macOS will resize the window if the application calls `setSize` but it doesn't not let users resize the window. You can still freely change the min/max size while the window has a non-resizable style.

ChrisThrasher:
Most of our examples disable resizing but one of then did not so I added some minimum and maximum bounds to that example just for the hell of it so at least one example would show these APIs in action.

ChrisThrasher:

dogunbound:
Well, I approve of these changes! We can also close #2124 once this is merged! Gotta leave a comment on the new API

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -843,6 +844,28 @@ void WindowImplX11::setSize(const Vector2u& size)
...
 
 
+////////////////////////////////////////////////////////////
+void WindowImplX11::setMinimumSize(const std::optional<Vector2u>& minimumSize)
+{
+    // Do nothing if resizing is disabled
+    if (m_useSizeHints)
+        return;
+
+    WindowImpl::setMinimumSize(minimumSize);
```
kimci86, line 854: I am wondering if this call should happen even when resizing is disabled, so that asserts about a potential following setMaximumSize call would be be triggered like on other platforms.

ChrisThrasher: If we don't disable `WindowImpl::setMinimumSize(minimumSize);` then a non-resizeable window suddenly becomes resizable by the user.

kimci86: I think this is what `setWindowSizeConstraints()` would do, but not `WindowImpl::setMinimumSize(minimumSize)`?
Anyways, maybe we can call it good enough for now and polish when we implement more window states and styles.

ChrisThrasher:

ChrisThrasher:
When dragging a window to the edge of the screen on Linux it still manages to exceed the maximum size constraints as it snaps to fill half or all of the screen depending on which edge you drag the window towards. Not sure what we can do about that. Not sure if X11 lets us detect that and override it.

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -780,14 +786,25 @@ void WindowImplWin32::processEvent(UINT message, WPARAM wParam, LPARAM lParam)
...
+            const auto maximumSize = contentSizeToWindowSize(getMaximumSize().value_or(Vector2u(10'000, 10'000)), m_handle);
+
+            MINMAXINFO& minMaxInfo      = *reinterpret_cast<PMINMAXINFO>(lParam);
+            minMaxInfo.ptMaxTrackSize.x = maximumSize.x;
+            minMaxInfo.ptMaxTrackSize.y = maximumSize.y;
+            if (getMaximumSize().has_value())
+            {
+                minMaxInfo.ptMaxSize.x = maximumSize.x;
+                minMaxInfo.ptMaxSize.y = maximumSize.y;
+            }
```
kimci86, line 801: This part seems redundant.

kimci86 (commented):
Some smalls remarks.

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -780,14 +786,25 @@ void WindowImplWin32::processEvent(UINT message, WPARAM wParam, LPARAM lParam)
...
-        // The system request the min/max window size and position
+        // Fix violations of minimum or maximum size
         case WM_GETMINMAXINFO:
         {
-            // We override the returned information to remove the default limit
-            // (the OS doesn't allow windows bigger than the desktop by default)
-            auto* info             = reinterpret_cast<MINMAXINFO*>(lParam);
-            info->ptMaxTrackSize.x = 50000;
-            info->ptMaxTrackSize.y = 50000;
+            const auto maximumSize = contentSizeToWindowSize(getMaximumSize().value_or(Vector2u(10'000, 10'000)), m_handle);
```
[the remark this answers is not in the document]
ChrisThrasher: I picked 10k since 8k displays were as big as I ever expected anyone using. I don't mind restoring it to 50k though.

src/SFML/Window/Win32/WindowImplWin32.cpp
kimci86: Nevermind, I thought it was doing the same as the line before but ptMaxTrackSize is not ptMaxSize :facepalm:

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -38,6 +38,7 @@
 // expects lowercase, and a native compile on windows, whether via msvc
 // or mingw-w64 addresses files in a case insensitive manner.
 #include <dbt.h>
+#include <limits>
```
kimci86, line 41: It looks like this is not needed.

kimci86 (approved):
Code looks good to me.

src/SFML/Window/Win32/WindowImplWin32.cpp
ChrisThrasher: Thank you yes that's right. I previously used that head but forgot to remove it.

ChrisThrasher:

ChrisThrasher:

dogunbound:
> When dragging a window to the edge of the screen on Linux it still manages to exceed the maximum size constraints as it snaps to fill half or all of the screen depending on which edge you drag the window towards. Not sure what we can do about that. Not sure if X11 lets us detect that and override it.

That's an interesting quirk. It's possible the desktop environment you are using sends min/max size constraints to x11 to allow it to snap to the edges. There really isn't much we can do about that.

ChrisThrasher:
@vittorioromeo @Bromeon @binary1248 Would you please take a look at this?

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -780,14 +785,28 @@ void WindowImplWin32::processEvent(UINT message, WPARAM wParam, LPARAM lParam)
...
+        // Fix violations of minimum or maximum size
         case WM_GETMINMAXINFO:
         {
             // We override the returned information to remove the default limit
             // (the OS doesn't allow windows bigger than the desktop by default)
-            auto* info             = reinterpret_cast<MINMAXINFO*>(lParam);
-            info->ptMaxTrackSize.x = 50000;
-            info->ptMaxTrackSize.y = 50000;
+
+            const auto maximumSize = contentSizeToWindowSize(getMaximumSize().value_or(Vector2u(50'000, 50'000)), m_handle);
```
Bromeon, line 794: I know the magic number was already there, is it just supposed to be "big enough" to exceed all realistic window sizes?

Also, variable name seems confusing, should probably be
```suggestion
            const auto maximumSize = getMaximumSize().value_or(Vector2u(50'000, 50'000));
            const auto windowSize = contentSizeToWindowSize(maximumSize, m_handle);
```

Bromeon (approved):
Looks good! Had 1 question, but not necessarily warranting a change, so I'll approve 🙂

src/SFML/Window/Win32/WindowImplWin32.cpp
dogunbound: That most definitely is for that purpose. It would be better if you use u32: :MAX size for this if you can. That would more semantically state that it is for that purpose.

Bromeon: It's unfortunately not Rust, so `std::numeric_limits<unsigned int>::max()` will have to do 😉

(`unsigned int` because we're using `Vector2u` here)

ChrisThrasher: > is it just supposed to be "big enough" to exceed all realistic window sizes?

Supposedly? In the year 2023 it's a safe bet that few SFML users have 50k displays. Maybe they'll be commonplace in 10 years though ¯\\\_(ツ)\_/¯.

I tried to replace `50'000` with `std::numeric_limits<unsigned int>::max()` but I got a bunch of GL errors so I don't think we can do that. The magic numbers will have to stay.

> Also, variable name seems confusing, should probably be

Lower down I use that same pattern for `minimumSize`. It would be asymmetric to change `maximumSize` to `windowSize` when `minimumSize` is also a type of window size. I could use `maximumWindowSize` and `minimumWindowSize` if you'd like though.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ShadowMarker789 (approved):
Looks good, happy to help write and test the Win32 implementation.

src/SFML/Window/Win32/WindowImplWin32.cpp
ShadowMarker789: > That most definitely is for that purpose. It would be better if you we have u32::MAX size for this if you can. That would more semantically state that it is for that purpose.
> 
> @ChrisThrasher or @ShadowMarker789 what do you think? Maybe there is a windows thing that doesn't let it get too big? If so, I would think u16::MAX would be better.

As it happens, by default several fields for the maximum values of the window tracking and/or maximized values will be defaulted to the working size of the machine. 

Where the working size of the machine is defined as the total display rectangle of all combined virtual displays minus the rectangle of the taskbar if it is not auto-hiding. 

If you are testing the Win32 implementation, set a breakpoint prior to the assigning of the values of the fields of the tracking sizes - you'll see that they have prepopulated default values that are "reasonable" for the current computer environment.

src/SFML/Window/OSX/WindowImplCocoa.mm
```diff
@@ -437,6 +438,27 @@ void showMouseCursor()
...
+}
+
+
+////////////////////////////////////////////////////////////
+void WindowImplCocoa::setMaximumSize(const std::optional<Vector2u>& maximumSize)
+{
+    WindowImpl::setMaximumSize(maximumSize);
+    const AutoreleasePool pool;
+    const NSSize          size = maximumSize ? NSMakeSize(maximumSize->x, maximumSize->y)
+                                             : NSMakeSize(std::numeric_limits<CGFloat>::max(), std::numeric_limits<CGFloat>::max());
```
eXpl0it3r, line 457: Why do we set a maximum size of numeric_limits when no maximumSize is provided?

src/SFML/Window/Unix/WindowImplX11.cpp
eXpl0it3r: I'm not sure about this one. You currently can make the window not resizable but still call `setSize` on it. Or do the limits not consider `setSize` calls?

src/SFML/Window/WindowBase.cpp
```diff
@@ -207,18 +211,67 @@ void WindowBase::setSize(const Vector2u& size)
 {
     if (m_impl)
     {
-        m_impl->setSize(size);
+        // Constrain requested size within minimum and maximum bounds
+        const auto minimumSize = m_impl->getMinimumSize().value_or(sf::Vector2u());
+        const auto maximumSize = m_impl->getMaximumSize().value_or(
+            sf::Vector2u(std::numeric_limits<unsigned int>::max(), std::numeric_limits<unsigned int>::max()));
```
eXpl0it3r, line 217: Are `{}` not enough to init a `sf::Vector2u`?

eXpl0it3r: `sf::` namespace is not needed here

src/SFML/Window/WindowBase.cpp
```diff
@@ -207,18 +211,67 @@ void WindowBase::setSize(const Vector2u& size)
...
+        const auto width  = std::clamp(size.x, minimumSize.x, maximumSize.x);
+        const auto height = std::clamp(size.y, minimumSize.y, maximumSize.y);
+
+        // Do nothing if requested size matches current size
+        const sf::Vector2u clampedSize(width, height);
```
eXpl0it3r, line 222: Nor here

src/SFML/Window/WindowBase.cpp
```diff
@@ -207,18 +211,67 @@ void WindowBase::setSize(const Vector2u& size)
...
+{
+    if (m_impl)
+    {
+        [[maybe_unused]] const auto validateMinimumSize = [this, minimumSize]()
+        {
+            if (!minimumSize.has_value() || !m_impl->getMaximumSize().has_value())
+                return true;
+            return minimumSize->x <= m_impl->getMaximumSize()->x && minimumSize->y <= m_impl->getMaximumSize()->y;
+        };
+        assert(validateMinimumSize() && "Minimum size cannot be bigger than the miximum size along either axis");
```
eXpl0it3r, line 248: `maximum size`*

eXpl0it3r (changes requested):
Some questions and some tiny changes. 🙂 

Have we tested the following scenarios:
- Calling `setMin/Max` when the window has visibility set to false?
- Calling `setMin/Max` directly after window creation?

src/SFML/Window/Unix/WindowImplX11.cpp
dogunbound: WindowBase.cpp:setSize implementation takes care of this situation by clamping. 

If the user sets a size for the x dimension at 50, but they previously set the lower limit to 200, it will clamp their own setSize to 200.

src/SFML/Window/OSX/WindowImplCocoa.mm
ChrisThrasher: It's just what the Cocoa API expects. `FLT_MAX` is the default maximum size.

https://developer.apple.com/documentation/appkit/nswindow/1419595-maxsize

src/SFML/Window/WindowBase.cpp
ChrisThrasher: > Are {} not enough to init a sf::Vector2u?

For whatever reason, no.

eXpl0it3r (approved):

eXpl0it3r:

ChrisThrasher:
> I'm not sure about this one. You currently can make the window not resizable but still call setSize on it. Or do the limits not consider setSize calls?

This is a feature and has been true even before this PR if I recall correctly. The idea is that if the programmer disables resizing, they themselves are still allowed to programmatically resize the window but users are not.

> Have we tested the following scenarios:
> 
> Calling setMin/Max when the window has visibility set to false?
> Calling setMin/Max directly after window creation?

I had not tested those but just tried doing both. On macOS and Ubuntu I don't observe any anomalous behavior.

ChrisThrasher:
Thank you to everyone who helped write, test, and review this! These are the kinds of changes SFML needs most. Let's keep it up!

PR #2520, opened 2023-04-13
Simplify 'FontHandles' management

vittorioromeo:
Behavior should be equivalent to what it was before. Tested all examples as well.


src/SFML/Graphics/Font.cpp
```diff
@@ -91,31 +92,28 @@ std::uint64_t combine(float outlineThickness, bool bold, std::uint32_t index)
...
+    FontHandles(const FontHandles&)            = delete;
+    FontHandles& operator=(const FontHandles&) = delete;
 
-public:
-    std::unique_ptr<std::remove_pointer_t<FT_Library>, Deleter> library;   //< Pointer to the internal library interface
-    std::unique_ptr<FT_StreamRec>                               streamRec; //< Pointer to the stream rec instance
-    std::unique_ptr<std::remove_pointer_t<FT_Face>, Deleter>    face;      //< Pointer to the internal font face
-    std::unique_ptr<std::remove_pointer_t<FT_Stroker>, Deleter> stroker;   //< Pointer to the stroker
+    FontHandles(FontHandles&&)            = delete;
+    FontHandles& operator=(FontHandles&&) = delete;
```
vittorioromeo, line 111: These guarantee that the destructor is good enough for cleanup, as we don't ever perform ownership transfer.

src/SFML/Graphics/Font.cpp
```diff
@@ -292,37 +284,34 @@ bool Font::loadFromStream(InputStream& stream)
     }
 
     // Prepare a wrapper for our stream, that we'll pass to FreeType callbacks
-    fontHandles->streamRec                     = std::make_unique<FT_StreamRec>();
```
vittorioromeo, line 295: Not sure why this was being allocated separately before -- it looks unnecessary to me unless I am missing something.

src/SFML/Graphics/Font.cpp
```diff
@@ -388,7 +377,7 @@ float Font::getKerning(std::uint32_t first, std::uint32_t second, unsigned int c
     if (first == 0 || second == 0)
         return 0.f;
 
-    auto* face = m_fontHandles ? m_fontHandles->face.get() : nullptr;
+    FT_Face face = m_fontHandles ? m_fontHandles->face : nullptr;
```
vittorioromeo, line 380: `auto` here was obfuscating the original type, it is not obvious from the initialization expression.

vittorioromeo:

src/SFML/Graphics/Font.cpp
```diff
@@ -91,31 +92,28 @@ std::uint64_t combine(float outlineThickness, bool bold, std::uint32_t index)
...
-        }
-        void operator()(FT_Stroker theStroker)
-        {
-            FT_Stroker_Done(theStroker);
-        }
-    };
+        FT_Stroker_Done(stroker);
+        FT_Done_Face(face);
+        // `streamRec` doesn't need to be explicitly freed.
+        FT_Done_FreeType(library);
```
Bromeon, line 104: In code below, if initialization of some members fails, the function returns and the destructor is run -- with the remaining members default-initialized.

Is invoking each of these `FT_*` functions on `nullptr` save, with defined behavior?

src/SFML/Graphics/Font.cpp
```diff
@@ -155,36 +153,32 @@ bool Font::loadFromFile(const std::filesystem::path& filename)
     // Cleanup the previous resources
     cleanup();
 
-    auto fontHandles = std::make_unique<FontHandles>();
+    auto fontHandles = std::make_shared<FontHandles>();
```
Bromeon, line 156: This probably still needs to be `shared_ptr` for dependency reasons (only declare destructor in the .cpp file)? Or do we actually use shared semantics?

src/SFML/Graphics/Font.cpp
```diff
@@ -155,36 +153,32 @@ bool Font::loadFromFile(const std::filesystem::path& filename)
...
     {
         err() << "Failed to load font (failed to initialize FreeType)\n" << formatDebugPathInfo(filename) << std::endl;
         return false;
     }
-    fontHandles->library.reset(library);
 
     // Load the new font face from the specified file
     FT_Face face;
-    if (FT_New_Face(library, filename.string().c_str(), 0, &face) != 0)
+    if (FT_New_Face(fontHandles->library, filename.string().c_str(), 0, &face) != 0)
```
kimci86, line 169: Is there a reason you are getting the font face in a temporary `FT_Face` instance which you later assign to `fontHandles->face` instead of getting it directly into `fontHandles->face` like you do with `fontHandles->library` and `fontHandles->stroker`?

src/SFML/Graphics/Font.cpp
vittorioromeo: This is poorly documented, so I had to check the implementation. 

- `FT_Stroker_Done` is safe and returns void.

- `FT_Done_Face` is safe, it returns non-zero for a null argument.

- `FT_Done_FreeType` is safe, it returns non-zero for a null argument.

src/SFML/Graphics/Font.cpp
vittorioromeo: It has always been a `std::shared_ptr`, this PR only makes it more obvious by avoiding a `unique_ptr` to `shared_ptr` conversion. 

I investigated a bit:

- `m_fontHandles` does get copied when `Font` is copied, implying shared semantics.

- Loading the same font multiple times (e.g. using the same path) re-creates a new `FontHandles` shared object. This might be an opportunity for optimization (i.e. caching the `FontHandles` shared pointer).

src/SFML/Graphics/Font.cpp
Bromeon: Thanks a lot for checking! Could you add this as a comment, so we don't have to re-check in the future? 🙂

src/SFML/Graphics/Font.cpp
```diff
@@ -41,6 +41,7 @@
 #include FT_OUTLINE_H
 #include FT_BITMAP_H
 #include FT_STROKER_H
+#include <cassert>
```
MarioLiebisch, line 44: Is this a leftover from testing or sometimes I'm missing? I don't see any new asserts and it compiles fine for me without.

vittorioromeo: It is -- will remove it.

ChrisThrasher (approved):
I don't feel qualified to thoroughly review this but I trust that this is a worthwhile change. I'll others make the final call on when this is ready to merge.

ChrisThrasher:
CI formatting jobs are in conflict which is rare. Long story short, you can fix this by adding our typical Doxygen headers above each and ever special member function. Then all the formatting jobs will be in agreement.

vittorioromeo:
> CI formatting jobs are in conflict which is rare. Long story short, you can fix this by adding our typical Doxygen comment blocks above each and every special member function. Then all the formatting jobs will be in agreement.

That would be overly verbose IMHO, let me try with `// clang-format off`

vittorioromeo:

MarioLiebisch (approved):

PR #2522, opened 2023-04-15
Enable clang-tidy `modernize-pass-by-value` check

ChrisThrasher:
## Description

https://clang.llvm.org/extra/clang-tidy/checks/modernize/pass-by-value.html

This ensures that we're not copying potentially large types more often than necessary.

### DIFF
--- a/examples/shader/Effect.hpp
+++ b/examples/shader/Effect.hpp
@@ -7,6 +7,7 @@
 
 #include <cassert>
 #include <string>
+#include <utility>
 
 
 ////////////////////////////////////////////////////////////
@@ -54,7 +55,7 @@ public:
     }
 
 protected:
-    Effect(const std::string& name) : m_name(name)
+    Effect(std::string name) : m_name(std::move(name))
     {
     }
 
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -30,7 +30,6 @@ Checks: >
   -misc-non-private-member-variables-in-classes,
   -modernize-avoid-c-arrays,
   -modernize-macro-to-enum,
-  -modernize-pass-by-value,
   -modernize-return-braced-init-list,
   -modernize-use-nodiscard,
   -modernize-use-trailing-return-type,
--- a/include/SFML/Graphics/Text.hpp
+++ b/include/SFML/Graphics/Text.hpp
@@ -77,13 +77,13 @@ public:
     /// \param characterSize  Base size of characters, in pixels
     ///
     ////////////////////////////////////////////////////////////
-    Text(const Font& font, const String& string = "", unsigned int characterSize = 30);
+    Text(const Font& font, String string = "", unsigned int characterSize = 30);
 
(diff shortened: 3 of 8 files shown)


binary1248 (approved):

vittorioromeo:
Not 100% convinced about this because for some types this can be a pessimization, especially strings. 

Given a string type supporting SBO (small buffer optimization), taking parameters of such types by value will be twice as expensive for strings fitting the small buffer, as you would have to copy them twice as opposed to binding them to a `const&` and then copying them once.

Would this be an issue in our case?

ChrisThrasher:
> Given a string type supporting SBO (small buffer optimization), taking parameters of such types by value will be twice as expensive for strings fitting the small buffer, as you would have to copy them twice as opposed to binding them to a `const&` and then copying them once.

That's a good point. I can think of two different perspective for making a decision.

1. Pick moves or copies depending on which is more commonly used with these APIs. I have no way of knowing whether "big" or "small" strings are more commonly used with these APIs so I'm flying in the dark here. This implies that we cannot add this clang-tidy check because we're stating that we need to exercise discretion. Maybe we make this change in select places but we would not _necessarily_ use this pattern.
2. Optimize for the worst case. This implies we'd rather move to avoid copying arbitrarily large strings or other types. Maybe this isn't ideal for small strings but it means we're not unnecessarily copying. This has the benefit of letting us enable this check so that we're less likely to forget to use this pattern in the future.

Both perspectives are valid in different ways, we just need to decide what we care about.

Bromeon (approved):

Bromeon:
> Given a string type supporting SBO (small buffer optimization), taking parameters of such types by value will be twice as expensive for strings fitting the small buffer, as you would have to copy them twice as opposed to binding them to a `const&` and then copying them once.

Copying the SBO string would only `memcpy` at most `sizeof(string)` bytes, no? So it's similarly expensive to a move constructor for large strings?

I see that we could avoid one such move with `const String&`, but then we have the extra dereferencing. With compiler optimizations, both might again look different, so it's probably hard to say without godbolting or measuring it 🤔

vittorioromeo:
> Copying the SBO string would only `memcpy` at most `sizeof(string)` bytes, no? So it's similarly expensive to a move constructor for large strings?

Taking a SBO string by value requires **two** `memcpy` invocations instead of one (if the string were to be taken by `const&`). That's my counter-argument to this PR, it might be an unnecessary performance pessimization if most strings fit the SBO.

examples/shader/Effect.hpp
```diff
@@ -54,7 +54,7 @@ class Effect : public sf::Drawable
     }
 
 protected:
-    Effect(const std::string& name) : m_name(name)
+    Effect(std::string name) : m_name(std::move(name))
```
vittorioromeo, line 57: `<utility>` needs to be explicitly included everywhere `std::move` is used.

ChrisThrasher:

Issue #2483, opened 2023-03-31
Finish supporting move semantics


ChrisThrasher:
Here are the types that still lack move semantics or have incomplete support for it. Some of these types may not apply but they're listed regardless until we decide to not bother.

The addition of move operations constitutes a new feature and not an API break so we're free to keep working on this after 3.0.0 ships.

Window:
- [ ] `sf::Context` (https://github.com/SFML/SFML/pull/2805)
- [x] `sf::Cursor` (https://github.com/SFML/SFML/pull/2487 https://github.com/SFML/SFML/pull/2525)
- [ ] `sf::GlResource`
- [ ] `sf::Window` (https://github.com/SFML/SFML/pull/2529)
- [ ] `sf::WindowBase` (https://github.com/SFML/SFML/pull/2529)

Graphics:
- [x] `sf::Drawable`
- [x] `sf::RenderTarget` (https://github.com/SFML/SFML/pull/2656)
- [x] `sf::RenderTexture` (https://github.com/SFML/SFML/pull/2656)
- [x] `sf::VertexBuffer`

Network:
- [ ] `sf::Ftp`
- [ ] `sf::Http`
- [x] `sf::Socket` (https://github.com/SFML/SFML/pull/2658)
- [x] `sf::SocketSelector`
- [x] `sf::TcpListener` (https://github.com/SFML/SFML/pull/2658)
- [x] `sf::TcpSocket` (https://github.com/SFML/SFML/pull/2658)
- [x] `sf::UdpSocket` (https://github.com/SFML/SFML/pull/2658)

Audio:
- [ ] `sf::AlResource` (https://github.com/SFML/SFML/pull/2541)
- [ ] `sf::Music`
- [ ] `sf::Sound`
- [ ] `sf::SoundBuffer`
- [ ] `sf::SoundBufferRecorder`
- [ ] `sf::SoundRecorder`
- [ ] `sf::SoundSource`
- [ ] `sf::SoundStream`

shakedechn:
Hey @ChrisThrasher. How would a contributor go about taking one of these? Should I ask here to take one and declare which one, or is there a dedicated place on the forums?

ChrisThrasher:
You don't need to ask for permission. Start working on a class and if you think you have a viable implementation then submit a PR.

boothr16:
[...]

PR #2525, opened 2023-04-19
fixed move semantics for sf::Cursor and added test for it.

alirezafour:
## Description

because of the forward declaration of `sf::priv::CursorImpl` on header file implementation of move constructor and move assignment operator on header file cause the code not to compile because `std::unique_ptr` need the implementation of `sf::priv::CursorImpl` to be able to delete it properly.

I moved the implementation into `Cursor.cpp` and added test for it.

----

----
## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>
#include <SFML/Window/Cursor.hpp>

int main()
{
    window.setFramerateLimit(60);
    sf::Cursor cursor(std::move(sf::Cursor()));


	if (cursor.loadFromSystem(sf::Cursor::Hand))


    while (window.isOpen())
    {
        {
                window.close();
        }

        window.clear();
        window.display();
    }
}
```


eXpl0it3r:

ChrisThrasher:
Thank you for fixing this! I think I see what's going on now. I assumed that 

```cpp
static_assert(std::is_nothrow_move_constructible_v<sf::Cursor>);
static_assert(std::is_nothrow_move_assignable_v<sf::Cursor>);
```

was enough to verify that moving a Cursor would compile but I now see that that we also need to compile code that moves a given type to be sure that moving that type will actually compiler.

I see a few increasingly cumbersome steps for verifying correct move semantics. Not all types need to take all steps but some types need more thorough testing than others.

1. Assert the correct type traits for move semantics (we already do this for all types) (good enough for types following the Rule of Zero)
1. Compile tests that perform move construction and move assignment (good enough for types whose internals allow for `= default`'ed move operations) (this is often done by our examples which use large swaths of the API but not all of the API)
1. Run said tests to confirm nothing crashes (this isn't possible for types like `sf::Cursor` for which we can't run the tests thanks to incomplete DRM support)
1. Assert some post condition to confirm the move worked as intended (this is difficult when many types are too opaque to inspect the internals to confirm the move operation succeeded)
1. Perform non-automated tested with a sample project to confirm the move operations are correct (we have done this before but it's very expensive)

test/Window/Cursor.test.cpp
```diff
@@ -19,6 +19,19 @@ TEST_CASE("[Window] sf::Cursor" * doctest::skip(true))
...
+        const sf::Cursor movedCursor(std::move(cursor));
+    }
+
+    SUBCASE("Move assigment operator")
+    {
+        sf::Cursor cursor;
+        sf::Cursor movedCursor;
+        movedCursor = std::move(cursor);
+    }
+
```
ChrisThrasher, line 34:
```suggestion
    SUBCASE("Move semantics")
    {
        SUBCASE("Construction")
        {
            sf::Cursor       movedCursor;
            const sf::Cursor cursor(std::move(movedCursor));
        }

        SUBCASE("Assignment")
        {
            sf::Cursor movedCursor;
            sf::Cursor cursor;
            cursor = std::move(movedCursor);
        }
    }

```

ChrisThrasher (changes requested):
Other than a minor style issue this is good to go

ChrisThrasher:

ChrisThrasher (approved):
Thank you!

vittorioromeo:
@ChrisThrasher: this might be of interest: https://stackoverflow.com/questions/51901837/how-to-get-if-a-type-is-truly-move-constructible/51912859#51912859

eXpl0it3r (approved):

eXpl0it3r:
Thanks! 🙂

PR #2529, opened 2023-04-22
Move semantics for `sf::Window`, `sf::WindowBase`

alirezafour:
## Description

Move semantics implementation for `sf::Window`, `sf::WindowBase`.

----

----

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Window.hpp>

int main()
{
	sf::Window window;
	{
		sf::Window movedWindow(sf::VideoMode({ 1280, 720 }), "Window Title");
		movedWindow.setFramerateLimit(60);
		window = std::move(movedWindow);
	}

	while (window.isOpen())
	{
		{
				window.close();
		}

		window.display();
	}
}
```


ChrisThrasher:

src/SFML/Window/Window.cpp
```diff
@@ -61,6 +61,17 @@ Window::~Window()
 }
 
 
+////////////////////////////////////////////////////////////
+Window::Window(Window&& other) noexcept
+{
+    *this = std::move(other);
+}
```
ChrisThrasher, line 68: Why isn't this defaulted?

test/Window/WindowBase.test.cpp
```diff
@@ -29,6 +29,27 @@ TEST_CASE("[Window] sf::WindowBase" * doctest::skip(skipDisplayTests))
...
+            }
+
+            SUBCASE("Assignment")
+            {
+                sf::WindowBase windowBase;
+                {
+                    sf::WindowBase movedWindowBase(sf::VideoMode({360, 240}), "WindowBase Tests");
+                    windowBase = std::move(movedWindowBase);
+                }
+                CHECK(windowBase.getSize() == sf::Vector2u(360, 240));
```
ChrisThrasher, line 49:
```suggestion
                sf::WindowBase movedWindowBase(sf::VideoMode({360, 240}), "WindowBase Tests");
                sf::WindowBase windowBase;
                windowBase = std::move(movedWindowBase);
                CHECK(windowBase.getSize() == sf::Vector2u(360, 240));
```
I think it's fine to let the moved-from window stay in scope for the sake of keeping the tests simple.

src/SFML/Window/Window.cpp
alirezafour: the default move constructor was also deleting `shareContext` after goes out of scope. for testing when the previous test goes out of scope the next window creating fail assert finding `shareContext` on creating new window.  
you can test the default move constructor with the test below. the second `movedWindow` find `shareContext` nullptr while creating `m_context`.   
It seems the default move constructor gives the ownership of the `shareContext` to the `sf::Window` class and it gets deleted after goes out of scope.  
  
```c++
SUBCASE("Construction")
{
    sf::Window       movedWindow(sf::VideoMode(sf::Vector2u(256, 256), 32), "Window Title");
    const sf::Window window(std::move(movedWindow));
    CHECK(window.getSize() == sf::Vector2u(256, 256));
}

SUBCASE("Assignment")
{
    sf::Window movedWindow(sf::VideoMode(sf::Vector2u(256, 256), 32), "Window Title");
    sf::Window window;
    window = std::move(movedWindow);
    CHECK(window.getSize() == sf::Vector2u(256, 256));
}
```

kimci86: Couldn't it be cleaner by making GlResource class movable?

ChrisThrasher: I always prefer making base classes moveable first if possible since you can end up with weird implementations if base classes or class members are not moveable.

alirezafour: thanks, I did that.
the problem was `resourceCount` was going to negative since the moved object was still exist and call destructor.

include/SFML/Window/GlResource.hpp
```diff
@@ -53,6 +66,12 @@ class SFML_WINDOW_API GlResource
     ////////////////////////////////////////////////////////////
     ~GlResource();
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Move constructor
+    ///
+    ////////////////////////////////////////////////////////////
+    GlResource(GlResource&&) noexcept = default;
```
kimci86, line 73: `default` does nothing here. This is not correct. We need to steal ownership of the implicit link with `GlContext` so that reference count remains correct.

alirezafour: `GlResource` doesn't own any `GlContext`. Its constructor and destructor only create global variable of `sharedContext` if not available and it delete them if `recourseCount` reach 0. the `GlContext` lives in Window class that is moved.

src/SFML/Window/GlResource.cpp
```diff
@@ -38,6 +38,13 @@ GlResource::GlResource()
 }
 
 
+////////////////////////////////////////////////////////////
+GlResource::GlResource(GlResource&&) noexcept
+{
+    priv::GlContext::initResource();
```
kimci86, line 44: Nice, this is a solution to get `resourceCount` updated correctly.
However, a mutex lock is involved, which might throw according to [cppreference](https://en.cppreference.com/w/cpp/thread/recursive_mutex/lock), so I am wondering if it is suitable for a `noexcept` function.

alirezafour: I am still new to source code, but wouldn't be more cleaner to have a singleton class to manage `resourceCount` and other global variable related? if it was like that we had a ref to the singleton class in `GlResource` that would be moved and we did not needed to increase `resourceCount` to 2 and moved object doesn't have valid ref so it couldn't do anything when goes out of scope.

src/SFML/Window/WindowBase.cpp
```diff
@@ -70,6 +70,14 @@ WindowBase::~WindowBase()
 }
 
 
+////////////////////////////////////////////////////////////
+WindowBase::WindowBase(WindowBase&&) noexcept = default;
```
shakedechn, line 74: Wouldn't you need to address `WindowsBaseImpl::fullscreenWindow`?
If you move an object that this global variable currently points to, the new object would not take its place.

src/SFML/Window/GlResource.cpp
alirezafour: @kimci86 I removed `noexcept` but static analyzer is unhappy.
what do you suggest ?

kimci86: Maybe there is a smarter way but one solution I can think of is to keep `noexcept` on move operations and add a boolean variable in `GlResource` telling if `priv::GlContext::cleanupResource` needs to be called in `GlResource` destructor. This makes possible to implement move operations with simple boolean variable assignments.

src/SFML/Graphics/Texture.cpp
```diff
@@ -105,6 +106,7 @@ Texture::~Texture()
 
 ////////////////////////////////////////////////////////////
 Texture::Texture(Texture&& right) noexcept :
+GlResource(right),
```
kimci86, line 109:
```suggestion
GlResource(std::move(right)),
```

src/SFML/Window/GlResource.cpp
kimci86: Actually it will be cleaner to wait for #2503 to get merged as it adds move semantics for `GlResource`.

test/Window/Window.test.cpp
```diff
@@ -28,4 +28,26 @@ TEST_CASE("[Window] sf::Window", runDisplayTests())
                                 sf::ContextSettings());
         CHECK(window.getSize() == sf::Vector2u(256, 256));
     }
+
+    SECTION("Move construction")
+    {
+        sf::Window       movedWindow(sf::VideoMode(sf::Vector2u(256, 256), 32),
```
kimci86, line 34: Maybe the video mode size could be defined as `{256, 256}` to be consistent with what is done in `WindowBase.test.cpp`? If we do so, other test cases in `Window.test.cpp` would be changed too.

kimci86 (approved):
LGTM besides minor style comments.

src/SFML/Window/Window.cpp
```diff
@@ -55,13 +55,21 @@ Window::Window(WindowHandle handle, const ContextSettings& settings)
...
 
 
+////////////////////////////////////////////////////////////
+sf::Window::Window(Window&&) noexcept = default;
+
+
 ////////////////////////////////////////////////////////////
 Window::~Window()
 {
     close();
```
ChrisThrasher, line 65: Typically when there is a non-default destructor we need non-default move operations. How are you sure the default operations are the correct choice?

ChrisThrasher:
Please rebase on `master`. We've merged a lot since this PR was opened.

ChrisThrasher:
Please rebase instead of merging. If you don't know how to do that I can do it for you.

ChrisThrasher:

test/Window/WindowBase.test.cpp
```diff
@@ -48,6 +48,25 @@ TEST_CASE("[Window] sf::WindowBase", runDisplayTests())
...
             CHECK(windowBase.getSystemHandle() != sf::WindowHandle());
         }
+
+        SECTION("Move construction")
+        {
+            sf::WindowBase       movedWWindowBase(sf::VideoMode({360, 240}), "WindowBase Tests");
+            const sf::WindowBase windowBase(std::move(movedWWindowBase));
+            CHECK(windowBase.isOpen());
+            CHECK(windowBase.getSize() == sf::Vector2u(360, 240));
+            CHECK(windowBase.getSystemHandle() != sf::WindowHandle());
```
ChrisThrasher, line 58: Heads up, we renamed `getSystemHandle` to `getNativeHandle`.

ChrisThrasher:
I rebased and fixed an instance of an API changing after this PR was opened.

ChrisThrasher:

src/SFML/Window/Window.cpp
alirezafour: yeah, it is problematic because of the fullscreen and future changes on close function. I will take a look and fix it.

test/Window/WindowBase.test.cpp
alirezafour: Thanks, I will fix it. and learn the rebase as you mentioned. 👍

ChrisThrasher:
Run `git fetch --all` then `git pull --rebase` to update your local branch without having to create another merge commit. I won't touch the branch after this. I just wanted to get that rebase done since CI has improved a lot and I wanted to make sure those improvements were being used.

src/SFML/Window/Window.cpp
ChrisThrasher: I think we need to make sure that if the window is fullscreen that the address of the new window is used as the address of the current fullscreen window instead of the moved-from window. Other than that I think we can rely on the default move semantics for the remaining data members of `sf::WindowBase` and `sf::Window`.

src/SFML/Window/Window.cpp
```diff
@@ -58,7 +58,10 @@ Window::Window(WindowHandle handle, const ContextSettings& settings)
 
 
 ////////////////////////////////////////////////////////////
-sf::Window::Window(Window&&) noexcept = default;
+sf::Window::Window(Window&& other) noexcept : WindowBase(std::move(other)), GlResource(std::move(other))
+{
+    other.close();
```
ChrisThrasher, line 63: I think we can retain default move operations on `sf::Window`, right? When moved-from the `unique_ptr` inside is set to null which is already all of what `sf::Window::close` was doing before. The move constructor for the base class ought to be automatically invoked. I'm not totally sure about this all though.

src/SFML/Window/Window.cpp
alirezafour: I did the fullscreen check on `sf::WindowBase`  move constructor and call close in both move after that for future safety.

ChrisThrasher: I'd rather not preemptively add code we don't know if we'll ever need.

src/SFML/Window/Window.cpp
alirezafour: I added only added close for future maybe something else added in close function. but you are right it can also be default.

src/SFML/Window/WindowBase.cpp
```diff
@@ -68,13 +69,28 @@ WindowBase::WindowBase(WindowHandle handle)
...
+
 ////////////////////////////////////////////////////////////
 WindowBase::~WindowBase()
 {
     WindowBase::close();
 }
 
 
+////////////////////////////////////////////////////////////
+WindowBase& WindowBase::operator=(WindowBase&&) noexcept = default;
```
ChrisThrasher, line 91: Does this need to include similar logic for setting the full screen window? I would assume so.

ChrisThrasher:

I made some pedantic cleanup changes. If CI passes then I think I'm happy with this PR. I still want some more feedback from others though to make sure we haven't missed anything.

ChrisThrasher:

src/SFML/Window/WindowBase.cpp
```diff
@@ -75,6 +76,37 @@ WindowBase::~WindowBase()
...
+WindowBase::WindowBase(WindowBase&& windowBase) noexcept :
+m_impl(std::exchange(windowBase.m_impl, nullptr)),
+m_size(std::exchange(windowBase.m_size, {}))
+{
+    // Check fullscreen style and set to new object
+    if (&windowBase == getFullscreenWindow())
+        setFullscreenWindow(this);
+
+    windowBase.close();
+}
```
ChrisThrasher, line 89:
```suggestion
WindowBase::WindowBase(WindowBase&& windowBase) noexcept
{
    *this = std::move(windowBase);
}
```
I'm not totally sure if this is correct and a good idea but we can maybe implement move construction in terms of move assignment and thus deduplicate some code.

alirezafour: the different is in my code resources constructed with new values but in suggested code they constructed with default and then re-assigned.

include/SFML/Window/Window.hpp
```diff
@@ -110,6 +110,30 @@ class SFML_WINDOW_API Window : public WindowBase, GlResource
...
+    /// \brief Deleted copy constructor
+    ///
+    ////////////////////////////////////////////////////////////
+    Window(const Window&) = delete;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Deleted copy assignment
+    ///
+    ////////////////////////////////////////////////////////////
+    Window& operator=(const Window&) = delete;
```
kimci86, line 123: Deleting those special methods manually is not technically required.

src/SFML/Window/WindowBase.cpp
```diff
@@ -75,6 +76,37 @@ WindowBase::~WindowBase()
...
+////////////////////////////////////////////////////////////
+WindowBase::WindowBase(WindowBase&& windowBase) noexcept :
+m_impl(std::exchange(windowBase.m_impl, nullptr)),
+m_size(std::exchange(windowBase.m_size, {}))
+{
+    // Check fullscreen style and set to new object
+    if (&windowBase == getFullscreenWindow())
+        setFullscreenWindow(this);
+
+    windowBase.close();
```
kimci86, line 88: In the case of a derived class with additional data, calling this virtual method might release resources before having a chance to move them to the new instance. I confirmed with a debugger this actually happens with `m_context` in `sf::Window` class. I suggest not calling this method.
```suggestion
```

src/SFML/Window/WindowBase.cpp
```diff
@@ -75,6 +76,37 @@ WindowBase::~WindowBase()
...
+        return *this;
+
+    m_impl = std::exchange(windowBase.m_impl, nullptr);
+    m_size = std::exchange(windowBase.m_size, {});
+
+    // Check fullscreen style and set to new object
+    if (&windowBase == getFullscreenWindow())
+        setFullscreenWindow(this);
+
+    windowBase.close();
```
kimci86, line 105: Same reason as move constructor:
```suggestion
```

kimci86 (changes requested):

kimci86:
Thinking more about the virtual functions involved, it might be a bad idea to have public move semantics for WindowBase and Window. See [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c67-a-polymorphic-class-should-suppress-public-copymove).

L0laapk3:
Found this issue since having move semantics on `RenderWindow` would be useful for me.

> Thinking more about the virtual functions involved, it might be a bad idea to have public move semantics for WindowBase and Window. See [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c67-a-polymorphic-class-should-suppress-public-copymove).

One solution to this offered [here](https://stackoverflow.com/questions/70336995/how-to-best-define-copy-move-constructors-assignment-operators-in-a-derived-clas), is to make the move semantic in the base classes `protected`. Interested in hearing your thoughts.

ChrisThrasher:
I found a much simpler way to implement move semantics for windows in PR #3142

Issue #2490, opened 2023-04-04
Header organization rules -- a coding style suggestion


vittorioromeo:
We currently organize headers in only two groups, and then let `clang-format` alphabetically sort within groups. I suggest a refinement of this technique as our coding style.

The rules I suggest are as follows:

1. Headers are organized in groups.
2. Within a group, headers are sorted alphabetically.
3. Header groups are separated by a newline.
4. The groups themselves are ordered from higher level to lower level, in terms of dependencies.

For example, this is a possible ordering of groups:

1. `sfml-graphics`
2. `sfml-audio`, `sfml-network`, `sfml-window` *(the order between these is alphabetical)*
3. `sfml-system`
4. Third-party libraries
5. Platform libraries
6. C++ Standard Library 
7. C Standard Library

Here's an example of these rules in action:

```cpp
#include <SFML/Graphics/Font.hpp> // GROUP 1
#include <SFML/Graphics/Text.hpp>
#include <SFML/Graphics/RenderWindow.hpp>

#include <SFML/Audio/Music.hpp> // LEVEL 2

#include <SFML/Window/Event.hpp> // LEVEL 2
#include <SFML/Window/Mouse.hpp>
#include <SFML/Window/Keyboard.hpp>

#include <SFML/System/Clock.hpp> // LEVEL 3
#include <SFML/System/Time.hpp>

#include <boost/stacktrace.hpp> // LEVEL 4
#include <glad/gl.h>

#include <windows.h> // LEVEL 5

#include <filesystem> // LEVEL 6
#include <map>
#include <vector>

#include <cstdlib> // LEVEL 7
#include <cstdio>
```

Reasons:

- Clear separation of headers belonging to the same group/level
- Visual indication of the hierarchy of dependencies
- Including from the most dependent to the least dependent helps to catch transitive includes and prevent regressions

vittorioromeo:
Here's an example of how https://www.sfml-dev.org/documentation/2.5.1/ would look using these rules, and also applying the ideas in #2489:

```cpp
#include <SFML/Graphics/Font.hpp>
#include <SFML/Graphics/RenderWindow.hpp>
#include <SFML/Graphics/Sprite.hpp>
#include <SFML/Graphics/Text.hpp>
#include <SFML/Graphics/Texture.hpp>

#include <SFML/Audio/Music.hpp>

#include <SFML/Window/Event.hpp>
#include <SFML/Window/VideoMode.hpp>

int main()
{
    // Create the main window
[...]

PR #2535, opened 2023-04-23
Reorder includes hierarchically

vittorioromeo:
Automatically implements the suggestions described in #2490 as closely as possible through changes to `.clang-format`.


ChrisThrasher:
> Formatted via clang-format version 16.0.1.

This may be a problem. clang-format versions newer than v14 result in differently formatted code particularly where vertical alignment is happening.

ChrisThrasher (commented):
I'm not sure what this buys us. Looking at the diff we appear to be inserting many more empty newlines but otherwise we have no issues with header ordering that need solving. I'm afraid I'm not onboard with this. Exploiter may have to make the final call.

vittorioromeo:
> I'm not sure what this buys us. Looking at the diff we appear to be inserting many more empty newlines but otherwise we have no issues with header ordering that need solving. I'm afraid I'm not onboard with this. Exploiter may have to make the final call.

Check https://github.com/SFML/SFML/issues/2490 for rationale. 

> - Clear separation of headers belonging to the same group/level
> - Visual indication of the hierarchy of dependencies
> - Including from the most dependent to the least dependent helps to catch transitive includes and prevent regressions

vittorioromeo:
> > Formatted via clang-format version 16.0.1.
> 
> This may be a problem. clang-format versions newer than v14 result in differently formatted code particularly where vertical alignment is happening.

I guess I can reformat everything with v14. Is there a good reason why we are not using v16?

ChrisThrasher:
> Check https://github.com/SFML/SFML/issues/2490 for rationale.

I read that rationale. I just wasn't convinced, sorry. Because this is an automated change that won't add any extra work to my workflow I can mostly forget about it so I'm not veto'ing it.

> I guess I can reformat everything with v14.

You have to reformat with v14 (or 12 or 13) otherwise the CI jobs will fail. That's a hard requirement I'm afraid. I'm guessing you're using the Visual Studio clang-format integration instead of our `format` target because our `format` target detects and forbids the use of anything but versions 12, 13, or 14. You should consider using our in-house formatting solution.

> Is there a good reason why we are not using v16?

No one has proposed upgrading clang-format version(s) since the original clang-format PR was merged. I'm open to upgrading if someone wants to lead that effort.

eXpl0it3r:
Can you please rebase it instead of merging master into it?

vittorioromeo:

vittorioromeo:
> Can you please rebase it instead of merging master into it?

Done :) 

Are you happy to approve?

eXpl0it3r:
Rebased to ensure that the latest commit doesn't cause the merge to fail

vittorioromeo:
@ChrisThrasher: could you run clang-format 14 on the PR again, please?

ChrisThrasher:
> @ChrisThrasher: could you run clang-format 14 on the PR again, please?

https://github.com/llvm/llvm-project/releases/tag/llvmorg-14.0.6

I downloaded clang-format-14 from here if you'd like to do the same.

eXpl0it3r:
I'm merging this now, might still be interesting to have a comment on #2490 why you don't like this style change, @ChrisThrasher

PR #2536, opened 2023-04-24
Add `sf::Shape::getGeometricCenter()`

TedLyngmo:
This adds

    virtual Vector2f sf::Shape::getGeometricCenter() const = 0;

as discussed here:
https://en.sfml-dev.org/forums/index.php?topic=29011

One practical use can be to set a rotation point in the middle of a circle, rectangle or convex shape.

Example:
```cpp
#include <SFML/Graphics.hpp>
#include <initializer_list>

class Boulder : public sf::ConvexShape {
public:
    inline Boulder(std::initializer_list<sf::Vector2f> pnts) :
        sf::ConvexShape(pnts.size())
    {
        for(std::size_t idx = 0; idx < pnts.size(); ++idx)
        {
            setPoint(idx, pnts.begin()[idx]);
        }

        // Set the rotation point at the geometrical center of the convex shape:
        setOrigin(getGeometricCenter());
    }
};

int main()
{
    window.setFramerateLimit(60);

    Boulder boulder({{20,0}, {60,0}, {80,40}, {70,60}, {30,80}, {10,60}, {0,30}});
    boulder.setPosition({1280/2, 720/2});

    while (window.isOpen())
    {
        {
                window.close();
        }

        boulder.rotate(sf::degrees(2));

        window.clear();
        window.draw(boulder);
        window.display();
    }
}
```

----

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android


include/SFML/Graphics/RectangleShape.hpp
```diff
@@ -92,6 +92,17 @@ class SFML_GRAPHICS_API RectangleShape : public Shape
     ///
     ////////////////////////////////////////////////////////////
     Vector2f getPoint(std::size_t index) const override;
+    ////////////////////////////////////////////////////////////
```
vittorioromeo, line 95:
```suggestion

    ////////////////////////////////////////////////////////////
```

src/SFML/Graphics/RectangleShape.cpp
```diff
@@ -78,4 +78,10 @@ Vector2f RectangleShape::getPoint(std::size_t index) const
     }
 }
 
+////////////////////////////////////////////////////////////
+Vector2f RectangleShape::getGeometricCenter() const
+{
+    return Vector2f(m_size.x / 2.f, m_size.y / 2.f);
```
vittorioromeo, line 84:
```suggestion
    return m_size / 2.f;
```

src/SFML/Graphics/ConvexShape.cpp
```diff
@@ -66,4 +66,44 @@ Vector2f ConvexShape::getPoint(std::size_t index) const
...
+        f = twicearea * 3.f;
+
+        return Vector2f(x / f + offx, y / f + offy);
+    }
+
+    if (m_points.size() == 2)
+    { // a line
+        const auto& [x1, y1] = m_points.front();
+        const auto& [x2, y2] = m_points.back();
+        return Vector2f((x1 + x2) / 2.f, (y1 + y2) / 2.f);
```
vittorioromeo, line 100:
```suggestion
        return (m_points[0] + m_points[1]) / 2.f;
```

src/SFML/Graphics/ConvexShape.cpp
```diff
@@ -66,4 +66,44 @@ Vector2f ConvexShape::getPoint(std::size_t index) const
...
+    }
+
+    if (not m_points.empty())
+        return m_points.front(); // a single point
```
vittorioromeo, line 104:
```suggestion
    if (!m_points.empty())
        return m_points.front(); // a single point
```

src/SFML/Graphics/ConvexShape.cpp
```diff
@@ -66,4 +66,44 @@ Vector2f ConvexShape::getPoint(std::size_t index) const
...
+
+    return Vector2f{}; // empty
```
vittorioromeo, line 106: Not too sure about this one. It's impossible to differentiate between a polygon whose geometrical center is in `0,0` and a polygon with no points. Maybe this function should return `std::optional<Vector2f>`?

src/SFML/Graphics/CircleShape.cpp
```diff
@@ -75,4 +75,12 @@ Vector2f CircleShape::getPoint(std::size_t index) const
     return Vector2f(m_radius, m_radius) + Vector2f(m_radius, angle);
 }
 
+////////////////////////////////////////////////////////////
+Vector2f CircleShape::getGeometricCenter() const
+{
+    if (m_pointCount == 3)
+        return (getPoint(0) + getPoint(1) + getPoint(2)) / 3.f;
```
vittorioromeo, line 82: What's the significance of special-casing for 3 points?

src/SFML/Graphics/ConvexShape.cpp
```diff
@@ -66,4 +66,44 @@ Vector2f ConvexShape::getPoint(std::size_t index) const
...
+        for (auto it = m_points.begin(), prev = std::prev(m_points.end()); it != m_points.end(); prev = it++)
+        {
+            const auto& [p1x, p1y] = *it;
+            const auto& [p2x, p2y] = *prev;
+
+            f = (p1x - offx) * (p2y - offy) - (p2x - offx) * (p1y - offy);
+            twicearea += f;
+            x += (p1x + p2x - 2.f * offx) * f;
+            y += (p1y + p2y - 2.f * offy) * f;
+        }
```
vittorioromeo, line 89:
```suggestion
        for (std::size_t i = 0; i < m_points.size() - 1; ++i)
        {
            const auto& [p1x, p1y] = m_points[i];
            const auto& [p2x, p2y] = m_points[i + 1];

            f = (p1x - offx) * (p2y - offy) - (p2x - offx) * (p1y - offy);
            twicearea += f;
            x += (p1x + p2x - 2.f * offx) * f;
            y += (p1y + p2y - 2.f * offy) * f;
        }
```

Honestly I find an index-based version easier and simpler to read. Also, we should use vector operations in the body (i.e. instead of manually acting on `x` and `y`, use a `Vector2f` for your result).

vittorioromeo (changes requested):
Indifferent on whether we should add this to SFML or not in general

src/SFML/Graphics/ConvexShape.cpp
TedLyngmo: That's an option or it could throw an exception or leave the case UB

src/SFML/Graphics/CircleShape.cpp
TedLyngmo: @vittorioromeo It's the one case where the geometric center is not @ `Vector2f(m_radius, m_radius)`.

src/SFML/Graphics/ConvexShape.cpp
TedLyngmo: @vittorioromeo This will however not include the line between the last point and the first point - but I can rewrite it using idices instead.

src/SFML/Graphics/ConvexShape.cpp
TedLyngmo: @Bromeon  Sounds like a good idea. What to do in `NDEBUG` builds? `std::terminate`?

TedLyngmo:
@vittorioromeo & @Bromeon Would it perhaps be better to place the generic algorithm in `Shape` instead of `ConvexShape`?

TedLyngmo:
> Which PR do you want to keep open. This one or #2537? I don't think we need both.

I wasn't sure if you'd be ok with adding the implementation to the `Shape` class so I made two separate patches - but it seems we all agree that the other patch is better so I'll close this.

PR #2537, opened 2023-04-24
Add `sf::Shape::getGeometricCenter()`

TedLyngmo:
This adds

    virtual Vector2f sf::Shape::getGeometricCenter() const;

as discussed here:
https://en.sfml-dev.org/forums/index.php?topic=29011

It's similar to https://github.com/SFML/SFML/pull/2536 but the generic implementation is here placed in `sf::Shape` instead of in `sf::ConvexShape`.

One practical use can be to set a rotation point in the middle of a circle, rectangle or convex shape.

Example:
```cpp
#include <SFML/Graphics.hpp>

#include <initializer_list>

class Boulder : public sf::ConvexShape {
public:
    inline Boulder(std::initializer_list<sf::Vector2f> pnts) :
        sf::ConvexShape(pnts.size())
    {
        for(std::size_t idx = 0; idx < pnts.size(); ++idx)
        {
            setPoint(idx, pnts.begin()[idx]);
        }

        // Set the rotation point at the geometrical center of the convex shape:
        setOrigin(getGeometricCenter());
    }
};

int main()
{
    window.setFramerateLimit(60);

    Boulder boulder1({{20,0}, {60,0}, {80,40}, {70,60}, {30,80}, {10,60}, {0,30}});
    boulder1.setPosition({1280/3, 720/3});

    Boulder boulder2({{20,0}, {60,0}, {180,40}, {70,60}, {30,80}, {10,60}, {0,30}});
    boulder2.setPosition({1280*2/3, 720/3});

    sf::RectangleShape rectangle({100,50});
    rectangle.setOrigin(rectangle.getGeometricCenter());
    rectangle.setPosition({1280/2, 720*2/3});

    sf::CircleShape circle1(40); // center slightly off to wobble a bit:
    circle1.setOrigin(circle1.getGeometricCenter() + sf::Vector2(5.f,5.f));
    circle1.setPosition({1280/3, 720*2/3});

    sf::CircleShape circle2(40,3);
    circle2.setOrigin(circle1.getGeometricCenter());
    circle2.setPosition({1280*2/3, 720*2/3});

    while (window.isOpen())
    {
        {
                window.close();
        }

        boulder1.rotate(sf::degrees(2));
        boulder2.rotate(sf::degrees(3));
        rectangle.rotate(sf::degrees(4));
        circle1.rotate(sf::degrees(5));
        circle2.rotate(sf::degrees(5));

        window.clear();
        window.draw(boulder1);
        window.draw(boulder2);
        window.draw(rectangle);
        window.draw(circle1);
        window.draw(circle2);
        window.display();
    }
}
```

----

## Tasks

* [x] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/include/SFML/Graphics/CircleShape.hpp
+++ b/include/SFML/Graphics/CircleShape.hpp
@@ -105,6 +105,18 @@ public:
     ////////////////////////////////////////////////////////////
     Vector2f getPoint(std::size_t index) const override;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the geometric center of the circle
+    ///
+    /// The returned point is in local coordinates, that is,
+    /// the shape's transforms (position, rotation, scale) are
+    /// not taken into account.
+    ///
+    /// \return The geometric center of the shape
+    ///
+    ////////////////////////////////////////////////////////////
+    Vector2f getGeometricCenter() const override;
+
 private:
     ////////////////////////////////////////////////////////////
     // Member data
--- a/include/SFML/Graphics/RectangleShape.hpp
+++ b/include/SFML/Graphics/RectangleShape.hpp
@@ -93,6 +93,18 @@ public:
     ////////////////////////////////////////////////////////////
     Vector2f getPoint(std::size_t index) const override;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the geometric center of the rectangle
+    ///
+    /// The returned point is in local coordinates, that is,
+    /// the shape's transforms (position, rotation, scale) are
+    /// not taken into account.
+    ///
+    /// \return The geometric center of the shape
+    ///
+    ////////////////////////////////////////////////////////////
+    Vector2f getGeometricCenter() const override;
+
 private:
     ////////////////////////////////////////////////////////////
     // Member data
--- a/src/SFML/Graphics/CircleShape.cpp
+++ b/src/SFML/Graphics/CircleShape.cpp
@@ -75,4 +75,11 @@ Vector2f CircleShape::getPoint(std::size_t index) const
     return Vector2f(m_radius, m_radius) + Vector2f(m_radius, angle);
 }
 
+
+////////////////////////////////////////////////////////////
+Vector2f CircleShape::getGeometricCenter() const
+{
+    return Vector2f(m_radius, m_radius);
+}
+
 } // namespace sf
--- a/src/SFML/Graphics/RectangleShape.cpp
+++ b/src/SFML/Graphics/RectangleShape.cpp
@@ -78,4 +78,11 @@ Vector2f RectangleShape::getPoint(std::size_t index) const
     }
 }
 
+
+////////////////////////////////////////////////////////////
+Vector2f RectangleShape::getGeometricCenter() const
+{
+    return m_size / 2.f;
+}
+
 } // namespace sf
--- a/src/SFML/Graphics/Shape.cpp
+++ b/src/SFML/Graphics/Shape.cpp
@@ -29,8 +29,7 @@
 #include <SFML/Graphics/Shape.hpp>
 #include <SFML/Graphics/Texture.hpp>
 
-#include <cmath>
-
+#include <algorithm>
 
 namespace
 {
@@ -127,6 +126,56 @@ void Shape::setOutlineThickness(float thickness)
 float Shape::getOutlineThickness() const
 {
     return m_outlineThickness;
+}
+
+
+////////////////////////////////////////////////////////////
+Vector2f Shape::getGeometricCenter() const
+{
+    const auto count = getPointCount();
+
+    switch (count)
+    {
+        case 0:
+            assert(false);
+            return Vector2f{};
+        case 1:
+            return getPoint(0);
+        case 2:
+            return (getPoint(0) + getPoint(1)) / 2.f;
+        default: // more than two points
+            Vector2f centroid;
+            float    twiceArea = 0;
+
+            auto previousPoint = getPoint(count - 1);
+            for (std::size_t i = 0; i < count; ++i)
+            {
+                const auto  currentPoint = getPoint(i);
+                const float product      = previousPoint.cross(currentPoint);
+                twiceArea += product;
+                centroid += (currentPoint + previousPoint) * product;
+
+                previousPoint = currentPoint;
+            }
+
+            if (twiceArea != 0.f)
+            {
+                return centroid / 3.f / twiceArea;
+            }
+
+            // Fallback for no area - find the center of the bounding box
+            auto minPoint = getPoint(0);
+            auto maxPoint = minPoint;
+            for (std::size_t i = 1; i < count; ++i)
+            {
+                const auto currentPoint = getPoint(i);
+                minPoint.x              = std::min(minPoint.x, currentPoint.x);
+                maxPoint.x              = std::max(maxPoint.x, currentPoint.x);
+                minPoint.y              = std::min(minPoint.y, currentPoint.y);
+                maxPoint.y              = std::max(maxPoint.y, currentPoint.y);
+            }
+            return (maxPoint + minPoint) / 2.f;
+    }
 }
 
 
--- a/test/Graphics/CircleShape.test.cpp
+++ b/test/Graphics/CircleShape.test.cpp
@@ -22,6 +22,7 @@ TEST_CASE("[Graphics] sf::CircleShape")
         CHECK(circle.getPointCount() == 30);
         for (std::size_t i = 0; i < circle.getPointCount(); ++i)
             CHECK(circle.getPoint(i) == sf::Vector2f(0, 0));
+        CHECK(circle.getGeometricCenter() == sf::Vector2f(0, 0));
     }
 
     SECTION("Radius constructor")
@@ -59,6 +60,7 @@ TEST_CASE("[Graphics] sf::CircleShape")
         CHECK(circle.getPoint(27) == Approx(sf::Vector2f(6.183218002f, 2.864748001f)));
         CHECK(circle.getPoint(28) == Approx(sf::Vector2f(8.898950577f, 1.296818733f)));
         CHECK(circle.getPoint(29) == Approx(sf::Vector2f(11.881320953f, 0.327786446f)));
+        CHECK(circle.getGeometricCenter() == sf::Vector2f(15.f, 15.f));
     }
 
     SECTION("Radius and point count constructor")
@@ -74,6 +76,7 @@ TEST_CASE("[Graphics] sf::CircleShape")
         CHECK(circle.getPoint(5) == Approx(sf::Vector2f(1.464466095f, 8.535533905f)));
         CHECK(circle.getPoint(6) == Approx(sf::Vector2f(0.000000000f, 4.999999523f)));
         CHECK(circle.getPoint(7) == Approx(sf::Vector2f(1.464465857f, 1.464466572f)));
+        CHECK(circle.getGeometricCenter() == sf::Vector2f(5.f, 5.f));
     }
 
     SECTION("Set radius")
@@ -88,6 +91,7 @@ TEST_CASE("[Graphics] sf::CircleShape")
         CHECK(circle.getPoint(3) == Approx(sf::Vector2f(10.000000000f, 20.000000000f)));
         CHECK(circle.getPoint(4) == Approx(sf::Vector2f(1.339746475f, 15.000000000f)));
         CHECK(circle.getPoint(5) == Approx(sf::Vector2f(1.339745522f, 5.000000000f)));
+        CHECK(circle.getGeometricCenter() == sf::Vector2f(10.f, 10.f));
     }
 
     SECTION("Set point count")
@@ -100,6 +104,7 @@ TEST_CASE("[Graphics] sf::CircleShape")
         CHECK(circle.getPoint(1) == Approx(sf::Vector2f(8.000000000f, 4.000000000f)));
         CHECK(circle.getPoint(2) == Approx(sf::Vector2f(3.999999762f, 8.000000000f)));
         CHECK(circle.getPoint(3) == Approx(sf::Vector2f(0.000000000f, 3.999999762f)));
+        CHECK(circle.getGeometricCenter() == sf::Vector2f(4.f, 4.f));
     }
 
     SECTION("Equilateral triangle")
@@ -110,5 +115,19 @@ TEST_CASE("[Graphics] sf::CircleShape")
         CHECK(triangle.getPoint(0) == Approx(sf::Vector2f(1.999999881f, 0.000000000f)));
         CHECK(triangle.getPoint(1) == Approx(sf::Vector2f(3.732050896f, 3.000000000f)));
         CHECK(triangle.getPoint(2) == Approx(sf::Vector2f(0.267949224f, 3.000000000f)));
+        CHECK(triangle.getGeometricCenter() == sf::Vector2f(2.f, 2.f));
+    }
+
+    SECTION("Geometric center")
+    {
+        SECTION("2 points")
+        {
+            CHECK(sf::CircleShape(2.f, 2).getGeometricCenter() == sf::Vector2f(2.f, 2.f));
+        }
+
+        SECTION("3 points")
+        {
+            CHECK(sf::CircleShape(4.f, 3).getGeometricCenter() == sf::Vector2f(4.f, 4.f));
+        }
     }
 }
--- a/test/Graphics/ConvexShape.test.cpp
+++ b/test/Graphics/ConvexShape.test.cpp
@@ -1,5 +1,8 @@
 #include <SFML/Graphics/ConvexShape.hpp>
 
+// Other 1st party headers
+#include <SFML/Graphics/CircleShape.hpp>
+
 #include <catch2/catch_test_macros.hpp>
 
 #include <SystemUtil.hpp>
@@ -44,5 +47,147 @@ TEST_CASE("[Graphics] sf::ConvexShape")
         convex.setPointCount(1);
         convex.setPoint(0, {3, 4});
         CHECK(convex.getPoint(0) == sf::Vector2f(3, 4));
+    }
+
+    SECTION(
+        "Construct clockwise ConvexShapes from CircleShapes to verify that they get approx. the same geometric center")
+    {
+        sf::ConvexShape convex;
+        for (unsigned int i = 2; i < 10; ++i)
+        {
+            const sf::CircleShape circle(4.f, i);
+            convex.setPointCount(i);
+            for (unsigned int j = 0; j < i; ++j)
+            {
+                convex.setPoint(j, circle.getPoint(j));
+            }
+            CHECK(convex.getGeometricCenter() == Approx(circle.getGeometricCenter()));
+        }
+    }
+
+    SECTION(
+        "Construct counterclockwise ConvexShapes from CircleShapes to verify that they get approx. the same geometric "
+        "center")
+    {
+        sf::ConvexShape convex;
+        for (unsigned int i = 2; i < 10; ++i)
+        {
+            const sf::CircleShape circle(4.f, i);
+            convex.setPointCount(i);
+            for (unsigned int j = 0; j < i; ++j)
+            {
+                convex.setPoint(i - 1 - j, circle.getPoint(j));
+            }
+            CHECK(convex.getGeometricCenter() == Approx(circle.getGeometricCenter()));
+        }
+    }
+
+    SECTION("Geometric center for one point")
+    {
+        sf::ConvexShape convex(1);
+        convex.setPoint(0, {1.f, 1.f});
+        CHECK(convex.getGeometricCenter() == sf::Vector2f(1.f, 1.f));
+    }
+
+    SECTION("Geometric center for two points")
+    {
+        sf::ConvexShape convex(2);
+        convex.setPoint(0, {0.f, 0.f});
+        convex.setPoint(1, {4.f, 2.f});
+        CHECK(convex.getGeometricCenter() == sf::Vector2f(2.f, 1.f));
+    }
+
+    SECTION("Geometric center for three points with a small area")
+    {
+        sf::ConvexShape convex(3);
+        convex.setPoint(0, {-100000.f, 0.f});
+        convex.setPoint(1, {100000.f, 0.f});
+        convex.setPoint(2, {100000.f, 0.000001f});
+        CHECK(convex.getGeometricCenter() == Approx(sf::Vector2f(100000.f / 3.f, 0.f)));
+    }
+
+    SECTION("Geometric center for aligned points")
+    {
+        SECTION("Geometric center for partly aligned points")
+        {
+            sf::ConvexShape convex(3);
+            convex.setPoint(0, {-100.f, 0.f});
+            convex.setPoint(1, {0.f, 0.f});
+            convex.setPoint(2, {100.f, 1.f});
+            CHECK(convex.getGeometricCenter() == Approx(sf::Vector2f(0.f, 1.f / 3.f)));
+        }
+
+        SECTION("Geometric center for aligned points with the two furthest apart not first and last")
+        {
+            sf::ConvexShape convex(4);
+            convex.setPoint(0, {-50.f, -50.f});
+            convex.setPoint(1, {-150.f, -150.f});
+            convex.setPoint(2, {150.f, 150.f});
+            convex.setPoint(3, {50.f, 50.f});
+            CHECK(convex.getGeometricCenter() == sf::Vector2f(0.f, 0.f));
+        }
+
+        SECTION("Geometric center for aligned points increasing x and y")
+        {
+            sf::ConvexShape convex(3);
+            convex.setPoint(0, {1.f, 1.f});
+            convex.setPoint(1, {5.f, 3.f});
+            convex.setPoint(2, {9.f, 5.f});
+            CHECK(convex.getGeometricCenter() == sf::Vector2f(5.f, 3.f));
+        }
+
+        SECTION("Geometric center for aligned points increasing x, decreasing y")
+        {
+            sf::ConvexShape convex(3);
+            convex.setPoint(0, {1.f, 5.f});
+            convex.setPoint(1, {5.f, 3.f});
+            convex.setPoint(2, {9.f, 1.f});
+            CHECK(convex.getGeometricCenter() == sf::Vector2f(5.f, 3.f));
+        }
+
+        SECTION("Geometric center for aligned points decreasing x and y")
+        {
+            sf::ConvexShape convex(3);
+            convex.setPoint(0, {9.f, 5.f});
+            convex.setPoint(1, {5.f, 3.f});
+            convex.setPoint(2, {1.f, 1.f});
+            CHECK(convex.getGeometricCenter() == sf::Vector2f(5.f, 3.f));
+        }
+
+        SECTION("Geometric center for aligned points decreasing x, increasing y")
+        {
+            sf::ConvexShape convex(3);
+            convex.setPoint(0, {9.f, 1.f});
+            convex.setPoint(1, {5.f, 3.f});
+            convex.setPoint(2, {1.f, 5.f});
+            CHECK(convex.getGeometricCenter() == sf::Vector2f(5.f, 3.f));
+        }
+
+        SECTION("Geometric center for aligned points with the same x value")
+        {
+            sf::ConvexShape convex(3);
+            convex.setPoint(0, {1.f, 2.f});
+            convex.setPoint(1, {1.f, 3.f});
+            convex.setPoint(2, {1.f, 1.f});
+            CHECK(convex.getGeometricCenter() == sf::Vector2f(1.f, 2.f));
+        }
+
+        SECTION("Geometric center for aligned points with the same y value")
+        {
+            sf::ConvexShape convex(3);
+            convex.setPoint(0, {2.f, 5.f});
+            convex.setPoint(1, {3.f, 5.f});
+            convex.setPoint(2, {1.f, 5.f});
+            CHECK(convex.getGeometricCenter() == sf::Vector2f(2.f, 5.f));
+        }
+
+        SECTION("Geometric center for aligned points out of order")
+        {
+            sf::ConvexShape convex(3);
+            convex.setPoint(0, {5.f, 3.f});
+            convex.setPoint(1, {1.f, 5.f});
+            convex.setPoint(2, {9.f, 1.f});
+            CHECK(convex.getGeometricCenter() == sf::Vector2f(5.f, 3.f));
+        }
     }
 }
--- a/test/Graphics/RectangleShape.test.cpp
+++ b/test/Graphics/RectangleShape.test.cpp
@@ -24,6 +24,7 @@ TEST_CASE("[Graphics] sf::RectangleShape")
         CHECK(rectangle.getPoint(1) == sf::Vector2f(0, 0));
         CHECK(rectangle.getPoint(2) == sf::Vector2f(0, 0));
         CHECK(rectangle.getPoint(3) == sf::Vector2f(0, 0));
+        CHECK(rectangle.getGeometricCenter() == sf::Vector2f(0, 0));
     }
 
     SECTION("Size constructor")
@@ -35,6 +36,7 @@ TEST_CASE("[Graphics] sf::RectangleShape")
         CHECK(rectangle.getPoint(1) == sf::Vector2f(9, 0));
         CHECK(rectangle.getPoint(2) == sf::Vector2f(9, 8));
         CHECK(rectangle.getPoint(3) == sf::Vector2f(0, 8));
+        CHECK(rectangle.getGeometricCenter() == sf::Vector2f(9.f, 8.f) / 2.f);
     }
 
     SECTION("Set size")
@@ -42,5 +44,6 @@ TEST_CASE("[Graphics] sf::RectangleShape")
         sf::RectangleShape rectangle({7, 6});
         rectangle.setSize({5, 4});
         CHECK(rectangle.getSize() == sf::Vector2f(5, 4));
+        CHECK(rectangle.getGeometricCenter() == sf::Vector2f(5.f, 4.f) / 2.f);
     }
 }
--- a/include/SFML/Graphics/Shape.hpp
+++ b/include/SFML/Graphics/Shape.hpp
@@ -209,6 +209,18 @@ public:
     virtual Vector2f getPoint(std::size_t index) const = 0;
 
     ////////////////////////////////////////////////////////////
+    /// \brief Get the geometric center of the shape
+    ///
+    /// The returned point is in local coordinates, that is,
+    /// the shape's transforms (position, rotation, scale) are
+    /// not taken into account.
+    ///
+    /// \return The geometric center of the shape
+    ///
+    ////////////////////////////////////////////////////////////
+    virtual Vector2f getGeometricCenter() const;
+
+    ////////////////////////////////////////////////////////////
     /// \brief Get the local bounding rectangle of the entity
     ///
     /// The returned rectangle is in local coordinates, which means
--- a/test/Graphics/Shape.test.cpp
+++ b/test/Graphics/Shape.test.cpp
@@ -88,13 +88,14 @@ TEST_CASE("[Graphics] sf::Shape")
         CHECK(triangleShape.getOutlineThickness() == 3.14f);
     }
 
-    SECTION("Virtual functions: getPoint, getPointCount")
+    SECTION("Virtual functions: getPoint, getPointCount, getGeometricCenter")
     {
         const TriangleShape triangleShape({2, 2});
         CHECK(triangleShape.getPointCount() == 3);
         CHECK(triangleShape.getPoint(0) == sf::Vector2f(1, 0));
         CHECK(triangleShape.getPoint(1) == sf::Vector2f(0, 2));
         CHECK(triangleShape.getPoint(2) == sf::Vector2f(2, 2));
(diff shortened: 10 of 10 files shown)


ChrisThrasher:
I think I prefer this approach because it doesn't force derived classes to implement this non-trivial algorithm if they'd rather stick to the generic implementation. Are there any other use cases for this you can envision other than rotating about the geometric center?

TedLyngmo:
> Are there any other use cases for this you can envision other than rotating about the geometric center?

If one would like to connect shapes I guess the centroid or the centroid + an offset is often a good place to put a joint.

But, it was for the case of making rotating complex shapes easy that I wanted it myself.

src/SFML/Graphics/ConvexShape.cpp
```diff
@@ -66,4 +66,5 @@ Vector2f ConvexShape::getPoint(std::size_t index) const
     return m_points[index];
 }
 
+
```
ChrisThrasher, line 69:
```suggestion
```

src/SFML/Graphics/RectangleShape.cpp
```diff
@@ -78,4 +78,10 @@ Vector2f RectangleShape::getPoint(std::size_t index) const
     }
 }
 
+////////////////////////////////////////////////////////////
```
ChrisThrasher, line 81:
```suggestion

////////////////////////////////////////////////////////////
```

src/SFML/Graphics/Shape.cpp
```diff
@@ -143,6 +144,48 @@ FloatRect Shape::getGlobalBounds() const
...
+
+    if (count == 2) // a line
+    {
+        return (getPoint(0) + getPoint(1)) / 2.f;
+    }
+
+    if (count == 1) // a single point
+    {
+        return getPoint(0);
+    }
```
ChrisThrasher, line 182: Might this be better expressed as a `switch` statement?

src/SFML/Graphics/Shape.cpp
```diff
@@ -143,6 +144,48 @@ FloatRect Shape::getGlobalBounds() const
...
 }
 
+Vector2f Shape::getGeometricCenter() const
+{
+    auto count = getPointCount();
+    if (count > 2)
+    {
+        const auto off  = getPoint(0);
+        const auto off2 = 2.f * off;
+        Vector2f   xy{};
```
ChrisThrasher, line 154:
```suggestion
        Vector2f   xy;
```

src/SFML/Graphics/Shape.cpp
```diff
@@ -143,6 +144,48 @@ FloatRect Shape::getGlobalBounds() const
...
+    }
+
+    if (count == 1) // a single point
+    {
+        return getPoint(0);
+    }
+
+    // empty
+    assert(false);
+    throw std::out_of_range("Called getGeometricCenter() on empty Shape");
```
ChrisThrasher, line 186: I think I'd prefer `std::terminate` with some error text written to `sf::err` beforehand since SFML currently does not intentionally throw anywhere in its implementation.

test/Graphics/CircleShape.test.cpp
```diff
@@ -108,4 +108,12 @@ TEST_CASE("[Graphics] sf::CircleShape")
         CHECK(triangle.getPoint(1) == Approx(sf::Vector2f(3.732050896f, 3.000000000f)));
         CHECK(triangle.getPoint(2) == Approx(sf::Vector2f(0.267949224f, 3.000000000f)));
     }
+
+    SUBCASE("Geometrical center")
+    {
+        const sf::CircleShape circle1(2.f);
+        CHECK(circle1.getGeometricCenter() == sf::Vector2f(2.f, 2.f));
+        const sf::CircleShape circle2(4.f, 3);
+        CHECK(circle2.getGeometricCenter() == sf::Vector2f(4.f, 4.f));
```
ChrisThrasher, line 117:
```suggestion
        CHECK(sf::CircleShape(2.f).getGeometricCenter() == sf::Vector2f(2.f, 2.f));
        CHECK(sf::CircleShape(4.f, 3).getGeometricCenter() == sf::Vector2f(4.f, 4.f));
```
Let's avoid intermediate variables if possible.

test/Graphics/RectangleShape.test.cpp
```diff
@@ -40,4 +40,10 @@ TEST_CASE("[Graphics] sf::RectangleShape")
         rectangle.setSize({5, 4});
         CHECK(rectangle.getSize() == sf::Vector2f(5, 4));
     }
+
+    SUBCASE("Geometrical center")
```
ChrisThrasher, line 44:
```suggestion
    SUBCASE("Geometric center")
```

test/Graphics/ConvexShape.test.cpp
```diff
@@ -1,3 +1,4 @@
+#include <SFML/Graphics/CircleShape.hpp>
```
ChrisThrasher, line 1: `ConvexShape.hpp` must be the first included header. Look [here](https://github.com/SFML/SFML/blob/97c00d42adbc2934b09d6b6120f4188a80d4803c/test/Graphics/Texture.test.cpp#L4) to see how we're handling header order in tests.

test/Graphics/ConvexShape.test.cpp
```diff
@@ -42,4 +43,31 @@ TEST_CASE("[Graphics] sf::ConvexShape")
         convex.setPoint(0, {3, 4});
         CHECK(convex.getPoint(0) == sf::Vector2f(3, 4));
     }
+
+    SUBCASE("Reasonable geometric center")
```
ChrisThrasher, line 47: This appears to be creating circles of varying size and testing that their centroid is correct. Can you reflect that in the `SUBCASE` name somehow?

test/Graphics/ConvexShape.test.cpp
```diff
@@ -42,4 +43,31 @@ TEST_CASE("[Graphics] sf::ConvexShape")
...
+            {
+                convex.setPoint(j, circle.getPoint(j));
+            }
+            CHECK(convex.getGeometricCenter() == Approx(circle.getGeometricCenter()));
+        }
+    }
+
+    SUBCASE("Edge case geometric centers")
+    {
+        sf::ConvexShape convex2(1);
```
ChrisThrasher, line 64: Instead of having `convex2` and `convex3` (which btw did you mean to start counting at 2?) can there be two SUBCASEs within this SUBCASE that describe what each one is testing? Presumably these two tests are covering different edge cases so lets spell out what those edge cases are.

test/Graphics/ConvexShape.test.cpp
```diff
@@ -42,4 +43,31 @@ TEST_CASE("[Graphics] sf::ConvexShape")
         convex.setPoint(0, {3, 4});
         CHECK(convex.getPoint(0) == sf::Vector2f(3, 4));
     }
+
+    SUBCASE("Reasonable geometric center")
+    {
+        sf::ConvexShape convex;
+        for (unsigned i = 2; i < 10; ++i)
```
ChrisThrasher, line 50:
```suggestion
        for (unsigned int i = 2; i < 10; ++i)
```
We always use `unsigned int` rather than `unsigned` for historical style reasons.

src/SFML/Graphics/Shape.cpp
```diff
@@ -143,6 +144,48 @@ FloatRect Shape::getGlobalBounds() const
     return getTransform().transformRect(getLocalBounds());
 }
 
+Vector2f Shape::getGeometricCenter() const
+{
+    auto count = getPointCount();
+    if (count > 2)
+    {
+        const auto off  = getPoint(0);
+        const auto off2 = 2.f * off;
```
ChrisThrasher, line 153: Can you use more expressive variable names?

src/SFML/Graphics/Shape.cpp
```diff
@@ -143,6 +144,48 @@ FloatRect Shape::getGlobalBounds() const
     return getTransform().transformRect(getLocalBounds());
 }
 
+Vector2f Shape::getGeometricCenter() const
+{
+    auto count = getPointCount();
```
ChrisThrasher, line 149:
```suggestion
    const auto count = getPointCount();
```

ChrisThrasher (changes requested):

src/SFML/Graphics/Shape.cpp
ChrisThrasher: It's not okay to use that because it's a C++20 feature. Let's focus on readability and expressiveness first. Performance is a tertiary concern.

TedLyngmo: Ok, so, I took a look at it using a `switch` and I didn't like it, but I'll change if you want it.

src/SFML/Graphics/Shape.cpp
```diff
@@ -143,6 +145,44 @@ FloatRect Shape::getGlobalBounds() const
...
+            float      f;
+
+            auto p2 = getPoint(count - 1);
+            for (std::size_t i = 0; i < count; ++i)
+            {
+                const auto p1 = getPoint(i);
+
+                f = (p1.x - offset.x) * (p2.y - offset.y) - (p2.x - offset.x) * (p1.y - offset.y);
+                twicearea += f;
+                centroid += ((p1 + p2) - offsetX2) * f;
```
kimci86, line 175: I don't think offset is needed. Also this can be more readable using `cross` method.

```suggestion
                const float product = p2.cross(p1);
                twicearea += product;
                centroid += (p1 + p2) * product;
```

src/SFML/Graphics/CircleShape.cpp
```diff
@@ -75,4 +75,12 @@ Vector2f CircleShape::getPoint(std::size_t index) const
     return Vector2f(m_radius, m_radius) + Vector2f(m_radius, angle);
 }
 
+////////////////////////////////////////////////////////////
+Vector2f CircleShape::getGeometricCenter() const
+{
+    if (m_pointCount == 3)
+        return (getPoint(0) + getPoint(1) + getPoint(2)) / 3.f;
```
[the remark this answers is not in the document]
TedLyngmo: @kimci86 Hmm, no, apparently not. I could have sworn that the centroid was misplaced for circles with only three points when I started. :-) Removing it and adding a test for it instead

src/SFML/Graphics/Shape.cpp
TedLyngmo: @kimci86 Without the factor `f` and `offsetX2` the centroid is placed very far away. I'm sure I misunderstood how to make the change. Can you please show the whole `default` section of the function?

test/Graphics/CircleShape.test.cpp
```diff
@@ -108,4 +108,15 @@ TEST_CASE("[Graphics] sf::CircleShape")
...
+
+    SUBCASE("Geometric center")
+    {
+        CHECK(sf::CircleShape(2.f).getGeometricCenter() == sf::Vector2f(2.f, 2.f));
+    }
+
+    SUBCASE("Geometric center for three points")
+    {
+        sf::CircleShape circle(4.f, 3);
+        CHECK(circle.getGeometricCenter() == (circle.getPoint(0) + circle.getPoint(1) +  circle.getPoint(2)) / 3.f);
```
ChrisThrasher, line 120: Instead of replicating the implementation which creates a tautological test, will you please directly construct a vector with the correct value?

src/SFML/Graphics/Shape.cpp
kimci86: I imagine this should work:
```c++
default:
{
    Vector2f centroid;
    float    twicearea{};

    auto p2 = getPoint(count - 1);
    for (std::size_t i = 0; i < count; ++i)
    {
        const auto p1 = getPoint(i);

        const float product = p2.cross(p1);
        twicearea += product;
        centroid += (p1 + p2) * product;

        p2 = p1;
    }

    return centroid / 3.f / twicearea;
}
```

TedLyngmo: @kimci86 Yes, very nice! Updated the patch now! Thanks!

eXpl0it3r:

TedLyngmo:
It says _"1 change requested"_ but I _think_ I've fixed all the comments and marked everything as resolved. Is there anything I should do?

eXpl0it3r:
It just means that the reviewer who requested changes, as to review the applied changes and either request additional changes or approve them.

src/SFML/Graphics/Shape.cpp
```diff
@@ -143,6 +145,38 @@ FloatRect Shape::getGlobalBounds() const
...
+        case 2:
+            return (getPoint(0) + getPoint(1)) / 2.f;
+        case 1:
+            return getPoint(0);
+        case 0:
+            sf::err() << "Shape::getGeometricCenter() called on empty Shape";
+            std::terminate();
+        default:
+            Vector2f centroid;
+            float    twicearea{};
```
ChrisThrasher, line 163:
```suggestion
            float    twicearea = 0;
```

src/SFML/Graphics/Shape.cpp
```diff
@@ -143,6 +145,38 @@ FloatRect Shape::getGlobalBounds() const
...
+    const auto count = getPointCount();
+
+    switch (count)
+    {
+        case 2:
+            return (getPoint(0) + getPoint(1)) / 2.f;
+        case 1:
+            return getPoint(0);
+        case 0:
+            sf::err() << "Shape::getGeometricCenter() called on empty Shape";
```
ChrisThrasher, line 159:
```suggestion
            err() << "Shape::getGeometricCenter() called on empty Shape";
```

test/Graphics/ConvexShape.test.cpp
```diff
@@ -42,4 +45,52 @@ TEST_CASE("[Graphics] sf::ConvexShape")
...
+            for (unsigned j = 0; j < i; ++j)
+            {
+                convex.setPoint(j, circle.getPoint(j));
+            }
+            CHECK(convex.getGeometricCenter() == Approx(circle.getGeometricCenter()));
+        }
+    }
+
+    SUBCASE(
+        "Construct counterclockwise ConvexShapes from CircleShapes to verify that they get approx. the same geometric "
```
ChrisThrasher, line 66: Why do this same thing clockwise and counterclockwise?

ChrisThrasher (changes requested):
More minor nitpicks

test/Graphics/ConvexShape.test.cpp
TedLyngmo: @ChrisThrasher Some of the algorithms people use need to detect if the points are stored in clockwise or counterclockwise order and do a unary `-` on the result at the end. It's just a precaution if someone wants to optimize/replace the algorithm in the future.

src/SFML/Graphics/Shape.cpp
```diff
@@ -143,6 +145,38 @@ FloatRect Shape::getGlobalBounds() const
...
 
+Vector2f Shape::getGeometricCenter() const
+{
+    const auto count = getPointCount();
+
+    switch (count)
+    {
+        case 0:
+            err() << "Shape::getGeometricCenter() called on empty Shape";
+            std::terminate();
```
ChrisThrasher, line 156: Is it better to `assert(false)` and return a default-constructed `Vector2f`? I'm still on the fence about termination as opposed to returning some value. The goal is to avoid UB and returning `(0, 0)` constitutes defined behavior in release builds. This is the only place we call `std::terminate` within SFML so it stands out in that regard.

src/SFML/Graphics/Shape.cpp
```diff
@@ -143,6 +145,38 @@ FloatRect Shape::getGlobalBounds() const
...
+        case 1:
+            return getPoint(0);
+        case 2:
+            return (getPoint(0) + getPoint(1)) / 2.f;
+        default: // more than two points
+            Vector2f centroid;
+            float    twicearea = 0;
```
ChrisThrasher, line 163:
```suggestion
            float    twiceArea = 0;
```

src/SFML/Graphics/Shape.cpp
```diff
@@ -143,6 +145,38 @@ FloatRect Shape::getGlobalBounds() const
...
+
+            auto p2 = getPoint(count - 1);
```
ChrisThrasher, line 165: Is there any way to give this a more descriptive name? I'm not sure how to interpret `p1` and `p2`. Is `p2` supposed to be the previous point as you go through all points in the shape?

test/Graphics/ConvexShape.test.cpp
```diff
@@ -42,4 +45,52 @@ TEST_CASE("[Graphics] sf::ConvexShape")
...
         CHECK(convex.getPoint(0) == sf::Vector2f(3, 4));
     }
+
+    SUBCASE(
+        "Construct clockwise ConvexShapes from CircleShapes to verify that they get approx. the same geometric center")
+    {
+        sf::ConvexShape convex;
+        for (unsigned int i = 2; i < 10; ++i)
+        {
+            sf::CircleShape circle(4.f, i);
```
ChrisThrasher, line 55:
```suggestion
            const sf::CircleShape circle(4.f, i);
```

test/Graphics/ConvexShape.test.cpp
```diff
@@ -42,4 +45,52 @@ TEST_CASE("[Graphics] sf::ConvexShape")
...
+    }
+
+    SUBCASE(
+        "Construct counterclockwise ConvexShapes from CircleShapes to verify that they get approx. the same geometric "
+        "center")
+    {
+        sf::ConvexShape convex;
+        for (unsigned int i = 2; i < 10; ++i)
+        {
+            sf::CircleShape circle(4.f, i);
```
ChrisThrasher, line 72:
```suggestion
            const sf::CircleShape circle(4.f, i);
```

test/Graphics/ConvexShape.test.cpp
```diff
@@ -42,4 +45,52 @@ TEST_CASE("[Graphics] sf::ConvexShape")
...
+
+    SUBCASE(
+        "Construct clockwise ConvexShapes from CircleShapes to verify that they get approx. the same geometric center")
+    {
+        sf::ConvexShape convex;
+        for (unsigned int i = 2; i < 10; ++i)
+        {
+            sf::CircleShape circle(4.f, i);
+            convex.setPointCount(i);
+            for (unsigned j = 0; j < i; ++j)
```
ChrisThrasher, line 57:
```suggestion
            for (unsigned int j = 0; j < i; ++j)
```

test/Graphics/ConvexShape.test.cpp
```diff
@@ -42,4 +45,52 @@ TEST_CASE("[Graphics] sf::ConvexShape")
...
+    SUBCASE(
+        "Construct counterclockwise ConvexShapes from CircleShapes to verify that they get approx. the same geometric "
+        "center")
+    {
+        sf::ConvexShape convex;
+        for (unsigned int i = 2; i < 10; ++i)
+        {
+            sf::CircleShape circle(4.f, i);
+            convex.setPointCount(i);
+            for (unsigned j = 0; j < i; ++j)
```
ChrisThrasher, line 74:
```suggestion
            for (unsigned int j = 0; j < i; ++j)
```

test/Graphics/RectangleShape.test.cpp
```diff
@@ -40,4 +40,10 @@ TEST_CASE("[Graphics] sf::RectangleShape")
         rectangle.setSize({5, 4});
         CHECK(rectangle.getSize() == sf::Vector2f(5, 4));
     }
+
+    SUBCASE("Geometric center")
+    {
+        sf::RectangleShape rectangle({20, 10});
```
ChrisThrasher, line 46:
```suggestion
        const sf::RectangleShape rectangle({20, 10});
```
This ensures that `getGeometricCenter` is marked as `const`.

ChrisThrasher (changes requested):
I like how this PR has turned out. I was worried at first that it might make certain future changes more difficult but that does not appear to be the case. It nicely fits in with the other features of `sf::Shape` and puts no addition onerous requirements on derivations of `sf::Shape`. I'm leaning towards a "yes" but want to hear more from the other maintainers before I can approve this.

src/SFML/Graphics/Shape.cpp
TedLyngmo: I'm not sure what's best. I had a default constructed `Vector2f` as the return value before but was adviced to change it to `std::terminate` in a CR. `{0, 0}` wouldn't be _too_ crazy if there are no points, at least not in this context I think.

ChrisThrasher: Haha yeah sorry if we've taken you in circles. Don't go changing it yet until we can have a final discussion about it.

TedLyngmo: No worries! :-) I'll leave it as-is for now.

src/SFML/Graphics/Shape.cpp
TedLyngmo: Yepp, that's it. Renamed them now.

kimci86 (commented):
There is something odd with the current implementation:
- The center of a shape with two points is the middle of the segment.
- The formula for more points which are aligned on a segment is meaningless. (it give NaN coordinates)

Maybe we need an alternative algorithm for aligned points, or decide that shapes with no surface do not have a center of mass.

ChrisThrasher:
Are the `NaN`s because `twiceArea` may be zero? Let's `assert` against that postcondition at a minimum.

TedLyngmo:
> Are the `NaN`s because `twiceArea` may be zero? Let's `assert` against that postcondition at a minimum.

Yes that's why andI think it's possible to give a proper "middle of the line" result without too much fuzz. I will give it a think. Darn corner cases :-)

src/SFML/Graphics/Shape.cpp
Bromeon: I would also not use `std::terminate`, as you say it'd be the only place where this happens.

`assert` + returning a zero-vector is probably sane. https://github.com/SFML/SFML/issues/2139 is unfortunately not concluded yet...

ChrisThrasher:
Divide by zero is of course not acceptable, but what happens as `twiceArea` approaches zero? Do we still get a sensible stable result even if `twiceArea` is something tiny like `1e-10`?

TedLyngmo:
> Maybe we need an alternative algorithm for aligned points, or decide that shapes with no surface do not have a center of mass.

Yes, the correct centroid would be in the middle of the two points furthest apart. I made an update to the patch that works as long as the first and last point are those furthest apart. That's not bullet proof to say the least. I'll give it another thought later.

src/SFML/Graphics/Shape.cpp
TedLyngmo: Ok, I've changed back to an `assert` and returning `Vector2f{}` in my latest patch :)

test/Graphics/ConvexShape.test.cpp
```diff
@@ -42,4 +45,88 @@ TEST_CASE("[Graphics] sf::ConvexShape")
...
+    {
+        SUBCASE("Geometric center for aligned points increasing")
+        {
+            sf::ConvexShape convex(3);
+            convex.setPoint(0, {1.f, 1.f});
+            convex.setPoint(1, {5.f, 3.f});
+            convex.setPoint(2, {9.f, 5.f});
+            CHECK(convex.getGeometricCenter() == sf::Vector2f(5.f, 3.f));
+        }
+        SUBCASE("Geometric center for aligned points decreasing")
```
ChrisThrasher, line 107:
```suggestion

        SUBCASE("Geometric center for aligned points decreasing")
```

test/Graphics/ConvexShape.test.cpp
```diff
@@ -42,4 +45,88 @@ TEST_CASE("[Graphics] sf::ConvexShape")
...
+        {
+            sf::ConvexShape convex(3);
+            convex.setPoint(0, {9.f, 5.f});
+            convex.setPoint(1, {5.f, 3.f});
+            convex.setPoint(2, {1.f, 1.f});
+            CHECK(convex.getGeometricCenter() == sf::Vector2f(5.f, 3.f));
+        }
+        SUBCASE("Geometric center for aligned points with the same x value")
```
ChrisThrasher, line 115:
```suggestion

        SUBCASE("Geometric center for aligned points with the same x value")
```

test/Graphics/ConvexShape.test.cpp
```diff
@@ -42,4 +45,88 @@ TEST_CASE("[Graphics] sf::ConvexShape")
...
+        {
+            sf::ConvexShape convex(3);
+            convex.setPoint(0, {1.f, 2.f});
+            convex.setPoint(1, {1.f, 3.f});
+            convex.setPoint(2, {1.f, 1.f});
+            CHECK(convex.getGeometricCenter() == sf::Vector2f(1.f, 2.f));
+        }
+        SUBCASE("Geometric center for aligned points with the same y value")
```
ChrisThrasher, line 123:
```suggestion

        SUBCASE("Geometric center for aligned points with the same y value")
```

src/SFML/Graphics/Shape.cpp
```diff
@@ -143,6 +142,49 @@ FloatRect Shape::getGlobalBounds() const
...
+        default: // more than two points
+            Vector2f centroid;
+            float    twiceArea = 0;
+
+            auto previous_point = getPoint(count - 1);
+            auto minPoint       = previous_point;
+            auto maxPoint       = previous_point;
+            for (std::size_t i = 0; i < count; ++i)
+            {
+                const auto current_point = getPoint(i);
```
ChrisThrasher, line 167:
```suggestion
                const auto currentPoint = getPoint(i);
```
`clang-tidy` is failing because `current_point` is the wrong variable name style. Same goes for `previous_point`.

ChrisThrasher (changes requested):

TedLyngmo:
Hmm, there seems to be a problem with the codecov uploader and some formulae on the Mac side. Is there anything I should do?

ChrisThrasher:
> Hmm, there seems to be a problem with the codecov uploader and some formulae on the Mac side. Is there anything I should do?

You can ignore that. It's a problem with Codecov itself that we can't entirely fix.

ChrisThrasher:

test/Graphics/ConvexShape.test.cpp
```diff
@@ -42,4 +45,146 @@ TEST_CASE("[Graphics] sf::ConvexShape")
...
+    SUBCASE(
+        "Construct counterclockwise ConvexShapes from CircleShapes to verify that they get approx. the same geometric "
+        "center")
+    {
+        sf::ConvexShape convex;
+        for (unsigned int i = 2; i < 10; ++i)
+        {
+            const sf::CircleShape circle(4.f, i);
+            convex.setPointCount(i);
+            for (unsigned j = 0; j < i; ++j)
```
ChrisThrasher, line 74:
```suggestion
            for (unsigned int j = 0; j < i; ++j)
```

src/SFML/Graphics/Shape.cpp
```diff
@@ -143,6 +142,57 @@ FloatRect Shape::getGlobalBounds() const
...
+            {
+                const auto currentPoint = getPoint(i);
+                if (currentPoint.x < minPoint.x)
+                    minPoint.x = currentPoint.x;
+                else if (currentPoint.x > maxPoint.x)
+                    maxPoint.x = currentPoint.x;
+                if (currentPoint.y < minPoint.y)
+                    minPoint.y = currentPoint.y;
+                else if (currentPoint.y > maxPoint.y)
+                    maxPoint.y = currentPoint.y;
```
ChrisThrasher, line 191: Can this be simplified with `std::min` and `std::max`?

ChrisThrasher (changes requested):

src/SFML/Graphics/Shape.cpp
Bromeon: Or `std::clamp`...

ChrisThrasher: Not sure how clamp could come in handy here. The point between the min and max is not what we’re trying to change. We’re trying to change the min and max themselves.

TedLyngmo: > Can this be simplified with `std::min` and `std::max`?

Do you mean `minPoint.x = std::min(minPoint.x, currPoint.x);` etc.? Sure. I don't like it, but if you want it that way, I'll change it.

> Or `std::clamp` ...

I don't see how I could clamp here.

TedLyngmo: > > Do you mean `minPoint.x = std::min(minPoint.x, currPoint.x);` etc.?
> 
> Yes. It's more expressive than rewriting that same logic by hand.

It's not exactly the same logic though. In the current version, if the first condition is `true` it will skip the second. Minor detail perhaps, but that's why I don't like using `min`/`max` in these situations.

Would it ok to make it expressive by hiding it like this:

    template<class T>
    std::pair<T, T> setMinMax(const T valueToTest, T& currentMin, T& currentMax) {
        if (valueToTest > currentMax) currentMax = valueToTest;
        else if (valueToTest < currentMin) currentMin = valueToTest;
    }

The code in the function with then become

            for (std::size_t i = 1; i < count; ++i)
            {
                const auto currentPoint = getPoint(i);
                setMinMax(currentPoint.x, minPoint.x, maxPoint.x);
                setMinMax(currentPoint.y, minPoint.y, maxPoint.y);
            }

If not, I'll go with `min`/`max`.

ChrisThrasher: > It's not exactly the same logic though.

The result is the same, right?

> Would it ok to make it expressive by hiding it like this:

This abstraction is not as intuitive to me and feels excessively complicated when all we need are a few calls to `min` and `max` functions.

TedLyngmo: Ok, `min`/`max` it is then :-)

ChrisThrasher:

TedLyngmo:
> SFMLBuildMaster: Build this please.

@ChrisThrasher An unrelated thing: I'm following https://www.sfml-dev.org/tutorials/2.5/compile-with-cmake.php to build on `master` to get a native installation of the current development on native Windows too (WSL works fine) but even though it produces the expected `.lib` files, 

```
./lib/sfml-audio-s.lib
./lib/sfml-graphics-s.lib
./lib/sfml-main.lib
./lib/sfml-network-s.lib
./lib/sfml-system-s.lib
./lib/sfml-window-s.lib
...
```
and VS2022 can find them, I get unresolved symbols for all the SFML functions. Any idea what could be wrong? I tried with `.dll`s too but got the same issue. I'm not very used to VS2022 so there could be some magic option that I don't know about :)

ChrisThrasher:
@eXpl0it3r @Bromeon @vittorioromeo @binary1248 Any thoughts on this feature and its implementation?

TedLyngmo:
> > Any idea what could be wrong?
> 
> Not off the top of my head I'm afraid. I don't use Windows myself so I have limited experience debugging Visual Studio build errors. In theory CMake should just handle all of this for you. You're certainly not the first Visual Studio user so this should be solvable. Maybe hop in the [Discord](https://discord.gg/nr4X7Fh) and we can figure it out.

Hmm, I'm trying to get in with my existing Discord account but so far no luck. It doesn't accept my password for some reason and it does not send me a reset link when I request it. I can't register a new account because ... well, it says that my email is already registered :-)

ChrisThrasher:

src/SFML/Graphics/Shape.cpp
```diff
@@ -143,6 +144,53 @@ FloatRect Shape::getGlobalBounds() const
...
+            {
+                const auto currentPoint = getPoint(i);
+                minPoint.x              = std::min(minPoint.x, currentPoint.x);
+                maxPoint.x              = std::max(maxPoint.x, currentPoint.x);
+                minPoint.y              = std::min(minPoint.y, currentPoint.y);
+                maxPoint.y              = std::max(maxPoint.y, currentPoint.y);
+            }
+            return (maxPoint + minPoint) / 2.f;
+    }
+}
```
kimci86, line 193:
```suggestion
}

```

src/SFML/Graphics/Shape.cpp
```diff
@@ -143,6 +144,53 @@ FloatRect Shape::getGlobalBounds() const
     return getTransform().transformRect(getLocalBounds());
 }
 
+Vector2f Shape::getGeometricCenter() const
```
kimci86, line 147: For consistency,
```suggestion

////////////////////////////////////////////////////////////
Vector2f Shape::getGeometricCenter() const
```

kimci86, line 147: This implementation could be moved before `getLocalBounds` to better match the order of methods in the class definition.

src/SFML/Graphics/Shape.cpp
```diff
@@ -143,6 +144,53 @@ FloatRect Shape::getGlobalBounds() const
...
+
+                previousPoint = currentPoint;
+            }
+
+            if (twiceArea != 0.f)
+            {
+                return centroid / 3.f / twiceArea;
+            }
+
+            // Fallback for no area - find the midpoint between the two points furthest apart
```
kimci86, line 180:
```suggestion
            // Fallback for no area - find the center of the bounding box
```

kimci86 (commented):
Nitpicking

include/SFML/Graphics/CircleShape.hpp
```diff
@@ -105,6 +105,18 @@ class SFML_GRAPHICS_API CircleShape : public Shape
...
     Vector2f getPoint(std::size_t index) const override;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the geometric center of the circle
+    ///
+    /// The returned point is in local coordinates, that is,
+    /// the shape's transforms (position, rotation, scale) are
+    /// not taken into account.
+    ///
+    /// \return The geometric center of the shape
```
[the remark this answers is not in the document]
TedLyngmo: Should that perhaps be in a separate commit where all the overridden functions descriptions are changes to mention the specific type instead?

Example:
```cpp
    /// \return index-th point of the shape
    ///
    Vector2f getPoint(std::size_t index) const override;
```

include/SFML/Graphics/RectangleShape.hpp
```diff
@@ -93,6 +93,18 @@ class SFML_GRAPHICS_API RectangleShape : public Shape
...
     Vector2f getPoint(std::size_t index) const override;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the geometric center of the rectangle
+    ///
+    /// The returned point is in local coordinates, that is,
+    /// the shape's transforms (position, rotation, scale) are
+    /// not taken into account.
+    ///
+    /// \return The geometric center of the shape
```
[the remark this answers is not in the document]
TedLyngmo: Same as previous query

include/SFML/Graphics/CircleShape.hpp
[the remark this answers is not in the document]
kimci86: That's a good point! Feel free to ignore my comment for now then.

eXpl0it3r:

include/SFML/Graphics/CircleShape.hpp
```diff
@@ -105,6 +105,18 @@ class SFML_GRAPHICS_API CircleShape : public Shape
...
+    ///
+    ////////////////////////////////////////////////////////////
```
eXpl0it3r, line 117: If we don't change the doxygen comment, then we might as well just copy the docs from the parent class with:

```
    /// \copydoc sf::Shape::getGeometricCenter()
    ///
```

include/SFML/Graphics/RectangleShape.hpp
```diff
@@ -93,6 +93,18 @@ class SFML_GRAPHICS_API RectangleShape : public Shape
...
+    ///
+    ////////////////////////////////////////////////////////////
```
eXpl0it3r, line 105: If we don't change the doxygen comment, then we might as well just copy the docs from the parent class with:

```
    /// \copydoc sf::Shape::getGeometricCenter()
    ///
```

eXpl0it3r (changes requested):
Just a minor thing regarding documentation text, then we're ready to merge, sorry for the delay 🙂

src/SFML/Graphics/Shape.cpp
```diff
@@ -137,6 +136,56 @@ FloatRect Shape::getLocalBounds() const
 }
 
 
+////////////////////////////////////////////////////////////
+Vector2f Shape::getGeometricCenter() const
```
kimci86, line 140: This implementation is still not just before `getLocalBounds` implementation, contrary to the order in the class definition.

include/SFML/Graphics/RectangleShape.hpp
TedLyngmo: I've tried to follow the existing praxis so if I do this, the comment for the added function will look a bit "rouge".  Would it be ok to go through the doc comments and do this work in a separate commit?

src/SFML/Graphics/Shape.cpp
TedLyngmo: Odd... I could have sworn I moved it... but, no. :-) Fixed now.

include/SFML/Graphics/RectangleShape.hpp
eXpl0it3r: If you think the current way is more consistent then let's stick with that

ChrisThrasher:
Fixed the compilation errors building the tests

TedLyngmo:
> Fixed the compilation errors building the tests

Oups, sloppy of me. Thanks!

ChrisThrasher:

eXpl0it3r (approved):

ChrisThrasher (approved):

eXpl0it3r:
Thank you for suggesting this feature, implementing it and keeping on during the whole review process 🎉

PR #2538, opened 2023-04-28
Fix macOS Window Resizing Bug

ChrisThrasher:
## Description

Fixes bug where resize events only got triggered upon manual window resizing OR if one called `sf::WindowBase::setSize` with a height that was too large. This was caught during the testing of #2519. The reason we think it's a bug is that other windowing backends like X11 do emit a resize event when `setSize` is called.

Brief testing proves that this works but I'm not sure if this has any unintended side effects so I'm not eager to merge until I understand the code better. This bug is present on 2.6.x as well in case we choose to backport this.

### DIFF
--- a/src/SFML/Window/OSX/SFWindowController.mm
+++ b/src/SFML/Window/OSX/SFWindowController.mm
@@ -457,18 +457,9 @@
         // Add titlebar height.
         height += static_cast<unsigned int>([self titlebarHeight]);
 
-        // Corner case: don't set the window height bigger than the screen height
-        // or the view will be resized _later_ without generating a resize event.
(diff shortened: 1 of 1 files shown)


ChrisThrasher:

eXpl0it3r:
I cherry-picked this on top of the changes in #2519, since it has such a nice test application, and it worked correctly.

eXpl0it3r (approved):

ChrisThrasher:
> And it might indeed be a candidate for SFML 2.6.1

How about we use SFML 3 to test this out for a while? If it work well in `master` for a while then we can easily cherry-pick it onto 2.6.x.

PR #2540, opened 2023-04-30
Modernize memory management of X11 types

ChrisThrasher:
## Description

This PR does two main things:

1. Remove unnecessary heap allocations
2. Manage necessary heap allocations with smart pointers

The goal at the start was to get rid of as many uses of `XFree` as possible. I quickly realized that often times we're freeing something that was heap allocated for no reason other than we chose to use an X11 factory function that happens to heap allocate the type. `XAllocWMHints` is a good example of this. It just `calloc`s a `XWMHints` the gives you a pointer to that allocation which it expects you to free. There are a few other `XAlloc` functions which all do the same thing. It initializes all members to zero which may be convenient to C programmers but we can use `T t{}` to zero-initialize all members of `T` so that's not useful to use in C++ land.

https://github.com/mirror/libX11/blob/ff8706a5eae25b8bafce300527079f68a201d27f/src/PropAlloc.c#L52-L55

Along the way I realized that a lot of other `XFree`s are cleaning up owning raw pointers that we're forced to deal with. Those are fixed by using `std::unique_ptr` to automatically clean up the allocation. In a few places X11 is actually returning a pointer to an array in which case I used `std::unique_ptr<T[]>`.

(This will get squashed. I'm keeping the verbose history in case I need to remove or revert anything.)

### DIFF
--- a/src/SFML/Window/Unix/Utils.hpp
+++ b/src/SFML/Window/Unix/Utils.hpp
@@ -1,0 +1,61 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2023 Laurent Gomila (laurent@sfml-dev.org)
+//
+// This software is provided 'as-is', without any express or implied warranty.
+// In no event will the authors be held liable for any damages arising from the use of this software.
+//
+// Permission is granted to anyone to use this software for any purpose,
+// including commercial applications, and to alter it and redistribute it freely,
+// subject to the following restrictions:
+//
+// 1. The origin of this software must not be misrepresented;
+//    you must not claim that you wrote the original software.
+//    If you use this software in a product, an acknowledgment
+//    in the product documentation would be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such,
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+#pragma once
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#include <X11/Xlib.h>
+
+#include <memory>
+#include <type_traits>
+
+
+namespace sf::priv
+{
+////////////////////////////////////////////////////////////
+/// \brief Class template for freeing X11 pointers
+///
+/// Specialized elsewhere for types that are freed through
+/// other means than XFree(). XFree() is the most common use
+/// case though so it is the default.
+///
+////////////////////////////////////////////////////////////
+template <typename T>
+struct XDeleter
+{
+    void operator()(T* data) const
+    {
+        XFree(data);
+    }
+};
+
+////////////////////////////////////////////////////////////
+/// \brief Class template for wrapping owning raw pointers from X11
+///
+////////////////////////////////////////////////////////////
+template <typename T>
+using X11Ptr = std::unique_ptr<T, XDeleter<std::remove_all_extents_t<T>>>;
+} // namespace sf::priv
--- a/src/SFML/Window/Unix/WindowImplX11.cpp
+++ b/src/SFML/Window/Unix/WindowImplX11.cpp
@@ -30,6 +30,7 @@
 #include <SFML/Window/Unix/Display.hpp>
 #include <SFML/Window/Unix/InputImpl.hpp>
 #include <SFML/Window/Unix/KeyboardImpl.hpp>
+#include <SFML/Window/Unix/Utils.hpp>
 #include <SFML/Window/Unix/WindowImplX11.hpp>
 
 #include <SFML/System/Err.hpp>
@@ -375,6 +376,50 @@ bool isWMAbsolutePositionGood()
 namespace sf::priv
 {
 ////////////////////////////////////////////////////////////
+template <>
+struct XDeleter<XImage>
+{
+    void operator()(XImage* image) const
+    {
+        XDestroyImage(image);
+    }
+};
+
+
+////////////////////////////////////////////////////////////
+template <>
+struct XDeleter<XRRScreenResources>
+{
+    void operator()(XRRScreenResources* res) const
+    {
+        XRRFreeScreenResources(res);
+    }
+};
+
+
+////////////////////////////////////////////////////////////
+template <>
+struct XDeleter<XRROutputInfo>
+{
+    void operator()(XRROutputInfo* outputInfo) const
+    {
+        XRRFreeOutputInfo(outputInfo);
+    }
+};
+
+
+////////////////////////////////////////////////////////////
+template <>
+struct XDeleter<XRRCrtcInfo>
+{
+    void operator()(XRRCrtcInfo* crtcInfo) const
+    {
+        XRRFreeCrtcInfo(crtcInfo);
+    }
+};
+
+
+////////////////////////////////////////////////////////////
 WindowImplX11::WindowImplX11(WindowHandle handle) : m_isExternal(true)
 {
     using namespace WindowsImplX11Impl;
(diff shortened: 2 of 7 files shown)


ChrisThrasher:

ChrisThrasher:

kimci86:
About your first point,
> 1. Remove unnecessary heap allocations

Those allocation might be unnecessary but it is recommended at least here https://tronche.com/gui/x/xlib/ICC/client-to-window-manager/wm-normal-hints.html for this reason:
> The size of the XSizeHints structure may grow in future releases, as new components are added to support new ICCCM features. Passing statically allocated instances of this structure into Xlib may result in memory corruption when running against a future release of the library. As such, it is recommended that only dynamically allocated instances of the structure be used.

About your second point,
> 2. Manage necessary heap allocations with smart pointers

Good idea, but make sure not to create `unique_ptr`s with a default deleter. That would mix C's malloc with C++'s delete, which is undefined behavior as far as I know.

TedLyngmo:
> > 2. Manage necessary heap allocations with smart pointers
> 
> Good idea, but make sure not to create `unique_ptr`s with a default deleter. That would mix C's malloc with C++'s delete, which is undefined behavior as far as I know.

Perhaps a helper would simplify this:
```cpp
template<class T>
struct Xfreer {
    void operator()(T* data) const noexcept { Xfree(data); }
};

template<class T, class Deleter = Xfreer<T>>
using x11_ptr = std::unique_ptr<T, Deleter>;
```

ChrisThrasher:
> it is recommended at least here https://tronche.com/gui/x/xlib/ICC/client-to-window-manager/wm-normal-hints.html for this reason:

I think what they're getting at is that using pointers makes your own code more resilient to ABI breaks. I understand why they would say that, but SFML currently does nothing to guard itself against ABI breaks in external libraries so I don't see a compelling reason to do that when using X11.

> Good idea, but make sure not to create unique_ptrs with a default deleter. That would mix C's malloc with C++'s delete, which is undefined behavior as far as I know.

This is frustrating but you're right. I'll probably try TedLyngmo's suggestion for a simple wrapper type.

ChrisThrasher:

ChrisThrasher:
I added a `std::unique_ptr` specialization that frees X11 pointers the proper way.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -899,8 +894,7 @@ void WindowImplX11::setIcon(const Vector2u& size, const std::uint8_t* pixels)
 {
     // X11 wants BGRA pixels: swap red and blue channels
     // Note: this memory will be freed by XDestroyImage
-    auto* iconPixels = static_cast<std::uint8_t*>(
-        std::malloc(static_cast<std::size_t>(size.x) * static_cast<std::size_t>(size.y) * 4));
+    auto iconPixels = std::vector<std::uint8_t>(size.x * size.y * 4);
```
kimci86, line 897: Notice the comment above. This is supposed to be allocated with malloc and freed in XDestroyImage.
Maybe our XPointerDeleter operator() could have an overload for XImage* that calls XDestroyImage?

ChrisThrasher: I was annoyed that `XDestroyImage` forced us to use `malloc` to create that buffer so by not using `XDestroyImage` we also get to use C++ idioms for allocating and freeing that buffer while `XFree` continues to clean up the image.

Here is the [`XDestroyImage`](https://github.com/mirror/libX11/blob/ff8706a5eae25b8bafce300527079f68a201d27f/src/Macros.c#L245-L249) implementation for reference. It doesn't appear to be doing anything magical but maybe I'm missing something.

kimci86: Having to look at a dependency's the source code (and actually not looking at [it](https://github.com/mirror/libX11/blob/ff8706a5eae25b8bafce300527079f68a201d27f/src/ImUtil.c#L435-L441)) instead of using the documented public interface seems wrong. It goes past libX's abstraction barrier, increasing coupling with a specific implementation. Using malloc may feel annoying, but I think it is the correct thing to do here.

ChrisThrasher:
Dealing with X11's memory management scheme is proving more complicated than expected so I'm marking this as a draft so we can sit on the code for while until we're more confident in how to approach this.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:
Rebase + a little bit of cleanup. I removed some changes I made to an `XImage` out of an abundance of caution. The remaining changes seem reasonable though so I'm opening this PR back up to reviews. I've tested this with a project of mine running in Linux and so far can't observe any issues but I'm sure I'm not hitting all code paths I've changed in this PR.

ChrisThrasher:

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -955,7 +950,6 @@ void WindowImplX11::setIcon(const Vector2u& size, const std::uint8_t* pixels)
     GC        iconGC = XCreateGC(m_display, m_iconPixmap, 0, &values);
     XPutImage(m_display, m_iconPixmap, iconGC, iconImage, 0, 0, 0, 0, size.x, size.y);
     XFreeGC(m_display, iconGC);
-    XDestroyImage(iconImage);
```
kimci86, line 958: This should not be removed.

src/SFML/Window/Unix/Utils.hpp
```diff
@@ -0,0 +1,57 @@
...
+#include <type_traits>
+
+
+namespace sf::priv
+{
+////////////////////////////////////////////////////////////
+/// \brief Class template for freeing X11 pointers
+///
+////////////////////////////////////////////////////////////
+template <class T>
```
kimci86, line 42: Note that `XFree` takes a `void*` so a template deleter is not required.

However, maybe we can make use of template specialization or function overloading to support types that are not freed with `XFree` such as `XImage` that is freed with `XDestroyImage`?

ChrisThrasher: Oh yeah we can overload that on different X pointer types. That's a great point. Let me play around with that and see what I can do.

src/SFML/Window/Unix/WindowImplX11.cpp
ChrisThrasher: Thanks. I forgot them when reverting the rest of the XImage stuff.

ChrisThrasher:

src/SFML/Window/Unix/Utils.hpp
ChrisThrasher: That wasn't too bad. I specialized the deleter for `XImage`.

ChrisThrasher:
I found many more types doing manual memory management that could be better replaced with custom `std::unique_ptr` deleters. In some places this replaced some sketchy looking C-style cleanup code where a given pointer was freed on 3 separate code paths. Those are the scenarios we're most likely to forget to free something or free it twice. I used this regex to find all the places we're declaring X11 pointers and I think I've fixed them all now. 

```
X([a-zA-Z]*)\* 
```

There are still places we're calling `XFree` but I'm not sure how we can replace those with `std::unique_ptr`. We need C++23's `std::out_ptr`. Function like `XGetWindowProperty` requires taking the address of a pointer which is something `std::unique_ptr` will not let you do since that would skirt `std::unique_ptr`'s ownership promises.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:
I wanted to bump this thread again because I'm working through [some windowing API enhancements](https://github.com/SFML/SFML/wiki/SD%3A-States-and-Styles) which will inevitably touch lots of X11 code. It would great to make a decision on this before I go modifying that X11 code since it may cause conflicts with this PR.

ChrisThrasher:

ChrisThrasher:

eXpl0it3r (approved):

PR #2541, opened 2023-04-30
Move semantics for AlResource

shakedechn:
* [V] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before?
* [V] Does the code follow the SFML [Code Style Guide](https://www.sfml-dev.org/style.php)?
* [V] Have you provided some example/test code for your changes? - A static assert, no other code is required.

----

## Description

## Tasks

* [ ] Tested on Linux
* [V] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

I am unsure why the destructor was protected and not public. The original pull request does not elaborate on the matter, which suggests that either I am missing a common use case I've never run into or this detail just isn't that important.
Defining the destructor as protected would not allow the class to pass the `std::is_nothrow_move_constructible_v` check, which is desirable. Please correct me if I am mistaken.

Edit: Sorry for the force push without changes mess, I was trying to get my commits verified.


shakedechn:
It seems appropriate to perform similar changes to classes that inherit from `AlResource` - `SoundBuffer`, `SoundRecorder`, and `SoundSource`. Would that be appropriate as part of one pull request, or should I wait for this one to be merged and then address the inheriting classes?

ChrisThrasher:
> It seems appropriate to perform similar changes to classes that inherit from `AlResource` - `SoundBuffer`, `SoundRecorder`, and `SoundSource`. Would that be appropriate as part of one pull request, or should I wait for this one to be merged and then address the inheriting classes?

Don't do that until we can make a decision on this PR

shakedechn:
The errors appear to be related to gpg key verification. I will switch to another solution to be verified shortly.

eXpl0it3r:
> The errors appear to be related to gpg key verification. I will switch to another solution to be verified shortly.

If you mean the codecov upload error, just ignore them. GitHub was having issues earlier, so a re-run will fix them.

But the formatting and analyze builds still show errors that would need to be resolved.

shakedechn:
> > The errors appear to be related to gpg key verification. I will switch to another solution to be verified shortly.
> 
> If you mean the codecov upload error, just ignore them. GitHub was having issues earlier, so a re-run will fix them.
> 
> But the formatting and analyze builds still show errors that would need to be resolved.

Oh, thank you! I already finished switching my signing key, but no matter.
Regarding pending formatting and compilation errors, I believe I have addressed them all by now.

ChrisThrasher:
I'm not yet sure why clang-tidy and GCC disagree on what to do here since this is code I don't often write and thus have never seen this particular quandary before. My first suggestion would be to use `// NOLINT(check-name-goes-here)` on the offending line or `// NOLINTNEXTLINE(check-name-goes-here)` on the preceding line to tell clang-tidy to ignore the fact that we're writing unnecessary code since that's cleaner than silencing a compiler warning.

shakedechn:
> 

Oh, it is clang-tidy! sorry, I was misled by the name of the action "Analyzing on Windows/Linux/etc" and thought this was some static code analysis error. I had not read around the error enough to see it is that. I definitely agree that silencing clang-tidy makes more sense.
Thanks for your input, I will update the PR shortly.

ChrisThrasher:
> I was misled by the name of the action "Analyzing on Windows/Linux/etc" and thought this was some static code analysis error.

It _is_ a static code analysis error. clang-tidy is a static analyzer. It's different from Clang the compiler.

shakedechn:
> > I was misled by the name of the action "Analyzing on Windows/Linux/etc" and thought this was some static code analysis error.
> 
> It _is_ a static code analysis error. clang-tidy is a static analyzer. It's different from Clang the compiler.

Huh, I thought it was just a linter. thanks for the clarification.

ChrisThrasher:
Closing because #2749 gets rid of `sf::AlResource`.

eXpl0it3r:
Technically, it replaces it with `sf::AudioResource`, but yes, the PR would need to be, majorly adjusted, so a new PR is more justified.

PR #2543, opened 2023-05-08
Upgrade to CMake 3.22

### LINKED ISSUE #2511: Raise minimum CMake version to 3.22
## Subject of the issue

The oldest OS that someone may reasonably use with SFML 3 is Ubuntu 20 LTS which ships with CMake 3.16, hence the fact that SFML 3 currently requires CMake 3.16. However if you ignore Ubuntu 20 LTS, the next oldest CMake version we expect to find in the wild is CMake 3.22 which ships with Ubuntu 22 LTS. Anyone on Windows or macOS is likely to be using something even newer than 3.22.

**I propose we raise the minimum CMake require to 3.22.** In this issue I want to keep tabs on all the features I'd like to use but can't until we raise this requirement. I plan on growing this list over time to help reinforce my argument. Bear in mind that users on Ubuntu 20 can still use SFML 3, they simply need to install a newer version of CMake which is [quite easy](https://alexreinking.com/blog/how-to-use-cmake-without-the-agonizing-pain-part-1.html).

 * [Numerous improved defaults](https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html) - See all the new policies that have been introduced between 3.16 and 3.22. Each new policy represents an improved default value which we opt into by increasing the minimum required vision.
 * [Presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) - Arguably we could start using this today since it doesn't necessarily impact the minimum CMake version to build SFML but nonetheless it's a CMake 3.19 features that has had significant updates in every minor release since then. Presets as of CMake 3.22 are fleshed out and very useful.
 * [`PROJECT_IS_TOP_LEVEL`](https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html) - There are two places where we are using a less expressive alternative to approximate what this built-in variable does.
 * [`TARGET_RUNTIME_DLLS`](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:TARGET_RUNTIME_DLLS) - This generator expression provides a more convenient option for automatically copying all upstream DLLs to wherever they are needed. The CMake project template already takes advantage of this.
 * [Policy 0110](https://cmake.org/cmake/help/latest/policy/CMP0110.html) - Allows for spaces and special characters in test names. This lets us add new tests that match Catch2's test name style that uses brackets and spaces.

ChrisThrasher:
## Description

Closes #2511

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 3.22)
 
 # define a macro that helps defining an option
 macro(sfml_set_option var default type docstring)
@@ -491,12 +491,7 @@ set(CPACK_NSIS_INSTALLER_MUI_ICON_CODE "!define MUI_WELCOMEFINISHPAGE_BITMAP \\\
 include(CPack)
 
 # configure extras by default when building SFML directly, otherwise hide them
-if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
-    set(SFML_CONFIGURE_EXTRAS_DEFAULT TRUE)
-else()
-    set(SFML_CONFIGURE_EXTRAS_DEFAULT FALSE)
-endif()
-sfml_set_option(SFML_CONFIGURE_EXTRAS ${SFML_CONFIGURE_EXTRAS_DEFAULT} BOOL "TRUE to configure extras, FALSE to ignore them")
+sfml_set_option(SFML_CONFIGURE_EXTRAS ${PROJECT_IS_TOP_LEVEL} BOOL "TRUE to configure extras, FALSE to ignore them")
 
 if(NOT SFML_CONFIGURE_EXTRAS)
     return()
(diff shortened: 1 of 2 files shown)


ChrisThrasher:

ChrisThrasher:
Some of the extended CI jobs are failing because they only have 3.18.4 installed

eXpl0it3r:
I upgraded CMake on my Raspberry Pi: https://snapcraft.io/install/cmake/raspbian

binary1248:
All builders should be upgraded now.

For reference if it is too hard to get a newer CMake somewhere you can always build it using the existing one:

```
sudo apt install libssl-dev
wget https://github.com/Kitware/CMake/releases/download/v3.26.3/cmake-3.26.3.tar.gz
tar xzf cmake-3.26.3.tar.gz
cd cmake-3.26.3/
cmake .
make -j5
sudo make install
hash -r
cmake --version
```

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:
The raspbian job still needs it's CMake updated

eXpl0it3r:
> All builders should be upgraded now.

Your RaPi is still finding CMake 3.18 in its path somehow.

binary1248:
Should be fixed now.

binary1248 (approved):

eXpl0it3r (approved):

PR #2546, opened 2023-05-13
Manage `AAsett*` with `std::unique_ptr`

ChrisThrasher:
## Description

I discovered this simplification while working on #2280. ~~I removed all the `if`/`else` blocks since there is no reason `m_file` should ever be null. I checked the [Android docs](https://developer.android.com/ndk/reference/group/asset#aassetmanager_open) and that function makes no mention of a potential `nullptr` return value.~~ Apparently that pointer can be null if the file fails to load. I still think it's better to `assert` against such a failure but for the sake of not rocking the boat so much I reverted those changes.


ChrisThrasher:

ChrisThrasher:
As an aside, why is `sf::priv::ResourceStream` used as a pointer within`sf::Font`? Is it to hide the ResourceStream.hpp header from users?

binary1248 (approved):

ChrisThrasher:

ChrisThrasher:

Bromeon (approved):

JonnyPtn:
Does it make sense to take ownership of a pointer we don’t own? It also seems necessary to close the asset so I don’t think that should be removed

src/SFML/System/Android/ResourceStream.hpp
```diff
@@ -98,10 +92,18 @@ class SFML_SYSTEM_API ResourceStream : public InputStream
     std::int64_t getSize() override;
 
 private:
+    ////////////////////////////////////////////////////////////
+    // Types
+    ////////////////////////////////////////////////////////////
+    struct AAssetDeleter
```
kimci86, line 98: Not entirely sure but I guess exporting this is needed for compiler-generated destructor to find this when ResourceStream is used in other modules (e.g. Graphics module).
```suggestion
    struct SFML_SYSTEM_API AAssetDeleter
```

ChrisThrasher: You can thank MinGW for that. No other compiler demanded we export this symbol. I know we've had to do this once before for another private type but I forget where.

kimci86: Nevermind, the problem we had in a similar case in `InputSoundFile` (see b48f4b70a7a9781af9b69c6f8a4382a45594999f) was only with MinGW. I don't think it will be problem for this Android-specific code.

PR #2547, opened 2023-05-13
Fix broken `sf::swap` overloads

ChrisThrasher:
## Description

`std::swap` won't call any overloads of `swap` in other namespaces which means these two overloads never got called. Since they never got called, their lack of linker visibility went unnoticed. Discovered while writing #2280.


ChrisThrasher:

eXpl0it3r (approved):

binary1248 (approved):

Issue #2552, opened 2023-05-16
2.6.x cannot compile because of missing cstdint library.


dogunbound:
## Subject of the issue

Compile sfml 2.6.x branch on a fresh clone with GCC version 13 causes uint64_t missing errors.

It is due to https://gcc.gnu.org/gcc-13/porting_to.html

## Your environment

* linux / kde
* sfml 2.6.x
* gcc / g++ 13.1.1
* `cmake --build build -j20 && sudo cmake --install build`

## Steps to reproduce

Install newest version of gcc / g++ and build the SFML 2.6.x branch.

## Expected behavior

Compiles and installs with the commands: `cmake --build build -j20 && sudo cmake --install build`

## Actual behavior
```
/home/dogunbound/SFML/SFML/src/SFML/Audio/SoundFileReaderMp3.cpp:68:23: error: ‘uint64_t’ is not a member of ‘std’; did you mean ‘wint_t’?
   68 | int seekCallback(std::uint64_t offset, void* data)
      |                       ^~~~~~~~
      |                       wint_t
/home/dogunbound/SFML/SFML/src/SFML/Audio/SoundFileReaderMp3.cpp:68:40: error: expected primary-expression before ‘void’
   68 | int seekCallback(std::uint64_t offset, void* data)
      |                                        ^~~~
/home/dogunbound/SFML/SFML/src/SFML/Audio/SoundFileReaderMp3.cpp:68:50: error: expression list treated as compound expression in initializer [-fpermissive]
   68 | int seekCallback(std::uint64_t offset, void* data)
      |                                                  ^
/home/dogunbound/SFML/SFML/src/SFML/Audio/SoundFileReaderMp3.cpp: In constructor ‘sf::priv::SoundFileReaderMp3::SoundFileReaderMp3()’:
/home/dogunbound/SFML/SFML/src/SFML/Audio/SoundFileReaderMp3.cpp:111:17: error: invalid conversion from ‘int’ to ‘MP3D_SEEK_CB’ {aka ‘int (*)(long unsigned int, void*)’} [-fpermissive]
  111 |     m_io.seek = seekCallback;
      |                 ^~~~~~~~~~~~
      |                 |
      |                 int
make[2]: *** [src/SFML/Audio/CMakeFiles/sfml-audio.dir/build.make:286: src/SFML/Audio/CMakeFiles/sfml-audio.dir/SoundFileReaderMp3.cpp.o] Error 1
[...]

PR #2553, opened 2023-05-16
Fix 2.6.x failing to compile on gcc 13.1.1 because of missing headers

dogunbound:
I was getting this error until I added this header: /home/dogunbound/SFML/SFML/src/SFML/Audio/SoundFileReaderMp3.cpp:68:23: error: ‘uint64_t’ is not a member of ‘std’;

This is a fix

----

## Description

There is a bug where compile sfml 2.6.x fails because of missing stdint library. This pull request fixes this bug. This is probably a GCC v13 issue and probably why some have yet to encounter it.

Fix is just to include the header in this file.

https://github.com/SFML/SFML/issues/2552

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Run this command to compile SFML 2.6.x:
```
cmake -B build -DCMAKE_BUILD_TYPE=Release
cmake --build build -j20
sudo cmake --install build
```

### DIFF
--- a/src/SFML/Audio/SoundFileReaderMp3.cpp
+++ b/src/SFML/Audio/SoundFileReaderMp3.cpp
@@ -55,6 +55,7 @@
 #include <SFML/System/Err.hpp>
 #include <algorithm>
 #include <cstring>
+#include <cstdint>
 
 
 namespace


ChrisThrasher:

ChrisThrasher (approved):

PR #2554, opened 2023-05-16
Add `sf::Window::getFrameTime`

ChrisThrasher:
## Description

We already do all the work to calculate the frame time so it's really easy to expose that to users if they want it. This PR is merely finishing a feature that has been mostly complete for many years.

The main use case for this is simplifying how users calculate frame time and thus calculate FPS. While a correct implementation of measuring frame time only requires 1 extra `sf::Clock` (see the Island example in this PR) the bigger cost is in how much time it takes for new users to figure out the correct way to do it. It's a better user experience to be told "here a feature for measuring FPS" then be told how to implement it themselves.

### DIFF
--- a/include/SFML/Window/Window.hpp
+++ b/include/SFML/Window/Window.hpp
@@ -198,6 +198,20 @@ public:
     const ContextSettings& getSettings() const;
 
     ////////////////////////////////////////////////////////////
+    /// \brief Get the duration of the previous frame
+    ///
+    /// Before the first call to display() this will return a
+    /// zero-length duration. After the first call to display()
+    /// it returns the duration since constructing the window.
+    /// After that, it returns the duration between the previous
+    /// two display() calls.
+    ///
+    /// \return Duration of previous frame
+    ///
+    ////////////////////////////////////////////////////////////
+    Time getFrameTime() const;
+
+    ////////////////////////////////////////////////////////////
     /// \brief Enable or disable vertical synchronization
     ///
     /// Activating vertical synchronization will limit the number
@@ -284,6 +298,7 @@ private:
     std::unique_ptr<priv::GlContext> m_context;        //!< Platform-specific implementation of the OpenGL context
     Clock                            m_clock;          //!< Clock for measuring the elapsed time between frames
     Time                             m_frameTimeLimit; //!< Current framerate limit
+    Time                             m_frameTime;      //!< Current frame time
 };
 
 } // namespace sf
--- a/src/SFML/Window/Window.cpp
+++ b/src/SFML/Window/Window.cpp
@@ -165,6 +165,13 @@ const ContextSettings& Window::getSettings() const
 
 
 ////////////////////////////////////////////////////////////
+Time Window::getFrameTime() const
+{
+    return m_frameTime;
+}
+
+
+////////////////////////////////////////////////////////////
 void Window::setVerticalSyncEnabled(bool enabled)
 {
     if (setActive())
@@ -213,10 +220,10 @@ void Window::display()
 
     // Limit the framerate if needed
     if (m_frameTimeLimit != Time::Zero)
-    {
         sleep(m_frameTimeLimit - m_clock.getElapsedTime());
-        m_clock.restart();
-    }
+
+    // Record the latest frame time
+    m_frameTime = m_clock.restart();
 }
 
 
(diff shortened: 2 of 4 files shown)


ChrisThrasher:

MarioLiebisch (approved):
Guess this falls under "too obvious".😉

src/SFML/Window/Window.cpp
```diff
@@ -215,7 +222,7 @@ void Window::display()
     if (m_frameTimeLimit != Time::Zero)
     {
         sleep(m_frameTimeLimit - m_clock.getElapsedTime());
-        m_clock.restart();
+        m_frameTime = m_clock.restart();
```
kimci86, line 225: This does not work when there is no framerate limit.

kimci86 (changes requested):

src/SFML/Window/Window.cpp
ChrisThrasher: Oops you're right. Let me fix that.

ChrisThrasher: Should be fixed now

src/SFML/Window/Window.cpp
```diff
@@ -212,11 +219,10 @@ void Window::display()
         m_context->display();
 
     // Limit the framerate if needed
-    if (m_frameTimeLimit != Time::Zero)
-    {
-        sleep(m_frameTimeLimit - m_clock.getElapsedTime());
-        m_clock.restart();
-    }
+    sleep(m_frameTimeLimit - m_clock.getElapsedTime());
```
ChrisThrasher, line 222: `sf::sleep` does nothing if the requested duration is less than or equal to zero so we don't need an extra conditional here. That's already taken care of for us.

https://github.com/SFML/SFML/blob/9855552f6431016f1953e36b2f058fbd239ed725/src/SFML/System/Sleep.cpp#L46

ChrisThrasher: The first display call will still have a non-zero frame time since the timer starts upon construction of the window.

kimci86: Nevermind, you are right. The clock having microsecond precision, it seems very unlikely for elapsed time to be zero between window creation and first display call.

include/SFML/Window/Window.hpp
```diff
@@ -197,6 +197,14 @@ class SFML_WINDOW_API Window : public WindowBase, GlResource
     ////////////////////////////////////////////////////////////
     const ContextSettings& getSettings() const;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the time between the last two display() calls
```
kimci86, line 201: Nitpicking, but it is not clear if calling this is allowed before display has been called tow times, and what it does if it is allowed.

src/SFML/Window/Window.cpp
ChrisThrasher: In either case, I think `sleep` should be modified to instead read ``if (duration > Time::Zero)` because if the requested duration is zero, we don't want to be calling into all the actual sleeping logic.

include/SFML/Window/Window.hpp
ChrisThrasher: That's a good point. This edge behavior is not necessarily intuitive.

ChrisThrasher:

src/SFML/Window/Window.cpp
MarioLiebisch: That's tricky. I think I remember `sleep(0)` still having the effect of signaling the OS's scheduler "I'm done for now". Less important these days with multiple cores, but just doing nothing in general might be the wrong approach here as it might impact perceived/expected behavior. Or in other words: I'm not sure `sleep(0)` should be a generic no-op.

ChrisThrasher:

src/SFML/Window/Window.cpp
ChrisThrasher: I put the conditional back out of an abundance of caution.

ChrisThrasher:

ChrisThrasher:
Closing since it doesn't have enough support from the SFML Team. Hopefully we can reconsider this decision later since this feature feels like such a natural addition to me.

ChrisThrasher:
I routinely see questions in Discord about measuring delta time and point them to this PR as a way of saying "sorry you have to do it yourself". That's not a good answer to give. Evidently this is something users want. I think we should reconsider this PR or something similar to it.

PR #2556, opened 2023-05-17
Use .hpp for Objective-C++ headers

ChrisThrasher:
## Description

All the files I changed are using at least some amount of C++ so they should have this extension to properly signify their contents. I'm struggling to find some documentation for Objective-C++ header file extension conventions. It's well documented that Objective-C uses .m and .h which makes sense. It's also well documented that Objective-C++ source files use .mm. If you extend that pattern using SFML naming conventions that means you end up with .hpp for Objective-C++ headers.


ChrisThrasher:

src/SFML/Window/OSX/SFOpenGLView.hpp
```diff
@@ -68,8 +68,7 @@ class WindowImplCocoa;
 /// the cursor (that was disconnected from the system).
 ///
 ////////////////////////////////////////////////////////////
-@interface SFOpenGLView : NSOpenGLView
-{
+@interface SFOpenGLView : NSOpenGLView {
```
ChrisThrasher, line 71: This got reformatted after changing the extension presumably because now clang-format knows that it's an Objective-C++ file but I'm not entirely certain. Same goes for SFViewController and SFWindowController.

ChrisThrasher: https://github.com/SFML/SFML/blob/9855552f6431016f1953e36b2f058fbd239ed725/.clang-format#L171

That's what `AfterObjCDeclaration` controls. It just bugs out when the file extension isn't `.h`. You can set this to `false` and you'll see that another .h Objective-C file changes how it formats this brace after `@interface` so I know this is the right setting to use in this situation.

ChrisThrasher:
Closing this because it's too unimportant to bother figuring out why a different extension is messing with our tools. We can reopen later if we get a better grip of Objective-C++ conventions or figure out what clang-format is up to.

PR #2558, opened 2023-05-19
Fix spacing of comment

dogunbound:
I was reading the primitives header file to see what v3 deems deprecated. Noticed this out of line comment. Just wanted to fix it


ChrisThrasher:
Sometimes clang-format doesn’t produce perfect results and this is unfortunately one of those instances. Thank you for looking out for docs fix ups but in this case we have to deal with the slightly janky formatting.

PR #2559, opened 2023-05-20
Fix clang-tidy-16 failures

ChrisThrasher:
## Description

I recently started using Fedora 38 which ships clang-tidy-16. That means for the first time we used something newer than clang-tidy-14 on the Linux (and Linux DRM) code. Turns out the `misc-const-correctness` check has improved since then so it caught a lot of places we're missing `const`.


ChrisThrasher:

Bromeon (approved):
Thanks for testing this already before we use it in our CI! 🙂

binary1248 (approved):

PR #2562, opened 2023-05-21
Test for correct network ordering of packets

ChrisThrasher:
## Description

We were previously testing for round-trip correctness after serialization and deserialization but we were not testing that the serialized forms matches network order. It's a safe assumption that all of our development machines and CI runners are Little Endian so these tests won't be truly meaningful until we can run on Big Endian hardware. Given how rare BE hardware is, arguably it doesn't even matter whether we honor network ordering but as it stands now, the API promises as such so it's a good idea to test it.

These tests will give us more feedback on the changes made in https://github.com/SFML/SFML/pull/2380.

### DIFF
--- a/test/Network/Packet.test.cpp
+++ b/test/Network/Packet.test.cpp
@@ -5,6 +5,9 @@
 #include <array>
 #include <limits>
 #include <type_traits>
+#include <vector>
+
+#include <cstddef>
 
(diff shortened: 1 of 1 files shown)


ChrisThrasher:

binary1248 (approved):

PR #2563, opened 2023-05-21
Remove workaround for support old versions of Visual Studio

ChrisThrasher:
## Description

This appears to be a backward compatibility fix for support VS 2015 and _older_ versions. Because VS 2019 is the older version you can use and still compile SFML 3, we don't need to worry about this any more.


ChrisThrasher:

Bromeon (approved):
We have supported versions in CI and that seems green, so... should be good? 🙂

PR #2565, opened 2023-05-24
Prefer named colors

ChrisThrasher:
## Description

Seeing named colors, particularly in public headers, is a nice readability win.

### DIFF
--- a/include/SFML/Graphics/Image.hpp
+++ b/include/SFML/Graphics/Image.hpp
@@ -55,7 +55,7 @@ public:
     /// \param color Fill color
     ///
     ////////////////////////////////////////////////////////////
-    void create(const Vector2u& size, const Color& color = Color(0, 0, 0));
+    void create(const Vector2u& size, const Color& color = Color::Black);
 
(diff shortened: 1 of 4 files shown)


Bromeon:
Great idea! 👍
 
Was there not an issue with potentially undefined global initialization order? We still have the constants as `static const` and not `constexpr`, so I'm not 100% sure what C++ guarantees here:

https://github.com/SFML/SFML/blob/865a50899c5867ddd6956791c7590b24c894b800/include/SFML/Graphics/Color.hpp#L84-L92

ChrisThrasher:
> Was there not an issue with potentially undefined global initialization order?

Not that I'm aware of. I don't think we're initializing anything static variables with these colors.

> We still have the constants as static const and not constexpr, so I'm not 100% sure what C++ guarantees here:

They're declared as `static const` in Color.hpp then defined as `inline constexpr` in Color.inl. Honestly not sure what's going on there.

binary1248 (approved):

eXpl0it3r (approved):

PR #2567, opened 2023-05-30
Test MinGW build in CI

ChrisThrasher:
## Description

The Windows CI image already comes with MinGW installed so it's very simple to start using. Since MinGW is one of the configs most prone to fail, this should help speed up PRs by providing important feedback sooner than otherwise. I hope this can let us remove some jobs from the extended pipeline.


ChrisThrasher:

binary1248 (approved):

PR #2573, opened 2023-06-05
Optimize compilation time by removing unnecessary includes in sfml-network

abd-ennour:

----

## Description
* This pull request addresses the issue #1893, which aims to enhance the compilation time of the SFML.
* The changes in this pull request optimize the include dependencies within the SFML-Network module, reducing the number of included headers to only those that are required.
* This pull request specifically targets the SFML-Network module. The scope of changes is limited to the **CPP files**, and the include statements in the **HPP files** remain untouched to keep the changes more targeted and reduce the overall scope.
## Tasks

* [ ] Tested on Linux
* [X] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

* For testing the changes, I utilized the existing test suite available in the repository.

Thank you in advance for taking the time to review!


ChrisThrasher:
> The 3 failing are all related to formating ?

Yes. They test different versions of our auto formatter. Just make the changes I recommended and they'll pass.

> I expected to find the usage of std::min, std::max functions, ..etc in the codebase

Did you also check that these .cpp are not using anything from those other headers?

abd-ennour:
> > The 3 failing are all related to formating ?
> 
> Yes. They test different versions of our auto formatter. Just make the changes I recommended and they'll pass.
> 
> > I expected to find the usage of std::min, std::max functions, ..etc in the codebase
> 
> Did you also check that these .cpp are not using anything from those other headers?
Yes, i checked the other includes. I meant to say that I found  #include<algorithm> which lead me to check the headers in the other CPP files of the module.

ChrisThrasher:
I fixed the formatting myself.

ChrisThrasher:
@vittorioromeo has done a lot of work with how we handle headers so hopefully he can take a look at this.

abd-ennour:
> I fixed the formatting myself so don't worry about that.

Think you !

src/SFML/Network/Ftp.cpp
```diff
@@ -30,15 +30,8 @@
 
 #include <SFML/System/Err.hpp>
 
-#include <algorithm>
```
vittorioromeo, line 33: Are you sure you can remove this? There's a call to `std::copy` in this file.

vittorioromeo (changes requested):

src/SFML/Network/Ftp.cpp
abd-ennour: You're right, I missed a few lines. my bad 😕 
This means that "Algorithm" is already included indirectly in "SFML/Network/Ftp.hpp", "SFML/Network/IpAddress.hpp" or "SFML/System/Err.hpp", so deleting it will not result in any significant savings in compilation time.

I found two inclues in the "Unix" and "Win32" folders of the module Network: 
[Unix/SocketImpl.cpp](src/SFML/Network/Unix/SocketImpl.cpp) : the header `<ostream>`
[Win32/SocketImpl.cpp](src/SFML/Network/Win32/SocketImpl.cpp) : the header `<cstring>`

After examination, I found that they are not required.  What are your thoughts on this?

eXpl0it3r:
This seems to be contradicting a bit what #2902 implements.

@vittorioromeo if you think this PR doesn't make sense, feel free to close it.

vittorioromeo:
> This seems to be contradicting a bit what #2902 implements.
> 
> @vittorioromeo if you think this PR doesn't make sense, feel free to close it.

I'm going to re-review this one manually. Removing a header in a `X.cpp` file when the same header is included in `X.hpp` file doesn't help at all, so those are not really useful optimizations. But there might be a header or two that I missed in the IWYU PR, as I did apply the changes manually after all.

src/SFML/Network/Win32/SocketImpl.cpp
```diff
@@ -27,8 +27,6 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Network/Win32/SocketImpl.hpp>
 
-#include <cstring>
```
vittorioromeo, line 30: This one can be removed.

src/SFML/Network/TcpSocket.cpp
```diff
@@ -32,7 +32,6 @@
 
 #include <SFML/System/Err.hpp>
 
-#include <algorithm>
```
vittorioromeo, line 35: This one cannot be removed, there is a `std::min` call below.

src/SFML/Network/SocketSelector.cpp
```diff
@@ -31,10 +31,7 @@
 
 #include <SFML/System/Err.hpp>
 
-#include <algorithm>
-#include <memory>
```
vittorioromeo, line 35: Cannot be removed, we use `make_unique`.

src/SFML/Network/SocketSelector.cpp
```diff
@@ -31,10 +31,7 @@
 
 #include <SFML/System/Err.hpp>
 
-#include <algorithm>
```
vittorioromeo, line 34: Already removed in #2902

src/SFML/Network/SocketSelector.cpp
```diff
@@ -31,10 +31,7 @@
...
-#include <memory>
 #include <ostream>
-#include <utility>
```
vittorioromeo, line 37: Cannot be removed, we use `swap`.

src/SFML/Network/Http.cpp
```diff
@@ -30,8 +30,6 @@
 #include <SFML/System/Err.hpp>
 #include <SFML/System/Utils.hpp>
 
-#include <iterator>
```
vittorioromeo, line 33: Cannot be removed, we use `istreambuf_iterator`.

src/SFML/Network/Http.cpp
```diff
@@ -30,8 +30,6 @@
...
-#include <limits>
```
vittorioromeo, line 34: Cannot be removed, we use `numeric_limits`.

src/SFML/Network/Ftp.cpp
```diff
@@ -32,13 +32,7 @@
 
 #include <algorithm>
 #include <fstream>
-#include <iterator>
-#include <ostream>
 #include <sstream>
-#include <utility>
-
-#include <cctype>
-#include <cstdio>
```
vittorioromeo, line 41: These were already tweaked in #2902.

vittorioromeo (changes requested):

vittorioromeo:
I missed one `<cstring>` header. I am gonna open another PR for that one, this one can be closed. Thanks for the PR anyway!

Generally speaking, if a `X.cpp` file includes the same header included in `X.hpp`, it's not worthwhile removing it from the `X.cpp` file -- that will not improve compilation times and just create a transitive dependency on `X.hpp`.

PR #2577, opened 2023-06-20
Don't overwrite module path

ChrisThrasher:
## Description

The Conan package for SFML has to fix this so apparently it's a real world problem.

https://github.com/conan-io/conan-center-index/blob/master/recipes/sfml/all/patches/0001-cmake-robust-find-deps.patch


ChrisThrasher:
Retargeted for 2.6.1 to allow for a little more cleanup of how SFML 2 is packaged

binary1248 (approved):

Issue #2208, opened 2022-09-20
Make sf::VertexArray a plain struct with public members


therocode:
At the moment `sf::VertexArray` has two data members `std::vector<Vertex> m_vertices` and `PrimitiveType m_primitiveType` which are encapsulated by being private along with some public methods like a plain get/set pair for primitive type, and a limited set of vector-like accessors/modifiers for vertices.

My argument: This encapsulation buys nothing for the user and actually severly limits the power of `sf::VertexArray` unnecessarily. Even conceptually, this is just a list of vertices + a primitive type. This is perfectly represented by a public `std::vector<Vertex>` and a public `PrimitiveType`. 
- There is no validation behind the API that ensures that the primitive type and vertices are set in a way that makes sense to one another.
- There is no real restriction on what data can be put in, since primitive type is a plain get/set which is fully permissive. And for the vertices, you have `clear()` and `push_back()` and `operator[]` which all also end up being fully permissive. So nothing significant is encapsulated.

My proposal: 
- Change `class VertexArray` into `struct VertexArray`
- Make the members public, and rename them to `vertices` and `primitiveType` (stripping `m_` since they are now part of the pulbic API)
- Remove:
- - `getVertexCount` (now `.vertices.size()`)
- - `operator[]` (now `.vertices[]`)
- - `clear()` (now `.vertices.clear()`)
- - `resize()` (now `.vertices.resize()`)
- - `append()` (now `.vertices.push_back()`)
- - `[set/get]PrimitiveType()` (now `.primitiveType` and `.primitiveType = `)
- Keep:
- - `virtual void draw()`, will still inherit drawable and should hence be kept.
- - `getBounds()`, useful piece of functionality that is not provided by the now public members.

The main benefit of this is that the user gets the full power of the `vector` API when working with vertices:
- Iterators (use with `<algorithm>`, range-based for, fits with 3rd party APIs, etc)
- `.reserve()` for allocation-free `push_back`
- `.at()` safe indexing
- Moving of vertex data in/out of the struct, useful f.ex. if vertices arrive from other parts of the codebase
- ...and so on

Another benefit is familiarity. Every C++ programmer knows how to use `vector`. Why have a stripped down restrictive version of it that is different? `.getVertexCount()` and `.append()` instead of the familiar `.size()` and `.push_back()`. 

[...]

Issue #1752, opened 2021-02-19
(suggestion) sf::Packet need a "reserve" method


JonathSpirit:
class sf::Packet need a reserve method.

It can be really great to have a sf::Packet::reserve(size_t n) method to optimize big workload on packet.

Currently, any call to operator << (or append) will trigger a reallocation. This can be improve if the devlopper know exactly the size we want to append when having multiple variables to packed.

A simple implementation will be like that : 
`
void Packet::reserve(std::size_t sizeInBytes)
{
        m_data.reserve(sizeInBytes);
}
`

This is simply a suggestion, please tell me what you think

Bromeon:
> Currently, any call to operator << (or append) will trigger a reallocation. 

Not exactly. `sf::Packet` uses an underlying `std::vector`, which has an exponential growth strategy. This means adding an element is amortized O(1). 

`reserve()` would only be helpful if you create lots of packets, and for each of them you add very small elements (thus rather `<<` than `append()`). However, there's a much better way to migitate this: reuse the `sf::Packet` instance with `clear()`. Not only will this avoid reallocations (the buffer doesn't shrink), but it will also make the initial allocation unnecessary (which `reserve()` still requires).

Unless you experienced real performance issues with `sf::Packet`, I would say the above-mentioned usage gives better optimization results -- I would thus advocate to keep the API simple.

1aam2am1:
O would say that keeping and reusing packet is hard. In my program as it is now i can create packet, then place my data and send it. 

But with reusing i would need static packet. As many thread could use my function i would need some mutex to guard my static variable or create static local thread variable so that each thread in this function has his own packet. Now there is second problem i should check all code that this function use so that there would not be any recursion as this would corupt my data.

Another idea for reusing packet would be to create shared packet in factory and then when we have used it completely releasing it to factory or checking in factory each time when we need new one if some shared packet has one use_cound meaning only factory has pointer to it.

Third idea is too keep packet as member of some class as this class will use this packed for sending data. It don't have as many problems as above, but it can't be used everywhere. And it will hoar memory as long as it exists. Example: Program sends data with size of 10kb. And one time it send 1gb file. Now our packet will have vector with 2gb declared space as long as it exists. Even when we need only 10kb for running.

Each and every idea how to reuse packet is somewhat wrong or problematic.

The best idea would be to declare how much wee will need. 

Allocations in vector are in 2*existing size as mentioned above. But wee need to copy existing data. While this takes away time.

[...]

PR #2587, opened 2023-06-23
Adding reserve() to VertexArray

HerrNamenlos123:

----

## Description

Hi there, thank you so much for this amazing library!
I am trying to implement a Batch Renderer and therefore need all the performance I can get. I added a small change to also expose the `reserve()` function besides the `resize()` function. Any problems with that? :)

I also added a unit test, although we cannot really check if the allocation works, we can only check that the function is callable.

## Tasks

* [ ] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
void resize() {
    sf::VertexArray va;
    va.reserve(50000);       // Thanks to this function all 50000 elements are dynamically appended, with only a single heap allocation
    for (int i = 0; i < 50000; i++) {
        va.append(sf::Vertex(sf::Vector2f(0, 0), sf::Color::Red));
    }
}
```

### DIFF
--- a/include/SFML/Graphics/VertexArray.hpp
+++ b/include/SFML/Graphics/VertexArray.hpp
@@ -123,10 +123,10 @@ public:
     /// used to reduce the number of memory allocations when
     /// a high number of vertices is appended.
     ///
-    /// \param maxVertexCount New capacity of the array (number of vertices)
+    /// \param capacity New capacity of the array (number of vertices)
     ///
     ////////////////////////////////////////////////////////////
-    void reserve(std::size_t maxVertexCount);
+    void reserve(std::size_t capacity);
 
     ////////////////////////////////////////////////////////////
     /// \brief Resize the vertex array
--- a/include/SFML/Graphics/VertexArray.hpp
+++ b/include/SFML/Graphics/VertexArray.hpp
@@ -115,6 +115,20 @@ public:
     void clear();
 
     ////////////////////////////////////////////////////////////
+    /// \brief Reserve memory space for the vertex array
+    ///
+    /// Allocates enough memory so that at least the given number
+    /// of vertices can be stored, without another memory allocation.
+    /// No vertices are added or removed. This function is solely
+    /// used to reduce the number of memory allocations when
+    /// a high number of vertices is appended.
+    ///
+    /// \param maxVertexCount New capacity of the array (number of vertices)
+    ///
+    ////////////////////////////////////////////////////////////
+    void reserve(std::size_t maxVertexCount);
+
+    ////////////////////////////////////////////////////////////
     /// \brief Resize the vertex array
     ///
     /// If \a vertexCount is greater than the current size, the previous
--- a/src/SFML/Graphics/VertexArray.cpp
+++ b/src/SFML/Graphics/VertexArray.cpp
@@ -70,6 +70,13 @@ void VertexArray::clear()
 
 
 ////////////////////////////////////////////////////////////
+void VertexArray::reserve(std::size_t maxVertexCount)
+{
+    m_vertices.reserve(maxVertexCount);
+}
+
+
+////////////////////////////////////////////////////////////
 void VertexArray::resize(std::size_t vertexCount)
 {
     m_vertices.resize(vertexCount);
--- a/src/SFML/Graphics/VertexArray.cpp
+++ b/src/SFML/Graphics/VertexArray.cpp
@@ -70,9 +70,9 @@ void VertexArray::clear()
 
 
 ////////////////////////////////////////////////////////////
-void VertexArray::reserve(std::size_t maxVertexCount)
+void VertexArray::reserve(std::size_t capacity)
 {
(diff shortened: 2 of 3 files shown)


Marioalexsan:
There's a [pull request](https://github.com/SFML/SFML/issues/2208) related to VertexArrays that aims to outright expose the internal array, since there is little point in encapsulating it. Could you look at this too and see whenever this could be more useful?

HerrNamenlos123:
Thank you, didn't know the discussion exists. If that PR gets resolved at some point, my PR is irrelevant. I think making it a struct or removing it completely is the best option.

Bromeon:
See also related discussion in the past in `sf::Packet`: https://github.com/SFML/SFML/issues/1752#issuecomment-820194385

danieljpetersen:
Does this have to be blocked on that other issue? This seems like a very uncontroversial change and purely a net benefit. There is pushback and no consensus on turning the vertex array into a plain struct, I can imagine that proposal getting stuck in limbo for a long time (and indeed, it already has been a while). Seems like the preferable course would be to merge this in and take the win now, and if later it's decided to do something more drastic this isn't harming that effort.

Bambo-Borris:
> Does this have to be blocked on that other issue? This seems like a very uncontroversial change and purely a net benefit. There is pushback and no consensus on turning the vertex array into a plain struct, I can imagine that proposal getting stuck in limbo for a long time (and indeed, it already has been a while). Seems like the preferable course would be to merge this in and take the win now, and if later it's decided to do something more drastic this isn't harming that effort.


I agree about the concerns of it being a great idea lost if blocked by the struct one, but also the struct one would entirely undo the need for this. I think I'd rather see someone open a prospective PR for the struct change and have it championed. I took a look, and there's lots of discussion on the issue, but I didn't see any link to a potential implementation. An actual PR for the struct issue would have a fighting chance. Just remaining an issue will lead to never-ending conversation.

include/SFML/Graphics/VertexArray.hpp
```diff
@@ -114,6 +114,20 @@ class SFML_GRAPHICS_API VertexArray : public Drawable
...
+    /// Allocates enough memory so that at least the given number
+    /// of vertices can be stored, without another memory allocation.
+    /// No vertices are added or removed. This function is solely
+    /// used to reduce the number of memory allocations when
+    /// a high number of vertices is appended.
+    ///
+    /// \param maxVertexCount New capacity of the array (number of vertices)
+    ///
+    ////////////////////////////////////////////////////////////
+    void reserve(std::size_t maxVertexCount);
```
ChrisThrasher, line 129:
```suggestion
    /// \param capacity New capacity of the array (number of vertices)
    ///
    ////////////////////////////////////////////////////////////
    void reserve(std::size_t capacity);
```

src/SFML/Graphics/VertexArray.cpp
```diff
@@ -69,6 +69,13 @@ void VertexArray::clear()
 }
 
 
+////////////////////////////////////////////////////////////
+void VertexArray::reserve(std::size_t maxVertexCount)
+{
+    m_vertices.reserve(maxVertexCount);
```
ChrisThrasher, line 75:
```suggestion
void VertexArray::reserve(std::size_t capacity)
{
    m_vertices.reserve(capacity);
```

ChrisThrasher (changes requested):
I think I'm okay with this PR, but it comes with the caveat that we still may refactor `sf::VertexArray` into an aggregate type or something else different from its current form.

ChrisThrasher:
https://github.com/SFML/SFML/compare/master...ChrisThrasher:SFML:vertex_array

I reimplemted `sf::VertexArray` to have all public data and it has some nice upsides. We delete 200 lines of code and can now leverage `std::vector` features that were previously inaccessible.

ChrisThrasher:
@eXpl0it3r Do you want to close this based on your comments [here](https://github.com/SFML/SFML/issues/2208#issuecomment-1606299436)?

shane-tran-whitmire:
I mean, as long as there is a way to do this highly custom functionality (there is if you just make and manage your own std::vector<sfml::Vertex>), then i think it is fine not to have this function.

ChrisThrasher:
> I mean, as long as there is a way to do this highly custom functionality

I believe so but I have yet to someone illustrate that.

https://github.com/SFML/SFML/blob/11b73743c42cf7ecd7c596ba83fdbf1150ffa96c/include/SFML/Graphics/RenderTarget.hpp#L248-L249

The above function (that gets inherited by `sf::RenderWindow`) seems to be how you're supposed to draw a vector of vertices without using `sf::VertexArray`. I believe this is what you'll be forced to use if you want to use any `std::vector` interfaces that `sf::VertexArray` doesn't already expose.

ChrisThrasher:
I personally wish we could merge this feature but there isn't consensus from the other maintainers so I'm going to close this.

Issue #2588, opened 2023-06-23
Glyph corruption in sf::Font when using GL_SCISSOR_TEST


pekfos:
## Subject of the issue

I have observed rendering artifacts when using `sf::Font` with `GL_SCISSOR_TEST`. I tracked it down to `Texture::update(Texture&, ..)`  performed internally in `sf::Font` when rendering new glyphs, in the codepath using `GLEXT_framebuffer_blit`. Copying glyphs to larger texture with `GL_SCISSOR_TEST` enabled may corrupt them and they will stay that way even if `GL_SCISSOR_TEST` is later disabled, resulting in seemingly random glyphs going permanently missing or corrupted for particular font sizes.

## Your environment

Windows 10, Visual Studio 2019, master branch of SFML

## Steps to reproduce

```cpp
#include <SFML/Graphics.hpp>
#include <SFML/OpenGL.hpp>

int main()
{
    sf::RenderWindow okno( sf::VideoMode( {320, 240 }), "Scissors" );
    sf::Font font;
    // https://fonts.google.com/specimen/Charis+SIL/about
    if(!font.loadFromFile("CharisSIL-Regular.ttf"))
        return 0;

    sf::Text text(font, "0123456789ABCDEFGHIJKLMNOPQ", 30);
    text.setFillColor(sf::Color::White);

    glEnable(GL_SCISSOR_TEST);
    glScissor(0, 0, 30, 30);
    okno.draw(text); // Trigger caching several glyphs with GL_SCISSOR_TEST active.
    glDisable(GL_SCISSOR_TEST);

    while( okno.isOpen() )
    {
        sf::Event event;
        while( okno.pollEvent( event ) )
        {
            if( event.type == sf::Event::Closed )
                    okno.close();

        } //while
        okno.clear();
        okno.draw(text);
        okno.display();
    }
}
```

eXpl0it3r:
As mentioned on #2589 correctly handling of OpenGL states, while using the graphics module, is the user's responsibility. You can use `pushGLStates()` / `popGLStates()` or `resetGLStates()` to let SFML deal with its own OpenGL states, but you still have to enable/disable your own states, as SFML won't be tracking that.

See: https://www.sfml-dev.org/tutorials/2.6/window-opengl.php#using-opengl-together-with-the-graphics-module

pekfos:
[...]

PR #2589, opened 2023-06-23
Fixed GL_SCISSOR_TEST affecting sf::Font internals

pekfos:
## Description

Added a sentry object for disabling and re-enabling `GL_SCISSOR_TEST`, if it was enabled when calling `Texture::update`. This may be called from within `sf::Font` when new glyph is rendered. 

This PR is related to the issue #2588

## Tasks

* [ ] Tested on Linux
* [X] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Following code shows missing or corrupted glyphs when `okno.draw(text)` is called with `GL_SCISSOR_TEST` enabled. Tweaking font size or number of unique characters in definition of `text` may be required for reproduction, in case there are other factors I am not aware of.

```cpp
#include <SFML/Graphics.hpp>
#include <SFML/OpenGL.hpp>

int main()
{
    sf::RenderWindow okno( sf::VideoMode( {320, 240 }), "Scissors" );
    sf::Font font;
    // https://fonts.google.com/specimen/Charis+SIL/about
    if(!font.loadFromFile("CharisSIL-Regular.ttf"))
        return 0;

    sf::Text text(font, "0123456789ABCDEFGHIJKLMNOPQ", 30);

    glEnable(GL_SCISSOR_TEST);
    glScissor(0, 0, 30, 30);
    okno.draw(text); // Trigger caching several glyphs with GL_SCISSOR_TEST active.
    glDisable(GL_SCISSOR_TEST);

    while( okno.isOpen() )
    {
        sf::Event event;
        while( okno.pollEvent( event ) )
        {
            if( event.type == sf::Event::Closed )
                    okno.close();

        } //while
        okno.clear();
        okno.draw(text);
        okno.display();
    }
}
```

### DIFF
--- a/src/SFML/Graphics/Texture.cpp
+++ b/src/SFML/Graphics/Texture.cpp
@@ -57,6 +57,28 @@ std::uint64_t getUniqueId() noexcept
 
     return id.fetch_add(1);
 }
+
+// Automatic wrapper for temporarily disabling GL_SCISSOR_TEST
+class ScissorSaver
+{
+public:
+	ScissorSaver() :
+		m_wasEnabled(glIsEnabled(GL_SCISSOR_TEST))
+	{
+		if(m_wasEnabled)
+			glCheck(glDisable(GL_SCISSOR_TEST));
+	}
+
+	~ScissorSaver()
+	{
+		if(m_wasEnabled)
+			glCheck(glEnable(GL_SCISSOR_TEST));
+	}
+
+private:
+	bool m_wasEnabled;
+};
+
 } // namespace TextureImpl
 } // namespace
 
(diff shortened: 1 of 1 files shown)


ChrisThrasher:
https://www.sfml-dev.org/tutorials/2.6/window-opengl.php#using-opengl-together-with-the-graphics-module

I talked with @binary1248 about this and was told that you need to be careful when manipulating OpenGL states while using the Graphics module since SFML shares the same global state machine with your code. The above link should explain that a bit more and how to work around this limitation of OpenGL.

PR #2590, opened 2023-06-25
Make glyph a struct

dogunbound:
## Make Glyph a struct

This is probably the most egregious example of a struct masking as a class. The reason being the coding standard wants to use classes over structs, but do we want to keep following this rule? It's fine if we want to stick to what we are currently doing, but we could at least talk about this coding standard. If anything, we can reference back to here why we don't want to use raw structs for data types.

### DIFF
--- a/include/SFML/Graphics/Glyph.hpp
+++ b/include/SFML/Graphics/Glyph.hpp
@@ -38,12 +38,8 @@ namespace sf
 /// \brief Structure describing a glyph
 ///
 ////////////////////////////////////////////////////////////
-class SFML_GRAPHICS_API Glyph
+struct SFML_GRAPHICS_API Glyph
 {
-public:
-    ////////////////////////////////////////////////////////////
-    // Member data
-    ////////////////////////////////////////////////////////////
     float     advance{};   //!< Offset to move horizontally to the next character
     int       lsbDelta{};  //!< Left offset after forced autohint. Internally used by getKerning()
     int       rsbDelta{};  //!< Right offset after forced autohint. Internally used by getKerning()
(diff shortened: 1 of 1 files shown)


ChrisThrasher (approved):
Good find! If a type is 100% public data and nothing else, it's a textbook case of using a `struct`.

ChrisThrasher:

eXpl0it3r (approved):
Fine for me, I'm interested in hearing @Bromeon's view on this, since I assume, we hadn't picked `class` randomly over `struct`.

Bromeon (approved):
This has been a `class` [since before 2009.](https://github.com/SFML/SFML/commit/3a34f81561731f3e981b4e47fc40d5de67c0c734#diff-09707b91ad364dcf368783f94574e8b0cdb0286fc22885e37c054931dcf5dd6c)

I don't think there was a particular reason for it other than "everything is a class". So good for me.

Bromeon:
Nitpick: I don't think this qualifies as `bug`. In other projects I've made really good experience with a `qol` or `quality-of-life` label, which includes all sorts of refactorings and chores.

eXpl0it3r:
> Nitpick: I don't think this qualifies as `bug`. In other projects I've made really good experience with a `qol` or `quality-of-life` label, which includes all sorts of refactorings and chores.

As you can see from the history, we had different views on this categorization. 😄

IMHO we don't really need an additional label, as it won't exactly solve the "categorization" problem, but just spread it across three instead of two options - when does something count as "qol" and when as "feature".

I recommend to mark every "change" as feature and reserve bug really just for broken behavior or unintended effects.

So for the given example, using struct instead of class is an intentional decision and a change from how it has been, thus it's a feature.

ChrisThrasher:
I wasn't thinking very hard when I added the "bug" label ¯\\\_(ツ)_/¯

Bromeon:
> I recommend to mark every "change" as feature and reserve bug really just for broken behavior or unintended effects.

That sounds like a good rule 👍

PR #2591, opened 2023-06-27
Remove C++11 header

ChrisThrasher:
## Description

This was caught by someone writing a Conan package for SFML 2.6.0.

https://github.com/conan-io/conan-center-index/pull/18069#discussion_r1243817430

In practice this is a meaningless change but I'm fixing it for the sake of pedantry. Because C99 already provides fixed-width integers, we can rely on them. We can't use `sf::Uint64` here since it's an `unsigned long` on some platforms and `unsigned long long` on others which apparently causes issues with using it as a parameter in a function pointer as is the case on line 112.


ChrisThrasher:

PR #2596, opened 2023-06-29
Added assert messages to Vector3 functions

erikbolumburu11:
## Description
Adds some assert messages to 2 Vector3 functions to make why the program is failing clearer.

## Tasks

* [x] Tested on Windows

## Test

```cpp
void Test() {
	sf::Vector3f v1(0.f, 0.f, 0.f);
	v1.normalized();
	sf::Vector3f v2(1.f, 1.f, 1.f);
	v2.cwiseDiv({ 0.f, 0.f, 0.f });
	v2.cwiseDiv({ 1.f, 0.f, 0.f });
	v2.cwiseDiv({ 1.f, 1.f, 0.f });
}

// Errors:
Assertion failed: *this != Vector3<T>() && "Vector3::normalized() cannot normalize a zero vector", file C:\Users\erikf\dev\LIBRARIES\SFML3\src\SFML\System\Vector3.cpp, line 41
Assertion failed: rhs.x != 0 && "Vector3::cwiseDiv() cannot divide by 0", file C:\Users\erikf\dev\LIBRARIES\SFML3\include\SFML\System\Vector3.inl, line 92
Assertion failed: rhs.y != 0 && "Vector3::cwiseDiv() cannot divide by 0", file C:\Users\erikf\dev\LIBRARIES\SFML3\include\SFML\System\Vector3.inl, line 93
Assertion failed: rhs.z != 0 && "Vector3::cwiseDiv() cannot divide by 0", file C:\Users\erikf\dev\LIBRARIES\SFML3\include\SFML\System\Vector3.inl, line 94
```

PR #2593, opened 2023-06-28
Assert Messages For Vector2 Functions

erikbolumburu11:
## Description

My pull request adds some assert messages into the new Vector2 functions from SFML3 that make it clearer that using a zero vector with these functions will cause an error.

This PR is related to the issue #2592

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## Testing

```cpp
#include <SFML/Graphics.hpp>

void Test() {
	sf::Vector2<float> v1;
	sf::Vector2<float> v2(1.0f, 1.0f);
	v1.normalized();
	v1.angleTo({ 1.0f, 1.0f });
	v2.angleTo({ 0.0f, 0.0f });
	v1.angle();
	v2.projectedOnto({ 0.0f, 0.0f });
}
```
### Resulting Errors
```c++
Assertion failed: ("Vector2::normalized() cannot normalize a zero vector", *this != Vector2<T>()), file C:\Users\erikf\dev\LIBRARIES\SFML3\src\SFML\System\Vector2.cpp, line 41
Assertion failed: ("Vector2::angleTo() cannot calculate angle from a zero vector", *this != Vector2<T>()), file C:\Users\erikf\dev\LIBRARIES\SFML3\src\SFML\System\Vector2.cpp, line 52
Assertion failed: ("Vector2::angleTo() cannot calculate angle to a zero vector", rhs != Vector2<T>()), file C:\Users\erikf\dev\LIBRARIES\SFML3\src\SFML\System\Vector2.cpp, line 53
Assertion failed: ("Vector2::angle() cannot calculate angle from a zero vector", *this != Vector2<T>()), file C:\Users\erikf\dev\LIBRARIES\SFML3\src\SFML\System\Vector2.cpp, line 64
Assertion failed: ("Vector2::projectedOnto() cannot project onto a zero vector", axis != Vector2<T>()), file C:\Users\erikf\dev\LIBRARIES\SFML3\src\SFML\System\Vector2.cpp, line 90
```

PR #2598, opened 2023-07-02
Add doc strings to remaining assertions

ChrisThrasher:
## Description

Inspired by #2595 and #2596

The main rationale behind this PR is that if an assertion fails, it's often not obvious what went wrong. At best you can see _that_ something failed but unless you're already familiar with SFML's source code, chances are you won't immediately understand _why_ something failed.

I'm sure these messages will need some refining since I myself struggled to understand what some of these assertions mean since they were present in code I don't already understand.

One outstanding question I have is whether we bother including the function name in the assertion message. We did that in the `sf::Vector` assertions but that's not a pattern we've used elsewhere before. Is that a precedent we want to set across the codebase? Is it worth the effort?

### DIFF
--- a/include/SFML/System/Angle.inl
+++ b/include/SFML/System/Angle.inl
@@ -28,7 +28,7 @@ constexpr float pi = 3.141592654f;
 
 constexpr float positiveRemainder(float a, float b)
 {
-    assert(b > 0.0f);
+    assert(b > 0.0f && "Cannot calculate remainder with non-positive divisor");
     const float val = a - static_cast<float>(static_cast<int>(a / b)) * b;
     if (val >= 0.f)
         return val;
--- a/src/SFML/Graphics/Texture.cpp
+++ b/src/SFML/Graphics/Texture.cpp
@@ -453,8 +453,8 @@ void Texture::update(const std::uint8_t* pixels)
 ////////////////////////////////////////////////////////////
 void Texture::update(const std::uint8_t* pixels, const Vector2u& size, const Vector2u& dest)
 {
-    assert(dest.x + size.x <= m_size.x);
-    assert(dest.y + size.y <= m_size.y);
+    assert(dest.x + size.x <= m_size.x && "Destination x coordinate is outside of texture");
+    assert(dest.y + size.y <= m_size.y && "Destination y coordinate is outside of texture");
 
     if (pixels && m_texture)
     {
@@ -497,8 +497,8 @@ void Texture::update(const Texture& texture)
 ////////////////////////////////////////////////////////////
 void Texture::update(const Texture& texture, const Vector2u& dest)
 {
-    assert(dest.x + texture.m_size.x <= m_size.x);
-    assert(dest.y + texture.m_size.y <= m_size.y);
+    assert(dest.x + texture.m_size.x <= m_size.x && "Destination x coordinate is outside of texture");
+    assert(dest.y + texture.m_size.y <= m_size.y && "Destination y coordinate is outside of texture");
 
     if (!m_texture || !texture.m_texture)
         return;
@@ -630,8 +630,8 @@ void Texture::update(const Window& window)
 ////////////////////////////////////////////////////////////
 void Texture::update(const Window& window, const Vector2u& dest)
 {
-    assert(dest.x + window.getSize().x <= m_size.x);
-    assert(dest.y + window.getSize().y <= m_size.y);
+    assert(dest.x + window.getSize().x <= m_size.x && "Destination x coordinate is outside of texture");
+    assert(dest.y + window.getSize().y <= m_size.y && "Destination y coordinate is outside of texture");
 
     if (m_texture && window.setActive(true))
     {
--- a/src/SFML/Window/Unix/Display.cpp
+++ b/src/SFML/Window/Unix/Display.cpp
@@ -82,7 +82,7 @@ void closeDisplay(Display* display)
 {
     const std::lock_guard lock(mutex);
 
-    assert(display == sharedDisplay);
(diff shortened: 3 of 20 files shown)


include/SFML/System/Angle.inl
```diff
@@ -28,7 +28,7 @@ constexpr float pi = 3.141592654f;
 
 constexpr float positiveRemainder(float a, float b)
 {
-    assert(b > 0.0f);
+    assert(b > 0.0f && "Cannot calculate non-positive remainder");
```
kimci86, line 31: This description doesn't look right. Maybe something like this would be more correct: "Computing remainder with non-positive divisor is not allowed".
By the way I am not sure why we have this restriction.

src/SFML/Window/Unix/Display.cpp
```diff
@@ -132,7 +132,7 @@ void closeXim(XIM xim)
 {
     const std::lock_guard lock(mutex);
 
-    assert(xim == sharedXIM);
+    assert(xim == sharedXIM && "XIM must matched shared XIM");
```
kimci86, line 135:
```suggestion
    assert(xim == sharedXIM && "XIM must match shared XIM");
```

test/System/String.test.cpp
```diff
@@ -19,7 +19,7 @@ namespace
 template <typename T>
 auto select(const std::basic_string<T>& string16, const std::basic_string<T>& string32)
 {
-    assert(string16 != string32);
+    assert(string16 != string32 && "Invalid to select between identical inputs");
```
kimci86, line 22: I don't think this one is a correct sentence. What about something like: "Passing identical arguments to select is invalid."

ChrisThrasher: It's just a shorter version of "It is invalid to select between identical inputs". That sounds like correct grammar to me.

eXpl0it3r:
A few general thoughts:
- When an assert happens, does the developer get informed "where" the assert happens or do we need to include the function name, like it was done for #2593 / #2596?
- While we're asserting on internal variables, it might be more useful to have the message
    - Not just mention what failed, but also a bit more why. What's the most probably cause, why something is set to null
    - Consider hinting at what could be done about it. Like "correctly call the constructor" or something like that, so the developer doesn't have to first go and read all of SFML's source ode

Unless we kind of just have those messages for SFML devs 🤔

ChrisThrasher:
> "where" the assert happens or do we need to include the function name, like it was done for

They won't get the function name. I do believe they get file and line number though. I tried to find a clever way to automatically embed the function name but it's not possible as far as I can tell.

```
Assertion failed: (*this != Vector2<T>() && "Vector2::normalized() cannot normalize a zero vector"), function normalized, file Vector2.cpp, line 41
```

> Not just mention what failed, but also a bit more why. What's the most probably cause, why something is set to null

Agreed, I just lack the knowledge to write a more informative message for some of these asserts. I tried my best but there's room for improvement.

> Consider hinting at what could be done about it

I hope to do this where possible. It's a good idea for assertions that check preconditions in public APIs but no so much assertions hidden deep within the implementation put in place to catch implementation bugs. 

> Unless we kind of just have those messages for SFML devs

I have users in mind. Developers already have the source code and are (hopefully) on Discord where they can ask for help. It's users who will never see the source code who most need this.

aardhex:
You might wanna take a look at Boost stacktrace() to get the function name.

ChrisThrasher:
> You might wanna take a look at Boost stacktrace() to get the function name.

I don't believe that work since `assert` is merely showing the code contained within the `assert` so if you call a function, it will just show that you called a function. It won't show that function returned. This is why have to use string literals as the description.

ChrisThrasher:
> Can we assert null or rather nullptr consistently? Personally, I like being explicit (e.g. assert(bla != nullptr)).

I'm cool with a follow on PR to use a unified style. I personally prefer the more terse `assert(m_ptr)` style for what it's worth but am fine with either.

Bromeon:
Talking of Boost, this may be more useful if we're just interested in the function name:
https://www.boost.org/doc/libs/1_82_0/boost/current_function.hpp

ChrisThrasher:
I took another pass at improving a lot of the assertion messages. In many places I found out what function needed to be called to fix the assertion and mentioned that in the message.

I'm still on the fence about whether I care about putting the name of the function inside the message itself, particularly whether that should be a rule we strictly enforce or just something we do whenever we think it's helpful.

ChrisThrasher:
https://github.com/llvm/llvm-project/commit/5e807c38bf9b5c1528012ec9003953b352f99c79

A recent LLVM commit uses this pattern of including the function name in the assertion message

src/SFML/Graphics/Texture.cpp
```diff
@@ -453,8 +453,8 @@ void Texture::update(const std::uint8_t* pixels)
 ////////////////////////////////////////////////////////////
 void Texture::update(const std::uint8_t* pixels, const Vector2u& size, const Vector2u& dest)
 {
-    assert(dest.x + size.x <= m_size.x);
-    assert(dest.y + size.y <= m_size.y);
+    assert(dest.x + size.x <= m_size.x && "Destination x coordinate is outside of texture");
+    assert(dest.y + size.y <= m_size.y && "Destination y coordinate is outside of texture");
```
eXpl0it3r, line 457: Non-native speaker, can one really say "outside of texture" instead of "outside of the texture"?

eXpl0it3r (approved):

src/SFML/Graphics/Texture.cpp
ChrisThrasher: It's not something you'd say verbally but in code it's common to drop "the" for the sake of brevity.

PR #2601, opened 2023-07-04
Remove brew update until GitHub fixes the issue

eXpl0it3r:
## Description

Our pipelines have been failing of the past days, due to some issue with GitHub's homebrew-core and homebrew-cask repositories. The maintainers are in contact with GitHub, but it's unclear how long it takes, until it's fixed.

https://github.com/orgs/Homebrew/discussions/4612

This PR just removes the `brew update` call for now, so we won't run into those issues anymore.

Alternatively, we could also just wait it out.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,8 @@ jobs:
     - name: Install macOS Tools
       if: runner.os == 'macOS'
       run: |
-        brew update
+        # Do not run updates until GitHub has fixed all the issues: https://github.com/orgs/Homebrew/discussions/4612
+        # brew update
         brew install gcovr ninja || true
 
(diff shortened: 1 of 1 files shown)


binary1248 (approved):

ChrisThrasher (commented):
I believe I added this to be extra sure that we'd be able to download Clang 16 instead of Clang 15 but it now appears the Brew caches on the macOS images have updated such that we are getting Clang 16 without running `brew update` first.

eXpl0it3r:
There's also this alternative fix, that someone suggested on the linked discussion, which also seems to work: https://github.com/SFML/SFML/commit/db2666c7db4fd25160a843ef346998c9d8f5c0e1 

If you think that makes more sense, then I'm happy to merge that.

I just wanted to temporarily get rid of the failures 😄

Issue #2583, opened 2023-06-22
Intel compiler unsupported


VictorEijkhout:
It looks like you don't support the (relatively) new Intel 1API compiler:

```
cmake -D CMAKE_INSTALL_PREFIX=/work2/00434/eijkhout/sfml/installation-sfml-2.6.0-clx-intel23                                           -D CMAKE_VERBOSE_MAKEFILE=ON                     -D BUILD_SHARED_LIBS=TRUE                     /work2/00434/eijkhout/sfml/sfml-2.6.0
-- The C compiler identification is IntelLLVM 2023.1.0
-- The CXX compiler identification is IntelLLVM 2023.1.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /opt/intel/oneapi/compiler/2023.1.0/linux/bin/icx - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /opt/intel/oneapi/compiler/2023.1.0/linux/bin/icpx - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at cmake/Config.cmake:134 (message):
  Unsupported compiler
```

VictorEijkhout:
Even the Intel classic compiler:

```
cmake -D CMAKE_INSTALL_PREFIX=/work2/00434/eijkhout/sfml/installation-sfml-2.6.0-clx-intel22                                           -D CMAKE_VERBOSE_MAKEFILE=ON                     -D BUILD_SHARED_LIBS=TRUE                     /work2/00434/eijkhout/sfml/sfml-2.6.0
-- The C compiler identification is Intel 2021.7.0.20220726
-- The CXX compiler identification is Intel 2021.7.0.20220726
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /scratch1/projects/compilers/oneapi_2022.3.0/compiler/latest/linux/bin/intel64/icc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /scratch1/projects/compilers/oneapi_2022.3.0/compiler/latest/linux/bin/intel64/icpc - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
[...]

PR #2607, opened 2023-07-08
Add Intel oneAPI CI job

ChrisThrasher:
## Description

https://www.intel.com/content/www/us/en/developer/tools/oneapi/dpc-compiler.html

This is a pretty unique CI job since so few CI pipelines use compilers that aren't from the MSVC, GCC, or Clang family of compilers. The Intel oneAPI compiler is relatively new and not the same Intel compiler from years past. Getting it installed requires a chunk of shell code but that's as far as the additional complexity goes. It does a perfectly fine job compiling the library.

The only change I had to make was a slightly tweak to some floating point related tests. This isn't entirely surprising. This compiler produced results just slightly outside of the 1*10^-5 margin we apply to all floating point comparisons so I had to increase the margin for that one test. 

If this new job gives us trouble then I don't mind removing it but in light of issues like #2583 I'm eager to add it and see how well it works for us.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,7 @@ jobs:
         - { name: Windows MinGW,          os: windows-2022, flags: -DSFML_USE_MESA3D=TRUE -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -GNinja }
         - { name: Linux GCC,            os: ubuntu-22.04, flags: -DSFML_RUN_DISPLAY_TESTS=ON -GNinja }
         - { name: Linux Clang,          os: ubuntu-22.04, flags: -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DSFML_RUN_DISPLAY_TESTS=ON -GNinja , gcovr_options: '--gcov-executable="llvm-cov-$CLANG_VERSION gcov"' }
+        - { name: Linux Intel oneAPI,   os: ubuntu-22.04, flags: -DCMAKE_C_COMPILER=icx -DCMAKE_CXX_COMPILER=icpx -GNinja }
         - { name: MacOS,                os: macos-12, flags: -GNinja }
         - { name: MacOS Xcode,          os: macos-12, flags: -GXcode }
         - { name: iOS,                  os: macos-12, flags: -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_ARCHITECTURES=arm64 }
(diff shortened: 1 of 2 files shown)


ChrisThrasher:

test/Graphics/ConvexShape.test.cpp
```diff
@@ -103,7 +104,8 @@ TEST_CASE("[Graphics] sf::ConvexShape")
         convex.setPoint(0, {-100000.f, 0.f});
         convex.setPoint(1, {100000.f, 0.f});
         convex.setPoint(2, {100000.f, 0.000001f});
-        CHECK(convex.getGeometricCenter() == Approx(sf::Vector2f(100000.f / 3.f, 0.f)));
+        CHECK(convex.getGeometricCenter().x == Catch::Approx(100000. / 3.).margin(1e-2));
+        CHECK(convex.getGeometricCenter().y == Catch::Approx(0).margin(1e-5));
```
eXpl0it3r, line 108: I assume using something like `std::numeric_limits<T>::epsilon` would not be sufficent, right?

ChrisThrasher: No, we need a margin much larger than that.

eXpl0it3r (approved):

binary1248 (approved):

.github/workflows/ci.yml
```diff
@@ -89,6 +90,18 @@ jobs:
         echo "CLANG_VERSION=$CLANG_VERSION" >> $GITHUB_ENV
         sudo apt-get install gcovr ${{ matrix.platform.name == 'Linux Clang' && 'llvm-$CLANG_VERSION' || '' }}
 
+    - name: Install Intel oneAPI
+      if: matrix.platform.name == 'Linux Intel oneAPI'
+      run: |
+        wget https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
+        sudo apt-key add GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
```
kimci86, line 97: CI warns this is a deprecated command.
Looking it up online, I found an example using more secure commands: https://brave.com/linux/#debian-ubuntu-mint

kimci86 (changes requested):
Maybe we can add a condition not to output a warning from this line: https://github.com/SFML/SFML/blob/20f34d38fd3969bcfa471e11cad074bc3fa6d3d5/cmake/Config.cmake#L95

ChrisThrasher:
> Maybe we can add a condition not to output a warning from this line:

As in not output a warning if this particular compiler is detected? I'd rather not add even more compiler specific code. If anything I'd want to remove that warning entirely since this kind of compiler detection code doesn't belong in cross platform libraries like this.

.github/workflows/ci.yml
ChrisThrasher: For posterity, here's the docs link Kimci shared with me in Discord:

https://www.intel.com/content/www/us/en/docs/oneapi/installation-guide-linux/2023-1/apt.html

I switched over to this approach and it seems to be working!

.github/workflows/ci.yml
```diff
@@ -89,6 +90,17 @@ jobs:
...
         sudo apt-get install gcovr ${{ matrix.platform.name == 'Linux Clang' && 'llvm-$CLANG_VERSION' || '' }}
 
+    - name: Install Intel oneAPI
+      if: matrix.platform.name == 'Linux Intel oneAPI'
+      run: |
+        wget -O- https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB \
+          | gpg --dearmor | sudo tee /usr/share/keyrings/oneapi-archive-keyring.gpg > /dev/null
+        echo "deb [signed-by=/usr/share/keyrings/oneapi-archive-keyring.gpg] https://apt.repos.intel.com/oneapi all main" | sudo tee /etc/apt/sources.list.d/oneAPI.list
+        sudo apt-get update
+        sudo apt-get install intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic intel-oneapi-mpi intel-oneapi-mpi-devel intel-oneapi-mkl
```
kimci86, line 100: I guess those packages are not all needed.
Probably `sudo apt install intel-oneapi-compiler-dpcpp-cpp` is enough.

ChrisThrasher:

ChrisThrasher:

kimci86:
Do we want to make sure SFML compiles and tests pass or do we additionally want to have no warnings doing so?
We would need a few lines of CMake to configure compiler warnings.
Otherwise it looks good to me.

ChrisThrasher:
> Do we want to make sure SFML compiles and tests pass or do we additionally want to have no warnings doing so?
> We would need a few lines of CMake to configure compiler warnings.
> Otherwise it looks good to me.

I looked up the warnings this compiler offers but just never bothered actually adding them. I'm not opposed, I was just a bit lazy when originally writing this.

eXpl0it3r (approved):

ChrisThrasher:
https://github.com/SFML/SFML/actions/runs/5496668671/jobs/10016832805?pr=2607

I added then removed Intel compiler warnings because they open up a new can of worms that I'd rather save for a later PR.

ChrisThrasher:

PR #2608, opened 2023-07-08
Raise clang-format requirement to version 14

### LINKED ISSUE #2568: Raise clang-format requirement to version 14
## Subject of the issue

https://github.com/SFML/SFML/compare/master...ChrisThrasher:SFML:clang-format-14

Above is a commit that takes advantage of some clang-format-14 features. The big one is related to qualifier sorting. Adding that check actually found a few minor inconsistencies. Because we now use one release of clang-format it also let us drop 2 CI jobs for versions 12 and 13.

https://releases.llvm.org/14.0.0/tools/clang/docs/ClangFormatStyleOptions.html

Here are the docs for clang-format-14. We can use anything in here. See the labels that note the version that added a given feature. Search for "clang-format 13" or "clang-format 14" to find new features we can start using.

https://github.com/SFML/SFML/pull/2159#discussion_r916304612

As Bromeon mentioned in the original clang-format PR, we may also consider using `PackConstructorInitializers` which makes the code better align to the style that was present before we started using clang-format.

---

Why clang-format-14 now? In the [original PR](#2159) I pushed for version 12 in part to have seamless compatibility with Ubuntu 20 LTS which ships with clang-format-12 and was still commonly used in mid 2022 when this was merged, just a month after Ubuntu 22's release. Over time that OS is being used less and I don't believe any regular contributors are using it so it's safe to upgrade to something newer. Perhaps it was never a worthwhile version to support in the first place ¯\\\_(ツ)_/¯.

On top of that, there is reason to consider upgrading to clang-format-15 or 16 in the near future. Version 16 is what macOS (via `brew`) and Windows (via Visual Studio) provide so we're causing our regular maintainers on those OSes to either manually format code or install an older clang-format release to format the codebase. That inconvenience plus the new features in newer versions are compelling. Dropping version 12 and 13 is a small step towards that goal.

Why not support clang-format-14 and all newer versions? Sadly clang-format-15 on onwards changes how vertical alignment rules are handled. This results in code that formatted differently between both versions. clang-format does **not** claim to have perfect compatibility between versions. It was a nice coincidence that versions 12 through 14 all produce the same code for SFML in particular but that's not a promise we can rely on. Going forward I expect us to continue to have to use a single version of clang-format at a time unless we're willing to remove those vertical alignment rules and potentially other rules as well.

ChrisThrasher:
## Description

Closes #2568 

Support clang-format-12 mattered a bit a year ago before Ubuntu 22 was as ubiquitous but since then I've heard of precisely zero users who have clang-format 12 or 13 as their default version and can't easily upgrade to 14. Let's go ahead and 12 and 13 so we're one step closer to requiring something more modern like 15 or 16 which are more commonly found on modern OSes.


ChrisThrasher:

eXpl0it3r (approved):

PR #2611, opened 2023-07-09
Emit resize event any time size changes

dogunbound:
Fixes bug where resize events only got triggered upon manual window resizing OR if one called setSize with a height that was too large.

Extension of https://github.com/SFML/SFML/pull/2538 

Thrasher wanted me to put up a PR for 2.6.x fixes for this. I agree that such a simple bug fix should also be included in 2.6.x

### DIFF
--- a/src/SFML/Window/OSX/SFWindowController.mm
+++ b/src/SFML/Window/OSX/SFWindowController.mm
@@ -471,18 +471,9 @@
         // Add titlebar height.
         height += static_cast<unsigned int>([self titlebarHeight]);
 
-        // Corner case: don't set the window height bigger than the screen height
-        // or the view will be resized _later_ without generating a resize event.
-        NSRect screenFrame = [[NSScreen mainScreen] visibleFrame];
-        CGFloat maxVisibleHeight = screenFrame.size.height;
-        if (height > maxVisibleHeight)
-        {
-            height = static_cast<unsigned int>(maxVisibleHeight);
-
-            // The size is not the requested one, we fire an event
-            if (m_requester != 0)
-                m_requester->windowResized(width, height - static_cast<unsigned int>([self titlebarHeight]));
-        }
+        // Send resize event if size has changed
+        if (sf::Vector2u(width, height) != m_requester->getSize())
+            m_requester->windowResized(width, height - static_cast<unsigned int>([self titlebarHeight]));
 
         NSRect frame = NSMakeRect([m_window frame].origin.x,
                                   [m_window frame].origin.y,


src/SFML/Window/OSX/SFWindowController.mm
```diff
@@ -471,18 +471,9 @@ -(void)resizeTo:(unsigned int)width by:(unsigned int)height
         // Add titlebar height.
         height += static_cast<unsigned int>([self titlebarHeight]);
 
-        // Corner case: don't set the window height bigger than the screen height
-        // or the view will be resized _later_ without generating a resize event.
```
kimci86, line 475: Maybe this comment was there for a reason and we have to handle the corner case described here.
I will be able to check on a macbook this week.

ChrisThrasher: If it's resized later without generating an event that doesn't matter because we just emitted the event.

kimci86: Maybe the view will be resized to a different size, one that fits in screen height, without notifying the user.

ChrisThrasher:

kimci86:
I tested and I can confirm this is an improvement but there is some weird behavior when calling setSize with height being greater than available height on screen, as the comment mentioned.


This patch on 2.6.x instead of this PR gave me better results:
```patch
diff --git a/src/SFML/Window/OSX/SFWindowController.mm b/src/SFML/Window/OSX/SFWindowController.mm
index 1da9634a..f6cb3e99 100644
--- a/src/SFML/Window/OSX/SFWindowController.mm
+++ b/src/SFML/Window/OSX/SFWindowController.mm
@@ -483,6 +483,11 @@
             if (m_requester != 0)
                 m_requester->windowResized(width, height - static_cast<unsigned int>([self titlebarHeight]));
         }
+        else if (width != [self size].width || height != [self size].height + static_cast<CGFloat>([self titlebarHeight]))
+        {
+            if (m_requester != 0)
+                m_requester->windowResized(width, height - static_cast<unsigned int>([self titlebarHeight]));
+        }
 
         NSRect frame = NSMakeRect([m_window frame].origin.x,
                                   [m_window frame].origin.y,
```

dogunbound:
@kimci86 could you test this on master and see if it's better to do this on master too?

It might be best just do do this on both branches then.

dogunbound:
New pr up https://github.com/SFML/SFML/pull/2618 

we can compare and see what we like best

kimci86:
> @kimci86 could you test this on master and see if it's better to do this on master too?

@dogunbound Yes, I can test on master next week. (I am mostly away this week)

kimci86:
> @kimci86 could you test this on master and see if it's better to do this on master too?

I confirm there is a small issue on master when setting window size bigger than available height. I opened #2631 to solve it.

ChrisThrasher:
Superseded by #2618

PR #2612, opened 2023-07-10
Add tests for `SFML_USE_SYSTEM_DEPS` option

ChrisThrasher:
## Description

I just realized we have no tests for the `SFML_USE_SYSTEM_DEPS` option so I added this special case CI job.


ChrisThrasher:

ChrisThrasher:

binary1248 (approved):

eXpl0it3r (approved):
Not sure how useful this is to run on Linux, as we're always using system dependencies on Linux to begin with. 🤔

ChrisThrasher:
I'm now realizing that the Linux builds use system deps irrespective of the value of `SFML_USE_SYSTEM_DEPS`. That's a good point. This job is pretty useless as-is. I'll see about using macOS instead so that we're actually testing something meaningful.

ChrisThrasher:
I switched over to using macOS and it still automatically works without having to install anything extra.

ChrisThrasher:

PR #2615, opened 2023-07-13
refactor: add missing `#include`s

JohelEGP:

----

## Description

When building the CMake target `SFML::Graphics`
with a sufficiently recent Clang using [Clang Modules](https://clang.llvm.org/docs/Modules.html)
(with flags `-stdlib=libc++ -fmodules -fbuiltin-module-map -fimplicit-module-maps`)
compilation fails with these (source, error) pairs:
```output
<prefix>/include/SFML/System/Time.inl:47:80: error: no member named 'milli' in namespace 'std'
   47 |     return std::chrono::duration_cast<std::chrono::duration<std::int32_t, std::milli>>(m_microseconds).count();
      |                                                                           ~~~~~^
<prefix>/src/SFML/Window/GlResource.cpp:43:52: error: declaration of 'move' must be imported from module 'std.utility.__utility.move' before it is required
   43 |     priv::GlContext::registerUnsharedGlObject(std::move(object));
      |                                                    ^
```

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Adding missing `#include`s shouldn't require changing how testing is done.

To test the current state of `master` as reported,
follow the description to setup Clang
and see the errors when building the CMake target `SFML::Graphics`.


ChrisThrasher:

src/SFML/Window/GlResource.cpp
```diff
@@ -28,6 +28,7 @@
 #include <SFML/Window/GlContext.hpp>
 #include <SFML/Window/GlResource.hpp>
 
+#include <utility>
```
ChrisThrasher, line 31:
```suggestion
#include <utility>

```
To preserve the two empty lines between the last header and `namespace sf`

ChrisThrasher (changes requested):
Cool to see someone experimenting with newer tools! I follow LLVM's development and I know libc++ has done a lot of work lately to remove unnecessary transitive headers within libc++ so I'm guessing this was caused by that work upstream in LLVM.

One minor tweak then this should be good to merge.

ChrisThrasher (approved):

ChrisThrasher:

PR #2616, opened 2023-07-14
Optimize WGL context creation performance on some systems

binary1248:
Cache the results of WglContext::selectBestPixelFormat to speed up context creation on systems where pixel format selection is an expensive operation.

This is a simpler version of the change proposed in #2508.

While I can't test this change myself (doesn't make much of a difference when using an AMD GPU it seems), it is possible that it might make a more noticeable difference on systems with an Nvidia/Intel GPU.

To test this change just create a lot of contexts and see which variant finishes faster.


kimci86 (commented):
Maybe you could add a little comment in the code telling why this can be useful. The rest of the function has comments, it looks surprising not to have one on this new part.

src/SFML/Window/Win32/WglContext.cpp
```diff
@@ -276,6 +276,31 @@ void WglContext::setVerticalSyncEnabled(bool enabled)
...
+        unsigned int bitsPerPixel;
+        unsigned int depthBits;
+        unsigned int stencilBits;
+        unsigned int antialiasingLevel;
+        bool         pbuffer;
+        int          bestFormat;
+    };
+
+    static std::mutex                         cacheMutex;
+    static std::vector<PixelFormatCacheEntry> pixelFormatCache;
```
MarioLiebisch, line 290: Just random thinking, didn't consider this last time with the original PR: Shouldn't there be one cache per device context instead of one global one? Haven't tried it, but might this screw up things in case displays change at runtime? Let's say switching between a HDR screen and a non-HDR screen?

binary1248: Pixel formats are specific to the GPU that Windows chooses to use for rendering the application window. If you run multiple screens, some being HDR and others SDR you can move the window between them and the OpenGL/GDI/Compositor will take care of the magic for you. Since we can't even select the GPU to render on in OpenGL, having one cache should be good enough.

binary1248:
Updated.

eXpl0it3r:
@lucasz93 does this fix your slow start up?

eXpl0it3r (approved):

ChrisThrasher (approved):

PR #2618, opened 2023-07-15
Fix macOS resize bug

dogunbound:
https://github.com/SFML/SFML/pull/2611

kimci86 stated that this may have better behavior than the previous pull request. I'm going to leave this up and see how we feel about it.

### DIFF
--- a/src/SFML/Window/OSX/SFWindowController.mm
+++ b/src/SFML/Window/OSX/SFWindowController.mm
@@ -483,6 +483,11 @@
             if (m_requester != 0)
                 m_requester->windowResized(width, height - static_cast<unsigned int>([self titlebarHeight]));
         }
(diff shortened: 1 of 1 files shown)


shane-tran-whitmire:
@kimci86 This is what you recommend, right?

kimci86:
> @kimci86 This is what you recommend, right?

Yes this works best for me.

kimci86 (approved):

ChrisThrasher:
@kimci86 was telling me that instead of simply cherry picking #2631 we need to use this PR since some other changes in `sf::WindowBase` require that this PR be a bit different?

ChrisThrasher (approved):
I can confirm this PR makes resizing seem to work just fine on macOS

PR #2619, opened 2023-07-16
Remove unnecessary export() call

ChrisThrasher:
## Description

This call appears to be redundant with the below lines which already generate this file.

https://github.com/SFML/SFML/blob/00b6488c3e9350bda573e9a82a06a1d855fa494b/cmake/Macros.cmake#L446-L448

The install interface tests should confirm that this is a valid change.

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -429,9 +429,6 @@ function(sfml_export_targets)
     endif()
     set(targets_config_filename "SFML${config_name}Targets.cmake")
 
-    export(EXPORT SFMLConfigExport
-           FILE "${CMAKE_CURRENT_BINARY_DIR}/${targets_config_filename}")
-
     if(SFML_BUILD_FRAMEWORKS)
         set(config_package_location "SFML.framework/Resources/CMake")
     else()
@@ -441,7 +438,6 @@ function(sfml_export_targets)
         INSTALL_DESTINATION "${config_package_location}")
     configure_package_config_file("${CURRENT_DIR}/SFMLConfigDependencies.cmake.in" "${CMAKE_CURRENT_BINARY_DIR}/SFMLConfigDependencies.cmake"
         INSTALL_DESTINATION "${config_package_location}")
-
 
     install(EXPORT SFMLConfigExport
             FILE ${targets_config_filename}


binary1248 (approved):

texus:
The `export` call isn't the same as `install(EXPORT)` though.

With the `export` call, files like `SFMLSharedTargets.cmake` are generated in the build folder, allowing SFML_DIR to point to the SFML build directory, without even running the equivalent of "make install" first. After this change SFML always has to be installed in order to use it.

While I haven't tested finding SFML in the build folder yet after this was merged, I did just quickly verify that on Linux there is no more SFMLSharedTargets.cmake file in the build directory while it is there with the previous commit.

While I could live with the fact that I must first run the install step instead of being lazy, the line was not "redundant".

ChrisThrasher:
> allowing SFML_DIR to point to the SFML build directory, without even running the equivalent of "make install" first

Is the idea that you want to call `find_package(SFML)` without having to actually install SFML? Can I ask why you wanted to do this rather than just install the library?

texus:
> Is the idea that you want to call find_package(SFML) without having to actually install SFML?

Yes.

> Can I ask why you wanted to do this rather than just install the library?

Laziness I guess :smile: 

On any platform you need elevated permissions to install to the default CMAKE_INSTALL_PREFIX, so if I need to install the files then I also have to first change this option (or just run with sudo). And since I type my cmake commands in a terminal instead of using the gui (unless I need to set more than just the basic settings), having to add CMAKE_INSTALL_PREFIX makes the command a lot longer (e.g. `cmake .. && make` vs `cmake -DCMAKE_INSTALL_PREFIX=/home/texus/Documents/SFML3/install .. && make install`).

Pointing `SFML_DIR` to the build directory and then calling `find_package(SFML)` has worked mostly fine for me, the only issue that I encountered was that it wouldn't find FreeType automatically on Windows and macOS and that I had to set `freetype_DIR` as well to make `find_package(SFML)` work on those platforms.

If I'm the only person who does this then I can easily adapt of my workflow of course, CMAKE_INSTALL_PREFIX is something that only needs to be specified once, but in my current workflow I almost never build the INSTALL target for SFML or my own library.

ChrisThrasher:
> And since I type my cmake commands in a terminal instead of using the gui (unless I need to set more than just the basic settings), having to add CMAKE_INSTALL_PREFIX makes the command a lot longer

You don't have to be so verbose about it. You can specify a relative path for the install prefix. I use this:

```
cmake -B build -DCMAKE_INSTALL_PREFIX=build/install
```

Then I point `SFML_ROOT` at that install/ dir and it all works. We do this in CI as well. CI is continuously testing the install interface of SFML.

kimci86:
Note this official CMake tutorial shows how to use `export` to make the project is usable from a build directory.
https://cmake.org/cmake/help/latest/guide/tutorial/Adding%20Export%20Configuration.html#cmakelists-txt-export

Also, at the beginning of the tutorial is:
> The next step is to add the necessary information so that other CMake projects can use our project, be it from a build directory, a local install or when packaged.

So I guess using the project from a build directory is a valid approach.

texus:
Removing the export call apparently affects libraries that depend on SFML as well.

In some cases the `export()` call in my cmake script is broken since this PR and now fails with the following error:
```export called with target "tgui" which requires target "sfml-graphics" that is not in any export set```

While I could work around this, I feel like this PR should just be reverted. It removed existing functionality without really providing any benefits.

ChrisThrasher:
Are you doing what we discussed before where you're pointing CMake at the build dir of SFML rather than its install dir or is this some new problem we haven't yet discussed?

texus:
This is a different but related issue.

For my own library I still have an `export()` call. However CMake can't generate the files in my build directory anymore because my library depends on SFML. So by deciding that SFML can't be used without installing it first, you are also deciding that libraries that depend on SFML should remove the possibility as well.

Right now it doesn't even matter whether someone tries to use the library from the build directory or first installing it, the very existence of the `export()` in my code is already causing a build error. The build failure can be triggered by just using the CMake SFML Project Template and then using FetchContent for my library after using FetchContent for SFML.

This leaves me with 3 potential fixes for my project:
- Remove `export()` from my cmake script as well, and no longer allowing people to use my project without installing it.
- Require CMake 3.26 to link to SFML with `$<BUILD_LOCAL_INTERFACE:...>`, so that the export ignores the dependency.
- Export SFML myself. I'm not sure if this is even possible, and it probably requires duplicating stuff from SFML's cmake script.

If I have to then I would probably just remove the `export()` call for my library. I however don't like the fact that I'm being forced to remove existing functionality, especially since this all works fine with SFML 2 (and as kimci86 said, the export is even mentioned as last step in the official "Adding Export Configuration" CMake tutorial)

PR #2627, opened 2023-07-27
Remove PR template testing checklist

ChrisThrasher:
## Description

These check boxes have been ignored for years and asking that users test on all those platforms isn't reasonable to begin with.


eXpl0it3r:
I've experienced the opposite. Among most of the template, these boxes have been "used" the most. 😄 
I don't think the expectation is that the PR creator tests on all the platforms, but it should a) give a hint that SFML is cross-platform and b) can be used as a check list by the maintainers as well, to see that we tested the change on all the platforms (if required).

If we remove it, we should replace it with something else that prompts point a) from above.

PR #2629, opened 2023-07-28
Run clang-tidy on Android code

ChrisThrasher:
## Description

This mostly involved lots of style fixes with few other changes required to make clang-tidy happy.

PR #2632, opened 2023-07-31
Add iOS clang-tidy job

ChrisThrasher:
## Description

An extension of #2629


ChrisThrasher:

src/SFML/Window/iOS/ClipboardImpl.mm
```diff
@@ -37,13 +37,13 @@
 ////////////////////////////////////////////////////////////
 String ClipboardImpl::getString()
 {
-    UIPasteboard* pboard = [UIPasteboard generalPasteboard];
+    UIPasteboard* const pboard = [UIPasteboard generalPasteboard];
     if (pboard.hasStrings)
     {
-        NSString* data = pboard.string;
+        const NSString* const data = pboard.string;
```
ChrisThrasher, line 43: I know we don't usually `const`-ify the pointer itself (usually just the pointed-to type) but clang-tidy enforces this in Objective C++ in spite of not enforcing it in C++ files. It's probably just a bug but it's a bug that we're already dealing with in all the macOS Objective C++ files so it's nothing new to SFML. Erring on the side of extra `const` is not bad though, just a big verbose.

eXpl0it3r (approved):

PR #2633, opened 2023-08-01
Test `SFML_USE_STATIC_STD_LIBS`

ChrisThrasher:
## Description

Yet another PR in my journey to testing all these configuration options we have. View the logs to see these flags being applied.

Something weird about this is that in the CI logs you only see this flag appear on the compilation of executables but not on all of SFML's .cpp files. Is that how this is supposed to work? I can't really tell. The other weird thing is that these `-static` flags appear _twice_ in the examples. I believe the problem is that we call `sfml_set_stdlib` when compiling the core library _and_ when compiling the examples. This function is not called when building the tests and the tests don't see this `-static` flag appear twice so that lines up.


eXpl0it3r:
> Something weird about this is that in the CI logs you only see this flag appear on the compilation of executables but not on all of SFML's .cpp files. Is that how this is supposed to work?

Since you build static libraries, you don't really "link" the static standard library, until you create a tangible artifact such as an executable or DLL, so that sounds correct to me.

eXpl0it3r (approved):

PR #2635, opened 2023-08-05
Add tests for more `loadFromFile` functions

ChrisThrasher:
## Description

The only class with an untested `loadFromFile` function is `sf::SoundBuffer` but that class has no tests yet because testing OpenAL code always fails in CI. I've already already started writing [tests for Audio module types](https://github.com/SFML/SFML/compare/master...ChrisThrasher:SFML:test_audio)

### DIFF
--- a/test/Graphics/Shader.test.cpp
+++ b/test/Graphics/Shader.test.cpp
@@ -142,7 +142,7 @@ TEST_CASE("[Graphics] sf::Shader (Dummy Implementation)", skipShaderDummyTests()
         CHECK_FALSE(sf::Shader::isGeometryAvailable());
     }
 
-    SECTION("Load")
+    SECTION("loadFromMemory()")
     {
         sf::Shader shader;
         CHECK_FALSE(shader.loadFromMemory(vertexSource, sf::Shader::Type::Vertex));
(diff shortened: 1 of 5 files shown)


ChrisThrasher:
Screenshot 2023-08-04 at 8 49 34 PM

Not sure what to make of these Debian jobs failing.

binary1248 (approved):

test/Graphics/Shader.test.cpp
```diff
@@ -163,17 +163,51 @@ TEST_CASE("[Graphics] sf::Shader", skipShaderFullTests())
...
+        CHECK(shader.getNativeHandle() == 0);
+    }
+
+    SECTION("loadFromFile()")
+    {
+        SECTION("One shader")
+        {
+            sf::Shader shader;
+            CHECK(!shader.loadFromFile("does-not-exist.vert", sf::Shader::Type::Vertex));
+            CHECK(shader.loadFromFile("Graphics/shader.vert", sf::Shader::Type::Vertex) == sf::Shader::isAvailable());
```
kimci86, line 178: This is testing only `Type::Vertex`. What about `Type::Fragment` and `Type::Geometry`?
If `Type::Geometry` crashes on the debian worker, maybe there is a bug to fix?

ChrisThrasher: I added a test for loading a fragment shader but couldn't get a geometry shader to load. The Linux and Windows jobs are showing that loading the geometry shader is failing in spite of geometry shaders being available.

kimci86: I think this is expected. The preexisting test about loadFromMemory expects loading a geometry shader alone fails.

ChrisThrasher: I see what you mean. It's weird that geometry shaders always fail to load though, no? Why have this API if it always fails?

PR #2636, opened 2023-08-05
Show debug path information when loading a shader file fails

kimci86:
Noticed while testing #2635.
This makes the error log more helpful when loading a shader file fails because of wrong filename or working directory.


ChrisThrasher:

ChrisThrasher (approved):
This is such a good idea! This must have just been an oversight when I first added `formatDebugPathInfo`.

Issue #2507, opened 2023-04-09
Remove default constructor for types that require a view into a resource


ChrisThrasher:
## Subject of the issue

#2486 and #2494 remove default constructors for `sf::Text` and `sf::Sprite` which are types that are not usable unless they contain a pointer to a valid resource. Those resources are an `sf::Font` and a `sf::Texture` respectively. The rationale is that by allowing for default construction, we're making it too easy for users to write bugs where they forget to assign the correct resource to a given type. By removing the default constructor, we force users to provide that resource upon construction so we can be sure that instances of these types are valid at all points in time. The convenience of a default constructor is not worth the potentially incorrect code someone may write. It's not possible to set these pointers to `nullptr` after construction so why let them be null upon construction?

Here's a list of types that may need to get this same treatment before SFML 3.0.0.

- [x] `sf::Text` (#2486)
- [x] `sf::Sprite` (#2494)
- [x] `sf::Sound` (#2640)

I believe there are no more than just these three but I could have missed one. `sf::Sound` is harder because its copy assignment operator will set the underlying `sf::SoundBuffer` to `nullptr` so it actually have its resource set to `nullptr` after construction unlike the other two. A solution for `sf::Sound` may require altering how copy semantics work.

ChrisThrasher:
@kimci86 pointed out how the copy semantics of `sf::Sound` don't actually leave the instance in a state with a null `SoundBuffer`. However I did discover [`sf::Sound::resetBuffer`](https://github.com/SFML/SFML/blob/ea4c448a85fc1a93c71431faf4b4f12efcfe0976/src/SFML/Audio/Sound.cpp#L195) which sets the underlying sound buffer to null. If the API of `sf::Sound` explicitly allows for nullifying its underlying resource then it's valid for `sf::Sound::Sound()` to default to a null sound buffer.

ChrisThrasher:
https://github.com/SFML/SFML/commit/362f37473

This machinery was added in 2010 and very little has changed about `sf::SoundBuffer` since then so I think it's reasonable to revisit this decision.

PR #2637, opened 2023-08-05
Return reference instead of non-null pointer

kimci86:
Since #2486 and #2494, `Sprite` and `Text` instances cannot hold a null pointer to the corresponding resource.
I suggest that `getTexture` and `getFont` return a reference instead of a pointer to make it clear that it is not null.
What do you think?

Note this is slightly changing `Sprite` constructor behavior. If I understand correctly, `Sprite(texture, IntRect())` is equivalent to `Sprite(texture)` in master, whereas `Sprite(texture, IntRect())` would actually create a Sprite with an empty texture rectangle with this PR. It looks to me this would be the correct behavior one would expect.

I also reviewed corresponding tests and fixed a small inconsistency in `Text` tests while at it.

The same kind of change would apply to `Sound` and `getSoundBuffer` once the default constructor gets removed as planned in #2507.

### DIFF
--- a/include/SFML/Graphics/Sprite.hpp
+++ b/include/SFML/Graphics/Sprite.hpp
@@ -139,16 +139,15 @@ public:
     ////////////////////////////////////////////////////////////
     /// \brief Get the source texture of the sprite
     ///
-    /// If the sprite has no source texture, a null pointer is returned.
-    /// The returned pointer is const, which means that you can't
+    /// The returned reference is const, which means that you can't
     /// modify the texture when you retrieve it with this function.
     ///
-    /// \return Pointer to the sprite's texture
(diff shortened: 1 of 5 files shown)


ChrisThrasher:

ChrisThrasher (approved):
I really like this! It removes yet another potential source of bugs from the API and further reinforces the idea that SFML 3 is systematically eliminating invalid states from being representable in the type system.

binary1248 (approved):

ChrisThrasher:
https://github.com/SFML/imgui-sfml/pull/250/files

Already found a real world example this is helpful. It's useful to know that a sprite always has a texture. It saves you some manual error checking.

PR #2640, opened 2023-08-06
Remove sound default constructor

kimci86:
## Description

This PR is related to the issue #2507.

By removing `Sound` default constructor, we make sure it is always bound to some `SoundBuffer`.

Doing some operations on an OpenAL sound buffer while OpenAL Sources are attached to it triggers error messages, so we have to keep track of `Sound` instances which are bound to a given `SoundBuffer` so that we can detach and attach sounds when needed.

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

- Running SFML audio examples `voip`, `sound`, `sound_capture` and `tennis`.
- Playing around with the following program:

```c++
#include <SFML/Graphics.hpp>
#include <SFML/Audio.hpp>

int main()
{
    auto window = sf::RenderWindow{sf::VideoMode{{200, 200}}, "Test"};
    window.setFramerateLimit(60);

    auto buffer1 = sf::SoundBuffer{};
    if (!buffer1.loadFromFile("resources/jingles_SAX02.ogg"))
        return EXIT_FAILURE;

    auto buffer2 = sf::SoundBuffer{};
    if (!buffer2.loadFromFile("resources/jingles_SAX07.ogg"))
        return EXIT_FAILURE;

    auto buffer = buffer1;

    auto sound1 = sf::Sound{buffer};
    auto sound2 = sound1;
    sound2.setPitch(0.5f);

    while (window.isOpen())
    {
        {
            switch (event.type)
            {
                case sf::Event::Closed:
                    window.close();
                    break;
                case sf::Event::KeyPressed:
                    switch (event.key.scancode)
                    {
                        case sf::Keyboard::Scan::Num1:
                            sound1.play();
                            break;
                        case sf::Keyboard::Scan::Num2:
                            sound2.play();
                            break;
                        case sf::Keyboard::Scan::Q:
                            buffer = buffer1;
                            break;
                        case sf::Keyboard::Scan::W:
                            buffer = buffer2;
                            break;
                        case sf::Keyboard::Scan::E:
                        {
                            auto buffer3 = sf::SoundBuffer{};
                            if (!buffer3.loadFromFile("resources/jingles_SAX03.ogg"))
                                return EXIT_FAILURE;
                            buffer = buffer3;
                            break;
                        }
                        case sf::Keyboard::Scan::A:
                            sound1.setBuffer(buffer1);
                            sound2 = sound1;
                            sound2.setPitch(0.5f);
                            break;
                        case sf::Keyboard::Scan::S:
                            sound1.setBuffer(buffer);
                            sound2 = sound1;
                            sound2.setPitch(0.5f);
                            break;
                        default:
                            break;
                    }
                    break;
                default:
                    break;
            }
        }

        window.clear();
        window.display();
    }
}
```

[resources.zip](https://github.com/SFML/SFML/files/12270235/resources.zip)
Sounds are from [kenney.nl](https://kenney.nl/assets/music-jingles)

### DIFF
--- a/include/SFML/Audio/Sound.hpp
+++ b/include/SFML/Audio/Sound.hpp
@@ -47,12 +47,6 @@ class SFML_AUDIO_API Sound : public SoundSource
 {
 public:
     ////////////////////////////////////////////////////////////
-    /// \brief Default constructor
-    ///
-    ////////////////////////////////////////////////////////////
-    Sound();
-
-    ////////////////////////////////////////////////////////////
     /// \brief Construct the sound with a buffer
     ///
     /// \param buffer Sound buffer containing the audio data to play with the sound
@@ -213,18 +207,27 @@ public:
     ////////////////////////////////////////////////////////////
     Sound& operator=(const Sound& right);
 
-    ////////////////////////////////////////////////////////////
-    /// \brief Reset the internal buffer of the sound
-    ///
-    /// This function is for internal use only, you don't have
-    /// to use it. It is called by the sf::SoundBuffer that
-    /// this sound uses, when it is destroyed in order to prevent
-    /// the sound from using a dead buffer.
-    ///
-    ////////////////////////////////////////////////////////////
-    void resetBuffer();
-
 private:
+    friend class SoundBuffer;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Detach sound from its internal buffer
+    ///
+    /// This allows the sound buffer to temporarily detach the
+    /// sounds that use it when the sound buffer gets updated.
+    ///
+    ////////////////////////////////////////////////////////////
+    void detachBuffer();
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Re-attach sound to its internal buffer
+    ///
+    /// This allows the sound buffer to attach back the sounds
+    /// that use it after the sound buffer has been updated.
+    ///
+    ////////////////////////////////////////////////////////////
+    void reattachBuffer();
+
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
@@ -252,7 +255,7 @@ private:
 ///
 /// In order to work, a sound must be given a buffer of audio
 /// data to play. Audio data (samples) is stored in sf::SoundBuffer,
-/// and attached to a sound with the setBuffer() function.
+/// and attached to a sound when it is created or with the setBuffer() function.
 /// The buffer object attached to a sound must remain alive
 /// as long as the sound uses it. Note that multiple sounds
 /// can use the same sound buffer at the same time.
@@ -265,8 +268,7 @@ private:
 ///     // Handle error...
 /// }
 ///
-/// sf::Sound sound;
-/// sound.setBuffer(buffer);
+/// sf::Sound sound(buffer);
 /// sound.play();
 /// \endcode
 ///
--- a/include/SFML/Audio/Sound.hpp
+++ b/include/SFML/Audio/Sound.hpp
@@ -164,10 +164,10 @@ public:
     ////////////////////////////////////////////////////////////
     /// \brief Get the audio buffer attached to the sound
     ///
-    /// \return Sound buffer attached to the sound (can be a null pointer)
+    /// \return Sound buffer attached to the sound
     ///
     ////////////////////////////////////////////////////////////
-    const SoundBuffer* getBuffer() const;
+    const SoundBuffer& getBuffer() const;
 
     ////////////////////////////////////////////////////////////
     /// \brief Tell whether or not the sound is in loop mode
--- a/src/SFML/Audio/Sound.cpp
+++ b/src/SFML/Audio/Sound.cpp
@@ -36,21 +36,18 @@
 namespace sf
 {
 ////////////////////////////////////////////////////////////
-Sound::Sound() = default;
-
-
-////////////////////////////////////////////////////////////
-Sound::Sound(const SoundBuffer& buffer)
+Sound::Sound(const SoundBuffer& buffer) : m_buffer(&buffer)
 {
-    setBuffer(buffer);
+    m_buffer->attachSound(this);
+    alCheck(alSourcei(m_source, AL_BUFFER, static_cast<ALint>(m_buffer->m_buffer)));
 }
 
 
 ////////////////////////////////////////////////////////////
-Sound::Sound(const Sound& copy) : SoundSource(copy)
+Sound::Sound(const Sound& copy) : SoundSource(copy), m_buffer(copy.m_buffer)
 {
-    if (copy.m_buffer)
-        setBuffer(*copy.m_buffer);
+    m_buffer->attachSound(this);
+    alCheck(alSourcei(m_source, AL_BUFFER, static_cast<ALint>(m_buffer->m_buffer)));
     setLoop(copy.getLoop());
 }
 
@@ -59,8 +56,7 @@ Sound::Sound(const Sound& copy) : SoundSource(copy)
 Sound::~Sound()
 {
     stop();
-    if (m_buffer)
-        m_buffer->detachSound(this);
+    m_buffer->detachSound(this);
 }
 
 
@@ -89,11 +85,8 @@ void Sound::stop()
 void Sound::setBuffer(const SoundBuffer& buffer)
 {
     // First detach from the previous buffer
-    if (m_buffer)
-    {
-        stop();
-        m_buffer->detachSound(this);
-    }
+    stop();
+    m_buffer->detachSound(this);
 
     // Assign and use the new buffer
     m_buffer = &buffer;
@@ -164,17 +157,8 @@ Sound& Sound::operator=(const Sound& right)
     // Delegate to base class, which copies all the sound attributes
     SoundSource::operator=(right);
 
-    // Detach the sound instance from the previous buffer (if any)
-    if (m_buffer)
-    {
-        stop();
-        m_buffer->detachSound(this);
-        m_buffer = nullptr;
-    }
-
     // Copy the remaining sound attributes
-    if (right.m_buffer)
-        setBuffer(*right.m_buffer);
+    setBuffer(*right.m_buffer);
     setLoop(right.getLoop());
 
     return *this;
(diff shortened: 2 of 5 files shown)


ChrisThrasher:

kimci86:
I added a test program. I found no issue playing around with it.

ChrisThrasher:
```cmake
project(audio-test CXX)

include(FetchContent)
FetchContent_Declare(SFML
    GIT_TAG feature/remove_sound_default_constructor)
FetchContent_MakeAvailable(SFML)

add_executable(audio-test main.cpp)
target_link_libraries(audio-test PRIVATE SFML::Graphics SFML::Audio)
```
Here's how you can build this branch to run the example program.

I ran this and couldn't notice any oddities so this change appears to hold up to testing.

ChrisThrasher (approved):

include/SFML/Audio/Sound.hpp
```diff
@@ -213,18 +207,27 @@ class SFML_AUDIO_API Sound : public SoundSource
     ////////////////////////////////////////////////////////////
     Sound& operator=(const Sound& right);
 
+private:
+    friend class SoundBuffer;
```
eXpl0it3r, line 211: This is needed because..?

src/SFML/Audio/Sound.cpp
```diff
@@ -117,9 +110,9 @@ void Sound::setPlayingOffset(Time timeOffset)
 
 
 ////////////////////////////////////////////////////////////
-const SoundBuffer* Sound::getBuffer() const
+const SoundBuffer& Sound::getBuffer() const
 {
-    return m_buffer;
+    return *m_buffer;
```
eXpl0it3r, line 115: If ever sound is required to have a buffer, why do we use a pointer still to reference the buffer?

src/SFML/Audio/SoundBuffer.cpp
```diff
@@ -213,7 +207,13 @@ SoundBuffer& SoundBuffer::operator=(const SoundBuffer& right)
...
     std::swap(m_buffer, temp.m_buffer);
     std::swap(m_duration, temp.m_duration);
-    std::swap(m_sounds, temp.m_sounds); // swap sounds too, so that they are detached when temp is destroyed
+
+    // Reattach sounds that use this buffer so they get bound to the new OpenAL buffer
+    for (Sound* soundPtr : m_sounds)
+    {
+        soundPtr->stop();
+        soundPtr->reattachBuffer();
+    }
```
eXpl0it3r, line 216: What's the point of reattaching if we previously assert when the m_sounds isn't empty?

eXpl0it3r (changes requested):

src/SFML/Audio/Sound.cpp
ChrisThrasher: We can't use a reference because in `sf::Sound::setBuffer` we reassign the buffer. We could consider using a std::reference_wrapper. It has the semantics we want. It can't be null yet does support reassignment. It's just a fancy wrapper around a raw pointer.

include/SFML/Audio/Sound.hpp
kimci86: When the SoundBuffer updates its OpenAL buffer, we need first to dettach sounds currently bound to it to avoid OpenAL errors. Previously we had `Sound::resetBuffer` method for that same reason. That method was public but documented as "for internal use only". I replaced it with methods `Sound::detachBuffer` and `Sound::reattachBuffer` and figured it would be better to make them private and having friend class SoundBuffer because the methods are only useful to SoundBuffer.

src/SFML/Audio/Sound.cpp
ChrisThrasher: https://github.com/ChrisThrasher/SFML/commit/fc5f0c1083561d431aee2a0a6dfcb19b57610cef

kimci86: Having a reference instead would make the Sound type non-copyable because references cannot be reassigned.

src/SFML/Audio/SoundBuffer.cpp
kimci86: At this point in the code, sounds are still correctly related to the SoundBuffer, but underlying OpenAL sources are still incorrectly bound to the previous OpenAL buffer. Calling reattachBuffer here makes the sounds bind to the new OpenAL buffer whose id was put in the SoundBuffer's m_buffer member.

kimci86: Sorry, I now realize I didn't quite answer the question. So, we assert when m_sounds is empty in SoundBuffer destructor. Here `temp` is about to get desctucted but it has not sounds attached. I don't see any problem about iterating on m_sounds here. m_sounds content is not touched by the assignment operator.

eXpl0it3r: I'm sorry, I somehow didn't have all the code expanded on GitHub and made some wrong connections (read the beginning of `~SoundBuffer()` and thought this code was at the end of the function...) 🤦

src/SFML/Audio/Sound.cpp
eXpl0it3r: I mean, if we can guarantee that the pointer can never be invalidated or `nullptr` then we can keep it. The issue I see with the change is, that we remove any null-checks, yet the keep the pointer. So if we have an invalid pointer, you could get some ugly crashes.

eXpl0it3r:
I just inserted this code to the testing example from above: 

```cpp
                case sf::Keyboard::Scan::Num3:
                {
                    auto bufferTemp = sf::SoundBuffer{};
                    if (!bufferTemp.loadFromFile("resources/jingles_SAX03.ogg"))
                        return EXIT_FAILURE;
                    sound2.setBuffer(bufferTemp);
                    sound2.play();
                    break;
                }
```

In debug mode this fails in the added `assert()`, but in release mode it plays just fine. 🤔 
I wonder whether there shouldn't be harder fail also in release mode, otherwise people might just not care, or we get a lot of "but it works in release".

ChrisThrasher:
> I wonder whether there shouldn't be harder fail also in release mode, otherwise people might just not care, or we get a lot of "but it works in release".

This is getting at how `sf::Sound` and `sf::SoundBuffer` properly ensure correct lifetimes are maintained whereas something like `sf::Text` and `sf::Font` just let those lifetime issues persist undetected. This is an inconsistency worth digging into at some point in the future before SFML 3's release.

The only way to turn this into a hard failure in a destructor in release mode is to throw an exception or call some variety of program-ending function like `std::terminate()`. I think a good improvement would be to elevate that assertion into an `sf::err` message so that those using release builds at least get some indication that something weird is going on. After the log we can `assert(false)`. We already use this pattern elsewhere.

src/SFML/Audio/SoundBuffer.cpp
```diff
@@ -68,15 +70,7 @@ SoundBuffer::SoundBuffer(const SoundBuffer& copy) : m_samples(copy.m_samples), m
...
-    // To prevent the iterator from becoming invalid, move the entire buffer to another
-    // container. Otherwise calling resetBuffer would result in detachSound being
-    // called which removes the sound from the internal list.
-    SoundList sounds;
-    sounds.swap(m_sounds);
-
-    // Detach the buffer from the sounds that use it (to avoid OpenAL errors)
-    for (Sound* soundPtr : sounds)
-        soundPtr->resetBuffer();
+    assert(m_sounds.empty() && "sf::SoundBuffer must not be destructed while it is used by a sf::Sound");
```
ChrisThrasher, line 73: How about we make this an `sf::err` log so that even those using release builds are notified that their program is violating lifetime rules?

kimci86:
> In debug mode this fails in the added assert(), but in release mode it plays just fine.

I guess in that case, OpenAL detects that the OpenAL buffer is still in use and refuses to delete it, so we are leaking the resource then.
I will implement a stronger fail, thank you for the feedback.

kimci86:
Here we go. Now debug or release are the same.
I choose to call `std::terminate` instead of `assert` to avoid undefined behavior in release mode.
FYI, throwing an exception from a desctructor also calls `std::terminate`.

src/SFML/Audio/SoundBuffer.cpp
```diff
@@ -68,15 +69,14 @@ SoundBuffer::SoundBuffer(const SoundBuffer& copy) : m_samples(copy.m_samples), m
...
-
-    // Detach the buffer from the sounds that use it (to avoid OpenAL errors)
-    for (Sound* soundPtr : sounds)
-        soundPtr->resetBuffer();
+    // Make sure no sound is attached to this buffer
+    if (!m_sounds.empty())
+    {
+        const char* plural = 2 <= m_sounds.size() ? "s" : "";
+        err() << "Failed to destruct sound buffer because it is used by " << m_sounds.size() << " sound" << plural
+              << std::endl;
```
ChrisThrasher, line 77:
```suggestion
        err() << "Failed to destruct sound buffer because it is used by " << m_sounds.size() << " sound(s)" << plural
              << std::endl;
```
There's already a simpler way of expressing potential plurality.

kimci86: Good idea.

ChrisThrasher:

src/SFML/Audio/Sound.cpp
```diff
@@ -36,21 +36,18 @@
...
-Sound::Sound() = default;
-
-
-////////////////////////////////////////////////////////////
-Sound::Sound(const SoundBuffer& buffer)
+Sound::Sound(const SoundBuffer& buffer) : m_buffer(&buffer)
 {
-    setBuffer(buffer);
+    m_buffer->attachSound(this);
+    alCheck(alSourcei(m_source, AL_BUFFER, static_cast<ALint>(m_buffer->m_buffer)));
```
eXpl0it3r, line 42: Not really a deal breaker, but we've duplicated these two lines in three places now.

eXpl0it3r (approved):

src/SFML/Audio/Sound.cpp
kimci86: I agree this duplication it is not extremely elegant. The motivation here is that only constructors are responsible for setting up the invariant (i.e. sound it attached to a sound buffer), and methods (especially setBuffer) can be simpler by assuming the invariant is true.

Issue #2594, opened 2023-06-28
Compiler errors with gcc 5


jcar87:
## Subject of the issue

Cannot build sfml 2.6.0 from source with gcc 5.

## Your environment

sfml 2.6.0, Linux with gcc 5

## Steps to reproduce

Build with the cmake option `WARNINGS_AS_ERRORS` set to `OFF`

## Expected behavior

A successful build (perhaps with warnings)

## Actual behavior

```
g++: error: unrecognized command line option ‘-Wnull-dereference’
g++: error: unrecognized command line option ‘-Wmisleading-indentation’
g++: error: unrecognized command line option ‘-Wduplicated-cond’
make[2]: *** [src/SFML/System/CMakeFiles/sfml-system.dir/Clock.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
g++: error: unrecognized command line option ‘-Wnull-dereference’
g++: error: unrecognized command line option ‘-Wmisleading-indentation’
g++: error: unrecognized command line option ‘-Wduplicated-cond’
g++: error: unrecognized command line option ‘-Wnull-dereference’
g++: error: unrecognized command line option ‘-Wmisleading-indentation’
g++: error: unrecognized command line option ‘-Wduplicated-cond’
make[2]: *** [src/SFML/System/CMakeFiles/sfml-system.dir/Err.cpp.o] Error 1
make[2]: *** [src/SFML/System/CMakeFiles/sfml-system.dir/Lock.cpp.o] Error 1
make[1]: *** [src/SFML/System/CMakeFiles/sfml-system.dir/all] Error 2
make: *** [all] Error 2
```

I can't find a mention in the documentation about gcc5 not being supported, clarification would be appreciated, thanks!

ChrisThrasher:
We never made an intentional decision to stop supporting GCC 5. It's simply a matter of nobody using GCC 5 and reporting this issue until now. In theory GCC 5 works if you stop applying those warnings.

Is there a GCC 5 flag that lets you ignore unknown warnings?

jcar87:
I see, thanks @ChrisThrasher ! Indeed removing the warnings actually works successfully - but it's a bit unergonomic as the only way is to patch the CMake scripts rather than via a configuration.
I can't find a universal option gcc that would ignore _all_ unknown warnings, if my understanding of the gcc documentation is correct I think you still need to know the _name_ of the warning to disable it.

[...]

PR #2642, opened 2023-08-10
Use CMake Presets to Disambiguate Developer Builds from User Builds

ChrisThrasher:
## Description

https://cmake.org/cmake/help/v3.22/manual/cmake-presets.7.html

After our [recent upgrade to CMake 3.22](https://github.com/SFML/SFML/pull/2543) we have access to CMake presets, a feature added in CMake 3.19 that is meant to simply the process of adding arguments during configuration. Something important to note about presets is that they're intended to be a developer feature. Users are not expected to use a given project's presets. In CI we can see what this looks like. The diff shows how we can take our most common configuration arguments and encode them in a new CMakePresets.json file.

So why bother with this? Aren't we just shuffling code around for no benefit?

The major benefit here is that presets give us a way of **separating how users build code from how developers build code**. The two parties have very different needs. Users want just the core library. They may specify their own build type or library type but that's about it. Developers meanwhile want the core library and examples and tests and compiler warnings and docs and potentially even more.

As it stands now we're in an awkward place where we want to enable some things by default for the sake of developer convenience but that may get in the way of the user experience. We want devs to have tests and examples enabled for example but that's just unnecessary bloat for users until they figure out how to disable it.

A recent real world problem is the fact that we enable warnings as error by default. [This recently caught out the folks packaging SFML into Vcpkg](https://github.com/microsoft/vcpkg/issues/32574). Conan developers had a [similar issue](https://github.com/SFML/SFML/issues/2594) related to how we enable many compiler warnings by default, something we do for the sake of developer convenience. It would be nice if we could disable warnings-as-errors by default to make users happy while keeping it enabled by default for developers and presets are how we can do that.

This PR takes a neutral stance by maintaining the status quo when it comes to configuration arguments, but if we merge this we'll have the infrastructure to easily change out mind about how developers build SFML versus our how our users do.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,7 +126,7 @@ jobs:
 
     - name: Configure CMake
       shell: bash
-      run: cmake -S $GITHUB_WORKSPACE -B $GITHUB_WORKSPACE/build -DCMAKE_INSTALL_PREFIX=$GITHUB_WORKSPACE/install -DSFML_BUILD_EXAMPLES=TRUE -DCMAKE_VERBOSE_MAKEFILE=ON -DSFML_BUILD_TEST_SUITE=TRUE ${{matrix.platform.flags}} ${{matrix.config.flags}} ${{matrix.type.flags}}
+      run: cmake --preset dev -DCMAKE_VERBOSE_MAKEFILE=ON ${{matrix.platform.flags}} ${{matrix.config.flags}} ${{matrix.type.flags}}
 
     - name: Build
       shell: bash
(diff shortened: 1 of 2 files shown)


ChrisThrasher:

ChrisThrasher:

.github/workflows/ci.yml
```diff
@@ -126,7 +126,7 @@ jobs:
 
     - name: Configure CMake
       shell: bash
-      run: cmake -S $GITHUB_WORKSPACE -B $GITHUB_WORKSPACE/build -DCMAKE_INSTALL_PREFIX=$GITHUB_WORKSPACE/install -DSFML_BUILD_EXAMPLES=TRUE -DCMAKE_VERBOSE_MAKEFILE=ON -DSFML_BUILD_TEST_SUITE=TRUE ${{matrix.platform.flags}} ${{matrix.config.flags}} ${{matrix.type.flags}}
+      run: cmake --preset dev -DCMAKE_VERBOSE_MAKEFILE=TRUE ${{matrix.platform.flags}} ${{matrix.config.flags}} ${{matrix.type.flags}}
```
eXpl0it3r, line 129: Did you intentionally change from `-DCMAKE_VERBOSE_MAKEFILE=ON` to `-DCMAKE_VERBOSE_MAKEFILE=TRUE`?

CMakePresets.json
```diff
@@ -0,0 +1,16 @@
+{
+  "version": 3,
+  "configurePresets":[
+    {
+      "name": "dev",
+      "binaryDir": "build",
+      "installDir": "${sourceDir}/build/install",
```
eXpl0it3r, line 7: Isn't installing inside the build directory rather bad practice?

.github/workflows/ci.yml
ChrisThrasher: Probably just an incidental thing. I'll use whichever style you prefer.

CMakePresets.json
ChrisThrasher: I've never heard anyone discourage it or argue against doing so for developer builds.

It's great for development because it means we can install SFML without needing root access and you don't run the risk of polluting your machine with potentially bugged installs of SFML. Plus it means you don't need to add yet another directory to your .gitignore. All build products remain in the binary directory which is already ignored. When you want to delete your build dir you also delete this install dir so it won't stick around polluting anything.

.github/workflows/ci.yml
eXpl0it3r: I think I'd just go with ON/OFF

ChrisThrasher: Okay I changed it back to using `ON`

CMakePresets.json
eXpl0it3r: Feels a bit weird, but I guess we can keep it for now

ChrisThrasher:

eXpl0it3r (approved):

PR #2644, opened 2023-08-11
Remove `$GITHUB_WORKSPACE` where not needed

ChrisThrasher:
## Description

Our CI shell commands tend to be pretty long and shortening them has readability benefits.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,7 +132,7 @@ jobs:
 
     - name: Build
       shell: bash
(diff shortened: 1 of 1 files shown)


ChrisThrasher:

binary1248 (approved):

Bromeon (approved):
Looks good!

A small thing to keep in mind is that the new code would no longer work when `working-directory:` is used, but at that point that's probably intended.

eXpl0it3r (approved):

PR #2646, opened 2023-08-14
Default to building static libraries

### LINKED ISSUE #2389: Stop changing the default value of `BUILD_SHARED_LIBS`
[Here](https://github.com/SFML/SFML/blob/master/CMakeLists.txt#L47) we default to building shared libraries for most platforms. Not only is this unnecessary to specify, it actively causes lots of pain for Windows users who routinely struggle to understand how DLLs work. Such users would be better off if we stuck to the CMake default of static libraries so they didn't have to try so hard to get their programs building. We can achieve this by simply removing any reference to `BUILD_SHARED_LIBS` and letting it return to its default false-y state.

It's been discussed on the forums [here](https://en.sfml-dev.org/forums/index.php?topic=27996.msg175824#msg175824).

Are there any benefits to defaulting to shared libraries that I'm missing?

ChrisThrasher:
## Description

Closes #2389

I've looked around at a few CI jobs and they seem to be behaving the same as before. If any jobs are broken it would be the jobs that aren't part of the core build matrix.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,10 +34,8 @@ include(cmake/Config.cmake)
 # https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html
 include(GNUInstallDirs)
 
-# add an option for choosing the build type (shared or static)
-if(NOT SFML_OS_ANDROID)
-    sfml_set_option(BUILD_SHARED_LIBS TRUE BOOL "TRUE to build SFML as shared libraries, FALSE to build it as static libraries")
-else()
+# disable static libraries on Android
+if(SFML_OS_ANDROID)
     set(BUILD_SHARED_LIBS TRUE)
 endif()
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
           config: { name: OpenGL ES, flags: -DSFML_USE_MESA3D=TRUE -DBUILD_SHARED_LIBS=TRUE -DSFML_OPENGL_ES=ON }
           type: { name: Debug, flags: -DCMAKE_BUILD_TYPE=Debug -DSFML_ENABLE_COVERAGE=TRUE }
         - platform: { name: Windows MinGW, os: windows-2022 }
-          config: { name: Static Standard Libraries, flags: -GNinja -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DBUILD_SHARED_LIBS=FALSE -DSFML_USE_STATIC_STD_LIBS=TRUE }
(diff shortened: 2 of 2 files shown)


ChrisThrasher:

CMakeLists.txt
```diff
@@ -34,10 +34,8 @@ include(cmake/Config.cmake)
...
 include(GNUInstallDirs)
 
-# add an option for choosing the build type (shared or static)
-if(NOT SFML_OS_ANDROID)
-    sfml_set_option(BUILD_SHARED_LIBS TRUE BOOL "TRUE to build SFML as shared libraries, FALSE to build it as static libraries")
-else()
+# disable static libraries on Android
+if(SFML_OS_ANDROID)
     set(BUILD_SHARED_LIBS TRUE)
 endif()
```
ChrisThrasher, line 40: https://github.com/SFML/SFML/commit/60894d1c1a5e51c6a02f85edbedf8ec91db1006b

I dug back in the commit history to find where this was added. I'm not totally sure why we have this here. Anything related to cross compilation belongs in a toolchain file. For what it's worth the Android CI jobs still succeed even if we remove this and build them as static libraries. Hopefully someone can point me to some Android docs explaining this requirement so I can add that link to the comment.

ChrisThrasher: @intjelic Do you know if we still need this?

CMakeLists.txt
```diff
@@ -34,13 +34,6 @@ include(cmake/Config.cmake)
 # https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html
 include(GNUInstallDirs)
 
-# add an option for choosing the build type (shared or static)
-if(NOT SFML_OS_ANDROID)
-    sfml_set_option(BUILD_SHARED_LIBS TRUE BOOL "TRUE to build SFML as shared libraries, FALSE to build it as static libraries")
-else()
-    set(BUILD_SHARED_LIBS TRUE)
```
ChrisThrasher, line 41: I decided to remove this setting for Android. It seems heavy-handed to forbid Android users from using static libraries.

ChrisThrasher:

CMakeLists.txt
kimci86: Android programs are java programs. Native code always come from a dynamic library.

ChrisThrasher: I don't think we _need_ to do any of that. Those on systems with particular requires like this can easily modify `BUILD_SHARED_LIBS` or `CMAKE_POSITION_INDEPENDENT_CODE` as needed. I will acknowledge however that removing this is a bit outside the scope of this PR so I'll roll it back if it risks stalling the PR.

kimci86: Right, configuring the android build might belong to android specific build scripts (maybe build.gradle or Android.mk, not sure). I suggest we test the android example before breaking stuff.

ChrisThrasher: Fair. I don't personally want to test those Android build scripts nor wait on someone else to do that so I'll remove that change for the sake of this PR being more mergeable.

ChrisThrasher:

eXpl0it3r (approved):

PR #2647, opened 2023-08-14
Add CMakeUserPresets to .gitignore

kimci86:
Now that we use a CMakePresets.json file for CI, it makes sense to support users having their own presets in a CMakeUserPresets.json file.

From CMake documentation:

> CMakePresets.json may be checked into a version control system, and
> CMakeUserPresets.json should NOT be checked in. For example, if a
> project is using Git, CMakePresets.json may be tracked, and
> CMakeUserPresets.json should be added to the .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,6 @@ cmake-build-*/
 .vscode/
 .idea/
 .vs/
+
+# User configuration
+CMakeUserPresets.json


ChrisThrasher (approved):

ChrisThrasher:

ChrisThrasher:
Unrelated to this PR but our .gitignore contains tons of useless boilerplate. It doesn't need to be any bigger than simply `build` and `CMakeUserPresets.json`. I'll even compromise and let all the IDE-specific build dirs stay in there ([although everyone ought to take care of that themselves via a global .gitignore file](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/)). But we can absolutely get rid of all that nonsense in the first 32 lines of .gitignore.

```
build/
cmake-build-*/
.cache/
.vscode/
.idea/
.vs/
CMakeUserPresets.json
```

This is all we need.

Issue #2459, opened 2023-03-21
Use `IMPORTED` libraries for vendored dependencies instead of `INTERFACE` libraries


ChrisThrasher:
## Subject of the issue

This will make `sfml_find_package` more similar to have `find_package` works which conventionally will give you an `IMPORTED` library. `IMPORTED` libraries are libraries which already exist on your system somewhere and you're merely creating a target for them which is exactly what we're doing. We're not trying to create any libraries that don't already exist. We just want a target for an existing pair of link libraries and include directories.

I've attempted this before and it gets messy because `sfml_find_package` also has some special behavior for iOS. This is one step towards being able to get rid of `sfml_find_package` in lieu of merely using `find_package`.

ChrisThrasher:
https://github.com/ChrisThrasher/SFML/commit/d16af6a8c970030f3d602dbbdb3e77f14e13d49e

#2359 makes some simplifications that let me inline all of `sfml_add_external` which takes us one step towards a less janky dependency management story and being able to use `IMPORTED` targets.

JonnyPtn:
This somewhat relates to the problem with the openAL dll not being added to `TARGET_RUNTIME_DLLS` - if we use an imported library and set the imported properties correctly then it will be added to that list

ChrisThrasher:
See #2650 for some meaningful progress on this. I replaced 6 `INTERFACE` libraries with `IMPORTED` targets.

ChrisThrasher:
I was able to complete this :)

Issue #2078, opened 2022-04-14
cmake: Target xxx is already defined


JonathSpirit:
## Subject of the issue

While working on a cmake project that required SFML, I discoverd that adding external library used by SFML with add_subdirectory doesn't work cause SFML try to create a INTERFACE target with the same name.

This cause the impossibility to build a subdirectory with cmake (useful for flexibility and workflow) and the only possibility I see is to build the library out of the top CMakeLists.txt by hand (or maybe an cmake addon like ExternalProject but this is not a solution for me). 

This cause extra overhead for nothing.

## Your environment

* Windows 10, 21H2
* SFML branch 2.6.x
* MSYS2 : (Rev10, Built by MSYS2 project) 11.2.0
* -D SFML_BUILD_NETWORK=OFF
* cmake version 3.22.3

## Steps to reproduce

An good example is to clone the openal-soft project from https://github.com/kcat/openal-soft.git and do in your top CMakeLists : 

```cmake

project(MyProject)


set(ALSOFT_UTILS OFF CACHE BOOL "" FORCE) #OpenAL-soft
set(ALSOFT_EXAMPLES OFF CACHE BOOL "" FORCE)
add_subdirectory("libs/openal-soft" EXCLUDE_FROM_ALL)

set(OPENAL_INCLUDE_DIR "libs/openal-soft/include" CACHE PATH "" FORCE) #needed for being detected by findPackage
set(OPENAL_LIBRARY "OpenAL" CACHE STRING "" FORCE)

#And then add SFML

add_subdirectory("libs/SFML" EXCLUDE_FROM_ALL)

```

## Expected behavior

SFML should correctly use the target created by the add_subdirectory with the help of findPackage

## Actual behavior

SFML cause an Cmake error : 

```cmake
CMake Error at libs/SFML/cmake/Macros.cmake:379 (message):
  Target 'OpenAL' is already defined

```

JonathSpirit:
Maybe a good fix is to create the INTERFACE target with a suffix/prefix like "OpenAL_sfml" in Macros.cmake

ChrisThrasher:
Why do you want to include OpenAL from two locations? If SFML already depends on it, why express that dependency again in your own build? Does this problem exist when you use `find_package` to depend on SFML or only when you include SFML sources in your build?

eXpl0it3r:
[...]

PR #2650, opened 2023-08-15
Replace more uses of `sfml_find_package` with `find_package`

ChrisThrasher:
## Description

In any green field CMake projects these days the expectation for dependencies is that they're found in the build scripts via `find_package` and consumed by downstream users in the config module via `find_dependency`. These two features handle a lot of tricky problems that make it feasible to compose large dependency graphs out of CMake projects. They're recommended for good reason.

My goal in this PR is to take a few more steps toward that solution. This is not the first PR of its kind and won't be the last. One of my goals is to entirely remove `sfml_find_package`.

Under the hood `sfml_find_package` is using `find_package` which is good but it was also doing a few other things we didn't need. It was supporting more complicated use cases than we actually needed so I simplified its input handling to only support how we're currently using it. There's no reason to believe our use of `sfml_find_package` will get more complicated over time.

After that I went on to convert our OpenGL, DRM, GBM, EGL, GLES, and UDev dependencies to use pure `find_package`/`find_dependency`. This required first making `IMPORTED` targets inside those dependency's respective find modules. This is canonical CMake usage. All good Find modules and Config modules these days are producing `IMPORTED` targets. This differs from how `sfml_find_package` worked where it generated `INTERFACE` targets which we then had to export. These `IMPORTED` targets correctly express to the build system how SFML doesn't "own" these targets. All targets found through `find_package` (conventionally) behave like this.

Along the way I realized that now we need downstream users to have access to these custom find modules we wrote. I'm referring to all the .cmake files in cmake/Modules. In the same way we need them in the build, we need to expose them to users so they too can find these dependencies. Typically projects don't think about this because they're using a config module from a 3rd party dependency or using a find module that ships with CMake (like FindBoost.cmake). If CMake already provides the find module then projects can expect that all consumers already have it. In our case we're writing find modules that don't exist anywhere but SFML so we need to make sure users get that code.

How was it we weren't doing this before? `sfml_bind_dependency` is roughly approximating what our find modules already do. In a sense we were duplicating code in our find modules and in `sfml_bind_dependency` to make sure that downstream users could find the libraries we were using. As soon as `sfml_find_package` is gone, `sfml_bind_dependency` will also disappear.

One more small step towards achieving the vision in https://github.com/SFML/SFML/issues/1991. One we're using `find_package` exclusively we'll be much better equipped to experiment with ways of removing all the vendored binaries and headers in extlibs.

See also:
https://github.com/SFML/SFML/issues/2459
https://github.com/SFML/SFML/issues/2078


ChrisThrasher:

ChrisThrasher:
I was able to switch OpenGL over to pure `find_package` and thus use the imported `OpenGL::GL` target instead of the target we were creating ourselves.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:
https://github.com/microsoft/vcpkg/blob/86bdd8f99a9dab84e8eae40c92ae20ebc9045e83/ports/sfml/fix-dependencies.patch#L34-L36

Vcpkg seems to claim that we need to ensure the value of `OpenGL_GL_PREFERENCE` is propagated to consumers.

ChrisThrasher:
> https://github.com/microsoft/vcpkg/blob/86bdd8f99a9dab84e8eae40c92ae20ebc9045e83/ports/sfml/fix-dependencies.patch#L34-L36
> 
> Vcpkg seems to claim that we need to ensure the value of `OpenGL_GL_PREFERENCE` is propagated to consumers.

We're not currently doing this in `master` so this PR doesn't make this situation any worse as far as I can tell. I'm still curious what are the ramifications of not doing that. I don't know about OpenGL to say.

eXpl0it3r:
> The [FindOpenGL](https://cmake.org/cmake/help/latest/module/FindOpenGL.html#module:FindOpenGL) module provides an OpenGL::GL target and an OPENGL_LIBRARIES variable for projects to use for legacy GL interfaces. When both a legacy GL library (e.g. libGL.so) and GLVND libraries for OpenGL and GLX (e.g. libOpenGL.so and libGLX.so) are available, the module must choose between them. It documents an OpenGL_GL_PREFERENCE variable that can be used to specify an explicit preference. When no such preference is set, the module must choose a default preference.
~ [CMP0072](https://cmake.org/cmake/help/latest/policy/CMP0072.html)

So it depends on the user's system of what's available and what to link to if both exist.

ChrisThrasher:
On second thought, I maybe do see how need to copy what Vcpkg is doing. I think we may need to change it to

```cmake
            if(NOT OpenGL_GL_PREFERENCE)
                set(OpenGL_GL_PREFERENCE "LEGACY")
            endif()
            find_dependency(OpenGL COMPONENTS OpenGL)
```

In `master` as it stands today, users are not calling `find_dependency(OpenGL)` so they're instead getting our own `OpenGL` target that we create inside `sfml_find_package` which has the correct link libraries encoded in it. Once they start using `OpenGL::OpenGL` from FindOpenGL.cmake then they're getting a target we not longer control so we need to be sure to set the same parameters before `find_package(OpenGL)` as we do `find_dependency(OpenGL)`.

ChrisThrasher:
https://gitlab.kitware.com/vtk/vtk/-/blob/master/CMake/vtk-config.cmake.in#L150

A maintainer of VTK confirmed to me that we do need to ensure that `OpenGL_GL_PREFERENCE` gets set before we call `find_dependency(OpenGL)`. I'm now confident again that this PR is in a mergeable state.

cmake/Modules/FindDRM.cmake
```diff
@@ -16,3 +16,8 @@ find_library(DRM_LIBRARY NAMES drm)
 
 include(FindPackageHandleStandardArgs)
 find_package_handle_standard_args(DRM DEFAULT_MSG DRM_LIBRARY DRM_INCLUDE_DIR)
+
+add_library(DRM::DRM IMPORTED UNKNOWN)
```
Rosme, line 20: *chef kiss*

Rosme (approved):
Looks good for me. What would be the next step for this after?

eXpl0it3r (approved):

PR #2651, opened 2023-08-15
Skip ClientMessage events with other window ID unless it is for IM

### LINKED ISSUE #2649: With multiple `sf::RenderWindow`s, the first one to poll receives the Closed event (Ubuntu / X11)
## Subject of the issue

When managing multiple `sf::RenderWindow` objects, only one window receives the `Closed` event (the same window, regardless of which window was closed by the user). _Edit: Based on further testing, the window that receives the closed event is almost certainly the one that polls first._

## Your environment

* Your OS / distro / window manager used
* 2.6.x, also confirmed on SFML 3
* GCC 12
* No custom flags (can repro with basic example code below)

## Steps to reproduce

```cpp
#include <SFML/Window.hpp>
#include <SFML/Graphics.hpp>
#include <iostream>

int main()
{
    // Create the main rendering window
    sf::RenderWindow App(sf::VideoMode(800, 600), "App1");
    sf::RenderWindow App2(sf::VideoMode(800, 600), "App2");
    
    // Start game loop
    while (App.isOpen() || App2.isOpen())
    {
        // Process events
        sf::Event Event;
        while (App.pollEvent(Event))
        {
            // Close window : exit
            if (Event.type == sf::Event::Closed) {
                App.close();
                std::cout << "app close" << std::endl;
            }
        }
        while (App2.pollEvent(Event))
        {
            // Close window : exit
            if (Event.type == sf::Event::Closed) {
                App2.close();
                std::cout << "app2 close" << std::endl;
            }
        }

        // Clear the screen (fill it with black color)
        App.clear();
        App2.clear();

        // Display window contents on screen
        App.display();
        App2.display();
    }
}
```

## Expected behavior

Clicking the `x` button on either window should close that window & only that window.

## Actual behavior

Clicking the `x` button on the second window closes the first window. Clicking the `x` button on the first window also closes the first window.

## Notes

I have a few ideas about what could be happening to cause this behaviour at a high level, but can't test them right now. My guess from the behaviour is that the first window that polls is getting the close event, which would indicate that there isn't distinction at a window level when receiving these particular X messages. This should be easily testable by adding sleeps before the polls to guarantee which poll is checked first after a close.

kimci86:
## Description

This is an attempt to fix #2649.

## Tasks

* [ ] Tested on Linux

## How to test this PR?

Running example program from issue #2649.

### DIFF
--- a/src/SFML/Window/Unix/WindowImplX11.cpp
+++ b/src/SFML/Window/Unix/WindowImplX11.cpp
@@ -85,13 +85,25 @@ namespace
 
         static const unsigned int             maxTrialsCount = 5;
 
-        // Filter the events received by windows (only allow those matching a specific window)
+        // Filter the events received by windows (only allow those matching a specific window or those needed for the IM to work)
         Bool checkEvent(::Display*, XEvent* event, XPointer userData)
         {
-            // Just check if the event matches the window
-            // The input method sometimes sends ClientMessages with a different window ID,
-            // our event loop has to process them for the IM to work
-            return (event->xany.window == reinterpret_cast< ::Window >(userData)) || (event->type == ClientMessage);
+            if (event->xany.window == reinterpret_cast<::Window>(userData))
+            {
+                // The event matches the current window so pick it up
+                return true;
+            }
+            if (event->type == ClientMessage)
+            {
+                // The input method sometimes sends ClientMessage with a different window ID.
+                // Our event loop has to process them for the IM to work.
+                // We assume ClientMessage events not having WM_PROTOCOLS message type are such events.
+                // ClientMessage events having WM_PROTOCOLS message type should be handled by their own window,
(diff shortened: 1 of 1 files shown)


kimci86:
This solves the issue for me.

@MJJBennett could you try this branch?

ChrisThrasher:
I can no longer reproduce the error

MJJBennett:
Tested on my personal project and this fixes the problem I was seeing. Also tested to make 100% sure on my test project, and this definitely fixes the issue (the event is polled by the correct window).

The change itself & comments also look great 👍 Thanks for fixing this!

ChrisThrasher (approved):

ChrisThrasher:
Thank you! Glad to see 2.6.1 get yet another great bug fix like this.

PR #2572, opened 2023-06-04
Move semantics for `sf::Drawable` , `sf::RenderTarget`, `sf::RenderTexture` and `sf::VertexBuffer`

Griezn:
## Description

I implemented the move semantics for `sf::Drawable` , `sf::RenderTarget`, `sf::RenderTexture` and `sf::VertexBuffer`.

This PR is related to the issue #2483 

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Describe how to best test these changes. Please provide a [minimal, complete and verifiable example](https://stackoverflow.com/help/mcve) if possible, you can use the follow template as a start:

```cpp
#include <SFML/Graphics.hpp>
#include <iostream>

class CustomDrawable : public sf::Drawable {
public:
    explicit CustomDrawable(sf::VertexArray vertices) : m_vertices(std::move(vertices)) {}

private:
    void draw(sf::RenderTarget& target,const sf::RenderStates& states) const override {
        target.draw(m_vertices, states);
    }

    sf::VertexArray m_vertices;
};

int main() {
    // Create a render window
    sf::RenderWindow window(sf::VideoMode({ 800, 600 }), "Move Semantics Example");

    // Create a RenderTexture
    sf::RenderTexture renderTexture;
    if (!renderTexture.create({400, 300})) {
        std::cerr << "Failed to create RenderTexture" << std::endl;
        return 1;
    }
    sf::RenderTexture movedRenderTexture(std::move(renderTexture));
    std::cout << "moved RenderTexture" << std::endl;

    // Create a VertexBuffer
    sf::VertexBuffer vertexBuffer(sf::VertexBuffer::Static);
    if (!vertexBuffer.create(4)) {
        std::cerr << "Failed to create VertexBuffer" << std::endl;
        return 1;
    }
    sf::VertexBuffer movedVertexBuffer(std::move(vertexBuffer));
    std::cout << "moved vertexBuffer" << std::endl;

    // Create a custom drawable object
    sf::VertexArray vertices(sf::PrimitiveType::Triangles, 4);
    vertices[0].position = sf::Vector2f(0, 0);
    vertices[1].position = sf::Vector2f(100, 0);
    vertices[2].position = sf::Vector2f(100, 100);
    vertices[3].position = sf::Vector2f(0, 100);
    CustomDrawable customDrawable(std::move(vertices));

[...]

PR #2656, opened 2023-08-22
Add move semantics to `sf::RenderTarget` and `sf::RenderTexture`

ChrisThrasher:
## Description

These classes had their move operations implicitly deleted because they both defined destructors. `sf::RenderTexture` defined a destructor because it contained a pointer to an opaque type. `sf::RenderTarget` has a destructor to ensure derived classes get properly destroyed under all circumstances.

Both situations resulted in move semantics being implicitly disabled. I don't believe either class contains anything that prevent us from enabling the default move operations because neither class uses a non-default destructor.

Closes #2572


ChrisThrasher:

kimci86:
My only concern about this PR is whether it is valid or not to move those objects between threads.
I tried to make it break and I have the following code that crashes at the end of main function when the shared OpenGL context is being cleaned up. I am not sure if this is misusing the API or a bug. I will try to investigate further in the coming days.

```c++
#include <SFML/Graphics.hpp>

#include <future>

int main()
{
    auto rt = sf::RenderTexture{};
    assert(rt.create({200, 150}));
    rt.clear(sf::Color::Red);
    assert(rt.setActive(false));

    std::async(
        std::launch::async,
        [](sf::RenderTexture rt2)
        {
            assert(rt2.setActive(true));
            rt2.clear(sf::Color::Green);
        },
        std::move(rt))
        .get();
}

```

ChrisThrasher:
That test program segfaults for me in Debug mode (but not Release) on macOS.

ChrisThrasher:
Screenshot 2023-08-23 at 6 18 10 PM

The crash is because `m_impl` is null when we try to access it.

ChrisThrasher:
Screenshot 2023-08-23 at 6 27 12 PM

Sometimes I get this crash in a slightly different place. I believe in this case it's trying to call a pure virtual function after the class implementing that pure virtual has been destroyed but I'm not positive about that.

ChrisThrasher:
```
thrasher ~/Projects/mandelbrot/build$ /Users/thrasher/Projects/mandelbrot/build/mandelbrot
/Users/thrasher/Projects/mandelbrot/build/_deps/sfml-src/src/SFML/Window/GlContext.cpp:879:27: runtime error: member call on address 0x00010751adc0 which does not point to an object of type 'sf::priv::GlContext'
0x00010751adc0: note: object has invalid vptr
 00 00 00 00  5c 2b 00 00 03 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 02 00 00 00  01 00 00 00
              ^~~~~~~~~~~~~~~~~~~~~~~
              invalid vptr
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Users/thrasher/Projects/mandelbrot/build/_deps/sfml-src/src/SFML/Window/GlContext.cpp:879:27 in 
Abort trap: 6
```

UB San caught this as well.

ChrisThrasher:

test/Graphics/RenderTarget.test.cpp
```diff
@@ -8,6 +8,12 @@
 
 class RenderTarget : public sf::RenderTarget
 {
+public:
+    RenderTarget() // NOLINT(modernize-use-equals-default)
```
ChrisThrasher, line 12: Idk why but using ` = default` breaks compilation. Maybe it's related to me moving some `= default`s from .cpp files to .hpp files.

```
/Users/thrasher/Projects/sfml/test/Graphics/RenderTarget.test.cpp:34:28: error: default initialization of an object of const type 'const RenderTarget' without a user-provided default constructor
        const RenderTarget renderTarget;
```

kimci86: This is an interesting error. The full error message with GCC 12.3 tells that it refuses to instantiate a const RenderTarget because:
> the implicitly-defined constructor does not initialize ‘bool sf::RenderTarget::StatesCache::enable’

Maybe `sf::RenderTarget::m_cache`  member variable should be initialized in `sf::RenderTarget` to fix the error instead of avoiding it?

ChrisThrasher: Good catch, that totally fixed it!

ChrisThrasher:

eXpl0it3r:
> My only concern about this PR is whether it is valid or not to move those objects between threads.

SFML doesn't really make any guarantees for multi-threading. I guess we could/should document the limitation somewhere, but at the end of the day, 99.9% of SFML usages should be fine in a single thread and any other cases, you should really not be multi-threading your OpenGL/render code.

eXpl0it3r (approved):

ChrisThrasher:
> I guess we could/should document the limitation somewhere, but at the end of the day

I added a blurb to the render target class docs about this. Let me know if I captured it well.

ChrisThrasher:

eXpl0it3r (approved):

include/SFML/Graphics/RenderTarget.hpp
```diff
@@ -525,6 +537,12 @@ class SFML_GRAPHICS_API RenderTarget
 /// OpenGL states are not messed up by calling the
 /// pushGLStates/popGLStates functions.
 ///
+/// While render targets are moveable, it is not valid to move them
+/// between threads. This will cause your program to crash. The
+/// problem boils down to OpenGL being limited with regard to how it
+/// works in multithreaded environments. Please ensure you only move
+/// render targets within the same thread.
```
kimci86, line 544: I am not sure but I naively imagine we could build abstractions that transfer information from one thread's OpenGL context to another, so maybe it would be more accurate to tell the problem is that moving an instance between threads is not supported by SFML rather than an fundamental OpenGL limitation. Can someone familiar with OpenGL confirm?

PR #2657, opened 2023-08-22
Clean up implementation of `sf::Shader`

ChrisThrasher:
## Description

I found these while working on `sf::Shader::loadFromFile` tests a little while back.

### DIFF
--- a/src/SFML/Graphics/Shader.cpp
+++ b/src/SFML/Graphics/Shader.cpp
@@ -120,12 +120,11 @@ bool getStreamContents(sf::InputStream& stream, std::vector<char>& buffer)
 }
 
 // Transforms an array of 2D vectors into a contiguous array of scalars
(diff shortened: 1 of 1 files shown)


kimci86 (commented):
This is assuming a file cannot be empty but it is not true. Being empty is not the same as being missing.

ChrisThrasher:
> This is assuming a file cannot but empty but it is not true. Being empty is not the same as being missing.

Both function push back a null character before returning the buffer so even if the file is empty these should still work.

kimci86:
> Both function push back a null character before returning the buffer so even if the file is empty these should still work.

Ok that makes sense. However that is not the case in `getStreamContents` when the stream size is 0.

ChrisThrasher:

eXpl0it3r (approved):

PR #2661, opened 2023-08-23
Disable faulty Windows Clang job

ChrisThrasher:
## Description

The problem is that the Windows CI image hasn't yet be updated to Clang 16. See this PR for progress on fixing it

https://github.com/actions/runner-images/pull/8134

Running clang-tidy in Debug mode also means we avoid running another Windows Clang Static Release job. In general it's probably good to keep running clang-tidy across debug configurations though so that change can stay.

Here's what CI on `master` currently looks like: https://github.com/SFML/SFML/actions/runs/5954008177

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
         - { name: Windows VS2022 x86,     os: windows-2022, flags: -DSFML_USE_MESA3D=TRUE -A Win32 }
         - { name: Windows VS2022 x64,     os: windows-2022, flags: -DSFML_USE_MESA3D=TRUE -A x64 }
         - { name: Windows VS2022 ClangCL, os: windows-2022, flags: -DSFML_USE_MESA3D=TRUE -T ClangCL }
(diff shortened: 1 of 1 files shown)


eXpl0it3r (approved):

ChrisThrasher:
Even though the fix for this has been merged we still sometimes get failures. The failures are pretty sporadic so I'm erring on the side of caution and disabling all jobs that use Clang on Windows. We will need to keep these workarounds in place until GitHub Actions runners are finally shipping with Clang 16.

ChrisThrasher:
https://github.com/SFML/SFML/actions/runs/6053687811

I'm glad I removed all Clang Windows jobs because now all five of them are consistency failing. Previously only a subset of these five were failing.

ChrisThrasher:
https://github.com/ChrisThrasher/argon/actions/runs/6104730037/job/16567174833#step:3:15

In another project I finally can see it using Clang 16. Most jobs still use Clang 15 but hopefully this will be fixed shortly and we can restore the Windows Clang jobs.

PR #2663, opened 2023-08-24
Modernize android example

texus:
The android example hasn't been updated for a long time, so I decided to modernize it.

Changes compared to the old example:
- Replaced ndk-build with CMake
- Updated Groovy DSL (.gradle files) to Kotlin DSL (.gradle.kts files)
- Updated Android Gradle plugin from 3.0 to 7.4
- Config file now targets a more recent Android SDK (and NDK)
- Remove deprecated settings from AndroidManifest

The most important files to look at are `app/src/main/jni/CMakeLists.txt` and `app/build.gradle.kts`

Currently the CMake script has to manually copy the libopenal.so into the apk, but it should probably be possible somehow to do this automatically when linking to SFML::Audio.

## How to test this PR?

First build SFML and install it into the NDK. The ARCH_ABI will likely be arm64-v8a for a real device (i.e. when not building for a simulator).
```
cmake -DCMAKE_SYSTEM_NAME=Android -DCMAKE_ANDROID_NDK=/path/to/ndk -DCMAKE_ANDROID_ARCH_ABI=x86_64 -DCMAKE_ANDROID_STL_TYPE=c++_static -DCMAKE_BUILD_TYPE=Debug ..
make && make install
```

In the SFML/example/android directory, edit `app/build.gradle.kts` with the correct settings (ndkVersion and abiFilters properties need to match what was used to build SFML).

Import the SFML/example/android directory in Android Studio and run it from there, or alternatively use `gradle buildDebug` and `gradle installDebug` from the terminal to build the project and install it to a connected Android device or running AVD simulator.


examples/android/app/src/main/jni/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+cmake_minimum_required(VERSION 3.22)
+project(sfml-example)
+
+set(SFML_DIR "${CMAKE_ANDROID_NDK}/sources/third_party/sfml/lib/${CMAKE_ANDROID_ARCH_ABI}/cmake/SFML")
```
ChrisThrasher, line 4: Is this a standard android path or something specific to your particular machine?

examples/android/app/src/main/jni/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+cmake_minimum_required(VERSION 3.22)
+project(sfml-example)
```
ChrisThrasher, line 2:
```suggestion
project(sfml-example CXX)
```

examples/android/app/src/main/jni/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+
+set(SFML_DIR "${CMAKE_ANDROID_NDK}/sources/third_party/sfml/lib/${CMAKE_ANDROID_ARCH_ABI}/cmake/SFML")
+
+# Create the libapp.so library that contains the application's c++ code
+add_library(app SHARED)
+target_sources(app PRIVATE main.cpp)
```
ChrisThrasher, line 8:
```suggestion
add_library(app SHARED main.cpp)
```

examples/android/app/src/main/jni/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+
+# Find and link SFML.
+# We must link to the network and audio components even if we don't use them, because sfml-activity will try to load them
+find_package(SFML 3 REQUIRED COMPONENTS Graphics Audio Network CONFIG)
```
ChrisThrasher, line 12:
```suggestion
find_package(SFML 3 REQUIRED COMPONENTS Graphics Audio Network)
```

It's unlikely that anyone uses a fine module to find SFML 3 but it's not invalid to do so let's not require using a config module.

examples/android/app/src/main/jni/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+# our libapp.so file so that is can be loaded at runtime by sfml-activity.
+target_link_libraries(app PUBLIC
+  -Wl,--whole-archive
+  SFML::Main
+  -Wl,--no-whole-archive
+)
+
+# We need to manually copy the OpenAL library into the apk, even if we don't use it, because sfml-activity will try to load it
+add_library(ImportOpenAL SHARED IMPORTED)
+set_target_properties(ImportOpenAL PROPERTIES IMPORTED_LOCATION "${CMAKE_ANDROID_NDK}/sources/third_party/sfml/extlibs/lib/${CMAKE_ANDROID_ARCH_ABI}/libopenal.so")
```
ChrisThrasher, line 25: Also curious about this path and whether it's portable across any machine someone may use to build this.

examples/android/app/src/main/jni/main.cpp
```diff
@@ -153,6 +153,8 @@ int main(int argc, char* argv[])
 #endif
                     }
                     break;
+                default:
+                    break;
```
ChrisThrasher, line 157: Why did you add this?

examples/android/.gitignore
```diff
@@ -1,8 +1,12 @@
 local.properties
 .idea
 .gradle
+gradle
+gradlew
+gradlew.bat
```
ChrisThrasher, line 6: I don't see any files matching these patterns in the repo. Are they generated as part of the Gradle build?

ChrisThrasher (changes requested):

examples/android/app/src/main/jni/main.cpp
texus: Without it, clang shows a warning during compilation:
`warning: 15 enumeration values not handled in switch: 'TextEntered', 'KeyReleased', 'MouseWheelScrolled'... [-Wswitch]`

examples/android/.gitignore
texus: Those were generated when I imported and built the project with Android Studio.

After thinking about it, those probably need to be checked into git instead of ignored, as they provide the gradle version that was selected when building the project. It is often a good idea to use such a gradlew file instead of running the native gradle from the OS because it guarantees that the used version is compatible with the project.

examples/android/app/src/main/jni/CMakeLists.txt
texus: It is the default value for CMAKE_INSTALL_PREFIX when building SFML: https://github.com/SFML/SFML/blob/master/CMakeLists.txt#L92

So as long as the user doesn't manually specify another path, it will be where SFML is installed to. It is a standard location for android libraries to install to, so I imagine most people will keep the path.

The same goes for the openal.so path, it is the path when SFML places that file by default.

ChrisThrasher: Ah okay I see. I didn't know the android build did that. We can keep this.

texus:
I'm using Android Studio 2022.3.1

According to https://developer.android.com/build/releases/gradle-plugin?buildsystem=ndk-build#android_gradle_plugin_and_android_studio_compatibility, Electric Eel | 2022.1.1 supports up to AGP 7.4 (which is what I selected, and I did so because it looked like any version from 2022 and 2023 would support it according to those docs). Do they perhaps support 7.4.0 but not my chosen 7.4.1?

There is no reason it has to be this high though. I started by copying an existing project from me that was using AGP 7.3.1 and only updated to AGP 7.4.1 because Android Studio suggested it. I'm sure it will work if I go even lower though. I'll look into it later, I might set it to e.g. 7.1 or 7.0 to also have compatibility with Android Studio versions from 2021 as well.

Zombieschannel:
I don't think the issue is in AGP, but the general Gradle version. Since I get this error specifically:
> An exception occurred applying plugin request [id: 'com.android.application']
Failed to apply plugin 'com.android.internal.version-check'.
Minimum supported Gradle version is 7.5. Current version is 7.4. 

Changing the Gradle version in project settings does not seem to do anything, no idea why.

Zombieschannel (approved):
Builds and works well.

examples/android/app/src/main/jni/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+# The ANativeActivity_onCreate function from SFML::Main needs to be exposed in
+# our libapp.so file so that is can be loaded at runtime by sfml-activity.
+target_link_libraries(app PUBLIC
+  -Wl,--whole-archive
+  SFML::Main
+  -Wl,--no-whole-archive
+)
+
+# We need to manually copy the OpenAL library into the apk, even if we don't use it, because sfml-activity will try to load it
+add_library(ImportOpenAL SHARED IMPORTED)
```
ChrisThrasher, line 23:
```suggestion
add_library(OpenAL::OpenAL SHARED IMPORTED)
```
Let's use the same target name as what's in [FindOpenAL.cmake](https://cmake.org/cmake/help/latest/module/FindOpenAL.html).

.github/workflows/ci.yml
```diff
@@ -133,7 +139,17 @@ jobs:
     - name: Build
       shell: bash
       run: cmake --build build --config ${{ matrix.type.name == 'Debug' && 'Debug' || 'Release' }} --target install
-      
+
+    - name: Build Android example
+      if: matrix.platform.name == 'Android'
+      shell: bash
+      run: |
+        cd examples/android
```
ChrisThrasher, line 147: Is it possible to do this without changing directories? If this is required, let's be sure to run `cd -` at the end of this job so that the current directory isn't affected for later jobs.

.github/workflows/ci.yml
```diff
@@ -133,7 +139,17 @@ jobs:
...
+        ARCH_ABI=$(echo ${{ matrix.config.flags }} | sed 's/.*-DCMAKE_ANDROID_ARCH_ABI=\([[:graph:]]*\).*/\1/')
+        sed "s/abiFilters.add(\"[[:alnum:]_-]*\")/abiFilters.add(\"${ARCH_ABI}\")/g" -i app/build.gradle.kts
+        sed "s/ndkVersion = \"[[:digit:].]*\"/ndkVersion = \"25.2.9519653\"/g" -i app/build.gradle.kts
```
ChrisThrasher, line 150: Why are all these find/replace `sed` calls necessary?

.github/workflows/ci.yml
kimci86: Following steps are not impacted by `cd`.
> Each run keyword represents a new process and shell in the runner environment.

See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun

It is possible to replace `cd` by `working-directory:`.

.github/workflows/ci.yml
texus: The settings in build.gradle.kts need to be changed. The architecture to use (e.g. x86) is hardcoded in the example, but the CI needs to test different architectures. I'm open for alternatives on this. I do realize that the last two sed commands are harder than they need to be, I'll change them into something slightly simpler.

The ndkVersion technically doesn't need to be changed. It is already the same on the CI and in the example, so the last sed command can be removed. The question is: will you guarantee that the example remains in sync with the NDK version used on the CI, or will you allow them to be independent from each other?

.github/workflows/ci.yml
texus: It can be done without changing directory, but I'll change it to `working-directory` so that the path doesn't need to be repeated on each line.

.github/workflows/ci.yml
ChrisThrasher: > but the CI needs to test different architectures

Is it worth testing on all architectures versus just 1? Is it possible that one of these Gradle jobs fails but the others don't? If they either all pass or all fail then we can maybe get away with a single one of them. I'd rather not fight with Gradle.

.github/workflows/ci.yml
kimci86: To clarify, you wouldn't have to repeat `cd` as long as you are in the same `run` step. You can have several commands in a `run` step as you do, which are executed in the same shell. I was answering the ChrisThrasher who was worried it would impact the rest of the workflow. But `working-directory` is an alternative and looks nice.

.github/workflows/ci.yml
texus: > Is it worth testing on all architectures versus just 1?

Probably not. If we only test one, and I can't change the example with sed commands, then the example will just have to build with the exact same ABI and NDK as is tested by the CI. That's not really an issue, it's just something to keep in mind if the CI or example are ever changed later.

ChrisThrasher: > the example will just have to build with the exact same ABI and NDK as is tested by the CI

I think I'm cool with that. I'd rather deal with that then have to patch the build scripts to make CI work.

texus:
I figured out how to pass parameters to Gradle, so the CI can now test multiple configurations again.

While it is unlikely that one test fails while the others succeed, it could happen in some rare cases (e.g. I noticed that SFML can be build with any API level except 25, that version looks broken). If the CI would be able to run the executable instead of only building it then it would find additional issues (as some of the tested configurations currently won't run until we remove sfml-activity later).

eXpl0it3r:

ChrisThrasher:
One of the extended CI jobs for Android is failing. Do you know what that is about? Do we need to update our Buildbot to handle the new way we're building the Android example?

texus:
The Buildbot script should run `./gradlew assembleDebug` instead of  `gradle assembleDebug` (and similar for release). Then it will automatically download and run with a working gradle version.

You could also update the gradle version on the system, but the gradle wrapper (gradlew) script that I included in the example exists for the very reason that it will work no matter whether the gradle version on the system is too low or too high.

eXpl0it3r:
> the gradle wrapper (gradlew) script that I included

Would that mean, we wouldn't even need to have gradle pre-installed?

One of the issues with buildbot is, that the configuration is the same for all version. So I [made that change](https://github.com/SFML/SFML-Buildbot/pull/42), it will cause failures for any 2.x based branches, non-rebased branches, or even master until this PR is merged.

Would updating gradle allow for a use with the new and old branches? What version would we need to upgrade to?

texus:
> Would that mean, we wouldn't even need to have gradle pre-installed?

I think it should work even without gradle installed.

> it will cause failures for any 2.x based branches

Then maybe a gradle wrapper should be used for 2.x as well (possibly with an older configured gradle version as used in master)

> Would updating gradle allow for a use with the new and old branches

The problem with gradle is that it is neither forward-compatible nor backwards-compatible for more than several versions. While it might work, if you upgrade the gradle version to a too recent version then you risk that the 2.x project doesn't run anymore.

The best solution to this problem might be to run `gradle wrapper` before running the `./gradlew` command in the Buildbot script. It will generate the gradlew files. On the master branch I've configured this command to generate a wrapper for Gradle 7.4 (the same wrapper as I included in git), on the 2.x branch this should generate a wrapper with Gradle 7.3.1 (the same as the installed version, as no `--gradle-version` is being passed as parameter to `gradle wrapper`). Afterwards you should be able to run `./gradlew` on any branch.

Originally I was thinking of not committing the gradlew files at all and just required the users to run `gradle wrapper` themselves, but from what I have read there might be gradle versions that can't generate the wrapper file if the gradle version can't build the project (which is a big issue because you need to generate the wrapper file exactly because it generates a version that can build the project). So you may need the account for the possibility in buildbot that the `gradle wrapper` command can fail.

binary1248:
Buildbot is able to build both versions now. A new patch was necessary to build the example due to the CMake version requirement hard coded [here](https://github.com/SFML/SFML/pull/2663/files#diff-94771b3d681ac0bcc1cdd493438a27b36c2262487b5d5014591bea46ae59b094R41). It would be nice if this wouldn't be necessary.

ChrisThrasher:
The conflicts between this branch and `master` are a bit tricky. Not entirely sure yet how to resolve them. Part of the conflicts are caused by #2768.

texus:
> A new patch was necessary to build the example due to the CMake version requirement hard coded here. It would be nice if this wouldn't be necessary.

I did some testing and was able to remove the version if I bump the AGP to 7.4 and Gradle version to 7.5. With AGP 7.3 (which this PR was using), gradle would automatically download CMake 3.18 which was too old for SFML 3. CMake 3.22.1 is the only CMake version available in the Android SDK manager that is new enough, which is why I hardcoded that version. You just needed to run `sdkmanager --install "cmake;3.22.1"` to get the correct version installed.

When AGP 7.4 is used, CMake 3.22.1 is automatically downloaded when build.gradle.kts doesn't explicitly set a CMake version.

Since the gradle wrapper is now used, it shouldn't be an issue to increase the gradle version, as it no longer depends on what is installed on the system.

> The conflicts between this branch and master are a bit tricky. Not entirely sure yet how to resolve them

I'll see if I can rebase this branch.

texus:
I've recreated this PR on top of the latest master version. A few changes were made compared to the earlier version:
- Gradle, AGP and NDK versions were increased
- CMake version is no longer hardcoded build.gradle.kts
- libapp.so is renamed back to libsfml-example.so, since this was also the name used in the old ndk-build (naming it libapp.so was never really intentional)

eXpl0it3r:

.github/workflows/ci.yml
```diff
@@ -95,8 +100,8 @@ jobs:
     - name: Install Android Components
       if: matrix.platform.name == 'Android'
       run: |
-        wget -nv https://dl.google.com/android/repository/android-ndk-r24-linux.zip -P $GITHUB_WORKSPACE
-        unzip -qq -d $GITHUB_WORKSPACE android-ndk-r24-linux.zip
+        echo "y" | ${ANDROID_HOME}/tools/bin/sdkmanager --install "build-tools;33.0.2"
+        echo "y" | ${ANDROID_HOME}/tools/bin/sdkmanager --install "ndk;26.1.10909125"
```
ChrisThrasher, line 104:
```suggestion
        yes | ${ANDROID_HOME}/tools/bin/sdkmanager --install "build-tools;33.0.2"
        yes | ${ANDROID_HOME}/tools/bin/sdkmanager --install "ndk;26.1.10909125"
```
There's a built-in program for spamming `y` at CLI prompts.

ChrisThrasher (approved):

.github/workflows/ci.yml
texus: Apparently there is a reason why `yes` wasn't used here. I tried it but it results in a `yes: standard output: Broken pipe` error.

ChrisThrasher:
Thanks for your help with Android support!

PR #2665, opened 2023-08-27
Add tests for `sf::Event`

ChrisThrasher:
## Description

Not a whole lot to test here but it's still good to have the basics in place in case we want to change `sf::Event`.

I do want to point out how all data in events are uninitialized by default which opens the door for UB if you accidentally read data out of an event before it's initialized. I'm not aware of this being a bug in practice but it's still something we should patch so that it's never possible to read uninitialized data. It's trivial to give `sf::Event::type` an initial value but I'm not yet sure how to give an initial value to that anonymous `union`.


kimci86:
The solution to avoid UB in this case is `std::variant`.

binary1248 (approved):

ChrisThrasher:
> The solution to avoid UB in this case is `std::variant`.

I'd like to do that at some point

PR #2666, opened 2023-08-27
Convert all dependencies to `IMPORTED` targets

ChrisThrasher:
## Description

This PR is the culmination of about a year of me trying to get rid of `sfml_find_package`. Continuation of #2650. 

The main issue with `sfml_find_package` is that it creates `INTERFACE` targets rather than `IMPORTED` targets. CMake convention is to ensure all external targets are `IMPORTED`. This tells CMake that a given target is not owned by the project but is rather imported into the project. By convention any time you call `find_package` it ought to produce 1 or more `IMPORTED` targets. SFML's own config module produces `IMPORTED` targets when someone calls `find_package(SFML)`.

This has the bonus of fixing bugs like how the CMake SFML Template fails to copy openal32.dll into the users' build making the Audio module cumbersome to use on Windows. The template project can only copy any `IMPORTED` targets but won't copy `INTERFACE` targets.

---

The mechanics of how I finally got rid of `sfml_find_package` are a little messy so I'll walk through them.

Converting `FLAC::FLAC` was one of the easier ones. I had to add some code to FindFLAC.cmake to create the imported target which includes some special case code to handle macOS Frameworks. Once that's in place we can directly call `find_package(FLAC)` and we'll get the `FLAC::FLAC` `IMPORTED` target. Alongside this change I had to edit SFMLConfigDependencies.cmake to stop using our old solution for handling `INTERFACE` library dependencies. The new solution is simpler and more canonical.

`Freetype::Freetype` was about the same. In this case I wanted to use version of FindFreetype.cmake that ships with CMake but that version is broken when using macOS Frameworks so I had to instead edit our existing find module for Freetype.

`OpenAL::OpenAL` was messy. FindOpenAL.cmake already ships in CMake and that's what we were using but the problem with that is that it doesn't provide the `OpenAL::OpenAL` target until CMake 3.25 and we're only requiring CMake 3.22. To fix this I copied the source code for this find module out of the CMake repo and made some minor tweaks to make it work for our purposes.

Vorbis and Ogg weren't difficult per se but a bit tedious since they amount to 4 total `IMPORTED` targets. The target names are lifted from the [reference implementation of Vorbis](https://github.com/xiph/vorbis). While I was at it I allowed all these targets to be found on iOS as well because it appears to be a harmless change and lets me make the find module simpler.

---

In the end, all uses of `sfml_find_package` and `sfml_bind_dependency` were removed so I was able to delete them. We will never have to use either ever again and can instead focus on using CMake's official machinery for handling 3rd party dependencies.

All of these changes in total will make SFML easier to package and makes SFML more suited to adopt a package manager itself. I'm excited I was able to complete all this cleanup work without disrupting the user workflow whatsoever.


ChrisThrasher:
I want to port the `OpenAL::OpenAL` changes back to SFML 2 so that users of the Audio module have an easier time automatically copying that DLL into their builds. Remember that there are 2 copies of openal32.dll so not only do users have to copy it out of the source tree, they also have to make sure they copy the right version which is yet another thing to go wrong.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:
I still need someone to confirm that this fixes the issue with openal32.dll not being copied. I hope that's fixed now but I can't confirm it.

Also I need to change FindVorbis.cmake back to its old state where certain libraries aren't found on iOS. That's causing an iOS job to fail.

ChrisThrasher:

eXpl0it3r:
One thing that I wonder is, we're editing Find* modules that have their own copyright and license on it, do we somehow need to highlight our changes and/or did the original authors (e.g. Kitware) update their modules in the last many years?

ChrisThrasher:
> One thing that I wonder is, we're editing Find* modules that have their own copyright and license on it, do we somehow need to highlight our changes and/or did the original authors (e.g. Kitware) update their modules in the last many years?

That's a good point. [CMake is permissively licensed](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Licenses/README.rst?ref_type=heads) so I'm inclined to believe that copying the file and making tweaks while retaining the copyright is the right move and doesn't affect SFML's own license.

ChrisThrasher:
> I still need someone to confirm that this fixes the issue with openal32.dll not being copied. I hope that's fixed now but I can't confirm it.

https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:TARGET_RUNTIME_DLLS

image

I can confirm that converting OpenAL from an `INTERFACE` target to an `IMPORTED` target is getting us closer to it being automatically copied so this PR does make some progress towards that end. However we still have to meet some more requirements as outlined above. Currently `OpenAL::OpenAL`'s library type is `UNKNOWN` upon import so CMake can't be certain whether or not it's a DLL. Also it's `IMPORTED_LOCATION` points to openal32.lib, not openal32.dll. Not sure how to find the .dll file instead.

ChrisThrasher:

ChrisThrasher:

eXpl0it3r (approved):
Nice to remove some of that custom macro! 🥳

PR #2667, opened 2023-08-27
Add tests for Network module types

ChrisThrasher:
## Description

This PR adds some unit tests for `sf::TcpListener`, `sf::TcpSocket`, `sf::UdpSocket`, `sf::Http`, and `sf::SocketSelector`. The coverage of each given type varies since some types are fundamentally easier to test than others. Regardless, this still increases code coverage for all five types and lays more groundwork for adding future tests if we figure out ways to further test these network types.


binary1248 (approved):

include/SFML/Network/Http.hpp
```diff
@@ -338,7 +338,7 @@ class SFML_NETWORK_API Http
     /// \brief Default constructor
     ///
     ////////////////////////////////////////////////////////////
-    Http();
+    Http() = default;
```
eXpl0it3r, line 341: When do we `= default` in the header and when in the source file?

ChrisThrasher: We put it in the header as much as possible unless we're using opaque pointers in which case we're forced to put it in the .cpp file.

eXpl0it3r (approved):

PR #2674, opened 2023-09-01
Add tests for `sf::RenderTarget` and `sf::RenderTexture`

ChrisThrasher:
## Description

We're close to 50 test cases!
Screenshot 2023-09-01 at 9 58 13 PM

I could use a hand with the `sf::RenderTarget::mapPixelToCoords` and `sf::RenderTarget::mapCoordsToPixel` if anyone has a better idea for what inputs are the most meaningful for testing. I just pulled out some random numbers for the sake of covering those code paths but I don't understand these functions well enough to test them any better than that.

These two classes should be the last two Graphics module classes without tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
           config: { name: OpenGL ES, flags: -DSFML_USE_MESA3D=TRUE -DBUILD_SHARED_LIBS=TRUE -DSFML_OPENGL_ES=ON }
           type: { name: Debug, flags: -DCMAKE_BUILD_TYPE=Debug -DSFML_ENABLE_COVERAGE=TRUE }
         - platform: { name: Windows MinGW, os: windows-2022 }
-          config: { name: Static Standard Libraries, flags: -GNinja -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DSFML_USE_STATIC_STD_LIBS=TRUE }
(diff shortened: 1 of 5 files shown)


ChrisThrasher:
Screenshot 2023-09-01 at 11 19 54 PM

nice

binary1248 (approved):

eXpl0it3r:
> I could use a hand with the `sf::RenderTarget::mapPixelToCoords` and `sf::RenderTarget::mapCoordsToPixel` if anyone has a better idea for what inputs are the most meaningful for testing. I just pulled out some random numbers for the sake of covering those code paths but I don't understand these functions well enough to test them any better than that.

As the point of these function is to convert between screen space and world/render space, it might make sense to move or otherwise transform the view, so that there isn't a 1:1 mapping.

ChrisThrasher:
> As the point of these function is to convert between screen space and world/render space, it might make sense to move or otherwise transform the view, so that there isn't a 1:1 mapping. For example, if you move the view by 100px, then the top left (0, 0) corner of the render target isn't (0, 0) in the world space anymore, but it's (100, 0).

Good idea. I added an offset to the views to further test more of the internal logic.

ChrisThrasher:
I added a few more `RenderTexture` tests after seeing them missing from the codecov report. While I was in there I also added tests for `sf::Texture::generateMipmap` which is used by `sf::RenderTexture::generateMipmap`.

eXpl0it3r (approved):

PR #2675, opened 2023-09-02
Add tests for `sf::Keyboard`

ChrisThrasher:
## Description

50 test cases! 🥳 (pending #2674)

These are pretty straightforward as to cover the basic use cases of these APIs. In theory we could go all-out and exhaustively cover all keys and scan codes would be neat. Perhaps an eager new contributor can take that task up :)

### DIFF
--- a/test/Window/Keyboard.test.cpp
+++ b/test/Window/Keyboard.test.cpp
@@ -1,0 +1,61 @@
+#include <SFML/Window/Keyboard.hpp>
+
+// Other 1st party headers
+#include <SFML/System/String.hpp>
+
+#include <catch2/catch_test_macros.hpp>
+
+#include <WindowUtil.hpp>
+
+// We're limited on what can be tested. Without control over the hardware and the
+// configuration of the operating system, certain things cannot be tested. In
+// general, the mapping between keys and scancodes is a user configuration. Our
+// tests cannot assume any particular configuration.
+//
+// Regardless this test case represents a best faith effort to cover some of this
+// code in a way that is hopefully not prone to fail on different machines.
+
+TEST_CASE("[Window] sf::Keyboard", runDisplayTests())
+{
+    SECTION("isKeyPressed(Key)")
+    {
+        CHECK(!sf::Keyboard::isKeyPressed(sf::Keyboard::W));
+        CHECK(!sf::Keyboard::isKeyPressed(sf::Keyboard::A));
+        CHECK(!sf::Keyboard::isKeyPressed(sf::Keyboard::S));
+        CHECK(!sf::Keyboard::isKeyPressed(sf::Keyboard::D));
+    }
+
+    SECTION("isKeyPressed(Scancode)")
+    {
+        CHECK(!sf::Keyboard::isKeyPressed(sf::Keyboard::Scan::W));
+        CHECK(!sf::Keyboard::isKeyPressed(sf::Keyboard::Scan::A));
+        CHECK(!sf::Keyboard::isKeyPressed(sf::Keyboard::Scan::S));
+        CHECK(!sf::Keyboard::isKeyPressed(sf::Keyboard::Scan::D));
+    }
+
+    SECTION("localize(Scancode)")
+    {
+        CHECK(sf::Keyboard::localize(sf::Keyboard::Scan::Space) == sf::Keyboard::Space);
+    }
+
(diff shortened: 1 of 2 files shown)


test/Window/Keyboard.test.cpp
```diff
@@ -0,0 +1,66 @@
...
+    {
+        CHECK(!sf::Keyboard::isKeyPressed(sf::Keyboard::Scan::W));
+        CHECK(!sf::Keyboard::isKeyPressed(sf::Keyboard::Scan::A));
+        CHECK(!sf::Keyboard::isKeyPressed(sf::Keyboard::Scan::S));
+        CHECK(!sf::Keyboard::isKeyPressed(sf::Keyboard::Scan::D));
+    }
+
+    SECTION("localize(Scancode)")
+    {
+        CHECK(sf::Keyboard::localize(sf::Keyboard::Scan::W) == sf::Keyboard::W);
```
kimci86, line 32: This kind of check will fail with some keyboard layouts that are not QWERTY.

ChrisThrasher: Good point. I suppose we can't reliably test this then.

ChrisThrasher: Are there any keys and scans that always map to each other on all keyboard layouts?

eXpl0it3r: Keys like Esc, Space, Enter should match in most cases, but in theory layouts can map to anything else 😄

ChrisThrasher: I'm going to use the spacebar as a very conservative choice so that we can continue to have non-zero coverage of these functions. If that's a problem and we're forced to remove these tests then so be it.

test/Window/Keyboard.test.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+    SECTION("delocalize(Key)")
+    {
+        CHECK(sf::Keyboard::delocalize(sf::Keyboard::Space) == sf::Keyboard::Scan::Space);
+    }
+
+    SECTION("getDescription(Scancode)")
+    {
+        CHECK(std::toupper(static_cast<int>(sf::Keyboard::getDescription(sf::Keyboard::Scan::W)[0])) == 'W');
+        CHECK(std::toupper(static_cast<int>(sf::Keyboard::getDescription(sf::Keyboard::Scan::A)[0])) == 'A');
```
kimci86, line 43: This fails with AZERTY keyboard.

ChrisThrasher: I also switched to using the spacebar here

test/Window/Keyboard.test.cpp
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    SECTION("delocalize(Key)")
+    {
+        CHECK(sf::Keyboard::delocalize(sf::Keyboard::Space) == sf::Keyboard::Scan::Space);
+    }
+
+    SECTION("getDescription(Scancode)")
+    {
+        CHECK(sf::Keyboard::getDescription(sf::Keyboard::Scan::Space) == "Space");
```
kimci86, line 40: Descriptions are translated in user language at least on Windows. This test would fail then.

ChrisThrasher: Can I at least keep the function key tests or are those prone to fail as well?

kimci86: I guess function keys are a good choice for this test. I am not aware of a setup that would make the test fail with function keys.

test/Window/Keyboard.test.cpp
```diff
@@ -0,0 +1,61 @@
...
+// Other 1st party headers
+#include <SFML/System/String.hpp>
+
+#include <catch2/catch_test_macros.hpp>
+
+#include <WindowUtil.hpp>
+
+// We're limited on what can be tested. For example the description for certain
+// scancodes is language-specific and the mapping between keys and scancodes is
+// hardware-specific. Without control over the hardware and the configuration of
```
kimci86, line 12: To be more specific, descriptions depend not only on language but also on OS.
I would say that mapping between keys and scancodes is more generally a user configuration rather than a hardware setting. Users probably choose a configuration that matches their hardware but not necessarily.

kimci86 (approved):

eXpl0it3r (approved):

PR #2677, opened 2023-09-05
Add tests for `sf::Clipboard`

ChrisThrasher:
## Description

Not too much to see here. I think these tests should do the trick just fine.


kimci86 (approved):
As we discussed directly, having an open window with an event loop is the documented way to have setString working. Writing a test doing that would be tedious compared to the simple workaround used here, so I guess it can justify this little hack.

eXpl0it3r (approved):
I'm surprised it works everywhere 😄

Issue #2575, opened 2023-06-14
Build warning (error) when an old version of freetype2 is installed (on Linux)


mojca:
Thanks for raising your issue here! 🙂

Before you submit the issue however, we'd like you to consider the follow points.

* We like to use the issue tracker for confirmed issues.
* If you're stuck with SFML, please use [the forum](https://en.sfml-dev.org/forums/index.php#c3) to get help.

----

## Subject of the issue

I'm trying to build some package for Python on `manylinux2014` (which is basically CentOS 7 with some bells and whistles) for an increased backward compatibility.

The software authors decided to go bold and added `-Werror -Wall -Wextra` (as well as some extra flags) which makes the build break with

```
/opt/rh/devtoolset-10/root/usr/bin/g++ -DSFML_STATIC -DSTBI_FAILURE_USERMSG -I/path/to/build/__external/src/sfml/include -I/path/to/build/__external/src/sfml/src -isystem /path/to/build/__external/src/sfml/extlibs/headers/stb_image -isystem /path/to/build/__external/src/sfml/extlibs/headers/glad/include -isystem /usr/include/freetype2 -pthread -Wall -Wextra -Wno-comment -Wno-unused-parameter -Wno-missing-field-initializers -Wno-unknown-pragmas -Wno-parentheses -Wno-misleading-indentation -Wno-unused-variable -Wno-switch -Wno-maybe-uninitialized -Wno-psabi  -Werror=return-type -fexec-charset=UTF-8 -finput-charset=UTF-8 -O3 -DNDEBUG -std=gnu++17 -fPIC -fvisibility=hidden -Werror  -Wall -Wextra -Wshadow -Wnon-virtual-dtor -Wcast-align -Wunused -Woverloaded-virtual -Wconversion -Wsign-conversion -Wdouble-promotion -Wformat=2 -Wnull-dereference -Wold-style-cast -Wpedantic -Wmisleading-indentation -Wduplicated-cond -Wlogical-op -Wduplicated-branches -MD -MT _deps/sfml-build/src/SFML/Graphics/CMakeFiles/sfml-graphics.dir/Font.cpp.o -MF _deps/sfml-build/src/SFML/Graphics/CMakeFiles/sfml-graphics.dir/Font.cpp.o.d -o _deps/sfml-build/src/SFML/Graphics/CMakeFiles/sfml-graphics.dir/Font.cpp.o -c /path/to/build/__external/src/sfml/src/SFML/Graphics/Font.cpp
[...]

PR #2678, opened 2023-09-07
Disable `SFML_WARNINGS_AS_ERRORS` by default

ChrisThrasher:
## Description

This is not the best default value because it imposes additional requirements on user builds that are not strictly necessary. This has caused complaints in the past as people encounter build failures that are merely due to warnings and not hard compiler errors. Changing this default value makes it more likely that someone trying to use SFML can use it without issue.

Here are some issues caused by us enabling `-Werror`/`/WX` by default:

 * https://github.com/conan-io/conan-center-index/pull/18069
 * https://github.com/microsoft/vcpkg/pull/32944
 * https://github.com/SFML/SFML/issues/2100
 * https://github.com/SFML/SFML/issues/2575
 * https://github.com/SFML/SFML/issues/2217
 * https://github.com/SFML/SFML/issues/2046
 * https://github.com/SFML/SFML/issues/1998 

None of these warnings were revealing true issues with the code. If we did fix any of these warnings it was done as a service to users so that they're not forced to see the warnings. It's exceptional that issues or PRs about warnings are catching legit bugs. Our CI pipeline with its static analysis already does a great job catching such things before users can detect them.

It's critical that developers and CI continue to use `-Werror`/`/WX` and luckily it's trivial to keep doing that thanks to us now using CMake presets. We stand to benefit by not requiring users build SFML in this overly strict manner. 

---

We should consider back porting this change to SFML 2 as well since it's easy enough to do. It only requires changing 2 lines of code.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -9,7 +9,8 @@
         "CMAKE_CXX_EXTENSIONS": "OFF",
         "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
         "SFML_BUILD_EXAMPLES": "ON",
-        "SFML_BUILD_TEST_SUITE": "ON"
+        "SFML_BUILD_TEST_SUITE": "ON",
+        "SFML_WARNINGS_AS_ERRORS": "ON"
       }
     }
   ]
--- a/cmake/CompilerWarnings.cmake
+++ b/cmake/CompilerWarnings.cmake
@@ -4,7 +4,7 @@
 
 # Helper function to enable compiler warnings for a specific target
 function(set_target_warnings target)
-    option(SFML_WARNINGS_AS_ERRORS "Treat compiler warnings as errors" TRUE)
+    option(SFML_WARNINGS_AS_ERRORS "Treat compiler warnings as errors" FALSE)
 
     if(SFML_COMPILER_MSVC)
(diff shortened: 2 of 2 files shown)


eXpl0it3r:
Any reason to not directly target 2.6.x and have flow to master with a back-merge?

ChrisThrasher:
I suppose I'd just start here since changing defaults in v3 is generally less controversial. If you like this change then I'll target 2.6.x and then we can back merge it later. It would be nice to get this change out before 2.6.1 as to minimize how many complaints I get from package managers.

eXpl0it3r (approved):
I honestly don't really mind either way. Forcing a failure up warnings, can help us detect potential issues earlier at the cost of user complaints. We can also keep the complaints low and hope our massive CI will detect new things early enough.

I'd directly go with the change on 2.6.x, but up to you.

ChrisThrasher:
The changes are pretty different between master and 2.6.x so I'll stick to 2 separate PRs. For example I have to edit CI in 2.6.x to add this option but in master that's encoded the preset. Plus the options have different names in the two branches.

ChrisThrasher:
https://github.com/SFML/SFML/issues/2744

Not that we needed even more convincing but here's another issue caused by warnings-as-errors enabled by default.

PR #2684, opened 2023-09-13
Add clang-tidy `modernize-return-braced-init-list` check

ChrisThrasher:
## Description

Yet another modern C++ feature we can be taking advantage of. The fact that clang-tidy can automatically detect this makes it easier to justify its addition. It's not something we'll have to keep in mind during code reviews.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -29,7 +29,6 @@ Checks: >
   -misc-non-private-member-variables-in-classes,
   -modernize-avoid-c-arrays,
   -modernize-macro-to-enum,
-  -modernize-return-braced-init-list,
   -modernize-use-nodiscard,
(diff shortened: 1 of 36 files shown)


eXpl0it3r (approved):
As an AAA-style (almost always auto) advocate this seems like a natural "extension". So might be good to also get a view from someone else on the team, maybe @Bromeon?

binary1248 (approved):

Bromeon:
It's not the style I'd personally use -- I would often have to check the return type, especially in longer functions. Could also have to do with languages I use daily (Kotlin, Rust) that are more explicit in this regard.

So, I'm not going to bikeshed it if everyone else finds it more readable 😊

PR #2685, opened 2023-09-15
Restructured and optimized GitHub actions workflow

binary1248:
- Make use of Ninja generator by default for all configurations
- Deduplicated CMake flag specification by setting them based on the full name of the job being run
- Cache data that is repeatedly downloaded every time a job is run
- Use ccache to cache compiler output
- Parse compiler warnings/messages and create GitHub annotations that are displayed inline when viewing the diff

Workflow run time should be reduced by quite a bit if the caches are hit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ concurrency:
 env:
   DISPLAY: ":99" # Display number to use for the X server
   GALLIUM_DRIVER: llvmpipe # Use Mesa 3D software OpenGL renderer
+  CCACHE_NOHASHDIR: true # Tell ccache not to care about embedded directory information
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -19,87 +20,171 @@ jobs:
...
+        elif [[ "${{ matrix.platform.name }}" == *"Intel"* ]]; then
+          echo "CC=icx" >> $GITHUB_ENV
+          echo "CXX=icpx" >> $GITHUB_ENV
+        fi
+
+    - name: Get Latest CMake and Ninja
+      uses: lukka/get-cmake@latest
+      with:
+        cmakeVersion: latest
+        ninjaVersion: latest
```
eXpl0it3r, line 112: Question for CMake is, whether we should go with the latest at all times, or whether we should actually stick to the minimum required version.

binary1248: `CMAKE_MSVC_DEBUG_INFORMATION_FORMAT` was only introduced in CMake 3.25 and isn't really relevant for most users who don't intend to use `ccache`. It speeds up workflows by quite a bit but wouldn't be a reason for me to bump up the minimum required version. Since CMake configuration changes are very infrequent and mostly very conservative I don't think it is really necessary to test the minimum version requirement as part of CI. If we see CMake changes in a patch that are concerning we would already check compatibility there.

eXpl0it3r (approved):
The speed up in build time is quite obvious (from 5-6min down to 1-2min).

I think @ChrisThrasher wants to test each change individually, so we can  keep it open as reference for that.

ChrisThrasher:
I have mixed feelings about this PR. The CI speedups are fantastic but I don't love what is required to get there. This is a pretty radical change from how we were previously handling CI. I'd prefer we do more small PRs like #2686 where we can measure the impact of a given change and weigh that against the additional implementation complexity required to achieve it.

binary1248:
Going the easy route and letting CMake take care of toolset and architecture selection like was done before doesn't let you do much in terms of optimization. There is no middle ground between "do everything for me" and "do everything myself". Besides, I don't know about others but I found the excessive duplication of command line arguments for common configurations like building shared libs and using Mesa 3D really unnecessary.

binary1248:
All changes in this PR have been integrated into master pending #2765.

PR #2692, opened 2023-09-18
Use more `sf::Vector` operations

ChrisThrasher:
## Description

When dealing with calculations that are fundamentally two dimensional it's helpful to use a 2D data type as much as possible rather than decomposing the calculations into x and y components. The more we use vector operations the better chance we have of easiliy reasoning about what the code is doing.

### DIFF
--- a/src/SFML/Graphics/RenderTarget.cpp
+++ b/src/SFML/Graphics/RenderTarget.cpp
@@ -211,10 +211,9 @@ Vector2f RenderTarget::mapPixelToCoords(const Vector2i& point) const
 Vector2f RenderTarget::mapPixelToCoords(const Vector2i& point, const View& view) const
 {
     // First, convert from viewport coordinates to homogeneous coordinates
-    Vector2f        normalized;
-    const FloatRect viewport = FloatRect(getViewport(view));
-    normalized.x             = -1.f + 2.f * (static_cast<float>(point.x) - viewport.left) / viewport.width;
-    normalized.y             = 1.f - 2.f * (static_cast<float>(point.y) - viewport.top) / viewport.height;
+    const FloatRect viewport   = FloatRect(getViewport(view));
+    const Vector2f  normalized = Vector2f(-1, 1) +
+                                Vector2f(2, -2).cwiseMul(Vector2f(point) - viewport.getPosition()).cwiseDiv(viewport.getSize());
 
     // Then transform by the inverse of the view matrix
     return view.getInverseTransform().transformPoint(normalized);
@@ -235,12 +234,9 @@ Vector2i RenderTarget::mapCoordsToPixel(const Vector2f& point, const View& view)
     const Vector2f normalized = view.getTransform().transformPoint(point);
 
     // Then convert to viewport coordinates
-    Vector2i        pixel;
     const FloatRect viewport = FloatRect(getViewport(view));
-    pixel.x                  = static_cast<int>((normalized.x + 1.f) / 2.f * viewport.width + viewport.left);
-    pixel.y                  = static_cast<int>((-normalized.y + 1.f) / 2.f * viewport.height + viewport.top);
-
-    return pixel;
+    return Vector2i((normalized.cwiseMul({1, -1}) + sf::Vector2f(1, 1)).cwiseDiv({2, 2}).cwiseMul(viewport.getSize()) +
+                    viewport.getPosition());
 }
 
 
--- a/src/SFML/Graphics/Shape.cpp
+++ b/src/SFML/Graphics/Shape.cpp
@@ -221,8 +221,7 @@ void Shape::update()
     m_insideBounds = m_vertices.getBounds();
 
     // Compute the center and make it the first vertex
-    m_vertices[0].position.x = m_insideBounds.left + m_insideBounds.width / 2;
-    m_vertices[0].position.y = m_insideBounds.top + m_insideBounds.height / 2;
+    m_vertices[0].position = m_insideBounds.getPosition() + m_insideBounds.getSize() / 2.f;
 
     // Color
     updateFillColors();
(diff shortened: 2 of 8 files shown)


src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -211,10 +211,9 @@ Vector2f RenderTarget::mapPixelToCoords(const Vector2i& point) const
 Vector2f RenderTarget::mapPixelToCoords(const Vector2i& point, const View& view) const
 {
     // First, convert from viewport coordinates to homogeneous coordinates
-    Vector2f        normalized;
     const FloatRect viewport = FloatRect(getViewport(view));
-    normalized.x             = -1.f + 2.f * (static_cast<float>(point.x) - viewport.left) / viewport.width;
-    normalized.y             = 1.f - 2.f * (static_cast<float>(point.y) - viewport.top) / viewport.height;
+    const Vector2f  normalized(-1.f + 2.f * (static_cast<float>(point.x) - viewport.left) / viewport.width,
+                              1.f - 2.f * (static_cast<float>(point.y) - viewport.top) / viewport.height);
```
Bromeon, line 216: Wanted to refactor this to _not_ use individual components, but the different signs for X and Y made it a bit more complex 😕 so not sure if it's truly more readable:
```cpp
Vector2f RenderTarget::mapPixelToCoords(const Vector2i& pixel, const View& view) const {
    // These could also be inlined, at the cost of more complex expression
    const Vector2f viewPos = viewPort.getPosition();
    const Vector2f viewSize = viewPort.getSize();

    const Vector2f point(pixel);
    const Vector2f normalized = Vector2f(-1.f, 1.f) 
        + Vector2f(2.f, -2.f).cwiseMul(point - viewPos).cwiseDiv(viewSize);
    ...
```
If we had proper operators, this would be much less ugly:
```cpp
    const Vector2f normalized = Vector2f(-1.f, 1.f) + Vector2f(2.f, -2.f) * (point - viewPos) / viewSize;
```

src/SFML/Graphics/Shape.cpp
```diff
@@ -221,8 +221,8 @@ void Shape::update()
     m_insideBounds = m_vertices.getBounds();
 
     // Compute the center and make it the first vertex
-    m_vertices[0].position.x = m_insideBounds.left + m_insideBounds.width / 2;
-    m_vertices[0].position.y = m_insideBounds.top + m_insideBounds.height / 2;
+    m_vertices[0].position = Vector2f(m_insideBounds.left, m_insideBounds.top) +
+                             Vector2f(m_insideBounds.width, m_insideBounds.height) / 2.f;
```
Bromeon, line 225: We should really think about `Rect::getEnd()` and `Rect::getCenter()` methods.

This would be:
```suggestion
    m_vertices[0].position = m_insideBounds.getCenter();
```

But at least, we can already do now -- which is of course much less expressive:

```suggestion
    m_vertices[0].position = m_insideBounds.getPosition() + m_insideBounds.getSize() / 2.f;
```

Bromeon (commented):
Thank you!

Made two examples where vector-based operations might be an alternative. One is much more controversial than the other 😉 

Also, we should really reconsider some utility methods to get/set end and center of rectangles. They cost nothing and their absence makes code less readable for no good reason.

src/SFML/Graphics/Shape.cpp
ChrisThrasher: I agree that having named APIs for these other parts of a rectangle would be useful. I support their addition.

src/SFML/Graphics/RenderTarget.cpp
ChrisThrasher: > Wanted to refactor this to not use individual components, but the different signs for X and Y made it a bit more complex 😕 so not sure if it's truly more readable:

I think this is an improvement. I have an easier time reasoning about these operations in a 2D space. I'd be happy to add this to the PR. Plus it means getting rid of some `static_cast`s :)

ChrisThrasher:

src/SFML/Graphics/RenderTarget.cpp
eXpl0it3r: I assume this is still open?

eXpl0it3r (changes requested):
See the open comment thread

src/SFML/Graphics/RenderTarget.cpp
ChrisThrasher: I just added that to the PR.

eXpl0it3r (approved):

src/SFML/Graphics/RenderTarget.cpp
```diff
@@ -235,12 +234,9 @@ Vector2i RenderTarget::mapCoordsToPixel(const Vector2f& point, const View& view)
...
 
     // Then convert to viewport coordinates
-    Vector2i        pixel;
     const FloatRect viewport = FloatRect(getViewport(view));
-    pixel.x                  = static_cast<int>((normalized.x + 1.f) / 2.f * viewport.width + viewport.left);
-    pixel.y                  = static_cast<int>((-normalized.y + 1.f) / 2.f * viewport.height + viewport.top);
-
-    return pixel;
+    return Vector2i(Vector2f((normalized.x + 1.f) / 2.f * viewport.width + viewport.left,
+                             (-normalized.y + 1.f) / 2.f * viewport.height + viewport.top));
```
ChrisThrasher, line 239: I want to change this to match the new version we're using in `mapPixelToCoords`

src/SFML/Graphics/Shape.cpp
```diff
@@ -270,14 +270,14 @@ void Shape::updateTexCoords()
...
-        m_vertices[i].texCoords.x = convertedTextureRect.left + convertedTextureRect.width * xratio;
-        m_vertices[i].texCoords.y = convertedTextureRect.top + convertedTextureRect.height * yratio;
+        const float xratio      = m_insideBounds.width > 0
+                                      ? (m_vertices[i].position.x - m_insideBounds.left) / m_insideBounds.width
+                                      : 0;
+        const float yratio      = m_insideBounds.height > 0
+                                      ? (m_vertices[i].position.y - m_insideBounds.top) / m_insideBounds.height
+                                      : 0;
+        m_vertices[i].texCoords = {convertedTextureRect.left + convertedTextureRect.width * xratio,
+                                   convertedTextureRect.top + convertedTextureRect.height * yratio};
```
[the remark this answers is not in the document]
ChrisThrasher: Really satisfying simplifications 👍🏻
I now see what is actually going on. Goes to show how these non-vector based equations are a lot harder to read than the vector-based counterpart. Thanks!

PR #2695, opened 2023-09-19
Add more tests for `sf::Window`

ChrisThrasher:
## Description

`sf::Window` has some overloads that `sf::WindowBase` does not so it's worth ensuring we explicitly test them.

### DIFF
--- a/test/Window/Window.test.cpp
+++ b/test/Window/Window.test.cpp
@@ -22,10 +22,90 @@ TEST_CASE("[Window] sf::Window", runDisplayTests())
 
     SECTION("Construction")
     {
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
> Id like to test the overloads that use `WindowHandle`s. I'm not sure how to do that though without creating a window, getting its handle, then using that handle to create a new window. Not sure if that's a valid way of testing those overloads though.

I think to properly test this, as in the way as the API has been intended for, you'd have to write platform specific code and use the handle.

ChrisThrasher:
> you'd have to write platform specific code and use the handle.

I figured this was the case. This is out of scope at the moment although it's theoretically possible we could write such tests in the future.

binary1248 (approved):

PR #2696, opened 2023-09-19
Remove trailing whitespace

ChrisThrasher:
## Description

Some of these files either aren't processed by clang-format or the trailing whitespace existed within blocks of code where we disabled clang-format.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,7 +127,7 @@ jobs:
     - name: Build
       shell: bash
       run: cmake --build build --config ${{ matrix.type.name == 'Debug' && 'Debug' || 'Release' }} --target install
-      
+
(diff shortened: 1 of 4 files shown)


Bromeon (approved):
Looks good!

Does our [`.editorconfig`](https://github.com/SFML/SFML/blob/master/.editorconfig) not catch those?

ChrisThrasher:
Perhaps the problem is that these files haven't been modified in a while? I'm not really sure how `.editorconfig` is supposed to be used though. I don't pay it any attention nor have I worked in other codebases that use it.

Bromeon:
I've only used it via IDE plugins, but you might be right that not touching the files in a while might not detect any non-conformities.

binary1248 (approved):

PR #2698, opened 2023-09-20
Use one InputImpl.hpp header for all implementations

ChrisThrasher:
## Description

There were 6 copies of InputImpl.hpp for Windows, macOS, Linux, Linux DRM, iOS, and Android. All but Windows and Linux DRM were identical so there's already an easy opportunity for consolidation.

The Windows header included some additional private static functions and data. An equivalent way of expressing this is via more data and functions instead the anonymous namespace of Win32/InputImpl.cpp so that's what I did. To fix some issues with functions being used before they're defined I moved number of functions into the anonymous namespace of that file which resulted in adding `sf::` in lots of places.

InputImplUDev.hpp used by the DRM backend was trickier because includes WindowImplDRM as a friend. This creates a weird dependency graph where a distant class can call private functions that mutate private state. I kept the definitions of those functions inside DRM/InputImpl.cpp while moving their declaration to WindowImplDRM.cpp which is the only file where they're used. This is a bit odd but it avoids using the preprocessor and ensures that only the file that needs these functions can call them.

In the end I was able to delete all 6 separate copies of InputImpl.hpp and consolidate them into src/SFML/Window/InputImpl.hpp **saving nearly 1,000 lines of code.**

Because `sf::priv::InputImpl` is merely a set of static functions I converted this class to a simple namespace. The use of a namespace is what ultimately allowed me to untangle the DRM functions that were not present in other implementations


Bromeon (commented):
Very nice! 👍 

This is a pure refactoring without semantic/doc changes, right?

The `namespace { }` in .cpp files means that you need to qualify `sf::` everywhere, which feels a bit like "outside the library" but I guess that's the convention that SFML uses. In case we ever want to address that, there are a few options:
* Specific `using` directives, like `using sf::Keyboard;`
* Use named namespaces (this changes linkage and may accidentally make symbols available)
* Anonymous namespace inside named namespace, i.e. `namespace sf { namespace { ... } }`

If we change anything, might need revisit of other similar places in SFML (out of scope for this PR). But I'm also OK with keeping it.

ChrisThrasher:
> This is a pure refactoring without semantic/doc changes, right?

Correct. If it changes semantics then that's certainly not intentional.

> The namespace { } in .cpp files means that you need to qualify sf:: everywhere

Adding `sf::` in so many places was annoying but that's just the reality of us putting as much private code inside anonymous namespaces as possible.

Bromeon:
> but that's just the reality of us putting as much private code inside anonymous namespaces as possible.

As mentioned, there are two ways around it, while still keeping anonymous namespaces 🙂
Not sure if we see that as something worthwhile though.

ChrisThrasher:
> As mentioned, there are two ways around it, while still keeping anonymous namespaces 🙂 Not sure if we see that as something worthwhile though.

What I'm saying is that I'm fine keeping `sf::` inside our anonymous namespaces. It's not something I'm motivated to change.

binary1248 (approved):

PR #2699, opened 2023-09-20
Fix incompatibilities with newer clang-tidy releases

ChrisThrasher:
## Description

LLVM 17 was just released so I decided to build LLVM from `HEAD` and run clang-tidy to see what it finds so that we can proactively address new warnings. I ignored a few new warnings that are either opinions we do not endorse or problems that are of dubious value to fix.

`performance-enum-size` was added in clang-tidy-18 which detects instances of enumerations being larger than necessary. Remember that enums are 4 bytes by default yet we seldom need all that space. In most cases we only need 1 byte and thus can use a `std::int8_t` or `std::uint8_t` to save space. The compiler will detect if any enum values cannot be presented by the underlying storage so there's no risk of us accidentally picking an enum storage class that is too small. In a few places a 16 bit integer was required.

We've been using `readability-container-size-empty` for a while but I was updated to detect more instances of this anti-pattern so I fixed that as well.

The two checks I may want to enable later are `bugprone-switch-missing-default-case` and `misc-header-include-cycle`. I haven't yet looked closely into whether they're right for us but they look promising.


ChrisThrasher:
Once we get around to reenabling `brew update` in CI then we'll start to see these errors pop up in all CI jobs since Homebrew is now shipping clang-tidy-17.

Bromeon:
> `performance-enum-size` was added in clang-tidy-18 which detects instances of enumerations being larger than necessary. Remember that enums are 4 bytes by default yet we seldom need all that space. In most cases we only need 1 byte and thus can use a `std::int8_t` or `std::uint8_t` to save space. The compiler will detect if any enum values cannot be presented by the underlying storage so there's no risk of us accidentally picking an enum storage class that is too small. In a few places a 16 bit integer was required.

This is mostly an optimization for size and not speed, correct? It can influence speed when storing lots of adjacent enums, which wouldn't otherwise fit in cache. But iteration might also be slower if the compiler needs to insert bitwise logic to extract 1-byte enums from 4/8-byte words.

Just a bit confused because for a very long time, [the general recommendation](https://stackoverflow.com/questions/3694131/fastest-integer-type-for-common-architectures) was to use `int/long` as they are "native" and most optimized for their architectures. People discouraged use of `unsigned char`/`short`/etc for integers. This StackOverflow thread also contains a very good argument that in theory, `std::uint_fast8_t` is always the fastest integer type ever.

ChrisThrasher:
> This is mostly an optimization for size and not speed, correct?

https://clang.llvm.org/extra/clang-tidy/checks/performance/enum-size.html

According to the clang-tidy docs the main focus is on memory footprint with potential runtime performance gains. I'm taking it on face value that those who wrote this check know what they're talking about. I don't have any empirical evidence to corroborate what the docs say about runtime gains. Based on first principles it seems reasonable to me to not use more memory than we need. My rationale doesn't go much deeper than that.

I won't die on the hill to use this check. It seems reasonable but the potential gains are probably relatively small. It's not the most important part of this PR. The most important aspect is disabling the new checks that are failing. That's making it infeasible to run clang-tidy locally since I'm already using clang-tidy-17 and thus interfering with my ability to develop.

eXpl0it3r:
I think the Core Guidelines recommend to only specify a type for enums when necessary: http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-underlying

ChrisThrasher:
So what's the final call on whether we use this enum size check?

Bromeon:
Aren't C++ compilers already allowed to determine the underlying type autonomously, as long as it's able to represent all values in an enum? In that sense, I would expect an optimizing configuration to select the one which is fastest for the target architecture.

However, memory footprint is also something we can consider important. If you have a large `std::vector<sf::SomeEnum>`, a factor of 4 could make quite a difference. But this might again be depending on compiler/flags/etc.

So, from my point of view, we need to prioritize:
* memory -> use smallest possible type (+ enable clang-tidy flag)
* speed -> use no underlying type or `std::uint_fast8_t`
* don't care -> no underlying type

ChrisThrasher:
`performance-enum-size` isn't even released yet and there is not consensus about it's use so I remove any changes recommended by that check. Even if we did like it, we still can't enforce it in CI since it's a clang-tidy-18 feature and CI doesn't have access to anything newer than clang-tidy-17.

Bromeon (approved):
Looks good, thanks!

PR #2706, opened 2023-09-24
Add tests for `sf::Joystick`

ChrisThrasher:
## Description

See a comment I wrote in Joystick.test.cpp about why we can reliably test `sf::Joystick` even if someone currently has a joystick connected. I confirmed this locally by connecting an Xbox controller and observing the tests passing. If I added a call to `sf::Joystick::update()` I can confirm it fails as one would expect.

While I was at it, I also added some more assertion to catch invalid user inputs to public APIs. This expands on what I did in https://github.com/SFML/SFML/pull/2682. Yay for more checks against UB!

### DIFF
--- a/test/Window/Joystick.test.cpp
+++ b/test/Window/Joystick.test.cpp
@@ -1,0 +1,76 @@
+#include <SFML/Window/Joystick.hpp>
+
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/generators/catch_generators_range.hpp>
+
+TEST_CASE("[Window] sf::Joystick")
+{
+    SECTION("Constants")
+    {
+        STATIC_CHECK(sf::Joystick::Count == 8);
+        STATIC_CHECK(sf::Joystick::ButtonCount == 32);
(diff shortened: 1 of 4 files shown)


test/Window/Joystick.test.cpp
```diff
@@ -0,0 +1,76 @@
...
+        CHECK(identification.name == "No Joystick");
+        CHECK(identification.vendorId == 0);
+        CHECK(identification.productId == 0);
+    }
+
+    // By avoiding calling sf::Joystick::update() we can guarantee that
+    // no joysticks will be detected. This is how we can ensure these
+    // tests are portable and reliable.
+
+    const auto joystick = GENERATE(range(0u, sf::Joystick::Count - 1));
```
ChrisThrasher, line 27: `GENERATE` is a Catch2 macro that, in this case, generates a range of values from 0 to 7 (inclusive) so that we can test various joystick index inputs without having to write `CHECK` 8 times more.

binary1248 (approved):

Issue #246, opened 2012-07-02
Investigate support for complex text layouts


LaurentGomila:
SFML currently only support simple text layouts. It is therefore unable to correctly display, for instance, Arabic which requires right-to-left layout, glyphs joined together, selecting the correct glyph variant according to its position in the word, etc.

This is a complex task and SFML will probably not be able to render all texts 100% correctly. However, FreeType has maybe more options to offer, and other libraries such as HarfBuzz could be used as well.

Bromeon:
Are there currently alternatives to Harfbuzz to realize this feature request? If not, should we close it for the time being?

eXpl0it3r:
The linked PR is already closed and this feature request should stay open, since it's something that we should really find a solution to. Additionally I still don't see an issue with adding Harfbuzz. Sure it's one dependency more, but not being able to properly write in Arabic, Chinese or similar is just bad and makes some of the unicode support rather useless.

eXpl0it3r:
13 years later, SFML finally has complex text layout support! 🎉

PR #2713, opened 2023-09-28
add text alignment to sf::Text

Hapaxia:
Adds ability to align all lines of sf::Text to the left, centre or the right.

It provides a single enum choice for the alignment and simply (horizontally) offsets the lines so they align according to the choice.

I has been tested with empty lines and also with underlines and strikethroughs as well to confirm those lines behave correctly.

Here is an image of it in action:
An image showing the different types of line alignment with different types of decorations

I spoke to a member of the team about if this would be wanted and they said it would be welcome.

I hope it's exactly the correct code style.

----

----

## Description

## Tasks

* [ ] Tested on Linux
* [x] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::Font font;
    if (!font.loadFromFile("arial.ttf"))
        return EXIT_FAILURE;

    sf::Text text(font);
    text.setString("THIS\nIS\nA\nMULTI-LINE\nTEXT");
    text.setLineAlignment(sf::Text::Center);

    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        {
                window.close();
        }

        window.clear();
        window.draw(text);
        window.display();
    }
}
```


ChrisThrasher:
I fixed up some minor style issues to get those out of the way early. Thanks for submitting this! I'll take a closer look when I have some more time later to dig into this. Unfortunately that may be a few days from now.

Hapaxia:
Thanks!

I thought it was about time I got around to doing this. 😃 

Let me know if you would like me to adjust anything.

lapinozz:
This is a cool feature and it seems to fit nicely in sf::Text without too many changes, sweet!
I think we could change it so that instead of having a m_lineOffsets member the line offset update function could return the vector.
Since it doesn't seem to be used/useful outside the geometry update and it's recreated each time anyways

ChrisThrasher:
Looks like clang-tidy caught something suspicious. Check out those CI logs to see if there's anything we need to fix.

https://github.com/SFML/SFML/actions/runs/6343566229/job/17231737846?pr=2713#step:8:76

Hapaxia:
I don't think it likes the fact that I've added something called lineOffset and passed it to the same function that an underlineOffset is passed.
If required, I can change it to something more verbose: horizontalOffset perhaps?


> This is a cool feature and it seems to fit nicely in sf::Text without too many changes, sweet! I think we could change it so that instead of having a m_lineOffsets member the line offset update function could return the vector. Since it doesn't seem to be used/useful outside the geometry update and it's recreated each time anyways
Thanks!
It's also used by findCharacterPos but it does re-generate it there too. I did think that maybe the m_lineOffsets vector could be used elsewhere (even for getting the number of lines) but I didn't want to change too much!
Part of the reason of storing it was to avoid it being constantly created and destroyed when nothing is changing. Rather, it gets resized to the same size (so stays the same) and just updates the values.

Of course, if it's preferred, I can switch it to create the vector on the fly every update; it is pretty small, after all.

lapinozz:
> Thanks! It's also used by findCharacterPos but it does re-generate it there too. I did think that maybe the m_lineOffsets vector could be used elsewhere (even for getting the number of lines) but I didn't want to change too much! Part of the reason of storing it was to avoid it being constantly created and destroyed when nothing is changing. Rather, it gets resized to the same size (so stays the same) and just updates the values.

Oh I didn't notice it was used in that other function. If we want to keep the vector as a member then we could get rid of the local vector in the function and use the member directly. Because right now it always allocates a new vector.

Hapaxia:
Indeed. That temporary vector is used for a different semantic purpose but I believe the member vector could be re-used for that purpose until it's needed without any issues.

ChrisThrasher:
> If required, I can change it to something more verbose: horizontalOffset perhaps?

Addressing this clang-tidy error in some way will be required. Either fix whatever it's complaining about or make a compelling case for why we should disable this error.

Hapaxia:
> > If required, I can change it to something more verbose: horizontalOffset perhaps?
> 
> Addressing this clang-tidy error in some way will be required. Either fix whatever it's complaining about or make a compelling case for why we should disable this error.

Okay, no problem. I'll change it to what I suggested and hope that it's happy with that thing instead!

Hapaxia:
Updated to change that variable name. Hopefully, the clang-tidy checks won't have an issue with that now.

ChrisThrasher:
If you have clang-tidy installed you can build the `tidy` target to run clang-tidy. I'd recommend doing that since it's much faster than waiting for CI to run.

Hapaxia:
I don't, unfortunately, sorry.

I didn't know that. I rarely use command line; I tend to use the Desktop app.

Hapaxia:
It has failed now on formatted. It seems to be suggesting to spread a line over multiple lines. Is that correct? I gather there's maybe a hard maximum limit on line length?

ChrisThrasher:
> It has failed now on formatted. It seems to be suggesting to spread a line over multiple lines. Is that correct? I gather there's maybe a hard maximum limit on line length?

That's because when you made that merge commit you reverted the formatting fixes I made.

Hapaxia:
That's not exactly true. All of your corrections are still present. I made sure to take all of your versions when merging.

ChrisThrasher:
My bad it wasn't related to that merge commit. It was some new code hit the line limit and started wrapping. If you have clang-format-14 installed you can build the `format` target to fix formatting. I went ahead and fixed it myself and pushed a new commit.

Hapaxia:
Thanks. I've also pulled and rebased your change this time as you suggested previously 😃

kimci86:
So this has no impact for a single-line text. What about positioning text relative to the local origin so that it has impact on a single-line text too? It would also make is simple to center or right-align text on a given horizontal line by setting the text position on said horizontal line.

Hapaxia:
It's designed to align with the other lines within the text block only so a single line should not be affected indeed.

To centre around - or align to the right using - the origin, we can simply offset its position by -width/2 or -width respectively.

Whether this is up to the user to do or SFML is a design decision that the team would have to make but it should be possible without too much 'upset' of the code. However, I didn't want to change too much at once just for a single small feature.

eXpl0it3r:

src/SFML/Graphics/Text.cpp
```diff
@@ -540,4 +577,74 @@ void Text::ensureGeometryUpdate() const
...
+        // For regular characters, add the advance offset of the glyph
+        position.x += m_font->getGlyph(curChar, m_characterSize, isBold).advance + letterSpacing;
+    }
+
+    // add final part of the string since last newline as the final line
+    // (this is the entire string if text is single line without newlines)
+    lineWidths.push_back(position.x);
+
+    // update line offsets from widths depending on line alignment
+    m_lineOffsets.resize(lineWidths.size());
```
ChrisThrasher, line 623: Could we use 1 vector here instead of 2? As in, could we remove `lineWidths` and replace that with a more clever use of `m_ lineWidths`?

ChrisThrasher (commented):
I added some unit tests and removed some duplicate code.

ChrisThrasher:

kimci86:
> It's designed to align with the other lines within the text block only so a single line should not be affected indeed.

In my opinion this is a sign that the design is not good. I suggest we make a conscious choice about the wanted behavior and not realize after the fact that it is a weird behavior. What do you all think?

kimci86:
On another topic: What about non-integer positions? Should we round them to get a sharp look?

src/SFML/Graphics/Text.cpp
```diff
@@ -463,7 +498,9 @@ void Text::ensureGeometryUpdate() const
                     break;
                 case U'\n':
                     y += lineSpacing;
-                    x = 0;
+                    ++line;
+                    horizontalOffset = (line < m_lineOffsets.size()) ? m_lineOffsets[line] : 0.f;
```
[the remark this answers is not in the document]
Hapaxia: It shouldn't. I was being super-extra-cautious.

I had the test elsewhere but removed it when I was certain it was never going to exceed it.

It increases the line parameter before using it as an index and I haven't done a test to check if this still works for text where the last line end with \n but I believe the added method should force it to consider that an extra line.

src/SFML/Graphics/Text.cpp
Hapaxia: Yes. It was briefly mentioned early here but wasn't sure if it would be clear to use the same vector for two semantically different things: widths and offsets, even if it is temporary, private and "internal".

Hapaxia:
> > It's designed to align with the other lines within the text block only so a single line should not be affected indeed.
> 
> In my opinion this is a sign that the design is not good (i.e. a function that does nothing in many cases). I suggest we make a conscious choice about the wanted behavior and not realize after the fact that it is a weird behavior. What do you all think?

I, personally, don't think it's weird to expect that telling a text object to align its (multiple) lines to not affect single lines. However, I can also see how this could benefit from automatically moving it around the origin.

Please discuss and I'll get right on it if wanted.

Hapaxia:
> On another topic: What about non-integer positions? Should we round them to get a sharp look? We can see some lines are blurry on the centered blocks in this PR description's screenshot.

This is clearly a simple enough fix as well. I think this might be a good idea indeed.

Would rounding to integer be affected by views? I don't think so (because this is the local co-ordinate system) but I'm not 100% on this.

Hapaxia:
> I rebased on `master` because we recently merged some new `sf::Text` tests. I also added unit tests for this new API and removed some duplicate code. Run `git pull -r` to update your local branch since I had to change the commit history.

I've done exactly that so we should be synchronized. I like the update; it removes duplication of similar code.

Hapaxia:
In addition, does sf::Text support vertical text? It doesn't look like it does (looking at its code) but if it does, this alignment inclusion might need some modification.

Hapaxia:
I've done some slight modifications to address so of the minor concerns raised.

That is, the pointless test for being out of range, the vector being re-used instead of creating a temporary one and internal rounding of the offsets.

However, the decision as whether the alignment affects the position of the visible text object (and its bounds) still needs to be made before I just go ahead and add that.

Hapaxia:
Updated screenshot (identical test code) to show how the rounding affects the text:
An updated image showing the different types of line alignment with different types of decorations including rounding

Hapaxia:
NOTE: the tests now need to check that local bounds match their integer positions, not the fractional positions. Global bounds in the check will also be affected.

Hapaxia:
Can someone please let me know what I need to do to get this ready to be included?

In addition, I'd like to add the suggested idea of aligning to its origin (e.g. aligning to the right would mean that text would be left of its position) and would like confirmation that this is the wanted behaviour.

include/SFML/Graphics/Text.hpp
```diff
@@ -63,6 +63,17 @@ class SFML_GRAPHICS_API Text : public Drawable, public Transformable
         StrikeThrough = 1 << 3  //!< Strike through characters
     };
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Enumeration of the text alignment options
+    ///
+    ////////////////////////////////////////////////////////////
+    enum LineAlignment
```
vittorioromeo, line 70:
```suggestion
    enum class LineAlignment
```

include/SFML/Graphics/Text.hpp
```diff
@@ -415,11 +475,13 @@ class SFML_GRAPHICS_API Text : public Drawable, public Transformable
     Color                 m_fillColor{Color::White};                   //!< Text fill color
     Color                 m_outlineColor{Color::Black};                //!< Text outline color
     float                 m_outlineThickness{0.f};                     //!< Thickness of the text's outline
+    LineAlignment         m_lineAlignment{Left};                       //!< Line alignment for a multi-line text
```
vittorioromeo, line 478:
```suggestion
    LineAlignment         m_lineAlignment{LineAlignment::Left};        //!< Line alignment for a multi-line text
```

src/SFML/Graphics/Text.cpp
```diff
@@ -280,22 +293,29 @@ float Text::getOutlineThickness() const
...
-    float       whitespaceWidth = m_font->getGlyph(U' ', m_characterSize, isBold).advance;
-    const float letterSpacing   = (whitespaceWidth / 3.f) * (m_letterSpacingFactor - 1.f);
-    whitespaceWidth += letterSpacing;
-    const float lineSpacing = m_font->getLineSpacing(m_characterSize) * m_lineSpacingFactor;
+    const bool isBold                                        = m_style & Bold;
+    const auto [whitespaceWidth, letterSpacing, lineSpacing] = getSpacing();
 
     // Compute the position
-    Vector2f      position;
+    Vector2f      position(m_lineOffsets[0], 0.f); // There will always be at least one line
```
vittorioromeo, line 318:
```suggestion
    assert(!m_lineOffsets.empty());
    Vector2f      position(m_lineOffsets[0], 0.f); // There will always be at least one line
```

Let's assert it then :)

src/SFML/Graphics/Text.cpp
```diff
@@ -402,19 +425,18 @@ void Text::ensureGeometryUpdate() const
...
 
     // Precompute the variables needed by the algorithm
-    float       whitespaceWidth = m_font->getGlyph(U' ', m_characterSize, isBold).advance;
-    const float letterSpacing   = (whitespaceWidth / 3.f) * (m_letterSpacingFactor - 1.f);
-    whitespaceWidth += letterSpacing;
-    const float lineSpacing = m_font->getLineSpacing(m_characterSize) * m_lineSpacingFactor;
-    float       x           = 0.f;
-    auto        y           = static_cast<float>(m_characterSize);
+    const auto [whitespaceWidth, letterSpacing, lineSpacing] = getSpacing();
+    float x = m_lineOffsets[0]; // there will always be at least one line
```
vittorioromeo, line 429:
```suggestion
    assert(!m_lineOffsets.empty());
    float x = m_lineOffsets[0]; // there will always be at least one line
```

test/Graphics/Text.test.cpp
```diff
@@ -54,6 +54,7 @@ TEST_CASE("[Graphics] sf::Text", runDisplayTests())
             CHECK(text.getFillColor() == sf::Color::White);
             CHECK(text.getOutlineColor() == sf::Color::Black);
             CHECK(text.getOutlineThickness() == 0);
+            CHECK(text.getLineAlignment() == sf::Text::Left);
```
vittorioromeo, line 57:
```suggestion
            CHECK(text.getLineAlignment() == sf::Text::LineAlignment::Left);
```

test/Graphics/Text.test.cpp
```diff
@@ -71,6 +72,7 @@ TEST_CASE("[Graphics] sf::Text", runDisplayTests())
...
```
vittorioromeo, line 75:
```suggestion
            CHECK(text.getLineAlignment() == sf::Text::LineAlignment::Left);
```

test/Graphics/Text.test.cpp
```diff
@@ -88,6 +90,7 @@ TEST_CASE("[Graphics] sf::Text", runDisplayTests())
...
```
vittorioromeo, line 93:
```suggestion
            CHECK(text.getLineAlignment() == sf::Text::LineAlignment::Left);
```

test/Graphics/Text.test.cpp
```diff
@@ -158,6 +161,19 @@ TEST_CASE("[Graphics] sf::Text", runDisplayTests())
         CHECK(text.getOutlineThickness() == 3.14f);
     }
 
+    SECTION("Set get line alignment")
+    {
+        sf::Text text(font, "QWERTY");
+        text.setPosition({50, 25});
+        CHECK(text.findCharacterPos(0) == sf::Vector2f(50, 25));
+        text.setLineAlignment(sf::Text::Center);
+        CHECK(text.getLineAlignment() == sf::Text::Center);
```
vittorioromeo, line 170:
```suggestion
        text.setLineAlignment(sf::Text::LineAlignment::Center);
        CHECK(text.getLineAlignment() == sf::Text::LineAlignment::Center);
```

test/Graphics/Text.test.cpp
```diff
@@ -158,6 +161,19 @@ TEST_CASE("[Graphics] sf::Text", runDisplayTests())
...
+        CHECK(text.findCharacterPos(0) == sf::Vector2f(-6.5, 25));
+        text.setLineAlignment(sf::Text::Right);
+        CHECK(text.getLineAlignment() == sf::Text::Right);
```
vittorioromeo, line 173:
```suggestion
        text.setLineAlignment(sf::Text::LineAlignment::Right);
        CHECK(text.getLineAlignment() == sf::Text::LineAlignment::Right);
```

test/Graphics/Text.test.cpp
```diff
@@ -199,5 +215,15 @@ TEST_CASE("[Graphics] sf::Text", runDisplayTests())
             CHECK(text.getLocalBounds() == sf::FloatRect({1, 5}, {33, 13}));
             CHECK(text.getGlobalBounds() == Approx(sf::FloatRect({66, 182}, {33, 13})));
         }
+
+        SECTION("Change alignment")
+        {
+            text.setLineAlignment(sf::Text::Center);
```
vittorioromeo, line 221:
```suggestion
            text.setLineAlignment(sf::Text::LineAlignment::Center);
```

test/Graphics/Text.test.cpp
```diff
@@ -199,5 +215,15 @@ TEST_CASE("[Graphics] sf::Text", runDisplayTests())
...
+            CHECK(text.getLocalBounds() == sf::FloatRect({-16.5, 5}, {33, 13}));
+            CHECK(text.getGlobalBounds() == sf::FloatRect({83.5, 205}, {33, 13}));
+            text.setLineAlignment(sf::Text::Right);
```
vittorioromeo, line 224:
```suggestion
            text.setLineAlignment(sf::Text::LineAlignment::Right);
```

vittorioromeo (changes requested):
Minor suggestions

eXpl0it3r:
> Can someone please let me know what I need to do to get this ready to be included?

As it's "just" an addition it has a bit of a lower priority. So it mostly takes time to get around to it.

If you can rebase it and apply @vittorioromeo's suggestions, then we can hopefully get around to it soon.

Hapaxia:
I just meant so that I could make it ready when it's needed.

Implemented @vittorioromeo suggestions. Thanks!
I only used enum as it seems to be SFML's style; I would've used enum class otherwise.

It's the first time I've seen that test file!

src/SFML/Graphics/Text.cpp
```diff
@@ -540,4 +573,83 @@ void Text::ensureGeometryUpdate() const
...
+                break;
+            case LineAlignment::Center:
+                m_lineOffsets[i] = (maxWidth - m_lineOffsets[i]) / 2.f;
+                break;
+            case LineAlignment::Left:
+                m_lineOffsets[i] = 0.f;
+                break;
+        }
+        m_lineOffsets[i] = std::round(m_lineOffsets[i]);
+    }
```
vittorioromeo, line 640:
```suggestion
    // convert widths into offsets depending on line alignment
    const auto applyLineOffset = [&](float value)
    {
        if (m_lineAlignment == LineAlignment::Left)
            return 0.f;
        
        if (m_lineAlignment == LineAlignment::Center)
            return (maxWidth - lineOffset) / 2.f;
        
        assert(m_lineAlignment == LineAlignment::Right);
        return maxWidth - lineOffset;
    };
        
    for (float& lineOffset : m_lineOffsets)
        lineOffset = std::round(applyLineOffset(lineOffset));
```

vittorioromeo (changes requested):
Just one more stylistic improvement. LGTM otherwise, I will test it soon

vittorioromeo:
> In addition, I'd like to add the suggested idea of aligning to its origin (e.g. aligning to the right would mean that text would be left of its position) and would like confirmation that this is the wanted behaviour.

I'm not too sure about this. Thoughts? @eXpl0it3r @ChrisThrasher @Bromeon

ChrisThrasher:
I’m about to go traveling without my computer so I can’t do any serious code reviews until the week of February 19th.

src/SFML/Graphics/Text.cpp
Hapaxia: This is quite an elegant use of more modern features.
However, I'd question why the multiple ifs are in place of a switch. Also, maybe we could test for center and right and default to left (i.e. 0 offset) if those options are not given. Which brings to the point that even the switch should have a default label.

eXpl0it3r:
A heads up: This PR might become obsolete, as a feature for complex text layout (#246) is currently being worked on, which will provide among other things basic layouting and shaping information.

eXpl0it3r:
With #3543 we could laverage HarfBuzz for proper text alignment and thus introduced the following API:
- `text.setLineAlignment(sf::Text::LineAlignment::Left)`
- `text.setLineAlignment(sf::Text::LineAlignment::Center)`
- `text.setLineAlignment(sf::Text::LineAlignment::Right)`

As such this PR is superseded

PR #2714, opened 2023-10-01
Remove Intel oneAPI CI jobs

ChrisThrasher:
## Description

These started spontaneously failing due to some Debian package repository key issues.

https://github.com/SFML/SFML/actions/runs/6366791146/job/17284823832#step:6:40

Perhaps we fix this problem later but honestly these jobs have brought little value as far as I'm aware. I'm not sure they're worth the time they take to run. Downloading the compiler takes about 2 minutes and that happens in every single job. We're saving a meaningful amount of CI time by removing them plus we get to remove more ugly CI code that I doubt anyone wants to maintain.


eXpl0it3r:
You'd probably have to add the correct key, but I'm also okay with removing them.

eXpl0it3r (approved):

PR #2717, opened 2023-10-03
Fix CMake minimum required version being too low

binary1248:
Support for specifying `SYSTEM` in `target_include_directories` for Visual Studio generators when using VS 2019 Update 11 or later was only [added in CMake 3.24](https://cmake.org/cmake/help/latest/release/3.24.html#generators). Building with a CMake version that is older will cause stbi warnings generated by MSVC to fail the build, see [this job which was run with CMake 3.23](https://github.com/SFML/SFML/actions/runs/6387751178/job/17336532042).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,12 @@
-cmake_minimum_required(VERSION 3.22)
+if(CMAKE_HOST_WIN32)
+    # We require a CMake version greater than or equal to 3.24 on Windows in order to
+    # support specifying target_include_directories with SYSTEM for Visual Studio Generators
+    # see: https://cmake.org/cmake/help/latest/release/3.24.html#generators
+    cmake_minimum_required(VERSION 3.24)
+else()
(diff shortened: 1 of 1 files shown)


ChrisThrasher:
This is something I’ve known about for a while. The problem is that if we do this then we will no longer seamlessly support Ubuntu 22 which ships with CMake 3.22 thus requiring those users to manually install a newer version through some means other than apt. 

Now that we disable warnings by default, what problem does it pose that the select few Windows users still on this old CMake release see a warning appear?

ChrisThrasher (approved):
I don't love this (or maybe I just don't love that we _have_ to do this) but I can't really argue against it. We do technically need 3.24 for some Windows configs. If the build fails because users somehow met all the requirements to invoke this error then it's really not obvious how to fix it.

As I now know this is because Microsoft changed VS in a way that broke the previously functional `SYSTEM` argument. It won't hurt to push more Windows users to a newer CMake release since it's plausible we upgrade even further before v3 ships.

PR #2718, opened 2023-10-03
Added missing SFML_USE_STATIC_STD_LIBS checks to CMake configuration.

binary1248:
4970af89992951eb2b4973bf32447445e9b38f64 changed to the newer way of linking to the MSVC static runtime, however it left out a few places in the CMake configuration where the check should have also been added. Without these changes, building statically with tests enabled and `SFML_USE_STATIC_STD_LIBS` enabled will fail with MSVC complaining about different runtime library linkage.

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -20,11 +20,17 @@ endmacro()
 # example: sfml_set_stdlib(sfml-system)
 function(sfml_set_stdlib target)
     # for gcc on Windows, apply the SFML_USE_STATIC_STD_LIBS option if it is enabled
-    if(SFML_OS_WINDOWS AND SFML_COMPILER_GCC)
-        if(SFML_USE_STATIC_STD_LIBS AND NOT SFML_COMPILER_GCC_TDM)
-            target_link_libraries(${target} PRIVATE "-static-libgcc" "-static-libstdc++")
-        elseif(NOT SFML_USE_STATIC_STD_LIBS AND SFML_COMPILER_GCC_TDM)
-            target_link_libraries(${target} PRIVATE "-shared-libgcc" "-shared-libstdc++")
+    if(SFML_OS_WINDOWS)
+        if(SFML_COMPILER_GCC)
+            if(SFML_USE_STATIC_STD_LIBS AND NOT SFML_COMPILER_GCC_TDM)
+                target_link_libraries(${target} PRIVATE "-static-libgcc" "-static-libstdc++")
+            elseif(NOT SFML_USE_STATIC_STD_LIBS AND SFML_COMPILER_GCC_TDM)
+                target_link_libraries(${target} PRIVATE "-shared-libgcc" "-shared-libstdc++")
+            endif()
+        elseif(SFML_COMPILER_MSVC)
+            if(SFML_USE_STATIC_STD_LIBS)
+                set_property(TARGET ${target} PROPERTY MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+            endif()
         endif()
     endif()
 endfunction()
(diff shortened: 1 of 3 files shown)


eXpl0it3r:
Just realized that this piece is misplaced, as it's in the else-branch, meaning it would come into effect for non-Windows OS, which is actually not compatible with `SFML_USE_STATIC_STD_LIBS`. Can you change that as well?

https://github.com/SFML/SFML/blob/75be07fa342d833bed07fe51effada18afd0c7df/cmake/Macros.cmake#L138-L140

Would this make sense to add here as well?

https://github.com/SFML/SFML/blob/75be07fa342d833bed07fe51effada18afd0c7df/cmake/Macros.cmake#L19-L30

Is it enough that we only define the flag when on Windows, or do we need additional Windows checks?

https://github.com/SFML/SFML/blob/75be07fa342d833bed07fe51effada18afd0c7df/CMakeLists.txt#L178-L179

And finally, does this not just apply to MSVC or is it simply ignored on say GCC or Clang?

binary1248:
The joke is that `sfml_set_stdlib` is only used when building macOS frameworks... so the code in that block is basically never run.

As for your first code block, if you look carefully the Windows check is actually in the inner `if` so I don't see the problem there.

To my knowledge CMake just ignores compiler specific variables when generating for other compilers.

binary1248:
Updated.

eXpl0it3r (approved):

ChrisThrasher:
Why not set [`CMAKE_MSVC_RUNTIME_LIBRARY `](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html#variable:CMAKE_MSVC_RUNTIME_LIBRARY) globally to avoid all this per-target boilerplate?

eXpl0it3r:
> Why not set [`CMAKE_MSVC_RUNTIME_LIBRARY`](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html#variable:CMAKE_MSVC_RUNTIME_LIBRARY) globally to avoid all this per-target boilerplate?

I thought per target setting is the correct way to go? If you set it globally , it will aslo affect anyone consuming SFML through CMake. I guess one could argue that mixing runtimes doesn't make sense.

ChrisThrasher:
> I thought per target setting is the correct way to go?

This is undoubtedly the case for anything `INTERFACE` or `PUBLIC` but for private properties like the, sometimes it's a lot more convenient to use the global alternative.

> If you set it globally, it will also affect anyone consuming SFML through CMake. I guess one could argue that mixing runtimes doesn't make sense.

I can't think of a reason this would be a problem but I'm admittedly not very familiar with all these flavors of MSVC runtime libraries and when/where they matter.

ChrisThrasher:
https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html#variable:CMAKE_MSVC_RUNTIME_LIBRARY

Screenshot 2023-10-15 at 4 43 37 PM

Are we sure this is something we need to set? It seems like CMake will automatically default the value we're already specifying. Why is this apparently not sufficient?

eXpl0it3r:
> This is undoubtedly the case for anything `INTERFACE` or `PUBLIC` but for private properties like this, sometimes it's a lot more convenient to use the global alternative.

How is this a private property?

ChrisThrasher:
> How is this a private property?

Properties themselves are not `PUBLIC`, `PRIVATE`, or `INTERFACE`. Those words only apply to select commands like `target_include_directories` or `target_link_libraries`. Under the hood those commands manipulate target properties like `INTERFACE_INCLUDE_DIRECTORIES`. Certain target properties get passed on to downstream targets, particularly those starting with `INTERFACE_`. I do not believe `MSVC_RUNTIME_LIBRARY` is a property that is used by downstream targets.

eXpl0it3r:
I don't understand your comment then, how does the usage of the target property make an argument for using global settings instead? If someone wants to use something else in the main project and then SFML changes the global, doesn't that have impact on the main project?

ChrisThrasher:
My point is that this is not an instance where a target-based solution is necessarily better. A global-based solution could be just as good but with way less code.

> If someone wants to use something else in the main project and then SFML changes the global, doesn't that have impact on the main project?
We can make this an `option` if we need to ensure users can override our choice. It'll take someone on a Windows machine to test this out. I'm not totally sure how this variable works and how it affects those consuming SFML via `add_subdirectory`.

eXpl0it3r:
I still don't see how it's a totally no-go to modify some `CMAKE_` variables, but then suddenly a better solution in other cases.

> We can make this an `option` if we need to ensure users can override our choice.

This feels a bit like a double standard. On one hand you want to reduce the option/config surface in general, but then here it's not a concern if it saves us a few lines of CMake code?

ChrisThrasher (approved):
I still think a simpler solution is possible but I can't yet prove that and don't want to keep stalling this PR until I find it.

Issue #2419, opened 2023-02-17
RenderTexture flipped on Android


Zombieschannel:
## Subject of the issue

Basically, when using render textures on Android, the image gets flipped (for some reason), however on Windows this works normally.
Images of the code below:

Windows 10

![image](https://user-images.githubusercontent.com/54750550/219801780-24e22edf-debb-44a9-9338-ff0fda366e60.png)

Android 8 and Android 13

![image](https://user-images.githubusercontent.com/54750550/219801830-cd3ebc49-f0a7-4d34-9ce9-5043b09fb9f5.png)

## Your environment

Android Studio 2022.1.1 on Windows 10
SFML 2.6.x branch
Used CMake, Ninja, Android NDK 25

## Steps to reproduce

```cpp
#include <SFML/Graphics.hpp>
int main()
{
    sf::RenderWindow window(sf::VideoMode::getDesktopMode(), "SFML works!", sf::Style::Fullscreen);
    window.setFramerateLimit(10);

    sf::CircleShape circle(400.f);
    sf::RectangleShape corner;
    sf::RectangleShape screen;
    circle.setFillColor(sf::Color::Green);
    corner.setSize(sf::Vector2f(400, 400));

    sf::RenderTexture texture;
    texture.create(window.getSize().x, window.getSize().y);
    texture.clear();
    screen.setSize((sf::Vector2f)window.getSize());

    sf::Texture t;

    bool active = window.hasFocus();
    while (window.isOpen())
    {
        sf::Event event;
        while (window.pollEvent(event))
        {
            switch (event.type)
            {
            case sf::Event::Closed:
                window.close();
                break;
            case sf::Event::LostFocus:
                window.setActive(0);
                active = 0;
                break;
            case sf::Event::GainedFocus:
                window.setActive(1);
                active = 1;
                break;
            case sf::Event::KeyReleased:
                if (event.key.code == sf::Keyboard::Escape)
                    window.close();
                break;
            }
        }
        if (!active)
            continue;
        t = texture.getTexture();
        corner.setTexture(&t);

        texture.clear();
        texture.draw(circle);
        texture.draw(corner);
        texture.display();
        screen.setTexture(&texture.getTexture());

        window.clear();
        window.draw(screen);
        window.display();

    }
}
```

## Expected behavior

As mentioned above, the texture should not get flipped.

## Actual behavior

On Android the texture that we got from rendertexture.getTexture() gets flipped.

Nathan-M-code:
[...]

PR #2719, opened 2023-10-03
Fixed texture being upside down on Android when copying the texture of a RenderTexture

texus:
## Description

RenderTexture was flipped on Android after the texture was copied, as described in https://github.com/SFML/SFML/issues/2419

This PR fixes https://github.com/SFML/SFML/issues/2419

## Platforms

The affected code is only compiled when `SFML_OPENGL_ES` is defined. The code was only tested on Android, but the bug should have existed on other platforms that use OpenGL ES. This was however not tested.

## How to test this PR?

The following code is supposed to render a green circle in the top left corner of the screen.
On Android, before the change from this PR is applied, the green circle will appear in the bottom left corner instead.

```cpp
#include <SFML/Graphics.hpp>
int main()
{
    sf::RenderWindow window(sf::VideoMode::getDesktopMode(), "SFML works!", sf::Style::Fullscreen);

    sf::CircleShape circle(400.f);
    circle.setFillColor(sf::Color::Green);

    sf::RenderTexture texture;
    texture.create({window.getSize().x, window.getSize().y});

    texture.clear();
    texture.draw(circle);
    texture.display();

    sf::Texture copiedTexture = texture.getTexture();
    sf::Sprite sprite(copiedTexture);

    bool active = window.hasFocus();
    while (window.isOpen())
    {
        sf::Event event;
        {
            switch (event.type)
            {
            case sf::Event::Closed:
                window.close();
                break;
            case sf::Event::LostFocus:
                window.setActive(0);
                active = 0;
                break;
            case sf::Event::GainedFocus:
                window.setActive(1);
                active = 1;
                break;
            case sf::Event::KeyReleased:
                    window.close();
                break;
            }
        }
        if (!active)
            continue;

        window.clear();
        window.draw(sprite);
        window.display();
    }
}
```

### DIFF
--- a/src/SFML/Graphics/Texture.cpp
+++ b/src/SFML/Graphics/Texture.cpp
@@ -35,6 +35,7 @@
 
 #include <SFML/System/Err.hpp>
 
+#include <algorithm>
 #include <atomic>
 #include <ostream>
 #include <utility>
@@ -400,6 +401,22 @@ Image Texture::copyToImage() const
         glCheck(GLEXT_glDeleteFramebuffers(1, &frameBuffer));
 
         glCheck(GLEXT_glBindFramebuffer(GLEXT_GL_FRAMEBUFFER, static_cast<GLuint>(previousFrameBuffer)));
+
+        if (m_pixelsFlipped)
+        {
+            // Flip the texture vertically
+            const auto stride             = static_cast<std::ptrdiff_t>(m_size.x * 4);
+            auto       currentRowIterator = pixels.begin();
+            auto       nextRowIterator    = pixels.begin() + stride;
+            auto       reverseRowIterator = pixels.begin() + (stride * static_cast<std::ptrdiff_t>(m_size.y - 1));
+            for (unsigned int i = 0; i < m_size.y / 2; ++i)
+            {
+                std::swap_ranges(currentRowIterator, nextRowIterator, reverseRowIterator);
+                currentRowIterator = nextRowIterator;
+                nextRowIterator += stride;
+                reverseRowIterator -= stride;
+            }
+        }
     }
 
 #else


binary1248:
Is allocating a temporary buffer necessary? It should be possible to perform the swap in-place using something like `std::swap_ranges`.

texus:
I've updated the code to use an in-place swap as suggested.

eXpl0it3r:

binary1248 (approved):

Zombieschannel (approved):
Can confirm that it fixes https://github.com/SFML/SFML/issues/2419
Just wanted to ask could this fix be added to the 2.6.x branch as well?

ChrisThrasher:
Fixed a minor conflict with `master`. Ready to merge pending CI passing.

src/SFML/Graphics/Texture.cpp
```diff
@@ -400,6 +401,22 @@ Image Texture::copyToImage() const
...
 
         glCheck(GLEXT_glBindFramebuffer(GLEXT_GL_FRAMEBUFFER, static_cast<GLuint>(previousFrameBuffer)));
+
+        if (m_pixelsFlipped)
+        {
+            // Flip the texture vertically
+            const auto stride             = static_cast<std::ptrdiff_t>(m_size.x * 4);
+            auto       currentRowIterator = pixels.begin();
+            auto       nextRowIterator    = pixels.begin() + stride;
+            auto       reverseRowIterator = pixels.begin() + (stride * static_cast<std::ptrdiff_t>(m_size.y - 1));
```
ChrisThrasher, line 411: Maybe we can use [`std::vector::rbegin()`](https://en.cppreference.com/w/cpp/container/vector/rbegin) here?

PR #2720, opened 2023-10-03
Manually install MinGW when running GitHub Actions workflows.

binary1248:
Title.

The recent changes to the workflow to work around the UCRT issue had the side effect of sometimes adding upwards of 10 minutes to every MinGW build.

Instead of relying on `choco` to manage MinGW for us, this change lets us do it ourself. The only time spent is decompressing the cache. Whenever we decide we want to support more/other MinGW versions it should be way easier to do so with this method rather than depending on the runner images or a `choco` package being available.


.github/workflows/ci.yml
```diff
@@ -108,11 +108,29 @@ jobs:
...
+
+    - name: Setup MinGW
       if: matrix.platform.name == 'Windows MinGW'
+      shell: bash
       run: |
-        choco uninstall mingw
-        choco install mingw --version 11.2.0
+        echo "CC=$GITHUB_WORKSPACE\\mingw64\\bin\\gcc.exe" >> $GITHUB_ENV
+        echo "CXX=$GITHUB_WORKSPACE\\mingw64\\bin\\g++.exe" >> $GITHUB_ENV
+        echo "RC=$GITHUB_WORKSPACE\\mingw64\\bin\\windres.exe" >> $GITHUB_ENV
```
ChrisThrasher, line 133: Why add this extra step instead of just changing the CMAKE_CXX_COMPILER configure param? I'd prefer we not not modify the environment when non-envvar options exist.

binary1248: This seemed like the cleaner of [both endorsed possibilities](https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#how-do-i-use-a-different-compiler) considering we specify the location where we unpack MinGW in the step itself. I'd rather not have to think of multiple places to have to make adjustments if we decide to move the directory at some point.

ChrisThrasher: I think it's cleaner to simply change existing arguments than add a whole new CI step that requires another 8 lines of code, plus it's consistent with all other jobs which specify the compiler this way. I'd rather not mix and match how we specify such things.

I know in your mega-CI-rewrite PR you made heavy use of environment variables but I hope we don't go down that route. I'd rather specify things via existing CMake interfaces wherever possible. I want that because when you look at the [Configure CMake](https://github.com/SFML/SFML/actions/runs/6387165344/job/17335087763) CI step you can see all configure parameters listed out whereas environment variables remain invisible.

ChrisThrasher:
I was able to figure out how we can make the caching of this new MinGW installation working without having to change any of the CMake configuration parameters. I'm happy with the current state of the PR. It's a massive CI savings to remove ~40 minutes of Windows CI time in every pipeline run. This is so worth the bit of extra boilerplate required to make it happen.

ChrisThrasher (approved):

eXpl0it3r (approved):

PR #2721, opened 2023-10-03
Cache packages downloaded when GitHub actions workflow is run.

binary1248:
Title.

This change aims to minimize the time spent downloading/installing packages. Bottlenecks requiring a larger amount of time in previous workflow runs were targeted and cached to reduce in some cases minutes down to seconds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,9 +74,15 @@ jobs:
     - name: Checkout Code
       uses: actions/checkout@v3
 
+    - name: Get CMake and Ninja
+      uses: lukka/get-cmake@latest
+      with:
+        cmakeVersion: ${{ runner.os == 'Windows' && '3.24' || '3.22' }} # Set to minimum required version (3.24 on Windows and 3.22 elsewhere) as specified in CMakeLists.txt to test compatibility
+        ninjaVersion: latest
+
     - name: Install Linux Dependencies
       if: runner.os == 'Linux'
-      run: sudo apt-get update && sudo apt-get install ninja-build xorg-dev libxrandr-dev libxcursor-dev libudev-dev libopenal-dev libflac-dev libvorbis-dev libgl1-mesa-dev libegl1-mesa-dev libdrm-dev libgbm-dev xvfb fluxbox
+      run: sudo apt-get update && sudo apt-get install xorg-dev libxrandr-dev libxcursor-dev libudev-dev libopenal-dev libflac-dev libvorbis-dev libgl1-mesa-dev libegl1-mesa-dev libdrm-dev libgbm-dev xvfb fluxbox
 
     - name: Install Android Components
       if: matrix.platform.name == 'Android'
@@ -96,18 +102,23 @@ jobs:
       run: |
         # Do not run updates until GitHub has fixed all the issues: https://github.com/orgs/Homebrew/discussions/4612
         # brew update
-        brew install gcovr ninja || true
+        brew install gcovr || true
 
-    - name: Install Windows Tools
+    - name: Cache OpenCppCoverage
+      if: matrix.type.name == 'Debug' && runner.os == 'Windows'
+      id: opencppcoverage-cache
+      uses: actions/cache@v3
+      with:
+        path: C:\Program Files\OpenCppCoverage
+        key: opencppcoverage
+
+    - name: Install OpenCppCoverage
       uses: nick-fields/retry@v2
-      if: runner.os == 'Windows'
+      if: matrix.type.name == 'Debug' && runner.os == 'Windows' && steps.opencppcoverage-cache.outputs.cache-hit != 'true'
       with:
         max_attempts: 10
         timeout_minutes: 3
-        command: |
-          choco install ninja OpenCppCoverage -y
-          echo "C:\Program Files\OpenCppCoverage" >> $env:GITHUB_PATH
-          echo "C:\Program Files\mingw64\bin" >> $env:GITHUB_PATH
+        command: choco install OpenCppCoverage -y
 
     - name: Cache MinGW
       if: matrix.platform.name == 'Windows MinGW'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -233,15 +265,24 @@ jobs:
...
         curl.exe -o run-clang-tidy https://raw.githubusercontent.com/llvm/llvm-project/llvmorg-15.0.7/clang-tools-extra/clang-tidy/tool/run-clang-tidy.py
 
     - name: Install Linux Dependencies
       if: runner.os == 'Linux'
-      run: sudo apt-get update && sudo apt-get install libxrandr-dev libxcursor-dev libudev-dev libopenal-dev libflac-dev libvorbis-dev libgl1-mesa-dev libegl1-mesa-dev libdrm-dev libgbm-dev
+      uses: awalsh128/cache-apt-pkgs-action@15d0235a4147963374ec3a9334d73adf0cb9eabc
+      with:
+        packages: libxrandr-dev libxcursor-dev libudev-dev libopenal-dev libflac-dev libvorbis-dev libgl1-mesa-dev libegl1-mesa-dev libdrm-dev libgbm-dev
+        version: 1.0
+        execute_install_scripts: true
```
ChrisThrasher, line 285: I tested this in isolation and proved that caching apt packages has no effect. Let's not bother doing this and instead keep our pipeline as simple as possible.

.github/workflows/ci.yml
```diff
@@ -96,24 +106,46 @@ jobs:
...
         choco install mingw --version 11.2.0
 
+    - name: Cache Mesa 3D
+      if: runner.os == 'Windows'
+      uses: actions/cache@v3
+      with:
+        path: |
+          build/mesa3d-23.0.0-release-msvc/x86
+          build/mesa3d-23.0.0-release-msvc/x64
+        key: mesa3d-23.0.0-release-msvc
```
ChrisThrasher, line 147: How much time does this save?

binary1248: Hard to say... if you try to compare side by side workflow runs then the timing is all over the place. Having to download and unpack the archive will always take some non-zero amount of time, but if the total time running CMake (discounting the Mesa3D stuff) depends on how generous GitHub is willing to provide us CPU time it is hard to provide any quantifiable value. Percentage-wise, the less time CMake spends doing other stuff the greater this will make a difference, e.g. it provides a bigger benefit when generating for Ninja than for MSBuild.

In the end, the upper limit of time saving is not going to exceed several seconds, so if you think it isn't worth it I can also get rid of the changes.

.github/workflows/ci.yml
binary1248: I'm guessing the packages we install are all pretty small compared to some of the bigger packages out there whose download time would dominate the installation process.

ChrisThrasher: Before caching: https://github.com/SFML/SFML/actions/runs/6387165344/job/17335087763

After caching: https://github.com/SFML/SFML/actions/runs/6388201930/job/17337670809

The difference in runtime is not significant. Perhaps it saves a few seconds but it's hard to say. I can cherry pick other examples where it's actually slower than the status quo. The evidence just doesn't prove that caching these packages is meaningful. 

Now how in the first link the packages are pulled from http://azure.archive.ubuntu.com/ubuntu. It appears that GitHub has their own apt mirror which may improve download speeds compared to pulling from Canonical's repositories. Downloading `apt` package is a popular thing to do so this would make sense.

binary1248:
Updated.

.github/workflows/ci.yml
```diff
@@ -232,10 +251,15 @@ jobs:
     - name: Checkout Code
       uses: actions/checkout@v3
 
+    - name: Get CMake and Ninja
+      uses: lukka/get-cmake@latest
+      with:
+        cmakeVersion: ${{ runner.os == 'Windows' && '3.24' || '3.22' }} # Set to minimum required version (3.24 on Windows and 3.22 elsewhere) as specified in CMakeLists.txt to test compatibility
+        ninjaVersion: latest
```
ChrisThrasher, line 258: We don't really need to use the lowest possible CMake when configuring for clang-tidy but at least this means avoiding downloading Ninja again which saves about 20 seconds.

.github/workflows/ci.yml
```diff
@@ -96,18 +102,25 @@ jobs:
...
-      if: runner.os == 'Windows'
+      if: matrix.type.name == 'Debug' && runner.os == 'Windows' && steps.opencppcoverage-cache.outputs.cache-hit != 'true'
       with:
         max_attempts: 10
         timeout_minutes: 3
         command: |
-          choco install ninja OpenCppCoverage -y
-          echo "C:\Program Files\OpenCppCoverage" >> $env:GITHUB_PATH
-          echo "C:\Program Files\mingw64\bin" >> $env:GITHUB_PATH
+          choco install OpenCppCoverage -y
```
ChrisThrasher, line 123: This can be moved onto the same line as `command`.

.github/workflows/ci.yml
```diff
@@ -96,18 +102,25 @@ jobs:
...
-        brew install gcovr ninja || true
+        brew install gcovr || true
+
+    - name: Cache OpenCppCoverage
+      if: matrix.type.name == 'Debug' && runner.os == 'Windows'
+      id: opencppcoverage-cache
+      uses: actions/cache@v3
+      with:
+        path: |
+          C:\Program Files\OpenCppCoverage
```
ChrisThrasher, line 113:
```suggestion
        path: C:\Program Files\OpenCppCoverage
```

ChrisThrasher (approved):
Just a few nit picks I'd like to see cleaned up but otherwise this is good stuff. I'm glad that we can start using the minimum CMake version in CI. I've wanted to do that for a while.

binary1248:
Updated.

ChrisThrasher (approved):

PR #2722, opened 2023-10-04
Added support for running the GitHub actions workflow using CCache

binary1248:
Title.

This changeset adds `ccache` support to speed up building by caching the produced object files whenever possible. Because it works by instructing the chosen CMake generator to insert it as the compiler launcher it won't be supported on vendor-specific generators like MSBuild or Xcode. MSVC support can be added later if we decide to move most of it over to Ninja.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,10 @@ jobs:
     name: ${{ matrix.platform.name }} ${{ matrix.config.name }} ${{ matrix.type.name }}
     runs-on: ${{ matrix.platform.os }}
 
+    env:
+      CMAKE_C_COMPILER_LAUNCHER: ccache # Use ccache to cache C compiler output
+      CMAKE_CXX_COMPILER_LAUNCHER: ccache # Use ccache to cache C++ compiler output
+
     strategy:
       fail-fast: false
       matrix:
@@ -72,9 +76,12 @@ jobs:
         cmakeVersion: ${{ runner.os == 'Windows' && '3.24' || '3.22' }} # Set to minimum required version (3.24 on Windows and 3.22 elsewhere) as specified in CMakeLists.txt to test compatibility
         ninjaVersion: latest
 
-    - name: Install Linux Dependencies
+    - name: Install Linux Dependencies and Tools
       if: runner.os == 'Linux'
-      run: sudo apt-get update && sudo apt-get install xorg-dev libxrandr-dev libxcursor-dev libudev-dev libopenal-dev libflac-dev libvorbis-dev libgl1-mesa-dev libegl1-mesa-dev libdrm-dev libgbm-dev xvfb fluxbox
+      run: |
+        CLANG_VERSION=$(clang++ --version | sed -n 's/.*version \([0-9]\+\)\..*/\1/p')
+        echo "CLANG_VERSION=$CLANG_VERSION" >> $GITHUB_ENV
+        sudo apt-get update && sudo apt-get install xorg-dev libxrandr-dev libxcursor-dev libudev-dev libopenal-dev libflac-dev libvorbis-dev libgl1-mesa-dev libegl1-mesa-dev libdrm-dev libgbm-dev xvfb fluxbox ccache gcovr ${{ matrix.platform.name == 'Linux Clang' && 'llvm-$CLANG_VERSION' || '' }}
 
     - name: Install Android Components
       if: matrix.platform.name == 'Android'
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -86,6 +86,22 @@ jobs:
         cmakeVersion: ${{ runner.os == 'Windows' && '3.24' || '3.22' }} # Set to minimum required version (3.24 on Windows and 3.22 elsewhere) as specified in CMakeLists.txt to test compatibility
         ninjaVersion: latest
 
+    # The pre-installed CCache on Windows seems to be broken (i.e. tests show that it doesn't work as intended)
+    # We get rid of the stock executable if any can be found and install a known working version
+    - name: Remove Broken Pre-Installed ccache
+      if: runner.os == 'Windows'
+      shell: bash
+      run: |
+        if [ -x "$(command -v ccache)" ]; then rm $(which ccache); fi
```
ChrisThrasher, line 95:
```suggestion
      run: if [ -x "$(command -v ccache)" ]; then rm $(which ccache); fi
```

.github/workflows/ci.yml
```diff
@@ -86,6 +86,22 @@ jobs:
         cmakeVersion: ${{ runner.os == 'Windows' && '3.24' || '3.22' }} # Set to minimum required version (3.24 on Windows and 3.22 elsewhere) as specified in CMakeLists.txt to test compatibility
         ninjaVersion: latest
 
+    # The pre-installed CCache on Windows seems to be broken (i.e. tests show that it doesn't work as intended)
+    # We get rid of the stock executable if any can be found and install a known working version
+    - name: Remove Broken Pre-Installed ccache
+      if: runner.os == 'Windows'
+      shell: bash
```
ChrisThrasher, line 93:
```suggestion
```

`master` now sets Bash as the default shell for all jobs.

CMakeLists.txt
```diff
@@ -31,7 +31,8 @@ set(VERSION_IS_RELEASE OFF)
 find_program(CCACHE_PROGRAM ccache)
 if(CCACHE_PROGRAM)
     message(STATUS "Found ccache in ${CCACHE_PROGRAM}")
-    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_PROGRAM}")
+    set(CMAKE_C_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
+    set(CMAKE_CXX_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
```
ChrisThrasher, line 35: What about `RULE_LAUNCH_COMPILE` isn't sufficient?

As an aside, none of this is necessary. See: https://github.com/SFML/SFML/pull/2032#issuecomment-1529981268

binary1248: [CMake documentation states that it is intended for internal use.](https://cmake.org/cmake/help/latest/prop_gbl/RULE_LAUNCH_COMPILE.html)

> Note: This property is intended for internal use by ctest(1). Projects and developers should use the `<LANG>_COMPILER_LAUNCHER` target properties or the associated `CMAKE_<LANG>_COMPILER_LAUNCHER` variables instead.

Also, the executable passed to `RULE_LAUNCH_COMPILE` would be used for everything that ends up being "compiled", this includes non-C/C++/C-like files which ccache does not support. This is definitely not the intention.

Also, I don't think anyone is realistically going to set up ccache via environment variable. As a C/C++ developer you are likely going to end up building third party libraries using CMake more often than not. Just setting up an environment variable would result in these libraries also taking space up in the cache even though you only intended to build them once (or at least very infrequently). The advantage ccache provides is speeding up frequent rebuilds, and that is something you only care about for your own projects and not everything that is built on the system.

I am also in favour of removing this block from the configuration and just passing the variables on the command line (or in our special case, even via environment variable since the CI workers are sandboxes). As long as this block is in the configuration the user isn't given a choice of choosing to build SFML with or without ccache if it is installed on their system.

ChrisThrasher: > [CMake documentation states that it is intended for internal use.](https://cmake.org/cmake/help/latest/prop_gbl/RULE_LAUNCH_COMPILE.html)

👍🏻

> I am also in favour of removing this block from the configuration and just passing the variables on the command line (or in our special case, even via environment variable since the CI workers are sandboxes). As long as this block is in the configuration the user isn't given a choice of choosing to build SFML with or without ccache if it is installed on their system.

In a later PR we can consider removing this but I'm fine leaving this in place for this PR.

binary1248:
Updated.

ChrisThrasher (approved):

binary1248:
Moved the environment variables into the `build` job since we don't install `ccache` when running the other jobs.

ChrisThrasher:
Do you have any numbers for how much this speeds up any particular CI jobs?

binary1248:
Comparing [building master](https://github.com/SFML/SFML/actions/runs/6725794846) to [this](https://github.com/SFML/SFML/actions/runs/6730959350), the total usage went from 6h 58min down to 5h 40min and total run duration went from 32min down to 22min. Bear in mind that this PR doesn't enable CCache for MSVC yet. I have follow-up commits for that and that will be another big chunk of time saved.

CMakeLists.txt
```diff
@@ -27,11 +27,12 @@ project(SFML VERSION 3.0.0)
...
-if(CCACHE_PROGRAM)
-    message(STATUS "Found ccache in ${CCACHE_PROGRAM}")
-    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_PROGRAM}")
+# Check if we are using any compiler launchers
+if(CMAKE_C_COMPILER_LAUNCHER)
+    message(STATUS "Using ${CMAKE_C_COMPILER_LAUNCHER} as C compiler launcher")
+endif()
+if(CMAKE_CXX_COMPILER_LAUNCHER)
+    message(STATUS "Using ${CMAKE_CXX_COMPILER_LAUNCHER} as C++ compiler launcher")
 endif()
```
ChrisThrasher, line 36: For what it's worth we can verify the use of ccache by looking at the compiler commands in CI or the compile_commands.json database.

Screenshot 2023-11-03 at 10 29 12 AM

binary1248: This output isn't only for our CI.

ChrisThrasher: Right, that's why I also mentioned `compile_commands.json` which can be checked locally as well.

binary1248:
Updated.

ChrisThrasher (approved):

PR #2723, opened 2023-10-05
Fixed compiling on Linux with SFML_OPENGL_ES

texus:
A typo was made in `EglContext::selectBestVisual` in [#2540](https://github.com/SFML/SFML/pull/2540) which breaks compilation on Linux when SFML_OPENGL_ES=ON

This PR adds the missing bracket to make the code compile again.

PR #2724, opened 2023-10-06
Added GitHub Actions jobs to build with OpenGL ES on Linux GCC.

binary1248:
Title.

Had to fix a few warnings.

Should prevent issues like #2723 happening in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,6 +68,12 @@ jobs:
           config: { name: Static DRM, flags: -GNinja -DSFML_USE_DRM=TRUE -DSFML_RUN_DISPLAY_TESTS=FALSE }
         - platform: { name: Linux GCC, os: ubuntu-22.04  }
           config: { name: Shared DRM, flags: -GNinja -DBUILD_SHARED_LIBS=TRUE -DSFML_USE_DRM=TRUE -DSFML_RUN_DISPLAY_TESTS=FALSE }
+        - platform: { name: Linux GCC, os: ubuntu-22.04 }
+          config: { name: OpenGL ES, flags: -GNinja -DSFML_OPENGL_ES=ON -DSFML_RUN_DISPLAY_TESTS=ON }
+          type: { name: Release }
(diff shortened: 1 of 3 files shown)


eXpl0it3r (approved):

.github/workflows/ci.yml
```diff
@@ -68,6 +68,12 @@ jobs:
           config: { name: Static DRM, flags: -GNinja -DSFML_USE_DRM=TRUE -DSFML_RUN_DISPLAY_TESTS=FALSE }
         - platform: { name: Linux GCC, os: ubuntu-22.04  }
           config: { name: Shared DRM, flags: -GNinja -DBUILD_SHARED_LIBS=TRUE -DSFML_USE_DRM=TRUE -DSFML_RUN_DISPLAY_TESTS=FALSE }
+        - platform: { name: Linux GCC, os: ubuntu-22.04 }
+          config: { name: OpenGL ES, flags: -GNinja -DSFML_OPENGL_ES=ON -DSFML_RUN_DISPLAY_TESTS=ON }
+          type: { name: Release }
+        - platform: { name: Linux GCC, os: ubuntu-22.04 }
+          config: { name: OpenGL ES, flags: -GNinja -DSFML_OPENGL_ES=ON -DSFML_RUN_DISPLAY_TESTS=ON }
+          type: { name: Debug, flags: -DCMAKE_BUILD_TYPE=Debug -DSFML_ENABLE_COVERAGE=TRUE }
```
ChrisThrasher, line 76: What do you think about simply adding a new line to the main build matrix for this? that's simpler than writing these 6 lines plus it means we get 4 jobs instead of 2 so better coverage of all possible configs. I'd also like to promote the Windows OpenGL ES jobs and the Linux DRM jobs. It would let us deduplicate a lot of code.

binary1248: We can do that all in a dedicated PR.

ChrisThrasher: That works for me

ChrisThrasher (approved):

PR #2725, opened 2023-10-07
Assert against division by zero in public APIs

ChrisThrasher:
## Description

Yet another class of UB we can eliminate in debug builds. I'm not concerned about the case where users provide the literal `0` but rather the cases where the divisor is the result of some calculation and may be zero in some cases.

### DIFF
--- a/include/SFML/System/Angle.inl
+++ b/include/SFML/System/Angle.inl
@@ -191,6 +191,7 @@ constexpr Angle& operator*=(Angle& left, float right)
 ////////////////////////////////////////////////////////////
 constexpr Angle operator/(Angle left, float right)
 {
+    assert(right != 0 && "Angle::operator/ cannot divide by 0");
     return degrees(left.asDegrees() / right);
 }
 
@@ -198,6 +199,7 @@ constexpr Angle operator/(Angle left, float right)
 ////////////////////////////////////////////////////////////
 constexpr Angle& operator/=(Angle& left, float right)
 {
+    assert(right != 0 && "Angle::operator/= cannot divide by 0");
     return left = left / right;
 }
 
@@ -205,6 +207,7 @@ constexpr Angle& operator/=(Angle& left, float right)
 ////////////////////////////////////////////////////////////
 constexpr float operator/(Angle left, Angle right)
 {
+    assert(right.asDegrees() != 0 && "Angle::operator/ cannot divide by 0");
     return left.asDegrees() / right.asDegrees();
 }
 
@@ -212,6 +215,7 @@ constexpr float operator/(Angle left, Angle right)
 ////////////////////////////////////////////////////////////
 constexpr Angle operator%(Angle left, Angle right)
 {
+    assert(right.asDegrees() != 0 && "Angle::operator% cannot divide by 0");
     return degrees(priv::positiveRemainder(left.asDegrees(), right.asDegrees()));
 }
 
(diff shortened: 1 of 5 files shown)


binary1248 (approved):

include/SFML/System/Time.inl
```diff
@@ -213,34 +213,39 @@ constexpr Time& operator*=(Time& left, std::int64_t right)
...
 {
+    assert(right != 0 && "Cannot divide by zero");
     return left = left / right;
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr float operator/(Time left, Time right)
 {
+    assert(right.asMicroseconds() != 0 && "Cannot divide by zero");
```
eXpl0it3r, line 248: Should these also mentioned the function name?

include/SFML/System/Angle.inl
```diff
@@ -191,20 +191,23 @@ constexpr Angle& operator*=(Angle& left, float right)
...
 {
+    assert(right != 0 && "Cannot divide by zero");
     return left = left / right;
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr float operator/(Angle left, Angle right)
 {
+    assert(right.asDegrees() != 0 && "Cannot divide by zero");
```
eXpl0it3r, line 210: Should these also mention the method name?

include/SFML/System/Time.inl
ChrisThrasher: We do that in Vector2 and Vector3 but not elsewhere. I made sure to keep the vectors internally consistent but we have no established convention for whether assertions in public APIs ought to name the public API they're contained within.

Bromeon (commented):
Thanks!

Modulo with 2nd operand zero could also be protected against, what do you think?

ChrisThrasher:
> Thanks!
> 
> Modulo with 2nd operand zero could also be protected against, what do you think?

Good idea! I forget that modulo operations involve divisions. Looks like only `sf::Time` and `sf::Angle` include that so it's an easy fix. I looked around and couldn't find any other public APIs where the right side of a modulo operation uses user input.

include/SFML/System/Time.inl
ChrisThrasher: Would you like to establish a convention about whether or not we do this? I'm open to that idea.

include/SFML/System/Time.inl
```diff
@@ -213,48 +213,55 @@ constexpr Time& operator*=(Time& left, std::int64_t right)
...
     return left.asSeconds() / right.asSeconds();
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr Time operator%(Time left, Time right)
 {
+    assert(right.asMicroseconds() != 0 && "Cannot divide by zero");
```
Bromeon, line 256: Maybe "modulo by zero"? 

Even though closely related to division, modulo is a separate operation, and I imagine that people who get "division by zero" errors would look for divisions, not modulos.

include/SFML/System/Angle.inl
```diff
@@ -191,34 +191,39 @@ constexpr Angle& operator*=(Angle& left, float right)
...
 {
+    assert(right.asDegrees() != 0 && "Angle::operator/ cannot divide by 0");
     return left.asDegrees() / right.asDegrees();
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr Angle operator%(Angle left, Angle right)
 {
+    assert(right.asDegrees() != 0 && "Angle::operator% cannot divide by 0");
```
kimci86, line 218: Inconsistency here. This is a modulo operator, not a division.

PR #2728, opened 2023-10-08
Added 'm_cacheId' getter to sf::Texture

KeithDobbelaere:
## Description
This member function would be very useful for checking whether a sf::Texture has been swapped/created between calls to a class that contains it.  An example would be custom font or text classes that operate on some format not supported directly by the SFML API, e.g., SVG files or some custom raster font.  My use case was a font class that loaded individual PNG files from an existing application (Super Mario 64 port) and built a dynamically sized texture from them, mapping their texture rects to characters. These could then be rendered to the screen via a custom text class with a vertex buffer, minimizing draw calls. I need a way to check whether the texture has been resized between calls the the custom text object, which may not be accessed every frame. I don't need direct access to an internal variable like m_cacheId; that would be stupid, but I don't see a reason to preclude someone from utilizing its value, and the additional code is pretty minuscule.  I've searched for this topic online and found one example of someone bringing it up, but it didn't seem like they made a very good case for it, and I believe they even suggested making the variable public, which again, like I said, is pretty silly.

### DIFF
--- a/include/SFML/Graphics/Texture.hpp
+++ b/include/SFML/Graphics/Texture.hpp
@@ -548,6 +548,20 @@ public:
     unsigned int getNativeHandle() const;
 
     ////////////////////////////////////////////////////////////
+    /// \brief Get the unique number that identifies the texture to
+    /// the render target's cache
+    ///
+    /// Can be stored in classes that utilize the sf::Texture class
+    /// and compared on subsequent calls to member functions of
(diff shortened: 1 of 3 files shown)


ChrisThrasher:
I spoke with some other SFML maintainers and this is not something we'd like to do because it exposes implementation details that we would rather users not be able to access. A change like this hinders our ability to make implementation changes in the future.

PR #2734, opened 2023-10-17
Add `cppcoreguidelines-pro-type-member-init` clang-tidy check

ChrisThrasher:
## Description

https://releases.llvm.org/16.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/cppcoreguidelines/pro-type-member-init.html

This marks another good step towards systematically rooting out undefined behavior in the form of reading uninitialized memory.

> The check takes assignment of fields in the constructor body into account but generates false positives for fields initialized in methods invoked in the constructor body.

I noticed a few of these false positives but found satisfactory workarounds.

### DIFF
--- a/include/SFML/Window/Event.hpp
+++ b/include/SFML/Window/Event.hpp
@@ -35,6 +35,8 @@
 #include <SFML/Window/Sensor.hpp>
 
 
+// NOLINTBEGIN(cppcoreguidelines-pro-type-member-init)
+
 namespace sf
 {
 ////////////////////////////////////////////////////////////
@@ -200,7 +202,7 @@ struct Event
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    EventType type; //!< Type of the event
+    EventType type{}; //!< Type of the event
 
     union
     {
@@ -219,6 +221,8 @@ struct Event
 };
 
 } // namespace sf
+
+// NOLINTEND(cppcoreguidelines-pro-type-member-init)
 
 
 ////////////////////////////////////////////////////////////
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -42,7 +42,6 @@ Checks: >
   -cppcoreguidelines-pro-bounds-constant-array-index,
   -cppcoreguidelines-pro-bounds-pointer-arithmetic,
   -cppcoreguidelines-pro-type-const-cast,
-  -cppcoreguidelines-pro-type-member-init,
   -cppcoreguidelines-pro-type-reinterpret-cast,
   -cppcoreguidelines-pro-type-static-cast-downcast,
   -cppcoreguidelines-pro-type-union-access,
--- a/examples/vulkan/Vulkan.cpp
+++ b/examples/vulkan/Vulkan.cpp
@@ -2595,7 +2595,7 @@ private:
     VkImageView                     depthImageView{};
     VkShaderModule                  vertexShaderModule{};
     VkShaderModule                  fragmentShaderModule{};
-    VkPipelineShaderStageCreateInfo shaderStages[2];
+    VkPipelineShaderStageCreateInfo shaderStages[2]{};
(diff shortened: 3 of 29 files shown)


binary1248:
Wouldn't it be simpler/cleaner to just give `sf::Event` a proper constructor instead of having to add braces to 100s of places?

ChrisThrasher:
> Wouldn't it be simpler/cleaner to just give `sf::Event` a proper constructor instead of having to add braces to 100s of places?

I played with this before to no avail but I tried again and got something to work. Thanks for bringing this up.

include/SFML/Window/Event.hpp
```diff
@@ -200,11 +200,11 @@ struct Event
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    EventType type; //!< Type of the event
+    EventType type{}; //!< Type of the event
 
     union
     {
-        SizeEvent      size;      //!< Size event parameters (Event::Resized)
+        SizeEvent      size{};    //!< Size event parameters (Event::Resized)
```
binary1248, line 207: This would only initialize the first 8 bytes of the union to 0. The rest remains uninitialized. Why not just provide a union constructor that `memset`s itself to all 0s?

ChrisThrasher: Only one field can be active at a time. Why would we need to initialize more than one field? I don't care about the other bits because accessing them without changing the active field is UB.

kimci86: Why initializing a specific field when default event type (Event::Closed) is not associated to any data?

ChrisThrasher: Otherwise no field is initialized. Now the initial value of an event is fully defined. It's just the reality of `sf::Event` that its default value is kind of weird.

binary1248: Then why bother initializing the union at all? `type` is initialized to 0 corresponding to `Closed` which would mean that a `sf::Event` in its default state wouldn't allow anything in the union to be accessed. Accessing the initialized size member in this scenario would be by your definition also UB. It's either we fill everything with 0s so we know what we are getting even in an unintended access case, or we shift the blame to the user for not checking the type before accessing the field. I prefer to err on the side of caution and not return garbage. Better yet replace the union with `std::variant` and let it do all the heavy lifting. In any correct usage case nobody should have to fear exceptions being thrown.

ChrisThrasher: > Then why bother initializing the union at all?

Because you wanted to remove `event{}` everywhere so I figured out how to do that.

binary1248: I can only assume the guidelines were written with `std::variant` usage in mind. Getting rid of the unions is probably the intended solution for conforming to this rule.

ChrisThrasher: I doubt that. What I wrote is a valid solution to clang-tidy detecting a struct with an uninitialized member. Users won't really know or care about this. This is just a terse way of satisfying clang-tidy. Would you rather we go back to writing `event{};` everywhere?

ChrisThrasher: I was about to get clang-tidy to ignore the uninitialized data in `sf::Event`. We can continue that conversation later or better yet, come up with a variant-based alternative and avoid this problem entirely. I'll resolve this thread since I was able to sidestep this problem.

binary1248: I was curious about this behaviour so I threw [some code into compiler explorer](https://gcc.godbolt.org/z/j4jEPE5j7).

Brace initializing the objects all over the place would have zeroed out the whole `struct`, whereas only adding the initializer to the first `union` member would have only partially initialized the `struct`. Strangely enough the compilers don't all produce the same output, which would mean that union initialization seems to be implementation-specific beyond the first member if it has a constructor. The GCC output produces deterministic behaviour, while clang and MSVC just resort to UB.

binary1248 (approved):

PR #2736, opened 2023-10-20
Add tests for `sf::Image::saveToFile`

ChrisThrasher:
## Description

The `<filesystem>` library makes it feasible to create and clean up temporary files so this didnt' end up being too difficult. It's not entirely exception-safe since `std::filesystem::remove` is not called within a destructor but (1) SFML does not throw exceptions and (2) it's not a big deal if a few files litter up someone's temporary directory.

While I was at it, I found 3 places to improve diagnostics when users incorrectly use the `sf::Image` API including 2 more instances of `assert`ing against UB.

Codecov confirms that this PR has perfect coverage of `sf::Image::saveToFile` and the other internal functions it calls into.

### DIFF
--- a/test/Graphics/Image.test.cpp
+++ b/test/Graphics/Image.test.cpp
@@ -129,6 +129,82 @@ TEST_CASE("[Graphics] sf::Image")
         CHECK(image.getPixelsPtr() != nullptr);
     }
 
+    SECTION("saveToFile()")
+    {
+        sf::Image image;
+
+        SECTION("Empty")
+        {
+            CHECK(!image.saveToFile("test.jpg"));
+        }
+
+        SECTION("Invalid size")
+        {
+            image.create({10, 0}, sf::Color::Magenta);
+            CHECK(!image.saveToFile("test.jpg"));
+            image.create({0, 10}, sf::Color::Magenta);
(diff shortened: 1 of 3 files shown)


binary1248 (approved):

test/Graphics/Image.test.cpp
```diff
@@ -129,6 +129,82 @@ TEST_CASE("[Graphics] sf::Image")
...
+                filename /= "test.jpeg";
+                CHECK(image.saveToFile(filename));
+            }
+
+            sf::Image loadedImage;
+            REQUIRE(loadedImage.loadFromFile(filename));
+            CHECK(image.getPixel({0, 0}) == sf::Color::Magenta);
+            CHECK(image.getPixel({255, 255}) == sf::Color::Magenta);
+            CHECK(image.getSize() == sf::Vector2u(256, 256));
+            CHECK(image.getPixelsPtr() != nullptr);
```
kimci86, line 202: I guess you meant to check this on loadedImage.

ChrisThrasher: Oops I meant to fix that but must have gotten distracted. 

https://github.com/SFML/SFML/pull/2738

PR #2739, opened 2023-10-21
Unify `SocketImpl` headers

ChrisThrasher:
## Description
The two headers only differed in their inclusions and two type aliases. It's easier to use the preprocessor than the maintain two nearly identical files.


ChrisThrasher:

eXpl0it3r (approved):
The whole point of separate impl files per OS is to prevent this varying ifdef-ing. I guess it's okay-is here, as it's very localized and affecting only some types.

Generally speaking, I don't see it as a good way to remove "duplicated" code if we sacrifice separation of "concerns" to achieve it. I much rather have some duplicated code, than having to carefully weave different OS code in one file.

ChrisThrasher:
> The whole point of separate impl files per OS is to prevent this varying ifdef-ing. I guess it's okay-is here, as it's very localized and affecting only some types.

I wasn't sure if that was intentional. Given how the majority of the code is identical it seemed like an oversight to repeat so much. If we end up needing more complicated preprocessor logic in the future I'm totally open to moving this back into separate files.

PR #2741, opened 2023-10-21
Bevel shape outline beyond threshold

### LINKED ISSUE #2727: `sf::ConvexShape` outline glitching when shape is too small
## Subject of the issue

I've tried to render some simple 3D rotating cube using triangles.
Triangles were implemented with `sf::ConvexShape`.  When i added `sf::ConvexShape.setOutlineColor`.  I noticed graphical glitch that occurs when the side of a cube becomes too small when rotated. I will add video to be clear. 

## Your environment

* Distro: Ubuntu 22.04.3 LTS
* SFML version: 2.6.x
* Compiler: gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0

## Steps to reproduce

```cpp
#include <iostream>
#include <SFML/Graphics.hpp>

int main() {
    sf::RenderWindow window(sf::VideoMode(800u, 800u), "");

    sf::ConvexShape shape;
    shape.setPointCount(3);
    sf::Vector2f p1(100.f, 100.f);
    sf::Vector2f p2(200.f, 200.f);
    sf::Vector2f p3(100.f, 400.f);

    shape.setPoint(0, p1);
    shape.setPoint(1, p2);
    shape.setPoint(2, p3);

    shape.setOutlineColor(sf::Color::White);
    shape.setOutlineThickness(2);

    sf::Clock clock;
    while(window.isOpen()) {
        float dt = clock.restart().asSeconds();
        sf::Event event;
            if (event.type == sf::Event::Closed) {
                window.close();
            }
        }
        if (sf::Keyboard::isKeyPressed(sf::Keyboard::S)) {
            p1.y += 100.f * dt;        
        }
        if (sf::Keyboard::isKeyPressed(sf::Keyboard::W)) {
            p1.y -= 100.f * dt;        
        }
        if (sf::Keyboard::isKeyPressed(sf::Keyboard::A)) {
            p1.x -= 100.f * dt;        
        }
        if (sf::Keyboard::isKeyPressed(sf::Keyboard::D)) {
            p1.x += 100.f * dt;        
        }

        shape.setPoint(0, p1);
        shape.setPoint(1, p2);
        shape.setPoint(2, p3);
        
        window.clear();
        window.draw(shape);
        window.display();
    }
}
```
https://github.com/SFML/SFML/assets/94391872/682691c8-3dc0-4833-811f-d0069f9464a6

kimci86:
## Description

This changes `sf::Shape` outline computation by beveling outline around small angles.
This is to fix #2727.

I noticed and fixed another issue that was visible with my test program below. See the first commit.

Here are videos showing the behavior of the test program before and after this PR commits.

https://github.com/SFML/SFML/assets/1485306/6d60a966-7ad8-4eee-b5db-74214eaaaf98

https://github.com/SFML/SFML/assets/1485306/9895a0be-400d-40d4-89e4-f557e1fc94ac

<details> 
  <summary>Initial result of the PR before review was:</summary>

https://github.com/SFML/SFML/assets/1485306/51148f17-77c0-4da7-a331-cc8d5253e115

</details>

## Tasks

- [x] Write more documentation

## How to test this PR?

Moving the mouse and scrolling when running this program, seeing that the outline looks good.

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    constexpr auto worldSize = sf::Vector2f{800, 600};

    auto shape = sf::ConvexShape{3};
    shape.setPosition(worldSize / 2.f);
    shape.setPoint(0, {0, 100});
    shape.setPoint(1, {-100, 0});
    shape.setPoint(2, {100, 0});
    shape.setFillColor(sf::Color::White);
    shape.setOutlineColor({0, 255, 0, 128});
    shape.setOutlineThickness(20);

    auto window = sf::RenderWindow{sf::VideoMode{sf::Vector2u{worldSize}}, "Shape example"};
    window.setMouseCursorVisible(false);

    while (window.isOpen())
    {
        window.handleEvents([&](const sf::Event::Closed&) { window.close(); },
                            [&](const sf::Event::MouseMoved& mouseMove)
                            {
                                const auto pixels      = mouseMove.position;
                                const auto coords      = window.mapPixelToCoords(pixels);
                                const auto shapeCoords = shape.getTransform().getInverse() * coords;
                                shape.setPoint(0, shapeCoords);
                            },
                            [&](const sf::Event::MouseWheelScrolled& scroll)
                            { shape.setOutlineThickness(shape.getOutlineThickness() + scroll.delta); });

        window.clear();
        window.draw(shape);
        window.display();
    }
}
```
Program can be built with:
```cmake
project(shape-test CXX)

include(FetchContent)
FetchContent_Declare(SFML
    GIT_TAG feature/shape_miter_limit)
FetchContent_MakeAvailable(SFML)

add_executable(shape-test main.cpp)
target_link_libraries(shape-test PRIVATE SFML::Graphics)
```

### DIFF
--- a/include/SFML/Graphics/Shape.hpp
+++ b/include/SFML/Graphics/Shape.hpp
@@ -135,6 +135,37 @@ public:
     void setOutlineThickness(float thickness);
 
     ////////////////////////////////////////////////////////////
+    /// \brief Set the limit on the ratio between miter length and outline thickness
+    ///
+    /// Outline segments around each shape corner are joined either
+    /// with a miter or a bevel join.
+    /// - A miter join is formed by extending outline segments until
+    ///   they intersect. The distance between the point of
+    ///   intersection and the shape's corner is the miter length.
+    /// - A bevel join is formed by connecting outline segments with
+    ///   a straight line perpendicular to the corner's bissector.
+    ///
+    /// The miter limit is used to determine whether ouline segments
+    /// around a corner are joined with a bevel or a miter.
+    /// When the ratio between the miter length and outline thickness
+    /// exceeds the miter limit, a bevel is used instead of a miter.
+    ///
+    /// The miter limit is linked to the maximum inner angle of a
+    /// corner below which a bevel is used by the following formula:
+    ///
+    /// miterLimit = 1 / sin(angle / 2)
+    ///
+    /// The miter limit must be greater than or equal to 1.
+    /// By default, the miter limit is 10.
+    ///
+    /// \param miterLimit New miter limit
+    ///
+    /// \see getMiterLimit
+    ///
+    ////////////////////////////////////////////////////////////
+    void setMiterLimit(float miterLimit);
+
+    ////////////////////////////////////////////////////////////
     /// \brief Get the source texture of the shape
     ///
     /// If the shape has no source texture, a `nullptr` is returned.
@@ -187,6 +218,16 @@ public:
     ///
     ////////////////////////////////////////////////////////////
     [[nodiscard]] float getOutlineThickness() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the limit on the ratio between miter length and outline thickness
+    ///
+    /// \return Limit on the ratio between miter length and outline thickness
+    ///
+    /// \see setMiterLimit
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] float getMiterLimit() const;
 
     ////////////////////////////////////////////////////////////
     /// \brief Get the total number of points of the shape
@@ -315,6 +356,7 @@ private:
     Color          m_fillColor{Color::White};    //!< Fill color
     Color          m_outlineColor{Color::White}; //!< Outline color
     float          m_outlineThickness{};         //!< Thickness of the shape's outline
+    float          m_miterLimit{10.f};           //!< Limit on the ratio between miter length and outline thickness
     VertexArray    m_vertices{PrimitiveType::TriangleFan};          //!< Vertex array containing the fill geometry
     VertexArray    m_outlineVertices{PrimitiveType::TriangleStrip}; //!< Vertex array containing the outline geometry
     FloatRect      m_insideBounds;                                  //!< Bounding rectangle of the inside (fill)
--- a/src/SFML/Graphics/Shape.cpp
+++ b/src/SFML/Graphics/Shape.cpp
@@ -37,12 +37,14 @@
 namespace
 {
 // Compute the normal of a segment
-sf::Vector2f computeNormal(sf::Vector2f p1, sf::Vector2f p2)
+sf::Vector2f computeNormal(sf::Vector2f p1, sf::Vector2f p2, bool flipped)
 {
     sf::Vector2f normal = (p2 - p1).perpendicular();
     const float  length = normal.length();
     if (length != 0.f)
         normal /= length;
+    if (flipped)
+        normal = -normal;
     return normal;
 }
 } // namespace
@@ -291,6 +293,15 @@ void Shape::updateOutline()
     const std::size_t count = m_vertices.getVertexCount() - 2;
     m_outlineVertices.resize((count + 1) * 2);
 
+    // Determine if points are defined clockwise or counterclockwise. This will impact normals computation.
+    const bool flipNormals = [this, count]()
+    {
+        float twiceArea = 0.f;
+        for (std::size_t i = 0; i < count; ++i)
+            twiceArea += m_vertices[i + 1].position.cross(m_vertices[i + 2].position);
+        return twiceArea >= 0.f;
+    }();
+
     for (std::size_t i = 0; i < count; ++i)
     {
         const std::size_t index = i + 1;
@@ -300,16 +311,9 @@ void Shape::updateOutline()
         const Vector2f p1 = m_vertices[index].position;
         const Vector2f p2 = m_vertices[index + 1].position;
 
-        // Compute their normal
-        Vector2f n1 = computeNormal(p0, p1);
-        Vector2f n2 = computeNormal(p1, p2);
-
-        // Make sure that the normals point towards the outside of the shape
-        // (this depends on the order in which the points were defined)
-        if (n1.dot(m_vertices[0].position - p1) > 0)
-            n1 = -n1;
-        if (n2.dot(m_vertices[0].position - p1) > 0)
-            n2 = -n2;
+        // Compute their normal pointing towards the outside of the shape
+        const Vector2f n1 = computeNormal(p0, p1, flipNormals);
+        const Vector2f n2 = computeNormal(p1, p2, flipNormals);
 
         // Combine them to get the extrusion direction
         const float    factor = 1.f + (n1.x * n2.x + n1.y * n2.y);
--- a/src/SFML/Graphics/Shape.cpp
+++ b/src/SFML/Graphics/Shape.cpp
@@ -296,10 +296,21 @@ void Shape::updateOutline()
     // Determine if points are defined clockwise or counterclockwise. This will impact normals computation.
     const bool flipNormals = [this, count]()
     {
-        float twiceArea = 0.f;
+        // p0 is either strictly inside the shape, or on an edge.
+        const sf::Vector2f p0 = m_vertices[0].position;
         for (std::size_t i = 0; i < count; ++i)
-            twiceArea += m_vertices[i + 1].position.cross(m_vertices[i + 2].position);
-        return twiceArea >= 0.f;
+        {
+            const sf::Vector2f p1      = m_vertices[i + 1].position;
+            const sf::Vector2f p2      = m_vertices[i + 2].position;
+            const float        product = (p1 - p0).cross(p2 - p0);
+            if (product == 0.f)
+            {
+                // p0 is on the edge p1-p2. We cannot determine shape orientation yet, so continue.
+                continue;
+            }
+            return product > 0.f;
+        }
+        return true;
     }();
 
     for (std::size_t i = 0; i < count; ++i)
--- a/src/SFML/Graphics/Shape.cpp
+++ b/src/SFML/Graphics/Shape.cpp
@@ -32,20 +32,19 @@
 #include <algorithm>
 
 #include <cassert>
+#include <cmath>
 #include <cstddef>
 
 namespace
 {
-// Compute the normal of a segment
-sf::Vector2f computeNormal(sf::Vector2f p1, sf::Vector2f p2, bool flipped)
+// Compute the direction of a segment
+sf::Vector2f computeDirection(sf::Vector2f p1, sf::Vector2f p2)
 {
-    sf::Vector2f normal = (p2 - p1).perpendicular();
-    const float  length = normal.length();
+    sf::Vector2f direction = p2 - p1;
+    const float  length    = direction.length();
     if (length != 0.f)
-        normal /= length;
-    if (flipped)
-        normal = -normal;
-    return normal;
+        direction /= length;
+    return direction;
 }
 } // namespace
 
@@ -123,7 +122,7 @@ Color Shape::getOutlineColor() const
 void Shape::setOutlineThickness(float thickness)
 {
     m_outlineThickness = thickness;
-    update(); // recompute everything because the whole shape must be offset
+    updateOutline();
 }
 
 
@@ -131,6 +130,22 @@ void Shape::setOutlineThickness(float thickness)
 float Shape::getOutlineThickness() const
 {
     return m_outlineThickness;
+}
+
+
+////////////////////////////////////////////////////////////
+void Shape::setMiterLimit(float miterLimit)
+{
+    assert(miterLimit >= 1.f && "Shape::setMiterLimit(float) cannot set miter limit to a value lower than 1");
+    m_miterLimit = miterLimit;
+    updateOutline();
+}
+
+
+////////////////////////////////////////////////////////////
+float Shape::getMiterLimit() const
+{
+    return m_miterLimit;
 }
 
 
@@ -282,8 +297,8 @@ void Shape::updateTexCoords()
 ////////////////////////////////////////////////////////////
 void Shape::updateOutline()
 {
-    // Return if there is no outline
-    if (m_outlineThickness == 0.f)
+    // Return if there is no outline or no vertices
+    if (m_outlineThickness == 0.f || m_vertices.getVertexCount() < 2)
     {
         m_outlineVertices.clear();
         m_bounds = m_insideBounds;
@@ -291,7 +306,8 @@ void Shape::updateOutline()
     }
 
     const std::size_t count = m_vertices.getVertexCount() - 2;
-    m_outlineVertices.resize((count + 1) * 2);
+    m_outlineVertices.resize((count + 1) * 2); // We need at least that many vertices.
+                                               // We will add two more vertices each time we need a bevel.
 
     // Determine if points are defined clockwise or counterclockwise. This will impact normals computation.
     const bool flipNormals = [this, count]()
@@ -313,6 +329,7 @@ void Shape::updateOutline()
         return true;
     }();
 
+    std::size_t outlineIndex = 0;
     for (std::size_t i = 0; i < count; ++i)
     {
         const std::size_t index = i + 1;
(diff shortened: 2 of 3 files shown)


kimci86:
I forgot about getGeometricCenter being virtual. I will rework my code to take advantage of this potential optimization.

ChrisThrasher:
I'm so glad you implemented this :) I have a few design questions

1. Why not make the miter size a parameter? Is there a particularly useful reason why it ought to be a hardcoded constant? Is it simply not useful to add an API for it?
1. Why make this feature on by default? My impression at first was that this would be an opt-in feature since it changes how drawing works to something that is perhaps not what users expect at first. This ties into the first question since to opt into this feature entails adding a new API.
1. The transition from a non-mitered to a mitered state is rather abrupt. Is there anything we can do to make that a bit smoother? What I was initially picturing was that as an internal angle gets more acute, instead of the external point on the outlet moving farther out, we see the miter open up. This is in contrast to the current implementation where the point on the triangle get shorter, so to speak. That abrupt shortening is what I'd prefer to eliminate if feasible.

I can help figure out how to test this once we talk through more of the higher level API concerns.

kimci86:
1. No reason besides me not being brave enough to write the documentation right away. The limit probably should be a parameter.
2. The feature being off means outlining a shape around a zero angle is a bug. I think the limit should never be off but I agree users may want to set a higher limit.
3. I took inspiration from SVG behavior not to need to get creative but we can implement whatever we want for sure. I will experiment with some smoother alternative soon, hopefully tonight.

Thank you for your feedback!

ChrisThrasher:
> I noticed and fixed another issue that was visible with my test program below. See https://github.com/SFML/SFML/commit/71826dec051cee424f67f4f1cf8300405d27a13d.

I just realized you did this. Nice find! This is a huge vote of confident for `getGeometricCenter` as a feature.

ChrisThrasher:
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-miterlimit

In the graphic that shows 3 different miter sizes, the middle one is what looks correct to me. Is this a fundamentally different behavior or simply demonstrating a different miter value? If the middle behavior is achievable via a different miter size then I don't think you need to change any of the algorithms.

kimci86:
> In the graphic that shows 3 different miter sizes, the middle one is what looks more desirable to me. Is this a fundamentally different behavior or simply demonstrating a different miter value?

It is the same behavior but with different miter limit value.

---

I started over. I was not really satisfied with the using geometric center because, even though we are not supposed to draw crossed polygons, it can look very wrong using the geometric center instead of looking just a little wrong with the bounding box's center.

Regarding the transition between miter and bevel, I came up with something smooth. Here is what it looks like:

https://github.com/SFML/SFML/assets/1485306/209ee83d-05d1-4f2c-8b00-8a1d1a7b7de2

Next step: adding method to set the miter limit, and also some optimization (vectors are computed twice instead of being reused).

kimci86:
Now works with negative outline too as it was supposed to be.
Added getter and setter for the miter limit.

lapinozz:
https://github.com/SFML/SFML/assets/7329064/0fd9bf49-fc13-4e25-99d5-157964832f18

In paint.net when I reach the miter limit it doesn't "shrink" the end. imo it looks cleaner.
I don't know what kind of algorithm it involves. Just thought it might be something to look at.

kimci86:
@lapinozz Thank you for the suggestion, I like the idea. I just added a commit implementing it.

kimci86:
> Maybe it's too close to all points being collinear and shouldn't be considered a bug?

I don't see a problem in the screenshot. What would you expect?

> There are other somewhat glitchy behaviours around that point.

True... Not sure what can be done about that. At least it is not worse than on master.

> I also think it shouldn't crash for miter limit under 1

What do you suggest? We could clamp the user provided value to be at least 1. Would that deserve printing a warning to the console?

ChrisThrasher (changes requested):

kimci86:
> Does that trigger UB? Does it create visual glitches? Some other problem?

For an angle that is large enough, a miter limit lower than 1 will give a glitchy look. It can trigger assert in Vector2 division operator in debug mode.

I don't really see a valid use case.

Note that `stroke-miterlimit` SVG property [documentation](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-miterlimit#usage_context) mentions that constraint:
> The value of stroke-miterlimit must be greater than or equal to 1.

lapinozz:
> I don't see a problem in the screenshot. What would you expect?

The clipped off corner don't make any sense

kimci86:
> The clipped off corner don't make any sense

The outlined shape is a right triangle with one very small side. For me it makes sense. What result would you expect?

kimci86:
Rebased on master, squashed related commits and updated video of the result.

src/SFML/Graphics/Shape.cpp
```diff
@@ -284,49 +306,87 @@ void Shape::updateTexCoords()
...
+
+            // Combine normals to get bevel edge's direction and normal vector pointing towards the outside of the shape
+            const float    twoSin2   = 1.f - n1.dot(n2);
+            const float    sin       = std::sqrt(twoSin2 / 2.f);
+            const Vector2f direction = (n2 - n1) / twoSin2; // Length is 1 / sin
+            const Vector2f extrusion = (flipNormals != (0.f <= d1.dot(n2)) ? direction : -direction).perpendicular();
+
+            // Compute bevel corner position in (direction, extrusion) coordinates
+            const float u = m_miterLimit * sin;
+            const float v = 1.f - std::sqrt(squaredLengthRatio);
```
ChrisThrasher, line 364:
```suggestion
            // Combine normals to get bevel edge's direction and normal vector pointing towards the outside of the shape
            const float    twoSin2   = 1.f - n1.dot(n2);
            const Vector2f direction = (n2 - n1) / twoSin2; // Length is 1 / sin
            const Vector2f extrusion = (flipNormals != (0.f <= d1.dot(n2)) ? direction : -direction).perpendicular();

            // Compute bevel corner position in (direction, extrusion) coordinates
            const float sin = std::sqrt(twoSin2 / 2.f);
            const float u   = m_miterLimit * sin;
            const float v   = 1.f - std::sqrt(squaredLengthRatio);
```
Lets move `sin` farther down closer to where it's used.

src/SFML/Graphics/Shape.cpp
```diff
@@ -284,49 +306,87 @@ void Shape::updateTexCoords()
...
+        if (needsBevel)
+        {
+            // Make room for two more vertices
+            m_outlineVertices.resize(m_outlineVertices.getVertexCount() + 2);
+
+            // Combine normals to get bevel edge's direction and normal vector pointing towards the outside of the shape
+            const float    twoSin2   = 1.f - n1.dot(n2);
+            const float    sin       = std::sqrt(twoSin2 / 2.f);
+            const Vector2f direction = (n2 - n1) / twoSin2; // Length is 1 / sin
+            const Vector2f extrusion = (flipNormals != (0.f <= d1.dot(n2)) ? direction : -direction).perpendicular();
```
ChrisThrasher, line 360:
```suggestion
            const Vector2f extrusion = (flipNormals != (d1.dot(n2) >= 0.f ) ? direction : -direction).perpendicular();
```

ChrisThrasher (commented):
I finally ran this PR and the results look so good! This feature behaves exactly how I hoped it would. I have a few minor nits about code style but no major comments on the structure the implementation.

Two requests/questions:

What was the rationale for picking `4` as the default value? I think I'd prefer a value closer to `10` judging by some local tests I ran. I think 4 is a bit too aggressive of a default value.

Can you write some `sf::ConvexShape` tests which test the bounds of a shape before and after setting the miter limit so we can ensure that the bounds do in fact change? You can use `https://github.com/SFML/SFML/pull/2762` for inspiration.

include/SFML/Graphics/Shape.hpp
```diff
@@ -127,6 +127,24 @@ class SFML_GRAPHICS_API Shape : public Drawable, public Transformable
...
+    /// the join is converted from a miter to a bevel.
+    /// The miter limit must be greater than or equal to 1.
+    /// By default, the miter limit is 4.
+    ///
+    /// \param miterLimit New miter limit
+    ///
+    /// \see getMiterLimit
+    ///
+    ////////////////////////////////////////////////////////////
+    void setMiterLimit(float miterLimit);
```
ChrisThrasher, line 146: Does this make more sense as a `sf::ConvexShape` feature? Mitering the outline will never be necessary for our prebuilt circle and rectangle shapes. I suppose you may want to miter the edges of a rectangle but that aesthetic consideration is not what we're setting out to solve.

src/SFML/Graphics/Shape.cpp
```diff
@@ -284,49 +306,87 @@ void Shape::updateTexCoords()
...
-        // Update the outline points
-        m_outlineVertices[i * 2 + 0].position = p1;
-        m_outlineVertices[i * 2 + 1].position = p1 + normal * m_outlineThickness;
+        // Compute its direction and normal pointing towards the outside of the shape
+        const auto [d2, n2] = computeDirectionAndNormal(p1, p2, flipNormals);
+
+        // Decide whether to add a bevel or not
+        const float twoCos2            = 1.f + n1.dot(n2);
+        const float squaredLengthRatio = m_miterLimit * m_miterLimit * twoCos2 / 2.f;
+        const bool  isConvexCorner     = 0.f <= d1.dot(n2) * m_outlineThickness;
```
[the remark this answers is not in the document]
kimci86: I find it easier to always use < or <= comparison (it reads from left to right like a sorted sequence of two items), but ok I will change if you want.

src/SFML/Graphics/Shape.cpp
kimci86: It is "used" in a comment but ok.

kimci86:
> What was the rationale for picking 4 as the default value?

I opened Inkscape editor and looked at the default value there. I will change to 10.

include/SFML/Graphics/Shape.hpp
kimci86: It would be more complex to implement and it would be missing from user class derived from `sf::Shape` so I think it is better to keep it in `sf::Shape`.

src/SFML/Graphics/Shape.cpp
```diff
@@ -31,16 +31,22 @@
...
 {
-// Compute the normal of a segment
-sf::Vector2f computeNormal(const sf::Vector2f& p1, const sf::Vector2f& p2)
+// Compute the direction and normal unit vectors of a segment
+std::pair<sf::Vector2f, sf::Vector2f> computeDirectionAndNormal(const sf::Vector2f& p1, const sf::Vector2f& p2, bool flipped)
 {
-    sf::Vector2f normal = (p2 - p1).perpendicular();
-    const float  length = normal.length();
+    sf::Vector2f direction = (p2 - p1);
+    const float  length    = direction.length();
```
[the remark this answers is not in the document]
kimci86: It is probably not very meaningful to have overlapping points, but it is a possible state.
If two consecutive points are identical, their (direction, normal) pair would be a pair of zero vectors. Again not very meaningful but the following computations will not explode.

ChrisThrasher:

ChrisThrasher (approved):

include/SFML/Graphics/Shape.hpp
```diff
@@ -181,6 +199,16 @@ class SFML_GRAPHICS_API Shape : public Drawable, public Transformable
...
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the limit on the ratio between miter length and outline thickness
+    ///
+    /// \return Limit on the ratio between miter length and outline thickness
+    ///
+    /// \see setMiterLimit
+    ///
+    ////////////////////////////////////////////////////////////
+    float getMiterLimit() const;
```
eXpl0it3r, line 210: I want to challenge the "Miter" name usage. I believe it makes on a "domain" level a lot of sense to call it that, but as a simple users that wants to draw shapes, it's not easy to understand what "miter" or "miter limit" means.

Just searching the term "miter" leads to the "tall headdress of a bishop" or "joint made between two pieces of wood". You really have to search for "miter limit" or similar until you start getting documentation from SVG/canvas.

I don't have a better name ready just yet, and am also open to using "miter limit", but maybe we can brainstorm a bit, if there's something that describes it in a way that doesn't necessarily require a documentation study or web search.

include/SFML/Graphics/Shape.hpp
```diff
@@ -181,6 +199,16 @@ class SFML_GRAPHICS_API Shape : public Drawable, public Transformable
     ////////////////////////////////////////////////////////////
     float getOutlineThickness() const;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the limit on the ratio between miter length and outline thickness
+    ///
+    /// \return Limit on the ratio between miter length and outline thickness
```
eXpl0it3r, line 205: This documentation could need a bit more explanation on what "miter" is. If I don't know what "miter" is, the documentation essentially explains things by self reference as one also wouldn't understand what "miter length" is.

include/SFML/Graphics/Shape.hpp
ChrisThrasher: The PR name uses the word `bevel` which I think is a good descriptor. This parameter controls how far the outline corner can stick out before it starts being beveled.

src/SFML/Graphics/Shape.cpp
```diff
@@ -129,6 +135,22 @@ float Shape::getOutlineThickness() const
 }
 
 
+////////////////////////////////////////////////////////////
+void Shape::setMiterLimit(float miterLimit)
+{
+    assert(miterLimit >= 1.f && "Shape::setMiterLimit(float) cannot set miter limit to a value lower than 1");
+    m_miterLimit = std::max(1.f, miterLimit);
```
eXpl0it3r, line 142: Not sure if we should write "defensive" code like that. Yes, it will warn in debug mode, but it will also silently "work" in release mode when you set a value less than 1.

What's the effect of allowing negative numbers? Would it crash or just render distorted?

src/SFML/Graphics/Shape.cpp
```diff
@@ -31,16 +31,22 @@
...
 
+#include <cassert>
+#include <cmath>
+
 namespace
 {
-// Compute the normal of a segment
-sf::Vector2f computeNormal(const sf::Vector2f& p1, const sf::Vector2f& p2)
+// Compute the direction and normal unit vectors of a segment
+std::pair<sf::Vector2f, sf::Vector2f> computeDirectionAndNormal(const sf::Vector2f& p1, const sf::Vector2f& p2, bool flipped)
```
eXpl0it3r, line 40: I'm not particularly a fan of boolean parameters. I think they can exist if they have a major impact on the calculation and it would cost a lot maintenance-wise to duplicate the code, but here we're just flipping one of the returned results.

So the consumer of the function defines the output of the function, which it just as well could do on the consuming end of the function.

If anything, it should have at least a default value for `flipped`.

src/SFML/Graphics/Shape.cpp
```diff
@@ -284,49 +306,87 @@ void Shape::updateTexCoords()
...
+            // Compute bevel corner position in (direction, extrusion) coordinates
+            const float sin = std::sqrt(twoSin2 / 2.f);
+            const float u   = m_miterLimit * sin;
+            const float v   = 1.f - std::sqrt(squaredLengthRatio);
+
+            // Update the outline points
+            m_outlineVertices[outlineIndex++].position = p1;
+            m_outlineVertices[outlineIndex++].position = p1 + (u * extrusion - v * direction) * m_outlineThickness;
+            m_outlineVertices[outlineIndex++].position = p1;
+            m_outlineVertices[outlineIndex++].position = p1 + (u * extrusion + v * direction) * m_outlineThickness;
```
eXpl0it3r, line 370: Do you have a resource link which you could reference as a comment that explains the math some more, so mere mortals like me can find some more information further down the line? 😅

eXpl0it3r (dismissed):
Just a few small comments 😃

include/SFML/Graphics/Shape.hpp
kimci86: I believe "miter limit" it is the correct name because it is already used by SVG as well as [cairo library](https://www.cairographics.org/manual/cairo-cairo-t.html#cairo-set-miter-limit). It would be more confusing to come up with some other name. I agree the meaning is not obvious and it deserves more documentation. I will work on it.

eXpl0it3r: I understand that it's used like that within the domain of SVG/graphics, but we shouldn't just stop there and do as everyone else does. The term is not intuitively understood and as such it deserves a bit of a brainstorming in my opinion.

src/SFML/Graphics/Shape.cpp
kimci86: Miter limit lower than 1 is asking outline not to grow bigger around angle than the outline thickness. Basically the limit is always reached. It makes little sense.
Having miter limit lower than 1 gives strange look and can run into Vector2 assert about division by 0.

Here are some example of what it looks like starting at 1 (the lowest valid limit) and going down. The result is especially wrong when angles are large enough (3rd example).

miter limit

src/SFML/Graphics/Shape.cpp
kimci86: I worked it out on a whiteboard. Not sure where to look for material on this specific problem.

eXpl0it3r: Hehe okay, I mean that's valid as well. Was just thinking, maybe you had some formula mentioned somewhere on the web that we could link.

ChrisThrasher: Is it feasible to write out the formula as a code comment?

src/SFML/Graphics/Shape.cpp
kimci86: I made code simple and removed the boolean parameter :+1:

ChrisThrasher:

src/SFML/Graphics/Shape.cpp
eXpl0it3r: @Bromeon Since we've had discussion about bad input handling, what's your view on doing assert + limit?

include/SFML/Graphics/Shape.hpp
eXpl0it3r: I asked a bit on social media, to see what people would more intuitively call this.

- Two mentioned miter, but more because they've learned the word from graphics applications.
- sanitizeOutline
- setOutlineExtensionLimit
- limitOutlineExtension
- clamp
- cutting corners
- corner style
- capping (used in Godot)
- Corner length
- Cap
- Clip
- Limit
- bevel

https://swiss.social/@darkcisum/111391908664038204 
https://twitter.com/DarkCisum/status/1723282675878789343

One interesting thought that came out of these discussions is that this could in theory be generalized further and allow for different "styling" of the corners. I wouldn't to do this right now of course, but it might influence the thinking of naming.

src/SFML/Graphics/Shape.cpp
ChrisThrasher: I'm not sure if we do `assert` + limit anywhere else. It may be more in line with other APIs if we assert but otherwise allow the original value to be used. For example `sf::WindowBase::setMinimumSize` asserts a valid minimum size (it can't be large than the maximum window size) but it doesn't clamp the minimum size to be valid. The invalid value continues to be used and whatever happens after that is unspecified.

Bromeon: Difficult question. In an ideal world, the developer would catch the assert during development.

But _if_ it slips through, it's probably better if it doesn't cause catastrophic damage.
However, "distorted rendering" is not catastrophic in my opinion (on the contrary, it could _increase_ chances of finding such bugs).

This is a hard problem, and I haven't seen any library/framework/engine handle errors always gracefully. Attempts to do so often result in excessively verbose error handling code, and other approaches (exceptions, panics, aborts) often crash the app/game for minuscule bugs.

kimci86: Ok, I removed the `std::max` call.

include/SFML/Graphics/Shape.hpp
kimci86: I tried to make is clearer adding documentation to `setMiterLimit` method. Feedback is welcome.

kimci86:
Rebased on master and cleaned up history.

wiiqwertyuiop:
So I've been noticing some weirdness with this still. I have this branch pulled, and using it in this project. It yields this result:

https://github.com/SFML/SFML/assets/33212768/6af3fcd3-7e91-45b9-9f74-923588176a4c

As you can see there are still some jagged edges, which I would not expect. Previously before this, I was having the issue where they would go across the entire screen and was much worse.

For a comparison, here is what it looks like using sf::Vertex:


https://github.com/SFML/SFML/assets/33212768/10d6b485-1893-4e3e-8cea-9aecd2d12261

This is how I would expect the shapes to function/look as well, but as you can see there is a big difference here.

If there is any interest in the code, I am just using this: https://github.com/wiiqwertyuiop/3DProjector

Hapaxia:
Since I rarely use sf::Shape and pretty much never use sf::ConvexShape, I wasn't aware this issue exists. Adding an outline to a sf::Shape is effectively creating a thick polyline (spline) around that shape and there are a few issues I've encountered with thick polylines. I'll try to recall what I can and share some insight if I remember anything of use.

As far naming, I'd say, "bevel" is a better name (for corner cutting) and it's certainly what I've experienced most of the time when working with vector graphics and their editors.

kimci86:
Rebased and updated example code in PR description with new events API.

ChrisThrasher:

ChrisThrasher:
I re-ran the example program and this looks good! I'm fine with either the terms "bezel" or "miter" so I'll let Exploiter make the final call on naming.

How confident are we that this won't introduce performance issues? If a user sets the limit to an especially high value to effectively turn this feature off, what additional time is spent on miters, if any? I'm not asking you to write benchmarks but I want to get a better idea of what this may cost to calculate.

ChrisThrasher:

kimci86:
> How confident are we that this won't introduce performance issues?

I had a closer look at performance implications. The fix about normal vector orientation was potentially expensive, computing the shape area. I implemented and pushed a cheaper alternative.
The other changes seem to have no significant impact on performance, when corners are not cut off.

ChrisThrasher:

ChrisThrasher (approved):
I think I'm good with this but I want @eXpl0it3r's buy-in before we merge.

ChrisThrasher:
I want to merge this sooner rather than later because it fixes some real-world bugs while adding a feature that we're eager to ship in v3.1. If we would like to tweak naming before we ship it, I'm okay with that.

PR #2745, opened 2023-10-23
Specify `CMAKE_RUNTIME_OUTPUT_DIRECTORY` from outside of the project

ChrisThrasher:
## Description

Hardcoding this into the library itself means that outside consumers cannot choose a different value. This in particular breaks the CMake template project of ours which also tries to set this value. It resulted in the DLLs not being placed in build/bin as expected but rather build/_deps/sfml-build/bin.

As conventional wisdom dictates, it's best to not touch `CMAKE_` variables unless necessary. Luckily our use of presents means we can easy move this into the development preset so there are no ergonomics disadvantages to this.

This problem is not present in SFML 2 because that version does not manipulate `CMAKE_RUNTIME_OUTPUT_DIRECTORY`. Luckily now that the template project defaults to static libs these DLL issues are harder to run into.

Thanks to @Bambo-Borris for reporting this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -216,9 +216,6 @@ if(SFML_OS_MACOS)
     endif()
 endif()
 
-# set the output directory for SFML DLLs and executables
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
-
 # enable project folders
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 set_property(GLOBAL PROPERTY PREDEFINED_TARGETS_FOLDER "CMake")
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -8,6 +8,7 @@
       "cacheVariables": {
         "CMAKE_CXX_EXTENSIONS": "OFF",
         "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
+        "CMAKE_RUNTIME_OUTPUT_DIRECTORY": "${sourceDir}/build/bin",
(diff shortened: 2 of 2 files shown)


Bambo-Borris:
Found this whilst using the template project, altered to target master. I regularly use the current state of master. I noticed that the DLLs weren't being output to the expected directory `build_directory/bin` when using master but were when using 2.6.0. Thanks to @ChrisThrasher for the quick spot and potential resolution for this

eXpl0it3r:
What do you want to do about buildbot?

ChrisThrasher:
The Mesa3D stuff hardcode a reference to `${PROJECT_BINARY_DIR}/bin`. To make this PR work I think I need to figure out a more generic solution that doesn't depend on using that particular directory.

JonnyPtn:
Better not to set it at all - as you said it's typically best practice to leave these for users to set themselves, and the DLL problem can be solved with `TARGET_RUNTIME_DLLS`

CesarBerriot:
Just ran into this issue after integrating SFML as a dynamic library using FetchContent preceded by `set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})`. This works flawlessly for CSFML but forces SFML dynamic binaries users to add something along the lines of the following to their targets :
```cmake
add_custom_command(TARGET hello_world POST_BUILD COMMAND "${CMAKE_COMMAND}" -E copy_directory "$<TARGET_FILE_DIR:sfml-system>" "${CMAKE_BINARY_DIR}")
```
Letting consumers set the value of `CMAKE_RUNTIME_OUTPUT_DIRECTORY` would greatly improve SFML's integration when built dynamically.

PR #2749, opened 2023-10-24
Replace audio backend with miniaudio

binary1248:
By popular demand.

Work in progress.

Here be dragons.


ChrisThrasher:
https://github.com/SFML/SFML/compare/master...miniaudio_rebase

I took a shot at rebasing and fixing merge conflicts. Some of the conflicts were easily fixed but I'm still not sure I correctly fixed the conflicts in src/SFML/Audio/Sound.cpp. I can't get it to compile.

ChrisThrasher:
I have most of the build errors fixed and removed all the imgui and imgui-sfml source code. I still need to fix compiler warnings and clang-tidy errors but this PR is at least in a state where it passes most of CI which means more people should be able to test it out.

binary1248:
miniaudio officially supports iOS and building for it even has a [dedicated subsection in its documentation](https://miniaud.io/docs/manual/index.html#Building).

ChrisThrasher:
Screenshot 2023-10-24 at 11 35 22 PM

I think my hunch was right. This should be fixable by telling CMake what specific files need to be compiled as Objective-C++.

ChrisThrasher:
I've fixed almost all of the compiler warnings. The last one to address is [here](https://github.com/SFML/SFML/actions/runs/6659820044/job/18099722291) where MinGW appears to have found a null dereference inside of mini audio. I'm not sure if this warning is triggered by code we wrote or what. 

We'll probably just `pragma` around this but I wanted to highlight it in case someone had an idea for a different fix.

Lastly I still have to fix those iOS jobs and then'll have 100% green CI :)

ChrisThrasher:
I was able to fix the iOS build errors and added and workaround to deal with that MinGW warning that I didn't otherwise know how to fix. CI should be 100% green now!

src/SFML/Audio/SoundBuffer.cpp
```diff
@@ -165,59 +181,56 @@ bool SoundBuffer::saveToFile(const std::filesystem::path& filename) const
...
 
 ////////////////////////////////////////////////////////////
 SoundBuffer& SoundBuffer::operator=(const SoundBuffer& right)
 {
-    SoundBuffer temp(right);
-
-    std::swap(m_samples, temp.m_samples);
-    std::swap(m_buffer, temp.m_buffer);
-    std::swap(m_duration, temp.m_duration);
+    const SoundBuffer& temp(right);
```
lapinozz, line 226: Is this supposed to be a reference?

src/SFML/Audio/SoundFileReaderWav.cpp
```diff
@@ -307,6 +307,63 @@ bool SoundFileReaderWav::parseHeader(Info& info)
...
+                static constexpr auto SPEAKER_TOP_FRONT_CENTER      = 0x2000u;
+                static constexpr auto SPEAKER_TOP_FRONT_RIGHT       = 0x4000u;
+                static constexpr auto SPEAKER_TOP_BACK_LEFT         = 0x8000u;
+                static constexpr auto SPEAKER_TOP_BACK_CENTER       = 0x10000u;
+                static constexpr auto SPEAKER_TOP_BACK_RIGHT        = 0x20000u;
+                // NOLINTEND(readability-identifier-naming)
+
+                info.channelMap.clear();
+
+                auto checkChannel = [&](auto bit, auto soundChannel)
```
lapinozz, line 334: I didn't see in the code style guideline if we always const variables when possible but if so maybe consider a const here.

src/SFML/Audio/SoundBuffer.cpp
ChrisThrasher: clang-tidy warned about this not being a `const&` so I added that but it did confuse me too. Swapping out of `const&` doesn't make sense to me.

include/SFML/Audio/Sound.hpp
```diff
@@ -228,10 +230,19 @@ class SFML_AUDIO_API Sound : public SoundSource
...
+    ///
+    ////////////////////////////////////////////////////////////
+    void* getSound() const override;
+
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    const SoundBuffer* m_buffer{}; //!< Sound buffer bound to the source
+    struct Impl;
+    const std::unique_ptr<Impl> m_impl; //!< Implementation details
```
ChrisThrasher, line 245: Why hide the implementation behind this layer indirection rather than put all of `Sound::Impl`'s data members inside the class directly?

The one reason I can think of doing that is to keep miniaudio headers out of of Sound.hpp.

include/SFML/Audio/SoundBuffer.hpp
```diff
@@ -127,16 +126,18 @@ class SFML_AUDIO_API SoundBuffer : AlResource
...
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool loadFromSamples(const std::int16_t* samples,
-                                       std::uint64_t       sampleCount,
-                                       unsigned int        channelCount,
-                                       unsigned int        sampleRate);
+    [[nodiscard]] bool loadFromSamples(const std::int16_t*              samples,
+                                       std::uint64_t                    sampleCount,
+                                       unsigned int                     channelCount,
+                                       unsigned int                     sampleRate,
+                                       const std::vector<SoundChannel>& channelMap);
```
kimci86, line 140: Having both `channelCount` and `channelMap` seems redundant as `channelCount` can be derived from `channelMap.size()`. Same remark applies to other places which have some `channelMap`.

src/SFML/Audio/CMakeLists.txt
```diff
@@ -57,18 +56,21 @@ set(CODECS_SRC
...
 
-# let CMake know about our additional audio libraries paths (on Windows and macOS)
-if(SFML_OS_WINDOWS)
-    list(APPEND CMAKE_INCLUDE_PATH "${PROJECT_SOURCE_DIR}/extlibs/headers/AL")
-elseif(SFML_OS_MACOS)
+# Ensure certain files are compiled as Objective-C++
+# See: https://miniaud.io/docs/manual/index.html#Building
+if(SFML_OS_IOS)
+    enable_language(OBJCXX)
+    set_source_files_properties(${SRCROOT}/AudioDevice.cpp PROPERTIES LANGUAGE OBJCXX)
```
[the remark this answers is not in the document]
ChrisThrasher: I think I'd be okay with that since this is the kind of header-only library that requires defining its implementation in just one file.

ChrisThrasher:
Rebased on `master` and applied a few changes recommended by Kimci.

src/SFML/Audio/CMakeLists.txt
```diff
@@ -57,18 +57,21 @@ set(CODECS_SRC
...
-elseif(SFML_OS_MACOS)
+# Ensure certain files are compiled as Objective-C++
+# See: https://miniaud.io/docs/manual/index.html#Building
+if(SFML_OS_IOS)
+    enable_language(OBJCXX)
+    set_source_files_properties(${SRCROOT}/Miniaudio.cpp PROPERTIES LANGUAGE OBJCXX)
+endif()
+
+# let CMake know about our additional audio libraries paths (on Android and OSX)
+if(SFML_OS_MACOSX)
```
kimci86, line 68:
```suggestion
# let CMake know about our additional audio libraries paths (on Android and macOS)
if(SFML_OS_MACOS)
```

src/SFML/Audio/CMakeLists.txt
```diff
@@ -77,19 +80,28 @@ sfml_add_library(Audio
                  SOURCES ${SRC} ${CODECS_SRC})
 
 # avoids warnings in vorbisfile.h
-target_compile_definitions(sfml-audio PRIVATE OV_EXCLUDE_STATIC_CALLBACKS FLAC__NO_DLL)
+target_compile_definitions(sfml-audio PRIVATE OV_EXCLUDE_STATIC_CALLBACKS FLAC__NO_DLL MA_NO_DECODING MA_NO_ENCODING)
```
kimci86, line 83:
```suggestion
# avoids warnings in vorbisfile.h
target_compile_definitions(sfml-audio PRIVATE OV_EXCLUDE_STATIC_CALLBACKS FLAC__NO_DLL)
# disables miniaudio features we do not use
target_compile_definitions(sfml-audio PRIVATE MA_NO_DECODING MA_NO_ENCODING)
```

src/SFML/Audio/CMakeLists.txt
```diff
@@ -65,7 +65,7 @@ if(SFML_OS_IOS)
 endif()
 
 # let CMake know about our additional audio libraries paths (on Android and OSX)
```
kimci86, line 67: That part of my previous suggestion was overlooked:
```suggestion
# let CMake know about our additional audio libraries paths (on Android and macOS)
```

include/SFML/Audio/SoundSource.hpp
```diff
@@ -55,6 +58,24 @@ class SFML_AUDIO_API SoundSource : AlResource
...
+    /// positioned outside the outer angle of the cone.
+    /// The gain declines linearly from 1 to outerGain as the
+    /// listener moves from the inner angle to the outer angle.
+    ///
+    ////////////////////////////////////////////////////////////
+    struct Cone
+    {
+        Angle innerAngle;  //!< Inner angle, default is 360 degrees
+        Angle outerAngle;  //!< Outer angle, default is 360 degrees
+        float outerGain{}; //!< Outer gain, default is 1
```
kimci86, line 76: Same remark as Listener::Cone documentation: documented default values do not match default-constructed Cone values.

src/SFML/Audio/SoundSource.cpp
```diff
@@ -189,22 +383,12 @@ SoundSource& SoundSource::operator=(const SoundSource& right)
...
-            return Stopped;
-        case AL_PAUSED:
-            return Paused;
-        case AL_PLAYING:
-            return Playing;
-    }
+    if (const auto* sound = static_cast<const ma_sound*>(getSound()); sound && (ma_sound_is_playing(sound) == MA_TRUE))
+        return Playing;
 
     return Stopped;
```
kimci86, line 389: Issue here: `Status::Paused` is never returned.

src/SFML/Audio/AudioDevice.cpp
```diff
@@ -25,211 +25,348 @@
...
+    {
+        m_impl->m_context.reset();
+        err() << "Failed to initialize the audio context: " << ma_result_description(result) << std::endl;
+        return;
+    }
+
+    // Create the playback device
+    m_impl->m_playbackDevice.emplace();
+
+    auto playbackDeviceConfig         = ma_device_config_init(ma_device_type_playback);
```
kimci86, line 137: The `sound` example is very noisy on my machine, but it sounds good if I set `playbackDeviceConfig.playback.format = ma_format_f32;` here.

include/SFML/Audio/Music.hpp
```diff
@@ -244,10 +248,8 @@ class SFML_AUDIO_API Music : public SoundStream
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    InputSoundFile            m_file;     //!< The streamed music file
-    std::vector<std::int16_t> m_samples;  //!< Temporary buffer of samples
-    std::recursive_mutex      m_mutex;    //!< Mutex protecting the data
-    Span<std::uint64_t>       m_loopSpan; //!< Loop Range Specifier
+    struct Impl;
+    const std::unique_ptr<Impl> m_impl; //!< Implementation details
```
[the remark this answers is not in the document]
ChrisThrasher: I'd also love to stop using pimpl but I was under the impression it would require adding mini audio headers to AudioDevice.hpp. How is that not true? We need to know the definitions of `ma_log` and other miniaudio types.

src/SFML/Audio/SoundBuffer.cpp
ChrisThrasher: I think it's actually correct. The thing being swapped is the pointed-to data owned by `temp.m_impl`. `temp` and `m_impl` don't actually change.

lapinozz: Why even have the temp variable at all then? And I guess it would mean the comment about temp being destructed doesn't make sense?

include/SFML/Audio/Music.hpp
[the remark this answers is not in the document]
kimci86: We need pimpl sometimes because we want miniaudio to be remain an implementation detail and not leak from public header, but in the case of Music and SoundBuffer (and maybe more), miniaudio is not involved at all (unless I missed something).

src/SFML/Audio/SoundBuffer.cpp
kimci86: This is indeed moving from temp.m_impl but temp is the same instance as right, so this is moving from right.m_impl, which is not correct for a copy constructor.

include/SFML/Audio/Music.hpp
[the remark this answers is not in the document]
ChrisThrasher: My bad I was looking at AudioDevice.

ChrisThrasher:
There are some new clang-tidy errors because the implementation structs use an `m_` prefix in spite of being `struct`s with exclusively `public` data.

ChrisThrasher:
Adding ImGUi-SFML creates a circular dependency. ImGui-SFML depends on SFML as shown by `ImGui-SFML::ImGui-SFML` linking to `SFML::Graphics`. However SFML is now downloading ImGui-SFML.

If we break the API and ImGui-SFML needs updating as sometimes happens, then we can’t merge the PR because SFML CI will fail. However if we try to fix this API change in ImGui-SFML first then ImGui-SFML CI will fail because SFML hasn’t yet merged this change. That feels a bit messy to me. As long as we don't break the API we're fine but it's not great that this design doesn't withstand SFML API changes gracefully.

I'm questioning whether it's worth adding ImGui-SFML just for an example. I get its value. It's useful to have sliders and other GUI elements on screen but it comes at the cost of some annoying new complexities that we could avoid by simplifying the example. Perhaps simplifying this example to work with keyboard controls is a cleaner solution.

ChrisThrasher:
This PR increases the clean build time on my machine from ~20 seconds to ~25 seconds. Most of that can be attribute to us downloading and building ImGui-SFML for the new example. I'm not yet sure if this also means a ~25% increase to how long CI jobs take since we do lots of caching in CI.

vittorioromeo:
@eXpl0it3r @binary1248: I've rebased on the current `master` branch and I've also removed a lot of code repetition between `SoundStream.cpp` and `Sound.cpp` by introducing a `src/MiniaudioUtils.{h,cpp}` utility component.

I've kept my commit separate for now so that my refactoring can be reviewed.

include/SFML/Audio/Sound.hpp
vittorioromeo: > The one reason I can think of doing that is to keep miniaudio headers out of of Sound.hpp.

Yup, and I think that's a good decision, even if it introduces one extra layer of indirection.

include/SFML/Audio/InputSoundFile.hpp
```diff
@@ -119,6 +120,19 @@ class SFML_AUDIO_API InputSoundFile
...
+    ///
+    /// This is used to map a sample in the sample stream to a
+    /// position during spatialisation.
+    ///
+    /// \return Map of position in sample frame to sound channel
+    ///
+    /// \see getSampleRate, getChannelCount, getDuration
+    ///
+    ////////////////////////////////////////////////////////////
+    std::vector<SoundChannel> getChannelMap() const;
```
vittorioromeo, line 134: Can we make this return a `const std::vector<SoundChannel>&`? I see that `SoundRecorder::getChannelMap()` creates one on the spot, but maybe we could just cache it as a data member in `SoundRecorder` to support this more efficient API?

src/SFML/Audio/SoundFileReaderFlac.cpp
```diff
@@ -172,6 +172,69 @@ void streamMetadata(const FLAC__StreamDecoder*, const FLAC__StreamMetadata* meta
...
+                                         sf::SoundChannel::BackLeft,
+                                         sf::SoundChannel::BackRight};
+                break;
+            case 6:
+                data->info.channelMap = {sf::SoundChannel::FrontLeft,
+                                         sf::SoundChannel::FrontRight,
+                                         sf::SoundChannel::FrontCenter,
+                                         sf::SoundChannel::LowFrequencyEffects,
+                                         sf::SoundChannel::BackLeft,
+                                         sf::SoundChannel::BackRight};
```
vittorioromeo, line 212: I am assuming that the order of these channels matters, and that's why these three huge switches cannot be refactored to a separate function. Is that right?

src/SFML/Audio/SoundFileWriterWav.cpp
```diff
@@ -82,8 +84,129 @@ SoundFileWriterWav::~SoundFileWriterWav()
...
+        targetChannelMap.push_back(SupportedChannel{SPEAKER_BACK_CENTER, SoundChannel::BackCenter});
+        targetChannelMap.push_back(SupportedChannel{SPEAKER_SIDE_LEFT, SoundChannel::SideLeft});
+        targetChannelMap.push_back(SupportedChannel{SPEAKER_SIDE_RIGHT, SoundChannel::SideRight});
+        targetChannelMap.push_back(SupportedChannel{SPEAKER_TOP_CENTER, SoundChannel::TopCenter});
+        targetChannelMap.push_back(SupportedChannel{SPEAKER_TOP_FRONT_LEFT, SoundChannel::TopFrontLeft});
+        targetChannelMap.push_back(SupportedChannel{SPEAKER_TOP_FRONT_CENTER, SoundChannel::TopFrontCenter});
+        targetChannelMap.push_back(SupportedChannel{SPEAKER_TOP_FRONT_RIGHT, SoundChannel::TopFrontRight});
+        targetChannelMap.push_back(SupportedChannel{SPEAKER_TOP_BACK_LEFT, SoundChannel::TopBackLeft});
+        targetChannelMap.push_back(SupportedChannel{SPEAKER_TOP_BACK_CENTER, SoundChannel::TopBackCenter});
+        targetChannelMap.push_back(SupportedChannel{SPEAKER_TOP_BACK_RIGHT, SoundChannel::TopBackRight});
```
vittorioromeo, line 157: We could use initializer list initialization here and avoid the individual push backs.

src/SFML/Audio/SoundSource.cpp
```diff
@@ -25,155 +25,349 @@
...
 }
 
 
 ////////////////////////////////////////////////////////////
-SoundSource::~SoundSource()
+void SoundSource::setPitch(float pitch)
 {
-    alCheck(alSourcei(m_source, AL_BUFFER, 0));
-    alCheck(alDeleteSources(1, &m_source));
+    if (auto* sound = static_cast<ma_sound*>(getSound()); sound)
```
vittorioromeo, line 63:
```suggestion
    if (auto* sound = static_cast<ma_sound*>(getSound()))
```

Ditto for all other occurrences

vittorioromeo (changes requested):

src/SFML/Audio/SoundSource.cpp
vittorioromeo: Applied this suggestion myself.

src/SFML/Audio/SoundFileWriterWav.cpp
vittorioromeo: Applied this suggestion myself.

src/SFML/Audio/SoundFileReaderFlac.cpp
binary1248: Yes, FLAC specifies channel ordering based on the number of channels.

https://xiph.org/flac/format.html#frame_header

include/SFML/Audio/InputSoundFile.hpp
binary1248: Done.

binary1248:
Addressed comments and replaced sound stage with sound effects example that no longer relies on ImGui.

src/SFML/Audio/MiniaudioUtils.cpp
```diff
@@ -0,0 +1,264 @@
...
+    ma_sound_set_cone(&sound, degrees(360).asRadians(), degrees(360).asRadians(), 0.f); // inner = 360 degrees, outer = 360 degrees, gain = 0
+    ma_sound_set_directional_attenuation_factor(&sound, 1.f);
+    ma_sound_set_velocity(&sound, 0.f, 0.f, 0.f);
+    ma_sound_set_doppler_factor(&sound, 1.f);
+    ma_sound_set_positioning(&sound, ma_positioning_absolute);
+    ma_sound_set_min_distance(&sound, 1.f);
+    ma_sound_set_max_distance(&sound, std::numeric_limits<float>::max());
+    ma_sound_set_min_gain(&sound, 0.f);
+    ma_sound_set_max_gain(&sound, 1.f);
+    ma_sound_set_rolloff(&sound, 1.f);
```
ChrisThrasher, line 190: A simpler option would be to provide default values to the data in `SavedSettings`. Then this function can be implanted as one line that simply calls `applySettings` with a default-constructed `SavedSettings`.

ChrisThrasher (commented):
I love the new sound effects example. I have no structural issues with this PR. I just pushed some style fixes and made sure to apply them to the appropriate commits. 

The one blocker is that on macOS the sound effects example emits no sound. Any recommendations for how I can go about debugging this? I've already tried playing the sound file by itself through another program and it works fine.

ChrisThrasher:
If I run sound-effects and press right arrow about once per second to cycle through its states, I quickly get this UBSan error
```
UndefinedBehaviorSanitizer:DEADLYSIGNAL
==98515==ERROR: UndefinedBehaviorSanitizer: SEGV on unknown address 0x00012da7f780 (pc 0x0001038f567c bp 0x00016d5bcfb0 sp 0x00016d5bcf80 T5810290)
==98515==The signal is caused by a READ memory access.
    #0 0x1038f567c in vorbis_synthesis+0x1b0 (vorbis:arm64+0xd67c)
    #1 0x1038d8fdc in _fetch_and_process_packet+0xbc (vorbisfile:arm64+0x4fdc)
    #2 0x1038d97a4 in ov_read_filter+0x90 (vorbisfile:arm64+0x57a4)
    #3 0x1038d9abc in ov_read+0x14 (vorbisfile:arm64+0x5abc)
    #4 0x102d04b6c in sf::priv::SoundFileReaderOgg::read(short*, unsigned long long) SoundFileReaderOgg.cpp:199
    #5 0x102c68268 in sf::InputSoundFile::read(short*, unsigned long long) InputSoundFile.cpp:261
    #6 0x102c5ff80 in sf::Music::onGetData(sf::SoundStream::Chunk&) Music.cpp:189
    #7 0x102c81204 in sf::SoundStream::Impl::read(void*, void*, unsigned long long, unsigned long long*) SoundStream.cpp:118
    #8 0x102bac92c in ma_data_source_read_pcm_frames_within_range(void*, void*, unsigned long long, unsigned long long*) miniaudio.h:57163
    #9 0x102baaedc in ma_data_source_read_pcm_frames miniaudio.h:57278
    #10 0x102c574d8 in ma_engine_node_process_pcm_frames__sound(void*, float const**, unsigned int*, float**, unsigned int*) miniaudio.h:74347
    #11 0x102c5429c in ma_node_process_pcm_frames_internal(void*, float const**, unsigned int*, float**, unsigned int*) miniaudio.h:72453
    #12 0x102bd8684 in ma_node_read_pcm_frames(void*, unsigned int, float*, unsigned int, unsigned int*, unsigned long long) miniaudio.h:72557
    #13 0x102c54bc8 in ma_node_input_bus_read_pcm_frames(void*, ma_node_input_bus*, float*, unsigned int, unsigned int*, unsigned long long) miniaudio.h:71701
    #14 0x102bd8a10 in ma_node_read_pcm_frames(void*, unsigned int, float*, unsigned int, unsigned int*, unsigned long long) miniaudio.h:72575
    #15 0x102bd7c50 in ma_node_graph_read_pcm_frames miniaudio.h:71232
    #16 0x102bf2a78 in ma_engine_read_pcm_frames miniaudio.h:75207
    #17 0x102aa4cfc in sf::priv::AudioDevice::AudioDevice()::$_1::operator()(ma_device*, void*, void const*, unsigned int) const AudioDevice.cpp:144
    #18 0x102aa4b20 in sf::priv::AudioDevice::AudioDevice()::$_1::__invoke(ma_device*, void*, void const*, unsigned int) AudioDevice.cpp:138
    #19 0x102c31754 in ma_device__on_data_inner(ma_device*, void*, void const*, unsigned int) miniaudio.h:18675
    #20 0x102c30938 in ma_device__on_data(ma_device*, void*, void const*, unsigned int) miniaudio.h:18756
    #21 0x102c2de6c in ma_device__handle_data_callback(ma_device*, void*, void const*, unsigned int) miniaudio.h:18808
    #22 0x102ae3b1c in ma_device__read_frames_from_client(ma_device*, unsigned int, void*) miniaudio.h:18838
    #23 0x102c2d0f8 in ma_device_audio_thread__default_read_write(ma_device*) miniaudio.h:19328
    #24 0x102adcfec in ma_worker_thread(void*) miniaudio.h:40874
    #25 0x102bfdff4 in ma_thread_entry_proxy(void*) miniaudio.h:16494
    #26 0x18318af90 in _pthread_start+0x84 (libsystem_pthread.dylib:arm64e+0x6f90)
    #27 0x183185d30 in thread_start+0x4 (libsystem_pthread.dylib:arm64e+0x1d30)

==98515==Register values:
 x[0] = 0x0000000437460000   x[1] = 0x0000000000000000   x[2] = 0x0000000137460000   x[3] = 0x000000016d5bce80  
 x[4] = 0x0000000003000001   x[5] = 0x0000000000000003   x[6] = 0x0000000000000000   x[7] = 0x0000000000000000  
 x[8] = 0xfffffffffb67fb56   x[9] = 0x000000012da7f558  x[10] = 0x0000000000000018  x[11] = 0x0000000000000252  
x[12] = 0x000000011e4b77b0  x[13] = 0x0000000000000000  x[14] = 0x0000000000007f73  x[15] = 0x0000000000000000  
x[16] = 0xfffffffffffffff1  x[17] = 0x00000001f52030f0  x[18] = 0x0000000000000000  x[19] = 0x000000013f7534b8  
x[20] = 0x0000000000000002  x[21] = 0x000000013f7534c0  x[22] = 0x000000013f7535a0  x[23] = 0x0000000140080800  
x[24] = 0x0000000140080828  x[25] = 0x000000013f753280  x[26] = 0x0000000000000038  x[27] = 0xffffffffffffffff  
x[28] = 0x0000000000000004     fp = 0x000000016d5bcfb0     lr = 0x00000001038f5658     sp = 0x000000016d5bcf80  
UndefinedBehaviorSanitizer can not provide additional info.
SUMMARY: UndefinedBehaviorSanitizer: SEGV (vorbis:arm64+0xd67c) in vorbis_synthesis+0x1b0
==98515==ABORTING
Abort trap: 6
```

If I spam inputs in slightly different ways I can get other errors. Not sure how these errors relate to each other but it points to there being something going on with how this example program manages changing states.

binary1248:
Updated.

Tested on Windows and Linux (Manjaro). Does not trigger any sanitizers and @texus example terminates cleanly.

src/SFML/Audio/AudioDevice.cpp
binary1248: The miniaudio implementation seems to be a bit bugged when using any other device format but `ma_format_f32`. The format is forced to `ma_format_f32` in the latest commit.

include/SFML/Audio/Listener.hpp
```diff
@@ -39,6 +40,24 @@
...
+/// outer angle of the cone. The gain declines linearly
+/// from 1 to outerGain as the sound moves from the inner
+/// angle to the outer angle.
+///
+////////////////////////////////////////////////////////////
+struct Cone
+{
+    Angle innerAngle;  //!< Inner angle, default is 360 degrees
+    Angle outerAngle;  //!< Outer angle, default is 360 degrees
+    float outerGain{}; //!< Outer gain, default is 1
```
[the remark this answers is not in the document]
binary1248: Should be fixed now.

src/SFML/Audio/SoundSource.cpp
binary1248: Fixed.

include/SFML/Audio/SoundBuffer.hpp
binary1248: Fixed.

binary1248:
Fixed issues mentioned by @kimci86.

texus:
I can confirm that the mentioned issues are fixed (tested on Linux). The noise is gone, all the effects in the example work and it no longer crashes.

ChrisThrasher:
The crashes seem to be fixed now. However I still cannot get any audio to play. Can anyone else try this example on macOS and let me know if you hear anything?

binary1248:
I updated the miniaudio header to a newer version, it includes a fix for an issue that users reported causing audio not to play on macOS systems.

I also made the console output when running sound code without an available playback device a bit more helpful.

If you uncomment `MA_DEBUG_OUTPUT` in Miniaudio.cpp, miniaudio should dump a bunch of information to the console to help narrow down the problem.

src/SFML/Audio/MiniaudioUtils.cpp
ChrisThrasher: https://github.com/SFML/SFML/compare/feature/miniaudio...saved_settings

Here's what that looks like. What do you think?

ChrisThrasher:
It's working now!

src/SFML/Audio/MiniaudioUtils.cpp
binary1248: Implemented.

ChrisThrasher (approved):
👍🏻👍🏻👍🏻

Bambo-Borris:
Just tested this on my current project, felt sweet to be able to delete `openal32.dll` and have this just work. Noticed no problems, and also had to make 0 changes in my code to get it to work. Great job so far folks!

Marioalexsan:
https://github.com/Marioalexsan/ProjectSilver/blob/7d328858e5d664af14a8891b9de71de620f76651/ProjectSilver/src/ProjectSilver/AudioEngine.cpp#L165

This check now returns a status of `Paused` instead of `Stopped` when using the miniaudio branch, which doesn't really seem correct given that I don't even use `pause()` anywhere.

This is the state in which the audio is when doing the stop check:
image

ChrisThrasher:

src/SFML/Audio/AudioDevice.hpp
```diff
@@ -56,28 +58,16 @@ class AudioDevice
...
+    /// There should only be a single instance of AudioDevice.
+    /// As long as an AudioResource exists, this function should
+    /// always return a valid pointer to the audio engine.
     ///
-    /// \return Corresponding format
+    /// \return The audio engine
     ///
     ////////////////////////////////////////////////////////////
-    static int getFormatFromChannelCount(unsigned int channelCount);
+    static void* getEngine();
```
ChrisThrasher, line 70: Why does this return a `void*` instead of a `ma_engine*`? I think it'd be cleaner and safer to instead forward declare `ma_engine` and use the type directly in the API.

ChrisThrasher (changes requested):

ChrisThrasher:
https://github.com/SFML/SFML/compare/feature/miniaudio...feature/miniaudio_remove_pimpl

I removed the use of PIMPL from AudioDevice. Here are the compile time differences

`feature/miniaudio`:
```
Benchmark 1: ninja clean && time ninja sfml-audio
  Time (mean ± σ):      2.109 s ±  0.024 s    [User: 11.991 s, System: 1.479 s]
  Range (min … max):    2.059 s …  2.141 s    10 runs
```

`feature/miniaudio_remove_pimpl`:
```
Benchmark 1: ninja clean && time ninja sfml-audio
  Time (mean ± σ):      2.105 s ±  0.029 s    [User: 12.066 s, System: 1.482 s]
  Range (min … max):    2.046 s …  2.150 s    10 runs
```

PIMPL in this case seems to make zero measurable difference to compile times. We're better off removing that pointer indirection since it serves no purpose.

ChrisThrasher:
Screenshot 2024-04-06 at 3 13 51 PM

I rebased my tests in #2927 onto this PR and am getting some failures. The above failure in `CHECK(music.getStatus() == sf::SoundSource::Status::Paused);` sounds similar to what Marioalexsan brought up.

ChrisThrasher:
Rebased on `master` and fixed a minor merge conflict.

binary1248:
- Rebased on master
- Removed PIMPL from `AudioDevice`
- Fixed `getStatus` returning unexpected values, emulates previous OpenAL behaviour now

@Marioalexsan @LumineBot The issues you were having should be solved now.

As far as I can tell, Miniaudio doesn't impose any kind of limit on the number of sounds that can exist or simultaneously play at the same time. All mixing and audio effects are performed in software so it should only be limited by available CPU/RAM.

danieljpetersen:
I'm going to add my voice to the choir in telling you all that you're awesome. I updated my game to use this branch, I haven't noticed any issues. The new functionality is great, we can do some cool dynamic audio transitions very easily now. Not needing a dll for openal is also very nice.

binary1248:
> I'm going to add my voice to the choir in telling you all that you're awesome. I updated my game to use this branch, I haven't noticed any issues. The new functionality is great, we can do some cool dynamic audio transitions very easily now. Not needing a dll for openal is also very nice.

And there is nothing more rewarding than seeing all the new features being put to good use. 😃

danieljpetersen:
Ah, unfortunately I may have found an issue. The following program crashes on the miniaudio branch but not on the master branch
```cpp
#include <SFML/Audio.hpp>
#include <vector>
#include <iostream>

int main() {
	std::vector<std::string> files = {
		"example1.wav",
		"example2.wav"
	};

	std::vector<sf::SoundBuffer> buffers(files.size());
	std::vector<sf::Sound> sounds;

	for (size_t i = 0; i < files.size(); ++i) {
		std::cout << files[i] << std::endl;
		if (!buffers[i].loadFromFile(files[i])) {
			std::cerr << "Failed to load " << files[i] << std::endl;
			return -1;
		}

		sounds.emplace_back(buffers[i]); // crashes here
	}
}
```

pure virtual method called
terminate called without an active exception

Process finished with exit code 134 (interrupted by signal 6:SIGABRT)

__pthread_kill_implementation 0x00007ffff69b29fc
sf::SoundSource::setPitch SoundSource.cpp:63
sf::SoundSource::SoundSource SoundSource.cpp:42
sf::Sound::Sound Sound.cpp:230
main main.cpp:26

This is on commit 8c15fafdb19bb9ff49e544c0c33ee2b6798c5dfc

Bambo-Borris:
> Ah, unfortunately I may have found an issue. The following program crashes on the miniaudio branch but not on the master branch
> 
> ```c++
> #include <SFML/Audio.hpp>
> #include <vector>
> #include <iostream>
> 
> int main() {
> 	std::vector<std::string> files = {
> 		"example1.wav",
> 		"example2.wav"
> 	};
> 
> 	std::vector<sf::SoundBuffer> buffers(files.size());
> 	std::vector<sf::Sound> sounds;
> 
> 	for (size_t i = 0; i < files.size(); ++i) {
> 		std::cout << files[i] << std::endl;
> 		if (!buffers[i].loadFromFile(files[i])) {
> 			std::cerr << "Failed to load " << files[i] << std::endl;
> 			return -1;
> 		}
> 
> 		sounds.emplace_back(buffers[i]); // crashes here
> 	}
> 	return 0;
> }
> ```
> 
> pure virtual method called terminate called without an active exception
> 
> Process finished with exit code 134 (interrupted by signal 6:SIGABRT)
> 
> __pthread_kill_implementation 0x00007ffff69b29fc sf::SoundSource::setPitch SoundSource.cpp:63 sf::SoundSource::SoundSource SoundSource.cpp:42 sf::Sound::Sound Sound.cpp:230 main main.cpp:26
> 
> This is on commit [8c15faf](https://github.com/SFML/SFML/commit/8c15fafdb19bb9ff49e544c0c33ee2b6798c5dfc)

Looks like `sf::SoundSource` copy constructor relies on being able to call `setPitch`, which will try invoke `getSound()`, but `getSound()` is virtual on SoundSource and is overriden by `sf::Sound`, which itself hasn't been fully constructed since `sf::SoundSource` must be constructed first. Also `m_impl` won't be valid until after `sf::SoundSource` copy constructor has run.

Threw some screenshots below of what I spotted in the debugger, don't have much more to offer than that but hope it helps! Loving the miniaudio work and wanna help it anyway I can! 

image
image

src/SFML/Audio/SoundSource.cpp
```diff
@@ -25,155 +25,349 @@
...
-#endif
+#include <miniaudio.h>
+
+#include <algorithm>
+
 
 namespace sf
 {
 // NOLINTBEGIN(readability-make-member-function-const)
+// NOLINTBEGIN(clang-analyzer-cplusplus.PureVirtualCall)
```
ChrisThrasher, line 38: > Looks like sf::SoundSource copy constructor relies on being able to call setPitch, which will try invoke getSound(), but getSound() is virtual on SoundSource and is overriden by sf::Sound, which itself hasn't been fully constructed since sf::SoundSource must be constructed first. 

clang-tidy caught this pure virtual function call bug a while ago but I silenced this check for the sake of getting CI to pass. Here's what clang-tidy reports when I reenable this check.

```
/Users/thrasher/Projects/sfml/src/SFML/Audio/SoundSource.cpp:62:46: error: Call to pure virtual method 'SoundSource::getSound' during construction has undefined behavior [clang-analyzer-cplusplus.PureVirtualCall,-warnings-as-errors]
   62 |     if (auto* sound = static_cast<ma_sound*>(getSound()))
      |                                              ^
/Users/thrasher/Projects/sfml/src/SFML/Audio/SoundSource.cpp:41:5: note: Calling 'SoundSource::setPitch'
   41 |     setPitch(copy.getPitch());
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~
/Users/thrasher/Projects/sfml/src/SFML/Audio/SoundSource.cpp:62:46: note: Call to pure virtual method 'SoundSource::getSound' during construction has undefined behavior
   62 |     if (auto* sound = static_cast<ma_sound*>(getSound()))
      |                        
```

The tests I wrote in #2927 also catch this bug. Those tests crash when I rebased them on this PR.

ChrisThrasher:
Rebased on `master` and fixed minor merge conflict.

ChrisThrasher:
I reenabled the clang-tidy check which is catching this pure virtual function call bug so that we can get better feedback once we start trying to fix it.

binary1248:
@danieljpetersen The pure virtual method call issue should be fixed now.

ChrisThrasher:
My Audio unit tests are now passing without any crashes. There is however this oddity in the `sf::SoundSource` tests

```cpp
    SECTION("Set/get pitch")
    {
        SoundSource soundSource;
        soundSource.setPitch(42);
        CHECK(soundSource.getPitch() == 0);
    }

    SECTION("Set/get volume")
    {
        SoundSource soundSource;
        soundSource.setVolume(0.5f);
        CHECK(soundSource.getVolume() == 0);
    }

    SECTION("Set/get position")
    {
        SoundSource soundSource;
        soundSource.setPosition({1, 2, 3});
        CHECK(soundSource.getPosition() == sf::Vector3f());
    }

    SECTION("Set/get relative to listener")
    {
        SoundSource soundSource;
        soundSource.setRelativeToListener(true);
        CHECK(!soundSource.isRelativeToListener());
    }

    SECTION("Set/get min distance")
    {
        SoundSource soundSource;
        soundSource.setMinDistance(12.34f);
        CHECK(soundSource.getMinDistance() == 0);
    }

    SECTION("Set/get attenuation")
    {
        SoundSource soundSource;
        soundSource.setAttenuation(10);
        CHECK(soundSource.getAttenuation() == 0);
    }
```

The setters and getters seem to just not work. Setting a value does not result in being about to get the old value. The setters and getters behave as one would expect with the OpenAL backend.

binary1248:
`SoundSource` was downgraded from being a semi-concrete class managing the actual OpenAL sound source to being an abstract interface with common functionality that is used by `Sound` and `SoundStream`. If you work around the fact that you shouldn't even be able to instantiate a `SoundSource` object by defining an empty derived class you can't expect it to work properly because there is no backing implementation behind it. If you want to test the `SoundSource` interface instantiate a proper object (`Sound` or `Music`) and test through a `SoundSource&` to the object.

ChrisThrasher:
CI is failing due to Homebrew updating clang-tidy which includes some new checks. I fixed that in https://github.com/SFML/SFML/pull/2953.

danieljpetersen:
Thank you Binary! That did indeed fix it. And thank you Bambo, thank you Thrasher!

Yesterday I wrote a little sound interpolator for the audio section of my codebase, where you give a duration and an easing formula for any given music/sound property(s). I've got it working in my game now to transition music on scene changes, it's especially nice when going from high tempo to low tempo music.

I'm looking forward to whenever this gets merged into master.

ChrisThrasher (approved):

eXpl0it3r (changes requested):
The assets of the new example need to be referenced in the asset_licenses.md file.

It's a really cool PR! 😎 

Beyond getting rid or OpenAL and its LGPL implications, it's awesome to see the natural use of SFML 3 features like Angle and C++17(-ish). 🎉

src/SFML/Audio/SoundBuffer.cpp
```diff
@@ -25,60 +25,45 @@
...
 
-#if defined(__APPLE__)
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-#endif
 
 namespace sf
 {
 ////////////////////////////////////////////////////////////
-SoundBuffer::SoundBuffer()
+SoundBuffer::SoundBuffer(const SoundBuffer& copy) : SoundBuffer()
```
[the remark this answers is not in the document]
ChrisThrasher: I'm not sure either. The default constructor doesn't do anything meaningful. I tried removing this and let CI run but nothing fails so it's not there to silence a warning from some tool or compiler.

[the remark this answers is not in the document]
binary1248: https://github.com/SFML/SFML/actions/runs/8789198971/job/24118432275#step:16:103

GCC complained if is was left away. 🤷‍♂️

src/SFML/Audio/Sound.cpp
```diff
@@ -56,99 +241,136 @@ Sound::Sound(const Sound& copy) : SoundSource(copy), m_buffer(copy.m_buffer)
...
 }
 
 
 ////////////////////////////////////////////////////////////
 void Sound::pause()
 {
-    alCheck(alSourcePause(m_source));
+    if (const ma_result result = ma_sound_stop(&m_impl->sound); result != MA_SUCCESS)
+    {
+        err() << "Failed to stop playing sound: " << ma_result_description(result) << std::endl;
```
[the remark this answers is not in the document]
binary1248: The message refers to what actually went wrong in the backend (stopping the sound), thus the miniaudio specific error message as well. It wasn't meant to provide feedback on the user facing function call.

src/SFML/Audio/SoundBuffer.cpp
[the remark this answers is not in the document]
eXpl0it3r: For Sound and SoundSource makes kind of sense, but this here is SoundBuffer calling SoundBuffer, so no base class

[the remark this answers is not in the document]
binary1248: Probably a leftover from the transition period. Will remove.

binary1248:
Rebased and updated.

eXpl0it3r (approved):
The asset_licenses.md file still needs updating, but we can do that later as well. 😊

eXpl0it3r:
Big thanks to @binary1248 for taking on this work and everyone for reviewing! 🎉

PR #2756, opened 2023-10-28
Remove output parameter from `sf::Image::saveToMemory`

ChrisThrasher:
## Description

This is a textbook use case for a `std::optional` return value. This operation might return a buffer or it might not. It depends on whether the inputs are valid.

### DIFF
--- a/include/SFML/Graphics/Image.hpp
+++ b/include/SFML/Graphics/Image.hpp
@@ -33,6 +33,7 @@
 #include <SFML/Graphics/Rect.hpp>
 
 #include <filesystem>
+#include <optional>
 #include <string_view>
 #include <vector>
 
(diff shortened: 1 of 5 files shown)


ChrisThrasher:
`sf::priv::ImageLoader` uses output parameters a fair bit as well. If we like this idea then maybe I can add more similar changes to this PR.

binary1248 (approved):

include/SFML/Graphics/Image.hpp
```diff
@@ -148,15 +149,14 @@ class SFML_GRAPHICS_API Image
     /// This function fails if the image is empty, or if
     /// the format was invalid.
     ///
-    /// \param output Buffer to fill with encoded data
     /// \param format Encoding format to use
     ///
-    /// \return True if saving was successful
+    /// \return Buffer with encoded data if saving was successful
```
eXpl0it3r, line 154: Maybe we can add "`std::nullopt` otherwise" or similar 🤔

eXpl0it3r (approved):

kimci86:
It is now impossible to reuse or reserve vector memory upfront to save some memory allocations. No big deal I guess but I just wanted to point it out.

PR #2757, opened 2023-10-29
Manually inline all of `sf::priv::ImageLoader`

ChrisThrasher:
Another ~250 lines of code gone!

This layer of indirection was unnecessary. All of its functions were only used once somewhere inside of Image.cpp. The code is shorter and simpler and easier to reason about if we put the implementations of these functions directly in Image.cpp.

Along the same lines as #2756, this is removing even more output parameters.

---

https://github.com/SFML/SFML/compare/image_loader_namespace...image_memory_management

I have more improvements lined up once we can merge this. I realized that we're working with owning, raw pointers and that we can easily replace them with `unique_ptr` instead.


kimci86:
General remark: this is difficult to review because you moved code and made changes to the moved code in a single commit.

ChrisThrasher:
The changes were mostly just replacing `std::memcpy` with `std::vector::assign`.

eXpl0it3r:
> The changes were mostly just replacing `std::memcpy` with `std::vector::assign`. Let me know if that is an unreasonable reimplementation to make.

Not sure if you misunderstood kimci86, I believe the change itself is okay, their point was more that it's easier to review something, if each step is split into multiple commits (e.g. commit 1: Moving code, commit 2: Cleaning up code). It's now much harder to verify that the code was moved correctly, as one has to consider the modified code on the fly.

I don't think, you need to change it here, just something for the future.

eXpl0it3r (approved):

ChrisThrasher:
Yeah I'm totally aware, thanks.

PR #2760, opened 2023-10-31
Backmerge 2.6.x

eXpl0it3r:
## Description

Back merging [changes from 2.6.x](https://github.com/SFML/SFML/compare/master...2.6.x) to master.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,38 @@
 # Changelog
 
+## SFML 2.6.1
+
+Also available on the website: https://www.sfml-dev.org/changelog.php#sfml-2.6.1
+
+### General
+
+-   Stop overwriting CMAKE_MODULE_PATH (#2577)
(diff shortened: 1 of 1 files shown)


src/SFML/Graphics/Texture.cpp
```diff
@@ -44,6 +44,10 @@
 #include <climits>
 #include <cstring>
 
+#if defined(__GNUC__)
+    #pragma GCC diagnostic ignored "-Wduplicated-branches"
+#endif
+
```
ChrisThrasher, line 50: I think we can remove this in `master`. We have OpenGL ES jobs in v3 and this warning doesn't appear.

ChrisThrasher (changes requested):
One small tweak but otherwise good to go 👍🏻

ChrisThrasher (approved):

PR #2761, opened 2023-11-04
Perform the majority of GitHub Actions workflow MSVC building using Ninja

binary1248:
Perform the majority of GitHub Actions workflow MSVC building using Ninja with the exception of ClangCL+MSBuild.

Specifying Ninja as the generator prevents specying ClangCL as the toolset. The `clang-cl.exe` binary would have to be manually specifyied as C and CXX compilers when configuring CMake in order to build using Ninja and ClangCL.

Since we want to test building using MSBuild as well, we use the opportunity to test MSBuild with ClangCL which simplifies things for us.

Building MSVC with Ninja enables ccache to be used for MSVC builds as well, as shown in the logs. The reason why nothing can currently be cached is because the debugging information is stored in a separate file when building with the default `/Zi`. This will be handled in a follow-up PR to this one.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,23 +27,23 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-        - { name: Windows VS2019 x86,     os: windows-2019, flags: -DSFML_USE_MESA3D=TRUE -A Win32 }
-        - { name: Windows VS2019 x64,     os: windows-2019, flags: -DSFML_USE_MESA3D=TRUE -A x64 }
-        - { name: Windows VS2022 x86,     os: windows-2022, flags: -DSFML_USE_MESA3D=TRUE -A Win32 }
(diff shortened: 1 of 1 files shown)


ChrisThrasher:
Approximately how much CI time does this save us?

binary1248:
By itself, probably not that much (although on paper ninja should be faster than MSBuild), but it is a requirement to enable ccache to do its thing for MSVC.

ChrisThrasher:
Including the benefits of ccache, what runtime differences are we looking at?

binary1248:
Comparing [this](https://github.com/SFML/SFML/actions/runs/6755404699) to [this](https://github.com/SFML/SFML/actions/runs/6756640801), total run time drops from 4h 37min to 3h 33min and total duration drops from 18min 37s to 16min 35s.

eXpl0it3r (approved):

ChrisThrasher (approved):

PR #2763, opened 2023-11-04
Add more tests for `sf::Context`

ChrisThrasher:
## Description

While improving the test coverage of `sf::Context` I found out that certain `nullptr` arguments would cause a crash so I added `assert`s to protect against it.


ChrisThrasher:
I prematurely submitted this PR. The new tests are not ready.

PR #2766, opened 2023-11-06
Rewrite `sf::Event` API to improve type safety

### LINKED ISSUE #2668: `sf::Event` API + type-safe unions
Since the topic came up again in #2665, I'd like to discuss different possible approaches for `sf::Event`.

### 0. `union`

The current approach based on `union` has multiple problems.

First and foremost, it's not type-safe and allows to easily create undefined behavior by reading the wrong variant. The fact that things are not very distinctly named doesn't help here:
```cpp
if (event.type == sf::Event::MouseWheelScrolled) {
   use(event.mouseWheel); // BAM UB, should have been mouseWheelScrolled
}
```

Second, unions are very limited regarding constructor/destructor, so we cannot e.g. store `std::string` inside a variant.
So, I'd like to use this issue to brainstorm different approaches and then converge on the one we like best 🙂 

### 1. `std::variant`

An option is to directly declare an `std::variant` of all possible event types:
```cpp
using Event = std::variant<Event::Closed, Event::Resized, ... Event::SensorChanged>;
```

`std::variant` has the advantage of being standard. However, there are a few issues:
* The visitor pattern is very verbose. There is the [`overload{}` life hack](https://en.cppreference.com/w/cpp/utility/variant/visit) which makes things a bit nicer, but IIRC it had quite some compile-time impact (this may no longer be true). At the very least, we always need the standard `<variant>` header.
* The visitor must be a lambda (or function object), meaning it interrupts control flow.
  This makes it harder to do things like `if (/* event is XY */) return;`.
* Not sure if this is still an issue, but there were quite some runtime performance impacts ([1](https://www.reddit.com/r/cpp/comments/kst2pu/with_stdvariant_you_choose_either_performance_or/), [2](https://orangejuiceliberationfront.com/surprising-slowness-of-cpp-std-variant/), [3](https://stackoverflow.com/questions/57726401/stdvariant-vs-inheritance-vs-other-ways-performance)). It might not be relevant because event handling involves expensive OS calls, but I'm listing it for completeness.
* By exposing a `std::variant` directly, implementation details may leak:
   * Changing the index of a variant is a breaking change for users.
   * We cannot add variants later, as the public-facing type is then different.

### 2. the OOP way, per event

Have a base `Event` with virtual destructor and inherit concrete events from it. Return the object inside a `std::unique_ptr<sf::Event>`. This needs to be downcast by the user.

Possible but not ideal as all the objects needs to be boxed, although it might not matter performance-wise. Would allow to have hierarchies in case we need that (e.g. intermediate class for different mouse events, containing mouse position) -- but doubtful that it's worth it.

### 3. the OOP way, as an event handler

Instead of having polymorphism on the `Event` itself, we could provide an abstract base class:
```cpp
class MyHandler : public sf::EventHandler {
    // Handle specific events -- return true if handled
    virtual bool onEvent(Event::MouseMove e) override;
    virtual bool onEvent(Event::KeyPressed e) override;

    // Have a fallback for unhandled events -- probably not needed?
    virtual void onUnhandledEvent(??);
};
```
Again a bit the issue of non-locality and verbosity. Possibly a bit too java-style, just thought I'd mention it as a non-generic alternative to `std::visit()`, but it shares similar problems.

### 4. generic get interface, hidden implementation

We would just expose a method to get a concrete type, based on `std::optional`. If the user fetches the wrong type, he will get an empty optional.

```cpp
class Event {
   ...
public:
    // Returns T if it's that variant, or empty
    template <typename T>
    std::optional<T> get() const;
};
```
```cpp
if (auto e = event.get<Event::MouseMove>()) {
    // procedural code
} else if (auto e = event.get<Event::KeyPressed>()) {
    // ...
}
```
This even allows us to use `std::variant` internally, but it's hidden as an implementation detail. If we forward the `get<T>()` invocation to `variant::get<T>()`, we'd implicitly check whether `T` is one of the allowed variants (and in a sense, the internal `std::variant` is then a leaky abstraction).

The disadvantage is that there's no visiting API, but a) this could be added/exposed if really needed, and b) events are typically not queried exhaustively. Users want to know if something they're interested in (like a key press) has happened. 

On the other hand, what this _does_ allow is a "catch-all" `else` branch. If ever we add debugging information such as `event.toString()`, that might be a possible choice to do in an `else` branch. Or when someone has their own event system, to enqueue `std::vector<sf::Event>` or so.

### 5. specific get interface

Same as 3, but with named methods.
```cpp
class Event {
   ...
public:
    std::optional<Event::MouseMove> mouseMove() const;
};
```
```cpp
if (auto e = event.mouseMove()) {
    // procedural code
} else if (auto e = event.keyPressed()) {
    // ...
}
```
Very concise but still safe. It's unlikely that people use `sf::Event` in a generic way, so having a generic interface only benefits us a small bit on the implementation side (reducing some code duplication). On the other hand, we can define everything in the .cpp file here and don't need a `<variant>` header.

This API is almost "boring", but it might come with the least surprises and has no issues regarding compile time or template error messages.

Also, it makes the API very discoverable: for example, when typing `event.` in an IDE, auto-complete will suggest all possible getters. This is not the case for `get<T>()`.

ChrisThrasher:
## Description

Closes #2668 

This new API is built on top of std::variant. This allows us to store many different event types in a space-efficient way and access the active event type in a type-safe manner that eliminates the types of UB that are possible with unions.

I tried to find docs that referenced the old API but may have missed a few places. Initially my focus is on consensus for API design. After that we can focus harder on updating all the docs. There are also some missing comment blocks I need to fill in and more tests to write so there is some finishing work to do.

---

> What are all these new structs you added?

The current implementation has fewer structs than values in the enumeration. This means there is not a 1-to-1 mapping between event types and data structures. By ensuring there is a 1-to-1 mapping it lets us use the variant index to get the enumeration value. This is very easy to write, read, and maintain. Believe me this is a lot better than alternative implementations that don't add new structs.

> What about `std::visit`?

I'm reluctant to add this because adding an element to a variant constitutes an API break for anyone visiting that variant. You must visit a variant with the perfect set of callbacks for the elements in the variant. If we allow users to visit this variant then we're effectively preventing ourselves from adding new events in a given major release. For now I'd prefer to retain that ability. We can always add this later if desired.

EDIT 28 Jan: Apparently it is possible to support `std::visit` without prohibiting us from adding new events so I'm open to this idea.

> Why bother keeping the enumeration in the API instead of the alternatives in #2668?

I highly value letting users continue to use `switch` to process potential event states. I think that's a very useful, powerful way to list all the actions one may take in response to a given event type. It's also nice to know that users don't have to remove the `switch` statements they've been writing for years. While the syntax for event processing are different, a lot of their existing knowledge can be carried forward. Users can still write `if`/`else if` chains if they prefer but they're not forced to.

> Invalid variant access results in an exception being thrown. We should not let exceptions escape/this leaks implementation details.

I don't see this as a problem. Any exceptions only occur in incorrect programs so no one writing correct code will ever deal with them. The exceptions will server to very quickly catch bugs in user programs. To avoid leaking this detail would require we jump through extra hoops and possibly introduce new avenues for UB that I would rather not do.

---

EDIT 16 Apr 2024: The rest of the team disagrees about the value of returning an enumeration so I removed the APIs that used the enum. I still stand by this as a good API but don't want to stall this PR indefinitely so I removed it.

### DIFF
--- a/examples/android/app/src/main/jni/main.cpp
+++ b/examples/android/app/src/main/jni/main.cpp
@@ -115,46 +115,49 @@ int main(int argc, char* argv[])
     {
         for (sf::Event event; active ? window.pollEvent(event) : window.waitEvent(event);)
         {
-            switch (event.type)
+            if (event.is<sf::Event::Closed>())
             {
-                case sf::Event::Closed:
+                window.close();
+            }
+            else if (const auto* keyPressed = event.getIf<sf::Event::KeyPressed>())
+            {
+                if (keyPressed->code == sf::Keyboard::Key::Escape)
                     window.close();
-                    break;
-                case sf::Event::KeyPressed:
-                    if (event.key.code == sf::Keyboard::Key::Escape)
-                        window.close();
-                    break;
-                case sf::Event::Resized:
-                    view.setSize(sf::Vector2f(event.size.width, event.size.height));
-                    view.setCenter(sf::Vector2f(event.size.width, event.size.height) / 2.f);
-                    window.setView(view);
-                    break;
-                case sf::Event::LostFocus:
-                    background = sf::Color::Black;
-                    break;
-                case sf::Event::GainedFocus:
-                    background = sf::Color::White;
-                    break;
-
-                // On Android MouseLeft/MouseEntered are (for now) triggered,
-                // whenever the app loses or gains focus.
-                case sf::Event::MouseLeft:
-                    active = false;
-                    break;
-                case sf::Event::MouseEntered:
-                    active = true;
-                    break;
-                case sf::Event::TouchBegan:
-                    if (event.touch.finger == 0)
-                    {
-                        image.setPosition({static_cast<float>(event.touch.x), static_cast<float>(event.touch.y)});
+            }
+            else if (const auto* resized = event.getIf<sf::Event::Resized>())
+            {
+                const auto size = sf::Vector2f(resized->size);
+                view.setSize(size);
+                view.setCenter(size / 2.f);
+                window.setView(view);
+            }
+            else if (event.is<sf::Event::FocusLost>())
+            {
+                background = sf::Color::Black;
+            }
+            else if (event.is<sf::Event::FocusGained>())
+            {
+                background = sf::Color::White;
+            }
+            // On Android MouseLeft/MouseEntered are (for now) triggered,
+            // whenever the app loses or gains focus.
+            else if (event.is<sf::Event::MouseLeft>())
+            {
+                active = false;
+            }
+            else if (event.is<sf::Event::MouseEntered>())
+            {
+                active = true;
+            }
+            else if (const auto* touchBegan = event.getIf<sf::Event::TouchBegan>())
+            {
+                if (touchBegan->finger == 0)
+                {
+                    image.setPosition(sf::Vector2f(touchBegan->position));
 #if defined(USE_JNI)
-                        vibrate(sf::milliseconds(10));
+                    vibrate(sf::milliseconds(10));
 #endif
-                    }
-                    break;
-                default:
-                    break;
+                }
             }
         }
 
--- a/examples/island/Island.cpp
+++ b/examples/island/Island.cpp
@@ -182,17 +182,17 @@ int main()
         for (sf::Event event; window.pollEvent(event);)
         {
             // Window closed or escape key pressed: exit
-            if ((event.type == sf::Event::Closed) ||
-                ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Key::Escape)))
+            if (event.is<sf::Event::Closed>() || (event.is<sf::Event::KeyPressed>() &&
+                                                  event.getIf<sf::Event::KeyPressed>()->code == sf::Keyboard::Key::Escape))
             {
                 window.close();
                 break;
             }
 
             // Arrow key pressed:
-            if (prerequisitesSupported && (event.type == sf::Event::KeyPressed))
+            if (prerequisitesSupported && event.is<sf::Event::KeyPressed>())
             {
-                switch (event.key.code)
+                switch (event.getIf<sf::Event::KeyPressed>()->code)
                 {
                     case sf::Keyboard::Key::Enter:
                         generateTerrain(terrainStagingBuffer.data());
--- a/examples/opengl/OpenGL.cpp
+++ b/examples/opengl/OpenGL.cpp
@@ -211,21 +211,23 @@ int main()
             for (sf::Event event; window.pollEvent(event);)
             {
                 // Close window: exit
-                if (event.type == sf::Event::Closed)
+                if (event.is<sf::Event::Closed>())
                 {
                     exit = true;
                     window.close();
                 }
 
                 // Escape key: exit
-                if ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Key::Escape))
+                if (const auto* keyPressed = event.getIf<sf::Event::KeyPressed>();
+                    keyPressed && keyPressed->code == sf::Keyboard::Key::Escape)
                 {
                     exit = true;
                     window.close();
                 }
 
                 // Return key: toggle mipmapping
-                if ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Key::Enter))
+                if (const auto* keyPressed = event.getIf<sf::Event::KeyPressed>();
+                    keyPressed && keyPressed->code == sf::Keyboard::Key::Enter)
                 {
                     if (mipmapEnabled)
                     {
@@ -242,14 +244,15 @@ int main()
                 }
 
                 // Space key: toggle sRGB conversion
-                if ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Key::Space))
+                if (const auto* keyPressed = event.getIf<sf::Event::KeyPressed>();
+                    keyPressed && keyPressed->code == sf::Keyboard::Key::Space)
                 {
                     sRgb = !sRgb;
                     window.close();
                 }
 
                 // Adjust the viewport when the window is resized
-                if (event.type == sf::Event::Resized)
+                if (const auto* resized = event.getIf<sf::Event::Resized>())
                 {
                     const sf::Vector2u textureSize = backgroundTexture.getSize();
 
@@ -260,10 +263,11 @@ int main()
                         return EXIT_FAILURE;
                     }
 
-                    glViewport(0, 0, static_cast<GLsizei>(event.size.width), static_cast<GLsizei>(event.size.height));
+                    const auto [width, height] = resized->size;
+                    glViewport(0, 0, static_cast<GLsizei>(width), static_cast<GLsizei>(height));
                     glMatrixMode(GL_PROJECTION);
                     glLoadIdentity();
-                    const GLfloat newRatio = static_cast<float>(event.size.width) / static_cast<float>(event.size.height);
+                    const GLfloat newRatio = static_cast<float>(width) / static_cast<float>(height);
 #ifdef SFML_OPENGL_ES
                     glFrustumf(-newRatio, newRatio, -1.f, 1.f, 1.f, 500.f);
 #else
--- a/examples/window/Window.cpp
+++ b/examples/window/Window.cpp
@@ -144,20 +144,22 @@ int main()
         for (sf::Event event; window.pollEvent(event);)
         {
             // Close window: exit
-            if (event.type == sf::Event::Closed)
+            if (event.is<sf::Event::Closed>())
                 window.close();
 
             // Escape key: exit
-            if ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Key::Escape))
-                window.close();
+            if (const auto* keyPressed = event.getIf<sf::Event::KeyPressed>())
+                if (keyPressed->code == sf::Keyboard::Key::Escape)
+                    window.close();
 
             // Resize event: adjust the viewport
-            if (event.type == sf::Event::Resized)
+            if (const auto* resized = event.getIf<sf::Event::Resized>())
             {
-                glViewport(0, 0, static_cast<GLsizei>(event.size.width), static_cast<GLsizei>(event.size.height));
+                const auto [width, height] = resized->size;
+                glViewport(0, 0, static_cast<GLsizei>(width), static_cast<GLsizei>(height));
                 glMatrixMode(GL_PROJECTION);
                 glLoadIdentity();
-                const GLfloat newRatio = static_cast<float>(event.size.width) / static_cast<float>(event.size.height);
+                const GLfloat newRatio = static_cast<float>(width) / static_cast<float>(height);
 #ifdef SFML_OPENGL_ES
                 glFrustumf(-newRatio, newRatio, -1.f, 1.f, 1.f, 500.f);
 #else
--- a/include/SFML/Window/Clipboard.hpp
+++ b/include/SFML/Window/Clipboard.hpp
@@ -93,16 +93,16 @@ SFML_WINDOW_API void setString(const String& text);
 /// // or use it in the event loop
 /// for (sf::Event event; window.pollEvent(event);)
 /// {
-///     if(event.type == sf::Event::Closed)
+///     if(event.is<sf::Event::Closed>())
 ///         window.close();
-///     if(event.type == sf::Event::KeyPressed)
+///     if(const auto* keyPressed = event.getIf<sf::Event::KeyPressed>())
 ///     {
 ///         // Using Ctrl + V to paste a string into SFML
-///         if(event.key.control && event.key.code == sf::Keyboard::Key::V)
+///         if(keyPressed->control && keyPressed->code == sf::Keyboard::Key::V)
 ///             string = sf::Clipboard::getString();
 ///
 ///         // Using Ctrl + C to copy a string out of SFML
-///         if(event.key.control && event.key.code == sf::Keyboard::Key::C)
+///         if(keyPressed->control && keyPressed->code == sf::Keyboard::Key::C)
 ///             sf::Clipboard::setString("Hello World!");
 ///     }
 /// }
--- a/include/SFML/Window/Event.hpp
+++ b/include/SFML/Window/Event.hpp
@@ -34,8 +34,10 @@
 #include <SFML/Window/Mouse.hpp>
 #include <SFML/Window/Sensor.hpp>
 
+#include <SFML/System/Vector2.hpp>
 
-// NOLINTBEGIN(cppcoreguidelines-pro-type-member-init)
+#include <variant>
+
 
 namespace sf
 {
@@ -43,186 +45,282 @@ namespace sf
 /// \brief Defines a system event and its parameters
 ///
 ////////////////////////////////////////////////////////////
-struct Event
+class SFML_WINDOW_API Event
 {
+public:
     ////////////////////////////////////////////////////////////
-    /// \brief Size events parameters (Resized)
+    /// \brief Empty event
     ///
     ////////////////////////////////////////////////////////////
-    struct SizeEvent
+    struct Empty
     {
-        unsigned int width;  //!< New width, in pixels
-        unsigned int height; //!< New height, in pixels
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Keyboard event parameters (KeyPressed, KeyReleased)
+    /// \brief Closed event
     ///
     ////////////////////////////////////////////////////////////
-    struct KeyEvent
+    struct Closed
     {
-        Keyboard::Key      code;     //!< Code of the key that has been pressed
-        Keyboard::Scancode scancode; //!< Physical code of the key that has been pressed
-        bool               alt;      //!< Is the Alt key pressed?
-        bool               control;  //!< Is the Control key pressed?
-        bool               shift;    //!< Is the Shift key pressed?
-        bool               system;   //!< Is the System key pressed?
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Text event parameters (TextEntered)
+    /// \brief Resized event
     ///
     ////////////////////////////////////////////////////////////
-    struct TextEvent
+    struct Resized
     {
-        std::uint32_t unicode; //!< UTF-32 Unicode value of the character
+        Vector2u size; //!< New size, in pixels
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Mouse move event parameters (MouseMoved)
+    /// \brief Lost focus event
     ///
     ////////////////////////////////////////////////////////////
-    struct MouseMoveEvent
+    struct FocusLost
     {
-        int x; //!< X position of the mouse pointer, relative to the left of the owner window
-        int y; //!< Y position of the mouse pointer, relative to the top of the owner window
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Mouse buttons events parameters
-    ///        (MouseButtonPressed, MouseButtonReleased)
+    /// \brief Gained focus event
     ///
     ////////////////////////////////////////////////////////////
-    struct MouseButtonEvent
+    struct FocusGained
     {
-        Mouse::Button button; //!< Code of the button that has been pressed
-        int           x;      //!< X position of the mouse pointer, relative to the left of the owner window
-        int           y;      //!< Y position of the mouse pointer, relative to the top of the owner window
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Mouse wheel events parameters (MouseWheelScrolled)
+    /// \brief Text event
     ///
     ////////////////////////////////////////////////////////////
-    struct MouseWheelScrollEvent
+    struct TextEntered
     {
-        Mouse::Wheel wheel; //!< Which wheel (for mice with multiple ones)
-        float        delta; //!< Wheel offset (positive is up/left, negative is down/right). High-precision mice may use non-integral offsets.
-        int          x; //!< X position of the mouse pointer, relative to the left of the owner window
-        int          y; //!< Y position of the mouse pointer, relative to the top of the owner window
+        std::uint32_t unicode{}; //!< UTF-32 Unicode value of the character
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Joystick connection events parameters
-    ///        (JoystickConnected, JoystickDisconnected)
+    /// \brief KeyChanged events
     ///
     ////////////////////////////////////////////////////////////
-    struct JoystickConnectEvent
+    struct KeyChanged
     {
-        unsigned int joystickId; //!< Index of the joystick (in range [0 .. Joystick::Count - 1])
+        Keyboard::Key      code{};     //!< Code of the key
+        Keyboard::Scancode scancode{}; //!< Physical code of the key
+        bool               alt{};      //!< Is the Alt key pressed?
+        bool               control{};  //!< Is the Control key pressed?
+        bool               shift{};    //!< Is the Shift key pressed?
+        bool               system{};   //!< Is the System key pressed?
+    };
+    struct KeyPressed : KeyChanged
+    {
+    };
+    struct KeyReleased : KeyChanged
+    {
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Joystick axis move event parameters (JoystickMoved)
+    /// \brief Mouse wheel scrolled event
     ///
     ////////////////////////////////////////////////////////////
-    struct JoystickMoveEvent
+    struct MouseWheelScrolled
     {
-        unsigned int   joystickId; //!< Index of the joystick (in range [0 .. Joystick::Count - 1])
-        Joystick::Axis axis;       //!< Axis on which the joystick moved
-        float          position;   //!< New position on the axis (in range [-100 .. 100])
+        Mouse::Wheel wheel{}; //!< Which wheel (for mice with multiple ones)
+        float        delta{}; //!< Wheel offset (positive is up/left, negative is down/right). High-precision mice may use non-integral offsets.
+        Vector2i     position; //!< Position of the mouse pointer, relative to the top left of the owner window
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Joystick buttons events parameters
-    ///        (JoystickButtonPressed, JoystickButtonReleased)
+    /// \brief Mouse button changed events
     ///
     ////////////////////////////////////////////////////////////
-    struct JoystickButtonEvent
+    struct MouseButtonChanged
     {
-        unsigned int joystickId; //!< Index of the joystick (in range [0 .. Joystick::Count - 1])
-        unsigned int button; //!< Index of the button that has been pressed (in range [0 .. Joystick::ButtonCount - 1])
+        Mouse::Button button{}; //!< Code of the button that has been pressed
+        Vector2i      position; //!< Position of the mouse pointer, relative to the top left of the owner window
+    };
+    struct MouseButtonPressed : MouseButtonChanged
+    {
+    };
+    struct MouseButtonReleased : MouseButtonChanged
+    {
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Touch events parameters (TouchBegan, TouchMoved, TouchEnded)
+    /// \brief Mouse move event
     ///
     ////////////////////////////////////////////////////////////
-    struct TouchEvent
+    struct MouseMoved
     {
-        unsigned int finger; //!< Index of the finger in case of multi-touch events
-        int          x;      //!< X position of the touch, relative to the left of the owner window
-        int          y;      //!< Y position of the touch, relative to the top of the owner window
+        Vector2i position; //!< Position of the mouse pointer, relative to the top left of the owner window
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Sensor event parameters (SensorChanged)
+    /// \brief Mouse entered event
     ///
     ////////////////////////////////////////////////////////////
-    struct SensorEvent
+    struct MouseEntered
     {
-        Sensor::Type type; //!< Type of the sensor
-        float        x;    //!< Current value of the sensor on X axis
-        float        y;    //!< Current value of the sensor on Y axis
-        float        z;    //!< Current value of the sensor on Z axis
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Enumeration of the different types of events
+    /// \brief Mouse left event
     ///
     ////////////////////////////////////////////////////////////
-    enum EventType
+    struct MouseLeft
     {
-        Closed,                 //!< The window requested to be closed (no data)
-        Resized,                //!< The window was resized (data in event.size)
-        LostFocus,              //!< The window lost the focus (no data)
-        GainedFocus,            //!< The window gained the focus (no data)
-        TextEntered,            //!< A character was entered (data in event.text)
-        KeyPressed,             //!< A key was pressed (data in event.key)
-        KeyReleased,            //!< A key was released (data in event.key)
-        MouseWheelScrolled,     //!< The mouse wheel was scrolled (data in event.mouseWheelScroll)
-        MouseButtonPressed,     //!< A mouse button was pressed (data in event.mouseButton)
-        MouseButtonReleased,    //!< A mouse button was released (data in event.mouseButton)
-        MouseMoved,             //!< The mouse cursor moved (data in event.mouseMove)
-        MouseEntered,           //!< The mouse cursor entered the area of the window (no data)
-        MouseLeft,              //!< The mouse cursor left the area of the window (no data)
-        JoystickButtonPressed,  //!< A joystick button was pressed (data in event.joystickButton)
-        JoystickButtonReleased, //!< A joystick button was released (data in event.joystickButton)
-        JoystickMoved,          //!< The joystick moved along an axis (data in event.joystickMove)
-        JoystickConnected,      //!< A joystick was connected (data in event.joystickConnect)
-        JoystickDisconnected,   //!< A joystick was disconnected (data in event.joystickConnect)
-        TouchBegan,             //!< A touch event began (data in event.touch)
-        TouchMoved,             //!< A touch moved (data in event.touch)
-        TouchEnded,             //!< A touch event ended (data in event.touch)
-        SensorChanged,          //!< A sensor value changed (data in event.sensor)
-
-        Count //!< Keep last -- the total number of event types
     };
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Joystick button events
+    ///
+    ////////////////////////////////////////////////////////////
+    struct JoystickButtonChanged
+    {
+        unsigned int joystickId{}; //!< Index of the joystick (in range [0 .. Joystick::Count - 1])
+        unsigned int button{}; //!< Index of the button that has been pressed (in range [0 .. Joystick::ButtonCount - 1])
+    };
+    struct JoystickButtonPressed : JoystickButtonChanged
+    {
+    };
+    struct JoystickButtonReleased : JoystickButtonChanged
+    {
+    };
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Joystick axis move event
+    ///
+    ////////////////////////////////////////////////////////////
+    struct JoystickMoved
+    {
+        unsigned int   joystickId{}; //!< Index of the joystick (in range [0 .. Joystick::Count - 1])
+        Joystick::Axis axis{};       //!< Axis on which the joystick moved
+        float          position{};   //!< New position on the axis (in range [-100 .. 100])
+    };
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Joystick connection events
+    ///
+    ////////////////////////////////////////////////////////////
+    struct JoystickChanged
+    {
+        unsigned int joystickId{}; //!< Index of the joystick (in range [0 .. Joystick::Count - 1])
+    };
+    struct JoystickConnected : JoystickChanged
+    {
+    };
+    struct JoystickDisconnected : JoystickChanged
+    {
+    };
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Touch events
+    ///
+    ////////////////////////////////////////////////////////////
+    struct TouchChanged
+    {
+        unsigned int finger{}; //!< Index of the finger in case of multi-touch events
+        Vector2i     position; //!< Position of the touch, relative to the top left of the owner window
+    };
+    struct TouchBegan : TouchChanged
+    {
+    };
+    struct TouchMoved : TouchChanged
+    {
+    };
+    struct TouchEnded : TouchChanged
+    {
+    };
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Sensor event
+    ///
+    ////////////////////////////////////////////////////////////
+    struct SensorChanged
+    {
+        Sensor::Type type{}; //!< Type of the sensor
+        Vector3f     value;  //!< Current value of the sensor on the X, Y, and Z axes
+    };
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Default constructor
+    ///
+    /// Sets the event to sf::Event::Empty
+    ///
+    ////////////////////////////////////////////////////////////
+    Event() = default;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Construct from a given sf::Event subtype
+    ///
+    /// \param t Event subtype
+    ///
+    ////////////////////////////////////////////////////////////
+    template <typename T>
+    Event(const T& t);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Check current event type
+    ///
+    /// \return True if template parameter is current event type
+    ///
+    ////////////////////////////////////////////////////////////
+    template <typename T>
+    [[nodiscard]] bool is() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Get particular event type
+    ///
+    /// \return Address of current event type, otherwise nullptr
+    ///
+    ////////////////////////////////////////////////////////////
+    template <typename T>
+    [[nodiscard]] const T* getIf() const;
+
+private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    EventType type{}; //!< Type of the event
+    std::variant<Empty,
+                 Closed,
+                 Resized,
+                 FocusLost,
+                 FocusGained,
+                 TextEntered,
+                 KeyPressed,
+                 KeyReleased,
+                 MouseWheelScrolled,
+                 MouseButtonPressed,
+                 MouseButtonReleased,
+                 MouseMoved,
+                 MouseEntered,
+                 MouseLeft,
+                 JoystickButtonPressed,
+                 JoystickButtonReleased,
+                 JoystickMoved,
+                 JoystickConnected,
+                 JoystickDisconnected,
+                 TouchBegan,
+                 TouchMoved,
+                 TouchEnded,
+                 SensorChanged>
+        m_data; //!< Event data
 
-    union
+    ////////////////////////////////////////////////////////////
+    // Helper functions
+    ////////////////////////////////////////////////////////////
+    template <typename T, typename... Ts>
+    static constexpr bool isInParameterPack(const std::variant<Ts...>&)
     {
-        SizeEvent      size;      //!< Size event parameters (Event::Resized)
-        KeyEvent       key;       //!< Key event parameters (Event::KeyPressed, Event::KeyReleased)
-        TextEvent      text;      //!< Text event parameters (Event::TextEntered)
-        MouseMoveEvent mouseMove; //!< Mouse move event parameters (Event::MouseMoved)
-        MouseButtonEvent mouseButton; //!< Mouse button event parameters (Event::MouseButtonPressed, Event::MouseButtonReleased)
-        MouseWheelScrollEvent mouseWheelScroll; //!< Mouse wheel event parameters (Event::MouseWheelScrolled)
-        JoystickMoveEvent     joystickMove;     //!< Joystick move event parameters (Event::JoystickMoved)
-        JoystickButtonEvent joystickButton; //!< Joystick button event parameters (Event::JoystickButtonPressed, Event::JoystickButtonReleased)
-        JoystickConnectEvent joystickConnect; //!< Joystick (dis)connect event parameters (Event::JoystickConnected, Event::JoystickDisconnected)
-        TouchEvent  touch;  //!< Touch events parameters (Event::TouchBegan, Event::TouchMoved, Event::TouchEnded)
-        SensorEvent sensor; //!< Sensor event parameters (Event::SensorChanged)
-    };
+        return (std::is_same_v<T, Ts> || ...);
+    }
+
+    template <typename T>
+    static constexpr bool isEventType = isInParameterPack<T>(decltype(m_data)());
 };
 
-} // namespace sf
+#include <SFML/Window/Event.inl>
 
-// NOLINTEND(cppcoreguidelines-pro-type-member-init)
+} // namespace sf
 
 
 ////////////////////////////////////////////////////////////
@@ -235,30 +333,32 @@ struct Event
 ///
 /// A sf::Event instance contains the type of the event
 /// (mouse moved, key pressed, window closed, ...) as well
-/// as the details about this particular event. Please note that
-/// the event parameters are defined in a union, which means that
-/// only the member matching the type of the event will be properly
-/// filled; all other members will have undefined values and must not
-/// be read if the type of the event doesn't match. For example,
-/// if you received a KeyPressed event, then you must read the
-/// event.key member, all other members such as event.mouseMove
-/// or event.text will have undefined values.
+/// as the details about this particular event. Each event
+/// corresponds to a different struct which contains the data
+/// required to process that event.
 ///
-/// Usage example:
+/// The way to access the current active event is via
+/// sf::Event::getIf. This member function returns the address
+/// of the event struct if the event type matches the active
+/// event, otherise it returns nullptr. sf::Event::is is used
+/// to check the active event type without actually reading
+/// any of the corresponding event data.
+///
 /// \code
 /// for (sf::Event event; window.pollEvent(event);)
 /// {
 ///     // Request for closing the window
-///     if (event.type == sf::Event::Closed)
+///     if (event.is<sf::Event::Closed>())
 ///         window.close();
 ///
 ///     // The escape key was pressed
-///     if ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Key::Escape))
-///         window.close();
+///     if (const auto* keyPressed = event.getIf<sf::Event::KeyPressed>())
+///         if (keyPressed->code == sf::Keyboard::Key::Escape)
+///             window.close();
 ///
 ///     // The window was resized
-///     if (event.type == sf::Event::Resized)
-///         doSomethingWithTheNewSize(event.size.width, event.size.height);
+///     if (const auto* resized = event.getIf<sf::Event::Resized>())
+///         doSomethingWithTheNewSize(resized->size);
 ///
 ///     // etc ...
 /// }
--- a/include/SFML/Window/Event.inl
+++ b/include/SFML/Window/Event.inl
@@ -1,0 +1,60 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2024 Laurent Gomila (laurent@sfml-dev.org)
+//
+// This software is provided 'as-is', without any express or implied warranty.
+// In no event will the authors be held liable for any damages arising from the use of this software.
+//
+// Permission is granted to anyone to use this software for any purpose,
+// including commercial applications, and to alter it and redistribute it freely,
+// subject to the following restrictions:
+//
+// 1. The origin of this software must not be misrepresented;
+//    you must not claim that you wrote the original software.
+//    If you use this software in a product, an acknowledgment
+//    in the product documentation would be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such,
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+// Repeating isEventType<T> allows for cleaner compiler errors.
+// It is not strictly necessary but it's useful nonetheless.
+// It works by ensuring that the code within the conditional is
+// only compiled when the condition is met. Otherwise you get
+// a static_assert failure in addition to the compiler failing
+// to compile the code within the compiletime conditional when
+// an incorrect template parameter is provided.
+
+////////////////////////////////////////////////////////////
+template <typename T>
+Event::Event(const T& t)
+{
+    static_assert(isEventType<T>, "T must be a subtype of sf::Event");
+    if constexpr (isEventType<T>)
+        m_data = t;
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+bool Event::is() const
+{
+    static_assert(isEventType<T>, "T must be a subtype of sf::Event");
+    if constexpr (isEventType<T>)
+        return std::holds_alternative<T>(m_data);
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+const T* Event::getIf() const
+{
+    static_assert(isEventType<T>, "T must be a subtype of sf::Event");
+    if constexpr (isEventType<T>)
+        return std::get_if<T>(&m_data);
+}
--- a/src/SFML/Window/Android/WindowImplAndroid.cpp
+++ b/src/SFML/Window/Android/WindowImplAndroid.cpp
@@ -234,14 +234,14 @@ void WindowImplAndroid::forwardEvent(const Event& event)
     {
         ActivityStates& states = getActivity();
 
-        if (event.type == Event::GainedFocus)
+        if (event.is<Event::FocusGained>())
         {
             WindowImplAndroid::singleInstance->m_size = Vector2u(
                 Vector2i(ANativeWindow_getWidth(states.window), ANativeWindow_getHeight(states.window)));
             WindowImplAndroid::singleInstance->m_windowBeingCreated = true;
             WindowImplAndroid::singleInstance->m_hasFocus           = true;
         }
-        else if (event.type == Event::LostFocus)
+        else if (event.is<Event::FocusLost>())
         {
             WindowImplAndroid::singleInstance->m_windowBeingDestroyed = true;
             WindowImplAndroid::singleInstance->m_hasFocus             = false;
@@ -392,13 +392,10 @@ int WindowImplAndroid::processScrollEvent(AInputEvent* inputEvent, ActivityState
     lJNIEnv->DeleteLocalRef(objectMotionEvent);
 
     // Create and send our mouse wheel event
-    Event event;
-    event.type                   = Event::MouseWheelScrolled;
-    event.mouseWheelScroll.wheel = Mouse::Wheel::Vertical;
-    event.mouseWheelScroll.delta = static_cast<float>(delta);
-    event.mouseWheelScroll.x     = static_cast<int>(AMotionEvent_getX(inputEvent, 0));
-    event.mouseWheelScroll.y     = static_cast<int>(AMotionEvent_getY(inputEvent, 0));
-
+    Event::MouseWheelScrolled event;
+    event.wheel    = Mouse::Wheel::Vertical;
+    event.delta    = static_cast<float>(delta);
+    event.position = Vector2i(Vector2(AMotionEvent_getX(inputEvent, 0), AMotionEvent_getY(inputEvent, 0)));
     forwardEvent(event);
 
     // Detach this thread from the JVM
@@ -416,36 +413,28 @@ int WindowImplAndroid::processKeyEvent(AInputEvent* inputEvent, ActivityStates& 
     std::int32_t key     = AKeyEvent_getKeyCode(inputEvent);
     std::int32_t metakey = AKeyEvent_getMetaState(inputEvent);
 
-    Event event;
-    event.key.code    = androidKeyToSF(key);
-    event.key.alt     = metakey & AMETA_ALT_ON;
-    event.key.control = false;
-    event.key.shift   = metakey & AMETA_SHIFT_ON;
+    Event::KeyChanged keyChanged;
+    keyChanged.code    = androidKeyToSF(key);
+    keyChanged.alt     = metakey & AMETA_ALT_ON;
+    keyChanged.control = false;
+    keyChanged.shift   = metakey & AMETA_SHIFT_ON;
 
     switch (action)
     {
         case AKEY_EVENT_ACTION_DOWN:
-            event.type = Event::KeyPressed;
-            forwardEvent(event);
+            forwardEvent(Event::KeyPressed{keyChanged});
             return 1;
         case AKEY_EVENT_ACTION_UP:
-            event.type = Event::KeyReleased;
-            forwardEvent(event);
+            forwardEvent(Event::KeyReleased{keyChanged});
 
             if (auto unicode = static_cast<std::uint32_t>(getUnicode(inputEvent)))
-            {
-                event.type         = Event::TextEntered;
-                event.text.unicode = static_cast<std::uint32_t>(unicode);
-                forwardEvent(event);
-            }
+                forwardEvent(Event::TextEntered{static_cast<std::uint32_t>(unicode)});
             return 1;
         case AKEY_EVENT_ACTION_MULTIPLE:
             // Since complex inputs don't get separate key down/up events
             // both have to be faked at once
-            event.type = Event::KeyPressed;
-            forwardEvent(event);
-            event.type = Event::KeyReleased;
-            forwardEvent(event);
+            forwardEvent(Event::KeyPressed{keyChanged});
+            forwardEvent(Event::KeyReleased{keyChanged});
 
             // This requires some special treatment, since this might represent
             // a repetition of key presses or a complete sequence
@@ -457,8 +446,7 @@ int WindowImplAndroid::processKeyEvent(AInputEvent* inputEvent, ActivityStates& 
             }
             else if (auto unicode = static_cast<std::uint32_t>(getUnicode(inputEvent))) // This is a repeated sequence
             {
-                event.type         = Event::TextEntered;
-                event.text.unicode = static_cast<std::uint32_t>(unicode);
+                const Event event(Event::TextEntered{static_cast<std::uint32_t>(unicode)});
 
                 std::int32_t repeats = AKeyEvent_getRepeatCount(inputEvent);
                 for (std::int32_t i = 0; i < repeats; ++i)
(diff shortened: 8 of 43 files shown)


include/SFML/Window/Event.hpp
```diff
@@ -34,196 +34,298 @@
...
+    }
+
+    template <typename T>
+    void set(const T& t)
+    {
+        m_data = t;
+    }
+
+    template <typename T>
+    [[nodiscard]] const T& get() const
```
MarioLiebisch, line 292: I'm not sure about the verb `get`. Maybe I'm just more used to APIs using `is()` combined with `as()` (e.g. jsoncpp).

include/SFML/Window/Event.hpp
```diff
@@ -34,196 +34,298 @@
...
-        JoystickConnected,      //!< A joystick was connected (data in event.joystickConnect)
-        JoystickDisconnected,   //!< A joystick was disconnected (data in event.joystickConnect)
-        TouchBegan,             //!< A touch event began (data in event.touch)
-        TouchMoved,             //!< A touch moved (data in event.touch)
-        TouchEnded,             //!< A touch event ended (data in event.touch)
-        SensorChanged,          //!< A sensor value changed (data in event.sensor)
-
-        Count //!< Keep last -- the total number of event types
+        Empty,                  //!< Default type
+        Closed,                 //!< The window requested to be closed
```
MarioLiebisch, line 243: Since we're throwing around the whole event structure, I'd really love to change `closed`. I remember this confusing me from day one, since it doesn't signal the window is gone, it only signals the someone or something wanting to close the window. I'm just not sure about a new name. Not a real fan of it, but maybe something along the lines of `WindowCloseRequests`? (Also see next comment.)

include/SFML/Window/Event.hpp
```diff
@@ -34,196 +34,298 @@
...
+        Resized,                //!< The window was resized
+        LostFocus,              //!< The window lost the focus
+        GainedFocus,            //!< The window gained the focus
```
MarioLiebisch, line 246: Theseshould be changed/swapped to be more in line with other events:
* `WindowResized`
* `FocusLost`
* `FocusGained`

Also should there be extra events for more window manager requests like minimizing, maximizing, restoring?

include/SFML/Window/Event.hpp
```diff
@@ -34,196 +34,298 @@
...
+        TextEntered,            //!< A character was entered
+        KeyPressed,             //!< A key was pressed
+        KeyReleased,            //!< A key was released
+        MouseWheelScrolled,     //!< The mouse wheel was scrolled
+        MouseButtonPressed,     //!< A mouse button was pressed
+        MouseButtonReleased,    //!< A mouse button was released
```
MarioLiebisch, line 252: Think these could be simplified:

* `MouseScrolled`
* `MousePressed`
* `MouseReleased`

include/SFML/Window/Event.hpp
```diff
@@ -34,196 +34,298 @@
...
+        MouseMoved,             //!< The mouse cursor moved
+        MouseEntered,           //!< The mouse cursor entered the area of the window
+        MouseLeft,              //!< The mouse cursor left the area of the window
+        JoystickButtonPressed,  //!< A joystick button was pressed
+        JoystickButtonReleased, //!< A joystick button was released
+        JoystickMoved,          //!< The joystick moved along an axis
+        JoystickConnected,      //!< A joystick was connected
+        JoystickDisconnected,   //!< A joystick was disconnected
```
MarioLiebisch, line 260: Maybe replace `Joystick` with `Controller`, since actual gamepads are now way more common and it's a more generic term.

include/SFML/Window/Event.hpp
```diff
@@ -34,196 +34,298 @@
...
+        TouchBegan,             //!< A touch event began
+        TouchMoved,             //!< A touch moved
+        TouchEnded,             //!< A touch event ended
+        SensorChanged,          //!< A sensor value changed
     };
 
+    Event() = default;
+
+    template <typename T>
+    explicit Event(const T& t) : m_data(t)
```
MarioLiebisch, line 270: I know the `explicit` improves type safety here, but at the same time I also feel like being allowed to omit `sf::Event()` around a value feels more natural. I'm not fully up to date with template shenanigans, but I'm pretty sure there'd be some way more fancy approach for this.

examples/android/app/src/main/jni/main.cpp
```diff
@@ -115,39 +115,41 @@ int main(int argc, char* argv[])
     {
         for (sf::Event event; active ? window.pollEvent(event) : window.waitEvent(event);)
         {
-            switch (event.type)
+            switch (event.getType())
             {
-                case sf::Event::Closed:
+                case sf::Event::Type::Closed:
```
MarioLiebisch, line 120: I don't see a way around the extra `Type::` here, but I think this is still my least favorite part in this change.

examples/android/app/src/main/jni/main.cpp
```diff
@@ -115,39 +115,41 @@ int main(int argc, char* argv[])
...
                     active = false;
                     break;
-                case sf::Event::MouseEntered:
+                case sf::Event::Type::MouseEntered:
                     active = true;
                     break;
-                case sf::Event::TouchBegan:
-                    if (event.touch.finger == 0)
+                case sf::Event::Type::TouchBegan:
+                    const auto& touchBegan = even.get<sf::Event::TouchBegan>();
```
MarioLiebisch, line 149: `even`?

include/SFML/Window/Event.hpp
```diff
@@ -34,196 +34,298 @@
...
+    {
+    }
+
+    [[nodiscard]] Type getType() const
+    {
+        return static_cast<Type>(m_data.index());
+    }
+
+    template <typename T>
+    [[nodiscard]] bool is() const
```
MarioLiebisch, line 280: For this (and the other similar members) I feel like there should be a safety to explicitly prevent anyone from using `sf::Event::Type::...` here. Haven't tried it yet, but this feels like it would compile and silently fail?

include/SFML/Window/Event.hpp
ChrisThrasher: I was thinking the same thing. There's lots of `Event(event)` code that I didn't like writing.

The problem is that the template is currently unconstrained to removing `explicit` means the compiler will try to convert _anything_ into an Event. Kimci shared an idea for containing this template which I may look into.

include/SFML/Window/Event.hpp
ChrisThrasher: Just like the Event constructor this template ought to be constrained to just the types that exist in the underlying variant.

include/SFML/Window/Event.hpp
ChrisThrasher: So long as we still use the name `Joystick` in `sf::Joystick` I'd prefer we keep using that term here as well.

include/SFML/Window/Event.hpp
```diff
@@ -34,196 +34,298 @@
...
+    {
+        return std::holds_alternative<T>(m_data);
+    }
+
+    template <typename T>
+    void set(const T& t)
```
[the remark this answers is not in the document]
ChrisThrasher: Is the idea to then depend on the compiler-generated assignment operators?

include/SFML/Window/Event.hpp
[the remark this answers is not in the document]
ChrisThrasher: Yeah that's reasonable. It's valid for someone to want to do that.

include/SFML/Window/Event.hpp
```diff
@@ -34,196 +34,298 @@
...
     /// \brief Size events parameters (Resized)
     ///
     ////////////////////////////////////////////////////////////
-    struct SizeEvent
+    struct Resized
     {
-        unsigned int width;  //!< New width, in pixels
-        unsigned int height; //!< New height, in pixels
+        unsigned int width{};  //!< New width, in pixels
+        unsigned int height{}; //!< New height, in pixels
```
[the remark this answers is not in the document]
ChrisThrasher: I've wanted to do that for a while. I forget why but last time I tried that a long time ago it got messy. I'm open to trying that again since it would be really nice to remove most of these `width`/`height` and `x`/`y` pairs.

include/SFML/Window/Event.hpp
MarioLiebisch: I'd change both of course.😉

include/SFML/Window/Event.hpp
MarioLiebisch: Yeah, I'd definitely use some constraints in some form or shape.

ChrisThrasher:
I contained the templates so you can't just pass any random template argument into it.

include/SFML/Window/Event.hpp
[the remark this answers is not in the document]
MarioLiebisch: Agree, this might be unnecessary, but as long as there's a `get()` (name wise), having a `set()` (even with duplicated functionality) doesn't sound inherently wrong to me.

include/SFML/Window/Event.hpp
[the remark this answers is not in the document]
MarioLiebisch: I'd definitely go with `Vector2` here, even if it's just to have a simple way to pass it along other functions expecting a `Vector2` (same for other similar event data).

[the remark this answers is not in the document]
ChrisThrasher: I like this idea and want to play with it again but this PR is already pretty big so I'd rather not get into this as to ensure the scope of the PR doesn't expand too much more. This will be a great PR for later when we can make this change in isolation.

ChrisThrasher:
I removed `sf::Event::set` since it's just unnecessary and I'd rather not have too many ways of doing the same thing. I added a non-`const` `sf::Event::get` since otherwise manipulating the data in-place is not possible. You'd have to copy it out first before modifying it then reassigning it into the event.

include/SFML/Window/Event.hpp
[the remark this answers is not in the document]
ChrisThrasher: I added this but am not sure of any internal places we'd use it.

Bromeon (commented):
Thanks a lot for writing this PoC! :muscle: 

Looking at the event handling,
```cpp
switch (event.type) 
{
    case sf::Event::Type::KeyPressed:
        if (event.get<sf::Event::KeyPressed>().code == sf::Keyboard::Escape)
            window.close();
        break;
    case ...
}
```

I wonder if we could consider some alternative idioms to reduce repetition a bit?
Example:
```cpp
if (auto* key = event.as<sf::Event::KeyPressed>()) {
        if (key.code == sf::Keyboard::Escape)
            window.close();
}
// no break needed

else ...
```
or even:
```cpp
if (auto* key = event.keyPressed(); key && key.code == sf::Keyboard::Escape)
     window.close();
}
// no break needed

else ...
```
This would also avoid the age-old type safety problem: you cannot accidentally check for type X and process Y. The concrete type is directly returned.

ChrisThrasher:
I'm fine adding an analog to [`std::get_if`](https://en.cppreference.com/w/cpp/utility/variant/get_if) which would do what you're looking for. It's only a few lines of code and very easy to test and maintain as well. Would that work?

ChrisThrasher:
I added `sf::Event::get_if` and found some places internally to use it which help prove its value. If you're already using `if` statements then this is a very useful addition.

I'm still questioning whether I want to keep the non-`const` overloads of `get` and `get_if`. I can't find any use cases within this repo for using them.

ChrisThrasher:
https://github.com/ChrisThrasher/mandelbrot/compare/event_api

Quick experiment trying the new API in a project of mine. I love how I can still use a `switch`. The changes to my code are not drastic yet I'm getting a huge upgrade in type safety.

ChrisThrasher:
I pushed a commit that improves error messages. Previously you'd get a failed `static_assert` alongside a lot of `std::variant` template errors which were not necessary to understand the problem. Now thanks to `if constexpr` we can skip all those template errors and go straight to the failed `static_assert`. Using Clang 17 on my machine this results in about 1/5th as many lines of error text in the console.

ChrisThrasher:
https://github.com/SFML/imgui-sfml/compare/master...event_api

Here is what it looks like to apply this API to ImGui-SFML. Writing this uncovered a potential shortcoming of this API

```cpp
        case sf::Event::Type::KeyPressed: // fall-through
        case sf::Event::Type::KeyReleased: {
            const bool down = (event.type == sf::Event::KeyPressed);

            const ImGuiKey mod = keycodeToImGuiMod(event.key.code);
            // The modifier booleans are not reliable when it's the modifier
            // itself that's being pressed. Detect these presses directly.
            if (mod != ImGuiKey_None) {
                io.AddKeyEvent(mod, down);
            } else {
                io.AddKeyEvent(ImGuiKey_ModCtrl, event.key.control);
                io.AddKeyEvent(ImGuiKey_ModShift, event.key.shift);
                io.AddKeyEvent(ImGuiKey_ModAlt, event.key.alt);
                io.AddKeyEvent(ImGuiKey_ModSuper, event.key.system);
            }

            const ImGuiKey key = keycodeToImGuiKey(event.key.code);
            io.AddKeyEvent(key, down);
            io.SetKeyEventNativeData(key, event.key.code, -1);
        } break;
```
In the event where fall through behavior is desired, this new API requires duplicating the code inside each `case` or moving the code out into a function template. 

The only solution I can think to this is (other than having users simply deal with this) is to put both press and release events into one even with a new field for disambiguating between whether the key was pressed or released. I'm not sure I like this though.

Bambo-Borris:
Pointed one of my personal projects towards the `event_api` branch to try out these changes locally. Didn't notice any issues with mouse/keyboard events on X11. Had to change very little existing code, it is essentially like working with the old form. I found myself using the `sf::Event::getIf` method as it aligned with the existing event handling code I had. 

For someone used to SFML I think this kind of API change will be mostly unnoticed but still yielding better type safety. For new users I think this poses no more diffficulty to grasp than the old form, but with the benefit of better type safety. Think this is a great PoC for showing how SFML could modernise events, hope to see it get further along and potentially become the new way in SFML3.

include/SFML/Window/Event.hpp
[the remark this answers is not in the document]
ChrisThrasher: https://github.com/SFML/SFML/compare/event_api...resized_event

This change works very well in practice. I'm willing to add this to this PR but won't yet for the sake of keeping scope small. This change will certainly happen at some point though.

ChrisThrasher:
I took the remaining events that include `x`, `y`, and `z` fields and converted them to instead hold an `sf::Vector`. This has had downstream simplifying effects on anything using those events.

src/SFML/Window/macOS/WindowImplCocoa.mm
```diff
@@ -255,87 +239,50 @@ void showMouseCursor()
...
+    pushEvent(Event::MouseWheelScrolled{Mouse::VerticalWheel, deltaY, {x, y}});
 
-    event.type                   = Event::MouseWheelScrolled;
-    event.mouseWheelScroll.wheel = Mouse::HorizontalWheel;
-    event.mouseWheelScroll.delta = deltaX;
-    event.mouseWheelScroll.x     = x;
-    event.mouseWheelScroll.y     = y;
-    scaleOutXY(event.mouseWheelScroll, m_delegate);
-    pushEvent(event);
+    scaleOutWidthHeight(x, y, m_delegate);
```
kimci86, line 268: Bug here: this is scaling the same variables a second time. You can just remove the line.

src/SFML/Window/DRM/InputImpl.cpp
```diff
@@ -409,22 +406,36 @@ bool eventProcess(sf::Event& event)
...
-                        event.key.system   = systemDown();
+                        if (inputEvent.value)
+                        {
+                            sf::Event::KeyPressed keyPressed;
+                            keyPressed.code     = kb;
+                            keyPressed.scancode = sf::Keyboard::Scan::Unknown; // TODO: not implemented
+                            keyPressed.alt      = altDown();
+                            keyPressed.control  = controlDown();
+                            keyPressed.shift    = shiftDown();
+                            keyPressed.system   = systemDown();
```
kimci86, line 425: This can be done into a `Event::KeyChanged` instance to lower code duplication with else branch.

src/SFML/Window/Android/WindowImplAndroid.cpp
```diff
@@ -534,42 +527,33 @@ int WindowImplAndroid::processPointerEvent(bool isDown, AInputEvent* inputEvent,
     {
         if (device == AINPUT_SOURCE_MOUSE)
         {
-            event.type               = Event::MouseButtonPressed;
-            event.mouseButton.button = static_cast<Mouse::Button>(id);
-            event.mouseButton.x      = x;
-            event.mouseButton.y      = y;
+            event = Event::MouseButtonPressed{{static_cast<Mouse::Button>(id), {x, y}}};
```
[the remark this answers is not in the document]
ChrisThrasher: https://github.com/SFML/SFML/actions/runs/6802085671/job/18494421323#step:15:61

CI required this :(

Not sure why it's only the Android jobs that deal with this though.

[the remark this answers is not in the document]
kimci86: Checking on compiler explorer, it looks like it was a fixed in clang 13 and android jobs use clang 12.0.8 :frowning_face:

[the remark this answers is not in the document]
ChrisThrasher: https://github.com/SFML/SFML/actions/runs/6802085671/job/18494421323#step:14:20

This Android job is using Clang 12 which is older than I would have expected to see in our CI pipeline. I think this was a Clang bug that got fixed in either version 13 or 14. Upgrading our Android CI toolchains might do the trick.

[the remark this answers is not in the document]
ChrisThrasher: I disabled `-Wmissing-braces` on Android to work around this for now.

ChrisThrasher:
I wrote most of the remaining docs required. Along the way I also removed all function definitions from inside of Event.hpp which entailed creating Event.cpp where we define the one non-templated member function. This is required to avoid multiple definitions of the same function.

ChrisThrasher:
I revamped the docs in Event.hpp to better illustrate the main two different ways you can use this API to process events. Those docs should provide users with a clear picture of how to use this new API.

ChrisThrasher:
I moved the definitions of the function templates to a `.inl` file to match how we define templates for other types.

include/SFML/Window/Event.hpp
eXpl0it3r: I think the current naming is more specific and more aligned with the joystick event names

include/SFML/Window/Event.inl
```diff
@@ -0,0 +1,67 @@
...
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+const T* Event::getIf() const
+{
+    if constexpr (isEventType<T>)
+        return std::get_if<T>(&m_data);
+    else
```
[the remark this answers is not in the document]
ChrisThrasher: It's there for better error messages. If you remove that `else` then you get dramatically longer compiler errors that make it hard to find the failed `static_assert`. We trade some extra library complexity for better user ergonomics.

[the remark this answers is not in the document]
kimci86: I think eXpl0it3r suggests to remove the else but keep the static assert and that would work the same.
Maybe it would be more readable this way:
```c++
static_assert(isEventType<T>, "T must be a subtype of sf::Event");
if constexpr (isEventType<T>)
    return std::get_if<T>(&m_data);
```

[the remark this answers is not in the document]
ChrisThrasher: Oh okay. I just tried that and yes that works! Great idea.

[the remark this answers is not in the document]
Bromeon: Is the `if constexpr` even needed if the condition has been checked one line before?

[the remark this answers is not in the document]
ChrisThrasher: It’s not necessary but it produces better error messages when the template parameter is invalid.

Jim-Marsden (approved):

Bromeon:
Responding to the earlier questions:

> I'm fine adding an analog to [`std::get_if`](https://en.cppreference.com/w/cpp/utility/variant/get_if) which would do what you're looking for. It's only a few lines of code and very easy to test and maintain as well. Would that work?
>
> I added `sf::Event::getIf` and found some places internally to use it which help prove its value. If you're already using `if` statements for event handling then this is a very useful addition.

The `getIf` is a good start, although I'm still interested what you guys think about specific methods that would allow
```cpp
if (auto* key = event.keyPressed(); key && key.code == sf::Keyboard::Escape)
     window.close();
}
```
instead of
```cpp
if (auto* key = event.getIf<sf::Event::KeyPressed>()) {
        if (key.code == sf::Keyboard::Escape)
            window.close();
}
```
Considering that this is unlikely to be used in generic contexts 🙂 

---

> I'm still questioning whether I want to keep the non-`const` overloads of `get` and `getIf`. I can't find any use cases within this repo for using them.

I don't think this is needed, events are almost always read, not written.

ChrisThrasher:
```cpp
if (auto* key = event.keyPressed(); key && key.code == sf::Keyboard::Escape)
     window.close();
}

if (auto* key = event.getIf<sf::Event::KeyPressed>(); key && key.code == sf::Keyboard::Escape)
     window.close();
}
```
Both of the above snippets are possible, FYI. No need for that extra inner `if` when using `getIf`.

I'll speak to why I chose `get<T>`. In no particular order:

1. The function template solution is very easy to implement since we're just grabbing the template argument from the underlying variant. I wasn't keen on writing tons of functions for each function type.
2. I care a lot about users have an easy time knowing what how `sf::Event::Type` enumerations map to the correct function call to access that event type. I thought this template solution made that mapping easy to understand since `sf::Event::Type::Foo` always corresponds to a template parameter of `sf::Event::Foo`.
3. I want to support returning references. When returning a reference we have the chance to `assert` when the user calls the wrong function which provides a convenient automatically correctness check. I'm not keen on an API that exclusively returns pointers.

I don't hate the idea of having a specific member function for each event type but I'm not yet convinced I want to switch to that. If enough of the team prefers named functions then we can do that.

ChrisThrasher:

Jim-Marsden:
I personally prefer the get<...>() and getIf<...>() over wrapped functions.

1. As stated, it shadows std::variant
2. As it's a template, it shouldn't add much to the runtime costs. It maybe inline, or other compile time optimizations may be better applied over a non templated function. 
3. In a strange way, I think it would make my code that consumes SFML more consistent.

In fairness, it does seem like SFML does avoid template parameters. I can't think of a single place where you'd use them, and the places you could, SFML provides aliases.

ChrisThrasher:

Bromeon (approved):
I think this is an improvement over the current state. Unless there is concrete criticism, we should merge it so @ChrisThrasher doesn't have to keep rebasing for a year 😬 

---

There are still some things that deserve follow-up discussion, but this can be done in separate issues/PRs, in my opinion. Mostly:

1. Like already mentioned:
    > If enough of the team prefers named functions then we can do that. You and I have spoken plenty about this, but no one else from the team has commented on #2668 so I'm not sure what they think about the various different designs.

   Would really be good to hear opinions from other team members here.

2. Do we want to keep `sf::Event::Type::Closed` that long? Should we consider alternatives like `sf::EventType::Closed`?

3. Now that there are multiple "styles" of usage, which one should be recommended in docs/tutorials? The `switch`/`case` one still brings quite some error potential.

---

Btw, not having tagged unions really sucks sometimes in C++. The amount of discussion and workarounds we need for things like these, that would be solved in a few hours in a language like [Kotlin](https://kotlinlang.org/docs/sealed-classes.html#sealed-classes-and-when-expression), [Swift](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/enumerations/) or [Rust](https://doc.rust-lang.org/book/ch06-01-defining-an-enum.html). `std::variant` is a joke in comparison 😕

ChrisThrasher:
Rebased on `master`. Found a few more places that we can utilize vector operations thanks to the event types using vectors more as well.

ChrisThrasher:
Rebased on `master` and fixed merge conflicts

src/SFML/Window/Android/WindowImplAndroid.cpp
```diff
@@ -535,42 +522,40 @@ int WindowImplAndroid::processPointerEvent(bool isDown, AInputEvent* inputEvent,
...
         if (device == AINPUT_SOURCE_MOUSE)
         {
+<<<<<<< HEAD
             event.type               = Event::MouseButtonReleased;
             event.mouseButton.button = button;
             event.mouseButton.x      = x;
             event.mouseButton.y      = y;
+=======
+            event = Event::MouseButtonReleased{static_cast<Mouse::Button>(id), {x, y}};
+>>>>>>> ada0a21e3 (Rewrite `sf::Event` API to improve type safety)
```
kimci86, line 551: Merge conflit
```suggestion
            event = Event::MouseButtonReleased{button, {x, y}};
```

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1929,19 +1907,25 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
...
                     case Button3: event.mouseButton.button = Mouse::Button::Right;  break;
                     case 8:       event.mouseButton.button = Mouse::Button::Extra1; break;
                     case 9:       event.mouseButton.button = Mouse::Button::Extra2; break;
+=======
+                    case Button1: event.button = Mouse::Button::Left;     break;
+                    case Button2: event.button = Mouse::Button::Middle;   break;
+                    case Button3: event.button = Mouse::Button::Right;    break;
+                    case 8:       event.button = Mouse::Button::XButton1; break;
+                    case 9:       event.button = Mouse::Button::XButton2; break;
+>>>>>>> ada0a21e3 (Rewrite `sf::Event` API to improve type safety)
```
kimci86, line 1928: merge conflict
```suggestion
                    case Button1: event.button = Mouse::Button::Left;     break;
                    case Button2: event.button = Mouse::Button::Middle;   break;
                    case Button3: event.button = Mouse::Button::Right;    break;
                    case 8:       event.button = Mouse::Button::Extra1;   break;
                    case 9:       event.button = Mouse::Button::Extra2;   break;
```

ChrisThrasher:
Rebased on `master` and fixed merge conflicts

ChrisThrasher:

include/SFML/Window/Event.inl
```diff
@@ -0,0 +1,64 @@
...
+//
+////////////////////////////////////////////////////////////
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+Event::Event(const T& t)
+{
+    static_assert(isEventType<T>, "T must be a subtype of sf::Event");
+    if constexpr (isEventType<T>)
```
vittorioromeo, line 31: Probably missing something, but why have the `if constexpr` at all if the `static_assert` above performs the same check?

include/SFML/Window/Event.hpp
```diff
@@ -34,195 +34,350 @@
...
+    ///
+    ////////////////////////////////////////////////////////////
+    template <typename T>
+    [[nodiscard]] const T* getIf() const;
+
+private:
+    ////////////////////////////////////////////////////////////
+    // Member data
+    ////////////////////////////////////////////////////////////
+    VariantType m_data; //!< Event data
```
vittorioromeo, line 375: At the risk of losing some encapsulation, it would be nice to have a public getter that returns the raw `VariantType&` to the user. This would allow users to use `std::visit` if they so desire. Could be a separate PR, though.

ChrisThrasher: See the PR description for why I don't want to add that.

include/SFML/Window/Event.inl
ChrisThrasher: I addressed that previously in this thread. TL;DR It allows for better compiler diagnostics when users do something incorrect.

vittorioromeo: > I addressed that previously in this thread. TL;DR It allows for better compiler diagnostics when users do something incorrect.

Probably worth adding a comment in the code itself, I can imagine a future maintainer having the same question and possibly removing the redundant-appearing checks.

include/SFML/Window/Event.hpp
vittorioromeo: This?

> You must visit a variant with the perfect set of callbacks for the elements in the variant. If we allow users to visit this variant then we're effectively preventing ourselves from adding new events in a given major release. For now I'd prefer to retain that ability. We can always add this later if desired.

I'm not sure I get it. If a user decides to use `std::visit` with an exhaustive set of callbacks, they are intentionally saying *"please break my build if a new event is added, because I care about exhaustiveness"*. They are using `std::visit` exactly to get notified that something new was added.

Note that a user can still use a catch-all callback in a `std::visit` if they don't care about future-proof exhaustiveness:

```cpp
std::visit(overload{
    [](const A&) { ... },
    [](const B&) { ... },
    [](const auto&) { ...fallback... },
})
```

src/SFML/Window/WindowImpl.cpp
```diff
@@ -277,9 +273,10 @@ void WindowImpl::processJoystickEvents()
                 if (prevPressed ^ currPressed)
                 {
                     Event event;
-                    event.type = currPressed ? Event::JoystickButtonPressed : Event::JoystickButtonReleased;
-                    event.joystickButton.joystickId = i;
-                    event.joystickButton.button     = j;
+                    if (currPressed)
+                        event = Event::JoystickButtonPressed{i, j};
+                    else
+                        event = Event::JoystickButtonReleased{i, j};
```
vittorioromeo, line 279: How about:

```cpp
if (currPressed)
    pushEvent(Event::JoystickButtonPressed{i, j});
else
    pushEvent(Event::JoystickButtonReleased{i, j});
```

?

tools/xcode/templates/SFML/SFML App.xctemplate/main.cpp
```diff
@@ -67,13 +67,13 @@ int main()
...
             // Close window: exit
-            if (event.type == sf::Event::Closed)
+            if (event.is<sf::Event::Closed>())
             {
                 window.close();
             }
 
             // Escape pressed: exit
-            if (event.type == sf::Event::KeyPressed && event.key.code == sf::Keyboard::Key::Escape)
+            if (event.is<sf::Event::KeyPressed>() && event.get<sf::Event::KeyPressed>().code == sf::Keyboard::Key::Escape)
```
vittorioromeo, line 76: Maybe:

```suggestion
            if (const auto* e = event.getIf<Event::KeyPressed>(); e && e.code == sf::Keyboard::Key::Escape)
```

include/SFML/Window/Event.inl
ChrisThrasher: I don't plan on abandoning the project plus I read every PR so I'm not particularly worried about someone removing this without me realizing it. A comment isn't a bad idea though.

include/SFML/Window/Event.hpp
ChrisThrasher: My impression was that visiting a variant requires overloads for all types. Some quick tests confirms that suspicion of mine. What I think I was missing was your `const auto&` overload. It's basically a `default:` case in a switch statement. Nobody would say that adding a new value to an enum constitutes an API break just because someone chose to write a `switch` without a `default:` case.

This may suffice to assuage my concerns about supporting `std::visit` since we can point to this if and when a user complains that a new event type causes their program to stop compiling.

tools/xcode/templates/SFML/SFML App.xctemplate/main.cpp
ChrisThrasher: Fine by me I suppose

include/SFML/Window/Event.hpp
ChrisThrasher: https://github.com/SFML/SFML/compare/event_api...visit_event

Here's what that would look like

Bromeon: It's not just a forward-compatibility hazard in terms of adding new events, but at the off chance we'd ever use a different implementation than `std::variant`. We would lock ourselves into this specific implementation, which begs the question of why having an abstraction in the first place.

I would expose the internal variant only if there is concrete demand, not prematurely. It's easy to add this, hard to remove. And if there is a concrete, real-world use case, we can also try to understand the usage pattern better and consider if the existing API can be improved somehow.

ChrisThrasher: Agreed. I'm glad to know that supporting `std::visit` is in fact feasible but I'd like to see more proof of usefulness before we expose such an implementation detail. As demonstrated this is really easy to add at any point in the future.

ChrisThrasher:
Rebased on `master` since we just merged #1453 which was a large PR. No changes required. Only changes since my last comments were to add some missing comment blocks to better describe some implementation details.

tools/xcode/templates/SFML/SFML CLT.xctemplate/main.cpp
```diff
@@ -65,13 +65,13 @@ int main()
...
             // Close window: exit
-            if (event.type == sf::Event::Closed)
+            if (event.is<sf::Event::Closed>())
             {
                 window.close();
             }
 
             // Escape pressed: exit
-            if (event.type == sf::Event::KeyPressed && event.key.code == sf::Keyboard::Key::Escape)
+            if (event.is<sf::Event::KeyPressed>() && event.get<sf::Event::KeyPressed>().code == sf::Keyboard::Key::Escape)
```
vittorioromeo, line 74: I suggest using `getIf` + C++17 `if` with initializer here.

src/SFML/Window/WindowImpl.cpp
vittorioromeo: @ChrisThrasher: you put thumbs-up and resolved the comment, but I don't see the suggestion applied. Did you change your mind?

include/SFML/Window/Event.inl
vittorioromeo: @ChrisThrasher: ditto, you seemed in favour of the idea of adding a comment, and the conversation was resolved, yet I do not see the suggestion applied.

examples/opengl/OpenGL.cpp
```diff
@@ -210,21 +210,21 @@ int main()
...
-                if (event.type == sf::Event::Closed)
+                if (event.is<sf::Event::Closed>())
                 {
                     exit = true;
                     window.close();
                 }
 
                 // Escape key: exit
-                if ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Key::Escape))
+                if (event.is<sf::Event::KeyPressed>() && event.get<sf::Event::KeyPressed>().code == sf::Keyboard::Key::Escape)
```
vittorioromeo, line 220: I suggest using `getIf` + C++17 `if` with initializer here.

examples/opengl/OpenGL.cpp
```diff
@@ -210,21 +210,21 @@ int main()
...
                 {
                     exit = true;
                     window.close();
                 }
 
                 // Return key: toggle mipmapping
-                if ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Key::Enter))
+                if (event.is<sf::Event::KeyPressed>() && event.get<sf::Event::KeyPressed>().code == sf::Keyboard::Key::Enter)
```
vittorioromeo, line 227: I suggest using `getIf` + C++17 `if` with initializer here.

examples/opengl/OpenGL.cpp
```diff
@@ -241,14 +241,14 @@ int main()
                 }
 
                 // Space key: toggle sRGB conversion
-                if ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Key::Space))
+                if (event.is<sf::Event::KeyPressed>() && event.get<sf::Event::KeyPressed>().code == sf::Keyboard::Key::Space)
```
vittorioromeo, line 244: I suggest using `getIf` + C++17 `if` with initializer here.

tools/xcode/templates/SFML/SFML CLT.xctemplate/main.cpp
ChrisThrasher: I was fine changing the first instance you pointed out but don't want to change them all.

include/SFML/Window/Event.inl
```diff
@@ -0,0 +1,64 @@
...
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
+const T& Event::get() const
+{
+    static_assert(isEventType<T>, "T must be a subtype of sf::Event");
+    assert(is<T>() && "Event::get() Cannot access an inactive event type");
+    if constexpr (isEventType<T>)
+        return std::get<T>(m_data);
```
vittorioromeo, line 53: Consider: 

```suggestion
        return *std::get_if<T>(&m_data);
```

This likely generates better code and results in UB in case `!is<T>()`, which exposes even less details about variant and prevents users from thinking that "catching" the exception might be a way of handling the error.

ChrisThrasher:
> @ChrisThrasher: you put thumbs-up and resolved the comment, but I don't see the suggestion applied. Did you change your mind?

@vittorioromeo I did make that change. There was more than one instance of this particular pattern in WindowImpl.cpp and I didn't change the second because I didn't know it existed.

vittorioromeo:
Here are my thoughts about this PR:

1. It's a massive improvement over the status quo. Kudos.

2. The choice of using `.is<T>`, `.get<T>`, and `.getIf<T>` is the right one. It is simple enough to understand for beginners, keeps SFML's internals clean and DRY, and it allow advanced users to generalize over `T` with templates/metaprogramming.

3. The `if (const auto* resized = event.getIf<sf::Event::Resized>())` pattern in particular is really ergonomic.

4. I really dislike the fact that we're keeping the enumeration here. The rationale you've provided in the OP is not strong enough, in my opinion. We are offering two separate APIs to consume events that achieve the same result -- the enum-based one (error-prone, requires duplication both on the implementation side and user side) and the type-based one (the new API). I think this is a great opportunity to clean-up the events API and offer a single solution which is superior in most of the cases. Please reconsider!

Future ideas:

1. Either expose `std::variant`, or go full "encapsulation" mode and completely hide it from the header. The latter can be done by precalculating the size of a in-place `char` buffer, and emplacing/destroying the variant via `.cpp`. Benefit of the latter is that we do not expose `<variant>` in the header.

ChrisThrasher:
> I really dislike the fact that we're keeping the enumeration here. ... Please reconsider!

Sorry, but no. I strongly value keeping the ability to `switch` on all possible events. Otherwise you're forcing users to write giant `if`/`else if` chains that I find unsightly. If we drop that enum then what we're left with is an awkward wrapper around a `std::variant` at which point we should remove `sf::Event` and just return the variant directly.

vittorioromeo:
> There is no feasible way this code is removed without it going through me first so I'm not worried. I've read every single PR from the last >2 years and don't plan on stopping reading them.

1. It's not just about the code being removed. Anyone perusing SFML's codebase who is not you will be puzzled by the presence of the `if constexpr`. That will result in confusion, and wasted time in creating a PR/issue that you're end up going to reject because *you* know the reason why the `if constexpr` is there.

2. Life is unpredictable. Anyone of us might get hit by a bus tomorrow, you included. (Hopefully not!) I'd rather see a short comment `// redundant check to improve diagnostic quality` rather than rely on "insider knowledge".

ChrisThrasher:
> Anyone perusing SFML's codebase who is not you will be puzzled by the presence of the if constexpr.

This is a much more compelling argument. I'll add the comment then.

vittorioromeo:
> I strongly value keeping the ability to `switch` on all possible events. 

Why? If you want to check what's inside the event, you have to call `.get<T>` inside the branch of the `switch` anyway. If you only care about the type, the behavior is equivalent, the only thing that changes is the syntax.

The only counter-argument I can think of is that fall-through is sometimes useful, but it's a small advantage over a large list of disadvantages (coming in this post).

---

> Otherwise you're forcing users to write giant `if`/`else if` chains that I find unsightly. 

Finding something "unsightly" is not a strong technical argument. Even if `switch` is objectively better syntactically speaking (which is not -- it's a subjective consideration), there are real advantages in having a *single* new API that is superior to the previous one in 90% of the cases, instead of having two parallel APIs so that the legacy one can be worse in 90% of the cases but better in the remaining 10%.

`switch` is not a great construct to begin with -- it's equivalent to an `if...else` chain with different syntax, yet users want to pattern-match. Once you figure out the "type" of something, you still have to "get" it inside the corresponding branch, and you can easily make a mistake here if the `.get<T>` is mismatched with the type you switched on!

`switch` is also generally error-prone -- users can forget to put a `break;` where needed, and that's actually a very common case of mistakes for beginners.

Maintaining two separate APIs is also a bad idea for the SFML team. We have to make sure that event types and event enumerators are kept in sync. We have to maintain more code, and test two separate APIs.

Again, I strongly suggest to consider my perspective. It would be nice if other SFML team members would also weigh in on the matter.

> If we drop that enum then what we're left with is an awkward wrapper around a `std::variant` at which point we should remove `sf::Event` and just return the variant directly.

This is not a compelling technical argument either: you're saying that one reason for not removing the `enum` is so that `sf::Event` has more code that justifies it being it's own class compared to just using a `std::variant`.

That line of thinking seems backwards to me -- if the right thing to do is actually get rid of `sf::Event` and just use `std::variant`, we should do that. I'm not saying we should, I'm just criticizing the logic of the argument here.

Regardless, one possible advantage of keeping `sf::Event` without the enum is encapsulation, as multiple people have suggested. As I also suggested above, we could make `std::variant` entirely private and confined to the source file if we wanted to.

include/SFML/Window/Event.inl
ChrisThrasher: I'm cool with that. That way we can remove a reference to `std::bad_variant_access` from the header. We've set a precedent that the library won't go to lengths to guard against UB in release builds and since we're still protecting against UB in debug builds via the `assert` that's acceptable to me.

ChrisThrasher:
> Maintaining two separate APIs is also a bad idea for the SFML team. We have to make sure that event types and event enumerators are kept in sync. We have to maintain more code, and test two separate APIs.

Thanks to Kimci the implementation of this is very simple and easy to maintain. The tests already exists so future maintenance will be a breeze.

> you're saying that one reason for not removing the enum is so that sf::Event has more code that justifies it being it's own class compared to just using a std::variant.

It's not what I'm saying. What I'm saying is that the current PR is the best I can make it. The next best solution is a raw `std::variant`. After that I'd rank what you're proposing 3rd. If we remove the enum the let's just drop down to a raw variant.

tools/xcode/templates/SFML/SFML CLT.xctemplate/main.cpp
vittorioromeo: It's exactly the same code as above. HOw is the one above worth changing but this one isn't?

ChrisThrasher: I didn't think the first was worth changing per se. I thought you only wanted that pattern change in one place so I obliged rather than argue about it. However now that you want this pattern changed in many places I'm pushing back a bit harder.

vittorioromeo: The reason why I believe the pattern I suggested is technically superior to the one you have here is that it prevents mismatches of the type `e.is<X>() && e.get<Y>()`. 

Even if a mismatch is not present, any reader of the source code needs to visually check if the two types are matching. The pattern I suggested prevents the mismatch from happening in the first place and prevents readers from having to double-check.

My recommendation is to apply that pattern wherever possible. 

If you do not agree with this recommendation, please explain why you believe that the current pattern is superior to the one I am suggesting, given my technical argumentations above.

If you still do not agree with the recommendation even after that, please revert the place that you have changed above to the pattern I consider inferior to at least maintain consistency in the codebase.

vittorioromeo:
> > I strongly value keeping the ability to `switch` on all possible events.
> 
> Why? If you want to check what's inside the event, you have to call `.get<T>` inside the branch of the `switch` anyway. If you only care about the type, the behavior is equivalent, the only thing that changes is the syntax.
> 
> The only counter-argument I can think of is that fall-through is sometimes useful, but it's a small advantage over a large list of disadvantages (coming in this post).
> 
> > Otherwise you're forcing users to write giant `if`/`else if` chains that I find unsightly.
> 
> Finding something "unsightly" is not a strong technical argument. Even if `switch` is objectively better syntactically speaking (which is not -- it's a subjective consideration), there are real advantages in having a _single_ new API that is superior to the previous one in 90% of the cases, instead of having two parallel APIs so that the legacy one can be worse in 90% of the cases but better in the remaining 10%.
> 
> `switch` is not a great construct to begin with -- it's equivalent to an `if...else` chain with different syntax, yet users want to pattern-match. Once you figure out the "type" of something, you still have to "get" it inside the corresponding branch, and you can easily make a mistake here if the `.get<T>` is mismatched with the type you switched on!
> 
> `switch` is also generally error-prone -- users can forget to put a `break;` where needed, and that's actually a very common case of mistakes for beginners.
> 
> Maintaining two separate APIs is also a bad idea for the SFML team. We have to make sure that event types and event enumerators are kept in sync. We have to maintain more code, and test two separate APIs.
> 
> Again, I strongly suggest to consider my perspective. It would be nice if other SFML team members would also weigh in on the matter.
> 
> > If we drop that enum then what we're left with is an awkward wrapper around a `std::variant` at which point we should remove `sf::Event` and just return the variant directly.
> 
> This is not a compelling technical argument either: you're saying that one reason for not removing the `enum` is so that `sf::Event` has more code that justifies it being it's own class compared to just using a `std::variant`.
> 
> That line of thinking seems backwards to me -- if the right thing to do is actually get rid of `sf::Event` and just use `std::variant`, we should do that. I'm not saying we should, I'm just criticizing the logic of the argument here.
> 
> Regardless, one possible advantage of keeping `sf::Event` without the enum is encapsulation, as multiple people have suggested. As I also suggested above, we could make `std::variant` entirely private and confined to the source file if we wanted to.

I would like to get some more opinions on this matter, as deciding whether we want to advertise and maintain two separate APIs that do the same exact thing is something that should be finalized prior to SFML3's release.

I have offered a few technical reasons why the `switch` API is inferior to the new one, but to quickly reiterate:
1. Users can `case X: event.get<Y>()` without realizing their mistake until run-time. The new API prevents this at compile-time.
2. Users can forget a `break;` and unintentionally fall-through. The new API also prevents this.

These two issues are possible **run-time** bugs that the new API prevents.

Another usability issue is that with two possible choices to achieve the same result, users will now have to pick one or the other, instead of being offered a single possible choice. This will lead to users not knowing what to pick and when, wasting time arguing/deciding on the best API to use, and so on. 

In terms of the implementation quality, any maintainer or reviewer would have to manually ensure that the enumeration and the types are kept in sync. Maybe it will never be a problem, but it creates more surface area for mistakes and more mental overhead when trying to understand how SFML internally works.

Another idea I had in mind, but haven't thought deeply about, is completely eliminating `.get<T>`, and only keeping `.is<T>` and `.getIf<T>`. Rationale being that "empty" event types only require `.is<T>` to be properly used, while in any other case you want to both check the type and the data, which `.getIf<T>` does without giving the user a chance to make a mistake. Maybe there are patterns that benefit from `.get<T>` being available, so consider this an idea that needs to be explored further.

@Bromeon @eXpl0it3r @kimci86 @binary1248: thoughts?

Bromeon:
I strongly agree with @vittorioromeo and pretty much expressed the same opinion [half a year ago](https://github.com/SFML/SFML/issues/2668#issuecomment-1694756334). Interestingly, our arguments were mostly the same, too.

> The other approach is shorter and easier to read, safer, works without repetition, requires fewer API symbols (no enum), and most importantly leaves a single way to do things. So if I read other SFML code, I immediately understand what it's doing and it's not 50/50 based on the convention used.
> 
> All that's speaking for `switch` at this point is that people have been used to it before. But is that alone worth the downsides?

To add a new argument:

In the real world it is extremely rare that people want to catch _all_ possible events. Users are interested in events relevant for their game. A game that doesn't use joysticks doesn't care about `JoystickDisconnected`; a non-mobile game doesn't care about `TouchMoved`.

Without the need for exhaustive matching, there is really nothing objective left to maintain `switch` + separate `enum`. But there are very objective reasons in favor of maintaining a single, unified API.

kimci86:
Overall I am not sure about what API I would prefer, but I wrote some thoughts below hoping it can help :)

As already discussed, writing if statements with `is` and `get` requires to repeat the type parameter, which is boring and error-prone. Errors can be found easily at run time with asserts though.
Similarly, switch statement with `getType` and `get` also requires repeating the same information twice.

It seems using `is` for unit types and `getIf` for other types allows to write if statements in a nice way without such repetition.

```c++
if(event.is<sf::Event::Closed>())
{
    // good
}
if (const auto* pressed = event.getIf<sf::Event::KeyPressed>(); pressed && pressed->code == sf::Keyboard::Key::A)
{
    // this is nice
}
```

Yet I am not a huge fan because I imagine it can go wrong in some ways:
```c++
if (const auto* pressed = event.getIf<sf::Event::KeyPressed>(); pressed->code == sf::Keyboard::Key::A)
{
    // oops, condition is incomplete, can be dereferencing a null pointer
}
```
or, less obviously:
```c++
if (const auto* pressed = event.getIf<sf::Event::KeyPressed>(); pressed && pressed->code == sf::Keyboard::Key::A)
{
    // ok
}
else if (const auto* released = event.getIf<sf::Event::KeyReleased>(); pressed && pressed->code == sf::Keyboard::Key::B)
{
    // bad copy paste
    // this code was intented to run when releasing B but it will run when pressing B.
}
```
I don't think such mistakes would be as easy to make with `is` and `get`.

-----

What about avoiding `getIf`'s pointer check with a template function like so?
```c++
event.doIf<sf::Event::KeyPressed>(
    [&](const auto& pressed)
    {
        // I feel safe and comfy in here.
        if (pressed.code == sf::Keyboard::Key::A)
        {
        }
    });
```
A downside is it interrupts control flow. (e.g. `return` is limited to the lambda)
Maybe this is overkill. Anyways, any of the suggested API's is a huge improvement over the raw unsafe union.

vittorioromeo:
@kimci86 

> Yet I am not a huge fan because I imagine it can go wrong in some ways:
> 
> ```c++
> if (const auto* pressed = event.getIf<sf::Event::KeyPressed>(); pressed->code == sf::Keyboard::Key::A)
> {
>     // oops, condition is incomplete, can be dereferencing a null pointer
> }
> ```

This mistake is similar to calling `get` without calling `if` beforehand:

```cpp
if (event.get<sf::Event::KeyPressed>().code == sf::Keyboard::Key::A)
{
    // UB
}
```

Both APIs can be misused, but I think that the `getIf` pattern is less likely to cause an error in practice (the "if" in the name and pointer return type suggest that it should be checked). 

We could technically detect this error with assertions in a proxy type: https://gcc.godbolt.org/z/zqxYf51dq

But the drawback is that `auto*` wouldn't work anymore for deduction as we're returning a proxy type. However, it would be safer. Maybe we could have a `SafePtr<T>` abstraction in SFML that we use throughout the library for this purpose.

---

> or, less obviously:
> 
> ```c++
> if (const auto* pressed = event.getIf<sf::Event::KeyPressed>(); pressed && pressed->code == sf::Keyboard::Key::A)
> {
>     // ok
> }
> else if (const auto* released = event.getIf<sf::Event::KeyReleased>(); pressed && pressed->code == sf::Keyboard::Key::B)
> {
>     // bad copy paste
>     // this code was intented to run when releasing B but it will run when pressing B.
> }
> ```

You also have a similar copy-paste mistake with `is` and `get`:

```cpp
if (event.is<sf::Event::KeyPressed>() && event.get<sf::Event::KeyPressed>()->code == sf::Keyboard::Key::A)
{ 
    // ok
}
else if (event.is<sf::Event::KeyReleased>() && event.get<sf::Event::KeyPressed>()->code == sf::Keyboard::Key::A)
{ 
    // bad copy paste
    // undefined behavior!
}
```

---

> What about avoiding getIf's pointer check with a template function like so?

This would be the safest option, but unfortunately I think it's too unergonomic in practice to be a good choice. :(

---

> a pattern similar to doIf can be achieved by using only if statements with no init-statement. Then there is no need to explicitly check for the pointer being non null.

Yes, the main drawback of this pattern is the extra verbosity and added layer of indentation. Rust has a nice language feature that solves this, but is based on pattern matching not available in C++: https://doc.rust-lang.org/rust-by-example/flow_control/if_let.html

examples/window/Window.cpp
```diff
@@ -144,20 +144,21 @@ int main()
         for (sf::Event event; window.pollEvent(event);)
         {
             // Close window: exit
-            if (event.type == sf::Event::Closed)
+            if (event.is<sf::Event::Closed>())
                 window.close();
 
             // Escape key: exit
-            if ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Key::Escape))
+            if (event.is<sf::Event::KeyPressed>() && event.get<sf::Event::KeyPressed>().code == sf::Keyboard::Key::Escape)
```
vittorioromeo, line 151: Another opportunity to use `getIf` here.

include/SFML/Window/Clipboard.hpp
```diff
@@ -93,16 +93,17 @@ SFML_WINDOW_API void setString(const String& text);
 /// // or use it in the event loop
 /// for (sf::Event event; window.pollEvent(event);)
 /// {
-///     if(event.type == sf::Event::Closed)
+///     if(event.is<sf::Event::Closed>())
 ///         window.close();
-///     if(event.type == sf::Event::KeyPressed)
+///     if(event.is<sf::Event::KeyPressed>())
 ///     {
+///         const auto& keyPressed = event.get<sf::Event::KeyPressed>();
```
vittorioromeo, line 100: Another opportunity to use `getIf` here.

vittorioromeo:
Here is a visual comparison between the `switch` in the PR versus my proposed API that does not support `getType` anymore. I personally like the new proposed API more than the `switch`:

{6B7E60DD-C817-400E-8832-ADB64A8572CB}

Bromeon:
Small nitpick: even though it doesn't matter much in practice, the
```cpp
case sf::Event::Type::KeyPressed:
    if (event.get<sf::Event::KeyPressed>().code == sf::Keyboard::Key::Escape)
    {
        ...
    }
    break;
```
is not exactly equivalent to
```cpp
if (const auto* pressed = event.getIf<sf::Event::KeyPressed>();
    pressed && pressed->code == sf::Keyboard::Key::Escape)
{
   ...
}
```
In an if-else chain, the latter will check following conditions even if the event type _is_ `KeyPressed` but the code doesn't match. A nested `if` statement would be the better equivalent. As mentioned, it doesn't typically matter unless there are side effects or expensive computations in the other conditions, but I thought I'd mention it.

I'm not sure if we want to encourage the `; pressed && pressed->code ...` pattern or rather a nested `if` (which also doesn't suffer from the null-dereference problem).

ChrisThrasher:
Rebased on master and fixed minor merge conflict.

ChrisThrasher:
Periodic rebase. No changes made. Just waiting on a final decision to be made.

Bromeon:
I think most arguments have been stated; is there any input from @eXpl0it3r, @binary1248 or someone else? 🙂

vittorioromeo:
> I think most arguments have been stated; is there any input from @eXpl0it3r, @binary1248 or someone else? 🙂

A major point you missed is that we're offering two different user APIs that achieve the same result. Even if the newer one is not perfect, I see a great amount of value in having only a single "generally better" API offered to the end user.

binary1248:
I like it when my friend the compiler can prevent me from doing stupid things, but to do that it needs to have a sense of what "stupid" is. One of C++'s greatest strengths is its type system. I consider it a "parallel dimension" to the dimension in which we write our procedural code that eventually does its stuff at runtime. Everything we "program" into the type system is "run" at compile time thus its ability to catch programming errors before a binary is even built. I'm a fan of catching my mistakes as early as possible. In order for the compiler to understand there is a link between checking the type of an event and retrieving its contents we have to convey this relationship to it. This can only be done using the type system. It is impossible for the compiler to infer this link when we perform what for the compiler looks like a series of simple integral comparisons and execute follow up code based on the result. These comparisons look like any of the 1000s in the rest of the code base. The only way for errors to be caught in this scenario is at runtime which is a little too late for me.

Maybe it's due to the fact that I don't write many tests for my own projects, but I don't like to rely on unit tests to tell me that my code "seems" to be correct when I can instead leverage the type system to provide me with definitive guarantees. I prefer code that is correct by construction (bless Dijkstra) instead of code that is correct by exhaustion.

Based on this principle I have to support the `getIf` proposal. I must admit, I am also one of the guys who doesn't mind a `switch` statement here or there, maybe its just purely out of habit. However, in this case we are really stuck right in the middle between runtime programming and compile time programming and in such a case where there is a solution that can make use of the type system, I am more than willing to give up the good old `switch` for whatever is safer for me in the end. Look on the bright side, we're probably not the only ones who finds long if-else chains unsightly, thus the [upcoming pattern matching feature](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p2688r1.pdf) which if you ask me has a bit of a `switch`y look to it.

In reading the comments I haven't seen anyone mention the examples/tutorials. If we were really to have 2 ways of doing the same thing, how do we decide which we use in the examples/tutorials? At the bare minimum we should at least try to stay consistent. There is nothing worse than tutorial code not lining up with the examples we provide. I also don't think it is a good idea to provide examples/tutorials for both, so eventually we would have to pick a side anyway, indirectly "deprecating" the other. If there was a side to deprecate I would prefer the `switch` side. It has served us well for a long time, however it will not benefit from future C++ language updates nearly as much as the other solution would. Why chain ourselves to the past when we could set the foundation for what is to come in the future?

ChrisThrasher:

eXpl0it3r:
### API Design
From an API design point of view I have three comments:
- I do agree that having multiple ways to do the same-ish thing is bad and can/will cause a lot of confusion
- Adding APIs is a non-breaking change, as such I propose to pick one API for SFML 3 and if we notice a lot of ergonomics, we always have the possibility to expand the API surface in SFML 3.1. This of course would prevent certain nearly "free" upgrade paths from SFML 2 to 3.
- `getIf<T>()` is in my opinion not a simple API. It's not very complex either, but it would be one of the first and very central APIs that will require *some* understand of templates. And while C++ dev familiar with `std::variant` will be used to `get_if` it's really not a great speaking API. We're going from having to know if/switch/enums to if/"templates"/"variants"/optional. I wonder if there's a way to provide the "advanced" API in addition to a more speaking/simpler API, close to what @Bromeon proposed - consider how we provide `Vector2f` usings for the same simplicity in usage.

### Switch Ergonomics
Personally, I've often fallen back to `if` & `else if` blocks, mostly to circumvent the non-exhaustive enum/missing default case warnings. Given that for small projects (most common type of SFML projects) you rarely will handle all events and given that enabling these warnings are generally recommended, it disarms the argument in favor of `switch-case` ergonomics for me quite a bit.  
Keep in mind though, that I've a rather limited experience with C++, so maybe this view doesn't line up with what others experience.

### My Proposal
Seeing most of the voice speaking up for support of `getIf<T>` and the points listed above, I would propose to get with `getIf<T>` only in this PR as a first step.

I'd then be very interested in seeing potential proposals to "simplify" the API.

ChrisThrasher:
I removed `getType` and `get` from the API.

binary1248 (approved):

examples/island/Island.cpp
```diff
@@ -182,17 +182,17 @@ int main()
         for (sf::Event event; window.pollEvent(event);)
         {
             // Window closed or escape key pressed: exit
-            if ((event.type == sf::Event::Closed) ||
-                ((event.type == sf::Event::KeyPressed) && (event.key.code == sf::Keyboard::Key::Escape)))
+            if (event.is<sf::Event::Closed>() || (event.is<sf::Event::KeyPressed>() &&
+                                                  event.getIf<sf::Event::KeyPressed>()->code == sf::Keyboard::Key::Escape))
```
Bromeon, line 186: This is an interesting proposal of the formatter... 😬 

One way to keep it more explicit would be to add something like `//.` at the end of the first line, after the `||`. Not sure if it's worth it... Either way, definitely not a blocker, feel free to ignore 🙂

include/SFML/Window/Event.inl
Bromeon: This is still open; adding a comment would help this seeming duplication.

ChrisThrasher: There’s already a comment about this higher up in the file

include/SFML/Window/Event.hpp
```diff
@@ -34,195 +34,299 @@
...
+    struct SensorChanged
+    {
+        Sensor::Type type{}; //!< Type of the sensor
+        Vector3f     value;  //!< Current value of the sensor on the X, Y, and Z axes
+    };
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Default constructor
+    ///
+    /// Sets the event to Type::Empty
```
kimci86, line 247:
```suggestion
    /// Sets the event to sf::Event::Empty
```

include/SFML/Window/Event.hpp
```diff
@@ -235,30 +339,33 @@ struct Event
...
+/// corresponds to a different struct which contains the data
+/// required to process that event.
+///
+/// The way to access the current active event is via
+/// sf::Event::getIf which returns the address of the event struct
+/// if it's active or a nullptr otherwise. sf::Event::is is used
+/// to check the active event type without actually reading any
+/// of the corresponding event data. These are useful if you need
+/// to check only a single possible active event or prefer to
+/// process events in a series of if/else if blocks.
```
kimci86, line 352: Does not apply anymore as there are no other methods.
```suggestion
/// of the corresponding event data.
```

test/Window/Event.test.cpp
```diff
@@ -16,7 +16,253 @@ TEST_CASE("[Window] sf::Event")
...
+        const auto& touchBegan = *event.getIf<sf::Event::TouchBegan>();
+        CHECK(touchBegan.finger == 99);
+        CHECK(touchBegan.position == sf::Vector2i(98, 97));
+
+        event = sf::Event::TouchMoved{96, {95, 94}};
+        CHECK(event.is<sf::Event::TouchMoved>());
+        CHECK(event.getIf<sf::Event::TouchMoved>());
+        const auto& touchMoved = *event.getIf<sf::Event::TouchMoved>();
+        CHECK(touchMoved.finger == 96);
+        CHECK(touchBegan.position == sf::Vector2i(95, 94));
```
kimci86, line 167:
```suggestion
        CHECK(touchMoved.position == sf::Vector2i(95, 94));
```
funny that it worked nonetheless because being at the same address.

test/Window/Event.test.cpp
```diff
@@ -16,7 +16,253 @@ TEST_CASE("[Window] sf::Event")
...
+
+        event = sf::Event::TouchEnded{93, {92, 91}};
+        CHECK(event.is<sf::Event::TouchEnded>());
+        CHECK(event.getIf<sf::Event::TouchEnded>());
+        const auto& touchEnded = *event.getIf<sf::Event::TouchEnded>();
+        CHECK(touchEnded.finger == 93);
+        CHECK(touchBegan.position == sf::Vector2i(92, 91));
```
kimci86, line 174:
```suggestion
        CHECK(touchEnded.position == sf::Vector2i(92, 91));
```

ChrisThrasher: Oops I made this same copy paste but twice in this file actually

ChrisThrasher:
Incorporated Kimci's review comments and rebased on `master`.

kimci86 (approved):

ChrisThrasher:
Rebased on `master` and fixed merge conflicts.

include/SFML/Window/Event.hpp
```diff
@@ -34,195 +34,299 @@
...
+    [[nodiscard]] bool is() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Get particular event type
+    ///
+    /// \return Address of current event type, otherwise nullptr
+    ///
+    ////////////////////////////////////////////////////////////
+    template <typename T>
+    [[nodiscard]] const T* getIf() const;
```
MarioLiebisch, line 277: I'd prefer `tryGet()`, but there's no regular `get()` (returning a value). So I'd probably just stick to `get()`? But since it returns a pointer (and not some uninitialized value), I'd say this is obvious enough with a simple `get()`. But maybe that's just me.

So essentially one of these:
```cpp
template <typename T>
[[nodiscard]] const T* get() const;
```

```cpp
template <typename T>
[[nodiscard]] const &T get() const;
template <typename T>
[[nodiscard]] const T* tryGet() const;
```

include/SFML/Window/Event.hpp
```diff
@@ -235,30 +339,31 @@ struct Event
...
-/// if you received a KeyPressed event, then you must read the
-/// event.key member, all other members such as event.mouseMove
-/// or event.text will have undefined values.
+/// as the details about this particular event. Each event
+/// corresponds to a different struct which contains the data
+/// required to process that event.
+///
+/// The way to access the current active event is via
+/// sf::Event::getIf which returns the address of the event struct
+/// if it's active or a nullptr otherwise. sf::Event::is is used
```
MarioLiebisch, line 348: > [...] returns the address of the event struct if it's active or a nullptr otherwise.
This sounds really off. How about this one?
> [...] returns the address of the event struct, if the event type matches or is related, or a null pointer otherwise.

MarioLiebisch (approved):

include/SFML/Window/Event.hpp
MarioLiebisch: Ah, wasn't aware of that. However, it **does** offer a simple [`get()`](https://en.cppreference.com/w/cpp/utility/variant/get) in addition, too.

ChrisThrasher: My original and preferred API did include a `get` function and I hope we can add it one day. For now I still think it makes sense to align with the `std::variant` API since users may already be familiar with that.

include/SFML/Window/Event.hpp
```diff
@@ -34,195 +34,299 @@
...
+    ////////////////////////////////////////////////////////////
+    template <typename T>
+    [[nodiscard]] const T* getIf() const;
+
+private:
+    ////////////////////////////////////////////////////////////
+    /// \brief Alias for std::variant template specialization
+    ///
+    ////////////////////////////////////////////////////////////
+    using VariantType = std::variant<
```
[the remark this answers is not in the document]
ChrisThrasher: I found a way we can remove that alias entirely which sidesteps the naming problem and simplifies the implementation slightly.

include/SFML/Window/Event.hpp
```diff
@@ -34,195 +34,299 @@
...
-        TouchMoved,             //!< A touch moved (data in event.touch)
-        TouchEnded,             //!< A touch event ended (data in event.touch)
-        SensorChanged,          //!< A sensor value changed (data in event.sensor)
-
-        Count //!< Keep last -- the total number of event types
     };
 
     ////////////////////////////////////////////////////////////
-    // Member data
+    /// \brief Joystick button events
```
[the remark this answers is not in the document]
ChrisThrasher: I don't have strong feelings about the use of the name "joystick" and open to changing it if the team wants to do so. However, doing a rename like that will be pretty invasive and is better suited for a separate PR. Think of all the variable names and docs we'd have to update as well.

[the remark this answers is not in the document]
MarioLiebisch: Fair point.

Bromeon (approved):
This has been open for a long time and discussed even longer before that :slightly_smiling_face: 
I think it's in a good state to be merged; if there are more things to refine, they could be done as follow-up PRs.

Thanks for the initiative, as well as all the continuous updates and rebases @ChrisThrasher! :+1:

ChrisThrasher:
Rebased on `master` since some new clang-tidy checks were added. No changes made.

eXpl0it3r (approved):
I don't think, this will be the final form of the API for SFML 3, but let us collect some more feedback from people using it, by getting it into master.

Thank you for all the work put into @ChrisThrasher and all the discuss around it so far from everyone! I believe it's a good thing that we the "old" API is being challenged and forces us to reflect some more about the design and safety.

PR #2769, opened 2023-11-09
move semantics for sf::Context

tbetten:
## Description

## Tasks

-   [ ] Tested on Linux
-   [x] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    window.setFramerateLimit(60);
sf::Context context;
sf::Context other { context };

    while (window.isOpen())
    {
        {
                window.close();
        }

        window.clear();
        window.display();
    }
}
```


src/SFML/Window/Context.cpp
```diff
@@ -48,7 +48,7 @@ namespace sf
 ////////////////////////////////////////////////////////////
 Context::Context() : m_context(priv::GlContext::create())
 {
-    if (!setActive(true))
```
ChrisThrasher, line 51: Based on [your forum post](https://en.sfml-dev.org/forums/index.php?topic=29252.0) did you mean to put this check in the destructor?

src/SFML/Window/Context.cpp
```diff
@@ -64,6 +64,11 @@ Context::~Context()
     }
 }
 
+////////////////////////////////////////////////////////////
+Context::Context(Context&& other) noexcept = default;
+
+////////////////////////////////////////////////////////////
+Context& Context::operator=(Context&& other) noexcept = default;
```
ChrisThrasher, line 71: Please put two newlines between these. Look at the rest of this .cpp and try to copy the code style you see in there. Also remove the `other` parameter. Because it's not used let's not even name it. Remove it from the .hpp file too please.

test/Window/Context.test.cpp
```diff
@@ -32,6 +32,23 @@ TEST_CASE("[Window] sf::Context", runDisplayTests())
         CHECK(context.getSettings().majorVersion > 0);
     }
 
+    SECTION("Move Construction")
+    {
+        sf::Context first;
+        sf::Context second{std::move(first)};
```
ChrisThrasher, line 38:
```suggestion
        sf::Context second(std::move(first));
```
Please match SFML code style by using parentheses for calling constructors.

include/SFML/Window/Context.hpp
```diff
@@ -84,6 +84,9 @@ class SFML_WINDOW_API Context : GlResource
     ////////////////////////////////////////////////////////////
     Context& operator=(const Context&) = delete;
 
+    Context(Context&& other) noexcept;
+    Context& operator=(Context&& other) noexcept;
+
```
ChrisThrasher, line 89: Please match the code style of the rest of this file. There are comment blocks you're supposed to add. See something like Shader.hpp for what to include in those comment blocks.

ChrisThrasher (changes requested):

ChrisThrasher:
Now that `m_context` can be null we need `assert`s in all member functions that ensure that nobody tries to call these functions from a moved-from `sf::Context` instance. Here's a good example of what that should look like:

https://github.com/SFML/SFML/blob/39ba64cfc63866a868c94a41ca331c12235c190e/src/SFML/System/Vector2.cpp#L90C7-L90C7

ChrisThrasher:
If you're move-assigning into an existing context, don't you still need to run the code in the destructor before the moved-to context steals the state of the moved-from context? In general this is how you do move assignment for types with non-empty destructors. The moved-to type has to call the cleanup code in its destructor first before being assigned to. This is why I question whether default move operations are correct.

At this point in SFML's development I think we've already used default move operations in as many places as we can. Now we're left with a bunch of immovable types with custom destructors that likewise need custom move operations that are tricky to correctly implement.

src/SFML/Window/Context.cpp
kimci86, line 71: The default implementation is not correct. We need to deactivate the context before destroying it.
I think this would be a correct implementation:
```c++
Context& Context::operator=(Context&& other) noexcept
{
    if (&other == this)
        return *this;

    if (m_context && !setActive(false))
        err() << "Failed to set context as inactive during move assignment" << std::endl;

    m_context = std::move(other.m_context);

    return *this;
}```

ChrisThrasher:
@tbetten Do you plan on revisiting this PR?

ChrisThrasher:
Closed for inactivity

PR #2770, opened 2023-11-10
Provide an option to specify the number of parallel jobs clang-tidy should be run with

binary1248:
Title.

This would allow me to get rid of the patching step necessary to limit CPU usage when running CI using Buildbot.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -549,6 +549,7 @@ add_custom_target(format
     WORKING_DIRECTORY ${PROJECT_SOURCE_DIR} VERBATIM)
 
 sfml_set_option(CLANG_TIDY_EXECUTABLE clang-tidy STRING "Override clang-tidy executable, requires minimum version 14")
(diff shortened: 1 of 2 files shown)


ChrisThrasher:
Why does the build bot rub clang-tidy?

binary1248:
It always ran clang-tidy as part of its static analysis run, even before we started using GitHub Actions. After clang-tidy support was added to our CMake configuration buildbot would just use that instead of manually invoking clang-tidy like it originally did.

ChrisThrasher:
I think the Buildbot ought to stop running clang-tidy. Is it analyzing any codepaths that aren't already analyzed by our 6 clang-tidy jobs in GitHub Actions?

binary1248:
I would also have preferred to cut down on buildbot jobs, but @eXpl0it3r prefers that we keep everything redundant just in case.

ChrisThrasher:
Running clang-tidy is not hard. Should we ever somehow lose our GitHub Actions CI pipeline it is very easy to add it to our Buildbot pipeline. There is no need to be continuously running clang-tidy while we wait on GitHub to cut us off.

binary1248:
Disabled running clang-tidy on Buildbot for now.

PR #2771, opened 2023-11-10
Use `inline` to define constant in header

ChrisThrasher:


eXpl0it3r (approved):
I think that's correct, but didn't fully understand the reasoning and none was given otherwise.

ChrisThrasher:
The use of `inline` here is a C++17 feature that lets us avoid defining these constants in a particular translation unit.

include/SFML/Audio/SoundFileFactory.hpp
```diff
@@ -152,8 +152,8 @@ class SFML_AUDIO_API SoundFileFactory
     // Static member data
     ////////////////////////////////////////////////////////////
     // NOLINTBEGIN(readability-identifier-naming)
-    static ReaderFactoryArray s_readers; //!< List of all registered readers
-    static WriterFactoryArray s_writers; //!< List of all registered writers
+    static inline ReaderFactoryArray s_readers; //!< List of all registered readers
```
kimci86, line 155: I don't see the point. I would rather see this removed from header and put in anonymous namespace in cpp file.

ChrisThrasher: These constants are also used in SoundFileFactory.inl so moving them to the anonymous namespace of SoundFileFactory.cpp is not a possibility.

PR #2773, opened 2023-11-10
Replace `std::memcpy` with a more expressive alternative

ChrisThrasher:
## Description

Use `std::vector` member functions is easier to read and maintain.

### DIFF
--- a/examples/voip/Server.cpp
+++ b/examples/voip/Server.cpp
@@ -133,11 +133,9 @@ private:
                 // (so we protect any operation on it with the mutex)
                 {
                     const std::lock_guard lock(m_mutex);
-                    const std::size_t     oldSize = m_samples.size();
-                    m_samples.resize(oldSize + sampleCount);
-                    std::memcpy(&(m_samples[oldSize]),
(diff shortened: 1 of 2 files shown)


examples/voip/Server.cpp
```diff
@@ -133,11 +133,9 @@ class NetworkAudioStream : public sf::SoundStream
                 // (so we protect any operation on it with the mutex)
                 {
                     const std::lock_guard lock(m_mutex);
-                    const std::size_t     oldSize = m_samples.size();
-                    m_samples.resize(oldSize + sampleCount);
```
eXpl0it3r, line 137: Might want to replace that with a `reserve` call

ChrisThrasher: `std::vector::insert` will allocate more space if needed.

eXpl0it3r: I guess it's fine when we specify a begin/end range. For an InputIterator there doesn't seem to be a guarantee

eXpl0it3r (approved):

PR #2775, opened 2023-11-12
Added missing VS_DEBUGGER_WORKING_DIRECTORY to sfml_add_test macro

binary1248:
Added missing VS_DEBUGGER_WORKING_DIRECTORY to `sfml_add_test` macro since tests now rely on loading file resources.

Recently, tests were added that rely on loading file resources. Running a single test target outside of the test runner would not run it in the right working directory as expected by the paths provided in the test code. Setting `VS_DEBUGGER_WORKING_DIRECTORY` analogous to the examples fixes this problem.

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -350,6 +350,9 @@ function(sfml_add_test target SOURCES DEPENDS)
     # set the target flags to use the appropriate C++ standard library
     sfml_set_stdlib(${target})
 
+    # set the Visual Studio startup path for debugging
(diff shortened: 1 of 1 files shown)


ChrisThrasher:
Does this change only change the experience for those using Visual Studio?

binary1248:
Well... yeah... there is currently no other variable in CMake that takes care of other IDEs. Visual Studio users get the best CMake support of all IDEs.

ChrisThrasher (approved):
I see we're already doing this for the examples so there is precedent for this.

PR #2776, opened 2023-11-13
Add `sf::Rect<T>::getCenter()`

ChrisThrasher:
## Description

Inspired by a comment from Bromeon: https://github.com/SFML/SFML/pull/2692#discussion_r1329828733

I think this is a good convenience to give users. There are ever 2 places we needed it internally.

---

> Why not add `sf::Rect<T>::getEnd()` too?
I'm not opposed to this, but I could not find anywhere inside this repo that we would use it so I decided to hold off. I'm open to adding it if others think it's worth having.

### DIFF
--- a/include/SFML/Graphics/Rect.inl
+++ b/include/SFML/Graphics/Rect.inl
@@ -129,6 +129,14 @@ constexpr Vector2<T> Rect<T>::getSize() const
 
 ////////////////////////////////////////////////////////////
 template <typename T>
+constexpr Vector2<T> Rect<T>::getCenter() const
+{
+    return getPosition() + getSize() / T{2};
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
 constexpr bool operator==(const Rect<T>& left, const Rect<T>& right)
 {
     return (left.left == right.left) && (left.width == right.width) && (left.top == right.top) &&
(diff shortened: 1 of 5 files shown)


ChrisThrasher:
https://github.com/SFML/SFML/compare/rect_functions...templated_tests

After this I have some new changes including changes to the Rect<T> tests that ensure we do a better job testing our class templates for multiple types instead of just one. This is a really neat Catch2 feature that lets you write tests that get ran for multiple template parameters. It's great for testing class templates like the handful we have.

include/SFML/Graphics/Rect.inl
```diff
@@ -127,6 +127,14 @@ constexpr Vector2<T> Rect<T>::getSize() const
 }
 
 
+////////////////////////////////////////////////////////////
+template <typename T>
+constexpr Vector2<T> Rect<T>::getCenter() const
+{
+    return getPosition() + getSize() / T{2};
```
Bromeon, line 134: The idea of `T{2}` is that it works for all types without possible warnings (`int`, `unsigned int`, `float`, `double`, ...)?

Bromeon (approved):
Great addition, thank you! 😊

include/SFML/Graphics/Rect.inl
eXpl0it3r: Yes 🙂

ChrisThrasher: That's right, that's to ensure we're dividing by the same type as the `Vector2` being returned by `getSize()`. It's required to ensure compilation succeeds.

PR #2779, opened 2023-11-14
Ensure GNUInstallDirs cache vars are included before first used

### LINKED ISSUE #2778: Incorrect installation directory for pkg-config files.
### Prerequisite Checklist

### Describe your issue here

The pkg-config scripts (`tools/pkg-config/*.pc`) are always installed in the wrong location. This led to all Linux distros trying different tricks to override the wrong installation.

This happens because `SFML_PKGCONFIG_DIR` is not being expanded correctly.
In `Config.cmake`:
```cmake
set(SFML_PKGCONFIG_DIR "/${CMAKE_INSTALL_LIBDIR}/pkgconfig")
```
After the **first** `cmake` run, this variable expands to
```
//pkgconfig
```
After the **second** `cmake` run, it changes to:
```
/lib64/pkgconfig
```
It appears that `CMAKE_INSTALL_LIBDIR` is set AFTER it's used to set `SFML_PKGCONFIG_DIR`.
I think the problem is caused by the wrong order of inclusion in `CMakeLists.txt`:

```cmake
# include the configuration file
include(cmake/Config.cmake)

# we use the paths from the cmake GNUInstallDirs module as defaults
# you can override these if you like
# https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html
include(GNUInstallDirs)
```

If I change it to include `GNUInstallDirs` before `cmake/Config.cmake`, the value seems to be set correctly.

### Your Environment

- OS / distro / window manager: Fedora 39, Mageia 9, Ubuntu Mantic, Debian Bookworm, Arch Linux (latest stable releases of all of them)
- SFML version: 2.6.1
- Compiler / toolchain: GCC

### Steps to reproduce

1. Edit `cmake/Config.cmake`, appending these lines at the bottom:
```cmake
message(STATUS ":::: CMAKE_INSTALL_LIBDIR = ${CMAKE_INSTALL_LIBDIR}")
message(STATUS ":::: SFML_PKGCONFIG_DIR = ${SFML_PKGCONFIG_DIR}")
```
2. `mkdir build`
3. `cd build`
4. `cmake ..`
5. `cmake ..`

### Expected behavior

It should always print
```
-- :::: CMAKE_INSTALL_LIBDIR = lib64
-- :::: SFML_PKGCONFIG_DIR = /lib64/pkgconfig
```
each time `cmake` is run.

### Actual behavior

First `cmake` run prints:
```
-- :::: CMAKE_INSTALL_LIBDIR = 
-- :::: SFML_PKGCONFIG_DIR = //pkgconfig
```
The second run prints:
```
-- :::: CMAKE_INSTALL_LIBDIR = lib64
-- :::: SFML_PKGCONFIG_DIR = /lib64/pkgconfig
```

ChrisThrasher:
## Description

Closes #2778


ChrisThrasher:
@dkosmari I retargeted this PR to 2.6.x. That seems like a reasonable thing to do. Ensuring distros don't have to patch SFML is a goal of ours.

dkosmari (approved):

ChrisThrasher:
image

This bug would have been caught if we were using CMake's uninitialized variable warnings.

eXpl0it3r (approved):

Issue #2394, opened 2023-02-06
Ability to choose which SFML modules to use on Android


Zombieschannel:
## Subject of the issue

Currently, on Android, in the file SFMLActivity.cpp in SFML/Main, there is a function loadLibrary() which loads all SFML modules + openal. If it fails, exit(1) is called. Unfortunately, there is no way to remove loading these libraries without modifying SFML code itself. For example, if there is no audio in an app, sfml-audio, and openal aren't needed. Removing these in the final apk/aab, can reduce the file size quite a bit. If you don't use sfml-audio anywhere, removing the loadLibrary("sfml-audio"...); does not effect the app, it works anyway.

![image](https://user-images.githubusercontent.com/54750550/217066889-3b25af46-4c85-4266-8125-9703c271e0c1.png)

A quick idea of how to fix this:

Just remove the line exit(1);

- if those ".so" files are actually needed, and they are just missing, the app will crash anyway but you'll still be able to see that they are missing in the logs
- if it isn't needed, the app will still complain, but at least it won't get stuck in a reboot while loop

## Your environment

Android Studio 2022.1.1 on Windows 10
SFML 2.6.x branch
Used CMake, Ninja, Android NDK 25

## Steps to reproduce

Just don't use sfml-audio anywhere and don't include the sfml-audio and openal in the final apk/aab.

## Expected behavior

It should ignore the fact that sfml-audio is missing since it isn't needed anyway.

## Actual behavior

It gets stuck in a forever loop.

ChrisThrasher:
> if those ".so" files are actually needed, and they are just missing, the app will crash anyway but you'll still be able to see that they are missing in the logs

What logs do you get when it crashes this way? Is it still clear what went wrong?

Zombieschannel:
If I have that exit(1) line commented out:

So first of all, when I remove sfml-audio and openal from cmakelists in the project, as expected, it does not include these files in the final apk. 

- Now if I don't use anything from sfml-audio in my code, compilation is successful, running on the device works, you just get those warnings that it was unable to find libsfml-audio.so and libopenal.so.
- If I #include <SFML/Audio.hpp> and write some basic code that opens an audio file and plays it, compilation is not successful, and I get errors. 
[...]

PR #2783, opened 2023-11-15
sfml-activity is no longer needed on Android

texus:
## Description

sfml-activity was used to bootstrap Android applications: it manually loaded the .so libraries. This is however no longer needed.

By removing sfml-activity, all issues that currently existed with it will be fixed:
- Fixes https://github.com/SFML/SFML/issues/2394
- Closes https://github.com/SFML/SFML/pull/2461 

Impact on user code:
- `sfml.app.lib_name` should no longer occur in AndroidManifest.xml. Previously `android.app.lib_name` was set to sfml-activity which then loaded the app via `sfml.app.lib_name`. Now the app should be set to `android.app.lib_name` directly.
- SFML::Activity should no longer be passed to target_link_libraries
- You no longer need to link SFML modules that you don't use

This PR also includes several other minor changes (in separate commits):
- Most commits are about old variables which were no longer being used.
- There are also 2 commits related to static linking. While static linking isn't supported yet (when building for Android), those are two trivial changes that would prevent static linking from working (these commits are part of a branch that I have where static linking does work, but the other changes are more tricky/hacky and might require more discussion before merging).

## Tasks

-   [x] Tested on Android (both in emulator and on real device)

## How to test this PR?

Build SFML for android and install it into the NDK:
```
cmake -DCMAKE_SYSTEM_NAME=Android -DCMAKE_SYSTEM_VERSION=21 -DCMAKE_ANDROID_NDK=/path/to/ndk -DCMAKE_ANDROID_ARCH_ABI=x86_64 -DCMAKE_ANDROID_STL_TYPE=c++_shared -DCMAKE_BUILD_TYPE=Debug ..
make && make install
```

Build and run the SFML/examples/android project (requires a running emulator or connected device with the correct ABI):
```
./gradlew -P ARCH_ABI=x86_64 installDebug
```


ChrisThrasher (approved):
Screenshot 2023-11-15 at 10 47 46 AM

🤩

Thanks for the PR! It's awesome to see so much Android-specific stuff get replaced with more generic, modern alternatives.

ChrisThrasher:

Issue #1528, opened 2018-12-23
Font copy constructor not working on Android


acsbendi:
Thanks for raising your issue here! 🙂

Before you submit the issue however, we'd like you to consider the follow points.

* We like to use the issue tracker for confirmed issues.
* If you're stuck with SFML, please use [the forum](https://en.sfml-dev.org/forums/index.php#c3) to get help.

----

## Subject of the issue

`sf::Font`'s copy constructor does not work on Android, when the original object get destructed the copy constructed object won't be usable anymore. Of course its use is not irreplaceable at all, but I just would not expect this to happen and thus wanted to open an issue.

## Your environment

Windows 8.1 using Android Studio with ndk r15 and clang 5.0
I compiled SFML git master with GCC 4.9.
I used the following command for compilation: 
```cmake -stdlib=libc++ -DCMAKE_SYSTEM_NAME=Android -DCMAKE_ANDROID_NDK=D:/NVPACK/android-ndk-r15c -DCMAKE_ANDROID_ARCH_ABI=x86 -DCMAKE_ANDROID_STL_TYPE=c++_static -DCMAKE_BUILD_TYPE=Debug -G "MinGW Makefiles" ../..```

## Steps to reproduce

```cpp
#include <SFML/Graphics.hpp>

using std::string;
using sf::Font;


void LoadFontInto(Font& font){
    sf::Font local;
    local.loadFromFile("sansation.ttf");
    font = Font{local};
}

int main(int argc, char *argv[])
{
    sf::VideoMode screen(sf::VideoMode::getDesktopMode());

    sf::RenderWindow window(screen, "");
    window.setFramerateLimit(30);

    sf::View view = window.getDefaultView();

    sf::Color background = sf::Color::White;

    Font font{};
    LoadFontInto(font);

    sf::Text text("Tap anywhere to move the logo.", font , 64);
    text.setFillColor(sf::Color::Black);
    text.setPosition(10, 10);
    
    bool active = true;

    while (window.isOpen())
    {
        sf::Event event;

        while (active ? window.pollEvent(event) : window.waitEvent(event))
        {
            switch (event.type)
            {
                case sf::Event::Closed:
                    window.close();
                    break;
                case sf::Event::KeyPressed:
                    if (event.key.code == sf::Keyboard::Escape)
                        window.close();
                    break;
                case sf::Event::Resized:
                    view.setSize(event.size.width, event.size.height);
                    view.setCenter(event.size.width/2, event.size.height/2);
[...]

PR #2355, opened 2023-01-12
Added modernize-use-equals-default, modernize-use-equals-delete

gnawme:
Thanks a lot for making a contribution to SFML! 🙂

Before you create the pull request, we ask you to check the follow boxes. (For small changes not everything needs to ticked, but the more the better!)

* [x] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before?
* x] Does the code follow the SFML [Code Style Guide](https://www.sfml-dev.org/style.php)?
* [ ] Have you provided some example/test code for your changes?
* [ ] If you have additional steps which need to be performed list them as tasks!

----

## Description

Please describe your pull request.

This PR is related to the issue #

## Tasks

* [x] Tested on Linux
* [ ] Tested on Windows
* [ ] Tested on macOS
* [ ] Tested on iOS
* [ ] Tested on Android

## How to test this PR?

Describe how to best test these changes. Please provide a [minimal, complete and verifiable example](https://stackoverflow.com/help/mcve) if possible, you can use the follow template as a start:

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode({1280, 720}), "Minimal, complete and verifiable example");
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        for (sf::Event event; window.pollEvent(event);)
        {
            if (event.type == sf::Event::Closed)
                window.close();
        }

        window.clear();
        window.display();
    }
}
```

PR #2791, opened 2023-11-17
Fixed crash on Android when using copied font while original font is destroyed

texus:
## Description

On android, the Font class opens a stream to the .ttf file in loadFromFile. When copying the font, this stream isn't copied.

The copied font can be used as long as the original font (which contains the stream) still exists. Destroying the original font will destroy the stream and cause a crash when attempting to use the copied font.

This fixes https://github.com/SFML/SFML/issues/1528

Remarks about the code:
- I added code to the body of the copy constructor instead of using the initializer list because otherwise a comma would need to exist on the line above depending on whether or not you are building for android.
- As far as I can tell the copy constructor and copy assignment operator are no longer special and we could just define them as "= default". Since all constructor, assignment operators and destructor would become defaulted, they can probably all be removed completely.

## Tasks

-   [x] Tested on Android

## How to test this PR?

Build the android example with the following code in main.cpp, if you see the "Tap anywhere to move the logo" text at the top then the code is working. Without this PR the example will crash or it may run but without the text being rendered.
```c++
#include <SFML/Graphics.hpp>

using std::string;
using sf::Font;

void LoadFontInto(Font& font){
    sf::Font local;
    local.loadFromFile("tuffy.ttf");
    font = Font{local};
}

{

    window.setFramerateLimit(30);

    Font font;
    LoadFontInto(font);

    sf::Text text(font, "Tap anywhere to move the logo.", 64);
    text.setPosition({10, 10});

    bool active = true;

    while (window.isOpen())
    {
        sf::Event event;

        {
            switch (event.type)
            {
                case sf::Event::Closed:
                    window.close();
                    break;
                case sf::Event::KeyPressed:
                        window.close();
                    break;
                case sf::Event::Resized:
                    view.setSize({static_cast<float>(event.size.width), static_cast<float>(event.size.height)});
                    view.setCenter({event.size.width/2.f, event.size.height/2.f});
                    window.setView(view);
                    break;
                case sf::Event::LostFocus:
                    break;
                case sf::Event::GainedFocus:
                    break;
                case sf::Event::MouseLeft:
                    active = false;
                    break;
                case sf::Event::MouseEntered:
                    active = true;
                    break;
                default:
                    break;
            }
        }

        if (active)
        {
            window.clear(background);
            window.draw(text);
            window.display();
        }
        else {
        }
    }
}
```

### DIFF
--- a/include/SFML/Graphics/Font.hpp
+++ b/include/SFML/Graphics/Font.hpp
@@ -427,7 +427,7 @@ private:
     mutable PageTable            m_pages;          //!< Table containing the glyphs pages by character size
     mutable std::vector<std::uint8_t> m_pixelBuffer; //!< Pixel buffer holding a glyph's pixels before being written to the texture
 #ifdef SFML_SYSTEM_ANDROID
-    std::unique_ptr<priv::ResourceStream> m_stream; //!< Asset file streamer (if loaded from file)
+    std::shared_ptr<priv::ResourceStream> m_stream; //!< Asset file streamer (if loaded from file)
 #endif
 };
 
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -136,6 +136,9 @@ m_info(copy.m_info),
 m_pages(copy.m_pages),
 m_pixelBuffer(copy.m_pixelBuffer)
 {
+#ifdef SFML_SYSTEM_ANDROID
+    m_stream = copy.m_stream;
+#endif
 }
 
 
@@ -204,7 +207,7 @@ bool Font::loadFromFile(const std::filesystem::path& filename)
 
 #else
 
-    m_stream = std::make_unique<priv::ResourceStream>(filename);
+    m_stream = std::make_shared<priv::ResourceStream>(filename);
     return loadFromStream(*m_stream);
 
 #endif


ChrisThrasher:

include/SFML/Graphics/Font.hpp
```diff
@@ -427,7 +427,7 @@ class SFML_GRAPHICS_API Font
     mutable PageTable            m_pages;          //!< Table containing the glyphs pages by character size
     mutable std::vector<std::uint8_t> m_pixelBuffer; //!< Pixel buffer holding a glyph's pixels before being written to the texture
 #ifdef SFML_SYSTEM_ANDROID
-    std::unique_ptr<priv::ResourceStream> m_stream; //!< Asset file streamer (if loaded from file)
+    std::shared_ptr<priv::ResourceStream> m_stream; //!< Asset file streamer (if loaded from file)
```
ChrisThrasher, line 430: My first impression was to be wary of this introduction of shared ownership but a few lines above this we already have a shared pointer in `m_fontHandles` so it seems like `sf::Font` already sets the precedent that some data has shared ownership.

src/SFML/Graphics/Font.cpp
```diff
@@ -136,6 +136,9 @@ m_info(copy.m_info),
 m_pages(copy.m_pages),
 m_pixelBuffer(copy.m_pixelBuffer)
 {
+#ifdef SFML_SYSTEM_ANDROID
+    m_stream = copy.m_stream;
+#endif
```
kimci86, line 141: I guess we can just implement copy with `= default` now.

ChrisThrasher: Yeah that does seem to be the case. It makes me wonder why it wasn't always defaulted to begin with. Is there some intent because this hand-written copy constructor that we're not recognizing?

kimci86: The reason was it was actually not copyable when a `std::unique_ptr` was used. See https://github.com/SFML/SFML/pull/2355#discussion_r1067644147

ChrisThrasher:
https://github.com/SFML/SFML/commit/6e18bb4c8986d31b7f6fd4e9a243025146ddd394

I'm not yet 100% if this is correct but if so, we can delete a _lot_ of code.

texus:
Seems correct to me. In SFML 2 the copy and cleanup code in sf::Font was more complex, but that has all been replaced by the `FontHandles` class. So now the Font class itself no longer needs special handling.

Just to double check I also build the `default_font_operations` branch on Android and it still displays the text correctly after copying the font. So if you merge the `default_font_operations` then this PR and the associated issue can be closed.

ChrisThrasher:
See https://github.com/SFML/SFML/pull/2792

Thanks for your PR!

PR #2792, opened 2023-11-17
Default all `sf::Font` special member functions

ChrisThrasher:
## Description

Closes #2791

There were apparently historical reasons why `sf::Font` had custom copy operations but we don't need that anymore. Once we switch to using a `shared_ptr` to own the Android `ResourceStream` then we can allow to compiler to generate all special member functions.

### DIFF
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -124,34 +124,6 @@ struct Font::FontHandles
 
 
 ////////////////////////////////////////////////////////////
-Font::Font() = default;
-
-
-////////////////////////////////////////////////////////////
-// NOLINTNEXTLINE(modernize-use-equals-default)
-Font::Font(const Font& copy) :
-m_fontHandles(copy.m_fontHandles),
-m_isSmooth(copy.m_isSmooth),
-m_info(copy.m_info),
-m_pages(copy.m_pages),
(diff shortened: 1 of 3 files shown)


binary1248 (approved):

src/SFML/Graphics/Font.cpp
```diff
@@ -502,25 +474,6 @@ bool Font::isSmooth() const
...
-    std::swap(m_info, temp.m_info);
-    std::swap(m_pages, temp.m_pages);
-    std::swap(m_pixelBuffer, temp.m_pixelBuffer);
-
-#ifdef SFML_SYSTEM_ANDROID
-    std::swap(m_stream, temp.m_stream);
-#endif
-
-    return *this;
-}
```
ChrisThrasher, line 521: I'm not sure why this was originally implemented as a swap but this PR would change it to a normal copy assignment. Is that prone to break anything?

texus: The code was just using the Copy-and-Swap Idiom.
Instead of duplicating all the exceptional code in both the copy constructor and the assignment operator, the assignment operator calls the copy constructor (by creating the temp object) and then swaps the result of the copy constructor into `this`.

So this was just to reuse the code in the copy constructor, it is functionally equivalent to a normal copy assignment.

PR #2798, opened 2023-11-20
Implicitly define default constructors

ChrisThrasher:
## Description

Yay for deleting code!


include/SFML/Graphics/Shape.hpp
```diff
@@ -256,12 +256,6 @@ class SFML_GRAPHICS_API Shape : public Drawable, public Transformable
     FloatRect getGlobalBounds() const;
 
 protected:
-    ////////////////////////////////////////////////////////////
-    /// \brief Default constructor
-    ///
-    ////////////////////////////////////////////////////////////
-    Shape();
```
ChrisThrasher, line 263: There is no need to hide this constructor. This class already contains some pure virtual functions making it not constructible by default. The `sf::Shape` tests already assert that this type is not constructible.

src/SFML/Audio/SoundFileReaderOgg.cpp
```diff
@@ -87,13 +87,6 @@ bool SoundFileReaderOgg::check(InputStream& stream)
 }
 
 
-////////////////////////////////////////////////////////////
-SoundFileReaderOgg::SoundFileReaderOgg()
-{
-    m_vorbis.datasource = nullptr;
```
ChrisThrasher, line 93: There is no need to set this to `nullptr` because we already initialize `m_vorbis` with `{}` which ensures this particular data member is set to `nullptr`.

binary1248 (approved):

PR #2805, opened 2023-11-27
Add move semantics to `sf::Context`

ChrisThrasher:
## Description

As always implementing custom move operations is tricky but I think the tests I've written should do a good job proving the correctness of this PR.


test/Window/Context.test.cpp
```diff
@@ -21,28 +21,24 @@ TEST_CASE("[Window] sf::Context", runDisplayTests())
...
-
         CHECK(context.setActive(true));
 
         using glGetStringFuncType = const char*(GLAPI*)(unsigned int);
-
-        auto glGetStringFunc = reinterpret_cast<glGetStringFuncType>(sf::Context::getFunction("glGetString"));
-
-        REQUIRE(glGetStringFunc);
+        auto glGetStringFunc      = reinterpret_cast<glGetStringFuncType>(sf::Context::getFunction("glGetString"));
+        REQUIRE(!!glGetStringFunc);
```
ChrisThrasher, line 41: This converts the function pointer to a boolean which fixes an issue that leads to UBSan failing under certain conditions. I've discussed this with the Catch2 maintainer and it appears to be an issue with my Clang 17 installation. 

There are actually other places in our tests that trigger this same error. I'm unsure if I bother fixing this here if I don't also fix it elsewhere.

ChrisThrasher:
@kimci86 What do you think of this?

kimci86:
@ChrisThrasher Code looks mostly ok assuming moving between threads is not supported.

However, I am just not sure about the behavior of `getActiveContext` and underlying `currentContext`.
Consider the following code:
```c++
sf::Context c1;
CHECK(sf::Context::getActiveContext() == &c1);

sf::Context c2 = std::move(c1);
CHECK(sf::Context::getActiveContext() == nullptr); // <- This doesn't look right
```

src/SFML/Window/Context.cpp
```diff
@@ -120,12 +110,22 @@ GlFunctionPointer Context::getFunction(const char* name)
...
 
+
+////////////////////////////////////////////////////////////
+void Context::ContextDeleter::operator()(priv::GlContext* context) const
+{
+    if (context->setActive(false))
+        ContextImpl::currentContext = nullptr;
+    else
+        err() << "Failed to set context as inactive during destruction" << std::endl;
+}
```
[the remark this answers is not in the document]
ChrisThrasher: 🤦🏻‍♂️

ChrisThrasher:
> @ChrisThrasher Code looks mostly ok assuming moving between threads is not supported.
> 
> However, I am just not sure about the behavior of `getActiveContext` and underlying `currentContext`. Consider the following code:
> 
> ```c++
> sf::Context c1;
> CHECK(sf::Context::getActiveContext() == &c1);
> 
> sf::Context c2 = std::move(c1);
> CHECK(sf::Context::getActiveContext() == nullptr); // <- This doesn't look right
> ```

That's a good point. I don't see how this `unique_ptr` deleter-based solution could ensure the newly move-constructed `sf::Context` sets the current active context. I might have to roll this back in favor of a more conventional solution.

The fact that I keep forgetting to write `delete ptr;` and the fact that we have to write `delete` at all has me calling into question whether this is the right solution. I feel like I'm abusing what deleters are meant for.

kimci86:
Another way to question this situation is whether tracking currentContext globally (for each thread) is of any use. It is well known that global states are bad and getCurrentContext is not used by SFML itself. Maybe we could consider removing this method.

ChrisThrasher:
> Another way to question this situation is whether tracking currentContext globally (for each thread) is of any use. It is well known that global states are bad and getCurrentContext is not used by SFML itself. Maybe we could consider removing this method.

You might be onto something there. I'll let @binary1248 weight in on the value on maintaining this API.

binary1248:
We have to be careful not to assume too much. The difference between `getActiveContext` and `getActiveContextId` is that `getActiveContext` either returns a pointer to an actual constructed `sf::Context` object or `nullptr` whereas `getActiveContextId` can also return IDs to internal contexts that SFML constructs for housekeeping purposes.

If someone tried to make use of a weird "context pool" in their own code and kept track of which context was active using only `getActiveContextId` they would realize that it sometimes returns IDs of contexts they didn't even create. AFAIK SFML doesn't construct many if any `sf::Context` objects internally (we make use of `TransientContextLock` instead), so it is not far fetched to be able to assume that all contexts returned by `getActiveContext` were ones the user created themselves.

I think everybody who has been writing C++ code for a while can agree that any global state is something that should be avoided. However, we are working with an almost 30-year-old system API at this point. The damage was done long before SFML even existed. All this ugliness is just our feeble attempt at hiding this ugliness from the user under a nice C++ API. If here and there OpenGL manages to peek its head out of the library then it is something we will have to live with until we come up with a new rendering backend.

kimci86:
@binary1248 It is true that we cannot avoid global states with OpenGL, but it is not clear to me what is the purpose of `getActiveContext` in particular and the corresponding `currentContext` pointer. This method is not used in SFML nor its examples. What are the arguments for not removing this method?

binary1248:
I would have less of an issue getting rid of `getActiveContext` if `sf::Context` got a `isActive` and/or `getId` method. Checking for whether a `sf::Context` you create is currently active shouldn't require you to activate them right after construction just to save its ID. Right now the `sf::Context` API seems half-baked.

You are right that many of its currently public functions were introduced to be used by SFML at some point in time and end up being very technical in their usage, but we mustn't forget that our goal should also be to improve usability of the public API in user code as well.

While we have always actively recommended against using multiple OpenGL contexts, there are some legitimate very advanced use cases where using multiple contexts can improve performance, and having a semi-usable `sf::Context` API makes already complex use cases more complicated to implement than they need to be.

src/SFML/Window/Context.cpp
```diff
@@ -49,19 +49,40 @@ namespace sf
...
         err() << "Failed to set context as inactive during destruction" << std::endl;
-    }
+}
+
+
+////////////////////////////////////////////////////////////
+Context::Context(Context&& other) noexcept : m_context(std::move(other.m_context))
+{
+    if (!setActive(true))
+        err() << "Failed to set context as active during construction" << std::endl;
```
kimci86, line 68: Are you sure this is wanted? When constructing from a context that is not active, wouldn't it make more sense to get a context that is also not active? What about this:
```suggestion
    if (ContextImpl::currentContext == &other)
        ContextImpl::currentContext = this;
```

Remark also applies to move assignment operator.

---

I tried it locally. It became apparent to me that the logic about `currentContext` in `setActive` looks wrong: `setActive(false)` sets `currentContext` to `nullptr` even when the context being deactivated is not the active one.
This implementation of `setActive` looks more correct to me:
```c++
bool Context::setActive(bool active)
{
    const bool result = m_context->setActive(active);

    if (result)
    {
        if (active)
            ContextImpl::currentContext = this;
        else if (ContextImpl::currentContext == this)
            ContextImpl::currentContext = nullptr;
    }

    return result;
}
```

src/SFML/Window/Context.cpp
```diff
@@ -49,19 +49,40 @@ namespace sf
...
+{
+    if (this == &other)
+        return *this;
+
+    if (m_context && !setActive(false))
+        err() << "Failed to set context as inactive during move assignment" << std::endl;
+
+    m_context = std::move(other.m_context);
+    if (m_context && !setActive(true))
+        err() << "Failed to set context as active during construction" << std::endl;
```
kimci86, line 83: Copy-paste detected :nerd_face:
```suggestion
        err() << "Failed to set context as active during move assignment" << std::endl;
```

test/Window/Context.test.cpp
```diff
@@ -33,6 +33,28 @@ TEST_CASE("[Window] sf::Context", runDisplayTests())
...
 
+    SECTION("Move semantics")
+    {
+        SECTION("Construction")
+        {
+            sf::Context       movedContext;
+            const sf::Context context(std::move(movedContext));
+            CHECK(context.getSettings().majorVersion > 0);
+            CHECK(context.getActiveContext() == &context);
+            CHECK(context.getActiveContextId() != 0);
```
kimci86, line 44: Those are static methods.
```suggestion
            CHECK(sf::Context::getActiveContext() == &context);
            CHECK(sf::Context::getActiveContextId() != 0);
```
Also applies to other test cases.

kimci86 (changes requested):

src/SFML/Window/Context.cpp
ChrisThrasher: 😰

src/SFML/Window/Context.cpp
ChrisThrasher: > It became apparent to me that the logic about currentContext in setActive looks wrong: setActive(false) sets currentContext to nullptr even when the context being deactivated is not the active one.

I can reproduce this with a test. Create a 2nd context then call `.setActive(false)` on the first context and then `sf::Context::getActiveContext` returns false. Good catch.

I think I want to make a separate PR to address this.

ChrisThrasher: See #2816

ChrisThrasher:
I rebased on #2816 and added more tests that should fail thus demonstrating some issues Kimci brought up.

src/SFML/Window/Context.cpp
```diff
@@ -49,19 +49,37 @@ namespace sf
...
         err() << "Failed to set context as inactive during destruction" << std::endl;
-    }
+}
+
+
+////////////////////////////////////////////////////////////
+Context::Context(Context&& context) noexcept : m_context(std::move(context.m_context))
+{
+    if (!setActive(&context == ContextImpl::currentContext))
+        err() << "Failed to set context as active during move construction" << std::endl;
```
kimci86, line 68: Because `setActive` requires `m_context` not to be null, the following code would crash:
```c++
sf::Context c1;
sf::Context c2 = std::move(c1);
sf::Context c3 = std::move(c1);
```
Maybe we can ignore this problem because it is generally bad to use a variable after move, but on the other hand it is simple to fix with the following code:

```suggestion
    if (&context == ContextImpl::currentContext)
        ContextImpl::currentContext = this;
```

Same applies to move assignment operator.

kimci86 (commented):
Besides one last comment, it looks good!

src/SFML/Window/Context.cpp
ChrisThrasher: I did actually think of that but my impression is that moving out of `c1` twice is something we can assume users aren't allowed to do and thus we don't have to guard against it.

ChrisThrasher: Perhaps we `assert` against moving out of an empty context so that this is a hard failure and we prevent UB in Debug builds?

ChrisThrasher: I added asserts to protect against this (mis)use of contexts. If someone later shows a valid use case for double-moving out of a context then we can remove the asserts and instead add an `if (m_context)` check to define this behavior in release builds.

ChrisThrasher: Apparently we are required to let users move out of the same object many times so we’re forced to add that pointer check back.

kimci86 (approved):

PR #2808, opened 2023-11-29
Avoid taking lock when not necessary

ChrisThrasher:
## Description

We can avoid taking this lock when using OpenGL ES.

### DIFF
--- a/src/SFML/Graphics/RenderTextureImplFBO.cpp
+++ b/src/SFML/Graphics/RenderTextureImplFBO.cpp
@@ -116,17 +116,18 @@ bool RenderTextureImplFBO::isAvailable()
 ////////////////////////////////////////////////////////////
 unsigned int RenderTextureImplFBO::getMaximumAntialiasingLevel()
 {
+#ifdef SFML_OPENGL_ES
+
+    return 0;
+
+#else
+
     const TransientContextLock lock;
-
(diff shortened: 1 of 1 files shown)


src/SFML/Graphics/RenderTextureImplFBO.cpp
```diff
@@ -116,17 +116,16 @@ bool RenderTextureImplFBO::isAvailable()
...
 
+    const TransientContextLock lock;
+    GLint                      samples = 0;
     glCheck(glGetIntegerv(GLEXT_GL_MAX_SAMPLES, &samples));
+    return static_cast<unsigned int>(samples);
 
 #endif
 
-    return static_cast<unsigned int>(samples);
+    return 0;
```
eXpl0it3r, line 128: Surprised that Clang isn't complaining about unreachable code.

I wonder if it weren't more readable, by having a `#ifdef SFML_OPENGL_ES` section at the top, that just returns 0.

binary1248 (approved):

src/SFML/Graphics/RenderTextureImplFBO.cpp
ChrisThrasher: I'd prefer that too

eXpl0it3r (approved):

PR #2810, opened 2023-11-29
Remove some references to SFML 2

ChrisThrasher:
## Description

Mentioning the version number isn't always necessary so I removed some instances of the version number to prevent us redoing this work for new major releases going forward.

### DIFF
--- a/tools/xcode/templates/readme.txt
+++ b/tools/xcode/templates/readme.txt
@@ -1,7 +1,7 @@
-XCODE TEMPLATES FOR SFML 2
-==========================
+XCODE TEMPLATES FOR SFML
+========================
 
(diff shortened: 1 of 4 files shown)


tools/xcode/templates/readme.txt
```diff
@@ -82,12 +82,6 @@ FAQ
 
 
 
-  * I'm still using SFML 1.6. Can I use these templates anyway?
-
-    No, these templates were made for SFML 2 only.
```
ChrisThrasher, line 87: It's safe to assume no users are still using 1.6 these days. Let's just not bother mentioning it.

binary1248 (approved):

Issue #2462, opened 2023-03-22
SFML crash on Galaxy S22


Zombieschannel:
## Subject of the issue

So this gets quite complicated but here is what I got.
While testing I found that for some unknown reason the SFML 2 app crashes on the Samsung galaxy S22. Since I've previously tested the same code on a dozen different devices from different manufacturers, it was quite weird. Well, it seems to crash in the window.draw() function no matter what you ask it to draw. I've tested another app that just clears the screen to a specific color and doesn't draw anything and it works without any issues.
I've captured the call stack of the crash:

![image](https://user-images.githubusercontent.com/54750550/227060152-9b139ce2-4e73-4ec4-8fef-ddc60897f447.png)

as well as the specific line that it crashed on:

![image](https://user-images.githubusercontent.com/54750550/227060213-a4eb5b7a-ed7a-49eb-959e-f8853ad7bcc0.png)

## My conclusion:
It seems like the galaxy S22 series has a completely new GPU compared to literally any other smartphone.
The Xclipse 920 that it has is the only Xclipse GPU ever. The now new s23 series no longer has this Xclipse GPU.
So I guess it could be a graphics driver issue. Or something similar. I haven't tried just removing this line that the program crashes on as I'm unsure what it does exactly.

## Your environment

Android Studio 2022.1.1 on Windows 10
SFML 2.6.x branch
Used CMake, Ninja, Android NDK 25

Zombieschannel:
After analyzing this issue further, it seems like the S22 indeed does not have this extension `glBlendEquation` (or at least not when using GLES 1.0) resulting in a crash. Simply changing the else { ... } to else if (GLEXT_glBlendEquation) { ... } should be a fix. I'll make a pull request soon.

Zombieschannel:
Just an interesting observation, after "upgrading" SFML to use GLES 2, all of the issues are gone.

vittorioromeo:
> Just an interesting observation, after "upgrading" SFML to use GLES 2, all of the issues are gone.

Are you planning to PR your https://github.com/Zombieschannel/SFML/tree/emcc branch into SFML upstream eventually? It would be a fantastic addition to SFML 3.x, and I've wanted Emscripten support for a long time!

It would actually make SFML usable for web browser games, greatly increasing its usage.

Zombieschannel:
> > Just an interesting observation, after "upgrading" SFML to use GLES 2, all of the issues are gone.
> 
[...]

PR #2814, opened 2023-12-05
Fix crash when glBlendEquation is missing

Zombieschannel:
## Description

Fix to this issue: https://github.com/SFML/SFML/issues/2462

image

Basically, I noticed that the Galaxy S22 Exynos crashes when running any SFML apps, and it seems like it does not have the glBlendEquation extension, SFML does not check whether it has it or not, it just tries to use it, resulting in a crash. Adding an else if statement fixes this.

## Tasks

-   [ ] Tested on Linux
-   [ ] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [x] Tested on Android

### DIFF
--- a/src/SFML/Graphics/RenderTarget.cpp
+++ b/src/SFML/Graphics/RenderTarget.cpp
@@ -638,7 +638,7 @@ void RenderTarget::applyBlendMode(const BlendMode& mode)
                 equationToGlConstant(mode.colorEquation),
                 equationToGlConstant(mode.alphaEquation)));
         }
-        else
+        else if (GLEXT_glBlendEquation)
         {
             glCheck(GLEXT_glBlendEquation(equationToGlConstant(mode.colorEquation)));
         }


ChrisThrasher:
Do you know why this seemingly modern phone doesn't support this feature?

Is it fine to simply do nothing when this GL feature is not present? Do we need a new `else` block that does some fallback behavior for devices without this feature?

Zombieschannel:
> Do you know why this seemingly modern phone doesn't support this feature?

I have no idea, but I mentioned my theory in https://github.com/SFML/SFML/issues/2462 under `My conclusion:`

> Is it fine to simply do nothing when this GL feature is not present?

Since it is an extension, I don't think it will have any impact on rendering (other than the fact that changing the blend equation won't do anything). I guess could add a warning saying that it is missing from the device.

> Do we need a new else block that does some fallback behavior for devices without this feature?

I don't think there is anything special that can be done if this function is missing. It is included in GLES 2.0+.

ChrisThrasher:

Zombieschannel:
So here are the results: https://opengles.gpuinfo.org/displayreport.php?id=6660
And yes, EXT_blend_minmax seems to be available.
Also this function is mentioned: 
image

Zombieschannel:
Interestingly, [EXT_blend_minmax](https://registry.khronos.org/OpenGL/extensions/EXT/EXT_blend_minmax.txt) mentions this:

> Dependencies on OpenGL ES
    If the GL is OpenGL ES, only the new MIN_EXT and MAX_EXT blend equations
    are introduced by this extension.  BlendEquationOES, FUNC_ADD_OES, and
    BLEND_EQUATION_OES are introduced by the OES_blend_subtract extension,
    which is _required for this extension to operate._  Alternatively,
    OpenGL ES 2.0 is required, which introduces BlendEquation, FUNC_ADD, and
    BLEND_EQUATION without the suffixes.

OES_blend_subtract doesn't seem to be supported, could this be the issue?

binary1248:
In the gpuinfo report your provided, it seems like all < OpenGL ES 3.0 extensions were removed since their functionality was merged into the core specification. Can you check what version the context is that SFML creates on your system?

Zombieschannel:
I'm not sure how could I check that. Since if I use 2.6, I'll get that created version 0.0 error, and if I use SFML 3, I'll get that "defaulting to 1.1" as mentioned here: https://github.com/SFML/SFML/pull/2438

binary1248:
Based on your gpuinfo, the `GL_VERSION` string should be `OpenGL ES 3.2 ANGLE git hash: 50101db6df97`, so I really don't understand how the version query in SFML can fail as you described unless context creation itself fails in some way and prevents us from getting any function pointers. Can you try creating a context and getting its version with the current master revision of SFML?

Zombieschannel:
So when using SFML from master, window.getSettings().majorVersion and minorVersion return 0 0 for some reason. I thought it was fixed a few months ago. Anyways, using glGetString() with GL_VERSION GL_RENDERER GL_VENDOR GL_EXTENSIONS I got the data I provided in the txt. Hopefully these are now the GLES 1 extensions.

[glesdata.txt](https://github.com/SFML/SFML/files/13589518/glesdata.txt)

PR #2818, opened 2023-12-10
Add `sf::State` for specifying fullscreen or floating windows

ChrisThrasher:
## Description

This is the first step towards implementing https://github.com/SFML/SFML/wiki/SD%3A-States-and-Styles

It converts `Fullscreen` from a `Style` to a `State`. While I was touching this code I fixed an inconsistency where `sf::Style`s were sometimes passed as a `std::uint32_t` and sometimes an `unsigned long`.

### DIFF
--- a/examples/opengl/OpenGL.cpp
+++ b/examples/opengl/OpenGL.cpp
@@ -47,7 +47,11 @@ int main()
         contextSettings.sRgbCapable = sRgb;
 
         // Create the main window
-        sf::RenderWindow window(sf::VideoMode({800, 600}), "SFML graphics with OpenGL", sf::Style::Default, contextSettings);
+        sf::RenderWindow window(sf::VideoMode({800, 600}),
+                                "SFML graphics with OpenGL",
+                                sf::Style::Default,
+                                sf::State::Windowed,
+                                contextSettings);
         window.setVerticalSyncEnabled(true);
         window.setMinimumSize(sf::Vector2u(400, 300));
         window.setMaximumSize(sf::Vector2u(1200, 900));
--- a/examples/window/Window.cpp
+++ b/examples/window/Window.cpp
@@ -30,7 +30,7 @@ int main()
     contextSettings.depthBits = 24;
 
     // Create the main window
-    sf::Window window(sf::VideoMode({640, 480}), "SFML window with OpenGL", sf::Style::Default, contextSettings);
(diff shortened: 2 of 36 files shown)


ChrisThrasher:
Exploiter and I talked about whether the order of `style` and `state` should be reversed. I put `state` after `style` for no particular reason. I'm not sure if one parameter order makes more sense than the other.

ChrisThrasher:

MarioLiebisch:
> Exploiter and I talked about whether the order of `style` and `state` should be reversed. I put `state` after `style` for no particular reason. I'm not sure if one parameter order makes more sense than the other.

Since it's more likely you want to adjust `style` than `state` by default, I'd go with `style` first.

eXpl0it3r:
While I didn't fully realize or verbalize in the past discussion, I think the main issue is, that you currently *have to* set a style for fullscreen window, even though the fullscreen window won't have a style.  
By reversing the order, you could switch to fullscreen without providing a style (if you don't need a context).

Currently asking myself if we just need another overload `(VideoMode, Title, State = default, ContextSettings = default)`, which has the style set to the default style.

Bromeon:
Maybe it goes a bit into off-topic (we can gladly discuss it somewhere else), but did we ever consider a fluent builder API for these lengthy constructors?

Instead of
```cpp
sf::RenderWindow window(
    sf::VideoMode({800, 600}),
    "SFML graphics with OpenGL",
    sf::Style::Default,
    sf::State::Windowed,
    contextSettings);
```
we'd have

```cpp
auto window = sf::RenderWindow::builder()
    .videoMode({800, 600})
    .title("SFML graphics with OpenGL")
    .state(sf::State::Windowed),
    .context(contextSettings)
    .build();
// no .style() needed if it's the default
```

It may be a bit unidiomatic for SFML, but it would solve the "anything can be default" problem.

ChrisThrasher:

ChrisThrasher:
I implemented addition constructor overloads that let users specify a state without specifying a style. While I was looking at the code I found some issues with the docs that needed correcting and added some missing tests so the PR is overall in a better state now.

MarioLiebisch (approved):

ChrisThrasher:

ChrisThrasher:
Added missing constructor overload to `sf::RenderWindow` so it matches the constructors that `sf::Window` and `sf::WindowBase` have.

eXpl0it3r (approved):

PR #2821, opened 2023-12-10
Issue CMake Warning when using UCRT MinGW

eXpl0it3r:
## Description

[WinLibs](https://winlibs.com/) has popularized or even introduced MinGW versions using the UCRT instead of the previously MSVCRT. Unfortunately, as our MinGW dependencies are built for MSVCRT, the linking will fail.  
Maybe eventually we'll provide compatible binaries or always build from source, but until then, we'll output a warning.  

We only issue an error if the user tries to use the built-in libraries. If they provide their own, we don't issue anything.

## Tasks

-   [x] Tested on Windows

## How to test this PR?

Build SFML with MinGW using the UCRT.

Expected output:
```
CMake Error at CMakeLists.txt:199 (message):
  The pre-compiled SFML dependencies for MinGW are not compatible with the
  UCRT.  Either switch to the MSVCRT or build the dependencies yourself.


-- Configuring incomplete, errors occurred!
```

### DIFF
--- a/cmake/Config.cmake
+++ b/cmake/Config.cmake
@@ -8,6 +8,7 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
     # detect the architecture (note: this test won't work for cross-compilation)
     include(CheckTypeSize)
     check_type_size(void* SIZEOF_VOID_PTR)
+
     if(${SIZEOF_VOID_PTR} STREQUAL "4")
         set(ARCH_32BITS 1)
     elseif(${SIZEOF_VOID_PTR} STREQUAL "8")
@@ -18,25 +19,31 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
     endif()
 elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
     set(SFML_OS_UNIX 1)
+
     if(ANDROID)
         set(SFML_OS_ANDROID 1)
+
         # use the OpenGL ES implementation on Android
         set(OPENGL_ES 1)
     else()
         set(SFML_OS_LINUX 1)
+
         # don't use the OpenGL ES implementation on Linux
         set(OPENGL_ES 0)
     endif()
 elseif(CMAKE_SYSTEM_NAME MATCHES "^k?FreeBSD$")
     set(SFML_OS_FREEBSD 1)
+
     # don't use the OpenGL ES implementation on FreeBSD
     set(OPENGL_ES 0)
 elseif(CMAKE_SYSTEM_NAME MATCHES "^OpenBSD$")
     set(SFML_OS_OPENBSD 1)
+
     # don't use the OpenGL ES implementation on OpenBSD
     set(OPENGL_ES 0)
 elseif(CMAKE_SYSTEM_NAME MATCHES "^NetBSD$")
     set(SFML_OS_NETBSD 1)
+
     # don't use the OpenGL ES implementation on NetBSD
     set(OPENGL_ES 0)
 elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
@@ -54,6 +61,7 @@ elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
         # detect OS X version. (use '/usr/bin/sw_vers -productVersion' to extract V from '10.V.x'.)
         EXEC_PROGRAM(/usr/bin/sw_vers ARGS -productVersion OUTPUT_VARIABLE MACOSX_VERSION_RAW)
         STRING(REGEX REPLACE "10\\.([0-9]+).*" "\\1" MACOSX_VERSION "${MACOSX_VERSION_RAW}")
+
         if(${MACOSX_VERSION} LESS 7)
             message(FATAL_ERROR "Unsupported version of OS X: ${MACOSX_VERSION_RAW}")
             return()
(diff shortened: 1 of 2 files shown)


binary1248 (approved):

cmake/Config.cmake
```diff
@@ -122,6 +131,10 @@ elseif(CMAKE_COMPILER_IS_GNUCXX)
     if(GCC_MACHINE MATCHES ".*w64.*")
         set(SFML_COMPILER_GCC_W64 1)
     endif()
+
+    if(GCC_COMPILER_VERSION MATCHES "ucrt")
+        message(WARNING "The pre-compiled SFML dependencies for Windows are not compatible with the UCRT. Either switch to the MSVCRT or build the dependencies yourself.")
```
ChrisThrasher, line 136: Do we want to take this a step further and emit a fatal error when `SFML_USE_SYSTEM_DEPS` is false and then emit no warnings when it's true? That way those trying to link to our precompiled deps get an error they cannot miss while those using system deps are not annoyed with any warning?

ChrisThrasher:
```cmake
if(GCC_COMPILER_VERSION MATCHES "ucrt" AND NOT SFML_USE_SYSTEM_DEPS)
    message(FATAL_ERROR "The pre-compiled SFML dependencies for Windows are not compatible with the UCRT. Either switch to the MSVCRT or build the dependencies yourself.")
endif()
```

This is what I had in mind.

I also think the error message ought to mention MinGW as to not be confused with the binaries that MSVC or Clang use.

ChrisThrasher (approved):

cmake/Config.cmake
eXpl0it3r: As we don't really have access to `SFML_USE_SYSTEM_DEPS` yet in the Config.cmake, I moved it one level higher after the `SFML_USE_SYSTEM_DEPS` check. That way it's also kind of grouped.

PrashanthKumar0:
Let me know if this is problematic before I file it as issue.

This merge actually doesnt fully resolve ( #3054  )

Here's my compiler Information
```
clang version 18.1.0rc (https://github.com/llvm/llvm-project.git 6c90f8dd546334b01c9a86387950e8c72d459f1e)
Target: x86_64-w64-windows-gnu
Thread model: posix
InstalledDir: D:/Apps/llvm-mingw-20240221-ucrt-x86_64/bin
```

and i still following linker error instead of warning at cmake time. 
the reasong being libfreetype.a is not compatible with clang ucrt 

These are the linker errors comming from libfreetype
```
ld.lld: error: undefined symbol: _setjmp
>>> referenced by libfreetype.a(sfnt.c.obj):(tt_face_build_cmaps>>> referenced by libfreetype.a(smooth.c.obj):(gray_convert_glyph_inner)
clang-18: error: linker command failed with exit code 1 (use -v to see invocation)
```

## Request

Can I file an issue and work on resolving this?

freetype is a small library so maybe can we distribute with source so it would be more cross platform and work across all compilers.

eXpl0it3r:
Ah you're using Clang 🤔 

Not sure it's worth opening an issue, we always link it to here or the other issue.

How do you plan to resolve it? Do you know of a way to detect UCRT for clang?

Alternatively, I guess it's just time to bite the bullet and build UCRT versions for the extlibs.

> freetype is a small library so maybe can we distribute with source so it would be more cross platform and work across all compilers.

We're working on a way to have the dependencies built from source, so we no longer have to ship binaries, but it's not trivial if you want to consider all the config configurations.

https://github.com/SFML/SFML/compare/master...feature/build_dependencies

PrashanthKumar0:
i have a Good news !
 
I have successfully modified cmake files so freetype is build from source.

Hopefully it would work on all compilers now ( tested on clang and wsl's gcc). Need one suggestion before i push it on my fork and asking for review on it 

I have currently added freetype as git submodule which you might want to change to your fork of freetype.

since im new to opensource dont know what would be best practice. ( adding as submodule under a vendor directory or adding as git submodule under it ?)

eXpl0it3r:
> Hopefully it would work on all compilers now ( tested on clang and wsl's gnu). Need one suggestion before i push it on my fork and asking for review on it

You're very optimistic, which I don't want to reduce, but it's very unlikely that it will just work for all the configurations (compile, shared/static, debug/release, x86/x64, all the OS). 😅 

> I have currently added freetype as git submodule which you might want to change to your fork of freetype.

We wouldn't want to use submodules. The linked approach uses CMake's FetchContent API.

> since im new to opensource dont know what would be best practice. ( adding as submodule under a vendor directory or adding as git submodule under it ?)

You could make a PR, but it's very unlikely that we'll choose this approach. For SFML 2 we certainly won't change the way we distribute dependencies and for SFML 3, we're currently exploring the linked approach with FetchContent.

PrashanthKumar0:
I'd love to work on linking with FetchContent
will let you know if i succeded

eXpl0it3r:
Looks like someone bit the bullet and started to add support #3115

PrashanthKumar0:
I still believe this is not the way to do it
shipping prebuild libs is bad like someone may inject some evil stuff in the library and no one would know.

building from source is better option.

either using fetch content or submodule ..
either would be ok 

since im not very much used with cmake i dont know what exactly is problem with submodules
since i've seen many big projects like chromium using it.

eXpl0it3r:
I agree and hope we'll eventually get to that as well, unfortunately it takes a bit more than just "it works on my PC" 😄

vittorioromeo:
> I still believe this is not the way to do it shipping prebuild libs is bad like someone may inject some evil stuff in the library and no one would know.

This is intended as a stopgap measure until dependencies are built from source, I never intended it to be a permanent solution. Will make it more clear in the PR description.

PR #2822, opened 2023-12-10
Use scoped enumerations in Window module

kimci86:
Part of C++17 modernization for SFML 3, related to [SD: CPP 17 Support](https://github.com/SFML/SFML/wiki/SD%3A-CPP-17-Support#scoped-enumerations) and #2131.
Solution design suggests `enum struct` instead of `enum class` but this is not what we have been doing until now so I used `enum class` for consistency.

I did not convert all enums:
- `sf::Style` and `sf::ContextSettings::Attribute` because they represent flags.
- `sf::Event::EventType` because it is done in #2766.
- `sf::Keyboard::Key` because this PR is already big enough.

### DIFF
--- a/include/SFML/Window/Mouse.hpp
+++ b/include/SFML/Window/Mouse.hpp
@@ -46,32 +46,33 @@ namespace Mouse
 /// \brief Mouse buttons
 ///
 ////////////////////////////////////////////////////////////
-enum Button
+enum class Button
 {
     Left,     //!< The left mouse button
     Right,    //!< The right mouse button
     Middle,   //!< The middle (wheel) mouse button
     XButton1, //!< The first extra mouse button
-    XButton2, //!< The second extra mouse button
-
-    ButtonCount //!< Keep last -- the total number of mouse buttons
+    XButton2  //!< The second extra mouse button
 };
+
+// NOLINTNEXTLINE(readability-identifier-naming)
+static constexpr unsigned int ButtonCount{5}; //!< The total number of mouse buttons
 
 ////////////////////////////////////////////////////////////
 /// \brief Mouse wheels
 ///
 ////////////////////////////////////////////////////////////
-enum Wheel
+enum class Wheel
 {
-    VerticalWheel,  //!< The vertical mouse wheel
-    HorizontalWheel //!< The horizontal mouse wheel
+    Vertical,  //!< The vertical mouse wheel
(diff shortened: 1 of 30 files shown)


ChrisThrasher:

I think I like this. The benefits of scoped enumerations are well known and they certainly apply here to these public API types. 

My one reservation is all the `static_cast`ing that now is required. While potentially outside the scope of this PR, we ought to reconsider what underlying data storage we use. If we used a map instead of an array, we'd benefit in three ways
1. We use fewer C-style arrays. This has safety and expressivity benefits.
2. We can remove the `Count` constants because we will no longer used fixed-capacity data structures.
3. We can stop `static_cast`ing as much since the key pf the map will be the scoped enumeration type instead of its underlying integer representation.

This does mean some extra heap allocation but I have no reason to believe that heap usage will be a problem. The heap usage will go down over time as more and more keys are added to the map so it's just a modest up-front efficiency hit. We could even get fancier and pre-populated the map with values to get all the heap allocations out of the way if need be.

ChrisThrasher (approved):

ChrisThrasher:
Kimci and I spoke about adding a new `std::array` wrapper type that will let us remove all the `static_cast`s while maintaining the benefits of using enumerations to index into arrays. I hope we can get that merged soon since there are other places internally that would really benefit from that.

include/SFML/Window/Mouse.hpp
```diff
@@ -46,32 +46,33 @@ namespace Mouse
...
 ///
 ////////////////////////////////////////////////////////////
-enum Button
+enum class Button
 {
     Left,     //!< The left mouse button
     Right,    //!< The right mouse button
     Middle,   //!< The middle (wheel) mouse button
     XButton1, //!< The first extra mouse button
-    XButton2, //!< The second extra mouse button
```
ChrisThrasher, line 55: Would it make sense to rename these enumerations to `X1` and `X2`?

kimci86: Extra1 and Extra2 would be more meaningful.

ChrisThrasher: I like that too. I'd happily approve a PR with that change.

kimci86: Here we go: #2838

PR #2835, opened 2023-12-14
Fix pkg-config support on Nix

### LINKED ISSUE #2815: Wrong usage of GNUInstallDirs when generating pkg-config files
### Prerequisite Checklist

### Describe your issue here

SFML uses GNUInstallDirs variables (such as CMAKE_INSTALL_LIBDIR) to generate pkg-config files (located in tools/pkg-config).
The problem is that variables like CMAKE_INSTALL_LIBDIR aren't guaranteed to be a path relative to CMAKE_INSTALL_PREFIX, they can even be an absolute path!
This is described in the [official documentation](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html), where it says "However, an absolute path is also allowed" under CMAKE_INSTALL_\<dir>.

All files in tools/pkg-config/*.pc.in define libdir as `${exec_prefix}/@CMAKE_INSTALL_LIBDIR@`, which might concatenate two absolute paths together. This is the case on NixOS for example.

This problem should be easy to fix using cmake_path() as described below:
https://github.com/jtojnar/cmake-snips#concatenating-paths-when-building-pkg-config-files

Also, the install location of *.pc files is affected. Check: [this line in SFML 2.6.x](https://github.com/SFML/SFML/blob/c1c65b53ec2c13e25c83fed5fee9c22562736a60/CMakeLists.txt#L306C15-L306C15) and also [this line in SFML master](https://github.com/SFML/SFML/blob/4e2f6ae655e224b67e4721b99fd52d7f16f49442/cmake/Config.cmake#L69C9-L69C9).
Both define SFML_PKGCONFIG_DIR variable using CMAKE_INSTALL_LIBDIR which later gets concatenated with CMAKE_INSTALL_PREFIX.

### Your Environment

- OS / distro / window manager: Linux / NixOS / XMonad
- SFML version: 2.6.1
- Compiler / toolchain: g++ / gcc
- Special compiler / CMake flags: -DSFML_INSTALL_PKGCONFIG_FILES=TRUE -DSFML_USE_SYSTEM_DEPS=TRUE

### Steps to reproduce

If it helps, I'm packaging SFML 2.6.1 on NixOS [here](https://github.com/ThwyIgo/cmake-sfml-bloat-project/blob/1d62e3ad110bf95e3489c1599e715bd27b63adfa/nix/sfml.nix).
I made a temporary workaround to fix *.pc.in files using sed, but the install location is still wrong.

You can build the derivation and check where pkg-config files were installed. In my system, they're at
/nix/store/\<hash>-sfml-2.6.1/nix/store/\<hash>-sfml-2.6.1/lib/pkgconfig
Clearly two absolute paths were concatenated.

### Expected behavior

pkg-config files installed in /nix/store/\<hash>-sfml-2.6.1/lib/pkgconfig
libdir defined as ${exec_prefix}/lib

### Actual behavior

pkg-config files installed in /nix/store/\<hash>-sfml-2.6.1/nix/store/\<hash>-sfml-2.6.1/lib/pkgconfig
libdir defined as ${exec_prefix}/\<CMAKE_INSTALL_PREFIX>/lib

ChrisThrasher:
## Description

Closes #2815. I also added a missing changelog entry that we meant to add earlier.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,7 @@
 -   Ensure GNUInstallDirs cache vars are included before first used (#2778, #2779)
 -   [macOS] Fix incorrect variable expansion (#2780)
 -   Issue warning when trying to use UCRT MinGW with precompiled MSVCRT depenencies (#2821)
+-   Fix Nix pkg-config support
 
 ### Audio
 
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,7 @@
 
 -   Ensure GNUInstallDirs cache vars are included before first used (#2778, #2779)
 -   [macOS] Fix incorrect variable expansion (#2780)
+-   Issue warning when trying to use UCRT MinGW with precompiled MSVCRT depenencies (#2821)
 
 ### Audio
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -303,9 +303,12 @@ endif()
 sfml_set_option(SFML_INSTALL_PKGCONFIG_FILES ${SFML_INSTALL_PKGCONFIG_DEFAULT} BOOL "TRUE to automatically install pkg-config files so other projects can find SFML")
 
 if(SFML_INSTALL_PKGCONFIG_FILES)
+    # account for CMAKE_INSTALL_LIBDIR potentially being an absolute path
+    file(RELATIVE_PATH SFML_RELATIVE_INSTALL_LIBDIR ${CMAKE_INSTALL_PREFIX} ${CMAKE_INSTALL_FULL_LIBDIR})
+
     # set pkgconfig install directory
     # this could be e.g. macports on mac or msys2 on windows etc.
-    set(SFML_PKGCONFIG_DIR "/${CMAKE_INSTALL_LIBDIR}/pkgconfig")
+    set(SFML_PKGCONFIG_DIR "/${SFML_RELATIVE_INSTALL_LIBDIR}/pkgconfig")
 
     if(SFML_OS_FREEBSD OR SFML_OS_OPENBSD OR SFML_OS_NETBSD)
         set(SFML_PKGCONFIG_DIR "/libdata/pkgconfig")
--- a/tools/pkg-config/sfml-all.pc.in
+++ b/tools/pkg-config/sfml-all.pc.in
@@ -1,6 +1,6 @@
 prefix=@CMAKE_INSTALL_PREFIX@
 exec_prefix=${prefix}
-libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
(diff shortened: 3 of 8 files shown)


ChrisThrasher:
@dkosmari You've previously helped debug pkg-config support. Do you mind testing this to make sure it hasn't broken your workflows?

ChrisThrasher:
@ThwyIgo Please check this out and approve the PR if it looks good to you.

ThwyIgo:
> @ThwyIgo Please check this out and approve the PR if it looks good to you.

LGTM, I approve. But as a random person who doesn't even have a profile picture, I don't have permission to "approve" the PR.

BTW, I'm not the pkg-config guy, but I tested it compiling a simple project I have here with the command:
`g++ src/*.cpp $(pkg-config --cflags --libs sfml-all) -lboost_filesystem -lboost_system -Isrc/include`
Success. (boost has nothing to do with SFML, it's just my project dependency).
Compiling with CMake also works.

The output of `pkg-config --cflags --libs sfml-all` on my system (NixOS) is:
`-I/nix/store/aiy5hvk2ha3l24qx2rssf00zxklq00q9-sfml-2.6.1/include -L/nix/store/aiy5hvk2ha3l24qx2rssf00zxklq00q9-sfml-2.6.1/lib -lsfml-graphics -lsfml-window -lsfml-audio -lsfml-network -lsfml-system`
Which is correct for NixOS.

ThwyIgo:
Oh, actually, I have a suggestion not directly related to the bug.

Why are we using GNUInstallDirs to define libdir in pkg-config, but not for includedir?
There's CMAKE_INSTALL_INCLUDEDIR which we can convert to a relative path and use it in pkg-config.

This would be great, because it would be easy to install .so files in one place, and .hpp files in another place. That's what some libraries do in NixOS. Usually the derivation (package) is divided in "dev" and "out", dev contains .hpp and out contains .so. This allows users of the distro to save space by not having .hpp files in their system.

dkosmari:
> @dkosmari You've previously helped debug pkg-config support. Do you mind testing this to make sure it hasn't broken your workflows?

Tested on Fedora 39, the pkg-config paths are still fine, on both native and cross mingw builds.

changelog.md
```diff
@@ -6,6 +6,8 @@
 
 -   Ensure GNUInstallDirs cache vars are included before first used (#2778, #2779)
 -   [macOS] Fix incorrect variable expansion (#2780)
+-   Issue warning when trying to use UCRT MinGW with precompiled MSVCRT depenencies (#2821)
```
eXpl0it3r, line 9: <3

eXpl0it3r (approved):

PR #2838, opened 2023-12-16
Rename XButton1 and XButton2 into Extra1 and Extra2

kimci86:
This is to avoid repeating Button when using those values now that the sf::Mouse::Button enumeration is a scoped enumeration.

Follow-up of review comment https://github.com/SFML/SFML/pull/2822#discussion_r1428646949

### DIFF
--- a/include/SFML/Window/Mouse.hpp
+++ b/include/SFML/Window/Mouse.hpp
@@ -48,11 +48,11 @@ namespace Mouse
 ////////////////////////////////////////////////////////////
 enum class Button
 {
-    Left,     //!< The left mouse button
-    Right,    //!< The right mouse button
-    Middle,   //!< The middle (wheel) mouse button
-    XButton1, //!< The first extra mouse button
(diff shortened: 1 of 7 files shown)


ChrisThrasher:

ChrisThrasher (approved):

Bambo-Borris (approved):

MarioLiebisch:
I'd actually be tempted to call them `Back` and `Forward`. Or does anyone actually know a mouse where this modern de-facto standard won't match? I think most if not all popular browsers map this for example.

kimci86:
All mice I use do not have extra buttons, but I see from [wikipedia](https://en.wikipedia.org/wiki/Mouse_button#Additional_buttons) those buttons can have other meanings such as changing mouse sensitivity.

MarioLiebisch:
Not exactly. If they influence sensitivity, they're not mapped to those buttons (or they only do so with a modifier/switch). But yes, it's not really standardized.

PR #2839, opened 2023-12-16
Add EnumArray to implement enum-indexed arrays

kimci86:
In a few places, we have arrays indexed by enumerated types. When those enums are scoped enumerations, it means indexing those arrays becomes verbose with the required `static_cast`.

I suggest `EnumArray` template type which wraps an `std::array` and takes enumeration values as array index, doing the `static_cast` internally.
This allows to have less verbose code, more type safety and even add bounds checking for debug builds.

```c++
// Before:
std::array<T, EnumCount> array{};
T x = array[static_cast<std::size_t>(Enum::Value)]; // verbose
T y = array[42]; // danger!

// After:
EnumArray<Enum, T, EnumCount> array{};
T x = array[Enum::Value]; // pure bliss
T y = array[42]; // compilation error for scoped enumeration
                 // assertion failure for unscoped enumeration with too few values
```

In this PR, I use the `EnumArray` type to replace C-style arrays indexed by `Joystick::Axis`, `Mouse::Button` and `Sensor::Type`.
It could eventually also be used for arrays indexed by `Keyboard::Key` and `Keyboard::Scan`.

Note this change is only about internal SFML implementation, not the public API.
This is a follow-up of #2822.

### DIFF
--- a/src/SFML/System/EnumArray.hpp
+++ b/src/SFML/System/EnumArray.hpp
@@ -1,0 +1,75 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2023 Laurent Gomila (laurent@sfml-dev.org)
+//
+// This software is provided 'as-is', without any express or implied warranty.
+// In no event will the authors be held liable for any damages arising from the use of this software.
+//
+// Permission is granted to anyone to use this software for any purpose,
+// including commercial applications, and to alter it and redistribute it freely,
+// subject to the following restrictions:
+//
+// 1. The origin of this software must not be misrepresented;
+//    you must not claim that you wrote the original software.
+//    If you use this software in a product, an acknowledgment
+//    in the product documentation would be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such,
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+#pragma once
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#include <array>
+#include <type_traits>
+
+#include <cassert>
+#include <cstddef>
+
+
+namespace sf::priv
+{
+////////////////////////////////////////////////////////////
+/// \brief Fixed-size array container indexed by an enumeration
+///
+////////////////////////////////////////////////////////////
+template <typename Enum, typename Value, std::size_t Count>
+struct EnumArray : public std::array<Value, Count>
+{
+    static_assert(std::is_enum_v<Enum>, "Enum type parameter must be an enumeration");
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Returns a reference to the element associated to specified \a key
+    ///
+    /// No bounds checking is performed in release builds.
+    ///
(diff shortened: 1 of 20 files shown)


src/SFML/System/EnumArray.hpp
```diff
@@ -0,0 +1,96 @@
...
+        return std::array<Value, Count>::operator[](index);
+    }
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Returns a reference to the element associated to specified \a key
+    ///
+    /// \throws std::out_of_range if \a key underlying value is not within array bounds.
+    ///
+    ////////////////////////////////////////////////////////////
+    constexpr Value& at(Enum key)
```
ChrisThrasher, line 77: Why add `.at()` if these functions are not used internally?

ChrisThrasher (commented):
I like this! Just a few minor comments and then we can get this merged.

src/SFML/System/EnumArray.hpp
kimci86: I added this as a way of making sure we do not use `at(std::size_t)` for type safety. Should I remove it?

ChrisThrasher: Does this prevent the use of `at(std::size_t)`? My impression is that `std::array::at` is still accessible.

kimci86: It accessible with qualified name or casting to base class, like:
```c++
array.std::array<Value, Count>::at(42);
static_cast<const std::array<Value, Count>&>(array).at(42);
```
but not like that:
```c++
array.at(42); // compilation error
```

src/SFML/System/EnumArray.hpp
```diff
@@ -0,0 +1,96 @@
...
+namespace sf::priv
+{
+////////////////////////////////////////////////////////////
+/// \brief Fixed-size array container indexed by an enumeration
+///
+////////////////////////////////////////////////////////////
+template <typename Enum, typename Value, std::size_t Count>
+struct EnumArray : public std::array<Value, Count>
+{
+    ////////////////////////////////////////////////////////////
```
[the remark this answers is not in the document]
kimci86: done

src/SFML/System/EnumArray.hpp
ChrisThrasher: It seems like the type safety benefits are not dependent on these new `at()` overloads existing. If we end up wanting `at()` functions at a later time I'm happy to add them. I just don't want to add code before we need it. Let's go ahead and remove them.

ChrisThrasher (approved):

ChrisThrasher:

src/SFML/Window/SensorManager.hpp
```diff
@@ -30,6 +30,8 @@
 #include <SFML/Window/Sensor.hpp>
 #include <SFML/Window/SensorImpl.hpp>
 
+#include <SFML/System/EnumArray.hpp>
+
```
MarioLiebisch, line 34: Extra empty line.

MarioLiebisch (approved):

src/SFML/Window/SensorManager.hpp
ChrisThrasher: This is preserving the two empty newlines between the last head and `namespace sf::priv`.

MarioLiebisch: Ah, right!

PR #2841, opened 2023-12-17
Use EnumArray for Key or Scancode-indexed arrays

kimci86:
Follow-up of #2839, using `sf::priv::EnumArray` in internal SFML code to implement arrays indexed by `sf::Keyboard::Key` or `sf::Keyboard::Scancode` enumerations.

---

Regarding public API, it was the opportunity to move `sf::Keyboard::Key::KeyCount` and `sf::Keyboard::Scan::ScancodeCount` out of their enumerations.
So they become respectively:
- `sf::Keyboard::KeyCount`
- `sf::Keyboard::ScancodeCount`

This is consistent with other enumerations:
- `sf::Joystick::Axis` -> `sf::Joystick::AxisCount`
- `sf::Mouse::Button` -> `sf::Mouse::ButtonCount`

See commit message for more arguments in favor of this change.

### DIFF
--- a/include/SFML/Window/Keyboard.hpp
+++ b/include/SFML/Window/Keyboard.hpp
@@ -153,8 +153,15 @@ enum Key
     F15,          //!< The F15 key
     Pause,        //!< The Pause key
 
-    KeyCount, //!< Keep last -- the total number of keyboard keys
+    Last = Pause, //!< Keep equal to last enumerator -- used to define KeyCount
 };
+
+////////////////////////////////////////////////////////////
+/// \brief The total number of keyboard keys, ignoring Key::Unknown
+///
+////////////////////////////////////////////////////////////
+// NOLINTNEXTLINE(readability-identifier-naming)
+static constexpr unsigned int KeyCount{static_cast<unsigned int>(Key::Last) + 1};
 
 ////////////////////////////////////////////////////////////
 /// \brief Scancodes
@@ -320,10 +327,17 @@ enum class Scan
     LaunchMail,         //!< Keyboard Launch Mail key
     LaunchMediaSelect,  //!< Keyboard Launch Media Select key
 
-    ScancodeCount //!< Keep last -- the total number of scancodes
+    Last = LaunchMediaSelect, //!< Keep equal to last enumerator -- used to define ScancodeCount
 };
 
 using Scancode = Scan;
+
+////////////////////////////////////////////////////////////
+/// \brief The total number of scancodes, ignoring Scan::Unknown
+///
+////////////////////////////////////////////////////////////
+// NOLINTNEXTLINE(readability-identifier-naming)
+static constexpr unsigned int ScancodeCount{static_cast<unsigned int>(Scan::Last) + 1};
 
 ////////////////////////////////////////////////////////////
 /// \brief Check if a key is pressed
--- a/src/SFML/Window/Unix/KeyboardImpl.cpp
+++ b/src/SFML/Window/Unix/KeyboardImpl.cpp
@@ -48,8 +48,8 @@ namespace
 
 const KeyCode          nullKeyCode = 0;
 const int              maxKeyCode  = 256;
-KeyCode                scancodeToKeycode[static_cast<std::size_t>(sf::Keyboard::Scan::ScancodeCount)]; ///< Mapping of SFML scancode to X11 KeyCode
-sf::Keyboard::Scancode keycodeToScancode[maxKeyCode]; ///< Mapping of X11 KeyCode to SFML scancode
+KeyCode                scancodeToKeycode[sf::Keyboard::ScancodeCount]; ///< Mapping of SFML scancode to X11 KeyCode
+sf::Keyboard::Scancode keycodeToScancode[maxKeyCode];                  ///< Mapping of X11 KeyCode to SFML scancode
 
 ////////////////////////////////////////////////////////////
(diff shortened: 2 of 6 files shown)


src/SFML/Window/Unix/KeyboardImpl.cpp
```diff
@@ -460,11 +462,8 @@ void ensureMapping()
         return;
 
     // Phase 1: Initialize mappings with default values
-    for (auto& keycode : scancodeToKeycode)
-        keycode = nullKeyCode;
-
-    for (auto& scancode : keycodeToScancode)
-        scancode = sf::Keyboard::Scan::Unknown;
+    scancodeToKeycode.fill(nullKeyCode);
+    keycodeToScancode.fill(sf::Keyboard::Scan::Unknown);
```
ChrisThrasher, line 466: 🤩

src/SFML/Window/Unix/KeyboardImpl.cpp
```diff
@@ -46,10 +48,10 @@
...
 {
 
-const KeyCode          nullKeyCode = 0;
-const int              maxKeyCode  = 256;
-KeyCode                scancodeToKeycode[static_cast<std::size_t>(sf::Keyboard::Scan::ScancodeCount)]; ///< Mapping of SFML scancode to X11 KeyCode
-sf::Keyboard::Scancode keycodeToScancode[maxKeyCode]; ///< Mapping of X11 KeyCode to SFML scancode
+const KeyCode                                                                     nullKeyCode = 0;
+const int                                                                         maxKeyCode  = 256;
+sf::priv::EnumArray<sf::Keyboard::Scancode, KeyCode, sf::Keyboard::ScancodeCount> scancodeToKeycode; ///< Mapping of SFML scancode to X11 KeyCode
+std::array<sf::Keyboard::Scancode, maxKeyCode> keycodeToScancode; ///< Mapping of X11 KeyCode to SFML scancode
```
ChrisThrasher, line 54: Why isn't this an `EnumArray` as well?

kimci86: `KeyCode` type from Xlib is a typedef to `unsigned char`, but `EnumArray` is meant to be used with enumerations.

ChrisThrasher:

ChrisThrasher (approved):

binary1248 (approved):

include/SFML/Window/Keyboard.hpp
```diff
@@ -153,9 +153,16 @@ enum Key
     F15,          //!< The F15 key
     Pause,        //!< The Pause key
 
-    KeyCount, //!< Keep last -- the total number of keyboard keys
+    Last = Pause, //!< Keep equal to last enumerator -- used to define KeyCount
```
eXpl0it3r, line 156: What's the point of moving out KeyCount when we still have it in there under a different, imho more confusing name?

If we keep it, we should:
- Not have a comment explaining what it was previously (otherwise just use the old name)
- Have a more speaking name, so nobody confuses it with an actual key name. Like one could easily think `Last` to be some key. So I'd probably with something like `LastKeyEntry` or so, which again begs the question why we can't stick with `KeyCount` to begin with

include/SFML/Window/Keyboard.hpp
```diff
@@ -153,9 +153,16 @@ enum Key
...
 };
 
+////////////////////////////////////////////////////////////
+/// \brief The total number of keyboard keys, ignoring Key::Unknown
+///
+////////////////////////////////////////////////////////////
+// NOLINTNEXTLINE(readability-identifier-naming)
+static constexpr unsigned int KeyCount{static_cast<unsigned int>(Key::Last) + 1};
```
eXpl0it3r, line 164: Is this correct? Because it certainly changes the meaning to the past. Before `KeyCount == Key::Last`, but now `KeyCount == Key::Last + 1` which is not the same.

include/SFML/Window/Keyboard.hpp
kimci86: Previously `KeyCount` was listed as any other key, which all have a distinct underlying value. So `sf::Keybaord::KeyCount` was a valid value for the type `sf::Keyboard::Key`, but it did not represent any actual key, which is weird.

Here `Last` is not the same as `KeyCount`. It is defined equal to another valid key value `Pause` so it is not adding an additional bogus value to the enumeration.
The reason I introduced this `Last` enumerator is to make it obvious that it needs to be updated whenever we add more key values by being visually close to where we would add such values. We could as well not have this `Last` enumerator but we would need to be careful to update `KeyCount` definition below whenever we add more values at the end of the enumeration.

What do you prefer? I will be happy to submit another pull request to improve this.

include/SFML/Window/Keyboard.hpp
kimci86: Before we had `KeyCount == Key::Pause + 1`,
After we have `KeyCount == Key::Last + 1` but `Key::Last` is `Key::Pause` so `KeyCount == Key::Pause + 1`.

eXpl0it3r: Ah, I get it now

include/SFML/Window/Keyboard.hpp
eXpl0it3r: I'd rather not duplicate the information. Either store the value in the enum or outside of it. If there was a way to hide a value from the public API, then that might have been the even better option.

PR #2845, opened 2023-12-17
Remove unhelpful return values

ChrisThrasher:
## Description

These two private functions always return `true`. It's simpler to have no return value than to return the same thing every time.


binary1248 (approved):

src/SFML/Audio/SoundFileWriterWav.hpp
```diff
@@ -89,7 +89,7 @@ class SoundFileWriterWav : public SoundFileWriter
     /// \return True on success, false on error
```
kimci86, line 89: Documentation should be updated.

kimci86 (approved):

PR #2850, opened 2023-12-21
Make Keyboard::Key a scoped enumeration

kimci86:
Continuing with C++17 modernization, follow-up of #2822.

It might be worth doing something to access those enumerators without having to write that much code, but for now it gets consistent with other SFML enumerations (e.g. `sf::Mouse::Button`, `sf::Mouse::Wheel`, `sf::Keyboard::Scan`, `sf::Joystick::Axis`).


ChrisThrasher:

ChrisThrasher (approved):
I like it! I'm glad we can make the key code and scan code APIs symmetric now. This is a big step in the right direction.

eXpl0it3r (approved):

PR #2853, opened 2023-12-22
Simplify `sf::Packet::clear()`

ChrisThrasher:
## Description

Just like `sf::View::reset` perhaps this function should just be removed. If someone wants to clear a packet they can use the assignment operator plus the default constructor to do the same thing. We should not provide APIs that the language already naturally provides. Regardless we can at least simplify our existing implementation.

```cpp
sf::Packet packet;
... // Do stuff with the packet
packet = {}; // This clears it
```

### DIFF
--- a/src/SFML/Network/Packet.cpp
+++ b/src/SFML/Network/Packet.cpp
@@ -59,9 +59,7 @@ std::size_t Packet::getReadPosition() const
 ////////////////////////////////////////////////////////////
 void Packet::clear()
 {
-    m_data.clear();
-    m_readPos = 0;
-    m_isValid = true;
+    *this = {};
 }
 
 


binary1248 (approved):

ChrisThrasher:
On second though, `std::vector::clear` leave the capacity of the vector untouched whereas this PR I believe will drop the capacity back down to zero resulting in more allocations as the packet gets refilled with data.

ChrisThrasher:
On third though, I can't find any information on what happens to the capacity of a vector when you assign to it. Does the capacity necessarily shrink? That's possible but standard library implementations try to be very efficient and not reducing the capacity is one more way to ensure efficiency without breaking the contract of the `std::vector` API so perhaps this PR does not result in any reduction in efficiency.

ChrisThrasher:
What about `sf::Packet::m_sendPos`? It's not touched by this function yet it's used in some distant classes. Does this PR fix a bug where `m_sendPos` doesn't get zeroed or does it introduce a bug where `m_sendPos` shouldn't get zeroed?

eXpl0it3r:
We may want a bit more tests to ensure this doesn't break performance.

PR #2859, opened 2023-12-24
Use vectors for expressing size parameters

ChrisThrasher:
## Description

I found some Objective-C++ interfaces that still used parameter pairs for expressing sizes.


binary1248 (approved):

ChrisThrasher:
I found some more Obj-C interfaces that needed this treatment

PR #2860, opened 2023-12-27
Deduplicate `Window{Base}::create` implementations

### LINKED ISSUE #2851: Deduplicate implementations of `WindowBase::create` and `Window::create`
### Prerequisite Checklist

These two functions are nearly identical in their implementation. It would make the library easier to maintain if the common code was extracted out into one place.

### Use Cases

Gotta add some text here so GitHub will let me submit the issue.

### API Example

_No response_

ChrisThrasher:
## Description

Closes #2851

### DIFF
--- a/include/SFML/Window/WindowBase.hpp
+++ b/include/SFML/Window/WindowBase.hpp
@@ -464,6 +464,18 @@ private:
     friend class Window;
 
     ////////////////////////////////////////////////////////////
+    /// \brief Create (or recreate) the window
+    ///
+    /// Implementation detail for sharing underlying implementation
+    /// with sf::Window
+    ///
+    /// \param mode  Video mode to use (defines the width, height and depth of the rendering area of the window)
+    /// \param style %Window style, a bitwise OR combination of sf::Style enumerators
+    ///
+    ////////////////////////////////////////////////////////////
+    void create(VideoMode mode, std::uint32_t& style);
+
+    ////////////////////////////////////////////////////////////
     /// \brief Processes an event before it is sent to the user
     ///
     /// This function is called every time an event is received
--- a/src/SFML/Window/Window.cpp
+++ b/src/SFML/Window/Window.cpp
@@ -74,45 +74,8 @@ void Window::create(VideoMode mode, const String& title, std::uint32_t style)
 ////////////////////////////////////////////////////////////
 void Window::create(VideoMode mode, const String& title, std::uint32_t style, const ContextSettings& settings)
 {
-    // Destroy the previous window implementation
-    close();
-
-    // Fullscreen style requires some tests
-    if (style & Style::Fullscreen)
-    {
-        // Make sure there's not already a fullscreen window (only one is allowed)
-        if (getFullscreenWindow())
-        {
-            err() << "Creating two fullscreen windows is not allowed, switching to windowed mode" << std::endl;
-            style &= ~static_cast<std::uint32_t>(Style::Fullscreen);
-        }
-        else
-        {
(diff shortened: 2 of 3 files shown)


ChrisThrasher:

binary1248:
This change adds an OpenGL specific header to `WindowBase` although it was specifically designed to be OpenGL-free. If you really want to reuse that common block of code in `::create` you should move the differentiating lines of code into `::initialize` instead of adding `ContextSettings` to `::create`.

ChrisThrasher:
> This change adds an OpenGL specific header to WindowBase

What's the problem with doing this? WindowBase.cpp already used it.

binary1248:
Not in the header. Everyone should know by now that OpenGL forces one to have to make dirty implementation decisions from time to time. Creating a "dummy context" in `::create` was necessary even when not using OpenGL in order to not break stuff on specific OSes. This is unavoidable, but we still get to choose whether we want to expose this dirtiness via the public header. The header was clean before this change, and would no longer be after it.

ChrisThrasher:
> The header was clean before this change, and would no longer be after it.

It's only "dirty" in some philosophical sense. I cannot think of any technical issues with what I did.

ChrisThrasher:
I used a forward declaration to avoid including ContextSettings.hpp.

ChrisThrasher:

ChrisThrasher:

binary1248:
Deduplicating code no matter the design cost can also devolve into a philosophical debate. The code before worked in every technical sense. We can't preach a different standard every day depending on what metric we are looking at.

Like I said, the only reason `WindowBase` exists is to provide Vulkan users with a class that is from their point of view OpenGL-free. Having to break this requirement just to save a few lines of code makes the entire class useless. If we were really hellbent on reducing the total lines of code we could just throw out `WindowBase` entirely and tell people to use `Window` instead.

When designing new public interfaces and eventually writing the accompanying implementation we should never let the implementation dictate what the public interface will look like if there are viable alternatives. Having a clean API philosophy is what separates good/learnable libraries from those that were written by people who "just needed to get the job done". We should really hold ourselves to higher standards.

ChrisThrasher:
Let me know if the latest state of the PR works for you or if that forward declaration is unacceptable.

binary1248:
Forward declarations just alleviate the need to have to include a header which is C-legacy, having to expose private function signatures in the publicly visible header is also C++-legacy. The semantic dependency they force is still there. Hypothetically, if we were to rename `ContextSettings` to `SomeOtherContextSettings` or one day even completely get rid of the OpenGL implementation, we would also have to touch this header although the class really shouldn't have anything to do with OpenGL.

This is a uniquely C/C++ problem, as you noticed it is much harder to design clean APIs around weird OS implementation requirements than would be necessary in other languages. Over the decades C++ developers have designed workarounds for this problem like PIMPL, but they all come with their own tradeoffs. The tradeoff that was made in the current implementation was duplication of that small block of code, however IMHO it is the lesser of all evils if we really value having a clean _C++_ API.

ChrisThrasher:
> you should move the differentiating lines of code into ::initialize instead of adding ContextSettings to ::create.

If you explain this some more then perhaps we can find a solution that satisfies your ideals of cleanliness. Otherwise this is the solution we're best off moving forward with.

binary1248:
https://github.com/SFML/SFML/compare/master...window_create

ChrisThrasher:
Your solution looks pretty obvious in hindsight. Thanks!

binary1248 (approved):

ChrisThrasher:

PR #2870, opened 2023-12-31
Modernize management of X11 resources

ChrisThrasher:
## Description

This approach uses RAII to manage resources rather than relying on correctly calling cleanup functions at all places where scope exits.

See #2540 and #2801 for similar PRs. Both of those have aged well if you ask me so I'm confident this pattern is worth employing in more places.

### DIFF
--- a/src/SFML/Window/Unix/Display.cpp
+++ b/src/SFML/Window/Unix/Display.cpp
@@ -42,27 +42,27 @@
 
 namespace
 {
-// The shared display and its reference counter
-Display*             sharedDisplay     = nullptr;
-unsigned int         referenceCount    = 0;
-XIM                  sharedXIM         = nullptr;
-unsigned int         referenceCountXIM = 0;
-std::recursive_mutex mutex;
-
-using AtomMap = std::unordered_map<std::string, Atom>;
-AtomMap atoms;
+// A nested named namespace is used here to allow unity builds of SFML.
+namespace UnixDisplayImpl
+{
+std::weak_ptr<Display> weakSharedDisplay;
+std::recursive_mutex   mutex;
+} // namespace UnixDisplayImpl
 } // namespace
+
 
 namespace sf::priv
 {
 ////////////////////////////////////////////////////////////
-Display* openDisplay()
+std::shared_ptr<Display> openDisplay()
 {
-    const std::lock_guard lock(mutex);
+    const std::lock_guard lock(UnixDisplayImpl::mutex);
 
-    if (referenceCount == 0)
+    auto sharedDisplay = UnixDisplayImpl::weakSharedDisplay.lock();
+    if (!sharedDisplay)
     {
-        sharedDisplay = XOpenDisplay(nullptr);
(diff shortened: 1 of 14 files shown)


ChrisThrasher:
https://quick-bench.com/q/svYpyGa7FyheM2iiu0S3VDCLGgE

Incrementing and decrementing the atomic ref count inside a `std::shared_ptr` should be faster than locking and unlocking the mutex like we were previously doing.

src/SFML/Window/Unix/Display.cpp
```diff
@@ -73,31 +67,20 @@ Display* openDisplay()
...
-
-////////////////////////////////////////////////////////////
-XIM openXim()
+std::shared_ptr<_XIM> openXim()
 {
     const std::lock_guard lock(mutex);
 
     assert(sharedDisplay != nullptr && "Shared display is null. Call priv::openDisplay() to initialize it.");
 
-    if (referenceCountXIM == 0)
```
ChrisThrasher, line 100: I'm not sure why we need this `assert`. It seems more reasonable to call `openDisplay` inside this function to ensure a display is open rather than hope this precondition is always true. That seemed out of scope for this PR so I didn't make that change.

src/SFML/Window/Unix/Display.cpp
```diff
@@ -42,27 +42,21 @@
...
-// The shared display and its reference counter
-Display*             sharedDisplay     = nullptr;
-unsigned int         referenceCount    = 0;
-XIM                  sharedXIM         = nullptr;
-unsigned int         referenceCountXIM = 0;
-std::recursive_mutex mutex;
-
-using AtomMap = std::unordered_map<std::string, Atom>;
-AtomMap atoms;
+std::shared_ptr<Display> sharedDisplay;
```
kimci86, line 45: The previous behavior would be better reproduced with an `std::weak_ptr`. For reference, we already use `std::weak_ptr` to manage shared OpenGL context in `GlContext.cpp`.

ChrisThrasher: My previous implementation was keeping the display and XIM alive forever because the global shared_ptr didn't go out of scope. Now I'm using a weak_ptr and reassigning that weak pointer when the underlying shared state gets rebuilt.

ChrisThrasher:

adgajda:
You removed every occurrence of `closeDisplay(display)`, shouldn't it be replaced with display.reset() to decrement shared_ptr counter?
In that case initialization of display could look like this:
```
struct XDisplayDeleter
{
    void operator()(Display* display)
    {
        XCloseDisplay(display);
    }
};
sharedDisplay = std::shared_ptr<Display>(XOpenDisplay(nullptr), XDisplayDeleter{});
```

Please correct me if I am wrong. @ChrisThrasher

ChrisThrasher:
The destruction of the shared pointer decrements the ref count. No need to manually do that.

eXpl0it3r (approved):

PR #2873, opened 2024-01-08
Fixed F13-F24 and consumer key names not being returned on Windows.

### LINKED ISSUE #2872: `sf::Keyboard::getDescription` returns "Unknown" for F13+ keys on Win32
### Prerequisite Checklist

### Describe your issue here

When calling `sf::Keyboard::getDescription` on Scancode `sf::Keyboard::Scan::F13`, and other higher function keys, the function returns "Unknown" on Win32. 

Multiple users reported this problem in my implementation, but only on Win32.

### Your Environment

- OS / distro / window manager: Win32
- SFML version: 2.6.1
- Compiler / toolchain: msbuild + vcpkg

### Steps to reproduce

1. No specific steps needed aside from the code below
```cpp
#include <SFML/Graphics.hpp>
#include <SFML/Window/Event.hpp>
#include <SFML/Window/Keyboard.hpp>

int main()
{
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        {
                window.close();

            if (event.type == sf::Event::KeyPressed) {
                std::cout << sf::Keyboard::getDescription(sf::Keyboard::Scancode::F13).toAnsiString() << std::endl;
            }

        }

        window.clear();
        window.display();
    }
}
```

### Expected behavior

Returns a proper key description, such as "F13" or "Function key 13".

### Actual behavior

Returns "Unknown"

binary1248:
Title.

Fixes #2872.

Test with any keyboard that has consumer keys. For F13-F24 remap some key using [powertoys](https://learn.microsoft.com/en-us/windows/powertoys/).

### DIFF
--- a/src/SFML/Window/Win32/InputImpl.cpp
+++ b/src/SFML/Window/Win32/InputImpl.cpp
@@ -512,6 +512,38 @@ UINT sfScanToVirtualKey(sf::Keyboard::Scancode code)
     // clang-format on
 }
 
+std::optional<sf::String> sfScanToConsumerKeyName(sf::Keyboard::Scancode code)
+{
+    // Convert an SFML scancode to a Windows consumer keyboard key name
+    // Reference: https://learn.microsoft.com/en-us/windows/win32/inputdev/about-keyboard-input#keystroke-messages
+    // clang-format off
+    switch (code)
+    {
+        case sf::Keyboard::Scan::MediaNextTrack:     return "Next Track";
+        case sf::Keyboard::Scan::MediaPreviousTrack: return "Previous Track";
+        case sf::Keyboard::Scan::MediaStop:          return "Stop";
+        case sf::Keyboard::Scan::MediaPlayPause:     return "Play/Pause";
+        case sf::Keyboard::Scan::VolumeMute:         return "Mute";
+        case sf::Keyboard::Scan::VolumeUp:           return "Volume Increment";
+        case sf::Keyboard::Scan::VolumeDown:         return "Volume Decrement";
+        case sf::Keyboard::Scan::LaunchMediaSelect:  return "Consumer Control Configuration";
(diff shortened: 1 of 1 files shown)


src/SFML/Window/Win32/InputImpl.cpp
```diff
@@ -512,6 +512,38 @@ UINT sfScanToVirtualKey(sf::Keyboard::Scancode code)
     // clang-format on
 }
 
+const char* sfScanToConsumerKeyName(sf::Keyboard::Scancode code)
```
ChrisThrasher, line 515: Would it be more expressive to return an optional? Perhaps `std::optional<std::string_view>`? If I see an API return a `const char*` I'm not automatically going to assume it may return a null pointer.

binary1248:
Updated.

src/SFML/Window/Win32/InputImpl.cpp
```diff
@@ -595,10 +627,22 @@ Keyboard::Scancode delocalize(Keyboard::Key key)
 ////////////////////////////////////////////////////////////
 String getDescription(Keyboard::Scancode code)
 {
-    const WORD winCode = sfScanToWinScanExtended(code);
-    const int  bufSize = 1024;
-    WCHAR      name[bufSize];
-    const int  result = GetKeyNameText(winCode << 16, name, bufSize);
+    // Try to translate the scan code to a consumer key
+    if (const auto consumerKeyName = sfScanToConsumerKeyName(code))
+        return consumerKeyName->data();
```
ChrisThrasher, line 632: I now see that the return value is only used here to create an `sf::String`. Would it be easier if `sfScanToConsumerKeyName` already returned an `sf::String`?

binary1248:
Updated.

ChrisThrasher (approved):

PR #2880, opened 2024-01-14
Add missing `std::` namespace

ChrisThrasher:
## Description

I wish there was a systematic way to find uses of C standard library functions that have C++ standard library equivalents. The only way I find this is by staring at enough code for a long time.

Eventually once we can use `import std;` then any remaining uses of C standard library functions will be detected since `import std;` does not include the C standard library.


ChrisThrasher:
By the way, `std::fwrite` and a few other standard library functions we use have a return value that we're ignoring. I came across a clang-tidy check in the `cert-` category that can flag such ignored return values. That may be something we should enable.

binary1248 (approved):

PR #2883, opened 2024-01-16
Clarify instructions for how to use `SFML_ROOT`

ChrisThrasher:
## Description

Apologies for accidentally pushing https://github.com/SFML/SFML/commit/30a986e63261eea3ae90e5d16b5b894f2f712403 straight to `master`. Luckily that commit is mostly good but it needs some refinement.

`SFML_ROOT` can simply point to the directory in which SFML was installed. This is verified by our install interface tests which do exactly that, even when SFML is built as a framework. CMake can figure out where to find the config module if you give it this much information.

### DIFF
--- a/cmake/SFMLConfig.cmake.in
+++ b/cmake/SFMLConfig.cmake.in
@@ -41,9 +41,8 @@
 # Additionally, keep in mind that SFML frameworks are only available as release libraries unlike dylibs which
 # are available for both release and debug modes.
 #
-# If SFML is not installed in a standard path, you can use the SFML_ROOT CMake variable
-# to tell CMake where SFML's config file is located (PREFIX/lib/cmake/SFML for a library-based installation,
-# and PREFIX/SFML.framework/Resources/CMake on macOS for a framework-based installation).
+# If SFML is not installed in a standard path, you can use the SFML_ROOT CMake variable to tell CMake in what directory
+# SFML was installed.
 #
 # Output
 # ------


binary1248 (approved):

kimci86:
For the record, using `SFML_DIR` stopped working with static SFML libraries on Windows after #2666 was merged.
I noticed this when trying to upgrade [SFML-Input](https://github.com/eXpl0it3r/SFML-Input) to SFML master version.

Using `SFML_DIR`, configuring SFML-Input fails with this error (see [workflow run](https://github.com/eXpl0it3r/SFML-Input/actions/runs/7453580133/job/20279261861)).

```
cmake -S $GITHUB_WORKSPACE -B $GITHUB_WORKSPACE/build -DCMAKE_INSTALL_PREFIX=$GITHUB_WORKSPACE/install -DCMAKE_VERBOSE_MAKEFILE=ON -DSFML_DIR=$GITHUB_WORKSPACE/SFML/install/lib/cmake/SFML -DBUILD_SHARED_LIBS=OFF -DSFML_USE_STATIC_STD_LIBS=ON
```

```
-- Could NOT find Freetype (missing: FREETYPE_LIBRARY) 
-- Could NOT find OpenAL (missing: OPENAL_LIBRARY OPENAL_INCLUDE_DIR) 
-- Could NOT find Vorbis (missing: OGG_INCLUDE_DIR OGG_LIBRARY VORBIS_INCLUDE_DIR VORBIS_LIBRARY) 
-- Could NOT find FLAC (missing: FLAC_LIBRARY FLAC_INCLUDE_DIR) 
-- Found SFML 3.0.0 in D:/a/SFML-Input/SFML-Input/SFML/install/lib/cmake/SFML
-- Configuring done (5.1s)
CMake Error at SFML/install/lib/cmake/SFML/SFMLStaticTargets.cmake:110 (set_target_properties):
-- Generating done (0.0s)
  The link interface of target "SFML::Audio" contains:

    Vorbis::vorbisfile

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

  SFML/install/lib/cmake/SFML/SFMLConfig.cmake:125 (include)
  CMakeLists.txt:20 (find_package)
```

Using `SFML_ROOT`, configuring SFML-Input goes fine (see [workflow run](https://github.com/eXpl0it3r/SFML-Input/actions/runs/7547191416/job/20546600448)).
Dependency headers are not found but that is expected as they are not part of SFML installation tree and are not needed to use SFML.

```
cmake -S $GITHUB_WORKSPACE -B $GITHUB_WORKSPACE/build -DCMAKE_INSTALL_PREFIX=$GITHUB_WORKSPACE/install -DCMAKE_VERBOSE_MAKEFILE=ON -DSFML_ROOT=$GITHUB_WORKSPACE/SFML/install -DBUILD_SHARED_LIBS=OFF -DSFML_USE_STATIC_STD_LIBS=ON
```

```
-- Found Freetype: D:/a/SFML-Input/SFML-Input/SFML/install/lib/freetype.lib  
-- Could NOT find OpenAL (missing: OPENAL_INCLUDE_DIR) 
-- Could NOT find Vorbis (missing: OGG_INCLUDE_DIR VORBIS_INCLUDE_DIR) 
-- Could NOT find FLAC (missing: FLAC_INCLUDE_DIR) 
```

So I think documenting that `SFML_ROOT` is preferred is welcome :+1:

kimci86:
About 30a986e63261eea3ae90e5d16b5b894f2f712403:
https://github.com/SFML/SFML/blob/30a986e63261eea3ae90e5d16b5b894f2f712403/examples/android/app/src/main/jni/CMakeLists.txt#L4

Isn't SFML_ROOT supposed to be `${CMAKE_ANDROID_NDK}/sources/third_party/sfml` in that case?

ChrisThrasher:
> Isn't SFML_ROOT supposed to be `${CMAKE_ANDROID_NDK}/sources/third_party/sfml` in that case?

I tried that but CI was failing because it couldn't find the SFML installation. I think the extra directory used to specify the architecture is messing things up. Keeping the same directory works so I'll leave it there unless someone finds a better solution.

texus:
I was expecting it to work because CMake's documentation of find_package mentions that it searches in `<prefix>/lib/<arch>/cmake/<name>*/`. However, apparently `<arch>` is specified by `CMAKE_LIBRARY_ARCHITECTURE` which will have a value of e.g. `arch64-linux-android` when `CMAKE_ANDROID_ARCH_ABI` is set to `arm64-v8a`.

If I call `set(CMAKE_LIBRARY_ARCHITECTURE ${CMAKE_ANDROID_ARCH_ABI})` before find_package, then setting SFML_ROOT to `${CMAKE_ANDROID_NDK}/sources/third_party/sfml` actually does work. Although I don't believe that overwriting CMAKE_LIBRARY_ARCHITECTURE would be the right thing to do.

ChrisThrasher:
I must have been mistaken because the current directory does not result in SFML being found.

> Edit: maybe it should be re-evaluated if the following line is still the correct way to install the library. I wonder how other c++ libraries that support Android are handling this.

I also think we should reevaluate this choice. In general the installation location is something by whoever is building the library. It seems odd that SFML dictates the install location on Android. This may be convenient but it seems unnecessary.

PR #2894, opened 2024-02-03
Improve PCH and enable PCH for Catch2 and tests

vittorioromeo:
tl;dr:

- Compilation speed unchanged with `-DSFML_BUILD_TEST_SUITE=1 -DSFML_ENABLE_PCH=0`
- Significant speedup with `-DSFML_BUILD_TEST_SUITE=1 -DSFML_ENABLE_PCH=1`

Full analysis logs available here:

- https://gist.github.com/vittorioromeo/480aa5ca9b940754844b07b4ea69c70e


vittorioromeo:

CMakeLists.txt
```diff
@@ -211,6 +211,15 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 set_property(GLOBAL PROPERTY PREDEFINED_TARGETS_FOLDER "CMake")
 
+# add an option for building the test suite
+sfml_set_option(SFML_BUILD_TEST_SUITE FALSE BOOL "TRUE to build the SFML test suite, FALSE to ignore it")
```
kimci86, line 215: Should this be defined only if `SFML_CONFIGURE_EXTRAS` option is set?

ChrisThrasher: Yes it is

test/CMakeLists.txt
```diff
@@ -21,6 +21,13 @@ set_target_properties(Catch2WithMain PROPERTIES EXPORT_COMPILE_COMMANDS OFF)
 get_target_property(CATCH2_INCLUDE_DIRS Catch2 INTERFACE_INCLUDE_DIRECTORIES)
 target_include_directories(Catch2 SYSTEM INTERFACE ${CATCH2_INCLUDE_DIRS})
 
+# enable precompiled headers
+if (SFML_ENABLE_PCH)
+    message(VERBOSE "enabling PCH for Catch2")
```
eXpl0it3r, line 26: Is the message needed?

```suggestion
    message(VERBOSE "Enabling PCH for Catch2")
```

CMakeLists.txt
eXpl0it3r: Can't options also be declared later on, so that `if` statement still work in CMake anyways?

Meaning, there's no need to move the option

test/CMakeLists.txt
vittorioromeo: We have the same message for every other PCH, with lowercase "enabling". This message is consistent with the status quo.

CMakeLists.txt
vittorioromeo: @eXpl0it3r: you are correct, I will move it back.

vittorioromeo:

vittorioromeo:

eXpl0it3r:
Fails to build for me on Windows with VS2022:

```
cmake -B build -S . -DSFML_BUILD_TEST_SUITE=1 -DSFML_ENABLE_PCH=1
cmake --build build
```

```
MSBuild version 17.8.3+195e7f5a3 for .NET Framework

  1>Checking Build System
  Building Custom Rule SFML/build/_deps/catch2-src/src/CMakeLists.txt
  cmake_pch.cxx
SFML\src\SFML\PCH.hpp(31,10): error C1083: Cannot open include file: 'SFML/Config.hpp': No such file o
r directory [SFML\build\_deps\catch2-build\src\Catch2.vcxproj]
  (compiling source file 'CMakeFiles/Catch2.dir/cmake_pch.cxx')

  Building Custom Rule SFML/src/SFML/System/CMakeLists.txt
  cmake_pch.cxx
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.22000.0\um\dinput.h: DIRECTINPUT_VERSION undefined. Defaulting to ve
  rsion 0x0800
SFML\src\SFML\PCH.hpp(74,10): error C1083: Cannot open include file: 'catch2/catch_test_macros.hpp': N
o such file or directory [SFML\build\src\SFML\System\sfml-system.vcxproj]
  (compiling source file 'CMakeFiles/sfml-system.dir/cmake_pch.cxx')
```

eXpl0it3r:
Are we running any build with tests AND PCH enabled in the CI?

vittorioromeo:
> Are we running any build with tests AND PCH enabled in the CI?

Duh, I always forget we are not. We should probably add a job or two...

test/CMakeLists.txt
```diff
@@ -21,6 +21,21 @@ set_target_properties(Catch2WithMain PROPERTIES EXPORT_COMPILE_COMMANDS OFF)
...
+    target_precompile_headers(Catch2 PRIVATE ${PROJECT_SOURCE_DIR}/src/SFML/PCH.hpp)
+
+    # necessary to find `WindowsHeader.hpp` on Windows
+    target_include_directories(Catch2 PRIVATE ${PROJECT_SOURCE_DIR}/src)
+
+    # necessary to find SFML headers from Catch2 through the PCH header
+    target_include_directories(Catch2 PRIVATE ${PROJECT_SOURCE_DIR}/include)
+
+    # necessary to find Catch2 headers from SFML through the PCH header
+    target_include_directories(sfml-system PRIVATE ${CATCH2_INCLUDE_DIRS})
```
vittorioromeo, line 36: @eXpl0it3r: this seems to solve the build on Windows, if you have any suggestions on a more elegant solution I'm all ears :)

eXpl0it3r:
Win 11 / VS 2022 (no Ninja or parallelization)

Before this change: 1min 4s
After this change: 37s

(Very scientifically measured with a stop watch 😄)

test/CMakeLists.txt
eXpl0it3r: I wonder if there isn't some include directory variable thingy from the target we can use. Maybe @ChrisThrasher has an idea.

src/SFML/PCH.hpp
```diff
@@ -62,3 +63,16 @@
...
 #include <cstdlib>
 #include <cstring>
+
+
+////////////////////////////////////////////////////////////
+// Precompiled Headers (Test Suite Only)
+////////////////////////////////////////////////////////////
+
+#ifdef SFML_BUILD_TEST_SUITE
+
```
ChrisThrasher, line 73: Why not just use a separate target_precompiled_headers call within test/CMakeLists.txt? Feels out of place to add non-SFML headers to PCH.hpp.

vittorioromeo: You mean to directly pass `catch_test_macros' instead of `PCH.hpp`?

That was my first approach, but I noticed that Catch2 shares quite a lot of headers in common with SFML (e.g. `ostream`, `algorithm`, `vector`, `memory`), so it made sense to reuse the existing `PCH.hpp` rather than creating a new ad-hoc one for Catch2.

The other issue is that we want `catch_test_macros` to also be part of SFML's own PCH because we use that PCH in our own test suite. 

So keeping everything in the same PCH seemed the most reasonable choice here.

We'd still have to keep `catch2/catch_test_macros` in `SFML`'s own PCH to speed up compilation of our test drivers.

vittorioromeo: > Can’t you have multiple PCH sets? And then reuse the PCH set for the core library alongside the PCH for catch_test_macros?

I don't see how that solves the issue, it would limit PCH reuse unless I am mistaken...

There are basically three categories of targets:

1. SFML library: needs PCH_STDLIB + PCH_SFML
2. SFML tests: needs PCH_STDLIB + PCH_SFML + PCH_CATCH
3. Catch2 library: needs PCH_STDLIB + PCH_CATCH

Right now, my PCH is PCH_STDLIB + PCH_SFML + PCH_CATCH. This allows me to reuse the same PCH for all three categories of targets. 

I definitely want "SFML library" and "SFML tests" to reuse the same PCH, to avoid recompiling two different PCHs -- but that requires that PCH to have access to Catch2 headers anyway. 

If we were to go with your solution, then we'd have a few PCH files:

1. `PCHStdlib.hpp`
2. `PCHStdlibSFML.hpp` (includes `PCHStdlib.hpp`)
3. `PCHStdlibCatch.hpp` (includes `PCHStdlib.hpp`)
4. `PCHStdlibSFMLCatch.hpp` (include `PCHSFML.hpp` and `PCHCatch.hpp`)

- SFML libraries would use `PCHStdlibSFML.hpp`.
- SFML tests would use `PCHStdlibSFMLCatch.hpp`.
- Catch would use `PCHStdlibCatch.hpp`.

I would be able to use `PCHStdlibSFML.hpp` for SFML libraries, and use `REUSE_FROM` from `sfml-system`.

But if I had to use `PCHStdlibSFMLCatch.hpp` for tests I would lose the ability to `REUSE_FROM` from `sfml-system`, which is a step backwards compared to the current approach. The same also applies to SFML examples.

I don't think going down this route would be better... in the end PCHs are an optimization of build times which already heavily screws up header hygiene, I think it's acceptable to use a few `target_include_directories` to make everything work when PCHs are enabled.

vittorioromeo:

vittorioromeo:

vittorioromeo:

vittorioromeo:
@ChrisThrasher: any feedback on this? It makes a significant difference on my machine

ChrisThrasher:
Sorry, but I'm not in favor of this PR. I think the complexity it adds outweighs whatever benefits it may have. A clean build of the test suite (including Catch2) takes under 10 seconds on my 2021 laptop. Adding complexity to improve compile time for users is one thing but I'm much less motivated to add complexity to improve compile times for the small number of people, myself included, who are building the tests on a regular basis.

vittorioromeo:
> Sorry, but I'm not in favor of this PR. I think the complexity it adds outweighs whatever benefits it may have. A clean build of the test suite (including Catch2) takes under 10 seconds on my 2021 laptop. Adding complexity to improve compile time for users is one thing but I'm much less motivated to add complexity to improve compile times for the small number of people, myself included, who are building the tests on a regular basis.

What exactly do you find most complicated about this PR, and what do you think would need to be simplified in order for you to consider approving it?

ChrisThrasher:
I can't picture a way of transforming this PR into something I'd be willing to approve

vittorioromeo:
> Win 11 / VS 2022 (no Ninja or parallelization)
> 
> Before this change: 1min 4s After this change: 37s
> 
> (Very scientifically measured with a stop watch 😄)

I think it's quite sad to ignore these sort of improvements. I also think we should start using PCH as our default in CI jobs and only have one or two sanity-check non-PCH jobs to save time and computing power (let's be green!). 

Anyone else on my side? @eXpl0it3r @Bromeon @binary1248

PR #2896, opened 2024-02-04
Improve SocketSelector in-documentation example

vittorioromeo:
No need for `std::unique_ptr` nor `std::list` now that we have move semantics. Also need to check return value for `.listen`.


vittorioromeo:

ChrisThrasher (approved):

vittorioromeo:

PR #2897, opened 2024-02-04
Self-contained '.inl' headers + minor refactoring

vittorioromeo:
This PR changes the way we deal with `.inl` files. 

Prior to the PR, we included `.inl` files inside the scope of either the `sf` or `sf::priv` namespace, and the code inside an `.inl` file always relied on headers defined in the corresponding `.hpp` file.

That approach has several drawbacks:

1. When looking directly at the `.inl` file, it is impossible to tell what namespace we're in.
2. Header dependencies that are not used in a component's `.hpp` file but are used only for the purpose of the `.inl` file are not obvious, as they are listed with all other headers in the `.hpp` file.
3. IDE support is pretty much broken when looking directly at a `.inl` file.

This PR changes our approach to the following:

1. In a `.hpp` file, an `.inl` file is included outside of any namespace, last thing in the file before the footer comment.
2. At the beginning of an `.inl` file, the corresponding `.hpp` header is included first. This doesn't cause a circular dependency inclusion due to the fact that the corresponding header has a `#pragma once`, and that the `.inl` file is only included from the corresponding `.hpp` header. **IMPORTANT:** this change makes IDEs work seamlessly (`clangd`, autocomplete, etc...) when looking directly at the `.inl` file.
3. Any header dependency that is used only for the code defined inside a `.inl` file is included directly in the `.inl` file, and not in the corresponding `.hpp` file.
4. Namespaces are explicitly open directly in the `.inl` file.

Excluding point (2), we were already doing this for `SoundFileFactory.inl`, so there already was precedent for this approach in SFML. Not sure why the rest of `.inl` file used the weird approach described at the beginning.

I've also renamed a `byte` variable to `testByte` as `g++` complains about a shadowing warning on MSYS2 + MinGW2. This is inside of a test, so it doesn't really matter.

The only possible issue I can see with this PR is that including the corresponding `.hpp` file in an `.inl` file might cause some issue on some compilers/platforms, but CI will confirm that :)


vittorioromeo:

test/Graphics/Font.test.cpp
```diff
@@ -73,8 +73,8 @@ TEST_CASE("[Graphics] sf::Font", runDisplayTests())
         SECTION("Invalid data and size")
         {
             CHECK(!font.loadFromMemory(nullptr, 1));
-            const std::byte byte{0xCD};
```
ChrisThrasher, line 76: What did the compiler think was being shadowed?

ChrisThrasher (approved):
If this makes language servers work with .inl files then I'm a fan. It's always been mildly annoying how .inl files are full of red squiggly lines.

test/Graphics/Font.test.cpp
eXpl0it3r: Maybe some evil global? 🤔

vittorioromeo: > Maybe some evil global? 🤔

Yep, exactly that :(

vittorioromeo:
> If this makes language servers work with .inl files then I'm a fan. It's always been mildly annoying how .inl files are full of red squiggly lines.

It does :)

I need to figure out a way to silence the circular dependency analysis... the only thing I can think of is to add some sort of macro wrapper around the include that an IDE can at least define to be true...

ChrisThrasher:
> I need to figure out a way to silence the circular dependency analysis...

I pushed a fix. If you like it then you can squash the commit into your commit. Feel free to remove it if you find a better solution as long as we get to keep using the check.

vittorioromeo:
> > I need to figure out a way to silence the circular dependency analysis...
> 
> I pushed a fix. If you like it then you can squash the commit into your commit. Feel free to remove it if you find a better solution as long as we get to keep using the check.

I like the solution, will squash.

vittorioromeo:

PR #2898, opened 2024-02-04
Turn `sf::Vertex` into an aggregate `struct`

vittorioromeo:
As the title says. Figured it'd be faster to make the change and show you what it looks like rather than discuss whether it's worthwhile doing it.

IMHO, this is a good simplification that promotes the use of simple `struct`s and aggregate initialization instead of exposing several manually-crafted constructors that do the same thing.

Advantages:

1. Much simpler implementation.
2. Automatic `constexpr`, `noexcept`, etc...
3. Better run-time performance in debug mode (no function call penalty for constructors and binding references).
4. Promotes more concise aggregate initialization.
5. Designated initializer support (standard in C++20, supported pretty much everywhere as an extension prior to C++20).

Disadvantages:

1. We lose the `(position, texCoords)` constructor, however I think it's pretty rare and we can achieve the same thing with designated initializers.
2. `emplace_back` and similar emplacement functions will not work prior to C++20, i.e.

    ```cpp
    std::vector<sf::Vertex> v;
    v.push_back({0.0f, 0.0f});    //    OK in C++17, OK in C++20
    v.emplace_back({0.0f, 0.0f}); //    OK in C++17, OK in C++20
    v.emplace_back(0.0f, 0.0f);   // ERROR in C++17, OK in C++20
    ``` 

    However this is also not a big deal as most people use `sf::VertexArray` which does not support emplacement in the first place.
    
I think it's a worthwhile change for SFML 3.x. Thoughts?


vittorioromeo:

ChrisThrasher:
I support this refactor. I don't think a type with entirely public data and zero invariants needs to be more complicated than an aggregate type. We've discussed this a few times in the past and couldn't get consensus, so I'm curious what the team has to think of it this time around.

eXpl0it3r:
> We lose the `(position, texCoords)` constructor, however I think it's pretty rare and we can achieve the same thing with designated initializers.

Isn't this more common than just position & color? Like when you want a textured vertex array, you wouldn't really explicitly set a color, which seems like a more common mode, than just a colored vertex array.

Would it make sense to reverse the order?

vittorioromeo:
> > We lose the `(position, texCoords)` constructor, however I think it's pretty rare and we can achieve the same thing with designated initializers.
> 
> Isn't this more common than just position & color? Like when you want a textured vertex array, you wouldn't really explicitly set a color, which seems like a more common mode, than just a colored vertex array.
> 
> Would it make sense to reverse the order?

I am not sure, do we have any sort of data to make this decision? If not, I'd rather keep it safe and keep the order as it is

eXpl0it3r:
> I am not sure, do we have any sort of data to make this decision? If not, I'd rather keep it safe and keep the order as it is

What makes this "safe" / safer compared to the other order?

vittorioromeo:
> > I am not sure, do we have any sort of data to make this decision? If not, I'd rather keep it safe and keep the order as it is
> 
> What makes this "safe" / safer compared to the other order?

Honestly, just paranoid about changing the memory layout of such a fundamental type. Maybe there is code out there that relies on the order of members...

Bromeon:
> Honestly, just paranoid about changing the memory layout of such a fundamental type. Maybe there is code out there that relies on the order of members...

Access could run into UB, but most code would probably need to construct `sf::Vertex` objects first, thus get compile errors, anyway?

If we change the order, is there any syntax that could previously be used for construction and would _still_ compile now? I think not 🤔

texus:
My code would break if the layout is changed, but I'm very aware that I'm doing something that isn't supported. So I have no problem updating my code if you decide to change the order.

I'm currently doing a `reinterpret_cast` from a list of my own vertex class to a list of sf::Vertex. I had 2 reasons for doing so:
- When I measured my drawing performance in the past, I noticed that constructing sf::Vertex objects had an impact on performance in some cases because the constructors weren't inlined. Creating my own object and then casting to sf::Vertex was actually faster. This reason will no longer exist after this PR, because the constructors are being removed.
- The [SDL_Vertex struct](https://github.com/libsdl-org/SDL/blob/451dc414278dc7b19a997ef7a6d1acf1f951162d/include/SDL3/SDL_render.h#L91-L96) happens to have the exact same memory layout as sf::Vertex. I need my own vertex class in backend-independent parts of my code, but because all vertex classes have identical memory layouts, I was able to just `reinterpret_cast` between them instead of having to copy all vertices to the backend-specific type when drawing.

eXpl0it3r:
> @eXpl0it3r: do you still strongly feel like the order of the members should be changed?

No, I think it's okay as is

include/SFML/Graphics/Vertex.hpp
```diff
@@ -37,73 +37,24 @@ namespace sf
...
-    /// \param theTexCoords Vertex texture coordinates
-    ///
-    ////////////////////////////////////////////////////////////
-    constexpr Vertex(const Vector2f& thePosition, const Color& theColor, const Vector2f& theTexCoords);
-
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    Vector2f position;            //!< 2D position of the vertex
+    Vector2f position{};          //!< 2D position of the vertex
```
ChrisThrasher, line 48:
```suggestion
    Vector2f position;          //!< 2D position of the vertex
```
The default constructor already does what we want.

vittorioromeo:

test/Graphics/Vertex.test.cpp
```diff
@@ -13,45 +13,48 @@ TEST_CASE("[Graphics] sf::Vertex")
...
-            constexpr sf::Vertex vertex({1, 2}, {3, 4, 5, 6});
+            constexpr sf::Vertex vertex{{1.0f, 2.0f}, {3, 4, 5, 6}};
             STATIC_CHECK(vertex.position == sf::Vector2f(1.0f, 2.0f));
             STATIC_CHECK(vertex.color == sf::Color(3, 4, 5, 6));
             STATIC_CHECK(vertex.texCoords == sf::Vector2f(0.0f, 0.0f));
         }
 
-        SECTION("Position and coords constructor")
+#if __cplusplus >= 202002L
+        SECTION("Aggregate initialization -- Position and coords")
```
ChrisThrasher, line 46: I don’t believe any CI job is compiling this. I’d rather we not write conditionally compiled C++20 code. It’s sufficient to assert that this is an aggregate type then we can be certain that designated initializers will work.

ChrisThrasher (commented):
Ready to approve pending some minor comments

ChrisThrasher:
Rebased on `master`, fix two of my review comments, and silenced a warning that I don't think we care to use. Not sure why this warning appears in my local macOS build but not in CI. Perhaps it's because CI still uses an older macOS version with AppleClang 14 instead of 15.

ChrisThrasher (approved):

include/SFML/Graphics/Vertex.hpp
ChrisThrasher: I stand corrected! Removing `{}` here triggers `-Wmissing-field-initializers` in Clang and GCC. Much easier to keep these braces here since being explicit seems to make compilers happier.

ChrisThrasher:

ChrisThrasher:

vittorioromeo:
I'm fine to merge as-is -- @eXpl0it3r are you good with it now?

eXpl0it3r (approved):

PR #2908, opened 2024-02-15
Add support for macOS arm64 builds in CI

eXpl0it3r:
## Description

The GitHub Actions runner image support for `macos-13-xlarge` and `macos-14` (beta) arm64 builds.
This PR updates all the macOS runner images to `macos-14`
Afaik we can't use the `macos-13-xlarge` ones, not sure if we want to wait for `macos-14` to get out of beta.

See also: https://github.com/actions/runner-images

## How to test this PR?

Have CI be green

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,8 +39,9 @@ jobs:
         - { name: Linux Clang,                    os: ubuntu-22.04, flags: -DCMAKE_CXX_COMPILER=clang++ -DSFML_RUN_DISPLAY_TESTS=ON -GNinja , gcovr_options: '--gcov-executable="llvm-cov-$CLANG_VERSION gcov"' }
         - { name: Linux GCC DRM,                  os: ubuntu-22.04, flags: -DSFML_USE_DRM=ON -DSFML_RUN_DISPLAY_TESTS=OFF -GNinja }
         - { name: Linux GCC OpenGL ES,            os: ubuntu-22.04, flags: -DSFML_OPENGL_ES=ON -DSFML_RUN_DISPLAY_TESTS=OFF -GNinja }
-        - { name: macOS,                          os: macos-12, flags: -GNinja }
-        - { name: macOS Xcode,                    os: macos-12, flags: -GXcode }
+        - { name: macOS x64,                      os: macos-12, flags: -GNinja }
+        - { name: macOS x64 Xcode,                os: macos-12, flags: -GXcode }
+        - { name: macOS arm64,                    os: macos-14, flags: -GNinja }
         - { name: iOS,                            os: macos-12, flags: -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_ARCHITECTURES=arm64 }
         - { name: iOS Xcode,                      os: macos-12, flags: -DCMAKE_SYSTEM_NAME=iOS -GXcode -DCMAKE_XCODE_ATTRIBUTE_CODE_SIGNING_ALLOWED=NO }
         config:


.github/workflows/ci.yml
```diff
@@ -39,10 +39,12 @@ jobs:
         - { name: Linux Clang,                    os: ubuntu-22.04, flags: -DCMAKE_CXX_COMPILER=clang++ -DSFML_RUN_DISPLAY_TESTS=ON -GNinja , gcovr_options: '--gcov-executable="llvm-cov-$CLANG_VERSION gcov"' }
         - { name: Linux GCC DRM,                  os: ubuntu-22.04, flags: -DSFML_USE_DRM=ON -DSFML_RUN_DISPLAY_TESTS=OFF -GNinja }
         - { name: Linux GCC OpenGL ES,            os: ubuntu-22.04, flags: -DSFML_OPENGL_ES=ON -DSFML_RUN_DISPLAY_TESTS=ON -GNinja }
-        - { name: macOS,                          os: macos-12, flags: -GNinja }
-        - { name: macOS Xcode,                    os: macos-12, flags: -GXcode }
```
ChrisThrasher, line 43: It's already questionable whether having a macOS Ninja job and macOS Xcode job is useful. I think for these ARM builds we don't need to duplicate them for Xcode. That'll save us some CI time.

.github/workflows/ci.yml
```diff
@@ -39,10 +39,12 @@ jobs:
...
-        - { name: iOS,                            os: macos-12, flags: -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_ARCHITECTURES=arm64 }
-        - { name: iOS Xcode,                      os: macos-12, flags: -DCMAKE_SYSTEM_NAME=iOS -GXcode -DCMAKE_XCODE_ATTRIBUTE_CODE_SIGNING_ALLOWED=NO }
+        - { name: macOS x64,                      os: macos-14, flags: -GNinja -DCMAKE_OSX_ARCHITECTURES=x86_64 }
```
ChrisThrasher, line 42: If all macOS 14 images are using ARM then I'd rather we just use the macOS 14 jobs to test ARM and macOS 13 jobs to test x86. That further deduplicates things and lets us skip setting CMAKE_OSX_ARCHITECTURES since we'll always use the native architecture.

eXpl0it3r:
Since it can't find Vorbis as system lib on `macos-14` does that mean, we'd have to install it manually?

ChrisThrasher (changes requested):

ChrisThrasher:
> Since it can't find Vorbis as system lib on `macos-14` does that mean, we'd have to install it manually?

Shouldn't it be finding what we have in extlibs? That's what gets found when I configure the library.

.github/workflows/ci.yml
eXpl0it3r: Eventually `macos-13` will go away, but we can also stick to it for as long as it exists. I kind of like that flag, as it makes it explicit what's going on, rather than relying on hidden knowledge that `macos-13` is x64 based and `macos-14` is arm64 based.

ChrisThrasher: We can use the CI job name to make it clear what's going on. I find that more expressive than adding unnecessary parameters.

eXpl0it3r:
> Shouldn't it be finding what we have in extlibs? That's what gets found when I configure the library.

I'm referring to the failing `macOS System Deps` builds, which should find the libraries on the system and not the ones SFML brings along.

.github/workflows/ci.yml
ChrisThrasher: It's hard to write CMake code that works with one generator on a given OS but not another generator.

> not that CI time is really a useful or strong argument here.

What do you mean? We've had _many_ PRs spent optimizing CI runtime. The fewer jobs we have the faster CI completes.

ChrisThrasher:
> > Shouldn't it be finding what we have in extlibs? That's what gets found when I configure the library.
> 
> I'm referring to the failing `macOS System Deps` builds, which should find the libraries on the system and not the ones SFML brings along.

Oh my bad. In that case I'd try installing libvorbis from Homebrew to see if that works.

ChrisThrasher:
https://github.com/SFML/SFML/compare/master...arm_mac_builds

I propose we do something simple like this.

eXpl0it3r:
> [master...arm_mac_builds](https://github.com/SFML/SFML/compare/master...arm_mac_builds)
> 
> I propose we do something simple like this.

Why remain on `macos-12` with the rest?

ChrisThrasher:
> > [master...arm_mac_builds](https://github.com/SFML/SFML/compare/master...arm_mac_builds)
> > I propose we do something simple like this.
> 
> Why remain on `macos-12` with the rest?

It's worth updating those but it's not strictly necessary if we just want to add an ARM job. I'm fine adding the one line from my arm_mac_builds and also bumping all macos-12 images to macos-13 while we're at it.

ChrisThrasher (approved):

PR #2909, opened 2024-02-15
Fix CI deprecation warnings

ChrisThrasher:
## Description

Who knew checking out a repo was so complicated but we're getting loads of deprecation warnings for using actions/checkout@v3 so I upgraded us to v4. 

Screenshot 2024-02-15 at 4 57 47 PM


eXpl0it3r (approved):
I always thought the checkout was just a git clone, but I recently learned that it actually fetches the data through the GitHub API 😄

PR #2910, opened 2024-02-19
Fix assert not firing when only x index is out of bounds

dogunbound:
## Description

Original protection to UB here was to simply check if the produced index is out of bounds, but that isn't entirely correct imo. What is correct is to fire if either the x or the y go out of bounds.

This shouldn't cause the UB problem to reappear, and be a little more "correct" from the user's perspective. I believe that the most correct way would be to use a boolean return in the setPixel(), and a std::optional and return None on a failed fetch for the getPixel() function. I don't know if ya'll are willing to do this, but this seems sufficient too. Atleast there is a clear place where the programs fails to work.

## Tasks

-   [x] Tested on Linux
-   [ ] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

```cpp
#include "SFML/Graphics/Image.hpp"
#include <SFML/Graphics.hpp>
#include <stdio.h>

const static sf::Vector2u IMAGE_SIZE = sf::Vector2u(64, 64);

int main()
{
    sf::Image black_image;
    black_image.create(IMAGE_SIZE, sf::Color::Black);

    for (int y = 0; y < 65; y++) {
        for (int x = 0; x < 65 ; x++) {
            sf::Color pixel = black_image.getPixel(sf::Vector2u(x,y));
            printf("[%u %u %u]", pixel.r, pixel.g, pixel.b);
        }
        printf("\n");
    }
    return 1;
}

```

### DIFF
--- a/src/SFML/Graphics/Image.cpp
+++ b/src/SFML/Graphics/Image.cpp
@@ -490,8 +490,10 @@ void Image::createMaskFromColor(const Color& color, std::uint8_t alpha)
 ////////////////////////////////////////////////////////////
 void Image::setPixel(const Vector2u& coords, const Color& color)
 {
-    const auto index = (coords.x + coords.y * m_size.x) * 4;
-    assert(index < m_pixels.size() && "Image::setPixel() cannot access out of bounds pixel");
+    assert(coords.x < m_size.x && "Image::setPixel() x coordinate is out of bounds");
+    assert(coords.y < m_size.y && "Image::setPixel() y coordinate is out of bounds");
+
+    const auto    index = (coords.x + coords.y * m_size.x) * 4;
     std::uint8_t* pixel = &m_pixels[index];
     *pixel++            = color.r;
     *pixel++            = color.g;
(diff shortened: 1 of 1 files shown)


src/SFML/Graphics/Image.cpp
```diff
@@ -490,8 +490,10 @@ void Image::createMaskFromColor(const Color& color, std::uint8_t alpha)
 ////////////////////////////////////////////////////////////
 void Image::setPixel(const Vector2u& coords, const Color& color)
 {
+    assert(coords.x < getSize().x && "Image::setPixel() x coordinate is out of bounds");
```
ChrisThrasher, line 493:
```suggestion
    assert(coords.x < m_size.x && "Image::setPixel() x coordinate is out of bounds");
```

ChrisThrasher (changes requested):
Seems reasonable to me. New assertions are a superset of the old one and catch more undesirable code.

src/SFML/Graphics/Image.cpp
dogunbound: This guy. so picky. lol. It's fixed.

ChrisThrasher (approved):

ChrisThrasher:

PR #2912, opened 2024-02-20
Enable `bugprone-suspicious-include`

ChrisThrasher:
## Description

We already comply with this check so we ought to include it. I'm not sure why we previously had to disable this.


eXpl0it3r (approved):

ChrisThrasher:
image

I just learned that CMake automatically silences this warning when doing unity builds. This is why CI is not failing.

PR #2914, opened 2024-02-26
Use `std::{min|max}`

ChrisThrasher:
## Description

A new clang-tidy-18 check led me to this. These files already use \<algorithm\> so we might as well use these functions where appropriate.

### DIFF
--- a/src/SFML/Graphics/Texture.cpp
+++ b/src/SFML/Graphics/Texture.cpp
@@ -313,14 +313,10 @@ bool Texture::loadFromImage(const Image& image, const IntRect& area)
 
         // Adjust the rectangle to the size of the image
         IntRect rectangle = area;
-        if (rectangle.left < 0)
-            rectangle.left = 0;
-        if (rectangle.top < 0)
-            rectangle.top = 0;
-        if (rectangle.left + rectangle.width > width)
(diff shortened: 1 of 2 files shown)


src/SFML/Graphics/Texture.cpp
```diff
@@ -313,10 +313,8 @@ bool Texture::loadFromImage(const Image& image, const IntRect& area)
 
         // Adjust the rectangle to the size of the image
         IntRect rectangle = area;
-        if (rectangle.left < 0)
-            rectangle.left = 0;
-        if (rectangle.top < 0)
-            rectangle.top = 0;
+        rectangle.left    = std::max(rectangle.left, 0);
+        rectangle.top     = std::max(rectangle.top, 0);
         if (rectangle.left + rectangle.width > width)
```
kimci86, line 318: While at it, let's use `std::min` here too.
```c++
rectangle.width = std::min(rectangle.width, width - rectangle.left);
rectangle.height = std::min(rectangle.height, height - rectangle.top);
```

ChrisThrasher: Thanks!

kimci86 (approved):

eXpl0it3r (approved):

PR #2917, opened 2024-03-10
Fix pkgconfig files install path

texus:
## Description

When running CMake without setting CMAKE_INSTALL_PREFIX, and then running CMake again with a different prefix, the install path for the pkgconfig files still points to the old location.

Because the default `/usr/local` prefix on Linux requires root access, installing SFML without sudo will fail with an error like this:
```
-- Install configuration: "Release"
-- Installing: /home/texus/Documents/SFML/whatever/lib/libsfml-system-s.a
-- Installing: /home/texus/Documents/SFML/whatever/lib/libsfml-window-s.a
-- Installing: /home/texus/Documents/SFML/whatever/lib/libsfml-network-s.a
-- Installing: /home/texus/Documents/SFML/whatever/lib/libsfml-graphics-s.a
-- Installing: /home/texus/Documents/SFML/whatever/lib/libsfml-audio-s.a
-- Installing: /usr/local/lib/pkgconfig/sfml-all.pc
CMake Error at build/cmake_install.cmake:59 (file):
  file INSTALL cannot copy file
  "/home/texus/Documents/SFML/build/tools/pkg-config/sfml-all.pc" to
  "/usr/local/lib/pkgconfig/sfml-all.pc": Permission denied.
```

This was fixed in the CSFML project a few weeks ago (see https://github.com/SFML/CSFML/issues/233 and https://github.com/SFML/CSFML/pull/234), this PR applies the same fix to SFML.

This change is NOT backwards compatible. The existing `SFML_PKGCONFIG_INSTALL_PREFIX` variable will no longer work and is being replaced with a new `SFML_PKGCONFIG_INSTALL_DIR` variable in cmake.

## Tasks

-   [x] Tested on Linux

## How to test this PR?

On linux you can run the following commands to get the error:
```bash
cmake -S . -B build
cmake --build build
cmake --install build --prefix whatever
```

On other platforms you can probably also witness the files being installed to the wrong directory, but you will need to set `SFML_INSTALL_PKGCONFIG_FILES` to `ON` yourself because it is only enabled by default on Linux and BSD.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -213,13 +213,13 @@ if(SFML_INSTALL_PKGCONFIG_FILES)
 
     # set pkgconfig install directory
     # this could be e.g. macports on mac or msys2 on windows etc.
-    set(SFML_PKGCONFIG_DIR "/${SFML_RELATIVE_INSTALL_LIBDIR}/pkgconfig")
+    set(SFML_PKGCONFIG_DIR "${SFML_RELATIVE_INSTALL_LIBDIR}/pkgconfig")
 
     if(SFML_OS_FREEBSD OR SFML_OS_OPENBSD OR SFML_OS_NETBSD)
-        set(SFML_PKGCONFIG_DIR "/libdata/pkgconfig")
+        set(SFML_PKGCONFIG_DIR "libdata/pkgconfig")
     endif()
 
-    sfml_set_option(SFML_PKGCONFIG_INSTALL_PREFIX "${CMAKE_INSTALL_PREFIX}/${SFML_PKGCONFIG_DIR}" PATH "Install directory for SFML's pkg-config .pc files")
+    sfml_set_option(SFML_PKGCONFIG_INSTALL_DIR "${SFML_PKGCONFIG_DIR}" PATH "Install directory for SFML's pkg-config .pc files")
 
     foreach(sfml_module IN ITEMS all system window graphics audio network)
         configure_file(
@@ -227,7 +227,7 @@ if(SFML_INSTALL_PKGCONFIG_FILES)
             "tools/pkg-config/sfml-${sfml_module}.pc"
             @ONLY)
         install(FILES "${CMAKE_CURRENT_BINARY_DIR}/tools/pkg-config/sfml-${sfml_module}.pc"
-            DESTINATION "${SFML_PKGCONFIG_INSTALL_PREFIX}")
+            DESTINATION "${SFML_PKGCONFIG_INSTALL_DIR}")
     endforeach()
 endif()
 


eXpl0it3r:

eXpl0it3r:

eXpl0it3r:
@ChrisThrasher has since implemented this in #2835 

Does that also work for you @texus?

texus:
The issue still occurs.
PR https://github.com/SFML/SFML/pull/2835 appears to be about different problem.

I've rebased my PR on the master branch.

texus:
I'm starting to think that we don't actually need SFML_RELATIVE_INSTALL_LIBDIR anymore when SFML_PKGCONFIG_INSTALL_PREFIX is replaced with SFML_PKGCONFIG_INSTALL_DIR. The old issue was about appending CMAKE_INSTALL_FULL_LIBDIR to CMAKE_INSTALL_PREFIX failing when SFML_PKGCONFIG_INSTALL_DIR was an absolute path. However this PR removes the CMAKE_INSTALL_PREFIX variable from this code.

I'm not familiar enough with pkg-config files to actually check whether the generated files are correct though.

ChrisThrasher:
> I'm waiting until there is a consensus on https://github.com/SFML/CSFML/pull/374 to keep the SFML implementation the same as in CSFML.

SFML sets the precedent that CSFML follows so this PR does not need to concern itself too much with being consistent with CSFML. 

> I'm starting to think that we don't actually need SFML_RELATIVE_INSTALL_LIBDIR anymore when SFML_PKGCONFIG_INSTALL_PREFIX is replaced with SFML_PKGCONFIG_INSTALL_DIR.

Do you still think this may be true? If we can fix this bug while also getting rid of some extra intermediate variables then I'd prefer that.

eXpl0it3r:

texus:
> SFML sets the precedent that CSFML follows so this PR does not need to concern itself too much with being consistent with CSFML.

The same changes are needed in both libraries though, so the exact same discussion in CSFML can be had right here for SFML. Which is why I still think that the CSFML discussion is relevant.

> > I'm starting to think that we don't actually need SFML_RELATIVE_INSTALL_LIBDIR anymore when SFML_PKGCONFIG_INSTALL_PREFIX is replaced with SFML_PKGCONFIG_INSTALL_DIR.
> 
> Do you still think this may be true? If we can fix this bug while also getting rid of some extra intermediate variables then I'd prefer that.

It still has a use in the existing code, but it can be removed depending on which decision is taken in the `@CMAKE_INSTALL_FULL_LIBDIR@` vs `${exec_prefix}/@CSFML_RELATIVE_INSTALL_LIBDIR@` situation.

- The advantage of `@CMAKE_INSTALL_FULL_LIBDIR@` is requiring less code, the `SFML_RELATIVE_INSTALL_LIBDIR` variable can be removed.
- The advantage of `${exec_prefix}/@CSFML_RELATIVE_INSTALL_LIBDIR@` is that it is more consistent with other pkg-config files. From the pkg-config files installed on my linux, almost 90% use some kind of prefix for their `libdir` value where only 10% use a hardcoded path (which is what `@CMAKE_INSTALL_FULL_LIBDIR@` would do). It also makes the .pc file somewhat more relocatable, but with `prefix` being a hardcoded value the files have never been relocatable anyway (so I don't think it is that important).

eXpl0it3r:

eXpl0it3r (approved):
I believe we should go with this for SFML 3. We can still have a discussion on the topic of variables.

PR #2931, opened 2024-04-09
Spend longer attempting to get codecov upload to succeed

ChrisThrasher:
## Description

If this fails then we're on the hook to manually retrigger CI. I'd rather just spend more time waiting on codecov to work because most times it fails it succeeds as soon as we re-trigger. GitHub Actions won't let you restart a failed CI job until all jobs have succeeded which further wastes time when we're waiting on a few codecov uploads to succeed.

25 attempts at 2 minutes between attempt mean we're allowing for about an hour for codecov to work. That should be long enough to deal with intermittant network failures or rate limiting in most cases.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -222,8 +222,8 @@ jobs:
       uses: nick-fields/retry@v3
       with:
         timeout_seconds: 30
-        max_attempts: 10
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
Alternatively we could try the v4 of the GitHub Action with our own token: #2932

PR #2932, opened 2024-04-09
Switch back to Codecov GitHub Action with our own token

eXpl0it3r:
## Description

Alternative approach to #2931 

As the coverage upload keeps failing even with way more retries, it might make sense to switch back to a token based approach, which is also the only supported way with the v4 GitHub action: https://github.com/codecov/codecov-action

We originally went with our own scripts so we could use the retry-action, as we were experiencing many failures. Question is, whether v4 performs better than v3...

## How to test this PR?

All CI runs with coverage upload are successful.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,7 +209,7 @@ jobs:
       uses: codecov/codecov-action@v4
       if: matrix.type.name == 'Debug' && github.repository == 'SFML/SFML' # Disable upload in forks
       with:
-        fail_ci_if_error: true
+        fail_ci_if_error: false
         name: ${{ matrix.platform.name }} ${{ matrix.config.name }} ${{ matrix.type.name }}
         directory: ./build
         files: ./build/coverage.out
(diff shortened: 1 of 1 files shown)


binary1248 (approved):

ChrisThrasher (approved):
This looks like a much cleaner solution. Fingers crossed this isn't flakey but we won't know until we merge so let's go ahead and do that.

ChrisThrasher:
Screenshot 2024-04-09 at 11 02 24 AM

It is a little concerning however that it's showing a change in coverage. I wonder if some reports are actually not getting uploaded properly or something like that.

eXpl0it3r:
I've opened an feedback issue regarding the "random" rate limiting we're getting from Codecov: https://github.com/codecov/feedback/issues/322

For now, I've marked the report upload as not required to pass the check. This however means that in certain cases the number might not be reported correctly.  
This seems to be like a better option, than to just fail over and over or do multi-hour retries in hopes to not get hit by rate limiting. If need be, we should be able to re-run jobs on master, if it fails there and gives us a wrong number.

Thoughts?

Maybe in the long run we have to move to some other service like [coveralls](https://coveralls.io/)

PR #2938, opened 2024-04-11
Fixed coverage data not being collected for MinGW builds

binary1248:
OpenCPPCoverage relies on pdb data being available. These are generated by MSVC and recently LLVM on Windows as well. When building with GCC on MinGW it will use GCC's native debugging information which is incompatible with OpenCPPCoverage.

This change just makes use of Gcovr to collect the GCC profiling information on MinGW just like is already done for GCC on Linux.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,6 +128,12 @@ jobs:
         verbose: 2
         key: ${{ matrix.platform.name }}-${{ matrix.config.name }}-${{ matrix.type.name }}
 
+    - name: Install Gcovr for MinGW
+      if: matrix.type.name == 'Debug' && contains(matrix.platform.name, 'MinGW')
+      uses: threeal/pipx-install-action@v1.0.0
+      with:
+        packages: gcovr
+
     - name: Cache OpenCppCoverage
       if: matrix.type.name == 'Debug' && runner.os == 'Windows'
       id: opencppcoverage-cache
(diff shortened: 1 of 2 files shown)


ChrisThrasher:
What would a MinGW coverage report include that an MSVC coverage report wouldn't already have? We don't need 50 coverage reports. We only need 1 unique report for each unique set of source code. I think we need maybe as few as like 6 reports if we were really careful about picking the right ones.

binary1248:
GCC is fundamentally a different compiler from MSVC or Clang and the only way to get GCC running on Windows is either using MinGW or Cygwin. It has happened before that tests have revealed problems when building/running GCC code on Windows in the past. In order for us to verify our tests help to prevent such problems in the future we need to know if they are actually covering the relevant parts of our code that might lead to issues.

ChrisThrasher:
> In order for us to verify our tests help to prevent such problems in the future we need to know if they are actually covering the relevant parts of our code that might lead to issues.

We already know this because we can see MSVC overage reports. That tells us what code is being covered on Windows. It's implied that code that is called by MSVC is also called by MinGW.

binary1248:
Well... see that's the thing, you can't make that assumption. Fundamentally different compilers will have different optimizers that among other things inline, fold and elide code differently. You can take a sufficiently complex example throw it into compiler explorer add see that the output isn't 1:1 comparable.

ChrisThrasher:
Optimizations are not relevant since we only measure coverage in debug builds.

ChrisThrasher:
Screenshot 2024-04-11 at 4 48 18 PM

I'm trying to figure out why this PR results in coverage changing. Coveralls says that the above 5 files are where coverage changed and none of them are Windows files.

binary1248:
The standard library implementations will also lead to different codegen because they have optimized paths depending on type traits among other things. Even with all optimizations disabled, what e.g. standard containers do when you modify them, whether they move or copy our types is all implementation dependant. If MinGW were to use the Microsoft standard library then sure, you could make the argument that the codegen should look very similar if not identical, but this is not the case. This is all ignoring the fact that the linker doesn't have flags that disable optimizations. If you look in `Macros.cmake` we already have to instruct the MSVC linker to build the tests with `/WHOLEARCHIVE` in order to get usable coverage output when linking statically. Who's to say that the GNU linker will not uncover lines that were never present in the MSVC/Clang coverage reports?

ChrisThrasher:
> you could make the argument that the codegen should look very similar if not identical

I don't think codegen is relevant to this discussion. Code coverage is about which lines of C++ are ran, not the machine code that is generated by the compiler backend.

> Who's to say that the GNU linker will not uncover lines that were never present in the MSVC/Clang coverage reports?

Who's to say the inverse of this is not true? So far we still have no evidence that MinGW is going to cover lines that MSVC does not. I'm very open to changing my opinion if we find any evidence that suggests that compiling with MinGW results in different lines of SFML's source code being ran than when compiling with MSVC. Without said evidence I'm not compelled to believe that is true.

ChrisThrasher (approved):
I'm still under the impression that these new reports do not add any value but they also come at very small cost so I'm fine merging this. The new code coverage reports are not going to hurt anything. At worst we just spend a few extra seconds of CI time creating redundant reports. Perhaps sometime in the future these new reports will demonstrate their value.

PR #2942, opened 2024-04-13
Remove `sf::View::reset` in favor of assignment operations

ChrisThrasher:
## Description
It's rare that a type truly needs a .reset function. Copy/move assignment typically accomplishes the same thing with less code and is easier to maintain since it doesn't require updating your .reset() function as new data members are added.

To reset a type is conceptually the same thing as simply assigning from a newly constructed instance of the same type.

### DIFF
--- a/include/SFML/Graphics/View.hpp
+++ b/include/SFML/Graphics/View.hpp
@@ -144,18 +144,6 @@ public:
     void setScissor(const FloatRect& scissor);
 
     ////////////////////////////////////////////////////////////
-    /// \brief Reset the view to the given rectangle
-    ///
-    /// Note that this function resets the rotation angle to 0.
-    ///
-    /// \param rectangle Rectangle defining the zone to display
-    ///
-    /// \see setCenter, setSize, setRotation
-    ///
-    ////////////////////////////////////////////////////////////
-    void reset(const FloatRect& rectangle);
-
-    ////////////////////////////////////////////////////////////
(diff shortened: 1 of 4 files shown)


ChrisThrasher:

kimci86:
Worth noting that reset method was not doing the same as assigning a new constructed view because it was not changing viewport and scissor.

ChrisThrasher:
> Worth noting that reset method was not doing the same as assigning a new constructed view because it was not changing viewport and scissor.

Good point. I wonder if that is an oversight or intentional. What could be the use case for resetting part of a view but not all of it?

ChrisThrasher:

ChrisThrasher:
> Worth noting that reset method was not doing the same as assigning a new constructed view because it was not changing viewport and scissor.

I did some digging in the commit history.

`m_viewport` was added in [`7cc00085`](https://github.com/SFML/SFML/commit/7cc00085) which came after the addition of the `reset` function. It's feasible it was an oversight that `reset` was not amended to include this new data member.

A similar story is feasible for `m_scissor` which was added much more recently in https://github.com/SFML/SFML/pull/1451. It's so easy to forget to update this kind of reset function when adding more data members.

I think `.reset()` was never intended to skip those two data members.

ChrisThrasher:

ChrisThrasher:

vittorioromeo (approved):

PR #2944, opened 2024-04-13
Compile unit tests for iOS and Android

ChrisThrasher:
## Description

While you can't easily run these unit tests, it's easy enough to still compile them so we don't need to go out of way to prohibit this. CMake has some support for running tests on the target OS (see [`CROSSCOMPILING_EMULATOR`](https://cmake.org/cmake/help/latest/prop_tgt/CROSSCOMPILING_EMULATOR.html)) so perhaps in the future we find a compelling way to run the test suite on a mobile OS.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,16 +65,16 @@ jobs:
         - platform: { name: macOS, os: macos-12 }
           config: { name: Frameworks, flags: -GNinja -DSFML_BUILD_FRAMEWORKS=TRUE -DBUILD_SHARED_LIBS=TRUE }
         - platform: { name: Android, os: ubuntu-22.04 }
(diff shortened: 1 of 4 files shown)


test/TestUtilities/SystemUtil.cpp
```diff
@@ -87,7 +87,8 @@ std::vector<std::byte> loadIntoMemory(const std::filesystem::path& path)
     const auto size = file.tellg();
     file.seekg(0, std::ios::beg);
     std::vector<std::byte>       buffer(static_cast<std::size_t>(size));
-    [[maybe_unused]] const auto& result = file.read(reinterpret_cast<char*>(buffer.data()), size);
+    [[maybe_unused]] const auto& result = file.read(reinterpret_cast<char*>(buffer.data()),
+                                                    static_cast<std::streamsize>(size));
```
ChrisThrasher, line 91: When compiling on Android there was a new conversion warning due to platform-specific difference in the definition of types like `std::size_t` and `std::streamsize`.

.github/workflows/ci.yml
```diff
@@ -211,7 +211,7 @@ jobs:
         fi
 
     - name: Upload Coverage Report to Coveralls
-      if: matrix.type.name == 'Debug' && github.repository == 'SFML/SFML' # Disable upload in forks
```
ChrisThrasher, line 214: This is the only part of this PR that annoys me. Feels a bit ugly to add even more conditions to CI but I don't see a better alternative.

binary1248 (approved):

PR #2945, opened 2024-04-14
Run Android tests using GitHub actions

binary1248:
Title.

Runs Android tests on x86, x86_64 and arm64-v8a. armeabi-v7a is already quite old and there isn't emulator support for it.

To run the tests, an Android Virtual Device (AVD) is created with a default Android system image using the specified emulator API and emulator architecture. Not all combinations are available, the emulator API and architectures in this change are the latest I could find that would still be able to run the tests.

Because we only dynamic link on Android, we also need to copy all the executables and dynamic libraries (including libc++) over to the device as well as the resources that the tests themselves need to pass. The directory containing the libraries is specified using `LD_LIBRARY_PATH` when running the executable. We can only copy files into a subdirectory of `/data/local/tmp/` on the device since everything else is read-only.

Because Catch doesn't currently support a nice way of running tests on a foreign system, we have to make use of the `CROSSCOMPILING_EMULATOR` target property that Catch uses when building the command line that it executes. Instead of directly executing the test executable, it calls a specially crafted proxy shell script that takes care of setting the working directory on the device, setting `LD_LIBRARY_PATH` and adding the `/data/local/tmp/` directory prefix onto the absolute path to the executable that Catch passes to us. Hopefully Catch supports such use cases better in the future.

Coverage data collection is missing because, well... it doesn't seem to be supported, or at least it wasn't obvious to me how to activate generation of profiling data when the executable is run.

### DIFF
--- a/src/SFML/Graphics/Image.cpp
+++ b/src/SFML/Graphics/Image.cpp
@@ -31,6 +31,7 @@
 #include <SFML/System/InputStream.hpp>
 #include <SFML/System/Utils.hpp>
 #ifdef SFML_SYSTEM_ANDROID
+#include <SFML/System/Android/Activity.hpp>
 #include <SFML/System/Android/ResourceStream.hpp>
 #endif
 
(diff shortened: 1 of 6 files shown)


src/SFML/Graphics/Image.cpp
```diff
@@ -157,7 +158,15 @@ void Image::create(const Vector2u& size, const std::uint8_t* pixels)
 ////////////////////////////////////////////////////////////
 bool Image::loadFromFile(const std::filesystem::path& filename)
 {
-#ifndef SFML_SYSTEM_ANDROID
+#ifdef SFML_SYSTEM_ANDROID
+
+    if (priv::getActivityStatesPtr() != nullptr)
```
ChrisThrasher, line 163: What does it mean if this condition evaluates false? I'm not familiar with Android activity states.

binary1248: If there is an activity state, we are running as an activity, meaning the process was started using the activity manager. If we open up a console or just use adb to run a binary it will run natively without the activity manager, thus there will be no activity state in that case.

ChrisThrasher (approved):

Issue #2943, opened 2024-04-13
(Linux) Window Close Event not triggered with sf::Style None, Titlebar, Resize


danieljpetersen:
### Prerequisite Checklist

- [X] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [X] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [X] I am here to report an issue and not to just ask a question or look for help (use the [forum](https://en.sfml-dev.org/forums/index.php#c3) or [Discord](https://discord.gg/nr4X7Fh) instead)

### Describe your issue here

The window event sf::Event::Closed never get triggered when creating a window with sf::Style::None, sf::Style::Titlebar, sf::Style::Resize.

Most distros come with a hotkey to close the in-focus window. These hotkeys should allow you to trigger a window close event, but with SFML one never occurs. I tested both GLFW and SDL2 for the same behavior, they both behave as one would expect (the close event is propagated / the window closes).

I initially thought it was styles without a close button on the decorations, but sf::Style::Fullscreen works.

### Your Environment

**- OS:** I tested against Linux Mint 23.1 Cinnamon x11. KDE was also tested by Flareon on Discord. I didn't test against Wayland.

**- SFML version:** Latest master, https://github.com/SFML/SFML/commit/30fcb3523c9d685f5c35eeae829da91daf6ff7ce

**- Compiler / toolchain:** GCC 13.1

### Steps to reproduce

Make sure hotkey is bound which closes in-focus window, compile / run this code, press hotkey -- the window will not get sf::Event::Close, the window will never close.

```cpp
#include <SFML/Window.hpp>
 
int main()
{
    sf::Window window; // also happens with RenderWindow
 
    // The following are all not working: Titlebar, Resize, None
    window.create(sf::VideoMode({800, 600}), "My window", sf::Style::None);
 
    while (window.isOpen()) {
        sf::Event event{};
        while (window.pollEvent(event)) {
            // sf::Event::Closed is never given to us from window.pollEvent(event)
            if (event.type == sf::Event::Closed)
                window.close();
        }
 
        window.display();
    }
 
    return 0;
}
```

### Expected behavior

The event sf::Close should be given via pollEvent when the user attempts to quit (such as by pressing hotkeys).

### Actual behavior

The event sf::Close is never triggered

[...]

PR #2947, opened 2024-04-14
Fix bug on x11 where window close events were not propagated with certain window styles

danieljpetersen:
## Description

This is intended to fix https://github.com/SFML/SFML/issues/2943.

It's a single line change which adds a flag for x11 telling it to propagate close events for styles None, Resize, Titlebar.

In truth I do not understand x11 very well, but the behavior / fix seems to make sense to me at least. It fixes the bug on my machine, and doesn't visually impact the window decorations.

I've only confirmed the initial bug on Cinnamon. Flareon from the Discord confirmed it was present on KDE. 

I've only tested this PR Cinnamon.

## How to test this PR?
Make sure hotkey is bound which closes in-focus window, compile / run this code, press hotkey 
A) without this PR, the window will not get sf::Event::Close, the window will never close.
B) with this PR, the window will get sf::Event::Close, the window will successfully close.

```cpp
#include <SFML/Window.hpp>
 
int main()
{
    sf::Window window; // also happens with RenderWindow
 
    // The following are all not working: Titlebar, Resize, None
    window.create(sf::VideoMode({800, 600}), "My window", sf::Style::None);
 
    while (window.isOpen()) {
        sf::Event event{};
            // sf::Event::Closed is never given to us from window.pollEvent(event)
                window.close();
        }
 
        window.display();
    }
 
    return 0;
}
```

### DIFF
--- a/src/SFML/Window/Unix/WindowImplX11.cpp
+++ b/src/SFML/Window/Unix/WindowImplX11.cpp
@@ -574,7 +574,7 @@ m_cursorGrabbed(m_fullscreen)
             auto hints        = WMHints();
             hints.flags       = MWM_HINTS_FUNCTIONS | MWM_HINTS_DECORATIONS;
             hints.decorations = 0;
-            hints.functions   = 0;
+            hints.functions   = MWM_FUNC_CLOSE;
 
             if (style & Style::Titlebar)
             {


danieljpetersen:
Ah, hmm, this actually alters the the decorations for resize and titlebar. I only tested against none before submitting the PR, should have tested all three cases.

With that in mind then, I could alter the PR so the bug is at least fixed for style::None, but I'm not sure how you'd fix it for the other two.

eXpl0it3r:
> Ah, hmm, this actually alters the the decorations for resize and titlebar. I only tested against none before submitting the PR, should have tested all three cases.

How is it changed?

danieljpetersen:
> > Ah, hmm, this actually alters the the decorations for resize and titlebar. I only tested against none before submitting the PR, should have tested all three cases.
> 
> How is it changed?

It adds the close button to the window decorations for resize and titlebar. sf::Style::None however remains without decorations.

So the PR as it currently stands has regressions for resize and titlebar. We could modify the PR to just target sf::Style::None. This would effectively fix the bug for the sf::Style::None window state with no regressions. sf::Style::Resize and sf::Style::Titlebar would obviously remain broken if we did this (unable to close via hotkey). I do not know if there is a way to actually fix the bug for these two states.

To be as clear as possible,

* sf::Style::None
    * Without current PR:
        * No window decorations
        *  does not close via quit hotkey
    * With current PR: 
        * No window decorations
        * Closes via window hotkey
* sf::Style::Resize
    * Without current PR: 
        * Has window decorations with minimize, maximize
        * Does not close via quit hotkey
    * With current PR: 
        * Has window decorations with minimize, maximize, close
        * Closes via quit hotkey
* sf::Style::Titlebar
    * Without current PR: 
        * Has window decorations with minimize
        * Does not close via quit hotkey
    * With current PR: 
        * Has window decorations with minimize, close
        * Closes via quit hotkey

I'm sure this is unnecessary, but to be abundantly clear here are some images

None with PR as it currently stands:
none_with_fix

None without PR as it currently stands:
none_without_fix

Resize with PR as it currently stands (regression, erroneously adds close):
resize_with_fix

None without PR as it currently stands:
resize_without_fix

Titlebar with PR as it currently stands (regression, erroneously adds close):
titlebar_with_fix

Titlebar without PR as it currently stands:
titlebar_without_fix

eXpl0it3r:
Superseded by #3684

PR #2949, opened 2024-04-16
Revert change to `Drawable::draw` function signature

### LINKED ISSUE #2940: Revert change to `sf::Drawable::draw` signature which made `sf::RenderStates` a const reference, and make it a copy again
### Prerequisite Checklist

I've noticed that the change around the signature of `sf::Drawable::draw` in SFML master for v3 does not improve the users experience when implementing custom drawable types, especially if they inherit from `sf::Transformable`. 

For example, I was producing a custom type to represent a particle emitter. I want to be able to position the emitter and scale it, so I made it inherit from `sf::Transformable`, and to render it I made it inherit from `sf::Drawable`. In 2.x I would have been able to adjust the `sf::RenderStates` parameter to set the transform (see API example below) without making a copy, but when using SFML master the signature has changed to a const reference parameter. This makes the more shorter more appealing form of the method I'd like to write impossible, and forces me to produce a copy. 

I don't think the magnitude of the original change was severe, but there was a useful pattern that emerged from the previous version that I think lots of poweruser with existing projects (using 2.x) would like to retain.

### Use Cases

The usecases for this are that people who inherit from `sf::Drawable` will often want to adjust the value of the `sf::RenderStates` parameter (to apply shaders, textures, transform etc). 

With the current signature a user is forced to make a copy themselves, where previously this was already done by the method signature taking a parameter. 

### API Example

```cpp
// Revert from:
void MyCustomType::draw(sf::RenderTarget& target, const sf::RenderStates& states) const
{
    sf::RenderStates statesCopy = states;
    statesCopy.transform *= getTransform();
    target.draw(m_vertexBuffer, statesCopy);
}

// Back to:
void MyCustomType::draw(sf::RenderTarget& target, sf::RenderStates states) const
{
    states.transform *= getTransform();
    target.draw(m_vertexBuffer, states);
}
```

ChrisThrasher:
## Description

Closes #2940 

This change was made in 359fe90 due to recommendations from tooling. On its face this change makes sense since it removes a copy that isn't always necessary. In practice it caused ergonomic issues due to now being forced to make a copy of the render states when needed.

The performance gains of eliding this copy are unsubstantiated. We have not done any profiling to measure its impact. For lack of such measurements I'd rather err on the side of improved user experience. If future benchmarks prove this copy is rather expensive then we can reconsider removing it with that evidence in mind.

### DIFF
--- a/examples/shader/Effect.hpp
+++ b/examples/shader/Effect.hpp
@@ -38,7 +38,7 @@ public:
             onUpdate(time, x, y);
     }
 
-    void draw(sf::RenderTarget& target, const sf::RenderStates& states) const override
+    void draw(sf::RenderTarget& target, sf::RenderStates states) const override
     {
         if (m_isLoaded)
         {
@@ -66,9 +66,9 @@ protected:
 
 private:
     // Virtual functions to be implemented in derived effects
-    virtual bool onLoad()                                                               = 0;
-    virtual void onUpdate(float time, float x, float y)                                 = 0;
-    virtual void onDraw(sf::RenderTarget& target, const sf::RenderStates& states) const = 0;
+    virtual bool onLoad()                                                        = 0;
+    virtual void onUpdate(float time, float x, float y)                          = 0;
+    virtual void onDraw(sf::RenderTarget& target, sf::RenderStates states) const = 0;
 
     std::string m_name;
     bool        m_isLoaded{};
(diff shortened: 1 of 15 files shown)


Bambo-Borris:
Thanks to the rapid response from the maintainers here ❤️ as an SFML power user this is a lovely QoL change to see return

Bambo-Borris (approved):

binary1248 (approved):

eXpl0it3r (approved):

PR #2950, opened 2024-04-16
Ensure struct data members are given an initial value

ChrisThrasher:
## Description

While I don't suspect there are any uninitialize variable bugs present, it's still good to err on the side of safety and provide an initial value nonetheless.

I did my best to search for data structures but might have missed something so please let me know if I did. The goal is that after constructing any SFML type, public or private, it's impossible to read uninitialized data.


test/Audio/SoundStream.test.cpp
```diff
@@ -1,9 +1,24 @@
...
+        STATIC_CHECK(!std::is_copy_assignable_v<sf::SoundStream>);
+        STATIC_CHECK(!std::is_nothrow_move_constructible_v<sf::SoundStream>);
+        STATIC_CHECK(!std::is_nothrow_move_assignable_v<sf::SoundStream>);
+    }
+
+    SECTION("Chunk")
+    {
+        const sf::SoundStream::Chunk chunk;
+        CHECK(chunk.samples == nullptr);
+        CHECK(chunk.sampleCount == 0);
```
ChrisThrasher, line 22: Now that `sf::SoundStream::Chunk` has initial values we can write tests for those initial values. Because these tests happen at runtime, we needed to create a `TEST_CASE` for `sf::SoundStream` which didn't already have. I thought it made sense to include this in the same commit as the other changes.

binary1248 (approved):

PR #2951, opened 2024-04-21
Replace C arrays with `std::array`

ChrisThrasher:
## Description

SFML uses C style arrays quite a lot so I wanted to start chipping away at replacing them with `std::array`. I chose to start with instances where switching to `std::array` was particularly straightforward. Eventually it would be nice to enable the clang-tidy check that forbids C arrays.

For more reading on why C style arrays are not ideal, see the C++ Core Guidelines. 

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#slcon1-prefer-using-stl-array-or-vector-instead-of-a-c-array

### DIFF
--- a/include/SFML/Graphics/Transform.inl
+++ b/include/SFML/Graphics/Transform.inl
@@ -55,7 +55,7 @@ constexpr Transform::Transform(float a00, float a01, float a02,
 ////////////////////////////////////////////////////////////
 constexpr const float* Transform::getMatrix() const
 {
-    return m_matrix;
+    return m_matrix.data();
 }
 
 
@@ -133,8 +133,8 @@ constexpr FloatRect Transform::transformRect(const FloatRect& rectangle) const
 ////////////////////////////////////////////////////////////
 constexpr Transform& Transform::combine(const Transform& transform)
 {
-    const float* a = m_matrix;
-    const float* b = transform.m_matrix;
+    const auto& a = m_matrix;
+    const auto& b = transform.m_matrix;
 
     // clang-format off
     *this = Transform(a[0] * b[0]  + a[4] * b[1]  + a[12] * b[3],
--- a/include/SFML/System/Utf.inl
+++ b/include/SFML/System/Utf.inl
@@ -56,7 +56,7 @@ In Utf<8>::decode(In begin, In end, std::uint32_t& output, std::uint32_t replace
 {
     // clang-format off
     // Some useful precomputed data
-    static constexpr int trailing[256] =
+    static constexpr std::array<std::uint8_t, 256> trailing =
     {
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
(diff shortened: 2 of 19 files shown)


src/SFML/Window/macOS/JoystickImpl.hpp
```diff
@@ -121,7 +121,7 @@ class JoystickImpl
     Joystick::Identification m_identification; ///< Joystick identification
 
     // NOLINTNEXTLINE(readability-identifier-naming)
-    static Location m_locationIDs[sf::Joystick::Count]; ///< Global Joystick register
+    static inline std::array<Location, Joystick::Count> m_locationIDs{}; ///< Global Joystick register
```
ChrisThrasher, line 124: Adding `inline` here means we can stop defining this static member in the .cpp file. `inline` constants are a great C++17 feature that we've made use of elsewhere.

src/SFML/Audio/SoundFileWriterWav.cpp
```diff
@@ -43,25 +43,25 @@ namespace
 
 void encode(std::ostream& stream, std::int16_t value)
 {
-    const std::byte bytes[] = {static_cast<std::byte>(value & 0xFF), static_cast<std::byte>(value >> 8)};
-    stream.write(reinterpret_cast<const char*>(bytes), sizeof(bytes));
+    const std::array bytes = {static_cast<char>(value & 0xFF), static_cast<char>(value >> 8)};
+    stream.write(bytes.data(), bytes.size());
```
ChrisThrasher, line 47: I think I'm the one who started using `std::byte` here. It wasn't the right move since we immediately have to `reintrepret_cast` to a `char*`. Might as well just keeping using `char` to avoid extra casts.

include/SFML/System/Utf.inl
```diff
@@ -56,7 +56,7 @@ In Utf<8>::decode(In begin, In end, std::uint32_t& output, std::uint32_t replace
 {
     // clang-format off
     // Some useful precomputed data
-    static constexpr int trailing[256] =
+    static constexpr std::array<std::uint8_t, 256> trailing =
```
ChrisThrasher, line 59: The template type was changed from `int` to `std::uint8_t` for two reasons. First of all using a smaller type saves memory. We don't need numbers bigger than `5` so `std::uint8_t` is a more efficient representation. Second, by using an unsigned type it avoids a sign conversion warning when using one of these array members to index into an array. C-style arrays let you index with signed types but `std::array` expects a `std::size_t`.

src/SFML/Window/macOS/JoystickImpl.hpp
ChrisThrasher: I opened #2952 which makes this change in isolation as well as finding another place where this refactor can take place.

ChrisThrasher:
Rebased on `master` and fixed merge conflicts

eXpl0it3r (approved):
Just one question, but maybe you want to tackle that separately.

include/SFML/Graphics/Transform.inl
```diff
@@ -55,7 +55,7 @@ constexpr Transform::Transform(float a00, float a01, float a02,
 ////////////////////////////////////////////////////////////
 constexpr const float* Transform::getMatrix() const
```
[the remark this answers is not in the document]
ChrisThrasher: I think it's fine as-is. If we change to something like a `std::array` return type then we're returning an entire copy of the array instead of just a view into that array which is unnecessary work.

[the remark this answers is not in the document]
ChrisThrasher: Ideally we make a note to change this to a `std::span<float, 16>` in SFML 4 once we can get our hands on C++20 features.

PR #2825, opened 2023-12-10
Unify TODO comment style for better searchability

ChrisThrasher:
## Description

In my ~2 years working on SFML I've heard anyone ask about lingering TODOs or try to fix them. Regardless this is a good first step towards bringing more attention to TODO comments. Now they all use the same style so they're easy to search for in the codebase.

PR #2958, opened 2024-04-27
Unify TODO comment style for better searchability

ChrisThrasher:
## Description

Continuing precedent set in https://github.com/SFML/SFML/pull/2825


ChrisThrasher:
https://github.com/SFML/SFML/blob/5da286487ae79703c4b918c00f9542be67d8e660/src/SFML/Window/iOS/JoystickImpl.mm#L34-L37

I found some more functions that might need `TODO` statements since they lack an implementation but in theory could be implemented.

eXpl0it3r (approved):

PR #2960, opened 2024-04-27
Prohibit use of malloc

ChrisThrasher:
## Description

We only use `std::malloc` in one place due to a C API requiring it. In the past [we used to have more `malloc`s](https://github.com/SFML/SFML/pull/2743) so lets prohibit any further use.

### DIFF
--- a/src/SFML/Window/Unix/WindowImplX11.cpp
+++ b/src/SFML/Window/Unix/WindowImplX11.cpp
@@ -957,8 +957,10 @@ void WindowImplX11::setIcon(const Vector2u& size, const std::uint8_t* pixels)
 {
     // X11 wants BGRA pixels: swap red and blue channels
     // Note: this memory will be freed by X11Ptr<XImage> deleter
+    // NOLINTBEGIN(cppcoreguidelines-no-malloc)
     auto* iconPixels = static_cast<std::uint8_t*>(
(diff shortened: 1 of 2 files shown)


src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -957,8 +957,10 @@ void WindowImplX11::setIcon(const Vector2u& size, const std::uint8_t* pixels)
 {
     // X11 wants BGRA pixels: swap red and blue channels
     // Note: this memory will be freed by X11Ptr<XImage> deleter
+    // NOLINTBEGIN(cppcoreguidelines-no-malloc)
     auto* iconPixels = static_cast<std::uint8_t*>(
         std::malloc(static_cast<std::size_t>(size.x) * static_cast<std::size_t>(size.y) * 4));
+    // NOLINTEND(cppcoreguidelines-no-malloc)
```
ChrisThrasher, line 963: This was pretty annoying. If you add an inline `// NOLINT` comment on the same line as the `std::malloc` call, clang-format ends up reformatting that line such that the `std::malloc` line is on the top line. If you switch to a `// NOLINTNEXTLINE` then clang-format keeps the current formatting where `std::malloc` is on the second line of the expression.

eXpl0it3r (approved):

PR #2964, opened 2024-04-29
Use `std::optional` rather than sentinel values

ChrisThrasher:
## Description

`sf::InputStream` and derived classes have four functions whose return value operates in the same way, `open`, `read`, `seek`, and `tell`. Upon success, they return a non-negative number which means different things for different functions. Upon failure, they return `-1`. `-1` is a [sentinel value](https://en.wikipedia.org/wiki/Sentinel_value). This error value is incompatible with other successful return values. It's not valid to, for example, add `-1` to a non-negative return value from `getSize()`. However, _the type system allows this_. Because both return values are of type `std::int64_t` it's very easy to ignore a potential `-1` return value and instead write code that behaves unexpected in the face of errors.

This PR replaces those `std::int64_t` return values with `std::optional<std::size_t>`. Using `std::optional` forces all callers of these functions to reckon with a potential error. This has major type safety implications since it's no longer possible to silently ignore the possibility of `-1` being returned. 

As it turns out, there are places where these functions were called without taking into consideration the possibility of failure. For example the following snippet assumes that `file.getSize()` succeeds.

https://github.com/SFML/SFML/blob/73126c93a372d0ec867f9bc621b213094750ac40/examples/vulkan/Vulkan.cpp#L934

What happens if it does not succeed? In the case that `-1` is returned then that value is underflowed to the maximum value of `std::size_t`. When divided by `sizeof(std::uint32_t)` you end up with a value that is certainly too large to be successfully allocated. Failure to check this return value morphs into an allocation failure which is a confusing user experience.

Here's another example. The return values of `tell` and `getSize` are used without handling the potential error case. It is not valid to increment `offset` by `-1`. `-1` does not have any arithmetic meaning. It is a placeholder value. Subtracting the offset from `-1` leads to a number that is even more negative. Yet another nonsense value yet the existing API did nothing to force us to handle this error, highlighting its lack of safety.

https://github.com/SFML/SFML/blob/73126c93a372d0ec867f9bc621b213094750ac40/src/SFML/Audio/SoundFileReaderOgg.cpp#L55-L58

In this case I chose to use `.value()` to ensure that the optional has a value or else let `std::bad_optional_access` be thrown. We could instead simply use `operator*` but then we open ourselves up to undefined behavior in the event that the optional is null which seemed like a worse user experience that an uncaught exception causing a program exit.

In some places we're working with C APIs that still expect `-1` to signal error and those circumstances are still handled, albeit handled in a more explicit way that makes it clear that the C APIs handle errors differently than SFML itself.

In performing this refactor I reduced the number of `static_cast`s by a count of 19. Removing the need for casts is a good sign that `std::optional<std::size_t>` is a more natural fit than `std::int64_t`.

### DIFF
--- a/examples/vulkan/Vulkan.cpp
+++ b/examples/vulkan/Vulkan.cpp
@@ -931,9 +931,10 @@ public:
                 return;
             }
 
-            std::vector<std::uint32_t> buffer(static_cast<std::size_t>(file.getSize()) / sizeof(std::uint32_t));
+            const auto                 fileSize = file.getSize().value();
+            std::vector<std::uint32_t> buffer(fileSize / sizeof(std::uint32_t));
 
-            if (file.read(buffer.data(), file.getSize()) != file.getSize())
+            if (file.read(buffer.data(), fileSize) != file.getSize())
             {
                 vulkanAvailable = false;
                 return;
@@ -959,9 +960,10 @@ public:
                 return;
             }
 
-            std::vector<std::uint32_t> buffer(static_cast<std::size_t>(file.getSize()) / sizeof(std::uint32_t));
+            const auto                 fileSize = file.getSize().value();
+            std::vector<std::uint32_t> buffer(fileSize / sizeof(std::uint32_t));
 
-            if (file.read(buffer.data(), file.getSize()) != file.getSize())
+            if (file.read(buffer.data(), fileSize) != file.getSize())
             {
                 vulkanAvailable = false;
                 return;
--- a/include/SFML/System/FileInputStream.hpp
+++ b/include/SFML/System/FileInputStream.hpp
@@ -111,36 +111,36 @@ public:
     /// \param data Buffer where to copy the read data
     /// \param size Desired number of bytes to read
     ///
-    /// \return The number of bytes actually read, or -1 on error
+    /// \return The number of bytes actually read, or `std::nullopt` on error
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] std::int64_t read(void* data, std::int64_t size) override;
+    [[nodiscard]] std::optional<std::size_t> read(void* data, std::size_t size) override;
 
     ////////////////////////////////////////////////////////////
     /// \brief Change the current reading position
     ///
     /// \param position The position to seek to, from the beginning
     ///
-    /// \return The position actually sought to, or -1 on error
+    /// \return The position actually sought to, or `std::nullopt` on error
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] std::int64_t seek(std::int64_t position) override;
+    [[nodiscard]] std::optional<std::size_t> seek(std::size_t position) override;
 
     ////////////////////////////////////////////////////////////
     /// \brief Get the current reading position in the stream
     ///
-    /// \return The current position, or -1 on error.
+    /// \return The current position, or `std::nullopt` on error.
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] std::int64_t tell() override;
+    [[nodiscard]] std::optional<std::size_t> tell() override;
 
     ////////////////////////////////////////////////////////////
     /// \brief Return the size of the stream
     ///
-    /// \return The total number of bytes available in the stream, or -1 on error
+    /// \return The total number of bytes available in the stream, or `std::nullopt` on error
     ///
     ////////////////////////////////////////////////////////////
-    std::int64_t getSize() override;
+    std::optional<std::size_t> getSize() override;
 
 private:
     ////////////////////////////////////////////////////////////
--- a/src/SFML/Audio/SoundFileReaderFlac.cpp
+++ b/src/SFML/Audio/SoundFileReaderFlac.cpp
@@ -44,15 +44,17 @@ FLAC__StreamDecoderReadStatus streamRead(const FLAC__StreamDecoder*, FLAC__byte 
 {
     auto* data = static_cast<sf::priv::SoundFileReaderFlac::ClientData*>(clientData);
 
-    const std::int64_t count = data->stream->read(buffer, static_cast<std::int64_t>(*bytes));
-    if (count > 0)
+    if (const std::optional count = data->stream->read(buffer, *bytes))
     {
-        *bytes = static_cast<std::size_t>(count);
-        return FLAC__STREAM_DECODER_READ_STATUS_CONTINUE;
-    }
-    else if (count == 0)
-    {
-        return FLAC__STREAM_DECODER_READ_STATUS_END_OF_STREAM;
+        if (*count > 0)
+        {
+            *bytes = *count;
+            return FLAC__STREAM_DECODER_READ_STATUS_CONTINUE;
+        }
+        else
+        {
+            return FLAC__STREAM_DECODER_READ_STATUS_END_OF_STREAM;
+        }
     }
     else
     {
@@ -64,8 +66,7 @@ FLAC__StreamDecoderSeekStatus streamSeek(const FLAC__StreamDecoder*, FLAC__uint6
 {
     auto* data = static_cast<sf::priv::SoundFileReaderFlac::ClientData*>(clientData);
 
-    const std::int64_t position = data->stream->seek(static_cast<std::int64_t>(absoluteByteOffset));
-    if (position >= 0)
+    if (data->stream->seek(static_cast<std::size_t>(absoluteByteOffset)).has_value())
         return FLAC__STREAM_DECODER_SEEK_STATUS_OK;
     else
         return FLAC__STREAM_DECODER_SEEK_STATUS_ERROR;
@@ -75,10 +76,9 @@ FLAC__StreamDecoderTellStatus streamTell(const FLAC__StreamDecoder*, FLAC__uint6
 {
     auto* data = static_cast<sf::priv::SoundFileReaderFlac::ClientData*>(clientData);
 
-    const std::int64_t position = data->stream->tell();
-    if (position >= 0)
+    if (const std::optional position = data->stream->tell())
     {
-        *absoluteByteOffset = static_cast<FLAC__uint64>(position);
+        *absoluteByteOffset = *position;
         return FLAC__STREAM_DECODER_TELL_STATUS_OK;
     }
     else
@@ -91,10 +91,9 @@ FLAC__StreamDecoderLengthStatus streamLength(const FLAC__StreamDecoder*, FLAC__u
 {
     auto* data = static_cast<sf::priv::SoundFileReaderFlac::ClientData*>(clientData);
 
-    const std::int64_t count = data->stream->getSize();
-    if (count >= 0)
+    if (const std::optional count = data->stream->getSize())
     {
-        *streamLength = static_cast<FLAC__uint64>(count);
+        *streamLength = *count;
         return FLAC__STREAM_DECODER_LENGTH_STATUS_OK;
     }
     else
--- a/src/SFML/Audio/SoundFileReaderMp3.cpp
+++ b/src/SFML/Audio/SoundFileReaderMp3.cpp
@@ -68,14 +68,14 @@ namespace
 std::size_t readCallback(void* ptr, std::size_t size, void* data)
 {
     auto* stream = static_cast<sf::InputStream*>(data);
-    return static_cast<std::size_t>(stream->read(ptr, static_cast<std::int64_t>(size)));
+    return stream->read(ptr, size).value_or(-1);
 }
 
 int seekCallback(std::uint64_t offset, void* data)
 {
-    auto*              stream   = static_cast<sf::InputStream*>(data);
-    const std::int64_t position = stream->seek(static_cast<std::int64_t>(offset));
-    return position < 0 ? -1 : 0;
+    auto*               stream   = static_cast<sf::InputStream*>(data);
+    const std::optional position = stream->seek(static_cast<std::size_t>(offset));
+    return position ? 0 : -1;
 }
 
 bool hasValidId3Tag(const std::uint8_t* header)
@@ -92,7 +92,7 @@ bool SoundFileReaderMp3::check(InputStream& stream)
 {
     std::uint8_t header[10];
 
-    if (static_cast<std::size_t>(stream.read(header, static_cast<std::int64_t>(sizeof(header)))) < sizeof(header))
+    if (stream.read(header, sizeof(header)) != sizeof(header))
         return false;
 
     if (hasValidId3Tag(header))
--- a/src/SFML/Audio/SoundFileReaderOgg.cpp
+++ b/src/SFML/Audio/SoundFileReaderOgg.cpp
@@ -41,29 +41,32 @@ namespace
 std::size_t read(void* ptr, std::size_t size, std::size_t nmemb, void* data)
 {
     auto* stream = static_cast<sf::InputStream*>(data);
-    return static_cast<std::size_t>(stream->read(ptr, static_cast<std::int64_t>(size * nmemb)));
+    return stream->read(ptr, size * nmemb).value_or(-1);
 }
 
-int seek(void* data, ogg_int64_t offset, int whence)
+int seek(void* data, ogg_int64_t signedOffset, int whence)
 {
     auto* stream = static_cast<sf::InputStream*>(data);
+    auto  offset = static_cast<std::size_t>(signedOffset);
     switch (whence)
     {
         case SEEK_SET:
             break;
         case SEEK_CUR:
-            offset += stream->tell();
+            offset += stream->tell().value();
             break;
         case SEEK_END:
-            offset = stream->getSize() - offset;
+            offset = stream->getSize().value() - offset;
     }
-    return static_cast<int>(stream->seek(offset));
+    const std::optional position = stream->seek(offset);
+    return position ? static_cast<int>(*position) : -1;
 }
 
 long tell(void* data)
 {
-    auto* stream = static_cast<sf::InputStream*>(data);
-    return static_cast<long>(stream->tell());
+    auto*               stream   = static_cast<sf::InputStream*>(data);
+    const std::optional position = stream->tell();
+    return position ? static_cast<long>(*position) : -1;
 }
 
 ov_callbacks callbacks = {&read, &seek, nullptr, &tell};
--- a/src/SFML/Audio/SoundFileReaderWav.cpp
+++ b/src/SFML/Audio/SoundFileReaderWav.cpp
@@ -43,13 +43,13 @@ namespace
 {
 ma_result onRead(ma_decoder* decoder, void* buffer, size_t bytesToRead, size_t* bytesRead)
 {
-    auto*      stream = static_cast<sf::InputStream*>(decoder->pUserData);
-    const auto count  = stream->read(buffer, static_cast<std::int64_t>(bytesToRead));
+    auto*               stream = static_cast<sf::InputStream*>(decoder->pUserData);
+    const std::optional count  = stream->read(buffer, bytesToRead);
 
-    if (count < 0)
+    if (!count.has_value())
         return MA_ERROR;
 
-    *bytesRead = static_cast<size_t>(count);
+    *bytesRead = static_cast<std::size_t>(*count);
     return MA_SUCCESS;
 }
 
@@ -61,19 +61,17 @@ ma_result onSeek(ma_decoder* decoder, ma_int64 byteOffset, ma_seek_origin origin
     {
         case ma_seek_origin_start:
         {
-            if (stream->seek(byteOffset) < 0)
+            if (!stream->seek(static_cast<std::size_t>(byteOffset)).has_value())
                 return MA_ERROR;
 
             return MA_SUCCESS;
         }
         case ma_seek_origin_current:
         {
-            const auto currentPosition = stream->tell();
-
-            if (currentPosition < 0)
+            if (!stream->tell().has_value())
                 return MA_ERROR;
 
-            if (stream->seek(stream->tell() + byteOffset) < 0)
+            if (!stream->seek(stream->tell().value() + static_cast<std::size_t>(byteOffset)).has_value())
                 return MA_ERROR;
 
             return MA_SUCCESS;
--- a/src/SFML/Graphics/Image.cpp
+++ b/src/SFML/Graphics/Image.cpp
@@ -55,14 +55,15 @@ namespace
 // stb_image callbacks that operate on a sf::InputStream
 int read(void* user, char* data, int size)
 {
-    auto& stream = *static_cast<sf::InputStream*>(user);
-    return static_cast<int>(stream.read(data, size));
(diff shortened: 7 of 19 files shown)


ChrisThrasher:

ChrisThrasher:

src/SFML/Audio/SoundFileReaderWav.cpp
```diff
@@ -109,7 +109,7 @@ namespace sf::priv
 bool SoundFileReaderWav::check(InputStream& stream)
 {
     char header[mainChunkSize];
-    if (stream.read(header, sizeof(header)) < static_cast<std::int64_t>(sizeof(header)))
+    if (stream.read(header, sizeof(header)).value_or(-1) < sizeof(header))
```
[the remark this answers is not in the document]
ChrisThrasher: I think this is the right move. `!= sizeof(foo)` is already used in other places.

ChrisThrasher:

ChrisThrasher:

src/SFML/System/MemoryInputStream.cpp
```diff
@@ -36,19 +36,19 @@ namespace sf
...
+std::optional<std::size_t> MemoryInputStream::read(void* data, std::size_t size)
 {
     if (!m_data)
-        return -1;
+        return std::nullopt;
 
-    const std::int64_t endPosition = m_offset + size;
-    const std::int64_t count       = endPosition <= m_size ? size : m_size - m_offset;
+    const std::size_t endPosition = m_offset + size;
+    const std::size_t count       = endPosition <= m_size ? size : m_size - m_offset;
```
ChrisThrasher, line 51: Changing this from signed to unsigned arithmetic gives me pause. I want to be sure I didn't accidentally change how this expression evaluates.

ChrisThrasher:

src/SFML/Graphics/Shader.cpp
```diff
@@ -103,20 +103,20 @@ bool getFileContents(const std::filesystem::path& filename, std::vector<char>& b
 // Read the contents of a stream into an array of char
 bool getStreamContents(sf::InputStream& stream, std::vector<char>& buffer)
 {
-    bool               success = false;
-    const std::int64_t size    = stream.getSize();
-    if (size > 0)
+    bool       success = false;
+    const auto size    = stream.getSize();
+    if (size.has_value() && *size > 0)
```
kimci86, line 108: What about this:
```suggestion
    if (size > 0)
```

src/SFML/System/MemoryInputStream.cpp
kimci86: It looks ok, but IMO it would be more readable with `std::min(m_size, m_offset + size)` to get the end position, then computing count as `end - m_offset`.

src/SFML/Graphics/Shader.cpp
ChrisThrasher: Screenshot 2024-04-30 at 12 38 42 PM

This is valid but we have to be careful because `operator<` has the opposite behavior and will return `true` if `size` is null.

src/SFML/System/FileInputStream.cpp
```diff
@@ -78,39 +78,39 @@ bool FileInputStream::open(const std::filesystem::path& filename)
...
 
 ////////////////////////////////////////////////////////////
-std::int64_t FileInputStream::tell()
+std::optional<std::size_t> FileInputStream::tell()
 {
     if (!m_file)
-        return -1;
+        return std::nullopt;
 #ifdef SFML_SYSTEM_ANDROID
     return m_file->tell();
```
ChrisThrasher, line 115: This is a bug since it may return `-1` which when implicitly converted to a `std::optional<std::size_t>` leads to the wrong result. The optional needs to be null when `-1` is returned.

ChrisThrasher:
I added more `sf::MemoryInputStream` tests to better cover edge case behavior.

examples/vulkan/Vulkan.cpp
```diff
@@ -931,9 +931,9 @@ class VulkanExample
                 return;
             }
 
-            std::vector<std::uint32_t> buffer(static_cast<std::size_t>(file.getSize()) / sizeof(std::uint32_t));
+            std::vector<std::uint32_t> buffer(file.getSize().value() / sizeof(std::uint32_t));
 
-            if (file.read(buffer.data(), file.getSize()) != file.getSize())
+            if (file.read(buffer.data(), file.getSize().value()) != file.getSize())
```
vittorioromeo, line 936: Nitpick but we probably don't want to promote excessive use of `.value()`:

```suggestion
const auto fileSize = file.getSize().value();
std::vector<std::uint32_t> buffer(fileSize / sizeof(std::uint32_t));

if (file.read(buffer.data(), fileSize) != file.getSize())
```

examples/vulkan/Vulkan.cpp
```diff
@@ -959,9 +959,9 @@ class VulkanExample
...
```
vittorioromeo, line 964: Ditto

include/SFML/System/FileInputStream.hpp
```diff
@@ -111,36 +111,36 @@ class SFML_SYSTEM_API FileInputStream : public InputStream
     /// \param data Buffer where to copy the read data
     /// \param size Desired number of bytes to read
     ///
-    /// \return The number of bytes actually read, or -1 on error
+    /// \return The number of bytes actually read, or std::nullopt on error
```
vittorioromeo, line 114:
```suggestion
    /// \return The number of bytes actually read, or `std::nullopt` on error
```

Doxygen should support inline code blocks
https://www.doxygen.nl/manual/markdown.html#md_codespan

vittorioromeo (commented):
Seems like a good idea to me!

include/SFML/System/FileInputStream.hpp
ChrisThrasher: Is it out of scope to fix that in other places where `std::nullopt` is used in a comment without surrounding backticks?

vittorioromeo: I wouldn't care that much, but probably best to leave it for a separate PR. I'm quite sure I've inadvertently used single quotes before...

ChrisThrasher:

ChrisThrasher:
Found some Android APIs that return `-1` on failure and thus need some extra code to check for that case to return `std::nullopt`.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

src/SFML/Audio/SoundFileReaderFlac.cpp
```diff
@@ -44,15 +44,17 @@ FLAC__StreamDecoderReadStatus streamRead(const FLAC__StreamDecoder*, FLAC__byte
 {
     auto* data = static_cast<sf::priv::SoundFileReaderFlac::ClientData*>(clientData);
 
-    const std::int64_t count = data->stream->read(buffer, static_cast<std::int64_t>(*bytes));
-    if (count > 0)
+    if (const auto count = data->stream->read(buffer, *bytes))
```
vittorioromeo, line 47:
```suggestion
    if (const auto count = data->stream->read(buffer, *bytes); count.has_value())
```

Nitpick, but this suggestion makes it obvious that `count` is an optional in this context

src/SFML/Audio/SoundFileReaderFlac.cpp
```diff
@@ -75,10 +76,9 @@ FLAC__StreamDecoderTellStatus streamTell(const FLAC__StreamDecoder*, FLAC__uint6
 {
     auto* data = static_cast<sf::priv::SoundFileReaderFlac::ClientData*>(clientData);
 
-    const std::int64_t position = data->stream->tell();
-    if (position >= 0)
+    if (const auto position = data->stream->tell())
```
vittorioromeo, line 79: Ditto
```suggestion
    if (const auto position = data->stream->tell(); position.has_value())
```

src/SFML/Audio/SoundFileReaderFlac.cpp
```diff
@@ -91,10 +91,9 @@ FLAC__StreamDecoderLengthStatus streamLength(const FLAC__StreamDecoder*, FLAC__u
 {
     auto* data = static_cast<sf::priv::SoundFileReaderFlac::ClientData*>(clientData);
 
-    const std::int64_t count = data->stream->getSize();
-    if (count >= 0)
+    if (const auto count = data->stream->getSize())
```
vittorioromeo, line 94:
```suggestion
    if (const auto count = data->stream->getSize(); count.has_value())
```

src/SFML/Audio/SoundFileReaderMp3.cpp
```diff
@@ -68,14 +68,14 @@ namespace
 std::size_t readCallback(void* ptr, std::size_t size, void* data)
 {
     auto* stream = static_cast<sf::InputStream*>(data);
-    return static_cast<std::size_t>(stream->read(ptr, static_cast<std::int64_t>(size)));
+    return stream->read(ptr, size).value_or(-1);
```
vittorioromeo, line 71: Nice use of `value_or`! Maybe simplify to:

```cpp
return static_cast<sf::InputStream*>(data)->read(ptr, size).value_or(-1);
```

src/SFML/Audio/SoundFileReaderMp3.cpp
```diff
@@ -68,14 +68,14 @@ namespace
...
 }
 
 int seekCallback(std::uint64_t offset, void* data)
 {
-    auto*              stream   = static_cast<sf::InputStream*>(data);
-    const std::int64_t position = stream->seek(static_cast<std::int64_t>(offset));
-    return position < 0 ? -1 : 0;
+    auto*      stream   = static_cast<sf::InputStream*>(data);
+    const auto position = stream->seek(static_cast<std::size_t>(offset));
+    return position ? 0 : -1;
```
vittorioromeo, line 78:
```suggestion
    return position.has_value() ? 0 : -1;
```

src/SFML/Audio/SoundFileReaderMp3.cpp
```diff
@@ -92,7 +92,7 @@ bool SoundFileReaderMp3::check(InputStream& stream)
 {
     std::uint8_t header[10];
 
-    if (static_cast<std::size_t>(stream.read(header, static_cast<std::int64_t>(sizeof(header)))) < sizeof(header))
+    if (stream.read(header, sizeof(header)) != sizeof(header))
```
vittorioromeo, line 95: Is there a reason why this used to be `<`? Could `read` actually ever return more bytes than the header?

src/SFML/Audio/SoundFileReaderOgg.cpp
```diff
@@ -41,29 +41,32 @@ namespace
...
-            offset += stream->tell();
+            offset += stream->tell().value();
             break;
         case SEEK_END:
-            offset = stream->getSize() - offset;
+            offset = stream->getSize().value() - offset;
     }
-    return static_cast<int>(stream->seek(offset));
+    const auto position = stream->seek(offset);
+    return position ? static_cast<int>(*position) : -1;
```
vittorioromeo, line 62:
```suggestion
    return stream->seek(offset).value_or(-1);
```

src/SFML/Audio/SoundFileReaderOgg.cpp
```diff
@@ -41,29 +41,32 @@ namespace
...
 }
 
 long tell(void* data)
 {
-    auto* stream = static_cast<sf::InputStream*>(data);
-    return static_cast<long>(stream->tell());
+    auto*      stream   = static_cast<sf::InputStream*>(data);
+    const auto position = stream->tell();
+    return position ? static_cast<long>(*position) : -1;
```
vittorioromeo, line 69:
```suggestion
    return static_cast<sf::InputStream*>(data)->tell().value_or(-1L);
```

src/SFML/Graphics/Image.cpp
```diff
@@ -55,14 +55,15 @@ namespace
 // stb_image callbacks that operate on a sf::InputStream
 int read(void* user, char* data, int size)
 {
-    auto& stream = *static_cast<sf::InputStream*>(user);
-    return static_cast<int>(stream.read(data, size));
+    auto&      stream = *static_cast<sf::InputStream*>(user);
+    const auto count  = stream.read(data, static_cast<std::size_t>(size));
+    return count ? static_cast<int>(*count) : -1;
```
vittorioromeo, line 60:
```suggestion
    return static_cast<sf::InputStream*>(user)->read(data, static_cast<std::size_t>(size)).value_or(-1);
```

Or keep the `stream` variable here if the line seems too complicated.

ChrisThrasher:

src/SFML/System/Android/ResourceStream.cpp
```diff
@@ -41,62 +41,41 @@ ResourceStream::ResourceStream(const std::filesystem::path& filename)
...
-        return AAsset_read(m_file.get(), data, static_cast<std::size_t>(size));
-    }
-    else
-    {
-        return -1;
-    }
+    const auto numBytesRead = AAsset_read(m_file.get(), data, size);
+    if (numBytesRead < 0)
+        return std::nullopt;
+    return numBytesRead;
```
vittorioromeo, line 54:
```suggestion
    const std::size_t numBytesRead = AAsset_read(m_file.get(), data, size);
    return numBytesRead < 0 ? std::nullopt : std::optional{numBytesRead};
```

src/SFML/System/Android/ResourceStream.cpp
```diff
@@ -41,62 +41,41 @@ ResourceStream::ResourceStream(const std::filesystem::path& filename)
...
-        return AAsset_seek(m_file.get(), static_cast<off_t>(position), SEEK_SET);
-    }
-    else
-    {
-        return -1;
-    }
+    const auto newPosition = AAsset_seek(m_file.get(), static_cast<off_t>(position), SEEK_SET);
+    if (newPosition < 0)
+        return std::nullopt;
+    return newPosition;
```
vittorioromeo, line 64: Ditto, like above

src/SFML/System/Android/ResourceStream.cpp
```diff
@@ -41,62 +41,41 @@ ResourceStream::ResourceStream(const std::filesystem::path& filename)
...
 {
-    if (m_file)
-    {
-        return getSize() - AAsset_getRemainingLength(m_file.get());
-    }
-    else
-    {
-        return -1;
-    }
+    return getSize().value() - static_cast<std::size_t>(AAsset_getRemainingLength(m_file.get()));
```
vittorioromeo, line 71: Does this have the same semantics as before? Wouldn't this be closer?

```cpp
if (!m_file) { return std::nullopt; }
return getSize().value() - static_cast<std::size_t>(AAsset_getRemainingLength(m_file.get()));

File: src/SFML/System/MemoryInputStream.cpp [hunk 20]
```diff
@@ -27,6 +27,8 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/System/MemoryInputStream.hpp>
 
+#include <algorithm>
```
vittorioromeo, line 30: 😔

src/SFML/System/MemoryInputStream.cpp
```diff
@@ -61,31 +61,31 @@ std::int64_t MemoryInputStream::read(void* data, std::int64_t size)
...
 ////////////////////////////////////////////////////////////
-std::int64_t MemoryInputStream::getSize()
+std::optional<std::size_t> MemoryInputStream::getSize()
 {
     if (!m_data)
-        return -1;
+        return std::nullopt;
 
     return m_size;
 }
```
vittorioromeo, line 91: I'd use ternaries here

vittorioromeo (commented):
Generally looks good, some suggestions

src/SFML/Audio/SoundFileReaderOgg.cpp
ChrisThrasher: `std::optional<std::size_t>::value_or` returns a `std::size_t` so `.value_or(-1)` will return a `std::numeric_limits<std::size_t>::max()` which is not what I want to do here.

src/SFML/Graphics/Image.cpp
ChrisThrasher: std::optional<std::size_t>::value_or returns a std::size_t so .value_or(-1) will return a std::numeric_limits<std::size_t>::max() which is not what I want to do here.

src/SFML/System/Android/ResourceStream.cpp
ChrisThrasher: `m_file` cannot be null so we're free to omit that check.

src/SFML/System/MemoryInputStream.cpp
```diff
@@ -36,20 +38,18 @@ namespace sf
...
+std::optional<std::size_t> MemoryInputStream::read(void* data, std::size_t size)
 {
     if (!m_data)
-        return -1;
-
-    const std::int64_t endPosition = m_offset + size;
-    const std::int64_t count       = endPosition <= m_size ? size : m_size - m_offset;
+        return std::nullopt;
 
+    const std::size_t count = std::min(size, m_size - m_offset);
```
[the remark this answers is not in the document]
ChrisThrasher: Yes. Kimci recommended this transformation and I double checked that it's correct.

ChrisThrasher (commented):
With respect to the use of `has_value()` I've gone back and forth while writing this PR. Sometimes I like explicitly saying _this is an optional_ but other times I prefer the elegance of letting the type system quietly handle things even if the reader doesn't necessarily know what's going on. I'm currently erring on the side of preferring we write less code since doing so doesn't reduce type safety whatsoever.

With respect to the use of ternaries, I don't have a strong opinion. In the past we haven't made a point to maximize the use of ternaries. If the code was already written to use ternaries I wouldn't replace them with `if` statements. I don't really see either as particularly superior to another so I just kept using whatever control flow already existed prior to this PR to reduce how much code I changed.

In both cases if the rest of the team agrees in one direction or the other I'll go along with what they prefer.

src/SFML/Audio/SoundFileReaderMp3.cpp
ChrisThrasher: I don't think the use of `<` here mattered. I think it was just an alternative way of saying "did we read the same number of bytes as the header?", in which case we ought to use `!=` like we do in many other places for this same thing.

ChrisThrasher:

src/SFML/Audio/SoundFileReaderFlac.cpp
vittorioromeo: @ChrisThrasher: The reason why I suggested this change (and similar ones) is that if I am just reading this line of code in isolation I will assume that `getSize()` returns a number and that `count` is just a number. Nothing in the name of the variable nor the name of the function suggests optionality.

I would rather read 

```cpp
if (const auto count = data->stream->getSize(); count.has_value())
```

or 

```
if (const std::optional count = data->stream->getSize())
```

or 

```cpp
if (const std::optional<std::size_t> count = data->stream->getSize())
```

ChrisThrasher:
Rebased on master and fixed conflicts

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

src/SFML/Audio/SoundFileReaderFlac.cpp
vittorioromeo: I still'd like this one to be somehow addressed because in this context it's not really obvious if `count` is an optional or if it's an integer value and we're relying on implicit conversion from integer to bool.

If you don't like my current suggestion, consider also:

```cpp
if (const std::optional count = data->stream->read(buffer, *bytes))
if (const std::optional<std::size_t> count = data->stream->read(buffer, *bytes))
```

src/SFML/Audio/SoundFileReaderFlac.cpp
vittorioromeo: Same here, I'm not trying to be needlessly pedantic but I can imagine reading this code and thinking that we're relying on `position != 0` instead of `position` being an optional. It's also the variable name that doesn't suggest any optionality.

src/SFML/Audio/SoundFileReaderFlac.cpp
vittorioromeo: Ditto

src/SFML/Audio/SoundFileReaderMp3.cpp
vittorioromeo: Same here, another possible idea is changing the name to `maybePosition` or `optionalPosition` but that's my least favorite option of everything I've suggested.

src/SFML/Audio/SoundFileReaderOgg.cpp
vittorioromeo: If `value_or` cannot be used here, I have the same readability concern as above -- it's not obvious that `position` is an optional here.

src/SFML/Audio/SoundFileReaderOgg.cpp
vittorioromeo: Ditto, either use CTAD with `std::optional` instead of `auto`, or use `.has_value()` to add some local context about `position` being an optional.

src/SFML/Graphics/Image.cpp
vittorioromeo: Same readability concern

src/SFML/System/Android/ResourceStream.cpp
vittorioromeo: Feel free to ignore this one, this is just style.

src/SFML/System/Android/ResourceStream.cpp
vittorioromeo: Feel free to ignore

src/SFML/System/MemoryInputStream.cpp
vittorioromeo: Feel free to ignore

vittorioromeo (changes requested):
See inline comments

ChrisThrasher:
Replaced `const auto foo = returns_optional();` with `const std::optional foo = returns_options();` and rebased on `master`.

vittorioromeo (approved):
LGTM, thanks for making the readability related changes!

vittorioromeo:

eXpl0it3r:
> In performing this refactor I reduced the number of `static_cast`s by a count of 19. Removing the need for casts is a good sign that `std::optional<std::size_t>` is a more natural fit than `std::int64_t`.

We'll have to look at this again. `size_t` is differently sized than `int64_t` and the stream APIs (`fseek`, `ftell`) use `long`

PR #2966, opened 2024-04-29
Fall back to using the NULL audio backend if the default backends don't provide any devices.

binary1248:
Title.

Because audio context creation will succeed even if the selected context cannot provide any devices, the NULL backend is never selected even on systems on which we know no audio device exists. We have to check for the existence of devices ourselves after creating a normal context and force miniaudio to create a NULL context if we determine a normal context would not provide us with any devices.

Should help with testing on audio device-less systems such as when running CI.

### DIFF
--- a/src/SFML/Audio/AudioDevice.cpp
+++ b/src/SFML/Audio/AudioDevice.cpp
@@ -30,6 +30,7 @@
 #include <SFML/System/Err.hpp>
 
 #include <algorithm>
+#include <array>
 #include <ostream>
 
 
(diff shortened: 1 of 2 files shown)


ChrisThrasher:
https://github.com/ChrisThrasher/SFML/actions/runs/8886742515/job/24400760325

I rebased my tests onto this PR and we're still getting Linux and Windows CI failures. It did fix the macOS issues though.

eXpl0it3r (approved):

eXpl0it3r:
Seems to work fine
image

PR #2967, opened 2024-05-01
Simplify `Event` usage by implementing `Event::operator bool`

trustytrojan:
## Description

This PR aims to simplify the usage of `Event` when polling events from a `RenderWindow`. Instead of having to initialize an `Event` on the stack and pass it by reference to `pollEvent`/`waitEvent`, this PR makes the methods return an `Event`, which can be converted to a `bool` based on whether its subtype is `Empty`.

There was discussion about having `RenderWindow::pollEvent` and `RenderWindow::waitEvent` return `std::optional<Event>`, but considering `Event` already uses subtyping to determine if it contains event data, letting `Event` be convertible to `bool` (based on whether the inner data was of type `Event::Empty`) sounded like the better option. Plus, we can avoid the clunky `->` syntax used for accessing members of an `std::optional`'s inner type.

Most of the changed files are changing comments containing example code of using `pollEvent`/`waitEvent`. The relevant change is in [this commit](https://github.com/SFML/SFML/commit/9c6d7d646c8113a7ca2ffa2d58ffab49289204bb).

## Tasks

-   [x] Tested on Linux
-   [ ] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

Use an `sf::RenderWindow` while processing events.

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        {
                window.close();
        }

        window.clear();
        window.display();
    }
}
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -57,7 +57,7 @@ body:
 
                   while (window.isOpen())
                   {
-                      for (sf::Event event; window.pollEvent(event);)
+                      while (const auto event = window.pollEvent())
                       {
                           if (event.is<sf::Event::Closed>())
                               window.close();
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -54,7 +54,7 @@ body:
 
                   while (window.isOpen())
                   {
-                      for (sf::Event event; window.pollEvent(event);)
+                      while (const auto event = window.pollEvent())
                       {
                           if (event.is<sf::Event::Closed>())
                               window.close();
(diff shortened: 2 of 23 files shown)


ChrisThrasher:
I had previously played around with a similar idea [here](https://github.com/SFML/SFML/commit/068ae6f5757c26eb4634923595793fe9aad7cc57) so this is something I've been wanting to do. Using `operator bool` to avoid the `operator->` usage that `std::optional` requires is a nice bonus since it retains values semantics. So long as `sf::Event` contains an internal null state then I think it's reasonable we exploit that to our advantage.

I'm definitely open to this idea. Curious to hear what @binary1248 and @eXpl0it3r think about this.

ChrisThrasher (commented):
examples/android/app/src/main/jni/main.cpp still needs to be update for the new API

eXpl0it3r:
Superseded by #2968

PR #2968, opened 2024-05-02
`Event::operator bool`

trustytrojan:
## Description

This PR aims to simplify the usage of `Event` when polling events from a `RenderWindow`. Instead of having to initialize an `Event` on the stack and pass it by reference to `pollEvent`/`waitEvent`, this PR makes the methods return an `Event`, which can be converted to a `bool` based on whether its subtype is `Empty`.

There was discussion about having `pollEvent` and `waitEvent` return `std::optional<Event>`, but considering `Event` already uses subtyping to determine if it contains event data, letting `Event` be convertible to `bool` (based on whether the inner data was of type `Event::Empty`) sounded like the better option. Plus, we can avoid the clunky `->` syntax used for accessing members of an `std::optional`'s inner type.

## Tasks

-   [x] Tested on Linux
-   [ ] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

Use an `sf::RenderWindow` while processing events.

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        {
                window.close();
        }

        window.clear();
        window.display();
    }
}
```

### DIFF
--- a/include/SFML/Window/Event.hpp
+++ b/include/SFML/Window/Event.hpp
@@ -276,6 +276,17 @@ public:
     template <typename T>
     [[nodiscard]] const T* getIf() const;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Check if current event type is not `Empty`
+    ///
+    /// \return True if current event type is not `Empty`
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] explicit operator bool() const
+    {
+        return !is<Empty>();
+    }
+
 private:
     ////////////////////////////////////////////////////////////
     // Member data
@@ -345,7 +356,7 @@ private:
 /// any of the corresponding event data.
 ///
 /// \code
-/// for (sf::Event event; window.pollEvent(event);)
+/// while (const auto event = window.pollEvent())
 /// {
 ///     // Request for closing the window
 ///     if (event.is<sf::Event::Closed>())
--- a/include/SFML/Window/WindowBase.hpp
+++ b/include/SFML/Window/WindowBase.hpp
@@ -180,52 +180,46 @@ public:
     /// \brief Pop the next event from the front of the FIFO event queue, if any, and return it
     ///
     /// This function is not blocking: if there's no pending event then
-    /// it will return false and leave \a event unmodified.
-    /// Note that more than one event may be present in the event queue,
-    /// thus you should always call this function in a loop
-    /// to make sure that you process every pending event.
+    /// it will return an empty event. Note that more than one event
+    /// may be present in the event queue, thus you should always call
+    /// this function in a loop to make sure that you process every
+    /// pending event.
     /// \code
-    /// for (sf::Event event; window.pollEvent(event);)
+    /// while (const auto event = window.pollEvent())
     /// {
     ///    // process event...
     /// }
     /// \endcode
     ///
-    /// \param event Event to be returned
-    ///
-    /// \return True if an event was returned, or false if the event queue was empty
+    /// \return The event; will be `Empty` (convertible to `false`) if no events are pending
     ///
     /// \see waitEvent
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool pollEvent(Event& event);
+    [[nodiscard]] Event pollEvent();
 
     ////////////////////////////////////////////////////////////
     /// \brief Wait for an event and return it
     ///
     /// This function is blocking: if there's no pending event then
-    /// it will wait until an event is received.
-    /// After this function returns (and no error occurred),
-    /// the \a event object is always valid and filled properly.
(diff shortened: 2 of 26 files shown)


include/SFML/Window/Event.hpp
```diff
@@ -276,6 +276,17 @@ class SFML_WINDOW_API Event
     template <typename T>
     [[nodiscard]] const T* getIf() const;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Check if current event type is not `Empty`
+    ///
+    /// \return True if current event type is not `Empty`
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] operator bool() const
```
ChrisThrasher, line 285:
```suggestion
    [[nodiscard]] explicit operator bool() const
```
Adding `explicit` here prevents conversions to `bool` outside of conditionals which is more in line with what we want. We only want this conversion happening inside a `while`.

ChrisThrasher (commented):
I like this! It's great to remove output parameters without adding any extra ergonomic issues.

ChrisThrasher:

include/SFML/Window/Event.hpp
trustytrojan: Didn't know `explicit` could be put on conversion operators. Tested it and looks good to me

ChrisThrasher (approved):

src/SFML/Window/WindowImpl.hpp
```diff
@@ -130,11 +130,12 @@ class WindowImpl
...
     /// returns false to indicate that no event is available.
     ///
-    /// \param event Event to be returned
     /// \param block Use true to block the thread until an event arrives
     ///
+    /// \return The event; can be `Empty` (convertible to `false`) if not blocking
+    ///
     ////////////////////////////////////////////////////////////
-    bool popEvent(Event& event, bool block);
+    Event popEvent(bool block);
```
vittorioromeo, line 138:
```suggestion
    [[nodiscard]] Event popEvent(bool block);
```

ChrisThrasher:

ChrisThrasher:
Cleaned up the commit history to prepare for merge. No changes made.

ChrisThrasher:

include/SFML/Window/Event.hpp
```diff
@@ -276,6 +276,17 @@ class SFML_WINDOW_API Event
     template <typename T>
     [[nodiscard]] const T* getIf() const;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Check if current event type is not `Empty`
+    ///
+    /// \return True if current event type is not `Empty`
```
kimci86, line 282: I wonder if doxygen makes a link to the Empty struct when using backticks.

kimci86: I am not sure it is worth adding backticks actually. We do not usually do it so it would look inconsistent.

trustytrojan: I can understand; I usually use backticks in my personal code because most renderers will treat it as a line of code (aka a <code> element in HTML). It ends up looking nicer in my opinion. If for consistency you revert it, that's fine.

vittorioromeo: > I am not sure it is worth adding backticks actually. We do not usually do it so it would look inconsistent.

It makes the documentation easier to browse, I think it's worth it assuming that there are going to be subsequent PRs to improve the existing docs.

trustytrojan: Leaving it as is sounds good to me.

trustytrojan: > > I am not sure it is worth adding backticks actually. We do not usually do it so it would look inconsistent.
> 
> It makes the documentation easier to browse, I think it's worth it assuming that there are going to be subsequent PRs to improve the existing docs.

I'd be down to make those PRs.

include/SFML/Window/WindowBase.hpp
```diff
@@ -185,20 +185,18 @@ class SFML_WINDOW_API WindowBase
     /// thus you should always call this function in a loop
     /// to make sure that you process every pending event.
```
kimci86, line 186: This paragraphs needs to be updated, not to reference the former event argument.
Comment applies to waitEvent documentation too.

ChrisThrasher:

ChrisThrasher:

ChrisThrasher:

trustytrojan:
Looks like a macos build failed because it took too long to clone?

eXpl0it3r:
Nothing you can do, I've rerun the failed job.

PR #2248, opened 2022-10-16
Reimplementing `Rect` in terms of position and size

max-aider:
## Description

Class `sf::Rect` is a simple and straightforward construct, widely used in SFML library. However, it lacks many basic features and doesn't follow object oriented intuition, as well as standards established in, for example, PR #2055. This PR improves the `Rect` class by indroducing the following changes:

* A `Rect` is now defined in terms of it's `Vector2` `position` and `size`, rather than `left`, `top`, `width` and `height`.
* The class now has 4 methods for acquiring it's edges: `getLeft()`, `getTop()`, `getRight()` and `getBottom()`.
* The class now has 3 methods for acquiring it's measurements: `getStart()`, `getEnd()` and `getCenter()`.
* The class now has `getAbsoluteSize()` method.
* The class now has `getAbsoluteRect()` method, which returns the same bounding rectangle, but described with it's actual starting position and positive size.
* All mentioned methods are designed to handle rectangles with negative size.
* The implementation of old methods (`contains` and `findIntersection`) as well as methods in other classes, which are dependent on the `Rect` class, have been refactored using the new functionality.

## Tasks

* [x] Tested on Windows (first-hand)
* [x] Tested on Linux, macOS, iOS and Android (CI)

## How to test this PR?

```cpp
#include <iostream>
#include <SFML/Graphics/Rect.hpp>

// This is basically a unit test. Every line should say "1"

int main()
{
  // Positive size
  {
    const sf::IntRect rectangle({0, 0}, {10, 10});
    
    std::cout << rectangle.getLeft() == 0 << "\n";
    std::cout << rectangle.getTop() == 0 << "\n";
    std::cout << rectangle.getRight() == 10 << "\n";
    std::cout << rectangle.getBottom() == 10 << "\n";
    
    std::cout << rectangle.getStart() == sf::Vector2i(0, 0) << "\n";
    std::cout << rectangle.getEnd() == sf::Vector2i(10, 10) << "\n";
    std::cout << rectangle.getCenter() == sf::Vector2i(5, 5) << "\n";
    
    std::cout << rectangle.getAbsoluteSize() == sf::Vector2i(10, 10) << "\n";
    std::cout << rectangle.getAbsoluteRect() == rectangle << "\n";
    
    std::cout << rectangle.contains({0, 0}) << "\n";
    std::cout << !rectangle.contains({10, 10}) << "\n";
    
[...]

PR #2972, opened 2024-05-04
Implement Rect as position and size vectors

kimci86:
This groups `sf::Rect` members `left`, `top`, `width`, `height` into two vectors: `position` and `size`.

This makes `sf::Rect` more aligned with all SFML API working with `Vector2` instead of separate values.
Related: #2055, #2248

### DIFF
--- a/examples/island/Island.cpp
+++ b/examples/island/Island.cpp
@@ -152,8 +152,7 @@ int main()
     }
 
     // Center the status text
-    statusText.setPosition({(windowWidth - statusText.getLocalBounds().size.x) / 2.f,
-                            (windowHeight - statusText.getLocalBounds().size.y) / 2.f});
+    statusText.setPosition((sf::Vector2f(windowWidth, windowHeight) - statusText.getLocalBounds().size) / 2.f);
 
     // Set up an array of pointers to our settings for arrow navigation
     constexpr std::array<Setting, 9> settings = {
--- a/examples/island/Island.cpp
+++ b/examples/island/Island.cpp
@@ -152,8 +152,8 @@ int main()
     }
 
     // Center the status text
-    statusText.setPosition({(windowWidth - statusText.getLocalBounds().width) / 2.f,
-                            (windowHeight - statusText.getLocalBounds().height) / 2.f});
+    statusText.setPosition({(windowWidth - statusText.getLocalBounds().size.x) / 2.f,
+                            (windowHeight - statusText.getLocalBounds().size.y) / 2.f});
 
     // Set up an array of pointers to our settings for arrow navigation
     constexpr std::array<Setting, 9> settings = {
--- a/include/SFML/Graphics/Rect.hpp
+++ b/include/SFML/Graphics/Rect.hpp
@@ -105,26 +105,6 @@ public:
     constexpr std::optional<Rect<T>> findIntersection(const Rect<T>& rectangle) const;
 
     ////////////////////////////////////////////////////////////
-    /// \brief Get the position of the rectangle's top-left corner
-    ///
-    /// \return Position of rectangle
-    ///
-    /// \see getSize, getCenter
-    ///
-    ////////////////////////////////////////////////////////////
-    constexpr Vector2<T> getPosition() const;
-
-    ////////////////////////////////////////////////////////////
-    /// \brief Get the size of the rectangle
-    ///
-    /// \return Size of rectangle
-    ///
-    /// \see getPosition, getCenter
-    ///
-    ////////////////////////////////////////////////////////////
-    constexpr Vector2<T> getSize() const;
-
-    ////////////////////////////////////////////////////////////
     /// \brief Get the position of the center of the rectangle
     ///
     /// \return Center of rectangle
--- a/include/SFML/Graphics/Rect.hpp
+++ b/include/SFML/Graphics/Rect.hpp
@@ -137,10 +137,8 @@ public:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    T left{};   //!< Left coordinate of the rectangle
-    T top{};    //!< Top coordinate of the rectangle
-    T width{};  //!< Width of the rectangle
-    T height{}; //!< Height of the rectangle
+    Vector2<T> position{}; //!< Position of the top-left corner of the rectangle
+    Vector2<T> size{};     //!< Size of the rectangle
 };
 
 ////////////////////////////////////////////////////////////
--- a/include/SFML/Graphics/Rect.hpp
+++ b/include/SFML/Graphics/Rect.hpp
@@ -109,8 +109,6 @@ public:
     ///
     /// \return Center of rectangle
     ///
-    /// \see getSize, getPosition
-    ///
     ////////////////////////////////////////////////////////////
     constexpr Vector2<T> getCenter() const;
 
@@ -166,18 +164,18 @@ using FloatRect = Rect<float>;
 ///
 /// A rectangle is defined by its top-left corner and its size.
 /// It is a very simple class defined for convenience, so
-/// its member variables (left, top, width and height) are public
+/// its member variables (position and size) are public
 /// and can be accessed directly, just like the vector classes
 /// (Vector2 and Vector3).
 ///
 /// To keep things simple, sf::Rect doesn't define
 /// functions to emulate the properties that are not directly
-/// members (such as right, bottom, center, etc.), it rather
+/// members (such as right, bottom, etc.), it rather
 /// only provides intersection functions.
 ///
 /// sf::Rect uses the usual rules for its boundaries:
 /// \li The left and top edges are included in the rectangle's area
-/// \li The right (left + width) and bottom (top + height) edges are excluded from the rectangle's area
+/// \li The right and bottom edges are excluded from the rectangle's area
 ///
 /// This means that sf::IntRect({0, 0}, {1, 1}) and sf::IntRect({1, 1}, {1, 1})
 /// don't intersect.
@@ -206,7 +204,7 @@ using FloatRect = Rect<float>;
 /// // Test the intersection between r1 and r2
 /// std::optional<sf::IntRect> result = r1.findIntersection(r2);
 /// // result.has_value() == true
-/// // result.value() == (4, 2, 16, 3)
+/// // result.value() == sf::IntRect({4, 2}, {16, 3})
 /// \endcode
 ///
 ////////////////////////////////////////////////////////////
--- a/include/SFML/Graphics/Rect.hpp
+++ b/include/SFML/Graphics/Rect.hpp
@@ -125,14 +125,14 @@ public:
 ///
 /// This operator compares strict equality between two rectangles.
 ///
-/// \param left  Left operand (a rectangle)
-/// \param right Right operand (a rectangle)
+/// \param lhs Left operand (a rectangle)
+/// \param rhs Right operand (a rectangle)
 ///
-/// \return True if \a left is equal to \a right
+/// \return True if \a lhs is equal to \a rhs
 ///
 ////////////////////////////////////////////////////////////
 template <typename T>
-[[nodiscard]] constexpr bool operator==(const Rect<T>& left, const Rect<T>& right);
+[[nodiscard]] constexpr bool operator==(const Rect<T>& lhs, const Rect<T>& rhs);
 
 ////////////////////////////////////////////////////////////
 /// \relates Rect
@@ -140,14 +140,14 @@ template <typename T>
 ///
 /// This operator compares strict difference between two rectangles.
 ///
-/// \param left  Left operand (a rectangle)
-/// \param right Right operand (a rectangle)
+/// \param lhs Left operand (a rectangle)
+/// \param rhs Right operand (a rectangle)
 ///
-/// \return True if \a left is not equal to \a right
+/// \return True if \a lhs is not equal to \a rhs
 ///
 ////////////////////////////////////////////////////////////
 template <typename T>
-[[nodiscard]] constexpr bool operator!=(const Rect<T>& left, const Rect<T>& right);
+[[nodiscard]] constexpr bool operator!=(const Rect<T>& lhs, const Rect<T>& rhs);
 
 // Create type aliases for the most common types
 using IntRect   = Rect<int>;
--- a/include/SFML/Graphics/Rect.inl
+++ b/include/SFML/Graphics/Rect.inl
@@ -112,25 +112,9 @@ constexpr std::optional<Rect<T>> Rect<T>::findIntersection(const Rect<T>& rectan
 
 ////////////////////////////////////////////////////////////
 template <typename T>
-constexpr Vector2<T> Rect<T>::getPosition() const
-{
-    return Vector2<T>(position.x, position.y);
-}
-
-
-////////////////////////////////////////////////////////////
-template <typename T>
-constexpr Vector2<T> Rect<T>::getSize() const
-{
-    return Vector2<T>(size.x, size.y);
-}
-
-
-////////////////////////////////////////////////////////////
-template <typename T>
 constexpr Vector2<T> Rect<T>::getCenter() const
 {
-    return getPosition() + getSize() / T{2};
+    return position + size / T{2};
 }
 
 
--- a/include/SFML/Graphics/Rect.inl
+++ b/include/SFML/Graphics/Rect.inl
@@ -120,17 +120,17 @@ constexpr Vector2<T> Rect<T>::getCenter() const
 
 ////////////////////////////////////////////////////////////
 template <typename T>
-constexpr bool operator==(const Rect<T>& left, const Rect<T>& right)
+constexpr bool operator==(const Rect<T>& lhs, const Rect<T>& rhs)
 {
-    return (left.position == right.position) && (left.size == right.size);
+    return (lhs.position == rhs.position) && (lhs.size == rhs.size);
 }
 
 
 ////////////////////////////////////////////////////////////
 template <typename T>
-constexpr bool operator!=(const Rect<T>& left, const Rect<T>& right)
+constexpr bool operator!=(const Rect<T>& lhs, const Rect<T>& rhs)
 {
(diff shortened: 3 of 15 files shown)


examples/island/Island.cpp
```diff
@@ -156,8 +156,7 @@ int main()
     }
 
     // Center the status text
-    statusText.setPosition({(windowWidth - statusText.getLocalBounds().width) / 2.f,
-                            (windowHeight - statusText.getLocalBounds().height) / 2.f});
+    statusText.setPosition((sf::Vector2f(windowWidth, windowHeight) - statusText.getLocalBounds().size) / 2.f);
```
vittorioromeo, line 159: Might be worthwhile changing `windowWidth` and `windowHeight` into a `windowSize` vector as well.

include/SFML/Graphics/Rect.hpp
```diff
@@ -104,43 +104,19 @@ class Rect
...
 
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    T left{};   //!< Left coordinate of the rectangle
-    T top{};    //!< Top coordinate of the rectangle
-    T width{};  //!< Width of the rectangle
-    T height{}; //!< Height of the rectangle
+    Vector2<T> position{}; //!< Position of the top-left corner of the rectangle
+    Vector2<T> size{};     //!< Size of the rectangle
```
vittorioromeo, line 119: Have you considered adding `.getLeft()`, `.getTop()`, `.getWidth()`, `.getHeight()`, etc accessors to at least ease migration and provide some generally useful functions?

include/SFML/Graphics/Rect.inl
```diff
@@ -118,36 +110,19 @@ constexpr std::optional<Rect<T>> Rect<T>::findIntersection(const Rect<T>& rectan
...
 }
 
 
 ////////////////////////////////////////////////////////////
 template <typename T>
 constexpr bool operator==(const Rect<T>& left, const Rect<T>& right)
 {
-    return (left.left == right.left) && (left.width == right.width) && (left.top == right.top) &&
-           (left.height == right.height);
+    return (left.position == right.position) && (left.size == right.size);
```
vittorioromeo, line 125: Might be a good opportunity to change `left` and `right` into `lhs` and `rhs` respectively, as it can be confusing to read `left` in this context.

include/SFML/Graphics/Transform.inl
```diff
@@ -104,29 +104,27 @@ constexpr Vector2f Transform::transformPoint(const Vector2f& point) const
...
 {
     // Transform the 4 corners of the rectangle
-    const std::array points = {transformPoint({rectangle.left, rectangle.top}),
-                               transformPoint({rectangle.left, rectangle.top + rectangle.height}),
-                               transformPoint({rectangle.left + rectangle.width, rectangle.top}),
-                               transformPoint({rectangle.left + rectangle.width, rectangle.top + rectangle.height})};
+    const std::array points = {transformPoint(rectangle.position),
+                               transformPoint(rectangle.position + Vector2f(0.f, rectangle.size.y)),
+                               transformPoint(rectangle.position + Vector2f(rectangle.size.x, 0.f)),
+                               transformPoint(rectangle.position + rectangle.size)};
```
vittorioromeo, line 110: If we add some extra accessors, we could do something like:

```suggestion
    const std::array points = {transformPoint(rectangle.getTopLeft()),
                               transformPoint(rectangle.getBottomLeft()),
                               transformPoint(rectangle.getTopRight())),
                               transformPoint(rectangle.getBottomRight())};
```

Maybe something to consider for a future PR.

src/SFML/Graphics/Font.cpp
```diff
@@ -637,24 +631,21 @@ Glyph Font::loadGlyph(std::uint32_t codePoint, unsigned int characterSize, bool
...
             }
         }
 
         // Write the pixels to the texture
-        const unsigned int x = static_cast<unsigned int>(glyph.textureRect.left) - padding;
-        const unsigned int y = static_cast<unsigned int>(glyph.textureRect.top) - padding;
-        const unsigned int w = static_cast<unsigned int>(glyph.textureRect.width) + 2 * padding;
-        const unsigned int h = static_cast<unsigned int>(glyph.textureRect.height) + 2 * padding;
-        page.texture.update(m_pixelBuffer.data(), {w, h}, {x, y});
+        const Vector2u dest = Vector2u(glyph.textureRect.position) - Vector2u(padding, padding);
```
vittorioromeo, line 647:
```suggestion
        const auto dest = Vector2u(glyph.textureRect.position) - Vector2u(padding, padding);
```

src/SFML/Graphics/Shape.cpp
```diff
@@ -267,14 +267,13 @@ void Shape::updateTexCoords()
 
     for (std::size_t i = 0; i < m_vertices.getVertexCount(); ++i)
     {
-        const float xratio      = m_insideBounds.width > 0
-                                      ? (m_vertices[i].position.x - m_insideBounds.left) / m_insideBounds.width
+        const float xratio      = m_insideBounds.size.x > 0
```
vittorioromeo, line 270: Sucks that we cannot express this as:

```cpp
const auto ratio = (m_vertices[i].position - m_insideBounds.position) / m_insideBounds.size;
```

Maybe this suggests that `m_insideBounds.size` should not be zero here, or that we should check that before entering the loop? Haven't thought about this deeply.

src/SFML/Graphics/Sprite.cpp
```diff
@@ -111,8 +111,8 @@ const Color& Sprite::getColor() const
 ////////////////////////////////////////////////////////////
 FloatRect Sprite::getLocalBounds() const
 {
-    const auto width  = static_cast<float>(std::abs(m_textureRect.width));
-    const auto height = static_cast<float>(std::abs(m_textureRect.height));
+    const auto width  = static_cast<float>(std::abs(m_textureRect.size.x));
```
vittorioromeo, line 114: It might be nice to have `.compAbs()` function on `Vector2`:

```cpp
return {{0.f, 0.f}, Vector2f(m_textureRect.size.compAbs())};
```

src/SFML/Graphics/Text.cpp
```diff
@@ -65,22 +65,19 @@ void addGlyphQuad(sf::VertexArray& vertices, sf::Vector2f position, const sf::Co
...
-    vertices.append({{position.x + right - italicShear * top, position.y + top}, color, {u2, v1}});
-    vertices.append({{position.x + left - italicShear * bottom, position.y + bottom}, color, {u1, v2}});
-    vertices.append({{position.x + left - italicShear * bottom, position.y + bottom}, color, {u1, v2}});
-    vertices.append({{position.x + right - italicShear * top, position.y + top}, color, {u2, v1}});
-    vertices.append({{position.x + right - italicShear * bottom, position.y + bottom}, color, {u2, v2}});
+    const sf::Vector2f p1 = glyph.bounds.position - sf::Vector2f(padding, padding);
+    const sf::Vector2f p2 = glyph.bounds.position + glyph.bounds.size + sf::Vector2f(padding, padding);
+
+    const sf::Vector2f uv1 = sf::Vector2f(glyph.textureRect.position) - sf::Vector2f(padding, padding);
+    const sf::Vector2f uv2 = sf::Vector2f(glyph.textureRect.position + glyph.textureRect.size) +
```
vittorioromeo, line 72:
```suggestion
    const auto uv1 = sf::Vector2f(glyph.textureRect.position) - sf::Vector2f(padding, padding);
    const auto uv2 = sf::Vector2f(glyph.textureRect.position + glyph.textureRect.size) +
```

src/SFML/Graphics/Text.cpp
```diff
@@ -65,22 +65,19 @@ void addGlyphQuad(sf::VertexArray& vertices, sf::Vector2f position, const sf::Co
...
-    const float u2 = static_cast<float>(glyph.textureRect.left + glyph.textureRect.width) + padding;
-    const float v2 = static_cast<float>(glyph.textureRect.top + glyph.textureRect.height) + padding;
-
-    vertices.append({{position.x + left - italicShear * top, position.y + top}, color, {u1, v1}});
-    vertices.append({{position.x + right - italicShear * top, position.y + top}, color, {u2, v1}});
-    vertices.append({{position.x + left - italicShear * bottom, position.y + bottom}, color, {u1, v2}});
-    vertices.append({{position.x + left - italicShear * bottom, position.y + bottom}, color, {u1, v2}});
-    vertices.append({{position.x + right - italicShear * top, position.y + top}, color, {u2, v1}});
-    vertices.append({{position.x + right - italicShear * bottom, position.y + bottom}, color, {u2, v2}});
+    const sf::Vector2f p1 = glyph.bounds.position - sf::Vector2f(padding, padding);
```
vittorioromeo, line 68: `sf::Vector2f(padding, padding)` is used everywhere, maybe change the above `float` to a `Vector2f`?

vittorioromeo (commented):
Not sure how I feel about this yet, just a few random comments after a first skim.

eXpl0it3r:

ChrisThrasher:
I really like this idea. Conceptually a `Rect` has always been a position and a size. For a long time we've agreed it's better to represent positions and sizes as vectors since that is more expressive and lends itself to better composition with other APIs that return and accept vectors for positions and sizes. I see no reason that `Rect` ought to be recused from that same precedent.

Any ergonomic issues introduced by this refactor can be fixed by adding the right new member functions.

include/SFML/Graphics/Rect.hpp
eXpl0it3r: Personally, I wouldn't go with this API. It's really not more difficult to write `position.x` / `position.y` / `size.x` / `size.y` than the respective getters, and using getters would create an asymmetrical API, unless you also add settings, which I would be even more more against.

examples/island/Island.cpp
kimci86: I see more possible simplifications (https://github.com/kimci86/SFML/compare/feature/rect_as_two_vectors...kimci86:SFML:feature/simplify_island_example) that I would rather not add to this PR.

include/SFML/Graphics/Transform.inl
eXpl0it3r: Not an API I'd personally want to see. We define the rectangle in one way and shouldn't need a whole matrix of getter functions, and it would create an asymmetrical API, unless you add setters as well, which I'd be even more against.

include/SFML/Graphics/Rect.inl
kimci86: Done. Is that a convention we would everywhere else?

src/SFML/Graphics/Shape.cpp
kimci86: I just added a commit simplifying this function.

eXpl0it3r:

src/SFML/Graphics/Font.cpp
```diff
@@ -573,42 +573,36 @@ Glyph Font::loadGlyph(std::uint32_t codePoint, unsigned int characterSize, bool
     glyph.lsbDelta = static_cast<int>(face->glyph->lsb_delta);
     glyph.rsbDelta = static_cast<int>(face->glyph->rsb_delta);
 
-    unsigned int width  = bitmap.width;
-    unsigned int height = bitmap.rows;
+    Vector2u size(bitmap.width, bitmap.rows);
 
-    if ((width > 0) && (height > 0))
+    if (size != Vector2u(0, 0))
```
kimci86, line 578: This is wrong, will fix.

include/SFML/Graphics/Rect.hpp
vittorioromeo: Counterpoint: having `.getRight()` and `.getBottom()` would be useful, but it would be weird to have those two and not left and top.

include/SFML/Graphics/Transform.inl
vittorioromeo: I think getting the corners of a rectangle is a common enough operation to warrant its own helper function. I've used it quite a lot when building UIs or aligning game elements. I also don't think that "getters" are required to have a symmetric "setter", especially when they're computing a value on the fly.

ChrisThrasher (approved):

ChrisThrasher:

src/SFML/Graphics/Font.cpp
```diff
@@ -637,24 +631,21 @@ Glyph Font::loadGlyph(std::uint32_t codePoint, unsigned int characterSize, bool
...
                 }
                 pixels += bitmap.pitch;
             }
         }
 
         // Write the pixels to the texture
-        const unsigned int x = static_cast<unsigned int>(glyph.textureRect.left) - padding;
-        const unsigned int y = static_cast<unsigned int>(glyph.textureRect.top) - padding;
-        const unsigned int w = static_cast<unsigned int>(glyph.textureRect.width) + 2 * padding;
-        const unsigned int h = static_cast<unsigned int>(glyph.textureRect.height) + 2 * padding;
```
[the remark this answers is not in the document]
kimci86: By reading the code above.
In the original code:
- `glyph.textureRect` is first set to `findGlyphRect(page, {width, height})`, which returns a rectangle of the given size (_when there is no error_). At this moment `{width, height}` is `{bitmap.width + 2 * padding, bitmap.rows + 2 * padding}`.
- Then the padding is removed. So `glyph.textureRect` is of size `{bitmap.width, bitmap.rows}`
- So when `w` and `h` are computed with padding, they are set to `{bitmap.width + 2 * padding, bitmap.rows + 2 * padding}`
- Conclusion: `page.texture.update` is called with size `{bitmap.width + 2 * padding, bitmap.rows + 2 * padding}`.

In my suggested changes:
- `size` is initialized to `{bitmap.width, bitmap.rows}`
- Then padding is added, so `size` becomes `{bitmap.width + 2 * padding, bitmap.rows + 2 * padding}`
- Conclusion: `page.texture.update` is also called with size `{bitmap.width + 2 * padding, bitmap.rows + 2 * padding}`.

I decided to pass `size` to the `update` method because `size` is used to define the size of `m_pixelBuffer`, which is what we are supposed to pass to `update`.

BUT, _when there is an error_ in `findGlyphRect` (no space left on the page texture), the previous code would copy a few transparent pixels of `m_pixelBuffer` into `page.texture` whereas my suggested changes would copy the glyph on the top left corner of the page texture.
So I need to fix my code, thank you for bringing this to my attention.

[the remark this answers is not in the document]
kimci86: Instead of reverting my changes to the previous behavior, I think I improved it.

To test it, I added underline style to the tennis example text (to check interaction with the white pixels in the top left corner of the page texture used for underline),
```patch
diff --git a/examples/tennis/Tennis.cpp b/examples/tennis/Tennis.cpp
index 41b05b8e..48a0c96a 100644
--- a/examples/tennis/Tennis.cpp
+++ b/examples/tennis/Tennis.cpp
@@ -95,6 +95,7 @@ int main()
     pauseMessage.setCharacterSize(40);
     pauseMessage.setPosition({170.f, 200.f});
     pauseMessage.setFillColor(sf::Color::White);
+    pauseMessage.setStyle(sf::Text::Style::Underlined);
 
 #ifdef SFML_SYSTEM_IOS
     pauseMessage.setString("Welcome to SFML Tennis!\nTouch the screen to start the game.");
```

and I forced `findGlyphRect` to fail with the following patches.

On master branch: Underline is not showing and characters that could not be copied on the page texture are invisible.
```patch
diff --git a/src/SFML/Graphics/Font.cpp b/src/SFML/Graphics/Font.cpp
index b009d65e..e69a9b49 100644
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -668,6 +668,10 @@ Glyph Font::loadGlyph(std::uint32_t codePoint, unsigned int characterSize, bool
 IntRect Font::findGlyphRect(Page& page, const Vector2u& size) const
 {
+    static int x = 0;
+    if (x++ % 2 == 1)
+        return {{0, 0}, {2, 2}};
+
     // Find the line that fits well the glyph
     Row*  row       = nullptr;
     float bestRatio = 0;
```
main

On this branch: Underline is visible and characters that could not be copied on the page texture are shown in white.
```patch
diff --git a/src/SFML/Graphics/Font.cpp b/src/SFML/Graphics/Font.cpp
index d6408dba..006e2ed5 100644
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -663,6 +663,10 @@ Glyph Font::loadGlyph(std::uint32_t codePoint, unsigned int characterSize, bool
 IntRect Font::findGlyphRect(Page& page, const Vector2u& size) const
 {
+    static int x = 0;
+    if (x++ % 2 == 1)
+        return {{0, 0}, {0, 0}};
+
     // Find the line that fits well the glyph
     Row*  row       = nullptr;
     float bestRatio = 0;
```
branch

Does that look good to you?

kimci86:
Note that I'm keeping separate commits for the review but I'll squash related commits once we are ready to merge.

ChrisThrasher:

src/SFML/Graphics/Font.cpp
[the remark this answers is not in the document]
kimci86: I am saving this for follow-up bugfix PR to keep this one focused.
https://github.com/kimci86/SFML/compare/feature/rect_as_two_vectors...kimci86:SFML:bugfix/handle_findGlyphRect_error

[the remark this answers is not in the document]
ChrisThrasher: Now that you're saving the proper bug fix for a later PR does that mean you need to revert your usage of `size` in the `page.texture.update` call?

ChrisThrasher (approved):

ChrisThrasher:

kimci86:
Simplified history a little bit and rebased

eXpl0it3r:

ChrisThrasher:

kimci86:
Rebased and fixed conflict.

vittorioromeo (approved):

vittorioromeo:

PR #2973, opened 2024-05-04
Added support for user defined sound effect implementations.

binary1248:
Title.

Because I found it a pity that #1708 had to be closed due to the miniaudio rewrite and all of @fallahn's work gone to waste, I went ahead and implemented equivalent (maybe even better?) support for user defined sound effects in the spirit of aforementioned PR.

The example has been accordingly extended with demonstrations of how to use the new API to implement basically any effect one can think of. I am by no means a DSP expert (the example implementations were taken from people far more knowledgeable than me 😁), so maybe someone can improve the example processing code or even add more interesting effects.

All of the action happens when an effect processor is attached to an `sf::SoundSource` using:
```c++
using EffectProcessor = std::function<void(const float*, unsigned int&, float*, unsigned int&, unsigned int)>;

void setEffectProcessor(EffectProcessor effectProcessor);
```
The effect processor is passed the raw audio data from the engine and can manipulate it before sending it back to be output through the audio device. Refer to the `sf::SoundSource` documentation and sound effects example for detailed information on API usage.

### DIFF
--- a/include/SFML/Audio/SoundSource.hpp
+++ b/include/SFML/Audio/SoundSource.hpp
@@ -34,6 +34,8 @@
 #include <SFML/System/Angle.hpp>
 #include <SFML/System/Vector3.hpp>
 
+#include <functional>
+
 
 namespace sf
 {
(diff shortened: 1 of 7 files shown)


ChrisThrasher:
Is there anything we can do to reduce the code duplication between Sound.cpp and SoundStream.cpp?

binary1248:
> Is there anything we can do to reduce the code duplication between Sound.cpp and SoundStream.cpp?

Not really... most of the duplicated code references internal members of `Impl` and since `Impl` is local to each TU trying to refactor out those blocks would just make the code unnecessarily more complex.

binary1248:
Rebased onto master.

src/SFML/Audio/SoundStream.cpp
```diff
@@ -395,6 +507,14 @@ bool SoundStream::getLoop() const
 }
 
 
+////////////////////////////////////////////////////////////
+void SoundStream::setEffectProcessor(EffectProcessor effectProcessor)
+{
+    m_impl->effectProcessor = std::move(effectProcessor);
+    m_impl->connectEffect(!!m_impl->effectProcessor);
+}
```
ChrisThrasher, line 515: How does this API respond to passing in a null `std::function` object? Does that simply disable effect processing?

binary1248: `effectProcessor` is set to `{}` and `connectEffect` is called with `false` resulting in the effect node being removed from the node graph. Without it being in the node graph it is completely bypassed.

include/SFML/Audio/SoundSource.hpp
```diff
@@ -74,6 +76,75 @@ class SFML_AUDIO_API SoundSource : protected AudioResource
...
+    ///
+    /// Because this function is stored by the SoundSource
+    /// object it will be able to be called as long as the
+    /// SoundSource object hasn't yet been destroyed. Make sure
+    /// that any data this function references outlives the
+    /// SoundSource object otherwise use-after-free errors will
+    /// occur.
+    ///
+    ////////////////////////////////////////////////////////////
+    using EffectProcessor = std::function<void(const float*, unsigned int&, float*, unsigned int&, unsigned int)>;
```
vittorioromeo, line 146:
```suggestion
    using EffectProcessor = std::function<void(const float* inputFrames, unsigned int& inputFrameCount, float* outputFrames, unsigned int& outputFrameCount, unsigned int frameChannelCount)>;
```

We can use parameter names in typedefs

src/SFML/Audio/SoundStream.cpp
```diff
@@ -91,6 +92,34 @@ struct SoundStream::Impl
...
+        {
+            err() << "Failed to initialize effect node: " << ma_result_description(result) << std::endl;
+            return;
+        }
+
+        effectNode.impl         = this;
+        effectNode.channelCount = nodeChannelCount;
+
+        // Route the sound through the effect node depending on whether an effect processor is set
+        connectEffect(!!effectProcessor);
```
vittorioromeo, line 121:
```suggestion
        connectEffect(static_cast<bool>(effectProcessor));
```

Personal preference, a bit longer but easier to understand if not familiar with `!!` idiom

src/SFML/Audio/SoundStream.cpp
```diff
@@ -111,7 +140,80 @@ struct SoundStream::Impl
...
+    {
+        // If a processor is set, call it
+        if (effectProcessor)
+        {
+            if (!framesIn)
+                frameCountIn = 0;
+
+            effectProcessor(framesIn ? framesIn[0] : nullptr, frameCountIn, framesOut[0], frameCountOut, effectNode.channelCount);
+
+            return;
```
vittorioromeo, line 161:
```suggestion
            effectProcessor(framesIn ? framesIn[0] : nullptr, frameCountIn, framesOut[0], frameCountOut, effectNode.channelCount);
            return;
```

src/SFML/Audio/SoundStream.cpp
```diff
@@ -395,6 +507,14 @@ bool SoundStream::getLoop() const
 }
 
 
+////////////////////////////////////////////////////////////
+void SoundStream::setEffectProcessor(EffectProcessor effectProcessor)
+{
+    m_impl->effectProcessor = std::move(effectProcessor);
+    m_impl->connectEffect(!!m_impl->effectProcessor);
```
vittorioromeo, line 514:
```suggestion
    m_impl->connectEffect(static_cast<bool>(m_impl->effectProcessor));
```

Personal preference, a bit longer but easier to understand if not familiar with `!!` idiom

vittorioromeo (commented):
Mostly nitpicks

binary1248:
Updated with @vittorioromeo's suggestions.

ChrisThrasher:
Rebased on `master` to inherit the new Audio module tests

ChrisThrasher (approved):

PR #2975, opened 2024-05-05
fix borderless flicker

L0laapk3:
Attempt to fix flickering in borderless mode. Likely solves issue #1284.

**Simple demo code to reproduce the flickering can be found [in this gist](https://gist.github.com/L0laapk3/7af0ba7b5f4f711fcc7874a3a1bfda58).**

On my machine, this results in black flickering everytime the window gains or loses focus. Most obvious when switching focus by clicking with the mouse on a different window on another screen.

Below is my best understanding of the issue, it contains some speculation.

Just like in #1284, I had originally recognised the flickering issue as `sf::Style::None` still using what I understood as some kind of "pseudo-fullscreen" mode (since the flicker is still present, but much shorter than real fullscreen switches). After doing some experimentation with the code, I concluded that this "pseudo-fullscreen" mode is actually what `WS_POPUP` does.

I speculated that the reason `WS_POPUP` was initially added, is that if you try to specify a true borderless style in windows API's `CreateWindow()`, it does not let you and instead gives you some border styles anyways. But specifying `WS_POPUP` blocks this behavior.

From analyzing the source code of the 'borderless gaming' program, which is able to turn almost any windowed game into borderless windows without flickering, they never set anything like `WS_POPUP` either.

Instead of doing that, in my opinion it is better to specify some known style first, and remove it immediately after creating the window.

## Tasks

-   [N/A] Tested on Linux
-   [X] Tested on Windows
-   [N/A] Tested on macOS
-   [N/A] Tested on iOS
-   [N/A] Tested on Android

Tested only on my machine: Windows 11 23H2, 1080 ti driver 551.61, 13900kf.


Bambo-Borris:
I tested this with the template project and pointed it to your fork, using the source below I noticed that the titlebar may not be present but it still attempts to render.

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode({ 200, 200 }), "SFML works!", sf::Style::None);
    sf::CircleShape shape(100.f);
    shape.setOrigin({100.f, 100.f});

    while (window.isOpen()) {
        while (sf::Event event = window.pollEvent()) {
                window.close();

            if (auto mousMove = event.getIf<sf::Event::MouseMoved>()) {
                shape.setPosition(sf::Vector2f(mousMove->position));
            }
        }

        window.clear();
        window.draw(shape);
        window.display();
    }

    return 0;
}
```

The green circle is set to track the mouse position, when I put the mouse at the top left of the window I get the following screenshot

image

Appears as though the region allocated to the titlebar isn't available to be rendered to.

PR #2977, opened 2024-05-05
Work around failing network test on Android x86_64 API 24

ChrisThrasher:
## Description

Emergency fix for CI being broken on `master` currently. I'm not sure how these tests originally passed in the PR that introduced them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,7 +85,7 @@ jobs:
             api: 24
             libcxx: x86_64-linux-android/libc++_shared.so
             emuarch: x86_64
-            emuapi: 34
+            # emuapi: 34 # Removing this causes the tests to not run. This works around an issue that causes the Network module tests to fail.
           type: { name: Release }
         - platform: { name: Android, os: ubuntu-latest }
           config:


eXpl0it3r (approved):

.github/workflows/ci.yml
```diff
@@ -80,12 +80,11 @@ jobs:
         - platform: { name: Android, os: ubuntu-latest }
           config:
             name: x86_64 (API 24)
-            flags: -GNinja -DCMAKE_ANDROID_ARCH_ABI=x86_64 -DCMAKE_SYSTEM_NAME=Android -DCMAKE_SYSTEM_VERSION=24 -DCMAKE_ANDROID_NDK=$ANDROID_NDK_ROOT -DBUILD_SHARED_LIBS=TRUE -DCMAKE_ANDROID_STL_TYPE=c++_shared -DSFML_RUN_DISPLAY_TESTS=OFF
+            flags: -GNinja -DCMAKE_ANDROID_ARCH_ABI=x86_64 -DCMAKE_SYSTEM_NAME=Android -DCMAKE_SYSTEM_VERSION=24 -DCMAKE_ANDROID_NDK=$ANDROID_NDK_ROOT -DBUILD_SHARED_LIBS=TRUE -DCMAKE_ANDROID_STL_TYPE=c++_shared -DSFML_BUILD_TEST_SUITE=OFF
             arch: x86_64
             api: 24
             libcxx: x86_64-linux-android/libc++_shared.so
             emuarch: x86_64
-            emuapi: 34
```
eXpl0it3r, line 88: Did you intentionally remove this part?

ChrisThrasher: Yes. This ensure that we don’t try to run the tests. There’s a conditional that includes `&& emuapi` which checks that this exists before running the tests.

eXpl0it3r: But if you disable the test suite, why do you also need to remove this for a check?

ChrisThrasher: > But if you disable the test suite, why do you also need to remove this for a check?

I first tried disabling the test suite but that only disables compilation. Later in CI there is still a job that attempts to run the tests and it fails when the tests don't get compiled. This PR is as minimal as I can make it as far as I'm aware.

Issue #2976, opened 2024-05-05
Store angles internally as radians


L0laapk3:
### Prerequisite Checklist

- [X] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [X] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [X] I understand [SFML's scope](https://www.sfml-dev.org/contribute.php#general-considerations) and believe the feature fits it

### Describe your feature request here

It has been mildly bothering me that angles are internally stored as degrees instead of radians.

It makes more sense to store them internally in radians, since [thats what hardware is optimized for](https://www.felixcloutier.com/x86/fsin). Rather than [having to convert them to radians all the time](https://github.com/SFML/SFML/blob/2386653bd3df5639b9ae537e671d7c9557a774c1/src/SFML/System/Vector2.cpp#L65), only to [have to convert them back immediately after](https://github.com/SFML/SFML/blob/2386653bd3df5639b9ae537e671d7c9557a774c1/src/SFML/Graphics/Transform.cpp#L40).

Especially now that in 3.0 angles are unit-aware with `sf::Angle`, I don't see any reason to use degrees anymore.

Its a fairly trivial change and I'd be happy to PR it myself, but I suspect this is the kind of thing that needs to be discussed first ;)

### Use Cases

`transform.rotate(vector.angle())`

vittorioromeo:
I've done the work so we can discuss pros/cons:
https://github.com/SFML/SFML/pull/2978

ChrisThrasher:
I'm open to this idea if it means fewer lossy conversions between degrees and radians. The fact that the standard library exclusively uses radians compels me to prefer radians. I can't imagine users noticing or caring what underlying representation we choose.

kimci86:
> I can't imagine users noticing or caring what underlying representation we choose.

Well, no need to imagine, this issue is a proof :smile:

On a more serious note, I agree that this is a good idea.

PR #2978, opened 2024-05-05
Store angles internally as radians

vittorioromeo:
Discuss in #2976

### DIFF
--- a/include/SFML/System/Angle.hpp
+++ b/include/SFML/System/Angle.hpp
@@ -144,20 +144,20 @@ private:
     friend constexpr Angle radians(float angle);
 
     ////////////////////////////////////////////////////////////
-    /// \brief Construct from a number of degrees
+    /// \brief Construct from a number of radians
     ///
     /// This function is internal. To construct angle values,
     /// use sf::radians or sf::degrees instead.
     ///
-    /// \param degrees Angle in degrees
+    /// \param radians Angle in radians
     ///
     ////////////////////////////////////////////////////////////
-    constexpr explicit Angle(float degrees);
+    constexpr explicit Angle(float radians);
 
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    float m_degrees{}; //!< Angle value stored as degrees
+    float m_radians{}; //!< Angle value stored as radians
 };
 
 ////////////////////////////////////////////////////////////
@@ -187,6 +187,7 @@ private:
 ////////////////////////////////////////////////////////////
 /// \relates Angle
 /// \brief Overload of == operator to compare two angle values
+/// \note Does not automatically wrap the angle value
 ///
 /// \param left  Left operand (an angle)
 /// \param right Right operand (an angle)
@@ -199,6 +200,7 @@ private:
 ////////////////////////////////////////////////////////////
 /// \relates Angle
 /// \brief Overload of != operator to compare two angle values
+/// \note Does not automatically wrap the angle value
 ///
 /// \param left  Left operand (an angle)
 /// \param right Right operand (an angle)
@@ -211,6 +213,7 @@ private:
 ////////////////////////////////////////////////////////////
 /// \relates Angle
 /// \brief Overload of < operator to compare two angle values
+/// \note Does not automatically wrap the angle value
 ///
 /// \param left  Left operand (an angle)
 /// \param right Right operand (an angle)
@@ -223,6 +226,7 @@ private:
 ////////////////////////////////////////////////////////////
 /// \relates Angle
 /// \brief Overload of > operator to compare two angle values
+/// \note Does not automatically wrap the angle value
 ///
 /// \param left  Left operand (an angle)
 /// \param right Right operand (an angle)
@@ -235,6 +239,7 @@ private:
 ////////////////////////////////////////////////////////////
 /// \relates Angle
 /// \brief Overload of <= operator to compare two angle values
+/// \note Does not automatically wrap the angle value
 ///
 /// \param left  Left operand (an angle)
 /// \param right Right operand (an angle)
@@ -247,6 +252,7 @@ private:
 ////////////////////////////////////////////////////////////
 /// \relates Angle
 /// \brief Overload of >= operator to compare two angle values
+/// \note Does not automatically wrap the angle value
 ///
 /// \param left  Left operand (an angle)
 /// \param right Right operand (an angle)
--- a/include/SFML/System/Angle.inl
+++ b/include/SFML/System/Angle.inl
@@ -34,16 +34,14 @@ namespace sf
 {
 namespace priv
 {
-constexpr float pi = 3.141592654f;
+constexpr float pi  = 3.141592654f;
+constexpr float tau = pi * 2.f;
 
 constexpr float positiveRemainder(float a, float b)
 {
-    assert(b > 0.0f && "Cannot calculate remainder with non-positive divisor");
+    assert(b > 0.f && "Cannot calculate remainder with non-positive divisor");
     const float val = a - static_cast<float>(static_cast<int>(a / b)) * b;
-    if (val >= 0.f)
-        return val;
-    else
-        return val + b;
+    return val >= 0.f ? val : val + b;
 }
 } // namespace priv
 
@@ -55,33 +53,33 @@ constexpr Angle::Angle() = default;
 ////////////////////////////////////////////////////////////
 constexpr float Angle::asDegrees() const
 {
-    return m_degrees;
+    return m_radians * (180.f / priv::pi);
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr float Angle::asRadians() const
 {
-    return m_degrees * (priv::pi / 180);
+    return m_radians;
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr Angle Angle::wrapSigned() const
 {
-    return degrees(priv::positiveRemainder(m_degrees + 180, 360) - 180);
+    return radians(priv::positiveRemainder(m_radians + priv::pi, priv::tau) - priv::pi);
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr Angle Angle::wrapUnsigned() const
 {
-    return degrees(priv::positiveRemainder(m_degrees, 360));
+    return radians(priv::positiveRemainder(m_radians, priv::tau));
 }
 
 
 ////////////////////////////////////////////////////////////
-constexpr Angle::Angle(float degrees) : m_degrees(degrees)
+constexpr Angle::Angle(float radians) : m_radians(radians)
 {
 }
 
@@ -89,70 +87,70 @@ constexpr Angle::Angle(float degrees) : m_degrees(degrees)
 ////////////////////////////////////////////////////////////
 constexpr Angle degrees(float angle)
 {
-    return Angle(angle);
+    return Angle(angle * (priv::pi / 180.f));
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr Angle radians(float angle)
 {
-    return Angle(angle * (180 / priv::pi));
+    return Angle(angle);
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr bool operator==(Angle left, Angle right)
 {
-    return left.asDegrees() == right.asDegrees();
+    return left.asRadians() == right.asRadians();
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr bool operator!=(Angle left, Angle right)
 {
-    return left.asDegrees() != right.asDegrees();
+    return left.asRadians() != right.asRadians();
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr bool operator<(Angle left, Angle right)
 {
-    return left.asDegrees() < right.asDegrees();
+    return left.asRadians() < right.asRadians();
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr bool operator>(Angle left, Angle right)
 {
-    return left.asDegrees() > right.asDegrees();
+    return left.asRadians() > right.asRadians();
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr bool operator<=(Angle left, Angle right)
 {
-    return left.asDegrees() <= right.asDegrees();
+    return left.asRadians() <= right.asRadians();
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr bool operator>=(Angle left, Angle right)
 {
-    return left.asDegrees() >= right.asDegrees();
+    return left.asRadians() >= right.asRadians();
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr Angle operator-(Angle right)
 {
-    return degrees(-right.asDegrees());
+    return radians(-right.asRadians());
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr Angle operator+(Angle left, Angle right)
 {
-    return degrees(left.asDegrees() + right.asDegrees());
+    return radians(left.asRadians() + right.asRadians());
 }
 
 
@@ -166,7 +164,7 @@ constexpr Angle& operator+=(Angle& left, Angle right)
 ////////////////////////////////////////////////////////////
 constexpr Angle operator-(Angle left, Angle right)
 {
-    return degrees(left.asDegrees() - right.asDegrees());
+    return radians(left.asRadians() - right.asRadians());
 }
 
 
@@ -180,7 +178,7 @@ constexpr Angle& operator-=(Angle& left, Angle right)
 ////////////////////////////////////////////////////////////
 constexpr Angle operator*(Angle left, float right)
 {
-    return degrees(left.asDegrees() * right);
+    return radians(left.asRadians() * right);
 }
 
 
@@ -201,15 +199,15 @@ constexpr Angle& operator*=(Angle& left, float right)
 ////////////////////////////////////////////////////////////
 constexpr Angle operator/(Angle left, float right)
 {
-    assert(right != 0 && "Angle::operator/ cannot divide by 0");
-    return degrees(left.asDegrees() / right);
+    assert(right != 0.f && "Angle::operator/ cannot divide by 0");
+    return radians(left.asRadians() / right);
 }
 
 
 ////////////////////////////////////////////////////////////
 constexpr Angle& operator/=(Angle& left, float right)
 {
-    assert(right != 0 && "Angle::operator/= cannot divide by 0");
+    assert(right != 0.f && "Angle::operator/= cannot divide by 0");
     return left = left / right;
 }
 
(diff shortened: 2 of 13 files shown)


Bambo-Borris:
Personally, I can't see a downside to this change. Think it makes sense, and is a nice, small improvement atop the already great work done by adding `sf::Angle` to the API.

include/SFML/System/Angle.hpp
```diff
@@ -507,4 +519,9 @@ namespace Literals
 /// sf::Angle a6 = 3.14_rad; // 3.14 radians
 /// \endcode
 ///
+/// Internally, sf::Angle stores the angle value as radians.
+///
+/// For the best performance and to minimize run-time conversions,
+/// it is recommended to work with radians whenever possible.
+///
```
ChrisThrasher, line 526: I don't necessarily recommend this. I want users to do whatever they find most ergonomic. Besides, it breaks encapsulation to start making recommendations based on implementation details.

src/SFML/Audio/SoundSource.cpp
```diff
@@ -25,6 +25,7 @@
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
+#include "SFML/System/Angle.hpp"
```
ChrisThrasher, line 28:
```suggestion
```

`SoundSource::Cone` already uses `sf::Angle` to define its data layout so we can guarantee it's included in SoundSource.hpp. I'm guessing something like clangd automatically added this as evidence by the quotes instead of angle brackets.

test/Graphics/Transformable.test.cpp
```diff
@@ -34,11 +34,11 @@ TEST_CASE("[Graphics] sf::Transformable")
...
 
         transformable.setRotation(sf::degrees(3.14f));
-        CHECK(transformable.getRotation() == sf::degrees(3.14f));
+        CHECK(transformable.getRotation() == Approx(sf::degrees(3.14f)));
         transformable.setRotation(sf::degrees(540));
-        CHECK(transformable.getRotation() == sf::degrees(180));
+        CHECK(transformable.getRotation() == Approx(sf::degrees(180)));
         transformable.setRotation(sf::degrees(-72));
-        CHECK(transformable.getRotation() == sf::degrees(288));
+        CHECK(transformable.getRotation() == Approx(sf::degrees(288)));
```
ChrisThrasher, line 41: Let's change the `setRotation` arguments to use `sf::radians` so we can avoid adding approximate comparisons.

test/System/Angle.test.cpp
```diff
@@ -202,27 +202,27 @@ TEST_CASE("[System] sf::Angle")
...
-            STATIC_CHECK(sf::degrees(10) * 2.5f == sf::degrees(25));
-            STATIC_CHECK(sf::degrees(100) * 10.0f == sf::degrees(1000));
+            CHECK(sf::degrees(10) * 2.5f == Approx(sf::degrees(25)));
+            CHECK(sf::degrees(100) * 10.0f == Approx(sf::degrees(1000)));
 
             STATIC_CHECK(10 * sf::radians(0) == sf::Angle::Zero);
-            STATIC_CHECK(2.5f * sf::degrees(10) == sf::degrees(25));
-            STATIC_CHECK(10.0f * sf::degrees(100) == sf::degrees(1000));
+            CHECK(2.5f * sf::degrees(10) == Approx(sf::degrees(25)));
+            CHECK(10.0f * sf::degrees(100) == Approx(sf::degrees(1000)));
```
ChrisThrasher, line 218: Anywhere we previously did exact comparison with degrees I'd rather we do exact comparisons with radians rather than start doing approximate comparisons.

test/System/Angle.test.cpp
```diff
@@ -25,33 +25,33 @@ TEST_CASE("[System] sf::Angle")
...
+        CHECK(sf::degrees(1).wrapUnsigned() == Approx(sf::degrees(1)));
+        CHECK(sf::degrees(-1).wrapUnsigned() == Approx(sf::degrees(359)));
+        CHECK(sf::degrees(90).wrapUnsigned() == Approx(sf::degrees(90)));
+        CHECK(sf::degrees(-90).wrapUnsigned() == Approx(sf::degrees(270)));
+        CHECK(sf::degrees(180).wrapUnsigned() == Approx(sf::degrees(180)));
+        CHECK(sf::degrees(-180).wrapUnsigned() == Approx(sf::degrees(180)));
+        CHECK(sf::degrees(360).wrapUnsigned() == Approx(sf::degrees(0)));
+        CHECK(sf::degrees(-360).wrapUnsigned() == Approx(sf::degrees(0)));
+        CHECK(sf::degrees(720).wrapUnsigned() == Approx(sf::degrees(0)));
+        CHECK(sf::degrees(-720).wrapUnsigned() == Approx(sf::degrees(0)));
```
ChrisThrasher, line 54: These changes show how the new API is perhaps less intuitive. Humans have an easier time reasoning about angles in degrees. It's not intuitive to me that 360 degrees wrapped doesn't yield precisely 0 degrees. At least with radians we know that pi cannot be perfectly represented by floating point numbers so it's not confusing why some number close to 2pi doesn't wrap exactly to 0 radians.

ChrisThrasher (commented):
I'm not sure whether I prefer this new implementation. While this change may theoretically lead to fewer internal conversions, it also means using this API in terms of degrees yields results that are less exact than before.

test/System/Angle.test.cpp
L0laapk3: Just wanted to put in my two cents because I disagree with this point.

In my opinion, all of these tests should do approx comparisons either way as with any other float test.

By testing for exact values of float, you are no longer only testing if the API works but also testing for implementation details.

test/System/Angle.test.cpp
L0laapk3: float additions never guaranteed no rounding errors and you will never be able to guarantee that 360 degrees wrapped is exactly 0 degrees. It just happened to be that way for these tests but the result was technically speaking undefined behavior.

floating point rounding errors are a very basic concept in computer programming and I feel like if we need to protect users from them in arbitrary situations then we are doing it wrong.

include/SFML/System/Angle.hpp
L0laapk3: and also, in many cases there is no cost as the compiler will convert constants at compile time.

test/System/Angle.test.cpp
ChrisThrasher: > but the result was technically speaking undefined behavior.

Its we're going to get into the pedantry of this, it is not undefined. IEEE float math is perfectly well defined.

> you will never be able to guarantee that 360 degrees wrapped is exactly 0 degrees.

Actually yes we can guarantee that `360.f` will wrap to `0.f`. Floats do a perfect job representing integers. It's once you start using values in between integers or try to use really big numbers that the "weird" stuff comes into play.

L0laapk3: You are indeed correct that floating point arithmetic rounding errors are well defined, i was not aware of this.

Still, the whole concept of relying on float arithmetic to give an exact result is wild to me.

It is not defined in the sf::Angle documentation anywhere that it will give exact results for some arithmetic operations, therefore in my opinion by not using approx comparisons breaks encapsulation by testing implementation details.

Bambo-Borris:
> I'm not sure whether I prefer this new implementation. While this change may theoretically lead to fewer internal conversions, it also means using this API in terms of degrees yields results that are less exact than before.

This may very well be true, but given the usecases of SFML within games is quite prevalent, the use of radians is preferred in this scenario in my view. Also tying back to the fact the standard library maths functions also prefer radians, SFML internally using radians seems sensible. The fact it will be slightly less exact when you do `sf::degrees(90.f)` is really going to be a question of what scale does it take for a user to notice this inexact angle? I think by the time it becomes noticeably inexact, there are probably other visual problems a user has that far exceed the angle problem. 

If we look to raylib it requires users pass an angle in degrees, but internally always convert that to radians for use in their own maths libraries. GLM will always require radians for rotations. I think the fact we have this nice abstraction means users can still work within the realm of degrees, but internally I think it's sensible for SFML to utilise radians.

L0laapk3:
> Other than the iOS sensor impl I'm not really seeing any less conversions, so does this actually change anything?
> 
> As far as I can tell the only motivation is that some people may think the implementation is unintuitive, but the same could easily be said for using radians. It seems subjective without any technical reason to change

I have linked a few now removed conversions here: https://github.com/SFML/SFML/issues/2976

vittorioromeo:
> Other than the iOS sensor impl I'm not really seeing any less conversions, so does this actually change anything?

Every call to `sf::Angle::asRadians` used to incur a conversion cost prior to this PR.

JonnyPtn:
And every call to `asDegrees` will in cur a conversion cost with this PR too right? So it's an arbitrary decision really

vittorioromeo:
> And every call to `asDegrees` will in cur a conversion cost with this PR too right? So it's an arbitrary decision really

It's really not because most of the programming/gamedev world has already "agreed" on using radians.

JonnyPtn:
I don't really see what relation that has, as SFML doesn't use radians or degrees in it's API now (would have been relevant with 2.x API where degrees was used). This is just about which is used internally

Worth noting that everybody upgrading to SFML 3 will be passing degrees in too due to the existing API

eXpl0it3r:
I believe one of the main benefits is, that for certain internal calculations (anything using `<cmath>` functions), we wouldn't need to convert from degrees to radians. Meaning, every `asRadians()` that is there currently, won't require a conversion now - this doesn't necessarily show up in the PR, as no change is needed.

As for the API usage perspective, I don't think there's a point in making guesses in who and how many are using which API. The point of `sf::Angles` is people are free to choose, thus there will always be one or the other that requires a conversion.

include/SFML/System/Angle.hpp
```diff
@@ -138,26 +138,32 @@ class Angle
     ////////////////////////////////////////////////////////////
     // NOLINTNEXTLINE(readability-identifier-naming)
     static const Angle Zero; //!< Predefined 0 degree angle value
+    // NOLINTNEXTLINE(readability-identifier-naming)
+    static const Angle QuarterCircle; //!< Predefined 90 degree angle value
+    // NOLINTNEXTLINE(readability-identifier-naming)
+    static const Angle HalfCircle; //!< Predefined 180 degree angle value
+    // NOLINTNEXTLINE(readability-identifier-naming)
+    static const Angle FullCircle; //!< Predefined 360 degree angle value
```
eXpl0it3r, line 146: Please always highlight new APIs in the PR description, at best keep them as separate commit or even PR.

vittorioromeo:
> using `<cmath>` functions), we wouldn't need to convert from degrees to radians. Meaning, every `asRadians()` that is there currently, won't require a conversion now - this doesn't necessarily show up in the PR, as no change is needed.
> 
> As for the API usage perspective, I don't think there's a point in making

SFML can be and is often used in conjuction with other libraries such as Vulkan, glm, or even the Standard Library that all expect angles in radians.

include/SFML/System/Angle.hpp
ChrisThrasher: I'd also rather save such additions for a later PR as to keep this PR as minimal as possible

vittorioromeo: OK, will remove for a later PR, I think these are beneficial to avoid using `degrees` and relying on compile-time conversions that might not be done at compile-time

include/SFML/System/Angle.hpp
vittorioromeo: I think it's good to have this but will leave for a future PR to minimize friction in this one

test/Graphics/Transformable.test.cpp
vittorioromeo: I gave it a try and some tests still failed due to small floating point precision errors. I don't think it's worthwhile spending time on avoiding the use of `Approx` in the test, I agree with @L0laapk3's technical arguments that `Approx` is what we want to test for, not an exact value.

vittorioromeo:
Removed the controversial changes

ChrisThrasher:
Did a quick accounting of where we need angles as radians versus where we need angles as degrees to help understand which format gets more use, at least within SFML itself. Is it true we use `.asRadians()` to a similar degree as `.asDegrees()`? Let's read the code.

As radians:
- Trigonometry in 4 example programs (see https://github.com/SFML/SFML/pull/2028 for a better API for common trig with `sf::Angle`)
- Use with miniaudio which expects radians
- Italic text calculations in `sf::Text`
- 2D matrix math in `sf::Transform`
- 2D matrix math in `sf::Transformable`
- 2D matrix math in `sf::View`
- 2D matrix math in `sf::Vector2<T>`

As degrees:
- Reporting gyroscope rates on iOS as degrees per second (See #3001 for why this may actually be a bug)

As it stands now users are getting free construction from degrees and then repeated conversions to radians under the hood since as evident by the source code. Upon changing to a radians-based implementation, construction from degrees is possibly the only place where a conversion between degrees and radians (in either direction) would be necessary. We have to expect users will continue to prefer writing `sf::degrees(45)` rather than `sf::radians(0.78539816339)` since humans have a much easier time reasoning about angles in terms of degrees. Not necessarily saying anything new but I wanted to provide some more evidence for this fact. There is seemingly no place in the code (outside of that one iOS snippet) where a degree-based implementation reduces the number of degree<->radian conversions.

The only place degrees would be superior is if a user has lots of code that performs arithmetic of angles in terms of degrees. This is code that theoretically exists but I have no evidence to believe that users are in fact routinely writing code like `sf::degrees(360) % sf::degrees(15) + sf::degrees(45)` or anything like that so I can't lean too heavily on previous arguments I made to this end.

I'll also mention as a historical fact that `sf::Angle` is currently implemented in terms of degrees because SFML 2 represented all angles as degrees. We did not ever discuss changing the underlying representation of angles in that PR. It was not something that ever crossed our mind if I recall correctly.

---

In conclusion, my hesitation about a radians-based implementation are probably outweighed by the benefits of performing fewer casts between radians and degrees. I think I'm leaning in favor of this PR, certainly enough so that I don't want to block it. If the team is in favor then let's go for it.

ChrisThrasher:

src/SFML/Audio/AudioDevice.hpp
```diff
@@ -27,6 +27,8 @@
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
+#include "SFML/System/Angle.hpp"
```
ChrisThrasher, line 30: Looks like another automatic inclusion from your IDE or language server. We don't need this because Listener.hpp already includes Angle.hpp.

kimci86:
> We have to expect users will continue to prefer writing sf::degrees(45) rather than sf::radians(0.78539816339) since humans have a much easier time reasoning about angles in terms of degrees.

It is worth noting that the conversion can happen at compile time when using hardcoded contant angles or constant expressions because Angle uses constexpr everywhere.

ChrisThrasher (changes requested):
The core of the PR is good.

My one nit is your use of `.f` suffixes. We don't need to do that. `sf::degrees(360)` is perfectly safe and well defined. No need to add `.f` in so many places. I find it no easier to read and it has no technical benefits.

src/SFML/Audio/AudioDevice.hpp
vittorioromeo: Sorry, removing

ChrisThrasher (approved):

include/SFML/System/Angle.inl
```diff
@@ -34,16 +34,14 @@ namespace sf
 {
 namespace priv
 {
-constexpr float pi = 3.141592654f;
+constexpr float pi  = 3.141592654f;
```
L0laapk3, line 37: I wanted to suggest `std::numbers::pi_v` but I guess we are not C++20.

include/SFML/System/Angle.inl
```diff
@@ -55,104 +53,104 @@ constexpr Angle::Angle() = default;
 ////////////////////////////////////////////////////////////
 constexpr float Angle::asDegrees() const
 {
-    return m_degrees;
+    return m_radians * (180.f / priv::pi);
```
L0laapk3, line 56: Note that these constructs [already carry a baked in rounding error](https://godbolt.org/z/ThfWPfM1d).

L0laapk3: Same with calls to `asDegrees()`, I guess it could have a double overload, I'm not sure if we have to care that much? Just pointing it out

L0laapk3 (approved):
just some thoughts

include/SFML/System/Angle.inl
ChrisThrasher: Are you saying that we could get marginally better accuracy if we used a single literal to represent 180 / pi rather than using this particular expression?

include/SFML/System/Angle.inl
ChrisThrasher: https://github.com/search?q=repo%3ASFML%2FSFML%20cxx_std_17&type=code

Alas still using C++17 since C++20 support isn't sufficient in the compilers we're targeting.

eXpl0it3r:
> My one nit is your use of `.f` suffixes. We don't need to do that. `sf::degrees(360)` is perfectly safe and well defined. No need to add `.f` in so many places. I find it no easier to read and it has no technical benefits.

Personally, I like types to be explicit and not rely on implicit conversion, especially since you then have to memorize in which scenario it's not considered a narrowing conversion.

> @eXpl0it3r Any final thoughts?

I too think it's a sensible change, given there's no real reason to stick to degrees when the interface is abstracted and allows us to do fewer conversions internally.

include/SFML/System/Angle.inl
eXpl0it3r: `57.295776` - float (Compiler Explorer)  
`57.2957802` - double (Compiler Explorer)  
`57.295779513082323402053960025447` - actual (Windows calculator)

`0.000003513082323402053960025447` - difference float  
`0.000000686917676597946039974553` - difference double

I guess in theory you could also do the full math in doubles and only covert at the end, but I feel like if you care about such a precision, you may want to implement the mathy bits yourself.

The change makes sense to me, but could also be a separate PR.

ChrisThrasher:
> > My one nit is your use of `.f` suffixes. We don't need to do that. `sf::degrees(360)` is perfectly safe and well defined. No need to add `.f` in so many places. I find it no easier to read and it has no technical benefits.
> 
> Personally, I like types to be explicit and not rely on implicit conversion, especially since you then have to memorize in which scenario it's not considered a narrowing conversion.

You don’t have to memorize anything. The compiler won’t allow narrowing conversions to happen since we have all the conversion warnings turned on.

include/SFML/System/Angle.hpp
```diff
@@ -187,6 +187,7 @@ class Angle
 ////////////////////////////////////////////////////////////
 /// \relates Angle
 /// \brief Overload of == operator to compare two angle values
+/// \note Does not automatically wrap the angle value
```
eXpl0it3r, line 190: Did it do that before?

ChrisThrasher: This behavior did not change in this PR.

eXpl0it3r (approved):

PR #2980, opened 2024-05-06
Prepare API for multimonitor support

L0laapk3:
# Multimonitor support

This is my proposal for a plan to support multiple monitors in the future. To keep a good overview about what changes, I propose to split it up in several PRs:
1. A PR `Monitor` implementation, breaking API changes (**this PR**).
2. PRs populating each of the platforms `MonitorImpl`s.
3. A PR adding a `RenderWindow` constructor overload with a `Monitor` parameter.
4. PRs adding new functionality into `Monitor` such as listing all available monitors and getting information about monitors.

Comments about the general plan as lined out above are welcome, as well as comments on 1. this PR. **However, please try to keep specific comments about PRs 2-4 out of this page if they have no impact on this specific PR.** These comments could possibly be directed at #2136.

# This PR

## Description

- A new `Monitor` API class has been created.
- `VideoMode`s functions have been moved to `Monitor` as they are all monitor dependent.
- `VideoModeImpl`s have been moved to `MonitorImpl`, and this class has been made non-static.
- Only `MonitorImplUnix` has actually been implemented. For the other platforms, placeholders have been created with identical behavior as before the PR to limit the scope. See PR 2 in the plan above.

This PR is related to the issue #2136 

## Tasks

-   [ ] Tested on Linux
-   [ ] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        {
                window.close();
        }

        window.clear();
        window.display();
    }
}
```


ChrisThrasher:
> a plan to support multiple monitors in the future

Do we have any confidence that the team has the expertise and resources to actually implement these features? My concern is that we're building a road to nowhere by creating an API without a plan to ever implement new features with it.

vittorioromeo:
> > a plan to support multiple monitors in the future
> 
> Do we have any confidence that the team has the expertise and resources to actually implement these features? My concern is that we're building a road to nowhere by creating an API without a plan to ever implement new features with it.

Without commenting on the "expertise and resources" part, I'd like to say that it's pretty embarassing to not support multiple monitors in a graphics library in 2024, no matter how "simple" it is.

src/SFML/Window/Unix/MonitorImplX11.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+
+////////////////////////////////////////////////////////////
+std::vector<VideoMode> MonitorImplX11::getFullscreenModes()
+{
+    std::vector<VideoMode> modes;
+
+    // Check if the XRandR extension is present
+    int version = 0;
+    if (XQueryExtension(m_display.get(), "RANDR", &version, &version, &version))
```
vittorioromeo, line 85: Invert `if` and early return to reduce nesting

src/SFML/Window/Unix/MonitorImplX11.cpp
```diff
@@ -0,0 +1,193 @@
...
+    {
+        // Get the current configuration
+        const auto config = X11Ptr<XRRScreenConfiguration>(
+            XRRGetScreenInfo(m_display.get(), RootWindow(m_display.get(), m_screen)));
+        if (config)
```
vittorioromeo, line 90: Invert `if` and early return to reduce nesting

src/SFML/Window/Unix/MonitorImplX11.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+
+////////////////////////////////////////////////////////////
+VideoMode MonitorImplX11::getDesktopMode()
+{
+    VideoMode desktopMode;
+
+    // Check if the XRandR extension is present
+    int version = 0;
+    if (XQueryExtension(m_display.get(), "RANDR", &version, &version, &version))
```
vittorioromeo, line 150: Invert `if` and early return to reduce nesting

src/SFML/Window/Unix/MonitorImplX11.cpp
```diff
@@ -0,0 +1,193 @@
...
+    {
+        // Get the current configuration
+        const auto config = X11Ptr<XRRScreenConfiguration>(
+            XRRGetScreenInfo(m_display.get(), RootWindow(m_display.get(), m_screen)));
+        if (config)
```
vittorioromeo, line 155: Invert `if` and early return to reduce nesting

src/SFML/Window/Unix/MonitorImplX11.cpp
```diff
@@ -0,0 +1,193 @@
...
+                Rotation modeRotation = 0;
+                XRRConfigRotations(config.get(), &modeRotation);
+
+                if (modeRotation == RR_Rotate_90 || modeRotation == RR_Rotate_270)
+                    std::swap(desktopMode.size.x, desktopMode.size.y);
+            }
+        }
+        else
+        {
+            // Failed to get the screen configuration
```
vittorioromeo, line 179: Redundant comment, I think the message below is enough

src/SFML/Window/Unix/MonitorImplX11.cpp
```diff
@@ -0,0 +1,193 @@
...
+            err() << "Failed to retrieve the screen configuration while trying to get the desktop video modes"
+                    << std::endl;
+        }
+    }
+    else
+    {
+        // XRandr extension is not supported: we cannot get the video modes
```
vittorioromeo, line 186: Redundant comment, I think the message below is enough

src/SFML/Window/Unix/MonitorImplX11.cpp
```diff
@@ -0,0 +1,193 @@
...
+        else
+        {
+            // Failed to get the screen configuration
+            err() << "Failed to retrieve the screen configuration while trying to get the supported video modes"
+                    << std::endl;
+        }
+    }
+    else
+    {
+        // XRandr extension is not supported: we cannot get the video modes
```
vittorioromeo, line 135: Redundant comment, I think the message below is enough

src/SFML/Window/Unix/MonitorImplX11.cpp
```diff
@@ -0,0 +1,193 @@
...
+                            if (std::find(modes.begin(), modes.end(), mode) == modes.end())
+                                modes.push_back(mode);
+                        }
+                    }
+                }
+            }
+        }
+        else
+        {
+            // Failed to get the screen configuration
```
vittorioromeo, line 128: Redundant comment, I think the message below is enough

vittorioromeo (commented):
Some minor improvements to our original code

vittorioromeo:
@L0laapk3: can you help me understand the relationship between `VideoMode` and `Monitor` a bit more?

Does `VideoMode` become just a simple struct to represent width, height, and bpp? 

If I create a `RenderWindow` with a specific `Monitor` and `VideoMode`, how do I ensure that the video mode is supported by that monitor? What would happen if it isn't?

src/SFML/Window/Unix/MonitorImplX11.cpp
ChrisThrasher: I love this practice and hope one day we can enable the clang-tidy [`readability-else-after-return`](https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html) to programmatically enforce its usage. I've played with this before in SFML and it results in a _lot_ of code moving towards the left margin and reduces line count as well.

vittorioromeo: Yes, for me it's a no-brainer. As soon as I see the early return, I know that I can focus on the rest of the function. 

With the nesting, my mind needs to keep track of the fact *"hey, an else is coming up eventually"* and it just makes everything harder to understand.

Fun fact: at least 2 professors in my uni who taught C were in the "never early return, only one return" camp. I feel sorry for the students that took that advice to heart without considering that the professor might be wrong.

eXpl0it3r:
> Some minor improvements to our original code

Before we jump too much into reviewing code that was mostly meant as a jump off point for API discussions, lets focus on that first.

L0laapk3:
@vittorioromeo 
> Does `VideoMode` become just a simple struct to represent width, height, and bpp?
The way I changed it now, yes.

> @L0laapk3: can you help me understand the relationship between `VideoMode` and `Monitor` a bit more?
Each monitor has a list of supported fullscreen VideoModes. In fullscreen mode it must be an exact match with one of the supported modes to work afaik. In windowed/borderless mode it doesn't matter (more on this below)

> If I create a `RenderWindow` with a specific `Monitor` and `VideoMode`, how do I ensure that the video mode is supported by that monitor? What would happen if it isn't?
Nothing is strictly stopping you from entering custom numbers and creating your own `VideoMode` object, but if you're using fullscreen, [the `RenderWindow` constructor will validate it first.](https://github.com/SFML/SFML/blob/master/src/SFML/Window/WindowBase.cpp#L359-L383).

As a sidenote, in my opinion the `RenderWindow` constructor `VideoMode` parameter is being abused a little as it is doubling up as the window size in windowed mode. For that reason, maybe the constructor should also accept `Rect` when not in fullscreen mode, or some other solution could be thought up.

src/SFML/Window/Unix/MonitorImplX11.cpp
L0laapk3: While I definitely agree and it is something I noticed as well, to keep this PR on-topic, I have moved the monitor X11 implementation onto its own branch: https://github.com/L0laapk3/SFML/tree/feature/monitor-unix where I will definitely be addressing your comments in the future.

.gitignore
Bambo-Borris, line 1: Pretty sure the SFML policy for stuff like this is that it should be in your own global .gitignore rather than adding new things to SFML's

L0laapk3 (commented):
Looks ready for review to me

ChrisThrasher (commented):
Do you expect any of these `MonitorImpl` classes to have data members in the future? Currently none of them contain any data which means we could simplify this significantly by converting all these classes into simply a collection of free functions.

L0laapk3:
> Do you expect any of these `MonitorImpl` classes to have data members in the future?

@ChrisThrasher yes, they will definitely have data members once they are populated. That is why I went through all the trouble of non-static impl classes :D

For reference, I have an unix implementation prepared here: https://github.com/L0laapk3/SFML/pull/1/files
But that is blocked until this PR is more mature, I do not want to spend a lot of effort into something that I may still be asked to change.

L0laapk3:
whoops, pushed on the wrong branch. reverted.

JonnyPtn:
I'm not sure `Monitor` is the best name here as it's a bit specific, something like `VideoDevice` seems better to me, and fits names we use elsewhere

L0laapk3:
> I'm not sure `Monitor` is the best name here as it's a bit specific, something like `VideoDevice` seems better to me, and fits names we use elsewhere

Im open to name suggestions and not super attached to any names.
Specifically for the `VideoDevice` suggestion, I did google it quickly and looked at the first 8 results, they all referred to camera devices, most of the results were about APIs. So I don't think thats a great idea.

L0laapk3:
See also https://github.com/SFML/SFML/wiki/SD:-Multi-Monitor-Support

ChrisThrasher:
Given the impending release of SFML 3, this PR has been moved back to the SFML 4 project. We do need to improve our story about multi monitor support but it's too late in the development process for this to land in SFML 3.

WarlockD:
I am interested in this.  How does it handle say, monitor getting disconnected and not redetected?  I say this because having a USB type device that is plugged into the usb (not a usb-c alt device mind you but still using usb3 bandwith) is somewhat interest to me.  Obviously I would have to write my own Monitor class but I would like some kind of API to overrite when it gets disconnected or a failure is detected.

L0laapk3:
Due to the glacial progess making this happen in the official repo, I have long since abandoned it. I have a fork with working multimonitor support, but I am no longer interested in leading the PR.

PR #2987, opened 2024-05-08
[CMAKE] disable cpack by default

alexv-ds:
## Description
Added a new option that allows to enable/disable Cpack for SFML.

There should be an option to disable CPack, since it interferes with projects that use CPack and add SFML as a submodule (add_subdirrectory). 

## Tasks

-   [X] Tested on Linux
-   [x] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -465,6 +465,13 @@ endif()
 
 sfml_export_targets()
 
+# configure extras by default when building SFML directly, otherwise hide them
+sfml_set_option(SFML_CONFIGURE_EXTRAS ${PROJECT_IS_TOP_LEVEL} BOOL "TRUE to configure extras, FALSE to ignore them")
+
+if(NOT SFML_CONFIGURE_EXTRAS)
+    return()
+endif()
+
 set(CPACK_PACKAGE_NAME_SUMMARY "Simple and Fast Multimedia Library")
 set(CPACK_PACKAGE_VENDOR "SFML Team")
 set(CPACK_PACKAGE_FILE_NAME "SFML-${PROJECT_VERSION}-${CMAKE_CXX_COMPILER_ID}-${CMAKE_CXX_COMPILER_VERSION}-${CMAKE_BUILD_TYPE}")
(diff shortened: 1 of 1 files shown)


alexv-ds:
> Can you be more specific in how it interferes with your CPack setup?

Cmake Warn message

```console

CMake Warning at C:/Program Files/CMake/share/cmake-3.29/Modules/CPack.cmake:507 (message):
  CPack.cmake has already been included!!
  D:/!AlexV/cpm-cache/sfml/82d37dc09cecdb2dafafc55bf7e9e6a572fa3a7d/SFML/CMakeLists.txt:486 (include)
  
CPack warning: both CPACK_COMPONENTS_ALL and CPACK_MONOLITHIC_INSTALL have been set.
```

This also overwrites the default name to **SFML-3.0.0-MSVC-19.39.33523.0-Debug**

CMakeLists.txt
```diff
@@ -465,25 +465,27 @@ endif()
...
+    set(CPACK_NSIS_DISPLAY_NAME "SFML ${PROJECT_VERSION} (${CMAKE_CXX_COMPILER_ID} ${CMAKE_CXX_COMPILER_VERSION})")
+    set(CPACK_NSIS_CONTACT "team@sfml-dev.org")
+    set(NSIS_IMAGE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/tools/nsis/")
+    string(REGEX REPLACE "/" "\\\\\\\\" NSIS_IMAGE_PATH ${NSIS_IMAGE_PATH})
+    set(CPACK_NSIS_INSTALLER_MUI_ICON_CODE "!define MUI_WELCOMEFINISHPAGE_BITMAP \\\"${NSIS_IMAGE_PATH}sidebar.bmp\\\"\n!define MUI_HEADERIMAGE_BITMAP \\\"${NSIS_IMAGE_PATH}header.bmp\\\"\n!define MUI_ICON \\\"${NSIS_IMAGE_PATH}sfml.ico\\\"")
+    include(CPack)
+endif()
 
 # configure extras by default when building SFML directly, otherwise hide them
 sfml_set_option(SFML_CONFIGURE_EXTRAS ${PROJECT_IS_TOP_LEVEL} BOOL "TRUE to configure extras, FALSE to ignore them")
```
ChrisThrasher, line 491: If we end up doing this, the better solution is to hide this behind `SFML_CONFIGURE_EXTRAS` where we already have lots of extra CMake things that `FetchContent`/`add_subdirectory` users ought not see. SFML already has many configuration options and I don't want to add more.

alexv-ds: good idea

ChrisThrasher:

ChrisThrasher:

ChrisThrasher (approved):
I see no reason that someone consuming SFML via FetchContent or `add_subdirectory` ought to inherit our CPack configuration so it makes sense we hide that from them. Thanks for the PR.

Issue #1184, opened 2017-01-19
Window::waitEvent does not return when window is closed


norgor:
# Description
Window::waitEvent does not return after Window::close is called, which causes the eventual dedicated event thread to wait for an event that will never arrive.

# Solution
    bool WindowImpl::popEvent(Event& event, bool block)
    {
    ...
            // In blocking mode, we must process events until one is triggered
            if (block)
            {
                // Here we use a manual wait loop instead of the optimized
                // wait-event provided by the OS, so that we don't skip joystick
                // events (which require polling)
                while (m_events.empty())
                {
                    sleep(milliseconds(10));
    		
    		//Check if window is open
    
                    processJoystickEvents();
                    processSensorEvents();
                    processEvents();
                }
            }
        }
    ...
    }

eXpl0it3r:
We like to discuss unconfirmed issues and design questions (like this one here) on the [forum](http://en.sfml-dev.org/forums/) as stated in the [contribution guidelines](http://www.sfml-dev.org/contribute.php).

While I at first thought that this might need adjustments to the [documented](http://www.sfml-dev.org/documentation/2.4.1/classsf_1_1Window.php#aaf02ab64fbc1d374eef3696df54137bc) or adding your suggested `// Check if window is open`, I wonder one a second thought, if this is not intentional.

Sure, if you close the window, ignore the close event and then call `waitEvent` again, it's essentially a programming error, as long as the function returns again after you create a new window (this needs to be tested!).
Similar situation can be constructed for `pollEvent`: If you use your own game-loop condition and don't set the flag to false when the close event gets triggered, then you'll end up with an infinite loop as well. The only difference is, that you can control this loop directly (user code), while for `waitEvent` the loop occurs within SFML (library code) itself.

**My conclusion**

It's working as intended (test for the window recreation situation needed) and if you get stuck in an infinite loop, it's a programming error that can easily be prevented.

norgor:
AFAIK closing a game using an in-game button and calling only Window::close will not cause the Closed event to be pushed into the event queue. The thread used to wait for events will therefore be stuck in an infinite loop unable to exit gracefully.

LaurentGomila:
SFML classes are not thread-safe (and I hope it is documented properly). So you have to take care of concurrent access, such as closing a window from one thread while the other thread is processing events.

[...]

PR #2992, opened 2024-05-11
Remove default empty state of `sf::Event`

vittorioromeo:
Proof of concept right now as I don't have access to a compiler, but I think this is a superior design compared to #2991. 

Removing the empty state from event results in a much more type-safe API. Events cannot be default-constructed anymore, and APIs that poll events now return an optional event, making it clear that an empty state can be present without polluting usages of events that cannot be in such an empty state.

@ChrisThrasher: feel free to continue working on this branch to adjust the usages of event in tests and examples.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -57,9 +57,9 @@ body:
 
                   while (window.isOpen())
                   {
-                      while (const auto event = window.pollEvent())
+                      while (const std::optional event = window.pollEvent())
                       {
-                          if (event.is<sf::Event::Closed>())
+                          if (event->is<sf::Event::Closed>())
                               window.close();
                       }
 
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -54,9 +54,9 @@ body:
 
                   while (window.isOpen())
                   {
-                      while (const auto event = window.pollEvent())
+                      while (const std::optional event = window.pollEvent())
                       {
-                          if (event.is<sf::Event::Closed>())
+                          if (event->is<sf::Event::Closed>())
                               window.close();
                       }
 
--- a/examples/raw_input/RawInput.cpp
+++ b/examples/raw_input/RawInput.cpp
@@ -34,9 +34,9 @@ int main()
 
     while (window.isOpen())
     {
-        while (const auto event = window.pollEvent())
+        while (const std::optional event = window.pollEvent())
         {
-            if (event.is<sf::Event::Closed>())
+            if (event->is<sf::Event::Closed>())
             {
                 window.close();
                 break;
@@ -45,10 +45,10 @@ int main()
             static const auto vec2ToString = [](const sf::Vector2i& vec2)
             { return '(' + std::to_string(vec2.x) + ", " + std::to_string(vec2.y) + ')'; };
 
-            if (const auto* const mouseMoved = event.getIf<sf::Event::MouseMoved>())
+            if (const auto* const mouseMoved = event->getIf<sf::Event::MouseMoved>())
                 mousePosition.setString("Mouse Position: " + vec2ToString(mouseMoved->position));
 
-            if (const auto* const mouseMovedRaw = event.getIf<sf::Event::MouseMovedRaw>())
+            if (const auto* const mouseMovedRaw = event->getIf<sf::Event::MouseMovedRaw>())
             {
                 log.emplace_back("Mouse Movement: " + vec2ToString(mouseMovedRaw->delta));
 
--- a/include/SFML/Window/Event.hpp
+++ b/include/SFML/Window/Event.hpp
@@ -49,14 +49,6 @@ class SFML_WINDOW_API Event
 {
 public:
     ////////////////////////////////////////////////////////////
-    /// \brief Empty event
-    ///
-    ////////////////////////////////////////////////////////////
-    struct Empty
-    {
-    };
-
-    ////////////////////////////////////////////////////////////
     /// \brief Closed event
     ///
     ////////////////////////////////////////////////////////////
@@ -275,14 +267,6 @@ public:
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Default constructor
-    ///
-    /// Sets the event to sf::Event::Empty
-    ///
-    ////////////////////////////////////////////////////////////
-    Event() = default;
-
-    ////////////////////////////////////////////////////////////
     /// \brief Construct from a given sf::Event subtype
     ///
     /// \param t Event subtype
@@ -309,23 +293,11 @@ public:
     template <typename T>
     [[nodiscard]] const T* getIf() const;
 
-    ////////////////////////////////////////////////////////////
-    /// \brief Check if current event type is not `Empty`
-    ///
-    /// \return True if current event type is not `Empty`
-    ///
-    ////////////////////////////////////////////////////////////
-    [[nodiscard]] explicit operator bool() const
-    {
-        return !is<Empty>();
-    }
-
 private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    std::variant<Empty,
-                 Closed,
+    std::variant<Closed,
                  Resized,
                  FocusLost,
                  FocusGained,
@@ -390,19 +362,16 @@ private:
 /// any of the corresponding event data.
 ///
 /// \code
-/// while (const auto event = window.pollEvent())
+/// while (const std::optional event = window.pollEvent())
 /// {
-///     // Request for closing the window
-///     if (event.is<sf::Event::Closed>())
+///     // Window closed or escape key pressed: exit
+///     if (event->is<sf::Event::Closed>() ||
+///         (event->is<sf::Event::KeyPressed>() &&
+///          event->getIf<sf::Event::KeyPressed>()->code == sf::Keyboard::Key::Escape))
 ///         window.close();
 ///
-///     // The escape key was pressed
-///     if (const auto* keyPressed = event.getIf<sf::Event::KeyPressed>())
-///         if (keyPressed->code == sf::Keyboard::Key::Escape)
-///             window.close();
-///
 ///     // The window was resized
-///     if (const auto* resized = event.getIf<sf::Event::Resized>())
+///     if (const auto* resized = event->getIf<sf::Event::Resized>())
 ///         doSomethingWithTheNewSize(resized->size);
 ///
 ///     // etc ...
--- a/include/SFML/Window/WindowBase.hpp
+++ b/include/SFML/Window/WindowBase.hpp
@@ -181,36 +181,36 @@ public:
     /// \brief Pop the next event from the front of the FIFO event queue, if any, and return it
     ///
     /// This function is not blocking: if there's no pending event then
-    /// it will return an empty event. Note that more than one event
+    /// it will return a `std::nullopt`. Note that more than one event
     /// may be present in the event queue, thus you should always call
     /// this function in a loop to make sure that you process every
     /// pending event.
     /// \code
-    /// while (const auto event = window.pollEvent())
+    /// while (const std::optional event = window.pollEvent())
     /// {
     ///    // process event...
     /// }
     /// \endcode
     ///
-    /// \return The event; will be `Empty` (convertible to `false`) if no events are pending
+    /// \return The event, otherwise `std::nullopt` if no events are pending
     ///
     /// \see waitEvent
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] Event pollEvent();
+    [[nodiscard]] std::optional<Event> pollEvent();
 
     ////////////////////////////////////////////////////////////
     /// \brief Wait for an event and return it
     ///
     /// This function is blocking: if there's no pending event then
     /// it will wait until an event is received or until the provided
-    /// timeout elapses. After this function returns if no error nor
-    /// timeout occurred, the returned event will not be empty.
+    /// timeout elapses. Only if an error or a timeout occurs the
+    /// returned event will be `std::nullopt`.
     /// This function is typically used when you have a thread that is
     /// dedicated to events handling: you want to make this thread sleep
     /// as long as no new event is received.
     /// \code
-    /// if (const auto event = window.waitEvent())
+    /// while (const std::optional event = window.waitEvent())
     /// {
     ///    // process event...
     /// }
@@ -218,12 +218,12 @@ public:
     ///
     /// \param timeout Maximum time to wait (`Time::Zero` for infinite)
     ///
-    /// \return The event; will be `Empty` (convertible to `false`) on timeout or if window was closed
+    /// \return The event, otherwise `std::nullopt` on timeout or if window was closed
     ///
     /// \see pollEvent
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] Event waitEvent(Time timeout = Time::Zero);
+    [[nodiscard]] std::optional<Event> waitEvent(Time timeout = Time::Zero);
 
     ////////////////////////////////////////////////////////////
     /// \brief Get the position of the window
@@ -561,10 +561,10 @@ private:
 /// while (window.isOpen())
 /// {
 ///    // Event processing
-///    while (const auto event = window.pollEvent())
+///    while (const std::optional event = window.pollEvent())
 ///    {
 ///        // Request for closing the window
-///        if (event.is<sf::Event::Closed>())
+///        if (event->is<sf::Event::Closed>())
 ///            window.close();
 ///    }
 ///
--- a/src/SFML/Window/Android/WindowImplAndroid.cpp
+++ b/src/SFML/Window/Android/WindowImplAndroid.cpp
@@ -462,28 +462,20 @@ int WindowImplAndroid::processKeyEvent(AInputEvent* inputEvent, ActivityStates& 
 ////////////////////////////////////////////////////////////
 int WindowImplAndroid::processMotionEvent(AInputEvent* inputEvent, ActivityStates& states)
 {
-    const std::int32_t device = AInputEvent_getSource(inputEvent);
-
-    Event event;
-
-    if (device == AINPUT_SOURCE_MOUSE)
-        event = Event::MouseMoved{};
-    else if (static_cast<std::uint32_t>(device) & AINPUT_SOURCE_TOUCHSCREEN)
-        event = Event::TouchMoved{};
-
-    const std::size_t pointerCount = AMotionEvent_getPointerCount(inputEvent);
+    const std::int32_t device       = AInputEvent_getSource(inputEvent);
+    const std::size_t  pointerCount = AMotionEvent_getPointerCount(inputEvent);
 
     for (std::size_t p = 0; p < pointerCount; ++p)
     {
         const std::int32_t id = AMotionEvent_getPointerId(inputEvent, p);
 
-        int x = static_cast<int>(AMotionEvent_getX(inputEvent, p));
-        int y = static_cast<int>(AMotionEvent_getY(inputEvent, p));
+        const int x = static_cast<int>(AMotionEvent_getX(inputEvent, p));
+        const int y = static_cast<int>(AMotionEvent_getY(inputEvent, p));
 
         if (device == AINPUT_SOURCE_MOUSE)
         {
             const Event::MouseMoved mouseMoved{{x, y}};
-            event = mouseMoved;
+            forwardEvent(mouseMoved);
 
             states.mousePosition = mouseMoved.position;
         }
@@ -493,13 +485,12 @@ int WindowImplAndroid::processMotionEvent(AInputEvent* inputEvent, ActivityState
                 continue;
 
             const Event::TouchMoved touchMoved{static_cast<unsigned int>(id), {x, y}};
-            event = touchMoved;
+            forwardEvent(touchMoved);
 
             states.touchEvents[id] = touchMoved.position;
         }
-
-        forwardEvent(event);
     }
+
     return 1;
 }
 
@@ -517,45 +508,42 @@ int WindowImplAndroid::processPointerEvent(bool isDown, AInputEvent* inputEvent,
     int x = static_cast<int>(AMotionEvent_getX(inputEvent, index));
     int y = static_cast<int>(AMotionEvent_getY(inputEvent, index));
 
-    Event event;
-
     if (isDown)
     {
         if (device == AINPUT_SOURCE_MOUSE)
         {
-            event = Event::MouseButtonPressed{button, {x, y}};
-
             if (id >= 0 && id < static_cast<int>(Mouse::ButtonCount))
                 states.isButtonPressed[button] = true;
+
+            forwardEvent(Event::MouseButtonPressed{button, {x, y}});
         }
         else if (static_cast<unsigned int>(device) & AINPUT_SOURCE_TOUCHSCREEN)
         {
             Event::TouchBegan touchBegan;
             touchBegan.finger   = static_cast<unsigned int>(id);
             touchBegan.position = {x, y};
-            event               = touchBegan;
 
             states.touchEvents[id] = touchBegan.position;
+
+            forwardEvent(touchBegan);
         }
     }
     else
     {
         if (device == AINPUT_SOURCE_MOUSE)
         {
-            event = Event::MouseButtonReleased{button, {x, y}};
-
             if (id >= 0 && id < static_cast<int>(Mouse::ButtonCount))
                 states.isButtonPressed[button] = false;
+
+            forwardEvent(Event::MouseButtonReleased{button, {x, y}});
         }
         else if (static_cast<std::uint32_t>(device) & AINPUT_SOURCE_TOUCHSCREEN)
         {
-            event = Event::TouchEnded{static_cast<unsigned int>(id), {x, y}};
-
             states.touchEvents.erase(id);
+            forwardEvent(Event::TouchEnded{static_cast<unsigned int>(id), {x, y}});
         }
     }
 
-    forwardEvent(event);
     return 1;
 }
 
--- a/src/SFML/Window/DRM/InputImpl.cpp
+++ b/src/SFML/Window/DRM/InputImpl.cpp
@@ -191,7 +191,7 @@ sf::Keyboard::Key toKey(int code)
 {
     switch (code)
     {
-        // clang-format off
+            // clang-format off
         case KEY_ESC:           return sf::Keyboard::Key::Escape;
         case KEY_1:             return sf::Keyboard::Key::Num1;
         case KEY_2:             return sf::Keyboard::Key::Num2;
@@ -341,7 +341,7 @@ void processSlots()
     }
 }
 
-bool eventProcess(sf::Event& event)
+std::optional<sf::Event> eventProcess()
 {
     const std::lock_guard lock(inputMutex);
 
@@ -353,9 +353,10 @@ bool eventProcess(sf::Event& event)
     static unsigned int doDeferredText = 0;
     if (doDeferredText)
     {
-        event          = sf::Event::TextEntered{doDeferredText};
+        const auto event = sf::Event::TextEntered{doDeferredText};
+
         doDeferredText = 0;
-        return true;
+        return event;
     }
 
     ssize_t bytesRead = 0;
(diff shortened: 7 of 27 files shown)


ChrisThrasher:
I considered this in the past and decided I didn’t like it because it requires excessive `operator->` usage. I find the current `operator bool` API more ergonomic even if it’s less ideologically pure.

ChrisThrasher:
https://github.com/SFML/SFML/compare/master...optional_events

Here's my branch from back in November. I looked at this and didn't think it was worth doing.

L0laapk3:
What are the advantages of this over std::monostate?

vittorioromeo:
> What are the advantages of this over std::monostate?

Monostate is just another symbol for ab empty state, which is a state a valid event should never be in.

As @ChrisThrasher mentioned in his PR, event shouldn't even have an empty state and most importantly such a state should not be public and accessible to users.

The reason why we had an empty state to begin with is because some APIa such as pollEvent need to return immediately and tell the user that no event was polled.

In C++17 we have an idiomatic way of expressing the possible absence of a result: optional. We are using that throughout the library already and using it for events solves the "public empty state but not really" issue and, even better, makes event a type that's always guaranteed to be valid and non empty.

L0laapk3:
> ... since users already should not be using `sf::Event::Empty`.

@ChrisThrasher Can you explain the rationale behind this? What is wrong `sf::Event::Empty`?

vittorioromeo:
> > ... since users already should not be using `sf::Event::Empty`.
> 
> @ChrisThrasher Can you explain the rationale behind this? What is wrong with `sf::Event::Empty`? Is it just about having to implement similar safeties to `std::monostate`?

It's a meaningless implementation detail. Operating systems or windowing systems do not produce empty events. We simply needed a sentinel to signal "hey, you tried to poll events but nothing was available" in user loops.

vittorioromeo:
@eXpl0it3r @binary1248: could you please express your thoughts on this PR? It follows exactly the same philosophy as these already-merged PRs:

- #3020 
- #3021 
- #3022
- #3025
- #3028
- #3030
- #3032

And the same philosophy as the soon-to-be-merged PRs:

- #3027

In fact, I don't understand @ChrisThrasher's concerns given the fact that he's the author of all the PRs mentioned above.

It is true that the user experience gets a tiny bit worse due to the fact that the `->` token must be used compared to the `.` token, and due to the fact that the user needs to be familiar with `std::optional`, but I genuinely believe that is a blessing in disguise. 

Our users' will benefit from the type safety of `std::optional`, and once they learn the pattern of using it, they will appreciate the consistency of such pattern being used everywhere in the library, `sf::Event` included.

There are a few more benefits to this PR:

- We avoid exposing the `sf::Event::Empty` class publicly to the users, which was just an implementation detail;

- We avoid a public `operator bool` conversion on `sf::Event`, which is again an implementation detail exposed to the user, as we were pretty much reimplementing a `std::optional` on our own;

- `sf::Window::waitEvent` now guarantees to return a non-empty event, simplifying code that uses `waitEvent` instead of `pollEvent`;

- The signature of `sf::Window::pollEvent` now clearly shows that an event might not be returned, making it clear how it differs from `waitEvent` and alerting users that they need to check for the empty case;

- This PR opens the door to a very nice declarative visitation-based event processing technique, see #3015 and #3014.

In conclusion, I strongly believe that this PR helps us fulfill our goal of making invalid states unrepresentable in the type system. Not merging it would not only hinder us from reaching that goal, but would undermine the work done by @ChrisThrasher with his previously merged PRs due to a glaring inconsistency between most SFML types and `sf::Event`.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -44,8 +44,8 @@ body:
           label: Steps to reproduce
           description: Tell us how to reproduce this issue and provide a [minimal, complete and verifiable example](https://stackoverflow.com/help/mcve), you can use the template below
           value: |
-              1. 
-              2.
```
ChrisThrasher, line 48: I think these trailing newlines may be intentional

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -57,10 +57,13 @@ body:
 
                   while (window.isOpen())
                   {
-                      while (const auto event = window.pollEvent())
+                      while (const std::optional event = window.pollEvent())
```
ChrisThrasher, line 60: As a point of style I'd prefer we continue to use `auto` here and in all other similar instances

vittorioromeo:
> Can you elaborate on how this PR relates to supporting visitation?

I overstated my point, the same visitation API can be achieved even without this PR. The only difference is that without this PR, we'd have to inject a `sf::Event::Empty` overload ourselves to make the visitor passed to `std::visit` exhaustive.

This PR removes the need for such a "hack", sort of proving that everything fits better together.

Edited my original post.

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -54,10 +54,13 @@ body:
...
                   while (window.isOpen())
                   {
-                      while (const auto event = window.pollEvent())
+                      while (const std::optional event = window.pollEvent())
                       {
-                          if (event.is<sf::Event::Closed>())
+                          if (event->is<sf::Event::Closed>())
+                          {
                               window.close();
+                              break;
```
ChrisThrasher, line 62: I suppose this `break` ensures that event processing ceases when a close is requested? In theory this reduces the latency between clicking the close button the application actually closing. Is that your intent?

include/SFML/Window/WindowBase.hpp
```diff
@@ -180,35 +180,33 @@ class SFML_WINDOW_API WindowBase
...
     /// \code
-    /// while (const auto event = window.pollEvent())
+    /// while (const std::optional<sf::Event> event = window.pollEvent())
     /// {
     ///    // process event...
     /// }
     /// \endcode
     ///
-    /// \return The event; will be `Empty` (convertible to `false`) if no events are pending
+    /// \return The event; will be `std::nullopt` (convertible to `false`) if no events are pending
```
ChrisThrasher, line 194:
```suggestion
    /// \return The event, otherwise `std::nullopt` if no events are pending
```
Tweaking the language to be more in line with how we document other `optional`-y APIs

src/SFML/Window/Android/WindowImplAndroid.cpp
```diff
@@ -462,28 +462,20 @@ int WindowImplAndroid::processKeyEvent(AInputEvent* inputEvent, ActivityStates&
...
 int WindowImplAndroid::processMotionEvent(AInputEvent* inputEvent, ActivityStates& states)
 {
-    const std::int32_t device = AInputEvent_getSource(inputEvent);
-
-    Event event;
-
-    if (device == AINPUT_SOURCE_MOUSE)
-        event = Event::MouseMoved{};
-    else if (static_cast<std::uint32_t>(device) & AINPUT_SOURCE_TOUCHSCREEN)
-        event = Event::TouchMoved{};
```
ChrisThrasher, line 472: How is it that we can get rid of this?

src/SFML/Window/DRM/InputImpl.cpp
```diff
@@ -353,9 +353,9 @@ bool eventProcess(sf::Event& event)
     static unsigned int doDeferredText = 0;
     if (doDeferredText)
     {
-        event          = sf::Event::TextEntered{doDeferredText};
+        auto event     = sf::Event::TextEntered{doDeferredText};
```
ChrisThrasher, line 356: Is this non-`const` to ensure a move can occur?

src/SFML/Window/WindowBase.cpp
```diff
@@ -146,22 +146,29 @@ bool WindowBase::isOpen() const
...
 Event WindowBase::waitEvent()
 {
-    Event event;
-    if (m_impl && (event = m_impl->popEvent(true)))
-        filterEvent(event);
-    return event;
+    assert(m_impl);
+
+    std::optional<sf::Event> event = m_impl->popEvent(true /* blocking */);
+    assert(event.has_value());
```
ChrisThrasher, line 168: Why `assert` here instead of keeping the original conditional? Can we safely assume `popEvent` never returns `std::nullopt`?

src/SFML/Window/WindowImpl.cpp
```diff
@@ -201,14 +201,12 @@ Event WindowImpl::popEvent(bool block)
...
     // Pop the first event of the queue, if it is not empty
-    if (!m_events.empty())
-    {
-        event = m_events.front();
-        m_events.pop();
-    }
+    if (m_events.empty())
+        return std::nullopt;
+
+    sf::Event event = m_events.front();
```
ChrisThrasher, line 208:
```suggestion
    Event event = m_events.front();
```

src/SFML/Window/WindowImpl.hpp
```diff
@@ -128,14 +128,14 @@ class WindowImpl
...
     /// if no event is available: if it is true then the function
     /// doesn't return until a new event is triggered; otherwise it
-    /// returns an empty event to indicate that no event is available.
+    /// returns `std::nullopt` to indicate that no event is available.
     ///
-    /// \param block Use true to block the thread until an event arrives
+    /// \param block Use `true` to block the thread until an event arrives
     ///
-    /// \return The event; can be `Empty` (convertible to `false`) if not blocking
+    /// \return The event; can be `std::nullopt` (convertible to `false`) if not blocking
```
ChrisThrasher, line 135:
```suggestion
    /// \return The event, otherwise `std::nullopt` if not blocking
```

ChrisThrasher (changes requested):
I suppose I can come around to this change. It is technically correct after all. My main argument about ergonomics is hard to support. Hopefully I'm wrong and users won't complain too much about using `->` so much in their event loops. 

Still curious to hear from the rest of the team.

binary1248:
People can't simultaneously praise other hyped up languages for their at times really esoteric syntax and at the same time complain that C++ isn't staying up to date while complaining when ever so often we use already existing language constructs to make the language "less bad" than what its undeserved reputation has many laypeople believe.

Sure I also prefer having to type less, but not at the cost of higher mental load because I have to think of something that wouldn't have been necessary if I would have typed a bit more. I prefer typing over thinking if there was a choice.

I have used SFML for more than a decade at this point, and never have I spent more than 1% of total project time messing with event handling. If anyone spends so much time on event handling that this actually makes a difference they are doing something wrong anyway.

src/SFML/Window/WindowImpl.hpp
vittorioromeo: This suggested change doesn't seem correct, it's not always `std::nullopt` if non blocking, but it *can be* `std::nullopt` only if it's non blocking

src/SFML/Window/WindowBase.cpp
vittorioromeo: Yes, `popEvent` in blocking mode will wait until an event is available. Will add a comment

vittorioromeo, line 168:
```suggestion
    assert(event.has_value()); // `popEvent` in blocking mode guarantees to return an event
```

.github/ISSUE_TEMPLATE/bug_report.yml
vittorioromeo: We're basically doing that, only that we're constraining the type inference to a `std::optional` of some `T`. I think it's worthwhile using `std::optional` here instead of `auto` to clearly suggest that an event might or might not be polled.

I considered two alternatives:

- `auto`: hides too much information, not clear that a `std::optional` is returned;
- `std::optional<sf::Event>`: too verbose and redundant, the information about being an event is already clearly available from the name of the variable `event`.

Using `std::optional` strikes a good balance between conciseness and providing useful information to the reader.

examples/raw_input/RawInput.cpp
```diff
@@ -36,7 +36,7 @@ int main()
     {
         while (const auto event = window.pollEvent())
```
vittorioromeo, line 37:
```suggestion
        while (const std::optional event = window.pollEvent())
```

src/SFML/Window/WindowBase.cpp
ChrisThrasher: Is it overkill to refactor this into two functions where one returns `Event` and the other `std::optional<Event>`?

.github/ISSUE_TEMPLATE/feature_request.yml
vittorioromeo: That's not my intent, but it's a nice side benefit of the change.

The intention is to prevent event processing from causing issues when using `waitEvent`, as `waitEvent` now asserts `m_impl != nullptr`, but `WindowBase::close` immediately invokes `m_impl.reset()`.

We need to call `break;` to avoid accessing a deleted `m_impl` on the next iteration.

Now that I wrote this, it does sound like a subtle trap if the `break;` is forgotten... But the only way I can think of fixing this is to revert `waitEvent` to return a `std::optional<sf::Event>`, which loses a nice API improvement that this PR would provide.

I think a decent alternative would be to do this in `waitEvent`:

```cpp
Event WindowBase::waitEvent()
{
    if (m_impl == nullptr)
    {
        sf::err() << "waitEvent invoked on destroyed window -- did you forget a `break`?" << std::endl;
        std::terminate();
    }

    std::optional<sf::Event> event = m_impl->popEvent(true /* blocking */);
    assert(event.has_value());

    filterEvent(*event);
    return *event;
}
```

src/SFML/Window/Android/WindowImplAndroid.cpp
vittorioromeo: That code did nothing, it's a no-op -- the same conditions are checked below when `forwardEvent` is called. I believe my change is equivalent in behavior to the previous code.

.github/ISSUE_TEMPLATE/feature_request.yml
vittorioromeo: I've added the additional safety measure to `waitEvent` and updated its documentation.

src/SFML/Window/WindowBase.cpp
```diff
@@ -146,22 +147,36 @@ bool WindowBase::isOpen() const
...
-    if (m_impl && (event = m_impl->popEvent(true)))
-        filterEvent(event);
-    return event;
+    if (m_impl == nullptr)
+    {
+        err() << "Fatal error: `waitEvent` invoked on destroyed window -- did you forget a `break` after "
+                 "`window.close()` in your event handling loop?"
+              << std::endl;
+
+        std::abort();
```
vittorioromeo, line 172: @ChrisThrasher: alternatively here we could just return another artificial `sf::Event::Closed`, but I'd rather promote the good practice of breaking after closing the window.

vittorioromeo: @ChrisThrasher: I decided to do that in the end, it's much friendlier for the user compared to termination.

.github/ISSUE_TEMPLATE/bug_report.yml
eXpl0it3r: Trailing spaces are intentional, so you start directly typing

.github/ISSUE_TEMPLATE/bug_report.yml
eXpl0it3r: Before bringing up the same discussion point in every PR, maybe let's leave it as it is right now. Start a separate discussion and the align the code pieces afterwards.

It's not helpful when someone writes code in one way and someone else then goes and keeps changing it afterwards in their way.

.github/ISSUE_TEMPLATE/feature_request.yml
vittorioromeo: > We've found an edge case where `waitEvent` can return an empty state

It's more of a precondition failure -- the precondition is that `waitEvent` should not be called  on a window that's already been closed. Returning a `sf::Event::Closed` is a way of not terminating nor getting UB, but it's not behavior the user should rely on.

Perhaps I can add an `err()` before the return to make it clear that the user is doing something wrong and we're just being nice?

.github/ISSUE_TEMPLATE/bug_report.yml
vittorioromeo: That's fair, but there was a reason I made this change specifically in this PR, as it was now unobvious that `pollEvent` returns an optional with plain `auto'.

.github/ISSUE_TEMPLATE/bug_report.yml
vittorioromeo: Will fix when I get back home

.github/ISSUE_TEMPLATE/feature_request.yml
eXpl0it3r: Actually, this just reminded me of the longstanding issue of `waitEvent` never returning if there are no more events to be produced. For example if you run `waitEvent` in a separate thread and in the other thread close the window. The event thread will continuing spinning as it doesn't get any events to process: https://github.com/orgs/SFML/projects/4?pane=issue&itemId=23482732# / https://github.com/SFML/SFML/issues/1184

Since SFML is anyways doing a spinning while loop, due to Joystick event processing, I'd actually propose to change `waitEvent` to take a custom timeout parameter. Blocking methods should always have a way to cancellation token or timeout option.

In that case you'll have another situation where waitEvent has a "failure case".

The user isn't doing anything wrong, when calling any of the window methods after closing the window. Because you still have a valid object and as such any operation on that object should remain valid for the lifetime of the object.

vittorioromeo: > Actually, this just reminded me of the longstanding issue of `waitEvent` never returning if there no more events are being produced. For example if you run `waitEvent` in a separate thread and in the other thread closes the window. The event thread will continuing spinning as it doesn't get any events to process: https://github.com/orgs/SFML/projects/4?pane=issue&itemId=23482732# / #1184
> 
> Since SFML is anyways doing a spinning while loop, due to Joystick event processing, I'd actually propose to change `waitEvent` to take a custom timeout parameter. Blocking methods should always have a way a cancellation token or timeout option.
> 
> In that case you'll have another situation where waitEvent has a "failure case".
> 
> The user isn't doing anything wrong, when calling any of the window methods after closing the window. Because you still have a valid object and as such any operation on that object should remain valid for the lifetime of the object.

I like the timeout idea, I will work on a PR for that specifically (against master) and then integrate that into this one as they are separate concerns.

vittorioromeo: Created PR for the timeout here: https://github.com/SFML/SFML/pull/3094

ChrisThrasher: > It's more of a precondition failure -- the precondition is that waitEvent should not be called on a window that's already been closed.

In my opinion calling `waitEvent` on a closed window is precondition violation. We should log an error, `assert(false)`, then return `sf::Event::Closed` to ensure behavior is defined in release builds. I still want that function to return an `sf::Event` rather than `std::optional<sf::Event>`.

eXpl0it3r: I strongly disagree. As long as an object was initialized, it's in a valid state. We cannot suddenly have functions semi-fail because we assign random preconditions.

Besides, with the timeout proposal, you actually get an expected failure case.

I'm also against adding `break;` everywhere.

vittorioromeo: My views:

- It's very unfortunate that we lose the nice non-`optional` API for `waitEvent` if we add the timeout, but I think it's a good idea to have one;

- I agree with @ChrisThrasher that calling `waitEvent` on a closed window should be a precondition violation and I agree with his proposed solution -- however if we do add a timeout it would make sense to return `std::nullopt` there as well;

- I disagree with both of you regarding `break;` -- I think that the only sensible thing to do when you close a window is to break out of the event handling loop, and I'd like to promote that practice regardless of whether or not it's strictly necessary for the correctness of the program.

ChrisThrasher: I think adding `break` in all these places is unrelated to the core mission of this PR. If you want to make the case for it, I'd rather that happen in a separate PR because discussing it is stalling the more important API changes. In general it gets back to a personal preference of mine that PRs not try to solve multiple orthogonal problems at once.

As for `waitEvent` changes, does that PR need to be settled first before this PR's review can continue? Help us understand the best order of operations for all the PRs you have recently opened.

vittorioromeo: #3094 was merged in, I've rebased this PR on that. I've also merged in #3096 into this PR, I would like to merge that as a prerequisite but I can extract it out if required. Resolving for now.

include/SFML/Window/Event.hpp
```diff
@@ -275,12 +267,10 @@ class SFML_WINDOW_API Event
...
 
     ////////////////////////////////////////////////////////////
-    /// \brief Default constructor
-    ///
-    /// Sets the event to sf::Event::Empty
+    /// \brief Deleted default constructor
     ///
     ////////////////////////////////////////////////////////////
-    Event() = default;
+    Event() = delete;
```
ChrisThrasher, line 273:
```suggestion
```
This can be removed since it is already implicitly deleted. To ensure we don't regress in this respect we should add a corresponding test.

https://github.com/search?q=repo%3ASFML%2FSFML%20is_default_constructible_v&type=code

I added tests for `!std::is_default_constructible_v<sf::Foo>` for all types for which we removed their default constructor so making this change is in line with existing precedent.

include/SFML/Window/WindowBase.hpp
```diff
@@ -210,20 +210,20 @@ class SFML_WINDOW_API WindowBase
     /// dedicated to events handling: you want to make this thread sleep
     /// as long as no new event is received.
     /// \code
-    /// if (const auto event = window.waitEvent())
+    /// while (const std::optional<sf::Event> event = window.waitEvent())
```
ChrisThrasher, line 213:
```suggestion
    /// while (const std::optional event = window.waitEvent())
```

include/SFML/Window/WindowBase.hpp
```diff
@@ -181,23 +181,23 @@ class SFML_WINDOW_API WindowBase
...
     ///
     /// This function is not blocking: if there's no pending event then
-    /// it will return an empty event. Note that more than one event
+    /// it will return a `std::nullopt`. Note that more than one event
     /// may be present in the event queue, thus you should always call
     /// this function in a loop to make sure that you process every
     /// pending event.
     /// \code
-    /// while (const auto event = window.pollEvent())
+    /// while (const std::optional<sf::Event> event = window.pollEvent())
```
ChrisThrasher, line 189:
```suggestion
    /// while (const std::optional event = window.pollEvent())
```

ChrisThrasher (changes requested):
I think I'm ready to approve this if the `break` additions are removed. I'm still not convinced I want to add them and would rather leave them out of this PR in the meantime.

include/SFML/Window/Event.hpp
vittorioromeo: Hmm, do you actually think it would be better to not have the explicit `= delete`? I think it conveys intent, and we do the same thing for some classes that cannot be copied or moved because of some of their members being non-copyable or non-movable. Technically the parent class doesn't need their copy/move constructors deleted, but we are still explicit about it.

I'd prefer leaving the explicit `= delete` in, but not a big deal for me if you believe it's best to rely on the implicit deletion.

ChrisThrasher: What I'd say is that the tests also convey intent but do so in a more terse manner. The type trait tests already cover many implicitly defined operations. Here's a good example of that: https://github.com/SFML/SFML/pull/2874

vittorioromeo: > What I'd say is that the tests also convey intent but do so in a more terse manner.

The tests do not appear in the public API or documentation of `sf::Event`. I still think it's worth keeping the deleted ctor, but I'd rather remove it and resolve this conversation and potentially re-introduce it later to avoid friction.

vittorioromeo: Removed the ctor and added a test

vittorioromeo:
@ChrisThrasher: I have removed all the extra `break`s and rebased on `master`.

vittorioromeo:

src/SFML/Window/WindowImpl.cpp
```diff
@@ -212,13 +212,13 @@ Event WindowImpl::pollEvent()
...
-Event WindowImpl::popEvent()
+std::optional<Event> WindowImpl::popEvent()
 {
-    Event event;
+    std::optional<Event> event; // Use a single local variable for NRVO
 
     if (!m_events.empty())
     {
-        event = m_events.front();
+        event.emplace(m_events.front());
```
ChrisThrasher, line 221: What does `emplace` offer here over `operator=`?

src/SFML/Window/WindowBase.cpp
```diff
@@ -146,22 +168,22 @@ bool WindowBase::isOpen() const
...
+std::optional<Event> WindowBase::pollEvent()
 {
-    Event event;
-    if (m_impl && (event = m_impl->pollEvent()))
-        filterEvent(event);
-    return event;
+    return WindowsBaseImpl::pollOrWaitEventImpl(
+        m_impl.get(),
+        [this] { return m_impl->pollEvent(); },
+        [this](sf::Event& e) { filterEvent(e); });
```
ChrisThrasher, line 176: I personally think this is overabstraction, but won't demand its removal. I get how it works now but don't love how long it took to understand how it works compared to the naive alternative of having two functions with nearly identical implementations. I don't wish to see more of this pattern going forward.

ChrisThrasher (changes requested):
Thanks for the changes. I'm close to approving this pending a few minor comments. Does @eXpl0it3r have any final thoughts to share about this?

src/SFML/Window/WindowBase.cpp
vittorioromeo: Fair enough, the repetition is not that bad here, so I will remove the template function. I think this pattern is still valuable when there's a larger amount of repetition to ensure things are kept in sync and there are no copy-paste mistakes, hopefully you'll agree with that for future cases where it's more valuable.

include/SFML/Window/WindowBase.hpp
```diff
@@ -210,20 +210,20 @@ class SFML_WINDOW_API WindowBase
     /// dedicated to events handling: you want to make this thread sleep
     /// as long as no new event is received.
     /// \code
-    /// if (const auto event = window.waitEvent())
+    /// while (const std::optional event = window.waitEvent())
```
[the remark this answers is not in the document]
vittorioromeo: Will fix

src/SFML/Window/WindowImpl.cpp
vittorioromeo: It makes it clear that `m_events.front()` returns an `sf::Event` and not a `std::optional<sf::Event>`. If I used `operator=`, it wouldn't be clear from this line of code whether `m_event` was a container of events or of optional events. `.emplace` makes it clear.

ChrisThrasher (approved):
I'd still like to get Exploiter's approval but this LGTM. Thanks for your patience over all the rounds of review and extensive discussion. This should be the last `std::optional` API change which means we'll have officially completed all our v3 API goals! I'm very excited as this marks a huge step towards freezing the v3 API in preparation of shipping.

src/SFML/Window/DRM/InputImpl.cpp
```diff
@@ -510,7 +503,7 @@ bool eventProcess(sf::Event& event)
     if (ready > 0 && FD_ISSET(STDIN_FILENO, &readFDSet))
         bytesRead = read(STDIN_FILENO, &code, 1);
 
-    if ((code == 127) || (code == 8)) // Suppress 127 (DEL) to 8 (BACKSPACE)
+    if ((code == 127) || (code == 8)) // Suppress127 (DEL) to 8 (BACKSPACE)
```
kimci86, line 506: Unintended space removal here

src/SFML/Window/DRM/InputImpl.cpp
```diff
@@ -603,7 +592,7 @@ String getDescription(Keyboard::Scancode /* code */)
 }
 
 
-////////////////////////////////////////////////////////////
+///////////////////////////////////////////////////////////
```
kimci86, line 595: Looks like an unintended change too.

src/SFML/Window/WindowBase.cpp
```diff
@@ -43,6 +43,7 @@
 #include <vector>
 
 #include <cassert>
+#include <cstdlib>
```
kimci86, line 46: Why?

kimci86 (commented):
Too late but here are some minor comments.

src/SFML/Window/DRM/InputImpl.cpp
ChrisThrasher: https://github.com/SFML/SFML/pull/3127

PR #2995, opened 2024-05-12
Switch out minimp3 with the dr_mp3 implementation

### LINKED ISSUE #2215: Certain MP3 files don't keep looping when using sf::Music
## Certain MP3 files not looping

Certain mp3 files are not looping correctly when playing them with sf::Music and looping enabled. One of them doesn't loop at all (issue 1), the other stops looping after the first time (issue 2).

## Your environment

* Windows 10 Pro
* SFML 2.6.X
* VS 17.3.3 Pro, MSVC C++ 20
* Static linking

## Steps to reproduce

```cpp
#include "sfml/Audio/Music.hpp"

#include <Windows.h>

using namespace std;

{
	sf::Music music;
	music.setLoop(true);
	music.setVolume(10);

	// This file stops playing after the first playthrough (issue 1)
	//if (music.openFromFile("SIERRA - UNBROKEN (mp3cut.net).mp3"))
	//	music.play();

	// This file stops playing after the second playthrough (issue 2)
	if (music.openFromFile("Feeling - Tung Wasabi (mp3cut.net).mp3"))
		music.play();
	
	MSG msg = {};
	while (GetMessage(&msg, 0, 0, 0))
	{
		TranslateMessage(&msg);
		DispatchMessage(&msg);
	}

	return 0;
}
```

## Expected behavior

The music should never stop playing.

## Actual behavior

The file "SIERRA - UNBROKEN (mp3cut.net).mp3" stops playing after the first playthrough (issue 1), the file "Feeling - Tung Wasabi (mp3cut.net).mp3" stops playing after the second playthrough (issue 2).

## Analysis
In both cases the problem is that SFML only loops a file if the expected EOF point is reached and minimp3 doesn't reach it, instead it fails reading more data.

Regarding issue 1, I believe the file may report the incorrect number of samples and minimp3 takes it at face value, then just stops streaming too early when EOF is reached. Since I currently needed a solution to this, I made a workaround for myself so SFML continues looping when EOF is reached too early: https://github.com/Biohazard90/SFML/commit/bcd67988848fe76c0e498fdc8778fd38428b8612 Note, this is not a fix I am proposing, it's just to illustrate the issue and how I'm solving it since I cannot fix the mp3 files I'm working with.

Regarding issue 2, minimp3 seems to be incorrectly seeking to 0 and I proposed a fix here already: https://github.com/lieff/minimp3/issues/101

Here are the two mp3 files that have these issues:
[Example mp3 files](https://github.com/SFML/SFML/files/9658753/Release.zip)

eXpl0it3r:
## Description

*There are still some crashes with certain MP3*

Since @lieff seems to have stopped working on [minimp3](https://github.com/lieff/minimp3), it might be wise to switch to dr_mp3, which simplifies our implementation and might potentially have fixes for certain issues (tbd).

I hope this would resolve #2215

## Tasks

-   [ ] Tested on Linux
-   [x] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Audio.hpp>
#include <SFML/Graphics.hpp>

int main()
{
    sf::Music music;
    music.setLoop(true);
    music.setVolume(10);

    if (music.openFromFile("file.mp3"))
        music.play();

    auto window = sf::RenderWindow{ {200, 200}, "MP3 Test" };
    window.setFramerateLimit(30);

    while (window.isOpen())
    {
        for(auto event = sf::Event{}; window.pollEvent(event);)
        {
            {
                window.close();
            }
        }
    }
}
```


include/SFML/Audio/InputSoundFile.hpp
```diff
@@ -66,10 +66,10 @@ class SFML_AUDIO_API InputSoundFile : NonCopyable
     /// The supported audio formats are: WAV (PCM only), OGG/Vorbis, FLAC, MP3.
     /// The supported sample sizes for FLAC and WAV are 8, 16, 24 and 32 bit.
     ///
-    /// Because of minimp3_ex limitation, for MP3 files with big (>16kb) APEv2 tag,
+    /// Because of dr_mp3 limitation, for MP3 files with big (>16kb) APEv2 tag,
     /// it may not be properly removed, tag data will be treated as MP3 data
     /// and there is a low chance of garbage decoded at the end of file.
-    /// See also: https://github.com/lieff/minimp3
+    /// See also: https://github.com/mackron/dr_libs
```
ChrisThrasher, line 72: We ought to double check this limitation or at least err on the side of caution and remove this disclaimer if we're not sure it's true.

ChrisThrasher:
I'm open to this change. It's certainly a liability to use abandoned dependencies. I believe in the past we may have hit sanitizer failures in minimp3 but I could be misremembering.

include/SFML/Audio/InputSoundFile.hpp
eXpl0it3r: I agree. Actually wanted to make a draft PR since I had crashing MP3s, but then it somehow reverted to a normal PR again.

dr_mp3 however just wrapped minimp3's decoding functionality in a nicer API, so I'd assume thr same limitations apply.

eXpl0it3r: dr_mp3 still operates on 16kb buffers, which I assume is where this issue comes from. I tried to reproduce it by writing large APEv2 tags but was unable to get it to produce garbage.

Since ID3 tags are anyways more common than APEv2, at least in my experience, I'd say we don't necessarily have to document this edge case.

eXpl0it3r:
Tested the implementation with various MP3 of different bitrates, mono, stereo, ID3 tags, APEv2 tags, large files, small files. So far everything played back correctly. As such for me the PR is ready for review.

ChrisThrasher:
Are we sure this isn't too big to put into v2?

I got `master` caught up with `2.6.x` (see #3002), merged this branch into that PR branch, fixed the new merge conflicts, then ran the tests and am seeing some test failures. Can you make sense of what is going on here? The lack of tests in v2 makes it harder to land a change like this in v2 versus v3.

https://github.com/SFML/SFML/actions/runs/9073215514

eXpl0it3r:
In the test you've just hard coded the returned samples. There's of course no guarantee that two different decoder implementation return the exact same four samples values at the beginning of a file.

Not sure what the "correct" values should be. Looking at the files in Audacity fully zoomed in, it seems like the first few samples would equal to zero, yet on FLAC it's the same and it too does report values to start with:

image

Either way, I don't think this is too big. We've added audio tests in master only some months ago, so it's not like SFML didn't work before that. And of course the tests need to be adapted. minimp3 is pretty broken in a few ways and there's no fix in sight.

eXpl0it3r:
The author of dr_mp3 mentioned to me that there is one known issue with LAME tags: https://github.com/mackron/dr_libs/issues/263

Not if this warrants a mention somewhere or if we just hope for a fix before the 2.6.2 release 😄

binary1248:
Just in case nobody noticed yet, the author of dr_libs is also the author of miniaudio and all the decoders (mp3, FLAC, wav) are also integrated into miniaudio. This PR wouldn't make sense targeting master since we would just use miniaudio for decoding instead.

eXpl0it3r:
Cherry picking the commit onto master with the tests, we seem to run into the linked issue. The test MP3 has been encoded with LAME and the tags aren't parsed, so the timing and sample count are wrongly reported.

```
  CHECK( inputSoundFile.getSampleCount() == 87'798 )
with expansion:
  91008 (0x16380) == 87798 (0x156f6)

  CHECK( inputSoundFile.getDuration() == sf::microseconds(1'990'884) )
with expansion:
  2063673us == 1990884us
```

eXpl0it3r:
I don't see a movement on dr_libs' side. As such, I'll move this into the backlog for now. Maybe if we choose to do a 2.6.3 one day and it gets fixed, we can pick it up, otherwise it may find its way into SFML 3.x

eXpl0it3r:
dr_mp3 has gotten meta data detection, which fixes the looping issue.

MirrasHue:
That's great. Do you see this being merged in the master branch?

MirrasHue:
I think he updated it last week, maybe all those fixes are in place now?

eXpl0it3r:
dr_mp3 released v0.7.0 but when testing the looping issue, it failed to loop again 🫠 

Will have to investigate...

eXpl0it3r:
I still believe it makes sense to backport this, less because of the originally discussed issue and more, because minimp3 is no longer maintained and cause unwanted crashes with certain files.

Update dr_mp3 to v0.7.3

PR #2999, opened 2024-05-13
Make user aware that findCharactarPos returns top of the line position

dogunbound:
This doesn't need any testing, just documentation changes. I really do think a lot of this confusion may be avoided if we atleast mention this "top of the line" stuff

### DIFF
--- a/include/SFML/Graphics/Text.hpp
+++ b/include/SFML/Graphics/Text.hpp
@@ -372,8 +372,9 @@ public:
     ///
     /// This function computes the visual position of a character
     /// from its index in the string. The returned position is
-    /// in global coordinates (translation, rotation, scale and
-    /// origin are applied).
+    /// at the character's ascender height in global coordinates
+    /// (translation, rotation, scale and origin are applied).
+    ///
     /// If \a index is out of range, the position of the end of
     /// the string is returned.
     ///


ChrisThrasher:

dogunbound:
Linking https://github.com/SFML/SFML/issues/2988# to this

include/SFML/Graphics/Text.hpp
```diff
@@ -372,8 +372,9 @@ class SFML_GRAPHICS_API Text : public Drawable, public Transformable
     ///
     /// This function computes the visual position of a character
     /// from its index in the string. The returned position is
-    /// in global coordinates (translation, rotation, scale and
-    /// origin are applied).
+    /// the global coordinates (translation, rotation, scale and
+    /// origin are applied) at the top of the character's text
+    /// line.
```
eXpl0it3r, line 377: I'd swap it around:

> ...position is at the top of the character's line in global coordinates...

eXpl0it3r: Is "character's text line" descriptive enough? If you hadn't had these discussions over the past few days, would you understand what it means?

dogunbound: I'm thinking about adding some information to the SFML-Website repo with like an image that we can use to define:
"top of the line"
"bottom of the line"
"line height"
"glyph bounds"

dogunbound: ascender height is very clear and googlable way of documenting what it's doing. I'm happy with this imho

PR #3005, opened 2024-05-14
Use PImpl idiom for 'AudioDevice'

vittorioromeo:
Pretty self-explanatory, reduce some transitive includes and hide implementation details.

### DIFF
--- a/src/SFML/Audio/AudioDevice.cpp
+++ b/src/SFML/Audio/AudioDevice.cpp
@@ -28,33 +28,59 @@
 #include <SFML/Audio/AudioDevice.hpp>
 
 #include <SFML/System/Err.hpp>
+#include <SFML/System/Vector3.hpp>
+
+#include <miniaudio.h>
 
 #include <algorithm>
 #include <array>
+#include <memory>
+#include <optional>
 #include <ostream>
-
 
 namespace sf::priv
 {
 ////////////////////////////////////////////////////////////
-AudioDevice::AudioDevice()
+struct AudioDevice::ListenerProperties
+{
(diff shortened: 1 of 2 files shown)


ChrisThrasher:
These are all implementation details so nothing is hidden from users. I don't think adding this extra complexity is worth it.

vittorioromeo:
> These are all implementation details so nothing new is hidden from users.

Internal components that use `AudioDevice` are *users* of `AudioDevice`. Neither `AudioResource` nor `Listener` use `miniaudio.h` or `<optional>` yet they're getting included transitively from `AudioDevice.hpp`. This PR removes those unnecessary transitive dependencies.

> I don't think adding this extra complexity is worth it.

I'd argue that it is worth it. The added complexity is minimal, this is an idiom we have used many times in the library and we're familiar with it. Using PImpl wherever possible and sensible (i.e. we can afford the allocation) moves us closer to ideal header and dependency hygiene.

ChrisThrasher:
Sorry but I'm a firm "no" on this. Header hygiene for the sake of header hygiene isn't enough justification for me to tolerate this increase in complexity.

ChrisThrasher:
https://github.com/SFML/SFML/pull/2749#issuecomment-2041186467

We already discussed this in the original Miniaudio PR. Here is some hard evidence showing that PIMPL in this exact case was not helpful.

vittorioromeo:
We clearly disagree on the added complexity/benefit ratio. I still genuinely think it's worth it *even* if the compilation time stays unchanged, but it's definitely not worth arguing over this. I'll apply these improvements and others to my own fork.

PR #3006, opened 2024-05-14
Remove unimplemented function

ChrisThrasher:
## Description

Removing some dead code. This function is declared in `src/` so this removal is not observable by users.


ChrisThrasher:
Oops my bad this _is_ used. I could have sworn CI passed in a previous run. Nothing to see here!

Issue #3003, opened 2024-05-14
`sf::err()` is not thread-safe


eXpl0it3r:
### Prerequisite Checklist

- [X] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [X] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [X] I am here to report an issue and not to just ask a question or look for help (use the [forum](https://en.sfml-dev.org/forums/index.php#c3) or [Discord](https://discord.gg/nr4X7Fh) instead)

### Describe your issue here

So basically, `sf::Err` is not thread-safe. `thread_local` (or SFML's C++98 `sf::ThreadLocalPtr` equivalent) will make sure there's no race on the `DefaultErrStreamBuf` itself, but not on the underlying `stderr` object.

If I'm not mistaken, C++11 guarantees thread safety for `stdout`, `stderr`, etc. in the sense that no corruption will occur due to multi-threaded access. However, it's still possible that unbuffered characters are interleaved from two places. In C++98, there was no threading model, so I guess no guarantees (might still work in practice).

I referenced a PR which is still incomplete, let's continue the discussion there. Unfortunately, pre-C++11 we cannot use `thread_local`.

### Your Environment

- OS / distro / window manager: 
- SFML version: 
- Compiler / toolchain: 
- Special compiler / CMake flags:

### Steps to reproduce

```cpp
#include <SFML/Graphics/Image.hpp>
#include <thread>

const int num_images = 3;
static std::string image_paths[3] = { "working-image.png", "non-existing-image.png", "main.cpp" };

int main()
{
    const int num_threads = 4;
    std::thread threads[num_threads];

    for(int i = 0; i < num_threads; ++i) {
        threads[i] = std::thread([]() {
            while(true) {
                for(int i = 0; i < num_images; ++i) {
                    sf::Image image;
                    image.loadFromFile(image_paths[i]);
                }
            }
        });
    }

    std::this_thread::sleep_for(std::chrono::seconds(10000000));
}
```

### Expected behavior

Application runs through without issues

### Actual behavior

You can run into crashes:

```
[...]

PR #3008, opened 2024-05-14
Simplify 'sf::Err' and prevent data races

vittorioromeo:
Should fix the crash in #3003 and also simplifies `sf::Err`'s implementation by removing `DefaultErrStreamBuf`. AFAICS, that custom stream buffer was used to make error printing buffered, as `std::cerr` is unbuffered by default.

But I don't think the extra complexity is worth it, we're basically optimizing the performance of... error message printing.

Am I missing something?

### DIFF
--- a/src/SFML/System/Err.cpp
+++ b/src/SFML/System/Err.cpp
@@ -28,84 +28,15 @@
 #include <SFML/System/Err.hpp>
 
 #include <iostream>
-#include <streambuf>
 
-#include <cstdio>
-
-
-namespace
-{
-// This class will be used as the default streambuf of sf::Err,
-// it outputs to stderr by default (to keep the default behavior)
-class DefaultErrStreamBuf : public std::streambuf
-{
-public:
-    DefaultErrStreamBuf()
-    {
-        // Allocate the write buffer
-        constexpr int size   = 64;
-        char*         buffer = new char[size];
-        setp(buffer, buffer + size);
-    }
-
-    ~DefaultErrStreamBuf() override
-    {
-        // Synchronize
-        sync();
-
-        // Delete the write buffer
-        delete[] pbase();
-    }
-
-private:
-    int overflow(int character) override
-    {
-        if ((character != EOF) && (pptr() != epptr()))
-        {
-            // Valid character
-            return sputc(static_cast<char>(character));
-        }
-        else if (character != EOF)
-        {
-            // Not enough space in the buffer: synchronize output and try again
-            sync();
-            return overflow(character);
-        }
-        else
-        {
-            // Invalid character: synchronize output
-            return sync();
-        }
-    }
-
-    int sync() override
-    {
-        // Check if there is something into the write buffer
-        if (pbase() != pptr())
-        {
-            // Print the contents of the write buffer into the standard error output
-            const auto size = static_cast<std::size_t>(pptr() - pbase());
-            std::fwrite(pbase(), 1, size, stderr);
-
-            // Reset the pointer position to the beginning of the write buffer
-            setp(pbase(), epptr());
-        }
-
-        return 0;
-    }
-};
-} // namespace
 
 namespace sf
 {
 ////////////////////////////////////////////////////////////
 std::ostream& err()
 {
-    static DefaultErrStreamBuf buffer;
-    static std::ostream        stream(&buffer);
-
+    thread_local std::ostream stream(std::cerr.rdbuf());
     return stream;
 }
-
 
 } // namespace sf


eXpl0it3r:
It's been implemented like that since the pre git days, unfortunately I don't know how to find FS#10 😄  
But it might be safe to assume, that this might not be needed anymore.

https://github.com/SFML/SFML/commit/76de05af611580c66be1f7e02fd933f0f3b641b8#diff-bd73b69ba08ad3c155e32487a2d553ed69f7ec4cbea6fa3ef7a5fcb5a95ecab8

ChrisThrasher:
> we're basically optimizing the performance of... error message printing.

I'm not stressed about the performance of error paths so if we can fix threading issues _and_ removes lots of code at the cost of error-path performance then I'm cool with that. Seems like an easy win.

ChrisThrasher (commented):
If we simplify `sf::err` down this far, then if I'm understanding correctly the only reason this API exists is to provide a way to redirect the SFML error messages which redirecting all of `std::cerr`. Is that right? I suppose in theory someone may want to send all SFML errors to a different stream or silence them entirely but not affect `std::cerr` which they may still be using for other purposes.

vittorioromeo:
> If we simplify `sf::err` down this far, then if I'm understanding correctly the only reason this API exists is to provide a way to redirect the SFML error messages without redirecting all of `std::cerr`. Is that right? I suppose in theory someone may want to send all SFML errors to a different stream or silence them entirely but not affect `std::cerr` which they may still be using for other purposes.

Yes, I believe this is done in `MainAndroid.cpp`:

    // Redirect error messages to logcat
    sf::err().rdbuf(&states->logcat);

ChrisThrasher:
If `sf::err` is directed at `std::cerr` then it inherits the thread safety guarantees of `std::cerr`. However if we redirect `sf::err` elsewhere, those thread safety guarantees don't necessary continue, right? In that case can we still call `sf::err` thread safe?

Somewhat tangential but I wonder if the `sf::err` tests ought to do some multithreading to better test this use case. It's the one place in SFML where I think multithreading tests may be worthwhile.

vittorioromeo:
> If `sf::err` is directed at `std::cerr` then it inherits the thread safety guarantees of `std::cerr`. However if we redirect `sf::err` elsewhere, those thread safety guarantees don't necessary continue, right? In that case can we still call `sf::err` thread safe?
> 
> Somewhat tangential but I wonder if the `sf::err` tests ought to do some multithreading to better test this use case. It's the one place in SFML where I think multithreading tests may be worthwhile.

The additional safety that this PR brings is that every thread has its own clone of `sf::err`. The thread-safety of the buffer `sf::err` writes to is unchanged from this PR.

Prior to this PR, if the buffer used was thread-safe, we would still get data races because `sf::err` was a global variable shared between threads without any synchronization.

After this PR, if the buffer used is thread-safe, `sf::err` will be safely usable concurrently.

ChrisThrasher:
> The additional safety that this PR brings is that every thread has its own clone of sf::err.

Right, duh, `thread_local` to the rescue.

For users who want to silence or redirect SFML log messages, they now have to do that in each thread individually? Not sure how much crossover there is between those who redirect `sf::err` and those who use `sf::err` indirectly in multiple threads but this may be something we want to explain in the API docs.

eXpl0it3r:
> For users who want to silence or redirect SFML log messages, they now have to do that in each thread individually.

Which you can't really, since for example the audio threads are spawned by SFML and not accessible for redirection?

vittorioromeo:
> > The additional safety that this PR brings is that every thread has its own clone of sf::err.
> 
> Right, duh, `thread_local` to the rescue.
> 
> For users who want to silence or redirect SFML log messages, they now have to do that in each thread individually. Not sure how much crossover there is between those who redirect `sf::err` and those who use `sf::err` concurrently but this may be something we want to explain in the API docs since it's not necessarily intuitive. Prior to this you only needed one call to `sf::err().rdbuf` to redirect all logs in all threads.

That's a very good point! I'll add something in the documentation for sure. The alternative is to have a single global buffer pointer that all thread-local instances share, but then we introduce back complexity that I don't think is needed.

ChrisThrasher:
> Which you can't really, since for example the audio threads are spawned by SFML and not accessible for redirection?

Interesting point. Not all threads are owned and accessible by users.

Vittorio floated the idea of a `std::print`-like interface. If we switched to this then we could have one function for printing messages and a separate function for setting the destination. This 2nd function could be implemented such that all thread-local instances of the underlying `ostream` are ensured to use that destination. Just thinking out load here so pardon me if this idea has some issues.

vittorioromeo:
> > For users who want to silence or redirect SFML log messages, they now have to do that in each thread individually.
> 
> Which you can't really, since for example the audio threads are spawned by SFML and not accessible for redirection?

True.

So we'd have to do something like

```cpp
std::atomic<std::streambuf*> errBufPtr{std::cerr.rdbuf()}; // global, private

void redirectErrStream(std::streambuf* bufPtr); // public

std::ostream& err()
{
    thread_local std::ostream stream;
    stream.rdbuf(errBufPtr.fetch());

    return stream;
}
```

ChrisThrasher:
Screenshot 2024-05-14 at 4 49 47 PM

I'm playing around with some multithreading tests and this is enough to cause a crash deep within `<streambuf>`.

ChrisThrasher:
Screenshot 2024-05-14 at 4 59 07 PM

I wrote some multithreading tests that are running against this PR. These tests _often_ pass but sometimes fail due to certain strings not being present in the final buffer. I suspect this is because we have 5 threads writing to the same `std::ostringstream` without synchronization. Is this a valid use case?

```
[System] sf::err
  Print from multiple threads
-------------------------------------------------------------------------------
/Users/thrasher/Projects/sfml/test/System/Err.test.cpp:46
...............................................................................

/Users/thrasher/Projects/sfml/test/System/Err.test.cpp:63: FAILED:
  CHECK_THAT( stream.str(), Catch::Matchers::ContainsSubstring(std::to_string(i)) )
with expansion:
  "1234" contains: "0"
  ```

vittorioromeo:
I think I can come up with an alternative design that uses a single err stream that locks at the beginning of a streaming chain and unlocks automatically at the end. That would avoid the issues with the buffer redirection. Will do it tomorrow

vittorioromeo:
Closing this in favour of https://github.com/SFML/SFML/pull/3010 which I think is superior because it has fewer caveats.

PR #3011, opened 2024-05-15
Do not use 'sf::err' in examples

vittorioromeo:
`sf::err` is an implementation detail, SFML applications (and examples) should use `std::cerr`


ChrisThrasher (approved):
I disagree about it being an implementation detail but think this is a good change regardless.

vittorioromeo:

vittorioromeo:
> I disagree about it being an implementation detail but think this is a good change regardless.

Users should definitely not be allowed to stream into `sf::err`. They should be able to redirect it, and that's it.

PR #3013, opened 2024-05-15
Refactor usage of 'err()' with miniaudio results

vittorioromeo:
Refactors the commonly used 

```cpp
err() << "...message..." << ma_result_description(result) << std::endl;
```

pattern into 

```cpp
MiniaudioUtils::printErr(result, "...message...");
```

This is not only nice to avoid some repetition and gain conciseness, but I am planning to make future changes to the `err` API that, after this PR, would only need to be applied once for all miniaudio error reporting code.

### DIFF
--- a/src/SFML/Audio/AudioDevice.cpp
+++ b/src/SFML/Audio/AudioDevice.cpp
@@ -26,6 +26,7 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Audio/AudioDevice.hpp>
+#include <SFML/Audio/MiniaudioUtils.hpp>
 
 #include <SFML/System/Err.hpp>
 
@@ -49,7 +50,7 @@ AudioDevice::AudioDevice()
     if (const auto result = ma_log_init(nullptr, &*m_log); result != MA_SUCCESS)
     {
         m_log.reset();
-        err() << "Failed to initialize the audio log: " << ma_result_description(result) << std::endl;
+        MiniaudioUtils::printErr(result, "Failed to initialize the audio log");
         return;
     }
 
@@ -64,7 +65,7 @@ AudioDevice::AudioDevice()
                                                          },
                                                          nullptr));
         result != MA_SUCCESS)
-        err() << "Failed to register audio log callback: " << ma_result_description(result) << std::endl;
+        MiniaudioUtils::printErr(result, "Failed to register audio log callback");
 
     // Create the context
     m_context.emplace();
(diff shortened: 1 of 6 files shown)


eXpl0it3r:
Maybe we can discuss what API changes you're thinking of first.

On its own, I find this less readable, as I have to go and look up what printErr does exactly, plus it potentially removes some flexibility, in case you wanted to stream-out more information. This seems to outweigh the repeated `ma_result_description` call for me.

vittorioromeo:

vittorioromeo:
> Maybe we can discuss what API changes you're thinking of first.

First step would be to hide `sf::err` in a `priv` namespace. Users should not stream to `sf::err` directly, it is an implementation error stream of our library.

With that, an explicit `sf::redirectErrStream` function would be added as part of the public API, allowing users to redirect the output of `sf::priv::err` without actually having to enter the `priv` namespace.

I also want to explore a format-based API in the future for `err`, obviating the whole arcane guard-based approach of #3010.

All of these changes would end up affecting every single usage of `err`.

binary1248:
Technically, miniaudio returns `ma_result` values. If we want to cut down the verbosity of printing them then maybe it would be simpler to just provide an `operator<<(ostream&, ma_result)` until we have a newer way of reporting errors?

vittorioromeo:
> Technically, miniaudio returns `ma_result` values. If we want to cut down the verbosity of printing them then maybe it would be simpler to just provide an `operator<<(ostream&, ma_result)` until we have a newer way of reporting errors?

That would couple us even further to a stream-based API and would not help to make it easier to make changes to `err` without affecting the entire codebase, I don't think it's the right way forward.

binary1248:
I don't think it is a strong argument that this change makes changing `err()` significantly less work-intensive. `err()` usage in audio is just a fraction of the total usage throughout the library. Considering that this kind of an overhaul should not really occur that often (hopefully not again for at least 10 years), optimizing `err()` usage in a single module to save a bit of time for such a rare occurrence doesn't pay off for me, especially since as @eXpl0it3r mentioned you won't be able to determine exactly what is printed by just looking at the site where the error is produced.

I would just bite the bullet, make the `err()` changes, adjust all the necessary bits of code however many they might be and call it a day for the next 10 years.

vittorioromeo:
> I don't think it is a strong argument that this change makes changing `err()` significantly less work-intensive. `err()` usage in audio is just a fraction of the total usage throughout the library. Considering that this kind of an overhaul should not really occur that often (hopefully not again for at least 10 years), optimizing `err()` usage in a single module to save a bit of time for such a rare occurrence doesn't pay off for me, especially since as @eXpl0it3r mentioned you won't be able to determine exactly what is printed by just looking at the site where the error is produced.
> 
> I would just bite the bullet, make the `err()` changes, adjust all the necessary bits of code however many they might be and call it a day for the next 10 years.

I only see that as an added benefit, in my opinion the PR has value regardless of that as a general DRY refactor.

ChrisThrasher:
I'd rather we solidify our final error logging API before making changes like this. I summarized all my thoughts on the matter in [this comment](https://github.com/SFML/SFML/pull/3010#issuecomment-2116295145).

vittorioromeo:
Not worth arguing over this one.

Issue #3014, opened 2024-05-15
Exploration of `sf::Event::visit` use cases


vittorioromeo:
### Prerequisite Checklist

- [X] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [X] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [X] I understand [SFML's scope](https://www.sfml-dev.org/contribute.php#general-considerations) and believe the feature fits it

### Describe your feature request here

In #2990 and event-related PRs, there was some discussion about providing a `sf::Event::visit` function. Would it actually be worth it?

Here are two examples of what could be achieved with such a function. Note that both examples are based on #2992, which I believe it's the right way forward for events anyway.

First of all, the *user* defines this convenience function:

```cpp
void pollEvents(sf::Window& window, auto... lambdas)
{   
    auto visitor = overload(lambdas...);
    while (const std::optional event = window.pollEvent())
        event.visit(visitor);
}
```

Now, example 1:

```cpp
//
//
// CURRENT API 
while (window.isOpen())
{
    while (const std::optional event = window.pollEvent())
    {
        if (event->is<sf::Event::Closed>())
        {
            window.close();
            break;
        }

        if (const auto* keyPressed = event->getIf<sf::Event::KeyPressed>();
            keyPressed && keyPressed->code == sf::Keyboard::Key::Escape)
        {
            window.close();
            break;
        }
    }
}

//
//
// WITH VISITATION
while (window.isOpen())
{
    pollEvents(window, 
        [&](sf::Event::Closed)
        { 
            window.close(); 
        },
        [&](sf::Event::KeyPressed e)
        { 
            if (e.code == sf::Keyboard::Key::Escape)
                window.close();
        }
    );
}
```

Example 2 (from Tennis):

```cpp
//
//
// CURRENT API
while (const auto event = window.pollEvent())
{
    // Window closed or escape key pressed: exit
    if (event.is<sf::Event::Closed>() || (event.is<sf::Event::KeyPressed>() &&
                                            event.getIf<sf::Event::KeyPressed>()->code == sf::Keyboard::Key::Escape))
    {
        window.close();
        break;
    }

    // Space key pressed: play
[...]

PR #3015, opened 2024-05-15
Implement `sf::Event::visit`

vittorioromeo:
See https://github.com/SFML/SFML/issues/3014.

### DIFF
--- a/include/SFML/Window/Event.hpp
+++ b/include/SFML/Window/Event.hpp
@@ -320,6 +320,15 @@ public:
         return !is<Empty>();
     }
 
+    ////////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


ChrisThrasher:
In the original event API there was significant pushback to there being more than one way to process events. Have we moved past that?

vittorioromeo:
> In the original event API there was significant pushback to there being more than one way to process events. Have we moved past that?

My personal gripe was having both a `switch`-based interface and one based on types, because it created duplication of `sf::Event::TYPE` and `sf::Event::Type::TYPE` and provided two ways of achieving the same imperative sort of checking.

I don't have a problem with supporting visitation as it is a type-based interface which is compatible with our new event handling model.

PR #3018, opened 2024-05-16
Allow more GitHub Actions runners to run the audio tests.

binary1248:
Title.

Getting rid of libasound2 allows the audio tests to run on Linux as well. We don't get rid of it when building for Android because Java relies on it and the tests wouldn't be run on the building system anyway.


ChrisThrasher (approved):

eXpl0it3r (approved):

ChrisThrasher:
Screenshot 2024-05-16 at 3 15 32 PM

Weird to see my name on this commit

binary1248:
That's what happens when you just amend stuff without actually creating new commits. 🙄

PR #3019, opened 2024-05-17
Replaced SoundFileReaderWav implementation with miniaudio (dr_)wav decoder

### LINKED ISSUE #1777: Suggestion: Replace default wav decoder implementation with dr_wav
Since SFML switched to its own implementation of wav file loading, it's gotten slower to load big wave files.

I've been overriding the default implementation with one that uses dr_wav, which loads files much faster and supports more exotic wav files.

dr_wav is dual licensed (public domain or MIT-0):
https://github.com/mackron/dr_libs/blob/master/dr_wav.h

I suggest you replace SFML's internal implementation with this one or a similar one.

On my project (DGEngine), the difference in speed is noticeable when loading a few wav files (seconds).

Wave2.h
```cpp
#pragma once

#define DR_WAV_NO_STDIO
#include "dr_wav.h"
#include <SFML/Audio/SoundFileReader.hpp>

// Wave file reader using a different decoder from SFML's default.
class Wave2FileReader : public sf::SoundFileReader
{
private:
	drwav wave{};
	bool isInitialized{ false };

public:
	Wave2FileReader() = default;
	~Wave2FileReader() override;

	static bool check(sf::InputStream& stream_);

	bool open(sf::InputStream& stream_, Info& info) override;
	void seek(sf::Uint64 sampleOffset) override;
	sf::Uint64 read(sf::Int16* samples, sf::Uint64 maxCount) override;
};
```

Wave2.cpp
```cpp
#include "Wave2.h"
#define DR_WAV_NO_STDIO
#define DR_WAV_IMPLEMENTATION
#include "dr_wav.h"
#include <SFML/System/InputStream.hpp>

static size_t readWave2(void* pUserData, void* pBufferOut, size_t bytesToRead)
{
	sf::InputStream* stream = (sf::InputStream*)pUserData;
	if (stream == nullptr)
	{
		return -1;
	}
	return (size_t)stream->read(pBufferOut, bytesToRead);
}

static drwav_bool32 seekWave2(void* pUserData, int offset, drwav_seek_origin origin)
{
	sf::InputStream* stream = (sf::InputStream*)pUserData;
	if (stream == nullptr)
	{
		return DRWAV_FALSE;
	}
	if (origin == drwav_seek_origin_current)
	{
		offset += (int)stream->tell();
	}
	auto ret = stream->seek(offset);
	if (ret >= 0)
	{
		return DRWAV_TRUE;
	}
	return DRWAV_FALSE;
}

Wave2FileReader::~Wave2FileReader()
{
	if (isInitialized == true)
	{
		drwav_uninit(&wave);
		isInitialized = false;
	}
}

bool Wave2FileReader::check(sf::InputStream& stream_)
{
	drwav wave;
	if (!drwav_init(&wave, readWave2, seekWave2, &stream_, nullptr))
	{
		return false;
	}
	drwav_uninit(&wave);
	return true;
}

bool Wave2FileReader::open(sf::InputStream& stream_, Info& info)
{
	if (!drwav_init(&wave, readWave2, seekWave2, &stream_, nullptr))
	{
		return false;
	}
	isInitialized = true;
	info.channelCount = wave.channels;
	info.sampleCount = wave.totalPCMFrameCount * wave.channels;
	info.sampleRate = wave.sampleRate;
	return true;
}

void Wave2FileReader::seek(sf::Uint64 sampleOffset)
{
	if (isInitialized == true)
	{
		drwav_seek_to_pcm_frame(&wave, sampleOffset / wave.channels);
	}
}

sf::Uint64 Wave2FileReader::read(sf::Int16* samples, sf::Uint64 maxCount)
{
	if (isInitialized == true)
	{
		return drwav_read_pcm_frames_s16le(&wave, maxCount / wave.channels, samples) * wave.channels;
	}
	return 0;
}
```

binary1248:
Title.

Closes #1777.

If anyone is wondering why the writer wasn't also replaced, our current implementation supports writing in extensible format which supports embedding channel mapping data, whereas the miniaudio implementation doesn't as far as I can tell. The reasoning behind replacing the decoder (see #1777: slow, doesn't support exotic formats) also doesn't really apply to replacing the writer as well. If we ever get reports about performance issues with the writer it will be easy to optimize.

### DIFF
--- a/src/SFML/Audio/SoundFileReaderWav.cpp
+++ b/src/SFML/Audio/SoundFileReaderWav.cpp
@@ -25,82 +25,64 @@
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
+#include <SFML/Audio/MiniaudioUtils.hpp>
 #include <SFML/Audio/SoundFileReaderWav.hpp>
 
 #include <SFML/System/Err.hpp>
 #include <SFML/System/InputStream.hpp>
-#include <SFML/System/Utils.hpp>
 
+#include <array>
 #include <ostream>
+#include <vector>
 
 #include <cassert>
 #include <cstddef>
-#include <cstring>
 
 
 namespace
 {
-// The following functions read integers as little endian and
-// return them in the host byte order
-
-bool decode(sf::InputStream& stream, std::uint8_t& value)
+ma_result onRead(ma_decoder* decoder, void* buffer, size_t bytesToRead, size_t* bytesRead)
 {
-    return static_cast<std::size_t>(stream.read(&value, sizeof(value))) == sizeof(value);
+    auto*      stream = static_cast<sf::InputStream*>(decoder->pUserData);
+    const auto count  = stream->read(buffer, static_cast<std::int64_t>(bytesToRead));
+
+    if (count < 0)
+        return MA_ERROR;
+
+    *bytesRead = static_cast<size_t>(count);
+    return MA_SUCCESS;
 }
 
-bool decode(sf::InputStream& stream, std::int16_t& value)
+ma_result onSeek(ma_decoder* decoder, ma_int64 byteOffset, ma_seek_origin origin)
 {
-    std::byte bytes[sizeof(value)];
-    if (static_cast<std::size_t>(stream.read(bytes, static_cast<std::int64_t>(sizeof(bytes)))) != sizeof(bytes))
-        return false;
+    auto* stream = static_cast<sf::InputStream*>(decoder->pUserData);
 
-    value = sf::toInteger<std::int16_t>(bytes[0], bytes[1]);
+    switch (origin)
+    {
+        case ma_seek_origin_start:
+        {
+            if (stream->seek(byteOffset) < 0)
+                return MA_ERROR;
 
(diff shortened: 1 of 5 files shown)


ChrisThrasher (approved):

src/SFML/Audio/SoundFileReaderWav.cpp
```diff
@@ -25,414 +25,179 @@
...
-    if (!info)
-        err() << "Failed to open WAV sound file (invalid or unsupported file)" << std::endl;
+    if (ma_decoder_init(&onRead, &onSeek, &stream, &config, &decoder) == MA_SUCCESS)
+    {
+        ma_decoder_uninit(&decoder);
+        return true;
+    }
 
-    return info;
+    return false;
```
eXpl0it3r, line 104: Should we provide a message up on failure?

eXpl0it3r (approved):
Tested it even with some surround audio that our implementation couldn't decode correctly.
Whether it's faster as the issue mentioned, I couldn't tell, but it's certainly less code for us to maintain. 😊

eXpl0it3r:
Thank you for switch the decoder! 🙂

PR #3020, opened 2024-05-18
Remove default empty state of `sf::Font`

ChrisThrasher:
## Description

This is the first step towards replacing the `loadFromFile` family of member functions with factory functions. There is no use case for a default-constructed `sf::Font` so this PR systematically eliminates a category of bug where users forget to or fail to initialize a font.

In our examples I chose to use `.value()` to extract the font from the `std::optional<sf::Font>`. This has the same effect as returning `EXIT_FAILURE` since in both cases the program is immediately terminated with a non-zero error code. I chose to make that same recommendation in the documentation since I think it's the best default choice for users who don't otherwise know how to handle such errors.

This method of error handling was decided upon in https://github.com/SFML/SFML/issues/2139.

Supersedes https://github.com/SFML/SFML/pull/2231 which has accumulated conflicts.

### DIFF
--- a/include/SFML/Graphics/Font.hpp
+++ b/include/SFML/Graphics/Font.hpp
@@ -37,6 +37,7 @@
 
 #include <filesystem>
 #include <memory>
+#include <optional>
 #include <string>
 #include <unordered_map>
 #include <vector>
@@ -87,12 +88,12 @@ public:
     ///
     /// \param filename Path of the font file to load
     ///
-    /// \return True if loading succeeded, false if it failed
+    /// \return Font if loading succeeded, `std::nullopt` if it failed
     ///
     /// \see loadFromMemory, loadFromStream
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool loadFromFile(const std::filesystem::path& filename);
+    [[nodiscard]] static std::optional<Font> loadFromFile(const std::filesystem::path& filename);
 
     ////////////////////////////////////////////////////////////
     /// \brief Load the font from a file in memory
@@ -108,12 +109,12 @@ public:
     /// \param data        Pointer to the file data in memory
     /// \param sizeInBytes Size of the data to load, in bytes
     ///
-    /// \return True if loading succeeded, false if it failed
+    /// \return Font if loading succeeded, `std::nullopt` if it failed
     ///
     /// \see loadFromFile, loadFromStream
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool loadFromMemory(const void* data, std::size_t sizeInBytes);
+    [[nodiscard]] static std::optional<Font> loadFromMemory(const void* data, std::size_t sizeInBytes);
 
     ////////////////////////////////////////////////////////////
     /// \brief Load the font from a custom stream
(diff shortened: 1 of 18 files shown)


include/SFML/Graphics/Font.hpp
```diff
@@ -322,12 +329,6 @@ class SFML_GRAPHICS_API Font
         std::vector<Row> rows;       //!< List containing the position of all the existing rows
     };
 
-    ////////////////////////////////////////////////////////////
-    /// \brief Free all the internal resources
-    ///
-    ////////////////////////////////////////////////////////////
-    void cleanup();
```
ChrisThrasher, line 329: If an API refactor lets us remove manual cleanup functions then we know we're doing something right.

include/SFML/Graphics/Font.hpp
```diff
@@ -288,6 +289,12 @@ class SFML_GRAPHICS_API Font
     bool isSmooth() const;
 
 private:
+    ////////////////////////////////////////////////////////////
+    /// Default constructor
+    ///
+    ////////////////////////////////////////////////////////////
+    Font() = default;
```
vittorioromeo, line 296:
```suggestion
    ////////////////////////////////////////////////////////////
    /// \brief Create a font from font handles and a family name
    ///
    ////////////////////////////////////////////////////////////
    explicit Font(std::shared_ptr<FontHandles>&& fontHandles, std::string&& familyName);
```

src/SFML/Graphics/Font.cpp
```diff
@@ -280,31 +273,32 @@ bool Font::loadFromStream(InputStream& stream)
...
-    m_fontHandles = std::move(fontHandles);
+    Font font;
+    font.m_fontHandles = std::move(fontHandles);
 
     // Store the font information
-    m_info.family = face->family_name ? face->family_name : std::string();
+    font.m_info.family = face->family_name ? face->family_name : std::string();
 
-    return true;
+    return font;
```
vittorioromeo, line 301: GitHub won't let me suggest here, but:

```cpp
return Font(std::move(fontHandles), std::string(face->family_name ? face->family_name : ""));

File: src/SFML/Graphics/Font.cpp [hunk 4]
```diff
@@ -209,40 +204,38 @@ bool Font::loadFromMemory(const void* data, std::size_t sizeInBytes)
...
```
vittorioromeo, line 232: GitHub won't let me suggest here, but:

```cpp
return Font(std::move(fontHandles), std::string(face->family_name ? face->family_name : ""));

File: src/SFML/Graphics/Font.cpp [hunk 5]
```diff
@@ -139,56 +138,52 @@ bool Font::loadFromFile(const std::filesystem::path& filename)
...
 
 #else
 
-    m_stream = std::make_shared<priv::ResourceStream>(filename);
-    return loadFromStream(*m_stream);
+    font.m_stream = std::make_shared<priv::ResourceStream>(filename);
+    return loadFromStream(*font.m_stream);
```
vittorioromeo, line 178:
```suggestion
    auto stream = std::make_shared<priv::ResourceStream>(filename);
    
    auto font = loadFromStream(*stream);
    font.m_stream = std::move(stream)
    return font;
```

vittorioromeo (changes requested):
Great change! This makes the API much nicer and safer to use. The only thing I think we can improve is that we don't need a default constructor at all, we always need font handles and a family name to construct the font AFAICS. Please let me know if I am mistaken!

ChrisThrasher:
> Great change! This makes the API much nicer and safer to use. The only thing I think we can improve is that we don't need a default constructor at all, we always need font handles and a family name to construct the font AFAICS. Please let me know if I am mistaken!

Sounds good. Happy to drop that default ctor in favor of something more useful.

include/SFML/Graphics/Font.hpp
ChrisThrasher: Why use `explicit` when there are 2 parameters? I'm only familiar with `explicit` on single-parameter constructors.

vittorioromeo (approved):
LGTM!

PR #3021, opened 2024-05-18
Remove default empty state of `sf::Image`

ChrisThrasher:
## Description

Sister PR to #3020 

Just like `sf::Font`, there is little value in having a default-constructed `sf::Image`. Its presence leaves the door open for bugs due to forgetting to initialize the image or failing to properly check the return value of a function like `loadFromFile`.

Supersedes #2228 which has accumulated conflicts and also makes changes beyond the scope of a pure API change. This PR is a simpler implementation of this idea that does not change semantics. Reducing the scope of this PR should make it easier to review. We can discuss semantic changes at a later time if desired.

### DIFF
--- a/examples/vulkan/Vulkan.cpp
+++ b/examples/vulkan/Vulkan.cpp
@@ -1801,13 +1801,15 @@ public:
     void setupTextureImage()
     {
         // Load the image data
-        sf::Image imageData;
+        const auto maybeImageData = sf::Image::loadFromFile("resources/logo.png");
 
-        if (!imageData.loadFromFile("resources/logo.png"))
+        if (!maybeImageData)
         {
             vulkanAvailable = false;
             return;
         }
+
+        const auto& imageData = *maybeImageData;
 
         // Create a staging buffer to transfer the data with
         const VkDeviceSize imageSize = imageData.getSize().x * imageData.getSize().y * 4;
--- a/include/SFML/Graphics/Image.hpp
+++ b/include/SFML/Graphics/Image.hpp
@@ -55,16 +55,16 @@ class SFML_GRAPHICS_API Image
 {
 public:
     ////////////////////////////////////////////////////////////
-    /// \brief Create the image and fill it with a unique color
+    /// \brief Construct the image and fill it with a unique color
     ///
     /// \param size  Width and height of the image
     /// \param color Fill color
     ///
     ////////////////////////////////////////////////////////////
-    void create(const Vector2u& size, const Color& color = Color::Black);
+    explicit Image(const Vector2u& size, const Color& color = Color::Black);
 
     ////////////////////////////////////////////////////////////
-    /// \brief Create the image from an array of pixels
+    /// \brief Construct the image from an array of pixels
     ///
     /// The \a pixel array is assumed to contain 32-bits RGBA pixels,
     /// and have the given \a width and \a height. If not, this is
@@ -75,7 +75,7 @@ public:
     /// \param pixels Array of pixels to copy to the image
     ///
     ////////////////////////////////////////////////////////////
-    void create(const Vector2u& size, const std::uint8_t* pixels);
+    Image(const Vector2u& size, const std::uint8_t* pixels);
 
     ////////////////////////////////////////////////////////////
     /// \brief Load the image from a file on disk
@@ -87,12 +87,12 @@ public:
     ///
     /// \param filename Path of the image file to load
     ///
-    /// \return True if loading was successful
+    /// \return Image if loading was successful, `std::nullopt` otherwise
     ///
     /// \see loadFromMemory, loadFromStream, saveToFile
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool loadFromFile(const std::filesystem::path& filename);
+    [[nodiscard]] static std::optional<Image> loadFromFile(const std::filesystem::path& filename);
 
     ////////////////////////////////////////////////////////////
     /// \brief Load the image from a file in memory
@@ -105,12 +105,12 @@ public:
     /// \param data Pointer to the file data in memory
     /// \param size Size of the data to load, in bytes
     ///
-    /// \return True if loading was successful
+    /// \return Image if loading was successful, `std::nullopt` otherwise
     ///
     /// \see loadFromFile, loadFromStream
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool loadFromMemory(const void* data, std::size_t size);
+    [[nodiscard]] static std::optional<Image> loadFromMemory(const void* data, std::size_t size);
 
     ////////////////////////////////////////////////////////////
     /// \brief Load the image from a custom stream
@@ -122,12 +122,12 @@ public:
     ///
     /// \param stream Source stream to read from
     ///
-    /// \return True if loading was successful
+    /// \return Image if loading was successful, `std::nullopt` otherwise
     ///
     /// \see loadFromFile, loadFromMemory
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool loadFromStream(InputStream& stream);
+    [[nodiscard]] static std::optional<Image> loadFromStream(InputStream& stream);
 
     ////////////////////////////////////////////////////////////
     /// \brief Save the image to a file on disk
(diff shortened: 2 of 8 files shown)


include/SFML/Graphics/Texture.hpp
```diff
@@ -110,15 +110,6 @@ class SFML_GRAPHICS_API Texture : GlResource
...
     /// \brief Load the texture from a file on disk
     ///
-    /// This function is a shortcut for the following code:
-    /// \code
-    /// sf::Image image;
-    /// if (!image.loadFromFile(filename))
-    ///     return false;
-    /// if (!texture.loadFromImage(image, area))
-    ///     return false;
-    /// \endcode
```
ChrisThrasher, line 120: It felt odd that our docs basically spell out the implementation. The issue with this is what when the implementation changes (as is the case in this PR) these docs need updating. I thought it better to remove this than continue to keep it up to date as the implementation evolves.

examples/vulkan/Vulkan.cpp
```diff
@@ -1801,14 +1801,16 @@ class VulkanExample
     void setupTextureImage()
     {
         // Load the image data
-        sf::Image imageData;
+        const auto maybeImageData = sf::Image::loadFromFile("resources/logo.png");
 
-        if (!imageData.loadFromFile("resources/logo.png"))
+        if (!maybeImageData)
```
vittorioromeo, line 1806:
```suggestion
        if (!maybeImageData.has_value())
```

Personal preference, but it provides local context about `maybeImageData` being a `std::optional`.

include/SFML/Graphics/Image.hpp
```diff
@@ -55,16 +55,16 @@ class SFML_GRAPHICS_API Image
...
     ////////////////////////////////////////////////////////////
-    /// \brief Create the image and fill it with a unique color
+    /// \brief Construct the image and fill it with a unique color
     ///
     /// \param size  Width and height of the image
     /// \param color Fill color
     ///
     ////////////////////////////////////////////////////////////
-    void create(const Vector2u& size, const Color& color = Color::Black);
+    Image(const Vector2u& size, const Color& color = Color::Black);
```
vittorioromeo, line 64:
```suggestion
    explicit Image(const Vector2u& size, const Color& color = Color::Black);
```

include/SFML/Graphics/Image.hpp
```diff
@@ -75,7 +75,7 @@ class SFML_GRAPHICS_API Image
     /// \param pixels Array of pixels to copy to the image
     ///
     ////////////////////////////////////////////////////////////
-    void create(const Vector2u& size, const std::uint8_t* pixels);
+    Image(const Vector2u& size, const std::uint8_t* pixels);
```
vittorioromeo, line 78:
```suggestion
    explicit Image(const Vector2u& size, const std::uint8_t* pixels);
```

`explicit` is a good default even for multiple-argument constructors as it still forces `Image` to be spelled out when constructing from a braced initialization list.

include/SFML/Graphics/Image.hpp
```diff
@@ -281,6 +281,12 @@ class SFML_GRAPHICS_API Image
     void flipVertically();
 
 private:
+    ////////////////////////////////////////////////////////////
+    /// \brief Directly initialize data members
+    ///
+    ////////////////////////////////////////////////////////////
+    Image(Vector2u size, std::vector<std::uint8_t>&& pixels);
```
vittorioromeo, line 288:
```suggestion
    explicit Image(Vector2u size, std::vector<std::uint8_t>&& pixels);
```

Not a big deal because `private`, but I'd still add it.

include/SFML/Graphics/Texture.hpp
vittorioromeo: +1

src/SFML/Graphics/Image.cpp
```diff
@@ -217,41 +211,32 @@ bool Image::loadFromMemory(const void* data, std::size_t size)
 
         if (ptr)
```
vittorioromeo, line 212: Can we invert these conditions and reduce the nesting here since we're touching this function?

src/SFML/Graphics/Image.cpp
```diff
@@ -179,34 +182,25 @@ bool Image::loadFromFile(const std::filesystem::path& filename)
 
     if (ptr)
```
vittorioromeo, line 183: Can we invert these conditions and reduce the nesting here since we're touching this function?

src/SFML/Graphics/Image.cpp
```diff
@@ -268,19 +253,13 @@ bool Image::loadFromStream(InputStream& stream)
...
```
vittorioromeo, line 254: Ditto, suggest inversion and early return here to be consistent.

src/SFML/Graphics/Texture.cpp
```diff
@@ -265,24 +265,28 @@ bool Texture::create(const Vector2u& size)
 ////////////////////////////////////////////////////////////
 bool Texture::loadFromFile(const std::filesystem::path& filename, const IntRect& area)
 {
-    Image image;
-    return image.loadFromFile(filename) && loadFromImage(image, area);
+    const auto image = sf::Image::loadFromFile(filename);
+    return loadFromImage(*image, area);
```
vittorioromeo, line 269: Uh, can't `image` fail to load here? Shouldn't we check and propagate a possible error like so?

```suggestion
    const auto image = sf::Image::loadFromFile(filename);
    
    if (!image.has_value())
        return false;
    
    return loadFromImage(*image, area);
```

src/SFML/Graphics/Texture.cpp
```diff
@@ -265,24 +265,28 @@ bool Texture::create(const Vector2u& size)
...
 }
 
 
 ////////////////////////////////////////////////////////////
 bool Texture::loadFromMemory(const void* data, std::size_t size, const IntRect& area)
 {
-    Image image;
-    return image.loadFromMemory(data, size) && loadFromImage(image, area);
+    const auto image = sf::Image::loadFromMemory(data, size);
+    if (!image)
```
vittorioromeo, line 277:
```suggestion
    if (!image.has_value())
```

src/SFML/Graphics/Texture.cpp
```diff
@@ -265,24 +265,28 @@ bool Texture::create(const Vector2u& size)
...
 }
 
 
 ////////////////////////////////////////////////////////////
 bool Texture::loadFromStream(InputStream& stream, const IntRect& area)
 {
-    Image image;
-    return image.loadFromStream(stream) && loadFromImage(image, area);
+    const auto image = sf::Image::loadFromStream(stream);
+    if (!image)
```
vittorioromeo, line 287:
```suggestion
    if (!image.has_value())
```

src/SFML/Graphics/Texture.cpp
```diff
@@ -456,11 +459,7 @@ Image Texture::copyToImage() const
 
 #endif // SFML_OPENGL_ES
 
-    // Create the image
-    Image image;
-    image.create(m_size, pixels.data());
-
-    return image;
+    return {m_size, pixels.data()};
```
vittorioromeo, line 462:
```suggestion
    return Image{m_size, pixels.data()};
```

vittorioromeo (changes requested):
Very good change here as well, I really like the new design. Some comments.

include/SFML/Graphics/Image.hpp
ChrisThrasher: I'm personally not a fan of forcing the use of the type name for multi-param constructors. SFML conventionally does not do this so I'd rather not start doing so now. If we want to revisit our use of `explicit` then I want that to happen in a separate PR.

src/SFML/Graphics/Image.cpp
ChrisThrasher: I chose to not rearrange control flow in this PR for the sake of minimizing the patch size and making it as easy as possible to merge. I certainly prefer what you're proposing but don't want to go too far making unrelated style changes in this PR.

My goal to this end is still to enable clang-tidy [`readability-else-after-return`](https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html) which would force us to transform this code into something simpler. I'd love to get your support in enabling that clang-tidy check so we can make this transformation across the whole codebase and systematically enforce it in all future PRs so we no longer have to manually point this out in code review.

ChrisThrasher: I opened https://github.com/SFML/SFML/issues/3024 to discuss this.

vittorioromeo (approved):
LGTM

PR #3022, opened 2024-05-18
Remove default empty state of `sf::SoundBuffer`

ChrisThrasher:
## Description

Sister PR to #3020 and #3021. The rationale from those two PRs carries over here too.

PR #3025, opened 2024-05-19
Remove default empty state of `sf::Shader`

ChrisThrasher:
## Description

Expanding on #3020, #3021, and #3022.

### DIFF
--- a/examples/island/Island.cpp
+++ b/examples/island/Island.cpp
@@ -94,11 +94,10 @@ int main()
     const auto font = sf::Font::loadFromFile("resources/tuffy.ttf").value();
 
     // Create all of our graphics resources
-    sf::Text               hudText(font);
-    sf::Text               statusText(font);
-    sf::Shader             terrainShader;
-    const sf::RenderStates terrainStates(&terrainShader);
-    sf::VertexBuffer       terrain(sf::PrimitiveType::Triangles, sf::VertexBuffer::Usage::Static);
+    sf::Text                  hudText(font);
+    sf::Text                  statusText(font);
+    std::optional<sf::Shader> terrainShader;
+    sf::VertexBuffer          terrain(sf::PrimitiveType::Triangles, sf::VertexBuffer::Usage::Static);
 
     // Set up our text drawables
     statusText.setCharacterSize(28);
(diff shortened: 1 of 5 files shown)


examples/island/Island.cpp
ChrisThrasher, line 1: I don't love the way I had to change this file. I hope to find a cleaner solution but I think this technically works.

examples/island/Island.cpp
```diff
@@ -123,14 +123,16 @@ int main()
...
-    else if (!terrainShader.loadFromFile("resources/terrain.vert", "resources/terrain.frag"))
+    else if (!(terrainShader = sf::Shader::loadFromFile("resources/terrain.vert", "resources/terrain.frag")))
     {
         prerequisitesSupported = false;
 
         statusText.setString("Failed to load shader program");
     }
     else
     {
+        terrainStates = sf::RenderStates(&*terrainShader);
```
vittorioromeo, line 134: Don't keep `terrainStates` as a data member, just create it on the fly around line 243.

examples/island/Island.cpp
```diff
@@ -123,14 +123,16 @@ int main()
     {
         statusText.setString("Shaders and/or Vertex Buffers Unsupported");
     }
-    else if (!terrainShader.loadFromFile("resources/terrain.vert", "resources/terrain.frag"))
+    else if (!(terrainShader = sf::Shader::loadFromFile("resources/terrain.vert", "resources/terrain.frag")))
     {
         prerequisitesSupported = false;
```
vittorioromeo, line 128: Get rid of this `prerequisitesSupported` variable altogether, and inline the check on line 122. Then before using the shader you can check if `terrainShader.has_value`, so the shader itself basically tells you whether it's available or not.

src/SFML/Graphics/Shader.cpp
```diff
@@ -896,16 +894,16 @@ bool Shader::compile(const char* vertexShaderCode, const char* geometryShaderCod
...
 
-    m_shaderProgram = castFromGlHandle(shaderProgram);
+    Shader shader(castFromGlHandle(shaderProgram));
 
     // Force an OpenGL flush, so that the shader will appear updated
     // in all contexts immediately (solves problems in multi-threaded apps)
     glCheck(glFlush());
 
-    return true;
+    return shader;
```
vittorioromeo, line 906: `castFromGlHandle` is just a cast, so you can: 

```cpp
// Force an OpenGL flush, so that the shader will appear updated
// in all contexts immediately (solves problems in multi-threaded apps)
glCheck(glFlush());

return Shader(castFromGlHandle(shaderProgram));
```

vittorioromeo (changes requested):
Good approach as usual, just a few nitpicks and a suggestion on how to improve the `Island` example

examples/shader/Shader.cpp
```diff
@@ -151,7 +147,7 @@ class StormBlink : public Effect
         }
 
         // Load the shader
-        return m_shader.loadFromFile("resources/storm.vert", "resources/blink.frag");
+        return true;
```
vittorioromeo, line 150: This seems a bit weird... At the very least, the comment needs to be changed. It also seems that `onLoad` doesn't have much of a purpose now, as the `load` function is as follows:

    void load()
    {
        m_isLoaded = sf::Shader::isAvailable() && onLoad();
    }
    
I'm not sure what the best course of action is here, maybe add a TODO in this PR and we can have a new one to revamp the example with a more appropriate design?

vittorioromeo (changes requested):

examples/shader/Shader.cpp
ChrisThrasher: > At the very least, the comment needs to be changed.

Fixed

Other `onLoad` functions have valid reasons for potentially failing. This particular function won't fail but others might so we can't yet change the signature of `onLoad`.

vittorioromeo (approved):
LGTM

PR #3026, opened 2024-05-20
Reduce required maths for transfrom by hardcoding the bottom row

L0laapk3:
## Description

Earlier I noted that all the constructors of `sf::Transform` in SFML follow the property of the bottom row being 0, 0, 1.  
All available operations on any `Transform`s with this property, results in the property also holding true for the resulting `Transform`. This includes the dot product `Transform::combine()` and `Transform.getInverse()`.  
The only way to break this property, is if the API user manually uses the `Transform::Transform(...)` constructor to create a Transform with a different bottom row. I could not find any useseful use cases for this.

This property can be exploited to significantly optimize the math of the operators of `Transform`. Especially in the heavily used `Transform::combine()` but also notably in `Transform::getInverse()`.

I did this because I could on a branch and have been using it in my own game for some time now. I understand that this is the lowest priority of low priority PR's, but since I already made the change I at least wanted to bring up the discussion rather than letting it go to waste.

## Tasks

-   [ ] Tested on Linux
-   [X] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

Open for suggestions.

### DIFF
--- a/include/SFML/Graphics/Transform.inl
+++ b/include/SFML/Graphics/Transform.inl
@@ -131,11 +131,11 @@ constexpr Transform& Transform::combine(const Transform& transform)
     const auto& b = transform.m_matrix;
 
     // clang-format off
-    *this = Transform(a[0] * b[0]  + a[4] * b[1]         ,
-                      a[0] * b[4]  + a[4] * b[5]         ,
+    *this = Transform(a[0] * b[0]  + a[4] * b[1],
+                      a[0] * b[4]  + a[4] * b[5],
                       a[0] * b[12] + a[4] * b[13] + a[12],
-                      a[1] * b[0]  + a[5] * b[1]         ,
-                      a[1] * b[4]  + a[5] * b[5]         ,
+                      a[1] * b[0]  + a[5] * b[1],
+                      a[1] * b[4]  + a[5] * b[5],
                       a[1] * b[12] + a[5] * b[13] + a[13]);
     // clang-format on
 
--- a/include/SFML/Graphics/Transform.inl
+++ b/include/SFML/Graphics/Transform.inl
@@ -41,12 +41,11 @@ constexpr Transform::Transform() = default;
 ////////////////////////////////////////////////////////////
 // clang-format off
 constexpr Transform::Transform(float a00, float a01, float a02,
-                               float a10, float a11, float a12,
-                               float a20, float a21, float a22)
-    : m_matrix{a00, a10, 0.f, a20,
-               a01, a11, 0.f, a21,
+                               float a10, float a11, float a12)
+    : m_matrix{a00, a10, 0.f, 0.f,
+               a01, a11, 0.f, 0.f,
                0.f, 0.f, 1.f, 0.f,
-               a02, a12, 0.f, a22}
+               a02, a12, 0.f, 1.f}
 {
 }
 // clang-format on
@@ -64,9 +63,7 @@ constexpr Transform Transform::getInverse() const
 {
     // clang-format off
     // Compute the determinant
-    const float det = m_matrix[0] * (m_matrix[15] * m_matrix[5] - m_matrix[7] * m_matrix[13]) -
-                      m_matrix[1] * (m_matrix[15] * m_matrix[4] - m_matrix[7] * m_matrix[12]) +
-                      m_matrix[3] * (m_matrix[13] * m_matrix[4] - m_matrix[5] * m_matrix[12]);
+    const float det = m_matrix[0] * m_matrix[5] - m_matrix[1] * m_matrix[4];
     // clang-format on
 
     // Compute the inverse if the determinant is not zero
@@ -74,15 +71,12 @@ constexpr Transform Transform::getInverse() const
     if (det != 0.f)
     {
         // clang-format off
-        return {(m_matrix[15] * m_matrix[5] - m_matrix[7] * m_matrix[13]) / det,
-               -(m_matrix[15] * m_matrix[4] - m_matrix[7] * m_matrix[12]) / det,
-                (m_matrix[13] * m_matrix[4] - m_matrix[5] * m_matrix[12]) / det,
-               -(m_matrix[15] * m_matrix[1] - m_matrix[3] * m_matrix[13]) / det,
-                (m_matrix[15] * m_matrix[0] - m_matrix[3] * m_matrix[12]) / det,
-               -(m_matrix[13] * m_matrix[0] - m_matrix[1] * m_matrix[12]) / det,
-                (m_matrix[7]  * m_matrix[1] - m_matrix[3] * m_matrix[5])  / det,
-               -(m_matrix[7]  * m_matrix[0] - m_matrix[3] * m_matrix[4])  / det,
-                (m_matrix[5]  * m_matrix[0] - m_matrix[1] * m_matrix[4])  / det};
+		return {(               m_matrix[5]                             ) / det,
+               -(               m_matrix[4]                             ) / det,
+				(m_matrix[13] * m_matrix[4] - m_matrix[5] * m_matrix[12]) / det,
+               -(               m_matrix[1]                             ) / det,
+                (               m_matrix[0]                             ) / det,
+               -(m_matrix[13] * m_matrix[0] - m_matrix[1] * m_matrix[12]) / det};
         // clang-format on
     }
     else
@@ -137,15 +131,12 @@ constexpr Transform& Transform::combine(const Transform& transform)
     const auto& b = transform.m_matrix;
 
     // clang-format off
-    *this = Transform(a[0] * b[0]  + a[4] * b[1]  + a[12] * b[3],
-                      a[0] * b[4]  + a[4] * b[5]  + a[12] * b[7],
-                      a[0] * b[12] + a[4] * b[13] + a[12] * b[15],
-                      a[1] * b[0]  + a[5] * b[1]  + a[13] * b[3],
-                      a[1] * b[4]  + a[5] * b[5]  + a[13] * b[7],
-                      a[1] * b[12] + a[5] * b[13] + a[13] * b[15],
-                      a[3] * b[0]  + a[7] * b[1]  + a[15] * b[3],
-                      a[3] * b[4]  + a[7] * b[5]  + a[15] * b[7],
-                      a[3] * b[12] + a[7] * b[13] + a[15] * b[15]);
+    *this = Transform(a[0] * b[0]  + a[4] * b[1]         ,
++                      a[0] * b[4]  + a[4] * b[5]         ,
++                      a[0] * b[12] + a[4] * b[13] + a[12],
++                      a[1] * b[0]  + a[5] * b[1]         ,
++                      a[1] * b[4]  + a[5] * b[5]         ,
++                      a[1] * b[12] + a[5] * b[13] + a[13]);
     // clang-format on
 
     return *this;
@@ -157,8 +148,7 @@ constexpr Transform& Transform::translate(const Vector2f& offset)
 {
     // clang-format off
     const Transform translation(1, 0, offset.x,
-                                0, 1, offset.y,
-                                0, 0, 1);
+                                0, 1, offset.y);
     // clang-format on
 
     return combine(translation);
@@ -170,8 +160,7 @@ constexpr Transform& Transform::scale(const Vector2f& factors)
 {
     // clang-format off
     const Transform scaling(factors.x, 0,         0,
-                            0,         factors.y, 0,
-                            0,         0,         1);
+                            0,         factors.y, 0);
     // clang-format on
 
     return combine(scaling);
@@ -183,8 +172,7 @@ constexpr Transform& Transform::scale(const Vector2f& factors, const Vector2f& c
 {
     // clang-format off
     const Transform scaling(factors.x, 0,         center.x * (1 - factors.x),
-                            0,         factors.y, center.y * (1 - factors.y),
-                            0,         0,         1);
+                            0,         factors.y, center.y * (1 - factors.y));
     // clang-format on
 
     return combine(scaling);
@@ -219,9 +207,9 @@ constexpr bool operator==(const Transform& left, const Transform& right)
     const float* b = right.getMatrix();
 
     // clang-format off
-    return ((a[0]  == b[0])  && (a[1]  == b[1])  && (a[3]  == b[3]) &&
-            (a[4]  == b[4])  && (a[5]  == b[5])  && (a[7]  == b[7]) &&
-            (a[12] == b[12]) && (a[13] == b[13]) && (a[15] == b[15]));
+    return ((a[0]  == b[0])  && (a[1]  == b[1])
+	     && (a[4]  == b[4])  && (a[5]  == b[5])
+		 && (a[12] == b[12]) && (a[13] == b[13]));
     // clang-format on
 }
 
--- a/include/SFML/Graphics/Transform.inl
+++ b/include/SFML/Graphics/Transform.inl
@@ -132,11 +132,11 @@ constexpr Transform& Transform::combine(const Transform& transform)
 
     // clang-format off
     *this = Transform(a[0] * b[0]  + a[4] * b[1]         ,
-+                      a[0] * b[4]  + a[4] * b[5]         ,
-+                      a[0] * b[12] + a[4] * b[13] + a[12],
-+                      a[1] * b[0]  + a[5] * b[1]         ,
-+                      a[1] * b[4]  + a[5] * b[5]         ,
-+                      a[1] * b[12] + a[5] * b[13] + a[13]);
+                      a[0] * b[4]  + a[4] * b[5]         ,
+                      a[0] * b[12] + a[4] * b[13] + a[12],
+                      a[1] * b[0]  + a[5] * b[1]         ,
+                      a[1] * b[4]  + a[5] * b[5]         ,
+                      a[1] * b[12] + a[5] * b[13] + a[13]);
     // clang-format on
 
     return *this;
--- a/test/Graphics/Transform.test.cpp
+++ b/test/Graphics/Transform.test.cpp
@@ -8,7 +8,6 @@
 #include <sstream>
 #include <type_traits>
 #include <vector>
-#include <iostream>
 
 #include <cassert>
 
--- a/test/Graphics/Transform.test.cpp
+++ b/test/Graphics/Transform.test.cpp
@@ -47,8 +47,7 @@ TEST_CASE("[Graphics] sf::Transform")
     SECTION("getInverse()")
     {
         STATIC_CHECK(sf::Transform::Identity.getInverse() == sf::Transform::Identity);
-        STATIC_CHECK(sf::Transform(1.0f, 2.0f, 3.0f, 4.0f, 8.0f, 6.0f).getInverse() ==
-                     sf::Transform::Identity);
+        STATIC_CHECK(sf::Transform(1.0f, 2.0f, 3.0f, 4.0f, 8.0f, 6.0f).getInverse() == sf::Transform::Identity);
         STATIC_CHECK(sf::Transform(1.0f, 2.0f, 3.0f, 5.0f, 2.0f, 4.0f).getInverse() ==
                      sf::Transform(-0.25f, 0.25f, -0.25f, 0.625f, -0.125f, -1.375f));
     }
@@ -153,8 +152,7 @@ TEST_CASE("[Graphics] sf::Transform")
             constexpr sf::Transform transform(1.0f, 2.0f, 3.0f, 4.0f, 5.0f, 4.0f);
             STATIC_CHECK(sf::Transform::Identity * transform == transform);
             STATIC_CHECK(transform * sf::Transform::Identity == transform);
-            STATIC_CHECK(transform * transform ==
-                         sf::Transform(9.0f, 12.0f, 14.0f, 24.0f, 33.0f, 36.0f));
+            STATIC_CHECK(transform * transform == sf::Transform(9.0f, 12.0f, 14.0f, 24.0f, 33.0f, 36.0f));
             STATIC_CHECK(transform * sf::Transform(10.0f, 2.0f, 3.0f, 4.0f, 50.0f, 40.0f) ==
                          sf::Transform(18.0f, 102.0f, 86.0f, 60.0f, 258.0f, 216.0f));
         }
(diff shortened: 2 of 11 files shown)


ChrisThrasher:
Are you able to write a benchmark to demonstrate the extent to which this improves performance?

L0laapk3:
> Are you able to write a benchmark to demonstrate the extent to which this improves performance?

I can try this, sure.

test/Graphics/Transform.test.cpp
```diff
@@ -8,6 +8,7 @@
 #include <sstream>
 #include <type_traits>
 #include <vector>
+#include <iostream>
```
ChrisThrasher, line 11:
```suggestion
```

include/SFML/Graphics/Transform.inl
```diff
@@ -137,15 +131,12 @@ constexpr Transform& Transform::combine(const Transform& transform)
...
-                      a[1] * b[12] + a[5] * b[13] + a[13] * b[15],
-                      a[3] * b[0]  + a[7] * b[1]  + a[15] * b[3],
-                      a[3] * b[4]  + a[7] * b[5]  + a[15] * b[7],
-                      a[3] * b[12] + a[7] * b[13] + a[15] * b[15]);
+    *this = Transform(a[0] * b[0]  + a[4] * b[1]         ,
+                      a[0] * b[4]  + a[4] * b[5]         ,
+                      a[0] * b[12] + a[4] * b[13] + a[12],
+                      a[1] * b[0]  + a[5] * b[1]         ,
+                      a[1] * b[4]  + a[5] * b[5]         ,
+                      a[1] * b[12] + a[5] * b[13] + a[13]);
```
ChrisThrasher, line 139: Please move the commas to the left

ChrisThrasher (changes requested):
This PR is promising in how it simplifies the API, simplifies the implementation, and likely improves performance. I'm still not entirely sure whether we're losing out on potentially useful functionality but so far it seems that the use cases we care about are still supported.

eXpl0it3r:
Not exactly a fan of this, as we're moving away from the mathematical definition in the name of some (unproven) "optimization".

Until now `sf::Transform` was really just a 4x4 transformation matrix. Now it's just "something". Any interoperability where a normal transformation matrix is expected, would need to readd the bottom row. I'm not even sure if the conversion to `sf::Glsl::Mat4` is correct now.

> The only way to break this property, is if the API user manually uses the Transform::Transform(...) constructor to create a Transform with a different bottom row. I could not find any useseful use cases for this.

For a normal transform changing the value might not matter, but I don't see why we should remove the mathematical possibilities from users.

If I understood this correctly, if you change the last row, you're no longer transforming within the 2D plane (R^2), but would map it into another space.
- https://math.stackexchange.com/questions/335932/what-is-the-sense-of-bottom-row-of-affine-transform-matrix
- https://stackoverflow.com/questions/32565827/whats-the-purpose-of-magic-4-of-last-row-in-matrix-4x4-for-3d-graphics

vittorioromeo:
I like the simplicity the optimization behind this PR, but the only way I would be comfortable accepting this is if instead of `sf::Transform` you introduced a brand new "transform type" that is convertible to `sf::Transform` that is only used internally by SFML.

`sf::Transform` is a general-purpose transformation matrix that is part of the public API, we cannot predict if users will always have a bottom row of `[0, 0, 1]`.

I think it's worthwhile investigating if all internal uses of `sf::Transform` by SFML can be changed to some `sf::priv::Transform2D` instead that has an hardcoded bottom row of `[0, 0, 1]`.

include/SFML/Graphics/Transform.inl
```diff
@@ -64,25 +63,20 @@ constexpr Transform Transform::getInverse() const
 {
     // clang-format off
     // Compute the determinant
-    const float det = m_matrix[0] * (m_matrix[15] * m_matrix[5] - m_matrix[7] * m_matrix[13]) -
-                      m_matrix[1] * (m_matrix[15] * m_matrix[4] - m_matrix[7] * m_matrix[12]) +
-                      m_matrix[3] * (m_matrix[13] * m_matrix[4] - m_matrix[5] * m_matrix[12]);
+    const float det = m_matrix[0] * m_matrix[5] - m_matrix[1] * m_matrix[4];
```
vittorioromeo, line 66: Also, have you considered just... checking here and branching?

```cpp
if (matrix[3] == 0.f && matrix[7] == 0.f && matrix[15] == 1.f)
{
    // special optimized case
}
else
{
    // original code here
}
```

This might be a good optimization that doesn't affect `sf::Transform`'s flexibility, but you'd have to prove it's worth it via benchmark.

L0laapk3:
> Not exactly a fan of this, as we're moving away from the mathematical definition in the name of some (unproven) "optimization".

I believe there is some misunderstanding here. I did not "remove" the bottom row from the calculations. I made it have fixed the values to 0, 0, 1 rather than them being configurable.

1. It is a mathematically proveable fact that when using only square matrices with a bottom row of `0, .., 0, 1`, for all the `sf::Transform` operations, the result is always another matrix with a bottom row of `0, ..., 0, 1`.

include/SFML/Graphics/Transform.inl
L0laapk3: Adding a branch on the hot path does not seem like good idea.

L0laapk3:
> Not exactly a fan of this, as we're moving away from the mathematical definition in the name of some (unproven) "optimization".

I believe there is some misunderstanding here. I did not "remove" the bottom row from the calculations. I made it have fixed the values to `0 0 1` rather than them being configurable.

1. It is a mathematically proveable fact that when using only square matrices with a bottom row of `0 .. 0 1`, for all the `sf::Transform` operations, the result is always another matrix with a bottom row of `0 .. 0 1`. I encourage you to fill in these values yourself for the bottom row and observe that in all cases the bottom row becomes `0 0 1` again and you will end up with the exact same functions as present in the PR.
2. It is easily checkable that indeed all code paths only construct `sf::Transform`s with a bottom row of `0 0 1`. The one notable exception being the freeform `sf::Transform(a00, a01, .., a21, a22)` constructor which I propose to remove this flexibility from (See below on what functionality is removed).

> Until now sf::Transform was really just a 4x4 transformation matrix.

This is not true, the third row and column in the 4x4 underlying memory object are already hardcoded to identity values, and these values were already optimized out of all the operators. As far as all mathematical operations are concerned, `sf::Transform` behaves as a 3x3 matrix. This PR does not change that into ""something"". It still behaves mathematically as a 3x3 matrix, only with the constraint that the bottom row can not be changed to anything other than `0 0 1`.

> If I understood this correctly, if you change the last row, you're no longer transforming within the 2D plane (R^2), but would map it into another space.

This is incorrect since I did not "change" the bottom row, at least in a mathematical sense. It just restricts the values to be `0 0 1`. `sf::Transform`. In the mathematical sense, `sf::Transform` is now a 2D affine transformation matrix able to do all affine transformations.

> For a normal transform changing the value might not matter, but I don't see why we should remove the mathematical possibilities from users.

**What functionality is actually lost by restricting the bottom row to `0 0 1`?**

As mentioned, all affine transformations are still supported (This includes all scaling, translation, rotation). So functionality is actually lost by not supporting other linear transformations?

One non-affine use case is projections. In a 3D environment, 4x4 non-affine transformations can be used to project a 3D space onto a 2D plane (Read: a non-isometric 2D camera of a 3D world). Since SFML already removes a dimension, this would mean with this change it is no longer possible to project a 2D space onto a 1D (line) camera in a non-isometric way.

I did also find this demo, where someone made some artistic warping effects: https://github.com/IRCSS/Non-Affine-Transformation-Matrices/blob/master/Assets/Shaders/Resources/nonAffineBending.shader. As far as I can tell, this is neither an affine nor a linear transformation, he uses a different (parametrized by theta) affine transformation for each vertex in his model, and this is not less possible after this PR compared to before this PR.

L0laapk3:
Oops, closed it by accident..

eXpl0it3r:
The first step of any optimization discussion is to do actual benchmarking. If they don't yield a significant difference, there's no need in discussing anything further, even if the calculation is "technically" simplified.

> the only way I would be comfortable accepting this is if instead of `sf::Transform` you introduced a brand new "transform type" that is convertible to `sf::Transform` that is only used internally by SFML.

I'm generally not a fan of introducing internal constructs for optimizations like this. You're trading complexity for (potential) performance gains. I much rather keep the complexity low and defer people to highly optimized math libraries if they require performance.

L0laapk3:
Ran some quick and dirty benchmarks. https://github.com/L0laapk3/SFML-benchmark

# test results

Ran on a 13900kf under WSL. Compiled with clang 18.1.4.

### Old (SFML/SFML 53ade4b)

```cpp
-------------------------------------------------------------------------------------------------
Benchmark                                       Time             CPU   Iterations UserCounters...
-------------------------------------------------------------------------------------------------
BM_transform_combine/min_time:4.000          2.64 ns         2.64 ns   2121118680 items_per_second=378.547M/s
BM_transform_getInverse/min_time:4.000       1.12 ns         1.12 ns   4978865605 items_per_second=890.346M/s
BM_transform_translate/min_time:4.000        2.09 ns         2.09 ns   2673725429 items_per_second=479.336M/s
BM_transform_rotate/min_time:4.000           8.66 ns         8.66 ns    651674658 items_per_second=115.518M/s
BM_transform_scale/min_time:4.000            2.57 ns         2.57 ns   2182173196 items_per_second=388.433M/s
```

### New (L0laapk3/SFML 3f84010)

```cpp
-------------------------------------------------------------------------------------------------
Benchmark                                       Time             CPU   Iterations  UserCounters...
-------------------------------------------------------------------------------------------------
BM_transform_combine/min_time:4.000          1.69 ns         1.69 ns   3325438781  items_per_second=593.03M/s
BM_transform_getInverse/min_time:4.000      0.553 ns        0.553 ns   10134886287 items_per_second=1.80678G/s
BM_transform_translate/min_time:4.000        1.39 ns         1.39 ns   3978370735  items_per_second=718.688M/s
BM_transform_rotate/min_time:4.000           7.58 ns         7.58 ns    737395540  items_per_second=131.899M/s
BM_transform_scale/min_time:4.000            1.69 ns         1.69 ns   3315908628  items_per_second=590.678M/s
```

ChrisThrasher:
When judging PRs some green flags I look for are reducing code and reducing complexity in general. SFML's commit history is _full_ of commits that remove unnecessary complexity. Do those commits improve performance? Perhaps marginally if they're doing something like removing a redundant conditional. Do we require those PRs to prove a performance gain before approval? Certainly not. We accept the PR because it's good in spite of whether or not it has performance implications. We make an implied judgement call that removing code is not likely to make performance significantly worse.

In the case of this PR it removes a lot of potentially unneeded complexity. (Still not 100% sure if this is necessarily complexity or not but I'm inclined to believe at the moment it's not needed). That alone is compelling evidence to consider this PR. Whether or not it speeds up runtime isn't all that important since this PR has benefits beyond performance. At a minimum it's not _adding_ complexity. Luckily there is evidence which suggests performance is improved so I think we can judge this PR on matters other than performance.

binary1248:
So these are the results I got when building with VS2022 ClangCL x64 and specifying flags `/arch:AVX512 /Gv /fp:fast` to give the optimizer as much freedom as it needs to get work done. Run on a Ryzen 9 7950X.
### Old
```
-------------------------------------------------------------------------------------------------
Benchmark                                       Time             CPU   Iterations UserCounters...
-------------------------------------------------------------------------------------------------
BM_transform_combine/min_time:4.000          5.30 ns         5.28 ns   1054117647 items_per_second=189.504M/s
BM_transform_getInverse/min_time:4.000       1.30 ns         1.30 ns   4318072289 items_per_second=767.657M/s
BM_transform_translate/min_time:4.000        4.66 ns         4.66 ns   1194666667 items_per_second=214.772M/s
BM_transform_rotate/min_time:4.000           8.56 ns         8.57 ns    676226415 items_per_second=116.654M/s
BM_transform_scale/min_time:4.000            4.99 ns         4.98 ns   1120000000 items_per_second=200.784M/s
```
### New
```
-------------------------------------------------------------------------------------------------
Benchmark                                       Time             CPU   Iterations UserCounters...
-------------------------------------------------------------------------------------------------
BM_transform_combine/min_time:4.000          5.25 ns         5.25 ns   1054117647 items_per_second=190.575M/s
BM_transform_getInverse/min_time:4.000       1.12 ns         1.12 ns   4977777778 items_per_second=892.375M/s
BM_transform_translate/min_time:4.000        4.69 ns         4.70 ns   1194666667 items_per_second=212.977M/s
BM_transform_rotate/min_time:4.000           7.78 ns         7.78 ns    702745098 items_per_second=128.502M/s
BM_transform_scale/min_time:4.000            4.37 ns         4.38 ns   1280000000 items_per_second=228.189M/s
```

Just a disclaimer up front, normally people will know me as the guy who has optimized performance in many places in the past. One rule that I never broke however, is that any optimization I perform must never have an effect on the public API of the library.

When writing a library there are multiple goals that have to be kept in mind. Obviously, a library that has horrible performance is not going to be one that many people will be willing to use. But a library that people have a hard time learning/adopting will also be a huge turn off. Keeping all the design and implementation goals in balance is stuff library maintainers have to deal with all the time when making changes.

In this case, your benchmarks have proven that at an instruction level, yes, forcing the CPU to run less instructions does have an impact on the performance of these functions when observing them under a microscope. However, we must not forget about considering the global context: where are these functions used, and is this really a bottleneck in real-world non-synthetic benchmarks?

The main reason why `sf::Transform` exists is because sfml-graphics was built on top of legacy OpenGL. In legacy OpenGL the only way users could _transform_ vertices without manipulating vertex data was by using the provided matrix stack. Rotations, translations and scaling would all be performed by the CPU and the final matrix fed to OpenGL. This was the model from the beginnings of OpenGL in the 1990s, when GPUs weren't really that specialized in doing matrix maths and not nearly as massively parallel as they are today. All of this maths we are trying to optimize here really belongs on the GPU where it was meant to be run. The only reason we are still where we are right now is because we maintainers just haven't gotten around to modernizing the graphics module just yet.

Coming back to how and where these functions are used, the only place where `::combine` is currently used in the library is by the other `Transform` functions. When the user decides to transform a `Transformable` they end up using one of either `::rotate`, `::translate` or `::scale` never actually calling `::combine` directly. As you said all of theses operations are affine transformations, so indeed we could optimize the maths behind them if we know an entire row is always fixed. Now comes the question: if it is so obvious to us that the maths can be simplified, why hasn't the compiler performed the optimizations for us and thrown out some of the unnecessary instructions?

From arithmetic we learned in school we know that `0 * x = 0` and `1 * x = x`. It just so happens the compiler also knows these rules... _for integer arithmetic_. When we dive into IEEE 754 floating point arithmetic we get `0 * x = <it depends>` and `1 * x = <it depends>`. These are not rules that are helpful to the optimizer. If anyone wants to read up more on this, you can do so [here](https://en.wikipedia.org/wiki/IEEE_754). To cut a long story short, IEEE 754 requires special handling of "special values" that breaks the rules we learned in school. We can see this in action using [compiler explorer](https://gcc.godbolt.org/z/6baPG9Y3v). Only when telling the compiler that it can perform "fast floating point operations" can it treat floats using the same arithmetic rules as it does for integers.

What does this all mean for us? Well, if we were to be really really pedantic, the changes you made would actually lead to different results if we fed it IEEE 754 special values. Does it really matter for everyday use? Probably not, but that is not a decision for the compiler to make, hence us having to specify explicitly that we don't care using the appropriate compiler flags.

If we can assume that we only have to work with "normal" values, we can still optimize the performance when performing the affine transformations, all we need to do is... be the optimizer ourselves. What you have done by hand is what the optimizer would have done if it was allowed to but could not because we passively forbid it. If we performed the inlining and eliding of mathematical terms ourselves we would end up with the performance that your code demonstrates without having to touch `::combine` and thus leaving the public API as it is. I think this would be a change that would encounter less resistance than the current proposal.

Additionally, the benchmark results I ran were built using ClangCL as noted above. Why didn't I build them using the Microsoft compiler? Well, what I saw confirmed the notorious reputation that the Microsoft compiler is just _really really bad_ at inlining code where it would make complete sense. Being able to inline code is also a prerequisite for a number of other optimizations, which also wouldn't be able to be performed by the Microsoft compiler, optimizations like auto-vectorization using SIMD instructions like [FMA](https://en.wikipedia.org/wiki/FMA_instruction_set). Basically, the `Transform` code as it is written now exploits the "Achilles Heel" of the Microsoft compiler if you can call it that, by making a call to `::combine` instead of performing the operations inline within the function body. Because it is a complete function call, the function body will have to have a prologue and epilogue that also costs CPU instructions, the function arguments will have to be passed using indirect memory accesses if they can't all fit in registers, etc. etc. This is the reason why I tried building with `/Gv` forcing the compiler to use the `__vectorcall` convention making the call overhead less of a problem. When building with ClangCL, this was not a problem because the `::combine` code was always inlined.

Taking it further, I told ClangCL that it could use up to AVX512 instructions to optimize the maths that was being performed. This actually worked out pretty well. When inspecting the instructions that were generated, you could see it was using a bunch of FMA instructions to perform the multiplications and additions. However if you look closer, and use tools like AMD uProf to instrument the code with the help of CPU counters, the next problem arises.

Because we aren't performing a "real" 4x4 matrix multiply, instead choosing to skip certain rows/columns we are only performing 3x32bit = 96bit operations. Since the SSE instruction set was introduced, SIMD registers have always been at least 128 bits wide. This means that you would perform SIMD operations optimally in groups of 4 floats. We can't tell the processor to ignore a value because that's... not how it works. Now you might be asking, if we only work on groups of 3 floats, how does ClangCL still manage to make use of 128bit instructions? Well, it just has to make sure that the memory/register that wasn't meant to be touched is appropriately restored. Compilers know the cost of each instruction, and if they deem that it is still a better option to run a 128bit instruction + restore instead of 4 32bit instructions it will generate the appropriate code. If you think of it, this is how [SMR hard drives](https://en.wikipedia.org/wiki/Shingled_magnetic_recording) also write their data, and everyone knows their performance is worse than that of [CMR hard drives](https://en.wikipedia.org/wiki/Perpendicular_recording).

What I am trying to say is that, sometimes, we shouldn't try to be too clever. Trying to help the compiler save work only leads to it being able to perform less optimizations on its own. Sometimes, doing more work actually takes _less time_ because of the "weird" ways in which modern processors work. In the end, without diving into the generated instructions using e.g. compiler explorer and instrumenting using actual CPU counters, any performance optimizations are at best "hand wavy" and at worst can actually lead to worse performance when enabling all the instruction set optimizations that are available to the compiler. Without diving into the whole "optimizing debug performance" discussion that is ongoing, I myself am not a big fan of optimizing... well... non-optimized code. The benchmark data I showed above shows that the gains become marginal when all compiler optimizations are actually enabled.

As a general rule of thumb, before performing any kind of optimizations, I like to ask myself if the problem I am trying to solve has been solved by countless others before me. In this case, multiplying 4x4 matrices on CPU is probably one of those solved problems, for which the CPU manufacturers have even helped us over the years. We just have to make use of the new tools they have provided us to get the work done. If developers know they are going to be performing a lot of linear algebra, I don't think SFML is the library of choice either. There are a multitude of better linear algebra libraries out there that even contain hand-written assembly that exploits all of the aforementioned optimizations. Here at SFML, performance is really good enough for us as long as it doesn't become a bottleneck in real-world scenarios.

To sum up, like I said, if you were to inline your optimizations into `::rotate`, `::scale` and `::translate` and manually simplify the computations there, we would be able to benefit from your performance gains without having to change the public API. This is a change that @eXpl0it3r and I would take less issue with than the current proposal that touches the public API.

L0laapk3:
> So these are the results I got when building with VS2022 ClangCL x64 and specifying flags /arch:AVX512 /Gv /fp:fast to give the optimizer as much freedom as it needs to get work done. Run on a Ryzen 9 7950X.

Without commenting on the rest of your post, I find it extremely suspicious that your results are up to 3x slower than mine for what should be more or less matched processors. Do you have any explanation for this?

binary1248:
Like I said, I ran the benchmark on Windows directly (not WSL), built with VS2022 using ClangCL because the default compiler produced even slower results. I don't know why you assume that the Ryzen 9 7950X and Intel 13900KF should produce the exact same results when built with different compilers, especially in synthetic tests. It is known that especially in synthetic single-threaded tests where everything fits in core-local L1D cache, Intel always came out in front because of their higher L1D bandwidth. AMD favours having larger shared L3 because in their philosophy it benefits real-world workloads more since they are becoming more multi-threaded these days. What makes matters worse is that, like I explained in my previous comment, we have a high amount of unaligned accesses here due to the way the code is written, another benefit for the Intel cache architecture. There have been enough articles on exactly this topic, but doing a quick search I found [this good one](https://chipsandcheese.com/2022/11/08/amds-zen-4-part-2-memory-subsystem-and-conclusion/).

Just for sake of discussion, I ran the benchmarks on a Ryzen 7 3600 (I don't have access to any Intel processors) running Debian and built with GCC and `-mavx2 -ffast-math` and these are the results.
### Old
```
-------------------------------------------------------------------------------------------------
Benchmark                                       Time             CPU   Iterations UserCounters...
-------------------------------------------------------------------------------------------------
BM_transform_combine/min_time:4.000          4.70 ns         4.70 ns   1190730302 items_per_second=212.782M/s
BM_transform_getInverse/min_time:4.000       4.84 ns         4.84 ns   1164881903 items_per_second=206.803M/s
BM_transform_translate/min_time:4.000        3.42 ns         3.42 ns   1628767426 items_per_second=292.309M/s
BM_transform_rotate/min_time:4.000           14.1 ns         14.1 ns    393183950 items_per_second=70.8252M/s
BM_transform_scale/min_time:4.000            3.18 ns         3.18 ns   1758419394 items_per_second=314.159M/s
```
### New
```
-------------------------------------------------------------------------------------------------
Benchmark                                       Time             CPU   Iterations UserCounters...
-------------------------------------------------------------------------------------------------
BM_transform_combine/min_time:4.000          2.74 ns         2.74 ns   2044065744 items_per_second=365.195M/s
BM_transform_getInverse/min_time:4.000       2.83 ns         2.83 ns   1974025570 items_per_second=353.804M/s
BM_transform_translate/min_time:4.000        2.69 ns         2.69 ns   2084774380 items_per_second=372.146M/s
BM_transform_rotate/min_time:4.000           12.4 ns         12.4 ns    452448636 items_per_second=80.8359M/s
BM_transform_scale/min_time:4.000            2.45 ns         2.45 ns   2279562321 items_per_second=407.835M/s
```
Because the difference is larger when run on a processor with higher memory latency (Zen 2 vs Zen 4) I can only assume that it means that the `Transform` code is pretty heavy on L1 accesses, especially misaligned ones as AMD uProf reported. Obviously removing L1 accesses as your code does will have a larger impact the worse the latency/bandwidth will be.

It is also a known fact that for CPU intensive workloads, Linux tends to win over Windows for various reasons.

When optimizing code, I wouldn't get too hung up on the absolute values that the benchmarks provide. As we have seen, it is highly dependent on compiler, operating system and CPU. What still has value however is a side-to-side performance comparison when the benchmarks are run in an identical environment. Some environments may have other bottlenecks which might skew the results in one direction or the other, but I would not label them "suspicious" just because they don't fit my expectations. Not everyone is going to run their SFML application on a 13900KF in WSL compiled with Clang (especially considering that SFML has known issues running in WSL anyway). Cherry picking results due to confirmation bias isn't really healthy for this kind of discussion considering many SFML users will be running their applications on Windows, probably built with the standard Microsoft compiler, running on perhaps even older AMD hardware.

If there is anyone out there with a 7950X running Windows, I have provided all the information necessary to be able to hopefully reproduce my results.

L0laapk3:
@binary1248 

>  I don't know why you assume that the Ryzen 9 7950X and Intel 13900KF should produce the exact same results

This is not at all what I said. I said they should be more or less matched.

> but I would not label them "suspicious" just because they don't fit my expectations

a 300% difference on platforms that should be in the same ballpark is absolutely a red flag.
`593M/s` to `190M/s` is the difference between a 13600K and an AMD bulldozer FX6100. It is ridiculous to ignore this.

> Windows

It is true that linux is generally faster, however these differences are typically measured in single digit percentage. Not 300%.

> L1

The memory footprint of the benchmark is measured in bytes..

> Cherry picking results

I had no such intentions. Not including `march` was a genuine mistake on my part and I intend to post new results. I have already reran the benchmarks, and I saw a smaller but still very significant gap (as I do not have AVX512). However I did not publish it for the following reasons:

### Benchmark issues

In the meantime I have made some new discoveries which make me want to discard all the current benchmark results entirely:

* I actually managed to somewhat reproduce your results: `-Ofast`/`-ffast-math` absolutely wrecks the score, resulting in `230M/s` for both versions on my machine.
* I also tried the benchmark after making the transform operations no longer in-place (`*this = ` -> `return`, return type no longer reference). I observed speeds of `2G/s` for this PR, or even `5.5G/s` when I implemented a full 4x4 matrix multiplication in `combine` (+ `-Ofast`). These numbers are implausible and I suspect `benchmark::DoNotOptimize()` may not be having the effect I intended it to have.

Therefore, I think more effort needs to go into defining a good benchmark first before reaching any conclusions.

### Sidenote

I think a conclusion should not be made based solely on testing with AVX512. Current gen intel devices do not support it and as a SFML user, it is therefore unreasonable to ship any binaries with AVX512 instructions in the forseeable future. Plus there are the mobile platforms which have weaker SIMD support afaik.

PR #3027, opened 2024-05-20
Remove default empty state of `sf::Texture`

ChrisThrasher:
## Description

Next in line after #3025 and others before it.

### DIFF
--- a/examples/android/app/src/main/jni/main.cpp
+++ b/examples/android/app/src/main/jni/main.cpp
@@ -86,9 +86,7 @@ int main(int argc, char* argv[])
     sf::RenderWindow window(screen, "");
     window.setFramerateLimit(30);
 
-    sf::Texture texture;
-    if (!texture.loadFromFile("image.png"))
-        return EXIT_FAILURE;
+    const auto texture = sf::Texture::loadFromFile("image.png").value();
 
     sf::Sprite image(texture);
     image.setPosition(sf::Vector2f(screen.size) / 2.f);
--- a/examples/cocoa/CocoaAppDelegate.mm
+++ b/examples/cocoa/CocoaAppDelegate.mm
@@ -41,9 +41,6 @@ struct SFMLmainWindow
 {
     SFMLmainWindow(sf::WindowHandle win) : renderWindow(win)
     {
-        if (!logo.loadFromFile(resPath / "logo.png"))
-            NSLog(@"Couldn't load the logo image");
-
         logo.setSmooth(true);
 
         sprite.setOrigin(sprite.getLocalBounds().getCenter());
@@ -57,7 +54,7 @@ struct SFMLmainWindow
     sf::RenderWindow      renderWindow;
     sf::Font              font{sf::Font::loadFromFile(resPath / "tuffy.ttf").value()};
     sf::Text              text{font};
-    sf::Texture           logo;
+    sf::Texture           logo{sf::Texture::loadFromFile(resPath / "logo.png").value()};
     sf::Sprite            sprite{logo};
     sf::Color             background{sf::Color::Blue};
 };
--- a/examples/opengl/OpenGL.cpp
+++ b/examples/opengl/OpenGL.cpp
@@ -58,10 +58,7 @@ int main()
         window.setMaximumSize(sf::Vector2u(1200, 900));
 
         // Create a sprite for the background
-        sf::Texture backgroundTexture;
-        backgroundTexture.setSrgb(sRgb);
-        if (!backgroundTexture.loadFromFile(resourcesDir() / "background.jpg"))
-            return EXIT_FAILURE;
+        const auto       backgroundTexture = sf::Texture::loadFromFile(resourcesDir() / "background.jpg", sRgb).value();
         const sf::Sprite background(backgroundTexture);
 
         // Create some text to draw on top of our OpenGL object
@@ -78,9 +75,7 @@ int main()
         mipmapInstructions.setPosition({200.f, 550.f});
 
         // Load a texture to apply to our 3D cube
-        sf::Texture texture;
-        if (!texture.loadFromFile(resourcesDir() / "logo.png"))
-            return EXIT_FAILURE;
+        auto texture = sf::Texture::loadFromFile(resourcesDir() / "logo.png").value();
 
         // Attempt to generate a mipmap for our cube texture
         // We don't check the return value here since
@@ -230,8 +225,7 @@ int main()
                     if (mipmapEnabled)
                     {
                         // We simply reload the texture to disable mipmapping
-                        if (!texture.loadFromFile(resourcesDir() / "logo.png"))
-                            return EXIT_FAILURE;
+                        texture = sf::Texture::loadFromFile(resourcesDir() / "logo.png").value();
 
                         mipmapEnabled = false;
                     }
--- a/include/SFML/Graphics/Font.hpp
+++ b/include/SFML/Graphics/Font.hpp
@@ -315,7 +315,8 @@ private:
     ////////////////////////////////////////////////////////////
     struct Page
     {
-        explicit Page(bool smooth);
+        [[nodiscard]] static std::optional<Page> make(bool smooth);
+        explicit Page(Texture&& texture);
 
         GlyphTable       glyphs;     //!< Table mapping code points to their corresponding glyph
         Texture          texture;    //!< Texture containing the pixels of the glyphs
--- a/include/SFML/Graphics/Texture.hpp
+++ b/include/SFML/Graphics/Texture.hpp
@@ -37,6 +37,7 @@
 #include <SFML/System/Vector2.hpp>
 
 #include <filesystem>
+#include <optional>
 
 #include <cstddef>
 #include <cstdint>
@@ -55,14 +56,6 @@ class Image;
 class SFML_GRAPHICS_API Texture : GlResource
 {
 public:
-    ////////////////////////////////////////////////////////////
-    /// \brief Default constructor
-    ///
-    /// Creates an empty texture.
-    ///
-    ////////////////////////////////////////////////////////////
-    Texture();
-
     ////////////////////////////////////////////////////////////
     /// \brief Destructor
     ///
@@ -101,11 +94,12 @@ public:
     /// If this function fails, the texture is left unchanged.
     ///
     /// \param size Width and height of the texture
+    /// \param sRgb True to enable sRGB conversion, false to disable it
     ///
-    /// \return True if creation was successful
+    /// \return Texture if creation was successful, otherwise `std::nullopt`
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool create(const Vector2u& size);
+    [[nodiscard]] static std::optional<Texture> create(const Vector2u& size, bool sRgb = false);
 
     ////////////////////////////////////////////////////////////
     /// \brief Load the texture from a file on disk
@@ -122,14 +116,17 @@ public:
     /// If this function fails, the texture is left unchanged.
     ///
     /// \param filename Path of the image file to load
+    /// \param sRgb     True to enable sRGB conversion, false to disable it
     /// \param area     Area of the image to load
     ///
-    /// \return True if loading was successful
+    /// \return Texture if loading was successful, otherwise `std::nullopt`
     ///
     /// \see loadFromMemory, loadFromStream, loadFromImage
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool loadFromFile(const std::filesystem::path& filename, const IntRect& area = {});
+    [[nodiscard]] static std::optional<Texture> loadFromFile(const std::filesystem::path& filename,
+                                                             bool                         sRgb = false,
+                                                             const IntRect&               area = {});
 
     ////////////////////////////////////////////////////////////
     /// \brief Load the texture from a file in memory
@@ -147,14 +144,19 @@ public:
     ///
     /// \param data Pointer to the file data in memory
     /// \param size Size of the data to load, in bytes
+    /// \param sRgb True to enable sRGB conversion, false to disable it
     /// \param area Area of the image to load
     ///
-    /// \return True if loading was successful
+    /// \return Texture if loading was successful, otherwise `std::nullopt`
     ///
     /// \see loadFromFile, loadFromStream, loadFromImage
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool loadFromMemory(const void* data, std::size_t size, const IntRect& area = {});
+    [[nodiscard]] static std::optional<Texture> loadFromMemory(
+        const void*    data,
+        std::size_t    size,
+        bool           sRgb = false,
+        const IntRect& area = {});
 
     ////////////////////////////////////////////////////////////
     /// \brief Load the texture from a custom stream
@@ -171,14 +173,15 @@ public:
     /// If this function fails, the texture is left unchanged.
     ///
     /// \param stream Source stream to read from
+    /// \param sRgb   True to enable sRGB conversion, false to disable it
     /// \param area   Area of the image to load
     ///
-    /// \return True if loading was successful
+    /// \return Texture if loading was successful, otherwise `std::nullopt`
     ///
     /// \see loadFromFile, loadFromMemory, loadFromImage
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool loadFromStream(InputStream& stream, const IntRect& area = {});
+    [[nodiscard]] static std::optional<Texture> loadFromStream(InputStream& stream, bool sRgb = false, const IntRect& area = {});
 
     ////////////////////////////////////////////////////////////
     /// \brief Load the texture from an image
@@ -195,14 +198,15 @@ public:
     /// If this function fails, the texture is left unchanged.
     ///
     /// \param image Image to load into the texture
+    /// \param sRgb   True to enable sRGB conversion, false to disable it
     /// \param area  Area of the image to load
     ///
-    /// \return True if loading was successful
+    /// \return Texture if loading was successful, otherwise `std::nullopt`
     ///
     /// \see loadFromFile, loadFromMemory
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool loadFromImage(const Image& image, const IntRect& area = {});
+    [[nodiscard]] static std::optional<Texture> loadFromImage(const Image& image, bool sRgb = false, const IntRect& area = {});
 
     ////////////////////////////////////////////////////////////
     /// \brief Return the size of the texture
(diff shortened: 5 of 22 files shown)


include/SFML/Graphics/Texture.hpp
```diff
@@ -98,14 +91,12 @@ class SFML_GRAPHICS_API Texture : GlResource
     ////////////////////////////////////////////////////////////
     /// \brief Create the texture
     ///
-    /// If this function fails, the texture is left unchanged.
-    ///
     /// \param size Width and height of the texture
     ///
-    /// \return True if creation was successful
+    /// \return Texture if loading was successful, otherwise `std::nullopt`
```
vittorioromeo, line 96:
```suggestion
    /// \return Texture if loading was successful, `std::nullopt` otherwise
```

include/SFML/Graphics/Texture.hpp
```diff
@@ -119,17 +110,16 @@ class SFML_GRAPHICS_API Texture : GlResource
     /// The maximum size for a texture depends on the graphics
     /// driver and can be retrieved with the getMaximumSize function.
     ///
-    /// If this function fails, the texture is left unchanged.
-    ///
     /// \param filename Path of the image file to load
     /// \param area     Area of the image to load
     ///
-    /// \return True if loading was successful
+    /// \return Texture if loading was successful, otherwise `std::nullopt`
```
vittorioromeo, line 116:
```suggestion
    /// \return Texture if loading was successful, `std::nullopt` otherwise
```

include/SFML/Graphics/Texture.hpp
```diff
@@ -143,18 +133,18 @@ class SFML_GRAPHICS_API Texture : GlResource
...
     /// driver and can be retrieved with the getMaximumSize function.
     ///
-    /// If this function fails, the texture is left unchanged.
-    ///
     /// \param data Pointer to the file data in memory
     /// \param size Size of the data to load, in bytes
     /// \param area Area of the image to load
     ///
-    /// \return True if loading was successful
+    /// \return Texture if loading was successful, otherwise `std::nullopt`
```
vittorioromeo, line 140:
```suggestion
    /// \return Texture if loading was successful, `std::nullopt` otherwise
```

include/SFML/Graphics/Texture.hpp
```diff
@@ -119,17 +110,16 @@ class SFML_GRAPHICS_API Texture : GlResource
...
     ///
     /// \see loadFromMemory, loadFromStream, loadFromImage
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool loadFromFile(const std::filesystem::path& filename, const IntRect& area = IntRect());
+    [[nodiscard]] static std::optional<Texture> loadFromFile(const std::filesystem::path& filename,
+                                                             const IntRect&               area = IntRect());
```
vittorioromeo, line 122:
```suggestion
                                                             const IntRect&               area = {});
```

include/SFML/Graphics/Texture.hpp
```diff
@@ -143,18 +133,18 @@ class SFML_GRAPHICS_API Texture : GlResource
...
-    /// \return True if loading was successful
+    /// \return Texture if loading was successful, otherwise `std::nullopt`
     ///
     /// \see loadFromFile, loadFromStream, loadFromImage
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool loadFromMemory(const void* data, std::size_t size, const IntRect& area = IntRect());
+    [[nodiscard]] static std::optional<Texture> loadFromMemory(const void*    data,
+                                                               std::size_t    size,
+                                                               const IntRect& area = IntRect());
```
vittorioromeo, line 147:
```suggestion
                                                               const IntRect& area = {});
```

include/SFML/Graphics/Texture.hpp
```diff
@@ -168,17 +158,15 @@ class SFML_GRAPHICS_API Texture : GlResource
     /// The maximum size for a texture depends on the graphics
     /// driver and can be retrieved with the getMaximumSize function.
     ///
-    /// If this function fails, the texture is left unchanged.
-    ///
     /// \param stream Source stream to read from
     /// \param area   Area of the image to load
     ///
-    /// \return True if loading was successful
+    /// \return Texture if loading was successful, otherwise `std::nullopt`
```
vittorioromeo, line 164:
```suggestion
    /// \return Texture if loading was successful, `std::nullopt` otherwise
```

include/SFML/Graphics/Texture.hpp
```diff
@@ -168,17 +158,15 @@ class SFML_GRAPHICS_API Texture : GlResource
...
     ///
     /// \see loadFromFile, loadFromMemory, loadFromImage
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool loadFromStream(InputStream& stream, const IntRect& area = IntRect());
+    [[nodiscard]] static std::optional<Texture> loadFromStream(InputStream& stream, const IntRect& area = IntRect());
```
vittorioromeo, line 169:
```suggestion
    [[nodiscard]] static std::optional<Texture> loadFromStream(InputStream& stream, const IntRect& area = {});
```

include/SFML/Graphics/Texture.hpp
```diff
@@ -192,17 +180,15 @@ class SFML_GRAPHICS_API Texture : GlResource
     /// The maximum size for a texture depends on the graphics
     /// driver and can be retrieved with the getMaximumSize function.
     ///
-    /// If this function fails, the texture is left unchanged.
-    ///
     /// \param image Image to load into the texture
     /// \param area  Area of the image to load
     ///
-    /// \return True if loading was successful
+    /// \return Texture if loading was successful, otherwise `std::nullopt`
```
vittorioromeo, line 186:
```suggestion
    /// \return Texture if loading was successful, `std::nullopt` otherwise
```

include/SFML/Graphics/Texture.hpp
```diff
@@ -192,17 +180,15 @@ class SFML_GRAPHICS_API Texture : GlResource
...
     ///
     /// \see loadFromFile, loadFromMemory
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool loadFromImage(const Image& image, const IntRect& area = IntRect());
+    [[nodiscard]] static std::optional<Texture> loadFromImage(const Image& image, const IntRect& area = IntRect());
```
vittorioromeo, line 191:
```suggestion
    [[nodiscard]] static std::optional<Texture> loadFromImage(const Image& image, const IntRect& area = {});
```

src/SFML/Graphics/Texture.cpp
```diff
@@ -175,21 +176,20 @@ bool Texture::create(const Vector2u& size)
...
-    m_actualSize    = actualSize;
-    m_pixelsFlipped = false;
-    m_fboAttachment = false;
+    Texture texture;
+    texture.m_size       = size;
+    texture.m_actualSize = actualSize;
 
     // Create the OpenGL texture if it doesn't exist yet
-    if (!m_texture)
+    if (!texture.m_texture)
```
vittorioromeo, line 188:
```suggestion
```

This check is useless now, right?

src/SFML/Graphics/Texture.cpp
```diff
@@ -198,7 +198,7 @@ bool Texture::create(const Vector2u& size)
     static const bool textureEdgeClamp = GLEXT_texture_edge_clamp || GLEXT_GL_VERSION_1_2 ||
                                          Context::isExtensionAvailable("GL_EXT_texture_edge_clamp");
 
-    if (!m_isRepeated && !textureEdgeClamp)
+    if (!texture.m_isRepeated && !textureEdgeClamp)
```
vittorioromeo, line 201: Ditto for the other checks -- unless I am missing something, they should all have their default value, as this is a brand new texture. Right?

vittorioromeo (dismissed):

include/SFML/Graphics/Texture.hpp
ChrisThrasher: What makes you feel strongly about this order of words? There are lots of SFML APIs that use slightly different variations of this same sentence so there is no one standard phrasing for me to pick from. Because there is little consistency I don't put too much thought into it. Perhaps it would be worthwhile to go through all our docs and unify our phrasing for APIs with the same semantics.

src/SFML/Graphics/Texture.cpp
```diff
@@ -175,21 +176,20 @@ bool Texture::create(const Vector2u& size)
...
     }
 
     // All the validity checks passed, we can store the new texture settings
-    m_size          = size;
-    m_actualSize    = actualSize;
-    m_pixelsFlipped = false;
-    m_fboAttachment = false;
+    Texture texture;
+    texture.m_size       = size;
+    texture.m_actualSize = actualSize;
```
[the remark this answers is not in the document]
ChrisThrasher: `sf::RenderTexture` takes advantage of this default constructor. To remove `Texture::Texture()` would result in more downstream changes to `RenderTexture`. However, if we also change `RenderTexture::create` to return a `std::optional<RenderTexture>` then this problem isn't so bad. I wonder if this PR should be postponed until I can perform this refactor inside `RenderTexture`. That order of operations may make more sense.

include/SFML/Graphics/Texture.hpp
vittorioromeo: I don't feel strongly about it, it just sounds better in my mind. Feel free to ignore and I agree with your last sentence

src/SFML/Graphics/Font.cpp
```diff
@@ -752,23 +752,28 @@ bool Font::setCurrentSize(unsigned int characterSize) const
 
 
 ////////////////////////////////////////////////////////////
-Font::Page::Page(bool smooth)
+Font::Page::Page(bool smooth) :
+texture(
+    [smooth]()
```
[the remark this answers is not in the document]
ChrisThrasher: I thought this was a C++20 feature. I suppose I'll remove it and wait for CI to yell at me if it is in fact a C++20 feature.

src/SFML/Graphics/Texture.cpp
ChrisThrasher: Yeah I think you're right. Good catch.

src/SFML/Graphics/Texture.cpp
ChrisThrasher: `m_isRepeated` defaults to `false` and is never touched prior to this line so yeah that will always evaluate to `true` and thus can be removed.

I think this whole section can be removed as well.
```cpp
    if (texture.m_sRgb && !textureSrgb)
    {
        static bool warned = false;

        if (!warned)
        {
#ifndef SFML_OPENGL_ES
            err() << "OpenGL extension EXT_texture_sRGB unavailable" << '\n';
#else
            err() << "OpenGL ES extension EXT_sRGB unavailable" << '\n';
#endif
            err() << "Automatic sRGB to linear conversion disabled" << std::endl;

            warned = true;
        }

        texture.m_sRgb = false;
    }
    ```
`m_sRgb` defaults to `false` and is never touched prior to this line so I don't think there is ever a way to enter this conditional. This is a good sign because it shows that this API change systematically eliminates certain errors that were made possible by the previous API.

Comment (ChrisThrasher) [reply src/SFML/Graphics/Texture.cpp:201]: As I look closer, `Texture::setSrgb` is only useful if you call it prior to loading the texture. I think that needs to be added as a function parameter to ensure there is still a way to set this. I believe the same can be said for `Texture::setSmooth` and `Texture::setRepeated`.

**Comment by ChrisThrasher** [comment]
I discovered a big issue with the current implementation. `sf::Texture::setSrgb`, `sf::Texture::setSmooth`, and `sf::Texture::setRepeated` are effectively used as constructor parameters and thus need to be reflected as parameters of the `create()` function. The definitions of those functions need to be inlined into `create()` as well so that functionality is not lost.

**Comment by ChrisThrasher** [comment]
> I discovered a big issue with the current implementation. `sf::Texture::setSrgb`, `sf::Texture::setSmooth`, and `sf::Texture::setRepeated` are effectively used as constructor parameters and thus need to be reflected as parameters of the `create()` function.

I fixed this by adding parameters to all four static factory functions for setting the smooth, sRbg, and repeated properties. I believe this fixes the functionality issues. Now you can construct a texture with those settings set to something non-default.

The new problem is that an API with 3 consecutive `bool` parameters _sucks_. We need a better solution. I'm thinking it may be best to create a `sf::Texture::Settings` struct with 3 fields for these 3 parameters. Then we can use a default parameter to set this so that users who don't care don't have to write it out. See all the `false, false, false` I had to add hurts, but I'm sure a better solution is possible.

File: include/SFML/Graphics/Texture.hpp [hunk 14]
```diff
@@ -663,9 +690,7 @@ SFML_GRAPHICS_API void swap(Texture& left, Texture& right) noexcept;
 /// // drawing a sprite
 ///
 /// // Load a texture from a file
-/// sf::Texture texture;
-/// if (!texture.loadFromFile("texture.png"))
-///     return -1;
+/// const auto texture = sf::Texture::texture.loadFromFile("texture.png").value();
```
texus, line 693:
```suggestion
/// const auto texture = sf::Texture::loadFromFile("texture.png").value();
```

ChrisThrasher:
I need to take some more time with this PR so I'm marking it as a draft to indicate that it's not ready for review in its current state.

ChrisThrasher:
I took another shot at this PR. This time I removed `sf::Texture::setSrgb` since it seemingly had no more purpose as a member function. The only place that value is read is within the creation factory function. Setting that value after creation would do nothing so it seemed bug-prone to leave it there. Thus I added 1 parameter to the texture factory functions for specifying sRgb or not so this functionality is not lost.

test/Graphics/Texture.test.cpp
```diff
@@ -284,25 +252,20 @@ TEST_CASE("[Graphics] sf::Texture", runDisplayTests())
...
         texture2.update(green);
-        texture2.setSrgb(false);
         texture2.setSmooth(true);
         texture2.setRepeated(false);
 
         sf::swap(texture1, texture2);
         CHECK_FALSE(texture1.isSrgb());
         CHECK(texture1.isSmooth());
         CHECK_FALSE(texture1.isRepeated());
-        CHECK(texture2.isSrgb());
```
ChrisThrasher, line 305: Something to know about `setSrgb()` prior to this PR is that it must be called before `create()` for it to work. In this case it was called after `create()` meaning the texture didn't actually attempt to enable SRGB.

My API refactor means this parameter must be provided upon creation. As it turns out this feature doesn't work on Windows OpenGL ES so the `create()` function automatically disables it and emits a warning. That means we can't test for `isSrgb` being `true` since on that one config it's false.

src/SFML/Graphics/Font.cpp
```diff
@@ -752,23 +752,28 @@ bool Font::setCurrentSize(unsigned int characterSize) const
...
+                image.setPixel({x, y}, Color::White);
+
+        // Create the texture
+        auto maybeTexture = sf::Texture::loadFromImage(image);
+        if (!maybeTexture)
+            err() << "Failed to load font page texture" << std::endl;
+
+        auto newTexture = maybeTexture.value();
+        newTexture.setSmooth(smooth);
+        return newTexture;
```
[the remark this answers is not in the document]
ChrisThrasher: I kept this in my latest revision of this PR. I don't like what I wrote but can't picture anything better at the moment.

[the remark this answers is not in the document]
kimci86: The natural next step would be that Font::Page is created with a factory function.

test/Graphics/Texture.test.cpp
kimci86: A comment explaining this in the code would be valuable so it doesn't look like testing it was just overlooked.

src/SFML/Graphics/Font.cpp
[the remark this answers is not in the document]
ChrisThrasher: I went ahead and made this change because it seems sensible. What happened is now the null check occurs downstream in `sf::Font::loadPage` where the pages are actually constructed. If construction fails inside `loadPage` then what do we do? I chose to `assert` that loading succeeds but we could just as well use `.value()` in which case we're back to the same solution as what I previously wrote. 

Should we go so far as to make `sf::Font::loadPage` return something nullable and put the onus users? That's possible but I don't love the idea.

ChrisThrasher:
Rebased on `master` and fixed merge conflicts.

examples/android/app/src/main/jni/main.cpp
```diff
@@ -86,9 +86,7 @@ int main(int argc, char* argv[])
     sf::RenderWindow window(screen, "");
     window.setFramerateLimit(30);
 
-    sf::Texture texture;
-    if (!texture.loadFromFile("image.png"))
-        return EXIT_FAILURE;
+    const auto texture = sf::Texture::loadFromFile("image.png").value();
```
vittorioromeo, line 89: This changes the behavior to throwing an exception rather than returning `EXIT_FAILURE`. Given that some behavior changes ended up being reverted in #3051, do we want to maintain the same behavior as before here too?

examples/cocoa/CocoaAppDelegate.mm
```diff
@@ -57,7 +54,7 @@
     sf::RenderWindow      renderWindow;
     sf::Font              font{sf::Font::loadFromFile(resPath / "tuffy.ttf").value()};
     sf::Text              text{font};
-    sf::Texture           logo;
+    sf::Texture           logo{sf::Texture::loadFromFile(resPath / "logo.png").value()};
```
vittorioromeo, line 57: Ditto about behavioral changes, we also lose the `NSLog` call here. Note that we already changed behavior in line 55 for the font, so I'm not sure...

examples/opengl/OpenGL.cpp
```diff
@@ -58,10 +58,7 @@ int main()
         window.setMaximumSize(sf::Vector2u(1200, 900));
 
         // Create a sprite for the background
-        sf::Texture backgroundTexture;
-        backgroundTexture.setSrgb(sRgb);
-        if (!backgroundTexture.loadFromFile(resourcesDir() / "background.jpg"))
-            return EXIT_FAILURE;
+        const auto       backgroundTexture = sf::Texture::loadFromFile(resourcesDir() / "background.jpg", sRgb).value();
```
vittorioromeo, line 61: Not going to repeat myself for all the other ones, but all of these `.value()` calls in the examples change behavior and such changes were reverted in #3051 IIRC.

Maybe we should have a policy about throwing in examples?

include/SFML/Graphics/Font.hpp
```diff
@@ -315,7 +315,8 @@ class SFML_GRAPHICS_API Font
     ////////////////////////////////////////////////////////////
     struct Page
     {
-        explicit Page(bool smooth);
+        [[nodiscard]] static std::optional<Page> make(bool smooth);
+        Page(Texture&& texture);
```
vittorioromeo, line 319:
```suggestion
        explicit Page(Texture&& texture);
```

vittorioromeo (changes requested):
Minor considerations but looks good to me otherwise, happy to approve once you take a look and decide what to do.

examples/android/app/src/main/jni/main.cpp
ChrisThrasher: > This changes the behavior to throwing an exception

In practice the result is the same. Failure to load means the program stops immediately. 

> Given that some behavior changes ended up being reverted in https://github.com/SFML/SFML/pull/3051, do we want to maintain the same behavior as before here too?

#3051 was different. In that case I truly did change behavior in a way that totally broke the example on certain platforms. For now I want to use the same `.value()` pattern in our examples until we come to a consensus on a better alternative. I'm not sure exactly what alternative you or others may prefer.

examples/cocoa/CocoaAppDelegate.mm
ChrisThrasher: To be honest my attitude here was "meh whatever" about removing those logs. I'm not married to this solution and can go back to the previous solution if you feel strongly about keeping that.

vittorioromeo: I never used/targeted Cocoa, I have no idea if those logs provide any real value or not, but honestly your solution is simpler and probably good enough. Unless you know anyone who targets Cocoa that can provide more feedback, I don't feel strongly about keeping the logs at all.

examples/android/app/src/main/jni/main.cpp
vittorioromeo: Makes sense, you are correct that #3051 was a special case. Still, would be nice to have a policy regarding exceptions in our examples. Resolving for now.

vittorioromeo (approved):

src/SFML/Graphics/Font.cpp
```diff
@@ -757,7 +760,7 @@ bool Font::setCurrentSize(unsigned int characterSize) const
 
 
 ////////////////////////////////////////////////////////////
-Font::Page::Page(bool smooth)
+std::optional<Font::Page> Font::Page::make(bool smooth)
```
eXpl0it3r, line 763: Any reason we didn't use `create` as with most other factory functions?

ChrisThrasher: [Good point](https://github.com/SFML/SFML/pull/3103)

PR #3028, opened 2024-05-21
Remove default empty state of `sf::Cursor`

ChrisThrasher:
## Description

Not much to say about this. It was pretty simple.

### DIFF
--- a/test/Window/Cursor.test.cpp
+++ b/test/Window/Cursor.test.cpp
@@ -10,60 +10,38 @@ TEST_CASE("[Window] sf::Cursor", runDisplayTests())
 {
     SECTION("Type traits")
     {
+        STATIC_CHECK(!std::is_default_constructible_v<sf::Cursor>);
         STATIC_CHECK(!std::is_copy_constructible_v<sf::Cursor>);
         STATIC_CHECK(!std::is_copy_assignable_v<sf::Cursor>);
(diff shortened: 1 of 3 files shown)


test/Window/Cursor.test.cpp
```diff
@@ -10,60 +10,38 @@ TEST_CASE("[Window] sf::Cursor", runDisplayTests())
...
-            sf::Cursor       movedCursor;
-            const sf::Cursor cursor(std::move(movedCursor));
-        }
-
-        SECTION("Assignment")
-        {
-            sf::Cursor movedCursor;
-            sf::Cursor cursor;
-            cursor = std::move(movedCursor);
-        }
```
eXpl0it3r, line 37: Does that mean `sf::Cursor` is no longer movable? Do we always create a copy now?

ChrisThrasher: https://github.com/SFML/SFML/blob/7234fc149bf21b12fbb10deec73025fdfec07808/test/Window/Cursor.test.cpp#L15-L16

It's still moveable.

I removed these tests because we use `= default` implementations of move operations and by convention in SFML we don't write tests for defaulted operations. Instead of rewriting these tests for the new API I just removed them. They don't add value.

eXpl0it3r (approved):

PR #3029, opened 2024-05-21
Added support for changing the audio playback device during runtime

binary1248:
Added functions to:
- Get a list of available audio playback devices
- Get the default audio playback device is one is marked as such
- Get the currently used audio playback device
- Set the audio playback device to use

The API mirrors that which is already part of `sf::SoundRecorder` for capture devices, with a few "upgrades" for the quirks that might occur with playback devices (such as duplicate names).

Test by plugging in multiple audio devices, running the sound effects example and spamming the F1 key to your heart's content.

### DIFF
--- a/examples/sound_effects/SoundEffects.cpp
+++ b/examples/sound_effects/SoundEffects.cpp
@@ -1089,13 +1089,24 @@ int main()
 
     // Create the description text
     sf::Text description(font, "Current effect: " + effects[current]->getName(), 20);
-    description.setPosition({10.f, 530.f});
+    description.setPosition({10.f, 522.f});
     description.setFillColor(sf::Color(80, 80, 80));
 
     // Create the instructions text
     sf::Text instructions(font, "Press left and right arrows to change the current effect", 20);
-    instructions.setPosition({280.f, 555.f});
+    instructions.setPosition({280.f, 544.f});
     instructions.setFillColor(sf::Color(80, 80, 80));
+
+    // Create the playback device text
+    auto     playbackDeviceName = sf::PlaybackDevice::getDevice();
+    sf::Text playbackDevice(font, "Current playback device: " + playbackDeviceName.value_or("None"), 20);
+    playbackDevice.setPosition({10.f, 566.f});
+    playbackDevice.setFillColor(sf::Color(80, 80, 80));
+
+    // Create the playback device instructions text
+    sf::Text playbackDeviceInstructions(font, "Press F1 to change device", 20);
+    playbackDeviceInstructions.setPosition({565.f, 566.f});
+    playbackDeviceInstructions.setFillColor(sf::Color(80, 80, 80));
 
     // Start the game loop
     const sf::Clock clock;
@@ -1139,6 +1150,37 @@ int main()
                         description.setString("Current effect: " + effects[current]->getName());
                         break;
 
+                    // F1 key: change playback device
+                    case sf::Keyboard::Key::F1:
+                    {
+                        // We need to query the list every time we want to change
+                        // since new devices could have been added in the mean time
+                        const auto devices       = sf::PlaybackDevice::getAvailableDevices();
+                        const auto currentDevice = sf::PlaybackDevice::getDevice();
+                        auto       next          = currentDevice;
+
+                        for (auto iter = devices.begin(); iter != devices.end(); ++iter)
+                        {
+                            if (*iter == currentDevice)
+                            {
+                                const auto nextIter = std::next(iter);
+                                next                = (nextIter == devices.end()) ? devices.front() : *nextIter;
+                                break;
+                            }
+                        }
+
+                        if (next)
+                        {
+                            if (!sf::PlaybackDevice::setDevice(*next))
+                                std::cerr << "Failed to set the playback device to: " << *next << std::endl;
+
+                            playbackDeviceName = sf::PlaybackDevice::getDevice();
+                            playbackDevice.setString("Current playback device: " + playbackDeviceName.value_or("None"));
+                        }
+
+                        break;
+                    }
+
                     default:
                         effects[current]->handleKey(keyPressed->code);
                         break;
@@ -1160,6 +1202,8 @@ int main()
         window.draw(textBackground);
         window.draw(instructions);
         window.draw(description);
+        window.draw(playbackDevice);
+        window.draw(playbackDeviceInstructions);
 
         // Finally, display the rendered frame on screen
         window.display();
--- a/include/SFML/Audio/PlaybackDevice.hpp
+++ b/include/SFML/Audio/PlaybackDevice.hpp
@@ -1,0 +1,110 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2024 Laurent Gomila (laurent@sfml-dev.org)
+//
+// This software is provided 'as-is', without any express or implied warranty.
+// In no event will the authors be held liable for any damages arising from the use of this software.
+//
+// Permission is granted to anyone to use this software for any purpose,
+// including commercial applications, and to alter it and redistribute it freely,
+// subject to the following restrictions:
+//
+// 1. The origin of this software must not be misrepresented;
+//    you must not claim that you wrote the original software.
+//    If you use this software in a product, an acknowledgment
+//    in the product documentation would be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such,
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+#pragma once
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#include <SFML/Audio/Export.hpp>
+
+#include <optional>
+#include <string>
+#include <vector>
+
+
+namespace sf::PlaybackDevice
+{
+////////////////////////////////////////////////////////////
+/// \brief Get a list of the names of all available audio playback devices
+///
+/// This function returns a vector of strings containing
+/// the names of all available audio playback devices.
+///
+/// If the operating system reports multiple devices with
+/// the same name, a number will be appended to the name
+/// of all subsequent devices to distinguish them from each
+/// other. This guarantees that every entry returned by this
+/// function will represent a unique device.
+///
+/// For example, if the operating system reports multiple
+/// devices with the name "Sound Card", the entries returned
+/// would be:
+///   - Sound Card
+///   - Sound Card 2
+///   - Sound Card 3
+///   - ...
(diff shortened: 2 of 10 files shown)


Bambo-Borris:
Awesome work here Binary! I'm looking forward to testing this in my own project later today, this is the kind of feature with real value to power users.

include/SFML/Audio/PlaybackDevice.hpp
```diff
@@ -0,0 +1,94 @@
...
+/// \brief Get a list of the names of all available audio playback devices
+///
+/// This function returns a vector of strings, containing
+/// the names of all available audio playback devices.
+///
+/// If the operating system reports multiple devices with
+/// the same name, starting from the second device a number
+/// is appended to the name to distinguish the devices from
+/// each other. This guarantees that every entry returned
+/// by this function will represent a unique device.
```
vittorioromeo, line 48: Would be nice to show an example here to see how the number actually looks in the computed final name.

include/SFML/Audio/PlaybackDevice.hpp
```diff
@@ -0,0 +1,94 @@
...
+///
+/// The default device, if one is marked as such, will be
+/// placed at the beginning of the vector.
```
vittorioromeo, line 51: May be also worthwhile to clarify what happens if no audio devices are detected. I assume an empty vector is returned...?

include/SFML/Audio/PlaybackDevice.hpp
```diff
@@ -0,0 +1,94 @@
...
+///
+/// \return A vector of strings containing the names
```
vittorioromeo, line 53:
```suggestion
/// \return A vector of strings containing the device names
```

include/SFML/Audio/PlaybackDevice.hpp
```diff
@@ -0,0 +1,94 @@
...
+////////////////////////////////////////////////////////////
+std::string SFML_AUDIO_API getDefaultDevice();
+
+////////////////////////////////////////////////////////////
+/// \brief Set the audio playback device
+///
+/// This function sets the audio playback device to the device
+/// with the given \a name. It can be called on the fly (i.e:
+/// while recording).
+///
```
vittorioromeo, line 76: What happens if there any currently playing sound/music while the device is switched? Might be a good note to add in the documentation.

include/SFML/Audio/PlaybackDevice.hpp
```diff
@@ -0,0 +1,94 @@
...
+////////////////////////////////////////////////////////////
+std::vector<std::string> SFML_AUDIO_API getAvailableDevices();
+
+////////////////////////////////////////////////////////////
+/// \brief Get the name of the default audio playback device
+///
+/// This function returns the name of the default audio
+/// playback device. If none is available, an empty string
+/// is returned.
```
vittorioromeo, line 63: Have you considered returning `std::optional<std::string>` instead to clearly signal that there might not be a default device and preventing users from assuming that the string has a useful value?

src/SFML/Audio/AudioDevice.hpp
```diff
@@ -217,6 +297,22 @@ class AudioDevice
...
+    ////////////////////////////////////////////////////////////
+    std::optional<ma_device_id> getSelectedDeviceId() const;
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Initialize the audio device and engine
+    ///
+    /// \return True if initialization was successful, false if it failed
+    ///
+    ////////////////////////////////////////////////////////////
+    bool initialize();
```
vittorioromeo, line 314:
```suggestion
    [[nodiscard]] bool initialize();
```

vittorioromeo (changes requested):
Thank you a lot for doing this work! I think it's a great improvement from the status quo.

My concern with this design is that we always have one global set-up audio playback device. 

A user wouldn't be able to simultaneously send a few sounds to a device and few other sounds to another device, is that correct?

Use cases such as sending voice comms to headphones and game audio to speakers would not be supported if that is the case.

Have you considered a different design that would allow users to set a "default SFML playback device" but then also support the use case I described, where users can play sounds/music on specific audio devices without having to change the global set "playback device"?

binary1248:
Updated with suggestions.

binary1248:
> Use cases such as sending voice comms to headphones and game audio to speakers would not be supported if that is the case.

> Have you considered a different design that would allow users to set a "default SFML playback device" but then also support the use case I described, where users can play sounds/music on specific audio devices without having to change the global set "playback device"?

This is a very very advanced topic, one so advanced that not many AAA games even dare to support simultaneously outputting different audio streams to different audio playback devices.

While technically possible, this would require an entire redesign of the sfml-audio public API, and that is definitely out of scope for this PR and SFML 3.

src/SFML/Audio/AudioDevice.cpp
```diff
@@ -359,6 +453,126 @@ Vector3f AudioDevice::getUpVector()
...
+    {
+        m_playbackDevice.reset();
+        getCurrentDevice() = std::nullopt;
+        err() << "Failed to initialize the audio playback device: " << ma_result_description(result) << std::endl;
+        return false;
+    }
+
+    // Update the current device string from the the device we just initialized
+    {
+        char   deviceName[MA_MAX_DEVICE_NAME_LENGTH + 1];
```
ChrisThrasher, line 511: Can we use a `std::array` here? I'd prefer if new code didn't use C arrays.

src/SFML/Audio/AudioDevice.hpp
```diff
@@ -71,6 +75,110 @@ class AudioDevice
...
+    ///
+    /// \see getAvailableDevices, getDefaultDevice
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] static bool setDevice(const std::string& name);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the name of the current audio playback device
+    ///
+    /// \return The name of the current audio playback device or std::nullopt if there is none
```
ChrisThrasher, line 134:
```suggestion
    /// \return The name of the current audio playback device or `std::nullopt` if there is none
```

src/SFML/Audio/AudioDevice.hpp
```diff
@@ -217,6 +325,22 @@ class AudioDevice
     static Vector3f getUpVector();
 
 private:
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the device ID of the currently selected device
+    ///
+    /// \return The device ID of the currently selected device or std::nullopt if none could be found
```
ChrisThrasher, line 331:
```suggestion
    /// \return The device ID of the currently selected device or `std::nullopt` if none could be found
```

src/SFML/Audio/MiniaudioUtils.cpp
```diff
@@ -25,81 +25,59 @@
...
 #include <ostream>
 
 #include <cassert>
+#include <cstring>
 
 
 namespace sf::priv
 {
 namespace
 {
```
ChrisThrasher, line 46: Why is there an anonymous namespace inside of `sf::priv`? That's not typically how we do things.

ChrisThrasher (changes requested):
No structural complaints but I have a number of smaller items of feedback

src/SFML/Audio/AudioDevice.cpp
```diff
@@ -221,6 +177,144 @@ ma_engine* AudioDevice::getEngine()
...
+
+    const std::lock_guard lock(instance->m_resourcesMutex);
+
+    // Deinitialize all audio resources
+    for (const auto& entry : instance->m_resources)
+        entry.deinitializeFunc(entry.resource);
+
+    // Destroy the old engine
+    if (instance->m_engine)
+        ma_engine_uninit(&*instance->m_engine);
```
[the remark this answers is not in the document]
binary1248: This won't work. `std::unique_ptr` is not meant to store full objects like `std::optional` is.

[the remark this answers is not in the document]
ChrisThrasher: What do you mean? The unique_ptr will store a pointer to that full object. We’re changing allocation strategies but resource being managed can be the same.

src/SFML/Audio/AudioDevice.cpp
```diff
@@ -221,6 +177,144 @@ ma_engine* AudioDevice::getEngine()
...
+    auto getDevices = [](auto& context)
+    {
+        ma_device_info* deviceInfos{};
+        ma_uint32       deviceCount{};
+
+        // Get the playback devices
+        if (const auto result = ma_context_get_devices(&context, &deviceInfos, &deviceCount, nullptr, nullptr);
+            result != MA_SUCCESS)
+        {
+            ma_context_uninit(&context);
```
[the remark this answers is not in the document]
binary1248: Same as above, all these objects are not dynamically allocated by us. They could just as well be placed on the stack and still have to be passed to miniaudio using pointers. If you want to introduce RAII semantics around them you would have to write your own wrapper or wait until [`scope_exit`](https://en.cppreference.com/w/cpp/experimental/scope_exit) makes it into the standard.

binary1248:
Updated.

src/SFML/Audio/AudioDevice.cpp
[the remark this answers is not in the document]
binary1248: miniaudio doesn't allocate the objects itself, we would still have to `make_unique` them ourselves before passing them on to be initialized, and in the case initialization fails we would still have to immediately deallocate the uninitialized objects _without calling the `uninit` functions_. Besides the fact that having to dynamically allocate objects that are perfectly fine with automatic storage duration is just pure wasteful, this really isn't the place to abuse `std::unique_ptr` just for its deleter functionality. If we want RAII semantics then we should go the whole way and write proper wrappers instead of browsing through the standard library looking for constructs we can abuse just to save a few lines of code.

binary1248:
Only if they don't have any expectations about what is returned from them... so probably not.

ChrisThrasher (approved):

vittorioromeo (commented):
Minor nitpick, LGTM otherwise

examples/sound_effects/SoundEffects.cpp
```diff
@@ -1139,6 +1150,38 @@ int main()
...
+                        }
+
+                        if (next)
+                        {
+                            if (!sf::PlaybackDevice::setDevice(*next))
+                                std::cerr << "Failed to set the playback device to: " << *next << std::endl;
+
+                            playbackDeviceName = sf::PlaybackDevice::getDevice();
+                            playbackDevice.setString(
+                                "Current playback device: " + (playbackDeviceName ? *playbackDeviceName : "None"));
```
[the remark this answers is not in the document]
binary1248: Done

vittorioromeo (approved):
LGTM. This is some great work. Still need to test this with Open Hexagon but I think it's worth merging now and fixing any issues that arise later.

PR #3030, opened 2024-05-21
Remove default empty state of `sf::InputSoundFile` and `sf::OutputSoundFile`

ChrisThrasher:
## Description

The one place this got a little messy was `sf::Music`. Its default constructor requires all data members to also be default constructible so I had to make `m_file` optional. In the future this complication should go away once we remove `sf::Music` default constructor. I added asserts to ensure we don't try to dereference a null optional.

### DIFF
--- a/include/SFML/Audio/InputSoundFile.hpp
+++ b/include/SFML/Audio/InputSoundFile.hpp
@@ -33,6 +33,7 @@
 
 #include <filesystem>
 #include <memory>
+#include <optional>
 #include <vector>
 
 #include <cstddef>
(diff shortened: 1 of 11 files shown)


vittorioromeo (changes requested):
Great work as usual. Like prior PRs, let's avoid default constructors if we know that the types must have their members initialized to be valid, even if the constructors are private.

include/SFML/Audio/InputSoundFile.hpp
```diff
@@ -211,6 +212,14 @@ class SFML_AUDIO_API InputSoundFile
     void close();
 
 private:
+    ////////////////////////////////////////////////////////////
+    /// \brief Default constructor
+    ///
+    /// Useful for implementing close()
+    ///
+    ////////////////////////////////////////////////////////////
+    InputSoundFile() = default;
```
vittorioromeo, line 221: I'm really not a fan of this but I can try to improve it in a separate PR.

vittorioromeo (approved):
LGTM.

include/SFML/Audio/InputSoundFile.hpp
ChrisThrasher: I'm open to better alternatives for implementing `close()` that don't require having to list all data members.

PR #3033, opened 2024-05-22
Use '{}' for default parameter initialization

vittorioromeo:
Small refactoring. `ContextSettings` could not be adjusted because it's `explicit`. `ContextSettings` maybe should be a simple aggregate struct...


vittorioromeo:
@ChrisThrasher: is there a way to only restart failed coverage jobs?

eXpl0it3r:
Once all GitHub Actions jobs are through, you can rerun the failed jobs.

ChrisThrasher (approved):

PR #3034, opened 2024-05-22
Turn 'ContextSettings' into an aggregate and update usages

vittorioromeo:
Another refactoring/simplification PR on the same line as #3033.

### DIFF
--- a/src/SFML/Window/WindowBase.cpp
+++ b/src/SFML/Window/WindowBase.cpp
@@ -95,7 +95,17 @@ void WindowBase::create(VideoMode mode, const String& title, std::uint32_t style
     WindowBase::create(mode, style, state);
 
     // Recreate the window implementation
-    m_impl = priv::WindowImpl::create(mode, title, style, state, ContextSettings(0, 0, 0, 0, 0, 0xFFFFFFFF, false));
+    m_impl = priv::WindowImpl::create(mode,
+                                      title,
+                                      style,
+                                      state,
+                                      ContextSettings{/* depthBits */ 0,
+                                                      /* stencilBits */ 0,
+                                                      /* antialiasingLevel */ 0,
+                                                      /* majorVersion */ 0,
+                                                      /* minorVersion */ 0,
+                                                      /* attributeFlags */ 0xFFFFFFFF,
+                                                      /* sRgbCapable */ false});
 
     // Perform common initializations
     initialize();
--- a/examples/stencil/Stencil.cpp
+++ b/examples/stencil/Stencil.cpp
@@ -19,7 +19,7 @@ int main()
                             "SFML Stencil",
                             sf::Style::Titlebar | sf::Style::Close,
                             sf::State::Windowed,
-                            sf::ContextSettings(0, 8));
+                            sf::ContextSettings{0 /* depthBits */, 8 /* stencilBits */});
     window.setVerticalSyncEnabled(true);
 
     sf::RectangleShape red({500, 50});
--- a/include/SFML/Graphics/RenderTexture.hpp
+++ b/include/SFML/Graphics/RenderTexture.hpp
@@ -101,8 +101,7 @@ public:
     /// \return Render texture if creation has been successful, otherwise `std::nullopt`
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] static std::optional<RenderTexture> create(const Vector2u&        size,
-                                                             const ContextSettings& settings = ContextSettings());
+    [[nodiscard]] static std::optional<RenderTexture> create(const Vector2u& size, const ContextSettings& settings = {});
 
     ////////////////////////////////////////////////////////////
     /// \brief Get the maximum anti-aliasing level supported by the system
--- a/include/SFML/Graphics/StencilMode.hpp
+++ b/include/SFML/Graphics/StencilMode.hpp
@@ -214,7 +214,7 @@ SFML_GRAPHICS_API bool operator!=(const StencilMode& left, const StencilMode& ri
 /// Usage example:
 /// \code
 /// // Make sure we create a RenderTarget with a stencil buffer by specifying it via the context settings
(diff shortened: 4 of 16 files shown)


src/SFML/Window/WindowBase.cpp
```diff
@@ -95,7 +95,17 @@ void WindowBase::create(VideoMode mode, const String& title, std::uint32_t style
...
+                                      title,
+                                      style,
+                                      state,
+                                      ContextSettings{/* depthBits */ 0,
+                                                      /* stencilBits */ 0,
+                                                      /* antialiasingLevel */ 0,
+                                                      /* majorVersion */ 0,
+                                                      /* minorVersion */ 0,
+                                                      /* attributeFlags */ 0xFFFFFFFF,
+                                                      /* sRgbCapable */ false});
```
eXpl0it3r, line 108: Not sure, I'm a fan of this.

It doesn't look particularly nice and is rather brittle. Renaming will not update the comments, changing the field order will lead to silent failures. We're making the filed order part of the public API.

ChrisThrasher (commented):
Not a fan of all the new `/* comments */` but I think it's reasonable to convert `ContextSettings` to an aggregate. There is not especially much value to be had in its existing constructor. SFML 3 already has changed a number of data types into aggregates so this is certainly not unprecedented. Users can take advantage of C++20 designated initializers if they want to name the data members while initializing the structure.

src/SFML/Window/WindowBase.cpp
vittorioromeo: From the code above, I have no idea what all the specified values mean unless I go and check the definition of `ContextSettings` and I carefully match the arguments provided here with the fields of `ContextSettings` by manually counting.

Adding the comments provides clarity on what these values actually mean. In C++20, designated initializers would be even better, but we're targeting C++17.

> Renaming will not update the comments, changing the field order will lead to silent failures. We're making the filed order part of the public API.

These arguments are all extremely weak. The field order and the names of the field **are** part of the public API, because `ContextSettings` is now an aggregate!

If we ever decide to change the name of a field or the field order, keeping some comments up-to-date is going to be the least of our worries because we will be breaking existing users and potentially changing run-time behavior silently, without causing a compilation failure.

vittorioromeo:
Rebased on `master`. I still think that the comments add value and do not introduce any additional vulnerability, see https://github.com/SFML/SFML/pull/3034#discussion_r1612486369 for rationale.

vittorioromeo:

ChrisThrasher (approved):
I'd still prefer you not add those extra `/* memberName */` comments since I think they're distracting and there is no established precedent for doing it inside this codebase. Otherwise I'm happy with this.

src/SFML/Window/WindowBase.cpp
eXpl0it3r: > Adding the comments provides clarity on what these values actually mean. In C++20, designated initializers would be even better, but we're targeting C++17.

And why is this needed now and not before with a constructor?

> These arguments are not compelling. The field order and the names of the field are part of the public API, because `ContextSettings` is now an aggregate!

By pointing out that I'm not sure the field order should be part of the API, I'm of course indirectly challenging whether it should be an aggregate...

Are there any benefits beyond just having a different way to construct it?

vittorioromeo: > And why is this needed now and not before with a constructor?

It's not needed now, I would have also done the same if we kept the constructor. It's mostly a drive-by small change that I realized would improve readability as I was adapting the existing uses of `ContextSettings` to being an aggregate.

> By pointing out that I'm not sure the field order should be part of the API, I'm of course indirectly challenging whether it should be an aggregate...

Sorry, it wasn't clear because you specifically selected the `/* ... */` comments in your review and the rest of your feedback was directed specifically to the comments.

> Are there any benefits beyond just having a different way to construct it?

Yes: C++20 projects can use designated initializers, aggregates can be automatically destructured and reflected on, the implementation is much simpler.

Prior to this PR `ContextSettings` was essentially used like an aggregate without being one, as its constructor had all its arguments defaulted. 

Please let me know if you're OK with moving forward before merging.

vittorioromeo:
> I'd still prefer you not add those extra `/* memberName */` comments since I think they're distracting and there is no established precedent for doing it inside this codebase. Otherwise I'm happy with this.

Why do you think they're distracting? Their purpose is to add local context when reading the code, without having to remember/check what the `n`th field of `ContextSetting` is. We do have precedent in SFML's source code, e.g.:

```cpp
static constexpr ma_data_source_vtable vtable{read, seek, getFormat, getCursor, getLength, setLooping, /* flags */ 0};
```

We also use the same comment style for every elided parameter name, which is not exactly the same, but the purpose is very similar (aiding local readability), e.g.:

```cpp
std::optional<Shader> Shader::loadFromFile(const std::filesystem::path& /* vertexShaderFilename */,
                                           const std::filesystem::path& /* fragmentShaderFilename */)
{
    return std::nullopt;
}
```

ChrisThrasher (changes requested):

ChrisThrasher:

src/SFML/Window/WindowBase.cpp
eXpl0it3r: I take a bit issue with doing a change with the main argument that someone using the latest compile could potentially benefit from it.

Personally, I don't see something wrong with having a constructor on an aggregate type.

But I'm okay with the change.

vittorioromeo: > I take a bit issue with doing a change with the main argument that someone using the latest compile could potentially benefit from it.

Perhaps I was not very clear -- what I meant to say is that making `ContextSettings` an aggregate has the extra benefit of allowing users targeting C++20 to use designated initializers, but most importantly it's a simplification of our implementation and API. 
 
> Personally, I don't see something wrong with having a constructor on an aggregate type.

An aggregate type, by definition, cannot have any user-declared constructor.

PR #3035, opened 2024-05-22
Fix sound file docs

ChrisThrasher:
## Description

Oversight from #3030

### DIFF
--- a/include/SFML/Audio/InputSoundFile.hpp
+++ b/include/SFML/Audio/InputSoundFile.hpp
@@ -275,9 +275,7 @@ private:
 /// Usage example:
 /// \code
 /// // Open a sound file
-/// sf::InputSoundFile file;
-/// if (!file.openFromFile("music.ogg"))
-///     /* error */;
+/// auto file = sf::InputSoundFile::openFromFile("music.ogg").value();
 ///
 /// // Print the sound attributes
 /// std::cout << "duration: " << file.getDuration().asSeconds() << '\n'
--- a/include/SFML/Audio/OutputSoundFile.hpp
+++ b/include/SFML/Audio/OutputSoundFile.hpp
@@ -111,9 +111,7 @@ private:
 /// Usage example:
 /// \code
 /// // Create a sound file, ogg/vorbis format, 44100 Hz, stereo
-/// sf::OutputSoundFile file;
-/// if (!file.openFromFile("music.ogg", 44100, 2))
-///     /* error */;
+/// auto file = sf::OutputSoundFile::openFromFile("music.ogg", 44100, 2).value();
 ///
 /// while (...)
 /// {


binary1248 (approved):

include/SFML/Audio/InputSoundFile.hpp
```diff
@@ -275,9 +275,7 @@ class SFML_AUDIO_API InputSoundFile
 /// Usage example:
 /// \code
 /// // Open a sound file
-/// sf::InputSoundFile file;
-/// if (!file.openFromFile("music.ogg"))
-///     /* error */;
+/// auto file = sf::InputSoundFile::openFromFile("music.ogg").value();
```
eXpl0it3r, line 278: But this isn't exactly equivalent. Previously it showed that younhad to handle an error, now it would just let you run into a crash if it didn't load correctly

ChrisThrasher: This is what we've done in all the previous PRs. It's what I'd recommend to users by default since by they should be made immediately aware of such errors. More advanced users will no doubt use the `std::optional` API as they wish.

eXpl0it3r: That's a bad practice though, no? It's like retrieving a pointer from a function and it may return a nullptr, but you never check it and just use it. Sure, you don't get the guaranteed failure, but you have the chance to check for validity and don't do it.

Why would we recommend people to not check the validity of the optional and just have the application crash (i.e. throw `std::bad_optional_access`)?

ChrisThrasher: > but you have the chance to check for validity and don't do it.

That's the thing. This code _is_ checking for validity. I'm not sure why you think code somehow constitutes not checking for an error state. All our examples already do basically the same thing. Upon failure to load from disk, the program immediately exits. What I wrote is a convenient shorthand for what our examples already recommend.

ChrisThrasher: I believe @vittorioromeo is on the same page as I am. Perhaps he has another perspective on why this is an appropriate error handling technique to put in our examples and our documentation.

eXpl0it3r: Sure, it checks, but if it doesn't have a value, it throws an exception ... which we aren't handling or showing how to handle.

> If `*this` contains a value, returns a reference to the contained value.
> Otherwise, throws a `std::bad_optional_access` exception.
> - https://en.cppreference.com/w/cpp/utility/optional/value

The behavior is defined, but we're not actually showing how users should be writing error handling code, which is pretty clear since the new comments do not have a `// error` or even branching.

If you consider throwing an exception a good enough error handling strategy, I wonder why we aren't throwing exceptions ourselves? There's no point in using `std::optional` if in general we recommend to directly call `value()` on it.

ChrisThrasher: > which we aren't handling or showing how to handle.

We don't have to. Their program will terminate after which they can read the console logs from `sf::err` to see that their asset failed to load.

> but we're not actually showing how users should be writing error handling code

This _is_ how I want users writing error handling code when they're getting started.

> If you consider throwing an exception a good enough error handling strategy, I wonder why we aren't throwing exceptions ourselves? There's no point in using std::optional if in general we recommend to directly call value() on it.

There is so much value in `std::optional`. There are other valid ways to handle these errors and it's important we allow those alternatives to exist. However, our docs can't spend 20 lines enumerating 4 different ways of handling errors. That is overkill. We must pick one default choice and I picked the shortest option that still avoids UB in all circumstances.

PR #3032, opened 2024-05-21
Remove default empty state of `sf::RenderTexture`

ChrisThrasher:
## Description

Merging this first may simplify some things about #3027.

PR #3037, opened 2024-05-23
Simplify implementations of types without an empty state

ChrisThrasher:
## Description

Now that these types have no default empty state, we can make assumptions about certain data members which allow us to skip doing certain checks.

I discovered this while iterating on #3032. It was something I overlooked in previous PRs.

### DIFF
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -301,13 +301,13 @@ const Font::Info& Font::getInfo() const
 ////////////////////////////////////////////////////////////
 const Glyph& Font::getGlyph(std::uint32_t codePoint, unsigned int characterSize, bool bold, float outlineThickness) const
 {
+    assert(m_fontHandles);
+
     // Get the page corresponding to the character size
     GlyphTable& glyphs = loadPage(characterSize).glyphs;
 
     // Build the key by combining the glyph index (based on code point), bold flag, and outline thickness
-    const std::uint64_t key = combine(outlineThickness,
-                                      bold,
-                                      FT_Get_Char_Index(m_fontHandles ? m_fontHandles->face : nullptr, codePoint));
+    const std::uint64_t key = combine(outlineThickness, bold, FT_Get_Char_Index(m_fontHandles->face, codePoint));
 
     // Search the glyph into the cache
     if (const auto it = glyphs.find(key); it != glyphs.end())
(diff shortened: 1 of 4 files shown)


vittorioromeo (changes requested):
This is a good change and a sign that removing empty states was the right thing to do, but we should replace every check with an `assert` rather than removing it outright.

Even if we know that, e.g., `m_shaderProgram` cannot be empty in `setUniform`, its type still allows the empty state as part of its domain.

I am happy to approve if the assertions are added.

ChrisThrasher:
> I am happy to approve if the assertions are added.

Sure I can add those asserts. In doing so I realized that the `Shader` destructor should still check for `m_shaderProgram` being zero since a moved-from shader sets that value to zero. Is it a bug to call `GLEXT_glDeleteObject(castToGlHandle(m_shaderProgram))` when `m_shaderProgarm` is zero? Idk. But I'd rather play it safe for now.

vittorioromeo:
> > I am happy to approve if the assertions are added.
> 
> Sure I can add those asserts. In doing so I realized that the `Shader` destructor should still check for `m_shaderProgram` being zero since a moved-from shader sets that value to zero. Is it a bug to call `GLEXT_glDeleteObject(castToGlHandle(m_shaderProgram))` when `m_shaderProgarm` is zero? Idk. But I'd rather play it safe for now.

I found this: https://developer.download.nvidia.com/opengl/specs/GL_ARB_shader_objects.txt

>   37) What should the behavior of glDeleteObject(0) be?
> 
>     DISCUSSION: It would be desirable to allow 0 as a valid input value, and
>     not raise an error.
> 
>     RESOLUTION: glDeleteObject() will silently ignore deleting the value 0.
>     This is consistent with glDeleteTextures() that also silently ignores
>     deleting the value 0.

src/SFML/Graphics/Font.cpp
```diff
@@ -412,7 +419,7 @@ float Font::getUnderlinePosition(unsigned int characterSize) const
 ////////////////////////////////////////////////////////////
 float Font::getUnderlineThickness(unsigned int characterSize) const
 {
-    FT_Face face = m_fontHandles ? m_fontHandles->face : nullptr;
+    FT_Face face = m_fontHandles->face;
```
vittorioromeo, line 422:
```suggestion
    assert(m_fontHandles);

    FT_Face face = m_fontHandles->face;
```

vittorioromeo (changes requested):
Missed one :)

vittorioromeo (approved):

PR #3039, opened 2024-05-23
Use `std::string_view` to construct `sf::Shader`s without allocating

ChrisThrasher:
## Description

Using a `const std::string&` parameter means that loading a shader from a string literal requires reallocating that entire string into a `std::string`. By using `std::string_view` we can avoid those allocations.

The way I pulled this off was by tweaking some OpenGL calls to provide the length of the string view rather than relying on a null terminator to signal the end of the string since string views do not include null terminators.

### DIFF
--- a/include/SFML/Graphics/Shader.hpp
+++ b/include/SFML/Graphics/Shader.hpp
@@ -36,6 +36,7 @@
 #include <filesystem>
 #include <optional>
 #include <string>
+#include <string_view>
 #include <unordered_map>
 
 #include <cstddef>
@@ -200,7 +201,7 @@ public:
     /// \see loadFromFile, loadFromStream
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] static std::optional<Shader> loadFromMemory(const std::string& shader, Type type);
+    [[nodiscard]] static std::optional<Shader> loadFromMemory(std::string_view shader, Type type);
 
     ////////////////////////////////////////////////////////////
     /// \brief Load both the vertex and fragment shaders from source codes in memory
@@ -221,8 +222,7 @@ public:
     /// \see loadFromFile, loadFromStream
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] static std::optional<Shader> loadFromMemory(const std::string& vertexShader,
-                                                              const std::string& fragmentShader);
+    [[nodiscard]] static std::optional<Shader> loadFromMemory(std::string_view vertexShader, std::string_view fragmentShader);
 
     ////////////////////////////////////////////////////////////
     /// \brief Load the vertex, geometry and fragment shaders from source codes in memory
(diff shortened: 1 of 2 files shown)


vittorioromeo:
There's a caveat with this -- `std::string_view` is not guaranteed to be `NULL`-terminated. Is there anything in our shader compiling code that assumes `NULL`-termination?

ChrisThrasher:
> There's a caveat with this -- `std::string_view` is not guaranteed to be `NULL`-terminated. Is there anything in our shader compiling code that assumes `NULL`-termination?

Yeah I mentioned this in the description.

> The way I pulled this off was by tweaking some OpenGL calls to provide the length of the string view rather than relying on a null terminator to signal the end of the string since string views do not include null terminators.

ChrisThrasher:

vittorioromeo (approved):
Ah, I missed that in the description, my bad! LGTM.

vittorioromeo:
BTW, is there anything similar that we can use for `getUniformLocation`?

ChrisThrasher:
> BTW, is there anything similar that we can use for `getUniformLocation`?

```cpp
int Shader::getUniformLocation(const std::string& name)
{
    // Check the cache
    if (const auto it = m_uniforms.find(name); it != m_uniforms.end())
    {
        // Already in cache, return it
        return it->second;
    }
    else
    {
        // Not in cache, request the location from OpenGL
        const int location = GLEXT_glGetUniformLocation(castToGlHandle(m_shaderProgram), name.c_str());
        m_uniforms.emplace(name, location);

        if (location == -1)
            err() << "Uniform " << std::quoted(name) << " not found in shader" << std::endl;

        return location;
    }
}
```

Seeing `m_uniform.emplace(name, location);` gave me pause since that line of code entails saving the string view to a member variable. I didn't do a thorough analysis but on first glance it seemed risky so I didn't bother changing it.

ChrisThrasher:

vittorioromeo:
> > BTW, is there anything similar that we can use for `getUniformLocation`?
> 
> ```c++
> int Shader::getUniformLocation(const std::string& name)
> {
>     // Check the cache
>     if (const auto it = m_uniforms.find(name); it != m_uniforms.end())
>     {
>         // Already in cache, return it
>         return it->second;
>     }
>     else
>     {
>         // Not in cache, request the location from OpenGL
>         const int location = GLEXT_glGetUniformLocation(castToGlHandle(m_shaderProgram), name.c_str());
>         m_uniforms.emplace(name, location);
> 
>         if (location == -1)
>             err() << "Uniform " << std::quoted(name) << " not found in shader" << std::endl;
> 
>         return location;
>     }
> }
> ```
> 
> Seeing `m_uniform.emplace(name, location);` gave me pause since that line of code entails saving the string view to a member variable. I didn't do a thorough analysis but on first glance it seemed risky so I didn't bother changing it.

The biggest problem there is `name.c_str()`, though... I didn't find a replacement of `GLEXT_glGetUniformLocation` that supports pointer + length.

PR #3046, opened 2024-05-25
Reduce pointer usage

ChrisThrasher:
## Description

MainAndroid.cpp uses pointers in many places where it's not necessary

### DIFF
--- a/src/SFML/Main/MainAndroid.cpp
+++ b/src/SFML/Main/MainAndroid.cpp
@@ -98,9 +98,9 @@ void onStart(ANativeActivity* /* activity */)
 }
 
 ////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


src/SFML/Main/MainAndroid.cpp
```diff
@@ -185,16 +183,16 @@ void goToFullscreenMode(ANativeActivity* activity)
 }
 
 ////////////////////////////////////////////////////////////
-void getScreenSizeInPixels(ANativeActivity* activity, int* width, int* height)
+void getScreenSizeInPixels(ANativeActivity& activity, int* width, int* height)
```
vittorioromeo, line 186: Can you do the same for the two integers?

vittorioromeo (approved):

PR #3050, opened 2024-05-29
Use more type safe interfaces

ChrisThrasher:
## Description

`sf::Time::asMicroseconds`, `sf::Time::asMilliseconds`, and `sf::Time::asSeconds` are functions that break out of the type safety of `sf::Time`. Once you call these functions you're dealing with weakly typed numbers that can easily be misused.


eXpl0it3r (approved):

vittorioromeo (approved):
LGTM, this should also be more precise as internally it uses microseconds rather than milliseconds.

PR #3051, opened 2024-05-30
Restore Shader and Island examples back to equivalent pre-API-change functionality

binary1248:
Title.

It isn't very "friendly" when examples `abort` due to `bad_optional_access` being thrown. The mechanism that kicks in when shaders aren't supported already exists, it just has to be triggered like it was before.

Test on any system that doesn't support shaders (OpenGL ES systems basically).

### DIFF
--- a/examples/shader/Shader.cpp
+++ b/examples/shader/Shader.cpp
@@ -38,26 +38,29 @@ public:
             return false;
         m_sprite.emplace(m_texture);
 
-        m_shader.setUniform("texture", sf::Shader::CurrentTexture);
+        // Load the shader
+        if (!(m_shader = sf::Shader::loadFromFile("resources/pixelate.frag", sf::Shader::Type::Fragment)))
+            return false;
+        m_shader->setUniform("texture", sf::Shader::CurrentTexture);
 
         return true;
     }
 
     void onUpdate(float, float x, float y) override
     {
-        m_shader.setUniform("pixel_threshold", (x + y) / 30);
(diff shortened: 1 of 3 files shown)


ChrisThrasher (approved):

examples/shader/Shader.cpp
```diff
@@ -38,26 +38,29 @@ class Pixelate : public Effect
...
     void onUpdate(float, float x, float y) override
     {
-        m_shader.setUniform("pixel_threshold", (x + y) / 30);
+        m_shader->setUniform("pixel_threshold", (x + y) / 30);
     }
 
     void onDraw(sf::RenderTarget& target, sf::RenderStates states) const override
     {
-        states.shader = &m_shader;
+        states.shader = &*m_shader;
```
MarioLiebisch, line 56: I'm not sure about using `&*` to get a raw pointer here (and in the other instances), at least not in a tutorial/example context. For someone rather new this might feel confusing and unnecessary as you don't necessarily consider the overloaded operator in `std::optional`. Maybe use `&m_shader.value()` after all?

examples/shader/Shader.cpp
```diff
@@ -146,27 +150,28 @@ class StormBlink : public Effect
             m_points.append({{x, y}, {r, g, b}});
         }
 
-        return true;
+        // Load the shader
```
MarioLiebisch, line 153: Nitpick: Shader or Shader**s**?

PR #3052, opened 2024-05-30
Perform sanity checks on OpenGL extensions to make sure the required entry points are actually available

### LINKED ISSUE #2462: SFML crash on Galaxy S22

binary1248:
Title.

We've seen in #2814 and #2462 that some OpenGL (ES) implementations don't behave as expected and misreport extensions as being available  when they don't even provide the corresponding entry points. This change protects us from such implementations by making sure the required entry points are available in order to be able to assume that an extension is supported.

When the issue was originally reported, I had a hard time reproducing the issue, but recently I grabbed the ANGLE .dlls from a Google Chrome installation, dumped them into the binaries folder and ran the examples built using OpenGL ES, and it crashed in the same way as described in aforementioned issue. If you want to test these changes, the easiest way would be to do the same.

Closes #2814
Closes #2462

@Zombieschannel can you test this change to see if it really fixes the issue?

### DIFF
--- a/src/SFML/Graphics/GLExtensions.cpp
+++ b/src/SFML/Graphics/GLExtensions.cpp
@@ -53,6 +53,38 @@
 namespace sf::priv
 {
 ////////////////////////////////////////////////////////////
+void extensionSanityCheck()
+{
+    static const auto check = [](int& flag, auto... entryPoints)
+    {
+        // If a required entry point is missing, flag the whole extension as unavailable
+        if (!(entryPoints && ...))
+            flag = 0;
+    };
+#ifdef SFML_OPENGL_ES
+    check(GLEXT_multitexture_dependencies);
+    check(GLEXT_vertex_buffer_object_dependencies);
+    check(GLEXT_OES_blend_subtract_dependencies);
+    check(GLEXT_blend_func_separate_dependencies);
+    check(GLEXT_blend_equation_separate_dependencies);
+    check(GLEXT_framebuffer_object_dependencies);
+    check(GLEXT_EXT_blend_minmax_dependencies);
+#else
+    check(GLEXT_blend_minmax_dependencies);
+    check(GLEXT_multitexture_dependencies);
+    check(GLEXT_blend_func_separate_dependencies);
+    check(GLEXT_vertex_buffer_object_dependencies);
+    check(GLEXT_shader_objects_dependencies);
+    check(GLEXT_blend_equation_separate_dependencies);
+    check(GLEXT_framebuffer_object_dependencies);
+    check(GLEXT_framebuffer_blit_dependencies);
+    check(GLEXT_framebuffer_multisample_dependencies);
+    check(GLEXT_copy_buffer_dependencies);
+#endif
+}
+
+
+////////////////////////////////////////////////////////////
 void ensureExtensionsInit()
 {
     static bool initialized = false;
@@ -65,6 +97,13 @@ void ensureExtensionsInit()
 #else
         gladLoadGL(Context::getFunction);
 #endif
+        // Some GL implementations don't fully follow extension specifications
+        // and advertise support for extensions although not providing the
+        // entry points specified for the corresponding extension.
+        // In order to protect ourselves from such implementations, we perform
+        // a sanity check to ensure an extension is _really_ supported, even
+        // from an entry point perspective.
+        extensionSanityCheck();
 
         // Retrieve the context version number
         int majorVersion = 0;
--- a/src/SFML/Graphics/GLExtensions.hpp
+++ b/src/SFML/Graphics/GLExtensions.hpp
@@ -41,28 +41,43 @@
 // All functionality beyond that is optional
 // and has to be checked for prior to use
 
+namespace sf::priv
+{
+// Stand-ins for desktop OpenGL feature-check flags
+// NOLINTBEGIN(readability-identifier-naming)
+inline int SF_GL_OES_multitexture         = 1;
+inline int SF_GL_OES_vertex_buffer_object = 1;
+// NOLINTEND(readability-identifier-naming)
+} // namespace sf::priv
+
 // Core since 1.0
-#define GLEXT_multitexture           true
-#define GLEXT_texture_edge_clamp     true
-#define GLEXT_EXT_texture_edge_clamp true
-#define GLEXT_glClientActiveTexture  glClientActiveTexture
-#define GLEXT_glActiveTexture        glActiveTexture
-#define GLEXT_GL_TEXTURE0            GL_TEXTURE0
-#define GLEXT_GL_CLAMP               GL_CLAMP_TO_EDGE
-#define GLEXT_GL_CLAMP_TO_EDGE       GL_CLAMP_TO_EDGE
+#define GLEXT_multitexture          ::sf::priv::SF_GL_OES_multitexture
+#define GLEXT_glClientActiveTexture glClientActiveTexture
+#define GLEXT_glActiveTexture       glActiveTexture
+#define GLEXT_GL_TEXTURE0           GL_TEXTURE0
+
+#define GLEXT_multitexture_dependencies ::sf::priv::SF_GL_OES_multitexture, glClientActiveTexture, glActiveTexture
+
+// Core since 1.0
+#define GLEXT_texture_edge_clamp true
+#define GLEXT_GL_CLAMP           GL_CLAMP_TO_EDGE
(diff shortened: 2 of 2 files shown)


src/SFML/Graphics/GLExtensions.hpp
```diff
@@ -41,28 +41,43 @@
 // All functionality beyond that is optional
 // and has to be checked for prior to use
 
+namespace sf::priv
+{
+// Stand-ins for desktop OpenGL feature-check flags
+// NOLINTBEGIN(readability-identifier-naming)
+extern int SF_GL_OES_multitexture;
```
ChrisThrasher, line 48: Let’s use C++17 `inline` variables to avoid the need for `extern` plus a definition in a .cpp file.

ChrisThrasher (changes requested):

src/SFML/Graphics/GLExtensions.hpp
```diff
@@ -41,28 +41,43 @@
 // All functionality beyond that is optional
 // and has to be checked for prior to use
 
+namespace sf::priv
+{
+// Stand-ins for desktop OpenGL feature-check flags
+// NOLINTBEGIN(readability-identifier-naming)
+inline int SF_GL_OES_multitexture         = 1;
```
ChrisThrasher, line 48: Where do these variables get mutated? Otherwise LGTM

ChrisThrasher (approved):

src/SFML/Graphics/GLExtensions.hpp
```diff
@@ -41,28 +41,43 @@
...
+inline int SF_GL_OES_vertex_buffer_object = 1;
+// NOLINTEND(readability-identifier-naming)
+} // namespace sf::priv
+
+// Core since 1.0
+#define GLEXT_multitexture          sf::priv::SF_GL_OES_multitexture
```
vittorioromeo, line 54:
```suggestion
#define GLEXT_multitexture          ::sf::priv::SF_GL_OES_multitexture
```

I think that full qualification in macros is the clearest choice.

src/SFML/Graphics/GLExtensions.hpp
```diff
@@ -41,28 +41,43 @@
...
+#define GLEXT_glClientActiveTexture glClientActiveTexture
+#define GLEXT_glActiveTexture       glActiveTexture
+#define GLEXT_GL_TEXTURE0           GL_TEXTURE0
+
+#define GLEXT_multitexture_dependencies sf::priv::SF_GL_OES_multitexture, glClientActiveTexture, glActiveTexture
```
vittorioromeo, line 59:
```suggestion
#define GLEXT_multitexture_dependencies ::sf::priv::SF_GL_OES_multitexture, glClientActiveTexture, glActiveTexture
```

src/SFML/Graphics/GLExtensions.hpp
```diff
@@ -41,28 +41,43 @@
...
+#define GLEXT_GL_CLAMP_TO_EDGE   GL_CLAMP_TO_EDGE
 
 // Core since 1.1
 // 1.1 does not support GL_STREAM_DRAW so we just define it to GL_DYNAMIC_DRAW
-#define GLEXT_vertex_buffer_object true
-#define GLEXT_GL_ARRAY_BUFFER      GL_ARRAY_BUFFER
-#define GLEXT_GL_DYNAMIC_DRAW      GL_DYNAMIC_DRAW
-#define GLEXT_GL_STATIC_DRAW       GL_STATIC_DRAW
-#define GLEXT_GL_STREAM_DRAW       GL_DYNAMIC_DRAW
+#define GLEXT_vertex_buffer_object sf::priv::SF_GL_OES_vertex_buffer_object
```
vittorioromeo, line 68:
```suggestion
#define GLEXT_vertex_buffer_object ::sf::priv::SF_GL_OES_vertex_buffer_object
```

src/SFML/Graphics/GLExtensions.hpp
```diff
@@ -41,28 +41,43 @@
...
 #define GLEXT_glBufferSubData      glBufferSubData
 #define GLEXT_glDeleteBuffers      glDeleteBuffers
 #define GLEXT_glGenBuffers         glGenBuffers
+#define GLEXT_GL_ARRAY_BUFFER      GL_ARRAY_BUFFER
+#define GLEXT_GL_DYNAMIC_DRAW      GL_DYNAMIC_DRAW
+#define GLEXT_GL_STATIC_DRAW       GL_STATIC_DRAW
+#define GLEXT_GL_STREAM_DRAW       GL_DYNAMIC_DRAW
+
+#define GLEXT_vertex_buffer_object_dependencies \
+    sf::priv::SF_GL_OES_vertex_buffer_object, glBindBuffer, glBufferData, glBufferSubData, glDeleteBuffers, glGenBuffers
```
vittorioromeo, line 80:
```suggestion
    ::sf::priv::SF_GL_OES_vertex_buffer_object, glBindBuffer, glBufferData, glBufferSubData, glDeleteBuffers, glGenBuffers
```

src/SFML/Graphics/GLExtensions.cpp
```diff
@@ -52,6 +52,38 @@
 
 namespace sf::priv
 {
+////////////////////////////////////////////////////////////
+void extensionSanityCheck()
+{
+    static const auto check = [](int& flag, auto... entryPoints)
```
vittorioromeo, line 58: The `XXX_dependencies` macros aro only used in this file, right? Can we define them in this file rather than the `.hpp` to avoid leaking more macros? It would also make it more obvious why `entryPoints` is variadic.

vittorioromeo (changes requested):

src/SFML/Graphics/GLExtensions.cpp
binary1248: The idea is that they are defined right under the definition of the extension functionality so in case it has to be adjusted (add/remove functions) we would only have to make changes to 1 place (that fits on a single screen) instead of 2. Splitting them leads to a higher risk that something is accidentally left out. Also, `gl.h` already defines 1000s of macros, so us adding a handful more with carefully chosen prefixes shouldn't really matter that much.

vittorioromeo (approved):

binary1248:
The code is written. Someone else can port it back to C++98 code, but that someone isn't me. 😁

Zombieschannel:
> Would it make sense to backport this to SFML 2, since it was originally reported for it?
> 
> Also, still interested in hearing @Zombieschannel feedback on the fix

I'll test it as soon as I can, however, I don't physically have the device, my friend has it, so I'll maybe have answers in a few days but it might be longer. 😐 Sorry for the inconvenience.

Zombieschannel:
I have good and bad news.
The crashing issue seems to have been fixed with this PR.
Unfortunately, there is still something wrong with rendering.
I tested the basic green circle example, and while it does not crash, it shows a black screen.

Did everything work normally after these changes with those Chrome dlls?

binary1248:
I tested all the (applicable) examples myself using the Chrome ANGLE .dlls and they all worked as expected.

The changes here didn't affect any of the rendering logic at all, it stayed unchanged. The only thing that was changed is that extensions are disabled if we find out the implementation doesn't strictly follow their specifications by not providing the specified entry points.

If even less is rendered now than before, it might be a sign that there was more non-conforming stuff than we initially assumed. Like I said, the ANGLE build shipped with Chrome on Windows does not have any issues with these changes, so it might be something specific to the Samsung branch of ANGLE.

Maybe something is output to the console in debug which might help narrow down the issue. If nothing else helps you can try to step through the new code from this PR and try to find out exactly which extension is causing the issue.

PR #3057, opened 2024-06-03
Added raw mouse input support

binary1248:
This is a squashed/rebased/modernized version of #1383. Commit authorship information is retained. All credit goes to @sro5h.

For more information refer to #1383. Test on Linux/Windows using the provided example.

Since macOS doesn't support raw input as far as I can tell, raw mouse input events will only be generated on Windows and Linux.

Closes #1383.


ChrisThrasher:
Can you update the tests to test for this new event?

binary1248:
Updated.

examples/raw_input/RawInput.cpp
```diff
@@ -0,0 +1,82 @@
...
+{
+std::filesystem::path resourcesDir()
+{
+#ifdef SFML_SYSTEM_IOS
+    return "";
+#else
+    return "resources";
+#endif
+}
+} // namespace
```
vittorioromeo, line 19: Feel free to ignore but simply using `static` here would be less verbose:

```suggestion
static std::filesystem::path resourcesDir()
{
#ifdef SFML_SYSTEM_IOS
    return "";
#else
    return "resources";
#endif
}
```

examples/raw_input/RawInput.cpp
```diff
@@ -0,0 +1,82 @@
...
+    window.setVerticalSyncEnabled(true);
+
+    // Load the application font and pass it to the Effect class
+    const auto font = sf::Font::loadFromFile(resourcesDir() / "tuffy.ttf").value();
+
+    auto vectorToString = [](const auto& vector)
+    { return '(' + std::to_string(vector.x) + ", " + std::to_string(vector.y) + ')'; };
+
+    // Create the mouse position text
+    sf::Text mousePosition(font, "Mouse Position: " + vectorToString(sf::Mouse::getPosition()), 20);
```
vittorioromeo, line 41:
```suggestion
    sf::Text mousePosition(font, "", 20);
```

Probably not worth stringifying the position here as it will immediately be updated on the next frame, using an empty string here makes it so that we have one single place where the text is updated.

examples/raw_input/RawInput.cpp
```diff
@@ -0,0 +1,82 @@
...
+    mousePosition.setPosition({400.f, 300.f});
+    mousePosition.setFillColor(sf::Color::White);
+
+    // Create the mouse raw movement text
+    sf::Text mouseRawMovement(font, "", 20);
+    mouseRawMovement.setFillColor(sf::Color::White);
+
+    std::vector<sf::String> log;
```
vittorioromeo, line 49:
```suggestion
    std::vector<std::string> log;
```

Any reason why this is `sf::String`?

examples/raw_input/RawInput.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+    while (window.isOpen())
+    {
+        while (const auto event = window.pollEvent())
+        {
+            if (event.is<sf::Event::Closed>())
+                window.close();
+
+            if (const auto* const mouseMoved = event.getIf<sf::Event::MouseMoved>(); mouseMoved)
```
vittorioromeo, line 58:
```suggestion
            if (const auto* const mouseMoved = event.getIf<sf::Event::MouseMoved>())
```

examples/raw_input/RawInput.cpp
```diff
@@ -0,0 +1,82 @@
...
+                mousePosition.setString("Mouse Position: " + vectorToString(mouseMoved->position));
+
+            if (const auto* const mouseMovedRaw = event.getIf<sf::Event::MouseMovedRaw>(); mouseMovedRaw)
```
vittorioromeo, line 61:
```suggestion
            if (const auto* const mouseMovedRaw = event.getIf<sf::Event::MouseMovedRaw>())
```

examples/raw_input/RawInput.cpp
```diff
@@ -0,0 +1,82 @@
...
+{
+    // Create the main window
+    sf::RenderWindow window(sf::VideoMode({800u, 600u}), "SFML Raw Mouse Input", sf::Style::Titlebar | sf::Style::Close);
+    window.setVerticalSyncEnabled(true);
+
+    // Load the application font and pass it to the Effect class
+    const auto font = sf::Font::loadFromFile(resourcesDir() / "tuffy.ttf").value();
+
+    auto vectorToString = [](const auto& vector)
+    { return '(' + std::to_string(vector.x) + ", " + std::to_string(vector.y) + ')'; };
```
vittorioromeo, line 38:
```suggestion
    auto vec2ToString = [](const auto& vec2)
    { return '(' + std::to_string(vec2.x) + ", " + std::to_string(vec2.y) + ')'; };
```

Minor, just to avoid confusion with `std::vector`

examples/raw_input/RawInput.cpp
```diff
@@ -0,0 +1,82 @@
...
+    mouseRawMovement.setFillColor(sf::Color::White);
+
+    std::vector<sf::String> log;
+
+    while (window.isOpen())
+    {
+        while (const auto event = window.pollEvent())
+        {
+            if (event.is<sf::Event::Closed>())
+                window.close();
```
vittorioromeo, line 56:
```suggestion
            if (event.is<sf::Event::Closed>())
            {
                window.close();
                break;
            }
```

examples/raw_input/RawInput.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+                if (log.size() > 24u)
+                    log.erase(log.begin());
+            }
+        }
+
+        window.clear();
+        window.draw(mousePosition);
+
+        for (auto i = 0u; i < log.size(); ++i)
```
vittorioromeo, line 73:
```suggestion
        for (std::size_t i = 0; i < log.size(); ++i)
```

Not a big deal, but `0u` is `unsigned int` which does not match `log.size()`'s type. `std::size_t` does.

include/SFML/Window/Event.hpp
```diff
@@ -154,6 +154,15 @@ class SFML_WINDOW_API Event
         Vector2i position; //!< Position of the mouse pointer, relative to the top left of the owner window
     };
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Mouse move raw event
+    ///
```
vittorioromeo, line 159: Can we add some details on when/why this should be used compared to the "normal" mouse events?

Also, if this event is never generated on MacOS, it should be clearly documented here.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -367,6 +369,35 @@ bool isWMAbsolutePositionGood()
                        std::end(wmAbsPosGood),
                        [&](const sf::String& name) { return name == windowManagerName; });
 }
+
+// Initialize raw mouse input
+void initRawMouse(::Display* disp)
+{
+    auto opcode = 0;
+    auto event  = 0;
+    auto error  = 0;
```
vittorioromeo, line 378:
```suggestion
    int opcode = 0;
    int event  = 0;
    int error  = 0;
```

I don't see any benefit from using type inference here. I believe we've always been quite conservative with the use of auto and type deduction in SFML -- i.e. only use it when it provides a benefit and doesn't obfuscate the type.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -367,6 +369,35 @@ bool isWMAbsolutePositionGood()
...
+
+    if (XQueryExtension(disp, "XInputExtension", &opcode, &event, &error))
+    {
+        auto major = 2;
+        auto minor = 0;
```
vittorioromeo, line 383:
```suggestion
        int major = 2;
        int minor = 0;
```

Also, can these be `const`?

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -367,6 +369,35 @@ bool isWMAbsolutePositionGood()
                        std::end(wmAbsPosGood),
                        [&](const sf::String& name) { return name == windowManagerName; });
 }
+
+// Initialize raw mouse input
+void initRawMouse(::Display* disp)
```
vittorioromeo, line 374:
```suggestion
bool initRawMouse(::Display* disp)
```

Might be useful to propagate success/failure to the caller.

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -2025,6 +2060,33 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
...
+        // Raw input
+        case GenericEvent:
+        {
+            if (XGetEventData(m_display.get(), &windowEvent.xcookie))
+            {
+                if (windowEvent.xcookie.evtype == XI_RawMotion)
+                {
+                    const auto* rawEvent       = static_cast<const XIRawEvent*>(windowEvent.xcookie.data);
+                    auto        relativeValueX = 0;
+                    auto        relativeValueY = 0;
```
vittorioromeo, line 2073:
```suggestion
                    int         relativeValueX = 0;
                    int         relativeValueY = 0;
```

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -1085,6 +1103,23 @@ void WindowImplWin32::processEvent(UINT message, WPARAM wParam, LPARAM lParam)
...
+        case WM_INPUT:
+        {
+            RAWINPUT input;
+            UINT     size = sizeof(input);
+
+            GetRawInputData(reinterpret_cast<HRAWINPUT>(lParam), RID_INPUT, &input, &size, sizeof(RAWINPUTHEADER));
+
+            if (input.header.dwType == RIM_TYPEMOUSE)
+            {
+                if (const auto* rawMouse = &input.data.mouse; (rawMouse->usFlags & 0x01) == MOUSE_MOVE_RELATIVE)
```
vittorioromeo, line 1116:
```suggestion
                if (const RAWMOUSE* rawMouse = &input.data.mouse; (rawMouse->usFlags & 0x01) == MOUSE_MOVE_RELATIVE)
```

vittorioromeo (changes requested):

src/SFML/Window/Unix/WindowImplX11.cpp
binary1248: `XIQueryVersion` takes their pointers as in/out parameters, so they can't be `const`.

binary1248:
Updated with suggestions from @vittorioromeo.

examples/raw_input/RawInput.cpp
binary1248: Tried it, `clang-tidy` will complain due to `misc-use-anonymous-namespace`. Also, this is the way the function has been defined in other examples, probably for the same reason.

include/SFML/Window/Event.hpp
```diff
@@ -154,6 +154,39 @@ class SFML_WINDOW_API Event
...
+    /// operating system hence "raw". While the MouseMoved
+    /// position value is dependent on the screen resolution,
+    /// raw data is not. If the physical mouse is moved too
+    /// little to cause the screen cursor to move at least a
+    /// single pixel, no MouseMoved event will be generated. In
+    /// contrast, any movement information generated by the
+    /// mouse independent of its sensor resolution will always
+    /// generate a MouseMovedRaw event.
+    ///
+    /// In addiiton to screen resolution independence, raw
```
vittorioromeo, line 170:
```suggestion
    /// In addition to screen resolution independence, raw
```

Small typo, but the new description is really good and informative!!

vittorioromeo (approved):
Small typo, but approving anyway as it's fine by me after it's fixed. Good work :)

include/SFML/Window/Event.hpp
binary1248: Fixed.

examples/raw_input/RawInput.cpp
```diff
@@ -0,0 +1,85 @@
...
+int main()
+{
+    // Create the main window
+    sf::RenderWindow window(sf::VideoMode({800u, 600u}), "SFML Raw Mouse Input", sf::Style::Titlebar | sf::Style::Close);
+    window.setVerticalSyncEnabled(true);
+
+    // Load the application font and pass it to the Effect class
+    const auto font = sf::Font::loadFromFile(resourcesDir() / "tuffy.ttf").value();
+
+    auto vec2ToString = [](const auto& vec2)
```
ChrisThrasher, line 37:
```suggestion
    const auto vec2ToString = [](const sf::Vector2i& vec2)
```
There's no need for this function to be generic. We're only stringifiying this one template specialization. We can also mark this `const`. Might also be worth moving this closer to the two callsites. If we're going to use lambdas then we might as well minimize the scope of the lambda.

examples/raw_input/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+# all source files
+set(SRC RawInput.cpp)
+if(SFML_OS_IOS)
+    set(RESOURCES resources/tuffy.ttf)
+    set_source_files_properties(${RESOURCES} PROPERTIES MACOSX_PACKAGE_LOCATION Resources)
+endif()
```
ChrisThrasher, line 6: No need for this copy pasta `if(SFML_OS_IOS)` conditional when this example doesn't even get compiled on iOS.

examples/raw_input/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+
+# define the raw_input target
+sfml_add_example(raw_input GUI_APP
+                 SOURCES ${SRC}
+                 BUNDLE_RESOURCES ${RESOURCES}
+                 DEPENDS SFML::Graphics
+                 RESOURCES_DIR resources)
```
ChrisThrasher, line 13:
```suggestion
                 DEPENDS SFML::Graphics)
```

examples/raw_input/RawInput.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+namespace
+{
+std::filesystem::path resourcesDir()
+{
+#ifdef SFML_SYSTEM_IOS
+    return "";
+#else
+    return "resources";
+#endif
```
ChrisThrasher, line 17: `resourcesDir` can be removed since iOS is not supported

ChrisThrasher (changes requested):

binary1248:
Updated.

ChrisThrasher (approved):

include/SFML/Window/Event.hpp
```diff
@@ -154,6 +154,39 @@ class SFML_WINDOW_API Event
...
+    ///
+    /// Raw mouse movement data is intended for controlling
+    /// non-cursor movement, e.g. controlling the camera
+    /// orientation in a first person view, whereas MouseMoved
+    /// is intended primarily for controlling things related to
+    /// the screen cursor hence the additional processing
+    /// applied to it.
+    ///
+    /// Currently, raw mouse input events will only be generated
+    /// on Windows and Linux.
```
eXpl0it3r, line 182: I feel like this would require some further highlighting as a hint or whatever doxygen provides.

PR #3058, opened 2024-06-04
Add clang-tidy `readability-redundant-member-init` check

ChrisThrasher:
## Description

This has to be silenced for `sf::Vertex` because in some places within SFML we initialize only a subset of this aggregate type. If we remove the `{}` from `texCoords` then we get a compiler warning from Clang. It feels like these two clang-based tools are somewhat contradictory.
    
```
    error: missing field 'texCoords' initializer [-Werror,-Wmissing-field-initializers]
                m_points.append({{x, y}, {r, g, b}});
```

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -71,7 +71,6 @@ Checks: >
   -readability-magic-numbers,
   -readability-named-parameter,
   -readability-redundant-casting,
-  -readability-redundant-member-init,
   -readability-uppercase-literal-suffix,
 CheckOptions:
   - { key: readability-identifier-naming.ClassCase,           value: CamelCase }
--- a/include/SFML/Graphics/Vertex.hpp
+++ b/include/SFML/Graphics/Vertex.hpp
@@ -45,9 +45,9 @@ struct Vertex
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    Vector2f position{};          //!< 2D position of the vertex
+    Vector2f position;            //!< 2D position of the vertex
     Color    color{Color::White}; //!< Color of the vertex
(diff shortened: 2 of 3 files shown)


eXpl0it3r:
I don't see the usefulness of this change. It's not more or less readable because of (i.e. `readability-*`). On the contrary, I feel like it's confusing why some members do use `{}` and others don't.

In light of then having to silence other places, this doesn't seem worth it to me.

ChrisThrasher:
This PR enforces something we already have to bring up in code reviews. It will save time in future PRs. We will still be removing `{}` where possible in future PRs but will instead have to do so manually.

ChrisThrasher:
> I don't see how this helps with readability to be less consistent

This PR improves consistently. Go look at any class that contains a `Vector2`. It does not use `{}` and instead relies on the default constructor being called. We have always written code this way.

eXpl0it3r (approved):
My arguments weren't specific to the existing SFML code base. Personally, I don't see how this flag *in general* makes things more readable. You'll now always have to know which types have or don't have a default constructor, without really gaining anything at the point of initialization. 🤷

ChrisThrasher:
> You'll now always have to know which types have or don't have a default constructor

No you won't. clang-tidy will tell you. For built-in types that lack default values or C structures, clang-tidy will tell you to provide a default value. For types with default constructors, clang-tidy will let you know if you redundantly call the default constructor. The whole point of this tooling is that it reduces how much you have to keep in your head when writing code.

eXpl0it3r:
We seem to have different views on how to use tooling.

Tools should help me detect potential mistakes, but it should still be reasonably possible to write correct code without just writing something randomly and then waiting for tooling to correct you. As such I find it unreasonable if the default case is more complicated/requires more knowledge without actual benefit.

PR #3059, opened 2024-06-04
Assume X11 UTF-8 support is present

ChrisThrasher:
## Description

https://linux.die.net/man/3/xutf8setwmproperties

UTF-8 support was added by XFree86 v4.0.2 [which was released in the year 2000](https://en.wikipedia.org/wiki/XFree86#Release_history).

This line was added [here](https://github.com/SFML/SFML/commit/9996b7abb6c6e216dc40d527b0339c89880709c5#diff-3373daee8e5b8c4a748f9035abd44a47c467702a388e2887b095946ca6eaf6b2).


eXpl0it3r (approved):
I feel like it's reasonable to assume if you use SFML 3, that your X WM also supports UTF-8

PR #3061, opened 2024-06-04
Remove unnecessary static casts

ChrisThrasher:


test/Network/Packet.test.cpp
```diff
@@ -22,15 +22,15 @@
         CHECK(packet.getData() != nullptr);                  \
         CHECK(packet.getDataSize() == sizeof(expected));     \
         CHECK(!packet.endOfPacket());                        \
-        CHECK(static_cast<bool>(packet));                    \
+        CHECK(bool{packet});                                 \
```
eXpl0it3r, line 25: Didn't we decide against this style of casting?

ChrisThrasher: Not that I recall. In this case `bool{packet}` is performing a lossless cast because this syntax is forbidden from doing anything that throws away information. Because `sf::Packet` defines `operator bool` you can convert an `sf::Packet` to a boolean without a static cast. This `{}` syntax is less powerful than a static cast and we like to use the least powerful tool that solves the problem.

vittorioromeo: Hmm, I'm not too sure about this PR. 

`static_cast` denotes an ambiguous conscious decision to trigger a conversion. It adds clarity of intention. A side benefit is that it's also easier to spot/search every conversion, as you can simply grep for `static_cast`.

We would also go against this core guideline: [ES.49: If you must use a cast, use a named cast](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-casts-named).

This other guideline ([https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es64-use-the-tenotation-for-construction](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es64-use-the-tenotation-for-construction)) also suggests that `T{...}` should be used for construction, not conversion. 

I know that construction can be conversion, but semantically speaking constructing something and converting something are two different intentions that can be conveyed with separate tools, `T{...}` and `static_cast<T>(...)` respectively.

If we were to go down this route, why *ever* use `static_cast<T>(...)` instead of `T{...}`? What would be our policy to prefer one over the other?

I'd rather see an issue first discussing the pros/cons of both approaches, and then we can create a policy that we'd apply library-wide instead of in a few selected cases.

vittorioromeo (changes requested):

ChrisThrasher:
> why ever use static_cast<T>(...) instead of T{...}?

Because in many many circumstances the latter does not work.

vittorioromeo:
> > why ever use static_cast(...) instead of T{...}?
> 
> Because in many many circumstances the latter does not work. It won't even compile. I believe it was you, Vittorio, who said that C++ developers ought to use the least powerful tool possible to solve a given problem. `T{...}` is less powerful than `static_cast<T>(...)`.

That was indeed me :) 

I'm not against the rationale behind the change, but how about we compare pros/cons and then we decide to apply a policy on the entire library rather than just on `Packet.test.cpp`?

vittorioromeo:
> I have a history of PRs that remove unnecessary static casts so this is not unprecedented. I also already searched the codebase for static casts to bool and found no other such casts that can be simplified into a `bool{...}` conversion.

Why are you special-casing `bool`?

If you are indeed correct that `T{...}` should be preferred over `static_cast<T>{...}` whenever possible, I believe there are many places in SFML where we perform arithmetic conversions that could use your proposed syntax. Example from `Island.cpp`:


    const float scalingFactorX = static_cast<float>(windowWidth) / static_cast<float>(resolutionX);
    const float scalingFactorY = static_cast<float>(windowHeight) / static_cast<float>(resolutionY);
    
Could become:

    const float scalingFactorX = float{windowWidth} / float{resolutionX};
    const float scalingFactorY = float{windowHeight} / float{resolutionY};
    
I'd rather see an automated large-scale change than cherry-picked smaller ones.

vittorioromeo:
> Pros:
> 
> * Shorter
> * Safer
> * Simpler
> 
> Cons:
> 
> * Can't think of any

I'm not sure that "shorter" is a pro -- we want to draw attention to conversions. That's one possible con of using a more subtle syntax. As I also mentioned above one possible con is also losing greppability of conversions.

Are these deal-breakers? Very likely not, but it'd be nice to first decide our policy for the library based on all the facts.

ChrisThrasher:
> Why are you special-casing bool?

Because it's an easy place to start. This PR aims to be an improvement even if it doesn't fix 100% of the instances of this issue.

> I believe there are many places in SFML where we perform arithmetic conversions that could use your proposed syntax.

Awesome let's do that! No arguments from me. If it compiles then let's ship it.

ChrisThrasher:
> we want to draw attention to conversions

We want to draw attention to *lossy* conversions. This braced syntax as you are probably aware does not truncate data (unless you explicitly wrote an operator that truncates data). There is no reason to call attention to lossless operations.

binary1248:
Whenever I see an explicit cast, I always assume that the additional characters had to be typed because it is a narrowing conversion. If something is implicitly convertible without narrowing I normally just rely on the compiler doing the right thing, otherwise if I have to convince the compiler to perform an explicit non-narrowing conversion (pretty uncommon) I guess I would also prefer not using an explicit cast just so I can see it is non-narrowing from a glance.

vittorioromeo:
I'm OK with this change after putting some thought into it. Can you add a few more to this PR to get the ball rolling? I think that grepping for `static_cast<float>` yields quite a few places where this change could be made.

ChrisThrasher:
> I'm OK with this change after putting some thought into it. Can you add a few more to this PR to get the ball rolling? I think that grepping for `static_cast<float>` yields quite a few places where this change could be made.

I took a look at lots of different static casts and found more than can be removed.

vittorioromeo (approved):
LGTM, there are probably more but we can have another PR for those.

eXpl0it3r:
I'll need to get used to this syntax, but it's just an internal thing so why not. 😄

PR #3062, opened 2024-06-04
Rewrite `Shader.cpp` example

vittorioromeo:
Based on #3027


ChrisThrasher:
Rebased on `master` now that #3027 is merged. No changes made.

examples/shader/Shader.cpp
```diff
@@ -21,46 +21,58 @@ std::random_device rd;
 std::mt19937       rng(rd());
 } // namespace
 
+
+////////////////////////////////////////////////////////////
+// Base class for effects
+////////////////////////////////////////////////////////////
+struct Effect : public sf::Drawable
```
ChrisThrasher, line 28:
```suggestion
struct Effect : sf::Drawable
```
Did a quick audit of the codebase and it looks like in all circumstances except 1 (EnumArray) we omit `public` when a struct inherits from another type.

examples/shader/Shader.cpp
```diff
@@ -21,46 +21,58 @@ std::random_device rd;
...
-        // Load the shader
-        if (!(m_shader = sf::Shader::loadFromFile("resources/pixelate.frag", sf::Shader::Type::Fragment)))
-            return false;
-        m_shader->setUniform("texture", sf::Shader::CurrentTexture);
+        return Pixelate{std::move(*texture), std::move(*shader)};
+    }
 
-        return true;
+    explicit Pixelate(sf::Texture&& texture, sf::Shader&& shader) :
+    m_texture{std::move(texture)},
```
ChrisThrasher, line 54:
```suggestion
    m_texture(std::move(texture)),
```
By convention we use `( ... )` in initializer lists.

examples/shader/Shader.cpp
```diff
@@ -21,46 +21,58 @@ std::random_device rd;
...
 
-        // Load the shader
-        if (!(m_shader = sf::Shader::loadFromFile("resources/pixelate.frag", sf::Shader::Type::Fragment)))
-            return false;
-        m_shader->setUniform("texture", sf::Shader::CurrentTexture);
+        return Pixelate{std::move(*texture), std::move(*shader)};
+    }
 
-        return true;
+    explicit Pixelate(sf::Texture&& texture, sf::Shader&& shader) :
```
ChrisThrasher, line 53: Can we move all these new constructors into the `private:` section of the class?

Also it doesn't _really_ matter but by convention we don't use `explicit` on multi-parameter constructors so this is a bit inconsistent with the rest of the codebase.

examples/shader/Shader.cpp
```diff
@@ -343,28 +378,38 @@ class Geometry : public Effect
...
+                                   optionalToPtr(edgeEffect),
+                                   optionalToPtr(geometryEffect)};
 
-    // Initialize them
-    for (Effect* effect : effects)
-        effect->load();
+    const std::array<std::string, 5>
+        effectNames{"Pixelate", "Wave + Blur", "Storm + Blink", "Edge Post-effect", "Geometry Shader Billboards"};
+
+    std::size_t current = 0;
```
ChrisThrasher, line 412: Why did you get rid of the `getName` function? I think that's a better solution than a second array that owns the effect names.

ChrisThrasher (changes requested):
In general I like this change. I find this easier to reason about than the existing solution that relies on two-phase initialization.

examples/shader/Shader.cpp
vittorioromeo: Because now I'm only conditionally creating the `Effect` objects if their loading succeeds, otherwise I'm storing `nullptr` for the ones that failed to load. The name has to be stored separately because of that.

ChrisThrasher (approved):

vittorioromeo:
@ChrisThrasher: can you run the example locally? I'm getting an issue getting the first texture displayed but I'm not sure if I screwed up something or if my build environment has some sort of issue.

ChrisThrasher:
I'm getting weird behavior. First thing I get is this UBSan error:
```
/Users/thrasher/Projects/sfml/src/SFML/Graphics/RenderTarget.cpp:863:63: runtime error: load of value 136, which is not a valid value for type 'bool'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Users/thrasher/Projects/sfml/src/SFML/Graphics/RenderTarget.cpp:863:63 in 
Abort trap: 6
```
If I disable UBSan then the program will launch but the Pixelate and Edge Post-effect shaders don't work anymore. Here's a console log that may be of use
```
UNSUPPORTED (log once): POSSIBLE ISSUE: unit 0 GLD_TEXTURE_INDEX_2D is unloadable and bound to sampler type (Float) - using zero texture because texture unloadable
```

vittorioromeo:
@ChrisThrasher: thanks for confirming the weird behavior. My guess is some sort of misuse of `sf::Texture` or maybe a bug in `sf::Texture`'s move constructor

vittorioromeo:
@ChrisThrasher: the issue disappears when changing the `tryLoad` functions to use guaranteed RVO copy elision, e.g.:

```cpp
static std::optional<Pixelate> tryLoad()
{
    // ...
    return std::optional<Pixelate>{std::in_place, std::move(*texture), std::move(*shader)};
}
```

So the issue is definitely in either `Texture(Texture&&)`, or `Sprite(Sprite&&)`, or `Shader(Shader&&)`. 

It's very late here, so you're more than welcome to try to figure this out while I get some rest :) 

Otherwise I'll give it a shot tomorrow.

vittorioromeo:
> @ChrisThrasher: the issue disappears when changing the `tryLoad` functions to use guaranteed RVO copy elision, e.g.:
> 
> ```c++
> static std::optional<Pixelate> tryLoad()
> {
>     // ...
>     return std::optional<Pixelate>{std::in_place, std::move(*texture), std::move(*shader)};
> }
> ```
> 
> So the issue is definitely in either `Texture(Texture&&)`, or `Sprite(Sprite&&)`, or `Shader(Shader&&)`.
> 
> It's very late here, so you're more than welcome to try to figure this out while I get some rest :)
> 
> Otherwise I'll give it a shot tomorrow.

@ChrisThrasher: BTW, we should use this pattern in all our optional-returning functions. Basically, when we do

```cpp
std::optional<T> f()
{
    return T{args...};
}
```

we should instead do:

```cpp
std::optional<T> f()
{
    return std::optional<T>{std::in_place, args...};
}
```

The former always invokes a move constructor, the latter never does so -- it is guaranteed RVO copy elision. We should make this change library-wide.

vittorioromeo:
@ChrisThrasher: oh, I think I get it. `sf::Sprite` stores a `sf::Texture*`. But when the entirety of `Pixelate` gets moved, the pointer doesn't get updated. This is very unfortunate

vittorioromeo:

vittorioromeo:

ChrisThrasher:
> but it worries me that this sort of bug can end up unnoticed...

I suppose this is another instance of the existing API limitation which states that the texture must outline the sprite. Moving the texture implies ending the lifetime of the texture (at its original address) before the sprite's lifetime ends. I'm still bothered by how hard this is to detect though.

> we should instead do:
> ```
> std::optional<T> f()
> {
>     return std::optional<T>{std::in_place, args...};
> }
> ```

Fine by me. This pattern is new to me but it looks like a reasonable solution for use internally.

ChrisThrasher:
https://github.com/SFML/SFML/pull/3070

I tried testing this new `assert` with the older, buggy start of this PR and can't get the assert to fire so this doesn't seem to fix the problem. It may still be useful but it's not sufficient to catch the bug we were previously seeing.

PR #3063, opened 2024-06-04
Simplify `sf::Clock` usage

ChrisThrasher:
## Description

`sf::Clock::restart` exists so we might as well use it internally. In addition to this being less code, it also results in marginally less work being performed.


binary1248 (approved):

kimci86:
But restarting before or after the sleep is not the same!

ChrisThrasher:
Oops

PR #3067, opened 2024-06-05
Simplify `sf::Sprite` implementation and reduce branches

vittorioromeo:
Few various improvements:
- Initialize data members on construction, rather than default-constructing them and then assigning
- Remove unnecessary branches during construction
- Define simpler constructor via delegation
- Merge `updatePositions` and `updateTexCoords` in the same function, as they're always called together
- Remove unnecessary use of `auto`
- Remove redundant comment

### DIFF
--- a/include/SFML/Graphics/Sprite.hpp
+++ b/include/SFML/Graphics/Sprite.hpp
@@ -211,22 +211,16 @@ private:
     void draw(RenderTarget& target, RenderStates states) const override;
 
     ////////////////////////////////////////////////////////////
-    /// \brief Update the vertices' positions
+    /// \brief Update the vertices' positions and texture coordinates
     ///
     ////////////////////////////////////////////////////////////
-    void updatePositions();
-
-    ////////////////////////////////////////////////////////////
-    /// \brief Update the vertices' texture coordinates
-    ///
-    ////////////////////////////////////////////////////////////
-    void updateTexCoords();
+    void updateVertices();
 
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
     std::array<Vertex, 4> m_vertices;    //!< Vertices defining the sprite's geometry
-    const Texture*        m_texture{};   //!< Texture of the sprite
+    const Texture*        m_texture;     //!< Texture of the sprite
     IntRect               m_textureRect; //!< Rectangle defining the area of the source texture to display
 };
 
@@ -271,9 +265,9 @@ private:
 ///
 /// // Create a sprite
 /// sf::Sprite sprite(texture);
-/// sprite.setTextureRect(sf::IntRect({10, 10}, {50, 30}));
-/// sprite.setColor(sf::Color(255, 255, 255, 200));
-/// sprite.setPosition({100, 25});
+/// sprite.setTextureRect({{10, 10}, {50, 30}});
+/// sprite.setColor({255, 255, 255, 200});
+/// sprite.setPosition({100.f, 25.f});
 ///
 /// // Draw it
 /// window.draw(sprite);
--- a/src/SFML/Graphics/Sprite.cpp
+++ b/src/SFML/Graphics/Sprite.cpp
@@ -30,25 +30,20 @@
 #include <SFML/Graphics/Texture.hpp>
 
 #include <cmath>
-#include <cstdlib>
 
 
 namespace sf
 {
 ////////////////////////////////////////////////////////////
-Sprite::Sprite(const Texture& texture)
+Sprite::Sprite(const Texture& texture) : Sprite(texture, IntRect({0, 0}, Vector2i(texture.getSize())))
 {
-    setTexture(texture, true);
 }
 
 
 ////////////////////////////////////////////////////////////
-Sprite::Sprite(const Texture& texture, const IntRect& rectangle)
+Sprite::Sprite(const Texture& texture, const IntRect& rectangle) : m_texture(&texture), m_textureRect(rectangle)
 {
-    // Compute the texture area
-    setTextureRect(rectangle);
-    // Assign texture
-    setTexture(texture, false);
+    updateVertices();
 }
 
 
@@ -57,9 +52,7 @@ void Sprite::setTexture(const Texture& texture, bool resetRect)
 {
     // Recompute the texture area if requested
     if (resetRect)
-    {
         setTextureRect(IntRect({0, 0}, Vector2i(texture.getSize())));
-    }
 
     // Assign the new texture
     m_texture = &texture;
@@ -72,8 +65,7 @@ void Sprite::setTextureRect(const IntRect& rectangle)
     if (rectangle != m_textureRect)
     {
         m_textureRect = rectangle;
-        updatePositions();
-        updateTexCoords();
+        updateVertices();
     }
 }
 
@@ -81,8 +73,7 @@ void Sprite::setTextureRect(const IntRect& rectangle)
 ////////////////////////////////////////////////////////////
 void Sprite::setColor(const Color& color)
 {
-    // Update the vertices' color
-    for (auto& vertex : m_vertices)
+    for (Vertex& vertex : m_vertices)
         vertex.color = color;
 }
 
@@ -111,10 +102,8 @@ const Color& Sprite::getColor() const
 ////////////////////////////////////////////////////////////
 FloatRect Sprite::getLocalBounds() const
 {
-    const auto width  = static_cast<float>(std::abs(m_textureRect.width));
-    const auto height = static_cast<float>(std::abs(m_textureRect.height));
-
-    return {{0.f, 0.f}, {width, height}};
+    // Last vertex posiion is equal to texture rect size absolute value
+    return {{0.f, 0.f}, m_vertices[3].position};
 }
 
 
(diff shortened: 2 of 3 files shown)


vittorioromeo:

include/SFML/Graphics/Sprite.hpp
```diff
@@ -108,7 +108,7 @@ class SFML_GRAPHICS_API Sprite : public Drawable, public Transformable
     /// \brief Disallow setting from a temporary texture
     ///
     ////////////////////////////////////////////////////////////
-    void setTexture(Texture&& texture, bool resetRect = false) = delete;
+    void setTexture(Texture&& texture, bool resetRect) = delete;
```
kimci86, line 111: This is going against the purpose of this declaration. Calling setTexture with an rvalue reference but no boolean argument will not match this signature, so the setTexture taking const reference will be used instead, missing the opportunity to show an error.

vittorioromeo: Good catch, I didn't think that one through! Fixing...

vittorioromeo:

kimci86 (approved):

src/SFML/Graphics/Sprite.cpp
```diff
@@ -36,19 +36,15 @@
...
 
 ////////////////////////////////////////////////////////////
-Sprite::Sprite(const Texture& texture, const IntRect& rectangle)
+Sprite::Sprite(const Texture& texture, const IntRect& rectangle) : m_texture(&texture), m_textureRect(rectangle)
 {
-    // Compute the texture area
-    setTextureRect(rectangle);
-    // Assign texture
-    setTexture(texture, false);
+    updateVertices();
```
ChrisThrasher, line 47: Before this PR `updateVertices()` was not called when using the `const Texture& texture` overload. It was also not called when the 2-parameter was called with a zero-size rectangle. Now `updateVertices` is called unconditionally. This is not incorrect but it's a bit unnecessary. Is it worth adding an `if (m_textureRect != IntRect())` conditional here to skip calling `updateVertices` when it is not needed?

src/SFML/Graphics/Sprite.cpp
```diff
@@ -72,17 +68,15 @@ void Sprite::setTextureRect(const IntRect& rectangle)
...
     }
 }
 
 
 ////////////////////////////////////////////////////////////
 void Sprite::setColor(const Color& color)
 {
-    // Update the vertices' color
-    for (auto& vertex : m_vertices)
+    for (Vertex& vertex : m_vertices)
```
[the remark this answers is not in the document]
vittorioromeo: It absolutely is a simplification, as we were using a more powerful tool (type inference) in a situation where a more limited one (simple concrete type) suffices -- it's not about "style" at all, "style" is picking Allman brace style over K&R brace style. 

Type deduction is not necessary here, nor it reduces repetition, nor it increases readability. All it does is obfuscate the type of `vertex`. As you reminded me of [this principle yourself](https://github.com/SFML/SFML/pull/3061#issuecomment-2147960874), now it is my turn to remind you of the same philosophy.

Opening up an issue to discuss what has been [already discussed and decided upon](https://github.com/SFML/SFML/pull/3061#issuecomment-2147960874)...

> type inference -- `auto` and `decltype` should only be used where they considerably increase readability (e.g. iterators), lack of types for basic variables can make code harder to understand

...is a complete waste of time, and this PR is perhaps being "slowed down" only because you brought up a concern about a valid removal of `auto` that is in line with our Modern C++ feature usage policy.

src/SFML/Graphics/Sprite.cpp
vittorioromeo: You raise a good point, however... when would you ever create a zero-sized sprite?

Now that `Texture` is guaranteed to be non-empty due to #3027, I would expect `texture.getSize()` on L39 to never be `{0, 0}`, right?

Furthermore, what would be the point of creating a sprite with a texture and a zero-sized rectangle?

It seems to me that we were "optimizing" for a very niche scenario, and that such "optimization" might not even be one as it's an additional branch.

I agree with your reasoning in this comment https://github.com/SFML/SFML/pull/3026#issuecomment-2123624265 -- the complexity added by an extra branch like the one you're describing would need to be motivated by a compelling benchmark, and I don't think that benchmarking the creation of zero-sized sprites makes much sense, unless I am missing something...

src/SFML/Graphics/Sprite.cpp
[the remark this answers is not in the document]
eXpl0it3r: I agree with Chris and basically disagree with every points you make Vittorio, but I also fall into the [AAA (almost always auto) camp](https://duerrenberger.dev/blog/2021/11/19/almost-always-auto/), so I'm certainly very bias on the topic.

> It absolutely is a simplification, as we were using a more powerful tool (type inference) in a situation where a more limited one (simple concrete type) suffices -- it's not about "style" at all, "style" is picking Allman brace style over K&R brace style.

In the most obtuse, pedantic sense you can say yes it's not a "style" choice, but it falls into the same category. The change has no effect on the outcome as the compiler will do the work for you.

Either way it's a "stylistic", personal choice of how to use the language feature and there's little sense in trying to make technical arguments about it.

> Opening up an issue to discuss what has been [already discussed and decided upon](https://github.com/SFML/SFML/pull/3061#issuecomment-2147960874)...
> 
> > type inference -- `auto` and `decltype` should only be used where they considerably increase readability (e.g. iterators), lack of types for basic variables can make code harder to understand

I've no idea where that quote is from, but the linked discussion didn't talk about `auto` and `decltype` and certainly didn't decide or set on any policy.

> ...is a complete waste of time, and this PR is perhaps being "slowed down" only because you brought up a concern about a valid removal of `auto` that is in line with our Modern C++ feature usage policy.

Let's maybe not start with such blaming games. We should be open for any feedback, even if you don't like the topic.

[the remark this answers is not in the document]
vittorioromeo: > I agree with Chris and basically disagree with every points you make Vittorio

Of course...

> but I also fall into the [AAA (almost always auto) camp](https://duerrenberger.dev/blog/2021/11/19/almost-always-auto/), so I'm certainly very "bias" on the topic (i.e. holding a personal stylistic opinion).

Then you misundestand what AAA actually means.

The stylistical choice is:

```cpp
auto i0 = int{0};
int i1 = 0;
```

Those two are equivalent and only differ in style. `i0` is an example of AAA. 

It is absolutely not the same as the comparison betwen:

```cpp
auto i0 = 0;
int i1 = 0;
```

Because now `i0`'s type is not visible anymore and we are relying on type inference.

AAA does not imply type inference.

> > It absolutely is a simplification, as we were using a more powerful tool (type inference) in a situation where a more limited one (simple concrete type) suffices -- it's not about "style" at all, "style" is picking Allman brace style over K&R brace style.
> 
> In the most obtuse, pedantic sense you can say yes it's not a "style" choice, but it falls into the same category. The change has no effect on the outcome as the compiler will do the work for you.

Implying that any change that doesn't have an impact on the final generated binary is stylistical is wrong and dangerous.

We could use `auto&&` in most places where `auto` suffices without any change to the program's behavior. Is it style?

We could use `auto` in most places where `T` suffices. Is it style? 

> Either way it's a "stylistic", personal choice of how to use the language feature and there's little sense in trying to make technical arguments about it.

Absolutely not, there are real readability and usability impacts when using type inference over a concrete type, sometimes positively and sometimes negatively.

Technical arguments on the use of type inference have been made ever since `auto` was introduced and for good reason, as it is a powerful feature whose use should be judicious.

> > Opening up an issue to discuss what has been [already discussed and decided upon](https://github.com/SFML/SFML/pull/3061#issuecomment-2147960874)...
> > > type inference -- `auto` and `decltype` should only be used where they considerably increase readability (e.g. iterators), lack of types for basic variables can make code harder to understand
> 
> I've no idea where that quote is from, but the linked discussion didn't talk about `auto` and `decltype` and certainly didn't decide or set on any policy on how to use and when to use `auto`/type inference.

I meant to link to this. https://github.com/SFML/SFML/wiki/SD:-CPP-17-Support#implementation-focused

A page *you* created and revised. 😅

> > ...is a complete waste of time, and this PR is perhaps being "slowed down" only because you brought up a concern about a valid removal of `auto` that is in line with our Modern C++ feature usage policy.
> 
> Let's maybe not start with such blaming games. We should be open for any feedback, even if you don't like the topic.

The feedback could have been delivered without accusing my change of being stylistical (it isn't) and implying that I was wasting the reviewer's time.

Had the initial feedback been delivered purely on a technical basis I would have not responded in tone.

src/SFML/Graphics/Sprite.cpp
```diff
@@ -131,36 +122,35 @@ void Sprite::draw(RenderTarget& target, RenderStates states) const
...
-    const float right  = left + convertedTextureRect.width;
-    const float top    = convertedTextureRect.top;
-    const float bottom = top + convertedTextureRect.height;
-
-    m_vertices[0].texCoords = Vector2f(left, top);
-    m_vertices[1].texCoords = Vector2f(left, bottom);
-    m_vertices[2].texCoords = Vector2f(right, top);
-    m_vertices[3].texCoords = Vector2f(right, bottom);
+    assert(m_textureRect.width >= 0);
+    assert(m_textureRect.height >= 0);
```
vittorioromeo, line 134: I don't think this can ever be negative unless a user explicitly passes a negative texture rect in `setTextureRect`, because the constructor from `sf::Texture` converts the texture's size which is unsigned.

src/SFML/Graphics/Sprite.cpp
```diff
@@ -131,36 +122,35 @@ void Sprite::draw(RenderTarget& target, RenderStates states) const
...
+
+    const auto width  = static_cast<float>(m_textureRect.width);
+    const auto height = static_cast<float>(m_textureRect.height);
+    const auto left   = static_cast<float>(m_textureRect.left);
+    const auto top    = static_cast<float>(m_textureRect.top);
+    const auto right  = float{left + width};
+    const auto bottom = float{top + height};
```
vittorioromeo, line 141: I have to admit that @ChrisThrasher's point regarding using `T{...}` instead of `static_cast<T>(...)` was great, I was wrong in doubting it.

I can now easily tell that the first four conversions are lossy, while the last two are lossless.

src/SFML/Graphics/Sprite.cpp
eXpl0it3r: Which begs the question whether we should "allow" negative texture rects in the first place.

Since we don't allow it here, I think it would be better to just move the asserts into `setTextureRect` and adjust the documentation.

eXpl0it3r:

src/SFML/Graphics/Sprite.cpp
vittorioromeo: `setTextureRect` is not the only place where `m_textureRect` is set, there's also the constructor -- I decided to put the assertion here because both the constructor and `setTextureRect` end up calling `updateVertices`.

I will adjust the documentation to clarify that negative rects are not valid. I think the right solution here would be to accept a `UIntRect` instead of `IntRect`, but I'd prefer leaving that to a separate PR as it's a more breaking change than the ones presented here.

src/SFML/Graphics/Sprite.cpp
vittorioromeo: @ChrisThrasher: are you OK to resolve this comment? 

I also think that the whole "store `m_textureRect` and check if we need to recompute the vertices" is very unnecessary, it's a micro-optimization that might even be dentrimental to performance due to branching, and we're talking about `sf::Sprite` here which is already one of the slowest ways you could draw a quad to the screen.

src/SFML/Graphics/Sprite.cpp
kimci86: Negative texture rect size can be used to flip the represented portion of the texture horizontally and/or vertically. Why disallow this? It looks like a valid use case to me.

vittorioromeo: @kimci86: is this well-defined and/or documented anywhere in SFML? I searched a bit in the codebase but couldn't find anything. If it's something we want to support, we could document it in `setTextureRect` and the sprite ctor. Thoughts? @eXpl0it3r

kimci86: I thought that was a well-known behavior but now I don't see documentation or tutorial about it. I guess I saw the implementation a while ago and learned from that.

vittorioromeo: > I thought that was a well-known behavior but now I don't see documentation or tutorial about it. I guess I saw the implementation a while ago and learned from that.

Do you know if this works on every platform and is compatible with GLES2 and our eventual plans of moving to a more modern GL implementation?

I've briefly looked around and didn't find much information, I did see some people saying that GLES2 on Android might not support UVs outside [0..1] range, but that wasn't very conclusive either.

eXpl0it3r: I don't really see a reason to restrict it. I can in theory do the same with texture coordinates when using a vertex array, so why artificially restrict it here. Especially in light that it was previously not constraint.

If we want, we could potentially assert that `size - pos` doesn't land in a negative quadrant, which is the constraint I was thinking about, while not considering that position can of course be non-zero.

I don't think it's necessary to know what GLES2 will or will not support. It's and edge case limitation we can introduce at the time when it becomes a constraint.

ChrisThrasher (approved):

vittorioromeo:

Issue #3044, opened 2024-05-25
Support moving `sf::Music`s


ChrisThrasher:
### Describe your feature request here

PRs like #3021 require move operations so that the objects can be moved into and out of `std::optional`s. `sf::Music` needs to support move operations before we can remove its default constructor and switch to a factory function approach for constructing it.

I briefly looked into how we might achieve this but it wasn't obvious to me. We need to make sure when moving a music, the sound being played does not stop.

vittorioromeo:
PImpl `sf::Music`?

ChrisThrasher:
> PImpl `sf::Music`?

I'd prefer a solution that doesn't require PIMPL since I find it pretty annoying have to hide everything behind a pointer indirection but it certainly would work. My desire to finish the `std::optional`-ification of SFML 3 is greater than my annoyance with using PIMPL in one more place.

vittorioromeo:
Will be fixed by #3075.

ChrisThrasher:
Closed by #3075

PR #3068, opened 2024-06-05
Optimize `std::optional<T>` factory function returns for `T` with public ctor

vittorioromeo:
Basically, this pattern we frequently use

```cpp
class Widget
{
private:
    explicit Widget(Dependency&&);

public:
    static std::optional<Widget> make(std::filesystem::path p)
    {
        std::optional<Dependency> d = loadDependencyFromPath(p);
        if (!d.has_value()) { return std::nullopt; }
        
        return Widget{std::move(*d)};
    }
};
```

has a flaw -- it is not eligible for C++17 guaranteed copy elision (RVO) due to the fact that we're not returning a prvalue of type `std::optional<Widget>`, but a prvalue of type `Widget` instead, whose type does *not* match the function's return type.

That has a few practical drawbacks:

- Non-copyable/non-movable types cannot be returned in this way;
- Performance is sligthly worse, as extra move/copy operations will be invoked;
- Code that previously relied on address stability might break after switching to this pattern due to the extra move/copy operations (see #3062).

The way of improving the pattern is to return a prvalue of `std::optional<Widget>` type, thus gaining eligibility for guaranteed copy elision. E.g.,

```cpp
static std::optional<Widget> make(std::filesystem::path p)
{
    std::optional<Dependency> d = loadDependencyFromPath(p);
    if (!d.has_value()) { return std::nullopt; }
    
    return std::make_optional<Widget>(std::move(*d));
}
```

This works, but only if the constructor of `Widget` is `public`, as internally `std::make_optional` tries to construct `Widget` outside of `Widget` itself's scope. The same issue applies to `std::optional<T>{std::in_place, ...}`.

This PR aims to improve our usage of the factory pattern *only* for types where the constructor is public and where the extra copies/moves are not trivial. 

For all the other types, we have two options AFAICS:

1. Make the constructors `public`, allowing users to call them directly;
2. Make the constructors `public`, but using the passkey idiom to prevent users from calling them.

Number (2) seems like the best choice, but it adds a bit of extra boilerplate and we'd have to use some sort of Doxygen annotations to suppress documentation generation for those constructors.

I think it's worth doing as all the heavyweight types such as `Texture`, `Font`, et cetera, all have `private` constructors and use the inefficient factory pattern that does not support guaranteed copy elision. But that's for another PR.

See also: https://stackoverflow.com/questions/78581072/stdoptional-factory-function-with-guaranteed-copy-elision-and-private-cons

### DIFF
--- a/include/SFML/Graphics/Rect.inl
+++ b/include/SFML/Graphics/Rect.inl
@@ -109,7 +109,8 @@ constexpr std::optional<Rect<T>> Rect<T>::findIntersection(const Rect<T>& rectan
     // If the intersection is valid (positive non zero area), then there is an intersection
     if ((interLeft < interRight) && (interTop < interBottom))
     {
-        return Rect<T>({interLeft, interTop}, {interRight - interLeft, interBottom - interTop});
+        return std::make_optional<Rect<T>>(Vector2<T>{interLeft, interTop},
+                                           Vector2<T>{interRight - interLeft, interBottom - interTop});
     }
     else
     {
--- a/src/SFML/Audio/PlaybackDevice.cpp
+++ b/src/SFML/Audio/PlaybackDevice.cpp
@@ -54,7 +54,7 @@ std::optional<std::string> getDefaultDevice()
     for (const auto& device : priv::AudioDevice::getAvailableDevices())
     {
         if (device.isDefault)
-            return device.name;
+            return std::make_optional(device.name);
     }
 
     return std::nullopt;
--- a/src/SFML/Audio/SoundBuffer.cpp
+++ b/src/SFML/Audio/SoundBuffer.cpp
@@ -113,7 +113,7 @@ std::optional<SoundBuffer> SoundBuffer::loadFromSamples(
         // Update the internal buffer with the new samples
         if (!soundBuffer.update(channelCount, sampleRate, channelMap))
             return std::nullopt;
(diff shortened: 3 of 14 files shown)


vittorioromeo:

kimci86:
Is there a reason to use std::make_optional instead of std::optional constructor?

vittorioromeo:
> Is there a reason to use std::make_optional instead of std::optional constructor?

```cpp
std::make_optional<T>(args...)
// ...versus...
std::optional<T>(std::in_place, args...)
```

I find the former a bit more concise/readable than the latter.

ChrisThrasher:
> Non-copyable/non-movable types cannot be returned in this way;

Does this imply that I was mistaken about https://github.com/SFML/SFML/issues/3044?

vittorioromeo:
> > Non-copyable/non-movable types cannot be returned in this way;
> 
> Does this imply that I was mistaken about #3044?

Not really. You can indeed make a factory function work for a type that does not support move semantics:
https://gcc.godbolt.org/z/e7odWjdqj

But that's the extent of it. It will only work specifically for prvalues, it will not work with NRVO, you won't be able to move music in any other context or to move a class that contains a music, etc...

vittorioromeo:
Superseded by https://github.com/SFML/SFML/pull/3092.

PR #3069, opened 2024-06-05
Optimize remaining `std::optional<T>` factory function returns

vittorioromeo:
Follow-up from #3068, see that PR for context.

### DIFF
--- a/include/SFML/Audio/InputSoundFile.hpp
+++ b/include/SFML/Audio/InputSoundFile.hpp
@@ -31,6 +31,8 @@
 
 #include <SFML/Audio/SoundFileReader.hpp>
 
+#include <SFML/System/PassKey.hpp>
+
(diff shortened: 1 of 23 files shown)


vittorioromeo:
Alternative solution that doesn't require making the constructors public: https://godbolt.org/z/vcG89f4q4

Source:
https://twitter.com/foonathan/status/1798347550568976518

vittorioromeo:
> Alternative solution that doesn't require making the constructors public: https://godbolt.org/z/vcG89f4q4
> 
> Source: https://twitter.com/foonathan/status/1798347550568976518

Nevermind, this doesn't seem to work with most MSVC versions.

vittorioromeo:
Superseded by https://github.com/SFML/SFML/pull/3092.

PR #3071, opened 2024-06-06
More `Shader.cpp` example simplifications

vittorioromeo:
I realized that there's no reason for the `tryLoad` to be members of the effect classes, and that there's no reason for the effect classes' constructors to be `private`, as it makes sense to create these classes directly if the dependencies are available.

We can also wrap everything in the anonymous namespace, and not just the random generator, and provide the factory functions as free functions before `main`.

I also changed the factory functions to use `std::make_optional`, so they're now eligible for guaranteed copy elision.

Finally, I changed the clear color from relying on a string comparison to actually comparing the currently active example.

### DIFF
--- a/examples/shader/Shader.cpp
+++ b/examples/shader/Shader.cpp
@@ -18,7 +18,6 @@ namespace
 {
 std::random_device rd;
 std::mt19937       rng(rd());
-} // namespace
 
 
 ////////////////////////////////////////////////////////////
@@ -36,20 +35,6 @@ struct Effect : sf::Drawable
 class Pixelate : public Effect
 {
 public:
-    static std::optional<Pixelate> tryLoad()
-    {
-        auto texture = sf::Texture::loadFromFile("resources/background.jpg");
-        if (!texture.has_value())
-            return std::nullopt;
-
-        auto shader = sf::Shader::loadFromFile("resources/pixelate.frag", sf::Shader::Type::Fragment);
-        if (!shader.has_value())
-            return std::nullopt;
-
-        return Pixelate{std::move(*texture), std::move(*shader)};
-    }
-
-private:
     explicit Pixelate(sf::Texture&& texture, sf::Shader&& shader) :
     m_texture(std::move(texture)),
     m_shader(std::move(shader))
@@ -68,6 +53,7 @@ private:
         target.draw(sf::Sprite{m_texture}, states);
     }
 
+private:
     sf::Texture m_texture;
     sf::Shader  m_shader;
 };
@@ -79,15 +65,6 @@ private:
 class WaveBlur : public Effect
 {
 public:
-    static std::optional<WaveBlur> tryLoad(const sf::Font& font)
-    {
-        auto shader = sf::Shader::loadFromFile("resources/wave.vert", "resources/blur.frag");
-        if (!shader.has_value())
-            return std::nullopt;
-
-        return WaveBlur{font, std::move(*shader)};
-    }
-
     void update(float time, float x, float y) override
     {
         m_shader.setUniform("wave_phase", time);
@@ -101,7 +78,6 @@ public:
         target.draw(m_text, states);
     }
 
-private:
     explicit WaveBlur(const sf::Font& font, sf::Shader&& shader) :
     m_text(font,
            "Praesent suscipit augue in velit pulvinar hendrerit varius purus aliquam.\n"
@@ -128,6 +104,7 @@ private:
         m_text.setPosition({30.f, 20.f});
     }
 
+private:
     sf::Text   m_text;
     sf::Shader m_shader;
 };
@@ -139,15 +116,6 @@ private:
 class StormBlink : public Effect
 {
 public:
-    static std::optional<StormBlink> tryLoad()
-    {
-        auto shader = sf::Shader::loadFromFile("resources/storm.vert", "resources/blink.frag");
-        if (!shader.has_value())
-            return std::nullopt;
-
-        return StormBlink{std::move(*shader)};
-    }
-
     void update(float time, float x, float y) override
     {
         const float radius = 200 + std::cos(time) * 150;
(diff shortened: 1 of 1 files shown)


examples/shader/Shader.cpp
```diff
@@ -359,13 +265,115 @@ class Geometry : public Effect
...
+
+    // Set the render resolution (used for proper scaling)
+    shader->setUniform("resolution", sf::Vector2f(800, 600));
+
+    return std::make_optional<Geometry>(std::move(*logoTexture), std::move(*shader));
+}
+
+} // namespace
+
+
```
eXpl0it3r, line 376: I don't see how going from type-specific loading functions inside the type (`T::tryLoad()`)  to free functions with type names in the function name (`tryLoadT()`) is "simpler". If it would do away with the type itself, I could see it, but this way we uncouple what should naturally be coupled together as seen by the inclusion of the type name in the function name.

eXpl0it3r (changes requested):

examples/shader/Shader.cpp
vittorioromeo: Did you read the PR description?

Coupling the loading function with the type itself was unnecessary, as it makes sense for the type itself to be created directly with its required dependencies if those are available to the user.

Loading the dependencies is not a concern of the effect class itself, and embedding the loading function in the effect type reduces encapsulation (the usual member vs free function argument applies here).

Another way of thinking about it is that a user might want to copy paste one of these effect classes and use them in their own projects, but they might not need the loading function at all. They're unrelated concerns and should not be bundled together. It was my mistake to do so in the first PR, and now I'm trying to improve over my own mistake.

The calling function of the syntax is irrelevant here and mostly a red herring -- the load functions are not meant to be used generically as their signatures can differ and there's no useful overarching abstraction that loads a 'T' in this context.

eXpl0it3r: I still don't see anything wrong with using factory methods for the effects. We're not writing a generic effect library. If someone copy pastes the code and don't know how to adjust it, they may have bigger issues.

Yes, if I wrote some generic effects classes and have my separate resource handling classes, I wouldn't embed things either, but don't really see an issue doing it here, as it was done before your other changes.

> The calling function of the syntax is irrelevant here and mostly a red herring -- the load functions are not meant to be used generically as their signatures can differ and there's no useful overarching abstraction that loads a 'T' in this context.

It goes exactly at the core of my interjection. As long as we don't have a "generic" way of handling the loading (which we shouldn't write, this is an example after all) and thus we end up with n loading functions matching our n classes, there's no point in artificially separating the two and then still having to match them up by using matching variable names.

eXpl0it3r: Anyways, doesn't really matter, both works, you like it that way and I the other way. Unlikely you'd want to change it. 🤷

eXpl0it3r (approved):

eXpl0it3r:

examples/shader/Shader.cpp
vittorioromeo: > I still don't see anything wrong with using factory methods for the effects.

It's not "wrong", it's just unnecessary. 

I want SFML's codebase to be an example of good engineering practices. I don't want to suggest to our users that loading the dependencies of a class should be done inside that class, because it's a poor programming pattern that harms encapsulation and adds unnecessary coupling.

It's exactly the same principle described here in the Core Guidelines: [C.4: Make a function a member only if it needs direct access to the representation of a class](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rc-member)

> We're not writing a generic effect library. If someone copy pastes the code and doesn't know how to adjust it, they may have bigger issues.

Let's say they like the `Pixelate` shader and want to reuse it in their own project, but they already had a texture loaded, they just want to pixelate that existing one. 

Prior to this PR, having a `private` constructor for `Pixelate` and forcing the user to go through the `Pixelate::tryLoad` function would have prevented that use case from being possible.

It was my mistake -- there's no need for that constructor to be `private`, and there's no need to force the creation of the `Pixelate` effect through the factory function. They're both valid choices and they should both be accessible and decoupled:

1. If you already have a loaded texture and shader laying around, use the constructor.

2. If you need to load the texture and shader, use the factory function.

> It goes exactly at the core of my interjection. As long as we don't have a "generic" way of handling the loading (which we shouldn't write, this is an example after all) and thus we end up with n loading functions matching our n classes, there's no point in artificially separating the two and then still having to match them up by using matching variable names.

Having multiple functions that can be "almost" invoked generically through the same syntax suggests genericity when there's really none. Case in point, I might be tempted to refactor

```cpp
std::optional pixelateEffect   = Pixelate::tryLoad();
std::optional waveBlurEffect   = WaveBlur::tryLoad(font);
std::optional stormBlinkEffect = StormBlink::tryLoad();
std::optional edgeEffect       = Edge::tryLoad();
std::optional geometryEffect   = Geometry::tryLoad();
```

to something like:

```cpp
auto loadedEffects = tryLoadAll<Pixelate, WaveBlur, StormBlink, Edge, Geometry>();
```

Where `tryLoadAll` is:

```cpp
template <typename... Effects>
auto tryLoadAll()
{
    return std::tuple{Effects::tryLoad()...};
}
```

This is an attractive nuisance, because the functions all have the same syntax, or so it seems... Then we realize that one of them actually takes an extra `font`, and we might try to solve the issue with metaprogramming to detect when to pass the `font`, or pass the `font` in all interfaces even if it's unused, and the rabbit hole just goes deeper and deeper.

On the other hand, the new code...

```cpp
std::optional pixelateEffect   = tryLoadPixelate();
std::optional waveBlurEffect   = tryLoadWaveBlur(font);
std::optional stormBlinkEffect = tryLoadStormBlink();
std::optional edgeEffect       = tryLoadEdge();
std::optional geometryEffect   = tryLoadGeometry();
```

...does not lend itself to genericity, and that's a good thing, because all these functions are not meant to be handled generically or to conform to a `concept`.

One more counter-argument is that we might have multiple ways of loading the same shader. Maybe `Pixelate` could be loaded with different textures or shaders. Or maybe we can embed the shader in the source code.

There's no guaranteed 1-1 relationship between a class that needs dependencies and a function that loads its dependencies. Having a consistent `T::tryLoad` syntax suggests otherwise, and I think that's counterproductive.

Issue #3073, opened 2024-06-07
A deep copy occurs in the Drawable member draw function parameter RenderStates


Nrtyck:
### Prerequisite Checklist

- [X] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [X] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [X] I am here to report an issue and not to just ask a question or look for help (use the [forum](https://en.sfml-dev.org/forums/index.php#c3) or [Discord](https://discord.gg/nr4X7Fh) instead)

### Describe your issue here

![image](https://github.com/SFML/SFML/assets/37080373/aade6917-bad8-4e31-8e56-6cb832e50d16)
![image](https://github.com/SFML/SFML/assets/37080373/74b98e64-e860-4ae4-9b3c-f06c2181cc9c)

Whether to use `const RenderStates& states`, copy on demand when rendering?
 

### Your Environment

- OS / distro / window manager:  windows 11
- SFML version:  master 
- Compiler / toolchain: VS 2022
- Special compiler / CMake flags:

### Steps to reproduce

1. 
2. 

```cpp

```

### Expected behavior

![image](https://github.com/SFML/SFML/assets/37080373/66ef1e38-fee7-45ec-bb33-6706c96336da)
![image](https://github.com/SFML/SFML/assets/37080373/e030ea6b-ab8c-4cc1-8a42-f917664c33d5)

### Actual behavior

![image](https://github.com/SFML/SFML/assets/37080373/74b98e64-e860-4ae4-9b3c-f06c2181cc9c)

Bambo-Borris:
I'd want to see some performance data here before we went down this route again. The drawable API survived many years with this copy instead of a reference, and because of it a common pattern of use occurred. The temporary move away to a reference meant a long standing way of making custom drawables didn't work. In fact, anyone making a custom drawable that they wanted to texture or apply a shader to would have to deep copy anyway, so the reference argument was barely saving anything in that case.

vittorioromeo:
I still think it's worthwhile to look into this, the fact that we're needlessly copying the entirety of  `sf::RenderStates` even when not needed suggests that maybe there's an improvement to our API that we can make, even if the performance gain is not significant.

[...]

PR #3074, opened 2024-06-07
Pass `sf::RenderStates` by mutable reference

vittorioromeo:
See #3073 for context.

Basically, `sf::RenderTarget::draw` creates the first `sf::RenderStates`, then that same object is passed down the chain by mutable reference.

This prevents the unnecessary copies, allows mutability, and keeps the syntax simple for the end user.

As far as I understand, passing by reference here is the desired behavior -- if I have a `sf::Drawable` that internally stores some child drawable objects, I would expect those children to "inherit" the same render states as the parent.

### DIFF
--- a/examples/shader/Shader.cpp
+++ b/examples/shader/Shader.cpp
@@ -62,7 +62,7 @@ private:
         m_shader.setUniform("pixel_threshold", (x + y) / 30);
     }
 
-    void draw(sf::RenderTarget& target, sf::RenderStates states) const override
+    void draw(sf::RenderTarget& target, sf::RenderStates& states) const override
     {
         states.shader = &m_shader;
         target.draw(sf::Sprite{m_texture}, states);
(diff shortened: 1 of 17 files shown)


vittorioromeo:
@Nrtyck would this PR address your concerns?

eXpl0it3r:
I don't understand why we're spending time on this.

- SFML itself has more important issues to tackle
- SFML 3's focus isn't on improving the rendering API
- Passing around non-const references to custom types will most certainly lead to unexpected side effects

Passing objects by non-const references is maybe not directly a code smell, but it's certain a great source of potential errors.

I'd also argue, that if you check existing code bases that make use of SFML, you will find few if any places, where people duplicated this behavior. You don't generally want to have the modified states inside one drawable to affect other drawables.

The changes "work" if you only ever operate with the default render state. As soon as you construct your own render state or reuse one within a custom drawable, you'll run into many unwanted scenarios.

```
draw(parent) - default render state (RS)
+ Multiple transform
- draw(child1, states) - includes the transform of the parent
  + Multiple transform
  - draw(sprite, states) - includes the transform of the parent and child
- draw(child2, states) - includes the transform of the parent and child1
  - ????
```

vittorioromeo:
Clearly the consensus is to not pursue this idea :D

Issue #3077, opened 2024-06-07
`sf::Texture` move semantics implementation inconsistencies


vittorioromeo:
I noticed that `sf::Texture`'s move operations do not exchange `m_pixelsFlipped` nor `m_hasMipmap`, is this intended or an oversight?

One more inconsistency is that the move operations exchange `m_cacheId` with `0`, but `swap` does this:

```cpp
void Texture::swap(Texture& right) noexcept
{
    std::swap(m_size, right.m_size);
    // ... 
    m_cacheId       = TextureImpl::getUniqueId();
    right.m_cacheId = TextureImpl::getUniqueId();
}
```

Should `m_cacheId` be always reset to a unique id even as a consequence of a move?

vittorioromeo:
@binary1248: what is the reason that `swap` does not simply swap the cache IDs, instead generating two new ones? Are our move operations incorrect in setting the moved-from texture's cache ID to `0`? See #3077 for context.

vittorioromeo:
Some previous discussion here: https://github.com/SFML/SFML/pull/2479#pullrequestreview-1363542042

I find it hard to understand. If moving cache IDs works, then swapping should as well...

binary1248:
#3078 was merged, closing this.

PR #3078, opened 2024-06-07
Fix inconsistencies in `sf::Texture` move and swap

vittorioromeo:
See #3077. I am not sure about swapping the cache ID, but since we've been doing it in move operations and it worked, it should also work for `swap`. @binary1248 can confirm/deny.

### DIFF
--- a/src/SFML/Graphics/Texture.cpp
+++ b/src/SFML/Graphics/Texture.cpp
@@ -117,7 +117,9 @@ m_texture(std::exchange(right.m_texture, 0)),
 m_isSmooth(std::exchange(right.m_isSmooth, false)),
 m_sRgb(std::exchange(right.m_sRgb, false)),
 m_isRepeated(std::exchange(right.m_isRepeated, false)),
+m_pixelsFlipped(std::exchange(right.m_pixelsFlipped, false)),
 m_fboAttachment(std::exchange(right.m_fboAttachment, false)),
+m_hasMipmap(std::exchange(right.m_hasMipmap, false)),
 m_cacheId(std::exchange(right.m_cacheId, 0))
 {
 }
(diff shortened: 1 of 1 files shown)


ChrisThrasher (approved):
LGTM but I want Binary to approve it before we merge since he understands this better than I do.

binary1248:
`swap` was introduced as a refactoring in 6b71456a55c9cce3eae12ee9cdacbf96fe1c1a71. The assignment in `operator=` in the original implementation goes back to when caching was introduced in 191730ac0d639cfdef8fae560cdbc7d7e8c67db1. I didn't write the original code, but from what I can tell, it would have been possible to `swap` from the `temp` object there as well, so I guess swapping here should be safe? 😁

Generally speaking, if there is no technical reason to force `RenderTarget` to rebind the texture the next time it is used, new cache IDs don't have to be generated. Based on anecdotal evidence, some OpenGL implementations need textures to be rebound in order to trigger internal stuff which is why the cache ID is regenerated in special circumstances. However, if every data member is swapped without actually changing the contents of the texture itself, it should be relatively safe to swap the cache IDs as well. No guarantees though, be ready to revert if we get complaints. 😁

vittorioromeo:
> `swap` was introduced as a refactoring in [6b71456](https://github.com/SFML/SFML/commit/6b71456a55c9cce3eae12ee9cdacbf96fe1c1a71). The assignment in `operator=` in the original implementation goes back to when caching was introduced in [191730a](https://github.com/SFML/SFML/commit/191730ac0d639cfdef8fae560cdbc7d7e8c67db1). I didn't write the original code, but from what I can tell, it would have been possible to `swap` from the `temp` object there as well, so I guess swapping here should be safe? 😁
> 
> Generally speaking, if there is no technical reason to force `RenderTarget` to rebind the texture the next time it is used, new cache IDs don't have to be generated. Based on anecdotal evidence, some OpenGL implementations need textures to be rebound in order to trigger internal stuff which is why the cache ID is regenerated in special circumstances. However, if every data member is swapped without actually changing the contents of the texture itself, it should be relatively safe to swap the cache IDs as well. No guarantees though, be ready to revert if we get complaints. 😁

Sounds good to me, thanks for the explanation!

vittorioromeo:

PR #3082, opened 2024-06-08
Consistently print errors in factory functions

### LINKED ISSUE #3081: Ensure that all `std::optional` factory functions print errors to `sf::err()` in case of failure
### Prerequisite Checklist

As `std::optional` does not carry any additional error information with it, and as we decided to keep the API simple and not use a in-house version of `std::expected`, we should ensure that errors are propagated through the `sf::err()` stream.

There are a few places where errors are not being printed out:

- `sf::InputSoundFile::openFromFile`
- `sf::InputSoundFile::openFromMemory`
- `sf::InputSoundFile::openFromStream`
- `sf::Music::tryOpenFromFile`
- `sf::OutputSoundFile::openFromFile`
- `sf::SoundBuffer::loadFromFile`
- `sf::SoundBuffer::loadFromMemory`
- `sf::SoundBuffer::loadFromStream`
- `sf::SoundBuffer::initialize`
- `sf::Texture::loadFromFile`
- `sf::Texture::loadFromMemory`
- `sf::Texture::loadFromStream`
- `sf::Texture::loadFromImage`
- `sf::IpAddress::resolve` (arguable, it's not really loading a resource)
- `sf::IpAddress::getLocalAddress`
- `sf::IpAddress::getGlobalAddress`
- `sf::Cursor::loadFromPixels`
- `sf::Cursor::loadFromSystem`

### Use Cases

Giving users feedback on why resource loading failed.

### API Example

N/A

vittorioromeo:
Fixes #3081, see the issue for context.

### DIFF
--- a/src/SFML/Audio/InputSoundFile.cpp
+++ b/src/SFML/Audio/InputSoundFile.cpp
@@ -34,6 +34,7 @@
 #include <SFML/System/InputStream.hpp>
 #include <SFML/System/MemoryInputStream.hpp>
 #include <SFML/System/Time.hpp>
+#include <SFML/System/Utils.hpp>
 
 #include <ostream>
 #include <utility>
@@ -71,19 +72,32 @@ std::optional<InputSoundFile> InputSoundFile::openFromFile(const std::filesystem
     // Find a suitable reader for the file type
     auto reader = SoundFileFactory::createReaderFromFilename(filename);
     if (!reader)
+    {
+        // Error message generated in called function.
         return std::nullopt;
+    }
 
     // Wrap the file into a stream
     auto file = std::make_unique<FileInputStream>();
 
     // Open it
     if (!file->open(filename))
+    {
+        err() << "Failed to open input sound file from file (couldn't open file input stream)\n"
+              << formatDebugPathInfo(filename) << std::endl;
+
         return std::nullopt;
+    }
 
     // Pass the stream to the reader
     auto info = reader->open(*file);
     if (!info)
+    {
+        err() << "Failed to open input sound file from file (reader open failure)\n"
+              << formatDebugPathInfo(filename) << std::endl;
+
         return std::nullopt;
+    }
 
     return InputSoundFile(std::move(reader), std::move(file), info->sampleCount, info->sampleRate, std::move(info->channelMap));
 }
@@ -95,7 +109,10 @@ std::optional<InputSoundFile> InputSoundFile::openFromMemory(const void* data, s
     // Find a suitable reader for the file type
     auto reader = SoundFileFactory::createReaderFromMemory(data, sizeInBytes);
     if (!reader)
+    {
+        // Error message generated in called function.
         return std::nullopt;
+    }
 
     // Wrap the memory file into a stream
     auto memory = std::make_unique<MemoryInputStream>(data, sizeInBytes);
@@ -103,7 +120,10 @@ std::optional<InputSoundFile> InputSoundFile::openFromMemory(const void* data, s
     // Pass the stream to the reader
     auto info = reader->open(*memory);
     if (!info)
+    {
+        err() << "Failed to open input sound file from memory (reader open failure)" << std::endl;
         return std::nullopt;
+    }
 
     return InputSoundFile(std::move(reader), std::move(memory), info->sampleCount, info->sampleRate, std::move(info->channelMap));
 }
@@ -115,7 +135,10 @@ std::optional<InputSoundFile> InputSoundFile::openFromStream(InputStream& stream
     // Find a suitable reader for the file type
     auto reader = SoundFileFactory::createReaderFromStream(stream);
     if (!reader)
+    {
+        // Error message generated in called function.
         return std::nullopt;
+    }
 
     // Don't forget to reset the stream to its beginning before re-opening it
     if (stream.seek(0) != 0)
@@ -127,7 +150,10 @@ std::optional<InputSoundFile> InputSoundFile::openFromStream(InputStream& stream
     // Pass the stream to the reader
     auto info = reader->open(stream);
     if (!info)
+    {
+        err() << "Failed to open input sound file from stream (reader open failure)" << std::endl;
         return std::nullopt;
+    }
 
     return InputSoundFile(std::move(reader), {&stream, false}, info->sampleCount, info->sampleRate, std::move(info->channelMap));
 }
--- a/src/SFML/Audio/Music.cpp
+++ b/src/SFML/Audio/Music.cpp
@@ -79,10 +79,13 @@ Music& Music::operator=(Music&&) noexcept = default;
 
 
 ////////////////////////////////////////////////////////////
-std::optional<Music> Music::tryOpenFromFile(std::optional<InputSoundFile>&& optFile)
+std::optional<Music> Music::tryOpenFromInputSoundFile(std::optional<InputSoundFile>&& optFile, const char* errorContext)
 {
     if (!optFile.has_value())
+    {
+        err() << "Failed to open music from " << errorContext << std::endl;
         return std::nullopt;
+    }
 
     // TODO: apply RVO here via passkey idiom
     return Music(std::move(*optFile));
@@ -92,21 +95,21 @@ std::optional<Music> Music::tryOpenFromFile(std::optional<InputSoundFile>&& optF
 ////////////////////////////////////////////////////////////
 std::optional<Music> Music::openFromFile(const std::filesystem::path& filename)
 {
-    return tryOpenFromFile(InputSoundFile::openFromFile(filename));
+    return tryOpenFromInputSoundFile(InputSoundFile::openFromFile(filename), "file");
 }
 
 
 ////////////////////////////////////////////////////////////
 std::optional<Music> Music::openFromMemory(const void* data, std::size_t sizeInBytes)
 {
-    return tryOpenFromFile(InputSoundFile::openFromMemory(data, sizeInBytes));
+    return tryOpenFromInputSoundFile(InputSoundFile::openFromMemory(data, sizeInBytes), "memory");
 }
 
 
 ////////////////////////////////////////////////////////////
 std::optional<Music> Music::openFromStream(InputStream& stream)
 {
-    return tryOpenFromFile(InputSoundFile::openFromStream(stream));
+    return tryOpenFromInputSoundFile(InputSoundFile::openFromStream(stream), "stream");
 }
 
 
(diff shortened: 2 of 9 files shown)


vittorioromeo:

src/SFML/Audio/InputSoundFile.cpp
```diff
@@ -106,7 +123,10 @@ std::optional<InputSoundFile> InputSoundFile::openFromMemory(const void* data, s
     // Pass the stream to the reader
     auto info = reader->open(*memory);
     if (!info)
+    {
+        err() << "Failed to open input sound file from memory (reader open failure)\n" << std::endl;
```
eXpl0it3r, line 127: In other places (e.g. in the new audio impl.), we only specify the actual error that happens without prefacing in what context this happened, i.e. "Failed to open reader"

src/SFML/Audio/Music.cpp
```diff
@@ -79,10 +79,13 @@ Music& Music::operator=(Music&&) noexcept = default;
 
 
 ////////////////////////////////////////////////////////////
-std::optional<Music> Music::tryOpenFromFile(std::optional<InputSoundFile>&& optFile)
+std::optional<Music> Music::tryOpenFromInputSoundFile(std::optional<InputSoundFile>&& optFile, const char* errorContext)
```
eXpl0it3r, line 82: I'm against providing an errorContext parameter. Handle/lig the error at the level you have the information.

Also I don't think we should use char* as param

vittorioromeo: I think that both using an `errorContext` parameter and using `const char*` are the right choices in this case. Consider what the code would look like if we did not use an `errorContext` parameter:

```cpp
std::optional<Music> Music::tryOpenFromInputSoundFile(std::optional<InputSoundFile>&& optFile, const char* errorContext)
{
    if (!optFile.has_value())
        return std::nullopt;

    return Music(std::move(*optFile));
}


std::optional<Music> Music::openFromFile(const std::filesystem::path& filename)
{
    if (auto result = tryOpenFromInputSoundFile(InputSoundFile::openFromFile(filename)))
        return result;

    err() << "Failed to open music from file" << std::endl;
    return std::nullopt;
}


std::optional<Music> Music::openFromMemory(const void* data, std::size_t sizeInBytes)
{
    if (auto result = tryOpenFromInputSoundFile(InputSoundFile::openFromMemory(data, sizeInBytes)))
        return result;

    err() << "Failed to open music from memory" << std::endl;
    return std::nullopt;
}


std::optional<Music> Music::openFromStream(InputStream& stream)
{
    if (auto result = tryOpenFromInputSoundFile(InputSoundFile::openFromStream(stream)))
        return result;

    err() << "Failed to open music from stream" << std::endl;
    return std::nullopt;
}
```

Compare that with the currently proposed change:

```cpp
std::optional<Music> Music::tryOpenFromInputSoundFile(std::optional<InputSoundFile>&& optFile, const char* errorContext)
{
    if (!optFile.has_value())
    {
        err() << "Failed to open music from " << errorContext << std::endl;
        return std::nullopt;
    }

    return Music(std::move(*optFile));
}


std::optional<Music> Music::openFromFile(const std::filesystem::path& filename)
{
    return tryOpenFromInputSoundFile(InputSoundFile::openFromFile(filename), "file");
}


std::optional<Music> Music::openFromMemory(const void* data, std::size_t sizeInBytes)
{
    return tryOpenFromInputSoundFile(InputSoundFile::openFromMemory(data, sizeInBytes), "memory");
}


std::optional<Music> Music::openFromStream(InputStream& stream)
{
    return tryOpenFromInputSoundFile(InputSoundFile::openFromStream(stream), "stream");
}
```

Not using `errorContext` would be a clear violation of the DRY principle. There are situations where having a bit of repetition is better than having one more layer of abstraction, but this is not the case because:

- We already have a DRY helper function (`tryOpenFromInputSoundFile`)
- The helper function is `private` 
- The error printing mechanism does not capture file/line information, so producing the error in the helper function does not affect the accuracy of the error reporting

`const char*` here is also the simplest type that allows us to reduce the repetition without introducing any overhead, complexity, or having to include any additional header. Since the helper function is `private` and all of its uses are right below it, we can clearly see that only string literals are passed to `errorContext`, so the use of `const char*` is appropriate.

src/SFML/Audio/InputSoundFile.cpp
vittorioromeo: I think the additional context provided here can be useful in tracking down an eventual issue or bug. Can you point me to the place you're referring to? I've checked some Miniaudio-related `err()` calls and they all seem to provide quite a bit of context/information.

src/SFML/Audio/Music.cpp
eXpl0it3r: DRY shouldn't be blindly applied. We do have a difference in implementation, as such it's not a simple repetition, which again is indicated by the fact that we have to provide context information. If you need context information from multiple code locations, it's a strong indication that you should in fact "duplicate" the code.

In my experiencing passing around such error information is a bad design.

Anyways, feel free to just "Resolve conversation", if you don't want to adjust it. No need to drag it out into another bikeshedding "discussion".

src/SFML/Audio/InputSoundFile.cpp
eXpl0it3r: Maybe it was something else, can't remember where I had the discussion.
Looking at the other examples, it might be more in line to do:

```
Failed to open input sound file from memory: reader open failure
```

Also there shouldn't be both a `\n` and `std::endl`

src/SFML/Audio/InputSoundFile.cpp
```diff
@@ -130,7 +153,10 @@ std::optional<InputSoundFile> InputSoundFile::openFromStream(InputStream& stream
     // Pass the stream to the reader
     auto info = reader->open(stream);
     if (!info)
+    {
+        err() << "Failed to open input sound file from stream (reader open failure)\n" << std::endl;
```
eXpl0it3r, line 157:
```suggestion
        err() << "Failed to open input sound file from stream (reader open failure)" << std::endl;
```

src/SFML/Network/IpAddress.cpp
```diff
@@ -163,10 +176,13 @@ std::optional<IpAddress> IpAddress::getPublicAddress(Time timeout)
...
     const Http::Response page = server.sendRequest(request, timeout);
-    if (page.getStatus() == Http::Response::Status::Ok)
+
+    const Http::Response::Status status = page.getStatus();
+
+    if (status == Http::Response::Status::Ok)
         return IpAddress::resolve(page.getBody());
 
-    // Something failed: return an invalid address
+    err() << "Failed to retrieve public address (HTTP response status" << static_cast<int>(status) << ")" << std::endl;
```
eXpl0it3r, line 185: This might be rather confusing/leaking implementation. Why does my public IP resolving failed with an HTTP error code?

src/SFML/Network/IpAddress.cpp
```diff
@@ -81,6 +86,7 @@ std::optional<IpAddress> IpAddress::resolve(std::string_view address)
         return IpAddress(ntohl(ip));
     }
 
+    // Not generating en error message here to keep this function lightweight.
```
eXpl0it3r, line 89: Not sure if "lightweight" is a good argument for not logging errors.

Why is the context information in the situation not useful?

eXpl0it3r (changes requested):

src/SFML/Audio/InputSoundFile.cpp
vittorioromeo, line 127:
```suggestion
        err() << "Failed to open input sound file from memory (reader open failure)" << std::endl;
```

src/SFML/Network/IpAddress.cpp
vittorioromeo: We are already clearly documenting that `getPublicAddress` depends on an external server:

```cpp
    /// ...
    /// this function depends on the network connection and on a distant
    /// server
    /// ...
```

I think knowing why access to the server failed is useful in debugging unexpected `getPublicAddress` failures. I will tweak the message a bit to make it much clear.

src/SFML/Network/IpAddress.cpp
vittorioromeo: I agree that this comment is kinda bad. The reason why I don't want to include an error message here is because `resolve` returning an empty optional is not really a failure, it might be a totally fine scenario if we are trying to resolve user-entered addresses.

What I am trying to say is that this is one of the situations where "empty optional" does not mean "unexpected failure" but rather "possibly expected failure to parse".

eXpl0it3r (changes requested):

src/SFML/Network/IpAddress.cpp
```diff
@@ -163,10 +176,15 @@ std::optional<IpAddress> IpAddress::getPublicAddress(Time timeout)
...
     const Http::Response page = server.sendRequest(request, timeout);
-    if (page.getStatus() == Http::Response::Status::Ok)
+
+    const Http::Response::Status status = page.getStatus();
+
+    if (status == Http::Response::Status::Ok)
         return IpAddress::resolve(page.getBody());
 
-    // Something failed: return an invalid address
+    err() << "Failed to retrieve public address from external IP resolution server (HTTP response status"
```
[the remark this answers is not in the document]
vittorioromeo: Nice catch, fixed

src/SFML/Audio/InputSoundFile.cpp
vittorioromeo: > Looking at the other examples, it might be more in line to do:

Most of the examples I've found use parentheses to add additional information/context to an error message. I've grepped for `err() <<`. The language we use is not very consistent, but I'd rather deal with the consistency issue in a separate PR -- resolving for now.

vittorioromeo:

eXpl0it3r (approved):

ChrisThrasher:
This PR got me thinking that we should perhaps add tests for error messages. This is possible due to the fact that `sf::err` can be redirected. For starters it would be good to add error messages for all of our factory function APIs. Some tests attempt to exercise a particular error path so by testing the error messages we're also testing that the expected error path is taken.

Here's what that might look like. If we like this idea we can discuss it further in a separate issue. It will require a substantial amount of work to add this to all factory function tests so I won't go much farther than this without buy-in from the team.

https://github.com/SFML/SFML/tree/test_error_messages

eXpl0it3r:
I feel like this would make it unnecessarily complicated to change the code.

I believe I'd be fine with just checking that there is "something" in the error stream.

Issue #3079, opened 2024-06-07
`SoundEffects.cpp` example segfaults due to channel count mismatch


vittorioromeo:
### Prerequisite Checklist

- [X] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [X] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [X] I am here to report an issue and not to just ask a question or look for help (use the [forum](https://en.sfml-dev.org/forums/index.php#c3) or [Discord](https://discord.gg/nr4X7Fh) instead)

### Describe your issue here

The `SoundEffects.cpp` example segfaults on my machine, with invalid accesses at these lines. It seems like, in the first case, there is a mismatch in the size of `filters` (which is `music.getChannelCount()`) and the index `channel` (which goes up to `frameChannelCount`). 

In my machine, `music.getChannelCount() == 2` and `frameChannelCount == 8`.

```cpp
// L926
    outputFrames[channel] = *enabled ? filters[channel](input) : input;
//                                     ^~~~~~~~~~~~~~~~

// L770
    outputFrames[channel] = *enabled ? yn : xn;
//  ^~~~~~~~~~~~~~~~~~~~~
```

<details>
<summary>
Example stack trace
</summary>

```
0x00007ff652f16e3c in Reverb::ReverbFilter<float>::operator() (this=0x9f75450, input=0) at C:/OHWorkspace/SFML/examples/sound_effects/SoundEffects.cpp:1011
1011                auto output = static_cast<T>(0.7f * input + m_feedbackGain * m_buffer[m_cursor]);
(gdb) bt
#0  0x00007ff652f16e3c in Reverb::ReverbFilter<float>::operator() (this=0x9f75450, input=0) at C:/OHWorkspace/SFML/examples/sound_effects/SoundEffects.cpp:1011
#1  0x00007ff652f6fbfa in Reverb::Reverb()::{lambda(float const*, unsigned int&, float*, unsigned int&, unsigned int)#1}::operator()(float const*, unsigned int&, float*, unsigned int&, unsigned int) (this=0x9f17c30, inputFrames=0x9f100e0,
[...]

PR #3083, opened 2024-06-08
Improve/simplify `SoundEffects.cpp` example

vittorioromeo:
Same principle as #3062 and #3071, basically:

- Remove unnecessary layers of abstraction 
- Pass dependencies of effects explicitly rather than relying on global state or initialization in constructors
- Avoid unnecessary usage of `std::optional` for music resources
- Minor readability improvements

Note that I still get segfaults in this example, with or without this PR. See #3079 for context.


vittorioromeo:

vittorioromeo:

vittorioromeo:

PR #3084, opened 2024-06-08
Update self-referential owner ptr when moving `SoundStream`

vittorioromeo:
Missed this self-referential pointer when implementing move semantics for `sf::Music` in #3075. Handles the latest ASAN issue in #3079, but the first reported one still occurs.

### DIFF
--- a/src/SFML/Audio/SoundStream.cpp
+++ b/src/SFML/Audio/SoundStream.cpp
@@ -211,7 +211,7 @@ struct SoundStream::Impl : priv::MiniaudioUtils::SoundBase
     // Member data
     ////////////////////////////////////////////////////////////
     static constexpr ma_data_source_vtable vtable{read, seek, getFormat, getCursor, getLength, setLooping, /* flags */ 0};
-    SoundStream* const                     owner;        //!< Owning SoundStream object
+    SoundStream*                           owner;        //!< Owning SoundStream object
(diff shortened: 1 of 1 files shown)


vittorioromeo:

vittorioromeo:

vittorioromeo:

src/SFML/Audio/SoundStream.cpp
```diff
@@ -234,11 +234,26 @@ SoundStream::~SoundStream() = default;
 
 
 ////////////////////////////////////////////////////////////
-SoundStream::SoundStream(SoundStream&&) noexcept = default;
+SoundStream::SoundStream(SoundStream&& rhs) noexcept : m_impl(std::move(rhs.m_impl))
```
eXpl0it3r, line 237: Any reason we don't implement move semantics for the impl?

vittorioromeo: I'm not entirely sure what you mean here -- the impl is stored as a `std::unique_ptr`, so internally the type `Impl` is never moved nor copied.

eXpl0it3r: Ah, I missed that it was a unique_ptr

eXpl0it3r:

eXpl0it3r (approved):

vittorioromeo:

PR #3086, opened 2024-06-10
Fix optional comparison to integer value

ChrisThrasher:
## Description

Oversight from #2964. This code attempts to compare `-1` to a `std::optional<std::size_t>` which is not a comparison we meant to make. Checking against `-1` is checking against the old sentinel value.


eXpl0it3r (approved):

vittorioromeo:
Nice catch. It's really unfortunate that `std::optional<T>` implicitly converts to `T` in comparisons...

ChrisThrasher:
> It's really unfortunate that std::optional<T> implicitly converts to T in comparisons...

The real unfortunate part seems to be that this implicit conversion doesn't trigger `-Wsign-conversion`. Comparing `-1` (which is implicitly an `int`) to `std::size_t` is something we _attempt_ to disallow but these conversions in `std::optional` circumvent the safety measures we're taking.

Issue #3054, opened 2024-05-31
Windows GCC 13.2 compilation failed


Nathan-M-code:
### Prerequisite Checklist

- [X] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [X] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [X] I am here to report an issue and not to just ask a question or look for help (use the [forum](https://en.sfml-dev.org/forums/index.php#c3) or [Discord](https://discord.gg/nr4X7Fh) instead)

### Describe your issue here

Compilation failed during linking

### Your Environment

- OS / distro / window manager: Windows 64bits
- SFML version: 2.6.1
- Compiler / toolchain: gcc 13.2 seh posix urct
- Special compiler / CMake flags: Release

### Steps to reproduce

```ps
$repoUrl = "https://github.com/SFML/SFML"
$gitTag = "2.6.1"


# Clone the repository at the specified tag
git clone -b $gitTag $repoUrl

# Change the working directory to external_deps/SFML
cd SFML
md build
cd build
$SFML_build = $PWD

# Execute the cmake command
cmake -G "MinGW Makefiles" -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX="$installPath" ..

# Build the project
cmake --build . --target all

# Install the project
cmake --install .
```

### Expected behavior

Compilation success

### Actual behavior

[ 15%] Built target sfml-system
[ 17%] Built target sfml-main
[ 42%] Built target sfml-window
[ 53%] Built target sfml-network
[ 54%] Linking CXX shared library ..\..\..\lib\sfml-graphics-2.dll
C:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/13.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: .../external_deps/SFML/extlibs/libs-mingw/x64/libfreetype.a(sfnt.c.obj):sfnt.c:(.text+0x5614): undefined reference to `_setjmp'
[...]

PR #3087, opened 2024-06-10
libfreetype build from source

PrashanthKumar0:
## Description

previously there was no prebuild static libraries for clang UCRT
I believe instead of supplying prebuild static libs building from source would be
more cross compiler and cross platform. `freetype` library isn't that big so 
compile time shouldn't be any issue 

I resolved it by including freetype as git submodule.
using fetch content is bad as it wastes bandwidth if I clear CMake cache which is annoying.
so I included it as git submodule.

This PR is related to the issue #2821 #3054 

## Tasks

-   [x] Tested on Linux
-   [x] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

to test it build it (only Graphics) because this pull request only adds freetype as submodule

a minimal working example
build it using clang ucrt 

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        {
                window.close();
        }

        window.clear();
        window.display();
    }
}
```


vittorioromeo:
Thanks for the contribution! I'm generally in favour of this PR as it makes SFML easier to build on multiple platforms, and likely will make it easier in the future to support web platforms via Emscripten.

The only issue I see is that we never used `.gitmodules` in the past, and that there are some well-known annoyances with git submodules. 

I think that despite `FetchContent` being suboptimal in terms of bandwidth usage, it might be the best option to get consensus for this PR as we already have a precedent for using it to fetch `Catch2`.

Thoughts? @ChrisThrasher @binary1248 @eXpl0it3r

vittorioromeo:
Some discussion on the woes of submodules: https://diziet.dreamwidth.org/14666.html

eXpl0it3r:
[As mentioned](https://github.com/SFML/SFML/pull/2821#issuecomment-2154837531) already, git submodule won't be the solution we'll go with.

It's kind of funny though, that in [the past discussion](https://en.sfml-dev.org/forums/index.php?topic=25053.0) this was the most voted for solution. Personally I've not made great experience with submodules. It can work quite well on small projects where you know exactly what is needed, but even then, removing a submodule and switching between before and after is a nightmare.

A solution with FetchContent is already being worked on, but it's not trivial to cover our massive matrix of build type, config and OS: https://github.com/SFML/SFML/compare/master...feature/build_dependencies

vittorioromeo:
@eXpl0it3r: should this PR be closed in favour of @binary1248's upcoming work?

ChrisThrasher:
> @eXpl0it3r: should this PR be closed in favour of @binary1248's upcoming work?

Yes. This PR lacks many of the changes required for it begin to be ready to merge. For example, we need to make sure clang-tidy ignores these new sources files we're adding to the build. Analyzing 3rd party source code will lead to a myriad of false positives. We also have to figure out how adding new build targets will impact installation which is currently still an unanswered question. I appreciate the PR author's enthusiasm but this problem is a lot more complicated than they suspected.

PR #3085, opened 2024-06-10
Remove default empty state of `sf::MemoryInputStream`

ChrisThrasher:

PR #3088, opened 2024-06-10
Remove default empty state of `sf::FileInputStream`

ChrisThrasher:
## Description

Continuing in the tradition of

- #2992 
- #3020 
- #3021 
- #3022
- #3025
- #3028
- #3030
- #3032
- #3027
- #3075
- #3085

Was it necessary that I list out all of these PRs? No. But I thought it would be fun to include a really long list of PRs that this one refers to. Look how satisfying all these matching PR titles are.

### DIFF
--- a/include/SFML/System/FileInputStream.hpp
+++ b/include/SFML/System/FileInputStream.hpp
@@ -57,12 +57,6 @@ class SFML_SYSTEM_API FileInputStream : public InputStream
 {
 public:
     ////////////////////////////////////////////////////////////
-    /// \brief Default constructor
-    ///
-    ////////////////////////////////////////////////////////////
-    FileInputStream();
-
-    ////////////////////////////////////////////////////////////
     /// \brief Default destructor
     ///
     ////////////////////////////////////////////////////////////
@@ -97,10 +91,10 @@ public:
     ///
     /// \param filename Name of the file to open
     ///
-    /// \return True on success, false on error
+    /// \return File input stream on success, `std::nullopt` on error
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] bool open(const std::filesystem::path& filename);
+    [[nodiscard]] static std::optional<FileInputStream> open(const std::filesystem::path& filename);
 
     ////////////////////////////////////////////////////////////
     /// \brief Read data from the stream
@@ -144,12 +138,6 @@ public:
 
 private:
     ////////////////////////////////////////////////////////////
-    // Member data
-    ////////////////////////////////////////////////////////////
-#ifdef SFML_SYSTEM_ANDROID
-    std::unique_ptr<priv::ResourceStream> m_androidFile;
-#endif
-    ////////////////////////////////////////////////////////////
     /// \brief Deleter for stdio file stream that closes the file stream
     ///
     ////////////////////////////////////////////////////////////
@@ -157,6 +145,27 @@ private:
     {
         void operator()(std::FILE* file);
     };
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Construct from file
+    ///
+    ////////////////////////////////////////////////////////////
+    explicit FileInputStream(std::unique_ptr<std::FILE, FileCloser>&& file);
+
+#ifdef SFML_SYSTEM_ANDROID
+    ////////////////////////////////////////////////////////////
+    /// \brief Construct from resource stream
+    ///
+    ////////////////////////////////////////////////////////////
+    explicit FileInputStream(std::unique_ptr<priv::ResourceStream>&& androidFile);
+#endif
+
+    ////////////////////////////////////////////////////////////
+    // Member data
+    ////////////////////////////////////////////////////////////
+#ifdef SFML_SYSTEM_ANDROID
+    std::unique_ptr<priv::ResourceStream> m_androidFile;
+#endif
 
     std::unique_ptr<std::FILE, FileCloser> m_file; //!< stdio file stream
 };
@@ -188,9 +197,9 @@ private:
 /// \code
 /// void process(InputStream& stream);
 ///
-/// FileInputStream stream;
-/// if (stream.open("some_file.dat"))
-///    process(stream);
+/// std::optional stream = sf::FileInputStream::open("some_file.dat");
+/// if (stream)
+///    process(*stream);
 /// \endcode
 ///
 /// InputStream, MemoryInputStream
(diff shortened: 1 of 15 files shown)


include/SFML/System/FileInputStream.hpp
```diff
@@ -143,6 +137,12 @@ class SFML_SYSTEM_API FileInputStream : public InputStream
     std::optional<std::size_t> getSize() override;
 
 private:
+    ////////////////////////////////////////////////////////////
+    /// \brief Default constructor
+    ///
+    ////////////////////////////////////////////////////////////
+    FileInputStream();
```
ChrisThrasher, line 144: When constructing a `FileInputStream`, sometimes you construct from an `std::unique_ptr<priv::ResourceStream>` and sometimes from a `std::unique_ptr<std::FILE, FileCloser>`. I suppose we could add two constructors to handle both cases for now chose the lazy solution. Open to to changing if someone feels like have both of those constructors is the right call.

src/SFML/Audio/InputSoundFile.cpp
```diff
@@ -73,13 +73,14 @@ std::optional<InputSoundFile> InputSoundFile::openFromFile(const std::filesystem
...
-
-    // Open it
-    if (!file->open(filename))
+    // Open the file
+    auto fileInputStream = FileInputStream::open(filename);
+    if (!fileInputStream)
         return std::nullopt;
 
+    // Wrap the file into a stream
+    auto file = std::make_unique<FileInputStream>(std::move(*fileInputStream));
```
ChrisThrasher, line 82: This was a little awkward. We start with a fallible factory function, check if it succeeded, and then move that file input stream into the unique pointer. I'm not aware of a more graceful way to do this.

src/SFML/System/FileInputStream.cpp
```diff
@@ -84,13 +86,11 @@ std::optional<std::size_t> FileInputStream::read(void* data, std::size_t size)
 #ifdef SFML_SYSTEM_ANDROID
     if (priv::getActivityStatesPtr() != nullptr)
     {
-        if (!m_androidFile)
-            return std::nullopt;
+        assert(m_androidFile);
```
ChrisThrasher, line 89: I added a lot of assertions to this file. As far as I'm aware and as far as the tests seem to prove, these assertions never fire and thus they're not necessary. We should be able to remove them. It doesn't seem possible to construct a `FileInputStream` in such a way that we ever try to call a function on a null pointer.

Doesn't hurt much to include them but I'm also of the opinion that I'd rather remove assertions when we can statically prove they're always unnecessary.

src/SFML/System/FileInputStream.cpp
```diff
@@ -156,4 +150,8 @@ std::optional<std::size_t> FileInputStream::getSize()
     return size;
 }
 
+
+////////////////////////////////////////////////////////////
+FileInputStream::FileInputStream() = default;
```
ChrisThrasher, line 155: I moved this farther down in the file to ensure the order of declarations in FileInputStream.hpp matches the order of definitions in FileInputStream.cpp. This is not necessary but it helps when flipping back and forth between both files.

include/SFML/System/FileInputStream.hpp
vittorioromeo: Unfortunately I feel like having those constructors is the right call, I'd like to keep empty states unrepresentable even in our own implementation, and to be consistent with the great work we've done so far.

It seems like you should get away with:

```cpp
// FileInputStream.hpp

explicit FileInputStream(std::unique_ptr<std::FILE, FileCloser>&& file);

#ifdef SFML_SYSTEM_ANDROID
explicit FileInputStream(std::unique_ptr<priv::ResourceStream>&& androidFile);
#endif
```

```cpp
// FileInputStream.cpp

FileInputStream::FileInputStream(std::unique_ptr<std::FILE, FileCloser>&& file)
    : m_file(std::move(file))
{
}

#ifdef SFML_SYSTEM_ANDROID
FileInputStream::FileInputStream(std::unique_ptr<priv::ResourceStream>&& androidFile)
    : m_androidFile(std::move(androidFile))
{
}
#endif
```

Which I don't think it's too bad :)

src/SFML/Audio/InputSoundFile.cpp
vittorioromeo: Hmm, I think the way you did it is fine. 

I think the real issue here is that we're unnecessarily using a unique ptr, a `std::variant<FileInputStream, MemoryInputStream, InputStream*>` would suffice. Maybe I'll send a PR for that

src/SFML/System/FileInputStream.cpp
```diff
@@ -60,21 +57,26 @@ FileInputStream& FileInputStream::operator=(FileInputStream&&) noexcept = defaul
...
 #ifdef SFML_SYSTEM_WINDOWS
-    m_file.reset(_wfopen(filename.c_str(), L"rb"));
+    fileInputStream.m_file.reset(_wfopen(filename.c_str(), L"rb"));
 #else
-    m_file.reset(std::fopen(filename.c_str(), "rb"));
+    fileInputStream.m_file.reset(std::fopen(filename.c_str(), "rb"));
 #endif
-    return m_file != nullptr;
+    if (fileInputStream.m_file != nullptr)
+        return fileInputStream;
```
vittorioromeo, line 78: Same idea

```suggestion
    if (file != nullptr)
        return FileInputStream(std::move(file));
```

src/SFML/System/FileInputStream.cpp
vittorioromeo: I'm generally in favour of more assertions, even if they're "unnecessary" they help give local reasoning context when reading the code.

In this particular case they are extremely cheap at run-time as well, so I'd say keep them.

src/SFML/Audio/InputSoundFile.cpp
vittorioromeo: I briefly tried the idea above but it's not worth doing, the reader relies on the stream's address being stable.

vittorioromeo (approved):
LGTM

src/SFML/System/FileInputStream.cpp
kimci86: Those assertions can fire when with a moved-from FileInputStream.

vittorioromeo: @kimci86: probably it's the 3 hours of sleep, but I'm not seeing the problem. Are you saying these assertions can fire when a moved-from input stream is destroyed? They can definitely fire if a moved-from input stream is used, but it's not meant to be used.

kimci86: Sorry I should have been more explicit.
I am answering to this:
> these assertions never fire and thus they're not necessary
This is not true: the assertions can fire when using a moved-from instance. It is correct to assert because a moved-from instance should not be used. This is an argument in favor of having those asserts.

PR #3089, opened 2024-06-10
Simplify/optimize `sf::AudioResource` implementation

vittorioromeo:
I believe this incantation doesn't have the same issues encountered in #2974 (@binary1248 please confirm), as we still dynamically construct/destroy the `sf::AudioDevice` instance on demand at run-time.

However, this PR brings several advantages compared to the previous implementation:

- Considerably simpler, no use of IIFE lambda + `std::weak_ptr` combination, just a mutex + global reference count

- More efficient in terms of memory usage and runtime for every class that derives from `AudioResource`

- No compile-time dependency on `<memory>` anymore

- No dynamic allocation


vittorioromeo:

src/SFML/Audio/AudioResource.cpp
```diff
@@ -28,34 +28,45 @@
...
-        // Ensure we only ever create a single instance of an
-        // AudioDevice that is shared between all AudioResources
-        static std::mutex                       mutex;
-        static std::weak_ptr<priv::AudioDevice> weakAudioDevice;
+AudioResource::AudioResource()
+{
+    const std::lock_guard guard{deviceMutex};
+
+    if (deviceRC++ == 0u)
+        device = new sf::priv::AudioDevice;
```
eXpl0it3r, line 51: I'm very much surprised that you'd suggest a custom reference counting implementation in a C++17 code base, especially while actively replacing a shared/weak_ptr implementation. To what end? Save a std header include.

This goes too far for me. I don't want this in SFML, nor would I ever recommend to anyone to do this.

src/SFML/Audio/AudioResource.cpp
```diff
@@ -28,34 +28,45 @@
...
+
 namespace sf
 {
 ////////////////////////////////////////////////////////////
-AudioResource::AudioResource() :
-m_device(
-    []
-    {
-        // Ensure we only ever create a single instance of an
-        // AudioDevice that is shared between all AudioResources
```
eXpl0it3r, line 43: According to @binary1248, the lambda was deliberately chosen to avoid static init order fiasco.

Additionally, he mentioned that in order to support globals in user code the best we can, we avoid dumping stuff into anon namespaces and if something needs to be static, hide it behind a function

eXpl0it3r (changes requested):
This doesn't really solve anything, it's veery questionably simplified and optimize what exactly?

src/SFML/Audio/AudioResource.cpp
vittorioromeo: Thank you for your feedback. I'd like to clarify my rationale behind suggesting a custom reference counting implementation for this specific use case.

Our requirements are straightforward:

- Maintain a single `sf::priv::AudioDevice` global object for the entire program.

- Ensure the first `AudioResource` object initializes the audio device if none are active.

- De-initialize the audio device when the last `AudioResource` object is destroyed.

While `std::shared_ptr` and `std::weak_ptr` are powerful tools, they offer functionalities that exceed our needs, such as support for atomics, weak ownership, type erasure, and polymorphic conversions. 

Given the simplicity of our requirements, I proposed a custom implementation using a simple `unsigned int` counter and a `std::mutex`, which I believe is a more appropriate solution here. 

My goal is to use the simplest and most efficient tool for the job. If our requirements involved managing multiple audio devices simultaneously, then the added complexity of `std::shared_ptr` would indeed be justified.

I am open to further discussion and review to ensure the best possible solution for our project.

src/SFML/Audio/AudioResource.cpp
vittorioromeo: > According to @binary1248, the lambda was deliberately chosen to avoid static init order fiasco.

"Static initialization order fiasco" specifically refers to the fact that the order of initialiation of globals *across* translation units is not well-defined.

It does not apply in this case, because the globals inside the anonymous namespace are limited to the current translation unit. 

Global variables in a single translation unit  are initialized in the order in which they are defined.

Perhaps you were thinking of when @binary1248 mentioned that the use of the lambda *alongside* the `std::shared_ptr` and `std::weak_ptr` was a way of *avoiding* static initialization order fiasco compared to simply using a `sf::priv::AudioDevice` singleton as I tried in #2974?

---

> Additionally, he mentioned that in order to support globals in user code the best we can, we avoid dumping stuff into anon namespaces and if something needs to be static, hide it behind a function

I believe you're thinking of a different use case. We use globals in anonymous namespaces for implementation details quite a lot, and there's really nothing wrong with that nor any need to wrap them in functions. E.g.,

- https://github.com/SFML/SFML/blob/master/src/SFML/Window/Unix/KeyboardImpl.cpp#L48-L54

- https://github.com/SFML/SFML/blob/master/src/SFML/Window/Unix/WindowImplX11.cpp#L80-L91

- https://github.com/SFML/SFML/blob/master/src/SFML/Window/Unix/VulkanImplX11.cpp#L102-L103

...and many more.

vittorioromeo:

src/SFML/Audio/AudioResource.cpp
eXpl0it3r: I'm totally okay with this perhaps slightly overpowered use of `std::shared_ptr` and `std::weak_ptr`. @ChrisThrasher has spent a considerable amount of time replacing all the manual memory managements with smart pointers, I'm not about to add our custom MMM just to avoid using a stdlib construct.

Maybe functionality-wise this is "simpler", but it's certainly not simpler implementation wise. Now you have to again ensure that `new` and `delete` are always guaranteed and everything that comes with MMM.

vittorioromeo: > I'm totally okay with this perhaps slightly overpowered use of `std::shared_ptr` and `std::weak_ptr`. @ChrisThrasher has spent a considerable amount of time replacing all the manual memory managements with smart pointers, I'm not about to add our custom MMM just to avoid using a stdlib construct.
> 
> Maybe functionality-wise this is "simpler", but it's certainly not simpler implementation wise. Now you have to again ensure that `new` and `delete` are always guaranteed and everything that comes with MMM.

You raise a great point here -- there's not even any need for dynamic allocation at all in this case. Updated the PR.

src/SFML/Audio/AudioResource.cpp
eXpl0it3r: I hope @binary1248 can provide some more insights, as I'm mostly just paraphrasing what we've chatted about this.

A call of a function that accesses these variables before they're initialized will also cause issues.

You also need to consider that all of this has to hold true for the whole list of all the possible initializations and OS specific implementations.

I honestly trust @binary1248's experience on this topic more than what the standard or intuition reveals. Yes, maybe it's not always absolutely necessary, but it helps keep it simple, when following the same pattern everywhere and not having to reevaluate all the possible initialization states.

image

> * https://github.com/SFML/SFML/blob/master/src/SFML/Window/Unix/KeyboardImpl.cpp#L48-L54
> * https://github.com/SFML/SFML/blob/master/src/SFML/Window/Unix/WindowImplX11.cpp#L80-L91
> * https://github.com/SFML/SFML/blob/master/src/SFML/Window/Unix/VulkanImplX11.cpp#L102-L103
> 
> ...and many more.

All places we should change to this pattern.

vittorioromeo: I am happy to hear @binary1248's thoughts on the matter, but I disagree with changing/rejecting `static` TU-specific global objects without a good reason to do so. 

The two approaches are very different:

- `static` function-scope variables are initialized the first time control flow reaches them, and their initialization is guaranteed to be synchronized (thread-safe, but also slower and often unnecessarily so)

- `static` global-scope variables are defined to be initialized before `main()`, but primitive types are usually initialized at compile-time via constant initialization -- there's also no thread safety added to to the initialization

In a situation like this one, where the global variable is an implementation detail that's only accessed from the TU it's defined in and already protected by a mutex, there's little reason to use a `static` function-scope variable unless I am missing something. 

Looking for @binary1248's confirmation, though, I can always be wrong...

vittorioromeo:

vittorioromeo:

src/SFML/Audio/AudioResource.cpp
binary1248: https://godbolt.org/z/cxEb8cM54

Even if symbols don't have external linkage, any call from outside of the TU to a function inside the TU that makes use of those internal symbols will be subject to initialization order issues.

It is true that this wouldn't be an issue if all variables we dump into the anonymous namespace are able to be constant initialized, but who is going to guarantee this? We already saw in #3076 that compiler vendors have no issue building in backdoors that subtly change behaviour that such a change would rely on. Having to build in additional (in my opinion unnecessary) assumptions to guarantee that our code works decreases the maintainability of the codebase and probably will make the library less attractive for less experienced users to contribute to.

We have to be careful not to venture down the path of doing things just for the sake of it as long as it is supported by some obscure paragraph in the language specification. As Bjarne Stroustrup has mentioned on several occasions, "Be clever, but don't be too clever."

The only concrete benefit of this change is that it reduces the memory overhead of audio objects. However, instead of looking at the reduction in absolute terms, we should look at it in relative terms. Audio objects are already pretty large (they also indirectly allocate memory through miniaudio). Couple that with the fact that almost all the time you will have some data buffer storing the audio data and all of a sudden the savings of this change will slip under 0.1% of the actual memory requirement of the complete audio object, maybe even 0.01% or 0.001% for larger audio objects. In today's environment and especially in the context of typical SFML applications, I don't think audio object memory requirements is going to be anyone's top concern, or even a concern at all. It is far easier for the user to waste memory by writing suboptimal code in their own project than by having many audio objects lying around.

As I said above, we should focus more on keeping the codebase maintainable and extensible, even for contributors that are less experienced than ourselves. We shouldn't rely on language specification minutiae unless there is absolutely no other way to provide users with features and functionality they actually care about, and at this time audio memory requirements isn't one of them.

vittorioromeo: @binary1248: thanks for sharing your thoughts and for the example!

The code you posted is a clear and good example of static order initialization fiasco -- there are *two* TUs having global variables with internal linkage (namely, `foo.cpp::(anon)::value` and `bar.cpp::(anon)::value). 

The issue with that code is that the initialization order between those two `value` variables is unspecified (as they are defined in different TUs), and that the two TUs end up calling each other.

I don't see how the example represents what's happening in this PR or the other examples in SFML that I posted, as there is only *one* TU having global variables with internal linkage, and those variables can only be accessed from that TU specifically.

Every access to either `deviceMutex`, `optDevice`, or `deviceRC` is guaranteed to happen in `AudioResource.cpp`, and that file does not access any other global variable.

Therefore, I think that this PR is safe as is. **Am I correct, or am I missing something very subtle here?**

Regardless, this PR is still completely valid even if you want to use function-scope `static` variables, that do not suffer from the issue you mentioned. I think it's unnecessary, but if it makes you feel safer, I can adapt the PR to look like this:

<details>

<summary>
Same PR, but with function-scope `static`
</summary>

```cpp
namespace
{
struct DeviceAccess
{
    std::mutex                           deviceMutex;
    std::optional<sf::priv::AudioDevice> optDevice;
    unsigned int                         deviceRC{};
};

DeviceAccess& deviceAccess()
{
    static DeviceAccess deviceAccess;
    return deviceAccess;
}
} // namespace

namespace sf
{
AudioResource::AudioResource()
{
    const auto& [deviceMutex, optDevice, deviceRC] = deviceAccess();
    const std::lock_guard guard{deviceMutex};

    if (deviceRC++ == 0u)
        optDevice.emplace();
}

AudioResource::~AudioResource()
{
    const auto& [deviceMutex, optDevice, deviceRC] = deviceAccess();
    const std::lock_guard guard{deviceMutex};

    if (--deviceRC == 0u)
        optDevice.reset();
}

AudioResource::AudioResource(const AudioResource&) : AudioResource{}
{
}

AudioResource::AudioResource(AudioResource&&) noexcept : AudioResource{}
{
}
```

</details>

This would satisfy my original goals of (1) simplifying the code, (2) avoiding dynamic allocation, and (3) reducing memory and run-time footprint of `AudioResource`, while being 100% guaranteed of not being subject to static order initialization fiasco.

**Would you consider accepting the PR in that alternative form shown above?**

binary1248: The problem is that we don't have control over the way the user writes their code. Using the example I provided, you can imagine `foo.cpp` being in our codebase and `bar.cpp` being in the user's codebase. If they also have objects that have to be initialized in namespace scope that make calls into SFML it will be the equivalent to the behaviour in the example.

We can't assume we are the only ones using objects in namespace scope. Depending on the complexity of the relationship, it might or might not work for the user. Sometimes it will come down to sheer luck on the side of the user whether the TUs are linked in a specific order or not.

@eXpl0it3r and I would not bring up such issues if they had not originated from actual reports from users constructing their objects in namespace scope over the years. Our usual answer was just "not to do that" since it was not officially supported by SFML, but now that we have an implementation that guarantees support for such use cases it would be a functionality regression if we had to claim that they are again not officially supported because of this change.

Regarding the change you suggested, yes that would resolve this issue. Whether the net change in code compared to the current implementation warrants it being made I leave up to @eXpl0it3r to decide.

vittorioromeo: @binary1248: I've posted on StackOverflow (https://stackoverflow.com/questions/78643699) and I've done some research, and these are the conclusions I've gathered from those two things and from your feedback.

---

Basically, the reason why the original version of this PR using a file-scope `static` global wasn't 100% safe was due to the possibility of users creating file-scope `static` global `AudioResource` objects. 

Therefore, this code below could cause undefined behavior:

```cpp
//
//
// AudioResource.cpp

namespace { /* ...file-scope statics here... */ }
AudioResource::AudioResource() { /* ...use file-scope statics... */ }
// ...


//
//
// SomeUserTU.cpp

static AudioResource globalAudioResource;
    // ISSUE: at this point in time, `AudioResource()` could
    // be invoked before `AudioResource.cpp`'s file-scope statics
   // are initialized!
```

An easy solution to this mess would be not supporting global `AudioResource` objects, but as you mentioned this is functionality we already provide and we don't want to revert that.

---

The suggestion of using a function-local `static` *seems* to work, but it is also subtly flawed! Consider the following code:

```cpp
//
//
// AudioResource.cpp

namespace { Globals& getGlobals() { /* ...function-scope statics... */ } }
AudioResource::AudioResource() { /* ...use `getGlobals()`... */ }
// ...


//
//
// SomeUserTU.cpp

static AudioResource globalAudioResource;
    // ISSUE: static *destruction* order fiasco!
    // 1. `AudioResource::AudioResource()` is called
    // 2. `<anon>::getGlobals()` is called
    // 3. `Globals::Globals()` is called
    // 4. `std::mutex::mutex()` is called
    // 5. ...all good so far...
    // 6. ...program termination starts...
    // 7. `Globals::~Globals()` is called
    // 8. `std::mutex::~mutex()` is called
    // 9. `AudioResource::~AudioResource()` is called
    // 10. WHOOPS! Undefined behavior, we're using destroyed globals!
```

---

The correct way of supporting global `AudioResource` objects is to leak the global variables as I've done in the latest iteration of this PR. 

I've used a file-scope `static` buffer to avoid dynamic allocation altogether, but the principle is the same: do not let the globals get destroyed at all to prevent the destruction order fiasco.

---

I understand that the PR is a bit of a tougher sell now, but I still think it's a good change, as:

- All forms of dynamic allocation are avoided
- `AudioResource`'s ownership model is much simpler (no shared or weak pointers)
- `AudioResource`'s memory footprint is zero
- `AudioResource`'s run-time overhead is diminished (shared/weak pointer overhead is larger due to control block, jumping around in memory, etc)
- No compile-time dependency on `<memory>`

binary1248: Why do you assume that the global destruction order is not the opposite of construction order? When function static objects are fully constructed, an entry is inserted into a destructor list that is called by the runtime when it is shutting down the application. Naturally this list is gone through in reverse order so that the destructors are called in the opposite order in which they were constructed to mimic objects with automatic storage duration and the general ordering of global objects within a TU.

Also don't forget that only fully constructed objects will eventually be destroyed. If object A depends on object B first being fully constructed then object A will be fully constructed after object B and thus the destructor for object A will eventually be called before that of object B. This is exactly the same behaviour for inheritance hierarchies and object composition... I don't really see how this can be an issue.

vittorioromeo: > Why do you assume that the global destruction order is not the opposite of construction order? 

I am assuming that, and that's the issue. Unless I am missing something, below is my understanding.

If a global `AudioResource` gets initialized first, its constructor will initialize the globals.

So the construction order list will be: (1) `AudioResource`, (2) globals.

On program termination, the opposite order will be taken.

So the globals are destroyed first, then `~AudioResource()` is called. The destructor of `AudioResource` will at that point access the destroyed globals, resulting in UB.

binary1248: > So the construction order list will be: (1) AudioResource, (2) globals.

This is not the case, the globals will have to be fully constructed before AudioResource can be fully constructed so the destructors of the globals will also be called after those of AudioResource.

Like I said in my previous comment, it is not the order of _calling_ the constructors that matters but the order of their _completion_. The AudioResource constructor cannot complete before those of the globals thus the AudioResource constructor will be inserted into the destruction list after those of the globals and since destruction takes place in reverse order the AudioResource destructor will be called before those of the globals.

See:
- https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables
- https://en.cppreference.com/w/cpp/utility/program/exit

Everything revolves around the _completion_ of either construction/initialization or destruction, never based on the order in which they are called.

vittorioromeo: You are correct and I was wrong. I will simplify this ASAP.

vittorioromeo: Done, resolving. Thanks!

src/SFML/Audio/AudioResource.cpp
vittorioromeo: @eXpl0it3r: After talking to binary, this PR is now ready for your review again.

eXpl0it3r:

eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:

PR #3092, opened 2024-06-10
Optimize `std::optional` creation, enable more [N]RVO opportunities

vittorioromeo:
*(Supersedes #3068 and #3069, because they were getting stale, and some changes were conflicting between the two PRs.)*

The goal of this PR is to improve the performance of factory functions, primarily resource-loading ones based on `std::optional`, but also a few other ones of various nature.

---

To improve `std::optional` creation, I've changed our usual pattern. Basically, the pattern we frequently use...

```cpp
class Widget
{
private:
    explicit Widget(Dependency&&);

public:
    static std::optional<Widget> make(std::filesystem::path p)
    {
        std::optional<Dependency> d = loadDependencyFromPath(p);
        if (!d.has_value()) { return std::nullopt; }
        
        return Widget{std::move(*d)};
    }
};
```

...has a flaw -- it is not eligible for C++17 guaranteed copy elision (RVO) due to the fact that we're not returning a prvalue of type `std::optional<Widget>`, but a prvalue of type `Widget` instead, whose type does *not* match the function's return type.

That has a few practical drawbacks:

- Non-copyable/non-movable types cannot be returned in this way;
- Performance is worsened, as extra move/copy operations will be invoked;
- Code that previously relied on address stability might break after switching to this pattern due to the extra move/copy operations (see #3062).

The way of improving the pattern is to return a *prvalue* of `std::optional<Widget>` type, thus gaining eligibility for guaranteed copy elision. E.g.,

```cpp
static std::optional<Widget> make(std::filesystem::path p)
{
    std::optional<Dependency> d = loadDependencyFromPath(p);
    if (!d.has_value()) { return std::nullopt; }
    
    return std::make_optional<Widget>(std::move(*d));
}
```

This works, but only if the constructor of `Widget` is `public`, as internally `std::make_optional` tries to construct `Widget` outside of `Widget` itself's scope. The same issue applies to `std::optional<T>{std::in_place, ...}`.

Therefore, I added a new `sf::priv::PassKey<T>` helper class to give access to the `private` constructors to `std::optional`'s internals. It adds a bit of complexity, and requires us to mark those constructors `public:` and annotate them with `\private` for Doxygen, but I strongly believe it's worth doing as all the heavyweight types such as `Texture`, `Font`, et cetera, all have `private` constructors and use the inefficient factory pattern that does not support guaranteed copy elision. 

See also: https://stackoverflow.com/questions/78581072/stdoptional-factory-function-with-guaranteed-copy-elision-and-private-cons

---

To enable more NRVO opportunities, I was guided by the newly added `-Wnrvo` GCC warning, that produces diagnostics like these:

```
In file included from C:/OHWorkspace/SFML/build/src/SFML/Audio/CMakeFiles/sfml-audio.dir/Unity/unity_0_cxx.cxx:7:
C:/OHWorkspace/SFML/src/SFML/Audio/AudioDevice.cpp: In static member function 'static std::vector<sf::priv::AudioDevice::DeviceEntry> sf::priv::AudioDevice::getAvailableDevices()':
C:/OHWorkspace/SFML/src/SFML/Audio/AudioDevice.cpp:275:12: warning: not eliding copy on return in 'static std::vector<sf::priv::AudioDevice::DeviceEntry> sf::priv::AudioDevice::getAvailableDevices()' [-Wnrvo]
  275 |     return deviceList;
      |            ^~~~~~~~~~
```

Note that I did not address the one above, because it was not worth it and it would have complicated the control flow too much.

The new pattern is as follows:

```cpp
std::optional<SoundBuffer> SoundBuffer::loadFromSamplesImpl(
    std::vector<std::int16_t>&&      samples,
    unsigned int                     channelCount,
    unsigned int                     sampleRate,
    const std::vector<SoundChannel>& channelMap)
{
    std::optional<SoundBuffer> soundBuffer; // Use a single local variable for NRVO

    if (channelCount == 0 || sampleRate == 0 || channelMap.empty())
    {
        err() << "...";
        return soundBuffer; // Same as `std::nullopt`, but NRVO-friendly
    }

    // Take ownership of the audio samples
    soundBuffer.emplace(priv::PassKey<SoundBuffer>{}, std::move(samples));

    // ...

    return soundBuffer; // Eligible for NRVO
}

### DIFF
--- a/include/SFML/Audio/InputSoundFile.hpp
+++ b/include/SFML/Audio/InputSoundFile.hpp
@@ -31,6 +31,8 @@
 
 #include <SFML/Audio/SoundFileReader.hpp>
 
+#include <SFML/System/PassKey.hpp>
+
 #include <filesystem>
 #include <memory>
 #include <optional>
@@ -237,16 +239,21 @@ private:
         bool owned{true};
     };
 
+public:
     ////////////////////////////////////////////////////////////
+    /// \private
+    ///
     /// \brief Constructor from reader, stream, and attributes
     ///
     ////////////////////////////////////////////////////////////
-    InputSoundFile(std::unique_ptr<SoundFileReader>&&            reader,
+    InputSoundFile(priv::PassKey<InputSoundFile>&&,
+                   std::unique_ptr<SoundFileReader>&&            reader,
                    std::unique_ptr<InputStream, StreamDeleter>&& stream,
                    std::uint64_t                                 sampleCount,
                    unsigned int                                  sampleRate,
                    std::vector<SoundChannel>&&                   channelMap);
 
+private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
--- a/include/SFML/Audio/Music.hpp
+++ b/include/SFML/Audio/Music.hpp
@@ -31,6 +31,8 @@
 
 #include <SFML/Audio/SoundStream.hpp>
 
+#include <SFML/System/PassKey.hpp>
+
 #include <filesystem>
 #include <memory>
 #include <optional>
@@ -238,12 +240,16 @@ private:
     [[nodiscard]] static std::optional<Music> tryOpenFromInputSoundFile(std::optional<InputSoundFile>&& optFile,
                                                                         const char*                     errorContext);
 
+public:
     ////////////////////////////////////////////////////////////
+    /// \private
+    ///
     /// \brief Initialize the internal state after loading a new music
     ///
     ////////////////////////////////////////////////////////////
-    explicit Music(InputSoundFile&& file);
+    explicit Music(priv::PassKey<Music>&&, InputSoundFile&& file);
 
+private:
     ////////////////////////////////////////////////////////////
     /// \brief Helper to convert an sf::Time to a sample position
     ///
--- a/include/SFML/Audio/OutputSoundFile.hpp
+++ b/include/SFML/Audio/OutputSoundFile.hpp
@@ -32,6 +32,8 @@
 #include <SFML/Audio/SoundChannel.hpp>
 #include <SFML/Audio/SoundFileWriter.hpp>
 
+#include <SFML/System/PassKey.hpp>
+
 #include <filesystem>
 #include <memory>
 #include <optional>
@@ -83,13 +85,15 @@ public:
     ////////////////////////////////////////////////////////////
     void close();
 
-private:
     ////////////////////////////////////////////////////////////
+    /// \private
+    ///
     /// \brief Constructor from writer
     ///
     ////////////////////////////////////////////////////////////
-    explicit OutputSoundFile(std::unique_ptr<SoundFileWriter>&& writer);
+    explicit OutputSoundFile(priv::PassKey<OutputSoundFile>&&, std::unique_ptr<SoundFileWriter>&& writer);
 
+private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
--- a/include/SFML/Audio/SoundBuffer.hpp
+++ b/include/SFML/Audio/SoundBuffer.hpp
@@ -31,6 +31,7 @@
 
 #include <SFML/Audio/SoundChannel.hpp>
 
+#include <SFML/System/PassKey.hpp>
 #include <SFML/System/Time.hpp>
 
 #include <filesystem>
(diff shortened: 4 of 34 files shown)

### CODE REVIEW

**Comment by eXpl0it3r** [comment]

**Comment by vittorioromeo** [comment]

**Comment by vittorioromeo** [comment]

**Comment by eXpl0it3r** [comment]
I can't accept such a solution.

- It spills `sf::priv` into the public API
- It introduces a hard to understand concept of PassKey (what is even a PassKey?)
- It breaks API protection level to some degree

**Comment by ChrisThrasher** [comment]
I'm unsure if the performance gains warrant the additional complexity and the fact that we're exposing functions as `public:` that we don't actually want anyone to use. I'm leaning towards a "no" on this PR.

**Comment by vittorioromeo** [comment]
@eXpl0it3r, @ChrisThrasher: thanks for considering the change, I believe it is worth merging as our users are now *required* to go through our `std::optional`-based factories and without these changes they'd incur a performance penalty compared to SFML 2.x due to additional avoidable constructor calls. 

I also mentioned before how I want SFML 3.x to be an example of an idiomatic and performant C++17 library, and this pattern is the proper way of constructing `std::optional` instances without unnecessary overhead.

I will attempt once to provide some counterpoints to have you reconsider your decision, if it doesn't change then I will close this PR due to lack of consensus. Regardless, I will open an issue to choose what our definition of "consensus" is and how many SFML team members are required to agree/disagree to a proposed change in order to reach that "consensus".

Counterpoints:

- > It spills `sf::priv` into the public API

  - `sf::priv` is already technically part of the public API, every symbol defined in that namespace is already accessible by our users. The point of using the `priv` namespace is to inform users that any use of entities defined inside `priv` is not officially supported, but there's nothing that stops them from doing that.
  
    On the other hand, this PR does not suffer from that issue because the functions that take a `sf::priv::PassKey` are not invocable by the user at all. User code attempting to invoke those functions will not compile.
    
- > It introduces a hard to understand concept of PassKey (what is even a PassKey?)

  - The PassKey idiom (a.k.a. Attoney/Agent idiom) is a widely used and known idiom in the C++ ecosystem. You can easily find mentions of this idiom [as early as 2006](https://www.drdobbs.com/friendship-and-the-attorney-client-idiom/184402053), or [on StackOverflow questions from 13 years ago](https://stackoverflow.com/questions/3217390/clean-c-granular-friend-equivalent-answer-attorney-client-idiom). I've seen this idiom used widely in production in my professional experience and I've seen it multiple times in the wild in open-source libraries. A GitHub search [reveals its widespread use](https://github.com/search?q=%22passkey%3C%22+language%3AC%2B%2B&type=code). 
  
  I also do not think the concept is hard to understand. Could you please elaborate on what parts of the idiom you find hard to understand?
  
- > It breaks API protection level to some degree

  - I don't see how, as the functions accepting a `sf::priv::PassKey` are not invocable by the user, and do not appear in our generated documentation. Could you elaborate on what practical risks you believe might arise?
  
- > we're exposing functions as public: that we don't actually want anyone to use

  - These functions cannot be invoked by our users and will not appear in our generated documentation.

**Comment by vittorioromeo** [comment]
> Regardless, I will open an issue to choose what our definition of "consensus" is and how many SFML team members are required to agree/disagree to a proposed change in order to reach that "consensus".

I created a forum thread as I thought that'd be more appropriate compared to a GitHub issue:
https://en.sfml-dev.org/forums/index.php?topic=29571.new#new

**Comment by vittorioromeo** [comment]
> > This works, but only if the constructor of Widget is public, as internally std::make_optional tries to construct Widget outside of Widget itself's scope. The same issue applies to std::optional{std::in_place, ...}.
> 
> In our case all of our types using factory functions do support being moved. Is it reasonable to use `std::make_optional` and use the public move constructor to move the type into the optional? All of our move operations are public and thus eligible for use by `make_optional`. Does that negative the performance gains of whatever addition elisions we get?

The elision would avoid the move completely, and directly construct the optional in the caller's stack frame. Basically, you would avoid an invocation of `T(T&&)` and `std::optional<T>(std::optional<T>&&)`.

Here is a demo showing that the impact can be huge, and scales with the amount of data members that a type has: https://gcc.godbolt.org/z/Ene11nWEv

```text
WITHOUT ELISION:
Dependency()
Dependency(Dependency&&)
~Dependency()
Dependency(Dependency&&)
Widget(Dependency&&)
Dependency(Dependency&&)
Widget(Widget&&)
~Widget()
~Dependency()
~Dependency()
~Widget()
~Dependency()


WITH ELISION:
Dependency()
Dependency(Dependency&&)
Widget(Dependency&&)
~Dependency()
~Widget()
~Dependency()


WITH NRVO:
Dependency()
Dependency(Dependency&&)
Widget(Dependency&&)
~Dependency()
~Widget()
~Dependency()
```

PR #3094, opened 2024-06-11
Add `timeout` parameter to `waitEvent`

vittorioromeo:
See https://github.com/SFML/SFML/pull/2992/files?w=1#r1634861828 and #1184 for context.

### DIFF
--- a/include/SFML/Window/WindowBase.hpp
+++ b/include/SFML/Window/WindowBase.hpp
@@ -33,6 +33,7 @@
 #include <SFML/Window/WindowEnums.hpp>
 #include <SFML/Window/WindowHandle.hpp>
 
+#include <SFML/System/Time.hpp>
 #include <SFML/System/Vector2.hpp>
 
 #include <memory>
@@ -202,11 +203,12 @@ public:
     /// \brief Wait for an event and return it
     ///
     /// This function is blocking: if there's no pending event then
-    /// it will wait until an event is received. After this function
-    /// returns if no error occurred, the returned event will not be
-    /// empty. This function is typically used when you have a thread
-    /// that is dedicated to events handling: you want to make this
-    /// thread sleep as long as no new event is received.
+    /// it will wait until an event is received or until the provided
+    /// timeout elapses. After this function returns if no error nor
+    /// timeout occurred, the returned event will not be empty.
+    /// This function is typically used when you have a thread that is
+    /// dedicated to events handling: you want to make this thread sleep
+    /// as long as no new event is received.
     /// \code
     /// if (const auto event = window.waitEvent())
     /// {
@@ -214,12 +216,14 @@ public:
     /// }
     /// \endcode
     ///
-    /// \return The event
+    /// \param timeout Maximum time to wait (`Time::Zero` for infinite)
+    ///
+    /// \return The event; will be `Empty` (convertible to `false`) on timeout or if window was closed
     ///
     /// \see pollEvent
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] Event waitEvent();
+    [[nodiscard]] Event waitEvent(Time timeout = Time::Zero);
 
     ////////////////////////////////////////////////////////////
     /// \brief Get the position of the window
--- a/src/SFML/Window/WindowImpl.cpp
+++ b/src/SFML/Window/WindowImpl.cpp
@@ -35,6 +35,7 @@
 #include <SFML/System/Time.hpp>
 
 #include <array>
+#include <chrono>
 #include <memory>
 
 #include <cmath>
@@ -175,35 +176,46 @@ void WindowImpl::setMaximumSize(const std::optional<Vector2u>& maximumSize)
 
 
 ////////////////////////////////////////////////////////////
-Event WindowImpl::popEvent(bool block)
+Event WindowImpl::waitEvent(Time timeout)
+{
+    const auto timedOut = [&, startTime = std::chrono::steady_clock::now()]
+    {
+        const bool infiniteTimeout = timeout == Time::Zero;
+        return !infiniteTimeout && (std::chrono::steady_clock::now() - startTime) >= timeout.toDuration();
+    };
+
+    // If the event queue is empty, let's first check if new events are available from the OS
+    if (m_events.empty())
+        populateEventQueue();
+
+    // Here we use a manual wait loop instead of the optimized wait-event provided by the OS,
+    // so that we don't skip joystick events (which require polling)
+    while (m_events.empty() && !timedOut())
+    {
+        sleep(milliseconds(10));
+        populateEventQueue();
+    }
+
+    return popEvent();
+}
+
+
+////////////////////////////////////////////////////////////
+Event WindowImpl::pollEvent()
 {
     // If the event queue is empty, let's first check if new events are available from the OS
     if (m_events.empty())
-    {
-        // Get events from the system
-        processJoystickEvents();
-        processSensorEvents();
-        processEvents();
+        populateEventQueue();
 
-        // In blocking mode, we must process events until one is triggered
-        if (block)
-        {
-            // Here we use a manual wait loop instead of the optimized
-            // wait-event provided by the OS, so that we don't skip joystick
-            // events (which require polling)
-            while (m_events.empty())
-            {
-                sleep(milliseconds(10));
-                processJoystickEvents();
-                processSensorEvents();
-                processEvents();
-            }
-        }
-    }
+    return popEvent();
+}
 
+
+////////////////////////////////////////////////////////////
+Event WindowImpl::popEvent()
+{
     Event event;
 
-    // Pop the first event of the queue, if it is not empty
     if (!m_events.empty())
     {
         event = m_events.front();
@@ -308,6 +320,15 @@ void WindowImpl::processSensorEvents()
                 pushEvent(Event::SensorChanged{sensor, m_sensorValue[sensor]});
         }
     }
+}
+
+
+////////////////////////////////////////////////////////////
+void WindowImpl::populateEventQueue()
+{
+    processJoystickEvents();
+    processSensorEvents();
+    processEvents();
 }
 
 
--- a/test/Window/WindowBase.test.cpp
+++ b/test/Window/WindowBase.test.cpp
@@ -9,6 +9,7 @@
 #include <catch2/catch_test_macros.hpp>
 
 #include <WindowUtil.hpp>
+#include <chrono>
 #include <type_traits>
 
 TEST_CASE("[Window] sf::WindowBase", runDisplayTests())
@@ -104,14 +105,38 @@ TEST_CASE("[Window] sf::WindowBase", runDisplayTests())
 
     SECTION("pollEvent()")
     {
-        sf::WindowBase windowBase;
-        CHECK(!windowBase.pollEvent());
+        SECTION("Uninitialized window")
+        {
+            sf::WindowBase windowBase;
+            CHECK(!windowBase.pollEvent());
+        }
     }
 
     SECTION("waitEvent()")
     {
-        sf::WindowBase windowBase;
-        CHECK(!windowBase.waitEvent());
+        SECTION("Uninitialized window")
+        {
+            sf::WindowBase windowBase;
+            CHECK(!windowBase.waitEvent());
+        }
+
+        SECTION("Initialized window")
+        {
+            sf::WindowBase windowBase(sf::VideoMode({360, 240}), "WindowBase Tests");
+
+            constexpr auto timeout = sf::milliseconds(100);
+
+            const auto startTime = std::chrono::steady_clock::now();
+            const auto event     = windowBase.waitEvent(timeout);
+            const auto elapsed   = std::chrono::steady_clock::now() - startTime;
+
+            REQUIRE(elapsed < (timeout + sf::milliseconds(50)).toDuration());
+
+            if (elapsed <= timeout.toDuration())
+                CHECK(event);
+            else
+                CHECK(!event);
+        }
     }
 
     SECTION("Set/get position")
--- a/src/SFML/Window/WindowBase.cpp
+++ b/src/SFML/Window/WindowBase.cpp
@@ -149,17 +149,17 @@ bool WindowBase::isOpen() const
 Event WindowBase::pollEvent()
 {
     Event event;
-    if (m_impl && (event = m_impl->popEvent(false)))
+    if (m_impl && (event = m_impl->pollEvent()))
         filterEvent(event);
     return event;
 }
 
 
 ////////////////////////////////////////////////////////////
-Event WindowBase::waitEvent()
+Event WindowBase::waitEvent(Time timeout)
 {
     Event event;
-    if (m_impl && (event = m_impl->popEvent(true)))
+    if (m_impl && (event = m_impl->waitEvent(timeout)))
         filterEvent(event);
     return event;
 }
--- a/src/SFML/Window/WindowImpl.hpp
+++ b/src/SFML/Window/WindowImpl.hpp
@@ -55,6 +55,7 @@
 namespace sf
 {
 class String;
+class Time;
 
 namespace priv
 {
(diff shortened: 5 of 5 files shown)


src/SFML/Window/WindowImpl.cpp
```diff
@@ -175,30 +176,50 @@ void WindowImpl::setMaximumSize(const std::optional<Vector2u>& maximumSize)
 
 
 ////////////////////////////////////////////////////////////
-Event WindowImpl::popEvent(bool block)
+Event WindowImpl::popEventBlocking(Time timeout)
 {
+    const auto startTime = std::chrono::steady_clock::now();
```
vittorioromeo, line 181: I considered and tried using `sf::Clock` here but the code was just not as readable / understandable due to `sf::Clock` being stateful.

src/SFML/Window/WindowImpl.cpp
```diff
@@ -175,30 +176,50 @@ void WindowImpl::setMaximumSize(const std::optional<Vector2u>& maximumSize)
...
+
+    const auto timedOut = [&]
+    { return timeout != Time::Zero && (std::chrono::steady_clock::now() - startTime) < timeout.toDuration(); };
```
vittorioromeo, line 184: I considered taking a `std::optional<Time>` instead of using the `Time::Zero` magic value, but it would have been inconsistent with any other timeout.

vittorioromeo:

src/SFML/Window/WindowImpl.cpp
eXpl0it3r: What about: 

```suggestion
    static Clock clock;
    const auto startTime = clock.restart();
```

src/SFML/Window/WindowImpl.cpp
eXpl0it3r, line 184: If the lambda spreads across multiple lines, can we use:

```suggestion
    {
        return timeout != Time::Zero && (std::chrono::steady_clock::now() - startTime) < timeout.toDuration();
    };
```

src/SFML/Window/WindowImpl.cpp
```diff
@@ -175,30 +176,50 @@ void WindowImpl::setMaximumSize(const std::optional<Vector2u>& maximumSize)
...
+    Event event;
+
+    // Pop the first event of the queue, if it is not empty
+    if (!m_events.empty())
+    {
+        event = m_events.front();
+        m_events.pop();
+    }
+
+    return event;
```
eXpl0it3r, line 211: Maybe this shared code piece can be moved out to e.g. `popEvent`

include/SFML/Window/WindowBase.hpp
```diff
@@ -202,24 +203,28 @@ class SFML_WINDOW_API WindowBase
...
-    /// \return The event
+    /// \param timeout Maximum time to wait (`Time::Zero` for infinite)
+    ///
+    /// \return The event; will be `Empty` (convertible to `false`) on timeout or if window was closed
     ///
     /// \see pollEvent
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] Event waitEvent();
+    [[nodiscard]] Event waitEvent(Time timeout = Time::Zero);
```
eXpl0it3r, line 227: We shouldn't recommend Zero as default value, since we don't want threads to get stuck. I'd go with 5 or 10s.

```suggestion
    [[nodiscard]] Event waitEvent(Time timeout = seconds(5.f));
```

eXpl0it3r (changes requested):

src/SFML/Window/WindowImpl.cpp
vittorioromeo: Does not seem like an improvement to me TBH, and it introduces `static` function-object synchronization overhead.

src/SFML/Window/WindowImpl.cpp
vittorioromeo: This is unfortunately up to `clang-format`, not me...

src/SFML/Window/WindowImpl.cpp
vittorioromeo: Will extract

include/SFML/Window/WindowBase.hpp
vittorioromeo: Isn't that quite long? How about one second? Threads would use a loop anyway

src/SFML/Window/WindowImpl.cpp
eXpl0it3r: I argue it simplifies things, as we're not reinventing the wheel. Can you give a benchmark that this overhead is measurable in any meaningful way for event handling?

src/SFML/Window/WindowImpl.cpp
eXpl0it3r: Thought so, @ChrisThrasher any option we can tweak? 😄

src/SFML/Window/WindowImpl.cpp
vittorioromeo: I'd argue the opposite, `sf::Clock` is reinventing the wheel: the standard C++ way of measuring time since C++11 is `<chrono>`. 

`sf::Clock` is implemented on top of `<chrono>` to provide additional functionality that `<chrono>` does not provide which is unneeded here. We don't need a restartable/toggleable stopwatch, we just need the `::now()` function.

Following the principle to use the simplest tool to get the job done, I'm strongly in favour of sticking to `<chrono>` here.

include/SFML/Window/WindowBase.hpp
eXpl0it3r: I even considered a higher timeout. You're not necessarily *expecting* events at a constant rate. If a user doesn't touch their keyboard, mouse, controller for some time, you may not get any events.

The timeout is really there as a ripcord in case we end up in a state where it's guaranteed that we'll never be receiving any events anymore.

So going down to 1s would IMHO trigger the timeout too often, while I think if you had to wait for the process to terminate 5s is about acceptable.

src/SFML/Window/WindowImpl.cpp
vittorioromeo: If you feel strongly about the formatting here, I can also manually address it via `// clang-format off`

eXpl0it3r: No it's fine for now.

include/SFML/Window/WindowBase.hpp
vittorioromeo: > I even considered a higher timeout. You're not necessarily expecting events at a constant rate. If a user doesn't touch their keyboard, mouse, controller for some time, you may not get any events. [...]

But that's fine even with a shorter timeout, isn't it? The pattern would be:

```cpp
int main()
{
    std::atomic<bool> running{true};
    sf::RenderWindow rw(/* ... */);

    std::jthread eventHandlingThread{[&]
    {
        while (running)
            if (const sf::Event e = rw.waitEvent())
                // ...handle event...
    }};
    
    // ...do stuff that takes time...
    
    running.store(false);
}
```

The timeout only exists as a way of saying *"every X seconds, I'm going to check whether we can clean-up the event handling thread"*. 

I think that having a long default timeout of 5s or 10s might lead people to believe their app is hanging/freezing. I know I wouldn't have the patience to wait 10s and I'd start figuring out why my app is freezing (while in reality it isn't). 

> So going down to 1s would IMHO trigger the timeout too often

Triggering the timeout should not be an issue, what's your concern with it being triggered more often?

JonnyPtn: I'd say zero/infinite is a sensible default timeout - it matches the existing behaviour and I'd say is typical for blocking functions when a timeout isn't specified, rather than picking some arbitrary value

vittorioromeo: How about we just don't have a default at all? @eXpl0it3r

src/SFML/Window/WindowImpl.cpp
ChrisThrasher: Perhaps we revisit this library-wide in a later PR. Using `Time::Zero` as a sentinel value isn't something I love.

include/SFML/Window/WindowBase.hpp
eXpl0it3r: No, I think Jonny is right. We're adding this for an edge case in some people's code. We shouldn't change the behavior for everyone because of that.

src/SFML/Window/WindowImpl.cpp
```diff
@@ -175,42 +176,59 @@ void WindowImpl::setMaximumSize(const std::optional<Vector2u>& maximumSize)
...
+}
+
+
+////////////////////////////////////////////////////////////
+Event WindowImpl::popEventBlocking(Time timeout)
+{
+    const auto startTime = std::chrono::steady_clock::now();
+
+    const auto timedOut = [&]
```
ChrisThrasher, line 198:
```suggestion
    const auto timedOut = [timeout, startTime = std::chrono::steady_clock::now()]
```
I'd prefer with be explicit with lambda captures.

src/SFML/Window/WindowImpl.cpp
```diff
@@ -175,42 +176,59 @@ void WindowImpl::setMaximumSize(const std::optional<Vector2u>& maximumSize)
...
+    { return timeout != Time::Zero && (std::chrono::steady_clock::now() - startTime) < timeout.toDuration(); };
```
ChrisThrasher, line 199: Shouldn't that `<` become a `>=`? If `true` == `a timeout has occurred` then we want to return true when the actual duration exceeds or equals the timeout.

src/SFML/Window/WindowImpl.cpp
```diff
@@ -175,42 +176,59 @@ void WindowImpl::setMaximumSize(const std::optional<Vector2u>& maximumSize)
...
-                processSensorEvents();
-                processEvents();
-            }
-        }
+    // Here we use a manual wait loop instead of the optimized
+    // wait-event provided by the OS, so that we don't skip joystick
+    // events (which require polling)
+    while (m_events.empty() && !timedOut())
+    {
+        sleep(milliseconds(10));
```
ChrisThrasher, line 213: This sleep duration imposes a sort of minimum timeout. If you want a timeout of 1 millisecond like in the tests you're still going to wait > 10 ms. Perhaps we document that timeouts under 10 milliseconds won't really work as intended. Alternatively maybe we make this duration some fraction of the provided timeout?

test/Window/WindowBase.test.cpp
```diff
@@ -104,14 +104,29 @@ TEST_CASE("[Window] sf::WindowBase", runDisplayTests())
...
+            CHECK(!windowBase.waitEvent());
+        }
+
+        SECTION("Initialized window")
+        {
+            sf::WindowBase windowBase(sf::VideoMode({360, 240}), "WindowBase Tests");
+
+            // Wait for an event with a timeout, doesn't matter if it is received or not,
+            // the purpose of this code is to test the timeout itself
+            [[maybe_unused]] const sf::Event event = windowBase.waitEvent(sf::milliseconds(1));
```
ChrisThrasher, line 128: I think we can take these tests further. How about we measure how long the function takes to return and ensure it takes less than the given timeout value with some margin of error. The margin of error will take some consideration. Here's what I came up with. Unfortunately I still get occasional failures so these tests are too flaky in their current state. Hopefully we can find a more robust way of doing this.

```cpp
        SECTION("Initialized window")
        {
            sf::WindowBase windowBase(sf::VideoMode({360, 240}), "WindowBase Tests");

            constexpr auto timeout = sf::milliseconds(200);
            const auto     now     = std::chrono::steady_clock::now();
            const auto     event   = windowBase.waitEvent(timeout);
            const auto     elapsed = std::chrono::steady_clock::now() - now;
            REQUIRE(elapsed < (timeout + sf::milliseconds(10)).toDuration());
            if (elapsed <= timeout.toDuration())
                CHECK(event);
            else
                CHECK(!event);
        }
```

ChrisThrasher (changes requested):

src/SFML/Window/WindowImpl.cpp
eXpl0it3r: Choosing timeouts in the range of milliseconds defeats the purpose of the timeout. While they technically could set it, I don't see it as problematic to have a not documented 10ms delay

vittorioromeo: @eXpl0it3r: how about making the poll interval a parameter as well, and mentioning the relationship between timeout and polling in the documentation?

eXpl0it3r: No, this is an implementation detail. If we didn't decide to track sf::Joystick states via event data, this wouldn't even exist and we'd be using OS-level waiting.

include/SFML/Window/WindowBase.hpp
ChrisThrasher: I think this timeout should remain infinite by default. Let's not change the default behavior.

src/SFML/Window/WindowImpl.cpp
vittorioromeo: This lambda does not outlive the scope of the function it's defined in, so it's entirely safe to use `[&]` and it would be overly verbose to provide the captures explicitly. However, I like how using generalized lambda captures reduces the scope of `startTime`, so I will add that in.

src/SFML/Window/WindowImpl.cpp
vittorioromeo: Resolving, feel free to re-open if you strongly feel we should use `sf::Clock`.

src/SFML/Window/WindowImpl.cpp
vittorioromeo: > Perhaps we revisit this library-wide in a later PR. Using `Time::Zero` as a sentinel value isn't something I love.

I completely agree, `Time::Zero` can be interpreted as "immediate timeout" (i.e. behave as if polling) instead of "infinite timeout". I'd prefer to see `std::optional<Time>` in all our interfaces and use the empty state as "absence of a timeout".

I can work on that on a separate PR after this one.

include/SFML/Window/WindowBase.hpp
vittorioromeo: Fair enough, I'll change the timeout to be consistent with the other functions here in preparation of a PR for https://github.com/SFML/SFML/pull/3094#discussion_r1637119797

src/SFML/Window/WindowImpl.cpp
vittorioromeo: Yes, this was a poor copy-paste mistake on my part while extracting the loop condition into a lambda.

test/Window/WindowBase.test.cpp
vittorioromeo: I'm going to try this out, I like the idea.

vittorioromeo:
> Might as well merge `waitEvent` and `pollEvent` at this point - Fundamentally both functions are just used to get the next event.
> 
> Previously it was necessary to have different functions because they have the same signature, and the only difference was behaviour, but now the timeout parameter dictates the behaviour I don't see any benefit in separating them, as `pollEvent` is effectively a specialisation of `waitEvent` with a zero timeout

I think there is still value in having two separate names for two conceptually different actions. Implementation-wise, I'd still keep them separate to avoid unnecessarily getting the current time and verifying a timeout check that I know it will always evaluate to `false`.

vittorioromeo:

ChrisThrasher (approved):

vittorioromeo:
Waiting for @eXpl0it3r's final feedback on this PR.

src/SFML/Window/WindowImpl.cpp
```diff
@@ -175,35 +176,10 @@ void WindowImpl::setMaximumSize(const std::optional<Vector2u>& maximumSize)
 
 
 ////////////////////////////////////////////////////////////
-Event WindowImpl::popEvent(bool block)
+Event WindowImpl::popEventImpl()
```
eXpl0it3r, line 179: Why popEvent**Impl**?

If you think it's too confusing with the other names, I'd recommend taking something longer than just appending `Impl`.

I also recommend moving the function below the other two functions.

test/Window/WindowBase.test.cpp
```diff
@@ -104,14 +105,38 @@ TEST_CASE("[Window] sf::WindowBase", runDisplayTests())
...
+
+        SECTION("Initialized window")
+        {
+            sf::WindowBase windowBase(sf::VideoMode({360, 240}), "WindowBase Tests");
+
+            constexpr auto timeout = sf::milliseconds(250);
+
+            const auto startTime = std::chrono::steady_clock::now();
+            const auto event     = windowBase.waitEvent(timeout);
+            const auto elapsed   = std::chrono::steady_clock::now() - startTime;
```
eXpl0it3r, line 131: I recommend using an `sf::Clock`

test/Window/WindowBase.test.cpp
```diff
@@ -104,14 +105,38 @@ TEST_CASE("[Window] sf::WindowBase", runDisplayTests())
...
+        {
+            sf::WindowBase windowBase;
+            CHECK(!windowBase.waitEvent());
+        }
+
+        SECTION("Initialized window")
+        {
+            sf::WindowBase windowBase(sf::VideoMode({360, 240}), "WindowBase Tests");
+
+            constexpr auto timeout = sf::milliseconds(250);
```
eXpl0it3r, line 127: Can we take a lower value? I wouldn't want to delay the test run by a quarter of a second. This also increases the risk that on a real system you end up getting an event within those 250ms.

eXpl0it3r (changes requested):

test/Window/WindowBase.test.cpp
vittorioromeo: I would recommend against it for the same reasons described in https://github.com/SFML/SFML/pull/3094#discussion_r1635519805. Resolving right now, feel free to re-open if you feel strongly about this recommendation.

src/SFML/Window/WindowImpl.cpp
vittorioromeo: I'll move the function and see if I can come up with a better name. Suggestions?

test/Window/WindowBase.test.cpp
vittorioromeo: I will tweak the values.

eXpl0it3r (approved):

vittorioromeo:

PR #3096, opened 2024-06-12
`break` event handling loop after `window.close()`

vittorioromeo:
Extracted from #2992. Adds a `break;` right after a `window.close()` call in event handling loops, as there's nothing useful that can be done since the window will be closed immediately.

Rationale:
- Technically a tiny run-time optimization.
- Promotes good practice of exiting from the event handling loop when the window is closed, as neither `pollEvent` nor `waitEvent` will produce any event if the window has been closed.


vittorioromeo:

vittorioromeo:

PR #3097, opened 2024-06-13
Testable debug run-time lifetime tracking

vittorioromeo:
Add a new `SFML/System/LifetimeTracking.hpp` header that provides testable run-time lifetime tracking. 

---

The tracking is enabled only when `SFML_ENABLE_LIFETIME_TRACKING` is defined. That macro is controlled by a CMake option of the same name, enabled by default when building in debug mode.

The tracking is also enabled for our CI test suite, regardless of the build mode.

---

The tracking catches common lifetime mistakes between dependee types (e.g. `Texture`) and dependant types (e.g. `Sprite`) at run-time, providing the user with a readable error message:

> ```
> FATAL ERROR: a texture object was destroyed while existing sprite objects depended on it.
> 
> Please ensure that every texture object outlives all of the sprite objects associated with it,
> otherwise those sprites will try to access the memory of the destroyed texture, 
> causing undefined behavior (e.g., crashes, segfaults, or unexpected run-time behavior).
>  
> One of the ways this issue can occur is when a texture object is created as a local variable 
> in a function and passed to a sprite object. When the function has finished executing, the 
> local texture object will be destroyed, and the sprite object associated with it will now be
> referring to invalid memory. Example:
> 
>     sf::Sprite createSprite()
>     {
>         sf::Texture texture(/* ... */);
>         sf::Sprite sprite(texture, /* ... */);
> 
>         return sprite;
>         //     ^~~~~~
>         // ERROR: `texture` will be destroyed right after
>         //        `sprite` is returned from the function!
>     }
> 
> Another possible cause of this error is storing both a texture and a sprite together in a
> data structure (e.g., `class`, `struct`, container, pair, etc...), and then moving that 
> data structure (i.e., returning it from a function, or using `std::move`) -- the internal 
> references between the texture and sprite will not be updated, resulting in the same 
> lifetime issue.
> 
> In general, make sure that all your texture objects are destroyed *after* all the 
> sprite objects depending on them to avoid these sort of issues.
> ```

Note that the error message generation dynamically changes depending on the object types, it's not hardcoded for sprites and textures specifically.

---

Adding lifetime tracking to SFML types is trivial and requires minimum changes, as shown by the diff. It consists in a few simple steps:

- In the dependee type (e.g. `Texture`):
    1. Forward-declare the dependent type (e.g. `Sprite`);
    2. Add the following code at the end of the dependee type's `private` section:
    
    ```cpp
    SFML_DEFINE_LIFETIME_DEPENDEE(Texture, Sprite);
    ```
    
- In the dependant type (e.g. `Sprite`):
    1. Add the following code at the end of the dependant type's `private` section:

    ```cpp
    SFML_DEFINE_LIFETIME_DEPENDANT(Texture);
    ```
    
    2. Add the following code in any function that changes the dependant type's pointer-to-dependee:
    
    ```cpp
    SFML_UPDATE_LIFETIME_DEPENDANT(Texture, Sprite, m_texture);
    ```
    
That's it!

---

For testing, use the provided `sf::priv::LifetimeDependee::TestingModeGuard`:

```cpp
SECTION("Return local from function")
{
    const auto badFunction = []
    {
        const auto texture = sf::Texture::create({64, 64}).value();
        return sf::Sprite(texture);
    };

    const sf::priv::LifetimeDependee::TestingModeGuard guard;
    CHECK(!guard.fatalErrorTriggered());

    badFunction();

    CHECK(guard.fatalErrorTriggered());
}
```

---

Drawbacks:

- Some valid constructs might need to be slightly rearranged to please the lifetime tracker. E.g.:

    ```cpp
    // Techincally valid, but triggers a fatal error
    {
        sf::Sprite sprite(texture);
        const sf::Texture otherTexture = sf::Texture::create({64, 64}).value();
        sprite.setTexture(otherTexture);
    }

    // Valid, and pleases the lifetime tracker
    {
        const sf::Texture otherTexture = sf::Texture::create({64, 64}).value();
    
        sf::Sprite sprite(texture);
        sprite.setTexture(otherTexture);
    }
    ```
    
    However, this is more of a design problem with `sf::Sprite` itself than with the lifetime tracker, as `sf::Sprite` only needs a reference to the texture during its draw call, but SFML artificially extends the relationship between `sf::Sprite` and `sf::Texture` way beyond that. See #3072 and #3080 for an ample discussion on that.
    
---
    
FAQ:

- > Is this useful for newbies?

  - Yes. Newbies might not be familiar with the concepts of lifetime and ownership in C++, and might inadvertedly get into a situation where a dependee is destroyed before a dependant. This PR ensures that they will be alerted of the issue as soon as possible, saving them minutes or hours of frustration, debugging, saving us from having to help them out on Discord, and prompting them to learn more about lifetimes and ownership in C++.

- > Is this useful for power users?
 
  - Yes. While powerusers are familiar with lifetimes and don’t typically struggle with them, they can still make mistakes. This PR aims to alert powerusers to subtle lifetime issues that might arise from complex interactions, refactoring, or occasional human errors. It's not about teaching lifetimes but providing a safety net for those rare mistakes.
  
- > Is there any overhead if lifetime tracking is disabled?

  - Nope. Disabled lifetime tracking incurs no memory or run-time overhead at all.

### DIFF
--- a/include/SFML/Audio/Sound.hpp
+++ b/include/SFML/Audio/Sound.hpp
@@ -31,6 +31,8 @@
 
 #include <SFML/Audio/SoundSource.hpp>
 
+#include <SFML/System/LifetimeTracking.hpp>
+
 #include <memory>
 
 #include <cstdlib>
@@ -244,6 +246,11 @@ private:
     ////////////////////////////////////////////////////////////
     struct Impl;
     const std::unique_ptr<Impl> m_impl; //!< Implementation details
+
+    ////////////////////////////////////////////////////////////
+    // Lifetime tracking
+    ////////////////////////////////////////////////////////////
+    SFML_DEFINE_LIFETIME_DEPENDANT(SoundBuffer);
 };
 
 } // namespace sf
--- a/test/Graphics/Text.test.cpp
+++ b/test/Graphics/Text.test.cpp
@@ -3,10 +3,13 @@
 // Other 1st party headers
 #include <SFML/Graphics/Font.hpp>
 
+#include <SFML/System/LifetimeTracking.hpp>
+
 #include <catch2/catch_test_macros.hpp>
 
 #include <GraphicsUtil.hpp>
 #include <WindowUtil.hpp>
+#include <optional>
 #include <type_traits>
 
 TEST_CASE("[Graphics] sf::Text", runDisplayTests())
@@ -182,4 +185,47 @@ TEST_CASE("[Graphics] sf::Text", runDisplayTests())
             CHECK(text.getGlobalBounds() == Approx(sf::FloatRect({66, 182}, {33, 13})));
         }
     }
+
+#ifdef SFML_ENABLE_LIFETIME_TRACKING
+    SECTION("Lifetime tracking")
+    {
+        SECTION("Return local from function")
+        {
+            const auto badFunction = []
+            {
+                const auto localFont = sf::Font::loadFromFile("Graphics/tuffy.ttf").value();
+                return sf::Text(localFont);
+            };
+
+            const sf::priv::LifetimeDependee::TestingModeGuard guard;
+            CHECK(!guard.fatalErrorTriggered());
+
+            badFunction();
+
+            CHECK(guard.fatalErrorTriggered());
+        }
+
+        SECTION("Move struct holding both dependee and dependant")
+        {
+            struct BadStruct
+            {
+                sf::Font memberFont{sf::Font::loadFromFile("Graphics/tuffy.ttf").value()};
+                sf::Text memberText{memberFont};
+            };
+
+            const sf::priv::LifetimeDependee::TestingModeGuard guard;
+            CHECK(!guard.fatalErrorTriggered());
+
+            std::optional<BadStruct> badStruct0;
+            badStruct0.emplace();
+            CHECK(!guard.fatalErrorTriggered());
+
+            const BadStruct badStruct1 = std::move(badStruct0.value());
+            CHECK(!guard.fatalErrorTriggered());
+
+            badStruct0.reset();
+            CHECK(guard.fatalErrorTriggered());
+        }
+    }
+#endif
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
         - { name: Shared, flags: -DBUILD_SHARED_LIBS=TRUE }
         - { name: Static, flags: -DBUILD_SHARED_LIBS=FALSE }
         type:
-        - { name: Release }
+        - { name: Release, flags: -DSFML_ENABLE_LIFETIME_TRACKING=TRUE }
         - { name: Debug, flags: -DCMAKE_BUILD_TYPE=Debug -DSFML_ENABLE_COVERAGE=TRUE }
 
         include:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,12 @@ if(SFML_OS_ANDROID)
     set(BUILD_SHARED_LIBS TRUE)
 endif()
 
+# add an option to enable lifetime tracking
+sfml_set_option(SFML_ENABLE_LIFETIME_TRACKING $<CONFIG:Debug> BOOL "TRUE to enable lifetime tracking, FALSE to disable it")
+if(SFML_ENABLE_LIFETIME_TRACKING)
+    add_definitions(-DSFML_ENABLE_LIFETIME_TRACKING)
+endif()
+
 # add options to select which modules to build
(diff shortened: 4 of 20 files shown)


vittorioromeo:
@Bambo-Borris: Thanks for the feedback!

You raised three main points:

1. The error message could be improved.
2. Newbies might not benefit since they're not familiar with lifetimes.
3. Powerusers don’t struggle with lifetimes.

Here are my thoughts:

1. I agree the error message has room for improvement. We could use simpler language tailored to beginners, include a code example, or link to resources like the "white square problem" page on the SFML website. This could make the message more accessible and helpful for newbies.

2. By tweaking the language of the error message to be newbie-friendly, avoiding terms like "dependee" or "dependant", and providing clear examples or links, I believe beginners will find the message valuable and easier to understand.

3. While powerusers are familiar with lifetimes and don’t typically struggle with them, they can still make mistakes. This PR aims to alert powerusers to subtle lifetime issues that might arise from complex interactions, refactoring, or occasional human errors. It's not about teaching lifetimes but providing a safety net for those rare mistakes.

vittorioromeo:

vittorioromeo:
New error message example:

> ```
> FATAL ERROR: a texture object was destroyed while existing sprite objects depended on it.
> 
> Please ensure that every texture object outlives all of the sprite objects associated with it,
> otherwise those sprites will try to access the memory of the destroyed texture, 
> causing undefined behavior (e.g., crashes, segfaults, or unexpected run-time behavior).
>  
> One of the ways this issue can occur is when a texture object is created as a local variable 
> in a function and passed to a sprite object. When the function has finished executing, the 
> local texture object will be destroyed, and the sprite object associated with it will now be
> referring to invalid memory. Example:
> 
>     sf::Sprite createSprite()
>     {
>         sf::Texture texture(/* ... */);
>         sf::Sprite sprite(texture, /* ... */);
> 
>         return sprite;
>         //     ^~~~~~
> 
>         // ERROR: `texture` will be destroyed right after
>         //        `sprite` is returned from the function!
>     }
> 
> Another possible cause of this error is storing both a texture and a sprite together in a
> data structure (e.g., `class`, `struct`, container, pair, etc...), and then moving that 
> data structure (i.e., returning it from a function, or using `std::move`) -- the internal 
> references between the texture and sprite will not be updated, resulting in the same 
> lifetime issue.
> 
> In general, make sure that all your texture objects are destroyed *after* all the 
> sprite objects depending on them to avoid these sort of issues.
> ```

Note that the error message generation dynamically changes depending on the object types, it's not hardcoded for sprites and textures specifically.

@Bambo-Borris: What do you think?

vittorioromeo:

test/Graphics/Text.test.cpp
```diff
@@ -182,4 +185,47 @@ TEST_CASE("[Graphics] sf::Text", runDisplayTests())
...
+    {
+        SECTION("Return local from function")
+        {
+            const auto badFunction = []
+            {
+                const auto localFont = sf::Font::loadFromFile("Graphics/tuffy.ttf").value();
+                return sf::Text(localFont);
+            };
+
+            const sf::priv::LifetimeDependee::TestingModeGuard guard;
```
ChrisThrasher, line 200: The tests currently make no mention of the `priv::` namespace and I'd like to keep it that way. There are definitely some `sf::priv::` things I wish we could test but I'm rather dogmatic about the unit tests only using the public API and am not comfortable testing implementation details like this which users are not intended to use or even know about.

How much code could be removed from this PR if `TestingModeGuard` was removed?

This code will still be tested although instead of explicitly testing for failure we will be implicitly testing for the lack of failure. I know that's not the same but by simply using types like `sf::Sprite` we will still be testing that your lifetime tracking has no false positives. We cannot test for true negatives quite as well though but I think I'm okay with that. It's the same as all of our `assert`s. We can't test for assertion failure but we can test that assertions don't have false positives. In practice this seems to work quite well.

include/SFML/Audio/Sound.hpp
```diff
@@ -244,6 +246,11 @@ class SFML_AUDIO_API Sound : public SoundSource
     ////////////////////////////////////////////////////////////
     struct Impl;
     const std::unique_ptr<Impl> m_impl; //!< Implementation details
+
+    ////////////////////////////////////////////////////////////
+    // Lifetime tracking
+    ////////////////////////////////////////////////////////////
+    SFML_DEFINE_LIFETIME_DEPENDANT(SoundBuffer);
```
ChrisThrasher, line 253: `sf::Sound` and `sf::SoundBuffer` already perform their own internal lifetime checks where sound buffers are de-registered with the `Sound`s that use them when the `SoundBuffer` is destroyed. Do these new lifetime checks means we can remove all that machinery from `Sound` and `SoundBuffer`?

vittorioromeo:
@ChrisThrasher example of lifetime issue that this PR would have caught: https://discord.com/channels/175298431294636032/175298431294636032/1262381178523222086

PR #3099, opened 2024-06-16
Create a new font page in Font::loadPage only when needed

kimci86:
While updating [SFML-Input](https://github.com/eXpl0it3r/SFML-Input) to work with latest in-development SFML version, I noticed performance was suddenly very bad.
It turns out `Font::loadPage` was creating a `Font::Page` instance and its corresponding `Texture` instance for nothing most of the time.
This commit solves the performance problem.
For reference, the bug was introduced in #3027.

### DIFF
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -464,9 +464,12 @@ bool Font::isSmooth() const
 ////////////////////////////////////////////////////////////
 Font::Page& Font::loadPage(unsigned int characterSize) const
 {
+    if (const auto it = m_pages.find(characterSize); it != m_pages.end())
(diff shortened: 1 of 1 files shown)


kimci86:
A simpler way to test this is running the `island` example as see the frame time text display big erratic values without this commit, or stable values around 16 ms with this commit.

vittorioromeo:

vittorioromeo:
Thanks for this PR. It also explains a performance degradation I saw in Open Hexagon recently!

vittorioromeo (approved):
LGTM. I also checked if using `emplace_hint` would make sense in this case, but for unordered containers it's not really worth it.

PR #3104, opened 2024-06-17
Fix sf::VideoMode usage example

kimci86:
A small inconsistency I noticed while looking at `.width` and `.height` occurrences during my work on `sf::Rect` type.
For reference, `width` and `height` were replaced by a `size` vector in #2055.


ChrisThrasher (approved):

vittorioromeo:

PR #3106, opened 2024-06-17
Replace sentinel values with `std::optional`

ChrisThrasher:
## Description

In both of these cases a signed int is used to gain access to `-1` as a sentinel value even though every time these variables are used they're cast to an unsigned integer. This is a good sign that `std::optional` is a better representation.

### DIFF
--- a/examples/vulkan/Vulkan.cpp
+++ b/examples/vulkan/Vulkan.cpp
@@ -701,12 +701,12 @@ public:
 
             if ((queueFamilyProperties[i].queueFlags & VK_QUEUE_GRAPHICS_BIT) && (surfaceSupported == VK_TRUE))
             {
-                queueFamilyIndex = static_cast<int>(i);
+                queueFamilyIndex = static_cast<std::uint32_t>(i);
                 break;
             }
         }
 
-        if (queueFamilyIndex < 0)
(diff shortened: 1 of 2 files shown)


vittorioromeo (approved):
LGTM

kimci86:
What about [line 468](https://github.com/SFML/SFML/pull/3106/files#diff-ac0f2ef01d1c9ccb45629cd3abb0707fb7b5a0388e5ce47084e8454b669ba9b5R465)?
This is silently converting a signed int (I guess it could be -1) to optional unsigned int.
I did not test but I suspect we now send a `TouchBegan` event after a `TouchEnded` because `slot.id` would have a value (-1 as unsigned int) instead of being an empty optional.

ChrisThrasher:
> What about [line 465](https://github.com/SFML/SFML/pull/3106/files#diff-ac0f2ef01d1c9ccb45629cd3abb0707fb7b5a0388e5ce47084e8454b669ba9b5R465)? This is silently converting a signed int (I guess it could be -1) to optional unsigned int. I did not test but I suspect we now send a `TouchBegan` event after a `TouchEnded` because `slot.id` would have a value (-1 as unsigned int) instead of being an empty optional.

https://github.com/SFML/SFML/pull/3112

PR #3110, opened 2024-06-18
Replace `static_cast`s with vector operations

ChrisThrasher:
## Description

Along the same lines as #3061 and #2692. The resulting code is shorter and in my opinion easier to read.

### DIFF
--- a/src/SFML/Window/iOS/SensorImpl.mm
+++ b/src/SFML/Window/iOS/SensorImpl.mm
@@ -127,51 +127,42 @@ void SensorImpl::close()
 ////////////////////////////////////////////////////////////
 Vector3f SensorImpl::update()
 {
-    Vector3f               value;
     CMMotionManager* const manager = [SFAppDelegate getInstance].motionManager;
 
     switch (m_sensor)
     {
         case Sensor::Type::Accelerometer:
             // Acceleration is given in G, convert to m/s^2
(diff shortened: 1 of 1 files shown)


src/SFML/Window/iOS/SensorImpl.mm
```diff
@@ -127,51 +127,42 @@
...
+                                           manager.deviceMotion.userAcceleration.y,
+                                           manager.deviceMotion.userAcceleration.z));
 
         case Sensor::Type::Orientation:
             // Absolute rotation (Euler) angles are given in radians
-            value.x = static_cast<float>(manager.deviceMotion.attitude.yaw);
-            value.y = static_cast<float>(manager.deviceMotion.attitude.pitch);
-            value.z = static_cast<float>(manager.deviceMotion.attitude.roll);
-            break;
+            return Vector3f(Vector3<double>(manager.deviceMotion.attitude.yaw,
```
kimci86, line 159: Is there a reason to be explicit about the template argument `double` only here?

ChrisThrasher: I tried omitting it but CTAD failed for some reason when I did that.

kimci86: Ok. Maybe this is related to `yaw`, `pitch` and `roll` being Objective-C properties. :thinking:

eXpl0it3r (approved):

PR #3111, opened 2024-06-18
Windows ARM64 support for SFML with MSVC

Zombieschannel:
## Description

As discussed in the SFML Discord server, recently there have been some Windows on ARM64 devices showing up like [Microsoft Surface Pro 10](https://www.xda-developers.com/microsoft-surface-pro-10-arm-may-20/). So I've build ARM64 libs & dlls for SFML dependencies (due to some bugs with OpenAL 1.21.0, I upgraded it to 1.21.1 which has a [fix](https://github.com/eXpl0it3r/SFML-dependencies/pull/13)).

I added these files to msvc-universal folder considering that the regular msvc folder is used if MSVC version is less than 14 (Visual Studio 2013 and below) and I cannot find a msvc compiler for ARM64 older than Visual Studio 2017.

## Tasks

-   [ ] Tested on Linux
-   [X] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android


vittorioromeo:
I'm getting one of those machines this week, so this is highly appreciated! I believe our long-term plan is to build everything from source, but this seems a good stopgap measure until we get there.

cmake/Config.cmake
```diff
@@ -17,6 +17,11 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
         message(FATAL_ERROR "Unsupported architecture")
         return()
     endif()
+
+    #compile for ARM64 if set
+    if(${CMAKE_GENERATOR_PLATFORM} MATCHES "ARM64")
+        set(ARCH_ARM64 1)
+    endif()
```
ChrisThrasher, line 24: I think we need to change some variable names to make it clear we're talking about 32 and 64 bit Intel architectures. Also I think all 3 of these architectures, which are independent, should be checked in the same `if`/`else if` chain because we don't want `ARCH_64BITS` and `ARCH_ARM64` set at the same time which is something that is currently happening.

This is what I propose.

```cmake
    if(${CMAKE_GENERATOR_PLATFORM} MATCHES "ARM64")
        set(ARCH_ARM64 1)
    elseif(${SIZEOF_VOID_PTR} STREQUAL "4")
        set(ARCH_X86 1)
    elseif(${SIZEOF_VOID_PTR} STREQUAL "8")
        set(ARCH_X64 1)
    else()
        message(FATAL_ERROR "Unsupported architecture")
        return()
    endif()
```

src/SFML/CMakeLists.txt
```diff
@@ -21,10 +21,14 @@ if (SFML_OS_WINDOWS)
                 set(CMAKE_LIBRARY_PATH ${CMAKE_LIBRARY_PATH} "${PROJECT_SOURCE_DIR}/extlibs/libs-msvc/x64")
             endif()
         else()
-            if(ARCH_32BITS)
-                set(CMAKE_LIBRARY_PATH ${CMAKE_LIBRARY_PATH} "${PROJECT_SOURCE_DIR}/extlibs/libs-msvc-universal/x86")
-            elseif(ARCH_64BITS)
-                set(CMAKE_LIBRARY_PATH ${CMAKE_LIBRARY_PATH} "${PROJECT_SOURCE_DIR}/extlibs/libs-msvc-universal/x64")
+            if(ARCH_ARM64)
+                set(CMAKE_LIBRARY_PATH ${CMAKE_LIBRARY_PATH} "${PROJECT_SOURCE_DIR}/extlibs/libs-msvc-universal/ARM64")
```
ChrisThrasher, line 25: As a minor point of style, is it `ARM64` or `Arm64`? What is the right capitalization to use?

CMakeLists.txt
```diff
@@ -446,13 +446,18 @@ if(SFML_OS_WINDOWS)
                 install(DIRECTORY extlibs/libs-mingw/x86/ DESTINATION ${CMAKE_INSTALL_LIBDIR})
             endif()
         elseif(ARCH_64BITS)
-            install(DIRECTORY extlibs/bin/x64/ DESTINATION ${CMAKE_INSTALL_BINDIR})
-            if(SFML_COMPILER_MSVC AND SFML_MSVC_VERSION LESS 14)
-                install(DIRECTORY extlibs/libs-msvc/x64/ DESTINATION ${CMAKE_INSTALL_LIBDIR})
-            elseif(SFML_COMPILER_MSVC OR (SFML_COMPILER_CLANG AND NOT MINGW))
-                install(DIRECTORY extlibs/libs-msvc-universal/x64/ DESTINATION ${CMAKE_INSTALL_LIBDIR})
+            if(ARCH_ARM64)
```
ChrisThrasher, line 449: Move this up one logical level so it's on the same level with the 2 Intel architectures we're already checking.

ChrisThrasher (changes requested):
Can you talk more about exactly how you compiled all of these deps? I'd like to know more about how these binaries were made.

Zombieschannel:
> Can you talk more about exactly how you compiled all of these deps? I'd like to know more about how these binaries were made.

Sure, so eXpl0it3r has an SFML-dependencies repo and if you clone it and do a `cmake -B build -A ARM64` + a build of everything you'll unfortunately get this OpenAL error: https://github.com/kcat/openal-soft/issues/494. It turns out that this issue has been fixed in 1.21.1, so I made a fork of eXpl0it3r's repo and upgraded OpenAL version from 1.21.0 to 1.21.1: https://github.com/eXpl0it3r/SFML-dependencies/pull/13. After these changes, ARM64 version of OpenAL compiles fine and there you have all of the libs and openal32.dll for ARM64.

I used ARM64 in all caps since CMake uses it with all caps.

ChrisThrasher:

Zombieschannel:
Everything should be fixed now.

eXpl0it3r:

ChrisThrasher (approved):
Thank you for filling this gap in SFML's platform support! I think this looks good to me. For now I think it's reasonable to treat Windows ARM support as "community maintained" unless Vittorio volunteers to take it upon himself to maintain this going forward. I'll let him decide if he wants to take that upon himself.

I still want to wait until Vittorio and/or Exploiter approve this before we merge since they're more invested in SFML's Windows support than I am.

eXpl0it3r (approved):
We'll only be seeing an uptick in questions around  Windows ARM64, so I think this is the right approach, even if we don't really have the hardware to test it.

Two points:
- We may want to update OpenAL to the same version for all the other platforms
- I created #3113 for the failing macOS builds

CMakeLists.txt
```diff
@@ -436,7 +436,10 @@ if(SFML_OS_WINDOWS)
 
     if(NOT SFML_USE_SYSTEM_DEPS)
         # install the binaries of SFML dependencies
-        if(ARCH_32BITS)
+        if(ARCH_ARM64)
+            install(DIRECTORY extlibs/bin/ARM64/ DESTINATION ${CMAKE_INSTALL_BINDIR})
+            install(DIRECTORY extlibs/libs-msvc-universal/ARM64/ DESTINATION ${CMAKE_INSTALL_LIBDIR})
```
ChrisThrasher, line 441:
```suggestion
            install(DIRECTORY extlibs/libs-msvc-universal/Arm64/ DESTINATION ${CMAKE_INSTALL_LIBDIR})
```
https://www.arm.com/company/policies/trademarks/guidelines-trademarks#:~:text=Never%20alter%20Arm's%20trademarks%3A,)%2C%20or%20using%20improper%20capitalization.

Arm themselves don't capitalize all three letters. I think it's correct to keep using all caps for CMake variables but I think for directories we can use the more proper "Arm" style.

eXpl0it3r: It's tricky. If you look around you'll find a lot of different handling of this.

I searched a bit earlier today and came to the conclusion that ARM64 is a good fit. Windows documentation and IDE mostly refer to it as ARM64, so does CMake, etc.

ChrisThrasher: Fine by me if we stick with the all-caps directory name.

vittorioromeo:
I managed to compile SFML with your prebuilt ARM64 libraries on my Surface Pro 11, seems to work fine -- nice work! 

I had to change a few things for SFML 3.x (see https://github.com/SFML/SFML/compare/master...vittorioromeo:SFML:win11_arm_support?expand=1). Notably, I had to use CMAKE_SYSTEM_PROCESSOR because the CMAKE_GENERATOR_PLATFORM macro for me evaluated to an empty string. Not sure which one is more robust/portable, but worth researching for sure.

Zombieschannel:
> I managed to compile SFML with your prebuilt ARM64 libraries on my Surface Pro 11, seems to work fine -- nice work!
> 
> I had to change a few things for SFML 3.x (see https://github.com/SFML/SFML/compare/master...vittorioromeo:SFML:win11_arm_support?expand=1). Notably, I had to use CMAKE_SYSTEM_PROCESSOR because the CMAKE_GENERATOR_PLATFORM macro for me evaluated to an empty string. Not sure which one is more robust/portable, but worth researching for sure.

Yeah that makes sense, since I'm quite sure `${CMAKE_GENERATOR_PLATFORM}` is connected with the `-A ARM64` argument when compiling from the command line, so if you don't specify an architecture it will be empty and therefore not match the `ARM64`.

Could we just add an elseif statement for the default architecture on ARM64 to be ARM64? Something like this:
```
if(${CMAKE_GENERATOR_PLATFORM} MATCHES "ARM64")
    set(ARCH_ARM64 1)
elseif("${CMAKE_GENERATOR_PLATFORM}" STREQUAL "" AND ${CMAKE_SYSTEM_PROCESSOR} MATCHES "ARM64")
    set(ARCH_ARM64 1)
elseif(${SIZEOF_VOID_PTR} STREQUAL "4")
    set(ARCH_X86 1)
elseif(${SIZEOF_VOID_PTR} STREQUAL "8")
    set(ARCH_X64 1)
else()
    message(FATAL_ERROR "Unsupported architecture")
    return()
endif()
```

eXpl0it3r:

eXpl0it3r:
> ```
> if(${CMAKE_GENERATOR_PLATFORM} MATCHES "ARM64")
>     set(ARCH_ARM64 1)
> elseif("${CMAKE_GENERATOR_PLATFORM}" STREQUAL "" AND ${CMAKE_SYSTEM_PROCESSOR} MATCHES "ARM64")
>     set(ARCH_ARM64 1)
> elseif(${SIZEOF_VOID_PTR} STREQUAL "4")
>     set(ARCH_X86 1)
> elseif(${SIZEOF_VOID_PTR} STREQUAL "8")
>     set(ARCH_X64 1)
> else()
>     message(FATAL_ERROR "Unsupported architecture")
>     return()
> endif()
> ```

I think it's a good start, but not sure it's enough, given that `${CMAKE_GENERATOR_PLATFORM}` is only really set with the `-A` flag, which is only really available with the Visual Studio generators, and given that relying on `CMAKE_SYSTEM_PROCESSOR` might not be enough when cross compiling from ARM64 to something else.

Some suggest to use `uname -m`, but that would also just get you the system arch.

So for now, I have no better suggestion and would probably just go with the simplest approach and hope nobody tries cross-compiling on ARM64 (aka it's a problem for future us). 😅

Zombieschannel:
Done.

eXpl0it3r:

vittorioromeo:
How about `(${CMAKE_GENERATOR_PLATFORM} MATCHES "ARM64" OR ${CMAKE_SYSTEM_PROCESSOR} MATCHES "ARM64")`?

ChrisThrasher:
https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html

The documentation leads you to believe that this variable should represent the architecture of the target CPU as opposed to always corresponding to your host CPU architecture. Perhaps the `-A` flag doesn't constitute a "cross compile" according to CMake?

ChrisThrasher:
> How about `(${CMAKE_GENERATOR_PLATFORM} MATCHES "ARM64" OR ${CMAKE_SYSTEM_PROCESSOR} MATCHES "ARM64")`?

Would this fail when trying to compile for x64 on an ARM64 machine?

vittorioromeo:
> > How about `(${CMAKE_GENERATOR_PLATFORM} MATCHES "ARM64" OR ${CMAKE_SYSTEM_PROCESSOR} MATCHES "ARM64")`?
> 
> Would this fail when trying to compile for x64 on an ARM64 machine?

Looking at the docs (https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html), possibly. But it does seem that cross-compiling requires a bit more effort anyway, and that the relevant variables should be set in a toolchain file. 

I'm fine either way.

ChrisThrasher:
I asked some CMake gurus about this problem and was referred to [this CMake issue](https://gitlab.kitware.com/cmake/cmake/-/issues/15170) that explains the same issue we're having. This thread provides some useful context.

TL;DR CMake doesn't offer a single unified for checking the CPU architecture of the target platform. The situation here is messy.

Let's look at the scenarios we have to support that involve Windows Arm platforms in some capacity.
1. Intel host + `-A ARM64` (implies MSVC)
2. Intel host + non-MSVC cross compilation via CMake toolchain file
3. Arm native build with MSVC
4. Arm native build with non-MSVC
5. Arm host + `-A AMD64` (implies MSVC)

Am I missing anything? Once we have a good grasp on all the possible scenarios to support we'll have an easier time making sure we cover all cases.

eXpl0it3r:
- Arm host + `-A x64/x86`
- Arm host + non-MSVC cross compilation via CMake toolchain file

Not sure if cross compilation can only ever happen with just toolchain files, or if I can somehow pass a flag to GCC or use a different cross-compiling compiler.

eXpl0it3r (approved):

eXpl0it3r:
I've updated it to resolve the conflicts and squashed the commits into one.

Cross-compiled it on my Windows 11 x64 machine. Of course I couldn't run it.

eXpl0it3r:

PR #3113, opened 2024-06-19
Update macOS GitHub Actions Image

eXpl0it3r:
## Description

The `macos-11` image is deprecated and will be removed towards the end of the month, as such I've updated the image to `macos-12` as we have it on `master`.

Additionally, I add macOS arm64 builds, using the new `macos-14` images.

And finally, I silenced the GitHub Action warnings, by update the checkout action to v4.

## How to test this PR?

CI passes


ChrisThrasher (approved):

eXpl0it3r:
Let's see if trying to backport 741fe219da8a04acc9a84098d21648a452b77bd3 to 2.6.x would work

ChrisThrasher:
#2359 claims that its changes require "modern CMake". I'm not sure exactly what version was required, but I'm now wondering if we accidentally merged some changes that mean our minimum CMake version is no longer correct.

eXpl0it3r:
I checked the used functions and they all seem to be part of 3.7

PR #3114, opened 2024-06-21
Fixed cache state not correctly being ignored when switching between RenderTargets within a single context

### LINKED ISSUE #3107: sf::StencilBuffer not applied when drawn from RenderTexture to RenderWindow
### Prerequisite Checklist

### Describe your issue here

sf::StencilBuffer appears to not work when drawn onto an sf::RenderTexture which is then drawn to an sf::RenderWindow. 

It behaves correctly when drawn directly onto an sf::RenderWindow. The issue is strictly sf::RenderTexture -> sf::RenderWindow.

It behaves correctly when drawn onto an sf::RenderTexture which is then copied to an sf::Image which explains why we are not getting failing tests.

I don't understand all the intricacies of SFML's OpenGL implementation (or OpenGL for that matter) so I have limited ability to dive into this. I did find that the issue is fixed if we force a resetGLStates() first thing within RenderTarget::setupDraw(bool useVertexCache, const RenderStates& states).

https://github.com/SFML/SFML/blob/master/src/SFML/Graphics/RenderTarget.cpp#L833

To me this implies that the issue may be with caching the state, but this is just a shot in the dark.

### Your Environment

- OS / distro / window manager: Linux / Fedora 40 / Gnome 46.2
- SFML version: The latest master, d99d673330fa1b3d37d3dc09a147c6ad45df2c1b
- Compiler / toolchain: gcc 14.1.1

### Steps to reproduce

This is the [official StencilBuffer example](https://github.com/SFML/SFML/blob/master/examples/stencil/Stencil.cpp) minimally modified to draw onto an intermediate RenderTexture

```cpp
#include <SFML/Graphics.hpp>
#include <cstdlib>

int main()
{
    auto contextSettings = sf::ContextSettings{0 /* depthBits */, 8 /* stencilBits */};

    sf::RenderWindow window(sf::VideoMode({600, 600}),
                            "SFML Stencil",
                            sf::Style::Titlebar | sf::Style::Close,
                            sf::State::Windowed,
                            contextSettings);

    sf::RenderTexture renderTexture = sf::RenderTexture::create(sf::Vector2u(600, 600), contextSettings).value();

    sf::RectangleShape red({500, 50});
    red.setFillColor(sf::Color::Red);
    red.setPosition({270, 70});
    red.setRotation(sf::degrees(60));

    sf::RectangleShape green({500, 50});
    green.setFillColor(sf::Color::Green);
    green.setPosition({370, 100});
    green.setRotation(sf::degrees(120));

    sf::RectangleShape blue({500, 50});
    blue.setFillColor(sf::Color::Blue);
    blue.setPosition({550, 470});
    blue.setRotation(sf::degrees(180));

    while (window.isOpen())
    {
        {
            {
                window.close();
                break;
            }
        }

    	window.clear(sf::Color::Black, 0);
    	renderTexture.clear(sf::Color::Black, 0);

        renderTexture.draw(red,
                    sf::StencilMode{sf::StencilComparison::Always, sf::StencilUpdateOperation::Replace, 3, ~0u, false});

        renderTexture.draw(green,
                    sf::StencilMode{sf::StencilComparison::Always, sf::StencilUpdateOperation::Replace, 1, ~0u, false});

        renderTexture.draw(blue,
                    sf::StencilMode{sf::StencilComparison::Greater, sf::StencilUpdateOperation::Replace, 2, ~0u, false});

        renderTexture.display();
        window.draw(sf::Sprite(renderTexture.getTexture()));
        window.display();
    }

    return EXIT_SUCCESS;
}
```

### Expected behavior

expected

### Actual behavior

actual

binary1248:
Title.

Due to historic reasons, each `RenderTarget` has a state cache it uses to attempt to reduce unnecessary OpenGL function calls. This worked out in the days when each `RenderTarget` lived in its own separate context. Due to the high overhead of context switching, changes were made so that `RenderTarget`s could share a single context in order to save on the context switching. The original per-`RenderTarget` caching strategy was retained.

In order to fix cache synchronization issues, when we detect that we have switched from one `RenderTarget` to another within the same context, the cache is disabled for the first draw call via `m_cache.enabled = false`. This will ensure that all relevant states are set and their cache state synchronized. At the end of the draw call, the cache for the current `RenderTarget` is re-enabled via `m_cache.enabled = true` so that we don't have to go through setting all states again in subsequent draw calls _as long as we don't switch `RenderTarget`s_.

When adding scissor and stencil functionality, this synchronization mechanism was overlooked.

This fix just adds `!m_cache.enable` checks to the scissor and stencil apply functions in order to be able to force the cache state to be synchronized for them as well.

Fixes #3107

@danieljpetersen can test to verify this fixes the issue.

### DIFF
--- a/src/SFML/Graphics/RenderTarget.cpp
+++ b/src/SFML/Graphics/RenderTarget.cpp
@@ -662,7 +662,7 @@ void RenderTarget::applyCurrentView()
     // Set the scissor rectangle and enable/disable scissor testing
     if (m_view.getScissor() == FloatRect({0, 0}, {1, 1}))
     {
-        if (m_cache.scissorEnabled)
+        if (!m_cache.enable || m_cache.scissorEnabled)
         {
             glCheck(glDisable(GL_SCISSOR_TEST));
             m_cache.scissorEnabled = false;
@@ -674,7 +674,7 @@ void RenderTarget::applyCurrentView()
         const int     scissorTop   = static_cast<int>(getSize().y) - (pixelScissor.position.y + pixelScissor.size.y);
         glCheck(glScissor(pixelScissor.position.x, scissorTop, pixelScissor.size.x, pixelScissor.size.y));
 
-        if (!m_cache.scissorEnabled)
+        if (!m_cache.enable || !m_cache.scissorEnabled)
         {
             glCheck(glEnable(GL_SCISSOR_TEST));
             m_cache.scissorEnabled = true;
@@ -754,7 +754,7 @@ void RenderTarget::applyStencilMode(const StencilMode& mode)
     // Fast path if we have a default (disabled) stencil mode
     if (mode == StencilMode())
     {
-        if (m_cache.stencilEnabled)
+        if (!m_cache.enable || m_cache.stencilEnabled)
         {
             glCheck(glDisable(GL_STENCIL_TEST));
             glCheck(glColorMask(GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE));
@@ -765,7 +765,7 @@ void RenderTarget::applyStencilMode(const StencilMode& mode)
     else
     {
         // Apply the stencil mode
-        if (!m_cache.stencilEnabled)
+        if (!m_cache.enable || !m_cache.stencilEnabled)
             glCheck(glEnable(GL_STENCIL_TEST));
 
         glCheck(glStencilOp(GL_KEEP,


Bambo-Borris (approved):
Reran the sample code that @danieljpetersen provided on the original issue, I was able to reproduce it previously but with this fix I can no longer reproduce the issue. Solution also seems sensible to me.

ChrisThrasher (approved):

danieljpetersen:
This does indeed fix the issue. Thank you!

vittorioromeo (approved):

PR #3116, opened 2024-06-21
Allow Sprite to share ownership of underlying Texture data.

binary1248:
Title.

This change allows the underlying `sf::Texture` data to be shared between the user `sf::Texture` object and the `sf::Sprite`s it is attached to. In the case the user relinquishes ownership of their `sf::Texture` object, the underlying data will still be available to any `sf::Sprite` objects that it is attached to, avoiding the infamous "white square problem" without any public API changes.

Test with any code that causes the "white square problem" to manifest itself, e.g.:
```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode({200, 200}), "Test");

    auto texture = sf::Texture::loadFromImage(sf::Image({100, 100}, sf::Color::Red));
    sf::Sprite sprite(*texture);
    texture.reset();

    while (window.isOpen())
    {
        {
                return 0;
        }

        window.clear();
        window.draw(sprite);
        window.display();
    }
}
```

### DIFF
--- a/include/SFML/Graphics/Sprite.hpp
+++ b/include/SFML/Graphics/Sprite.hpp
@@ -32,6 +32,7 @@
 #include <SFML/Graphics/Drawable.hpp>
 #include <SFML/Graphics/Rect.hpp>
 #include <SFML/Graphics/RenderStates.hpp>
+#include <SFML/Graphics/Texture.hpp>
 #include <SFML/Graphics/Transformable.hpp>
 #include <SFML/Graphics/Vertex.hpp>
 
@@ -40,8 +41,6 @@
 
 namespace sf
 {
-class Texture;
-
 ////////////////////////////////////////////////////////////
 /// \brief Drawable representation of a texture, with its
 ///        own transformations, color, etc.
@@ -220,7 +219,7 @@ private:
     // Member data
     ////////////////////////////////////////////////////////////
     std::array<Vertex, 4> m_vertices;    //!< Vertices defining the sprite's geometry
-    const Texture*        m_texture;     //!< Texture of the sprite
+    Texture               m_texture;     //!< Texture of the sprite
     IntRect               m_textureRect; //!< Rectangle defining the area of the source texture to display
 };
 
--- a/include/SFML/Graphics/Texture.hpp
+++ b/include/SFML/Graphics/Texture.hpp
@@ -37,6 +37,7 @@
 #include <SFML/System/Vector2.hpp>
 
 #include <filesystem>
+#include <memory>
 #include <optional>
 
 #include <cstddef>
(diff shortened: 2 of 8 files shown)


vittorioromeo:
Thank you for experimenting with solving the sprite lifetime issues -- I believe it's a problem that requires some attention and I appreciate you putting some effort into it.

---

I am not in favour of this particular direction, as previously discussed in https://github.com/SFML/SFML/issues/2385 (specifically, see https://github.com/SFML/SFML/issues/2385#issuecomment-1925372755). I don't like the idea of SFML forcing a particular ownership model on its users -- a texture should just be a texture, and if someone wants to share its ownership through run-time reference counting they can always use `std::shared_ptr<sf::Texture>`.

If we make `sf::Texture` internally use a `std::shared_ptr`, then we're forcing a particular ownership model on every users, even those who only need basic scope-based lifetimes (simplest, and most efficient), or `std::unique_ptr`-based unique ownership (simpler, and more efficient shared ownership).

By storing the `sf::Texture` by value inside `sf::Sprite`, we are making `sf::Sprite` a heavyweight object, as it now contains a `std::shared_ptr`. Applications/games that do not need a shared ownership model based on atomic operations (meant for non-deterministic lifetimes in multithreaded contexts) will now be forced to use such an ownership model. Copying a sprite will be a much heavier operation compared to before. Similarly, creating a temporary `sf::Sprite` on the spot to render something will also be a much heavier operation compared to before.

Few other minor drawbacks of this change, for completeness:

- `Texture.hpp` becomes a physical dependency of `Sprite.hpp`.
- `<memory>` becomes a physical dependency of `Texture.hpp`.
- `sf::Texture` and `sf::Sprite` become more coupled together due to friendship.
- Any access to a `sf::Texture`'s data members requires one extra indirection through `std::shared_ptr`'s control block.

---

In general, it would be very difficult for me to accept this PR. 

I am still convinced that this is all a self-inflicted problem due to an unhealthy overuse of OOP -- the existence of `sf::Sprite` itself is just surface area for footguns. 

I re-propose my PRs #3072 and #3080, which solve the problem at design-time by completely eliminating the possibility of the lifetime issue of arising in the first place, without any extra overhead or imposing ownership model constraints on our users. 

Making "illegal states unrepresentable" at compile-time should be our direction, as we have done with our `std::optional`-based factory APIs. Adding shared ownership to work around a self-inflicted lifetime issue is something I will very likely never be happy about.

I would also like to point your attention to #3097, which introduces a lightweight debug-only lifetime tracking mechanism that does not impact any API, does not introduce any run-time or memory overhead in release mode (or when turned off), and prevents lifetime issues not only for `sf::Sprite` but for all SFML types that can have the same issue, presenting users with a helpful message that educates them on the concepts of lifetime and ownership without sweeping them under a rug.

binary1248:
Closing in favour of #3122.

PR #3117, opened 2024-06-21
Refactor `sf::Text` geometry update code

vittorioromeo:
Pure refactoring change that does not change the behavior of `sf::Text` whatsoever. These are some improvements I made when researching the performance benefit of merging outline and fill vertices into a single vertex array (and single draw call). Those benefits still need to be measured, but I wanted to capture the refactoring work in a separate PR regardless of whether those rendering changes are going to be made.

### DIFF
--- a/include/SFML/Graphics/Text.hpp
+++ b/include/SFML/Graphics/Text.hpp
@@ -31,14 +31,15 @@
 
 #include <SFML/Graphics/Color.hpp>
 #include <SFML/Graphics/Drawable.hpp>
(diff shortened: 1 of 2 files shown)


vittorioromeo:

vittorioromeo:
Just to be clear, this PR is independent from #3118 -- I've changed `sf::VertexArray` to `std::vector<sf::Vertex>` here to benefit from multiple-element insertion via `std::vector::insert(iterator, std::initializer_list)`.

Issue #3098, opened 2024-06-16
Enable `libstdc++` assertions and source fortification in our CI when applicable


vittorioromeo:
### Prerequisite Checklist

- [X] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [X] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [X] I understand [SFML's scope](https://www.sfml-dev.org/contribute.php#general-considerations) and believe the feature fits it

### Describe your feature request here

When building with `libstdc++`, there are two preprocessor definitions that can detect misuses of the standard library *(i.e., precondition violations)* and buffer overruns: 

- `-D_GLIBCXX_ASSERTIONS=1`
- `-D_FORTIFY_SOURCE=3`

There is also a flag called `-fhardened` in GCC that enables those, but it also enables many more checks that I think are unnecessary in our case -- see [*"GCC Preparing To Introduce "-fhardened" Security Hardening Option"*](https://www.phoronix.com/news/GCC-fhardened-Hardening-Option).

Furthermore, there is a preprocessor define called `-D_GLIBCXX_DEBUG` that implies `-D_GLIBCXX_ASSERTIONS`, but it adds heavyweight checks that can considerably slow down a program -- see [https://gcc.gnu.org/wiki/LibstdcxxDebugMode](https://gcc.gnu.org/wiki/LibstdcxxDebugMode).

`libc++` has something similar -- see [`_LIBCPP_DEBUG` Macro](https://releases.llvm.org/8.0.0/projects/libcxx/docs/DesignDocs/DebugMode.html#libcpp-debug-macro).

I'm not sure what the best way of enabling this would be -- we could add a `sfml_add_option` for the two preprocessor defines and enable it in the CI, I guess...

### Use Cases

Catching more classes of bugs early in our development process.

PR #3119, opened 2024-06-21
Enable standard library assertions in CI

vittorioromeo:
Addresses #3098. I've chosen flags that enable unexpensive assertions (e.g. assert when dereferencing a smart pointer or `std::optional`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,7 @@ jobs:
         - platform: { name: Windows MinGW, os: windows-2022 }
           config: { name: Static Standard Libraries, flags: -GNinja -DSFML_USE_MESA3D=TRUE -DCMAKE_CXX_COMPILER=g++ -DSFML_USE_STATIC_STD_LIBS=TRUE }
         - platform: { name: Windows MinGW, os: windows-2022 }
-          config: { name: Static with PCH (GCC), flags: -GNinja -DSFML_USE_MESA3D=TRUE -DCMAKE_CXX_COMPILER=g++ -DBUILD_SHARED_LIBS=FALSE -DSFML_ENABLE_PCH=1 }
+          config: { name: Static with PCH (GCC), flags: -GNinja -DSFML_USE_MESA3D=TRUE -DCMAKE_CXX_COMPILER=g++ -DBUILD_SHARED_LIBS=FALSE -DSFML_ENABLE_PCH=1 -DSFML_ENABLE_STDLIB_ASSERTIONS=FALSE } # disabling stdlib assertions due to false positive
         - platform: { name: macOS, os: macos-12 }
           config: { name: Frameworks, flags: -GNinja -DSFML_BUILD_FRAMEWORKS=TRUE -DBUILD_SHARED_LIBS=TRUE }
         - platform: { name: macOS , os: macos-12 }
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -204,6 +204,13 @@ if(SFML_ENABLE_SANITIZERS)
     string(APPEND CMAKE_CXX_FLAGS " -fno-omit-frame-pointer -fno-sanitize-recover=all -fsanitize=undefined")
 endif()
 
+option(SFML_ENABLE_STDLIB_ASSERTIONS "Enable standard library assertions" OFF)
+if(SFML_ENABLE_STDLIB_ASSERTIONS)
+    # intentionally using `add_definitions` here to propagate defines to subdirectories
+    add_definitions(-D_GLIBCXX_ASSERTIONS=1)                                   # see https://gcc.gnu.org/wiki/LibstdcxxDebugMode
+    add_definitions(-D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_EXTENSIVE) # see https://libcxx.llvm.org/Hardening.html
+endif()
+
 # set the output directory for SFML DLLs and executables
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
 
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -10,6 +10,7 @@
         "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
         "SFML_BUILD_EXAMPLES": "ON",
         "SFML_BUILD_TEST_SUITE": "ON",
+        "SFML_ENABLE_STDLIB_ASSERTIONS": "ON",
         "SFML_WARNINGS_AS_ERRORS": "ON"
       }
     }


kimci86:
Can't we determine which definition is needed with our CMake scripts depending on the platform and compiler? Then there could be a single option instead of three.

vittorioromeo:
> Can't we determine which definition is needed with our CMake scripts depending on the platform and compiler? Then there could be a single option instead of three.

We'd have to detect the standard library being used, not the platform or compiler. You could e.g. compile with GCC on a MacOS using either libstdc++ or libc++.

I'm not sure how to reliably detect the standard library with CMake

Alternatively, we could just define all of them at once with a generic `SFML_ENABLE_STDLIB_ASSERTIONS` option. It doesn't hurt to have unused defines. 

What do you think? @eXpl0it3r

vittorioromeo:
BTW, here's an example of an assertion that fires after intentionally adding a bad `std::optional` dereference in `sf::Image` as a test:

```
C:/msys64/ucrt64/include/c++/14.1.0/optional:475: _Tp &std::_Optional_base_impl<sf::Image, std::_Optional_base<sf::Image>>::_M_get() [_Tp = sf::Image, _Dp = std::_Optional_base<sf::Image>]: Assertion 'this->_M_is_engaged()' failed.
```

eXpl0it3r:
CMake magic is @ChrisThrasher's specialty 😄  
Not sure what CMake offers for detection.

But I'm certain he'd also prefer one option (if any at all).

binary1248:
> CMake magic is @ChrisThrasher's specialty 😄

😮

Combine [`iso646`](https://en.cppreference.com/w/cpp/header/ciso646) with CMake's `try_compile` to see what headers get consumed.

ChrisThrasher:
> > Can't we determine which definition is needed with our CMake scripts depending on the platform and compiler? Then there could be a single option instead of three.
> 
> We'd have to detect the standard library being used, not the platform or compiler. You could e.g. compile with GCC on a MacOS using either libstdc++ or libc++.
> 
> I'm not sure how to reliably detect the standard library with CMake
> 
> Alternatively, we could just define all of them at once with a generic `SFML_ENABLE_STDLIB_ASSERTIONS` option. It doesn't hurt to have unused defines.
> 
> Basically, this:
> 
> ```cmake
> option(SFML_ENABLE_STDLIB_ASSERTIONS "Enable standard library assertions" OFF)
> if(SFML_ENABLE_STDLIB_ASSERTIONS)
>     add_definitions(-D_GLIBCXX_ASSERTIONS=1) # see https://gcc.gnu.org/wiki/LibstdcxxDebugMode
>     add_definitions(-D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_EXTENSIVE) # see https://libcxx.llvm.org/Hardening.html
>     add_definitions(-D_DEBUG=1) # see https://learn.microsoft.com/en-us/cpp/c-runtime-library/debug?view=msvc-170
> endif()
> ```
> 
> What do you think? @eXpl0it3r

I'd much prefer one CMake option that enables all 3 at once and then amend our `dev` preset to enable this by default so that the CI scripts remain unchanged.

vittorioromeo:
> I'd much prefer one CMake option that enables all 3 at once and then amend our `dev` preset to enable this single option so that the CI scripts remain unchanged.

@ChrisThrasher 

This is a much much better solution! I've made the change you requested. For now, I removed `_DEBUG` for MSVC as it was causing linker errors, and I need to explore the consequences of enabling the full-fledged MSVC debug mode prior to doing so.

I would still recommend accepting this change for libstdc++ and libc++ to benefit from the assertions ASAP, and add MSVC support (if reasonable) in a later PR.

.github/workflows/ci.yml
```diff
@@ -63,7 +63,7 @@ jobs:
         - platform: { name: Windows MinGW, os: windows-2022 }
           config: { name: Static Standard Libraries, flags: -GNinja -DSFML_USE_MESA3D=TRUE -DCMAKE_CXX_COMPILER=g++ -DSFML_USE_STATIC_STD_LIBS=TRUE }
         - platform: { name: Windows MinGW, os: windows-2022 }
-          config: { name: Static with PCH (GCC), flags: -GNinja -DSFML_USE_MESA3D=TRUE -DCMAKE_CXX_COMPILER=g++ -DBUILD_SHARED_LIBS=FALSE -DSFML_ENABLE_PCH=1 }
+          config: { name: Static with PCH (GCC), flags: -GNinja -DSFML_USE_MESA3D=TRUE -DCMAKE_CXX_COMPILER=g++ -DBUILD_SHARED_LIBS=FALSE -DSFML_ENABLE_PCH=1 -SFML_ENABLE_STDLIB_ASSERTIONS=FALSE } # disabling stdlib assertions due to false positive
```
ChrisThrasher, line 66: What's the false positive?

CMakeLists.txt
```diff
@@ -204,6 +204,12 @@ if(SFML_ENABLE_SANITIZERS)
     string(APPEND CMAKE_CXX_FLAGS " -fno-omit-frame-pointer -fno-sanitize-recover=all -fsanitize=undefined")
 endif()
 
+option(SFML_ENABLE_STDLIB_ASSERTIONS "Enable standard library assertions" OFF)
+if(SFML_ENABLE_STDLIB_ASSERTIONS)
+    add_definitions(-D_GLIBCXX_ASSERTIONS=1)                                   # see https://gcc.gnu.org/wiki/LibstdcxxDebugMode
+    add_definitions(-D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_EXTENSIVE) # see https://libcxx.llvm.org/Hardening.html
```
ChrisThrasher, line 210:
```suggestion
    add_compile_definitions(_GLIBCXX_ASSERTIONS=1)                                   # see https://gcc.gnu.org/wiki/LibstdcxxDebugMode
    add_compile_definitions(_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_EXTENSIVE) # see https://libcxx.llvm.org/Hardening.html
```
CMake recommends using `add_compile_definitions` in lieu of `add_definitions`.

Source: https://cmake.org/cmake/help/latest/command/add_definitions.html

CMakePresets.json
```diff
@@ -10,7 +10,8 @@
         "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
         "SFML_BUILD_EXAMPLES": "ON",
         "SFML_BUILD_TEST_SUITE": "ON",
-        "SFML_WARNINGS_AS_ERRORS": "ON"
+        "SFML_WARNINGS_AS_ERRORS": "ON",
+        "SFML_ENABLE_STDLIB_ASSERTIONS": "ON"
```
ChrisThrasher, line 14:
```suggestion
        "SFML_ENABLE_STDLIB_ASSERTIONS": "ON",
        "SFML_WARNINGS_AS_ERRORS": "ON"
```
I'd like to keep these in alphabetical order if you don't mind

ChrisThrasher (changes requested):
LGTM pending a few minor comments

.github/workflows/ci.yml
vittorioromeo: This one: https://github.com/SFML/SFML/actions/runs/9629381587/job/26558744647#step:17:110

It looks like either complete bogus to me, or something inside the standard library implementation that we cannot really fix. 

Regardless, I cannot reproduce locally with a newer version of MinGW.

CMakeLists.txt
```diff
@@ -204,6 +204,12 @@ if(SFML_ENABLE_SANITIZERS)
     string(APPEND CMAKE_CXX_FLAGS " -fno-omit-frame-pointer -fno-sanitize-recover=all -fsanitize=undefined")
 endif()
 
+option(SFML_ENABLE_STDLIB_ASSERTIONS "Enable standard library assertions" OFF)
+if(SFML_ENABLE_STDLIB_ASSERTIONS)
+    add_compile_definitions(_GLIBCXX_ASSERTIONS=1)                                   # see https://gcc.gnu.org/wiki/LibstdcxxDebugMode
+    add_compile_definitions(_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_EXTENSIVE) # see https://libcxx.llvm.org/Hardening.html
```
vittorioromeo, line 210: Oh wait @ChrisThrasher, this apparently does not propagate to subdirectories. Should I revert to `add_definitions`?

ChrisThrasher (approved):

CMakeLists.txt
ChrisThrasher: Oh really? That's no good. If true then I suppose we have to revert this.

vittorioromeo: Reverted and added a comment.

vittorioromeo:

.github/workflows/ci.yml
ChrisThrasher, line 66:
```suggestion
          config: { name: Static with PCH (GCC), flags: -GNinja -DSFML_USE_MESA3D=TRUE -DCMAKE_CXX_COMPILER=g++ -DBUILD_SHARED_LIBS=FALSE -DSFML_ENABLE_PCH=1 -DSFML_ENABLE_STDLIB_ASSERTIONS=FALSE } # disabling stdlib assertions due to false positive
```

ChrisThrasher:

Issue #3120, opened 2024-06-22
We should bring back default constructors


danieljpetersen:
### Prerequisite Checklist

- [X] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [X] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [X] I understand [SFML's scope](https://www.sfml-dev.org/contribute.php#general-considerations) and believe the feature fits it

### Describe your feature request here

I wanted to share my thoughts on using some of the recent changes with SFML 3. There are positives to the changes and I know a lot of work has gone into it, but I'm going to focus on why I think these changes make the experience of using SFML worse.

Am I correct that if we want to store sfml objects on classes, our options are
	* wrap it in a smart pointer
	* store it as an optional
	* manually manage memory via new and delete

I believe this may backfire. A large amount of people are going to remedy the (confusing, potentially hundreds of lines of dense repetitive compile errors) by changing the type to a raw pointer. I believe the bug class we're attempting to solve is something that people who understand the lifetime model behind things like textures and sprites are unlikely to make. I think it's possible that, in our effort to reduce bugs among folks who are new, we're going to push them down a path which is likely to cause even more bugs than the amount that currently exist due to misuse. I know that when I was new I would have changed it to use a raw pointer. The way I resolved this today was by using smart pointers, but this actually caused a handful of runtime crashes that I had to track down due to not having called make_unique as I overlooked a few.

I think we're attempting to solve something which is a problem on paper and not one in practice. I think we get a more miserable API in exchange. In order to get Binary's stencil buffer fix I updated my SFML branch from prior to the event change to the latest master. Here's an example of the type of thing which I think is going to trip people up.

I have an unordered_map<string, sf::Image>. I thought for a moment this would work
```
asyncLoadedImages[workOrder.path] = std::move(img.value());
```
But the solution is using insert_or_assign, so we apparently can't use bracket operators here anymore. A small trap here is to make sure that you're not using insert and not using emplace because that's different behavior than what we originally had.
``` 
asyncLoadedImages.insert_or_assign(workOrder.path, std::move(img.value()));
```
[...]

PR #3121, opened 2024-06-23
Improve `sf::Event` documentation and parameter naming

vittorioromeo:
Inspired by https://github.com/SFML/SFML/issues/3120. This PR improves the documentation and parameter naming of `sf::Event` in a few ways:

- Consistently uses the "subtype" terminology to talk about event variant alternatives
- Renames `T` template parameters to a more descriptive `EventSubType` and documents them
- Uses backticks for inline code in Doxygen comments
- Minor wording improvements

### DIFF
--- a/include/SFML/Window/Event.hpp
+++ b/include/SFML/Window/Event.hpp
@@ -49,7 +49,7 @@ class SFML_WINDOW_API Event
 {
 public:
     ////////////////////////////////////////////////////////////
-    /// \brief Closed event
+    /// \brief Closed event subtype
     ///
     ////////////////////////////////////////////////////////////
     struct Closed
@@ -57,7 +57,7 @@ public:
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Resized event
+    /// \brief Resized event subtype
     ///
     ////////////////////////////////////////////////////////////
     struct Resized
@@ -66,7 +66,7 @@ public:
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Lost focus event
+    /// \brief Lost focus event subtype
     ///
     ////////////////////////////////////////////////////////////
     struct FocusLost
@@ -74,7 +74,7 @@ public:
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Gained focus event
+    /// \brief Gained focus event subtype
     ///
     ////////////////////////////////////////////////////////////
     struct FocusGained
@@ -82,7 +82,7 @@ public:
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Text event
+    /// \brief Text event subtype
     ///
     ////////////////////////////////////////////////////////////
     struct TextEntered
@@ -91,7 +91,7 @@ public:
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief KeyChanged events
+    /// \brief KeyChanged event subtypes
     ///
     ////////////////////////////////////////////////////////////
     struct KeyChanged
@@ -111,7 +111,7 @@ public:
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Mouse wheel scrolled event
+    /// \brief Mouse wheel scrolled event subtype
     ///
     ////////////////////////////////////////////////////////////
     struct MouseWheelScrolled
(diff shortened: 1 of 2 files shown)


vittorioromeo:

include/SFML/Window/Event.hpp
```diff
@@ -277,42 +277,48 @@ class SFML_WINDOW_API Event
...
     ///
     ////////////////////////////////////////////////////////////
     Event() = default;
 
     ////////////////////////////////////////////////////////////
-    /// \brief Construct from a given sf::Event subtype
+    /// \brief Construct from a given `sf::Event` subtype
     ///
-    /// \param t Event subtype
+    /// \tparam EventSubType Type of event subtype used to construct the event
```
ChrisThrasher, line 288:
```suggestion
    /// \tparam EventSubtype Type of event subtype used to construct the event
```
`EventSubType` -> `EventSubtype`. "Subtype" is one word so I think the "t" in "type" does not deserve to be capitalized.

I'm also not in love with the "type of event subtype" phrasing. I think we can just say 
```
/// \tparam EventSubtype Event subtype used to construct the event
```

ChrisThrasher (changes requested):
I think cleaning this up is worthwhile. I've also been a little confused at how `sf::Event` is "an event" but also `sf::Event::Closed` is "an event". It's good to be more clear about the hierarchy at play here.

include/SFML/Window/Event.hpp
eXpl0it3r: As a template type it should be prefixes with T: TEventSubtype

vittorioromeo: We don't follow that convention anywhere else -- if you feel strongly about that naming convention I can apply it here and then create a separate PR to apply it to the rest of the library. What do you decide @eXpl0it3r?

vittorioromeo: @ChrisThrasher:

> EventSubType -> EventSubtype. "Subtype" is one word so I think the "t" in "type" does not deserve to be capitalized.

Good point, will fix.

> I'm also not in love with the "type of event subtype" phrasing. I think we can just say [...]

I'm not in love with it either, but I'd prefer keeping it. Consider

```cpp
/// \tparam EventSubtype Type of event subtype used to construct the event
///
/// \param eventSubtype Event subtype used to construct the event
```

Our meaning of "event subtype" here is not the type (for example) `sf::Event::Closed`, but it is the actual object of type `sf::Event::Closed`. So our meaning of "event subtype" is "object of type `sf::Event::*`. Therefore to differentiate `eventSubtype` (object) and `EventSubtype` (type) it is more correct to say *"type of event subtype"*, despite the cacophony.

ChrisThrasher (approved):

vittorioromeo:

include/SFML/Window/Event.hpp
eXpl0it3r: The only place where I found that we don't just use `T` or `U` is in the new `Time` API.

```
    template <typename Rep, typename Period>
    constexpr Time(const std::chrono::duration<Rep, Period>& duration);
```

Yes, I think it makes sense to prefix them with `T`. Also give `Rep` a more speaking name? I'd have to look up `std::chrono` just to understand what type is expected here.

include/SFML/Window/Event.inl
kimci86, line 1: There is also an occurrence of `isEventType` in the comment on line 25.

ChrisThrasher:

include/SFML/Window/Event.hpp
```diff
@@ -267,31 +267,37 @@ class SFML_WINDOW_API Event
     };
 
     ////////////////////////////////////////////////////////////
-    /// \brief Construct from a given sf::Event subtype
+    /// \brief Construct from a given `sf::Event` subtype
     ///
-    /// \param t Event subtype
+    /// \tparam TEventSubtype Type of event subtype used to construct the event
+    ///
+    /// \param eventSubtype Event subtype used to construct the event
```
kimci86, line 274: What about adding the word "instance" to clarify when we are referring to values instead of types?

kimci86 (approved):
LGTM besides a minor suggestion.

ChrisThrasher:

eXpl0it3r (approved):

PR #3129, opened 2024-06-24
Rename Font::loadFromFile into Font::openFromFile

kimci86:
Similar renaming for Font::loadFromMemory and Font::loadFromStream.
The goal is to better express the need to keep the source available, similar to Music::openFromFile for example.

I suggest this renaming while we can still make some breaking changes to the API.

What do you think?

### DIFF
--- a/examples/joystick/Joystick.cpp
+++ b/examples/joystick/Joystick.cpp
@@ -93,8 +93,8 @@ int main()
     sf::RenderWindow window(sf::VideoMode({400, 775}), "Joystick", sf::Style::Close);
     window.setVerticalSyncEnabled(true);
 
-    // Load the text font
-    const auto font = sf::Font::loadFromFile("resources/tuffy.ttf").value();
+    // Open the text font
+    const auto font = sf::Font::openFromFile("resources/tuffy.ttf").value();
 
     // Set up our string conversion parameters
     sstr.precision(2);
--- a/examples/raw_input/RawInput.cpp
+++ b/examples/raw_input/RawInput.cpp
@@ -18,8 +18,8 @@ int main()
     sf::RenderWindow window(sf::VideoMode({800u, 600u}), "SFML Raw Mouse Input", sf::Style::Titlebar | sf::Style::Close);
     window.setVerticalSyncEnabled(true);
 
-    // Load the application font and pass it to the Effect class
-    const auto font = sf::Font::loadFromFile("resources/tuffy.ttf").value();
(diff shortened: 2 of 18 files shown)


ChrisThrasher:
I've never totally understood the difference between `loadFromFile` and `openFromFile`. Perhaps @eXpl0it3r knows more about the history of these names?

eXpl0it3r:
It was used to indicate that the source needs to remain available. For example a SoundBuffer loads the whole file into memory, while the Music continuously streams the data (from disk or your own stream).

Font always used `loadFrom*` yet we actually require the file/stream to outlive the font.

As such, I believe it aligns more to use `openFrom*` here as well.

examples/joystick/Joystick.cpp
```diff
@@ -94,7 +94,7 @@ int main()
     window.setVerticalSyncEnabled(true);
 
     // Load the text font
```
kimci86, line 96: Update comment

examples/raw_input/RawInput.cpp
```diff
@@ -19,7 +19,7 @@ int main()
     window.setVerticalSyncEnabled(true);
 
     // Load the application font and pass it to the Effect class
```
kimci86, line 21: Update comment

ChrisThrasher:
> It was used to indicate that the source needs to remain available. For example a SoundBuffer loads the whole file into memory, while the Music continuously streams the data (from disk or your own stream).
> 
> Font always used `loadFrom*` yet we actually require the file/stream to outlive the font.
> 
> As such, I believe it aligns more to use `openFrom*` here as well.

To make sure I understand...

`openFromX` -> stream content from source, therefore the source must outlive the class using it

`loadFromX` -> load entire content into memory, therefore the source **does not** need to outlive the class using it

ChrisThrasher:
@kimci86 Did you happen to check if other classes violate this convention?

ChrisThrasher:

kimci86:
> Did you happen to check if other classes violate this convention?

Doing a quick search search over the API:
- with loadFrom* : SoundBuffer, _Font_, Image, Shader, Texture, Cursor
- with openFrom*:  InputSoundFile, Music, OutputSoundFile

I think Font is the only odd one.

eXpl0it3r:

ChrisThrasher (approved):

eXpl0it3r (approved):

eXpl0it3r:
Thank you for aligning this inconsistency!

PR #3131, opened 2024-06-25
Fix SFML_USE_STATIC_STD_LIBS behaviour with newer cmake (2.6.x)

JonnyPtn:
CMAKE_CXX_FLAGS does not always contain the `/MD` flags so using a string replace doesn't have any effect in that case.

Instead, remove the default flag if it's present, and always add the correct flag


eXpl0it3r (approved):

eXpl0it3r:

Lorenzooone:
Confirmed working on my end!

PR #3132, opened 2024-06-25
Backmerge 2.6.x to master

eXpl0it3r:
Getting all the changes from the `2.6.x` to master


kimci86:
The OpenAL binaries can be ignored as we use miniaudio instead on master branch.

ChrisThrasher (approved):

src/SFML/Window/Win32/ClipboardImpl.cpp
```diff
@@ -35,6 +35,26 @@
...
+                        error,
+                        0,
+                        reinterpret_cast<PTCHAR>(&buffer),
+                        0,
+                        nullptr) == 0)
+        return "Unknown error.";
+
+    sf::String message = buffer;
+    LocalFree(buffer);
+    return message.toAnsiString();
```
ChrisThrasher, line 55: We now have 3 nearly identical implementations of this function in src/Window/Win32/ClipboardImpl.cpp, src/Window/Win32/JoystickImpl.cpp, and src/Window/Win32/WglContext.cpp. I propose we add a new src/Window/Win32/Utils.hpp file where we can this function. This will reduce code duplication and avoid these unity builds failures we're getting. There is already precedent for this. We already have a src/Window/Unix/Utils.hpp header for some X11 utilities.

ChrisThrasher:
I pushed my CI fixes as a new commit so they're separate from the work you've already done in case we want change our minds on the solution I picked.

In SFML 2, our CI pipeline does do unity builds. As a matter of coincidence, our three separate definitions of `getErrorString` don't get compiled in the same translation unit so the unity build never failed. That's why the v2 CI pipeline has always passed. Something about merging into `master` caused CMake to rearrange which files get compiled together during the unity build causing this error to manifest.

No matter what, we need to collapse these three nearly identical functions into one. This is required to fix our unity build errors in v3 plus it's just good software engineering. Having two identical functions is potentially acceptable but I draw the line at 3 identical functions. The only question is whether we bother also fixing this in v2. We can get away with not fixing it in v2 if we want. Perhaps we just wrap some extra namespaces around these functions as a short term fix in v2 while we do the proper fix in v3 as I already wrote.

src/SFML/Window/Win32/ClipboardImpl.cpp
```diff
@@ -45,21 +46,22 @@ String ClipboardImpl::getString()
 
     if (!IsClipboardFormatAvailable(CF_UNICODETEXT))
     {
-        err() << "Failed to get the clipboard data in Unicode format." << std::endl;
+        err() << "Failed to get the clipboard data in Unicode format: " << priv::getErrorString(GetLastError())
```
eXpl0it3r, line 49: `priv::` isn't needed here, because we're already in `sf::priv`

src/SFML/Window/Win32/Utils.hpp
```diff
@@ -0,0 +1,51 @@
...
+{
+    PTCHAR buffer = nullptr;
+    if (FormatMessage(FORMAT_MESSAGE_MAX_WIDTH_MASK | FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM,
+                      nullptr,
+                      error,
+                      0,
+                      reinterpret_cast<PTCHAR>(&buffer),
+                      0,
+                      nullptr) == 0)
+        return "Unknown error.";
```
eXpl0it3r, line 45: Due to the line breaking formatting of the if, I'd prefer a `{}` wrapping of this return statement, unless clang-tidy forbids this

src/SFML/Window/Win32/WglContext.cpp
```diff
@@ -234,7 +211,7 @@ bool WglContext::makeCurrent(bool current)
     if (wglMakeCurrent(m_deviceContext, current ? m_context : nullptr) == FALSE)
     {
         err() << "Failed to " << (current ? "activate" : "deactivate")
-              << " OpenGL context: " << getErrorString(GetLastError()).toAnsiString() << std::endl;
+              << " OpenGL context: " << priv::getErrorString(GetLastError()) << std::endl;
```
eXpl0it3r, line 214: We're already inside `sf::priv`

src/SFML/Window/Win32/Utils.hpp
```diff
@@ -0,0 +1,51 @@
...
+{
+inline std::string getErrorString(DWORD error)
+{
+    PTCHAR buffer = nullptr;
+    if (FormatMessage(FORMAT_MESSAGE_MAX_WIDTH_MASK | FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM,
+                      nullptr,
+                      error,
+                      0,
+                      reinterpret_cast<PTCHAR>(&buffer),
+                      0,
```
eXpl0it3r, line 43: 0 is fine here

> If the FORMAT_MESSAGE_ALLOCATE_BUFFER flag is not set, this parameter specifies the size of the output buffer, in TCHARs. If FORMAT_MESSAGE_ALLOCATE_BUFFER is set, this parameter specifies the minimum number of TCHARs to allocate for an output buffer.
> 
> The output buffer cannot be larger than 64K bytes.

We can let Windows figure out the necessary size of the buffer.

src/SFML/Window/Win32/Utils.hpp
```diff
@@ -0,0 +1,51 @@
...
+namespace sf::priv
+{
+inline std::string getErrorString(DWORD error)
+{
+    PTCHAR buffer = nullptr;
+    if (FormatMessage(FORMAT_MESSAGE_MAX_WIDTH_MASK | FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM,
+                      nullptr,
+                      error,
+                      0,
+                      reinterpret_cast<PTCHAR>(&buffer),
```
eXpl0it3r, line 42: The cast should be fine, but it's really not a great look for the WinAPI, so many different "types" for strings and it's all kind of convertible...

`PTCHAR` == `TCHAR*`  
`TCHAR` == `WCHAR` (if `UNICODE`) or `char`  
`LPTSTR` == `LPWSTR` (if `UNICODE`) or `LPSTR`  
`LPWSTR` == `WCHAR*`  
`LPSTR` == `CHAR*`  
`WCHAR` == `wchar_t`  
`CHAR` == `char`

So the param takes `wchar_t* or `char*` which can be either represented by `LPTSTR` or `PTCHAR`

src/SFML/Window/Win32/Utils.hpp
```diff
@@ -0,0 +1,51 @@
...
+////////////////////////////////////////////////////////////
+#include <SFML/System/String.hpp>
+#include <SFML/System/Win32/WindowsHeader.hpp>
+
+namespace sf::priv
+{
+inline std::string getErrorString(DWORD error)
+{
+    PTCHAR buffer = nullptr;
+    if (FormatMessage(FORMAT_MESSAGE_MAX_WIDTH_MASK | FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM,
```
eXpl0it3r, line 38: Since we're not using `printf` or similar, we don't need `FORMAT_MESSAGE_IGNORE_INSERTS`.

See an explanation here: https://devblogs.microsoft.com/oldnewthing/20071128-00/?p=24353

eXpl0it3r (commented):
Looks good to me

PR #3134, opened 2024-06-26
Added throwing constructor analogs of load/open factory functions.

binary1248:
Title.

Enables better integration with the STL without having to constantly unwrap `optional`s.

Also due to the fact that object memory is only allocated once and resources don't have to be moved back and forth (not only for the object itself but also within the constructor implementations), the throwing constructor variants might provide better runtime performance and cause less memory fragmentation.

```cpp
std::make_unique<sf::Texture>(...);
std::vector<sf::Texture>::emplace_back(...);
std::unordered_map<std::string, sf::Texture>::try_emplace(key, ...);
...
```

Also allows concise and better support for programmatic handling of errors, without having to rely on `err()` or other debug-only output.

```cpp
try
{
    texture = std::make_unique<sf::Texture>(...);
    imageMap.try_emplace("someKey", ...);
    ...
}
catch (const std::runtime_error& ex)
{
    std::cout << "Something went wrong during loading: " << ex.what() << std::endl;
}
```

If we add support for our own differentiated exceptions, the user will have finer grained control over how they might want to handle them.

Slightly related to #3120.
@danieljpetersen might be interested in this.

### DIFF
--- a/include/SFML/Audio/InputSoundFile.hpp
+++ b/include/SFML/Audio/InputSoundFile.hpp
@@ -53,6 +53,51 @@ class SFML_AUDIO_API InputSoundFile
 {
 public:
     ////////////////////////////////////////////////////////////
+    /// \brief Construct a sound file from the disk for reading
+    ///
+    /// The supported audio formats are: WAV (PCM only), OGG/Vorbis, FLAC, MP3.
+    /// The supported sample sizes for FLAC and WAV are 8, 16, 24 and 32 bit.
+    ///
+    /// Because of minimp3_ex limitation, for MP3 files with big (>16kb) APEv2 tag,
+    /// it may not be properly removed, tag data will be treated as MP3 data
+    /// and there is a low chance of garbage decoded at the end of file.
+    /// See also: https://github.com/lieff/minimp3
+    ///
+    /// \param filename Path of the sound file to load
+    ///
+    /// \throws std::runtime_error if opening the file was unsuccessful
+    ///
+    ////////////////////////////////////////////////////////////
+    InputSoundFile(const std::filesystem::path& filename);
+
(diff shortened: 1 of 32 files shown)


MarioLiebisch:
> If we add support for our own differentiated exceptions, the user will have finer grained control over how they might want to handle them.

If we add exceptions, I'd absolutely vote for using custom exceptions, not just standard ones.

And should it be possible to use a config value to exclude these constructors for builds without exceptions?

binary1248:
Disabling exceptions isn't technically supported even by the standard library (at least in the sense that disabling exceptions removes functionality). I don't think it makes sense to have a configuration option to remove throwing code. If the user decides to disable exceptions in their project then it is their responsibility to avoid causing anything to throw, not only when calling SFML functions but standard library functions as well.

danieljpetersen:
I appreciate the effort you've put into SFML over the years Binary. I don't think exception support would change my misgivings about lack of default constructors, but that's not to say I think exceptions are a bad idea. I hesitated posting since I didn't want to leave a negative comment (especially since not entirely related). I do think this will be appreciated by some folks.

trustytrojan:
Is this finally the time we make use of a C++ feature that's been in the language for years? I'm all in. Even before the `std::optional` mess that we have now, I've been wrapping SFML classes in my own that simply throw exceptions on failure. I really don't get the point of forcing C-style error-handling semantics down everyone's throat. This isn't C. Now that my project no longer compiles I think I'd rather spend my time contributing to the exception movement. Keep it up Binary.

binary1248:
Alternative including default constructors and old load/open functionality is provided in #3152.

ChrisThrasher:
Closing in favor of #3152

PR #3137, opened 2024-06-27
Use `[[nodiscard]]` in more places

ChrisThrasher:
## Description

This PR does two things. Primarily it adds `[[nodiscard]]` to many APIs (both public and private) where it would be unambiguously incorrect to ignore the return value. This makes it more difficult to write programs that misuse the API. GCC and Clang will warn about this by default so it's safe to assume once we add this all SFML 3 users will start potentially receiving these warnings. Whether or not those warnings are converted into errors is a matter of user preference. There are likely more APIs where `[[nodiscard]]` would be appropriate so we may add it in more places after this PR once a precedent has been established.

The second thing this PR does is use clang-tidy's [`modernize-use-nodiscard`](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-nodiscard.html) check to automatically flag APIs that ought to be marked with `[[nodiscard]]` but are not. This only applies to class member functions so this check will not touch free functions. We will have to manually audit the codebase to find free functions that need the `[[nodiscard]]` attribute. This is doable albeit rather labor intensive.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -54,7 +54,6 @@ Checks: >
   -misc-non-private-member-variables-in-classes,
   -modernize-avoid-c-arrays,
   -modernize-macro-to-enum,
-  -modernize-use-nodiscard,
   -modernize-use-trailing-return-type,
   -performance-avoid-endl,
   -performance-enum-size,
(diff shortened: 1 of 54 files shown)


binary1248 (approved):

ChrisThrasher:

PR #3138, opened 2024-06-27
Use more `[[nodiscard]]`

ChrisThrasher:
## Description

These functions were not flagged by clang-tidy. I searched for patterns like functions that start with `get` or `is` and are marked as `const`. Others were found just by reading various header files looking for more unmarked functions. As it turns out a _lot_ of our functions require using the return value. I think this is a good thing. I'd find it suspicious if we had a lot of functions with optional return values.

There are many, many private functions inside the src/ dir that could theoretically use `[[nodiscard]]` but I won't bother changing those since it would be a massive effort and offer relatively little marginal value. I plan on stopping the `[[nodiscard]]` work with the interface and therefore stop with this PR.

### DIFF
--- a/src/SFML/Window/FreeBSD/JoystickImpl.hpp
+++ b/src/SFML/Window/FreeBSD/JoystickImpl.hpp
@@ -85,7 +85,7 @@ public:
     /// \return Joystick capabilities
     ///
     ////////////////////////////////////////////////////////////
-    JoystickCaps getCapabilities() const;
+    [[nodiscard]] JoystickCaps getCapabilities() const;
 
     ////////////////////////////////////////////////////////////
     /// \brief Get the joystick identification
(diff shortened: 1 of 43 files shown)


src/SFML/Window/FreeBSD/JoystickImpl.hpp
```diff
@@ -85,15 +85,15 @@ class JoystickImpl
...
+    [[nodiscard]] JoystickCaps getCapabilities() const;
 
     ////////////////////////////////////////////////////////////
     /// \brief Get the joystick identification
     ///
     /// \return Joystick identification
     ///
     ////////////////////////////////////////////////////////////
-    Joystick::Identification getIdentification() const;
+    [[nodiscard]] Joystick::Identification getIdentification() const;
```
ChrisThrasher, line 96: If we ran clang-tidy on the FreeBSD, NetBSD, and OpenBSD codepaths then these changes would have been fixed in #3137 but we don't so I had to manually add them here.

binary1248 (approved):

Issue #3023, opened 2024-05-19
SFML Event Intermediate classes are not fully featured Events.


L0laapk3:
This issue is meant as a discussion starting point.

Currently, the Event API makes some attempts to introduce a class structure into the events. There are for example already `TouchChanged`, `MouseButtonChanged` event classes exposed.

However, these are not treated the same way as other events, and are as a result of that not very convenient to use. I would like to see this change.

Ideally, I would like the API to be able to do something like this:  
(Listed in comments are limitations of the current API as to why this is not possible right now in a neat way.)
```cpp
void main() {
  sf::Event baseEvent = window.pollEvent();
  // Problem 1: event.getIf<parentClass>() is not possible, you have to exhaustively try getIf on all the event types.
  // Problem 2: There are not enough event parent classes: `MouseButtonChanged` exists, `MousePosition` doesn't.
  if (auto mouseEvent = baseEvent.getIf<sf::Event::MousePosition>()) {
	auto* myWorldBlock = myWorld.getBlockAtLocation(mouseEvent.position);
	if (myWorldBlock)
		if (myWorldBlock->handleEvent(mouseEvent))
		  return true;
    
	// handle other events
  }
}

// Problem 3: Cannot be explicit about only accepting MousePosition event types
bool myWorldBlock::handleEvent(const sf::Event::MousePosition& event) {
	// Problem 4: `event.getIf()` cannot be called on the sub event classes
	if (auto& scrollEvent = event.getIf<sf::Event::MouseWheelScrolled>()) {
		// handle event
		return true;
	}
	// handle other mouse events
	return false;
}
```

To achieve this, I propose the following Event changes:
1. Add more Event parent class groups such as `Event::Mouse`, `Event::Joystick`, `Event::MousePosition`.
2. All Events inherit some common base class which provides `.is()`, `.getIf()`.

vittorioromeo:
My understanding is that the types `KeyChanged`, `MouseButtonChanged', `JoystickButtonChanged`, `JoystickChanged`, and `TouchChanged` are just implementation details to avoid repetition when writing the event fields. Is that true, @ChrisThrasher? If so, we can hide them under `priv` or (gasp) use a macro to avoid the repetition.

ChrisThrasher:
They’re meant to be used and are not something I want to hide.

vittorioromeo:
> They’re meant to be used and are not something I want to hide.

How can they be used?

ChrisThrasher:
[...]

Issue #2990, opened 2024-05-09
What are you doing with Event?


kelbon:
### Prerequisite Checklist

- [X] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [X] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [X] I am here to report an issue and not to just ask a question or look for help (use the [forum](https://en.sfml-dev.org/forums/index.php#c3) or [Discord](https://discord.gg/nr4X7Fh) instead)

### Describe your issue here

SFML was good until
https://github.com/SFML/SFML/pull/2766

Firstly. No one (literaly no one) had problems with some 'type safety', no one asks to make such great changes which is fully broke all existing code ever.

Problems:
-1. it "solves" problems which never existed
 0. it breaks all code
 2. It is unusable (if if if if if) and no way to make clear code with switch and field access. Code will be less readable (both for me and compiler) and less maintanable 
 3. it is not efficient.

Author says, that he does not want to add .visit because it will break code later with adding events. But he break all existing code now and it is unusable. I think real reason why he does not add .visit is because Rust does not have .visit and it is impossible to implement

REMOVE RUST FROM MY C++

### Your Environment

any environment

### Steps to reproduce

0. have some code with using SFML

### Expected behavior

Library is usefull

### Actual behavior

Library unusable

Bambo-Borris:
The API change is still in the works, it isn't its final form. There was also ample time to voice concerns and discuss it during the PR of #2766. The changes do solve problems of users being able to misuse the `sf::Event` union from 2.x days, many users enter the SFML Discord & forums with bugs and issues that trace back to misusing the `sf::Event` union. It is perfectly usable, but if you prefer the switch case there was an argument put forward by people about this during the PR, and since, and I've seen maintainers are more than open to discussion about this. Also master is the in development branch of SFML 3.0.0, which is a major version update that features breaking API changes. Which under sem ver is the only time SFML **can** undertake such changes. If you've been using master, you've been utilising an in development version which is still subject to change, expect breaking changes to come if you choose to use that branch.

[...]

PR #3139, opened 2024-06-28
Added event visitation support

### LINKED ISSUE #3014: Exploration of `sf::Event::visit` use cases

binary1248:
Title.

This change adds event visitation support to `sf::WindowBase` (not `sf::Event` itself). Instead of exposing the internals of `sf::Event` itself, the user can pass multiple types of callables to a new `handleEvents` method to do all the dirty work for them.

Yes... this is purely syntactic sugar built on top of the new optional-based API, but in my opinion this will sweeten up the experience of beginning users who would not have to familiarize themselves with `optional`s or `variant`s to have a type-safe way of handling events they are interested in. This might also be familiar to users coming from other programming languages where such patterns are more commonplace (e.g. Java etc.).

To quote @eXpl0it3r from #2766:

> I don't think, this will be the final form of the API for SFML 3, but let us collect some more feedback from people using it, by getting it into master.

For a demonstration of the multiple ways that users can choose from to handle events refer to the newly added Event Handling example.

By leveraging the type system and compile-time programming, experienced users will be able to fine-tune their event handling code to their likings.

Related to #3015.
Related to #2990.
Related to #3023.
Closes #3014.

@xparq, @kelbon and @L0laapk3 might be interested in this.


examples/event_handling/EventHandling.cpp
```diff
@@ -0,0 +1,224 @@
...
+#include <vector>
+
+
+namespace
+{
+std::string vec2ToString(const sf::Vector2i& vec2)
+{
+    return '(' + std::to_string(vec2.x) + ", " + std::to_string(vec2.y) + ')';
+};
+} // namespace
```
MarioLiebisch, line 15: Not really related to this PR, but I wouldn't be against finally adding simple things like string conversion to SFML classes like Vector2D or Vector3D (i.e. where it makes sense and does more than default `std::to_string()`.

examples/event_handling/EventHandling.cpp
```diff
@@ -0,0 +1,224 @@
...
+        auto handlerType = 0;
+
+        while (m_window.isOpen())
+        {
+            if (handlerType == 0)
+            {
+                // The "classical form" of event handling
+                // Poll/Wait for events in a loop and handle them
+                // individually based on their concrete type
+                while (const std::optional event = m_window.pollEvent())
```
MarioLiebisch, line 51: I haven't really used `std::optional`, so this is kind of a new user perspective: This feels weird without template arguments. And wouldn't it be closer to other variables here if it would just be `auto`?

examples/event_handling/EventHandling.cpp
```diff
@@ -0,0 +1,224 @@
...
+                        }
+                        else if constexpr (std::is_same_v<T, sf::Event::KeyPressed>)
+                        {
+                            m_log.emplace_back("2: Key Pressed: " + sf::Keyboard::getDescription(event.scancode));
+                        }
+                        else if constexpr (std::is_same_v<T, sf::Event::MouseMoved>)
+                        {
+                            m_log.emplace_back("2: Mouse Moved: " + vec2ToString(event.position));
+                        }
+                        if constexpr (std::is_same_v<T, sf::Event::MouseButtonPressed>)
```
MarioLiebisch, line 123: `else if`?

examples/event_handling/EventHandling.cpp
```diff
@@ -0,0 +1,224 @@
...
+    {
+        // This example is for demonstration purposes only
+        // All the following forms of event handling have equivalent behavior
+        // In your own code you should decide which form of event handling
+        // suits your needs best and use a single form of event handling
+        auto handlerType = 0;
+
+        while (m_window.isOpen())
+        {
+            if (handlerType == 0)
```
MarioLiebisch, line 46: Even though it's just a rotating number, I'd still use `enum` values to name the "magic numbers".

examples/event_handling/EventHandling.cpp
```diff
@@ -0,0 +1,224 @@
...
+                // When choosing this method a default "catch-all" handler must
+                // be available for unhandled events to be forwarded to
+                // If you don't want to provide an empty "catch-all" handler
+                // you will have to make sure (e.g. via if constexpr) that this
+                // lambda doesn't attempt to call a member function that doesn't exist
+                m_window.handleEvents([this](const auto& event) { handle(event); });
+            }
+
+            // Switch to next form of event handling
+            handlerType = (handlerType + 1) % 4;
```
MarioLiebisch, line 152: I expected some input switch to change implementation, which would also allow/invite experimentation and comparison from user side. With constant switching it's kind of random which handler type is active at any given time, making it way less predictable/reproducable.

examples/event_handling/EventHandling.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+            // Draw the contents of the log to the window
+            m_window.clear();
+
+            for (std::size_t i = 0; i < m_log.size(); ++i)
+            {
+                m_logText.setPosition({50.f, static_cast<float>(i * 20) + 50.f});
+                m_logText.setString(m_log[i]);
+                m_window.draw(m_logText);
+            }
```
MarioLiebisch, line 165: I would have opted to only update the `sf::Text`, if the log actually changed.

examples/event_handling/EventHandling.cpp
```diff
@@ -0,0 +1,224 @@
...
+    {
+        m_log.emplace_back("3: Touch Began: " + vec2ToString(touchBegan.position));
+    }
+
+    template <typename T>
+    void handle(const T&)
+    {
+        // All unhandled events will end up here
+        // m_log.emplace_back("3: Other Event");
+    }
```
MarioLiebisch, line 201: This block should have its own comment explaining how it's being used. Currently this feels a bit out of the blue, if you haven't read the specific case's details/implementation above.

examples/event_handling/EventHandling.cpp
```diff
@@ -0,0 +1,224 @@
...
+/// Entry point of application
+///
+/// \return Application exit code
+///
+////////////////////////////////////////////////////////////
+int main()
+{
+    Application application;
+    application.run();
+}
```
MarioLiebisch, line 224: I'd always use explicit returns in example codes.

include/SFML/Window/WindowBase.hpp
```diff
@@ -220,11 +220,56 @@ class SFML_WINDOW_API WindowBase
...
+    ////////////////////////////////////////////////////////////
+    /// \brief Handle all pending events
+    ///
+    /// This function is not blocking: if there's no pending event then
+    /// it will return without calling any of the handlers.
+    ///
+    /// This function takes a variadic list of event handlers that
+    /// each take a concrete event type as a single parameter. The
+    /// event handlers can be any kind of callable object that has
+    /// an operator() defined for a specific event type.
```
MarioLiebisch, line 237: > that has an `operator()` taking a specific event type as its sole argument.

Maybe?

include/SFML/Window/WindowBase.hpp
```diff
@@ -220,11 +220,56 @@ class SFML_WINDOW_API WindowBase
...
+    ///
+    /// Examples:
+    /// - Lambda expressions: `[&](const sf::Event::KeyPressed) { ... }`
+    /// - Free functions: `void handler(const sf::Event::KeyPressed&) { ... }`
+    ///
+    /// Calling member functions is supported through lambda
+    /// expressions.
```
MarioLiebisch, line 244: I know what it's meant to express, but at the same time it feels kind of (un-)specific, too. Shouldn't it work with something like `std::mem_fn()`, too?

include/SFML/Window/WindowBase.hpp
```diff
@@ -220,11 +220,56 @@ class SFML_WINDOW_API WindowBase
...
+    ///     }
+    ///     else
+    ///     {
+    ///         // Handle non-KeyPressed
+    ///         handleOtherEvents(event);
+    ///     }
+    /// }
+    /// \endcode
+    ///
+    /// \param handlers A variadic list of callables that take an event as their only parameter
```
MarioLiebisch, line 265: I think this should repeat:

> take a specific event as their only parameter

examples/event_handling/EventHandling.cpp
ChrisThrasher: https://github.com/SFML/SFML/issues/2128

I tried (and failed) to get support for that.

examples/event_handling/EventHandling.cpp
ChrisThrasher: We stopped doing that a while ago. C++ guarantees a return value of `0` if `main` finishes so I prefer omitting that unnecessary `return 0;` line. Oddly C _does_ require it. It's a subtle difference between languages.

examples/event_handling/EventHandling.cpp
binary1248: This is how it is written in all other examples, so the same pattern is followed here.

examples/event_handling/EventHandling.cpp
binary1248: Fixed.

examples/event_handling/EventHandling.cpp
binary1248: Fixed.

examples/event_handling/EventHandling.cpp
binary1248: Fixed.

examples/event_handling/EventHandling.cpp
binary1248: That would make the example longer and more complex than necessary. Performance isn't really a concern here.

examples/event_handling/EventHandling.cpp
binary1248: Fixed.

include/SFML/Window/WindowBase.hpp
binary1248: Fixed.

include/SFML/Window/WindowBase.hpp
binary1248: Fixed.

include/SFML/Window/WindowBase.hpp
binary1248: `mem_fn` doesn't really work here. The provided callable has to be callable when passed an event object and currently the event object doesn't have any member functions that would be useful in such a context. In general, lambda expressions can be used everywhere where `mem_fn` would have been used anyway.

MarioLiebisch:
Just as a last single random idea: Maybe show the currently active handler type on screen or in the title bar text? Or at least print it to the log, too?

include/SFML/Window/WindowBase.inl
```diff
@@ -0,0 +1,75 @@
...
+
+#include <variant>
+
+
+namespace sf
+{
+namespace priv
+{
+template <typename... Ts>
+struct Overloaded : Ts...
```
ChrisThrasher, line 39:
```suggestion
struct OverloadSet : Ts...
```

include/SFML/Window/WindowBase.inl
```diff
@@ -0,0 +1,75 @@
...
+{
+    using Ts::operator()...;
+};
+template <typename... Ts>
+Overloaded(Ts...) -> Overloaded<Ts...>;
+
+template <typename... Ts>
+struct OverloadedWrapper : Overloaded<Ts...>
```
ChrisThrasher, line 47:
```suggestion
struct OverloadSetWithDefault : OverloadSet<Ts...>
```

include/SFML/Window/WindowBase.inl
```diff
@@ -0,0 +1,75 @@
...
+{
+    // By providing our own operator() and forwarding based
+    // on invocability of Overloaded<Ts...> on the concrete type
+    // of the value, we save the user from having to provide
+    // their own catch-all overload if they don't want to
+    template <typename T>
+    void operator()(T&& value) // NOLINT(cppcoreguidelines-missing-std-forward)
```
ChrisThrasher, line 54:
```suggestion
    void operator()([[maybe_unused]] T&& value) // NOLINT(cppcoreguidelines-missing-std-forward)
```

include/SFML/Window/WindowBase.inl
```diff
@@ -0,0 +1,75 @@
...
+    {
+        if constexpr (std::is_invocable_v<Overloaded<Ts...>, T>)
+            Overloaded<Ts...>::operator()(std::forward<T>(value));
+
+        (void)value;
```
ChrisThrasher, line 59:
```suggestion
```

include/SFML/Window/WindowBase.inl
```diff
@@ -0,0 +1,75 @@
...
+    }
+};
+template <typename... Ts>
+OverloadedWrapper(Ts...) -> OverloadedWrapper<Ts...>;
+} // namespace priv
+
+
+////////////////////////////////////////////////////////////
+template <typename... Ts>
+void WindowBase::handleEvents(Ts&&... handlers) // NOLINT(cppcoreguidelines-missing-std-forward)
```
ChrisThrasher, line 69: Why did you need to disable this clang-tidy check in a couple places?

ChrisThrasher (changes requested):

ChrisThrasher:
> Just as a last single random idea: Maybe show the currently active handler type on screen or in the title bar text? Or at least print it to the log, too?

I like this idea. When running the example I wasn't sure which event handling mechanism I was using.

include/SFML/Window/WindowBase.inl
binary1248: clang-tidy on specific platforms would still complain about missing `std::forward` even though it was there.

Or in other words, clang-tidy was spitting out false positives on certain systems (not all).

https://github.com/SFML/SFML/actions/runs/9714239737/job/26812967949

binary1248:
Updated.

include/SFML/Window/Event.hpp
```diff
@@ -41,6 +41,8 @@
 
 namespace sf
 {
+class WindowBase;
+
```
ChrisThrasher, line 45: You don't need this. `friend class` will forward declare `WindowBase` if need be.

ChrisThrasher (commented):
I'm think I'm okay with this. The new example program does a good job explaining how to use the API. I thought about how to write any tests for this and couldn't think of anything. I'm guessing you came to the same conclusion.

@eXpl0it3r what do you think?

include/SFML/Window/Event.hpp
binary1248: Fixed.

binary1248:
> I thought about how to write any tests for this and couldn't think of anything. I'm guessing you came to the same conclusion.

Since this just adds a bit of template boilerplate on top of the already existing event handling, the fact that we didn't have event handling tests before means that this will also not be able to be tested.

ChrisThrasher (approved):

include/SFML/Window/WindowBase.inl
```diff
@@ -0,0 +1,73 @@
...
+OverloadSetWithDefault(Ts...) -> OverloadSetWithDefault<Ts...>;
+} // namespace priv
+
+
+////////////////////////////////////////////////////////////
+template <typename... Ts>
+void WindowBase::handleEvents(Ts&&... handlers) // NOLINT(cppcoreguidelines-missing-std-forward)
+{
+    while (const std::optional event = pollEvent())
+        std::visit(priv::OverloadSetWithDefault{std::forward<Ts>(handlers)...}, event->m_data);
```
kimci86, line 70: Looks like a potential use-after-move.
A handler could own a movable resource (imagine some std::unique_ptr for example). In that case, said resource would be usable in the handler only on the first iteration that uses that handler.
I guess we should instantiate the OverloadSetWithDefault once, not in the loop.

include/SFML/Window/WindowBase.inl
```diff
@@ -0,0 +1,73 @@
...
+template <typename... Ts>
+OverloadSetWithDefault(Ts...) -> OverloadSetWithDefault<Ts...>;
+} // namespace priv
+
+
+////////////////////////////////////////////////////////////
+template <typename... Ts>
+void WindowBase::handleEvents(Ts&&... handlers) // NOLINT(cppcoreguidelines-missing-std-forward)
+{
+    while (const std::optional event = pollEvent())
```
kimci86, line 69: What if the user wants event visitation and `waitEvent`?
Couldn't `visit` simply be a method of the `sf::Event` class?

binary1248: Making visit part of `sf::Event` would mean the user would still have to pick up the events themselves and check for validity. `handleEvents` handles pending events without waiting by design. The main goal of this addition is to alleviate the majority of users from having to deal with validity checking completely. The users who need to wait for events in their application can still handle them the "classical" way using `waitEvent`. I consider using `waitEvent` "advanced usage" so it is also acceptable for those who need that functionality to have to implement more of the event handling themselves.

include/SFML/Window/WindowBase.inl
binary1248: Fixed.

include/SFML/Window/WindowBase.inl
ChrisThrasher: > What if the user wants event visitation and waitEvent?
Couldn't visit simply be a method of the sf::Event class?

That's a good point. It begs the question why we don't just handle visitation through the `sf::Event` API like previous proposals. It would let us avoid using a `friend` which obscure the relationship between classes.

```cpp
while (const std::optional event = window.{pollEvent or waitEvent}())
    event->visit([&](const sf::Event::Closed&){ ... }, ...);
```

> I consider using waitEvent "advanced usage"

I consider visitation to be "advanced usage" as well. It takes a certain level of C++ proficiency to structure your app around callbacks rather than more conventional control flow structures.

binary1248: > I consider visitation to be "advanced usage" as well. It takes a certain level of C++ proficiency to structure your app around callbacks rather than more conventional control flow structures.

Which is why this change goes out of its way to keep this concept under the hood. For a beginning user, they will not have to know what a variant is, nor what visitation means. The fact that _we_ know what is happening behind the scenes doesn't make it harder for beginners to learn. You can pick up any beginner Java book and this is a pattern they will learn on day 1 when designing UI applications. The concept of passing callables to a function is not a hard one to learn, but as usual when some straight-forward concept is introduced to C++-land it becomes "spooky" and for whatever reason we start to worry that it will be too hard for beginners to grasp.

Having to check less things in order to get code that works will always be more beginner friendly in my book. If anyone has read "Programming: Principles and Practice Using C++", they will know that dynamic allocation only comes deep into the latter half of the book, for good reason. Even though we know that a `std::optional` doesn't dynamically allocate, the fact that you have to check if it is valid and then use the same operator to "dereference" it as you would a pointer just screams "advanced concept" to me.

I find it worrying that there has been a trend lately that disregards the little things like this. While a beginning user was perfectly capable of writing complete games in SFML 2 without having to use the `->` operator a single time in their code base or having to check if any "pointer-like-object" was valid, we just litter the current API at every corner we can with these concepts all in the name of "safety". What use is a safe API if the users we are targeting don't even know what half the symbols on the screen mean? I would even go so far as to claim that the new API would lead to even more "problems" if a novice user who is unfamiliar with these concepts were to jump head first into it. Sure we could claim that it is still "safer" but replacing one problem with another is still not going to leave a good impression with the user whose program doesn't work for reasons they don't yet understand.

I guess I'll have to repeat the motivation of this change once more: This change aims to reduce the amount of assumed knowledge required to handle events properly and also reduce the number of mandatory "boilerplate" steps required to write code that doesn't crash. If having to write less code to achieve the same goal makes the experience more challenging and error prone then I don't know what else to say.

If the main concern is that the `waitEvent` enjoyers are left in the rain with this change, we can still add another visitation function for them as well. Just don't make the beginner experience more horrible than it has to be because of some really niche use cases.

binary1248: `visit` has been added to `sf::Event` and is used by `handleEvents` in the current revision.

ChrisThrasher (approved):
I love this new design! Moving `.visit()` into `sf::Event` fixes the issue of not supporting `waitEvent` plus it means we can write tests for more of this logic all while removing a hidden `friend`.

kelbon:
1. non const version required
2. remove old api and std::variant, just switch instead

binary1248:
> non const version required

`visit` is `const` because it doesn't make any sense that a visitor would have to mutate the event itself. Even in the old API, nobody would ever modify the event data themselves unless they were misusing the event for purposes it was never intended for, which was never officially supported anyway.

> return old api and remove std::variant, just switch instead

You can't simultaneously ask for `visit` support and say we have to remove `std::variant`, or do you expect us to implement our own variant functionality instead of just using the standard library? The fact the implementation uses `std::variant` is just an implementation detail and it could have just as well been implemented in numerous other ways and still lead to the current API.

You can express your issues with the public API, but asking us to change the private implementation because you don't like bits of it isn't very constructive.

The main reason for getting rid of switch support is because it was still possible to check for an event type and query the wrong event data because they weren't linked. Now, you might claim that you are such a good programmer that you would never make that mistake, but just remember, you aren't the only one using this library and we have gotten many reports of users making this exact mistake over the years, thus the current solution to this very common problem.

To be honest, I also find the new API a bit more verbose than the older one, but this slight increase in verbosity is the price we are willing to pay to eliminate a very common source of errors. I also wish there was a solution that would be as terse as the switch statement and provide the same safety as the current implementation, but currently such a solution is not supported in the C++ language yet.

You can be absolutely sure that after [pattern matching](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p1371r3.pdf#subsection.4.4) is standardized, we will implement support for it into the library and you will be able to replace your visitors with the new pattern matching statement which looks very similar to a switch statement. Until then we are just making do the best we can with what the language offers us right now.

danieljpetersen:
It really is unfortunate because pattern matching would probably make literally everyone happy, it's the perfect solution to this problem.

kelbon:
> You can express your issues with the public API, but asking us to change the private implementation because you don't like

Its not implementation detail, since all users 100% already saw and used this 'implementation detail'

> visit is const because it doesn't make any sense that a visitor would have to mutate the event itself. Even in the old API, nobody would ever modify the event

That the problem. Some people for some reason think, that 'its not how it should be used' and forbid. Thats Rust's minded. In fact, while there are no real problem it should be allowed, since it was allowed already

>  you might claim that you are such a good programmer that you would never make that mistake

firstly, users will get visit, if they want to use it. Now you will force newbie programmer to know about:

std::variant sumtypes and other buzzwords
overloading
inheriting
"overloaded" pattern

and force them to deal with worst compiler errors ever, additionaly slow their compilation

binary1248:
> Its not implementation detail, since all users 100% already saw and used this 'implementation detail'

Maybe there is some misunderstanding on what "implementation detail" means. Implementation detail in this context means that the way the publicly accessible functionality is implemented in code is chosen at the discretion of the library and can be subject to change without notice. Users should not rely on undocumented implementation details such as this. The only guarantee libraries make is that the _public API_ will stay stable. I don't know what exactly you mean by "used this implementation detail" but if you are referring to users circumventing the private access specifier to get direct access to the private variant member, then they are just setting themselves up for all kinds of trouble hence this is never going to be a good idea.

> That the problem. Some people for some reason think, that 'its not how it should be used' and forbid. Thats Rust's minded. In fact, while there are no real problem it should be allowed, since it was allowed already

I don't know how much programming experience you have, but if you have used many other libraries (or even operating system APIs like the Windows API), you should probably know that relying on undocumented behaviour that isn't explicitly forbidden by the API documentation will never be supported by the maintainers and can/will lead to code eventually breaking at some time in the future. SFML is not the first library that has adopted this policy nor will it be the last. Also, any programming language that makes efforts to improve the programmer experience/productivity will eventually also try their best to reduce the possibilities of writing incorrect code. This isn't specific to Rust and has been a trend for decades, even before Rust existed.

> firstly, users will get visit, if they want to use it. Now you will force newbie programmer to know about:
>
> std::variant sumtypes and other buzzwords
> overloading
> inheriting
> "overloaded" pattern
>
> and force them to deal with worst compiler errors ever, additionaly slow their compilation

It seems to me like the issues you have with the proposed change are actually issues you have with C++ as a programming language. Yes... it has been known for a very long time that C++ compile times and compiler errors are probably not its strong point, but improving these areas is the responsibility of the compiler vendors and they have made some good improvements in the last years.

SFML is a C++ library, and while it would be nice for beginners to be able to use SFML without prior C++ knowledge, that is never going to happen and was also never the case in SFML 2. When designing the library API, we have to consider what we expect beginners to know in order to be able to use it effectively. We already established that no prior C++ knowledge is out of the question, so the real question is where we draw the line regarding minimum requirements.

Overloading and inheritance are one of the main additions of C++ over C since C++ provides language-level support for object oriented programming. Because they are part of the core principles of C++, they are taught very early to beginners in any of the teaching materials I have seen so far. Because of this, we should be able to expect SFML users to have a firm grasp of these concepts before trying to use SFML.

As I already said above, there is currently nowhere in the documented public API where the user is expected to understand what `std::variant` is nor how to use it in their own code. The fact that you know how the functionality is currently implemented in code doesn't make it a requirement for new users who just want to use the library without knowing how it works under the hood.

You have to understand, SFML is a C++ library. SFML 2 limited itself to C++98 for more than a decade at this point. I don't think there are many libraries out there that are still in active development that artificially limit themselves to such an old C++ standard. As time moves on, the C++ standard committee evolves the language to address the issues that have been plaguing C++ since C++98. When languages evolve, they adopt new ideas and concepts. Sometimes these concepts might bear some resemblance to those found in other languages (e.g. Python, Rust, Java etc.) because it turns out those are good ideas. Looking back in history you will find that the way many programming languages have eventually died out was from their inability to evolve over time, and as a C++ user myself, this is not an outcome I want to see with C++ for the foreseeable future.

Because SFML is a C++ library and C++ is constantly evolving there will inevitably be a point at which SFML will choose to adopt functionality from the newer language standards. That point is SFML 3.

You are free to continue using SFML 2 for as long as you please if you are not fond of modern C++, but just remember that you aren't really doing yourself any favours by sticking to older language standards. The majority of C++ libraries out there will have chosen to go down the same path as SFML and adopt newer language features, in fact SFML is probably one of the latest libraries to do so. We are not and have never advocated for legacy codebases to migrate to SFML 3. If the functionality that SFML 2 provides was enough to satisfy the older project requirements then there is no need to migrate to SFML 3. When starting a new project however, we recommend considering using a newer C++ standard which not only provides access to newer language features and the newer standard library but also the third party library ecosystem that requires compiler support for newer language standards as well.

I can't stress enough, all of the arguments you have presented so far lead me to believe that you have fundamental issues with the C++ language itself and not SFML as a C++ library. The fact that you have mentioned multiple times that you consider everything you dislike about C++ "Rust-like" shows that you have a different opinion on how the language should evolve (or if the language should even evolve at all). Everyone is entitled to their own opinion and as I said, you are free to continue using C++98 and SFML 2 in your own project, but you have to understand that this does not reflect the sweeping majority of C++ programmers I have encountered. If you are going to make further technical arguments about this change, it would be more productive to do so from the perspective of a programmer that embraces modern C++ instead of the opposite.

kelbon:
> but if you are referring to users circumventing the private access specifier to get direct access to the private variant member

Im not talking about private variant member, im talking about cuurrent api with union, which is totaly broken and its not easy to rewrite one to another, even event names are changed

> > firstly, users will get visit, if they want to use it. Now you will force newbie programmer to know about:
> > std::variant sumtypes and other buzzwords
> > overloading
> > inheriting
> > "overloaded" pattern
> > and force them to deal with worst compiler errors ever, additionaly slow their compilation
> 
> It seems to me like the issues you have with the proposed change are actually issues you have with C++ as a programming language

Im? No, but new users will try your library and never try again without sending a message, because they will not understand how to use it

> You have to understand, SFML is a C++ library. SFML 2 limited itself to C++98

And even in C+98 you can write Event::visit without any problems, if you understand C++

You want safety? Okay, add () after each field, for example:

```cpp
struct Event {
...
  int index;
  union {
    ield_t _field;
...
  }
  field_t& field() {
    assert(index == _field_index);
    return _field;
  }
}

```
> Because SFML is a C++ library and C++ is constantly evolving there will inevitably be a point at which SFML will choose to adopt functionality from the newer language standards. That point is SFML 3.

New language features should be used to improve user expirience and performance

But now its used just because of R* language. For example, removed default state of Event. Whats a point? There are no point, just to use Option\<Event\> oh, sorry, std::optional\<Event\>. But C++ has its generaly accepted rules, you should write understandable code, understandable in this case == explicit operaton bool and default state

> it would be more productive to do so from the perspective of a programmer that embraces modern C++ instead of the opposite.

if all will use modern C++ as an excuse to breaking api, compilation speed and performance, people will not use C++

include/SFML/Window/Event.hpp
```diff
@@ -299,6 +299,17 @@ class SFML_WINDOW_API Event
     template <typename TEventSubtype>
     [[nodiscard]] const TEventSubtype* getIf() const;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Apply a visitor to the event
+    ///
+    /// \param visitor The visitor to apply
+    ///
+    /// \return The result of the applying the visitor to the event
```
kimci86, line 307:
```suggestion
    /// \return The result of applying the visitor to the event
```

kimci86:
This branch looks ok to me, but it brings to light a problem with event subtypes hierarchy.

There is some inconsistency between visitation and "classical" `is`/`getIf` approach.
With visitation, it is possible to write a handler for an event subtype base class, such as `sf::Event::KeyChanged` for example.

```c++
window.handleEvents([&](const sf::Event::KeyChanged& keyChanged) { /* use keyChanged... */ });
```

But this is not possible with `is`/`getIf` which only accept event subtypes (not their base classes). We have to call both `getIf<sf::Event::KeyPressed>` and `event->getIf<sf::Event::KeyReleased>` separately.

To my understanding, class hierarchy of event subtypes is only a way to avoid code duplication in SFML implementation, but not a meaningful part of public API. This is even causing unhelpful auto-completion hints when looking for event subtypes to use as `is` or `getIf` parameter, and may lead to compiler error "TEventSubtype must be a subtype of sf::Event" which is confusing when the type parameter actually starts with `sf::Event::`.

One solution would be to pour some template magic to allow `is`/`getIf` to take event subtype base class, but I doubt it is worth the effort for an uncertain usefulness.
This is why I suggest not to use inheritance between event subtypes. It avoids the inconsistency between visitation and `is`/`getIf` approach, it makes API simple for users and auto-completion, and it avoids the need for decisions about designing the most correct hierarchy (like, why don't FocusLost and FocusGained have a common base class?).

It is not blocking for this current PR but I think it is important before releasing SFML 3 API.

ChrisThrasher:
> To my understanding, class hierarchy of event subtypes is only a way to avoid code duplication in SFML implementation, but not a meaningful part of public API.

This is not the case. The use of inheritance was an intentional part of the API. Search the codebase for `KeyChanged` and you'll see a few places we make use of that polymorphism. That being said I'm open to reconsidering the use of polymorphism unless we can find a way to reconcile the asymmetry you pointed out.

include/SFML/Window/WindowBase.inl
```diff
@@ -0,0 +1,77 @@
...
+
+template <typename... Ts>
+struct OverloadSetWithDefault : OverloadSet<Ts...>
+{
+    // By providing our own operator() and forwarding based
+    // on invocability of OverloadSet<Ts...> on the concrete type
+    // of the value, we save the user from having to provide
+    // their own catch-all overload if they don't want to
+    template <typename T>
+    void operator()([[maybe_unused]] T&& value) // NOLINT(cppcoreguidelines-missing-std-forward)
```
eXpl0it3r, line 54: Is the check failing, because not all code paths do a forward?  
Is this bad?

include/SFML/Window/Event.hpp
```diff
@@ -299,6 +299,17 @@ class SFML_WINDOW_API Event
     template <typename TEventSubtype>
     [[nodiscard]] const TEventSubtype* getIf() const;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Apply a visitor to the event
+    ///
+    /// \param visitor The visitor to apply
+    ///
+    /// \return The result of applying the visitor to the event
```
eXpl0it3r, line 307: This should be a bit more descriptive. What kind of results can the user expect?

examples/event_handling/EventHandling.cpp
```diff
@@ -0,0 +1,338 @@
...
+                        m_log.emplace_back("Mouse Pressed");
+                    }
+                    else if (const auto* touchBegan = event->getIf<sf::Event::TouchBegan>())
+                    {
+                        m_log.emplace_back("Touch Began: " + vec2ToString(touchBegan->position));
+                    }
+                    else
+                    {
+                        // All unhandled events will end up here
+                        // m_log.emplace_back("Other Event");
```
eXpl0it3r, line 143: Why comment it out?

examples/event_handling/EventHandling.cpp
```diff
@@ -0,0 +1,338 @@
...
+            // Limit the log to 24 entries
+            if (m_log.size() > 24u)
+                m_log.erase(m_log.begin(), m_log.begin() + static_cast<int>(m_log.size() - 24u));
+
+            // Draw the contents of the log to the window
+            m_window.clear();
+
+            for (std::size_t i = 0; i < m_log.size(); ++i)
+            {
+                m_logText.setPosition({50.f, static_cast<float>(i * 20) + 50.f});
```
eXpl0it3r, line 250: `i * 20.f` would remove the need for a cast

examples/event_handling/EventHandling.cpp
```diff
@@ -0,0 +1,338 @@
...
+    {
+        m_window.close();
+    }
+
+    void handle(const sf::Event::KeyPressed& keyPress)
+    {
+        m_log.emplace_back("Key Pressed: " + sf::Keyboard::getDescription(keyPress.scancode));
+
+        // When the enter key is pressed, switch to the next handler type
+        if (keyPress.code == sf::Keyboard::Key::Enter)
```
eXpl0it3r, line 276: Personally recommend scancode usage as much as possible, not that it would really matter here

include/SFML/Window/WindowBase.inl
```diff
@@ -0,0 +1,77 @@
...
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#include <SFML/Window/Event.hpp>
+#include <SFML/Window/WindowBase.hpp> // NOLINT(misc-header-include-cycle)
```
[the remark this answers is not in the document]
binary1248: To stay consistent with Event.inl 😁

include/SFML/Window/WindowBase.inl
binary1248: It failed only on clang-tidy on Windows, so I have no idea.

include/SFML/Window/Event.hpp
binary1248: "The result of applying the visitor to the event" is the description of the return type, since the user gets to decide this there is no other way to describe it.

examples/event_handling/EventHandling.cpp
binary1248: To reduce spam by a tiny bit, and to encourage experimentation.

examples/event_handling/EventHandling.cpp
binary1248: Just following the other examples.

include/SFML/Window/Event.hpp
eXpl0it3r: One can still be more precise in specifying what is returning here, as in the result as specified by the visitors implementation.

examples/event_handling/EventHandling.cpp
binary1248: This is weird... thought the compiler would have preferred promoting `std::size_t * float` to `double` instead, I guess the standard doesn't mind that precision is lost here.

include/SFML/Window/Event.hpp
binary1248: But... isn't that just saying the same thing just with different words? 😁 

Not to mention, maybe the implementation doesn't even specify the return type also just relying on it being automatically deduced.

eXpl0it3r: Yes, but right now it's very unclear how it happens, unless you go check under the hood. The result just seems to "spring into existence" and it's not clear that it's defined by the provided visitor, it might as well be defined by the event or derived from cosmic rays.

I think the difference in understanding comes from the point that "applying the visitor to the event" communicates to you a clear concept, while it might not be as clear to others.

examples/event_handling/EventHandling.cpp
eXpl0it3r: I mean if it starts erroring with loss in conversion, then I guess we need this.

examples/event_handling/EventHandling.cpp
eXpl0it3r: Maybe the policy should be that we use scancodes if it's layout dependent (e.g. characters), but keycode is fine generic keys like "Return".

include/SFML/Window/Event.hpp
binary1248: I am open to suggestions for how the whole line should be written verbatim so there is no ambiguity. 😁

examples/event_handling/EventHandling.cpp
binary1248: The current code works with a cast and is guaranteed not to produce any conversion warnings so...

examples/event_handling/EventHandling.cpp
binary1248: Sounds reasonable, good thing we don't have a FPS example yet. 😁

include/SFML/Window/WindowBase.inl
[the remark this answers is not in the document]
eXpl0it3r: Was done as part of #2897

Issue #3047, opened 2024-05-27
pass sf::Vector2<T> and sf::Vector3<T> by value instead of by references


ZXShady:
### Prerequisite Checklist

- [X] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [X] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [X] I understand [SFML's scope](https://www.sfml-dev.org/contribute.php#general-considerations) and believe the feature fits it

### Describe your feature request here

Simple structs like these should be passed by value instead of by references in many places and get* functions should return by value too.

### Use Cases

it is just for extra performance for functions defined out of line.
as the that function could internally use a `const_cast`  on the reference passed legally that would disable many optimizations and the compiler had to take into account aliasing issues with global variables.

Simple trivially copyable structs/classes should be passed by value to avoid rhe issue above.

 functions that are defined in another translation unit and takes a parameter by const reference the caller has to assume that the called function could change the paramter and the compiler has to take that into account
`void f(int const&)` f could change its arguement and that is legal only if the passed int arguement was not orignally const.

vittorioromeo:
Can you provide a benchmark where this matters? 

Generally speaking, if the function is called so frequently that it becomes significant in the benchmark, it should be inlined and whether the vector is passed by value or by reference shouldn't be important.

ChrisThrasher:
As far as I’m aware there’s no noticeable difference between passing a `sf::Vector2i` or `sf::Vector2f` by value or const reference. However I do think 3D vectors are large enough that passing by const reference is superior so I don’t see why we’d ever change 3D vector parameter to be passed by value.

ChrisThrasher:
Closing this issue until we receive more evidence for why passing by value is an improvement.

ZXShady:
hey sorry for the extremely late reply yikes 3 whole weeks.

passing by reference means if the compiler cannot know the body of the function means that your variable could be theoritcally changed by the function


```
void print(sf::Vector2f const&); // defined in another translation unit

int main() {
    sf::Vector2f v{4,2};
    print(v);
    return v.x; // won't be optimized into return 4 since this could change by the function
// since the function could const_cast it away!
}

```
[by reference](https://godbolt.org/z/axTbo8hMo)
[by value](https://godbolt.org/z/zEPx47T8s )

[...]

PR #3140, opened 2024-06-29
Pass `sf::Vector2<T>`s by value

ChrisThrasher:
## Description

Addresses part of #3047.

As a rule of thumb, if the type is trivially copyable and is less than or equal to the CPU register width times two then you ought to pass it by value. This will lead to more efficient code generation. We already do this with `sf::Time` and `sf::Angle` with are both 4 bytes so I added tests for trivial copyability to those two types as well.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-in

### DIFF
--- a/examples/raw_input/RawInput.cpp
+++ b/examples/raw_input/RawInput.cpp
@@ -42,7 +42,7 @@ int main()
                 break;
             }
 
-            static const auto vec2ToString = [](const sf::Vector2i& vec2)
+            static const auto vec2ToString = [](const sf::Vector2i vec2)
             { return '(' + std::to_string(vec2.x) + ", " + std::to_string(vec2.y) + ')'; };
 
             if (const auto* const mouseMoved = event->getIf<sf::Event::MouseMoved>())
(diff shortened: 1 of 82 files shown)


ZXShady:
still passing by reference Glsl::Vec2 in Shader.hpp at line 333

ZXShady:
still passing by reference Glsl::IVec and BVec in Shader.hpp

the number is the line number

  `378 void setUniform(const std::string& name, const Glsl::Ivec2& vector);
// alias to sf::Vector2i`

`422  void setUniform(const std::string& name, const Glsl::Bvec2& vector);
// alias to sf::Vector2<bool>`


`431    void setUniform(const std::string& name, const Glsl::Bvec3& vector);
// sf::Vector3<bool>`

`440  void setUniform(const std::string& name, const Glsl::Bvec4& vector);
// alias to some type having 4 bools`


Ofcourse sizeof bool is implementation defined  but every system in existence that has actually a bool type (note  old visual studio used to typedef bool to an int).
sizeof(bool) is usually 1 so passing bools by reference is bad.

ZXShady (approved):

eXpl0it3r (approved):
Would still be kind of interesting to run a before and after test in debug and release for some hot paths. I don't see any issues for kind of one-off calls, just wonder whether for example the Vector2 operators can cause issues.

PR #3142, opened 2024-06-30
Add move semantics to `sf::WindowBase` class heirarchy

ChrisThrasher:
## Description

Puling this off takes two steps spilt into two commits. The first moves the bookkeeping for fullscreen windows from `WindowBase` to `WindowImpl`. This is a worthwhile cleanup on its own. It's a net reduction in code and lets us use a default destructor for `sf::Window` and `sf::WindowBase`. 

Lastly once those destructors are defaulted then we can add default move operations to the classes to get move semantics for free!


src/SFML/Window/WindowImpl.cpp
```diff
@@ -90,9 +90,19 @@ using WindowImplType = sf::priv::WindowImplAndroid;
 #endif
 
 
-namespace sf::priv
+namespace
+{
+// A nested named namespace is used here to allow unity builds of SFML.
+// Yes, this is a rather weird namespace name. ¯\_(ツ)_/¯
+namespace WindowsImplImpl
```
binary1248, line 97: `WindowImplImpl`

src/SFML/Window/WindowImpl.cpp
```diff
@@ -90,9 +90,19 @@ using WindowImplType = sf::priv::WindowImplAndroid;
 #endif
 
 
-namespace sf::priv
+namespace
+{
+// A nested named namespace is used here to allow unity builds of SFML.
+// Yes, this is a rather weird namespace name. ¯\_(ツ)_/¯
```
binary1248, line 96: It would be a pity if this was the only non-ASCII line in this file.

binary1248 (approved):

PR #3144, opened 2024-07-01
Disallow constructing `sf::String` from `nullptr`

ChrisThrasher:
## Description

This is not a sensible operation to perform. If a user tries to do this then their program is likely in a bad state. It's better to assert against this behavior than quietly handle it.

The first way to protect against this is by deleting a construct overload using `std::nullptr_t`. If a user tries to construct a string from `nullptr`, this will result in a compiler error. If the user creates a null `char*`, `wchar_t*`, or `char32_t*`, then there are new runtime `assert`s that will detect that error in Debug builds.

### DIFF
--- a/include/SFML/System/String.hpp
+++ b/include/SFML/System/String.hpp
@@ -111,6 +111,14 @@ public:
     String() = default;
 
     ////////////////////////////////////////////////////////////
+    /// \brief Deleted nullptr constructor
+    ///
+    /// Disallow construction from nullptr literal
+    ///
+    ////////////////////////////////////////////////////////////
+    String(std::nullptr_t, const std::locale& = {}) = delete;
+
+    ////////////////////////////////////////////////////////////
     /// \brief Construct from a single ANSI character and a locale
     ///
     /// The source character is converted to UTF-32 according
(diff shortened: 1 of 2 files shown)


eXpl0it3r:
I'm fine with deleting the constructor, as this gives up a compile-time error.

But I'm against making nullptr via `char*` undefined behavior. Sure, in debug you get an assert now, but in release you just invoke undefined behavior by calling `std::strlen(nullptr)`. I don't think we have to follow the standard library design of adding undefined behavior, when we have the chance of defining a valid behavior.

binary1248:
I also think we should try not to introduce new places where undefined behaviour can happen.

While there is precedent for places where passing in a `nullptr` is just silently ignored like in the current implementation e.g. `std::free`, I still think that passing in any nulled pointer would be almost always unintentional and we should provide a "diagnostic".

I say keep the runtime `nullptr` check to avoid UB but throw an `assert(false, ...)` into the else path so that the user can get notified that nothing happened when running in debug.

eXpl0it3r:
> I say keep the runtime `nullptr` check to avoid UB but throw an `assert(false && ...)` into the else path so that the user can get notified that nothing happened when running in debug.

Wouls be fine for me as well.

ChrisThrasher:
I love asserting against UB but am less inclined to assert against behavior that we already define. I removed the assert, kept the new deleted constructor, and added new tests that cover the case where users provide a null pointer since we want that to be an explicitly defined behavior of the library. Now the tests cover that.

binary1248 (approved):

PR #3149, opened 2024-07-04
Modernize management of Udev pointers

ChrisThrasher:
## Description

Follows the same pattern as #2540, #2759, and #2801. Using `unique_ptr` makes for safer and more expressive code that can more easily be modified without introducing subtle resource management bugs.

### DIFF
--- a/src/SFML/Window/Unix/JoystickImpl.cpp
+++ b/src/SFML/Window/Unix/JoystickImpl.cpp
@@ -32,6 +32,7 @@
 #include <fcntl.h>
 #include <libudev.h>
 #include <linux/joystick.h>
+#include <memory>
 #include <ostream>
 #include <poll.h>
 #include <string>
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
Just reminded me - not blocking for this PR - wasn't there a discussion once to replace udev with something not as ancient?

ChrisThrasher:
> Just reminded me - not blocking for this PR - wasn't there a discussion once to replace udev with something not as ancient?

Not that I recall. Perhaps you're thinking of talks about using alternative Windows APIs for joysticks?

eXpl0it3r:
Ah yeah, I think it might have been in connection with that.

binary1248:
There is no need for such heavy templating:
```cpp
struct UdevDeleter
{
    void operator()(udev_device* device) const
    {
        udev_device_unref(device);
    }
    void operator()(udev_monitor* monitor) const
    {
        udev_monitor_unref(monitor);
    }
    void operator()(udev_enumerate* enumerate) const
    {
        udev_enumerate_unref(enumerate);
    }
    void operator()(udev* context) const
    {
        udev_unref(context);
    }
};

template <typename T>
using UdevPtr = std::unique_ptr<T, UdevDeleter>;
```

binary1248 (approved):

PR #3151, opened 2024-07-06
Fix how macOS fullscreen video modes are detected

### LINKED ISSUE #2300: [MacOS] Empty `std::vector<VideoMode>` returned from `getFullscreenModes()`, causing segfault
## Subject of the issue

This issue is described in the forums, here:
https://en.sfml-dev.org/forums/index.php?topic=28601.0

I am not the OP of that thread, but I (and others) are seeing the same issue. Essentially, on MacOS (latest versions), a call to `getFullScreenModes()` (https://github.com/SFML/SFML/blob/2.6.x/src/SFML/Window/OSX/VideoModeImpl.cpp#L40) returns an empty vector of video modes. This is because each candidate is rejected by the `continue` at line 65, owing to all of them being larger than the reported desktop width/height.

Adding some debug output at that point shows the following modes are all rejected:

```
mode.size.x = 3840, desktop.size.x = 3420, mode.size.y = 2400, desktop.size.y = 2224
mode.size.x = 4096, desktop.size.x = 3420, mode.size.y = 2560, desktop.size.y = 2224
mode.size.x = 4096, desktop.size.x = 3420, mode.size.y = 2664, desktop.size.y = 2224
mode.size.x = 5120, desktop.size.x = 3420, mode.size.y = 3200, desktop.size.y = 2224
mode.size.x = 5120, desktop.size.x = 3420, mode.size.y = 3328, desktop.size.y = 2224
(Empty std::vector of modes returned)
```

This appears to be related to the question about whether Apple's `CGDisplayModeGetWidth()` function returns its value in points or pixels.

## Your environment

* OS:  MacOS 13.1 "Ventura"
* Version of SFML: I've tried 3.0 and 2.5.x branches
* Apple clang version 14.0.0 (clang-1400.0.29.202)
* Special compiler flags used: just the usual `-std=c++17 -g -Wall -Wextra -Werror -Wpedantic`

## Steps to reproduce

This is using SFML 3.0 but a similar program for 2.5.x exhibits the same behaviour. Note, this issue is on MacOS only.

```cpp
#include <SFML/Graphics.hpp>
#include <SFML/Window.hpp>

#include <iostream>
#include <vector>

int main()
{
    std::vector<sf::VideoMode> modes = sf::VideoMode::getFullscreenModes();
    if (modes.empty())
    {
        std::cout << "Empty set of modes returned\n";
    }
    else
    {
        for (std::size_t i = 0; i < modes.size(); ++i)
        {
            sf::VideoMode mode = modes[i];
            std::cout << "Mode #" << i << ": " << mode.size.x << "x" << mode.size.y << " - " << mode.bitsPerPixel
                      << " bpp" << std::endl;
        }
    }

    sf::RenderWindow win(sf::VideoMode(sf::Vector2u{800, 600}), "Test", sf::Style::Fullscreen);
    return 0;
}

```

## Expected behavior

We should not see a segfault, but SFML is assuming it will always have a valid mode as a default (after it states "The requested video mode is not available, switching to a valid mode") - but `getFullscreenModes()` is returning an empty vector, so it segfaults.

## Actual behavior

Segfault owing to the empty vector.

Do please let me know if anything needs clarification.

ChrisThrasher:
## Description

Attempt to fix #2300. Some quick local tests show that this does prevent crashing when trying to make a fullscreen window on macOS.

The equivalent fix for v3 is a bit simpler. https://github.com/SFML/SFML/compare/master...macos_fullscreen

### DIFF
--- a/src/SFML/Window/OSX/SFWindowController.mm
+++ b/src/SFML/Window/OSX/SFWindowController.mm
@@ -181,7 +181,6 @@
 {
     // Create a screen-sized window on the main display
     sf::VideoMode desktop = sf::VideoMode::getDesktopMode();
-    sf::priv::scaleInWidthHeight(desktop, nil);
     NSRect windowRect = NSMakeRect(0, 0, desktop.width, desktop.height);
     m_window = [[SFWindow alloc] initWithContentRect:windowRect
                                            styleMask:NSBorderlessWindowMask
@@ -448,7 +447,6 @@
         // Special case when fullscreen: only resize the opengl view
         // and make sure the requested size is not bigger than the window.
         sf::VideoMode desktop = sf::VideoMode::getDesktopMode();
-        sf::priv::scaleInWidthHeight(desktop, nil);
 
         width = std::min(width, desktop.width);
         height = std::min(height, desktop.height);
--- a/src/SFML/Window/OSX/Scaling.h
+++ b/src/SFML/Window/OSX/Scaling.h
@@ -58,13 +58,6 @@ void scaleIn(T& in, id<WindowImplDelegateProtocol> delegate)
 }
 
 template <class T>
-void scaleInWidthHeight(T& in, id<WindowImplDelegateProtocol> delegate)
-{
-    scaleIn(in.width, delegate);
-    scaleIn(in.height, delegate);
-}
-
-template <class T>
 void scaleInXY(T& in, id<WindowImplDelegateProtocol> delegate)
 {
     scaleIn(in.x, delegate);
--- a/src/SFML/Window/OSX/cg_sf_conversion.mm
+++ b/src/SFML/Window/OSX/cg_sf_conversion.mm
@@ -28,6 +28,7 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Window/OSX/cg_sf_conversion.hpp>
 #include <SFML/System/Err.hpp>
+#include <SFML/System/Vector2.hpp>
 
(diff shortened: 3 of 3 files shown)


Lorenzooone:
Just going to add my two cents here from the testing of a couple of users:
CGDisplayModeGetWidth seems to have the issue, on some setups, of returning 1x scaled resolution for DesktopMode, and 2x scaled resolutions for FullscreenModes. This makes them all incompatible.
From my tests, CGDisplayModeGetPixelWidth actually always returns all the resolutions with the same scaling factor applied to them. That can be 2x or 1x, depending on the setup, but the benefit is that they're always the same.
This prevents having empty vectors returned, or a DesktopMode which is not accepted because it's not present in the FullscreenModes (the latter being why most times applications crashed even when using stuff like NSHighResolutionCapable:false).

The "big" standing issue with this is that if the 2x scaling factor is applied, it will create Fullscreen modes which are too big for the screen, so they'll end up outsizing the screen, and being off-center.
If this can be somehow fixed, it would be amazing.

The PR in general is already an improvement, though.

ChrisThrasher:
I incorporated the changes from #3193 and tested again. All seems to work just fine on my high DPI MacBook. I don't have access to a low DPI screen to test on though.

@Lorenzooone Can you test this PR again now that I've merged your changes? If it's working then I think we can go ahead and merge this.

Lorenzooone:
Seems to work fine with both the HiDPI and the non-HiDPI tests! :+1:

eXpl0it3r (approved):

PR #3159, opened 2024-07-13
Specify `sf::Image`'s save format via a scoped enumeration

### LINKED ISSUE #3157: Suggestion: `sf::Image::saveToMemory` should accept the format as `enum class`
### Prerequisite Checklist

Currently we have:

```cpp
class Image 
{
    // ...
    [[nodiscard]] std::optional<std::vector<std::uint8_t>> saveToMemory(std::string_view format) const;
    // ...
};
```

Users can specify any string in `format`, including empty strings, invalid formats, or even invalid memory.

I suggest:

```cpp
class Image 
{
    // ...
    enum class [[nodiscard]] SaveFormat
    {
        BMP,
        TGA,
        PNG,
        JPG
    };

    [[nodiscard]] std::vector<std::uint8_t> saveToMemory(SaveFormat format) const;
    // ...
};
```

This change would: 

- Prevents misuse of `sf::Image::saveToMemory` at compile-time
- Removes `std::optional` from `sf::Image::saveToMemory`'s interface if we assume `*this` is a valid image *(BTW, this can also be ensured by changing `sf::Image`'s constructor to a factory function that validates size and pixels)*
- Simplifies `sf::Image::saveToMemory`'s implementation
- Simplifies `sf::Image` tests
- Removes compile time dependency on `<string_view>` from `<SFML/Graphics/Image.hpp>` 

I can provide a PR if there is agreement on this suggestion.

### Use Cases

.

### API Example

_No response_

ChrisThrasher:
## Description

Closes #3157 

The rationale in Vittorio's issue lays out the value case for this change so I won't repeat that.

The one place this PR differs from Vittorio's issue is that I chose to not change the return type of `sf::Image::saveToMemory` since the `stbi_write_` functions are prone to fail and to avoid having to make bigger changes to `sf::image`. This PR does not aim to change what we deem a valid image nor change the semantics of `sf::Image::Image`. Such a change can be discussed in a separate PR like #3152.

---

We have one other `saveToFile` function in `sf::SoundBuffer::saveToFile` but that function seems unfit for such a change since it supported an open set of file formats rather than the closed set of formats that `sf::Image::saveToFile` supports.

### DIFF
--- a/src/SFML/Graphics/Image.cpp
+++ b/src/SFML/Graphics/Image.cpp
@@ -317,44 +317,34 @@ bool Image::loadFromStream(InputStream& stream)
 
 
 ////////////////////////////////////////////////////////////
-bool Image::saveToFile(const std::filesystem::path& filename) const
+bool Image::saveToFile(const std::filesystem::path& filename, SaveFormat format) const
 {
     // Make sure the image is not empty
     if (!m_pixels.empty() && m_size.x > 0 && m_size.y > 0)
     {
-        // Deduce the image type from its extension
+        const Vector2i convertedSize = Vector2i(m_size);
 
-        // Extract the extension
-        const std::filesystem::path extension     = filename.extension();
-        const Vector2i              convertedSize = Vector2i(m_size);
-
-        if (extension == ".bmp")
+        switch (format)
         {
-            // BMP format
-            if (stbi_write_bmp(filename.string().c_str(), convertedSize.x, convertedSize.y, 4, m_pixels.data()))
-                return true;
-        }
-        else if (extension == ".tga")
-        {
-            // TGA format
-            if (stbi_write_tga(filename.string().c_str(), convertedSize.x, convertedSize.y, 4, m_pixels.data()))
-                return true;
-        }
-        else if (extension == ".png")
-        {
-            // PNG format
-            if (stbi_write_png(filename.string().c_str(), convertedSize.x, convertedSize.y, 4, m_pixels.data(), 0))
-                return true;
-        }
-        else if (extension == ".jpg" || extension == ".jpeg")
-        {
-            // JPG format
-            if (stbi_write_jpg(filename.string().c_str(), convertedSize.x, convertedSize.y, 4, m_pixels.data(), 90))
-                return true;
-        }
-        else
-        {
-            err() << "Image file extension " << extension << " not supported\n";
+            case SaveFormat::BMP:
+                if (stbi_write_bmp(filename.string().c_str(), convertedSize.x, convertedSize.y, 4, m_pixels.data()))
+                    return true;
+                break;
+            case SaveFormat::TGA:
+                if (stbi_write_tga(filename.string().c_str(), convertedSize.x, convertedSize.y, 4, m_pixels.data()))
+                    return true;
+                break;
+            case SaveFormat::PNG:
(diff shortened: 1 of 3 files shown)


src/SFML/Graphics/Image.cpp
```diff
@@ -260,44 +260,31 @@ std::optional<Image> Image::loadFromStream(InputStream& stream)
...
+                    return true;
+                break;
+            case SaveFormat::PNG:
+                if (stbi_write_png(filename.string().c_str(), convertedSize.x, convertedSize.y, 4, m_pixels.data(), 0))
+                    return true;
+                break;
+            case SaveFormat::JPG:
+                if (stbi_write_jpg(filename.string().c_str(), convertedSize.x, convertedSize.y, 4, m_pixels.data(), 90))
+                    return true;
+                break;
```
eXpl0it3r, line 287: Despite being exhaustive, we'd still need to handle the default case, as `static_cast<sf::Image::SaveFormat>(42)` will compile just fine.

test/Graphics/Image.test.cpp
```diff
@@ -190,44 +190,32 @@ TEST_CASE("[Graphics] sf::Image")
...
-        {
-            CHECK(!image.saveToFile("wheresmyextension"));
-            CHECK(!image.saveToFile("pls/add/extension"));
-        }
-
-        SECTION("Invalid extension")
-        {
-            CHECK(!image.saveToFile("test.ps"));
-            CHECK(!image.saveToFile("test.foo"));
-        }
```
eXpl0it3r, line 209: We could still have an invalid extension test: `CHECK(!image.saveToFile("test", static_cast<sf::Image::SaveFormat>(42)));`

test/Graphics/Image.test.cpp
```diff
@@ -190,44 +190,32 @@ TEST_CASE("[Graphics] sf::Image")
...
+            CHECK(!sf::Image({0, 10}, sf::Color::Magenta).saveToFile("test.jpg", sf::Image::SaveFormat::JPG));
         }
 
         const sf::Image image({256, 256}, sf::Color::Magenta);
 
-        SECTION("No extension")
-        {
-            CHECK(!image.saveToFile("wheresmyextension"));
-            CHECK(!image.saveToFile("pls/add/extension"));
-        }
```
eXpl0it3r, line 203: No extension is still a valid test case, but now it should pass, given the passed SaveFormat

src/SFML/Graphics/Image.cpp
ChrisThrasher: If anything we ought to `assert` against that. It's a pretty unambiguous bug if you someone manage to use an invalid enum value. If you want we can add a `default` case with that assertion.

eXpl0it3r: I think an assert should be fine, then we wouldn't need a default case.

ChrisThrasher: I'm not sure where to add that `assert` without adding a `default` case which I was hoping to avoid. Where are you picturing we add the `assert`?

eXpl0it3r: I'd have put them at the beginning of the function like any other argument checks. If you pass in invalid argument values, the function should fail in precondition asserts.

ChrisThrasher:

ChrisThrasher:
This is an interesting idea but I think we have to be strict with ourselves about cutting off API changes so late in the v3 development process so I will be closing this PR.

PR #3161, opened 2024-07-15
Proof-of-concept `sf::Texture` "recycling" API

vittorioromeo:
@binary1248's #3152 allows for "efficient object recycling after construction". 

Such feature can coexist with `std::optional`-based factory construction, and can coexist with the principle of having SFML objects always in a valid state. This PR is a proof-of-concept for that, only for `sf::Texture`.

Changes:

- Renamed factory functions to `createFromXXX`
- Added a recycling function for every factory function, named `recreateFromXXX`
- Factory functions are implemented in terms of recycling functions

I'm still not convinced that this is worthwhile doing, but it's definitely worth exploring. The advantage of "recycling" compared to assignment is that internal handles/buffers can be reused. In other words:

```cpp
auto t0 = sf::Texture::createFromFile("somepath").value();
t0 = sf::Texture::createFromFile("somepath").value(); 
    // Semantically correct, simple, but cannot reuse resources such as GL texture
    // handle, or internal containers (will need to be reallocated and moved)
    
auto t1 = sf::Texture::createFromFile("somepath").value();
if (!t1.recreateFromFile("somepath")) { /* ... */ }
    // Allows internal resources to be reused (e.g. no need to create a new GL 
    // texture, internal containers can be cleared instead of being reallocated)
```

Reasons why I am not convinced this is worthwhile doing:
1. Increases the surface area of SFML objects API
2. Somewhat increases complexity of SFML object implementation
3. Users now have to consciously choose between `operator=` or `.recreateFromXXX`
4. No evidence that recycling is meaningful/useful in any real application

Point (4) is the most important. If a program's performance bottleneck is texture loading, image creation, font loading, soundbuffer loading, and so on... there's something severely wrong. 

These things should not be done on a per-frame basis, resources should be loaded sparingly and only when needed. 

Furthermore, the overhead of using the file system to read an image file is going to be much larger than the one for calling `glGenTextures` or allocating a `std::vector<std::uint8_t>` for an image's pixels.

Therefore, I think that `operator=` is good enough in most cases, even if not optimal, but I can be convinced otherwise if provided with compelling evidence.

### DIFF
--- a/doc/mainpage.hpp
+++ b/doc/mainpage.hpp
@@ -21,7 +21,7 @@
 ///     sf::RenderWindow window(sf::VideoMode({800, 600}), "SFML window");
 ///
 ///     // Load a sprite to display
-///     const auto texture = sf::Texture::loadFromFile("cute_image.jpg").value();
+///     const auto texture = sf::Texture::createFromFile("cute_image.jpg").value();
 ///     sf::Sprite sprite(texture);
 ///
 ///     // Create a graphical text to display
--- a/examples/android/app/src/main/jni/main.cpp
+++ b/examples/android/app/src/main/jni/main.cpp
@@ -86,7 +86,7 @@ int main(int argc, char* argv[])
     sf::RenderWindow window(screen, "");
     window.setFramerateLimit(30);
 
-    const auto texture = sf::Texture::loadFromFile("image.png").value();
+    const auto texture = sf::Texture::createFromFile("image.png").value();
 
     sf::Sprite image(texture);
     image.setPosition(sf::Vector2f(screen.size) / 2.f);
--- a/examples/cocoa/CocoaAppDelegate.mm
+++ b/examples/cocoa/CocoaAppDelegate.mm
@@ -54,7 +54,7 @@ struct SFMLmainWindow
     sf::RenderWindow      renderWindow;
     sf::Font              font{sf::Font::openFromFile(resPath / "tuffy.ttf").value()};
     sf::Text              text{font};
-    sf::Texture           logo{sf::Texture::loadFromFile(resPath / "logo.png").value()};
+    sf::Texture           logo{sf::Texture::createFromFile(resPath / "logo.png").value()};
     sf::Sprite            sprite{logo};
     sf::Color             background{sf::Color::Blue};
 };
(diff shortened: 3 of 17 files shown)


Bambo-Borris:
You've missed some key details here that should be considered:

Games may often stream textures from disk, regularly loading and unloading assets during a scene. This is more common in 3D or large 2D games. Texture reuse may be preferable under the requirements of 3D games.

Yes, the loading from disk is the larger concern but the one aspect you can perform on a thread with minimal setup (with regards to graphics objects) is to load from disk -> image. The loading to GPU must be done on the thread with the active GL context. So in fact it becomes possible to bottleneck your core loop on loading textures to the GPU.

vittorioromeo:
@Bambo-Borris: thanks for the additional context/information!

> Texture reuse may be preferable under the requirements of 3D games.

When you're talking about "texture reuse", are you referring to reusing the same texture ID without generating a new one?

> Yes, the loading from disk is the larger concern but the one aspect you can perform on a thread with minimal setup (with regards to graphics objects) is to load from disk -> image. The loading to GPU must be done on the thread with the active GL context. So in fact it becomes possible to bottleneck your core loop on loading textures to the GPU.

This is useful information, but I do not understand how it is relevant. What I mean is: would having or not having a recycling API change anything regarding what you wrote above?

Bambo-Borris:
> When you're talking about "texture reuse", are you referring to reusing the same texture ID without generating a new one?

If someone is utilising SFML 2.x in a 3D application and was relying on the ability to load their textures via `sf::Texture`, then yes, they may have become genuinely reliant on this behaviour since it allowed them to avoid a texture rebind (which does enough computation to make avoiding a rebind preferable). 


> This is useful information, but I do not understand how it is relevant. What I mean is: would having or not having a recycling API change anything regarding what you wrote above?

Was more responding to your point about the overhead on disk loading which is somewhat possible to mitigate the severity of, and that the concern of loading into VRAM is still one that developers must consider.


As we discussed in Discord, I think that the reloading by calling `create` again should invalidate the previous IDs and force a user to rebind. But maybe allowing the old behaviour which mitigates a rebind is something we can leave open here so that people who are producing more complex applications with more constrained performance limits can benefit here. Recycling allows greater scope for optimization for those users.

PR #3164, opened 2024-07-15
Modernize memory management of Core Foundation pointers

ChrisThrasher:
## Description

Follows the same pattern as https://github.com/SFML/SFML/pull/2540, https://github.com/SFML/SFML/pull/2759, https://github.com/SFML/SFML/pull/2801, and https://github.com/SFML/SFML/pull/3149. 

Core Foundation pointers are used in a slightly more complicated way than in the previously cited PRs. In some cases `CFRetain` is called on a given pointer. I don't entirely understand how retains and releases work in the Apple Objective-C world so I chose not to touch pointers which are retained.

I believe this fixes a memory leak here:

https://github.com/SFML/SFML/blob/5eb57ae5f103947ab5ff4ad244d9c3fb640a3422/src/SFML/Window/macOS/HIDInputManager.mm#L771-L776

If `underlying` is non-null and `CFArrayGetCount(underlying)` returns zero then we exit the function without releasing `underlying`. This PR's new RAII-based approach guarantees that `underlying` gets freed.

---

Edit: I tried improving upon the pointers that did use `CFRetain` now that I better understand what retains and releases are. I'm still not entirely what I wrote is correct so this PR will have to sit for a bit longer before I'm ready to mark it as ready for review.

### DIFF
--- a/src/SFML/Window/macOS/HIDInputManager.hpp
+++ b/src/SFML/Window/macOS/HIDInputManager.hpp
@@ -30,6 +30,7 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Window/JoystickImpl.hpp>
 #include <SFML/Window/Keyboard.hpp>
+#include <SFML/Window/macOS/Utils.hpp>
 
 #include <IOKit/hid/IOHIDDevice.h>
 #include <IOKit/hid/IOHIDManager.h>
@@ -40,7 +41,7 @@
 namespace sf::priv
 {
 
-using IOHIDElements = std::vector<IOHIDElementRef>;
+using IOHIDElements = std::vector<CFPtr<IOHIDElementRef>>;
 
 ////////////////////////////////////////////////////////////
 class HIDInputManager
@@ -81,7 +82,7 @@ public:
     /// \return a retained CFDictionaryRef
     ///
     ////////////////////////////////////////////////////////////
-    static CFDictionaryRef copyDevicesMask(std::uint32_t page, std::uint32_t usage);
+    static CFPtr<CFDictionaryRef> copyDevicesMask(std::uint32_t page, std::uint32_t usage);
 
     ////////////////////////////////////////////////////////////
     /// \brief Try to convert a character into a SFML key code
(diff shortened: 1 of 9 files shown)


src/SFML/Window/macOS/HIDInputManager.hpp
```diff
@@ -40,7 +41,7 @@
 namespace sf::priv
 {
 
-using IOHIDElements = std::vector<IOHIDElementRef>;
+using IOHIDElements = std::vector<CFPtr<__IOHIDElement>>;
```
eXpl0it3r, line 44: I don't know much about Obj-C, but should we really be using types that start with `__`?

ChrisThrasher: I was hesitant to use types with those names too. Does Apple not want users to directly use such types? I'm not sure. A much more verbose alternative would be `std::remove_pointer_t<IOHIDElementRef>`.

eXpl0it3r: I'd probably go with that. Not a fan of using "reserved"/internal symbols

ChrisThrasher:
Moving this to the v3.1 milestone. I want to finish this eventually but for now it's not a priority to land in v3.0.

WarlockD:
Just curious but could this be wrapped in a simple std::shared_ptr type class?  Hell even unique_ptr can be given a function (or in C++17 case a class) that could call release once its done with the pointer.

eXpl0it3r:
Anything speaking against merging this at this stage?

ChrisThrasher:
I haven’t looked at this code in ages. I think it’s mostly good but idk if it’s complete (I might still have more places I want to use CFPtr) and idk if it’s correct (I don’t recall doing any testing beyond letting CI run). If you’re keen on this for v3.1 I can take another look at it to jog my memory as to whether more work is required.

eXpl0it3r:
3.1 is as good as any other. I think it's also fine if we expand the usage later on.

Would be good if you could at least run an example on your macOS system that touches the changeset.

PR #3165, opened 2024-07-16
Use `struct` when all type members are public

ChrisThrasher:
## Description

This follows an existing precedent where any type that exclusively contains `public` functions and data is marked as a `struct`. At this point I think we've found all the types that are entirely public but please let me know if I missed any.

If one day we change our mind and want to add private data, we can do that without issue so this change doesn't preclude any future change we may want to make.

https://github.com/SFML/SFML/commit/73a5a7382df0a82fb74bc8ffcca60eac8d3416a5
https://github.com/SFML/SFML/pull/2590
https://github.com/SFML/SFML/pull/2889
https://github.com/SFML/SFML/pull/2898


eXpl0it3r:
I thought the policy was to use `struct` for pure data-structures, i.e. when they have no member functions or constructors.

ChrisThrasher:
> I thought the policy was to use `struct` for pure data-structures, i.e. when they have no member functions or constructors.

I saw past precedent like `sf::RenderStates` being a struct with constructors so that's what motivated me. My rationale was as simple as "if everything is public use struct". Ultimately this is not a very important PR so if you're not a fan of this change then that's no problem with me. While we could theoretically sit down and construct a robust rationale for using `class` vs `struct` I'll admit that's not a great use of our time.

ZXShady:
will this be documented? so users who used to forward declare them on MSVC for example won't get crazy linker errors

eXpl0it3r:
Yes, we'll be writing a migration guide.

ChrisThrasher:
I don't think this PR is important enough to spend time debating in time to land before v3 ships.

PR #3171, opened 2024-07-20
Fix event handling overloadSet stack corruption on MSVC compiler

ZXShady:
## Description

A fix to stack corruption due to incorrect compiler behavior from MSVC that it will never fix due to backward compatability

## Tasks

-   [N] Tested on Linux
-   [Y] Tested on Windows
-   [N] Tested on macOS
-   [N] Tested on iOS
-   [N] Tested on Android

## How to test this PR?

Compile with MSVC (in debug mode) run examples/event_handling and press Enter until you switch to Overload it will instantly crash saying stack corruption around variable overloadSet in src/WindowBase.inl this commit is to fix this bug by making sure no base classes are empty in the sf::priv::OverloadSet for MSVC only since MSVC does not do the optimizations correctly but leaves Clang/GCC alone.

### DIFF
--- a/examples/event_handling/EventHandling.cpp
+++ b/examples/event_handling/EventHandling.cpp
@@ -1,7 +1,12 @@
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
-#include <SFML/Graphics.hpp>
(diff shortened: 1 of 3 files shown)


eXpl0it3r:
Superseded by #3172

PR #3176, opened 2024-07-25
Fix issue with openal32 dll for Windows ARM64 requiring vc_redist

Lorenzooone:
## Description

Fixes an issue where the openal32 dll required the vc_redistributables to be installed in the system on ARM64, but not on x64 and x86. With this, the ARM64 dll also does not require the vc_redistributables to be installed.

This is important because otherwise building with SFML_USE_STATIC_STD_LIBS on Windows ARM64 wouldn't really do much, as the default openal32 dll would require the vc_redistributables anyways. With this change, one can build applications which do not require the users to install anything even on ARM64 (when also shipping the openal32 dll), like it already was on x86 and x64.

This PR is related to: https://github.com/eXpl0it3r/SFML-dependencies/pull/15

## Tasks

-   [X] Tested on Windows

## How to test this PR?

Tested on a clean ARM64 Windows VM. Build a minimal application with SFML_USE_STATIC_STD_LIBS set to TRUE. 
Before the changes, the vc_redistributables were required for the user. Now, they aren't.


ChrisThrasher:

eXpl0it3r (approved):

eXpl0it3r:
No idea why we still got to macos-11 builds in the action list, given that we've updated all the images to macos-12 as far as I know.

I'll just ignore those, as the image has been deprecated and is unlikely to build.

eXpl0it3r:
Thanks for fixing this! 🥳

Lorenzooone:
> No idea why we still got to macos-11 builds in the action list, given that we've updated all the images to macos-12 as far as I know.
> 
> I'll just ignore those, as the image has been deprecated and is unlikely to build.

Master is what got updated. You should update the ci.yml file in 2.6.x as well. 👍

ChrisThrasher:
> > No idea why we still got to macos-11 builds in the action list, given that we've updated all the images to macos-12 as far as I know.
> > I'll just ignore those, as the image has been deprecated and is unlikely to build.
> 
> Master is what got updated. You should update the ci.yml file in 2.6.x as well. 👍

https://github.com/SFML/SFML/pull/3182

eXpl0it3r:
I updated 2.6.x, but I guess I missed those. 😅

Thanks!

PR #3186, opened 2024-07-31
Prevent crash when setting empty uniform array

### LINKED ISSUE #3185: setUniformArray with vector crash when buffer is empty even when size is 0
### Prerequisite Checklist

### Describe your issue here

image
I have this crash and error when using setUniformArray on a shader with a sf::Glsl::Vec2 type.
it happens when the pointer I pass is empty, so it would be normal to have an read access violation, but I still gave the size 0, so it should not be accessing data in my array

### Your Environment

- OS / distro / window manager: Windows
- SFML version: 2.6.1
- Compiler / toolchain: Cmake, MSVC 2019
- Special compiler / CMake flags: SFML linked statically

### Steps to reproduce

```cpp
#include <SFML/Graphics.hpp>
#include <vector>

int main() {
	sf::RectangleShape shape;
	sf::RenderWindow window(sf::VideoMode(800, 600), "My window");
	sf::Shader shader;
	shader.loadFromFile("test.frag", sf::Shader::Type::Fragment);
	sf::RenderStates state(&shader);

	std::vector<sf::Glsl::Vec2> uniformArray;
	shader.setUniformArray("testUniform", uniformArray.data(), 0);

	while (window.isOpen()) {
		sf::Event event;
				window.close();
		}

		window.clear();
		window.draw(shape, state);
		window.display();
	}

	return 0;
}
```

### Expected behavior

the function should verify that the given size is greater than 0 before accessing data in my array

### Actual behavior

the function just acces the element 0 of they array before doing any check.

ChrisThrasher:
## Description

Closes #3185

### DIFF
--- a/src/SFML/Graphics/Shader.cpp
+++ b/src/SFML/Graphics/Shader.cpp
@@ -603,7 +603,7 @@ void Shader::setUniformArray(const std::string& name, const Glsl::Vec2* vectorAr
 
     UniformBinder binder(*this, name);
     if (binder.location != -1)
-        glCheck(GLEXT_glUniform2fv(binder.location, static_cast<GLsizei>(length), &contiguous[0]));
+        glCheck(GLEXT_glUniform2fv(binder.location, static_cast<GLsizei>(length), !contiguous.empty() ? &contiguous[0] : NULL));
 }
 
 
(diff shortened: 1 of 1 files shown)


ChrisThrasher:
Retargeting 2.6.x since this crash does not apparently occur in v3.

kimci86:
I think the fix could be a simple as replacing `&contiguous[0]` by `length ? &contiguous[0] : NULL`.

We use this pattern in other places like here:
https://github.com/SFML/SFML/blob/779b96353017c884212b6d54ad20ddc77dc5e25a/src/SFML/Network/Packet.cpp#L85

binary1248 (approved):

ChrisThrasher:
The author of #3185 confirmed that this PR fixes their problem and that their problem is not present in SFML 3 so this commit does not need to be backmerged into `master`.

Issue #3181, opened 2024-07-27
ALooper_pollAll is deprecated on Android NDK 27


Zombieschannel:
### Prerequisite Checklist

- [X] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [X] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [X] I am here to report an issue and not to just ask a question or look for help (use the [forum](https://en.sfml-dev.org/forums/index.php#c3) or [Discord](https://discord.gg/nr4X7Fh) instead)

### Describe your issue here

Basically everything [here](https://github.com/raysan5/raylib/issues/4166) applies to SFML (2 & 3) as well. In src/SFML/Window/Android/SensorImpl.cpp in the update function there is this line: `ALooper_pollAll(0, nullptr, nullptr, nullptr);` which seems to be deprecated. NDK 26 still compiles SFML fine, but NDK 27 throws out this error:
```
error: 'ALooper_pollAll' is unavailable: obsoleted in Android 1 - ALooper_pollAll may ignore wakes. Use ALooper_pollOnce instead. See The API documentation for more information
```

### Your Environment

SFML version: 2.6.1
Ninja
Android studio 2024.1.1
NDK: 27.0.11902837

### Steps to reproduce

Try to compile SFML for Android with NDK 27.

### Expected behavior

Compiles fine.

### Actual behavior

error: 'ALooper_pollAll' is unavailable: obsoleted in Android 1

ChrisThrasher:
https://github.com/ChrisThrasher/SFML/actions/runs/10221306711/job/28283554085

I'm starting to see this sometimes fail CI jobs so we'll probably be forced to take action on this soon.

ChrisThrasher:
@Zombieschannel Could you test this out for me? https://github.com/SFML/SFML/pull/3188

ChrisThrasher:
Closed by #3188

PR #3188, opened 2024-08-05
Remove use of deprecated `ALooper_pollAll`

ChrisThrasher:
## Description

An attempt at fixing #3181

### DIFF
--- a/src/SFML/Window/Android/InputImpl.cpp
+++ b/src/SFML/Window/Android/InputImpl.cpp
@@ -166,7 +166,8 @@ void setVirtualKeyboardVisible(bool visible)
 ////////////////////////////////////////////////////////////
 bool isMouseButtonPressed(Mouse::Button button)
 {
-    ALooper_pollAll(0, nullptr, nullptr, nullptr);
+    while (ALooper_pollOnce(0, nullptr, nullptr, nullptr) >= 0)
+        ;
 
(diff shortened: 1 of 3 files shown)


eXpl0it3r:
Would also need to be backported to 2.6.2

src/SFML/Window/Android/InputImpl.cpp
```diff
@@ -166,7 +166,8 @@ void setVirtualKeyboardVisible(bool visible)
 ////////////////////////////////////////////////////////////
 bool isMouseButtonPressed(Mouse::Button button)
 {
-    ALooper_pollAll(0, nullptr, nullptr, nullptr);
+    while (ALooper_pollOnce(0, nullptr, nullptr, nullptr) != ALOOPER_POLL_CALLBACK)
+        ;
```
eXpl0it3r, line 170: Shouldn't this be the other way around?  
For as long as you get `ALOOPER_POLL_CALLBACK` you should continue processing, every other result should lead to a break.

What's curious is that the official NDK sample, doesn't do it in a loop: https://github.com/android/ndk-samples/pull/1008/files

Zombieschannel:
All errors/warnings are fixed!

ChrisThrasher:

eXpl0it3r (approved):

ChrisThrasher:

PR #3194, opened 2024-08-20
Use imported target for openal to simplify copying dll's

JonnyPtn:
One of the big problems people face (although now solved with openal Replaced for SFML3) is copying required dlls to the required location. A common and convenient way to do this is with the `TARGET_RUNTIME_DLLS` generator expression which works great for SFML dlls but not for openAL as it's not an imported target

Cmake's built-in module for finding openal creates an imported target but doesn't set the `IMPORTED_LOCATION` property, there's an issue for that here: https://gitlab.kitware.com/cmake/cmake/-/issues/24033. It's unlikely to be fixed any time soon there, as they have to consider that the dll next to the lib isn't guaranteed to be a match, however for SFML we bundle these dependencies so we can guarantee that

So this change adds a lightly modified copy of cmake's find module which sets the IMPORTED_LOCATION appropriately and also handles the different arch variations

the `sfml_find_package` macro doesn't really work with existing targets so I've just reverted to using the standard one which I believe is fine in this case - this all goes away for SFML 3 anyway so don't think we need to be too precious about it

### DIFF
--- a/cmake/Modules/FindOpenAL.cmake
+++ b/cmake/Modules/FindOpenAL.cmake
@@ -1,0 +1,134 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
+# file Copyright.txt or https://cmake.org/licensing for details.
+
+#[=======================================================================[.rst:
+FindOpenAL
+----------
+
+Finds Open Audio Library (OpenAL).
+
+Projects using this module should use ``#include "al.h"`` to include the OpenAL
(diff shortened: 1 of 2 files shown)


ChrisThrasher (approved):
https://github.com/SFML/SFML/pull/2666/files#diff-d0df7de42fe96428ddf8d50eb199423bf66318895af78b2726a7acc65e175526

This is something we already did in v3 before eventually removing OpenAL so there is precedent behind this change.

PR #3195, opened 2024-08-20
gitignore vs folder

JonnyPtn:
just noticed it wasn't ignored so tried to add a bunch of files - it's already ignored in master


ChrisThrasher:

PR #3196, opened 2024-08-20
Throw custom `sf::Exception` type

ChrisThrasher:
## Description

It seems wise to throw a custom exception type for a few minor reasons.

1. Allows users to disambiguate exceptions from SFML code from other sources
2. Gives us an abstraction upon which to add more functionality later. Perhaps in a future version we'll want to add something with `std::source_location` to this or maybe `std::stacktrace`.

### DIFF
--- a/cmake/CompilerWarnings.cmake
+++ b/cmake/CompilerWarnings.cmake
@@ -35,6 +35,7 @@ function(set_target_warnings target)
             /wd4068 # disable warnings about unknown pragmas (e.g. #pragma GCC)
             /wd4505 # disable warnings about unused functions that might be platform-specific
             /wd4800 # disable warnings regarding implicit conversions to bool
+            /wd4275 # disable warnings about exporting non DLL-interface classes
         )
     endif()
 
--- a/include/SFML/System/Exception.hpp
+++ b/include/SFML/System/Exception.hpp
@@ -27,29 +27,26 @@
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
+#include <SFML/System/Export.hpp>
 
-#include <SFML/Config.hpp>
-
-#include <SFML/System/Angle.hpp>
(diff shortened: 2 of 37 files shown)


cmake/CompilerWarnings.cmake
```diff
@@ -35,6 +35,7 @@ function(set_target_warnings target)
             /wd4068 # disable warnings about unknown pragmas (e.g. #pragma GCC)
             /wd4505 # disable warnings about unused functions that might be platform-specific
             /wd4800 # disable warnings regarding implicit conversions to bool
+            /wd4275 # disable warnings about exporting non DLL-interface classes
```
eXpl0it3r, line 38: The alternative and maybe better way (?) is to not export the Exception and just have the constructor defined in the header, see https://stackoverflow.com/a/48453230/1034248 or https://stackoverflow.com/a/24512237/1034248

I'm okay with both ways. Ignoring the warning globally could theoretically hide some other issues, but not sure we'd really run into those.

eXpl0it3r (changes requested):

ChrisThrasher:
Okay so lesson learned, if we want to `throw` a type across shared library boundaries, that type must be exported.

include/SFML/System/Exception.hpp
```diff
@@ -0,0 +1,54 @@
...
+public:
+    ////////////////////////////////////////////////////////////
+    /// \brief Construct with message
+    ///
+    /// \param message Exception message
+    ///
+    ////////////////////////////////////////////////////////////
+    explicit Exception(const std::string& message) : std::runtime_error(message)
+    {
+    }
```
eXpl0it3r, line 52: Should we keep the impl in the header?

ZXShady: I would say if there is not a reason to make it in the header then why not do it in the implementation file and imrpove compile times and be consistent

eXpl0it3r (approved):

PR #3197, opened 2024-08-21
Reimplement event assertions to loosen constexpr restrictions on event subtypes

ChrisThrasher:
## Description

The current implementation of these static assertions requires that `std::Event::m_data` be `constexpr` constructible. This transitively requires that all event subtypes are `constexpr` constructible as well. This restrictions currently causes no issues but it means that we cannot add future event subtypes that are not `constexpr` constructible.

To test this out, add a new event subtype containing a `std::string`, add it to `m_data`, then observe compilation fail.

### DIFF
--- a/include/SFML/Window/Event.hpp
+++ b/include/SFML/Window/Event.hpp
@@ -373,13 +373,13 @@ private:
     // Helper functions
     ////////////////////////////////////////////////////////////
     template <typename T, typename... Ts>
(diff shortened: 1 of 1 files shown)


vittorioromeo:
Whoever came up with the pointer idea is pretty clever!

eXpl0it3r (approved):

PR #3202, opened 2024-08-26
Fix incorrect clang-format documentation

ChrisThrasher:


ChrisThrasher:
As an aside it's probably time to upgrade to a newer clang-format release. I'll probably want to use version 18 since that's what Ubuntu 24 ships with. Version 14 is what Ubuntu 22 used so it made sense when we picked that years ago.

eXpl0it3r (approved):

PR #3190, opened 2024-08-14
Mark standard library types as code in doc strings

eXpl0it3r:
## Description

I regex-ed for `/// .* std::[^ ]*` and added backticks. Maybe there are more types that I've missed (e.g. standard types).

Partially solves #3123 

## How to test this PR?

Generate API docs and look at the output.

PR #3213, opened 2024-09-08
Use backticks for SFML types in doc comments

### LINKED ISSUE #3123: Use backticks to distinguish code from prose in Doxygen documentation
In a number of recent issues we've made a point to update the Doxygen documentation to use backticks around things like `std::null opt` or references to other SFML APIs to ensure that C++ constructors are differentiated from prose. I'm opening this issue to make a note to ourselves that we ought to audit the codebase to find more places where this pattern can be employed.

Here's a before and after to illustrate how this change improves the readability of the docs. The effect is subtle but useful and requires minimum changes to our code. Note how the style of `std::nullopt` has changed.

Screenshot 2024-06-23 at 9 52 51 AM

Screenshot 2024-06-23 at 9 52 29 AM

eXpl0it3r:
## Description

Similar to #3190, I added backticks to all the SFML types.  
Since this was manual work, there's a high chance that I've missed some spots.

I've also spotted a few inconsistencies and tried to align them, but might have also still missed some.

This PR closes #3123

## How to test this PR?

Generate the docs and ensure it looks correct.


include/SFML/Audio/Listener.hpp
```diff
@@ -61,13 +61,13 @@ struct Cone
 ////////////////////////////////////////////////////////////
 /// \brief Change the global volume of all the sounds and musics
 ///
-/// The volume is a number between 0 and 100; it is combined with
-/// the individual volume of each sound / music.
+/// The \a `volume` is a number between 0 and 100; it is combined
```
ChrisThrasher, line 64:
```suggestion
/// \a `volume` is a number between 0 and 100; it is combined
```

include/SFML/Audio/Listener.hpp
```diff
@@ -77,19 +77,19 @@ SFML_AUDIO_API void setGlobalVolume(float volume);
...
+/// \see `setGlobalVolume`
 ///
 ////////////////////////////////////////////////////////////
 [[nodiscard]] SFML_AUDIO_API float getGlobalVolume();
 
 ////////////////////////////////////////////////////////////
 /// \brief Set the position of the listener in the scene
 ///
-/// The default listener's position is (0, 0, 0).
+/// The default listener's position is `(0, 0, 0)`.
```
ChrisThrasher, line 88: I'm not sure if "(0, 0, 0)" counts as code. I suppose it count be interpreted as part of a bigger expression like `sf::Vector3(0, 0, 0)`.

include/SFML/Audio/Listener.hpp
```diff
@@ -211,11 +211,11 @@ SFML_AUDIO_API void setUpVector(const Vector3f& upVector);
 ///
 /// The audio listener defines the global properties of the
 /// audio environment, it defines where and how sounds and musics
-/// are heard. If sf::View is the eyes of the user, then sf::Listener
-/// is his ears (by the way, they are often linked together --
-/// same position, orientation, etc.).
+/// are heard. If `sf::View` is the eyes of the user, then
+/// `sf::Listener` is his ears (by the way, they are often linked
```
ChrisThrasher, line 215:
```suggestion
/// `sf::Listener` are their ears (by the way, they are often linked
```

ChrisThrasher (changes requested):
I like the premise of this PR and think the liberal use of monospacing will make the generated documentation easier to read.

ChrisThrasher:
I generated the docs and am clicking around to see what I can find. Here are some observations.

* "*this" is often not monospaced
* "True"/"true"/"False"/"false" is often not monospaced within `\return` lines
* Look for instances of "operator *", "operator ==" and other similar strings in the docs. Note the extra space after "operator" which I think should be removed
* There are other uses of stuff like "== operator" which I think ought to be "`operator==`".
* The following regex `///.* sf::` finds more uses of non-monospaced references to 1st party types. For example, InputSoundFile.hpp class docs at the bottom of the file.
* Glsl.hpp referes to types like `Mat3` without a `sf::` namespace so they missed the regexes you used to find them. I think that file ought to add `sf::` prefixes when referring to 1st party types.
* The `sf::ContextSettings` docs refer to data members. We could monospace those as well. I'm not sure how to easily find other instances of this occurring so doing this repo-wide will take a lot of manual inspection.

I'm not necessarily sure that all of these points should result in a change to this PR but I thought they were worth sharing nonetheless.

eXpl0it3r:
> * Glsl.hpp referes to types like `Mat3` without a `sf::` namespace so they missed the regexes you used to find them. I think that file ought to add `sf::` prefixes when referring to 1st party types.
> * [ ]  The `sf::ContextSettings` docs refer to data members. We could monospace those as well. I'm not sure how to easily find other instances of this occurring so doing this repo-wide will take a lot of manual inspection.

I feel like these take way too much time to figure out and there's little regex magic to apply. We can try to improve as we go or if someone is bored, they can tackle it, but I'd leave it as is for the current PR.

include/SFML/Audio/InputSoundFile.hpp
```diff
@@ -178,11 +178,11 @@ class SFML_AUDIO_API InputSoundFile
     /// \brief Get the map of position in sample frame to sound channel
     ///
     /// This is used to map a sample in the sample stream to a
-    /// position during spatialisation.
+    /// position during spatialization.
```
ChrisThrasher, line 181: This spelling is also found in AudioDevice.hpp

include/SFML/Window/WindowBase.hpp
```diff
@@ -604,14 +603,14 @@ class SFML_WINDOW_API WindowBase
 /// \class sf::WindowBase
 /// \ingroup window
 ///
-/// sf::WindowBase serves as the base class for all Windows.
+/// `sf::WindowBase` serves as the base class for all Windows.
 ///
 /// A sf::WindowBase can create its own new window, or be embedded into
```
ChrisThrasher, line 608:
```suggestion
/// An `sf::WindowBase` can create its own new window, or be embedded into
```
Should we also audit the docs for the use of "a" prior to an `sf::` type? I think it reads better to as "an" as the article.

ChrisThrasher (changes requested):
See also HIDInputManager.hpp:101 and HIDInputManager.hpp:125

ChrisThrasher:
Switching to macOS (and therefore upgrading to Doxygen 1.12) fixes the Doxygen error we were getting.

ChrisThrasher:
Adding backticks around "sf::Exception" does not work. Those backticks also result in a hyperlink to `sf::Exception` not being formed.

Screenshot 2024-09-11 at 1 31 13 PM

ChrisThrasher (approved):
I think in its current state this PR makes enough progress that it's ready to merge. We will keep iterating on the docs in subsequent PRs.

ChrisThrasher (approved):

PR #3217, opened 2024-09-10
Update v2.6.2 changelog

ChrisThrasher:
## Description

The changelog was missing a few important fixes

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,13 +1,34 @@
 # Changelog
 
-## Unreleased
+## Unreleased SFML 2.6.2
 
 ### General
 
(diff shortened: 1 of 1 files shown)


changelog.md
```diff
@@ -1,13 +1,34 @@
 # Changelog
 
-## Unreleased
+## SFML 2.6.2
```
eXpl0it3r, line 3: SFML 2.6.2 is still unreleased 😄

ChrisThrasher: I figured I would omit the "Unreleased" so that we have one fewer thing to change when we do release. What are the odds we forget to remove that word when we ship?

eXpl0it3r (approved):

PR #3225, opened 2024-09-12
Extend the migration guide regarding the sf::Text constructor change

eXpl0it3r:
## Description

Just noticed this small change and thought it was still worth mentioning.

```cpp
auto text = sf::Text{"Hello World", font};
```
vs
```cpp
auto test = sf::Text{font, "Hello World"};
```

### DIFF
--- a/migration.md
+++ b/migration.md
@@ -425,6 +425,7 @@ SFML 3 capitalizes the `A` of `aliasing` for all the APIs.
 
 SFML 3 includes various smaller changes that ought to be mentioned.
 
+* Changed the parameter order of the `sf::Text` constructor, so that the provided font is always the first parameter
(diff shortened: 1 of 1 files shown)


migration.md
```diff
@@ -425,6 +425,7 @@ SFML 3 capitalizes the `A` of `aliasing` for all the APIs.
 
 SFML 3 includes various smaller changes that ought to be mentioned.
 
+* Changed the parameter order of the `sf::Text` constructor, so that the passed-in font is always the first parameter
```
ChrisThrasher, line 428:
```suggestion
* Changed the parameter order of the `sf::Text` constructor, so that the provided font is always the first parameter
```

ChrisThrasher (changes requested):
Otherwise LGTM

ChrisThrasher (approved):

Issue #2843, opened 2023-12-17
`sf::WindowBase::pollEvent` triggers input monitoring prompt on macOS even though it's not needed


orazioedoardo:
### Prerequisite Checklist

- [X] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [X] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [X] I am here to report an issue and not to just ask a question or look for help (use the [forum](https://en.sfml-dev.org/forums/index.php#c3) or [Discord](https://discord.gg/nr4X7Fh) instead)

### Describe your issue here

As I understand, input monitoring permission is required when using `sf::Keyboard::isKeyPressed` since checking for key press makes possible to create a keylogger, however when [just watching for events](https://en.sfml-dev.org/forums/index.php?topic=27734.0) delivered to the process looking for `sf::Event::KeyPressed` events, no permission is needed.

I'm not familiar with SFML but by checking around with the code I believe that the permission prompt is triggered [here](https://github.com/SFML/SFML/blob/a474654ea7782d0e861e7a498a469f70da811702/src/SFML/Window/macOS/HIDInputManager.mm#L749) and the function involved is [IOHIDManagerCopyDevices](https://developer.apple.com/documentation/iokit/1438391-iohidmanagercopydevices). If I simply uncomment `HIDInputManager::initializeKeyboard` [here](https://github.com/SFML/SFML/blob/a474654ea7782d0e861e7a498a469f70da811702/src/SFML/Window/macOS/HIDInputManager.mm#L719) and recompile SFML I can run the minimal example with no permission prompt.

Perhaps the keyboard should not be initialized when receiving keyboard events? As it stands, it's bad UX to open an app or game and seeing a spurious prompt requesting eavesdropping capabilities for no apparent reason as soon as a key is pressed, all while keyboard input works anyway.

### Your Environment

- OS / distro / window manager: macOS 13
- SFML version: 2.6.1
- Compiler / toolchain: clang
- Special compiler / CMake flags: none

### Steps to reproduce

1. Compile and run the following code
2. Focus on the window
3. Press Esc on the keyboard

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode({1280, 720}), "Minimal, complete and verifiable example");
[...]

PR #3232, opened 2024-09-14
Use lazy loading for keyboard m_keys on macos

Lorenzooone:
## Description

On macos, initializing the keyboard's m_keys (a dictionary to convert scancodes to IOHIDElementRef) has the unfortunate side effect of triggering a request for "input monitor" permissions.
The m_keys dictionary is exclusively used when using isKeyPressed, which is a function that already triggers the input monitor permission request.
This change makes it so initialize_keyboard is only lazily called if isKeyPressed is called.

The benefit of this change is that using keyboard events on macos won't trigger permission requests, anymore (but isKeyPressed still will, like before).
initialize_keyboard is protected by a bool, so there are no risks of calling it more than needed (that is... once).

This PR fixes https://github.com/SFML/SFML/issues/2843

The change should be backported to 2.6.x.

## Tasks

-   [X] Tested on macOS

## How to test this PR?

Try running a simple program on macos (like the Island example) which uses events for keyboard inputs.
You should not see any permission request popup.

Then, try running another program on macos which uses isKeyPressed.
After confirming input monitoring is allowed through the popup, you should be able to use isKeyPressed just fine.

### DIFF
--- a/src/SFML/Window/macOS/HIDInputManager.hpp
+++ b/src/SFML/Window/macOS/HIDInputManager.hpp
@@ -286,7 +286,8 @@ private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    IOHIDManagerRef                                                       m_manager{}; ///< Underlying HID Manager
+    IOHIDManagerRef m_manager{};         ///< Underlying HID Manager
+    bool            m_keysInitialized{}; ///< Has initializeKeyboard been called at least once?
     EnumArray<Keyboard::Scancode, IOHIDElements, Keyboard::ScancodeCount> m_keys; ///< All the keys on any connected keyboard
     EnumArray<Keyboard::Key, Keyboard::Scancode, Keyboard::KeyCount> m_keyToScancodeMapping{}; ///< Mapping from Key to Scancode
     EnumArray<Keyboard::Scancode, Keyboard::Key, Keyboard::ScancodeCount> m_scancodeToKeyMapping{}; ///< Mapping from Scancode to Key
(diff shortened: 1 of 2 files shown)


eXpl0it3r:

src/SFML/Window/macOS/HIDInputManager.hpp
```diff
@@ -286,7 +286,8 @@ class HIDInputManager
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    IOHIDManagerRef                                                       m_manager{}; ///< Underlying HID Manager
+    IOHIDManagerRef m_manager{};           ///< Underlying HID Manager
+    bool            m_keysInitializeTried; // Has initializeKeyboard been called at least once?
```
ChrisThrasher, line 290:
```suggestion
    bool            m_hasInitializedKeys{}; // Has initializeKeyboard been called at least once?
```
Let's use a variable name that reads a bit more like English.

src/SFML/Window/macOS/HIDInputManager.mm
```diff
@@ -705,6 +711,7 @@
 HIDInputManager::HIDInputManager()
 {
     // Create an HID Manager reference
+    m_keysInitializeTried     = false;
```
ChrisThrasher, line 714:
```suggestion
```

ChrisThrasher (changes requested):
Glad to see someone take a shot at fixing this. This will be a good user experience win.

ChrisThrasher (approved):
I was just able to confirm this patch works on my machine. Thank you!

ChrisThrasher:

PR #3234, opened 2024-09-15
Replace `swap` with move where appropriate

vittorioromeo:
Should not have any effect on behavior -- these swaps seems remnant of C++03-like move semantics emulation.

### DIFF
--- a/src/SFML/Graphics/Image.cpp
+++ b/src/SFML/Graphics/Image.cpp
@@ -46,6 +46,7 @@
 #include <memory>
 #include <ostream>
 #include <string>
+#include <utility>
 
 #include <cassert>
 #include <cstring>
@@ -154,7 +155,7 @@ void Image::resize(Vector2u size, Color color)
         }
 
         // Commit the new pixel buffer
-        m_pixels.swap(newPixels);
+        m_pixels = std::move(newPixels);
 
         // Assign the new size
         m_size = size;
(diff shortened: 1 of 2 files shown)


src/SFML/Graphics/Image.cpp
```diff
@@ -154,7 +155,7 @@ void Image::resize(Vector2u size, Color color)
         }
 
         // Commit the new pixel buffer
-        m_pixels.swap(newPixels);
```
ChrisThrasher, line 157: > // Create a new pixel buffer first for exception safety's sake

This comment is somewhat tangential to the goal of your PR, but I read this comment earlier this year when working on something else and couldn't make sense of it. What about making a new buffer first has any affect on exception safety?

ChrisThrasher (approved):

src/SFML/Graphics/Image.cpp
vittorioromeo: Uhhh... I guess the idea was that if the memory allocation of `newPixels` ends up throwing an exception, the state of `Image::m_pixels` would be unchanged. 

Probably the initial implementation was something like `m_pixels.resiez(static_cast<std::size_t>(size.x) * static_cast<std::size_t>(size.y) * 4)`, but that should have the strong exception guarantee anyway, so I guess we could just do that?

ChrisThrasher: I would prefer a simpler implementation using `.resize` instead of doing a dance with extra intermediate vectors.

vittorioromeo: Unless we are missing something else, I agree. I'd be happy to review a PR for that.

PR #3235, opened 2024-09-16
Use lazy loading for keyboard scancodes on macos (backport)

Lorenzooone:
## Description

Backport of #3232 for 2.6.x.
To test (I will in a few minutes/hours... Or if you guys are faster, feel free to do it in the meantime).

## Tasks

-   [ ] Tested on macOS

## How to test this PR?

Try running a simple program on macos (like the Island example) which uses events for keyboard inputs.
You should not see any permission request popup.

Then, try running another program on macos which uses isKeyPressed.
After confirming input monitoring is allowed through the popup, you should be able to use isKeyPressed just fine.

(Same as #3232 )

### DIFF
--- a/src/SFML/Window/OSX/HIDInputManager.hpp
+++ b/src/SFML/Window/OSX/HIDInputManager.hpp
@@ -281,6 +281,7 @@ private:
     // Member data
     ////////////////////////////////////////////////////////////
     IOHIDManagerRef    m_manager;                                             ///< Underlying HID Manager
+    bool               m_keysInitialized;                                     ///< Has initializeKeyboard been called at least once?
     IOHIDElements      m_keys[Keyboard::Scan::ScancodeCount];                 ///< All the keys on any connected keyboard
     Keyboard::Scancode m_keyToScancodeMapping[Keyboard::KeyCount];            ///< Mapping from Key to Scancode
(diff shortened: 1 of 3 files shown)


src/SFML/Window/OSX/HIDInputManager.hpp
```diff
@@ -281,6 +281,7 @@ class HIDInputManager : NonCopyable
     // Member data
     ////////////////////////////////////////////////////////////
     IOHIDManagerRef    m_manager;                                             ///< Underlying HID Manager
+    bool               m_keysInitialized = false;                             ///< Has initializeKeyboard been called at least once?
```
ChrisThrasher, line 284: https://en.cppreference.com/w/cpp/language/data_members

NSDMI (yay for C++ initialisms) is a C++11 feature so we'll have to initialize this member in the construct initializer list instead.

ChrisThrasher (changes requested):

src/SFML/Window/OSX/HIDInputManager.hpp
Lorenzooone: Oh shit! Had no idea... Okay, easy fix...

ChrisThrasher (approved):

ChrisThrasher (approved):

ChrisThrasher:

ChrisThrasher (approved):

PR #3238, opened 2024-09-19
Add CapsLock, NumLock and ScrollLock support for Events

eXpl0it3r:
## Description

This adds support for getting CapsLock, NumLock and ScrollLock modifier/state information on the events.

If this is accepted, once could also look into adding some functions to `sf::Keyboard`.

This came up as discussion on Discord, which originated from a TGUI issue: https://github.com/texus/TGUI/issues/245

Note: ScrollLock isn't really a concept that macOS has implemented, as such it's always false.

I also looked into the function (Fn) key and while most OS actually has a filter for it, Windows doesn't and since we don't have a Scancode for it, it would be a bit a special case. Additionally, a lot of keyboard firmware won't even trigger as normal key input on the Fn key, so it might not be worth it.

## Tasks

-   [x] Tested on Linux
-   [x] Tested on Windows
-   [x] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    auto window = sf::RenderWindow{ sf::VideoMode({1280, 720}), "Minimal, complete and verifiable example" };
    window.setFramerateLimit(60);

    auto font = sf::Font{ "C:\\Windows\\Fonts\\Arial.ttf" }; // Change me
    auto text = sf::Text{ font };

    while (window.isOpen())
    {
        {
            {
                window.close();
            }
            else if (auto keyPressed = event->getIf<sf::Event::KeyPressed>())
            {
                auto output = std::string{ "Key Pressed: " + sf::Keyboard::getDescription(keyPressed->scancode) };
                output += "\nCapsLock: ";
                output += keyPressed->capsLock ? "true" : "false";
                output += "\nNumLock: ";
                output += keyPressed->numLock ? "true" : "false";
                output += "\nScrollLock: ";
                output += keyPressed->scrollLock ? "true" : "false";
                text.setString(output);
            }
            else if (auto keyReleased = event->getIf<sf::Event::KeyReleased>())
            {
                auto output = std::string{ "Key Released: " + sf::Keyboard::getDescription(keyReleased->scancode) };
                output += "\nCapsLock: ";
                output += keyReleased->capsLock ? "true" : "false";
                output += "\nNumLock: ";
                output += keyReleased->numLock ? "true" : "false";
                output += "\nScrollLock: ";
                output += keyReleased->scrollLock ? "true" : "false";
                text.setString(output);
            }
        }

        window.clear();
        window.draw(text);
        window.display();
    }
}
```

### DIFF
--- a/include/SFML/Window/Event.hpp
+++ b/include/SFML/Window/Event.hpp
@@ -99,12 +99,15 @@ public:
     ////////////////////////////////////////////////////////////
     struct KeyPressed
     {
-        Keyboard::Key      code{};     //!< Code of the key that has been pressed
-        Keyboard::Scancode scancode{}; //!< Physical code of the key that has been pressed
-        bool               alt{};      //!< Is the Alt key pressed?
-        bool               control{};  //!< Is the Control key pressed?
-        bool               shift{};    //!< Is the Shift key pressed?
-        bool               system{};   //!< Is the System key pressed?
+        Keyboard::Key      code{};       //!< Code of the key that has been pressed
+        Keyboard::Scancode scancode{};   //!< Physical code of the key that has been pressed
+        bool               alt{};        //!< Is the Alt key pressed?
+        bool               control{};    //!< Is the Control key pressed?
+        bool               shift{};      //!< Is the Shift key pressed?
+        bool               system{};     //!< Is the System key pressed?
+        bool               capsLock{};   //!< Is the CapsLock key toggled?
+        bool               numLock{};    //!< Is the NumLock key toggled? (Not supported on macOS)
+        bool               scrollLock{}; //!< Is the ScrollLock key toggled?
     };
 
     ////////////////////////////////////////////////////////////
@@ -113,12 +116,15 @@ public:
     ////////////////////////////////////////////////////////////
     struct KeyReleased
     {
-        Keyboard::Key      code{};     //!< Code of the key that has been released
-        Keyboard::Scancode scancode{}; //!< Physical code of the key that has been released
-        bool               alt{};      //!< Is the Alt key pressed?
-        bool               control{};  //!< Is the Control key pressed?
-        bool               shift{};    //!< Is the Shift key pressed?
-        bool               system{};   //!< Is the System key pressed?
+        Keyboard::Key      code{};       //!< Code of the key that has been released
+        Keyboard::Scancode scancode{};   //!< Physical code of the key that has been released
+        bool               alt{};        //!< Is the Alt key pressed?
+        bool               control{};    //!< Is the Control key pressed?
+        bool               shift{};      //!< Is the Shift key pressed?
+        bool               system{};     //!< Is the System key pressed?
+        bool               capsLock{};   //!< Is the CapsLock key toggled?
+        bool               numLock{};    //!< Is the NumLock key toggled? (Not supported on macOS)
+        bool               scrollLock{}; //!< Is the ScrollLock key toggled?
     };
 
     ////////////////////////////////////////////////////////////
--- a/src/SFML/Window/Win32/WindowImplWin32.cpp
+++ b/src/SFML/Window/Win32/WindowImplWin32.cpp
@@ -887,12 +887,15 @@ void WindowImplWin32::processEvent(UINT message, WPARAM wParam, LPARAM lParam)
             if (m_keyRepeatEnabled || ((HIWORD(lParam) & KF_REPEAT) == 0))
             {
                 Event::KeyPressed event;
-                event.alt      = HIWORD(GetKeyState(VK_MENU)) != 0;
-                event.control  = HIWORD(GetKeyState(VK_CONTROL)) != 0;
-                event.shift    = HIWORD(GetKeyState(VK_SHIFT)) != 0;
-                event.system   = HIWORD(GetKeyState(VK_LWIN)) || HIWORD(GetKeyState(VK_RWIN));
-                event.code     = virtualKeyCodeToSF(wParam, lParam);
-                event.scancode = toScancode(wParam, lParam);
+                event.alt        = GetKeyState(VK_MENU) & 0x8000;
+                event.control    = GetKeyState(VK_CONTROL) & 0x8000;
+                event.shift      = GetKeyState(VK_SHIFT) & 0x8000;
+                event.system     = GetKeyState(VK_LWIN) & 0x8000 || GetKeyState(VK_RWIN) & 0x8000;
+                event.capsLock   = GetKeyState(VK_CAPITAL) & 0x0001;
+                event.numLock    = GetKeyState(VK_NUMLOCK) & 0x0001;
+                event.scrollLock = GetKeyState(VK_SCROLL) & 0x0001;
+                event.code       = virtualKeyCodeToSF(wParam, lParam);
+                event.scancode   = toScancode(wParam, lParam);
                 pushEvent(event);
             }
             break;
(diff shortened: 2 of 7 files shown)


eXpl0it3r:
This doesn't actually work the way it's supposed to. 😔

The modifiers are only set, if you actually press any of the buttons. So the event properties will only change, if the key you're pressing is the *Lock itself or when you're holding a *Lock key, while pressing another key. It doesn't detect if CapsLock is in general engaged.

texus:
I just tested the X11 implementation and that one does work as intended.
The numLock, capsLock and scrollLock values in the KeyPressed event do represent the actual lock state on Linux and not just if the key is currently pressed or not.

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -886,12 +886,15 @@ void WindowImplWin32::processEvent(UINT message, WPARAM wParam, LPARAM lParam)
...
-                event.system   = HIWORD(GetKeyState(VK_LWIN)) || HIWORD(GetKeyState(VK_RWIN));
-                event.code     = virtualKeyCodeToSF(wParam, lParam);
-                event.scancode = toScancode(wParam, lParam);
+                event.alt        = HIWORD(GetKeyState(VK_MENU)) != 0;
+                event.control    = HIWORD(GetKeyState(VK_CONTROL)) != 0;
+                event.shift      = HIWORD(GetKeyState(VK_SHIFT)) != 0;
+                event.system     = HIWORD(GetKeyState(VK_LWIN)) || HIWORD(GetKeyState(VK_RWIN));
+                event.capsLock   = HIWORD(GetKeyState(VK_CAPITAL)) != 0;
+                event.numLock    = HIWORD(GetKeyState(VK_NUMLOCK)) != 0;
+                event.scrollLock = HIWORD(GetKeyState(VK_SCROLL)) != 0;
```
texus, line 895:
```suggestion
                event.capsLock   = (LOWORD(GetKeyState(VK_CAPITAL)) & 1) != 0;
                event.numLock    = (LOWORD(GetKeyState(VK_NUMLOCK)) & 1) != 0;
                event.scrollLock = (LOWORD(GetKeyState(VK_SCROLL)) & 1) != 0;
```

You need LOWORD instead of HIWORD to get the toggled state instead of the current pressed state (see [GetKeyState function return value](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getkeystate#return-value)).

When pressing one of the lock keys, the LOWORD values were `65408` or `65409` instead of just `0` or `1`, which is why I've added the `& 1`.

Note that when pressing the lock key itself, Windows will give you the NEW state while Linux will give you the OLD state. I don't think this matters much in practice though, the important thing is that all key presses afterwards have the correct state.

eXpl0it3r: That's what I get for copy pasting and not reading the docs properly 🙃 

Thank you!

eXpl0it3r: I might need to update the documentation for macOS.  
macOS doesn't really know "Num Lock" it's not something they support.

I'll try to test it on our Mac Mini.

texus: > That's what I get for copy pasting and not reading the docs properly 🙃

Talking about reading the docs, I think the code is still completely wrong (even though it works).
I used LOWORD because the existing code was using HIWORD, but the usage of HIWORD already seems wrong.

HIWORD and LOWORD are used to get the 16-bit words from a 32-bit DWORD, but `GetKeyState` only returns a 16-bit short to begin with. My first instinct was that HIWORD should always have returned 0, but apparently the returned 16-bit value is negative and the implicit conversion from WORD to DWORD in the HIWORD call causing all high-bits to become 1 (which caused to existing code to actually work).

When pressing e.g. the F1 key, the result of `GetKeyState(VK_F1)` is always `-127` or `-128` (i.e. `10000001` or `10000000` in binary). The highest bit (which happens to be the sign bit) is 1 in both these values (because the key is being pressed). The lowest bit alternates between each key press as the state is toggled.

The same two values (`-127` and `-128`) of course also appear from `GetKeyState(VK_CAPITAL)` when pressing the caps lock key itself. When calling `GetKeyState(VK_CAPITAL)` while pressing another key though, we instead get the value `0` or `1`, as now the caps lock key isn't being held down and only the lowest bit changes depending on whether the caps lock is on or off.

So the code should probably look like this:
```c++
event.shift      = (GetKeyState(VK_SHIFT) & 0x8000) != 0;  // Checking if key is currently down
event.scrollLock = (GetKeyState(VK_CAPITAL) & 0x0001) != 0;  // Checking if lock state is on
```

eXpl0it3r: You're right!

I've also removed the `!= 0` check, since it doesn't really do anything.

ChrisThrasher (dismissed):
I'm on macOS playing around with how this works when the window is defocused. This API seems a bit unique from other keys in how a key released event happens when you disable the caps lock state as opposed to when you release the physical key. Toggling the caps lock state requires two key presses and two key releases with your physical hand but only registers as one key pressed event and one key released event. This can lead to the app's understanding of the state of the caps lock key to differ from the actual OS. For example:

1. Press caps lock (toggling on)
2. Defocus window
3. Press caps lock (toggling off)
4. Observe how the test application still thinks the caps lock key is pressed

This is not a problem per se. The next time a key press is registered with the app, the app will be able to see that caps lock is no longer engaged. I can only think of this being a problem if a user tries to use `isKeyPressed` using that stale caps lock state (which the app must choose to store itself) but at this point the user is pretty far away from recommended practices so I don't see this warranting us taking action to resolve.

TL;DR macOS implementation seems to behave as expected and the implementation looks good 👍🏻

ChrisThrasher:
Testing this under my Ubuntu VM I get different behavior. I get a KeyReleased event when I lift my finger from the caps lock key. Additionally, the only time the test app reports that caps lock is engaged as off is when caps is not engaged then I hold down the caps lock key.

ChrisThrasher:
By the way, after this PR was opened we merged https://github.com/SFML/SFML/pull/3322 which is the perfect program to test this kind of keyboard behavior.

eXpl0it3r:
> By the way, after this PR was opened we merged #3322 which is the perfect program to test this kind of keyboard behavior.

Not really, because we aren't showing key event states, which this PR is exclusively about.

The *lock keys behavior when checking the key itself with scancodes should be unaffected by this change, we're really just getting these *lock keys states for any given key press, allowing the user to determine whether they want to process the input differently given a *lock keys' state.

> This is not a problem per se. The next time a key press is registered with the app, the app will be able to see that caps lock is no longer engaged. I can only think of this being a problem if a user tries to use isKeyPressed using that stale caps lock state (which the app must choose to store itself) but at this point the user is pretty far away from recommended practices so I don't see this warranting us taking action to resolve.

I don't quite understand what you mean with this. The event states shouldn't be relevant for `isKeyPressed`.

Thraka:
It seems to me that the design wouldn't account for when I start the game and the caps lock is already on. I don't know how you would detect this apart from querying the OS about if the functionality of the key is considered on or off.

eXpl0it3r:
We aren't manually tracking the key presses and deriving the state from it, the states are provided / queried with every key event, as such it doesn't matter of it's active at the start of the application or not.

binary1248 (approved):

include/SFML/Window/Event.hpp
```diff
@@ -99,12 +99,15 @@ class Event
...
-        bool               system{};   //!< Is the System key pressed?
+        Keyboard::Key      code{};       //!< Code of the key that has been pressed
+        Keyboard::Scancode scancode{};   //!< Physical code of the key that has been pressed
+        bool               alt{};        //!< Is the Alt key pressed?
+        bool               control{};    //!< Is the Control key pressed?
+        bool               shift{};      //!< Is the Shift key pressed?
+        bool               system{};     //!< Is the System key pressed?
+        bool               capsLock{};   //!< Is the CapsLock key toggled?
+        bool               numLock{};    //!< Is the NumLock key toggled? (Not supported on macOS)
+        bool               scrollLock{}; //!< Is the ScrollLock key toggled?
```
kimci86, line 110: Looking at the implementation, it looks like macOS supports num lock but not scroll lock, contrary to the documentation.

eXpl0it3r: Woops, good catch. Didn't even double check that before merging. 😅

eXpl0it3r: See https://github.com/SFML/SFML/pull/3712

PR #3242, opened 2024-09-24
Fix clang-tidy-19 errors

ChrisThrasher:
## Description

Homebrew now ships the new LLVM 19 release. As with any new clang-tidy release that means we have to fix or disable some new warnings. If you run CI on `master` you can see the failures [here](https://github.com/SFML/SFML/actions/runs/11001329795/job/30545764694). Fixing this is required to unblock all other PRs.

Most changes are related to ensuring file-local functions are within an anonymous namespace. This resulted in a rather large diff because a lot of functions had to move where they were in the file. The contents of those functions did not change though.


src/SFML/Window/GlContext.cpp
```diff
@@ -388,7 +388,7 @@ struct GlContext::TransientContext
     ////////////////////////////////////////////////////////////
     static std::optional<TransientContext>& get()
     {
-        thread_local std::optional<TransientContext> transientContext;
+        thread_local std::optional<TransientContext> transientContext; // NOLINT(misc-use-internal-linkage)
```
ChrisThrasher, line 391: I assumed this was a false positive.

binary1248: Could also just add `static` in front of it even though `thread_local` already implies this according to the standard.

examples/sockets/CMakeLists.txt
```diff
@@ -1,7 +1,5 @@
 # all source files
-set(SRC Sockets.cpp
-        TCP.cpp
-        UDP.cpp)
+set(SRC Sockets.cpp)
```
eXpl0it3r, line 3: Not sure I agree with this change. Why can't we keep this separated for an example?

eXpl0it3r (approved):

PR #3244, opened 2024-09-25
Prevent reading from an uninitialized character buffer

ChrisThrasher:
## Description

If `GLEXT_glGetInfoLog` fails then `log` is left unchanged which can lead to a reading from an uninitialized buffer on the following line. Initializing the buffer ensures this never happens. While I was touching this code I elected to switch to `std::array` as well since SFML has made previous efforts to stop using C-style arrays and I wanted to continue that effort.

Coverity caught this in a recent scan
```
/home/buildbot-worker/coverity/build/src/SFML/Graphics/Shader.cpp: 943 in sf::Shader::compile(std::basic_string_view<char, std::char_traits<char>>, std::basic_string_view<char, std::char_traits<char>>, std::basic_string_view<char, std::char_traits<char>>)()
937             GLint success = 0;
938             glCheck(GLEXT_glGetObjectParameteriv(fragmentShader, GLEXT_GL_OBJECT_COMPILE_STATUS, &success));
939             if (success == GL_FALSE)
940             {
941                 char log[1024];
942                 glCheck(GLEXT_glGetInfoLog(fragmentShader, sizeof(log), nullptr, log));
   CID 420681:    (UNINIT)
   Using uninitialized value "*log" when calling "operator <<". [Note: The source code implementation of the function has been overridden by a builtin model.]
943                 err() << "Failed to compile fragment shader:" << '\n' << log << std::endl;
944                 glCheck(GLEXT_glDeleteObject(fragmentShader));
945                 glCheck(GLEXT_glDeleteObject(shaderProgram));
946                 return false;
947             }
948     
```

### DIFF
--- a/src/SFML/Graphics/Shader.cpp
+++ b/src/SFML/Graphics/Shader.cpp
@@ -40,6 +40,7 @@
 #include <SFML/System/Vector2.hpp>
 #include <SFML/System/Vector3.hpp>
 
+#include <array>
 #include <fstream>
 #include <iomanip>
 #include <ostream>
@@ -882,9 +883,9 @@ bool Shader::compile(std::string_view vertexShaderCode, std::string_view geometr
         glCheck(GLEXT_glGetObjectParameteriv(vertexShader, GLEXT_GL_OBJECT_COMPILE_STATUS, &success));
         if (success == GL_FALSE)
         {
-            char log[1024];
-            glCheck(GLEXT_glGetInfoLog(vertexShader, sizeof(log), nullptr, log));
-            err() << "Failed to compile vertex shader:" << '\n' << log << std::endl;
+            std::array<char, 1024> log{};
+            glCheck(GLEXT_glGetInfoLog(vertexShader, sizeof(log), nullptr, log.data()));
+            err() << "Failed to compile vertex shader:" << '\n' << log.data() << std::endl;
             glCheck(GLEXT_glDeleteObject(vertexShader));
             glCheck(GLEXT_glDeleteObject(shaderProgram));
             return false;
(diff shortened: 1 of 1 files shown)


binary1248 (approved):

src/SFML/Graphics/Shader.cpp
```diff
@@ -882,9 +883,9 @@ bool Shader::compile(std::string_view vertexShaderCode, std::string_view geometr
         glCheck(GLEXT_glGetObjectParameteriv(vertexShader, GLEXT_GL_OBJECT_COMPILE_STATUS, &success));
         if (success == GL_FALSE)
         {
-            char log[1024];
-            glCheck(GLEXT_glGetInfoLog(vertexShader, sizeof(log), nullptr, log));
-            err() << "Failed to compile vertex shader:" << '\n' << log << std::endl;
+            std::array<char, 1024> log{};
+            glCheck(GLEXT_glGetInfoLog(vertexShader, sizeof(log), nullptr, log.data()));
```
kimci86, line 887: `log.size()` would be more obviously correct than `sizeof(log)`

ChrisThrasher: `std::array:size` says "I want the number of elements". `sizeof` says "I want the number of bytes". In the case of `std::array<char>` and similar specializations, the two both return the same value but I still think they convey different meanings. In the case of C APIs that want a size (in bytes) and an address, I think it makes sense to use `sizeof`, but we can still use `.size()` without issue since realistically nobody will refine `log` to be an array of `wchar_t` or something else larger than 1 byte.

kimci86: My thinking was that `sizeof(some_container)` is generally not related to the valid range `[data(), data() + size())`. It happens to be ok with `std::array<char>` but it is dependent on the kind of container. It would not work with `std::vector<char>`. Anyways this is minor, let's ignore this if I am the only one confused with this `sizeof`.

ChrisThrasher: > It happens to be ok with std::array<char> but it is dependent on the kind of container. It would not work with std::vector<char>.

For whatever reason that hadn't crossed my mind. I was guarding against the admittedly less likely situation where the array type was changed to something larger than 1 byte. I see no reason that would ever happen but I can certainly imagine a situation where the container is replaced with `std::vector`. I think I'm coming around to your perspective on this.

PR #3246, opened 2024-09-26
DRY shader creation via helper lambda

vittorioromeo:
Should not change any existing behavior, just DRY refactoring. Useful to clearly demonstrate that there are no difference between the creation/attachment of different shader types.

### DIFF
--- a/src/SFML/Graphics/Shader.cpp
+++ b/src/SFML/Graphics/Shader.cpp
@@ -868,89 +868,50 @@ bool Shader::compile(std::string_view vertexShaderCode, std::string_view geometr
     // Create the program
     const GLEXT_GLhandle shaderProgram = glCheck(GLEXT_glCreateProgramObject());
 
-    // Create the vertex shader if needed
-    if (!vertexShaderCode.empty())
+    // Helper function for shader creation
+    const auto createAndAttachShader =
+        [shaderProgram](GLenum shaderType, const char* shaderTypeStr, std::string_view shaderCode)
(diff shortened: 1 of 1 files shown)


src/SFML/Graphics/Shader.cpp
```diff
@@ -868,89 +868,49 @@ bool Shader::compile(std::string_view vertexShaderCode, std::string_view geometr
     // Create the program
     const GLEXT_GLhandle shaderProgram = glCheck(GLEXT_glCreateProgramObject());
 
-    // Create the vertex shader if needed
-    if (!vertexShaderCode.empty())
+    // Helper function for shader creation
+    const auto createAndAttachShader = [&](GLenum shaderType, const char* shaderTypeStr, std::string_view shaderCode)
```
ChrisThrasher, line 872:
```suggestion
    const auto createAndAttachShader = [shaderProgram](GLenum shaderType, const char* shaderTypeStr, std::string_view shaderCode)
```
If a lambda only captures 1 or 2 things I'd prefer we explicitly capture those variables.

ChrisThrasher (changes requested):
One nitpick but LGTM

src/SFML/Graphics/Shader.cpp
vittorioromeo: Good point :)

ChrisThrasher (approved):

ChrisThrasher:

PR #3248, opened 2024-09-26
Fix `island` example freeze when submitting work too quickly

vittorioromeo:
As the title says, if you run the `island` example and keep pressing Enter in quick succession, the example would get stuck prior to this PR.

### DIFF
--- a/examples/island/Island.cpp
+++ b/examples/island/Island.cpp
@@ -398,7 +398,7 @@ void generateTerrain(sf::Vertex* buffer)
         {
             const std::lock_guard lock(workQueueMutex);
 
-            if (workQueue.empty())
+            if (pendingWorkCount == 0u)
                 break;
         }
 


examples/island/Island.cpp
```diff
@@ -534,7 +534,8 @@ int main()
                 switch (event->getIf<sf::Event::KeyPressed>()->code)
                 {
                     case sf::Keyboard::Key::Enter:
-                        generateTerrain(terrainStagingBuffer.data());
+                        if (pendingWorkCount == 0 && !bufferUploadPending)
+                            generateTerrain(terrainStagingBuffer.data());
```
eXpl0it3r, line 538: Does the reading of `pendingWorkCount` not require a lock?

What about clearing the work items and queuing the new changes?

eXpl0it3r:
When you say "getting stuck", does it just take a while for all the items to be processed, or does it never return control back?  
Are running into some deadlock?

vittorioromeo:
Basically the application doesn't actually freeze, but `pendingWorkCount` becomes negative therefore it always displays the *"Generating Terrain..."* message. So it's basically a soft-lock: `generateTerrain` sets `pendingWorkCount` to `32` (block count), and the `threadFunction` decrements `pendingWorkCount` by one per block... however if we press `Enter` while this is happening, we reset the `pendingWorkCount` to 32 in the middle of things and I think it screws up the counting.

vittorioromeo:

ChrisThrasher (approved):
I can reproduce the problem and demonstrate that this PR fixes the problem.

PR #3249, opened 2024-09-26
Prefer try_emplace over emplace for maps

eXpl0it3r:
## Description

[Reported by SonarCloud](https://sonarcloud.io/project/issues?rules=cpp%3AS6030&issueStatuses=OPEN%2CCONFIRMED&id=SFML_SFML), `try_emplace` should be preferred over `emplace` for `std::map` and `std::unordered_map` as it doesn't unnecessarily instantiate the value, if the key already exists.

## Tasks

-   [ ] Tested on Linux
-   [x] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

Building and running the examples should be enough.


ChrisThrasher (approved):
> [Reported by SonarCloud](https://sonarcloud.io/project/issues?rules=cpp%3AS6030&issueStatuses=OPEN%2CCONFIRMED&id=SFML_SFML), try_emplace should be preferred over emplace for std::map and std::unordered_map as it doesn't unnecessarily instantiate the value, if the key already exists.

I suppose `try_emplace` is better for those reasons but in this case the PR is a no-op because I don't see any instances where a potential object construction is being avoided. No harm done though so I'll approve this if it means fewer SonarCloud issues to sort through. That does have some value. We certainly do similar things to appease clang-tidy.

eXpl0it3r:
That's true, but it should also just be the default. So reflecting that in the code base, will likely lead to other usages not being copy-pasted from a suboptimal source.

PR #3250, opened 2024-09-26
Use explicit constructors when implicit construction isn't wanted

eXpl0it3r:
## Description

[Reported by SonarCloud](https://sonarcloud.io/project/issues?impactSeverities=HIGH&rules=cpp%3AS1709&issueStatuses=OPEN%2CCONFIRMED&id=SFML_SFML), we should be using `explicit` on constructors with single parameters, to prevent implicit constructions.

There are of course a few places, where we actually do want implicit constructions:

- String overloads (wchar_t, char32_t, const wchar_t*, const char32_t*, const std::wstring&, std::u32string)
- `Ftp::DirectoryResponse(const Ftp::Response&)`
- `Time(const std::chrono::duration&)`
- `StencileValue(int)` & `StencileValue(unsigned int)`
- `RenderStates(...)`
- `Event(const TEventSubtype&)`

Is there a type on the list that shouldn't be there or one that you think should be there?

## Tasks

-   [ ] Tested on Linux
-   [x] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

Should compile with examples.


ChrisThrasher (approved):

ZXShady:
I would argue that single char sf::String constructors being implicit is a bad idea



also constructors that have default arguements can be invoked implicitly like this one

```cpp
    Text(const Font& font, String string = "", unsigned int characterSize = 30);


void f(sf::Text const&);
sf::Font font;
f(font); // compiles!
```

eXpl0it3r:
> I would argue that single char sf::String constructors being implicit is a bad idea

Why?

eXpl0it3r:
> also constructors that have default arguements can be invoked implicitly like this one
> 
> ```c++
>     Text(const Font& font, String string = "", unsigned int characterSize = 30);
> 
> 
> void f(sf::Text const&);
> sf::Font font;
> f(font); // compiles!
> ```

I went over the public API again and change those constructors as well. This required a change to the joystick example, is this okay this way, @ChrisThrasher?

ZXShady:
@eXpl0it3r 

putting explicit on the Text constructor prohibits this code from compiling

```cpp

sf::Text getText(const Font& f)
{
   return {f,"Hello Exploiter",30}; // fails to xompile no implicit conversion
}

```

the solution is to provide another overload that only takes sf::Font and is marked explicit and uses constructor delegation to delegate to the Font,String,Size constructor but this requires adding docs to the new constructor

eXpl0it3r:
A text constructor only taking a font doesn't make sense though. You need to set some string and setting it implicitly isn't a road I'd want to go down.

Since the effect of this change isn't wanted, I'll revert it again.

ChrisThrasher:
> A text constructor only taking a font doesn't make sense though.

That's the status quo. Are you saying that we should remove the default argument on `string`? 

```cpp
Text(const Font& font, String string = "", unsigned int characterSize = 30);
```

eXpl0it3r:
The status quo as default arguments. It's clear from the function signature that they are there and what they're set to. Creating an overload with just the front and no other params with default values didn't make sense, because you'll then set a default string in "secret".

I'm just saying we should keep the status quo then.

vittorioromeo:
> @ZXShady 
>
> or I would make the user type sf::Text infront of the return statement

That would be my vote.

ZXShady:
also to note that not all sfml types provide `{}` syntax to default intiialize

```cpp
sf::Text getText()
{
     return {}; // works fine
}

sf::CircleShape getCircle()
{
      return {}; // error no implicit conversion
}
```
why doesn't this work? it is because the default constructor of CircleShape is a constructor marked explicit with efault parameters

```cpp
explicit CircleShape(float radius = 0, std::size_t pointCount = 30);
```

eXpl0it3r:
I've removed the explicit constructors with default params.

ChrisThrasher (approved):

ZXShady:
> > I would argue that single char sf::String constructors being implicit is a bad idea
> 
> Why?



Actually the class relies on implicit conversions from chars like this line

```cpp

sf::String a;
a += 'A'; // the char is implicitly converted into a sf::String then operator+= is called

```

PR #3251, opened 2024-09-29
Remove unnecessary ternaries

ChrisThrasher:
## Description

It's simpler to let the boolean get promoted to an integer

### DIFF
--- a/src/SFML/Window/DRM/DRMContext.cpp
+++ b/src/SFML/Window/DRM/DRMContext.cpp
@@ -683,7 +683,7 @@ void DRMContext::display()
 ////////////////////////////////////////////////////////////
 void DRMContext::setVerticalSyncEnabled(bool enabled)
 {
-    eglCheck(eglSwapInterval(m_display, enabled ? 1 : 0));
+    eglCheck(eglSwapInterval(m_display, enabled));
 }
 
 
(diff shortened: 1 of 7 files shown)


src/SFML/Audio/SoundSource.cpp
```diff
@@ -63,7 +63,7 @@ void SoundSource::setVolume(float volume)
 void SoundSource::setSpatializationEnabled(bool enabled)
 {
     if (auto* sound = static_cast<ma_sound*>(getSound()))
-        ma_sound_set_spatialization_enabled(sound, enabled ? MA_TRUE : MA_FALSE);
```
eXpl0it3r, line 66: I wouldn't change this, as we should use the API variables and not assume the conversion

ChrisThrasher: What's dangerous about assuming `MA_TRUE` continues to be defined as `1`? If one day Miniaudio upgrades to a C standard that has booleans (how crazy that some libs still don't assume the presence of booleans) and `MA_TRUE` is redefined as `true` then we'll still be fine.

eXpl0it3r: It's mostly just technically more correct. Maybe one day MA_TRUE is 0 and MA_FALSE -1

eXpl0it3r (approved):

PR #3252, opened 2024-09-30
delegate to operator equals in operator not equals

ZXShady:
all other `operator!=` implementations did this except these two.

~~cannot wait for C++20~~


eXpl0it3r:

PR #3253, opened 2024-09-30
Update the macOS images for the CI analysis

eXpl0it3r:
## Description

[It looks like](https://github.com/SFML/SFML/actions/runs/11099776203/job/30834545898) brew on `macos-12` doesn't ship the wanted LLVM version anymore. So each build triggers a complete build of clang and clang-tools, which can take 3+h.

By updating to the `macos-14` image, we should get pre-built binaries again.

## How to test this PR?

The CI Analyzer steps for iOS and macOS shouldn't build LLVM and shouldn't take 2+h.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -323,9 +323,6 @@ jobs:
     name: Formatting
     runs-on: ubuntu-22.04
 
-    strategy:
-      fail-fast: false
-
     steps:
     - name: Checkout Code
       uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
I'm apparently not smart enough to get clang-tidy working on the `macos-14` image...

ChrisThrasher:
I tried a few things to fix this to no avail:
* Stop doing unity builds
* Switch from AppleClang to LLVM Clang (this was achieved by removing `-DCMAKE_OSX_ARCHITECTURES=arm64`)
* `brew remove mono` (This brew package is not even installed in the macOS 14 images)

The problem seems to be related to a system installation of Mono which contains Freetype headers that are erroneously being picked up in the build. The proper fix probably entails something involving the priority with which CMake looks for the path to freetype headers but I'm not sure what that specifically looks like.

ChrisThrasher:
https://github.com/SFML/SFML/blob/71594d402e8fa297fbf5b964a2921e514e28e5eb/cmake/Modules/FindFreetype.cmake#L54-L71

I suspect the problem is to do with how `FREETYPE_INCLUDE_DIR_ft2build` is being set. We're not taking the right precautions to ensure our vendored copy is being used.

ChrisThrasher:
I wrote https://github.com/SFML/SFML/pull/3258 to work around the issues of building LLVM from scratch. Eventually we do need to go back to the latest LLVM release and the latest macOS release but for now this fixes the excessively long CI times.

eXpl0it3r:
Superseded by #3258

PR #3255, opened 2024-10-01
Use `if constexpr` instead of a `switch` statement

ZXShady:
The condition is a constant expression, it makes more sense here to use `if constexpr` than a `switch` statement.


ChrisThrasher (approved):

ChrisThrasher:

PR #3257, opened 2024-10-01
Use explicit conversion operator instead of explicit constructor

ZXShady:
I also added a type trait test to check whether the conversion operator is marked explicit. (via using std::is_convertible)

this change has the advantage of being  clearer imo as it is a conversion operator for conersions not a constructor.

and also it allows users to take the address of convesion operator which I don't know if someone might find that useful. but it is extra utility for free

### DIFF
--- a/include/SFML/Graphics/Glsl.inl
+++ b/include/SFML/Graphics/Glsl.inl
@@ -128,18 +128,13 @@ struct Vector4
 #endif
 
     ////////////////////////////////////////////////////////////
-    /// \brief Conversion constructor
-    ///
-    /// \param other 4D vector of different type
+    /// \brief Converts the vector to another type of vector
     ///
     ////////////////////////////////////////////////////////////
     template <typename U>
-    constexpr explicit Vector4(const Vector4<U>& other) :
-    x(static_cast<T>(other.x)),
-    y(static_cast<T>(other.y)),
-    z(static_cast<T>(other.z)),
-    w(static_cast<T>(other.w))
+    constexpr explicit operator Vector4<U>() const
     {
+        return Vector4<U>(static_cast<U>(x), static_cast<U>(y), static_cast<U>(z), static_cast<U>(w));
     }
 
     ////////////////////////////////////////////////////////////
--- a/test/Graphics/Rect.test.cpp
+++ b/test/Graphics/Rect.test.cpp
@@ -35,15 +35,6 @@ TEMPLATE_TEST_CASE("[Graphics] sf::Rect", "", int, float)
             STATIC_CHECK(rectangle.position == position);
             STATIC_CHECK(rectangle.size == dimension);
         }
-
-        SECTION("Conversion constructor")
-        {
-            constexpr sf::FloatRect sourceRectangle({1.0f, 2.0f}, {3.0f, 4.0f});
-            constexpr sf::IntRect   rectangle(sourceRectangle);
-
-            STATIC_CHECK(rectangle.position == sf::Vector2i(1, 2));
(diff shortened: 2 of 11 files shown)


test/Graphics/Rect.test.cpp
```diff
@@ -38,6 +38,8 @@ TEMPLATE_TEST_CASE("[Graphics] sf::Rect", "", int, float)
 
         SECTION("Conversion constructor")
```
ChrisThrasher, line 39:
```suggestion
        SECTION("Conversion operator")
```

test/System/Vector2.test.cpp
```diff
@@ -38,6 +38,8 @@ TEMPLATE_TEST_CASE("[System] sf::Vector2", "", int, float)
 
         SECTION("Conversion constructor")
```
ChrisThrasher, line 39:
```suggestion
        SECTION("Conversion operator")
```

test/System/Vector3.test.cpp
```diff
@@ -35,6 +35,8 @@ TEMPLATE_TEST_CASE("[System] sf::Vector3", "", int, float)
 
         SECTION("Conversion constructor")
```
ChrisThrasher, line 36:
```suggestion
        SECTION("Conversion operator")
```

ChrisThrasher (changes requested):
I think I'm in favor of this. I'm curious to hear what others think of it.

eXpl0it3r:
To me this moves the operation to the wrong side. On an API level we're saying now, that a vector can be converted into another, instead of being constructable from another type.

vittorioromeo:
I immediately cannot think of anything wrong with this approach, it's simpler than the status quo and moves us closer to making vectors into aggregate types, so I'm in favour.

vittorioromeo:
> To me this moves the operation to the wrong side. On an API level we're saying now, that a vector can be converted into another, instead of being constructable from another type.

Why is it "wrong" to say that *"a `Vector2f` can convert to a `Vector2u`"* and why is it "right" to say that *"a `Vector2u` can be constructed from a `Vector2f`"*?

If anything, the conversion feels more natural/intuitive to me.

vittorioromeo:
Codegen is affected with `-O0` on both GCC and Clang, but I cannot tell if it's significant. With `-O1` and above, codegen is the same. See https://gcc.godbolt.org/z/8bPfos17z

ChrisThrasher (approved):

ChrisThrasher:

ChrisThrasher:
For the sake of consistency I think `sf::priv::Vector4` also needs this treatment.

include/SFML/Graphics/Glsl.inl
```diff
@@ -134,12 +134,9 @@ struct Vector4
     ///
     ////////////////////////////////////////////////////////////
     template <typename U>
-    constexpr explicit Vector4(const Vector4<U>& other) :
-    x(static_cast<T>(other.x)),
-    y(static_cast<T>(other.y)),
-    z(static_cast<T>(other.z)),
-    w(static_cast<T>(other.w))
+    constexpr explicit operator Vector4<U>() const
```
ChrisThrasher, line 137: Doc comment above needs to be updated to remove the parameter.

ZXShady:
> To me this moves the operation to the wrong side. On an API level we're saying now, that a vector can be converted into another, instead of being constructable from another type.

Yes, because this is what it is do you say when you see this `int(1.5f)` that you are constructing an int from a float or do you say I convert the float to the int? same here I say I convert this Vector2f to a Vector2i.

ChrisThrasher (approved):

ChrisThrasher:

PR #3258, opened 2024-10-01
Revert to LLVM 18 until we can upgrade to macOS 14 images

ChrisThrasher:
## Description

This fixes the problem where CI was building LLVM from source which takes a _long_ time. If you use LLVM 18 on macOS 12, Homebrew still has to build it from source but if we bump that up to macOS 13 then Homebrew can use a prebuilt version.


eXpl0it3r (approved):
Do you want to update the macOS images for the regular builds as well?

ChrisThrasher:
> Do you want to update the macOS images for the regular builds as well?

I do in a later PR.

PR #3261, opened 2024-10-02
Upgrade all macOS 12 CI jobs to macOS 13

ChrisThrasher:
## Description

There are further issues upgrading to macOS 14 but this gets us one step closer to that.


eXpl0it3r (approved):
Note that once we go to macos-14 we need specify x64 for our 64bits builds as the default arch on macos-14 is arm64

ChrisThrasher:
> Note that once we go to macos-14 we need specify x64 for our 64bits builds as the default arch on macos-14 is arm64

I was planning to leave the x64 jobs on macOS 13 actually. Is that okay with you? I want to upgrade all other macOS/iOS jobs to macOS 14 though.

eXpl0it3r:
Yeah that works as well! 🙂

PR #3273, opened 2024-10-05
Run format before any other build jobs

Marioalexsan:
# Description

Checking formatting takes a really short time compared to all of the other jobs. It would make sense to run that first before proceeding with builds.

# How to test

Check the resulting job graph, for example [this one](https://github.com/SFML/SFML/actions/runs/11196163182?pr=3273).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,7 @@ defaults:
 jobs:
   build:
     name: ${{ matrix.platform.name }} ${{ matrix.config.name }} ${{ matrix.type.name }}
+    needs: format
     runs-on: ${{ matrix.platform.os }}
 
     env:
@@ -338,6 +339,7 @@ jobs:
 
   tidy:
     name: Analyzing on ${{ matrix.platform.name }}
+    needs: format
     runs-on: ${{ matrix.platform.os }}
 
     strategy:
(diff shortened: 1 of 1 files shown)


ChrisThrasher:
For whatever reason, the formatting job tends to finish very shortly after submitting a PR. If you want formatting feedback then you can quickly get that by looking at the pipeline run. I'm not sure I want to stop the entire pipeline on formatting though since the majority of the value of the pipeline exists in all the other jobs. For example if you code doesn't even compile that's a bigger deal than the formatting being wrong.

Marioalexsan:
I'm thinking that even if the code compiles, if the formatting is wrong, you'll have to push the branch again and retrigger all jobs, which is wasteful from a resource perspective (even if the runners are free for public repositories). 
I might be wrong on this one, but I think the formatting job is also not guaranteed to run first. The scheduler might pick up 20 of the build steps first, then do those before finally touching the formatting job and failing the pipeline.

ChrisThrasher:
For me it comes down to priorities. Correct formatting is one of the least important things in the CI pipeline. I can fix that myself in a few seconds and it has no affect on the functionality of the code. What I most care about is that people submit code that compiles, passes our tests, and passes the other important checks in CI.

What commonly happens is a PR is submitted while I'm away from my computer so when I open it, CI has already finished and I see everything that happened. I would be irked to see formatting failed for some tiny reason and thus I would get zero feedback on whether the code compiles and passes the test suite.

PR #3274, opened 2024-10-06
Remove unnecessary friend declarations

ZXShady:
These functions do not access any private internals they do not need to be `friend`ed


vittorioromeo:

vittorioromeo (approved):

ChrisThrasher (approved):
Good catch! Looks like the need for a `friend` function was removed in https://github.com/SFML/SFML/pull/2133.

PR #3275, opened 2024-10-06
Use more correct iterator comparison operator

ZXShady:
Switching the iterator comparison operator used lowers the requirement for iterators from *RandomAccessIterators* 
to *ForwardIterators*, which is a benefit. It is also more accurate; instead of saying "if `begin` is less than the end, stop." it states "if `begin` is not equal to end, stop."

Additionally, the old '<' did not protect us from undefined behavior because it is undefined to compare two iterators relationally that are not:

1. past the end
2. within the array

### DIFF
--- a/include/SFML/System/Utf.inl
+++ b/include/SFML/System/Utf.inl
@@ -168,7 +168,7 @@ template <typename In>
 std::size_t Utf<8>::count(In begin, In end)
 {
     std::size_t length = 0;
-    while (begin < end)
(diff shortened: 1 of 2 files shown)


vittorioromeo:

vittorioromeo (approved):

ChrisThrasher (approved):

ChrisThrasher (approved):

PR #3278, opened 2024-10-08
Remove specified trailing return types in lamdbas

ZXShady:
The lamdba can deduce the return type correctly having it deducing it automaticly makes less rooms for mistakes 
later when we change the impl.


ChrisThrasher (approved):

ChrisThrasher:

PR #3279, opened 2024-10-08
Make true and false keywords lowercase and monospaced in docs

ZXShady:
Continues the work of pr #3213

I tried to find every instance of True or False being capitalized and them not being inclosed in `'`


ChrisThrasher:
I've always been a little confused why we have full Doxygen docs for private interfaces. Regardless, as long as we have private documentation then we ought to keep them consistent with and up to the quality of our public documentation.

ChrisThrasher:

eXpl0it3r (approved):

PR #3281, opened 2024-10-08
Stop uploading code coverage from non-pull request CI runs

ChrisThrasher:
We only care about code coverage in PRs so let's save some CI time and stop doing this for non-pull request CI pipelines. It's just another job that is prone to run for a long just to fail for some silly reason.


ChrisThrasher:

eXpl0it3r (approved):

kimci86:
It seems this broke coverage reports on pull requests. I guess we need to upload coverage results on pushes to master so pull requests targeting master have a baseline for comparison.

PR #3283, opened 2024-10-08
Add getter for `sf::Texture::m_pixelsFlipped`

### LINKED ISSUE #1770: Feature request: implement getter for sf::Texture::m_pixelsFlipped
Hello.

There's an issue that I've encountered during [ImGui-SFML](https://github.com/eliasdaler/imgui-sfml) development. 

When users use `sf::RenderTexture` and set it on `sf::Sprite`, it's impossible to know if `sf::Sprite` has texture set from regular `Texture` of from `RenderTexture`, which results in inverted image in the former case. 
If `sf::Texture::m_pixelsFlipped` was accessible, I'd be able to check this an invert UVs that I pass to `ImGui::Image` [here](https://github.com/eliasdaler/imgui-sfml/blob/master/imgui-SFML.cpp#L532).

I think that this might be also useful for some shaders and other OpenGL interactions where it matters if pixels are inverted or not.

ChrisThrasher:
Closes #1770 

There is reason to believe that larger rendering changes in SFML 4 will render (pun intended) this API moot and thus it will get removed. Nevertheless I think there is value in providing this feature even if this feature only lives for as long as SFML 3. It's trivial to write, test, and maintain and I see no reason why this would limit any implementation changes we wish to make in the v3 release cycle.

### DIFF
--- a/include/SFML/Graphics/Texture.hpp
+++ b/include/SFML/Graphics/Texture.hpp
@@ -711,6 +711,14 @@ public:
     ////////////////////////////////////////////////////////////
     [[nodiscard]] static unsigned int getMaximumSize();
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the state of whether the Y orientatino of pixels are flipped
+    ///
+    /// \return `true` if pixels are flipped, otherwise `false`
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] bool getPixelsFlipped() const;
+
 private:
     friend class Text;
     friend class RenderTexture;
--- a/src/SFML/Graphics/Texture.cpp
+++ b/src/SFML/Graphics/Texture.cpp
@@ -1032,6 +1032,13 @@ unsigned int Texture::getMaximumSize()
 
 
 ////////////////////////////////////////////////////////////
+bool Texture::getPixelsFlipped() const
+{
+    return m_pixelsFlipped;
+}
+
+
+////////////////////////////////////////////////////////////
 Texture& Texture::operator=(const Texture& right)
 {
     Texture temp(right);
--- a/test/Graphics/Texture.test.cpp
+++ b/test/Graphics/Texture.test.cpp
@@ -132,6 +132,7 @@ TEST_CASE("[Graphics] sf::Texture", runDisplayTests())
             CHECK(!texture.isSrgb());
             CHECK(!texture.isRepeated());
             CHECK(texture.getNativeHandle() == 0);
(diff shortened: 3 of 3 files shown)


include/SFML/Graphics/Texture.hpp
```diff
@@ -711,6 +711,14 @@ class SFML_GRAPHICS_API Texture : GlResource
     ////////////////////////////////////////////////////////////
     [[nodiscard]] static unsigned int getMaximumSize();
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the state of whether the Y orientatino of pixels are flipped
+    ///
+    /// \return `true` if pixels are flipped, otherwise `false`
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] bool getPixelsFlipped() const;
```
ZXShady, line 720: the name is confusing, when I read it I thought it returns a pointer to the pixels flipped but no this is just a boolean getter I think an isPixelsFlipped or something else is better name

ChrisThrasher: `isPixelsFlipped` or maybe `arePixelsFlipped`? I agree that a better name probably exists.

vittorioromeo:
Is there a need / actual use case to expose this? Or are we exposing an implementation detail?

eXpl0it3r:
See the linked issue and issue in imgui-sfml

vittorioromeo:
> See the linked issue and issue in imgui-sfml

Completely missed that, my bad. Will take a look at it.

include/SFML/Graphics/Texture.hpp
vittorioromeo: How about:
- `m_isYFlipped`
- `bool isYFlipped()`

or something like that?

JonnyPtn: I'd say `arePixelsFlipped` or `pixelsAreFlipped`

include/SFML/Graphics/Texture.hpp
```diff
@@ -711,6 +711,14 @@ class SFML_GRAPHICS_API Texture : GlResource
     ////////////////////////////////////////////////////////////
     [[nodiscard]] static unsigned int getMaximumSize();
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the state of whether the Y orientatino of pixels are flipped
```
JonnyPtn, line 715:
```suggestion
    /// \brief Get the state of whether the Y orientation of pixels are flipped
```

JonnyPtn (changes requested):

PR #3288, opened 2024-10-12
Reduce the scope of variables

ChrisThrasher:
See also https://github.com/SFML/SFML/pull/2119 and https://github.com/SFML/SFML/pull/3060 for past precedent.

### DIFF
--- a/examples/island/Island.cpp
+++ b/examples/island/Island.cpp
@@ -240,11 +240,10 @@ sf::Vector2f computeNormal(float left, float right, float bottom, float top)
 /// coordinates.
 ///
 ////////////////////////////////////////////////////////////
-const auto scalingFactors = sf::Vector2f(windowSize).componentWiseDiv(sf::Vector2f(resolution));
-
 sf::Vertex computeVertex(sf::Vector2u position)
 {
-    sf::Vertex vertex;
+    static const auto scalingFactors = sf::Vector2f(windowSize).componentWiseDiv(sf::Vector2f(resolution));
+    sf::Vertex        vertex;
     vertex.position  = sf::Vector2f(position).componentWiseMul(scalingFactors);
     vertex.color     = getTerrainColor(getElevation(position), getMoisture(position));
     vertex.texCoords = computeNormal(getElevation(position - sf::Vector2u(1, 0)),
--- a/src/SFML/Audio/AudioDevice.cpp
+++ b/src/SFML/Audio/AudioDevice.cpp
@@ -462,9 +462,9 @@ std::optional<ma_device_id> AudioDevice::getSelectedDeviceId() const
     if (!deviceName)
         deviceName = PlaybackDevice::getDefaultDevice();
 
-    auto iter = std::find_if(devices.begin(),
-                             devices.end(),
-                             [&](const auto& device) { return device.name == deviceName; });
+    const auto iter = std::find_if(devices.begin(),
+                                   devices.end(),
+                                   [&deviceName](const auto& device) { return device.name == deviceName; });
 
     if (iter != devices.end())
         return iter->id;
--- a/src/SFML/Audio/PlaybackDevice.cpp
+++ b/src/SFML/Audio/PlaybackDevice.cpp
@@ -66,8 +66,8 @@ bool setDevice(const std::string& name)
 {
     // Perform a sanity check to make sure the user isn't passing us a non-existent device name
     const auto devices = priv::AudioDevice::getAvailableDevices();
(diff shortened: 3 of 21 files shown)


src/SFML/Audio/PlaybackDevice.cpp
```diff
@@ -66,7 +66,9 @@ bool setDevice(const std::string& name)
 {
     // Perform a sanity check to make sure the user isn't passing us a non-existent device name
     const auto devices = priv::AudioDevice::getAvailableDevices();
-    if (auto iter = std::find_if(devices.begin(), devices.end(), [&](const auto& device) { return device.name == name; });
+    if (auto iter = std::find_if(devices.begin(),
+                                 devices.end(),
+                                 [name](const auto& device) { return device.name == name; });
```
ZXShady, line 71:
```suggestion
                                 [&name](const auto& device) { return device.name == name; });
```

no need to copy it unnecessarily

src/SFML/Audio/AudioDevice.cpp
```diff
@@ -464,7 +464,7 @@ std::optional<ma_device_id> AudioDevice::getSelectedDeviceId() const
 
     auto iter = std::find_if(devices.begin(),
                              devices.end(),
-                             [&](const auto& device) { return device.name == deviceName; });
+                             [deviceName](const auto& device) { return device.name == deviceName; });
```
ZXShady, line 467:
```suggestion
                             [&deviceName](const auto& device) { return device.name == deviceName; });
```

no need to copy an `std::optional` of a `std::string`

ZXShady (changes requested):

src/SFML/Window/macOS/SFContext.mm
```diff
@@ -215,9 +215,7 @@
 
     // 1.x/2.x are mapped to 2.1 since Apple only support that legacy version.
     // >=3.0 are mapped to a 3.2 core profile.
-    const bool legacy = m_settings.majorVersion < 3;
-
-    if (legacy)
+    if (const bool legacy = m_settings.majorVersion < 3)
```
ZXShady, line 218: is this bool strictly necessary?

src/SFML/Window/Unix/JoystickImpl.cpp
```diff
@@ -315,10 +309,8 @@ unsigned int getUdevAttributeUint(udev_device* udevDevice, const std::string& at
     if (!udevDevice)
         return 0;
 
-    const char*  attribute = getUdevAttribute(udevDevice, attributeName);
-    unsigned int value     = 0;
-
-    if (attribute)
+    unsigned int value = 0;
```
ZXShady, line 312: The variable `value` could be removed and simply return the result without assigning

src/SFML/Window/Unix/JoystickImpl.cpp
```diff
@@ -300,10 +296,8 @@ unsigned int getUsbAttributeUint(udev_device* udevDevice, const std::string& att
     if (!udevDevice)
         return 0;
 
-    const char*  attribute = getUsbAttribute(udevDevice, attributeName);
-    unsigned int value     = 0;
-
-    if (attribute)
+    unsigned int value = 0;
```
ZXShady, line 299: this variable could be removed by returning directly

ZXShady (approved):

examples/island/Island.cpp
```diff
@@ -240,11 +240,10 @@ sf::Vector2f computeNormal(float left, float right, float bottom, float top)
 /// coordinates.
 ///
 ////////////////////////////////////////////////////////////
-const auto scalingFactors = sf::Vector2f(windowSize).componentWiseDiv(sf::Vector2f(resolution));
-
 sf::Vertex computeVertex(sf::Vector2u position)
 {
-    sf::Vertex vertex;
+    static const auto scalingFactors = sf::Vector2f(windowSize).componentWiseDiv(sf::Vector2f(resolution));
```
vittorioromeo, line 245: These could all be made `constexpr`

vittorioromeo (approved):
LGTM

examples/island/Island.cpp
ZXShady: this is not related to the pr though and it would require changing the `windowSize` to be constexpr which I would constexpr heavily unrelated

vittorioromeo: Yes, this is why I mentioned it as a drive by comment and I did not request changes.

ChrisThrasher: https://github.com/SFML/SFML/commit/5be64c3d6c2b9fccdf8ee5f4182f5e7cb55e69bd

I have a commit for this which might land in a future PR so this slight improvement won't be forgotten.

PR #3289, opened 2024-10-13
Upgrade to clang-format-17

ChrisThrasher:
## Description

We've been using clang-format-14 as our sole formatting program for a while but it's starting to get old. LLVM 19 is now out which macOS Homebrew ships. For years Homebrew has not shipped version 14 so I had to get a standalone clang-format-14 executable to keep formatting SFML. The latest stable Ubuntu release ships with versions 14 through 18 so Ubuntu hasn't quite yet abandoned version 14 but it's on the cusp. We periodically need to upgrade the version of clang-format we use to make sure developers have a somewhat easy time acquiring the right version of clang-format. I'm unsure what the situation is for shipping and packaging clang-format on Windows but surely this version is easier to get than 14.

Along with this upgrade we have access to 3 more major versions worth of new features. I chose to not touch the .clang-format file but it's worth looking at what features were added in case we'd like to take advantage of any of them. For future reference, here are the release notes for clang-format versions 15, 16, and 17.

- https://releases.llvm.org/15.0.0/tools/clang/docs/ReleaseNotes.html#clang-format
- https://releases.llvm.org/16.0.0/tools/clang/docs/ReleaseNotes.html#clang-format
- https://releases.llvm.org/17.0.1/tools/clang/docs/ReleaseNotes.html#clang-format

_Why not clang-format-18 if Homebrew and Ubuntu ship it?_

Weirdly enough, the two versions shipped on each OS have a different patch version which surprisingly results in differing formatting between the two. It would be pretty annoying if two devs, one on Debian and one on macOS, both used the "same version" of clang-format just to get different results. Version 17 does not have this problem. Homebrew and Debian ship the same version down to the patch number so that won't be a problem.

_Why does the formatting keep changing as we upgrade?_

clang-format does not try to produce identical results between releases. There are certainly unit tests which catch regressions to some extent but for some combinations of rules, there are not tests to prevent changes or perhaps the changes are intentional. My hope is that over time these "random" or perhaps unplanned changes between releases slow down but I don't have firm evidence to prove that's the case. The vertical alignment rules seem to be a hot spot for these differences. In codebases of mine that don't use vertical alignment rules I've noticed many fewer changes in clang-format results between versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -269,17 +269,17 @@ jobs:
 
   format:
     name: Formatting
-    runs-on: ubuntu-22.04
-
-    strategy:
-      fail-fast: false
+    runs-on: ubuntu-24.04
 
     steps:
+    - name: Install Clang Format
+      run: sudo apt-get install clang-format-17
+
     - name: Checkout Code
       uses: actions/checkout@v4
 
     - name: Format Code
-      run: cmake -DCLANG_FORMAT_EXECUTABLE=clang-format-14 -P cmake/Format.cmake
+      run: cmake -DCLANG_FORMAT_EXECUTABLE=clang-format-17 -P cmake/Format.cmake
 
     - name: Check Formatting
       run: git diff --exit-code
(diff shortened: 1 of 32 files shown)


vittorioromeo (approved):
LGTM, but going to let the others take a look at this one before merging

eXpl0it3r (approved):

PR #3291, opened 2024-10-16
Disallow C-style arrays

ChrisThrasher:
## Description

See https://github.com/SFML/SFML/pull/2951 for past precedent.

Now that all C style arrays are gone (minus a few in the tests which aren't a big deal) we can tell clang-tidy to forbid their use.


examples/sockets/TCP.cpp
```diff
@@ -34,17 +35,17 @@ void runTcpServer(unsigned short port)
     std::cout << "Client connected: " << socket.getRemoteAddress().value() << std::endl;
 
     // Send a message to the connected client
-    const char out[] = "Hi, I'm the server";
-    if (socket.send(out, sizeof(out)) != sf::Socket::Status::Done)
+    static constexpr std::string_view out = "Hi, I'm the server";
```
eXpl0it3r, line 38: Is `static` really necessary?

examples/vulkan/Vulkan.cpp
```diff
@@ -26,21 +26,21 @@
...
-using Matrix = float[4][4];
+using Matrix = std::array<std::array<float, 4>, 4>;
 
 // Multiply 2 matrices
 void matrixMultiply(Matrix& result, const Matrix& left, const Matrix& right)
 {
     Matrix temp;
 
-    for (int i = 0; i < 4; ++i)
+    for (std::size_t i = 0; i < 4; ++i)
```
eXpl0it3r, line 36: Sure use the matrix size, instead of magic number `4`

examples/vulkan/Vulkan.cpp
```diff
@@ -26,21 +26,21 @@
...
     {
-        for (int j = 0; j < 4; ++j)
+        for (std::size_t j = 0; j < 4; ++j)
```
eXpl0it3r, line 38: Sure use the matrix size, instead of magic number `4`

examples/vulkan/Vulkan.cpp
```diff
@@ -2380,7 +2385,8 @@ class VulkanExample
     void updateUniformBuffer(float elapsed)
     {
         // Construct the model matrix
-        Matrix model = {{1.0f, 0.0f, 0.0f, 0.0f}, {0.0f, 1.0f, 0.0f, 0.0f}, {0.0f, 0.0f, 1.0f, 0.0f}, {0.0f, 0.0f, 0.0f, 1.0f}};
+        Matrix model = {
+            {{1.0f, 0.0f, 0.0f, 0.0f}, {0.0f, 1.0f, 0.0f, 0.0f}, {0.0f, 0.0f, 1.0f, 0.0f}, {0.0f, 0.0f, 0.0f, 1.0f}}};
```
eXpl0it3r, line 2389: We could consider disabling formatting here, to show the matrix

src/SFML/Audio/SoundFileReaderFlac.cpp
```diff
@@ -40,7 +40,7 @@
 
 namespace
 {
-FLAC__StreamDecoderReadStatus streamRead(const FLAC__StreamDecoder*, FLAC__byte buffer[], std::size_t* bytes, void* clientData)
+FLAC__StreamDecoderReadStatus streamRead(const FLAC__StreamDecoder*, FLAC__byte* buffer, std::size_t* bytes, void* clientData)
```
eXpl0it3r, line 43: I’d rather retain the official function signature. Yes, it’s equivalent, but I don’t agree with changing it, just to make our tooling happy.

test/System/String.test.cpp
```diff
@@ -57,6 +57,8 @@ struct StringMaker<sf::U8String>
 };
 } // namespace Catch
 
+// NOLINTBEGIN(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays)
```
eXpl0it3r, line 60: Why should it not apply here?

test/Network/Packet.test.cpp
```diff
@@ -239,7 +239,7 @@ TEST_CASE("[Network] sf::Packet")
 
         SECTION("char*")
         {
-            const char string[] = "testing";
+            const char string[] = "testing"; // NOLINT(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays)
```
eXpl0it3r, line 242: It’s kind of funny how the section uses `char*`, but then in the code we use `const char string[]`. Given that we test C-style strings here, which are usually written as `char*`, wouldn’t it make sense to just use `const char* string`?

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -863,8 +864,8 @@ void WindowImplWin32::processEvent(UINT message, WPARAM wParam, LPARAM lParam)
                     if ((character >= 0xDC00) && (character <= 0xDFFF))
                     {
                         // Convert the UTF-16 surrogate pair to a single UTF-32 value
-                        std::uint16_t utf16[] = {m_surrogate, static_cast<std::uint16_t>(character)};
-                        sf::Utf16::toUtf32(utf16, utf16 + 2, &character);
+                        std::array utf16 = {m_surrogate, static_cast<std::uint16_t>(character)};
+                        sf::Utf16::toUtf32(utf16.data(), utf16.data() + 2, &character);
```
eXpl0it3r, line 868: Do we want to keep that pointer math?

src/SFML/Window/Win32/WglContext.cpp
```diff
@@ -295,21 +295,27 @@ int WglContext::selectBestPixelFormat(HDC deviceContext, unsigned int bitsPerPix
...
-                               GL_TRUE,
-                               WGL_SUPPORT_OPENGL_ARB,
-                               GL_TRUE,
-                               WGL_DOUBLE_BUFFER_ARB,
-                               GL_TRUE,
-                               WGL_PIXEL_TYPE_ARB,
-                               WGL_TYPE_RGBA_ARB,
-                               0,
-                               0};
+        std::array intAttributes =
```
eXpl0it3r, line 298:
```suggestion
        const std::array intAttributes =
```

test/System/String.test.cpp
ChrisThrasher: Honestly just being a bit lazy. No real excuse for not converting these C arrays as well.

examples/vulkan/Vulkan.cpp
```diff
@@ -1242,9 +1243,9 @@ class VulkanExample
         for (std::size_t i = 0; i < swapchainFramebuffers.size(); ++i)
         {
             // Each framebuffer consists of a corresponding swapchain image and the shared depth image
-            VkImageView attachments[] = {swapchainImageViews[i], depthImageView};
+            std::array attachments = {swapchainImageViews[i], depthImageView};
```
ZXShady, line 1246:
```suggestion
            const std::array attachments = {swapchainImageViews[i], depthImageView};
```

src/SFML/Graphics/Shader.cpp
```diff
@@ -756,11 +756,11 @@ void Shader::setUniformArray(const std::string& name, const Glsl::Mat3* matrixAr
 ////////////////////////////////////////////////////////////
 void Shader::setUniformArray(const std::string& name, const Glsl::Mat4* matrixArray, std::size_t length)
 {
-    const std::size_t matrixSize = 4 * 4;
+    static const std::size_t matrixSize = matrixArray[0].array.size();
```
ZXShady, line 759:
```suggestion
    static constexpr std::size_t matrixSize = matrixArray[0].array.size();
```

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1856,10 +1856,10 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
                         // There might be more than 1 characters in this event,
                         // so we must iterate it
                         std::uint32_t unicode = 0;
-                        std::uint8_t* iter    = keyBuffer;
-                        while (iter < keyBuffer + length)
+                        std::uint8_t* iter    = keyBuffer.data();
```
ZXShady, line 1859: I think we should use iterators here.

ZXShady (changes requested):

examples/vulkan/Vulkan.cpp
```diff
@@ -1432,7 +1433,7 @@ class VulkanExample
         VkBuffer       stagingBuffer       = {};
         VkDeviceMemory stagingBufferMemory = {};
 
-        if (!createBuffer(sizeof(vertexData),
+        if (!createBuffer(vertexData.size(),
```
ZXShady, line 1436: This is not correct, this gives the vertex count but not the size of the vertex this should be multiplied by `sizeof(vertexData[0]);` or continue using `sizeof`

examples/vulkan/Vulkan.cpp
```diff
@@ -1455,13 +1456,13 @@ class VulkanExample
...
         // Copy the vertex data into the buffer
-        std::memcpy(ptr, vertexData.data(), sizeof(vertexData));
+        std::memcpy(ptr, vertexData.data(), vertexData.size());
 
         // Unmap the buffer
         vkUnmapMemory(device, stagingBufferMemory);
 
         // Create the GPU local vertex buffer
-        if (!createBuffer(sizeof(vertexData),
+        if (!createBuffer(vertexData.size(),
```
ZXShady, line 1465: This is not correct, this gives the vertex count but not the size of the vertex this should be multiplied by sizeof(vertexData[0]); or continue using sizeof

examples/vulkan/Vulkan.cpp
```diff
@@ -1445,7 +1446,7 @@ class VulkanExample
         void* ptr = nullptr;
 
         // Map the buffer into our address space
-        if (vkMapMemory(device, stagingBufferMemory, 0, sizeof(vertexData), 0, &ptr) != VK_SUCCESS)
```
ZXShady, line 1448: This is not correct, this gives the vertex count but not the size of the vertex this should be multiplied by sizeof(vertexData[0]); or continue using sizeof

ZXShady (changes requested):
Changing some `sizeof` to `.size` will calculate the wrong thing

src/SFML/Graphics/Shader.cpp
ChrisThrasher: `matrixArray[0]` is not a compile time expression

src/SFML/Window/Unix/WindowImplX11.cpp
ChrisThrasher: Probably wise but I may hold off on that change for now. I think in general we should audit SFML for the use of pointer arithmetic where iterators would be idiomatic.

examples/sockets/TCP.cpp
eXpl0it3r: Does the compiler not already optimize this?

ZXShady (approved):
Looks good to me.

examples/sockets/TCP.cpp
ChrisThrasher: I'm not sure. In some cases it may be smart enough to do it. I'm really not sure how to check that without looking at the generated code.

src/SFML/Window/DRM/DRMContext.cpp
```diff
@@ -751,7 +759,7 @@ void DRMContext::destroySurface()
 EGLConfig DRMContext::getBestConfig(EGLDisplay display, unsigned int bitsPerPixel, const ContextSettings& settings)
 {
     // Set our video settings constraint
-    const EGLint attributes[] =
+    const std::array attributes =
     { EGL_BUFFER_SIZE,
```
eXpl0it3r, line 763: That’s really odd formatting, but I guess that’s what clang format wants. 😄

src/SFML/Window/Unix/KeyboardImpl.cpp
```diff
@@ -481,10 +481,10 @@ void ensureMapping()
             continue;
         }
 
-        std::memcpy(name, descriptor->names->keys[keycode].name, XkbKeyNameLength);
+        std::memcpy(name.data(), descriptor->names->keys[keycode].name, XkbKeyNameLength);
```
eXpl0it3r, line 484: Kind of wonder whether we can’t just construct a string out of it, instead of doing some memcpy…

eXpl0it3r (approved):

PR #3303, opened 2024-11-09
Close the open window first

eXpl0it3r:
## Description

This prevents race conditions where for a brief moment two different window implementations can exist and cause uniqueness issues. Or likely rather related to having two contexts "active" at the same time.

@Lorenzooone reported on Discord that they sometimes get some OpenGL errors when recreating a window with SFML 3

Windows
image

Linux
image

## Tasks

-   [ ] Tested on Linux
-   [x] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        {
                window.close();
            if (event->is<sf::Event::KeyPressed>())
                window.create(sf::VideoMode({1280, 720}), "Minimal, complete and verifiable example");
        }

        window.clear();
        window.display();
    }
}
```

### DIFF
--- a/src/SFML/Window/Window.cpp
+++ b/src/SFML/Window/Window.cpp
@@ -84,6 +84,9 @@ void Window::create(VideoMode mode, const String& title, std::uint32_t style, St
 ////////////////////////////////////////////////////////////
 void Window::create(VideoMode mode, const String& title, std::uint32_t style, State state, const ContextSettings& settings)
 {
+    // Ensure the open window is closed first
+    close();
+
     // Recreate the window implementation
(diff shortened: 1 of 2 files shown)


binary1248 (approved):

ChrisThrasher (approved):
Do we merge the into 2.6.x as well?

eXpl0it3r:
2.6 already does this. We changed it for SFML 3, since the unique_ptr replaces the and thus destroys the old impl, but at the time of creation we end up with two instances

PR #3307, opened 2024-11-11
Add more robust tests for `sf::Music::setLoopPoints`

ChrisThrasher:
## Description

While writing https://github.com/SFML/SFML-Website/pull/238 I realized that we don't test for certain edge cases.

### DIFF
--- a/test/Audio/Music.test.cpp
+++ b/test/Audio/Music.test.cpp
@@ -241,11 +241,42 @@ TEST_CASE("[Audio] sf::Music", runAudioDeviceTests())
 
     SECTION("setLoopPoints()")
     {
-        sf::Music music("Audio/killdeer.wav");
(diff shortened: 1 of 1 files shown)


ChrisThrasher:
Screenshot 2024-11-11 at 2 18 40 PM

I'm surprised to see zero code coverage changed. There are no other tests which call this function, and I'm fairly certain the new tests exercise new codepaths. I expected to see an increase in coverage. Has something about our code coverage config broken since the last time we were adding lots of tests?

eXpl0it3r:
Wonder if uploading only on PRs and not on push is causing some issue?

ChrisThrasher:
> Wonder if uploading only on PRs and not on push is causing some issue?

That's a good point. It might be worth reverting https://github.com/SFML/SFML/pull/3281 to see if this problem resolves itself over time.

eXpl0it3r (approved):

PR #3317, opened 2024-11-19
Revert "Stop uploading code coverage from non-pull request CI runs"

eXpl0it3r:
## Description

This reverts commit 46a1b568a981d7854f498b417a2ab6a0e3f1ae3e as it seems to cause wrong reporting on pull request coverage info, likely due to missing baselines from the master branch "push".

This PR is related to the issue #3281

I've heard @kimci86 has some additional ideas, but maybe it's worth to revert first.

## How to test this PR?

Have CI pass and coverage report uploaded for both push and pull request.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -259,7 +259,7 @@ jobs:
     name: Finalize Coverage Upload
     needs: build
     runs-on: ubuntu-22.04
-    if: github.repository == 'SFML/SFML' && github.event_name == 'pull_request' # Only run for PRs to main repository
+    if: github.repository == 'SFML/SFML' # Disable upload in forks
 
     steps:
     - name: Coveralls Finished


ChrisThrasher (approved):
We can merge a more refined solution later but for now let's just get normal code coverage numbers back.

kimci86:
> I've heard @kimci86 has some additional ideas, but maybe it's worth to revert first.

We now upload code coverage results for both push and pull request jobs. I think we are really just interested in the pull request coverage results. So uploading coverage results for push (to a PR branch) is not needed and makes CI checks less reliable because the upload sometimes fails.
But we want to compare PR coverage with target branch (usually master) and that info was missing when we did not upload on push.
So I think we could minimize coverage uploads to avoid some flakiness while keeping comparison with master working by uploading coverage result on pull request events (any branch) and push events on master branch only.

PR #3323, opened 2024-11-24
Update migration guide

ChrisThrasher:
## Description

Mentioned the change in how `sf::Rect<T>` is defined.


migration.md
```diff
@@ -85,6 +85,9 @@ circle.setPosition({10, 20});
 sf::IntRect rect({250, 400}, {50, 100});
 ```
 
+This change extends to the data members of `sf::Rect<T>`.
+The `top`, `left`, `width`, and `height` data members were replaced with two `position` and `size` data members of type `sf::Vector2<T>`.
+
```
eXpl0it3r, line 90: We have a dedicated section for `sf::Rect<T>` it believe it makes more sense to just mention it there in more detail:

```
`sf::Rect<T>` has been refactored from 4 scalar values into 2 `sf::Vector2<T>`s.
Its two data members are `sf::Vector2<T>`s named `position` and `size`.
This means that `sf::Rect<T>::getPosition()` and `sf::Rect<T>::getSize()` have been removed in favor of directly accessing the `position` and `size` data members.
The 4-parameter constructor was also removed in favor of the constructor which takes two `sf::Vector2<T>`s.
```
vs
```
`sf::Rect<T>` has been refactored from the four scalar values `top`, `left`, `width`, and `height` into two `sf::Vector2<T> `position` and `size` data members.
Its two data members are `sf::Vector2<T>`s named `position` and `size`.
This means that `sf::Rect<T>::getPosition()` and `sf::Rect<T>::getSize()` have been removed in favor of directly accessing the `position` and `size` data members.
The 4-parameter constructor was also removed in favor of the constructor which takes two `sf::Vector2<T>`s.
```

Or similar

eXpl0it3r (changes requested):

migration.md
ChrisThrasher: Oh my bad I forgot we have a dedicated section for `Rect<T>` changes.

kimci86 (approved):

PR #3326, opened 2024-11-24
Add more tests for `sf::SoundStream`

ChrisThrasher:
## Description

A lot of these tests are copied from #3325 but with different return values


test/Audio/SoundStream.test.cpp
```diff
@@ -8,7 +8,7 @@
 
 namespace
 {
-class SoundStream : public sf::SoundStream
+struct SoundStream : sf::SoundStream
```
eXpl0it3r, line 11: I'd stay consistent and not do this. Also most of these methods should actually remain private in their implementation.

eXpl0it3r (approved):

Issue #3330, opened 2024-11-28
Regression: MacOS fullscreen mode 2.6.1 -> 2.6.2


md81544:
### Prerequisite Checklist

- [X] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [X] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [X] I am here to report an issue and not to just ask a question or look for help (use the [forum](https://en.sfml-dev.org/forums/index.php#c3) or [Discord](https://discord.gg/nr4X7Fh) instead)

### Describe your issue here

On ARM64 MacOS, 2.6.1 worked as expected in full screen mode. In 2.6.2, when I enter full screen mode, there is a black border around the writeable area.

### Your Environment

- OS / distro / window manager: MacOS Sequoia
- SFML version: 2.6.1 and 2.6.2
- Compiler / toolchain: clang++
- Special compiler / CMake flags: standard cmake build

### Steps to reproduce

I built the following small app against both 2.6.1 and 2.6.2

```cpp
#include <SFML/Graphics.hpp>
#include <SFML/Window.hpp>

#include <filesystem>
#include <iostream>
#include <vector>

int main(int, char* argv[])
{
    // Get the location of the executable (this will follow symlinks)
    std::filesystem::path executable(argv[0]);
    std::filesystem::path cwd = executable.remove_filename();

    std::vector<sf::VideoMode> modes = sf::VideoMode::getFullscreenModes();
    if (modes.empty()) {
        std::cout << "Empty set of modes returned\n";
    } else {
        for (std::size_t i = 0; i < modes.size(); ++i) {
            sf::VideoMode mode = modes[i];
            std::cout << "Mode #" << i << ": " << mode.width << "x" << mode.height << " - "
                      << mode.bitsPerPixel << " bpp" << std::endl;
        }
    }

    sf::RenderWindow window;
    if (!modes.empty()) {
        window.create(
            sf::VideoMode(modes[0].width, modes[0].height), "Test", sf::Style::Fullscreen);
    } else {
        // So we fall back to windowed. Any attempt at fullscreen will cause a segfault
        // owing to the bug outlined in my report at https://github.com/SFML/SFML/issues/2300
        window.create(sf::VideoMode(1280, 800), "Test");
    }
    window.setFramerateLimit(24);

    sf::Font font;
[...]

PR #3331, opened 2024-11-28
Add Desktop mode to valid MacOS Fullscreen modes

Lorenzooone:
## Description

Adds the current Desktop mode to the valid Fullscreen Modes (which makes sense, as being <= of the Desktop mode is the parameter for being a valid Fullscreen Mode).

MacOS fullscreen mode just takes the desktop mode and creates black borders around it to match the wanted resolution, in its current implementation. So this change is fully compatible.

The change also has the benefit of making it so there is always going to be a valid Fullscreen mode, unless a catastrofic error occurred.
I didn't add this line of code in the previous PR because I thought for sure the current Desktop mode would also be a Fullscreen mode. However, that doesn't seem to be guaranteed.

This PR is related to the issue #3330 .

Should be ported to master as well.

## Tasks

-   [X] Tested on macOS

## How to test this PR?

You probably need to set some intermediate HiDPI mode, like the user who reported the issue and tested the fix.
As for the code, look at issue #3330 .
I say probably because I couldn't experience this...

### DIFF
--- a/src/SFML/Window/OSX/VideoModeImpl.cpp
+++ b/src/SFML/Window/OSX/VideoModeImpl.cpp
@@ -51,6 +51,7 @@ std::vector<VideoMode> VideoModeImpl::getFullscreenModes()
     }
 
     VideoMode desktop = getDesktopMode();
+    modes.push_back(desktop);
 
     // Loop on each mode and convert it into a sf::VideoMode object.
     const CFIndex modesCount = CFArrayGetCount(cgmodes);


ChrisThrasher:
I did some more testing and can now clearly see that when you set your MacBook display to the highest resolution setting, the desktop resolution does not appear in the list of fullscreen modes. I'm not sure why this is, but this PR should address that.

ChrisThrasher (approved):

ChrisThrasher:

PR #3332, opened 2024-11-28
Use `char32_t` for UTF-32 characters

### LINKED ISSUE #3324: Complete transition from `std::uint32_t` to `char32_t` for UTF-32 characters
In https://github.com/SFML/SFML/pull/2480 we started using C++11's `char32_t` keyword to represent UTF-32 characters. This aligns with how `std::u32string` is a specialization of `std::basic_string<T>` with `char32_t`. The `sf::String` interface was changed to use `char32_T` as well. 

However, there are many places where we still use `std::uint32_t` to represent UTF-32 characters both in the public API and the private implementation.

`sf::Event::TextEntered` still uses `std::uint32_t`
https://github.com/SFML/SFML/blob/d78cec0236d066275b63b65de069ca64d734034b/include/SFML/Window/Event.hpp#L86-L89

Many `sf::Utf` functions are using `std::uint32_t`
https://github.com/SFML/SFML/blob/d78cec0236d066275b63b65de069ca64d734034b/include/SFML/System/Utf.hpp#L72-L73

This branch shows a number internal uses of `std::uint32_t` and there are even more that I haven't yet changed.
https://github.com/SFML/SFML/compare/master...char32_t

This is an API inconsistency that ought to be rectified. The problem is, with [SFML 3's release imminent](https://github.com/SFML/SFML/issues/3314), I'm unsure if it's worth trying to finalize this work and merge it. It represents a rather small risk but the library is also in a very risk-averse phase right before our biggest release ever.

A smaller and safer change would be to simply change `TextEntered` to use `char32_t` and nothing else. This means our events are consistent with `sf::String`. I'm unsure about changing `sf::Utf` nor changing internal uses of `std::uint32_t` to `char32_t`. That is a larger and riskier change. I'm no expert in string conversion so I can't be confident I'd make that refactor without introducing a subtle regression.

---

Expanding the scope of this issue, I believe there may also be places we're using `std::uint16_t` for UTF-16 characters which ought to be replaced with C++11's `char16_t` keyword. I have not looked into this yet but I believe it's another minor inconsistency or imperfection worth fixing at some point.

ChrisThrasher:
## Description

Closes #3324

This is probably an oversight from https://github.com/SFML/SFML/pull/2480

### DIFF
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -338,7 +338,7 @@ const Font::Info& Font::getInfo() const
 
 
 ////////////////////////////////////////////////////////////
-const Glyph& Font::getGlyph(std::uint32_t codePoint, unsigned int characterSize, bool bold, float outlineThickness) const
+const Glyph& Font::getGlyph(char32_t codePoint, unsigned int characterSize, bool bold, float outlineThickness) const
 {
     // Get the page corresponding to the character size
     GlyphTable& glyphs = loadPage(characterSize).glyphs;
@@ -362,7 +362,7 @@ const Glyph& Font::getGlyph(std::uint32_t codePoint, unsigned int characterSize,
 
 
 ////////////////////////////////////////////////////////////
-bool Font::hasGlyph(std::uint32_t codePoint) const
+bool Font::hasGlyph(char32_t codePoint) const
 {
     return FT_Get_Char_Index(m_fontHandles ? m_fontHandles->face : nullptr, codePoint) != 0;
 }
@@ -503,7 +503,7 @@ Font::Page& Font::loadPage(unsigned int characterSize) const
 
 
 ////////////////////////////////////////////////////////////
-Glyph Font::loadGlyph(std::uint32_t codePoint, unsigned int characterSize, bool bold, float outlineThickness) const
+Glyph Font::loadGlyph(char32_t codePoint, unsigned int characterSize, bool bold, float outlineThickness) const
 {
     // The glyph to return
     Glyph glyph;
--- a/src/SFML/Window/Unix/WindowImplX11.cpp
+++ b/src/SFML/Window/Unix/WindowImplX11.cpp
@@ -1855,7 +1855,7 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
                     {
                         // There might be more than 1 characters in this event,
                         // so we must iterate it
-                        std::uint32_t unicode = 0;
+                        char32_t      unicode = 0;
                         std::uint8_t* iter    = keyBuffer.data();
                         while (iter < keyBuffer.data() + length)
                         {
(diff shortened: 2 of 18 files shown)


src/SFML/Graphics/Font.cpp
```diff
@@ -503,7 +503,7 @@ Font::Page& Font::loadPage(unsigned int characterSize) const
 
 
 ////////////////////////////////////////////////////////////
-Glyph Font::loadGlyph(std::uint32_t codePoint, unsigned int characterSize, bool bold, float outlineThickness) const
+Glyph Font::loadGlyph(char32_t codePoint, unsigned int characterSize, bool bold, float outlineThickness) const
```
eXpl0it3r, line 506: Is `char32_t` implicitly convertible to `FT_ULong` which is `unsigned long`?

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1870,7 +1870,7 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
                     static XComposeStatus status;
                     std::array<char, 16>  keyBuffer{};
                     if (XLookupString(&windowEvent.xkey, keyBuffer.data(), keyBuffer.size(), nullptr, &status))
-                        pushEvent(Event::TextEntered{static_cast<std::uint32_t>(keyBuffer[0])});
+                        pushEvent(Event::TextEntered{static_cast<char32_t>(keyBuffer[0])});
```
eXpl0it3r, line 1873: Kind of wonder if this is correct. Not relevant for this PR, but assuming this return UTF-8 (only thing fitting `char`), I guess we'd need to actually convert it like we do for the other implementations, instead of just casting it.

eXpl0it3r (changes requested):
I'm only unsure about the Glyph/Font. Because the implementation doesn't seem to change at all, even though you have a different type. 🤔

src/SFML/Window/Unix/WindowImplX11.cpp
ChrisThrasher: I assume that static casting to character types maintains the integer value of the character so if that's what we were doing before, I suspect that's still the correct thing to do now.

src/SFML/Graphics/Font.cpp
ChrisThrasher: I suppose so? We have all the compiler warnings cranked up so I'm pretty sure we'd get a CI failure if that conversion was potentially lossy.

src/SFML/Window/Unix/WindowImplX11.cpp
eXpl0it3r: With "Not relevant for this PR" I referred to the current implementation and not to the change.

Since I assume Linux returns UTF-8 `char` and we just do a cast to `char32_t` and represent it as UTF-32 character, I doubt that his is correct. (Completely unrelated whether it's `std::uint32_t` or `char32_t`.)

eXpl0it3r (approved):

src/SFML/Window/Unix/WindowImplX11.cpp
ChrisThrasher: Oh I see what you mean now. My bad.

PR #3333, opened 2024-12-01
Fixes for the Xorg dummy driver

jcowgill:
## Description

Some fixes for some small bugs I found while playing around running the SFML tests with the Xorg dummy video driver.

* The first commit fixes a `BadAtom` error if we cache a `None` atom. This could happen if `onlyIfExists == true` and if we're the very first client to connect to the sever so some atoms don't exist yet.
* The second commit fixes a segfault which happens if `XOpenIM` fails.

## Tasks

-   [x] Tested on Linux (note: I have only been testing this on bleeding edge Debian sid)

N/A to other OSes

### DIFF
--- a/src/SFML/Window/Unix/Display.cpp
+++ b/src/SFML/Window/Unix/Display.cpp
@@ -104,7 +104,12 @@ std::shared_ptr<_XIM> openXim()
         XSetLocaleModifiers("");
 
         // Create the input context
-        sharedXIM.reset(XOpenIM(UnixDisplayImpl::weakSharedDisplay.lock().get(), nullptr, nullptr, nullptr), XCloseIM);
+        const auto closeIM = [](XIM im)
+        {
+            if (im)
+                XCloseIM(im);
+        };
+        sharedXIM.reset(XOpenIM(UnixDisplayImpl::weakSharedDisplay.lock().get(), nullptr, nullptr, nullptr), closeIM);
         xim = sharedXIM;
 
         // Restore the previous locale
--- a/src/SFML/Window/Unix/Display.cpp
+++ b/src/SFML/Window/Unix/Display.cpp
@@ -129,7 +129,8 @@ Atom getAtom(const std::string& name, bool onlyIfExists)
 
     const auto display = openDisplay();
     const Atom atom    = XInternAtom(display.get(), name.c_str(), onlyIfExists ? True : False);
-    atoms[name]        = atom;
+    if (atom)
+        atoms[name] = atom;
 
     return atom;
 }


src/SFML/Window/Unix/Display.cpp
```diff
@@ -104,8 +104,12 @@ std::shared_ptr<_XIM> openXim()
...
 
         // Create the input context
-        sharedXIM.reset(XOpenIM(UnixDisplayImpl::weakSharedDisplay.lock().get(), nullptr, nullptr, nullptr), XCloseIM);
-        xim = sharedXIM;
+        XIM newXIM = XOpenIM(UnixDisplayImpl::weakSharedDisplay.lock().get(), nullptr, nullptr, nullptr);
+        if (newXIM)
+        {
+            sharedXIM.reset(newXIM, XCloseIM);
+            xim = sharedXIM;
+        }
```
ChrisThrasher, line 112:
```suggestion
        sharedXIM.reset(XOpenIM(UnixDisplayImpl::weakSharedDisplay.lock().get(), nullptr, nullptr, nullptr), XCloseIM);
        if (sharedXIM.reset)
            xim = sharedXIM;
```

ChrisThrasher (changes requested):

src/SFML/Window/Unix/Display.cpp
```diff
@@ -129,7 +133,8 @@ Atom getAtom(const std::string& name, bool onlyIfExists)
 
     const auto display = openDisplay();
     const Atom atom    = XInternAtom(display.get(), name.c_str(), onlyIfExists ? True : False);
-    atoms[name]        = atom;
+    if (atom)
+        atoms[name] = atom;
 
     return atom;
```
eXpl0it3r, line 139: Am I seeing this right, that we'd need to error handling in the ClipboardImpl as well?

https://github.com/SFML/SFML/blob/ae87def93e7b0280b5826b0fa5e1a8ef7156e9e7/src/SFML/Window/Unix/ClipboardImpl.cpp#L83-L87

src/SFML/Window/Unix/Display.cpp
kimci86: @ChrisThrasher This would still segfault if I understand the error correctly.
According to the commit message:
> Passing nullptr to XCloseIM segfaults, so check the result of XOpenIM first.

Maybe a simpler solution would be to have a custom deleter that checks for `nullptr` before calling `XCloseIM`, instead of passing directly `XCloseIM` as a custom deleter?

jcowgill: Yes that code would still segfault. The issue is doing this is broken:
```
sharedXIM.reset(nullptr, XCloseIM);
```

Because `unique_ptr` always calls the deleter on destruction (even if it contains a `nullptr`) and `XCloseIM` doesn't bother checking if the input is null first:
https://gitlab.freedesktop.org/xorg/lib/libx11/-/blob/master/src/xlibi18n/IMWrap.c#L128

> Maybe a simpler solution would be to have a custom deleter that checks for `nullptr` before calling `XCloseIM`, instead of passing directly `XCloseIM` as a custom deleter?

I think that's a perfectly valid solution as well.

jcowgill: > Or maybe not, taking ownership of nullptr is weird.

For some reason I was thinking about `unique_ptr` while writing the above and confusing myself a little...

For `shared_ptr` this is the strange case where `shared_ptr(nullptr)` and `shared_ptr(<pointer which may be null>)` do different things. It's the latter which calls the deleter which is what the problem is here.

src/SFML/Window/Unix/Display.cpp
jcowgill: Can you clarify this a bit? I'm not sure where you want the error handling?

src/SFML/Window/Unix/Display.cpp
ChrisThrasher: > Maybe a simpler solution would be to have a custom deleter that checks for nullptr before calling XCloseIM, instead of passing directly XCloseIM as a custom deleter?

I think this is the right solution.

src/SFML/Window/Unix/Display.cpp
eXpl0it3r: Doesn't this just delay the failure? Now the atoms are set to nullptr and then blindly used later, e.g. here: https://github.com/SFML/SFML/blob/master/src/SFML/Window/Unix/ClipboardImpl.cpp#L134-141

Or does this cause no issue?

jcowgill: Oh I see. I think we only need special handling if `onlyIfExists` is `true`. Otherwise `XInternAtom` always returns a valid atom, or you would get X error would would cause the app to exit.

So in `ClipboardImpl.cpp` it looks like `m_utf8String` is the only atom which could be null, and I think it's handled correctly with lots of `!= None` checks.

The original trigger for this patch is because SFML uses `getAtom("UTF8_STRING", true)` and `getAtom("UTF8_STRING", false)` in different places. The first call caches `None` causing the second call to unexpectedly fail.

eXpl0it3r: I trust your judgement on this 😄

src/SFML/Window/Unix/Display.cpp
jcowgill: I changed the second commit to use a custom deleter instead

ChrisThrasher:

eXpl0it3r (approved):

eXpl0it3r:
Thank you for fixing this! 🙂

PR #3339, opened 2024-12-08
Store IpAddress::m_address in host byte order

jcowgill:
## Description

I'm looking at some big-endian failures in the Debian builds. One of them concerns the `sf::IpAddress` ordering being different in big-endian systems.

I found this issue where the ordering of `sf::IpAddress` was discussed and how it's a bit strange and maybe it should be fixed in SFML 3: https://github.com/SFML/SFML/pull/2134#discussion_r901088579

Switching to host ordering fixes the ordering and has the nice side effect of working on big-endian systems "for free", so that's what I've done in this PR. If the ordering is considered part of the API, then this is a breaking change.

## Tasks

-   [x] Tested on Linux
-   [ ] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

Run this before and after to see the new ordering:

```cpp
#include <SFML/Network.hpp>
#include <algorithm>
#include <iostream>
#include <random>

int main()
{
    std::mt19937 gen;
    std::uniform_int_distribution<uint32_t> distrib;

    std::vector<sf::IpAddress> addresses;
    for (int i = 0; i < 50; i++)
        addresses.emplace_back(distrib(gen));

    std::sort(addresses.begin(), addresses.end());

    for (auto addr : addresses)
        std::cout << addr << std::endl;

    return 0;
}
```

### DIFF
--- a/test/Network/IpAddress.test.cpp
+++ b/test/Network/IpAddress.test.cpp
@@ -120,9 +120,9 @@ TEST_CASE("[Network] sf::IpAddress")
         {
             CHECK(sf::IpAddress(1) < sf::IpAddress(2));
             CHECK(sf::IpAddress(0, 0, 0, 0) < sf::IpAddress(1, 0, 0, 0));
-            CHECK(sf::IpAddress(1, 0, 0, 0) < sf::IpAddress(0, 1, 0, 0));
-            CHECK(sf::IpAddress(0, 1, 0, 0) < sf::IpAddress(0, 0, 1, 0));
-            CHECK(sf::IpAddress(0, 0, 1, 0) < sf::IpAddress(0, 0, 0, 1));
+            CHECK(sf::IpAddress(0, 1, 0, 0) < sf::IpAddress(1, 0, 0, 0));
+            CHECK(sf::IpAddress(0, 0, 1, 0) < sf::IpAddress(0, 1, 0, 0));
+            CHECK(sf::IpAddress(0, 0, 0, 1) < sf::IpAddress(0, 0, 1, 0));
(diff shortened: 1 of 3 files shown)


eXpl0it3r (approved):
Thanks fo fixing that!

eXpl0it3r:

test/Network/IpAddress.test.cpp
```diff
@@ -120,29 +120,29 @@ TEST_CASE("[Network] sf::IpAddress")
         {
             CHECK(sf::IpAddress(1) < sf::IpAddress(2));
             CHECK(sf::IpAddress(0, 0, 0, 0) < sf::IpAddress(1, 0, 0, 0));
-            CHECK(sf::IpAddress(1, 0, 0, 0) < sf::IpAddress(0, 1, 0, 0));
-            CHECK(sf::IpAddress(0, 1, 0, 0) < sf::IpAddress(0, 0, 1, 0));
-            CHECK(sf::IpAddress(0, 0, 1, 0) < sf::IpAddress(0, 0, 0, 1));
+            CHECK(sf::IpAddress(0, 1, 0, 0) < sf::IpAddress(1, 0, 0, 0));
+            CHECK(sf::IpAddress(0, 0, 1, 0) < sf::IpAddress(0, 1, 0, 0));
+            CHECK(sf::IpAddress(0, 0, 0, 1) < sf::IpAddress(0, 0, 1, 0));
```
ChrisThrasher, line 125: Would you please preserve the left hand side of these comparisons and change the right hand side as needed to make them pass?

ChrisThrasher (approved):

ChrisThrasher:

PR #3341, opened 2024-12-09
Prevent third-party files from being installed

eXpl0it3r:
## Description

I noticed while updating CSFML, that the CMake install target for SFML also includes all the third-party files - from headers, libs, to documentation files.

Starting with [CMake 3.28](https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_declare) the `EXCLUDE_FROM_ALL` option has been introduced, which will ensure that these files are not considered for the install target.

Ubuntu 24.04 ships with CMake 3.28.  
Debian 13 ships with CMake 3.30.  
See Repology: https://repology.org/project/cmake/versions

## Tasks

-   [ ] Tested on Linux
-   [x] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

Build the `install` target, e.g. `cmake --build build --config Release --target install`


eXpl0it3r:
@jcowgill if we went with CMake 3.28, what impact does this have for packaging SFML 3 in the Debian world? I guess in theory you can just patch it out, given that you will be using system dependencies 🤔

ChrisThrasher:
A solution that doesn't require raising the CMake version is to instead use [this technique](https://stackoverflow.com/a/65529161). I've used this before. It's a bit more verbose, but I find it preferable to an increase in our CMake requirement.

jcowgill:
> @jcowgill if we went with CMake 3.28, what impact does this have for packaging SFML 3 in the Debian world? I guess in theory you can just patch it out, given that you will be using system dependencies 🤔

It doesn't have any impact really. I'm packaging SFML against the development version of Debian which is already on CMake 3.31. I expect most other Linux distros will be using something fairly recent as well when packaging new library versions.

It is an issue for any users trying to compile SFML themselves on older distros though (like Ubuntu 22.04). CMake 3.28 was released December 2023, so this change drops support for all pre-2024 distros. I guess it's a question of where you draw the line.

ChrisThrasher:
https://github.com/SFML/SFML/compare/master...exclude_from_all

Here is an implementation of what I was saying. It's pretty simple and ensures Ubuntu 22.04 LTS users still have a seamless experience building the library.

PR #3354, opened 2024-12-21
Update state of development

ChrisThrasher:
## Description

Now that SFML 3 is released, there is little reason to mention the SFML 2 branch. Users can simply focus on version 3.


eXpl0it3r:
#3355 should be merged first

eXpl0it3r (approved):

PR #3358, opened 2024-12-24
Make `glCheckError` use `std::string_view` instead for file path argument

Bambo-Borris:
(Note: Discussions on this topic were done on the SFML Discord [here](https://discord.com/channels/175298431294636032/477173543281491968/1321093199640658021))

To reduce the overhead for v3 debug builds this PR alters the arguments for `glCheckError` function to take the file path as a `std::string_view` instead of a `std::filesystem::path`. The path construction in debug builds was causing noticeable performance impact for users. By swapping to `std::string_view` we can remove this overhead for the happy path and only construct the path when a gl error is actually encountered.

### DIFF
--- a/src/SFML/Graphics/GLCheck.cpp
+++ b/src/SFML/Graphics/GLCheck.cpp
@@ -36,11 +36,11 @@
 namespace sf::priv
 {
 ////////////////////////////////////////////////////////////
-bool glCheckError(const std::filesystem::path& file, unsigned int line, std::string_view expression)
+bool glCheckError(std::string_view file, unsigned int line, std::string_view expression)
 {
     const auto logError = [&](const char* error, const char* description)
     {
-        err() << "An internal OpenGL call failed in " << file.filename() << "(" << line << ")."
+        err() << "An internal OpenGL call failed in " << std::filesystem::path(file).filename() << "(" << line << ")."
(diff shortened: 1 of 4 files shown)


src/SFML/Graphics/GLCheck.cpp
```diff
@@ -36,11 +36,12 @@
...
 {
 ////////////////////////////////////////////////////////////
-bool glCheckError(const std::filesystem::path& file, unsigned int line, std::string_view expression)
+bool glCheckError(std::string_view file, unsigned int line, std::string_view expression)
 {
     const auto logError = [&](const char* error, const char* description)
     {
-        err() << "An internal OpenGL call failed in " << file.filename() << "(" << line << ")."
+        const std::filesystem::path asPath(file);
+        err() << "An internal OpenGL call failed in " << asPath.filename() << "(" << line << ")."
```
ZXShady, line 44: I would use an rvalue here instead of making a local variable.

```cpp
std::filesystem::path(file).filename()
```

ChrisThrasher:
What about making the same change in `eglCheckError`? It does the same thing.

Bambo-Borris:
> What about making the same change in `eglCheckError`? It does the same thing.

Ah yes, I forgot that footnote from ZX. Will apply to there also

ChrisThrasher (approved):

ChrisThrasher:

PR #3362, opened 2024-12-25
Hide `<filesystem>` dependency in [E]GLCheck `.cpp`

vittorioromeo:
Small compilation time improvement over #3359.

PR #3364, opened 2024-12-26
Avoid string reallocations in glCheckError

MarioLiebisch:
Didn't expect `std::filesystem::path` to have that much overhead behind the scenes (string conversion/reallocation). This replaces the `filename()` logic with a simplified approach that should be fully sufficient for any and all file paths within the SFML project following common naming conventions.

### DIFF
--- a/src/SFML/Graphics/GLCheck.cpp
+++ b/src/SFML/Graphics/GLCheck.cpp
@@ -30,18 +30,17 @@
 
 #include <SFML/System/Err.hpp>
 
-#include <filesystem>
 #include <ostream>
-
 
 namespace sf::priv
 {
 ////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


ZXShady:
But the string was only constructed in an error case unless you are erroring out so much this shouldn't matter much or does it?

MarioLiebisch:
Yes, should only happen in case there are issues (or the driver isn't ready for whatever reason, which might actually pile up on window creation on Android as far as I remember).

ChrisThrasher:
I'm not sure I can behind a performance optimization for code that only runs when an error occurs in debug builds. We're pretty far away from hot path code at this point.

eXpl0it3r:
Unless you have a specific use case where this causes issues, I also don't see it as a problem to do more work on a fail path in debug mode.

The hot path issues were already solved with #3359 and #3358 and #3362

PR #3366, opened 2024-12-26
Enable range based looping of `sf::VertexArray`

ChrisThrasher:
## Description

`begin` and `end` iterators are all that is required of a container to supported C++11 ranged-for loops. This PR also uses this feature as much as possible internally. Let me know if I missed any more places we can be using this.

### DIFF
--- a/test/Graphics/VertexArray.test.cpp
+++ b/test/Graphics/VertexArray.test.cpp
@@ -4,6 +4,7 @@
 
 #include <GraphicsUtil.hpp>
 #include <type_traits>
+#include <utility>
 
 TEST_CASE("[Graphics] sf::VertexArray")
 {
@@ -39,11 +40,11 @@ TEST_CASE("[Graphics] sf::VertexArray")
             CHECK(vertexArray.getVertexCount() == 10);
             CHECK(vertexArray.getPrimitiveType() == sf::PrimitiveType::Lines);
             CHECK(vertexArray.getBounds() == sf::FloatRect({0, 0}, {0, 0}));
-            for (std::size_t i = 0; i < vertexArray.getVertexCount(); ++i)
+            for (const auto& vertex : vertexArray)
             {
-                CHECK(vertexArray[i].position == sf::Vertex{}.position);
-                CHECK(vertexArray[i].color == sf::Vertex{}.color);
-                CHECK(vertexArray[i].texCoords == sf::Vertex{}.texCoords);
+                CHECK(vertex.position == sf::Vertex{}.position);
+                CHECK(vertex.color == sf::Vertex{}.color);
+                CHECK(vertex.texCoords == sf::Vertex{}.texCoords);
             }
         }
     }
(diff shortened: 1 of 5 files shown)


kimci86:
`VertexArray::operator[]` is not covered by tests now.

kimci86:
> This operator is still used indirectly by some other tests.

Which one? Coveralls report does not see it.

ChrisThrasher:
> > This operator is still used indirectly by some other tests.
> 
> Which one? Coveralls report does not see it.

https://github.com/SFML/SFML/blob/3c084bf66186378dc52a53175ad396b344cb6a73/test/Graphics/VertexArray.test.cpp#L78

It’s used in a few places.

kimci86:
Ok, this is because the non-const variant is used by tests but the const variant is not.

ChrisThrasher:
> Ok, this is because the non-const variant is used by tests but the const variant is not.

Ah okay this makes sense. I added tests explicitly for this case to make sure it remains covered. Thanks for catching this!

test/Graphics/VertexArray.test.cpp
```diff
@@ -111,4 +133,34 @@ TEST_CASE("[Graphics] sf::VertexArray")
...
+        SECTION("Non const")
+        {
+            for (const auto& vertex : vertexArray)
+            {
+                CHECK(vertex.position == sf::Vector2f(1, 1));
+                ++count;
+            }
+        }
+
+        CHECK(count == vertexArray.getVertexCount());
```
ZXShady, line 164: count is incremented twice it would be 2 * getVertexCount()

ZXShady (changes requested):

test/Graphics/VertexArray.test.cpp
```diff
@@ -111,4 +133,34 @@ TEST_CASE("[Graphics] sf::VertexArray")
...
+            for (const auto& vertex : std::as_const(vertexArray))
+            {
+                CHECK(vertex.position == sf::Vector2f(1, 1));
+                ++count;
+            }
+        }
+
+        SECTION("Non const")
+        {
+            for (const auto& vertex : vertexArray)
```
ZXShady, line 157: I think it is better to test whether `auto&` non const compiles because it means that the dereferenced iterators correctly return a non const lvalue reference.

ZXShady (approved):

eXpl0it3r:
Am I understanding it correctly, that this opens up additional opens to manipulate the VertexArray's vertices? Anything that takes a `begin()`-`end()` "range", right? Is there anything we need to consider in those cases?

What I mean, the PR focuses on ranged-based looping, but the feature itself goes beyond that. So I just want to make sure when people do manipulate the vertices through that interface, nothing breaks. I currently don't see anything, but maybe someone else can think of some potential problem.

ZXShady:
> Am I understanding it correctly, that this opens up additional opens to manipulate the VertexArray's vertices? Anything that takes a `begin()`-`end()` "range", right? Is there anything we need to consider in those cases?
> 
> What I mean, the PR focuses on ranged-based looping, but the feature itself goes beyond that. So I just want to make sure when people do manipulate the vertices through that interface, nothing breaks. I currently don't see anything, but maybe someone else can think of some potential problem.

they can't break anything 

begin and end are just equalivent semanticly to `&vertex_array[0]` and `&vertex_array[vertex_array.size()]` respectivly but less annoying to type.

eXpl0it3r (approved):
Seems like a useful addition 👍

PR #3367, opened 2024-12-29
Add non-const overload of `sf::Event::getIf`

ChrisThrasher:
## Description

While [upgrading SFGUI to SFML 3](https://github.com/TankOs/SFGUI/compare/master...ChrisThrasher:SFGUI:feature/sfml-three) I found a number of places where event data is being modified. This caused some problems because `sf::Event` currently provides on easy way to modify the event short of creating a brand new `Event` object then using `operator=` to assign it to the initial event. The `sf::Event` class is just some data with a type-safe interface. I see no reason to prevent easy modification of the underlying data. Most users will never need this, but for some more advanced use cases it will come in handy.

### DIFF
--- a/include/SFML/Window/Event.hpp
+++ b/include/SFML/Window/Event.hpp
@@ -325,6 +325,17 @@ public:
     ///
     ////////////////////////////////////////////////////////////
     template <typename TEventSubtype>
+    [[nodiscard]] TEventSubtype* getIf();
+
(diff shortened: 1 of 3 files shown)


vittorioromeo (approved):
Generally in favour as it seems like an unnecessary limitation to prevent mutation.

binary1248 (approved):

PR #3368, opened 2024-12-30
Suggest fix when incorrect library type is found by `find_package`

ChrisThrasher:
## Description

This is what it currently looks like when `find_package(SFML)` fails due to locating the incorrect library type:

```
CMake Error at <path>/<to>/lib/cmake/SFML/SFMLConfig.cmake:182 (message):
  Requested SFML configuration (Shared) was not found
  CMakeLists.txt:7 (find_package)

CMake Error at CMakeLists.txt:7 (find_package):
  Found package configuration file:

    <path>/<to>/lib/cmake/SFML/SFMLConfig.cmake

  but it set SFML_FOUND to FALSE so package "SFML" is considered to be NOT
  FOUND.
```

After this change the "Requested SFML configuration" line is followed by either

```
Set SFML_STATIC_LIBRARIES to ON for static libraries
```

or

```
Set SFML_STATIC_LIBRARIES to OFF for shared libraries
```

depending on the value of SFML_STATIC_LIBRARIES. This should help clear up a common source of confusion when users build SFML from source. The library follows CMake convention of building static libraries by default but our config module assumes shared libraries by default so those who build SFML from source are prone to run into this error.
ave additional steps which need to be performed, please list them as tasks!


ChrisThrasher:
Perhaps we change the config module so that if `SFML_STATIC_LIBARARIES` is empty then either static or shared libraries can be used. I think that's more user-friendly behavior that assuming an empty `SFML_STATIC_LIBARARIES` variable requires shared libraries be found. For the most part when you call `find_package(Foo)` you don't really care whether you're getting static or shared libraries. At the point of consuming a library, that much is typically out of your control so you take what you get. In SFML's case we support the use case of dual installation of static and shared libraries so we ought to treat `SFML_STATIC_LIBARARIES` as an opt-in feature for those who demand specificity but not impose that specificity on those who do not care.

eXpl0it3r:
> Perhaps we change the config module so that if `SFML_STATIC_LIBARARIES` is empty then either static or shared libraries can be used. [...] For the most part when you call `find_package(Foo)` you don't really care whether you're getting static or shared libraries.

Sounds like a good idea to me!

It's a bit odd to have to set `SFML_STATIC_LIBARARIES` to `OFF` if you want to force shared only, but as you said, usually one doesn't care to be specific.

(However, this will lead to some support requests, where people link SFML shared or static unintentionally)

eXpl0it3r:

ChrisThrasher:
> > Perhaps we change the config module so that if `SFML_STATIC_LIBARARIES` is empty then either static or shared libraries can be used. [...] For the most part when you call `find_package(Foo)` you don't really care whether you're getting static or shared libraries.
> 
> Sounds like a good idea to me!

Great, then I'll implement that in a later PR.

eXpl0it3r (approved):

PR #3373, opened 2025-01-03
Make pi/tau variables inline

JonnyPtn:
Fixes build errors when using SFML with modules which would cause conflicting linkage for these variables.

Generally best practice for constexpr variables in headers to be declared inline anyway so all uses share the same address

### DIFF
--- a/include/SFML/System/Angle.inl
+++ b/include/SFML/System/Angle.inl
@@ -34,8 +34,8 @@ namespace sf
 {
 namespace priv
 {
-constexpr float pi  = 3.141592654f;
-constexpr float tau = pi * 2.f;
+inline constexpr float pi  = 3.141592654f;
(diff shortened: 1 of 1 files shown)


ChrisThrasher:
Could you share those build errors? I was under the impression that `constexpr` implied `inline`.

JonnyPtn:
constexpr implies inline for functions but not variables.

error message would be something like

>  warning C5260: the constant variable 'sf::priv::pi' has internal linkage in an included header file context, but external linkage in imported header unit context; consider declaring it 'inline' as well if it will be shared across translation units, or 'static' to express intent to use it local to this translation unit

ChrisThrasher (approved):

ChrisThrasher:

ChrisThrasher:

Issue #3372, opened 2025-01-03
WindowBase::handleEvents does not capture events without const


Pixel-Tony:
### Prerequisite Checklist

- [X] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [x] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [X] I am here to report an issue and not to just ask a question or look for help (use the [forum](https://en.sfml-dev.org/forums/index.php#c3) or [Discord](https://discord.gg/nr4X7Fh) instead)

### Describe your issue here

When using SFML 3.x feature of passing lambdas to `handleEvents` method, if an event argument is not const-qualified, lambda is not visited when event happens, but such code compiles successfully with no warnings.

### Your Environment

- OS / distro / window manager:  Windows 10
- SFML version: 3.0.0
- Compiler / toolchain: checked with `MSVC MSBuild 17` and `MinGW G++ 13.1.0`
- Special compiler / CMake flags: `/W4` for most warnings

### Steps to reproduce

1. Use specified version
2. Try to compile next sample code:
```cpp
#include <SFML/Graphics.hpp>

int main() {
    sf::RenderWindow window(sf::VideoMode({800, 600}), "Test");
    while (window.isOpen()) {
        window.handleEvents([&](/* const */ sf::Event::Closed &) { window.close(); });
        window.clear();
        window.display();
    }
    return 0;
}
```
3. Launch executable
4. Try to close the window
5. Try to compile the same code with `const` qualifier uncommented
6. Launch executable
7. Try to close the window

### Expected behavior

steps 2-4: Window does close / code does not compile with a warning or an error specifying incorrect use of lambda as handler.
steps 5-7: Window does close.

### Actual behavior

steps 2-4: Code compiles and window cannot be closed with cross button.

Pixel-Tony:
Additionally checked for MinGW G++ 13.1.0, same behaviour

ZXShady:
This is because `sf::Event::visit` is const qualified and `WindowBase::handleEvent` has a default handler that does nothing if the arguments do not match and since `WindowClosed&` event doesn't match since it you can't pass a const type to a non const reference the default handler (does nothing) is chosen.

How to solve this issue? I honestly don't know maybe we can iterate over the handlers list and check if all of them are callable?

[...]

PR #3375, opened 2025-01-03
Provide additional checks for arguments in `WindowBase::handleEvents`

Pixel-Tony:
## Description

This PR provides fix for invalid handlers being accepted in `WindowBase::handleEvents`, as discussed in  #3372. Handlers will be checked for whether they can be invoked with at least one of existing event subtypes as an const reference argument. Thus, handlers accepting `sf::Event::<subclass> &` will be invalid, and this error will be visible to end user.

## Tasks
-   [ ] Tested on Linux
-   [x] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

Using next sample code, uncomment one line with `handleEvents` call. First line should invoke a compiler error with descriptive message, other two should compile and specified Closed event should be handled correctly (window closes).

```cpp
#include <SFML/Graphics.hpp>

int main() {
    sf::RenderWindow window(sf::VideoMode({800, 600}), "Test");
    while (window.isOpen()) {
        // Uncomment one line and compile to verify
        // window.handleEvents([&](sf::Event::Closed&) { window.close(); });       //< example does not compile, error is shown
        // window.handleEvents([&](const sf::Event::Closed&) { window.close(); }); //< example compiles
        // window.handleEvents([&](sf::Event::Closed) { window.close(); });        //< example compiles

        window.clear();
        window.display();
    }
    return 0;
}
```

### DIFF
--- a/include/SFML/Window/Event.hpp
+++ b/include/SFML/Window/Event.hpp
@@ -34,6 +34,7 @@
 
 #include <SFML/System/Vector2.hpp>
 
+#include <type_traits>
 #include <variant>
 
 
@@ -339,26 +340,26 @@ public:
     [[nodiscard]] const TEventSubtype* getIf() const;
 
     ////////////////////////////////////////////////////////////
-    /// \brief Apply a visitor to the event
+    /// \brief Apply handlers to the event
     ///
-    /// \param visitor The visitor to apply
+    /// \param handlers Handlers to apply
     ///
-    /// \return The result of applying the visitor to the event
+    /// \return The result of applying the handlers to the event
     ///
     ////////////////////////////////////////////////////////////
-    template <typename T>
-    decltype(auto) visit(T&& visitor);
+    template <typename... Handlers>
+    decltype(auto) visit(Handlers&&... handlers);
 
     ////////////////////////////////////////////////////////////
-    /// \brief Apply a visitor to the event
+    /// \brief Apply handlers to the event
     ///
-    /// \param visitor The visitor to apply
+    /// \param handlers Handlers to apply
     ///
-    /// \return The result of applying the visitor to the event
+    /// \return The result of applying the handlers to the event
     ///
     ////////////////////////////////////////////////////////////
-    template <typename T>
-    decltype(auto) visit(T&& visitor) const;
+    template <typename... Handlers>
+    decltype(auto) visit(Handlers&&... handlers) const;
 
 private:
     ////////////////////////////////////////////////////////////
@@ -395,11 +396,20 @@ private:
     template <typename T, typename... Ts>
     [[nodiscard]] static constexpr bool isInParameterPack(const std::variant<Ts...>*)
     {
-        return (std::is_same_v<T, Ts> || ...);
+        return std::disjunction_v<std::is_same<T, Ts>...>;
     }
 
     template <typename T>
-    static constexpr bool isEventSubtype = isInParameterPack<T>(decltype (&m_data)(nullptr));
+    static constexpr bool isEventSubtype = isInParameterPack<T>(static_cast<decltype(&m_data)>(nullptr));
+
+    template <typename Handler, typename... Ts>
+    static constexpr bool isInvokableWithAnyOf(std::variant<Ts...>*)
+    {
+        return std::disjunction_v<std::is_invocable<Handler, Ts&>...>;
+    }
+
+    template <typename Handler>
+    static constexpr bool isValidHandler = isInvokableWithAnyOf<Handler>(static_cast<decltype(&m_data)>(nullptr));
 };
 
 } // namespace sf
--- a/include/SFML/Window/Event.inl
+++ b/include/SFML/Window/Event.inl
@@ -38,6 +38,20 @@
 
 namespace sf
 {
+namespace priv
+{
+template <typename... Ts>
+struct OverloadSet : Ts...
+{
+    using Ts::operator()...;
+#if defined(_MSC_VER) && !defined(__clang__)
+    unsigned char dummy; // Dummy variable to ensure that this struct is not empty thus avoiding a crash due to an MSVC bug
+#endif
+};
+template <typename... Ts>
+OverloadSet(Ts...) -> OverloadSet<Ts...>;
+} // namespace priv
+
 ////////////////////////////////////////////////////////////
 template <typename TEventSubtype>
 Event::Event(const TEventSubtype& eventSubtype)
@@ -79,18 +93,20 @@ const TEventSubtype* Event::getIf() const
 
 
 ////////////////////////////////////////////////////////////
-template <typename T>
-decltype(auto) Event::visit(T&& visitor)
+template <typename... Handlers>
+decltype(auto) Event::visit(Handlers&&... handlers)
 {
-    return std::visit(std::forward<T>(visitor), m_data);
+    static_assert((isValidHandler<Handlers> && ...), "All handlers must accept a single event subtype parameter");
+    return std::visit(priv::OverloadSet{std::forward<Handlers>(handlers)...}, m_data);
 }
 
 
 ////////////////////////////////////////////////////////////
-template <typename T>
-decltype(auto) Event::visit(T&& visitor) const
+template <typename... Handlers>
+decltype(auto) Event::visit(Handlers&&... handlers) const
 {
-    return std::visit(std::forward<T>(visitor), m_data);
+    static_assert((isValidHandler<Handlers> && ...), "All handlers must accept a single event subtype parameter");
+    return std::visit(priv::OverloadSet{std::forward<Handlers>(handlers)...}, m_data);
 }
 
 } // namespace sf
--- a/include/SFML/Window/WindowBase.inl
+++ b/include/SFML/Window/WindowBase.inl
@@ -35,17 +35,6 @@ namespace sf
 {
 namespace priv
 {
-template <typename... Ts>
-struct OverloadSet : Ts...
-{
-    using Ts::operator()...;
-#if defined(_MSC_VER) && !defined(__clang__)
-    unsigned char dummy; // Dummy variable to ensure that this struct is not empty thus avoiding a crash due to an MSVC bug
-#endif
-};
-template <typename... Ts>
-OverloadSet(Ts...) -> OverloadSet<Ts...>;
-
 struct DelayOverloadResolution
 {
     template <typename T>
@@ -55,21 +44,14 @@ struct DelayOverloadResolution
 };
 } // namespace priv
 
-
 ////////////////////////////////////////////////////////////
-template <typename... Ts>
-void WindowBase::handleEvents(Ts&&... handlers) // NOLINT(cppcoreguidelines-missing-std-forward)
+template <typename... Handlers>
+void WindowBase::handleEvents(Handlers&&... handlers)
 {
-    static_assert(sizeof...(Ts) > 0, "Must provide at least one handler");
-
-    // Disable misc-const-correctness for this line since clang-tidy
-    // complains about it even though the code would become uncompilable
-
-    // NOLINTNEXTLINE(misc-const-correctness)
-    priv::OverloadSet overloadSet{std::forward<Ts>(handlers)..., [](const priv::DelayOverloadResolution&) { /* ignore */ }};
+    static_assert(sizeof...(Handlers) > 0, "Must provide at least one handler");
 
     while (std::optional event = pollEvent())
-        event->visit(overloadSet);
+        event->visit(std::forward<Handlers>(handlers)..., [](priv::DelayOverloadResolution) { /* ignore */ });
 }
 
 } // namespace sf


include/SFML/Window/Event.hpp
```diff
@@ -389,6 +389,19 @@ class Event
 
     template <typename T>
     static constexpr bool isEventSubtype = isInParameterPack<T>(decltype (&m_data)(nullptr));
+
+    template<class T, class ...Ts>
```
ChrisThrasher, line 393:
```suggestion
    template<class Callable, class ...Ts>
```
Let's name these callable template parameters. The template argument name `T` is already used in this file to refer to event subtypes so let's try to disambiguate a bit.

ChrisThrasher:
```diff
diff --git a/test/Window/WindowBase.test.cpp b/test/Window/WindowBase.test.cpp
index 66b01bef..470e0d2b 100644
--- a/test/Window/WindowBase.test.cpp
+++ b/test/Window/WindowBase.test.cpp
@@ -213,6 +213,7 @@ TEST_CASE("[Window] sf::WindowBase", runDisplayTests())
 
         // Should compile if user provides only a specific handler
         windowBase.handleEvents([](const sf::Event::Closed&) {});
+        windowBase.handleEvents([](sf::Event::Closed) {});
 
         // Should compile if user provides only a catch-all
         windowBase.handleEvents([](const auto&) {});
```
Could you please add this line to the tests? Otherwise this PR is pretty cool. I tested it out with some other things. It also prevents issues like providing `[](int&){}` as a callable which is currently possible.

By the way there are some minor formatting issues. Typically I would fix these formatting issues on your behalf but because you submitted your PR from your fork's master branch, I'm not allowed to make any changes.

include/SFML/Window/Event.hpp
ZXShady, line 393: sfml uses `typename` instead of `class` I would like to keep that convention

ChrisThrasher: Agreed

Pixel-Tony:
> By the way there are some minor formatting issues. Typically I would fix these formatting issues on your behalf but because you submitted your PR from your fork's master branch, I'm not allowed to make any changes.

@ChrisThrasher I will gladly fix them right now, this is my first PR and I hope it won't be the last one.

Pixel-Tony:
I noticed diff in workflow output, will fix in a second

include/SFML/Window/WindowBase.inl
```diff
@@ -62,6 +62,9 @@ void WindowBase::handleEvents(Ts&&... handlers) // NOLINT(cppcoreguidelines-miss
 {
     static_assert(sizeof...(Ts) > 0, "Must provide at least one handler");
 
+    static_assert((Event::isValidHandler<Ts>() && ...),
```
ChrisThrasher, line 65:
```suggestion
    static_assert((Event::isValidHandler<Ts>() && ...),
```

ChrisThrasher (changes requested):

include/SFML/Window/Event.hpp
```diff
@@ -389,6 +389,19 @@ class Event
...
     static constexpr bool isEventSubtype = isInParameterPack<T>(decltype (&m_data)(nullptr));
+
+    template <typename Callable, typename... Ts>
+    static constexpr bool isInvokableWithAnyOf(std::variant<Ts...>*)
+    {
+        return (std::is_invocable_v<Callable, const Ts&> || ...);
+    }
+
+public:
+    template <typename Callable>
```
ChrisThrasher, line 400:
```suggestion
    template <typename Handler>
```
Sorry for the churn but I just realized this name makes more sense given the function's name. I'd also just fix this myself to save you the effort but again, I can't edit PR's submitted from someone's `master` branch :(

include/SFML/Window/WindowBase.inl
```diff
@@ -62,6 +62,9 @@ void WindowBase::handleEvents(Ts&&... handlers) // NOLINT(cppcoreguidelines-miss
...
+                  "All event handlers must accept single parameter, either a const reference or a value");
+
```
ChrisThrasher, line 67:
```suggestion
    static_assert(sizeof...(Ts) > 0, "Must provide at least one handler");
    static_assert((Event::isValidHandler<Ts>() && ...),
                  "All event handlers must accept a single parameter, either a const reference or a value");

```

ChrisThrasher (changes requested):
Last round of nitpicks but otherwise this looks good!

ChrisThrasher (dismissed):

ChrisThrasher:

include/SFML/Window/Event.hpp
Pixel-Tony: This way I finally set up "trim trailing whitespace" in CLion, so the hustle gave me something good too :D

include/SFML/Window/Event.hpp
```diff
@@ -389,6 +389,19 @@ class Event
...
+
+    template <typename Handler, typename... Ts>
+    static constexpr bool isInvokableWithAnyOf(std::variant<Ts...>*)
+    {
+        return (std::is_invocable_v<Handler, const Ts&> || ...);
+    }
+
+public:
+    template <typename Handler>
+    static constexpr bool isValidHandler()
```
kimci86, line 401: Is it worth making this public? It could as well be hidden inside `handleEvents`.

kimci86: On second thought, I am not sure how to hide it.

Pixel-Tony: That's my concern: possible event variants are stored in Event but needed in Window. We *could* (*) transfer the whole multilambda composition process from `WindowBase.hpp` to `Event.hpp`, but I wasn't sure about it so I left it as is. 

It sure would be nice to not expose unneeded things, and we already have `Event::visit` method that is only used in `WindowBase`, so I think (*) is a viable option. I am open to suggestions and opinions.

kimci86: We expose `Event::visit` on purpose. It is meant to be part of the public `Event` interface.
On the other hand, `isValidHandler` is an implementation detail needed by `WindowBase::handleEvents` only.
Maybe the least dirty solution is to add `friend class WindowBase;` into `Event`?

Pixel-Tony: We could add an overload for `visit` method with variadic arguments, move all static checks to the overload along with adding a default handler (created currently in `WindowBase`), and expose it too for people to use. What do you think?

kimci86: That is possible as long as we pick a different name. Otherwise `handleEvents` with one handler would use the normal visit instead of the variadic overload, and fail to compile. The question is, which name?

Honestly I am not a fan of the `handleEvents` method. It is doing two things at once: abstracting the event loop, and providing helper types to make a visitor from lambdas. Maybe expecting users to write the event loop as before and providing helper types to make a visitor for `Event::visit` would be a better approach?

That being said, `handleEvents` is now part of the API so we have to fix it.
I still think it is the least bad to add `friend class WindowBase` and not add anything to the public API that we might regret.

ChrisThrasher: I'd maybe consider using a `friend` to hide this but I am not keen on friendships in C++. I think keeping this `public` poses little risk.

Pixel-Tony: I am not a fan of `handleEvents` too: it does take a piece of functionality which should be done by `Event`, and outside of said functionality it is just a wrapper around a loop.

ChrisThrasher:

include/SFML/Window/Event.hpp
eXpl0it3r: I guess, I don't understand enough about the C++ magic here, but why can't the function not be implemented within WindowBase?

ChrisThrasher: > but why can the function not be implemented within WindowBase

The function needs access to the type of `sf::Event::m_data` which is `private` and thus inaccessible to `sf::WindowBase`.

Pixel-Tony: We can add a `Event::visitAny`-like method and move all of the overload-creating functionality there. Then just pass it instead of visit to `WindowBase::handleEvents`. This way we keep the functionality but extend API in a non-conflicting way. 

`visitAny` would also create overloads in another way: it will add one `()` overload taking `auto param` and doing a static check that no lambda passed can be called with this parameter. This way it can only be used for non-covered subtypes and we don't need to use `DelayOverloadResolution`. Opinions?

ChrisThrasher: > We can add a Event::visitAny-like method and move all of the overload-creating functionality there. Then just pass it instead of visit to WindowBase::handleEvents. This way we keep the functionality but extend API in a non-conflicting way.

This is a really compelling solution to me. I'd love to see this implemented if you have something in mind for how to do it.

Pixel-Tony: I pushed a proposition

ChrisThrasher:

ChrisThrasher:
Please rebase on `master`. We fixed some other stuff that is currently causing the extended CI pipeline to fail.

ChrisThrasher:

ChrisThrasher:
```cpp
    SECTION("visit()")
    {
        SECTION("Exhaustive visitor")
        {
            // Using a non-void return type requires providing callbacks for all types
            CHECK(sf::Event(sf::Event::Closed{}).visit(visitor) == "Closed");
            CHECK(sf::Event(sf::Event::Resized{}).visit(visitor) == "Resized");
            CHECK(sf::Event(sf::Event::FocusLost{}).visit(visitor) == "Other");
            CHECK(sf::Event(sf::Event::FocusGained{}).visit(visitor) == "Other");
            CHECK(sf::Event(sf::Event::KeyPressed{}).visit(visitor) == "KeyPressed");
        }

        SECTION("Non-exhausive visitor")
        {
            // If you omit certain callbacks then the callbacks cannot have a return value
            const sf::Event event = sf::Event::Closed{};

            bool closedVisited = false;
            event.visit([&closedVisited](sf::Event::Closed) { closedVisited = true; });
            CHECK(closedVisited);

            bool resizedVisited = false;
            event.visit([&resizedVisited](sf::Event::Resized) { resizedVisited = true; });
            CHECK(not resizedVisited);
        }
    }
```

Would you please paste in these new `sf::Event::visit` tests? Just overwrite the existing `visit()` section in Event.test.cpp. After that I think I'm ready to approve this.

@eXpl0it3r @binary1248 any more thoughts?

include/SFML/Window/Event.inl
```diff
@@ -79,10 +93,15 @@ const TEventSubtype* Event::getIf() const
 
 
 ////////////////////////////////////////////////////////////
-template <typename T>
-decltype(auto) Event::visit(T&& visitor) const
+template <typename... Ts>
+decltype(auto) Event::visit(Ts&&... visitors) const
```
ZXShady, line 97: should the argument variable be named `visitors`?

I think a visitor means the entire overload set but this takes the handlers of the visitor so I think it should be called that.

Pixel-Tony: Correct.

include/SFML/Window/Event.hpp
```diff
@@ -339,15 +339,15 @@ class Event
...
+    /// \param handlers Handlers to apply
     ///
-    /// \return The result of applying the visitor to the event
+    /// \return The result of applying selected handler to the event
     ///
     ////////////////////////////////////////////////////////////
-    template <typename T>
-    decltype(auto) visit(T&& visitor) const;
+    template <typename... Ts>
+    decltype(auto) visit(Ts&&... handlers) const;
```
ChrisThrasher, line 350:
```suggestion
    template <typename... Handlers>
    decltype(auto) visit(Handlers&&... handlers) const;
```

include/SFML/Window/Event.hpp
```diff
@@ -339,15 +339,15 @@ class Event
     [[nodiscard]] const TEventSubtype* getIf() const;
 
     ////////////////////////////////////////////////////////////
-    /// \brief Apply a visitor to the event
+    /// \brief Apply one of handlers to the event, if applicable
```
ChrisThrasher, line 342:
```suggestion
    /// \brief Apply handlers to the event
```

include/SFML/Window/Event.hpp
```diff
@@ -339,15 +339,15 @@ class Event
...
     ///
-    /// \param visitor The visitor to apply
+    /// \param handlers Handlers to apply
     ///
-    /// \return The result of applying the visitor to the event
+    /// \return The result of applying selected handler to the event
```
ChrisThrasher, line 346:
```suggestion
    /// \return The result of applying the handlers to the event
```

include/SFML/Window/Event.inl
```diff
@@ -79,10 +93,15 @@ const TEventSubtype* Event::getIf() const
 
 
 ////////////////////////////////////////////////////////////
-template <typename T>
-decltype(auto) Event::visit(T&& visitor) const
+template <typename... Ts>
+decltype(auto) Event::visit(Ts&&... handlers) const
```
ChrisThrasher, line 97:
```suggestion
template <typename... Handlers>
decltype(auto) Event::visit(Handlers&&... handlers) const
```

include/SFML/Window/WindowBase.inl
```diff
@@ -60,16 +47,8 @@ struct DelayOverloadResolution
 template <typename... Ts>
 void WindowBase::handleEvents(Ts&&... handlers) // NOLINT(cppcoreguidelines-missing-std-forward)
```
ChrisThrasher, line 48:
```suggestion
template <typename... Handlers>
void WindowBase::handleEvents(Handlers&&... handlers) // NOLINT(cppcoreguidelines-missing-std-forward)
```
Please also update WindowBase.hpp to use this `Handlers` template name.

test/Window/Event.test.cpp
```diff
@@ -306,10 +306,30 @@ TEST_CASE("[Window] sf::Event")
...
+            bool closedVisited = false;
+            event.visit([&closedVisited](sf::Event::Closed) { closedVisited = true; },
+                        [](auto) {});
+            CHECK(closedVisited);
+
+            bool resizedVisited = false;
+            event.visit([&resizedVisited](sf::Event::Resized) { resizedVisited = true; },
+                        [](auto) {});
+            CHECK(!resizedVisited);
+        }
```
ChrisThrasher, line 333:
```suggestion
        SECTION("Non-exhausive visitor")
        {
            const auto matchAll = [](auto) {};

            // If you omit certain callbacks then the callbacks cannot have a return value
            const sf::Event event = sf::Event::Closed{};

            bool closedVisited = false;
            event.visit([&closedVisited](sf::Event::Closed) { closedVisited = true; }, matchAll);
            CHECK(closedVisited);

            bool resizedVisited = false;
            event.visit([&resizedVisited](sf::Event::Resized) { resizedVisited = true; }, matchAll);
            CHECK(!resizedVisited);
        }
```

ChrisThrasher (changes requested):

include/SFML/Window/Event.hpp
```diff
@@ -389,6 +389,15 @@ class Event
 
     template <typename T>
     static constexpr bool isEventSubtype = isInParameterPack<T>(decltype (&m_data)(nullptr));
+
+    template <typename Handler, typename... Ts>
+    static constexpr bool isInvokableWithAnyOf(std::variant<Ts...>*)
+    {
+        return (std::is_invocable_v<Handler, const Ts&> || ...);
```
vittorioromeo, line 396: Missing include for `type_traits`

vittorioromeo (changes requested):

include/SFML/Window/WindowBase.inl
```diff
@@ -57,19 +44,11 @@ struct DelayOverloadResolution
...
-
-    // Disable misc-const-correctness for this line since clang-tidy
-    // complains about it even though the code would become uncompilable
-
-    // NOLINTNEXTLINE(misc-const-correctness)
-    priv::OverloadSet overloadSet{std::forward<Ts>(handlers)..., [](const priv::DelayOverloadResolution&) { /* ignore */ }};
-
     while (const std::optional event = pollEvent())
-        event->visit(overloadSet);
+        event->visit(handlers..., [](const priv::DelayOverloadResolution&) { /* ignore */ });
```
vittorioromeo, line 51: `handlers...` should be forwarded here, and `NOLINT(cppcoreguidelines-missing-std-forward)` should be removed

vittorioromeo (dismissed):

include/SFML/Window/WindowBase.inl
```diff
@@ -28,48 +28,15 @@
...
-
-    // Disable misc-const-correctness for this line since clang-tidy
-    // complains about it even though the code would become uncompilable
-
-    // NOLINTNEXTLINE(misc-const-correctness)
-    priv::OverloadSet overloadSet{std::forward<Ts>(handlers)..., [](const priv::DelayOverloadResolution&) { /* ignore */ }};
-
     while (const std::optional event = pollEvent())
-        event->visit(overloadSet);
+        event->visit(handlers...);
```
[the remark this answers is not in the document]
Pixel-Tony: @vittorioromeo should handlers be forwarded here too? I thought `std::forward` should prevent moving when argument passed is a non-temporary value

vittorioromeo:
@ChrisThrasher: I believe this is superseded by #3396, right? Do we still want to look into adding extra checks?

ChrisThrasher:
> @ChrisThrasher: I believe this is superseded by #3396, right? Do we still want to look into adding extra checks?

There is still a discussion to be had about the ergonomic improvements this PR makes with regard to making `sf::Event::visit` a variadic function. However the bits about testing for callback parameters has to be dropped.

include/SFML/Window/Event.hpp
```diff
@@ -399,7 +400,16 @@ class Event
     }
 
     template <typename T>
-    static constexpr bool isEventSubtype = isInParameterPack<T>(decltype (&m_data)(nullptr));
+    static constexpr bool isEventSubtype = isInParameterPack<T>(static_cast<decltype(&m_data)>(nullptr));
+
+    template <typename Handler, typename... Ts>
+    static constexpr bool isInvokableWithAnyOf(std::variant<Ts...>*)
+    {
+        return (std::is_invocable_v<Handler, Ts&> || ...);
```
ChrisThrasher, line 408:
```suggestion
        return std::disjunction_v<std::is_invocable<Handler, Ts&>...>;
```
Could we do something like this instead?

include/SFML/Window/Event.inl
```diff
@@ -79,18 +93,26 @@ const TEventSubtype* Event::getIf() const
...
 ////////////////////////////////////////////////////////////
-template <typename T>
-decltype(auto) Event::visit(T&& visitor)
+template <typename... Handlers>
+decltype(auto) Event::visit(Handlers&&... handlers)
 {
-    return std::visit(std::forward<T>(visitor), m_data);
+    static_assert(sizeof...(Handlers) > 0, "Must provide at least one handler");
+    static_assert((isValidHandler<Handlers> && ...),
+                  "All event handlers must accept a single parameter, either a const reference or a value");
```
ChrisThrasher, line 101:
```suggestion
                  "All event handlers must accept a single event subtype parameter");
```

include/SFML/Window/Event.inl
```diff
@@ -79,18 +93,26 @@ const TEventSubtype* Event::getIf() const
...
 ////////////////////////////////////////////////////////////
-template <typename T>
-decltype(auto) Event::visit(T&& visitor) const
+template <typename... Handlers>
+decltype(auto) Event::visit(Handlers&&... handlers) const
 {
-    return std::visit(std::forward<T>(visitor), m_data);
+    static_assert(sizeof...(Handlers) > 0, "Must provide at least one handler");
+    static_assert((isValidHandler<Handlers> && ...),
+                  "All event handlers must accept a single parameter, either a const reference or a value");
```
ChrisThrasher, line 113:
```suggestion
                  "All event handlers must accept a single event subtype parameter");
```

test/Window/Event.test.cpp
```diff
@@ -340,5 +340,29 @@ TEST_CASE("[Window] sf::Event")
...
+            const sf::Event event = sf::Event::Closed{};
+
+            bool closedVisited = false;
+            event.visit([&closedVisited](sf::Event::Closed) { closedVisited = true; }, matchAll);
+            CHECK(closedVisited);
+
+            bool resizedVisited = false;
+            event.visit([&resizedVisited](sf::Event::Resized) { resizedVisited = true; }, matchAll);
+            CHECK(!resizedVisited);
+        }
```
ChrisThrasher, line 366:
```suggestion
```
I believe the existing tests are good enough to handle these changes. Because this PR bans certain code from compiling, we don't have much to write unit tests. We don't have a way to test that certain function arguments will not compile.

test/Window/WindowBase.test.cpp
```diff
@@ -213,6 +213,7 @@ TEST_CASE("[Window] sf::WindowBase", runDisplayTests())
 
         // Should compile if user provides only a specific handler
         windowBase.handleEvents([](const sf::Event::Closed&) {});
+        windowBase.handleEvents([](sf::Event::Closed) {});
```
ChrisThrasher, line 216:
```suggestion
```
This is no longer needed.

ChrisThrasher (changes requested):

include/SFML/Window/Event.hpp
Pixel-Tony: Yes, this feels more modern

include/SFML/Window/Event.hpp
```diff
@@ -399,7 +400,16 @@ class Event
     }
 
     template <typename T>
-    static constexpr bool isEventSubtype = isInParameterPack<T>(decltype (&m_data)(nullptr));
+    static constexpr bool isEventSubtype = isInParameterPack<T>(static_cast<decltype(&m_data)>(nullptr));
+
+    template <typename Handler, typename... Ts>
+    static constexpr bool isInvokableWithAnyOf(std::variant<Ts...>*)
+    {
+        return std::disjunction_v<std::is_invocable<Handler, Ts&>...>;
```
ChrisThrasher, line 408: Above, let's please change

```cpp
return (std::is_same_v<T, Ts> || ...);
```

to 

```cpp
return std::disjunction_v<std::is_same<T, Ts>...>;
```

ChrisThrasher (changes requested):
One nitpick + some formatting fixes and I'll be ready to approve this 👍🏻

ChrisThrasher (approved):
LGTM!

PR #3376, opened 2025-01-04
Implement Event::WindowMoved for Win32

learn-more:
-   [x] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before?
-   [x] Have you provided some example/test code for your changes?

## Description

Add an event that will be raised after a window is moved (just like the size event works).
This has been discussed / promised multiple times before, at least:
https://en.sfml-dev.org/forums/index.php?topic=2914.msg19032
https://en.sfml-dev.org/forums/index.php?topic=3598.msg23525#msg23525
https://en.sfml-dev.org/forums/index.php?topic=19658.msg141513#msg141513

## Tasks

-   [ ] Tested on Linux
-   [x] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>

int main()
{

    while (window.isOpen())
    {
        {
                window.close();
            else if (const auto* moved = event->getIf<sf::Event::WindowMoved>())
                printf("New position: %d, %d\n", moved->position.x, moved->position.y);
        }

        window.clear();
        window.display();
    }
}
```


ChrisThrasher:
What is your use case for window moved events?

learn-more:
> What is your use case for window moved events?

I wanted to use this as a basis for writing desktop tools,
in which I sometimes use this as 'overlay' over the actual desktop:
image

If this is denied because there is no use for it that would be fair as well,
but considering it was said to 'probably come in sfml2' I assumed it was still something that had at least some interest.

binary1248:
While I don't have anything against such functionality (other libraries provide similar), I still want to see at least one other operating system implementation. Designing the API around what a single operating system supports goes against the idea that SFML is a cross-platform library that provides portable abstractions.

It has to also be considered whether the event has to contain the new window position or not. While there might be valid use cases for being notified when the window has been moved, not everybody is going to be interested in its new position, and for those who are, they can just call `getPosition()` to get the value themselves.

learn-more:
> While I don't have anything against such functionality (other libraries provide similar), I still want to see at least one other operating system implementation. Designing the API around what a single operating system supports goes against the idea that SFML is a cross-platform library that provides portable abstractions.
> 
> It has to also be considered whether the event has to contain the new window position or not. While there might be valid use cases for being notified when the window has been moved, not everybody is going to be interested in its new position, and for those who are, they can just call `getPosition()` to get the value themselves.

The `WindowMoved` contains the position, because the `Resized` event contains the window size.
If you want me to change this, please let me know.

As for implementing others, I'll have to spin up a VM to get anything other than Windows running, this would probably be some linux distro.

binary1248:
Normally I tend to only add data to a structure if I know the data will be often used, but in this case I can see the consistency argument when comparing to `Resized`. The overhead shouldn't be an issue since `Moved` events should happen relatively rarely during the lifetime of an application anyway. If anybody is having performance issues due to `Moved` events being too big I think they have other problems to solve first.

learn-more:
> Normally I tend to only add data to a structure if I know the data will be used often, but in this case I can see the consistency argument when comparing to `Resized`. The overhead shouldn't be an issue since `Moved` events should happen relatively rarely during the lifetime of an application anyway. If anybody is having performance issues due to `Moved` events being too big I think they have other problems to solve first.

Rebased on the latest master to fix the merge conflicts in `Event.test.cpp`
Added an X11 implementation, and I temporarily monkeypatched the `window` example to print the coordinates:

image

include/SFML/Window/Event.hpp
```diff
@@ -366,6 +375,7 @@ class Event
     ////////////////////////////////////////////////////////////
     std::variant<Closed,
                  Resized,
+                 WindowMoved,
```
eXpl0it3r, line 378: Wonder if it should just be `Moved` given that it also isn't `WindowResized`, then again Resized doesn't stand in name conflict with say `MouseMoved`.

Any opinions on this? 🤔

binary1248: If I had to guess, `Resized` isn't `WindowResized` because it didn't require the disambiguation, whereas `Moved` does.

learn-more: I chose this name because it is more explicit about what is being moved.
Imo resized should be renamed to WindowResized, but that would probably also require a typedef for backwards compatibility

test/Window/Event.test.cpp
```diff
@@ -50,14 +55,22 @@ TEST_CASE("[Window] sf::Event")
...
             CHECK(resized.size == sf::Vector2u(1, 2));
         }
+        SECTION("Template constructor window move")
+        {
+            const sf::Event event = sf::Event::WindowMoved{{1, 2}};
+            CHECK(event.is<sf::Event::WindowMoved>());
+            CHECK(event.getIf<sf::Event::WindowMoved>());
+            const auto& windowMoved = *event.getIf<sf::Event::WindowMoved>();
+            CHECK(windowMoved.position == sf::Vector2i(1, 2));
+        }
```
ChrisThrasher, line 73:
```suggestion
```

test/Window/Event.test.cpp
```diff
@@ -50,14 +55,22 @@ TEST_CASE("[Window] sf::Event")
 
     SECTION("Construction")
     {
-        SECTION("Template constructor")
+        SECTION("Template constructor resize")
```
ChrisThrasher, line 58:
```suggestion
        SECTION("Template constructor")
```

test/Window/Event.test.cpp
```diff
@@ -307,6 +329,12 @@ TEST_CASE("[Window] sf::Event")
         REQUIRE(mouseMoved);
         mouseMoved->position = sf::Vector2i(6, 9);
         CHECK(mouseMoved->position == sf::Vector2i(6, 9));
+
+        event               = sf::Event::WindowMoved{{1, 2}};
+        auto* windowMoved   = event.getIf<sf::Event::WindowMoved>();
+        REQUIRE(windowMoved);
+        windowMoved->position = sf::Vector2i(3, 4);
+        CHECK(windowMoved->position == sf::Vector2i(3, 4));
```
ChrisThrasher, line 337:
```suggestion
```

src/SFML/Window/Unix/WindowImplX11.cpp
```diff
@@ -1757,6 +1757,13 @@ bool WindowImplX11::processEvent(XEvent& windowEvent)
                 m_previousSize.x = windowEvent.xconfigure.width;
                 m_previousSize.y = windowEvent.xconfigure.height;
             }
+            if (windowEvent.xconfigure.x != m_previousPosition.x || windowEvent.xconfigure.y != m_previousPosition.y)
+            {
+                pushEvent(Event::WindowMoved{Vector2i(windowEvent.xconfigure.x, windowEvent.xconfigure.y)});
+
+                m_previousPosition.x = windowEvent.xconfigure.x;
+                m_previousPosition.y = windowEvent.xconfigure.y;
+            }
```
ChrisThrasher, line 1766:
```suggestion
            const sf::Vector2i previousPosition(windowEvent.xconfigure.x, windowEvent.xconfigure.y);
            if (m_previousPosition != previousPosition)
            {
                pushEvent(Event::WindowMoved{previousPosition});
                m_previousPosition = previousPosition;
            }
```

ChrisThrasher (changes requested):

test/Window/Event.test.cpp
kimci86: Chris is suggesting to delete those lines.

PR #3378, opened 2025-01-05
Remove unused RESOURCES_DIR cmake option

learn-more:
-   [ ] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before?
-   [ ] Have you provided some example/test code for your changes?

## Description

Since 27a4c83ebc724a978e554e77f5f231b887e0bdd9 the examples are no longer installed.
Now the `RESOURCES_DIR` option is still parsed but is fully ignored, so remove it.

## Tasks

-   [ ] Tested on Linux
-   [x] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -280,12 +280,11 @@ endmacro()
 # example: sfml_add_example(ftp
 #                           SOURCES ftp.cpp ...
 #                           BUNDLE_RESOURCES MainMenu.nib ...    # Files to be added in target but not installed next to the executable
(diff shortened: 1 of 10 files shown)


ChrisThrasher (approved):
Good find! This also implies that we can collapse the directory structure of the resources themselves. We can simply put the resources in the same directory as the C++ code. I might do that in a later PR.

PR #3379, opened 2025-01-06
Make keyboard key/scancode count variables inline

JonnyPtn:
Similar to #3373 , We want these constants to be shared between all translation units

While this doesn't cause compiler errors immediately when using SFML header modules, it does when you try to use these variables as they have internal linkage, so you get an error like `'ScancodeCount': is not a member of 'sf::Keyboard'`


eXpl0it3r (approved):

ChrisThrasher (approved):

ChrisThrasher:

PR #3381, opened 2025-01-07
Use correct type for accessing array

JonnyPtn:
Error sporadically flagged by the buildbot freebsd checks:

> src/SFML/Window/FreeBSD/JoystickImpl.cpp:318:48: warning: implicit conversion changes signedness: 'int' to 'size_type' (aka 'unsigned long') [-Wsign-conversion]

Not immediately clear why sometimes it's not treated as an error but fix seems simple enough

### DIFF
--- a/src/SFML/Window/FreeBSD/JoystickImpl.cpp
+++ b/src/SFML/Window/FreeBSD/JoystickImpl.cpp
@@ -304,8 +304,8 @@ JoystickState JoystickImpl::JoystickImpl::update()
         if (!data)
             continue;
 
-        int        buttonIndex = 0;
-        hid_item_t item;
+        std::size_t buttonIndex = 0;
+        hid_item_t  item;
 
         while (hid_get_item(data, &item))
         {


eXpl0it3r (approved):

ChrisThrasher:
https://github.com/SFML/SFML/blob/b56604c940fa4f46214aab12157ba995c3373cfa/src/SFML/Window/NetBSD/JoystickImpl.cpp#L312

For what it's worth the NetBSD version of this file is nearly identical and does this same thing. Do we even compile the NetBSD code in the extended CI pipeline?

eXpl0it3r:
> For what it's worth the NetBSD version of this file is nearly identical and does this same thing.

Should also be fixed. It's due to changing from C-style array to `std::array`, so the they is now `std::size_t`.

> Do we even compile the NetBSD code in the extended CI pipeline?

We don't (yet?) 😄

ChrisThrasher:
https://github.com/SFML/SFML/pull/3384

PR #3382, opened 2025-01-07
Use approximation when comparing floats in tests

eXpl0it3r:
## Description

This should fix the MinGW 32-bits test failures, see: https://ci.sfml-dev.org/#/builders/18/builds/64/steps/5/logs/stdio

Question is, whether we shouldn't be doing this for *all* float tests.

## How to test this PR?

Buildbot CI succeeds


ChrisThrasher (approved):
What caused this issues to just now manifest? These tests are rather old at this point. Did y'all add a new BuiltBot CI job?

It's also worth mentioning how this change marginally reduces the `constexpr` code coverage. This doesn't matter a lot but perhaps if we pick different test values we can avoid floating point precision losses and continue to run these tests at compile time. Remember that code ran at compile time is guaranteed to be free of UB so we gain some additional assurances about the library if we do so.

eXpl0it3r:
Yes, we updated MinGW

PR #3383, opened 2025-01-09
Add a C++ language version check

eXpl0it3r:
## Description

We've had a few people not change their Visual Studio settings to `/std:c++17`. Unfortunately, the compiler sometimes only throws rather confusing errors, such as not knowing what to do with `std::optional`.

This check should provide a bit of a better error.

I'm however not sure, if this is fully portable.

## Tasks

-   [ ] Tested on Linux
-   [ ] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

Build a sample application that doesn't set C++17 or newer as language.


binary1248 (approved):

ChrisThrasher:
```
error: a type specifier is required for all declarations
    2 | static_assert(false,
      | ^
1 error generated.
```
When I compile this with AppleClang in C++14 mode, I get this error. Compilation still fails, which is what we want, but we're not getting a static assertion failure. It's just a compilation error.

Might I recommend we use an `#error` directive instead?

include/SFML/Config.hpp
```diff
@@ -111,6 +111,15 @@
 #endif
 
 
+////////////////////////////////////////////////////////////
+// Ensure correct C++ language selection
+////////////////////////////////////////////////////////////
+#if (defined(_MSVC_LANG) && _MSVC_LANG < 201703L) || (!defined(_MSVC_LANG) && __cplusplus < 201703L)
+static_assert(false,
+              "C++17 required, enable C++17 for your compiler (e.g. -std=c++17 for GCC/Clang or /std:c++17 for MSVC)");
```
ChrisThrasher, line 119:
```suggestion
#error "C++17 required. Enable C++17 for your compiler (e.g. -std=c++17 for GCC/Clang or /std:c++17 for MSVC)");
```

ChrisThrasher (changes requested):

eXpl0it3r:
> we're not getting a static assertion failure. It's just a compilation error.

That's odd, but I guess `#error` is also fine for me.

ChrisThrasher:
I don't think `static_assert(false)` is valid in C++17. Maybe in C++23 that made well-formed but I don't recall those details exactly.

include/SFML/Config.hpp
```diff
@@ -111,6 +111,14 @@
 #endif
 
 
+////////////////////////////////////////////////////////////
+// Ensure correct C++ language selection
+////////////////////////////////////////////////////////////
+#if (defined(_MSVC_LANG) && _MSVC_LANG < 201703L) || (!defined(_MSVC_LANG) && __cplusplus < 201703L)
+#error "C++17 required, enable C++17 for your compiler (e.g. -std=c++17 for GCC/Clang or /std:c++17 for MSVC)")
```
ChrisThrasher:
```suggestion
#error "C++17 or newer required. Enable via -std=c++17 for GCC/Clang or /std:c++17 for MSVC"
```
Also note the removal of the trailing paren. That would probably cause issues.

include/SFML/Config.hpp
```diff
@@ -111,6 +111,14 @@
 #endif
 
 
+////////////////////////////////////////////////////////////
+// Ensure correct C++ language selection
```
ChrisThrasher, line 115:
```suggestion
// Ensure minimum C++ language standard version is met
```

ChrisThrasher (changes requested):

ChrisThrasher (approved):

ChrisThrasher (approved):

PR #3384, opened 2025-01-09
Fix conversion warning

ChrisThrasher:
Continuing what was started in #3381


eXpl0it3r (changes requested):
Can you adjust the commit message to not include a full URL to GitHub? 
I'd just use `#3381` or leave it out completely.

(Personally, I don't like issue IDs as much in open source projects, as the Git history can theoretically live on much longer than the platform/issue tracker)

eXpl0it3r (approved):

PR #3390, opened 2025-01-16
Hack up an inital POC of a blocking eventloop

learn-more:
Messages are delivered using a callback instead of polled, this allows a window to continue to process events while it is being dragged on Windows

*Only one example builds, the new main*

There is a delay of ~1 sec before the Window is processing events when the left button is clicked on the title bar, I suspect this has to do with the window having the input capture.

https://github.com/user-attachments/assets/bad5e149-3d35-4825-8fbd-12cc9b4675e5

The main loop now looks like this:

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    auto window = sf::RenderWindow(sf::VideoMode({800u, 600u}), "SFML Event Loop");
    const sf::Font m_font{"resources/tuffy.ttf"};
    sf::Text       m_text(m_font);
    int            m_frame = 0;
    window.setFramerateLimit(60);

    window.eventLoop(
        [&](const sf::Event& evt)
        {
            if (evt.is<sf::Event::Idle>())
            {
                window.clear();
                m_frame += 1;
                m_text.setString("Frame: " + std::to_string(m_frame));
                window.draw(m_text);
                window.display();
            }
            else if (evt.is<sf::Event::Closed>())
            {
                window.close();
            }
        });
}
```


trustytrojan:
Callbacks are great, and I agree with this change given that the `pollEvent` API is still there so that the user has choice.

I wanted to share something interesting though:

SDL3 employs callbacks as an optional way to both receive events **and** enter your program: https://github.com/libsdl-org/SDL/blob/main/docs%2Fhello.c

Quite a bit cleaner because it's literally "implementing the interface" but not in an OOP context. It could be an SFML4 prospect to also employ this pattern (although still optional to use) as an entry point into SFML apps.

PR #3395, opened 2025-01-20
Allow event handlers to accept non-const references

vittorioromeo:
Alternative approach to #3375 that allows handlers to take non-const references. In other words

```cpp
while (window.isOpen())
{
    window.clear();
    window.handleEvents([&](sf::Event::Closed&) { window.close(); });
    window.display();
}
```

will compile and will work as expected (the window will be closed when `[X]` is pressed). 

Not claiming this approach is strictly better, but it seems simpler to me and it is a bit artificial to prevent specifically non-const-ref handlers to work via type-trait detection.

### DIFF
--- a/include/SFML/Window/Event.inl
+++ b/include/SFML/Window/Event.inl
@@ -82,7 +82,8 @@ const TEventSubtype* Event::getIf() const
 template <typename T>
 decltype(auto) Event::visit(T&& visitor) const
 {
-    return std::visit(std::forward<T>(visitor), m_data);
+    auto data = m_data; // Copy event to allow visitors to take it by non-const reference
+    return std::visit(std::forward<T>(visitor), data);
 }
 
 } // namespace sf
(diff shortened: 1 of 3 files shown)


ZXShady:
is there a reason for us having a const qualified visit function?

vittorioromeo:
> is there a reason for us having a const qualified visit function instead of both a const and non const like normal `std::visit` given that we also implemented one for `getIf`?

Perhaps having two overloads makes even more sense, I cannot think of a good reason why `Event` should be `const`-only -- it is a value semantic class representing any event. I will post one more alternative

vittorioromeo:
> is there a reason for us having a const qualified visit function instead of both a const and non const like normal `std::visit` given that we also implemented one for `getIf`?
> 
> also given that you can take a non const lvalue I would expect that I can modify the event inside but no this modifies a temporary

Opened #3396, my preferred solution now.

ChrisThrasher:
Closing in favor of #3396

PR #3396, opened 2025-01-20
Add non-`const` overload of `Event::visit`

vittorioromeo:
Alternative approach to #3375 and #3395 that allows handlers to take non-const references. In other words

```cpp
while (window.isOpen())
{
    window.clear();
    window.handleEvents([&](sf::Event::Closed&) { window.close(); });
    window.display();
}
```

will compile and will work as expected (the window will be closed when `[X]` is pressed). 

This approach gives `Event` a more sensible and homogeneous API. It was already possible to get a mutable reference to the event by using `getIf`, now you can also do the same with `visit`. Thanks to @ZXShady for the suggestion.

### DIFF
--- a/include/SFML/Window/Event.hpp
+++ b/include/SFML/Window/Event.hpp
@@ -347,6 +347,17 @@ public:
     ///
     ////////////////////////////////////////////////////////////
     template <typename T>
+    decltype(auto) visit(T&& visitor);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Apply a visitor to the event
+    ///
+    /// \param visitor The visitor to apply
+    ///
+    /// \return The result of applying the visitor to the event
+    ///
+    ////////////////////////////////////////////////////////////
+    template <typename T>
     decltype(auto) visit(T&& visitor) const;
 
 private:
--- a/include/SFML/Window/Event.inl
+++ b/include/SFML/Window/Event.inl
@@ -80,6 +80,14 @@ const TEventSubtype* Event::getIf() const
 
 ////////////////////////////////////////////////////////////
 template <typename T>
+decltype(auto) Event::visit(T&& visitor)
+{
+    return std::visit(std::forward<T>(visitor), m_data);
+}
+
+
+////////////////////////////////////////////////////////////
+template <typename T>
 decltype(auto) Event::visit(T&& visitor) const
 {
     return std::visit(std::forward<T>(visitor), m_data);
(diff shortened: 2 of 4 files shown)


ChrisThrasher:
I'm playing around with the unit tests on this PR and noticed that the `const` overload of `sf::Event::visit` will still use callbacks with non-const reference parameters. That doesn't seem to be what we want. If you have a `const` event then you shouldn't be able to mutate it like that. Is this the fault of `std::visit` or something we can address?

ChrisThrasher (approved):
> This approach gives `Event` a more sensible and homogeneous API. It was already possible to get a mutable reference to the event by using `getIf`, now you can also do the same with `visit`. Thanks to @ZXShady for the suggestion.

Good point. There is no reason to prohibit mutation via `visit` either. This is a good step forward that addresses the "my ref param callback isn't used!" problem in an elegant way. PR #3375 still has interesting ideas with regard to extending how `sf::Event::visit` behaves but let's tackle one problem at a time.

I would like @Exploiter's feedback before we merge.

ChrisThrasher:

ChrisThrasher:

ZXShady:
> I'm playing around with the unit tests on this PR and noticed that the `const` overload of `sf::Event::visit` will still use callbacks with non-const reference parameters. That doesn't seem to be what we want. If you have a `const` event then you shouldn't be able to mutate it like that. Is this the fault of `std::visit` or something we can address?
> 
> This is the visitor I'm using to test this. When I swap this in, the existing tests keep compiling and passing.
> 
> ```cpp
> struct
> {
>     // Const ref parameter
>     std::string_view operator()(const sf::Event::Closed&) const
>     {
>         return "Closed";
>     }
> 
>     // Non-const ref parameter
>     std::string_view operator()(sf::Event::Resized&) const
>     {
>         return "Resized";
>     }
> 
>     // Value parameter
>     std::string_view operator()(sf::Event::KeyPressed) const
>     {
>         return "KeyPressed";
>     }
> 
>     template <typename T>
>     std::string_view operator()(const T&) const
>     {
>         return "Other";
>     }
> } visitor;
> ```
> 
> Edit: When I stop using a temporary event and instead use a const stack variable, the mutable reference callback stops being used. Perhaps I'm just misunderstanding something?
> 
> Edit 2: You can disregard this comment. For whatever reason, a temporary `sf::Event` can use the non-const ref callbacks. I suppose this is not a problem. 

an rvalue object can call a non const membee function which in turns makes it an lvalue

```cpp
class Obj
{
// *this is always an lvalue inside the member functions
   void hello() { std::cout << "Hello" ;}
   void hello() const { std::cout << "Const Hello";}
}

Obj().hello(); // says Hello
```

well you can fix it by marking the non const `hello` member function to be an lvalue member function

`void hello() & {...}`

PR #3398, opened 2025-01-24
Use `true` and `false` literals instead of API specific macros

ZXShady:
Replace api specific macros with C++ literals.

Previously, we used API specific macros and way too many different ways to compare with API booleans we used 
```cpp
Var != API_TRUE
Var != API_FALSE
Var == API_FALSE
Var == API_TRUE
```

now it is uniform with the rest of our modern C++ codebase and I added `static_assert`s just incase `API_TRUE` gets defined into `42` at some point.

### DIFF
--- a/src/SFML/Graphics/GLCheck.cpp
+++ b/src/SFML/Graphics/GLCheck.cpp
@@ -33,6 +33,10 @@
 #include <filesystem>
 #include <ostream>
 
+// NOLINTBEGIN(readability-simplify-boolean-expr)
+static_assert(GL_TRUE == true);
+static_assert(GL_FALSE == false);
+// NOLINTEND(readability-simplify-boolean-expr)
 
 namespace sf::priv
 {
--- a/src/SFML/Graphics/RenderTarget.cpp
+++ b/src/SFML/Graphics/RenderTarget.cpp
@@ -611,7 +611,7 @@ void RenderTarget::resetGLStates()
         glCheck(glEnableClientState(GL_VERTEX_ARRAY));
         glCheck(glEnableClientState(GL_COLOR_ARRAY));
         glCheck(glEnableClientState(GL_TEXTURE_COORD_ARRAY));
-        glCheck(glColorMask(GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE));
+        glCheck(glColorMask(true, true, true, true));
         m_cache.scissorEnabled = false;
         m_cache.stencilEnabled = false;
         m_cache.glStatesSet    = true;
@@ -760,7 +760,7 @@ void RenderTarget::applyStencilMode(const StencilMode& mode)
         if (!m_cache.enable || m_cache.stencilEnabled)
         {
             glCheck(glDisable(GL_STENCIL_TEST));
-            glCheck(glColorMask(GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE));
+            glCheck(glColorMask(true, true, true, true));
 
             m_cache.stencilEnabled = false;
         }
@@ -860,7 +860,7 @@ void RenderTarget::setupDraw(bool useVertexCache, const RenderStates& states)
 
     // Mask the color buffer off if necessary
     if (states.stencilMode.stencilOnly)
-        glCheck(glColorMask(GL_FALSE, GL_FALSE, GL_FALSE, GL_FALSE));
+        glCheck(glColorMask(false, false, false, false));
 
     // Apply the texture
     if (!m_cache.enable || (states.texture && states.texture->m_fboAttachment))
@@ -913,7 +913,7 @@ void RenderTarget::cleanupDraw(const RenderStates& states)
 
     // Mask the color buffer back on if necessary
     if (states.stencilMode.stencilOnly)
-        glCheck(glColorMask(GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE));
+        glCheck(glColorMask(true, true, true, true));
 
     // Re-enable the cache at the end of the draw if it was disabled
     m_cache.enable = true;
--- a/src/SFML/Graphics/RenderTextureImplFBO.cpp
+++ b/src/SFML/Graphics/RenderTextureImplFBO.cpp
@@ -614,10 +614,10 @@ void RenderTextureImplFBO::updateTexture(unsigned int)
             {
                 // Scissor testing affects framebuffer blits as well
                 // Since we don't want scissor testing to interfere with our copying, we temporarily disable it for the blit if it is enabled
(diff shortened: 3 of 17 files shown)


MarioLiebisch:
Feeling mixed on this. I totally get the idea, but at least in case of OpenGL I'd keep `GL_TRUE` and `GL_FALSE` as a convention kind of thing. For stuff like X11's `True` and `False` this makes sense, though.

ZXShady:
> Feeling mixed on this. I totally get the idea, but at least in case of OpenGL I'd keep `GL_TRUE` and `GL_FALSE` as a convention kind of thing. 

I understand, but the  only and sole reason for these 2 macros existing is because of the fact that the opengl API is not C99 or higher meaning no `<stdbool.h>` so no `true` or `false`.if opengl API was not C89 or whatever it is then these wouldn't exist so given we have `true` and `false` I ought to replace them. It also doesn't give us a concrete way to compare for truthiness.

> For stuff like X11's `True` and `False` this makes sense, though.

I’m curious why wouldn’t X11’s `True` and `False` be considered part of the same convention? Both are API-specific macros defined for similar reasons (lack of `<stdbool.h>`), and both can be replaced with C++ `true` and `false` without losing clarity or functionality. If not using X11 macros makes sense, I think it’s consistent to extend the same principle to OpenGL.

MarioLiebisch:
You're absolutely right and it feels hard to explain and I absolutely don't want to to rule out this change. Maybe I'm just too used to writing OpenGL. It certainly makes sense to simplify/reduce all the comparisons.

I would have said it just feels more consistent with other macroed parameters or values, but there are many examples in SFML where the opposite would be true, too, e.g.:

```cpp
// Here GL_TRUE feels "better" to me,
static constexpr std::array<int, 10> intAttributes =
    {WGL_DRAW_TO_WINDOW_ARB,
    GL_TRUE,
    WGL_SUPPORT_OPENGL_ARB,
    GL_TRUE,
    WGL_DOUBLE_BUFFER_ARB,
    GL_TRUE,
    WGL_PIXEL_TYPE_ARB,
    WGL_TYPE_RGBA_ARB,
    0,
    0};

// regular booleans not so much
static constexpr std::array<int, 10> intAttributes =
    {WGL_DRAW_TO_WINDOW_ARB,
    true,
    WGL_SUPPORT_OPENGL_ARB,
    true,
    WGL_DOUBLE_BUFFER_ARB,
    true,
    WGL_PIXEL_TYPE_ARB,
    WGL_TYPE_RGBA_ARB,
    0,
    0};

// Here it's the opposuite
glCheck(GLEXT_glUniformMatrix3fv(binder.location, 1, GL_FALSE, matrix.array.data()));
// false fits in just as fine (if not better)
glCheck(GLEXT_glUniformMatrix3fv(binder.location, 1, false, matrix.array.data()));
```

As for the others, the change/difference feels just way more subtle.

But different question, should this then continue using `GLboolean`?

ZXShady:
> ```cpp
> // Here GL_TRUE feels "better" to me,
> static constexpr std::array<int, 10> intAttributes =
>     {WGL_DRAW_TO_WINDOW_ARB,
>     GL_TRUE,
>     WGL_SUPPORT_OPENGL_ARB,
>     GL_TRUE,
>     WGL_DOUBLE_BUFFER_ARB,
>     GL_TRUE,
>     WGL_PIXEL_TYPE_ARB,
>     WGL_TYPE_RGBA_ARB,
>     0,
>     0};
> 
> // regular booleans not so much
> static constexpr std::array<int, 10> intAttributes =
>     {WGL_DRAW_TO_WINDOW_ARB,
>     true,
>     WGL_SUPPORT_OPENGL_ARB,
>     true,
>     WGL_DOUBLE_BUFFER_ARB,
>     true,
>     WGL_PIXEL_TYPE_ARB,
>     WGL_TYPE_RGBA_ARB,
>     0,
>     0};


I have to agree that GL macro is better than the `true` here as the array type is `int` and it just feels "out of place" if you know what I mean (btw the array size and type was implicitly deduced and I don't think that idea is good). 


> // Here it's the opposuite
> glCheck(GLEXT_glUniformMatrix3fv(binder.location, 1, GL_FALSE, matrix.array.data()));
> // false fits in just as fine (if not better)
> glCheck(GLEXT_glUniformMatrix3fv(binder.location, 1, false, matrix.array.data()));
> ```



> As for the others, the change/difference feels just way more subtle.

I agree it's sublet and this pr is not that important. I may remove the replacement of the constants if unwated and just keep the simplification of if checks.



> But different question, should this then continue using `GLboolean`?

Sadly no, I wish I could but `GLboolean` if I rember correctly is just a typedef to `int` and OpenGL's `get` functions take the return variable as an out pointer parameter and a `bool*` cannot convert to an `int*`.

MarioLiebisch:
Please definitely wait for other opinions first.

> I may remove the replacement of the constants if unwated and just keep the simplification of if checks.

I'd have to try, but I could even imagine super pedantic settings complaining about using integer values as boolean values directly. (I think C# does so by default.)

ZXShady:
> I'd have to try, but I could even imagine super pedantic settings complaining about using integer values as boolean values directly. (I think C# does so by default.)

Tested with clang,gcc and msvc with /Wall and no warnings but clang tidy has `readability-implicit-bool-conversion` which we disable already

ChrisThrasher:
> I could even imagine super pedantic settings complaining about using integer values as boolean values directly.

All of those pedantic settings have been enabled for years so if CI doesn't complain then we're good.

binary1248:
A pedant could make the argument that substituting the definitions with their values is technically incorrect because we start to rely on details that were never really exposed through the documented API and thus can be subject to future change.

One has to consider not only our own code but everything else in the ecosystem. Any documentation/guides/tutorials/etc. written for these APIs will always use the definitions because they can also be used in C code where there aren't any better alternatives. Searching online for anything other than what is specified by the documentation will yield less precise results than if the definitions are used. Debugging tools also assume that the definitions are used (see any OpenGL debugger).

Replacing the definitions with their values also doesn't really increase type safety as @ChrisThrasher already mentioned. If it did, our CI would have complained a long time ago.

Also, for newcomers who are willing to contribute changes, using the documented definitions will be less surprising and thus easier to read/comprehend than having to translate the values back into the definitions in order to understand what is happening.

There are wrappers that encapsulate C APIs into C++ APIs with improved type safety. Compare https://github.com/KhronosGroup/Vulkan-Hpp to the standard Vulkan headers. Since OpenGL and Xlib are slowly dying I doubt anybody is going to put in the effort to develop/maintain C++ wrappers for those.

ZXShady:
Well, it seems not that popular and at the end nothing more than a style change.

PR #3399, opened 2025-01-25
Improved `Event::visit` and `WindowBase::handleEvents`

kimci86:
## Description

PR #3375 made valid code not compilable.

For example, the following code failed to compile because the implementation needed to make a copy of the visitor to initialize the `sf::priv::OverloadSet` instance.

```c++
const auto visitor = [ptr = std::make_unique<int>()](const auto&) {};
sf::Event event;
event.visit(visitor);
```

So I suggested to revert it -> This has been merged already in #3400.

What remains in this PR is about:
- Implementing sanity checks on handlers #3375 was about, but in a different way: `isEventHandler` in `Event` accessible in `WindowBase::handleEvents` via `friend`.
- Adding support for move-only event handlers in `WindowBase::handleEvents`. Thinking about the regression with `Event::visit` made me realize we can also support move-only handlers in `WindowBase::handleEvents`.
- Adding support for free functions in `WindowBase::handleEvents`. This was documented but never worked as far as I understand.
- Some renaming for clarity.

## Tasks

-   [x] Tested on Linux
-   [ ] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

See unit tests

### DIFF
--- a/include/SFML/Window/Event.hpp
+++ b/include/SFML/Window/Event.hpp
@@ -346,8 +346,8 @@ public:
     /// \return The result of applying the visitor to the event
     ///
     ////////////////////////////////////////////////////////////
-    template <typename T>
-    decltype(auto) visit(T&& visitor);
+    template <typename Visitor>
+    decltype(auto) visit(Visitor&& visitor);
 
     ////////////////////////////////////////////////////////////
     /// \brief Apply a visitor to the event
@@ -357,8 +357,8 @@ public:
     /// \return The result of applying the visitor to the event
     ///
     ////////////////////////////////////////////////////////////
-    template <typename T>
-    decltype(auto) visit(T&& visitor) const;
+    template <typename Visitor>
+    decltype(auto) visit(Visitor&& visitor) const;
 
 private:
     ////////////////////////////////////////////////////////////
--- a/include/SFML/Window/Event.hpp
+++ b/include/SFML/Window/Event.hpp
@@ -34,6 +34,7 @@
 
 #include <SFML/System/Vector2.hpp>
 
+#include <type_traits>
 #include <variant>
 
 
@@ -395,11 +396,22 @@ private:
     template <typename T, typename... Ts>
     [[nodiscard]] static constexpr bool isInParameterPack(const std::variant<Ts...>*)
     {
-        return (std::is_same_v<T, Ts> || ...);
+        return std::disjunction_v<std::is_same<T, Ts>...>;
     }
 
     template <typename T>
     static constexpr bool isEventSubtype = isInParameterPack<T>(decltype (&m_data)(nullptr));
+
+    friend class WindowBase;
+
+    template <typename Handler, typename... Ts>
+    [[nodiscard]] static constexpr bool isInvocableWithEventSubtype(const std::variant<Ts...>*)
+    {
+        return std::disjunction_v<std::is_invocable<Handler&, Ts&>...>;
+    }
+
+    template <typename Handler>
+    static constexpr bool isEventHandler = isInvocableWithEventSubtype<Handler>(decltype (&m_data)(nullptr));
 };
 
 } // namespace sf
--- a/include/SFML/Window/WindowBase.inl
+++ b/include/SFML/Window/WindowBase.inl
@@ -28,6 +28,7 @@
 #include <SFML/Window/Event.hpp>
 #include <SFML/Window/WindowBase.hpp> // NOLINT(misc-header-include-cycle)
 
+#include <type_traits>
 #include <utility>
 
 
@@ -53,6 +54,40 @@ struct DelayOverloadResolution
     {
     }
 };
+
+// Wrapper providing a callable type suitable as OverloadSet base type for different kinds of handlers.
+// By default, we derive from the handler type and inherit its call operators.
+template <typename Handler>
+struct Caller : Handler
+{
+    using Handler::operator();
+};
+
+// Inheritance is not possible with reference types.
+// In this case, we capture the reference and forward arguments to it.
+template <typename Handler>
+struct Caller<Handler&>
+{
+    // Use SFINAE so that the call operator exists only for arguments the captured handler accepts
+    template <typename Argument, std::enable_if_t<std::is_invocable_v<Handler&, Argument>, int> = 0>
+    decltype(auto) operator()(Argument&& argument)
+    {
+        return handler(std::forward<Argument>(argument));
+    }
+    Handler& handler;
+};
+
+// Inheritance is not possible with function pointers either.
+// In this case, we capture the function pointer to call it.
+template <typename Return, typename Argument>
+struct Caller<Return (*)(Argument)>
+{
+    Return operator()(Argument&& argument)
+    {
+        return function(std::forward<Argument>(argument));
+    }
+    Return (*function)(Argument);
+};
 } // namespace priv
 
 
@@ -67,7 +102,7 @@ void WindowBase::handleEvents(Handlers&&... handlers)
     // complains about it even though the code would become incorrect
 
     // NOLINTNEXTLINE(misc-const-correctness)
-    priv::OverloadSet overloadSet{std::forward<Handlers>(handlers)...,
+    priv::OverloadSet overloadSet{priv::Caller<Handlers>{std::forward<Handlers>(handlers)}...,
                                   [](const priv::DelayOverloadResolution&) { /* ignore */ }};
 
     while (std::optional event = pollEvent())
--- a/include/SFML/Window/WindowBase.inl
+++ b/include/SFML/Window/WindowBase.inl
@@ -61,6 +61,7 @@ template <typename... Handlers>
 void WindowBase::handleEvents(Handlers&&... handlers)
 {
     static_assert(sizeof...(Handlers) > 0, "Must provide at least one handler");
+    static_assert((Event::isEventHandler<Handlers> && ...), "Handlers must accept at least one subtype of sf::Event");
 
     // Disable misc-const-correctness for this line since clang-tidy
     // complains about it even though the code would become incorrect
--- a/include/SFML/Window/WindowBase.inl
+++ b/include/SFML/Window/WindowBase.inl
@@ -57,16 +57,17 @@ struct DelayOverloadResolution
 
 
 ////////////////////////////////////////////////////////////
-template <typename... Ts>
-void WindowBase::handleEvents(Ts&&... handlers) // NOLINT(cppcoreguidelines-missing-std-forward)
+template <typename... Handlers>
+void WindowBase::handleEvents(Handlers&&... handlers)
 {
-    static_assert(sizeof...(Ts) > 0, "Must provide at least one handler");
+    static_assert(sizeof...(Handlers) > 0, "Must provide at least one handler");
 
     // Disable misc-const-correctness for this line since clang-tidy
-    // complains about it even though the code would become uncompilable
+    // complains about it even though the code would become incorrect
 
     // NOLINTNEXTLINE(misc-const-correctness)
-    priv::OverloadSet overloadSet{std::forward<Ts>(handlers)..., [](const priv::DelayOverloadResolution&) { /* ignore */ }};
+    priv::OverloadSet overloadSet{std::forward<Handlers>(handlers)...,
+                                  [](const priv::DelayOverloadResolution&) { /* ignore */ }};
 
     while (std::optional event = pollEvent())
         event->visit(overloadSet);
--- a/test/Window/WindowBase.test.cpp
+++ b/test/Window/WindowBase.test.cpp
@@ -12,6 +12,7 @@
 #include <chrono>
 #include <memory>
 #include <type_traits>
+#include <utility>
 
 TEST_CASE("[Window] sf::WindowBase", runDisplayTests())
 {
@@ -240,5 +241,13 @@ TEST_CASE("[Window] sf::WindowBase", runDisplayTests())
             void operator()(const sf::Event::Closed&) && = delete;
         };
         windowBase.handleEvents(LvalueOnlyHandler{});
+
+        // Should compile if user provides a reference to a handler
+        auto handler = [](const sf::Event::Closed&) {};
+        windowBase.handleEvents(handler);
+        windowBase.handleEvents(std::as_const(handler));
+
+        // Should compile if user provides a function pointer
+        windowBase.handleEvents(+[](const sf::Event::Closed&) {});
     };
 }
--- a/test/Window/WindowBase.test.cpp
+++ b/test/Window/WindowBase.test.cpp
@@ -10,6 +10,7 @@
 
 #include <WindowUtil.hpp>
 #include <chrono>
+#include <memory>
 #include <type_traits>
 
 TEST_CASE("[Window] sf::WindowBase", runDisplayTests())
@@ -219,5 +220,25 @@ TEST_CASE("[Window] sf::WindowBase", runDisplayTests())
 
         // Should compile if user provides both a specific handler and a catch-all
         windowBase.handleEvents([](const sf::Event::Closed&) {}, [](const auto&) {});
+
+        // Should compile if user provides a handler taking an event subtype by value or reference,
+        // but not rvalue reference because it would never be called.
+        windowBase.handleEvents([](sf::Event::Closed) {});
+        windowBase.handleEvents([](const sf::Event::Closed) {});
+        windowBase.handleEvents([](sf::Event::Closed&) {});
+        windowBase.handleEvents([](const sf::Event::Closed&) {});
+
+        // Should compile if user provides a move-only handler
+        windowBase.handleEvents([p = std::make_unique<int>()](const sf::Event::Closed&) {});
+
+        // Should compile if user provides a handler with deleted rvalue ref-qualified call operator
+        struct LvalueOnlyHandler
+        {
+            void operator()(const sf::Event::Closed&) &
+            {
+            }
+            void operator()(const sf::Event::Closed&) && = delete;
+        };
+        windowBase.handleEvents(LvalueOnlyHandler{});
     };
 }
--- a/include/SFML/Window/Event.inl
+++ b/include/SFML/Window/Event.inl
@@ -79,18 +79,18 @@ const TEventSubtype* Event::getIf() const
 
 
 ////////////////////////////////////////////////////////////
-template <typename T>
-decltype(auto) Event::visit(T&& visitor)
+template <typename Visitor>
+decltype(auto) Event::visit(Visitor&& visitor)
 {
-    return std::visit(std::forward<T>(visitor), m_data);
+    return std::visit(std::forward<Visitor>(visitor), m_data);
 }
 
 
 ////////////////////////////////////////////////////////////
-template <typename T>
-decltype(auto) Event::visit(T&& visitor) const
+template <typename Visitor>
+decltype(auto) Event::visit(Visitor&& visitor) const
 {
-    return std::visit(std::forward<T>(visitor), m_data);
+    return std::visit(std::forward<Visitor>(visitor), m_data);
 }
 
 } // namespace sf
--- a/include/SFML/Window/WindowBase.hpp
+++ b/include/SFML/Window/WindowBase.hpp
@@ -326,8 +326,8 @@ public:
     /// \see `waitEvent`, `pollEvent`
     ///
     ////////////////////////////////////////////////////////////
-    template <typename... Ts>
-    void handleEvents(Ts&&... handlers);
+    template <typename... Handlers>
+    void handleEvents(Handlers&&... handlers);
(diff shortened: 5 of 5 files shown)


include/SFML/Window/Event.hpp
```diff
@@ -340,26 +340,26 @@ class Event
...
+    /// \param visitor The visitor to apply
     ///
-    /// \return The result of applying the handlers to the event
+    /// \return The result of applying the visitor to the event
     ///
     ////////////////////////////////////////////////////////////
-    template <typename... Handlers>
-    decltype(auto) visit(Handlers&&... handlers);
+    template <typename T>
+    decltype(auto) visit(T&& visitor);
```
ZXShady, line 352: I would like the `T` template type parameters to be named `Visitor`

include/SFML/Window/Event.hpp
```diff
@@ -340,26 +340,26 @@ class Event
...
+    /// \param visitor The visitor to apply
     ///
-    /// \return The result of applying the handlers to the event
+    /// \return The result of applying the visitor to the event
     ///
     ////////////////////////////////////////////////////////////
-    template <typename... Handlers>
-    decltype(auto) visit(Handlers&&... handlers) const;
+    template <typename T>
+    decltype(auto) visit(T&& visitor) const;
```
ZXShady, line 362: Same here

kimci86:
Diff is simpler to review with master parent commit: https://github.com/SFML/SFML/compare/SFML:master^1..kimci86:bugfix/event_visit_and_handleEvents

ZXShady (commented):
Name change

include/SFML/Window/WindowBase.inl
```diff
@@ -35,23 +35,83 @@ namespace sf
...
+template <typename T>
+struct Caller<const T&>
+{
+    template <typename Arg, std::enable_if_t<std::is_invocable_v<T, Arg>>>
+    decltype(auto) operator()(Arg&& arg)
+    {
+        return handler(std::forward<Arg>(arg));
+    }
+    const T& handler;
+};
```
ZXShady, line 83: This specilization is not necessary, `T&` specilization will match `const T&` as well

kimci86: Right!

include/SFML/Window/WindowBase.inl
ZXShady, line 1: Missing include for `<type_traits>` it uses `std::enable_if_t`

ZXShady (changes requested):

include/SFML/Window/WindowBase.inl
```diff
@@ -35,23 +35,83 @@ namespace sf
...
+template <typename T>
+struct Caller : T
+{
+    using T::operator();
+};
+
+template <typename T>
+struct Caller<T&>
+{
+    template <typename Arg, std::enable_if_t<std::is_invocable_v<T, Arg>>>
```
ZXShady, line 66:
```suggestion
    template <typename Arg, std::enable_if_t<std::is_invocable_v<T, Arg>,int> = 0>
```

SFINAE requires the `= 0` to force the compiler to check and sfinae this overload out

ZXShady (changes requested):

include/SFML/Window/Event.hpp
kimci86: Done

include/SFML/Window/Event.hpp
kimci86: Done

include/SFML/Window/WindowBase.inl
kimci86: Added

include/SFML/Window/WindowBase.inl
kimci86: I didn't know, I trust you about that. Done.

include/SFML/Window/Event.hpp
```diff
@@ -400,16 +400,18 @@ class Event
...
 
     template <typename T>
-    static constexpr bool isEventSubtype = isInParameterPack<T>(static_cast<decltype(&m_data)>(nullptr));
+    static constexpr bool isEventSubtype = isInParameterPack<T>(decltype (&m_data)(nullptr));
+
+    friend class WindowBase;
 
     template <typename Handler, typename... Ts>
-    static constexpr bool isInvokableWithAnyOf(std::variant<Ts...>*)
+    [[nodiscard]] static constexpr bool isInvokableWithEventSubType(const std::variant<Ts...>*)
```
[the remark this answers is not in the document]
ZXShady: is the const variant pointer necessary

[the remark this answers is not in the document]
kimci86: No, it is not. I copy-pasted from `isInParameterPack`. :shrug:

include/SFML/Window/WindowBase.inl
```diff
@@ -28,30 +28,79 @@
...
+    {
+        return handler(std::forward<Argument>(argument));
+    }
+    Handler& handler;
+};
+
+template <typename Return, typename Argument>
+struct Caller<Return (*)(Argument)>
+{
+    decltype(auto) operator()(Argument&& argument)
```
ZXShady, line 78: should this use `Return` instead of `decltype(auto)` since we already have the return type

kimci86: I guess `Return` would be enough.

kimci86: Done.

kimci86:
Rebased and updated PR description since a part of it was merged in #3400.

include/SFML/Window/Event.hpp
```diff
@@ -395,11 +396,22 @@ class Event
...
 
     template <typename T>
     static constexpr bool isEventSubtype = isInParameterPack<T>(decltype (&m_data)(nullptr));
+
+    friend class WindowBase;
+
+    template <typename Handler, typename... Ts>
+    [[nodiscard]] static constexpr bool isInvocableWithEventSubtype(const std::variant<Ts...>*)
+    {
+        return std::disjunction_v<std::bool_constant<std::is_invocable_v<Handler, Ts&> || std::is_invocable_v<Handler, Ts&&>>...>;
```
ChrisThrasher, line 410: What does this implementation offer compared to this?
```cpp
return std::disjunction_v<std::is_invocable_v<Handler, Ts&>...>;
```

ChrisThrasher (commented):
Does `std::visit` support references to lambdas and function pointers? Are we adding this infrastructure to make sure that we don't accidentally remove any of the functionality of `std::visit`? Or are expanding upon what `std::visit` natively supports?

include/SFML/Window/Event.hpp
ZXShady: It checks whether it can be called with an rvalue which I don't think is correct.

as the event is an lvalue in both cases the lamdba should take an lvalue or a copy

ZXShady:
> Does `std::visit` support references to lambdas and function pointers? Are we adding this infrastructure to make sure that we don't accidentally remove any of the functionality of `std::visit`? Or are expanding upon what `std::visit` natively supports?

std::visit takes the entire overload set not individual handlers. 

From what I understand we are expanding what types the `WindowBase::handleEvents` can take

include/SFML/Window/Event.hpp
kimci86: @ChrisThrasher The difference is that this is accepted:
```c++
windowBase.handleEvents([](sf::Event::Closed&&) {});
```
but as @ZXShady explains, such a handler is never actually called. It is a mistake.
So I will use what Chris suggests so that `[](sf::Event::Closed&&) {}` is rightfully not allowed.

test/Window/WindowBase.test.cpp
```diff
@@ -219,5 +220,24 @@ TEST_CASE("[Window] sf::WindowBase", runDisplayTests())
...
+        windowBase.handleEvents([](sf::Event::Closed) {});
+        windowBase.handleEvents([](const sf::Event::Closed) {});
+
+        // Should compile if user provides a mutable reference to a move-only handler
+        auto handler = [p = std::make_unique<int>()](const sf::Event::Closed&) {};
+        windowBase.handleEvents(handler);
+
+        // Should compile if user provides a const reference to a move-only handler
+        const auto constHandler = [p = std::make_unique<int>()](const sf::Event::Closed&) {};
+        windowBase.handleEvents(constHandler);
```
ZXShady, line 238: can't this use `std::as_const`?

kimci86: Done

include/SFML/Window/WindowBase.inl
```diff
@@ -53,20 +54,49 @@ struct DelayOverloadResolution
...
+
+template <typename Return, typename Argument>
+struct Caller<Return (*)(Argument)>
+{
+    Return operator()(Argument&& argument)
+    {
+        return function(std::forward<Argument>(argument));
+    }
+    Return (*function)(Argument);
+};
```
ChrisThrasher, line 83: I want to make sure I understand what's going on here so let me know if the following summary is correct:

The `Caller<Handler&>` class template specialization exists to create a closure for a reference to a lambda. This enables the use of move-only lambdas because it means we can simply hold a reference to said lambdas instead of actually moving them. 

The `struct Caller<Return (*)(Argument)>` class template specialization is the same idea but for function pointers.

And then there's the non-specialized class template which catches all other callable which do not need special handling.

include/SFML/Window/WindowBase.inl
```diff
@@ -53,20 +54,49 @@ struct DelayOverloadResolution
...
+    {
+        return handler(std::forward<Argument>(argument));
+    }
+    Handler& handler;
+};
+
+template <typename Return, typename Argument>
+struct Caller<Return (*)(Argument)>
+{
+    Return operator()(Argument&& argument)
```
ChrisThrasher, line 78: Should this specialization also use the same SFINAE type checking that `Caller<Handler&>` uses?

ZXShady: Not Kimi, but no the sfinae was necessary for the other specilization because we didn't have the type of the arguement (you cant get the type of the arguement of a lamdba) and we need to SFINAE out uncallable overloads if we cannot call the lamdba with the type so `std::visit` warns about invalid handler

include/SFML/Window/WindowBase.inl
```diff
@@ -53,20 +54,49 @@ struct DelayOverloadResolution
...
+};
+
+template <typename Handler>
+struct Caller<Handler&>
+{
+    template <typename Argument, std::enable_if_t<std::is_invocable_v<Handler, Argument>, int> = 0>
+    decltype(auto) operator()(Argument&& argument)
+    {
+        return handler(std::forward<Argument>(argument));
+    }
```
ZXShady, line 71: the explicit SFINAE could be rewritten to

```cpp
template<typename Arg>
auto operator()(Arg&& arg) -> decltype(handler(std::forward<Arg>(arg)))
{
    return handler(std::forward<Arg>(arg));
}
```


I don't like it (and not suggesting it) but thought to say it

include/SFML/Window/WindowBase.inl
ChrisThrasher:
```
/Users/thrasher/Projects/sfml/include/SFML/Window/WindowBase.inl:92:20: error: static assertion failed due to requirement 'Event::isEventHandler<void (*)(const std::string &)>': Handlers must accept at least one subtype of sf::Event
   92 |     static_assert((Event::isEventHandler<Handlers> && ...), "Handlers must accept at least one subtype of sf::Event");
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/thrasher/Projects/sfml/test/Window/WindowBase.test.cpp:239:20: note: in instantiation of function template specialization 'sf::WindowBase::handleEvents<void (*)(const std::string &)>' requested here
  239 |         windowBase.handleEvents(+[](const std::string&) {});
  ```
  I tried using a function pointer with invalid arguments and it fired that `static_assert` so all is well 👍🏻
  
  So then why do we need to SFINAE out uncallable overloads here when `handleEvents` already does that?

Comment (ZXShady) [reply include/SFML/Window/WindowBase.inl:78]: > So then why do we need to SFINAE out uncallable overloads here when handleEvents already does that?

Rethinked about it and it is because otherwise the handler will catch everything and cause ambiguities so we need to SFINAE out the overload.

try this and remove the SFINAE and you will get ambiguous errors

```cpp
auto lam = [](sf::Event::WindowClosed){};
auto lam2 = [](sf::Event::MousePressed){};
window.handleEvents(lam,lam2);
```

include/SFML/Window/WindowBase.inl
```diff
@@ -53,20 +54,49 @@ struct DelayOverloadResolution
...
+template <typename Handler>
+struct Caller : Handler
+{
+    using Handler::operator();
+};
+
+template <typename Handler>
+struct Caller<Handler&>
+{
+    template <typename Argument, std::enable_if_t<std::is_invocable_v<Handler, Argument>, int> = 0>
```
ZXShady, line 67:
```suggestion
    template <typename Argument, std::enable_if_t<std::is_invocable_v<Handler&, Argument>, int> = 0>
```


lets imagine this

```cpp
struct Handler {
    void operator()(event-type) & {}
    void operator()(event-type) && = delete;
};

Handler handler;
window.handleEvent(handler); // errors out although it shouldn't
```

(didn't test it I am not at laptop)

ZXShady (changes requested):

include/SFML/Window/WindowBase.inl
kimci86: You understand correctly.
You are right, it deserves a few comments. I will add some.

include/SFML/Window/WindowBase.inl
kimci86: Nice find, you are right.

include/SFML/Window/WindowBase.inl
kimci86: Done

include/SFML/Window/WindowBase.inl
kimci86: Similar change was also needed in `isInvocableWithEventSubtype`. Done.

include/SFML/Window/Event.hpp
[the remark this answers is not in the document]
kimci86: Renaming is done. I am assuming keeping `const` is ok.

test/Window/WindowBase.test.cpp
```diff
@@ -219,5 +221,30 @@ TEST_CASE("[Window] sf::WindowBase", runDisplayTests())
...
+
+        // Should compile if user provides a handler with deleted rvalue ref-qualified call operator
+        struct LvalueOnlyHandler
+        {
+            void operator()(const sf::Event::Closed&) &
+            {
+            }
+            void operator()(const sf::Event::Closed&) && = delete;
+        };
+        windowBase.handleEvents(LvalueOnlyHandler{});
```
ZXShady, line 240: This shouldn't compile though you are passing an rvalue to it.

You should test when the handlerItself is an Lvalue


```suggestion
LvalueOnlyHandler handler; windowBase.handleEvents(handler);
```

ZXShady (changes requested):

test/Window/WindowBase.test.cpp
kimci86: The thing is this rvalue is used to initialize the OverloadSet that is then used as an lvalue to visit events.

ZXShady: > The thing is this rvalue is used to initialize the OverloadSet that is then used as an lvalue to visit events.

so OverloadSet doesn't preserve reference category is that wanted?

kimci86: I think so, because it is used potentially many times in the event loop.

test/Window/WindowBase.test.cpp
```diff
@@ -219,5 +221,30 @@ TEST_CASE("[Window] sf::WindowBase", runDisplayTests())
...
+
+        // Should compile if user provides a reference to a move-only handler
+        auto handler = [p = std::make_unique<int>()](const sf::Event::Closed&) {};
+        windowBase.handleEvents(handler);
+        windowBase.handleEvents(std::as_const(handler));
```
kimci86, line 245: This is test and corresponding implementation is actually not about move-only, but just about passing by reference. I will simplify.

kimci86: Done.

include/SFML/Window/WindowBase.inl
```diff
@@ -53,20 +54,56 @@ struct DelayOverloadResolution
...
+void WindowBase::handleEvents(Handlers&&... handlers)
 {
-    static_assert(sizeof...(Ts) > 0, "Must provide at least one handler");
+    static_assert(sizeof...(Handlers) > 0, "Must provide at least one handler");
+    static_assert((Event::isEventHandler<Handlers> && ...), "Handlers must accept at least one subtype of sf::Event");
 
     // Disable misc-const-correctness for this line since clang-tidy
     // complains about it even though the code would become uncompilable
 
     // NOLINTNEXTLINE(misc-const-correctness)
```
ChrisThrasher, line 104: I searched https://github.com/SFML/SFML/pull/3139 for an explanation of this `NOLINTNEXTLINE` line but couldn't find an explanation. Do you think we need to keep this? I'm not sure why `overloadSet` would need to be mutable.

test/Window/WindowBase.test.cpp
```diff
@@ -219,5 +221,33 @@ TEST_CASE("[Window] sf::WindowBase", runDisplayTests())
...
+
+        // Should compile if user provides a handler taking an event subtype by value or reference,
+        // but not rvalue reference because it would never be called.
+        windowBase.handleEvents([](sf::Event::Closed) {});
+        windowBase.handleEvents([](const sf::Event::Closed) {});
+        windowBase.handleEvents([](sf::Event::Closed&) {});
+        windowBase.handleEvents([](const sf::Event::Closed&) {});
+
+        // Should compile if user provides a move-only handler
+        windowBase.handleEvents([p = std::make_unique<int>()](const sf::Event::Closed&) {});
```
ChrisThrasher, line 233: Should we store this move-only lambda on the stack and test for the state of the lambda after `handleEvents` is used? Should we expect that unique pointer to be null after it's passed to `handleEvents`?

ChrisThrasher (commented):
I have a few more small questions about this PR but I'm generally in favor and should be ready to approve soon!

include/SFML/Window/WindowBase.inl
kimci86: We don't want `overloadSet` to be const because that would make it impossible to use mutable handlers, for which the call operator is not const.
I though clang-tidy would suggest to make this object const but I don't reproduce locally, so I will remove the comment and check what CI finds.

ChrisThrasher: I removed that `NOLINT` and re-ran clang-tidy without errors so I think we're good.

test/Window/WindowBase.test.cpp
kimci86: This test is merely making sure that we are not missing an `std::forward` in `handleEvents` implementation. 
If we put this handler on the stack, then `handleEvents` would take it by reference. The `unique_ptr` would be left intact. This test is not run anyways. It is just checking the code compiles.

ChrisThrasher:

include/SFML/Window/WindowBase.inl
kimci86: CI is not happy:

```
/home/runner/work/SFML/SFML/include/SFML/Window/WindowBase.inl:101:5: error: variable 'overloadSet' of type [...] can be declared 'const' [misc-const-correctness,-warnings-as-errors]
```

https://github.com/SFML/SFML/actions/runs/12978770634/job/36193726247?pr=3399#step:8:13

ChrisThrasher: A workaround is to use `auto`. clang-tidy does not recommend adding `const` to any local variables declared with `auto`.

[Ask me how I know :)](https://github.com/llvm/llvm-project/issues/60789)

kimci86: I would rather add a NOLINT comment than rely on a clang-tidy bug.
I will reset my branch to the commit with the comment.

ChrisThrasher: The comment is still incorrect though. Adding `const` does not make this code uncompilable as the comment claims.

kimci86: I changed "uncompilable" into "incorrect".

ChrisThrasher (approved):

ChrisThrasher:

ZXShady (approved):

eXpl0it3r:
Might have also fixed a VS issue reported in #3454

PR #3400, opened 2025-01-25
Revert `Event::visit` regression and add test

kimci86:
## Description

This is the first part of #3399 that reverts #3375 and adds a corresponding non-regression test.

@ChrisThrasher suggested to me I submit this part separately to merge it first.
Then the discussion can continue in #3399.

## Tasks

-   [x] Tested on Linux
-   [ ] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

- [x] Unit test

### DIFF
--- a/include/SFML/Window/Event.hpp
+++ b/include/SFML/Window/Event.hpp
@@ -34,7 +34,6 @@
 
 #include <SFML/System/Vector2.hpp>
 
-#include <type_traits>
 #include <variant>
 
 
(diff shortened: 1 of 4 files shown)


ChrisThrasher (approved):
While a move-only lambda is perhaps not a common use case, it seems like a valid one nonetheless so I think it's wise to revert #3375 and revisit it with this requirement codified in the test suite.

ChrisThrasher:

PR #3403, opened 2025-01-28
Fix #647 `sf::Image` support for Unicode filenames save/load

### LINKED ISSUE #647: sf::Image::saveToFile() fails when user has special characters in his username
When a user has special characters in his username (e.g. "Sébastien", notice the "é"), and he chooses to save an image under a path which contains the username (e.g. C:\Users\%USERNAME%\Desktop under Windows), sf::Image::saveToFile() fails with the error: "Failed to save image FILENAME" (file ImageLoader.cpp, line 261, SFML 2.1).

It would be good if sf::Image::saveToFile() was overloaded in order to accept an std::wstring instead of an std::string.

FRex:
-   [x] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before?
-   [x] Have you provided some example/test code for your changes?

## Description

Here's a fix issue https://github.com/SFML/SFML/issues/647
All the tests have passed on https://github.com/FRex/SFML/actions/runs/13015976076

It has ~~3~~ commits (for explanation of each see below):
1. adding tests for saving, loading and failing to load a file with Unicode filename.
1. fix for `sf::Image` saving/loading to/from Unicode filenames.
~~1. fix for printing `std::filesystem::path` to `std::string` to then print error to `sf::err()`.~~

EDIT: The last commit from above was put in PR https://github.com/SFML/SFML/pull/3407 and merged, now this PR has jus `sf::Image` tests and fixes for Unicode load/save filenames.

---

The tests check for 4 different interesting Unicode filenames: one with a Spanish letter, one with a Polish letter, one with a CJK character, and one with an emoji.

The Spanish letter is a Latin 1 character, and often works on Windows due to codepage 850 aka Latin 1 (which has values that overlap with Unicode, by design) being set. The Polish character will fail with that codepage since it's above `0xff`. The CJK character will not work in any of these extended ASCII/codepage settings, it requires real Unicode support to work.

The emoji makes sure characters above `0xffff` (outside of Unicode's first plane, the Basic Multilingual Plane) are handled properly. These characters take 3 UTF-8 bytes/codeunits or 2 wchars/UTF-16 codeunits (using surrogate pairs). Some older software (especially made before emojis became so popular online) sometimes fails on them, if they assume that UTF-16 is fixed width, so this is a good test for robust Unicode handling.

---

The fix in `sf::Image` uses `std::ifstream`/`std::ofstream` which accept `std::filesystem::path` so the Unicode stuff never leaves `std::` parts, `stb_`/`stbi_` then use callbacks to read to/from these streams, just how SFML's own streams.

---

**EDIT: see above, this commit is no longer in this PR.**

~~In `Utils.cpp` there is a helper function for making error messages that puts `std::filesystem::path` into a `char` stream, which throws an exception on Windows if that path contains Unicode. The fix is to always convert to UTF-8. The `reinterpret_cast` is because in C++17 `u8path` is `char` string, but in C++20 it's `char8_t` string which cannot be put into a `char` stream. It's same data and `char8_t` is same size as `char` so it works.~~

## Tasks

-   [ ] Tested on Linux
-   [ ] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

Run the below program on Windows and see that files `🐌.png`, `日.png`, `ń.png` and `ñ.png` appear and load, as expected, and error message is printed properly in UTF-8 (if redirected to file) or at least doesn't throw an exception and just has some boxy characters in place of Unicode (if in default terminal).

Before this PR due to using `.string()` all these operations would throw a "No mapping for the unicode character exists in the target multi-byte code page" exception when path contains Unicode outside of current codepage (so most likely `ñ` works, but the other three don't, for most Western/American developers on default English language Windows).

```cpp
#include <SFML/Graphics/Image.hpp>
#include <iostream>

int main()
{
    std::cerr << std::boolalpha; // to print true/false to std::cerr

    const char32_t totry[] = {
        0xf1,    // small n with tilde, outside of ascii but inside common latin1 codepage
        0x144,   // small n with acute accent, outside of latin1 codepage
        0x65E5,  // CJK symbol for sun, outside of any european codepage
        0x1F40C, // snail emoji, outside of Basic Multilingual Plane
    };

    char32_t utf32name[6] = {'X', '.', 'p', 'n', 'g', '\0'};

    for(const auto c : totry)
    {
        utf32name[0] = c;
        utf32name[1] = '.';
        sf::Image img = sf::Image(sf::Vector2u(32, 32), sf::Color::Red);
        std::cerr << img.saveToFile(utf32name) << std::endl; // successful save
        std::cerr << img.loadFromFile(utf32name) << std::endl; // successful load
        utf32name[1] = ',';
        std::cerr << !img.loadFromFile(utf32name) << std::endl; // failed load
    }
}
```

### DIFF
--- a/src/SFML/Graphics/Image.cpp
+++ b/src/SFML/Graphics/Image.cpp
@@ -42,6 +42,7 @@
 #include <stb_image_write.h>
 
 #include <algorithm>
+#include <fstream>
 #include <iomanip>
 #include <memory>
 #include <ostream>
@@ -212,11 +213,41 @@ bool Image::loadFromFile(const std::filesystem::path& filename)
     // Clear the array (just in case)
     m_pixels.clear();
 
+    // Set up the stb_image callbacks for the std::ifstream
+    const auto readStdIfStream = [](void* user, char* data, int size)
+    {
+        auto& file = *static_cast<std::ifstream*>(user);
+        file.read(data, size);
+        return static_cast<int>(file.gcount());
+    };
+    const auto skipStdIfStream = [](void* user, int size)
+    {
+        auto& file = *static_cast<std::ifstream*>(user);
+        if (!file.seekg(size, std::ios_base::cur))
+            err() << "Failed to seek image loader std::ifstream" << std::endl;
+    };
+    const auto eofStdIfStream = [](void* user)
+    {
+        auto& file = *static_cast<std::ifstream*>(user);
+        return static_cast<int>(file.eof());
+    };
+    const stbi_io_callbacks callbacks{readStdIfStream, skipStdIfStream, eofStdIfStream};
+
+    // Open file
+    std::ifstream file(filename, std::ios::binary);
+    if (!file.is_open())
+    {
+        // Error, failed to open the file
+        err() << "Failed to load image\n"
+              << formatDebugPathInfo(filename) << "\nReason: " << std::strerror(errno) << std::endl;
+        return false;
+    }
+
     // Load the image and get a pointer to the pixels in memory
     int width    = 0;
     int height   = 0;
     int channels = 0;
-    if (const auto ptr = StbPtr(stbi_load(filename.string().c_str(), &width, &height, &channels, STBI_rgb_alpha)))
+    if (const auto ptr = StbPtr(stbi_load_from_callbacks(&callbacks, &file, &width, &height, &channels, STBI_rgb_alpha)))
     {
         // Assign the image properties
         m_size = Vector2u(Vector2i(width, height));
@@ -324,28 +355,42 @@ bool Image::saveToFile(const std::filesystem::path& filename) const
         const std::filesystem::path extension     = filename.extension();
         const Vector2i              convertedSize = Vector2i(m_size);
 
+        // Callback to write to std::ofstream
+        auto writeStdOfstream = [](void* context, void* data, int size)
+        {
+            auto& file = *static_cast<std::ofstream*>(context);
+            if (file)
+                file.write(static_cast<const char*>(data), static_cast<std::streamsize>(size));
+        };
+
         if (extension == ".bmp")
         {
             // BMP format
-            if (stbi_write_bmp(filename.string().c_str(), convertedSize.x, convertedSize.y, 4, m_pixels.data()))
+            std::ofstream file(filename, std::ios::binary);
+            if (stbi_write_bmp_to_func(writeStdOfstream, &file, convertedSize.x, convertedSize.y, 4, m_pixels.data()) && file)
                 return true;
         }
         else if (extension == ".tga")
         {
             // TGA format
-            if (stbi_write_tga(filename.string().c_str(), convertedSize.x, convertedSize.y, 4, m_pixels.data()))
+            std::ofstream file(filename, std::ios::binary);
+            if (stbi_write_tga_to_func(writeStdOfstream, &file, convertedSize.x, convertedSize.y, 4, m_pixels.data()) && file)
                 return true;
         }
         else if (extension == ".png")
         {
             // PNG format
-            if (stbi_write_png(filename.string().c_str(), convertedSize.x, convertedSize.y, 4, m_pixels.data(), 0))
+            std::ofstream file(filename, std::ios::binary);
+            if (stbi_write_png_to_func(writeStdOfstream, &file, convertedSize.x, convertedSize.y, 4, m_pixels.data(), 0) &&
+                file)
                 return true;
         }
         else if (extension == ".jpg" || extension == ".jpeg")
         {
             // JPG format
-            if (stbi_write_jpg(filename.string().c_str(), convertedSize.x, convertedSize.y, 4, m_pixels.data(), 90))
+            std::ofstream file(filename, std::ios::binary);
+            if (stbi_write_jpg_to_func(writeStdOfstream, &file, convertedSize.x, convertedSize.y, 4, m_pixels.data(), 90) &&
+                file)
                 return true;
         }
         else
--- a/test/Graphics/Image.test.cpp
+++ b/test/Graphics/Image.test.cpp
@@ -265,6 +265,18 @@ TEST_CASE("[Graphics] sf::Image")
             CHECK(!image.loadFromFile("."));
             CHECK(!image.loadFromFile("this/does/not/exist.jpg"));
 
+            // small n with tilde, from Spanish, outside of ASCII, inside common Latin 1 codepage
+            CHECK(!image.loadFromFile(std::filesystem::path(U"missing-file-\u00f1.png")));
+
+            // small n with acute accent, from Polish, outside of Latin 1 codepage
+            CHECK(!image.loadFromFile(std::filesystem::path(U"missing-file-\u0144.png")));
+
+            // CJK symbol for Sun, outside of any European language codepage
+            CHECK(!image.loadFromFile(std::filesystem::path(U"missing-file-\u65E5.png")));
+
+            // snail emoji, outside of Unicode Basic Multilingual Plane
+            CHECK(!image.loadFromFile(std::filesystem::path(U"missing-file-\U0001F40C.png")));
+
             CHECK(image.getSize() == sf::Vector2u(0, 0));
             CHECK(image.getPixelsPtr() == nullptr);
         }
@@ -410,22 +422,46 @@ TEST_CASE("[Graphics] sf::Image")
             SECTION("To .bmp")
             {
                 filename /= "test.bmp";
-                CHECK(image.saveToFile(filename));
             }
 
             SECTION("To .tga")
             {
                 filename /= "test.tga";
-                CHECK(image.saveToFile(filename));
             }
 
             SECTION("To .png")
             {
                 filename /= "test.png";
-                CHECK(image.saveToFile(filename));
+            }
+
+            SECTION("To Spanish Latin1 filename .png")
+            {
+                // small n with tilde, from Spanish, outside of ASCII, inside common Latin 1 codepage
+                filename /= U"test-\u00f1.png";
+            }
+
+            SECTION("To Polish filename .png")
+            {
+                // small n with acute accent, from Polish, outside of Latin 1 codepage
+                filename /= U"test-\u0144.png";
+            }
+
+            SECTION("To Japanese CJK filename .png")
+            {
+                // CJK symbol for Sun, outside of any European language codepage
+                filename /= U"test-\u65E5.png";
+            }
+
+            SECTION("To emoji non-BMP Unicode filename .png")
+            {
+                // snail emoji, outside of Unicode Basic Multilingual Plane
+                filename /= U"test-\U0001F40C.png";
             }
 
             // Cannot test JPEG encoding due to it triggering UB in stbiw__jpg_writeBits
+
+            REQUIRE(image.saveToFile(filename));
+            REQUIRE(std::filesystem::exists(filename));
 
             const sf::Image loadedImage(filename);
             CHECK(loadedImage.getSize() == sf::Vector2u(256, 256));


ChrisThrasher:

src/SFML/Graphics/Image.cpp
```diff
@@ -75,6 +76,27 @@ int eof(void* user)
...
+    auto& stream = *static_cast<std::ifstream*>(user);
+    if (!stream.seekg(size, std::ios_base::cur))
+        sf::err() << "Failed to seek image loader std::ifstream" << std::endl;
+}
+
+int eofStdIfStream(void* user)
+{
+    auto& stream = *static_cast<std::ifstream*>(user);
+    return stream.eof();
+}
```
ChrisThrasher, line 98: Can you convert these to lambdas that are defined right before they're used?

src/SFML/System/Utils.cpp
```diff
@@ -44,8 +44,10 @@ std::string toLower(std::string str)
 std::string formatDebugPathInfo(const std::filesystem::path& path)
 {
     std::ostringstream oss;
-    oss << "    Provided path: " << path << '\n' //
-        << "    Absolute path: " << std::filesystem::absolute(path);
+    // convert to UTF-8 to handle non-ascii/non-latin1 filenames on windows
+    // cast to work in C++20 where u8string is char8_t which can't be printed to char stream
```
ChrisThrasher, line 48:
```suggestion
    // cast is required to work in C++20 where u8string is char8_t which can't be printed to char stream
```

test/Graphics/Image.test.cpp
```diff
@@ -265,6 +265,20 @@ TEST_CASE("[Graphics] sf::Image")
             CHECK(!image.loadFromFile("."));
             CHECK(!image.loadFromFile("this/does/not/exist.jpg"));
 
+            std::array<char32_t, 10> utf32name = {'X', 'f', 'a', 'k', 'e', '.', 'p', 'n', 'g', '\0'};
```
ChrisThrasher, line 268:
```suggestion
            utf32name = U"Xfake.png";
```
No need for arrays. We can use string literals.
https://en.cppreference.com/w/cpp/language/string_literal

test/Graphics/Image.test.cpp
```diff
@@ -265,6 +265,20 @@ TEST_CASE("[Graphics] sf::Image")
...
+
+            utf32name[0] = 0xf1; // small n with tilde, outside of ascii but inside common latin1 codepage
+            CHECK(!image.loadFromFile(std::filesystem::path(utf32name.data())));
```
ChrisThrasher, line 271:
```suggestion
            // small n with tilde, outside of ascii but inside common latin1 codepage
            CHECK(!image.loadFromFile(std::filesystem::path(U"\xF1fake.png")));
```
Because we can use string literals, we can just pass the string to the function rather than store it on the stack and give it a name.

test/Graphics/Image.test.cpp
```diff
@@ -405,7 +419,8 @@ TEST_CASE("[Graphics] sf::Image")
 
         SECTION("Successful save")
         {
-            auto filename = std::filesystem::temp_directory_path();
+            auto                    filename  = std::filesystem::temp_directory_path();
+            std::array<char32_t, 6> utf32name = {'X', '.', 'p', 'n', 'g', '\0'};
```
ChrisThrasher, line 423: Another place we can use string literals instead. The comments are still great to explain what the hex codes mean so keep those please.

ChrisThrasher (changes requested):

test/Graphics/Image.test.cpp
```diff
@@ -265,6 +265,20 @@ TEST_CASE("[Graphics] sf::Image")
...
 
+            std::array<char32_t, 10> utf32name = {'X', 'f', 'a', 'k', 'e', '.', 'p', 'n', 'g', '\0'};
+
+            utf32name[0] = 0xf1; // small n with tilde, outside of ascii but inside common latin1 codepage
+            CHECK(!image.loadFromFile(std::filesystem::path(utf32name.data())));
+
+            utf32name[0] = 0x144; // small n with acute accent, outside of latin1 codepage
+            CHECK(!image.loadFromFile(std::filesystem::path(utf32name.data())));
+
+            utf32name[0] = 0x65E5; // CJK symbol for sun, outside of any european codepage
```
learn-more, line 276: Please note that installed codepages can differ per installation.
This is one of the reasons that the testbots for the Wine project have many different installations:
image

(Scroll down on https://testbot.winehq.org/ for the full list)

test/Graphics/Image.test.cpp
learn-more: This is a different string, because your first char here is `\xF1fa`

FRex: Yes, and `\x{}` and `\u{}` is from C++23. I used ASCII in code and arrays to be 100% sure no funny stuff like that happens.

test/Graphics/Image.test.cpp
learn-more: `wchar_t` is commonly 2 bytes on Windows-like systems, and 4 bytes on Linux.
`std::filesystem::path` should accept the 4-bytes type, but will do a conversion on Windows.

test/Graphics/Image.test.cpp
FRex: That's totally fine and we don't want or need any codepages to be present and/or missing. This code deals with only Unicode and using std:: libs so it should be totally independent of locales, codepages, everything.

The whole point is to not use `std::filesystem::path::string()` at all since it uses the current codepage to do the conversion and will fail for characters that can't be represented, like Russian or Polish letters if you have the usual Western 850 codepage that has Western European characters for English, German, French, Spanish, etc. characters only.

Most American (I guess?) developers will probably have that codepage on their system and most exposure to French or Spanish languages because they are popular in Canada/USA, so they might test their Unicode handling code with some character like é, and see it working, and think it's done, but no, it was just luck.

That's why I have those 4 specific test cases: Latin 1 (just for completness), non-Latin 1 (to catch anyone who has Latin 1 code page set and passed first test), CJK (no West/East Europe codepage will save you from this test) and non-BMP (to be sure characters that take 2 UTF-16 code units also work, this was a problem historically since almost no one ever used non-BMP chars in the wild, even CJK languages are mostly within BMP plane of Unicode, but since emojis became super popular this made softwares fix this).

src/SFML/Graphics/Image.cpp
FRex: Yes, but old callbacks (for SFML stream class) do not do that. So do we make all callbacks be functions or convert the SFML ones to lambdas for consistency, or do we keep SFML callbacks as functions and add new ifstream ones as lambdas (so there's a mismatch)?

test/Graphics/Image.test.cpp
FRex: Yes, I know `wchar_t` is UTF-16 on Windows (some call it WTF-16, since it allows unpaired surrogates, some think it's broken or limited to UCS-2 but it's not).

The `char32_t` is supposed to be UTF-32 in C++ and `std::filesystem::path` is supposed to them take care of converting that to native format (16 on Windows, 8 on Linux) and stuff like `std::ifstream` is supposed to do the right thing.

I used 32 bit on purpose here, since I later put an emoji (which wouldn't fit in Windows `wchar`) in a name for a test, to make sure such characters (emojis, super niche languages, anything out of Basic Multilingual Plane) do work too, since they take 2 `wchar_t` to encode on Windows in UTF-16.

Some software used to break once emojis started becoming popular since they assumed Unicode is a 16-bit mapping, each char is single `wchar_t` on Windows (and no one ever used the higher numbered characters since they are dead language or very niche CJK stuff or such, emojis are first time these high codes are really spotted in the wild and among normal people), etc. and this test would catch such mistake. We never substr, take length with `wcslen`, etc. of the native `wchar` path on Windows, so we have 0 potential for such bugs, but it's for completness, it's a Unicode filename test, so a non-BMP character names should be tested too.

It also makes test 100% portable for any OS, it's always UTF-32 from us -> path native -> open file. We don't care or see what native format is, once UTF-32 goes inside std:: classes it's done from our point of view, it's then up to them.

test/Graphics/Image.test.cpp
learn-more: You could do something like `U"\xF1"  "fake.png"` (like what I did in godbolt) to try to ensure it, but that already loses the 'readability' advantage imo.

FRex: Yes, or use `"missingX.png"`. I put X first so it's `arr[0] = 0xWhatever`, no need to think/find/remember offset, and used "fake" instead of "missing" or "does-not-exist" to make array init less typing. There's no real reason it's just 1 unicode char or that it's first, so with literal we can just put unicode last (before dot) and use more than 1 char per string too.

src/SFML/Graphics/Image.cpp
ChrisThrasher: Just make thew new ones lambdas. We can change the existing ones later if we want.

test/Graphics/Image.test.cpp
ChrisThrasher: The `U` prefix connotes an UTF-32 string literal. Am I mistaken?

test/Graphics/Image.test.cpp
FRex: I put in a whole other filename "missing" + unicode char + the extension (not that extension even matters when missing file).

src/SFML/Graphics/Image.cpp
FRex: I made all new load/save ones lambda.

I also added printing error reason in loadFromFile, from strerror, if opening file fails, so it will say something informative, not the "Image not of any known type, or corrupt" from STB.

test/Graphics/Image.test.cpp
FRex: It does mean `char32_t` which is UTF-32 yes and available from C++11, on my own CI I have just new tests, no fix yet, to check that the `\u` and `\U` fail as expected because they work to bring in those unicode chars. https://github.com/FRex/SFML/actions/runs/13043068592/job/36388867375

The \u is 4 digit Unicode escape and \U is an 8 digit one. Seems to be all fine.

src/SFML/System/Utils.cpp
FRex: This comment change was done and merged in separate PR https://github.com/SFML/SFML/pull/3407 and this commit is now gone from this PR (we're down to 2).

ChrisThrasher:

src/SFML/Graphics/Image.cpp
```diff
@@ -212,11 +213,43 @@ bool Image::loadFromFile(const std::filesystem::path& filename)
     // Clear the array (just in case)
     m_pixels.clear();
 
+    std::ifstream file(filename, std::ios::binary);
+    if (!file.is_open())
+    {
+
```
kimci86, line 219: Unwanted empty line

test/Graphics/Image.test.cpp
```diff
@@ -425,6 +437,36 @@ TEST_CASE("[Graphics] sf::Image")
...
+                // CJK symbol for Sun, outside of any European language codepage
+                filename /= U"test-\u65E5.png";
+                CHECK(image.saveToFile(filename));
+            }
+
+            SECTION("To emoji non-BMP Unicode filename .png")
+            {
+                // snail emoji, outside of Unicode Basic Multilingual Plane
+                filename /= U"test-\U0001F40C.png";
+                CHECK(image.saveToFile(filename));
```
kimci86, line 465: Each section is ending with this line. It could as well be written once out of sections.

src/SFML/Graphics/Image.cpp
FRex: That's true, I'll try to not forget to remove it next time we make changes to this PR.

test/Graphics/Image.test.cpp
FRex: @ChrisThrasher  will have to say on this. I have no opinion or knowledge what's better in catch2.

ChrisThrasher (approved):

ChrisThrasher:

test/Graphics/Image.test.cpp
ChrisThrasher: Good idea. I implemented that in my recent change.

hernandezcalzadillamijail-ux (commented):
@eXpl0it3r

Issue #3406, opened 2025-01-28
Unicode Support within SFML


eXpl0it3r:
Every C++ dev will agree that unicode support for C++ is a pain, inconsistent across platforms and non-trivial to get right.

SFML has places where proper unicode support is lacking. This issue serves as a catch-all for this topic, so we can more easily see what has been fixed and what remains open.

## Unicode Path Handling

- [x] `sf::Image::saveToFile()` / `sf::Image::loadFromFile()` / `sf::Texture::loadFromFile()`
    - Issue: #647 
    - PR: #3403 
- [x] `sf::SoundBuffer::saveToFile()`
    - PR: #3421
- [x] `sf::SoundBuffer::loadFromFile()`
    - Tests: #3416
- [x] `sf::Shader::loadFromFile()`
    - Tests: #3410
- [x] `sf::Music::openFromFile()`
    - Tests: #3416
- [x] `sf::Font::openFromFile()`
    - PR: #3422 
- [x] `sf::InputSoundFile::openFromFile()` 
    - Tests: #3415
- [x] `sf::OutputSoundFile::openFromFile()`
   - Tests: #3419
   - PR: #3421
- [ ] `sf::Ftp` ?
- [x] `sf::FileInputStream` 
    - Tests: #3413

## Unicode Rendering

- [x] Complex Text Layout
    - Issue: #246
    - PR: #3543
- [ ] Font fallback
    - Issue: #1536
- [ ] Multi-texture rendering
    - Note: Prerequisite for font fallback support
    - PR: #2282
- [ ] Emoji support & colored fonts support

## Unicode String Handling

- [ ] Correct conversion
    - Issue: #407 
- [x] `sf::Utf<8>` handling
    - Comment: https://github.com/SFML/SFML/issues/3406#issuecomment-2625052363
- [ ] Invalid sequence handling
    - Comment: https://github.com/SFML/SFML/issues/3406#issuecomment-2625052363
- [ ] Consider replacing SFML unicode conversion with cpp-unicodelib
- [x] `sf::Clipboard`
    - PR: #3435
- [ ] `sf::String::String(wchar_t)` allows for constructing from one half of a UTF-16 surrogate pair
- [x] `sf::WindowBase::setTitle`
    - PR: https://github.com/SFML/SFML/pull/3437

## Resources

- Unicode tool: https://frex.github.io/unicode.html
- https://tonsky.me/blog/unicode/
- https://www.unicode.org/standard/principles.html
[...]

PR #3407, opened 2025-01-29
Always print filepaths as UTF-8 to a char stream #3406

FRex:
[x] Has this change been discussed on [the forum](https://en.sfml-dev.org/forums/index.php#c3) or in an issue before?
[x] Does the code follow the SFML [Code Style Guide](https://www.sfml-dev.org/style.php)?
[x] Have you provided some example/test code for your changes?
[ ] If you have additional steps which need to be performed, please list them as tasks!

## Description

This PR is related to the issue https://github.com/SFML/SFML/issues/3406 and https://github.com/SFML/SFML/issues/647

This PR is one of the commits from the PR https://github.com/SFML/SFML/pull/3403

The only commit in this PR is the one that uses UTF-8 to print filepath to `sf::err()`.

I was asked to make it its own PR on Discord by @ChrisThrasher 

## Tasks

-   [ ] Tested on Linux
-   [x] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

Run this program. `sf::Shader` is used since it already uses `std::ifstream` and supports Unicode
filenames, so there is no exception due to using `.string()` on Windows there.

```cpp
#include <SFML/Graphics.hpp>
#include <iostream>

int main(void)
{
    const char32_t totry[] = {
        0xF1, // spanish n tilde
        0x144, // polish n accent
        0x65E5, // sun cjk char
        0x1F40C, // snail emoji
    };

    sf::Shader shader;
    for(const auto c : totry)
    {
        char32_t fname[] = U"missing-file-X.txt";
        fname[13] = c;
        shader.loadFromFile(fname, sf::Shader::Type::Vertex);
    }
}
```

Without this PR the result will be abort (due to uncaught exception) on the first Unicode path that is printed that isn't possible to represent in current codepage (I have Western Europe codepage set so ñ passes but ń doesn't):

image

image

---

With this PR change applied the output in cmd will be (by default for me since I have Western Europe codepage set):

image

With `chcp.com 65001` ran before the program, the console is set to "codepage" for UTF-8, which results in:
image

The boxes are due to missing fonts for CJK and emojis in cmd terminal, but if the text is redirected to file or copy pasted from the terminal, it's fully correct:

```
Failed to open shader file
    Provided path: missing-file-ñ.txt
    Absolute path: D:\vs\SFML-3-testing\missing-file-ñ.txt
Failed to open shader file
    Provided path: missing-file-ń.txt
    Absolute path: D:\vs\SFML-3-testing\missing-file-ń.txt
Failed to open shader file
    Provided path: missing-file-日.txt
    Absolute path: D:\vs\SFML-3-testing\missing-file-日.txt
Failed to open shader file
    Provided path: missing-file-🐌.txt
    Absolute path: D:\vs\SFML-3-testing\missing-file-🐌.txt
```

### DIFF
--- a/src/SFML/System/Utils.cpp
+++ b/src/SFML/System/Utils.cpp
@@ -44,8 +44,10 @@ std::string toLower(std::string str)
 std::string formatDebugPathInfo(const std::filesystem::path& path)
 {
     std::ostringstream oss;
-    oss << "    Provided path: " << path << '\n' //
-        << "    Absolute path: " << std::filesystem::absolute(path);
+    // convert to UTF-8 to handle non-ascii/non-latin1 filenames on windows
+    // cast is required to work in C++20 where u8string is char8_t which can't be printed to char stream
+    oss << "    Provided path: " << reinterpret_cast<const char*>(path.u8string().c_str()) << '\n' //
+        << "    Absolute path: " << reinterpret_cast<const char*>(std::filesystem::absolute(path).u8string().c_str());
     return oss.str();
 }
 


ChrisThrasher (approved):

ChrisThrasher:

src/SFML/System/Utils.cpp
```diff
@@ -44,8 +44,10 @@ std::string toLower(std::string str)
 std::string formatDebugPathInfo(const std::filesystem::path& path)
 {
     std::ostringstream oss;
-    oss << "    Provided path: " << path << '\n' //
-        << "    Absolute path: " << std::filesystem::absolute(path);
+    // convert to UTF-8 to handle non-ascii/non-latin1 filenames on windows
+    // cast is required to work in C++20 where u8string is char8_t which can't be printed to char stream
+    oss << "    Provided path: " << reinterpret_cast<const char*>(path.u8string().c_str()) << '\n' //
+        << "    Absolute path: " << reinterpret_cast<const char*>(std::filesystem::absolute(path).u8string().c_str());
```
ChrisThrasher, line 50: The cast to `const char*` makes sense to me. After that though, how does the `ostream` "know" that we're giving it UTF-8 text? Is it possible that someone's terminal is not configured to handle or expect UTF-8 leading to mangled console text?

FRex: The cast to `char*` is only for C++20 and up, because they changed UTF-8 string to be u8string of `char8_t`...

The `ostream` doesn't know. It's all opaque/binary to it. Only part that "knows" anything is terminal. In screenshot above with codepage 850 (Western Europe so Spanish, German, French, etc.) it shows bad chars there, but doesn't crash. Even with terminal set to Unicode due to missing fonts I have no CJK or emojis in terminal.

Short of changing codepage ourselves (very iffy for a library to do IMO, since it's global state) this is best we can do.

If you know what you are doing you can recover original text. The reason it shows weird chars there is that the UTF-8 bytes (so 2, 3 or 4 for each non-ASCII Unicode) are interpreted as bytes in that codepage and corresponding glyps are shown (including 2 or more per Unicode char, like ñ has 2 byte UTF-8 representation so in codepage 850 it comes out as 2 chars: `├▒`).

If you copy that text, encode it as the codepage you have used (850 for me) and decode it as UTF-8, you can get the text back:

```
text = r"""
Failed to open shader file
    Provided path: missing-file-├▒.txt
    Absolute path: D:\vs\SFML-3-testing\missing-file-├▒.txt
Failed to open shader file
    Provided path: missing-file-┼ä.txt
    Absolute path: D:\vs\SFML-3-testing\missing-file-┼ä.txt
Failed to open shader file
    Provided path: missing-file-µùÑ.txt
    Absolute path: D:\vs\SFML-3-testing\missing-file-µùÑ.txt
Failed to open shader file
    Provided path: missing-file-­ƒÉî.txt
    Absolute path: D:\vs\SFML-3-testing\missing-file-­ƒÉî.txt
"""

print(text.encode("CP850").decode("UTF-8"))
```

gives me the original Unicode names:

```

Failed to open shader file
    Provided path: missing-file-ñ.txt
    Absolute path: D:\vs\SFML-3-testing\missing-file-ñ.txt
Failed to open shader file
    Provided path: missing-file-ń.txt
    Absolute path: D:\vs\SFML-3-testing\missing-file-ń.txt
Failed to open shader file
    Provided path: missing-file-日.txt
    Absolute path: D:\vs\SFML-3-testing\missing-file-日.txt
Failed to open shader file
    Provided path: missing-file-🐌.txt
    Absolute path: D:\vs\SFML-3-testing\missing-file-🐌.txt
```

PR #3408, opened 2025-01-30
Quote include paths in SFML and FreeType

eXpl0it3r:
## Description

Someone reported a build issue with FreeType, because their build directory path contained a space.
This adds missing quotes to FreeType and also ensures that they're all quoted in SFML itself.

## Tasks

-   [ ] Tested on Linux
-   [x] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

Clone the SFML CMake template into a directory with a space in it.


binary1248 (approved):

ChrisThrasher:
My impression was that CMake always quotes paths so spaces in paths don't cause problems. I've tested this a lot myself. I've asked others for reproducible examples. I even asked the inventor of CMake about this in person and so far nobody could prove that quoting path variables necessary. Is there any way we can reproduce this problem so I can look deeper into it? Quoting these variables certainly won't hurt, but I'd like to know more about what is causing this.

eXpl0it3r:
Here's the link to the [Discord chat](https://discord.com/channels/175298431294636032/175298431294636032/1334449131262967902).

The user was using MinGW (i.e. GCC) on Windows. Might be that it's not a problem for CMake, but if it's passed directly to GCC without quotes, GCC might not like it.

ChrisThrasher:
Has that user tested this PR yet to confirm it works?

eXpl0it3r:
I have reproduced the issue and tested the resolution of the FreeType stuff

kimci86:
This got me curious because quoting variables shouldn't be needed.

As Craig Scott says,
> Whitespace argument separators are found before variable evaluation.
> Semicolon argument separators are found after variable evaluation.

https://crascit.com/2022/01/25/quoting-in-cmake/

After some investigation, I could reproduce the problem with:
- MinGW compiler linked on SFML download page: [WinLibs UCRT 14.2.0 (64-bit)](https://github.com/brechtsanders/winlibs_mingw/releases/download/14.2.0posix-19.1.1-12.0.0-ucrt-r2/winlibs-x86_64-posix-seh-gcc-14.2.0-mingw-w64ucrt-12.0.0-r2.7z)
- CMake with Ninja generator
- space in build folder path. I tried with `C:\spa ce\bu ild`

The problem happens when building `ftver.rc` file with `windres.exe`. For example:
```
C:\SPACE~1\mingw64\bin\windres.exe -O coff -DFT2_BUILD_LIBRARY -D_CRT_NONSTDC_NO_WARNINGS -D_CRT_SECURE_NO_WARNINGS -I "C:/spa ce/bu ild/_deps/freetype-build/include" -I "C:/spa ce/bu ild/_deps/freetype-src/include" -I "C:/spa ce/bu ild/_deps/freetype-build/include/freetype/config"  "C:/spa ce/bu ild/_deps/freetype-src/src/base/ftver.rc" _deps/freetype-build/CMakeFiles/freetype.dir/src/base/ftver.rc.obj
```

This PR does not solve the issue.
It turns out it is a known issue in windres. See for example: https://gitlab.kitware.com/cmake/cmake/-/issues/20920

Strangely enough, there is no issue when using "MinGW Makefiles" CMake generator instead of Ninja. That is because somehow it uses DOS short names, so there are no spaces.
```
cd /d C:\SPACE~1\BUILD2~1\_deps\FREETY~3 && C:\SPACE~1\mingw64\bin\windres.exe -O coff -DFT2_BUILD_LIBRARY -D_CRT_NONSTDC_NO_WARNINGS -D_CRT_SECURE_NO_WARNINGS -I C:\SPACE~1\BUILD2~1\_deps\FREETY~3\include -I C:\SPACE~1\BUILD2~1\_deps\FREETY~2\include -I C:\SPACE~1\BUILD2~1\_deps\FREETY~3\include\freetype\config  "C:\spa ce\bu ild2\_deps\freetype-src\src\base\ftver.rc"
```

Maybe an idea to avoid the problem would be to patch dependencies to exclude .rc files (maybe they are not critical), but SFML also adds its own .rc files when building DLLs, which add nice metadata (version, copyright, etc, see [resource.rc.in](https://github.com/SFML/SFML/blob/master/tools/windows/resource.rc.in)). SFML's own .rc files trigger the windres issue too.

To summarize, adding quotes does not solve the issue, but we can suggest workarounds:
- do not have spaces in the build folder path
- or use "MinGW Makefiles" generator

ChrisThrasher:
Thanks for the analysis! That's really insightful.

I'm curious how it is that Exploiter came to a different conclusion about the efficacy of this PR.

eXpl0it3r:
Closing this, as adding quotes doesn't actually do anything and it seems to be a [CMake / MinGW / ninja / winres.exe issue](https://gitlab.kitware.com/cmake/cmake/-/issues/20920).

I don't think it's worth patching any third-party code for this and rather tell people who run into this sort of issue, to use MinGW Makefiles or a source/build directory that doesn't contain spaces.

PR #3418, opened 2025-02-04
Revamp `sf::String` tests

ChrisThrasher:
## Description

Related to #3406 

I wrote the `sf::String` tests (see #2466) with no knowledge of how Unicode works, therefore the tests did a poor job actually testing all the Unicode-aware features of the class. I rewrote most of the tests to use Unicode codepoints more often. These new tests will do a much better job catching regressions.

In this same PR I also included a separate commit that removes hex code from strings in favor of the actual Unicode characters those hex values correspond to. Nobody would know that `\U0001F40D` means the 🐌 emoji. I know that because `\U0001F40D` is actually 🐍! The snail is `\U0001F40C`. Aren't you glad you can just see the character with your own eyes without having to check the web while reading the tests? Adding these characters required adding the `/utf-8` code to MSVC builds.

These tests mostly exercise happy path code. There are many more tests we could theoretically add with error cases and non-default `std::locale` values but I chose to hold off on those for now.

### DIFF
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -274,6 +274,10 @@ macro(sfml_add_library module)
         target_compile_definitions(${target} PUBLIC "SFML_STATIC")
     endif()
 
+    # Enable support for UTF-8 characters in source code
+    if(SFML_COMPILER_MSVC)
+        target_compile_options(${target} PRIVATE /utf-8)
+    endif()
 endmacro()
 
 # add a new target which is a SFML example
@@ -350,6 +354,11 @@ macro(sfml_add_example target)
     if(SFML_OS_WINDOWS AND SFML_USE_MESA3D)
         add_dependencies(${target} "install-mesa3d")
     endif()
+
+    # Enable support for UTF-8 characters in source code
+    if(SFML_COMPILER_MSVC)
+        target_compile_options(${target} PRIVATE /utf-8)
+    endif()
 endmacro()
 
 # add a new target which is a SFML test
@@ -405,6 +414,11 @@ function(sfml_add_test target SOURCES DEPENDS)
         if(SFML_OS_ANDROID)
             set_target_properties(${target} PROPERTIES CROSSCOMPILING_EMULATOR "${PROJECT_BINARY_DIR}/run-in-adb-shell.sh")
         endif()
+    endif()
+
+    # Enable support for UTF-8 characters in source code
+    if(SFML_COMPILER_MSVC)
+        target_compile_options(${target} PRIVATE /utf-8)
     endif()
 
     # Add the test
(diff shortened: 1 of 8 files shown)


ChrisThrasher:
Attn: @FRex

cmake/Macros.cmake
```diff
@@ -274,6 +274,10 @@ macro(sfml_add_library module)
         target_compile_definitions(${target} PUBLIC "SFML_STATIC")
     endif()
 
+    # Enable support for UTF-8 characters in source code
+    if(SFML_COMPILER_MSVC)
+        target_compile_options(${target} PRIVATE /utf-8)
```
kimci86, line 279: This is needed only for tests, right? Why adding it for libraries and examples too?
Also, the comment could tell why this option is needed.

test/System/String.test.cpp
```diff
@@ -256,56 +256,68 @@ TEST_CASE("[System] sf::String")
...
-                const sf::String string = "def";
-                CHECK(std::string(string) == "def"s);
-                CHECK(std::wstring(string) == L"def"s);
-                CHECK(string.toAnsiString() == "def"s);
-                CHECK(string.toWideString() == L"def"s);
-                CHECK(string.toUtf8() == sf::U8String{'d', 'e', 'f'});
-                CHECK(string.toUtf16() == u"def"s);
-                CHECK(string.toUtf32() == U"def"s);
-                CHECK(string.getSize() == 3);
+                const sf::String string = "Escargo";
```
kimci86, line 260: In case you wanted to use a French word, it should be "Escargot".

cmake/Macros.cmake
ChrisThrasher: > This is needed only for tests, right?

Correct. It's because of the viralitiy of pre-compiled headers. For those to work all TUs must use the same arguments. Not my preference but otherwise CI would fail.

eXpl0it3r: I guess one question is whether `/source-charset:utf-8` is enough. Are they an downsides to using utf8 in the executing charset?

https://learn.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-170

ChrisThrasher: https://github.com/ChrisThrasher/SFML/actions/runs/13162292727/job/36733841275

CI is upset if we do that.

eXpl0it3r (approved):
I feel like `select` is quite a non-speaking name. I had to scroll up and read the comment just to understand what the what's supposed to do.

ChrisThrasher:
> I feel like `select` is quite a non-speaking name. I had to scroll up and read the comment just to understand what the what's supposed to do.

There's definitely room for improvement here. I'm open to suggestions. I can't think of a good name off the top of my head.

PR #3422, opened 2025-02-07
Fix opening `sf::Font` from non-ASCII path

ChrisThrasher:
## Description

Related to https://github.com/SFML/SFML/issues/3406

This PR is more complicated than most Unicode support PRs because we have to create a stream that remains alive for the lifetime of the `sf::Font` object. Most code is factored out into an `openFromStreamImpl` function that all 3 opening functions use.


src/SFML/Graphics/Font.cpp
```diff
@@ -211,122 +192,38 @@ bool Font::openFromMemory(const void* data, std::size_t sizeInBytes)
...
-    auto fontHandles = std::make_shared<FontHandles>();
-
-    // Initialize FreeType
-    // Note: we initialize FreeType for every font instance in order to avoid having a single
-    // global manager that would create a lot of issues regarding creation and destruction order.
-    if (FT_Init_FreeType(&fontHandles->library) != 0)
+    if (!data)
     {
-        err() << "Failed to load font from memory (failed to initialize FreeType)" << std::endl;
+        err() << "Failed to load font from memory - data pointer is nullptr" << std::endl;
```
eXpl0it3r, line 197:
```suggestion
        err() << "Failed to load font from memory (provided data pointer is nullptr)" << std::endl;
```

src/SFML/Graphics/Font.cpp
```diff
@@ -211,122 +192,38 @@ bool Font::openFromMemory(const void* data, std::size_t sizeInBytes)
...
-                           static_cast<FT_Long>(sizeInBytes),
-                           0,
-                           &face) != 0)
-    {
-        err() << "Failed to load font from memory (failed to create the font face)" << std::endl;
-        return false;
-    }
-    fontHandles->face = face;
+    // Create memroy stream, the memory is owned by the user
+    const auto memorystream = std::make_shared<MemoryInputStream>(data, sizeInBytes);
```
eXpl0it3r, line 202:
```suggestion
    const auto memoryStream = std::make_shared<MemoryInputStream>(data, sizeInBytes);
```

eXpl0it3r (changes requested):
Just some small things, looks good otherwise 👍

eXpl0it3r (approved):

PR #3425, opened 2025-02-12
Win32: Recover from invalid device context (DC)

jkeller51:
## Description

This code change will check the result of `SwapBuffers()`. If `SwapBuffers()` fails, it will retrieve a new device context from Windows and proceed with the new context.

I made this change on my fork in 2021, evidently facing an issue where the window wasn't redrawing in some scenario because the device context needed to be updated. I no longer recall the exact scenario so I can't currently reproduce, but I thought it might be useful anyway.

In short, in case Windows decides to give the window a new DC, this change allows SFML to recover from that without freezing up.

### DIFF
--- a/src/SFML/Window/Win32/WglContext.cpp
+++ b/src/SFML/Window/Win32/WglContext.cpp
@@ -224,8 +224,17 @@ bool WglContext::makeCurrent(bool current)
 ////////////////////////////////////////////////////////////
 void WglContext::display()
 {
-    if (m_deviceContext && m_context)
-        SwapBuffers(m_deviceContext);
+    if (m_deviceContext && m_context) 
+    {
+        bool success = SwapBuffers(m_deviceContext);
+        if (!success) 
+        {
+            err() << "SwapBuffers error: " << GetLastError() << std::endl;
+            ReleaseDC(m_window, m_deviceContext);
+            m_deviceContext = GetDC(m_window);
+            success = SwapBuffers(m_deviceContext);
+        }
+    }
 }
 
 
--- a/src/SFML/Window/Win32/WglContext.cpp
+++ b/src/SFML/Window/Win32/WglContext.cpp
@@ -226,8 +226,7 @@ void WglContext::display()
 {
     if (m_deviceContext && m_context)
     {
-        bool success = SwapBuffers(m_deviceContext);
-        if (!success)
+        if (!SwapBuffers(m_deviceContext))
         {
             err() << "SwapBuffers error: " << GetLastError() << std::endl;
             ReleaseDC(m_window, m_deviceContext);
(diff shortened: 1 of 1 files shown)


src/SFML/Window/Win32/WglContext.cpp
```diff
@@ -224,8 +224,17 @@ bool WglContext::makeCurrent(bool current)
...
-        SwapBuffers(m_deviceContext);
+    if (m_deviceContext && m_context) 
+    {
+        bool success = SwapBuffers(m_deviceContext);
+        if (!success) 
+        {
+            err() << "SwapBuffers error: " << GetLastError() << std::endl;
+            ReleaseDC(m_window, m_deviceContext);
+            m_deviceContext = GetDC(m_window);
+            success = SwapBuffers(m_deviceContext);
```
ChrisThrasher, line 235: This is a dead store. `success` is never read after this. Let's not store this boolean at all since we don't need it. Also formatting is slightly incorrect. I tried to fix this and force-update the branch but I seem to not be allowed to do that so please go check out the Formatting CI job and make the changes it requests.

ChrisThrasher (changes requested):

src/SFML/Window/Win32/WglContext.cpp
```diff
@@ -225,7 +225,16 @@ bool WglContext::makeCurrent(bool current)
 void WglContext::display()
 {
     if (m_deviceContext && m_context)
-        SwapBuffers(m_deviceContext);
+    {
+        bool success = SwapBuffers(m_deviceContext);
+        if (!success)
```
ChrisThrasher, line 230:
```suggestion
        if (!SwapBuffers(m_deviceContext))
```

ChrisThrasher (changes requested):

binary1248:
I really don't see what this is supposed to protect against. If `SwapBuffers` actually fails in some rare scenario it is an indicator something else went wrong way earlier in code that wasn't caught properly. Recreating the device context in this situation is equivalent to sweeping the real problem under the rug. We don't even know if recreating the device context in this way can lead to other negative side effects. Device contexts don't only belong to windows but pbuffers as well, which this change completely ignores.

In all the years I have been debugging OpenGL applications I have never once seen `SwapBuffers` fail, and to be honest I wouldn't even know how to provoke such a failure without doing something obviously wrong with some other code. If you also can't provide more information about how to reproduce the issue that this change is supposed to solve we can't even test if this change actually makes things better and whether it introduces any other negative side effects in such a case.

For all we know what you observed could have just been the result of some undefined behaviour or data race introduced somewhere else in your code that you eventually fixed and thus can't reproduce any longer.

jkeller51:
My application runs as a plugin to other software (digital audio workstations), so if what you're saying is true, it's possible that the other software was causing the problem and my tweak was just keeping my plugin chugging along anyway.

Because it prints an error, I wouldn't say it's sweeping anything under the rug, but rather adds an additional check & recovers gracefully from an error state in the field.

PR #3429, opened 2025-02-16
Add a CMake option to disable internet tests

jcowgill:
* The `getLocalAddress` test requires a functional network connection or it will fail.
* The `getPublicAddress` test attempts to access the internet. Although it has a fallback, Debian (and probably other distros) doesn't allow any package builds to use the internet in their tests so it would be nice to have an option to avoid it.

So add a CMake option to avoid running these tests similar to the existing `SFML_RUN_*_TESTS` options.

### DIFF
--- a/test/Network/IpAddress.test.cpp
+++ b/test/Network/IpAddress.test.cpp
@@ -71,15 +71,19 @@ TEST_CASE("[Network] sf::IpAddress")
 
     SECTION("Static functions")
     {
-        SECTION("getLocalAddress")
+        // These functions require external network access to work thus imposing an additional
+        // requirement on our test suite of internet access. This causes issues for developers
+        // trying to work offline and for package managers who may be building and running the
+        // tests offline as well.
+        (void)[]
         {
             const std::optional<sf::IpAddress> ipAddress = sf::IpAddress::getLocalAddress();
             REQUIRE(ipAddress.has_value());
             CHECK(ipAddress->toString() != "0.0.0.0");
             CHECK(ipAddress->toInteger() != 0);
-        }
+        };
 
-        SECTION("getPublicAddress")
+        (void)[]
         {
             const std::optional<sf::IpAddress> ipAddress = sf::IpAddress::getPublicAddress(sf::seconds(1));
             if (ipAddress.has_value())
@@ -87,7 +91,7 @@ TEST_CASE("[Network] sf::IpAddress")
                 CHECK(ipAddress->toString() != "0.0.0.0");
                 CHECK(ipAddress->toInteger() != 0);
             }
-        }
+        };
     }
 
     SECTION("Static constants")


ChrisThrasher:
Why do packagers need to run the tests? Is that not out of scope of what a package manager does?

jcowgill:
> Why do packagers need to run the tests? Is that not out of scope of what a package manager does?

Why not run the tests? I also want to check the package actually works in the environment I'm building it in. There are so many things that could break the package that you won't be testing: different compilers, different library versions, installing files in slightly different places, maybe I applied a small patch to the package...

(The change in this MR is already patched in Debian in a different way - I am trying to upstream my changes)

ChrisThrasher:
Can we apply the same solution as https://github.com/SFML/CSFML/pull/393? That would be much preferable to all this additional complexity.

jcowgill:
> Can we apply the same solution as [SFML/CSFML#393](https://github.com/SFML/CSFML/pull/393)? That would be much preferable to all this additional complexity.

In that PR I could remove the old tests and replace them with something else that worked just as well.

With `getLocalAddress` and `getPublicAddress` I think I would have to remove the tests entirely? That would solve what I'm trying to do but then you have no tests.

ChrisThrasher:
> > Can we apply the same solution as [SFML/CSFML#393](https://github.com/SFML/CSFML/pull/393)? That would be much preferable to all this additional complexity.
> 
> In that PR I could remove the old tests and replace them with something else that worked just as well.
> 
> With `getLocalAddress` and `getPublicAddress` I think I would have to remove the tests entirely? That would solve what I'm trying to do but then you have no tests.

Pardon me I now see that CSFML PR was touching different functions. Ignore what I said.

I think I'd just like to see these tests removed. I've also been annoyed in the by this function not working when I'm not on the internet. It won't be too harmful to just stop running it. Let's keep compiling them but omit running them. We've done this once before. Here is what that pattern looks like:

https://github.com/SFML/SFML/blob/f59913e997c9f0eef9ac88f721acd72f5be36c89/test/Window/WindowBase.test.cpp#L209-L213

https://github.com/SFML/SFML/blob/f59913e997c9f0eef9ac88f721acd72f5be36c89/test/Window/WindowBase.test.cpp#L252

Replace the `SECTION` with `(void)[]` to create a temporary lambda and add a semicolon at the end.

ChrisThrasher (approved):

ChrisThrasher:

ChrisThrasher:
Thanks for taking the time to match the style for how we do optional tests! Sorry that work went to waste, but I think the new version of this PR should suit your needs. I'm ready to merge this as soon as you can confirm it will fix your problem.

eXpl0it3r:
Just a heads-up, as we introduce more network capabilities and new tests, we've opted for introducing CMake options with #3667 after all.

- `SFML_RUN_IPV4_LINK_TESTS`
- `SFML_RUN_IPV4_INTERNET_TESTS`
- `SFML_RUN_IPV6_LINK_TESTS`
- `SFML_RUN_IPV6_INTERNET_TESTS`

PR #3431, opened 2025-02-17
Revert to no -s suffix for sfml-main

eXpl0it3r:
## Description

In #2787 we removed this override as it mentioned compatibility with FindSFML.cmake this however now also changed in the way that `sfml-main` will generate a `sfml-main-s` and `sfml-main-s-d` variant, instead of the previous `sfml-main` and `sfml-main-d`. As `sfml-main` is always a static library, I'm not sure if it makes sense to mark it with the `-s` suffix.

Note, that if want to keep this behavior, we'd have to revert #3346, as I assumed we didn't use the `-s` suffix for `sfml-main`.

## How to test this PR?

Build SFML on Windows and check the create library names.

### DIFF
--- a/src/SFML/Main/CMakeLists.txt
+++ b/src/SFML/Main/CMakeLists.txt
@@ -15,6 +15,13 @@ endif()
 # define the sfml-main target
 sfml_add_library(Main STATIC SOURCES ${SRC})
 
+# don't use -s suffix for sfml-main
+set_target_properties(sfml-main PROPERTIES
+                      DEBUG_POSTFIX -d
+                      RELEASE_POSTFIX ""
+                      MINSIZEREL_POSTFIX ""
+                      RELWITHDEBINFO_POSTFIX "")
+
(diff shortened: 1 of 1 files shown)


ChrisThrasher (approved):

ChrisThrasher:
```
/usr/local/etc/buildbot-worker/freebsd-gcc-64/build/src/SFML/Window/FreeBSD/JoystickImpl.cpp:318:48: error: implicit conversion changes signedness: 'int' to 'size_type' (aka 'unsigned long') [-Werror,-Wsign-conversion]                    m_state.buttons[buttonIndex++] = hid_get_data(m_buffer.data(), &item);                    ~~~~~~~         ~~~~~~~~~~~^~1 error generated.
```
I think we need to cherry-pick https://github.com/SFML/SFML/pull/3381. We might as well do that in this PR too to save us some time.

eXpl0it3r:
Now that we backported all the other commits, I've dropped that fix again.

PR #3435, opened 2025-02-18
Fix using Emoji with `sf::Clipboard`

ChrisThrasher:
## Description

Related to https://github.com/SFML/SFML/issues/3406

I was able to fix using Emoji with the Windows clipboard by switching from a wide string to a UTF-16 string. This makes sense because wide strings are UCS-2 on Windows which is a fixed-width encoding, not able to encode Emoji.

### DIFF
--- a/src/SFML/Window/Win32/ClipboardImpl.cpp
+++ b/src/SFML/Window/Win32/ClipboardImpl.cpp
@@ -33,6 +33,7 @@
 #include <SFML/System/Win32/WindowsHeader.hpp>
 
 #include <ostream>
+#include <string_view>
 
 #include <cstring>
 
@@ -65,7 +66,8 @@ String ClipboardImpl::getString()
         return text;
     }
 
-    text = String(static_cast<wchar_t*>(GlobalLock(clipboardHandle)));
+    const std::u16string_view string(static_cast<const char16_t*>(GlobalLock(clipboardHandle)));
+    text = String::fromUtf16(string.begin(), string.end());
     GlobalUnlock(clipboardHandle);
 
     CloseClipboard();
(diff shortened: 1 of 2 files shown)


ChrisThrasher:
Attn: @FRex

FRex:
At quick glance it looks good. Technically it's not `wchar`/`wstring` at fault, but `sf::String::toWideString`. We could fix it in general too, for anyone using it around WinAPI, or those people can just use UTF-16 directly like we do here. Technically (and othan straight dump to/from disk or similar) a `wchar`/`wstring` you can't even use 100% correctly without knowing if you're on Windows or other OS.

The important part in this PR is `stringSize` calculation. It's correct (encode then check length of encoded string), but before PR it was wrong - it assumed length of `wstring` is same as of UTF-32 `sf::String`, which isn't true for chars above `0xffff` (outside BMP) which are 1 UTF-32 char/code unit, but two UTF-16 (as surrogate pair) or 4 bytes in UTF-8: https://frex.github.io/unicode.html?text=%F0%9F%90%8C

The commit message and PR title say "emoji", but it's not just emojis that are outside of Basic Multilingual Plane (BMP). There's obscure languages, private use area, musical marks, Egyptian Hieroglyphs, rare or new CJK chars, etc. there too. Emojis just happen to be the most widespread use of these high codepoints and once they became common online, a lot of software that assumed Unicode = UCS-2/BMP/3-byte-UTF-8-max got caught by it and then fixed. A few non-emoji non-BMP chars: https://frex.github.io/unicode.html?text=%F0%93%80%80%F0%9D%84%9E

eXpl0it3r (approved):

PR #3436, opened 2025-02-19
iOS fixes

JonnyPtn:
Been a while since I've properly looked at iOS bits so there's a few things accumulated:
- Removed the CI tests for shared libraries on iOS as they [aren't supported by apple](https://developer.apple.com/library/archive/technotes/tn2435/_index.html#//apple_ref/doc/uid/DTS40017543-CH1-TROUBLESHOOTING-BUNDLE_ERRORS)
- SFML main module depends on SFML window and requires a fix to make sure the delegate is loaded correctly
- We need to use the main module to run the tests
- We *can* use command line arguments for iOS builds under some circumstances (debugging mostly)
- We need to add the test resources to the bundle
- Cursor, Keyboard and Vulkan test source files are excluded as not supported on iOS
- ~~Test validating the window size is set correctly excluded by preprocessor as on iOS it always returns display size~~ Now just excluded all windows tests as requested

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,6 +99,11 @@ jobs:
             arch: arm64-v8a
             api: 33
           type: { name: Debug, flags: -DCMAKE_BUILD_TYPE=Debug -DSFML_FATAL_OPENGL_ERRORS=ON }
+        exclude: # Shared libraries are not supported on iOS
+        - platform: { name: iOS }
+          config: { name: Shared }
+        - platform: { name: iOS Xcode }
+          config: { name: Shared }
 
 
     steps:
--- a/cmake/Macros.cmake
+++ b/cmake/Macros.cmake
@@ -383,7 +383,7 @@ function(sfml_add_test target SOURCES DEPENDS)
     source_group("" FILES ${SOURCES})
 
     # create the target
-    add_executable(${target} ${SOURCES})
+    add_executable(${target} ${SOURCES} ${PROJECT_SOURCE_DIR}/test/main.cpp)
 
     # enable precompiled headers
     if (SFML_ENABLE_PCH)
--- a/src/SFML/Main/CMakeLists.txt
+++ b/src/SFML/Main/CMakeLists.txt
@@ -28,4 +28,6 @@ if(SFML_OS_ANDROID)
 
     # glad sources
     target_include_directories(sfml-main SYSTEM PRIVATE "${PROJECT_SOURCE_DIR}/extlibs/headers/glad/include")
+elseif(SFML_OS_IOS)
+    target_link_libraries(sfml-main PUBLIC SFML::Window)
 endif()
--- a/src/SFML/Main/MainiOS.mm
+++ b/src/SFML/Main/MainiOS.mm
@@ -43,17 +43,19 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Config.hpp>
 
+#include <SFML/Window/iOS/SFAppDelegate.hpp>
+
 #include <UIKit/UIKit.h>
 
 
 ////////////////////////////////////////////////////////////
 int main(int argc, char** argv)
 {
-    // Note: we intentionally drop command line arguments,
-    // there's no such thing as a command line on an iOS device :)
-
-    // Important: "SFAppDelegate" must always match the name of the
-    // application delegate class defined in sfml-window
-
-    return UIApplicationMain(argc, argv, nil, @"SFAppDelegate");
+    NSString* appDelegateClassName = {};
+    @autoreleasepool
+    {
+        // Setup code that might create autoreleased objects goes here.
+        appDelegateClassName = NSStringFromClass([SFAppDelegate class]);
+    }
+    return UIApplicationMain(argc, argv, nil, appDelegateClassName);
 }
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -32,7 +32,15 @@ add_library(sfml-test-main STATIC
     TestUtilities/AudioUtil.cpp
 )
 target_include_directories(sfml-test-main PUBLIC TestUtilities)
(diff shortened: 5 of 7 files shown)


eXpl0it3r:

src/SFML/Main/CMakeLists.txt
```diff
@@ -21,4 +21,6 @@ if(SFML_OS_ANDROID)
 
     # glad sources
     target_include_directories(sfml-main SYSTEM PRIVATE "${PROJECT_SOURCE_DIR}/extlibs/headers/glad/include")
+elseif(SFML_OS_IOS)
+    target_link_libraries(sfml-main PUBLIC SFML::Window)
```
eXpl0it3r, line 25: Don't think we should create this dependency.

src/SFML/Main/MainiOS.mm
```diff
@@ -43,17 +43,19 @@
...
-    // application delegate class defined in sfml-window
-
-    return UIApplicationMain(argc, argv, nil, @"SFAppDelegate");
+    NSString* appDelegateClassName = nullptr;
+    @autoreleasepool
+    {
+        // Setup code that might create autoreleased objects goes here.
+        appDelegateClassName = NSStringFromClass([SFAppDelegate class]);
+    }
+    return UIApplicationMain(argc, argv, nil, appDelegateClassName);
```
eXpl0it3r, line 60: Why is it not enough to use a fixed string?

eXpl0it3r (changes requested):

src/SFML/Main/CMakeLists.txt
JonnyPtn: It already exists - the main module on iOS requires the app delegate class from the window module, so are you proposing we re-organise the modules so that this dependency isn't necessary?

.github/workflows/ci.yml
```diff
@@ -99,6 +99,11 @@ jobs:
             arch: arm64-v8a
             api: 33
           type: { name: Debug, flags: -DCMAKE_BUILD_TYPE=Debug -DSFML_FATAL_OPENGL_ERRORS=ON }
+        exclude:
```
eXpl0it3r: I'd add a comment to explain the exclusion

```suggestion
        exclude: # Shared libraries are not supported on iOS
```

cmake/Macros.cmake
```diff
@@ -383,7 +383,7 @@ function(sfml_add_test target SOURCES DEPENDS)
     source_group("" FILES ${SOURCES})
 
     # create the target
-    add_executable(${target} ${SOURCES})
+    add_executable(${target} ${SOURCES} ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/../test/main.cpp)
```
eXpl0it3r:
```suggestion
    add_executable(${target} ${SOURCES} ${PROJECT_SOURCE_DIR}/test/main.cpp)
```

eXpl0it3r (changes requested):
Just one small change.

@ChrisThrasher would be interested in hearing your opinion w.r.t. the test changes and disabling of some things. For example, any better ideas for the window size checks?

cmake/Macros.cmake
eXpl0it3r: Did you forget to push that commit or add the file to the commit? 😄

JonnyPtn: main.cpp? Looks to me like it's added correctly, and the tests would fail without it

eXpl0it3r:

eXpl0it3r (approved):

eXpl0it3r:

test/Window/WindowBase.test.cpp
```diff
@@ -86,31 +101,43 @@ TEST_CASE("[Window] sf::WindowBase", runDisplayTests())
...
         }
 
         SECTION("Mode, title, style, and state")
         {
             windowBase.create(sf::VideoMode({240, 360}), "WindowBase Tests", sf::Style::Resize, sf::State::Windowed);
             CHECK(windowBase.isOpen());
+#ifndef SFML_SYSTEM_IOS
+            // Fails on iOS as the window size always returns the display size
             CHECK(windowBase.getSize() == sf::Vector2u(240, 360));
+#endif
```
ChrisThrasher: I'd rather omit these tests from iOS than add these preprocessor conditionals.

test/main.cpp
```diff
@@ -0,0 +1,10 @@
+#include <catch2/catch_session.hpp>
+
+#include <SFML/Main.hpp>
+#include <array>
```
ChrisThrasher:
```suggestion
```
Can this be removed?

test/CMakeLists.txt
```diff
@@ -131,6 +156,27 @@ if(SFML_RUN_DISPLAY_TESTS)
     target_compile_definitions(test-sfml-graphics PRIVATE SFML_RUN_DISPLAY_TESTS)
 endif()
 
+if (XCODE AND SFML_OS_IOS)
+    set(GRAPHICS_RESOURCE
```
ChrisThrasher:
```suggestion
    set(GRAPHICS_RESOURCES
```

ChrisThrasher (dismissed):

test/Window/WindowBase.test.cpp
JonnyPtn: You don't see value in any of the other tests? So just exclude this file as per the other test files in this PR?

eXpl0it3r: @ChrisThrasher So you don't want to run these tests on iOS? Or is there something else that one could do that wouldn't require these preprocessor conditionals?

ChrisThrasher: > So you don't want to run these tests on iOS?

No, I don't think so. I'd rather keep it simple and just omit tests on iOS that require significant mods.

eXpl0it3r (approved):

eXpl0it3r:
Thank you for those fixes! 🥳

PR #3437, opened 2025-02-19
Fix using Emoji in Windows window titles

ChrisThrasher:
## Description

Related to #3406 

Just like #3435, the fix is to use UTF-16 instead of UCS-2 which is what wide strings current encode. Perhaps we should change `sf::String::toWideString` to return a Unicode string instead. Windows is already treating a `wchar_t[]` as being UTF-16 so I think it would be worth doing.

I have manually tested that Emoji titles work on macOS and Linux. I do need someone to help me test on Windows though.

EDIT: FRex has confirmed this PR fixes the bug on Windows 👍🏻

### DIFF
--- a/src/SFML/Window/Win32/WindowImplWin32.cpp
+++ b/src/SFML/Window/Win32/WindowImplWin32.cpp
@@ -213,7 +213,7 @@ m_cursorGrabbed(m_fullscreen)
 
     // Create the window
     m_handle = CreateWindowW(className,
-                             title.toWideString().c_str(),
+                             reinterpret_cast<const wchar_t*>(title.toUtf16().c_str()),
                              win32Style,
                              left,
                              top,
@@ -358,7 +358,7 @@ void WindowImplWin32::setSize(Vector2u size)
 ////////////////////////////////////////////////////////////
 void WindowImplWin32::setTitle(const String& title)
 {
-    SetWindowTextW(m_handle, title.toWideString().c_str());
+    SetWindowTextW(m_handle, reinterpret_cast<const wchar_t*>(title.toUtf16().c_str()));
 }
 
 


ChrisThrasher:
Attn: @FRex

FRex:
Yes, it works as expected with test string 🐌𓀀👻 https://frex.github.io/unicode.html?text=%F0%9F%90%8C%F0%93%80%80%F0%9F%91%BB

image

I tried a 𝄞 too ( https://frex.github.io/unicode.html?text=%F0%9D%84%9E ) and it shown a box there, but using WinSpy I can see it's actually there (by copying it out, WinSpy also is missing this font character and shows a box, it must be WinAPI shortcoming).

image

Test program:

```cpp
#include <SFML/Graphics.hpp>
int main()
{
    const std::u32string utf32str(U"\U0001F40C\U00013000\U0001F47B");
    const sf::String sfstr(utf32str);
    sf::RenderWindow win(sf::VideoMode(sf::Vector2u(800, 600)), sfstr);
    while(win.isOpen())
    {
        while(const auto eve = win.pollEvent())
            if(eve->is<sf::Event::Closed>())
                win.close();
        win.clear();
        win.display();
    }
}
```

eXpl0it3r (approved):
Wasn't the point of `toWideString` to do this conversion to `wchar_t*`? Why is it not working?

ChrisThrasher:
> Wasn't the point of `toWideString` to do this conversion to `wchar_t*`? Why is it not working?

`toWideString` converts the string to UCS-2. This encoding is similar to UTF-16 but it's a fixed-width encoding which cannot encoding codepoints outside of the Basic Multilingual Plane, which means it cannot encode Emoji and other codepoints with particularly high values.

Issue #3433, opened 2025-02-17
sf::Packet checkSize() Integer Overflow DoS on 32-bit Targets


grapurabaki:
### Prerequisite Checklist

- [x] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [x] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [x] I am here to report an issue and not to just ask a question or look for help (use the [forum](https://en.sfml-dev.org/forums/index.php#c3) or [Discord](https://discord.gg/nr4X7Fh) instead)

### Describe your issue here

A malicious user can currently cause any receiver of arbitrary `sf::Packet`s (presumably most applications using SFML for networking) to crash on 32-bit systems. In every deserialisation method in `sf::Packet` which uses a payload size read from the packet (`std::string`, `char *` etc.) a length of `UINT32_MAX` will cause the following logic in `sf::Packet::checkSize()` to break on 32-bit systems:
```cpp
m_isValid = m_isValid && (m_readPos + size <= m_data.size());
```
The length stored in the packet is defined as `uint32_t`, and on 32-bit systems both `m_readPos` and `size` (defined as `size_t`) are also 32-bit. With a crafted packet with length `UINT32_MAX` the addition of `m_readPos` will overflow the comparison to a smaller number and pass the check, creating e.g. a `std::string` which far surpasses the maximum length or, perhaps more worringly, a `memcpy` with an effectively unlimited size into the location of a `char *`.

This is trivially exploited to cause crashes, though I assume anything more impactful would be impractical. The sheer size makes it very difficult to avoid crashing, unless you for example managed to send an `sf::Packet` in the gigabytes to bring down the size value required to overflow, but this seems unlikely.

This is not an issue on 64-bit systems where `size_t` is 64-bits, as the 32-bit size read from the packet will never overflow that comparison.

### Your Environment

N/A

### Steps to reproduce

```cpp
#include <iostream>
#include <cstdlib>
#include <cstdint>

#include <SFML/Network/Packet.hpp>

struct serial_str {
	uint32_t length;
	char data[2];
};

int main(void)
{
	serial_str str;
	str.length = 0xffffffff;
	str.data[0] = 'h';
	str.data[1] = 'i';

	sf::Packet packet;
	packet.append(&str, sizeof(str));

	std::string out;
[...]

PR #3441, opened 2025-02-20
Fix unsigned overflow in sf::Packet size check

grapurabaki:
## Description

This PR fixes the issue described in #3433.

## Tasks

-   [X] Tested on Linux
-   [ ] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

```cpp
#include <iostream>
#include <cstdlib>
#include <cstdint>

struct serial_str {
	uint32_t length;
	char data[2];
};

int main(void)
{
	serial_str str;
	str.length = 0xffffffff;
	str.data[0] = 'h';
	str.data[1] = 'i';

	sf::Packet packet;
	packet.append(&str, sizeof(str));

	std::string out;
	packet >> out;

	std::cout << out << std::endl;

	return EXIT_SUCCESS;
}
```
This code, which emulates a received crafted packet from the network, would previously crash when compiled in 32-bit mode. This is now fixed, the library no longer crashes and the packet is marked as invalid.

### DIFF
--- a/src/SFML/Network/Packet.cpp
+++ b/src/SFML/Network/Packet.cpp
@@ -571,6 +571,8 @@ Packet& Packet::operator<<(const String& data)
 ////////////////////////////////////////////////////////////
 bool Packet::checkSize(std::size_t size)
 {
+    // Unsigned overflow is defined behavior
+    m_isValid = m_isValid && !(m_readPos + size < m_readPos);
     m_isValid = m_isValid && (m_readPos + size <= m_data.size());
 
     return m_isValid;
--- a/src/SFML/Network/Packet.cpp
+++ b/src/SFML/Network/Packet.cpp
@@ -571,9 +571,9 @@ Packet& Packet::operator<<(const String& data)
 ////////////////////////////////////////////////////////////
 bool Packet::checkSize(std::size_t size)
 {
-    // Unsigned overflow is defined behavior
-    m_isValid = m_isValid && !(m_readPos + size < m_readPos);
-    m_isValid = m_isValid && (m_readPos + size <= m_data.size());
+    // Determine if size is big enough to trigger an overflow
+    const bool overflowDetected = m_readPos + size < m_readPos;
+    m_isValid = m_isValid && (m_readPos + size <= m_data.size()) && ! overflowDetected
 
     return m_isValid;
 }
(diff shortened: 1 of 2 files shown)


ChrisThrasher:
https://github.com/SFML/SFML/actions/runs/13440614127

I turned your sample code into a unit test and was able to demonstrate that this code fails on 32 bit platforms but succeeds on 64 bit platforms. I then cherry-picked this PR and can see that those tests now pass so I think this fix is good!

https://github.com/SFML/SFML/actions/runs/13441434031

Can you cherry-pick [this commit](https://github.com/SFML/SFML/commit/b4093d128a593aa2d7c7e186f37051546908264c) into your PR? I would do this myself but I can't because you opened your PR from your fork's `master` branch. In the future I would prefer you make a branch first then open a PR from that branch. That way maintainers can edit your PR which speeds up the review process. This will also save you time in the future when you have to update your fork after your PRs are merged.

grapurabaki:
Ah, sorry about that. Commit merged.

ChrisThrasher:

src/SFML/Network/Packet.cpp
```diff
@@ -571,6 +571,8 @@ Packet& Packet::operator<<(const String& data)
 ////////////////////////////////////////////////////////////
 bool Packet::checkSize(std::size_t size)
 {
+    // Unsigned overflow is defined behavior
+    m_isValid = m_isValid && !(m_readPos + size < m_readPos);
     m_isValid = m_isValid && (m_readPos + size <= m_data.size());
```
ChrisThrasher, line 576:
```suggestion
    // Determine if size is big enough to trigger an overflow
    const bool overflowDetected = (m_readPos + size < m_readPos);
    m_isValid = m_isValid && (m_readPos + size <= m_data.size()) && ! overflowDetected
```
Let's put a name to this new expression we're adding. Otherwise I like this and we can get this merged!

ChrisThrasher (changes requested):

ChrisThrasher:

ChrisThrasher:
¯\_(ツ)_/¯

ChrisThrasher:

ChrisThrasher:

ChrisThrasher (approved):

PR #3443, opened 2025-02-23
Use digit separators for large constants or hexadecimal doublets

vittorioromeo:
I have applied C++14 digit separators to large integral constants (over 10000) and to hexadecimal constants where the doublets are significant (e.g. colors or IPv4 addresses).

I think this improves readability quite a bit, especially in places where large microsecond constants are used (it's quite easy to accidentally read one million as 100'000).

### DIFF
--- a/include/SFML/Graphics/Color.inl
+++ b/include/SFML/Graphics/Color.inl
@@ -42,10 +42,10 @@ a(alpha)
 
 ////////////////////////////////////////////////////////////
 constexpr Color::Color(std::uint32_t color) :
-r(static_cast<std::uint8_t>((color & 0xff000000) >> 24)),
-g(static_cast<std::uint8_t>((color & 0x00ff0000) >> 16)),
-b(static_cast<std::uint8_t>((color & 0x0000ff00) >> 8)),
-a(static_cast<std::uint8_t>(color & 0x000000ff))
+r(static_cast<std::uint8_t>((color & 0xff'00'00'00) >> 24)),
+g(static_cast<std::uint8_t>((color & 0x00'ff'00'00) >> 16)),
+b(static_cast<std::uint8_t>((color & 0x00'00'ff'00) >> 8)),
+a(static_cast<std::uint8_t>((color & 0x00'00'00'ff) >> 0))
 {
 }
 
--- a/src/SFML/Audio/SoundFileWriterWav.cpp
+++ b/src/SFML/Audio/SoundFileWriterWav.cpp
@@ -59,10 +59,10 @@ void encode(std::ostream& stream, std::uint16_t value)
 void encode(std::ostream& stream, std::uint32_t value)
 {
     const std::array bytes = {
-        static_cast<char>(value & 0x000000FF),
-        static_cast<char>((value & 0x0000FF00) >> 8),
-        static_cast<char>((value & 0x00FF0000) >> 16),
-        static_cast<char>((value & 0xFF000000) >> 24),
+        static_cast<char>((value & 0x00'00'00'FF) >> 0),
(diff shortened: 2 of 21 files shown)


ChrisThrasher:
I'm a fan of this. I've started using this in newer code and really like how it helps me mentally parse bigger numbers.

https://clang.llvm.org/docs/ClangFormatStyleOptions.html#integerliteralseparator

By the way, there's a clang-format check to automate this. I've played with it in the past to automatically make this change to SFML and liked the results.

include/SFML/Graphics/Color.inl
```diff
@@ -42,10 +42,10 @@ a(alpha)
...
 ////////////////////////////////////////////////////////////
 constexpr Color::Color(std::uint32_t color) :
-r(static_cast<std::uint8_t>((color & 0xff000000) >> 24)),
-g(static_cast<std::uint8_t>((color & 0x00ff0000) >> 16)),
-b(static_cast<std::uint8_t>((color & 0x0000ff00) >> 8)),
-a(static_cast<std::uint8_t>(color & 0x000000ff))
+r(static_cast<std::uint8_t>((color & 0xff'00'00'00) >> 24)),
+g(static_cast<std::uint8_t>((color & 0x00'ff'00'00) >> 16)),
+b(static_cast<std::uint8_t>((color & 0x00'00'ff'00) >> 8)),
+a(static_cast<std::uint8_t>((color & 0x00'00'00'ff) >> 0))
```
ChrisThrasher, line 48: I definitely approve of all the changes to integers but I'm not yet sure what I think about applying to hexidecal literals. I'm not sure if I don't like or if it's just unfamiliar. I suppose all the same arguments apply to hex values.

vittorioromeo: In general I don't think *every* hexadecimal constant benefits from digit separators.

I picked the hexadecimal constants where doublets have a meaning (e.g. in the case above the each doublet is the value of a channel). I think it helps to visually separate those, as a potential typo/mistake such as `0x00000ff` is hard to spot compared to `0x00'00'0f'f`.

ChrisThrasher:
https://github.com/SFML/SFML/compare/master...ChrisThrasher:SFML:integer_literal_separators

For curiosity's sake I added the `IntegerLiteralSeparator` check to see what else it would change and it caught many more >5 digit integer literals that this PR did not cover. The only annoying change was how it formatted the preprocessor expression below but this isn't a big deal.

Screenshot 2025-02-23 at 10 26 54 AM

ChrisThrasher:

src/SFML/Audio/SoundFileWriterWav.cpp
```diff
@@ -126,8 +126,8 @@ bool SoundFileWriterWav::open(const std::filesystem::path&     filename,
         static constexpr auto speakerTopFrontCenter     = 0x2000u;
         static constexpr auto speakerTopFrontRight      = 0x4000u;
         static constexpr auto speakerTopBackLeft        = 0x8000u;
-        static constexpr auto speakerTopBackCenter      = 0x10000u;
-        static constexpr auto speakerTopBackRight       = 0x20000u;
+        static constexpr auto speakerTopBackCenter      = 0x10'000u;
+        static constexpr auto speakerTopBackRight       = 0x20'000u;
```
ChrisThrasher, line 130: I believe this hex value was mistakenly treated as a decimal. Did you intend to make this change?

ChrisThrasher:

ChrisThrasher (approved):

eXpl0it3r:
I'd claim that anyone used to reading hexadecimal values will find these separators more distracting than useful, but we can keep them for now.

PR #3451, opened 2025-02-26
Install pkgconfig only if module is built

eXpl0it3r:
## Description

Another upstream fix that was patched in vcpkg. Currently SFML always installs all pkgconfig files, even if you don't build for example sfml-audio. Moving the install command into the `Macros.cmake` should make sure that they're only installed if the module is also built.

This PR is related to the issue https://github.com/microsoft/vcpkg/pull/42913

## Tasks

-   [ ] Tested on Linux

## How to test this PR?

Install SFML on a platform supporting pkgconfig and don't build one module. the sfml-all.pc and the specifically deselected sfml-module.pc should not be installed.


ChrisThrasher (approved):

oomek:
This PR has caused a regression. I cannot say if it's related to how I build SFML by cross-compiling for Win32 with MXE on Ubuntu.
image

Issue #3454, opened 2025-03-02
WindowBase::handleEvents() causes C4789 / Stack corruption in VS2022


n67-exe:
### Prerequisite Checklist

- [x] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [x] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [x] I am here to report an issue and not to just ask a question or look for help (use the [forum](https://en.sfml-dev.org/forums/index.php#c3) or [Discord](https://discord.gg/nr4X7Fh) instead)

## The problem
Calling WindowBase::handleEvents(...) with lambdas causes a compiler error in Visual Studio 2022:
```
...\WindowBase.inl(69,1): error C4789: in function "void WindowBase::handleEvents<... ::<lambda_1> const &>(... ::<lambda_1> const &)" buffer "overloadSet" of size 8 bytes will be overrun; 4 bytes will be written starting at offset 5
```
## Investigation
I tried forcing the compiler to ignore the error using `#pragma warning(disable : 4789)`. 
In Debug builds the program compiles, however when i run it a "Debug Error!" message box pops up with the text:
``Run-Time Check Failure #2 - Stack around the variable 'overloadSet' was corrupted.``
In the debugger the same error can be seen, marked as an exception originating from the line 73 of [WindowBase.inl](https://github.com/SFML/SFML/blob/44339c3d6766f63b9563955ab4cbe33465fc8542/include/SFML/Window/WindowBase.inl):
```cpp
}
```
Release builds seem to compile and run fine.

Inspecting the commit history i [found](https://github.com/SFML/SFML/commit/44339c3d6766f63b9563955ab4cbe33465fc8542) that this piece of code inside ``struct OverloadSet`` was added to prevent "stack corruption on MSVC":
```cpp
#if defined(_MSC_VER) && !defined(__clang__)
    unsigned char dummy; // Dummy variable to ensure that this struct is not empty thus avoiding a crash due to an MSVC bug
#endif
```
## Solution
I was able to compile and run the program without issues after commenting out the dummy variable on line 43 in [WindowBase.inl](https://github.com/SFML/SFML/blob/44339c3d6766f63b9563955ab4cbe33465fc8542/include/SFML/Window/WindowBase.inl):
```cpp
[...]

PR #3456, opened 2025-03-08
Backport Remaining 3.0.1 Patches

ChrisThrasher:
## Description

See https://github.com/orgs/SFML/projects/8 for a list of PRs I back ported. I had to add a few more patches to avoid merge conflicts. 3.0.1 sticks to the [SemVer](https://semver.org) convention of fixing bugs without adding features.


ChrisThrasher:
I cherry-picked all commits from https://github.com/SFML/SFML/pull/3399 because it was necessary to fix a CI failure as well as it seems to be the fix for https://github.com/SFML/SFML/issues/3454.

eXpl0it3r (approved):
The Caller change in the WindowBase.inl was the only thing I was concerned about, but since it's a `priv` implementation, I'm fine with expanding it for SFML 3.0.1 

I guess in theory this would still be "ABI" compatible, as it's a header-only change and should be able to link against a new version of SFML, even if your application hasn't compile in this header. Then again I'm not ABI expert, nor do I believe it's as beneficial to SFML.

PR #3461, opened 2025-03-15
Fixed normalized texture coordinates when NPOT textures aren't supported

### LINKED ISSUE #3460: CoordinateType::Normalized broken if NPOT textures not supported (e.g. on Android)
### Prerequisite Checklist

- [x] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [x] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [x] I am here to report an issue and not to just ask a question or look for help (use the [forum](https://en.sfml-dev.org/forums/index.php#c3) or [Discord](https://discord.gg/nr4X7Fh) instead)

### Describe your issue here

When using CoordinateType::Normalized, the `textCoords` in vertices are relative to the internal `m_actualSize` of the texture instead of the size returned by `texture.getSize()`.

On platforms where NPOT textures aren't supported (which seems to be the case for Android with GLES 1), loading an image of size 768x896 will result in the actual size being 1024x1024.

The wrong part of the image is thus being rendered when specifying normalized texture coordinates when using textures with a size that isn't a multiple of 2.

### Your Environment

- Android
- SFML 3

While I'm testing on an Android emulator, the issue was originally reported to me when running on a Samsung A54, so the issue is not related to the emulation.

### Steps to reproduce

The following code would be broken on Android:
```cpp
#include <SFML/Graphics.hpp>

int main(int, char**)
{
    sf::RenderWindow window(sf::VideoMode::getDesktopMode(), "");
    window.setFramerateLimit(30);

    const sf::Texture texture("loading.png");

    std::vector<sf::Vertex> verticesPixels;
    verticesPixels.push_back({{  0,   0}, sf::Color::White, {  0, 5*128}});
    verticesPixels.push_back({{128,   0}, sf::Color::White, {128, 5*128}});
    verticesPixels.push_back({{  0, 128}, sf::Color::White, {  0, 6*128}});
    verticesPixels.push_back({{128, 128}, sf::Color::White, {128, 6*128}});

    std::vector<sf::Vertex> verticesNormalized;
    for (const auto& vertexPixels : verticesPixels)
    {
        auto& vertexNormalized = verticesNormalized.emplace_back(vertexPixels);
        vertexNormalized.texCoords.x /= texture.getSize().x;
        vertexNormalized.texCoords.y /= texture.getSize().y;
    }

    while (window.isOpen())
    {
        {
                window.close();
        }

        window.clear();

        sf::RenderStates states;
        states.texture = &texture;
        states.transform.translate({50, 50});

        // The first vertex array always renders correctly
        states.coordinateType = sf::CoordinateType::Pixels;
        window.draw(verticesPixels.data(), verticesPixels.size(), sf::PrimitiveType::TriangleStrip, states);

        // The second vertex array only renders correctly on systems supporting NPOT textures
        states.transform.translate({0, 150});
        states.coordinateType = sf::CoordinateType::Normalized;
        window.draw(verticesNormalized.data(), verticesNormalized.size(), sf::PrimitiveType::TriangleStrip, states);

        window.display();
    }
}
```

The `loading.png` image that was used:
Image

### Expected behavior

This is how it looks on Linux, and how it is supposed to look:
Image

### Actual behavior

This is how it looks on Android:
Image

texus:
This fixes #3460
See the original issue for details.

I also tested first rendering to an `sf::RenderTexture` and then using it's texture with normalized coordinates, in order to verify that the code is still correct when `m_pixelsFlipped` is true.

### DIFF
--- a/src/SFML/Graphics/Texture.cpp
+++ b/src/SFML/Graphics/Texture.cpp
@@ -931,7 +931,8 @@ void Texture::bind(const Texture* texture, CoordinateType coordinateType)
         glCheck(glBindTexture(GL_TEXTURE_2D, texture->m_texture));
 
         // Check if we need to define a special texture matrix
-        if ((coordinateType == CoordinateType::Pixels) || texture->m_pixelsFlipped)
+        if ((coordinateType == CoordinateType::Pixels) || texture->m_pixelsFlipped ||
+            ((coordinateType == CoordinateType::Normalized) && (texture->m_size != texture->m_actualSize)))
         {
             // clang-format off
             std::array matrix = {1.f, 0.f, 0.f, 0.f,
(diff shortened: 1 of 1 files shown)


binary1248 (approved):

eXpl0it3r:
Thank you for the fix!

PR #3462, opened 2025-03-18
Patch dependencies to fix unacceptably low minimum CMake requirement

ChrisThrasher:
## Description

Newer versions of CMake have ended or deprecated support for scripts that require 3.5 or 3.10 so I added a max policy version just barely new enough to silence the warning. This is required to ensure that those using CMake 4 don't have their builds suddenly break.


eXpl0it3r (approved):
This not only gets rid of the failures with CMake 4, but also removes the warnings in CMake 3 regarding the old versions

Issue #3143, opened 2024-06-30
C++20 Enhancements for SFML 4


ChrisThrasher:
### Prerequisite Checklist

- [X] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [X] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [X] I understand [SFML's scope](https://www.sfml-dev.org/contribute.php#general-considerations) and believe the feature fits it

### Describe your feature request here

SFML 4 will require C++20 or newer. This issue catalogs places where SFML could benefit from utilizing C++20 features. Some of these changes imply breaking the API while others are merely implementation improvements.

- Replace `sf::U8string` with `std::u8string`
- Use ranged algorithms (`std::ranges::{sort|copy|is_permutation|find|find_if|remove_if|is_sorted}`)
- Use `operator<=>` to define comparison operations
- Use `operator==() = default` to define (in)equality operations
- Use `<numbers>` for Pi constant
- Replace pointer/size parameter pairs with `std::span`

### Use Cases

N/A

### API Example

N/A

ZXShady:
also changing Shader::setUniform to use string_view instead of const string& since C++20 allows heterogenous lookups.

ZXShady:
sf::Vector2 can't be made an aggregate since it has another constructor that cannot be replaced

`SFML_SYSTEM_API Vector2(T r, Angle phi);`

TerensTare:
Just brainstorming, I think it's better if this case can use a named constructor (ie. a static method that returns a vector), which makes `Vector2` an aggregate. The code would probably like something like this:
```cpp
template <typename T>
struct Vector2
{
    static constexpr Vector2 fromPolar(T r, Angle phi)
    {
        auto phi_r = phi.asRadians();
        auto cos = std::cosf(phi_r);
        auto sin = std::sinf(phi_r);
        return Vector2{.x = r * cos, .y = r * sin};
    }

    T x, y;
};
```

ZXShady:
@TerensTare 


the uses of an aggregate Vector2 is not particularly helpful, as designated initializers really shine with multiple members not just 2 members that are known in order making it an aggregate really just saves us 10 lines of code or so.


also we also need to remove the conversion constructor to make it an aggregate and turn it into the explcicit conversion operator instead

```cpp

template<typename T>
[...]

PR #3467, opened 2025-03-26
Add modules support

mikomikotaishi:
## Description

This pull request adds support for C++20 modules, which are enabled using `SFML_BUILD_MODULES`. 

This PR is related to the issue #3143

## Tasks

-   [x] Tested on Linux
-   [ ] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

This PR can be tested by importing any of the following modules that now exist (wrapping around the existing headers):
- `sfml.audio`
- `sfml.graphics`
- `sfml.network`
- `sfml.system`
- `sfml.window`

The following should work:

```cpp
import sfml.graphics;

int main()
{
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        {
                window.close();
        }

        window.clear();
        window.display();
    }
}
```

Feel free to make any changes as necessary.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -445,3 +445,21 @@ sfml_set_option(CLANG_TIDY_EXECUTABLE clang-tidy STRING "Override clang-tidy exe
 add_custom_target(tidy
     COMMAND ${CMAKE_COMMAND} -DCLANG_TIDY_EXECUTABLE=${CLANG_TIDY_EXECUTABLE} -DPROJECT_BINARY_DIR=${PROJECT_BINARY_DIR} -P ./cmake/Tidy.cmake
     WORKING_DIRECTORY ${PROJECT_SOURCE_DIR} VERBATIM)
+
+# Option to build C++20 modules (requires CMake 3.28+)
+sfml_set_option(SFML_BUILD_MODULES OFF BOOL "ON to build SFML's C++20 modules (requires CMake 3.28+)")
+
+if(SFML_BUILD_MODULES)
+    # Check CMake version requirement for C++20 modules
+    if(CMAKE_VERSION VERSION_LESS 3.28)
+        message(FATAL_ERROR "Building SFML modules requires CMake 3.28 or newer. Your version: ${CMAKE_VERSION}")
+    endif()
+    
+    # Make sure all SFML components are built before modules
+    if(NOT TARGET sfml-system)
+        message(FATAL_ERROR "SFML modules require all SFML components to be built first")
+    endif()
+    
+    # Add modules directory
+    add_subdirectory(modules)
+endif()
--- a/modules/SFML/CMakeLists.txt
+++ b/modules/SFML/CMakeLists.txt
@@ -1,0 +1,50 @@
+cmake_minimum_required(VERSION 3.28)
+
+file(GLOB_RECURSE SFML_MODULES *.cppm)
+
+# Define configure_cpp_module_target function if not already defined
+if(NOT COMMAND configure_cpp_module_target)
+    function(configure_cpp_module_target target)
+        if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.28)
+            target_sources(${target} PUBLIC FILE_SET CXX_MODULES FILES ${SFML_MODULES})
+        else()
+            message(WARNING "C++ modules require CMake 3.28+. Using standard compilation.")
+            target_sources(${target} PRIVATE ${SFML_MODULES})
+        endif()
+    endfunction()
+endif()
+
+# Create the module library
+add_library(sfml)
+
+# Configure the module target
+configure_cpp_module_target(sfml)
+
+# Link with SFML libraries
+target_link_libraries(sfml
+    PUBLIC
+        SFML::Audio 
+        SFML::Graphics 
+        SFML::Network 
+        SFML::System 
+        SFML::Window
+)
+
+# Set target properties
+set_target_properties(sfml PROPERTIES 
+    CXX_STANDARD 23 
+    CXX_STANDARD_REQUIRED ON
+    CXX_EXTENSIONS OFF 
+    CXX_MODULE_DYNDEP OFF
+)
+
+# Create an alias target
+add_library(SFML::modules ALIAS sfml)
+
+# Installation configuration
+install(TARGETS sfml
+    EXPORT sfml-targets
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    FILE_SET CXX_MODULES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/SFML/modules
+)
--- a/modules/SFML/Config.cppm
+++ b/modules/SFML/Config.cppm
@@ -1,0 +1,20 @@
+/**
+ * @file Config.cppm
+ * @brief Module file for SFML macro constants.
+ */
+
+module;
+
+#include <SFML/Config.hpp>
+
+export module sfml.Config;
+
+/**
+ * @brief The SFML sf:: namespace 
+ */
+export namespace sf {
+    constexpr inline int VERSION_MAJOR = SFML_VERSION_MAJOR;
+    constexpr inline int VERSION_MINOR = SFML_VERSION_MINOR;
+    constexpr inline int VERSION_PATCH = SFML_VERSION_PATCH;
+    constexpr inline bool VERSION_IS_RELEASE = SFML_VERSION_IS_RELEASE;
+}
--- a/modules/SFML/audio.cppm
+++ b/modules/SFML/audio.cppm
@@ -1,0 +1,23 @@
+/**
+ * @file audio.cppm
+ * @brief Module file for SFML audio operations.
+ */
+
+export module sfml.audio;
+
+export import sfml.audio.InputSoundFile;
+export import sfml.audio.Listener;
+export import sfml.audio.Music;
+export import sfml.audio.OutputSoundFile;
+export import sfml.audio.PlaybackDevice;
+export import sfml.audio.Sound;
+export import sfml.audio.SoundBuffer;
+export import sfml.audio.SoundBufferRecorder;
+export import sfml.audio.SoundChannel;
+export import sfml.audio.SoundFileFactory;
+export import sfml.audio.SoundFileReader;
+export import sfml.audio.SoundFileWriter;
+export import sfml.audio.SoundRecorder;
+export import sfml.audio.SoundSource;
+export import sfml.audio.SoundStream;
+export import sfml.system;
--- a/modules/SFML/audio/InputSoundFile.cppm
+++ b/modules/SFML/audio/InputSoundFile.cppm
@@ -1,0 +1,18 @@
+/**
+ * @file InputSoundFile.cppm
+ * @brief Module file for the InputSoundFile class of the Audio module
+ */
+
+module;
+
+#include <SFML/Audio/InputSoundFile.hpp>
+
+export module sfml.audio.InputSoundFile;
+
+/**
+ * @namespace sf
+ * @brief The SFML sf:: namespace 
+ */
+export namespace sf {
+    using sf::InputSoundFile;
+}
--- a/modules/CMakeLists.txt
+++ b/modules/CMakeLists.txt
@@ -1,0 +1,1 @@
+add_subdirectory(SFML)
--- a/modules/SFML/audio/Listener.cppm
+++ b/modules/SFML/audio/Listener.cppm
@@ -1,0 +1,37 @@
+/**
+ * @file Listener.cppm
+ * @brief Module file for the Listener namespace of the Audio module
+ */
+
+module;
+
+#include <SFML/Audio/Listener.hpp>
+
+export module sfml.audio.Listener;
+
+/**
+ * @namespace sf
+ * @brief The SFML sf:: namespace 
+ */
+export namespace sf {
+    /**
+     * @namespace Listener
+     * @brief The SFML sf::Listener:: namespace
+     */
+    namespace Listener {
+        using sf::Listener::Cone;
+
+        using sf::Listener::setGlobalVolume;
+        using sf::Listener::getGlobalVolume;
+        using sf::Listener::setPosition;
+        using sf::Listener::getPosition;
+        using sf::Listener::setDirection;
+        using sf::Listener::getDirection;
+        using sf::Listener::setVelocity;
+        using sf::Listener::getVelocity;
+        using sf::Listener::setCone;
+        using sf::Listener::getCone;
+        using sf::Listener::setUpVector;
+        using sf::Listener::getUpVector;
+    }
+}
--- a/modules/SFML/audio/Music.cppm
+++ b/modules/SFML/audio/Music.cppm
@@ -1,0 +1,18 @@
+/**
+ * @file Music.cppm
+ * @brief Module file for the Music class of the Audio module
+ */
+
+module;
+
+#include <SFML/Audio/Music.hpp>
+
+export module sfml.audio.Music;
+
+/**
+ * @namespace sf
+ * @brief The SFML sf:: namespace 
+ */
+export namespace sf {
+    using sf::Music;
+}
--- a/modules/SFML/audio/OutputSoundFile.cppm
+++ b/modules/SFML/audio/OutputSoundFile.cppm
@@ -1,0 +1,18 @@
+/**
+ * @file OutputSoundFile.cppm
+ * @brief Module file for the OutputSoundFile class of the Audio module
+ */
+
+module;
+
+#include <SFML/Audio/OutputSoundFile.hpp>
+
+export module sfml.audio.OutputSoundFile;
+
+/**
+ * @namespace sf
+ * @brief The SFML sf:: namespace
+ */
+export namespace sf {
+    using sf::OutputSoundFile;
+}
--- a/modules/SFML/audio/PlaybackDevice.cppm
+++ b/modules/SFML/audio/PlaybackDevice.cppm
@@ -1,0 +1,27 @@
+/**
+ * @file PlaybackDevice.cppm
+ * @brief Module file for the PlaybackDevice namespace of the Audio module
+ */
+
+module;
+
+#include <SFML/Audio/PlaybackDevice.hpp>
+
+export module sfml.audio.PlaybackDevice;
+
+/**
+ * @namespace sf
+ * @brief The SFML sf:: namespace
+ */
+export namespace sf {
+    /**
+     * @namespace PlaybackDevice
+     * @brief The SFML sf::PlaybackDevice:: namespace
+     */
+    namespace PlaybackDevice {
+        using sf::PlaybackDevice::getAvailableDevices;
+        using sf::PlaybackDevice::getDefaultDevice;
+        using sf::PlaybackDevice::setDevice;
+        using sf::PlaybackDevice::getDevice;
+    }
+}
--- a/modules/SFML/audio/Sound.cppm
+++ b/modules/SFML/audio/Sound.cppm
@@ -1,0 +1,18 @@
+/**
+ * @file Sound.cppm
+ * @brief Module file for the Sound class of the Audio module
+ */
+
+module;
+
+#include <SFML/Audio/Sound.hpp>
+
+export module sfml.audio.Sound;
+
+/**
+ * @namespace sf
+ * @brief The SFML sf:: namespace 
+ */
+export namespace sf {
+    using sf::Sound;
+}
--- a/modules/SFML/audio/SoundBuffer.cppm
+++ b/modules/SFML/audio/SoundBuffer.cppm
@@ -1,0 +1,18 @@
+/**
+ * @file SoundBuffer.cppm
+ * @brief Module file for the SoundBuffer class of the Audio module
+ */
+
+module;
+
+#include <SFML/Audio/SoundBuffer.hpp>
+
+export module sfml.audio.SoundBuffer;
+
+/**
+ * @namespace sf
+ * @brief The SFML sf:: namespace
+ */
+export namespace sf {
+    using sf::SoundBuffer;
+}
--- a/modules/SFML/audio/SoundBufferRecorder.cppm
+++ b/modules/SFML/audio/SoundBufferRecorder.cppm
@@ -1,0 +1,18 @@
+/**
+ * @file SoundBufferRecorder.cppm
+ * @brief Module file for the SoundBufferRecorder class of the Audio module
+ */
+
+module;
+
+#include <SFML/Audio/SoundBufferRecorder.hpp>
+
+export module sfml.audio.SoundBufferRecorder;
+
+/**
+ * @namespace sf
+ * @brief The SFML sf:: namespace 
+ */
+export namespace sf {
+    using sf::SoundBufferRecorder;
+}
--- a/modules/SFML/audio/SoundChannel.cppm
+++ b/modules/SFML/audio/SoundChannel.cppm
@@ -1,0 +1,18 @@
+/**
+ * @file SoundChannel.cppm
+ * @brief Module file for the SoundChannel enum of the Audio module
+ */
+
+module;
+
+#include <SFML/Audio/SoundChannel.hpp>
+
+export module sfml.audio.SoundChannel;
+
+/**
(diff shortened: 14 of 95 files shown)


modules/SFML/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+        SFML::Audio 
+        SFML::Graphics 
+        SFML::Network 
+        SFML::System 
+        SFML::Window
+)
+
+# Set target properties
+set_target_properties(sfml PROPERTIES 
+    CXX_STANDARD 23
```
eXpl0it3r, line 35: Is C++20 not sufficient?

I'd prefer to use `target_compile_features(sfml PUBLIC cxx_std_20)`

modules/SFML/Config.cppm
```diff
@@ -0,0 +1,20 @@
+/**
+ * @file Config.cppm
+ * @brief Module file for SFML macro constants.
+ */
```
eXpl0it3r, line 4: That's not the header head style we're using

modules/SFML/Config.cppm
```diff
@@ -0,0 +1,20 @@
...
+
+module;
+
+#include <SFML/Config.hpp>
+
+export module sfml.Config;
+
+/**
+ * @brief The SFML sf:: namespace 
+ */
```
eXpl0it3r, line 14: We're using different documentation comment style

modules/SFML/audio/InputSoundFile.cppm
eXpl0it3r, line 1: I'd want to stick to capitalized directories, i.e. `modules/SFML/Audio/...`

modules/SFML/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
+cmake_minimum_required(VERSION 3.28)
+
+file(GLOB_RECURSE SFML_MODULES *.cppm)
```
eXpl0it3r, line 3: We don't use GLOB for SFML. The files should be listed explicitly.

modules/SFML/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+
+# Link with SFML libraries
+target_link_libraries(sfml
+    PUBLIC
+        SFML::Audio 
+        SFML::Graphics 
+        SFML::Network 
+        SFML::System 
+        SFML::Window
+)
```
eXpl0it3r, line 31: This currently doesn't  consider that `SFML_BUILD_<module>` can disable some modules, which would then fail here

modules/SFML/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+
+# Set target properties
+set_target_properties(sfml PROPERTIES 
+    CXX_STANDARD 23 
+    CXX_STANDARD_REQUIRED ON
+    CXX_EXTENSIONS OFF
```
eXpl0it3r, line 37: Do we have to enforce the standard and disable extensions for modules to work?

modules/SFML/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+        SFML::System 
+        SFML::Window
+)
+
+# Set target properties
+set_target_properties(sfml PROPERTIES 
+    CXX_STANDARD 23 
+    CXX_STANDARD_REQUIRED ON
+    CXX_EXTENSIONS OFF 
+    CXX_MODULE_DYNDEP OFF
```
eXpl0it3r, line 38: Is this required? Is there an equivalent CMake feature option?

eXpl0it3r:
Thank you for contributing this! We've already a few people try/use SFML as modules, so it would be quite interesting to be able to add optional support.

I'll add a few review comments that should get the PR into a good state. The final decision on whether we actually want to support this already, is still outstanding however.

CMakeLists.txt
```diff
@@ -445,3 +445,21 @@ sfml_set_option(CLANG_TIDY_EXECUTABLE clang-tidy STRING "Override clang-tidy exe
 add_custom_target(tidy
     COMMAND ${CMAKE_COMMAND} -DCLANG_TIDY_EXECUTABLE=${CLANG_TIDY_EXECUTABLE} -DPROJECT_BINARY_DIR=${PROJECT_BINARY_DIR} -P ./cmake/Tidy.cmake
     WORKING_DIRECTORY ${PROJECT_SOURCE_DIR} VERBATIM)
+
+# Option to build C++20 modules (requires CMake 3.28+)
+sfml_set_option(SFML_BUILD_MODULES OFF BOOL "ON to build SFML's C++20 modules (requires CMake 3.28+)")
```
eXpl0it3r, line 450: Don't think we need the `(requires CMake 3.28+)` in the option description

modules/SFML/Config.cppm
eXpl0it3r, line 1: Wondering if we should really create a separate directory `modules` or whether we can include this in the normal `src` and integrate the modules configuration in the existing CMake setup

modules/SFML/Config.cppm
```diff
@@ -0,0 +1,20 @@
+/**
+ * @file Config.cppm
+ * @brief Module file for SFML macro constants.
+ */
+
+module;
+
+#include <SFML/Config.hpp>
+
+export module sfml.Config;
```
eXpl0it3r, line 10: Is lower case a C++ recommendation or why not `SFML.Config` / `SFML.Audio` / etc. ?

eXpl0it3r (changes requested):

modules/SFML/Config.cppm
mikomikotaishi: I went with it to keep things separated and because I'll admit the pull request was hastily glued together after I realised that it worked with my project, nothing wrong with having `modules` be part of `src` but I left it separated for ease of convenience in writing the CMakeLists files

modules/SFML/audio/InputSoundFile.cppm
mikomikotaishi: Not a problem at all

mikomikotaishi:
Thanks for the comments! I'll follow up with the requested changes in due time.

migration.md
```diff
@@ -460,6 +460,20 @@ This a breaking change for the following APIs:
 - `bool sf::OutputSoundFile::openFromFile(...)`
 - `bool sf::SoundFileWriter::open(...)`
 
+## C++20 Modules Support
```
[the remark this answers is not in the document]
mikomikotaishi: Where do you recommend I include the notes otherwise?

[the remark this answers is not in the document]
eXpl0it3r: We could mention it in the changelog.md

modules/SFML/CMakeLists.txt
ChrisThrasher: No. Neither of those are required. target_compile_features is sufficient.

modules/SFML/Config.cppm
ChrisThrasher: I’d rather put this code in src/

modules/SFML/audio.cppm
```diff
@@ -0,0 +1,23 @@
...
+export import sfml.audio.Sound;
+export import sfml.audio.SoundBuffer;
+export import sfml.audio.SoundBufferRecorder;
+export import sfml.audio.SoundChannel;
+export import sfml.audio.SoundFileFactory;
+export import sfml.audio.SoundFileReader;
+export import sfml.audio.SoundFileWriter;
+export import sfml.audio.SoundRecorder;
+export import sfml.audio.SoundSource;
+export import sfml.audio.SoundStream;
```
ChrisThrasher, line 22: Please add trailing newlines to all files. I see you omitted that in a lot of these new files.

modules/SFML/audio/InputSoundFile.cppm
```diff
@@ -0,0 +1,18 @@
...
+
+#include <SFML/Audio/InputSoundFile.hpp>
+
+export module sfml.audio.InputSoundFile;
+
+/**
+ * @namespace sf
+ * @brief The SFML sf:: namespace 
+ */
+export namespace sf {
```
ChrisThrasher, line 16: Two comments on this file. First, the path needs to use `Audio` instead of `audio`. Note the capital letter.

Second, why have one file per SFML header? Why not one file per SFML module, so 5 total? I’m seeing a lot of boilerplate that will be tedious to maintain. Let’s cut down on as much of that as possible. Have you looked at how `fmt` does this? They have a `FMT_EXPORT` macro in all headers that lets the headers be used in a non-modules context but also can be used to add `export` every so that the modularized form of the library works.

ChrisThrasher (changes requested):
I’m open to this idea, but the current implementation needs some work before I can consider it.

modules/SFML/audio/InputSoundFile.cppm
mikomikotaishi: I specifically designed the modules to correspond to each header to offer granularity, even though strictly speaking module size is not proportional to compilation time, I wanted to offer this as an option. For me it just seemed simplest to compile modules on top of the headers to allow switching between the two at any time possible.

ChrisThrasher: Given how including large modules does not incur the same efficiently penalty as including large headers, I’d prefer to be no more granular than one module per SFML library.

ChrisThrasher:
Have you considered how this new modular build of SFML will fit into our existing unit test and CI infrastructure? To land a change of this size, we need some pretty substantial testing. I know from experience upgrading SFML to C++20 (in a separate unmerged branch) that upgrading all the CI tooling to support C++20 is actually pretty tricky. I’m curious how `libfmt` tackles this or if they simply don’t do any automated testing of their modular build.

modules/SFML/audio/InputSoundFile.cppm
mikomikotaishi: We could alternatively package them as partitions, which the C++ standard does recommend for large modules

mikomikotaishi:
> Have you considered how this new modular build of SFML will fit into our existing unit test and CI infrastructure? To land a change of this size, we need some pretty substantial testing. I know from experience upgrading SFML to C++20 (in a separate unmerged branch) that upgrading all the CI tooling to support C++20 is actually pretty tricky. I’m curious how `libfmt` tackles this or if they simply don’t do any automated testing of their modular build.

There's one test on libfmt for modules, we could follow this format
https://github.com/fmtlib/fmt/blob/master/test/module-test.cc

modules/SFML/audio/InputSoundFile.cppm
mikomikotaishi: So what do you think of using partitions instead? There would no longer be importing the submodules of each module, as partitions cannot be imported individually, but maintaining some level of separation between each header's corresponding partition would be a good idea for organisation whenever a new feature is added to easily identify where it needs to be re-exported from.

ChrisThrasher: > a good idea for organisation whenever a new feature is added to easily identify where it needs to be re-exported from.

I was picturing a solution where we have an `SFML_EXPORT` macro that expands into `export` when doing a modular build. That way we don't have to write some line of code far away to ensure a new class or function gets exported. I also want to prevent having one C++ modules file per C++ header. I'm not sure what is required to achieve that because I don't have much first hand experience writing C++20 modules.

I hope that better explains what I'm looking for.

ChrisThrasher: If the partitions are required to drastically cut down on the boilerplate then I'm cool with that. I just don't want one C++ modules source file per header. Is that possible?

mikomikotaishi: Yes, that works, but what should be done with regards to allowing projects to both `#include` and `import` both headers and modules within the same build, if this is a concern at all?

ChrisThrasher: Idk. I haven't fully thought through every implication. I'm still figuring out how I want to do this since it's somewhat uncharted territory for me.

mikomikotaishi: I'll change as many of the aforementioned requests up to now and leave the undetermined things up to you to decide then

mikomikotaishi:
Okay, I changed everything that was requested up to now, except for completely merging module re-exports of existing declarations in header files that @ChrisThrasher requested, and tests are still yet to be written. Seeing as these are still an experimental feature, I think for the time being it should be pulled and then as time progresses and more users give feedback more concrete changes can be made to the modules.

Seeing as the addition of modules is probably a worthwhile feature to mention, I suggest adding this somewhere to the changelogs.
```
## C++20 Modules Support

SFML 3 now adds support for C++20 modules, which can be imported using the `import` keyword. The library itself is a single module, `sfml`.
```

mikomikotaishi:
Hi, are there anything missing changes from my recent update that you want addressed before you are ready to accept and merge the pull request?

src/SFML/Modules/System.cppm
```diff
@@ -0,0 +1,79 @@
...
+
+export namespace sf {
+    using sf::Angle;
+    using sf::Clock;
+    using sf::Exception;
+    using sf::FileInputStream;
+    using sf::InputStream;
+    using sf::MemoryInputStream;
+    #ifdef SFML_SYSTEM_ANDROID
+    using ::ANativeActivity;
```
kimci86, line 46: Why is this needed? It probably shouldn't be in sf namespace.

src/SFML/Modules/Window.cppm
```diff
@@ -0,0 +1,114 @@
...
+        using sf::Touch::getPosition;
+    }
+    using sf::VideoMode;
+    namespace Vulkan {
+        using sf::Vulkan::isAvailable;
+        using sf::Vulkan::getFunction;
+        using sf::Vulkan::getGraphicsRequiredInstanceExtensions;
+    }
+    using ::VkInstance;
+    using ::VkSurfaceKHR;
```
kimci86, line 94: Same remark: this probably shouldn't be in sf namespace.

src/SFML/Modules/CMakeLists.txt
```diff
@@ -0,0 +1,82 @@
...
+        Network.cppm
+    )
+endif()
+
+# Define configure_cpp_module_target function if not already defined
+if(NOT COMMAND configure_cpp_module_target)
+    function(configure_cpp_module_target target)
+        target_sources(${target} PUBLIC FILE_SET CXX_MODULES FILES ${SFML_MODULES})
+    endfunction()
+endif()
```
kimci86, line 42: This function is used only once. Let's remplace its usage by its body and not define the function.

src/SFML/Modules/Config.cppm
```diff
@@ -0,0 +1,40 @@
...
+
+////////////////////////////////////////////////////////////
+// Module preamble
+////////////////////////////////////////////////////////////
+
+module;
+
+#include <SFML/Config.hpp>
+
+export module sfml.Config;
```
kimci86, line 33: I don't see the point of having a separate module for config. It can be part of sfml.system module.

src/SFML/Modules/Audio.cppm
```diff
@@ -0,0 +1,81 @@
...
+    using sf::SoundStream;
+}
+
+////////////////////////////////////////////////////////////
+/// \defgroup audio Audio module
+///
+/// Sounds, streaming (musics or custom sources), recording,
+/// spatialization.
+///
+////////////////////////////////////////////////////////////
```
kimci86, line 81: This is copied from Audio.hpp. It doesn't bring anything to the doxygen documentation. Let's remove those comments here and in other modules.

kimci86:
> Hi, are there anything missing changes from my recent update that you want addressed before you are ready to accept and merge the pull request?

I guess this should be automatically tested to some extent. Maybe having an example program using those modules would be enough.

mikomikotaishi:
> > Hi, are there anything missing changes from my recent update that you want addressed before you are ready to accept and merge the pull request?
> 
> I guess this should be automatically tested to some extent. Maybe having an example program using those modules would be enough.

Where would be an appropriate place for one?

src/SFML/Modules/Config.cppm
mikomikotaishi: These actually contain a bunch of macros, and macros shouldn't be exported from modules. I'm contemplating just deleting it altogether.

mikomikotaishi:
Okay, after a bit of testing I have found that GCC still cannot compile modules correctly (Clang works just fine). I'm leaving a message in CMake that will prevent it from trying to compile if the compiler is GCC.

ChrisThrasher:
The SFML team is interested in knowing what it will take to provide a dual header-based and module-based interface to the library. That being said, using C++20 modules requires raising the C++ requirement, CMake requirement, and compiler requirements of the library as of the current major version of SFML, version 3.

Your PR if merged into SFML 3 would require some complicated infrastructure to raise those requirements if and only if a C++20 modules build is occurring. On top of that the testing infrastructure (both unit tests and CI) would have to adapt to the raised requirement. This entails a level of additional complexity that I have a hard time picturing merging into SFML 3.

The good news is that SFML 4 is likely going to target a toolset (C++ standard/CMake/compiler versions) that will allow the use of C++20 modules so such a PR as yours would fit much more neatly into that major version. However, SFML 4 work has not begun and we cannot tell you when it may begin. If you're willing to be patient and slowly iterate this PR with us, it has a chance to land early in the SFML 4 development cycle. 

Whether you choose to keep working on this is up to you but you ought to know where our attitudes lie and the approximate timeline for a change as what you're proposing.

mikomikotaishi:
I understand, I was hoping to see it integrated to the next release (i.e. 3.1), but certainly I could wait until 4 for this to be added
> The SFML team is interested in knowing what it will take to provide a dual header-based and module-based interface to the library.
Could you elaborate a bit more on what you mean? 
> Most of the configurations we currently support are too old to compiler C++20 modules code.
I am a little bit confused - the code itself may be old but that doesn't stop users of C++20 and above from being able to import it as a module. See [this](https://github.com/mikomikotaishi/openJuice) project of mine, it can import/export SFML as modules (as of currently it does not, but I have tested importing the SFML modules and it still compiles fine).

mikomikotaishi:
I understand, I was hoping to see it integrated to the next release (i.e. 3.1), but certainly I could wait until 4 for this to be added
> The SFML team is interested in knowing what it will take to provide a dual header-based and module-based interface to the library.
Could you elaborate a bit more on what you mean? 
> Most of the configurations we currently support are too old to compiler C++20 modules code.
I am a little bit confused - the code itself may be old but that doesn't stop users of C++20 and above from being able to import it as a module.

mikomikotaishi:
Update on module names: @eXpl0it3r , you asked earlier about whether the modules should be named sfml.system or SFML.System, etc. It seems the C++  standard does prefer lowercase modules, see [here](https://isocpp.org/files/papers/P1634R0.html)

mikomikotaishi:
Also, for convenience, what do you think of also having a module named only sfml that re-exports all five of the modules, so that files requiring all five can just write import sfml; and get all five modules? Obviously this will not replace the existing five individual modules, it is only a convenience feature.

ChrisThrasher:
> Also, for convenience, what do you think of also having a module named only `sfml` that re-exports all five of the modules, so that files requiring all five can just write `import sfml;` and get all five modules? Obviously this will not replace the existing five individual modules, it is only a convenience feature (there would be six modules, `sfml` and the five `sfml.*` submodules).

Let's skip this idea for now, but I may want to come back to it later once we solve the bigger problems.

> > The SFML team is interested in knowing what it will take to provide a dual header-based and module-based interface to the library.
>
> Could you elaborate a bit more on what you mean?

SFML 3's interface comes in the form of header files. SFML 4 will also ship with header files. SFML 4 will not be a module-only library. Therefore SFML 4, if it is to provide a C++20 modules interface, must support both interfaces.

> > Most of the configurations we currently support are too old to compile C++20 modules code.
> 
> I am a little bit confused - the code itself may be old but that doesn't stop users of C++20 and above from being able to import it as a module. For instance, see my project [here](https://github.com/mikomikotaishi/openJuice), I have tested importing SFML which compiles fine and runs fine.

As library authors, we make promises like this to users: "SFML 3 can be compiled from source on Ubuntu 22 using its default compiler and packages available from `apt`". That means we're saying GCC 11, the default compiler for Ubuntu 22, must be able to compile SFML 3. We do let ourselves write any code that breaks with GCC 11. What you're asking to do is adding code to SFML 3 which breaks support for these compilers. This is no small ask and it's why I can't support a PR like this prior to SFML 4.

mikomikotaishi:
OK, fair enough. I just hope that not too much will change in the meantime so the existing code here can be merged seamlessly when the time comes for SFML 4. I just wonder why it cannot be used as an optional extension for the time being.

> SFML 3's interface comes in the form of header files. SFML 4 will also ship with header files. SFML 4 will not be a module-only library. Therefore SFML 4, if it is to provide a C++20 modules interface, must support both interfaces. In other words, we are not getting rid of our header files in SFML 4. A C++20 modules interface has to coexist with those header files.

The existing layout, with `.cppm` files that `#include` the headers and then re-export all contained definitions, will accomplish this exactly, and is probably the simplest and easiest to maintain solution.

Odex64:
> As library authors, we make promises like this to users: "SFML 3 can be compiled from source on Ubuntu 22 using its default compiler and packages available from apt". That means we're saying GCC 11, the default compiler for Ubuntu 22, must be able to compile SFML 3.

I find that a pretty stupid reason personally. I don't see nothing wrong with using newer compilers.
Oh well, I still wish the best for SFML and looking up modules support for future never nevertheless.

ChuanqiXu9:
@mikomikotaishi I just noticed you contributed several nice patches for C++20 modules and I am maintaining https://arewemodulesyet.org/ . It'll be appreciated if you'd like to update https://arewemodulesyet.org so that I don't have to chase your trace : )

mikomikotaishi:
> @mikomikotaishi I just noticed you contributed several nice patches for C++20 modules and I am maintaining https://arewemodulesyet.org/ . It'll be appreciated if you'd like to update https://arewemodulesyet.org so that I don't have to chase your trace : )

Hi @ChuanqiXu9, thank you for letting me know! I will make an effort to update it any time I make a new modules port

ChrisThrasher:
> I just wonder why it cannot be used as an optional extension for the time being.

I don't want SFML to have optional extensions that are untested and of unspecified quality. If a feature if merged, it needs to meet the same level of testing and quality as all other features. We're not going to just toss something into the repo and say "eh whatever you can use it if you want but no promises". It's a bad user experience if the library offers some good features that are well tested and others that are not well tested and may be broken.

> I just hope that not too much will change in the meantime so the existing code here can be merged seamlessly when the time comes for SFML 4.

You should expect a lot to change about SFML by the time we reconsider this PR. I appreciate your enthusiasm to keep this PR updated, but I wanted to let you know that this will not be merged in the 3.x release series. On a long enough timeline we'll eventually offer a C++20 modules interface but I cannot promise you when that will happen. Maybe that will be in SFML 4, but nothing can be guaranteed at this time.

mikomikotaishi:
I understand. I will continue to keep this PR updated in the meantime, even if it will be a while before SFML 4 enters development.

Ignis92:
@mikomikotaishi If it can be of any help, I used most of your files to quickly put together a wrapping static library for SFML that simply reinjects headers into their corresponding modules. You can look up the details in [this minimal repo](https://github.com/Ignis92/SFMLModules).

Regarding my environment, I'm using MSVC and I had some issues compiling, but that seemed to be more a general problem with the modules' ABI processing rather than something specifically related to SFML or my wrapper. Anyway, I was able to get it working only by changing to the preview version of VS2022, where what seems to be the fix has been sitting for almost 3 months now.

mikomikotaishi:
Hmm, interesting. Modules don't have a standardised intermediate form/ABI between compilers though that probably has nothing to do with it. I've been able to use these modules myself in some projects though so I wonder if something I did is specific to Clang, though module sources should be compiler-agnostic. Nonetheless I'm glad to hear there is some interest in using SFML modules.

mikomikotaishi:
I am interested in whether or not for SFML 4 method chaining and returning `*this` from methods will be considered for addition. I doubt it would break existing code (as calls to setters returning `void` would just be returning a pointer to the object that gets discarded).

mikomikotaishi:
I've decided to just make them partitions, rather than full modules. It's far simpler to just write `import sfml;` rather than complicate things.

mikomikotaishi:
Now that there is more discussion on the direction of SFML 4, can we re-open discussion on C++ modules integration?

Issue #3369, opened 2024-12-31
RenderWindow::create() with Qt's QWidget::winId() crashes program on Windows


Wolfmyths:
### Prerequisite Checklist

- [X] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [X] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [X] I am here to report an issue and not to just ask a question or look for help (use the [forum](https://en.sfml-dev.org/forums/index.php#c3) or [Discord](https://discord.gg/nr4X7Fh) instead)

### Describe your issue here

I am creating a program with Qt 6.8 and SFML, I started off with SFML 2.6 and I decided to upgrade to SFML 3.0.
Migrating gave me no issues other than a new bug that involves `sf::Window::create(sf::WindowHandle Handle)`.
The program becomes unresponsive only when I use this specific overload.

This appears to only happen on the Windows build, as Linux doesn't experience this issue.

I found a forum post from 2016 that talks about the same issue that appears to have returned:
https://en.sfml-dev.org/forums/index.php?topic=21049.msg150558#msg150558

Sadly there are no error messages that I know of, the program just becomes unresponsive.

### Your Environment

- OS / distro / window manager: Windows 10
- SFML version: 3.0.0
- Compiler / toolchain: MingGW
- Special compiler / CMake flags: 

### Steps to reproduce

1. Start Program

```cpp
#include <QApplication>
#include <QWidget>
#include <QObject>
#include <QScopedPointer>

#include <SFML/Graphics.hpp>

class GameDisplay : public QWidget
{
    Q_OBJECT
public:
    explicit GameDisplay(QWidget *parent = nullptr, QString rom_url = QString()) : QWidget{parent},
                                                                                    render_window(new sf::RenderWindow(sf::VideoMode({800, 600}), "OpenGL", sf::Style::Default)),
                                                                                    texture({800, 600}),
                                                                                    sprite(texture),
                                                                                    m_initialized(false)
    {
        // Window attributes
        setAttribute(Qt::WA_PaintOnScreen);
        setAttribute(Qt::WA_OpaquePaintEvent);
        setAttribute(Qt::WA_NoSystemBackground);
        setAttribute(Qt::WA_DeleteOnClose);

        setWindowFlags(Qt::Window);
        resize(800, 600);
    };
[...]

PR #3468, opened 2025-03-26
Added support for making OpenGL errors fatal

binary1248:
Since any detected OpenGL error is an indication that something went wrong somewhere, we should have support for strict error checking that `assert(false)`s when running a debug build. Since not everybody will be able to make sense of these errors, this strict mode is only enabled when explicitly requested via CMake option.

Since we have seen in #3369 that different driver implementations will have different interpretations of what to do when making OpenGL calls without an active context, this change also additionally checks if an OpenGL context was active on the current thread at the time of the function call. While the implementations that silently ignore this scenario are technically still within specification, I think we can all agree that making an OpenGL call without an active context is never done intentionally and thus constitutes programmer error.

The new CMake option `SFML_FATAL_OPENGL_ERRORS` will enable making OpenGL errors fatal. This includes our newly detected no-active-context errors. Until #3369 is resolved `SFML_FATAL_OPENGL_ERRORS` should not be enabled in CI since that might lead to CI never completing successfully. After #3369 is resolved we can add `SFML_FATAL_OPENGL_ERRORS` to the debug build options in CI as well to have the additional checks enabled.

Whoever wants to tackle #3369 can enable `SFML_FATAL_OPENGL_ERRORS` to make the error visible even when testing on implementations that silently ignore the error (AMD, Mesa3D, etc.).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
         - { name: Static, flags: -DBUILD_SHARED_LIBS=OFF }
         type:
         - { name: Release }
-        - { name: Debug, flags: -DCMAKE_BUILD_TYPE=Debug -DSFML_ENABLE_COVERAGE=ON }
+        - { name: Debug, flags: -DCMAKE_BUILD_TYPE=Debug -DSFML_ENABLE_COVERAGE=ON -DSFML_FATAL_OPENGL_ERRORS=ON }
 
         include:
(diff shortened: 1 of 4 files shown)


binary1248:
If #3469 gets merged before this `SFML_FATAL_OPENGL_ERRORS` can be activated in the CI configuration in this change as well.

binary1248:
Added `-DSFML_FATAL_OPENGL_ERRORS=ON` to CI configuration.

ChrisThrasher (approved):

Issue #2428, opened 2023-02-27
Windows Joystick Impl for Gamepad Trigger Separate Polling


ShadowMarker789:
# Change Windows Joystick Implementation to a modern API that allows the consuming programmer to query the two analog triggers of the gamepad device separately. 

The current sf::Joystick implementation uses Windows DirectInput v8 which has limitations and is deprecated. 

Below exerpt from [Microsoft Documentation](https://learn.microsoft.com/en-us/windows/win32/xinput/xinput-and-directinput)

> Note
> Use of legacy [DirectInput](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/ee416842(v=vs.85)) is not recommended, and DirectInput is not available for Windows Store apps. 
> ... 
> The Xbox Controller is properly enumerated on [DirectInput](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/ee416842(v=vs.85)), and can be used with the DirectInputAPIs. However, some functionality provided by XInput will be missing from the DirectInput implementation:
> The left and right trigger buttons will act as a single button, not independently
> The vibration effects will not be available
> Querying for headset devices will not be available
> The combination of the left and right triggers in [DirectInput](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/ee416842(v=vs.85)) is by design. Games have always assumed that DirectInput device axes are centered when there is no user interaction with the device. However, the Xbox controller was designed to register minimum value, not center, when the triggers are not being held. Older games would therefore assume user interaction.

Microsoft recommended moving to XInput from DirectInput a while ago, really pushing for Gamepad standardization on windows gaming machines, but it's also simultaneously reasonable to want to be able to support multiple different kinds of gaming controllers and not just XInput-capable ones. 

The newer gaming-input APIs in modern windows have newer APIs that let one do this. 

The most compelling candidate to me is [Windows.Gaming.Input::RawGameController](https://learn.microsoft.com/en-us/uwp/api/windows.gaming.input.rawgamecontroller?view=winrt-22621) 

Which itself has an API not too dissimilar to that of sf::Joystick 

> AxisCount
> The number of axes on the raw game controller.
> ButtonCount
> The number of buttons on the raw game controller.
> DisplayName
> The name of the controller, provided by the hardware.
> GetCurrentReading(Boolean[], GameControllerSwitchPosition[], Double[])
> Gets a snapshot of the raw game controller's state.

Being the ones that immediately caught my eye. 

## Limitations and negatives 

[...]

PR #3471, opened 2025-03-27
Win32 Xbox Controller Fix

ShadowMarker789:
## Description

This PR fixes issue #2428 by dual wielding RawInput and XInput, foregoing DirectInput entirely. 

## How to test this PR?

Test this PR on Windows by launching the Joystick demo with a controller and playing around, using the left and right triggers of an Xbox controller, in particular.

### DIFF
--- a/src/SFML/Window/WindowImpl.cpp
+++ b/src/SFML/Window/WindowImpl.cpp
@@ -297,7 +297,7 @@ void WindowImpl::processJoystickEvents()
 
         // Connection state
         const bool connected = m_joystickStatesImpl->states[i].connected;
-        if (previousState.connected ^ connected)
+        if (previousState.connected != connected)
         {
             if (connected)
                 pushEvent(Event::JoystickConnected{i});
@@ -335,7 +335,7 @@ void WindowImpl::processJoystickEvents()
                 const bool prevPressed = previousState.buttons[j];
                 const bool currPressed = m_joystickStatesImpl->states[i].buttons[j];
 
-                if (prevPressed ^ currPressed)
+                if (prevPressed != currPressed)
                 {
                     if (currPressed)
                         pushEvent(Event::JoystickButtonPressed{i, j});


Bambo-Borris:
I have a lot of peripherals kicking around for gaming that I can try with this PR for some testing, if I get any bugs/issues I'll do my best to debug far enough to give you some information to work on a fix. Glad to see someone take up the mantle on this, will keep you posted on how the testing goes 🫡

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -45,1114 +45,717 @@
...
-            }
+            // looks weird, but the BitSize tells us how many bits they're actually using and which ones to ignore 
+            LONG logicalMax = expectedMax & pValueCaps[i].LogicalMax;
+            LONG logicalMin = expectedMax & pValueCaps[i].LogicalMin;
 
-            // Enumerate device objects (axes/povs/buttons)
-            result = m_device->EnumObjects(&JoystickImpl::deviceObjectEnumerationCallback,
-                                           this,
-                                           DIDFT_AXIS | DIDFT_BUTTON | DIDFT_POV);
+            float min = (float)logicalMin;
```
Bambo-Borris, line 461: I don't think we want these C-style casts either

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -45,1114 +45,717 @@
...
+                    state.buttons[8]  = !!(gamepad.wButtons & XINPUT_GAMEPAD_START);
+                    state.buttons[9]  = !!(gamepad.wButtons & XINPUT_GAMEPAD_BACK);
+                    state.buttons[10] = !!(gamepad.wButtons & XINPUT_GAMEPAD_LEFT_SHOULDER);
+                    state.buttons[11] = !!(gamepad.wButtons & XINPUT_GAMEPAD_RIGHT_SHOULDER);
+                    state.buttons[12] = !!(gamepad.wButtons & XINPUT_GAMEPAD_LEFT_THUMB);
+                    state.buttons[13] = !!(gamepad.wButtons & XINPUT_GAMEPAD_RIGHT_THUMB);
+
+                    SHORT deadzone = XINPUT_GAMEPAD_LEFT_THUMB_DEADZONE / 4;
+
+                    state.axes[Joystick::Axis::X] = (abs(gamepad.sThumbLX) < deadzone) ? 0.0f : gamepad.sThumbLX / 327.670f;
```
Bambo-Borris, line 521: Magic number of `327.670` here, can we convert this to a const with some context as to what this value means in the context of the conversion being done here

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -45,1114 +45,717 @@
...
+                    state.axes[Joystick::Axis::Y] = (abs(gamepad.sThumbLY) < deadzone) ? 0.0f : gamepad.sThumbLY / 327.670f;
+                    state.axes[Joystick::Axis::Z] = (abs(gamepad.sThumbRX) < deadzone) ? 0.0f : gamepad.sThumbRX / 327.670f;
+                    state.axes[Joystick::Axis::R] = (abs(gamepad.sThumbRY) < deadzone) ? 0.0f : gamepad.sThumbRY / 327.670f;
+                    state.axes[Joystick::Axis::U] = gamepad.bLeftTrigger / 2.550f;
```
Bambo-Borris, line 525: Same here for `2.550`, what does this value mean in this context?

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -45,1114 +45,717 @@
...
 
-
 ////////////////////////////////////////////////////////////
-JoystickState JoystickImpl::updateDInputPolled()
+void JoystickImpl::initialize()
 {
-    JoystickState state;
-
-    if (m_device)
+    // we'll be churning through data routinely, so it's simply more efficient to allocate a chunk once and reuse it rather than mallocing over and over again.
```
Bambo-Borris, line 637: I'm going to have to revisit this or see what the stance of some other folks is on this. I am certain we didn't need `malloc` in this context. Others will come along with better suggestions here so I'll just leave this as a note for an area to dig into.

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -45,1114 +45,717 @@
...
+        {
+            jstkImpl.m_caps.axes[GetAxis(index)] = true;
+        }
     }
 
-    keyDataSize = sizeof(keyData);
+    unsigned int xInputIndex = 0;
+    for (size_t i = 0; i < m_sJoysticks.size(); i++)
+    {
+        auto& jstk = m_sJoysticks.at(i);
```
Bambo-Borris, line 169: Any specific reason for using `.at(i)` over `[]` operator on a `std::vector<>`? I'm not sure if there's a specific convention within the SFML codebase of when to use one over the other, or if to always prefer one.

Bambo-Borris (commented):
I've tried to get a quick overview of the flow of what's going on here, I need to spend more time digging into the documentation side of raw input since I think the best place for me on this review is in the use of the APIs. I've worked with XInput a bunch. I've left some questions and I'll move onto the testing of the code next.

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,716 @@
...
+    if (errno != 0 || *endptr != L'\0' || vidLong > 0xFFFF)
     {
-        JOYINFOEX joyInfo;
-        joyInfo.dwSize         = sizeof(joyInfo);
-        joyInfo.dwFlags        = 0;
-        ConnectionCache& cache = connectionCache[i];
-        cache.connected        = joyGetPosEx(JOYSTICKID1 + i, &joyInfo) == JOYERR_NOERROR;
-
-        cache.timer.restart();
+        const sf::String vidStrSfml = {vidStr};
```
ChrisThrasher, line 243:
```suggestion
        const sf::String vidStrSfml(vidStr);
```

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,716 @@
...
+    HRESULT hresult{};
+    UINT    bufferSize = sizeof(RAWINPUT);
+    UINT    uiResult{};
+    // a bigger stack-alloc, but since we spawned a separate thread for this we have plenty of stack memory
+    RAWINPUT  rawInput{};
+    HIDP_CAPS hCaps{};
+    DWORD     lastError = 0x0;
+
+    // It's nice being able to allocate an entire RAWINPUT on the stack, apparently RawInput has been around so long that folk used to malloc for it.
+    uiResult = GetRawInputData(inputDevice, RID_INPUT, &rawInput, &bufferSize, sizeof(RAWINPUTHEADER));
```
ChrisThrasher, line 291:
```suggestion
    UINT uiResult = GetRawInputData(inputDevice, RID_INPUT, &rawInput, &bufferSize, sizeof(RAWINPUTHEADER));
```
It's best to declare a variable at the same place you initialize it rather than declaring it beforehand.

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,716 @@
...
+            return;
+        }
 
-    // Search for a joystick with the given index in the connected list
-    for (const JoystickRecord& record : joystickList)
-    {
-        if (record.index == index)
+        std::uint16_t    capsLength = hCaps.NumberInputButtonCaps;
+        HIDP_BUTTON_CAPS hButtonCaps{};
+        if (FAILED(hresult = HidP_GetButtonCaps(HidP_Input, &hButtonCaps, &capsLength, preparsedData)))
```
ChrisThrasher, line 364:
```suggestion
        if (const auto result = HidP_GetButtonCaps(HidP_Input, &hButtonCaps, &capsLength, preparsedData); FAILED(result))
```
This ensures `result`'s scope is limited to this `if` statement instead of being visible to the larger scope. This applies to other places where `FAILED` is used so please check those as well.

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,716 @@
...
-            auto property              = DIPROPDWORD();
-            property.diph.dwSize       = sizeof(property);
-            property.diph.dwHeaderSize = sizeof(property.diph);
-            property.diph.dwHow        = DIPH_DEVICE;
+        if (buttonCount == 0)
+        {
+            // INFO: No buttons are down.
 
-            if (SUCCEEDED(m_device->GetProperty(DIPROP_VIDPID, &property.diph)))
+            for (std::size_t buttonIndex = 0; buttonIndex < Joystick::ButtonCount; ++buttonIndex)
```
ChrisThrasher, line 395: Can this be converted to a range-for loop?

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,716 @@
...
-
-            static bool         formatInitialized = false;
-            static DIDATAFORMAT format;
-
-            if (!formatInitialized)
+        }
+        else
+        {
+            std::size_t rawInputButtonIndex = 0;
+            for (std::size_t buttonIndex = 0; buttonIndex < Joystick::ButtonCount; ++buttonIndex)
```
ChrisThrasher, line 403: Can this be converted to a range-for loop?

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,716 @@
...
+                    state.buttons[8]  = !!(gamepad.wButtons & XINPUT_GAMEPAD_START);
+                    state.buttons[9]  = !!(gamepad.wButtons & XINPUT_GAMEPAD_BACK);
+                    state.buttons[10] = !!(gamepad.wButtons & XINPUT_GAMEPAD_LEFT_SHOULDER);
+                    state.buttons[11] = !!(gamepad.wButtons & XINPUT_GAMEPAD_RIGHT_SHOULDER);
+                    state.buttons[12] = !!(gamepad.wButtons & XINPUT_GAMEPAD_LEFT_THUMB);
+                    state.buttons[13] = !!(gamepad.wButtons & XINPUT_GAMEPAD_RIGHT_THUMB);
+
+                    const SHORT deadzone = XINPUT_GAMEPAD_LEFT_THUMB_DEADZONE / 4;
+
+                    state.axes[Joystick::Axis::X] = (abs(gamepad.sThumbLX) < deadzone) ? 0.0f : gamepad.sThumbLX / 327.670f;
```
ChrisThrasher, line 516:
```suggestion
                    state.axes[Joystick::Axis::X] = (std::abs(gamepad.sThumbLX) < deadzone) ? 0.0f : gamepad.sThumbLX / 327.670f;
```
This applies to other uses of `abs` as well

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,716 @@
...
 
 
 ////////////////////////////////////////////////////////////
-void JoystickImpl::closeDInput()
+DWORD WINAPI JoystickImpl::Win32JoystickDispatchThread(LPVOID /* lpParam */)
 {
-    if (m_device)
+    // Required
+    auto coInitResult = CoInitializeEx(nullptr, COINIT_APARTMENTTHREADED);
+    if (!SUCCEEDED(coInitResult))
```
ChrisThrasher, line 541:
```suggestion
    if (const auto coInitResult = CoInitializeEx(nullptr, COINIT_APARTMENTTHREADED); FAILED(coInitResult))
```

ChrisThrasher (changes requested):
I pushed some changes that fixed much of CI but there are still some warnings and clang-tidy failures that did not have an obvious solution. Most are related to messy implicit casting. Please fix those if you can.

src/SFML/Window/Win32/JoystickImpl.hpp
```diff
@@ -117,111 +102,38 @@ class JoystickImpl
...
-    [[nodiscard]] JoystickState updateDInputBuffered();
+private:
+    /// \brief informs system of device added
+    static void DispatchDeviceConnected(HANDLE deviceHandle);
+    /// \brief informs system of device removed
+    static void DispatchDeviceRemoved(HANDLE deviceHandle);
+    /// \brief Performs XInput Operations
+    static void DispatchXInput();
+    /// \brief Performs RawInput Operations
+    static void DispatchRawInput(HRAWINPUT inputDevice);
```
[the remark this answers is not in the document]
ShadowMarker789: But can we? These methods utilize private instance fields.

I'd have to make those fields public to move these static functions into the anonymous namespace wouldn't I? My initial attempt has me rethinking this with almost all of these methods complaining about inability to access private member data.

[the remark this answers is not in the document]
ShadowMarker789: Those that can be moved have been moved.

src/SFML/Window/Win32/JoystickImpl.hpp
```diff
@@ -30,12 +30,11 @@
 #include <SFML/System/EnumArray.hpp>
 #include <SFML/System/Win32/WindowsHeader.hpp>
 
-#include <dinput.h>
-#include <mmsystem.h>
-
+#include <vector>
```
ChrisThrasher, line 33: I don't think we need this inclusion anymore.

src/SFML/Window/Win32/JoystickImpl.hpp
```diff
@@ -117,111 +102,33 @@ class JoystickImpl
...
-    ///
-    ////////////////////////////////////////////////////////////
-    static BOOL CALLBACK deviceEnumerationCallback(const DIDEVICEINSTANCE* deviceInstance, void* userData);
+    /// \brief informs system of device added
+    static void dispatchDeviceConnected(HANDLE deviceHandle);
+    /// \brief informs system of device removed
+    static void dispatchDeviceRemoved(HANDLE deviceHandle);
+    /// \brief Performs XInput Operations
+    static void dispatchXInput();
+    /// \brief Performs RawInput Operations
```
ChrisThrasher, line 114: Eventually you'll need to add the same doxygen headers that all other functions have which means tweaking the formatting and adding docs for each parameter.

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,724 @@
...
-        return isConnectedDInput(index);
-
-    ConnectionCache&   cache                  = connectionCache[index];
-    constexpr sf::Time connectionRefreshDelay = sf::milliseconds(500);
-    if (!lazyUpdates && cache.timer.getElapsedTime() > connectionRefreshDelay)
+    // Find VID
+    std::size_t vidPos = devicePath.find(L"VID_");
+    if (vidPos == std::wstring::npos)
+        return false;
+    vidPos += 4;
```
ChrisThrasher, line 104: Where does this `4` come from? Can we replace some of these magic numbers with more expressive expressions or a named constant if that's the right solution? This goes for some other magic numbers in this file. It's hard for me to figure out what the code is trying to do or how I can change it without breaking an internal invariant.

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,724 @@
...
+                                   GENERIC_READ | GENERIC_WRITE,
+                                   FILE_SHARE_READ | FILE_SHARE_WRITE,
+                                   nullptr,
+                                   OPEN_EXISTING,
+                                   0x0,
+                                   nullptr);
+    if (HidD_GetProductString(fileHandle, deviceHumanNameDataChunk.data(), rawInputChunkSize))
     {
-        if (m_buffered)
+        joystickImpl.m_identification.name = String{reinterpret_cast<wchar_t*>(deviceHumanNameDataChunk.data())};
```
ChrisThrasher, line 176:
```suggestion
        joystickImpl.m_identification.name = String(reinterpret_cast<wchar_t*>(deviceHumanNameDataChunk.data()));
```

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,724 @@
...
+        USHORT vid{};
+        USHORT pid{};
+        bool   isXInput = false;
+        if (extractVidPid(devicePath, vid, pid, isXInput))
```
ChrisThrasher, line 180: Perhaps instead of this function returning a bool + using output parameter, we could put all the output values into a structure then return a `std::optional<SomeNewStruct>`. Then we can do something like this
```cpp
if (const auto foo = extractVidPid(devicePath))
{
    joystickImpl.m_identification.vendorId  = foo->vid;
    joystickImpl.m_identification.productId = foo->pid;
    joystickImpl.m_useXInput                = true;
}
```

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,724 @@
...
+        // RawInput query capabilities
+        UINT buffSize{};
+        if (const auto result = GetRawInputDeviceInfoW(deviceHandle, RIDI_PREPARSEDDATA, nullptr, &buffSize);
+            FAILED(result))
         {
-            const float angle                = static_cast<float>(pos.dwPOV) / 18000.f * 3.141592654f;
-            state.axes[Joystick::Axis::PovX] = std::sin(angle) * 100;
-            state.axes[Joystick::Axis::PovY] = std::cos(angle) * 100;
+            // that didn't work!
+            auto lastError = GetLastError();
```
ChrisThrasher, line 206:
```suggestion
            const auto lastError = GetLastError();
```

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,724 @@
...
+            return;
         }
-        else
+        if (const auto result = GetRawInputDeviceInfoW(deviceHandle, RIDI_PREPARSEDDATA, preparsedDataChunk.data(), &buffSize);
+            FAILED(result))
         {
-            state.axes[Joystick::Axis::PovX] = 0;
-            state.axes[Joystick::Axis::PovY] = 0;
+            // that didn't work!
+            auto lastError = GetLastError();
```
ChrisThrasher, line 215:
```suggestion
            const auto lastError = GetLastError();
```

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,724 @@
...
+            err() << "GetRawInputData returned [" << result << "] unexpectedly! GetLastError: [" << lastError << "]"
+                  << std::endl;
+            return;
+        }
+        auto*     preparsedData = reinterpret_cast<PHIDP_PREPARSED_DATA>(preparsedDataChunk.data());
+        HIDP_CAPS hCaps{};
+        if (const auto result = HidP_GetCaps(preparsedData, &hCaps); FAILED(result))
+        {
+            // that didn't work!
+            auto lastError = GetLastError();
```
ChrisThrasher, line 225:
```suggestion
            const auto lastError = GetLastError();
```

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,724 @@
...
+    {
+        // RawInput query capabilities
+        UINT buffSize{};
+        if (const auto result = GetRawInputDeviceInfoW(deviceHandle, RIDI_PREPARSEDDATA, nullptr, &buffSize);
+            FAILED(result))
         {
-            const float angle                = static_cast<float>(pos.dwPOV) / 18000.f * 3.141592654f;
-            state.axes[Joystick::Axis::PovX] = std::sin(angle) * 100;
-            state.axes[Joystick::Axis::PovY] = std::cos(angle) * 100;
+            // that didn't work!
```
ChrisThrasher, line 205:
```suggestion
```
I appreciate the comment but I think the `FAILED(result)` right above this makes it clear enough that this condition did not succeed.

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,724 @@
...
+                  << std::endl;
+            return;
         }
-        else
+        if (const auto result = GetRawInputDeviceInfoW(deviceHandle, RIDI_PREPARSEDDATA, preparsedDataChunk.data(), &buffSize);
+            FAILED(result))
         {
-            state.axes[Joystick::Axis::PovX] = 0;
-            state.axes[Joystick::Axis::PovY] = 0;
+            // that didn't work!
```
ChrisThrasher, line 214:
```suggestion
```

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,724 @@
...
+            auto lastError = GetLastError();
+            err() << "GetRawInputData returned [" << result << "] unexpectedly! GetLastError: [" << lastError << "]"
+                  << std::endl;
+            return;
+        }
+        auto*     preparsedData = reinterpret_cast<PHIDP_PREPARSED_DATA>(preparsedDataChunk.data());
+        HIDP_CAPS hCaps{};
+        if (const auto result = HidP_GetCaps(preparsedData, &hCaps); FAILED(result))
+        {
+            // that didn't work!
```
ChrisThrasher, line 224:
```suggestion
```

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,724 @@
...
+            err() << "HidP_GetCaps returned [" << result << "] unexpectedly! GetLastError: [" << lastError << "]"
+                  << std::endl;
+            return;
         }
 
-        // Buttons
-        for (unsigned int i = 0; i < Joystick::ButtonCount; ++i)
-            state.buttons[i] = (pos.dwButtons & (1u << i)) != 0;
+        joystickImpl.m_caps.buttonCount = hCaps.NumberInputButtonCaps;
+        auto axes                       = hCaps.NumberInputValueCaps;
```
ChrisThrasher, line 232:
```suggestion
        const auto axes                 = hCaps.NumberInputValueCaps;
```

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,724 @@
...
-////////////////////////////////////////////////////////////
-void JoystickImpl::cleanupDInput()
-{
-    // Release the DirectInput interface
-    if (directInput)
+    if (fileHandle)
     {
-        directInput->Release();
-        directInput = nullptr;
+        CloseHandle(fileHandle);
```
ChrisThrasher, line 256: I see a number of codepaths where the function exits without this being called. Perhaps we wrap `fileHandle` in a `std::unique_ptr` with a custom deleter so that this is guaranteed to be called when the function exits. 

See `sf::FileInputStream::m_file` and `sf::FileInputStream::FileCloser` as a good example of this pattern.

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,724 @@
...
+            // we're getting the information from another API, return.
+            return;
+        }
 
-    m_axes.fill(-1);
-    m_buttons.fill(-1);
+        if (const auto uiResult = GetRawInputDeviceInfoW(deviceHandle, RIDI_PREPARSEDDATA, nullptr, &bufferSize);
+            uiResult == UINT_MAX)
+        {
+            // that didn't work!
```
ChrisThrasher, line 328:
```suggestion
```

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,724 @@
...
+            auto lastError = GetLastError();
```
ChrisThrasher, line 329:
```suggestion
            const auto lastError = GetLastError();
```

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,724 @@
...
 
-    m_deviceCaps        = {};
-    m_deviceCaps.dwSize = sizeof(DIDEVCAPS);
-    m_state             = JoystickState();
-    m_buffered          = false;
+        if (const auto uiResult = GetRawInputDeviceInfoW(deviceHandle, RIDI_PREPARSEDDATA, preparsedDataChunk.data(), &bufferSize);
+            uiResult == UINT_MAX)
+        {
+            // that didn't work!
+            auto lastError = GetLastError();
```
ChrisThrasher, line 339:
```suggestion
            const auto lastError = GetLastError();
```

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,724 @@
...
-    {
-        if (record.index == index)
+        auto* preparsedData = reinterpret_cast<PHIDP_PREPARSED_DATA>(preparsedDataChunk.data());
+
+        if (const auto result = HidP_GetCaps(preparsedData, &hCaps); FAILED(result))
         {
-            // Create device
-            HRESULT result = directInput->CreateDevice(record.guid, &m_device, nullptr);
+            // that didn't work!
+            auto lastError = GetLastError();
```
ChrisThrasher, line 350:
```suggestion
            const auto lastError = GetLastError();
```

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -44,1115 +44,724 @@
...
 
-            if (FAILED(result))
-            {
-                err() << "Failed to create DirectInput device: " << result << std::endl;
+        std::uint16_t    capsLength = hCaps.NumberInputButtonCaps;
+        HIDP_BUTTON_CAPS hButtonCaps{};
+        if (const auto result = HidP_GetButtonCaps(HidP_Input, &hButtonCaps, &capsLength, preparsedData); FAILED(result))
+        {
+            // that didn't work!
+            auto lastError = GetLastError();
```
ChrisThrasher, line 361:
```suggestion
            const auto lastError = GetLastError();
```

ChrisThrasher (changes requested):

src/SFML/Window/Win32/JoystickImpl.cpp
ChrisThrasher: I went ahead and made this change but oddly `is`XInput` is never used. Is that a bug or should we remove that variable?

src/SFML/Window/Win32/JoystickImpl.cpp
ChrisThrasher: On second thought this is a bit tricker since `HANDLE` is a `void*` and you can't have `std::unique_ptr<void>`. I suppose we could use `std::shared_ptr<void>` but that's less appealing. It may still be necessary though given the multiple exit paths of this function.

ChrisThrasher:
https://github.com/SFML/SFML/actions/runs/14147972602/job/39637631720?pr=3471

MinGW is not ABI-compatible with the HID DLL we're using. Not sure how to fix that since compiling it ourselves seems well out of scope.

CMakePresets.json
```diff
@@ -11,7 +11,8 @@
         "SFML_BUILD_EXAMPLES": "ON",
         "SFML_BUILD_TEST_SUITE": "ON",
         "SFML_ENABLE_STDLIB_ASSERTIONS": "ON",
-        "SFML_WARNINGS_AS_ERRORS": "ON"
+        "SFML_WARNINGS_AS_ERRORS": "ON",
+        "CMAKE_BUILD_TYPE": "Debug"
```
learn-more, line 15: I don't think you should change this?

eXpl0it3r:
Superseded by #3477

PR #3476, opened 2025-03-30
Require two-step initialization of `sf::priv::ResourceStream`

### LINKED ISSUE #3473: Different Behavior when File Doesn't Exist with Android
The `ResourceStream` implementation for Android triggers in debug mode on an assert if a file doesn't exist, which means that the behavior is different when using `loadFromFile`/`openFromFile`/etc. on Android (which uses `ResourceStream` underneath), than it is on any other platform.

https://github.com/SFML/SFML/blob/1c803b9e174762146c0beead4badbe7501cf7e0d/src/SFML/System/Android/ResourceStream.cpp#L39-L45

Has been reported on Discord by Nafffen & @Zombieschannel (see original [discussion #1](https://discord.com/channels/175298431294636032/175298431294636032/1340350137465049120) & [discussion #2](https://discord.com/channels/175298431294636032/175298431294636032/1354835459628470513))

## Your Environment

- OS / distro / window manager: Android
- SFML version: SFML 3

## Steps to reproduce

1. Be in debug mode, i.e. have asserts enabled
2. Load any of the SFML resources where the file doesn't exist

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::Texture texture;
    texture.loadFromFile("nope.jpg"); // Assert
}
```

## Expected behavior

`loadFromFile` returns `false` and execution continues

## Actual behavior

`loadFromFile` asserts and execution stops

ChrisThrasher:
## Description

Closes #3473

The fix is basically to just make `ResourceStream` behave like `FileInputStream`. Frankly I'm not sure why we directly use `ResourceStream` when `FileInputStream` already uses `ResourceStream` under the hood, but regardless this PR should fix the issue without otherwise changing any behavior.

### DIFF
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -154,24 +154,21 @@ bool Font::openFromFile(const std::filesystem::path& filename)
     // Cleanup the previous resources
     cleanup();
 
-#ifndef SFML_SYSTEM_ANDROID
-
     // Create the input stream and open the file
+#ifndef SFML_SYSTEM_ANDROID
     const auto stream = std::make_shared<FileInputStream>();
     const auto type   = "file"sv;
+#else
+    const auto stream = std::make_shared<priv::ResourceStream>();
+    const auto type   = "Android resource stream"sv;
+#endif
+
     if (!stream->open(filename))
     {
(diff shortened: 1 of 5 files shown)


ChrisThrasher:
@Zombieschannel could you test this out?

eXpl0it3r (approved):

PR #3477, opened 2025-04-01
Win32 Xbox XInput Controller Fix (Revised)

ShadowMarker789:
## Description
This PR fixes issue https://github.com/SFML/SFML/issues/2428 (#2428) by adding XInput filtering and support in addition to the current DirectInput functionality.

## How to test this PR?
Test this PR on Windows by launching the Joystick demo with a controller and playing around, using the left and right triggers of an Xbox controller, in particular.

Other information: 

`WM_DEVICECHANGE` is inappropriate place to perform COM interop in order to determine if the Xbox controller is an XInput device or not. As such, the device-change procedure has been moved away from when the window message occurs and onto when the update routine is called. This was identified as a problem and advice was given to move the functionality away from `WM_DEVICECHANGE` in the DirectX Discord. 

No further linking targets were added as MinGW builds were failing on that - instead the procedures for XInput are loaded dynamically. `wbemuuid.lib` was also similarly avoided as a link target to avoid build errors with MinGW, instead hardcoding the relevant UUIDs for the IID and CLSID directly, following the conventions already present for the UUIDs involved in DirectInput.

### DIFF
--- a/src/SFML/Window/Win32/JoystickImpl.cpp
+++ b/src/SFML/Window/Win32/JoystickImpl.cpp
@@ -45,6 +45,22 @@
 
 #include <cmath>
 
+// Used for XInput matching
+#include <Xinput.h>
+#include <oleauto.h>
+#include <wbemidl.h>
+
+#ifndef SAFE_RELEASE
+#define SAFE_RELEASE(p)     \
+    {                       \
+        if (p)              \
+        {                   \
+            (p)->Release(); \
+            (p) = nullptr;  \
+        }                   \
+    }
+#endif
+
 
 ////////////////////////////////////////////////////////////
 // DirectInput
@@ -84,6 +100,7 @@ struct JoystickRecord
     GUID         guid{};
     unsigned int index{};
     bool         plugged{};
+    bool         xInputDevice{};
 };
 
 using JoystickList = std::vector<JoystickRecord>;
@@ -99,6 +116,218 @@ using JoystickBlacklist = std::vector<JoystickBlacklistEntry>;
 JoystickBlacklist joystickBlacklist;
 
 const DWORD directInputEventBufferSize = 32;
+} // namespace
+
+// for XInput
+namespace
+{
+constexpr std::size_t xinputMaxDevices{4};
+std::vector<bool>     xInputSlots{};
+bool                  directInputNeedsInvalidation{};
+
+[[nodiscard]] sf::Joystick::Axis getAxis(std::size_t index)
+{
+    switch (index)
+    {
+        case 0:
+            return sf::Joystick::Axis::X;
+        case 1:
+            return sf::Joystick::Axis::Y;
+        case 2:
+            return sf::Joystick::Axis::Z;
+        case 3:
+            return sf::Joystick::Axis::R;
+        case 4:
+            return sf::Joystick::Axis::U;
+        case 5:
+            return sf::Joystick::Axis::V;
+        case 6:
+            return sf::Joystick::Axis::PovX;
+        case 7:
+            return sf::Joystick::Axis::PovY;
+        default:
+            throw std::out_of_range("index is out of range.");
+    }
+}
+
+struct XInputCleanupData
+{
+    VARIANT                           var            = {};
+    IWbemLocator*                     pIWbemLocator  = nullptr;
+    IEnumWbemClassObject*             pEnumDevices   = nullptr;
+    std::array<IWbemClassObject*, 20> pDevices       = {};
+    IWbemServices*                    pIWbemServices = nullptr;
+    BSTR                              bstrNamespace  = nullptr;
+    BSTR                              bstrDeviceID   = nullptr;
+    BSTR                              bstrClassName  = nullptr;
+};
+
+// NOLINTBEGIN(readability-identifier-naming)
+// keeping GUIDs and UUIDs consistent
+// Define CLSID_WbemLocator
+const CLSID CLSID_WbemLocator = {0x4590f811, 0x1d3a, 0x11d0, {0x89, 0x1f, 0x00, 0xaa, 0x00, 0x4b, 0x2e, 0x24}};
+
+// Define IID_IWbemLocator
+const IID IID_IWbemLocator = {0xdc12a687, 0x737f, 0x11cf, {0x88, 0x4d, 0x00, 0xaa, 0x00, 0x4b, 0x2e, 0x24}};
+// NOLINTEND(readability-identifier-naming)
+
+// Function pointer type for XInputGetState
+using XInputGetState_t = DWORD(WINAPI*)(DWORD dwUserIndex, XINPUT_STATE* pState);
+
+XInputGetState_t mXInputGetState = nullptr;
+
+void safeCleanup(XInputCleanupData& data)
+{
+    VariantClear(&data.var);
+
+    if (data.bstrNamespace)
+        SysFreeString(data.bstrNamespace);
+    if (data.bstrDeviceID)
+        SysFreeString(data.bstrDeviceID);
+    if (data.bstrClassName)
+        SysFreeString(data.bstrClassName);
+
+    for (auto* device : data.pDevices)
+        SAFE_RELEASE(device);
+
+    SAFE_RELEASE(data.pEnumDevices);
+    SAFE_RELEASE(data.pIWbemLocator);
+    SAFE_RELEASE(data.pIWbemServices);
+}
+
+// See also https://learn.microsoft.com/en-us/windows/win32/xinput/xinput-and-directinput?redirectedfrom=MSDN
+[[nodiscard]] BOOL isXInputDevice(const GUID* pGuidProductFromDirectInput)
+{
+    XInputCleanupData data;
+    bool              bisXInputDevice = false;
+    HRESULT           hr              = {};
+
+    // So we can call VariantClear() later, even if we never had a successful IWbemClassObject::Get().
+    VARIANT var = {};
+
+    const HRESULT comInit = CoInitialize(nullptr);
+    if (FAILED(comInit))
+    {
+        safeCleanup(data);
+        return false;
+    }
+
+    VariantInit(&var);
+
+    // Create WMI
+    hr = CoCreateInstance(CLSID_WbemLocator,
+                          nullptr,
+                          CLSCTX_INPROC_SERVER,
+                          IID_IWbemLocator,
+                          reinterpret_cast<LPVOID*>(&data.pIWbemLocator));
+    if (FAILED(hr) || data.pIWbemLocator == nullptr)
+    {
+        safeCleanup(data);
+        return false;
+    }
+
+    data.bstrNamespace = SysAllocString(L"\\\\.\\root\\cimv2");
+    if (data.bstrNamespace == nullptr)
+    {
+        safeCleanup(data);
+        return false;
+    }
+    data.bstrClassName = SysAllocString(L"Win32_PNPEntity");
+    if (data.bstrClassName == nullptr)
+    {
+        safeCleanup(data);
+        return false;
+    }
+    data.bstrDeviceID = SysAllocString(L"DeviceID");
+    if (data.bstrDeviceID == nullptr)
+    {
+        safeCleanup(data);
+        return false;
+    }
+
+    // Connect to WMI
+    hr = data.pIWbemLocator
+             ->ConnectServer(data.bstrNamespace, nullptr, nullptr, nullptr, 0L, nullptr, nullptr, &data.pIWbemServices);
+    if (FAILED(hr) || data.pIWbemServices == nullptr)
+    {
+        safeCleanup(data);
+        return false;
+    }
+
+    // Switch security level to IMPERSONATE.
+    hr = CoSetProxyBlanket(data.pIWbemServices,
+                           RPC_C_AUTHN_WINNT,
+                           RPC_C_AUTHZ_NONE,
+                           nullptr,
+                           RPC_C_AUTHN_LEVEL_CALL,
+                           RPC_C_IMP_LEVEL_IMPERSONATE,
+                           nullptr,
+                           EOAC_NONE);
+    if (FAILED(hr))
+    {
+        safeCleanup(data);
+        return false;
+    }
+
+    hr = data.pIWbemServices->CreateInstanceEnum(data.bstrClassName, 0, nullptr, &data.pEnumDevices);
+    if (FAILED(hr) || data.pEnumDevices == nullptr)
+    {
+        safeCleanup(data);
+        return false;
+    }
+
+
+    // Loop over all devices
+    for (;;)
+    {
+        ULONG uReturned = 0;
+        hr = data.pEnumDevices->Next(10000, static_cast<ULONG>(data.pDevices.size()), data.pDevices.data(), &uReturned);
+        if (FAILED(hr))
+        {
+            safeCleanup(data);
+            return false;
+        }
+        if (uReturned == 0)
+            break;
+
+        for (size_t iDevice = 0; iDevice < uReturned; ++iDevice)
+        {
+            // For each device, get its device ID
+            hr = data.pDevices[iDevice]->Get(data.bstrDeviceID, 0L, &var, nullptr, nullptr);
+            if (SUCCEEDED(hr) && var.vt == VT_BSTR && var.bstrVal != nullptr)
+            {
+                // Check if the device ID contains "IG_".  If it does, then it's an XInput device
+                // This information cannot be found from DirectInput
+                if (wcsstr(var.bstrVal, L"IG_"))
+                {
+                    // If it does, then get the VID/PID from var.bstrVal
+                    DWORD  dwPid  = 0;
+                    DWORD  dwVid  = 0;
+                    WCHAR* strVid = wcsstr(var.bstrVal, L"VID_");
+                    if (strVid && swscanf_s(strVid, L"VID_%4X", &dwVid) != 1)
+                        dwVid = 0;
+                    WCHAR* strPid = wcsstr(var.bstrVal, L"PID_");
+                    if (strPid && swscanf_s(strPid, L"PID_%4X", &dwPid) != 1)
+                        dwPid = 0;
+
+                    // Compare the VID/PID to the DInput device
+                    const auto dwVidPid = static_cast<DWORD>(MAKELONG(dwVid, dwPid));
+                    if (dwVidPid == pGuidProductFromDirectInput->Data1)
+                    {
+                        bisXInputDevice = true;
+                        safeCleanup(data);
+                        return bisXInputDevice;
+                    }
+                }
+            }
+            VariantClear(&var);
+            SAFE_RELEASE(data.pDevices[iDevice]);
+        }
+    }
+
+    safeCleanup(data);
+    return bisXInputDevice;
+}
 } // namespace
 
 
@@ -206,6 +435,18 @@ namespace sf::priv
 ////////////////////////////////////////////////////////////
 void JoystickImpl::initialize()
 {
+    auto* xinputModule = LoadLibraryA("XInput1_4.dll");
+    if (!xinputModule)
+    {
+        // this always succeeds.
+        xinputModule = LoadLibraryA("XINPUT9_1_0.DLL");
+    }
+    assert(xinputModule != nullptr);
+    mXInputGetState = reinterpret_cast<XInputGetState_t>(
+        reinterpret_cast<void*>(GetProcAddress(xinputModule, "XInputGetState")));
+
+    xInputSlots.resize(xinputMaxDevices);
+
     // Try to initialize DirectInput
     initializeDInput();
 
@@ -228,6 +469,12 @@ void JoystickImpl::cleanup()
 ////////////////////////////////////////////////////////////
 bool JoystickImpl::isConnected(unsigned int index)
 {
+    if (directInputNeedsInvalidation)
+    {
+        directInputNeedsInvalidation = false;
+        updateConnections();
+    }
+
     if (directInput)
         return isConnectedDInput(index);
 
@@ -249,6 +496,11 @@ bool JoystickImpl::isConnected(unsigned int index)
 void JoystickImpl::setLazyUpdates(bool status)
 {
     lazyUpdates = status;
+}
+
+void JoystickImpl::invalidateDevices()
+{
+    directInputNeedsInvalidation = true;
 }
 
 ////////////////////////////////////////////////////////////
@@ -305,6 +557,16 @@ void JoystickImpl::close()
 ////////////////////////////////////////////////////////////
 JoystickCaps JoystickImpl::getCapabilities() const
 {
+    if (m_useXInput)
+    {
+        // XInput has 10 Buttons (since we exclude the DPad) and all 8 Axes
+        JoystickCaps caps{0};
+        caps.buttonCount    = 10;
+        constexpr auto axes = 8;
+        for (unsigned int i = 0; i < axes; ++i)
+            caps.axes[getAxis(i)] = true;
+        return caps;
+    }
     if (directInput)
         return getCapabilitiesDInput();
 
@@ -335,6 +597,17 @@ Joystick::Identification JoystickImpl::getIdentification() const
 ////////////////////////////////////////////////////////////
 JoystickState JoystickImpl::update()
 {
+    if (directInputNeedsInvalidation)
+    {
+        directInputNeedsInvalidation = false;
+        updateConnections();
+    }
+
+    if (m_useXInput)
+    {
+        return updateXInput();
+    }
+
     if (directInput)
     {
         if (m_buffered)
@@ -530,6 +803,7 @@ bool JoystickImpl::openDInput(unsigned int index)
     {
         if (record.index == index)
         {
+            m_useXInput = record.xInputDevice;
             // Create device
             HRESULT result = directInput->CreateDevice(record.guid, &m_device, nullptr);
 
@@ -578,6 +852,12 @@ bool JoystickImpl::openDInput(unsigned int index)
 
             if (SUCCEEDED(m_device->GetProperty(DIPROP_PRODUCTNAME, &stringProperty.diph)))
                 m_identification.name = stringProperty.wsz;
+
+            if (m_useXInput)
+            {
+                // use XInput instead of DirectInput for obtaining caps and data
+                return true;
+            }
 
             static bool         formatInitialized = false;
             static DIDATAFORMAT format;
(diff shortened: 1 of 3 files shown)


src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -861,6 +1137,115 @@ JoystickCaps JoystickImpl::getCapabilitiesDInput() const
...
+    state.buttons[0] = !!(gamepad.wButtons & XINPUT_GAMEPAD_A);
+    state.buttons[1] = !!(gamepad.wButtons & XINPUT_GAMEPAD_B);
+    state.buttons[2] = !!(gamepad.wButtons & XINPUT_GAMEPAD_X);
+    state.buttons[3] = !!(gamepad.wButtons & XINPUT_GAMEPAD_Y);
+    state.buttons[4] = !!(gamepad.wButtons & XINPUT_GAMEPAD_START);
+    state.buttons[5] = !!(gamepad.wButtons & XINPUT_GAMEPAD_BACK);
+    state.buttons[6] = !!(gamepad.wButtons & XINPUT_GAMEPAD_LEFT_SHOULDER);
+    state.buttons[7] = !!(gamepad.wButtons & XINPUT_GAMEPAD_RIGHT_SHOULDER);
+    state.buttons[8] = !!(gamepad.wButtons & XINPUT_GAMEPAD_LEFT_THUMB);
+    state.buttons[9] = !!(gamepad.wButtons & XINPUT_GAMEPAD_RIGHT_THUMB);
```
ZXShady, line 1183: I prefer if we compared against `0`

`0 !=  (gamepad.wButtons & XINPUT_GAMEPAD_A);`

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -101,6 +117,215 @@ JoystickBlacklist joystickBlacklist;
...
+    if (FAILED(comInit))
+    {
+        safeCleanup(data);
+        return false;
+    }
+
+    VariantInit(&data.var);
+
+    // Create WMI
+    hr = CoCreateInstance(CLSID_WbemLocator,
```
ZXShady, line 214: Prefer to directly initialize instead of declarations at top and put variables near their first use.
```suggestion
    bool              bisXInputDevice = false;

    const HRESULT comInit = CoInitialize(nullptr);
    XInputCleanupData data;
    if (FAILED(comInit))
    {
        safeCleanup(data);
        return false;
    }

    VariantInit(&data.var);

    // Create WMI
    HRESULT hr = CoCreateInstance(CLSID_WbemLocator,
```

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -101,6 +117,215 @@ JoystickBlacklist joystickBlacklist;
...
+    SafeRelease(data.pEnumDevices);
+    SafeRelease(data.pIWbemLocator);
+    SafeRelease(data.pIWbemServices);
+}
+
+// See also https://learn.microsoft.com/en-us/windows/win32/xinput/xinput-and-directinput?redirectedfrom=MSDN
+[[nodiscard]] BOOL isXInputDevice(const GUID* pGuidProductFromDirectInput)
+{
+    XInputCleanupData data;
+    bool              bisXInputDevice = false;
```
ZXShady, line 201: avoid C style Hungarian naming the `is` clearly tells us it is a boolean.


```suggestion
    bool              isXInputDevice = false;
```

ZXShady (commented):
Style changes

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -101,6 +117,215 @@ JoystickBlacklist joystickBlacklist;
...
+        case 6:
+            return sf::Joystick::Axis::PovX;
+        case 7:
+            return sf::Joystick::Axis::PovY;
+        default:
+            throw std::out_of_range("index is out of range.");
+    }
+}
+
+struct XInputCleanupData
```
learn-more, line 152: If you add a destructor in this struct, and have that do what `safeCleanup` is doing,
it will be done automatically on all paths leaving the function where it is used.

src/SFML/Window/Win32/JoystickImpl.cpp
ShadowMarker789: Suggestion taken and implemented.

src/SFML/Window/Win32/JoystickImpl.cpp
ShadowMarker789: This has been done, take a look and hopefully it matches what you want.

src/SFML/Window/Win32/JoystickImpl.cpp
ShadowMarker789: Good idea, that variable name was a holdover from the MSFT example that I copied over. Implemented.

src/SFML/Window/Win32/JoystickImpl.cpp
ShadowMarker789: Great suggestion, taken and implemented. (Can you check that I did it right?)

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -101,6 +106,211 @@ JoystickBlacklist joystickBlacklist;
...
+
+template <typename T>
+void safeRelease(T*& p)
+{
+    if (p)
+    {
+        p->Release();
+        p = nullptr;
+    }
+}
```
ZXShady, line 124: this should no longer be needed

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -101,6 +106,211 @@ JoystickBlacklist joystickBlacklist;
...
+        if (bstrClassName)
+            SysFreeString(bstrClassName);
+
+        for (auto* device : pDevices)
+            safeRelease(device);
+
+        safeRelease(pEnumDevices);
+        safeRelease(pIWbemLocator);
+        safeRelease(pIWbemServices);
+    }
```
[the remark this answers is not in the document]
ShadowMarker789: Okay, simplified cleanup code.

src/SFML/Window/Win32/JoystickImpl.cpp
ShadowMarker789: Oh I see, we can just do the same thing inline. Done.

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -101,6 +106,208 @@ JoystickBlacklist joystickBlacklist;
...
+
+// Function pointer type for XInputGetState
+using XInputGetState_t = DWORD(WINAPI*)(DWORD dwUserIndex, XINPUT_STATE* pState);
+
+XInputGetState_t mXInputGetState = nullptr;
+
+// See also https://learn.microsoft.com/en-us/windows/win32/xinput/xinput-and-directinput?redirectedfrom=MSDN
+[[nodiscard]] BOOL isXInputDevice(const GUID* pGuidProductFromDirectInput)
+{
+    const HRESULT comInit = CoInitialize(nullptr);
```
ZXShady, line 194: shouldn't there be a mirrored call to `CoUninitialize`?

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -236,6 +236,24 @@ m_cursorGrabbed(m_fullscreen)
         {
             JoystickImpl::setLazyUpdates(true);
 
+            std::array<RAWINPUTDEVICE, 3> rids{};
+            ZeroMemory(rids.data(), sizeof(rids));
```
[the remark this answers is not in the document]
ShadowMarker789: Was a holdover from example Microsoft code that didn't use std::array
Taken and implemented.

src/SFML/Window/Win32/JoystickImpl.cpp
ShadowMarker789: I wasn't so sure, but after testing and adding that to the destructor of the cleanup data everything continues to work. 

There doesn't HAVE to be a matching call to CoUninitialize because we never initialize COM to be MTA on this thread, but I've put it there. CoInitialize with the same apartment type (nullptr defaults to STA) is a no-op.

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -101,6 +106,212 @@ JoystickBlacklist joystickBlacklist;
...
+                {
+                    // If it does, then get the VID/PID from var.bstrVal
+                    DWORD  dwPid  = 0;
+                    DWORD  dwVid  = 0;
+                    WCHAR* strVid = wcsstr(data.var.bstrVal, L"VID_");
+                    if (strVid && swscanf_s(strVid, L"VID_%4X", &dwVid) != 1)
+                        dwVid = 0;
+                    WCHAR* strPid = wcsstr(data.var.bstrVal, L"PID_");
+                    if (strPid && swscanf_s(strPid, L"PID_%4X", &dwPid) != 1)
+                        dwPid = 0;
```
learn-more, line 295: This seems a weird way to retrieve a VID and PID from a device?

ShadowMarker789: > This seems a weird way to retrieve a VID and PID from a device?

That's how it's done in the Microsoft Documentation I linked above.

learn-more: looks like msdn suggests this, nvm.

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -101,6 +106,212 @@ JoystickBlacklist joystickBlacklist;
...
+            throw std::out_of_range("index is out of range.");
+    }
+}
+
+struct XInputCleanupData
+{
+    VARIANT                           var            = {};
+    IWbemLocator*                     pIWbemLocator  = nullptr;
+    IEnumWbemClassObject*             pEnumDevices   = nullptr;
+    std::array<IWbemClassObject*, 20> pDevices       = {};
```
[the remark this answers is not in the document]
ShadowMarker789: > If you make this an std::vector<IWbemClassObject*

I don't think that really gains us anything here. 

https://learn.microsoft.com/en-us/windows/win32/api/wbemcli/nf-wbemcli-ienumwbemclassobject-next

Right now, we're enumerating devices in batches of up to 20, hence passing in the size of the array as well as the pointer to the array data. 

I do not see what being able to resize the backing array does for us here. We're enumerating devices in a batch size - does having a variable batch size instead of a hard-coded reasonable 20 value gain us anything here? Would we want to change that? The MSFT example-code used 20. 20 works in everyone's testing so far. 

Are you sure this is doing what you think it's doing?

ChrisThrasher:
I rebased on `master` and made some style fixes.

texus:
I did a quick test and found that there are still some issues with this PR.

First the good news: the PR does improve the trigger identification. I ran the joystick example and checked which values changed when interacting with the controller:
```
With DirectInput controller:
  Left joystick:     X and Y
  Right joystick:    Z and R
  Triggers:          identified as buttons

Before PR with Xinput controller:
  Left joystick:     X and Y
  Right joystick:    U and V
  Triggers:          Z (0 = neither pressed, -100 = left trigger fully pressed, 100 = right trigger fully pressed)

After PR with Xinput controller:
  Left joystick:     X and Y
  Right joystick:    Z and R
  Triggers:          U and V (-100 = not pressed, 100 = fully pressed)
```

Is it intentional that the trigger values go from -100 to 100 by the way? Wouldn't the range from 0 to 100 make more sense?

I did find several issues in the few minutes that I tried the new code (which didn't occur with the version in the master branch):
- With 2 joysticks connected, the name of the joystick being used is wrong (Xbox controller identifies as F310 while F310 identifies as Xbox controller).
- Switching the F310 controller from Xintput to DirectInput mode (which is equivalent of removing and inserting a different controller) causes the program to show "\<Not connected\>" and controller actions do nothing until the program is restarted. Unplugging and replugging the F310 controller actually leads to a crash in that case in the XInputCleanupData destructor. The old version handled the switching and unplugging without issues.

I'll probably have a deeper look into these issues later if I find the time for it.

ShadowMarker789:
> Unplugging and replugging the F310 controller actually leads to a crash 

Particularly concerning, which makes me wonder if there was something messed up about the initialization order of the controllers. I don't have the hardware to test multiple xinput and non-xinput controllers simultaneously, all I had was one xinput controller and one PS5 controller, which means that I've only ever had one controller only in slot 1 in my personal testing. 

I will need some help from testers to identify and fix the issues with XInput here, but I might need to rethink my approach...

ShadowMarker789:
> Is it intentional that the trigger values go from -100 to 100 by the way? Wouldn't the range from 0 to 100 make more sense?

Yes - When I tested SFML DirectInput with a PS5 controller (wired) the Triggers of the PS5 controller were -100 to 100 

The consensus seems to be to try and avoid vendor-specific behaviours of controllers to make end-consumers' life easier - no need to write code specific to one vendor or another. 

It was first this reason also that the DPad was mapped to POVX and POVY - the PS5 controller did that, and consistency makes for an easier to use library.

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -861,6 +1081,100 @@ JoystickCaps JoystickImpl::getCapabilitiesDInput() const
...
+            {
+                xInputSlots[i] = true;
+                m_xInputIndex  = static_cast<DWORD>(i);
+                break;
+            }
+        }
+    }
+
+    XINPUT_STATE xinputState{};
+    const auto   result = mXInputGetState(m_xInputIndex, &xinputState);
```
learn-more, line 1103: You seem to be choosing your own index here, but it is actually XInput that assigns this ID.
You can't just query on a random ID and expect that to be the actual controller you are working with.

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -861,6 +1081,100 @@ JoystickCaps JoystickImpl::getCapabilitiesDInput() const
...
+    state.buttons[4] = 0 != (gamepad.wButtons & XINPUT_GAMEPAD_START);
+    state.buttons[5] = 0 != (gamepad.wButtons & XINPUT_GAMEPAD_BACK);
+    state.buttons[6] = 0 != (gamepad.wButtons & XINPUT_GAMEPAD_LEFT_SHOULDER);
+    state.buttons[7] = 0 != (gamepad.wButtons & XINPUT_GAMEPAD_RIGHT_SHOULDER);
+    state.buttons[8] = 0 != (gamepad.wButtons & XINPUT_GAMEPAD_LEFT_THUMB);
+    state.buttons[9] = 0 != (gamepad.wButtons & XINPUT_GAMEPAD_RIGHT_THUMB);
+
+    // The standard deadzone felt, too... dead. This feels reasonable, but should it be configurable?
+    // Threshold is ... different, and NOT a deadzone, but instead a required delta between inputs to event an update.
+    static constexpr SHORT deadzone = XINPUT_GAMEPAD_LEFT_THUMB_DEADZONE / 2;
```
learn-more, line 1131: Please note that there is also `XINPUT_GAMEPAD_RIGHT_THUMB_DEADZONE`

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -861,6 +1081,100 @@ JoystickCaps JoystickImpl::getCapabilitiesDInput() const
...
+    else if (dpadUp)
+        state.axes[Joystick::Axis::PovY] = 100.0f;
+    else if (dpadDown)
+        state.axes[Joystick::Axis::PovY] = -100.0f;
+    else
+        state.axes[Joystick::Axis::PovY] = 0.0f;
+
+    // XInput triggers range between 0 and 255 - to scale them to -100.0f .. 100.0f divide by the below factor (255 / 200)
+    static constexpr auto triggerScaleFactor = 1.275f;
+    state.axes[Joystick::Axis::U]            = (gamepad.bLeftTrigger / triggerScaleFactor) - 100.0f;
```
learn-more, line 1170: There is also the `XINPUT_GAMEPAD_TRIGGER_THRESHOLD`

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -99,6 +104,166 @@ using JoystickBlacklist = std::vector<JoystickBlacklistEntry>;
...
+                }
+            }
+
+            VariantClear(&data.var);
+            auto* device = data.pDevices[iDevice];
+            if (device)
+            {
+                device->Release();
+                // important to prevent a double-free in cleanup data destructor
+                device = nullptr;
```
learn-more, line 261: This is incorrect, it should zero out `data.pDevices[iDevice]`.
Now you are just clearing your local variable `device`, but that does not update the actual pDevices.

ChrisThrasher: Calling a release function then setting a pointer to null makes me think this could be a good place to use `std::unique_ptr` with a custom deleter.

src/SFML/Window/Win32/JoystickImpl.cpp
learn-more: The XInput sample from the DirectX SDK uses another dead zone:
https://github.com/microsoft/DirectX-SDK-Samples/blob/main/C%2B%2B/XInput/SimpleController/SimpleController.cpp#L39
which translates to `7864`
That is slightly bigger than the `7849` from `XINPUT_GAMEPAD_LEFT_THUMB_DEADZONE`

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -99,6 +104,166 @@ using JoystickBlacklist = std::vector<JoystickBlacklistEntry>;
...
+const IID IID_IWbemLocator = {0xdc12a687, 0x737f, 0x11cf, {0x88, 0x4d, 0x00, 0xaa, 0x00, 0x4b, 0x2e, 0x24}};
+// NOLINTEND(readability-identifier-naming)
+
+// Function pointer type for XInputGetState
+using XInputGetStateFunc = DWORD(WINAPI*)(DWORD dwUserIndex, XINPUT_STATE* pState);
+
+XInputGetStateFunc mXInputGetState = nullptr;
+
+// See also https://learn.microsoft.com/en-us/windows/win32/xinput/xinput-and-directinput?redirectedfrom=MSDN
+[[nodiscard]] BOOL isXInputDevice(const GUID* pGuidProductFromDirectInput)
```
learn-more, line 167: The DirectInput SDK sample caches this data instead of querying it for every device, that might also make sense?
https://github.com/walbourn/directx-sdk-samples/blob/main/DirectInput/Joystick/joystick.cpp#L288

learn-more: And this stackoverflow post has another way of getting the vid/pid from the XInput index:
https://stackoverflow.com/a/68879988/4928207

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -302,9 +496,19 @@ void JoystickImpl::close()
...
+    {
+        // XInput has 10 Buttons (since we exclude the DPad) and all 8 Axes
+        JoystickCaps caps{0};
+        caps.buttonCount = 10;
+        caps.axes.fill(true);
+        return caps;
+    }
+
     if (directInput)
         return getCapabilitiesDInput();
```
learn-more, line 513: Maybe `DInput` should be the fallback, and the entire `joyXXX` legacy api can be dropped?
@ChrisThrasher what do you think about that?

src/SFML/Window/Win32/JoystickImpl.cpp
ShadowMarker789: Yes, it seems that I was naive and tried to preempt which slot a newly added device would occupy. 

I have pushed further changes, please test them when you have the time.

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -99,6 +105,181 @@ using JoystickBlacklist = std::vector<JoystickBlacklistEntry>;
...
 
 const DWORD directInputEventBufferSize = 32;
+
+// For XInput
+
+struct XInputJoystickEntry
+{
+    XInputJoystickEntry()
+    {
+    }
```
ZXShady, line 115: is this needed? the default initializer for members is enough

src/SFML/Window/Win32/JoystickImpl.cpp
ShadowMarker789: Yes, but I dislike using undocumented APIs unless there's an exceptionally compelling usecase for it. 

In this case, it isn't compelling enough.

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -99,6 +105,211 @@ using JoystickBlacklist = std::vector<JoystickBlacklistEntry>;
...
+
+[[nodiscard]] std::optional<DWORD> getXInputIndexFromVidPid(WORD vid, WORD pid) noexcept
+{
+    DWORD slot = 0;
+    for (const DWORD xInputSlot : {0, 1, 2, 3})
+    {
+        XINPUT_CAPABILITIES_EX capsEx = {};
+        if (mXInputGetCapabilitiesEx(1, xInputSlot, 0, &capsEx) == ERROR_SUCCESS)
+        {
+            if (capsEx.vendorId == vid && capsEx.productId == pid)
```
learn-more, line 203:
```suggestion
            if (capsEx.vendorId == 0x045e && capsEx.productId == 0 && capsEx.Capabilities.Flags & XINPUT_CAPS_WIRELESS)
            {
                capsEx.productId = 0x02a1;
            }

            if (capsEx.vendorId == vid && capsEx.productId == pid)
```

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -579,6 +869,12 @@ bool JoystickImpl::openDInput(unsigned int index)
             if (SUCCEEDED(m_device->GetProperty(DIPROP_PRODUCTNAME, &stringProperty.diph)))
                 m_identification.name = stringProperty.wsz;
 
+            if (m_useXInput)
+            {
+                // use XInput instead of DirectInput for obtaining caps and data
```
ChrisThrasher: Please capitalize your comments

src/SFML/Window/Win32/WindowImplWin32.cpp
```diff
@@ -150,6 +150,25 @@ WindowImplWin32::WindowImplWin32(WindowHandle handle) : m_handle(handle)
...
+            {
+                // We are using this, talking to "generic" input devices.
+                rid.usUsagePage = 0x01; // HID_USAGE_PAGE_GENERIC;
+                // We want Windows to notify us when devices are added or removed.
+                rid.dwFlags = RIDEV_DEVNOTIFY;
+                // And send notifications to THIS specific HWND, which we registered earlier.
+                rid.hwndTarget = m_handle;
+            }
+
+            rids[0].usUsage = 0x05; // HID_USAGE_GENERIC_GAMEPAD;
```
ChrisThrasher, line 165: Why rewrite the value of the macro as opposed to just using the macro?

ChrisThrasher (changes requested):

src/SFML/Window/Win32/JoystickImpl.cpp
ShadowMarker789: You know what? Never mind! I think you're right, and I'm going to have to use this to get the right description for each device we find on XInput, because otherwise we have no idea which device is which.

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -99,14 +110,224 @@ using JoystickBlacklist = std::vector<JoystickBlacklistEntry>;
...
+// Function pointer type for XInputGetState
+using XInputGetStateFunc = DWORD(WINAPI*)(DWORD dwUserIndex, XINPUT_STATE* pState);
+// Function pointer type for XInputGetCapabilitiesEx
+using XInputGetCapabilitiesExFunc =
+    DWORD(WINAPI*)(DWORD a1, DWORD dwUserIndex, DWORD dwFlags, XINPUT_CAPABILITIES_EX* pCapabilities);
+
+XInputGetCapabilitiesExFunc xInputGetCapabilitiesEx = nullptr;
+XInputGetStateFunc          xInputGetState          = nullptr;
+
+[[nodiscard]] std::optional<DWORD> getXInputIndexFromVidPid(WORD vid, WORD pid) noexcept
```
learn-more, line 204: This is still an estimate, not an actual 'get input id'

src/SFML/Window/Win32/JoystickImpl.cpp
```diff
@@ -99,6 +108,219 @@ using JoystickBlacklist = std::vector<JoystickBlacklistEntry>;
...
+    DWORD                   xinputIndex{0xFFFFFFFF}; // cannot be zero because 0 is a valid index!
+    sf::priv::JoystickImpl* joystick{};
+    XINPUT_STATE            state{};
+    unsigned int            joystickIndex{};
+    WORD                    vendorId;
+    WORD                    productId;
+};
+// NOLINTBEGIN(readability-identifier-naming)
+// Since this struct is a reference to MSFT Windows
+struct XINPUT_CAPABILITIES_EX
```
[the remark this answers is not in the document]
ShadowMarker789: This struct is "foreign" to SFML, but it's not defined in a Windows header. It's used for the undocumented function `XInputGetCapabilitiesEx` Which takes in an a `XINPUT_CAPABILITIES_EX` which we have to define ourselves since there's no windows header for it. 

Using this lets us guess "less wrong" about which XInput device is likely to be a given DirectInput device - because now we can check that the VID and PID match, so we're at least talking about the same kind of device.

src/SFML/Window/Win32/WindowImplWin32.cpp
ShadowMarker789: Quite right, I was expecting that to cause linking problems like with the UUIDs, but it seems that the macros evaluate find without linking to HID directly! 

This works fine, and I'm now using the macro. 👍

ShadowMarker789: Nope! Looks like MinGW is unhappy with that macro. 
```
D:/a/SFML/SFML/src/SFML/Window/Win32/WindowImplWin32.cpp
D:/a/SFML/SFML/src/SFML/Window/Win32/WindowImplWin32.cpp: In constructor 'sf::priv::WindowImplWin32::WindowImplWin32(sf::WindowHandle)':
D:/a/SFML/SFML/src/SFML/Window/Win32/WindowImplWin32.cpp:171:31: error: 'HID_USAGE_GENERIC_MULTI_AXIS_CONTROLLER' was not declared in this scope
  171 |             rids[2].usUsage = HID_USAGE_GENERIC_MULTI_AXIS_CONTROLLER; // 0x08
      |                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
D:/a/SFML/SFML/src/SFML/Window/Win32/WindowImplWin32.cpp: In constructor 'sf::priv::WindowImplWin32::WindowImplWin32(sf::VideoMode, const sf::String&, uint32_t, sf::State, const sf::ContextSettings&)':
D:/a/SFML/SFML/src/SFML/Window/Win32/WindowImplWin32.cpp:276:31: error: 'HID_USAGE_GENERIC_MULTI_AXIS_CONTROLLER' was not declared in this scope
  276 |             rids[2].usUsage = HID_USAGE_GENERIC_MULTI_AXIS_CONTROLLER; // 0x08
      |                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
How should I proceed from here? Should I use a `#ifndef` to define it?

PR #3487, opened 2025-04-15
Move joystick example code around to support multiple controllers

learn-more:
Keep a state per joystick, so that we can display all of them at once. The window is simply resized for the number of controllers, so it might get fairly wide.

## Tasks

-   [ ] Tested on Linux
-   [x] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

Connect multiple controllers, and enjoy!

> [!NOTE]
> Currently (as of 15-apr-2025) devices that are enabled before starting the application will only be 'discovered' after receiving inputs from them.
> This is being worked on in #3477 

Example screenshots:
No controller:
No controller

1 controller:
1 controller

3 controllers:
3 controllers

### DIFF
--- a/examples/joystick/Joystick.cpp
+++ b/examples/joystick/Joystick.cpp
@@ -16,68 +16,157 @@ struct JoystickObject
 {
     sf::Text label;
     sf::Text value;
+
+    void draw(sf::RenderWindow& window, const sf::RenderStates& states = sf::RenderStates::Default) const
+    {
+        window.draw(label, states);
+        window.draw(value, states);
+    }
 };
 
-using Texts = std::unordered_map<std::string, JoystickObject>;
-Texts              texts;
-std::ostringstream sstr;
-float              threshold = 0.1f;
+[[nodiscard]] auto getStream()
+{
+    std::ostringstream sstr;
+    // Set up our string conversion parameters
+    sstr.precision(2);
+    sstr.setf(std::ios::fixed | std::ios::boolalpha);
+    return sstr;
+}
 
 // Axes labels in as strings
 const std::array<std::string, 8> axislabels = {"X", "Y", "Z", "R", "U", "V", "PovX", "PovY"};
 
-// Helper to set text entries to a specified value
-template <typename T>
-void set(const std::string& label, const T& value)
+class JoystickDisplay
 {
-    sstr.str("");
-    sstr << value;
-    texts.at(label).value.setString(sstr.str());
-}
+public:
+    JoystickDisplay() = default;
 
-// Update joystick identification
-void updateIdentification(unsigned int index)
-{
-    sstr.str("");
-    sstr << "Joystick " << index << ":";
-    auto& [label, value] = texts.at("ID");
-    label.setString(sstr.str());
-    value.setString(sf::Joystick::getIdentification(index).name);
-}
-
-// Update joystick axes
-void updateAxes(unsigned int index)
-{
-    for (unsigned int j = 0; j < sf::Joystick::AxisCount; ++j)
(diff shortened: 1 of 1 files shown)


examples/joystick/Joystick.cpp
```diff
@@ -16,145 +16,191 @@ struct JoystickObject
...
+        window.draw(value, states);
+    }
 };
 
 using Texts = std::unordered_map<std::string, JoystickObject>;
-Texts              texts;
-std::ostringstream sstr;
-float              threshold = 0.1f;
+
+auto getStrm()
```
ChrisThrasher, line 29: We prefer identifiers that use non-abbreviated works. Could you use a more verbose function name?

examples/joystick/Joystick.cpp
```diff
@@ -16,145 +16,191 @@ struct JoystickObject
...
-template <typename T>
-void set(const std::string& label, const T& value)
+struct JoystickDisplay
 {
-    sstr.str("");
-    sstr << value;
-    texts.at(label).value.setString(sstr.str());
-}
+    unsigned int index   = 0;
+    bool         present = false;
```
ChrisThrasher, line 44:
```suggestion
    unsigned int index{};
    bool         present{};
```

examples/joystick/Joystick.cpp
```diff
@@ -16,145 +16,191 @@ struct JoystickObject
...
-    auto& [label, value] = texts.at("ID");
-    label.setString(sstr.str());
-    value.setString(sf::Joystick::getIdentification(index).name);
-}
 
-// Update joystick axes
-void updateAxes(unsigned int index)
-{
-    for (unsigned int j = 0; j < sf::Joystick::AxisCount; ++j)
+    void init(unsigned int idx, const sf::Font& font)
```
ChrisThrasher, line 48: Why an `init` function instead of a constructor?

ChrisThrasher (changes requested):

examples/joystick/Joystick.cpp
learn-more: Added 2 constructors.

examples/joystick/Joystick.cpp
```diff
@@ -16,145 +16,190 @@ struct JoystickObject
...
-    label.setString(sstr.str());
-    value.setString(sf::Joystick::getIdentification(index).name);
-}
+    JoystickDisplay() = default;
 
-// Update joystick axes
-void updateAxes(unsigned int index)
-{
-    for (unsigned int j = 0; j < sf::Joystick::AxisCount; ++j)
+    JoystickDisplay(unsigned int idx, const sf::Font& font)
```
vittorioromeo, line 49:
```suggestion
    JoystickDisplay(unsigned int idx, const sf::Font& font) : index{idx}
```

examples/joystick/Joystick.cpp
```diff
@@ -96,64 +185,23 @@ int main()
...
-        {
-            updateValues(i);
-            break;
-        }
+        joysticks[i] = JoystickDisplay{i, font};
+        joysticks[i].updateValues();
+        if (joysticks[i].isPresent())
+            ++joysticksPresent;
     }
+    window.setSize({400 * std::max(joysticksPresent, 1u), 775});
```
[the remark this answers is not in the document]
learn-more: No, because it needs to scale up and down depending on the number of controllers.
This was a quick way to be able to show multiple controllers without re-designing the entire app.

I could make two constants for it if you prefer that?

ChrisThrasher:

ChrisThrasher (approved):

PR #3489, opened 2025-04-18
Reduce warnings inside Windows Vm

Lorenzooone:
## Description

This PR fixes some minor issues when it comes to using SFML on Windows inside of a VM.
In particular, it reduces the amount of spam that is output to the console.

These are the two changes in this PR:
- Avoids trying to disable SRGB in a RenderTexture if the SRGB extension is not available. This was causing an OpenGL EnumError beforehand, which would get spammed to the console.
- It makes it so the warning about the GDI Windows Generic driver is only printed once. Before, it would get constantly spammed.

## Tasks

-   [ ] Tested on Linux
-   [X] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

Just launch any SFML program from inside a Windows VM without GPU passthrough and check how much the console gets spammed by messages.

It should be significantly less. Meaning you're going to be able to better understand what is happening with your program.

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        {
                window.close();
        }

        window.clear();
        window.display();
    }
}
```

### DIFF
--- a/src/SFML/Window/Win32/WglContext.cpp
+++ b/src/SFML/Window/Win32/WglContext.cpp
@@ -510,13 +510,17 @@ void WglContext::updateSettingsFromPixelFormat()
     {
         m_isGeneric = true;
 
-        err() << "Warning: Detected \"Microsoft Corporation GDI Generic\" OpenGL implementation" << std::endl;
+        [[maybe_unused]] static const bool hasWarnedAboutGeneric = [actualFormat]() -> bool
+        {
+            err() << "Warning: Detected \"Microsoft Corporation GDI Generic\" OpenGL implementation" << std::endl;
 
-        // Detect if the generic GDI implementation is not accelerated
-        if (!(actualFormat.dwFlags & PFD_GENERIC_ACCELERATED))
-            err() << "Warning: The \"Microsoft Corporation GDI Generic\" OpenGL implementation is not "
-                     "hardware-accelerated"
-                  << std::endl;
+            // Detect if the generic GDI implementation is not accelerated
+            if (!(actualFormat.dwFlags & PFD_GENERIC_ACCELERATED))
+                err() << "Warning: The \"Microsoft Corporation GDI Generic\" OpenGL implementation is not "
+                         "hardware-accelerated"
+                      << std::endl;
+            return true;
+        }();
     }
 
     if (SF_GLAD_WGL_ARB_pixel_format)


Bambo-Borris:
I personally don't see value in this, VMs cause lots of problems since they aren't a physical host. Most users aren't trying SFML in a VM, and I think keeping the code simpler and ensuring it reports problems clearly is preferable for users running on physical machines. I think realistically VM users may just have to 'live with it'.

Lorenzooone:
> I personally don't see value in this, VMs cause lots of problems since they aren't a physical host. Most users aren't trying SFML in a VM, and I think keeping the code simpler and ensuring it reports problems clearly is preferable for users running on physical machines. I think realistically VM users may just have to 'live with it'.

I see your point, however look at it this way:
- Change number 1 fixes an issue in the way SFML called a function.
- Change number 2 is simply about not spamming the console while reporting an error.

The PR is VM centric, as that is how I discovered these issues.
However, they also happen on hardware in Windows which doesn't have a proper graphics driver (like a Raspberry Pi 5 running Windows for ARM).

Considering I (and many others) do not have one of the new Snapdragon Laptops, VMs and hardware without a proper graphics driver are the only ways to test my code on Windows for ARM.

Finally, these are not changes which ruins the experience of regular users in favor of VM users. The experience of VM users (which, let's face it, is actually going to be developers 99% of the time) will still be worse. But at least you're going to be able to look at the output of your console to debug stuff in a controlled environment.

binary1248:
I wouldn't mind these changes for 2 reasons: correctness and consistency.

The `GL_FRAMEBUFFER_SRGB` enumeration should only be used when the corresponding [`ARB_framebuffer_sRGB`]([ARB_framebuffer_sRGB](https://registry.khronos.org/OpenGL/extensions/ARB/ARB_framebuffer_sRGB.txt)) is actually available. Because the `glEnable` path is guarded behind `isSrgb()` and can only be run if sRGB support is available that only leaves the `glDisable` path that currently doesn't have a check. The correct extension to check for here would be `ARB_framebuffer_sRGB` and not `EXT_texture_sRGB`. I added all available extensions in #3491 so this should be based on that once it merges.

Regarding the warning message output, this is not the only place we output warnings when we detect a sub-optimal environment. In other places we already make use of a flag to prevent spamming the same message over and over again so doing the same here would just bring it in line with what we are already doing elsewhere.

Lorenzooone:
Updated (and tested) the PR to the new identifier.
After that, I also changed to "static local variable initialization with IIFE" after a suggestion about it on SFML's Discord.

binary1248:
We add a corresponding section to `GLExtensions.hpp` when we want to access tokens from a new extension. This way we can use a single `GLEXT_` token for both OpenGL and OpenGL ES builds without having to add `#ifdef`s everywhere else in the code.

Lorenzooone:
> We add a corresponding section to `GLExtensions.hpp` when we want to access tokens from a new extension. This way we can use a single `GLEXT_` token for both OpenGL and OpenGL ES builds without having to add `#ifdef`s everywhere else in the code.

A bit of an issue: OpenGL ES doesn't seem to have GL_ARB_framebuffer_sRGB. The ability to glEnable or glDisable sRGB is related to EXT_sRGB_write_control instead (https://registry.khronos.org/OpenGL/extensions/EXT/EXT_sRGB_write_control.txt), which doesn't seem to be in the updated header though...?

Should I make two, GLEXT_ARB_framebuffer_sRGB and GLEXT_sRGB_write_control and place them in logical OR for the glDisable function to be called?
Shouldn't GLEXT_sRGB_write_control also be checked when calling glEnable?

I am not an expert when it comes to this, to be clear. Just something I noticed while looking into how to implement the change to GLExtensions.hpp.

Lorenzooone:
Changes implemented and tested.
Let me know if there is anything.

FYI: From what I've seen while searching online, with OpenGL ES, only if a texture is SRGB, it be SRGB (automatically)...

Though if we wanted to mantain the pre-PR behaviour for OpenGL ES, GLEXT_ARB_framebuffer_sRGB should be set to true... :thinking:

binary1248:
We strip the extension type from the definition. `GLEXT_ARB_framebuffer_sRGB` should be `GLEXT_framebuffer_sRGB`.

Lorenzooone:
> We strip the extension type from the definition. `GLEXT_ARB_framebuffer_sRGB` should be `GLEXT_framebuffer_sRGB`.

I did notice that, however, I've also seen that there is an EXT_framebuffer_sRGB which is different from ARB_framebuffer_SRGB: https://registry.khronos.org/OpenGL/extensions/EXT/EXT_framebuffer_sRGB.txt https://registry.khronos.org/OpenGL/extensions/ARB/ARB_framebuffer_sRGB.txt

How should I handle this?

binary1248:
The `ARB` variants are more or less "official" extensions that get added to later standards. They are advertised by more implementations than their `EXT` predecessors. If both exist we always favour the `ARB` variant over `EXT`.

Lorenzooone:
I see... Should I check for both inside GLEXT? Or just for the ARB one?

binary1248:
Just `ARB`. There are systems where only `ARB` is advertised. Checking for both would cause sRGB to not be detected if `EXT` is not advertised but `ARB` is.

Lorenzooone:
Great! Then that is taken care of by the last push I did.

binary1248 (approved):

PR #3363, opened 2024-12-26
Reduce pointer usage

ChrisThrasher:
## Description

Continuing what I started in https://github.com/SFML/SFML/pull/3046

PR #3497, opened 2025-04-29
Feature: Android joystick support

nerudaj:
## Description

This PR adds the support for physical gamepads on Android devices through sf::Joystick API. The implementation is limited to only one physical device with layout of a standard Xbox/Playstation gamepad.

## Tasks

-   [ ] Tested on Linux
-   [ ] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [X] Tested on Android
- [x] Tested with Backbone One
- [x] Tested with íPega-9167
- [ ] Tested with Xbox One (wired) controller
- [ ] Fix implementation not noticing disconnected controller
- [ ] Make sure axii bindings are alinged with Windows behavior

## How to test this PR?

Download the attached zipfile and unpack it somewhere.

1) In the root dir, create `_build` directory and run command `cmake ..` in it. This will generate AppManifest, gradle config and local properties in the `bin-android` folder.
2) Edit `bin-android/cppcore/CMakeLists.txt` so the `add_subdirectory` call on line 5 points to cloned repo from this branch (or just FetchContent on this branch)
3) Open `bin-android` as a project in Android Studio and build/run the project.

[_ControllerTest.zip](https://github.com/user-attachments/files/19966191/_ControllerTest.zip)

The application will print identification for all 4 controller slots (only the first one is supported) and all axii/button values for the first controller. If you connect a controller, the app should immediately recognize it. If you press buttons / sticks, you should see immediate changes to displayed values.

I've tested this with Backbone One, íPega-9167

### DIFF
--- a/src/SFML/System/Android/Activity.hpp
+++ b/src/SFML/System/Android/Activity.hpp
@@ -81,8 +81,8 @@ struct ActivityStates
     Vector2i                                           mousePosition;
     EnumArray<Mouse::Button, bool, Mouse::ButtonCount> isButtonPressed{};
 
-    std::map<uint32_t, EnumArray<Joystick::Axis, float, Joystick::AxisCount>> joyAxii;
-    std::map<uint32_t, std::array<bool, Joystick::ButtonCount>> isJoystickButtonPressed{};
+    EnumArray<Joystick::Axis, float, Joystick::AxisCount> joyAxii;
+    std::array<bool, Joystick::ButtonCount> isJoystickButtonPressed{};
 
     bool mainOver{};
 
--- a/src/SFML/System/Android/Activity.hpp
+++ b/src/SFML/System/Android/Activity.hpp
@@ -29,6 +29,7 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Window/EglContext.hpp>
 #include <SFML/Window/Event.hpp>
+#include <SFML/Window/Android/JoystickButton.hpp>
 
 #include <SFML/System/EnumArray.hpp>
 
@@ -40,6 +41,7 @@
 #include <string>
 #include <unordered_map>
 #include <vector>
+#include <map>
 
 #include <cstddef>
 
@@ -78,6 +80,9 @@ struct ActivityStates
     std::unordered_map<int, Vector2i>                  touchEvents;
     Vector2i                                           mousePosition;
     EnumArray<Mouse::Button, bool, Mouse::ButtonCount> isButtonPressed{};
+
+    std::map<uint32_t, EnumArray<Joystick::Axis, float, Joystick::AxisCount>> joyAxii;
+    std::map<uint32_t, std::array<bool, Joystick::ButtonCount>> isJoystickButtonPressed{};
 
     bool mainOver{};
 
--- a/src/SFML/Window/Android/JoystickButton.hpp
+++ b/src/SFML/Window/Android/JoystickButton.hpp
@@ -1,0 +1,56 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2013 Jonathan De Wachter (dewachter.jonathan@gmail.com)
+//
+// This software is provided 'as-is', without any express or implied warranty.
+// In no event will the authors be held liable for any damages arising from the use of this software.
+//
+// Permission is granted to anyone to use this software for any purpose,
+// including commercial applications, and to alter it and redistribute it freely,
+// subject to the following restrictions:
+//
+// 1. The origin of this software must not be misrepresented;
+//    you must not claim that you wrote the original software.
+//    If you use this software in a product, an acknowledgment
+//    in the product documentation would be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such,
+//    and must not be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
+////////////////////////////////////////////////////////////
+
+#pragma once
+
+namespace sf::Joystick
+{
+////////////////////////////////////////////////////////////
+/// \brief Abstract names for Joystick button codes
+///
+////////////////////////////////////////////////////////////
+enum class Button : size_t
+{
+    A = 0,
+    B,
+    C,
+    X,
+    Y,
+    Z,
+    L1,
+    R1,
+    L2,
+    R2,
+    L3,
+    R3,
+    Start,
+    Select,
+    Capture,
+    DpadUp,
+    DpadDown,
+    DpadLeft,
+    DpadRight,
+    DpadCenter,
+};
+}
--- a/src/SFML/Window/Android/JoystickImpl.cpp
+++ b/src/SFML/Window/Android/JoystickImpl.cpp
@@ -29,16 +29,166 @@
 #include <SFML/System/Err.hpp>
 #include <SFML/System/Android/Activity.hpp>
 
-static std::string javaStringToStd(JNIEnv* lJNIEnv, jstring str)
+static std::string javaStringToStd(JNIEnv* env, jstring str)
 {
-    const char* utfChars = lJNIEnv->GetStringUTFChars(str, nullptr);
+    const char* utfChars = env->GetStringUTFChars(str, nullptr);
     std::string result(utfChars);
-    lJNIEnv->ReleaseStringUTFChars(str, utfChars);
+    env->ReleaseStringUTFChars(str, utfChars);
     return result;
 }
 
+template<class T>
+class JniArray {
+public:
+    JniArray(JNIEnv* env, jintArray array)
+        : env(env)
+        , array(array)
+        , length(env->GetArrayLength(array))
+        , data(env->GetIntArrayElements(array, nullptr))
+    {}
+
+    JniArray(JniArray&& other)
+    {
+        std::swap(env, other.env);
+        std::swap(array, other.array);
+        std::swap(length, other.length);
+        std::swap(data, other.data);
+    }
+
+    ~JniArray()
+    {
+        if (data)
+            env->ReleaseIntArrayElements(array, data, 0);
+    }
+
+public:
+    T operator[](ssize_t idx)
+    {
+        assert(0 <= idx && idx <= length);
+        return data[idx];
+    }
+
+    const T operator[](ssize_t idx) const
+    {
+        assert(0 <= idx && idx <= length);
+        return data[idx];
+    }
+
+    ssize_t getLength() const noexcept
+    {
+        return length;
+    }
+
+private:
+    JNIEnv * env = nullptr;
+    jintArray array = nullptr;
+    ssize_t length = 0;
+    T* data = nullptr;
+};
+
 namespace sf::priv
 {
+class JniInputDevice
+{
+public:
+    JniInputDevice(
+        JNIEnv* env,
+        jobject inputDevice,
+        jmethodID getNameMethod,
+        jmethodID getVendorIdMethod,
+        jmethodID getProductIdMethod,
+        jmethodID supportsSourceMethod)
+        : env(env)
+        , inputDevice(inputDevice)
+        , getNameMethod(getNameMethod)
+        , getVendorIdMethod(getVendorIdMethod)
+        , getProductIdMethod(getProductIdMethod)
+        , supportsSourceMethod(supportsSourceMethod)
+    {}
+
+public:
+    static std::optional<JniArray<jint>> getDeviceIds(JNIEnv *env, jclass inputDeviceClass)
+    {
+        jmethodID getDeviceIdsMethod = env->GetStaticMethodID(inputDeviceClass, "getDeviceIds", "()[I");
+        if (!getDeviceIdsMethod)
+        {
+            err() << "Could not locate InputDevice.getDeviceIds method" << std::endl;
+            return std::nullopt;
+        }
+
+        jintArray deviceIdsArray = (jintArray)env->CallStaticObjectMethod(inputDeviceClass, getDeviceIdsMethod);
+        if (deviceIdsArray == nullptr)
+        {
+            err() << "No input devices found." << std::endl;
+            return std::nullopt;
+        }
+
+        return JniArray<jint>(env, deviceIdsArray);
+    }
+
+    static std::optional<JniInputDevice> getDevice(JNIEnv *env, jclass inputDeviceClass, jint idx)
+    {
+        jmethodID getDeviceMethod = env->GetStaticMethodID(inputDeviceClass, "getDevice", "(I)Landroid/view/InputDevice;");
+        jmethodID getNameMethod = env->GetMethodID(inputDeviceClass, "getName", "()Ljava/lang/String;");
+        jmethodID getVendorIdMethod = env->GetMethodID(inputDeviceClass, "getVendorId", "()I");
+        jmethodID getProductIdMethod = env->GetMethodID(inputDeviceClass, "getProductId", "()I");
+        jmethodID supportsSourceMethod = env->GetMethodID(inputDeviceClass, "supportsSource", "(I)Z");
+
+        if (!getDeviceMethod || !getNameMethod || !getVendorIdMethod || !getProductIdMethod || !supportsSourceMethod)
+        {
+            err() << "Could not locate required InputDevice methods" << std::endl;
+            return std::nullopt;
+        }
+
+        jobject inputDevice = env->CallStaticObjectMethod(inputDeviceClass, getDeviceMethod, idx);
+        if (!inputDevice)
+        {
+            // Can happen normally, no log needed
+            return std::nullopt;
+        }
+
+        return JniInputDevice(
+                env,
+                inputDevice,
+                getNameMethod,
+                getVendorIdMethod,
+                getProductIdMethod,
+                supportsSourceMethod);
+    }
+
+    unsigned getVendorId() const
+    {
+        return (unsigned)env->CallIntMethod(inputDevice, getVendorIdMethod);
+    }
+
+    unsigned getProductId() const
+    {
+        return (unsigned)env->CallIntMethod(inputDevice, getProductIdMethod);
+    }
+
+    std::string getName() const
+    {
+        return javaStringToStd(
+            env,
+            (jstring)env->CallObjectMethod(
+                inputDevice,
+                getNameMethod));
+    }
+
+    bool supportsSource(size_t sourceFlags) const
+    {
+        return env->CallBooleanMethod(inputDevice, supportsSourceMethod, jint(sourceFlags));
+    }
+
+private:
+    JNIEnv* env;
+    jobject inputDevice;
+    jmethodID getNameMethod;
+    jmethodID getVendorIdMethod;
+    jmethodID getProductIdMethod;
+    jmethodID supportsSourceMethod;
+};
+
 ////////////////////////////////////////////////////////////
 void JoystickImpl::initialize()
 {
@@ -54,38 +204,18 @@ void JoystickImpl::cleanup()
 
 
 ////////////////////////////////////////////////////////////
-bool JoystickImpl::isConnected(unsigned int /* index */)
+bool JoystickImpl::isConnected(unsigned int index)
 {
-    // To implement
-    return true;
+    // This is called as a prefilter before open, but it
+    // would just duplicate the logic of open. Furthermore
+    // how many physical devices are you going to attach to
+    // an android device?
+    return index == 0;
 }
 
 
 ////////////////////////////////////////////////////////////
-bool JoystickImpl::open(unsigned int /* index */)
-{
-    // To implement
-    return true;
-}
-
-
-////////////////////////////////////////////////////////////
-void JoystickImpl::close()
-{
-    // To implement
-}
-
-
-////////////////////////////////////////////////////////////
-JoystickCaps JoystickImpl::getCapabilities() const
-{
-    // To implement
-    return {};
-}
-
-
-////////////////////////////////////////////////////////////
-Joystick::Identification JoystickImpl::getIdentification() const
+bool JoystickImpl::open(unsigned int joy_index)
 {
     // Retrieve activity states
     ActivityStates&       states = getActivity();
@@ -106,55 +236,66 @@ Joystick::Identification JoystickImpl::getIdentification() const
     lResult = lJavaVM->AttachCurrentThread(&lJNIEnv, &lJavaVMAttachArgs);
 
     if (lResult == JNI_ERR)
-        err() << "Failed to initialize JNI, couldn't get the Unicode value" << std::endl;
+        err() << "Failed to initialize JNI" << std::endl;
 
     jclass inputDeviceClass = lJNIEnv->FindClass("android/view/InputDevice");
     if (inputDeviceClass == nullptr) {
         err() << "Failed to find InputDevice class." << std::endl;
-        return m_identification;
+        return false;
     }
 
-    // Locate required InputDevice methods
-    jmethodID getDeviceIdsMethod = lJNIEnv->GetStaticMethodID(inputDeviceClass, "getDeviceIds", "()[I");
-    jmethodID getDeviceMethod = lJNIEnv->GetStaticMethodID(inputDeviceClass, "getDevice", "(I)Landroid/view/InputDevice;");
-    jmethodID getNameMethod = lJNIEnv->GetMethodID(inputDeviceClass, "getName", "()Ljava/lang/String;");
-    jmethodID getVendorIdMethod = lJNIEnv->GetMethodID(inputDeviceClass, "getVendorId", "()I");
-    jmethodID getProductIdMethod = lJNIEnv->GetMethodID(inputDeviceClass, "getProductId", "()I");
-    jmethodID supportsSourceMethod = lJNIEnv->GetMethodID(inputDeviceClass, "supportsSource", "(I)Z");
+    auto deviceIds = JniInputDevice::getDeviceIds(lJNIEnv, inputDeviceClass);
+    if (!deviceIds) return false;
 
-    if (!getDeviceIdsMethod || !getDeviceMethod || !getNameMethod || !getVendorIdMethod || !getProductIdMethod || !supportsSourceMethod) {
-        err() << "Failed to locate required methods from android/view/InputDevice" << std::endl;
-        return m_identification;
-    }
-
-    // Call getDeviceIds() to get the array of device IDs
-    jintArray deviceIdsArray = (jintArray)lJNIEnv->CallStaticObjectMethod(inputDeviceClass, getDeviceIdsMethod);
-    if (deviceIdsArray == nullptr) {
-        err() << "No input devices found." << std::endl;
-        return m_identification;
-    }
-
-    // Get the array length and elements
-    jsize arrayLength = lJNIEnv->GetArrayLength(deviceIdsArray);
-    jint* deviceIds = lJNIEnv->GetIntArrayElements(deviceIdsArray, nullptr);
-
-    for (jsize i = 0; i < arrayLength; ++i) {
-        jobject inputDevice = lJNIEnv->CallStaticObjectMethod(inputDeviceClass, getDeviceMethod, deviceIds[i]);
+    for (jsize i = 0; i < deviceIds->getLength(); ++i) {
+        auto inputDevice = JniInputDevice::getDevice(lJNIEnv, inputDeviceClass, (*deviceIds)[i]);
         if (!inputDevice) continue;
 
-        jboolean supportsGamepad = lJNIEnv->CallBooleanMethod(inputDevice, supportsSourceMethod, jint(AINPUT_SOURCE_GAMEPAD | AINPUT_SOURCE_JOYSTICK));
-        if (!(bool)supportsGamepad) continue;
+        if (!inputDevice->supportsSource(AINPUT_SOURCE_GAMEPAD | AINPUT_SOURCE_JOYSTICK)) continue;
+
+        // Get state of nth gamepad
+        if (joy_index > 0)
+        {
+            --joy_index;
+            continue;
+        }
 
         m_identification = Joystick::Identification{
-            javaStringToStd(lJNIEnv, (jstring)lJNIEnv->CallObjectMethod(inputDevice, getNameMethod)),
-            (unsigned)lJNIEnv->CallIntMethod(inputDevice, getVendorIdMethod),
-            (unsigned)lJNIEnv->CallIntMethod(inputDevice, getProductIdMethod),
+                inputDevice->getName(),
+                inputDevice->getVendorId(),
+                inputDevice->getProductId(),
         };
+
+        m_currentDeviceIdx = int(i);
+
+        return true;
     }
 
-    // Release the array elements
-    lJNIEnv->ReleaseIntArrayElements(deviceIdsArray, deviceIds, 0);
+    return false;
+}
 
+
+////////////////////////////////////////////////////////////
+void JoystickImpl::close()
+{
+    // To implement
+}
+
+
+////////////////////////////////////////////////////////////
+JoystickCaps JoystickImpl::getCapabilities() const
+{
+    // To implement
+    return JoystickCaps{
+        Joystick::ButtonCount,
+        EnumArray<Joystick::Axis, bool, Joystick::AxisCount>{ true }
+    };
+}
+
+
+////////////////////////////////////////////////////////////
+Joystick::Identification JoystickImpl::getIdentification() const
+{
     return m_identification;
 }
 
--- a/src/SFML/Window/Android/JoystickImpl.cpp
+++ b/src/SFML/Window/Android/JoystickImpl.cpp
@@ -29,61 +29,53 @@
 #include <SFML/System/Err.hpp>
 #include <SFML/System/Android/Activity.hpp>
 
-static std::string javaStringToStd(JNIEnv* env, jstring str)
-{
-    const char* utfChars = env->GetStringUTFChars(str, nullptr);
-    std::string result(utfChars);
-    env->ReleaseStringUTFChars(str, utfChars);
-    return result;
-}
-
 template<class T>
 class JniArray {
 public:
     JniArray(JNIEnv* env, jintArray array)
-        : env(env)
-        , array(array)
-        , length(env->GetArrayLength(array))
-        , data(env->GetIntArrayElements(array, nullptr))
+        : m_env(env)
+        , m_array(array)
+        , m_length(env->GetArrayLength(array))
+        , m_data(env->GetIntArrayElements(array, nullptr))
     {}
 
-    JniArray(JniArray&& other)
+    JniArray(JniArray&& other) noexcept
     {
-        std::swap(env, other.env);
-        std::swap(array, other.array);
-        std::swap(length, other.length);
-        std::swap(data, other.data);
+        std::swap(m_env, other.m_env);
+        std::swap(m_array, other.m_array);
+        std::swap(m_length, other.m_length);
+        std::swap(m_data, other.m_data);
     }
 
     ~JniArray()
     {
-        if (data)
-            env->ReleaseIntArrayElements(array, data, 0);
+        if (m_data)
+            m_env->ReleaseIntArrayElements(m_array, m_data, 0);
     }
 
 public:
     T operator[](ssize_t idx)
     {
-        assert(0 <= idx && idx <= length);
-        return data[idx];
+        assert(0 <= idx && idx <= m_length);
+        return m_data[idx];
     }
 
     const T operator[](ssize_t idx) const
     {
-        assert(0 <= idx && idx <= length);
-        return data[idx];
+        assert(0 <= idx && idx <= m_length);
+        return m_data[idx];
     }
 
-    ssize_t getLength() const noexcept
+    [[nodiscard]] ssize_t getLength() const noexcept
     {
-        return length;
+        return m_length;
     }
 
 private:
-    JNIEnv * env = nullptr;
-    jintArray array = nullptr;
-    ssize_t length = 0;
-    T* data = nullptr;
+    JNIEnv * m_env = nullptr;
+    jintArray m_array = nullptr;
+    ssize_t m_length = 0;
+    T* m_data = nullptr;
 };
 
 namespace sf::priv
@@ -98,12 +90,12 @@ public:
         jmethodID getVendorIdMethod,
         jmethodID getProductIdMethod,
         jmethodID supportsSourceMethod)
-        : env(env)
-        , inputDevice(inputDevice)
-        , getNameMethod(getNameMethod)
-        , getVendorIdMethod(getVendorIdMethod)
-        , getProductIdMethod(getProductIdMethod)
-        , supportsSourceMethod(supportsSourceMethod)
+        : m_env(env)
+        , m_inputDevice(inputDevice)
+        , m_getNameMethod(getNameMethod)
+        , m_getVendorIdMethod(getVendorIdMethod)
+        , m_getProductIdMethod(getProductIdMethod)
+        , m_supportsSourceMethod(supportsSourceMethod)
     {}
 
 public:
@@ -116,7 +108,8 @@ public:
             return std::nullopt;
         }
 
-        jintArray deviceIdsArray = (jintArray)env->CallStaticObjectMethod(inputDeviceClass, getDeviceIdsMethod);
+        auto deviceIdsArray = static_cast<jintArray>(
+                env->CallStaticObjectMethod(inputDeviceClass, getDeviceIdsMethod));
         if (deviceIdsArray == nullptr)
         {
             err() << "No input devices found." << std::endl;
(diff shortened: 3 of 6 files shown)


src/SFML/System/Android/Activity.hpp
```diff
@@ -40,6 +41,7 @@
 #include <string>
 #include <unordered_map>
 #include <vector>
+#include <map>
```
nerudaj, line 44: TODO: remove this include

src/SFML/Window/Android/JoystickImpl.cpp
```diff
@@ -26,52 +26,264 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Window/JoystickImpl.hpp>
+#include <SFML/System/Err.hpp>
+#include <SFML/System/Android/Activity.hpp>
 
+template<class T>
```
nerudaj, line 32: TODO: move JNI helpers to a dedicated file

nerudaj:
@eXpl0it3r can you please approve rerun of the workflows? The previous was green on everything except some Windows static build UTs, I merged last two months worth of upstream into my branch, hopefully it'll fix the issue. Changes to Android part of the codebase shouldn't affect Windows builds.

eXpl0it3r:

nerudaj:
@eXpl0it3r Who shall I ask for feedback/review? Zombieschannel, Chris Trasher? Sorry, first time contributor here, not sure who to involve in the conversation.

eXpl0it3r:
All the above 😄 

Unfortunately, we don't have an expert in Android development and/or peripherals as part of the SFML Team, so while Chris and I will be able to do C++ code reviews, we'll likely need to depend on Zombieschannel and/or texus etc. who have more Android experience to help up test this feature.

@Zombieschannel / @texus do you have some gamepad that can connect to Android to test this?

nerudaj:
@eXpl0it3r Thanks for the response. Any gamepad should do, I tried with a bluetooth one (íPega), I tried wired options (Xbox), I tried smartphone-first (Backbone). Android version 14 (13 or perhaps even older should be fine as well).

Frankly, this was my first developing anything for Android. If the build process for the example code is too contrived or doesn't work, I can port it to this project of mine: https://github.com/nerudaj/dgm-template which at this point can generate APKs from Github Actions (slightly better than offering you unsigned APKs built by hand 😀), or we can arrange some Discord session.

Ofcs, take your time, we all have our lives 😉

Zombieschannel:
I can test on Android 5/8/10/13/15 with the Sony PS4 controller.

Zombieschannel:
Ok, I tested input on Android 13 (just with sf::Event), and while buttons do all get correctly detected, only the axis 0 is detected, no events happen for other axes for some reason.

I am getting a lot of these warnings when I move other axes though:
image
Not sure if it is of any use to you.

Tested connecting multiple controllers, and they do behave as one, as you mentioned.

Zombieschannel:
Testing with sf::Joystick seems to mostly work, the only issue being that hasAxis returns false on any axis other than 0, but getAxisPosition does return correct data for the "not available" axes.

Also interesting observation, the PS4 controller detects buttons from 0 to 14 except 3, no idea why, probably not connected to SFML.

nerudaj:
@Zombieschannel Thanks for the testing and feedback. As for the "hasAxis", EnumArray constructor probably works differently than I thought. I initially wanted to pretend that all 8 axii are supported, but I now realized I can get the actual capabilities from JNI (InputDevice.getMotionRanges). I'll address that.

I am a bit surprised that you got it working with `sf::Event`, I was under the impression that I didn't propagate any events (I didn't call `forwardEvent`). I did that on purpose as all axii are read from a single java event.

And yeah, just one controller is being registered, on purpose, as I didn't find any mappings between events and device IDs that caused them.

nerudaj:
* I've fixed the hasAxis, not it should correctly report all supported. Note that triggers are reported both as buttons and axes.
* I've also figured out how to support multiple gamepads, so far I've been able to test with two gamepads connected at the same time.

texus:
I quickly tested the Main.zip sample with an EasySMX controller over bluetooth on Android 15 and I can confirm that it works. Axes, triggers and button presses all changed one of the displayed values in the sample app.

src/SFML/Window/JoystickManager.cpp
```diff
@@ -112,7 +112,7 @@ JoystickManager::JoystickManager()
 ////////////////////////////////////////////////////////////
 JoystickManager::~JoystickManager()
 {
-    for (Item& item : m_joysticks)
+    for (const Item& item : m_joysticks)
```
nerudaj, line 115: @eXpl0it3r How can I supress the linter here? This can be const on Android, but not on any other platform.

ChrisThrasher: Yeah we can do that. Fix the rest of the CI pipeline then I can help with clang-tidy in this one particular circumstance.

nerudaj: It's ready.

Zombieschannel (approved):
Ok, I tested this on Android 8/13/15 (forgot SFML 3 cannot work on Android 5, I can test on Android 10 as well if it is needed).

Everything seems to fully work, either through `sf::Joystick` or `sf::Event`. At least with PS4 controllers. I tested with 3 of them at the same time.

One interesting observation, it seems like something changed between Androids 8 and 13 in how joysticks are handled, since on these controllers the Android 8 device did not assign controllers their own colors (these controllers have a color bar), the trigger buttons? (not just on/off, but anything in between) are detected as an axis in recent versions of android, while just being detected as buttons on Android 8.

Awesome to see `sf::Joystick` being added to Android, as this was one of the major missing features. :)

src/SFML/Window/JoystickManager.cpp
nerudaj: @ChrisThrasher Hi, how do you prefer to resolve that lint issue? `// NOLINT`, `ifdef` for Android, or some other way?

nerudaj: Hi @ChrisThrasher @eXpl0it3r , sorry to bother you, I'd like to fix that linter error so the PR is ready for review. As mentioned previously, linter wants me to make the Item ref const, since all relevant methods in Android joystick impl are const. But that only applies to the Android platform as the others are implemented in different ways.

ChrisThrasher:
```cpp
// NOLINTNEXTLINE(the clang check you want to silence goes here)
```

eXpl0it3r:
Happens sometimes, I re-ran the failed jobs

nerudaj:
@ChrisThrasher @eXpl0it3r Friendly reminder, this PR is waiting for review.

src/SFML/Window/Android/JniHelper.hpp
```diff
@@ -0,0 +1,340 @@
...
+    Jni(Jni&& other) noexcept
+    {
+        std::swap(m_vm, other.m_vm);
+    }
+    Jni(const Jni&&) = delete;
+    ~Jni()
+    {
+        if (m_vm)
+            m_vm->DetachCurrentThread();
+    }
```
eXpl0it3r:
```suggestion
    Jni(Jni&& other) noexcept
    {
        std::swap(m_vm, other.m_vm);
    }

    Jni(const Jni&&) = delete;

    ~Jni()
    {
        if (m_vm)
            m_vm->DetachCurrentThread();
    }
```

src/SFML/Window/Android/JoystickImpl.cpp
```diff
@@ -25,53 +25,147 @@
...
+            continue;
+
+        if (foundGamepadsSoFar < joyIndex)
+        {
+            ++foundGamepadsSoFar;
+            continue;
+        }
+
+        if (states.joystickStates.find(deviceId) != states.joystickStates.end())
+            continue; // this gamepad is already registered
```
eXpl0it3r: Won't this cause issues with the indexing? Meaning, shouldn't this be `return true` instead?

What if I call `open(3)` twice in a row? Wouldn't it first register it and then on the second call register a different device for the same index, because we don't return here?

src/SFML/Window/Android/JoystickImpl.cpp
```diff
@@ -25,53 +25,147 @@
...
+
+        if (const auto capabilities = getCapabilitiesFromJni(*inputDevice))
+            m_capabilities = *capabilities;
+        else
+            continue;
```
eXpl0it3r: In what scenarios can `getCapabilitiesFromJni` fail and is the right option to ignore this device? Or would it be better to call `return false` here?

src/SFML/Window/Android/JniHelper.hpp
```diff
@@ -0,0 +1,340 @@
...
+    T operator[](ssize_t idx) const
+    {
+        assert(0 <= idx && idx <= m_length);
+        return m_data[idx];
+    }
+
+    [[nodiscard]] ssize_t getLength() const noexcept
+    {
+        return m_length;
+    }
```
eXpl0it3r: It's a bit more work, but it would align better with SFML's way to separating code, by having only declarations in the header, move the template implementations into an *.inl file and have the non-template implementation in the *.cpp file.

At least it should be consistent and not mixed as it currently is.

src/SFML/Window/Android/WindowImplAndroid.cpp
```diff
@@ -330,19 +331,9 @@ int WindowImplAndroid::processEvent(int /* fd */, int /* events */, void* /* dat
...
-    lJavaVMAttachArgs.name    = "NativeThread";
-    lJavaVMAttachArgs.group   = nullptr;
-
-    lResult = lJavaVM->AttachCurrentThread(&lJNIEnv, &lJavaVMAttachArgs);
-
-    if (lResult == JNI_ERR)
+    auto    jni     = Jni::attachCurrentThread(states.activity->vm, &lJNIEnv);
+    if (!jni)
     {
         err() << "Failed to initialize JNI, couldn't get the Unicode value" << std::endl;
```
eXpl0it3r:
```suggestion
        err() << "Failed to initialize JNI" << std::endl;
```

Copy-paste from `WindowImplAndroid::getUnicode`

src/SFML/Window/Android/WindowImplAndroid.cpp
```diff
@@ -492,6 +503,34 @@ int WindowImplAndroid::processMotionEvent(AInputEvent* inputEvent, ActivityState
...
+
+            Also also, a same gamepad connected via same method can report different axii
+            on Windows and Android. Xbox One for example will report triggers as negative/positive
+            values on the single axis, while Android reports them on two separate axii.
+            */
+            const auto deviceId = AInputEvent_getDeviceId(inputEvent);
+            if (states.joystickStates.find(deviceId) == states.joystickStates.end())
+                return 1;
+
+            const float factor = 100.f; // Windows code normalizes axis to range <-100, 100> instead of sane <-1, 1>
```
eXpl0it3r: No need for judgmental comments 😉 

This is how the SFML API defines it and isn't Windows specific.

```suggestion
            const float factor = 100.f; // SFML normalizes axis to the range <-100, 100> instead of <-1, 1>
```

src/SFML/Window/Android/WindowImplAndroid.hpp
```diff
@@ -228,18 +230,23 @@ class WindowImplAndroid : public WindowImpl
...
+    static int processJoystickButtonEvent(AInputEvent*     inputEvent,
+                                          std::int32_t     action,
+                                          Joystick::Button button,
+                                          ActivityStates&  states);
     static int processMotionEvent(AInputEvent* inputEvent, ActivityStates& states);
     static int processPointerEvent(bool isDown, AInputEvent* event, ActivityStates& states);
 
     ////////////////////////////////////////////////////////////
-    /// \brief Convert a Android key to SFML key code
+    /// \brief Convert a Android key to SFML key code / joy button code
```
eXpl0it3r:
```suggestion
    /// \brief Convert a Android key to SFML key code / joystick button code
```

src/SFML/Window/Android/WindowImplAndroid.cpp
```diff
@@ -492,6 +503,34 @@ int WindowImplAndroid::processMotionEvent(AInputEvent* inputEvent, ActivityState
...
+            auto&       axes   = states.joystickStates.at(deviceId).axes;
+
+            for (unsigned axisIdx = 0; axisIdx < Joystick::AxisCount; ++axisIdx)
+            {
+                const auto axis = static_cast<Joystick::Axis>(axisIdx);
+                axes[axis] = AMotionEvent_getAxisValue(inputEvent, JoystickImpl::sfAxisToAndroid(axis), p) * factor;
```
eXpl0it3r, line 540: What happens if `sfAxisToAndroid` returns `-1`?

src/SFML/Window/Android/WindowImplAndroid.cpp
```diff
@@ -492,6 +503,34 @@ int WindowImplAndroid::processMotionEvent(AInputEvent* inputEvent, ActivityState
...
+            in the single input event, meaning that I can poll values for all axii in this single
+            loop iteration and it'll work.
+
+            Also, whoever coded Windows gamepad support had normalized axii values to <-100, 100>
+            instead of sane interval of <-1, 1>.
+
+            Also also, a same gamepad connected via same method can report different axii
+            on Windows and Android. Xbox One for example will report triggers as negative/positive
+            values on the single axis, while Android reports them on two separate axii.
+            */
```
eXpl0it3r:
```suggestion
            // There seems to be no direct mapping between input event and the ID of the device that
            // caused the event. However, as the single input event contains all axii changes, it's possible
            // to poll values for all axii in a single loop iteration.
            //
            // Additionally, some controllers such as the Xbox One controller will report triggers as
            // negative/positive values on the single axis on Windows, while Android reports them
            // on two separate axii.
```

The middle part is mentioned below and thus isn't needed here

src/SFML/Window/Android/JoystickImpl.cpp
nerudaj: API documentation for Android is not the best, but this one doesn't mention any failure cases: https://developer.android.com/reference/android/view/InputDevice#getMotionRanges()

Which should imply that this function can only fail in case something is downright wrong with the JNI invocation. In which case, `return false` would be better indeed.

src/SFML/Window/Android/WindowImplAndroid.cpp
nerudaj: Sorry, this one grinds my gear a little, it got a better of me :( Seriously, I'd like to hear why SFML opted for -100..100 instead of -1..1, just to satisfy my curiousity, but that's besides the point of this PR.

src/SFML/Window/Android/WindowImplAndroid.cpp
nerudaj: Once again, Android NDK leaves us guessing: https://developer.android.com/ndk/reference/group/input#amotionevent_getaxisvalue

But I tried to hardcode -1 and the function returns 0.f.

Furthermore, -1 should only ever happen if somebody adds a new axis to SFML and doesn't update the translation function.

src/SFML/Window/Android/JoystickImpl.cpp
nerudaj: There needs to be a skip (or a complete logic rewrite).

Assume this:
* `open(0)` -> registers device with id 42
* `open(1)` -> registers device with id 43
* device 42 disconnects
* `open(0)` -> finds device 43, but that is already registered as joystick 1! if you returned true, it would pop up in both slots. Instead, I skip, searching for any other device that might be there.

If you'd called `open(0)` twice in a row (without calling open(1) inbetween), that could be an issue. But that shouldn't happen. Not without the device disconnecting in the meanwhile. In that case, JoystickManager calls `close` and removes the device idx from the map so the device might be found again.

src/SFML/Window/Android/JniHelper.hpp
nerudaj: done

src/SFML/Window/Android/JoystickImpl.cpp
nerudaj: I've updated the comment in code to hopefully better communicate why there is a skip

src/SFML/Window/Android/JoystickImpl.cpp
nerudaj: changed to return false

src/SFML/Window/Android/JniHelper.hpp
nerudaj: You're the first person who's asking me to break the C++ code into even more files 😃 ...done

src/SFML/Window/Android/WindowImplAndroid.cpp
nerudaj: done

src/SFML/Window/Android/WindowImplAndroid.cpp
nerudaj: updated

src/SFML/Window/Android/WindowImplAndroid.hpp
nerudaj: done

src/SFML/Window/Android/WindowImplAndroid.cpp
nerudaj: done

nerudaj:
@eXpl0it3r Thanks for the review, all comments should be addressed now ;)

src/SFML/Window/Android/JniHelper.hpp
```diff
@@ -0,0 +1,204 @@
...
+
+#include <android/native_activity.h>
+
+#include <algorithm>
+#include <optional>
+#include <ostream>
+#include <string>
+
+#include <cassert>
+
```
eXpl0it3r, line 38: We're missing a namespace here, wouldn't want to put this into the global namespace, even if it shouldn't be an issue.

I suggest: `namespace sf::priv`

src/SFML/Window/Android/JoystickImpl.hpp
```diff
@@ -94,13 +94,19 @@ class JoystickImpl
...
+
+    [[nodiscard]] static int sfAxisToAndroid(Joystick::Axis axis);
 
 private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
     Joystick::Identification m_identification; ///< Joystick identification
+    JoystickCaps             m_capabilities;
+    int32_t                  m_currentDeviceIdx = -1; ///< Physical device ID
```
eXpl0it3r: Maybe instead of using `-1` as surrogate, we could use `std::optional` instead.

src/SFML/Window/Android/JoystickImpl.cpp
```diff
@@ -83,10 +179,74 @@ Joystick::Identification JoystickImpl::getIdentification() const
...
+    {
+        // This technically shouldn't happen, but when I have connect physical gamepad
+        // and then connect/disconnect a bluetooth one, states for the physical one
+        // disappears for a frame and then it reconnects.
+        return {false};
+    }
+
+    return {isConnected,
+            states.joystickStates.at(m_currentDeviceIdx).axes,
+            states.joystickStates.at(m_currentDeviceIdx).buttons};
```
eXpl0it3r: Direct access should be fine, as we just check it above

```suggestion
            states.joystickStates[m_currentDeviceIdx].axes,
            states.joystickStates[m_currentDeviceIdx].buttons};
```

src/SFML/Window/Android/WindowImplAndroid.cpp
```diff
@@ -492,6 +503,28 @@ int WindowImplAndroid::processMotionEvent(AInputEvent* inputEvent, ActivityState
...
+            //
+            // Additionally, some controllers such as the Xbox One controller will report triggers as
+            // negative/positive values on the single axis on Windows, while Android reports them
+            // on two separate axii.
+            const auto deviceId = AInputEvent_getDeviceId(inputEvent);
+            if (states.joystickStates.find(deviceId) == states.joystickStates.end())
+                return 1;
+
+            const float factor = 100.f; // SFML normalizes axis to the range <-100, 100> instead of <-1, 1>
+            auto&       axes   = states.joystickStates.at(deviceId).axes;
```
eXpl0it3r: Direct access should be fine here, as we just checked it above:

```suggestion
            auto&       axes   = states.joystickStates[deviceId].axes;
```

src/SFML/Window/Android/WindowImplAndroid.cpp
```diff
@@ -461,6 +458,20 @@ int WindowImplAndroid::processKeyEvent(AInputEvent* inputEvent, ActivityStates&
...
+                                                  std::int32_t     action,
+                                                  Joystick::Button button,
+                                                  ActivityStates&  states)
+{
+    const auto deviceId = AInputEvent_getDeviceId(inputEvent);
+    if (states.joystickStates.find(deviceId) == states.joystickStates.end())
+        return 1;
+
+    const auto buttonIdx = static_cast<std::underlying_type_t<decltype(button)>>(button);
+    states.joystickStates.at(deviceId).buttons[buttonIdx] = action == AKEY_EVENT_ACTION_DOWN;
```
eXpl0it3r: Direct access should be fine here, as we just checked it above:

```suggestion
    states.joystickStates[deviceId].buttons[buttonIdx] = action == AKEY_EVENT_ACTION_DOWN;
```

eXpl0it3r (changes requested):
Noticed some small things 😅

ChrisThrasher:

src/SFML/Window/Android/JniHelper.hpp
```diff
@@ -0,0 +1,199 @@
...
+class JniMotionRange
+{
+public:
+    [[nodiscard]] int getAxis() const;
+
+private:
+    friend class JniMotionRangeClass;
+
+    JniMotionRange(JNIEnv* env, jobject motionRange, jmethodID getAxisMethod);
```
ChrisThrasher: What's with all the private constructors and friend classes? Why not just use public constructors and drop the friends? What are we protecting against by doing this?

src/SFML/Window/Android/JniHelper.cpp
```diff
@@ -0,0 +1,286 @@
...
+m_getDeviceIdsMethod(getDeviceIdsMethod),
+m_getDeviceMethod(getDeviceMethod)
+{
+}
+
+
+////////////////////////////////////////////////////////////
+std::optional<JniInputDeviceClass> JniInputDeviceClass::findClass(JNIEnv* env)
+{
+    assert(env);
```
ChrisThrasher: Instead of using a pointer param and asserting it's not null, let's just use a `JNIEnv&` param.

src/SFML/Window/Android/JniHelper.cpp
```diff
@@ -0,0 +1,286 @@
...
+    if (m_vm)
+        m_vm->DetachCurrentThread();
+}
+
+
+////////////////////////////////////////////////////////////
+std::optional<Jni> Jni::attachCurrentThread(JavaVM* vm, JNIEnv** env)
+{
+    assert(vm);
+    assert(env);
```
ChrisThrasher: See previous comment about using reference params instead of pointers.

src/SFML/Window/Android/JniHelper.hpp
nerudaj: The intention here is that if somebody ever needs this helpers, there is a clear flow on how to instantiate them - they can't do so directly, they have to use `JniMotionRangeClass` in this case. If the constructor was public, they might not notice the `JniMotionRangeClass` and go down the rabbit hole of figuring out how to get all of these jobjects and jmethodIDs.

I'll make it public if you insist.

src/SFML/Window/Android/JniHelper.cpp
nerudaj: I will change it at other places but in this particular instance, JavaVM uses a pointer so the wrapper class can be movable and to not call detach twice.

src/SFML/Window/Android/JoystickImpl.hpp
nerudaj: changed

src/SFML/Window/Android/JniHelper.cpp
nerudaj: Changed JNIEnv to references where possible. JniList maintains a pointer due to swaps in the move constructor. I've commented on `Jni` in the other thread. A) both its imputs are coming directly from activity states where they are pointers anyway, B) `Jni` needs to be movable.

src/SFML/Window/Android/JniHelper.hpp
ChrisThrasher: I think I'd prefer a simpler solution with public constructors and no `friend`s. I see what you're going for, but I think it's more than we need for private implementation details. Perhaps add a comment near the top of JniHelper.hpp with a few sentences taking about how these classes are supposed to be used.

src/SFML/Window/Android/JniHelper.cpp
ChrisThrasher: I'm only talking about function parameters. I've already done work to reduce pointer usage in Android code and would like to keep doing that. If a function only needs a reference to data, let's use a reference parameter even if some Android API returns a pointer. If we know that is never null then we can always safely dereference it.

https://github.com/SFML/SFML/pull/3046
https://github.com/SFML/SFML/pull/3363

src/SFML/Window/Android/JniHelper.cpp
ChrisThrasher: The parameter itself ought to be a pointer. I'm not asking you to change how data is stored in the class.

src/SFML/Window/Android/JniHelper.hpp
```diff
@@ -145,16 +145,16 @@ class JniInputDevice
...
     [[nodiscard]] std::optional<JniArray<jint>> getDeviceIds();
 
     [[nodiscard]] std::optional<JniInputDevice> getDevice(jint idx);
 
 private:
-    JniInputDeviceClass(JNIEnv* env, jclass inputDeviceClass, jmethodID getDeviceIdsMethod, jmethodID getDeviceMethod);
+    JniInputDeviceClass(JNIEnv& env, jclass inputDeviceClass, jmethodID getDeviceIdsMethod, jmethodID getDeviceMethod);
 
-    JNIEnv*   m_env;
+    JNIEnv&   m_env;
```
ChrisThrasher, line 179: I'm not asking you to change how data is stored in the class. If it's important that these classes be reassignable, we can keep using a pointer instead of a reference.

ChrisThrasher:
To emphasize my point about pointers *as function parameters*, if a pointer argument must not be null, let's use a reference instead. I'd rather use the type system to enforce that the data is not null than have to add an `assert` to enforce that. I am only talking about function parameters. I am not talking about uses of pointers elsewhere. I'm letting you use your best judgment about whether these new classes should store a pointer or a reference.

nerudaj:
@ChrisThrasher please revert your last commit. Jni is a RAII wrapper that detaches the current thread when destroyed. Your changes will just crash all android apps.

We can debate whether that's the best approach. I'd argue that runtime crash that can be caught in QA is better than manual detaching that one can forget to do.

But if you'd like me to change it to manual detach instead of RAII wrapper, that's ultimately your call. I will respect that.

ChrisThrasher:
> @ChrisThrasher please revert your last commit. Jni is a RAII wrapper that detaches the current thread when destroyed. Your changes will just crash all android apps.
> 
> We can debate whether that's the best approach. I'd argue that runtime crash that can be caught in QA is better than manual detaching that one can forget to do.
> 
> But if you'd like me to change it to manual detach instead of RAII wrapper, that's ultimately your call. I will respect that.

I reverted I bug I introduced but did not revert the other style changes.

src/SFML/Window/Android/JniHelper.cpp
nerudaj: All functions should accept references instead of pointers now.

src/SFML/Window/Android/JniHelper.hpp
nerudaj: All private ctors have been made public, friends removed, documentation on preferred usage added.

src/SFML/Window/Android/JniHelper.cpp
```diff
@@ -242,39 +242,49 @@ std::optional<JniList<JniMotionRange, JniMotionRangeClass>> JniInputDevice::getM
...
 {
+    std::swap(m_vm, other.m_vm);
+    std::swap(m_env, other.m_env);
 }
 
 
 ////////////////////////////////////////////////////////////
 Jni::~Jni()
 {
+    m_env = nullptr;
```
ChrisThrasher: There is no need to change the value of data members in a destructor. This is just dead code. I went ahead and removed this line.

src/SFML/Window/Android/JniHelper.cpp
```diff
@@ -242,39 +242,49 @@ std::optional<JniList<JniMotionRange, JniMotionRangeClass>> JniInputDevice::getM
...
     if (m_vm)
```
ChrisThrasher, line 261: Is there a situation where `m_vm` would be null upon destruction? I believe this pointer is guaranteed to be non-null to added an `assert` to enforce this. Do we want to store a reference internally here?

src/SFML/Window/Android/WindowImplAndroid.cpp
```diff
@@ -716,11 +717,12 @@ char32_t WindowImplAndroid::getUnicode(AInputEvent* event)
     const std::lock_guard lock(states.mutex);
 
     // Prepare the Java virtual machine
-    JNIEnv* lJNIEnv = states.activity->env;
-    auto    jni     = Jni::attachCurrentThread(states.activity->vm, &lJNIEnv);
+    auto jni = Jni::attachCurrentThread(*states.activity);
     if (!jni)
         err() << "Failed to initialize JNI, couldn't get the Unicode value" << std::endl;
 
+    auto& lJNIEnv = jni->getEnv();
```
ChrisThrasher, line 860: What was the reasoning for adding `getEnv`? It appears that anywhere we use `getEnv` we could also access the environment through `states.activity`.

src/SFML/Window/Android/JniHelper.cpp
nerudaj: I fear that this change
image
could lead to runtime crashes.

If that `std::optional<Jni>` moves when being returned from `attachCurrentThread` or if anybody moves `Jni` for whatever reason, the moved-from object will contain nulls and trigger the assert.

It is fine for m_vm being null in destructor.

src/SFML/Window/Android/WindowImplAndroid.cpp
nerudaj: I thought I'll give the Jni object a little more reason to live. You know, as you've been simplifying those if conditions, if env was supplied by Jni, you wouldn't have the reason to do so in the first place. Also, env is only valid for as long as the Jni object itself, so it makes more sense.

src/SFML/Window/Android/JniHelper.cpp
ChrisThrasher: > the moved-from object will contain nulls and trigger the assert.

This is not necessarily true, but it's certainly possible. We can't know the value of `m_vm` when it's in a moved-from state. The broader question is, does this class really need to be moveable? I noticed you added some special members while deleting others. Can you talk me through your rationale there? In general I'd prefer to disable as many potentially problematic special members functions as possible if we don't need them.

ChrisThrasher:
I'm thinking more about the classes in JniHelper.hpp that contain references or pointers. Those refs/pointers cause us to have to think harder about special member functions and potential lifetime or memory bugs. Is it viable to instead remove all those ref and pointer members and instead just require that those things be provided via function arguments? This will make the calling code more complicated but it will let these classes be simpler and less bug prone. But perhaps it introduces more ugly problems I haven't yet considered. Was this a design you considered at some point?

nerudaj:
@ChrisThrasher 
> This is not necessarily true, but it's certainly possible. We can't know the value of m_vm when it's in a moved-from state.

Actually, we can. It's defined by the implementation of the move constructor. In this case, there is a std::swap between two pointers, one of them being default-constructed, thus null. How do we know the move happens? Because `attachCurrentThread` returns an optional, which triggers the move: https://godbolt.org/z/Ys7Tde8Gc

Since I don't want Jni to be copyable, I've deleted the copy constructor. If you delete the copy constructor, you must define the move constructor if you want your object to be returned from functions, even if RVO could optimize that return away.

> Those refs/pointers cause us to have to think harder about special member functions and potential lifetime or memory bugs.

Yes, they do. While writing this reply, I've realized that I've forgotten to delete the copy constructor of `JniArray`, my bad. The question is what kind of trade-off do you want - do you want to scrutinize each PR for whether it correctly detaches each thread, or frees all JNI resource,s or do you want to scrutinize each PR for whether an instance of Jni outlives all other helper classes and that it only lives till the end of the current scope?

I find the latter easier to reason about. Also, when put this way, you don't have to worry about the special functions either. If nobody is trying to keep Jni alive, they are not directly working with move constructors or copy constructors because they have no reason to.

**Disclaimer:** I am not even sure whether I am releasing all resources correctly with regards to the JNI code. The Android NDK docs are minimalistic and don't answer half of the questions I have. I sort of hoped that somebody on review who wrote the existing Android code would be able to tell me. That said, if there is something missing, it'll be much easier to just write a destructor than hunt for it in the code.

> Is it viable to instead remove all those ref and pointer members and instead just require that those things be provided via function arguments? Was this a design you considered at some point?

I didn't start with all these helper classes. I started with calling the JNI C API directly and only refactored from there because it was an awful mess. I don't think there is any meaningful middle ground, because:

* You can invoke JNI directly in code, as WindowImplAndroid.cpp does - This is "fine" for 2-3 method invocations, but grows out of proportion fast. I've considered refactoring these as well, but I didn't want to further complicate this PR.
* Group methodIDs/jobjects into structs and create helper methods for invoking those - This is essentially wrapping a C API with yet another C API. You don't get the RAII guarantees, and the only thing it gives you is the fact the `env` is supplied directly from NativeActivity.
* Embrace encapsulation/RAII - which is what I did

One extra point. If you use these helpers normally:

```
{
    auto jni = Jni::attachCurrentThread(nativeActivity);
    if (!jni) // error
    
    auto someClass = JniSomeClass::findClass(jni.getEnv());
    if (!someClass) // error
    
    auto result = someClass->foo();
} // scope ends, everything gets cleaned up
```

`env` is tied to the lifetime of `Jni`, at least visually, which reflects the fact it is only valid for as long as the thread is attached.

---

**Memory safety**

What will happen if I keep a reference to `someClass` dangling somewhere around? `env` will be always valid, because it lives inside NativeActivity. If a thread is attached, it will even work normally (I assume). If a thread is detached, it will crash the application, but it will do so in a well-defined manner (no nullptr dereference), you get a stacktrace, etc.

What will happen if I keep Jni alive? First, it'll be somewhat harder to do, because it is only movable. I don't have a better tool in C++ to tell people not to keep certain objects alive longer than they should. Second, stuff will generally continue working, up until a point where another piece of code tries to attach the current thread. I haven't tried that case, but I assume that `env` pointer will be simply overwritten (unless JNI can detect it wasn't detached earlier).

After that, anything goes. But that's also why attaching the current thread is protected by mutexes - these are not supposed to be long-term affairs. I don't have any idea how to make that more foolproof.

---

Ugh, that took me a solid hour to write :D

nerudaj:
@eXpl0it3r @ChrisThrasher @Zombieschannel Hi, are there any changes you want to be made to the PR based on that wall of text I've posted two weeks ago?

JonnyPtn:
Tested this by adding the joystick example to #3613  (which can be found here if wanted: https://github.com/JonnyPtn/SFML/tree/android_examples_joystick) and it all seems to work well

eXpl0it3r:
I dropped the master branch merge and squashed all the commits so I could more easily rebase onto master.

src/SFML/Window/Android/JoystickImpl.cpp
```diff
@@ -25,53 +25,173 @@
...
-    return false;
+    // This is called as a prefilter before open, but it
+    // would just duplicate the logic of open.
+    return index < Joystick::Count;
 }
 
 
 ////////////////////////////////////////////////////////////
-bool JoystickImpl::open(unsigned int /* index */)
+bool JoystickImpl::open(unsigned int joyIndex)
```
eXpl0it3r: Name should match declaration in header:


```suggestion
bool JoystickImpl::open(unsigned int index)
```

src/SFML/Window/Android/JoystickImpl.cpp
```diff
@@ -25,53 +25,173 @@
...
+
+    sf::priv::JoystickCaps capabilities = {sf::Joystick::ButtonCount, {false}};
+
+    for (ssize_t axisIdx = 0; axisIdx < motionRanges->getSize(); ++axisIdx)
+    {
+        auto motionRange = (*motionRanges)[axisIdx];
+        if (!motionRange)
+        {
+            sf::err() << "Gamepad was found, but its capabilities couldn't be read, skipping" << std::endl;
+            return std::nullopt;
```
eXpl0it3r, line 77: We're looping here, should we give up on the first missing capability or try to continue?

src/SFML/Window/Android/JoystickImpl.cpp
```diff
@@ -25,53 +25,173 @@
...
 ////////////////////////////////////////////////////////////
-void JoystickImpl::close()
+void JoystickImpl::close() const
 {
-    // To implement
+    ActivityStates&       states = getActivity();
+    const std::lock_guard lock(states.mutex);
+
+    if (m_currentDeviceIdx)
+        states.joystickStates.erase(*m_currentDeviceIdx);
```
eXpl0it3r, line 187: Do we need to reset `m_currentDeviceIdx`?

src/SFML/Window/Android/JniHelper.inl
```diff
@@ -0,0 +1,149 @@
...
+        return m_data[idx];
+    }
+
+    [[nodiscard]] ssize_t getLength() const
+    {
+        return m_length;
+    }
+
+private:
+    JNIEnv*   m_env{};
```
eXpl0it3r, line 92: Can we align the type to be `ref` or `ptr` across the different classes?

src/SFML/Window/Android/JniHelper.hpp
```diff
@@ -0,0 +1,224 @@
...
+///
+/// Thread is detached automatically when this class is destroyed.
+/// Construct this class by calling attachCurrentThread method.
+////////////////////////////////////////////////////////////
+struct Jni
+{
+public:
+    Jni(JavaVM& vm, JNIEnv& env);
+
+    Jni(Jni&& other) noexcept;
```
eXpl0it3r, line 196: Don't think there's a good reason to make `Jni` movable

src/SFML/Window/Android/JoystickButton.hpp
```diff
@@ -0,0 +1,61 @@
+////////////////////////////////////////////////////////////
+//
+// SFML - Simple and Fast Multimedia Library
+// Copyright (C) 2007-2025 Laurent Gomila (laurent@sfml-dev.org)
```
eXpl0it3r: Year needs to be updated everyhwere

src/SFML/Window/JoystickManager.cpp
```diff
@@ -112,7 +112,7 @@ JoystickManager::JoystickManager()
 ////////////////////////////////////////////////////////////
 JoystickManager::~JoystickManager()
 {
-    for (Item& item : m_joysticks)
+    for (Item& item : m_joysticks) // NOLINT(misc-const-correctness)
```
eXpl0it3r: I think using 
```suggestion
    for (const Item& item : m_joysticks)
```
should still work

src/SFML/Window/Android/JoystickImpl.cpp
nerudaj: This is working over an `std::optional<JniMotionRange` returned from `JniList::operator[]`:

```cpp
[[nodiscard]] std::optional<T> operator[](ssize_t idx) const
  {
      auto cls = TClass::findClass(m_env);
      if (!cls)
          return std::nullopt;

      jobject motionRange = m_env.CallObjectMethod(m_list, m_getMethod, idx);
      if (!motionRange)
          return std::nullopt;

      return cls->makeFromJava(motionRange);
  }
 ```
 
 as you can see, this method asks JNI for a bunch of internals. It is expected that all calls succeed, but in case any of these fail and return an empty optional, I assume something went horribly wrong and abort the computation.

File: src/SFML/Window/Android/JniHelper.hpp (hunk 29 above)
Comment (nerudaj) [reply src/SFML/Window/Android/JniHelper.hpp:196]: Answered here:
https://github.com/SFML/SFML/pull/3497#issuecomment-2975285457

**tl;dr** Jni is returned from `attachCurrentThread` and as such, it needs move constructor defined.

File: src/SFML/Window/Android/JniHelper.inl (hunk 28 above)
Comment (nerudaj) [reply src/SFML/Window/Android/JniHelper.inl:92]: Since JNI has to be movable, I wanted to have `m_vm` nullable, so it doesn't detach prematurely in the destructor.

File: src/SFML/Window/JoystickManager.cpp (hunk 31 above)
Comment (nerudaj) [reply src/SFML/Window/JoystickManager.cpp]: Since I am now resetting `m_currentDeviceIdx`, close is no longer `const` so this loop operates on a mutable reference.

File: src/SFML/Window/Android/JoystickButton.hpp (hunk 30 above)
Comment (nerudaj) [reply src/SFML/Window/Android/JoystickButton.hpp]: Updated

File: src/SFML/Window/Android/JoystickImpl.cpp (hunk 27 above)
Comment (nerudaj) [reply src/SFML/Window/Android/JoystickImpl.cpp:187]: Added the reset. It doesn't really matter since any subsequent `open` would overwrite it, but it makes more sense to reset it.

File: src/SFML/Window/Android/JoystickImpl.cpp (hunk 25 above)
Comment (nerudaj) [reply src/SFML/Window/Android/JoystickImpl.cpp]: Updated

File: src/SFML/Window/Android/JniHelper.inl [hunk 32]
```diff
@@ -0,0 +1,149 @@
...
+    ~JniArray()
+    {
+        if (m_data)
+            m_env->ReleaseIntArrayElements(m_array, m_data, 0);
+    }
+
+    [[nodiscard]] T operator[](ssize_t idx) const
+    {
+        assert(0 <= idx);
+        assert(idx <= m_length);
```
eXpl0it3r: This should actually be
```suggestion
        assert(idx < m_length);
```

src/SFML/Window/Android/WindowImplAndroid.cpp
```diff
@@ -810,22 +849,13 @@ char32_t WindowImplAndroid::getUnicode(AInputEvent* event)
...
-    lJavaVMAttachArgs.name    = "NativeThread";
-    lJavaVMAttachArgs.group   = nullptr;
-
-    lResult = lJavaVM->AttachCurrentThread(&lJNIEnv, &lJavaVMAttachArgs);
-
-    if (lResult == JNI_ERR)
+    // Prepare the Java virtual machine
+    auto jni = Jni::attachCurrentThread(*states.activity);
+    if (!jni)
         err() << "Failed to initialize JNI, couldn't get the Unicode value" << std::endl;
```
eXpl0it3r, line 856: Missing `return 0;` statement

src/SFML/Window/Android/JniHelper.cpp
```diff
@@ -0,0 +1,290 @@
...
+Jni::Jni(JavaVM& vm, JNIEnv& env) : m_vm(&vm), m_env(&env)
+{
+}
+
+
+////////////////////////////////////////////////////////////
+Jni::Jni(Jni&& other) noexcept
+{
+    std::swap(m_vm, other.m_vm);
+    std::swap(m_env, other.m_env);
```
eXpl0it3r: I think Thrasher already hinted at that, but this will lead to the moved-from object to retain a valid pointer, that could potentially corrupt the moved-to object, once the destructor is called.


```suggestion
Jni::Jni(Jni&& other) noexcept :
    m_vm(other.m_vm),
    m_env(other.m_env)
{
    other.m_vm = nullptr;
    other.m_env = nullptr;
```

src/SFML/Window/Android/JniHelper.inl
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    JniArray(const JniArray&) = delete;
+
+    JniArray(JniArray&& other) noexcept
+    {
+        std::swap(m_env, other.m_env);
+        std::swap(m_array, other.m_array);
+        std::swap(m_length, other.m_length);
+        std::swap(m_data, other.m_data);
```
eXpl0it3r: Same issue as with the Jni class

```suggestion
    JniArray(JniArray&& other) noexcept :
        m_env(other.m_env),
        m_array(other.m_array),
        m_length(other.m_length),
        m_data(other.m_data)
    {
        other.m_env = nullptr;
        other.m_array = nullptr;
        other.m_length = 0;
        other.m_data = nullptr;
```

eXpl0it3r (changes requested):
Some comments for you, but I've already implemented fixes for it that I'll push right after.

I believe there's still room for improvement with the whole JNI topic, but I think it's fine for now and I really want to ship SFML 3.1 😄

eXpl0it3r:
It took a while, but it's finally merged! 🥳 

Thanks a lot for adding this and sticking with us through this lengthy processes 😅

PR #1393, opened 2018-03-20
Android: Updated the example to build with gradle

MarioLiebisch:
This PR is mostly for visibility and you should consider it a work in progress.
It basically addresses #1311 and supercedes/includes #1315.

This updates the Android example to use Gradle rather than Ant for building.

To configure the example, simply run `gradle`. To build it, run `gradle build`, and to install it, run `gradle installDebug` or `gradle installRelease`.

Here are some known and so far not fixed issues (at least on my Nexus 5), but I don't think they're actually demo project related:

- Loading `canary.wav` fails for me.
- Once you switch to another app, the EGL context is no longer valid/restored.
- There's something broken with loading the shared STL libraries.

Feedback and bugfixes are appreciated, I'm really short on time right now. :(

PR #3498, opened 2025-04-30
Remove disabling of minifying in release

eXpl0it3r:
## Description

As reported by code scanning tools:

- https://github.com/SFML/SFML/security/code-scanning/22
- https://sonarcloud.io/project/issues?id=SFML_SFML&issues=AZYWVNYobgobvcurtOwm&open=AZYWVNYobgobvcurtOwm

This was added with the original rewrite with gradle support, see #1393

## Tasks

-   [ ] Tested on Android

## How to test this PR?

- Build for Android in release mode
- App should still work

### DIFF
--- a/examples/android/app/build.gradle.kts
+++ b/examples/android/app/build.gradle.kts
@@ -31,7 +31,7 @@ android {
     }
     buildTypes {
         release {
-            isMinifyEnabled = false
+            isMinifyEnabled = true
             proguardFiles(getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro")
         }
     }


eXpl0it3r:
@Zombieschannel as our personal Android master chief, can you confirm that removing this flag is okay?

ChrisThrasher (approved):

Zombieschannel:
Tbh I've never heard of nor used this flag, but from what info I gathered online and from Android studio it seems to be related to code shrinking and **it is off by default** so removing it shouldn't change anything.
I can confirm that a release build with either having this option explicitly off or just not having that line of code generates an **identical** apk.

And also while I do appreciate the compliment, I don't think I am nearly as big of an expert in the Android department as you probably think.😅

texus:
I would expect that this change doesn't resolve the sonarcloud "issue".
In the ["How can I fix it?" tab](https://sonarcloud.io/project/issues?id=SFML_SFML&issues=AZYWVNYobgobvcurtOwm&open=AZYWVNYobgobvcurtOwm&tab=how_to_fix) they give an example where the "bad" code doesn't specify the isMinifyEnabled option at all and the solution is to explicitly set `isMinifyEnabled` to `true`. So it feels like they aren't suggesting that you remove the line (which as Zombieschannel said is the same as setting it to false), but instead they suggest turning the option on.

I tried building the android example with the option set to true and it still worked. Which makes sense because the option is about minifying and obfuscating java code based on what I've read, and SFML isn't using java.

eXpl0it3r:
Fair point, I've enabled it now.

Lorenzooone:
Question: why would you need to make it the default in the example?

I feel like if someone has the need to obfuscate their code, they should do it themselves...

eXpl0it3r:
Mostly because it's a recommended default. The argument works the other way around. If someone doesn't want to obfuscate it, they can remove it.

It's always good to provide reasonable defaults.

PR #3501, opened 2025-05-03
Add support for getting and setting window states with size overload

eXpl0it3r:
## Description

This is an alternative/extended API design for #3496.
Main description & discussion on the other PR

`sf::WindowBase` implements the following three functions:
- `setState(sf::State state)`
- `setState(sf::State state, sf::Vector2u size)`
- `sf::State getState()`


include/SFML/Window/WindowBase.hpp
```diff
@@ -424,6 +424,44 @@ class SFML_WINDOW_API WindowBase
     ////////////////////////////////////////////////////////////
     void setIcon(Vector2u size, const std::uint8_t* pixels);
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Set the window state
+    ///
+    /// Change between windowed and fullscreen mode.
+    ///
+    /// \note This function will reuse the window's current size
+    /// when switching modes. If this behavior is no desired use the
```
ChrisThrasher:
```suggestion
    /// when switching modes. If this behavior is not desired, use the
```

include/SFML/Window/WindowBase.hpp
```diff
@@ -424,6 +424,44 @@ class SFML_WINDOW_API WindowBase
     ////////////////////////////////////////////////////////////
     void setIcon(Vector2u size, const std::uint8_t* pixels);
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Set the window state
+    ///
+    /// Change between windowed and fullscreen mode.
```
ChrisThrasher: This phrasing is brittle. Once we add more states as we have planned, this comment will go out of date. Can we use more generic language that doesn't require mentioning the windowed and fullscreen states?

include/SFML/Window/WindowBase.hpp
```diff
@@ -424,6 +424,44 @@ class SFML_WINDOW_API WindowBase
...
+    ///
+    /// \see `setState`
+    ///
+    ////////////////////////////////////////////////////////////
+    void setState(State state, sf::Vector2u size);
+
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the window's current state
+    ///
+    /// \return Current state the window is in
```
ChrisThrasher:
```suggestion
    /// \return Current state of the window
```

ChrisThrasher:
I implemented `setState` on macOS.

JonnyPtn:
So one of the problems this highlights on Mac is that SFML doesn't properly handle the two different "fullscreen" options on Mac:
- Classic fullscreen, as SFML currently uses, where it's a fullscreen window on the main desktop akin to borderless fullscreen
- Fullscreen space, the more modern approach to fullscreen on Mac where the app gets it's own space/desktop to itself

This is really an existing bug, as you can make the window fullscreen via the Mac menu bar, keyboard shortcut or window button but internally SFML still thinks it's not fullscreen (although you do get a resized event)

This API highlights the problem more, as you can get the window state which won't say it's fullscreen if you've made it a fullscreen space. Options are:
- Maintain the existing behaviour but disable the default NSWindow fullscreen menu option so users can't use the fullscreen space
- Change to use fullscreen space only and remove the legacy behaviour
- Expose the option to the users via some flag (build or runtime)

Personally I would rule out the first option, as it's the worst user experience on Mac. For reference, SDL gives the user the choice via a hint: https://wiki.libsdl.org/SDL2/SDL_HINT_VIDEO_MAC_FULLSCREEN_SPACES

PR #3507, opened 2025-05-18
Fix issue with Re-creation of Windows on Android

Lorenzooone:
## Description

This PR fixes an issue with SFML on Android, where it is not possible to close and re-create a Window (unless the user paused and resumed the application).
This happened due to the fact that for Surface creation, SFML waited for the onNativeWindowCreated callback to be called. However, the required callback before that one (onNativeWindowDestroyed) isn't called if you close the window from within the program itself.

To fix that, a call to create the Surface is added, which only happens if onNativeWindowCreated has already been called (and onNativeWindowDestroyed has not been called yet).

## Tasks

-   [ ] Tested on Linux
-   [ ] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [X] Tested on Android

## How to test this PR?

Use the basic Android example, and add to it the ability to recreate the window (so call create a second time).

One could make it so recreating the window can be triggered by either holding the touchscreen for a set amount of time, or by touching the top half of the screen.

### DIFF
--- a/src/SFML/Window/Android/WindowImplAndroid.cpp
+++ b/src/SFML/Window/Android/WindowImplAndroid.cpp
@@ -70,11 +70,16 @@ m_size(mode.size)
     if (state == State::Fullscreen)
         states.fullscreen = true;
 
+    const bool shouldCreateSurface = (states.window != nullptr) && (WindowImplAndroid::singleInstance == nullptr);
+
     WindowImplAndroid::singleInstance = this;
     states.forwardEvent               = forwardEvent;
 
     // Register process event callback
     states.processEvent = processEvent;
+
+    if (shouldCreateSurface)
+        states.forwardEvent(sf::Event::FocusGained{});
 
     states.initialized = true;
 }


src/SFML/Window/Android/WindowImplAndroid.cpp
```diff
@@ -70,12 +70,17 @@ m_size(mode.size)
...
+    const bool shouldCreateSurface = (states.window != nullptr) && (WindowImplAndroid::singleInstance == nullptr);
+
     WindowImplAndroid::singleInstance = this;
     states.forwardEvent               = forwardEvent;
 
     // Register process event callback
     states.processEvent = processEvent;
 
+    if (shouldCreateSurface)
+        states.forwardEvent(sf::Event::FocusGained{});
```
eXpl0it3r, line 82: Does this trigger the creation?

Lorenzooone: Yes. It is the same mechanism onNativeWindowCreated uses.

Processing the FocusGained event sets m_windowBeingCreated to true here: https://github.com/SFML/SFML/blob/387082237bda73c8cf0f47c27ae3f06c5082c6c2/src/SFML/Window/Android/WindowImplAndroid.cpp#L243

That, in turn, causes the Surface creation here: https://github.com/SFML/SFML/blob/387082237bda73c8cf0f47c27ae3f06c5082c6c2/src/SFML/Window/Android/WindowImplAndroid.cpp#L115

Lorenzooone: For my thoughs on using opaque event types like this to achieve window creation, see: https://github.com/SFML/SFML/issues/3508

eXpl0it3r:

PR #3513, opened 2025-05-28
Linux DRM - Fix issues with current implementation

Lorenzooone:
## Description

This PR fixes three issues I noticed while testing the DRM setup by myself:
- A segmentation fault if the initDrm function fails (i.e. due to the file specified being the wrong one).
- A segmentation fault if libEgl is not loaded (the program will still not work properly, to be clear... But at least now the user will be able to know why. Before this PR, SFML said nothing to the user, and it just crashed)
- The lack of the ability to select which screen the program should target.

The first two points are fixed thanks to the new function inside DRMContext.cpp: getEglLoaded(), which returns whether Egl was dynamically loaded correctly.
The last point is solved by adding the SFML_DRM_CONNECTOR_ID environment variable, which SFML checks at runtime. If this variable is set, it will specify the ID for the connector (and the screen that is plugged into said connector).

## Tasks

-   [X] Tested on Linux
-   [ ] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

To test this, a Linux environment without a Desktop active is required.
This can be achieved by using certain combinations on some devices, like CTRL + Alt + F3.

For point 1, to test this, simply set SFML_DRM_DEVICE to a bad file value (anything will work). It shouldn't crash inside DRMContext.

For point 2, launch a program in an environment without libegl1. I used a Raspberry Pi loaded with the Raspbian OS Lite image, with libgm1 and libgl1 installed, but without libegl1 installed. Again, the result is that it shouldn't crash inside DRMContext.
A helpful error message should also get printed.

For point 3, it requires a second monitor. Check the ID of the connector via the drm_info command, and then the user should be able to change the screen on which the application is displayed on.

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        {
                window.close();
        }

        window.clear();
        window.display();
    }
}
```

### DIFF
--- a/src/SFML/Window/DRM/DRMContext.cpp
+++ b/src/SFML/Window/DRM/DRMContext.cpp
@@ -68,11 +68,6 @@ int             contextCount   = 0;
 EGLDisplay      display        = EGL_NO_DISPLAY;
 int             waitingForFlip = 0;
 
-bool isEglLoaded()
-{
-    return eglLoaded;
-}
-
 void pageFlipHandler(int /* fd */, unsigned int /* frame */, unsigned int /* sec */, unsigned int /* usec */, void* data)
 {
     int* temp = static_cast<int*>(data);
@@ -555,7 +550,7 @@ DRMContext::DRMContext(DRMContext* shared)
 
     // Get the initialized EGL display
     m_display = getInitializedDisplay();
-    if (!isEglLoaded())
+    if (!eglLoaded)
         return;
 
     // Get the best EGL config matching the default video settings
@@ -583,7 +578,7 @@ DRMContext::DRMContext(DRMContext* shared, const ContextSettings& settings, cons
 
     // Get the initialized EGL display
     m_display = getInitializedDisplay();
-    if (!isEglLoaded())
+    if (!eglLoaded)
         return;
 
     // Get the best EGL config matching the requested video settings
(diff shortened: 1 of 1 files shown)


eXpl0it3r:

src/SFML/Window/DRM/DRMContext.cpp
```diff
@@ -67,6 +68,11 @@ int             contextCount   = 0;
 EGLDisplay      display        = EGL_NO_DISPLAY;
 int             waitingForFlip = 0;
 
+bool getEglLoaded()
```
eXpl0it3r, line 71:
```suggestion
bool isEglLoaded()
```

eXpl0it3r (changes requested):

src/SFML/Window/DRM/DRMContext.cpp
Lorenzooone: Changed accordingly! :+1:

src/SFML/Window/DRM/DRMContext.cpp
```diff
@@ -522,6 +555,8 @@ DRMContext::DRMContext(DRMContext* shared)
 
     // Get the initialized EGL display
     m_display = getInitializedDisplay();
+    if (!isEglLoaded())
```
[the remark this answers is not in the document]
Lorenzooone: Done!

PR #3518, opened 2025-05-31
Deprecate `sf::Font::getKerning` overload which uses `std::uint32_t` to represent Unicode codepoints

ChrisThrasher:
## Description

Before SFML 3, we switched from using std::uint32_t to char32_t to represent Unicode codepoints. This was an oversight that got missed during that transition in PR https://github.com/SFML/SFML/pull/2480.


include/SFML/Graphics/Font.hpp
```diff
@@ -251,6 +251,31 @@ class SFML_GRAPHICS_API Font
     ////////////////////////////////////////////////////////////
     [[nodiscard]] bool hasGlyph(char32_t codePoint) const;
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the kerning offset of two glyphs
+    ///
+    /// \deprecated Use `char32_t` overload of `getKerning` instead.
```
eXpl0it3r:
```suggestion
    /// \deprecated Use the `getKerning(char32_t, char32_t, unsigned int, bool bold)` overload instead.
```

I recommend to use the function signature, this way you get a clickable link:

image

include/SFML/Graphics/Font.hpp
```diff
@@ -251,6 +251,31 @@ class SFML_GRAPHICS_API Font
...
+    ///
+    /// \param first         Unicode code point of the first character
+    /// \param second        Unicode code point of the second character
+    /// \param characterSize Reference character size
+    /// \param bold          Retrieve the bold version or the regular one?
+    ///
+    /// \return Kerning value for `first` and `second`, in pixels
+    ///
+    ////////////////////////////////////////////////////////////
+    [[deprecated("Use char32_t overload")]] [[nodiscard]] float getKerning(
```
eXpl0it3r: Personally I'd align the messages between code and doxygen

```suggestion
    [[deprecated("Use the getKerning(char32_t, char32_t, unsigned int, bool bold) overload")]] [[nodiscard]] float getKerning(
```

PR #3519, opened 2025-06-01
Use const ref instead of cloning channelMap

dogunbound:
getChannelMap was cloning the data unnecessarily. It wasn't building the data in the function, so I saw no reason why it had to pass by value instead of simply passing by const reference.

This is just a proposal, so nothing too serious. The one major benefit is simply being more consistent. There are other uses of the `getChannelMap` (such as in SoundRecorder and InputSoundFile) that was passing by const reference instead of cloning the data.

This may also improve the CSFML code. I see that ya'll are cloning the data in the CSFML layer because of this.

Looks like there is some proof that it was mistakenly forgotten:
https://github.com/SFML/SFML/pull/2749#discussion_r1478634083

There is an initiative to use the const references :)

### DIFF
--- a/include/SFML/Audio/SoundBuffer.hpp
+++ b/include/SFML/Audio/SoundBuffer.hpp
@@ -291,7 +291,7 @@ public:
     /// \see `getSampleRate`, `getChannelCount`, `getDuration`
     ///
     ////////////////////////////////////////////////////////////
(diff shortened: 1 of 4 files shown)


ChrisThrasher:
¯\_(ツ)_/¯

ChrisThrasher:

ChrisThrasher (approved):
Thanks for catching this

PR #3523, opened 2025-06-06
Improve audio device and stream rerouting management

### LINKED ISSUE #3490: No audio after resuming from sleep
### Prerequisite Checklist

- [x] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [x] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [x] I am here to report an issue and not to just ask a question or look for help (use the [forum](https://en.sfml-dev.org/forums/index.php#c3) or [Discord](https://discord.gg/nr4X7Fh) instead)

### Describe your issue here

The audio device stops playing any sounds or music after sleep or hibernation then resuming my PC.
Confirmed by others.

### Your Environment

- OS / distro / window manager: Windows 11
- SFML version: 3.0.0
- Compiler / toolchain: x86_64-w64-mingw32.static 

### Steps to reproduce

1. Just play any audio with SFML
2. Press the sleep button
3. Resume

### Expected behavior

The audio file continues playing.

### Actual behavior

The audio stops playing. Setting play state does not have any effect.

binary1248:
This continues from #3490.

In the initial implementation using miniaudio, device selection and rerouting wasn't working as most users would have expected. Because SFML was always explicitly passing an audio device ID to miniaudio during device creation, miniaudio would interpret it as a request to only use that specific device and stop the device completely if it went missing. The functionality that most users would be accustomed to which automatically switches devices to whichever is the new default on the system is known as "automatic stream routing". This automatic stream routing is disabled by miniaudio when specifying a device ID.

In order to enable automatic stream routing we must not pass any device ID when initializing the audio device. Doing so requests miniaudio to use the "Default Device" as defined by the operating system. Indeed many multimedia applications and games provide this "Default Device" or "Default" entry when allowing their users to select from a list of devices even though such a "Default Device" isn't listed as a device in the operating system sound device management GUI. It is an abstract device that allows the user to enable automatic stream routing without having to understand how it works under the hood.

In the comments in #3490, I listed 3 possible solutions to the issue. 2 of them were in our control and the 3rd was within miniaudio's control. After much deliberation I have come to the conclusion that miniaudio is very unlikely to implement what I described as a possible solution instead opting to provide us with the tools necessary to implement something that comes pretty close ourselves.

For almost all users who don't care about setting an audio device, leaving everything at their default values will "just work", selecting the "Default Device" and thus enabling automatic stream routing on the systems that support it.

For those users that have reasons to select a non-default device we provide the option to do so as well. The question is what happens if this device goes missing at some point while the application is still running and playing audio. In this case, miniaudio will stop the device (which has possibly gone completely missing) and require us to explicitly tell it what to do next. In order to provide application developers with maximum flexibility I chose to offer forwarding the device state notifications all the way to application code so this scenario can be handled by the application developers themselves. Every application will have its own requirements and procedures when it comes to handling device loss.

From application code, the developer can choose to reroute to the "Default Device". Doing so will automatically continue playing audio on some other audio device available on the system. Sometimes end users might find this surprising or maybe even outright annoying e.g. when audio starts playing from speakers in the middle of the night. In such cases, the developer can also opt to route the audio to the null device. Before this change, the null device was not able to be explicitly requested by the application developer, instead being reserved as a fallback device in case no other devices were available. Because there is now a legitimate reason for allowing the developer to select the null device it has been exposed through the public API as well.

A demonstration of how such stream rerouting code might look like is provided in the newly added sound-device example. Unfortunately, as of this time, miniaudio does not provide a cross-platform way to notify us of when new audio devices are added to the system. Thus, as can be seen in the example, we have to resort to periodically polling for devices until the original device shows up again.

Audio device selection and routing is a very advanced topic. In my analysis of how many applications and games handle device selection, removal and rerouting I have come to the conclusion that not many of them actually care about a seamless user experience. Most of the time removing the device that was selected in the application just lead to audio not working any more until the application was completely restarted. Because we want to offer SFML users maximum flexibility to develop robust applications I am confident that the API offered in this PR is the way to go. Feel free to suggest/make adjustments as necessary.

I am not providing any test code here since that is already provided in the added example.

Closes #3490

@oomek @Bambo-Borris


src/SFML/Audio/AudioDevice.cpp
```diff
@@ -484,6 +572,60 @@ std::optional<ma_device_id> AudioDevice::getSelectedDeviceId() const
...
+
+    // If the NULL audio backend also doesn't provide a device we give up
+    if (deviceCount == 0)
+    {
+        m_context.reset();
+        return false;
+    }
+
+    if (!getCurrentDeviceSelection().useNull && (m_context->backend == ma_backend_null))
+        err() << "Using NULL audio backend for playback" << std::endl;
```
eXpl0it3r: As selecting the NULL device is a state the user can choose, the error message is no longer sufficient, as it doesn't indicate that the NULL device is used despite a different selection.

Additionally, one might consider using the term "device" instead of "audio backend"

src/SFML/Audio/AudioDevice.cpp
```diff
@@ -502,39 +644,50 @@ bool AudioDevice::initialize()
...
+        const std::lock_guard lock(notificationCallback.mutex);
+        const auto&           callback = notificationCallback.callback;
+
+        if (!callback)
+            return;
+
+        switch (notification->type)
+        {
+            case ma_device_notification_type_started:
+                callback(PlaybackDevice::Notification::DeviceStarted);
```
eXpl0it3r, line 670: Would it make sense and/or is it possible to also translate `notification->pDevice` and pass it to the callback?

Or is that easily determined with `sf::PlaybackDevice::getDevice()`?

src/SFML/Audio/AudioDevice.cpp
```diff
@@ -254,39 +228,152 @@ std::vector<AudioDevice::DeviceEntry> AudioDevice::getAvailableDevices()
...
+    std::size_t                                     deviceNameLength{};
+
+    if (const auto result = ma_device_get_name(&*instance->m_playbackDevice,
+                                               ma_device_type_playback,
+                                               deviceName.data(),
+                                               deviceName.size(),
+                                               &deviceNameLength);
+        result != MA_SUCCESS)
+    {
+        err() << "Failed to get name of audio playback device: " << ma_result_description(result) << std::endl;
```
eXpl0it3r:
```suggestion
        err() << "Failed to get the name of audio playback device: " << ma_result_description(result) << std::endl;
```

eXpl0it3r (changes requested):

src/SFML/Audio/AudioDevice.cpp
binary1248: Some of these notifications come from the active backend itself, independent of what or whether there is still a device available. Even if there was information in `notification->pDevice` it wouldn't be reliable enough to base any algorithmic decisions on.

oomek:
I know this is only partially related, but I have to ask.
How can I get the sample rate of the current playback device?
I need it to scale the output of the fft.

binary1248:
@oomek SFML will always request the device at its native sample rate. After miniaudio successfully initializes the device, it stores the sample rate of the device in the `sampleRate` member of the playback device object. We just have to expose this to the user through the public API. This seems like a simple enough task that I will give others the opportunity to write code as well. 😄

oomek:
@binary1248 I'm unable to find `sampleRate` member in `sf::PlaybackDevice`

binary1248:
`sampleRate` is a member of `m_playbackDevice` which is the miniaudio `ma_device` that functions as our playback device.

ChrisThrasher:
Rebased on `master` and fixed conflicts.

src/SFML/Audio/AudioDevice.cpp
eXpl0it3r: I suggest:

```suggestion
        err() << "Using NULL audio playback device, even though it wasn't requested" << std::endl;
```

eXpl0it3r (approved):

PR #3524, opened 2025-06-06
Update miniaudio to v0.11.22.

binary1248:
Title. Not much to be said.

Release notes can be found here: https://github.com/mackron/miniaudio/releases/tag/0.11.22

There is a non-zero probability that this update might fix audio related issues reported to us since we switched over to miniaudio.


src/SFML/Audio/Miniaudio.cpp
```diff
@@ -29,5 +29,4 @@
 #pragma GCC diagnostic ignored "-Wnull-dereference" // Work around MinGW warning
 #endif
 // #define MA_DEBUG_OUTPUT // Uncomment to enable miniaudio debug output to console
-#define MINIAUDIO_IMPLEMENTATION
-#include <miniaudio.h>
+#include <miniaudio.c>
```
ChrisThrasher: I’d prefer we not change this file. Creating a .c file just to include that seems unnecessary and unconventional.

binary1248: This change will have to be made whether we like it or not. The author has already stated that `MINIAUDIO_IMPLEMENTATION` is deprecated and will no longer be supported in the future. Including `miniaudio.c` is the officially sanctioned method to do what we are already doing now in a future proof way.

ChrisThrasher: Is the expectation that miniaudio.c is going to change in the future to actually include the function definitions instead of merely being the two lines that it is now?

src/SFML/Audio/Miniaudio.cpp
```diff
@@ -29,5 +29,5 @@
 #pragma GCC diagnostic ignored "-Wnull-dereference" // Work around MinGW warning
 #endif
 // #define MA_DEBUG_OUTPUT // Uncomment to enable miniaudio debug output to console
-#define MINIAUDIO_IMPLEMENTATION
-#include <miniaudio.h>
+// NOLINTNEXTLINE(bugprone-suspicious-include)
```
ChrisThrasher: We can more simply just put a `// NOLINT(check name here)` on the same line

src/SFML/Audio/Miniaudio.cpp
binary1248: Yes. The content of this file is going to stay like that only during the transition period. The breaking change will be made in v0.12 of miniaudio.

src/SFML/Audio/Miniaudio.cpp
binary1248: Done.

ChrisThrasher (approved):

PR #3525, opened 2025-06-12
Add `sf::PlaybackDevice::getDeviceSampleRate`

oomek:
## Description

Not related to a specific issue, but mentioned here #https://github.com/SFML/SFML/pull/3523#issuecomment-2963869250

## Tasks

-   [ ] Tested on Linux
-   [x] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

### DIFF
--- a/include/SFML/Audio/PlaybackDevice.hpp
+++ b/include/SFML/Audio/PlaybackDevice.hpp
@@ -107,4 +107,12 @@ namespace sf::PlaybackDevice
 ////////////////////////////////////////////////////////////
 [[nodiscard]] SFML_AUDIO_API std::optional<std::string> getDevice();
 
+////////////////////////////////////////////////////////////
+/// \brief Get the sample rate of the current audio playback device
+///
+/// \return The sample rate of the current audio playback device or 0 if there is none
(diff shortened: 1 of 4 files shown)


include/SFML/Audio/PlaybackDevice.hpp
```diff
@@ -107,4 +107,12 @@ namespace sf::PlaybackDevice
 ////////////////////////////////////////////////////////////
 [[nodiscard]] SFML_AUDIO_API std::optional<std::string> getDevice();
 
+////////////////////////////////////////////////////////////
+/// \brief Get the sample rate of the current audio playback device
+///
+/// \return The sample rate of the current audio playback device or 0 if there is none
```
ChrisThrasher: If it’s possible for this function to potentially not return a value, why not use `std::optional<std::uint32_t>`? That will clearly express how this function may not be able to return something.

oomek: Done.

ChrisThrasher:

PR #3534, opened 2025-06-20
[Backport] Fix pkg-config when SFML_PKGCONFIG_INSTALL_DIR is unset

dkosmari:
This is just commit a87763becbc4672b38f1021418ed94caa0f6540a grafted to the 3.0.x branch, so it can generate the .pc scripts on the first run of cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -201,9 +201,6 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 set_property(GLOBAL PROPERTY PREDEFINED_TARGETS_FOLDER "CMake")
 
-# add the subdirectories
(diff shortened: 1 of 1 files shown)


ChrisThrasher:
I don’t foresee a 3.0.2 release so I don’t think there is reason to further merge anything into 3.0.x.

dkosmari:
Right now the 3.0 releases don't build correctly if you need pkg-config scripts, which is ubiquitous on Linux systems.
Even worse, when packaging them, pkg-config is often used to automatically generate dependencies, and will create bogus "requires" dependencies on `sfml-audio`, `sfml-graphics`, etc. In one instance, the resulting RPM packages for SFML 3.0.1 x86-64 pulled these missing dependencies from existing SFML 2.6.x x86 packages (which in turn pull in all other 32-bit dependencies.)

ChrisThrasher:
This will be fixed when we release v3.1.0 so there is no need to add this to the 3.0.x branch.

Issue #3537, opened 2025-07-01
Release Automation


eXpl0it3r:
### Prerequisite Checklist

- [x] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [x] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [x] I understand [SFML's scope](https://www.sfml-dev.org/contribute.php#general-considerations) and believe the feature fits it

### Describe your feature request here

The full release process is currently holding back a higher cadence in SFML releases, because it's a lot of tedious work and basically just @eXpl0it3r doing the work, thus the bottleneck doubles. Various attempts have been made to automate this in the past, but it was never fully executed. Seeing the backlog for SFML 3.x slowly filling up, it becomes apparently, that if we want to be faster in creating SFML releases, some of the work needs to be automated.

This issue collects ideas, issues and PRs for release automation in one place.

- [ ] Release Builds
- [ ] Documentation Updates
- [ ] Download Page

### Use Cases

a

### API Example

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode({1280, 720}), "Minimal, complete and verifiable example");
    window.setFramerateLimit(60);

    while (window.isOpen())
    {
        while (const std::optional event = window.pollEvent())
        {
            if (event->is<sf::Event::Closed>())
                window.close();
        }

        window.clear();
        window.display();
    }
}
```

ZXShady:
Solution: get more exploiters

shivam-dhir:
Hi, I would love to work on this issue

eXpl0it3r:
The release build creation has been implemented. What's open is the automation for the documentation on the website and download page.

PR #3538, opened 2025-07-01
Add release builds

eXpl0it3r:
## Description

This adds a new CI workflow that creates SFML builds for all the configuration (and more) that we usually provide pre-compiled builds for. It makes sure that the content is packages in the same way we usually ship our pre-compiled builds. The whole process is a bit messy, which is also why automating it can help a lot.

- MinGW version matches the one used for SFML 3.0.0
- macOS builds can no longer ship old SDK compatibilities, as that process is not supported by Apple and hard to replicated in CI
- This targets the `3.0.x` branch, as the 3.0.1 builds are still waiting to be created
    - I had to backport the removal of windows-2019 and update of windows-2025 images, to fix the failing builds
- The artifacts are zipped twice, because on Linux and macOS we want to retain symlinks and use tar.gz
- The archive names use either the git tag or the SHA if it's not a tag build

I'm open to ideas on cleaning this up, but I fear a lot of it, is required to match the usual SDK distributions.

This PR is related to #3537 

## How to test this PR?

You should be able to manually trigger the workflow. Alternatively, push a modified YAML to your fork, that executes the release builds every time.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,345 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - .*
+  workflow_dispatch:
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build:
+    name: ${{ matrix.platform.name }} ${{ matrix.config.name }}
+    runs-on: ${{ matrix.platform.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+        - { name: Windows VS2022 x86, os: windows-2025 }
+        - { name: Windows VS2022 x64, os: windows-2025 }
+        - { name: Windows MinGW x86,  os: windows-2025, flags: -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ }
+        - { name: Windows MinGW x64,  os: windows-2025, flags: -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ }
+        - { name: Linux GCC,          os: ubuntu-24.04 }
+        - { name: macOS x64,          os: macos-13 }
+        - { name: macOS arm64,        os: macos-15 }
+        config:
+        - { name: Shared, flags: -DBUILD_SHARED_LIBS=ON }
+        - { name: Static, flags: -DBUILD_SHARED_LIBS=OFF }
+        type:
+        - { name: Release, flags: -DCMAKE_BUILD_TYPE=Release }
+        - { name: Debug, flags: -DCMAKE_BUILD_TYPE=Debug }
+        include:
+        - platform: { name: macOS x64, os: macos-13 }
+          config: { name: Frameworks, flags: -DSFML_BUILD_FRAMEWORKS=ON -DBUILD_SHARED_LIBS=ON }
+          type: { name: Release }
+        - platform: { name: macOS arm64, os: macos-15 }
+          config: { name: Frameworks, flags: -DSFML_BUILD_FRAMEWORKS=ON -DBUILD_SHARED_LIBS=ON }
+          type: { name: Release }
+        exclude:
+        - platform: { name: Linux GCC }
+          type: { name: Debug }
+        - platform: { name: macOS x64 }
+          type: { name: Debug }
+        - platform: { name: macOS arm64 }
+          type: { name: Debug }
+
+    steps:
+    - name: Checkout Code
+      uses: actions/checkout@v4
+
+    - name: Set Visual Studio Architecture
+      if: contains(matrix.platform.name, 'Windows VS')
+      uses: ilammy/msvc-dev-cmd@v1
+      with:
+        arch: ${{ contains(matrix.platform.name, 'x86') && 'x86' || 'x64' }}
+
+    - name: Get CMake and Ninja
+      uses: lukka/get-cmake@latest
+      with:
+        cmakeVersion: '3.25'
+        ninjaVersion: latest
+
+    - name: Install Linux Dependencies and Tools
+      if: runner.os == 'Linux'
+      run: sudo apt-get update && sudo apt-get install xorg-dev libxrandr-dev libxcursor-dev libxi-dev libudev-dev libflac-dev libvorbis-dev libgl1-mesa-dev libegl1-mesa-dev
+
+    - name: Install macOS Tools
+      if: runner.os == 'macOS'
+      run: |
+        brew update
+
+    - name: Cache MinGW x86
+      if: matrix.platform.name == 'Windows MinGW x86'
+      id: mingw32-cache
+      uses: actions/cache@v4
+      with:
+        path: "C:\\Program Files\\mingw32"
+        key: winlibs-i686-posix-dwarf-gcc-14.2.0-mingw-w64ucrt-12.0.0-r2
+
+    - name: Install MinGW x86
+      if: matrix.platform.name == 'Windows MinGW x86' && steps.mingw32-cache.outputs.cache-hit != 'true'
+      run: |
+        curl -Lo mingw32.zip https://github.com/brechtsanders/winlibs_mingw/releases/download/14.2.0posix-19.1.1-12.0.0-ucrt-r2/winlibs-i686-posix-dwarf-gcc-14.2.0-mingw-w64ucrt-12.0.0-r2.zip
+        unzip -qq -d "C:\Program Files" mingw32.zip
+
+    - name: Add MinGW x86 to PATH
+      if: matrix.platform.name == 'Windows MinGW x86'
+      run: |
+        echo "C:\Program Files\mingw32\bin" >> $GITHUB_PATH
+
+    - name: Cache MinGW x64
+      if: matrix.platform.name == 'Windows MinGW x64'
+      id: mingw64-cache
+      uses: actions/cache@v4
+      with:
+        path: "C:\\Program Files\\mingw64"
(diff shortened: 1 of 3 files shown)


Lorenzooone:
Is the output of this available anywhere to check how it is similar/differs from the official releases?

eXpl0it3r:
Not for the latest commit and the manually triggering only works after the change has been merged, but I've first pushed a version where the filter wasn't yet available. The output for this can be found here: https://github.com/SFML/SFML/actions/runs/16009339412

Relevant are the `SFML-a7201d2f3ba01b8ac46afea5cbdd132c62c9321f-*` archives.

For macOS and Linux it differs largely as this also ships static builds for those platforms, which we never did in the past.

.github/workflows/release.yml
```diff
@@ -0,0 +1,360 @@
...
+
+    # Although the CMake configuration will run with 3.24 on Windows and 3.22
+    # elsewhere, we install 3.25 on Windows in order to support specifying
+    # CMAKE_MSVC_DEBUG_INFORMATION_FORMAT which allows CCache to cache MSVC object
+    # files, see: https://cmake.org/cmake/help/latest/release/3.25.html#variables
+    - name: Get CMake and Ninja
+      uses: lukka/get-cmake@latest
+      with:
+        cmakeVersion: ${{ runner.os == 'Windows' && '3.25' || '3.22' }}
+        ninjaVersion: latest
```
ChrisThrasher: This code is import in the CI pipeline because we want to test with the lowest possible CMake version. Doing so ensures that we don't accidentally use features from a newer version. However here in the CD pipeline, we're not longer trying to test the code and rather just want to ship it. I think we can omit all this and use the default CMake version or at least we can simplify this down to just using 3.25 on all machines without the conditional logic.

.github/workflows/release.yml
```diff
@@ -0,0 +1,360 @@
...
+      run: sudo apt-get update && sudo apt-get install xorg-dev libxrandr-dev libxcursor-dev libxi-dev libudev-dev libflac-dev libvorbis-dev libgl1-mesa-dev libegl1-mesa-dev ccache
+
+    - name: Install macOS Tools
+      if: runner.os == 'macOS'
+      run: |
+        brew update
+        brew install ccache || true
+
+    # In addition to installing a known working version of CCache, this action also takes care of saving and restoring the cache for us
+    # Additionally it outputs information at the end of each job that helps us to verify if the cache is working properly
```
ChrisThrasher: Given how the CD pipeline runs much less often than CI, I don't think we need to bother with caching. Assuming it works it doesn't hurt, but it's complexity that doesn't really buy us a lot if you ask me.

.github/workflows/release.yml
```diff
@@ -0,0 +1,360 @@
...
+    - name: Add MinGW x64 to PATH and remove MinGW-supplied CCache
+      if: matrix.platform.name == 'Windows MinGW x64'
+      run: |
+        echo "C:\Program Files\mingw64\bin" >> $GITHUB_PATH
+        rm -f "C:\Program Files\mingw64\bin\ccache.exe"
+        echo "Using $(which ccache)"
+        ccache --version
+
+    - name: Configure CMake
+      run: cmake -B build -GNinja -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=Release -DSFML_BUILD_EXAMPLES=${{matrix.config.name == 'Static' && matrix.type.name == 'Release' && 'ON' || 'OFF'}} -DSFML_BUILD_TESTS=OFF -DSFML_BUILD_DOC=OFF -DCMAKE_INSTALL_PREFIX=./install ${{ matrix.platform.flags }} ${{matrix.config.flags }} ${{ matrix.type.flags }}
```
ChrisThrasher, line 129: I think we can hardcode the examples off.

.github/workflows/release.yml
```diff
@@ -0,0 +1,360 @@
...
+        - { name: macOS arm64,        os: macos-15 }
+        config:
+        - { name: Shared, flags: -DBUILD_SHARED_LIBS=ON }
+        - { name: Static, flags: -DBUILD_SHARED_LIBS=OFF }
+        type:
+        - { name: Release }
+        - { name: Debug, flags: -DCMAKE_BUILD_TYPE=Debug }
+        include:
+        - platform: { name: macOS x64, os: macos-13 }
+          config: { name: Frameworks, flags: -GNinja -DSFML_BUILD_FRAMEWORKS=ON -DBUILD_SHARED_LIBS=ON }
```
ChrisThrasher:
```suggestion
          config: { name: Frameworks, flags:-DSFML_BUILD_FRAMEWORKS=ON -DBUILD_SHARED_LIBS=ON }
```
We already hardcode `-GNinja` for all builds so no need to put it here too.

.github/workflows/release.yml
```diff
@@ -0,0 +1,360 @@
...
+        mkdir 'SFML-${{ matrix.platform.name }}-${{ matrix.config.name }}-${{ matrix.type.name }}/examples'
+        cp -r ./examples/* 'SFML-${{ matrix.platform.name }}-${{ matrix.config.name }}-${{ matrix.type.name }}/examples'
+        rm './SFML-${{ matrix.platform.name }}-${{ matrix.config.name }}-${{ matrix.type.name }}/examples/CMakeLists.txt'
+        find './SFML-${{ matrix.platform.name }}-${{ matrix.config.name }}-${{ matrix.type.name }}/examples' -type f -name CMakeLists.txt -delete
+        rm -r './SFML-${{ matrix.platform.name }}-${{ matrix.config.name }}-${{ matrix.type.name }}/examples/android'
+        rm -r './SFML-${{ matrix.platform.name }}-${{ matrix.config.name }}-${{ matrix.type.name }}/examples/include'
+        rm -r './SFML-${{ matrix.platform.name }}-${{ matrix.config.name }}-${{ matrix.type.name }}/examples/vulkan/vulkan.h'
+        rm -r './SFML-${{ matrix.platform.name }}-${{ matrix.config.name }}-${{ matrix.type.name }}/examples/island/stb_perlin.h'
+
+    - name: Packaging - Prepare Example Files for Windows
```
ChrisThrasher, line 146: I don't think the value of shipping examples outweighs the effort and complexity required to do so. I expect that a rather small percentage of users of the prebuilt binaries take advantage of this. I'd rather we omit this from our CD pipeline so that our pipeline remains easier to maintain and our install packages are lighter.

.github/workflows/release.yml
```diff
@@ -0,0 +1,360 @@
...
+        - { name: Windows MinGW x64,  os: windows-2025, flags: -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ }
+        - { name: Linux GCC,          os: ubuntu-24.04 }
+        - { name: macOS x64,          os: macos-13 }
+        - { name: macOS arm64,        os: macos-15 }
+        config:
+        - { name: Shared, flags: -DBUILD_SHARED_LIBS=ON }
+        - { name: Static, flags: -DBUILD_SHARED_LIBS=OFF }
+        type:
+        - { name: Release }
+        - { name: Debug, flags: -DCMAKE_BUILD_TYPE=Debug }
```
ChrisThrasher, line 34: Debug libs help with MSVC since you cannot mix Release and Debug but for all other platforms it may be fine otherwise only ship Release since Clang and GCC are happy to link Release libs to a Debug application.

ChrisThrasher (changes requested):
This is cool! In principle I think it's great to automate this task. I have some thoughts on the scope of this pipeline, in particular with regard to examples. I expect very few users to care about prebuilt examples. I don't think it's worth our time to continuously ship them.

eXpl0it3r:
One interesting idea proposed by @kimci86 was to use CPack to define the packaging and then all you need to do is `cmake --build build --target package`. Currently our CPack implementation only support NSIS installers. Maybe this could also be changed to support zip packaging.

.github/workflows/release.yml
eXpl0it3r: I do still think it has value for:
- Checking that SFML "works" on the user's PC
- Easily accessible SFML showcase

binary1248 also had the idea, that I kind of like, to ship the examples separately

.github/workflows/release.yml
eXpl0it3r: Right, copy paste error 😅

.github/workflows/release.yml
eXpl0it3r: I tried to remove it, but ran into failures. Can it be that Ninja requires it?

.github/workflows/release.yml
eXpl0it3r: Removed debug for macOS and Linux

I think it also doesn't work for MinGW, due to the runtime usage

.github/workflows/release.yml
eXpl0it3r: Tried again and it worked now 🤷

JonnyPtn:
Yeah cpack is definitely the way I'd approach this - A basic zip setup is very simple and would work across all platforms, and you can use separate components for examples + libraries so it consistently produces 2 zips across all platforms than can be easily distributed

eXpl0it3r:
Personally, I'd like to merge this more or less as is and provide improvements along the way.

It currently prevents the publishing of new pre-built releases, as I'm not willing to go the manual step anymore when we essentially have a fully working automation and as history shows, it's unlikely anyone else will help with this step of the release process.

I guess the alternative is, that I maintain this on my own personal fork, until it's at an acceptable level, but what's the point of hiding this process from the main repository?

Issue #3539, opened 2025-07-13
Unhandled invalid areas when loading texture from image


swordfatih:
### Prerequisite Checklist

- [x] I searched for [existing issues](https://github.com/search?q=repo%3ASFML%2FSFML&type=issues) to prevent duplicates
- [x] I searched for [existing discussions on the forum](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fen.sfml-dev.org) to prevent duplicates
- [x] I am here to report an issue and not to just ask a question or look for help (use the [forum](https://en.sfml-dev.org/forums/index.php#c3) or [Discord](https://discord.gg/nr4X7Fh) instead)

### Describe your issue here

Some cases are not handled when loading a texture from an image and an area, causing an error & exception :
- Area's size is negative 
- Area's position is out of the bounds of the image

I explain why it happens in the following sections.

### Negative size

Area's size is directly casted into `sf::Vector2u` : `resize(Vector2u(rectangle.size), sRgb)`

When negative, this causes an attempt to create a huge texture (conversion with modulo), thus the exception.

To address this, we can either:
- load the entire image if the size is negative
- use the absolute value of negative sizes, just like in `sf::Sprite::updateVertices`
- not accept negative values at all for area, by taking `sf::Rect<unsigned int>` instead of `sf::IntRect`

The issue with the latter is that it will require few `unsigned int` to `int` casts for OpenGL, which can create worse issues in this case (and it's implementation-defined).

### Position out of image bounds

Position greater than image's size creates negative size later in `sf::Texture::loadFromImage` : 
```
rectangle.size.x     = std::min(rectangle.size.x, size.x - rectangle.position.x);
rectangle.size.y     = std::min(rectangle.size.y, size.y - rectangle.position.y);
```

This then causes the same issue as with direct negative sizes.

To address this, we can also load the entire image if the area is out of bounds. 

Currently, we load the entire image only if either :
- Size is `{0, 0}`
- The area fully contains the image

### Your Environment

- OS / distro / window manager: Windows 11
- SFML version: master (3.1.0)
- Compiler / toolchain: GCC 15.1.0 x86_64-w64-mingw32 (mingw64)
- Special compiler / CMake flags: c++17

### Steps to reproduce

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow window(sf::VideoMode({920, 680}), "SFML window");

    sf::Image image;
    if (!image.loadFromFile("./unnamed.png"))
    {
[...]

PR #3541, opened 2025-07-15
Assert positive size and in-bounds position for area

swordfatih:
## Description

Added asserts to handle negative size and out of bounds position for the image sub-area in Texture::loadFromImage. 

Size: assert strictly positive sizes, as zero sizes are already handled earlier and are invalid as well.
Position: assert positions strictly less than the image size, ensuring the remaining space is non-zero.

This PR is related to the issue #3539

For reference, negative positions are adjusted to 0. So maybe we should consider asserting on negative positions as well, or consistently adjust size and position for uniform behavior.

## Tasks

-   [ ] Tested on Linux
-   [X] Tested on Windows
-   [ ] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    const sf::Image image(sf::Vector2u(10, 15));
    const sf::Texture position_out_of_bounds(image, false, {{20, 20}, {5, 10}}); // assertion
    const sf::Texture negative_size(image, false, {{20, 20}, {-5, 10}}); // assertion
}
```

### DIFF
--- a/src/SFML/Graphics/Texture.cpp
+++ b/src/SFML/Graphics/Texture.cpp
@@ -394,6 +394,10 @@ bool Texture::loadFromImage(const Image& image, bool sRgb, const IntRect& area)
     }
 
     // Load a sub-area of the image
+    assert(area.size.x > 0 && "Area size x cannot be negative");
+    assert(area.size.y > 0 && "Area size y cannot be negative");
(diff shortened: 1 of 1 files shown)


ChrisThrasher:

ChrisThrasher (approved):

PR #3542, opened 2025-07-15
Added boolean return value to sf::Http::setHost. Added test to sf::Http::setHost. Made sf::Http::sendRequest const.

Rosme:
This PR has no related issue. It was discussed in Discord as a nice to have.

## Tasks

-   [X] Tested on Linux
-   [X] Tested on Windows
-   [X] Tested on macOS
-   [ ] Tested on iOS
-   [ ] Tested on Android

## How to test this PR?

This PR does not add any side effect. It provide the result of setHost. Existing code will still work.
Unit tests have been added for the result of setHost with both a http prefix and not, both for a valid and a non valid host.

```cpp
#include <SFML/Network.hpp>

int main()
{
    sf::Http http;
    if(!http.setHost("dummy"))
    {
        sf::err() << "Failed to set host";
    }
}
```


test/Network/Http.test.cpp
```diff
@@ -14,6 +14,33 @@ TEST_CASE("[Network] sf::Http")
         STATIC_CHECK(!std::is_nothrow_move_assignable_v<sf::Http>);
     }
 
+    SECTION("Host")
+    {
+        SECTION("Valid Host Http Prefix")
+        {
+            sf::Http http;
```
ChrisThrasher: You can move `sf::Http http;` up one level such that each of these four sections can reuse this variable.

test/Network/Http.test.cpp
```diff
@@ -14,6 +14,33 @@ TEST_CASE("[Network] sf::Http")
...
+            CHECK(http.setHost("http://google.com") == true);
```
ChrisThrasher:
```suggestion
            CHECK(http.setHost("http://google.com"));
```

include/SFML/Network/Http.hpp
```diff
@@ -404,13 +407,12 @@ class SFML_NETWORK_API Http
...
     ///
     ////////////////////////////////////////////////////////////
-    [[nodiscard]] Response sendRequest(const Request& request, Time timeout = Time::Zero);
+    [[nodiscard]] Response sendRequest(const Request& request, Time timeout = Time::Zero) const;
 
 private:
     ////////////////////////////////////////////////////////////
     // Member data
     ////////////////////////////////////////////////////////////
-    TcpSocket                m_connection; //!< Connection to the host
```
ChrisThrasher, line 413: It’s cool that we can remove some internal state from this class. Can you help me understand why storing this as a class member is not necessary?

include/SFML/Network/Http.hpp
```diff
@@ -383,8 +384,10 @@ class SFML_NETWORK_API Http
     /// \param host Web server to connect to
     /// \param port Port to use for connection
     ///
+    /// \return `true` if the host has been resolved and is valid, `false` otherwise
+    ///
     ////////////////////////////////////////////////////////////
-    void setHost(const std::string& host, unsigned short port = 0);
+    bool setHost(const std::string& host, unsigned short port = 0);
```
ChrisThrasher, line 390: Do you think adding `[[nodiscard]]` here would be appropriate? Is there a valid way to use this function even if you never check the return value?

include/SFML/Network/Http.hpp
ZXShady: the variable wasn't needed in any other function except this one so why store it.

ChrisThrasher: I was wondering if it carried any relevant persistent state. Apparently not.

include/SFML/Network/Http.hpp
eXpl0it3r: Wouldn't that then be considered a breaking change?

Rosme: I would think it makes sense to add `[[nodiscard]]`. However it would be a breaking change and as such I didn't want to take the decision. I will let you decide that.

include/SFML/Network/Http.hpp
Rosme: It does not. It was a swap 1 to 1 with a local variable. Further more, it makes sense from a design point of view that sending a request would not modify the class internals.

ChrisThrasher:

include/SFML/Network/Http.hpp
ChrisThrasher: I think I'm fine with that. @eXpl0it3r thoughts?

eXpl0it3r: Even if changing the signature from `void` to `bool` is arguably a non-breaking change, I would argue that using `[[nodiscard]]` is a breaking change, as it would cause people's code to fail to compile.

Meaning, I don't want to use `[[nodiscard]]` here.

ChrisThrasher: I just want to make it clear that user code will only fail to compile if they use a specific compiler warning and enforce warnings as errors. ISO C++ lets you ignore `[[nodiscard]]` return values without your program becoming ill-formed. Due to personal tooling decisions compilation may break, but as per the rules of C++ no changes to user code are required.

As someone who uses the highest level of compiler warnings I understand I’m saying that I want the most feedback and I want the build to break for as many reasonable reasons as possible.

eXpl0it3r: Technically I agree with you. I do still see two points that let me consider `[[nodiscard]]` as a breaking change:

- The point of `[[nodiscard]]` is to communicate an "invalid" use of an API. The intention is clear that the function signature changed in a breaking way, even if toolchains can be configured to ignore it.
- I can't claim to know the default config of the standard compilers, but I'd assume that this would at least generate a warning. Relying on defaults and expecting that users who want a stricter rule set will be happy about breaks is one view/opinion, I think it could lead to more unhappy users.

ChrisThrasher: A quick test with the latest versions of GCC, Clang, and MSVC show that all of them emit warnings about unused return values by default. I was not aware they did. That shifts the calculus a bit towards your side and makes me less zealous to push for `[[nodiscard]]` here. I thought you had to opt into this warning on all compilers.

Issue #2711, opened 2023-09-27
sf::Text bounds are incorrect in case of underline and strikethrough


FRex:
## Subject of the issue

Local (and thus global) bounds of SFML text do not take underline and strikethrough into account.

## Your environment

* Windows 10, but the issue is OS independent.
* SFML from git commit 3f6403e279d3f0965938250f263cdd9fdfef5f4a (issue is present in 2.4.1 and 2.5.0 too)
* MSVC 19.37.32822  via Visual Studio 2022 but issue is compiler independent.
* No special compiler flags.

## Steps to reproduce

Make an sf::Text instance with style underline set on it, get bounds, try display that bounding box using sf::RectangleShape or similar, see that underline is outside the rectangle drawn.

A more through example with interactive changing of styles, text and outline size, and my (bad - I don't recommend it) attempt at a fix is available at https://github.com/FRex/sfml-text-bounds-with-underline

```cpp
#include <SFML/Graphics.hpp>

int main()
{
    sf::RenderWindow app(sf::VideoMode(640u, 480u), "Test");

    sf::Font font;
    font.loadFromFile("DejaVuSans.ttf");

    sf::Text txt("Test", font, 200);
    txt.setStyle(sf::Text::Underlined);
    const auto bounds = txt.getLocalBounds();

    sf::RectangleShape sha;
    sha.setFillColor(sf::Color::Red);
    sha.setPosition(bounds.left, bounds.top);
    sha.setSize(sf::Vector2f(bounds.width, bounds.height));

    txt.move(100.f, 100.f);
    sha.move(100.f, 100.f);

    while(app.isOpen())
    {
        sf::Event eve;
        while(app.pollEvent(eve))
            if(eve.type == sf::Event::Closed)
                app.close();

        app.clear();
        app.draw(sha);
        app.draw(txt);
        app.display();
    }
}
```

## Expected behavior

Underline should be included in the bounding box, so it should fall inside the drawn rectangle shape. Use case I have is having a background drawn under sf::Text using these bounds. Without underline being taken into account I can't know how big the box should be to be under all space touched by the sf::Text (underline included).

## Actual behavior

The bounds should take underline (and strikethrough) into account, like they already take bold and italic styles, and outline thickness (including negative) into account. Also scree screenshot and repo linked above for more examples.

![issuescreenshot](https://github.com/SFML/SFML/assets/2415079/2090ae42-ee41-4b3c-bfeb-83a43103c56b)

FRex:
[...]

PR #3543, opened 2025-07-16
Improved text layout and shaping

### LINKED ISSUE #246: Investigate support for complex text layouts

### LINKED ISSUE #2988: `sf::Text::findCharacterPos()` doesn't report the correct Y location

binary1248:
This change completely reworks how text is shaped i.e. glyphs are positioned to form the final geometry that ends up being drawn.

The previous implementation was adequate for simple text, focused on the Latin character set and primarily left-to-right scripts. Unicode is a very complex topic, and "text isn't just text". There are so many edge cases that have to be handled properly that the previous implementation would just start to fall apart once "non-typical" text was to be rendered.

The core of this change consists of replacing the hand-rolled shaping that has been used until now with the HarfBuzz shaping engine. HarfBuzz is used in basically every widely used piece of software that exists today, so I don't think I have to explain why using it for our purposes makes sense. It is industrial-grade and battle tested.

A pre-requisite for this change was to detach ourselves from the concept of a "character". The problem with the term "character" is that it is a very ambiguous term. Sometimes it can refer to the drawn glyph, sometimes the Unicode codepoint and when used in the context of a `char` it is merely a single-byte-wide character data type. Because Unicode aims to support every single script in existence they can't afford to work with such ambiguous terms.

Starting with how data is stored on the system, we have a stream consisting of single bytes. Without knowing how this stream is encoded, we won't be able to make sense of the data. Typical text encodings are e.g. UTF-8 and UTF-32. For simplicity sake SFML stores text data internally using UTF-32 encoding. As of today, UTF-32 is sufficient to represent all Unicode codepoints, but who can tell if this will always be the case (remember IPv4?). UTF-8 is a variable length encoding, meaning that a single Unicode codepoint can be represented by one of more bytes. UTF-32 is as of today a fixed length encoding, which means there is a 1:1 mapping between the length of a UTF-32 string and the number of codepoints in that string. Now that we have a stream of codepoints, they will have to be interpreted to form graphemes. Because it is common for scripts to have glyphs consisting of a base character and one or more modifiers, many graphemes can be represented by base and mark codepoint sequences. Basically, a base codepoint determines the base glyph that is to be rendered and any following mark codepoints will be applied to that base glyph. There isn't really a limit on how many mark codepoints are allowed to follow, enabling things like [Zalgo text](https://en.wikipedia.org/wiki/Zalgo_text).

So one might think, OK now that we have a stream of graphemes we have everything we need to draw text right? Well... yes and no. In a naive Latin script centric implementation (like the previous one), rendering a single line of graphemes would yield passable results. While it wouldn't look horrible, it also won't look as good as it could be. Staying in Latin script for the moment, graphemes can actually be combined to form so-called ligatures. Such ligatures are e.g. æ and ﬃ. Many fonts have dedicated ligature replacement tables that allow substitution of eligible grapheme sequences with their corresponding ligature. This means that it is possible that multiple Unicode codepoints end up being represented by a single glyph in the final geometry. What is an exception in Latin script is much more commonplace in other scripts, which is one of the many reasons why the older implementation was not suited to render them.

Going further, when we consider things like line-breaking or bi-directional text, we will need additional script-based rules on where segmentation boundaries occur. The main Unicode text segmentation boundaries are: grapheme, word and sentence. These boundaries are relevant when it comes to cursor positioning and line-breaking. As an example, in contrast to Latin script there exist other scripts that are also alphabetic but don't use whitespace to separate words.

Because of the above, functions like `sf::Text::findCharacterPos` and any other functions referring to characters just don't make much sense when dealing with Unicode text. This is the reason why `sf::Text::findCharacterPos` is deprecated in this change.

There are way more intricacies that have to be taken into consideration than the ones I listed above if we want to properly handle and render Unicode text, but the above basics should allow anyone to understand what the implementation is actually doing.

Another complex topic is bi-directional text. The two major right-to-left scripts in use today are Arabic and Hebrew. For a multitude of reasons, it has become the standard that right-to-left scripts incorporate left-to-right segments of text from other scripts to be able to represent e.g. numbers in an international format. For this reason, in order to properly support right-to-left scripts, bi-directional segmentation has to be performed first. Based on standardized rules, a text sequence is broken up into multiple runs with uniform direction. Arabic text with an embedded 10-digit number will be broken up into 3 runs: A right-to-left run, a left-to-right run for the number and finally a right-to-left run again. The fact that right-to-left scripts are displayed on the screen or printed on paper in right-to-left direction does not mean that their corresponding byte sequence in storage is also reversed. All code unit sequences follow a single order independent of the way their concrete script is rendered in the end. Bi-directional segmentation is performed by the SheenBidi library. It is passed a UTF-32 string of text and returns the runs of uniform direction. We then have to individually pass these runs to HarfBuzz for shaping since HarfBuzz only shapes text with uniform properties.

HarfBuzz does not perform multi-line shaping. From the perspective of HarfBuzz, it is always given a single line of text of possibly infinite length. For this reason, we still have to perform newline and tab handling ourselves. Newline characters are basically another position at which the input string has to be split into yet more runs before being passed to HarfBuzz to be shaped. Analogous to the previous implementation, tabs are still represented as 4 spaces. This substitution also has to be performed before being shaped by HarfBuzz. After shaping, the implementation goes through the effort of re-combining the 4 space "glyphs" back into a single "tab glyph". In practice, whitespace glyphs don't really exist. They are represented as either an x-advance without a y-advance or a y-advance without an x-advance depending on whether we are dealing with horizontal or vertical text.

I think those are the major points regarding the implementation. Everything else is just handling of corner cases in corner cases as is typical with Unicode text handling.

The last complex topic that had to be dealt with is the circular dependency between FreeType and HarfBuzz. HarfBuzz relies on font metrics and other tables to perform accurate shaping, and while HarfBuzz could use its own internal font implementation, since we are using FreeType already it makes sense to allow HarfBuzz to use FreeType's more advanced implementation to provide metrics and font data. What is less obvious is that FreeType can depend on HarfBuzz as well. This can be enabled to allow FreeType auto-hinting to improve its output by using the shaping data provided by HarfBuzz for complex scripts that would not have been covered by auto-hinting on its own.

CMake has ways to deal with such circular dependencies, however what is less documented is how they interact with FetchContent and export package generation. The problem of building SFML itself is solved by breaking up the dependency before running FetchContent and manually linking the created targets as necessary after it is complete. Because the export target configurations for both FreeType and HarfBuzz depend on how they are linked up during configuration time we can't use the standard `target_link_libraries` to do so. Linking them using `target_link_libraries` would cause both package configurations to require the other to be present for `find_dependency` to succeed and since a failure in `find_dependency` would cause the whole `find_package` call to abort, this wouldn't work. I searched a long time for a solution that allows us to use `find_dependency` with circular dependencies but couldn't find a solution that worked. I ended up just manually re-linking them after the targets have been created in the export package configuration as well. If anybody knows of a better alternative, feel free to provide feedback.

A top-level view of the additions in this change:
- Unicode text shaping
- Support for bi-directional text
- Support for right-alignment of multi-line right-to-left text
- Support for mixed-alignment of multi-line bi-directional text
- Support for forced right/center/left-alignment of text regardless of script
- Support for vertical text orientations (no multi-line vertical text support yet)
- Better post-shaping glyph data provided via the public API
- Added functions to retrieve the ascent/descent of a font for purposes of e.g. positioning a cursor within a line
- Added fields to sf::Font::Info to indicate whether a font supports kerning and native vertical metrics
- Grapheme/Word/Sentence boundary detection in `sf::String`
- Support for modifying per-glyph properties after shaping but before geometry is generated for them
- Support for retrieving geometry data and manipulating it using glyph shaping data
- A new text example demonstrating all of the above and providing a reference implementation of text cursor positioning and glyph hit-testing
- And a bunch more stuff that I can't think of right now

No example code will be provided here, have a look at and run the text example.

The tests could probably be expanded a bit, especially to cover the new stuff.

As usual, feedback is welcome. The code is committed to this repository so make changes as necessary.

Closes #246
Closes #2713
Closes #3053
#244 has to be reviewed.
#2711 possibly fixed, needs testing
Closes #2988 since `sf::Text::findCharacterPos` becomes deprecated and has been replaced by shaped glyph access
#3406 status has to be updated

To do:
- Add support for Emoji's and other colored fonts
- Investigate #1536


include/SFML/Graphics/Font.hpp
```diff
@@ -378,6 +443,8 @@ class SFML_GRAPHICS_API Font
     [[nodiscard]] bool isSmooth() const;
 
 private:
+    friend class Text;
```
eXpl0it3r, line 446: Doesn't this (again?) limit the use of `sf::Font` given that custom classes won't have this `friend` benefit? 

If you recall the whole "is dirty flag" discussion

include/SFML/Graphics/Text.hpp
```diff
@@ -360,7 +490,252 @@ class SFML_GRAPHICS_API Text : public Drawable, public Transformable
...
+    /// text is composed and not subdivided into their individual
+    /// components or glyphs. If positioning of the cursor within
+    /// a single grapheme e.g. a ligature is required, a more
+    /// fine-grained cluster segmentation algorithm should be used.
+    ///
+    /// The returned glyph positions are in local coordinates
+    /// (translation, rotation, scale and origin are not applied).
+    ///
+    /// \return List of shaped glyphs that make up the text
+    ///
```
eXpl0it3r, line 536: Since the algorithm is mentioned above, maybe we should add `\see setClusterGrouping`

Bambo-Borris:
> Harfbuzz is quite massive dependency, did you take a look at kb? https://github.com/JimmyLefevre/kb
> 
> Sorry for the noise, feel free to delete my comment if it's an inappropriate place to discuss this

That may be so, but it's also incredibly depended on for text shaping and I don't see it being a "massive dependency" being bad if it's essentially the gold standard for what it achieves.

include/SFML/Graphics/Text.hpp
```diff
@@ -360,7 +490,252 @@ class SFML_GRAPHICS_API Text : public Drawable, public Transformable
...
+    /// \brief Get a reference to the vertex data of this text
+    ///
+    /// The vertex data is regenerated by the text whenever it is
+    /// necessary. Any changes made to the vertex data will be
+    /// discarded whenever this happens.
+    ///
+    /// \return Reference to the vertex data of this text
+    ///
+    ////////////////////////////////////////////////////////////
+    VertexArray& getVertexData() const;
```
eXpl0it3r, line 728: Do we want return a non-const reference?

src/SFML/Graphics/Text.cpp
```diff
@@ -385,105 +803,478 @@ void Text::ensureGeometryUpdate() const
...
+        {
+            lines.back().firstCodepointOffset = segment.offset;
+            lines.back().direction            = currentDirection;
         }
 
-        // Extract the current glyph's description
-        const Glyph& glyph = m_font->getGlyph(curChar, m_characterSize, isBold);
+        // We use the index into the input string as the input cluster IDs as well
+        // This way the user will be able to map the resulting cluster IDs back to
+        // characters in the input text they provided for advanced functionality
```
eXpl0it3r, line 1164: How can the user "map the resulting cluster IDs back to characters in the input text"?

include/SFML/Graphics/Font.hpp
```diff
@@ -468,6 +535,16 @@ class SFML_GRAPHICS_API Font
...
 
+    ////////////////////////////////////////////////////////////
+    /// \brief Get the current font handle
+    ///
+    /// This is used internally by Text to shape unicode text.
+    ///
+    /// \return The currently active font handle or nullptr if there is none
+    ///
+    ////////////////////////////////////////////////////////////
+    [[nodiscard]] void* getFontHandle() const;
```
eXpl0it3r: Would it make sense to use `using FontHandle = void*;` and return `FontHandle` instead?

src/SFML/Graphics/Text.cpp
```diff
@@ -385,105 +803,478 @@ void Text::ensureGeometryUpdate() const
...
+                // Remove unwanted x-offset when dealing with vertical spaces
+                if ((shapeGlyph.direction == HB_DIRECTION_TTB) || (shapeGlyph.direction == HB_DIRECTION_BTT))
+                    glyphEntry.position.x -= shapeGlyph.offset.x;
+
+                // Check if we are dealing with the start of a tab
+                if (std::find(currentLineTabIndices.begin(), currentLineTabIndices.end(), shapeGlyph.cluster) !=
+                    currentLineTabIndices.end())
+                {
+                    // Widen the size to 4 times the advance of the space glyph
+                    glyphEntry.glyph.advance *= 4.0f;
```
eXpl0it3r, line 959: I assume the "tab" alignment as you see in word processors, is something that would need be handled by the user, correct?

HB / FT doesn't provide anything for that?

src/SFML/Graphics/Text.cpp
```diff
@@ -385,105 +803,478 @@ void Text::ensureGeometryUpdate() const
...
+                                          underlineOffset,
+                                          underlineThickness,
+                                          m_outlineThickness);
+                }
+            }
         }
 
-        // Apply the outline
-        if (m_outlineThickness != 0)
+        // If we're using the strike through style and there's a new line, draw a line across all characters
```
eXpl0it3r:
```suggestion
        // If we're using the strikethrough style and there's a new line, draw a line across all characters
```

binary1248:
Counting the lines of code in the HarfBuzz *.cc, *.hh and *.h files in the src directory I ended up at 138757 lines. That is only around 6 times the 23984 lines of kb. Many features of HarfBuzz like rasterization and the "boring expansion" have been disabled, which means many of those 138757 don't actually contribute to the build. Factor in that HarfBuzz also has inline documentation and a copyright header in every one of its files and the lines of actual code shrinks even more. All the relevant files are built as part of the harfbuzz.cc unity file so in essence it is built very similarly to kb.

For those reasons, I wouldn't consider HarfBuzz a "massive dependency". If we wanted to do so then kb would also have to be a "massive dependency" since code-wise HarfBuzz doesn't even have an order of magnitude more code than kb.

For that, HarfBuzz provides support for everything that can be thrown at us. This is important because SFML is a library itself that is consumed by application developers. Making use of libraries that have limitations would just pass those limitations on to the application developers, and in my opinion, the less constraints SFML imposes on application developers the better.

Application developers who intend on doing text shaping and rasterization themselves in their own code and have full control over the kinds of fonts they use in their application can use kb if it fulfils their requirements. If I had to guess, it was written with this use case in mind. SFML has already been using FreeType for a very long time and has established and tested glyph rendering code. There is also good synergy between FreeType and HarfBuzz. HarfBuzz was very easy to drop into an existing codebase that already made use of FreeType and did rendering.

HarfBuzz is also basically ubiquitous at this point, which means it has been extensively tested in production environments as well. This is especially important for a library that deals with Unicode since there is always going to be that edge case one did not think of when initially writing the code (I know this from first hand experience). kb's first commit was just over a month ago at this point. I'm sure it was used internally somewhere before the author went public with it, but having so little exposure means that the edge cases were only able to start coming to light a month ago.

The TL;DR is that kb wasn't written with SFML's use case in mind. It was written to be a one-stop-shop, stb-like, batteries-included library to get font handling, rasterization and text shaping up and running in application code as fast as possible.

src/SFML/Graphics/Text.cpp
binary1248: There is no universal definition for how wide a tab stop should be. The previous implementation hard coded it to 4x the width of a space so it was adopted here as well. Shaping and rasterization doesn't really deal with tabs so they all had to be converted to the equivalent spaces before shaping was done. For that reason aligning multi-line text according to tab stops is just another thing that either SFML or the user will have to implement on top of what already exists.

include/SFML/Graphics/Font.hpp
binary1248: It would. But unlike the other *Handle types that already exist in SFML, it would be pretty dangerous for the user to simultaneously access the FreeType object while we are also making use of it. Even in single-threaded code, SFML makes certain assumptions that the state of the FreeType font doesn't change unless we do it ourselves.

src/SFML/Graphics/Text.cpp
binary1248: This should be described at length with examples in the header.

include/SFML/Graphics/Text.hpp
binary1248: This is the only opportunity the user has to manipulate geometry data after it has been generated. Combined with the shaped glyph data which contains cluster IDs and vertex offsets, the user is able to know exactly which vertices they have to modify to affect a certain glyph in the geometry. This is obviously a very very advanced feature that most users would never need to touch, but the possibility is there for building libraries on top of SFML.

include/SFML/Graphics/Text.hpp
binary1248: Makes sense.

include/SFML/Graphics/Font.hpp
binary1248: Alternatively we could couple `sf::Text` and `sf::Font` using public methods only and mark those methods as "advanced". But since they are mainly for internal use they might be subject to higher levels of breakage than the typical public method.

Unfortunately C/C++ doesn't provide a nice way for 2 TUs to communicate internally without touching the publicly visible API. `public` and `private` merely affect the access control of the members. They all have to have external linkage in order to be usable in other TUs.

include/SFML/Graphics/Font.hpp
eXpl0it3r: I didn't really mean it as public interface, just that it looks nicer for internal usage to have a handle alias instead of `void*`, but maybe it's also a better sign of "here be dragons" to keep `void*` internally.

binary1248: I wouldn't mind using `FontHandle`. We just have to make it very clear that this shouldn't be treated the same as all the other `*Handle` types and can be subject to very frequent changes and as already mentioned, if the user does go ahead and touch this in their own FreeType code they shouldn't be surprised if stuff starts to randomly break.

examples/text/Text.cpp
```diff
@@ -0,0 +1,823 @@
...
+                cursorPositions.emplace_back(previousClusterPosition + sf::Vector2f{-previousClusterAdvance, 0.0f});
+                break;
+            case sf::Text::TextDirection::TopToBottom:
+                cursorPositions.emplace_back(previousClusterPosition + sf::Vector2f{0.0f, previousClusterAdvance});
+                break;
+            case sf::Text::TextDirection::BottomToTop:
+                cursorPositions.emplace_back(previousClusterPosition + sf::Vector2f{0.0f, -previousClusterAdvance});
+                break;
+            default:
+                break;
```
eXpl0it3r:
```suggestion
            case sf::Text::TextDirection::LeftToRight:
                cursorPositions.emplace_back(previousClusterPosition + sf::Vector2f{std::max(previousClusterSize.x, previousClusterAdvance), 0.0f});
                break;
            case sf::Text::TextDirection::RightToLeft:
                cursorPositions.emplace_back(previousClusterPosition + sf::Vector2f{-std::max(previousClusterSize.x, previousClusterAdvance), 0.0f});
                break;
            case sf::Text::TextDirection::TopToBottom:
                cursorPositions.emplace_back(previousClusterPosition + sf::Vector2f{0.0f, std::max(previousClusterSize.y, previousClusterAdvance)});
                break;
            case sf::Text::TextDirection::BottomToTop:
                cursorPositions.emplace_back(previousClusterPosition + sf::Vector2f{0.0f, -std::max(previousClusterSize.y, previousClusterAdvance)});
                break;
            default:
                break;
```

This seems to fix wrong positions for RTL text and Zalgo text. Not sure if it's the "correct" fix though

Before

Image

After

Image

binary1248: If it works... 🤷‍♂️

eXpl0it3r: The advance would be "right", if you were to add another character - probably the shaper would give other values though - but we're actually interested in landing visually after the last character, which may not be, where the next character "would" be placed. So either take the full width of the glyph or the advance, whatever is larger

binary1248:
This change doesn't implement line wrapping. However, line wrapping should be easily implementable in application code based on the shaped glyph data that this change provides. We might consider adding additional functionality such as line wrapping to `sf::Text` in the future.

include/SFML/Graphics/Font.hpp
eXpl0it3r: I guess we can leave it like that for now and if a discussion arises in the future, consider alternatives.

include/SFML/Graphics/Font.hpp
eXpl0it3r: I still feel like we talk past each other. Whether you use `void*` or `FontHandle` doesn't really change the fact that we SFML devs need to be careful in the usage 😄 

My proposal is merely:

Font.hpp
```cpp
class SFML_GRAPHICS_API Font
{
// ...
private:
    using FontHandle = void*; //!< Font handle used by the shaper
// ...
    [[nodiscard]] FontHandle getFontHandle() const;
};
```

Font.cpp
```cpp
// ...
Font::FontHandle Font::getFontHandle() const
{
    if (!m_fontHandles)
        return nullptr;

    return m_fontHandles->face;
}
```

It basically just hides the `void*` for SFML devs and isn't part of the public API

binary1248: I understood your proposal. What I am saying is that the function documentation for `sf::Font::getFontHandle()` would need to contain an extra warning that using the handle without care might cause unwanted side effects.

eXpl0it3r: Ah okay. I can also add that, unless you beat me to it 😄

eXpl0it3r: Added a warning to the `getFontHandle` function

eXpl0it3r (approved):
I may play around with the vertical text bounding box alignment, but otherwise, I think this is an awesome addition! 🥳 

Image

Would be great to get some eyes on it though.

binary1248:
Added a few lines to demonstrate vertex position post-processing in the text example.

examples/asset_licenses.md
```diff
@@ -5,26 +5,28 @@ All assets are under public domain (CC0):
...
-| shader/resources/devices.png         | Kenney.nl                 | [Game Icons Pack][5]       |
-| sound/resources/ding.flac            | Kenney.nl                 | [Interface Sounds Pack][6] |
-| sound/resources/ding.mp3             | Kenney.nl                 | [Interface Sounds Pack][6] |
-| keyboard/resources/error_005.ogg     | Kenney.nl                 | [Interface Sounds Pack][6] |
-| keyboard/resources/mouseclick1.ogg   | Kenney.nl                 | [UI Audio Pack][7]         |
-| keyboard/resources/mouserelease1.ogg | Kenney.nl                 | [UI Audio Pack][7]         |
-| win32/resources/image1.jpg           | Kenney.nl                 | [Toon Character Pack][8]   |
-| win32/resources/image2.jpg           | Kenney.nl                 | [Toon Character Pack][8]   |
-| sound/resources/killdeer.wav         | US National Park Services | [Bird sounds][9]           |
+| Noto fonts                           | Google                    | [Google Fonts][2]
```
kimci86:
```suggestion
| Noto fonts                           | Google                    | [Google Fonts][2]          |
```

examples/asset_licenses.md
kimci86, line 1: > All assets are under public domain (CC0):

This is not true anymore as Noto fonts are provided under the Open Font License.

examples/asset_licenses.md
binary1248: Fixed.

examples/asset_licenses.md
binary1248: Fixed.

examples/asset_licenses.md
```diff
@@ -28,3 +28,11 @@ All assets are under public domain (CC0):
...
 [8]: https://www.kenney.nl/assets/toon-characters-1
 [9]: https://www.nps.gov/subjects/sound/sounds-killdeer.htm
+
+All the following assets are under the Open Font License:
+
+| Name       | Author | Link              |
+| ---------- | ------ | ----------------- |
+| Noto fonts | Google | [Google Fonts][1] |
+
+[1]: https://fonts.google.com/noto
```
kimci86: It should be `[10]` because the link id `[1]` is already used in this document.

binary1248: Fixed.

test/Graphics/Text.test.cpp
```diff
@@ -9,6 +9,13 @@
 #include <WindowUtil.hpp>
 #include <type_traits>
 
+// Allow testing deprecated functions
+#ifdef _MSC_VER
+#pragma warning(disable : 4996)
+#else
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
ChrisThrasher, line 16: When I wrote https://github.com/SFML/SFML/pull/3518 I just removed the tests for the deprecated APIs. I think it's reasonable to stop testing them if we're also telling users to stop using them. It has the nice side-effect of not needing to work around these compiler warnings.

test/TestUtilities/GraphicsUtil.hpp
```diff
@@ -38,3 +42,66 @@ bool operator==(const sf::Rect<T>& lhs, const Approx<sf::Rect<T>>& rhs)
 {
     return lhs.position == Approx(rhs.value.position) && lhs.size == Approx(rhs.value.size);
 }
+
+template <typename T>
+struct EqualsApproxVector2Matcher : Catch::Matchers::MatcherGenericBase
```
ChrisThrasher, line 47: Can you talk about it about how `Approx(vec)` was not sufficient and why you needed to add this new helper class template?

src/SFML/Graphics/Text.cpp
```diff
@@ -30,40 +30,81 @@
 #include <SFML/Graphics/Text.hpp>
 #include <SFML/Graphics/Texture.hpp>
 
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wold-style-cast"
+#elif defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wold-style-cast"
```
ChrisThrasher: Why is the use of `SYSTEM` include dirs not sufficient to silence these 3rd party warnings?

test/TestUtilities/GraphicsUtil.hpp
binary1248: Floating point error accumulates the more operations you perform. Especially when checking metrics values over longer strings of text, the error/rounding that is done to get to that point goes beyond the standard epsilon value of `Approx()` which is less than 1 pixel. Strict equality comparison like it was done previously does not work any more since shaping can and will return sub-1-pixel positions. If we were to use `Approx()` with `.margin(1)`, it would have also needed an overload for comparing `sf::Vector2f`s and `sf::FloatRect`s. Also, according to [Catch2's own documentation](https://github.com/catchorg/Catch2/blob/devel/docs/comparing-floating-point-numbers.md#approx) using `Approx()` is not recommended for newer code.

src/SFML/Graphics/Text.cpp
binary1248: CMake 3.25 introduced the [`SYSTEM` target property](https://cmake.org/cmake/help/latest/prop_tgt/SYSTEM.html) which indicates among other things that includes originating from `target_link_libraries()` should be marked as `SYSTEM`. Before that there was no built-in way of telling CMake to apply `SYSTEM` to the `target_include_directories()` if it was done automatically. Unfortunately our CMake configuration isn't targeting CMake 3.25+ yet.

test/Graphics/Text.test.cpp
binary1248: I wouldn't mind getting rid of tests for deprecated functionality, however it would mean that we also don't guarantee that deprecated functionality will keep working until it is removed, which kind of defeats the purpose of going through the deprecation process in the first place. Users relying on the API staying stable until the next major release would also be relying on the behaviour behind that API staying usable until the next major release.

behdad:
> [](/SFML/SFML/pull/3543/files/8dd956021567cb3ed791438f9de3247bf522b960#diff-22c1fd53bbab429b4a348662278b21abb1d7a063aa14b175a8f3a5451c783b6c)
> #pragma clang diagnostic ignored "-Wold-style-cast"
> #elif defined(__GNUC__)
> #pragma GCC diagnostic push
> #pragma GCC diagnostic ignored "-Wold-style-cast"

From our experience in HarfBuzz, clang processes GCC pragmas just fine. I don't think you need both.

binary1248:
> From our experience in HarfBuzz, clang processes GCC pragmas just fine. I don't think you need both.

While this is true when building with clang on Linux, when building with clang-cl i.e. the drop-in MSVC replacement, the GCC pragmas are ignored. Our CI runs would not pass successfully without separate pragmas for clang.

behdad:
> > From our experience in HarfBuzz, clang processes GCC pragmas just fine. I don't think you need both.
> 
> While this is true when building with clang on Linux, when building with clang-cl i.e. the drop-in MSVC replacement, the GCC pragmas are ignored. Our CI runs would not pass successfully without separate pragmas for clang.

Thanks for the information.

binary1248:
@behdad I just checked again and it was actually LLVM/Clang on Windows that was failing because of that, as can be seen here:
https://github.com/SFML/SFML/actions/runs/16263110741/job/45912721520#step:18:801

It was building 6baf3c0f5165ba5fd35bee0fd82c1f94844056fb where the code still looked like this: https://github.com/SFML/SFML/blob/6baf3c0f5165ba5fd35bee0fd82c1f94844056fb/src/SFML/Graphics/Text.cpp#L33-L40

CMake reported the detected compiler as `Clang 20.1.7 with GNU-like command-line`. I can't tell if ignoring GCC pragmas on Windows is intentional or a bug. It is something we have seen many times in the past though, when we had to suppress warnings coming from other dependencies.

behdad:
I'm pushing this PR just in case it helps: https://github.com/harfbuzz/harfbuzz/pull/5431

binary1248:
> I'm pushing this PR just in case it helps: [harfbuzz/harfbuzz#5431](https://github.com/harfbuzz/harfbuzz/pull/5431)

I don't think that change will be visible to our own translation units. We have to suppress it when we build our own source which only includes `hb-ft.h` and from what I understand the line you added only applies when building HarfBuzz itself.

```
In file included from D:/a/SFML/SFML/src/SFML/Graphics/Text.cpp:37:
In file included from D:/a/SFML/SFML/build/_deps/harfbuzz-src/src\hb-ft.h:32:
In file included from D:/a/SFML/SFML/build/_deps/harfbuzz-src/src\hb.h:31:
In file included from D:/a/SFML/SFML/build/_deps/harfbuzz-src/src\hb-blob.h:34:
D:/a/SFML/SFML/build/_deps/harfbuzz-src/src\hb-common.h:526:24: error: use of old-style cast [-Werror,-Wold-style-cast]
  526 |   HB_SCRIPT_COMMON                      = HB_TAG ('Z','y','y','y'), /*1.1*/
      |                                           ^~~~~~~~~~~~~~~~~~~~~~~~
```

behdad:
Right... And since that's a C header we can't avoid those casts. Unless we wrap it in __cplusplus__, but that sounds like overkill.

binary1248:
We are considering just configuring CMake to use the appropriate system include mechanism which would automatically disable warnings originating from the headers of our dependencies. That would solve this issue for us going forward, especially when including C headers. However it would require increasing our CMake requirement to 3.25 since that is when the `SYSTEM` target property got introduced. Ideally we wouldn't want to trip over warnings/errors caused by dependencies when activating them for our own code but the C/C++ textual inclusion mechanism makes this hard without build system support as we have seen.

test/Graphics/Text.test.cpp
eXpl0it3r: I think it's not unreasonable to keep the tests, as to guarantee that the APIs continue to work, until removed.

src/SFML/Graphics/Text.cpp
eXpl0it3r: Now that CMake was updated, can you implement this with the `SYSTEM` property? And/or explain what the necessary change is, because I haven't fully understood it.

binary1248: Done.

examples/text/Text.cpp
```diff
@@ -0,0 +1,840 @@
...
+            glyphBox.setPosition({(shapedGlyph.position + shapedGlyph.glyph.bounds.position).x,
+                                  localBounds.position.y + shapedGlyph.position.y - boxHeight});
+            glyphBox.setSize({shapedGlyph.glyph.bounds.size.x, boxHeight});
+            glyphBox.setOutlineColor(sf::Color(0, 127, 255));
+        }
+        else if (shapedGlyph.glyph.bounds.size.x == 0.0f)
+        {
+            // Vertical space
+            glyphBox.setPosition({-localBounds.size.x / 2.0f,
+                                  localBounds.position.y + shapedGlyph.position.y - shapedGlyph.glyph.bounds.size.y});
```
eXpl0it3r:
```suggestion
                                  localBounds.position.y + shapedGlyph.position.y - boxHeight});
```

Using the boxHeight produces a better result. 
Should we change that?

Image

vs

Image

binary1248: If you think so. 😁

eXpl0it3r (approved):

eXpl0it3r:
Thank you for this awesome feature! 🎉

From my testing and especially the great example, it seems to work quite well. So lets get the change into the hands of more users.

oomek:
I'm a bit confused. This feature is planned for 3.1.0 release, but it has 1 breaking change which is `sf::Text::findCharacterPos` deprecation, am I correct?

Or is it still working, but marked for deletion in 4.0.0?

eXpl0it3r:
A deprecation is exactly that, a deprecation. The API continues to exist, but produces warnings to use the improved APIs and will be removed with SFML 4.

vittorioromeo:
@binary1248 fantastic work -- the results speak for themselves!

However, I've tested this branch on my game BubbleByte, where `sf::Text` objects are created/destroyed every frame, and the performance tanked. 

Profiling shows that most of the time is spend in the `std::sort` call that was introduced, is that truly necessary?

binary1248:
This PR was primarily about getting the functionality into the library at all and designing a usable API. Performance optimization was not really a priority here. There are countless places where performance can definitely be optimized given adequate profiling data and benchmarking examples, but in the recent months I have been in more of a "implement features" mode rather than a "optimize performance" mode. If nobody else gets to it first I will surely work on optimizing the performance of `sf::Text` myself at some point.

vittorioromeo:
> This PR was primarily about getting the functionality into the library at all and designing a usable API. Performance optimization was not really a priority here. There are countless places where performance can definitely be optimized given adequate profiling data and benchmarking examples, but in the recent months I have been in more of a "implement features" mode rather than a "optimize performance" mode. If nobody else gets to it first I will surely work on optimizing the performance of `sf::Text` myself at some point.

No worries, I wasn't implying that the first iteration of the feature should be optimized.

I can take a stab at some optimization too at some point -- I'm mostly curious on why the sort is necessary. I see that, without it, the example cursor selection behavior breaks, but is it necessary for correct rendering?

binary1248:
The `std::sort` takes place after the geometry has already completely been updated and glyph data is not used by post-processing. It has no impact on rendering. However, in order to support the correct ordering of graphemes according to script direction the sort is necessary to ensure the user can iterate in ascending cluster ID order. This is necessary as you said for properly positioning and moving a text cursor. If there is no intention of iterating over the shaped glyphs then the sorting is entirely unnecessary and could be completely skipped.

